PR #3751, opened 2017-07-20
Widen type to Any when isinstance(..., Any) is True

### LINKED ISSUE #3474: If isinstance target type is Any, narrow down to Any
Currently if the second argument to `isinstance` has type `Any`, we don't narrow the type down. This doesn't work as expected with ignored imports. Example (assume that `foo` has a missing stub):

```py
from typing import Union

from foo import A  # type: ignore

def f(x: Union[A, str]) -> None:
    if isinstance(x, A):
        x.method_only_in_a()  # Item "str" of "Union[Any, str]" has no attribute "method_only_in_a"
```

I'd expect the type of `x` to be 'narrowed' (actually widened) to `Any`, and mypy should probably not generate an error.

miedzinski:
Fixes #3474.

### DIFF
--- a/test-data/unit/check-isinstance.test
+++ b/test-data/unit/check-isinstance.test
@@ -1977,6 +1977,17 @@ def f() -> None:
 [out]
 
 [case testIsinstanceWidensWithAnyArg]
+from typing import Any
+class A: ...
+B: Any
+x: A
+x.foo()  # E: "A" has no attribute "foo"
+assert isinstance(x, B)
+x.foo()
+reveal_type(x)  # E: Revealed type is 'Any'
+[builtins fixtures/isinstance.pyi]
+
+[case testIsinstanceWidensUnionWithAnyArg]
 from typing import Any, Union
 class A: ...
 B: Any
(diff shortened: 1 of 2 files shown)


test-data/unit/check-isinstance.test
```diff
@@ -1770,3 +1770,13 @@ def narrow_any_to_str_then_reassign_to_int() -> None:
         reveal_type(v)  # E: Revealed type is 'Any'
 
 [builtins fixtures/isinstance.pyi]
+
+[case testIsinstanceWidensWithAnyArg]
```
ilevkivskyi: I think this is a good fix, but could you please add few more tests:
* With an error and ``# type: ignore``
* With an "un-imported" ``Any``
* With an initial type that is not a union

ilevkivskyi:
This appeared again, are you still working on this PR?

JukkaL:
@miedzinski Are you interested in finishing this PR?

miedzinski:
@JukkaL Yes, I am. I have merged master into my branch and will look into this weekend.

miedzinski:
@ilevkivskyi Do you have any ideas for more test cases?

PR #3753, opened 2017-07-21
Update mypy development status to Alpha

JukkaL:


gvanrossum:
Honestly that's a pretty conservative use of Alpha (I think of it as close to Beta myself) but I guess we're not promising strict backwards compatibility so this is not unreasonable. When do you expect to declare it Beta? I suppose it should come out of Beta when we release version 1.0. Maybe 0.600 or 0.700 should be Beta?

ethanhs:
It makes more sense to me to make it beta when we feel it is at that level, not by an arbitrary deadline. But I think that is probably going to be close to 0.600/0.700 in my experiences at least.

JukkaL:
For beta we should at least have well documented distinction between experimental and non-experimental features, and probably some degree of backward compatibility guarantees for non-experimental features.

ilevkivskyi:
Looking at the current mypy roadmap I would draw the line for Beta after protocols (this is just a coincidence :-), plus I would move the support for ORM/Django sooner.

Issue #3269, opened 2017-04-27
Can't create a type alias targeting None


JukkaL:
An alias that refers to `None` doesn't work:

```py
N = type(None)
NN = None
n: N  # Invalid type "t.N"
nn: NN  # Invalid type "t.NN"
```

ilevkivskyi:
I think we could support one of these forms. The second one ``N = None`` looks much more like a variable assignment. The first one ``N = type(None)`` looks easy to support by a simple special casing, although this is not specified by PEP 484 and a bit arbitrary.

Issue #640, opened 2015-04-17
Support type(None) as a type


JukkaL:
`type(None)` should be equivalent to `None`, as per PEP 484 draft.

JukkaL:
Or maybe we should update PEP 484 to not suggest that `type(None)` is valid as a type, and `None` is the only correct spelling? There should one -- and preferably only one -- obvious way to do it etc.

msullivan:
I don't think anyone cares

PR #3754, opened 2017-07-21
 Allow creating type aliases for None

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/3269
Fixes https://github.com/python/typing/issues/451
Also a step in the direction of #640


gvanrossum:
Looks like we just forgot about this.

PR #3755, opened 2017-07-22
Make bare None a partial type again

### LINKED ISSUE #3741: Ask for an explicit annotation if the inferred type is None
Currently mypy behaves like this
```python
class C:
    def __init__(self) -> None:
        self.x = None

    def bar(self) -> None:
        self.x = 42  # Incompatible types in assignment (expression has type "int", variable has type None)
```
This may be non-intuitive for new users. Instead, it is proposed to either:
* Request an explicit annotation if the inferred type is ``None``, like we would do for ``lst = []``.
* Use partial ``None`` types, like we would do if both assignments were within the same function.

cc @vlasovskikh

ddfisher:
Fixes #3741.  Only merge after reading discussion in #3741 and deciding this is the right thing.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2545,12 +2545,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         partial_types = self.partial_types.pop()
         if not self.current_node_deferred:
             for var, context in partial_types.items():
-                if isinstance(var.type, PartialType) and var.type.type is None:
-                    # None partial type: assume variable is intended to have type None
-                    var.type = NoneTyp()
-                else:
-                    self.msg.fail(messages.NEED_ANNOTATION_FOR_VAR, context)
-                    var.type = AnyType()
+                self.msg.fail(messages.NEED_ANNOTATION_FOR_VAR, context)
+                var.type = AnyType()
 
     def find_partial_types(self, var: Var) -> Optional[Dict[Var, Context]]:
         for partial_types in reversed(self.partial_types):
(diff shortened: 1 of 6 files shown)


gvanrossum:
This gives lots of new errors in our internal codebase, so I think this PR is dead on arrival.

ilevkivskyi:
I think a better idea might be to only do this for ``self.attr = None`` assignments. IIUC this is how PyCharm does this.

gvanrossum:
I wish we could do partial types across methods (and include class variables too).

ddfisher:
Yeah.  See [my comment in 3741](https://github.com/python/mypy/issues/3741#issuecomment-317135650).  I'm not actually convinced this is the right idea, but I put it up in case you thought it was and wanted to merge it for 0.521.

I spot-checked some of the internal errors, and a decent number were from `self.attr = None` assignments, so I'm not sure that would help.

Is this a real user-confusion issue, or is it only theoretical at this point?

ilevkivskyi:
> Is this a real user-confusion issue, or is it only theoretical at this point?

FWIW, this came form @vlasovskikh on gitter.

vlasovskikh:
@ddfisher I haven't encountered anyone besides me who complains about this problem. In PyCharm we infer `Any` for such fields instead of `None` since we don't want to force users into using type hints just to make the warnings go away. But since you already have such warnings for empty lists, it makes sense to handle the None-initialized fields in the same way. It is highly unlikely that somebody wants an attributes of type `None`, so inferring `None` little sense.

gvanrossum:
@JukkaL, what do you think?

JukkaL:
If we change this, it will potentially affect a lot of existing code, so it looks we can't really include the change in the 0.521 release. Let's reconsider this for 0.530 after 0.521 has been released. Just doing this for `self.attr = None` is a potential compromise, but again we can't include this in 0.521 because it could start generating errors for code that's clean in 0.520.

ddfisher:
I think we've basically decided we're not going to do this.

Issue #3211, opened 2017-04-21
mypy can't find typeshed directory when installed using pyenv python/pip


mdeland:
mypy can't find typeshed directory when installed via a pyenv'd python version.

On OSX.
python version 3.5.2, installed with `pyenv`

```
$ which python
>> ~/.pyenv/shims/python
```

```
$ pip install mypy
$ pip show mypy
>> Name: mypy
>> Version: 0.501
>> ...
>> Location: ~/.pyenv/versions/3.5.2/Python.framework/Versions/3.5/lib/python3.5/site-packages
$ which mypy
>> ~/.pyenv/shims/mypy
```

```
$ mypy test.py
>> test.py: error: Name '__builtins__' is not defined
>> test.py:1: error: Name 'int' is not defined
>> test.py:1: error: Cannot find module named 'builtins'
>> test.py:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
>> test.py:1: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues
>> test.py:1: note: please use --show-traceback to print a traceback when reporting a bug
```

compare to:
```
mypy test.py  --custom-typeshed-dir ~/.pyenv/versions/3.5.2/Python.framework/Versions/3.5/lib/mypy/typeshed/
```
which works fine.

gvanrossum:
You're going to have to debug this yourself, I'm not familiar with pyenv.

mdeland:
Happy to close this, but I brought up the problem in the gitter room and @JelleZijlstra suggested it should be opened as an issue.

gvanrossum:
Sorry, it's fine as an issue, and if you come up with a fix we'll happily review a PR if you send it, I just don't think the core team can fix it without your help (since it's a config issue and it's happening on your computer).

mdeland:
Got it, will do.

Socialery:
I'm having a similar issue I believe, with a user install of mypy.

https://gist.githubusercontent.com/Socialery/d94693f1db33467bd2ab5ec7d6f3a569/raw/5a3800d3d8341245631d96806842332a91246427/gistfile1.txt

grafi-tt:
I have the same problem, but the cause is different from both of @mdeland and @Socialery 's.

On Gentoo Linux, user-installed Python locates `site-packages` on the directory named `lib64`.

[...]

PR #3758, opened 2017-07-23
Search lib directory suffixed by ABI (32 or 64) used in Gentoo Linux #3211

grafi-tt:
Explanation is found on (#3211).
Though this patch doesn't completely solve #3211 as there are many different platforms ans installs, it works on my platform and probably some other platforms.


gvanrossum:
Thanks! It looks good enough for now.

PR #3081, opened 2017-03-29
Add support for NamedTuple methods

JelleZijlstra:
This is the mypy implementation of python/typing#352. cc @ilevkivskyi who wrote the runtime implementation.

Also fixes #3075.

This was pretty simple to implement.

PR #3760, opened 2017-07-24
Fix crash on incompatible redefinition in named tuple

### LINKED ISSUE #3759: Crash on namedtuples with attribute named `count`
Repro (mypy 0.520):

```python
from typing import NamedTuple

class Crash(NamedTuple):
    count: int
```

No crash if you rename the attribute, or stop subclassing NamedTuple.

Output:
```
test.py:4: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.520
  File "<...>/venv/bin/mypy", line 11, in <module>
    sys.exit(console_entry())
  File "<...>/venv/lib/python3.6/site-packages/mypy/__main__.py", line 7, in console_entry
    main(None)
  File "<...>/venv/lib/python3.6/site-packages/mypy/main.py", line 50, in main
  File "<...>/venv/lib/python3.6/site-packages/mypy/main.py", line 97, in type_check_only
    options=options)
  File "<...>/venv/lib/python3.6/site-packages/mypy/build.py", line 196, in build
  File "<...>/venv/lib/python3.6/site-packages/mypy/build.py", line 1769, in dispatch
    process_graph(graph, manager)
  File "<...>/venv/lib/python3.6/site-packages/mypy/build.py", line 2012, in process_graph
  File "<...>/venv/lib/python3.6/site-packages/mypy/build.py", line 2115, in process_stale_scc
  File "<...>/venv/lib/python3.6/site-packages/mypy/build.py", line 1684, in type_check_first_pass
    self.type_checker.check_first_pass()
  File "<...>/venv/lib/python3.6/site-packages/mypy/checker.py", line 185, in check_first_pass
    self.accept(d)
  File "<...>/venv/lib/python3.6/site-packages/mypy/checker.py", line 273, in accept
    stmt.accept(self)
  File "<...>/venv/lib/python3.6/site-packages/mypy/nodes.py", line 754, in accept
  File "<...>/venv/lib/python3.6/site-packages/mypy/checker.py", line 1109, in visit_class_def
    self.accept(defn.defs)
  File "<...>/venv/lib/python3.6/site-packages/mypy/checker.py", line 273, in accept
    stmt.accept(self)
  File "<...>/venv/lib/python3.6/site-packages/mypy/nodes.py", line 815, in accept
  File "<...>/venv/lib/python3.6/site-packages/mypy/checker.py", line 1202, in visit_block
    self.accept(s)
  File "<...>/venv/lib/python3.6/site-packages/mypy/checker.py", line 273, in accept
    stmt.accept(self)
  File "<...>/venv/lib/python3.6/site-packages/mypy/nodes.py", line 859, in accept
  File "<...>/venv/lib/python3.6/site-packages/mypy/checker.py", line 1209, in visit_assignment_stmt
  File "<...>/venv/lib/python3.6/site-packages/mypy/checker.py", line 1241, in check_assignment
    if self.check_compatibility_all_supers(lvalue, lvalue_type, rvalue):
  File "<...>/venv/lib/python3.6/site-packages/mypy/checker.py", line 1331, in check_compatibility_all_supers
    base_type, base_node = self.lvalue_type_from_base(lvalue_node, base)
  File "<...>/venv/lib/python3.6/site-packages/mypy/checker.py", line 1418, in lvalue_type_from_base
    itype = map_instance_to_supertype(instance, base)
  File "<...>/venv/lib/python3.6/site-packages/mypy/maptype.py", line 15, in map_instance_to_supertype
    if instance.type == superclass:
AttributeError: 'TupleType' object has no attribute 'type'
```

ilevkivskyi:
Fixes #3759 

Currently we don't support generic named tuples, so using a fallback to instance is a reasonable solution.
(Later, when we will support generic named tuples, maybe we could use item types to get a more precise type).

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1440,8 +1440,11 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
 
             if base_type:
                 if not has_no_typevars(base_type):
-                    # TODO: Handle TupleType, don't cast
-                    instance = cast(Instance, self.scope.active_self_type())
+                    self_type = self.scope.active_self_type()
+                    if isinstance(self_type, TupleType):
+                        instance = self_type.fallback
+                    else:
+                        instance = self_type
                     itype = map_instance_to_supertype(instance, base)
                     base_type = expand_type_by_instance(base_type, itype)
 
--- a/test-data/unit/check-namedtuple.test
+++ b/test-data/unit/check-namedtuple.test
@@ -412,6 +412,12 @@ b = B._make([''])  # type: B
 
 [builtins fixtures/list.pyi]
 
+[case testNamedTupleIncompatibleRedefinition]
+from typing import NamedTuple
+class Crash(NamedTuple):
+    count: int  # E: Incompatible types in assignment (expression has type "int", base class "tuple" defined the type as Callable[[Tuple[Any, ...], Any], int])
+[builtins fixtures/tuple.pyi]
+
 [case testNamedTupleInClassNamespace]
 # https://github.com/python/mypy/pull/2553#issuecomment-266474341
 from typing import NamedTuple
--- a/test-data/unit/fixtures/tuple.pyi
+++ b/test-data/unit/fixtures/tuple.pyi
@@ -1,6 +1,6 @@
 # Builtins stub used in tuple-related test cases.
 
-from typing import Iterable, Iterator, TypeVar, Generic, Sequence
+from typing import Iterable, Iterator, TypeVar, Generic, Sequence, Any
 
 Tco = TypeVar('Tco', covariant=True)
 
@@ -13,6 +13,7 @@ class type:
 class tuple(Sequence[Tco], Generic[Tco]):
     def __iter__(self) -> Iterator[Tco]: pass
     def __getitem__(self, x: int) -> Tco: pass
+    def count(self, obj: Any) -> int: pass
 class function: pass
 
 # We need int and slice for indexing tuples.


gvanrossum:
This bug is a regression since 0.511, right? Is it worth cherrypicking the fix into the 0.521 release?

ilevkivskyi:
> This bug is a regression since 0.511, right?

I didn't do bisect, but I think this is very old, the code affected was last changed almost half year ago.

gvanrossum:
Actually it seems to be caused by 3589c08e4fc398ab2acab8f8d9ead3ee71d24ee2, Add support for NamedTuple methods (#3081) by @JelleZijlstra.

ilevkivskyi:
So actually few days after 0.511. Then formally speaking this is a regression.

gvanrossum:
So how confident are you? Anyway I'm testing this against the Dropbox codebase.

ilevkivskyi:
I think this should be quite safe (at least for sure not less safe than it was). There are only three possibilities: ``Instance``, ``TupleType``, and ``None``. ``None`` can't happen, since this function is only called from within a class.

gvanrossum:
Merged, as it did not introduce new errors in our codebase. I will also cp.

JelleZijlstra:
Thanks for fixing the crash Ivan!

Arguably we should give an error for this case, since it breaks LSP. (If you take a tuple as an argument and use `.count()` on it, it will break if your tuple is actually a NamedTuple with a `count` attribute.) But that's less urgent.

ilevkivskyi:
> But that's less urgent.

The error is shown (see test). I just re-use the fallback ``Instance`` to resolve the attribute type.

PR #3765, opened 2017-07-24
mypy and mypy_extensions aren't posix only

graingert:


ethanhs:
Alternatively, we can replace it with `platforms=['any'],`. Since we support windows and Mac OS, we probably should indicate in the package metadata that is so.

graingert:
I guess it depends on whether you take the flag to mean platforms required,
or platforms supported

ethanhs:
@graingert If you merge master and update this today, it could probably make it in time for 0.570.

gvanrossum:
I fixed the merge conflict. Once the tests pass it can be merged.

PR #3129, opened 2017-04-04
[WIP] Fix crash when joining tuple and NamedTuple

pkch:
Fixes #3117 

First commit intentionally will fail because I'm modifying builtins stub to make the test fail (to match what happens with production code).

Only after that, I'll push commit that actually fixes the failing test.

PR #3767, opened 2017-07-26
Make tuple generic in most stubs

pkch:
Per @ilevkivskyi suggestion, making tuples generic in stubs to reduce incompatibility between stubs and production.

This is a carve-out from #3129.

### DIFF
--- a/test-data/unit/fixtures/async_await.pyi
+++ b/test-data/unit/fixtures/async_await.pyi
@@ -1,6 +1,7 @@
 import typing
 
 T = typing.TypeVar('T')
+U = typing.TypeVar('U')
 class list(typing.Sequence[T]): pass
 
 class object:
@@ -9,8 +10,8 @@ class type: pass
 class function: pass
 class int: pass
 class str: pass
-class dict: pass
-class set: pass
+class dict(typing.Generic[T, U]): pass
+class set(typing.Generic[T]): pass
 class tuple(typing.Generic[T]): pass
 class BaseException: pass
 class StopIteration(BaseException): pass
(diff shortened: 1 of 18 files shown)


test-data/unit/fixtures/async_await.pyi
```diff
@@ -11,7 +11,7 @@ class int: pass
 class str: pass
 class dict: pass
```
ilevkivskyi: Maybe you could also fix ``dict`` and ``set`` (on next line)? There are few of them.

test-data/unit/fixtures/fine_grained.pyi
```diff
@@ -20,7 +23,7 @@ class str:
 
 class float: pass
 class bytes: pass
-class tuple: pass
+class tuple(Generic[T]): pass
 class function: pass
 class ellipsis: pass
-class list: pass
```
ilevkivskyi: It looks like there is still one more ``tuple`` not fixed in ``isinstancelist.pyi``.

test-data/unit/fixtures/async_await.pyi
ilevkivskyi: The idea is that if you also fix few ``dict``s and ``set``s, then we could turn on the asserts that are now commented out.

gvanrossum:
Can you please measure how much slower this made the tests?

ilevkivskyi:
@gvanrossum 
> Can you please measure how much slower this made the tests?

Before: 3m7s
After: 3m9s

The full test runs two more seconds, i.e. less than 1% of total time.

gvanrossum:
OK, that's fine then. In general the tests have become slower over time, in part because mypy itself became slower; but IIRC we had a recent test speedup, and this slowdown is negligible.

PR #3778, opened 2017-07-30
Remove redundant casts and unused ignores

gvanrossum:
This resulted from running with `--warn-redundant-casts` and
`--warn-unused-ignores`.  And turn those on in the self-check.

### DIFF
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -177,7 +177,7 @@ class ConditionalTypeBinder:
             assert type is not None
             if isinstance(self.declarations.get(key), AnyType):
                 # At this point resulting values can't contain None, see continue above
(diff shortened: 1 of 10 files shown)


ilevkivskyi:
From the test failures it looks like ``--warn-redundant-casts`` has some bugs in interactions with ``None``, and in union simplifications. Concerning the second, I could guess that it uses ``is_same_type``, while we need something like ``is_identical_type``, actually the protocols PR implements ``__eq__`` for most types, that will solve this problem.

gvanrossum:
I reverted a few changes, now it seems to work.

ilevkivskyi:
> I reverted a few changes, now it seems to work.

OK, the solution was much simpler than I thought :-)

PR #3779, opened 2017-07-30
Support from ... import ... with module level __getattr__

ethanhs:
PEP 484 specifies one can `def __getattr__(name): ...` to specify that
unresolved names should be Any (or the return type of the `__getattr__`
method). This change adds support for access to these attributes via the
`from module import name` pattern. The `module.name` pattern is already
supported.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1377,7 +1377,23 @@ class SemanticAnalyzer(NodeVisitor[None]):
                     self.add_submodules_to_parent_modules(possible_module_id, True)
                 elif possible_module_id in self.missing_modules:
                     missing = True
-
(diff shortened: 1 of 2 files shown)


ethanhs:
Ah, okay. I tried to keep it in the style, but perhaps I added too many. I'll make that change.

ethanhs:
Okay, I kept spaces between `[file...]`, `[case...]`, and `[builtins...]`, to be consistent.

ilevkivskyi:
I will merge when tests pass.

ilevkivskyi:
> I will merge when tests pass.

Except that they failed :-) Could you please fix the line numbers in the test output?

ethanhs:
Sorry about the line number problems. Should be good now!

Issue #1673, opened 2016-06-07
Switch to pytest from our homegrown test runner


kirbyfan64:
Ok, so when I saw the PR a while back that "parallelized" the test runner, I was super hyped. But the hype wore off when I realized that the unit tests were still run sequentially.

In particular, it continuously bugs me how, when running the full unit test suite, there's absolutely no sort of progress indicator of any kind. To make things worse, `runtests.py` is largely undocumented, leaving me to wonder why stuff like `./runtests.py unit-test -a '*something*'` works but not `./runtests.py -a '*something*'`.

In addition, to an extent, `-a` is slightly...useless. I mean, because the naming convention for tests is so inconsistent, it's rarely ever useful.

Also, test fixtures are a cruel mess. It took me 30 minutes of debugging to figure out why the tests were giving out obscure `KeyError`s that I couldn't reproduce in my own code. Turns out, the fixture I was using didn't define `list`. But defining `list` caused other tests to brutally fail, so then I just created a new fixture. IMO this whole thing should either be reworked or (better yet) completely gutted and thrown into a fire pit. A very _hot_ fire pit. With sulfur.

So I see two options:
- Move to a different test runner. Some have suggested pytest, which I personally _really_ like! Pytest has the [xdist plugin](http://pytest.readthedocs.io/en/latest/xdist.html), which supports running multiple tests in parallel.
- Revamp the current test runner. Ideally, this would include fixing fixtures (oh, the irony...), improving `runtests.py`, and making the actual unit tests run in parallel, preferably also with a progress bar of some sort.

Thoughts? I could try and help out here a bit if you guys come to a consensus of some sort.

gnprice:
Yep -- I think we all agree that the current test runner is pretty frustrating. Last week at the PyCon sprints where we had a bunch of new contributors getting involved, I found myself apologizing every time I explained how to filter tests (with the positional argument and with `-a`).

We were actually just discussing this also on #1668, and so far everyone agrees that the right answer will be to move to something like pytest. If you want to try putting together a PR, that would be very useful!

Pytest is (mostly for good reasons) kind of a complicated beast with a lot of choices in how to set it up, so I recommend discussing early and often, either on this thread or on a work-in-progress PR.

ddfisher:
I also hate hate hate the builtins test fixtures. +1 to killing them with fire.  They only exist for perf reasons, and I'm not at all sure they're worth it.

gvanrossum:
[...]

PR #3780, opened 2017-07-30
Move testparse to pytest

elazarg:
IIUC, this is an item from #1673. I tried to keep this minimal.

(If this goes well I can add more)

### DIFF
--- a/mypy/test/data.py
+++ b/mypy/test/data.py
@@ -556,7 +556,7 @@ class MypyDataCase(pytest.Item):  # type: ignore  # inheriting from Any
 
 
 class DataSuite:
-    def __init__(self, *, update_data: bool = False) -> None:
+    def __init__(self, *, update_data: bool) -> None:
         self.update_data = update_data
 
     @classmethod
--- a/mypy/test/data.py
+++ b/mypy/test/data.py
@@ -556,6 +556,9 @@ class MypyDataCase(pytest.Item):  # type: ignore  # inheriting from Any
 
 
 class DataSuite:
+    def __init__(self, *, update_data: bool = False) -> None:
+        self.update_data = update_data
+
     @classmethod
     def cases(cls) -> List[DataDrivenTestCase]:
         return []
--- a/mypy/test/testfinegrained.py
+++ b/mypy/test/testfinegrained.py
@@ -35,9 +35,6 @@ files = [
 
 
 class FineGrainedSuite(DataSuite):
-    def __init__(self, *, update_data: bool) -> None:
-        pass
-
     @classmethod
     def cases(cls) -> List[DataDrivenTestCase]:
         c = []  # type: List[DataDrivenTestCase]
--- a/mypy/test/testparse.py
+++ b/mypy/test/testparse.py
@@ -7,24 +7,28 @@ from typing import List
 from mypy import defaults
 from mypy.myunit import Suite, AssertionFailure
 from mypy.test.helpers import assert_string_arrays_equal
-from mypy.test.data import parse_test_cases, DataDrivenTestCase
+from mypy.test.data import parse_test_cases, DataDrivenTestCase, DataSuite
 from mypy.test import config
 from mypy.parse import parse
 from mypy.errors import CompileError
 from mypy.options import Options
 
 
-class ParserSuite(Suite):
+class ParserSuite(DataSuite):
     parse_files = ['parse.test',
                    'parse-python2.test']
 
-    def cases(self) -> List[DataDrivenTestCase]:
+    @classmethod
+    def cases(cls) -> List[DataDrivenTestCase]:
         # The test case descriptions are stored in data files.
         c = []  # type: List[DataDrivenTestCase]
-        for f in self.parse_files:
+        for f in cls.parse_files:
             c += parse_test_cases(
                 os.path.join(config.test_data_prefix, f), test_parser)
         return c
+
+    def run_case(self, testcase: DataDrivenTestCase) -> None:
+        test_parser(testcase)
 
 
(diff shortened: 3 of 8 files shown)


mypy/test/data.py
```diff
@@ -556,6 +556,9 @@ def repr_failure(self, excinfo: Any) -> str:
 
 
 class DataSuite:
+    def __init__(self, *, update_data: bool = False) -> None:
```
gvanrossum: Since you're adding this here you should update all the classes that derive from it. Or you can follow the existing pattern and leave it out here but add such a `__init__` method to `ParserSuite` (like all other subclasses of `DataSuite` do). It seems these classes all need an `update_date` keyword argument because of the call on line 533 in data.py (line number from master, it's in `runtest()` in `MypyDataCase`).

Note that these classes don't seem to be instantiated anywhere, I suppose they are recognized dynamically. Do you know how?

mypy/test/testparse.py
```diff
@@ -7,25 +7,29 @@
...
 from mypy.options import Options
 
 
-class ParserSuite(Suite):
+class ParserSuite(DataSuite):
     parse_files = ['parse.test',
                    'parse-python2.test']
 
-    def cases(self) -> List[DataDrivenTestCase]:
+    @classmethod
```
gvanrossum: Does it actually matter whether this is a class method or an instance method? (I don't know -- can you tell me?)

mypy/test/data.py
elazarg: I've added it here since I believe that at the end of the road no class will need it, or any other argument;. So for the time being, a prototype `__init__` in the base class (just for that keyword) should suffice.

elazarg: These classes are gathered by pytest, passed to pytest_pycollect_makeitem as the `obj` argument to `pytest_pycollect_makeitem` which returns a `MypyDataSuite` wrapping this `obj`. Later the `collect` method is called on the `MypyDataSuite`, and the (our) implementation iterates over `self.obj.cases()`, wrapping and yielding the results.

mypy/test/testparse.py
elazarg: It's called without an instance (line 515 in data.py, obj is the class itself); it could have been a static method. Better yet, it could not exist at all, since the needed information here is only the file names and 2-3 options.

elazarg:
I don't know if that would help the review in any way, but [here](https://github.com/elazarg/mypy/tree/pytest/mypy) is an almost-complete migration of everything to pytest, plus making individual tests run in 0.5s instead of 4.

It's "almost", and it's one huge step, and it's outdated. So I'm trying* to do it step by step.
(It was also a couple of months ago so I don't remember all the details).

\* (I really am :) )

elazarg:
Maybe there's a way to automate testing-the-tests to avoid regression. As a sanity check I randomly put errors in the tests and watch them fail. This is not systematic enough though.

mypy/test/testparse.py
elazarg: For now it has to be a classmethod since it's this is how it's declared in DataSuite.

elazarg:
Also, running

    pytest -n0 --collect-only

gives a list of collected tests:
```
<Module 'mypy/test/testparse.py'>
  <MypyDataSuite 'ParserSuite'>
    <MypyDataCase 'testEmptyFile'>
    ...
  <MypyDataSuite 'ParseErrorSuite'>
    <MypyDataCase 'testInvalidFunction'>
    ...
```

mypy/test/testparse.py
gvanrossum: Oh, I think I realize what confused me. There's a class `Suite` with a `cases()` instance method, and a class `DataSuite` with a cases()` class method. The two are not directly related (though I suspect that some duck typing may still be going on). Anyway, no need to change this back. :-)

mypy/test/data.py
gvanrossum: Yeah, that explanation of how they are called makes sense.

I still think that now you're adding `__init__` to `DataSuite`, all subclasses of it that also define `__init__` should be modified to call `super().__init__(update_data=update_data)`. Otherwise my spider-sense about inheritance is unhappy.

elazarg: The only class that actually defines a nontrivial `__init__` is `ASTMergeSuite` (and I'm not sure it should, since it does not receive arguments and does not seem to have any mutable state).
May I simply remove the empty `__init__` definitions?

gvanrossum: I think you should remove those `__init__` definitions that are now a copy of `DataSuite.__init__`, and make the others call `super().__init__(update_data=update_data)`. I think there may be one or two that ignore their `update_data` argument, please figure out if those simply don't need it or if setting it would just be ignored.

elazarg: Currently the only test that uses this option is `testcheck`, which uses `update_testcase_output` if it's set. The others simply ignore it.

mypy/test/testfinegrained.py
```diff
@@ -35,9 +35,6 @@
 
 
 class FineGrainedSuite(DataSuite):
-    def __init__(self, *, update_data: bool) -> None:
-        pass
-
```
elazarg: Or did you want here a call to `__init__`?

elazarg:
Ok. If everything goes well, please let me know how do you want to continue. There are 3 more data-driven test (testtransform, testtypegen, testsemanal), all of them are similarly straightforward. Do you want a dedicated PR for each of them or a single one?

gvanrossum:
Thanks! I much prefer to review smaller pieces (easier to get them done between meetings :-).

ilinum:
Thanks for doing this @elazarg! I will be so excited once we can fully switch to pytest.

PR #3782, opened 2017-07-31
Fix crash with unanalyzed branches and old-html report

ilinum:
I think the right solution would be to remove `--old-html-report` altogether. What do you think?


gvanrossum:
Fine with me to remote --old-html-report.

PR #3786, opened 2017-08-01
Tag Anys with their provenance

ilinum:
This PR tags each instance of `AnyType` with where it comes from.
Each `Any` can be one (and only one) of the following:
* implicit (if this Any type was inferred without a type annotation)
* explicit (if this Any comes from an explicit `Any` in a type annotation)
* from_unimported_type (if this `Any` comes from an unfollowed import)
* from_omitted_generics (if this `Any` comes from an ommitted generic suchs as `Any` in `List`)
* from_error (if this `Any` comes from an error)
* special_form (if this `Any` is  created for internal mypy purposes)

Plus, this PR contains an addition to `AnyExpressionsReport` to report most common "tags" of Anys (implicit, explicit, from_unimported_type, etc).

Note that I was not able to identify some of the `Any`s, specifically, `Any`s related to generators. I added a todo to each one with the following exact words(`# todo: is type of this Any correct?`). All of these should probably be resolved until this PR is ready to be merged.

Also, perhaps, it would be better to split the PR into two? (one would tag Anys with their provenance and the other PR would add the new kind of Any Exprs Report)

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3264,7 +3264,7 @@ def nothing() -> Iterator[None]:
 def is_typed_callable(c: Optional[Type]) -> bool:
     if not c or not isinstance(c, CallableType):
         return False
-    return not all(isinstance(t, AnyType) and t.implicit
+    return not all(isinstance(t, AnyType) and t.type_of_any == TypeOfAny.implicit
                    for t in c.arg_types + [c.ret_type])
 
 
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -740,9 +740,15 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
 
                 # Type check initialization expressions.
                 for arg in item.arguments:
-                    init = arg.initialization_statement
-                    if init:
-                        self.accept(init)
+                    if arg.initializer is not None:
+                        name = arg.variable.name()
+                        msg = 'Incompatible default for '
+                        if name.startswith('__tuple_arg_'):
+                            msg += "tuple argument {}".format(name[12:])
+                        else:
+                            msg += 'argument "{}"'.format(name)
+                        self.check_simple_assignment(arg.variable.type, arg.initializer,
+                            context=arg, msg=msg, lvalue_name='argument', rvalue_name='default')
 
             # Type check body in a new scope.
             with self.binder.top_frame_context():
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -32,7 +32,7 @@ from mypy import nodes
 from mypy.typeanal import has_any_from_unimported_type, check_for_explicit_any
 from mypy.types import (
     Type, AnyType, CallableType, FunctionLike, Overloaded, TupleType, TypedDictType,
-    Instance, NoneTyp, strip_type, TypeType,
+    Instance, NoneTyp, strip_type, TypeType, TypeOfAny,
     UnionType, TypeVarId, TypeVarType, PartialType, DeletedType, UninhabitedType, TypeVarDef,
     true_only, false_only, function_type, is_named_instance, union_items
 )
@@ -422,11 +422,12 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         """
         if is_coroutine:
             # This means we're in Python 3.5 or later.
-            at = self.named_generic_type('typing.Awaitable', [AnyType()])
+            at = self.named_generic_type('typing.Awaitable', [AnyType(TypeOfAny.special_form)])
             if is_subtype(at, typ):
                 return True
         else:
-            gt = self.named_generic_type('typing.Generator', [AnyType(), AnyType(), AnyType()])
+            any_type = AnyType(TypeOfAny.special_form)
+            gt = self.named_generic_type('typing.Generator', [any_type, any_type, any_type])
             if is_subtype(gt, typ):
                 return True
         return isinstance(typ, Instance) and typ.type.fullname() == 'typing.AwaitableGenerator'
@@ -437,7 +438,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         True if `typ` is a supertype of AsyncGenerator.
         """
         try:
-            agt = self.named_generic_type('typing.AsyncGenerator', [AnyType(), AnyType()])
+            any_type = AnyType(TypeOfAny.special_form)
+            agt = self.named_generic_type('typing.AsyncGenerator', [any_type, any_type])
         except KeyError:
             # we're running on a version of typing that doesn't have AsyncGenerator yet
             return False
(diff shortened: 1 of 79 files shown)


mypy/report.py
```diff
@@ -156,14 +160,66 @@ def on_file(self,
...
+                line += ('{:>{}} ' * len(widths)).format(*itertools.chain(*zip(values, widths)))
+                f.write(line + '\n')
+
+            f.write(separator)
+            line = '{:{filename_width}} '.format(total_row_name, filename_width=filename_width)
+            values = [total_counter[typ] for typ in any_types_column_names]
+            line += ('{:>{}} ' * len(widths)).format(*itertools.chain(*zip(values, widths)))
+            f.write(line + '\n')
+
+    def _report_any_exprs(self) -> None:
```
ddfisher: Nit: this should probably go above the previous function.

mypy/report.py
```diff
@@ -156,14 +160,66 @@ def on_file(self,
...
 
     def on_finish(self) -> None:
+        self._report_any_exprs()
+        self._report_types_of_anys()
+
+    def _report_types_of_anys(self) -> None:
+        # On the following line, mypy complains about implicit generic Any and wants us to use
+        # typing.Counter() instead of collections.Counter().
+        # We cannot use typing.Counter() because it doesn't work with python 3.5 and older.
+        total_counter = collections.Counter()  # type: ignore
```
ddfisher: Instead, give this `# type: typing.Counter[str]` (or whatever the item type is, exactly).

mypy/report.py
```diff
@@ -147,6 +150,7 @@ class AnyExpressionsReporter(AbstractReporter):
     def __init__(self, reports: Reports, output_dir: str) -> None:
         super().__init__(reports, output_dir)
         self.counts = {}  # type: Dict[str, Tuple[int, int]]
+        self.any_types_counter = {}  # type: Dict[str, Any]
```
ddfisher: This should be precisely typed.

mypy/report.py
```diff
@@ -156,14 +160,66 @@ def on_file(self,
...
+        any_types_column_names = collections.OrderedDict([
+            (TypeOfAny.implicit, "Unannotated"),
+            (TypeOfAny.explicit, "Explicit"),
+            (TypeOfAny.from_unimported_type, "Unimported"),
+            (TypeOfAny.from_omitted_generics, "Omitted Generics"),
+            (TypeOfAny.from_error, "Error"),
+            (TypeOfAny.special_form, "Special Form"),
+            (TypeOfAny.from_another_any, "From Another Any"),
+        ])  # type: collections.OrderedDict[TypeOfAny, str]
+        file_column_name = "Name"
```
ddfisher: I wonder if the rest of this function could be abstracted out/combined with similar looking code from the function below.

mypy/checker.py
```diff
@@ -422,11 +422,14 @@ def is_generator_return_type(self, typ: Type, is_coroutine: bool) -> bool:
...
         if is_coroutine:
             # This means we're in Python 3.5 or later.
-            at = self.named_generic_type('typing.Awaitable', [AnyType()])
+            # todo: is type of this Any correct?
+            at = self.named_generic_type('typing.Awaitable', [AnyType(TypeOfAny.special_form)])
             if is_subtype(at, typ):
                 return True
         else:
-            gt = self.named_generic_type('typing.Generator', [AnyType(), AnyType(), AnyType()])
+            # todo: is type of this Any correct?
```
ddfisher: These `Any`s don't escape the function -- special form seems right here.

mypy/checker.py
```diff
@@ -437,7 +440,9 @@ def is_async_generator_return_type(self, typ: Type) -> bool:
         True if `typ` is a supertype of AsyncGenerator.
         """
         try:
-            agt = self.named_generic_type('typing.AsyncGenerator', [AnyType(), AnyType()])
+            # todo: is type of this Any correct?
+            any_type = AnyType(TypeOfAny.special_form)
```
ddfisher: This `Any` doesn't escape the function -- special form seems right here.

mypy/checker.py
```diff
@@ -468,23 +474,25 @@ def get_generator_yield_type(self, return_type: Type, is_coroutine: bool) -> Typ
             # parameters (i.e. is `object`), then the yielded values can't
             # be accessed so any type is acceptable.  IOW, ty is Any.
             # (However, see https://github.com/python/mypy/issues/1933)
-            return AnyType()
+            # todo: is type of this Any correct?
+            return AnyType(TypeOfAny.special_form)
```
ddfisher: It's not clear to me that this should even return `Any`, but I think special form is fine here.

mypy/subtypes.py
```diff
@@ -204,7 +203,8 @@ def visit_tuple_type(self, left: TupleType) -> bool:
                 if right.args:
                     iter_type = right.args[0]
                 else:
-                    iter_type = AnyType()
+                    # todo: is type of this Any correct?
+                    iter_type = AnyType(TypeOfAny.from_error)
```
ddfisher: This should be special form.

ilevkivskyi: I think that _ideally_ this should be ``assert False``, since all these types have one type parameter and types with incorrect type argument numbers (e.g. zero) should be "patched" in the third pass of semantic analysis. There were some efforts recently to weed out all instances with incorrect type argument count (including test fixtures), but maybe some are still there.

ilinum: @ilevkivskyi, hmm I see. I think it's better to have a separate PR for that one. This doesn't seem to be relevant to this PR and this PR is already huge.

mypy/report.py
ilinum: Done!

mypy/report.py
ilinum: Done!

mypy/report.py
ilinum: Done! I think code is much cleaner now :)

mypy/report.py
ilinum: Yep!

mypy/checker.py
ilinum: Done!

mypy/checker.py
ilinum: Done!

mypy/checker.py
ilinum: Done!

ilinum:
All right, all code review feedback should be addressed now!

mypy/semanal.py
```diff
@@ -3520,7 +3531,7 @@ def builtin_type(self, fully_qualified_name: str) -> Instance:
         sym = self.lookup_fully_qualified(fully_qualified_name)
         node = sym.node
         assert isinstance(node, TypeInfo)
-        return Instance(node, [AnyType()] * len(node.defn.type_vars))
+        return Instance(node, [AnyType(TypeOfAny.implicit)] * len(node.defn.type_vars))
```
ddfisher: I think this should be special form.

mypy/types.py
```diff
@@ -1772,10 +1778,10 @@ def callable_type(fdef: mypy.nodes.FuncItem, fallback: Instance,
         name = '"{}"'.format(name)
 
     return CallableType(
-        [AnyType()] * len(fdef.arg_names),
+        [AnyType(TypeOfAny.implicit)] * len(fdef.arg_names),
         fdef.arg_kinds,
         [None if argument_elide_name(n) else n for n in fdef.arg_names],
-        ret_type or AnyType(),
+        ret_type or AnyType(TypeOfAny.implicit),
```
ddfisher: I think this (and the other `Any` in this function) should be special form.

mypy/semanal.py
```diff
@@ -1191,7 +1194,7 @@ def named_type(self, qualified_name: str, args: List[Type] = None) -> Instance:
         if args:
             # TODO: assert len(args) == len(node.defn.type_vars)
             return Instance(node, args)
-        return Instance(node, [AnyType()] * len(node.defn.type_vars))
+        return Instance(node, [AnyType(TypeOfAny.implicit)] * len(node.defn.type_vars))
```
ddfisher: I thiiiink this should be special form.

ilinum: Done!

mypy/semanal.py
ilinum: Done!

mypy/types.py
ilinum: Done!

ilinum: Okay, after making this change, tests were failing, so I made them implicit again.

ddfisher:
👍

ddfisher:
🎉

PR #3788, opened 2017-08-01
Move testtransform to pytest

elazarg:
Another item from #1673, following #3780

I have checked that some tests fail by changing `testGlobalVariable` from 'semanal-basic.test' and running

    ./runtests.py pytest -a -n0 -a -k -a testGlobalVariable

(`testsemanal` is not a pytest yet)

### DIFF
--- a/mypy/test/testtransform.py
+++ b/mypy/test/testtransform.py
@@ -6,20 +6,18 @@ from typing import Dict, List
 
 from mypy import build
 from mypy.build import BuildSource
-from mypy.myunit import Suite
 from mypy.test.helpers import (
     assert_string_arrays_equal, testfile_pyversion, normalize_error_messages
 )
-from mypy.test.data import parse_test_cases, DataDrivenTestCase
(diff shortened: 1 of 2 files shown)


elazarg:
@gvanrossum

mypy/test/testtransform.py
```diff
@@ -30,15 +28,19 @@ class TransformSuite(Suite):
...
                        'semanal-python2.test']
 
-    def cases(self) -> List[DataDrivenTestCase]:
+    @classmethod
+    def cases(cls) -> List[DataDrivenTestCase]:
         c = []  # type: List[DataDrivenTestCase]
-        for f in self.transform_files:
+        for f in cls.transform_files:
             c += parse_test_cases(os.path.join(test_data_prefix, f),
                                   test_transform,
```
elazarg: the argument `test_transform`, which corresponds to the parameter  `perform`, is not used in pytests. The parameter is propagated to DataDrivenTestCase which uses it in the method `run()`. This method is not called; maybe we should add `assert False` there.

Issue #626, opened 2015-03-29
Allow using # type: ignore to skip type checking in a file


JukkaL:
As per PEP 484 discussion, it should be possible to use `# type: ignore` to skip type checking in the rest of a file:

```
# type: ignore
... # ignore type errors here
```

Currently the comment can only be used to ignore errors on individual lines.

gvanrossum:
(Moved assignment and Milestone from the duplicate #964.)

JukkaL:
I wonder what we should do if this happens within a function. Should we then ignore all errors within the function only? Example:

```
def f() -> None:
    # type: ignore
    return 1 + 'x'   # don't report an error?

a = 1 + 'y'   # report an error here?
```

This could be nice, as it would let people annotate functions but ignore errors in the body of the function. `@no_type_check` is not a good match, as it also causes the annotations to be ignored.

gvanrossum:
Yeah, I agree that this should be block-scoped.

alexpirine:
Hello,

Alternatively, could there be a way to exclude certain files or directory from command line or in the configuration file?

Currently, if I type `mypy my_directory`, the entire folder is scanned. It would be very handy to be able to add an `--exclude` argument.

In my case, I want to exclude automatically created Django migration files, but still scan my entire Django app.

gvanrossum:
You can use the `ignore_errors` per-file option.

alexpirine:
Thank you @gvanrossum, this seems to be what I was looking for.

Could you please assist me a bit further? Something is still not working:

**mypy.ini**
```
[mypy]

[gw_box.migrations.*]
platformplatform = True
```

```console
$ mypy gw_box --ignore-missing-imports --config-file mypy.ini
gw_box/migrations/0001_initial.py:14: error: Need type annotation for variable
```

gvanrossum:
You need to add a section like this:
```
[mypy-dropbox.edgestore.datatypes.namespace]
ignore_errors = False
```
For docs see http://mypy.readthedocs.io/en/latest/config_file.html

alexpirine:
Thank you, I figured out I was just missing the `mypy-` part.

P.S. Just figured out who you were. Huge thanks you for creating Python :)

JukkaL:
@ddfisher Are you planning to work on this?

chingc:
[...]

PR #3789, opened 2017-08-01
add config option to skip package checking

yedpodtrzitko:
Hi,
this could be partially related to #626.

This PR adds a new option into the per-module section in config file, which would allow to skip typechecks in a given package. 

Example for skipping any packages in `kiwi.schemas`:

```
[mypy-kiwi.schemas.*]
skip = true
```

Thanks for any feedback.

### DIFF
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -187,6 +187,8 @@ overridden by the pattern sections matching the module name.
 - ``no_implicit_optional`` (Boolean, default false) changes the treatment of
   arguments with a default value of None by not implicitly making their type Optional
 
+- ``skip`` (Boolean, default False) skip typechecking in given package.
+
 Example
 *******
 
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1358,6 +1358,10 @@ class State:
             self.import_context = []
         self.id = id or '__main__'
         self.options = manager.options.clone_for_module(self.id)
+
+        if self.options.skip:
+            raise ModuleNotFound
+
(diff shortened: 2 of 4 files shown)


sobolevn:
Real life use-case: I need to exclude everything in `my_django_app/migrations/` module.

ethanhs:
Closing. This is already taken care of by e.g.
```
[mypy-kiwi.schemas.*]
ignore_missing_imports = True
follow_imports = skip
```

yedpodtrzitko:
@ethanhs thanks for your response, however the indended behaviour/purpose of `skip` is different than combination of `follow_imports` & `ignore_missing_imports` in your example.

Having the following directory structure (see below), and running mypy check for the whole `kw` package will go through the `schemas` submodule anyway. It wont report any errors from there whatsoever, but it will parse the content in there. My main reason for (not) doing that is #3796. And the behaviour of `skip` works exactly that way.
 
```
kw
|-bar
|-foo
|-schemas # <- want to skip parsing anything in here
  |-schema_1.py
  |-schema_n.py
```


However I can understand having such option is not a desired design, so until #3796 will be solved, I'm just gonna keep using my fork.

Cheers,
yed

gvanrossum:
1. If `# type: ignore` on a line by itself worked, would that satisfy you?
2. Maybe we could make `follow_imports = skip` work this way?

yedpodtrzitko:
> Maybe we could make `follow_imports = skip` work this way?

I'd be happy with this solution \^\^

Issue #3726, opened 2017-07-17
Stugben: Handling files with file encoding comment


ashwch:
I was trying to generate stubs for [`sre_compile`](https://hg.python.org/cpython/file/2.7/Lib/sre_compile.py) and it fails with:

```
$ python stubgen.py --py2 sre_compile
.../2.7/lib/python2.7/sre_compile.py:0: error: encoding declaration in Unicode string
```

This is probably happening because [`ast27_parse_impl `](https://github.com/python/typed_ast/blob/f4495f45d1d456aae7d01da9f9670fe01e974f17/ast27/Custom/typed_ast.c#L259) converts source from bytes to string and tries to parse it without caring about source code encoding comment.

PEP 263 says:

> If a Unicode string with a coding declaration is passed to `compile()` , a `SyntaxError` will be raised.

One quick fix would be to remove such comment from source before it is passed to [`mypy.parse.parse()` in stubgen.py](https://github.com/python/mypy/blob/master/mypy/stubgen.py#L202).

I am wondering if this is worth fixing? If yes, is the above the best way to do it.

gvanrossum:
IMO stubgen should be made to understand encoding declarations the same way [mypy does](https://github.com/python/mypy/blob/master/mypy/util.py#L48).

PR #3790, opened 2017-08-01
Use read_with_python_encoding in stubgen to handle file encoding

ashwch:
Fixes https://github.com/python/mypy/issues/3726


gvanrossum:
Thanks for the fix!

PR #3793, opened 2017-08-02
Invalidate metadata if errors were previously ignored

### LINKED ISSUE #3639: --incremental --follow-imports=silent run separately on different files does not produce errors
On mypy version 0.511

```
# a.py
import b

b.foo(1, 2)
```

```
# b.py

def foo(a, b):
    # type: (int, int) -> str
    return a + b
```

```
> mypy --incremental --py2 --follow-imports=silent a.py  # No error, as expected.
> mypy --incremental --py2 --follow-imports=silent b.py  # No error where it was expected!
> mypy --incremental --py2 --follow-imports=normal b.py  # Error
```

mhahnenberg:
Using --incremental with --follow-imports=silent triggered a bug where a
file was not typechecked due to the presence of a metadata file
and accompanying AST from a previous run and would therefore not print any
errors. This diff fixes the issue by recording whether errors were
silenced in the metadata file when it was created and ignoring these files if
errors are no longer silenced.

Fixes #3639

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -313,6 +313,7 @@ CacheMeta = NamedTuple('CacheMeta',
                         ('dep_prios', List[int]),
                         ('interface_hash', str),  # hash representing the public interface
                         ('version_id', str),  # mypy version for cache invalidation
+                        ('ignore_all', bool),  # if errors were ignored
                         ])
 # NOTE: dependencies + suppressed == all reachable imports;
 # suppressed contains those reachable imports that were prevented by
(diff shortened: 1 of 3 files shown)


gvanrossum:
Sorry, it's vacation season and there are few people around to review this. We'll get to it.

gvanrossum:
LG, waiting for the tests to pass after the resolve.

gvanrossum:
Thanks! This addresses a nasty obscure corner case.

Issue #3476, opened 2017-05-30
Document how to selectively ignore specific imports


JukkaL:
Currently mypy users often have to ignore all imports (`--ignore-missing-imports`) because mypy can't find stubs for a third party library module. Instead, we could provide an option to only ignore specific imports everywhere within a program. 

Potential way to support this:

```
mypy --ignore-imports=sqlalchemy,frobnicator program.py
```

This would have a few benefits over `--ignore-missing-imports`:
* It would be possible to have fine-grained control over ignored imports (without using a lot of `# type: ignore` comments).
* It would be easy to guard against accidentally adding additional ignored dependencies.
* Users might be more likely to contribute stubs for 3rd party libraries since it would be more obvious where this would be helpful. Once users start using `--ignore-missing-imports`, they lose a lot of visibility to which stubs are missing.
* Mypy would be able to catch misspelled module names in import statements.

OddBloke:
As discussed in #3795, this is actually already supported using a section that matches the import to be ignored; this should either be closed out or updated to be a documentation bug.

ScottWilliamAnderson:
Is there support for a standalone cli command option without a section in one of the configuration files? Or is the only option then the all-or-nothing of --ignore-missing-imports

PR #3795, opened 2017-08-03
Add --ignore-imports option to ignore imports of specific modules (#3476)

OddBloke:
This is essentially a per-module version of --ignore-missing-imports.
It only supports matching the full name that is used to import the
module (so `--ignore-import foo` will not cause `import foo.bar` to be
ignored).

This addresses #3476.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1401,7 +1401,8 @@ class State:
                 # misspelled module name, missing stub, module not in
                 # search path or the module has not been installed.
                 if caller_state:
-                    if not self.options.ignore_missing_imports:
+                    if (not self.options.ignore_missing_imports
+                            and id not in self.options.ignore_imports):
                         save_import_context = manager.errors.import_context()
                         manager.errors.set_import_context(caller_state.import_context)
                         manager.module_not_found(caller_state.xpath, caller_line, id)
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -246,6 +246,9 @@ def process_options(args: List[str],
                         const=defaults.PYTHON2_VERSION, help="use Python 2 mode")
     parser.add_argument('--ignore-missing-imports', action='store_true',
                         help="silently ignore imports of missing modules")
+    parser.add_argument('--ignore-imports', action='append',
+                        help="silently ignore imports of the given module (may"
+                             " be specified more than once)")
     parser.add_argument('--follow-imports', choices=['normal', 'silent', 'skip', 'error'],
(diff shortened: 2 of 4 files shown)


OddBloke:
This doesn't include tests for the new behaviour yet.

OddBloke:
(Now with test.)

ilevkivskyi:
Thanks for the PR! Just a small note for the future: we really prefer not to rebase PRs. It is easier to review PRs that have several smaller commits.

OddBloke:
@ilevkivskyi Understood! :-)

jaens:
This can already be achieved with or without wildcards by using `mypy.ini`, eg.:
```
[mypy-some_module]
ignore_missing_imports = True
```

(this will  only ignore imports of `some_module`)

Is this diff somehow different from that?
Can the `.ini` be represented as a command-line optio somehow so no code would need to be added?

gvanrossum:
@OddBloke Can I close this or do you disagree with @jaens' analysis?

graingert:
@jaens I think that's ignoring imports from some_module not of some_module

gvanrossum:
@OddBloke Please play around with it. I'm pretty sure it does both. Try it! (I don't have the time to try it but I do have time to close this PR. :-)

graingert:
@OddBloke if it does both it's a good opportunity to update the docs!

OddBloke:
Yep, looks like it works fine:

```shell
$ cat << EOF > test.py
heredoc> from no_stubs import Something
heredoc> from wont_ignore import SomethingElse
heredoc> EOF
$ mypy test.py
test.py:1: error: Cannot find module named 'no_stubs'
test.py:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
test.py:2: error: Cannot find module named 'wont_ignore'
$ cat << EOF > mypy.ini
heredoc> [mypy]
heredoc> 
heredoc> [mypy-no_stubs]
heredoc> ignore_missing_imports = True
heredoc> EOF
$ mypy test.py
test.py:2: error: Cannot find module named 'wont_ignore'
test.py:2: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
```

Closing this out, thanks for the catch @jaens!

sid-kap:
Do wildcards work here? In my project, I'm getting import errors in lots of pandas modules, and it seems like the only way to get rid of these errors is to add
```
[mypy-pandas.util.whatever]
ignore_missing_imports = True
```
for each module. (Something like `pandas.*` would be awesome.)

PR #3798, opened 2017-08-04
Report sources of Any in html-report per-line as hovertext

ilinum:
Now, when you hover over a line in html-report, mypy will show types of Anys that appear on that line.


ilinum:
Wohoo! thanks! :)

Issue #3800, opened 2017-08-04
disallow_subclassing_any should be configurable per-module


OddBloke:
I have a codebase that is annotated where I'm introducing an integration with an another project. This integration takes the form of subclassing a super-class from their codebase.

I don't want to allow subclassing of Any for my entire codebase, I really would like to just allow it for this specific module.

(Specifically, when I add the following to `setup.cfg`:

```
[mypy-jenkins_job_linter.jjb_subcommand]
disallow_subclassing_any = False
```

I see this in the output:

```
setup.cfg: [mypy-jenkins_job_linter.jjb_subcommand]: Per-module sections should only specify per-module flags (disallow_subclassing_any)
...
jenkins_job_linter/jjb_subcommand.py:14: error: Class cannot subclass 'TestSubCommand' (has type 'Any')
```
)

gvanrossum:
This seems a simple change to `PER_MODULE_OPTIONS` in mypy/options.py. Can you give that a try?

OddBloke:
Yep, trying that now!

PR #3801, opened 2017-08-04
Make disallow_subclassing_any a per-module option

OddBloke:
Fixing issue #3800.


gvanrossum:
Thanks! Simple and straightforward. Thanks for supplying tests! (Hm, I am just reminded that there may be something in the docs that needs to be changed. Could you check for that yourself?)

OddBloke:
@gvanrossum I added to the per-module section in https://github.com/python/mypy/pull/3801/commits/cb152330d5366c06a4246ab58f3816bd64ee9d01, and I _think_ that's the only place that it was needed...

gvanrossum:
Ah, thanks. I missed that, sorry.

PR #3837, opened 2017-08-16
Attempt to expand Type[C] calls using C's upper bound

OddBloke:
When C in Type[C] is a type variable with an upper bound, use that upper
bound to realise type variables in Type[C] instead of just erasing them.

This allows, for example, `instance` in the following to typecheck as
dict[str, str] instead of dict[Any, Any].

```
C = TypeVar('C', upper_bound=Mapping[str, str])
def func(param: Type[C]) -> C: ...
instance = func(dict)
```

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -156,7 +156,13 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             result = type_object_type(node, self.named_type)
             if isinstance(self.type_context[-1], TypeType):
                 # This is the type in a Type[] expression, so substitute type
-                # variables with Any.
+                # variables. Use the upper bound if applicable, or Any if we
+                # don't have a useful upper bound.
+                target_type = self.type_context[-1].item
+                if (isinstance(target_type, TypeVarType)
+                        and isinstance(target_type.upper_bound, Instance)):
+                    result = expand_type_by_instance(
+                        result, target_type.upper_bound)
                 result = erasetype.erase_typevars(result)
         elif isinstance(node, MypyFile):
             # Reference to a module object.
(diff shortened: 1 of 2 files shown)


OddBloke:
This builds on the change that was merged in #3833, and allows us to retain a little more type information than we did previously.

ilevkivskyi:
TBH this looks a bit arbitrary. I think we should wait with this, so that other can say their opinion.

OddBloke:
Yeah, this is definitely less obvious; I realised that I could do it whilst I was playing around in the code (and it was actually going to be the next issue that I filed once #3833 had landed), so I figured I would propose it to see what people thought.

elazarg:
Does this play well with metaclasses of C?

OddBloke:
@elazarg Could you give me an example of something you're worried about?

gvanrossum:
@elazarg can you answer @OddBloke's question?

elazarg:
No, sorry, I only had a vague feeling that this could be problematic, and couldn't find anything concrete regarding metaclasses.

elazarg:
I have a different concrete example of an issue - I'm not sure how important is this:
```python
from typing import Generic, Type, TypeVar

K = TypeVar('K')
V = TypeVar('V')
class ExampleDict(Generic[K, V]): pass

K1 = TypeVar('K1')
V1 = TypeVar('V1', bound=int)
class SubclassDict(ExampleDict[K1, V1]): pass

D = TypeVar('D', bound=ExampleDict[str, str])
def mkdict(dict_type: Type[D]) -> D:
    return dict_type()
reveal_type(mkdict(SubclassDict))  # E: Revealed type is 'SubclassDict*[builtins.str, builtins.str]'
```
But `SubclassDict[str, str]` is not a legal instantiation since `K1` is bound by an `int`, so it should have been an error (or `Intersection[str, int]`?).

ilevkivskyi:
I think this is in general unsound, class objects do not carry any generic information, only instances. Also I agree with @elazarg that this may be unsafe. Finally, I don't see any particular practical value in this.

PR #3842, opened 2017-08-18
Support six.add_metaclass

### LINKED ISSUE #3365: Support six.add_metaclass
Forked from #1764. I propose to give this less priority than `six.with_metaclass`, since the latter is more fundamental in some sense (it works for multiple bases with conflicting metaclasses where `add_metaclass` can't, see https://github.com/python/mypy/issues/1764#issuecomment-301077905).

elazarg:
Fix #3365

There are no special checks for improper use (similar to `with_metaclass`).
Consistency checks are handled like in any other metaclass.

### DIFF
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -3470,65 +3470,61 @@ reveal_type(y['b']) # E: Revealed type is '__main__.B'
 -- Special support for six
 -- -----------------------
 
-[case testSixWithMetaclass]
+[case testSixMetaclass]
 import six
 class M(type):
     x = 5
 class A(six.with_metaclass(M)): pass
+@six.add_metaclass(M)
+class B: pass
 reveal_type(type(A).x)  # E: Revealed type is 'builtins.int'
+reveal_type(type(B).x)  # E: Revealed type is 'builtins.int'
 
-[case testSixWithMetaclass_python2]
+[case testSixMetaclass_python2]
 import six
 class M(type):
     x = 5
 class A(six.with_metaclass(M)): pass
-reveal_type(type(A).x)  # E: Revealed type is 'builtins.int'
-
-[case testSixAddMetaclass]
-import six
-class M(type):
-    x = 5
-class X:
-    a = 'h'
 @six.add_metaclass(M)
-class A(X): pass
+class B: pass
 reveal_type(type(A).x)  # E: Revealed type is 'builtins.int'
-reveal_type(A().a)  # E: Revealed type is 'builtins.str'
+reveal_type(type(B).x)  # E: Revealed type is 'builtins.int'
 
-[case testSixAddMetaclass_python2]
-import six
-class M(type):
-    x = 5
-@six.add_metaclass(M)
-class A(object): pass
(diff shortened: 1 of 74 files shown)


test-data/unit/check-classes.test
```diff
@@ -3484,6 +3484,25 @@ class M(type):
...
+reveal_type(A().a)  # E: Revealed type is 'builtins.str'
+
+[case testSixAddMetaclass_python2]
+import six
+class M(type):
+    x = 5
+@six.add_metaclass(M)
+class A(object): pass
+reveal_type(type(A).x)  # E: Revealed type is 'builtins.int'
+
```
ilevkivskyi: I would add more tests. Some ideas:
* An invalid metaclass (not a subtype of ``type``)
* Something involving ``Any`` (for example a metaclass form silent import)
* ``__getattr__`` and/or other special attributes like ``__iter__``
* Multiple decorators on a class
* An incompatible metaclass

elazarg: But these will test other (later) mechanisms, so aren't they redundant (white-box-wise)?

ilevkivskyi: Look at the number of tests for ``with_metaclass``, they are there for a reason, tests are not for the present (implementation), they are for the future (when everything may be refactored in unpredictable ways).

ilevkivskyi:
OK, this one can be now updated after #3848 is merged (plus the suggested tests).

test-data/unit/check-classes.test
```diff
@@ -3539,25 +3569,50 @@ class ArcMeta(GenericMeta, DestroyableMeta):
     pass
 class Arc(six.with_metaclass(ArcMeta, Generic[T_co], Destroyable)):
     pass
+@six.add_metaclass(ArcMeta)
+class Arc1(Generic[T_co], Destroyable):
```
ilevkivskyi: This actually fails at runtime because of a metaclass conflict. Is it easy to detect this? If this is non-trivial, then I think we should not include it in this PR.

test-data/unit/check-classes.test
```diff
@@ -3539,25 +3569,50 @@ class ArcMeta(GenericMeta, DestroyableMeta):
...
-
-[case testSixWithMetaclassErrors_python2-skip]
-# No error here yet
+@six.add_metaclass(f())  # E: Dynamic metaclass not supported for 'D5'
+class D5: pass
+
+@six.add_metaclass(M)  # E: Multiple metaclass definitions
+class CD(six.with_metaclass(M)): pass
+
+def q(t):
```
ilevkivskyi: This is not a good way to test ``Any``, since the function is not checked at all. I would prefer something like import a metaclass and/or base class from a non-existing module. This will generate an error and the imported names will have types ``Any``. (Or maybe use ``--follow-imports=silent`` to silence the error.)

test-data/unit/check-classes.test
elazarg: I believe this is not caught because the base class Generic is removed earlier in the analysis. It doesn't seem obvious to me how this should be fixed.

ilevkivskyi: OK, them it is not easy to support this.

PR #3844, opened 2017-08-19
Switch typeshed submodule URL to `https`

gnprice:
Not only is this more hip to the future of the Web -- but right now
GitHub is failing to respond most of the time on port 80 for me, while
things work fine for the `https` URL.


gvanrossum:
Good idea!

PR #3846, opened 2017-08-19
Update pythoneval test according to PR typeshed/#1220 (ABCs -> Protocols)

ilevkivskyi:
The test will fail now. This PR contains the tests output after https://github.com/python/typeshed/pull/1220 is merged and synced (note that pythoneval tests are run with full typeshed stubs).

### DIFF
--- a/test-data/unit/pythoneval.test
+++ b/test-data/unit/pythoneval.test
@@ -1402,6 +1402,11 @@ o: object = p
 it2: Iterable[int] = p
 [out]
 _testCanConvertTypedDictToAnySuperclassOfMapping.py:11: error: Incompatible types in assignment (expression has type "Point", variable has type "Iterable[int]")
+_testCanConvertTypedDictToAnySuperclassOfMapping.py:11: note: Following member(s) of "Point" have conflicts:
+_testCanConvertTypedDictToAnySuperclassOfMapping.py:11: note:     Expected:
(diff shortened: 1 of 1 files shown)


ilevkivskyi:
The failure is expected, but AppVeyor has crashed in a strange way, this is a second time it happened today on random pushes. @ethanhs do you have any ideas what has happened.

ethanhs:
@ilevkivskyi Im taking a look. The error seem non-obvious, but I will try to see if I can figure out what it is.

elazarg:
I have encountered similar crash during #3870. IIRC it seemed to happen when multiple instances of pytest were active. It stopped happening after I've changed the PR to use only a single instance. I'm not sure if this information helps though...

ethanhs:
Moving to a new issue to not clutter this PR #3895

gvanrossum:
I propose to disable this test (by appending `-skip` to the test name), then merging the typeshed PR, then re-enabling the failing test. That way there's no race between the two PRs. (I find it more stressful to have failing tests than to have a TODO about re-enabling a test.)

ilevkivskyi:
@gvanrossum 
OK, I temporarily disabled the test. Now this PR re-enables it with the updated output.

gvanrossum:
OK, can you rebase/repush this so the tests pass? Then you can merge it.

ilevkivskyi:
All tests pass, so I am merging now.

PR #3848, opened 2017-08-19
Make ClassDef.metaclass an Expression

### LINKED ISSUE #3847: Make ClassDef.metaclass an Expression instead of str
Currently the metaclass node is assumed to be a string. This is inconsistent, requires workarounds such as passing `'<error>'` from the parser to the semantic analysis phase, and makes it awkward to handle metaclasses that come from different contexts (as in python2).

Making it an Expression works cleanly as far as I can tell, and reduces code duplication.

elazarg:
Fix #3847.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -465,12 +465,6 @@ class ASTConverter(ast3.NodeTransformer):
     @with_line
     def visit_ClassDef(self, n: ast3.ClassDef) -> ClassDef:
         self.class_nesting += 1
-        metaclass_arg = find(lambda x: x.arg == 'metaclass', n.keywords)
-        metaclass = None
-        if metaclass_arg:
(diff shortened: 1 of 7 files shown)


ilevkivskyi:
You have another PR https://github.com/python/mypy/pull/3842 that will have conflicts with this one, which one should be considered first?

elazarg:
This one. #3847 should be easier and cleaner afterwards. Should I close it for now?

ilevkivskyi:
> Should I close it for now?

No, it can be left open, but please don't rebase it afterwards, merge master instead.

ilevkivskyi:
I will take a look at this later, sorry, don't have time now.

gvanrossum:
No time to review but the stated goal seems right and the diff looks simple enough, so no objection to Ivan merging this.

PR #3854, opened 2017-08-21
Even more strict optional

ilevkivskyi:
This PR cleans another hundred of ``--strict-optional`` errors and makes four more  files completely clean:
* main.py
* build.py
* fastparse.py
* fastparse2.py

------------------------------
The next PR will be probably the last one, also it will be pretty "hardcore", since there are only three files left - checker/checkexpr/semanal - but these files are touched by almost every PR, so there will be lots of conflicts.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1545,8 +1545,7 @@ class State:
         self.manager.modules[self.id] = self.tree
 
     def fix_cross_refs(self) -> None:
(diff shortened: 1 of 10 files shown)


ilevkivskyi:
3.5.1 strikes again... will implement an uglier solution now.

gvanrossum:
Thanks! Reviewed only lightly, I trust you in this stuff.

PR #3861, opened 2017-08-22
Move testtypegen to pytest

elazarg:
Another item from #1673, following #3780 and #3788

As in previous PRs, I have checked that a test may fail by changing `testConstructorCall` and running

    ./runtests.py pytest -a -n0 -a -k -a testGlobalVariable


gvanrossum:
Thanks again!

PR #3862, opened 2017-08-22
Better error on empty typeshed directory

ethanhs:
If typeshed is empty (someone forgot to pull the submodule, or the install is broken) the path to typeshed will resolve correctly, but of course stubs will not be found. I added an assert to catch this case. In addition, I realized that before when analyzing 2.7 code, it would look for subdirectories of the typeshed directory for 2.0-2.6, which is pointless, so I modified it to only look in the 2.7 subdirectory.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -44,6 +44,7 @@ from mypy.stats import dump_type_stats
 from mypy.types import Type
 from mypy.version import __version__
 from mypy.plugin import Plugin, DefaultPlugin, ChainedPlugin
+from mypy.defaults import PYTHON3_VERSION_MIN
 
 
 # We need to know the location of this file to load data, but
@@ -283,11 +284,11 @@ def default_lib_path(data_dir: str,
         # We allow a module for e.g. version 3.5 to be in 3.4/. The assumption
         # is that a module added with 3.4 will still be present in Python 3.5.
         versions = ["%d.%d" % (pyversion[0], minor)
-                    for minor in reversed(range(3, pyversion[1] + 1))]
+                    for minor in reversed(range(PYTHON3_VERSION_MIN[1], pyversion[1] + 1))]
     else:
         # For Python 2, we only have stubs for 2.7
         versions = ["2.7"]
-    # E.g. for Python 3.2, try 3.2/, 3.1/, 3.0/, 3/, 2and3/.
+    # E.g. for Python 3.5, try 3.5/, 3.4/, 3.3/, 3/, 2and3/.
     # (Note that 3.1 and 3.0 aren't really supported, but we don't care.)
     for v in versions + [str(pyversion[0]), '2and3']:
         for lib_type in ['stdlib', 'third_party']:
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -279,10 +279,14 @@ def default_lib_path(data_dir: str,
         if os.path.isdir(auto):
             data_dir = auto
         typeshed_dir = os.path.join(data_dir, "typeshed")
-    # We allow a module for e.g. version 3.5 to be in 3.4/. The assumption
-    # is that a module added with 3.4 will still be present in Python 3.5.
-    versions = ["%d.%d" % (pyversion[0], minor)
-                for minor in reversed(range(pyversion[1] + 1))]
+    if pyversion[0] == 3:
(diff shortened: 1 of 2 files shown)


mypy/build.py
```diff
@@ -294,7 +298,9 @@ def default_lib_path(data_dir: str,
     # Add fallback path that can be used if we have a broken installation.
     if sys.platform != 'win32':
         path.append('/usr/local/lib/mypy')
-
+    assert path, "Could not resolve typeshed subdirectories. If you are using MyPy " \
```
gvanrossum: Is there a way to turn this into a regular (blocking) error message? Many people don't read the error when they get a traceback, and just instantly file a bug or go to a help forum.

mypy/build.py
```diff
@@ -279,10 +279,14 @@ def default_lib_path(data_dir: str,
...
         typeshed_dir = os.path.join(data_dir, "typeshed")
-    # We allow a module for e.g. version 3.5 to be in 3.4/. The assumption
-    # is that a module added with 3.4 will still be present in Python 3.5.
-    versions = ["%d.%d" % (pyversion[0], minor)
-                for minor in reversed(range(pyversion[1] + 1))]
+    if pyversion[0] == 3:
+        # We allow a module for e.g. version 3.5 to be in 3.4/. The assumption
+        # is that a module added with 3.4 will still be present in Python 3.5.
+        versions = ["%d.%d" % (pyversion[0], minor)
+                    for minor in reversed(range(pyversion[1] + 1))]
```
gvanrossum: That could probably be `range(2, pyversion[1] + 1)` since we only support 3.2 and up.

ethanhs: Typeshed only has 3.3+ so I will do `range(3, pyversion[1] + 1)`.

mypy/build.py
gvanrossum: I'd like to plead for starting at 2, since mypy still supports 3.2 (as the check target -- not to run mypy itself). It just avoids confusion for people who read the source. (Also it should probably be a DEFINED_CONSTANT somewhere.)

ethanhs: Im not sure we support 3.2... only 3.3+ are listed in `setup.py`, and the README states that we support 3.3+. We are planning on dropping 3.3 when it is EOL'd in a month I recall. And I agree it should be a constant.

As to the blocking error message, Im not sure. In general if we are in a state where things are "wrong" eg if there is a key error looking up builtins, that is an exception. I can see how the UX may overrule this. Perhaps raise using report_internal_error in mypy.errors?

gvanrossum: Heh, I forgot about it, but indeed we dropped support for 3.2. (And may it R.I.P. :-)

Regarding the error message, the Error instance hasn't even been constructed at the point this function is called, so the best we could do would be to `print(message, file=sys.stderr)`and then `sys.exit(1)`. Which isn't terrible, main.py even has a helper that does just that. (Though I think eventually we should refactor the fatal/blocking error handling; I've also got the idea that using sys.exit(2) for all fatal errors would be preferable.)

ethanhs: >we dropped support for 3.2. (And may it R.I.P. :-)

Indeed

RE errors, printing then exiting makes sense. I'll make those changes now.

ethanhs:
I addressed the two pieces of feedback so this is ready for review again @gvanrossum

mypy/build.py
```diff
@@ -283,7 +283,7 @@ def default_lib_path(data_dir: str,
         # We allow a module for e.g. version 3.5 to be in 3.4/. The assumption
         # is that a module added with 3.4 will still be present in Python 3.5.
         versions = ["%d.%d" % (pyversion[0], minor)
-                    for minor in reversed(range(pyversion[1] + 1))]
+                    for minor in reversed(range(3, pyversion[1] + 1))]
```
gvanrossum: It was my understanding that you would replace this magical constant '3' with some globally defined named constant.

ethanhs: Oh, sorry I must have misunderstood, I thought you wanted that as a separate change. I'll make that change now as part of this.

gvanrossum:
Thanks!

ethanhs:
My pleasure! I'll look through things and see if there is anywhere else that could use the minimum Python version global.

PR #3865, opened 2017-08-23
Add --warn-unused-configs flag

gvanrossum:
This is handy for debugging misspellings in mypy.ini, and other mishaps due to e.g. file moves.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -53,6 +53,11 @@ def main(script_path: Optional[str], args: Optional[List[str]] = None) -> None:
         a = e.messages
         if not e.use_stdout:
             serious = True
+    if options.warn_unused_configs and options.unused_configs:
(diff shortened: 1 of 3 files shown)


mypy/main.py
```diff
@@ -53,6 +53,11 @@ def main(script_path: Optional[str], args: Optional[List[str]] = None) -> None:
         a = e.messages
         if not e.use_stdout:
             serious = True
+    if options.warn_unused_configs and options.unused_configs:
+        print("warning: unused section(s) in %s: %s" %
```
gvanrossum: Note that the warning goes to stderr but doesn't get routed via the Errors object so it doesn't get included in the junit.xml file (if we write it). Then again it's not associated with any particular file, and I'm not sure if the Error  class supports such errors. (All other warnings related to the config file also just go to stderr directly.)

PR #3866, opened 2017-08-23
Move testsemanal to pytest

elazarg:
Another item from #1673, following #3780, #3788 and #3861.

As in previous PRs, I have checked that a test may fail by changing `testMissingModuleImport1` and running

    ./runtests.py pytest -a -n0 -a -k -a testMissingModuleImport1

If I am not mistaken this is the last "simple" file. After this there's the pyeval/cmd tests, refactoring/cleanup, and the non-datadriven tests which require special treatment.


gvanrossum:
Thanks again!

PR #3868, opened 2017-08-24
Use global minimum Python version for version check

ethanhs:
As a follow up to [this comment](https://github.com/python/mypy/pull/3862#discussion_r134635535) I thought it would be good to use this constant everywhere. Now to update the minimum supported Python version we just need to update the default in `defaults.py` and the trove classifiers in `setup.py`.

EDIT: It seems I accidentally based this off of my previous branch. If you want I can rebase to be based off of master.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -155,7 +155,8 @@ def parse_version(v: str) -> Tuple[int, int]:
     elif major == 3:
         if minor <= 2:
             raise argparse.ArgumentTypeError(
-                "Python 3.{} is not supported (must be 3.3 or higher)".format(minor))
+                "Python 3.{0} is not supported (must be {1}.{2} or higher)".format(minor,
+                                                                    *defaults.PYTHON3_VERSION_MIN))
     else:
         raise argparse.ArgumentTypeError(
             "Python major version '{}' out of range (must be 2 or 3)".format(major))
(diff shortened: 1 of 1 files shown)


gvanrossum:
Yeah, please merge from master.

ethanhs:
Okay that should be better, my apologies.

mypy/main.py
```diff
@@ -155,7 +155,8 @@ def parse_version(v: str) -> Tuple[int, int]:
     elif major == 3:
         if minor <= 2:
```
gvanrossum: I'm guessing that should be `defaults.PYTHON3_VERSION_MIN[1]`?

ethanhs: Yes. Good point.

mypy/main.py
```diff
@@ -153,7 +153,7 @@ def parse_version(v: str) -> Tuple[int, int]:
             raise argparse.ArgumentTypeError(
                 "Python 2.{} is not supported (must be 2.7)".format(minor))
     elif major == 3:
-        if minor <= 2:
+        if minor <= defaults.PYTHON3_VERSION_MIN[1]:
```
gvanrossum: Should be '<'.

ethanhs: 🤦‍♂️ sorry I'll fix that.

ethanhs: Fixed.

PR #2635, opened 2017-01-04
Speed up testpythoneval

ambv:
Splits eval-test into simple buckets by first letter of test name, enabling
parallel execution. This speeds up execution of the test suite by around 25% on
my laptop. The split enables more consistent loading of all CPU cores during
the entire run of ./runtests.py.

To achieve this, I had to modify testpythoneval.py to not write all testcase
inputs to the same temporary path.

Before:
```
  SUMMARY  all 204 tasks and 1811 tests passed
  *** OK ***
  total time in run: 554.571954
  total time in check: 214.105742
  total time in lint: 130.914682
  total time in pytest: 92.031659
  ./runtests.py -j4 -v  744.76s user 74.10s system 235% cpu 5:48.34 total
```
After:
```
  SUMMARY  all 225 tasks and 3823 tests passed
  *** OK ***
  total time in run: 640.698327
  total time in check: 178.758370
  total time in lint: 149.604402
  total time in pytest: 78.356671
  ./runtests.py -j4 -v  850.81s user 81.09s system 353% cpu 4:23.69 total
```
Total wall clock time fell from 5:48 to 4:23.

Note: the test sum is now over-reported. Looks like the driver counts also the
filtered out tests in eval-test. I don't have cycles now to hunt this down.

PR #3870, opened 2017-08-24
Move pythoneval and cmdline to pytest

elazarg:
(EDITED)
Another item from #1673, following #3780, #3788, #3861 and #3866.

After this, `StubgenPythonSuite` is the only data-driven test left in myunit.

As in previous PRs, I have checked that a test may fail by changing tests and running

    ./runtests.py eval -a -n0 -a -k -a testHello
    ./runtests.py eval -a -n0 -a -k -a testSimpleCoroutineSleep
    ./runtests.py cmd -a -n0 -a -k -a testCmdlinePackageContainingSubdir

I am somewhat less confident about this PR, since I am not entirely sure what special considerations are in play here (which aren't there for the proper unit tests).

@ambv @gvanrossum: The splitting of eval tests, introduced in PR #2635, is removed. IIUC this was the intention - please correct me if I'm wrong. (I did not time the difference).

Executing multiple pytest instances does not work well with appveyor.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -5,13 +5,10 @@ from typing import Dict, List, Optional, Set, Iterable
 
 from mypy.waiter import Waiter, LazySubprocess
 from mypy import util
-from mypy.test.config import test_data_prefix
-from mypy.test.testpythoneval import python_eval_files, python_34_eval_files
 
 import itertools
 import os
 from os.path import join, isdir
-import re
 import sys
 
 
@@ -92,7 +89,8 @@ class Driver:
     def add_mypy_string(self, name: str, *args: str, cwd: Optional[str] = None) -> None:
         self.add_mypy_cmd(name, ['-c'] + list(args), cwd=cwd)
 
-    def add_pytest(self, name: str, pytest_args: List[str], coverage: bool = False) -> None:
+    def add_pytest(self, name: str, pytest_files: List[str], coverage: bool = True) -> None:
+        pytest_args = pytest_files + self.arglist + self.pyt_arglist
         full_name = 'pytest %s' % name
         if not self.allow(full_name):
             return
@@ -197,7 +195,12 @@ def add_imports(driver: Driver) -> None:
             driver.add_python_string('import %s' % mod, 'import %s' % mod)
 
 
-PYTEST_FILES = [os.path.join('mypy', 'test', '{}.py'.format(name)) for name in [
+def test_path(*names: str):
+    return [os.path.join('mypy', 'test', '{}.py'.format(name))
+            for name in names]
+
+
+PYTEST_FILES = test_path(
     'testcheck',
     'testextensions',
     'testdeps',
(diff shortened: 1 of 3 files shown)


elazarg:
I will try to decipher appveyor's failure. Any help would be appreciated.

elazarg:
Looks like executing multiple pytest instances does not work well with appveyor.

runtests.py
```diff
@@ -208,57 +211,36 @@ def add_imports(driver: Driver) -> None:
...
-            coverage=True
-        )
-
-
-def add_cmdline(driver: Driver) -> None:
-    driver.add_python_mod('cmdline-test', 'mypy.myunit',
-                          '-m', 'mypy.test.testcmdline', *driver.arglist,
-                         coverage=True)
+        driver.add_python_mod('unit-test %s' % mod, 'mypy.myunit', '-m', mod,
+                              *driver.arglist, coverage=True)
```
elazarg: The diff is somewhat confusing - this part belongs to `add_myunit()`

elazarg:
I can start this PR from scratch, if it will be easier to review.

gvanrossum:
Sadly there's a problem with this. When I run the tests as follows:
```
pytest -n0 mypy/test/testcmdline.py
```
every test in that file fails. Apparently the `pytest-xdist` extension which runs tests in parallel does something that keeps the tests from failing. The error is essentially that every subprocess invoking scripts/mypy fails with the following traceback:
```
  File "/Users/guido/src/mypy/scripts/mypy", line 4, in <module>
    from mypy.main import main
ModuleNotFoundError: No module named 'mypy.main'
```
This makes sense as the current directory is set to the temporary directory where the test files are written, so the repo root is not on `sys.path`. I'm guessing that the xdist extension somehow retains the current directory?

The pythoneval tests fail in the same way.

gvanrossum:
(And yes, I spent 2 hours trying to debug unrelated code because I didn't realize `-n0` always made these tests fail.)

elazarg:
Ouch. Sorry.

I recall checking for that before opening this PR, and it did not fail on my machine :\ I use `-n0` pretty often. But maybe I didn't use it for `cmdline` specifically; for some reason, exactly one test (`testCoberturaParser`) fails for me now: 
```
Expected:
  <coverage timestamp="$TIMESTAMP" version="$VERSION" line-rate="0.8000" branch-rate="0"> (diff)
  ...
Actual:
  <coverage timestamp="$TIMESTAMP" version="0.530" line-rate="0.8000" branch-rate="0"> (diff)
  ...
```

elazarg:
That test indeed does not fail without `-n0`.

gvanrossum:
I wonder if you have mypy installed? That version looks like it comes from the installed version. If you `pip3 uninstall` it, do more tests fail?

elazarg:
Yes they are! Or actually, they didn't fail - until I've commented out the patch :)

elazarg:
But actually this means my patch is the wrong solution. We need to *make sure* the folder is the right one, and simply adding it to PYTHONPATH does not guarantee it, I assume.

gvanrossum:
Use mypy.api.run()? (See other issue.)

PR #3893, opened 2017-08-30
Format many-itemed tuples and unions consistently and more clearly

OddBloke:
When the string representation of a tuple or union is too long to display, a shorter representation is used.  This moves from two different representations to `<tuple: N items>` and `<union: N items>`, and consistently does not quote them.

This also includes a refactoring of the quoting logic in to a separate method, which enables the quoting of star arguments consistently with other types.


OddBloke:
This comes out of the conversation @ilevkivskyi and I had in #3873 (the relevant comments start here: https://github.com/python/mypy/pull/3873#issuecomment-325976972).

Issue #3890, opened 2017-08-30
`runtests.py pytest` doesn't work


JukkaL:
`./runtests.py -l` indicates that there is a task called `pytest` but `runtests.py pytest` doesn't run it:

```
$ ./runtests.py pytest
PARALLEL 8
SUMMARY  0 tasks selected
SUMMARY  all 0 tasks and 0 tests passed
*** OK ***
total runtime: 0.010855553999135736 sec
```

elazarg:
Another regression of mine. What is the ideal interface? I'd suggest
`./runtests.py unit-test` runs all unit test, myunit and pytest alike.
Do we need a way to run myunit tests (or pytest) specifically?

JukkaL:
If we name the tasks `pytest unit-test` and `legacy unit-test` then `./runtests.py unit-test` will run  both of them, because runtests uses a substring check. We can also have `pytest integration-test` for the slower tests.

JukkaL:
Fixed by #3894.

Issue #3891, opened 2017-08-30
Instructions for running unit tests obsolete in the README


JukkaL:
The README has these instructions which no longer work as expected:

```
For example, to run unit tests only, which run pretty quickly:

    $ ./runtests.py unit-test pytest
```

It currently runs no pytest tests (but this is just broken, which is a separate issue: #3890). Even if it would run all pytest tests, it would now also likely run integration tests, which are slow. Being able to run only unit tests is an important feature.

JukkaL:
cc @elazarg

JukkaL:
Fixed by #3894.

PR #3894, opened 2017-08-30
Fix unit-test and pytest lookup

elazarg:
Quick fix for <s>both #3890 and</s> #3891. I'm not sure it does the right thing, but it's will be easier to discuss it here. I'm also not sure about terminology.

`./runtests.py unit-test` runs all unit tests (both myunit and pytest) except pythoneval and cmdline
`./runtests.py integration` runs pythoneval and cmdline

Also possible `./runtests.py myunit` and `./runtests.py pytest`

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 """Mypy test runner."""
 
-from typing import Dict, List, Optional, Set, Iterable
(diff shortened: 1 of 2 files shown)


elazarg:
Actually this doesn't fix #3890; I'm not sure how it works yet.

JukkaL:
Thanks for fixing this! LGTM, though the pytest test selection logic is a bit convoluted -- but we can live with that.

PR #3900, opened 2017-08-31
Update credits to current state of affairs

gvanrossum:


gvanrossum:
Alternatively, we should just list the core team and ex-core-team members and leave the rest up to GitHub history?

CREDITS
```diff
@@ -1,15 +1,10 @@
...
-
 Core team:
 
-  Guido <guido@dropbox.com>
-  David Fisher <ddfisher@dropbox.com>
-  Greg Price <gregprice@dropbox.com>
+  Jukka Lehtosalo <jukka.lehtosalo@iki.fi>
+  Guido van Rossum <guido@python.org>
 
 Contributors (in alphabetical order, including typeshed):
```
JukkaL: At least mention that this is a partial list and suggest looking at the git history for a full set of contributors.

CREDITS
```diff
@@ -59,6 +57,7 @@ Contributors (in alphabetical order, including typeshed):
   Daniel Shaulov
   David Shea
   Vita Smid
+  Ethan Smith
```
JukkaL: Also add Naomi?

gvanrossum:
The more I think about it the stronger I feel we should not have this list at all. (TBH I didn't even know it existed.)

JukkaL:
Ok, let's drop this and just mention current and previous core team members. We could also add a links to the release blog where we have more detailed per-release contributor information and to the github contributions page (https://github.com/python/mypy/graphs/contributors).

gvanrossum:
How about this?

JukkaL:
These are now live on the website.

PR #3902, opened 2017-08-31
Adds --output-dir option to stubgen

brianbruggeman:
Currently, the output folder is hard-coded.

* Adds an option (--output-dir) to the stubgen command-line interface
  -- allows user to provide a different output folder other than 'out'
  -- defaults to 'out'
  -- allows ~user and relative '.' paths

* Now creates a new folder if it doesn't already exist

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -74,6 +74,7 @@ Options = NamedTuple('Options', [('pyversion', Tuple[int, int]),
                                  ('ignore_errors', bool),
                                  ('recursive', bool),
                                  ('include_private', bool),
+                                 ('output_dir', str),
                                  ])
 
 
@@ -623,8 +624,8 @@ def walk_packages(packages: List[str]) -> Iterator[str]:
 
 def main() -> None:
     options = parse_options(sys.argv[1:])
-    if not os.path.isdir('out'):
-        raise SystemExit('Directory "out" does not exist')
+    if not os.path.isdir(options.output_dir):
+        raise SystemExit('Directory "{}" does not exist'.format(options.output_dir))
     if options.recursive and options.no_import:
         raise SystemExit('recursive stub generation without importing is not currently supported')
     sigs = {}  # type: Any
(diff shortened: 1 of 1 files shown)


gvanrossum:
The default should really remain 'out'.

brianbruggeman:
Is there a reason/backstory on why `out` is currently hard-coded?

gvanrossum:
> why out is currently hard-coded?

This makes it clear what the output is from the script and makes it easy to throw it away if you prefer to start over.

mypy/stubgen.py
```diff
@@ -666,8 +670,19 @@ def parse_options(args: List[str]) -> Options:
...
     include_private = False
+    output_dir = 'out'
     while args and args[0].startswith('-'):
-        if args[0] == '--doc-dir':
+        if args[0] == '--output-dir':
+            output_dir = args[1]
+            args = args[1:]
+            # Resolve/expand output_dir; use symlinks
+            if output_dir.startswith('~'):
+                output_dir = os.path.expanduser(output_dir)
```
gvanrossum: Normally the shell takes care of this. I think on Windows it may not -- but is it really important enough to support?

mypy/stubgen.py
```diff
@@ -666,8 +670,19 @@ def parse_options(args: List[str]) -> Options:
...
+            elif output_dir == '.':
+                output_dir = os.getenv('PWD') or os.getcwd()
```
gvanrossum: What's the purpose of this? I don't believe stubgen ever changes the current directory.

mypy/stubgen.py
```diff
@@ -666,8 +670,19 @@ def parse_options(args: List[str]) -> Options:
...
+            else:
+                output_dir = os.path.abspath(output_dir)
```
gvanrossum: Ditto.

mypy/stubgen.py
```diff
@@ -697,6 +712,9 @@ def parse_options(args: List[str]) -> Options:
         usage()
     if not interpreter:
         interpreter = sys.executable if pyversion[0] == 3 else default_python2_interpreter()
+    # Build the output folder if it doesn't already exist.
```
gvanrossum: Build -> Create

mypy/stubgen.py
brianbruggeman: These are completely unnecessary.

gvanrossum:
Thanks!

PR #3300, opened 2017-05-02
Fix overload handling with union types in signatures

JukkaL:
Previously `None` was not considered as more precise than
`Optional[x]`, which was obviously incorrect. Fixed the
implementation of type precision checking to match the
description, now that proper subtype checking works.

Fixes the original example in #3295.

PR #3917, opened 2017-09-05
Infer better type for call to overload with Any actual arguments

### LINKED ISSUE #3662: Inconsistent return type when calling overloaded function with Any argument
The results of type checking the following fragment are unexpected:

```py
from typing import Any, overload, Union

@overload
def f(x: int) -> int: ...
@overload
def f(x: Union[int, float]) -> float: ...
def f(x): pass

@overload
def g(x: int) -> int: ...
@overload
def g(x: float) -> float: ...
def g(x): pass

a: Any
reveal_type(f(a)) # Any   <--- okay
reveal_type(g(a)) # int   <--- unexpected
```

The correct result would arguably be `Any` for both of the calls. `Union[int, float]` should be equivalent to just `float`.

This seems to interact with the result type of `os.path.join` in Python 2 mode with #3300.

JukkaL:
If the overload return type is ambiguous and there is an `Any` actual
argument that seems to cause the ambiguity, infer `Any` as the type of
the call expression instead of some precise type. Previously we
sometimes inferred a precise type even though the type was ambiguous,
resulting in false positives.

This is not quite the most correct implementation since we sometimes
incorrectly assume ambiguity caused by `Any` even though none actually
exists. This is acceptable since this only affects calls with `Any`
arguments that are imprecise anyway. Also, the correct logic would be
quite complex, hard to test and probably not worth it, at least right
now.

Fixes #3662.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1,7 +1,7 @@
 """Expression type checker. This file is conceptually part of TypeChecker."""
 
 from collections import OrderedDict
-from typing import cast, Dict, Set, List, Tuple, Callable, Union, Optional
+from typing import cast, Dict, Set, List, Tuple, Callable, Union, Optional, Iterable
 
 from mypy.errors import report_internal_error
 from mypy.typeanal import has_any_from_unimported_type, check_for_explicit_any, set_any_tvars
@@ -1076,8 +1076,10 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             if similarity > 0 and similarity >= best_match:
                 if (match and not is_same_type(match[-1].ret_type,
                                                typ.ret_type) and
-                    not mypy.checker.is_more_precise_signature(
-                        match[-1], typ)):
+                    (not mypy.checker.is_more_precise_signature(match[-1], typ)
+                     or (any(isinstance(arg, AnyType) for arg in arg_types)
+                         and any_arg_causes_overload_ambiguity(
+                             match + [typ], arg_types, arg_kinds, arg_names)))):
                     # Ambiguous return type. Either the function overload is
                     # overlapping (which we don't handle very well here) or the
                     # caller has provided some Any argument types; in either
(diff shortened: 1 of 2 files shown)


ilevkivskyi:
Somehow it seems to me that inferring the widest return type (if there is one) instead of ``Any`` might be better, although I see that this still can cause spurious errors if this return type will appear in a contravariant position. So all in all I think this is fine for now, since this problem blocks  #3300 which is important.

mypy/checkexpr.py
```diff
@@ -2764,3 +2766,71 @@ def overload_arg_similarity(actual: Type, formal: Type) -> int:
...
+        arg_kinds: Actual argument kinds
+        arg_names: Actual argument names
+    """
+    actual_to_formal = [
+        map_formals_to_actuals(
+            arg_kinds, arg_names, item.arg_kinds, item.arg_names, lambda i: arg_types[i])
+        for item in items
+    ]
+
+    for i, arg_type in enumerate(arg_types):
```
ilevkivskyi: It is quite hard to follow the logic in this place, maybe you could use something like ``act_pos, form_pos`` instead of ``i, j``?

test-data/unit/check-overloading.test
```diff
@@ -1157,3 +1157,116 @@ class Test(object):
...
+def f(x): pass
+
+a: Any
+# Any causes ambiguity
+reveal_type(f(a, 1, ''))  # E: Revealed type is 'Any'
+# Any causes no ambiguity
+reveal_type(f(1, a, a))  # E: Revealed type is 'builtins.int'
+reveal_type(f('', a, a))  # E: Revealed type is 'builtins.object'
+# Like above, but use keyword arguments.
+reveal_type(f(y=1, z='', x=a))  # E: Revealed type is 'Any'
```
ilevkivskyi: I would add one more test where the order of arguments is shuffled, just to be sure that the matching logic works right.

mypy/checkexpr.py
JukkaL: Done

test-data/unit/check-overloading.test
JukkaL: Added another call where arguments are shuffled

JukkaL:
The widest type may also generate false positives, since the wider type may be missing some attributes defined in the narrower type. Also, if a user creates an invariant container such as a list based on the result, it won't be compatible if the expected type is the narrower one.

ilevkivskyi:
> Also, if a user creates an invariant container such as a list based on the result, it won't be compatible if the expected type is the narrower one.

Yes, this is more important that contravariant, since invariant containers are abundant.

ilevkivskyi:
OK, Now we can go with #3300

Issue #3208, opened 2017-04-21
Reject covariant overriding of attributes


JukkaL:
Covariant overriding of attributes is allowed though it's clearly unsafe. Example:
```py
class A:
    x: float

class B(A):
    x: int  # no error, though unsafe

def f(a: A) -> None:
    a.x = 1.1

b = B()
f(b)
b.x  # float at runtime, though declared as int
```

Mypy should reject `x: int` in `B`. If this is actually what the programmer intended, they can use `# type: ignore`.

ilevkivskyi:
I agree with the idea, but I think we should have a flag to allow this (``--alow-mutable-overriding``?), otherwise it could be a bit annoying to put ``# type: ignore``s. It will be easy to implement such a flag.

JukkaL:
I don't agree with having a flag -- I don't see why this particular case of unsafety should have a flag dedicated to it. I don't expect that this will require many `# type: ignore`s for most users, and a better fix is to not redefine attributes covariantly (or to use a read-only property).

ilevkivskyi:
Probably a good measure for this would be to just test a future PR against mypy itself and Dropbox internal codebase. If there will be not many errors due to prohibiting covariant overriding of mutable attributes, then probably it is OK to not have a flag.

JukkaL:
We have previously had to make maybe hundreds of changes to Dropbox internal repos in response to fixes to mypy, so our bar for adding a backwards-compatibility flag is pretty high.

ilevkivskyi:
> We have previously had to make maybe hundreds of changes to Dropbox internal repos in response to fixes to mypy, so our bar for adding a backwards-compatibility flag is pretty high.

Ah, OK, I see.

JukkaL:
This is less controversial now, since in many use cases it's possible to use `Final` attributes, for which we can allow safe covariant overriding. This still wouldn't help when the attribute needs to be modified, but I believe that these use cases are somewhat rare, and it's still possible to use a `# type: ignore` comment as a workaround.

hauntsaninja:
If we ever want to do this (which I suspect we may not), https://github.com/python/mypy/issues/14588 points out that we should prohibit:
```
T = TypeVar("T", covariant=True)
class Foo(Generic[T]):
    def __init__(self, value: T) -> None:
        self.foo = value
```

tyralla:
Is covariant overriding of attributes already (possibly accidentally) rejected for multiple inheritance?  

```python
class A:
    x: float | str

[...]

PR #3918, opened 2017-09-05
Fix type checking decorated method overrides

### LINKED ISSUE #1441: Decorator on method in derived class
A decorator that changes the signature of a function is rejected if the old signature isn't consistent with the superclass. E.g.:

```
def convert_to_bool(f):
    # type: (Callable[..., int]) -> Callable[..., bool]
    def wrapper(*n, **kw):
        i = f(*n, **kw)
        return i < 10
    return wrapper

class Base(object):
    def f(self):
        # type: () -> bool
        return True

class Derived(Base):
    @convert_to_bool
    def f(self):
        # type: () -> int
        return 20
```

The interesting point here is that the base class doesn't use the decorator while the child does. Our particular issue is in code that uses contextlib.contextmanager, but I think we can work around it by adding the decorator to the base class.

JukkaL:
The fix turned out to be pretty complicated, as there were a bunch of
untested method overriding scenarios which weren't quite right. I
fixed any related issues that I encountered, though I'm not certain
whether some of the issues were hidden previously by other bugs. I
also added tests for some related, previously untested scenarios.

Fixes #1441.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1015,13 +1015,13 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         else:
             return [(defn, typ)]
 
-    def check_method_override(self, defn: FuncBase) -> None:
+    def check_method_override(self, defn: Union[FuncBase, Decorator]) -> None:
         """Check if function definition is compatible with base classes."""
         # Check against definitions in base classes.
         for base in defn.info.mro[1:]:
             self.check_method_or_accessor_override_for_base(defn, base)
 
-    def check_method_or_accessor_override_for_base(self, defn: FuncBase,
+    def check_method_or_accessor_override_for_base(self, defn: Union[FuncBase, Decorator],
                                                    base: TypeInfo) -> None:
         """Check if method definition is compatible with a base class."""
         if base:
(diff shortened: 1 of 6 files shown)


JukkaL:
This is ready for review, but don't merge this yet -- I have some outstanding fixes to internal Dropbox code bases that I want to land first before merging this.

ilevkivskyi:
It looks like you didn't update all tests, now all types are wrapped in quotes ``"..."``.

mypy/checker.py
```diff
@@ -1058,23 +1071,31 @@ def check_method_override_for_base_with_name(
...
                 self.check_override(typ,
                                     cast(FunctionLike, original),
                                     defn.name(),
                                     name,
                                     base.name(),
-                                    defn)
-            elif isinstance(original_type, AnyType):
+                                    context)
+            elif is_equivalent(original_type, typ):
+                # Assume invariance for a non-callable attribute here.
```
ilevkivskyi: Wait, so does this PR also fix https://github.com/python/mypy/issues/3208?

mypy/checker.py
```diff
@@ -1058,23 +1071,31 @@ def check_method_override_for_base_with_name(
...
+                #
+                # TODO: Allow covariance for read-only attributes?
```
ilevkivskyi: FWIW, protocols currently allow covariant overrides of read-only attributes (like ``@property``).

mypy/checker.py
JukkaL: No -- this only affects decorators that return a non-callable value (but `@property` is unaffected). This is a rare edge case. This also means that this will have only minor backward compatibility issues.

mypy/checker.py
JukkaL: `@property` is special and it can be overridden covariantly even outside protocols. I'll update the comment to mention that this doesn't affect `@property`.

JukkaL:
> What about writable properties?

These should be invariant, similar to regular attributes.

Issue #3895, opened 2017-08-31
Appveyor tests are being flaky


ethanhs:
This was first brought up in #3846. Moved here to avoid clutter in the PR.

You can see the results here: https://ci.appveyor.com/project/JukkaL/mypy/build/1.0.1042

The relevant exception is `OSError: [Errno 9] Bad file descriptor` which I believe is from pytest trying to read from stdout, however for some reason stdout is either locked, unavailable, or Appveyor is incorrectly reporting the handle of stdout.

@elazarg chimed in

>IIRC it seemed to happen when multiple instances of pytest were active. It stopped happening after I've changed the PR to use only a single instance. I'm not sure if this information helps though...

If true, AIUI, this should explain the issue. However, Im not sure we currently spawn multiple pytest instances concurrently, so this may indicate our scheduling logic is broken, or it is an entirely different issue.

gvanrossum:
What makes you think it's related to stdout? The tracebacks all seem to point to this line:
```
    def resume(self):
        self.syscapture.resume()
>       os.dup2(self.tmpfile_fd, self.targetfd)
```
This does appear to be part of a pytest internal class ([FDCapture.resume()](https://github.com/pytest-dev/pytest/blob/523bfa6151081d0aa2385d88d85015731fb0c3cc/_pytest/capture.py#L401)).

ethanhs:
The previous stack is
```python
    def resume_capturing(self):
        if self.out:
            self.out.resume()
        if self.err:
>           self.err.resume()
```

I suppose stderr is more accurate, my apologies, but the effect is the same. This is part of pytest capturing ouput to the console.

ethanhs:
Apparently [PyInstaller had similar issues with pytest-xdist](https://github.com/pytest-dev/pytest-xdist/issues/167), so I believe that we can put the source on issues with pytest-xdist's scheduling.

At pybay @pkch and I saw an interesting talk about [pytest-concurrent](https://github.com/reverbc/pytest-concurrent). It may be too early on for a switch, but it might be nice to use in the future. Eg "allowing certain tests to be grouped so that they are executed sequentially".

ethanhs:
Scrolling through twitter this morning I also ran into this https://twitter.com/pumpichank/status/903280328978178048

So maybe its Appveyor flakiness? Hard to tell.

ethanhs:
Since it has been 3 weeks, I think (I hope) this can be safe to close.

elazarg:
It happened again in #3973

[...]

PR #3919, opened 2017-09-05
Type check arguments for "super"

### LINKED ISSUE #526: Type check super() arguments
Currently the arguments in `super()` calls are not checked.

JukkaL:
This is not quite as strict as we could be, but this is still better
than what we had before. A full implementation will be somewhat
complicated.

Remaining issues:

* Assume that the first argument refers to the enclosing class, even
  though that's not generally the case.

* Assume that a type object as the second argument is always fine.

* Don't validate some kinds of type object types as the first argument.

* Single-argument variant is not supported.

Fixes #526 (modulo the open issues).

Also, I no longer think that defining messages as constants in
`mypy.messages` is generally a good idea so I'm not following
that pattern here. I'll officially update the convention in a
separate PR.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2105,7 +2105,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             elif isinstance(type_obj_type, AnyType):
                 return
             else:
-                self.chk.fail('First argument for "super" must be a type object', e)
+                self.msg.first_argument_for_super_must_be_type(type_obj_type, e)
                 return
 
             if isinstance(instance_type, (Instance, TupleType, TypeVarType)):
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2099,7 +2099,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 if not isinstance(item, Instance):
                     # A complicated type object type. Too tricky, give up.
                     # TODO: Do something more clever here.
-                    self.chk.fail(messages.UNSUPPORTED_ARGUMENT_2_FOR_SUPER, e)
+                    self.chk.fail('Unsupported argument 1 for "super"', e)
                     return
                 type_info = item.type
             elif isinstance(type_obj_type, AnyType):
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2145,9 +2145,12 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                         self.chk.fail('super() outside of a method is not supported', e)
                         return AnyType(TypeOfAny.from_error)
                     args = self.chk.scope.top_function().arguments
-                    # An empty args with super() is an error; we need something in declared_self
+                    # super() in a function with empty args is an error; we
+                    # need something in declared_self.
                     if not args:
-                        self.chk.fail('super() requires at least one positional argument', e)
+                        self.chk.fail(
+                            'super() requires one or more positional arguments in '
+                            'enclosing function', e)
                         return AnyType(TypeOfAny.from_error)
                     declared_self = args[0].variable.type
                     return analyze_member_access(name=e.name, typ=fill_typevars(e.info), node=e,
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2094,9 +2094,12 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 if isinstance(item, AnyType):
                     # Could be anything.
                     return
+                if isinstance(item, TupleType):
+                    item = item.fallback  # Handle named tuples and other Tuple[...] subclasses.
                 if not isinstance(item, Instance):
-                    # A complicated type object type. Assume it's fine.
+                    # A complicated type object type. Too tricky, give up.
                     # TODO: Do something more clever here.
+                    self.chk.fail(messages.UNSUPPORTED_ARGUMENT_2_FOR_SUPER, e)
                     return
                 type_info = item.type
             elif isinstance(type_obj_type, AnyType):
@@ -2111,10 +2114,11 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                     instance_type = instance_type.upper_bound
                     if not isinstance(instance_type, (Instance, TupleType)):
                         # Too tricky, give up.
-                        self.chk.fail('Unsupported argument 2 for "super"', e)
+                        # TODO: Do something more clever here.
+                        self.chk.fail(messages.UNSUPPORTED_ARGUMENT_2_FOR_SUPER, e)
                         return
                 if isinstance(instance_type, TupleType):
-                    # Needed for named tuples and classes derived from Tuple[...].
+                    # Needed for named tuples and other Tuple[...] subclasses.
                     instance_type = instance_type.fallback
                 if type_info not in instance_type.type.mro:
                     self.chk.fail('Argument 2 for "super" not an instance of argument 1', e)
@@ -2123,7 +2127,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 # TODO: Check whether this is a valid type object here.
                 pass
             elif not isinstance(instance_type, AnyType):
-                self.chk.fail('Unsupported argument 2 for "super"', e)
+                self.chk.fail(messages.UNSUPPORTED_ARGUMENT_2_FOR_SUPER, e)
 
     def analyze_super(self, e: SuperExpr, is_lvalue: bool) -> Type:
         """Type check a super expression."""
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2066,8 +2066,62 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
     def visit_super_expr(self, e: SuperExpr) -> Type:
         """Type check a super expression (non-lvalue)."""
+        self.check_super_arguments(e)
         t = self.analyze_super(e, False)
         return t
+
+    def check_super_arguments(self, e: SuperExpr) -> None:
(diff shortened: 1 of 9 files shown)


JukkaL:
This is ready for review but this is not ready to be merged yet, as this requires some changes to internal Dropbox code.

I'll create follow-up issues for any remaining open issues after this has been merged.

ethanhs:
Appveyor failure is a flake (see #3895)

mypy/checkexpr.py
```diff
@@ -2066,9 +2066,63 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+        if ARG_STAR in e.call.arg_kinds:
+            self.chk.fail('Varargs not supported with "super"', e)
+        elif e.call.args and set(e.call.arg_kinds) != {ARG_POS}:
+            self.chk.fail('"super" only accepts positional arguments', e)
+        elif len(e.call.args) == 1:
+            self.chk.fail('"super" with a single argument not supported', e)
+        elif len(e.call.args) > 2:
+            self.chk.fail('Too many arguments for "super"', e)
+        elif self.chk.options.python_version[0] == 2 and len(e.call.args) == 0:
+            self.chk.fail('Too few arguments for "super"', e)
```
ilevkivskyi: I would move all the logic until this point to ``semanal.py``, but it is optional.

mypy/checkexpr.py
```diff
@@ -2066,9 +2066,63 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+                    return
+                if not isinstance(item, Instance):
+                    # A complicated type object type. Assume it's fine.
+                    # TODO: Do something more clever here.
+                    return
+                type_info = item.type
+            elif isinstance(type_obj_type, AnyType):
+                return
+            else:
+                self.chk.fail('First argument for "super" must be a type object', e)
```
ilevkivskyi: I would add ``Got: "<actual type>"``

mypy/checkexpr.py
```diff
@@ -2066,9 +2066,63 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+                return
+
+            if isinstance(instance_type, (Instance, TupleType, TypeVarType)):
+                if isinstance(instance_type, TypeVarType):
+                    # Needed for generic self.
+                    instance_type = instance_type.upper_bound
+                    if not isinstance(instance_type, (Instance, TupleType)):
+                        # Too tricky, give up.
+                        self.chk.fail('Unsupported argument 2 for "super"', e)
```
ilevkivskyi: This is a bit inconsistent with the first argument logic above, where we give up by a simple ``return`` without any error message, why it is so?

mypy/checkexpr.py
```diff
@@ -2066,9 +2066,63 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+                    # Needed for named tuples and classes derived from Tuple[...].
+                    instance_type = instance_type.fallback
+                if type_info not in instance_type.type.mro:
+                    self.chk.fail('Argument 2 for "super" not an instance of argument 1', e)
+            elif isinstance(instance_type, TypeType) or (isinstance(instance_type, FunctionLike)
+                                                         and instance_type.is_type_obj()):
+                # TODO: Check whether this is a valid type object here.
+                pass
+            elif not isinstance(instance_type, AnyType):
+                self.chk.fail('Unsupported argument 2 for "super"', e)
```
ilevkivskyi: My question above about consistency also applies here. Also in this case the error message is repeated, so maybe it makes sense to define a constant for it?

mypy/checkexpr.py
```diff
@@ -2066,9 +2066,63 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+
     def analyze_super(self, e: SuperExpr, is_lvalue: bool) -> Type:
```
ilevkivskyi: I remember looking at the function ``analyze_super`` and something worried me. Namely, look at the error:
``super() requires at least one positional argument`` several lines below (can't comment there), it is misleading since it can appear on Python 3 as well but we support ``super()`` without arguments. Also the logic there seems a bit wrong.

mypy/treetransform.py
```diff
@@ -392,7 +392,7 @@ def visit_reveal_type_expr(self, node: RevealTypeExpr) -> RevealTypeExpr:
         return RevealTypeExpr(self.expr(node.expr))
 
     def visit_super_expr(self, node: SuperExpr) -> SuperExpr:
-        new = SuperExpr(node.name)
+        new = SuperExpr(node.name, cast(CallExpr, self.expr(node.call)))
```
ilevkivskyi: Why a cast needed here? Is it possible to replace it with an assert?

test-data/unit/check-super.test
```diff
@@ -115,3 +111,138 @@ class B(A):
...
+        reveal_type(super(x, x).f) # E: Revealed type is 'def ()'
+        reveal_type(super(C, x).f) # E: Revealed type is 'def ()'
+
+[case testSuperInUnannotatedMethod]
+class C:
+    def h(self):
+        super(C, self).xyz
+
+[case testSuperWithTypeType]
+from typing import Type
```
ilevkivskyi: You are not using this import, have you forgotten to test something here?

test-data/unit/check-super.test
```diff
@@ -115,3 +111,138 @@ class B(A):
...
+
+class C(B):
+    def f(self) -> None: pass
+
+    def g(self: T) -> T:
+        reveal_type(super(C, self).f) # E: Revealed type is 'def ()'
+        return self
+
+[case testSuperWithTypeVarValues]
+from typing import TypeVar, Generic
```
ilevkivskyi: You imported ``Generic`` but didn't use it. Did you want to add another test also here? I would actually add some tests with ``Generic``.

test-data/unit/check-super.test
```diff
@@ -115,3 +111,138 @@ class B(A):
...
+[case testSuperWithGenericSelf]
+from typing import TypeVar
+
+T = TypeVar('T', bound='C')
+
+class B:
+    def f(self) -> None: pass
+
+class C(B):
+    def f(self) -> None: pass
```
ilevkivskyi: I would make return types different here and in superclass (e.g. ``float`` and ``int``), so that we could be sure in ``reveal_type`` that the correct one is picked up.

mypy/checkexpr.py
JukkaL: I'm keeping the logic here because we want to execute the final `elif` block only if all the previous checks are false, and having all the checks here makes it explicit. Also, we check the argument counts of other calls during type checking (though `super()` is special so we could do it during semantic analysis as well).

mypy/checkexpr.py
JukkaL: Updated, but I still don't display the class name, since it looked confusing (due to type of instance `C` looking like a reference `C` to type object).

mypy/checkexpr.py
JukkaL: I now generate error above as well.

mypy/checkexpr.py
JukkaL: Updated to use a constant for the error message.

mypy/checkexpr.py
JukkaL: The message refers to the enclosing function not having any positional arguments. Updated the wording. If it's still incorrect, we can create a new issue about it, since this looks like a pretty marginal scenario. The logic is kind of convoluted though -- most of the for loop body is executed at most once, which is not immediately obvious.

mypy/treetransform.py
JukkaL: Replaced with an assert.

test-data/unit/check-super.test
JukkaL: Yes, added another thing to test.

test-data/unit/check-super.test
JukkaL: Updated.

test-data/unit/check-super.test
JukkaL: Added a test that uses `Generic`.

mypy/checkexpr.py
ilevkivskyi: You already have a list of things still left to fix in the PR description, if you feel this deserves a separate issue, then OK, otherwise you can just combine this with first item in your list.

mypy/checkexpr.py
```diff
@@ -2068,9 +2068,67 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+                item = type_obj_type.item
+                if isinstance(item, AnyType):
+                    # Could be anything.
+                    return
+                if isinstance(item, TupleType):
+                    item = item.fallback  # Handle named tuples and other Tuple[...] subclasses.
+                if not isinstance(item, Instance):
+                    # A complicated type object type. Too tricky, give up.
+                    # TODO: Do something more clever here.
+                    self.chk.fail(messages.UNSUPPORTED_ARGUMENT_2_FOR_SUPER, e)
```
ilevkivskyi: Should it be actually argument _1_ here?

JukkaL: Yes! Added a test case and fixed the error message.

mypy/checkexpr.py
JukkaL: It's still unclear to me what the issue is (other than the code being hard to understand). Can you create a new issue (or add a comment to an existing relevant issue) about this?

PR #3920, opened 2017-09-05
Update convention about implenting messages in the type checker

JukkaL:
Using constants for all type checker messages doesn't seem very
valuable and it introduces some unnecessary friction when defining
simple messages.

I don't think that it's useful to refactor away the existing
constants, though. The cost of the old convention is primarily in
writing new code, not in reading existing code.


ilevkivskyi:
I think this is right. I might even go as far as to remove all constants that used only once, since their only role now is to require one more ``grep`` when start working on a new issue :-) but probably this would be an overkill.

Issue #3881, opened 2017-08-26
Internal Error with forward references


garrettheel:
Reporting a bug I found.

```bash
$ mypy mypy_bug.py --show-traceback
mypy_bug.py:15: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.521
Traceback (most recent call last):
  File "/Users/gheel/.virtualenvs/mypy/bin/mypy", line 11, in <module>
    sys.exit(console_entry())
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/__main__.py", line 7, in console_entry
    main(None)
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/main.py", line 50, in main
    res = type_check_only(sources, bin_dir, options)
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/main.py", line 97, in type_check_only
    options=options)
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/build.py", line 196, in build
    graph = dispatch(sources, manager)
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/build.py", line 1801, in dispatch
    process_graph(graph, manager)
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/build.py", line 2044, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/build.py", line 2147, in process_stale_scc
    graph[id].type_check_first_pass()
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/build.py", line 1716, in type_check_first_pass
    self.type_checker.check_first_pass()
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/checker.py", line 185, in check_first_pass
    self.accept(d)
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/checker.py", line 273, in accept
    stmt.accept(self)
  File "/Users/gheel/.virtualenvs/mypy/lib/python3.6/site-packages/mypy/nodes.py", line 754, in accept
    return visitor.visit_class_def(self)
[...]

Issue #3836, opened 2017-08-16
INTERNAL ERROR: 'NoneType' object is not iterable


martincmartin:
```
mypy --show-traceback Deduction.py
Deduction.py:102: error: The return type of "__init__" must be None
Deduction.py:103: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.521-858f7512cf3b4e39c0f4e8de5a13eee0e1e138fb-dirty
Traceback (most recent call last):
  File "/usr/local/Cellar/mypy/0.521/libexec/bin/mypy", line 11, in <module>
    load_entry_point('mypy===0.521-858f7512cf3b4e39c0f4e8de5a13eee0e1e138fb-dirty', 'console_scripts', 'mypy')()
  File "/usr/local/Cellar/mypy/0.521/libexec/lib/python3.6/site-packages/mypy/__main__.py", line 7, in console_entry
    main(None)
  File "/usr/local/Cellar/mypy/0.521/libexec/lib/python3.6/site-packages/mypy/main.py", line 50, in main
    res = type_check_only(sources, bin_dir, options)
  File "/usr/local/Cellar/mypy/0.521/libexec/lib/python3.6/site-packages/mypy/main.py", line 97, in type_check_only
    options=options)
  File "/usr/local/Cellar/mypy/0.521/libexec/lib/python3.6/site-packages/mypy/build.py", line 196, in build
    graph = dispatch(sources, manager)
  File "/usr/local/Cellar/mypy/0.521/libexec/lib/python3.6/site-packages/mypy/build.py", line 1801, in dispatch
    process_graph(graph, manager)
  File "/usr/local/Cellar/mypy/0.521/libexec/lib/python3.6/site-packages/mypy/build.py", line 2044, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/usr/local/Cellar/mypy/0.521/libexec/lib/python3.6/site-packages/mypy/build.py", line 2147, in process_stale_scc
    graph[id].type_check_first_pass()
  File "/usr/local/Cellar/mypy/0.521/libexec/lib/python3.6/site-packages/mypy/build.py", line 1716, in type_check_first_pass
    self.type_checker.check_first_pass()
  File "/usr/local/Cellar/mypy/0.521/libexec/lib/python3.6/site-packages/mypy/checker.py", line 185, in check_first_pass
    self.accept(d)
[...]

Issue #3799, opened 2017-08-04
Internal error using Union and NamedTuple


n-n-n-n-n-n-n:
mypy version 0.521, python 3.6.2

I tried to find the smallest example that would reproduce the bug I'm getting.
Removing references to NamedTuple or moving the type alias below the classes seems to stop the error.

~~~
$ cat bug.py 
from typing import Union, NamedTuple

# Moving this below the class declarations stops the error
NodeType = Union['Foo', 'Bar']

# Removing NamedTuple from both classes stops the error
class Foo(NamedTuple):
    pass

class Bar(NamedTuple):
    pass

def foo(node: NodeType):
    x = node
~~~

I got the following error

~~~
$ mypy bug.py 
bug.py:14: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.521
bug.py:14: note: please use --show-traceback to print a traceback when reporting a bug
~~~

And the traceback

~~~
$ mypy --show-traceback bug.py 
bug.py:14: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.521
Traceback (most recent call last):
  File "/usr/bin/mypy", line 11, in <module>
    sys.exit(console_entry())
  File "/usr/lib/python3.6/site-packages/mypy/__main__.py", line 7, in console_entry
    main(None)
  File "/usr/lib/python3.6/site-packages/mypy/main.py", line 50, in main
    res = type_check_only(sources, bin_dir, options)
  File "/usr/lib/python3.6/site-packages/mypy/main.py", line 97, in type_check_only
    options=options)
  File "/usr/lib/python3.6/site-packages/mypy/build.py", line 196, in build
    graph = dispatch(sources, manager)
  File "/usr/lib/python3.6/site-packages/mypy/build.py", line 1801, in dispatch
    process_graph(graph, manager)
  File "/usr/lib/python3.6/site-packages/mypy/build.py", line 2044, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/usr/lib/python3.6/site-packages/mypy/build.py", line 2147, in process_stale_scc
    graph[id].type_check_first_pass()
  File "/usr/lib/python3.6/site-packages/mypy/build.py", line 1716, in type_check_first_pass
    self.type_checker.check_first_pass()
  File "/usr/lib/python3.6/site-packages/mypy/checker.py", line 185, in check_first_pass
    self.accept(d)
[...]

Issue #3685, opened 2017-07-10
Infinite recursion with recursive TypedDict


JukkaL:
Mypy crashes with `RecursionError` when I type check this program:

```py
from mypy_extensions import TypedDict

A = TypedDict('A', {'a': 'A'})

a: A
```

Output:

```py
t2.py:3: error: Invalid type "t2.A"
Traceback (most recent call last):
  File "/Users/jukka/src/mypy/scripts/mypy", line 6, in <module>
    main(__file__)
  File "/Users/jukka/src/mypy/mypy/main.py", line 50, in main
    res = type_check_only(sources, bin_dir, options)
  File "/Users/jukka/src/mypy/mypy/main.py", line 97, in type_check_only
    options=options)
  File "/Users/jukka/src/mypy/mypy/build.py", line 196, in build
    graph = dispatch(sources, manager)
  File "/Users/jukka/src/mypy/mypy/build.py", line 1769, in dispatch
    process_graph(graph, manager)
  File "/Users/jukka/src/mypy/mypy/build.py", line 2012, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/Users/jukka/src/mypy/mypy/build.py", line 2107, in process_stale_scc
    graph[id].semantic_analysis()
  File "/Users/jukka/src/mypy/mypy/build.py", line 1664, in semantic_analysis
    self.manager.semantic_analyzer.visit_file(self.tree, self.xpath, self.options, patches)
  File "/Users/jukka/src/mypy/mypy/semanal.py", line 295, in visit_file
    self.accept(d)
  File "/Users/jukka/src/mypy/mypy/semanal.py", line 3636, in accept
    node.accept(self)
  File "/Users/jukka/src/mypy/mypy/nodes.py", line 859, in accept
    return visitor.visit_assignment_stmt(self)
  File "/Users/jukka/src/mypy/mypy/semanal.py", line 1567, in visit_assignment_stmt
    s.type = self.anal_type(s.type, allow_tuple_literal=allow_tuple_literal)
  File "/Users/jukka/src/mypy/mypy/semanal.py", line 1555, in anal_type
    return t.accept(a)
  File "/Users/jukka/src/mypy/mypy/types.py", line 189, in accept
    return visitor.visit_unbound_type(self)
  File "/Users/jukka/src/mypy/mypy/typeanal.py", line 292, in visit_unbound_type
    return td.copy_modified(item_types=self.anal_array(list(td.items.values())),
[...]

PR #3952, opened 2017-09-13
Fix crashes and fails in forward references

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/3340
Fixes https://github.com/python/mypy/issues/3419
Fixes https://github.com/python/mypy/issues/3674
Fixes https://github.com/python/mypy/issues/3685
Fixes https://github.com/python/mypy/issues/3799
Fixes https://github.com/python/mypy/issues/3836
Fixes https://github.com/python/mypy/issues/3881
Fixes https://github.com/python/mypy/issues/867
Fixes https://github.com/python/mypy/issues/2241
Fixes https://github.com/python/mypy/issues/2399
Fixes https://github.com/python/mypy/issues/1701
Fixes https://github.com/python/mypy/issues/3016
Fixes https://github.com/python/mypy/issues/3054
Fixes https://github.com/python/mypy/issues/2762
Fixes https://github.com/python/mypy/issues/3575
Fixes https://github.com/python/mypy/issues/3990

Currently, forward references don't work with anything apart from classes, for example this doesn't work:
```python
x: A
A = NamedTuple('A', [('x', int)])
```
The same situation is with ``TypedDict``s, ``NewType``s, and type aliases. The root problem is that these synthetic types are neither detected in first pass, nor fixed in third pass. In certain cases this can lead to _crashes_ (first six issues above are various crash scenarios). I fix these crashes by applying some additional patches after third pass. Here is the summary of the PR:
* New simple wrapper type ``ForwardRef`` with only one field ``link`` is introduced (with updates to type visitors)
* When an unknown type is found in second pass, the corresponding ``UnboundType`` is wrapped in ``ForwardRef``, it is given a "second chance" in third pass.
* After third pass I record the "suspicious" nodes, where forward references and synthetic types have been encountered and append patches (callbacks) to fix them after third pass. Patches use the new visitor ``TypeReplacer`` (which is the core of this PR).

Here are two problems that I encountered:
* Third pass (both in ``semanal.py`` and in ``typeanal.py``) was more "shallow" than the second one, some visitor methods were literally ``pass``. It was necessary to update these to match the "depth" of the second pass.
* Now third pass has a link to second pass analyzer, ``self.sem``, same as for first pass. It would be nice to refactor the passes since all three share some code/functionality, there is already https://github.com/python/mypy/issues/3459 to track this.

NOTE:  self-referential types are still _not_ properly supported, but now we give a reasonable error for this, not a crash, and they still can be used to certain extent, for example:
```python
class MyNamedTuple(NamedTuple):
    parent: 'MyNamedTuple'

def get_parent(nt: MyNamedTuple) -> MyNamedTuple:
    return nt.parent

x: MyNamedTuple
reveal_type(x.parent)
reveal_type(x[0])
```
results in
```
main:2: error: Recursive types not fully supported yet, nested types replaced with "Any"
main:9: error: Revealed type is 'Tuple[Any, fallback=__main__.MyNamedTuple]
main:10: error: Revealed type is 'Tuple[Any, fallback=__main__.MyNamedTuple]
```
Proper support of recursive types would be much harder (mostly due to (de-)serialization that would require something similar to ``type_ref``), there is a separate issue for this https://github.com/python/mypy/issues/731.

@JukkaL sorry it took a bit longer than expected because of CPython sprint last week.


ilevkivskyi:
An important comment, for some reason forward references _between files_ still does not work properly (but at least the y don't crash).

ilevkivskyi:
This is ready for review, but I marked it WIP since I have found one more crash, will fix it shortly.

ilevkivskyi:
OK, I have fixed some remaining problems and TODO's, also added more tests.

JukkaL:
Thank you very much for implementing this! Based on a quick pass, it looks like this fixes many long-standing issues in a clean way. Forward reference handling will be also be useful for fine-grained incremental checking, which we are planning to continue working on later this year. Special thanks for writing many test cases.

I'll try to do a full review by mid next week. I'll also run this against internal Dropbox codebases to see if there are regressions (or performance issues).

Also, there's now a merge conflict. Can you fix it?

ilevkivskyi:
> Also, there's now a merge conflict. Can you fix it?

Fixed.
> Special thanks for writing many test cases.

I am glad you like the tests. But this is a subtle area, if you have ideas for more tests, then I will be grateful (I am sure this PR does not fix _all_ issues with forward references, for example across files as I mentioned above, but this could be a good start).

JukkaL:
I ran this against internal Dropbox repos and found one minor difference. Previously this code didn't generate an error:

```py
from typing import AnyStr, Dict

def f():   # Note no annotation
    x = {}  # type: Dict[str, AnyStr]
```

Now it generates this error:

```
t.py:4: error: Invalid type "typing.AnyStr"
```

Also I there seems to be no significant performance impact, which is good.

I'll continue tomorrow with a more detailed review.

ilevkivskyi:
@JukkaL In your absence I did the following:
* Fixed the thing you found by just skipping all the ``ForwardRef`` dance in unchecked functions
* Fixed the cross-file forward references I mentioned above (just one line, update gloabals on every file in third pass).
* Found more issue that will be fixed by this PR, updated the description and added corresponding tests.
* Fixed merge conflict.

ilevkivskyi:
@JukkaL One more thing, I just fixed a situation when recursive named tuple is _defined_ directly in base expressions, this is probably super rare, but I did it just for completeness.

ilevkivskyi:
Probably it makes sense to summarize what are the limitations of this PR, I think there are three:

**1. Proper support for recursive types:**
This has been deliberately postponed.

**2. Mutually recursive aliases between files:**
```python
# file a.py
form b import B
A = Union[int, Tuple[B]]
# file b.py
from a import A
B = Union[int, Tuple[A]]
```
This works with normal types (not aliases), will not crash, and the analyzed types are reasonable (expanded aliases with nested types replaced with ``Any``), but this will also emit a spurious ``Invalid type`` error. Suppressing it is hard, since this will require reprocessing imports in third pass, which may have speed penalties. I don't want this, since currently forward references are more like "prepaid" rather than "contract", if one does not use them, there should be no performance penalty.

**3. Double forward references in some contexts:**
In some special corner cases it is impossible to fix all the forward references in one pass of patches, for example things like this
```python
x: A
A = B
B = List[int]
```
will result in ``Invalid type`` kind of errors (but not crashes). There could be other special cases like this, but I think most situations with double forward references are covered:
```python
x: A
A = List[B]
B = List[C]
class C(NamedTuple):
    x: int

reveal_type(x[0][0].x) # Revealed type is 'builtins.int'
```

mypy/typeanal.py
```diff
@@ -712,13 +739,22 @@ def visit_type_list(self, t: TypeList) -> None:
...
         pass
 
     def visit_type_type(self, t: TypeType) -> None:
-        pass
+        t.item.accept(self)
+
+    def visit_forwardref_type(self, t: ForwardRef) -> None:
+        self.indicator['forward'] = True
+        if isinstance(t.link, UnboundType):
+            t.link = self.sem.anal_type(t.link, third_pass=True)
```
JukkaL: This is kind of hacky -- we call the second pass of semantic analysis from the third pass of type analysis, without having all the normal context for the second pass set up. This causes some errors:

```py
from typing import List, TypeVar

T = TypeVar('T')

def f(x: T) -> None:
    y: A[T]  # "Invalid type "t.T"

A = List[T]
```

Another example that doesn't work because of this issue:

```py
    from typing import List, TypeVar

    def f() -> None:
        X = List[int]
        x: A[X]  # "Invalid type X?"

    T = TypeVar('T')
    A = List[T]
```

Not sure what's the best way to approach this. We perhaps don't want to set up the full context for second pass during the third pass, as this can add a lot of extra complexity. A simpler workaround would be to not support forward references with type arguments, and give an error such as `Forward reference to "A" not supported`. This at least would fix the current error messages, which are confusing.

Also, I'd prefer if we didn't depend on `mypy.semanal` here. A better approach would be to remove the cyclic import to `mypy.semanal` and call `TypeAnalyser` directly here. `TypeAnalyserPass3` would take the lookup functions etc. as arguments instead of getting them indirectly from `SemanticAnalyzer`. The rationale is that this makes it explicit what context the third pass depends on, and it will make the code clearer. Also, we won't depend on the internals of `SemanticAnalyzer`.

mypy/types.py
```diff
@@ -1375,6 +1375,26 @@ def deserialize(cls, data: JsonDict) -> Type:
...
+        return visitor.visit_forwardref_type(self)
+
+    def serialize(self):
+        if isinstance(self.link, UnboundType):
+            name = self.link.name
+        if isinstance(self.link, Instance):
+            name = self.link.type.name()
+        else:
+            name = self.link.__class__.__name__
+        assert False, "Internal error: Unresolved forward reference to {}".format(name)
```
JukkaL: Add comment about why we shouldn't get here (all forward references should be resolved and removed during semantic analysis).

test-data/unit/check-classes.test
```diff
@@ -3515,9 +3515,224 @@ D = TypeVar('D')
...
+from typing import NamedTuple
+
+N = NamedTuple('N', [('x', N)]) # E: Recursive types not fully supported yet, nested types replaced with "Any"
+n: N
+[out]
+
+[case testCrashOnSelfRecursiveTypedDictVar]
+from mypy_extensions import TypedDict
+
+A = TypedDict('A', {'a': 'A'})  # type: ignore
```
JukkaL: Why use a `# type: ignore` here?

test-data/unit/check-classes.test
```diff
@@ -3515,9 +3515,224 @@ D = TypeVar('D')
...
+[case testCorrectIsinstanceInForwardRefToNewType]
+from typing import Dict, NewType
+
+NameDict = Dict[str, 'NameInfo']
+class NameInfoBase:
+    ast: bool
+NameInfo = NewType('NameInfo', NameInfoBase)
+
+def parse_ast(name_dict: NameDict) -> None:
+    if isinstance(name_dict[''], int):
```
JukkaL: Reveal tyhe type of `name_dict` and verify that it actually behaves like a `NewType`.

mypy/semanal.py
```diff
@@ -4162,37 +4201,96 @@ def visit_decorator(self, dec: Decorator) -> None:
                 dec.var.type = sig
 
     def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
-        self.analyze(s.type)
+        """Traverse the actual assignment statement and synthetic types
```
JukkaL: Style nit: docstring formatting (paragraph break after first line).

mypy/semanal.py
```diff
@@ -4162,37 +4201,96 @@ def visit_decorator(self, dec: Decorator) -> None:
...
+                self.analyze(analyzed.info.typeddict_type, analyzed, warn=True)
+            if isinstance(analyzed, NamedTupleExpr):
+                self.analyze(analyzed.info.tuple_type, analyzed, warn=True)
+                for name in analyzed.info.names:
+                    sym = analyzed.info.names[name]
+                    if isinstance(sym.node, (FuncDef, Decorator)):
+                        self.accept(sym.node)
+                    if isinstance(sym.node, Var):
+                        self.analyze(sym.node.type, sym.node)
+        if isinstance(s.lvalues[0], RefExpr) and isinstance(s.lvalues[0].node, Var):
```
JukkaL: It's unclear what this is for. Please add a comment.

mypy/semanal.py
```diff
@@ -4162,37 +4201,96 @@ def visit_decorator(self, dec: Decorator) -> None:
...
         super().visit_type_application(e)
 
     # Helpers
 
-    def analyze(self, type: Optional[Type]) -> None:
+    def perform_transform(self, node: Union[Node, SymbolTableNode],
+                          transform: Callable[[Type], Type]) -> None:
+        """Apply transform to all types associated with node."""
+        if isinstance(node, (FuncDef, CastExpr, AssignmentStmt, TypeAliasExpr, Var)):
+            node.type = transform(node.type)
```
JukkaL: What about `type_annotation` of `Argument`? Should we process it?

mypy/semanal.py
```diff
@@ -4162,37 +4201,96 @@ def visit_decorator(self, dec: Decorator) -> None:
...
         for type in e.types:
-            self.analyze(type)
+            self.analyze(type, e)
         super().visit_type_application(e)
 
     # Helpers
 
-    def analyze(self, type: Optional[Type]) -> None:
+    def perform_transform(self, node: Union[Node, SymbolTableNode],
+                          transform: Callable[[Type], Type]) -> None:
```
JukkaL: Is `type_vars` of `ClassDef` transformed somewhere? What about `index_type` of `ForStmt` and `target_type` of `WithStmt`? And `TypeVarExpr`?

ilevkivskyi:
@JukkaL Thanks for review! Here are some quick replies, I will push the commits later.

> Do forward references to (functional) enums work?
> Forward references to type variables don't seem to work.

Yes, I somehow forgot about these two. Enums should be easy, type variables are much harder, especially from the runtime point of view, forward references to type variables don't work at runtime. I think I will try to give a more meaningful message.

mypy/typeanal.py
ilevkivskyi: Yes, this is one of the hard parts. I will try to see what is the minimal context needed, basically we just need ``lookup`` to work in third pass. If it will be too hard, I will just prohibit things like ``F[<something>]``, where ``F`` is a forward reference, with a clearer error message.

test-data/unit/check-classes.test
ilevkivskyi: I had some problems with line numbers where ``Recursive types are not supported...`` error is reported. So I wanted to check that they are silenced correctly. Also in some tests I want to trigger serialization, IIUC serialization doesn't happen if there are errors.

mypy/semanal.py
ilevkivskyi: ``type_vars`` and ``TypeVarExpr`` are probably not needed (I will double-check), but it looks like I just forgot about ``WithStmt`` and ``ForStmt``. I will add processing.

mypy/semanal.py
ilevkivskyi: I was not able to trigger any errors with this, I will double check.

ilevkivskyi:
@JukkaL I started addressing your comments, I will try to finish before the end of this week.

While making changes you requested, I remembered something important I found before: third pass calls ``is_subtype`` and ``is_same_types``. I think this is not the right place to do this, and may be dangerous. Currently I overcome this problem with a hack, but I think we should move these calls to a later stage (either callbacks or type checker) in this PR or in a separate PR.

ethanhs:
Looks like travis flaked. I restarted that job.

mypy/semanal.py
```diff
@@ -4618,3 +4716,123 @@ def visit_any(self, t: AnyType) -> Type:
...
         return t
+
+
+class TypeReplacer(TypeTranslator):
+    """This is very similar TypeTranslator but tracks visited nodes to avoid
+    infinite recursion on potentially circular (self- or mutually-referential) types.
+    """
+    def __init__(self, fail: Callable[[str, Context], None],
+                 start: Union[Node, SymbolTableNode], warn: bool) -> None:
+        self.seen = []  # type: List[Type]
```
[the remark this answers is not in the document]
ilevkivskyi: This actually does not work, I get some weird failures, when I change this to a set. I am not sure why, but my guess is that Python reuses ``id``s for deleted objects.

ilevkivskyi:
@JukkaL I think I have now addressed all your comments. This is now ready for further review.

mypy/types.py
```diff
@@ -1376,8 +1376,21 @@ def deserialize(cls, data: JsonDict) -> Type:
...
-    link = None  # type: Type # the wrapped type
+    """Class to wrap forward references to other types.
+
+    This is used when a forward reference to an (unanalyzed) synthetic type is found,
+    for example:
+
+        x: A
+        A = TypedDict('A', {'x': int})
+
+    To avoid false positives and crashes in such situations, we first wrap the second
```
JukkaL: Should this be 'the first occurrence ...'?

JukkaL:
Most of my comments from the last round are shown as outdated because I screwed up things a little. I think that they should mostly be still relevant though.

ilevkivskyi:
> Most of my comments from the last round are shown as outdated because I screwed up things a little. I think that they should mostly be still relevant though.

By last round you mean few minutes ago ago or few days ago? I see almost everything as outdated, I assume you mean few minutes ago, I received an e-mail from GitHub, so that I can see these comments.

test-data/unit/check-typeddict.test
```diff
@@ -1326,6 +1326,22 @@ reveal_type(m['director']['name']) # E: Revealed type is 'builtins.str'
...
 [out]
 
+[case testTypedDictForwardAsUpperBound]
+from typing import TypeVar, Generic
+from mypy_extensions import TypedDict
+T = TypeVar('T', bound='M')
+class G(Generic[T]):
+    x: T
+
+yb: G[int]
```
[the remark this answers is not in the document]
ilevkivskyi: This is rejected, here and above. I see the problem, it looks like you reviewed only a commit range, not all changes in PR.

I will anyway go though all comments this evening. Thanks for review!

mypy/types.py
ilevkivskyi: Yes, will fix this docstring.

mypy/typeanal.py
```diff
@@ -672,12 +673,28 @@ def visit_instance(self, t: Instance) -> None:
...
                         arg_values = [arg]
                     self.check_type_var_values(info, arg_values, tvar.name, tvar.values, i + 1, t)
+                # TODO: These hacks will be not necessary when this will be moved to later stage.
                 if isinstance(arg, ForwardRef):
                     arg = arg.link
-                if not is_subtype(arg, tvar.upper_bound):
+                bound = tvar.upper_bound
+                if isinstance(bound, ForwardRef):
+                    bound = bound.link
+                if isinstance(bound, Instance) and bound.type.replaced:
```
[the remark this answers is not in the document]
ilevkivskyi: I will minimize duplication (but I can't use exactly the code from visitor in ``semanal.py`` it is too soon here).

ilevkivskyi:
@JukkaL I think I now implemented all your recent comments. In addition I fixed one more minor crash on MRO being ``None`` for ``NewType``s and improved tests. This includes:
* Added more ``reveal_type``s here and there.
* Used shorted and/or less distracting names.
* Fixed formatting to make tests more compact.

JukkaL:
Thanks for the updates! Glad to see so many bugs fixed.

PR #3957, opened 2017-09-14
[clean-strict-optional] Clean-up the type checking files

ilevkivskyi:
This makes another major piece of mypy ``--strict-optional`` clean: the type checking files.
After this PR there will be only single exception left ``semanal.py``.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -236,7 +236,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 isinstance(callee_type, CallableType)
                 and callee_type.implicit):
             return self.msg.untyped_function_call(callee_type, e)
-        # Figure out the full name of the callee for plugin loopup.
+        # Figure out the full name of the callee for plugin lookup.
         object_type = None
         if not isinstance(e.callee, RefExpr):
             fullname = None
@@ -260,6 +260,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                     # Apply plugin signature hook that may generate a better signature.
                     signature_hook = self.plugin.get_method_signature_hook(fullname)
                     if signature_hook:
+                        assert object_type is not None
                         callee_type = self.apply_method_signature_hook(
                             e, callee_type, object_type, signature_hook)
         ret_type = self.check_call_expr_with_callee_type(callee_type, e, fullname, object_type)
@@ -1715,7 +1716,6 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             if stride is None:
                 return self.nonliteral_tuple_index_helper(left_type, slic)
 
-        assert begin and stride and end
         return left_type.slice(begin, stride, end)
 
     def nonliteral_tuple_index_helper(self, left_type: TupleType, index: Expression) -> Type:
@@ -1960,6 +1960,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         vtdef = TypeVarDef('VT', -2, [], self.object_type())
         kt = TypeVarType(ktdef)
         vt = TypeVarType(vtdef)
+        rv = None  # type: Optional[Type]
         # Call dict(*args), unless it's empty and stargs is not.
         if args or not stargs:
             # The callable type represents a function like this:
@@ -1976,7 +1977,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             rv = self.check_call(constructor, args, [nodes.ARG_POS] * len(args), e)[0]
         else:
             # dict(...) will be called below.
-            rv = None
+            pass
         # Call rv.update(arg) for each arg in **stargs,
         # except if rv isn't set yet, then set rv = dict(arg).
         if stargs:
@@ -2176,8 +2177,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                             'super() requires one or more positional arguments in '
                             'enclosing function', e)
                         return AnyType(TypeOfAny.from_error)
-                    declared_self = args[0].variable.type
-                    assert declared_self is not None, "Internal error: type of self is None"
+                    declared_self = args[0].variable.type or fill_typevars(e.info)
                     return analyze_member_access(name=e.name, typ=fill_typevars(e.info), node=e,
                                                  is_lvalue=False, is_super=True, is_operator=False,
                                                  builtin_type=self.named_type,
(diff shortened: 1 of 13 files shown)


mypy/checkexpr.py
```diff
@@ -896,9 +900,10 @@ def apply_inferred_arguments(self, callee_type: CallableType,
...
 
     def check_argument_count(self, callee: CallableType, actual_types: List[Type],
-                             actual_kinds: List[int], actual_names: List[str],
+                             actual_kinds: List[int],
+                             actual_names: Optional[Sequence[Optional[str]]],
                              formal_to_actual: List[List[int]],
-                             context: Context,
+                             context: Optional[Context],
                              messages: Optional[MessageBuilder]) -> bool:
         """Check that there is a value for all required arguments to a function.
```
gvanrossum: Maybe note in the docstring that if `messages` is given, `context` must also be given?

mypy/checkexpr.py
```diff
@@ -1351,13 +1368,13 @@ def visit_comparison_expr(self, e: ComparisonExpr) -> Type:
...
         """
-        result = None  # type: mypy.types.Type
+        result = None  # type: Optional[mypy.types.Type]
 
         # Check each consecutive operand pair and their operator
         for left, right, operator in zip(e.operands, e.operands[1:], e.operators):
             left_type = self.accept(left)
 
-            method_type = None  # type: mypy.types.Type
+            method_type = None  # type: Optional[mypy.types.Type]
```
gvanrossum: Again (somehow).

mypy/checkexpr.py
```diff
@@ -1679,9 +1697,9 @@ def visit_index_expr_helper(self, e: IndexExpr) -> Type:
             return result
 
     def visit_tuple_slice_helper(self, left_type: TupleType, slic: SliceExpr) -> Type:
-        begin = None  # type: int
-        end = None  # type: int
-        stride = None  # type:int
+        begin = None  # type: Optional[int]
+        end = None  # type: Optional[int]
+        stride = None  # type: Optional[int]
```
gvanrossum: These three `# type` annotations seem unnecessary too...

mypy/checkexpr.py
```diff
@@ -1943,6 +1960,7 @@ def visit_dict_expr(self, e: DictExpr) -> Type:
         vtdef = TypeVarDef('VT', -2, [], self.object_type())
         kt = TypeVarType(ktdef)
         vt = TypeVarType(vtdef)
+        rv = None  # type: Optional[Type]
```
gvanrossum: Again.

mypy/checkmember.py
```diff
@@ -630,7 +630,7 @@ class B(A): pass
 
     """
     if isinstance(method, Overloaded):
-        return cast(F, Overloaded([bind_self(c, method) for c in method.items()]))
+        return cast(F, Overloaded([bind_self(c, original_type) for c in method.items()]))
```
gvanrossum: Wow, was that a bug?

mypy/myunit/__init__.py
```diff
@@ -112,7 +112,7 @@ def __init__(self, name: str, suite: 'Optional[Suite]' = None,
         self.name = name
         self.suite = suite
         self.old_cwd = None  # type: Optional[str]
-        self.tmpdir = None  # type: Optional[tempfile.TemporaryDirectory]
+        self.tmpdir = None  # type: Optional[tempfile.TemporaryDirectory[str]]
```
gvanrossum: Totally off-topic: this would be a case where if you converted it to PEP 526 style (`tmpdir: Optional[tempfile.TemporaryDirectory[str]]` at the class level) it would fail at runtime, because the real tempfile module doesn't declare TemporaryDirectory as generic -- but Łukasz's PEP 563 would save the day...

mypy_self_check.ini
```diff
@@ -8,21 +8,6 @@ disallow_any = generics, unimported
...
-[mypy-mypy.binder]
-disallow_any = unimported
-
-[mypy-mypy.checker]
-strict_optional = False
-
-[mypy-mypy.checkexpr]
-strict_optional = False
-
+; historical exception
```
gvanrossum: I prefer `#` as comment character in .ini files.

mypy/checkmember.py
elazarg: Indeed. I recall fixing it in at some point, but probably only on a local branch.

The reason it hadn't surface (AFAIK) is that it requires an overloaded method with generic self that's also has as bound a supertype of the class, without overlapping the other overload. This ish is a rare combination which is hard to get right anyway, and the checking for the "being a supertype" part is too strict in the case of overload, so I actually struggle to make an example where this is the _only_ error.

```python
from typing import TypeVar, overload

T = TypeVar('T')

class A:
    @overload
    def f(self: T) -> T: return self
    def f(self): return self

reveal_type(A().f())
```
result:
```
../tmp.py:6: error: Single overload definition, multiple required
../tmp.py:10: error: Revealed type is 'Overload(def [T] (self: T`9) -> T`9)'
```
Adding a valid overload is not obvious (to me).

mypy/myunit/__init__.py
ilevkivskyi: Yes, this is a good case for PEP 563.

mypy/checkexpr.py
```diff
@@ -703,7 +704,8 @@ def infer_arg_types_in_context2(
 
         Returns the inferred types of *actual arguments*.
         """
-        res = [None] * len(args)  # type: List[Type]
+        dummy = None  # type: Any
+        res = [dummy] * len(args)  # type: List[Type]
```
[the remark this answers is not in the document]
ilevkivskyi: I think this is a bug in _per-file_ ``--strict-optional``, @ddfisher you might be interested in looking at this. After removing the exclusion for ``semanal.py`` I have founds few other ``--strict-optional`` failures in previously clean files, I am now fixing those.

gvanrossum:
(I've restarted the AppVeyor build, we'll see how it fares.)

gvanrossum:
No, still the same failure. :-(

ethanhs:
The issue is there is a [new lxml release](https://pypi.python.org/pypi/lxml/4.0.0) and there are not Windows wheels out yet, as IIRC they are provided by a third party. For now we should pin to lxml 3.8. I'll make a PR.

ilevkivskyi:
OK, all tests pass now.

mypy/myunit/__init__.py
elazarg: This has surfaced in #5087 during self-check:
> mypy/test/data.py:232: error: "TemporaryDirectory" expects no type arguments, but 1 given

How did that pass before?

PR #3970, opened 2017-09-18
Update mypy roadmap

JukkaL:


ROADMAP.md
```diff
@@ -9,28 +9,16 @@ core team member that nobody else is working on the same thing.
 on, and everything is subject to change. Near-term plans are likely
 more accurate.
 
-## July-December 2017
+## September-December 2017
 
 - Fix remaining highest-priority TypedDict issues and make TypedDict
   an officially supported mypy feature.
```
ilevkivskyi: When it is planned to move ``TypedDict`` to ``typing``? There are still some bugs with ``TypedDict``, but it seems to be "feature complete".

gvanrossum: If you want this in Python 3.7 there should be a PEP (since it is a significant feature). But I think it's better to keep it in typing_extensions, so we can iterate on the feature more quickly. E.g. maybe we end up needing per-field flags for totality.

ilevkivskyi: > But I think it's better to keep it in typing_extensions

It is currently only in ``mypy_extensions`` (as well as extended ``Callable`` arguments). Maybe it actually makes sense to move these to ``typing_extensions``?

gvanrossum: We should probably debate that on the typing tracker. The difference between the two is that things in mypy_extensions can be unilaterally added or changed by the mypy team, and *may* forever remain mypy-specific, while things in typing_extensions need to be debated with the other projects using PEP 484 (e.g. PyCharm and pytype) and *must* be intended for eventual inclusion in typing.py.

PR #3973, opened 2017-09-20
Refactor test discovery

elazarg:
Test discovery logic is duplicated over all the different test suites. This PR refactors this logic to the plugin method `collect()`, leaving only the choice of configuration in the classes. This refactoring should enable further optimizations in the discovery of specific tests.

The argument `perform` to `parse_test_cases()` is (already) unused, so it is removed. Same goes for `include_path`.

The core of the changes can be tested by diffing the result of `py.test --collect-only -n0` with the main branch. The only differences should be the execution time and the name of the suite `PythonCmdlineSuite` which is called `PythonEvaluationSuite` on master.

`TestCase` is split into `BaseTestCase` to avoid sharing of fields and methods that's not really shared with myunit tests, such as `run()`. I believe this change need not last long.


mypy/myunit/__init__.py
```diff
@@ -105,30 +105,19 @@ def fail() -> None:
     raise AssertionFailure()
 
 
-class TestCase:
-    def __init__(self, name: str, suite: 'Optional[Suite]' = None,
-                 func: Optional[Callable[[], None]] = None) -> None:
-        self.func = func
+class ProtoTestCase:
```
JukkaL: I'd suggest renaming this to `BaseTestCase`, since `Proto` sounds like a protocol but this is a regular class. Also add a docstring. At least describe how these are run, since this class doesn't seem to have a hint about that.

mypy/myunit/__init__.py
```diff
@@ -140,6 +129,28 @@ def tear_down(self) -> None:
         self.tmpdir = None
 
 
+class TestCase(ProtoTestCase):
```
JukkaL: Add docstring. In particular, how is this different from the base class?

mypy/test/data.py
```diff
@@ -191,7 +188,7 @@ def parse_test_cases(
     return out
 
 
-class DataDrivenTestCase(TestCase):
+class DataDrivenTestCase(ProtoTestCase):
```
JukkaL: This is kind of confusing, since there seems to be no longer anything in this class about how an individual test case is run. This at least needs a docstring.

mypy/test/data.py
```diff
@@ -561,13 +549,37 @@ def pytest_addoption(parser: Any) -> None:
 def pytest_pycollect_makeitem(collector: Any, name: str, obj: Any) -> Any:
     if not isinstance(obj, type) or not issubclass(obj, DataSuite):
         return None
+    if obj is DataSuite:
```
JukkaL: Add comment here -- something like this perhaps: only classes derived from `DataSuite` contain test cases, not the `DataSuite` class.

mypy/test/data.py
```diff
@@ -561,13 +549,37 @@ def pytest_addoption(parser: Any) -> None:
...
+        return None
     return MypyDataSuite(name, parent=collector)
 
 
 class MypyDataSuite(pytest.Class):  # type: ignore  # inheriting from Any
     def collect(self) -> Iterator['MypyDataCase']:
-        for case in self.obj.cases():
-            yield MypyDataCase(case.name, self, case)
+        cls = self.obj
```
JukkaL: Add comment -- what is `self.obj` here and where does it come from? This apparently requires understanding some pytest magic, which many readers likely won't be too familiar with, including me.

mypy/test/data.py
```diff
@@ -561,13 +549,37 @@ def pytest_addoption(parser: Any) -> None:
 def pytest_pycollect_makeitem(collector: Any, name: str, obj: Any) -> Any:
     if not isinstance(obj, type) or not issubclass(obj, DataSuite):
         return None
+    if obj is DataSuite:
+        return None
     return MypyDataSuite(name, parent=collector)
 
 
 class MypyDataSuite(pytest.Class):  # type: ignore  # inheriting from Any
```
JukkaL: Could you add docstring here as well while you are at it and probably have more context on this than I have?

mypy/test/data.py
```diff
@@ -610,12 +622,15 @@ def repr_failure(self, excinfo: Any) -> str:
 
 
 class DataSuite:
+    files = None  # type: typing.ClassVar[List[str]]
```
JukkaL: I'd rather not have the `ClassVar` annotations here and elsewhere -- they add noise and distract from the main intent of the code. It would be okay to just add a comment here mentioning that all of these should be treated as class variables.

mypy/test/data.py
```diff
@@ -610,12 +622,15 @@ def repr_failure(self, excinfo: Any) -> str:
...
+    base_path = '.'  # type: typing.ClassVar[str]
+    optional_out = False  # type: typing.ClassVar[bool]
+    native_sep = False  # type: typing.ClassVar[bool]
+    require_stable = False  # type: typing.ClassVar[bool]
+    require_incremental = False  # type: typing.ClassVar[bool]
```
JukkaL: `require_stable` and `require_incremental` are too specialized to be included here as configuration attributes. It's okay to have `is_incremental` and `has_stable_flags` as utility functions in this module (I don't think that they need to be associated with any class), but I'd prefer if these would be explicitly handled using asserts in `run_case` methods, similar to how these were implemented previously.

mypy/test/testpythoneval.py
```diff
@@ -18,36 +18,25 @@
...
-        for f in python_eval_files:
-            c += parse_test_cases(os.path.join(test_data_prefix, f),
-                                  test_python_evaluation, test_temp_dir, True)
-        if sys.version_info.major == 3 and sys.version_info.minor >= 4:
-            for f in python_34_eval_files:
-                c += parse_test_cases(os.path.join(test_data_prefix, f),
-                    test_python_evaluation, test_temp_dir, True)
-        return c
+    files = ['pythoneval.test', 'python2eval.test']  # type: typing.ClassVar[List[str]]
+    if sys.version_info.major == 3 and sys.version_info.minor >= 4:
```
JukkaL: 3.3 is no longer supported for running mypy so this test can be dropped.

elazarg:
I think I've addressed the comments except the parts about configuration options. I hope to address the rest tomorrow.

mypy/test/data.py
elazarg: These are also used as filters in the test selection, and the asserts only validate it.  I will add a `filter` method then.

elazarg:
@JukkaL I believe this is ready.

mypy/myunit/__init__.py
```diff
@@ -105,30 +104,22 @@ def fail() -> None:
...
 
 
-class TestCase:
-    def __init__(self, name: str, suite: 'Optional[Suite]' = None,
-                 func: Optional[Callable[[], None]] = None) -> None:
-        self.func = func
+class BaseTestCase:
+    """Common base class for _MyUnitTestCase and DataDrivenTestCase.
+
+    Handles temporary folder creation and deletion"""
```
JukkaL: Style nit: In multi-line docstrings add the trailing `"""` on a separate line like this:

```py
class BaseTestCase:
    """Common base class for _MyUnitTestCase and DataDrivenTestCase.

    Handles temporary folder creation and deletion.
    """
    ...
```

mypy/myunit/__init__.py
```diff
@@ -140,35 +131,50 @@ def tear_down(self) -> None:
...
+    """A concrete, myunit-specific test case, a wrapper around a method to run."""
+
+    def __init__(self, name: str, suite: 'Suite', run: Callable[[], None]) -> None:
+        super().__init__(name)
+        self.run = run
+        self.suite = suite
+
+    def setup(self) -> None:
+        super().setup()
+        self.suite.set_up()
```
JukkaL: Inconsistent naming: `setup` vs `set_up`?

mypy/test/data.py
```diff
@@ -558,42 +551,73 @@ def pytest_addoption(parser: Any) -> None:
...
 # http://doc.pytest.org/en/latest/writing_plugins.html#collection-hooks
-def pytest_pycollect_makeitem(collector: Any, name: str, obj: Any) -> Any:
-    if not isinstance(obj, type) or not issubclass(obj, DataSuite):
-        return None
-    return MypyDataSuite(name, parent=collector)
+def pytest_pycollect_makeitem(collector: Any, name: str,
+                              obj: object) -> 'Optional[Any]':
+    """Called by pytest on each object in modules configured in conftest.py files.
+
+    collector is pytest.Collector, returns Optional[pytest.Class]"""
```
JukkaL: Style nit: `"""` should be on a separate line in multi-line docstrings.

mypy/test/data.py
```diff
@@ -558,42 +551,73 @@ def pytest_addoption(parser: Any) -> None:
...
+
+    if isinstance(obj, type):
+        # Only classes derived from DataSuite contain test cases, not the DataSuite class itself
+        if issubclass(obj, DataSuite) and obj is not DataSuite:
+            # Non-None result means this obj is a test case
+            # result.collect will be called, with self.obj being obj
```
JukkaL: Here it's a little unclear what `result.collect` refers to. Maybe rephrase as "The collect method of the returned MypyDataSuite instance will be called later, with self.obj being obj."

mypy/test/testcheck.py
```diff
@@ -83,14 +80,9 @@
...
 class TypeCheckSuite(DataSuite):
-
-    @classmethod
-    def cases(cls) -> List[DataDrivenTestCase]:
-        c = []  # type: List[DataDrivenTestCase]
-        for f in files:
-            c += parse_test_cases(os.path.join(test_data_prefix, f),
-                                  None, test_temp_dir, True)
-        return c
+    files = typecheck_files  # type: List[str]
```
JukkaL: All of the variable type annotations for in `DataSuite` subclasses now seem redundant. Can you remove them?

mypy/test/testdmypy.py
```diff
@@ -4,52 +4,47 @@
...
-            c += [case for case in tc
-                  if cls.has_stable_flags(case) and cls.is_incremental(case)]
-        return c
+    def filter(cls, testcase: DataDrivenTestCase) -> bool:
+        return has_stable_flags(testcase) and is_incremental(testcase)
 
     def run_case(self, testcase: DataDrivenTestCase) -> None:
-        assert self.is_incremental(testcase), "Testcase is not incremental"
-        assert self.has_stable_flags(testcase), "Testcase has varying flags"
+        assert has_stable_flags(testcase), "Testcase is not incremental"
```
JukkaL: The message should be "Testcase has varying flags" (the messages seem to have been switched on these two lines).

mypy/test/testdmypy.py
```diff
@@ -4,52 +4,47 @@
...
+        assert is_incremental(testcase), "Testcase has varying flags"
```
JukkaL: Here's the other switched message.

elazarg:
Thanks! Fixed.

Issue #3939, opened 2017-09-09
Prohibit issubclass() for runtime protocols with non-method members


ilevkivskyi:
As discussed in https://github.com/python/typing/pull/464, ``issubclass()`` with runtime protocols that declare non-method attributes is too ambiguous. Therefore it is proposed to prohibit such calls. Note that ``isinstance()`` is still fine.

Issue #3938, opened 2017-09-09
Apply the None rule only to callable protocol members.


ilevkivskyi:
Currently mypy applies the rule that things like ``__iter__ = None`` block protocol implementation. However, it is applied only to dunder attributes, which is too ad-hoc (also there are names like ``send`` and ``throw`` in ``collections.abc``). It is proposed in https://github.com/python/typing/pull/464 that it should be applied only to methods (callable members) which would be more consistent.

Issue #3906, opened 2017-09-01
None should be a subtype of an empty protocol


ilevkivskyi:
Currently this gives an incompatible type error with ``--strict-optional``, although it works with plain ``object``:
```python
class Empty(Protocol):
    pass

e: Empty
e = None  # Error

o: object
o = None # This is OK
```

gvanrossum:
Should we have a topic-protocols label?

JukkaL:
Yes. Do you want to add one?

gvanrossum:
Created the label and added some issues. Ivan may want to double-check, there may be more issues to be categorized this way.

ilevkivskyi:
Thanks! I added the label to few other issues.

PR #3996, opened 2017-09-24
Minor updates to protocol semantics

ilevkivskyi:
This updates protocol semantic according to the discussion in https://github.com/python/typing/pull/464:
* ``None`` should be considered a subtype of an empty protocol
* ``method = None`` rule should apply only to callable protocol members
* ``issublcass()`` is prohibited for protocols with non-method members.

Fixes https://github.com/python/mypy/issues/3906
Fixes https://github.com/python/mypy/issues/3938
Fixes https://github.com/python/mypy/issues/3939

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -264,31 +264,36 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                         callee_type = self.apply_method_signature_hook(
                             e, callee_type, object_type, signature_hook)
         ret_type = self.check_call_expr_with_callee_type(callee_type, e, fullname, object_type)
-        if (isinstance(e.callee, RefExpr) and len(e.args) == 2 and
-                e.callee.fullname in ('builtins.isinstance', 'builtins.issubclass')):
-            for expr in mypy.checker.flatten(e.args[1]):
-                tp = self.chk.type_map[expr]
-                if (isinstance(tp, CallableType) and tp.is_type_obj() and
-                        tp.type_object().is_protocol and
-                        not tp.type_object().runtime_protocol):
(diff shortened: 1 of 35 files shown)


gvanrossum:
@JukkaL Can you review this?

@ilevkivskyi What changed so that reveal_type() of `builtins.list` now comes out as `builtins.list[Any]`?

ilevkivskyi:
> What changed so that reveal_type() of ``builtins.list`` now comes out as ``builtins.list[Any]``?

This:
```
 def visit_type_type(self, t: TypeType) -> None:
-    pass
+    t.item.accept(self)
```
This fix will anyway happen in #3952, but without it this test (expectedly) crashed.

mypy/checkexpr.py
```diff
@@ -273,6 +273,16 @@ def visit_call_expr(self, e: CallExpr, allow_none_return: bool = False) -> Type:
                         not tp.type_object().runtime_protocol):
                     self.chk.fail('Only @runtime protocols can be used with'
                                   ' instance and class checks', e)
+        if (isinstance(e.callee, RefExpr) and len(e.args) == 2 and
+                e.callee.fullname == 'builtins.issubclass'):
+            for expr in mypy.checker.flatten(e.args[1]):
```
JukkaL: Move the body to a separate method, as this function is already quite long. Maybe also move the body of the above if statement (from line 269) to a separate method. You could also move the if statement to be nested under the above if statement (line 267). You'd only need the check `if e.callee.fullname == 'builtins.subclass'` which would simplify things slightly.

JukkaL:
Can you also fix the merge conflict?

ilevkivskyi:
Thanks! I addressed CR and fixed the conflict.

PR #4000, opened 2017-09-24
Allow subscripted aliases at function scope

### LINKED ISSUE #3145: Qualified alias names are mishandled
```
T = TypeVar('T')

class X(Generic[T]):
    A = X[T]
    def f(self, x: int = 1) -> X[T]: ...
    a: X[T]
    b: A = a # ok
    def g(self) -> None:
        a: X[T]
        b: X.A = a # ok


def g(x: X[T]) -> None:
    p: X[T] = x.f() # ok
    q: X.A = x.f() # fail
```

I think it should pass, unless qualified aliases are not allowed (in which case it should fail in `X.g()`). In addition, the error is confusing: `Incompatible types in assignment (expression has type m.X[T], variable has type m.X[T])`.

ilevkivskyi:
Fixes #3145 

This PR allows subscripted type aliases at function scope, as discussed in #3145 this will simplify the rules and have other pluses (apart from fixing the actual issue).

In addition, I prohibit reuse of _bound_ type variables to define generic aliases, situations like this were always ambiguous:
```python
T = TypeVar('T')
class C(Generic[T]):
    A = List[T]
    x: A
    reveal_type(x)  # on one hand one might expect this to be List[T],
                    # but on the other hand, non-subscripted generic aliases
                    # get variables replaced with 'Any', so this should be 'List[Any]'.
```
Also prohibiting this use is consistent with how we prohibit bound type variable reuse for nested classes, one can always just use another type variable, so this works:
```python
T = TypeVar('T')
S = TypeVar('S')
class C(Generic[T]):
    A = List[S]
    x: A[T]
```

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -635,7 +635,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             res = type_object_type(item.type, self.named_type)
             if isinstance(res, CallableType):
                 res = res.copy_modified(from_type_type=True)
-            return res
+            return expand_type_by_instance(res, item)
         if isinstance(item, UnionType):
             return UnionType([self.analyze_type_type_callee(item, context)
                               for item in item.relevant_items()], item.line)
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1710,11 +1710,10 @@ class SemanticAnalyzer(NodeVisitor[None]):
         return Instance(fb_info, [])
 
     def analyze_alias(self, rvalue: Expression,
-                      allow_unnormalized: bool) -> Tuple[Optional[Type], List[str]]:
+                      warn_bound_tvar: bool = False) -> Tuple[Optional[Type], List[str]]:
         """Check if 'rvalue' represents a valid type allowed for aliasing
         (e.g. not a type variable). If yes, return the corresponding type and a list of
         qualified type variable names for generic aliases.
-        If 'allow_unnormalized' is True, allow types like builtins.list[T].
         """
         res = analyze_type_alias(rvalue,
                                  self.lookup_qualified,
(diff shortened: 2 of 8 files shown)


mypy/semanal.py
```diff
@@ -1732,11 +1732,10 @@ def alias_fallback(self, tp: Type) -> Instance:
         return Instance(fb_info, [])
 
     def analyze_alias(self, rvalue: Expression,
-                      allow_unnormalized: bool) -> Tuple[Optional[Type], List[str]]:
+                      warn_bound_tvar: bool = False) -> Tuple[Optional[Type], List[str]]:
         """Check if 'rvalue' represents a valid type allowed for aliasing
         (e.g. not a type variable). If yes, return the corresponding type and a list of
         qualified type variable names for generic aliases.
-        If 'allow_unnormalized' is True, allow types like builtins.list[T].
```
JukkaL: Does this PR also affect whether `list[x]` is accepted in stubs?

mypy/checkexpr.py
```diff
@@ -650,7 +650,7 @@ def analyze_type_type_callee(self, item: Type, context: Context) -> Type:
             res = type_object_type(item.type, self.named_type)
             if isinstance(res, CallableType):
                 res = res.copy_modified(from_type_type=True)
-            return res
+            return expand_type_by_instance(res, item)
```
JukkaL: What's this change?

mypy/semanal.py
```diff
@@ -1770,11 +1770,12 @@ def check_and_set_up_type_alias(self, s: AssignmentStmt) -> None:
         lvalue = s.lvalues[0]
         if not isinstance(lvalue, NameExpr):
             return
-        if (len(s.lvalues) == 1 and not self.is_func_scope() and
-                not (self.type and isinstance(s.rvalue, NameExpr) and lvalue.is_def)
+        if (len(s.lvalues) == 1 and
+                not ((self.type or self.is_func_scope()) and
```
JukkaL: This condition is hard to understand. Add a comment describing what it does at a high level and maybe also give some concrete examples. It's also worth trying to simplify the nested boolean expression. Ideas:

* Make the top-level expression of form `e1 and (...) and not e2` instead of `e1 and not (...) and not e2` -- the first `not` operator makes this a bit harder to understand, I think.
* Use a temp variable to hold the result of `self.type or self.is_func_scope()` and give it a semantically meaningful name.

PR #4002, opened 2017-09-25
README.md: rearrange first paragraph to be more friendly

elazarg:
The discussion of different annotations need not be the first thing the reader see.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,8 +168,7 @@ In Windows, the script is generally installed in
     C:\>\Python34\python \Python34\Scripts\mypy PROGRAM
 
 If you are on Windows using Python 3.3 or 3.4, and would like to use XML reports
-download LXML from [PyPi](https://pypi.python.org/pypi/lxml). 
-XML based reports do not work on Python 3.3 and 3.4 on Windows.
+download LXML from [PyPi](https://pypi.python.org/pypi/lxml).
 
 ### Working with `virtualenv`
 
--- a/README.md
+++ b/README.md
@@ -47,9 +47,9 @@ See [the documentation](http://mypy.readthedocs.io/en/stable/introduction.html) 
 
 For Python 2.7, the standard annotations are written as comments:
 ```python
-def reverse(s):
-    # type: str -> str
-    return s[::-1]
+def is_palindrome(s):
+    # type: (str) -> str
+    return s == s[::-1]
 ```
 
 See [the documentation for Python 2 support](http://mypy.readthedocs.io/en/latest/python2.html).
--- a/README.md
+++ b/README.md
@@ -25,17 +25,8 @@ What is mypy?
 -------------
 
 Mypy is an optional static type checker for Python.  You can add type
-hints to your Python programs using the standard for type
-annotations introduced in Python 3.5 ([PEP 484](https://www.python.org/dev/peps/pep-0484/)), and use mypy to
-type check them statically. Find bugs in your programs without even
-running them!
-
-The type annotation standard has also been backported to earlier
-Python 3.x versions.  Mypy supports Python 3.3 and later.
-XML based reports do not work on Python 3.3 and 3.4 on Windows.
-
-For Python 2.7, you can add annotations as comments (this is also
-specified in [PEP 484](https://www.python.org/dev/peps/pep-0484/)).
+hints to your Python programs, and use mypy to type check them statically.
+Find bugs in your programs without even running them!
 
 You can mix dynamic and static typing in your programs. You can always
 fall back to dynamic typing when static typing is not convenient, such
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,6 +47,17 @@ def fib(n: int) -> Iterator[int]:
 Mypy is in development; some features are missing and there are bugs.
 See 'Development status' below.
 
+Annotations
```
obestwalter: Good move 🚜

gvanrossum:
Can you move the bullet about the xml reporter elsewhere? It's a distraction at this point. I'd also fold the discussion of supported Python 3 versions into one sentence.

elazarg:
I tried something slightly different - demonstration for python 2 too. I'm not sure this is a good idea, but I couldn't find the right way to phrase it.

README.md
```diff
@@ -160,7 +168,8 @@ In Windows, the script is generally installed in
     C:\>\Python34\python \Python34\Scripts\mypy PROGRAM
 
 If you are on Windows using Python 3.3 or 3.4, and would like to use XML reports
-download LXML from [PyPi](https://pypi.python.org/pypi/lxml).
+download LXML from [PyPi](https://pypi.python.org/pypi/lxml). 
+XML based reports do not work on Python 3.3 and 3.4 on Windows.
```
ethanhs: They do work, but only if you have LXML installed. The last sentence can be deleted.

elazarg:
So should I just remove the annotations? I think it's valuable to be aware of the way to annotate variables, since most people will not read the PEP. If the example will return a list we'll have both:
```python
from typing import List 

def fib(n: int) -> List[int]:
    a, b = 0, 1
    result: List[int] = []
    while a < n:
        result.append(a)
        a, b = b, a + b
    return result
```

ilevkivskyi:
@elazarg 
>  I think it's valuable to be aware of the way to annotate variables, since most people will not read the PEP.

I am not sure here, people will most likely not read read the PEP, but we should make them read the docs (at least to some extent).

gvanrossum:
The readme is not the place for this. It should point people to the real
docs, in particular an introduction, right after the example.

README.md
```diff
@@ -168,8 +168,7 @@ In Windows, the script is generally installed in
     C:\>\Python34\python \Python34\Scripts\mypy PROGRAM
 
 If you are on Windows using Python 3.3 or 3.4, and would like to use XML reports
```
ethanhs: If #4007 gets merged before this, this sentence can be removed, as there are wheels for all Python versions on Windows now, and installing via test-requirements will "just work".

elazarg:
Is this OK now? I tried to make it shorter so the first part will fit in a laptop screen.

README.md
```diff
@@ -52,11 +43,20 @@ def fib(n: int) -> Iterator[int]:
         yield a
         a, b = b, a + b
 ```
+See [the documentation](http://mypy.readthedocs.io/en/stable/introduction.html) for more examples.
+
+For Python 2.7, the standard annotations are written as comments:
+```python
+def reverse(s):
+    # type: str -> str
```
ilevkivskyi: This should be ``# type: (str) -> str`` (note parentheses), otherwise it is ``syntax error in type comment``.

README.md
```diff
@@ -52,11 +43,19 @@ def fib(n: int) -> Iterator[int]:
...
 ```
+This example uses Python 3 style of the standard type annotations, defined in [PEP 484](https://www.python.org/dev/peps/pep-0484/). 
+
+For Python 2.7, the standard annotations are written as comments:
+```python
+def reverse(s):  # type: str -> str
+    return s[::-1]
+```
+
+See [the documentation](http://mypy.readthedocs.io/en/stable/introduction.html) for more examples.
```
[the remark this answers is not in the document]
elazarg: I'm not sure how to keep it fluent, while staying clear and concise. I gave it another try.

Maybe there can be some explanation about PEP-484 in a dedicated section.

README.md
```diff
@@ -25,23 +25,14 @@ What is mypy?
...
-type check them statically. Find bugs in your programs without even
-running them!
-
-The type annotation standard has also been backported to earlier
-Python 3.x versions.  Mypy supports Python 3.3 and later.
-XML based reports do not work on Python 3.3 and 3.4 on Windows.
-
-For Python 2.7, you can add annotations as comments (this is also
-specified in [PEP 484](https://www.python.org/dev/peps/pep-0484/)).
+hints ([PEP 484](https://www.python.org/dev/peps/pep-0484/)) to your Python programs, and use mypy to type check them statically.
```
ilevkivskyi: Not super-important, but could you please make lines in your diff 79 chars?

README.md
```diff
@@ -52,11 +43,20 @@ def fib(n: int) -> Iterator[int]:
         yield a
         a, b = b, a + b
 ```
+See [the documentation](http://mypy.readthedocs.io/en/stable/introduction.html) for more examples.
+
+For Python 2.7, the standard annotations are written as comments:
+```python
+def reverse(s):
+    # type: str -> str
+    return s[::-1]
```
ilevkivskyi: Maybe use a slightly less trivial example? Like ``is_palindrome(s)`` that returns ``s == s[::-1]``?

elazarg: Nice! :)
I tried to find something like that and all I came up with was reverse :(

PR #4005, opened 2017-09-25
Support user install on Windows.

### LINKED ISSUE #3988: typeshed not being found on windows user install
```powershell
PS D:\testing> git clone https://github.com/python/mypy.git
# no errors

PS D:\testing> cd .\mypy\

PS D:\testing\mypy> git submodule update --init typeshed
# worked, checked out '55dbb967ad006d5835590c53ab023774dc5a1341'

PS D:\testing\mypy> pip install --user --upgrade .
Processing D:\testing\mypy
Requirement already up-to-date: typed-ast<1.2.0,>=1.1.0 in c:\users\[redacted!]\appdata\roaming\python\python36\site-packages (from mypy===0.530-dev-2bd11f60f0a0332b4c9537983a6ac7da4416e9ca)
Installing collected packages: mypy
  Running setup.py install for mypy ... done
Successfully installed mypy-0.530-dev-2bd11f60f0a0332b4c9537983a6ac7da4416e9ca

PS D:\testing\mypy> mypy setup.py  # to test
Could not resolve typeshed subdirectories. If you are using MyPyfrom source, you need to run "git submodule --init update".Otherwise your MyPy install is broken.
```

The data directory (after searching for `typeshed`) appears to be `%AppData%\Python\lib\mypy`

full path: `C:\Users\[redacted!]\AppData\Roaming\Python\lib\mypy`

```powershell
PS D:\testing\mypy> ls $Env:AppData\Python\lib\mypy

typeshed
xml
```

using this directory leads to more expected results

```powershell
PS D:\testing\mypy> mypy --custom-typeshed-dir $Env:AppData\Python\lib\mypy\typeshed setup.py
setup.py:15: error: No library stub file for module 'setuptools'
setup.py:15: note: (Stub files are from https://github.com/python/typeshed)
setup.py:16: error: No library stub file for module 'setuptools.command.build_py'
setup.py:74: error: Need type annotation for variable
```

ethanhs:
On Windows, the usual install paths are `sys.prefix/Lib/mypy` for a global
install. For user installs, the path is `site.getuserbase()/lib/mypy`. We
also fall back to the old method of installation due to the data dir being
put in the package directory on an egg install.

This restores the usual data directory resolution on other platforms.

Correctly fixes #3988, and keeps other platforms working.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -17,6 +17,7 @@ import hashlib
 import json
 import os.path
 import re
+import site
 import sys
 import time
 from os.path import dirname, basename
@@ -213,6 +214,11 @@ def default_data_dir(bin_dir: Optional[str]) -> str:
       bin_dir: directory containing the mypy script
     """
     if not bin_dir:
+        if os.name == 'nt':
+            for parent in map(os.path.join, (sys.prefix, 'Lib'), (site.getuserbase(), 'lib')):
+                    data_dir = os.path.join(parent, 'mypy')
+                    if os.path.exists(data_dir):
+                        return data_dir
         mypy_package = os.path.dirname(__file__)
         parent = os.path.dirname(mypy_package)
         if (os.path.basename(parent) == 'site-packages' or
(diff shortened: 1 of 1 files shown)


ethanhs:
Removed. Mypy wasn't happy with the map call either.

ethanhs:
The Appveyor failure was the usual #3895

Summertime:
Seems like it fixed the issue I was having, works for me!

davt:
Be aware that this change as shipped in 0.530 (installed from pypi) breaks when run under virtualenv on Windows.

The underlying reason is that - for reasons I don't understand - virtualenv overwrites Lib/site.py with an old version which does not include the function site.getuserbase().

Would a direct call to sysconfig.get_config_var('userbase') be a suitable workaround?

JukkaL:
@davt Can you file a new issue for the virtualenv-on-Windows problem?

PR #4006, opened 2017-09-25
Run mypy self-check only once

ethanhs:
Since we only need to test that mypy is type safe once, we skip the
mypy package check except on one test run. This is arbitrarily chosen as
3.5.1. 3.6 is not chosen as it already is running flake8.

Additionally, the old comment about pypy is removed as typed_ast doesn't run on pypy, so the comment is confusing and misleading.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,6 @@ python:
   - "3.5.1"
   - "3.6"
   - "3.7-dev"
-  # Pypy build is disabled because it doubles the travis build time, and it rarely fails
-  # unless one one of the other builds fails.
-  # - "pypy3"
 
 install:
   - pip install -U pip setuptools wheel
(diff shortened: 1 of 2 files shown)


ethanhs:
Okay this is ready for review. The appveyor failure is the usual #3895.

ethanhs:
Also travis timed out.

gvanrossum:
Well since this PR is specifically about Travis CI and AppVeyor it would be nice if we got the tests to pass at least once... I've restarted the Travis CI 3.5.1 job (the others passed), but I can't seem to find a "restart job" button on AppVeyor.

ethanhs:
I believe you need to go to "New Build" under the drop down and choose the branch. I believe the PR branch should show up. If not I can try to trigger it again.

gvanrossum:
Actually I think the menu item to use is "re-build". I'm not sure why it didn't appear for me before, possibly I switched accounts sloppily. I've now restarted it, hopefully it'll work.

gvanrossum:
The re-build either didn't have any effect or failed in the same way. And now I don't see the re-build button any more (nor new build). I'll just wait until tomorrow.

gvanrossum:
(Hm, I wonder if AppVeyor only lets me rebuild when there are no other builds running?)

PR #4007, opened 2017-09-25
Update LXML dependency to 4.0

ethanhs:
LXML 4.0.0 has wheels for all platforms/Python versions we support. It is faster and has several bug fixes.


gvanrossum:
Can you restart AppVeyor?

gvanrossum:
Seems this worked. I'll merge it now.

PR #3227, opened 2017-04-23
Fix handling of non-method callable attribute

elazarg:
Rewrite `check_method_type`:
<br />1. Reverse the subtyping check. It seems to be backwards. This change alone fixes #3223.
<br />2. Pass original_type for the subtype check, which is more precise - in particular, there seems no need to fallback from TupleType. It also yields better error messages.
<br />3. Make the error message clearer (I think. see the tests)
<br />4. The classmethod case goes backwards too - instead of falling back in the parameter, fall back in the argument. I think it's the straightforward thing.
<br />
<br />The PR does not pass stub tests due to [the following error](https://travis-ci.org/python/mypy/jobs/224949181#L1434):
<br />> typeshed/stdlib/2and3/fractions.pyi:80: error: Invalid self argument "Union[int, float, Decimal, Real]" to attribute function Callable[[Real, Any], bool]
<br />
<br />For this line:
<br />```python
<br />    def __gt__(self, other: _ComparableNum) -> bool: ...
<br />```
<br />And similarly for `__ge__`. I did not figure out yet why it happens.

PR #4016, opened 2017-09-26
Fix checking of self when accessing a non-method callable attribute (take 2)

### LINKED ISSUE #3223: self type on property creates a spurious error
Running mypy master on this file:
```python
from typing import TypeVar, List
_T = TypeVar('_T')

class X:
    @property
    def __members__(self: _T) -> List[_T]: ...

reveal_type(X().__members__)
```
gives
```
../bin/selftypeproperty.py:8: error: Revealed type is 'builtins.list[selftypeproperty.X*]'
../bin/selftypeproperty.py:8: error: Invalid method type
```
The revealed type is correct but the "Invalid method type" doesn't make sense.

elazarg:
Fix #3223. Minimal examples:
```python
from typing import Callable
class X:
    f: Callable[[object], None]
X().f  # E: Invalid method type
```
```python
class X:
    @property
    def g(self: object): pass
X().g  # E: Invalid method type
```
The problem is that for non-methods the check performed on access to `self` is whether `object <: X` instead of the other way around.

1. Reverse subtyping check. This change alone fixes the issue described above.
2. For classmethod, fallback on the argument instead of on the parameter
3. Mimic dispatch better: meet `original_type` with static class. This handles the case were instead of `X()` above there was something of type `Union[X, Y]`.
4. Better error message

Take 2 of #3227, but without the reverse-operator part

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -20,6 +20,8 @@ from mypy.typevars import fill_typevars
 from mypy.plugin import Plugin, AttributeContext
 from mypy import messages
 from mypy import subtypes
+from mypy import meet
+
 MYPY = False
 if MYPY:  # import for forward declaration only
     import mypy.checker
@@ -314,7 +316,9 @@ def analyze_var(name: str, var: Var, itype: Instance, info: TypeInfo, node: Cont
                 # methods: the former to the instance, the latter to the
                 # class.
                 functype = t
-                check_method_type(functype, itype, var.is_classmethod, node, msg)
+                # Use meet to simulate dispatch - e.g. reduce Union[A, B] to A on dispatch to A
+                dispatched_type = meet.meet_types(original_type, itype)
+                check_self_arg(functype, dispatched_type, var.is_classmethod, node, name, msg)
                 signature = bind_self(functype, original_type, var.is_classmethod)
                 if var.is_property:
                     # A property cannot have an overloaded type => the cast
@@ -370,33 +374,22 @@ def lookup_member_var_or_accessor(info: TypeInfo, name: str,
         return None
 
 
-def check_method_type(functype: FunctionLike, itype: Instance, is_classmethod: bool,
-                      context: Context, msg: MessageBuilder) -> None:
+def check_self_arg(functype: FunctionLike, original_type: Type, is_classmethod: bool,
+                   context: Context, name: str, msg: MessageBuilder) -> None:
+    """Check that the the most precise type of the self argument is compatible
+    with the declared type of each of the overloads.
+    """
     for item in functype.items():
         if not item.arg_types or item.arg_kinds[0] not in (ARG_POS, ARG_STAR):
             # No positional first (self) argument (*args is okay).
-            msg.invalid_method_type(item, context)
-        elif not is_classmethod:
-            # Check that self argument has type 'Any' or valid instance type.
-            selfarg = item.arg_types[0]
-            # If this is a method of a tuple class, correct for the fact that
-            # we passed to typ.fallback in analyze_member_access. See #1432.
-            if isinstance(selfarg, TupleType):
-                selfarg = selfarg.fallback
-            if not subtypes.is_subtype(selfarg, itype):
-                msg.invalid_method_type(item, context)
+            msg.fail('Attribute function with type %s does not accept self argument'
+                     % msg.format(item), context)
         else:
-            # Check that cls argument has type 'Any' or valid class type.
-            # (This is sufficient for the current treatment of @classmethod,
-            # but probably needs to be revisited when we implement Type[C]
-            # or advanced variants of it like Type[<args>, C].)
-            clsarg = item.arg_types[0]
-            if isinstance(clsarg, CallableType) and clsarg.is_type_obj():
-                if not subtypes.is_equivalent(clsarg.ret_type, itype):
-                    msg.invalid_class_method_type(item, context)
-            else:
-                if not subtypes.is_equivalent(clsarg, AnyType(TypeOfAny.special_form)):
-                    msg.invalid_class_method_type(item, context)
+            selfarg = item.arg_types[0]
+            if is_classmethod:
+                original_type = TypeType.make_normalized(original_type)
+            if not subtypes.is_subtype(original_type, erase_to_bound(selfarg)):
+                msg.invalid_method_type(name, original_type, item, is_classmethod, context)
 
 
 def analyze_class_attribute_access(itype: Instance,
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -288,6 +288,7 @@ def analyze_var(name: str, var: Var, itype: Instance, info: TypeInfo, node: Cont
 
     This is conceptually part of analyze_member_access and the arguments are similar.
 
+    itype is the class object in which var is dedined
     original_type is the type of E in the expression E.var
     """
     # Found a member variable.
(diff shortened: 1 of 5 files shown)


mypy/checkmember.py
```diff
@@ -370,33 +374,22 @@ def lookup_member_var_or_accessor(info: TypeInfo, name: str,
         return None
 
 
-def check_method_type(functype: FunctionLike, itype: Instance, is_classmethod: bool,
-                      context: Context, msg: MessageBuilder) -> None:
+def check_self_arg(functype: FunctionLike, original_type: Type, is_classmethod: bool,
+                   context: Context, name: str, msg: MessageBuilder) -> None:
+    """Check that the the most precise type of the self argument is compatible
```
JukkaL: Style nit: in a multi-line docstring there should be an empty line after the first line.

elazarg: I'll fix it on the next commit or before merge.

elazarg:
Updated the PR description and the title. I hope it's an improvement.

gvanrossum:
Sorry, this is too late to make it into the 0.530 release. It's a minor issue and it touches subtle parts of the codebase. Let's postpone it.

mypy/checkmember.py
```diff
@@ -314,7 +316,9 @@ def analyze_var(name: str, var: Var, itype: Instance, info: TypeInfo, node: Cont
                 # methods: the former to the instance, the latter to the
                 # class.
                 functype = t
-                check_method_type(functype, itype, var.is_classmethod, node, msg)
+                # Use meet to simulate dispatch - e.g. reduce Union[A, B] to A on dispatch to A
```
JukkaL: The comment is unclear. Can you add a comment with an example where this makes a difference?

mypy/checkmember.py
```diff
@@ -370,33 +374,22 @@ def lookup_member_var_or_accessor(info: TypeInfo, name: str,
...
-        elif not is_classmethod:
-            # Check that self argument has type 'Any' or valid instance type.
-            selfarg = item.arg_types[0]
-            # If this is a method of a tuple class, correct for the fact that
-            # we passed to typ.fallback in analyze_member_access. See #1432.
-            if isinstance(selfarg, TupleType):
-                selfarg = selfarg.fallback
-            if not subtypes.is_subtype(selfarg, itype):
-                msg.invalid_method_type(item, context)
+            msg.fail('Attribute function with type %s does not accept self argument'
```
JukkaL: Add a method to `Messages` instead of building the message here.

test-data/unit/check-selftype.test
```diff
@@ -354,16 +354,21 @@ class E:
...
+    def __members__(self: Q) -> Q: return self
+
+class A(X):
+    @property
+    def member(self: T) -> T: return self
 
 class B(A):
     pass
 
+reveal_type(X().__members__)  # E: Revealed type is '__main__.X*'
```
JukkaL: Would it make sense to add a test case similar to this, but so that `X` would a subtype of `Tuple`?

mypy/checkmember.py
JukkaL: The docstring could also be clarified a bit. The meaning of "the most precise type of the self argument" wasn't clear initially. Maybe something about generic self, and using the actual type of the self (based on the receiver type) rather than the declared type in the function signature? Would also be good to describe `original_type` here -- it would even make sense to rename it.  What about `actual_self_type`? It's always the type of self, and it's the actual instead of the formal/declared type. If you rename it, it would also make sense to rename it elsewhere.

elazarg: There are 4 potentially distinct types in play here:
1. The declared type of the first parameter `self` (potentially generic)
2. The inferred/declared type of the receiver, ignoring the function signature
3. The same type as (2) but after narrowing it using the current class
4. The type of the current class
(4) is almost irrelevant at this point, but it is used to construct (3), restricts the valid types for (2), and adds to the potential confusion.
So both (2) and (3) are "actual self type". Maybe I should elaborate on the comment, but it will become lengthy.

mypy/checkmember.py
elazarg: I added a comment and a test. This kind of dispatch with not-exactly-the-right-type occur in reverse operators.

I believe the same thing should also be done with plain functions, but it's not checked at all now. i.e. there should be a call to `check_self_arg` before `bind_self` [here](https://github.com/python/mypy/blob/master/mypy/checkmember.py#L93). For example:
```python
class A:
    def f(self: B) -> None: pass  # type: ignore - because reasons. probably useful for metaclasses

class B(A): pass
A().f()  # should be an error
```
but changing _plain functions_ breaks some typeshed code (specifically, AbstractSet operators and the `version_info` tests break).

elazarg:
@JukkaL I will only be available for a couple of hours from now, so I won't be able to respond to further comments in time, unless ~30 hours from now will be good enough. As I understand it, the behavior is OK, and I hope it will be possible to resolve further issues in docstrings and comments even after the branch is cut.

test-data/unit/check-selftype.test
```diff
@@ -351,21 +351,99 @@ class E:
...
+    @property
+    def f(self: B) -> int: pass
+x: Union[A, B]
+reveal_type(x.f)  # E: Revealed type is 'builtins.int'
+
+[builtins fixtures/property.pyi]
 
+[case testSelfTypeProperSupertypeAttribute]
+from typing import Callable, TypeVar
+T = TypeVar('T', bound=object)
```
JukkaL: `bound=object` is redundant -- this is equivalent to `T = TypeVar('T')`. It would be nice to have a test case with a non-`object` bound.

test-data/unit/check-selftype.test
```diff
@@ -351,21 +351,99 @@ class E:
...
+class B(A):
+    pass
+
+class X(metaclass=B): pass
+class Y(X): pass
+X1: Type[X]
+reveal_type(X.g)  # E: Revealed type is 'builtins.int'
+reveal_type(X.gt)  # E: Revealed type is 'def () -> __main__.X'
+reveal_type(X.f())  # E: Revealed type is 'builtins.int'
+reveal_type(X.ft())  # E: Revealed type is 'def () -> __main__.X'
```
JukkaL: Can you add a `__init__` to `X` that takes an argument so that it's obvious that the signature is derived from the constructor?

test-data/unit/check-selftype.test
elazarg: I know it's redundant, I just wanted to be clear about the intention of the test. But yes, I should add a non-object bound

test-data/unit/check-selftype.test
```diff
@@ -351,21 +351,130 @@ class E:
...
+    def gt(self: K[T]) -> T: return self.q
+    f: Callable[[object], int]
+    ft: Callable[[T], T]
+
+class B(A[Q]):
+    pass
+a: A[int]
+b: B[str]
+reveal_type(a.g)  # E: Revealed type is 'builtins.int'
+--reveal_type(a.gt)  # E: Revealed type is 'builtins.int'
```
JukkaL: Why are these commented out?

elazarg: The instantiation of selftype fails in the commented tests. I think it is not a problem in this PR, and fixing these requires changes in `bind_self`.

I just forced an update that removed two of the comments; I hope it doesn't cause a mess.

elazarg: Sorry - I didn't force it; I will add as a new commit in a moment.

elazarg: To elaborate, `bind_self` does not unify `K[T]` and `A[T]` even if it can. It just does not check for this case, and simply trims the first argument. It only performs such unification is for TypeType.

PR #4022, opened 2017-09-27
Restore the original logic on leaving partial types

ilevkivskyi:
This fixes a recently discovered crash introduced by #3995 in some internal codebases.


gvanrossum:
I'll just ignore the AppVeyor test failure -- it seems AppVeyor fails every time now.

gvanrossum:
Oh never mind it passed. Will wait for Travis CI test.

ilevkivskyi:
> I'll just ignore the AppVeyor test failure -- it seems AppVeyor fails every time now.

Actually it is green now. Sometimes it is red for several seconds, but then it maybe retries failed tests or something and then turns green.

PR #4023, opened 2017-09-27
Correct suggested command to checkout typeshed

ethanhs:
`git` will not allow `git submodule --init update`, only `git submodule update --init`.


ethanhs:
Appveyor failure is a flake and can safely be ignored (no code is changed).

Issue #1668, opened 2016-06-07
Speed up "pythoneval" tests


gnprice:
The one slowest test task in our test suite is `testpythoneval.py` aka `run eval-test` -- in fact, on my desktop with 8-way parallelism, the whole suite takes 1m28s, and if that one task is omitted it takes 51s. This is because for much of that time it's the single straggler task running, even though since commit 56bb4babb3 we start it at the very beginning. (There are also some quick wins we can get to make the suite much faster than 51s, once we make this task much faster and/or break it into smaller pieces.)

Looking inside the test code, it looks like it does two things for each of a number of test cases:
1. Run the type-checker on the test program
2. Run the test program
and then compares the output to an expected output.

Is step 2 still useful? It made a lot of sense when mypy was its own implementation of a Python-like language, but now it seems like it just serves as a regression check on CPython. In principle it could be useful to check that our tests run the way we think they do, but that's basically a test of those test cases themselves. I'm not sure how often that's useful; I think most of the test cases themselves are very straightforward. So I suggest we cut it out.

gvanrossum:
I don't think step 2 is useful. We should however keep step 1, since these are the only tests that don't use the fake builtins from mypy/test/data/fixtures/. Unfortunately this means somewhat painful surgery on the test case data files, since the checker output and the CPython output are just combined in the [out] sections. Probably you can run myunit with the -u or -i flag to make it generate new versions of the test case data files.

JukkaL:
Step 2 isn't essential any more. (I'd still consider it at least slightly useful as long as PEP 484 is provisional and the `typing` module is still evolving. However, it might be better to just move anything that seems valuable to the the `typing` module test suite.)

The biggest bottleneck is processing the stubs for builtins and typing for every test case. If we'd use incremental type checking to cache these stubs the test cases would be much faster. We could have a setup method that populates the cache. (This has been discussed elsewhere.)

Beyond that and removing step 2, we have at least these options for speeding up things:

* Don't run type checking in a subprocess but just call the `build` function directly.
* Split slowest tasks into multiple subtasks that are scheduled independently. Once the "evaluation" tests are faster, the type checker tests might even be the next long pole, so we may want to split them.
* Cache some of the fake stubs used in other tests.
* Try to reuse subprocesses to avoid some startup overhead. In particular, if we split tasks then running only a small subset of tests may become slower, since we pay the startup overhead multiple times.

ddfisher:
[...]

PR #4025, opened 2017-09-27
Move pythoneval to use mypy.api.run

### LINKED ISSUE #1671: Cut out evaluation in the "pythoneval" tests
The "pythoneval" tests are by far the biggest contributor to the time it takes to run our test suite. For each of 171 sample programs, we type-check them and then run them through the Python interpreter.

From discussion in #1668, running the programs is no longer important for our testing. So cut it out. This will mean
- cutting out that step in `testpythoneval.py`,
- cutting out the program output (as opposed to the type-checker output) from all the test cases, and
- cutting out entirely any cases that no longer seem to have any meaningful content when they're only for the type-checker.

Some of the test cases may exercise features of `typing` in interesting ways -- any of those should be moved to the `typing` repo's own test suite if they aren't covered there already.

After this is done, we can probably get a big further speedup by folding the remaining test cases into just more cases for `testcheck.py`, so that we're not forking a mypy subprocess for each one. Will make a separate issue for that.

ethanhs:
This replaces the old subprocess based method. In addition, the test cases
are no longer run, in order to reduce test time. This has led to a 10-20%
speedup in pytest. This might help with #3895, as it removes the
subprocesses.

Fixes #1671

Two follow up ideas:
* Move `testcmdline` to use `mypy.api.run` as well
* Add support for StringIO inputs to mypy and use it so we don't have to write out to files all the time.

### DIFF
--- a/mypy/test/testpythoneval.py
+++ b/mypy/test/testpythoneval.py
@@ -10,12 +10,9 @@ Note: These test cases are *not* included in the main test suite, as including
       this suite would slow down the main suite too much.
 """
 
-from contextlib import contextmanager
-import errno
 import os
 import os.path
 import re
-import subprocess
 import sys
 
 import pytest  # type: ignore  # no pytest in typeshed
@@ -25,6 +22,7 @@ from mypy.test.config import test_data_prefix, test_temp_dir
 from mypy.test.data import DataDrivenTestCase, parse_test_cases, DataSuite
 from mypy.test.helpers import assert_string_arrays_equal
 from mypy.util import try_find_python2_interpreter
+from mypy.api import run
 
 # Files which contain test case descriptions.
 python_eval_files = ['pythoneval.test',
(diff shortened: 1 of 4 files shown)


elazarg:
This PR looks like a good idea regardless of performance and flakes, since it makes the test more cohesive.

StringIO sounds very good! In addition to potential speedup of test execution, it will also speed up test discovery but an order of magnitude (I intended to do that by deferring the act of writing the files). I'm interested to know what are the potential downsides; I can't see any.

Regarding `testcmdline`, I don't think it's a good idea not to test the interface between the OS and the program. Instead it might be possible to split the tests - test that passed flags correspond to the internal flags (using subprocess), and test that internal flags have the desired effect (using API).
But maybe I'm overcomplicating things.

elazarg:
Do the tests re-analyze typeshed each time?

ethanhs:
>I don't think it's a good idea not to test the interface between the OS and the program

That is why I decided to not include that in this PR. I do like the idea of splitting them up.

StringIO would require a bit of work on the mypy internals, but I think it is worth it.

As for typeshed, yes, I believe they do re-analyze each run, but an optimization there would likely be complicated.

elazarg:
But incremental mode might help? (I think caching mypy's state after analysis might be very useful, but that's slightly OT)

ethanhs:
Incremental may help, if you want to continue this discussion the best place is probably #1668.

ethanhs:
This change has the additional benefit of making us completely independent of a runtime Python 2 interpreter, fixing #4039.

gvanrossum:
> In addition, the test cases are no longer run, in order to reduce test time.

While I appreciate the speedup, I worry that over time these test cases might stop working when executed. There are still quite a few places where mypy passes code that doesn't actually work at runtime. Every time a test case in this category is added or changed we would risk that we'd be testing something imaginary. This is particularly worrisome for tests that are trying to ascertain that mypy matches subtle runtime behavior.

ethanhs:
Okay, that makes sense. I believe several of the tests can be removed anyway (they test extremely basic things), so that will cause some speed increases.

gvanrossum:
So do you want to reuse this PR for the new approach, or open a new one?

ethanhs:
I will open a new one, I think it will be cleaner.

Issue #3999, opened 2017-09-24
No documentation for metaclass support


elazarg:
Metaclasses are only mentioned in passing in the docs:
* [Once](http://mypy.readthedocs.io/en/latest/generics.html#generic-class-internals) in the context of self-type (an orthogonal feature)
* [Once](http://mypy.readthedocs.io/en/latest/class_basics.html#abstract-base-classes-and-multiple-inheritance) where the use of `ABCMeta` is recommended - which IMHO is not the right way to go anyway

I'm actually not sure what there is to say about metaclasses, but it seems reasonable to dedicate to it a small section, accessible in the main page. I think it should also mention briefly what's _not_ supported.

gvanrossum:
Let's first make an inventory of what to say. First there are the standard limitations that follow directly from mypy's general lack of understanding of dynamically computed types, for example

- dynamically computing a metaclass isn't supported
- a metaclass that dynamically modifies the class isn't supported

I believe there are also other limitations which could be classified as bugs, where in principle mypy could do the right thing but we currently don't have the code. But I haven't carefully thought through which those might be. Perhaps the loss of type when iterating over an Enum class is an example? It's fine to document those limitations as long as we classify them separately from the first category.

ilevkivskyi:
In general I like the idea of having a separate subsection about metaclasses (probably in the section "Kinds of types").

Some additional thoughts:
* it looks like this page (referenced from RtD) is outdated: https://github.com/python/mypy/wiki/Unsupported-Python-Features
* also the first two items in this list are covered in the docs and can be removed http://mypy.readthedocs.io/en/latest/additional_features.html
* in the section about metaclass we can mention that ``six.with_metaclass`` and ``six.add_metaclass`` are supported.
* in addition to listing supported metaclass features we could show several examples of typical supported use cases.

gvanrossum:
Sounds good. The RtD should probably not refer to the wiki for a list of
unsupported features.

elazarg:
#4028

PR #4028, opened 2017-09-28
Document metaclass support

elazarg:
See #3999


docs/source/class_basics.rst
```diff
@@ -151,6 +151,112 @@ concrete. As with normal overrides, a dynamically typed method can
 implement a statically typed abstract method defined in an abstract
 base class.
 
+.. _metaclasses:
```
ilevkivskyi: I think ``class_basics`` is a wrong section for this. Metaclasses are not really basic (although they definitely related to classes). I propose to put this in ``kinds_of_types``.

docs/source/class_basics.rst
```diff
@@ -151,6 +151,112 @@ concrete. As with normal overrides, a dynamically typed method can
...
+***********
+A `metaclass <https://docs.python.org/3/reference/datamodel.html#metaclasses>`_
+is a class that describes the construction and behavior of other classes,
+similarly to how classes describe the construction and behavior of objects.
+The default metaclass is ``type``, but it's possible to use other metaclasses.
+Metaclasses allows one to create "a different kind of class", such as Enums,
+NamedTuples and singletons.
+
+Mypy has some special understanding of ``ABCMeta``, ``EnumMeta``,
+``NamedTupleMeta`` and of course ``GenericMeta``.
```
ilevkivskyi: I would not mention these two, the first is really internal to typing, and there is a chance the second will not live long.

docs/source/class_basics.rst
```diff
@@ -151,6 +151,112 @@ concrete. As with normal overrides, a dynamically typed method can
...
+
+        def make(cls: Type[T]) -> T:
+            M.count += 1
+            return cls()
+    
+    class A(metaclass=M):
+        pass
+    
+    a = A.make()  # make is looked up at M
+    print(A.count)
```
ilevkivskyi: Maybe show here the types of ``a`` and ``A.count``.

docs/source/class_basics.rst
```diff
@@ -151,6 +151,112 @@ concrete. As with normal overrides, a dynamically typed method can
...
+    
+    class A1(metaclass=M1): pass
+    class A2(metaclass=M2): pass
+    
+    class B1(A1, metaclass=M2): pass  # Mypy Error: Inconsistent metaclass structure for 'B1'
+    # At runtime the above definition raises an exception
+    # TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
+    
+    # Same runtime error as in B1, but mypy does not catch it yet
+    class B12(A1, A2): pass
```
ilevkivskyi: I think this is not a limitation of metaclass support, but rather a useful feature to detect possible metaclass conflicts, so it should not be in this section, but probably in the previous one.

docs/source/class_basics.rst
```diff
@@ -151,6 +151,112 @@ concrete. As with normal overrides, a dynamically typed method can
...
+    
+* Mypy does not understand dynamically-computed metaclasses,
+  such as ``class A(metaclass=f()): ...``
+* Mypy does not and cannot understand arbitrary metaclass code. Precise support
+  for metaclasses requires writing a mypy plugin.
```
ilevkivskyi: This statement is a bit ambiguous. I understand what you wanted to say, but I am not sure someone new to mypy will understand this correctly.

docs/source/class_basics.rst
```diff
@@ -151,6 +151,112 @@ concrete. As with normal overrides, a dynamically typed method can
...
+* Mypy only recognizes classes that inherit from `type` as potential metaclasses.
```
ilevkivskyi: What about inheriting from other metaclasses that inherit from ``type``? I think this statement is not 100% clear.

docs/source/class_basics.rst
```diff
@@ -151,6 +151,112 @@ concrete. As with normal overrides, a dynamically typed method can
...
+* Inner class ``__metaclass__`` are not supported yet.
+
+It is often possible to use decorators as an alternative to metaclasses.
+Support for decorators is even more limited, but they do not cause trouble with
+inheritance.
```
ilevkivskyi: I think we don't need this paragraph.

docs/source/class_basics.rst
elazarg: The fact that mypy does not catch the bad structure in the last example is a limitation. Apart from that, you are right.

docs/source/class_basics.rst
elazarg: I think that `a subclass of` should be clearer

docs/source/class_basics.rst
elazarg: Do you have a suggestion? Perhaps not mention plugins at all? or refer to the relevant methods?

docs/source/class_basics.rst
elazarg: I renamed it to "Gotchas and limitations of metaclass support". Maybe some other title will do. I want it to be separated from the "happy path" explanation.

docs/source/class_basics.rst
ilevkivskyi: Maybe we can add this comment later, when plugin API will be stable (or at least documented somewhere).

docs/source/class_basics.rst
elazarg: I agree that it is not class basics, but it's not some kind of type either. In my opinion "class basics" should be the first subsection in the section, which should be named something like "Classes and Protocols", or "Classes, Metaclasses and Protocols" or some other generic name. Or maybe just split this section to three.

ilevkivskyi: > but it's not some kind of type either

Why not? Everything that is allowed in type context is a type, and we have a special section there about ``Type[...]`` that seems related. (It may be even instructive to explain what is the difference between ``Type[A]`` and ``M`` for ``class A(metaclass=M): ...``)

docs/source/class_basics.rst
elazarg: I'm not sure that `reveal_type` is something that the reader wants to see. Perhaps I should add some lookup failures? Or maybe something like `print(A.count + " objects were created")` which demonstrates an actually helpful error message.

ilevkivskyi: For the first I mean ``a: A = A.make()`` or using a type comment. For the second, indeed showing an error like ``incompatible arguments for + "int" and "str"`` is better.

docs/source/class_basics.rst
elazarg: Because metaclass is a runtime thing, which is quite different from most of the stuff in that page. The difference is indeed important, but that explanation can be given independently.

ilevkivskyi: > Because metaclass is a runtime thing, which is quite different from most of the stuff in that page.

Most things there are different from other things to a similar extent, this is normal. I would also split this discussion and put the "Gotchas and limitations of metaclass support" in common_issues.

ethanhs:
@elazarg would you be interested in picking this back up? You may need to refactor this to fit with the new docs structure, FWIW.

elazarg:
@ethanhs I'm not sure what kind of refactoring is needed. If you want to take this on (using my version as a basis or otherwise) it's fine with me.

ethanhs:
@elazarg I was mostly meaning the new layout of the docs. https://mypy.readthedocs.io/en/latest/

I think it could be its own item after class basics.

elazarg:
@ethanhs is this what you meant? (Only I put it after protocols because I believe it's less important)

ethanhs:
@elazarg Yes, thank you for making this change.

msullivan:
So is this good to land?

PR #4034, opened 2017-09-29
Revert "Sync typeshed"

gvanrossum:
Reverts python/mypy#4033

This excludes python/typeshed#1549, which turned `__class__` into a property, and our internal codebases have some problems with that.


gvanrossum:
Instead, let's re-sync typeshed to include https://github.com/python/typeshed/pull/1632

Issue #3852, opened 2017-08-21
Crash with "Should never get here in normal mode" with incremental=True


lordmauve:
I was working in a file, adding types and replacing [attrs](http://www.attrs.org/en/stable/) with `typing.NamedTuple` in order to benefit from types, when I got this crash:

```
$ bin/mypy python/
Traceback (most recent call last):
  File "bin/mypy", line 33, in <module>
    sys.exit(console_entry())
  File "dist/lib/python3.6/mypy/__main__.py", line 7, in console_entry
    main(None)
  File "dist/lib/python3.6/mypy/main.py", line 50, in main
    res = type_check_only(sources, bin_dir, options)
  File "dist/lib/python3.6/mypy/main.py", line 97, in type_check_only
    options=options)
  File "dist/lib/python3.6/mypy/build.py", line 196, in build
    graph = dispatch(sources, manager)
  File "dist/lib/python3.6/mypy/build.py", line 1801, in dispatch
    process_graph(graph, manager)
  File "dist/lib/python3.6/mypy/build.py", line 2037, in process_graph
    process_fresh_scc(graph, prev_scc)
  File "dist/lib/python3.6/mypy/build.py", line 2106, in process_fresh_scc
    graph[id].fix_cross_refs()
  File "dist/lib/python3.6/mypy/build.py", line 1547, in fix_cross_refs
    self.manager.options.quick_and_dirty)
  File "dist/lib/python3.6/mypy/fixup.py", line 22, in fixup_module_pass_one
    node_fixer.visit_symbol_table(tree.names)
  File "dist/lib/python3.6/mypy/fixup.py", line 105, in visit_symbol_table
    self.visit_type_info(value.node)
  File "dist/lib/python3.6/mypy/fixup.py", line 62, in visit_type_info
    base.accept(self.type_fixer)
  File "dist/lib/python3.6/mypy/types.py", line 435, in accept
    return visitor.visit_instance(self)
  File "dist/lib/python3.6/mypy/fixup.py", line 176, in visit_instance
    assert self.quick_and_dirty, "Should never get here in normal mode"
AssertionError: Should never get here in normal mode
```

My mypi.ini contained

```
[mypy]
follow_imports = silent
incremental = True
```

However the error went away when I removed `incremental = True`.

gvanrossum:
Thanks for the report -- if you can still repro this, is your source code public, and can you save a tarball of your `.mypy_cache`?

lordmauve:
[...]

Issue #3052, opened 2017-03-24
incremental: Crash after deleting a function.


timabbott:
I ran into this crash with 0.501 today on the Zulip codebase when running mypy with this version of the code https://github.com/zulip/zulip/commit/09f66b5c6dfc5c8bf5d52fb51259189ade0f336e (which deleted the below function).  

I've attached the cache state tarball in case it's helpful: [mypy.tar.gz](https://github.com/python/mypy/files/866606/mypy.tar.gz)

```
Traceback (most recent call last):
  File "/srv/zulip-py3-venv/bin/mypy", line 6, in <module>
    main(__file__)
  File "/srv/zulip-venv-cache/015153c2b7c837e69f696afeaa7df4d23e89fed1/zulip-py3-venv/lib/python3.5/site-packages/mypy/main.py", line 42, in main
    res = type_check_only(sources, bin_dir, options)
  File "/srv/zulip-venv-cache/015153c2b7c837e69f696afeaa7df4d23e89fed1/zulip-py3-venv/lib/python3.5/site-packages/mypy/main.py", line 86, in type_check_only
    options=options)
  File "/srv/zulip-venv-cache/015153c2b7c837e69f696afeaa7df4d23e89fed1/zulip-py3-venv/lib/python3.5/site-packages/mypy/build.py", line 183, in build
    dispatch(sources, manager)
  File "/srv/zulip-venv-cache/015153c2b7c837e69f696afeaa7df4d23e89fed1/zulip-py3-venv/lib/python3.5/site-packages/mypy/build.py", line 1531, in dispatch
    process_graph(graph, manager)
  File "/srv/zulip-venv-cache/015153c2b7c837e69f696afeaa7df4d23e89fed1/zulip-py3-venv/lib/python3.5/site-packages/mypy/build.py", line 1761, in process_graph
    process_fresh_scc(graph, prev_scc)
  File "/srv/zulip-venv-cache/015153c2b7c837e69f696afeaa7df4d23e89fed1/zulip-py3-venv/lib/python3.5/site-packages/mypy/build.py", line 1830, in process_fresh_scc
    graph[id].fix_cross_refs()
  File "/srv/zulip-venv-cache/015153c2b7c837e69f696afeaa7df4d23e89fed1/zulip-py3-venv/lib/python3.5/site-packages/mypy/build.py", line 1311, in fix_cross_refs
    fixup_module_pass_one(self.tree, self.manager.modules)
[...]

PR #4038, opened 2017-09-30
Fix --incremental crash on Type[...]

### LINKED ISSUE #4058: self.__class__.__name__ has type Any in class derived from Any
Here's something that came up in https://github.com/python/typeshed/pull/1549#issuecomment-333227407.
```py
from typing import Any, Type, TypeVar
T = TypeVar('T')
class C(Any):
    def bar(self: T) -> Type[T]: ...
    def foo(self) -> None:
        reveal_type(self.bar())  # Type[C], as expected
        reveal_type(self.bar().__name__)  # Any, but expected str
```
@ilevkivskyi I think you muttered something about `analyze_member_access` for `TypeType` still being broken but I'm not sure I understand.

ilevkivskyi:
This fixes the second traceback reported in #3852. This also improves member lookup on ``Type[...]`` by moving it to a later stage (``checkmember.py`` instead of some manipulations in ``TypeType`` constructor that are dangerous during de-serialization).

By the way it looks like the first traceback reported in #3852 is just another manifestation of #3052, but instead of just deleting a class it is replaced by a ``Var`` or by a ``FunctionDef``.

Also fixes #4058.

### DIFF
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -717,8 +717,7 @@ class CallableType(FunctionLike):
         )
 
     def is_type_obj(self) -> bool:
-        t = self.fallback.type
-        return t is not None and t.is_metaclass()
+        return self.fallback.type.is_metaclass()
 
     def is_concrete_type_obj(self) -> bool:
         return self.is_type_obj() and self.is_classmethod_class
@@ -1341,10 +1340,7 @@ class TypeType(Type):
         type UnionType must be handled through make_normalized static method.
         """
         super().__init__(line, column)
-        if isinstance(item, CallableType) and item.is_type_obj():
-            self.item = item.fallback
-        else:
-            self.item = item
+        self.item = item
 
     @staticmethod
     def make_normalized(item: Type, *, line: int = -1, column: int = -1) -> Type:
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -177,26 +177,35 @@ def analyze_member_access(name: str,
     elif isinstance(typ, TypeType):
         # Similar to FunctionLike + is_type_obj() above.
         item = None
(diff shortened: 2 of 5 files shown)


gvanrossum:
Is this something urgent (fixing a regression in a recent change) that should be cherry-picked into the release branch? Or is this something that has been around for a long time? It seems the second crash in #3852 is related to pointing mypypath to site-packages, which is not recommended.

ilevkivskyi:
> Is this something urgent (fixing a regression in a recent change) that should be cherry-picked into the release branch? Or is this something that has been around for a long time?

The bug is quite old (around a year), it started to be a crash in 0.510. So I don't think it is something urgent.

> It seems the second crash in #3852 is related to pointing mypypath to site-packages, which is not recommended.

The crash doesn't depend on ``MYPYPATH`` setting, but the probability to find more dynamic code in libraries is higher. For example, the crash appears on ``type(f)`` , where ``f`` is a callable other than class object.

mypy/types.py
```diff
@@ -1341,10 +1340,7 @@ def __init__(self, item: Union[Instance, AnyType, TypeVarType, TupleType, NoneTy
         type UnionType must be handled through make_normalized static method.
         """
         super().__init__(line, column)
-        if isinstance(item, CallableType) and item.is_type_obj():
-            self.item = item.fallback
-        else:
-            self.item = item
+        self.item = item
```
JukkaL: Why does this change help? Is it because the `type` attribute of an `Instance` might refer to `FakeInfo` during deserialization?

ilevkivskyi: Yes.

ilevkivskyi: (Or previously to ``None``)

PR #4041, opened 2017-10-01
Require typing module before using a Py2 interpreter for tests.

### LINKED ISSUE #4039: Skip Py2 subprocess tests if Py2 interpreter found does not have typing installed
The behavior of mypy with regard to Python 2.7 tests is inconsistent. On the one hand, if no Python 2.7 interpreter can be found at all, https://github.com/python/mypy/blob/master/mypy/util.py#L66 will return `None` and mypy will happily skip the Py2 tests. But if a Python 2.7 interpreter is found, but is missing the `typing` backport (which mypy no longer bundles), mypy will go ahead and run the Py2 tests and they will fail.

Our build boxes are in the midst of an upgrade from an OS version with Python 2.6 as system Python to one with Python 2.7. On the older OS, mypy builds fine and just skips the Py2 tests (which is fine). On the newer OS, it finds the system Python 2.7 and those tests fail for lack of `typing`. It'sdifficult for us to install the `typing` module into the system Python on these hosts.

I propose that it would be more consistent if mypy considers "no Python 2.7 with typing available" equivalent to "no Python 2.7 available" and skips the Py2 tests. It would be quite easy to add a `python -c "import typing"` check to `try_find_python2_interpreter`. I can submit a PR for this, if the idea is approved.

carljm:
Fixes #4039.

Tested with:

a) Python 2.7 interpreter on PATH without typing installed: Py2 subprocess
tests skipped.

b) Python 2.7 interpreter on PATH with typing installed: Py2 subprocess tests
pass.

c) Temporarily change assertion to require `(2, 6)` instead of `(2, 7)`, with
no Python2.6 interpreter available, and `typing` installed in Python2.7 interpreter
on PATH (to verify that version assert works correctly): Py2 subprocess tests skipped.

### DIFF
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -69,10 +69,11 @@ def try_find_python2_interpreter() -> Optional[str]:
         return _python2_interpreter
     for interpreter in default_python2_interpreter:
         try:
-            process = subprocess.Popen([interpreter, '-V'], stdout=subprocess.PIPE,
-                                       stderr=subprocess.STDOUT)
-            stdout, stderr = process.communicate()
-            if b'Python 2.7' in stdout:
+            retcode = subprocess.Popen([
+                interpreter, '-c',
+                'import sys, typing; assert sys.version_info[:2] == (2, 7)'
+            ]).wait()
+            if not retcode:
                 _python2_interpreter = interpreter
                 return interpreter
         except OSError:


carljm:
I think the Travis failure here is some kind of flaky unrelated failure, but I'm not 100% sure. Can someone with the necessary access trigger a re-run?

carljm:
Any concerns about merging this that I can do some further testing to assuage?

mypy/util.py
```diff
@@ -69,10 +69,14 @@ def try_find_python2_interpreter() -> Optional[str]:
...
-                                       stderr=subprocess.STDOUT)
-            stdout, stderr = process.communicate()
-            if b'Python 2.7' in stdout:
+            retcode = subprocess.Popen(
+                [
+                    interpreter,
+                    '-c',
+                    'import sys, typing; assert sys.version_info[:2] == (2, 7)'
+                ],
+            ).wait()
```
gvanrossum: Please don't use this incredibly stretched vertical layout; it hurts my eyes and does not help me understand the code.

carljm:
Adjusted the formatting to be more vertically compact.

PR #4045, opened 2017-10-02
Delete cache for a module if errors are found

### LINKED ISSUE #3052: incremental: Crash after deleting a function.

### LINKED ISSUE #3852: Crash with "Should never get here in normal mode" with incremental=True

### LINKED ISSUE #4043: Errors not always reported for a file in incremental mode
Sometimes mypy doesn't generate an error for a module in incremental mode, even if a previous run generated errors and nothing has changed.

To reproduce, first create these two files:

```py
# a.py
import b
b.f()
```

```py
# b.py
def f() -> None: pass
```

Then run mypy, and no errors are reported (as expected):
```
$ mypy -i a.py
```

Now edit `b.py` like this:
```py
# b.py (2)
def f(x) -> None: pass
```

Run mypy again, and it correctly reports an error:
```
$ mypy -i a.py
a.py:3: error: Too few arguments for "f"
```

Finally, run mypy again without changing anything, and it incorrectly gives clean output:
```
$ mypy -i a.py
$    <--- no error reported
```

The expected output is the same error as in the previous run, since nothing has changed.

ilevkivskyi:
Fixes #4043 
Fixes #3852 
Fixes #3052

This is a simple-minded solution for inconsistent cache state problem discovered in #4043.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -21,6 +21,7 @@ import site
 import sys
 import time
 from os.path import dirname, basename
+import errno
 
 from typing import (AbstractSet, Dict, Iterable, Iterator, List, cast, Any,
                     NamedTuple, Optional, Set, Tuple, Union, Callable)
@@ -469,7 +470,7 @@ class BuildManager:
       all_types:       Map {Expression: Type} collected from all modules
       options:         Build options
       missing_modules: Set of modules that could not be imported encountered so far
-      stale_modules:   Set of modules that needed to be rechecked
+      stale_modules:   Set of modules that needed to be rechecked (only used by tests)
       version_id:      The current mypy version (based on commit id when possible)
       plugin:          Active mypy plugin(s)
       errors:          Used for reporting all errors
@@ -1180,7 +1181,7 @@ def delete_cache(id: str, path: str, manager: BuildManager) -> None:
         try:
             os.remove(filename)
         except OSError as e:
-            if e.errno != os.errno.ENOENT:
+            if e.errno != errno.ENOENT:
                 manager.log("Error deleting cache file {}: {}".format(filename, e.strerror))
 
 
@@ -1553,11 +1554,12 @@ class State:
         """Marks this module as having been fully re-analyzed by the type-checker."""
         self.manager.rechecked_modules.add(self.id)
 
-    def mark_interface_stale(self) -> None:
+    def mark_interface_stale(self, *, on_errors: bool = False) -> None:
         """Marks this module as having a stale public interface, and discards the cache data."""
         self.meta = None
         self.externally_same = False
-        self.manager.stale_modules.add(self.id)
+        if not on_errors:
+            self.manager.stale_modules.add(self.id)
 
     def check_blockers(self) -> None:
         """Raise CompileError if a blocking error is detected."""
@@ -1833,7 +1835,7 @@ class State:
             is_errors = self.manager.errors.is_errors()
         if is_errors:
             delete_cache(self.id, self.path, self.manager)
-            self.mark_interface_stale()
+            self.mark_interface_stale(on_errors=True)
             return
         dep_prios = [self.priorities.get(dep, PRI_HIGH) for dep in self.dependencies]
         new_interface_hash = write_cache(
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1165,6 +1165,24 @@ def write_cache(id: str, path: str, tree: MypyFile,
     return interface_hash
 
 
+def delete_cache(id: str, path: str, manager: BuildManager) -> None:
+    """Delete cache files for a module.
+
+    The cache files for a module are deleted when mypy finds errors there.
+    This avoids inconsistent states with cache files from different mypy runs,
+    see #4043 for an example.
+    """
+    path = os.path.abspath(path)
+    meta_json, data_json = get_cache_names(id, path, manager)
+    manager.log('Deleting {} {} {} {}'.format(id, path, meta_json, data_json))
+
+    for filename in [data_json, meta_json]:
+        try:
+            os.remove(filename)
+        except OSError:
+            manager.log("Error deleting cache file {}".format(filename))
+
+
 """Dependency manager.
 
 Design
@@ -1813,6 +1831,7 @@ class State:
         else:
             is_errors = self.manager.errors.is_errors()
         if is_errors:
+            delete_cache(self.id, self.path, self.manager)
             return
         dep_prios = [self.priorities.get(dep, PRI_HIGH) for dep in self.dependencies]
         new_interface_hash = write_cache(
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1179,8 +1179,9 @@ def delete_cache(id: str, path: str, manager: BuildManager) -> None:
     for filename in [data_json, meta_json]:
         try:
             os.remove(filename)
-        except OSError:
-            manager.log("Error deleting cache file {}".format(filename))
+        except OSError as e:
+            if e.errno != os.errno.ENOENT:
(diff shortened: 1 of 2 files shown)


mypy/build.py
```diff
@@ -1165,6 +1165,24 @@ def write_cache(id: str, path: str, tree: MypyFile,
...
+    """
+    path = os.path.abspath(path)
+    meta_json, data_json = get_cache_names(id, path, manager)
+    manager.log('Deleting {} {} {} {}'.format(id, path, meta_json, data_json))
+
+    for filename in [data_json, meta_json]:
+        try:
+            os.remove(filename)
+        except OSError:
+            manager.log("Error deleting cache file {}".format(filename))
```
gvanrossum: This would cause a bit noisy logging. If the file doesn't exist I'd rather not log anything (or specifically log that the file doesn't exist).

mypy/build.py
```diff
@@ -1813,6 +1831,7 @@ def write_cache(self) -> None:
         else:
             is_errors = self.manager.errors.is_errors()
         if is_errors:
+            delete_cache(self.id, self.path, self.manager)
```
gvanrossum: Shouldn't this also mark the interface as stale (as below on line 1845-1847) and set self.interface_hash to something unmatchable? Otherwise I worry we could still repro the same issues by adding another chain to the reference link.

ilevkivskyi: Yes, you are right. I even have found a longer crasher with four files that still happens unless interface is marked as stale (I added it to the tests). Do we also need to set the ``interface_hash`` to something? It looks like it is not necessary.

It looks like marking files with errors as stale breaks several tests. I will check if this can be fixed somehow.

ilevkivskyi: Hm, there is an explicit comment in tests:
```
Note that a file that ends up producing an error does not create
a new cache file and so is not considered stale.
```
Should I just mark them as stale now and update this comment?

gvanrossum: Yeah, I think the comment just reflects the state of the world at the time (i.e. that no cache file was written).

I presume this only affects some of the tests and not all tests with a `[stale]` marker?

Alternatively you could have a flag to `mark_interface_stale()` to skip adding the module to `stale_modules` since that is used purely by the tests. (The other two things it sets are used for the graph algorithm too.)

ilevkivskyi: > I presume this only affects some of the tests and not all tests with a [stale] marker?

It affects 42 tests, i.e. around one third of all tests with [stale].

> Alternatively you could have a flag to ``mark_interface_stale()`` to skip adding the module to ``stale_modules`` since that is used purely by the tests. (The other two things it sets are used for the graph algorithm too.)

OK, I did this. Now all tests should pass. By the way, what do you think about potential performance issues because of this change? Intuitively it should be minor (since it only affects situations where new errors appear after previous incremental runs), but maybe you could measure it somehow?

gvanrossum:
Whee!

PR #4047, opened 2017-10-03
Move pythoneval to mypy.api and increase timeout

### LINKED ISSUE #3543: Travis CI tests failing for 3.4 on master
See e.g.
- https://travis-ci.org/python/mypy/builds/242519706
- https://travis-ci.org/python/mypy/jobs/242880097

These are commits from PRs that passed all tests. The failure is always only on the Python 3.4 build, in different eval-test-* subtests. The "Actual" test output is always empty, which makes me wonder if the processes just die for some Travis-CI-specific reason. I see nothing at https://www.traviscistatus.com/.

ethanhs:
This is a re-do of #4025.
This change moves the mypy checking of the eval files away from subprocesses, instead using `mypy.api`.
 It also increases the timeout of the `run` function in an attempt to fix #3543.

### DIFF
--- a/mypy/test/testpythoneval.py
+++ b/mypy/test/testpythoneval.py
@@ -10,8 +10,6 @@ Note: These test cases are *not* included in the main test suite, as including
       this suite would slow down the main suite too much.
 """
 
-from contextlib import contextmanager
-import errno
 import os
 import os.path
 import re
@@ -25,6 +23,7 @@ from mypy.test.config import test_data_prefix, test_temp_dir
 from mypy.test.data import DataDrivenTestCase, parse_test_cases, DataSuite
 from mypy.test.helpers import assert_string_arrays_equal
 from mypy.util import try_find_python2_interpreter
+from mypy import api
 
 # Files which contain test case descriptions.
 python_eval_files = ['pythoneval.test',
@@ -61,11 +60,7 @@ def test_python_evaluation(testcase: DataDrivenTestCase) -> None:
     version.
     """
     assert testcase.old_cwd is not None, "test was not properly set up"
-    mypy_cmdline = [
-        python3_path,
-        os.path.join(testcase.old_cwd, 'scripts', 'mypy'),
-        '--show-traceback',
-    ]
+    mypy_cmdline = ['--show-traceback']
     py2 = testcase.name.lower().endswith('python2')
     if py2:
         mypy_cmdline.append('--py2')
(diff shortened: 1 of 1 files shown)


mypy/test/testpythoneval.py
```diff
@@ -80,21 +75,23 @@ def test_python_evaluation(testcase: DataDrivenTestCase) -> None:
...
+    mypy_cmdline.append(program_path)
     with open(program_path, 'w') as file:
         for s in testcase.input:
             file.write('{}\n'.format(s))
+    output = []
     # Type check the program.
-    # This uses the same PYTHONPATH as the current process.
-    returncode, out = run(mypy_cmdline)
+    out, err, returncode = api.run(mypy_cmdline)
+    output.extend([s.rstrip('\r\n').lstrip(test_temp_dir + os.sep)
```
gvanrossum: Looks like you're misunderstanding how `lstrip()` works. It removes characters from the left as long as they are in the argument, which is surely not what you meant. (Ditto for `rstrip()` but in this case it's harmless.)

ethanhs: Ah, you are right. Apologies. The rstrip should be fine (removing trailing newlines is what I want), but the lstrip I can replace with a slice. Thanks!

ethanhs:
The AppVeyor failure is a flake.

ethanhs:
Also, I have a change set for https://github.com/python/mypy/issues/3895 but it modifies testpythoneval (to allow for filtering of test names), so it is blocked by this.

gvanrossum:
OK I've restarted the failing build.

mypy/test/testpythoneval.py
```diff
@@ -80,21 +75,24 @@ def test_python_evaluation(testcase: DataDrivenTestCase) -> None:
...
     with open(program_path, 'w') as file:
         for s in testcase.input:
             file.write('{}\n'.format(s))
+    output = []
     # Type check the program.
-    # This uses the same PYTHONPATH as the current process.
-    returncode, out = run(mypy_cmdline)
+    out, err, returncode = api.run(mypy_cmdline)
+    # split lines, remove newlines, and remove directory of test case
+    output.extend([s.rstrip('\r\n')[len(test_temp_dir + os.sep):]
```
gvanrossum: Hm. I don't like blindly clipping. Please use a regular for-loop that inspects the start of each line before stripping the temp directory and separator.

ethanhs: Okay I will do that.

ethanhs:
>I am a little surprised that this still flaked the same way

This is because the change still uses subprocesses to evaluate the Python files. I have a change set to disable the python2eval tests that will make it go away, it is just waiting on this.

gvanrossum:
I am not 100% sure I believe the story about the AppVeyor tests (this would imply that the problem was always about the subprocess running the Python code rather than the subprocess running mypy) but at this point I am willing to try anything to stop the bleeding, so I'm merging now.

PR #4048, opened 2017-10-03
Add default Python 2 install on Windows to search

ethanhs:
Working on #3895, I realized that the tests were not resolving the installed Python 2 on my machine. This is a simple fix. A run of pytest went from `4549 passed, 72 skipped` to `4596 passed, 25 skipped`.


gvanrossum:
Merged -- but do we have any reason that Python 2 is always installed on AppVeyor?

ethanhs:
In their [build environment docs](https://www.appveyor.com/docs/build-environment/#python) 2.7 is the default Python.

PR #4053, opened 2017-10-03
Skip Python2 evaluation tests on AppVeyor

### LINKED ISSUE #3895: Appveyor tests are being flaky

ethanhs:
Fixes #3895

This splits the pytest run into a new command so that tests ending in `_python2eval` are skipped. This is done in the appveyor.yml so that these tests still can be run locally on Windows. Hopefully, we will no longer have issues with pytest crashing the test suite after this.

### DIFF
--- a/mypy/test/testpythoneval.py
+++ b/mypy/test/testpythoneval.py
@@ -10,8 +10,6 @@ Note: These test cases are *not* included in the main test suite, as including
       this suite would slow down the main suite too much.
 """
 
-from contextlib import contextmanager
-import errno
 import os
 import os.path
 import re
@@ -25,6 +23,7 @@ from mypy.test.config import test_data_prefix, test_temp_dir
 from mypy.test.data import DataDrivenTestCase, parse_test_cases, DataSuite
 from mypy.test.helpers import assert_string_arrays_equal
 from mypy.util import try_find_python2_interpreter
+from mypy import api
 
 # Files which contain test case descriptions.
 python_eval_files = ['pythoneval.test',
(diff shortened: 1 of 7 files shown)


test-data/unit/python2eval.test
```diff
@@ -5,7 +5,7 @@
 -- complete.
 
 
-[case testAbs2_python2]
+[case testAbs2_python2eval]
```
gvanrossum: I really would rather not have to rename all the tests to conform to this new naming convention. I think you can select for just the pythoneval tests with `-k PythonEvaluationSuite`, so to combine these two you can run
```
pytest -k "PythonEvaluationSuite and python2"
```
(the Windows script will require some extra quoting).

ethanhs: Okay I will try that route.

mypy/test/testpythoneval.py
```diff
@@ -61,7 +61,7 @@ def test_python_evaluation(testcase: DataDrivenTestCase) -> None:
     """
     assert testcase.old_cwd is not None, "test was not properly set up"
     mypy_cmdline = ['--show-traceback']
-    py2 = testcase.name.lower().endswith('python2')
+    py2 = 'python2' in testcase.name.lower()
```
gvanrossum: I think this should be restored to exactly what it was before. I'll see if I can do the edit online myself.

Issue #3241, opened 2017-04-25
Keyword arguments of __call__ are not interpreted correctly


Sterbic:
I looks like mypy drops all keyword arguments of the `__call__` function of a class and recognises only positional arguments.

Code example:

```
class Formatter:
    def __call__(
        self,
        message: str,
        *,  # same behaviour with and without 
        bold: bool = False,
    ) -> str:
        if bold:
            message = f"** {message} **"
        return message

formatter = Formatter()
print(formatter("test", bold=True))
```

Mypy error:
`test.py:13: error: Unexpected keyword argument "bold" for "__call__" of "Formatter"`

gvanrossum:
Yes, confirmed and seems specific to `__call__`.

ilevkivskyi:
Adding ``__call__`` to ``MAGIC_METHODS_ALLOWING_KWARGS`` in ``sharedparse.py`` fixes the issue. I wonder why it is not there now, do older versions of Python not allow kwargs in ``__call__``?

gvanrossum:
I assume it was just an oversight. Do the tests pass with that one change? Then it's worth adding a test specifically for `__call__` with kwargs and push a PR.

ilevkivskyi:
OK, here is the PR https://github.com/python/mypy/pull/3245

PR #2883, opened 2017-02-20
Allow arbitrary overriding of __init_subclass__

miedzinski:
Fixes #2882. Same change as b0f6171aa329138336b028242caaf51918746c58 regarding `__new__`.

PR #4054, opened 2017-10-03
Allow arbitrary overriding of __call__

naufraghi:
The list of whitelisted names in `checker.py` and `sharedparse.py` is
converging, with this patch we use the set of names in
`sharedparse.MAGIC_METHODS_ALLOWING_KWARGS` in both places.

See #2883 and #3241


ilevkivskyi:
Since objects with ``__call__`` are considered structural subtypes of ``Callable`` with corresponding signature, this PR is type unsafe. I would not do this, unless there is a good reason.

PR #4059, opened 2017-10-04
Check code after inferred UninhabitedType

### LINKED ISSUE #3994: Code is not cheked after some function calls: binder/inference bug
Any code after an ``UninhabitedType`` is marked as unreachable in binder. This leads to dangerous behaviour together with how type inference currently works "in all uncertain situations infer uninhabited" (see also https://github.com/python/mypy/issues/3283 for discussion). For example no error is reported in this code:
```python
from typing import TypeVar
T = TypeVar('T')

def f() -> T:
    ...

class C:
    f()
    1 + 'this is not checked'

def fun() -> None:
    f()
    1 + 'can do everything here'

if True:
    f()
    1 + 'whatever'
```
There are other function signatures (all involve type variables) that lead to this behaviour: type checking stops after a call to such function.

ilevkivskyi:
Fixes #3994

Currently code after inferred ``UninhabitedType`` is silently skipped, for example no errors are shown in this code:
```python
T = TypeVar('T')
def f(x: List[T]) -> T:
    ...

class C:
    x = f([])
    def m(self) -> str:
        return 42 # No error!

if bool():
    f([])
    1 + ''  # No error!
```
There are other scenarios in tests. With this PR errors are shown as expected, since ``UninhabitedType``s resulting from ambiguous inference don't influence binder.

### DIFF
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -357,6 +357,7 @@ class UninhabitedType(Type):
     can_be_true = False
     can_be_false = False
     is_noreturn = False  # Does this come from a NoReturn?  Purely for error messages.
+    ambiguous = False  # Is this a result of inference for a variable without constraints?
 
     def __init__(self, is_noreturn: bool = False, line: int = -1, column: int = -1) -> None:
         super().__init__(line, column)
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -357,6 +357,9 @@ class UninhabitedType(Type):
     can_be_true = False
     can_be_false = False
     is_noreturn = False  # Does this come from a NoReturn?  Purely for error messages.
+    # It is important to track whether this is an actual NoReturn type, or just a result
+    # of ambiguous type inference, in the latter case we don't want to mark a branch as
+    # unreachable in binder.
     ambiguous = False  # Is this a result of inference for a variable without constraints?
 
     def __init__(self, is_noreturn: bool = False, line: int = -1, column: int = -1) -> None:
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -269,7 +269,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 self.check_runtime_protocol_test(e)
             if e.callee.fullname == 'builtins.issubclass':
                 self.check_protocol_issubclass(e)
-        if isinstance(ret_type, UninhabitedType):
+        if isinstance(ret_type, UninhabitedType) and not ret_type.ambiguous:
             self.chk.binder.unreachable()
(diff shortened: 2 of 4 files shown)


mypy/types.py
```diff
@@ -357,6 +357,7 @@ class UninhabitedType(Type):
     can_be_true = False
     can_be_false = False
     is_noreturn = False  # Does this come from a NoReturn?  Purely for error messages.
+    ambiguous = False  # Is this a result of inference for a variable without constraints?
```
JukkaL: Also mention why we care about this and where this makes a difference.

ilevkivskyi: > Also mention why we care about this and where this makes a difference.

I added a short discussion.

JukkaL:
This generates a bunch of new errors in Dropbox internal codebases. Since this fixes false negatives, they may just be issues in our code, but I'll have to verify it.

ilevkivskyi:
> This generates a bunch of new errors in Dropbox internal codebases. Since this fixes false negatives, they may just be issues in our code, but I'll have to verify it.

I think this PR shouldn't bring anything false positive, but will be interested to see your findings.

JukkaL:
This PR seems to cause an invalid type to be inferred for the following code:

```py
from typing import Any

class C(dict):
    def f(self, x: Any) -> None:
        y = super(C, x).__new__(x)
        reveal_type(y)  # None, which is unexpected
```

Previously the `reveal_type` line was skipped. The underlying issue doesn't seem new though, and similar problems also happen on master. 

This works unexpectedly also on master (even when using `--strict-optional`):

```py
from typing import TypeVar

T = TypeVar('T')

def f() -> T: pass

x = f()
reveal_type(x)  # None
```

The issue seems rare enough that it doesn't block merging this, but it would be nice to at least understand what is going on.

ilevkivskyi:
> The issue seems rare enough that it doesn't block merging this, but it would be nice to at least understand what is going on.

I don't know if it is related or not, but I have seen some problems in ``meet.py`` with ``None`` and ``UninhabitedType``. Another possible suspect is in partial ``None`` types, I remember looking at it recently I didn't really like how ``NoneTyp`` vs ``UninhabitedType`` is treated, I will check tomorrow morning if this can be fixed quickly, if not, then I propose to merge this and take a look at this problem later (you already opened an issue about this).

JukkaL:
I have a potential fix in a local branch. I'll have a PR out tomorrow.

JukkaL:
Now that #4112 was merged, the motivating example generates an error (though not all errors that we'd like) so this seems a little less urgent now. I may postpone this until the next release.

PR #4068, opened 2017-10-06
Release 0.530

gvanrossum:


gvanrossum:
(No, we should not merge this. The only difference is in mypy/version.py anyway.)

PR #4072, opened 2017-10-08
Narrow types after 'in' operator

### LINKED ISSUE #4071: Narrow types after 'in' operator.
Sometimes it is useful to narrow down a type after an ``in``/``not in`` comparison. For example:
```python
class Option:
    ...
opt: Union[Option, str]
if opt in ('verbose', 'silent'):
    reveal_type(opt)  # we know this is 'str'
else:
    ...  # still a union here

presets: List[Option]
if opt not in presets:
    ...  # still a union
else:
    reveal_type(opt)  # we know this can be only 'Option'
```
This seems to be especially useful for ``--strict-optional``, for example:
```python
class Node:
    fullname: Optional[str]
n: Node

reverse_aliases: Dict[str, str]
if n.fullname in reverse_aliases:
    reveal_type(n.fullname)  # This should be 'str'
```
(related https://github.com/python/mypy/issues/2980)

A similar idea was discussed in https://github.com/python/mypy/issues/1749, but emitting an error on ``if 'a' in [1, 2, 3]: ...`` would require a larger change: emitting errors on unreachable code in binder.

ilevkivskyi:
Fixes #4071 

This is separated from #4070 plus added various tests.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -53,7 +53,7 @@ from mypy.semanal import set_callable_name, refers_to_fullname
 from mypy.erasetype import erase_typevars
 from mypy.expandtype import expand_type, expand_type_by_instance
 from mypy.visitor import NodeVisitor
-from mypy.join import join_types
+from mypy.join import join_types, join_type_list
 from mypy.treetransform import TransformVisitor
 from mypy.binder import ConditionalTypeBinder, get_declaration
 from mypy.meet import is_overlapping_types
@@ -2874,6 +2874,27 @@ def remove_optional(typ: Type) -> Type:
         return typ
 
 
+def builtin_item_type(tp: Type) -> Optional[Type]:
+    # This is only OK for built-in containers, where we know the behavior of __contains__.
+    if isinstance(tp, Instance):
+        if tp.type.fullname() in ['builtins.list', 'builtins.tuple', 'builtins.dict',
+                                  'builtins.set', 'builtins.frozenset']:
+            if not tp.args:
+                # TODO: make lib-stub/builtins.pyi define generic tuple.
+                return None
+            if not isinstance(tp.args[0], AnyType):
+                return tp.args[0]
+    elif isinstance(tp, TupleType) and all(not isinstance(it, AnyType) for it in tp.items):
+        return join_type_list(tp.items)
+    elif isinstance(tp, TypedDictType):
+        # TypedDict always has non-optional string keys.
+        if tp.fallback.type.fullname() == 'typing.Mapping':
+            return tp.fallback.args[0]
+        elif tp.fallback.type.bases[0].type.fullname() == 'typing.Mapping':
+            return tp.fallback.type.bases[0].args[0]
+    return None
+
+
 def and_conditional_maps(m1: TypeMap, m2: TypeMap) -> TypeMap:
     """Calculate what information we can learn from the truth of (e1 and e2)
     in terms of the information that we can learn from the truth of e1 and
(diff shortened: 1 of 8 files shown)


JukkaL:
As discussed in https://github.com/python/mypy/pull/4099#discussion_r144272445, let's only do this for optional types, since other cases can be unsafe because of operator overloading, and these other cases seem pretty rare anyway.

ilevkivskyi:
OK, I will do this later this evening or tomorrow morning.

ilevkivskyi:
@JukkaL 
> let's only do this for optional types

Done!

mypy/checker.py
```diff
@@ -2874,6 +2874,27 @@ def remove_optional(typ: Type) -> Type:
         return typ
 
 
+def builtin_item_type(tp: Type) -> Optional[Type]:
+    # This is only OK for built-in containers, where we know the behavior of __contains__.
```
JukkaL: Add docstring.

test-data/unit/check-isinstance.test
```diff
@@ -1770,3 +1769,180 @@ def narrow_any_to_str_then_reassign_to_int() -> None:
...
+    reveal_type(y) # E: Revealed type is 'Union[builtins.int, builtins.None]'
+else:
+    reveal_type(y) # E: Revealed type is 'builtins.int'
+[builtins fixtures/tuple.pyi]
+[out]
+
+[case testNarrowTypeAfterInDict]
+# flags: --strict-optional
+from typing import Dict, Optional
+x: Dict[str, str]
```
JukkaL: Use a different type for values to make sure that this works against the key type.

test-data/unit/check-isinstance.test
```diff
@@ -1770,3 +1769,180 @@ def narrow_any_to_str_then_reassign_to_int() -> None:
         reveal_type(v)  # E: Revealed type is 'Any'
 
 [builtins fixtures/isinstance.pyi]
+
+[case testNarrowTypeAfterInList]
+# flags: --strict-optional
+from typing import List, Optional
+
+x: List[int]
```
JukkaL: Add test for optional item type (e.g. `List[Optional[int]]`).

ilevkivskyi:
@JukkaL Thanks for review! I implemented the changes in the new commit.

test-data/unit/check-isinstance.test
```diff
@@ -1788,6 +1788,20 @@ else:
 [builtins fixtures/list.pyi]
 [out]
 
+[case testNarrowTypeAfterInListOfOptional]
+# flags: --strict-optional
+from typing import List, Optional
+
+x: List[Optional[int]]
```
JukkaL: Here's another special case: What if the container is a nested one, say `List[List[x]]`, where `x` might be `Any`? How do we deal with various item types, such as `List[int]` and `List[Any]`?

ilevkivskyi: a) If the container is ``List[Any]``, then we do nothing (there is a test for this).
b) If the container is ``List[List[Any]]`` we narrow the type (provided there is an overlap, this is consistent with how ``==`` currently treated), for example:
```python
x: Optional[int]
lst: Optional[List[int]]
nested: List[List[Any]]
if lst in nested:
    reveal_type(lst) # List[int]
if x in nested:
    reveal_type(x) # Optional[int]
```
There is already a test for non-overlapping items ``int`` vs ``str``. I will add one more test for other (nested) types?

ilevkivskyi: OK, added one more test.

PR #4073, opened 2017-10-08
Push correct enclosing class while deferring a method

### LINKED ISSUE #4069: Class level __getattr__ causes error: __getattr__ is not valid at the module level outside a stub file
After updating to mypy 0.530, I am getting an error on a `__getattr__()` method defined on a class:

```
$ mypy --version
mypy 0.530
$ python3 --version
Python 3.6.3
[~/cretonne/lib/cretonne/meta]
$ cat mypy.ini 
[mypy]
disallow_untyped_defs = True
warn_unused_ignores = True
warn_return_any = True
[~/cretonne/lib/cretonne/meta]
$ mypy cdsl/formats.py
cdsl/formats.py:143: error: __getattr__ is not valid at the module level outside a stub file
```

I don't have any module level `__getattr__()` functions.

This didn't reproduce with a smaller test case, but you can check out the [Cretonne repo](https://github.com/stoklund/cretonne) to reproduce.

ilevkivskyi:
Fixes #4069 

This pushes the correct enclosing class while deferring a method instead of just the top class.

A related comment: the name ``active_class`` is a bit misleading, it returns ``None`` from within methods, I think it would be more consistent if ``active_class`` behave the same way as ``self.type`` in semantic analysis, but it is too late to change this, it is already used in several other places with this semantics.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3326,7 +3326,7 @@ class Scope:
         top = self.top_function()
         assert top, "This method must be called from inside a function"
         index = self.stack.index(top)
-        assert index
+        assert index, "Scope stack must always start with a module"
         enclosing = self.stack[index - 1]
         if isinstance(enclosing, TypeInfo):
             return enclosing
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -259,15 +259,14 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
 
     def handle_cannot_determine_type(self, name: str, context: Context) -> None:
         node = self.scope.top_function()
-        if (self.pass_num < LAST_PASS and node is not None
-                and isinstance(node, (FuncDef, LambdaExpr))):
+        if self.pass_num < LAST_PASS and isinstance(node, (FuncDef, LambdaExpr)):
             # Don't report an error yet. Just defer.
             if self.errors.type_name:
                 type_name = self.errors.type_name[-1]
             else:
                 type_name = None
             # Shouldn't we freeze the entire scope?
-            active_class = self.scope.active_class()
+            active_class = self.scope.enclosing_class()
             self.deferred_nodes.append(DeferredNode(node, type_name, active_class))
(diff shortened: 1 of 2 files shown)


elazarg:
I think I am to blame for the name `active_class`,  and I agree it's not the best. But the intention is that `active` is  `active scope`, and inside a method there is no active class scope. I don't think it's the same as "the current type".

Can't the issue be resolved by freezing the entire scope, as hinted by the `TODO`?

Anyway, if you are leaving the method `active_class()`, and assigning `enclosing_scope()` to a variable, I think it can't be called `active_class`; it will be confusing.

ilevkivskyi:
> Can't the issue be resolved by freezing the entire scope, as suggested by the ``TODO``?

It is unnecessary, and will only make solution more complex.

> Anyway, if you are leaving the method ``active_class()``, and assigning ``enclosing_scope()`` to a variable, I think it can't be called ``active_class``; it will be confusing.

OK, I renamed this variable, I keep however the named tuple attribute name, ``active_typeinfo``, it makes sense, since the enclosing class for a method becomes the active class when the deferred node rechecking starts.

JukkaL:

ilevkivskyi:
@JukkaL 
> Can you fix the merge conflict?

Fixed.

JukkaL:
This generates unexpected errors when run against an internal Dropbox codebase. Here is a simplified example:

```py
class A:
    def f(self) -> None:
        def g(x: int) -> int:  # Self argument missing for a non-static method 
                               # (or an invalid type for self)
            return y

y = int()
```

ilevkivskyi:
> This generates unexpected errors when run against an internal Dropbox codebase. Here is a simplified example: ...

Indeed, since the ``active_typeinfo`` is only used by ``bind_self`` and friends, we should only push the _immediate_ enclosing class. Will push a fix in a second.

ilevkivskyi:
@JukkaL Fixed.

Issue #3785, opened 2017-07-31
`Argument` issues and needed clarifications


elazarg:
I have several comments/questions regarding `Argument` and `Var`
1. Following #3783, attribute `initialization_statement` is not needed. It is only being copied and printed. Without it, it becomes pure data. There's also redundant bookkeeping in other places
3. `FuncItem` steals the kinds and names from `Argument` on initialization, yet they are still accessed in several places - it's not clear who "owns" the information
2. `arg.type_annotation` is also not really used (the type is read from `arg.variable.type`) and is not even bound
4. In general, the relationship between `Argument` and `Var` (which is its first, well, argument) is not clear
5. More so, given that a single argument may be responsible for multiple variables in Python2.7. This is hacked away in the parser, but the way it is handled (making a regular argument and adding an unpacking statement) makes it very difficult to give precise error messages

PR #3783, opened 2017-07-31
Better error message for incompatible default argument (take 2)

elazarg:
Reopen #3773

PR #4074, opened 2017-10-08
Remove Argument.initialization_statement

elazarg:
As mentioned in #3785, since #3783 the initialization statement for Argument is not used anywhere, except when printing, copying etc. What's left is almost pure data.

Most of the changes in the PR is the expected output in semanal and parse tests.


gvanrossum:
Thanks for the cleanup!

Issue #2755, opened 2017-01-25
TypedDict tests don't typecheck


ddfisher:
Many of the TypedDict tests in `mypy/test/testextensions.py` don't actually typecheck, even under 3.6 with the eval'd code placed properly in the codebase.  This likely indicates a problem with the tests, as the constructs used are intended to typecheck.  This is also the only remaining part of mypy that isn't fully typechecked.

ddfisher:
@ilevkivskyi, would you mind taking a look at this at some point?

ilevkivskyi:
Yes, I will try to fix this.

ilevkivskyi:
This depends on https://github.com/python/typeshed/pull/990 and https://github.com/python/mypy/issues/2492. After these two will be fixed, this one should be easy.

97littleleaf11:
It seems that we don't have `mypy/test/testextensions.py` any more. Has this issue been fixed already? @ilevkivskyi

ilevkivskyi:
Yes, I think this is not relevant anymore.

PR #4076, opened 2017-10-09
Fix config file parsing of disallow_any/disallow_untyped_defs combination

### LINKED ISSUE #4075: Setting `disallow_any = generics` in mypy.ini breaks `disallow_untyped_defs = True`
I have the following code (`rpc_server.py`):
```
def get_tasks(self):
    return 'whatever'
```
And `mypy.ini`:
```
[mypy]
disallow_untyped_defs = True
disallow_any = generics
```
I would expect to get an error because `get_tasks` is missing an argument annotation and a return type annotation.

However, `mypy --config-file /path/to/mypy.ini /path/to/rpc_server.py` gives no errors. When I remove `disallow_any = generics`, I get the expected error. I also get the expected error if I simply do `mypy --disallow-untyped-defs --disallow-any=generics`—somehow it makes a difference that I use a config file.

JelleZijlstra:
Fixes #4075.

Due to the way `disallow_any` is parsed in config files, any `disallow_any` directive basically overrode `disallow_untyped_defs`. This confused me in #4075, and it also meant that `disallow_untyped_defs` was implicitly disabled for mypy itself, because of the way `mypy_self_check.ini` was written.

This PR solves the problem by making `disallow_any` override `disallow_untyped_defs` only if the config file either has `disallow_any = ` (disabling all `disallow_any` checks) or explicitly lists `unannotated` (enabling `disallow_untyped_defs`). This passes all tests (including a new one), but it's awkward because it makes `unannotated` behave differently from other `disallow_any` flags: if a more general config has `disallow_any =  A B` and a more specific one has `disallow_any = A`, then `B` should be disabled, but this is no longer true for `unannotated`.

Here are two other options for solving the problem:
* We could parse `disallow_untyped_defs` and `disallow_any` completely independently, and just AND them at runtime when we check whether to give an error for an untyped def. This leads to an awkward situation because now if you have both `disallow_any = unannotated` in a general config and `disallow_untyped_defs = False` in the specific one, untyped defs still produce errors.
* The root cause of the problem is that we have two options that do the same thing, so we should get rid of one of them so we don't have to worry about how they interact. I think that's what @ilinum originally did, but I argued against it because `disallow_untyped_defs` is more intuitively named. What if we just get rid of `disallow_any = unannotated` instead?

In this PR I also fix the untyped defs that crept into mypy itself due to the bug.

### DIFF
--- a/mypy_self_check.ini
+++ b/mypy_self_check.ini
@@ -7,7 +7,12 @@ no_implicit_optional = True
 disallow_any = generics, unimported
 warn_redundant_casts = True
 warn_unused_ignores = True
+warn_unused_configs = True
 
 # historical exception
 [mypy-mypy.semanal]
 strict_optional = False
+
+# needs py2 compatibility
+[mypy-mypy.test.testextensions]
+disallow_untyped_defs = False
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -750,7 +750,13 @@ def parse_section(prefix: str, template: Options,
             print("%s: %s: %s" % (prefix, key, err), file=sys.stderr)
             continue
         if key == 'disallow_any':
-            results['disallow_untyped_defs'] = v and 'unannotated' in v
+            # "disallow_any = " should disable all disallow_any options, including untyped defs,
+            # given in a more general config.
+            if not v:
+                results['disallow_untyped_defs'] = False
+            # If "unannotated" is explicitly given, turn on disallow_untyped_defs.
+            elif 'unannotated' in v:
+                results['disallow_untyped_defs'] = True
         if key == 'silent_imports':
             print("%s: silent_imports has been replaced by "
                   "ignore_missing_imports=True; follow_imports=skip" % prefix, file=sys.stderr)
--- a/mypy/test/testcmdline.py
+++ b/mypy/test/testcmdline.py
@@ -41,7 +41,7 @@ class PythonEvaluationSuite(DataSuite):
                                   native_sep=True)
         return c
 
-    def run_case(self, testcase: DataDrivenTestCase):
+    def run_case(self, testcase: DataDrivenTestCase) -> None:
         test_python_evaluation(testcase)
 
 
--- a/mypy/test/testpythoneval.py
+++ b/mypy/test/testpythoneval.py
@@ -49,7 +49,7 @@ class PythonEvaluationSuite(DataSuite):
                     test_python_evaluation, test_temp_dir, True)
         return c
 
-    def run_case(self, testcase: DataDrivenTestCase):
+    def run_case(self, testcase: DataDrivenTestCase) -> None:
         test_python_evaluation(testcase)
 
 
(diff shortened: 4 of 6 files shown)


mypy_self_check.ini
```diff
@@ -7,7 +7,12 @@ no_implicit_optional = True
 disallow_any = generics, unimported
 warn_redundant_casts = True
 warn_unused_ignores = True
+warn_unused_configs = True
 
 # historical exception
 [mypy-mypy.semanal]
 strict_optional = False
+
+# needs py2 compatibility
```
JelleZijlstra: I'm not sure this is true, but the file is syntactically valid in Python 2 and it tests a piece of code (mypy_extensions.py) that is supposed to work in Python 2. @ilevkivskyi can you confirm.

ilevkivskyi: Yes, this file supposed to run on both Python 3 _and_ Python 2. There are still some problem with type checking it, but there is a separate issue for this https://github.com/python/mypy/issues/2755

gvanrossum:
Thanks for finding this. I have to agree that I would rather get rid of one of the two spellings (your second bullet). I also dislike the double negative of "disallow_any = unannotated" so I would prefer to get rid of that.

FWIW I also would prefer to have a bunch of separate Boolean flags instead of the single complex flag that is `disallow_any`. I argued against the current set-based design during its inception but lost; IIRC @ilinum eventually found a problem that would have been easier with the separate-flags design, but he was too far into the implementation to switch.

(That problem, FWIW, is that if you want to have `disallow_any = X, Y, Z` in your main section and then disable `Y` in a module-specific section, you have to write `disallow_any = X, Z` in that subsection. And then later if you change your main section to `disallow_any = X, Y, Z, U` you have to remember to add `U` to all the module-specific sections that override it (assuming they just meant to remove `Y`).)

JelleZijlstra:
Let's do this for now:
* Remove `disallow_any = unannotated` in this PR, leaving only `disallow_untyped_defs = True`
* File an issue to replace the other `disallow_any` flags with boolean flags. (I might be able to work on that, but not immediately.)
Does that sound good?

ilinum:
That sounds good to me!
Removing `disallow_any=unannotated` seems like a good call. Do we need to deprecate it first? Or can we just remove it? What's the right approach here?

I am also okay with replacing `disallow_any` with boolean flags for the reasons that Guido pointed out.

gvanrossum:
I propose to just drop it. People who use it will get failures after they update (possibly at 0.540 already, which will be frozen this Friday and released Wednesday next week) but IMO that's fine, people always test new releases somewhat when they upgrade.

The only thing we might add would be to have a special case in the error message that checks if `unannotated` is among the given strings and then prints `use --disallow-untyped-defs` instead.

JukkaL:
The PR has merge conflicts.

JelleZijlstra:
Fixed.

PR #4080, opened 2017-10-09
Update documentation of SymbolTableNode

JukkaL:


mypy/nodes.py
```diff
@@ -2262,6 +2296,8 @@ class SymbolTableNode:
     # for other nodes, optionally the name of the referenced object.
     cross_ref = None  # type: Optional[str]
     # Was this node created by normalіze_type_alias?
+    #
+    # TODO: Write a better comment, this is confusing
```
ilevkivskyi: This one distinguishes ``typing.List`` and ``builtins.list`` (sorry this comment is my fault)

mypy/nodes.py
```diff
@@ -2234,21 +2234,55 @@ def __getattribute__(self, attr: str) -> None:
...
+
+    A few definitions get special kinds, including type variables (TVAR),
+    imported modules and module aliases (MODULE_REF), and type aliases
+    (TYPE_ALIAS).
+
+    Type aliases are very special and have additional attributes that
+    are only used for them ('type_override', 'alias_tvars' at least).
+    """
+    # TODO: This is a mess. Refactor!
+    # TODO: Describe how type aliases work.
```
ilevkivskyi: Type aliases are quite complicated (due to several corner cases). I tried to make some simplifications recently. It seems to me that we can introduce a dedicated ``SybmolNode`` (like ``Var`` or ``TypeInfo``) and deprecate ``type_override`` and friends and just point ``.node`` to this new symbol node. This is major refactoring, but I think it may help.

mypy/nodes.py
```diff
@@ -2234,21 +2234,55 @@ def __getattribute__(self, attr: str) -> None:
...
+    Symbol tables don't contain direct references to AST nodes primarily
+    because there can be multiple symbol table references to a single
+    AST node (due to imports and aliases), and different references can
+    behave differently. This class describes the unique properties of
+    each reference.
+
+    The most fundamental attributes are 'kind' and 'node'.  The 'node'
+    attribute defines the AST node that the name refers to.
+
+    For many bindings, including those targeting variables, functions
```
ilevkivskyi: Maybe it will be helpful to note that this is a bit similar to how CPython symbol tables work. There are no ``.node``, only ``.kind`` (and there are more than three kinds however). These kinds are used to emit correct lookup opcodes, so that the right ``.node`` will be found by the interpreter at runtime (I know this comparison is vague).

JukkaL: Hmm talking about CPython implementation details might not be helpful, as many readers likely aren't familiar with them.

mypy/nodes.py
JukkaL: Yeah, that's what I was thinking about. I'd also like to deprecate the special handling of module references and other special node kinds so that we could remove the `kind` attribute and replace it with a scope attribute, with only 3 possible values, corresponding to `LDEF`, `GDEF` and `MDEF`. Not sure how feasible this would be. It would be quite a big refactoring in any case.

mypy/nodes.py
JukkaL: Thanks for the hint! I'll update the comment.

mypy/nodes.py
ilevkivskyi: Maybe it makes sense to open an issue to track this?

JukkaL: Created https://github.com/python/mypy/issues/4082.

PR #4087, opened 2017-10-09
Infer constraints against tuple fallbacks

### LINKED ISSUE #1485: SupportsAbs not compatible with NamedTuple
```
from typing import SupportsAbs, NamedTuple

class Point(NamedTuple('Point', [('x', int), ('y', int)]), SupportsAbs[int]):
    def __abs__(p) -> int:
        return abs(p.x) + abs(p.y)

def test(a: Point) -> bool:
    return abs(a) == 2  # E: Argument 1 to "abs" has incompatible type "Point"; expected SupportsAbs[None]
```

There's no error in `test` if the `NamedTuple` superclass of `Point` is removed.

ilevkivskyi:
Fixes #1485

Tuple fallback was overlooked in constraint inference. As usual, I place nominal before structural.


mypy/constraints.py
```diff
@@ -373,11 +373,12 @@ def visit_instance(self, template: Instance) -> List[Constraint]:
                 cb = infer_constraints(template.args[0], item, SUPERTYPE_OF)
                 res.extend(cb)
             return res
+        elif isinstance(actual, TupleType) and self.direction == SUPERTYPE_OF:
+            return infer_constraints(template, actual.fallback, self.direction)
         elif (isinstance(actual, TupleType) and template.type.is_protocol and
```
Daenyth: I don't see how this branch can ever be entered - when this condition is true, the condition of the branch above will be true, and this won't get entered.

ilevkivskyi: Yes, this branch is redundant now.

JukkaL:

ilevkivskyi:
Fixed!

Appveyor failure is a flake.

PR #4092, opened 2017-10-10
Refactoring: Make the state of type forward references explicit

JukkaL:
Forward references can be either unbound or resolved. Also ensure
that each forward reference is only resolved at most once.

As a semi-related change, be more careful about forward references
leaking after semantic analysis.

### DIFF
--- a/mypy/indirection.py
+++ b/mypy/indirection.py
@@ -103,4 +103,7 @@ class TypeIndirectionVisitor(SyntheticTypeVisitor[Set[str]]):
         return self._visit(t.item)
 
     def visit_forwardref_type(self, t: types.ForwardRef) -> Set[str]:
-        return self._visit(t.link)
+        if t.resolved:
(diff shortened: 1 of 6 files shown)


ilevkivskyi:
A related note: as I mentioned in https://github.com/python/mypy/pull/3952#issuecomment-330804438 some double-forward references still don't work (stay unresolved/unbound). This is because third pass happens only once. In ``checker.py`` we have deferred nodes, and IIUC there are mini-passes over them until they are all done, I think we can adopt the same idea here and have repeated third passes. Of course in real world people will use such double forward references extremely rarely, but this is rather a question of "completeness". What do you think? Do we need to care?

JukkaL:
Let's not care about this yet. We have bigger fish to fry. Can you add an issue to track this?

ilevkivskyi:
> Can you add an issue to track this?

Opened #4095

PR #4099, opened 2017-10-12
Refactor: Move semantic analyzer pass 3 to a new module

JukkaL:
Just move things around, tweak inter-module references and add
necessary imports.

The third pass still directly accesses the second pass so there is
more work to do.


JukkaL:
I'm going to merge this soon if nobody says anything so that I can move on with follow-up refactorings. This is all straightforward.

ilevkivskyi:
Wait few minutes please. I am from the phone now.

12 Жов 2017 13:07 "Jukka Lehtosalo" <notifications@github.com> пише:

mypy_self_check.ini
```diff
@@ -11,3 +11,6 @@ warn_unused_ignores = True
 # historical exception
 [mypy-mypy.semanal]
 strict_optional = False
+
+[mypy-mypy.semanal_pass3]
+strict_optional = False
```
ilevkivskyi: This PR will cause several conflicts in last strict optional PR https://github.com/python/mypy/pull/4070

They are all however more or less straightforward to fix, so I am OK with merging this first.

By the way, if you are not happy with the ``if var in List[int]: ...`` narrowing that PR https://github.com/python/mypy/pull/4070 introduces, then I can limit it only to narrowing optional to non-optional, which is safe and consistent with existing narrowing optional to non-optional by ``==``. And then later we can consider the more general approach like https://github.com/python/mypy/pull/4072

JukkaL: Thanks! Sorry about the conflicts -- I'll try not to do any complex refactorings until the strict optional PR has been merged.

Just doing the narrowing for optional sounds like a good idea. Non-optional cases seem very rare and they are potentially unsafe. If you can update #4072 soon I should be able to merge it (and #4070) before some further refactorings land.

PR #4101, opened 2017-10-12
Refactor: Move semantic analyzer pass 1 to a new module

JukkaL:
Just move things around, tweak inter-module references and add
necessary imports.

The first pass still directly accesses the second pass so there is
more work to do.


JukkaL:
@ilevkivskyi If this causes tricky merge conflicts with your other PRs I can postpone this as recreating the changes in this PR is easy even if things have changed a lot.

ilevkivskyi:
It looks like this one should not cause additional conflicts at all. Btw, have you tried not adding the ``--strict-optional`` exception? There is a chance the new file will be strict-optional-clean.

JukkaL:
The new file wasn't quite strict-optional clean, but it should be easy enough to fix that.

PR #4112, opened 2017-10-13
Complain about inferring UninhabitedType for a variable

### LINKED ISSUE #4104: None inferred as variable type from UninhabitedType
The inferred type for `x` is unexpected (especially when using `--strict-optional`):

```py
from typing import TypeVar

T = TypeVar('T')

def f() -> T: pass

x = f()
reveal_type(x)  # None
```

It would be better to require an annotation for `x`.

JukkaL:
These don't really make sense as variable types, most of the time.
Also, don't turn them into partial None types, since UninhabitedType
and None no longer have any correspondence.

Fixes #4104.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1890,7 +1890,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             self.set_inferred_type(name, lvalue, init_type)
 
     def infer_partial_type(self, name: Var, lvalue: Lvalue, init_type: Type) -> bool:
-        if isinstance(init_type, (NoneTyp, UninhabitedType)):
+        if isinstance(init_type, NoneTyp):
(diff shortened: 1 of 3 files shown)


JukkaL:
Don't merge yet -- I want to fix an issue in internal Dropbox code first.

ilevkivskyi:
So does this fix the problem with:
```python
def f() -> T: pass

x = f()
reveal_type(x)  # None
```
If yes then maybe it makes sense to add a test for this?

JukkaL:
Yes, there is an existing test case that was skipped (`testUnderspecifiedInferenceResult`).

JukkaL:
Updated the description to mention that this fixes #4104.

ilevkivskyi:
Ah, OK, but then #4059 should be updated after you merge this, since the ``reveal_type`` you added will start working.

Issue #4083, opened 2017-10-09
Refactor semantic analyzer


JukkaL:
The current semantic analyzer is hard to understand and modify. Here some ideas for how to  improve things:

* Create separate modules for the three different passes (say, `mypy.analyzer[123]`).
* Don't pass a `SemanticAnalyzer` instance to the first and third passes. Instead, move any shared functionality to a utility class (or functions), and put them in a separate module such as `mypy.analyzeutil`.
* Rename the passes to, say, `SemanticAnalyzerPass[123]`.

ethanhs:
Should we close #3459 in favor of this? Or close this in favor of that?

ilevkivskyi:
We can keep #3459 to track documenting part, which is not mentioned here, but I leave this up to @JukkaL to decide.

elazarg:
Also
* We can try (again) to extract the creation of namedtuple, typeddict etc. to a separate module
* The analysis of a class can follow the interpreter: first analyzing the body, then passing the results to a "meta" function that constructs the class object. This will be cleaner, and a good plugin point (e.g. for special classes)

ethanhs:
I think this issue could use an update since there has been a lot of progress on this :) I believe we still want to not pass the second pass' instance to the first pass, but are there other concrete steps left in this issue?

JukkaL:
Here's what has been done:

* Semantic analysis passes have been moved to different modules and they have documentation.
* Named tuple, TypedDict, NewType and call-based Enum have been moved to new modules.

The main missing thing is getting rid of direct references between passes.

JukkaL:
Closed in favor of #6204, which has a wider scope.

PR #4114, opened 2017-10-13
Extract typeddict semantic analysis from to a separate class

elazarg:
Related to #4083.

The intention is to do the same thing to namedtuple and enum, then move them to a separate module, refactor common work to helper functions, and hopefully find the right abstraction so it will be easy to add plugins of this kind.


ilevkivskyi:
I think this should happen after last strict-optional PR is merged.

elazarg:
Makes sense.

JukkaL:
Note that I'm working on doing a similar extraction for named tuple. Mine is a little different -- it moves the new class into a new module in a way that there is no import cycle.

elazarg:
This PR can't be maintained.

PR #4115, opened 2017-10-13
Fix regression with empty package names

### LINKED ISSUE #4111: Mypy crashes with `from .. import *` import
In my example at #4110 , if I change the import in `backends/Bla.py` to `from .. import *` and remove references to `blamodule.`, like so: 

I have the following directory structure:
```
blamodule
|-- __init__.py
|-- backends
|   |-- Bla.py
|   `-- __init__.py
`-- call_command.py
```

With the following contents: 
* `blamodule/__init__.py`
```
from . import call_command
from . import backends
```
* `blamodule/backends/__init__.py`
```
from .Bla import Bla
```
* `blamodule/backends/Bla.py`
```
from .. import *
from typing import Iterable

class Bla:
    def __init__(self, path: str) -> None:
        self.path = path

    def method(self, options: Iterable[str]) -> str:
        return call_command.call(["BLA"])
```
* `blamodule/call_command.py`
```
import io

def call() -> str:

    output_buffer = io.StringIO()

    return output_buffer.getvalue()
```

I crash mypy with the following backtrace: 
```
  File "/path/py3env/bin/mypy", line 11, in <module>
    sys.exit(console_entry())
  File "/path/py3env/lib/python3.5/site-packages/mypy/__main__.py", line 7, in console_entry
    main(None)
  File "/path/py3env/lib/python3.5/site-packages/mypy/main.py", line 50, in main
  File "/path/py3env/lib/python3.5/site-packages/mypy/main.py", line 103, in type_check_only
    options=options)
  File "/path/py3env/lib/python3.5/site-packages/mypy/build.py", line 198, in build
  File "/path/py3env/lib/python3.5/site-packages/mypy/build.py", line 1833, in dispatch
    graph = load_graph(sources, manager)
  File "/path/py3env/lib/python3.5/site-packages/mypy/build.py", line 1946, in load_graph
    caller_state=st, caller_line=st.dep_line_map.get(dep, 1))
  File "/path/py3env/lib/python3.5/site-packages/mypy/build.py", line 1466, in __init__
    self.parse_file()
  File "/path/py3env/lib/python3.5/site-packages/mypy/build.py", line 1663, in parse_file
    self.ignore_all or self.options.ignore_errors)
  File "/usr/lib/python3.5/contextlib.py", line 77, in __exit__
    self.gen.throw(type, value, traceback)
  File "/path/py3env/lib/python3.5/site-packages/mypy/build.py", line 1554, in wrap_context
    yield
  File "/path/py3env/lib/python3.5/site-packages/mypy/build.py", line 1663, in parse_file
    self.ignore_all or self.options.ignore_errors)
  File "/path/py3env/lib/python3.5/site-packages/mypy/build.py", line 589, in parse_file
    tree = parse(source, path, self.errors, options=self.options)
  File "/path/py3env/lib/python3.5/site-packages/mypy/parse.py", line 25, in parse
    options=options)
  File "/path/py3env/lib/python3.5/site-packages/mypy/fastparse.py", line 96, in parse
    ).visit(ast)
  File "/path/py3env/lib/python3.5/site-packages/mypy/fastparse.py", line 169, in visit
    return super().visit(node)
  File "/path/py3env/lib/python3.5/site-packages/typed_ast/ast3.py", line 275, in visit
    return visitor(node)
  File "/path/py3env/lib/python3.5/site-packages/mypy/fastparse.py", line 289, in visit_Module
  File "/path/py3env/lib/python3.5/site-packages/mypy/fastparse.py", line 182, in translate_stmt_list
    stmt = self.visit(e)
  File "/path/py3env/lib/python3.5/site-packages/mypy/fastparse.py", line 169, in visit
    return super().visit(node)
  File "/path/py3env/lib/python3.5/site-packages/typed_ast/ast3.py", line 275, in visit
    return visitor(node)
  File "/path/py3env/lib/python3.5/site-packages/mypy/fastparse.py", line 126, in wrapper
    node = f(self, ast)
  File "/path/py3env/lib/python3.5/site-packages/mypy/fastparse.py", line 654, in visit_ImportFrom
    assert n.module is not None
AssertionError: 

```

matzipan:
Fixes #4111

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -651,7 +651,7 @@ class ASTConverter(ast3.NodeTransformer):
     def visit_ImportFrom(self, n: ast3.ImportFrom) -> ImportBase:
         assert n.level is not None
         if len(n.names) == 1 and n.names[0].name == '*':
-            assert n.module is not None
+            assert n.module is not ''
             i = ImportAll(n.module, n.level)  # type: ImportBase
         else:
             i = ImportFrom(self.translate_module_id(n.module) if n.module is not None else '',


mypy/fastparse.py
```diff
@@ -651,7 +651,7 @@ def visit_Import(self, n: ast3.Import) -> Import:
     def visit_ImportFrom(self, n: ast3.ImportFrom) -> ImportBase:
         assert n.level is not None
         if len(n.names) == 1 and n.names[0].name == '*':
-            assert n.module is not None
+            assert n.module is not ''
```
elazarg: Are you sure `is not ''` is the right thing to do? I'd expect `n.module != ''`.

ilevkivskyi: This is actually not what I meant. The fix should at least replace ``None`` with ``''``.

matzipan:
```
if n.module is None:
    n.module = ''
```

Is this what you mean?

ilevkivskyi: Yes, but it may be more subtle than that. ``--strict-optional`` handling requires care, so I need to investigate this myself in any case.

ilevkivskyi: OK, it looks like you can write
```python
        mod = n.module if n.module is not None else ''
        i = ImportAll(mod, n.level)
```
Also please make the corresponding change in ``fastparse2.py`` and add tests (check that the import does not crash in both Python 2 and Python mode).

(I think this is something we should get into the next release, so it is better not to wait with this.)

matzipan:
I shall do it on monday when I arrive at work.

mypy/fastparse.py
ilevkivskyi: @matzipan Is there any progress on this? It is a crash so I don't want the fix to wait indefinitely.

gvanrossum:
Can you work around it by removing the toplevel __init__.py?

gvanrossum:
This needs a test, so other bugs won't cause similar regressions in the future without being caught.

ilevkivskyi:
Superseded by #4259 
@matzipan Thank you for help!

PR #4126, opened 2017-10-17
Update config_file.rst

conanfanli:
This PR adds a note in the documentation to help prevent adding quotes around `mypy_path`.


gvanrossum:
Hm, this is too specific -- all config flags of type 'string' have the same issue. There should a note before the individual flag descriptions explaining that strings should not be quoted.

gvanrossum:
Also MYPYPATH is for more than stubs -- it can be used to combine code in different places.

PR #4130, opened 2017-10-18
Add daemon mode -- highly experimental

gvanrossum:
Many things probably need to change still, but I'd like to get this in before the merge conflicts get out of hand (there are a fair number of small changes to build.py, and 1-2 to main.py).


mypy/main.py
```diff
@@ -28,6 +28,21 @@ class InvalidPackageName(Exception):
     """Exception indicating that a package name was invalid."""
 
 
+orig_stat = os.stat
+
+
+def stat_proxy(path: str) -> os.stat_result:
```
gvanrossum: Hm, we probably don't need this function. (It was an experiment to count stat() calls, and while the numbers are useful, the implementation feels pretty wrong.)

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,308 @@
+"""Type checker test cases"""
```
gvanrossum: This started out as a clone of testcheck.py, but there are many small changes.

test-data/unit/check-incremental.test
```diff
@@ -3020,7 +3020,6 @@ tmp/mod.py:7: error: Revealed type is 'builtins.bytes'
 # cmd: mypy -m a
 # cmd2: mypy -m b
 # flags: --follow-imports=silent
-# flags2: --follow-imports=silent
```
gvanrossum: This was just redundant, and its presence thwarted my logic for skipping test that vary flags across runs.

setup.py
```diff
@@ -103,11 +103,15 @@ def run(self):
...
-                                        'stubgen=mypy.stubgen:main']},
+                                        'stubgen=mypy.stubgen:main',
+                                        'dmypy=mypy.dmypy:main',
+                                        ]},
       data_files=data_files,
       classifiers=classifiers,
       cmdclass={'build_py': CustomPythonBuild},
-      install_requires = ['typed-ast >= 1.1.0, < 1.2.0'],
+      install_requires = ['typed-ast >= 1.1.0, < 1.2.0',
+                          'psutil >= 5.4.0, < 5.5.0',
```
gvanrossum: This is used by dmypy only, to report memory usage.

scripts/dmypy
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python3
```
gvanrossum: Cloned from scripts/stubgen.

ethanhs: You might want to also add a entry_point in setup.py if you want something run from an installed mypy distribution.

gvanrossum: It's there: https://github.com/python/mypy/pull/4130/files#diff-2eeaed663bd0d25b7e608891384b7298R107

ethanhs: Ah! Sorry, I missed that.

mypy/test/testdmypy.py
elazarg: I'd really like to be able to refactor the duplications in the test files in order to eventually clean up the test logic and improve the usability (#3973 being a first step). A comment about the nature of these differences may help with that.

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,310 @@
...
+    'check-incremental.test',
+    'check-newtype.test',
+]
+
+
+class TypeCheckSuite(DataSuite):
+
+    @classmethod
+    def cases(cls) -> List[DataDrivenTestCase]:
+        if sys.platform == 'win32':
```
gvanrossum: This is new (dmypy is not supported on Windows, and doesn't work there).

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,310 @@
...
+            return []  # Nothing here works on Windows.
+        c = []  # type: List[DataDrivenTestCase]
+        for f in files:
+            tc = parse_test_cases(os.path.join(test_data_prefix, f),
+                                  None, test_temp_dir, True)
+            c += [case for case in tc
+                  if cls.has_stable_flags(case) and cls.is_incremental(case)]
+        return c
+
+    def run_case(self, testcase: DataDrivenTestCase) -> None:
```
gvanrossum: This method is specialized for incremental==True.

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,310 @@
...
+    @classmethod
+    def cases(cls) -> List[DataDrivenTestCase]:
+        if sys.platform == 'win32':
+            return []  # Nothing here works on Windows.
+        c = []  # type: List[DataDrivenTestCase]
+        for f in files:
+            tc = parse_test_cases(os.path.join(test_data_prefix, f),
+                                  None, test_temp_dir, True)
+            c += [case for case in tc
+                  if cls.has_stable_flags(case) and cls.is_incremental(case)]
```
gvanrossum: These filters were added.

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,310 @@
...
+        num_steps = max([2] + list(testcase.output2.keys()))
+        # Check that there are no file changes beyond the last run (they would be ignored).
+        for dn, dirs, files in os.walk(os.curdir):
+            for file in files:
+                m = re.search(r'\.([2-9])$', file)
+                if m and int(m.group(1)) > num_steps:
+                    raise ValueError(
+                        'Output file {} exists though test case only has {} runs'.format(
+                            file, num_steps))
+        self.server = None  # type: Optional[dmypy.Server]
```
gvanrossum: Reset the server here.

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,310 @@
...
+        for step in range(1, num_steps + 1):
+            self.run_case_once(testcase, step)
+
+    @classmethod
```
gvanrossum: Next two methods are new.

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,310 @@
...
+            if os.path.basename(filename).startswith('mypy.ini.'):
+                return False
+        return True
+
+    def clear_cache(self) -> None:
+        dn = defaults.CACHE_DIR
+        if os.path.exists(dn):
+            shutil.rmtree(dn)
+
+    def run_case_once(self, testcase: DataDrivenTestCase, incremental_step: int) -> None:
```
gvanrossum: No default for incremental_step, and assert it's >= 1 below.

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,310 @@
...
+                        new_time = os.stat(target).st_mtime + 1
+                        os.utime(target, times=(new_time, new_time))
+            # Delete files scheduled to be deleted in [delete <path>.num] sections.
+            for path in testcase.deleted_paths.get(incremental_step, set()):
+                # Use retries to work around potential flakiness on Windows (AppVeyor).
+                retry_on_error(lambda: os.remove(path))
+
+        # Parse options after moving files (in case mypy.ini is being moved).
+        options = self.parse_options(original_program_text, testcase, incremental_step)
+        if incremental_step == 1:
```
gvanrossum: Initialize server here.

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,310 @@
...
+            self.server = dmypy.Server([])  # TODO: Fix ugly API
+            self.server.options = options
+
+        sources = []
+        for module_name, program_path, program_text in module_data:
+            # Always set to none so we're forced to reread the module in incremental mode
+            sources.append(build.BuildSource(program_path, module_name, None))
+        response = self.server.check(sources, alt_lib_path=test_temp_dir)
```
gvanrossum: Call server.check() instead of build.build().

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,310 @@
...
+                incremental_step) + ' ({}, line {})')
+            output = testcase.output2.get(incremental_step, [])
+        else:
+            raise AssertionError()
+
+        if output != a and self.update_data:
+            update_testcase_output(testcase, a)
+        assert_string_arrays_equal(output, a, msg.format(testcase.file, testcase.line))
+
+        manager = self.server.last_manager
```
gvanrossum: Get manager from server instead of from BuildResult.

mypy/test/testdmypy.py
```diff
@@ -0,0 +1,310 @@
...
+                raise RuntimeError('Specifying targets via the flags pragma is not supported.')
+        else:
+            options = Options()
+
+        # Allow custom python version to override testcase_pyversion
+        if (not flag_list or
+                all(flag not in flag_list for flag in ['--python-version', '-2', '--py2'])):
+            options.python_version = testcase_pyversion(testcase.file, testcase.name)
+
+        options.use_builtins_fixtures = True
```
gvanrossum: Hardcoding these flags here (rather than way up there).

mypy/test/testdmypy.py
elazarg: Thanks.

mypy/dmypy.py
```diff
@@ -0,0 +1,585 @@
...
+import sys
+import time
+
+from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Tuple, TypeVar
+
+import psutil  # type: ignore  # It's not in typeshed yet
+
+import mypy.build
+import mypy.errors
+import mypy.main
```
JukkaL: Eventually we could import most mypy modules lazily so that the client would run quicker in case the server is already running.

mypy/dmypy.py
```diff
@@ -0,0 +1,585 @@
...
+    This is where mypy flags are set.  Setting flags is a bit awkward;
+    you have to use e.g.:
+
+      dmypy start -- --strict
+
+    since we don't want to duplicate mypy's huge list of flags.
+    """
+    try:
+        pid, sockname = get_status()
+    except SystemExit as err:
```
JukkaL: I'd prefer having a separate exception class for daemon errors, and a top-level try/except that will translate it to `SystemExit`.

mypy/dmypy.py
```diff
@@ -0,0 +1,585 @@
...
+        if not self.last_sources:
+            return {'error': "Command 'recheck' is only valid after a 'check' command"}
+        return self.check(self.last_sources)
+
+    last_mananager = None  # type: Optional[mypy.build.BuildManager]
+
+    def check(self, sources: List[mypy.build.BuildSource],
+              alt_lib_path: Optional[str] = None) -> Dict[str, Any]:
+        bound_gc_callback = lambda phase, info: self.gc_callback(phase, info)
+        self.gc_start_time = None  # type: Optional[float]
```
JukkaL: I'd recommend moving the stats handling code to separate methods to keep the logic of this method less cluttered.

JukkaL:
I encountered something that looks like a performance problem. I did these steps:

* Remove mypy cache.
* Create a tiny new file `mypy/x.py` with a single function that generates a type check error.
* Run `scripts/dmypy check mypy` (takes about 13 seconds). It generates an error as expected.
* Run `scripts/dmypy recheck`. It still takes about 13 seconds but I'd expect it to only take a fraction of a second, since only `mypy.x` needs to be processed.

gvanrossum:
The performance problem is apparently not new -- I can reproduce this with standard `-i` mode as well. No cache entries for the mypy package are written somehow -- perhaps because mypy.x precedes all other submodules of mypy. This could be due to an edge case in the topsort.

JukkaL:
In that case I'll create a new issue to track it.

JukkaL:
Created #4135 to track the performance issue.

mypy/dmypy.py
```diff
@@ -0,0 +1,585 @@
...
+        return sock
+
+    def run_command(self, command: str, data: Mapping[str, object]) -> Dict[str, object]:
+        """Run a specific command from the registry."""
+        key = 'cmd_' + command
+        method = getattr(self.__class__, key, None)
+        if method is None:
+            return {'error': "Unrecognized command '%s'" % command}
+        else:
+            return method(self, **data)
```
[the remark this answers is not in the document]
JukkaL: This is still pending, but this can be merged anyway.

mypy/build.py
```diff
@@ -339,6 +350,27 @@ def default_lib_path(data_dir: str,
 # silent mode or simply not found.
 
 
+def cache_meta_from_dict(meta: Dict[str, Any], data_json: str) -> CacheMeta:
+    sentinel = None  # type: Any  # the values will be post-validated below
```
ilevkivskyi: This comment should be now "validated by the caller" or similar.

Issue #4133, opened 2017-10-19
Callable names contains quotation marks


elazarg:
`CallableType.name` is initialized to be something like `"foo"` instead of simply `foo`. I believe this is a mistake, tying the type to the representation of error messages and makes the error messages harder to tune.

gvanrossum:
I've noticed this too, it can be a btt confusing. Cleaning it up may be complex though.

PR #4136, opened 2017-10-19
Encapsulate callable name access in messages.py

elazarg:
Trying to solve #4133 in two steps, in order to avoid conflicts on `semanal.py`. This step does not do much by itself - it replaces `callee.name` with `callable_name(callee)`. In later steps, the quotation marks can be added inside `callable_name` (which is a formatting function) instead of being part of `CallableType`.

As a side effect error messages are slightly more uniform, and `Function does not return a value` is replaced with `Expression evaluates to "None"` when the expression is not a function.

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -512,7 +512,7 @@ class MessageBuilder:
         return AnyType(TypeOfAny.from_error)
 
     def untyped_function_call(self, callee: CallableType, context: Context) -> Type:
-        name = callee.name if callee.name is not None else '(unknown)'
+        name = callable_name(callee) or '(unknown)'
         self.fail('Call to untyped function {} in typed context'.format(name), context)
         return AnyType(TypeOfAny.from_error)
 
@@ -526,8 +526,9 @@ class MessageBuilder:
         operator name in the messages.
         """
         target = ''
-        if callee.name:
-            name = callee.name
+        callee_name = callable_name(callee)
+        if callee_name is not None:
+            name = callee_name
             if callee.bound_args and callee.bound_args[0] is not None:
                 base = self.format(callee.bound_args[0])
             else:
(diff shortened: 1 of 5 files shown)


test-data/unit/check-expressions.test
```diff
@@ -1525,9 +1525,7 @@ from typing import Iterator
 def f() -> Iterator[int]:
     yield 5
 def g() -> Iterator[int]:
-    a = yield from f()
-[out]
-main:5: error: Function does not return a value
+    a = yield from f()  # E: Expression evaluates to "None"
```
ilevkivskyi: I don't think this message is better, it is still not clear what is the problematic expression. With previous wording, it may be better to add a function name.

elazarg: But it's not necessarily a named function. I changed it to `Expression "yield from <iterator>" does not return a value`. Is that better?

mypy/messages.py
```diff
@@ -656,61 +657,53 @@ def too_few_arguments(self, callee: CallableType, context: Context,
...
-        else:
-            self.fail('Function does not return a value', context)
+        name = callable_name(callee_type) or 'Function'
+        self.fail('{} does not return a value'.format(name), context)
+
+    def evaluates_to_none(self, context: Context) -> None:
+        self.fail('Expression evaluates to "None"', context)
+
+    def yield_from_iterator(self, context: Context) -> None:
+        self.fail('Expression "yield from <iterator>" does not return a value', context)
```
ilevkivskyi: This part is not directly related to the PR as I understand. So that it is better to make a separate PR/issue.
These error messages deserve a separate discussion.

elazarg: I thought I needed this change, but turns out you're right. Reverted.

(Although I do like the new messages better :) )

ilevkivskyi:
@elazarg OK, I let us make the second step you mentioned.

PR #4149, opened 2017-10-22
Fix broken installation on Windows in virtualenv

### LINKED ISSUE #4142: Virtualenv install broken on Windows
It seems that virtualenvs do not have a userbase prefix, so my recent changes in #4005 broke this. First reported [here](https://github.com/python/mypy/pull/4005#issuecomment-337577401).

AleksanderGondek:
Fixes #4142 
Python `site.py` from any virtualenv does not contain many functions added in python 2.7 (see: https://github.com/pypa/virtualenv/issues/355), which results in method `site.getsitepackages` being absent.

This commit replaces getsitepackages call with equivalent of `distutils.sysconfig.get_python_lib` which is supported in virtualenv.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -13,6 +13,7 @@ The function build() is the main interface to this module.
 import binascii
 import collections
 import contextlib
+from distutils.sysconfig import get_python_lib
 import gc
 import hashlib
 import json
@@ -229,7 +230,12 @@ def default_data_dir(bin_dir: Optional[str]) -> str:
     """
     if not bin_dir:
         if os.name == 'nt':
-            prefixes = [os.path.join(sys.prefix, 'Lib'), os.path.join(site.getuserbase(), 'lib')]
+            prefixes = [os.path.join(sys.prefix, 'Lib')]
+            try:
+                prefixes.append(os.path.join(site.getuserbase(), 'lib'))
+            except AttributeError:
+                # getuserbase in not available in virtualenvs
+                prefixes.append(os.path.join(get_python_lib(), 'lib'))
             for parent in prefixes:
                     data_dir = os.path.join(parent, 'mypy')
                     if os.path.exists(data_dir):


gvanrossum:
@ethanhs Can you review/test this?

mypy/build.py
```diff
@@ -229,7 +229,7 @@ def default_data_dir(bin_dir: Optional[str]) -> str:
     """
     if not bin_dir:
         if os.name == 'nt':
-            prefixes = [os.path.join(sys.prefix, 'Lib'), os.path.join(site.getuserbase(), 'lib')]
+            prefixes = [os.path.join(sys.prefix, 'Lib'), os.path.join(get_python_lib(), 'lib')]
```
ethanhs: Sadly, this will not work on `--user` installs. It might be better to try`site.getuserbase()`, then fall back on sysconfig's get_python_lib, so it is only invoked in virtualenvs.

AleksanderGondek: That is too bad.
Do you think that the following solution could do the trick?
``` python
getattr(site, 'getuserbase', None)
prefixes = [
    os.path.join(sys.prefix, 'Lib'), 
    os.path.join(getattr(site, 'getuserbase', get_python_lib)(), 'lib')
]
```

ethanhs: Hm, I think the getattr call, while concise, is a bit opaque and not obvious. I think
```python
prefixes = [os.path.join(sys.prefix, 'Lib')]
try:
    prefixes.append(os.path.join(site.getuserbase(), 'lib')
except:
    prefixes.append(os.path.join(get_python_lib(), 'lib')
``` 
or perhaps

```python
prefixes = [os.path.join(sys.prefix, 'Lib')]
if hasattr(site, 'getuserbase'):
    prefixes.append(os.path.join(site.getuserbase(), 'lib')
else:
    prefixes.append(os.path.join(get_python_lib(), 'lib')
```

would be much more readable.

AleksanderGondek: Your suggestion looks way more readable, thanks!
Will implement it *soonTM* (in about 12 hours).

ethanhs: Also, if you could, explain _why_ we fall back on `get_python_lib` please. eg in the except/else branch
`# getuserbase not available in virtualenvs`

Thanks so much for fixing this!

AleksanderGondek: No problem, glad to help and thanks for your quick feedback!

AleksanderGondek:
Ping

gvanrossum:
Ping what?

AleksanderGondek:
I was just wondering if we could merge the change - I have adjusted it accordingly to ethanhs advice. 
Sorry if this is being to pushy, but it pains me to not have mypy working on Windows in virtualenv :)

gvanrossum:
We're all volunteers here -- Ethan will get to it when he gets to it, and
it will make it into the release that follows. (So yes, a bit too pushy.)

AleksanderGondek:
Apologies - I have not intended to be disrespectful. I fully appreciate free time put into the project. I will now cease to spam and thanks for clarifying things for me .

Issue #4036, opened 2017-09-29
Drop support for Python 3.3


ethanhs:
Python 3.3's End of Life is today. [relevant docs](https://docs.python.org/devguide/index.html#branchstatus). I believe in the past there was at least general consensus that we should drop support around the time it is EOL'd.

With an impending release, I believe it is the perfect time to stop supporting 3.3, however, we may want to support it for the 0.530 series, since the release is so soon.

Thoughts?

gvanrossum:
Sounds like we should do this once 0.530 is released.

elazarg:
(When it's done: #1243)

deathaxe:
There are still some tools out there which use python3.3 and can't update soon. Dropping python 3.3 support would render this tool useless for developers contributing to such tools.

Sublime Text is one famous example.

JukkaL:
@deathaxe It's possible to have multiple versions of Python 3 installed at the same time, so running mypy should only be slightly more complicated if users are stuck with Python 3.3. Do you think that the extra hassle of needing multiple installed Python 3 versions (and perhaps using virtualenv) is enough to alienate some users?

I agree that some users may want to continue using Python 3.3. Let's announce our intention to drop Python 3.3 in the release notes of 0.530 and link to this issue. If many users are worried about this, we can reconsider and keep 3.3 supported for now.

@elazarg I don't think that I want #1243 after all. I'll elaborate in #1243.

deathaxe:
The question depends on what "dropping" means. I understand it as dropping support for checking legacy features/constructs which got obsolete in newer python versions.

If dropping targets the required python version to run mypy only, it's not that critical. Running different versions of python is no problem. Many people use virtualenv for isolated environments anyway.

gvanrossum:
The intent as I understand it is to require at least 3.4 to run mypy. It can still analyze 3.2 and 3.3 code.

JukkaL:
Even if mypy didn't officially support type checking 3.3 code, the differences between 3.3 and 3.4 are sufficiently minor that using the 3.4 target mode should be fine for most users. However, I don't see why we'd need to drop 3.3 as a type checking target, since the cost of maintaining it is minor -- mainly there may be some extra complexity in certain typeshed stubs.

ethanhs:
Yes, the intention has always been to drop support for running mypy on 3.3. It should continue to analyze 3.3 code however, since 3.3 code should be valid 3.4 code.

hugovk:
[...]

PR #4152, opened 2017-10-23
Running on Python 3.3 is no longer supported

hugovk:
For https://github.com/python/mypy/issues/4036.

### DIFF
--- a/docs/source/revision_history.rst
+++ b/docs/source/revision_history.rst
@@ -3,6 +3,12 @@ Revision history
 
 List of major changes:
 
+- Upcoming
+    * Publish ``mypy`` version 0.550 on PyPI.
+
+      * We soon intend to drop support for Python 3.3, which has reached end-of-life. Please comment on `#4036 <https://github.com/python/mypy/issues/4036>`_ if this is an issue.
+
+
 - October 2017
     * Publish ``mypy`` version 0.540 on PyPI.
 
--- a/docs/source/revision_history.rst
+++ b/docs/source/revision_history.rst
@@ -6,8 +6,7 @@ List of major changes:
 - Upcoming
     * Publish ``mypy`` version 0.550 on PyPI.
 
-      * We soon intend to drop support for Python 3.3, which has reached end-of-life. Please comment on `#4036 <https://github.com/python/mypy/issues/4036>`_ if this is an issue.
-
+      * Running mypy now requires Python 3.4 or higher. However Python 3.2 and 3.3 are still valid for the target of the analysis (i.e. the ``--python-version`` flag).
 
 - October 2017
     * Publish ``mypy`` version 0.540 on PyPI.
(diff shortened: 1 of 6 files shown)


docs/source/revision_history.rst
```diff
@@ -3,6 +3,12 @@ Revision history
 
 List of major changes:
 
+- Upcoming
```
gvanrossum: I like this -- it makes it easier to collect everything relevant to the next release in one place (rather than having to scour the logs a day before the release).

docs/source/revision_history.rst
```diff
@@ -3,6 +3,12 @@ Revision history
...
+    * Publish ``mypy`` version 0.550 on PyPI.
+
+      * We soon intend to drop support for Python 3.3, which has reached end-of-life. Please comment on `#4036 <https://github.com/python/mypy/issues/4036>`_ if this is an issue.
```
gvanrossum: This should just be "Running mypy now requires Python 3.4 or higher." Perhaps also add "However Python 3.2 and 3.3 are still valid for the target of the analysis (i.e. the `--python-version` flag)."

We already announced we would drop 3.3 support for running mypy in the blog post for 0.530 -- we just never followed through. FWIW the text there was:

> Note that future mypy versions will still support _checking_ older Python versions — they just won’t _run_ on 3.3 or earlier. It’s easy to have multiple Python versions installed, so this won’t prevent anyone from running mypy.

hugovk: Updated!

docs/source/revision_history.rst
hugovk: We do something similar for Pillow: https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst

docs/source/revision_history.rst
```diff
@@ -3,6 +3,11 @@ Revision history
 
 List of major changes:
 
+- Upcoming
+    * Publish ``mypy`` version 0.550 on PyPI.
+
+      * Running mypy now requires Python 3.4 or higher. However Python 3.2 and 3.3 are still valid for the target of the analysis (i.e. the ``--python-version`` flag).
```
gvanrossum: Can you break this long line? We don't like lines >= 80 characters in our docs; it makes reading on small screens or in side-by-side diffs cumbersome.

gvanrossum: Also to my embarrassment we actually *don't* support `--python-version 3.2`. You get
```
mypy: error: argument --python-version: Python 3.2 is not supported (must be 3.3 or higher)
```
However using 3.3 as a target is still supported.

hugovk: Oops! Updated!

hugovk:
.travis.yml and setup.py updated here, along with some other documentation files. I expect there's some more code changes possible, but I think they can be in another PR.

gvanrossum:
Great! Thanks for finding those other occurrences. The setup.py one was particularly embarrassing -- it would allow installing with 3.2 but then could not type-check 3.2. :-)

PR #4156, opened 2017-10-24
Clarify documentation about TypedDicts.

garetht:
This notes that `TypedDict`s are now officially supported and clarifies what is required to use the inheritance syntax for `TypedDict`s.


gvanrossum:
Thanks!

PR #4160, opened 2017-10-25
Make stubgen return status 0 on --help

gwerbin:
This lets package managers like [conda](https://conda.io/docs/) use `stubgen --help` as a quick-and-dirty test.

Also it makes more sense if you squint at it a little. The user _requested_ help, so they shouldn't be put in some an error state.


gvanrossum:
Can you address the lint error? Otherwise LG -- agreed that `--help` should
return exit status 0.

gvanrossum:
Thanks!

PR #4164, opened 2017-10-26
Fix classmethod access for TupleType

### LINKED ISSUE #4155: Item "type" of "Union[Type[Foo], Type[Bar]]" has no attribute "x"
```
# mypy 0.530
foo.py:14: warning: Returning Any from function declared to return "int"
foo.py:14: error: Item "type" of "Union[Type[Foo], Type[Bar]]" has no attribute "x"
```

```python
from typing import *

class Foo(NamedTuple):
    a: int
    @classmethod
    def x(cls) -> int: return 1

class Bar(NamedTuple):
    a: int
    @classmethod
    def x(cls) -> int: return 1

def foo(t: Type[Union[Foo, Bar]]) -> int:
    return t.x()
```

If I make them usual classes instead of namedtuples, it works

elazarg:
Fix #4155

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -191,6 +191,8 @@ def analyze_member_access(name: str,
         elif isinstance(typ.item, TypeVarType):
             if isinstance(typ.item.upper_bound, Instance):
                 item = typ.item.upper_bound
+        elif isinstance(typ.item, TupleType):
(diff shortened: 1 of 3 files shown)


gvanrossum:
Can you also add a separate test for NamedTuple? (Who knows if they actually use completely the same path in every case...)

Issue #4167, opened 2017-10-26
Properties, `setter`, and `Incompatible types in assignment`


nnadeau:
# Expected Behaviour
- Ability to use property `setter` with `Union[]`
- The intention is to overload the `setter`

# Actual Behaviour & Repro Case
## Python Code
```python
from typing import Union


class A:
    def __init__(self) -> None:
        self._foo = 0

    @property
    def foo(self) -> int:
        return self._foo

    @foo.setter
    def foo(self, value: Union[int, float]) -> None:
        self._foo = int(value)


if __name__ == '__main__':
    a = A()
    print(a.foo)

    a.foo = 123
    print(a.foo)

    a.foo = 456.789 # error thrown here
    print(a.foo)
```

## Stdout
```
0
123
456
```
(i.e., valid Python code)

## Mypy CLI
```
mypy --strict
```

## Result
```
error: Incompatible types in assignment (expression has type "float", variable has type "int")
```

# Notes
- `Mypy` seems to be type checking the assignment using the type declared in `def foo(self) -> int:` instead of the declared types in the `setter`: ` def foo(self, value: Union[int, float]) -> None:`

# Current Solution
- Silence the error with `# type: ignore`

gvanrossum:
Your hunch is correct: for properties, mypy uses the type declared in the initial `@property` definition as the final type for the attribute, regardless of what the `@property.setter` does.

I suppose we could fix this though it might require a bunch of one-off changes in odd corners of the code since properties are unique in this respect.

srittau:
This seems similar to issue #3004, which has more discussion.

gvanrossum:
Oh, yeah, it's a duplicate.

PR #4176, opened 2017-10-27
Properties & Incompatible Types in Assignment

nnadeau:
Documentation for https://github.com/python/mypy/issues/4167

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -476,37 +476,33 @@ Here's the above example modified to use ``MYPY``:
 Properties & Incompatible Types in Assignment
 ---------------------------------------------
 
-Consider the code below:
+Mypy infers the type of a property from the getter method (e.g., ``int``) as
+the final type for the attribute, regardless of what the ``@property.setter``
+declares (e.g., ``Union[int, float]``), as demonstrated below:
 
 .. code-block:: python
 
     from typing import Union
-    
-    
+
+
     class A:
-        def __init__(self) -> None:
-            self._foo = 0
-    
         @property
         def foo(self) -> int:
-            return self._foo
-    
+            ...
+
         @foo.setter
         def foo(self, value: Union[int, float]) -> None:
-            self._foo = int(value)
-    
-    
-    if __name__ == '__main__':
-        a = A()
-        a.foo = 123
-        a.foo = 456.789
-    
+            ...
 
-Running Mypy against the above code will generate the following error at ``a.foo = 456.789``: 
+
+    a = A()
+    a.foo = 123
+    a.foo = 456.789 # error here
+
+
+Running mypy against the above code will generate the following error:
 
 .. code-block:: none
-    
-    error: Incompatible types in assignment (expression has type "float", variable has type "int")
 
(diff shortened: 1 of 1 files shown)


docs/source/common_issues.rst
```diff
@@ -472,3 +472,41 @@ Here's the above example modified to use ``MYPY``:
 
    def listify(arg: 'bar.BarClass') -> 'List[bar.BarClass]':
        return [arg]
+
+Properties & Incompatible Types in Assignment
+---------------------------------------------
+
+Consider the code below:
```
gvanrossum: I think the whole thing is a bit too verbose. You should start with something like "The type of a property is taken from the getter method." Then the example should be more compact (compare some of the other sections in this file). The method bodies could just be `...` and the "main" section shouldn't have `if __name__...` .

docs/source/common_issues.rst
```diff
@@ -472,3 +472,41 @@ Here's the above example modified to use ``MYPY``:
...
+    
+        @foo.setter
+        def foo(self, value: Union[int, float]) -> None:
+            self._foo = int(value)
+    
+    
+    if __name__ == '__main__':
+        a = A()
+        a.foo = 123
+        a.foo = 456.789
```
gvanrossum: Add a comment here like "# error here".

docs/source/common_issues.rst
```diff
@@ -472,3 +472,41 @@ Here's the above example modified to use ``MYPY``:
...
+    
+
+Running Mypy against the above code will generate the following error at ``a.foo = 456.789``: 
+
+.. code-block:: none
+    
+    error: Incompatible types in assignment (expression has type "float", variable has type "int")
+
+While overloading class property setters in Python is valid, Mypy infers the type from the initial ``@property`` definition (e.g., ``int``) as the final type for the attribute, regardless of what the ``@property.setter`` declares (e.g., ``Union[int, float]``). This error may be silenced with `# type: ignore  <http://mypy.readthedocs.io/en/latest/common_issues.html#spurious-errors-and-locally-silencing-the-checker>`_
```
gvanrossum: Please break this line so each line is under 80 characters, and remove the remark about silencing it with `# type: ignore` -- by the time a user gets here they should know they can silence anything that way.

Also, "mypy" is not capitalized except at the start of a sentence.

ilevkivskyi:
I thought the section "Common issues" is not about _known issues_ (i.e. bugs), but for things that are either correct but unintuitive for some developers (like invariants of ``List[T]``), or are some fundamental limitations (like import cycles). Most sections have motivation and/or workarounds.

Do we really need this? It looks just like a duplicate of the issue.

nnadeau:
@ilevkivskyi , https://github.com/python/mypy/issues/4167 is currently being considered as a low-priority feature request and not a bug. IMHO, I think this documentation for the current state of mypy is useful for others who may stumble upon issues with properties, as it may change their approach to class architecture.

gvanrossum:
Hm, I agree with Ivan that the bar for inclusion of something in the common_issues page should be higher than "low priority feature request" (else half of the issues in the tracker would deserve inclusion here). So I'll close the PR now without merging. If we see an uptick in reports of users who are confused by this we can reconsider.

PR #4178, opened 2017-10-29
Replace `--disallow-any` flags with separate boolean flags.

### LINKED ISSUE #4089: Replace `--disallow-any` flags with separate boolean flags.
The `--disallow-any` flags interact in unexpected ways with other flags and they make configuration more complicated. We should replace them with a set of simple boolean flags. See discussion in #4076.

ilinum:
This PR also gets rid of `--disallow-any=unannotated` because it
is the same thing as `--disallow-untyped-defs`.

Fix #4089


mypy/main.py
```diff
@@ -272,10 +254,18 @@ def add_invertible_flag(flag: str,
...
     parser.add_argument('--follow-imports', choices=['normal', 'silent', 'skip', 'error'],
                         default='normal', help="how to treat imports (default normal)")
-    parser.add_argument('--disallow-any', type=disallow_any_argument_type, default=[],
-                        metavar='{{{}}}'.format(', '.join(disallow_any_options)),
-                        help="disallow various types of Any in a module. Takes a comma-separated "
-                             "list of options (defaults to all options disabled)")
+    parser.add_argument('--disallow-any-unimported', default=False, action='store_true',
+                        help="disallow usage of types that come from unfollowed imports")
+    parser.add_argument('--disallow-any-expr', default=False, action='store_true',
+                        help='disallow all expressions in the module that have type Any')
```
gvanrossum: I'd drop "in the module" since that's implicit (and you don't say it for the other flags).

mypy/main.py
```diff
@@ -272,10 +254,18 @@ def add_invertible_flag(flag: str,
                         help="silently ignore imports of missing modules")
     parser.add_argument('--follow-imports', choices=['normal', 'silent', 'skip', 'error'],
                         default='normal', help="how to treat imports (default normal)")
-    parser.add_argument('--disallow-any', type=disallow_any_argument_type, default=[],
-                        metavar='{{{}}}'.format(', '.join(disallow_any_options)),
-                        help="disallow various types of Any in a module. Takes a comma-separated "
-                             "list of options (defaults to all options disabled)")
+    parser.add_argument('--disallow-any-unimported', default=False, action='store_true',
+                        help="disallow usage of types that come from unfollowed imports")
```
gvanrossum: Maybe "disallow Any types resulting from unfollowed imports" ?

ilinum: Yep, this seems reasonable! Done!

mypy/main.py
ilinum: Yep!

gvanrossum:
Thanks! Now I have a whole bunch of internal mypy.ini files to clean up...

PR #4207, opened 2017-11-04
Move quoting of function names to messages.py

### LINKED ISSUE #4133: Callable names contains quotation marks

elazarg:
Fix #4133 (continuing #4136).

Also fix naming of namedtuple methods (to which I believe I am responsible :| ).

(I don't like the use of "method of T" as part of the name, but this PR doesn't change that).

I wonder if `callable_type()` shouldn't be using `set_callable_name()` to keep naming consistent.


elazarg:
@JukkaL note that this PR touches namedtuple creation.

gvanrossum:
OK, since there are no merge conflicts yet I'm merging this. It won't make it into the 0.550 release.

PR #4213, opened 2017-11-06
Improve error message for an invalid callable type

li-dan:
If the user gives the wrong number of arguments to Callable[], tell them
how many arguments are expected.

### DIFF
--- a/test-data/unit/check-generics.test
+++ b/test-data/unit/check-generics.test
@@ -812,7 +812,7 @@ class Node(Generic[T]):
     def __init__(self, x: T) -> None:
         ...
 
-BadC = Callable[T] # E: Invalid function type
+BadC = Callable[T] # E: Please use "Callable[[<parameters>], <return type>]" or "Callable"
 
 C = Callable[..., T]
 C2 = Callable[[T, T], Node[T]]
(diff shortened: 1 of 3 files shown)


test-data/unit/check-generics.test
```diff
@@ -812,7 +812,7 @@ class Node(Generic[T]):
     def __init__(self, x: T) -> None:
         ...
 
-BadC = Callable[T] # E: Invalid function type
+BadC = Callable[T] # E: Callable takes 0 or 2 arguments, not 1
```
elazarg: This error is ambiguous, since it looks like it is about the callable itself, not the the type-constructor "Callable". At least quote it: `"Callable" takes 0 or 2 arguments, not 1`.

Additionally, I think it's better not to talk about "number of arguments" - again, can be confused with function arguments. Instead it might be more obvious if the message demonstrated the two possibilities. Something like `Please use "Callable[[<parameters]>, <return type>]`.

What is the "0 arguments" case? The `...` form?

JelleZijlstra: I guess the "0 arguments" case is plain `Callable` with omitted arguments. I agree with you that the message is confusing.

PR #4219, opened 2017-11-07
add back options in mypy_self_check.ini

JelleZijlstra:
Followup from #4217 and python/typeshed#1664.


gvanrossum:
I synced typeshed, can you rebase?

JelleZijlstra:
Done.

I had to change a few types to "Any" because of `disallow_any_unimported`. An alternative could be to remove that option, because I don't think it adds much if we already have an error for imported modules without stubs.

PR #4228, opened 2017-11-09
Partial support for import cycles in fine-grained incremental mode

JukkaL:
When processing a module within an import cycle in fine-grained 
incremental mode, we have two versions of the AST in memory for a while.
Previously it was possible that we used stale references that referred to 
the old AST. This fixes many of these issues, with the goal of only having
non-stale references after each increment.

This makes three related changes to improve this:

* When looking up names during semantic analysis, rebind references to 
  the old AST to point to the new AST instead.
* Merge old and new ASTs after semantic analysis instead of after type 
  checking so that type checking won't have to worry about stale 
  references.
* Rewrite symbol table diff implementation to take snapshots of symbol 
  tables and compare them, instead of directly comparing symbol tables.
  This is needed since we now do the AST merge earlier, and when we
  calculate the diff, the old symbol table has already been invalidated by
  the AST merge and can't be used any more.

This doesn't fully resolve the issue, since a bunch of scenarios are not yet
tested. I'll refine this later, but getting everything right will require 
improvements in other parts of the implementation and I don't want to make 
this PR much larger.


mypy/semanal.py
```diff
@@ -3638,9 +3653,26 @@ def lookup_qualified(self, name: str, ctx: Context,
...
 
+    def rebind_symbol_table_node(self, n: SymbolTableNode) -> Optional[SymbolTableNode]:
+        """If node refers to old version of module, return reference to new version.
+
+        If the reference is removed in the new version, return None.
+        """
+        # TODO: Handle type aliases, type variables and other sorts of references
+        if isinstance(n.node, (FuncDef, OverloadedFuncDef, TypeInfo, Var)):
+            # TODO: Why is it possible for fullname() to be None, even though it's not
+            #   annotated as Optional[str]?
```
ilevkivskyi: I think this is because of pre-PEP 526 limitations for ``_fullname`` (note underscore) and the fact that ``_fullname`` is not set in ``__init__`` but later (first vs second pass IIRC). We already had few crashes because of this. I fixed some of them but maybe some nodes still don't have full names generated in second pass.

mypy/semanal.py
```diff
@@ -3638,9 +3653,26 @@ def lookup_qualified(self, name: str, ctx: Context,
...
+            if n.node.fullname() and get_prefix(n.node.fullname()) == self.cur_mod_id:
```
gvanrossum: Does this need the same TODO as above on line 3642? ("Do this for all modules in the set of modified files.")

Also what if it's inside a class or other namespace? get_prefix() only strips one component from the end. Maybe we need a utility to compare a name with a prefix (basically `name.startswith(prefix + '.')`).

mypy/semanal.py
```diff
@@ -3652,8 +3684,9 @@ def lookup_fully_qualified(self, name: str) -> SymbolTableNode:
 
         Assume that the name is defined. This happens in the global namespace -- the local
         module namespace is ignored.
+
+        NOTE: This is only supported for names known to be defined.
```
gvanrossum: This note seems to repeat what the start of the previous paragraph also states.

mypy/semanal.py
```diff
@@ -4156,3 +4189,8 @@ def visit_any(self, t: AnyType) -> Type:
         if t.type_of_any == TypeOfAny.explicit:
             return t.copy_modified(TypeOfAny.special_form)
         return t
+
+
+def get_prefix(fullname: str) -> str:
```
gvanrossum: This should really be moved to util so you won't need three copies of the same function.

mypy/server/astdiff.py
```diff
@@ -209,3 +135,210 @@ def visit_type_type(self, left: TypeType) -> bool:
         if isinstance(self.right, TypeType):
             return is_identical_type(left.item, self.right.item)
         return False
+
+
+SnapshotItem = Tuple[object, ...]
```
gvanrossum: It would be nice if you gave at least some examples of what various snapshots look like. Without knowing the internal structure of snapshots the code below is hard to follow. The representation may be easy to manipulate, but it seems to hide logical structure.

Longer term this should probably become a class rather than just a tuple.

mypy/server/astdiff.py
```diff
@@ -209,3 +135,210 @@ def visit_type_type(self, left: TypeType) -> bool:
...
+        node = symbol.node
+        # TODO: cross_ref, tvar_def, type_override?
+        fullname = node.fullname() if node else None
+        common = (fullname, symbol.kind, symbol.module_public)
+        if symbol.kind == MODULE_REF:
+            # This is a cross-reference to another module.
+            assert isinstance(node, MypyFile)
+            result[name] = ('Moduleref', common)
+        elif symbol.kind == TVAR:
+            assert False
```
gvanrossum: Is this really a TODO?

PR #4238, opened 2017-11-13
Add checks for overlapping TupleTypes

### LINKED ISSUE #4237: "Overloaded ... incompatible return types" error when one overloaded argument is a Tuple
Running mypy (0.550) on the following code:

```python
from typing import overload, Tuple, Any

class A: ...
class B: ...
class C: ...
class D: ...

@overload
def f(x: A) -> B: ...

@overload
def f(x: Tuple[C]) -> D: ...

def f(x: Any) -> Any:...
```

reports:

`error: Overloaded function signatures 1 and 2 overlap with incompatible return types`

I've tried a couple different types for the second overload and so far the error only seems to occur with `Tuple[...]` arguments.  For instance, I don't see the issue if I substitute `def f(x:List[C]) -> D: ...` as the 2nd overload.

elazarg:
Fix #4237.

This is not complete or bulletproof, but it's an improvement.

I don't see where should there be a special case for namedtuple; maybe there shouldn't.


mypy/meet.py
```diff
@@ -120,6 +120,17 @@ class C(A, B): ...
     if isinstance(s, UnionType):
         return any(is_overlapping_types(t, item)
                    for item in s.relevant_items())
+    if isinstance(t, TupleType) and isinstance(s, TupleType):
+        if t.length() == s.length():
+            if all(is_overlapping_types(ti, si, use_promotions)
+                   for ti, si in zip(t.items, s.items)):
+                return True
+        # TODO: Tuple[A, ...]
```
JukkaL: I think that this may work already -- `Tuple[A, ...]` is represented as an `Instance`, and the new check below should handle this (`TupleType` vs. `Instance`). Can you add a test case and remove this comment if I was correct?

elazarg: The case was not handled well, so I added support for it

elazarg:
Added support for `Tuple[A, ...]`.
I also fixed the use of fallback - I believe that comparing fallbacks is incorrect. 

@JukkaL the logic is quite different now, so it needs to be re-reviewed... Sorry about that :\

There are _many_ tests that can be added - involving combinations of tuples, namedtuple, varargs, instances and unions; but I prefer to do it in at a different time.

PR #4243, opened 2017-11-14
Check right operand of is expressions

### LINKED ISSUE #4055: Right 'is' operand not type checked
Mypy produces no error for this program, even though it clearly should:
```py
1 is 1()
```

ethanhs:
The right operand of `is` expressions were not being checked.
Fixes #4055


gvanrossum:
Thanks! Of course this found some issues in our internal codebase... But we'll survive.

ethanhs:
You're welcome. And it'd be a pretty poor linter to not catch bugs, so it's good bugs were caught :^)

PR #4248, opened 2017-11-15
Add LICENSE files to sdists

### LINKED ISSUE #4247: LICENSE file missing in pypi tarball
The source tarball https://pypi.io/packages/source/m/mypy/mypy-0.550.tar.gz doesn't have the LICENSE file. Please add it to the manifest.

ethanhs:
Fixes #4247

### DIFF
--- a/extensions/LICENSE
+++ b/extensions/LICENSE
@@ -4,7 +4,7 @@ Mypy extensions are licensed under the terms of the MIT license, reproduced belo
 
 The MIT License
 
-Copyright (c) 2015-2016 Jukka Lehtosalo and contributors
+Copyright (c) 2016-2017 Jukka Lehtosalo and contributors
 
 Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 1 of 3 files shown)


extensions/LICENSE
```diff
@@ -0,0 +1,27 @@
+Mypy extensions are licensed under the terms of the MIT license, reproduced below.
+
+= = = = =
+
+The MIT License
+
+Copyright (c) 2015-2016 Jukka Lehtosalo and contributors
```
ilevkivskyi: Shouldn't this be updated to year 2017?

ethanhs: AIUI, from what I have read updating it is not needed, but I will go ahead and make that change if you want.

ilevkivskyi: I am not an expert myself, but this is what I have seen typically.

ethanhs: Updated.

PR #4257, opened 2017-11-16
Do semantic analysis on lambda's bodies

### LINKED ISSUE #4098: Semantic analysis error not reported within lambda
This doesn't generate an error, though it's clearly invalid:

```py
lambda: xyz
```

This generates an error:

```py
lambda: 1 + ''
```

It looks like type checking errors are reported within a lambda but semantic analysis errors are lost. 

This is not a regression -- this bug has been around for a long time.

ethanhs:
Lambdas were seen as dynamic functions when going to report errors. This
changes them to not be dynamic functions and report semantic analysis
errors in their bodies.
Fixes #4098

The relevant check for `is_dynamic()` is here:
https://github.com/python/mypy/blob/b240125f96949b35dc2ad54c3707fd3dca5d170b/mypy/semanal.py#L3790

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -1480,6 +1480,9 @@ class LambdaExpr(FuncItem, Expression):
     def accept(self, visitor: ExpressionVisitor[T]) -> T:
         return visitor.visit_lambda_expr(self)
 
+    def is_dynamic(self) -> bool:
+        return False
+
 
 class ListExpr(Expression):
     """List literal expression [...]."""
--- a/test-data/unit/check-functions.test
+++ b/test-data/unit/check-functions.test
@@ -2212,3 +2212,19 @@ def i() -> List[Union[str, int]]:
     return x
 
 [builtins fixtures/dict.pyi]
+
+[case testLambdaSemanal]
+f = lambda: xyz
+[out]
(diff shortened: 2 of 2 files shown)


ilevkivskyi:
I think it is worth adding few tests for this.

ethanhs:
D'oh it appears I forgot to check in the tests I wrote. Just pushed them.

gvanrossum:
Nice, but it's introducing a new form of error. Here's an example:
```py
from typing import Callable
def f() -> Callable[[], int]:
    a = lambda: b  # error: Name 'b' is not defined
    b = 42
    return a
assert f()() == 42
```
I wouldn't mind so much except I'm seeing this example in a few places in our existing codebase where fixing it would be somewhat awkward (the lambda is passed as a callback to something that must exist before some local variable can be set -- but the callback won't be called until after that's taken care of).

ilevkivskyi:
I think with PEP 526 this can be safely fixed by ``b: int`` somewhere at the right scope. Otherwise ``# type: ignore`` here I think is still better than uncaught errors somewhere else. (There seem to be no middle ground here, either lambdas are considered dynamic and not typechecked, or they are always checked.)

gvanrossum:
Fair enough, I'll try to work around these.

Issue #4233, opened 2017-11-10
Split up JUnit XML output per file


kennydo:
Hello! I've recently started running mypy with `--junit-xml` since our CI system supports parsing JUnit output (and it's been a really useful feature!).

Currently, mypy's JUnit output lumps together all of the error output into one "testcase", like so:
```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuite errors="0" failures="1" name="mypy" skips="0" tests="1" time="0.826">
  <testcase classname="mypy" file="mypy" line="1" name="mypy" time="0.826">
    <failure message="mypy produced messages">testmodule/bar.py:2: error: The return type of "__init__" must be None
testmodule/bar.py:2: error: Missing return statement
testmodule/foo.py:4: error: Self argument missing for a non-static method (or an invalid type for self)
testmodule/foo.py:5: error: Incompatible return value type (got "str", expected "int")</failure>
  </testcase>
</testsuite>
```

Would anyone have any objection to breaking this output up into 1 testcase per file? I imagine it would look something like this:
```xml
<?xml version="1.0" encoding="utf-8"?>
<testsuite errors="0" failures="1" name="mypy" skips="0" tests="3" time="0.826">
  <testcase classname="testmodule.bar" file="testmodule/bar.py" line="0" name="testmodule.bar">
    <failure message="mypy produced messages">testmodule/bar.py:2: error: The return type of "__init__" must be None
testmodule/bar.py:2: error: Missing return statement</failure>
  </testcase>
  <testcase classname="testmodule.foo" file="testmodule/foo.py" line="0" name="testmodule.foo">
    <failure message="mypy produced messages">testmodule/foo.py:4: error: Self argument missing for a non-static method (or an invalid type for self)
testmodule/foo.py:5: error: Incompatible return value type (got "str", expected "int")</failure>
  </testcase>
  <testcase classname="testmodule.file_with_no_issues" file="testmodule/file_with_no_issues.py" line="0" name="testmodule.file_with_no_issues"></testcase>
</testsuite>
```

There are a few benefits to splitting it up by file:
- CI systems that support the JUnit format can nicely format each error message per file, making it easier for developers to see which files need fixing
- CI systems that collect and aggregate metrics about which test cases fail most often / how many tests are run can have more detailed information to play with
- We can account for files that don't have typing issues (like `testmodule/file_with_no_issues.py` in the example above).

[...]

PR #4263, opened 2017-11-18
Split up messages in JUnit XML by checked file

kennydo:
This implements the functionality mentioned in https://github.com/python/mypy/issues/4233.

If there were no serious errors, then mypy messages are grouped by which file they pertain to. (Because junit-xml does not expose a direct way to set just the elapsed time for an entire test suite, I faked it by dividing the total time by the number of files checked.)
If there was a serious error, then we don't have per-file output, so I just consider it one erroring test case.

I mimicked the cmdline test for my test cases.


kennydo:
Marking this as WIP. Didn't realize there was CI for Windows, so I'll update this when I've got that handled!

kennydo:
I finally set up a Windows VM and got this to work on windows too! This PR is ready for review.

gvanrossum:
Two high-level comments:

- It's a lot of code for this functionality (and adds a new dependency to boot). I'm still not sure if that's worth it to me. (Even though you've already written the code, somebody still has to review it. :-)

- I noticed that the reported time for each testcase is the same. That seems unavoidable, but can't you leave it out completely? Or do something more creative like use the file size or line count as weight? (Also the time for the testsuite[s] elements should be limited to microseconds.)

kennydo:
Regarding the functionality, I think the benefits I mentioned in [the issue](https://github.com/python/mypy/issues/4233) are worth it for developers typechecking in their CI systems. Having more detailed machine-readable mypy output opens a lot of doors for other tooling that might be interested in mypy's output. I might not have written the code in the most concise way, but I'm open to re-working it.

If we're concerned about the additional requirement, how does putting it under `extras_require` sound? People that don't want or need JUnit XML output wouldn't need to install anything extra.

After looking at a spec for JUnit XML (http://llg.cubic.org/docs/junit/), it looks like the `time` property on each test case is technically optional, but the JUnit XML library I used uses these values to compute the overall `time` on the `testsuite` (https://github.com/kyrus/python-junit-xml/blob/master/junit_xml/__init__.py#L126). I do like the idea of weighting it by line count though -- I'll try it out!

Could you clarify what you meant by "Also the time for the testsuite[s] elements should be limited to microseconds"? Do you mean rounding the values to the nearest 0.000001 ?

gvanrossum:
I'll ponder it more...

We actually use mypy in our own CI and this would perhaps make things mildly better but I don't think it would make a huge difference. (What would really help us is links that can be clicked through to the source code *as it is in the diff*, but this is not helping.)

> Could you clarify what you meant by "Also the time for the testsuite[s] elements should be limited to microseconds"? Do you mean rounding the values to the nearest 0.000001 ?

Indeed; the time there showed ridiculous "precision", something like 16 digits. Surely that's a floating point rounding error (especially now you've explained it is computed automatically from adding up the per-file times, which all use microseconds precision).

kennydo:
I tried to be more clever about distributing the times between the files, but because the set of "build source" files passed around isn't necessarily the same as the set of files actually showing up in the output (ex: when `pkg/foo.py` imports stuff from `pkg/bar.py`, but we only run the mypy command on `pkg/foo.py`, `pkg/bar.py` errors still show up, but it doesn't show up as a build source.), it ended up far too hacky and brittle.

`junit_xml` doesn't provide a way to round the summed time values. Even if I round the per-file values, I still end up getting that floating point issue in the summed value.

I could remove the `time` property entirely, but I think there is more value in knowing how long it took to typecheck everything than in trying to guess how long each file contributed.

gvanrossum:
Is it even necessary to use a 3rd party package (junit-xml) for this? The amount of code you've written to interface with it could just as easily be used to just construct the necessary XML by brute force, and then you have control over things like the time attribute.

kennydo:
Ah, good point. I've update this change to remove the junit-xml dependency and construct the XML just from stdlib.

gvanrossum:
Now looking (though note that this isn't going to make the upcoming 0.610 release).

mypy/junit.py
```diff
@@ -92,19 +210,52 @@ def create_serious_error_document(
...
+            (0x5FFFE, 0x5FFFF), (0x6FFFE, 0x6FFFF),
+            (0x7FFFE, 0x7FFFF), (0x8FFFE, 0x8FFFF),
+            (0x9FFFE, 0x9FFFF), (0xAFFFE, 0xAFFFF),
+            (0xBFFFE, 0xBFFFF), (0xCFFFE, 0xCFFFF),
+            (0xDFFFE, 0xDFFFF), (0xEFFFE, 0xEFFFF),
+            (0xFFFFE, 0xFFFFF), (0x10FFFE, 0x10FFFF),
+        ])
+
+    illegal_ranges = [
+        "%s-%s" % (chr(low), chr(high))
```
gvanrossum: This would seem to include a null byte in the regexp (since the very first range starts with 0x00). That seems asking for trouble, even if it works. I recommend putting \x or \u escapes in the regexp instead.

kennydo:
I've updated the branch per feedback. This should be ready for another look!

gvanrossum:
Thanks, I'll look into it later this week.

dragongame:
I'm truly disappointed by this decision, especially because you don't give any reason. Treating static code analysis as a single large unit test seams very counter intuitive. 
> UNIT TESTING is a level of software testing where individual units/ components of a software are tested.
[http://softwaretestingfundamentals.com/unit-testing/](http://softwaretestingfundamentals.com/unit-testing/)

> JUnit is a unit testing framework for the Java programming language.
[https://en.wikipedia.org/wiki/JUnit](https://en.wikipedia.org/wiki/JUnit)

The current solution is rather **ugly** and **complicated**. One can either run static code analysis for every single file and than merge the results, which is what we are using now, but this is **very time consuming** . Or one could split the xml afterwards and thereby also redoing all the inclusion detection, time consuming and error prone. Also ugly because it requires another tool to be developed.

One could also argue, that your case is the special case and that you could also easily make a script to collapse them into a single unit test.
I'm still in favor of supporting this with a simple switch, allowing both use cases.

However, my hope are quite low.

JukkaL:
> One can either run static code analysis for every single file and than merge the results

I think there may be some misunderstanding of mypy is supposed to be used. All the results from mypy are a global property of the whole program being analyzed, not specific to individual files. Fundamentally mypy doesn't attribute errors to causes but to symptoms. For example, a bad annotation in module `a` may generate no errors in `a` but several errors in `b` and `c`. Just looking at results for a single file can be confusing and doesn't necessarily tell much about the quality of that particular file.

The fact that we use an output format designed for unit tests doesn't change the fact that mypy is not a unit testing tool but basically a static analysis tool. Junit.xml files are only supported for convenience, since many tools happen to support them. The junit.xml integration necessarily has some level of impedance mismatch. 

Conceptually a mypy run is a single test, since there is generally no way of splitting up a run neatly to individual components/files (due to things like import cycles). Splitting the output per file may be a reasonable UI decision, but that's out of scope for mypy.

Finally, I want to avoid adding new command line options at the moment unless there is a very compelling reason, since the proliferation of options already makes it hard to find the right ones, and it's becoming a burden for us maintainers to keep all the options working correctly. This option doesn't quite make the cut, in particular since it's easy enough to get the desired results through a post-processing script.

gvanrossum:
Maybe we should remove the junit generation from mypy altogether. We have an internal use for this but we can easily replace it with a separate script -- such a script already exists at Dropbox.

JukkaL:
We may have many users relying on junit.xml generation, so I'd rather not drop it. I think the current level of functionality is fine, and makes it easy to get started with mypy in CI systems that support junit.xml. We might want to update the docs to say that only very basic level of functionality is provided, and in more complex cases the user should write a script.

wouterdb:
1. could you please be so kind as to provide the script to split the xml here?
2. a multi-line mypy report can be used by my IDE to click on the location and jump to it, making it much easier to process the hundreds of type errors I have.

gvanrossum:
A decent IDE should be able to parse the default output of mypy (not the junit.xml). It's just <FILE>:<LINE>: <ERROR>

PR #4268, opened 2017-11-20
extensions: Specify dependency on typing for py<35

### LINKED ISSUE #4264: mypy_extensions: missing dependency? typing (py27)
```
(mypytest) [root@80ca205d2ef4 recipe]# pip install mypy_extensions
Collecting mypy_extensions
  Downloading mypy_extensions-0.3.0.tar.gz
Building wheels for collected packages: mypy-extensions
  Running setup.py bdist_wheel for mypy-extensions ... done
  Stored in directory: /root/.cache/pip/wheels/58/47/b8/ffdcbc1959691d5ff42fa78937b36638ccbed59f2af8d35b18
Successfully built mypy-extensions
Installing collected packages: mypy-extensions
Successfully installed mypy-extensions-0.3.0
(mypytest) [root@80ca205d2ef4 recipe]# python -c "import mypy_extensions"
  File "<string>", line 1, in <module>
  File "/opt/conda/envs/mypytest/lib/python2.7/site-packages/mypy_extensions.py", line 8, in <module>
    from typing import Any
ImportError: No module named typing
```

nehaljwani:
Fixes #4264

### DIFF
--- a/extensions/setup.py
+++ b/extensions/setup.py
@@ -40,4 +40,7 @@ setup(
     platforms=['POSIX'],
     py_modules=['mypy_extensions'],
     classifiers=classifiers,
+    install_requires=[
+        'typing >= 3.5.3; python_version < "3.5"',
+    ],
 )


nehaljwani:
I set the minimum dependency on typing same as mypy did.

gvanrossum:
Seems to work, but I still wonder if we shouldn't put this in `extras_require`, like the main mypy setup.py. See https://hynek.me/articles/conditional-python-dependencies/ -- what do you think?

nehaljwani:
I think that method is deprecated now?
https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies
https://stackoverflow.com/a/32643122/1005215

gvanrossum:
But the blog I mentioned says "Sadly this – as of setuptools 21 – does
*silently* nothing." That blog is from May last year, so not that old, and
mypy doesn't actually specify which setuptools version it needs (I guess
that would be pointless since setuptools is already running by the time it
processes the requirements).

nehaljwani:
I have tested it for three other projects and it works. If will not work in the scenario when a dependency also specifies dependency on typing, but does not restrict it.

One of the comments at https://stackoverflow.com/a/32643122/1005215 also mentions that this worked for him and can't think of a reason as to why it didn't work for the author of that blog post.

gvanrossum:
Cool. Thanks!

hynek:
FTR, I ran some experiments the other day and it seems that the PEP 508 behaviour has been implemented as of setuptools **36.2** (released Jul 14, 2017) .  So you need setuptools 36.2 on the *build* machine to correctly encode the condition in your wheels.

For installation pip>1.4 should be fine so that should be a non-issue nowadays.  I’ve amended the blog post but I’m still waiting for confirmation of my findings.

gvanrossum:
Thanks! It would be nice if there was a wheel doctor app that showed the
details of what's in the wheel spec...

PR #4275, opened 2017-11-21
Thread line/col for lambda parameter to error about covariant typevar

### LINKED ISSUE #4270: Covariance error inferred for lambda has no associated line number
```
from typing import TypeVar, Generic, Callable

_Tco = TypeVar('_Tco', covariant=True)

class Thing(Generic[_Tco]):

    def chain(self, func):
        # type: (Callable[[_Tco], None]) -> None
        pass

    def end(self):
        # type: () -> None

        return self.chain(lambda _: None)
```

`example.py: error: Cannot use a covariant type variable as a parameter`

This error has no associated line number. (If the lambda is expanded into an explicit inner function typed as `(_Tco) -> None`, the line number is present.)

gvanrossum:
Fixes #4270

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -725,7 +725,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                             arg_type.variance == COVARIANT and
                             defn.name() not in ('__init__', '__new__')
                         ):
-                            ctx = arg_type
+                            ctx = arg_type  # type: Context
                             if ctx.line < 0:
                                 ctx = typ
                             self.fail(messages.FUNCTION_PARAMETER_CANNOT_BE_COVARIANT, ctx)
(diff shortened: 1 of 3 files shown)


gvanrossum:
@ilevkivskyi Do you think this is the right solution? I spiked on just this error, not sure how many others are similar.

mypy/checker.py
```diff
@@ -725,7 +725,10 @@ def check_func_def(self, defn: FuncItem, typ: CallableType, name: Optional[str])
                             arg_type.variance == COVARIANT and
                             defn.name() not in ('__init__', '__new__')
                         ):
-                            self.fail(messages.FUNCTION_PARAMETER_CANNOT_BE_COVARIANT, arg_type)
+                            ctx = arg_type  # type: Context
```
ilevkivskyi: I would add a test where problematic argument appears on a separate line.

Issue #2773, opened 2017-01-30
Warn about missing __init__.py


JukkaL:
If a package doesn't have an `__init__.py` file, the error message from mypy isn't very helpful (e.g., `Can't find package 'abc'`), and sometimes users can't figure this out by themselves. If mypy can't find a package, it could see if the directory exists somewhere in the module search path without an `__init__.py`, and give a message that says this explicitly. 

Examples of potential better messages:

```
$ mkdir foo
$ mypy -p foo
Can't find valid package 'foo' (foo/ has no __init__.py)
$ mypy -c "import foo"
<string>:1: error: Cannot find module named 'foo' (foo/ has no __init__.py)
```

gvanrossum:
Not also that in Python 3, a directory without `__init__.py` can be a package (there's a PEP for that). But it's intended only for a certain type of "namespace" packages. There once was an issue requesting support for this, but I rejected it because it would break the "crawl up until root of package" algorithm that mypy uses when passed a file inside a package. Maybe though there should be a flag to change that behavior? (My current workaround is "add an `__init__.pyi`.)

davidthornton:
I'm not seeing this message come through on release 0.521... I got stung by this, and wanted to  prevent anyone else from having to go through this by reminding y'all 😀

davidthornton:
Omg I just replied to Guido Van Rossum 😱

carljm:
At Instagram this behavior has been a common cause of new packages silently not being typechecked, and nobody realizing it until much later. Somebody (probably not very familiar with Python) will add a new package without an `__init__.py`, and it will work fine (because we are on Python 3, where `__init__.py` is not actually required), and they won't realize that their new code is not typechecked because mypy can't find it. We run with `ignore-missing-imports`, for other reasons we can't change at the moment, so it's totally silent.

It would be great to have an option to explicitly specify the package root and not require `__init__.py` files.

gvanrossum:
I agree it would be nice to have a way to override this. The "crawl-up" behavior is only needed for files passed on the command line. @carljm -- do you pass files inside `__init__`-less packages on the command line? If not, maybe in PY3 mode we could just change the behavior when you set `$MYPYPATH` and implement a module search (when following imports) that's more similar to what Python 3 itself does.

BTW what exactly do you mean by "package root"? For a module `m.py` inside a package `p`, would that be the directory `p` or its parent?

carljm:
[...]

PR #4277, opened 2017-11-22
[WIP] Namespace implementation

m-novikov:
This is currently very hacky namespace support.
So I would like to improve it or at least facilitate some discussion that will help to implement it.
As far I understand the correct way to code it would be to create an empty MypyFile tree for namespace modules. 
I don't understand the best way to workaround [process_stale_scc ](https://github.com/m-novikov/mypy/blob/7ae0f9cdf6dd3ac705139ea217ae14c76f78c562/mypy/build.py#L2426-L2442)
Which forces me to use ```if.namespace: return```

My use case pretty much described here https://github.com/python/mypy/issues/1645#issuecomment-318449340
What behavior should be if we run ```mypy -p namespace_pkg```
Ref: #2773 #1645


mypy/build.py
```diff
@@ -859,8 +859,12 @@ def find_modules_recursive(module: str, lib_path: List[str]) -> List[BuildSource
 
 def verify_module(id: str, path: str) -> bool:
     """Check that all packages containing id have a __init__ file."""
+    if path.endswith(tuple(PYTHON_EXTENSIONS)):
```
ethanhs: I think you should probably rewrite the rest of this function, and update the docstring, as this change makes the docstring incorrect, and much of the rest of the function duplicative (eg `path.endswith(('__init__.py', '__init__.pyi'))` will never be reached...

mypy/build.py
```diff
@@ -1759,9 +1774,13 @@ def parse_file(self) -> None:
                 except (UnicodeDecodeError, DecodeError) as decodeerr:
                     raise CompileError([
                         "mypy: can't decode file '{}': {}".format(self.path, str(decodeerr))])
-            assert source is not None
-            self.tree = manager.parse_file(self.id, self.xpath, source,
-                                           self.ignore_all or self.options.ignore_errors)
+
+            if not self.namespace:
```
ethanhs: A small nit:
```
if <condition>:
   action
else:
   other action
```
rather than
```
if not <condition>:
    other action
else:
    action
```

mypy/build.py
```diff
@@ -539,6 +544,7 @@ def __init__(self, data_dir: str,
         self.version_id = version_id
         self.modules = {}  # type: Dict[str, MypyFile]
         self.missing_modules = set()  # type: Set[str]
+        self.module_discovery = module_discovery
         self.plugin = plugin
         self.semantic_analyzer = SemanticAnalyzerPass2(self.modules, self.missing_modules,
                                                   lib_path, self.errors, self.plugin)
```
carljm: I'm not sure why BuildManager should get `lib_path` as a separate argument at all anymore; that should be encapsulated in `ModuleDiscovery`. Passing in both is just an opportunity for them to be out of sync, or for some code to use `lib_path` improperly, bypassing the `ModuleDiscovery` logic.

The only thing `BuildManager` seems to use `lib_path` for now is to pass it along to `SemanaticAnalyzerPass2`, but the latter doesn't seem to do anything with it either that I can find; it just stashes it on `self` and never looks at it again. So unless I miss something, I'd eliminate it from both.

mypy/build.py
```diff
@@ -778,59 +787,73 @@ def is_file(path: str) -> bool:
...
-def find_module(id: str, lib_path_arg: Iterable[str]) -> Optional[str]:
-    """Return the path of the module source file, or None if not found."""
-    lib_path = tuple(lib_path_arg)
+class ModuleDiscovery:
+    def __init__(self, lib_path: Iterable[str], namespaces_allowed: bool):
+        self.lib_path = tuple(lib_path)
+        self.namespaces_allowed = namespaces_allowed
 
-    def find() -> Optional[str]:
+    def find(self, id: str) -> Optional[str]:
```
carljm: Given that this class now handles both regular modules and namespaces, but this method handles only module-finding internals, I think it'd be clearer to name it `_find_module` instead of `find`; `find` as a public method (and first in source order) strongly implies it's the primary public API, which it's not.

mypy/build.py
```diff
@@ -1823,13 +1879,20 @@ def parse_file(self) -> None:
         self.check_blockers()
 
     def semantic_analysis(self) -> None:
+        if self.namespace:
```
carljm: Scattering these all through here seems ugly. I'm not very familiar with this logic, but why isn't it possible to just treat a namespace as an empty module and not need to skip all these methods?

test-data/unit/check-namespaces.test
```diff
@@ -0,0 +1,77 @@
...
+# flags: --config-file tmp/mypy.ini
+from mod_or_ns import a
+from mod_or_ns import b # E: Module 'mod_or_ns' has no attribute 'b'
+[file mypy.ini]
+[[mypy]
+namespace_packages = True
+mypy_path = ./tmp/dir:./tmp/other_dir
+[file dir/mod_or_ns.py]
+a = None
+[file other_dir/m/b.py]
```
carljm: Shouldn't `m` here be `mod_or_ns`  instead, for this test to be useful?

test-data/unit/check-namespaces.test
```diff
@@ -0,0 +1,77 @@
+-- Type checker test cases dealing with namespaces imports
```
carljm: Might be good to also have a test for a "namespace" inside a package with `__init__.py` (e.g. `a/b/c.py` where `a/__init__.py` exists but `a/b/__init__.py` does not.) This isn't a _useful_ deployment of a namespace package, but it works fine in Py3 and thus is likely to occur in the wild by accident of omission.

mypy/build.py
m-novikov: Yes, thanks, I will do that.

m-novikov:
So it seems there a lot of dimensions to this problem.
Right now mypy will discover the same module located in separate directories.
Given following directory structure:
dir1/mod1/__init__.py
dir2/mod1/a.py
dir2/mod1/__init__.py
dir2/mod1/b.py
Script:
```python
from m import a, b
dir1 = a  # type: str
dir2 = b  # type: str
```
Produces following errors:
```
test.py:3: error: Incompatible types in assignment (expression has type Module, variable has type "str")
test.py:4: error: Incompatible types in assignment (expression has type Module, variable has type "str")
```
But it should, in my opinion at least fail to import module a or module b depending of MYPYPATH directory order.

m-novikov:
Requirements for module discovery is following:
* merge stubs with existing modules.
* merge namespaces from different MYPYPATHS
* not merge modules located in different directories.

m-novikov:
@ethanhs could you help me with this error
```python
import typing as t

b = ['a', 'b', 'c']  # type: t.List[str]
a = tuple(p for p in b)  # type: t.Tuple[str]
```
fails with: 
```
error: Incompatible types in assignment (expression has type "Tuple[str, ...]", variable has type "Tuple[str]")
```
I don't understand

ilevkivskyi:
@m-novikov ``Tuple[str]`` is a tuple with exactly one element of type ``str``, ``Tuple[str, ...]`` is a tuple where all elements have type ``str`` (without length specified).

ethanhs:
@m-novikov  also, you may be interested in reading the docs here: https://docs.python.org/3/library/typing.html#typing.Tuple for more. The best way to fix that would be to change the type comment to be variable length as well (`Tuple[str, ...]`). Depending on how it is used later you may want to cast it.

ilevkivskyi:
@m-novikov Also why do you need ``tuple(p for p in b)`` and not just ``tuple(b)``?

m-novikov:
> Also why do you need tuple(p for p in b) and not just tuple(b)?
@ilevkivskyi it was the minimal sample that I came up with. I should have read more docs

m-novikov:
Should I remove find_module_clear_caches function?

m-novikov:
@ethanhs  could you review?
It was a bit more changes that I have anticipated.
But the previous version of ```find_module``` was behaving differently from python interpreter.
I am still not sure what needs to be done with find_module_*_cache. Probably all of this should be pulled to ModuleDiscovery and function find_module_clear_caches removed. As it used only in tests and ModuleDiscovery already has an empty cache on creation.

m-novikov:
@carljm could you try running it on your projects using namespaces to see if any problems arise?
it could be run with command line flag:
```
mypy --namespace-packages
```
Thanks.

ethanhs:
I'm afraid I won't have time to review this now. I will dismiss my review instead.

m-novikov:
@ilevkivskyi can you review this pull request?

carljm:
@m-novikov Yes, I will give it a try tomorrow and report back.

ilevkivskyi:
@m-novikov I can take a quick look later. I will wait for @carljm review.

carljm:
First issue I found testing this on our codebases: it doesn't prioritize normal packages or stubs over potential namespace packages (as Python does).

One way to reproduce this: create a file with `import locale; locale.setlocale(locale.LC_ALL, '')` and create an empty `locale/` directory next to it. With `--namespace-packages` option this will generate a pair of errors, because mypy is using the empty `locale/` directory as an empty package, instead of the typeshed stub for the `locale` module.

Another repro: create `path1/foo/` as an empty directory and `path2/foo/__init__.py` with `BAR = 1`, and `test.py` with `import foo; foo.BAR`. Run `MYPYPATH=path1:path2 mypy --namespace-packages test.py` and you'll get `Module has no attribute BAR`. This is wrong; mypy should prefer `path2/foo/__init__.py` (even though it comes later on the path) and ignore `path1/foo/`.

(Haven't finished analyzing other new issues for possible problems; will post separately if I find others, and make a note of when I'm done.)

carljm:
The new commit fixes the priority order I previously identified. I've gone through all the new errors generated by this branch (with the new `--namespace-packages` option) on our code and confirmed that the rest are all due to code in namespace packages that was previously not found and is now found.

ilevkivskyi:
@carljm I will take a look at this when you will an approving review.

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+            for ext in PYTHON_EXTENSIONS:
+                path = os.path.join(path_item, last_comp, sepinit + ext)
+                ctx.maybe_add_path(path, ModuleType.package)
+
+            for ext in PYTHON_EXTENSIONS:
+                path = os.path.join(path_item, last_comp + ext)
+                ctx.maybe_add_path(path, ModuleType.module)
+
+        if self.namespaces_allowed and not ctx.paths:
+            for path_item in paths:
```
carljm: rather than repeating the loop and the `is_pkg_path` check, I'd recommend building up the list of "possible namespace package paths" in its own list variable on your first iteration through `paths`, and then just using that already-collected list here.

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+                continue
+
+            if item.endswith(tuple(PYTHON_EXTENSIONS)):
+                item = item.split('.')[0]
+
+            yield module + '.' + item
+
+    def _collect_paths(self, paths: List[str], last_comp: str) -> List[str]:
+        """
+        Collect all available module paths
```
carljm: "Return the path (or namespace paths) for given (sub)module name, relative to given parent paths."

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+        for item in list_dir(path) or []:
+            if item == '__init__.py' or item == '__init__.pyi':
+                continue
+
+            if item.endswith(tuple(PYTHON_EXTENSIONS)):
+                item = item.split('.')[0]
+
+            yield module + '.' + item
+
+    def _collect_paths(self, paths: List[str], last_comp: str) -> List[str]:
```
carljm: I'd have understood this method sooner if `paths` were named `parent_paths` and `last_comp` were named `leaf_module_name` (paint the bikeshed as you will, but something clearer than `last_comp`).

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+        """
+        Collect all available module paths
+        """
+        sepinit = '__init__'
+        ctx = ImportContext()
+
+        # Detect modules in following order: package, module, namespace.
```
carljm: maybe also mention "stubs before .py" in this comment

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+        # First hit determines module type, consistency of paths to given type
```
carljm: this is an overly complex description based on the wrong idea that we may collect multiple stub files for a module. Maybe can just be eliminated

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+    def _verify_module(self, path: str) -> bool:
+        # At this point we already know that that it's valid python path
+        # We only need to check file existence
+        if not is_file(path):
             return False
-    return True
+
+        return True
+
+    def _verify_namespace(self, path: str) -> bool:
```
carljm: We don't really need this verification here either, because `_collect_paths` will only try to create a namespace package if it's already exhausted the possibility of a normal package.

In general I think most of the difficulty in understanding this code comes from splitting (or duplicating) this kind of logic between `ImportContext` and `ModuleDiscovery._collect_paths`; I recommend consolidating the logic in just one place.

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+        if not ok:
+            return None
+
+        if py_path:
+            self.has_py = True
+
+        self.type = type
+        self.paths.append(path)
+
+    def _verify_module(self, path: str) -> bool:
```
carljm: We don't need this verification at all, `_collect_paths` won't ever try to add a path that it hasn't already verified exists.

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+                hits.add(src.module)
+                result.append(src)
+        return result
+
+    def _find_modules_recursive(self, module: str) -> List[BuildSource]:
+        module_paths = self._find_module(module)
+
+        srcs = []  # type: List[BuildSource]
+        for path in module_paths:
+            if is_module_path(path) or is_pkg_path(path):
```
carljm: again here (and below) there's a lot of re-checking of the path type that is redundant; `_find_module` already had this information and threw it away. Can't we have `_find_module` throw away less info, return an object instead of just paths, and avoid unnecessary string path checking?

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+    package = 'package'
+    module = 'module'
+    namespace = 'namespace'
 
-    key = (id, lib_path)
-    if key not in find_module_cache:
-        find_module_cache[key] = find()
-    return find_module_cache[key]
 
+class ImportContext:
```
carljm: Why is this called `ImportContext` instead of just e.g. `Module`?

Also, I wonder if we really need both this and `BuildSource` as separate classes? There seems to be a lot of overlap, maybe just add a `type` (and the ability to hold multiple paths for namespace packages) to `BuildSource` and unify them?

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+
+        srcs = []  # type: List[BuildSource]
+        for path in module_paths:
+            if is_module_path(path) or is_pkg_path(path):
+                srcs.append(BuildSource(path, module, None))
+
+            if is_pkg_path(path):
+                path = dirname(path)
+                for submodule in self._find_submodules(module, path):
+                    srcs += self._find_modules_recursive(submodule)
```
carljm: by going all the way back to `_find_module` (which starts at the very beginning with all of `lib_path`) on every recursive call here, we are going to do a lot of repeated unnecessary work. It's worse with namespace packages than with the previous code. Imagine a namespace package that exists in three different paths, `dir1/foo`, `dir2/foo`, and `dir3/foo`. Consider that each of those has a different sub-package, `dir1/foo/a/`, `dir2/foo/b`, and `dir3/foo/c`. When we find `dir1/foo/a`, we throw away the context we already have (that we are within `dir1/foo/a`) and just call `_find_module('foo.a')`, which will end up unnecessarily looking at every single `lib_path` entry again to see if it has a `foo/a/` within it.

It may be that caching can reduce the impact of this multiplied repeated work, but not fully. And I think we can easily do better in the algorithm. `_collect_paths` already has the right logic to look for the final component of a submodule within a given limited set of parent paths. Can't we use `_collect_paths` here instead of recursive calls to be more efficient and only look within the relevant parent?

mypy/test/testmodulediscovery.py
```diff
@@ -0,0 +1,141 @@
+import os
+
+from unittest import mock, TestCase
+from typing import List, Set
+
+from mypy.build import ModuleDiscovery, find_module_clear_caches
+from mypy.myunit import Suite, assert_equal
```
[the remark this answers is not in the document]
carljm: @ilevkivskyi can you check me here?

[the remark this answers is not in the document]
ethanhs: Myunit usage is discouraged and we are trying to move to pytest to run our test suite.

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
+        self.type = None  # type: Optional[ModuleType]
+        # Paths can contain only one ".py" path, but multiple stubs
+        self.paths = []  # type: List[str]
+
+    def maybe_add_path(self, path: str, type: ModuleType) -> None:
+        """
+        Add path to import context.
+        Modifies self.paths in case if arguments satisfy import context state
+        """
+        assert path.endswith((os.path.sep,) + tuple(PYTHON_EXTENSIONS))
```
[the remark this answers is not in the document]
m-novikov: > Otherwise there is no circumstance where self.paths should ever have length > 1.

@carljm That's not true, as I understand.
Here are examples of packages split into multiple directories.
https://github.com/m-novikov/mypy-test/tree/master/split_package_stubs
Stubs and implementations located in separate directories.
You can even have split implementations:
https://github.com/m-novikov/mypy-test/tree/master/split_package_test
I am not sure if this is the desired behavior, but it is in master branch.
What if stubs are lagging in updates?

mypy/build.py
m-novikov: > We don't need this verification at all, _collect_paths won't ever try to add a path that it hasn't already verified exists.

That is incorrect we trying to add dynamically computed path and we don't know which of them exists at this moment yet. We are just trying all combinations in the specific order and hope that one of them fits.

mypy/build.py
```diff
@@ -1492,32 +1592,36 @@ def __init__(self,
...
-                        manager.log("Skipping %s (%s)" % (path, id))
-                        if follow_imports == 'error':
-                            if ancestor_for:
-                                self.skipping_ancestor(id, path, ancestor_for)
-                            else:
-                                self.skipping_module(id, path)
-                        path = None
-                        manager.missing_modules.add(id)
-                        raise ModuleNotFound
+                if not is_namespace_path(path):
```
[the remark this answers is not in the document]
m-novikov: Think so too, shied away from doing it initially to avoid bringing, even more, changes to find_module API.

mypy/build.py
m-novikov: Indeed it would greatly simplify the case if we didn't need to keep track of multiple stubs and implementations. As I wrote earlier, it would break existing behaviors.
Following directory structure would work on master:
```
dir1/pkg/__init__.py
dir1/pkg/a.py
dir2/pkg/__init__.py
dir2/pkg/b.py
dir3/pkg/__init__.pyi
dir3/pkg/c.pyi
```
And this code will not raise any errors:
```python
from pkg import a, b, c
```
Obviously, multiple implementations are incorrect. But in the case of stubs, I am not sure.

m-novikov: ImportContext is state machine in this case when it reaches final state it discards input.
And collect path provides it with the certain order of input.

mypy/build.py
```diff
@@ -778,95 +785,188 @@ def is_file(path: str) -> bool:
...
-        path = dirname(path)
-        if not any(os.path.isfile(os.path.join(path, '__init__{}'.format(extension)))
-                   for extension in PYTHON_EXTENSIONS):
+    Do we already discovered implementation?
+    What kind of module we discovered?
+    """
+    def __init__(self) -> None:
+        self.has_py = False  # type: bool
+        self.type = None  # type: Optional[ModuleType]
+        # Paths can contain only one ".py" path, but multiple stubs
```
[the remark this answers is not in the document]
m-novikov: https://github.com/python/mypy/pull/4277#discussion_r154532455

mypy/test/testmodulediscovery.py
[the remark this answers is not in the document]
ilevkivskyi: Yes, myunit is being gradually replaced by pytest.

mypy/build.py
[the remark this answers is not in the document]
carljm: I believe this is an accident of the current implementation. Python itself cannot find `pkg.b` in the `split_package_test` case (try `PYTHONPATH='dir1:dir2' python -c 'import pkg.b'`). Even for mypy itself, there is an inconsistency between `find_module` and `find_modules_recursive`; `find_module` can find either `pkg.a` or `pkg.b`, but `find_modules_recursive` on `pkg` will find only `pkg` and `pkg.a`, it will not find `pkg.b`. In this case `find_modules_recursive` is implementing the right Python (pre-namespace) import semantics (because of the "hits" tracking, it never accepts or uses more than one path for any module id), whereas `find_module` takes a performance shortcut (just append `pkg/` to all `lib_path` entries and see if the requested module exists there) instead of going module-by-module from the top level and verifying that we are using the correct path at each level.

In the no-stubs case, I think this can be considered purely a bug and should be fixed; mypy should not be able to find modules Python itself will not import, given the same lib path.

In the stubs case, I can imagine that some people may now be relying on this existing behavior in order to provide stubs for only some submodules in a package, while falling back to the implementation for submodules without stubs. I would leave it up to @ilevkivskyi / @gvanrossum / @JukkaL whether we need to preserve this behavior. If we do (and we do it consistently), it effectively means we need to treat packages with only `__init__.pyi` as a form of namespace package, stopping the search only when we find a package with `__init__.py`. (I don't think there's a use case for finding stub packages that occur _after_ the implementation on lib path).

Regardless, this exception should apply only for packages (ending with `__init__.pyi`); there's never any reason to track multiple paths for a regular module (we never make use of both `a.py` in one `lib_path` directory and `a.pyi` in another; one or the other always takes full priority).

mypy/build.py
carljm: Yes, I see the reasoning. It may have helped my understanding on first reading to have clarified in comment or docstring the critical dependency on paths being provided in the proper order of priority.

Regardless I think my initial comment here is still correct?

> We don't really need this verification here either, because _collect_paths will only try to create a namespace package if it's already exhausted the possibility of a normal package.

ethanhs:
@m-novikov I'm getting back to work on my PR #4278 on Wednesday. If you find time to respond to Carl's review and fix the merge conflict between now and then, that would be great. If you don't have time, that is entirely fine also, I will probably just work on my PR without namespace package support, and you may have to rebase on that. Thank you for your work on this so far!

m-novikov:
I plan to continue this Tuesday(12.12).

carljm:
Hi @m-novikov, any progress last week? I'm interested in seeing this land; if you won't have time for it soon I would be willing to pick it up and try to push it forward.

mypy/build.py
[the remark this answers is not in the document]
carljm: Any feedback from the core team here? Do we need to maintain the (I think accidentally implemented, but possibly useful) behavior that stub packages (with `__init__.pyi`) already in practice behave kind of like a namespace package? That is, if you have this structure:

```
.
├── src
│   └── pkg
│       ├── a.py
│       ├── b.py
│       └── __init__.py
└── stubs
    └── pkg
        ├── b.pyi
        └── __init__.pyi
```

with both `stubs` and `src` on `MYPYPATH` (in that order), mypy will still find and use `src/pkg/a.py`, even though that violates Python import's usual handling of non-namespace packages?

My feeling is that people may likely be relying on this (I think we currently are at Instagram, actually), but I don't want to put a lot of work into supporting it in this PR without confirmation that it should be formally supported, rather than fixed as a bug.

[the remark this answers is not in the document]
carljm: (The same is currently also true if `stubs` contains all `.py` files rather than `.pyi` -- this I am quite sure is a bug and should be fixed.)

gvanrossum:
This is sadly getting stale. @m-novikov do you want to continue to work on this? Then please resolve the conflicts. Alternatively, @carljm you can clone this PR and finish it if you want -- IIUC it's mostly ready to land? Maybe we can make the upcoming #4606 release (or else the one after that -- we plan to resume the ~3-week cadence we had towards the end of last year).

carljm:
@gvanrossum Sure, I'd be happy to pick this up and try to push it through to readiness, if @m-novikov doesn't plan to finish it.

There is one question about desired/intended behavior that came up that would be helpful to get feedback from you or other core devs. It's buried in a line-comment thread on an outdated revision, but should still be linkable: https://github.com/python/mypy/pull/4277#discussion_r157556653

(If you can't get to the question from that link, lmk and I'll copy it again here.)

carljm:
Hmm, that link doesn't seem to work if the thread isn't already expanded. Here's the issue:

Do we need to maintain the (I think accidentally implemented, but possibly useful) behavior that stub packages (with `__init__.pyi`) already in practice behave like a namespace package? That is, if you have this structure:

```
.
├── src
│   └── pkg
│       ├── a.py
│       ├── b.py
│       └── __init__.py
└── stubs
    └── pkg
        ├── b.pyi
        └── __init__.pyi
```

with both `stubs` and `src` on `MYPYPATH` (in that order), mypy will still find and use `src/pkg/a.py`, even though that violates Python import's usual handling of non-namespace packages?

My feeling is that people may likely be relying on this (I think we currently are at Instagram, actually), but I don't want to put a lot of work into supporting it in this PR without confirmation that it should be formally supported, rather than fixed as a bug.

gvanrossum:
IIRC I used to be convinced that mypy didn't have that feature until I
tried it. :-) Dropbox isn't depending on it and if Instagram can be made to
live without it I think it's better not to preserve that behavior. I also
think the mypy docs probably deny its existence, and PEP 484 may do the
same.

carljm:
Yes, we can live without it. I agree that it's better for mypy imports to consistently behave like Python imports.

carljm:
This has a fair amount of overlap/conflict with #4403 and #4623, so given that those are up to date and active, I'm holding off on bringing this up to date until they are merged.

carljm:
Now just waiting on #4693 to go in before I bring this up to date.

ethanhs:
@carljm now that PEP 561 is in, are you interested in picking this up?

carljm:
@ethanhs Yes, I'm interested and aware that the conflicting PRs have landed, just haven't had time yet. Not sure if I will before PyCon sprints.

ethanhs:
>ethanhs Yes, I'm interested and aware that the conflicting PRs have landed, just haven't had time yet. Not sure if I will before PyCon sprints.

Okay, no worries! I look forward to see you at the sprints.

gvanrossum:
So @carljm are you still interested in this? I could understand if not (maybe pyre does what you want). In that case I propose to close this -- it's probably too far out of date to be of much help to anyone else who wants to implement this feature.

carljm:
@gvanrossum For a while I was hoping to still follow through on this, just because I'd said I would, but I think I have to face reality that it's not going to make it to the top of my priority list now. Sorry for not clarifying that sooner.

I agree re closing this specific PR; I was planning to reuse the tests as a starting point, but expected to have to start over with the implementation.

gvanrossum:
I'm a little sad, but understand.

m-novikov:
Actually now I have time and willing to continue work on it, provided I have some consultation regarding edge cases.

ilevkivskyi:
@m-novikov I can't guarantee that I will be able to help you with all questions, but I think we have enough people following here who will be able to help. So please go ahead!

gvanrossum:
It's probably best to open a new PR then, once you're ready.

PR #4278, opened 2017-11-22
[WIP] Begin implementing PEP 561 checking

ethanhs:
This is the initial work to add checking using packages that opt into type checking via
PEP 561.

TODO:

~~- [ ] Namespace package support (awaiting #4277)~~ I'm going ahead without this.
- [ ] Tests (installing packages etc is going to be interesting...)
- [x] Check PEP 561 conformance to [resolution order](https://www.python.org/dev/peps/pep-0561/#type-checker-module-resolution-order)
- [x] ignore errors in these files.
- [ ] support running with alternate Python version.
- [ ] document how to make a package compatible.

As of right now, you can install regular packages with types and third party stub packages and they should be picked up. `$MYPYPATH` supersedes both types of packages, and third-party stubs supersede bundled type packages per PEP 561. Errors in installed packages should be ignored.

Questions:
~~- Should this be behind a flag?~~
~~- Should errors that are ignored in typeshed be ignored for these packages too? I'm not too keen on this, but open to recommendations.~~

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -792,6 +792,21 @@ def find_module(id: str, lib_path_arg: Iterable[str]) -> Optional[str]:
                 dir = os.path.normpath(os.path.join(pathitem, dir_chain))
                 if os.path.isdir(dir):
                     dirs.append(dir)
+            try:
+                user_dir = site.getusersitepackages()
+                package_dirs = site.getsitepackages() + [user_dir]
+            except AttributeError:
+                package_dirs = [get_python_lib()]
+
+            for pkg_dir in package_dirs:
+                # Third-party stub packages
+                stub_pkg = os.path.join(pkg_dir, components[0] + '_stubs')
+                if os.path.isfile(os.path.join(stub_pkg, 'py.typed')):
+                    components[0] = components[0] + '_stubs'
+                    dirs.append(os.path.join(pkg_dir, os.sep.join(components[:-1])))
+                elif os.path.isfile(os.path.join(pkg_dir, components[0], 'py.typed')):
+                    dirs.append(os.path.join(pkg_dir, dir_chain))
+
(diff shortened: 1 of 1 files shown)


mypy/build.py
```diff
@@ -792,6 +792,21 @@ def find() -> Optional[str]:
                 dir = os.path.normpath(os.path.join(pathitem, dir_chain))
                 if os.path.isdir(dir):
                     dirs.append(dir)
+            try:
+                user_dir = site.getusersitepackages()
```
JelleZijlstra: Doesn't this use the site packages of the mypy installation, not that of the code being typechecked?

JelleZijlstra:
I think we should suppress errors in these packages, although I haven't thought deeply about the exact semantics required.

The general reasoning is that third-party packages aren't code the user controls. If I use mypy with `--disallow-untyped-defs`, start using lxml, and install the lxml-stubs package, I don't want to get errors saying the lxml stubs are missing annotations, because I can't do anything about it.

JelleZijlstra:
Also, I don't think it needs to be behind a flag.

mypy/build.py
ethanhs: Yes I need to add that. Thanks for the reminder!

ethanhs:
Namespace package (PEP 420) support should probably wait until it is generally supported in mypy (see #4277), so I removed that from the TODO list for this PR.

It seems that `subprocess` is being rather uncooperative (not capturing stdout), thus I need to experiment a bit more about capturing the output of invoked Python.

Also, ignoring errors in installed packages should be an easy change, so I will implement that.

ethanhs:
I'm afraid I've run out of time, and will it will be a few weeks before I can finish this, which is probably better anyway so that #4277 can be done before I complete this.

ethanhs:
Closing in favor of #4403 for simplicity.

PR #4282, opened 2017-11-24
Modifying multiple files per update in fine-grained incremental mode

JukkaL:
The approach is to split multiple file updates into multiple single-file
updates, and only report errors for the final update. If a modified file
brings in additional files through an import, we'll process one of the
imported files instead, changing the target file as we go. The 
motivation is that this should result in less redundant work than 
postponing the imported file, as a missing imported file will result
in a lot of stale targets which would need to processed again on the
next update.

This still doesn't handle blockers well. I'll fix these issues separately.

This is inefficient as some files can be parsed multiple times, with 
O(n**2) worst case behavior (n is the number of changed files). It 
should be fine when a small number of files need to be processed.

### DIFF
--- a/mypy/server/update.py
+++ b/mypy/server/update.py
@@ -220,9 +220,8 @@ def update_single_isolated(module: str,
                                                              List[Tuple[str, str]]]:
     """Build a new version of one changed module only.
 
-    Don't propagate changes to elsewhere in the program.
-
-    Raise CompleError on encountering a blocking error.
+    Don't propagate changes to elsewhere in the program. Raise CompleError on
+    encountering a blocking error.
 
     Args:
         module: Changed module (modified, created or deleted)
@@ -241,9 +240,7 @@ def update_single_isolated(module: str,
           modules)
     """
     if module in manager.modules:
-        path1 = os.path.normpath(path)
-        path2 = os.path.normpath(manager.modules[module].path)
-        assert path1 == path2, '%s != %s' % (path1, path2)
+        assert_equivalent_paths(path, manager.modules[module].path)
 
     old_modules = dict(manager.modules)
     sources = get_sources(graph, [(module, path)])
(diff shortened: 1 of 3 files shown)


mypy/server/update.py
```diff
@@ -174,70 +211,85 @@ def get_all_dependencies(manager: BuildManager, graph: Dict[str, State],
...
-    for st in graph.values():
-        if st.id not in old_graph and st.id not in changed_set:
-            changed_set.add(st.id)
-            assert st.path
-            changed_modules.append((st.id, st.path))
-    # TODO: Handle multiple changed modules per step
-    assert len(changed_modules) == 1, changed_modules
+    changed_modules = get_all_changed_modules(module, path, old_graph, graph)
+    # If there are multiple modules to process, only process one of them and return the
+    # remaining ones to the caller.
```
gvanrossum: Maybe note that you process the last one in the list first? This is significant, because the first one in the list is guaranteed (by the implementation of `get_all_changed_modules()`) to be the original module.

mypy/server/update.py
```diff
@@ -174,70 +211,85 @@ def get_all_dependencies(manager: BuildManager, graph: Dict[str, State],
...
+    if len(changed_modules) > 1:
+        remaining_modules = changed_modules[:-1]
+        # The remaining modules haven't been processed yet so drop them.
+        for id, _ in remaining_modules:
+            del manager.modules[id]
+            del graph[id]
+        module, path = changed_modules[-1]
```
gvanrossum: Somehow this came as a small surprise to me -- I would chop off the designated item first. (Maybe using `pop()` so the slice on line 262 is unnecessary.)

mypy/server/update.py
```diff
@@ -174,70 +211,85 @@ def get_all_dependencies(manager: BuildManager, graph: Dict[str, State],
...
+    # Process the changed file.
     state.parse_file()
-
     # TODO: state.fix_suppressed_dependencies()?
-
-    # Run remaining passes of semantic analysis.
     try:
         state.semantic_analysis()
     except CompileError as err:
         # TODO: What if there are multiple changed modules?
```
gvanrossum: That TODO is currently irrelevant.

mypy/server/update.py
```diff
@@ -279,6 +336,20 @@ def get_sources(graph: Graph, changed_modules: List[Tuple[str, str]]) -> List[Bu
...
+def get_all_changed_modules(root_module: str,
+                            root_path: str,
+                            old_graph: Dict[str, State],
+                            new_graph: Dict[str, State]) -> List[Tuple[str, str]]:
+    changed_set = {root_module}
+    changed_modules = [(root_module, root_path)]
+    for st in new_graph.values():
+        if st.id not in old_graph and st.id not in changed_set:
+            changed_set.add(st.id)
+            assert st.path
```
gvanrossum: The assert looks odd in the middle of updating the two parallel data structures -- I'd put it at the top of the block.

JukkaL:
I'll add suggested improvements to my next PR if they are still relevant.

PR #4294, opened 2017-11-30
Two optimizations for load_graph()

gvanrossum:
Profiling load_graph() found that there were some unnecessary stat() calls going on, and, more importantly, that Options.clone_for_module() is very inefficient if you have many sections in your mypy.ini file. I solved the latter by introducing a cache. An alternative design would move the cache into BuildManager -- that would avoid a reference cycle, but it turns out that having the cache in the root Options object means that it survives between dmypy check runs, and that's a nice win there. (This explains at least partly why load_graph() was slow even with everything loaded in memory.)

@JukkaL

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -20,6 +20,7 @@ import json
 import os.path
 import re
 import site
+import stat
 import sys
 import time
 from os.path import dirname, basename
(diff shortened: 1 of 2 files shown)


mypy/options.py
```diff
@@ -177,17 +180,22 @@ def __repr__(self) -> str:
         return 'Options({})'.format(pprint.pformat(self.__dict__))
 
     def clone_for_module(self, module: str) -> 'Options':
+        res = self.clone_cache.get(module)
```
JukkaL: Add docstring that mentions that after this method has been called the options object should be treated as immutable, because otherwise the cache would become stale.

mypy/build.py
```diff
@@ -1056,11 +1058,10 @@ def validate_meta(meta: Optional[CacheMeta], id: str, path: Optional[str],
 
     # TODO: Share stat() outcome with find_module()
     path = os.path.abspath(path)
-    # TODO: Don't use isfile() but check st.st_mode
-    if not os.path.isfile(path):
+    st = manager.get_stat(path)  # TODO: Errors
```
JukkaL: What about adding a `is_file` method to `BuildManager` instead and using it here?

gvanrossum: That's a bigger effort, we'd have to give BuildManager a cache results and use that everywhere. There's also an inefficiency in find_module(), which seems to make way to many stat() and listdir() calls, which I haven't tracked down.

PR #4295, opened 2017-11-30
Do not warn on returning any from deferred node

### LINKED ISSUE #4296: Strange "Returning Any" warning
`mypy --warn-return-any` (ver. 0.550) gives a strange "Returning Any from function ..." warning for the following code:
```
def f(a: 'A') -> int:
    if a._x == 2:
        return 2
    n = 1
    return n

class A:
    def __init__(self) -> None:
        self._x = 2
```
```
foo.py:5: warning: Returning Any from function declared to return "int"
```
Combining `n = 1` `return n` into one statement `return 1` or putting the definition of `A` before `f` makes the warning disappear.

elazarg:
Fix #4296

The access to member defined in `__init__` causes the current node to be deferred, so mypy assumes expressions are `Any` but should not warn about it.

This is not the only conceivable fix - for example, I'm not sure why mypy should continue checking the function in the first place.

### DIFF
--- a/test-data/unit/check-warnings.test
+++ b/test-data/unit/check-warnings.test
@@ -209,5 +209,3 @@ def foo(a1: A) -> int:
 class A:
     def __init__(self, x: int) -> None:
         self._x = x
-
-
--- a/test-data/unit/check-warnings.test
+++ b/test-data/unit/check-warnings.test
@@ -197,3 +197,17 @@ class Test(object):
             return None
 [builtins fixtures/list.pyi]
 [out]
+
+[case testReturnAnyDeferred]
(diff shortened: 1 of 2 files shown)


test-data/unit/check-warnings.test
```diff
@@ -197,3 +197,17 @@ class Test(object):
...
+def foo(a1: A) -> int:
+    if a1._x:
+        return 1
+    n = 1
+    return n
+
+class A:
+    def __init__(self, x: int) -> None:
+        self._x = x
+
```
JukkaL: Extra empty lines at the end of the file.

PR #4313, opened 2017-12-03
Document empty tuple syntax

### LINKED ISSUE #4211: Is there a way to represent an empty tuple type?
`logging.handlers.SMTPHandler` arg `secure` accepts something like this:

    Union[EmptyTuple, Tuple[str], Tuple[str, str]]

Is there a way to define the `EmptyTuple` part?

ethanhs:
Fixes #4211.

### DIFF
--- a/docs/source/builtin_types.rst
+++ b/docs/source/builtin_types.rst
@@ -13,6 +13,7 @@ Type                Description
 ``bytes``           8-bit string
 ``object``          an arbitrary object (``object`` is the common base class)
 ``List[str]``       list of ``str`` objects
+``Tuple[int]``      a tuple of one int, ``Tuple[()]`` is the empty tuple
 ``Dict[str, int]``  dictionary from ``str`` keys to ``int`` values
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
I'm not sure this belongs in the cheatsheet, since the cheatsheet is meant to be a quick reference for common types, and the type of an empty tuple is definitely not a common need. Maybe there isn't a better place for it though.

ilevkivskyi:
> Maybe there isn't a better place for it though.

There is "Common issues" section. Maybe this discussion can be put there?

JelleZijlstra:
I doubt it's a common issue either. :)

Maybe it should be in https://mypy.readthedocs.io/en/latest/builtin_types.html, which is currently missing `Tuple`.

gvanrossum:
I think adding various forms of tuples to https://mypy.readthedocs.io/en/latest/builtin_types.html is a good idea, but I like adding this to the cheat sheets too (many people use *just* the cheat sheet for reference, and this is definitely an issue where everyone needs help).

PR #4316, opened 2017-12-04
Ignore specific module

### LINKED ISSUE #3476: Document how to selectively ignore specific imports

ethanhs:
Fixes #3476.

### DIFF
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -227,6 +227,17 @@ for all mypy runs in this tree, and also selectively turns on the
 package.  This issues an error for function definitions without
 type annotations in that subdirectory only.
 
+If you would like to ignore specific imports, instead of ignoring all missing
+imports with ``--ignore-missing-imports``, use a section of the configuration
+file per module such as the following to ignore missing imports from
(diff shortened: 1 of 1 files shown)


gvanrossum:
Is there a way to silence just a particular non-existing module without type-ignore on each import? That would be a good addition.

Issue #4311, opened 2017-12-03
Enum-like user classes


orenbenkiki:
Feature request: Allow annotating a class as enum-like for `mypy` type checking purposes.

I have classes dealing with a value-like (immutable) type.
One of the classes in that family is an enum-like class, whose values are of the new immutable type.
I tried to implement this by deriving from the standard library `Enum`, but it has pretty strong opinions about the implementation which do not match the way my value type works (it is too different from a plain integer).

No worries, I just implemented my own `MyEnumBase` base class (and `MyEnumMeta` for it), loosely inspired by `Enum`.
It works fine.

The problem is convincing `mypy` to type check it.
Specifically, if I write:

```python
class EnumExample(MyEnumBase):
    FOO = MyAuto()
```

Then `mypy` will not recognize that `EnumExample.FOO` has the type `EnumExample`.
I can work around this by the following horrible hack:

```python
class IsEnumBase:
    """Check whether a given class is an enumerated base class."""
    def __eq__(self, class_name: Any) -> bool:
        return class_name == 'enum.Enum' or class_name == 'my.module.MyEnumBase'

mypy.nodes.ENUM_BASECLASS = IsEnumBase()
```

Which mostly works, but, *shudder* (is there a better way to achieve the same effect?)

It would be nice if I could annotate `MyEnumBase` somehow to let `mypy` it is an enum base class.
For example, if the single `mypy.nodes.ENUM_BASECLASS` was replaced by a set of class names, so that it were possible to insert additional class names into it (and provide a command line flag to `mypy` for doing that).

ethanhs:
Have you tried inheriting from `EnumMeta`? I believe that was designed to be more customizable.

ilevkivskyi:
Can this be achieved by descriptors? mypy already supports ``__get__`` and ``__set__``.

orenbenkiki:
@ethanhs: I did try deriving from `EnumMeta`, but that was using Python 3.6.0. Perhaps it was modified since but I'm still supporting 3.6.0. Even if I could use it, from looking at the code in `mypy.nodes`, it seems it explicitly checks whether the base class name is `enum.Enum` rather than checking whether the base class uses `EnumMeta` as a metaclass, so I doubt that would solve the type checking problem.

[...]

PR #4319, opened 2017-12-04
Use EnumMeta instead of Enum to mark enum classes

elazarg:
Implement #4311.

The tests will fail until python/typeshed#1770 is synched.


test-data/unit/check-enum.test
```diff
@@ -6,10 +6,19 @@ class Medal(Enum):
...
     silver = 2
     bronze = 3
+reveal_type(Medal.bronze)  # E: Revealed type is '__main__.Medal'
 m = Medal.gold
-m = 1
-[out]
-main:7: error: Incompatible types in assignment (expression has type "int", variable has type "Medal")
+m = 1  # E: Incompatible types in assignment (expression has type "int", variable has type "Medal")
+
+[case testEnumFromEnumMetaBasics]
```
ethanhs: I think it would be helpful to have a testcase for something that inherits from a class that has `EnumMeta` as a metaclass, to check those classes are still found to be `Enum`s for all intents and purposes.

Also perhaps a test of a `EnumMeta` deriving class that tries to use generics?

elazarg: >I think it would be helpful to have a testcase for something that inherits from a class that has EnumMeta as a metaclass, to check those classes are still found to be Enums for all intents and purposes.

Will do, but note that this is already handled by classes that inherit `Enum` itself.

I'm not sure I understand the second suggestion. Should mypy support something like that?

ethanhs: Good point on the first, but just to be safe. Also on the second, I mean to check that it fails correctly :)

ethanhs:
Hm, I think you need to `git submodule update` to pull the latest changes from typeshed.

elazarg:
`git submodule update` doesn't do the trick for some reason. I tried `git checkout upstream/master typeshed` which feels horribly wrong, but seems to work.

test-data/unit/check-enum.test
```diff
@@ -6,10 +6,37 @@ class Medal(Enum):
...
+[case testEnumFromEnumMetaSubclass]
+from enum import EnumMeta
+class Achievement(metaclass=EnumMeta): pass
+class Medal(Achievement):
+    gold = 1
+    silver = "hello"
+    bronze = None
+reveal_type(Medal.bronze)  # E: Revealed type is '__main__.Medal'
+m = Medal.gold
+m = 1  # E: Incompatible types inassignment (expression has type "int", variable has type "Medal")
```
ethanhs: Needs a space between "in" and "assignment"

elazarg: Yep. Sorry about that.

mypy/semanal.py
```diff
@@ -1275,6 +1273,11 @@ def analyze_metaclass(self, defn: ClassDef) -> None:
             # do not declare explicit metaclass, but it's harder to catch at this stage
             if defn.metaclass is not None:
                 self.fail("Inconsistent metaclass structure for '%s'" % defn.name, defn)
+        else:
+            if defn.info.metaclass_type.type.fullname() == 'enum.EnumMeta':
```
gvanrossum: So if the metaclass is a subclass of EnumMeta, it's not an enum? (I'm not sure I care, but I wonder if this is an intentional choice.)

test-data/unit/check-enum.test
```diff
@@ -6,10 +6,37 @@ class Medal(Enum):
...
+reveal_type(Medal.bronze)  # E: Revealed type is '__main__.Medal'
 m = Medal.gold
-m = 1
-[out]
-main:7: error: Incompatible types in assignment (expression has type "int", variable has type "Medal")
+m = 1  # E: Incompatible types in assignment (expression has type "int", variable has type "Medal")
+
+[case testEnumFromEnumMetaBasics]
+from enum import EnumMeta
+class Medal(metaclass=EnumMeta):
```
gvanrossum: This code doesn't actually work -- you need to also inherit from `enum.Enum`, else the class definition fails at runtime with
```
TypeError: Medal().__init__() takes no arguments
```
(Note: I'm not asking that mypy enforce this -- but I still think the test should work as an example.)

test-data/unit/check-enum.test
```diff
@@ -6,10 +6,37 @@ class Medal(Enum):
...
+    gold = 1
+    silver = "hello"
+    bronze = None
+reveal_type(Medal.bronze)  # E: Revealed type is '__main__.Medal'
+m = Medal.gold
+m = 1  # E: Incompatible types in assignment (expression has type "int", variable has type "Medal")
+
+[case testEnumFromEnumMetaSubclass]
+from enum import EnumMeta
+class Achievement(metaclass=EnumMeta): pass
```
gvanrossum: This must also inherit from `enum.Enum`.

test-data/unit/check-enum.test
elazarg: I'm not sure what to do about it. If Medal inherits from `Enum`, the test doesn't check for what I want it to check. We might as well drop it.

elazarg: I can add a `def __init__(self, *args): pass` constructor, if you deem it better. It works, though might be considered an implementation detail.

msullivan: I'd say add a comment explaining that it doesn't work at runtime and what you are testing for.

JelleZijlstra:
This seems to have broken typeshed CI. https://travis-ci.org/python/typeshed/jobs/364931878 I'll look into it now.

PR #4323, opened 2017-12-04
Produce better error on invalid types like List(int)

### LINKED ISSUE #4172: Better message for invalid types like List(int)
The error message "invalid type comment or annotation" isn't very helpful if a user writes `List(int)` instead of `List[int]` (or `Optional(int)`).  This seems to be a common error.

Example:

```py
from typing import List

def f():
    # type: () -> List(int)   # invalid type comment or annotation
    return [1]
```

A better message could be something like this:

```
program.py:3: error: Syntax error in type annotation
program.py:3:note: Suggestion: Use List[...] instead of List(...)
```

msullivan:
Fixes #4172

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -1015,6 +1015,10 @@ class TypeConverter(ast3.NodeTransformer):
         if self.errors:
             self.errors.report(line, column, msg)
 
+    def note(self, msg: str, line: int, column: int) -> None:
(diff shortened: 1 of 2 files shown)


msullivan:
Added some more tests, including a python3 one (the existing one was already python2).

I'm taking a look at the other issues mentioned in the bug and will put up another PR for them.

PR #4325, opened 2017-12-05
Clean up decorator implementation

JukkaL:
Store decorated function as Decorator in symbol table, not as Var.
Previously module-level function were stored as Var nodes in symbol
tables.  That was inconsistent, as decorated methods were stored as
Decorators in symbol tables.

As a side effect, this also fixes some cases of conditional method
definition that were incorrectly rejected previously.

This helps with fine-grained incremental checking, as it gives us
access to a decorated function through the module symbol table.

### DIFF
--- a/test-data/unit/check-functions.test
+++ b/test-data/unit/check-functions.test
@@ -1411,6 +1411,20 @@ f()
 f(1)
 [builtins fixtures/list.pyi]
 
+[case testDefineConditionallyAsImportedAndDecorated]
+from typing import Callable
+
+def dec(f: Callable[[], None]) -> Callable[[], None]: ...
+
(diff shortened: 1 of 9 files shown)


test-data/unit/check-functions.test
```diff
@@ -1398,6 +1398,37 @@ f()
 f(1)
 [builtins fixtures/list.pyi]
 
+[case testDefineConditionallyAsImportedAndDecorated]
```
gvanrossum: These two tests also seems to pass without this change. Is that expected?

JukkaL:
Yes, earlier versions of this PR had these failing and that's why I added
these.

Issue #2088, opened 2016-09-02
Typechecking attrs-generated classes


lvh:
[attrs](https://github.com/hynek/attrs) removes object boilerplate:

``` python
>>> import attr
>>> @attr.s
... class C(object):
...     x = attr.ib(default=42)
```

This has many advantages, like eq/hash behavior working, reprs working, easy to add runtime validators, &c. However, it's not obvious to me how I can typecheck creating instances of this, because there's no `__init__`. (I'm using Python 2.7, so comment syntax, if it matters :))

gvanrossum:
Hi Laurens, this looks like something that should be doable with the (as yet hypothetical) plugin architecture: https://github.com/python/mypy/issues/1240

wsanchez:
A bit more detail, as the problem here isn't just that you can't validate `attrs` class initializer usage using `mypy` (and, in fact, you can validate types at runtime using `attrs`), but also that `mypy` is emitting false positives, which makes `mypy` harder to use if you use `attrs`.

Here's a script `thing.py`:

```python
from pathlib import Path

from attr import attrib, attrs
from attr.validators import instance_of


@attrs()
class ThingWithPath(object):
    _path = attrib(validator=instance_of(Path))

    def exists(self):
        return self._path.exists()


thing1 = ThingWithPath(Path("."))
print("thing1 exists:", thing1.exists())

thing2 = ThingWithPath(path=Path("."))
print("thing2 exists:", thing2.exists())
```

Which runs thusly (note the runtime type validation, if that's a thing you like):

```console
wsanchez$ python thing.py
thing1 exists: True
thing2 exists: True
Traceback (most recent call last):
  File "thing.py", line 21, in <module>
    thing3 = ThingWithPath(".")
  File "<attrs generated init 72ca13b0f10a00b12fb26accfd84d196fc2af0c6>", line 5, in __init__
  File "/Users/wsanchez/Dropbox/Developer/PIE/wsanchez/ConfigService/.tox/py35-coverage/lib/python3.5/site-packages/attr/validators.py", line 24, in __call__
    attr, self.type, value,
TypeError: ("'_path' must be <class 'pathlib.Path'> (got '.' that is a <class 'str'>).", Attribute(name='_path', default=NOTHING, validator=<instance_of validator for type <class 'pathlib.Path'>>, repr=True, cmp=True, hash=True, init=True, convert=None, metadata=mappingproxy({})), <class 'pathlib.Path'>, '.')
```

[...]

PR #4328, opened 2017-12-06
Add Class Decorator/Metaclass/Base Class plugin

euresti:
Work towards #2088

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -1,14 +1,12 @@
 """Plugin system for extending mypy."""
 
+from collections import OrderedDict
 from abc import abstractmethod
-from typing import Callable, List, Tuple, Optional, NamedTuple, TypeVar, Dict
+from typing import Callable, List, Tuple, Optional, NamedTuple, TypeVar
 
-from mypy.nodes import Expression, StrExpr, IntExpr, UnaryExpr, Context, \
-    DictExpr, TypeInfo, ClassDef, ARG_POS, ARG_OPT, Var, Argument, FuncDef, \
-    Block, SymbolTableNode, MDEF, CallExpr, RefExpr, MemberExpr, NameExpr, \
-    AssignmentStmt, TempNode
+from mypy.nodes import Expression, StrExpr, IntExpr, UnaryExpr, Context, DictExpr, ClassDef
 from mypy.types import (
-    Type, Instance, CallableType, TypedDictType, UnionType, NoneTyp, TypeVarType,
+    Type, Instance, CallableType, TypedDictType, UnionType, NoneTyp, FunctionLike, TypeVarType,
     AnyType, TypeList, UnboundType, TypeOfAny
 )
 from mypy.messages import MessageBuilder
@@ -62,6 +60,15 @@ class SemanticAnalyzerPluginInterface:
     def named_type(self, qualified_name: str, args: Optional[List[Type]] = None) -> Instance:
         raise NotImplementedError
 
+    @abstractmethod
+    def parse_bool(self, expr: Expression) -> Optional[bool]:
+        raise NotImplementedError
+
+    @abstractmethod
+    def fail(self, msg: str, ctx: Context, serious: bool = False, *,
+             blocker: bool = False) -> None:
+        raise NotImplementedError
+
 
 # A context for a function hook that infers the return type of a function with
 # a special signature.
@@ -108,12 +115,14 @@ AttributeContext = NamedTuple(
         ('context', Context),
         ('api', CheckerPluginInterface)])
 
+# A context for a class hook that modifies the class definition.
 ClassDefContext = NamedTuple(
     'ClassDecoratorContext', [
-        ('cls', ClassDef),
-        ('context', Optional[Expression]),
+        ('cls', ClassDef),       # The class definition
+        ('reason', Expression),  # The expression being applied (decorator, metaclass, base class)
         ('api', SemanticAnalyzerPluginInterface)
     ])
+
 
 class Plugin:
     """Base class of all type checker plugins.
@@ -217,7 +226,7 @@ class ChainedPlugin(Plugin):
         return self._find_hook(lambda plugin: plugin.get_class_metaclass_hook(fullname))
 
     def get_class_base_hook(self, fullname: str
-                                 ) -> Optional[Callable[[ClassDefContext], None]]:
+                            ) -> Optional[Callable[[ClassDefContext], None]]:
         return self._find_hook(lambda plugin: plugin.get_class_base_hook(fullname))
 
     def _find_hook(self, lookup: Callable[[Plugin], T]) -> Optional[T]:
@@ -251,12 +260,6 @@ class DefaultPlugin(Plugin):
             return typed_dict_get_callback
         elif fullname == 'builtins.int.__pow__':
             return int_pow_callback
-        return None
-
-    def get_class_decorator_hook(self, fullname: str
-                                 ) -> Optional[Callable[[ClassDefContext], None]]:
-        if fullname == 'attr.s':
-            return attr_s_callback
         return None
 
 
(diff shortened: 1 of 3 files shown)


mypy/plugin.py
```diff
@@ -332,3 +374,53 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+
+
+def attr_s_callback(ctx: ClassDefContext) -> None:
+    """Add an __init__ method to classes decorated with attr.s."""
+    info = ctx.cls.info
+    # TODO: Handle default arguments
+    has_default = {}  # type: Dict[str, Expression]
+    args = []
+
+    for name, table in info.names.items():
```
euresti: Just discovered this is overly broad, but you get the idea.

mypy/plugin.py
```diff
@@ -53,6 +54,14 @@ def named_generic_type(self, name: str, args: List[Type]) -> Instance:
         raise NotImplementedError
 
 
+class SemanticAnalyzerPluginInterface:
+    """Interface for accessing semantic analyzer functionality in plugins."""
+
+    @abstractmethod
+    def named_type(self, qualified_name: str, args: Optional[List[Type]] = None) -> Instance:
```
ilevkivskyi: Is this the only method that would be potentially used by plugins? Do we need to add some more?

mypy/plugin.py
```diff
@@ -332,3 +374,53 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+def attr_s_callback(ctx: ClassDefContext) -> None:
+    """Add an __init__ method to classes decorated with attr.s."""
+    info = ctx.cls.info
+    # TODO: Handle default arguments
+    has_default = {}  # type: Dict[str, Expression]
+    args = []
+
+    for name, table in info.names.items():
+        if table.type:
+            var = Var(name.lstrip("_"), table.type)
```
ilevkivskyi: Variable name ``table`` is a bit misleading, I would call this ``symbol`` or similar.

mypy/plugin.py
```diff
@@ -332,3 +374,53 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+
+def attr_s_callback(ctx: ClassDefContext) -> None:
+    """Add an __init__ method to classes decorated with attr.s."""
+    info = ctx.cls.info
+    # TODO: Handle default arguments
+    has_default = {}  # type: Dict[str, Expression]
+    args = []
+
+    for name, table in info.names.items():
+        if table.type:
```
ilevkivskyi: I think this should be treated differently. You should look at the right hand sides for assignments in the class and take those that have ``attr.ib`` (note however, that I am not an expert in ``attrs`` so I would like to see many tests for this reviewed by an ``attrs`` maintainer).

mypy/semanal.py
```diff
@@ -720,6 +720,35 @@ def analyze_class_body(self, defn: ClassDef) -> Iterator[bool]:
...
+                            defn.metaclass)
+                    hook = self.plugin.get_class_metaclass_hook(metaclass_name)
+                    if hook:
+                        hook(ClassDefContext(defn, self))
+
+                for type_info in defn.info.bases:
+                    hook = self.plugin.get_class_base_hook(type_info.type.fullname())
+                    if hook:
+                        hook(ClassDefContext(defn, self))
+
```
ilevkivskyi: I would factor out this whole block into a separate method, and call it here like ``self.apply_plugin_hooks(defn)`` or similar.

euresti: Sounds good. Will do.

mypy/plugin.py
euresti: If possible I'd like to remove all this code from the PR.  The actual plugin will require possibly more thought.  Is it weird to just have a commit that creates a plugin but doesn't actually use it?

Also it turns out I can't actually use a plugin unless attr.pyi exists.  So I'd have to figure out how to bring in https://github.com/python-attrs/attrs/pull/238

mypy/plugin.py
euresti: My current explorations also need `parse_bool` and `anal_type`

euresti: `fail` could be useful too. Should I add them as needed or add them now?

euresti:
All right. I cleaned up the code and everything passes.  I removed the actual plugin to implement in a different PR.

mypy/plugin.py
ilevkivskyi: > Should I add them as needed or add them now?

I am not sure why we actually need this class. Can't we just pass everything? Or am I missing something?

mypy/semanal.py
```diff
@@ -719,9 +719,44 @@ def analyze_class_body(self, defn: ClassDef) -> Iterator[bool]:
...
                 self.setup_type_promotion(defn)
-
+                self.apply_class_plugin_hooks(defn)
                 self.leave_class()
 
+    def apply_class_plugin_hooks(self, defn: ClassDef) -> None:
+        """Apply a plugin hook that may infer a more precise definition for a class."""
+        def get_fullname(expr: Expression) -> Optional[str]:
+            # We support @foo.bar(...) @foo.bar and @bar
+            # TODO: Support IndexExpressions?
```
ilevkivskyi: ``IndexExpression`` is prohibited in decorators by the Python parser.

euresti: But not base classes which also uses this.

mypy/plugin.py
euresti: I'm following the pattern of the other plugins.  Technically the object that's passed in is the full object. I believe this base class is here to limit what methods the plugins should be calling. i.e. if the method isn't listed here then the mypy checker will complain about it.

mypy/semanal.py
ilevkivskyi: For base classes yes, this is important, since they can be generic.

mypy/plugin.py
ilevkivskyi: Then we probably need to think what to add here. My expectation is that many methods of ``SemanticAnalyzer`` may be useful. You can play more with your implementation for ``attrs`` and add here whatever you ever used.

mypy/semanal.py
```diff
@@ -719,9 +719,54 @@ def analyze_class_body(self, defn: ClassDef) -> Iterator[bool]:
...
+            try:
+                base = expr_to_unanalyzed_type(base_expr)
+            except TypeTranslationError:
+                continue  # This will be reported later
+            if not isinstance(base, UnboundType):
+                continue
+            sym = self.lookup_qualified(base.name, base)
+            if sym is None or sym.node is None:
+                continue
+            base_name = sym.node.fullname()
```
ilevkivskyi: Why this is more complex than in other cases? I would expect you can just extract the name using ``get_full_name`` (if you would add one more case for ``IndexExpr``).

mypy/semanal.py
```diff
@@ -719,9 +719,54 @@ def analyze_class_body(self, defn: ClassDef) -> Iterator[bool]:
...
                 self.leave_class()
 
+    def apply_class_plugin_hooks(self, defn: ClassDef) -> None:
+        """Apply a plugin hook that may infer a more precise definition for a class."""
+
+        def get_fullname(expr: Expression) -> Optional[str]:
+            # We support foo.bar(...), foo.bar, and bar.
+            if isinstance(expr, CallExpr):
+                if isinstance(expr.callee, RefExpr):
+                    return expr.callee.fullname
```
ilevkivskyi: ``RefExpr`` can be both ``NameExpr`` and ``MemberExpr``. Why do you have the same code for both here, while below you need special casing?

mypy/semanal.py
euresti: Because of this:
```
    # Base class expressions (not semantically analyzed -- can be arbitrary expressions)
    base_type_exprs = None  # type: List[Expression]
```

If I ask for the `expr.base.fullname` of `class Foo(Mapping[str, str])` I get `Mapping` instead of `typing.Mapping`

mypy/semanal.py
euresti: I should probably just recurse.

euresti:
Ok. Tests pass and IndexExpr is parsed. Let me know if this needs anything else.

mypy/semanal.py
```diff
@@ -719,9 +719,48 @@ def analyze_class_body(self, defn: ClassDef) -> Iterator[bool]:
...
+
+        def get_fullname(expr: Expression) -> Optional[str]:
+            if isinstance(expr, CallExpr):
+                return get_fullname(expr.callee)
+            elif isinstance(expr, (MemberExpr, NameExpr)):
+                if expr.fullname:
+                    return expr.fullname
+
+                # If we don't have a fullname look it up.
+                sym = self.lookup_type_node(expr)
```
ilevkivskyi: Add a longer comment here explaining that currently base classes are analysed in different manner than other expressions (analysed by ``exprtotype.py`` etc.) and therefore some AST nodes doesn't have names set. Maybe also add a ``TODO`` item to unify this (this is however low priority since this will require major rewriting with minor benefits).

mypy/semanal.py
```diff
@@ -719,9 +719,48 @@ def analyze_class_body(self, defn: ClassDef) -> Iterator[bool]:
...
+                if sym:
+                    return sym.fullname
+            elif isinstance(expr, IndexExpr):
```
ilevkivskyi: Move this branch after ``CallExpr`` so the order will look more logical.

mypy/plugin.py
ilevkivskyi: I think we can keep only these methods, and then add more in your PR with the actual plugin (if necessary).

mypy/semanal.py
```diff
@@ -719,9 +719,48 @@ def analyze_class_body(self, defn: ClassDef) -> Iterator[bool]:
...
+    def apply_class_plugin_hooks(self, defn: ClassDef) -> None:
+        """Apply a plugin hook that may infer a more precise definition for a class."""
+
+        def get_fullname(expr: Expression) -> Optional[str]:
+            if isinstance(expr, CallExpr):
+                return get_fullname(expr.callee)
+            elif isinstance(expr, (MemberExpr, NameExpr)):
+                if expr.fullname:
+                    return expr.fullname
+
```
ilevkivskyi: I think this empty line is not needed here.

mypy/plugin.py
```diff
@@ -136,7 +161,17 @@ def get_attribute_hook(self, fullname: str
                            ) -> Optional[Callable[[AttributeContext], Type]]:
         return None
 
-    # TODO: metaclass / class decorator hook
+    def get_class_decorator_hook(self, fullname: str
+                                 ) -> Optional[Callable[[ClassDefContext], None]]:
+        return None
+
+    def get_class_metaclass_hook(self, fullname: str
```
ilevkivskyi: I think you can call this just ``get_metaclass_hook``.

mypy/plugin.py
```diff
@@ -136,7 +161,17 @@ def get_attribute_hook(self, fullname: str
...
-    # TODO: metaclass / class decorator hook
+    def get_class_decorator_hook(self, fullname: str
+                                 ) -> Optional[Callable[[ClassDefContext], None]]:
+        return None
+
+    def get_metaclass_hook(self, fullname: str
+                           ) -> Optional[Callable[[ClassDefContext], None]]:
+        return None
+
+    def get_class_base_hook(self, fullname: str
```
euresti: get_base_class_hook maybe?

ilevkivskyi: > get_base_class_hook maybe?

As you prefer, both are fine with me.

gvanrossum:
None from me, Jukka will let you know when he's up. :-)

mypy/plugin.py
JukkaL: Here is why we are using an ABC instead exposing the entire semantic analyzer interface:

* By only giving access to a subset of an interface, we declare the public part of the semantic analyzer interface. This has a few benefits related to modularity and loose coupling:
  * Everything else in the semantic analyzer can be more freely modified without breaking plugins. If we change the public interface, we may break 3rd party plugins. (Currently the plugin interface is still not officially supported so we can change things without worries, but I'd like to make the interface more stable in the future.)
  * It's less likely that plugins will rely on internal implementation details that are likely to change, or on poorly thought-out internal APIs that are hard to use correctly.
  * It's easier to reason about plugins in general, as the plugins are expected to stick to the exposed public interface instead of freely accessing mypy internals.
* This can avoid a cyclic import dependency (maybe not right now, but it will help untangle cyclic dependencies). They slow down incremental checking.

ilevkivskyi:
@JukkaL 
> And sorry for the late minute review!

No problem, thanks! I have opened #4363 to track the tests.

> I wonder how they will interact with things that happen later during semantic analysis? This might affect forward references at least.

Yes, it makes to have tests specifically for forward references. I created a separate issue for this #4364

@euresti 
Could you please make this renaming?
> Maybe rename ``AnalyzerPluginInterface`` to ``TypeAnalyzerPluginInterface`` to make it clear how it's different from the new ABC?

euresti:
Renamed!

PR #4329, opened 2017-12-06
Give better error message for ommited parens in function type

msullivan:
Partially fixes #4323 

The direct check for a paren feels janky, but I'm not sure if there is a nicer way.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -157,6 +157,9 @@ class ASTConverter(ast3.NodeTransformer):
         self.is_stub = is_stub
         self.errors = errors
 
+    def note(self, msg: str, line: int, column: int) -> None:
+        self.errors.report(line, column, msg, severity='note')
+
     def fail(self, msg: str, line: int, column: int) -> None:
         self.errors.report(line, column, msg)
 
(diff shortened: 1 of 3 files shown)


mypy/fastparse.py
```diff
@@ -356,6 +359,8 @@ def do_func_def(self, n: Union[ast3.FunctionDef, ast3.AsyncFunctionDef],
                     arg_types.insert(0, AnyType(TypeOfAny.special_form))
             except SyntaxError:
                 self.fail(TYPE_COMMENT_SYNTAX_ERROR, n.lineno, n.col_offset)
+                if n.type_comment and n.type_comment[0] != "(":
+                    self.note('Suggestion: wrap input types in parens', n.lineno, n.col_offset)
```
JukkaL: Have you tested what happens if there are extra spaces after `# type:`, such as `# type:<space><space>int -> int`?

Style nits: Use "argument types" instead of "input types". Spell parens as "parentheses".

msullivan: I tested it but probably should put a test for it into the test suite.

Roger.

PR #4331, opened 2017-12-06
Fix incremental speed regression

### LINKED ISSUE #4135: Slow incremental run when single module has errors
`mypy -i` is slower than expected in the following scenario:

* Check out fresh mypy repository
* Create tiny module `mypy.x` that generates an error: `echo '1 + ""' > mypy/x.py`
* Run `mypy -i mypy` 
  * This takes a while, generates an error for `mypy/x.py` as expected
* Run `mypy -i mypy`
  * This is as slow as the previous run, but caching should make this much faster

gvanrossum:
Fixes #4135.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1474,6 +1474,9 @@ class State:
     # Whether to ignore all errors
     ignore_all = False
 
+    # Whether the module has an error or any of its dependencies have one.
+    transitive_error = False
+
     # Type checker used for checking this file.  Use type_checker() for
     # access and to construct this on demand.
     _type_checker = None  # type: Optional[TypeChecker]
@@ -1944,7 +1947,7 @@ class State:
         if self.manager.options.quick_and_dirty:
             is_errors = self.manager.errors.is_errors_for_file(self.path)
         else:
-            is_errors = self.manager.errors.is_errors()
+            is_errors = self.transitive_error
         if is_errors:
             delete_cache(self.id, self.path, self.manager)
             self.meta = None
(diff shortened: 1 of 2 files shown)


gvanrossum:
Don't merge yet -- this has an O(N**2) component. We can do better by having an "errors" bit per file. (Though it's really per SCC that counts.)

gvanrossum:
What do you think of this version? It sets a flag `transitive_error` when there's an error in either the current module or any of its dependencies. (Except it does it it per SCC, so if any dependency of any module in an SCC has this flag set, all modules in the SCC get it set.)

There's one bit fof strangeness (see the last local commit) -- one test decided to have a different `[stale]` list.

(If you'd rather see the previous O(N**2) version, I can probably get it back using git reflog. :-)

JukkaL:
The approach looks good to me. I'm a little worried about the difference in the test case though -- might be worth figuring out why it has changed. Feel free to merge this once you are happy with this (and have fixed Travis).

gvanrossum:
Ah, I understand why that test (testIncrementalWithSilentImportsAndIgnore) had to be modified. There are no errors in `c/__init__.py`, and it does not depend on modules with errors. However it is processed after a module with errors (`c/submodule.py`), and with the old approach that meant no cache file would be written for `c/__init__.py` (in fact that was the bug we're fixing here). But with the new approach the file is not considered to have a (transitive) error, and the logic in write_cache() decides it can write the cache file. So it's a good thing.

PR #4336, opened 2017-12-07
Suggest spurious trailing commas as a cause for invalid tuple types

### LINKED ISSUE #4172: Better message for invalid types like List(int)

msullivan:
Fixes #4172 

This will generate the suggestion in some situations where it is probably not the right thing, unfortunately. (Like when the tuple type is wrapped in parens). That isn't *wrong*, since this is only a suggestion, but I don't have a good intuition for how likely those cases are.


gvanrossum:
Short and simple. I like it!

PR #4338, opened 2017-12-08
Various improvements to fine-grained incremental mode

JukkaL:
Summary of changes:

* Fixes to `NamedTuple`
* Fixes to overloads
* Fixes to generic classes
* Fixes to decorators
* Fixes to processing statements in class bodies and module top levels
* Detect endless fine-grained dependency loops
* Add support for testing which triggers get triggered

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -2333,11 +2333,11 @@ class SymbolTableNode:
         self.kind = kind
         self.node = node
         self.type_override = typ
-        self.module_hidden = module_hidden
         self.module_public = module_public
         self.normalized = normalized
         self.alias_tvars = alias_tvars
         self.implicit = implicit
+        self.module_hidden = module_hidden
 
     @property
     def fullname(self) -> Optional[str]:
@@ -2359,6 +2359,18 @@ class SymbolTableNode:
             return node.var.type
         else:
             return None
+
+    def copy(self) -> 'SymbolTableNode':
+        new = SymbolTableNode(self.kind,
+                              self.node,
+                              self.type_override,
+                              self.module_public,
+                              self.normalized,
+                              self.alias_tvars,
+                              self.implicit,
+                              self.module_hidden)
+        new.cross_ref = self.cross_ref
+        return new
 
     def __str__(self) -> str:
         s = '{}/{}'.format(node_kinds[self.kind], short_type(self.node))
@@ -2449,6 +2461,10 @@ class SymbolTable(Dict[str, SymbolTableNode]):
         a.insert(0, 'SymbolTable(')
         a[-1] += ')'
         return '\n'.join(a)
+
+    def copy(self) -> 'SymbolTable':
+        return SymbolTable((key, node.copy())
+                           for key, node in self.items())
 
     def serialize(self, fullname: str) -> JsonDict:
         data = {'.class': 'SymbolTable'}  # type: JsonDict
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -670,6 +670,8 @@ class ClassDef(Statement):
     type_vars = None  # type: List[mypy.types.TypeVarDef]
     # Base class expressions (not semantically analyzed -- can be arbitrary expressions)
     base_type_exprs = None  # type: List[Expression]
+    # Special base classes like Generic[...] get moved here during semantic analysis
+    removed_base_type_exprs = None  # type: List[Expression]
     info = None  # type: TypeInfo  # Related TypeInfo
     metaclass = None  # type: Optional[Expression]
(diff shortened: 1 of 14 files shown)


test-data/unit/deps-statements.test
```diff
@@ -549,3 +549,101 @@ def f() -> int:
...
+
+def f(x: A) -> A:
+    g()
+    return x
+
+def g() -> None: pass
+[builtins fixtures/isinstancelist.pyi]
+[out]
+<m.A.__init__> -> <m.B.__init__>
+<m.A> -> <m.f>, m, m.A, m.B, m.f
```
gvanrossum: Remind me, why does A trigger m (the whole module?) but not B?

test-data/unit/fine-grained.test
```diff
@@ -996,3 +996,232 @@ class B:
...
+[triggered]
+2: <b>
+3: <b>
+[out]
+main:2: error: Revealed type is 'contextlib.GeneratorContextManager[builtins.None]'
+==
+a.py:1: error: Cannot find module named 'b'
+a.py:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
+main:2: error: Revealed type is 'contextlib.GeneratorContextManager[builtins.None]'
+a.py:3: error: Cannot find module named 'b'
```
gvanrossum: There's only one "import b" in a.py (on line 3) -- why does that error get reported both for line 1 and for line 3?

test-data/unit/fine-grained.test
```diff
@@ -996,3 +996,232 @@ class B:
...
+
+@contextmanager
+def f(x: List[int]) -> Iterator[None]:
+    x.append(1)
+    yield
+[typing fixtures/typing-full.pyi]
+[builtins fixtures/list.pyi]
+[out]
+==
+a.py:1: error: Cannot find module named 'b'
```
gvanrossum: Ditto.

test-data/unit/fine-grained.test
```diff
@@ -996,3 +996,232 @@ class B:
...
+    from n import g
+[file m.py]
+class A: pass
+[file n.py]
+def g() -> None: pass
+[delete m.py.2]
+[delete n.py.2]
+[out]
+==
+main:1: error: Cannot find module named 'm'
```
gvanrossum: And ditto?

test-data/unit/merge.test
```diff
@@ -829,6 +829,110 @@ NameExpr:11: Overload(def (x: target.A<0>) -> target.A<0>, def (x: builtins.int<
...
+class B: pass
+
+@overload
+def f(x: A) -> B: pass
+@overload
+def f(x: str) -> str: pass
+
+def f(x): pass
+[out]
+MypyFile:1<0>(
```
gvanrossum: I hope you verified this file by hand -- I can't follow anything here in detail.

mypy/nodes.py
```diff
@@ -2360,6 +2363,18 @@ def type(self) -> 'Optional[mypy.types.Type]':
         else:
             return None
 
+    def copy(self) -> 'SymbolTableNode':
```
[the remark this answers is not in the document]
ethanhs: The dataclasses backport only works on 3.6 since dataclasses' class based declaration syntax requires PEP 526. It will be a couple years before we can use it (3.5 goes out of support in late 2020).

test-data/unit/fine-grained.test
JukkaL: That's a bug with the current implementation. There's a TODO comment elsewhere but I haven't repeated it everywhere where this happens.

test-data/unit/merge.test
JukkaL: Yeah I looked at it by hand. At some point I may update the output of these test cases, as currently it's super verbose.

mypy/semanal.py
```diff
@@ -2880,7 +2877,7 @@ def fail_enum_call_arg(self, message: str,
         self.fail(message, context)
         return [], [], False
 
-    def visit_decorator(self, dec: Decorator) -> None:
+    def visit_decorator(self, dec: Decorator, func_body: bool = True) -> None:
```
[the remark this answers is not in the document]
JukkaL: Looks like it's left over from earlier revisions. I'll remove it.

test-data/unit/deps-statements.test
JukkaL: It looks like it's because `A` is used as a base class in `B`, whereas `B` is not used as a base class. Base classes are considered to be in the top-level module scope. (Effectively both will trigger the module top level since `m.A` and `m.B` also include `m`.)

PR #4341, opened 2017-12-11
Some daemon code cleanup

### LINKED ISSUE #4194: Refactor some parts of dmypy.py
- Move class GcLogger to its own file -- it depends on nothing.
- Move the Server class to its own file, and the mypy imports with it -- could improve client startup time.

gvanrossum:
- Move do_restart() and wait_for_server() right after do_start()
- Introduce BadStatus exception, use instead of SystemExit
- Add a timeout option to request(), use for stop/status
- Include all socket ops in the try/except clause
- Rationalize and document exceptions raised by receive().
- Move GcLogger to its own file; fixes #4194
- Make status quiet by default; add -v to print full status
- Replace -q flags with -v flags
- If 'error' in response, sys.exit with that message
- Streamline the way command subparsers are updated
- Misc other cleanup

### DIFF
--- a/mypy/dmypy.py
+++ b/mypy/dmypy.py
@@ -26,42 +26,45 @@ parser = argparse.ArgumentParser(description="Client for mypy daemon mode",
 parser.set_defaults(action=None)
 subparsers = parser.add_subparsers()
 
-start_parser = subparsers.add_parser('start', help="Start daemon")
-start_parser.add_argument('--log-file', metavar='FILE', type=str,
+start_parser = p = subparsers.add_parser('start', help="Start daemon")
+p.add_argument('--log-file', metavar='FILE', type=str,
                           help="Direct daemon stdout/stderr to FILE")
-start_parser.add_argument('flags', metavar='FLAG', nargs='*', type=str,
+p.add_argument('flags', metavar='FLAG', nargs='*', type=str,
                           help="Regular mypy flags (precede with --)")
 
-restart_parser = subparsers.add_parser('restart',
+restart_parser = p = subparsers.add_parser('restart',
     help="Restart daemon (stop or kill followed by start)")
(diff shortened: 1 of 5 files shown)


mypy/dmypy.py
```diff
@@ -196,14 +162,67 @@ def wait_for_server(timeout: float = 5.0) -> None:
...
+
+@action(stop_parser)
+def do_stop(args: argparse.Namespace) -> None:
+    """Stop daemon via a 'stop' request."""
+    try:
+        response = request('stop', timeout=5)
+    except BadStatus:
+        raise
+    except Exception as err:
+        sys.exit("Daemon is stuck; consider %s kill" % sys.argv[0])
```
JukkaL: See my comment above about exceptions.

mypy/dmypy.py
```diff
@@ -196,14 +162,67 @@ def wait_for_server(timeout: float = 5.0) -> None:
...
+    status = read_status()
+    if args.verbose:
+        show_stats(status)
+    check_status(status)
+    try:
+        response = request('status', timeout=5)
+    except BadStatus:
+        raise
+    except Exception as err:
+        print("Daemon is stuck; consider %s kill" % sys.argv[0])
```
JukkaL: Should we display some information about the exception, since we are catching arbitrary exceptions? An exception could be caused by a bug in dmypy.

mypy/dmypy.py
```diff
@@ -99,79 +109,35 @@ def do_start(args: argparse.Namespace) -> None:
...
 
     We first try to stop it politely if it's running.  This also sets
-    mypy flags (and has the same issues as start).
+    mypy flags (like start).
     """
     try:
         response = request('stop')
-    except SystemExit:
+    except BadStatus:
         pass
```
JukkaL: Maybe add a comment about when we are likely to hit this (daemon not running?).

PR #4342, opened 2017-12-11
ListExpr is not an lvalue anymore

### LINKED ISSUE #4340: Mypy ignores type annotation in assignment to square bracket target list
```python
from typing import Tuple, List
(x1, y1) = (1, 2)  # type: Tuple[str, int]  # E: Incompatible types in assignment (expression has type "int", variable has type "str")
[x2, y2] = (1, 2)  # type: Tuple[str, int]  # no error - bad
reveal_type(x2)  # E: Revealed type is 'builtins.int' - bad
```
Since the notation `[x1, y1]` is just another syntax for target list, the above notation should work. But even if they shouldn't, mypy should at least warn about mismatched notation.

IMHO mypy should not differentiate at all between the cases after parsing - there should be only <s>TupleType</s>TupleExpr for lvalues.

elazarg:
Fix #4340 by considering TupleExpr as the only lvalue sequence.

(And maybe one day it will be LvalueSeq)

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -960,8 +960,11 @@ class ASTConverter(ast3.NodeTransformer):
 
     # List(expr* elts, expr_context ctx)
     @with_line
-    def visit_List(self, n: ast3.List) -> ListExpr:
-        return ListExpr([self.visit(e) for e in n.elts])
+    def visit_List(self, n: ast3.List) -> Union[ListExpr, TupleExpr]:
+        expr_list = [self.visit(e) for e in n.elts]  # type: List[Expression]
+        if isinstance(n.ctx, ast3.Store):
+            return TupleExpr(expr_list)
+        return ListExpr(expr_list)
 
     # Tuple(expr* elts, expr_context ctx)
     @with_line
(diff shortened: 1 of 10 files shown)


msullivan:
Could you add a python2 test and a test with a list lvalue and a `Tuple` annotation?

elazarg:
Thanks! Indeed there was a Python2 bug :)

elliott-beach:
For what it's worth, I think this is a cool change that should be merged.

mypy/fastparse.py
```diff
@@ -960,8 +960,11 @@ def visit_Name(self, n: ast3.Name) -> NameExpr:
 
     # List(expr* elts, expr_context ctx)
     @with_line
-    def visit_List(self, n: ast3.List) -> ListExpr:
-        return ListExpr([self.visit(e) for e in n.elts])
+    def visit_List(self, n: ast3.List) -> Union[ListExpr, TupleExpr]:
+        expr_list = [self.visit(e) for e in n.elts]  # type: List[Expression]
+        if isinstance(n.ctx, ast3.Store):
```
msullivan: Could you add a quick comment here and in fastparse2 explaining why?

elazarg: Done

Issue #3603, opened 2017-06-23
Multiple Inheritance Makes Mypy think an if branch is unreachable.


ilinum:
In the following example, the branch of an if-statement is not being typechecked.

```py
class Parent1: pass

class Parent2: pass

class Child(Parent1, Parent2): pass

def goo(p: Parent1) -> None:
    if isinstance(p, Parent2):
        # this branch is not typechecked
        1 + 'boom'  # no error from mypy here.


goo(Child())
```

Here is the output from `mypy` and `python`.
```
$ mypy i.py 
$ python3 i.py 
Traceback (most recent call last):
  File "i.py", line 13, in <module>
    goo(Child())
  File "i.py", line 10, in goo
    1 + 'boom'  # no error from mypy but there's an error at runtime
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

gvanrossum:
These bugs are pretty good finds! How'd you find them?

ilinum:
I am working on a patch for `--html-report` to mark unreachable code as `Any` (currently, it marks the code as typed).

So I just compared the output of new `--html-report` and the old one.

JukkaL:
This has come up before as well. Intersection types would perhaps be the cleanest way to support it, but just faking it by internally generating dummy multiply inheriting classes would perhaps work. So in the original example the type of `p` after `isinstance` would be a newly created empty class that looks like `Child` (it needs to be generated since generally something like `Child` can't be assumed to be visible to mypy at any given scope).

ilevkivskyi:
I would vote for implementing this using an _internal_ intersection type. It looks like ``Intersection`` appeared here and there several times, so that we might want to make it available in PEP 484 in the future.

JukkaL:
I'm still not excited about general intersection types, though I'm not quite sure if that is what you proposed. I expect that implementing them would require a large amount of work and would complicate future work on mypy features, whereas the hack I proposed would be a local change. It's hard to justify adding heavy machinery for this issue, which a relatively rare edge case.

Here's how we could implement semi-general intersection types without full-on general intersection types (similar to current `UnionType`). This would be a generalized version of my proposed hack -- we'd try to represent the intersection using only the types we already have at our disposal. 

We'd implement something like `make_fake_intersection(t1, t2)` that generates a dummy `TypeInfo` that basically is a subclass of `t1` and `t2` if both are `Instance` types. We can also support some other cases (these are just a few examples, and I'm not sure if all of these are worth it): 
[...]

PR #4343, opened 2017-12-12
Rework how we refine information from callable()

### LINKED ISSUE #3605: Calling `callable` in an if-condition makes mypy think branch in unreachable
In the following code, mypy thinking `callable(o)` always returns `False`. This might be related to #3603. 
```py
def f(o: object) -> None:
    if callable(o):
        # this branch is not typechecked
        o()
        1 + 'boom'  # no error from mypy
        o()

f(lambda: print('hello'))
```

Here is the output from `mypy` and `python`.
```
$ mypy n.py 
$ python3 n.py 
hello
  File "n.py", line 8, in <module>
    f(lambda: print('hello'))
  File "n.py", line 5, in f
    1 + 'boom'
TypeError: unsupported operand type(s) for +: 'int' and 'str'
```

msullivan:
Don't ever assume that something is *not* callable; basically anything
could be, and getting it wrong leaves code to not be typechecked.

When operating on an Instance that is not clearly callable, construct
a callable version of it by generating a fake subclass with a ``__call__``
method to be used.

Fixes #3605 

This is being posted to solicit feedback, I expect it might need some large changes.
The hack perpetrated here is somewhat grevious, and I suspect I missed something, so I am interested to find out what :).
Also, to reduce the change size during development I threaded a `TypeChecker` through a bunch of functions; the right thing might be to move all of those functions to be `TypeChecker` methods.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2659,7 +2659,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         # Build the fake ClassDef and TypeInfo together.
         # The ClassDef is full of lies and doesn't actually contain a body.
         # Use format_bare to generate a nice name for error messages.
-        # We skip filling fully filling out a handful of TypeInfo fields because they
+        # We skip fully filling out a handful of TypeInfo fields because they
         # should be irrelevant for a generated type like this:
         # is_protocol, protocol_members, is_abstract
         short_name = self.msg.format_bare(typ)
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2823,7 +2823,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         return function_type(func, self.named_type('builtins.function'))
 
     def find_isinstance_check(self, n: Expression) -> 'Tuple[TypeMap, TypeMap]':
-        return find_isinstance_check(n, self.type_map)
+        return find_isinstance_check(n, self.type_map, self)
 
     def push_type_map(self, type_map: 'TypeMap') -> None:
         if type_map is None:
@@ -2897,7 +2897,45 @@ def conditional_type_map(expr: Expression,
         return {}, {}
 
 
-def partition_by_callable(type: Type) -> Tuple[List[Type], List[Type]]:
+def intersect_instance_callable(type: Instance, callable_type: CallableType) -> Type:
+    """Creates a fake type that represents the intersection of an
+    Instance and a CallableType.
+
+    It operates by creating a bare-minimum dummy TypeInfo that
+    subclasses type and adds a __call__ method matching callable_type."""
+
+    # Build the fake ClassDef and TypeInfo together.
+    # The ClassDef is full of lies and doesn't actually contain a body.
+    cdef = ClassDef("<callable subtype of {}>".format(type.type.name()), Block([]))
+    info = TypeInfo(SymbolTable(), cdef, '<dummy>')
+    cdef.info = info
+    info.bases = [type]
+    info.calculate_mro()
+
+    # Build up a fake FuncDef so we can populate the symbol table.
+    func_def = FuncDef('__call__', [], Block([]), callable_type)
+    func_def.info = info
+    info.names['__call__'] = SymbolTableNode(MDEF, func_def, callable_type)
+
+    return Instance(info, [])
+
+
+def make_fake_callable(type: Instance, typechecker: TypeChecker) -> Type:
+    """Produce a new type that makes type Callable with a generic callable type."""
+
+    fallback = typechecker.named_type('builtins.function')
+    callable_type = CallableType([AnyType(TypeOfAny.explicit),
+                                  AnyType(TypeOfAny.explicit)],
+                                 [nodes.ARG_STAR, nodes.ARG_STAR2],
+                                 [None, None],
+                                 ret_type=AnyType(TypeOfAny.explicit),
+                                 fallback=fallback,
+                                 is_ellipsis_args=True)
+
+    return intersect_instance_callable(type, callable_type)
+
+
+def partition_by_callable(type: Type, typechecker: TypeChecker) -> Tuple[List[Type], List[Type]]:
     """Takes in a type and partitions that type into callable subtypes and
     uncallable subtypes.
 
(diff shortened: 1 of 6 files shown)


test-data/unit/check-callable.test
```diff
@@ -253,7 +253,7 @@ T = TypeVar('T', int, Callable[[], int], Union[str, Callable[[], str]])
 
 def f(t: T) -> None:
     if callable(t):
-        reveal_type(t())  # E: Revealed type is 'builtins.int'  # E: Revealed type is 'builtins.str'
+        reveal_type(t())  # E: Revealed type is 'Any'  # E: Revealed type is 'builtins.int'  # E: Revealed type is 'Union[Any, builtins.str]'
```
JukkaL: This change seems okay, but can you split the long line using `\`?

test-data/unit/check-callable.test
```diff
@@ -208,13 +208,13 @@ b = B()  # type: B
...
 if callable(a):
-    5 + 'test'
+    5 + 'test' # E: Unsupported operand types for + ("int" and "str")
 
 if not callable(b):
     5 + 'test'
 
 if callable(c):
-    reveal_type(c)  # E: Revealed type is '__main__.B'
+    reveal_type(c)  # E: Revealed type is 'Union[<callable subtype of A>, __main__.B]'
```
ilevkivskyi: Just a random comment: we need to be sure that this type ``<callable subtype of A>`` will never appear in error messages (apart from actually ``reveal_type``), since this can be confusing. I think it can be formatted just as ``A`` in other cases

ilevkivskyi:
> the right thing might be to move all of those functions to be ``TypeChecker`` methods.

I remember I also wanted this few times.

mypy/checker.py
```diff
@@ -2883,7 +2883,45 @@ def conditional_type_map(expr: Expression,
         return {}, {}
 
 
-def partition_by_callable(type: Type) -> Tuple[List[Type], List[Type]]:
+def intersect_instance_callable(type: Instance, callable_type: CallableType) -> Type:
```
JukkaL: Style nit: we generally prefer `typ` instead of `type` since `type` is a builtin :-( This doesn't apply to attribute access, since it's clear that `foo.type` doesn't refer to the builtin.

mypy/checker.py
```diff
@@ -2883,7 +2883,45 @@ def conditional_type_map(expr: Expression,
...
+    """Creates a fake type that represents the intersection of an
+    Instance and a CallableType.
+
+    It operates by creating a bare-minimum dummy TypeInfo that
+    subclasses type and adds a __call__ method matching callable_type."""
+
+    # Build the fake ClassDef and TypeInfo together.
+    # The ClassDef is full of lies and doesn't actually contain a body.
+    cdef = ClassDef("<callable subtype of {}>".format(type.type.name()), Block([]))
```
JukkaL: This is the trickiest thing in this PR. However, there generally seems to be no better way.

There are several things that should be handled here, such as:

* The original class may be generic.
* The original class may inherit from `Tuple[...]` and in that case we should return `TupleType`, not `Instance`.
* If the original class inherits from `Any`, the new class should also do that.

Basically you should check that every `TypeInfo` attribute is initialized correctly. 

You may be able to share code with `basic_new_typeinfo` in `mypy.semanal`, but you shouldn't add a dependency from type checker to the semantic analyzer. You could move `basic_new_typeinfo` to some third module, perhaps.

Additionally, it's possible to leak these types through type inference to class and module top level scopes. This means that the types can be serialized in incremental mode, and you should test that it works. It may be necessary to store the created class in a symbol table (with a generated unique dummy name) for serialization to work, but I'm not sure about this.

test-data/unit/check-callable.test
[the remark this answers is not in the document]
gvanrossum: I think nearly all users will be surprised by this. And I think in actual code it is very rare to find a situation where c (defined as Union[A, B]) was an instance of a callable subclass of A. We may see the occasional non-callable class that has a callable subclass, but it's unlikely to figure in type discriminations like this. The realistic scenarios probably involve either a union of callable and non-callable classes, where there is no concern that one of non-callable classes has a callable subclass; or a non-callable class that has a callable subclass (but then the discrimination is just between the base class and the callable subclass).

test-data/unit/check-callable.test
```diff
@@ -208,13 +208,13 @@ b = B()  # type: B
 c = A()  # type: Union[A, B]
 
 if callable(a):
-    5 + 'test'
+    5 + 'test' # E: Unsupported operand types for + ("int" and "str")
```
gvanrossum: Nit of nits: our coding standard has two spaces before `#`.

test-data/unit/check-callable.test
msullivan: Does there exist mechanism for having different internal and external names for instances?

ilevkivskyi: Yes, see ``messages.py``. There is a method ``MessageBuilder.format`` that controls most of the formatting. _But_, see other comments for this code that propose more radical solutions.

gvanrossum:
No, but you can special-case the way types are formatted with various
methods/functions in messages.py.

JukkaL:
On second thought, creating a new subclass might be bit of an overkill considering that this is not a very common use case. Here's an idea about a simpler "good enough" fix: 

* Only do anything with `callable` if the target type is a union.  For unions, use the existing approach to narrow them down.
* If the target is not a union, ignore the `callable` check -- the original type stays unmodified. This will fix the false negatives but can generate unwanted errors, but that's acceptable.

msullivan:
Oops, I wish I had noticed your last comment this morning.
That seems like a reasonable behavior, though this is somewhat nicer behavior to provide?

Pushed changes to address most of the comments. I still need to move the functions into the TypeChecker class, but I've put that off to keep the diffs easy to read.

msullivan:
Another way to structure this would be to return just a `Callable` from `conditional_callable_type_map` and push the generation of the intersecting class into `narrow_declared_type`. That would be a start towards implementing `make_fake_intersection` as discussed in #3603.

msullivan:
I think the `is_abstract` and protocol related fields aren't important. Doing `type(x)()` wouldn't try creating the abstract class, it would create the concrete class that x actually is. Similar with `isinstance` checks for a protocol type. The other stuff I think is now all either computed or correctly left blank. `_promote` being set in a parent class seems to do the right thing?

ilevkivskyi:
There are certain checks for protocols. For example, ``isinstance(x, P)`` is prohibited if ``P`` is a protocol (unless it is decorated with ``@runtime``).

msullivan:
Right, but the concern here was `isinstance(y, type(x))` where x is a protocol---but this isn't a problem, because `type(x)` is a regular concrete type.

mypy/checker.py
```diff
@@ -2615,6 +2641,306 @@ def visit_continue_stmt(self, s: ContinueStmt) -> None:
...
+        # have a valid fullname and a corresponding entry in a symbol table. We generate
+        # a unique name inside the symbol table of the current module.
+        cur_module = cast(MypyFile, self.scope.stack[0])
+        gen_name = gen_unique_name("<callable subtype of {}>".format(typ.type.name()),
+                                   cur_module.names)
+
+        # Build the fake ClassDef and TypeInfo together.
+        # The ClassDef is full of lies and doesn't actually contain a body.
+        # Use format_bare to generate a nice name for error messages.
+        # We skip filling fully filling out a handful of TypeInfo fields because they
```
JukkaL: Typo: 'filling fully filling out'.

gvanrossum:
🥇

Issue #4066, opened 2017-10-06
Stub functions should only typecheck for None


Wilfred:
This passes in mypy 0.521:

```python
def foo(): # type: () -> int
    pass
```

`mypy --py2 --strict demo.py` does not show any errors. If a function body does not contain a `return` statement or a `yield` statement it should only have a type of `None` I believe.

ilevkivskyi:
This is a special case in mypy. If a function have an empty body then it can have arbitrary return type. This is useful for testing, and is necessary for stubs. We have lots of tests that use this, so we can't disable this in non-stub files. Alternatively we can somehow detect whether we are running tests, and display an error otherwise.

JukkaL:
Mypy should probably eventually enforce this outside stubs. One way would be to make this an option (which doesn't need to be shown in help, perhaps) and turn it on by default, but we'd turn it off for most tests. Detecting stubs is easy, so it won't be a problem. We don't need to actually detect tests if all test runners just explicitly toggle the option.

ilevkivskyi:
@JukkaL 
> We don't need to actually detect tests if all test runners just explicitly toggle the option.

Yes, I was thinking about the same idea.

ilevkivskyi:
See also https://github.com/python/mypy/issues/2350

Wilfred:
Aha, I didn't realise that mypy treated stubs specially. My intent was to minimise my example, but I initially encountered this issue on a function that calls exit:

```python
import sys

def fatal_error(message):
    # type: (str) -> dict
    print "Oh no! {}".format(message)
    sys.exit(1)
```

I was expecting mypy to give a type error here.

JelleZijlstra:
That's expected: that function basically returns NoReturn, and NoReturn is a subtype of every type.

gvanrossum:
I propose to change the title of this issue to "Get rid of exception for functions with just pass as body" -- or something better (but I can't think of something better). The current title is too wide and in general incorrect.

Wilfred:
I've updated the issue title.

Wilfred:
@JelleZijlstra the NoReturn behaviour was surprising to me. Could I open an issue for a `--strict` option that disabled this behaviour, or is this working as intended?

ilevkivskyi:
> or is this working as intended?

This is intended. it is OK for a function to declare a return type wider than actual (``NoReturn`` in this case).

Wilfred:
[...]

Issue #2350, opened 2016-10-27
Handle empty function bodies safely


JukkaL:
Currently an empty function body is accepted if a function returns a non-`None` value (even with `--strict-optional`). This is mostly useful for abstract methods. This is unsound, since we don't prevent anybody from calling these empty functions. 

Here are some ideas for making this safe:
- Only allow empty function bodies for abstract methods. Other functions will need something like a return statement, a raise statement or `assert False` or `# type: ignore`.
- Tag abstract methods with an empty body with a flag in the AST, and disallow calls to them via `super()`.

This still wouldn't address empty functions defined in stubs, but perhaps can just assume that every function in a stub has a non-empty body.

See also discussion in #2339.

gvanrossum:
This would only apply to abstract methods whose return value *cannot* be None, right? I think this should be okay and allowed via super:
```python
class C:
    @abstractmethod
    def f(self) -> Optional[int]:
        pass
class D(C):
    def f(self) -> Optional[int]:
        return super().f()
```

JukkaL:
Yeah, we could safely allow those examples.

rwbarton:
The extensive use of functions with empty body and arbitrary type in the tests was another motivation for allowing empty bodies. Though it isn't a problem now, there would be a huge pain should `warn-no-return` (and perhaps `strict-optional`) be turned on by default. Not saying it should block this change, but it would be good to have some plan for dealing with this when it comes up.

gvanrossum:
I think @ddfisher would like both of these to become the default. David,
can you come up with a plan?

ddfisher:
I think it'd be reasonable not to worry about this too much for now: we're erring on the side of being more permissive (which is the better side to err on when introducing a new default), and I'd expect empty functions to be much more often intentional than not.

ilevkivskyi:
This keeps coming, so I propose to raise priority at least to normal.

ilevkivskyi:
Whatever we decide here might require updating thousands of tests cases, so adding the tests label.

Michael0x2a:
What if we modified the definition of a "trivial" function so that functions containing only `pass` are no longer considered to be trivial?

One additional benefit is that this change would now make the distinction between `...` and `pass` meaningful, rather than a matter of style. You'll now need to deliberately use the former if you want mypy to ignore the body.

gvanrossum:
That won't work in Python 2 though -- `...` by itself is not valid there.

ilevkivskyi:
[...]

PR #4344, opened 2017-12-12
Update docs to reflect the fact that Iterable etc. are protocols

### LINKED ISSUE #4339: Update docs to reflect the fact that Iterable etc. are now protocols

JukkaL:
Also update the discussion of ABCs, and don't describe protocols as
experimental.

Fixes #4339.

### DIFF
--- a/docs/source/class_basics.rst
+++ b/docs/source/class_basics.rst
@@ -117,11 +117,15 @@ annotations have no effect at runtime:
 Abstract base classes and multiple inheritance
 **********************************************
 
-Mypy uses Python abstract base classes for protocol types. There are
-several built-in abstract base classes types (for example,
-``Sequence``, ``Iterable`` and ``Iterator``). You can define abstract
-base classes using the ``abc.ABCMeta`` metaclass and the
-``abc.abstractmethod`` function decorator.
+Mypy supports Python abstract base classes (ABCs). Abstract classes
+have at least one abstract method or property that must be implemented
+by a subclass. You can define abstract base classes using the
+``abc.ABCMeta`` metaclass, and the ``abc.abstractmethod`` and
+``abc.abstractproperty`` function decorators.
+
+Note that mypy performs checking for unimplemented abstract methods
+even if you omit the ``ABCMeta`` metaclass. This can be useful if the
+metaclass would cause runtime metaclass conflicts.
 
 .. code-block:: python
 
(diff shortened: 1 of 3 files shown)


docs/source/class_basics.rst
```diff
@@ -267,6 +486,13 @@ be explicitly present if you are defining a protocol:
    # Error: nominal subtyping used by default
    x: NewProtocol = Concrete()  # Error!
 
+You can also include default implementations of methods in
+protocols. If you explicitly subclass these protocols you can inherit
+these default implementations. Explicitly including a protocol as a
+base class is also a way of documenting that your class implements a
+particular protocol, and it forces mypy to verify that your class
+implementation is actually compatible with the protocol.
```
ilevkivskyi: There is a caveat that this still works (although it shouldn't according to PEP 544):
```python
class P(Protocol):
    def meth(self) -> int: ...  # No error here...

class C(P):
    pass

C()  # ...and no error here!
```
mypy currently treats empty body (ellipsis) as a default implementation. There is an issue to allow this only in stubs (and  in tests), but it is not implemented yet. (The current workaround is to use ``@abstractmethod``, IIRC the similar problem with literal ellipsis in ``x: int = ...`` is already fixed.)

docs/source/class_basics.rst
```diff
@@ -174,16 +172,239 @@ a structural subtype of class ``C`` if the former has all attributes
 and methods of the latter, and with compatible types.
 
 Structural subtyping can be seen as a static equivalent of duck
-typing, which is well known to Python programmers. Mypy provides an
-opt-in support for structural subtyping via protocol classes described
+typing, which is well known to Python programmers. Mypy provides
+support for structural subtyping via protocol classes described
```
elazarg: Nitpicking question: the part about duck typing is technically not true (e.g. cross relations between variables are not captured even when statically analysable). What does it give to the reader that is not conveyed by the rest of the discussion? If all we want is help readers that search for the term "duck typing", perhaps we can phrase is differently - maybe "Structural subtyping is a closer approximation of duck typing than inheritance" or some other weak claim.
(Sorry, I know this is not important; the rest of the PR LGTM).

docs/source/class_basics.rst
ilevkivskyi: For the reference here are the issues: https://github.com/python/mypy/issues/4066, https://github.com/python/mypy/issues/2350. You can increase priority if you think this is important.

gvanrossum: I don't think we need to call this out in the docs. I do think it's important to fix. (Honestly I don't understand the difference between the two issues you link to.)

ilevkivskyi: > (Honestly I don't understand the difference between the two issues you link to.)

They are duplicates, but both have some discussion, so I don't know which one to close.

docs/source/class_basics.rst
```diff
@@ -174,16 +172,239 @@ a structural subtype of class ``C`` if the former has all attributes
...
+...............
+
+These protocols can be useful in async code.
+
+``Awaitable[T]``
+----------------
+
+.. code-block:: python
+
+   def __await__(self) -> Generator[Any, None, T]
```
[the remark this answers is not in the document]
ilevkivskyi: I just know that some people (for example @1st1) don't like mentioning generators in async docs.

PR #4353, opened 2017-12-13
[WIP] When a module has errors but the interface hash is the same, keep the cache

### LINKED ISSUE #4212: Daemon mode optimal behavior when there are no module interface changes
I'm recording an idea so that I won't lose track of it. It's possible be that this is how dmypy already works.

Assume that module `a` imports `b` and we are using dmypy. Assume that `a` is much larger than `b` (it could be many modules). Now we modify `b` so that the public interface doesn't change. We could consider `a` dirty and process it to ensure that there won't be any references to the old version of `b` in the in-memory cache. This would be expensive, however. An alternative would be to not process `a` but to just drop the module from the in-memory cache (while retaining the JSON cache file). This way any sequence of internal modifications to `b` would be very fast as they'd only require processing `b`.

@gvanrossum Please close this if this is how dmypy already works (or if this is a bad idea).

gvanrossum:
Closes #4212.

However, there are still many scenarios where a trivial error affects
the interface hash and hence causes a recheck of everything
downstream.

Also, this affects regular incremental mode too (making it faster, but
if there's a bug in my logic, it will also be hit by that bug.)

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1272,8 +1272,7 @@ def write_cache(id: str, path: str, tree: MypyFile,
     return interface_hash, cache_meta_from_dict(meta, data_json)
 
 
-def delete_cache(id: str, path: str, manager: BuildManager,
-                 keep_in_memory: bool = False) -> None:
+def delete_cache(id: str, path: str, manager: BuildManager) -> None:
     """Delete cache files for a module.
 
     The cache files for a module are deleted when mypy finds errors there.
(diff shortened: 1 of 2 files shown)


gvanrossum:
Looking at the test failures:
- testIncrementalInternalFunctionDefinitionChange is easy to fix (mod1 is unaffected by the error in mod2 since mod2's public API doesn't change)
- testCacheDeletedAfterErrorsFound3 however suggests I need to think harder, because here run 3 should still give the error (the cause is a change in b.py in run 2 that causes a.py to fail -- in run 3, b.py is unchanged, so the error in a.py should persevere)

gvanrossum:
No, the "different idea" may fix the last failing test but it doesn't consistently address the underlying problem. Tomorrow...

gvanrossum:
Well, this is really tricky. I think I may have to reconsider a lot more of the way dependencies and cache freshness are tracked...

PR #4356, opened 2017-12-13
[WIP] Fine-grained incremental mode integrated into daemon

JukkaL:
NOT READY FOR REVIEW

This has only been tested to work with mypy (for some level of working).

To start/restart the server:

```
$ python3 scripts/dmypy restart --log-file foo.log -- --experimental
```

To check or recheck mypy:

```
$ python3 scripts/dmypy check mypy
```

To verify that things are working, change a function signature in `mypy/nodes.py` and recheck. It should only take a few seconds to get results. Then revert your change and recheck. Again, this should be reasonably quick and faster than in normal daemon mode.

If the daemon crashes, look at `foo.log` for a traceback.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -19,6 +19,7 @@ from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence
 import mypy.build
 import mypy.errors
 import mypy.main
+import mypy.server.update
 from mypy.dmypy_util import STATUS_FILE, receive
 from mypy.gclogger import GcLogger
 
@@ -82,6 +83,12 @@ class Server:
     def __init__(self, flags: List[str]) -> None:
         """Initialize the server with the desired mypy flags."""
         self.saved_cache = {}  # type: mypy.build.SavedCache
+        if '--experimental' in flags:
+            self.fine_grained = True
+            self.fine_grained_initialized = False
+            flags.remove('--experimental')
+        else:
+            self.fine_grained = False
         sources, options = mypy.main.process_options(['-i'] + flags, False)
         if sources:
             sys.exit("dmypy: start/restart does not accept sources")
@@ -94,6 +101,10 @@ class Server:
         self.options = options
         if os.path.isfile(STATUS_FILE):
             os.unlink(STATUS_FILE)
+        if self.fine_grained:
+            options.incremental = True
+            options.show_traceback = True
+            options.cache_dir = os.devnull
 
     def serve(self) -> None:
         """Serve requests, synchronously (no thread or fork)."""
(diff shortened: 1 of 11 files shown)


mypy/dmypy_server.py
```diff
@@ -94,6 +101,10 @@ def __init__(self, flags: List[str]) -> None:
         self.options = options
         if os.path.isfile(STATUS_FILE):
             os.unlink(STATUS_FILE)
+        if self.fine_grained:
+            options.incremental = True
+            options.show_traceback = True
+            options.cache_dir = os.devnull
```
gvanrossum: Is this also going to be true for the "real" version of fine-grained?

mypy/dmypy_server.py
```diff
@@ -212,6 +236,72 @@ def check(self, sources: List[mypy.build.BuildSource],
...
+        self.previous_messages = messages[:]
+        if messages:
+            messages.append('.')
+        self.fine_grained_initialized = True
+        self.file_modified = {}  # type: Dict[str, float]
+        for source in sources:
+            assert source.path
+            self.file_modified[source.path] = os.stat(source.path).st_mtime
+        self.previous_sources = sources
+        return {'out': '\n'.join(messages), 'err': '', 'status': status}
```
gvanrossum: Could merge in manager.stats_summary() here (like in check_default()).

mypy/dmypy_server.py
```diff
@@ -212,6 +236,72 @@ def check(self, sources: List[mypy.build.BuildSource],
...
+            assert path
+            if not os.path.isfile(path):
+                changed.append((source.module, path))
+                if source.path in self.file_modified:
+                    del self.file_modified[source.path]
+        if not changed:
+            # Nothing changed -- just produce the same result as before.
+            messages = self.previous_messages
+        else:
+            messages = self.fine_grained_manager.update(changed)
```
gvanrossum: To get stats from this, just before this call add `self.fine_grained_manager.manager.stats.clear()` and just after add `stats = self.fine_grained_manager.manager.stats_summary()` (or some abstraction for this), then on line 298 below merge stats with the return dict.

mypy/dmypy_server.py
```diff
@@ -212,6 +236,72 @@ def check(self, sources: List[mypy.build.BuildSource],
...
+    def fine_grained_increment(self, sources: List[mypy.build.BuildSource]) -> Dict[str, Any]:
+        changed = []
+        for source in sources:
+            path = source.path
+            assert path
+            mtime = os.stat(path).st_mtime
+            if path not in self.file_modified or self.file_modified[path] != mtime:
+                self.file_modified[path] = mtime
+                changed.append((source.module, path))
+        omitted = self.find_omitted(sources)
```
gvanrossum: Seems odd to have an abstraction for find_omitted() but not for the code surrounding it here. :-)

JukkaL:
All of these changes have now been merged in separate PRs so this is no longer relevant.

PR #4362, opened 2017-12-14
Fix mypy self check on Windows

ethanhs:
Noticed this working on #4278 :)

Perhaps we should add self check back to appveyor?

### DIFF
--- a/mypy/test/testdmypy.py
+++ b/mypy/test/testdmypy.py
@@ -30,7 +30,7 @@ if sys.platform != 'win32':
         'check-newtype.test',
     ]
 else:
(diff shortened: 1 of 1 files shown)


mypy/test/testdmypy.py
```diff
@@ -30,7 +30,7 @@
         'check-newtype.test',
     ]
 else:
-    dmypy_files = []
+    dmypy_files = []  # List[str]
```
JelleZijlstra: Do you mean `# type: List[str]`?

ethanhs: How silly of me, yes I did.

PR #4367, opened 2017-12-15
Allow calling TupleTypes if the fallback can be called

### LINKED ISSUE #4360: NamedTuple subclasses are not considered callable even when they have a __call__ method
```
from typing import Tuple, NamedTuple

Thing = NamedTuple('Thing', [('s', str), ('n', int)])

class CallableTuple(Thing):
    def __call__(self) -> str:
        return self.s + str(self.n)

o = CallableTuple('hello ', 12)
print(o())
```
produces the message
```error: "CallableTuple" not callable```

msullivan:
Fixes #4360

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -639,6 +639,10 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             item = self.analyze_type_type_callee(callee.item, callee)
             return self.check_call(item, args, arg_kinds, context, arg_names,
                                    callable_node, arg_messages)
+        elif isinstance(callee, TupleType):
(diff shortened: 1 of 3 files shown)


gvanrossum:
Would you consider adding a test where `__call__` takes args and returns a relevant value? (Just being paranoid here.)

msullivan:
Added more testing. Our documentation claims that `Tuple` can't be subclassed directly, which is why I was testing with `NamedTuple`.

ilevkivskyi:
> Added more testing. Our documentation claims that ``Tuple`` can't be subclassed directly, which is why I was testing with ``NamedTuple``.

The docs need to be updated, would you like to make a PR?

PR #4369, opened 2017-12-15
Remove myunit

elazarg:
This should conclude #1673. It uses `unittest.TestCase` as a shortcut, as suggested by @JukkaL. It is aliased as `Suite`.

The transition is crude, aims at functionality; I intend to clean it up in a later PRs.

* Helper functions moved from from `myunit.__init__` to `test.helpers`. I think some of the asserts should be changed to simple `assert` to take advantage of pytest's introspection.
* BaseTestCase is inlined (and perhaps was never needed).
* testargs fails without the change to `option.py`. I'm not sure why it hadn't fail before.
* pytest fixtures are not used yet

As in previous changes, using `pytest -n0 --collect-only` can give you the list of tests; diffing it with the result of the same command on master shows difference of 155 test items. `grep '    def test_' mypy/test/test*.py | wc -l` gives 162 items, and there are 7 skipped items.


ethanhs:
This is excellent! I have a branch that moves several of the assert functions to bare asserts, if that is desired. Also, I believe the original idea was that replacement of the test framework would be concluded when `runtests.py` was replaced entirely by pytest? So I don't think this quite concludes #1673. But it certainly is great progress!

elazarg:
@ethanhs thanks! Yeah, maybe "conclude" is not accurate. I hope this PR does not conflict with your branch, it will be very useful.

ethanhs:
@elazarg well, I just checked and it already has quite a few conflicts with master, so this branch can't hurt more than the conflicts that already exist. No worries :)

mypy/test/helpers.py
```diff
@@ -308,3 +249,52 @@ def retry_on_error(func: Callable[[], Any], max_wait: float = 1.0) -> None:
                 # Done enough waiting, the error seems persistent.
                 raise
             time.sleep(wait_time)
+
+
+class AssertionFailure(Exception):
```
ethanhs: This Exception and the `assert*`s below are the targets I mentioned replacing. They entirely duplicate pytest functionality.

mypy/test/testargs.py
```diff
@@ -5,13 +5,15 @@
...
+from mypy.options import Options
 from mypy.main import process_options
 
 
 class ArgSuite(Suite):
     def test_coherence(self) -> None:
         options = Options()
         _, parsed_options = process_options([], require_targets=False)
-        assert_equal(options, parsed_options)
+        assert options.__dict__.keys() == parsed_options.__dict__.keys()
```
ethanhs: Could you explain why you chose this method of checking equality? `assert_equal` is essentially `assert not a != b`, I believe this could be a bit different, and is not as straightforward.

mypy/test/testsubtypes.py
```diff
@@ -67,10 +67,8 @@ def test_interface_subtyping(self) -> None:
         self.assert_equivalent(self.fx.f, self.fx.f)
         self.assert_not_subtype(self.fx.a, self.fx.f)
 
+    @skip
```
ethanhs: Perhaps still leave the TODO item comment here and for the other `@skip`s? I sometimes grep for them from time to time.

pytest.ini
```diff
@@ -14,7 +14,7 @@ python_files = test*.py
 
 # Because we provide our own collection logic, disable the default
 # python collector by giving it empty patterns to search for.
-python_classes =
+python_classes = *Suite
```
ethanhs: Perhaps it would be better to explicitly check via e.g. isinstance checks in the custom collector?

mypy/test/testargs.py
elazarg: I added these checks in order to find out why it failed. Since it was helpful for me, I thought it might be better not to switch back.

mypy/test/testsubtypes.py
elazarg: I am thinking about `skip` as some kind of TODO... I can leave the TODO, but you might want to grep for skips too.

pytest.ini
elazarg: The custom collector is for the data-driven tests. It doesn't know about tests derived from TestCase, and I think it might be awkward to make it handle the non-TestCase, non-data-driven tests. I am still thinking about the right way to collect tests, but I prefer to work on it in following PRs.

mypy/test/helpers.py
elazarg: Do you want me to do anything about this in current PR? I wanted to leave these changes to the following PRs.

ethanhs: No, I don't think that is needed.

mypy/options.py
```diff
@@ -129,7 +129,7 @@ def __init__(self) -> None:
         self.scripts_are_modules = False
 
         # Config file name
-        self.config_file = None  # type: Optional[str]
+        self.config_file = 'setup.cfg'  # type: Optional[str]
```
ethanhs: Isn't this just `str` now? (no longer `Optional`?)

elazarg: I think it isn't. Looks like it's possible to have None if there's no config file; but when running the test, there is. I'm not sure about that though; I wonder why ArgSuite doesn't fail on master.

elazarg:
I'm not sure why the conflicts don't go away.

ethanhs:
For some reason I've had to fix merge conflicts twice for github to register it. This seems like a bug in Github. I went ahead and did the second merge, things seem to not conflict now.

pytest.ini
msullivan: What is the reason for this change?

msullivan:
This looks pretty reasonable to me. It's always nice when you can delete a lot of code. @JukkaL, is there any reason to not merge this?

mypy/options.py
msullivan: What's the reason for this change?

JukkaL:
I had look at this a while ago and it looked good, but then I got distracted by fine-grained incremental work and unfortunately didn't finish my review. I'll do another pass this week. Sorry about the slow turnaround -- it will great be to finally drop myunit!

mypy/options.py
JukkaL: This change doesn't seem to be related to the rest of PR. Can you remove this?

mypy/test/data.py
```diff
@@ -647,8 +662,12 @@ class DataSuite:
     optional_out = False  # type: bool
     native_sep = False  # type: bool
 
-    def __init__(self, *, update_data: bool) -> None:
-        self.update_data = update_data
+    # Assigned from MypyDataCase.runtest
+    update_data = False  # type: bool
```
JukkaL: Style nit: redundant type annotation (you can also remove the `str` and `bool` annotations above).

mypy/test/helpers.py
```diff
@@ -312,6 +255,55 @@ def retry_on_error(func: Callable[[], Any], max_wait: float = 1.0) -> None:
             time.sleep(wait_time)
 
 
+class AssertionFailure(Exception):
```
JukkaL: You could probably use the built-in exception `AssertionError` instead of defining a new class here.

mypy/test/helpers.py
```diff
@@ -312,6 +255,55 @@ def retry_on_error(func: Callable[[], Any], max_wait: float = 1.0) -> None:
...
+    """Exception used to signal failed test cases."""
+    def __init__(self, s: Optional[str] = None) -> None:
+        if s:
+            super().__init__(s)
+        else:
+            super().__init__()
+
+
+def assert_true(b: bool, msg: Optional[str] = None) -> None:
```
JukkaL: Add comment about `assert_true` and `assert_false` being redundant (we could just use the assert statement).

mypy/test/testsolve.py
```diff
@@ -2,16 +2,15 @@
...
 from mypy.constraints import SUPERTYPE_OF, SUBTYPE_OF, Constraint
 from mypy.solve import solve_constraints
 from mypy.test.typefixture import TypeFixture
 from mypy.types import Type, TypeVarType, TypeVarId
 
 
 class SolveSuite(Suite):
-    def __init__(self) -> None:
-        super().__init__()
+    def setUp(self) -> None:
```
JukkaL: Elsewhere we had a `setup` method, which is a bit confusing. This is needed because of `unittest`. I wonder if it make sense to use this name variant uniformly?

mypy/test/helpers.py
```diff
@@ -3,14 +3,18 @@
...
 import time
 
-from typing import List, Dict, Tuple, Callable, Any
+from typing import List, Dict, Tuple, Callable, Any, Optional
 
 from mypy import defaults
-from mypy.myunit import AssertionFailure
+
+import pytest  # type: ignore  # no pytest in typeshed
+from unittest import TestCase as Suite
```
JukkaL: Add comment about this, since this may be kind of unexpected. In the future we'd probably want to directly import `TestCase` in test files and inherit from it instead of the `Suite` alias (and maybe rename some of the other helper classes to avoid ambiguity).

pytest.ini
JukkaL: The comment above this line is no longer up-to-date.

mypy/options.py
elazarg: It is needed because (as far as I understand) pytest uses the config file, so the default option is not None. I'm not sure this change is the right one; another option I can think of is changing this value inside the test.

JukkaL: But this is the mypy config file, which I don't expect to be directly relevant for pytest. Pytest config file should be configured somewhere else (not sure how this works exactly, though) -- and pytest already has a config file (`pytest.ini`) which seems to be used. Does this PR do something that requires changes to this? Or maybe I'm talking about a different config file?

mypy/test/testsolve.py
elazarg: If you insist, but I'm not sure that's desirable. The naming distinction makes it clear where it's required by unittest and where it isn't.

mypy/options.py
elazarg: I think I might have confused something in my previous response (it's been a while since I tried to figure this out), but note that `SHARED_CONFIG_FILES = ('setup.cfg',)` is hardcoded in `main.py`. Somewhere in `parse_config_file()` there's probably a difference between how myunit and pytest work, so one the former uses `None` and the latter uses `'setup.cfg'`.

elazarg: Running myunit, `os.path.exists('setup.cfg')` is false. Running pytest, it is true.

elazarg: I'm trying a different approach - simply create a new directory and chdir into it.

elazarg: Failed - changing directories did not work on Windows.

mypy/test/testsolve.py
JukkaL: Ok no need to change this. If this becomes a problem we can reconsider.

JelleZijlstra:
Hooray!

PR #4396, opened 2017-12-20
Flush error messages incrementally after processing a file

### LINKED ISSUE #1294: Flush buffered errors after each file is processed
Currently we only flush errors at the very end of a run. It might be nicer if we flushed errors after each file is fully processed (i.e. once the type checking pass has finished). We buffer errors so we can sort them and de-dupe them, but there shouldn't be any new errors for a file once it's been processed.

msullivan:
In order to avoid duplicate error messages for errors produced in both
load_graph() and process_graph() and to prevent misordered error
messages in a number of places, lists of error messages are now
tracked per-file.

These lists are collected and printed out when a file is complete.  To
maintain consistency with clients that use .messages() (namely,
tests), messages are generated file-at-a-time even when not printing
them out incrementally.

Fixes #1294

### DIFF
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -95,8 +95,8 @@ class Errors:
     # files were processed.
     error_info_map = None  # type: Dict[str, List[ErrorInfo]]
 
-    # The size of error_info the last time that error messages were flushed
-    new_errors_start_map = None  # type: Dict[str, int]
+    # Errors that haven't been flushed out yet
+    fresh_error_info_map = None  # type: Dict[str, List[ErrorInfo]]
 
     # A cache of the formatted messages
     formatted_messages = None  # type: List[str]
@@ -150,7 +150,7 @@ class Errors:
 
     def initialize(self) -> None:
         self.error_info_map = OrderedDict()
-        self.new_errors_start_map = defaultdict(int)
+        self.fresh_error_info_map = OrderedDict()
         self.import_ctx = []
         self.formatted_messages = []
         self.error_files = set()
@@ -302,7 +302,9 @@ class Errors:
     def _add_error_info(self, info: ErrorInfo) -> None:
         if info.file not in self.error_info_map:
             self.error_info_map[info.file] = []
+            self.fresh_error_info_map[info.file] = []
         self.error_info_map[info.file].append(info)
+        self.fresh_error_info_map[info.file].append(info)
 
     def add_error_info(self, info: ErrorInfo) -> None:
         (file, line) = cast(Tuple[str, int], info.origin)  # see issue 1855
@@ -365,12 +367,13 @@ class Errors:
         """
         # self.new_messages() will format all messages that haven't already
         # been returned from a new_module_messages() call. Count how many
-        # we've seen before that.
+        # we've seen before that so we can determine which are fresh.
         already_seen = len(self.formatted_messages)
-        raise CompileError(self.messages(),
+        messages = self.messages()
+        raise CompileError(messages,
                            use_stdout=True,
                            module_with_blocker=self.blocker_module(),
-                           num_already_seen=already_seen)
+                           fresh_messages=messages[already_seen:])
 
     def format_messages(self, error_info: List[ErrorInfo]) -> List[str]:
         """Return a string list that represents the error messages.
@@ -406,8 +409,8 @@ class Errors:
         """
         if path not in self.error_info_map:
             return []
-        msgs = self.format_messages(self.error_info_map[path][self.new_errors_start_map[path]:])
-        self.new_errors_start_map[path] = len(self.error_info_map[path])
+        msgs = self.format_messages(self.fresh_error_info_map[path])
+        self.fresh_error_info_map[path] = []
         self.formatted_messages += msgs
         return msgs
 
@@ -518,7 +521,7 @@ class Errors:
     def sort_messages(self, errors: List[ErrorInfo]) -> List[ErrorInfo]:
         """Sort an array of error messages locally by line number.
 
-        I.e., sort a run of consecutive messages with the same file
+        I.e., sort a run of consecutive messages with the same
         context by line number, but otherwise retain the general
         ordering of the messages.
         """
@@ -574,18 +577,18 @@ class CompileError(Exception):
     use_stdout = False
     # Can be set in case there was a module with a blocking error
     module_with_blocker = None  # type: Optional[str]
-    num_already_seen = 0
+    fresh_messages = None  # type: List[str]
 
     def __init__(self,
                  messages: List[str],
                  use_stdout: bool = False,
                  module_with_blocker: Optional[str] = None,
-                 num_already_seen: int = 0) -> None:
+                 fresh_messages: Optional[List[str]] = None) -> None:
         super().__init__('\n'.join(messages))
         self.messages = messages
         self.use_stdout = use_stdout
         self.module_with_blocker = module_with_blocker
-        self.num_already_seen = num_already_seen
+        self.fresh_messages = fresh_messages if fresh_messages is not None else messages
 
 
 class DecodeError(Exception):
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -51,6 +51,9 @@ class ErrorInfo:
     # Only report this particular messages once per program.
     only_once = False
 
+    # Actual origin of the error message
+    origin = None  # type: Tuple[str, int]
+
     # Fine-grained incremental target where this was reported
     target = None  # type: Optional[str]
 
@@ -291,13 +294,13 @@ class Errors:
                          target=self.current_target())
         self.add_error_info(info)
 
-    def _add_error_info(self, info: ErrorInfo) -> None:
-        if info.file not in self.error_info_map:
-            self.error_info_map[info.file] = []
-        self.error_info_map[info.file].append(info)
+    def _add_error_info(self, file: str, info: ErrorInfo) -> None:
+        if file not in self.error_info_map:
+            self.error_info_map[file] = []
+        self.error_info_map[file].append(info)
 
     def add_error_info(self, info: ErrorInfo) -> None:
-        (file, line) = cast(Tuple[str, int], info.origin)  # see issue 1855
+        file, line = info.origin
         if not info.blocker:  # Blockers cannot be ignored
             if file in self.ignored_lines and line in self.ignored_lines[file]:
                 # Annotation requests us to ignore all errors on this line.
@@ -309,7 +312,7 @@ class Errors:
             if info.message in self.only_once_messages:
                 return
             self.only_once_messages.add(info.message)
-        self._add_error_info(info)
+        self._add_error_info(file, info)
 
     def generate_unused_ignore_notes(self, file: str) -> None:
         ignored_lines = self.ignored_lines[file]
@@ -319,7 +322,7 @@ class Errors:
                 info = ErrorInfo(self.import_context(), file, self.current_module(), None,
                                  None, line, -1, 'note', "unused 'type: ignore' comment",
                                  False, False)
-                self._add_error_info(info)
+                self._add_error_info(file, info)
 
     def is_typeshed_file(self, file: str) -> bool:
         # gross, but no other clear way to tell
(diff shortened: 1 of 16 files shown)


mypy/errors.py
```diff
@@ -375,12 +396,48 @@ def messages(self) -> List[str]:
...
+        for key in self.error_info_map.keys():
+            msgs.extend(self.new_file_messages(key))
+        return msgs
+
+    def messages(self) -> List[str]:
+        """Return a string list that represents the error messages.
+
+        Use a form suitable for displaying to the user.
+        """
+        self.new_messages()
```
JukkaL: Add comment about `new_messages()` storing new messages as a side effect (or potentially rename the method to something that makes this more explicit).

mypy/test/testerrorstream.py
```diff
@@ -0,0 +1,76 @@
...
+    options.show_traceback = True
+
+    a = []
+
+    def flush_errors(msgs: List[str]) -> None:
+        nonlocal a
+        if msgs:
+            a.append('==== Errors flushed ====')
+            a += msgs
+    plugin = ChainedPlugin(options, [LoggingPlugin(options, flush_errors), DefaultPlugin(options)])
```
JukkaL: Style nit: Add empty line after nested function for clarity.

test-data/unit/errorstream.test
```diff
@@ -0,0 +1,46 @@
...
+-- a plugin when a call to it is checked, which can be used to verify that
+-- error messages are printed before doing later typechecking work.
+--
+-- The input file name in errors is "file".
+--
+-- Comments starting with "--" in this file will be ignored, except for lines
+-- starting with "----" that are not ignored. The first two dashes of these
+-- lines are interpreted as escapes and removed.
+
+[case testErrorStream]
```
JukkaL: Add test case with an import cycle? You can perhaps use "deferred nodes" to interleave messages from two modules (like 'error from module a', 'error from module b', 'error from module a') and then you can test that the errors are grouped by file correctly.

mypy/main.py
```diff
@@ -62,13 +62,27 @@ def main(script_path: Optional[str], args: Optional[List[str]] = None) -> None:
...
+        f = sys.stderr if serious else sys.stdout
+        try:
+            for m in a:
+                f.write(m + '\n')
+        except BrokenPipeError:
+            pass
+
     try:
-        res = type_check_only(sources, bin_dir, options)
+        res = type_check_only(sources, bin_dir, options, flush_errors)
```
JukkaL: I'd prefer if the main body of test cases (`mypy/test/testscheck.py`) would use streaming errors (without testing that flushing happens as expected, i.e. we'd still only test that the sequence of errors in the output is as expected). Not sure if this is feasible without many changes to test cases. If not, at least it would be good to do a one-off manual check that streaming produces the same errors as not streaming in test cases.

test-data/unit/errorstream.test
```diff
@@ -0,0 +1,46 @@
...
+import b
+[file a.py]
+1 + ''
+[file b.py]
+import a
+break
+[out]
+==== Errors flushed ====
+a.py:1: error: Unsupported operand types for + ("int" and "str")
+==== Blocking error ====
```
JukkaL: What about also adding a third error that would be get reported after the blocking error (but doesn't get reported, because there was a blocking error I assume)?

msullivan:
Time to report an error in `mypy/errors.py` was 4.3s with this patch and 11.8s without it.

mypy/messages.py
```diff
@@ -153,8 +153,9 @@ def copy(self) -> 'MessageBuilder':
     def add_errors(self, messages: 'MessageBuilder') -> None:
         """Add errors in messages to this builder."""
         if self.disable_count <= 0:
-            for info in messages.errors.error_info:
-                self.errors.add_error_info(info)
+            for errs in messages.errors.error_info_map.values():
```
gvanrossum: Oh. This is where `error_info_map` is used. Though of course in this case the errors will never be flushed...

mypy/test/testcheck.py
```diff
@@ -170,15 +170,22 @@ def run_case_once(self, testcase: DataDrivenTestCase, incremental_step: int = 0)
...
             res = build.build(sources=sources,
                               options=options,
-                              alt_lib_path=test_temp_dir)
+                              alt_lib_path=test_temp_dir,
+                              flush_errors=flush_errors)
             a = res.errors
         except CompileError as e:
             a = e.messages
         a = normalize_error_messages(a)
+        streamed_messages = normalize_error_messages(streamed_messages)
```
gvanrossum: I was quite surprised that this could work. Until I realized that `BuildResult` calls `errors.messages()` which returns *all* messages (redundantly). This complexity increase worries me again).

mypy/test/testerrorstream.py
```diff
@@ -0,0 +1,82 @@
...
+
+    plugin = ChainedPlugin(options, [LoggingPlugin(options, flush_errors), DefaultPlugin(options)])
+
+    sources = [BuildSource('main', '__main__', '\n'.join(testcase.input))]
+    try:
+        res = build.build(sources=sources,
+                          options=options,
+                          alt_lib_path=test_temp_dir,
+                          flush_errors=flush_errors,
+                          plugin=plugin)
```
gvanrossum: This is the only place where we pass a plugin to `build()`. I wish there was another way to do this... (It's pretty clever though. :-)

test-data/unit/cmdline.test
```diff
@@ -400,9 +400,9 @@ bla bla
 [file error.py]
 bla bla
 [out]
+normal.py:2: error: Unsupported operand types for + ("int" and "str")
```
gvanrossum: Ah, so this is an example where the errors for different files are interspersed. Nifty! :-)

test-data/unit/errorstream.test
```diff
@@ -0,0 +1,72 @@
+-- Test cases for incremental error streaming. Each test case consists of two
+-- sections.
+-- The first section contains [case NAME] followed by the input code, while
```
gvanrossum: This sentence is redundant, since all .test files work that way.

test-data/unit/errorstream.test
```diff
@@ -0,0 +1,72 @@
...
+-- the second section contains [out] followed by the output from the checker.
+-- Each time errors are reported, '==== Errors flushed ====' is printed.
+-- The log.log_checking() function will immediately emit a message from
+-- a plugin when a call to it is checked, which can be used to verify that
+-- error messages are printed before doing later typechecking work.
+--
+-- The input file name in errors is "file".
```
gvanrossum: What does this mean?

test-data/unit/errorstream.test
```diff
@@ -0,0 +1,72 @@
...
+--
+-- Comments starting with "--" in this file will be ignored, except for lines
```
gvanrossum: This is also redundant.

mypy/test/testerrorstream.py
```diff
@@ -0,0 +1,82 @@
...
+    options.show_traceback = True
+
+    logged_messages = []  # type: List[str]
+    real_messages = []  # type: List[str]
+
+    def flush_errors(msgs: List[str], serious: bool, is_real: bool=True) -> None:
+        if msgs:
+            logged_messages.append('==== Errors flushed ====')
+            logged_messages.extend(msgs)
+        if is_real:
```
gvanrossum: This block can also be indented.

mypy/test/testerrorstream.py
```diff
@@ -0,0 +1,82 @@
...
+
+    The argument contains the description of the test case.
+    """
+    options = Options()
+    options.show_traceback = True
+
+    logged_messages = []  # type: List[str]
+    real_messages = []  # type: List[str]
+
+    def flush_errors(msgs: List[str], serious: bool, is_real: bool=True) -> None:
```
gvanrossum: There should be spaces around the `=` (it's a special case in PEP 8 :-).

mypy/test/testerrorstream.py
msullivan: Another approach would be arrange to load a plugin through the regular mechanisms---that requires the plugin communicating with the test via global state, though.

Another another approach is to ditch the plugin because the possible failure tests for (errors being streamed out properly but not until all typing is done) seems pretty marginal anyways.

gvanrossum: I'm okay with ditching the plugin.

gvanrossum:
Another thought: let BuildResult *only* collect messages that haven't been flushed yet. This means that in general there are three ways you can get messages: (1) via the flush_errors callback, if set; (2) via BuildResult, if no flush_errors callback is set; (3) via CompileError. This way you can keep the old way of keeping track of messages, and you don't need to cache formatted messages. The tests simply pass flush_errors=None and get the message from BuildResult.

msullivan:
The benefit of the caching scheme and returning the messages even when streaming is on is that it makes it easy to ensure that the streaming and the fixed interfaces return the same messages in the same order and also to test that. If we think that is important, ditching the caching would I think require some other machinery.

gvanrossum:
I'm pretty confident about the ordering regardless, so I'd rather do without the cache scheme etc.

msullivan:
Getting to the ordering to match without caching will require some new machinery not in the current patches: the errors are streamed out in the order that SCCs are processed, but this might not match the order in the `OrderedDict` if errors were generated at parse time. The machinery might be simpler than the caching though, so.

msullivan:
Nevermind, I have what I think is a good plan.

gvanrossum:
Aren't parse-time errors always fatal (blocking)?

msullivan:
Actual parse errors are blocking but the first pass of semantic analysis is done immediately after parsing, and those can be nonblocking.

mypy/errors.py
```diff
@@ -554,7 +596,8 @@ def report_internal_error(err: Exception, file: Optional[str], line: int,
     # Dump out errors so far, they often provide a clue.
     # But catch unexpected errors rendering them.
     try:
-        for msg in errors.messages():
+        errors.flushed_files = set()  # Print out already flushed messages too
```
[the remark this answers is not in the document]
msullivan: My thought was that I wanted to always make sure that all of the messages printed, even in the cases where the messages were being buffered in build.build. But I think you are right and we would rather lose some messages while running tests than confuse matters by printing duplicate messages in actual use.

mypy/errors.py
```diff
@@ -90,15 +93,17 @@ class Errors:
     current error context (nested imports).
     """
 
-    # List of generated error messages.
-    error_info = None  # type: List[ErrorInfo]
+    # Map from files to generated error messages. Is an OrderedDict so
```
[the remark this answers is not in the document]
msullivan: Not all error infos have a module, unfortunately.

mypy/options.py
```diff
@@ -34,6 +34,7 @@ class Options:
         "show_none_errors",
         "warn_no_return",
         "warn_return_any",
+        "warn_unused_ignores",
```
gvanrossum: Oh, now the docs also need to be updated (it has separate sections for global and per-module flags).

gvanrossum:
Congrats! A nice piece of work. I'll merge the other thing too as soon as I remember what it was.

marcintustin:
This is breaking pytest integration, any chance of a bugfix release any time soon?

gvanrossum:
@marcintustin Please file a separate bug report with more details.

marcintustin:
@gvanrossum I'm not saying that this PR is causing a problem. I'm saying that the lack of a release including this PR is causing a problem. You still want a separate bug report?

JelleZijlstra:
There has been a release since this was merged, so I'm not sure what you're referring to.

marcintustin:
@JelleZijlstra @gvanrossum Well I suspect then that I've misunderstood the chatter on the issue where this is linked as the cause of the pytest integration problem. :( Apologies; I'll chase up on the integration first.

ethanhs:
@marcintustin this PR broke things as filed in #4681. That issue is missing that it causes errors in pytest-mypy. I'm going to go ahead and submit a fix for it. @msullivan I think we should probably have a .571 release.

marcintustin:
@ethanhs You're quite right. Thanks for clarifying.

Issue #3135, opened 2017-04-05
Class decorators are not type checked


JelleZijlstra:
```python
decorator: int = 3

@decorator  # no error
class A:
    pass

class B:
    pass
NewB = decorator(B)  # E: "int" not callable
```
Mypy doesn't appear to do any checking of class decorators. Applying the decorator manually works.

JelleZijlstra:
Guido and I found out about this in https://github.com/python/typeshed/pull/1136#discussion_r109725629. Looks like this can be fixed somewhere around `analyze_class_decorator` in semanal.py.

ilevkivskyi:
A use case reported in #3483
```python
from typing import Type

class A:
    pass

def deco(cls) -> Type[A]:
    return A

@deco
class B:
    pass

def func(cls: Type[A]):
    assert cls is A

func(B)
```
might be non-trivial to implement. I am adding this example here, so that we will not forget about it.

euresti:
Another possibly easier to solve problem is that the functions called are not type-checked

```
from typing import TypeVar, Callable
_C = TypeVar('_C', bound=type)
def blah(x:int=17) -> Callable[[_C], _C]: ...

@blah(y=18)
class A:
    pass

blah(y=18)
```

Or even more fun
```
x=17

@x(huh=17)
class B:
    pass

x(huh=17)
```

reinhrst:
I'm thinking how this could/should be used in combination with `Protocol`s to deal with class decorators adding some methods to a class

```
class FooBar(Protocol):
   def bar(self) -> int:
     return 1

T = TypeVar("T", bound=Type)
def add_bar(cls: T) -> Intersection[FooBar, T]:
   def bar(self) -> int:
       return 2
  cls.bar = bar

@add_bar
class Foo: pass

Foo().bar()
```

Now this depends on `Intersection` which was described in PEP483, but not implemented; I don't see another way of doing this cleanly right now.

erezsh:
Are there plans to fix this? It's strange that mypy doesn't support a core syntax.

rileyjohngibbs:
It looks like some of these examples (e.g. the `x=17` example above) may have been fixed, so in the interest of having a current, non-working example:
```python
from typing import Any

def foo(_: Any) -> int:
    return 5

@foo
class Foo:
    pass

def test_foo() -> None:
    assert Foo + 7 == 12  # Does not pass typechecking
```
[...]

PR #4397, opened 2017-12-20
Plugin to typecheck attrs-generated classes

### LINKED ISSUE #2088: Typechecking attrs-generated classes

euresti:
See http://www.attrs.org/en/stable/how-does-it-work.html for information on how attrs works.

The plugin walks the class declaration (including superclasses) looking for "attributes" then depending on how the decorator was called, makes modification to the classes as follows:
* `init=True` adds an `__init__` method.
* `cmp=True` adds all of the necessary `__cmp__` methods.
* `frozen=True` turns all attributes into properties to make the class read only.
* Remove any `@x.default` and `@y.validator` decorators which are only part of class creation.

Fixes #2088


mypy/plugin.py
```diff
@@ -379,3 +392,189 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+    from mypy.semanal import set_callable_name
+
+    first = [Argument(Var('self'), self_type, None, ARG_POS)]
+    args = first + args
+
+    arg_types = [arg.type_annotation for arg in args]
+    arg_names = [arg.variable.name() for arg in args]
+    arg_kinds = [arg.kind for arg in args]
+    assert None not in arg_types
+    signature = CallableType(cast(List[Type], arg_types), arg_kinds, arg_names,
```
ilevkivskyi: Is it possible to avoid the ``cast``, why it is necessary? We try to avoid using casts.

mypy/plugin.py
```diff
@@ -379,3 +392,189 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                return attr_value
+            if name and attr_name == name:
+                return attr_value
+        return None
+
+    def get_bool_argument(call: CallExpr, name: str,
+                          default: Optional[bool]) -> Optional[bool]:
+        for arg_name, arg_value in zip(call.arg_names, call.args):
+            if arg_name == name:
+                # TODO: Handle None being returned here.
```
ilevkivskyi: Don't forget to fix TODO items before the PR is merged.

mypy/plugin.py
```diff
@@ -379,3 +392,189 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+        cmp = get_bool_argument(decorator, "cmp", cmp)
+        auto_attribs = get_bool_argument(decorator, "auto_attribs",
+                                         auto_attribs)
+
+    if not init and not cmp:
+        # Nothing to add.
+        return
+
+    # To support nested classes we use fullname(). But fullname is <filename>.<variable name>
+    # and named_type() expects only the name.
```
ilevkivskyi: I don't understand this comment. Why do you need the self type here? It is normally bound at the call site (i.e. type variables of a generic type are fixed).

mypy/plugin.py
```diff
@@ -379,3 +392,189 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+            if isinstance(expr.node, Var):
+                return expr.node.is_classvar
+            return False
+
+        # Walk the mro in reverse looking for those yummy attributes.
+        for info in reversed(ctx.cls.info.mro):
+            for stmt in info.defn.defs.body:
+                if isinstance(stmt, AssignmentStmt) and isinstance(stmt.lvalues[0], NameExpr):
+                    lhs = stmt.lvalues[0]
+                    name = lhs.name.lstrip("_")
```
ilevkivskyi: Why this is necessary? Does ``attr`` has specific treatment of underscores? If yes, then please add a comment.

mypy/plugin.py
```diff
@@ -379,3 +392,189 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                    typ = stmt.type
+
+                    if called_function(stmt.rvalue) in attr_attrib_makers:
+                        assert isinstance(stmt.rvalue, CallExpr)
+
+                        # Is it an init=False argument?
+                        attr_init = get_argument(stmt.rvalue, "init", 5)
```
ilevkivskyi: Avoid any magic numbers, use a clearly named constant instead of 5.

mypy/plugin.py
```diff
@@ -379,3 +392,189 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                        if attr_init and ctx.api.parse_bool(attr_init) is False:
+                            continue
+
+                        # Look for default=  in the call.
+                        default = get_argument(stmt.rvalue, "default", 0)
```
ilevkivskyi: I think zero here should be also replaced by a constant.

mypy/plugin.py
```diff
@@ -379,3 +392,189 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                        attr_typ = get_argument(stmt.rvalue, "type", 15)
```
ilevkivskyi: Magic number again. Also 15 is quite large/strange, maybe you can add a comment with a link to docs where you get these numbers?

mypy/test/data.py
```diff
@@ -101,6 +101,13 @@ def parse_test_cases(
                     src_path = join(os.path.dirname(path), arg)
                     with open(src_path) as f:
                         files.append((join(base_path, 'typing.pyi'), f.read()))
+                elif p[i].id == 'add-module':
```
ilevkivskyi: This whole block is not necessary, just put the "pseudo stub" for ``attrs`` in ``mypy/test-data/unit/lib-stub``.

test-data/unit/fixtures/attr.pyi
```diff
@@ -0,0 +1,14 @@
+from typing import TypeVar, overload, Callable, Any
```
ilevkivskyi: Move this file to ``mypy/test-data/unit/lib-stub`` as I explained above.

test-data/unit/check-attr.test
```diff
@@ -0,0 +1,160 @@
...
+UnTyped('1', '2', '3')
+UnTyped('1', '2', '3', 4)
+UnTyped('1', '2', def_kwarg='5')
+UnTyped('1', '2', def_kwarg='5')
+UnTyped(normal=1, private=2, def_arg=3, def_kwarg=4)
+UnTyped()  # E: Too few arguments for "UnTyped"
+UnTyped(1, 2) == UnTyped(2, 3)
+UnTyped(1, 2) >= UnTyped(2, 3)
+[builtins fixtures/attr_builtins.pyi]
+[add-module fixtures/attr.pyi]
```
ilevkivskyi: You don't need these lines.

test-data/unit/check-attr.test
```diff
@@ -0,0 +1,160 @@
...
+    normal = attr.ib()  # E: Need type annotation for variable
+    _private = attr.ib()  # E: Need type annotation for variable
+    def_arg = attr.ib(18)  # E: Need type annotation for variable
+    _def_kwarg = attr.ib(validator=None, default=18)   # E: Need type annotation for variable
+
+    CLASS_VAR = 18
+[builtins fixtures/attr_builtins.pyi]
+[add-module fixtures/attr.pyi]
+
+[case testTypedAttrS]
```
ilevkivskyi: This and some other tests are too large. Try splitting it in half (like the previous test).

euresti:
Ok this supports quite a bit of features now.

Things that aren't working correctly:
a) subclasses that overwrite attributes from parents.
b) Setting a default using a decorator on a method. (i.e. @y.default)
c) The `@dataclass` decorator.  ;)

I think everything else is working fairly well.  I kind of wish everything wasn't so hard coded on the method names.  Many users of attrs (us included) create helper functions to add metadata and other stuff. I'll probably have to make the code a bit more adaptable to that.

test-data/unit/check-attr.test
```diff
@@ -0,0 +1,160 @@
...
+@attr.s
+class C:
+    y = attr.ib()
+    @attr.s
+    class D:
+        x: int = attr.ib()
+reveal_type(C)  # E: Revealed type is 'def (y: Any) -> __main__.C'
+reveal_type(C.D)  # E: Revealed type is 'def (x: builtins.int) -> __main__.C.D'
+[builtins fixtures/attr_builtins.pyi]
+[add-module fixtures/attr.pyi]
```
ilevkivskyi: Could you please add more tests? Check not only that definitions are processed correctly, but how they interact with various other concepts. For example, try to have good coverage at least for:
* type aliases (both inside and outside of a class)
* type variables
* decorated generic classes
* type inference
* forward references (both inside and outside of class)
* importing decorated classes
* instance and class methods in decorated types

ilevkivskyi:
> The ``@dataclass`` decorator. ;)

Support for data classes is a separate topic/PR. Also, they will probably be supported directly like named tuples (not via plugin).

euresti:
Actually attr has a `@dataclass` decorator, defined as 
```
dataclass = partial(attrs, auto_attribs=True)
```
That's why there was a smiley.  

Thanks for the review.  I'll probably play with things again after the break.

mypy/plugin.py
euresti: The named tuple code does the same thing.  arg_types is technically a `List[Optional[Type]]`  but note the `assert None not in arg_types` right above to make sure it's right.

mypy/plugin.py
euresti: Oh I see.  `AnyType(TypeOfAny.unannotated)` is what i need here!  Thanks!

euresti: Oh wait. I need it for 
`def __lt__(self, other: SelfType)`

mypy/plugin.py
euresti: This was made up. :)  Removing this useless code anyway.

mypy/test/data.py
euresti: Ok cool. I didn't know if that was ok since it's not a stdlib.

mypy/plugin.py
ilevkivskyi: I think the cast is not required with that assert. Have you actually tried removing it?

ilevkivskyi:
> Actually attr has a ``@dataclass`` decorator

Ah, OK, I didn't know.

mypy/plugin.py
ilevkivskyi: OK, then please use it only in methods that need it. Also please double-check that this reflects the actual runtime semantics/docs.

euresti: Actually, I found out that if i do `ctx.api.accept(func)` on the FuncDef I create it does a lot of things for me (setting the type of the self arg, being one of them).  Is that a reasonable thing to do?

ilevkivskyi: > Is that a reasonable thing to do?

I am not sure it is really needed, but if you find a case where this is necessary then go ahead.

test-data/unit/check-attr.test
euresti: The only tests I couldn't write because I wasn't sure what you meant were "type inference"  But I think I covered everything else.  Let me know if you want more tests.

mypy/plugin.py
euresti: Yeah I get:

```
mypy/plugin.py:580: error: Argument 1 to "CallableType" has incompatible type "List[Optional[Type]]"; expected "List[Type]"
```

When running `mypy mypy/plugin.py --strict`

euresti:
Ok I believe this is now feature complete.  It's pretty long because the attrs feature set is pretty large.  Let me know what other changes you want.

mypy/plugin.py
```diff
@@ -379,3 +407,289 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+)
+attrib_arguments = AttribMaker(
+    default=ArgumentInfo(None, 'default', 0),
+    init=ArgumentInfo(True, "init", 5),
+    type=ArgumentInfo(None, 'type', 8),
+)
+
+
+# The names of the different functions that create classes or arguments.
+attr_class_makers = {
```
euresti: I could use some advice/opinion here. It's fairly common when using attrs to write your own wrappers around the functions to make coding easier.  However the plugins only work on specific names so I tried to leave some way for users to add their own "class makers" and "attrib makers" to the plugin.  But that's probably not the right way to do this.

gvanrossum:
I think it would be good if @ambv looked at it. IIRC he is a big attrs user
(and co-author of PEP 484).

ilevkivskyi:
> I think it would be good if @ambv looked at it. IIRC he is a big attrs user (and co-author of PEP 484).

This is a good idea. I already mentioned that it would be great if someone familiar with ``attrs`` will look at this PR (in particular at the tests, to check that ``attrs`` "semantics" is right) I was also thinking about @hynek or @Tinche.

Anyway, I will also make another pass of review soon.

hynek:
This stuff is way over my head :) but one thing I’ve noticed: `convert` has been deprecated in favour of `converter` (to mirror `validator`).

ilevkivskyi:
> but one thing I’ve noticed: ``convert`` has been deprecated in favour of ``converter``

Thanks! This is exactly the kind of feedback I wanted to see. Also _later_ (when there will be more tests), you can comment if the error messages given by mypy are reasonable.

euresti:
I can add support for `converter`.  
And looking at 17.4.0 I'm also gonna have to change the MRO handling code.

euresti:
Ok I've added support for `converter` and fixed the MRO issue.

I'm a little unhappy because in order to support subclassing correctly I had to add something to the `TypeInfo` instance.  Basically I needed to add the list of Attributes somewhere so that a) I don't have to traverse the class bodies again for subclasses and b) I can modify the body as I traverse it e.g. to remove the `@x.default` decorators that don't make sense during static type checking.

Another way would be to add a `Dict[TypeInfo, Attributes]` to the `DefaultPlugin` and then all the attributes would be in that dictionary.  

I'm also starting to think that the code is pretty large to all live in one method and am wondering if I should put in a separate file and even maybe a separate plugin from `DefaultPlugin`.  I would like it to live in mypy though (rather than in attrs) because it'll be less likely to break.

gvanrossum:
(I'm hoping Ivan will be able to review this in the next few weeks and guide it to completion.)

euresti:
Ok I'm now mostly happy with the code.  I got rid of the magic numbers, the MRO works correctly, and the tests pass.  Feel free to review when you're ready.  Also this is my first big mypy project so I expect I did lots of things wrong.    

And a musing for v2:

The only other thing we might want is a way to allow users to say: this function returns an attrib or this function basically creates an attr class.  I was thinking it could be a decorator.  Something like:

```
@attr.register_attrib_func()
def type_checked(...)
    # Add my special validators.
    return attr.ib(...)

@attr.register_attrs_func(auto_attribs=True)
def myattr_maker(...)
    # Add my special validators.
    return attr.s(...)
```

At runtime the decorator does nothing, but in mypy we'd see the decorator and add the function to the list of functions we process.  Additionally the decorator could change the default values for the methods (e.g. auto_attribs) etc.

However since that requires commits in attrs, typeshed and mypy I'd figure that's best left to v2.  :)

euresti:
Sigh. Just noticed that my caching of `Attributes` in the plugin breaks incremental mode + subclassing.  I might have to move the caching back into `TypeInfo` so that I can serialize them in the mypy_cache.

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                callable_type = callee_type
+        return callable_type
+
+    def get_argument(call: CallExpr, name: str) -> Optional[Expression]:
+        """Return the expression for the specific argument."""
+        callee = get_callable_type(call)
+        if not callee:
+            return None
+
+        argument = callee.argument_by_name(name)
```
ilevkivskyi: I would try to inline this function instead of using ``get_callable_type`` (unless it is too long). I think the corresponding ``FuncDef`` should have all the information

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+        if not argument:
+            return None
+
+        if not argument.name and not argument.pos:
```
ilevkivskyi: This line isn't clear. Also I think we should report errors for incorrect calls to ``attr.s`` (with tests).

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+
+        for i, (attr_name, attr_value) in enumerate(zip(call.arg_names, call.args)):
+            if argument.pos is not None and not attr_name and i == argument.pos:
+                return attr_value
+            if attr_name == argument.name:
+                return attr_value
+        return None
+
+    def get_bool_argument(expr: Expression, name: str, default: bool) -> bool:
+        """Return the value of an argument name in the give Expression.
```
ilevkivskyi: Typo "give" -> "given".

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+
+        If it's a CallExpr and the argument is one of the args then return it.
+        Otherwise return the default value for the argument.
+        """
+        if isinstance(expr, CallExpr):
+            attr_value = get_argument(expr, name)
+            if attr_value:
+                ret = ctx.api.parse_bool(attr_value)
+                if ret is None:
+                    ctx.api.fail('"{}" argument must be True or False.'.format(name), expr)
```
ilevkivskyi: Add a test for this errors.

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                    return default
+                return ret
+        return default
```
ilevkivskyi: Add a comment that a decorator can be only ``CallExpr`` or ``RefExpr``, therefore we can get here with a plain ``attr.s``. Also I don't it is safe to apply the same logic to ``attr.ib`` as you do below for ``init``.

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+
+    # auto_attribs means we also generate Attributes from annotated variables.
+    auto_attribs = get_bool_argument(decorator, 'auto_attribs', auto_attribs_default)
+
+    def get_attributes_for_body(body: List[Statement]) -> List[Attribute]:
+        attributes = OrderedDict()  # type: OrderedDict[str, Attribute]
+
+        for stmt in body:
+            if isinstance(stmt, AssignmentStmt) and isinstance(stmt.lvalues[0], NameExpr):
+                lhs = stmt.lvalues[0]
```
ilevkivskyi: Shouldn't we require ``len(lvalues) == 1``? (Add a test case where this fails.)

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                assert isinstance(lhs, NameExpr)
```
ilevkivskyi: If you move definition of ``lhs`` above ``isinstance``, then I think you will not need this assert.

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                name = lhs.name
+                typ = stmt.type
+
+                func_name = called_function(stmt.rvalue)
```
ilevkivskyi: ``called_function`` is used only once here, doesn't it make sense to just inline it?

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                            and isinstance(converter.node.type, CallableType)
+                            and converter.node.type.arg_types):
+                        typ = converter.node.type.arg_types[0]
+
+                    # Does this even have to go in init?
+                    init = get_bool_argument(stmt.rvalue, 'init', True)
+
+                    if name in attributes:
+                        del attributes[name]
+                    attributes[name] = Attribute(name, typ, attr_has_default, init, stmt)
```
ilevkivskyi: Add a comment that we first delete and them set to get right order in ``OrderedDict`` (this is my guess).

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                elif auto_attribs and typ and stmt.new_syntax and not is_class_var(lhs):
+                    # `x: int` (without equal sign) assigns rvalue to TempNode(AnyType())
+                    has_rhs = not isinstance(stmt.rvalue, TempNode)
+                    if name in attributes:
+                        del attributes[name]
+                    attributes[name] = Attribute(name, typ, has_rhs, True, stmt)
+
+            elif isinstance(stmt, Decorator):
+                # Look for attr specific decorators.  ('x.default' and 'x.validator')
```
ilevkivskyi: You don't have a test case for ``x.validator``

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+
+                        if func_decorator.name == 'default':
+                            # This decorator lets you set a default after the fact.
+                            attributes[func_decorator.expr.name].has_default = True
+
+                        if func_decorator.name in ('default', 'validator'):
+                            # These are decorators on the attrib object that only exist during
+                            # class creation time.  In order to not trigger a type error later we
+                            # just remove them.  This might leave us with a Decorator with no
+                            # decorators (Emperor's new clothes?)
```
ilevkivskyi: I think this is fragile. Some code can crash because of this. A proper solution would be to replace it with a ``FuncDef``.

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+    if get_bool_argument(decorator, 'init', True):
+        # Generate the __init__ method.
+
+        # Check the init args for correct default-ness.  Note: This has to be done after all the
+        # attributes for all classes have been read, because subclasses can override parents.
+        last_default = False
+        for attribute in attributes:
+            if not attribute.has_default and last_default:
+                ctx.api.fail(
+                    "Non-default attributes not allowed after default attributes.",
```
ilevkivskyi: I didn't find a test for this error.

test-data/unit/check-attr.test
```diff
@@ -0,0 +1,444 @@
...
+import attr
+T = TypeVar('T')
+@attr.s(auto_attribs=True)
+class A(Generic[T]):
+    x: List[T]
+    y: T = attr.ib()
+reveal_type(A) # E: Revealed type is 'def [T] (x: builtins.list[T`1], y: T`1) -> __main__.A[T`1]'
+
+a = A([1], 2)
+reveal_type(a)  # E: Revealed type is '__main__.A[builtins.int*]'
```
ilevkivskyi: Please also reveal types of attributes.

test-data/unit/check-attr.test
```diff
@@ -0,0 +1,444 @@
...
+    y: Alias2 = attr.ib()
+reveal_type(A) # E: Revealed type is 'def (x: builtins.list[builtins.int], y: builtins.list[builtins.str]) -> __main__.A'
+[builtins fixtures/list.pyi]
+
+[case testAttrsTypeVariable]
+from typing import TypeVar, Generic, List
+import attr
+T = TypeVar('T')
+@attr.s(auto_attribs=True)
+class A(Generic[T]):
```
ilevkivskyi: Also please add a test where generic class has a method, and it contains access to an attribute on ``self``.

test-data/unit/check-attr.test
```diff
@@ -0,0 +1,444 @@
+[case testAttrsSimple]
+# TODO(David):
+# Add tests for:
+# o type inference
```
ilevkivskyi: Tests for inference should verify that a generic attrs class has correctly inferred type variables after a call. I think you already have one test, just add one more where it would lead to an error (like conflicting attributes to ``__init__``).

test-data/unit/check-attr.test
```diff
@@ -0,0 +1,444 @@
...
+
+
+@attr.s
+class A(C):
+    z: int = attr.ib(default=18)
+
+reveal_type(C)  # E: Revealed type is 'def (x: builtins.int =, y: builtins.int =) -> __main__.C'
+reveal_type(A)  # E: Revealed type is 'def (x: builtins.int =, y: builtins.int =, z: builtins.int =) -> __main__.A'
+
+[builtins fixtures/list.pyi]
```
ilevkivskyi: A general comment about tests: you are mostly testing situations when everything goes right (i.e. user code is correct). Try to add more tests where a user makes various errors, to be sure that we give a clear error in a correct location.

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                and isinstance(call.callee, RefExpr)
+                and isinstance(call.callee.node, Var)
+                and call.callee.node.type):
+            callee_type = call.callee.node.type
+            if isinstance(callee_type, Overloaded):
+                # We take the last overload.
+                callable_type = callee_type.items()[-1]
+            elif isinstance(callee_type, CallableType):
+                callable_type = callee_type
+        return callable_type
```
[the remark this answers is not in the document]
euresti: I was trying to programmatically extract the signature so that in the future I could support helper functions that return attribs.  (For example at my company we add a couple of arguments to do some special handling)  By looking at the stub rather than hardcoding the order and names, then this could be supported.

mypy/plugin.py
```diff
@@ -379,3 +421,332 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+        self.has_default = has_default
+        self.init = init
+        self.context = context
+
+    def argument(self) -> Argument:
+        """Return this attribute as an argument to __init__."""
+        # Convert type not set to Any.
+        _type = self.type or AnyType(TypeOfAny.unannotated)
+        # Attrs removes leading underscores when creating the __init__ arguments.
+        return Argument(Var(self.name.strip("_"), _type), _type,
```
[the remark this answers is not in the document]
euresti: Good catch. This should've been lstrip.

mypy/plugin.py
euresti: Which `FuncDef` are you referring to?

mypy/plugin.py
euresti: Oh wow. You can actually declare attrs in multiple ways (which this doesn't currently implement)

```
@attr.s
class A:
    x = attr.ib()
    z, y = attr.ib(), attr.ib()
    a = b = c = attr.ib()
```

@hynek Is the 3rd form I write a correct way to do it. Can 2 names use the same attr?

mypy/plugin.py
euresti: I've seen this done in the code base though:  https://github.com/python/mypy/blob/master/mypy/semanal.py#L2921

test-data/unit/check-attr.test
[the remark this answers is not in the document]
euresti: Sorry about the rebase. I noticed I was using the wrong email address so I rebased to get the correct name. Won't happen again.

mypy/plugin.py
ilevkivskyi: The one `callee` normally "points to". I think you see some `Var`s because you use function aliases, but even in that case you should be able to get to the original `FuncDef`.

euresti: So 2 problems with this approach.  

1. In the non-`Var` case, `call.callee.node` points to an `OverloadedFuncDef`.  So I still need to pick one of the overloads to get a `FuncDef` is that an ok thing to do?

2. For the life of me the only way I can see to get a `FuncDef` from a `Var` is to use `lookup_qualified` to get the `SymbolTableNode` get the `CallableType` out of that and then get the `FuncDef` from `definition` Is that what you had in mind?  I can also instead get rid of the assignments in the stubs and just repeat the function definitions for all of them.  (That's the way the real stubs are written anyway because of a bug in pycharm) But I'd rather not depend on that.

Another option is to forget the whole looking at the stub file and just hard code the positions.  This will mean this can't be used as is for future extensions but why code for something that hasn't been developed.

Let me know what you think.

mypy/plugin.py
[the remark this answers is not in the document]
ilevkivskyi: Yes, I see. There is a trade-off between code clarity and flexibility. I will make another pass of review now to see how it reads.

mypy/plugin.py
ilevkivskyi: My intuition here is that if there will be a major change in `attrs`, then this plugin will anyway needs to be updated. So hard-coding some function signatures is OK. Also IIUC this is the logic other smaller plugins use (like `@contextlib.contextmanager`).

But I don't want to push you on this, if you believe it is better to extract signatures from the stubs then it is also OK, if the code readability doesn't suffer too much.

mypy/plugin.py
ilevkivskyi: OK, let us hope this will not break other code.

mypy/plugin.py
ilevkivskyi: This is still not fixed. Add a comment here. Do you have a test for this?

mypy/plugin.py
ilevkivskyi: This is still not fixed.

mypy/plugin.py
```diff
@@ -379,3 +421,346 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                    if func_decorator.name == 'default':
+                        # This decorator lets you set a default after the fact.
+                        own_attrs[func_decorator.expr.name].has_default = True
+
+                    if func_decorator.name in ('default', 'validator'):
+                        # These are decorators on the attrib object that only exist during
+                        # class creation time.  In order to not trigger a type error later we
+                        # just remove them.  This might leave us with a Decorator with no
+                        # decorators (Emperor's new clothes?)
+                        # TODO: Any way to type-check these?
```
ilevkivskyi: If you want to type-check them, then you should not remove them at this stage., since type checking happens at a later stage.

ilevkivskyi: I think this may be actually a question of having good `attrs` stubs.

mypy/plugin.py
```diff
@@ -379,3 +421,346 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                attribute.context)
+        last_default = attribute.has_default
+
+    self_type = fill_typevars(info)
+    function_type = ctx.api.named_type('__builtins__.function')
+
+    def add_method(method_name: str, args: List[Argument], ret_type: Type,
+                   self_type: Type = self_type,
+                   tvd: Optional[TypeVarDef] = None) -> None:
+        """Create a method: def <method_name>(self, <args>) -> <ret_type>): ..."""
```
ilevkivskyi: Explain what is the use of `tvd` in the docstring, and why it is only one.

mypy/plugin.py
```diff
@@ -379,3 +421,346 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+        if not attribute.has_default and last_default:
+            ctx.api.fail(
+                "Non-default attributes not allowed after default attributes.",
+                attribute.context)
+        last_default = attribute.has_default
+
+    self_type = fill_typevars(info)
+    function_type = ctx.api.named_type('__builtins__.function')
+
+    def add_method(method_name: str, args: List[Argument], ret_type: Type,
```
ilevkivskyi: Can you move this function to the outer scope? It looks like it is useful on its own (i.e., can be used by other plugins in future). Also please add a TODO to re-use the same code for `NamedTuple`.

mypy/plugin.py
```diff
@@ -379,3 +421,346 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+            attr_value = get_argument(expr, name)
+            if attr_value:
+                ret = ctx.api.parse_bool(attr_value)
+                if ret is None:
+                    ctx.api.fail('"{}" argument must be True or False.'.format(name), expr)
+                    return default
+                return ret
+        return default
+
+    def is_class_var(expr: NameExpr) -> bool:
```
ilevkivskyi: Move this function to the outer scope (or maybe actually move it to `nodes.py` and import it from there). It can be useful on its own.

mypy/plugin.py
```diff
@@ -379,3 +421,346 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+            return None
+
+        for i, (attr_name, attr_value) in enumerate(zip(call.arg_names, call.args)):
+            if argument.pos is not None and not attr_name and i == argument.pos:
+                return attr_value
+            if attr_name == argument.name:
+                return attr_value
+        return None
+
+    def get_bool_argument(expr: Expression, name: str, default: bool) -> bool:
```
ilevkivskyi: Move this function to the outer scope and prefix it with `_attrs_` to indicate it is a private helper for `attrs` plugin.

mypy/plugin.py
```diff
@@ -379,3 +421,346 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                and call.callee.node.type):
+            callee_type = call.callee.node.type
+            if isinstance(callee_type, Overloaded):
+                # We take the last overload.
+                callable_type = callee_type.items()[-1]
+            elif isinstance(callee_type, CallableType):
+                callable_type = callee_type
+        return callable_type
+
+    def get_argument(call: CallExpr, name: str) -> Optional[Expression]:
```
ilevkivskyi: Move this function to the outer scope and prefix it with `_attrs_` to indicate it is a private helper for `attrs` plugin.

mypy/plugin.py
```diff
@@ -379,3 +421,346 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+
+    At a quick glance, the decorator searches the class body for assignments of `attr.ib`s (or
+    annotated variables if auto_attribs=True), then depending on how the decorator is called,
+    it will add an __init__ or all the __cmp__ methods.  For frozen=True it will turn the attrs
+    into properties.
+
+    See http://www.attrs.org/en/stable/how-does-it-work.html for information on how attrs works.
+    """
+
+    def get_callable_type(call: CallExpr) -> Optional[CallableType]:
```
ilevkivskyi: Move this function to the outer scope and prefix it with `_attrs_` to indicate it is a private helper for `attrs` plugin. Note that there is an agreement in mypy to place helper functions _after_ the  main function.

mypy/plugin.py
euresti: The actual problem here is that the code looks like this:

```
@attr.s
class A:
   x: int = attr.ib()

   @x.default
   def foo(self):
      return ...
```

So as far as mypy is concerned that `x` is an `int`.  So I definitely have to remove the `x.default` decorator.  I imagine I could replace it with a different decorator but I'm not even sure what that would look like.  I'll update the comment so that it's not a question.

mypy/plugin.py
euresti: Fixed now.  It was an impossible situation because `argument_by_name` always returns a `.name`

euresti:
Thanks for the review!  I believe I got to all your points.

The only thing I'm worried about now is that this plugin doesn't do the right thing in incremental mode when you subclass and the parent class is cached but the subclass isn't.

The only solution I've found is to add a `List[Attributes]` to `TypeInfo` so that it can be serialized/unserialized.  Then you have to figure out how to "fixup" the type to make it all work.  I don't know if there's another way.

test-data/unit/lib-stub/attr.pyi
```diff
@@ -0,0 +1,57 @@
...
+               init: bool = ...,
+               slots: bool = ...,
+               frozen: bool = ...,
+               str: bool = ...,
+               auto_attribs: bool = ...) -> Callable[[_C], _C]: ...
+
+# aliases
+s = attrs = attributes
+ib = attrib = attr
+dataclass = attrs # Technically, partial(attrs, auto_attribs=True) ;)
```
hynek: Oh man, you found my Easter egg. ;)

test-data/unit/lib-stub/attr.pyi
```diff
@@ -0,0 +1,57 @@
...
+_C = TypeVar('_C', bound=type)
+
+@overload
+def attr(default: Optional[_T] = ...,
+         validator: Optional[Any] = ...,
+         repr: bool = ...,
+         cmp: bool = ...,
+         hash: Optional[bool] = ...,
+         init: bool = ...,
+         convert: Optional[Callable[[Any], _T]] = ...,
```
hynek: I know other people have different opinions but I find it unfortunate that you’re adding a deprecated option to mypy. The sooner people get driven off using it, the better. Since nobody™ runs their tests with warnings activated, this might be a unique chance.

test-data/unit/lib-stub/attr.pyi
```diff
@@ -0,0 +1,57 @@
...
+         type: Optional[Type[_T]] = ...,
+         converter: Optional[Callable[[Any], _T]] = ...) -> _T: ...
+@overload
+def attr(default: None = ...,
+         validator: None = ...,
+         repr: bool = ...,
+         cmp: bool = ...,
+         hash: Optional[bool] = ...,
+         init: bool = ...,
+         convert: Optional[Callable[[Any], _T]] = ...,
```
hynek: same

test-data/unit/lib-stub/attr.pyi
euresti: Ok so how about I emit and error from the plugin `Using deprecated option "convert"` or something like that.  But I think we should keep the handling of it because of `--follow-imports=silent` 

Not handling `convert` means that `__init__` may not have the correct type.  

> silent follow imports normally and even “type check” them normally, but suppress any error messages. This is typically the best option for a new codebase.

euresti:
@hynek Thanks for your comments.  If you wouldn't mind looking over test-data/unit/check-attr.test and seeing that all the cases are being handled and that the error messages make sense.

All the `#E: ...` are the errors that will be printed out by mypy.
And `reveal_type(x)` will tell mypy to show you the type of `x` by outputting an error.

hynek:
OK I had a cup of coffee and worked myself through it. That’s a lot of code and I already feel guilty for moving to enums for repr/cmp/hash/… in medium term.

But as far as I can see: great work! There’s absolutely no way, I’d have done this myself.

***

I heard @ambv would like to have a look too. :)

test-data/unit/lib-stub/attr.pyi
euresti: This has been added.

test-data/unit/lib-stub/attr.pyi
euresti: I found a lot of things!  ;)

ambv:
> I heard @ambv would like to have a look too. :)

True, sorry, this is big. Looking now.

ambv:
Love the idea, implementation looks sensible. I have three things.

1. In terms of design, I'd like to see the attrs-related part of the plugin out of `mypy/plugin.py` and in its own file. I don't think we will want to keep implementations of everything in that file when time comes for Click, Django models, etc.

2. I would highly recommend splitting the main `attr_class_maker_callback()` into a number of smaller functions. It is the second most complex function in the codebase in terms of paths (47!). Close to 25% of this pull request is in the body of this function.

3. This doesn't work for me yet. Run [this Gist](https://gist.github.com/ambv/f22d7aef45633e2365db47966e293bc1) to create an example repro. The reason it doesn't work is that fullname reported to `get_class_decorator_hook()` is `pkg.subpkg.dep.attr` so it doesn't match any of the listed "attr.s" variants.

Also, curiously, when mypy is reporting the false positives here, it points at builtins.pyi for "definition of C":
```
pkg/subpkg/dep.py:1: error: Cannot find module named 'attr'
pkg/subpkg/dep.py:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
pkg/subpkg/dep.py:13: error: Unexpected keyword argument "field1" for "C"
/Users/ambv/Dropbox/Python/attrsmypy/mypy/typeshed/stdlib/3/builtins.pyi:39: note: "C" defined here
```

euresti:
As to 1.  Would you prefer a 2nd plugin or just the code moved out?  Not hard to move the code out.  

As to 2. Yep. Perfectly doable.  

As to 3. I believe it doesn't work because the typeshed doesn't have the annotations and so mypy is reporting the name as where it found it.  (But I have to investigate that in case the code changed at some point)  https://github.com/python/typeshed/pull/1838

euresti:
Ok just used your gist but pulled in the typeshed from that PR and I get no errors.

ambv:
1. Just move it out. That will demonstrate that other future plugins are not meant to be put in mypy/plugin.py either.

2. Great.

3. You are right! I just used your fork of mypy on this branch which was a mistake. After using `--custom-typeshed-dir` with a checkout of https://github.com/chadrik/typeshed/tree/attrs the plugin works, indeed. Sorry for the confusion.

4. Getting the plugin to work with the proper typeshed, I noticed that when `attr.s()` is used as a class decorator, mypy is not reporting invalid keyword arguments passed to it. When used as a function call, it does. The error message is very generic but at least it tells you that something is wrong.

gvanrossum:
Maybe we should try to land this before we release mypy 0.570.

euresti:
When is that?  I can find some time this weekend to refactor some of the code as ambv wants.

But as a warning Incremental mode with parent classes in other files is still broken. I don't entirely know how to fix that.

euresti:
@ambv can you post the code that exhibits this problem just so I understand exactly.

chadrik:
Regarding tests, I've added [pretty thorough tests](https://github.com/chadrik/attrs/blob/pyi_stubs/tests/test_stubs.tests.py) to my [attrs PR for pyi stubs](https://github.com/python-attrs/attrs/pull/238) which cover both the attrs stubs and mypy plugin.   I'm not sure where this should ultimately live:  mypy, typeshed, attrs, or some combination thereof?

ilevkivskyi:
@euresti There is no specific timeline IIUC, but I think we could get this in (including typeshed PR) in next few days. This is almost ready.

As for the incremental mode, maybe we can just allow plugins to add keys to a special `metadata` field of `TypeInfo`. Probably you can add it in already serialized form (like JSON), so it will be damped to cache as is, and then loaded as is. My understanding is that you just need to preserve flags for certain `Var`s in the class symbol table, their types are already (de-)serialized automatically. For example, if you have a class like this:
```python
@attr.s(frozen=True)
class C:
    x: int = attr.ib(init=False)
    y: str
```
Then you can just write (pseudocode, hopefully you get the idea):
```python
type_info.metadata['attrs'] = {'fileds': {'x': [], 'y': ['init']}, 'flags': ['frozen']}
```
Taking into account the flags (both global and per-field) and the types that you can find in class symbol table, you should be able to reconstruct all the necessary info from a deserialized base class.

euresti:
@ambv I think I know your issue. It's a mypy issue: https://github.com/python/mypy/issues/3135

ilevkivskyi:
@euresti 
> I think I know your issue. It's a mypy issue

Since decorators are important for `attrs`, maybe you can write a limited fix (in a separate PR)? Namely, just check the decorators that are function calls. They should be visited in semantic analysis and type check, you can see how it is done for a regular `CallExpr`. This would be a partial fix, but it will already catch the errors like @ambv discovered.

euresti:
Definitely on my mind.  Wanna get the serialization working first though.

euresti:
Dang that was hard.  But it works!  Comments welcome!

ambv:
Cool! This looks much better. I won't block you on this basis alone but we might want to split `attr_class_maker_callback` further in a future refactor. [Paraphrasing Linus Torvalds](https://www.kernel.org/doc/html/v4.10/process/coding-style.html#indentation), "if you need more than 3 levels of indentation, you’re screwed anyway, and should fix your program". Another rule of thumb that applies here: if you need comments to name separate sections in your function, those should really be separate functions :-)

As I said though, I won't bother you with doing this right now. Just something to keep in mind for future work.

euresti:
Ok I refactored a bit more and cleaned up some stuff related to the converter method.  Will leave this alone for a bit until I hear something.  Thanks!

euresti:
@ilevkivskyi @ambv Any more comments on this PR?  I've already started using it on our code base and it's been pretty great.

test-data/unit/check-incremental.test
```diff
@@ -3467,6 +3467,68 @@ tmp/main.py:2: error: Expression has type "Any"
...
+[file a.py]
+import attr
+@attr.s
+class A:
+    x: int = attr.ib()
+
+[out1]
+main:6: error: Revealed type is 'def (x: builtins.int) -> __main__.B'
+[out2]
+main:6: error: Revealed type is 'def (x: builtins.int) -> __main__.B'
```
ilevkivskyi: Please add more tests for incremental mode. Here are some ideas:
* Check situations with more attributes
* Checks that access to attributes after deserialization results in a correct type and "kind" (i.e. being frozen etc.)
* Check that signatures of deserialized dunder methods are correct.
* Check situations where base attr class in module `a.py` is not changed in the second run, but its subclass in module `b.py` is changed so that this produces a type error.
* Also vice-versa (error to no error)
* Add test with more than two files
* Add test with more than two incremental runs (for example: no error -> error -> no error in a subclass)

Here are two additional notes about incremental tests:
* If a file results in errors, it is never actually deserialized from cache, this _includes_ `reveal_type`. So although you can keep current tests (to check serialization, which happens always), please avoid `reveal_type` in future incremental tests, instead use an assignment to check that the type is as expected.
* To add changes between runs, use this pattern:
```
[file a.py]
import b
...
[file a.py.2]
import b
...
[file b.py]
...
```
Important note: you don't need to add `.2`, (and `.3` in one test) to all files, only to the files that are actually changed. It is better to actually _not_ do this (to avoid `touch`ing unchanged file, and therefore considering its cache stale).

Don't add to many tests, just one case for each item I listed above.

euresti:
Thanks for the review. I've added a couple more tests.  I hope I got all your points.

gvanrossum:
+1 on merging!

Issue #2305, opened 2016-10-24
Support iteration on enums


bbc2:
Iterating on enums makes mypy (0.4.5) raise errors even though it's OK:

``` python
import enum


class Colors(enum.Enum):
    RED = 0
    BLUE = 1


print([color.name for color in Colors])
```

```
> python3 color.py 
['RED', 'BLUE']
```

```
> mypy color.py 
color.py:9: error: Iterable expected
color.py:9: error: "Colors" has no attribute "__iter__"
```

(follow-up issue created as suggested in #529)

elazarg:
At first sight, we only need #2193 
```Python
class Enum:
    ...
    @classmethod
    def __iter__(cls: Type[T]) -> Iterator[T]: ...
```
And somehow making Enum itself iterable?

But it's not so simple, since `Enum` is a class appearing in the context of an expression, and therefore "coerce" to a callable somewhere in the code. Thus the above code still results in `error: Iterable expected`. At [the point where the error is emitted](https://github.com/python/mypy/blob/master/mypy/checker.py#L1784), the information that it's enum does not seem to exist anymore - it is a simple callable.

dokai:
In terms of making a class itself iterable it seems you have to use a metaclass instead of using just a `@classmethod` in the class definition.

```python
In [7]: class A:
   ...:     @classmethod
   ...:     def __iter__(cls):
   ...:         return iter(range(3))
   ...:

In [8]: list(A)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-ed10f13c6f39> in <module>()
----> 1 list(A)

TypeError: 'type' object is not iterable

In [9]: class MetaIter(type):
   ...:     def __iter__(self):
   ...:         return iter(range(3))
   ...:

In [10]: class B(metaclass=MetaIter): pass

In [11]: list(B)
Out[11]: [0, 1, 2]
```

gvanrossum:
Yes, that's what I meant. But that's runtime. How to do it in mypy is
probably quite different because it doesn't interpret the metaclass.

gvanrossum:
Did you check the is_type_obj() method? It should still know...

Also, I doubt that defining a classmethod is the same as having it on the
metaclass. IIRC that doesn't even work in CPython.

elazarg:
I didn't check yet.

As you said, it's not how mypy work. I think it should, but I guess it will require a big change.

[...]

PR #4400, opened 2017-12-21
Add enum iteration test cases

ngaya-ll:
Test cases for #2305.

### DIFF
--- a/test-data/unit/pythoneval.test
+++ b/test-data/unit/pythoneval.test
@@ -1459,11 +1459,8 @@ _testNoCrashOnGenericUnionUnpacking.py:11: error: Revealed type is 'Union[builti
 _testNoCrashOnGenericUnionUnpacking.py:15: error: Revealed type is 'Union[builtins.int*, builtins.str*]'
 _testNoCrashOnGenericUnionUnpacking.py:16: error: Revealed type is 'Union[builtins.int*, builtins.str*]'
 
-[case testEnumIteration]
+[case testEnumIterationAndPreciseElementType]
 # Regression test for #2305
-# Tests that:
-# - Mypy allows iteration over enum classes in comprehensions and for-loops
-# - The inferred element type in such iteration contexts is the enum type
 from enum import Enum
 class E(Enum):
     A = 'a'
@@ -1471,8 +1468,8 @@ class E(Enum):
 for e in E:
     reveal_type(e)
 [out]
-_testEnumIteration.py:8: error: Revealed type is '_testEnumIteration.E*'
-_testEnumIteration.py:10: error: Revealed type is '_testEnumIteration.E*'
+_testEnumIterationAndPreciseElementType.py:5: error: Revealed type is '_testEnumIterationAndPreciseElementType.E*'
+_testEnumIterationAndPreciseElementType.py:7: error: Revealed type is '_testEnumIterationAndPreciseElementType.E*'
 
 [case testEnumIterable]
 from enum import Enum
--- a/test-data/unit/pythoneval.test
+++ b/test-data/unit/pythoneval.test
@@ -1460,6 +1460,10 @@ _testNoCrashOnGenericUnionUnpacking.py:15: error: Revealed type is 'Union[builti
 _testNoCrashOnGenericUnionUnpacking.py:16: error: Revealed type is 'Union[builtins.int*, builtins.str*]'
 
 [case testEnumIteration]
+# Regression test for #2305
+# Tests that:
+# - Mypy allows iteration over enum classes in comprehensions and for-loops
+# - The inferred element type in such iteration contexts is the enum type
 from enum import Enum
 class E(Enum):
     A = 'a'
@@ -1467,8 +1471,8 @@ class E(Enum):
 for e in E:
     reveal_type(e)
 [out]
-_testEnumIteration.py:4: error: Revealed type is '_testEnumIteration.E*'
-_testEnumIteration.py:6: error: Revealed type is '_testEnumIteration.E*'
+_testEnumIteration.py:8: error: Revealed type is '_testEnumIteration.E*'
+_testEnumIteration.py:10: error: Revealed type is '_testEnumIteration.E*'
 
 [case testEnumIterable]
 from enum import Enum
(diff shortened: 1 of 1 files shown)


elazarg:
`Enum` should follow typeshed - `__iter__` should be added to an EnumMeta metaclass.

ngaya-ll:
@elazarg I added an iteration test that directly iterates over an enum type without checking `Iterable`, but for some reason that also fails even though equivalent code passes with mypy 0.560.  Do you know what's going on there?

elazarg:
@ngaya-ll see previous comment - the enum stub is incomplete, it does not have an `EnumMeta` metaclass with an `__iter__` method (which typeshed already has).

PR 1755 is about the removal of explicit baseclasses, which will fix the other tests; you can try mimicking it in the stub.

ngaya-ll:
@elazarg If I understand correctly, you're saying the unit tests here are independent of the typeshed definitions.

I observed a problem with mypy in the wild, so I thought it would make sense to add test cases here to demonstrate the problem and verify when it was fixed.  But it sounds like these tests use dummy type definitions that will not reflect the real-world behavior of mypy.

Is there a better way to test what I'm trying to test (i.e. that mypy + typeshed correctly handles enums in real life)?  Or is this not helpful?

elazarg:
This is what [python-eval](https://github.com/python/mypy/blob/master/test-data/unit/pythoneval.test) tests are for.

ngaya-ll:
@elazarg Thanks!  Moved the tests to pythoneval and they are now passing.  Can this PR be merged?

test-data/unit/pythoneval.test
```diff
@@ -1458,3 +1458,48 @@ _testNoCrashOnGenericUnionUnpacking.py:10: error: Revealed type is 'Union[builti
...
+[out]
+_testEnumIteration.py:5: error: Revealed type is '_testEnumIteration.E*'
+_testEnumIteration.py:7: error: Revealed type is '_testEnumIteration.E*'
+
+[case testEnumIterable]
+from enum import Enum
+from typing import Iterable
+class E(Enum):
+    a = 'a'
+def f(ie:Iterable[E]):
```
ilevkivskyi: Please follow PEP 8 (even in tests).

test-data/unit/pythoneval.test
```diff
@@ -1458,3 +1458,48 @@ _testNoCrashOnGenericUnionUnpacking.py:10: error: Revealed type is 'Union[builti
 _testNoCrashOnGenericUnionUnpacking.py:11: error: Revealed type is 'Union[builtins.str, builtins.int]'
 _testNoCrashOnGenericUnionUnpacking.py:15: error: Revealed type is 'Union[builtins.int*, builtins.str*]'
 _testNoCrashOnGenericUnionUnpacking.py:16: error: Revealed type is 'Union[builtins.int*, builtins.str*]'
+
+[case testEnumIteration]
```
ilevkivskyi: I would use longer/more detailed names for test cases (there are lots of tests in `pythoneval.test`).

test-data/unit/pythoneval.test
```diff
@@ -1458,3 +1458,48 @@ _testNoCrashOnGenericUnionUnpacking.py:10: error: Revealed type is 'Union[builti
...
+from enum import Enum
+class E(Enum):
+    A = 'a'
```
ilevkivskyi: Why do you use uppercase `A` here and lowercase `a` in other places?

test-data/unit/pythoneval.test
```diff
@@ -1458,3 +1458,48 @@ _testNoCrashOnGenericUnionUnpacking.py:10: error: Revealed type is 'Union[builti
...
+l = [e for e in E]
```
ilevkivskyi: Also reveal type of `list(E)`?

ilevkivskyi:
@ngaya-ll 
> Can this PR be merged?

Thank you for PR! Please see some suggestions I left above. After you consider them, we can add these tests.

test-data/unit/pythoneval.test
ngaya-ll: Hm, not sure what more detail would be helpful here.

test-data/unit/pythoneval.test
ngaya-ll: The test is more concerned with the element type of the comprehension, not the resulting list.  I'll edit the test to make that more explicit.

test-data/unit/pythoneval.test
ilevkivskyi: Something like `EnumIterationPreservesPreciseType`

test-data/unit/pythoneval.test
ilevkivskyi: > The test is more concerned with the element type of the comprehension, not the resulting list.

It is (almost) never bad to test more.

test-data/unit/pythoneval.test
ngaya-ll: That's only part of what this is aiming to test.  I'll add a comment in the test with more details.

test-data/unit/pythoneval.test
ngaya-ll: I was looking at it more in terms of trying to check the enum behavior with minimal reliance on other type-checking features.  But I can add the additional check if you feel strongly that it would be valuable.

test-data/unit/pythoneval.test
```diff
@@ -1458,3 +1458,51 @@ _testNoCrashOnGenericUnionUnpacking.py:10: error: Revealed type is 'Union[builti
 _testNoCrashOnGenericUnionUnpacking.py:11: error: Revealed type is 'Union[builtins.str, builtins.int]'
 _testNoCrashOnGenericUnionUnpacking.py:15: error: Revealed type is 'Union[builtins.int*, builtins.str*]'
 _testNoCrashOnGenericUnionUnpacking.py:16: error: Revealed type is 'Union[builtins.int*, builtins.str*]'
+
+[case testEnumIteration]
```
ilevkivskyi: > That's only part of what this is aiming to test.

Then choose a better name. Also the comment below is too long, just leave `# Regression test for #2305`

I am not arguing for fun here, there is `pytest -k` command that will run all test cases with a given pattern in names. This is why it is important to add some specific keyword like `Precise` or `Preserve` (or your choice). In addition, this avoids accidental test name collisions (we currently have thousands of test cases).

PR #4401, opened 2017-12-21
Add dmypy.json to .gitignore

msullivan:


msullivan:
The normal case for using the daemon will be that it will be running all the time, though, including while doing git operations. So I think it is worth cleaning that up.

Issue #4406, opened 2017-12-24
Stubgen generates "<ERROR>" if namedtuple's fields are declared in a tuple


hoefling:
Contents of `spam.py`:

```
from collections import namedtuple

FieldsStr = namedtuple('FieldsStr', 'spam, eggs')
FieldsList = namedtuple('FieldsList', ['spam', 'eggs'])
FieldsTuple = namedtuple('FieldsTuple', ('spam', 'eggs'))
```

`$ stubgen -o . spam` yields `spam.pyi`:

```
# Stubs for spam (Python 3.6)
#
# NOTE: This dynamically typed stub was automatically generated by stubgen.

from collections import namedtuple

FieldsStr = namedtuple('FieldsStr', 'spam, eggs')

FieldsList = namedtuple('FieldsList', ['spam', 'eggs'])

FieldsTuple = namedtuple('FieldsTuple', <ERROR>)
```

Is this deliberate?

gvanrossum:
Looks like a bug. Can you submit a PR?

hoefling:
Looks like it's just a case that was forgotten to be handled. In `mypy/stubgen.py`, only field names in a string or a list are being processed:

```python
def process_namedtuple(self, lvalue: NameExpr, rvalue: CallExpr) -> None:
    ...
    if isinstance(rvalue.args[1], StrExpr):
        items = repr(rvalue.args[1].value)
    elif isinstance(rvalue.args[1], ListExpr):
        list_items = cast(List[StrExpr], rvalue.args[1].items)
        items = '[%s]' % ', '.join(repr(item.value) for item in list_items)
    else:
        items = '<ERROR>'
```

Seems to be an easy fix even I could implement.

P.S. omg, I just noticed I got a reply from Guido himself...
<img src="https://user-images.githubusercontent.com/4455652/34328394-ab487f72-e8de-11e7-8f23-dc812edd79a1.png" alt="alt text" width="400">

PR #4407, opened 2017-12-24
Handling of field names tuple (#4406)

hoefling:
Is this fix for #4406 acceptable? My concern is that in this solution, the tuple case is merely a duplication of the list case, with only the output formatting being different.

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -593,7 +593,7 @@ class StubGenerator(mypy.traverser.TraverserVisitor):
         name = repr(getattr(rvalue.args[0], 'value', '<ERROR>'))
         if isinstance(rvalue.args[1], StrExpr):
             items = repr(rvalue.args[1].value)
-        elif isinstance(rvalue.args[1], ListExpr):
+        elif isinstance(rvalue.args[1], (ListExpr, TupleExpr)):
             list_items = cast(List[StrExpr], rvalue.args[1].items)
             items = '[%s]' % ', '.join(repr(item.value) for item in list_items)
         else:
--- a/test-data/unit/stubgen.test
+++ b/test-data/unit/stubgen.test
@@ -569,6 +569,20 @@ _X = namedtuple('_X', ['a', 'b'])
 
 class Y(_X): ...
 
+[case testNamedtupleAltSyntaxFieldsTuples]
+from collections import namedtuple, x
+X = namedtuple('X', ())
+Y = namedtuple('Y', ('a',))
+Z = namedtuple('Z', ('a', 'b', 'c', 'd', 'e'))
+[out]
+from collections import namedtuple
+
+X = namedtuple('X', [])
+
+Y = namedtuple('Y', ['a'])
+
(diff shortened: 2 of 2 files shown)


elazarg:
I don't think the formatting is important.

hoefling:
@elazarg: I thought about that too, but then I'm unsure about getting the proper formatting for the field names (`['spam', 'eggs']` vs. `('spam', 'eggs')`). Wouldn't I need to check the exact type again for that? Smth like this:

```python
elif isinstance(rvalue.args[1], (ListExpr, TupleExpr)):
    list_items = cast(List[StrExpr], rvalue.args[1].items)
    format = '[%s]' if isinstance(rvalue.args[1], ListExpr) else '(%s)'
    items = format % ', '.join(repr(item.value) for item in list_items)
```

elazarg:
Why use different formatting?

hoefling:
Because otherwise the stub would lie about the actual code? I mean, there must be a reason why a distinction between the list and string case was implemented the way it is now; I don't want to ignore it without knowing what's the intention behind it.

elazarg:
Well I think it's the same code with different syntax - just like difference in whitespace - but it's not very important either way.

mypy/stubgen.py
```diff
@@ -596,6 +596,9 @@ def process_namedtuple(self, lvalue: NameExpr, rvalue: CallExpr) -> None:
         elif isinstance(rvalue.args[1], ListExpr):
             list_items = cast(List[StrExpr], rvalue.args[1].items)
             items = '[%s]' % ', '.join(repr(item.value) for item in list_items)
+        elif isinstance(rvalue.args[1], TupleExpr):
+            list_items = cast(List[StrExpr], rvalue.args[1].items)
+            items = '(%s)' % ', '.join(repr(item.value) for item in list_items)
```
gvanrossum: This would actually be broken if the original tuple had only one item. Reusing the previous stanza would avoid that.

hoefling: @gvanrossum oh, now I see - without the trailing comma, it's a generator expression...

hoefling:
Changes: Using the suggestion @elazarg has made to avoid the error; added a unit test that validates cases where fields were passed as a tuple with no elements/one element/multiple elements. Try the added test out by running `pytest -v -n1 mypy/test/teststubgen.py::StubgenPythonSuite::testNamedtupleAltSyntaxFieldsTuples`.

elazarg:
(You probably want `-n0` - no parallelism at all)

hoefling:
Oh, this indeed makes much more sense - thanks! My learning curve for pytest and its plugins is still pretty steep...

hoefling:
@gvanrossum: @elazarg: is the PR now mature enough to be accepted or did I do something wrong? Sorry if I'm being annoying, this is my first time doing a PR so I don't know if I did everything right...

elazarg:
LGTM, but I am not familiar enough with stubgen.

gvanrossum:
Thanks!

Issue #4409, opened 2017-12-25
Proposal: treat "obvious" return type as annotated


elazarg:
(Not sure if it should be here or in typing)

Consider this function:
```python
def f():
    return "hello"
def g(x):
    if x: return A(1)
    else: return A(2)
```

`f` obviously (without any deep analysis) returns `str`, and `g` returns A. Why not use this information? This pattern is very common, and taking advantage of it can help in precise checking and remove clutter (for example ` -> Union[Tuple[int,  int, str], Tuple[int, str, bool]]`).

I propose treating calls to functions whose return expressions consists solely of literals and constructor calls as if they were declared with the returned type (join or union of the return types).

kirbyfan64:
Well, I always felt that the main reason for this was to allow mypy to be used on already-existing codebases, so the unannotated functions don't need to be type-safe (yet). This would change that behavior.

ethanhs:
@kirbyfan64 I think I agree with what you are saying, but I also think there isn't a reason something like this couldn't be behind a flag, I've heard from several people that this is a significant pain point, as people don't want to have to do a lot of work to annotate obvious functions, so big +1 from me.

elazarg:
@kirbyfan64 the proposal does not require unannotated functions to be type-safe. It simply allows annotated functions that use them to be slightly more type safe (peek into the immediate return expressions, if there is one).

elazarg:
To elaborate:
```python
class A: pass

def f(): return A(1)  # no error here, even though A has no no-arg __init__

def g() -> None:
    f().bar()  # error here
```

gvanrossum:
However when is it obvious enough? Where do you draw the line?

elazarg:
Either a literal or a direct constructor call on a globally defined class. (Part of the idea is that the return statements are also the documentation).

ethanhs:
Expression atoms would be straightforward enough (not certain about identifiers however). At some point this could probably be expanded to entire expressions I believe (down the road).

elazarg:
Yes, that could be a nice path. Atoms, then (recursively) tuples/lists, then simple names. My hope is this could be done as part of semantic analysis, just like the analysis of the annotations themselves.

kirbyfan64:
Could a director constructor call really be safe though? e.g.:

```python
def func():
    A = lambda x: x
    return A(None)
```

Or even:

```python
def func():
    return A(1)

func.__globals__['A'] = lambda x: x
```

[...]

PR #4411, opened 2017-12-26
[WIP] Treat obvious return types as annotated

elazarg:
Implementing #4409.

* Only literal types for now
* The join is very naive (of course, it _must_ be very simple)

`defn.type` is used on master to signify unannotated functions and bad signatures; I think distinguishing between the purposes is important regardless of the proposal.

Also:

- [ ] Find a good name for the feature.

(I open it early to accompany the discussion at #4409)


elazarg:
Closing for now as it isn't likely to see daylight anytime soon.

PR #4415, opened 2017-12-30
Update builtin_types.rst

gpkc:
The Markdown wasn't rendering the inline code correctly if it was not surrounded by spaces.


gvanrossum:
Thanks!

PR #4417, opened 2018-01-02
Removal of section regarding comparison between mypy and PyPy

a4ff7810:
(Hopefully) clarified and corrected the section on how mypy differs from PyPy.


gvanrossum:
Honestly I think we should just delete the question about PyPy. It probably stems from the time when mypy actually transpiled your code and ran it. I've never heard from anyone who was confused between the two in the past 2-3 years.

JukkaL:
Yes, the comparison really makes no sense any more and can be deleted.

gvanrossum:
Thanks!

a4ff7810:
Not at all :). Happy to help, even if it's just with the smallest bits and bobs. Thank _you_ and everybody else for your amazing and continued efforts.

Issue #2931, opened 2017-03-01
AssertionError in deeply nested class scenario (incremental with warm cache)


ilevkivskyi:
This is a small remainder of #2535 

Example to reproduce: run ``mypy -i main.py`` twice (without touching any files in between) where:
* main.py
```python
import ntcrash
```
* ntcrash.py
```python
from typing import NamedTuple

class C:
    def f(self) -> None:
        class D:
            class E:
                def g(self) -> None:
                    A = NamedTuple('A', [('x', int), ('y', int)])
                    self.a = A(1, 1)
        self.c = D.E().a
```
Possible fix would be to store all "anonymous" classes in ``globals``.

ilevkivskyi:
One needs to add ``reveal_type(ntcrash.C().c)`` between two runs to trigger the crash.

hauntsaninja:
I think this is fixed on master, possibly by #10431

PR #2855, opened 2017-02-12
Fix crash on access to local classes

ilevkivskyi:
Fixes #2559 

The idea is the same as in PR #2553. The only difference is that mangled names are always stored in ``globals``, otherwise there would be a problem for a method in a nested class that is itself inside a method.

PR #2553, opened 2016-12-11
Fix for NamedTuple in method [WIP]

gvanrossum:
Fixes #2535.

PR #4419, opened 2018-01-02
Allow subclassing NamedTuple as a nested class

### LINKED ISSUE #4349: mypy can't find type defined inside a function.
### Steps to reproduce

mypy can't seem to find a class defined inside a function.  For example:

```python
import typing

def foo() -> typing.Any:
    class Bar(typing.NamedTuple):
        member: str

    o = Bar(member='fee')
    return o
```

### Current behavior

I get the following when I run mypy on the above code (saved in a file named "t1.py"):

t1.py:4: error: Name 'Bar' is not defined
t1.py:7: error: Missing positional arguments "typename", "fields", "verbose", "rename", "kwargs" in call to "Bar"

### Expected behavior

No errors or warning from mypy.

### mypy --version output

mypy 0.540

elliott-beach:
Fixes #4349.

This 1-line change allows subclassing NamedTuple inside a function.

The current problem is that inner NamedTuples aren't added to the symbol table.
Inner classes get added in `self.setup_class_def_analysis(defn)` , so we just call that method before calling `analyze_namedtuple_classdef`, which looks up the node in the symbol table.

As far as I can tell, `setup_class_def_analysis` does the work that would be done in pass 1's `visit_class_defn`, which, for inner classes, adds the defn to the local symbol table and the global symbol table, under a mangled name. That should also happen for NamedTuples.

I added some semaneval unit tests for this change to verify the change in behavior, but that may not be the right place. I might want to add a test to prove that local NamedTuples need to be go in the global symbol table as well.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -956,8 +956,9 @@ class SemanticAnalyzerPass2(NodeVisitor[None], SemanticAnalyzerPluginInterface):
                 tvars.extend(base_tvars)
         return remove_dups(tvars)
 
-    def is_namedtuple_classdef(self, defn: ClassDef):
-        return 'typing.NamedTuple' in [ getattr(base_expr, 'fullname', None) for base_expr in defn.base_type_exprs ]
+    def is_namedtuple_classdef(self, defn: ClassDef) -> bool:
+        base_exprs = defn.base_type_exprs
+        return any(getattr(b, 'fullname', None) == 'typing.NamedTuple' for b in base_exprs)
 
     def analyze_namedtuple_classdef(self, defn: ClassDef) -> Optional[TypeInfo]:
         # special case for NamedTuple
@@ -1054,12 +1055,10 @@ class SemanticAnalyzerPass2(NodeVisitor[None], SemanticAnalyzerPluginInterface):
                 defn.info._fullname = self.cur_mod_id + '.' + local_name
                 defn.fullname = defn.info._fullname
                 if self.type and self.is_namedtuple_classdef(defn):
-                    # Special Case for NamedTuple: store in class because that is what NamedTuples expect.
+                    # Special case for NamedTuple.
                     self.type.names[local_name] = node
                 else:
                     self.globals[local_name] = node
-
-
 
     def analyze_base_classes(self, defn: ClassDef) -> None:
         """Analyze and set up base classes.
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -671,6 +671,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None], SemanticAnalyzerPluginInterface):
             if self.analyze_typeddict_classdef(defn):
                 yield False
                 return
+            self.setup_class_def_analysis(defn)
             named_tuple_info = self.analyze_namedtuple_classdef(defn)
             if named_tuple_info is not None:
                 # Temporarily clear the names dict so we don't get errors about duplicate names
@@ -704,7 +705,6 @@ class SemanticAnalyzerPass2(NodeVisitor[None], SemanticAnalyzerPluginInterface):
                     if key not in named_tuple_info.names or key != '__doc__'
                 })
             else:
-                self.setup_class_def_analysis(defn)
                 self.analyze_base_classes(defn)
                 self.analyze_metaclass(defn)
                 defn.info.is_protocol = is_protocol
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -956,6 +956,9 @@ class SemanticAnalyzerPass2(NodeVisitor[None], SemanticAnalyzerPluginInterface):
                 tvars.extend(base_tvars)
         return remove_dups(tvars)
 
+    def is_namedtuple_classdef(self, defn: ClassDef):
+        return 'typing.NamedTuple' in [ getattr(base_expr, 'fullname', None) for base_expr in defn.base_type_exprs ]
+
     def analyze_namedtuple_classdef(self, defn: ClassDef) -> Optional[TypeInfo]:
         # special case for NamedTuple
         for base_expr in defn.base_type_exprs:
@@ -964,10 +967,14 @@ class SemanticAnalyzerPass2(NodeVisitor[None], SemanticAnalyzerPluginInterface):
                 if base_expr.fullname == 'typing.NamedTuple':
                     node = self.lookup(defn.name, defn)
                     if node is not None:
+                        if self.type or self.is_func_scope():
+                            name = defn.name + '@' + str(defn.line)
+                        else:
+                            name = defn.name
                         node.kind = GDEF  # TODO in process_namedtuple_definition also applies here
                         items, types, default_items = self.check_namedtuple_classdef(defn)
                         info = self.build_namedtuple_typeinfo(
-                            defn.name, items, types, default_items)
+                            name, items, types, default_items)
                         node.node = info
                         defn.info.replaced = info
                         defn.info = info
@@ -1046,7 +1053,13 @@ class SemanticAnalyzerPass2(NodeVisitor[None], SemanticAnalyzerPluginInterface):
                 local_name = defn.info._fullname + '@' + str(defn.line)
                 defn.info._fullname = self.cur_mod_id + '.' + local_name
                 defn.fullname = defn.info._fullname
-                self.globals[local_name] = node
+                if self.type and self.is_namedtuple_classdef(defn):
+                    # Special Case for NamedTuple: store in class because that is what NamedTuples expect.
+                    self.type.names[local_name] = node
+                else:
+                    self.globals[local_name] = node
+
+
 
     def analyze_base_classes(self, defn: ClassDef) -> None:
         """Analyze and set up base classes.
--- a/test-data/unit/check-class-namedtuple.test
+++ b/test-data/unit/check-class-namedtuple.test
@@ -667,3 +667,11 @@ class HasStaticMethod(NamedTuple):
         return 4
 
 [builtins fixtures/property.pyi]
+
+[case testNamedTupleLocalScope]
+from typing import NamedTuple, Any, Tuple
+
+def function() -> Tuple:
+    class InnerNamedTuple(NamedTuple):
+        x: int
+    return InnerNamedTuple(x=0)
--- a/test-data/unit/check-incremental.test
+++ b/test-data/unit/check-incremental.test
@@ -1925,6 +1925,21 @@ tmp/crash.py:17: error: Revealed type is 'crash.B@13[builtins.int*]'
 main:2: error: Revealed type is 'crash.A@5'
 main:3: error: Revealed type is 'crash.B@13[builtins.int*]'
 
+[case testIncrementalSubclassNamedTuple]
+from a import C
+reveal_type(C().a)
+[file a.py]
+from typing import NamedTuple
+class C:
+    def __init__(self) -> None:
+        class A(NamedTuple):
+            x: int
+        self.a = A(1)
+[out1]
+main:2: error: Revealed type is 'Tuple[builtins.int, fallback=a.C.A@4]'
+[out2]
+main:2: error: Revealed type is 'Tuple[builtins.int, fallback=a.C.A@4]'
+
 [case testGenericMethodRestoreMetaLevel]
 from typing import Dict
 
--- a/test-data/unit/semanal-namedtuple.test
+++ b/test-data/unit/semanal-namedtuple.test
@@ -132,6 +132,48 @@ MypyFile:1(
       __main__.N@2)
     PassStmt:2()))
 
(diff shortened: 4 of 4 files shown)


JelleZijlstra:
Thanks! Could you also add a direct test for local namedtuples (i.e., not a semanal test)?

elliott-beach:
done!

elliott-beach:
friendly ping on this

gvanrossum:
I hope someone else can approve/merge this; I'd be lying if I said I understand the repercussions of moving `self.setup_class_def_analysis(defn)` up. Sorry. :-(

mypy/semanal.py
```diff
@@ -671,6 +671,7 @@ def analyze_class_body(self, defn: ClassDef) -> Iterator[bool]:
             if self.analyze_typeddict_classdef(defn):
                 yield False
                 return
+            self.setup_class_def_analysis(defn)
```
ethanhs: While this works, I'm not sure if this is the best change that could be made. As we can see in [`analyze_namedtuple_classdef`](https://github.com/elliott-beach/mypy/blob/ea573e23e08350783a4d3f8f98ee89cfdbb57983/mypy/semanal.py#L967):
(referencing [this TODO](https://github.com/elliott-beach/mypy/blob/ea573e23e08350783a4d3f8f98ee89cfdbb57983/mypy/semanal.py#L2287)), 
the root of the issue is that the analysis and processing of namedtuples is unconditionally global. If I understand things correctly, your change adds the node early to the symbol table so it is coerced into being a local definition by `setup_class_def_analysis`. That works, but is rather brittle.

elliott-beach: I agree that this is incomplete because we need to handle local definitions of namedutple in a few other places, but I would argue that we should call `setup_class_def_analysis` for namedtuples.  Local classes are added to the symbol table in `setup_class_def_analysis`, and there is no reason not to add local namedtuples at that point as well. 

This does not seem early to me because semantic analysis proceeds in two steps of (1) adding nodes with basic info to the symbol table and then (2) adding more complex type info. I think here is the right place to do step (1) for NamedTuples.

elliott-beach: The TODO comments by @JukkaL  appear to be referring to assigning `node.kind` as `GDEF` or `LDEF`.
If that is all, I can add commits to handle that, and then I think this would be complete.

ethanhs: Okay I think this is acceptable.

elliott-beach: Great, Let me fix the todos.

elliott-beach: Actually, my understanding is LDEF and GDEF for NamedTuples, TypedDicts, and Enums is a longstanding issue that is not the cause of the specific problem this PR is solving, and I don't know how to devise a correct solution with confidence. So I won't address those.

elliott-beach: @JukkaL Would you look at this? I think you could merge this / request changes with confidence.

elliott-beach:
@JukkaL Would you look at this? I think you could merge this / request changes with confidence.

JukkaL:
I'll have a look. This could have implications for incremental and fine-grained incremental modes. In particular, it's possible for a local named tuple to leak to the surrounding class, and this might plausibly cause trouble:

```py
class A:
    def f(self) -> None:
        class N(NamedTuple):
            ...
        self.n = N(...)
```

elliott-beach:
> it's possible for a local named tuple to leak to the surrounding class, and this might plausibly cause trouble

I'll be looking at what you've been doing on dmypy to gain a better understanding of this.

gvanrossum:
> I'll be looking at what you've been doing on dmypy to gain a better understanding of this.

I don't think dmypy has anything to do with this, does it?

elliott-beach:
My understanding was @jukkal was referring to dmypy when talking about
fine-grained incremental mode.

gvanrossum:
Not quite -- `dmypy` is a recent experimental "daemon" mode that is separate from incremental (`-i`, not new) and fine-grained (super experimental) modes.

The problem Jukka was referring to can occur with regular incremental mode. If there's a `NamedTuple` definition local to a function that's used to declare the function's return type, then (IIRC) the JSON cache file written by incremental mode will contain a reference to that type by name, which won't be resolvable when the cache file is read back in on a subsequent incremental run.

In any case, you nerd-sniped me, and I constructed the following example (with your patch applied):
- Create file `a.py`:
```py
from typing import NamedTuple
class C:
    def __init__(self) -> None:
        class A(NamedTuple):
            x: int
        self.a = A(1)
```
- Run `mypy a.py`
- Create file `main.py`:
```py
from a import C
reveal_type(C().a)
```
- Run `mypy -i a.py` -- crash!

Without your patch, the line `self.a = A(1)` fails.

Note that if instead of `class A...` we use `A = NamedTuple('A', [('x', int)])` things have a happy ending -- that's because class definitions are processed differently than `NamedTuple(...)` calls.

In case you're interested, here's the traceback:
```
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/guido/src/mypy/mypy/__main__.py", line 11, in <module>
    main(None)
  File "/Users/guido/src/mypy/mypy/main.py", line 66, in main
  File "/Users/guido/src/mypy/mypy/main.py", line 119, in type_check_only
    options=options)
  File "/Users/guido/src/mypy/mypy/build.py", line 218, in build
  File "/Users/guido/src/mypy/mypy/build.py", line 2001, in dispatch
    process_graph(graph, manager)
  File "/Users/guido/src/mypy/mypy/build.py", line 2298, in process_graph
    process_fresh_scc(graph, prev_scc, manager)
  File "/Users/guido/src/mypy/mypy/build.py", line 2375, in process_fresh_scc
    graph[id].fix_cross_refs()
  File "/Users/guido/src/mypy/mypy/build.py", line 1701, in fix_cross_refs
    self.manager.options.quick_and_dirty)
  File "/Users/guido/src/mypy/mypy/fixup.py", line 22, in fixup_module_pass_one
    node_fixer.visit_symbol_table(tree.names)
  File "/Users/guido/src/mypy/mypy/fixup.py", line 84, in visit_symbol_table
    self.quick_and_dirty)
  File "/Users/guido/src/mypy/mypy/fixup.py", line 284, in lookup_qualified_stnode
    assert key in names, "Cannot find %s for %s" % (key, name)
AssertionError: Cannot find A for a.C.A
```

elliott-beach:
I appreciate the very detailed explanation on this! I'm reading through the related discussion (like #2855) while thinking about a solution.

elliott-beach:
Sorry that this is so long, but I wanted to document my thought process as I debugged this. I'm planning on going with approach (1) below.

**TL;DR;** The problem is local namedtuples are stored under the classe's symbol table but local classes are stored under the global symbol table, and subclasses of namedtuple under my patch have inconsistent behavior that combines both.

Consider the file a.py: 
```python
from typing import NamedTuple
class C:
    def __init__(self) -> None:
        class A(NamedTuple):
            x: int
        class B:
            def __init__(self, x):
                self.x = x
        D = NamedTuple('D', [('x', int)])
        self.a = A(1)
        self.b = B(1)
        self.d = D(1)

```

`B` and `D` work, but `A` is broke for incremental.

Inner NamedTuples defined with `D = NamedTuple('D', ...)` expect to be stored under the _containing class_ because the `_fullname` attribute of the symbol table node is `a.C.D`, local to the class.

Whereas inner classes expect to be stored in the _global_ symbol table because the `_fullname` for classes is module.LocalClass, e.g., `a.B`.

This matters because if the node with fullname `a.B` were stored in `a.C.B` (`a.C`'s symbol table), then serialization will assume that this is a cross-reference and not store the full definition of `B`:
https://github.com/python/mypy/blob/6fce774f001045a654ad56b8934a51da95b48c32/mypy/nodes.py#L2418-L2421

Likewise, under my patch,  `A` has the fullname `a.C.A` but it is stored in the symbol table for `a`, not `a.C`, so the cache does not write the definition.

`build_namedtuple_typeinfo`, expects the class to be stored under `a.C`; after the namedtuple info is returned it has a bunch of internal references to `a.C.A`.

I see three  **possible approaches:**

1) Store `A` under the `C`'s namespace as `a.C.A`. This requires the least changes and would be the path I am inclined to, but I wanted to ask about the other approaches. I feel like it may leave technical debt because the behavior is different for NamedTuples and TypeDicts than inner classes.

2) Change the `fullname` for `A` to be `a.A` and store it under the global namespace. This would require modifying `build_namedtuple_typeinfo` and require the same change for `D`, as well as typedicts.  However I think it makes more sense to store inner namedtuples under the class scope.

3) Change the `fullname` for `A` and `B` to `a.C.A` and `a.C.B` respectively, storing inner classes and inner NamedTuple classes under the class symbol table. This makes sense to me because this is how `D` as well as inner TypedDicts are stored.

I believe the reason inner classes are stored under the global symbol table but inner namedtuples/typedicts are stored in the class symbol table is that the code was written by different developers. ilevkivskyi implemented caching local classes in #2855, whereas gvanrossum implemented caching inner TypedDicts  and `D`-style NamedTuples in #2553. My insight here is not 20/20 though.

elliott-beach:
Also related: https://github.com/python/mypy/issues/2931

> Possible fix would be to store all "anonymous" classes in globals.

mypy/semanal.py
```diff
@@ -964,10 +968,14 @@ def analyze_namedtuple_classdef(self, defn: ClassDef) -> Optional[TypeInfo]:
                 if base_expr.fullname == 'typing.NamedTuple':
                     node = self.lookup(defn.name, defn)
                     if node is not None:
+                        if self.type or self.is_func_scope():
+                            name = defn.name + '@' + str(defn.line)
```
elliott-beach: The name has to be mangled so that, for incremental mode, it matches the name of type (but it makes sense anyway so it does not leak outside the containing scope).

elliott-beach:
Bump for feedback. @JukkaL I did fix the type leaking issue by storing NamedTuples class definitions under the outer class, if it exists.

mypy/semanal.py
```diff
@@ -956,6 +956,10 @@ def get_all_bases_tvars(self, defn: ClassDef, removed: List[int]) -> TypeVarList
                 tvars.extend(base_tvars)
         return remove_dups(tvars)
 
+    def is_namedtuple_classdef(self, defn: ClassDef) -> bool:
+        base_exprs = defn.base_type_exprs
+        return any(getattr(b, 'fullname', None) == 'typing.NamedTuple' for b in base_exprs)
```
JukkaL: This duplicates similar logic in `analyze_namedtuple_classdef`. There should be only one implementation of looking for a `typing.NamedTuple` base class.

Style nit: We avoid the use of `getattr` if there's a reasonable way around it, since it can't be type checked.

mypy/semanal.py
```diff
@@ -1046,7 +1054,11 @@ def setup_class_def_analysis(self, defn: ClassDef) -> None:
                 local_name = defn.info._fullname + '@' + str(defn.line)
                 defn.info._fullname = self.cur_mod_id + '.' + local_name
                 defn.fullname = defn.info._fullname
-                self.globals[local_name] = node
+                if self.type and self.is_namedtuple_classdef(defn):
```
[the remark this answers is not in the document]
elliott-beach: The problem is local namedtuples are stored under the classes's symbol table but local classes are stored under the global symbol table. We might want to fix that first. (I mentioned this in https://github.com/python/mypy/pull/4419#issuecomment-357538358). As you suggested a little refactoring might be more effective than this haphazard fix, I would suggest storing local classes under the global symbol table. Then we wouldn't need this.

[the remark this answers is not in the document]
elliott-beach: I went a little it more into the technical details in https://github.com/python/mypy/pull/4419#issuecomment-357538358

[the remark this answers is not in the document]
elliott-beach: I would be delighted to make that change if you agree.

ethanhs:
So is the decision here to do the refactor first?

gvanrossum:
This is now quite stale -- there was no response to the code review from February, and there's now (of course) a merge conflict. Should we just close it? (That would make me sad given the effort put into this, but sometimes it's the only alternative.)

elliott-beach:
I would say the assertion that there was no response to the code review is just not truth. Look! https://github.com/python/mypy/pull/4419#discussion_r166819036
I did snub Ethan, apparently.

Anyway, after getting a real job I view this whole affair as nonessential tinkering that may benefit no-one, which is probably why Jukka doesn't have time to respond. So long, and thanks for all the fish.

gvanrossum:
Thanks for clarifying, and sorry for misreading the issue.

PR #4424, opened 2018-01-03
Mention distinction between Generator and Iterator

carljm:
As seen in https://github.com/python/typeshed/issues/1806, `Generator[..., None, None]` is not quite equivalent to `Iterator[...]`, due to the `close()` method on generators. In most cases this probably doesn't matter, since the `close()` method is infrequently used, so I've left the basic structure of the docs the same (recommending `Iterator[...]` first for simple generators). But it's at least worth a mention.

### DIFF
--- a/docs/source/kinds_of_types.rst
+++ b/docs/source/kinds_of_types.rst
@@ -988,7 +988,9 @@ annotated the first example as the following:
            yield i * i
            
 This is slightly different from using ``Iterable[int]`` or ``Iterator[int]``,
-since generators have a ``close()`` method that generic iterables don't.
+since generators have ``close()``, ``send()``, and ``throw()`` methods that
+generic iterables don't. If you will call these methods on the returned
+generator, use the ``Generator`` type instead of ``Iterable`` or ``Iterator``.
 
 .. _async-and-await:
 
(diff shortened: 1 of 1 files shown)


docs/source/kinds_of_types.rst
```diff
@@ -986,6 +986,9 @@ annotated the first example as the following:
    def squares(n: int) -> Generator[int, None, None]:
        for i in range(n):
            yield i * i
+           
+This is slightly different from using ``Iterable[int]`` or ``Iterator[int]``,
+since generators have a ``close()`` method that generic iterables don't.
```
ilevkivskyi: Generators have also ``throw`` and ``send``, maybe you could say explicitly that if one wants to call ``send``, ``throw`` or ``close``, then one should use ``Generator[T, None, None]`` (this can actually be a generic type alias). Maybe you can illustrate this with a short example, like:
```python
x: Iterator[int]
x.send(None) # Error!
x.close() # Error!
y: Generator[int, None, None]
... # etc.
```

carljm:
I think it should definitely be mentioned here regardless; if you also want it mentioned in Common Issues I can add that.

I added mention of ``send()`` and ``throw()``. I don't feel like this note is important enough to warrant its own example; I think the added text is clear enough. If you feel strongly that an example is needed, let me know and I will add it.

PR #4428, opened 2018-01-03
More improvements to fine-grained incremental mode

JukkaL:
This fixes several issues in fine-grained incremental mode. The biggest
change is adding additional type checking passes in fine-grained
incremental mode. This is not very principled, but it works around some
problems with type inference. A better longer-term approach might be to
increase the number of passes in all modes.

### DIFF
--- a/mypy/server/aststrip.py
+++ b/mypy/server/aststrip.py
@@ -156,6 +156,10 @@ class NodeStripVisitor(TraverserVisitor):
 
     def visit_member_expr(self, node: MemberExpr) -> None:
         self.strip_ref_expr(node)
+        # These need to cleared for member expressions but not for other RefExprs since
+        # these can change based on changed in a base class.
+        node.is_new_ref = False
+        node.is_inferred_def = False
         if self.is_duplicate_attribute_def(node):
             # This is marked as an instance variable definition but a base class
             # defines an attribute with the same name, and we can't have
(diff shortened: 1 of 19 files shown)


mypy/server/deps.py
```diff
@@ -235,6 +235,9 @@ def process_lvalue(self, lvalue: Expression) -> None:
         # TODO: star lvalue
 
     def get_non_partial_lvalue_type(self, lvalue: RefExpr) -> Type:
+        if lvalue not in self.type_map:
+            # Likely a block considered unreacable during type checking.
```
msullivan: `unreachable`

mypy/server/aststrip.py
```diff
@@ -109,6 +109,14 @@ def visit_assignment_stmt(self, node: AssignmentStmt) -> None:
             assert isinstance(lvalue, NameExpr)
             assert self.type is not None  # Because self.is_class_body is True
             del self.type.names[lvalue.name]
+        if self.type and not self.is_class_body:
+            # TODO: Handle multiple assignment
+            # TODO: Merge with above
+            if len(node.lvalues) == 1:
+                lvalue = node.lvalues[0]
+                if isinstance(lvalue, MemberExpr) and lvalue.is_new_def:
```
msullivan: Indicate that `is_new_def` implies that this must be a self assignment?

PR #4433, opened 2018-01-06
Create an issue template

ilevkivskyi:


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
+Please provide more information to help us understand the issue:
+
+* Are you reporting a bug, or opening a feature request?
+* Please insert below the code you are checking with mypy,
+  or a mock-up repro if the source is private. We would appreciate
+  if you try to simplify your case to a minimal repro.
+* What is the actual behavior/output?
+* What is the behavior/output you expect?
+* What are the versions of mypy and Python you are using?
```
ethanhs: You probably want to add how this can be checked eg `python -V` and `mypy -V`.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+  Do you see the same issue after installing mypy from Git master?
```
ethanhs: Also add instructions for this `python3 -m pip install git+https://github.com/python/mypy`

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,16 @@
...
+* What are the mypy flags you are using? (For example --strict-optional)
+* If mypy crashed with a traceback, please paste
```
ethanhs: Maybe add hint about `--show-traceback`?

ilevkivskyi:
Thank you for review @ethanhs! I however think that all three proposed additions should not be in the template for two reasons:
* First, it should be a template, i.e. it should list "what"s and not "how"s. Also ``--show-traceback`` flag and mypy version are printed by mypy itself on an internal error. As well, installing from Git is covered both on RtD and in mypy README.
* Second, I want it to be short. In my experience people tend to skip reading instructions if they are longer than few lines. I would rather make it shorter, not longer.

ethanhs:
@ilevkivskyi ok, I dismissed review. I prefer to have the information all there so people don't need to go looking for it (reducing the odds of them doing it). But your choice.

ilevkivskyi:
OK, it looks like there are no more comments. I will merge this now, and we will see how it goes.

PR #2375, opened 2016-10-29
Fix crash for deferred methods introduced by #2193 (self-type).

gvanrossum:
Fixes #2372.

@elazarg Please note.

PR #4434, opened 2018-01-07
Annotate functions in tests

elazarg:
Seems like some functions are accidentally unannotated. For example, `testInplaceSetitem` should have actually failed since the list fixture does not declare addition for `int`; but it doesn't, since mypy seems to silently infer `Any` for the attribute.

### DIFF
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -125,13 +125,18 @@ class A:
     def __init__(self) -> None:
         self.x = object()
 a = None # type: A
-a.y
-a.y = object()
+a.y  # E: "A" has no attribute "y"
+a.y = object()  # E: "A" has no attribute "y"
 a.x
 a.x = object()
-[out]
-main:6: error: "A" has no attribute "y"
-main:7: error: "A" has no attribute "y"
+
+[case testInferAttributeUnannotatedInit]
+class A:
+    def __init__(self):
+        self.x = object()
+
+a.x
+a.x = object()
 
 [case testArgumentTypeInference]
 
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -124,17 +124,20 @@ A().f = None # E: Cannot assign to a method
 class A:
     def __init__(self) -> None:
         self.x = object()
-a = None # type: A
+a: A
 a.y  # E: "A" has no attribute "y"
 a.y = object()  # E: "A" has no attribute "y"
 a.x
 a.x = object()
 
-[case testInferAttributeUnannotatedInit]
+[case testReferToInvalidAttributeUnannotatedInit]
 class A:
     def __init__(self):
         self.x = object()
 
+a: A
+a.y  # E: "A" has no attribute "y"
+a.y = object()  # E: "A" has no attribute "y"
 a.x
 a.x = object()
 
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -122,7 +122,7 @@ A().f = None # E: Cannot assign to a method
 [case testReferToInvalidAttribute]
 
 class A:
-    def __init__(self):
+    def __init__(self) -> None:
         self.x = object()
 a = None # type: A
 a.y
--- a/test-data/unit/check-inference.test
+++ b/test-data/unit/check-inference.test
@@ -1403,7 +1403,7 @@ b[{}] = 1
 
 [case testInferDictInitializedToEmptyAndUpdatedFromMethod]
 map = {}
-def add():
+def add() -> None:
     map[1] = 2
 [builtins fixtures/dict.pyi]
 [out]
@@ -1959,7 +1959,7 @@ class C:
 
 if bool():
     f()
-    1 + '' # E: Unsupported left operand type for + ("int")
+    1 + '' # E: Unsupported operand types for + ("int" and "str")
 [builtins fixtures/list.pyi]
 [out]
 
@@ -1976,7 +1976,7 @@ class C:
 
 if bool():
     f()
-    1 + '' # E: Unsupported left operand type for + ("int")
+    1 + '' # E: Unsupported operand types for + ("int" and "str")
 [builtins fixtures/list.pyi]
 [out]
 
@@ -1992,6 +1992,6 @@ class C:
 
 if bool():
     f([])
-    1 + '' # E: Unsupported left operand type for + ("int")
+    1 + '' # E: Unsupported operand types for + ("int" and "str")
 [builtins fixtures/list.pyi]
 [out]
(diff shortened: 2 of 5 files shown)


ilevkivskyi:
I am wondering if some of those tests don't have annotations intentionally, for example to check that something works/fails even in non-annotated code. Maybe a safer way would be to make copies of tests that you changed, and keep both versions: with and without annotations?

elazarg:
Tests for unannotated code would have named appropriately. I have left out some tests that seem to address unannotated code intentionally.

As for copies, they will define the expected behaviour, of which I am not sure. [PEP 484](https://www.python.org/dev/peps/pep-0484/#the-meaning-of-annotations) does not define what should and should not be checked, inferred or reported in unannotated functions and methods.

gvanrossum:
IIRC mypy analyzes the bodies of all unannotated functions just as if they were annotated with all `Any`s, and then suppresses errors -- but keeps inferences about things outside the function (in particular instance attributes). I would tread carefully here. Surely git blame can tell you who wrote these tests originally? (My guess would be Jukka. :-)

elazarg:
Mypy does indeed, and these tests were found in an attempt to move the decision to a single place.

What exactly do you mean by "tread carefully"? IMHO the real question is the specification: what is the expected behaviour? The tests I've touched do no seem to intentionally try to define expected behaviour for the special case of unannotated functions, otherwise they neglect the more common and important case.

The inference seems like a waste of computation time, especially at early stages of "gradualization" of a large codebase, since the inferred type of will always be `Any`. The fact that a function is "declared" to be un-type-checkable means that non-obvious inference - i.e. inferring an attribute for an inferred type - is expected to be wrong. Simple inference of attributes can be done at the semantic analysis phase. 

Blame:
* check-classes, testReferToInvalidAttribute: @JukkaL, [5 years ago](testReferToInvalidAttribute); mypy-style
* check-inference.test, testInferDictInitializedToEmptyAndUpdatedFromMethod: Guido, [Dec 2015](https://github.com/python/mypy/commit/25b614cab2a50067a47c88ba8f4739c08f00145a)
* check-modules.test, testDeferredClassContext: #2375 (Guido)
* check-statements.test, testInplaceSetitem: #2051 (@gnprice)

test-data/unit/check-classes.test
```diff
@@ -122,7 +122,7 @@ A().f = None # E: Cannot assign to a method
 [case testReferToInvalidAttribute]
 
 class A:
-    def __init__(self):
+    def __init__(self) -> None:
```
JukkaL: I think that omitting the return type was intentional here. It would be okay to have two variants of this method, one with an annotation for `__init__` and another without.

test-data/unit/check-inference.test
```diff
@@ -1403,7 +1403,7 @@ b[{}] = 1
 
 [case testInferDictInitializedToEmptyAndUpdatedFromMethod]
 map = {}
-def add():
+def add() -> None:
```
JukkaL: It's unclear if the annotation was omitted intentionally. I'd suggest creating two variants of the test case.

test-data/unit/check-inference.test
```diff
@@ -1959,7 +1959,7 @@ class C:
 
 if bool():
     f()
-    1 + '' # E: Unsupported left operand type for + ("int")
+    1 + '' # E: Unsupported operand types for + ("int" and "str")
```
JukkaL: I'd prefer changing the test case so that we don't depend on `int.__add__` instead (it's a little arbitrary to rely on it here). For example, use `1()` instead of `1 + ''` which doesn't require a fixture change.

test-data/unit/check-inference.test
```diff
@@ -1976,7 +1976,7 @@ class C:
...
```
JukkaL: Similar to above, I'd prefer using `1()` instead (and not changing the fixture).

test-data/unit/check-inference.test
```diff
@@ -1992,6 +1992,6 @@ class C:
 
 if bool():
     f([])
-    1 + '' # E: Unsupported left operand type for + ("int")
+    1 + '' # E: Unsupported operand types for + ("int" and "str")
```
JukkaL: Similar to above.

test-data/unit/check-modules.test
```diff
@@ -1312,7 +1312,7 @@ class A:
     def f(self) -> str: return 'foo'
 class B(A):
     def f(self) -> str: return self.x
-    def initialize(self): self.x = 'bar'
+    def initialize(self) -> None: self.x = 'bar'
```
JukkaL: Not having a return type may have well been intentional. I'd prefer either keeping the original test case or adding another variant with a return type annotation.

test-data/unit/check-statements.test
```diff
@@ -353,7 +353,7 @@ a @= 1  # E: Argument 1 to "__imatmul__" of "A" has incompatible type "int"; exp
 
 [case testInplaceSetitem]
 class A(object):
-    def __init__(self):
+    def __init__(self) -> None:
```
JukkaL: Good catch!

test-data/unit/fixtures/list.pyi
```diff
@@ -26,7 +26,8 @@ class list(Generic[T]):
 
 class tuple(Generic[T]): pass
 class function: pass
-class int: pass
+class int:
```
JukkaL: As discussed above, I wouldn't change the fixture.

elazarg:
Done.

I still think this "just in case" is misplaced here. What's so wrong about changing tests, then (worst case) finding the context through some breakage at some edge case, and then test for it thoroughly? The resulting system will be more stable this way. And this is worst-case scenario.

This PR now defines an expected behaviour that was not explicitly defined previously, just in case. IMO with this definition, the distinction between annotated and unannotated functions is not obvious.

JukkaL:
@elazarg The cost of having a few potentially redundant test cases (assuming that they aren't slow to run) is pretty minor. On the other hand, finding out whether a test case is potentially redundant (or invalid) can be hard, since it's 1) difficult to determine if the same feature or edge case is tested elsewhere and 2) a test case may test some edge case that isn't obvious from the name of the test case. So the default option is to leave a potentially poorly thought out test case instead of removing it -- but using a more descriptive name is still a good thing, or adding a comment such as "TODO: It's unclear if the missing annotation is intentional or not".

elazarg:
I don't care about redundancy. I am only interested in the decision to use unannotated functions as a source for type information; I don't think such a decision should be the result of a (possibly  accidental) omission in a test. But since it is now decided, what's left is to document this behaviour.

JukkaL:
There have been bugs that were triggered only by unannotated functions. Sometimes these can be pretty surprising. Mypy still does a bunch of analysis for unannotated functions (even though it rarely generates type errors).

PR #4436, opened 2018-01-08
Improve documentation of fine-grained incremental mode

JukkaL:
Major updates to docstrings/comments in `mypy.server.astdiff`,
`mypy.server.astmerge` and `mypy.server.aststrip`.

### DIFF
--- a/mypy/server/aststrip.py
+++ b/mypy/server/aststrip.py
@@ -1,6 +1,40 @@
-"""Strip AST from semantic information.
+"""Strip/reset AST in-place to match state after semantic analysis pass 1.
 
-This is used in fine-grained incremental checking to reprocess existing AST nodes.
+Fine-grained incremental mode reruns semantic analysis (passes 2 and 3)
+and type checking for *existing* AST nodes (targets) when changes are
+propagated using fine-grained dependencies.  AST nodes attributes are
+often changed during semantic analysis passes 2 and 3, and running
+semantic analysis again on those nodes would produce incorrect
+results, since these passes aren't idempotent. This pass resets AST
+nodes to reflect the state after semantic analysis pass 1, so that we
+can rerun semantic analysis.
+
+(The above is in contrast to behavior with modules that have source code
+changes, for which we reparse the entire module and reconstruct a fresh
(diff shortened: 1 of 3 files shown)


mypy/server/aststrip.py
```diff
@@ -1,6 +1,40 @@
...
+
+* This is currently pretty fragile, as we must carefully undo whatever
+  changes can be made in semantic analysis passes 2 and 3, including changes
+  to symbol tables.
+
+* We reuse existing AST nodes because it makes it relatively straightforward
+  to reprocess only a single target within a module efficiently. If there
+  was a way to parse a single target within a file, in time proportional to
+  the size of the target, we'd rather create fresh AST nodes than strip them.
+  Alas, no such facility exists and building it is non-trivial.
```
ilevkivskyi: Why it is hard? If we know that source code is not changed, then we can use the start line number from AST, and the last line number from tokenizer (matching ``DEDENT``, and maybe logical ``NEWLINE``). This however will still require tokenizing whole file. Alternatively, maybe we could update ``typed-ast`` to save the last line, not only the first line (but this would be a large project). Does it make sense to open an issue for this?

msullivan: Another possible approach is that if we could serialize full ASTs, we could snapshot serialized ASTs before semantic analysis and use deserialization to produce fresh ASTs. We're going to need AST serialization anyways in order to produce cache files for fine-grained incremental mode, right?

msullivan: Guido points out that this is probably not true--ASTs are big and so it is likely to be better to just parse files on-demand in such a mode.

JukkaL: There are at least a few complicating factors. First, the module top level is one target, and parsing only it (without processing functions) and then combining the AST fragments sounds pretty involved. Second, tokenizing the entire file is not efficient (proportional to the size of the file), and tokenizing just some stretches of a file sounds hard. For example, consider encodings, `__future__` imports, and keeping track of the spans of AST nodes efficiently. Finding line N in a text file is an O(n) operation without some additional index data structure.

PR #4438, opened 2018-01-08
Various fixes to the fine-grained incremental mode

JukkaL:
These fix various issues seen when testing with real-world code.

### DIFF
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -129,10 +129,20 @@ class Errors:
     # Set to True to show column numbers in error messages.
     show_column_numbers = False  # type: bool
 
-    # Stack of active fine-grained incremental checking targets within
-    # a module. The first item is always the current module id.
+    # State for keeping track of the current fine-grained incremental mode target.
     # (See mypy.server.update for more about targets.)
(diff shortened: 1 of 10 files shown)


mypy/errors.py
```diff
@@ -129,10 +129,20 @@ class Errors:
     # Set to True to show column numbers in error messages.
     show_column_numbers = False  # type: bool
 
-    # Stack of active fine-grained incremental checking targets within
-    # a module. The first item is always the current module id.
+    # State for keeping track of the current fine-grained incremental mode target.
```
msullivan: Is `Errors` really the best place to be tracking what target is currently being processed?

JukkaL: Not really. `Errors` is tracking this primarily for convenience -- it's the only place that currently has enough context and is available pretty much everywhere. A better approach might be to have a separate context class which would track the current scope and target, and `Errors` would get access to this context (but it would only need read-only access). Since this PR doesn't change the responsibilities of `Errors` significantly (the wart was already there before), I'd prefer to refactor this in a separate PR.

PR #4443, opened 2018-01-08
Exit with code 2 on blocking errors

### LINKED ISSUE #2754: Use a different error code on early exit
When running mypy from a script, it's sometimes nice to know when mypy successfully completed a full typechecking run vs exited early due to blocking errors.  Perhaps mypy could exit with exit code `2` when exiting due to blocking errors?

msullivan:
Exit with code 2 on blocking errors, while continuing to exit with 1
when there are nonblocking errors.  Add test infrastructure.

Fixes #2754.

### DIFF
--- a/mypy/test/testcmdline.py
+++ b/mypy/test/testcmdline.py
@@ -79,7 +79,9 @@ def test_python_cmdline(testcase: DataDrivenTestCase) -> None:
                                            path))
     else:
         out = normalize_error_messages(out)
-        out.append('== Return code: {}'.format(result))
+        obvious_result = 1 if out else 0
+        if obvious_result != result:
+            out.append('== Return code: {}'.format(result))
         assert_string_arrays_equal(testcase.output, out,
                                    'Invalid output ({}, line {})'.format(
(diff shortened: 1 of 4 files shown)


msullivan:
(This depends on #4396 because they would otherwise conflict)

mypy/test/testcmdline.py
```diff
@@ -78,6 +79,7 @@ def test_python_cmdline(testcase: DataDrivenTestCase) -> None:
                                            path))
     else:
         out = normalize_error_messages(out)
+        out.append('== Return code: {}'.format(result))
```
gvanrossum: Hm, this added a lot of noise to a lot of tests. Perhaps only print the return code when it's not equal to `not len(out)`?

msullivan:
Those cases don't add blockers to the errors object, but they raise CompileError directly, which causes us to exit with 2. One of the tests I added checks for this.

gvanrossum:
Ah, I see. But I'd still like to see fewer tests with different output.

gvanrossum:
Whee!

PR #4446, opened 2018-01-09
More improvements to documentation of fine-grained incremental mode

JukkaL:
This updates the documentation of fine-grained dependencies and the top-level
fine-grained incremental logic.

### DIFF
--- a/mypy/server/update.py
+++ b/mypy/server/update.py
@@ -56,10 +56,10 @@ Here's a summary of how a fine-grained incremental program update happens:
 * Determine which modules have changes in their source code since the
   previous update.
 
-* Process changed modules one at a time. Perform a full update for each
-  module, but only report the final set of error messages, since the
-  intermediate states can generate bogus errors due to only seeing a
-  partial set of changes.
+* Process changed modules one at a time. Perform a separate full update
+  for each changed module, but only report the errors after all modules
+  have been processed, since the intermediate states can generate bogus
+  errors due to only seeing a partial set of changes.
 
 * Each changed module is processed in full. We parse the module, and
   run semantic analysis to create a new AST and symbol table for the
--- a/mypy/server/update.py
+++ b/mypy/server/update.py
@@ -1,49 +1,119 @@
-"""Update build result by incrementally processing changed modules.
+"""Update build by processing changes using fine-grained dependencies.
 
 Use fine-grained dependencies to update targets in other modules that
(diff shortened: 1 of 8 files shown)


mypy/server/update.py
```diff
@@ -1,49 +1,119 @@
...
-* Replace old ASTs of the modules that we reprocessed earlier with
-  the new ones, but try to retain the identities of original externally
-  visible AST nodes so that we don't (always) need to patch references
-  in the rest of the program.
-* Semantically analyze and type check the stale targets.
-* Repeat the previous steps until nothing externally visible has changed.
+  previous update.
+
+* Process changed modules one at a time. Perform a full update for each
+  module, but only report the final set of error messages, since the
```
msullivan: What exactly is meant by final?

mypy/server/update.py
```diff
@@ -1,49 +1,119 @@
...
+* Fine-grained dependencies allow processing only the relevant parts
+  of modules indirectly affected by a change. Say, if only one function
+  in a large module is affected by a change in another module, only this
+  function is processed. The classic incremental mode always processes
+  an entire file as a unit, which is typically much slower.
 
-We perform a fine-grained incremental program update like this:
+* It's possible to independently process individual modules within an
+  import cycle (SCC). Small incremental changes can be fast independent
+  of the size of the related SCC. In classic incremental mode, any change
```
ilevkivskyi: As I understand, there still can be cycles in fine grained dependencies (like mutually recursive functions). So there should be a similar notion of an SCC made of smaller locations (functions instead of modules.) Or am I missing something?

JukkaL: The concept of SCC is currently implicit in fine-grained incremental mode. We can freely follow fine-grained dependencies independent of import relationships, since everything is in memory all the time. We can more or less treat the entire program as a single SCC. In case of module `x` depending on types inferred in module `y` and vice versa, we may need to perform multiple passes of type checking across the entire stale set of targets, I think. This isn't implemented yet, though.

Also, some dependency cycles are naturally handled by just following fine-grained dependencies. Say, `f` and `g` are mutually recursive, defined in different modules, and both signatures change. We first update module containing `f` and its signature changes. This causes `g` to be stale and we reprocess it (using the old definition for `g`). Then we update the module containing `g` and its signature changes as well. This triggers another update of `f`, and after than everything should be consistent.

PR #4447, opened 2018-01-09
Integrate fine-grained incremental checking with dmypy

JukkaL:
Fine-grained incremental mode is enabled through the `--experimental`
dmypy flag.

As a mostly unrelated change, print daemon tracebacks to log
to make debugging crashes easier.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -252,11 +252,11 @@ class Server:
             result = mypy.build.build(sources=sources,
                                       options=self.options)
         except mypy.errors.CompileError as e:
-            messages = ''.join(s + '\n' for s in e.messages)
+            output = ''.join(s + '\n' for s in e.messages)
             if e.use_stdout:
-                out, err = messages, ''
+                out, err = output, ''
             else:
-                out, err = '', messages
+                out, err = '', output
             return {'out': out, 'err': err, 'status': 2}
         messages = result.errors
         manager = result.manager
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -13,12 +13,14 @@ import os
 import socket
 import sys
 import time
+import traceback
 
-from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence
+from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Tuple
 
 import mypy.build
 import mypy.errors
 import mypy.main
+import mypy.server.update
 from mypy.dmypy_util import STATUS_FILE, receive
 from mypy.gclogger import GcLogger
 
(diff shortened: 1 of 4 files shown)


mypy/dmypy_server.py
```diff
@@ -212,6 +235,71 @@ def check(self, sources: List[mypy.build.BuildSource],
...
+        self.fine_grained_manager = mypy.server.update.FineGrainedBuildManager(manager, graph)
+        status = 1 if messages else 0
+        self.previous_messages = messages[:]
+        if messages:
+            messages.append('')
+        self.fine_grained_initialized = True
+        self.file_modified = {}  # type: Dict[str, float]
+        for source in sources:
+            assert source.path
+            self.file_modified[source.path] = os.stat(source.path).st_mtime
```
msullivan: We could populate this table before we do the build, and that would eliminate the race conditions where we miss changes that happen during the checking, right?

mypy/dmypy_server.py
```diff
@@ -212,6 +235,71 @@ def check(self, sources: List[mypy.build.BuildSource],
...
+        except mypy.errors.CompileError as e:
+            # TODO: Handle blockers here
+            assert False, str('\n'.join(e.messages))
+        messages = result.errors
+        manager = result.manager
+        graph = result.graph
+        self.fine_grained_manager = mypy.server.update.FineGrainedBuildManager(manager, graph)
+        status = 1 if messages else 0
+        self.previous_messages = messages[:]
+        if messages:
```
msullivan: Maybe return `''.join(s+'\n' for s in messages)` instead?

mypy/dmypy_server.py
```diff
@@ -212,6 +235,71 @@ def check(self, sources: List[mypy.build.BuildSource],
...
+            return self.fine_grained_increment(sources)
+
+    def initialize_fine_grained(self, sources: List[mypy.build.BuildSource]) -> Dict[str, Any]:
+        try:
+            # TODO: alt_lib_path
+            result = mypy.build.build(sources=sources,
+                                      options=self.options)
+        except mypy.errors.CompileError as e:
+            # TODO: Handle blockers here
+            assert False, str('\n'.join(e.messages))
```
msullivan: Something like this should work well enough, I think:
```
        except mypy.errors.CompileError as e:
            messages = ''.join(x+'\n' for x in e.messages)
            out, err = messages, '' if e.use_stdout else '', messages
            return {'out': out, 'err': err, 'status': 2}
```

mypy/dmypy_server.py
```diff
@@ -212,6 +235,71 @@ def check(self, sources: List[mypy.build.BuildSource],
...
+            mtime = os.stat(path).st_mtime
+            if path not in self.file_modified or self.file_modified[path] != mtime:
+                self.file_modified[path] = mtime
+                changed.append((source.module, path))
+        modules = {source.module for source in sources}
+        omitted = [source for source in self.previous_sources if source.module not in modules]
+        for source in omitted:
+            path = source.path
+            assert path
+            if not os.path.isfile(path):
```
msullivan: What cases other than the file being removed would cause being omitted?

JukkaL: The file could be removed from the root sources, but it could still continue to exist on the file system. I'll add a comment.

PR #4459, opened 2018-01-12
Update the core team

gvanrossum:


gvanrossum:
@msullivan Can you merge when the tests pass?

PR #4467, opened 2018-01-13
stubgen: infer None return type for functions without return statement

### LINKED ISSUE #4181: Infer return None in stubgen
Is it possible that when a function don't have any return statement, mypy can infer that the return type is `None`?

Both Pycharm and Typescript has this feature.

hoefling:
Hi, I added a proposal for the feature implementation requested in #4181, would be grateful for a review! I have doubts about the ellipsis case - should a function containing an ellipsis also have None return type?

Fixes #4181


hoefling:
@ilevkivskyi Wow, that was fast! Thank you for the response and for the hint, I will remember that!

PR #4473, opened 2018-01-16
Fine-grained: Fix refresh of class nested inside function

### LINKED ISSUE #4462: Crash on refreshing subclass defined in function in fine-grained incremental mode
Test case the reproduces the crash:

```
[case testRefreshSubclassNestedInFunction]
from a import C
def f() -> None:
    class D(C): pass
[file a.py]
class C: pass
[file a.py.2]
```

Traceback:

```
/Users/jukka/src/mypy/mypy/test/testfinegrained.py:73: in run_case
    new_messages = fine_grained_manager.update(modules)
/Users/jukka/src/mypy/mypy/server/update.py:222: in update
    result = self.update_single(next_id, next_path)
/Users/jukka/src/mypy/mypy/server/update.py:285: in update_single
    graph)
/Users/jukka/src/mypy/mypy/server/update.py:725: in propagate_changes_using_dependencies
    todo = find_targets_recursive(triggered, deps, manager.modules, up_to_date_modules)
/Users/jukka/src/mypy/mypy/server/update.py:785: in find_targets_recursive
    deferred = lookup_target(modules, target)
/Users/jukka/src/mypy/mypy/server/update.py:977: in lookup_target
    assert isinstance(node, (MypyFile, TypeInfo))
E   AssertionError
```

Apparently there is a dependency with target of form `mod.func.Cls`, which is not a valid target. The correct target to refresh would be `mod.func`.

JukkaL:
Fix crash on refreshing class nested inside function. Typically this
seemed to happen when the base class got triggered. Classes nested
within functions aren't supposed to be targets separate from the
surrounding function, and this PR effectively adds them to the 
surrounding target. Previously they were treated as separate targets,
which broke some assumptions when looking up targets.

Also refactor current target tracking in dependency generation since 
the logic would otherwise have been pretty messy. The main change 
is to not include classes defined within functions in target names.

Some weird dependencies are still generated for nested classes, but
they may be harmless -- extra dependencies should not be a problem.

Fixes #4462.

### DIFF
--- a/mypy/server/deps.py
+++ b/mypy/server/deps.py
@@ -116,7 +116,7 @@ def get_dependencies_of_target(module_id: str,
     """Get dependencies of a target -- don't recursive into nested targets."""
     # TODO: Add tests for this function.
     visitor = DependencyVisitor(type_map, python_version)
-    visitor.enter_file_scope(module_id)
+    visitor.scope.enter_file(module_id)
     if isinstance(target, MypyFile):
         # Only get dependencies of the top-level of the module. Don't recurse into
         # functions.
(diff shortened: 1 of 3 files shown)


JukkaL:
Using the scope class in Errors might work, at least with some tweaks. It would remove some duplicated logic. A good idea!

PR #4478, opened 2018-01-17
Add a note about typeshed to the issue template

ilevkivskyi:
It happened several times when users were reporting typeshed issues here. I think it makes sense to add a note about this to the issue template.


ISSUE_TEMPLATE.md
```diff
@@ -1,3 +1,6 @@
+Note: if you are reporting a wrong signature of a function or a class in
+the standard library, then typeshed tracker is better suited for this report.
```
gvanrossum: "... then _the_ typeshed tracker ..." Also I'd add a link.

ilevkivskyi: Note that the template is shown in plain text, so I don't use any Markdown for the link.

PR #4482, opened 2018-01-18
report columns in 1-based, not 0-based

### LINKED ISSUE #4475: Are column numbers base-0 or base-1?
IIUC Column numbers (as shown by `--show-column-numbers` are 0-based (as opposed to line numbers, which are 1-based). But in Emacs `*compilation*` mode, it seems column numbers (if present) are interpreted as 1-based, so the cursor is placed one character to the left of the actual error.

Do we know what other tools that understand/use column numbers do?

elliott-beach:
fixes #4475

### DIFF
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -411,8 +411,7 @@ class Errors:
             if file is not None:
                 if self.show_column_numbers and line is not None and line >= 0 \
                         and column is not None and column >= 0:
-                    col = 1 + column
-                    srcloc = '{}:{}:{}'.format(file, line, col)
+                    srcloc = '{}:{}:{}'.format(file, line, 1+column)
                 elif line is not None and line >= 0:
                     srcloc = '{}:{}'.format(file, line)
                 else:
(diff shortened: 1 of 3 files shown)


mypy/errors.py
```diff
@@ -411,7 +411,8 @@ def format_messages(self, error_info: List[ErrorInfo]) -> List[str]:
             if file is not None:
                 if self.show_column_numbers and line is not None and line >= 0 \
                         and column is not None and column >= 0:
-                    srcloc = '{}:{}:{}'.format(file, line, column)
+                    col = 1 + column
+                    srcloc = '{}:{}:{}'.format(file, line, col)
```
gvanrossum: Any reason why you're not just inlining `column + 1` here?

elliott-beach:
My thinking was doing math inside format string is awkward, but I have no
strong opinion.

gvanrossum:
No good deed goes unpunished, flake8 wants spaces around the `+`.

gvanrossum:
Thanks!

PR #4483, opened 2018-01-18
Caching for fine-grained incremental mode

msullivan:
Work in progress for cache loading for fine-grained incremental mode.

I've tested it some manually and it seems to work all right. This needs a test suite.

A good improvement would be to load from caches without doing a full build.

### DIFF
--- a/mypy/server/update.py
+++ b/mypy/server/update.py
@@ -378,7 +378,7 @@ def update_single_isolated(module: str,
     sources = get_sources(previous_modules, [(module, path)])
     invalidate_stale_cache_entries(manager.saved_cache, [(module, path)])
 
-    manager.missing_modules = set()
+    manager.missing_modules.clear()
     try:
         graph = load_graph(sources, manager)
     except CompileError as err:
@@ -497,7 +497,8 @@ def delete_module(module_id: str,
     # TODO: Remove deps for the module (this only affects memory use, not correctness)
     assert module_id not in graph
     new_graph = graph.copy()
-    del manager.modules[module_id]
+    if module_id in manager.modules:
+        del manager.modules[module_id]
     if module_id in manager.saved_cache:
         del manager.saved_cache[module_id]
     components = module_id.split('.')
--- a/mypy/server/update.py
+++ b/mypy/server/update.py
@@ -170,6 +170,10 @@ class FineGrainedBuildManager:
         self.blocking_error = None  # type: Optional[Tuple[str, str]]
         # Module that we haven't processed yet but that are known to be stale.
         self.stale = []  # type: List[Tuple[str, str]]
+        # Disable the cache so that load_graph doesn't try going back to disk
+        # for the cache. This is kind of a hack and it might be better to have
+        # this directly reflected in load_graph's interface.
+        self.options.cache_dir = os.devnull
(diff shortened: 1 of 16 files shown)


msullivan:
This now has a test suite and supports doing a fine-grained incremental build to initialize.

There is one pretty dodgy hack to disable reading from the cache after the initial build that might need rethinking, and a few test cases have been disabled, but I think this is reviewable.

gvanrossum:
(Be sure to remove the [WIP] from the title before submitting.)

mypy/server/update.py
```diff
@@ -717,6 +719,11 @@ def propagate_changes_using_dependencies(
         triggered: Set[str],
         up_to_date_modules: Set[str],
         targets_with_errors: Set[str]) -> List[Tuple[str, str]]:
+    """Transitively rechecks targets based on triggers and the dependency map.
+
+    Returns a list (module id, path) tuples representing modules that contain
+    a target that needs to be reprocessed but that has not been parsed yet."""
```
JukkaL: Style nit: move """ to a separate line.

mypy/test/testfinegrained.py
```diff
@@ -41,31 +42,40 @@ class FineGrainedSuite(DataSuite):
...
+    # Whether to use the fine-grained cache in the testing. This is overridden
+    # by a trivial subclass to produce a suite that uses the cache.
+    use_cache = False
+
+    # Decide whether to skip the test. This could have been structured
+    # as a filter() classmethod also, but we want the tests reported
+    # as skipped, not just elided.
+    def should_skip(self, testcase: DataDrivenTestCase) -> bool:
+        if self.use_cache:
+            if testcase.name.endswith("-skip-cache"): return True
```
JukkaL: Style nit: move `return True` to a separate line.

mypy/test/testfinegrained.py
```diff
@@ -41,31 +42,40 @@ class FineGrainedSuite(DataSuite):
...
+            # TODO: In caching mode we currently don't well support
+            # starting from cached states with errors in them.
+            if testcase.output and testcase.output[0] != '==': return True
```
JukkaL: Similar to above.

mypy/test/testfinegrained.py
```diff
@@ -41,31 +42,40 @@ class FineGrainedSuite(DataSuite):
...
+        else:
+            if testcase.name.endswith("-skip-nocache"): return True
```
JukkaL: Similar to above.

mypy/test/testfinegrained.py
```diff
@@ -90,7 +100,7 @@ def run_case_inner(self, testcase: DataDrivenTestCase, cache: bool) -> None:
             # cache, now we need to set it up
             if fine_grained_manager is None:
                 messages, manager, graph = self.build(main_src, testcase, sources_override,
-                                                      build_cache=False, enable_cache=cache)
+                                                      build_cache=False, enable_cache=True)
```
JukkaL: Is the change to a `True` argument on purpose?

msullivan: Yeah. This path only executes when the cache is on.

PR #4486, opened 2018-01-18
Fix type of forward reference to a decorated class method

### LINKED ISSUE #4485: Forward reference to decorated class method has type Any
In the program below, the type inferred for `cls.g` in `A.f` is `Any`, even though it should be `def (x: builtins.str)`:

```py

T = TypeVar('T')
def dec() -> Callable[[T], T]: pass

class A:
    @classmethod
    def f(cls) -> None:
        reveal_type(cls.g)  # Any  <-- unexpected

    @classmethod
    @dec()
    def g(cls, x: str) -> None:
        pass

    @classmethod
    def h(cls) -> None:
        reveal_type(cls.g)  # def (x: builtins.str)  <-- ok
```

This works correctly when the decorated method is accessed using an instance type (such as `self`).

In fine-grained incremental mode the types can be more precise, which can result in additional errors in fine-grained incremental mode.

JukkaL:
The decorator case wasn't properly handled before when accessing an attribute
through a type object -- the type of the forward reference defaulted to `Any`.

The fix doesn't work at module top level since module top levels can't
be deferred, but the reference would generally fail at runtime anyway.

Fixes #4485.


JukkaL:
This is now ready for review. I've addressed the issues in Dropbox internal code.

test-data/unit/check-inference.test
```diff
@@ -1785,6 +1785,29 @@ y = 0
...
+
+    @classmethod
+    @dec()
+    def g(cls, x: str) -> None:
+        pass
+
+    @classmethod
+    def h(cls) -> None:
+        reveal_type(cls.g)  # E: Revealed type is 'def (x: builtins.str)'
+[builtins fixtures/classmethod.pyi]
```
ilevkivskyi: Maybe also add `reveal_type(A.g)` after the class definition here?

mypy/checkmember.py
```diff
@@ -467,8 +467,12 @@ def analyze_class_attribute_access(itype: Instance,
...
 
     if is_decorated:
-        # TODO: Return type of decorated function. This is quick hack to work around #998.
-        return AnyType(TypeOfAny.special_form)
+        assert isinstance(node.node, Decorator)
+        if node.node.type:
+            return node.node.type
+        else:
+            not_ready_callback(name, context)
+            return AnyType(TypeOfAny.special_form)
```
ilevkivskyi: I thought that we normally return `TypeOfAny.from_error` in all cases where the may be an error.

Issue #3905, opened 2017-09-01
Unable to suppress `No library stub file for module...` error


henryJack:
Hi,

In the command line (on windows 10 using Python 3.6) I run `mypy <my_module_name>`. The output is very informative and has revealed a lot of missing types in my python module. 

However, There is currently no mypy switch to suppress the 300+ errors of the form:

-` No library stub file for module 'numpy'`
-`No library stub file for module 'scipy'`
-`No library stub file for module 'matplotlib'`

What is the best way to get around this? I need a *clean sheet* from the mypy command if I want to integrate a `mypy` test ainto the build process for my module on gitlab or github.

ethanhs:
You probably want to use `–ignore-missing-imports` or one of the more relaxed options specified here: https://mypy.readthedocs.io/en/latest/command_line.html#following-imports-or-not

That will cause Mypy to not raise errors if it doesn't find stubs for the modules.

henryJack:
Ah, So sorry. I thought I tried that! 

Yep, that switch works perfectly, thanks for the fantastic tool!

boompig:
Is this the only way to suppress this warning?

The documentation for `--ignore-missing imports` states:

> --ignore-missing-imports suppresses error messages about imports that cannot be resolved (see Following imports or not? for some examples).

This seems extremely broad, as (I believe) this also squelches errors on `import foo` when foo does not exist. What about a flag to suppress no stub files for selected modules that are known to not have stub files. Currently, a large number of python modules lack stub files, so having a way to surgically suppress this warning would be extremely helpful, especially in automated builds.

gvanrossum:
Yeah, an extra flag could help here. Do you want to help adding it by submitting a PR?

boompig:
Let me give this a try.

lingz:
In the end, we ended up suppressing stub warnings one by one, by just doing:

```
```

henryJack:
@boompig - To add to @lingz nice example... You can find more information about the current module specific settings in the [per module flags](https://mypy.readthedocs.io/en/latest/config_file.html#per-module-flags) section in the docs.

s1291:
> –ignore-missing-imports

You mean: `--ignore-missing-imports`

stuaxo:
It would be great if there was some way of specifying a requirements file as a source for packages to ignore.

thejaminator:
It seems said that https://github.com/python/mypy/pull/4491 got closed.

[...]

PR #4491, opened 2018-01-19
Adds the --ignore-missing-stubs command line option

boompig:
--ignore-missing-stubs MODULE-1 [MODULE-2 ...] will suppress missing typeshed information for the specified modules

Includes tests for both command-line and config file invocations

Fixes issue #3905

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -10,7 +10,7 @@ flag (or its long form ``--help``)::
   $ mypy -h
   usage: mypy [-h] [-v] [-V] [--python-version x.y] [--platform PLATFORM] [-2]
               [--ignore-missing-imports]
-              [--ignore-missing-stubs MODULE [MODULE ...]]
+              [--ignore-missing-stubs MODULE-1[,MODULE-2...]]
               [--follow-imports {normal,silent,skip,error}]
               [--disallow-any-{unimported,expr,decorated,explicit,generics}]
               [--disallow-untyped-calls] [--disallow-untyped-defs]
@@ -292,7 +292,7 @@ Here are some more useful flags:
   that cannot be resolved (see :ref:`follow-imports` for some examples).
 
 .. _ignore-missing-stubs:
-- ``--ignore-missing-stubs MODULE`` suppresses error messages about imports
+- ``--ignore-missing-stubs MODULE-1[,MODULE-2...]`` suppresses error messages about imports
   for ``MODULE`` when it can be resolved but it lacks library stub files.
 
 - ``--strict-optional`` enables experimental strict checking of ``Optional[...]``
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -10,6 +10,7 @@ flag (or its long form ``--help``)::
   $ mypy -h
   usage: mypy [-h] [-v] [-V] [--python-version x.y] [--platform PLATFORM] [-2]
               [--ignore-missing-imports]
+              [--ignore-missing-stubs MODULE [MODULE ...]]
               [--follow-imports {normal,silent,skip,error}]
               [--disallow-any-{unimported,expr,decorated,explicit,generics}]
               [--disallow-untyped-calls] [--disallow-untyped-defs]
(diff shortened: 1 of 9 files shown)


gvanrossum:
Thanks! I have one issue with the UI design though. The way argparse parses flags with an arbitrary number of values unfortunately makes it hard to specify that flag followed by a filename (meant as a positional arg):
```
mypy --ignore-missing-stubs foo bar x.py
```
will complain that there's no target module, package or file. My suggested solution is to use a comma as a separator instead; you'll have to do a little parsing of the (single) argument to make that work, and it'll require some more work for mypy.ini.

JukkaL:
What about naming it `--ignore-missing-modules` instead and making it possible to ignore arbitrary modules? The third party module list in mypy is not at all complete. Besides, it could be used for ignoring user modules as well. 

It would be nice if there was an easy way to ignore an entire package. Maybe the target should act as a prefix, ignoring all submodules as well? Or we could support the `*` wildcard.

The documentation should probably cover packages and submodules as well.

gvanrossum:
Note that IIRC it's possible to ignore specific modules by putting either `ignore_missing_imports = True` or `follow_imports = skip` in a module-specific section in mypy *for the module to be ignored*. (Can someone please verify that this works?)

boompig:
> My suggested solution is to use a comma as a separator instead

@gvanrossum What about specifying the argument multiple times in the command line, like `--ignore-missing-stub <module-1> --ignore-missing-stub <module-2>`?

boompig:
On further reflection I like what Guido suggested best since it's consistent with the config file. I'm going to revise the PR

boompig:
The above commit has the new syntax and additional tests to make sure behavior is consistent with the config files.

boompig:
Can anyone help merge this PR? This feature would be very useful to several projects I'm working on.

JukkaL:
@boompig To move forward with this, please respond to my comments above.

boompig:
@JukkaL 

> What about naming it --ignore-missing-modules instead and making it possible to ignore arbitrary modules? The third party module list in mypy is not at all complete. Besides, it could be used for ignoring user modules as well.

I think this may be confused with `--ignore-missing-imports`. Using stubs makes it clear that you're not ignoring the module, you're just ignoring the lack of a stubs file.

> It would be nice if there was an easy way to ignore an entire package. Maybe the target should act as a prefix, ignoring all submodules as well? Or we could support the * wildcard.

I like the idea of a * wildcard. This is a very good idea. Not sure if I have the bandwidth to work on this though, since it's a non-trivial change.

> The documentation should probably cover packages and submodules as well.

Can you explain further?

JukkaL:
> I think this may be confused with --ignore-missing-imports. Using stubs makes it clear that you're not ignoring the module, you're just ignoring the lack of a stubs file. But if you feel strongly, I don't mind changing it.

The current approach only works for certain third party modules that happen to be whitelisted in mypy. It doesn't allow ignoring missing stdlib stubs and non-whitelisted third party modules, which is too limiting (and also confusing).

We could continue to use `--ignore-missing-stubs` as the name of the option but allow ignoring arbitrary modules, not just whitelisted third party ones. The new note could be generated for known modules only. It would be better to generate the note for both known third party and stdlib modules, however, instead of just third party modules.

> >  The documentation should probably cover packages and submodules as well.

> Can you explain further?

You could mention that ignoring package `x` doesn't affect imports from submodules of `x` such as `x.y`. It's necessary to list each submodule that is being ignored separately, and this is not clear from the current documentation.

ethanhs:
I'm not so sure this is needed.

>Note that IIRC it's possible to ignore specific modules by putting either ignore_missing_imports = True or follow_imports = skip in a module-specific section in mypy for the module to be ignored. (Can someone please verify that this works?)

I did test this and it worked fine.

gvanrossum:
> I did test this and it worked fine.

If you're thinking of #3789, according to the OP it still parses those modules?

ethanhs:
>If you're thinking of #3789, according to the OP it still parses those modules?

Interesting. It seems if I have:

```
mod
| - __init__.py
| - lib.py
``` it will skip it with `mypy -m mod` but not `mypy mod`.

**Comment by gvanrossum** [comment]
> it will skip it with `mypy -m mod` but not `mypy mod`.

Ah, I know why. The latter invocation searches the folder named `mod` for `.py[i]` files, and it parses those regardless of the other settings. Though I'm now curious how the code in #3789 avoids that.

**Comment by msullivan** [comment]
Agree that this isn't needed.
To suppress use of a typeshed package that is *missing*, you can do
```
[mypy-module.*]
ignore_missing_imports = True
```
To suppress use of a typeshed package this is present but that you want to suppress anyways, you can use:
```
[mypy-module.*]
follow_imports = skip
follow_imports_for_stubs = True
```
(the `follow_imports_for_stubs` option is recent and post-dates this PR, but was a smaller hammer)

PR #4495, opened 2018-01-20
Don't error on re-exported imports in cycles. Fixes #4049.

### LINKED ISSUE #4049: Circular import + reexport = "error: Module '...' has no attribute '..."?
This is a minimal test case replicating what I'm experiencing in a larger project:

```
% tree
.
└── asd
    ├── __init__.py
    ├── one.py
    └── two.py

2 directories, 6 files

% cat asd/__init__.py 
from asd.one import One
from asd.two import Two

% cat asd/one.py 
class One:
    pass

% cat asd/two.py 
from asd import One

class Two:
    pass
```

Python can work with it:

```
% python -c 'from asd import Two; print(Two)'
<class 'asd.two.Two'>
```

mypy, however, doesn't like it:

```
% mypy .
asd/two.py:1: error: Module 'asd' has no attribute 'One'
```

It fails all the same when I change ``__init__.py`` to use ``import .. as ..`` (I saw this pattern mentioned in https://github.com/python/mypy/issues/3981):

```
% cat asd/__init__.py 
from asd.one import One as One
from asd.two import Two as Two
```

Apologies if there's already an issue filled for this, I searched for one briefly and haven't found any.

mypy 0.521, CPython 3.6.2

carljm:
Fixes #4049.


carljm:
So it turned out that just removing the blanket prohibition on re-exporting a node of kind `UNBOUND_IMPORTED` was all it took to make the example in #4049 (and a similar example I had independently discovered) work as expected. I had to reintroduce a specific check for a module importing from itself in order to make the `testImportFromSameModule` tests pass.

Without the changes in `semanal.py`, the added test here fails (as expected) with:

```
Expected:
  main:2: error: Revealed type is 'def () -> m.one.One' (diff)
Actual:
  tmp/m/two.py:1: error: Module 'm' has no attribute 'One' (diff)
  main:2: error: Revealed type is 'def () -> m.one.One' (diff)

  E: main:2: error: Revealed type is 'def () -> m.one.One'
  A: tmp/m/two.py:1: error: Module 'm' has no attribute 'One'
```

carljm:
Er, this isn't quite done yet; I hadn't asserted the resolved type at the right spot. On my other cyclic import example it works fully (all types resolve correctly), but on this example it leaves a type as `Any` where it shouldn't be; pushed a failing version of the test to demonstrate.

I think that bogus `Any` is a better failure mode than "Module has no attribute" false positives, so I still consider this an improvement over the status quo, but I'm looking into whether we can visit import nodes again in pass 3 to resolve these types correctly.

carljm:
Pushed my initial attempt to get the type corrected. Not working yet, will look into it more later. Any tips welcome!

elliott-beach:
I spent some time debugging this but could not figure it out. I'm eager to see what you come up with!

carljm:
This passes all tests, including the added one, but in internal Instagram codebases it now generates new "{name} is not defined" errors on certain inline imports. Haven't found a simple-case repro for that yet, tracking it down.

carljm:
Ok, this fixes the problem and causes no new errors (both in mypy test suite and in internal Instagram code). I think it's ready for an initial round of review.

carljm:
Anything I could do by way of testing or anything else to make this more attractive for review? Not complaining, I know everyone's got plenty to do (been on the "overloaded maintainer" side plenty often), just wondering if I can help move it along somehow.

ilevkivskyi:
This is a "controversial" topic, and some refactoring of symbol table kinds will happen soon. Therefore this may wait for some time, sorry.

carljm:
@ilevkivskyi Thanks for the explanation. This fixes a clear bug, the new behavior is unequivocally better, it's been thoroughly tested on a large real-world code base, and the PR is not complex; it doesn't change any major behaviors, just does some fixups in third pass on things that were missed in the second pass. It doesn't seem likely to add significant complexity to any upcoming refactor of symbol table kinds. Is there any strong reason to let it languish rather than just merging it?

JukkaL:
@ilevkivskyi Can you give specific reasons why this PR should be postponed? If you are worried about type alias refactoring, it looks to me like the required change to this PR should be pretty minor and is not enough to block this from being merged before the type alias refactor.

@carljm Fixing issues related to import cycles like #4049 is definitely high on our list of things we want to fix in the next few months or so.

mypy/semanal.py
```diff
@@ -1499,7 +1502,8 @@ def visit_import_from(self, imp: ImportFrom) -> None:
                     symbol = SymbolTableNode(GDEF, ast_node)
                     self.add_symbol(name, symbol, imp)
                     return
-            if node and node.kind != UNBOUND_IMPORTED and not node.module_hidden:
+
+            if node and not node.module_hidden and not (is_unbound and is_same_module):
```
ilevkivskyi: This line is complex (many conditions), so it deserves a comment.

mypy/semanal.py
```diff
@@ -1471,14 +1471,17 @@ def visit_import_from(self, imp: ImportFrom) -> None:
...
             node = module.names.get(id) if module else None
             missing = False
+            is_same_module = import_id == self.cur_mod_id
+            is_unbound = node and node.kind == UNBOUND_IMPORTED
             possible_module_id = import_id + '.' + id
 
             # If the module does not contain a symbol with the name 'id',
             # try checking if it's a module instead.
-            if not node or node.kind == UNBOUND_IMPORTED:
+            if not node or is_unbound:
```
ilevkivskyi: I find the original form simpler (don't need look above to `is_unbound`).

test-data/unit/check-modules.test
```diff
@@ -1943,3 +1943,24 @@ main:2: error: Name 'name' is not defined
...
+reveal_type(One)
+[file m/one.py]
+class One:
+    pass
+[file m/two.py]
+from m import One
+reveal_type(One)
+class Two:
+    pass
+[out]
```
ilevkivskyi: Add also tests for incremental mode.

test-data/unit/check-modules.test
```diff
@@ -1943,3 +1943,24 @@ main:2: error: Name 'name' is not defined
 main:3: error: Revealed type is 'Any'
 
 [builtins fixtures/module.pyi]
+
+
+[case testImportCycle]
```
ilevkivskyi: I would prefer a more descriptive name.

test-data/unit/check-modules.test
ilevkivskyi: Please also add tests for type aliases (you seem to care about them in `visit_import_from`).

test-data/unit/check-modules.test
```diff
@@ -1943,3 +1943,24 @@ main:2: error: Name 'name' is not defined
...
+[file m/__init__.py]
+from .one import One
+from .two import Two
+reveal_type(One)
+[file m/one.py]
+class One:
+    pass
+[file m/two.py]
+from m import One
+reveal_type(One)
```
JukkaL: Note that references to `One` in annotations still don't work correctly:

```
from m import One
o: One
reveal_type(o)  # Any
```

Similarly, if we make `One` a type alias and use it in an annotation, it will be treated as `Any`.

If we create a subclass of `One` in `m.two`, it seems to have an implicit `Any` base class, which is unexpected.

test-data/unit/check-modules.test
carljm: Ok, I can look at some existing incremental-mode tests to see how that's done.

carljm: So the type-alias stuff is there just because it seemed better to be complete about transferring node info (in the same way that `visit_import_from` in second pass is complete), rather than only transfer some of the node attributes. I have not had any real-world issue with type aliases and circular imports that those lines are intended to solve. I will try to see if I can repro some case involving type aliases that is fixed by those lines.

test-data/unit/check-modules.test
carljm: Good catches! I think these are probably fixable by adding more third-pass visits to different node types to propagate type info that we missed in second pass due to the cycle, but it does begin to feel a bit like a game of whack-a-mole.

carljm:
Thanks for the review comments! I see that this fix is going to get uglier as I make it more complete. I'm curious what the longer-term plan is to address import cycles in a more thorough way (I guess some kind of fixed-point iteration in semantic analysis might be needed?)

From a practical point of view doing gradual typing in a large codebase, spurious `Any` is a preferable failure mode to false positive import errors, so I'd have some preference for improving this and merging it (and I'm willing to work on getting it merge ready). It seems like at the very least the tests will be a useful addition regardless of future refactoring. But if there are specific ways that merging this will hinder progress towards a better fix, I'm open to dropping it.

ilevkivskyi:
> But if there are specific ways that merging this will hinder progress towards a better fix, I'm open to dropping it.

We discussed with @JukkaL and it looks like this PR is unlikely to cause any long-term harm, and it would be OK as a temporary partial solution (provided you fix some review comments).

JukkaL:
> I'm curious what the longer-term plan is to address import cycles in a more thorough way (I guess some kind of fixed-point iteration in semantic analysis might be needed?)

Here are a few ideas:

* Create "indirect cross-module references" (perhaps a new AST node class?) during the first pass of semantic analysis. These would be dereferenced on demand during later phases of semantic analysis. We would support multiple levels of indirection. For example, assume we have `from x import y` in module `m`. This would make `m.y` an indirect reference to `x.y`. `x.y` could be another indirect reference to, say, `a.b`. When analyzing a `y` reference in `m`, we'd follow the indirect reference chain until we reach `a.b`, which is the actual target.
* Create cross-module `ForwardRef` objects if there is a reference to a type in another module that hasn't been analyzed yet. We already support forward references within a module, and hopefully it wouldn't be hard to generalize these a bit.

carljm:
Ok, I think I've addressed all the feedback so far. I fixed several new cases (member expressions, use in type annotations) and added tests for type aliases, named tuples, and incremental mode. I left TODOs for a) inheriting from a re-exported-in-cycle import (it seemed too complex to fix up in third pass) and b) name expressions in scopes other than module-level variables. I certainly won't claim that this is a complete fix, but I think it does improve behavior incrementally for many common cases, and at the very least the tests and TODOs will be a useful starting point for a more complete fix. Review and further comments much appreciated.

I'll be very curious to see what form the more complete fix takes, having wrestled with this for awhile. I'm probably missing something key, but it feels to me like the mentioned approaches (cross-module references, using ForwardRef) are effectively just different ways to handle the book-keeping for unresolved cycles; they don't address the root problem with this diff: that we have to go back and fix up all kinds of things in third pass that normally should be handled in second pass, by effectively copying bits of code in from second pass. That doesn't feel like a path to "complete fix", it feels like whack-a-mole, where the asymptote is that third pass becomes a more and more "complete" copy of second pass. It seems like for a thorough and sustainable fix of these kinds of issues, some kind of scheme for fixed-point iteration in second pass will be needed instead. But I'm sure I'm missing something, and I look forward to learning a lot from seeing the better fix implemented :-)

Thanks @JukkaL and @ilevkivskyi for reviewing!

JukkaL:
@carljm Thanks for the updates! I'm currently prototyping what might become a cleaner fix, where we handle many exported imported names in semantic analyzer pass 2. If it works out, it might be at least a partial replacement of this PR, but your PR has been useful first step towards fixing the underlying issue. I'm going to try your test cases with my work-in-progress fix. I still don't have complete story for references to named tuples, type aliases and such with my approach.

JukkaL:
@carljm #4695 is an alternative fix to the same issue, which I believe to be somewhat more general. It handles base classes, at least, and it may be easier to generalize to fix certain other import cycle related issues.

ethanhs:
This is superseded by https://github.com/python/mypy/commit/0b7597cebcfdbe038f02c1c123c164712ab47cd1, so closing.

ilevkivskyi:
Interesting that it didn't happen automatically. Looking at the previous type when it worked, it looks like it auto-closes superseded PRs only on comments like `Fixes #PR` (which is weird).

ethanhs:
Yeah, I think github's syntax is a bit specific here. (I think closes may work, but not sure)

anowlcalledjosh:
`close`, `fix`, and `resolve`, as well as variants like `closes`, all work: https://help.github.com/articles/closing-issues-using-keywords/

PR #4496, opened 2018-01-21
Add variable name to "Need type annotation for variable" error

### LINKED ISSUE #4431: Add variable name to "Need type annotation for variable" error
It would be nice if the error message "Need type annotation for variable" showed the variable name. This helps the user in case the mypy output was from a slightly outdated version of the file, and in general seems helpful/reassuring.

elliott-beach:
fixes #4431 by changing "`need annotation for variable`" messages to "`need to annotation for 'x'`".

### DIFF
--- a/test-data/unit/check-expressions.test
+++ b/test-data/unit/check-expressions.test
@@ -1562,7 +1562,7 @@ d5 = dict(a=1, b='') # type: Dict[str, Any]
 
 [case testDictWithoutKeywordArgs]
 from typing import Dict
-d = dict() # E: Need type annotation for variable
+d = dict() # E: Need type annotation for 'd'
 d2 = dict() # type: Dict[int, str]
 dict(undefined) # E: Name 'undefined' is not defined
 [builtins fixtures/dict.pyi]
--- a/test-data/unit/check-expressions.test
+++ b/test-data/unit/check-expressions.test
@@ -1787,3 +1787,8 @@ reveal_type(a.__pow__(2)) # E: Revealed type is 'builtins.int'
 reveal_type(a.__pow__(a)) # E: Revealed type is 'Any'
 a.__pow__() # E: Too few arguments for "__pow__" of "int"
 [builtins fixtures/ops.pyi]
+
+[case testTypeAnnotationNeededMultipleAssignment]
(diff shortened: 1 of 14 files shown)


test-data/unit/fine-grained.test
```diff
@@ -1327,9 +1327,9 @@ def g() -> None: pass
 def g() -> int: pass
 [builtins fixtures/dict.pyi]
 [out]
-main:5: error: Need type annotation for variable
+main:5: error: Need type annotation for 'x'
 ==
-main:5: error: Need type annotation for variable
+main:5: error: Need type annotation for 'x'
```
ilevkivskyi: I think you should also add dedicated tests for more tricky situations like:
```python
x, y = [], []
self.a, self.b = None, None
```
(and maybe some others if you have more ideas).

elliott-beach: I added a case for multiple assignment. I think classes would be a valid case too but currently mypy doesn't appear to require annotations for `self` assignments. Other cases I can't think of. My understanding is the only assignments that can trigger this are assignments with variable names literal as lvalue, which is covered.

elliott-beach:
Sorry, I will lint before pushing next time. I had an interesting configuration issue with PyCharm causing the tabs which I just resolved.

test-data/unit/check-expressions.test
```diff
@@ -1787,3 +1787,8 @@ reveal_type(a.__pow__(2)) # E: Revealed type is 'builtins.int'
 reveal_type(a.__pow__(a)) # E: Revealed type is 'Any'
 a.__pow__() # E: Too few arguments for "__pow__" of "int"
 [builtins fixtures/ops.pyi]
+
+[case testTypeAnnotationNeededMultipleAssignment]
+x,y = [],[] # E: Need type annotation for 'x' \
```
ilevkivskyi: I know this is just a test but why not use PEP 8 rules here? (space after comma)

elliott-beach: yes yes

elliott-beach: because I need to read PEP 8 :P

PR #4502, opened 2018-01-23
Fix some fine-grained incremental bugs with newly imported files

msullivan:
I'm not yet super confident about these changes.


mypy/server/update.py
```diff
@@ -282,7 +282,7 @@ def update_single(self, module: str, path: str) -> Tuple[List[str],
         propagate_changes_using_dependencies(manager, graph, self.deps, triggered,
                                              {module},
                                              self.previous_targets_with_errors,
-                                             graph)
+                                             self.manager.modules)
```
JukkaL: `propagate_changes_using_dependencies` also gets `manager` and `graph` as arguments, so it looks like the final argument is not actually required at all.

PR #4504, opened 2018-01-24
Display arg name, not number, for functions called with kwargs

### LINKED ISSUE #4487: Sometimes it's better to display argument name instead of number
The current message for invalid argument type is sometimes not very helpful. Example:

```py
def f(a: int, b: int, c: int, d: int, e: str) -> None: pass

f(c=6, b=3, d=2, e=6, a=6)  # Error
```

This is the current message:

```
t.py:2: error: Argument 4 to "f" has incompatible type "int"; expected "str"
```

I think that including the argument name (but only if the caller used a keyword argument) would be more helpful when there are many arguments:

```
t.py:2: error: Argument "e" to "f" has incompatible type "int"; expected "str"
```

The current message is a bit problematic when a function takes many arguments. It's not rare to see functions that take 8+ arguments, and then you'll need to count the arguments pretty carefully to find the correct one.

Including the correct line/column would also help, but it's a separate issue.

elliott-beach:
Fixes #4487.
The new error message is in this format:
```python
 C(foo='')  # E: Argument "foo" to "C" has incompatible type "str"; expected "int"
```

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -25,8 +25,8 @@ from mypy.types import (
 from mypy.nodes import (
     TypeInfo, Context, MypyFile, op_methods, FuncDef, reverse_type_aliases,
     ARG_POS, ARG_OPT, ARG_NAMED, ARG_NAMED_OPT, ARG_STAR, ARG_STAR2,
-    ReturnStmt, NameExpr, Var, CONTRAVARIANT, COVARIANT, SymbolNode
-)
+    ReturnStmt, NameExpr, Var, CONTRAVARIANT, COVARIANT, SymbolNode,
+    CallExpr)
 
 
 # Constants that represent simple type checker error message, i.e. messages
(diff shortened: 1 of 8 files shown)


mypy/messages.py
```diff
@@ -25,8 +25,8 @@
 from mypy.nodes import (
     TypeInfo, Context, MypyFile, op_methods, FuncDef, reverse_type_aliases,
     ARG_POS, ARG_OPT, ARG_NAMED, ARG_NAMED_OPT, ARG_STAR, ARG_STAR2,
-    ReturnStmt, NameExpr, Var, CONTRAVARIANT, COVARIANT, SymbolNode
-)
+    ReturnStmt, NameExpr, Var, CONTRAVARIANT, COVARIANT, SymbolNode,
+    CallExpr)
```
msullivan: Nit: leave the closing paren on a line of its own

mypy/messages.py
```diff
@@ -634,8 +634,18 @@ def incompatible_argument(self, n: int, m: int, callee: CallableType, arg_type:
...
             elif arg_kind == ARG_STAR2:
                 arg_type_str = '**' + arg_type_str
-            msg = 'Argument {} {}has incompatible type {}; expected {}'.format(
-                n, target, self.quote_type_string(arg_type_str),
+
+            # For function calls with keyword arguments, use the argument name instead of the
+            # number.
+            arg_name = context.arg_names[n - 1] if isinstance(context, CallExpr) else None
+            if arg_name:
+                message_format = 'Argument "{}" {}has incompatible type {}; expected {}'
```
msullivan: I think it would be cleaner to make the computation of `arg_name` conditional (and either quoted or not) while leaving the main format string unconditional, to reduce duplication.

elliott-beach:
Thanks for the review @msullivan. I believe I addressed your changes.

PR #4505, opened 2018-01-24
Fine-grained: Treat files changed only if md5 changes

### LINKED ISSUE #4499: Fine-grained: Redundant work after git operations
Here's an example scenario where mypy daemon does a lot of unnecessary work in fine-grained incremental mode:

* I have daemon running on my git feature branch, and the daemon knows about all my latest changes. 
* I switch to an old branch, do a bit of work there (but don't invoke the daemon).
* I switch back to my original branch and invoke the daemon. 

Now the daemon thinks that a lot of files have changed. Checking out another branch touches file timestamps, and the daemon looks for file changes based on timestamps. However, there are actually no file content changes from the previous daemon update, since I'm back in the original branch. The daemon will process all files with changed timestamps, even though this work is redundant, and it can take a long time for a large codebase. It would be more efficient to recognize that the files actually have no changes and skip processing them.

(This would also work around some unreliability issues, since processing many files often triggers fine-grained incremental bugs.)

Here's an implementation sketch (similar to how the regular incremental mode already works):
* Keep track of file timestamps, sizes and sha1 hashes.
* If the timestamp and the size of a file hasn't changed, assume that it's unchanged (as an optimization).
* If the timestamp or the size of a file has changed, look at the hash of the file to determine if it has actually changed. Update the state of the file (timestamp, size and hash).

JukkaL:
Previously the daemon could do a lot of extra work if you switched
to another git branch and immediately back to the original branch,
since many file timestamps would change. Now this isn't sufficient
to consider a file as changed.

Also add a new cached file system abstraction (that is currently only
used in dmypy) to avoid redundant file system operations and to make
file system state easier to reason about. The idea is that we cache
output of previous file system operations during a single increment,
for both consistency and performance.  My plan is to eventually use it
about everywhere.

This should also make it slightly easier to switch to listening to
file system events instead of stat()ing everything, but that's not
a priority yet.

Fixes #4499.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -23,6 +23,8 @@ import mypy.main
 import mypy.server.update
 from mypy.dmypy_util import STATUS_FILE, receive
 from mypy.gclogger import GcLogger
+from mypy.fscache import FileSystemCache
+from mypy.fswatcher import FileSystemWatcher
 
 
 def daemonize(func: Callable[[], None], log_file: Optional[str] = None) -> int:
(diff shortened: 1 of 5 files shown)


JukkaL:
Travis is failing on master as well -- the failures seem unrelated to this.

JukkaL:
> My only concern is that now we have two separate mechanisms for tracking the hashes of files to see if they have changed, but they are at very different places in the abstraction stack, so it isn't too concerning.

This is a concern, but this shouldn't be significantly worse now than it was previously. In the near/medium term I want to get the regular build using the new abstraction as well. This should give a minor perf boost and also make things a bit cleaner in general.

MakonnenMak:
@JukkaL  is this item from the docs still relevant with this change?

https://mypy.readthedocs.io/en/stable/additional_features.html#remote-cache

> If you use the mypy daemon, you may want to restart the daemon each time after the merge base or local branch has changed to avoid processing a potentially large number of changes in an incremental build, as this can be much slower than downloading cache data and restarting the daemon.

PR #4506, opened 2018-01-24
Update lxml version

ilevkivskyi:


ilevkivskyi:
@JukkaL FWIW 4.1.1 works fine, let's stick with it (FYI previous pin was 4.0.0)

PR #4509, opened 2018-01-25
Tweak NamedTuple dep handling and remove it from TODO

msullivan:
For completeness, add dependency from member types to the trigger for
major namedtuple changes, since a change to a member type changes the
underlying tuple_type. In practice it seems that generating any
dependencies from NamedTuples might be redundant, since exposing the
underlying types as part of its TupleType means that any clients will
already have those dependencies.

Adds a bunch of tests because I was convinced there had to be a bug related to that.

### DIFF
--- a/mypy/server/deps.py
+++ b/mypy/server/deps.py
@@ -197,6 +197,8 @@ class DependencyVisitor(TraverserVisitor):
         # Add dependencies to base types.
         for base in o.info.bases:
             self.add_type_dependencies(base, target=target)
+        if o.info.tuple_type:
+            self.add_type_dependencies(base, target=make_trigger(target))
         # TODO: Add dependencies based on remaining TypeInfo attributes.
         super().visit_class_def(o)
         self.is_class = old_is_class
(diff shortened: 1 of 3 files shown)


mypy/server/deps.py
```diff
@@ -150,7 +150,6 @@ def __init__(self,
     # TODO (incomplete):
     #   from m import *
     #   await
-    #   named tuples
```
ilevkivskyi: If you are removing the TODO item also from here, then it is better to be sure that Python 3.6 class syntax for named tuples also works.

test-data/unit/deps-classes.test
```diff
@@ -20,7 +20,39 @@ class A: pass
...
+class A: pass
+class B: pass
+[out]
+<m.N.__init__> -> m.f
+<m.N.a> -> m.f
+<m.N> -> m.f
+<a.A> -> <m.N.a>, <m.N>, m
+<a.B> -> <m.N.a>, <m.N>, m
+
+[case namedTuple3]
```
JukkaL: Style nit: Rename test case to `testNamedTuple3`.

PR #4510, opened 2018-01-25
Handle TypedDict in diff and deps

msullivan:
Required performing a patchup in semanal to *actually* replace the
TypeInfo with the 'replaced' version.

### DIFF
--- a/mypy/server/deps.py
+++ b/mypy/server/deps.py
@@ -261,7 +261,7 @@ class DependencyVisitor(TraverserVisitor):
         elif isinstance(rvalue, CallExpr) and isinstance(rvalue.analyzed, TypedDictExpr):
             # Depend on the underlying typeddict type
             info = rvalue.analyzed.info
-            assert(info.typeddict_type is not None)
+            assert info.typeddict_type is not None
             prefix = '%s.%s' % (self.scope.current_full_target(), info.name())
             self.add_type_dependencies(info.typeddict_type, target=make_trigger(prefix))
         else:
--- a/mypy/server/deps.py
+++ b/mypy/server/deps.py
@@ -87,7 +87,7 @@ from mypy.nodes import (
     ImportFrom, CallExpr, CastExpr, TypeVarExpr, TypeApplication, IndexExpr, UnaryExpr, OpExpr,
     ComparisonExpr, GeneratorExpr, DictionaryComprehension, StarExpr, PrintStmt, ForStmt, WithStmt,
     TupleExpr, ListExpr, OperatorAssignmentStmt, DelStmt, YieldFromExpr, Decorator, Block,
-    TypeInfo, FuncBase, OverloadedFuncDef, RefExpr, SuperExpr, Var, NamedTupleExpr,
+    TypeInfo, FuncBase, OverloadedFuncDef, RefExpr, SuperExpr, Var, NamedTupleExpr, TypedDictExpr,
     LDEF, MDEF, GDEF,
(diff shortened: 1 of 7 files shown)


mypy/server/deps.py
```diff
@@ -199,6 +199,8 @@ def visit_class_def(self, o: ClassDef) -> None:
             self.add_type_dependencies(base, target=target)
         if o.info.tuple_type:
             self.add_type_dependencies(o.info.tuple_type, target=make_trigger(target))
+        if o.info.typeddict_type:
```
ilevkivskyi: I am mostly in favor of "unification" of ``TypedDict`` and ``NamedTuple``, but why can't you just use `o.info.replaced.typeddict_type` here? (Then you wouldn't need to change `semanal.py`.)

mypy/server/deps.py
```diff
@@ -258,6 +259,12 @@ def visit_assignment_stmt(self, o: AssignmentStmt) -> None:
                         self.add_type_dependencies(typ, target=make_trigger(prefix))
                         attr_target = make_trigger('%s.%s' % (prefix, name))
                         self.add_type_dependencies(typ, target=attr_target)
+        elif isinstance(rvalue, CallExpr) and isinstance(rvalue.analyzed, TypedDictExpr):
+            # Depend on the underlying typeddict type
+            info = rvalue.analyzed.info
+            assert(info.typeddict_type is not None)
```
JukkaL: Style nit: spurious parens around the expression.

test-data/unit/diff.test
```diff
@@ -562,3 +562,33 @@ B = Enum('B', 'y')
...
+[file next.py]
+from mypy_extensions import TypedDict
+Point = TypedDict('Point', {'x': int, 'y': str})
+p = Point(dict(x=42, y='lurr'))
+[builtins fixtures/dict.pyi]
+[out]
+__main__.Point
+__main__.p
+
+[case testTypedDict2]
```
JukkaL: Test also that removing or adding a field triggers a TypedDict type. And test that changing totality triggers the type.

test-data/unit/fine-grained.test
```diff
@@ -1669,3 +1669,59 @@ def f(x: a.N) -> None:
...
+[file b.py]
+from a import Point
+def foo(x: Point) -> int:
+    return x['x'] + x['y']
+[builtins fixtures/dict.pyi]
+[out]
+==
+b.py:3: error: Unsupported operand types for + ("int" and "str")
+
+[case testTypedDictUpdate]
```
JukkaL: Test case has same name as the previous test case.

test-data/unit/semanal-typeddict.test
```diff
@@ -46,8 +46,8 @@ MypyFile:1(
   ImportFrom:1(mypy_extensions, [TypedDict])
   ClassDef:2(
     A
-    BaseTypeExpr(
-      NameExpr(TypedDict [mypy_extensions.TypedDict]))
+    BaseType(
+      typing.Mapping[builtins.str, builtins.str])
```
JukkaL: What's this change?

mypy/server/deps.py
msullivan: `replaced` doesn't get serialized, so the right thing to do seemed to be following `NamedTuple` and actually doing the replacement.

test-data/unit/semanal-typeddict.test
msullivan: A consequence of the semanal changes that replace the TypeInfo with the `replaced` version.

PR #4513, opened 2018-01-26
Fix dep generation of super() and class names

msullivan:
Class names need to generate dependencies to __init__ even when not
directly being called, since they could be passed elsewhere and
called. We generate spurious dependencies in cases such as classes
being referenced in isinstance checks and expect statements, but
that's not a big deal.

Handling super() deps is bundled in with this because the __init__ fix
broke a super() using test that previously worked somewhat by accident.


mypy/server/deps.py
```diff
@@ -368,9 +368,17 @@ def visit_del_stmt(self, o: DelStmt) -> None:
...
-    # TODO
-    #   dependency on __init__ (e.g. ClassName())
-    #   super()
+    def process_global_ref_expr(self, o: RefExpr) -> None:
+        if o.fullname is not None:
+            self.add_dependency(make_trigger(o.fullname))
+
+        # If this is a reference to a type, generate a dependency to its
+        # constructor. This will be spurious in cases like isinstance
+        # calls and except statements, but that's no big deal.
```
JukkaL: What about class attribute references? If things like `Foobar.method()` depend on `FooBar.__init__`, this could have performance implications.

PR #4518, opened 2018-01-28
stubgen: don't append star arg when args list already has varargs appended

### LINKED ISSUE #3985: stubgen may generate (*args, *, ...)
I ran stubgen recursively on the asyncio package, and found this in the output:
```py
    def subprocess_exec(self, protocol_factory, program, *args, *, stdin: Any = ..., stdout: Any = ..., stderr: Any = ..., universal_newlines: bool = ..., shell: bool = ..., bufsize: int = ..., **kwargs): ...
```

hoefling:
Hi, this is a proposal to fix the issue when stubgen encounters function signatures with varargs and keyword args:
```
def f(x, *args, y=1)
```
becomes
```
def f(x, *args, *, y=1)
```
in the stub file. I also added a separate test that covers possible valid cases of having varargs mixed with keyword args in a signature.

Fixes #3985


hoefling:
You can invoke the new test with
```
pytest -v -n0 mypy/test/teststubgen.py::StubgenPythonSuite::testVarArgsWithKwVarArgs
```

gvanrossum:
Thanks!

Issue #3526, opened 2017-06-12
Assignment of Any does not get rid of Optional (add `AnyUnion` type?)


ddfisher:
Consider this code (run with `--strict-optional`):
```
from typing import Optional
def f():
  # untyped function; often from silent import
  return 0  # always returns an int

def g(x: Optional[int]) -> int
  if x is None:
    x = f()
  return x  # E: Incompatible return value type (got "Optional[int]", expected "int")
```
Mypy doesn't infer that `x` is not `None` in `g`.  I think this a bug -- mypy 
should be permissive here.

If we wanted to fix this, we'd run into a problem: what type should `x` have 
after `x = f()`?  We don't want it to be `Any`, because we don't want that 
behavior in the general case.  It shouldn't be `int` -- `f()` could be intended 
to unconditionally return `None` instead.  To fix this, I think we need to 
introduce a new special type, which I will call `AnyUnion`.  `AnyUnion[A, 
B, C]` would represent a type which is allowed to be used as any of `A`, `B`, or 
`C` (similarly to how `Any` is allowed to be used as any type).  In this case, 
we could have `x`'s type bound to `AnyUnion[int, None]`, which would fix this 
error.

Other helpful uses of `AnyUnion`:
- `pow` will no longer need to return `Any` in builtins.  Instead, it could 
	return the much less broad `AnyUnion[int, float]`.
- `AnyUnion[X, None]` would provide a per-variable opt-out of strict None 
	checking.  This could be helpful for classes with delayed initialization where 
	a lot of effort would be needed to refactor them into a state that would work 
	properly with strict None checking.

Cons:
- Additional complexity of the type system affects implementation complexity and 
	makes things slightly harder for users to understand.
- The motivating problem also exists with subclasses (with the way the binder 
	currently works), but `AnyUnion` doesn't help much with that.

ilevkivskyi:
> ``pow`` will no longer need to return ``Any`` in builtins. Instead, it could
return the much less broad ``AnyUnion[int, float]``.

But doesn't #3501 (or similar) provides a better solution?

> ``AnyUnion[X, None]`` would provide a per-variable opt-out of strict None
checking. This could be helpful for classes with delayed initialization where
a lot of effort would be needed to refactor them into a state that would work
properly with strict None checking.

Do I understand correctly that this has been solved by PEP 526?

[...]

PR #4523, opened 2018-01-30
Allow comparisons to refine Optional types without strict-optional

### LINKED ISSUE #4520: explicit `x is None` should narrow union candidates like `isinstance(x, type(None))`
```
from typing import Optional, List  # noqa


def union_test1(x):
    # type: (Optional[List[int]]) -> Optional[int]
    if x is None:
        return x
    else:
        return len(x)


def union_test2(x):
    # type: (Optional[List[int]]) -> Optional[int]
    if isinstance(x, type(None)):
        return x
    else:
        return len(x)
```

`union_test1` with default options gives the error `Incompatible return value type (got "Optional[List[int]]", expected "Optional[int]")` (with `--strict-optional`, it does not give this error). `union_test2` typechecks just fine. This strikes me as highly counterintuitive; even though (as the docs say) `None` is a valid value for each type, if one of the above checks allows the typechecker to learn that `x` is `None` and *therefore* not a list, it's hard to see why the other doesn't.

Failing that, I would suggest at least a documentation change to point out both the gotcha embodied in `union_test1` and the workaround in `union_test2`.

msullivan:
Fixes #4520.

### DIFF
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -267,7 +267,7 @@ class ConditionalTypeBinder:
             return
 
         enclosing_type = self.most_recent_enclosing_type(expr, type)
-        if (isinstance(enclosing_type, AnyType) and not restrict_any):
+        if isinstance(enclosing_type, AnyType) and not restrict_any:
             # If x is Any and y is int, after x = y we do not infer that x is int.
             # This could be changed.
             # Instead, since we narrowed type from Any in a recent frame (probably an
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -267,8 +267,7 @@ class ConditionalTypeBinder:
             return
 
         enclosing_type = self.most_recent_enclosing_type(expr, type)
-        if (isinstance(enclosing_type, AnyType)
-                and not restrict_any):
+        if (isinstance(enclosing_type, AnyType) and not restrict_any):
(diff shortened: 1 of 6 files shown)


gvanrossum:
(Let's hold off on this a bit, we've discovered some problems with it in our own codebase.)

ilevkivskyi:
> (Let's hold off on this a bit, we've discovered some problems with it in our own codebase.)

Please let me know when the problems are solved, so that I can proceed with a review.

gvanrossum:
IIUC that's in @msullivan's court.

msullivan:
Before this can land, we need to fix #3526, since this change exposes a bunch of instances of it.

msullivan:
@ilevkivskyi this is ready for review now

mypy/binder.py
```diff
@@ -271,11 +271,10 @@ def assign_type(self, expr: Expression,
                 and not restrict_any):
```
ilevkivskyi: This doesn't need to be on a separate line.

msullivan:
This causes a small handful of errors in internal codebases that will be easy to fix. A couple were weird second-order effects in which doing the refinement meant that a type was no longer a Union and so behaved differently when computing types in a ternary expression. Sigh.

PR #4525, opened 2018-01-31
Support type aliases in fine-grained incremental mode

### LINKED ISSUE #4394: Support type aliases in fine-grained incremental mode
This includes generic type aliases such as `X = Dict[str, T]` and aliases to all kinds of types.

Generating dependencies may be tricky.

ilevkivskyi:
Fixes #4394 

This was a bit more tricky than I expected. I think I cover all common cases and majority of corner cases:
* Simple aliases
* Generic aliases
* Forward references
* Nested aliases
* Chained aliases

Note that among tests I added some also pass without this addition (probably because some dependencies are added by coincidence). Still there are 17 test scenarios that would fail without this change. Note that I mainly focus on false negatives, since in my experience while playing with fine-grained daemon, this is the most typical problem.

There are two weird failures that I skip, but those seem to be independent of type aliases (I checked manually that alias dependence maps are correct). The problem goes away if I just "touch" the `aa.py` module.


test-data/unit/fine-grained.test
```diff
@@ -1776,3 +1776,242 @@ def foo(x: Point) -> int:
...
+import a
+B = Dict[str, a.A]
+[file b.py]
+import aa
+x: aa.B = {'first': {str(): int()}}
+[builtins fixtures/dict.pyi]
+[out]
+==
+b.py:2: error: Dict entry 0 has incompatible type "str": "int"; expected "str": "str"
+
```
JukkaL: Ideas for additional test cases:

* Alias changed into a generic alias (and vice versa)
* Changing the number of type args of a generic alias (e.g. `C[T]` -> `C[T, S]`)
* Deleting the target of a type alias or a component of the type alias target

mypy/semanal.py
```diff
@@ -1674,12 +1679,32 @@ def anal_type(self, t: Type, *,
...
+                return tp
+            self.add_alias_deps(a.aliases_used)
+        return tp
+
+    def add_alias_deps(self, aliases_used: Set[str]) -> None:
+        """Add full names of type aliases on which the current node depends.
+
+        This is used by fine-grained incremental mode to re-chek the corresponding nodes.
+        """
+        if self.is_class_scope():
```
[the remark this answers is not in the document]
ilevkivskyi: There is no `cur_node` in pass two (only in pass three). Instead, in pass two there are two separate scope stacks: function stack and class stack. Maybe this can be simplified, but I think it would be a large refactoring for relatively little benefits.

test-data/unit/deps-types.test
```diff
@@ -151,3 +151,150 @@ def h() -> None:
 <m.f> -> m, m.h
 <m.ff> -> m, m.h
 <m.g> -> m.f
+
+[case testAliasDepsNormalMod]
+A = int
```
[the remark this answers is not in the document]
ilevkivskyi: OK, I forgot about special nature of `builtins` and `typing`. I will change most of the tests to user defined classes.

test-data/unit/deps-types.test
```diff
@@ -151,3 +151,150 @@ def h() -> None:
...
+from typing import Dict
+A = Dict[str, int]
+B = Dict[str, A]
+def f(x: B) -> None:
+    pass
+[builtins fixtures/dict.pyi]
+[out]
+<m.A> -> m, m.f
+<m.B> -> m, m.f
+
```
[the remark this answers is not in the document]
ilevkivskyi: `TypeVar`s are currently supported only partially (independently of aliases), and `TypedDict` has only basic support. I will therefore add only few basic tests with them. For others yes, I will add more tests.

ilevkivskyi:
> Also, should there be some AST diff test cases for type aliases as well? What about AST merge test cases?

I thought that these are less important, since aliases should be completely expanded after semantic analysis, I think I will add few tests for these.

ilevkivskyi:
@JukkaL Thanks for a detailed review! I will try to finish this today.

ilevkivskyi:
Hm, strange, all tests passed locally, but here in both Travis and AppVeyor errors are sorted in different order in one test. Should we somehow fix the order they are shown? (To make tests more stable.)

ilevkivskyi:
Sorry, I give up, this is beyond my understanding, all tests fail consistently. Independently of what is the order, the messages are shown in the opposite order.

mypy/typeanal.py
```diff
@@ -63,8 +64,9 @@ def analyze_type_alias(node: Expression,
                        allow_unnormalized: bool = False,
                        in_dynamic_func: bool = False,
                        global_scope: bool = True,
-                       warn_bound_tvar: bool = False) -> Optional[Type]:
-    """Return type if node is valid as a type alias rvalue.
+                       warn_bound_tvar: bool = False) -> Optional[Tuple[Type, Set[str]]]:
+    """If `node` is valid as a type alias rvalue, return the resulting type and a set of
```
JukkaL: Style nit: first line of a multi-line docstring should be one line, followed by an empty line.

mypy/nodes.py
```diff
@@ -194,6 +201,8 @@ class MypyFile(SymbolNode):
     path = ''
     # Top-level definitions and statements
     defs = None  # type: List[Statement]
+    # Type alias dependencies as mapping from node to set of alias full names
+    alias_deps = None  # type: DefaultDict[DepNode, Set[str]]
```
JukkaL: These references should be merged after semantic analysis, or otherwise they can point to stale nodes.

Actually, is it necessary to have AST node references here? It looks like we could use the full name of the enclosing fine-grained target as the key instead, and this wouldn't have to be merged, and it would generally better match how things work in general). You could perhaps use `mypy.server.deps.Scope` to track the current target (it should be moved to a different module first since we don't want semantic analyzer to depend on anything in mypy.server).

mypy/server/update.py
```diff
@@ -677,7 +677,10 @@ def calculate_active_triggers(manager: BuildManager,
             names.add(id)
         else:
             snapshot2 = snapshot_symbol_table(id, new.names)
-        names |= compare_symbol_table_snapshots(id, snapshot1, snapshot2)
+        diff = compare_symbol_table_snapshots(id, snapshot1, snapshot2)
+        names |= diff
+        if diff:
+            names.add(id)
```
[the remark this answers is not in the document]
ilevkivskyi: > the problem looks to be somewhere else

Indeed, I add it for `FuncItem` and `ClassDef`, but not for module target, will fix.

mypy/nodes.py
```diff
@@ -2334,6 +2341,11 @@ class SymbolTableNode:
     normalized = False  # type: bool
     # Was this defined by assignment to self attribute?
     implicit = False  # type: bool
+    # Is this node refers to other node via node aliasing?
+    # (This is currently used for simple aliases like `A = int` instead of .type_override)
+    is_aliasing = False  # type: bool
+    # This includes full names of aliases used in this alias.
+    alias_depends_on = None  # type: Set[str]
```
JukkaL: I wonder if this actually is necessary. Even without this, the target that contains the alias definition should depend on these aliases, right? And change in the alias should be detected by `mypy.server.astdiff`. If you remove this, what will break? Maybe some symbol table differences won't then be propagated correctly -- but that could just be a bug elsewhere, which this papers over.

mypy/server/aststrip.py
```diff
@@ -217,6 +217,9 @@ def is_duplicate_attribute_def(self, node: MemberExpr) -> bool:
         return any(info.get(node.name) is not None for info in self.type.mro[1:])
 
     def strip_ref_expr(self, node: RefExpr) -> None:
+        if node.fullname and (node.fullname.startswith('typing') or
+                node.fullname.startswith('builtins')):
+            return
```
JukkaL: This doesn't seem right. It's possible that something that used to refer to a builtin will be updated to point to a user-defined definition after a refresh.

mypy/server/deps.py
```diff
@@ -264,6 +283,18 @@ def visit_assignment_stmt(self, o: AssignmentStmt) -> None:
             assert info.typeddict_type is not None
             prefix = '%s.%s' % (self.scope.current_full_target(), info.name())
             self.add_type_dependencies(info.typeddict_type, target=make_trigger(prefix))
+        elif isinstance(rvalue, IndexExpr) and isinstance(rvalue.analyzed, TypeAliasExpr):
+            assert len(o.lvalues) == 1
+            lvalue = o.lvalues[0]
+            assert isinstance(lvalue, NameExpr)
+            attr_trigger = make_trigger('%s.%s' % (self.scope.current_full_target(),
```
JukkaL: It would be better to avoid constructing triggers by concatenating strings, since there are edge cases where this is not correct (such as for things nested within functions). It's better to use the full name instead. Even if this is not always correct yet, at least we can hopefully fix it in one place (where we initialize the full name).

mypy/server/deps.py
```diff
@@ -264,6 +283,18 @@ def visit_assignment_stmt(self, o: AssignmentStmt) -> None:
...
+                                                   lvalue.name))
+            # We need to re-process the target where alias is defined to "refresh" the alias.
+            self.add_type_dependencies(rvalue.analyzed.type)
+            self.add_type_dependencies(rvalue.analyzed.type, attr_trigger)
+            if attr_trigger in self.alias_deps:
+                for dep in self.alias_deps[attr_trigger]:
```
JukkaL: I don't understand this. Can you give an example where this is needed? It seems like ast diff should pick up changes to aliases.

mypy/server/deps.py
```diff
@@ -506,6 +537,13 @@ def visit_yield_from_expr(self, e: YieldFromExpr) -> None:
 
     # Helpers
 
+    def add_type_alias_deps(self, target: str, add_trigger: bool = False) -> None:
+        if target in self.alias_deps:
```
JukkaL: Add comment discussing why type aliases are special cased here.

mypy/server/deps.py
```diff
@@ -506,6 +537,13 @@ def visit_yield_from_expr(self, e: YieldFromExpr) -> None:
 
     # Helpers
 
+    def add_type_alias_deps(self, target: str, add_trigger: bool = False) -> None:
```
JukkaL: The `add_trigger` argument name is unclear. Can you come up with another name? Maybe something about triggering recursively?

mypy/server/deps.py
```diff
@@ -172,11 +189,13 @@ def visit_func_def(self, o: FuncDef) -> None:
             else:
                 signature = o.type
             for trigger in get_type_triggers(signature):
+                # TODO: avoid adding both `mod.func` and `<mod.func>` here and below for aliases
                 self.add_dependency(trigger)
                 self.add_dependency(trigger, target=make_trigger(target))
         if o.info:
             for base in non_trivial_bases(o.info):
                 self.add_dependency(make_trigger(base.fullname() + '.' + o.name()))
+        self.add_type_alias_deps(self.scope.current_full_target(), add_trigger=True)
```
JukkaL: Add comment about why we have `add_trigger=True`. Again, I believe that this should be `current_target()` instead of `current_full_target()`.

mypy/server/deps.py
```diff
@@ -153,13 +170,13 @@ def __init__(self,
...
     #   protocols
     #   metaclasses
-    #   type aliases
     #   functional enum
     #   type variable with value restriction
 
     def visit_mypy_file(self, o: MypyFile) -> None:
         self.scope.enter_file(o.fullname())
         self.is_package_init_file = o.is_package_init_file()
+        self.add_type_alias_deps(self.scope.current_full_target())
```
JukkaL: This should be `self.scope.current_target()`? `current_full_target()` can refer to an entire class (all method bodies) which we want to avoid if possible, since it's expensive.

mypy/server/deps.py
```diff
@@ -202,6 +221,7 @@ def visit_class_def(self, o: ClassDef) -> None:
             self.add_type_dependencies(o.info.typeddict_type, target=make_trigger(target))
         # TODO: Add dependencies based on remaining TypeInfo attributes.
         super().visit_class_def(o)
+        self.add_type_alias_deps(self.scope.current_full_target())
```
JukkaL: Similar to above, this should probably be `current_target()`.

mypy/typeanal.py
```diff
@@ -171,6 +175,11 @@ def __init__(self,
         self.is_typeshed_stub = is_typeshed_stub
         self.warn_bound_tvar = warn_bound_tvar
         self.third_pass = third_pass
+        # Names of type aliases encountered while analysing a type will be collected here.
+        # (This also recursively includes the names of aliases they depend on.)
+        # TODO: adding the sub-dependencies shouldn't be necessary since the logic
+        # of fine-grained should propagate triggered dependencies.
```
JukkaL: Why is this necessary? It would be useful to understand the precise root cause for why this is included and mention it in the TODO comment.

mypy/server/aststrip.py
ilevkivskyi: This needs to be kept in sync since there are other places where `builtins` and `typing` are special-cased. If I remove this, one test will fail. Namely, if there is an alias like `A = Dict[str, int]`, then it is not reprocessed correctly.

mypy/nodes.py
ilevkivskyi: Ideally this shouldn't be necessary. But IIRC correctly some tests failed without this. I added a TODO about this in `typeanal.py` (that you also commented). I will try to remove this and re-run the tests, and if some will still fail, I will try to figure out where exactly is the problem in diff calculation and expand the TODO.

mypy/server/deps.py
ilevkivskyi: This is to generate dependencies for the alias name itself, for example this code
```python
A = B[C, D]
```
will generated the following dependencies in `semanal`:
```
python
alias_deps['m'] = {'B', 'C', 'D'}
alias_deps['<A>'] = {'B', 'C', 'D'}
```
The line in question adds the latter.

Again, maybe this is indeed not needed, I will double check now.

ilevkivskyi:
@JukkaL OK, I got rid of the "long-range" (i.e. indirect, recursive) dependencies for type aliases. There are few minor remaining questions we can discuss tomorrow (also I will add tests you asked). Here is a summary of how everything works:
* Definition of a type alias creates a dependency for the enclosing scope, so that the alias will be reprocessed if it is changed. So `A = int` generates `<A> -> __main__`.
* Use of a type alias creates a dependency for the using target on the alias. So `def func(x: A): ...` generates `<A> -> func` (for _functions_ this also creates `<A> -> <func>`, but this is not something special for aliases, the same happens for other types because functions are not correctly treated in `get_triggered_namespace_items`)
* Finally, type aliases depend on their expansion, aliases directly used, and type variables (for generic aliases). So, `B = Dict[A, Cls]` generates `<A> -> <B>, <Cls> -> <B>`

This is all, no other dependencies are generated, for "chained" aliases etc. indirect dependencies are re-triggered and re-processed in several passes.

There is one thing that concerns me: currently subscripted and un-subscripted aliases are processed differently so this also required some duplication of logic (this also happens in several other places). As we discussed, there will be refactoring of symbol table nodes, so there are two options:
1. First refactor symbol table nodes, then rebase this on top which will avoid duplication of logic.
2. First merge this, then remove duplication of logic here as part of the refactoring.

I prefer the second option, since all this picture of dependencies is now in my head and it will take time to rebuild it if I will switch to another thing.

JukkaL:
> Finally, type aliases depend on their expansion, aliases directly used, and type variables (for generic aliases). So, `B = Dict[A, Cls]` generates `<A> -> <B>`, `<Cls> -> <B>`

What about generating dependencies like `<A> -> m`, `<Cls> -> m` instead, if the alias is defined in `m`? This way we'd get rid of one processing step:

1. `A` is changed and it triggers `<A>`.
2. We reprocess `m`, the module containing the type alias `B`.
3. AST diff of `m` notices that the alias `m.B` has changed, so we trigger `<m.B>`.
4. We reprocess each target that uses `B` in an annotation.

Also, do you test that changing `A.__init__` causes calls to `B` to be reprocessed?

> I prefer the second option, since all this picture of dependencies is now in my head and it will take time to rebuild it if I will switch to another thing.

Yeah, let's go with the second option (at least unless the symbol table refactoring turns out to be a simple thing).

ilevkivskyi:
> What about generating dependencies like `<A> -> m`, `<Cls> -> m` instead, if the alias is defined in `m`?

I think the performance win will be pretty minor, since currently we already have `<A> -> <B>` and `<B> -> m`. So I would prefer not to do this now (I think correctness/simplicity is more important now), unless you think it is something important.

> Also, do you test that changing `A.__init__` causes calls to B to be reprocessed?

Yes, there are some tests still missing, I have a TODO item for this one too.

JukkaL:
> I think the performance win will be pretty minor, since currently we already have `<A> -> <B>` and `<B> -> m`. So I would prefer not to do this now (I think correctness/simplicity is more important now), unless you think it is something important.

The performance win isn't really the only motivation for preferring minimal dependencies. I think that this is important due to a few reasons:

* If we have non-minimal/optimal dependencies, it's harder to reason about bugs and correct behavior -- logs will have redundant steps, etc. This can result in developers tending to ignore suspicious-looking log data, since it's easy to just assume that it's because of known issue X, even if the logs are actually showing a previously unknown issue.
* If we have non-minimal/optimal dependencies, bugs elsewhere in fine-grained mode may get unnoticed because of the redundant work.
* It looks like we can reprocess everything that uses an alias `B` in an annotation twice: first after `A` changes, and second time after we've reprocessed `m`.  The performance impact could be pretty bad in some cases.
* Even if the performance impact of this is small, it might interact badly with other performance problems (say, if you have three issues that cause 2x extra work, you could suddenly have 8x extra work if they all happen at the same time).

ilevkivskyi:
This is almost ready. I still need to figure out what best to do with `builtins` and `typing` and add tests for `.__init__` deps.

test-data/unit/diff.test
```diff
@@ -618,3 +618,53 @@ p = Point(dict(x=42, y=1337))
 [out]
 __main__.Point
 __main__.p
+
+[case testTypeAliasSimple]
+A = int
```
JukkaL: In all of these test cases, also include a case where there is no difference to make sure that we don't just always consider type aliases different. Example:

```
[case testTypeAliasSimple]
A = int
B = int
[file next.py]
A = str
B = int
[out]
__main__.A
```

test-data/unit/fine-grained.test
```diff
@@ -1797,6 +1797,372 @@ def foo(x: Point) -> int:
...
+[file b.py.2]
+def f(x: A):
+    x = int()
+A = str
+[out]
+==
+b.py:2: error: Incompatible types in assignment (expression has type "int", variable has type "str")
+
+[case testAliasFineChainedFunc]
+# This and next don't work without <A> -> <B> intermediate deps
```
JukkaL: Is this comment out of date?

test-data/unit/fine-grained.test
```diff
@@ -1797,6 +1797,372 @@ def foo(x: Point) -> int:
...
+[file a.py.2]
+A = str
+[file b.py]
+import a
+def f() -> None:
+    class C:
+        x: a.A = int()
+[out]
+==
+b.py:4: error: Incompatible types in assignment (expression has type "int", variable has type "str")
```
JukkaL: Potential new tests (if they don't already exist):

* Create a type alias (initially there is a stale reference to a type alias).
* Delete type alias (after update there is a stale reference to a type alias).
* Switch a class to a type alias (or type alias to class).

test-data/unit/deps-types.test
```diff
@@ -151,3 +151,411 @@ def h() -> None:
...
+[file a.py]
+from mod import I
+A = I
+[file mod.py]
+class I: pass
+[out]
+<m.x> -> m
+<a.A> -> m
+<a> -> m
+<mod.I> -> <m.x>, m
```
JukkaL: Can you also dump dependencies in `a`? I'd like to see the dependency `<mod.I> -> a` at least.

test-data/unit/deps-types.test
```diff
@@ -151,3 +151,411 @@ def h() -> None:
...
+[case testAliasDepsNormalFuncExtended]
+import a
+def f(x: a.A) -> None:
+    pass
+[file a.py]
+from mod import I
+A = I
+[file mod.py]
+class I: pass
+[out]
```
JukkaL: Again would also be good to see dependencies in `a`.

test-data/unit/deps-types.test
```diff
@@ -151,3 +151,411 @@ def h() -> None:
...
+<m.A> -> m
+<m.B> -> m
+<m.C> -> m.C
+<mod.I.__init__> -> <m.C.__init__>
+<mod.I> -> m, m.C
+
+[case testAliasDepsNestedMod]
+from mod import I, S, D
+A = D[S, I]
+B = D[S, A]
```
JukkaL: This test case would be more interesting if `A` and `B` would be defined in different modules, and if we dumped the dependencies within both modules.

test-data/unit/deps-types.test
```diff
@@ -151,3 +151,411 @@ def h() -> None:
...
+<m.B> -> m
+<m.x> -> m
+<mod.D> -> <m.x>, m
+<mod.I> -> <m.x>, m
+<mod.S> -> <m.x>, m
+
+[case testAliasDepsNestedFunc]
+from mod import I, S, D
+A = D[S, I]
+B = D[S, A]
```
JukkaL: This is similar to above -- would be better to split across two modules.

test-data/unit/deps-types.test
```diff
@@ -151,3 +151,411 @@ def h() -> None:
...
+class I: pass
+class S: pass
+[out]
+<m.A> -> m.fun
+<mod.I> -> m, m.fun
+
+[case testAliasDepsRuntime]
+from mod import I, S, D
+A = I
+x = D[S, A]()
```
JukkaL: This would be more interesting if `A` would be defined in another module.

test-data/unit/deps-types.test
```diff
@@ -151,3 +151,411 @@ def h() -> None:
...
+class P(NamedTuple):
+    x: A
+[file mod.py]
+class I: pass
+[out]
+<m.A> -> m
+<m.P> -> m.P
+<mod.I> -> <m.P.x>, <m.P>, m, m.P
+
+[case testAliasDepsTypedDict]
```
JukkaL: Also test functional definition of typed dict.

test-data/unit/deps-types.test
```diff
@@ -151,3 +151,411 @@ def h() -> None:
...
+class S: pass
+[out]
+<m.A> -> m
+<m.x> -> m
+<mod.D.__init__> -> m
+<mod.D> -> <m.x>, m
+<mod.I> -> <m.x>, m
+<mod.S> -> <m.x>, m
+
+[case testAliasDepsNamedTuple]
```
JukkaL: Also test functional definition of named tuple.

test-data/unit/deps-types.test
```diff
@@ -566,6 +665,20 @@ class I: pass
...
+[case testAliasDepsTypedDictFunctional]
+from mypy_extensions import TypedDict
+from mod import I
+A = I
+P = TypedDict('P', {'x': A})
+[file mod.py]
+class I: pass
+[builtins fixtures/dict.pyi]
+[out]
+<m.A> -> m
```
JukkaL: Like above, does this actually test that the dependency is generated, or would the dependency generate for just `A = I`, without the typed dict definition?

test-data/unit/deps-types.test
```diff
@@ -550,6 +637,18 @@ class I: pass
...
 
+[case testAliasDepsNamedTupleFunctional]
+from typing import NamedTuple
+from mod import I
+A = I
+P = NamedTuple('P', [('x', A)])
+[file mod.py]
+class I: pass
+[out]
+<m.A> -> m
```
JukkaL: Does this actually test that the dependency is generated, or would the dependency generate for just `A = I`, without the named tuple definition?

PR #4526, opened 2018-01-31
Enable generation and caching of fine-grained dependencies from normal runs

msullivan:


mypy/build.py
```diff
@@ -1977,6 +1986,19 @@ def _patch_indirect_dependencies(self,
             elif dep not in self.suppressed and dep in self.manager.missing_modules:
                 self.suppressed.append(dep)
 
+    def find_fine_grained_deps(self) -> None:
```
gvanrossum: Would you mind renaming this so the name more clearly implies that it has a side-effect?

mypy/build.py
```diff
@@ -1183,6 +1184,7 @@ def compute_hash(text: str) -> str:
 
 
 def write_cache(id: str, path: str, tree: MypyFile,
+                fine_grained_deps: Dict[str, List[str]],
```
gvanrossum: I'd rename this to `serialized_fine_grained_deps` just to clarify that it does not have the same type as `State.fine_grained_deps` (List vs. Set).

Issue #1852, opened 2016-07-13
Importing unstubbed submodules of stubbed modules with silent imports gives inconsistent errors


ddfisher:
In this example, the `boto` package has a stub, but the `boto.foo` module does not:
```bash
$ python3 -m mypy --py2 --silent-imports -c 'import boto.foo'
# no error
$ python3 -m mypy --py2 --silent-imports -c 'from boto import foo'
<string>:1: error: Module has no attribute 'foo'
$ python3 -m mypy --py2 --silent-imports -c 'import boto.foo; boto.foo'
<string>:1: error: "module" has no attribute "foo"
# in real code, this error is on line where boto.foo is used
```

As you can see, the behavior here is inconsistent.  `import boto.foo` gives no error, but attempting to use `boto.foo` later in the program does give an error.

I think the correct behavior here is for `import x.y` with `--silent-imports` to be an error when `x` has a stub but `y` does not, because this feels conceptually similar to other cases of incomplete stubs.  Otherwise, you'd be in danger of importing nonexistent submodules from packages you have full stubs for, with no way to lock that down.  The only downside to this approach is it means you have to use a type ignore when the stubs are incomplete, but that's already the case most of the time.

gvanrossum:
There would still be an inconsistency: if instead of `boto.foo` we were talking about `boto_foo`, then `import boto_foo; boto_foo` would not generate any errors (and `boto_foo.bar` would not either, since `boto_foo` would have type `Any`).

Back to `boto.foo`, there might also be cases where `boto/foo.py` exists (though not listed on the command line) but there's no `boto/foo.pyi` -- then the intention of `--silent-imports` would be to suppress errors about any mention of `boto.foo`.

So there is some argument to be made that the problem here is actually that, even though the error on `import boto.foo` is suppressed, you still get an error about `boto.foo`. I suspect that error happens because while the failing import introduce a "tombstone" for `boto.foo`, it doesn't do anything about the `foo` variable in the `boto` namespace.

Note that `--almost-silent` does point out the error in the import itself.

JukkaL:
It seems to me that if `import boto.foo` (and other similar examples) would produce errors when using `--silent-imports`, we should try to enforce in typeshed that all stubs for packages include stubs for all submodules. Otherwise the behavior of `--silent-imports` would feel inconsistent. This is generally difficult to do as new submodules can get added in new package releases, but on the other hand we already have this problem with new functions and other things added to existing module stubs, so we could perhaps live with that.

[...]

PR #4541, opened 2018-02-04
Rename variable, OSError.filename will have union type (typeshed #1852)

intgr:
See https://github.com/python/typeshed/pull/1852, I want to change the type of `OSError.filename` to `Union[str, bytes, None]`, but since the `path` variable is inferred to have `str` type, this causes a Mypy error.

On the other hand, the `path = error.filename` assignment seems of questionable value anyway, since `path` already contained the path. But I may be missing something so I erred on the side of caution of not (potentially) changing the behavior.

### DIFF
--- a/mypy/test/data.py
+++ b/mypy/test/data.py
@@ -315,11 +315,11 @@ class DataDrivenTestCase(BaseTestCase):
                     # a bug in a test case -- we shouldn't leave
                     # garbage lying around. By nuking the directory,
                     # the next test run hopefully passes.
-                    path = error.filename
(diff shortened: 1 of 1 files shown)


gvanrossum:
I don't think this is the right thing to do. Unions are terrible from a gradual typing perspective. Your best bet may be to change the attribute in typeshed to `Any`, and then you won't need this PR.

intgr:
Merged in Typeshed; this is now unnecessary, closing. https://github.com/python/typeshed/pull/1852

PR #4544, opened 2018-02-05
Type checking of class decorators

euresti:
It checks that the calls are well-type and that applying the decorators works.
It does not check the end result because what should I do with that?

Helps with #3135

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1254,6 +1254,11 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                 # Otherwise we've already found errors; more errors are not useful
                 self.check_multiple_inheritance(typ)
 
+            for decorator in defn.decorators:
+                # Currently this only checks that the decorator itself is well typed.
+                # TODO: Check that applying the decorator to the class would do the right thing.
+                self.expr_checker.accept(decorator)
+
     def check_protocol_variance(self, defn: ClassDef) -> None:
         """Check that protocol definition is compatible with declared
         variances of type variables.
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -40,7 +40,7 @@ from mypy.types import (
 from mypy.sametypes import is_same_type, is_same_types
 from mypy.messages import MessageBuilder, make_inferred_type_note
 import mypy.checkexpr
-from mypy.checkmember import map_type_from_supertype, bind_self, erase_to_bound
+from mypy.checkmember import map_type_from_supertype, bind_self, erase_to_bound, type_object_type
 from mypy import messages
 from mypy.subtypes import (
     is_subtype, is_equivalent, is_proper_subtype, is_more_precise,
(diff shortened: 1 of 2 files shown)


test-data/unit/check-classes.test
```diff
@@ -4223,3 +4223,12 @@ class C(Any):
...
 [out]
+
+[case testClassDecoratorIsTypeChecked]
+from typing import Callable
+def decorate(x: int) -> Callable[[type], type]:  # N: "decorate" defined here
+    ...
+
+@decorate(y=17)  # E: Unexpected keyword argument "y" for "decorate"
+class A:
+    pass
```
ilevkivskyi: The PR is trivial, but I would add few more test cases for various kinds of errors, like non-existing name/attribute, calling non-callable, too many args, forward reference in a decorator type, etc.

test-data/unit/check-classes.test
```diff
@@ -4115,7 +4115,7 @@ def f() -> type: return M
 class C1(six.with_metaclass(M), object): pass  # E: Invalid base class
 class C2(C1, six.with_metaclass(M)): pass  # E: Invalid base class
 class C3(six.with_metaclass(A)): pass  # E: Metaclasses not inheriting from 'type' are not supported
-@six.add_metaclass(A)  # E: Metaclasses not inheriting from 'type' are not supported
+@six.add_metaclass(A)  # E: Metaclasses not inheriting from 'type' are not supported  # E: Argument 1 to "add_metaclass" has incompatible type "Type[A]"; expected "Type[type]"
```
ilevkivskyi: One more thing: Add `\` somewhere here and format it nicely with a newline, like this:
```
failing_func(wrong) # E: Bad function \
                    # E: Bad argument
```

mypy/checker.py
```diff
@@ -1254,6 +1254,24 @@ def visit_class_def(self, defn: ClassDef) -> None:
                 # Otherwise we've already found errors; more errors are not useful
                 self.check_multiple_inheritance(typ)
 
+            if defn.decorators:
+                sig = type_object_type(defn.info, self.named_type)
+                # Decorators are applied in reverse order.
+                for decorator in reversed(defn.decorators):
+                    if (isinstance(decorator, CallExpr)
+                            and isinstance(decorator.analyzed, PromoteExpr)):
```
ilevkivskyi: This exception deserves a comment.

mypy/checker.py
```diff
@@ -1254,6 +1254,24 @@ def visit_class_def(self, defn: ClassDef) -> None:
...
+                        continue
+
+                    dec = self.expr_checker.accept(decorator)
+                    temp = self.temp_node(sig)
+                    fullname = None
+                    if isinstance(decorator, RefExpr):
+                        fullname = decorator.fullname
+
+                    sig, t2 = self.expr_checker.check_call(dec, [temp],
```
ilevkivskyi: Better use an underscore for throw-away variable.

mypy/checker.py
```diff
@@ -1254,6 +1254,24 @@ def visit_class_def(self, defn: ClassDef) -> None:
                 # Otherwise we've already found errors; more errors are not useful
                 self.check_multiple_inheritance(typ)
 
+            if defn.decorators:
```
ilevkivskyi: I would leave some kind of `TODO` here. One of main use cases:
```python
dec: Callable[[Type[A]], Type[B]]

@dec
class A:
    pass

def fun(arg: B) -> None:
    pass

fun(A())  # Should be OK
```
is still not supported.

test-data/unit/check-classes.test
```diff
@@ -4223,3 +4224,35 @@ class C(Any):
...
+@decorate  # E: Argument 1 to "decorate" has incompatible type "Type[A2]"; expected "int"
+class A2: pass
+
+[case testClassDecoratorIncorrect]
+def not_a_class_decorator(x: int) -> int: ...
+@not_a_class_decorator(7)  # E: "int" not callable
+class A3: pass
+
+not_a_function = 17
+@not_a_function()  # E: "int" not callable
```
ilevkivskyi: Also add a test for directly applying `not_a_function`:
```python
@not_a_function
class C: ...
```

test-data/unit/check-classes.test
```diff
@@ -4223,3 +4224,35 @@ class C(Any):
...
+@decorate_forward_ref()
+@decorate(11)
+class A: pass
+
+@decorate  # E: Argument 1 to "decorate" has incompatible type "Type[A2]"; expected "int"
+class A2: pass
+
+[case testClassDecoratorIncorrect]
+def not_a_class_decorator(x: int) -> int: ...
+@not_a_class_decorator(7)  # E: "int" not callable
```
ilevkivskyi: The error message is quite obscure. I would somehow highlight that a class decorator must be a function that accepts a type. If this is too hard, then leave a TODO about this in code.

mypy/checker.py
euresti: Added below

mypy/checker.py
euresti: Done!

test-data/unit/check-classes.test
euresti: Left a TODO. Only thing I can figure out is collect all the errors with `self.msg.copy()` then modify all the messages to say what we want.  Or check the count of errors before and after and add another one if it has increased.

test-data/unit/check-classes.test
euresti: Done

euresti:
Done!

PR #4545, opened 2018-02-06
Fix --warn-return-any for NotImplemented

### LINKED ISSUE #4534: --warn-return-any should not complain on NotImplemented when using abc
I'm using `abc.abstractmethod`:
```python
@abc.abstractmethod
def encode(self) -> bytes:
    return NotImplemented
```

And I get a warning: `warning: Returning Any from function declared to return "bytes"`

I don't think something is wrong with the code, it's totally normal way to declare and describe an interface and mypy shouldn't complain about it.

drtyrsa:
fixes #4534


.gitignore
```diff
@@ -14,6 +14,7 @@ docs/build/
 .cache
 .runtest_log.json
 dmypy.json
+.pytest_cache/
```
gvanrossum: You can add this to you own global .gitignore -- I don't see a reason to include this in mypy's .gitignore.

gvanrossum:
Thanks!

PR #4550, opened 2018-02-07
Have -v enable DEBUG logging from fine-grained updates.

msullivan:


gvanrossum:
Can I beg you to refactor this to do without the global variable?

There are only a few places where DEBUG is used in a function that doesn't already have access to a BuildManager.

msullivan:
Refactored to just use manager.log everywhere

Issue #4492, opened 2018-01-19
What to do about inferring types from multiple scopes?


JukkaL:
Typically mypy infers the type of the variable from the first assignment. In some cases mypy infers the type from two assignments. The first assignment creates internally a *partial type*, which is missing some information. The second assignment completes the type. Example:

```py
x = []  # x gets a partial List type
x.append(1)  # x gets complete type List[int]
```

In some cases the assignments can be in two different scopes. Example:

```py
x = []

def f() -> None:
    x.append(1)

# type of x is List[int]
```

This feature has apparently been supported a while, but it has some issues:

* This behavior is somewhat messy to support in fine-grained incremental mode (#4464). 
* It can make it hard to figure out the inferred type of a variable, since the two assignments can be separated from each other by hundreds of lines and many functions.
* Moving functions around can affect the type of a module/class variable in cases where there are two potential functions that can complete a partial type.
* If the second assignment is missing or removed, and the first assignment initializes the variable with `None`, the inferred type will become `None`, which is usually not what the user wanted.
* It seems a bit fragile with some issues that took a long time to notice (#4484, #4455, #4416).

There are a few alternatives which might be reasonable, instead of the current behavior:

1. Require an annotation for the variable in cases such as the above, where the assignment that completes a partial type is in a different scope. This would break compatibility with previous mypy releases. Based on a quick experiment, the S internal codebase at Dropbox would require about 40 additional type annotations if we make this change.
2. Ignore assignments in nested scopes and assume `Any` for any missing partial type information. For example, if the initial assignment is `x = []` and the only additional assignments are in nested scopes, infer `List[Any]` as the type of `x`. This way the type of `x` is defined entirely by the contents of a single scope, which solves the above problems. This also introduces an additional problem -- implicit `Any` types. We might want to have a strictness flag that falls back to requiring a type annotation (as in option 1 above).

Currently my biggest concern is fine-grained incremental mode. There it may be possible to support the current behavior by introducing a new concept: "bound targets". I'll add a writeup about this to #4464.

@gvanrossum @msullivan @ilevkivskyi Thoughts?

ilevkivskyi:
[...]

PR #4553, opened 2018-02-08
Fine-grained: Don't infer partial types from multiple targets

JukkaL:
Previously if partial types were inferred from assignments in multiple 
scopes, like in the example below, fine-grained incremental mode could 
display bogus messages, since only one of the scopes could be 
reprocessed in one incremental update:

```
x = None

def f() -> None:
    global x
    x = ''
```

This PR fixes the issue by always inferring partial types within a single
fine-grained incremental target. This is always enabled in fine-grained
incremental mode and can also be turned on in other modes through
a new (but hidden) `--local-partial-types` option.

This was complicated by various edge cases that covered by the new
test cases and/or documented in comments.

If the new option is not enabled, the old behavior should (mostly) be
preserved, module some fixes that may affect also the default mode.

Work towards #4492.


JukkaL:
This is ready for review but don't merge yet since I want to fix various issues in internal Dropbox codebases first.

mypy/checker.py
```diff
@@ -107,6 +107,10 @@
         ('is_upper_bound', bool),  # False => precise type
     ])
 
+# Keeps track of partial types in a single scope
+PartialTypeScope = NamedTuple('PartialTypeMap', [('map', Dict[Var, Context]),
```
ilevkivskyi: Shouldn't `NAME` in `NAME = NamedTuple('NAME', ...)` always be the same in l.h.s. and r.h.s.?

mypy/checker.py
```diff
@@ -107,6 +107,10 @@
...
+                                                 ('is_function', bool)])
```
ilevkivskyi: I think a short comment about why we need to keep track about `is_function` will help here. (My understanding this is because fine-grained targets are always top level functions.)

test-data/unit/check-inference.test
```diff
@@ -2001,3 +2002,262 @@ if bool():
...
+    x: str
+
+class B(A):
+    x = None  # E: Incompatible types in assignment (expression has type "None", base class "A" defined the type as "str")
+
+[case testLocalPartialTypesWithAnyBaseClass]
+# flags: --local-partial-types --strict-optional
+from typing import Any
+
+A: Any
```
ilevkivskyi: I would add the same test where `A` is not `Any`, but a normal class. (And `None` in _two_ subclasses like here.)

test-data/unit/fine-grained.test
```diff
@@ -1796,3 +1796,51 @@ def foo(x: Point) -> int:
...
+import a
+[file a.py]
+[file a.py.2]
+y = None
+def f() -> None:
+    global y
+    y = ''
+[out]
+==
+a.py:1: error: Need type annotation for 'y'
```
ilevkivskyi: Just for completes I would add a tests where the error appears in the first run, but not in the second one (like if a user listened and actually added an annotation).

mypy/checker.py
```diff
@@ -3088,33 +3092,99 @@ def lookup_qualified(self, name: str) -> SymbolTableNode:
...
                         self.msg.need_annotation_for_var(var, context)
                         self.partial_reported.add(var)
+                    # Give the variable an 'Any' type to avoid generating multiple errors
+                    # from a single missing annotation.
                     var.type = AnyType(TypeOfAny.from_error)
 
+    def is_defined_in_base_class(self, var: Var) -> bool:
+        if var.info is not None:
+            for base in var.info.mro[1:]:
+                if base.get(var.name()) is not None:
```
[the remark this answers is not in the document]
JukkaL: Sorry, I'm not sure what you mean by this. Can you give an example where this would make a difference?

test-data/unit/check-inference.test
JukkaL: Added a test case.

test-data/unit/fine-grained.test
JukkaL: Added a test case.

mypy/checker.py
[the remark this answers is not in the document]
ilevkivskyi: I mean add `... and not isinstance(base.get(var.name()), NoneTyp)`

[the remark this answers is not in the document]
JukkaL: `base.get()` returns `Optional[SymbolTableNode]` so that exact check doesn't make sense. If you meant guarding agains a definition like `x = None  # type: None` in a base class, I don't agree -- it should be treated as a valid definition. If you still don't agree, can you give me a test case where this would make a difference?

JukkaL:
The travis failure seems like a known flake.

mypy/checker.py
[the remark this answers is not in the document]
ilevkivskyi: Yes, sorry, you are right, this is perfectly valid.

PR #4557, opened 2018-02-09
Add back support for simplified fine-grained logging

JukkaL:
A recent change merged fine-grained and general logging,
resulting in pretty verbose logging for fine-grained mode.
This change adds back a development-time option to use
less verbose logging specific to the fine-grained
incremental mode.

### DIFF
--- a/mypy/server/update.py
+++ b/mypy/server/update.py
@@ -123,7 +123,7 @@ from typing import Dict, List, Set, Tuple, Iterable, Union, Optional, Mapping, N
 
 from mypy.build import (
     BuildManager, State, BuildSource, Graph, load_graph, SavedCache, CacheMeta,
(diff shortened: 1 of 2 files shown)


mypy/server/update.py
```diff
@@ -1008,3 +1008,7 @@ def extract_type_maps(graph: Graph) -> Dict[str, Dict[Expression, Type]]:
     # This is used to export information used only by the testmerge harness.
     return {id: state.type_map() for id, state in graph.items()
             if state.tree}
+
+
+def is_verbose(manager: BuildManager) -> bool:
+    return manager.options.verbosity > 1 or DEBUG_FINE_GRAINED
```
ilevkivskyi: I think this should use `>=` instead of `>` for this to be an identical transformation if `DEBUG_FINE_GRAINED` is false.

PR #4561, opened 2018-02-09
Sync typeshed

gvanrossum:


gvanrossum:
Hold on, this caused some new failures on an internal repo.

PR #4566, opened 2018-02-12
Make TypeIndirectionVisitor depend on all classes in MROs

### LINKED ISSUE #4562: False negative in incremental mode
We don't yet fully understood this, but @msullivan and I observed the following:

- the code was of the form `assert x.y.z == <constant>`
- the expected error was of the form `Item "None" of "Optional[Y]" has no attribute "z"`
- the type of `x.y` was inferred from a method `f() -> Optional[Y]`
- previously, `f` (defined in a different file) did not have an annotation
- the file with the error was not modified

Apparently the cache for the file with the error was considered up-to-date, and not dependent on the file where `f` was defined. It's possible that interface hash calculation is at fault.

msullivan:
This fixes some nasty incremental mode bugs.
Fixes #4562.

### DIFF
--- a/mypy/indirection.py
+++ b/mypy/indirection.py
@@ -69,7 +69,11 @@ class TypeIndirectionVisitor(SyntheticTypeVisitor[Set[str]]):
     def visit_instance(self, t: types.Instance) -> Set[str]:
         out = self._visit(*t.args)
         if t.type is not None:
-            out.update(split_module_names(t.type.module_name))
+            # Uses of a class depend on everything in the MRO,
+            # as changes to classes in the MRO can add types to methods,
+            # change property types, change the MRO itself, etc.
+            for s in t.type.mro:
+                out.update(split_module_names(s.module_name))
         return out
 
     def visit_callable_type(self, t: types.CallableType) -> Set[str]:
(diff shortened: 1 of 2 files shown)


gvanrossum:
Can you understand why the tests fail on all builds except 3.6?

gvanrossum:
I've restarted the build, but it doesn't *feel* like a glitch. (There have been some glitches related to Trusty.)

gvanrossum:
After re-running the failing build, it still fails, but now with 3 out of 4 jobs passing (previously only 3.6 passed). But 3.4 still fails, even after re-running that one. Flake? Or something else?

gvanrossum:
After re-running the failing build, it still fails, but now with 3 out of 4 jobs passing (previously only 3.6 passed). But 3.4 still fails, even after re-running that one. Flake? Or something else?

msullivan:
I am deeply confused by this.

msullivan:
I can intermittently reproduce the failure locally---it seems to genuinely be a nondeterministic failure. I am investigating.

msullivan:
The test failures were caused by `STRICT_OPTIONAL` leaking out and being true for unit tests. This would happen if a test run under dmypy had strict optional on (and I added a test that did). It was nondeterministic because it only came up if the same runner that ran that test then ran the typing unit tests.

#4568 is up to fix that. I'm rebasing this diff but it needs to not land until the other has.

Issue #4527, opened 2018-01-31
Let plugins customize class MRO


snarkmaster:
The Python metaclass data model allows the `mro()` method to be overloaded. This is a genuinely useful feature for dealing with metaclasses whose classes must support inheritance. See "Example" below.

I wrote a `mypy` plugin for type-checking instances of my metaclass. I was able to model its semantics almost correctly with the existing plugin hooks, except for one thing — customizing the MRO.

The best way I can get my type-checker unblocked is by adding a plugin hook for customizing the MRO. My idea is roughly this:
 - Move `calculate_class_mro` to be a member of `SemanticAnalyzerPass2`
 - Add a plugin hook that gets to customize the computed MRO.

@JukkaL, does this sound like a feature you would take?

Example of needing to customize the MRO:
 - My metaclass `M` injects a implementation-detail base `M_classname_base` into each class it creates. `M`'s semantics demand that `M_classname_base` be at the very end of the MRO (to allow overloading of its features).
 - Let us define `class A(metaclass=M)` and `class  B(A)`.
 - Each class injects its own base: `M_A_base` and `M_B_base`. That means that `B` actually has both implementation-detail bases.
 - Unfortunately, `M_A_base` and `M_B_base` conflict, and `M_B_base` ends up **after** `M_A_base` in the MRO, which is the opposite of what you want (child must overload parent, not vice versa). 
 - At runtime, `M.mro()` can simply remove the `A`-produced base from `B`'s MRO, and all is well.
 - In current `mypy`, removing `M_A_base` from `B`'s MRO seems impossible.

Cc: @carljm, if you're curious.

JukkaL:
At first this sounds like a pretty simple change, but there are some edge cases:
* What if modifying the MRO changes the abstract status of the class?
* What if modifying the MRO turns the class into a named tuple (or a tuple base class)? Or makes it a non-tuple?
* What if metaclass structure becomes inconsistent?

There probably are more things that could go wrong. Maybe you could go through all `TypeInfo` attributes and analyze which of them might be stale after an MRO change? Maybe we should reject certain MRO changes if they are hard to support?

It would also help if we can identify other projects that would benefit from this (at at least single well-known project).

snarkmaster:
Here's a proof-of-concept patch: https://github.com/snarkmaster/mypy/commit/63f4aa493c0d262bdfc831a0aed7dd0fb95f21a0

[...]

PR #4567, opened 2018-02-13
Add a `customize_class_mro` plugin hook

snarkmaster:
The rationale for this MRO hook is documented on https://github.com/python/mypy/issues/4527

This patch completely addresses my need for customizing the MRO of types that use my metaclass, and I believe it is simple & general enough for other plugin authors.

NB My initial version of this change avoided the "get_*_hook -> Callable" indirection, and simply allowed each plugin to customize the MRO. It seemed like @gvanrossum was not a fan of breaking with the local convention, so I made the API conventional and more complex. Here is the diff from the original version: https://gist.github.com/snarkmaster/85076c9793fd5639fc7212a4c04e710a

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -214,6 +214,10 @@ class Plugin:
                             ) -> Optional[Callable[[ClassDefContext], None]]:
         return None
 
+    def get_customize_class_mro_hook(self, fullname: str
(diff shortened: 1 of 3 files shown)


snarkmaster:
Resolved the merge conflict.

snarkmaster:
Thanks to all! Acknowledging that there are no guarantees, and that the API may change significantly.

PR #4568, opened 2018-02-13
Clean up handling of STRICT_OPTIONAL flag

msullivan:
STRICT_OPTIONAL is managed with strict_optional_set based on
options.strict_optional. It doesn't need to be set directly anymore,
so get rid of the code that does it.

In addition to the code cleanup, this fixes a bug where using strict
optional in a test case run by testdmypy would cause the strict
optional state to leak, causing nondeterministic test failures in unit
tests.


msullivan:
The flakiness was triggered by adding an incremental strict-optional test, which we previously didn't have. So it has been lurking but not surfacing.

PR #4578, opened 2018-02-16
Remove now-unnecessary type: ignore.

EvanKrall:
These are made unnecessary by https://github.com/python/typeshed/pull/1881


JelleZijlstra:
I merged the typeshed change; this can now be merged after we sync typeshed.

PR #4583, opened 2018-02-17
Fix crash in attrs_plugin when using a converter

euresti:
Found this issue running mypy on our codebase.  

Sadly I don't know how to fix the problem completely *and* support subclasses *and* incremental mode but I'll leave that for another day.

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -93,6 +93,10 @@ class SemanticAnalyzerPluginInterface:
     def lookup_fully_qualified(self, name: str) -> SymbolTableNode:
         raise NotImplementedError
 
+    @abstractmethod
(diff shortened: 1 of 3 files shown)


euresti:
Any idea when the release might be? I have a better fix that makes it work correctly for local functions.  Just wanting to know if I should rush on that.

euresti:
My fix builds on this change so this should land too.

gvanrossum:
Here's the release schedule: https://github.com/python/mypy/issues/4606

euresti:
Can this be merged? I'd hate for someone who's testing out the attrs support in master getting crashes.

ilevkivskyi:
So if #4607 is merged, then this is unnecessary, right? In this case I would wait for whenever #4607 lands (I will review it now) rather than applying a temporary fix.

PR #4585, opened 2018-02-18
Treat divmod like a binary operator

euresti:


gvanrossum:
David, can you rebase this so it includes the needed typeshed change? That should make the tests pass. Otherwise LG.

euresti:
Woo!

gvanrossum:
Yay!

Issue #4587, opened 2018-02-19
Cannot detect when function returns None but should return int


hvnsweeting:
* Are you reporting a bug, or opening a feature request?
Bug

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.

```python

# mypytrueneg.py
def hello2():
    # type: () -> int
    return None


def hello():
    # type: () -> None
    return 1
```

* What is the actual behavior/output?

```
 $ mypy mypytrueneg.py
mypytrueneg.py:8: error: No return value expected
```

* What is the behavior/output you expect?
mypy should detect that `hello2` returns `None` and it is not `int` as statically type declared.

* What are the versions of mypy and Python you are using?
  Do you see the same issue after installing mypy from Git master?
```
$ python --version
Python 3.5.2

$ mypy --version
mypy 0.560

Also tried
$ mypy --version
mypy 0.570-dev-b0af822a79465a93d55e67b5aea74551a63c08f4
```

JukkaL:
By default, `None` is considered compatible with everything. You can run mypy with `--strict-optional` and it will catch the error.

PR #4594, opened 2018-02-20
[Doc] mypy skips check function return None as a common issue

hvnsweeting:
When I got problem (https://github.com/python/mypy/issues/4587) , I looked into the doc and noticed "common issues" and FAQ but non of these helped me. Add this as I think it is a common issue when new user evaluating using mypy, they might write simple functions to see how the tool work.

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -51,6 +51,11 @@ flagged as an error.
 
   If you don't know what types to add, you can use ``Any``, but beware:
 
+- **The function returns ``None`` value but it was annotated as having a return
+  type of other types.** By default, ``None`` value is considered compatible
+  with everything. See :ref:`optional` for details and an experimental mode
+  which allows mypy to check ``None`` values precisely.
+
 - **One of the values involved has type ``Any``.** Extending the above
   example, if we were to leave out the annotation for ``a``, we'd get
   no error:
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -51,12 +51,6 @@ flagged as an error.
 
   If you don't know what types to add, you can use ``Any``, but beware:
 
-- **A function annotated as non-optional returns ``None`` and mypy doesn't
-  complain** By default, ``None`` value is considered compatible
-  with everything. See :ref:`optional` for details on strict optional checking,
-  which allows mypy to check ``None`` values precisely, and will soon become
-  default.
-
 - **One of the values involved has type ``Any``.** Extending the above
   example, if we were to leave out the annotation for ``a``, we'd get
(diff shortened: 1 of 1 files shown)


ethanhs:
Hm, this is already pretty well documented at https://mypy.readthedocs.io/en/latest/kinds_of_types.html#the-type-of-none-and-optional-types, though it may be that a section in common issues is needed as well.

gvanrossum:
I actually like adding such things to "common issues" or "FAQ" since sometimes users are overwhelmed by the structure of the docs. (Up to a point -- a FAQ with 100s of sections is also overwhelming. :-)

I think the language needs to be cleaned up a bit, and `--strict-optional` is no longer experimental (in fact it is slated to become the default in the future, perhaps in the next half year).

docs/source/common_issues.rst
```diff
@@ -51,6 +51,11 @@ flagged as an error.
 
   If you don't know what types to add, you can use ``Any``, but beware:
 
+- **The function returns ``None`` value but it was annotated as having a return
```
ethanhs: I would re-word this to something like "A function annotated as non-Optional returns None and mypy doesn't complain".

docs/source/common_issues.rst
```diff
@@ -51,6 +51,11 @@ flagged as an error.
...
+  type of other types.** By default, ``None`` value is considered compatible
+  with everything. See :ref:`optional` for details and an experimental mode
```
ethanhs: ... for details on strict optional checking, which allows mypy to check `None` values precisely, and will soon become default.

docs/source/common_issues.rst
```diff
@@ -51,6 +51,12 @@ flagged as an error.
 
   If you don't know what types to add, you can use ``Any``, but beware:
 
+- **A function annotated as non-optional returns ``None`` and mypy doesn't
+  complain** By default, ``None`` value is considered compatible
```
ethanhs: "`None` value" > "`None`" or put "the" before `None`, since you are talking about a definite noun.

docs/source/common_issues.rst
```diff
@@ -85,6 +85,18 @@ flagged as an error.
   clarity about the latter use ``--follow-imports=error``.  You can
   read up about these and other useful flags in :ref:`command-line`.
 
+- **A function annotated as returning non-optional type but returns ``None``
```
ethanhs: Hm, perhaps this should be:
>annotated as returning **a** non-optional type returns ``None``...

gvanrossum: You can make this small edit on the web yourself and then merge. No need to wait for OP.

PR #4607, opened 2018-02-21
Better support for converter in attrs plugin.

euresti:
My first pass at this only supported converters that were reachable
from the root.  With this code the converter can be a local function
too.
    
Also now when we can't detect the type of the converter the init type
becomes Any.

Supersedes #4583


test-data/unit/check-attr.test
```diff
@@ -532,6 +532,21 @@ class C:
...
+    def do_it(self, int) -> str:
+        ...
+thing = Thing()
+def factory(default: int):
+   ...
+@attr.s
+class C:
+    x: str = attr.ib(converter=thing.do_it)
+    y: str = attr.ib(converter=lambda x: x)
+    z: str = attr.ib(converter=factory(8))
```
ilevkivskyi: These should give errors until supported.

By the way I am thinking about how to support this. One possible way is to just accept type checker at a later stage. This would require a second part of the plugin that would be triggered on visiting this `TypeInfo` in type checker. We also need to be sure to mark all nodes that need this type as deferred. TBH I am sure good support for converters worth all this (including potential performance problems). So maybe we can just propose a workaround like always explicitly define your converters i.e. use
```python
def str_first_item(seq: List[int]) -> int:
    """A docstring for why one needs this"""
    return str(seq[0])

@attr.s
class C:
    x: str = attr.ib(converter=str_first_item)
```
instead of
```python
attr.s
class C:
    x: str = attr.ib(converter=lambda x: str(x[0]))
```

test-data/unit/check-incremental.test
```diff
@@ -3745,6 +3745,183 @@ class C(A, B):
...
+def foo() -> None:
+    def foo(x: str) -> int:
+        ...
+    @attr.s
+    class A:
+        x: int = attr.ib(converter=foo)
+    reveal_type(A)
+[builtins fixtures/list.pyi]
+[out1]
+main:8: error: Revealed type is 'def (x: str?) -> __main__.A@5'
```
ilevkivskyi: Why `str` is shown as unbound type here? This may cause problems later.

test-data/unit/check-incremental.test
```diff
@@ -3745,6 +3745,183 @@ class C(A, B):
...
+def maybe_str(x: Optional[str]) -> str:
+   ...
+@attr.s
+class SubBB(SubBase):
+    xx: int = attr.ib(converter=maybe_int)
+    yy: str = attr.ib(converter=maybe_str)
+    zz: bool = attr.ib(converter=bar.maybe_bool)
+[builtins fixtures/list.pyi]
+[out1]
+main:2: error: Revealed type is 'def (x: Union[builtins.int, builtins.None], y: Union[builtins.str, builtins.None], z: Union[builtins.bool, builtins.None]) -> base.Base'
```
ilevkivskyi: This is a bit hard to read. Could you please change this to instead checking types of things by assignments that would cause an error on the _second_ run only? For example:
```
import a
[file a.py]
from b import thing
x: str = thing # OK
[file a.py.2]
from b import thing
x: int = thing # Error
[file b.py]
thing = 'thing'
```

ilevkivskyi:
@gvanrossum I would revert this and merge later, there are some suspicious thing (like unbound types where they should be bound).

test-data/unit/check-incremental.test
euresti: I'll investigate

test-data/unit/check-incremental.test
euresti: Ok. It'll be more like checking class creation though.

test-data/unit/check-attr.test
euresti: Yeah. That's what I was already thinking for my own code.  Should I mention this in the error message?  "only functions supported" or something like that?

ilevkivskyi: > Should I mention this in the error message? "only functions supported" or something like that?

Yes, this will be helpful. Maybe like "only named functions" (to make it clear that lambdas are not fully supported).

test-data/unit/check-incremental.test
ilevkivskyi: > I'll investigate

This one actually worries me most. It would be great to fix at least this one before the release.

Issue #4599, opened 2018-02-20
Crash in incremental mode when binder interacts with callable types


ilevkivskyi:
There is one more crash without a simple repro:
```
Traceback (most recent call last):
  File "[somewhere]/mypy/main.py", line 80, in main
    type_check_only(sources, bin_dir, options, flush_errors)
  File "[somewhere]/mypy/main.py", line 129, in type_check_only
    flush_errors=flush_errors)
  File "[somewhere]/mypy/build.py", line 172, in build
    result = _build(sources, options, alt_lib_path, bin_dir, saved_cache, flush_errors)
  File "[somewhere]/mypy/build.py", line 258, in _build
    graph = dispatch(sources, manager)
  File "[somewhere]/mypy/build.py", line 2048, in dispatch
    process_graph(graph, manager)
  File "[somewhere]/mypy/build.py", line 2349, in process_graph
    process_stale_scc(graph, scc, manager)
  File "[somewhere]/mypy/build.py", line 2525, in process_stale_scc
    graph[id].type_check_first_pass()
  File "[somewhere]/mypy/build.py", line 1923, in type_check_first_pass
    self.type_checker().check_first_pass()
  File "[somewhere]/mypy/checker.py", line 228, in check_first_pass
    self.accept(d)
  File "[somewhere]/mypy/checker.py", line 323, in accept
    stmt.accept(self)
  File "[somewhere]/mypy/nodes.py", line 701, in accept
    return visitor.visit_class_def(self)
  File "[somewhere]/mypy/checker.py", line 1251, in visit_class_def
    self.accept(defn.defs)
  File "[somewhere]/mypy/checker.py", line 323, in accept
    stmt.accept(self)
  File "[somewhere]/mypy/nodes.py", line 763, in accept
    return visitor.visit_block(self)
  File "[somewhere]/mypy/checker.py", line 1371, in visit_block
    self.accept(s)
  File "[somewhere]/mypy/checker.py", line 323, in accept
    stmt.accept(self)
  File "[somewhere]/mypy/nodes.py", line 574, in accept
    return visitor.visit_decorator(self)
  File "[somewhere]/mypy/checker.py", line 2537, in visit_decorator
    self.check_func_item(e.func, name=e.func.name())
  File "[somewhere]/mypy/checker.py", line 640, in check_func_item
    self.check_func_def(defn, typ, name)
  File "[somewhere]/mypy/checker.py", line 803, in check_func_def
    self.accept(item.body)
[...]

PR #4611, opened 2018-02-21
Fix callable types with inconsistent argument counts

JukkaL:
This fixes potential issues with callable types where there are a
different number of argument names compared to argument types. These
malformed callable types might have caused crashes.

This may address #4599. Even if the issue remains unfixed, future
tracebacks could be more useful due to an assert that I added.


gvanrossum:
Did you not find a small repro for a test yet? Even though you know how to fix it?

JukkaL:
The new assert caused a few failures in existing tests.

Issue #4590, opened 2018-02-19
"Incompatible return value type" error with TypeVar


mikez302:
I am using Python 3.6.4 and the latest version of Mypy from master (mypy 0.570-dev-3545a71ba11576b08007f70d8407bf7924738886).

I have a file with the following code:
```python
from typing import TypeVar

T = TypeVar('T', int, str)

def foo() -> T:
    return 1
```

If I run it through Mypy, I get this error: `typevar_test.py:6: error: Incompatible return value type (got "int", expected "str")`.

If I make the function return a string, I get a similar error: `typevar_test.py:6: error: Incompatible return value type (got "str", expected "int")`.

If I make the function return something not specified by the TypeVar (such as a list, for example), I get two errors:
```
typevar_test.py:6: error: Incompatible return value type (got "List[int]", expected "int")
typevar_test.py:6: error: Incompatible return value type (got "List[int]", expected "str")
```

gvanrossum:
That's not how type variables work. You seem to be looking for a type alias for a union, e.g.
```py
U = Union[int, str]
def foo() -> U:
    return 1
```
A type variable is used to create a generic function or class -- please read the docs:
http://mypy.readthedocs.io/en/latest/generics.html

mikez302:
I read the documentation about generic functions, both for Mypy and for the typing module.

That was a toy example. I tried to come up with the simplest thing that would trigger the error. I know that in this case, I use a union to do the same thing.

Regardless of that, does a type variable, used in the return hint of a function, absolutely need to also be used in the hint of one of its parameters? If so, then I am doing something invalid. But I don't see anything in the Mypy docs, in the typing module docs, or in PEP 484 that mentions this.

And even if I am doing something invalid, I think the error messages could be made clearer. I think it is strange that if I use an int, I'm told that a string is expected, but if I use a string, I'm told that an int is expected. That is what made me suspect that this might be a bug in Mypy, and not (only) incorrect usage.

I wish I could work on this and try to improve it, but I feel that I am still not familiar enough with Mypy to feel that I could realistically do it myself.

gvanrossum:
Even if you are using the type variable legitimately, you still cannot do what you have in your example.

There are indeed cases where the return type can contain a type variable without the arguments containing one, but the only use cases I'm familiar with are when returning a generic function, such as is done for certain decorators.

The problem with your example is that the type system doesn't have a way to describe the type of a value assigned from a call to `foo()`:
```py
x = foo()  # What is the type of x?
[...]

PR #4618, opened 2018-02-21
Clarify generic function documentation

elias6:
This explains that a variable of a generic type must be able to be resolved to a non-generic type, hopefully preventing the confusion that made me (and possibly others) think #4590 is a bug.

@gvanrossum, I used some of the wording in your explanation almost verbatim. I'm guessing you don't mind.

### DIFF
--- a/docs/source/generics.rst
+++ b/docs/source/generics.rst
@@ -224,19 +224,7 @@ example we use the same type variable in two generic functions:
        return seq[-1]
 
 A variable cannot have a type variable in its type unless the type
-variable is bound in a containing generic class or function:
-
-.. code-block:: python
-
-    from typing import TypeVar
-
-    T = TypeVar('T')      # Declare type variable
-
-    def bad_function() -> T:
-        return 1
-
-    x = bad_function()  # Error! There is no containing class or function binding T.
-                        # x has type T, which cannot be replaced with a non-generic type.
+variable is bound in a containing generic class or function.
 
 .. _generic-methods-and-generic-self:
 
--- a/docs/source/generics.rst
+++ b/docs/source/generics.rst
@@ -223,6 +223,26 @@ example we use the same type variable in two generic functions:
    def last(seq: Sequence[T]) -> T:
        return seq[-1]
 
+Note that a variable cannot have a type variable in its type unless the type variable is bound in a containing generic class or function:
+
+.. code-block:: python
+
+    from typing import Sequence, TypeVar
+
+    T = TypeVar('T')      # Declare type variable
+
+    def first(seq: Sequence[T]) -> T:
+        return seq[0]
+
+    s = first(['foo', 'bar'])  # s has type str.
+    n = first([1, 2, 3])       # n has type int
+
+    def bad_function() -> T:
+        return 1
+
+    x = bad_function()  # Error! There is no containing class or function binding T.
+                        # x has type T, which cannot be replaced with a non-generic type.
+
 .. _generic-methods-and-generic-self:
 
 Generic methods and generic self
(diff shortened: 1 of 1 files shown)


docs/source/generics.rst
```diff
@@ -223,6 +223,26 @@ example we use the same type variable in two generic functions:
...
 
+Note that a variable cannot have a type variable in its type unless the type variable is bound in a containing generic class or function:
+
+.. code-block:: python
+
+    from typing import Sequence, TypeVar
+
+    T = TypeVar('T')      # Declare type variable
+
+    def first(seq: Sequence[T]) -> T:
```
gvanrossum: There's no need to repeat the two "good" examples -- this example should be only concerned with what doesn't work.

docs/source/generics.rst
```diff
@@ -223,6 +223,26 @@ example we use the same type variable in two generic functions:
    def last(seq: Sequence[T]) -> T:
        return seq[-1]
 
+Note that a variable cannot have a type variable in its type unless the type variable is bound in a containing generic class or function:
```
gvanrossum: Please break the long line. Also, maybe drop the leading "Note that" -- the previous section also starts like that and it's a bit repetitive.

docs/source/generics.rst
```diff
@@ -223,20 +223,15 @@ example we use the same type variable in two generic functions:
    def last(seq: Sequence[T]) -> T:
        return seq[-1]
 
-Note that a variable cannot have a type variable in its type unless the type variable is bound in a containing generic class or function:
+A variable cannot have a type variable in its type unless the type
+variable is bound in a containing generic class or function:
```
gvanrossum: Hm, sadly I just realized this is not quite correct, since this is possible:
```py
T = TypeVar('T')
S = TypeVar('S')

def repeat(n: int) -> Callable[[Callable[[T], T]], Callable[[T], T]]:
    def outer(func: Callable[[S], S]) -> Callable[[S], S]:
        def inner(x: S) -> S:
            for i in range(n):
                x = func(x)
            return x
        return inner
    return outer

fff = repeat(3)
reveal_type(fff)  # 'def [T] (def (T`-1) -> T`-1) -> def (T`-1) -> T`-1'
```
But I don't want to include such an example here -- I just wish to find a way to write it that doesn't exclude it.

Also this much simpler example *should* be invalid but isn't:
```py
def foo() -> T:
    pass

a = foo()
```

So I'm not sure what to do. Maybe @ilevkivskyi has an idea?

kamahen: `def foo() -> T: pass`

seems to be an unidiomatic way of writing

`def foo() -> Any: pass`

gvanrossum: They don't mean the same thing in mypy. The former returns a type containing `<nothing>` (meaning it has no values).

mikez302:
@gvanrossum it seems like you don't need your strange nested example to contradict the wording in your explanation. I ran `reveal_type` on the `first` example in the "Generic functions" section of the documentation, and it told me ``Revealed type is 'def [T] (seq: typing.Sequence[T`-1]) -> T`-1'``. I don't know exactly what that means, but isn't that an example of a type with a type variable, even though that type variable is not bound in a containing generic class or function?

I kind of have an idea of where and how type variables can and can't be used, based on the existing documentation and my own tinkering. But I can't imagine how to explain it precisely and rigorously, in a way that most mypy users will be able to understand. I think someone will have to help me with this.

gvanrossum:
Yeah, I went overboard with the complicated example. You'll note that all the examples where a type variable legitimately appears are types that are (or contain) generic functions. So maybe we should use that?

gvanrossum:
I propose to drop the example and just keep the sentence explaining the constraint. I'll just edit this myself so we can finish this up.

gvanrossum:
And, belatedly, thanks!

Issue #2036, opened 2016-08-18
Imports are sometimes un-silenced when using silent_imports + incremental.


Michael0x2a:
I recently ran into a bug related to silenced modules sometimes being un-silenced and re-checked (?) when combining silent_imports and incremental mode and forcing an SCC to be rechecked (using commit [fc03d438d61adf95e6a1aed6fb9fbdf986a90602](https://github.com/python/mypy/commit/fc03d438d61adf95e6a1aed6fb9fbdf986a90602), the latest version on master as of time of writing). 

While I was able to find a fix and will be submitting a pull request shortly, I wasn't able to write a test case. It would be nice to fix that, though it's challenging to do so for a few reasons. More details:

--- 

This bug manifests after you...

1. Make mypy run with silent_imports and incremental mode enabled with a cold cache.
2. Re-run mypy to establish a baseline for what running mypy on a fully fresh cache should be. On large code-bases, this step should finish substantially faster then step 1.
3. Make a harmless change to some file which happens to be importing some files not provided on the command line (eg are not being typechecked). This file may need to be a part of some odd import cycle shenanigans. 
4. Re-run mypy.
5. Re-run mypy again.

The expected behavior is that on step 5, mypy should take about the same amount of time to complete as on step 2, since there was no change after step 4 finished. However, the actual behavior is that step 5 will take about the same amount of time to run as step 4. After examining the logs, it seems the reason why this is the case is because in step 4, mypy is treating the previously silenced imports from the modified file as unsilenced, which causes them to be rechecked.

Unfortunately, I wasn't able to provide a test case because this problem seems to manifest mainly in larger code bases (with import cycles?) that are hard to simplify and after running incremental mode at least 3 times -- mypy's test suite can currently only run incremental two times.

While I was able to reliably repro the issue against the large code-base and will be submitting a pull request with a fix momentarily, it would be nice if it were possible to add an actual test case for this bug. 

I'll try and working on adding a test case for this (probably after overhauling parts of the test framework), but I wanted to file an issue here in case this ends up dropping off my todo list.

timabbott:
I encountered this issue with `-i` as well; I can upload a copy of a  buggy mypy cache dir in Zulip (open source) if it'd be helpful for debugging.

gvanrossum:
If it's repeatable, please do. Also add the exact revs of zulip and mypy you're using. Repeatable cases have often helped me track down pernicious `-i` bugs.

timabbott:
[...]

PR #4622, opened 2018-02-22
Optimize fine-grained update by using Graph as the cache

msullivan:
Make fine-grained update use a `Graph` as the source of truth instead of a `SavedCache`. We modify `load_graph` to optionally use an existing graph as a starting point.

This allows us to skip the expensive full `load_graph` (~500ms on S) and `preserve_full_cache` (~200ms on S) operations that essentially only convert between `Graph` and `SavedCache`.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1660,7 +1660,7 @@ class State:
         else:
             # Parse the file (and then some) to get the dependencies.
             self.parse_file()
-            self.suppressed = []
+            self.compute_dependencies()
             self.child_modules = set()
 
     def skipping_ancestor(self, id: str, path: str, ancestor_for: 'State') -> None:
@@ -1815,6 +1815,8 @@ class State:
         """
         # TODO: See if it's possible to move this check directly into parse_file in some way.
         # TODO: Find a way to write a test case for this fix.
+        # TODO: I suspect that splitting compute_dependencies() out from parse_file
+        # obviates the need for this but lacking a test case for the problem this fixed...
         silent_mode = (self.options.ignore_missing_imports or
                        self.options.follow_imports == 'skip')
         if not silent_mode:
@@ -1881,30 +1883,34 @@ class State:
         # TODO: Why can't SemanticAnalyzerPass1 .analyze() do this?
         self.tree.names = manager.semantic_analyzer.globals
 
-        # Compute (direct) dependencies.
-        # Add all direct imports (this is why we needed the first pass).
-        # Also keep track of each dependency's source line.
-        dependencies = []
-        suppressed = []
-        priorities = {}  # type: Dict[str, int]  # id -> priority
-        dep_line_map = {}  # type: Dict[str, int]  # id -> line
         for pri, id, line in manager.all_imported_modules_in_file(self.tree):
-            priorities[id] = min(pri, priorities.get(id, PRI_ALL))
-            if id == self.id:
-                continue
-            # Omit missing modules, as otherwise we could not type-check
-            # programs with missing modules.
-            if id in manager.missing_modules:
-                if id not in dep_line_map:
-                    suppressed.append(id)
-                    dep_line_map[id] = line
-                continue
             if id == '':
                 # Must be from a relative import.
                 manager.errors.set_file(self.xpath, self.id)
                 manager.errors.report(line, 0,
                                       "No parent module -- cannot perform relative import",
                                       blocker=True)
+
+        self.check_blockers()
+
+    def compute_dependencies(self):
+        """Compute a module's dependencies after parsing it.
+
+        This is used when we parse a file that we didn't have
+        up-to-date cache information for. When we have an up-to-date
+        cache, we just use the cached info.
+        """
+        manager = self.manager
+
+        # Compute (direct) dependencies.
+        # Add all direct imports (this is why we needed the first pass).
+        # Also keep track of each dependency's source line.
+        dependencies = []
+        priorities = {}  # type: Dict[str, int]  # id -> priority
+        dep_line_map = {}  # type: Dict[str, int]  # id -> line
+        for pri, id, line in manager.all_imported_modules_in_file(self.tree):
+            priorities[id] = min(pri, priorities.get(id, PRI_ALL))
+            if id == self.id:
                 continue
             if id not in dep_line_map:
                 dependencies.append(id)
@@ -1913,17 +1919,17 @@ class State:
         if self.id != 'builtins' and 'builtins' not in dep_line_map:
             dependencies.append('builtins')
 
-        # If self.dependencies is already set, it was read from the
-        # cache, but for some reason we're re-parsing the file.
         # NOTE: What to do about race conditions (like editing the
         # file while mypy runs)?  A previous version of this code
         # explicitly checked for this, but ran afoul of other reasons
         # for differences (e.g. silent mode).
+
+        # Missing dependencies will be moved from dependencies to
+        # suppressed when they fail to be loaded in load_graph.
         self.dependencies = dependencies
-        self.suppressed = suppressed
+        self.suppressed = []
         self.priorities = priorities
         self.dep_line_map = dep_line_map
-        self.check_blockers()
 
     def semantic_analysis(self) -> None:
         assert self.tree is not None, "Internal error: method must be called on parsed file only"
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -389,7 +389,6 @@ def default_lib_path(data_dir: str,
 CacheMeta = NamedTuple('CacheMeta',
                        [('id', str),
                         ('path', str),
-                        ('memory_only', bool),  # no corresponding json files (fine-grained only)
                         ('mtime', int),
                         ('size', int),
                         ('hash', str),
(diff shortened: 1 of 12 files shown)


JukkaL:
This makes a big performance difference, excellent! Using my randomized tester script, a cold run was about 70s (several files with small changes, time includes cache download) in the Dropbox S repo, and incremental updates were typically 5-10s (and no more than 20s).

I did a fair amount of testing, and found some common false positives. I isolated them into these test cases that pass on master:

```
[case testRefreshImportOfIgnoredModule1]
# flags: --follow-imports=skip --ignore-missing-imports
# cmd: mypy c.py a/__init__.py b.py
[file c.py]
from a import a2
import b
b.x
[file a/__init__.py]
[file b.py]
x = 0
[file b.py.2]
x = ''
[file b.py.3]
x = 0
[file a/a2.py]
[out]
==
==

[case testRefreshImportOfIgnoredModule2]
# flags: --follow-imports=skip --ignore-missing-imports
# cmd: mypy c.py a/__init__.py b.py
[file c.py]
from a import a2
import b
b.x
[file a/__init__.py]
[file b.py]
x = 0
[file b.py.2]
x = ''
[file b.py.3]
x = 0
[file a/a2/__init__.py]
[out]
==
==
```

It looks like that for coarse-grained incremental the existence of a file not included in a build affects build results. Fine-grained incremental mode considers the file as non-existent even if it exists (but apparently only when propagating fine-grained dependencies).

Example failure:

```
______________________________ testRefreshImportOfIgnoredModule2_cached ______________________________
data: /Users/jukka/src/mypy/test-data/unit/fine-grained-modules.test:868:
/Users/jukka/venv/mypy/lib/python3.6/site-packages/pluggy/__init__.py:617: in __call__
    return self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
/Users/jukka/venv/mypy/lib/python3.6/site-packages/pluggy/__init__.py:222: in _hookexec
    return self._inner_hookexec(hook, methods, kwargs)
/Users/jukka/venv/mypy/lib/python3.6/site-packages/pluggy/__init__.py:216: in <lambda>
    firstresult=hook.spec_opts.get('firstresult'),
/Users/jukka/venv/mypy/lib/python3.6/site-packages/_pytest/runner.py:109: in pytest_runtest_call
    item.runtest()
/Users/jukka/src/mypy/mypy/test/data.py:647: in runtest
    suite.run_case(self.case)
/Users/jukka/src/mypy/mypy/test/testfinegrained.py:122: in run_case
    testcase.file, testcase.line))
/Users/jukka/src/mypy/mypy/test/helpers.py:95: in assert_string_arrays_equal
    raise AssertionError(msg)
E   AssertionError: Invalid output (/Users/jukka/src/mypy/test-data/unit/fine-grained-modules.test, line 868)
---------------------------------------- Captured stderr call ----------------------------------------
Expected:
  ==
  ==
Actual:
  ==
  ==
  c.py:1: error: Module 'a' has no attribute 'a2' (diff)
```

msullivan:
The problem is due to the handling of `missing_modules`. Hopefully I'll have a fix tomorrow.

msullivan:
Fixed the missing_modules related failures.

This has been a bit of an adventure.

Since we are no longer doing a full `load_graph()`, we can't rely on
it fully populating `missing_modules` if it is cleared between
updates. If `missing_modules` isn't fully populated, then semanal will
spuriously reject `from x import y` where `x.y` is a missing module.

We can't just *not* clear `missing_modules`, however, because
`parse_file` uses it to decide that modules are suppressed. But this
is actually wrong! It can lead to an import failure message not being
generated because some other module has already failed to import it
(and perhaps even ignored the error).

`parse_file()` shouldn't actually *need* to compute `suppressed`,
though. If it is called on a file with no cache, `load_graph()` will
handle moving deps to `suppressed`, and if called on a file that has
had cache info loaded, then the dependency information has all been
loaded from the cache.

So we refactor things so that dep information from the cache is used
when available and `parse_file` doesn't need to deal with it.

I strongly suspect that this refactor obviates the need for
`fix_suppressed_dependencies`, but I was not able to quickly produce a
test case from the description in #2036, so I am not comfortable
making that change.

mypy/build.py
```diff
@@ -1896,49 +1883,54 @@ def parse_file(self) -> None:
         # TODO: Why can't SemanticAnalyzerPass1 .analyze() do this?
         self.tree.names = manager.semantic_analyzer.globals
 
+        for pri, id, line in manager.all_imported_modules_in_file(self.tree):
+            if id == '':
```
JukkaL: It's not obvious why an empty string indicates a relative import. Add a longer description, probably here and in the docstring of `all_imported_modules_in_file`.

mypy/build.py
```diff
@@ -658,23 +656,15 @@ def correct_rel_imp(imp: Union[ImportFrom, ImportAll]) -> str:
...
                         if self.is_module(sub_id):
                             res.append((pri, sub_id, imp.line))
-                        else:
-                            all_are_submodules = False
-                    # If all imported names are submodules, don't add
-                    # cur_id as a dependency.  Otherwise (i.e., if at
-                    # least one imported name isn't a submodule)
-                    # cur_id is also a dependency, and we should
-                    # insert it *before* any submodules.
-                    if not all_are_submodules:
```
JukkaL: Why was this check removed?

mypy/checker.py
```diff
@@ -216,6 +216,16 @@ def __init__(self, errors: Errors, modules: Dict[str, MypyFile], options: Option
...
 
+    def reset(self) -> None:
+        """Cleanup stale state that might be left over from a typechecking run.
+
+        This allows us to reuse TypeChecker objects in fine-grained
+        incremental mode.
+        """
+        self.partial_reported.clear()
+        assert self.partial_types == []
+        assert self.deferred_nodes == []
```
JukkaL: Maybe also reset `type_map`? We could ensure that the binder and `scope` are clear (or reset them as well). Also, consider clearing `inferred_attribute_types` and `module_refs`. Or maybe it would be easier to create a new `TypeChecker` object (doesn't need to happen in this PR).

mypy/server/update.py
```diff
@@ -170,9 +170,7 @@ def __init__(self,
         # for the cache. This is kind of a hack and it might be better to have
         # this directly reflected in load_graph's interface.
         self.options.cache_dir = os.devnull
-        mark_all_meta_as_memory_only(graph, manager)
-        manager.saved_cache = preserve_full_cache(graph, manager)
-        self.type_maps = extract_type_maps(graph)
+        manager.saved_cache = {}
```
JukkaL: Update the docstring of `BuildManager` to reflect the fact that `saved_cache` is no longer used in fine-grained incremental mode.

test-data/unit/fine-grained-modules.test
```diff
@@ -846,3 +860,92 @@ x = 1
...
+[file b.py]
+import c
+x = c.Foo()
+[file c.py]
+class Foo:
+    def foo(self) -> None: pass
+[delete c.py.2]
+[file b.py.2]
+class Foo:
+    def foo(self) -> None: pass
```
JukkaL: Maybe add a third step where the signature of `foo` is changed, which would result in an error in the main module.

mypy/checker.py
msullivan: I got a noticeable (though not particularly big) performance win by reseting the `TypeChecker` instead of creating a new one.

mypy/server/update.py
msullivan: Will do. 

Once this is landed I'm also going to just totally rip out the coarse-grained daemon and all of the `saved_cache` machinery entirely.

mypy/build.py
msullivan: The lack of the parent module in the dependency list was involved in a weird interaction with the original changes that resulted in the wrong module name appearing in an error message in one case.

The other changes to build's dependency handling appear to solve this without this check, though, so I'm going to undo this change.

test-data/unit/fine-grained-modules.test
msullivan: :white_check_mark:

mypy/build.py
```diff
@@ -2186,13 +2186,16 @@ def dump_graph(graph: Graph) -> None:
...
+def load_graph(sources: List[BuildSource], manager: BuildManager,
+               old_graph: Optional[Graph] = None) -> Graph:
     """Given some source files, load the full dependency graph.
 
     As this may need to parse files, this can raise CompileError in case
     there are syntax errors.
     """
-    graph = {}  # type: Graph
+
+    graph = old_graph or {}  # type: Graph
```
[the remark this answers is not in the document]
msullivan: Yep, that's the goal.

mypy/build.py
```diff
@@ -1896,49 +1899,48 @@ def parse_file(self) -> None:
...
-        suppressed = []
         priorities = {}  # type: Dict[str, int]  # id -> priority
         dep_line_map = {}  # type: Dict[str, int]  # id -> line
         for pri, id, line in manager.all_imported_modules_in_file(self.tree):
             priorities[id] = min(pri, priorities.get(id, PRI_ALL))
             if id == self.id:
                 continue
-            # Omit missing modules, as otherwise we could not type-check
-            # programs with missing modules.
-            if id in manager.missing_modules:
```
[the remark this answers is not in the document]
msullivan: The graph loading code is pretty subtle, so I wouldn't say I am 100% confident, but I am reasonably so---I discussed my reasoning some in my comment above.

PR #4623, opened 2018-02-23
Overhaul file-system operations to use FileSystemCache

msullivan:
This overhauls our file-system accesses to always go through a `FileSystemCache` when stat()ing or reading source files. (We don't use it for reading cache json files, since we seem to only do a single read on each anyways and so caching doesn't help us any.) This eliminates a number of potential bugs having to do with seeing inconsistent file-system state and reduces the number of file reads+decodes in fine-grained incremental mode.

This also allows us to eliminate a number of caches maintained by `find_module` that duplicated functionality of `FileSystemCache`. Since I was reluctant to make the `FileSystemCache` global, I took the opportunity to make `find_module`'s caches non-global as well.


msullivan:
Argh: on windows python generates different error messages in the `FileNotFoundError` for `open` and `os.stat`, and this changes things to do `os.stat` first.

mypy/server/update.py
```diff
@@ -499,13 +505,14 @@ def get_module_to_path_map(manager: BuildManager) -> Dict[str, str]:
             for module, node in manager.modules.items()}
 
 
-def get_sources(modules: Dict[str, str],
+def get_sources(fscache: FileSystemCache,
+                modules: Dict[str, str],
                 changed_modules: List[Tuple[str, str]]) -> List[BuildSource]:
     # TODO: Race condition when reading from the file system; we should only read each
```
JukkaL: Can this TODO comment be removed now?

mypy/build.py
```diff
@@ -544,6 +548,23 @@ def find_config_file_line_number(path: str, section: str, setting_name: str) ->
     return -1
 
 
+class FindModuleCache:
```
JukkaL: Add docstring. Mention that this never flushes the FileSystemCache and caller is responsible for that.

mypy/build.py
```diff
@@ -802,74 +827,10 @@ def remove_cwd_prefix_from_path(p: str) -> str:
...
-            res = False
-        else:
-            names = list_dir(head)
-            res = names is not None and tail in names and os.path.isfile(path)
-        find_module_is_file_cache[path] = res
-    return res
-
-
-def find_module(id: str, lib_path_arg: Iterable[str]) -> Optional[str]:
+def find_module(cache: FindModuleCache, id: str, lib_path_arg: Iterable[str]) -> Optional[str]:
```
JukkaL: Should this be a method of `FindModuleCache`? If not, document in `FindModuleCache` that this function populates the data structures and is tightly coupled with the class.

mypy/build.py
```diff
@@ -878,19 +839,15 @@ def find() -> Optional[str]:
         # that will require the same subdirectory.
         components = id.split('.')
         dir_chain = os.sep.join(components[:-1])  # e.g., 'foo/bar'
-        if (dir_chain, lib_path) not in find_module_dir_cache:
+        if (dir_chain, lib_path) not in cache.dirs:
             dirs = []
             for pathitem in lib_path:
                 # e.g., '/usr/lib/python3.4/foo/bar'
-                isdir = find_module_isdir_cache.get((pathitem, dir_chain))
```
JukkaL: Why was this block of code removed?

msullivan: It's maintaining an `isdir` cache, which is handled by the `fscache` stat cache.

msullivan:
OK, all of the other changes I was making that conflicted with this are in now, so I've updated it all and made the requested cleanups.

Doesn't affect the performance of a trivial fine-grained incremental update at all.

Manual testing seems to pick up changes fine.

gvanrossum:
@ethanhs Note that this will most likely affect your PEP 561 PR (#4693) severely.

ethanhs:
@gvanrossum, indeed. I was originally hoping to get my old PR merged before this one was opened, but that of course didn't happen. The good news is my implementation in #4693 was in a single commit, so I should be able to revert that, rebase on master and re-apply the changes (with slight modification) without too much headache. Thanks for the heads up!

mypy/build.py
```diff
@@ -812,202 +820,122 @@ def remove_cwd_prefix_from_path(p: str) -> str:
...
-    find_module_listdir_cache.clear()
-    find_module_is_file_cache.clear()
-    find_module_isdir_cache.clear()
-
+class FindModuleCache:
+    """Module finder with integrated cache.
 
-def list_dir(path: str) -> Optional[List[str]]:
-    """Return a cached directory listing.
+       Module locations and some intermediate results are cached internally
```
JukkaL: Style nit: Indents seem off unless github diff view is acting up. Just to be extra clear, indents should be like this:

```py
class Foo:
    """First line.

    Second line (4 space indent).
    """
```

mypy/build.py
```diff
@@ -812,202 +820,122 @@ def remove_cwd_prefix_from_path(p: str) -> str:
...
+            # Subtle: this code prefers the .pyi over the .py if both
+            # exists, and also prefers packages over modules if both x/
+            # and x.py* exist.  How?  We sort the directory items, so x
+            # comes before x.py and x.pyi.  But the preference for .pyi
+            # over .py is encoded in find_module(); even though we see
+            # x.py before x.pyi, find_module() will find x.pyi first.  We
+            # use hits to avoid adding it a second time when we see x.pyi.
+            # This also avoids both x.py and x.pyi when x/ was seen first.
+            hits = set()  # type: Set[str]
+            for item in sorted(os.listdir(os.path.dirname(module_path))):
```
JukkaL: Should we use `FileSystemCache` here and below?

mypy/build.py
```diff
@@ -1891,11 +1809,14 @@ def parse_file(self) -> None:
             if self.path and source is None:
                 try:
                     path = manager.maybe_swap_for_shadow_path(self.path)
-                    source, self.source_hash = read_with_python_encoding(
-                        path, self.options.python_version)
+                    source = manager.fscache.read_with_python_encoding(path)
+                    self.source_hash = manager.fscache.md5(path)
                 except IOError as ioerr:
+                    # ioerr.strerror differs for os.stat failures between Windows and
+                    # other systems, but os.strerror(ioerr.errno) does not, so we use that.
```
gvanrossum: Maybe explain that we want the error to be platform-independent so tests can have predictable output?

mypy/fscache.py
```diff
@@ -97,12 +97,39 @@ def listdir(self, path: str) -> List[str]:
         return results
 
     def isfile(self, path: str) -> bool:
-        st = self.stat(path)
-        return stat.S_ISREG(st.st_mode)
+        try:
+            st = self.stat(path)
+            return stat.S_ISREG(st.st_mode)
```
gvanrossum: Possibly move this out of the try/except -- this can't raise OSError.

mypy/fscache.py
```diff
@@ -97,12 +97,39 @@ def listdir(self, path: str) -> List[str]:
...
+        except OSError:
+            return False
+
+    def isfile_case(self, path: str) -> bool:
+        """Return whether path exists and is a file.
+
+        On case-insensitive filesystems (like Mac or Windows) this returns
+        False if the case of the path's last component does not exactly
+        match the case found in the filesystem.
```
gvanrossum: Hm, this means that importing wrongly-cased *packages* will not be flagged, since we only compare the filename (which would be `__init__.py` for a package). Not actionable except perhaps add a TODO comment; this was a pre-existing problem. (Python itself gets it right so that's a fallback.)

mypy/dmypy_server.py
```diff
@@ -298,9 +302,7 @@ def initialize_fine_grained(self, sources: List[mypy.build.BuildSource]) -> Dict
                 out, err = '', output
             return {'out': out, 'err': err, 'status': 2}
         messages = result.errors
-        manager = result.manager
-        graph = result.graph
-        self.fine_grained_manager = FineGrainedBuildManager(manager, graph)
+        self.fine_grained_manager = mypy.server.update.FineGrainedBuildManager(result)
```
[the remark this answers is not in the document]
msullivan: Improperly resolved a merge conflict. When this patch first went up it used a prefix, but I changed it to not in another patch and must have undone it while merging.

msullivan:
I updated `remove_cwd_prefix_from_path` to use the cache, but `getmtime` is only used for cache json files, which we don't use the fscache for.

mypy/build.py
```diff
@@ -812,202 +820,122 @@ def remove_cwd_prefix_from_path(p: str) -> str:
...
-                mod = item.split('.')[0]
-                if mod not in hits:
-                    hits.add(mod)
-                    result += find_modules_recursive(
-                        module + '.' + mod, lib_path)
-    return result
-
-
-def verify_module(id: str, path: str) -> bool:
+def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
```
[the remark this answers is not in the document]
msullivan: I'd rather not add more knowledge about python-specific things to FileSystemCache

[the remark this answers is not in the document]
gvanrossum: OK.

gvanrossum:
Yay!

PR #4628, opened 2018-02-23
Fix issues with attr code.

euresti:
More meesages, no unbound types, better tests.

In response to comments in #4607

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -64,11 +64,13 @@ class Attribute:
                     and converter.type
                     and isinstance(converter.type, CallableType)
                     and converter.type.arg_types):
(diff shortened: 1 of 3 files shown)


mypy/plugins/attrs.py
```diff
@@ -334,6 +337,10 @@ def _get_converter_name(converter: Optional[Expression]) -> Optional[str]:
                 and converter.node.type.arg_types):
             return converter.node.fullname()
         # Signal that we have an unsupported converter.
+        ctx.api.fail(
+            "Unsupported converter function, only named functions are currently supported",
+            converter
+        )
```
ilevkivskyi: For future, we typically don't put parentheses on separate lines (no need to fix this now).

PR #4632, opened 2018-02-25
Documentation for attr support

euresti:


euresti:
I don't know if this is the right place for it, or into how much detail I should go. Let me know!

ethanhs:
Hm, I don't consider attrs a Python feature per-se. I think attrs documentation would probably be a better fit for the additional features page.

docs/source/supported_python_features.rst
```diff
@@ -18,3 +18,75 @@ features. However, you need to be careful if you decide to do this. If
...
+.. code-block:: python
+
+    import attr
+    @attr.s(auto_attribs=True)
+    class A:
+        one: int
+        two: int = 7
+        three: int = attr.ib(8)
+
+The typeshed has a couple of "white lie" annotations to make type checking
```
ethanhs: "The typeshed" -> "Typeshed", since there aren't other typesheds.

docs/source/supported_python_features.rst
```diff
@@ -18,3 +18,75 @@ features. However, you need to be careful if you decide to do this. If
...
+.. code-block:: python
+
+    import attr
+    from typing import Dict
+    @attr.s(auto_attribs=True)
+        one: int = attr.ib(8)
+        two: Dict[str, str] = attr.Factory(dict)
+        bad: str = attr.ib(16)   # Error: can't assign int to str
+
+Caveats/Known Issues
```
ethanhs: Is this supposed to be made a section heading?

docs/source/supported_python_features.rst
```diff
@@ -18,3 +18,75 @@ features. However, you need to be careful if you decide to do this. If
...
+
+* Currently, ``converter`` only supports named functions.  If mypy finds something else it
+  will complain about not understanding the argument and the type annotation in
+  ``__init__`` will be replaced by ``Any``.
+
+* `Validator decorators <http://www.attrs.org/en/stable/examples.html#decorator>`_
+  and `default decorators <http://www.attrs.org/en/stable/examples.html#defaults>`_
+  are not type-checked against the attribute they are setting/validating.
+
+* Methods added by mypy currently overwrite any existing methods.  This is how
```
ethanhs: This sounds like mypy is mutating things at runtime. AIUI, that is not the case. Perhaps re-wording to talk about mypy recognizing/understanding attrs generated methods throughout would avoid confusion?

euresti: Switched to using method definition.  Which seems to be what most of the documentation uses.

euresti:
Thanks for the quick review.  I moved it to the new section.

docs/source/additional_features.rst
```diff
@@ -7,3 +7,78 @@ including the following:
...
+.. code-block:: python
+
+    import attr
+    @attr.s(auto_attribs=True)
+    class A:
+        one: int
+        two: int = 7
+        three: int = attr.ib(8)
+
+The Typeshed has a couple of "white lie" annotations to make type checking
```
ethanhs: I might have been unclear (sorry), but I was mostly talking about removing the preceding article. Typeshed is a proper noun and the name is referred to without a preceding article everywhere else.

gvanrossum:
Yay! Thanks. Now we need to clone this for `@dataclass(...)` in Python 3.7...

Issue #708, opened 2015-06-06
Cannot assign to field of Callable type


spkersten:
Consider this example:

``` python
from typing import Callable

class A:
    f = None  # type: Callable[[int], int]

    def __init__(self, g: Callable[[int], int]) -> None:
        self.f = g
```

mypy reports:

```
trial.py: In member "__init__" of class "A":
trial.py, line 7: Cannot assign to a method
trial.py, line 7: Invalid method type
trial.py, line 7: Incompatible types in assignment (expression has type Callable[[int], int], variable has type Callable[[], int])
```

while I'd expect this to work.

Note that there is a second error in the reporting of the type of the variable.

Also, this passes type checking (correctly):

``` python
f = None  # type: Callable[[int], int]

def g(x: int) -> int: return 0

f = g
```

spkersten:
The mismatching types is probably because mypy thinks `f` is a method and therefore strips off the first argument (self).

JukkaL:
Hmm.. actually not sure what's the right thing to do here. Based on a quick experiment, the desired behavior seems to depend on what kind of a callable `g` is in your original example.

Consider this:

```python
class A:
    f = None
def m(x):
    return x + 1
print(m(1))  # 2
A.f = m
print(A().f(1))  # Error (2 args given, expects 1)
```

But now look at this:

```python
class A:
    f = None
class M:
    def __call__(self, x):
        return x + 1
m = M()
print(m(1) # 2
A.f = m
print(A().f(1))  # Okay => 2
```

JukkaL:
Taking off priority label because the fix doesn't seem obvious.

spkersten:
Before reporting it I also thought for a while that I could fix it quickly :smile:.

tharvik:
Is there any update? The issue is still present. For some real world example, it is used in the [ctypes](https://docs.python.org/3/library/ctypes.html#function-prototypes) module (in the examples).

```python
from typing import Any, Callable

class A():
    a = lambda: None  # type: Callable[..., None]

def f() -> None: return None

A().a = f
```
gives
```
/tmp/assign_callable.py:8: error: Cannot assign to a method
/tmp/assign_callable.py:8: error: Incompatible types in assignment (expression has type Callable[[], None], variable has type Callable[..., None])
```

Having `Callable[..., None]` or `Callable[[], None]` do not resolve it, having `Callable[..., Any]` remove the incompatible types line.

gvanrossum:
[...]

PR #4633, opened 2018-02-25
Make attributes in attr classes be instance variables.

euresti:
This is only a problem when trying to assign callables into attrs.
This avoids hitting #708


euresti:
Hang on. This might cause crashes.  Don't review yet.

gvanrossum:
(Am I glad we didn't rush this into the 0.570 release. :-)

PR #4647, opened 2018-02-27
Optionally check that we don't have duplicate nodes after AST merge

JukkaL:
It's hard to manually ensure that all relevant nodes are merged in AST merge
and that no references to stale objects remain. This automated opt-in check
looks for these duplicate nodes. It can be enabled in tests by switching 
`mypy.test.testfinegrained.CHECK_CONSISTENCY` to True.

This was originally implemented by @msullivan. I did various minor changes
and some refactoring.

### DIFF
--- a/mypy/server/mergecheck.py
+++ b/mypy/server/mergecheck.py
@@ -1,147 +1,14 @@
-from collections import deque
-from collections.abc import Iterable
-from typing import List, Dict, Iterator, Optional, Tuple, Mapping
-import weakref
-import types
+from typing import Dict, List, Tuple
 
 from mypy.nodes import SymbolNode, Var, Decorator, OverloadedFuncDef, FuncDef
-
-
-method_descriptor_type = type(object.__dir__)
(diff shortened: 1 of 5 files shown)


JukkaL:
Currently this will actually find many duplicate nodes. I will fix these issues in separate PRs after this has been merged.

mypy/server/mergecheck.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def path_to_str(path: List[Tuple[object, object]]) -> str:
+    result = '<root>'
+    for attr, obj in path:
+        t = type(obj).__name__
+        if t in ('dict', 'tuple', 'SymbolTable', 'list'):
+            result += '[%s]' % repr(attr)
+        else:
+            if t == 'Var':
```
msullivan: mypy doesn't like this because it doesn't know about the type.

What's the reason for doing this based on the name of the type instead of the type?

PR #4650, opened 2018-02-28
Properly format Callables in stubgen

### LINKED ISSUE #4638: Empty argument list of a Callable is translated to <TypeList> in stub file
[PEP 484](https://www.python.org/dev/peps/pep-0484/#callable) mentions an example for typing a function with no arguments (I added sample function body to be able to pass it to ``stubgen``):

```python
from typing import Callable

def feeder(get_next_item: Callable[[], str]) -> None:
    return None
```
Now run ``stubgen`` over the file with the above code:

```sh
$ python -m mypy.stubgen spam
Created out/spam.pyi
```
The following stub file is generated:
```python
# Stubs for spam (Python 3.6)
#
# NOTE: This dynamically typed stub was automatically generated by stubgen.

from typing import Callable

def feeder(get_next_item: Callable[<TypeList >, str]) -> None: ...
```
What confuses me is the ``<TypeList >`` instead of an empty list.

evhub:
Resolves #4638/#4640.

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -65,7 +65,7 @@ from mypy.nodes import (
 from mypy.stubgenc import parse_all_signatures, find_unique_signatures, generate_stub_for_c_module
 from mypy.stubutil import is_c_module, write_header
 from mypy.options import Options as MypyOptions
-from mypy.types import Type, TypeStrVisitor, AnyType, CallableType, UnboundType, NoneTyp, TupleType
+from mypy.types import Type, TypeStrVisitor, AnyType, CallableType, UnboundType, NoneTyp, TupleType, TypeList
 from mypy.visitor import NodeVisitor
 
 Options = NamedTuple('Options', [('pyversion', Tuple[int, int]),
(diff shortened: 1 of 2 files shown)


mypy/stubgen.py
```diff
@@ -65,7 +65,7 @@
 from mypy.stubgenc import parse_all_signatures, find_unique_signatures, generate_stub_for_c_module
 from mypy.stubutil import is_c_module, write_header
 from mypy.options import Options as MypyOptions
-from mypy.types import Type, TypeStrVisitor, AnyType, CallableType, UnboundType, NoneTyp, TupleType
+from mypy.types import Type, TypeStrVisitor, AnyType, CallableType, UnboundType, NoneTyp, TupleType, TypeList
```
ethanhs: This line is too long. You probably should use a multi-line import.

ethanhs:
@evhub thanks for working on this. Could you add some tests please? Definitely any minimal reproductions in the original issues, but also just a few common/more complex example tests of what was broken/solved. Thanks again!

PR #4651, opened 2018-02-28
Add instructions for encountering AssertionErrors

JakobGM:
The thought is that new users do not necessarily know that AssertionErrors are not how mypy is supposed to communicate with the end user. The two times I have encountered AssertionErrors the last year have been caused by cold caches, and it took my a while before realizing it (after debugging my own code for some time).

This pull request adds instructions for the user to clear their cache and report an issue if an AssertionError goes all the way up the stack to the CLI entrypoint.

I can amend the pull request if you have different answers for the following questions:
- Should only AssertionErrors be caught (as it is now), or should we catch BaseException instead, since any uncaught exception this far up the stack is probably caused by a bug.
- Is the ``console_entry`` function the correct place to put this. Should we guard ``mypy.__main__::__main__`` too?

Thanks in advance!
Jakob

### DIFF
--- a/mypy/__main__.py
+++ b/mypy/__main__.py
@@ -1,10 +1,20 @@
 """Mypy type checker command line tool."""
 
+import sys
+
 from mypy.main import main
 
 
 def console_entry() -> None:
-    main(None)
+    try:
+        main(None)
+    except AssertionError:
+        print("Uncaught AssertionError thrown by mypy. "
+              "This may be caused by a bug in mypy, and can sometimes be "
+              "worked around by clearing the '.mypy_cache' directory. "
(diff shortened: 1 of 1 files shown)


gvanrossum:
> ... caused by cold caches

I assume you meant they were caused by a *warm* cache?

In any case I don't want to encourage users to just clear the cache and move along -- we need the bug reports. Maybe you can use the same phrasing as in errors.py (`report_internal_error()`)? Maybe you can even reuse that code? (It's only invoked for certain specific errors, which is probably why you've seen a bare assertion failure -- if an assertion fails that's not caught you only get a traceback.)

gvanrossum:
I think we should just not do this. If you receive an assertion, you should report it to this tracker.

PR #4652, opened 2018-02-28
Fine-grained: Fix AST merge issues

JukkaL:
This fixes all but one AST merge checker test failures. The remaining
one seems tricky to fix so I'll leave it for later.

I decided not to add tests for things that can be caught using the
merge checker. We'll have to run the merge checker periodically
since it's not enabled by default in the test suite.

### DIFF
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -245,7 +245,7 @@ class TypeList(Type):
     This is used for the arguments of a Callable type, i.e. for
     [arg, ...] in Callable[[arg, ...], ret]. This is not a real type
     but a syntactic AST construct. UnboundTypes can also have TypeList
-    types before they are processed into Callable type.s
+    types before they are processed into Callable types.
     """
 
     items = None  # type: List[Type]
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -242,9 +242,10 @@ class CallableArgument(Type):
 class TypeList(Type):
     """Information about argument types and names [...].
 
-    This is only used for the arguments of a Callable type, i.e. for
+    This is used for the arguments of a Callable type, i.e. for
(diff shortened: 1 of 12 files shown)


mypy/types.py
```diff
@@ -242,9 +242,10 @@ def serialize(self) -> JsonDict:
 class TypeList(Type):
     """Information about argument types and names [...].
 
-    This is only used for the arguments of a Callable type, i.e. for
+    This is used for the arguments of a Callable type, i.e. for
     [arg, ...] in Callable[[arg, ...], ret]. This is not a real type
-    but a syntactic AST construct.
+    but a syntactic AST construct. UnboundTypes can also have TypeList
+    types before they are processed into Callable type.s
```
msullivan: Typo

mypy/types.py
```diff
@@ -1503,6 +1503,11 @@ def visit_type_type(self, t: TypeType) -> T:
     def visit_forwardref_type(self, t: ForwardRef) -> T:
         raise RuntimeError('Internal error: unresolved forward reference')
 
+    def visit_type_list(self, t: TypeList) -> T:
```
msullivan: Can other sorts of "synthetic" types show up in our traversal?

Would it be a better approach in the future to make the merger a synthetic type traverser?

JukkaL: Hmm looks like we could be hitting other such types as well. Maybe making it a "synthetic type visitor" is the right thing to do. I'll consider making a follow-up PR about this. 

On the other hand, we should perhaps only have one sort of type visitor. I don't really see the benefit of having two sorts of visitors.

JukkaL:
The remaining failure is `testFineGrainedCallable`.

PR #4656, opened 2018-03-01
Fine-grained: Support NewType and reset subtype caches

JukkaL:
NewType work highlighted an issue with subtype caches with stale
information leaking, and this fixes that issue as well. We reset the
subtype cache in two places:

* When calculating the MRO; we reset caches of all base classes as
  well.

* When merging a new version of a TypeInfo, which may have a
  different MRO; we reset all caches of base classes in the old
  MRO, as they might no longer be supertypes.

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -2080,6 +2080,11 @@ class TypeInfo(SymbolNode):
             return (left, right) in self._cache
         return (left, right) in self._cache_proper
 
+    def reset_subtype_cache(self) -> None:
+        for item in self.mro:
+            item._cache = set()
+            item._cache_proper = set()
+
     def __getitem__(self, name: str) -> 'SymbolTableNode':
         n = self.get(name)
         if n:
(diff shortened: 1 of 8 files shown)


mypy/nodes.py
```diff
@@ -2122,6 +2127,7 @@ def calculate_mro(self) -> None:
         self.is_enum = self._calculate_is_enum()
         # The property of falling back to Any is inherited.
         self.fallback_to_any = any(baseinfo.fallback_to_any for baseinfo in self.mro)
+        self.reset_subtype_cache()
```
msullivan: I don't fully understand why this is necessary (though it certainly isn't /wrong/). Adding a new subclass shouldn't invalidate any of the caches, since they store only positive information?

JukkaL: I think that some changes in generic base classes may require this. Say, originally the base class was `C[A]` and after update it is `C[B]`. This would change the subtype relation even though the MRO is unchanged. If we are doing a refresh of the class there would be no AST merge so without this the subtype cache would not get emptied.

PR #4662, opened 2018-03-02
Update revision history for 0.570

msullivan:


gvanrossum:
LGTM. That's the only doc change (apart from nits)?

docs/source/revision_history.rst
```diff
@@ -3,6 +3,11 @@ Revision history
 
 List of major changes:
 
+- March 2017
```
JelleZijlstra: I think it's 2018 now :)

msullivan:
The only doc change I had planned at least. Attrs docs and making strict optional and incremental non-experimental went in before the branch cut.

gvanrossum:
Go ahead, merge and then CP into the release branch. (Thanks Jelle for catching the mistaken year!)

PR #4664, opened 2018-03-02
Don't modify the on disk cache in fine-grained mode

msullivan:
This is a little subtle, because interface_hash still needs to be
computed, as it is a major driver of the coarse-grained build process.

Since metas are no longer computed for files that get rechecked during
build, to avoid spuriously reprocessing them we need to find initial
file state in cache mode as well.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -281,10 +281,14 @@ class Server:
             # Pull times and hashes out of the saved_cache and stick them into
             # the fswatcher, so we pick up the changes.
             for state in self.fine_grained_manager.graph.values():
+                # Only grab hashes from modules that came from the
+                # cache. For modules that actually got parsed &
+                # typechecked we rely on the data from the actual disk
+                # cache, since we don't generate metas for those.
+                assert state.tree is not None
+                if not state.tree.is_cache_skeleton: continue
                 meta = state.meta
-                # If there isn't a meta, that means the current
-                # version got checked in the initial build.
-                if meta is None: continue
+                assert meta is not None
                 assert state.path is not None
                 self.fswatcher.set_file_data(
                     state.path,
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -254,9 +254,8 @@ class Server:
         self.fscache = FileSystemCache(self.options.python_version)
         self.fswatcher = FileSystemWatcher(self.fscache)
         self.update_sources(sources)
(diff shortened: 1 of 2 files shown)


mypy/dmypy_server.py
```diff
@@ -254,9 +254,8 @@ def initialize_fine_grained(self, sources: List[mypy.build.BuildSource]) -> Dict
         self.fscache = FileSystemCache(self.options.python_version)
         self.fswatcher = FileSystemWatcher(self.fscache)
         self.update_sources(sources)
-        if not self.options.use_fine_grained_cache:
-            # Stores the initial state of sources as a side effect.
-            self.fswatcher.find_changed()
+        # Stores the initial state of sources as a side effect.
```
JukkaL: Why was this changed to be executed unconditionally?

mypy/dmypy_server.py
```diff
@@ -274,16 +273,17 @@ def initialize_fine_grained(self, sources: List[mypy.build.BuildSource]) -> Dict
...
+        # If we are using the fine-grained cache, build might not have
+        # actually done the typechecking on the updated files yet.
         # Run a fine-grained update starting from the cached data
         if self.options.use_fine_grained_cache:
             # Pull times and hashes out of the saved_cache and stick them into
             # the fswatcher, so we pick up the changes.
             for state in self.fine_grained_manager.graph.values():
                 meta = state.meta
+                # If there isn't a meta, that means the current
+                # version got checked in the initial build.
```
JukkaL: Relying on this logic here seems a bit fragile. Maybe move the `meta is None` check to a helper method in `State` and use the helper method here instead?

mypy/dmypy_server.py
msullivan: So that we have accurate `fswatcher` cache information for files that we didn't read from the on-disk cache, now that we don't generate `CacheMeta`s

msullivan:
The main rationale is closer to correctness than performance. Currently we write and delete cache files during the initial load of fine-grained mode, but never during fine-grained updates. This means that fine-grained can mess the cache up but won't fix it without a restart, which seems not great.

mypy/dmypy_server.py
msullivan: Changing it to use `is_cache_skeleton` instead.

msullivan:
This has a bad interaction with #4669 and needs changes. My plan is to land #4669 and revise this after.

PR #4669, opened 2018-03-02
 Don't have newly added files trigger a coarse-grained rebuild in fine-grained cache mode

msullivan:


test-data/unit/fine-grained-modules.test
```diff
@@ -971,3 +971,56 @@ x = Foo()
 ==
 ==
 main:2: error: Too few arguments for "foo" of "Foo"
+
+[case testAddModuleAfterCache1]
```
JukkaL: Would it make sense to use the `-skip-nocache` suffix for some of these test cases?

test-data/unit/fine-grained-modules.test
```diff
@@ -971,3 +971,56 @@ x = Foo()
...
+[file b.py.2]
+def foo() -> None: pass
+[file b.py.3]
+def foo(x: int) -> None: pass
+[out]
+==
+a.py:2: error: Too many arguments for "foo"
+==
+
+[case testAddModuleAfterCache3]
```
JukkaL: I'd prefer to have more tests around this. For example, test that errors are correctly generated in all the new modules and existing modified modules, and make sure that we can correctly bind references to things defined in all three classes of modules:

1. Cached but unmodified
2. Cached and modified
3. New modules

test-data/unit/fine-grained-modules.test
msullivan: I don't see any need to. The main thing under test is cache behavior but they are still tests of module adding, which is subtle and could potentially find regressions in non-cache mode.

Issue #4635, opened 2018-02-26
stubgen tests are flaking


ethanhs:
I had noticed this a few days ago, it seems that stugen tests are flaking every once in a while, I just was able to get a repro:

```
_________________________ testVarDefsNotInAll_import __________________________
[gw3] win32 -- Python 3.6.4 C:\Python36\python.exe
data: C:\Users\ethanhs\Documents\mypy\test-data\unit\stubgen.test:369:
..\..\..\..\Documents\mypy\mypy\test\teststubgen.py:97: in run_case
    test_stubgen(testcase)
..\..\..\..\Documents\mypy\mypy\test\teststubgen.py:136: in test_stubgen
    a = load_output(out_dir)
..\..\..\..\Documents\mypy\mypy\test\teststubgen.py:158: in load_output
    assert entries, 'No files generated'
E   AssertionError: No files generated
E   assert []
---------------------------- Captured stdout call -----------------------------
Failed to import prog_ha2bjeqt; skipping it
_______________________ testIncludeClassNotInAll_import _______________________
[gw3] win32 -- Python 3.6.4 C:\Python36\python.exe
data: C:\Users\ethanhs\Documents\mypy\test-data\unit\stubgen.test:379:
..\..\..\..\Documents\mypy\mypy\test\teststubgen.py:97: in run_case
    test_stubgen(testcase)
..\..\..\..\Documents\mypy\mypy\test\teststubgen.py:136: in test_stubgen
    a = load_output(out_dir)
..\..\..\..\Documents\mypy\mypy\test\teststubgen.py:158: in load_output
    assert entries, 'No files generated'
E   AssertionError: No files generated
E   assert []
---------------------------- Captured stdout call -----------------------------
Failed to import prog_fp1d8vup; skipping it
```

I haven't looked into what is causing the issue yet.

ethanhs:
See also this Travis log: https://gist.github.com/ethanhs/5169bf3bc1ecee491b1cd9bd3f1ebcc6

miedzinski:
How did you repro this? I did

```bash
until ! ./runtests.py stubgen; do :; done
```

for 20 minutes and nothing came up. I'm on `064c8d33` commit, Python 3.6.4, macOS 10.3.3.

ethanhs:
It happens ephemerally, and I'm not doing anything to repro at the moment, the errrors that have come up are random. If it was reproducible I would have looked at what was causing it :)

[...]

PR #4672, opened 2018-03-02
Add some debug code for tracking down the stubgen test flake

msullivan:
There is a stubgen test flake (#4635) that is showing up
intermittently, mostly in CI. Add some reporting code so we have a bit
more to work with the next time it happens.


msullivan:
Merged #4751 instead.

PR #4677, opened 2018-03-04
Add coroutines to the cheatsheet

rkday:
It took me a while to find http://mypy.readthedocs.io/en/stable/kinds_of_types.html#typing-async-await in the documentation, and it would have been easier to find if it was mentioned in the cheatsheet - this adds both `@asyncio.coroutine` and `async def` examples to the Python 3 cheatsheet.

I *think* it's right that the YieldType of a Python 3.4 generator-coroutine should always be Any, and the SendType always None, but if there are exceptions let me know and I can tweak the text.

### DIFF
--- a/docs/source/cheat_sheet_py3.rst
+++ b/docs/source/cheat_sheet_py3.rst
@@ -125,8 +125,9 @@ See :ref:`async-and-await` for the full detail on typing coroutines and asynchro
            count -= 1
        return "Blastoff!"
 
-   # mypy currently does not support converting functions into async generators
-   # in Python 3.4, so you need to add a 'yield' to make it typecheck.
+   # mypy currently does not support converting functions into generator-based
+   # coroutines in Python 3.4, so you need to add a 'yield' to make it
+   # typecheck.
    @asyncio.coroutine
    def async1(obj: object) -> Generator[None, None, str]:
        if False:
--- a/docs/source/cheat_sheet_py3.rst
+++ b/docs/source/cheat_sheet_py3.rst
@@ -105,8 +105,10 @@ Python 3 introduces an annotation syntax for function declarations in `PEP 3107 
        
        ...
 
-Coroutines
-**********
+Coroutines and asyncio
+**********************
+
+See :ref:`async-and-await` for the full detail on typing coroutines and asynchronous code.
 
 .. code-block:: python
 
--- a/docs/source/cheat_sheet_py3.rst
+++ b/docs/source/cheat_sheet_py3.rst
@@ -115,8 +115,8 @@ See :ref:`async-and-await` for the full detail on typing coroutines and asynchro
    import asyncio
    from typing import Generator, Any
 
-   # A Python 3.4 coroutine should have a return type of
-   # Generator[Any, None, T], where T is the type it returns.
+   # A generator-based coroutine created with @asyncio.coroutine should have a
+   # return type of Generator[Any, None, T], where T is the type it returns.
    @asyncio.coroutine
    def countdown34(tag: str, count: int) -> Generator[Any, None, str]:
        while count > 0:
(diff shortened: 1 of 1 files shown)


docs/source/cheat_sheet_py3.rst
```diff
@@ -94,6 +94,36 @@ Python 3 introduces an annotation syntax for function declarations in `PEP 3107
            yield i
            i += 1
 
+   # A Python 3.4 coroutine should have a return type of
+   # Generator[Any, None, T], where T is the type it returns.
+   @asyncio.coroutine
```
ethanhs: I think we try to keep the cheat sheet copy/paste runnable. Could you add an import for `asyncio`?

rkday: Sure! I've fixed up the imports, and also fixed the duplicate names that got thrown up when typechecking, so this now executes cleanly and the only error I get is the expected error for `quux` that was already there.

gvanrossum:
Since this is still a somewhat esoteric topic, can you make it a separate section (e.g. named "Couroutines and asyncio") after the "Functions" section? (I.e. right before "puzzled".)

rkday:
Yep, done.

docs/source/cheat_sheet_py3.rst
```diff
@@ -105,6 +105,45 @@ Python 3 introduces an annotation syntax for function declarations in `PEP 3107
...
+   # mypy currently does not support converting functions into coroutines in
+   # Python 3.4, so you need to add a 'yield' to make it typecheck.
+   @asyncio.coroutine
+   def async1(obj: object) -> Generator[None, None, str]:
+       if False:
+           yield
+       return "placeholder"
+
+   # A Python 3.5+ coroutine is typed like a normal function, and doesn't need
+   # a 'yield' to make it typecheck.
```
ethanhs: I don't think the `and doesn't need a 'yield' to make it typecheck` is needed. Coroutines don't need to be generator functions.

docs/source/cheat_sheet_py3.rst
```diff
@@ -105,6 +105,45 @@ Python 3 introduces an annotation syntax for function declarations in `PEP 3107
...
+**********************
+
+See :ref:`async-and-await` for the full detail on typing coroutines and asynchronous code.
+
+.. code-block:: python
+
+   import asyncio
+   from typing import Generator, Any
+
+   # A Python 3.4 coroutine should have a return type of
```
ethanhs: I think it would be better to say a "generator based coroutine" or "async generator" here instead of `Python 3.4 coroutine` which is a bit ambiguous.

docs/source/cheat_sheet_py3.rst
```diff
@@ -105,6 +105,45 @@ Python 3 introduces an annotation syntax for function declarations in `PEP 3107
...
+   @asyncio.coroutine
+   def countdown34(tag: str, count: int) -> Generator[Any, None, str]:
+       while count > 0:
+           print('T-minus {} ({})'.format(count, tag))
+           yield from asyncio.sleep(0.1)
+           count -= 1
+       return "Blastoff!"
+
+   # mypy currently does not support converting functions into coroutines in
+   # Python 3.4, so you need to add a 'yield' to make it typecheck.
```
ethanhs: I think it is important to say "dynamically convert a function into an async generator" as mypy *can* infer explicit generators are generators.

ethanhs:
@rkday if you have time to respond to my remaining comments, it would be appreciated!

rkday:
Sure - how's this?

docs/source/cheat_sheet_py3.rst
```diff
@@ -125,16 +125,15 @@ See :ref:`async-and-await` for the full detail on typing coroutines and asynchro
            count -= 1
        return "Blastoff!"
 
-   # mypy currently does not support converting functions into coroutines in
-   # Python 3.4, so you need to add a 'yield' to make it typecheck.
+   # mypy currently does not support converting functions into async generators
```
JelleZijlstra: Async generators are a different feature that was added in 3.6. Do you mean generator-based coroutines?

ethanhs: Yeah that is my bad. I suggested it and forgot about PEP 525. "generator based coroutine" is probably better.

rkday: Fixed.

PR #4680, opened 2018-03-05
Add support for --junit-xml=FILE to dmypy [re]check

### LINKED ISSUE #4663: Support writing junit.xml in dmypy
Daemon mode doesn't support the `--junit-xml=FILE` flag -- it is silently ignored. This is because the implementation happens at the very end of `main()` which is never called in daemon mode.

We should probably add this flag to `dmypy check` so multiple check calls can easily be made to write different XML files (handy as CI artifacts). And `dmypy [re]start` should special-case the flag and reject it.

gvanrossum:
Refuse a non-empty value for dmypy [re]start.

Also implement blocker logic in dmypy (CompileError === blocker).

Fixes #4663.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -95,6 +95,9 @@ class Server:
             sys.exit("dmypy: start/restart does not accept sources")
         if options.report_dirs:
             sys.exit("dmypy: start/restart cannot generate reports")
+        if options.junit_xml:
+            sys.exit("dmypy: start/restart does not support --junit-xml; "
+                     "pass it to check/recheck instead")
         if not options.incremental:
             sys.exit("dmypy: start/restart should not disable incremental mode")
         if options.quick_and_dirty:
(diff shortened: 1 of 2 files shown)


mypy/dmypy_server.py
```diff
@@ -95,6 +95,9 @@ def __init__(self, flags: List[str]) -> None:
             sys.exit("dmypy: start/restart does not accept sources")
         if options.report_dirs:
             sys.exit("dmypy: start/restart cannot generate reports")
+        if options.junit_xml:
+            sys.exit("dmypy: start/restart does not support --junit-xml; "
```
gvanrossum: Alternatively, we could keep ignoring this flag silently, or print a warning without exiting. Both would be slightly more convenient for the Dropbox use case, but sound like a possible cause for confusion for other users.

msullivan: Probably cleanest to fail on it

PR #4684, opened 2018-03-06
Add pytest cache to the .gitignore

ethanhs:


gvanrossum:
Interesting, I do not see such a folder in my workspace (despite having run pytest many times).

ethanhs:
That is interesting. AIUI, it is created when tests fail, so that it is straight forward to do `--lf`.

PR #4403, opened 2017-12-22
Implement PEP 561 searching

ethanhs:
This is an implementation of PEP 561.

- [x] Test functionality
- [x] Check PEP 561 conformance to [resolution order](https://www.python.org/dev/peps/pep-0561/#type-checker-module-resolution-order)
- [x] ignore errors in these files
- [x] support running with alternate Python executable
- [x] document PEP 561 feature
- [x] document ``--python-executable`` flag
- [x] test ``--python-executable`` flag
- [x] `--python-version` flag sets `--python-executable` if possible.

This branch *should* work as intended and be feature complete, but it is possible that I've overlooked something/made a mistake.
(Picked up from #4278 for simplicity).

Fixes #2625, #1190, #965.

PR #4685, opened 2018-03-06
Add --cache-fine-grained to docs

ethanhs:
This was discovered in #4403. I thought I'd add it while I am splitting changes out of that PR. It was originally introduced in #4526 but not added to the docs it seems.

(the rest of the diff is due to my using an actual output of the help command from master)


JukkaL:
Since this is related to an undocumented feature, a better option could be to remove the option from mypy help output. @msullivan What do you think?

PR #4691, opened 2018-03-07
Fix a handful of fine-grained incremental bugs

msullivan:
This fixes all of the bugs found while running fine-grained incremental on the S repo with every file changed trivially.

### DIFF
--- a/mypy/server/update.py
+++ b/mypy/server/update.py
@@ -117,7 +117,6 @@ Major todo items:
 - Use mypy.fscache to access file system
 """
 
-import time
 import os.path
 from typing import (
     Dict, List, Set, Tuple, Iterable, Union, Optional, Mapping, NamedTuple, Callable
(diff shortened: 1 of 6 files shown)


test-data/unit/fine-grained.test
```diff
@@ -2635,3 +2635,83 @@ def a(f: Callable[[Arg(int, 'x')], int]) -> None: pass
 [builtins fixtures/dict.pyi]
 [out]
 ==
+
+[case testImplicitTuple1]
+import a
+[file a.py]
+# Bogus annotation in nested function masked because
```
JukkaL: Something missing in the comment? "... function masked because [what]"

mypy/server/update.py
```diff
@@ -226,6 +228,11 @@ def update(self, changed_modules: List[Tuple[str, str]]) -> List[str]:
             changed_modules = [(id, path) for id, path in changed_modules
                                if id != next_id]
             changed_modules = dedupe_modules(remaining + changed_modules)
+            t1 = time.time()
+
+            self.manager.log_fine_grained(
```
JukkaL: This seems unrelated to the rest of thePR?

msullivan: Oh, indeed. I added it to track progress while running mypy-daemon on a big changed repo. I can split it out.

Issue #965, opened 2015-10-25
Specifying which Python interpreter stubgen should use


JukkaL:
Add a `-p <path-to-python>` option to stubgen that will replace the `--py2` option. Mypy would use this interpreter for these things:
- Determining Python version (2 or 3).
- Determining Python module search path.
- Runtime introspection (esp. for generating stubs for C modules).

As pointed out by @gnprice, the current behavior of using the current Python interpreter doesn't make sense when using a virtualenv, as the virtualenv would have mypy dependencies instead of the (third party) modules targeted by stubgen. This would also allow using an arbitrary Python 2 virtualenv.

We may also need a way of specifying PYTHONPATH for the target interpreter separately from the PYTHONPATH used to run mypy.

gnprice:
This came to mind again today, in the context not of stubgen but of mypy itself. Someone on IRC had trouble getting mypy to find the code they were type-checking, because they were depending on a virtualenv and so `--use-python-path` didn't help them. If we had a `-p` option, they could use that in the same way for `mypy` as we were discussing for stubgen.

Would it make sense for the same `-p` option to apply to both? Perhaps we should retitle this bug to include both.

JukkaL:
Good point -- retitled.

ethanhs:
#4693 added this for mypy.

hauntsaninja:
stubgen gained support for this at some point in the last four years

PR #4692, opened 2018-03-07
Add --python-executable and --no-infer-executable flags

ethanhs:
This implements a flag that allows users to point to a Python interpreter that mypy can use for type checking. `--no-infer-executable` is named as such because the old name (`--no-site-packages`) was rather misleading.

Fixes https://github.com/python/mypy/issues/965

Carried over from #4403, with some changes (renaming the flag, documentation).

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -360,11 +360,27 @@ Here are some more useful flags:
   updates the cache, but regular incremental mode ignores cache files
   written by quick mode.
 
+- ``--python-executable EXECUTABLE`` will have mypy collect type information
+  from `PEP 561`_ compliant packages installed for the Python executable
+  ``EXECUTABLE``. If not provided, mypy will use PEP 561 compliant packages
+  installed for the Python executable running mypy. See
+  :ref:`installed-packages` for more on making PEP 561 compliant packages. This
+  flag will attempt to set ``--python-version`` if not already set.
+
 - ``--python-version X.Y`` will make mypy typecheck your code as if it were
   run under Python version X.Y. Without this option, mypy will default to using
   whatever version of Python is running mypy. Note that the ``-2`` and
   ``--py2`` flags are aliases for ``--python-version 2.7``. See
-  :ref:`version_and_platform_checks` for more about this feature.
+  :ref:`version_and_platform_checks` for more about this feature. This flag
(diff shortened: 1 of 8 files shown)


ethanhs:
This should be ready for review now. I think the testing I have exercises the new flags well.

docs/source/command_line.rst
```diff
@@ -360,11 +360,21 @@ Here are some more useful flags:
...
 - ``--python-version X.Y`` will make mypy typecheck your code as if it were
   run under Python version X.Y. Without this option, mypy will default to using
   whatever version of Python is running mypy. Note that the ``-2`` and
   ``--py2`` flags are aliases for ``--python-version 2.7``. See
-  :ref:`version_and_platform_checks` for more about this feature.
+  :ref:`version_and_platform_checks` for more about this feature. This flag
+  will attempt to find a Python executable of the corresponding version. If
+  you'd like to disable this, see ``--no-infer-executable`` below.
+
+- ``--no-infer-executable`` will disable searching for a usable Python
```
eric-wieser: I don't think this flag makes any sense without PEP561, and that `--no-site-packages` was a better name in that scenario, as it indicated the purpose of the executable

ethanhs: Hm, I think I might agree in principle with the first (but I split this out for ease of review mostly). 

As for the flag name, I'm not so sure. If the executable is ever used for anything other than PEP 561, the `--no-site-packages` flag would be misleading. It is misleading now anyway, it doesn't actually disable searching, it sets `options.python_executable` to `None`, which happens to mean that searching for PEP 561 packages is not done.

eric-wieser: > it sets options.python_executable to None, which happens to mean that searching for PEP 561 packages is not done.

You've got this backwards. `options.python_executable is None` is an implementation-detail of how we represent "don't do pep561 searching", rather than storing a redundant boolean alongside it. We could add a `@property` to `Options` to make that more explicit, I suppose.

We shouldn't be exposing that detail in the command line arguments.

ethanhs: Hm, I suppose from a user perspective it would be better to name it `--no-site-packages`.

ethanhs:
Closing as this should be part of the main PEP implementation.

PR #4695, opened 2018-03-07
Fix issues related to indirect imports in import cycles

### LINKED ISSUE #4049: Circular import + reexport = "error: Module '...' has no attribute '..."?

### LINKED ISSUE #4429: Legal Python circular imports result in loss of type information
Example:

`test/__init__.py`:
```python
import typing

from . import a
from . import b

if typing.TYPE_CHECKING:
        reveal_type(a.foo())
```

`test/a.py`:
```python
import test

def foo() -> int:
    return test.b.x
```

`test/b.py`:
```python
import test
import typing
x = 42

def run() -> int:
    if typing.TYPE_CHECKING:
        reveal_type(test.a.foo())
    return test.a.foo()
```

`test/__main__.py`
```python
import test

print(test.b.run())
```

Python will execute the above code just fine:

```bash
$ python3 -m test
42
```

But `mypy` loses the type information for `test.a` in the context of `b.py`
```bash
$ mypy test/__init__.py test/a.py test/b.py
test/b.py:7: error: Revealed type is 'Any'
test/__init__.py:7: error: Revealed type is 'builtins.int'
```

What's odd is if I reorder the files to explicitly pass in `b.py` first, things work -

```bash
$ mypy test/b.py test/a.py test/__init__.py
test/__init__.py:7: error: Revealed type is 'builtins.int'
test/b.py:7: error: Revealed type is 'builtins.int'
```

### LINKED ISSUE #4682: Crash when serializing property with forward reference
This serialization test case fails with a crash:

```
[case testSerializeForwardReferenceToAliasInProperty]
import a
[file a.py]
import b
[file a.py.2]
import b
reveal_type(b.A().p)
[file b.py]

class A:
    @property
    def p(self) -> C: pass
    @p.setter
    def p(self, c: C) -> None: pass
    @p.deleter
    def p(self) -> None: pass

C = str
[builtins fixtures/property.pyi]
```

Here's the traceback:

```
/Users/jukka/src/mypy/mypy/test/testcheck.py:105: in run_case
    self.run_case_once(testcase, step)
/Users/jukka/src/mypy/mypy/test/testcheck.py:169: in run_case_once
    alt_lib_path=test_temp_dir)
/Users/jukka/src/mypy/mypy/build.py:180: in build
    result = _build(sources, options, alt_lib_path, bin_dir, saved_cache, flush_errors)
/Users/jukka/src/mypy/mypy/build.py:266: in _build
/Users/jukka/src/mypy/mypy/build.py:2114: in dispatch
    process_graph(graph, manager)
/Users/jukka/src/mypy/mypy/build.py:2421: in process_graph
/Users/jukka/src/mypy/mypy/build.py:2621: in process_stale_scc
    graph[id].write_cache()
/Users/jukka/src/mypy/mypy/build.py:2071: in write_cache
    self.manager)
/Users/jukka/src/mypy/mypy/build.py:1260: in write_cache
    data = {'tree': tree.serialize(),
/Users/jukka/src/mypy/mypy/nodes.py:247: in serialize
    'names': self.names.serialize(self._fullname),
/Users/jukka/src/mypy/mypy/nodes.py:2521: in serialize
    data[key] = value.serialize(fullname, key)
/Users/jukka/src/mypy/mypy/nodes.py:2456: in serialize
    data['node'] = self.node.serialize()
/Users/jukka/src/mypy/mypy/nodes.py:2229: in serialize
    'names': self.names.serialize(self.fullname()),
/Users/jukka/src/mypy/mypy/nodes.py:2521: in serialize
    data[key] = value.serialize(fullname, key)
/Users/jukka/src/mypy/mypy/nodes.py:2458: in serialize
    data['type_override'] = self.type_override.serialize()
/Users/jukka/src/mypy/mypy/types.py:955: in serialize
    'items': [t.serialize() for t in self.items()],
/Users/jukka/src/mypy/mypy/types.py:955: in <listcomp>
    'items': [t.serialize() for t in self.items()],
/Users/jukka/src/mypy/mypy/types.py:870: in serialize
    'ret_type': self.ret_type.serialize(),
/Users/jukka/src/mypy/mypy/types.py:1429: in serialize
    assert False, "Internal error: Unresolved forward reference to {}".format(name)
E   AssertionError: Internal error: Unresolved forward reference to C
```

It looks like we don't remove forward references in the `type_override` attribute of symbol table nodes.

JukkaL:
This adds supports for some cases of importing an imported name
within an import cycle. Originally they could result in false positives 
or false negatives.

The idea is to use a new node type `ImportedName` in semantic
analysis pass 1 to represent an indirect reference to a name in another
module. It will get resolved in semantic analysis pass 2.

`ImportedName` is not yet used everywhere where it could make 
sense and this doesn't fix all related issues with import cycles.

Also did a bit of refactoring of type semantic analysis to avoid passing
multiple callback functions.

Fixes #4049.
Fixes #4429.
Fixes #4682.

Borrowed test cases from #4495 by @carljm. This PR is an alternative 
for #4495.

Supersedes #4495

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1647,8 +1647,8 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
         if i_id in self.modules:
             m = self.modules[i_id]
             self.add_submodules_to_parent_modules(i_id, True)
-            for name, node in m.names.items():
-                node = self.dereference_module_cross_ref(node)
+            for name, orig_node in m.names.items():
+                node = self.dereference_module_cross_ref(orig_node)
                 if node is None:
                     continue
                 new_node = self.normalize_type_alias(node, i)
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -55,7 +55,7 @@ from mypy.nodes import (
     YieldFromExpr, NamedTupleExpr, TypedDictExpr, NonlocalDecl, SymbolNode,
     SetComprehension, DictionaryComprehension, TYPE_ALIAS, TypeAliasExpr,
     YieldExpr, ExecStmt, Argument, BackquoteExpr, ImportBase, AwaitExpr,
-    IntExpr, FloatExpr, UnicodeExpr, EllipsisExpr, TempNode, EnumCallExpr,
+    IntExpr, FloatExpr, UnicodeExpr, EllipsisExpr, TempNode, EnumCallExpr, ImportedName,
     COVARIANT, CONTRAVARIANT, INVARIANT, UNBOUND_IMPORTED, LITERAL_YES, ARG_OPT, nongen_builtins,
     collections_type_aliases, get_member_expr_fullname,
 )
@@ -1488,6 +1488,8 @@ class SemanticAnalyzerPass2(NodeVisitor[None], SemanticAnalyzerPluginInterface):
         module = self.modules.get(import_id)
         for id, as_id in imp.names:
             node = module.names.get(id) if module else None
+            node = self.dereference_module_cross_ref(node)
+
             missing = False
             possible_module_id = import_id + '.' + id
 
@@ -1516,11 +1518,14 @@ class SemanticAnalyzerPass2(NodeVisitor[None], SemanticAnalyzerPluginInterface):
                     ast_node = Var(name, type=typ)
                     symbol = SymbolTableNode(GDEF, ast_node)
                     self.add_symbol(name, symbol, imp)
-                    return
+                    continue
             if node and node.kind != UNBOUND_IMPORTED and not node.module_hidden:
                 node = self.normalize_type_alias(node, imp)
                 if not node:
-                    return
+                    # Normalization failed because target is not defined. Avoid duplicate
+                    # error messages by marking the imported name as unknown.
+                    self.add_unknown_symbol(as_id or id, imp, is_import=True)
+                    continue
                 imported_id = as_id or id
                 existing_symbol = self.globals.get(imported_id)
                 if existing_symbol:
@@ -1550,6 +1555,26 @@ class SemanticAnalyzerPass2(NodeVisitor[None], SemanticAnalyzerPluginInterface):
                 # Missing module.
                 self.add_unknown_symbol(as_id or id, imp, is_import=True)
 
+    def dereference_module_cross_ref(
+            self, node: Optional[SymbolTableNode]) -> Optional[SymbolTableNode]:
+        """Dereference cross references to other modules (if any).
+
+        If the node is not a cross reference, return it unmodified.
+        """
+        seen = set()  # type: Set[str]
+        # Continue until we reach a node that's nota cross reference (or until we find
+        # nothing).
+        while node and isinstance(node.node, ImportedName):
+            fullname = node.node.fullname()
+            if fullname in seen:
+                # Looks like a reference cycle. Just break it.
+                # TODO: Generate a more specific error message.
+                node = None
+                break
+            node = self.lookup_fully_qualified_or_none(fullname)
+            seen.add(fullname)
+        return node
+
     def process_import_over_existing_name(self,
                                           imported_id: str, existing_symbol: SymbolTableNode,
                                           module_symbol: SymbolTableNode,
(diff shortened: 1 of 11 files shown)


mypy/semanal.py
```diff
@@ -1573,6 +1603,14 @@ def process_import_over_existing_name(self,
 
     def normalize_type_alias(self, node: SymbolTableNode,
                              ctx: Context) -> Optional[SymbolTableNode]:
+        """If node refers to a built-intype alias, normalize it.
```
ilevkivskyi: Typo `buit-intype` (space mising).

mypy/semanal.py
```diff
@@ -3873,8 +3910,10 @@ def add_symbol(self, name: str, node: SymbolTableNode,
             self.type.names[name] = node
         else:
             existing = self.globals.get(name)
-            if existing and (not isinstance(node.node, MypyFile) or
-                             existing.node != node.node) and existing.kind != UNBOUND_IMPORTED:
+            if (existing
+                    and (not isinstance(node.node, MypyFile) or existing.node != node.node)
+                    and existing.kind != UNBOUND_IMPORTED
```
ilevkivskyi: I would expect that the goal is to get rid of `UNBOUND_IMPORTED` and replace all such checks with `isinstance(node.node, ImportedName)`, what I am missing here?

mypy/semanal_pass3.py
```diff
@@ -431,12 +438,28 @@ def check_for_omitted_generics(self, typ: Type) -> None:
...
+    def fail(self, msg: str, ctx: Context, serious: bool = False, *,
+             blocker: bool = False) -> None:
+        # TODO: Process all arguments
         self.errors.report(ctx.get_line(), ctx.get_column(), msg)
 
     def fail_blocker(self, msg: str, ctx: Context) -> None:
         self.fail(msg, ctx, blocker=True)
 
+    def note(self, msg: str, ctx: Context) -> None:
+        self.sem.note(msg, ctx)
```
ilevkivskyi: What is the reason we can reuse `self.sem.note` here but not `self.sem.fail` above?

mypy/types.py
```diff
@@ -284,6 +284,8 @@ class TypeOfAny(Enum):
     special_form = 'special_form'
     # Does this Any come from interaction with another Any?
     from_another_any = 'from_another_any'
+    # Does this Any come from an implementation limitation/bug?
+    implementation_artifact = 'implementation_artifact'
```
ilevkivskyi: Big +1 for adding this. I think I have seen other cases where this would be appropriate (but can't remember them now).

test-data/unit/check-modules.test
```diff
@@ -1955,3 +1957,217 @@ import c
 tmp/b.py:1: error: Cannot find module named 'c'
 main:1: error: Cannot find module named 'c'
 main:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
+
+[case testIndirectFromImportWithinCycle]
+import a
+[file a.py]
+from b import f
```
ilevkivskyi: Would it make sense to check that the result is independent on the order of import statements (just in case) here and/or in other tests?

ilevkivskyi:
(I updated the description to mention that this also fixes https://github.com/python/mypy/issues/4682 since I found the relevant code and tests.)

mypy/semanal.py
JukkaL: That's the end goal. This PR doesn't remove all instance of `UNBOUND_IMPORTED`. In particular, `import m` statements generate them still. I'll fix them in a separate PR.

ilevkivskyi: > I'll fix them in a separate PR.

OK, having smaller PRs makes sense.

mypy/semanal_pass3.py
JukkaL: These were different originally, but it seems like there's no good reason. I'll unify them.

test-data/unit/check-modules.test
JukkaL: I'll add a few test cases where we reverse the shape of the import cycle relative to the main module.

carljm:
Confirmed that this fixes the problems in our codebase. Thanks @JukkaL and @ilevkivskyi !

PR #4701, opened 2018-03-08
Improve performance of deletions in initial fine-grained cache load

msullivan:
To do this we just fully ditch `process_graph` for doing initial cache loads, since we don't want to do anything but load the graph and trees from the cache.

This has the side effect of disabling any writes to the cache while doing a successful cache load, so for consistency we also disable cache writes when a fine-grained cache load fails and we give up on it.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -571,7 +571,9 @@ class BuildManager:
       flush_errors:    A function for processing errors after each SCC
       saved_cache:     Dict with saved cache state for coarse-grained dmypy
                        (read-write!)
-      cache_enabled:   Whether cache usage is enabled
+      cache_enabled:   Whether cache usage is enabled. This is set based on options,
+                       but is disabled if fine-grained cache loading fails
+                       and after an initial fine-grained load.
       stats:           Dict with various instrumentation numbers
     """
 
@@ -609,7 +611,8 @@ class BuildManager:
         self.rechecked_modules = set()  # type: Set[str]
         self.plugin = plugin
         self.flush_errors = flush_errors
-        self.cache_enabled = options.incremental and options.cache_dir != os.devnull
+        self.cache_enabled = options.incremental and (
+            not options.fine_grained_incremental or options.use_fine_grained_cache)
         self.saved_cache = saved_cache if saved_cache is not None else {}  # type: SavedCache
         self.stats = {}  # type: Dict[str, Any]  # Values are ints or floats
 
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -588,7 +588,6 @@ class BuildManager:
         self.data_dir = data_dir
         self.errors = errors
         self.errors.set_ignore_prefix(ignore_prefix)
-        self.only_load_from_cache = options.use_fine_grained_cache
         self.lib_path = tuple(lib_path)
         self.source_set = source_set
         self.reports = reports
@@ -1675,10 +1674,10 @@ class State:
                                  for id, line in zip(self.meta.dependencies, self.meta.dep_lines)}
             self.child_modules = set(self.meta.child_modules)
         else:
-            # In fine-grained cache mode, pretend we only know about modules that
-            # have cache information and defer handling new modules until the
-            # fine-grained update.
-            if manager.only_load_from_cache:
+            # When doing a fine-grained cache load, pretend we only
+            # know about modules that have cache information and defer
+            # handling new modules until the fine-grained update.
+            if manager.options.use_fine_grained_cache:
                 manager.log("Deferring module to fine-grained update %s (%s)" % (path, id))
                 raise ModuleNotFound
 
@@ -1799,7 +1798,7 @@ class State:
         # cache load because we need to gracefully handle missing modules.
         fixup_module_pass_one(self.tree, self.manager.modules,
                               self.manager.options.quick_and_dirty or
-                              self.manager.only_load_from_cache)
+                              self.manager.options.use_fine_grained_cache)
 
     def calculate_mros(self) -> None:
         assert self.tree is not None, "Internal error: method must be called on parsed file only"
@@ -2107,9 +2106,13 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
     # This is a kind of unfortunate hack to work around some of fine-grained's
     # fragility: if we have loaded less than 50% of the specified files from
     # cache in fine-grained cache mode, load the graph again honestly.
+    # In this case, we just turn the cache off entirely, so we don't need
+    # to worry about some files being loaded and some from cache and so
+    # that fine-grained mode never *writes* to the cache.
     if manager.options.use_fine_grained_cache and len(graph) < 0.50 * len(sources):
-        manager.log("Redoing load_graph because too much was missing")
-        manager.only_load_from_cache = False
+        manager.log("Redoing load_graph without cache because too much was missing")
+        manager.options.use_fine_grained_cache = False
+        manager.options.cache_dir = os.devnull
         graph = load_graph(sources, manager)
 
     t1 = time.time()
@@ -2130,8 +2133,8 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
     if manager.options.dump_graph:
         dump_graph(graph)
         return graph
-    if manager.only_load_from_cache:
-        halfass_process_graph(graph, manager)
+    if manager.options.use_fine_grained_cache:
+        process_fine_grained_cache_graph(graph, manager)
     else:
         process_graph(graph, manager)
     updated = preserve_cache(graph)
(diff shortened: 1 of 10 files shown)


mypy/dmypy_server.py
```diff
@@ -280,6 +280,9 @@ def initialize_fine_grained(self, sources: List[mypy.build.BuildSource]) -> Dict
             result = mypy.build.build(sources=sources,
                                       options=self.options,
                                       alt_lib_path=self.alt_lib_path)
+            # build will clear use_fine_grained_cache if it needs to give up
```
JukkaL: Style nit: looks like it would be better to move this outside the try statement, since this can't generate a `CompileError` exception.

mypy/server/update.py
```diff
@@ -174,8 +174,8 @@ def __init__(self,
         # for the cache. This is kind of a hack and it might be better to have
         # this directly reflected in load_graph's interface.
         self.options.cache_dir = os.devnull
+        self.options.use_fine_grained_cache = False
```
JukkaL: Modifying the options object it's kind of ugly. Maybe at least mention that it gets mutated in the docstring?

mypy/dmypy_server.py
```diff
@@ -280,6 +280,9 @@ def initialize_fine_grained(self, sources: List[mypy.build.BuildSource]) -> Dict
...
+            # on doing a cache load.
+            cache_load_succeeded = self.options.use_fine_grained_cache
```
JukkaL: Instead of mutating the `use_fine_grained_cache` attribute, would it be cleaner to add another attribute to the build result? Having a side channel like this seems questionable.

mypy/build.py
```diff
@@ -2456,6 +2456,25 @@ def process_graph(graph: Graph, manager: BuildManager) -> None:
...
 
 
+def process_fine_grained_cache_graph(graph: Graph, manager: BuildManager) -> None:
+    """Finish loading everything for use in the fine-grained incremental cache"""
+
+    # If we are running in fine-grained incremental mode with caching,
+    # we process all SCCs as fresh SCCs so that we have all of the symbol
+    # tables and fine-grained dependencies available.
+    # We fail the loading of any SCC that we can't load a meta for, so we
+    # don't have anything *but* fresh SCCs.
```
JukkaL: Have you tested what happens if we fail loading some SCCs? Will they be picked up by the build later?

mypy/build.py
```diff
@@ -2105,9 +2106,13 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
     # This is a kind of unfortunate hack to work around some of fine-grained's
     # fragility: if we have loaded less than 50% of the specified files from
     # cache in fine-grained cache mode, load the graph again honestly.
+    # In this case, we just turn the cache off entirely, so we don't need
+    # to worry about some files being loaded and some from cache and so
+    # that fine-grained mode never *writes* to the cache.
     if manager.options.use_fine_grained_cache and len(graph) < 0.50 * len(sources):
-        manager.log("Redoing load_graph because too much was missing")
-        manager.only_load_from_cache = False
+        manager.log("Redoing load_graph without cache because too much was missing")
```
JukkaL: Did you test that this still gets triggered correctly? What about setting a flag when we get here and having a test case where we test the existence of the flag (enabled using a magic comment in test description, for example)? The test could also delete some subset of cache files (maybe based on module name or something).

mypy/build.py
```diff
@@ -2128,7 +2133,10 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
     if manager.options.dump_graph:
         dump_graph(graph)
         return graph
-    process_graph(graph, manager)
+    if manager.options.use_fine_grained_cache:
```
JukkaL: Add comment here describing why we have two different ways of processing the cache.

test-data/unit/fine-grained-modules.test
```diff
@@ -624,6 +624,22 @@ c.f(1)
 ==
 a.py:2: error: Too many arguments for "f"
 
+[case testRenameAndDeleteModule]
```
JukkaL: Was this test case failing on master? Do we have sufficient test coverage? The details of loading the import graph are pretty subtle. Maybe we could, say, keep track of the number of files parsed during `load_graph` and test that it is zero in certain cases when using the cache (and non-zero when not using the cache)?

msullivan: It wasn't failing, no. I ran this test with verbosity on to track that the right thing was happening. Which, of course, is a pretty good sign that there should be some automated scheme.

mypy/build.py
msullivan: Yeah, that is tested by the `testAddModuleAfterCache` tests.

msullivan:
I added some test hooks that I think should cover things, and fixed a bug in the fallback to doing a full load that it revealed.

test-data/unit/fine-grained.test
```diff
@@ -15,6 +15,11 @@
 --   <optional output from batch pass>
 --   ==
 --   <optional output from first incremental pass>
+--
+--
+-- Modules that are expected to be detected as changed can be checked with [stale ...]
```
JukkaL: Please make the description more detailed. Say, does changed mean that source code is changed or the file is added/deleted, or only some of these? How does it differ for the first run and successive runs, and what about tests using cache or not using cache?

Also mention `[stale2` etc.

test-data/unit/fine-grained.test
```diff
@@ -15,6 +15,11 @@
...
+-- while modules that are reprocessed by update (which can include cached files
+-- that need to be loaded) can be checked with [rechecked ...]
```
JukkaL: Make this more explicit -- here reprocessing only means full module reprocessing, not that we process only some target in the module when following fine-grained dependencies.

test-data/unit/fine-grained-modules.test
```diff
@@ -1080,18 +1115,64 @@ import a
...
+[file a.py]
+from b import f
+f()
+[file b.py]
+def f() -> None: pass
+[delete b.py.2]
+
+-- in no cache mode, there is no way to know about b yet,
+-- but a should still get triggered
+[stale a]
```
JukkaL: I don't fully understand why `a` is included here, since only `b` had file changes (it was deleted). Please explain in the comment why `a` is triggered.

test-data/unit/fine-grained-modules.test
```diff
@@ -975,10 +975,10 @@ x = Foo()
 main:2: error: Too few arguments for "foo" of "Foo"
 
 -- This series of tests is designed to test adding a new module that
--- doesn't appear in the cache, for cache mode. They aren't run only
--- in cache mode, though, because they are still perfectly good
--- regular tests.
-[case testAddModuleAfterCache1]
+-- doesn't appear in the cache, for cache mode. They are run in
+-- cache mode oly because stale and rechecked differ heavily between
```
JukkaL: Typo: `oly`

test-data/unit/fine-grained-modules.test
```diff
@@ -1008,30 +1016,51 @@ import b
...
 [file b.py.2]
 def foo() -> None: pass
+
+[stale b]
+[rechecked a, b]
+
 [file b.py.3]
 def foo(x: int) -> None: pass
+
+[stale2 b]
```
JukkaL: Include `[rechecked2 ...`?

test-data/unit/fine-grained-modules.test
```diff
@@ -1008,30 +1016,51 @@ import b
...
+def foo() -> None: pass
 [file c.py.2]
 [file d.py.2]
 [file e.py.2]
 [file f.py.2]
 [file g.py.2]
+
+-- No files should be stale or reprocessed in the first step since the large number
+-- of missing files will force build to give up on cache loading.
+[stale]
```
JukkaL: Should we check for `[rechecked` here as well?

test-data/unit/fine-grained-modules.test
```diff
@@ -1008,30 +1016,51 @@ import b
...
+
+[file b.py.3]
+def foo(x: int) -> None: pass
+[stale2 b]
```
JukkaL: Similar to above.

test-data/unit/fine-grained-modules.test
```diff
@@ -1064,14 +1093,20 @@ def foo() -> None: pass
 [file b.py.2]
 import a
 a.foo(10)
+
+[stale a, b]
```
JukkaL: Here and various other places below we don't check for `[rechecked`. Is this intentional?

mypy/build.py
```diff
@@ -569,6 +571,7 @@ class BuildManager:
       flush_errors:    A function for processing errors after each SCC
       saved_cache:     Dict with saved cache state for coarse-grained dmypy
                        (read-write!)
+      cache_enabled:   Whether cache usage is enabled
```
JukkaL: Could be worth mentioning how this relates to `use_fine_grained_cache` -- the latter means that we attempt to use the cache if it exists, and this indicates whether we actually used it or something.

test-data/unit/fine-grained-modules.test
msullivan: When there is no `rechecked`, the value of `stale` is used. This is behavior inherited from the regular incremental test suite. I can document this better in the block comment or just add all the recheckeds in.

test-data/unit/fine-grained-modules.test
msullivan: The initial cache mode run picks up changes to files that are not "fresh", in order to pick up modules with changed dependencies.

PR #4705, opened 2018-03-08
Fine-grained: Support "from m import *"

### LINKED ISSUE #4668: Fine-grained: Support "from m import *"
`from m import *` is only partially supported in fine-grained incremental mode.

### LINKED ISSUE #4671: Fine-grained: Error fixed in later steps may persist
I suspect that if we refresh the same target multiple times in a single fine-grained update, errors from the initial refreshes may persist in the generated errors even if those errors are later fixed. To fix this, we should remove any generated errors for a target before refreshing it. We may want to have errors indexed by the target to make this more efficient in case there are a large number of errors.

I have a test case that depends on unrelated local changes that may reproduce this, but it's still unclear. I'm creating this issue now to avoid duplicate work in case somebody else encounters this while I'm still working on this.

JukkaL:
This a adds a new per-module "wildcard" trigger that gets triggered if 
anything gets changed in module top-level. This allows tracking
"from m import *" dependencies which depend on every name in the
target module.

Fixes #4668.

Also fixed #4671 since it was triggered by some of the new tests.

### DIFF
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -352,8 +352,12 @@ class Errors:
     def clear_errors_in_targets(self, path: str, targets: Set[str]) -> None:
         """Remove errors in specific fine-grained targets within a file."""
         if path in self.error_info_map:
-            new_errors = [info for info in self.error_info_map[path]
-                          if info.target not in targets]
+            new_errors = []
+            for info in self.error_info_map[path]:
+                if info.target not in targets:
+                    new_errors.append(info)
+                elif info.only_once:
+                    self.only_once_messages.remove(info.message)
             self.error_info_map[path] = new_errors
 
     def generate_unused_ignore_notes(self, file: str) -> None:
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -21,7 +21,7 @@ class ErrorInfo:
     # related to this error. Each item is a (path, line number) tuple.
     import_ctx = None  # type: List[Tuple[str, int]]
 
-    # The source file that was the source of this error.
+    # The path to source file that was the source of this error.
     file = ''
 
     # The fully-qualified id of the source module for this error.
@@ -51,7 +51,7 @@ class ErrorInfo:
     # Only report this particular messages once per program.
     only_once = False
 
-    # Actual origin of the error message
+    # Actual origin of the error message as tuple (path, line number)
     origin = None  # type: Tuple[str, int]
 
     # Fine-grained incremental target where this was reported
@@ -348,6 +348,13 @@ class Errors:
                 return
             self.only_once_messages.add(info.message)
         self._add_error_info(file, info)
+
+    def clear_errors_in_targets(self, path: str, targets: Set[str]) -> None:
+        """Remove errors in specific fine-grained targets within a file."""
+        if path in self.error_info_map:
+            new_errors = [info for info in self.error_info_map[path]
+                          if info.target not in targets]
+            self.error_info_map[path] = new_errors
 
     def generate_unused_ignore_notes(self, file: str) -> None:
         ignored_lines = self.ignored_lines[file]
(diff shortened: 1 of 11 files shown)


JukkaL:
The test failure is the known stubgen flake.

mypy/errors.py
```diff
@@ -349,6 +349,17 @@ def add_error_info(self, info: ErrorInfo) -> None:
             self.only_once_messages.add(info.message)
         self._add_error_info(file, info)
 
+    def clear_errors_in_targets(self, path: str, targets: Set[str]) -> None:
```
msullivan: To try to improve my own understanding of this issue: which of the tests was breaking?

mypy/server/aststrip.py
```diff
@@ -196,6 +199,14 @@ def visit_import(self, node: Import) -> None:
                     symnode.kind = UNBOUND_IMPORTED
                     symnode.node = None
 
+    def visit_import_all(self, node: ImportAll) -> None:
```
msullivan: You might need an `is_unreachable` check here to avoid running into trouble with imports in `if MYPY` else blocks. (I needed to add such a check to `visit_import`  and `visit_import_from`.)

mypy/server/trigger.py
```diff
@@ -1,5 +1,17 @@
...
+# Used as a suffix for triggers to handle "from m import *" dependencies (see also
+# make_wildcard_trigger)
+WILDCARD_TAG = '[wildcard]'
+
 
 def make_trigger(name: str) -> str:
     return '<%s>' % name
+
+
+def make_wildcard_trigger(module: str) -> str:
```
msullivan: Could you add some comments about the semantic difference between module triggers and module wildcard triggers?

mypy/semanal.py
```diff
@@ -1659,7 +1659,7 @@ def visit_import_all(self, i: ImportAll) -> None:
                 # False, and we can skip checking '_' because it's been explicitly included.
                 if (new_node and new_node.module_public and
                         (not name.startswith('_') or '__all__' in m.names)):
-                    existing_symbol = self.globals.get(name)
+                    existing_symbol = self.lookup_current_scope(name)
```
msullivan: What's the story with this? Was there an existing bug here that just got exposed by one of these test cases, or is this only necessary with fine-grained mode?

mypy/server/update.py
```diff
@@ -594,7 +594,16 @@ def calculate_active_triggers(manager: BuildManager,
             names.add(id)
         else:
             snapshot2 = snapshot_symbol_table(id, new.names)
-        names |= compare_symbol_table_snapshots(id, snapshot1, snapshot2)
+        diff = compare_symbol_table_snapshots(id, snapshot1, snapshot2)
+        package_nesting_level = id.count('.')
+        for item in diff:
+            if (item.count('.') <= package_nesting_level + 1
+                    and not item.split('.')[-1].startswith('_')):
```
ilevkivskyi: Shouldn't we treat `.pyi` files specially here? For example ignore top level names imported from other modules (unless imported using `... as ...`).

mypy/errors.py
JukkaL: `testImportStarAddMissingDependency1` and `testImportStarAddMissingDependencyInsidePackage1` would fail without this.

mypy/server/aststrip.py
JukkaL: Good point, will add.

mypy/server/trigger.py
JukkaL: Will add.

mypy/semanal.py
JukkaL: I believe that there was an existing bug, but it could be a bigger problem in fine-grained incremental mode.

mypy/server/update.py
JukkaL: That would improve efficiency a bit but since `.pyi` files are usually shipped as part of typeshed, they don't change very frequently and this is a minor problem. Added a comment about this.

PR #4716, opened 2018-03-10
stubgenc: Generate PEP-526 style variable annotations

gclinch:
This makes the output from stubgen consistent between pure python modules and c modules.


gvanrossum:
LGTM, though I'm shocked there are no tests affected by this change. :-)

gvanrossum:
Thanks!

PR #4718, opened 2018-03-11
Inferred subclass override of a ClassVar is also a ClassVar. Fixes #4715.

### LINKED ISSUE #4715: False positive error on subclass overriding a ClassVar
This seems to be a bug.

With the following code:

```
from typing import ClassVar

class Base:
    name: ClassVar[str]

class Sub(Base):
    name = "Sub"
```

mypy complains that `:7: error: Cannot override class variable (previously declared on base class "Base") with instance variable`.

But the assignment on the subclass is clearly at the class level. Assigning a value to a class attribute that has already been declared as one should not result in an error. I'd expect this error only if the assignment took place e.g. inside `__init__`.

Explicitly annotating the subclass assignment (`name: ClassVar[str] = "Sub"`) gets rid of the error, but that repeated annotation should not be necessary.

Reproduced on mypy 0.580-dev-9374acf2d808905d392b2ec8ea52bfce03f9ecad under Python 3.6.4, with no flags or options.

I realize that inability to distinguish instance vs class attributes is a known mypy limitation (#1097), but in general it just results in false negatives. This seems worth a distinct issue since it appears that mypy is attempting to make the distinction, but doing so incorrectly, resulting in a false positive.

carljm:
Fixes #4715. Notes:

1. It seems to me that many Vars which have explicit type annotation are currently wrongly marked as `is_inferred`, thus the fix in `semanal.py`. This fix doesn't cause any test failures, but it makes me wonder if I missed the intended semantics of `is_inferred`?
2. If that fix is correct, it looks likely that the same line should be added in other branches of that method, but that seems out of scope for this PR; the fix is only strictly needed for class attributes in this PR.
3. I put the fixup in pass 3 because it seemed necessary in order to ensure populated MRO. If that's wrong and this could be set correctly reliably in pass 2, let me know how and I'm happy to update.
4. The three-layer testcase is needed in order to demonstrate why my first fix attempt here was inadequate. My first try was just to check `is_inferred` in `checker.py` and not raise the override error for a type-inferred assignment, but the three-layer testcase still fails with that fix; we need to actually update the value of `is_classvar` on the node so checks work correctly on further subclassing.

Thanks in advance for review!

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1995,6 +1995,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
                 v = Var(lval.name)
                 v.info = self.type
                 v.is_initialized_in_class = True
+                v.is_inferred = not explicit_type
                 v.set_line(lval)
                 v._fullname = self.qualified_name(lval.name)
                 lval.node = v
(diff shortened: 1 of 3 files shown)


carljm:
Ping? Any concerns here, or anything I can address?

JukkaL:
I can take this.

PR #4721, opened 2018-03-12
Refuse to run under 3.5.0 or anything before 3.4

### LINKED ISSUE #4714: Python 3.5.0 is not supported to run mypy
While in general we support running mypy with Python 3.4 and higher, 3.5.0 specifically is not supported (we just found this out rather painfully). This is because the 3.5.0 stdlib has a very old version of typing.py that does not support all the features mypy uses. Higher versions of Python 3.6 (i.e. 3.5.1 and up) work fine. 3.4.x works too because there typing.py is taken from PyPI.

We should probably document this and/or check for it when mypy starts. Currently the only runtime version check I can find is in setup.py, and I'm not convinced that it always triggers (setup.py isn't run when installing via wheels).

Note that our Travis-CI setup explicitly tests with Python 3.5.1 -- this keeps us honest (while I'm not 100% sure 3.6.0 works, at least its typing.py is newer than the one from 3.5.1, so I expect it's fine).

gvanrossum:
Fixes #4714

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -51,6 +51,13 @@ def main(script_path: Optional[str], args: Optional[List[str]] = None) -> None:
         args: Custom command-line arguments.  If not given, sys.argv[1:] will
         be used.
     """
+    # Check for known bad Python versions.
+    vi = tuple(sys.version_info)
+    if vi[:2] < (3, 4):
+        sys.exit("Running mypy with Python 3.3 or lower is not supported; please upgrade")
+    if vi[:3] == (3, 5, 0):
+        sys.exit("Running mypy with Python 3.5.0 is not supported; please upgrade")
+
     t0 = time.time()
     # To log stat() calls: os.stat = stat_proxy
     if script_path:
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -52,11 +52,12 @@ def main(script_path: Optional[str], args: Optional[List[str]] = None) -> None:
         be used.
     """
     # Check for known bad Python versions.
-    vi = tuple(sys.version_info)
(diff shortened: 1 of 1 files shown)


mypy/main.py
```diff
@@ -51,6 +51,13 @@ def main(script_path: Optional[str], args: Optional[List[str]] = None) -> None:
         args: Custom command-line arguments.  If not given, sys.argv[1:] will
         be used.
     """
+    # Check for known bad Python versions.
+    vi = tuple(sys.version_info)
+    if vi[:2] < (3, 4):
+        sys.exit("Running mypy with Python 3.3 or lower is not supported; please upgrade")
+    if vi[:3] == (3, 5, 0):
+        sys.exit("Running mypy with Python 3.5.0 is not supported; please upgrade")
```
ethanhs: I think it is worth it to specify that upgrading to a later version of `3.5` is possible. It isn't hard for me to imagine someone seeing this and thinking that they must use `3.6.x` or newer to use mypy, which is not the impression we want. Perhaps `... please upgrade to Python 3.5.1 or newer`?

mypy/main.py
```diff
@@ -51,6 +51,13 @@ def main(script_path: Optional[str], args: Optional[List[str]] = None) -> None:
         args: Custom command-line arguments.  If not given, sys.argv[1:] will
         be used.
     """
+    # Check for known bad Python versions.
+    vi = tuple(sys.version_info)
```
ethanhs: `sys.version_info` is already a tuple, so I think the `tuple` call is unnecessary.

gvanrossum:
OK, how about now?

PR #4725, opened 2018-03-13
Fine-grained: Fix crashes due to stale dependencies

### LINKED ISSUE #4723: Fine-grained: Crash when refreshing variable based on stale dependency
This fine-grained incremental mode test case causes a crash:

```
[case testChangeFunctionToVariableAndRefreshUsingStaleDependency]
import a
import c
[file a.py]
import c
def f() -> c.A: pass
[file a.py.2]
f = 1
[file c.py]
class A: pass
[file c.py.3]
[out]
==
```

Here's a traceback:

```
________________________ testChangeFunctionToVariableAndRefreshUsingStaleDependency _________________________
data: /Users/jukka/src/mypy/test-data/unit/fine-grained.test:2836:
/Users/jukka/src/mypy/mypy/test/testfinegrained.py:112: in run_case
    new_messages = self.run_check(server, sources)
/Users/jukka/src/mypy/mypy/test/testfinegrained.py:172: in run_check
    response = server.check(sources)
/Users/jukka/src/mypy/mypy/dmypy_server.py:244: in check
    return self.check_fine_grained(sources)
/Users/jukka/src/mypy/mypy/dmypy_server.py:283: in check_fine_grained
    return self.fine_grained_increment(sources)
/Users/jukka/src/mypy/mypy/dmypy_server.py:352: in fine_grained_increment
    messages = self.fine_grained_manager.update(changed, removed)
/Users/jukka/src/mypy/mypy/server/update.py:239: in update
    result = self.update_single(next_id, next_path, next_id in removed_set)
/Users/jukka/src/mypy/mypy/server/update.py:318: in update_single
    self.previous_targets_with_errors)
/Users/jukka/src/mypy/mypy/server/update.py:694: in propagate_changes_using_dependencies
    manager.modules, up_to_date_modules)
/Users/jukka/src/mypy/mypy/server/update.py:761: in find_targets_recursive
    deferred = lookup_target(modules, target)
/Users/jukka/src/mypy/mypy/server/update.py:944: in lookup_target
    OverloadedFuncDef)), 'unexpected type: %s' % type(node)
E   AssertionError: unexpected type: <class 'mypy.nodes.Var'>
```

The reason for the crash is that we are trying to refresh a function based on a stale dependency, but the function has been changed into a variable. The daemon doesn't expect to see a variable and dies. Instead, we should probably just ignore the variable node. Removing stale dependencies is probably not worth it.

The test case is derived from a real-world example.

JukkaL:
We can have a dependency pointing to a target that is no longer valid,
and this could trigger various assertions, causing a crash. Now we just
filter out bad targets to avoid crashes.

Fixes #4723 and other related issues.


mypy/nodes.py
```diff
@@ -458,7 +458,7 @@ def set_line(self, target: Union[Context, int], column: Optional[int] = None) ->
 
 
 class FuncItem(FuncBase):
-    arguments = []  # type: List[Argument]
+    arguments = []  # type: List[Argument]  # Note: Can be None if deserialized (type is a lie!)
```
msullivan: Unfortunate.

test-data/unit/fine-grained.test
```diff
@@ -2832,3 +2832,148 @@ class C(Enum):
 ==
 a.py:4: error: Argument 1 to "f" has incompatible type "C"; expected "int"
 a.py:5: error: Argument 1 to "f" has incompatible type "C"; expected "int"
+
```
msullivan: Which tests hit partially deserialized nodes?

Issue #4690, opened 2018-03-07
Disallow locally-generated types from escaping their function


msullivan:
We should ban locally generated types (such as NamedTuples declared inside of functions and <callable subtype of Whatever> types generated by uses of `callable`) from escaping the function where they are created. 

It isn't really a good idea to do in the first place and it has difficult interactions with both coarse and finen-grained incremental.

gvanrossum:
These may also interact poorly with 'from __future__ import annotations',
so perhaps that's another reason to ban them?

PR #4728, opened 2018-03-13
Fine-grained: Fix crash due to nested class

### LINKED ISSUE #4612: Fine-grained: Nested class causes hang
This fine-grained incremental mode test case hangs:

```
[case testRefreshNestedClassWithSelfReference]
import a
[file a.py]
import b

def f(self) -> None:
    b.y
    class C:
        z: C
[file b.py]
y = 0
[file b.py.2]
y = ''
[out]
==
```

JukkaL:
This works around the problem but doesn't fully resolve the underlying
issue.

Fixes #4612.


msullivan:
Proper fix is probably  implementing  #4690, but this seems fine as a workaround.

PR #4730, opened 2018-03-13
Fine-grained: Fix a number of fine-grained target tracking bugs, use Scope everywhere

msullivan:


mypy/checker.py
```diff
@@ -131,6 +132,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
     # Helper for type checking expressions
     expr_checker = None  # type: mypy.checkexpr.ExpressionChecker
 
+    tscope = None  # type: TargetScope
```
JukkaL: I'd prefer renaming the `Scope` class in `checker.py` instead of importing `mypy.scope.Scope` as `TargetScope`. Having two classes with the same name (at least in their definition) can be pretty confusing.

mypy/scope.py
```diff
@@ -25,30 +28,34 @@ def current_module_id(self) -> str:
...
         if self.function:
-            target += '.' + self.function.name()
-        return target
+            return self.function.fullname()
+        if self.classes:
+            return self.classes[-1].fullname()
+        return self.module
+
+    def current_type_name(self) -> Optional[str]:
+        """Return the current type name if it exists"""
```
JukkaL: Maybe say "the short name of the current type", as there could be confusion with the full name.

mypy/scope.py
```diff
@@ -25,30 +28,34 @@ def current_module_id(self) -> str:
...
+        return self.classes[-1].name() if self.classes else None
+
+    def current_function_name(self) -> Optional[str]:
+        """Return the current function name if it exists"""
```
JukkaL: Similar to above, maybe explicitly mention that this is the short name.

mypy/errors.py
```diff
@@ -180,9 +164,7 @@ def copy(self) -> 'Errors':
         new.type_name = self.type_name[:]
         new.function_or_member = self.function_or_member[:]
         new.target_module = self.target_module
-        new.target_class = self.target_class
-        new.target_function = self.target_function
-        new.target_ignore_depth = self.target_ignore_depth
+        new.scope = self.scope
```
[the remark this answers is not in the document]
msullivan: I think it makes more sense to have a reference, since making a copy means that using the scope that was passed in doesn't work right any more.

Both are a little dodgy in different situations, though.

Fortunately it seems to not actually matter, as copy only is used while checking expressions, so there won't be any scope changes.

PR #4731, opened 2018-03-13
Fine-grained: Only reprocess errored modules once

msullivan:


mypy/server/update.py
```diff
@@ -679,6 +696,8 @@ def propagate_changes_using_dependencies(
     Returns a list (module id, path) tuples representing modules that contain
     a target that needs to be reprocessed but that has not been parsed yet."""
 
+    t0 = time.time()
```
JukkaL: This seems unrelated to the rest of the PR.

mypy/server/update.py
```diff
@@ -246,20 +246,37 @@ def update(self,
...
+
+            if blocker_messages is not None:
                 self.blocking_error = (next_id, next_path)
                 self.stale = changed_modules
-                return messages
-
-        return messages
+                return blocker_messages
+
+            # It looks like we are done processing everything, so now
```
JukkaL: Style nit: This function is getting uncomfortably long. Can you move the new code (or some of the old code) to another function to keep this more manageable?

mypy/server/update.py
```diff
@@ -246,20 +246,37 @@ def update(self,
...
+            # reprocess all targets with errors. We are careful to
+            # support the possibility that reprocessing an errored module
+            # might trigger loading of a module, but I am not sure
+            # if this can really happen.
+            if not changed_modules:
+                # N.B: We just checked next_id, so manager.errors contains
+                # the errors from it. Thus we consider next_id up to date
+                # when propagating changes from the errored targets.
+                changed_modules = propagate_changes_using_dependencies(
```
JukkaL: Since `update_single` adds encountered targets with errors to `self.previous_targets_with_errors`, does this actually process all new targets with errors a second time? If yes, it seems wasteful -- but it should be easy enough to avoid.

msullivan: No, since we considered `next_id` up to date while propagating the changes.

JukkaL: Gotcha. Maybe update the comment to make this more explicit? I didn't make the connection when first reading the comment.

There is still a case where we seem to do extra work: Assume `m.f` has previous errors. Then we update something else that doesn't fully resolve the errors but triggers `m.f`. Now we refresh `m.f` again because it also had previous errors. This seems like a minor issue and probably not worth trying to fix.

msullivan: Yeah.

I have a WIP patch that addresses that sort of problem by deferring *all* propagation to the end, but it fails one test with a deeply bizarre NamedTuple related issue that I haven't had a chance to look into yet.

mypy/server/update.py
JukkaL: This is still an issue.

msullivan: Moved some of the old code out to a new function.

PR #4733, opened 2018-03-14
Add support for specifying multiple packages on the command line alongside modules

### LINKED ISSUE #4732: Repeated --package arguments are ignored
<s>Running `mypy -m module1 -p package1 -p package2` results in only `module1` being checked.</s>

Running `mypy -p package1 -p package2` results in only `package2` being checked.

Ideally this would check all of them, but it should at least error, rather than silently passing and tricking me into thinking my annotations are valid in `package1`<s> and `package2`</s>

eric-wieser:
Fixes #4732

Let me know how best to add tests for this, if you think they're needed.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -419,16 +419,16 @@ def process_options(args: List[str],
 
     code_group = parser.add_argument_group(title='How to specify the code to type check')
     code_group.add_argument('-m', '--module', action='append', metavar='MODULE',
+                            default=[],
                             dest='special-opts:modules',
                             help="type-check module; can repeat for more modules")
-    # TODO: `mypy -p A -p B` currently silently ignores A
-    # (last option wins).  Perhaps -c, -m and -p could just be
-    # command-line flags that modify how we interpret self.files?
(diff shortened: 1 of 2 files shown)


gvanrossum:
Are you going to fix the error?

eric-wieser:
CI output wasn't showing up when I looked before. I'll take a look later today

eric-wieser:
Alright fixed.

gvanrossum:
Hm, we don't seem to have tests for -m and -p at all... Could you add some? They would go in test-data/unit/cmdline.test.

eric-wieser:
Tests don't run at all locally for my machine, so I'm gonna hijack the CI to do my testing...

(fails with `AssertionError: De-serialization failure: TypeInfo not fixed` when I run `python setup.py test`)

eric-wieser:
Ok, test added

gvanrossum:
Don't use setup.py for tests, see README for instructions.

eric-wieser:
Are you happy with the test I've added here? Seems to pass locally and on travis

Issue #3916, opened 2017-09-04
Syntax for specifying plugins


rmcgibbo:
Currently, to use a plugin, the ini file has to specify the relative or absolute filesystem path to the plugin.

To make distributing `pip install`-able plugins easier, it would be useful to have something like a setuptools entrypoint type syntax too.

ethanhs:
Plugins are still rather experimental, so this would probably need to wait for a more stable plugin API to be implemented.

rmcgibbo:
Making it easier for developers to distribute easy-to-use plugins could help uncover any issues by getting more usage.

chadrik:
Hi all,
I'm afraid that this issue may become a blocker for getting the attrs project fully mypy-compliant. 

Users of mypy will most often want to use the very latest version in order to get new bugfixes and features, but users of attrs will likely have more restrictions that force them to remain on older versions of the library.  That means attrs users who upgrade to newer versions of mypy could (and will) inadvertently introduce a version of the attrs plugin which is not compatible with their version of attrs.  Solutions have been suggested along the lines of creating a new github project that just holds the attrs stub, but the main benefit of that is that it frees up the attrs project to release on its own schedule rather than being pinned to mypy's.  That's a good start, but it does not solve the primary problem of ensuring that attrs version X gets attrs plugin version X.  

I've invited @hynek to comment here, but I'm afraid that the level of inconvenience for both attrs developers and users may ultimately stall the feature entirely ([already 8 months in the making](https://github.com/python-attrs/attrs/pull/238)).  

I'd like to propose a compromise.  It's already possible for third-parties to author plugins and load them via mypy.ini, but it requires a path to a plugin file.  If #4734 (or something like it) is merged, but it is not advertised in the docs or release notes, the current situation changes only slightly.   Once merged, this feature will allow a few projects like attrs, which are at the bleeding edge of static type checking and well aware of the unsupported risks involved, to push the tooling forward, while shouldering the responsibility of making adjustments to our plugin if the API changes.  On our end we can run our stub tests against a matrix of the recent mypy versions, and advertise compatible versions with each release of attrs.  

So let's compare our options here:
- A. leave the plugin bundled here: users of attrs who upgrade mypy will eventually run into errors or bugs due to incompatibilities between the plugin version and their attrs version
- B. bundle the plugin with attrs: users of attrs who upgrade mypy will eventually run into errors due to changes in mypy's internal API

[...]

PR #4734, opened 2018-03-14
Load plugins using pip-installable entry points

chadrik:
Quick first draft to start the conversation about #3916.

This relies on `setuptools` entry_points to register a loader function, and `pkg_resources` to access the loader function.

An example registration would look like:

```python
from setuptools import setup

setup(
    entry_points={
        'mypy.plugins': [
            'my_plugin = my_module.mypy_plugin:get_plugin_class'
        ]
    }
)
```

As with the current config-based approach, the function identified by the entrypoint should take a single string argument which is the mypy version and return a `Plugin` class.

Let me know what you think.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -480,7 +480,7 @@ def load_plugins(options: Options, errors: Errors) -> Plugin:
         errors.report(line, 0, message)
         errors.raise_error()
 
-    custom_plugins = []  # type: List[Plugin]
+    custom_plugin_loaders = []  # type: List[Tuple[str, Any]]
     errors.set_file(options.config_file, None)
     for plugin_path in options.plugins:
         # Plugin paths are relative to the config file location.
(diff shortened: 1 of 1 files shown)


JukkaL:
The plugin API isn't yet at all stable so this feature could result in plugins that only work with a specific (potentially old) version of mypy, which would be a problem as it could slow down adoption of new mypy releases. 

We could partially work around this by having semantic versioning for the plugin API, but for that to make sense we'd need to declare which parts of internal APIs are public and can be relied on by plugins, since plugins get easy access to mypy internals that we have no intention of including in a stable API.

So for this to make sense as a supported feature, we'd likely need at least these things:

* Document the plugin API and explicitly enumerate which parts of internal APIs can be relied on.
* Implement semantic versioning for the plugin API and document the history of changes in the plugin API to make it easier to keep plugins up-to-date.
* Allow plugins to decide which versions of the mypy plugin API they support.

Maintaining a public plugin API would add some extra maintenance work. It's not clear yet when we are ready to commit to something like this.

Another consideration is mypy startup performance -- using `pkgresources` as in this PR seems to add about 100msec to mypy startup on my laptop, which would be unfortunate, as I'd rather like to make mypy start up faster in the long term. I expect that this can be made faster, but it would likely require some extra effort.

So unfortunately it seems a bit too early for a feature like this. For now, if you want to make a plugin easily accessible for mypy users, it needs to be maintained in the mypy repository and shipped with mypy. This clearly is not optimal, since the mypy core team needs to review all plugins.

ethanhs:
I look forward to the day when this is possible, but the API is not ready as laid out by Jukka above, and further more I'd like to see an alternative to `pkg_resources` for something like this ideally (perhaps built on `importlib.resources`?). Until the API is more stable, closing this, though thank you for the proposed changes @chadrik.

JukkaL:
Yeah, unfortunately we are not ready for this yet. Note that my above comment about startup speed should be taken with a grain of salt, since my measurements may be off the mark.

ethanhs:
>Note that my above comment about startup speed should be taken with a grain of salt, since my measurements may be off the mark.

It isn't just you, `pkg_resources` is notorious for being rather slow on import.

```
(mypyenv) [ethanhs@towerofart mypy]$ time python -c "import pkg_resources"

real    0m0.121s
user    0m0.104s
sys     0m0.017s
```

PR #4735, opened 2018-03-14
Fine-grained: Fix AST strip of multiple assignment and multiple lvalues

JukkaL:
These used to generate false positives when refreshed.


JukkaL:
> Was this revealed by the randomized tester?

Yes.

PR #4739, opened 2018-03-14
Get rid of the coarse-grained dmypy

msullivan:
This allows us to entirely remove testdmypy (replaced by testfinegrained for fine-grained mode) and all of the `SavedCache` infrastructure. 

Useful fine-grained dmypy tests are ported over to testfinegrained.

The `--experimental` flag is now superfluous but is still accepted temporarily to avoid needing to tightly synchronize this change with scripts that use mypy-daemon.


mypy/main.py
```diff
@@ -404,6 +404,7 @@ def add_invertible_flag(flag: str,
                         dest='special-opts:no_fast_parser',
                         help=argparse.SUPPRESS)
     if server_options:
+        # TODO: This flag is superfluous and should be removed after a short transition
```
JukkaL: Add current date to the comment so that if somebody sees this comment a year from now they will immediately know that the flag can be removed :-)

test-data/unit/check-dmypy-fine-grained.test
```diff
@@ -1,214 +0,0 @@
...
-# cmd: mypy -m a nonexistent_stub
-# NOTE: 'nonexistent_stub' is a magic module name understood by mypy.test.testdmypy
-# TODO: Should generate an error for missing file
-[file a.py]
-[file nonexistent_stub.pyi]
-[delete nonexistent_stub.pyi.2]
-[out1]
-[out2]
-
-[case testPartialNoneTypeFineGrainedIncremental]
```
JukkaL: Is this test case no longer needed? This wasn't migrated.

JukkaL:
There also is a merge conflict.

test-data/unit/check-dmypy-fine-grained.test
msullivan: There were already equivalent tests

PR #4743, opened 2018-03-15
Better support for type variables in fine grained mode

### LINKED ISSUE #4393: Support type variable with value restriction in fine-grained incremental mode

ilevkivskyi:
This adds support for type variables with value restrictions and upper bounds in fine grained mode.
This also fixes several related and unrelated (triggered by new tests) issues.

I try not to add too many new dependencies. In particular, this is done by making class type variables included fully in the class snapshot instead of just their names.

This also adds missing test cases.

Fixes #4393


test-data/unit/deps-generics.test
```diff
@@ -140,5 +140,24 @@ def f(x: T) -> None:
 [out]
 <m.A.__add__> -> m.f
 <m.A.f> -> m.f
-<m.A> -> <m.T>, m.A
+<m.A> -> <m.T>, <m.f>, m.A, m.f
+<m.T> -> <m.f>, m.f
+
+[case testTypeVarValues]
```
JukkaL: Also test a generic class where the class type variable has values and the type variable is used in a method.

ilevkivskyi:
Thanks for review! I added the tests requested and a clarifying comment that we discussed recently.

PR #4751, opened 2018-03-16
Add debug output to stubgen test flakes

JukkaL:
This might help debug stubgen test flake that occasionally
happens in CI (#4635).

Alternative to #4672. A notable difference is that this prints 
a traceback from importlib.

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -38,6 +38,7 @@ TODO:
 import glob
 import importlib
 import json
+import os
 import os.path
 import pkgutil
 import subprocess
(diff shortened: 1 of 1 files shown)


JukkaL:
@msullivan I forgot about #4672 so this is mostly duplicate of that, modulo a few differences. This is arguably slightly cleaner and can dump importlib tracebacks so maybe this should go in?

msullivan:
I'm happy to merge this instead of #4672.

I'm not sure how much it will help, though: the reason I never bothered merging #4672 was that it hit the bug on travis before I merged it and the spew didn't help much

PR #4752, opened 2018-03-16
Support metaclasses in fine grained mode

ilevkivskyi:
The PR is straightforward. It also fixes a metaclass dependency bug in normal incremental mode.

I add tests for both Python 2 and Python 3, although most of them are very similar, I feel safer to check both versions thoroughly.


JukkaL:

ilevkivskyi:
@JukkaL Fixed.

test-data/unit/deps-types.test
```diff
@@ -152,6 +152,140 @@ def h() -> None:
...
+    def __add__(self, other):
+        # type: (M) -> M
+        pass
+class C:
+    __metaclass__ = M
+[out]
+<mod.C.__init__> -> m.f
+<mod.C> -> m, m.f
+<mod.M.__add__> -> m.f
+<mod.M.__radd__> -> m.f
```
JukkaL: Also test dependency from metaclass to the class that has the metaclass. Could be something like `<M> -> <C>`.

test-data/unit/fine-grained.test
```diff
@@ -1743,6 +1743,292 @@ def f(x=1): pass
...
+class M(type):
+    x = None  # type: int
+[out]
+==
+a.py:5: error: Incompatible types in assignment (expression has type "int", variable has type "str")
+==
+a.py:5: error: "Type[C]" has no attribute "x"
+==
+
+[case testMetaclassAttributesDirect_python2]
```
JukkaL: Probably having just one or two Python 2 test cases would be sufficient, since only the syntax is different.

test-data/unit/deps-types.test
```diff
@@ -152,6 +152,140 @@ def h() -> None:
...
+class M(type):
+    x = None  # type: int
+class C:
+    __metaclass__ = M
+[out]
+<mod.C.x> -> m.f
+<mod.C> -> <m.f>, m, m.f
+<mod.M.x> -> m.f
+
+[case testMetaclassAttributesDirect_python2]
```
JukkaL: Fewer Python 2 test cases would be sufficient.

test-data/unit/fine-grained.test
```diff
@@ -1743,6 +1743,292 @@ def f(x=1): pass
...
+[file submod.py.2]
+class M(type):
+    def __add__(self, other):
+        # type: (M) -> M
+        pass
+[out]
+a.py:4: error: Unsupported operand types for + ("Type[C]" and "Type[C]")
+==
+
+[case testFineMetaclassUpdate]
```
JukkaL: This test case looks very similar to `testMetaclassAttributesDirect`. Maybe it's redundant?

test-data/unit/fine-grained.test
```diff
@@ -1743,6 +1743,292 @@ def f(x=1): pass
...
+[file b.py.2]
+import c
+class B(metaclass=c.M): pass
+
+[file c.py]
+class M(type):
+    pass
+[out]
+==
+a.py:3: error: Inconsistent metaclass structure for 'D'
```
JukkaL: Ideas for additional tests cases:

* Removing metaclass from a class.
* Change metaclass to another one.
* Initially metaclass definition doesn't point to a class (undefined or not a class). In update point metaclass to a suitable class.
* Update makes metaclass point to a non-class.

test-data/unit/fine-grained.test
ilevkivskyi: Hm, indeed. Maybe I just forgot that I already tested renamed attribute.

test-data/unit/deps-types.test
ilevkivskyi: I think I actually missing this dependency. Namely, there are two metaclasses, declared and actual (calculated). I think we don't need deps on the latter (also it appears as a part of class snapshot), but I think we need a dep on declared metaclass.

ilevkivskyi:
@JukkaL Thanks! I added the tests you proposed, and also added one more dependency.

(Test failure looks like a flake, I restarted the job.)

test-data/unit/fine-grained.test
```diff
@@ -1984,6 +1984,65 @@ class B:
...
+    pass
+[file b.py]
+from c import M
+[file c.py]
+M = 1
+[file c.py.2]
+class M(type):
+    pass
+[out]
+a.py:2: error: Invalid metaclass 'b.M'
```
JukkaL: Also test that reference to metaclass attribute gets fixed by adding the metaclass.

test-data/unit/fine-grained.test
```diff
@@ -1984,6 +1984,65 @@ class B:
...
+class C(metaclass=b.M):
+    pass
+[file b.py]
+class M(type):
+    pass
+[file b.py.2]
+M = 1
+[out]
+==
+a.py:2: error: Invalid metaclass 'b.M'
```
JukkaL: Also test metaclass attribute reference through `C` that should be refreshed using fine-grained deps. 

The more common case would be deletion of `M` (or undefined reference to metaclass that gets fixed) and these would also be worth testing.

PR #4754, opened 2018-03-17
Only allow snake_case keys in configuration file (#4753)

### LINKED ISSUE #4753: Configuration file recognizes dash as correct but results in broken config
mypy doesn't seem to be able to recognize dashed-options as incorrect within the context of a config file.

For example, suppose we have the following files `nothing.py` and `mypy.ini`

```python
def nothing():
    pass
```

```
[mypy]
X = True
```

When `X` is:
- `disallow_untyped_defs`
    - Correctly reports `nothing.py:1: error: Function is missing a type annotation`
- `disallow_untyped_defs1`
    - Correctly reports: `mypy.ini: [mypy]: Unrecognized option: disallow_untyped_defs1 = True`
- `disallow-untyped-defs`
    - Doesn't tell me I'm missing a type annotation, nor that the option is incorrect.  No errors at all!

Python 3.6.4 and mypy 0.570 were used.

ceh:
We must use the original key when reading the value from the config file, since it may have been specified in a dash-notation.

Fixes #4753.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -760,8 +760,6 @@ def parse_section(prefix: str, template: Options,
     results = {}  # type: Dict[str, object]
     report_dirs = {}  # type: Dict[str, str]
     for key in section:
-        orig_key = key
-        key = key.replace('-', '_')
         if key in config_types:
             ct = config_types[key]
         else:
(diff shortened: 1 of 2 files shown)


gvanrossum:
It looks like the code never worked. Seeing how many places would have to be modified, it feels brittle to try and support this. I suggest that it may be better to just get rid of orig_key and delete the normalization at line 764. We may have to update the docs too.

ceh:
I couldn't find any documentation where it's stated that the dash-notation is supported. See e.g. https://github.com/python/mypy/blob/10522cf/docs/source/config_file.rst

I added a test case as well, so that it's documented in some sense that it's not a valid syntax.

gvanrossum:
And thanks for all that! Agreed, it wasn't even documented (I think it was just a wishful feature in my mind when I wrote the code).

gvanrossum:
And thanks for all that! Agreed, it wasn't even documented (I think it was just a wishful feature in my mind when I wrote the code).

PR #4757, opened 2018-03-19
sync typeshed

JelleZijlstra:
@ilevkivskyi is it too late to include this in the release? I would like to make sure people's typeshed fixes make it into the mypy release.


ilevkivskyi:
I think this can go in. I will wait until all tests pass (I restarted one which looks like a flake).

JelleZijlstra:
Thanks!

Issue #3912, opened 2017-09-03
Bazel build integration


sixolet:
I've been playing with getting Mypy to integrate with bazel for typechecking. I've got a basic version of it working. I intend to describe it here for discussion of the general approach, and then post a series of PRs for individual things it needs as I clean them up, if the general approach seems right.

The general approach: Make a `mypy_library` and `mypy_binary` macro that puts a typechecking aspect on the python files it contains. The aspect calculates the mypy cache files; give mypy a mode where it stops considering itself responsible for calculating freshness of dependencies, and instead let bazel handle that.

Specific PRs I expect to make:
* A `--cache-hermetically` option to stop writing mtimes
* A  `--cache-mapping` option to specify a json file containing `.data.json` and `.meta.json` file locations by module id, both for cache files to read and new cache files to write (bazel likes being super specific about things like this; it seems cleaner to specify to mypy than to rely on mypy's cache directory layout). This option would also short-circuit the process of looking for source and checking it for freshness for the covered modules; just believe the cache files bazel is handing you.
* Possibly a `--module-without-init` option (or some name, I don't know what it should be called) to allow mypy to believe things are modules without an `__init__.py` file. Bazel for some reason doesn't care whether you have them in your python source when you build.  I can possibly hack around this by touching that file myself in the build directory in the bazel rule, instead.
* A `--ignore-module-prefix` option to understand that certain parts of the module path are already accounted for, and all modules within the prefix to ignore should be treated as if they have empty `__init__.py`. 
* An actual Skylark implementation of the python typechecking macros, and a `BUILD` file for mypy.

That would be enough to get it working (and I have a prototype of that, which I need to split into separate PRs and clean up a lot if it seems an ok approach).  To make it awesomer:
* Speed up mypy's startup time. There might be low-hanging fruit?
* Precompute the cache files for builtins and typeshed, so we never need to go looking for source there.
* Implement some dependency pruning, so you don't even need to load all the cache files all the time.

I'm submitting this description before I'm ready to submit PRs deliberately, so I am not super attached to any one approach. Suggestions?

ethanhs:
>--module-without-init

I don't think this is a good idea. The issue sounds more to be with Bazel than Mypy, and I think it makes more sense to modify the build script until Bazel fixes this.

>Precompute the cache files for builtins and typeshed, so we never need to go looking for source there.

[...]

PR #4759, opened 2018-03-19
Add several undocumented flags (--bazel, --package-root and --cache-map) in support of Bazel

gvanrossum:
The `--bazel` flag make cache files hermetic by using relative paths and setting mtime to zero; effectively the presence of a cache file prevents reading of the source file (though the source file must still exist).

The `--cache-map` flag specifies a mapping from source files to cache files that overrides the usual way of finding the locations for cache files; e.g.
```
mypy --cache-map foo.py foo.meta.json foo.data.json bar.py bar.meta.json bar.data.json -- foo.py bar.py
```
There must be any number of *triples* (source, meta, data) with the additional constraints that the source must end in `.py` or `.pyi`, meta must end in `.meta.json`, and data must end in `.data.json`. For files in the cache map, the default cache directory (`--cache-dir`) is ignored, and if the cache files named in the `--cache-map` don't exist they will be written. For files not in the cache map, the `--cache-dir` is still used.

(NOTE: You still need to pass the source files separately, and the list of `--cache-map` triples must be followed by another flag or by `--`, as in the example above.)


mypy/build.py
```diff
@@ -66,13 +66,38 @@
...
+
 Graph = Dict[str, 'State']
 
 
 def getmtime(name: str) -> int:
     return int(os.path.getmtime(name))
 
 
+def relpath(path: str) -> str:
+    if not path or path[0] not in SEPARATORS:
```
sixolet: """Returns the path relative to the current working directory."""

mypy/build.py
```diff
@@ -931,14 +961,17 @@ def get_cache_names(id: str, path: str, manager: BuildManager) -> Tuple[str, str
...
-      path: module path (used to recognize packages)
+      path: module path
       cache_dir: cache directory
       pyversion: Python version (major, minor)
 
     Returns:
       A tuple with the file names to be used for the meta JSON and the
       data JSON, respectively.
     """
+    pair = manager.options.cache_map.get(path)
```
sixolet: It feels naïvely like it'd be cleaner to always use the cache map in the presence of a cache map at all; I don't know that the hybrid mode of looking in a cache dir *and* a cache map file would be very meaningful, and it certainly sounds hard to debug.

mypy/main.py
```diff
@@ -381,6 +382,15 @@ def add_invertible_flag(flag: str,
     # --local-partial-types disallows partial types spanning module top level and a function
     # (implicitly defined in fine-grained incremental mode)
     parser.add_argument('--local-partial-types', action='store_true', help=argparse.SUPPRESS)
+    # --bazel changes some behaviors for use with Bazel (https://bazel.build).
+    parser.add_argument('--bazel', action='store_true', help=argparse.SUPPRESS)
+    # --cache-map FILE ... gives a mapping from source files to cache files.
```
sixolet: I am interested in what happened to the idea of dropping cache files next to source files. I remember choosing the same option you did here, but I forget why, because now-me had the idea that cache files next to source would be cleaner for some reason.

mypy/main.py
```diff
@@ -520,6 +530,18 @@ def add_invertible_flag(flag: str,
             report_dir = val
             options.report_dirs[report_type] = report_dir
 
+    # Parse cache map.  Uses assertions for checking.
+    if special_opts.cache_map:
+        n = len(special_opts.cache_map)
+        assert n % 3 == 0, "--cache-map requires one or more triples (see source)"
+        for i in range(0, n, 3):
+            source, meta_file, data_file = special_opts.cache_map[i:i + 3]
+            assert source not in options.cache_map
```
sixolet: In final code probably wants to be something other than an assert

mypy/build.py
gvanrossum: That might well be the case but I haven't solved what to do about typeshed yet, so without keeping cache_dir as a fallback, typeshed will always read the source.

mypy/main.py
gvanrossum: I tried to write this up in my overly long status report in #3912 -- Dropbox's build environment puts the output files in a different directory than the source files. It seems outputs (and inputs) must be `File` objects, the only way to create those is `ctx.actions.declare_file()`, which always puts it it in the output directory. (The `File` objects for the sources come from `ctx.files.srcs`, but I don't think I can add new ones via that approach.)

kamahen: That is correct -- you cannot add anything via ctx.files.srcs. You should assume that the source tree is read-only.

If you declare a file, you must create it within the action (Bazel checks this).

gvanrossum:
I have to revise this, see https://github.com/python/mypy/issues/3912#issuecomment-376286346 why. (I can't solve the implicit `__init__.py` problem in pure Skylark, I need to support this in mypy.)

gvanrossum:
I probably have to go back here and add a new flag, `--package-roots DIR DIR ...`. I've finally hit a large enough sample of our codebase to hit a situation where a "source" file is actually generated by another rule/action, and in that case the file exists in a different tree. Fortunately in Skylark I can figure out when this is the case: `File` objects have a `root` attribute which is another `File` object pointing there, and it's empty for genuine sources but points to the output tree for generated sources. So I can write some Skylark that computes the necessary `--package-roots` arguments. I was somewhat anticipating this with the naming of the arguments in the last diff, but I'll have to switch it from a `bool` to `List[str]`.

sixolet:
I'm following along. I think my version of this a few months ago didn't fake having `__init__.py`, it allowed dirs to be treated as packages even in the absence of `__init__.py`. I think I like your version better.

I also had to do something similar to your suggested `--package-roots` change, though at the time I thought it was just because of the weird way my particular codebase is organized.

mypy/build.py
```diff
@@ -1149,11 +1164,11 @@ def validate_meta(meta: Optional[CacheMeta], id: str, path: Optional[str],
     fine_grained_cache = manager.use_fine_grained_cache()
 
     size = st.st_size
-    if size != meta.size and not fine_grained_cache:
+    if size != meta.size and not bazel and not fine_grained_cache:
```
msullivan: Why drop this check in bazel mode?

mypy/build.py
```diff
@@ -1250,8 +1265,17 @@ def write_cache(id: str, path: str, tree: MypyFile,
...
+
     # Obtain file paths
-    path = os.path.abspath(path)
+    if not bazel:
+        # Normally, make all paths absolute.
+        path = os.path.abspath(path)
+    else:
+        # For Bazel, make all paths relative (else Bazel caching is thwarted).
+        # And also patch tree.path, which might have been made absolute earlier.
+        tree.path = path = os.path.relpath(path)
```
kamahen: Bazel uses a lot of symlinks, so you might want os.path.realpath here.
(Do you also want it for the "not bazel" case?)

mypy/build.py
```diff
@@ -1282,10 +1307,11 @@ def write_cache(id: str, path: str, tree: MypyFile,
         return interface_hash, None
 
     # Write data cache file, if applicable
+    # Note that for Bazel we don't record the data file's mtime.
     if old_interface_hash == interface_hash:
         # If the interface is unchanged, the cached data is guaranteed
         # to be equivalent, and we only need to update the metadata.
-        data_mtime = getmtime(data_json)
+        data_mtime = 0 if bazel else getmtime(data_json)
```
kamahen: Maybe put "0 if bazel else getmtime(path)" into a one-line function? (it appears in a lot of places and causes a bit of cognitive load each time).

mypy/build.py
```diff
@@ -1352,7 +1378,7 @@ def delete_cache(id: str, path: str, manager: BuildManager) -> None:
     This avoids inconsistent states with cache files from different mypy runs,
     see #4043 for an example.
     """
-    path = os.path.abspath(path)
+    path = os.path.relpath(path) if manager.options.bazel else os.path.abspath(path)
```
kamahen: Here also you might want to call realpath to get rid of symlinks.

mypy/build.py
gvanrossum: Because in Bazel mode we don't do any of the cache validation things. If the cache file exists it should be used, period.

kamahen: Perhaps add a comment like "# Bazel ensures the cache is vlaid"?

gvanrossum:
This is now a bit more battle-tested (see latest commit) and I'm ready for a full-scale review to land it.

gvanrossum:
Note that "gvanrossum added some commits 17 days ago" is a lie -- I merged today, fixing two (fortunately simple) conflicts. @msullivan it seems you're "it" for the review of this diff...

mypy/main.py
```diff
@@ -633,6 +650,57 @@ def add_invertible_flag(flag: str,
             report_dir = val
             options.report_dirs[report_type] = report_dir
 
+    # Validate and normalize --package-root.
```
msullivan: Pull this logic out to functions?

test-data/unit/check-incremental.test
```diff
@@ -4378,3 +4378,14 @@ from d import k
 [file t.py.2]
 from d import k
 # dummy change
+
+[case testBazelFlagIgnoresFileChanges]
```
msullivan: Are there any other `--bazel` behaviors we can check in tests without too much pain?

mypy/build.py
```diff
@@ -1405,10 +1430,11 @@ def write_cache(id: str, path: str, tree: MypyFile,
         return interface_hash, None
 
     # Write data cache file, if applicable
+    # Note that for Bazel we don't record the data file's mtime.
     if old_interface_hash == interface_hash:
         # If the interface is unchanged, the cached data is guaranteed
         # to be equivalent, and we only need to update the metadata.
-        data_mtime = getmtime(data_json)
```
msullivan: ^

mypy/build.py
```diff
@@ -71,6 +71,12 @@
 PYTHON_EXTENSIONS = ['.pyi', '.py']
 
 
+if os.altsep:
```
msullivan: Dead code?

mypy/fscache.py
```diff
@@ -133,12 +197,18 @@ def read(self, path: str) -> bytes:
...
 
-        try:
-            with open(path, 'rb') as f:
-                data = f.read()
-        except Exception as err:
-            self.read_error_cache[path] = err
-            raise
+        dirname, basename = os.path.split(path)
+        dirname = os.path.normpath(dirname)
+        if basename == '__init__.py' and dirname in self.fake_package_cache:
```
msullivan: A comment explaining what is going on and why?

gvanrossum:
I think I've done everything you asked for, and I also merged master.
Please re-review!

gvanrossum:
I'm unclear about the test failures. I can't repro them locally, except once, and after that they passed. So they seem to be flakes, or the result of imperfect test isolation.

gvanrossum:
The test flakes weren't due to a bad merge. In a recent commit I replaced an uncached `os.path.getmtime()` call with a call to `fscache.stat()` (and extracting the mtime). But that particular `getmtime()` call apparently needs to be uncached. This also (intuitively) explains why the failures were flakes -- they depended on filesystem state.

thejcannon:
For posterity "For files not in the cache map, the --cache-dir is still used." isn't true as of https://github.com/python/mypy/commit/b0319aff582d918ec6e5a3bf69860a958eb31261.

Just leaving it here because this cost me some headscratching :smile:

PR #4761, opened 2018-03-20
Fix adding a class in a new module and using it

### LINKED ISSUE #4512: Reexported imports not tracked properly in fine-grained incremental mode
We fail to propagate changes when the target of an import changes.

Here, we swap what module `a` imports `foo` from, but `m.f` doesn't get rechecked.
```
[case testImportSwap]
from a import foo
def f() -> int:
    return foo()

[file a.py]
from b1 import foo
[file a.py.2]
from b2 import foo

[file b1.py]
def foo() -> int:
    return 12
[file b2.py]
def foo() -> str:
    return 'uhoh'
[out]
==
main:3: error: Incompatible return value type (got "str", expected "int")
```

Same thing if we switch from `a` importing `foo` to just defining it.
```
[case testImportSwap2]
from b import foo
def f() -> int:
    return foo()

[file a.py]
from b import foo
[file a.py.2]
def foo() -> str:
    return 'uhoh'

[file b.py]
def foo() -> int:
    return 12
[out]
==
main:3: error: Incompatible return value type (got "str", expected "int")
```

Same problem without using `from` in `m`.
```
[case testImportSwap3]
import a
def f() -> int:
    return a.foo()
[file a.py]
from b1 import foo
[file a.py.2]
from b2 import foo

[file b1.py]
def foo() -> int:
    return 12
[file b2.py]
def foo() -> str:
    return 'uhoh'
[out]
==
main:3: error: Incompatible return value type (got "str", expected "int")
```

msullivan:
We add a field to AnyType indicating---if it was created because of a
failed import---what the full name of the imported name is (that is,
the name in the importing module).

An eventual better solution might be to *always* generate dependencies to the locally imported name of variables. That would fix #4512 as well.

### DIFF
--- a/test-data/unit/deps.test
+++ b/test-data/unit/deps.test
@@ -604,10 +604,18 @@ x = 0
 [out]
 <a[wildcard]> -> m
 
-[case testMissingModuleClass]
+[case testMissingModuleClass1]
 from b import A  # type: ignore
 def f(x: A) -> None:
     x.foo()
 [out]
 <m.A> -> <m.f>, m.f
 <b.A> -> m
+
+[case testMissingModuleClass2]
+from p.b import A  # type: ignore
+def f(x: A) -> None:
+    x.foo()
+[out]
+<m.A> -> <m.f>, m.f
+<p.b.A> -> m
--- a/test-data/unit/deps.test
+++ b/test-data/unit/deps.test
@@ -603,3 +603,11 @@ from a import *
 x = 0
 [out]
 <a[wildcard]> -> m
+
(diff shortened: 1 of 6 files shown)


test-data/unit/deps.test
```diff
@@ -603,3 +603,11 @@ from a import *
 x = 0
 [out]
 <a[wildcard]> -> m
+
+[case testMissingModuleClass]
```
JukkaL: Try with an import from a package as well (such as `from p.b import A`) and using `from p import *`.

PR #4763, opened 2018-03-20
Support __new__ in fine grained mode

ilevkivskyi:
This is a straightforward PR that essentially adds same dependencies on ``__new__`` as we have on ``__init__`` for a class as a callable.

Just in case I add few dedicated tests.

### DIFF
--- a/test-data/unit/fine-grained.test
+++ b/test-data/unit/fine-grained.test
@@ -3555,3 +3555,24 @@ class C:
 [out]
 ==
 a.py:3: error: Argument 1 to "C" has incompatible type "int"; expected "str"
+
+[case testDunderNewUpdatedCallable]
+import a
+[file a.py]
+from typing import Callable, Any
+import b
+
+def func(arg: Callable[[int], Any]) -> None:
+    pass
+func(b.C)
+[file b.py]
+class C:
+    def __new__(cls, x: int) -> C:
+        pass
+[file b.py.2]
+class C:
+    def __new__(cls, x: str) -> C:
+        pass
+[out]
+==
+a.py:6: error: Argument 1 to "func" has incompatible type "Type[C]"; expected "Callable[[int], Any]"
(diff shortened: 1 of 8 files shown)


test-data/unit/fine-grained.test
```diff
@@ -3458,3 +3458,121 @@ a.py:1: error: "int" not callable
...
+        pass
+[file b.py.2]
+class C:
+    def __new__(cls, x: str) -> C:
+        pass
+[out]
+==
+a.py:3: error: Argument 1 to "C" has incompatible type "int"; expected "str"
+
+[case testDunderNewUpdatedFunc]
```
JukkaL: Not sure if we need both this and the above test case (`testMetaclassAttributesDirect`) since they are pretty similar.

test-data/unit/fine-grained.test
```diff
@@ -3458,3 +3458,121 @@ a.py:1: error: "int" not callable
...
+        pass
+[file b.py.2]
+class C:
+    def __new__(cls, x: str) -> C:
+        pass
+[out]
+==
+a.py:3: error: Argument 1 to "C" has incompatible type "int"; expected "str"
+
+[case testDunderNewUpdatedMethod]
```
JukkaL: Again this test case seems redundant.

test-data/unit/fine-grained.test
```diff
@@ -3458,3 +3458,121 @@ a.py:1: error: "int" not callable
...
+[file b.py]
+class C:
+    def __new__(cls, x: int) -> C:
+        pass
+[file b.py.2]
+class C:
+    def __new__(cls, x: str) -> C:
+        pass
+[out]
+==
```
JukkaL: Ideas for additional test cases:
* Add `__new__` when originally there was nothing.
* Replace `__init__` with `__new__`.

ilevkivskyi:
@JukkaL Thanks for review! I left only one of three boring tests, and added the two other tests you proposed.

PR #4765, opened 2018-03-20
Add more tests for decorators in fine grained mode

ilevkivskyi:
It looks like decorators are already supported but some tests were missing. I added the dependency and update tests for decorator signature, function signature, and decorator arguments. I add these tests for module scope, nested functions scope, and for a decorated method.


JukkaL:
This PR has a merge conflict.

ilevkivskyi:
> This PR has a merge conflict.

Fixed now.

test-data/unit/deps.test
```diff
@@ -604,6 +604,84 @@ x = 0
...
+    pass
+class C:
+    pass
+[out]
+<m.D.func> -> m
+<m.D> -> m.D
+<mod.C.__init__> -> m
+<mod.C> -> m
+<mod.deca> -> m
+<mod.decb> -> m
```
JukkaL: Ideas for test cases: Method decorator in class nested within other class, decorator in a function nested within a function, method decorator in a class nested within a function.

ilevkivskyi:
OK, thanks! I added test cases you suggested plus one more, where a class in function signature is invalidated due to changed MRO.

PR #4768, opened 2018-03-20
Somehow mypy/api.py had DOS line endings. Fix this.

gvanrossum:


gvanrossum:
I suspect it was the original author of this file... (@JdeH)

JdeH:
Indeed, I develop on Windows.
Hadn't thought of that...

PR #4769, opened 2018-03-20
On a dmypy crash, report the crash to the client

msullivan:
Tested manually


mypy/dmypy_server.py
```diff
@@ -162,7 +162,14 @@ def serve(self) -> None:
                             resp = {'error': "Command is not a string"}
                         else:
                             command = data.pop('command')
-                        resp = self.run_command(command, data)
+                        try:
+                            resp = self.run_command(command, data)
+                        except:
```
gvanrossum: Let's just catch `except Exception:`.

PR #4778, opened 2018-03-21
Add more tests for generic callables in fine grained mode

ilevkivskyi:
This PR adds some more tests for generic callables in fine grained in addition to several tests added as part of type variables support PR.

These tests uncovered two bugs that I also fix here. For some reason, one of the tests fails without cache, I spent some time but still don't see why. @JukkaL I will be grateful for your advise.

### DIFF
--- a/test-data/unit/fine-grained.test
+++ b/test-data/unit/fine-grained.test
@@ -3886,6 +3886,173 @@ main:3: error: Value of type variable "submod.T" of "D" cannot be "S"
 main:3: error: Value of type variable "submod.U" of "D" cannot be "I"
 ==
 
+[case testGenericFineCallableNormal]
+import a
+[file a.py]
+import b
+x: int = b.f(int())
+[file b.py]
+from c import g
+f = g
+[file c.py]
+from typing import TypeVar
+class B: pass
+T = TypeVar('T')
+def g(x: T) -> T:
+    pass
+[file c.py.2]
+from typing import TypeVar
+class B: pass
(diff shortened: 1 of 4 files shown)


test-data/unit/fine-grained.test
```diff
@@ -3886,6 +3886,173 @@ main:3: error: Value of type variable "submod.T" of "D" cannot be "S"
...
+[file c.py.2]
+from typing import TypeVar
+class B: pass
+T = TypeVar('T')
+def g(y: T) -> T:
+    pass
+[out]
+==
+a.py:2: error: Unexpected keyword argument "x"
+c.py:4: note: called function defined here
```
JukkaL: Maybe capitalize the 'c' in 'called'?

test-data/unit/fine-grained.test
```diff
@@ -3886,6 +3886,173 @@ main:3: error: Value of type variable "submod.T" of "D" cannot be "S"
...
+from typing import TypeVar
+class B: pass
+T = TypeVar('T', bound=B)
+def g(x: T) -> T:
+    pass
+[out]
+==
+a.py:2: error: Value of type variable "T" of function cannot be "int"
+
+[case testGenericFineCallableBoundDeleted-skip-nocache]
```
JukkaL: You can create a separate issue about this failure and we look at it later.

ilevkivskyi: OK, will create a separate issue.

Issue #4477, opened 2018-01-17
Crash in fine-grained incremental mode if __init__.py is deleted


JukkaL:
Deleting package `__init__.py` causes a fine-grained incremental mode crash.

Test case:

```
[case testDeletePackageInitFile]
# cmd: mypy p/__init__.py p/a.py
[file p/__init__.py]
[file p/a.py]
[delete p/__init__.py.2]
```

Traceback:

```
/Users/jukka/src/mypy/mypy/server/update.py:378: in update_single_isolated
    graph = load_graph(sources, manager)
/Users/jukka/src/mypy/mypy/build.py:2187: in load_graph
    ancestor_for=st)
/Users/jukka/src/mypy/mypy/build.py:1610: in __init__
    raise CompileError(["mypy: can't find module '%s'" % id])
E   mypy.errors.CompileError: mypy: can't find module 'p'

During handling of the above exception, another exception occurred:
/Users/jukka/src/mypy/mypy/test/testfinegrained.py:73: in run_case
    new_messages = fine_grained_manager.update(modules)
/Users/jukka/src/mypy/mypy/server/update.py:222: in update
    result = self.update_single(next_id, next_path)
/Users/jukka/src/mypy/mypy/server/update.py:265: in update_single
    result = update_single_isolated(module, path, manager, previous_modules)
/Users/jukka/src/mypy/mypy/server/update.py:381: in update_single_isolated
    assert err.module_with_blocker
E   AssertionError
```

JukkaL:
This can also happen if a newly created package is missing `__init__.py`:

```py
[case testAddPackage4]
import p.a
p.a.f(1)
[file p/a.py.2]
def f(x: str) -> None: pass
[file p/__init__.py.3]
```

msullivan:
I think the fix here is that whenever an `__init__.py` is added, we need to remove and re-add any modules underneath it, since their names change.
To really get it right, we also need to do this for __init__ files that aren't in the build, unfortunately

PR #4785, opened 2018-03-23
Make create_source_list not exit, use it for testfinegrained

msullivan:
This fixes a testfinegrained bug in which the computation of module
names did not properly take into account the absence of `__init__.py`
files when file names were specified in the test.

This removes a confounding factor to fixing #4477.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -24,8 +24,8 @@ from mypy.version import __version__
 PY_EXTENSIONS = tuple(PYTHON_EXTENSIONS)
 
 
-class InvalidPackageName(Exception):
-    """Exception indicating that a package name was invalid."""
+class InvalidSourceList(Exception):
+    """Exception indicating a problem in the list of sources given to mypy."""
 
 
 orig_stat = os.stat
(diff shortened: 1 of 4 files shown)


mypy/main.py
```diff
@@ -555,18 +558,12 @@ def create_source_list(files: Sequence[str], options: Options) -> List[BuildSour
...
         elif os.path.isdir(f):
-            try:
-                sub_targets = expand_dir(f)
-            except InvalidPackageName as e:
-                fail(str(e))
+            sub_targets = expand_dir(f)
             if not sub_targets:
-                fail("There are no .py[i] files in directory '{}'"
-                     .format(f))
+                raise Exception("There are no .py[i] files in directory '{}'"
```
gvanrossum: Please define a custom exception for this purpose. `Exception` is for catching, not throwing. (I see that this also throws `InvalidPackageName`, from `crawl_up()`, maybe you can rename and reuse it?

mypy/main.py
```diff
@@ -555,18 +558,12 @@ def create_source_list(files: Sequence[str], options: Options) -> List[BuildSour
     targets = []
     for f in files:
         if f.endswith(PY_EXTENSIONS):
-            try:
-                targets.append(BuildSource(f, crawl_up(f)[1], None))
-            except InvalidPackageName as e:
-                fail(str(e))
+            targets.append(BuildSource(f, crawl_up(f)[1], None))
```
gvanrossum: I don't think it's necessarily an improvement to let `InvalidPackageName` (raised by `crawl_up()`) is now bubbled out from this. Maybe add a comment here, and another to the function header (in a docstring?) about which exception is raised if things go pear-shaped?

PR #4786, opened 2018-03-23
Rollback typeshed commit

ilevkivskyi:
It looks like the `ctypes` stubs cause several problems in some internal code bases. This is mostly due to mypy's own limitation in treatment of platform checks. But it looks like there may be some missing or problematic definitions:
* `CArgObject` missing
* `wintypes` missing
* It looks like `memmove` can also accept `str` and `_cparam`
* There are lots of errors like `Unsupported operand types for * ("Type[Array[<nothing>]]" and "int")`, so I suppose `Array` has a metaclass that implements `__mul__/__rmul__` like `_CDataMeta`.
* Signatures of `_CDataMeta.__mul__` and `_CDataMeta.__rmul__` are wrong, they contain unbound type variable resulting in `<nothing>`.
* Also it looks like mypy has problems with type inference for `*` types, so `Array.__init__` is currently problematic.

The stub looks good overall, so I don't want to revert it, but it needs some polishing before it gets into the next release.

@JelleZijlstra


gvanrossum:
Why roll it back? Do you want to cherry-pick the rollback into the 0.580
release?

ilevkivskyi:
> Do you want to cherry-pick the rollback into the 0.580
release?

Yes.

gvanrossum:
OK LGTM then.

JelleZijlstra:
Oops. I filed python/typeshed#1983 to get the typeshed issue fixed before the next mypy release.

PR #4790, opened 2018-03-26
Support protocols in fine grained mode

ilevkivskyi:
Support of protocols in fine grained mode is non-trivial because protocols are not very "modular" (can have action at a distance). To support them we introduce several new dependencies and a global state that represents protocol dependency map (serialized in a separate cache file).

There are three kinds of protocol dependencies. For example, after a subtype check:
```python
x: Proto = C()
```
the following dependencies will be generated:
```
1. ..., <SuperProto[wildcard]>, <Proto[wildcard]> -> <Proto>
2. ..., <B.attr>, <C.attr> -> <C> [for every attr in Proto members]
3. <C> -> Proto  # this one to invalidate the subtype cache
```
The first kind is generated immediately per-module. While two other kinds are generated
after all modules are type checked and we have recorded all the subtype checks.

Another change is that we reset subtype caches in protocols scheduled for recheck _before_ processing any other targets. This avoids some false negatives in corner cases that are sensitive to module processing order.

I didn't test the scenario when the protocol cache files are tampered with, it is also hard to write any tests for this.

_UPDATE:_ edited description to reflect the current state of PR.


mypy/build.py
```diff
@@ -926,6 +927,84 @@ def verify_module(fscache: FileSystemMetaCache, id: str, path: str) -> bool:
...
+    Serialize protocol dependencies map for fine grained mode. Also take the snapshot
+    of current sources to later check consistency between protocol cache and individual
+    cache files.
+    """
+    proto_meta, proto_cache = get_protocol_deps_cache_name(manager)
+    meta_snapshot = {}  # type: Dict[str, str]
+    for id in graph:
+        meta_snapshot[id] = graph[id].source_hash
+    if not atomic_write(proto_meta, json.dumps(meta_snapshot), '\n'):
+        manager.log("Error writing protocol meta JSON file {}".format(proto_cache))
```
JukkaL: Error handling is a bit sketchy -- the cache is unusable without a protocol meta JSON file. Other cache files can be regenerated by reprocessing individual modules, but the global cache can't be built incrementally. So if we write some cache files but not the global cache file we should probably fail the build. If we don't write any cache files at all, things are probably okay, since we'll need to a full rebuild which will also populate the protocol data structures correctly.

mypy/build.py
```diff
@@ -926,6 +927,84 @@ def verify_module(fscache: FileSystemMetaCache, id: str, path: str) -> bool:
...
+    listed_proto_deps = {k: list(v) for (k, v) in proto_deps.items()}
+    if not atomic_write(proto_cache, json.dumps(listed_proto_deps), '\n'):
+        manager.log("Error writing protocol deps JSON file {}".format(proto_cache))
```
JukkaL: Similar to above -- error checking needs to be more involved.

mypy/build.py
```diff
@@ -2052,8 +2131,13 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
     # just want to load in all of the cache information.
     if manager.use_fine_grained_cache():
         process_fine_grained_cache_graph(graph, manager)
+        manager.proto_deps = read_protocol_cache(manager, graph)
     else:
         process_graph(graph, manager)
+        if manager.options.cache_fine_grained or manager.options.fine_grained_incremental:
+            proto_deps = collect_protocol_deps(graph)
```
JukkaL: This needs a comment, since the special protocol dependencies are non-trivial.

mypy/build.py
```diff
@@ -2052,8 +2131,13 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
     # just want to load in all of the cache information.
     if manager.use_fine_grained_cache():
         process_fine_grained_cache_graph(graph, manager)
+        manager.proto_deps = read_protocol_cache(manager, graph)
```
JukkaL: `read_protocol_cache` can return `None` if it can't load the file. Can we handle it here? Without protocol dependencies we won't be able to produce correct results, so it looks like handling it somehow is the right thing to do. Maybe we should do a full build instead of using the cache?

mypy/server/deps.py
```diff
@@ -734,6 +749,43 @@ def visit_union_type(self, typ: UnionType) -> List[str]:
...
+                if 'typing' in trigger or 'builtins' in trigger:
+                    continue
+                # If any class that was checked against a protocol changes,
+                # we need to reset the subtype cache for the protocol.
+                # This however creates a "weak" dependency, the protocol doesn't
+                # need to be re-checked, and its uses elsewhere are still valid
+                # (unless invalidated by other deps). We mark this kind of deps
+                # by a star, since they are higher priority, i.e., they need
+                # to be processed before any other deps.
+                deps[trigger].add(info.fullname() + '*')
```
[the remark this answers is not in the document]
ilevkivskyi: OK, I will try to put this in a separate structure and generalize recursive triggering logic so that it can handle both normal and protocol deps.

mypy/server/deps.py
```diff
@@ -734,6 +749,43 @@ def visit_union_type(self, typ: UnionType) -> List[str]:
...
+def collect_protocol_attr_deps(names: SymbolTable,
+                               processed: Optional[Set[TypeInfo]] = None) -> Dict[str, Set[str]]:
+    """Recursively collect protocol attribute dependencies for classes in 'names'."""
+    # Currently there can be at most one target per trigger, but we might
+    # add more deps in future so we are using defaultdict.
+    if processed is None:
+        processed = set()
+    deps = defaultdict(set)  # type: DefaultDict[str, Set[str]]
+    for node in names.values():
+        if isinstance(node.node, TypeInfo) and node.node not in processed:
```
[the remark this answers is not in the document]
ilevkivskyi: Yes, this will improve speed a bit, will do.

mypy/subtypes.py
```diff
@@ -170,8 +170,11 @@ def visit_instance(self, left: Instance) -> bool:
                 if nominal:
                     right.type.record_subtype_cache_entry(left, right)
                 return nominal
-            if right.type.is_protocol and is_protocol_implementation(left, right):
-                return True
+            if right.type.is_protocol:
+                if not left.type.is_protocol:
+                    left.type.checked_against_members.update(right.type.protocol_members)
```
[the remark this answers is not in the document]
ilevkivskyi: Agree.

test-data/unit/deps-types.test
```diff
@@ -893,3 +893,65 @@ class B: pass
...
+import mod
+x: mod.P
+[file mod.py]
+from typing import Protocol
+class P(Protocol):
+    x: int
+[out]
+<m.x> -> m
+<mod.P> -> <m.x>, m, mod.P
+<mod.P[wildcard]> -> <mod.P>
```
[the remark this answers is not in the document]
ilevkivskyi: OK, although there are few changes `astdiff.py`.

mypy/build.py
```diff
@@ -612,6 +612,7 @@ def __init__(self, data_dir: str,
         self.stats = {}  # type: Dict[str, Any]  # Values are ints or floats
         self.fscache = fscache
         self.find_module_cache = FindModuleCache(self.fscache)
+        self.proto_deps = None  # type: Optional[Dict[str, Set[str]]]
```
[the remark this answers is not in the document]
ilevkivskyi: Oops, forgot to add a comment here, will fix.

mypy/build.py
```diff
@@ -926,6 +927,84 @@ def verify_module(fscache: FileSystemMetaCache, id: str, path: str) -> bool:
...
+    for id in graph:
+        meta_snapshot[id] = graph[id].source_hash
+    if not atomic_write(proto_meta, json.dumps(meta_snapshot), '\n'):
+        manager.log("Error writing protocol meta JSON file {}".format(proto_cache))
+    listed_proto_deps = {k: list(v) for (k, v) in proto_deps.items()}
+    if not atomic_write(proto_cache, json.dumps(listed_proto_deps), '\n'):
+        manager.log("Error writing protocol deps JSON file {}".format(proto_cache))
+
+
+def get_protocol_deps_cache_name(manager: BuildManager) -> Tuple[str, str]:
```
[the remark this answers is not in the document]
ilevkivskyi: OK, I will try to refactor and certainly will add the docstring.

mypy/build.py
ilevkivskyi: Yes, this is the part where I am not sure what to do. Also should we invalidate the protocol cache separately if cache was written with `--strict-optional` but is used without `--strict-optional`?

mypy/build.py
ilevkivskyi: I was thinking about showing a blocking error and stopping the build. There is a TODO about this in `update.py`.

mypy/server/deps.py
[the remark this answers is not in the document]
ilevkivskyi: Note however that this means we should never skip processing `typing`. Because e.g. subtype cache of `typing.Iterable` also needs to be invalidated if something tested against it changed.

ilevkivskyi:
@JukkaL I will push some updates, but please ignore them. I need to make some code clean-up, and one of your review comments is not addressed yet, I will let you know when this is ready for review again.

ilevkivskyi:
@JukkaL This is ready for review now. I addressed all previous comments and simplified the logic.

Re: moving `TypeInfo` state to a shared class in a new module: This is planned for a separate PR as we discussed before, but it looks like there may be some performance issues until the second PR, namely we re-collect protocol deps after every update, this can easily be several 100ms in S. After moving to the shared state the time will be negligible. Also note that moving `TypeInfo` state will have huge conflicts with your `semanal.py` refactoring, so I think we should do this in a non-overlapping way (e.g. start moving `TypeInfo` state after your PR is merged).

@msullivan Is there an easy way to test performance implications of a PR? In particular I am interested in precise numbers for this PR on warm runs.

ilevkivskyi:
(Updated the PR description to reflect the current state.)

ilevkivskyi:
I just merged master. I will update the PR to use `typestate` tomorrow (also maybe some optimizations will be possible due to deferred cache loading).

ilevkivskyi:
OK, now I updated the PR to use the newly added `TypeState`. Now protocol deps are stored there. This improves performance (we don't need to recollect all deps after every update) and simplifies the logic (most of it is now in one place).

@JukkaL This is ready for your review.

mypy/dmypy_server.py
```diff
@@ -196,6 +196,7 @@ def serve(self) -> None:
 
     def free_global_state(self) -> None:
         TypeState.reset_all_subtype_caches()
+        TypeState.reset_protocol_deps()
```
JukkaL: These who functions could be put in a function somewhere, since the two lines are repeated in `mypy.build` -- something like `reset_global_state`. It could be module-level function in mypy.typestate, and if we ever have multiple modules with global state, we can just move the function to another module.

mypy/build.py
```diff
@@ -975,6 +976,79 @@ def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
...
+        manager.errors.report(0, 0, "Error writing protocol dependencies cache.",
+                              blocker=True)
+
+
+def read_protocol_cache(manager: BuildManager,
+                        graph: Graph) -> Optional[Dict[str, Set[str]]]:
+    """Read and validate protocol dependencies cache."""
+    proto_meta, proto_cache = get_protocol_deps_cache_name(manager)
+    try:
+        with open(proto_meta, 'r') as f:
```
JukkaL: Use file system cache here (and elsewhere in this function)?

mypy/subtypes.py
```diff
@@ -392,6 +392,7 @@ def f(self) -> A: ...
     as well.
     """
     assert right.type.is_protocol
+    TypeState.record_protocol_subtype_check(left.type, right.type)
```
JukkaL: Please add short comment explaining why we do this ("We need to record this check to generate fine-grained dependencies." or something).

mypy/typestate.py
```diff
@@ -29,6 +33,26 @@ class TypeState:
...
+    proto_deps = {}  # type: Optional[Dict[str, Set[str]]]
+
+    # Protocols (full names) a given class attempted to implement.
+    # Used to calculate fine grained protocol dependencies and optimize protocol
+    # subtype cache invalidation in fine grained mode.
+    _attempted_protocols = {}  # type: Dict[TypeInfo, Set[str]]
+    # We also snapshot protocol members of the above protocols.
+    _checked_against_members = {}  # type: Dict[TypeInfo, Set[str]]
+    # TypeInfos that has been reprocessed since latest dependency snapshot update.
+    _reprocessed_types = set()  # type: Set[TypeInfo]
```
JukkaL: The term 'reprocessed' is a bit unclear here. Can you spell out what it means exactly? Does it only apply to fine-grained incremental mode?

mypy/dmypy_server.py
ilevkivskyi: OK, I will make it a global function in `typestate`.

mypy/typestate.py
ilevkivskyi: > Does it only apply to fine-grained incremental mode?

Essentially yes, this is is an optimization to avoid gathering dependencies from all registered types, but only from those re-checked during last update. For full build, this simply contains all registered types, we can't optimize this.

ilevkivskyi:
@JukkaL Thanks for review! I addressed all your comments and also made few minor updates (for better separation between normal deps and protocol deps, now they are mixed only immediately before using, or dumping in a test).

mypy/build.py
```diff
@@ -962,6 +962,77 @@ def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
...
+def read_protocol_cache(manager: BuildManager,
+                        graph: Graph) -> Optional[Dict[str, Set[str]]]:
+    """Read and validate protocol dependencies cache."""
+    proto_meta, proto_cache = get_protocol_deps_cache_name(manager)
+    try:
+        data = manager.fscache.read(proto_meta).decode()
+        manager.trace('Proto meta {}'.format(data.rstrip()))
+        meta_snapshot = json.loads(data)
+    except IOError:
+        manager.log('Could not load protocol meta: could not find {}'.format(proto_meta))
```
JukkaL: `IOError` can be raised due to various reasons even if the file exists, so the message "could not find ..." may not be accurate. Maybe just log something like "Could not load protocol metadata: <path>".

mypy/build.py
```diff
@@ -962,6 +962,77 @@ def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
...
+        return None
+    current_meta_snapshot = {}  # type: Dict[str, str]
+    for id in graph:
+        meta = graph[id].meta
+        assert meta is not None, 'Protocol cache should be read after all other.'
```
JukkaL: Style nit: Remove period from the end of the message.

mypy/build.py
```diff
@@ -962,6 +962,77 @@ def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
...
+        current_meta_snapshot[id] = meta.hash
+    common = set(meta_snapshot.keys()) & set(current_meta_snapshot.keys())
+    if any(meta_snapshot[id] != current_meta_snapshot[id] for id in common):
+        # TODO: invalidate also if options changed (like --strict-optional)?
+        manager.log('Protocol cache inconsistent, ignoring.')
```
JukkaL: Style nit: Remove period from the end of the message.

mypy/build.py
```diff
@@ -962,6 +962,77 @@ def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
...
+        return None
+    try:
+        data = manager.fscache.read(proto_cache).decode()
+        manager.trace('Proto deps {}'.format(data.rstrip()))
```
JukkaL: Like above, move this line and the following line outside the try block.

mypy/build.py
```diff
@@ -962,6 +962,77 @@ def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
...
+        deps = json.loads(data)  # TODO: Errors
+    except IOError:
+        manager.log('Could not load protocol cache: could not find {}'.format(proto_cache))
```
JukkaL: Similar to above, the error message "could not find ..." may not be accurate.

mypy/build.py
```diff
@@ -962,6 +962,77 @@ def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
...
+        return None
+    if not isinstance(deps, dict):
+        manager.log('Could not load protocol cache: cache is not a dict: {}'
+                    .format(repr(deps)))
```
JukkaL: There could be lots of deps, maybe just print out `type(deps)`?

mypy/build.py
```diff
@@ -962,6 +962,77 @@ def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
...
+    for id in graph:
+        meta = graph[id].meta
+        assert meta is not None, 'Protocol cache should be read after all other.'
+        current_meta_snapshot[id] = meta.hash
+    common = set(meta_snapshot.keys()) & set(current_meta_snapshot.keys())
+    if any(meta_snapshot[id] != current_meta_snapshot[id] for id in common):
+        # TODO: invalidate also if options changed (like --strict-optional)?
+        manager.log('Protocol cache inconsistent, ignoring.')
+        return None
+    try:
```
JukkaL: This try statement looks pretty similar to what was above. Maybe create a utility function that reads a JSON file or something to avoid repetition?

mypy/build.py
```diff
@@ -962,6 +962,77 @@ def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
...
+        manager.log('Could not load protocol cache: could not find {}'.format(proto_cache))
+        return None
+    if not isinstance(deps, dict):
+        manager.log('Could not load protocol cache: cache is not a dict: {}'
+                    .format(repr(deps)))
+        return None
+    return {k: set(v) for (k, v) in deps.items()}
+
+
+def _get_prefix(manager: BuildManager, id: Optional[str] = None) -> str:
```
JukkaL: Maybe rename to `_cache_dir_prefix` or something for more clarity/

mypy/build.py
```diff
@@ -988,6 +1057,17 @@ def get_cache_names(id: str, path: str, manager: BuildManager) -> Tuple[str, str
     return (prefix + '.meta.json', prefix + '.data.json', deps_json)
 
 
+def get_protocol_deps_cache_name(manager: BuildManager) -> Tuple[str, str]:
+    """Return file names for fine grained protocol dependencies cache.
+
+    Since these dependencies represent a global state of the program, they
+    are serialized per program, not per module, and the corresponding files
+    live at the root of the cache folder for a given Python version.
```
JukkaL: Maybe document that returns tuple (meta file path, data file path) and describe what's included in the meta file.

mypy/build.py
```diff
@@ -2237,8 +2317,24 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
     # just want to load in all of the cache information.
     if manager.use_fine_grained_cache():
         process_fine_grained_cache_graph(graph, manager)
+        # Fine grained protocol dependencies are serialized separately, so we read them
+        # after we loaded cache for whole graph. The `read_protocol_cache` will also validate
+        # the protocol cache against the loaded individual cache files.
+        TypeState.proto_deps = read_protocol_cache(manager, graph)
+        if TypeState.proto_deps is None and manager.stats.get('fresh_trees', 0):
```
JukkaL: Maybe write the latter condition could as `manager.stats.get(..., 0) > 0` to be more explicit? Also mention this in the comment, since it looked a bit confusing at first. Also, it could be helpful to document the `stats` attribute (in `BuildManager`, not here) so that it's clear that we rely on the stats for correctness.

mypy/build.py
```diff
@@ -2237,8 +2317,24 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
...
+            manager.errors.set_file(_get_prefix(manager), None)
+            manager.errors.report(0, 0, "Error reading protocol dependencies cache. Aborting.",
```
JukkaL: Again, leave out the final period (maybe like this: "Error reading protocol dependencies cache -- aborting").

mypy/build.py
```diff
@@ -2237,8 +2317,24 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
...
     else:
         process_graph(graph, manager)
+        if manager.options.cache_fine_grained or manager.options.fine_grained_incremental:
+            # If we are running a daemon or are going to write cache for further fine grained use,
+            # then we need to collect fine grained protocol dependencies.
+            # Since these are a global property of the program, they are calculated after we
+            # processed the whole graph.
+            TypeState.update_protocol_deps()
+            if TypeState.proto_deps is not None:
+                write_protocol_deps_cache(TypeState.proto_deps, manager, graph)
```
JukkaL: For consistency, we probably shouldn't write any cache files in fine-grained incremental mode.

mypy/server/update.py
```diff
@@ -944,12 +960,18 @@ def update_deps(module_id: str,
...
 
 def lookup_target(manager: BuildManager,
-                  target: str) -> List[DeferredNode]:
-    """Look up a target by fully-qualified name."""
+                  target: str) -> Tuple[List[DeferredNode], Optional[TypeInfo]]:
+    """Look up a target by fully-qualified name.
 
+    The first item in the return tuple is a list of deferred nodes that
+    needs to be reprocessed. If the target represents a TypeInfo corresponding
+    to a protocol. Return it as a second item in the return tuple, otherwise None.
```
JukkaL: Grammar: Should be "... to a protocol, return it as ..."

mypy/nodes.py
```diff
@@ -2211,6 +2208,18 @@ def get_containing_type_info(self, name: str) -> 'Optional[TypeInfo]':
                 return cls
         return None
 
+    @property
+    def protocol_members(self) -> List[str]:
+        # Protocol members are names of all attributes/methods defined in a protocol
```
JukkaL: Should this return an empty list if the class is not a protocol, similar to how this was defined earlier?

mypy/typestate.py
```diff
@@ -3,22 +3,26 @@
...
     from typing import ClassVar
 from mypy.nodes import TypeInfo
 from mypy.types import Instance
+from mypy.server.trigger import make_trigger
 
 
 class TypeState:
     """
     This class provides subtype caching to improve performance of subtype checks.
+    As well it holds protocol fine grained dependencies.
```
JukkaL: Style nit: There should be an empty line after the first line of the docstring (also we tend to write it as `"""<first line>` (no newline after `"""`).

Grammar nit: Would be better to write as "It also holds ...".

mypy/typestate.py
```diff
@@ -3,22 +3,26 @@
...
 
     Note: to avoid leaking global state, 'reset_all_subtype_caches()' should be called
-    after a build has finished and after a daemon shutdown. This class only exists for
+    after a build has finished and after a daemon shutdown. This subtype cache only exists for
     performance reasons, resetting subtype caches for a class has no semantic effect.
+    The protocol dependencies however are only stored here, and shouldn't be deleted unless
+    no more needed (e.g. during daemon shutdown).
```
JukkaL: Grammar: no more needed -> not needed any more.

mypy/typestate.py
```diff
@@ -29,6 +33,31 @@ class TypeState:
...
 
+    # This contains protocol dependencies generated after running a full build,
+    # or after an update. These dependencies are special because:
+    #   * They are a global property of the program; i.e. some dependencies for imported
+    #     classes can be generated in the importing modules.
+    #   * Because of the above, they are serialized separately, after a full run,
+    #     or a full update.
+    # `proto_deps` can be None if after deserialization it turns out that they are
+    # inconsistent with the other cache files (or an error occurred during deserialization).
+    # A blocking error will be generated in this case, since we can't proceed safely.
```
JukkaL: Document what kinds of dependencies are stored here (of the 3 kinds discussed in the PR message) and give examples. Preferably have real code fragments, and explain exactly which dependencies would be generated and why.

Potential comment update: Can they also be `None` if we aren't using fine-grained incremental checking?

mypy/typestate.py
```diff
@@ -29,6 +33,31 @@ class TypeState:
...
+    proto_deps = {}  # type: Optional[Dict[str, Set[str]]]
+
+    # Protocols (full names) a given class attempted to implement.
+    # Used to calculate fine grained protocol dependencies and optimize protocol
+    # subtype cache invalidation in fine grained mode.
+    _attempted_protocols = {}  # type: Dict[TypeInfo, Set[str]]
```
JukkaL: Should we use the full name instead of a `TypeInfo` reference here as the key? References are harder to reason about since they could refer to per-merged nodes if we are not careful.

Give a concrete example of what's stored here. For example, if we pass a value of type `a.A` to a function expecting something compatible with protocol `p.P`, we'd have a `a.A -> {'p.P', ...}` in the map.

Maybe also discuss the lifetime of this. Does this get flushed after every update?

mypy/typestate.py
```diff
@@ -29,6 +33,31 @@ class TypeState:
...
+    # We also snapshot protocol members of the above protocols.
+    _checked_against_members = {}  # type: Dict[TypeInfo, Set[str]]
```
JukkaL: Again a more detailed explanation would be helpful here. A concrete example would be nice and a explanation why this is needed.

mypy/typestate.py
```diff
@@ -29,6 +33,31 @@ class TypeState:
...
+    # TypeInfos that has been type-checked since latest dependency snapshot update.
```
JukkaL: Grammar nit: "TypeInfos that *have* been ..."

mypy/typestate.py
```diff
@@ -29,6 +33,31 @@ class TypeState:
...
+    # This is an optimisation for fine grained mode; during a full run we only take
+    # a dependency snapshot at the very end, so this set will contain all type-checked
+    # TypeInfos. After a fine grained update however, we can gather new dependencies only
+    # from few TypeInfos that were type-checked during this update, because these are
+    # the only that can generate new protocol dependencies.
```
JukkaL: This comment was a bit unclear. Which dependencies does the comment refer to (of the three kinds)? It's not clear why only type-checked TypeInfos can generate new dependencies, since elsewhere we've established that dependencies are a global property of a program. Please elaborate a bit. What does type-checked mean exactly here? It could plausibly mean, for example, some of 1) triggered `ClsName` 2) triggered `<ClsName>` 3) file containing the class was modified 4) the target containing the class definition was processed -- but I suspect it means none of these.

Style/grammar nits: "from few TypeInfos" -> "from the (typically) few TypeInfos". "are the only" -> "are the only ones"

mypy/typestate.py
```diff
@@ -62,3 +91,99 @@ def record_subtype_cache_entry(cls, left: Instance, right: Instance) -> None:
     @classmethod
     def record_proper_subtype_cache_entry(cls, left: Instance, right: Instance) -> None:
         cls._subtype_caches_proper.setdefault(right.type, set()).add((left, right))
+
+    @classmethod
+    def reset_protocol_deps(cls) -> None:
+        cls.proto_deps = {}
```
JukkaL: Mention when this is expected to be run in a docstring.

mypy/typestate.py
```diff
@@ -62,3 +91,99 @@ def record_subtype_cache_entry(cls, left: Instance, right: Instance) -> None:
...
+            x: Proto = C()
+
+        the following dependencies will be generated:
+            1. ..., <SuperProto[wildcard]>, <Proto[wildcard]> -> <Proto>
+            2. ..., <B.attr>, <C.attr> -> <C> [for every attr in Proto members]
+            3. <C> -> Proto  # this one to invalidate the subtype cache
+
+        The first kind is generated immediately per-module in deps.py. While two other kinds are
+        generated here after all modules are type checked anf we have recorded all the subtype
+        checks.
```
JukkaL: Can you give examples where each of these dependency kinds are needed to produce correct type checking results?

mypy/typestate.py
```diff
@@ -62,3 +91,99 @@ def record_subtype_cache_entry(cls, left: Instance, right: Instance) -> None:
...
+    def update_protocol_deps(cls, second_map: Optional[Dict[str, Set[str]]] = None) -> None:
+        """Update global protocol dependency map.
+
+        We update the global map incrementally, using a snapshot only from recently
+        type checked types. If second_map is given, update it as well. This is currently used
+        by FineGrainedBuildManager that maintains normal (non-protocol) dependencies.
+        """
+        if cls.proto_deps is None:
+            # Unsuccesful cache loading, nothing to do.
+            return
```
JukkaL: Is silently failing here the right thing to do? Should we instead assert that `cls.proto_deps` is not-`None`?

mypy/typestate.py
```diff
@@ -62,3 +91,99 @@ def record_subtype_cache_entry(cls, left: Instance, right: Instance) -> None:
...
+    @classmethod
+    def add_all_protocol_deps(cls, deps: Dict[str, Set[str]]) -> None:
+        """Add all known protocol dependencies to deps.
+
+        This is used by tests and debug output, and also when passing
+        all collected or loaded dependencies on to FineGrainedBuildManager
+        in its __init__.
+        """
+        cls.update_protocol_deps()  # just in case
+        if TypeState.proto_deps is None:
```
JukkaL: Similar to above, silently failing seems kind of suspicious. I'd rather perform the check in the call site perhaps, and only when it's actually safe to do so.

ilevkivskyi:
@JukkaL I addressed the previous round of comments. I abandoned the idea of using the `fscache` for reading meta files, because it also caches exceptions and it is not what we want, because it is normal for meta files to not exist at the first run and then be added later, also the exception caching logic is not clear to me. Anyway, as I mentioned before, the file system cache will not give any speed-up with the current design, since protocol caches are read only once.

mypy/server/deps.py
```diff
@@ -235,6 +233,13 @@ def process_type_info(self, info: TypeInfo) -> None:
             self.add_type_dependencies(info.typeddict_type, target=make_trigger(target))
         if info.declared_metaclass:
             self.add_type_dependencies(info.declared_metaclass, target=make_trigger(target))
+        if info.is_protocol:
+            for base_info in info.mro[:-1]:
+                self.add_dependency(make_wildcard_trigger(base_info.fullname()),
```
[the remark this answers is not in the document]
ilevkivskyi: 6 tests fail if I only add this for the class but not MRO. The idea is that we can have subprotocols, so that an actual interface is determined by the whole MRO, if something is changed in an explicit superprotocol we need to invalidate the protocol. A simple example is `[InvalidateProtocolViaSuperClass]`. There are other ways how to treat this (like including protocol members in a class snapshot), but it seems to me this one is conceptually simplest and the most similar to how no-protocol classes are treated.

mypy/nodes.py
ilevkivskyi: I would say it is not necessary, but I see no particular danger in this.

mypy/typestate.py
ilevkivskyi: > Can they also be `None` if we aren't using fine-grained incremental checking?

Currently they should be `None` only in case of error, otherwise it is an empty dict.

mypy/typestate.py
ilevkivskyi: > Should we use the full name instead of a `TypeInfo` reference here as the key? 

I was not sure, but maybe it makes sense to switch to full names, I will give it a try.

> Does this get flushed after every update?

Actually these stay until the daemon shutdown. I think they can be flushed after every update because we capture them in `proto_deps`, I thought it may make sense to keep them for debugging, but maybe it is not necessary if you think they will take too much space.

mypy/typestate.py
ilevkivskyi: I think you are right, we shouldn't call this after an unsuccessful cache load, I will change to an assert here and below.

msullivan:
Currently we don't use the fscache for any of the cache files. We only use it for source files.

test-data/unit/fine-grained.test
```diff
@@ -5416,6 +5416,785 @@ class C:
...
+a.py:2: error: Incompatible types in assignment (expression has type "C", variable has type "P")
+a.py:2: note: 'C' is missing following 'P' protocol member:
+a.py:2: note:     z
+==
+==
+a.py:2: error: Incompatible types in assignment (expression has type "C", variable has type "P")
+a.py:2: note: Following member(s) of "C" have conflicts:
+a.py:2: note:     z: expected "int", got "str"
+
+[case testWeAreCarefullWithBuiltinProtocols]
```
JukkaL: Typo: Carefull.

mypy/typestate.py
```diff
@@ -62,3 +99,114 @@ def record_subtype_cache_entry(cls, left: Instance, right: Instance) -> None:
...
+
+            class A:
+                def __iter__(self) -> Iterator[int]:
+                    ...
+
+            it: Iterable[int] = A()
+
+        We add <a.A.__iter__> -> <a.A> to invalidate the assignment (module target in this case),
+        whenever the signature of a.A.__iter__ changes. We also add <a.A> -> typing.Iterable,
+        to invalidate the subtype caches of the latter.
```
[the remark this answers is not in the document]
ilevkivskyi: By the way, I think I don't have any tests about meets, joins, and constraint inference with protocols in fine grained mode. And I am not sure there are such tests even for normal classes. Do you know if we have such tests for non-protocols?

test-data/unit/fine-grained.test
```diff
@@ -5416,6 +5416,785 @@ class C:
...
+    g(b.C())
+[file b.py]
+class C:
+    x: int
+[file b.py.2]
+class C:
+    x: str
+[out]
+==
+a.py:8: error: Argument 1 to "g" has incompatible type "C"; expected "P"
```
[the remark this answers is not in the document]
ilevkivskyi: I remember most of these added for a reason plus few variations in target, but I don't remember now which exactly were just variations. I think there shouldn't be many, if you are sure about certain test cases, then I can delete them.

ilevkivskyi:
I think all issues are addressed now and the follow-up issues are opened, so I am going to merge this now. Thanks @JukkaL for thorough review!

PR #4800, opened 2018-03-27
Make incremental mode the default

msullivan:


docs/source/command_line.rst
```diff
@@ -342,6 +342,9 @@ Here are some more useful flags:
   when most parts of your program haven't changed since the previous
   mypy run.
 
+  Incremental mode is now the default and may be disabled with
```
gvanrossum: This is not a news entry, I'd leave "now" off.

Also could you regenerate the option summary at the top of this file? There's a `[-i]` tucked in there somewhere that probably should go, and `[--no-incremental]` should appear.

docs/source/command_line.rst
```diff
@@ -342,6 +342,9 @@ Here are some more useful flags:
...
+  ``--no-incremental``.
+
 - ``--cache-dir DIR`` is a companion flag to ``-incremental``, which
```
ethanhs: Since incremental mode is the default, this should probably be updated, as people usually won't be passing `--incremental` itself. Also above, the first `--incremental` after the section heading should probably be changed to `Incremental mode`.

msullivan:
Made the suggested doc updates and rebased, but we still want  #4841 to go in first

PR #4802, opened 2018-03-27
Don't write cache files in fine-grained incremental mode

JukkaL:
I checked that this only affects one fine-grained test case that
uses cache, testAddModuleAfterCache3_cached-skip-nocache. This might
only affect cases where many cache files are missing.

I've seen cases were apparently some cache files are missing, and this
might help with the issue. However, this is bit of a shot in the dark.


msullivan:
Yeah, this will only affect cases where many cache files are missing and so fine-grained falls back to doing a traditional `process_graph`.

PR #4806, opened 2018-03-27
Fix some stale_info() related crashes by giving it a proper mro

msullivan:
`join` expects any pair of Instances to have a least upper bound, and
contains assertions to this effect. This is true because all instances
should eventually descend from object.  To preserve this invariant,
make the `stale_info()` TypeInfos properly descend from
builtins.object.

This fixes a fine-grained incremental crash observed in the wild
as well as a quick-and-dirty crash.

Bogus TypeInfo's are generated by `fixup.stale_info()` in
quick-and-dirty mode (when there is a reference to a deleted item) and
when loading from a fine fine-grained cache (when a module has been
deleted).

This changes the behavior of a couple --quick-and-dirty tests. But
since the whole premise of --quick-and-dirty is that it is wrong in
complicated cases (which certainly includes deleting objects referened
through an import cycle!), it doesn't really matter what the actual
output is, just that we don't *crash*.

### DIFF
--- a/mypy/fixup.py
+++ b/mypy/fixup.py
@@ -86,15 +86,15 @@ class NodeFixer(NodeVisitor[None]):
                         value.type_override = stnode.type_override
                         if (self.quick_and_dirty and value.kind == TYPE_ALIAS and
                                 stnode.type_override is None):
-                            value.type_override = Instance(stale_info(), [])
+                            value.type_override = Instance(stale_info(self.modules), [])
                         value.alias_tvars = stnode.alias_tvars or []
                     elif not self.quick_and_dirty:
                         assert stnode is not None, "Could not find cross-ref %s" % (cross_ref,)
                     else:
                         # We have a missing crossref in quick mode, need to put something
-                        value.node = stale_info()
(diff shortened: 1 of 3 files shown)


mypy/fixup.py
```diff
@@ -290,12 +290,14 @@ def lookup_qualified_stnode(modules: Dict[str, MypyFile], name: str,
...
     suggestion = "<stale cache: consider running mypy without --quick>"
     dummy_def = ClassDef(suggestion, Block([]))
     dummy_def.fullname = suggestion
 
     info = TypeInfo(SymbolTable(), dummy_def, "<stale>")
-    info.mro = [info]
-    info.bases = []
+    obj_type = lookup_qualified(modules, 'builtins.object', False)
+    assert isinstance(obj_type, TypeInfo)
+    info.bases = [Instance(obj_type, [])]
```
ilevkivskyi: Is this really necessary?

msullivan: Yeah, `join` looks explicitly at `bases`.

PR #4809, opened 2018-03-28
Clear the traceback when caching exceptions, to avoid creating reference cycles

### LINKED ISSUE #4807: FileSystemCache leaks memory via __traceback__
FileSystemCache caches exceptions, which I think means it also caches `__traceback__`. This is bad, because:

* `__traceback__` is a big memory leak (citation needed)
* The traceback is not representative of the actual failing call

This might be to blame for the OOM in #4693

eric-wieser:
The traceback isn't super useful here anyway.
I'm not sure if the effect of this is significant.

Fixes #4807

### DIFF
--- a/mypy/fscache.py
+++ b/mypy/fscache.py
@@ -54,6 +54,7 @@ class FileSystemMetaCache:
         try:
             st = os.stat(path)
         except Exception as err:
+            err.__traceback__ = None
             self.stat_error_cache[path] = err
             raise
         self.stat_cache[path] = st
@@ -67,6 +68,7 @@ class FileSystemMetaCache:
         try:
             results = os.listdir(path)
         except Exception as err:
+            err.__traceback__ = None
             self.listdir_error_cache[path] = err
             raise err
         self.listdir_cache[path] = results


ethanhs:
Hm, I was curious (because I was planning on doing a similar change), so I went ahead and used [memory_profiler](https://pypi.python.org/pypi/memory_profiler) to check. My exact command line was `mprof run --include-children runtests.py`.

This is master:
master

And this is this branch:
branch_mem

This is the branch of #4693:
pep561branch

It seems to me that neither this branch, nor completely removing caching exceptions (which is what my branch does) changes the total memory usage significantly.

gvanrossum:
Am I the only one reminded of The Little Prince here? :-)

eric-wieser:
Thanks for the graphs, @ethanhs!

There looks to be a pretty substantial improvement in memory usage at startup on this branch, but it does look the whole graph is just shifted to the right - so maybe it was some artifact of the measuring process.

I'm afraid that reference is completely lost on me, @gvanrossum, so you might be! (although I'd appreciate an explanation of the connection)

eric-wieser:
> mypy/fscache.py:71: error: Incompatible types in assignment (expression has type "None", variable has type "TracebackType")

Looks like typeshed needs an update too

JukkaL:
The memory usage would likely only be a (potential) problem in cases where there are a large number of IO errors for whatever reason. The test suite likely won't hit many error cases so I'm not surprised that the memory impact is small.

eric-wieser:
>  The test suite likely won't hit many error cases

Every invocation of `fscache.isfile` causes an error to be cached, which I imagine the testcases do hit.

ethanhs:
I'm going to go ahead and close this since I don't think we are going ahead with it. (Let me know if that is wrong).

eric-wieser:
Even if there is no substantial memory improvement, is there any reason not to merge this? If nothing else, even the most simple traceback gives `sys.getsizeof(e.__traceback__.tb_frame) == 536` on my machine, which seems like a lot to associate with every filename that doesn't exist.

JukkaL:
Memory use when running the tests isn't really my main concern -- we will soon be launching a mypy daemon, and optimizing the memory use of the daemon is pretty important, since it will often be running for a long time. However, it would be nice to see some concrete measurement of the potential impact. For example, you could call `fscache.isfile` against 1M nonexistent paths using a script and measure the memory use of the script with and without this change.

mypy/fscache.py
```diff
@@ -54,6 +54,7 @@ def stat(self, path: str) -> os.stat_result:
         try:
             st = os.stat(path)
         except Exception as err:
+            err.__traceback__ = None
```
JukkaL: This didn't actually reclaim the space used by the traceback in my experiments. I tried something different and was able to save some memory use. I'll create a PR once I've verified that my approach works.

eric-wieser: Even after a `gc.collect()`?

JukkaL:
Closing in favor of #4902.

Issue #4565, opened 2018-02-12
Function overloading example does not typecheck


tyrion:
The second code example given on the [function overloading documentation](http://mypy.readthedocs.io/en/latest/function_overloading.html)  gives the error (referred to the last definition of `__getitem__`): `Signature of __getitem__ incompatible with supertype Sequence`.

The whole snippet is:
```python
from typing import overload, Sequence, TypeVar, Union
T = TypeVar('T')

class MyList(Sequence[T]):
    @overload
    def __getitem__(self, index: int) -> T:
        pass  # Don't put code here

    @overload
    def __getitem__(self, index: slice) -> Sequence[T]:
        pass  # Don't put code here

    def __getitem__(self, index: Union[int, slice]) -> Union[T, Sequence[T]]:
        raise NotImplementedError
```

ilevkivskyi:
I think this is because mypy checks the _implementation_ signature against base class (while it shouldn't). If I remove the implementation annotation the error goes away.

PR #4810, opened 2018-03-28
Removes incorrect override check on overloaded methods

Michael0x2a:
This commit fixes https://github.com/python/mypy/issues/4565

Currently, when we override a method and annotate the implementation, mypy appears to check the legality of the override twice: once for the overloads, and once for the implementation.

This second check is unnecessary and causes a spurious error.


Michael0x2a:
@ilevkivskyi I added another test -- let me know if you think I should add more.

PR #4813, opened 2018-03-28
Add tests for overloads in fine grained mode

ilevkivskyi:
This PR adds tests for overloads in fine grained mode and fixes two (rather minor) uncovered crashes.

During work on this something came to my mind. The `TypeVarScope` and all the code around function type variables binding needs significant clean-up or refactoring, maybe we should add this to our plans?

### DIFF
--- a/mypy/server/astmerge.py
+++ b/mypy/server/astmerge.py
@@ -360,6 +360,7 @@ class TypeReplaceVisitor(SyntheticTypeVisitor[None]):
     def visit_overloaded(self, t: Overloaded) -> None:
         for item in t.items():
             item.accept(self)
+        # Fallback can be None for overloaded types that haven't been semantically analyzed.
         if t.fallback is not None:
             t.fallback.accept(self)
 
--- a/mypy/server/astmerge.py
+++ b/mypy/server/astmerge.py
@@ -360,7 +360,8 @@ class TypeReplaceVisitor(SyntheticTypeVisitor[None]):
     def visit_overloaded(self, t: Overloaded) -> None:
         for item in t.items():
             item.accept(self)
-        t.fallback.accept(self)
+        if t.fallback is not None:
+            t.fallback.accept(self)
 
     def visit_deleted_type(self, typ: DeletedType) -> None:
         pass
--- a/mypy/server/aststrip.py
+++ b/mypy/server/aststrip.py
@@ -107,7 +107,12 @@ class NodeStripVisitor(TraverserVisitor):
         if not self.recurse_into_functions:
             return
         node.expanded = []
+        print('func', node.type, node.unanalyzed_type)
         node.type = node.unanalyzed_type
+        # Type variable binder binds tvars before the type is analized.
+        # It should be refactored, before that we just undo this change here.
+        if node.type:
+            node.type.variables = []
         with self.enter_method(node.info) if node.info else nothing():
             super().visit_func_def(node)
 
--- a/mypy/server/aststrip.py
+++ b/mypy/server/aststrip.py
@@ -109,8 +109,9 @@ class NodeStripVisitor(TraverserVisitor):
             return
         node.expanded = []
         node.type = node.unanalyzed_type
-        # Type variable binder binds tvars before the type is analized.
+        # Type variable binder binds tvars before the type is analyzed.
         # It should be refactored, before that we just undo this change here.
+        # TODO: this will be not necessary when https://github.com/python/mypy/issues/4814 is fixed.
         if node.type:
             assert isinstance(node.type, CallableType)
             node.type.variables = []
--- a/mypy/server/aststrip.py
+++ b/mypy/server/aststrip.py
@@ -111,7 +111,7 @@ class NodeStripVisitor(TraverserVisitor):
         node.type = node.unanalyzed_type
         # Type variable binder binds tvars before the type is analyzed.
         # It should be refactored, before that we just undo this change here.
-        # TODO: this will be not necessary when https://github.com/python/mypy/issues/4814 is fixed.
+        # TODO: this will be not necessary when #4814 is fixed.
         if node.type:
             assert isinstance(node.type, CallableType)
             node.type.variables = []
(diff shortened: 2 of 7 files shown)


test-data/unit/fine-grained.test
```diff
@@ -4854,3 +4854,321 @@ class C:
...
+def f(x: int) -> None: pass
+@overload
+def f(x: str) -> int: pass
+def f(x):
+    pass
+[out]
+==
+a.py:3: error: Incompatible types in assignment (expression has type "int", variable has type "str")
+
+[case testOverloadsSimpleTo]
```
JukkaL: I'd recommend including something that refers to nested functions in the test case description, since nested functions are somewhat special.

test-data/unit/fine-grained.test
```diff
@@ -4854,3 +4854,321 @@ class C:
...
+import a
```
JukkaL: The `import a` / `[file a.py]` seems unnecessary. You can just move the contents of `a.py` to the main module.

test-data/unit/fine-grained.test
```diff
@@ -4854,3 +4854,321 @@ class C:
...
+    pass
+[file mod.py.2]
+def f() -> str:
+    pass
+[out]
+==
+a.py:9: error: Incompatible types in assignment (expression has type "str", variable has type "int")
+
+[case testOverloadsRemovedOverload]
+import a
```
JukkaL: Again `a.py` seems redundant -- you can move the contents to the main module.

test-data/unit/fine-grained.test
```diff
@@ -4854,3 +4854,321 @@ class C:
...
+@overload
+def f(x: C) -> int: pass
+def f(x):
+    pass
+[out]
+==
+a.py:3: error: No overload variant of "f" matches argument types [builtins.str]
+
+[case testOverloadsDeleted]
+import a
```
JukkaL: Again, merge `a.py` and the main module. (Also in various test cases below.)

test-data/unit/fine-grained.test
```diff
@@ -4854,3 +4854,321 @@ class C:
...
+def f(x: int) -> None: pass
+@overload
+def f(x: str) -> str: pass
+def f(x):
+    pass
+[file mod.py.2]
+from typing import overload
+[out]
+==
+a.py:3: error: "object" has no attribute "f"
```
JukkaL: Hint: You can use `[builtins fixtures/module.pyi]` to fix the `"object"` in the error message.

test-data/unit/fine-grained.test
```diff
@@ -4854,3 +4854,321 @@ class C:
...
+
+[case testOverloadsUpdatedTypeRecheckImplementation]
+import a
+[file a.py]
+from typing import overload
+import mod
+def outer() -> None:
+    @overload
```
JukkaL: I'd rather have a non-nested overload since it's by far the most common scenario.

test-data/unit/fine-grained.test
```diff
@@ -4854,3 +4854,321 @@ class C:
...
+==
+a.py:9: error: Incompatible types in assignment (expression has type "int", variable has type "str")
+
+[case testOverloadsUpdatedTypeRechekConsistency]
+import a
+[file a.py]
+from typing import overload
+import mod
+def outer() -> None:
+    @overload
```
JukkaL: Again, the nested overloaded function seems unnecessary. Maybe use a method instead? I'd expect them to be much more common.

test-data/unit/fine-grained.test
```diff
@@ -4854,3 +4854,321 @@ class C:
...
+[file mod.py.2]
+from typing import overload
+def f(x: int) -> int:
+    pass
+[out]
+==
+a.py:3: error: Incompatible types in assignment (expression has type "int", variable has type "str")
+a.py:3: error: Argument 1 to "f" has incompatible type "str"; expected "int"
+
+[case testOverloadsToOverloaded]
```
JukkaL: Test case name doesn't seem to describe the test case.

test-data/unit/fine-grained.test
```diff
@@ -4854,3 +4854,321 @@ class C:
...
+[file b.py.2]
+from typing import overload, Any
+class C: pass
+class Parent:
+    @overload
+    def f(self, arg: int) -> int: ...
+    @overload
+    def f(self, arg: str) -> C: ...
+    def f(self, arg: Any) -> Any: ...
+[out]
```
JukkaL: Maybe add a few additional overloaded method test cases?

mypy/server/astmerge.py
```diff
@@ -360,7 +360,8 @@ def visit_callable_type(self, typ: CallableType) -> None:
     def visit_overloaded(self, t: Overloaded) -> None:
         for item in t.items():
             item.accept(self)
-        t.fallback.accept(self)
+        if t.fallback is not None:
```
JukkaL: When can this be `None`? It sounds like a potential bug, and cause trouble elsewhere. Would it be possible to ensure that `fallback` always gets initialized with a non-`None` value before we reach AST merge? At least add a TODO comment.

mypy/server/aststrip.py
```diff
@@ -108,6 +109,11 @@ def visit_func_def(self, node: FuncDef) -> None:
             return
         node.expanded = []
         node.type = node.unanalyzed_type
+        # Type variable binder binds tvars before the type is analized.
+        # It should be refactored, before that we just undo this change here.
```
JukkaL: It's worth adding a TODO comment about this or creating an issue.

mypy/server/astmerge.py
ilevkivskyi: Few lines above there is a comment about `CallableType.fallback`, I think it equally applies to `Overloaded`. Or you want me to add a similar comment here?

JukkaL: That sounds reasonable.

ilevkivskyi:
@JukkaL Thanks! I addressed your comments, I only left few `import a`/`[file a.py]` just in case.

Issue #4815, opened 2018-03-28
User-specific config file


OmeGak:
Feature request. According to my search-fu, it hasn't been discussed yet.

`mypy` with no `--config-file` specified will check for `mypy.ini` first, use to `setup.cfg` otherwise, and ultimately fallback to the built-in defaults if none of them are present. I would like to have a user-default config that would be used instead. Would it be too crazy to check for `~/.mypy.ini` before falling back to the built-in defaults?

I would find this useful for quick prototyping without having to copy&paste my preferred configuration from other projects. It's also quite a common pattern in other static analysis tools like linters.

gvanrossum:
This looks like a simple project which you can probably do yourself in a
few lines -- just submit a PR.

OmeGak:
Sure! Happy to tackle it.

dsifford:
Nice work @OmeGak 

Curious what your thoughts are on adding to this commit to support the XDG base directory specification? 

[Details here](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support#XDG_Base_Directory_specification).
[Spec here](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).

My home directory thanks you in advance :pray:

OmeGak:
Not a strong opinion on XDG base directory. It wouldn't hurt complying with the specs, I guess. Maybe you could create a new issue proposing the idea for people to discuss it first?

OmeGak:
Not a strong opinion on XDG base directory. It wouldn't hurt complying with the specs, I guess. Maybe you could create a new issue proposing the idea for people to discuss it first?

PR #4817, opened 2018-03-28
remove warn_unused_ignores from mypy_self_check

JelleZijlstra:
This check causes CI failures for typeshed PRs that fix unused ignores, such as python/typeshed#1982, python/typeshed#1881 and python/typeshed#1937. Currently, merging such PRs without leaving CI broken requires a complicated dance of merging into typeshed, syncing typeshed into mypy, and removing the ignores here. Removing the config option removes the need for coordinated changes in the mypy and typeshed repos, at the cost of perhaps leaving some stray type: ignores in the mypy codebase.

### DIFF
--- a/mypy_self_check.ini
+++ b/mypy_self_check.ini
@@ -7,7 +7,6 @@ no_implicit_optional = True
 disallow_any_generics = True
 disallow_any_unimported = True
 warn_redundant_casts = True
-warn_unused_ignores = True
 warn_unused_configs = True
 
 # needs py2 compatibility


mypy_self_check.ini
```diff
@@ -7,7 +7,6 @@ no_implicit_optional = True
 disallow_any_generics = True
 disallow_any_unimported = True
 warn_redundant_casts = True
-warn_unused_ignores = True
```
gvanrossum: We want this for mypy itself though. Can you add it back as a package-specific option?
```
[mypy-mypy.*]
<blah>
```

JelleZijlstra: I actually *don't* want it for mypy itself—the problems with the linked typeshed PRs are that they make `# type: ignore`s in mypy itself unused, and that leads to test failures in typeshed.

ethanhs:
Would a better solution perhaps be https://github.com/python/mypy/issues/4815 once implemented? That way typeshed can have its own config that disables this, while mypy itself can check that we don't get unused ignores?

gvanrossum:
IIUC #4815 would look for ~/.mypy.ini -- I don't see how typeshed could have its own.

There's a different feature request that would allow multiple mypy.ini files all over a tree, and that would work.

ethanhs:
Ah, I suppose I was optimistically hoping there could be a script that would write to that location, but perhaps that isn't possible.

PR #4820, opened 2018-03-28
fine-grained: strip is_abstract

msullivan:
Fixes a crash bug


msullivan:
I actually just hacked something up to do a fairly basic randomization (just randomizing the order of files passed in), and after accounting for differing error message order and one test that has two possible files that can block it, I haven't been able to get a failure out of it, which I find pleasantly surprising.

Possibly randomizing the order of target reprocessing might cause more grief

PR #4821, opened 2018-03-29
fine-grained: Don't crash when module imported by a __init__ is deleted

msullivan:
This could occur if p/__init__.py is importing Foo from p.q
and p.q is deleted. Fixup will see that p.q doesn't exist and
so will look for q in p. p might be an unpatched cross-ref, which
we need to handle by returning stale_info().

### DIFF
--- a/mypy/fixup.py
+++ b/mypy/fixup.py
@@ -286,6 +286,11 @@ def lookup_qualified_stnode(modules: Dict[str, MypyFile], name: str,
         if not rest:
             return stnode
         node = stnode.node
+        # In fine-grained mode, could be a cross-reference to a deleted module
+        if node is None:
+            if not quick_and_dirty:
+                assert rest, "Cannot find %s" % (name,)
+            return None
         assert isinstance(node, TypeInfo)
         names = node.names
 
--- a/test-data/unit/fine-grained-modules.test
+++ b/test-data/unit/fine-grained-modules.test
@@ -322,6 +322,26 @@ def f() -> None: pass
 ==
 ==
 
+[case testDeleteDepOfDunderInit1]
+[file p/__init__.py]
+from .foo import Foo
+[file p/foo.py]
+class Foo: pass
+[file p/__init__.py.2]
+[delete p/foo.py.2]
+[out]
+==
+
+[case testDeleteDepOfDunderInit2]
+[file p/__init__.py]
+from p.foo import Foo
+[file p/foo.py]
+class Foo: pass
+[file p/__init__.py.2]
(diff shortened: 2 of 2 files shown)


mypy/fixup.py
```diff
@@ -286,6 +286,11 @@ def lookup_qualified_stnode(modules: Dict[str, MypyFile], name: str,
         if not rest:
             return stnode
         node = stnode.node
+        # In fine-grained mode, could be a cross-reference to a deleted module
+        if node is None:
+            if not quick_and_dirty:
+                assert rest, "Cannot find %s" % (name,)
```
ilevkivskyi: I am not sure I understand why it is `assert rest` here. Shouldn't it be `assert node`? Or is this assert necessary at all?

Also another random idea, since `stale_info` was initially designed only for `--quick` mode, but now is used also in fine grained, maybe it makes sense to check that it never gets to be printed in an error message if not in `--quick` mode by having an assert in `TypeInfo.__str__` (or wherever it is converted to string for an error)?

JukkaL: The assertion seems copied from above. Since we return if `not rest` is true on line 286, this assert will always be true and can perhaps be dropped? `assert node` would always fail since we just checked for `node is None`.

mypy/fixup.py
```diff
@@ -286,6 +286,11 @@ def lookup_qualified_stnode(modules: Dict[str, MypyFile], name: str,
         if not rest:
             return stnode
         node = stnode.node
+        # In fine-grained mode, could be a cross-reference to a deleted module
+        if node is None:
```
JukkaL: The comment for `SymbolTable.node` says that `node` is None for bound type variables, but here this is not what we are expecting, so apparently that comment is not quite right. I'm assuming that if this condition is true, we should have a cross reference. Maybe check/assert that `cross_ref` attribute is not None if that's the case? Also worth updating the comment for `SymbolTable.node` to document that the value can be None in this case as well (assuming I'm right in my reasoning here).

mypy/fixup.py
msullivan: Oh, oops. This is supposed to be `assert node`, which will always fail. That seems to be the existing idiom in fixup: do a check, assert the check if it is not quick_and_dirty, return a failure if it is quick_and_dirty.

msullivan: Some sort of check to prevent it seeping into error messages would be good, but it's not super obvious what the best way would be. Putting an assert in `__str__` I don't think would work, since it *might* get converted to an error string, but the errors should always get replaced with something else while reprocessing

Issue #4789, opened 2018-03-24
Variables declared as Type[Foo] doesn't support operators from Foo's metaclass


dgelessus:
When a variable's type is declared as `Type[Foo]`, and `Foo` has a metaclass that defines an operator, attempting to use that operator on the variable falsely causes an error. The metaclass methods can be used; only operators are broken. The bug does not happen when the variable's type is inferred (`var = Foo` instead of `var: Type[Foo] = Foo`).

I'm using Python 3.6.4 and the current mypy version from the repo (`0.580-dev-80a9b8f8e362902a93606ea0aabf9c18184fef3e`) with no special command-line flags.

Example code demonstrating the false error:

```python
from typing import Type

class CDataMeta(type):
    def __mul__(self, other: int) -> str:
        return "foo"
    
    def __rmul__(self, other: int) -> str:
        return "rfoo"
    
    def method(self) -> None:
        pass

class CData(metaclass=CDataMeta):
    pass

foo = CData
bar: Type[CData] = CData

def get_cdata_type() -> Type[CData]:
    return CData

# Correct behavior when the operand's type is inferred:
CData * 4  # No error
foo * 4  # No error

# Incorrect behavior when the operand's type is explicitly declared:
bar * 4  # error: Unsupported operand types for * ("Type[CData]" and "int")
get_cdata_type() * 4  # error: Unsupported operand types for * ("Type[CData]" and "int")

# Correct behavior - all of the following work:
CData.method()
foo.method()
bar.method()
get_cdata_type().method()

# Correct behavior - all of the following cause errors:
CData.nope()
foo.nope()
bar.nope()
get_cdata_type().nope()
```

This is a very simplified version of the Typeshed `ctypes` stub code (see python/typeshed#1983).

ilevkivskyi:
The difference between these two:
```python
foo = CData
bar: Type[CData] = CData
```
is that the first creates a type alias, while the second defines a variable. As a result, the type of `foo` expression in runtime context is `Callable` with the constructor signature, while the type of `bar` is just `Type[...]`. So indeed only `Type[...]` is problematic with operators like `__add__`, `__mul__` etc., other types work OK.

dgelessus:
[...]

PR #4823, opened 2018-03-29
Fix operator access on Type[...]

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/4789
This also fixes typeshed problems in internal codebases.

The idea is straightforward add overlooked `elif`. A better long-term solution may be to refactor `has_member` to use `checkmember.analyze_member_access`.

I use `pythoneval` tests because the original problem only appears currently with full stubs (apparently because `int` doesn't have operator methods in fixtures).


ilevkivskyi:
@JukkaL My understanding is that @msullivan wants to move the typeshed pin before Tuesday when you are back, so I am not waiting for your review (PR is quite straightforward), but if there is something bad please let me know.

Issue #4532, opened 2018-02-01
Fine-grained: failure to propagate change to module alias


JukkaL:
We don't track changes to module aliases correctly. This test case fails currently:

```
[case testRefreshModuleAlias]
import a
a.n.x
[file a.py]
from b import m
n = m
[file b.py]
import m
[file b.py.2]
import m2 as m
[file m.py]
x = 1
[file m2.py]
[builtins fixtures/module.pyi]
[out]
==
main:2: error: Module has no attribute "x"
```

JukkaL:
Looks like we aren't generating dependencies for module aliases.

ilevkivskyi:
Actually this one may be easier to fix after symbol table nodes will be refactored.

ilevkivskyi:
Looks like another repro:
```
[case testChangeModuleToTypeAlias]
import a
x: int = a.C.x
def f(x: int) -> None:
    x = a.C.x
class A:
    def meth(self) -> None:
        def inner() -> int:
            return a.C.x
[file a.py]
import b
C = b.D
[file b.py]
import D
[file b.py.2]
class D:
    x: str
[file D.py]
x: int
[builtins fixtures/module.pyi]
[out]
```
This test doesn't show any errors on the second run while it should.

ilevkivskyi:
Note that first attempt at fixing this https://github.com/python/mypy/pull/4861 has a nice set of test cases that can be used for future fix (I will keep the branch in my fork).

PR #4825, opened 2018-03-29
Add tests for change of node kind (like TypeVar -> module) in fine grained

ilevkivskyi:
This PR adds tests for change of node kind in fine grained mode. For example `f` can be a generic function or a type alias before an update and class or module alias after. This also fixes some bugs discovered by added tests:
* Missing deps from unsubscripted alias in runtime context.
* Always strip type of `Var` nodes (those are created in first pass of semantic analysis, but with `None` type).
* Keep `ClassDef.type_vars` and `TypeInfo.type_vars` in sync (as possible).
* Always carry over `SymbolTableNode.is_aliasing` during imports.

The last one is relatively serious, because type aliases will not work correctly in fine grained if used with `import *` or `from mod import Alias`. (I actually just noticed that I use `import mod` and then `mod.Alias` in 95% of all alias fine grained cases, therefore I added few tests for the former.)

This also fixes line endings in `typevars.py`.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3284,6 +3284,12 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
     # Expressions
     #
 
+    def add_symbol_alias_deps(self, n: Optional[SymbolTableNode]) -> None:
+        """Add type alias dependencies for a RefExpr node."""
+        if n and n.is_aliasing:
+            assert n.alias_name is not None
+            self.add_type_alias_deps([n.alias_name])
+
     def visit_name_expr(self, expr: NameExpr) -> None:
         n = self.lookup(expr.name, expr)
         if n:
@@ -3291,9 +3297,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
                 self.fail("'{}' is a type variable and only valid in type "
                           "context".format(expr.name), expr)
             else:
-                if n.is_aliasing:
-                    assert n.alias_name is not None
-                    self.add_type_alias_deps([n.alias_name])
+                self.add_symbol_alias_deps(n)
                 expr.kind = n.kind
                 expr.node = n.node
                 expr.fullname = n.fullname
@@ -3471,9 +3475,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
             # else:
             #     names = file.names
             n = file.names.get(expr.name, None) if file is not None else None
-            if n and n.is_aliasing:
-                assert n.alias_name is not None
-                self.add_type_alias_deps([n.alias_name])
+            self.add_symbol_alias_deps(n)
             n = self.dereference_module_cross_ref(n)
             if n and not n.module_hidden:
                 n = self.normalize_type_alias(n, expr)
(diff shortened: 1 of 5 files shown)


mypy/server/aststrip.py
```diff
@@ -228,13 +229,16 @@ def visit_import_all(self, node: ImportAll) -> None:
...
     def visit_name_expr(self, node: NameExpr) -> None:
-        # Global assignments are processed in semantic analysis pass 1, and we
+        # Global assignments are processed in semantic analysis pass 1 [*], and we
         # only want to strip changes made in passes 2 or later.
         if not (node.kind == GDEF and node.is_new_def):
             # Remove defined attributes so that they can recreated during semantic analysis.
             if node.kind == MDEF and node.is_new_def:
                 self.strip_class_attr(node.name)
             self.strip_ref_expr(node)
+        # [*] although we always strip type
```
JukkaL: Maybe explain why this is different?

mypy/semanal.py
```diff
@@ -3283,6 +3291,9 @@ def visit_name_expr(self, expr: NameExpr) -> None:
                 self.fail("'{}' is a type variable and only valid in type "
                           "context".format(expr.name), expr)
             else:
+                if n.is_aliasing:
```
JukkaL: These 3 lines are repeated 3 times. Would it make sense to move them to a utility method?

test-data/unit/deps-types.test
```diff
@@ -875,6 +875,29 @@ class I: pass
...
+[file b.py]
+class D:
+    pass
+[out]
+<m.A> -> m.A
+<m.x> -> m
+<a.C> -> m, m.A.meth, m.f
+<b.D.__init__> -> m.f
+<b.D.__new__> -> m.f
+<b.D> -> <m.A.meth>, <m.x>, m, m.A.meth, m.f
```
JukkaL: The `<m.A.meth>` on the RHS seems redundant, since the type is used in a nested function, but it's pretty benign.

test-data/unit/fine-grained.test
```diff
@@ -4914,6 +5069,76 @@ class A: pass
...
+T = TypeVar('T')
+class A(Generic[T]):
+    def __init__(self, x: T) -> None:
+        pass
+    def f(self) -> T:
+        pass
+[file a.py.2]
+from typing import TypeVar, Generic
+T = TypeVar('T')
+class A(Generic[T]):
```
JukkaL: Maybe also test changing a module to a non-module?

test-data/unit/deps-types.test
ilevkivskyi: TBH I am not 100% sure whether it is needed or not, but avoiding it probably will cause more complications than benefits.

test-data/unit/fine-grained.test
ilevkivskyi: Good suggestion! This uncovered one more bug.

ilevkivskyi: Hm, it looks like we need special handling of module aliases.

ilevkivskyi: OK, looks like it is a know issue https://github.com/python/mypy/issues/4532

ilevkivskyi: Hm, I added another test and found a different issue, will try to fix it.

ilevkivskyi: The solution is not very trivial and also solves #4532, so I will add these tests in a separate PR, together with a fix for them.

Issue #4392, opened 2017-12-19
Add better tests for various mypy features + fine-grained incremental mode


JukkaL:
These features don't have good (or any) test coverage (but they are expected to work to a certain extent):

- [x] `# type: ignore`
- [x] Partial types
- [x] Semantic analysis 3rd pass (and the "patching" pass)
- [x] Handle change of the kind of a definition in a symbol table (e.g. class -> module)
- [x] `NewType`
- [x] `TypedDict`
- [x] Nested classes
- [x] Nested functions
- [x] Overloaded functions
- [x] Strict optional (including per-file)
- [x] Decorators
- [x] `__new__`
- [x] Protocols
- [x] Metaclasses
- [x] Enums
- [x] Type variables 
- [x] Generic callables
- [x] Named tuples

ilevkivskyi:
Everything from this list is implemented, so this can be closed now.

PR #4826, opened 2018-03-29
Add deps and diff tests for nested functions

ilevkivskyi:
This is the last PR from the "feature completeness tests" series outlined in https://github.com/python/mypy/issues/4392

@JukkaL  I checked, nested classes are already well covered. But for nested functions, there are no deps and diff tests. So although it is rather a formality (tests are really trivial), I just add them for completeness.


ilevkivskyi:
Yes, there are quite a few tests in `fine-grained.test`, some of them are added by me for other reasons, but I like nested functions :-)

PR #4830, opened 2018-03-30
Add --always-true and --always-false flags

### LINKED ISSUE #3668: Option for force mypy to treat a constant as always true or always false
It's not uncommon to have constants such as `IS_MAC` or `IS_PROD` that are always true or false during a single program run, but the actual value depends on some potentially complex runtime check. Maybe mypy should provide a way of configuring the values of such constants. Example where this would be useful:

```py
...
if IS_MAC:  # only one of the definitions of Window should be type checked
    class Browser: ...  # mac-specific definition
else:
    class Browser: ...  # windows-specific definition
```

Mypy shouldn't generate `"Browser" redefined` errors for the second definition. It would be nice to be able to run mypy twice to type check both of the conditional definitions. Potential way to do it:

```
$ mypy --always-true=IS_MAC program.py
$ mypy --always-false=IS_MAC program.py
```

Some of these could be refactored into `sys.platform` checks that mypy already supports, but for existing code this can be a lot of work, and this doesn't work for non-platform checks such as dev vs. prod checks.

gvanrossum:
These work similarly to `if PY3` or `if MYPY`.

Fixes #3668.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -17,12 +17,14 @@ flag (or its long form ``--help``)::
               [--disallow-untyped-defs] [--disallow-incomplete-defs]
               [--check-untyped-defs] [--disallow-subclassing-any]
               [--warn-incomplete-stub] [--disallow-untyped-decorators]
-              [--warn-redundant-casts] [--no-warn-no-return] [--warn-return-any]
-              [--warn-unused-ignores] [--warn-unused-configs]
-              [--show-error-context] [--no-implicit-optional] [-i]
-              [--quick-and-dirty] [--cache-dir DIR] [--cache-fine-grained]
-              [--skip-version-check] [--strict-optional]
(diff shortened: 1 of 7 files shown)


gvanrossum:
@ilevkivskyi Yes, I will document it. And indeed the semantics are somewhat lame due to the need to work in the first pass (which is where unreachable blocks are computed).

@ethanhs I can't parse what you're saying. Are you requesting an error if the command line or config file is ambiguous, e.g. `--always-true FOO --always-false FOO`? I could do that. Storing them as sets seems overkill -- if you have more than a very small number of these you're probably doing it wrong?

ilevkivskyi:
> Are you requesting an error if the command line or config file is ambiguous, e.g. `--always-true FOO --always-false FOO`?

+1 for giving an error in this situation.

gvanrossum:
Redundancy is a much lesser sin than ambiguity. :-)

PR #4833, opened 2018-03-31
 Special case assignment to '_': always infer 'Any'

### LINKED ISSUE #465: Type error due to repeated use of underscore as dummy assignment target
Consider the following code:

``` python
from typing import Tuple

class A: pass
class B: pass
class C: pass

def f() -> Tuple[A, B, C]:
    pass

a = A()
a, _, _ = f()
```

mypy reports: `line 11: Incompatible types in assignment (expression has type "C", variable has type "B")`

Seems like mypy treats `_` as just a variable. That prevents it from being used as a dummy in assignments like the one above.

A solution might be to always give `_` the Any type.

gvanrossum:
Fixes #465. Also see discussion there about the exact specification and concerns.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1607,7 +1607,9 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
             # Set the type if the rvalue is a simple literal (even if the above error occurred).
             if len(s.lvalues) == 1 and isinstance(s.lvalues[0], NameExpr):
                 if s.lvalues[0].is_inferred_def:
-                    s.type = self.analyze_simple_literal_type(s.rvalue)
+                    # Except if it's an assignment to '_'.
+                    if s.lvalues[0].name != '_':
+                        s.type = self.analyze_simple_literal_type(s.rvalue)
         if s.type:
             # Store type into nodes.
             for lvalue in s.lvalues:
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1607,9 +1607,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
             # Set the type if the rvalue is a simple literal (even if the above error occurred).
             if len(s.lvalues) == 1 and isinstance(s.lvalues[0], NameExpr):
                 if s.lvalues[0].is_inferred_def:
-                    # Except if it's an assignment to '_'.
-                    if s.lvalues[0].name != '_':
-                        s.type = self.analyze_simple_literal_type(s.rvalue)
+                    s.type = self.analyze_simple_literal_type(s.rvalue)
         if s.type:
             # Store type into nodes.
             for lvalue in s.lvalues:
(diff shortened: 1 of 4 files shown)


test-data/unit/check-inference.test
```diff
@@ -2303,3 +2303,97 @@ class C:
     def f(self, x) -> None:
         # TODO: It would be better for the type to be Any here
         self.a.y  # E: "None" has no attribute "y"
+
+-- Special case for assignment to '_'
+-- ----------------------------------
+
+[case testUnusedTargetSingle]
+def foo() -> None:
```
ilevkivskyi: Why do you have only tests at the function scope? Is it something intentional?

test-data/unit/check-inference.test
```diff
@@ -2303,3 +2303,97 @@ class C:
...
+
+[case testUnusedTargetNotClass]
+class _:
+    pass
+_().method()  # E: "_" has no attribute "method"
+
+[case testUnusedTargetNotDef]
+def _() -> int:
+    pass
+_() + 0
```
ilevkivskyi: I would add `_() + 'hi'` to check that this generates an error.

test-data/unit/check-inference.test
gvanrossum: Oops, I thought it would all be treated the same, but it isn't. Will fix. (Since I've seen a fair number of global uses of this idiom.)

gvanrossum:
I still am hesitant about the safety rules -- I suspect there are too many different places. I'd almost rather add it behind a flag.

ilevkivskyi:
> Oh, pilot error, pushed the test updates (I hope).

Maybe now I am missing something, but I still don't see the tests updates in this PR.

gvanrossum:
I'm guessing I committed but didn't push (or the push failed and I didn't notice). I pushed again, we now have https://github.com/python/mypy/pull/4833/commits/67d84aa08a0cc8b9889ac86ca4dbe8022338d222 and the tests for that are running.

gvanrossum:
Recording a few ideas from Jukka:
- only do it on the *second* assignment?
- create a new `Var` on each assignment?
- do this analysis in semanal.py, using `is_func_scope()`?
Also need tests:
- multiple assignments to `_` on same line (e.g. `_, _, x = <tuple>`)
- make sure `import ... as _` doesn't trigger this
- make sure `_: <type> = <expr>` overrides

ilevkivskyi:
Some thoughts about these ideas (not sure if I should comment these, but just in case).

> only do it on the second assignment?

I am not sure I understand the benefit of this.

> create a new `Var` on each assignment?

I would say it is a bad idea to have two variables with the same name.

> do this analysis in semanal.py, using `is_func_scope()`?

I like this idea, this is similar to what I was thinking about this.

gvanrossum:
This is ready for another review. I moved the special-casing to semanal.py, where it seems to work fine, and limited it to locals only. There's one quirk: `except Exception as _:` doesn't treat `_` as `Any` because the checker overrides the type. The code in the checker looks a bit complex so I'd rather keep this as a limitation rather than trying to fix it. As the tests show, it works fine with the target in a `for` loop or in a `with` statement.

mypy/semanal.py
```diff
@@ -1826,6 +1826,10 @@ def analyze_lvalue(self, lval: Lvalue, nested: bool = False,
                 lval.kind = LDEF
                 lval.fullname = lval.name
                 self.add_local(v, lval)
+                if isinstance(lval, NameExpr) and lval.name == '_' and self.is_func_scope():
```
ilevkivskyi: First `isinstance()` should be unnecessary, since we are already in a huge `if isinstance(lval, NameExpr)` block. Also the last check `self.is_func_scope()` is not needed, because it is equivalent to `self.locals[-1] is not None` where we are now. But it is up to you.

PR #4840, opened 2018-04-02
Refactor import diagnostics, fix diagnostics in some fine-grained cases

### LINKED ISSUE #4798: Fine-grained mode doesn't properly respect `--follow-imports` settings when producing error messages
This test fails:
```
[case testSkipButDontIgnore]
# cmd: mypy a.py c.py
# flags: --follow-imports=skip
[file a.py]
import b
from c import x
[file b.py]
[file c.py]
x = 1
[file c.py.2]
x = '2'
[out]
==
```
producing
```
a.py:1: error: Cannot find module named 'b'
a.py:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
```
in the second iteration.
There will be similar problems with incorrect diagnostics with `--follow-imports=error`.

The root of the problem is that the `verify_dependencies` function used to produce missing modules messages during incremental updates operates solely in terms of the module graph, but producing the correct diagnostics here requires consulting the file system.

None of this is a problem if you use `--ignored-unused-imports`, of course.

msullivan:
Refactor all of the import diagnostics so that they can be used
outside of `State`'s `__init__` method.

Then use that to fix some problems with import diagnostics in fine-grained mode.
Fixes #4798.

This refactor also lays the ground work for fixing (coarse) incremental mode's interaction
with `--warn-unused-ignores`

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1492,7 +1492,7 @@ class State:
             assert id is not None
             try:
                 path, follow_imports = find_module_and_diagnose(
-                    manager, self.options, id, caller_state, caller_line,
+                    manager, id, self.options, caller_state, caller_line,
                     ancestor_for, root_source)
             except ModuleNotFound:
                 manager.missing_modules.add(id)
@@ -1910,14 +1910,15 @@ class State:
             self.interface_hash = new_interface_hash
 
     def verify_dependencies(self) -> None:
-        """Report errors for import targets in module that don't exist."""
+        """Report errors for import targets in modules that don't exist."""
         # Strip out indirect dependencies. See comment in build.load_graph().
         manager = self.manager
         dependencies = [dep for dep in self.dependencies
                         if self.priorities.get(dep) != PRI_INDIRECT]
         assert self.ancestors is not None
         for dep in dependencies + self.suppressed + self.ancestors:
-            if dep not in manager.modules and not manager.options.ignore_missing_imports:
+            options = manager.options.clone_for_module(dep)
+            if dep not in manager.modules and not options.ignore_missing_imports:
                 line = self.dep_line_map.get(dep, 1)
                 try:
                     if dep in self.ancestors:
@@ -1926,12 +1927,12 @@ class State:
                         state, ancestor = self, None
                     # Called just for its side effects of producing diagnostics.
                     find_module_and_diagnose(
-                        manager, self.options, dep,
+                        manager, dep, options,
                         caller_state=state, caller_line=line,
                         ancestor_for=ancestor)
-                except Exception:
-                    # Swallow up any exception while generating a diagnostic.
-                    # This can actually include CompileErrors that get generated in
+                except (ModuleNotFound, CompileError):
+                    # Swallow up any ModuleNotFounds or CompilerErrors while generating
+                    # a diagnostic. CompileErrors may get generated in
                     # fine-grained mode when an __init__.py is deleted, if a module
                     # that was in that package has targets reprocessed before
                     # it is renamed.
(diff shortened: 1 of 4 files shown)


mypy/build.py
```diff
@@ -2018,6 +1947,132 @@ def generate_unused_ignore_notes(self) -> None:
...
+def find_module_and_diagnose(manager: BuildManager,
+                             options: Options,
+                             id: str,
+                             caller_state: 'Optional[State]' = None,
+                             caller_line: int = 0,
+                             ancestor_for: 'Optional[State]' = None,
+                             root_source: bool = False) -> Tuple[str, str]:
+    """Find a module by name, respecting follow_imports and producing diagnostics.
+
+    """
```
JukkaL: Document the return value and arguments.

mypy/build.py
```diff
@@ -2018,6 +1947,132 @@ def generate_unused_ignore_notes(self) -> None:
...
+        errors.report(line, 0, "Cannot find module named '{}'".format(target))
+        errors.report(line, 0, '(Perhaps setting MYPYPATH '
+                      'or using the "--ignore-missing-imports" flag would help)',
+                      severity='note', only_once=True)
+    errors.set_import_context(save_import_context)
+
+
+def skipping_module(manager: BuildManager, line: int, caller_state: Optional[State],
+                    id: str, path: str) -> None:
+    assert caller_state, (id, path)
```
JukkaL: Add short docstring (no need to explain arguments in detail)?

mypy/build.py
```diff
@@ -2018,6 +1947,132 @@ def generate_unused_ignore_notes(self) -> None:
...
+                          "Import of '%s' ignored" % (id,),
+                          severity='note')
+    manager.errors.report(line, 0,
+                          "(Using --follow-imports=error, module not passed on command line)",
+                          severity='note', only_once=True)
+    manager.errors.set_import_context(save_import_context)
+
+
+def skipping_ancestor(manager: BuildManager, id: str, path: str, ancestor_for: 'State') -> None:
+    # TODO: Read the path (the __init__.py file) and return
```
JukkaL: Add short docstring (no need to explain arguments in detail)?

mypy/build.py
```diff
@@ -2008,6 +1909,34 @@ def write_cache(self) -> None:
...
+                    if dep in self.ancestors:
+                        state, ancestor = None, self  # type: (Optional[State], Optional[State])
+                    else:
+                        state, ancestor = self, None
+                    # Called just for its side effects of producing diagnostics.
+                    find_module_and_diagnose(
+                        manager, self.options, dep,
+                        caller_state=state, caller_line=line,
+                        ancestor_for=ancestor)
+                except Exception:
```
JukkaL: A catch-all except looks a bit suspicious. Would it be possible to explicitly enumerate the exceptions that we are expecting here? Some exceptions such as TypeErrors we probably don't want to silently ignore.

test-data/unit/fine-grained-modules.test
```diff
@@ -1632,3 +1632,112 @@ class Foo:
...
 ==
 a.py:3: error: Argument 1 to "foo" of "Foo" has incompatible type "int"; expected "str"
+
+[case testSkipButDontIgnore1]
+# cmd: mypy a.py c.py
+# flags: --follow-imports=skip
+[file a.py]
+import b
+from c import x
+[file b.py]
```
JukkaL: Maybe include something here that would trigger an error to verify that this actually gets skipped? Maybe also update `b.py`?

test-data/unit/fine-grained-modules.test
```diff
@@ -1632,3 +1632,112 @@ class Foo:
...
+a.py:2: error: Cannot find module named 'b'
+a.py:2: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
+
+[case testErrorButDontIgnore1]
+# cmd: mypy a.py c.py
+# flags: --follow-imports=error
+[file a.py]
+from c import x
+import b
+[file b.py]
```
JukkaL: Again maybe introduce an error here that shouldn't be reported?

test-data/unit/fine-grained-modules.test
```diff
@@ -1632,3 +1632,112 @@ class Foo:
...
+[file c.py]
+x = 1
+[file c.py.2]
+x = '2'
+[out]
+==
+a.py:2: note: Import of 'b' ignored
+a.py:2: note: (Using --follow-imports=error, module not passed on command line)
+
+-- This test fails because p.b doesn't seem to trigger p properly...
```
JukkaL: Create an issue about this (if one doesn't already exist) and mention it here.

Issue #4576, opened 2018-02-15
Support basic union math for overloaded functions


ilevkivskyi:
I think this idea appeared in context of other issues (mostly related to strict optional), but I hit this few times in another context. Consider this example:
```python
Dictable = Union[Dict[str, str], Iterable[Tuple[str, str]]]
def do(arg: Dictable) -> None:
    converted = dict(arg)  # mypy errors here, although this is valid
```
The problem is that `dict.__init__` is defined as an overload. I think we should at least support matching simple cases like:
```python
@overload
def fun(x: A) -> B: ...
@overload
def fun(x: C) -> D: ...

arg: Union[A, C]

fun(arg)  # we infer Union[B, D]
```

Michael0x2a:
I believe this is now fixed after a combination of https://github.com/python/mypy/pull/5255 and https://github.com/python/mypy/pull/5222.

Issue #3295, opened 2017-05-01
Should have a way to annotate an Optional carryover method.


euresti:
**Edited with a better example.**

Let's suppose you have a function that can take an Optional and returns something based on that or None if it's None.

```
from typing import overload, Optional

def bar(foo: Optional[int]) -> Optional[str]:
   if foo is None:
       return None
   else:
       return str(foo)

x = None  # type: Optional[int]

reveal_type(bar(None))  # Line 18
reveal_type(bar(x))
reveal_type(bar('foo'))
```

There's one problem with this being annotated this way.  If you pass a Non-Optional you get returned an Optional[str] even though it's not possible.  So I thought the new `@overload` stuff might help here but it doesn't:

```
from typing import overload, Optional

@overload
def bar(foo: int) -> str:
    ...

@overload
def bar(foo: Optional[int]) -> Optional[str]:
    ...

def bar(foo: Optional[int]) -> Optional[str]:
   if foo is None:
       return None
   else:
       return str(foo)

x = None  # type: Optional[int]

reveal_type(bar(None))
reveal_type(bar(x))
reveal_type(bar('foo'))
```

Sadly that gives me an error and the wrong types:
```
tt.py:4: error: Overloaded function signatures 1 and 2 overlap with incompatible return types
tt.py:19: error: Revealed type is 'Union[builtins.int, builtins.None]'
tt.py:20: error: Revealed type is 'Any'
tt.py:21: error: Revealed type is 'Any'
```

I also tried None as the argument to the 2nd overload.  That got me closer:

```
tt.py:19: error: Revealed type is 'Union[builtins.int, builtins.None]'
tt.py:20: error: Revealed type is 'Any'
tt.py:21: error: Revealed type is 'builtins.str'
```

I would love to be able to specify this as it would make a lot of our code easier to type.  Maybe there's a way to do it with a TypeVar but I couldn't figure it out.

gvanrossum:
And for anyone thinking that maybe type variables provide the solution, the 'str' type in the input is unrelated to the 'str' type in the output -- the real-world code from which this was derived has bytes -> str.

Also this reminds me of https://github.com/python/mypy/issues/1436.

kirbyfan64:
IIRC you can do:

```python
OptStr = TypeVar('OptStr', str, Optional[str])

# You might be able to instead do:
OptStr = TypeVar('OptStr', bound=Optional[str])
# but I'm not sure...

def bar(foo: OptStr) -> OptStr:
    # ...
[...]

Issue #1943, opened 2016-07-26
os.open doesn't accept Union[bytes, Text]


euresti:
The following code:

```
from typing import Text, Union
import os

def test(path):
    # type: (Union[bytes, Text]) -> None
    f = os.open(path, os.O_RDONLY)
```

Passes with --py2 but fails in python3 as follows:

`error: Type argument 1 of "open" has incompatible value "Union[bytes, str]"`

@JukkaL

gvanrossum:
FWIW if we change os.open() to take Union[str, bytes] the error goes away, but I'm sort of thinking that passing that union to something expecting AnyStr *should* work...

JukkaL:
Yes, `Union[str, bytes]` should probably be *sometimes* valid when `AnyStr` is expected (but it isn't right now). Example:

```
def f(x: AnyStr) -> AnyStr:
    return x

def g(x: Union[str, bytes]) -> None:
    f(x)  # should be fine, return type Union[str, bytes]
```

However, `List[Union[str, bytes]]` should not be compatible with `List[AnyStr]` because the latter is uniform but the prior could have a mix of `str` and `bytes`. Also, this shouldn't be valid, following similar reasoning:

```
def f2(x: AnyStr, y: AnyStr) -> AnyStr:
    return x + y

def g2(x: Union[str, bytes], y: Union[str, bytes]) -> None:
    f2(x, y)  # error, could be a mix of str and bytes
```

Here's how we could support these (and other things not involving type variables, such as overloads):

* If caller has arguments with union types, expand all the combinations of the union types into non-union types and test the call with each of these expansions. The call is valid if all combinations type check okay.

The call to `f2` in the second example would be checked 4 times:

* `f2(<str>, <str>)`
* `f2(<str>, <bytes>)`
* `f2(<bytes>, <str>)`
* `f2(<bytes>, <bytes>)`

The obvious problem with this is that this could be really slow if there are many union types in a call. An implementation could plausibly be clever and only expand those unions where it seems like a useful thing to do.

gnprice:
I think the right fix to @euresti's original issue is that the type of `os.open`'s parameter should not be `AnyStr` -- it should be `Union[bytes, str]`. In general, `AnyStr` doesn't provide any additional value when it appears only once in a type signature, and it introduces complexity because its semantics are somewhat complicated.

It's possible we should also handle the interaction of `AnyStr` with `Union` differently in general. I think that ought to be a separate issue if we want to pursue it further.

[...]

Issue #1322, opened 2016-04-01
Any incorrectly inferred for a call to an overloaded function


JukkaL:
This was reported by Alex Allain.

Mypy doesn't give an error for this program even though it should:

```
from typing import TypeVar, overload, List

T = TypeVar('T')

@overload
def f(x: T) -> List[T]: ...
@overload
def f(x: T, y: int) -> int: pass

def not_checked(text):
    # type: (str) -> str
    return f(*[])  # no error reported!
```

Mypy infers type `List[None]` for `[]` (which is arguably okay, though a little confusing), and this results in all overload variants matching, which is again okay. Now since multiple overload variants match, mypy incorrectly gives up and infers `Any` as the return type of `f(*[])`, even though in this case we could infer a more precise type: a union of all return types of all overload variants, i.e. `Union[List[None], int]`. Alternatively, mypy could give up and say that the return type is ambiguous, but it's less clear when this would be an okay thing to do (if ever).

gvanrossum:
Why is it okay if multiple overloads match? I thought mypy had some check that warns if overlaps are possible when the overload is defined, so I'd expect that that would produce an error (and I commend Alex for having found an example that matches multiple overloads).

Given that you generally don't like unions I'm a little surprised you propose one here. But I suppose if both variants were to return the same type we should at least infer that type.

Also, once we implement strict Optional checking, `List[None]` no longer would match all overload variants, right?

JukkaL:
Originally the idea was that overloads shouldn't be overlapping, but this got abandoned since it relied on multiple inheritance from built-in types being restricted so that you can't subclass `int` and `str`, for example. If almost any pair of classes can be subclassed, almost all pairs of types are potentially overlapping. Currently mypy doesn't keep track of which classes are built-in / extension classes so it can't do this sort of enforcement easily. Besides, this also seemed to be too restrictive, and the ambiguity due to overlapping overload variants doesn't seem to be a significant problem most of the time. The story is a bit more complex than this, as some forms of overlapping are always okay. Maybe I should write some more detailed docs about how overloaded functions works.

I still don't like union types, but here they just might be justified. However, let's start with an example where union types are plausible but problematic. Consider this fragment:

```
# stub for m
@overload
def f(x: str) -> str: ...
@overload
def f(x: str, y: str) -> int: ...

# program
from m import f
a = <something> # type: Iterable[str]
f(*a)  # Is this valid? What's the type?
```

[...]

PR #4842, opened 2018-04-02
Adds support for basic union math with overloads

Michael0x2a:
This commit adds support for very basic and simple union math when calling overloaded functions, resolving #4576.

One thing led to another, and this ended up accidentally fixing or touching on several different overload-related issues. In particular, I believe this pull request:

1. Fixes the bug (?) where calling overloaded functions can sometimes silently infer a return type of 'Any'
2. Changes the semantics of how mypy handles overlapping functions, which I believe is currently under discussion in python/typing#253

Although this change is functional and mergeable, I was planning on polishing it more -- adding more tests, fleshing out the union math behavior, etc.

However, I think these are sort of big changes and wanted to check in and make sure this pull request is actually welcome/is a good idea. If not, let me know, and I'd be happy to abandon it.

---

Details on specific changes made:

1.  The new algorithm works by modifying `checkexpr.overload_call_targets` to return all possible matches, rather then just one.

    We start by trying the first matching signature. If there was some error, we (conservatively) attempt to union all of the matching signatures together and repeat the typechecking process.

    If it doesn't seem like it's possible to combine the matching signatures in a sound way, we end and just output the errors we obtained from typechecking the first match.

    The "signature-unioning" code is currently deliberately very conservative. I figured it was better to start small and attempt to handle only basic cases like #1943 and relax the restrictions later as needed. For more details on this algorithm, see the comments in `checkexpr.union_overload_matches`.

2.  This change incidentally resolves any bugs related to how calling an overloaded function can sometimes silently infer a return type of Any. Previously, if a function call caused an overload to be
    less precise then a previous one, we gave up and returned a silent Any.

    This change removes this case altogether and only infers Any if either (a) the caller arguments explicitly contains Any or (b) if there was some error.

    For example, see #3295 and #1322 -- I believe this pull request touches on and maybe resolves (??) those two issues.

3.  As a result, this caused a few errors in mypy where code was relying on this "silently infer Any" behavior -- see the changes in checker.py and semanal.py. Both files were using expressions of the form `zip(*iterable)`, which ended up having a type of `Any` under the old algorithm. The new algorithm will instead infer `Iterable[Tuple[Any, ...]]` which actually matches the stubs in typeshed.

4.  Many of the attrs tests were also relying on the same behavior. It seemed that expressions of the form `a = attr.ib()` were evaluated to 'Any' not because of the definitions in the stubs, but because of the 'silent Any' bug.

    I couldn't find a clean way of fixing the stubs to infer the correct thing under this new behavior, so just gave up and removed the overloads altogether. I think this is fine though -- it seems like the attrs plugin infers the correct type for us anyways, regardless of what the stubs say.

    If this pull request is accepted, I plan on submitting a similar pull request to the stubs in typeshed.

4.  This pull request also probably touches on https://github.com/python/typing/issues/253. We still require the overloads to be written from the most narrow to general and disallow overlapping signatures.

    However, if a _call_ now causes overlaps, we try the "union" algorithm described above and default to selecting the first matching overload instead of giving up.


Michael0x2a:
@ilevkivskyi -- yep, will do. I plan on adding more tests either later today or over the weekend -- since this pull request does change the semantics of overloads slightly, I just wanted to make sure it wasn't going to be rejected out-of-hand first :)

No rush on reviewing -- it looks like people are pretty busy so I don't mind waiting.

Michael0x2a:
Superseded by https://github.com/python/mypy/pull/5084.

PR #4843, opened 2018-04-02
Use a FileSystemCache in create_source_list

msullivan:
I also moved the code to a new file. Probably best to look at the commits individually, because of that.

### DIFF
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -5,6 +5,7 @@ import os
 from typing import List, Sequence, Set, Tuple, Optional
 
 from mypy.build import BuildSource, PYTHON_EXTENSIONS
+from mypy.fscache import FileSystemMetaCache
 from mypy.options import Options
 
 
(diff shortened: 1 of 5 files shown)


mypy/find_sources.py
```diff
@@ -15,19 +16,20 @@ class InvalidSourceList(Exception):
     """Exception indicating a problem in the list of sources given to mypy."""
 
 
-# TODO: use a FileSystemCache for this
-def create_source_list(files: Sequence[str], options: Options) -> List[BuildSource]:
+def create_source_list(files: Sequence[str], options: Options,
+                       fscache: Optional[FileSystemMetaCache]=None) -> List[BuildSource]:
```
JukkaL: Style nit: add spaces around `=` in `]=None`.

mypy/find_sources.py
```diff
@@ -15,19 +16,20 @@ class InvalidSourceList(Exception):
...
     """
+    fscache = fscache or FileSystemMetaCache()
     targets = []
     for f in files:
         if f.endswith(PY_EXTENSIONS):
             # Can raise InvalidSourceList if a directory doesn't have a valid module name.
-            targets.append(BuildSource(f, crawl_up(f)[1], None))
-        elif os.path.isdir(f):
-            sub_targets = expand_dir(f)
+            targets.append(BuildSource(f, crawl_up(fscache, f)[1], None))
```
JukkaL: For clarity, change `import os` above to `import os.path` since we no longer directly access any functions in `os` in this module.

JukkaL:
Please have a look at my comments that GitHub considers outdated.

PR #4845, opened 2018-04-03
Fix stubgen --recursive to only generate stubs for requested submodules.

### LINKED ISSUE #4844: `stubgen --recursive` stubs entire package even when only subpackage was requested.
This is a request for a slight change of behavior for `stubgen`.  Right now, if I use `stubgen --recursive urllib.parse`, stubs will be generated for the entirety of `urllib`.

```sh
$ stubgen --recursive urllib.parse
Created out/urllib/__init__.pyi
Created out/urllib/error.pyi
Created out/urllib/parse.pyi
Created out/urllib/request.pyi
Created out/urllib/response.pyi
Created out/urllib/robotparser.pyi
```

This seems to be somewhat odd behavior -- I would have expected stubgen to only generate `out/urllib/parse.pyi`.  To generate all the stubs, I would expect `stubgen --recursive urllib` (currently there is no difference between passing `urllib.parse` and `urllib` as the module to be stubbed).

The behavior is because of the use of `__import__` rather than `importlib.import_module` in [`walk_packages`](https://github.com/python/mypy/blob/master/mypy/stubgen.py#L849).  

I'm not completely sure if this is an intended feature or a bug, but I figured that I'd report it and find out.

mgilson:
This fixes #4844 

I'm not completely sure where/how to add unit tests for this.  I'm happy to do that if someone can point me in the right direction though.

As it stands, I have tested this manually against `urllib` and `urllib.parse` and verified that the current code only generates `out/urllib/parse.pyi` in the latter case but generates all the stubs for `urllib` in the former case.

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -851,7 +851,7 @@ def walk_packages(packages: List[str]) -> Iterator[str]:
         path = getattr(package, '__path__', None)
         if path is None:
             # It's a module inside a package.  There's nothing else to walk/yield.
-            return
+            continue
         for importer, qualified_name, ispkg in pkgutil.walk_packages(path,
                                                                      prefix=package.__name__ + ".",
                                                                      onerror=lambda r: None):
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -846,9 +846,13 @@ def get_qualified_name(o: Expression) -> str:
 
 def walk_packages(packages: List[str]) -> Iterator[str]:
     for package_name in packages:
-        package = __import__(package_name)
+        package = importlib.import_module(package_name)
         yield package.__name__
-        for importer, qualified_name, ispkg in pkgutil.walk_packages(package.__path__,
+        path = getattr(package, '__path__', None)
+        if path is None:
+            # It's a module inside a package.  There's nothing else to walk/yield.
+            return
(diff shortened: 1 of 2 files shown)


mypy/stubgen.py
```diff
@@ -846,9 +846,13 @@ def get_qualified_name(o: Expression) -> str:
...
 def walk_packages(packages: List[str]) -> Iterator[str]:
     for package_name in packages:
-        package = __import__(package_name)
+        package = importlib.import_module(package_name)
         yield package.__name__
-        for importer, qualified_name, ispkg in pkgutil.walk_packages(package.__path__,
+        path = getattr(package, '__path__', None)
+        if path is None:
+            # It's a module inside a package.  There's nothing else to walk/yield.
+            return
```
gvanrossum: Use `continue` here, there may be more packages in the `packages` argument list, we want to process all of them.

gvanrossum:
Unit tests for stubgen are in mypy/test/teststubgen.py, but don't sweat it.

mypy/stubgen.py
mgilson: Yep, thanks.  Somehow I failed to notice the outer loop.

gvanrossum:
This LGTM -- were you going to try and add a test or not? Let me know.

gvanrossum:
This looks like a good start -- thanks for going the extra mile! I'm merging it now.

PR #4848, opened 2018-04-03
Cache package names in create_source_list

msullivan:
On my testing MBP, this shaves about 200ms off of every fine-grained check of S.

### DIFF
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -2,7 +2,7 @@
 
 import os.path
 
-from typing import List, Sequence, Set, Tuple, Optional
+from typing import List, Sequence, Set, Tuple, Optional, Dict
 
 from mypy.build import BuildSource, PYTHON_EXTENSIONS
 from mypy.fscache import FileSystemMetaCache
@@ -10,6 +10,9 @@ from mypy.options import Options
 
 
 PY_EXTENSIONS = tuple(PYTHON_EXTENSIONS)
+
+# A cache for package names, mapping from module id to directory path
+PackageCache = Dict[str, str]
 
 
 class InvalidSourceList(Exception):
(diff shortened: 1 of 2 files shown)


mypy/find_sources.py
```diff
@@ -85,27 +90,51 @@ def expand_dir(fscache: FileSystemMetaCache,
...
+    """Given a directory name, return the corresponding module name.
+
+    Use package_cache to cache results."""
+
+    if dir in package_cache:
+        return package_cache[dir]
+
+    parent_dir, base = os.path.split(dir)
+    if not dir or not get_init_file(fscache, dir) or not base:
+        return ''
```
gvanrossum: So this result should not be cached?

mypy/find_sources.py
```diff
@@ -52,60 +55,87 @@ def keyfunc(name: str) -> Tuple[int, str]:
...
-        if not base:
-            break
-        # Ensure that base is a valid python module name
-        if not base.isidentifier():
-            raise InvalidSourceList('{} is not a valid Python package name'.format(base))
-        if mod == '__init__' or not mod:
-            mod = base
-        else:
-            mod = base + '.' + mod
+        Use package_cache to cache results."""
```
gvanrossum: Move the """ to the next line (by itself). There's no need for a subsequent blank line then.

mypy/find_sources.py
```diff
@@ -52,60 +55,87 @@ def keyfunc(name: str) -> Tuple[int, str]:
...
-                sources.extend(sub_sources)
+class SourceFinder:
+    def __init__(self, fscache: FileSystemMetaCache) -> None:
+        self.fscache = fscache
+        # A cache for package names, mapping from module id to directory path
+        self.package_cache = {}  # type: Dict[str, str]
+
+    def expand_dir(self, arg: str, mod_prefix: str = '') -> List[BuildSource]:
+        """Convert a directory name to a list of sources to build."""
+        f = get_init_file(self.fscache, arg)
```
gvanrossum: Perhaps make get_init_file() also a method here? (Since it uses fscache.)

Issue #4615, opened 2018-02-21
Fine-grained: Inconsistent behavior for nested NewType


JukkaL:
The output of the second increment of this fine-grained incremental mode test case doesn't generate an error, even though nothing relevant has changed:

```
[case testRefreshNestedNewType]
import a
[file a.py]
from typing import Union, NewType
import b

def f(self) -> None:
    X = NewType('X', A)

A = str
[file b.py]
y = 0
[file b.py.2]
y = ''
[out]
```

Here's the output:

```
a.py:5: error: Argument 2 to NewType(...) must be subclassable (got A?)
==
```

The expected output would be identical for both runs.

The output from the first and second runs also does not match for this slightly different test case:

```
[case testRefreshNestedNewType]
import a
[file a.py]
from typing import Union, NewType
import b

def f(self) -> None:
    X = NewType('X', A)

A = Union[int, str]

[file b.py]
y = 0
[file b.py.2]
y = ''
[out]
```

Here's the output (again expected to be identical):

```
a.py:5: error: Argument 2 to NewType(...) must be subclassable (got A?)
==
a.py:5: error: Argument 2 to NewType(...) must be subclassable (got "Union[int, str]")
```

It looks like forward references within NewType declarations behave inconsistently.

ilevkivskyi:
I think that at least the second one (with union) should be fixed by my alias PR #4525. My guess is that things that should be stripped are not stripped (like `analyzed`).

ilevkivskyi:
Also there is in fact no big difference between two examples from point of view of forward refs vs `NewType`, in the first case the analyzed forward ref is `str` which is a valid base class, therefore no error is shown.

ilevkivskyi:
It looks like `strip_target` is not called on `a.py`, while it should. Therefore, the type aliases are not stripped, and appear in analyzed form on the second run. I am not yet sure how to fix this.

ilevkivskyi:
After a discussion with @JukkaL it looks like the right solution for now would be to support forward references in `NewType`s. We anyway allow them in many places where they should be prohibited, so this is consistent.

PR #4855, opened 2018-04-04
Refactor: Move code out of semantic analyzer pass 2

JukkaL:
This moves the semantic analysis of named tuples, TypedDicts, call-based 
Enum and NewType to separate modules. I chose these based on these 
criteria:

* The features have a significant amount of code so refactoring seems
  valuable.
* The features don't interact much with other features.
* The features are arguably not "core" mypy features.

There are still other bits we could move away in the future.

Also extended the abstract semantic analyzer API to support these
features.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -84,8 +84,9 @@ from mypy import experiments
 from mypy.plugin import Plugin, ClassDefContext, SemanticAnalyzerPluginInterface
 from mypy import join
 from mypy.util import get_prefix, correct_relative_import
-from mypy.semanal_shared import SemanticAnalyzerInterface, PRIORITY_FALLBACKS
+from mypy.semanal_shared import SemanticAnalyzerInterface, set_callable_name, PRIORITY_FALLBACKS
 from mypy.scope import Scope
+from mypy.semanal_namedtuple import NamedTupleAnalyzer, NAMEDTUPLE_PROHIBITED_NAMES
 
 
 T = TypeVar('T')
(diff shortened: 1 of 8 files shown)


mypy/semanal.py
```diff
@@ -277,6 +273,10 @@ def visit_file(self, file_node: MypyFile, fnam: str, options: Options,
         self.is_typeshed_stub_file = self.errors.is_typeshed_file(file_node.path)
         self.globals = file_node.names
         self.patches = patches
+        self.named_tuple_analyzer = NamedTupleAnalyzer(options, self)
```
ilevkivskyi: There is a bit strange asymmetry between `NamedTupleAnanlyzer` and `TypedDictAnalyzer`, because only the latter needs `self.msg`. It is needed by `check_for_explicit_any`, and it is a bit surprising that it doesn't appear in `NamedTupleAnalyzer`. I think it would be good or explain this asymmetry.

mypy/semanal.py
```diff
@@ -1794,11 +1603,11 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
             for lvalue in s.lvalues:
                 self.store_declared_types(lvalue, s.type)
         self.check_and_set_up_type_alias(s)
-        self.process_newtype_declaration(s)
+        self.newtype_analyzer.process_newtype_declaration(s)
```
ilevkivskyi: Rather a comment for myself: `NewType` and `TypeVar` not requiring `self.is_func_scope()` are by interesting coincidence(?) behave in weird way with forward references in fine grained, see e.g. https://github.com/python/mypy/issues/4615

PR #4861, opened 2018-04-05
Support module aliases in fine grained mode

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/4532

This also fixes two bugs discovered when a module is replaced with a function, or a module alias is replaced with a type alias.

The idea is trivial, we just record alternative names for a given module ("a.k.a" names) and then add dependencies on them as well, in addition to the original module name ("real" name). The non-trivial part is `aststrip.py` part, because module aliasing does some irreversible changes during second pass, so I record the "unanalized" (i.e. after first pass) node and kind, and then restore them in `aststrip.py` (I cannot just set them to `None`, because they are not `None` after first pass and some code relies on this).

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -211,6 +211,8 @@ class MypyFile(SymbolNode):
     is_stub = False
     # Is this loaded from the cache and thus missing the actual body of the file?
     is_cache_skeleton = False
+    # Full names under which this module appears in other modules.
+    aka_names = None  # type: Set[str]
 
     def __init__(self,
                  defs: List[Statement],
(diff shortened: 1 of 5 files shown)


ilevkivskyi:
> Hmm this looks like a bit too much extra state for such a minor edge case.

However only few `SymbolTableNodes` instances will have this state, so it is rather just a code complexity problem.
> I wonder if there is a simpler way to implement this. Maybe we should first refactor symbol table node kinds?

The dependency part is already very simple. However the node aliasing indeed can be made simpler after refactoring. I will check if there is some easy makeshift way to do this, if not then let's wait until refactoring.

JukkaL:
> However only few SymbolTableNodes instances will have this state, so it is rather just a code complexity problem.

Yeah, code complexity is what I'm worried about.

ilevkivskyi:
This has some merge conflicts and it is unlikely we will go this way, so I am just closing this.

PR #4878, opened 2018-04-09
Correct search order for installed packages

### LINKED ISSUE #4876: Installed stubs not always taking precedence in PEP 561 implementation
I installed a package with inline type annotations using `python setup.py install` and installed the corresponding stubs package using `python setup.py install --user`. Mypy picked up the inline type annotations instead of the stubs, even though I'd expect that a stub package takes precedence. I suspect that this is caused by the package and the stubs being in different directories, and that the order in which we search through the directories may affect which types will be used by mypy.

@ethanhs Can you take a look at this?

ethanhs:
Due to the method of looking for installed packages, it was possible for
the search to be unstable: an installed inline package would be chosen
before an installed stub package. This corrects that by storing a
seperate list for each, and putting the stub packages first.

Fixes #4876

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -858,7 +858,10 @@ class FindModuleCache:
         dir_chain = os.sep.join(components[:-1])  # e.g., 'foo/bar'
         # TODO (ethanhs): refactor each path search to its own method with lru_cache
 
-        third_party_dirs = []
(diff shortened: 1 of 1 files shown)


gvanrossum:
I'll merge this now, before cutting the release branch, since it looks right.

PR #4888, opened 2018-04-11
fix test_typed_pkg test failure on Mac

### LINKED ISSUE #4883: Test failure: test_typed_pkg
I'm running mypy tests from master ( python/mypy@af7e834 ) with python 3.6.2; When running tests I get a failure at `test_typed_package`

```
====================================================== FAILURES =======================================================
______________________________________________ TestPEP561.test_typed_pkg ______________________________________________
[gw1] linux -- Python 3.6.2 /home/machinalis/.virtualenvs/mypy-dev/bin/python3

self = <mypy.test.testpep561.TestPEP561 testMethod=test_typed_pkg>

    def test_typed_pkg(self) -> None:
        """Tests type checking based on installed packages.
    
            This test CANNOT be split up, concurrency means that simultaneously
            installing/uninstalling will break tests.
            """
        test_file = 'simple.py'
        if not os.path.isdir('test-packages-data'):
            os.mkdir('test-packages-data')
        old_cwd = os.getcwd()
        os.chdir('test-packages-data')
        with open(test_file, 'w') as f:
            f.write(SIMPLE_PROGRAM)
        try:
            with self.install_package('typedpkg-stubs'):
                check_mypy_run(
                    [test_file],
                    "simple.py:4: error: Revealed type is 'builtins.list[builtins.str]'\n"
                )
    
            # The Python 2 tests are intentionally placed after a Python 3 test to check
            # the package_dir_cache is behaving correctly.
            python2 = try_find_python2_interpreter()
            if python2:
                with self.install_package('typedpkg-stubs', python2):
                    check_mypy_run(
                        ['--python-executable={}'.format(python2), test_file],
                        "simple.py:4: error: Revealed type is 'builtins.list[builtins.str]'\n"
                    )
                with self.install_package('typedpkg', python2):
                    check_mypy_run(
                        ['--python-executable={}'.format(python2), 'simple.py'],
                        "simple.py:4: error: Revealed type is 'builtins.tuple[builtins.str]'\n"
                    )
    
                with self.install_package('typedpkg', python2):
                    with self.install_package('typedpkg-stubs', python2):
                        check_mypy_run(
                            ['--python-executable={}'.format(python2), test_file],
                            "simple.py:4: error: Revealed type is 'builtins.list[builtins.str]'\n"
                        )
    
            with self.install_package('typedpkg'):
                check_mypy_run(
                    [test_file],
>                   "simple.py:4: error: Revealed type is 'builtins.tuple[builtins.str]'\n"
                )

mypy/test/testpep561.py:103: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cmd_line = ['simple.py'], expected_out = "simple.py:4: error: Revealed type is 'builtins.tuple[builtins.str]'\n"
expected_err = '', expected_returncode = 1

    def check_mypy_run(cmd_line: List[str],
                       expected_out: str,
                       expected_err: str = '',
                       expected_returncode: int = 1) -> None:
        """Helper to run mypy and check the output."""
        out, err, returncode = mypy.api.run(cmd_line)
>       assert out == expected_out, err
E       AssertionError: 
E       assert "simple.py:4:...ltins.str]'\n" == "simple.py:4: ...ltins.str]'\n"
E         Skipping 37 identical leading characters in diff, use -v to show
E         - 'builtins.list[builtins.str]'
E         ?            ^^^
E         + 'builtins.tuple[builtins.str]'
E         ?           +++ ^

mypy/test/testpep561.py:27: AssertionError
============================================== 1 failed in 26.85 seconds ==============================================
```

JelleZijlstra:
Fixes #4883 

The check for whether we're in a venv turns out to be unreliable. See https://stackoverflow.com/questions/1871549/determine-if-python-is-running-inside-virtualenv for background.

Note that if you're testing this on a Mac that previously ran into the bug, you'll have to manually remove any global typedpkg installs for the test to work.

### DIFF
--- a/mypy/test/testpep561.py
+++ b/mypy/test/testpep561.py
@@ -29,6 +29,18 @@ def check_mypy_run(cmd_line: List[str],
     assert returncode == expected_returncode, returncode
 
 
+def is_in_venv() -> bool:
+    """Returns whether we are running inside a venv.
+
+    Based on https://stackoverflow.com/a/42580137.
+
+    """
+    if hasattr(sys, 'real_prefix'):
+        return True
+    else:
+        return hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix
+
+
 class TestPEP561(TestCase):
     @contextmanager
     def install_package(self, pkg: str,
@@ -38,9 +50,7 @@ class TestPEP561(TestCase):
         install_cmd = [python_executable, '-m', 'pip', 'install', '.']
         # if we aren't in a virtualenv, install in the
         # user package directory so we don't need sudo
-        # In a virtualenv, real_prefix is patched onto
-        # sys
-        if not hasattr(sys, 'real_prefix') or python_executable != sys.executable:
+        if not is_in_venv() or python_executable != sys.executable:
             install_cmd.append('--user')
         returncode, lines = run_command(install_cmd, cwd=working_dir)
         if returncode != 0:
--- a/mypy/test/testpep561.py
+++ b/mypy/test/testpep561.py
@@ -38,7 +38,8 @@ def is_in_venv() -> bool:
     if hasattr(sys, 'real_prefix'):
         return True
     else:
-        return hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix
+        # https://github.com/python/typeshed/pull/2047
+        return hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix  # type: ignore
 
 
 class TestPEP561(TestCase):


mypy/test/testpep561.py
```diff
@@ -29,6 +29,19 @@ def check_mypy_run(cmd_line: List[str],
...
+def is_in_venv() -> bool:
+    """Returns whether we are running inside a venv.
+
+    Based on https://stackoverflow.com/a/42580137.
+
+    """
+    if hasattr(sys, 'real_prefix'):
+        return True
+    else:
+        # https://github.com/python/typeshed/pull/2047
```
gvanrossum: I merged that typeshed PR, but yet another typeshed sync is also needed.

JelleZijlstra: I was expecting that we'd just merge this to get mypy master back in a good state. I can send another small PR to remove the type ignore later.

gvanrossum: (Actually I propose not to cp that onto the release branch, since the `# type: ignore` below silences the error. We can fix things on the master branch.)

gvanrossum:
So apparently the repro was to run the test *twice*...

I'm a little sad this test uses and modifies the environment.

ethanhs:
>So apparently the repro was to run the test *twice*...

Well, run the test twice in an environment where the global/user site packages directory was available from a virtualenv. In all of my testing environments, the tests could be run twice without failure.

>I'm a little sad this test uses and modifies the environment.

Yes, it is less than ideal. I am working on a version that creates, uses, then destroys a virtualenv of its own.

JelleZijlstra:
>I am working on a version that creates, uses, then destroys a virtualenv of its own.

That might also be tricky, because on Python 2 you need to install the third-party `virtualenv` library to even create a venv, so your test might still need to do things to the global env in order to even make a venv.

ethanhs:
>That might also be tricky, because on Python 2 you need to install the third-party virtualenv library to even create a venv, so your test might still need to do things to the global env in order to even make a venv.

I mean our Python 2 tests already depend on it having `typing` installed. We can add `virtualenv` as a requirement on top of that.

gvanrossum:
I cherry-picked this into release-0.590, but not #4884 (which you created to debug this). And I'm not going to cp the typeshed change for `sys.base*prefix`, nor hurry with another typeshed sync.

PR #4894, opened 2018-04-12
Rework (and heavily optimize!) mypy.ini per-module configuration

msullivan:
Currently, we compute the options for each module by scanning through
the full list of per-module configuration options and seeing if the
pattern matches against the module name.
This is incredibly slow for large configuration files. On the internal S repo,
mypy spends 23 seconds (!!) just computing per-module options.

To fix this, we instead precompute an Options object for each config
section, and in `clone_for_module` do a search for the most specific
configured Options (so for foo.bar, we try foo.bar, foo.bar.*,
foo.*, in that order).

This cuts down the processing time from 23s to about 80ms.

The catch is that this is actually a backwards-incompatible semantics
change, in two ways:
 * Patterns can be of the form `foo.bar` and `foo.bar.*`, but can no longer be
   general purpose globs. We produce an error message to detect this misusage.
 * Patterns are now always applied based on specificity and not on the order they
   appear in the file. This means that some poorly-formed configuration files that
   contained options that were always overridden may have their meaning changed.
   This is probably not too common, so we don't do anything to deal with it.
   We could emit a warning when specificity-order and file-order disagree, but
   it seems wrong to lock people into the old behavior, and it seems like a silly
   thing to make configurable.

I had a really pretty trie-based solution that I wrote before realizing I could do it this way and it would probably be good enough. The trie-based solution is faster, but not enough faster to justify the extra complication :(

### DIFF
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -29,10 +29,15 @@ characters.
   the global flags. The ``setup.cfg`` file is an exception to this.
 
 - Additional sections named ``[mypy-PATTERN1,PATTERN2,...]`` may be
-  present, where ``PATTERN1``, ``PATTERN2`` etc. are `fnmatch patterns
-  <https://docs.python.org/3.6/library/fnmatch.html>`_
-  separated by commas.  These sections specify additional flags that
-  only apply to *modules* whose name matches at least one of the patterns.
+  present, where ``PATTERN1``, ``PATTERN2`` etc. are comma-separated
+  patterns of the form ``dotted_module_name`` or ``dotted_module_name.*``.
+  These sections specify additional flags that only apply to *modules*
+  whose name matches at least one of the patterns.
+
+  A pattern of the form ``dotted_module_name`` matches only the named module,
+  while ``dotted_module_name.*`` matches ``dotted_module_name`` and any
+  submodules (so ``foo.bar.*`` would match all of ``foo.bar``,
+  ``foo.bar.baz``, and ``foo.bar.baz.quux``).
 
 .. note::
 
@@ -137,8 +142,12 @@ overridden by the pattern sections matching the module name.
 
 .. note::
 
-   If multiple pattern sections match a module they are processed in
-   order of their occurrence in the config file.
+   If multiple pattern sections match a module, the options from the
+   most specific section are used where they disagree.  This means
+   that ``foo.bar`` will take values from sections with the patterns
+   ``foo.bar``, ``foo.bar.*``, and ``foo.*``, but when they specify
+   different values, it will use values from ``foo.bar`` before
+   ``foo.bar.*`` before ``foo.*``.
 
 - ``follow_imports`` (string, default ``normal``) directs what to do
   with imports when the imported module is found as a ``.py`` file and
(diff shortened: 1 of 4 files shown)


docs/source/config_file.rst
```diff
@@ -29,10 +29,15 @@ characters.
   the global flags. The ``setup.cfg`` file is an exception to this.
 
 - Additional sections named ``[mypy-PATTERN1,PATTERN2,...]`` may be
-  present, where ``PATTERN1``, ``PATTERN2`` etc. are `fnmatch patterns
-  <https://docs.python.org/3.6/library/fnmatch.html>`_
-  separated by commas.  These sections specify additional flags that
-  only apply to *modules* whose name matches at least one of the patterns.
+  present, where ``PATTERN1``, ``PATTERN2`` etc. are comma-separated
```
JukkaL: Grammar nit: You are missing some commas around 'etc.'. Maybe: `PATTERN2`, etc., are ....

test-data/unit/cmdline.test
```diff
@@ -202,18 +203,15 @@ def g(a: int) -> int: return f(a)
...
 [out]
-yy.py:2: error: Call to untyped function "f" in typed context
-yx.py:1: error: Function is missing a type annotation
-yx.py:2: error: Call to untyped function "f" in typed context
-xy.py:1: error: Function is missing a type annotation
-xy.py:2: error: Call to untyped function "f" in typed context
-xx.py:1: error: Function is missing a type annotation
+mypy.ini: [mypy-*x*]: Invalid pattern. Patterns must be 'module_name' or 'module_name.*'
+mypy.ini: [mypy-*y*]: Invalid pattern. Patterns must be 'module_name' or 'module_name.*'
+== Return code: 0
```
JukkaL: Should the return code be non-zero?

test-data/unit/cmdline.test
```diff
@@ -1113,3 +1125,25 @@ follow_imports_for_stubs = True
...
+[[mypy-spam.eggs]
+[[mypy-emarg.*]
+[[mypy-emarg.hatch]
+[file foo.py]
+[file quux.py]
+[file spam/__init__.py]
+[file spam/eggs.py]
+[out]
+Warning: unused section(s) in mypy.ini: [mypy-bar], [mypy-baz.*], [mypy-emarg.*], [mypy-emarg.hatch]
+== Return code: 0
```
JukkaL: Again the 0 exit status is unexpected (but this was like this before this PR).

test-data/unit/cmdline.test
msullivan: Yes, but we don't currently error for most `mypy.ini` so I've stayed consistent for now.

msullivan:
Typeshed update was a misfire, will fix

felixc:
Hi; just to clarify the change that's needed to existing config files: Does this mean that if you have a directory (say, "repositories"), which contains a bunch of subdirs, (call them "a" through "e"), all of which contain a "migrations" directory that should be ignored, we need to go from:

```ini
[mypy-*migrations*]
ignore_errors = True
```

to

```ini
[mypy-repositories.a.migrations.*,repositories.b.migrations.*,repositories.c.migrations.*,repositories.d.migrations.*,repositories.e.migrations.*]
ignore_errors = True
```

Or is there a more concise option, or one that would be robust to the hypothetical future addition of "f/migrations", "g/migrations", etc? Thanks!

msullivan:
Yeah, that is correct.

PR #4898, opened 2018-04-13
Restore dict based typesafe caching

ethanhs:
This also restores `isfile_case` caching and fixes a type error (and legitimate bug) (!).

The last remaining follow up from #4693 is to run only one Python subprocess.

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -160,7 +160,8 @@ def find_module_path_and_all(module: str, pyversion: Tuple[int, int],
             module_all = getattr(mod, '__all__', None)
     else:
         # Find module by going through search path.
-        module_path = mypy.build.FindModuleCache().find_module(module, ['.'] + search_path,
+        module_path = mypy.build.FindModuleCache().find_module(module,
+                                                               ('.',) + tuple(search_path),
                                                                interpreter)
         if not module_path:
             raise SystemExit(
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -825,18 +825,26 @@ class FindModuleCache:
 
     def __init__(self, fscache: Optional[FileSystemMetaCache] = None) -> None:
         self.fscache = fscache or FileSystemMetaCache()
-        self.find_lib_path_dirs = functools.lru_cache(maxsize=None)(self._find_lib_path_dirs)
-        self.find_module = functools.lru_cache(maxsize=None)(self._find_module)
+        # Cache find_lib_path_dirs: (dir_chain, lib_path)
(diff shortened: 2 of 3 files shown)


gvanrossum:
Does this need to go into release-0.590? I.e. is it a critical bugfix to the PEP 561 code or just a performance improvement? (I haven't looked yet -- was there a perf regression in situations that don't invoke PEP 561?)

mypy/stubgen.py
```diff
@@ -160,7 +160,8 @@ def find_module_path_and_all(module: str, pyversion: Tuple[int, int],
             module_all = getattr(mod, '__all__', None)
     else:
         # Find module by going through search path.
-        module_path = mypy.build.FindModuleCache().find_module(module, ['.'] + search_path,
+        module_path = mypy.build.FindModuleCache().find_module(module,
+                                                               ('.',) + tuple(search_path),
```
ethanhs: @gvanrossum this is a bug on master. `lib_path` should be a tuple, and as you can see to the left, it is passed as a list. I can split this fix out if you don't want the rest.

JelleZijlstra: What is the concrete effect here? Could this cause problems to users?

ethanhs: Essentially, `lru_cache` won't be able to hash the passed `list`.

```
$ python -m mypy.stubgen --no-import mypy
  File "C:\Python36\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Python36\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\ethanhs\Documents\mypy\mypy\stubgen.py", line 1022, in <module>
    main()
  File "C:\Users\ethanhs\Documents\mypy\mypy\stubgen.py", line 901, in main
    raise e
  File "C:\Users\ethanhs\Documents\mypy\mypy\stubgen.py", line 898, in main
    include_private=options.include_private)
  File "C:\Users\ethanhs\Documents\mypy\mypy\stubgen.py", line 106, in generate_stub_for_module
    interpreter=interpreter)
  File "C:\Users\ethanhs\Documents\mypy\mypy\stubgen.py", line 164, in find_module_path_and_all
    interpreter)
TypeError: unhashable type: 'list'
```

JelleZijlstra: Then maybe make a separate PR for that change?

gvanrossum: OK, please split that out -- I can cherry-pick a 3-line fix to stubgen with a lot less risk and effort than this whole PR.

ethanhs: @gvanrossum https://github.com/python/mypy/pull/4899

PR #4902, opened 2018-04-13
Optimize file system cache memory usage 

JukkaL:
Copy exception objects to get rid of tracebacks and frame objects that
are not needed and consume memory.

Verified memory savings using the recently added memory profiler.

Alternative to #4809 (which didn't seem to reclaim the space in my
experiments).

### DIFF
--- a/mypy/fscache.py
+++ b/mypy/fscache.py
@@ -31,7 +31,7 @@ advantage of the benefits.
 import functools
 import os
 import stat
-from typing import Dict, List, Optional, Tuple, TypeVar
+from typing import Dict, List, Optional, Tuple, TypeVar, Type
 from mypy.util import read_with_python_encoding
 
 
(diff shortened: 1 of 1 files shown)


JukkaL:
This saves about 4% of total memory use when type checking mypy.

mypy/fscache.py
```diff
@@ -42,33 +42,36 @@ def __init__(self) -> None:
...
+            raise err
         self.stat_cache[path] = st
         return st
 
     def listdir(self, path: str) -> List[str]:
         if path in self.listdir_cache:
             return self.listdir_cache[path]
         if path in self.listdir_error_cache:
-            raise self.listdir_error_cache[path]
+            raise copy_os_error(self.listdir_error_cache[path])
```
eric-wieser: Why a new copy here?

PR #4904, opened 2018-04-13
Reduce memory usage by using __slots__

JukkaL:
In some cases this reduces the memory usage of mypy and mypy-daemon
over 30%. This seems to have only a minor effect on speed. Depending on
the workload, this seems to either speed up or slow down runtimes by a
few percent.

Notes:
- Refactor various attribute initializations away from class bodies, as they 
  conflict with `__slots__`.
- This only adds `__slots__` to classes that use the most memory, based
  on the output from the memory profiler.
- Some attributes had `None` defaults even though the type was not
  optional. Now we use a cast for some of these, which has the same 
  effect.
- Various `__dict__` manipulations had to be modified to support 
  `__slots__` as well.


JukkaL:
I think that I've addressed all comments (or at least most).

PR #4906, opened 2018-04-13
Split fine_grained_deps out into its own .deps.json cache file

msullivan:
The goal here is to decouple loading of of the AST trees from the data
cache and the fine-grained dependencies. This means that a regular
incremental run can use cache files produced with --cache-fine-grained
without needing to load the fine-grained dependencies.
(To enable this, we drop "cache_fine_grained" from the OPTIONS_AFFECTING_CACHE
that are checked when validating metadata. If we require a fine-grained cache
but don't have one, that will be caught by the lack of deps_mtime.)

More importantly, it will enable loading dependencies without needing
to parse the data json, which is a good optimization for for a planned
patch to load data caches on demand in fine-grained mode.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1110,6 +1110,7 @@ def validate_meta(meta: Optional[CacheMeta], id: str, path: Optional[str],
         return None
     deps_mtime = None
     if manager.options.cache_fine_grained:
+        assert meta.deps_json
         deps_mtime = getmtime(meta.deps_json)
         if deps_mtime != meta.deps_mtime:
             manager.log('Metadata abandoned for {}: deps cache is modified'.format(id))
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -396,7 +396,9 @@ CacheMeta = NamedTuple('CacheMeta',
                         ('hash', str),
                         ('dependencies', List[str]),  # names of imported modules
                         ('data_mtime', int),  # mtime of data_json
+                        ('deps_mtime', Optional[int]),  # mtime of deps_json
                         ('data_json', str),  # path of <id>.data.json
+                        ('deps_json', Optional[str]),  # path of <id>.deps.json
                         ('suppressed', List[str]),  # dependencies that weren't imported
                         ('child_modules', List[str]),  # all submodules of the given module
                         ('options', Optional[Dict[str, object]]),  # build options
@@ -413,7 +415,8 @@ CacheMeta = NamedTuple('CacheMeta',
 # silent mode or simply not found.
 
 
-def cache_meta_from_dict(meta: Dict[str, Any], data_json: str) -> CacheMeta:
+def cache_meta_from_dict(meta: Dict[str, Any],
+                         data_json: str, deps_json: Optional[str]) -> CacheMeta:
     sentinel = None  # type: Any  # Values to be validated by the caller
     return CacheMeta(
         meta.get('id', sentinel),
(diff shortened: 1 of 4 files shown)


msullivan:
Failures look like a dict iteration order nondeterminism bug

msullivan:
Oh no it was so much dumber than that. `deps_mtime = getmtime(meta.data_json)` *almost* always works...

mypy/build.py
```diff
@@ -396,7 +396,9 @@ def default_lib_path(data_dir: str,
                         ('hash', str),
                         ('dependencies', List[str]),  # names of imported modules
                         ('data_mtime', int),  # mtime of data_json
+                        ('deps_mtime', Optional[int]),  # mtime of deps_json
                         ('data_json', str),  # path of <id>.data.json
+                        ('deps_json', Optional[str]),  # path of <id>.deps.json
```
JukkaL: Also add comment with hint about what the `...deps.json` files is used for (fine-grained deps?).

mypy/build.py
```diff
@@ -413,7 +415,8 @@ def default_lib_path(data_dir: str,
 # silent mode or simply not found.
 
 
-def cache_meta_from_dict(meta: Dict[str, Any], data_json: str) -> CacheMeta:
+def cache_meta_from_dict(meta: Dict[str, Any],
+                         data_json: str, deps_json: Optional[str]) -> CacheMeta:
```
JukkaL: Can you add a docstring and describe arguments, including `deps_json`?

mypy/build.py
```diff
@@ -972,16 +977,20 @@ def get_cache_names(id: str, path: str, manager: BuildManager) -> Tuple[str, str
       pyversion: Python version (major, minor)
 
     Returns:
-      A tuple with the file names to be used for the meta JSON and the
-      data JSON, respectively.
+      A tuple with the file names to be used for the meta JSON, the
+      data JSON, and the deps JSON, respectively.
```
JukkaL: Maybe call the deps JSON "fine-grained deps JSON" to make it clear that this is different from module dependencies?

mypy/build.py
```diff
@@ -1098,6 +1108,13 @@ def validate_meta(meta: Optional[CacheMeta], id: str, path: Optional[str],
     if data_mtime != meta.data_mtime:
         manager.log('Metadata abandoned for {}: data cache is modified'.format(id))
         return None
+    deps_mtime = None
+    if manager.options.cache_fine_grained:
+        assert meta.deps_json
+        deps_mtime = getmtime(meta.deps_json)
```
JukkaL: Use the file system cache?

mypy/build.py
```diff
@@ -1265,6 +1286,15 @@ def write_cache(id: str, path: str, tree: MypyFile,
             return interface_hash, None
         data_mtime = getmtime(data_json)
 
+    deps_mtime = None
+    if deps_json:
+        deps_str = json_dumps(serialized_fine_grained_deps, manager.options.debug_cache)
+        if not atomic_write(deps_json, deps_str, '\n'):
+            manager.log("Error writing deps JSON file {}".format(deps_json))
+            return interface_hash, None
+            raise CompileError(["Error writing deps JSON file {}".format(deps_json)])
```
JukkaL: This raises after return -- looks like the raise statement is redundant.

test-data/unit/check-incremental.test
```diff
@@ -4235,3 +4235,34 @@ pass
...
+import a
+[file a.py]
+x = 0
+[file mypy.ini]
+[[mypy]
+cache_fine_grained = False
+[file mypy.ini.2]
+[[mypy]
+cache_fine_grained = True
+-- Nothing should get rechecked
```
JukkaL: The comment seems to conflict with what's below.

mypy/build.py
msullivan: We don't use the cache for meta files at all, currently

Issue #1907, opened 2016-07-19
Wrong checking of overload for keyword-only args


dmoisset:
While writing some stubs I found an issue that I simplified to the following:

``` python
from typing import overload

@overload
def f(*, out: str, flag: bool=False) -> str: ...

@overload
def f(*, flag: bool=False) -> bool: ...

reveal_type(f())  # I expect bool, get Any
reveal_type(f(flag=False))  # I expect bool, get Any
reveal_type(f(out="xxx"))   # I expect str, get str
```

The checker shows a

```
error: Overloaded function signatures 1 and 2 overlap with incompatible return types
```

but actually the signatures do not overlap (there's no call that can match both the signatures).

Something that's also confusing is that if I change the order of the arguments like this

``` python
# Move the "out" argument to the end
@overload
def f(*, flag: bool=False, out: str) -> str: ...
@overload
def f(*, flag: bool=False) -> bool: ...
```

I stop getting the "signature overlap" error, but I still get the "Any" type on the first two example calls. Given that these arguments are keyword only their positions shouldn't matter at all

(Comes from python/typing#248 )

euresti:
Very weirdly it's the kwonly that's causing issues:

This code has no issues:
```
@overload
def f(x: int) -> int: ...
@overload
def f(x: str) -> str: ...
```

But make those kwonly and boom:
```
@overload
def f(*, x: int) -> int: ...   # error: Overloaded function signatures 1 and 2 overlap with incompatible return types
@overload
def f(*, x: str) -> str: ...
```

ilevkivskyi:
The `Any` return is now fixed, but the overlap error still appears, so this is a false positive.

Michael0x2a:
This took way way longer to resolve then I originally thought it would take, but the above examples are working as expected now in master after https://github.com/python/mypy/commit/f61c2baac5915ab680c2f1a9858b7f608262a00e.

There's still a few todos related to handling edge cases w.r.t. star args, optional args, and such in https://github.com/python/mypy/pull/5163 that needs to land before we can really call this done, but I'm going to close this mostly so I can prune the list of overload-related things I have left to do.

Issue #4159, opened 2017-10-25
allow overloads to be distinguished by contained type of a generic


carljm:
This is a follow-up to the discussion beginning at https://github.com/python/typing/issues/253#issuecomment-235652393

Mypy's choice to not distinguish overload signatures by contained types rules out some natural API choices, e.g.

```
@overload
def map(ids: List[int]) -> Dict[int, MyObj]: ...

@overload
def map(ids: List[str]) -> Dict[str, MyObj]: ...
```
 
We've run into this limitation several times already at Instagram, and there aren't any nice workarounds for it. From a user point of view, the "empty list" rationale doesn't feel right; it seems like it should be possible to only error in specific calls with an unknown contained type.

There is some more discussion in the above-linked thread of how to handle this. In any case, I think it deserves tracking as a mypy limitation.

JelleZijlstra:
This particular case could be handled by a `TypeVar('T', int, str)`, unless I'm missing something. Is your actual use case more complicated?

carljm:
Yes, good point, I over-simplified the example. In our real cases the types involved are not always the same (there is a third overload for a different type of id, which still results in int keys in the returned mapping).

JukkaL:
The old worry about empty lists is no longer valid. In the early days of mypy it was a bigger concern since `@overload` actually did runtime type dispatch, but now that it's only used for static checking we can relax the rules.

ilevkivskyi:
> but now that it's only used for static checking we can relax the rules.

I think that it is not well specified in PEP 484 what overloads mean. I think the first step here should be to make a PR to PEP 484 with a detailed specification of how overloads work, what is allowed, and what is not.

JukkaL:
> I think that it is not well specified in PEP 484 what overloads mean. I think the first step here should be to make a PR to PEP 484 with a detailed specification of how overloads work, what is allowed, and what is not.

I agree that PEP 484 should specify how overloads should work, but I think that it's better to write the specification together with an implementation so that we can immediately test how the specification works in practice.

ilevkivskyi:
This is fixed on master now.

PR #4912, opened 2018-04-14
Make overloads respect keyword-only args

Michael0x2a:
This pull request resolves https://github.com/python/mypy/issues/1907.

Specifically, it turned out that we never support for non-positional args in overload was never implemented to begin with. Thankfully, it also turned out the bulk of the logic we wanted was already implemented within `mypy.subtypes.is_callable_subtype`. Rather then re-implementing that code, this pull request refactors that method to support any kind of check, instead of specifically subtype checks.

This, as a side-effect, ended up making some partial progress towards https://github.com/python/mypy/issues/4159 -- this is because unlike the existing checks, `mypy.subtypes.is_callable_subtype` *doesn't* erase types and has better support for typevars in general.

The reason this commit does not fully remove type erasure from overload checks is because the new implementation still calls `mypy.meet.is_overlapping_types` which *does* perform erasure. But fixing that seemed out-of-scope for this commit, so I stopped here.


test-data/unit/check-overloading.test
```diff
@@ -1026,16 +1033,21 @@ g(1, 'foo')
...
-[case testBadOverlapWithTypeVarsWithValues]
+[case testOverlapWithTypeVarsWithValuesOrdering]
 from foo import *
 [file foo.pyi]
 from typing import overload, TypeVar
 AnyStr = TypeVar('AnyStr', bytes, str)
 
 @overload
-def f(x: AnyStr) -> None: pass # E: Overloaded function signatures 1 and 2 overlap with incompatible return types
+def f(x: AnyStr) -> None: pass
```
ilevkivskyi: This example is not very useful, I would rather check what happens for `(AnyStr) -> AnyStr`.

Michael0x2a:
@ilevkivskyi -- I updated the test, fyi.

> A general comment, I would rather recommend spending some time in writing a strict specification of how overloads should work, otherwise these PRs feel a bit unsystematic.

I think my strategy was to stick to making changes that improve the status quo, avoid making drastic changes, and steadily iterate before trying to nail down a specification. (E.g. regardless of how we handle overloads, it'd be nice if they understood keyword-only arguments; partially removing the erasure thing was just a nice side-effect that also improves the status quo).

That said, I don't mind trying to write a specification and pushing for more systemic changes if you + the mypy team think it's a good idea -- I just wasn't sure if anybody had the bandwidth to discuss/review any proposed changes.

But if you think I should go for it anyways, let me know and I can give it a shot :)

ilevkivskyi:
Specifying how exactly overloads should work is a long standing issue, I was going to work on this myself anyway, also I think other will agree it is an important thing to fix before PEP 484 is no more provisional, see https://github.com/python/typing/issues/253

Concerning the bandwidth, discussing the specification will take less energy (at least for me) than reviewing code without specification. Also I think @JukkaL mentioned some time ago that specification and implementation should be developed in parallel, so that we don't specify something that is hard to implement. In this view you can use your PRs to test various aspects of the future proposed spec.

PR #4917, opened 2018-04-16
Support types as converters in attrs plugin

### LINKED ISSUE #4729: bug with attrs converters
@euresti I'm getting new errors when running mypy against the attrs doctests:

```
Line 1, in converters1
Failed example:
    import attr
    @attr.s
    class C(object):
        x = attr.ib(converter=int)
    o = C("1")
    o.x
Expected nothing
Got:
    main:4: error: Unsupported converter function, only named functions are currently supported
```

I assume this is trying to ward off lambdas, but getting tripped up by builtins?

Btw, Is there an existing issue for creating a mypy plugin search path, so that we can move the attrs plugin into the attrs project?

euresti:
Fixes #4729

### DIFF
--- a/test-data/unit/check-attr.test
+++ b/test-data/unit/check-attr.test
@@ -542,9 +542,9 @@ def factory(default: int):
    ...
 @attr.s
 class C:
-    x: str = attr.ib(converter=thing.do_it)  # E: Unsupported converter function, only named functions are currently supported
-    y: str = attr.ib(converter=lambda x: x)  # E: Unsupported converter function, only named functions are currently supported
-    z: str = attr.ib(converter=factory(8))  # E: Unsupported converter function, only named functions are currently supported
+    x: str = attr.ib(converter=thing.do_it)  # E: Unsupported converter, only named functions and types are currently supported
+    y: str = attr.ib(converter=lambda x: x)  # E: Unsupported converter, only named functions and types are currently supported
+    z: str = attr.ib(converter=factory(8))   # E: Unsupported converter, only named functions and types are currently supported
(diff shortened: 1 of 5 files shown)


chadrik:
This unblocks: https://github.com/python-attrs/attrs/pull/238

test-data/unit/check-attr.test
```diff
@@ -567,6 +567,31 @@ reveal_type(A)  # E: Revealed type is 'def (x: builtins.int) -> __main__.A'
 reveal_type(A(15).x)  # E: Revealed type is 'builtins.str'
 [builtins fixtures/list.pyi]
 
+[case testAttrsUsingConverterWithTypes]
+from typing import overload
+import attr
+
+class complex:
```
ethanhs: Perhaps this should be moved to a fixture?

test-data/unit/check-incremental.test
```diff
@@ -3983,6 +3983,40 @@ main:2: error: Revealed type is 'def (x: builtins.list[builtins.int]) -> a.a.A'
...
+class D(C):
+    a: str = attr.ib()
+reveal_type(D)
+[file a.py]
+from typing import overload
+import attr
+class complex:
+    @overload
+    def __init__(self, re: float = 0.0, im: float = 0.0) -> None: ...
+    @overload
```
ethanhs: Could you also add tests for an overloaded converter function?

euresti: Thanks for this note.  It was actually broken!

euresti:
Thanks for the review! I've addressed all your concerns and fixed some more things that were broken.

ethanhs:
Thank you again for working on this @euresti !

PR #4918, opened 2018-04-16
#4897: allow `raise E` where E: Type[Exn]

### LINKED ISSUE #4897: Confusing behavior with `raise Class`
```python
from typing import Type, Mapping, TypeVar


class Foo(Exception):
    pass


class Bar(Foo):
    pass


m = {1: Bar, 2: Foo}  # Mapping[int, Type[Foo]]


def f(i):
    # type: (int) -> Type[Foo]
    return m[i]


def x():
    # type: () -> None
    reveal_type(f(1))
    raise f(1)


def y():
    # type: () -> None
    reveal_type(m[1])
    raise m[1]


# for completeness's sake
def huh(*args, **kwargs):
    # type: (*object, **object) -> Foo
    return Bar()


def z():
    # type: () -> None
    reveal_type(huh)
    raise huh
```

The first `reveal_type` shows that `f(1)` returns a `Type[Foo]`; `x()` does not typecheck, with the error "Exception must be derived from BaseException".

The second `reveal_type` shows that `m[1]` is `def (*args: builtins.object, **kwargs: builtins.object) -> foo.Foo`. `y()` typechecks.

The third `reveal_type` shows that `huh` is `def (*args: builtins.object, **kwargs: builtins.object) -> foo.Foo`. `z()` does not typecheck (same error as `x()`).

bwo:
That is, where E is annotated as a Type, rather than being inferred as a
FunctionLike which is a type object.

This also allows `raise E` where E is Type[Any], which *seems* right but
I was not certain about.

Fixes #4897

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2357,15 +2357,20 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
     def type_check_raise(self, e: Expression, s: RaiseStmt,
                          optional: bool = False) -> None:
         typ = self.expr_checker.accept(e)
-        if isinstance(typ, FunctionLike):
-            if typ.is_type_obj():
-                # Cases like "raise/from ExceptionClass".
-                typeinfo = typ.type_object()
-                base = self.lookup_typeinfo('builtins.BaseException')
-                if base in typeinfo.mro or typeinfo.fallback_to_any:
-                    # Good!
-                    return
-                # Else fall back to the checks below (which will fail).
+        typeinfo = None
+        if isinstance(typ, FunctionLike) and typ.is_type_obj():
+            typeinfo = typ.type_object()
+        elif isinstance(typ, TypeType):
+            if isinstance(typ.item, AnyType):
+                return
+            typeinfo = typ.item.type
+        if typeinfo is not None:
+            # Cases like "raise/from ExceptionClass".
+            base = self.lookup_typeinfo('builtins.BaseException')
+            if base in typeinfo.mro or typeinfo.fallback_to_any:
+                # Good!
+                return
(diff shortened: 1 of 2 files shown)


test-data/unit/check-statements.test
```diff
@@ -410,6 +410,12 @@ raise object # E: Exception must be derived from BaseException
 raise f # E: Exception must be derived from BaseException
 [builtins fixtures/exception.pyi]
 
+[case testRaiseExceptionType]
+import typing
+x = None # type: typing.Type[BaseException]
+raise x
```
JelleZijlstra: Could you also add a negative test (e.g., `x: Type[int]; raise x` still errors)?

JelleZijlstra:
I think allowing `Type[Any]` is right: Any means any type, so it may be a valid exception type.

test-data/unit/check-statements.test
bwo: done

bwo:
done

bwo:
ping?

mypy/checker.py
```diff
@@ -2357,6 +2357,10 @@ def visit_raise_stmt(self, s: RaiseStmt) -> None:
     def type_check_raise(self, e: Expression, s: RaiseStmt,
                          optional: bool = False) -> None:
         typ = self.expr_checker.accept(e)
+        if isinstance(typ, TypeType):
```
ilevkivskyi: Shouldn't (for consistency) the same case be added to `check_except_handler_test` (with test cases)?

msullivan:
@bwo, do you have a chance to look at Ivan's feedback?

Issue #2777, opened 2017-01-31
Functions with same signature but different parameter names.


sharmaeklavya2:
```python
def is_prime(n):
    # type: (int) -> bool
    return all(n % p != 0 for p in range(2, n))

def is_palindrome(m):
    # type: (int) -> bool
    return str(m) == ''.join(reversed(str(m)))

list1 = [is_prime, is_palindrome]
list2 = [is_prime, is_prime]

reveal_type(list1)
reveal_type(list2)
```

In this code, `is_prime` and `is_palindrome` have the same signature `def (builtins.int) -> builtins.bool` but different parameter names.

These are the revealed types:

* `list1`: `builtins.list[builtins.function*]`
* `list2`: `builtins.list[def (n: builtins.int) -> builtins.bool]`

I think the type of `list1` should have been `builtins.list[def (builtins.int) -> builtins.bool]`.

sharmaeklavya2:
This behaviour was not present in mypy 0.4.6. In mypy 0.4.6, revealed type of `list1` was `builtins.list[def (m: builtins.int) -> builtins.bool]`, which was wrong.

JukkaL:
This is a side effect of recent changes to callable types. The previous behavior was better for this example. In technical terms, mypy should be more clever about joins of callable types in general.

A simple workaround is to give `list1` an explicit annotation:

```
...
list1 = [is_prime, is_palindrome]  # type: List[Callable[[int], bool]]
```

PR #4920, opened 2018-04-17
Make joins of callables respect positional parameter names

Michael0x2a:
This commit fixes https://github.com/python/mypy/issues/2777 -- specifically, it enhances joining callables to erase the names of positional arguments as necessary.

For example, consider the following program:

    def f(x: int) -> int: ...
    def g(y: int) -> int: ...

    lst = [f, g]

Previously mypy would treat the final line as an error since 'f' and 'g' have different types due to the different parameter names.

Now, mypy infers that `lst` has type `list[def (int) -> int]`, effectively erasing the parameter name from the inferred type.

This commit does not attempt change how handle keyword-only arguments are currently handled.


test-data/unit/check-inference.test
```diff
@@ -856,13 +856,41 @@ s = s_s() # E: Incompatible types in assignment (expression has type "Set[str]",
...
 
 [case testSetWithStarExpr]
-
 s = {1, 2, *(3, 4)}
 t = {1, 2, *s}
 reveal_type(s)  # E: Revealed type is 'builtins.set[builtins.int*]'
 reveal_type(t)  # E: Revealed type is 'builtins.set[builtins.int*]'
 [builtins fixtures/set.pyi]
 
+[case testListLiteralWithFunctionsErasesNames]
```
msullivan: Maybe add a test for the `join_similar_callables` path by joining functions whose argument types differ but have a nontrivial meet. (For example, if the arguments are `Union[bytes, float]` and `Union[str, int]`, the argument for the joined function should be `int`)

Michael0x2a:
@msullivan -- sorry for the delay! I made the changes you suggested.

mypy/join.py
```diff
@@ -391,7 +391,17 @@ def combine_similar_callables(t: CallableType, s: CallableType) -> CallableType:
...
 
 def combine_arg_names(t: CallableType, s: CallableType) -> List[Optional[str]]:
-    # Note: assumes is_similar_type(t, s) is true
+    """Takes two callables and produces the meet of their argument names.
+
+    For example, if 't' had the signature (a: int, b: str, c: str) -> None
+    and 's' had the signature (a: int, b: str, z: str) -> None, the
+    join of their argument names would be ["a", "b", None].
+
+    This information is then used to compute more accurate joins for
```
msullivan: Meets and joins, right?

mypy/join.py
```diff
@@ -391,7 +391,17 @@ def combine_similar_callables(t: CallableType, s: CallableType) -> CallableType:
 
 
 def combine_arg_names(t: CallableType, s: CallableType) -> List[Optional[str]]:
-    # Note: assumes is_similar_type(t, s) is true
+    """Takes two callables and produces the meet of their argument names.
```
msullivan: Extreme nit: this calls this operation a meet in one place and a join in another, and should use the same term twice. I think it is more of a meet than a join (since it is like an intersection, which are meets?). Just saying "combination" or something fine also.

mypy/join.py
Michael0x2a: I think it's just joins -- this pull request doesn't touch meet.py.

(iirc, the meet of two types 't' and 's' would be some type that's basically a valid subtype of both. I couldn't think of a way of creating a callable that would be a valid subtype of two callables with different arg names without doing something messy with keyword arguments, so didn't attempt to modify that file.)

mypy/join.py
Michael0x2a: Oh whoops, that's embarrassing -- I didn't mean to use the word 'meet' here...

But yeah, I agree it's confusing, especially since this function isn't even manipulating types. I reworded the comment completely and just used the phrase "compatible with" which is hopefully better.

Issue #4854, opened 2018-04-04
Implementation not checked when abstract method is added in fine-grained mode


JukkaL:
This fine-grained incremental mode test case fails (no errors are generated):

```

[case testAddAbstractMethod]
from b import D
D()
[file b.py]
from a import C
class D(C):
    def f(self) -> None: pass
[file a.py]
from abc import abstractmethod
class C:
    @abstractmethod
    def f(self) -> None: pass
[file a.py.2]
from abc import abstractmethod
class C:
    @abstractmethod
    def f(self) -> None: pass
    @abstractmethod
    def g(self) -> None: pass
[out]
==
main:2: error: Cannot instantiate abstract class 'D' with abstract attribute 'g'
```

Analysis: Looks like the addition of an abstract attribute should trigger the `__init__` method of the class.

ilevkivskyi:
It looks like there is a similar issue when both an abstract method is added, and an implementation is added to a base class. I was not able to find a simple repro however.

JukkaL:
Looks like this was fixed by #4977. If something is still broken, we can create a new issue once we have a repro.

PR #4922, opened 2018-04-17
Fine-grained: properly trigger when the set of abstract attrs changes

msullivan:


ilevkivskyi:
@JukkaL I will leave this to you as you are also working on https://github.com/python/mypy/issues/4854

JukkaL:
I can take this. I hadn't started work on #4854 yet.

mypy/server/astdiff.py
```diff
@@ -193,6 +193,8 @@ def snapshot_definition(node: Optional[SymbolNode],
                  node.fallback_to_any,
                  node.is_named_tuple,
                  node.is_newtype,
+                 # Whether subclasses are abstract can depend on the set of abstract attrs
+                 tuple(sorted(node.abstract_attributes)),
```
JukkaL: This change can make changes to abstract classes much more expensive to process that normal classes. I have an idea about how to do this without having the trigger the whole class. Let's not merge this for now.

PR #4931, opened 2018-04-18
Compute mangled fullnames properly for nested classes inside functions

msullivan:
This reverts #4927 and fixes the (a?) bug that it was working around.

Previously, mangled local fullnames were only computed if the class was directly inside a function, and not for nested classes inside a function.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -995,7 +995,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
             defn.info._fullname = defn.info.name()
         if self.is_func_scope() or self.type:
             kind = MDEF
-            if self.is_nested_func_scope():
+            if self.is_nested_within_func_scope():
                 kind = LDEF
             node = SymbolTableNode(kind, defn.info)
             self.add_symbol(defn.name, node, defn)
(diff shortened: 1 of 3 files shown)


mypy/semanal.py
```diff
@@ -995,7 +995,7 @@ def setup_class_def_analysis(self, defn: ClassDef) -> None:
             defn.info._fullname = defn.info.name()
         if self.is_func_scope() or self.type:
             kind = MDEF
-            if self.is_func_scope():
+            if self.is_nested_func_scope():
```
ilevkivskyi: I am not sure this is the right solution. The nested class is clearly an `MDEF`, but on the other hand all these kinds will be refactored soon, so maybe it is OK. @JukkaL what do you think?

JukkaL: It's okay to call these LDEFs, since these have access to the local variables of an enclosing function, which is the most important distinction for classes. Whether it's nested within a class or not is less important.

mypy/semanal.py
```diff
@@ -3109,6 +3109,10 @@ def leave(self) -> None:
     def is_func_scope(self) -> bool:
         return self.locals[-1] is not None
 
+    # Are we underneath a function scope, even if we are in a nested class also
```
JukkaL: Move the comment to a docstring.

mypy/semanal.py
```diff
@@ -3109,6 +3109,10 @@ def leave(self) -> None:
...
+    def is_nested_func_scope(self) -> bool:
```
JukkaL: I was a bit confused by the name. Maybe rename to `is_nested_within_func_scope`?

PR #4937, opened 2018-04-19
Document how to use a remote cache with mypy

JukkaL:


JukkaL:
We should add a discussion of `--cache-fine-grained` once we document the mypy daemon.

PR #4939, opened 2018-04-19
Add support for user-specific config file

### LINKED ISSUE #4815: User-specific config file

OmeGak:
Closes #4815.

This PR makes `mypy` fallback to reading the configuration from `~/.mypy.ini` if neither `mypy.ini` nor `setup.cfg` are found when invoked without `--config-file`. Also, `--help` will now mention all default config files under `--config-file CONFIG_FILE`, instead of mentioning only `mypy.ini`.

Missing:
- [x] Update docs.
- [ ] Tests?


gvanrossum:
Thanks! Especially thank you for also updating the docs.

PR #4953, opened 2018-04-23
Move subtype caches to a separate global class

ilevkivskyi:
This refactoring doesn't have much value in itself, but will be helpful to simplify and speed-up protocols in fine grained mode, see #4790.


ilevkivskyi:
I am intentionally using class variables and class methods instead of a singleton because I think creating a singleton somewhere in `build.py` will only complicate the code.

JukkaL:
My guess is that the travis failures are due to high memory use. You can try lowering travis parallelism. If it helps, then that's likely the cause.

ilevkivskyi:
Yes, it was memory problem, thanks! But instead of reducing parallelism, I reset caches completely on every `build.build` (i.e. on everything that supposed to be a cold run, but apparently tests don't restart mypy completely).

mypy/build.py
```diff
@@ -265,6 +266,7 @@ def _build(sources: List[BuildSource],
                            fscache=fscache)
 
     try:
+        TypeState.reset_all()
```
JukkaL: If we aren't running a fine-grained incremental build, I think that we can reset the state after the build (before returning the build results), since it won't be useful any more at than point but it will consume some memory. We should also reset the cache when shutting down the daemon, maybe in `reset()`. It's good manners to not leave global state lying around at the end of a program.

Style nit: Move this to be outside the try statement, since we don't expect this to fail.

mypy/nodes.py
```diff
@@ -2062,8 +2058,7 @@ def __init__(self, names: 'SymbolTable', defn: ClassDef, module_name: str) -> No
         self.assuming = []
         self.assuming_proper = []
         self.inferring = []
-        self._cache = set()
-        self._cache_proper = set()
+        TypeState.add_caches(self)
```
JukkaL: What if this call would be optional? The TypeState API would be a bit simpler if `TypeState` would deal with adding entries as needed. Generally I'd prefer to keep public APIs as simple as possible and hide all logic within the implementation. Now the API is leaking an implementation detail (cache entry needs to exist for a class before it can be used for caches).

mypy/nodes.py
```diff
@@ -2099,25 +2094,9 @@ def get_containing_type_info(self, name: str) -> 'Optional[TypeInfo]':
                 return cls
         return None
 
-    def record_subtype_cache_entry(self, left: 'mypy.types.Instance',
```
JukkaL: Good to have this logic moved away from this module!

mypy/nodes.py
```diff
@@ -2099,25 +2094,9 @@ def get_containing_type_info(self, name: str) -> 'Optional[TypeInfo]':
...
-                                proper_subtype: bool = False) -> bool:
-        if not proper_subtype:
-            return (left, right) in self._cache
-        return (left, right) in self._cache_proper
-
     def reset_subtype_cache(self) -> None:
         for item in self.mro:
-            item._cache = set()
-            item._cache_proper = set()
+            TypeState.reset_caches(item)
```
JukkaL: We could simplify module dependencies a bit by moving this method to `TypeState`. This would have the benefit of `TypeInfo` not needing any knowledge of `TypeState` (together with the above idea). Conceptually then `TypeState` would only hold information about speeding up certain `TypeInfo` operations, but it would be separate from `TypeInfo`. It's better if TypeInfo doesn't need to be aware of operations available for TypeInfos.

In the future we can move additional 'transient' global state from TypeInfos to TypeState (or some other place), such as `inferring` and `assuming` (it's best to not do it in this PR to avoid making this big).

mypy/typestate.py
```diff
@@ -0,0 +1,55 @@
+"""
+A shared state for all TypeInfo's that holds global cache and dependency information,
```
JukkaL: Can you make it a bit more explicit that this is mutable global state (basically global variables)?

Grammar nit: TypeInfos, not TypeInfo's.

mypy/typestate.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import Dict, Set, Tuple
+
+MYPY = False
+if MYPY:
+    from typing import ClassVar
+    from mypy.nodes import TypeInfo
+    from mypy.types import Instance
+
+
+class TypeState:
```
JukkaL: Add docstring. In the docstring, maybe discuss at least these things:

* Purpose of this class.
* How to avoid leaking global state: this needs to be explicitly cleared after a build has finished and after a daemon shutdown.
* Invariants: Resetting subtype caches for a class has no semantic effect (only perf).

mypy/typestate.py
```diff
@@ -0,0 +1,55 @@
...
+    # 'caches' and 'caches_proper' are subtype caches, implemented as sets of pairs
+    # of (subtype, supertype), where supertypes are instances of given TypeInfo.
+    # We need the caches, since subtype checks for structural types are very slow.
```
JukkaL: Explicitly mention that `caches_proper` if for proper subtype checks and `caches` for normal subtype checks.

mypy/typestate.py
```diff
@@ -0,0 +1,55 @@
...
+    caches = {}  # type: ClassVar[Dict[TypeInfo, Set[Tuple[Instance, Instance]]]]
```
JukkaL: Maybe rename these to `subtype_caches` and `proper_subtype_caches` (then the above comment becomes moot). Maybe also use `_` prefix since these should only be accessed through methods.

mypy/typestate.py
```diff
@@ -0,0 +1,55 @@
...
+    caches_proper = {}  # type: ClassVar[Dict[TypeInfo, Set[Tuple[Instance, Instance]]]]
+
+    @classmethod
+    def reset_all(cls) -> None:
```
JukkaL: Maybe rename to `clear` or `clear_caches`?

mypy/typestate.py
```diff
@@ -0,0 +1,55 @@
...
+        cls.caches = {}
+        cls.caches_proper = {}
+
+    @classmethod
+    def add_caches(cls, info: 'TypeInfo') -> None:
```
JukkaL: As discussed above, this seems unnecessary in the public API. We can create the caches on demand here.

mypy/typestate.py
```diff
@@ -0,0 +1,55 @@
...
+        cls.caches_proper[info] = set()
+
+    @classmethod
+    def reset_caches(cls, info: 'TypeInfo') -> None:
+        cls.caches[info].clear()
+        cls.caches_proper[info].clear()
+
+    @classmethod
+    def is_cached_subtype_check(cls, left: 'Instance', right: 'Instance',
+                                proper_subtype: bool = False) -> bool:
```
JukkaL: Style nit: I think that it would be better to have two methods, one for subtype checks and the other for proper subtype checks. Bool mode arguments are a minor code smell in Python.

mypy/typestate.py
```diff
@@ -0,0 +1,55 @@
...
+        info = right.type
+        if proper_subtype:
+            cache = cls.caches_proper[info]
+        else:
+            cache = cls.caches[info]
+        return (left, right) in cache
+
+    @classmethod
+    def record_subtype_cache_entry(cls, left: 'Instance', right: 'Instance',
+                                   proper_subtype: bool = False) -> None:
```
JukkaL: Like above, I'd prefer two variants of this method.

mypy/dmypy_server.py
```diff
@@ -152,6 +153,7 @@ def serve(self) -> None:
                         conn, addr = sock.accept()
                     except socket.timeout:
                         print("Exiting due to inactivity.")
+                        TypeState.reset_all_subtype_caches()
```
JukkaL: Maybe move this call to a method (say, `free_global_state`) since this call and the identical one below must stay in sync, and by having them in a single method makes that easier.

mypy/dmypy_server.py
```diff
@@ -182,6 +184,7 @@ def serve(self) -> None:
                     conn.close()
                     if command == 'stop':
                         sock.close()
+                        TypeState.reset_all_subtype_caches()
```
JukkaL: This is the other call I mentioned above.

mypy/nodes.py
```diff
@@ -2133,7 +2128,7 @@ def calculate_mro(self) -> None:
         self.mro = mro
         # The property of falling back to Any is inherited.
         self.fallback_to_any = any(baseinfo.fallback_to_any for baseinfo in self.mro)
-        self.reset_subtype_cache()
+        TypeState.reset_all_subtype_caches_for(self)
```
JukkaL: Bummer that we still have the circular dependency. We could move the `calculate_mro` method to another module (that `mypy.node` doesnt't depend on) to get rid of it. Can you do it or at least add a TODO comment about this refactor?

mypy/build.py
```diff
@@ -265,9 +265,11 @@ def _build(sources: List[BuildSource],
                            flush_errors=flush_errors,
                            fscache=fscache)
 
+    TypeState.reset_all_subtype_caches()
     try:
-        TypeState.reset_all()
         graph = dispatch(sources, manager)
+        if not options.cache_fine_grained or options.fine_grained_incremental:
```
JukkaL: Should we actually clear the caches when using `cache_fine_grained`? I think that it only generates some additional module metadata that is not related to subtype checks.

ilevkivskyi: Indeed, we don't need it.

PR #4957, opened 2018-04-24
Don't decode files while computing their hash

msullivan:
Instead, cache undecoded bytes and decode in client code.  Happily,
this allows us to get rid of FileSystemCache's dependency on the
python version.

This is a big performance win for daemon cold runs on codebases that
make use of custom encodings where decoding is expensive.

This gives about a 5s speed up on cold runs on S when the cache mtimes don't match.

### DIFF
--- a/mypy/fscache.py
+++ b/mypy/fscache.py
@@ -29,10 +29,10 @@ advantage of the benefits.
 """
 
 import functools
+import hashlib
 import os
 import stat
 from typing import Dict, List, Tuple
-from mypy.util import read_with_python_encoding
 
 
 class FileSystemMetaCache:
@@ -121,20 +121,20 @@ class FileSystemMetaCache:
         return True
 
 
+# TODO: Merge FileSystemMetaCache back into this?
 class FileSystemCache(FileSystemMetaCache):
-    def __init__(self, pyversion: Tuple[int, int]) -> None:
+    def __init__(self) -> None:
         super().__init__()
-        self.pyversion = pyversion
         self.flush()
 
     def flush(self) -> None:
         """Start another transaction and empty all caches."""
         super().flush()
-        self.read_cache = {}  # type: Dict[str, str]
+        self.read_cache = {}  # type: Dict[str, bytes]
         self.read_error_cache = {}  # type: Dict[str, Exception]
         self.hash_cache = {}  # type: Dict[str, str]
 
-    def read_with_python_encoding(self, path: str) -> str:
+    def read(self, path: str) -> bytes:
         if path in self.read_cache:
             return self.read_cache[path]
         if path in self.read_error_cache:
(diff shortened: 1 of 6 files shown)


mypy/util.py
```diff
@@ -68,40 +69,34 @@ class DecodeError(Exception):
...
+    f = BytesIO(source)
+    source_start = f.readline() + f.readline()
+
+    # check for BOM UTF-8 encoding and strip it out if present
+    if source_start.startswith(b'\xef\xbb\xbf'):
+        encoding = 'utf8'
+        source = source[3:]
+    else:
+        _encoding, _ = find_python_encoding(source_start, pyversion)
+        # check that the coding isn't mypy. We skip it since
```
gvanrossum: I'm pretty sure we got rid of the mypy codec in 2016.

mypy/util.py
```diff
@@ -3,6 +3,7 @@
 import re
 import subprocess
 import hashlib
+from io import BytesIO
```
gvanrossum: I'd prefer the style `import io` here (in general for stdlib modules, though I'm fine with `escape` below).

mypy/fscache.py
```diff
@@ -121,20 +121,20 @@ def exists(self, path: str) -> bool:
         return True
 
 
+# TODO: Merge FileSystemMetaCache back into this?
```
gvanrossum: Would it be much work to do that now? Why keep them separate?

mypy/util.py
```diff
@@ -68,40 +69,34 @@ class DecodeError(Exception):
...
-        m.update(remainder)
-        source_bytearray.extend(remainder)
-        try:
-            source_text = source_bytearray.decode(encoding)
-        except LookupError as lookuperr:
-            raise DecodeError(str(lookuperr))
-        return source_text, m.hexdigest()
+    # look at first two lines and check if PEP-263 coding is present
+    f = BytesIO(source)
+    source_start = f.readline() + f.readline()
```
gvanrossum: Do you really need to capture the first two lines to pass to `find_python_encoding()`? It would *seem* that function works just as well if you just pass it `source`.

mypy/util.py
```diff
@@ -68,40 +69,34 @@ class DecodeError(Exception):
...
+
+    # check for BOM UTF-8 encoding and strip it out if present
+    if source_start.startswith(b'\xef\xbb\xbf'):
+        encoding = 'utf8'
+        source = source[3:]
```
gvanrossum: (For a large file this copies (nearly) the entire file an extra time. But I don't know if anyone cares.)

msullivan: Nobody caring was my theory, yeah

mypy/fscache.py
msullivan: Not much work, but I didn't want to clutter up this PR. I'll merge them in another PR after this?

mypy/util.py
msullivan: What would the right way to fix that be, though? Is there a way to decode a substring of a bytes?

mypy/util.py
msullivan: Looks that way, yeah

PR #4958, opened 2018-04-24
allow __getattr__ in nested scopes; support module-level __getattr__

### LINKED ISSUE #4950: Naming a closure __getattr__ triggers `Invalid signature`
Mypy won't let me name a closure `__getattr__`. Since it's not a method I'm not sure why mypy would care what the signature is. This is somewhat important, though, as Pytho 3.7 supports `__getattr__` functions in modules.

* Are you reporting a bug, or opening a feature request? bug
```python
def foo():
    def __getattr__():
        pass
    return __getattr__
```
* What is the actual behavior/output? `tester.py:2: error: Invalid signature "def () -> Any"`
* What is the behavior/output you expect? No error
* What are the versions of mypy and Python you are using? CPython master, mypy 0.59
  Do you see the same issue after installing mypy from Git master? yes
* What are the mypy flags you are using? (For example --strict-optional) None

JelleZijlstra:
Fixes #4950 and implements PEP 562 for mypy.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1064,30 +1064,33 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
 
     def check_getattr_method(self, typ: CallableType, context: Context, name: str) -> None:
         if len(self.scope.stack) == 1:
-            # module-level __getattr__
+            # module scope
             if name == '__getattribute__':
                 self.msg.fail('__getattribute__ is not valid at the module level', context)
                 return
-            elif name == '__getattr__' and not self.is_stub:
-                self.msg.fail('__getattr__ is not valid at the module level outside a stub file',
-                              context)
-                return
+            # __getattr__ is fine at the module level as of Python 3.7 (PEP 562). We could
+            # show an error for Python < 3.7, but that would be annoying in code that supports
+            # both 3.7 and older versions.
             method_type = CallableType([self.named_type('builtins.str')],
                                        [nodes.ARG_POS],
                                        [None],
                                        AnyType(TypeOfAny.special_form),
                                        self.named_type('builtins.function'))
-        else:
+        elif self.scope.active_class():
             method_type = CallableType([AnyType(TypeOfAny.special_form),
                                         self.named_type('builtins.str')],
                                        [nodes.ARG_POS, nodes.ARG_POS],
                                        [None, None],
                                        AnyType(TypeOfAny.special_form),
                                        self.named_type('builtins.function'))
+        else:
+            return
         if not is_subtype(typ, method_type):
             self.msg.invalid_signature(typ, context)
 
     def check_setattr_method(self, typ: CallableType, context: Context) -> None:
+        if not self.scope.active_class():
+            return
         method_type = CallableType([AnyType(TypeOfAny.special_form),
                                     self.named_type('builtins.str'),
                                     AnyType(TypeOfAny.special_form)],
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1334,7 +1334,8 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
                     missing = True
             # If it is still not resolved, and the module is a stub
             # check for a module level __getattr__
-            if module and not node and module.is_stub and '__getattr__' in module.names:
+            if (module and not node and (module.is_stub or self.options.python_version >= (3, 7))
+                    and '__getattr__' in module.names):
                 getattr_defn = module.names['__getattr__']
                 if isinstance(getattr_defn.node, FuncDef):
                     if isinstance(getattr_defn.node.type, CallableType):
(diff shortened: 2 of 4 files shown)


mypy/checker.py
```diff
@@ -1064,30 +1064,33 @@ def check_inplace_operator_method(self, defn: FuncBase) -> None:
...
-            # module-level __getattr__
+            # module scope
             if name == '__getattribute__':
                 self.msg.fail('__getattribute__ is not valid at the module level', context)
                 return
-            elif name == '__getattr__' and not self.is_stub:
-                self.msg.fail('__getattr__ is not valid at the module level outside a stub file',
-                              context)
-                return
+            # __getattr__ is fine at the module level as of Python 3.7 (PEP 562). We could
```
ethanhs: Hm, I think this could actually bite people more than you'd think. People already get confused about why `__getattr__` works in stubs and not in runtime modules without mypy allowing it in non-stubs. Also code that supports both 3.7 and below probably wouldn't be able to use `__getattr__`, I can't think of a non-hacky way in which it would work at runtime...

ilevkivskyi: I think it is actually fine. We don't warn about `__setattr__` being unsupported etc.

test-data/unit/check-classes.test
```diff
@@ -1887,6 +1887,18 @@ b = a.bar
...
+    def __getattr__() -> None:  # no error because not in a class
+        pass
+    return __getattr__
+
+class X:
+    def foo(self) -> object:
+        def __getattr__() -> None:  # no error because not directly inside a class
+            pass
+        return __getattr__
+
```
ilevkivskyi: I would add  two more tests like this:
```python
def make_getattr_good() -> Callable[[str], int]: ...
__getattr__ = make_getattr_good()  # OK

def make_getattr_bad() -> Callable[[], int]: ...
__getattr__ = make_getattr_bad()  # E: Invalid signature...
```

Because this may be one of typical use cases for PEP 572. This will likely require to also call `check_getattr_method` from `visit_assignment_statement` (not only from function definition like now).

JelleZijlstra: Good point. I'll also check for `from other_module import __getattr__`.

JelleZijlstra: And another one I realized I don't cover in the code I just pushed: `from other_module import other_name as __getattr__`.

JelleZijlstra:
@ilevkivskyi this is ready for another review now.

mypy/semanal.py
```diff
@@ -1335,11 +1335,11 @@ def visit_import_from(self, imp: ImportFrom) -> None:
                     self.add_submodules_to_parent_modules(possible_module_id, True)
                 elif possible_module_id in self.missing_modules:
                     missing = True
-            # If it is still not resolved, and the module is a stub
-            # check for a module level __getattr__
-            if module and not node and module.is_stub and '__getattr__' in module.names:
+            # If it is still not resolved, check for a module level __getattr__
+            if (module and not node and (module.is_stub or self.options.python_version >= (3, 7))
```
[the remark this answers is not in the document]
JelleZijlstra: I think that's different actually. In `checker.py` we don't show an error if the user uses module-level `__getattr__` before 3.7, because it's harmless (it does nothing, but also doesn't break anything). Here though we're essentially implementing the equivalent of the runtime attribute lookup on modules, so we should only look at `__getattr__` in 3.7 or later.

mypy/checker.py
```diff
@@ -1483,6 +1486,19 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
         else:
             lvalue_type, index_lvalue, inferred = self.check_lvalue(lvalue)
 
+            if isinstance(lvalue, NameExpr):
```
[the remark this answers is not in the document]
JelleZijlstra: Rearranged the code a bit. I think the only logic change is that we may call `accept` on the `rvalue` more than once, which is probably fine because errors get deduplicated somewhere.

Relatedly, the way this method is structured makes it hard to verify that we actually call `accept` on the rvalue at all times. It might be worth it to refactor the code to just call `self.expr_checker.accept(rvalue)` once at the top.

mypy/checker.py
```diff
@@ -1483,6 +1486,23 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
...
+            if isinstance(lvalue, NameExpr) and lvalue.node:
+                name = lvalue.node.name()
+                if name in ('__setattr__', '__getattribute__', '__getattr__'):
+                    # If an explicit type is given, use that.
+                    if lvalue_type:
+                        signature = lvalue_type
+                    else:
+                        signature = self.expr_checker.accept(rvalue)
+                    if signature:
+                        name = lvalue.node.name()
```
ilevkivskyi: This line is now redundant, the name is already assigned above.

JelleZijlstra: Fixed, thanks for the review.

Issue #4951, opened 2018-04-23
Fine-grained: Python 3 features


JukkaL:
These Python 3 features should be tested with fine-grained incremental checking (may require code changes):

* [ ] yield from
* [ ] f-strings
* [ ] extended unpacking
* [ ] async and await related features
* [ ] data classes

ilevkivskyi:
Data classes could be problematic for 0.600 because we don't have them even in normal mode. It is not something hard, but it will take time to write the plugin, so I think we can postpone data classes to 0.610 (which will be still before Python 3.7.0 final).

JukkaL:
Oops will remove them from the list.

PR #4967, opened 2018-04-25
Fine-grained: Test f string

JukkaL:
Work towards #4951.


test-data/unit/fine-grained.test
```diff
@@ -5800,3 +5800,15 @@ class M(type):
 [out]
 ==
 a.py:2: error: Argument 1 to "f" of "M" has incompatible type "int"; expected "str"
+
+[case testFString]
+from a import g
+f'{g(1)}'
```
ilevkivskyi: I would add a test where f-string is at a different scope (class or function) and/or where `g` turns into a class, if you have time.

PR #4968, opened 2018-04-25
Fine-grained: Test "yield from"

JukkaL:
Work towards #4951.

### DIFF
--- a/test-data/unit/fine-grained.test
+++ b/test-data/unit/fine-grained.test
@@ -5800,3 +5800,31 @@ class M(type):
 [out]
 ==
 a.py:2: error: Argument 1 to "f" of "M" has incompatible type "int"; expected "str"
+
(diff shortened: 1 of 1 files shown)


test-data/unit/fine-grained.test
```diff
@@ -5800,3 +5800,31 @@ class M(type):
...
+[file a.py.2]
+from typing import Generator
+
+class A: pass
+
+def f() -> Generator[int, None, A]:
+    yield 5
+    return A()
+
+[out]
```
ilevkivskyi: I would add a test where `f` is turned into a normal function (just in case).

JukkaL: Verified that this works but left the test out since this seems unlikely to break (to conserve test running times).

PR #4969, opened 2018-04-25
Fine-grained: Support async def/for/with and await

JukkaL:
Some of the changes motivated cleaning up some duplicate error 
messages.

Also noticed missing `__next__`/`next` dependency in ordinary for
statements and fixed it as well.

Work towards #4951.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2530,12 +2530,6 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         """Analyse async iterable expression and return iterator and iterator item types."""
         echk = self.expr_checker
         iterable = echk.accept(expr)
-
-        self.check_subtype(iterable,
-                           self.named_generic_type('typing.AsyncIterable',
-                                                   [AnyType(TypeOfAny.special_form)]),
-                           expr, messages.ASYNC_ITERABLE_EXPECTED)
-
         method = echk.analyze_external_member_access('__aiter__', iterable, expr)
         iterator = echk.check_call(method, [], [], expr)[0]
         method = echk.analyze_external_member_access('__anext__', iterator, expr)
@@ -2558,11 +2552,6 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             return iterator, joined
         else:
             # Non-tuple iterable.
-            self.check_subtype(iterable,
-                               self.named_generic_type('typing.Iterable',
-                                                       [AnyType(TypeOfAny.special_form)]),
-                               expr, messages.ITERABLE_EXPECTED)
-
             if self.options.python_version[0] >= 3:
                 nextmethod = '__next__'
             else:
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2518,15 +2518,16 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
     def visit_for_stmt(self, s: ForStmt) -> None:
         """Type check a for statement."""
         if s.is_async:
-            item_type = self.analyze_async_iterable_item_type(s.expr)
+            iterator_type, item_type = self.analyze_async_iterable_item_type(s.expr)
         else:
-            item_type = self.analyze_iterable_item_type(s.expr)
+            iterator_type, item_type = self.analyze_iterable_item_type(s.expr)
         s.inferred_item_type = item_type
+        s.inferred_iterator_type = iterator_type
         self.analyze_index_variables(s.index, item_type, s.index_type is None, s)
         self.accept_loop(s.body, s.else_body)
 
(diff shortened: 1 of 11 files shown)


mypy/checker.py
```diff
@@ -2518,57 +2518,47 @@ def get_types_from_except_handler(self, typ: Type, n: Expression) -> List[Type]:
...
+        """Analyse iterable expression and return iterator and iterator item types."""
         echk = self.expr_checker
         iterable = echk.accept(expr)
+        method = echk.analyze_external_member_access('__iter__', iterable, expr)
+        iterator = echk.check_call(method, [], [], expr)[0]
 
         if isinstance(iterable, TupleType):
             joined = UninhabitedType()  # type: Type
             for item in iterable.items:
                 joined = join_types(joined, item)
```
ilevkivskyi: Ideally, this would be already available in the tuple fallback instance, but it doesn't work yet because of problems with forward refs (I will fix this later).

test-data/unit/check-async-await.test
```diff
@@ -247,14 +246,10 @@ async def wrong_iterable(obj: Iterable[int]):
...
 [out]
-main:18: error: AsyncIterable expected
-main:18: error: "Iterable[int]" has no attribute "__aiter__"; maybe "__iter__"?
-main:19: error: Iterable expected
-main:19: error: "asyncify[int]" has no attribute "__iter__"; maybe "__aiter__"?
-main:20: error: AsyncIterable expected
-main:20: error: "Iterable[int]" has no attribute "__aiter__"; maybe "__iter__"?
-main:21: error: Iterable expected
-main:21: error: "asyncify[int]" has no attribute "__iter__"; maybe "__aiter__"?
+main:18: error: "Iterable[int]" has no attribute "__aiter__"; maybe "__iter__"? (not async iterable)
```
ilevkivskyi: If it is easy to change, then I would skip alternative name suggestion for these particular names. If it is not easy then ignore this comment.

test-data/unit/fine-grained.test
```diff
@@ -5800,3 +5800,123 @@ class M(type):
 [out]
 ==
 a.py:2: error: Argument 1 to "f" of "M" has incompatible type "int"; expected "str"
+
+[case testAwaitAndAsyncDef-skip-cache]
```
ilevkivskyi: Why skip cache here and in some tests below?

test-data/unit/fine-grained.test
```diff
@@ -5800,3 +5800,123 @@ class M(type):
...
+    async def __aenter__(self) -> int: pass
+
+[builtins fixtures/async_await.pyi]
+[typing fixtures/typing-full.pyi]
+[out]
+==
+main:5: error: Incompatible return value type (got "str", expected "int")
+==
+==
+main:4: error: "C" has no attribute "__aexit__"
```
ilevkivskyi: I would add few tests where iterators and context managers change kind (i.e. from normal to async and back).

test-data/unit/fine-grained.test
```diff
@@ -5800,3 +5800,123 @@ class M(type):
...
+==
+main:6: error: Incompatible return value type (got "str", expected "int")
+==
+main:6: error: Incompatible return value type (got "object", expected "int")
+
+[case testAsyncWith2-skip-cache]
+from a import C
+
+async def f() -> int:
+    async with C() as x:
```
ilevkivskyi: I would add a test without assignment target (`x` in this case).

test-data/unit/fine-grained.test
JukkaL: To avoid running the test cases twice -- the second run doesn't seem very useful.

test-data/unit/check-async-await.test
JukkaL: Will tweak this a a bit. The `__aiter__` suggestion seems potentially useful though.

test-data/unit/fine-grained.test
JukkaL: I wrote these tests and they passed but they don't seem to increase test coverage in a meaningful way so I'd like to leave them out. (Not sure if you've noticed but I'm getting tired of the slow local test runs... I suspect that the tests are over 2x slower than not too long ago.)

test-data/unit/fine-grained.test
JukkaL: Will add to this same test.

test-data/unit/fine-grained.test
ilevkivskyi: I didn't notice they are so much slower. Last run took 155 sec. on my laptop, which is not much. Maybe we can think about making tests more lightweight, thinking about which tests to include may impose more mental burden than actually waiting, but this may be subjective.

PR #4971, opened 2018-04-25
Enable strict optional by default

### LINKED ISSUE #359: Type checking of None values
A common pattern in alot of functions is to use None to signify something alternative behavior.
It might be beneficial to include a convenience Maybe type that is a partial application of Union with None. 

I'm still familiarizing myself with the code and am not 100% what to do yet

JukkaL:
As an exception, tests still don't do this since there are
hundreds of tests that would have to be updated. If tests
explicitly define command line options, strict optional will
be on by default -- basically command line options in tests
default to `--no-strict-optional`.

Fixes #359.

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -51,7 +51,7 @@ flagged as an error.
 
   If you don't know what types to add, you can use ``Any``, but beware:
 
-- **One of the values involved has type Any.** Extending the above
+- **One of the values involved has type 'Any'.** Extending the above
   example, if we were to leave out the annotation for ``a``, we'd get
   no error:
 
@@ -85,7 +85,7 @@ flagged as an error.
   clarity about the latter use ``--follow-imports=error``.  You can
   read up about these and other useful flags in :ref:`command-line`.
 
-- **A function annotated as returning a non-optional type returns None
+- **A function annotated as returning a non-optional type returns 'None'
   and mypy doesn't complain**.
 
   .. code-block:: python
@@ -129,6 +129,17 @@ The second line is now fine, since the ignore comment causes the name
     type if mypy cannot find information about that particular module. So,
     if we did have a stub available for ``frobnicate`` then mypy would
     ignore the ``# type: ignore`` comment and typecheck the stub as usual.
+
+
+Unexpected errors about 'None' and/or 'Optional' types
+------------------------------------------------------
+
+Starting from mypy 0.600, mypy uses
+:ref:`strict optional checking <strict_optional>` by default,
+and ``None`` is not compatible with non-optional types.  It's
+easy to switch back to the older behavior where ``None`` was
+compatible with arbitrary types (see :ref:`no_strict_optional`).
+
 
 Types of empty collections
 --------------------------
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -51,7 +51,7 @@ flagged as an error.
 
   If you don't know what types to add, you can use ``Any``, but beware:
 
-- **One of the values involved has type ``Any``.** Extending the above
+- **One of the values involved has type Any.** Extending the above
   example, if we were to leave out the annotation for ``a``, we'd get
   no error:
 
@@ -85,7 +85,7 @@ flagged as an error.
   clarity about the latter use ``--follow-imports=error``.  You can
   read up about these and other useful flags in :ref:`command-line`.
 
-- **A function annotated as returning a non-optional type returns ``None``
+- **A function annotated as returning a non-optional type returns None
   and mypy doesn't complain**.
 
   .. code-block:: python
@@ -93,9 +93,8 @@ flagged as an error.
       def foo() -> str:
           return None  # No error!
 
-  By default, the ``None`` value is considered compatible with everything. See
-  :ref:`optional` for details on strict optional checking, which allows mypy to
-  check ``None`` values precisely, and will soon become default.
+  You may have disabled strict optional checking (see
+  :ref:`no_strict_optional` for more).
 
 .. _silencing_checker:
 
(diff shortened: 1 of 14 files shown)


docs/source/kinds_of_types.rst
```diff
@@ -345,50 +345,176 @@ narrow down the type to a specific type:
...
+       n = None  # Inferred type Optional[int] because of the assignment below
        if i > 0:
             n = i
        ...
 
-Often it's useful to know whether a variable can be
+.. note::
+
+    ``None`` is also used as the return type for functions that don't
+    return a value, i.e. that implicitly return ``None``. (Mypy doesn't
```
ilevkivskyi: PEP 484 already specify this, I am not sure we need to discuss this here again (on the other hand there was a recent issue that proposed to improve some wording there).

runtests.py
```diff
@@ -274,16 +274,17 @@ def add_stdlibsamples(driver: Driver) -> None:
...
-                                    cwd=stdlibsamples_dir)
+                                    cwd=stdlibsamples_dir, extra_args=['--no-strict-optional'])
 
 
 def add_samples(driver: Driver) -> None:
     for f in find_files(os.path.join('test-data', 'samples'), suffix='.py'):
         if f == os.path.join('test-data', 'samples', 'crawl2.py'):
             # This test requires 3.5 for async functions
-            driver.add_mypy_cmd('file {}'.format(f), ['--python-version=3.5', f])
+            driver.add_mypy_cmd('file {}'.format(f), ['--python-version=3.5',
```
ilevkivskyi: I am not sure we actually need the `--python-version=3.5` flag. I thought mypy now uses the latest Python version by default.

docs/source/kinds_of_types.rst
kamahen: Do you expect mypy users to read through PEP 484 (which is rather long and complex)? Especially for a fairly basic thing like this.

JukkaL: I'd like the mypy documentation to be self-contained and PEP 484 should be strictly optional reading.

docs/source/common_issues.rst
```diff
@@ -51,7 +51,7 @@ flagged as an error.
 
   If you don't know what types to add, you can use ``Any``, but beware:
 
-- **One of the values involved has type ``Any``.** Extending the above
```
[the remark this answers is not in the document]
JukkaL: Good idea, I'll give it a try.

docs/source/kinds_of_types.rst
ilevkivskyi: Then I would reformulate this to make message more clear. Currently it starts from mentioning return type, but then  moves to something else. Also I would not mention `NoneType` at all, it can only add to the confusion, I would just re-iterate that for simplicity in `None` in type context means a type with a single value `None`. And separately mention that Python (this is not specific to mypy) returns `None` from a function that either reaches its end without return statement, or has a bare return.

docs/source/kinds_of_types.rst
```diff
@@ -345,50 +345,176 @@ narrow down the type to a specific type:
...
+    use ``NoneType`` for this, since it would
+    look awkward, even though that is the real name of the type of ``None``
+    -- try ``type(None)`` in the interactive interpreter to see for yourself.)
+
+.. note::
+
+    ``Optional[...]`` *does not* mean a function argument with a default value.
+    However, if the default value of an argument is ``None``, you can use
+    an optional type for the argument.
```
gvanrossum: Maybe make this stronger? There's a flag `--no-implicit-optional` that requires using `Optional[]` here, and there's talk of deprecating the default behavior altogether.

docs/source/kinds_of_types.rst
```diff
@@ -345,50 +345,176 @@ narrow down the type to a specific type:
...
+this example, so it's not recommended if you can avoid it:
+
+.. code-block:: python
+
+   def inc(x: int) -> int:
+       return x + 1
+
+   x = inc(None)  # No error reported by mypy!
+
+In mypy versions before 0.600 this was the default mode, and if you
```
gvanrossum: Move this up?

docs/source/revision_history.rst
```diff
@@ -225,7 +225,7 @@ List of major changes:
 - July 2016
     * Publish ``mypy-lang`` version 0.4.3 on PyPI.
 
-    * Add :ref:`strict_optional`.
+    * Add ``--strict-optional``.
```
gvanrossum: Add link to chapter.

mypy/main.py
```diff
@@ -424,8 +424,10 @@ def add_invertible_flag(flag: str,
                         help="include fine-grained dependency information in the cache")
     parser.add_argument('--skip-version-check', action='store_true',
                         help="allow using cache written by older mypy version")
-    add_invertible_flag('--strict-optional', default=False, strict_flag=True,
-                        help="enable experimental strict Optional checks")
+    parser.add_argument('--strict-optional', action='store_true',
+                        help=argparse.SUPPRESS)
+    parser.add_argument('--no-strict-optional', action='store_false', dest='strict_optional',
+                        help="disable strict Optional checks (inverse: --no-strict-optional)")
```
gvanrossum: I think the inverse should be `--strict-optional`? Also can't this use add_invertible_flag()?

runtests.py
```diff
@@ -274,16 +274,17 @@ def add_stdlibsamples(driver: Driver) -> None:
                 modules.append(module)
         if modules:
             driver.add_mypy_modules('stdlibsamples (%s)' % (version,), modules,
-                                    cwd=stdlibsamples_dir)
+                                    cwd=stdlibsamples_dir, extra_args=['--no-strict-optional'])
```
gvanrossum: Add the TODO here too?

PR #4977, opened 2018-04-26
Fine-grained: Trigger when abstract attributes change

JukkaL:
Add new kind of trigger `<ClsName.(abstract)>` which gets triggered when
the set of abstract attributes in a class change.

This is a more efficient alternative to #4922.

### DIFF
--- a/mypy/server/astdiff.py
+++ b/mypy/server/astdiff.py
@@ -211,6 +211,8 @@ def snapshot_definition(node: Optional[SymbolNode],
                  snapshot_optional_type(node._promote))
         prefix = node.fullname()
         symbol_table = snapshot_symbol_table(prefix, node.names)
+        # Special dependency for abstract attribute handling.
+        symbol_table['(abstract)'] = ('Abstract', tuple(sorted(node.abstract_attributes)))
         return ('TypeInfo', common, attrs, symbol_table)
     else:
         # Other node types are handled elsewhere.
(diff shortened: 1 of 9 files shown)


mypy/server/astdiff.py
```diff
@@ -211,6 +211,8 @@ def snapshot_definition(node: Optional[SymbolNode],
                  snapshot_optional_type(node._promote))
         prefix = node.fullname()
         symbol_table = snapshot_symbol_table(prefix, node.names)
+        # Special dependency for abstract attribute handling.
```
msullivan: I approve of this trick!

mypy/server/deps.py
```diff
@@ -254,6 +254,14 @@ def process_type_info(self, info: TypeInfo) -> None:
...
             self.add_dependency(make_trigger(base_info.fullname() + '.__new__'),
                                 target=make_trigger(info.fullname() + '.__new__'))
+            # If the set of abstract attributes change, this may invalidate class
+            # instantiation, or change the generated error message, since Python checks
+            # class abstract status when creating an instance.
+            self.add_dependency(make_trigger(base_info.fullname() + '.(abstract)'),
+                                target=make_trigger(info.fullname() + '.__init__'))
+            # If the base class abstract attributes change, subclass abstract
+            # attributes need to be recalculated.
+            self.add_dependency(make_trigger(base_info.fullname() + '.(abstract)'))
```
msullivan: Could these be more targeted? I think we shouldn't need dependencies from all the base classes abstract attrs to `__init__`, just from the class itself, and only if it is abstract (since abstract changes triggers the whole class)? And likewise I think the dependency from base abstract attrs is only needed if the base is abstract.

Maybe add a TODO?

Issue #4946, opened 2018-04-20
Release 0.600 planning


JukkaL:
The current plan is to release mypy 0.600 around Wed, May 2 (**UPDATED**). The release branch will be cut on Thursday, April 26.

I'd like to include at least these major changes in the release:

* Mypy daemon (as an experimental feature)
* Strict optional on by default
* Remote caching documented and officially supported

Mypy daemon support depends on Python 3 feature completeness for daemon.

ethanhs:
I'm most of the way through updating `typed_ast` to 3.7. I have everything up to type comment parsing, for which I have a bug to squash. Should I aim to get a new version in the next couple days? Or do we want to wait?

JukkaL:
If typed-ast 3.7 support is ready when we cut the release branch we can include it in the release. If the schedule is too tight, we can leave it until the following release which should be about 3 weeks from the the 0.600 release.

ilevkivskyi:
I would like to try landing protocols support before daemon is released. Although the PR itself is ready it requires `TypeInfo` mutable state refactoring and it needs to be rebased. I will make the refactoring PR today/tomorrow. Another question about the fine grained do we have an issue with a list of missing Python 3 features (or missing tests for those)? In general I would like the fine grained mode to be more like beta quality (known bugs), rather than alpha quality (missing features). Also what do you think about postponing other refactorings to the week after the release, and instead trying to fix some high-priority fine grained issues?

JukkaL:
@ilevkivskyi I'll prepare an issue with a list of Python 3 features that might need work later today. I have some time this week to work on Python 3 features. I agree that we want to at least know how well various Python 3 features work. Python 3 feature support has the highest priority since these will likely affect a lot of users (but not us so much since mypy doesn't use any recent Python 3 features).

Protocols support in the daemon would also be nice but we can also mention it as a known issue. Various planned refactorings don't need to be included in the release.

JukkaL:
Python 3 feature support in fine-grained incremental mode: #4951

ilevkivskyi:
> Various planned refactorings don't need to be included in the release.

The protocol support requires one of those (to avoid performance problems), `TypeInfo` state. This is however the smallest/simplest one and I already started working on it.

JukkaL:
Created the 0.600 release branch. Some documentation updates are still missing (at least mypy daemon) but they can be cherry picked. Protocols support in fine-grained mode didn't get in due to the tight schedule, unfortunately.

[...]

PR #1865, opened 2016-07-14
Fix importing new files in incremental mode

Michael0x2a:
This pull request makes mypy re-parse `__init__.py` files within modules when a new file is added to that module before incremental mode is run a second time.

Previously, when a file was modified to contain a new "from x import y" statement where "y" is the newly created submodule, mypy would assume that "y" was instead a new attribute added to the "x" module.

See #1848 for more details.

PR #4980, opened 2018-04-27
Fix fine-grained crash caused by unneeded extra parse_file()s

msullivan:
load_graph() contains code to parse ancestor modules that
get new children. Since load_graph() did not communicate that
these modules had been parsed to fine-grained mode, this caused
problems with fine-grained mode.

That code was added in #1865 to fix a bug where 'import from'
of a newly added submodule would not work in incremental mode
because it didn't appear in the parent module's symbol table.
The need to reload the parent to fix this was obviated by #2061,
which makes semantic analysis directly inspect the modules map
when doing 'import from' to find possible submodules.

Since the code isn't needed any more and it is causing trouble,
we just remove it.

I would like to cherry-pick this into 0.600 (#4946).

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1669,10 +1669,6 @@ class State:
     def is_interface_fresh(self) -> bool:
         return self.externally_same
 
-    def has_new_submodules(self) -> bool:
-        """Return if this module has new submodules after being loaded from a warm cache."""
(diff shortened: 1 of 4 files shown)


msullivan:
@Michael0x2a If you still have any context on this and it seems like I'm missing something, feel free to let me know

PR #4989, opened 2018-05-01
Reprocess decorators instead of the function they decorate

msullivan:
This fixes an issue where decorators on overloaded methods
would sometimes report false positives.


msullivan:
Oh, darn. I didn't fix the commit message.

PR #4993, opened 2018-05-02
Fix refresh of functions with implicit optional arguments

msullivan:
Type analysis would clear t.optional, so it wouldn't be there on
future runs.

This bug also apparently caused `__new__` methods to not pick up
Optional from implicit optional arguments?

### DIFF
--- a/test-data/unit/fine-grained.test
+++ b/test-data/unit/fine-grained.test
@@ -6310,3 +6310,15 @@ class Foo(a.I):
     def g(self) -> None: pass
 [out]
 ==
+
+[case testImplicitOptionalRefresh1]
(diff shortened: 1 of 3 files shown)


test-data/unit/fine-grained.test
```diff
@@ -6310,3 +6310,15 @@ class Foo(a.I):
     def g(self) -> None: pass
 [out]
 ==
+
+[case testImplicitOptionalRefresh1]
+# flags: --strict-optional
+from x import f
+def foo(x: int=None) -> None:
```
gvanrossum: Spaces around `=` here.

PR #4994, opened 2018-05-02
Fix some aststrip import bugs

msullivan:
Strip `assignments` in ImportAll nodes and clear `names` even
if there are assignments.

Diff looks bigger than it really is because some conditional code was made unconditional. Suggest ignoring whitespace changes when looking at it, which github just added a handy checkbox under "diff settings" to do!

### DIFF
--- a/mypy/server/aststrip.py
+++ b/mypy/server/aststrip.py
@@ -180,49 +180,44 @@ class NodeStripVisitor(TraverserVisitor):
                 self.process_lvalue_in_method(item)
 
     def visit_import_from(self, node: ImportFrom) -> None:
-        if node.assignments:
-            node.assignments = []
-        else:
-            # If the node is unreachable, don't reset entries: they point to something else!
-            if node.is_unreachable: return
-            if self.names:
-                # Reset entries in the symbol table. This is necessary since
-                # otherwise the semantic analyzer will think that the import
-                # assigns to an existing name instead of defining a new one.
-                for name, as_name in node.names:
-                    imported_name = as_name or name
-                    # This assert is safe since we check for self.names above.
-                    assert self.file_node is not None
-                    sym = create_indirect_imported_name(self.file_node,
-                                                        node.id,
-                                                        node.relative,
-                                                        name)
-                    if sym:
-                        self.names[imported_name] = sym
+        node.assignments = []
+
+        # If the node is unreachable, don't reset entries: they point to something else!
+        if node.is_unreachable: return
+        if self.names:
+            # Reset entries in the symbol table. This is necessary since
+            # otherwise the semantic analyzer will think that the import
+            # assigns to an existing name instead of defining a new one.
+            for name, as_name in node.names:
+                imported_name = as_name or name
+                # This assert is safe since we check for self.names above.
+                assert self.file_node is not None
+                sym = create_indirect_imported_name(self.file_node,
+                                                    node.id,
+                                                    node.relative,
+                                                    name)
+                if sym:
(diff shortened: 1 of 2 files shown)


test-data/unit/fine-grained-modules.test
```diff
@@ -1588,6 +1588,39 @@ x = ''
 ==
 ==
 
+[case testImportStarSomethingMoved1]
```
ilevkivskyi: `1` at the end of test case name implies you either forgot to add second tests or forgot to remove this `1`.

mypy/server/aststrip.py
```diff
@@ -233,6 +228,11 @@ def visit_import_all(self, node: ImportAll) -> None:
                 del self.names[name]
         node.imported_names = []
 
+    def visit_for_stmt(self, node: ForStmt) -> None:
```
ilevkivskyi: It looks like we don't have `visit_with_stmt` while it also might need to be stripped (at least for consistency).

test-data/unit/fine-grained-modules.test
```diff
@@ -1588,6 +1588,39 @@ x = ''
...
+[file p.py]
```
ilevkivskyi: You never import `p` so I am not sure anything is actually checked in this test.

mypy/server/aststrip.py
```diff
@@ -180,49 +180,44 @@ def process_lvalue_in_method(self, lvalue: Node) -> None:
...
+                # This assert is safe since we check for self.names above.
+                assert self.file_node is not None
+                sym = create_indirect_imported_name(self.file_node,
+                                                    node.id,
+                                                    node.relative,
+                                                    name)
+                if sym:
+                    self.names[imported_name] = sym
 
     def visit_import(self, node: Import) -> None:
```
ilevkivskyi: I see logically three changes, but only two tests added. Maybe there is missing a test with a simple import? (i.e. the one visited by this method)

mypy/server/aststrip.py
```diff
@@ -180,49 +180,44 @@ def process_lvalue_in_method(self, lvalue: Node) -> None:
...
-                                                        node.relative,
-                                                        name)
-                    if sym:
-                        self.names[imported_name] = sym
+        node.assignments = []
+
+        # If the node is unreachable, don't reset entries: they point to something else!
+        if node.is_unreachable: return
+        if self.names:
+            # Reset entries in the symbol table. This is necessary since
```
ilevkivskyi: Maybe explain in comment why we need to always do this independently of `assignments`?

test-data/unit/fine-grained-modules.test
msullivan: `testfinegrained` picks up all the files involved rather than following imports. I'll add an import though to make it more clear, since most tests still do that.

mypy/server/aststrip.py
msullivan: The change was for consistency, but it turns out that `Import` can't even have `assignments`. Changing it to an assert instead.

mypy/server/aststrip.py
msullivan: Hm probably `target_type` needs to get cleared. I'll look at it for a different PR

test-data/unit/fine-grained-modules.test
msullivan: It was the latter, though I do also kind of like always having numbers on tests, because it means you don't need to rename the original one to add a second.

PR #5003, opened 2018-05-04
Create socket file in /tmp

myjian:
When the mypy daemon runs in Windows Subsystem for Linux (WSL) under a [DriveFs](https://blogs.msdn.microsoft.com/wsl/2016/04/22/windows-subsystem-for-linux-overview/) directory, it would fail to bind to the socket file.

From mypy/dmypy_server.py:
```
sock = socket.socket(socket.AF_UNIX)
sock.bind(self.sockname)  # <-- stuck here
sock.listen(1)
```

This pull request changes the path to `/tmp`, so on Windows it will be always created under the `VolFs`, which provides full support for Linux file system features.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -80,7 +80,7 @@ def daemonize(func: Callable[[], None], log_file: Optional[str] = None) -> int:
 
 # Server code.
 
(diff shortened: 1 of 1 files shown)


msullivan:
Thanks!

Issue #5002, opened 2018-05-03
Write tests for cache performance in fine-grained incremental mode


ilevkivskyi:
We don't have good tests for end-to-end cache performance. It would be good to add tests that check that cache files written in a normal run with `--cache-fine-grained` are then successfully loaded in a subsequent daemon cold start. The emphasis here is on end-to-end, we have tests that use cache but they don't exactly emulate what happens "in real world". An additional argument for this is that code in `build.py` responsible for cache loading is quite complicated, so it is better to be sure that we don't break it.

ilevkivskyi:
More ideas for cache tests: cache corruption tests, see https://github.com/python/mypy/pull/5004#discussion_r186033289

PR #5004, opened 2018-05-04
Fallback to full run if protocol deps are missing/invalid instead of crashing/failing

ilevkivskyi:
Also change the cache file names to `@proto_deps` to avoid clashes with user modules with the same name.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2348,8 +2348,7 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
             manager.cache_enabled = False
             TypeState.proto_deps = {}
             manager.log("Falling back to full run -- reloading graph...")
-            graph = load_graph(sources, manager)
-            process_graph(graph, manager)
+            return dispatch(sources, manager)
 
     else:
         process_graph(graph, manager)
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1088,7 +1088,7 @@ def get_protocol_deps_cache_name(manager: BuildManager) -> Tuple[str, str]:
     contains hashes of all source files at the time the protocol dependencies
     were written, and data file contains the protocol dependencies.
     """
-    name = os.path.join(_cache_dir_prefix(manager), 'proto_deps')
+    name = os.path.join(_cache_dir_prefix(manager), '@proto_deps')
     return name + '.meta.json', name + '.data.json'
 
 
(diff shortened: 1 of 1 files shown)


mypy/build.py
```diff
@@ -2342,11 +2342,15 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
         # after we loaded cache for whole graph. The `read_protocol_cache` will also validate
         # the protocol cache against the loaded individual cache files.
         TypeState.proto_deps = read_protocol_cache(manager, graph)
-        if TypeState.proto_deps is None and manager.stats.get('fresh_trees', 0) > 0:
+        if TypeState.proto_deps is None and manager.stats.get('fresh_metas', 0) > 0:
```
JukkaL: It would be good to have a test case for this, since this is something that could easily get stale. Maybe create an issue about adding a test (maybe also for other related cache corruption issues)?

mypy/build.py
```diff
@@ -2342,11 +2342,15 @@ def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
...
             # There were some cache files read, but no protocol dependencies loaded.
-            manager.errors.set_file(_cache_dir_prefix(manager), None)
-            manager.errors.report(0, 0, "Error reading protocol dependencies cache -- aborting",
-                                  blocker=True)
+            manager.log("Error reading protocol dependencies cache -- aborting cache load")
+            manager.cache_enabled = False
+            TypeState.proto_deps = {}
+            manager.log("Falling back to full run -- reloading graph...")
+            graph = load_graph(sources, manager)
```
JukkaL: Should we call `update_protocol_deps` as well? This might be less likely to get stale if we'd only have one `process_graph` callin this function. Maybe instead of calling `load_graph` and `process_graph` here, call `dispatch` recursively after changing the options?

mypy/build.py
ilevkivskyi: We already have https://github.com/python/mypy/issues/5002, do we need a separate issue? I can just add a comment that we need to also check performance/behavior with corrupted caches.

mypy/build.py
ilevkivskyi: > Should we call `update_protocol_deps` as well?

I think no, since `update.py` should do this. We call it manually below to have them up to date if we write caches. Anyway, I think calling `dispatch` recursively is a better option, so we don't need to worry about keeping the logic in sync.

PR #5005, opened 2018-05-04
Support factory= in attr plugin.

euresti:
This is a new feature in attrs 18.1 
The implementation is just like `default=`

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -323,7 +323,7 @@ def _attribute_from_attrib_maker(ctx: 'mypy.plugin.ClassDefContext',
     if attr_has_default and attr_has_factory:
         ctx.api.fail("Can't pass both `default` and `factory`.", rvalue)
     elif attr_has_factory:
-        attr_has_default = attr_has_factory
+        attr_has_default = True
 
     # If the type isn't set through annotation but is passed through `type=` use that.
     type_arg = _get_argument(rvalue, 'type')
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -318,6 +318,12 @@ def _attribute_from_attrib_maker(ctx: 'mypy.plugin.ClassDefContext',
     init = _get_bool_argument(ctx, rvalue, 'init', True)
     # TODO: Check for attr.NOTHING
     attr_has_default = bool(_get_argument(rvalue, 'default'))
+    attr_has_factory = bool(_get_argument(rvalue, 'factory'))
+
+    if attr_has_default and attr_has_factory:
+        ctx.api.fail("Can't pass both `default` and `factory`.", rvalue)
+    elif attr_has_factory:
+        attr_has_default = attr_has_factory
 
     # If the type isn't set through annotation but is passed through `type=` use that.
     type_arg = _get_argument(rvalue, 'type')
--- a/test-data/unit/check-attr.test
+++ b/test-data/unit/check-attr.test
@@ -771,3 +771,15 @@ class FFrozen(F):
     def bar(self) -> bool:
         return self._cb(5, 6)
 [builtins fixtures/callable.pyi]
+
+[case testAttrsWithFactory]
+from typing import List
+import attr
+def my_factory() -> int:
+    return 7
+@attr.s
+class A:
+    x: List[int] = attr.ib(factory=list)
+    y: int = attr.ib(factory=my_factory)
+A()
+[builtins fixtures/list.pyi]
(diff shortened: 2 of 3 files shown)


mypy/plugins/attrs.py
```diff
@@ -318,6 +318,12 @@ def _attribute_from_attrib_maker(ctx: 'mypy.plugin.ClassDefContext',
     init = _get_bool_argument(ctx, rvalue, 'init', True)
     # TODO: Check for attr.NOTHING
     attr_has_default = bool(_get_argument(rvalue, 'default'))
+    attr_has_factory = bool(_get_argument(rvalue, 'factory'))
+
+    if attr_has_default and attr_has_factory:
+        ctx.api.fail("Can't pass both `default` and `factory`.", rvalue)
```
ilevkivskyi: It is a good idea to always add tests for any newly added error messages.

test-data/unit/check-attr.test
```diff
@@ -771,3 +771,15 @@ class FFrozen(F):
...
+
+[case testAttrsWithFactory]
+from typing import List
+import attr
+def my_factory() -> int:
+    return 7
+@attr.s
+class A:
+    x: List[int] = attr.ib(factory=list)
+    y: int = attr.ib(factory=my_factory)
```
ilevkivskyi: Please add also a test where this fails due to a wrong return type of the factory function.

mypy/plugins/attrs.py
```diff
@@ -318,6 +318,12 @@ def _attribute_from_attrib_maker(ctx: 'mypy.plugin.ClassDefContext',
...
+    elif attr_has_factory:
+        attr_has_default = attr_has_factory
```
ilevkivskyi: I think `attr_has_default = True` would be shorter and cleaner.

euresti:
Done and done!

test-data/unit/lib-stub/attr.pyi
```diff
@@ -1,57 +1,98 @@
...
+           hash: Optional[bool] = ...,
+           init: bool = ...,
+           convert: Optional[_ConverterType[_T]] = ...,
+           metadata: Optional[Mapping[Any, Any]] = ...,
+           type: Optional[Type[_T]] = ...,
+           converter: Optional[_ConverterType[_T]] = ...,
+           factory: Optional[Callable[[], _T]] = ...,
+           ) -> _T: ...
+# This form catches explicit None or no default but with no other arguments returns Any.
+# This form catches explicit None or no default but with no other arguments returns Any.
```
ilevkivskyi: This line is repeated twice.

test-data/unit/lib-stub/attr.pyi
```diff
@@ -1,57 +1,98 @@
-from typing import TypeVar, overload, Callable, Any, Type, Optional
+from typing import TypeVar, overload, Callable, Any, Type, Optional, Union, Sequence, Mapping
 
 _T = TypeVar('_T')
 _C = TypeVar('_C', bound=type)
 
+_ValidatorType = Callable[[Any, Any, _T], Any]
```
ilevkivskyi: I thought validator should return `bool`, isn't this the case?

test-data/unit/check-attr.test
```diff
@@ -771,3 +771,29 @@ class FFrozen(F):
...
+@attr.s
+class A:
+    x: int = attr.ib(factory=int, default=7)  # E: Can't pass both `default` and `factory`.
+[builtins fixtures/bool.pyi]
+
+[case testAttrsFactoryBadReturn]
+import attr
+@attr.s
+class A:
+    x: int = attr.ib(factory=list)  # E: Incompatible types in assignment (expression has type "List[T]", variable has type "int")
```
ilevkivskyi: The error message looks a bit cryptic. Why it is `List[T]` and not `List[Any]`? There is no `T` in code. I suppose this might be a bug in using class as factory, is the error better if you use a generic function instead of class?

test-data/unit/lib-stub/attr.pyi
euresti: Oops.

test-data/unit/lib-stub/attr.pyi
euresti: The return value of validators is ignored.  A validation error is a raised exception.   However I don't want to force a "None"

test-data/unit/check-attr.test
euresti: This error come from the stubs not from the plugin (i.e. any method that takes a type would give this error message.)  I can add a generic function too.

euresti:
Thanks for the quick review. Your requests have been implemented.

PR #5009, opened 2018-05-06
Fix access to operators on metaclass through typevar

### LINKED ISSUE #4929: Metaclass operators not respected by TypeVar bounds
Using `mypy 0.590`, running `mypy` on the following code:

(EDIT: See https://github.com/python/mypy/issues/4929#issuecomment-383166651 for a smaller test case)

```python
import ctypes
from typing import Type, TypeVar

S = TypeVar("S", bound=ctypes.Structure)

def f(x: Type[ctypes.Structure]) -> ctypes.Array[ctypes.Structure]:
    return (x * 0)()
def g(x: Type[S]) -> ctypes.Array[S]:
    return (x * 0)()
```

fails with

```
test.py:9: error: Unsupported operand types for * ("Type[S]" and "int")
test.py:9: error: "int" not callable
```

But AFAICT, `g` should be perfectly valid -- because it has a bound of `ctypes.Structure` (which supports multiplication with integers), it should also support multiplication with integers.

elazarg:
Fix #4929: typevars were not handled correctly on `has_member`.

The test uses multiplication instead of addition since adding `__radd__` to the stub reveals several inconsistencies in unrelated tests.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2431,6 +2431,10 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         """Does type have member with the given name?"""
         # TODO: refactor this to use checkmember.analyze_member_access, otherwise
         # these two should be carefully kept in sync.
+        if isinstance(typ, TypeVarType):
+            typ = typ.upper_bound
+        if isinstance(typ, TupleType):
+            typ = typ.fallback
         if isinstance(typ, Instance):
             return typ.type.has_readable_member(member)
         if isinstance(typ, CallableType) and typ.is_type_obj():
@@ -2440,8 +2444,6 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         elif isinstance(typ, UnionType):
             result = all(self.has_member(x, member) for x in typ.relevant_items())
             return result
-        elif isinstance(typ, TupleType):
-            return self.has_member(typ.fallback, member)
         elif isinstance(typ, TypeType):
             # Type[Union[X, ...]] is always normalized to Union[Type[X], ...],
             # so we don't need to care about unions here.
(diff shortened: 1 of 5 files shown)


mypy/checkexpr.py
```diff
@@ -2445,9 +2445,14 @@ def has_member(self, typ: Type, member: str) -> bool:
         elif isinstance(typ, TypeType):
             # Type[Union[X, ...]] is always normalized to Union[Type[X], ...],
             # so we don't need to care about unions here.
-            if isinstance(typ.item, Instance) and typ.item.type.metaclass_type is not None:
-                return self.has_member(typ.item.type.metaclass_type, member)
-            if isinstance(typ.item, AnyType):
+            item = typ.item
+            if isinstance(item, TypeVarType):
```
ilevkivskyi: I wonder shouldn't we add the same fallback to upper bound at the very top of this function? For example if I have just
```python
def f(x: T) -> T:
    x + 0
```

test-data/unit/check-classes.test
```diff
@@ -3459,6 +3459,23 @@ class Concrete(metaclass=Meta):
...
+    def __mul__(cls, other: int) -> str:
+        return ""
+
+class Test(metaclass=MetaClass):
+    pass
+
+S = TypeVar("S", bound=Test)
+
+def f(x: Type[Test]) -> str:
+    return x * 0
```
ilevkivskyi: This will still pass if the inferred type will become `Any`. I would either add an explicit `reveal_type` or add `--warn-return-any` flag in this test.

mypy/checkexpr.py
```diff
@@ -2431,6 +2431,10 @@ def has_member(self, typ: Type, member: str) -> bool:
         """Does type have member with the given name?"""
         # TODO: refactor this to use checkmember.analyze_member_access, otherwise
         # these two should be carefully kept in sync.
+        if isinstance(typ, TypeVarType):
+            typ = typ.upper_bound
```
ilevkivskyi: Could you also add a test case also for this branch? (Similar to the one you already have.)

Issue #4792, opened 2018-03-26
Add support for dataclasses


gwk:
I have added preliminary support for dataclasses in https://github.com/python/typeshed/pull/1944.

These seem to work for declarations, but not for instantiation. For example:

```
@dataclass
class D0:
  x: int

d0 = D0(x=0)
```

The constructor is erroneously flagged, and points to the `__init__` annotation for `object`:
```
error: Unexpected keyword argument "x" for "D0"
/Users/gwk/work/python/mypy/typeshed/stdlib/3/builtins.pyi:38:5: note: "D0" defined here
```

It appears that we need to teach mypy about dataclasses more deeply.

JelleZijlstra:
Yes, we will definitely need a mypy plugin to support dataclasses. It can probably be somewhat similar to the existing one for attrs.

ilevkivskyi:
It is good that you opened this issue, we definitely wanted to have a dedicated support for this, so now we have a place to track it. (Btw, I wanted to do this.)

gwk:
I'm not in the position to develop this plugin, so by all means give it a whirl if you want. I am actively using dataclasses though so I can help with testing.

gwk:
One other point: I backported dataclasses to 3.6 as a third party module inside of a general-purpose utility package that I maintain. I wonder, will it be possible/easy to get this dataclass plugin to recognize that module for < 3.7?

ilevkivskyi:
Yes.

PR #5010, opened 2018-05-06
add support for dataclasses

Bogdanp:
This is a work in progress for adding support for dataclasses as a plugin (heavily inspired by and reusing some parts of the attrs plugin).  It's pretty close to complete, but I wanted to get some feedback before continuing as I'm new to the codebase.  I'll take this over the finish line next weekend if it's deemed desirable, but here are my open questions:

* Is the way I'm hooking the plugin in correct? `fullname` in `get_class_decorator_hook` doesn't seem to have the value I expect (i.e. the name is qualified to the module `dataclass` is imported in, so its value is `some_package.some_module.dataclass` rather than `dataclasses.dataclass`).
* Should I be attempting to re-use more code between the two plugins? I only re-used the things that seemed to be fully generic between the two.

I tried it out on a real-world Python 3.6 project that makes heavy use of dataclasses and it seems to work great so far.

This would close https://github.com/python/mypy/issues/4792.


mypy/plugin.py
```diff
@@ -302,13 +301,23 @@ def get_method_hook(self, fullname: str
...
+
+        if fullname in attrs.attr_class_makers:
+            return attrs.attr_class_maker_callback
+        elif fullname in attrs.attr_dataclass_makers:
             return partial(
-                mypy.plugins.attrs.attr_class_maker_callback,
+                attrs.attr_class_maker_callback,
                 auto_attribs_default=True
             )
+        # TODO: It's unclear if fullname should be a fully qualified
```
euresti: I believe this happens if the module is not in the typeshed.

Bogdanp: Thanks! Should I add the module to the typeshed first?

ilevkivskyi: I am fine with this solution. We can check later if this is fixed when the stub is in typeshed.

JelleZijlstra: There is already an open typeshed PR to add a dataclasses stub: https://github.com/python/typeshed/pull/1944. I found some issues in it, but the original author hasn't responded for a while. @Bogdanp if you're interested, you could take over that PR and I can help you get it merged into typeshed.

test-data/unit/lib-stub/dataclasses.pyi
```diff
@@ -0,0 +1,36 @@
...
+              *,
+              init: bool = ...,
+              repr: bool = ...,
+              eq: bool = ...,
+              order: bool = ...,
+              unsafe_hash: bool = ...,
+              frozen: bool = ...) -> Callable[[_C], _C]: ...
+
+
+def field(*,
```
ilevkivskyi: It is good that you added this to `lib-stub`. Be sure to make updates here, if https://github.com/python/typeshed/pull/1944 gets updated.

test-data/unit/check-dataclasses.test
```diff
@@ -0,0 +1,108 @@
...
+
+
+[case testDataclassesFields]
+# flags: --python-version 3.6
+from dataclasses import dataclass, field
+
+@dataclass
+class Person:
+    name: str
+    age: int = field(default=0, init=False)
```
ilevkivskyi: I would add more combinations of different values for `init` and present/absent defaults.

test-data/unit/check-dataclasses.test
```diff
@@ -0,0 +1,108 @@
...
+
+reveal_type(Person)  # E: Revealed type is 'def (name: builtins.str) -> __main__.Person'
+john = Person('John')
+john.age = 24
```
ilevkivskyi: I would move this below the next line to check the type is inferred from class, not from this assignment.

test-data/unit/check-dataclasses.test
```diff
@@ -0,0 +1,108 @@
...
+
+[case testDataclassesDefaults]
+# flags: --python-version 3.6
+from dataclasses import dataclass
+
+
+@dataclass
+class Application:
+    name: str = 'Unnamed'
+    rating: int = 0
```
ilevkivskyi: I would add a test that an error is given if a field without a default comes after the one with a default.

ilevkivskyi:
@Bogdanp are you still interested in working on this? If you don't have time or desire, then we can pick up the PR.

mypy/plugins/common.py
```diff
@@ -0,0 +1,69 @@
+from typing import Optional
+
+from mypy.nodes import CallExpr, Expression, FuncBase, RefExpr, Var
+from mypy.plugin import ClassDefContext
+from mypy.types import CallableType, Overloaded
+
+
+def _get_decorator_bool_argument(
```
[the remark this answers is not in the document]
Bogdanp: Yes, these are the same functions from the attrs module, unchanged.

mypy/plugin.py
```diff
@@ -302,13 +301,23 @@ def get_method_hook(self, fullname: str
 
     def get_class_decorator_hook(self, fullname: str
                                  ) -> Optional[Callable[[ClassDefContext], None]]:
-        if fullname in mypy.plugins.attrs.attr_class_makers:
-            return mypy.plugins.attrs.attr_class_maker_callback
-        elif fullname in mypy.plugins.attrs.attr_dataclass_makers:
+        from mypy.plugins import attrs
```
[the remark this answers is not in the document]
Bogdanp: There had already been a circular dependency that was being resolved by [this](https://github.com/python/mypy/pull/5010/files#diff-fafb55a715f1c5c1f6f4cbc247af56f9L5) line. When I extracted `mypy.plugins.common` the same thing continued to work fine for the test suite, but the dynamically-generated test to ensure `import mypy.plugins.common` could be imported failed. I think the circular dependency can be fixed by extracting `ClassDefContext` into a separate module (like `mypy.plugin_context`). Let me know if you'd like me to do that!

Bogdanp:
@ilevkivskyi I believe I have addressed all the points you raised except for

* the circular import (see [this comment](https://github.com/python/mypy/pull/5010#discussion_r189455379)) and
* adding support for `InitVar`s. 

I'm a bit out of my depth when it comes to adding support for `InitVars`. I could not find how to add a new generic alias type (which I think `InitVars` counts as). Help would be appreciated here or I would be happy to let someone more experienced implement that part.

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,279 @@
...
+                    return_type=order_return_type,
+                    self_type=order_other_type,
+                    tvar_def=order_tvar_def,
+                )
+
+        if decorator_arguments['frozen']:
+            self._freeze(attributes)
+
+        info.metadata['dataclass'] = {
+            'attributes': {attr.name: attr.serialize() for attr in attributes},
```
JelleZijlstra: This probably should be an OrderedDict in order to preserve order in pre-3.6 Python versions. I think that's what's causing the test failure on 3.4 and 3.5 in Travis.

Bogdanp: Ahh, that's right! I'll make the change.

Bogdanp: Change made. :)

JelleZijlstra: Thanks! For the future though, it's generally easier on the reviewer to push new commits instead of amending your previous commit and force-pushing as you're doing now. That way, it's easier for the reviewer to see what changes were made in response to the review.

We squash the commits anyway before merging, so it's OK to have a lot of commits in your PR branch.

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,279 @@
...
+        if decorator_arguments['eq'] and info.get('__eq__') is None:
+            cmp_tvar_def = TypeVarDef('T', 'T', 1, [], ctx.api.named_type('__builtins__.object'))
+            cmp_other_type = TypeVarType(cmp_tvar_def)
+            cmp_return_type = ctx.api.named_type('__builtins__.bool')
+
+            for method_name in ['__eq__', '__ne__']:
+                _add_method(
+                    ctx,
+                    method_name,
+                    args=[Argument(Var('other', cmp_other_type), cmp_other_type, None, ARG_POS)],
```
ilevkivskyi: I think it is dangerous to reuse the same type `cmp_other_type` for two different methods. (attrs plugin also does this, we can fix this separately).

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,279 @@
...
+                    return_type=cmp_return_type,
+                    self_type=cmp_other_type,
+                    tvar_def=cmp_tvar_def,
+                )
+
+        # Add <,>,<=,>=, but only if the class has an eq method.
```
ilevkivskyi: Please alway use space after comma.

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,279 @@
...
+
+            try:
+                node = cls.info.names[lhs.name].node
+                assert isinstance(node, Var)
+
+                # x: ClassVar[int] is ignored by dataclasses.
+                if node.is_classvar:
+                    continue
+
+                # Treat the assignment as an instance-level assignment.
```
ilevkivskyi: Why do we need this? Could you please add more details?

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,279 @@
...
+                node.is_initialized_in_class = False
+            except KeyError:
```
ilevkivskyi: Add a comment when this can happen.

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,279 @@
...
+                continue
+
+            has_field_call, field_args = _collect_field_args(stmt.rvalue)
+
+            try:
+                is_in_init = bool(ctx.api.parse_bool(field_args['init']))
+            except KeyError:
```
ilevkivskyi: There are too many things inside this except block. Move all the unnecessary code outside. Also I would probably use `if 'init' in field_args` here.

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,279 @@
...
+                    known_attrs.add(name)
+                    super_attrs.append(attr)
+
+        all_attrs = super_attrs + attrs
+
+        # Ensure that arguments without a default don't follow
+        # arguments that have a default.
+        found_default = False
+        for attr in all_attrs:
+            if found_default and attr.is_in_init and not attr.has_default:
```
ilevkivskyi: Please clarify in above comment why `and attr.is_in_init` is important.

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,279 @@
...
+        """Converts all attributes to @property methods in order to
+        emulate frozen classes.
+        """
+        info = self._ctx.cls.info
+        for attr in attributes:
+            try:
+                node = info.names[attr.name].node
+                assert isinstance(node, Var)
+                node.is_property = True
+            except KeyError:
```
ilevkivskyi: TBH, I am a big fan of having so many try/except blocks. Could you refactor t least some of them?

test-data/unit/check-dataclasses.test
ilevkivskyi: Have you added these tests?

test-data/unit/check-dataclasses.test
```diff
@@ -0,0 +1,362 @@
...
+[case testDataclassesDeepInheritance]
+# flags: --python-version 3.6
+from dataclasses import dataclass
+
+
+@dataclass
+class A:
+    a: int
+
+
```
ilevkivskyi: Please use at most one empty line in tests (even between tests).

test-data/unit/check-dataclasses.test
```diff
@@ -0,0 +1,362 @@
...
+@dataclass
+class SpecializedApplication(Application):
+  rating: int = 0
+
+
+reveal_type(SpecializedApplication)  # E: Revealed type is 'def (id: Union[builtins.int, None], name: builtins.str, rating: builtins.int =) -> __main__.SpecializedApplication'
+
+[builtins fixtures/list.pyi]
```
ilevkivskyi: Newline is missing at the end of file.

test-data/unit/check-dataclasses.test
```diff
@@ -0,0 +1,108 @@
...
+class Application:
+    name: str
+
+    @classmethod
+    def parse(cls, request: str) -> "Application":
+        return cls(name='...')
+
+app = Application.parse('')
+
+[builtins fixtures/list.pyi]
```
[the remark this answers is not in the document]
ilevkivskyi: It looks like you didn't yet added incremental tests. They live in `check-incremental.test` (you can look for example in `attrs` PR).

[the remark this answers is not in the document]
ilevkivskyi: Could you please add some tests for generic dataclasses, in particular type inference (again, you can look for examples in the original `attrs` PR)

test-data/unit/lib-stub/dataclasses.pyi
```diff
@@ -0,0 +1,36 @@
+from typing import Any, Callable, Optional, TypeVar, overload
+
+_T = TypeVar('_T')
+_C = TypeVar('_C', bound=type)
+
+
+@overload
+def dataclass(_cls: _C,
```
ilevkivskyi: You also need to add `InitVar` (here and on typeshed). You can define it as a normal generic class:
```python
class InitVar(Generic[_T]):
    pass
```
Then in the plugin you can test for it e.g. like this `if isinstance(some_type, Instance) and some_type.info.fullname() == 'dataclasses.InitVar'`.

test-data/unit/check-dataclasses.test
Bogdanp: I have now

mypy/plugin.py
```diff
@@ -302,13 +301,19 @@ def get_method_hook(self, fullname: str
...
+
+        if fullname in attrs.attr_class_makers:
+            return attrs.attr_class_maker_callback
+        elif fullname in attrs.attr_dataclass_makers:
             return partial(
-                mypy.plugins.attrs.attr_class_maker_callback,
+                attrs.attr_class_maker_callback,
                 auto_attribs_default=True
             )
+        # TODO: Drop the or clause once dataclasses lands in typeshed.
```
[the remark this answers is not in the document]
Bogdanp: Sorry, but I don't think I'll have the time to carry that through at the moment.

Bogdanp:
@ilevkivskyi do you have any advice with regards to how to best remove `InitVar` attributes from a class? I've tried removing them from the SymbolTable, but that breaks inheritance. The best I could come up with was to remove them from the class' symbol table and add the types to the `DataclassAttribute`s but then I wasn't sure how to best serialize/deserialize the types.

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,295 @@
...
+        return Var(self.name, info[self.name].type)
+
+    def serialize(self) -> JsonDict:
+        return {
+            'name': self.name,
+            'is_in_init': self.is_in_init,
+            'has_default': self.has_default,
+            'line': self.line,
+            'column': self.column,
+        }
```
ilevkivskyi: > The best I could come up with was to remove them from the class' symbol table and add the types to the DataclassAttributes but then I wasn't sure how to best serialize/deserialize the types.

You can add list of `InitVar`s here. Most types in mypy support (de-)serialization (those that are not supported like partial types, or forward reference types, should not appear in at the point where serialization happens). You can find signatures in `mypy.types.Type.serialize` and `mypy.types.Type.deserialize`.

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,295 @@
...
+        return all_attrs
+
+    def _freeze(self, attributes: List[DataclassAttribute]) -> None:
+        """Converts all attributes to @property methods in order to
+        emulate frozen classes.
+        """
+        info = self._ctx.cls.info
+        for attr in attributes:
+            try:
+                node = info.names[attr.name].node
```
ilevkivskyi: I think this will look cleaner if refactored via `info.names.get(attr.name)` which cannot raise.

setup.cfg
```diff
@@ -48,3 +48,6 @@ parallel = true
 
 [coverage:report]
 show_missing = true
+
+[isort]
+multi_line_output = 5
```
ilevkivskyi: Why do you want this addition? Can it be made in a separate PR?

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,295 @@
...
+            # Treat the assignment as an instance-level assignment
+            # even though it looks like a class-level assignment.
+            node.is_initialized_in_class = False
+
+            has_field_call, field_args = _collect_field_args(stmt.rvalue)
+
+            try:
+                # parse_bool returns an optional bool, so we corece it
+                # to a bool here in order to appease the type checker.
+                is_in_init = bool(ctx.api.parse_bool(field_args['init']))
```
ilevkivskyi: This will be definitely cleaner if refactored via `filed_args.get('init')` that can't raise. (Also the comment/coercion will be unnecessary after refactoring.)

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,295 @@
...
+            node = cls.info.names[lhs.name].node
+            assert isinstance(node, Var)
+
+            # x: ClassVar[int] is ignored by dataclasses.
+            if node.is_classvar:
+                continue
+
+            # Treat the assignment as an instance-level assignment
+            # even though it looks like a class-level assignment.
+            node.is_initialized_in_class = False
```
ilevkivskyi: Could you please add a test that fails without this change? Or if you already added, could you please tell me the name?

test-data/unit/check-dataclasses.test
```diff
@@ -0,0 +1,384 @@
...
+  def foo(self) -> List[T]:
+    return [self.x, self.y]
+
+  def bar(self) -> T:
+    return self.z[0]
+
+  def problem(self) -> T:
+    return self.z  # E: Incompatible return value type (got "List[T]", expected "T")
+
+reveal_type(A)  # E: Revealed type is 'def [T] (x: T`1, y: T`1, z: builtins.list[T`1]) -> __main__.A[T`1]'
```
ilevkivskyi: This check is not very useful, I would rather called it with incompatible args to see an error, e.g. `A('no', 'way', [1, 2])`.

test-data/unit/check-dataclasses.test
```diff
@@ -0,0 +1,384 @@
...
+a = A(1, 2, [1, 2])
+reveal_type(a)  # E: Revealed type is '__main__.A[builtins.int*]'
+reveal_type(a.x)  # E: Revealed type is 'builtins.int*'
+reveal_type(a.y)  # E: Revealed type is 'builtins.int*'
+reveal_type(a.z)  # E: Revealed type is 'builtins.list[builtins.int*]'
```
ilevkivskyi: I would add a test that `a.foo()` returns an `int`.

test-data/unit/check-dataclasses.test
```diff
@@ -0,0 +1,384 @@
...
+
+@dataclass
+class SpecializedApplication(Application):
+  rating: int = 0
+
+reveal_type(SpecializedApplication)  # E: Revealed type is 'def (id: Union[builtins.int, None], name: builtins.str, rating: builtins.int =) -> __main__.SpecializedApplication'
+
+[builtins fixtures/list.pyi]
+
+[case testDataclassGenerics]
```
ilevkivskyi: Thanks for adding some tests!

setup.cfg
Bogdanp: That was accidental, sorry! I'll remove it.

mypy/plugins/dataclasses.py
Bogdanp: I may be missing something but I don't think that's true. `field_args` is a dict from `str` -> `Expression` so I'll still need the `parse_bool` call, which means I'll also need the `bool` coercion since `parse_bool` returns an `Optional[bool]`. I'll refactor to drop the try-catch though (although I do think this approach is cleaner)!

mypy/plugins/dataclasses.py
Bogdanp: I had tried that, but deserialization failed with `De-serialization failure: TypeInfo not fixed`. Reading the docstring for `FakeInfo`, I understand this is supposed to be done in two passes, but it looks like the fixup step requires a lot of information I don't necessarily have/that it would be ugly to pass around. Am I missing something obvious?

mypy/plugins/dataclasses.py
Bogdanp: It appears that wasn't necessary after all so I removed it.

Bogdanp:
Thanks for the review @ilevkivskyi. I believe I've made all the requested changes apart from `InitVar` support which I'll try to tackle again towards the end of the week.

mypy/plugin.py
ilevkivskyi: The typeshed PR is landed. Is this still needed?

mypy/plugins/dataclasses.py
```diff
@@ -0,0 +1,291 @@
...
+        return Var(self.name, info[self.name].type)
+
+    def serialize(self) -> JsonDict:
+        return {
+            'name': self.name,
+            'is_in_init': self.is_in_init,
+            'has_default': self.has_default,
+            'line': self.line,
+            'column': self.column,
+        }
```
ilevkivskyi: > Am I missing something obvious?

No, you are right, this way it will not work. Here is another idea about `InitVar`s that I think should work. You can just store _names_ of `InitVar`s in a special filed in `.metadata`, then later you can find the types in the type of `__init__` method, since in mypy both arg types and and arg names are stored. This is a bit hacky, but should be robust and is much simpler than manually traversing types in `fixup` phase.

test-data/unit/lib-stub/dataclasses.pyi
```diff
@@ -0,0 +1,39 @@
...
+              *,
+              init: bool = ...,
+              repr: bool = ...,
+              eq: bool = ...,
+              order: bool = ...,
+              unsafe_hash: bool = ...,
+              frozen: bool = ...) -> Callable[[_C], _C]: ...
+
+
+def field(*,
```
ilevkivskyi: It looks like current definition in typeshed is a bit different, could you please sync this one?

test-data/unit/check-incremental.test
```diff
@@ -4362,6 +4362,297 @@ import b
...
+    a: int
+    _b: List[int]
+    c: str = '18'
+    _d: int = field(default=False)
+    E = 7
+    F: ClassVar[int] = 22
+
+[builtins fixtures/list.pyi]
+[out1]
+[out2]
```
ilevkivskyi: I think this test is not testing anything. If all files are unchanged like here, then mypy will at most load caches, but will not make any checks. What you should do is to put first part of the test in another module (e.g. `[file b.py]`). And then on the second run (i.e. in `[file b.py.2]`) just make a cosmetic change, like  comment somewhere, so that mypy rechecks `b.py` while using cache for (unchanged) `a.py`.

test-data/unit/check-incremental.test
```diff
@@ -4362,6 +4362,297 @@ import b
...
+[file a.py]
+from dataclasses import dataclass
+
+@dataclass
+class A:
+    x: int
+
+[builtins fixtures/list.pyi]
+[out1]
+main:7: error: Revealed type is 'def (x: builtins.int) -> __main__.B'
```
ilevkivskyi: Any error (including a reveal) prevents writing the cache, so you should use `reveal_type` only on second run.

test-data/unit/check-incremental.test
```diff
@@ -4362,6 +4362,297 @@ import b
...
+
+[builtins fixtures/list.pyi]
+[rechecked]
+[stale]
+[out1]
+main:4: error: Property "x" defined in "Frozen" is read-only
+main:13: error: Unsupported left operand type for < ("NoCmp")
+main:14: error: Unsupported left operand type for <= ("NoCmp")
+main:15: error: Unsupported left operand type for > ("NoCmp")
+main:16: error: Unsupported left operand type for >= ("NoCmp")
```
ilevkivskyi: The above comment applies here too, I would like the first run to be clean, and the second run to have errors.

test-data/unit/check-incremental.test
```diff
@@ -4362,6 +4362,297 @@ import b
...
+class NoInit:
+    x: int
+
+@dataclass(order=False)
+class NoCmp:
+    x: int
+
+[builtins fixtures/list.pyi]
+[rechecked]
+[stale]
```
ilevkivskyi: You don't need these `[stale]` and `[rechecked]` in your tests, since your PR doesn't change dependency tracking logic.

test-data/unit/check-incremental.test
```diff
@@ -4362,6 +4362,297 @@ import b
...
+[out2]
+main:4: error: Property "x" defined in "Frozen" is read-only
+main:13: error: Unsupported left operand type for < ("NoCmp")
+main:14: error: Unsupported left operand type for <= ("NoCmp")
+main:15: error: Unsupported left operand type for > ("NoCmp")
+main:16: error: Unsupported left operand type for >= ("NoCmp")
+
+[case testIncrementalDataclassesDunder]
+from a import A
+reveal_type(A)  # E: Revealed type is 'def (a: builtins.int) -> a.A'
```
ilevkivskyi: Also here, please don't add errors on the first run.

test-data/unit/check-incremental.test
```diff
@@ -4362,6 +4362,297 @@ import b
...
+from a import A
+from dataclasses import dataclass
+
+@dataclass
+class B(A):
+    y: int
+
+[builtins fixtures/list.pyi]
+[out1]
+[out2]
```
ilevkivskyi: I like this test case, simple and to the point.

test-data/unit/check-incremental.test
```diff
@@ -4362,6 +4362,297 @@ import b
...
+from a import A
+from dataclasses import dataclass
+
+@dataclass
+class B(A):
+    y: str
+
+[builtins fixtures/list.pyi]
+[out1]
+main:2: error: Argument 2 to "B" has incompatible type "str"; expected "int"
```
ilevkivskyi: I like this one too. Although it has an error on first run, we should keep it, just to be sure, the error is cleared when error is fixed.

test-data/unit/check-incremental.test
```diff
@@ -4362,6 +4362,297 @@ import b
...
+from b import B
+from dataclasses import dataclass
+
+@dataclass
+class C(A, B):
+    c: bool
+
+[builtins fixtures/list.pyi]
+[out1]
+[out2]
```
ilevkivskyi: I would introduce a simple change in one of `a` or `b` on the second run.

test-data/unit/check-dataclasses.test
```diff
@@ -0,0 +1,403 @@
...
+
+@dataclass
+class B:
+  x: int
+
+reveal_type(A)  # E: Revealed type is 'def (b: __main__.B) -> __main__.A'
+A(b=B(42))
+A(b=42)  # E: Argument "b" to "A" has incompatible type "int"; expected "B"
+
+[builtins fixtures/list.pyi]
```
ilevkivskyi: No newline at the end of file. We really should make this a lint error :-)

test-data/unit/lib-stub/dataclasses.pyi
Bogdanp: This definition diverges from the one in typeshed in that it returns `_T` rather than `Field[_T]`. Returning `Field[_T]` breaks things like:

```
@dataclass
class Foo:
  x: int = field(default=42)
```

Because `int` and `Field[int]` are incompatible.

Bogdanp:
> Note that we will still need to add function plugin hooks to make asdict, astuple, replace etc. more precise (they can't be typed precisely in typeshed) in a separate PR. Are you interested in working on the second PR?

I don't think I'll have the free time to tackle that. In fact, I won't have time to work on this PR either in the next two or three weeks. Feel free to take over if you feel there are any other changes that need to be made.

PR #5012, opened 2018-05-07
Update protocol caches when generating fg caches on incremental build

msullivan:
Previously using `--cache-fine-grained` and `--incremental` together
would produce invalid protocol caches, which would prevent using
`--incremental` to quickly generate fine-grained cache artifacts.

This also adds machinery to `testfinegrained` to test building cache
artifacts using incremental runs.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -965,6 +965,7 @@ def verify_module(fscache: FileSystemCache, id: str, path: str) -> bool:
 
 
 def write_protocol_deps_cache(proto_deps: Dict[str, Set[str]],
+                              old_meta_snapshot: Dict[str, str],
                               manager: BuildManager, graph: Graph) -> None:
     """Write cache files for protocol dependencies.
 
@@ -981,7 +982,7 @@ def write_protocol_deps_cache(proto_deps: Dict[str, Set[str]],
     meta_snapshot = {}  # type: Dict[str, str]
     error = False
     for id in graph:
-        meta_snapshot[id] = graph[id].source_hash
+        meta_snapshot[id] = graph[id].source_hash or old_meta_snapshot[id]
     if not atomic_write(proto_meta, json.dumps(meta_snapshot), '\n'):
         manager.log("Error writing protocol meta JSON file {}".format(proto_cache))
         error = True
@@ -996,7 +997,7 @@ def write_protocol_deps_cache(proto_deps: Dict[str, Set[str]],
 
 
 def read_protocol_cache(manager: BuildManager,
-                        graph: Graph) -> Optional[Dict[str, Set[str]]]:
+                        graph: Graph) -> Optional[Tuple[Dict[str, str], Dict[str, Set[str]]]]:
     """Read and validate protocol dependencies cache.
 
     See docstring for write_protocol_cache for details about which kinds of
@@ -1011,8 +1012,8 @@ def read_protocol_cache(manager: BuildManager,
     current_meta_snapshot = {}  # type: Dict[str, str]
     for id in graph:
         meta = graph[id].meta
-        assert meta is not None, 'Protocol cache should be read after all other'
-        current_meta_snapshot[id] = meta.hash
+        if meta:
+            current_meta_snapshot[id] = meta.hash
     common = set(meta_snapshot.keys()) & set(current_meta_snapshot.keys())
     if any(meta_snapshot[id] != current_meta_snapshot[id] for id in common):
         # TODO: invalidate also if options changed (like --strict-optional)?
@@ -1027,7 +1028,7 @@ def read_protocol_cache(manager: BuildManager,
         manager.log('Could not load protocol cache: cache is not a dict: {}'
                     .format(type(deps)))
         return None
-    return {k: set(v) for (k, v) in deps.items()}
+    return (meta_snapshot, {k: set(v) for (k, v) in deps.items()})
 
 
 def _load_json_file(file: str, manager: BuildManager,
@@ -1608,7 +1609,7 @@ class State:
     path = None  # type: Optional[str]  # Path to module source
     xpath = None  # type: str  # Path or '<string>'
     source = None  # type: Optional[str]  # Module source code
-    source_hash = None  # type: str  # Hash calculated based on the source code
+    source_hash = None  # type: Optional[str]  # Hash calculated based on the source code
     meta = None  # type: Optional[CacheMeta]
     data = None  # type: Optional[str]
     tree = None  # type: Optional[MypyFile]
@@ -2089,6 +2090,7 @@ class State:
             return
         dep_prios = self.dependency_priorities()
         dep_lines = self.dependency_lines()
+        assert self.source_hash is not None
         new_interface_hash, self.meta = write_cache(
             self.id, self.path, self.tree,
             {k: list(v) for k, v in self.fine_grained_deps.items()},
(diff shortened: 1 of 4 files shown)


mypy/build.py
```diff
@@ -1011,8 +1012,8 @@ def read_protocol_cache(manager: BuildManager,
     current_meta_snapshot = {}  # type: Dict[str, str]
     for id in graph:
         meta = graph[id].meta
-        assert meta is not None, 'Protocol cache should be read after all other'
-        current_meta_snapshot[id] = meta.hash
+        if meta:
```
ilevkivskyi: I am a bit worried about this change. My understanding is a following: if I first run mypy with only `--incremental` protocol cache is not written, if I then change some files and run again with `--incremental` and `--cache-fine-grained` it should still revert to a full run. In general, the caching logic is quite complex, and I think we need to first add tests that full run actually happens if:
* the above scenario happens
* protocol caches are inconsistent (we have wrong source hashes, because the protocol cache was copied from somewhere).

msullivan: I've manually tested those but will try to get a real test set up. Unfortunately our testing setup is really not good at this sort of thing.

The first case works fine in a pretty natural way without actually needing to trigger the retry logic really: every cache file fails to load because it doesn't have a fine-grained deps cache.

msullivan:
I added some tests for both of those conditions, though the corrupting the cache file is done via a somewhat dubious hack.

I added some more documentation, though mostly I tried to clean up the logic, eliminating `old_snapshot` entirely and checking the cached snapshot against the source hash of everything with a meta, not just things whose metas were fresh.

test-data/unit/check-incremental.test
```diff
@@ -4283,3 +4283,31 @@ def foo() -> None:
...
+-- the proto deps file with something with a hash mismatch.
+[file ../.mypy_cache/3.6/@proto_deps.meta.json.2]
+{"__main__": "00000000000000000000000000000000", "a": "d41d8cd98f00b204e9800998ecf8427e", "b": "d41d8cd98f00b204e9800998ecf8427e", "builtins": "3d22863e34b3d846306e1e76ebc82407"}
+[file b.py.2]
+# uh
+-- Every file should get reloaded, since the cache was invalidated
+[stale a, b, builtins]
+[rechecked a, b, builtins]
+
+[case testIncrementalBustedFineGrainedCache2]
```
ilevkivskyi: Maybe also add a test where everything works for `--cache-fine-grained` and `--incremental` (i.e. flag is used on both runs, we don't touch proto_deps, and then check that only relevant files are rechecked)?

test-data/unit/check-incremental.test
```diff
@@ -4283,3 +4283,31 @@ def foo() -> None:
...
+[case testIncrementalBustedFineGrainedCache1]
+# flags: --cache-fine-grained
+import a
+import b
+[file a.py]
+[file b.py]
+-- This is a heinous hack, but we simulate having a invalid cache by clobbering
+-- the proto deps file with something with a hash mismatch.
+[file ../.mypy_cache/3.6/@proto_deps.meta.json.2]
+{"__main__": "00000000000000000000000000000000", "a": "d41d8cd98f00b204e9800998ecf8427e", "b": "d41d8cd98f00b204e9800998ecf8427e", "builtins": "3d22863e34b3d846306e1e76ebc82407"}
```
ilevkivskyi: I think we need to somehow exclude `builtins` from the snapshot in tests, otherwise all such tests need to be updated when we change something in `lib_stub`. Or alternatively use a dedicated named file and `[builtins ...]`.

msullivan: It doesn't matter in this case really because the point is that it is going to mismatch anyways, but this would be a problem in any situation where we want a correct hash.

I'll make the hash all 0s, though, to make clear it doesn't matter whether it matches.

test-data/unit/check-incremental.test
msullivan: Done

mypy/build.py
```diff
@@ -1008,11 +1016,12 @@ def read_protocol_cache(manager: BuildManager,
...
     if meta_snapshot is None:
         return None
-    current_meta_snapshot = {}  # type: Dict[str, str]
-    for id in graph:
-        meta = graph[id].meta
-        assert meta is not None, 'Protocol cache should be read after all other'
-        current_meta_snapshot[id] = meta.hash
+    # Take a snapshot of the source hashes from all of the metas we found.
+    # (Including the ones we rejected because they were out of date.)
+    # We use this to verify that the match up with the proto_deps.
```
JukkaL: Looks like a typo: the -> they?

mypy/test/data.py
```diff
@@ -308,6 +308,11 @@ def teardown(self) -> None:
                 pycache = os.path.join(path, '__pycache__')
                 if os.path.isdir(pycache):
                     shutil.rmtree(pycache)
+                # As a somewhat nasty hack, ignore any dirs with .mypy_cache in the path,
+                # to allow test cases to intentionally corrupt the cache without provoking
+                # the test suite when there are still files left over.
+                if '/.mypy_cache' in path:
```
JukkaL: I wonder if the `/` might cause trouble on Windows.

mypy/test/data.py
```diff
@@ -358,8 +363,8 @@ def find_steps(self) -> List[List[FileOperation]]:
             for path in paths:
                 module = module_from_path(path)
                 steps.setdefault(num, []).append(DeleteFile(module, path))
-        max_step = max(steps)
-        return [steps[num] for num in range(2, max_step + 1)]
+        max_step = max(steps) if steps else 2
```
JukkaL: Add comment about the 2 default.

mypy/test/testcheck.py
```diff
@@ -103,18 +102,16 @@ def run_case(self, testcase: DataDrivenTestCase) -> None:
                         raise ValueError(
                             'Output file {} exists though test case only has {} runs'.format(
                                 file, num_steps))
+            steps = testcase.find_steps()
             for step in range(1, num_steps + 1):
-                self.run_case_once(testcase, step)
+                ops = steps[step - 2] if step - 2 < len(steps) else []
```
JukkaL: For step 1, the index will be -1, which seems wrong. Apparently the value of `ops` is ignored in that case so this doesn't affect behavior, but it's still worth fixing.

mypy/test/testfinegrained.py
```diff
@@ -205,6 +205,14 @@ def format_triggered(self, triggered: List[List[str]]) -> List[str]:
             result.append(('%d: %s' % (n + 2, ', '.join(filtered))).strip())
         return result
 
+    def get_build_steps(self, program_text: str) -> int:
```
JukkaL: Add docstring. In particular, it's worth explaining the return value.

mypy/test/testfinegrained.py
```diff
@@ -81,22 +81,13 @@ def run_case(self, testcase: DataDrivenTestCase) -> None:
...
-                config = name  # type: Optional[str]
-                break
-        else:
-            config = None
-        if config:
-            parse_config_file(options, config)
+        build_options = self.get_options(main_src, testcase, build_cache=True)
         server = Server(options)
 
+        build_steps = self.get_build_steps(main_src)
```
JukkaL: The naming is not very clear. Maybe rename to `num_regular_incremental_steps` or something.

test-data/unit/fine-grained-cache-incremental.test
```diff
@@ -0,0 +1,183 @@
+-- Test cases for building caches for fine-grained mode using incremental
+-- builds.
+--
+-- '# num_build_steps: N' specifies how many build steps to do before
```
JukkaL: Maybe make more explicit, such as "... how many regular build steps to do without the daemon...".

PR #4910, opened 2018-04-14
Load .data.json files on-demand in fine-grained mode

msullivan:
Depends on #4906.

This gives around a 13s speed up for cold runs on S.

Issue #4498, opened 2018-01-22
"Invalid type" when using type alias in import cycle


JukkaL:
Type checking the following files results in an unexpected "Invalid type" error:

`t.py`:

```py
MYPY = False
if MYPY:
    from t2 import A

x: A  # Invalid type "t2.A"   <---- unexpected
```

`t2.py`:

```py
import t
from typing import Callable

A = Callable[[], None]
```

The reason for the error is that `t` and `t2` form an import cycle and `t` is processed before `t2` in the cycle, and thus the type alias `A` hasn't been processed when we analyze `t`.

This is probably the same bug as #4442, but for users this may not be obvious, so it perhaps makes maintain a separate issue.

PR #5016, opened 2018-05-09
Restore `all_are_submodules` import logic as workaround for #4498

### LINKED ISSUE #5015: "Invalid Type" error in v0.600 on uncached runs involving file with same name as parent package
When I run mypy on my project, I sometimes get an invalid type error. This error seems to go away over time (some kind of caching issue?), or if I avoid having a file in a package of the same name, or if I do an explicit direct import of the type from the file instead of importing it via the package.

This is kind of a complicated bug to reproduce, but here is my best attempt at making a succinct repro. We will need five files.

0. Create a fresh virtual environment and test project directory.

    ```bash
    mkvirtualenv repro --python=/usr/bin/python3
    pip install mypy
   mkdir /tmp/whatever
   cd /tmp/whatever
   mkdir project
   ```

1. Have a `project/__init__.py` file containing

    ```
    from project.study import CustomType
    ```

2. Have an empty `project/root.py` file.

3. Have a `project/neighbor/__init__.py` file containing

    ```python
    from project.study import CustomType
    def m(arg: CustomType) -> str:
        del arg
        return 'test'
    ```

4. Have a `project/study/__init__.py` file containing

    ```python
    from project.study.study import CustomType
    ```

5. Have a `project/study/study.py` file containing

    ```python
    from project import root
    CustomType = str
    ```

    Note: renaming this file to something that doesn't match its directory's name will cause the repro to fail.

6. (If you already ran mypy in this directory.) Clear the mypy cache.

    ```bash
    rm -rf .mypy_cache
    ```

7. Run mypy on *exactly these files in exactly this order*.

    ```bash
    mypy project/root.py project/study/study.py project/neighbor/__init__.py
   ```

8. And you will get this output:

    ```
    project/neighbor/__init__.py:4: error: Invalid type "project.study.study.CustomType"
    ```

    If you run mypy again without clearing its cache, then there will be no error.

    or, depending on factors I don't understand you may get this different error instead:

    ```
    project/neighbor/__init__.py:2: error: Internal error (node is None, kind=1)
    ```

An easy workaround for this issue is to avoid defining types in files that have the same name as their parent directory. Everyone likes avoiding cache invalidation issues by naming things, right?

msullivan:
The logic in build to determine what imported modules are depended on
used to elide dependencies to m in `from m import a, b, c` if all of
a, b, c were submodules. This was removed in #4910 because it seemed
like it ought not be necessary (and that semantically there *was*
a dependency), and early versions of #4910 depended removing it.

The addition of this dependency, though, can cause cycles that
wouldn't be there otherwise, which can cause #4498 (invalid type when
using aliases in import cycles) to trip when it otherwise wouldn't.

We've seen this once in a bug report and once internally, so restore
the `all_are_submodules` logic in avoid triggering #4498 in these
cases.

Fixes #5015

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -696,7 +696,6 @@ class BuildManager:
                                    blocker=True)
 
             return new_id
-
         res = []  # type: List[Tuple[int, str, int]]
         for imp in file.imports:
             if not imp.is_unreachable:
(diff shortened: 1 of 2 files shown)


msullivan:
Oh right, I had forgotten that to help justify removing that code I found a super obscure bug that removing it fixed. This will take marginally more thought if I don't just want to reintroduce it.

msullivan:
Though it looks like this fixes the #5015 bug but not the one we've observed internally. (Eliding the module when `all_are_submodules` doesn't either.)

PR #5021, opened 2018-05-11
Fix interpretation of returned generator expressions

### LINKED ISSUE #5020: Generator expression returned from a function is interpreted as an Iterator
I have a function like this:

```python
def foo(some_list: List[Any]) -> Generator[Any, None, None]:
    return (element for element in some_list)
```

This should be correct according to the documentation of [`typing.Generator`](https://docs.python.org/3/library/typing.html#typing.Generator), yet mypy complains that it got an Iterator as the return value:

    error: Incompatible return value type (got "Iterator[Any]", expected "Generator[Any, None, None]")

Question was originally posted on Stackoverflow [here](https://stackoverflow.com/questions/50273130/mypy-error-when-returning-a-generator-expression?noredirect=1#comment87617942_50273130). Versions are Python 3.6.4, mypy 0.600.

gmolau:
Fixes #5020 and updates relevant tests.

### DIFF
--- a/test-data/unit/check-isinstance.test
+++ b/test-data/unit/check-isinstance.test
@@ -1348,7 +1348,7 @@ l = [x for x in a if isinstance(x, int)]
 g = (x for x in a if isinstance(x, int))
 d = {0: x for x in a if isinstance(x, int)}
 reveal_type(l) # E: Revealed type is 'builtins.list[builtins.int*]'
-reveal_type(g) # E: Revealed type is 'typing.Generator[builtins.int*]'
+reveal_type(g) # E: Revealed type is 'typing.Generator[builtins.int*, builtins.None, builtins.None]'
 reveal_type(d) # E: Revealed type is 'builtins.dict[builtins.int*, builtins.int*]'
 [builtins fixtures/isinstancelist.pyi]
 
--- a/test-data/unit/check-isinstance.test
+++ b/test-data/unit/check-isinstance.test
@@ -1348,7 +1348,7 @@ l = [x for x in a if isinstance(x, int)]
 g = (x for x in a if isinstance(x, int))
 d = {0: x for x in a if isinstance(x, int)}
 reveal_type(l) # E: Revealed type is 'builtins.list[builtins.int*]'
-reveal_type(g) # E: Revealed type is 'typing.Iterator[builtins.int*]'
+reveal_type(g) # E: Revealed type is 'typing.Generator[builtins.int*]'
 reveal_type(d) # E: Revealed type is 'builtins.dict[builtins.int*, builtins.int*]'
 [builtins fixtures/isinstancelist.pyi]
 
--- a/test-data/unit/check-isinstance.test
+++ b/test-data/unit/check-isinstance.test
@@ -1348,7 +1348,7 @@ l = [x for x in a if isinstance(x, int)]
 g = (x for x in a if isinstance(x, int))
 d = {0: x for x in a if isinstance(x, int)}
 reveal_type(l) # E: Revealed type is 'builtins.list[builtins.int*]'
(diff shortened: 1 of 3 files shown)


test-data/unit/check-isinstance.test
```diff
@@ -1348,7 +1348,7 @@ l = [x for x in a if isinstance(x, int)]
 g = (x for x in a if isinstance(x, int))
 d = {0: x for x in a if isinstance(x, int)}
 reveal_type(l) # E: Revealed type is 'builtins.list[builtins.int*]'
-reveal_type(g) # E: Revealed type is 'typing.Iterator[builtins.int*]'
+reveal_type(g) # E: Revealed type is 'typing.Generator[builtins.int*]'
```
JelleZijlstra: Even though tests pass, I don't think this is right, because Generator takes three arguments, not one. I don't think there's a really elegant way to handle this, but somehow `check_generator_or_comprehension` needs to pass in all three args.

gmolau: That is a bit strange, changing the signatures to the proper `'typing.Generator[builtins.int*, builtins.None, builtins.None]'` makes the test fail, even though this type is used in other [places](https://github.com/python/mypy/blob/a2fd96280ba9127426f24c17bad351a16a5e1c70/test-data/unit/check-optional.test#L458).

JelleZijlstra: To be clear, you'll need to change checkexpr.py, not here—this just shows the wrong output.

gmolau: Yeah, I was just wondering why the correct signature does work in other places, but not here. The AsyncGenerator has even more examples where that is the case:

```
$ grep -r -n "typing.AsyncGenerator\[.*,.*\]" test-data/unit/
test-data/unit//check-async-await.test:218:    reveal_type(lst)  # E: Revealed type is 'typing.AsyncGenerator[builtins.int*, builtins.None]'
test-data/unit//check-async-await.test:220:    reveal_type(lst2)  # E: Revealed type is 'typing.AsyncGenerator[builtins.int*, builtins.None]'
test-data/unit//check-async-await.test:444:reveal_type(g)  # E: Revealed type is 'def () -> typing.AsyncGenerator[builtins.int, builtins.None]'
test-data/unit//check-async-await.test:445:reveal_type(g())  # E: Revealed type is 'typing.AsyncGenerator[builtins.int, builtins.None]'
test-data/unit//check-async-await.test:484:    reveal_type(gen.__aiter__())  # E: Revealed type is 'typing.AsyncGenerator[builtins.int*, builtins.None]'
```

JelleZijlstra: I guess mypy just doesn't check the number of arguments when you construct a Generator instance internally.

ethanhs:
@gmolau are you interested in continuing this work?

gmolau:
@ethanhs I'll give it another go this weekend.

gmolau:
Sorry, I won't have time to work on this in the next few weeks. I suspect that [this](https://github.com/python/mypy/blob/4389341707fd5da128c48eb2440ec927190cd682/mypy/checkexpr.py#L2529) call is not producing the right signature for generator expressions, but but I don't know enough about the Mypy internals to correct this.

gvanrossum:
@JelleZijlstra Maybe you want to take this one over the finish line? I've got the feeling that @gmolau is not going to be able to get to it any time soon.

JelleZijlstra:
Yes, I can take it.

JelleZijlstra:
Should be mostly OK now but I'm confused over "None" vs. "builtins.None" in the test. I'll sort it out later today.

JelleZijlstra:
OK, now it's fine. This is ready for review (and it's still a pretty simple change).

ethanhs:
@JelleZijlstra can this be merged?

JelleZijlstra:
Yes, I guess I was waiting for somebody to push the button.

PR #5023, opened 2018-05-12
Allow multiple shadow files to be specified

garetht:
This pull request allows the `--shadow-file` argument to be specified multiple times, allowing multiple different source files to be replaced by their corresponding shadow files.

Tests have also been added to ensure the original and new behaviors both work.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -468,7 +468,8 @@ Here are some more useful flags:
   make transformations to a file before type checking without having to change
   the file in-place.  (For example, tooling could use this to display the type
   of an expression by wrapping it with a call to reveal_type in the shadow
-  file and then parsing the output.)
+  file and then parsing the output.) This argument may be specified multiple times
+  to make mypy substitute multiple different files with their shadow replacements.
 
 .. _no-implicit-optional:
 
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -655,9 +655,14 @@ class BuildManager:
         self.fscache = fscache
         self.find_module_cache = FindModuleCache(self.fscache)
 
-        # a mapping from files to their corresponding shadow files
+        # a mapping from source files to their corresponding shadow files
+        # for efficient lookup
         self.shadow_map = {}  # type: Dict[str, str]
-        # a mapping from incoming files to files that exist in the shadow map
+        if self.options.shadow_file is not None:
+            self.shadow_map = {source_file: shadow_file
+                               for (source_file, shadow_file)
+                               in self.options.shadow_file}
+        # a mapping from typechecked files to files to their possible shadow files
         self.shadow_equivalence_map = {}  # type: Dict[str, Optional[str]]
 
     def use_fine_grained_cache(self) -> bool:
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -662,7 +662,7 @@ class BuildManager:
             self.shadow_map = {source_file: shadow_file
                                for (source_file, shadow_file)
                                in self.options.shadow_file}
-        # a mapping from typechecked files to files to their possible shadow files
+        # a mapping from each file being typechecked to its possible shadow file
         self.shadow_equivalence_map = {}  # type: Dict[str, Optional[str]]
 
     def use_fine_grained_cache(self) -> bool:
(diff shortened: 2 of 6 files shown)


garetht:
Not sure why the tests are failing only in Python 3.5.1 - it seems the newly added method hasn't made its way into that test?

docs/source/command_line.rst
```diff
@@ -468,7 +468,8 @@ Here are some more useful flags:
   make transformations to a file before type checking without having to change
   the file in-place.  (For example, tooling could use this to display the type
   of an expression by wrapping it with a call to reveal_type in the shadow
-  file and then parsing the output.)
+  file and then parsing the output.) This argument may be specified multiple times
+  to make mypy substitute multiple different files with their shadow replacements.
```
gvanrossum: This needs an example -- do you write `--shadow-file X1 X2 Y1 Y2` or `--shadow-file X1 X2 --shadow-file Y1 Y2`? (I presume the latter, but some options use the former syntax, and argparse supports both.)

Also, this argument has caused many people to be confused. We should present an example explaining that `--shadow-file X1 X2` means that whenever the checker is asked to check `X1`, it actually reads and checks the contents of `X2`, but *diagnostics will refer to `X1`* (it may well be the latter bit that trips people up).

mypy/build.py
```diff
@@ -655,13 +655,36 @@ def __init__(self, data_dir: str,
...
+        self.shadow_equivalence_map = {}  # type: Dict[str, Optional[str]]
+
     def use_fine_grained_cache(self) -> bool:
         return self.cache_enabled and self.options.use_fine_grained_cache
 
     def maybe_swap_for_shadow_path(self, path: str) -> str:
-        if (self.options.shadow_file and
-                os.path.samefile(self.options.shadow_file[0], path)):
-            path = self.options.shadow_file[1]
+        if not self.options.shadow_file:
```
gvanrossum: IMO it would be better if this checked for `if not self.shadow_map`. The Options object has been processed already in `__init__()` above.

mypy/build.py
```diff
@@ -655,13 +655,36 @@ def __init__(self, data_dir: str,
...
+        previously_checked = path in self.shadow_equivalence_map
+        if not previously_checked:
+            for k in self.shadow_map.keys():
+                if self.fscache.samefile(path, k):
+                    self.shadow_equivalence_map[path] = self.shadow_map.get(k)
+                    break
+                else:
+                    self.shadow_equivalence_map[path] = None
+
+        shadow_file = self.shadow_equivalence_map.get(path)
```
gvanrossum: This and the rest of the function could be shortened to
```
return self.shadow_equivalence_map.get(path, path)
```

mypy/build.py
```diff
@@ -655,13 +655,36 @@ def __init__(self, data_dir: str,
...
     def maybe_swap_for_shadow_path(self, path: str) -> str:
-        if (self.options.shadow_file and
-                os.path.samefile(self.options.shadow_file[0], path)):
-            path = self.options.shadow_file[1]
+        if not self.options.shadow_file:
+            return path
+
+        previously_checked = path in self.shadow_equivalence_map
+        if not previously_checked:
+            for k in self.shadow_map.keys():
```
gvanrossum: I'd switch this to `for k, v in self.shadow_map.items()`, then you can use `v` instead of `self.shadow_map.get(k)` two lines below.

(Also perhaps use more imaginative names than `k, v` -- how about `source, shadow`?

mypy/fscache.py
```diff
@@ -149,6 +149,11 @@ def md5(self, path: str) -> str:
             self.read(path)
         return self.hash_cache[path]
 
+    def samefile(self, f1: str, f2: str) -> bool:
+        s1 = self.stat(f1)
+        s2 = self.stat(f2)
+        return os.path.samestat(s1, s2)  # type: ignore
```
gvanrossum: I'm curious -- why do you need `# type: ignore` here?

JelleZijlstra: We were missing this function in the stub. We just merged the PR adding it, so this ignore can be removed as soon as we sync typeshed.

gvanrossum:
Thanks! (I cancelled the tests of my final cleanup since Travis and AppVeyor are kind of busy.)

garetht:
Thanks for the quick review!

Issue #5007, opened 2018-05-05
Support egg installs with PEP 561 searching


FuegoFro:
When working on a repo containing types to be distributed in accordance with PEP 561 (such as https://github.com/numpy/numpy-stubs) it is surprisingly difficult to get mypy to actually see and respect those types. In particular, things that do not work (though I would expect them to):

- Adding the directory containing the `-stubs` directory to the `MYPYPATH`
- Running `python setup.py install`
- Running `pip install -e .`

I eventually was able to make it work by running `pip install .`, but (particularly for things like running `py.test` while developing one of these stubs repos) this seems non-optimal.

Interestingly, adding the directory containing the `-stubs` directory to `MYPYPATH` did allow me to resolve types in the root of the package we care about (in this case `numpy`), but not in nested packages (in this case `numpy.core`). It looks like this might have something to do with [this code](https://github.com/python/mypy/blob/master/mypy/build.py#L883) where `dir_chain` is an empty string when looking for a top-level package and therefore `find_lib_path_dirs` doesn't filter out the items in `lib_path`. When we're trying to get a nested package it's non-empty and can't be found in the `lib_path` dirs, which then get filtered out. We also don't populate `third_party_stubs_dirs` in any of the above cases because the items on `MYPYPATH` aren't looked at for that (only site packages) and the other two methods don't create a bare directory named `-stubs` in the site packages (one creates a `.egg` and the other creates a `.egg-link`).

I did these experiments on commit `7fe60a547c476c8b341a47b87d72fc4460532102` of mypy with Python 3.6.4 on macOS.

Let me know if there's anything else I can provide which would be helpful! I'm not sure if this is more of a bug or a feature request because I couldn't tell from my reading of PEP 561 whether this use case was covered by that.

ethanhs:
Running `python setup.py install` definitely should work and should be supported, as this does the same work that pip does to install packages. Can you please make sure that the package (the folder numpy-stubs) is installed after running `python setup.py install`? If not, then that is a setuptools issue.

Development installs on the other hand aren't intended to be supported, at least for now, as they don't put the files in the path, so you will need to handle this manually, just like before. I will look at the possibility of adding this.

FuegoFro:
Thank you for your quick reply!

[...]

PR #5026, opened 2018-05-13
Add support for setuptools eggs with PEP 561 searching

ethanhs:
Setuptools installs packages differently from wheels or normal
distutils, using "eggs". Eggs have three main formats that differ from the
normal method: editable, compressed, and uncompressed. This change adds
support for editable and uncompressed egg installs.

Before reviewing I *highly* recommend reading https://github.com/pypa/setuptools/blob/master/docs/formats.txt which describes the egg format.

There is also a work around for the fact that Ubuntu (and I believe all Debian based Linuxes) list non-existent site-package directories...

This fixes much of #5007, I will work on compressed egg installs on the flight to Cleveland :)

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -831,10 +831,12 @@ class FindModuleCache:
         self.dirs = {}  # type: Dict[Tuple[str, Tuple[str, ...]], List[str]]
         # Cache find_module: (id, lib_path, python_version) -> result.
         self.results = {}  # type: Dict[Tuple[str, Tuple[str, ...], Optional[str]], Optional[str]]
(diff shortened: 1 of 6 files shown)


ethanhs:
Well, I am flummoxed why this is failing, clearly the package is installed (the type information is picked up), but pip cannot uninstall it? Seems like a pip bug (I can repro locally too). I'll poke at it tomorrow.

ethanhs:
Based on more discussion with the packaging folks, I am first going to re-design how the pep561 tests are run, which will hopefully work around this issue.

Instead of interacting with one site-packages, I will create a virtualenv for each test. This also allows me to parrallelize the tests, hopefully making them run a bit faster.

PR #5036, opened 2018-05-14
Use `.dmypy.json` as the dmypy run file

msullivan:
No reason to clutter up the working directory.


msullivan:
(I've left `dmypy.json` in the `.gitignore` so that running an installed mypy 0.600 `dmypy` doesn't clutter things up)

PR #5038, opened 2018-05-14
Remove no longer used scripts

### LINKED ISSUE #1457: scripts/mypy doesn't work when run straight from the repo
When I don't have mypy installed, running scripts/mypy from the repo doesn't work. E.g.

```
$ ./scripts/mypy -h
  File "./scripts/mypy", line 4, in <module>
    from mypy.main import main
ImportError: No module named 'mypy'
$ 
```

This is unfortunate because README.md shows this as the way to run individual unit tests:

```
$ scripts/myunit -m mypy.test.testlex -v '*backslash*'
```

Also, unit tests that invoke mypy (e.g. mypy/test/testpythoneval.py) use scripts/mypy this way. The danger exists that those tests run the mypy that's installed rather than the one in the repo.

**UPDATE:** To run an individual unit test I use this instead of what the README.md suggests:

```
$ python3 -m mypy.myunit -m mypy.test.testlex -v '*backslash*'
51 test cases run, all passed.
*** OK ***
$ 
```

But this doesn't work for testpythoneval:

```
$ python3 -m mypy.myunit -m mypy.test.testpythoneval '*testHello'
Expected:
  hello, world                                  (diff)
Actual:
  Traceback (most recent call last):            (diff)
    File "/Users/guido/src/mypy/scripts/mypy", line 4, in <module> (diff)
      from mypy.main import main                (diff)
  ImportError: No module named 'mypy'           (diff)

  E: hello, world
  A: Traceback (most recent call last):
     ^
  File "/Users/guido/src/mypy/mypy/test/data.py", line 141, in run
    self.perform(self)
  File "/Users/guido/src/mypy/mypy/test/testpythoneval.py", line 102, in test_python_evaluation
    testcase.file, testcase.line))
  File "/Users/guido/src/mypy/mypy/test/helpers.py", line 86, in assert_string_arrays_equal
    raise AssertionFailure(msg)
AssertionFailure: Invalid output (mypy/test/data/pythoneval.test, line 8)

test_testpythoneval_PythonEvaluationSuite.testHello failed

1/1 test cases failed.
*** FAILURE ***
$
```

To work around this you have to set PYTHONPATH to an absolute path:

```
$ PYTHONPATH=`pwd` python3 -m mypy.myunit -m mypy.test.testpythoneval '*testHello'
1 test cases run, all passed.
*** OK ***
$ 
```

(Using `PYTHONPATH=.` doesn't work, because tests are run from a temporary directory.)

ethanhs:
These scripts were once useful, but now people should use `python -m mypy` instead.

Fixes #1457


scripts/stubgen
```diff
@@ -1,20 +0,0 @@
...
-
-file_dir = os.path.dirname(__file__)
-parent_dir = os.path.join(file_dir, os.pardir)
-if os.path.exists(os.path.join(parent_dir, '.git')):
-    # We are running from a git clone.
-    sys.path.insert(0, parent_dir)
-
-import mypy.stubgen
-
-mypy.stubgen.main()
```
ilevkivskyi: Will simple `stubgen file` still work after removing this?

ethanhs: No, but unless you modified your enviroment, it wouldn't anyway. You can do `python -m mypy.stubgen`. You will likely need to update some aliases.

Issue #3487, opened 2017-06-01
Type of conditional expression is join instead of union


JukkaL:
The type of expression such as `1 if foo else 'x'` is `object` (join of `int` and `str`). This is surprising and if such an expression is used in a union type context, the result is often not what's expected. Example:

```py
from typing import Union

def f(x: Union[int, str]) -> None: ...

c = 1
f(1 if c else 'x')  # Incompatible argument type "object"
```

This is also arguably inconsistent, as the `or` and `and` operators produce union types. For example, the type of `(1+2) or 'x'` is `Union[int, str]`.

Maybe conditional expressions should also produce unions? An alternative idea is to only produce union types if the type context is a union type, but this would be somewhat ad-hoc.

JukkaL:
cc @carljm

adsharma:
Comment related to #2464. The following shouldn't produce an error either.

```
class A:
    pass

class B:
    pass

class C:
    def __init__(self, a: A, b: B) -> None:
        if a and b:
            raise Exception("both a and b specified")

        if not a and not b:
            raise Exception("neither a nor b specified")

        self._obj: Union[A, B] = a if a else b
        reveal_type(a if a else b)
```

If A and B have a common ancestor, it makes sense to use it. But if it turns out to be `builtins.object` it's probably a good signal that we have a `NamedUnion` type of case where there are a couple of unrelated objects. So resolving to Union[A, B] makes more sense for that case?

JukkaL:
Bumped priority to high since the workaround is non-trivial and this seems to be affecting multiple users.

MentalMegalodon:
I'm working on this.

SKHecht:
So is the new behavior expected to be that the types of conditional expressions are unions instead of joins? I'm not seeing that. I'm on mypy 0.701 and when I run this code in mypy
```
class A:
    pass
class B:
    pass

def condition() -> bool:
    return True

reveal_type(A() if condition() else B())
```
I get
```
error: Revealed type is 'builtins.object'
```
Based on my understanding of the resolution of this issue I'd expect the type to `Union[A, B]`

ethanhs:
@SKHecht no. It originally was changed to make them Unions instead of joins, but with https://github.com/python/mypy/pull/5095 this was partially rolled back.

ilevkivskyi:
It is still a union in union context.

cjolowicz:
Should this be reopened?

Here is another use case:

```
s: str = "key=value"
[...]

PR #5041, opened 2018-05-14
Fix type of conditional statement to Union and not Join. (#3487)

MentalMegalodon:


test-data/unit/check-optional.test
```diff
@@ -385,7 +385,7 @@ def lookup_field(name, obj):
         attr = None
 
 [case testTernaryWithNone]
-reveal_type(None if bool() else 0)  # E: Revealed type is 'Union[builtins.int, builtins.None]'
+reveal_type(None if bool() else 0)  # E: Revealed type is 'Union[builtins.None, builtins.int]'
```
ilevkivskyi: Could you please add few dedicated tests? (Probably not in this test but in `check-unions`)

PR #5046, opened 2018-05-14
fix error message in pythoneval test

JelleZijlstra:
Needed after python/typeshed#2128; this will fail tests until the
typeshed change is merged.


JelleZijlstra:
@gvanrossum now that python/typeshed#2128 is merged, we'll have to sync typeshed and merge this to get CI working again for mypy and typeshed. I can do that tonight but don't have enough time right now.

gvanrossum:
I'll do it.

ethanhs:
Done!

PR #5051, opened 2018-05-15
Improve stub testing to cover all of typeshed

ethanhs:
This will fail until typeshed is synced with https://github.com/python/typeshed/pull/2130 and https://github.com/python/typeshed/pull/2121

Previously, we were only checking 2and3, 3.3-3.5 and 3. This checks 3.6, 3.7 and 2.


JelleZijlstra:
Would it be possible to replace this by an invocation of `tests/mypy_test.py` in the typeshed submodule using mypy from the current repository? That way we don't have to maintain two slightly different versions of the same logic.

ethanhs:
@JelleZijlstra IIRC you now want to do the opposite, replace the check in typeshed with this one? Is that right? Once this is merged it should be easy to just invoke this from typeshed.

JelleZijlstra:
I don't feel very strongly about this, but apparently this uncovered bugs in typeshed that typeshed's own tests didn't find, so it would be better to run this test in typeshed itself.

ilevkivskyi:
What is the status of this PR? `runtests.py` is going to be removed soon, so I am wondering if this is still relevant.

ethanhs:
test_stubs has moved in #5142, I think we should still change the tests to cover all of typeshed. I will modify this to change the pytest test instead of what is in runtests.py

ethanhs:
Okay this now targets the pytest test. I decided to split each version into its own test so pytest can parrallelize them.

mypy/test/testsamples.py
```diff
@@ -1,31 +1,57 @@
...
+
+    def test_2(self) -> None:
+        self.check_stubs("2.7", name="2")
+
+    def test_2and3_2(self) -> None:
+        self.check_stubs("2.7", name="2and3")
+
+    def test_2and3_3(self) -> None:
+        sys_ver_str = '.'.join(map(str, sys.version_info[:2]))
+        self.check_stubs(sys_ver_str, name="2and3")
```
Michael0x2a: Rather then checking the "2and3" directory in only these two cases, would it maybe make sense to pass in a list of directories we want to check into `check_stubs`?

E.g. we could do `self.check_stubs(["2.7", "2and3"])` inside `test_2`, do `self.check_stubs(["2and3", "3", "3.5"])` inside `test_35`...

I think this would probably make our tests more thorough, though the tradeoff is that each individual test would take longer.

(This is just an idea though -- maybe the tradeoff isn't worth it?)

ethanhs: Part of the reason I did this was we already run tests in a matrix of options on Travis, so we'd end up testing all the versions of Python we support if we use the system version info. I collapsed the 2/2and3 and 3/2and3 tests together, but I think otherwise, there isn't a point in adding 3/2and3 to each test.

PR #5054, opened 2018-05-15
Better error messages when calling overloaded function with invalid types.

### LINKED ISSUE #672: Give better error messages when calling overloaded function with invalid types
Currently when calling an overloaded function, the error message is something like `No overload variant of "foo" matches argument types`. Instead, we should try to give a message like `Argument 1 to "foo" has incompatible type "int"; expected "str" or "bytes"`, which is more informative. This may not be always possible, but we should be able to do it at least most of the time.

MentalMegalodon:
Fixes #672

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -752,6 +752,10 @@ class MessageBuilder:
                       .format(name, arg_types), context)
         else:
             self.fail('No overload variant matches argument types {}'.format(arg_types), context)
+        OFFSET = 2
+        MAX_ITEMS = 2
+        self.note('Possible overload variants:', context)
+        self.pretty_overload(overload, context, OFFSET, MAX_ITEMS)
 
     def wrong_number_values_to_unpack(self, provided: int, expected: int,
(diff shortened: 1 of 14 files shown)


Michael0x2a:
@MentalMegalodon -- would you be interested in continuing work on this? It'll unfortunately be a bit of a slog due to the merge conflicts (we've been making a variety of improvements to overloads over the past month).

The plus side is that addressing @gvanrossum's feedback should be easier now due to these changes: I think it basically boils down passing in the `plausible_targets` variable into the `no_variant_matches_arguments` function [here](https://github.com/python/mypy/blob/1b21e8c488c451d641a7e1da182ae96999a5af27/mypy/checkexpr.py#L1185) and using that info to display a more useful list of variants. In the case that the list is empty, we could fall back to the current behavior or something.

If you're not interested in continuing work on this, let me know and I can take over.

ilevkivskyi:
OK, this is now superseded by #5296 

Thanks for your work @MentalMegalodon!

PR #5058, opened 2018-05-15
Make local_partial_types per-module so that it triggers cache rebuilds

msullivan:


msullivan:
There are two different answers to that question:
1. It makes perfect semantic sense for it to be configurable per-module
2. I can't think of any reason why you would need to

I went with 1 because people think of lots of ways to use things that I wouldn't have and I didn't see any harm in it, but I'd be basically just as happy to just add it to `OPTIONS_AFFECTING_CACHE`

PR #5060, opened 2018-05-16
Make pep561 tests parrallel and sandboxed

ethanhs:
Each test is run in a seperate virtualenv which is deleted after the
tests. This means they can be run in parrallel, and caused about a 2x
speedup for pep561 on my machine (4x parrallel). This also will allow for the tests
to scale better with egg support, and bypass issues with uninstalling
some types of packages.

### DIFF
--- a/mypy/test/testpep561.py
+++ b/mypy/test/testpep561.py
@@ -1,8 +1,10 @@
 from contextlib import contextmanager
 import os
+import random
 import shutil
+import string
 import sys
-from typing import Iterator, List
+from typing import Iterator, List, Generator
 from unittest import TestCase, main
 
 import mypy.api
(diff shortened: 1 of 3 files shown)


ethanhs:
This is ready for review now. It isn't *as* important as I decided egg support serves little purpose, but it provides a nice speedup, and makes things cleaner.

ethanhs:
I just realized I might have been a bit vague in my last message. I'm hoping to get this merged soonish so that I can work on adding support for editable installs of packages (which I already have an implementation of), then support for partial stub packages. If someone finds the time I'd appreciate a review.

mypy/test/testpep561.py
```diff
@@ -19,112 +22,112 @@
...
         if returncode != 0:
             self.fail('\n'.join(lines))
-        try:
-            yield
-        finally:
-            returncode, lines = run_command([python_executable, '-m', 'pip', 'uninstall',
-                                             '-y', pkg], cwd=package_path)
-            if returncode != 0:
-                self.fail('\n'.join(lines))
+        yield
```
ilevkivskyi: I am not sure what is the benefit of context manager which has yield as a last statement. Can't be this just made a function?

ethanhs: Good point. I will refactor things into just using this as a function.

PR #5066, opened 2018-05-16
Clean up the traceback from the `run` mode

msullivan:
Don't start the server from inside a catch block, since that produces
and ugly traceback if the server crashes.


gvanrossum:
I believe we don't have any tests for dmypy behavior. :-( We'd have to design a new test harness probably.

msullivan:
Yeah, we would (and probably *do*) need a new harness. I expanded the fine grained testing to cover most of the tricky daemon logic, but there is still plenty that isn't.

Issue #4722, opened 2018-03-12
"Name already defined" error is confusing


sid-kap:
Mypy complains if an if-else statement redefines a global variable with a class definition.

Failing code:
```python
if 5 % 2 == 0:
  X = 5
else:
  class X:
    pass
```
Error:
```
test.py:4: error: Name 'X' already defined on line 2
```

mypy version: 0.570

gvanrossum:
That's intentional, because after the if-else X has a useless type.

sid-kap:
1. What do you mean by "has a useless type"?

2. Here's the code I'm depending on that's causing this error: https://github.com/ipython/ipython/blob/dc46d30ae2318456b5f3fb3cef37066ef2cc305e/IPython/core/history.py#L66. What do you suggest I do to fix this error?

gvanrossum:
I was referring to your example. Apparently you didn't capture the essence of the real code you were having trouble with. There are a number of special cases implemented when two branches define different but equivalent (to the user) types, and your example wasn't one of them.

It seems defining two different classes isn't one of them either. So I recommend putting `# type: ignore` on the exact line where you get the error.

sid-kap:
Ok, thanks!

sid-kap:
In any case, "Name already defined" seems like a very bad explanation for the error. The problem is not that X is already defined—mutating variables is something you do in Python all the time. The problem is that you're trying to reassign something to X that's of a different type.

gvanrossum:
OK, reopening as a usability issue -- long-term maintainers get blind for those little usability problems, so thanks for bringing that up!

glenjamin:
I've just run into the same error message, but in a different scenario.

I'm trying to generate some usable stubs for the `GitPython` project.

This is what `__init__.py` looks like:

# Stubs for git (Python 3.6)
#
# NOTE: This dynamically typed stub was automatically generated by stubgen.

```
from git.exc import *
from git.objects import *
from git.refs import *
from git.diff import *
from git.db import *
from git.remote import *
from git.index import *
from git.cmd import Git as Git
from git.config import GitConfigParser as GitConfigParser
from git.repo import Repo as Repo
from git.util import Actor as Actor, BlockingLockFile as BlockingLockFile, LockFile as LockFile, Stats as Stats, rmtree as rmtree
```

Which produces the following error:

```
type-stubs/git/__init__.pyi:6: error: Name 'util' already defined
type-stubs/git/__init__.pyi:7: error: Name 'tag' already defined
[...]

PR #5067, opened 2018-05-17
Improve "Name already defined" error message

sid-kap:
In `semanal.py`, `name_already_defined` takes an optional `original_ctx: Optional[SymbolTableNode]` argument which is used to print the line number of the original context where a name was defined already.

In many cases, the code doesn't pass anything for this argument, even though there is an original context that makes sense. This PR does this for the `name_already_defined` call in `add_symbol`, as discussed in #4722.

A few questions if you guys think this is a good idea:
- Should it also be done in `add_symbol` in `semanal_pass1.py` and `semanal_pass3.py`?
- I only made this change to cover the example that @glenjamin brought up in #4722, but there are other places in `semanal.py` where adding this additional argument could be helpful:
https://github.com/python/mypy/blob/fae1f255bee75cc031223690edb1137a6ea57987/mypy/semanal.py#L3177
https://github.com/python/mypy/blob/fae1f255bee75cc031223690edb1137a6ea57987/mypy/semanal.py#L3201
The function is also called with just two parameters in `_visit_func_def`, `_visit_overloaded_func_def`, and `analyze_lvalue`, but I'm not sure if there is a `original_ctx` that makes sense to pass there.
- Should I add tests?

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3258,7 +3258,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
         elif node and node.line != -1:
             extra_msg = ' on line {}'.format(node.line)
         else:
-            extra_msg = ''
+            extra_msg = ' (possibly by an import)'
         self.fail("Name '{}' already defined{}".format(name, extra_msg), ctx)
 
     def fail(self, msg: str, ctx: Context, serious: bool = False, *,
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3244,7 +3244,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
                 self.add_fixture_note(fullname, ctx)
 
     def name_already_defined(self, name: str, ctx: Context,
-                             original_ctx: Optional[Union[SymbolTableNode, SymbolNode]] = None) -> None:
+                    original_ctx: Optional[Union[SymbolTableNode, SymbolNode]] = None) -> None:
         if isinstance(original_ctx, SymbolTableNode):
             node = original_ctx.node
         elif isinstance(original_ctx, SymbolNode):
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -406,7 +406,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
                         # Redefinition. Conditional redefinition is okay.
                         n = self.locals[-1][defn.name()].node
                         if not self.set_original_def(n, defn):
-                            self.name_already_defined(defn.name(), defn)
+                            self.name_already_defined(defn.name(), defn, self.locals[-1][defn.name()])
                     else:
                         self.add_local(defn, defn)
             else:
@@ -3172,7 +3172,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
                 # Flag redefinition unless this is a reimport of a module.
                 if not (node.kind == MODULE_REF and
                         self.locals[-1][name].node == node.node):
-                    self.name_already_defined(name, context)
+                    self.name_already_defined(name, context, self.locals[-1][name])
             self.locals[-1][name] = node
         elif self.type:
             self.type.names[name] = node
@@ -3196,7 +3196,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
         assert self.locals[-1] is not None, "Should not add locals outside a function"
         name = node.name()
         if name in self.locals[-1]:
-            self.name_already_defined(name, ctx)
+            self.name_already_defined(name, ctx, self.locals[-1][name])
         node._fullname = name
         self.locals[-1][name] = SymbolTableNode(LDEF, node)
 
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -395,7 +395,8 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
                         # Redefinition. Conditional redefinition is okay.
                         n = self.type.names[defn.name()].node
                         if not self.set_original_def(n, defn):
-                            self.name_already_defined(defn.name(), defn)
+                            self.name_already_defined(defn.name(), defn,
+                                                      self.type.names[defn.name()])
                     self.type.names[defn.name()] = SymbolTableNode(MDEF, defn)
                 self.prepare_method_signature(defn, self.type)
             elif self.is_func_scope():
@@ -406,7 +407,8 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
                         # Redefinition. Conditional redefinition is okay.
                         n = self.locals[-1][defn.name()].node
                         if not self.set_original_def(n, defn):
-                            self.name_already_defined(defn.name(), defn, self.locals[-1][defn.name()])
+                            self.name_already_defined(defn.name(), defn,
+                                                      self.locals[-1][defn.name()])
                     else:
                         self.add_local(defn, defn)
             else:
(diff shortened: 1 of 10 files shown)


JelleZijlstra:
I think anywhere this message appears it's useful to give the original definition, if possible.

Tests would be good to add. You'll probably just have to change the existing tests though.

sid-kap:
Thanks for taking a look at this! I updated the tests to what I think should be the correct output, but in many cases the line numbers it outputs aren't what I expected. I'll have to figure out why these are wrong.

sid-kap:
It looks like when `SymbolTableNode` is created in `semanal.py` for an import, it stores the node of the referenced module, not the node of the import statement:
https://github.com/python/mypy/blob/master/mypy/semanal.py#L1310-L1314
Therefore, `symbol_table_node.node.line` is the line of the imported module, which will always be 1.

So it seems that the options are either:

1) modify `SymbolTableNode` to store line numbers for imports if `kind` is `MODULE_REF`, or 
2) before raising a "Name already defined" error, check if the original type is a module reference (i.e. check if `existing.kind == MODULE_REF`) and don't report the line number in that case.

Which would be the right thing to do?

ilevkivskyi:
> Which would be the right thing to do?

I would just not show the additional note in this case. Anyway, conditional imports will require special support which will probably add later. I will now make a review, sorry for waiting so long with this.

mypy/semanal.py
```diff
@@ -406,7 +407,8 @@ def _visit_func_def(self, defn: FuncDef) -> None:
                         # Redefinition. Conditional redefinition is okay.
                         n = self.locals[-1][defn.name()].node
                         if not self.set_original_def(n, defn):
-                            self.name_already_defined(defn.name(), defn)
+                            self.name_already_defined(defn.name(), defn,
+                                                      self.locals[-1][defn.name()])
```
ilevkivskyi: It looks like you missed one more case: global function definition, see `else: # Top-level function` below. You will likely need to patch also `self.check_no_global` in the same way you do above for `self.name_already defined`

mypy/semanal.py
```diff
@@ -1848,7 +1850,12 @@ def analyze_lvalue(self, lval: Lvalue, nested: bool = False,
                 self.type.names[lval.name] = SymbolTableNode(MDEF, v)
             elif explicit_type:
                 # Don't re-bind types
-                self.name_already_defined(lval.name, lval)
+                global_def = self.globals.get(lval.name)
+                local_def = cast(SymbolTable, self.locals[-1]).get(lval.name) if self.locals and self.locals[-1] else None
```
ilevkivskyi: If you rewrite it using an if statement, then the cast will be unnecessary.

sid-kap:
No worries! I can make that change, the tests should pass after I fix that.

mypy/semanal.py
```diff
@@ -3230,10 +3237,16 @@ def name_not_defined(self, name: str, ctx: Context) -> None:
                 self.add_fixture_note(fullname, ctx)
 
     def name_already_defined(self, name: str, ctx: Context,
-                             original_ctx: Optional[SymbolTableNode] = None) -> None:
+                             original_ctx: Optional[Union[SymbolTableNode, SymbolNode]] = None) -> None:
```
ilevkivskyi: In general, it is not a good idea to use a union here, but it seems to me the old signature was "wrong", so it is fine in this case. Maybe you can add a short comment about this?

mypy/semanal.py
```diff
@@ -3230,10 +3237,16 @@ def name_not_defined(self, name: str, ctx: Context) -> None:
...
+            if isinstance(original_ctx, SymbolTableNode):
+                node = original_ctx.node
+            elif isinstance(original_ctx, SymbolNode):
+                node = original_ctx
+            else:
+                node = None
+            if node and node.line != -1:
+                extra_msg = ' on line {}'.format(node.line)
             else:
                 extra_msg = ' (possibly by an import)'
```
ilevkivskyi: Do we have enough tests for this case? If not you can add one more.

test-data/unit/semanal-errors.test
```diff
@@ -1347,7 +1347,7 @@ def dec(x: Any) -> Any:
 @dec
 def f() -> None:
     pass
-@dec  # E: Name 'f' already defined
+@dec  # E: Name 'f' already defined on line 4
```
ilevkivskyi: Wow, good job updating all these error messages! 👍

mypy/semanal.py
sid-kap: I'm not sure when this would happen (when the node's line number is -1). There don't seem to be any tests covering it, so maybe this never happens?

mypy/semanal.py
sid-kap: The old signature was actually correct, the function used to only be called with `SymbolTableNode`s. I added the option for passing a `SymbolNode` instead because in `_visit_overloaded_func_def` I only have a `SymbolNode` to pass it, not a `SymbolTableNode`.

sid-kap: So I guess to fix this, the options are
- take `Optional[SymbolNode]` instead
- take `Optiona[SymbolTableNode]` only, and in `_visit_overloaded_func_def` create a `SymbolTableNode` to pass to the function

mypy/semanal.py
sid-kap: I think `self.check_no_global` already does the right thing?
https://github.com/python/mypy/blob/7eebb8f9d433488f7deff7b4cba14dc3a9650743/mypy/semanal.py#L3225

mypy/semanal.py
sid-kap: Fixed!

mypy/semanal.py
sid-kap: Actually, looks like there was one test where this happened. Do you know in what situations this happens? If so, I could add tests for those cases.

mypy/semanal.py
ilevkivskyi: OK, if the previous context is always correct in `check_no_global`, then no action is required here.

mypy/semanal.py
ilevkivskyi: I mean that `SymbolTableNode` is logically "wrong" because it is just an entry in the symbol table wrapper, while `SymbolNode` is an actual error context (well to be more precise, `SymbolNode`s are semantic objects, not syntactic, but at least they are closer, and should ideally point to the line of syntax node defining them). So I vote for `SymbolNode`.

mypy/semanal.py
ilevkivskyi: I think this might happen in case of an error or an unresolved circular import. If you can't find a simple additional test case than ignore this.

mypy/semanal.py
```diff
@@ -1848,7 +1850,19 @@ def analyze_lvalue(self, lval: Lvalue, nested: bool = False,
                 self.type.names[lval.name] = SymbolTableNode(MDEF, v)
             elif explicit_type:
                 # Don't re-bind types
-                self.name_already_defined(lval.name, lval)
+                global_def = self.globals.get(lval.name)
+                if self.locals:
+                    locals_last = self.locals[-1]
```
ilevkivskyi: @gvanrossum probably will be happy to see another case in mypy for PEP 572 :-)

@sid-kap You can ignore this comment.

mypy/semanal.py
ilevkivskyi: Actually you might need `.kind == MODULE_REF` which is only available on `SymbolTableNode`. Potentially, this can be replaced by `isinstance(node, (MypyFile, ImportedName))`, but I think it is better to just leave it as is.

sid-kap: Okay, I'll leave it as-is then.

mypy/semanal.py
sid-kap: Hmm, looks like there's already a test case for that situation (`testDuplicateDefFromImport` in `semanal-errors.test`). So I guess we can leave this as-is.

Issue #5072, opened 2018-05-17
Remove `builtins.` from error messages


JelleZijlstra:
Our error messages would be a bit more concise if we removed the `builtins.` part from type output, and perhaps `typing.` too. This would make error messages match up more closely with runtime syntax.

For example, we currently output `None` as `builtins.None`, which is a syntax error at runtime. Just outputting `None` would be more clear.

JelleZijlstra:
There's a couple of others:
```
unit/check-bound.test:x = None # type: G[str] # E: Type argument "builtins.str" of "G" must be a subtype of "__main__.A"
unit/check-bound.test:class D3(C[str]): pass # E: Type argument "builtins.str" of "C" must be a subtype of "__main__.A"
unit/check-classes.test:tmp/foo.pyi:5: error: Invalid signature "def (foo.C, Any, Any) -> builtins.int"
unit/check-classes.test:main:6: error: Invalid signature "def (__main__.C, builtins.str, builtins.str) -> __main__.C"
unit/check-classes.test:main:6: error: Invalid signature "def (__main__.C, builtins.str, builtins.str) -> __main__.C"
unit/check-classes.test:    def __setattr__(self, name: int, value: int) -> None: ...  # E: Invalid signature "def (__main__.C, builtins.int, builtins.int)"
unit/check-classes.test:NT = NewType('NT', N) # E: Argument 2 to NewType(...) must be subclassable (got TypedDict('__main__.N', {'x': builtins.int}))
unit/check-classes.test:main:4: error: t: builtins.str
unit/check-classes.test:main:4: error: y: builtins.float
unit/check-expressions.test:main:4: error: x: builtins.int
unit/check-expressions.test:main:4: error: y: builtins.int
unit/check-expressions.test:main:4: error: z: builtins.int
unit/check-functions.test:main:4: error: a: builtins.int
unit/check-functions.test:main:4: error: b: builtins.int
unit/check-functions.test:main:6: error: a: builtins.int
unit/check-functions.test:main:6: error: b: builtins.int
unit/check-functions.test:main:6: error: c: builtins.int
unit/check-functions.test:main:9: error: a: builtins.float
unit/check-modules.test:tmp/has_getattr.pyi:1: error: Invalid signature "def (builtins.int, builtins.str) -> builtins.str"
unit/check-namedtuple.test:yb: G[int] # E: Type argument "builtins.int" of "G" must be a subtype of "Tuple[builtins.int, fallback=__main__.M]"
[...]

PR #5073, opened 2018-05-17
Remove "builtins." from output for overloads and builtins.None

JelleZijlstra:
Part of #5072.

Overloads are probably the most common user-facing feature that still outputs "builtins" in error messages; this diff fixes them to use standard type formatting.

I also change `builtins.None` to `None`, since `builtins.None` is a syntax error. A comment claimed that we use `builtins.None` to distinguish the type from the value, but I don't think there are a lot of contexts where that confusion is likely.

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -754,7 +754,8 @@ class MessageBuilder:
         arg_types_str = ', '.join(self.format(arg) for arg in arg_types)
         num_args = len(arg_types)
         if num_args == 0:
-            self.fail('All overload variants{} require at least one argument'.format(name_str))
+            self.fail('All overload variants{} require at least one argument'.format(name_str),
+                      context)
         elif num_args == 1:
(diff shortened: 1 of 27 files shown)


ilevkivskyi:
Thanks for cleaning this up! I would propose to change the error format to something like this:
* `No overload variant of "f" of "A" matches argument type "B"` if there is a single argument
* `No overload variant of "f" of "A" matches argument types "A", "B", "C"` otherwise.

(Note that quotes are used by some editor plugins to better highlight types/methods in error messages.)

JelleZijlstra:
Yes, that would be better. What do you suggest for the case where the list is empty? (All overloads take at least one argument, but the user called `f()`.) I suggest `All overloads of "f" require at least one argument`.

ilevkivskyi:
(also error message in tests needs to be updated)

JelleZijlstra:
Oops, just pushed a commit fixing the two tests I messed up.

PR #5075, opened 2018-05-18
Refactoring: Cleanup DataSuite fields

elazarg:
Simple changes to DataSuite, I believe they make things somewhat more readable and explicit.

* Default `base_path` to `test_temp_dir`; this clearly marks tests that do something unusual 
* Rename `optional_out` to `required_out_section` and default to `False`
* Remove `update_data`, and inline the only place it is used. The flag might be reintroduced if the functionality will be shared with other suites
* Document reason for `data_prefix`, since it is not written anywhere

### DIFF
--- a/mypy/test/data.py
+++ b/mypy/test/data.py
@@ -659,7 +659,7 @@ class DataSuite:
 
     base_path = test_temp_dir
 
-    # Allow mypyc to update to using a recent version of mypy. See #4779
+    # Allow external users of the test code to override the data prefix
     data_prefix = test_data_prefix
 
     required_out_section = False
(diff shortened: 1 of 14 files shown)


mypy/test/data.py
```diff
@@ -657,17 +656,20 @@ def has_stable_flags(testcase: DataDrivenTestCase) -> bool:
 class DataSuite:
     # option fields - class variables
     files = None  # type: List[str]
-    base_path = '.'
+
+    base_path = test_temp_dir
+
+    # Allow mypyc to update to using a recent version of mypy. See #4779
```
msullivan: I would say "Allow external users of the test code to override the data prefix" or something

PR #5076, opened 2018-05-18
Refactoring: Remove handling of @include in test suites

elazarg:
This is very old code, and I couldn't find any hint of it doing anything.


mypy/test/data.py
```diff
@@ -505,24 +504,6 @@ def collapse_line_continuation(l: List[str]) -> List[str]:
     return r
 
 
-def expand_includes(a: List[str], base_path: str) -> List[str]:
-    """Expand @includes within a list of lines.
-
-    Replace all lies starting with @include with the contents of the
```
JelleZijlstra: lies indeed.

gvanrossum:
Didn't even knew we had that!

PR #5087, opened 2018-05-20
Add self-check to pytests

elazarg:
Yet another item for #1673.

This PR uses the api for the self-check. 

Example of a failed-run output:
```
~/workspace/mypy$ pytest -n0 -k test_mypy_package
================================ test session starts ================================
platform darwin -- Python 3.6.5, pytest-3.2.5, py-1.5.2, pluggy-0.4.0
rootdir: ~/workspace/mypy, inifile: pytest.ini
plugins: xdist-1.20.1, forked-0.2, cov-2.5.1
collected 6240 items                                                                 

mypy/test/testselfcheck.py F

===================================== FAILURES ======================================
_________________________ SelfCheckSuite.test_mypy_package __________________________
Self check failed
------------------------------- Captured stdout call --------------------------------
mypy/binder.py:254: error: No return value expected
=============================== 6239 tests deselected ===============================
==================== 1 failed, 6239 deselected in 23.75 seconds =====================
```

`$ ./runtests.py self-check` results in a similar output, though slightly more verbose, as usual.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,6 @@ install:
   - pip install .
 
 script:
-  - python runtests.py -j12 -x lint -x package
+  - python runtests.py -j12 -x lint -x self-check
   - if [[ $TRAVIS_PYTHON_VERSION == '3.6' ]]; then flake8; fi
   - if [[ $TRAVIS_PYTHON_VERSION == '3.5.1' ]]; then python runtests.py package; fi
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -21,7 +21,7 @@ build: off
 
 test_script:
     # Ignore lint and mypy self check (both run in Travis)
-    - "%PYTHON%\\python.exe runtests.py -x lint -x package -x pytest"
+    - "%PYTHON%\\python.exe runtests.py -x lint -x self-check -x pytest"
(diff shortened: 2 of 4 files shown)


elazarg:
3.4 fails the type check with this error: 
> mypy/test/data.py:232: error: "TemporaryDirectory" expects no type arguments, but 1 given

I have no idea why did it pass before. (See short discussion [here](https://github.com/python/mypy/pull/3957#discussion_r139049936))

JelleZijlstra:
Maybe we just didn't run a self-check with 3.4 before? That error is consistent with the stubs (https://github.com/python/typeshed/blob/master/stdlib/3/tempfile.pyi): TemporaryDirectory is generic only in 3.5+.

ilevkivskyi:
I think self check is only run on 3.6. Btw @elazarg I added a new label https://github.com/python/mypy/labels/topic-tests for test related issues.

elazarg:
Where's the appveyor check? I've only touched `.travis.yml`.

ethanhs:
@elazarg it is excluded on this line https://github.com/python/mypy/blob/master/appveyor.yml#L24

elazarg:
@ethanhs is it OK now?

PR #5094, opened 2018-05-22
Do not serialize symbol table nodes added by 'add_submodules_to_parent_modules'

ilevkivskyi:
A hot fix for a complicated crash due to a poisoned cache (discovered internally).

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -2500,6 +2500,7 @@ class SymbolTableNode:
                  'implicit',
                  'is_aliasing',
                  'alias_name',
+                 'no_serialize'
                  )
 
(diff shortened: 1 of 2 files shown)


mypy/nodes.py
```diff
@@ -2500,6 +2500,7 @@ class SymbolTableNode:
                  'implicit',
                  'is_aliasing',
                  'alias_name',
+                 'no_serialize'
```
gvanrossum: Trailing comma. Also add a comment to the end of the docstring about it please.

ilevkivskyi: Sorry, already merged before seeing your comment. Will fix this later.

PR #5095, opened 2018-05-22
Don't always infer unions for conditional expressions

JukkaL:
Use a join for type of a conditional expression unless the type context
is a union type.

A recent change (#5041) modified the type to always use a union, but this
caused many problems in internal Dropbox repos that we aren't ready to
fix yet.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2357,7 +2357,14 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             # branch's type.
             else_type = self.analyze_cond_branch(else_map, e.else_expr, context=if_type)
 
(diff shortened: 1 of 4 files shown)


mypy/checkexpr.py
```diff
@@ -2357,7 +2357,14 @@ def visit_conditional_expr(self, e: ConditionalExpr) -> Type:
             # branch's type.
             else_type = self.analyze_cond_branch(else_map, e.else_expr, context=if_type)
 
-        res = UnionType.make_simplified_union([if_type, else_type])
+        # Only create a union type if the type context is a union to be mostly
```
gvanrossum: comma before "to be mostly"

PR #5099, opened 2018-05-23
Explain how to work around a certain name conflict

### LINKED ISSUE #5097: Invalid type reported when class property matches valid top-level type.
Run mypy (0.600) on the following code:

```
class C: ...

class D: ...

class A:
    C: D

    def f(self, x: C) -> None: ...
```

This reports:

`mymodule.py:8: error: Invalid type "mymodule.A.C"`

for the line `def f() ...`

Note that I see the same issue if `C` is a method on `A`:

```
class C: ...

class A:
    def C(self) -> None: ...

    def f(self, x: C) -> None: ...
```

`mymodule.py:6: error: Invalid type "mymodule.A.C"`

But *not* if C is an attribute with type `Any`:

```
from typing import Any

class C: ...

class A:
    C: Any

    def f(self, x: C) -> None: ...
```

(no error)

My actual use case is I'm trying to write a stub for a library class that contains a property named `str` and this is making it a bit difficult.

gvanrossum:
Fixes #5097

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -539,21 +539,25 @@ Dealing with conflicting names
 
 Suppose you have a class with a method whose name is the same as an
 imported (or built-in) type, and you want to use the type in another
-method signature.  E.g.
+method signature.  E.g.:
 
 .. code-block:: python
 
-   class C:
-       def str(self): ...
-       def method(self: arg: str): ...
+   class Message:
+       def bytes(self):
+           ...
+       def register(self, path: bytes):  # error: Invalid type "mod.Message.bytes"
+           ...
 
 The third line elicits an error because mypy sees the argument type
-``str`` as a reference to the method by that name.  Other than
+``bytes`` as a reference to the method by that name.  Other than
 renaming the method, a work-around is to use an alias:
 
 .. code-block:: python
 
-   str_ = str
-   class C:
-       def str(self): ...
-       def method(self: arg: str_): ...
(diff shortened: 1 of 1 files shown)


docs/source/common_issues.rst
```diff
@@ -532,3 +532,28 @@ put the linter comment *after* the type comment:
...
+Dealing with conflicting names
+------------------------------
+
+Suppose you have a class with a method whose name is the same as an
+imported (or built-in) type, and you want to use the type in another
+method signature.  E.g.
+
+.. code-block:: python
+
+   class C:
```
ilevkivskyi: I would give these more descriptive name. For example, `C` -> `Message`, `str` -> `bytes` (these IIRC are from an older similar issue).

docs/source/common_issues.rst
```diff
@@ -532,3 +532,28 @@ put the linter comment *after* the type comment:
...
+       def str(self): ...
+       def method(self: arg: str): ...
```
ilevkivskyi: typo `self:` -> `self,`

docs/source/common_issues.rst
```diff
@@ -532,3 +532,28 @@ put the linter comment *after* the type comment:
...
+
+The third line elicits an error because mypy sees the argument type
+``str`` as a reference to the method by that name.  Other than
```
ilevkivskyi: Maybe mention the actual error that mypy gives in this case?

PR #5102, opened 2018-05-24
Add a `--mypy-verbose` option to pytest to turn on verbosity

msullivan:


mypy/test/data.py
```diff
@@ -578,6 +578,8 @@ def pytest_addoption(parser: Any) -> None:
     group.addoption('--update-data', action='store_true', default=False,
                     help='Update test data to reflect actual output'
                          ' (supported only for certain tests)')
+    group.addoption('--mypy-verbose', action='count',
```
ilevkivskyi: Since it is `'count'`, does it mean I need to write `--mypy-verbose` twice to emulate `-vv`?

msullivan: Yeah

Issue #4772, opened 2018-03-21
"<function> defined here" notes omitted when function is loaded from cache


msullivan:
mypy emits a nice "<function> defined here" note when calling a function with unexpected keywords arguments.

Unfortunately the note is omitted if the module the function is defined in has been loaded from the cache.
This both eliminates a nice message (which we should add more of!) and causes annoying deviations between modes.

The following coarse-grained incremental test fails
```
[case testCachedUnexpectedKeywordArgument]
import a
[file a.py]
import b
b.lol(uhhhh=12) # tweak
[file a.py.2]
import b
b.lol(uhhhh=12)
[file b.py]
def lol() -> None: pass
[out]
tmp/a.py:2: error: Unexpected keyword argument "uhhhh" for "lol"
tmp/b.py:1: note: "lol" defined here
[out2]
tmp/a.py:2: error: Unexpected keyword argument "uhhhh" for "lol"
tmp/b.py:1: note: "lol" defined here
```

As well as this fine-grained incremental test:
```
[case testUnexpectedKeywordArgument]
import a
[file a.py]
import b
[file a.py.2]
import b
b.lol(uhhhh=12)
[file b.py]
def lol() -> None: pass
[out]
==
a.py:2: error: Unexpected keyword argument "uhhhh" for "lol"
b.py:1: note: "lol" defined here
```

ethanhs:
It seems this may be blocking #1412.

ilevkivskyi:
I tried to fix this by serializing just the necessary parts for these better error messages (serializing full definition would be too expensive), but here is a problem. If I write the original definition line number to the cache, then it becomes part of public interface (what is called interface hash in `build.py`). This is problematic, since it would trigger to many other files to recheck. This only affects coarse grained runs, but in any case it seems wrong to make line number a part of public interface. So if we want better error messages, we need to rethink how we determine public interface, so that we can store some extra metadata.

Issue #1412, opened 2016-04-20
Use incremental mode in python evaluation test cases


JukkaL:
The Python evaluation test cases (which don't actually run the code if there are type errors...) use full stubs for `builtins` and `typing`. Using incremental mode would likely speed them up significantly, as processing stubs probably takes the majority of time. These tests are the long pole in the full test suite, so this could give a nice speed boost.

gvanrossum:
I looked into this for a few minutes and one problem is that currently each test is run in a clean directory. There currently is no way to change the name of the cache directory, and it's always `.mypy_cache` relative to the current directory. So we'd need something for this. (I've filed #1414 to track it.)

ethanhs:
I just tried making the pythoneval tests use a shared cache directory via this diff:
```
diff --git a/mypy/test/testpythoneval.py b/mypy/test/testpythoneval.py
index cf4cbafd..9c207d5d 100644
--- a/mypy/test/testpythoneval.py
+++ b/mypy/test/testpythoneval.py
@@ -14,7 +14,7 @@ import os
 import os.path
 import re
 import sys
-
+from tempfile import TemporaryDirectory
 import pytest  # type: ignore  # no pytest in typeshed

 from typing import List
@@ -37,12 +37,12 @@ class PythonEvaluationSuite(DataSuite):
              'pythoneval-asyncio.test']
     base_path = test_temp_dir
     optional_out = True
-
+    cache_dir = TemporaryDirectory()
     def run_case(self, testcase: DataDrivenTestCase) -> None:
-        test_python_evaluation(testcase)
+        test_python_evaluation(testcase, os.path.join(self.cache_dir.name, '.mypy_cache'))


-def test_python_evaluation(testcase: DataDrivenTestCase) -> None:
+def test_python_evaluation(testcase: DataDrivenTestCase, cache_dir: str) -> None:
     """Runs Mypy in a subprocess.

     If this passes without errors, executes the script again with a given Python
@@ -71,6 +71,7 @@ def test_python_evaluation(testcase: DataDrivenTestCase) -> None:
     with open(program_path, 'w') as file:
         for s in testcase.input:
             file.write('{}\n'.format(s))
+    mypy_cmdline.append('--cache-dir={}'.format(cache_dir))
     output = []
     # Type check the program.
     out, err, returncode = api.run(mypy_cmdline)
```
However I get a failure (consistently) with:
```
_______________ testCanConvertTypedDictToAnySuperclassOfMapping _______________
[...]

PR #5109, opened 2018-05-24
Preserve the name of first argument in methods when serializing

ilevkivskyi:
This is useful to give nicer error messages after serialization.

This is a step towards #4772 
Unblocks #1412 (much faster `pythoneval` tests).

### DIFF
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -698,6 +698,7 @@ class CallableType(FunctionLike):
                  special_sig: Optional[str] = None,
                  from_type_type: bool = False,
                  bound_args: Sequence[Optional[Type]] = (),
+                 def_extras: Optional[Dict[str, Any]] = None,
                  ) -> None:
         super().__init__(line, column)
         assert len(arg_types) == len(arg_kinds) == len(arg_names)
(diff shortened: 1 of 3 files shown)


ethanhs:
The lint issue seemed minor so I went ahead and fixed it. I can confirm this solves the issue I came across in #1412!

mypy/types.py
```diff
@@ -907,25 +923,28 @@ def serialize(self) -> JsonDict:
...
+                                 Instance.deserialize(data['fallback']),
+                                 name=data['name'],
+                                 variables=[TypeVarDef.deserialize(v) for v in data['variables']],
+                                 is_ellipsis_args=data['is_ellipsis_args'],
+                                 implicit=data['implicit'],
+                                 is_classmethod_class=data['is_classmethod_class'],
+                                 bound_args=[(None if t is None else deserialize_type(t))
+                                            for t in data['bound_args']],
+                                 )
+        call_type.def_extras = data['def_extras']
```
msullivan: Would it be cleaner to add `def_extras` to the `__init__`?

ilevkivskyi:
@ethanhs You can now go ahead with your #1412 fix.

PR #5113, opened 2018-05-25
Fix testcmdline when it is run with py.test -n0

### LINKED ISSUE #4127: Single threaded pytest spuriously fails for testcmdline and testpythoneval
[Reported](https://github.com/python/mypy/pull/3870#issuecomment-337110885) by @gvanrossum in #3870 :

> Sadly there's a problem with this. When I run the tests as follows:
> 
```
pytest -n0 mypy/test/testcmdline.py
```
> every test in that file fails. Apparently the pytest-xdist extension which runs tests in parallel does something that keeps the tests from failing. The error is essentially that every subprocess invoking scripts/mypy fails with the following traceback:
```
   File "/Users/guido/src/mypy/scripts/mypy", line 4, in <module>
     from mypy.main import main
 ModuleNotFoundError: No module named 'mypy.main'
```
> This makes sense as the current directory is set to the temporary directory where the test files are written, so the repo root is not on sys.path. I'm guessing that the xdist extension somehow retains the current directory?
> 
> The pythoneval tests fail in the same way.

(End quote)

On my machine most of the tests pass, except `testCoberturaParser`:
```
Expected:
  <coverage timestamp="$TIMESTAMP" version="$VERSION" line-rate="0.8000" branch-rate="0"> (diff)
  ...
Actual:
  <coverage timestamp="$TIMESTAMP" version="0.530" line-rate="0.8000" branch-rate="0"> (diff)
  ...
```
Apparently `pytest-xdist` adds mypy's base dir to PYTHONPATH.

This issue has two items:
- [ ] Fix the failures
- [ ] Add tests to prevent the reintroduction of such failures

msullivan:
The test runner sets PYTHONPATH when parallelized but not with -n0.
Our subprocess was depending on this to find mypy. We should set up
the state we care about ourselves, so just do that.

The common failure mode here was the tests failing, but I also ran
into a more terrifying mode where it was invoking my installed system
mypy!

This is basically the same fix elazarg proposed in #4127 but that
never went anywhere for unclear reasons.

Fixes #4127.

### DIFF
--- a/mypy/test/testcmdline.py
+++ b/mypy/test/testcmdline.py
@@ -11,7 +11,7 @@ import sys
 
 from typing import List
 
-from mypy.test.config import test_temp_dir
+from mypy.test.config import test_temp_dir, PREFIX
 from mypy.test.data import fix_cobertura_filename
 from mypy.test.data import DataDrivenTestCase, DataSuite
 from mypy.test.helpers import assert_string_arrays_equal, normalize_error_messages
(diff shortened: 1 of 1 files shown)


msullivan:
...

PR #5120, opened 2018-05-29
Allow wildcards inside of configuration section names

### LINKED ISSUE #5014: Version 0.600 disallows per module configuration globs like mypy-project.apps.*.migrations.*
I have a Django project with many apps and I want to disable mypy in all of `migrations` modules of each app. Up to `0.590` it was possible and since `0.600` this feature seems to be dropped. 

This is my `mypy.ini` config:
```mypy.ini
[mypy-project.apps.*.migrations.*]
ignore_errors=True
```

msullivan:
This is to support Django-style usecases with patterns like
`site.*.migrations.*`.

The implementation works by mixing in the old-style glob matching with
the new structured matching.

Fixes #5014.

### DIFF
--- a/mypy/options.py
+++ b/mypy/options.py
@@ -1,8 +1,10 @@
 from collections import OrderedDict
+import fnmatch
+import re
 import pprint
 import sys
 
-from typing import Dict, List, Mapping, MutableMapping, Optional, Set, Tuple
+from typing import Dict, List, Mapping, MutableMapping, Optional, Pattern, Set, Tuple
 
 from mypy import defaults
 
@@ -167,8 +169,8 @@ class Options:
         self.plugins = []  # type: List[str]
 
         # Per-module options (raw)
-        pm_opts = OrderedDict()  # type: OrderedDict[str, Dict[str, object]]
-        self.per_module_options = pm_opts
+        self.per_module_options = OrderedDict()  # type: OrderedDict[str, Dict[str, object]]
+        self.glob_options = []  # type: List[Tuple[str, Pattern[str]]]
         self.unused_configs = set()  # type: Set[str]
 
         # -- development options --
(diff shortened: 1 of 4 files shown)


mypy/options.py
```diff
@@ -208,27 +210,54 @@ def __ne__(self, other: object) -> bool:
...
+                                  if '*' in k[:-1]]
+        structured_keys = [k for k in self.per_module_options.keys()
+                           if '*' not in k[:-1]]
+        wildcards = sorted(k for k in structured_keys if k.endswith('.*'))
+        concrete = [k for k in structured_keys if not k.endswith('.*')]
+
+        for glob in unstructured_glob_keys:
+            self.glob_options.append((glob, re.compile(fnmatch.translate(glob))))
+
+        # We (for ease of implementation), treat unstructured glob
```
gvanrossum: No comma.

mypy/options.py
```diff
@@ -208,27 +210,54 @@ def __ne__(self, other: object) -> bool:
...
+        # sections as used if any real modules use them or if any
+        # concrete config sections use them. This means we need to
```
gvanrossum: Oh, this is a little shady, and explain why a.*.b is not an error in the test, right?

mypy/options.py
```diff
@@ -208,27 +210,54 @@ def __ne__(self, other: object) -> bool:
...
             options = self.clone_for_module(key)
             # And then update it with its per-module options.
-            new_options = Options()
-            new_options.__dict__.update(options.__dict__)
-            new_options.__dict__.update(self.per_module_options[key])
+            new_options = options.apply_changes(self.per_module_options[key])
             self.per_module_cache[key] = new_options
 
-        self.unused_configs = set(keys)
+        # Add the more structured sections into unused configs .
```
gvanrossum: Space before period

mypy/options.py
msullivan: Yeah I'll add a comment to the test.

mypy/options.py
```diff
@@ -250,18 +279,33 @@ def clone_for_module(self, module: str) -> 'Options':
...
         for i in range(len(path), 0, -1):
             key = '.'.join(path[:i] + ['*'])
             if key in self.per_module_cache:
                 self.unused_configs.discard(key)
-                return self.per_module_cache[key]
+                options = self.per_module_cache[key]
+                break
+
+        # OK and *now* we need to look for glob matches
+        if not module.endswith('.*'):
```
gvanrossum: When would module ever end in a star?

mypy/options.py
```diff
@@ -208,27 +210,54 @@ def __ne__(self, other: object) -> bool:
...
+        for key in wildcards + concrete:
             # Find what the options for this key would be, just based
             # on inheriting from parent configs.
             options = self.clone_for_module(key)
             # And then update it with its per-module options.
-            new_options = Options()
-            new_options.__dict__.update(options.__dict__)
-            new_options.__dict__.update(self.per_module_options[key])
+            new_options = options.apply_changes(self.per_module_options[key])
             self.per_module_cache[key] = new_options
```
gvanrossum: Perhaps combine with previous line? (Unless it doesn't fit.)

mypy/options.py
```diff
@@ -208,27 +210,54 @@ def __ne__(self, other: object) -> bool:
...
+        # Unstructured glob configs are stored and are all checked for each module.
+        unstructured_glob_keys = [k for k in self.per_module_options.keys()
+                                  if '*' in k[:-1]]
+        structured_keys = [k for k in self.per_module_options.keys()
+                           if '*' not in k[:-1]]
+        wildcards = sorted(k for k in structured_keys if k.endswith('.*'))
+        concrete = [k for k in structured_keys if not k.endswith('.*')]
+
+        for glob in unstructured_glob_keys:
+            self.glob_options.append((glob, re.compile(fnmatch.translate(glob))))
```
gvanrossum: IIUC this means that module `foo.bar` does not match the pattern `foo.*.bar`. And that's surprising since `foo.bar` does matchs `foo.bar.*`.

gvanrossum:
(So maybe it's fine if foo.bar doesn't match foo.*.bar, but then it should
be called out in the docs.)

mypy/options.py
```diff
@@ -302,11 +302,13 @@ def clone_for_module(self, module: str) -> 'Options':
...
-        # If the pattern is 'mod.*', we want 'mod' to match that too.
-        # (That's so that a pattern specifying a package also matches
-        # that package's __init__.)
-        return pattern.match(module) is not None or pattern.match(module + '.') is not None
+    def compile_glob(self, s: str) -> Pattern[str]:
+        # Compile one of the glob patterns to a regex so that '.*' can
+        # match *zero or more* module sections. This means we compile
+        # '.*' into '(\..*)?'. We also need to escape .s in the glob, so
+        # we hackily rewrite .s to ,s to accomplish this.
+        s = s.replace('.', ',').replace(',*', '(\..*)?').replace(',', '\.')
```
gvanrossum: This doesn't work right for *.foo I think. Also maybe the code would be more readable if you just started with module.split('.') and applied re.escape() to parts that aren't `*`, and something else for other parts, special-casing a `*` at the front.

PR #5132, opened 2018-06-01
Always patch indirect dependencies

ilevkivskyi:
The bug and fix are straightforward: previously caches generated in full (non-incremental) mode lacked _all_ indirect dependencies. This makes it also a bit scary, there are probably lots of invalid caches out there waiting to explode.

cc @dzbarsky


msullivan:
It frightens me that this never caused problems for us before.

PR #5139, opened 2018-06-03
option to merge stubs into source trees

### LINKED ISSUE #5028: Test stub files against real code
There are a few use cases when stubs may be compelling over type annotations or type comments:

- annotate code that needs to be Python 2 compatible, ideally by using type annotations (so no type comments), 
- often times users prefer to have type information outside of the source code (this is to avoid modifying legacy code change - and in practice lowers resistance from maintainers),
- the maintainer of the library does not want to take on the in-line type annotations, someone else trying to provide it as a separate package.

However at the moment this has some limitations: stub files are used only to check clients of the stub, not the code itself. 

```python
# a.py
def add_int(a, b):
    return a + b

add_int("a", "b") # mypy does not complain, actually ignores the entire file
```

```python
# a.pyi
def add_int(a:int, b:int) -> int: ...
```

```python
# client.py
from a import add_int
add_int(1, 2)
add_int(1, "2") # here mypy is going to complain
```

In case of the above files mypy will warn about bad usage of inside ``client.py``, but not in ``a.py`` (that is, mypy does not validate against the source files being annotated by the stub). This has serious drawbacks: 
- maintaining the stub files is hard, now users need to manually make sure whatever is in the stub file is an accurate representation of the matching source files (both interface - aka number of arguments, names - and type wise),
- the stub creator does not know if the source code itself is type correct or not.

Here I propose to implement a way to support testing stub files against the real code:

- build the syntax tree for the source file,
- build the syntax tree for the stub file, 
- merge stub tree into the source tree
  - source file only tree available: dynamically typed - noop
  - abstract syntax tree only: raise error - missing annotated sources
  - both stub and source AST exists:
    - for all source elements for what there is a matching stub element
      - copy inject over type annotations
      - complain if function interface missmatch
- now just run the existing type checking mechanism on the enriched source code ast

Merging the AST definitely is not trivial. The plan is to start with a POC implementation, that works for the most cases, and then see the viability of this approach. Note this would help a lot both stub maintainers I think (typeshed) and projects that need Python 2 support.

gaborbernat:
Resolves #5028. 

Need to clean up, both logic mostly there (tested with https://github.com/gaborbernat/tox/tree/type-hint-stubs/tox).

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -400,7 +400,7 @@ class ASTConverter:
         return MypyFile(body,
                         self.imports,
                         False,
-                        {*ignores},
+                        set(ignores),
                         )
 
     # --- stmt ---
--- a/mypy/fastparse2.py
+++ b/mypy/fastparse2.py
@@ -308,7 +308,7 @@ class ASTConverter:
         return MypyFile(body,
                         self.imports,
                         False,
-                        {*ignores},
+                        set(ignores),
                         )
 
     # --- stmt ---


mypy/nodes.py
```diff
@@ -270,6 +281,165 @@ def deserialize(cls, data: JsonDict) -> 'MypyFile':
...
+                if i.name not in src_names:
+                    defs.append(i)
+        defs.extend(self.defs)
+        self.defs = defs
+
+
+def merge_names(src: 'SymbolTable', stub: 'SymbolTable', errors: 'StubOrSrcErrors') -> None:
+    for key, value in stub.items():
+        if key not in src:
+            src.__setitem__(key, value)
```
JelleZijlstra: Why are you calling `__setitem__` directly instead of doing `src[key] = value`?

mypy/nodes.py
```diff
@@ -270,6 +281,165 @@ def deserialize(cls, data: JsonDict) -> 'MypyFile':
...
+            main_value = src.get(key)
+            if main_value is not value:
+                merge_symbols(main_value, value, errors)
+
+
+def merge_symbols(src: Any, stub: Any, errors: 'StubOrSrcErrors') -> None:
+    t = type(src)
+    if t != type(stub):
+        errors.report(src.line,
+                      "definition miss-match of src {} and stub (line {}) {}".format(
```
JelleZijlstra: mismatch

mypy/nodes.py
```diff
@@ -270,6 +281,165 @@ def deserialize(cls, data: JsonDict) -> 'MypyFile':
...
+                          src.__class__.__qualname__,
+                          stub.line,
+                          stub.__class__.__qualname__,
+                      ))
+        return
+    if src is None or t in (ImportedName,):
+        pass
+    elif t == SymbolTableNode:
```
JelleZijlstra: We usually prefer using `isinstance`

mypy/test/test_stub_src_merge.py
```diff
@@ -0,0 +1,119 @@
...
+        result_ast, result_res = _build_ast(BuildSource(result_path, None, None))
+        capsys.readouterr()
+        assert result_ast == combined_ast
+        return combined_res
+
+    def _build_ast(source):
+        path = source.merge_with.path if source.merge_with is not None else source.path
+        res = build.build(sources=[source], options=options, alt_lib_path=tmpdir)
+        regex = re.compile(r'\s+{}\s+^'.format(re.escape(path)), re.MULTILINE)
+        ast_str = line_nr.sub('\1',
```
JelleZijlstra: Needs to be `r'\1'`

mypy/test/test_stub_src_merge.py
```diff
@@ -0,0 +1,119 @@
...
+
+    yield _do
+
+
+def test_merge_module_func(checker):
+    result = checker("""
+    def fancy_add(a, b = None):
+        pass""", """
+    from typing import Union
+    def fancy_add(a : int, b : Union[None, int] = ...) -> int: ...""", """
```
JelleZijlstra: No space before :

mypy/test/testcmdline.py
```diff
@@ -58,6 +58,14 @@ def test_python_cmdline(testcase: DataDrivenTestCase) -> None:
     outb = process.stdout.read()
     # Split output into lines.
     out = [s.rstrip('\n\r') for s in str(outb, 'utf8').splitlines()]
+
+    is_running_in_py_charm = "PYCHARM_HOSTED" in os.environ
```
JelleZijlstra: This looks unrelated to your PR.

tox.ini
```diff
@@ -0,0 +1,12 @@
+[tox]
```
JelleZijlstra: Again unrelated to your PR.

.gitignore
```diff
@@ -1,3 +1,5 @@
+*.dirty
+.tox
```
JelleZijlstra: This is repeated (and I don't think it should be in this PR at all; see below).

mypy/build.py
```diff
@@ -59,18 +60,14 @@
 from mypy.server.deps import get_dependencies
 from mypy.fscache import FileSystemCache
 from mypy.typestate import TypeState, reset_global_state
-
```
JelleZijlstra: Avoid making style changes unrelated to your PR like these.

gvanrossum:
I think this would need a flag to enable it; I expect that not everybody will want this functionality.

This PR probably ought to be merged quickly to avoid running into merge conflicts; but it requires very thorough review (probably by Jukka or Ivan).

mypy/test/testcmdline.py
gaborbernat: It is slightly unrelated but needed to allow me run tests inside PyCharm, which helps me create the PR. Once the PR is in merge ready state I can pull it out into a different PR.

tox.ini
gaborbernat: Same as above; I will remove into a different PR once other code parts look good.

gaborbernat:
@gvanrossum fair enough, I'll hide it behind a flag.

JukkaL:
I can review this later this week. Agreed that we want to merge this quickly -- even if this is not 100% finished, if it's behind a flag and doesn't risk breaking other functionality we can hopefully merge it, and it can be iterated on afterwards.

mypy/test/testcmdline.py
ethanhs: How are you running things from PyCharm? I've been able to run the test suite for months without issue.

gaborbernat:
Running works fine. Debugging though generates this extra attached output
which breaks the tests. Probably should reformulate to debugging instead
running.

gaborbernat:
Hello @JukkaL,

1. Many of the changes unrelated to this pr I've done to help me ease the development/setup. Once we want to merge it I can easily remove it. It already sits on its own individual comment exactly for this reason. So sure👌
2. If core members can think of edge cases not covered now I'm open to handle/add them even part of this pr.
3. I believe hidden behind a feature flag we can make it experimental, meaning we documentat it but don't mark it mature yet.
4. I'm committed to finishing and maintaining this. I want to use this feature extensively in my projects, so might as well own it throughout.
5. Sounds good.

As far as style goes whatever you guys want. For all new files I've used the excellent fully automated black formatter with isort import ordering mostly because I prefer spending my mental power on the implementation difficulties rather than how to indent/space my code. That being said I'm not married or attached to any style on my side. Ideally we would have this change automated but if not possible I can alter manually here and there.

gvanrossum:
A suggestion: if changes unrelated to this topic would benefit others
developing mypy, why not submit then as separate PRs? Once accepted you can
merge and they won't be part of this PR yet. (Of course these need to be
things that don't get in the way of using mypy or other development
workflows -- checking for environment variables might be useful here.)

gaborbernat:
@gvanrossum that would be great and I plan to do that. I don't completely understand what's the current development workflow used by core devs, so my changes don't get in way. Those can be probably discussed inside that pr though.

gvanrossum:
Mostly we don't use PyCharm much. :-)

gaborbernat:
#5189 now contains the changes unrelated to this PR.

gaborbernat:
So this is now in a review ready state I would say (all tests now pass). Before v0.1 I think I just need to add proper cache handling, some documentation on it. And we're good to go I would say. However, the underlying merge logic is now the proposed one so ready for review.

JukkaL:
I can't get this to work on the command line. I created a stub file `t.pyi` and the corresponding source file `t.py`, and another module `t2` that imports `t`. All of the files are in the same directory. No matter what I do, only either the stub or the source file of `t` seems to be processed. For example, I added a syntax error to `t.py` and type checked `t2` and `t` and there was no error:

```
$ mypy --merge-stub-into-src t2.py
<no syntax error>
$ mypy --merge-stub-into-src -m t
<no syntax error>
```

Similarly, if I add a syntax error to `t.pyi` and type check `t.py` no merging is happening:

```
$ mypy --merge-stub-into-src t.py
<no syntax error>
```

gaborbernat:
@JukkaL it does not work like that; only support now specifying the directory. The source finder logic does the collection of files, and then matching stub files with source files; and that's only used if you do ``mypy --merge-stub-into-src .``.

JukkaL:
Thanks for the hint, now I got this to merge some files. 

> That being said I believe, the way could work:

These all look reasonable. In addition, if I don't include a module explicitly on the command line and the module is found by following imports, we should perform a merge (this corresponds to the `mypy --merge-stub-into-src t2.py` case above).

I've found a couple of additional issues. First, if I define function `f` in the .py file but the .pyi file is empty, I can't import `f` from the module. I'd expect the function to be available (with a dynamically typed signature). I guess it might be reasonable to generate an error about a missing definition in the stub, but this might be impractical, since stubs generally don't include functions not included in the public interface of a module.

Second, the type of an imported name from a merged module is wrong, and it can cause crashes. Here is an example:

```
# p/t.py
def f(x): pass
```

```
# p/t.pyi
def f(x: int) -> None: ...
```

```
# p/__init__.py
```

```
# t2.py
from p.t import f
reveal_type(f)
```

If I type check the above example I get a crash:

```
$ mypy --show-traceback --merge-stub-into-src t2.py p
t2.py:2: error: Revealed type is 'def (x: int?) -> None?'
t2.py: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.620+dev-2d887dd02ce234e579a2087fb4b84a45ef43c0f2
  File "/Users/jukka/venv/mypy/bin/mypy", line 11, in <module>
    load_entry_point('mypy==0.620+dev.2d887dd02ce234e579a2087fb4b84a45ef43c0f2', 'console_scripts', 'mypy')()
  File "/Users/jukka/src/mypy/mypy/__main__.py", line 7, in console_entry
    main(None)
  File "/Users/jukka/src/mypy/mypy/main.py", line 91, in main
    res = type_check_only(sources, bin_dir, options, flush_errors, fscache)  # noqa
  File "/Users/jukka/src/mypy/mypy/main.py", line 148, in type_check_only
    fscache=fscache)
  File "/Users/jukka/src/mypy/mypy/build.py", line 188, in build
    flush_errors, fscache)
  File "/Users/jukka/src/mypy/mypy/build.py", line 293, in _build
  File "/Users/jukka/src/mypy/mypy/build.py", line 2468, in dispatch
    process_graph(graph, manager)
  File "/Users/jukka/src/mypy/mypy/build.py", line 2767, in process_graph
  File "/Users/jukka/src/mypy/mypy/build.py", line 2896, in process_stale_scc
    graph[id].finish_passes()
  File "/Users/jukka/src/mypy/mypy/build.py", line 2116, in finish_passes
    manager.report_file(self.tree, self.type_map(), self.options)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/contextlib.py", line 99, in __exit__
    self.gen.throw(type, value, traceback)
  File "/Users/jukka/src/mypy/mypy/build.py", line 1867, in wrap_context
    yield
  File "/Users/jukka/src/mypy/mypy/build.py", line 2111, in finish_passes
    self._patch_indirect_dependencies(self.type_checker().module_refs, self.type_map())
  File "/Users/jukka/src/mypy/mypy/build.py", line 2121, in _patch_indirect_dependencies
    types = set(type_map.values())
  File "/Users/jukka/src/mypy/mypy/types.py", line 906, in __hash__
    return hash((self.ret_type, self.is_type_obj(),
  File "/Users/jukka/src/mypy/mypy/types.py", line 781, in is_type_obj
    return self.fallback.type.is_metaclass()
AttributeError: 'NoneType' object has no attribute 'type'
t2.py: : note: use --pdb to drop into pdb
```

The likely reason for the crash is on the first line of output (`def (x: int?) -> None?`) -- the question marks mean that the type of `f` is not semantically analyzed, which should not happen. Likely the merged type is not updated correctly during merge, and we are leaking the pre-merge type of `f` in `p/t.py` during import.

gaborbernat:
Will try to tackle the raised issues.

mypy/stub_src_merge.py
```diff
@@ -0,0 +1,387 @@
...
+                return cast(Union[NameExpr, MemberExpr], l_value).name
+            else:
+                if report:
+                    self.report(
+                        "l-values must be simple name expressions, is {}".format(
+                            type(l_value).__qualname__
+                        )
+                    )
+        elif report:  # pragma: no cover
+            # TODO: how can we have more than one l-value in an assignment
```
ilevkivskyi: `x = y = 1`

Btw, what is the status of this PR? We will have a release next week, is this ready for a review?

gvanrossum:
This seems stalled...?

gaborbernat:
Actually I'm just planning to pick it up again. @Jukkal change requests are non trivial so this got put back on my schedule a bit.

gvanrossum:
OK, both Jukka and I are on vacation, so be patient...

(I tried to send this before but somehow the mail bounced.)

ethanhs:
@gaborbernat you can do:
```
MYPY=False
if MYPY:
    from typing import Type
```
And then string escape `Type` when you use it.

gaborbernat:
@JukkaL picking this up again, so what blockers we need to get this merged? :+1:

JukkaL:
Based on a quick review pass, these may be sufficient to get this merged:
* Fix remaining failing tests.
* Make style consistent with project style in previously existing files at least, and preferably everywhere (in particular, multi-line function signatures).
* Add at least one command line test that does an end-to-end source/stub merge (in `pythoneval.test`, probably).
* Get module discovery reviewed in detail, since it has changes to existing functionality. @msullivan may have the most context on this.
* Maybe refactor the new pytest-style tests to use unittest-style test classes, for consistency.
* One of the core team members should do some manual testing and a final round of review. Fix any big enough issues found.

It would also be good to have some commitment for additional work, since this is a pretty big feature. As discussed above, there will likely be various features and edge cases that would still need more work after this PR.

mypy/find_sources.py
```diff
@@ -19,53 +20,76 @@ class InvalidSourceList(Exception):
     """Exception indicating a problem in the list of sources given to mypy."""
 
 
-def create_source_list(files: Sequence[str], options: Options,
-                       fscache: Optional[FileSystemCache] = None,
-                       allow_empty_dir: bool = False) -> List[BuildSource]:
+def partition(
+    pred: Callable[[str], bool], iterable: Iterable[str]
+) -> Tuple[Iterable[str], Iterable[str]]:
```
JukkaL: We don't use this style for multi-line signatures. The original signature conforms to our style.

mypy/find_sources.py
```diff
@@ -19,53 +20,76 @@ class InvalidSourceList(Exception):
...
+    t1, t2 = tee(iterable)
+    return filterfalse(pred, t1), filter(pred, t2)
+
+
+def create_source_list(
+    files: Sequence[str],
+    options: Options,
+    fscache: Optional[FileSystemCache] = None,
+    allow_empty_dir: bool = False,
+) -> List[BuildSource]:
```
JukkaL: Another signature with inconsistent style.

mypy/test/test_source_finder.py
```diff
@@ -0,0 +1,145 @@
...
+        options.merge_stub_into_src = merge
+        test_dir = str(tmpdir)
+        create_files(Path(test_dir), content)
+        targets = create_source_list(files=[test_dir] if args is None else args, options=options)
+        return Path(str(tmpdir)), targets
+
+    return _checker
+
+
+def test_source_finder_merge(merge_finder: MergeFinder) -> None:
```
JukkaL: Can you put these tests inside a test class derived from `unittest.TestCase`, for consistency?

gaborbernat: yes, will do 👍

JukkaL:
Sorry, I've been busy with other things to finish the review. If you are still interested in finishing this up, I can dedicate some time to reviewing this in May. (Or maybe during PyCon sprints, in case you are coming there.)

gaborbernat:
@JukkaL yeah I still plan to finish it, will try to make it in some merge-able state by the sprint is my current plan (and yes I'll be there 😄 - I actually have a talk accepted about mypy 🤔 )

PR #5142, opened 2018-06-04
Move test_stubs and test_samples to pytest

elazarg:
Using the api, similarly to #5087.

`stdlibsamples` is not included in this PR since it requires special handling of working directory and paths.

This is another item for #1673.


ethanhs:
Hm, we will have to be careful with this and https://github.com/python/mypy/pull/5051. I think it should be fine for me to rebase my PR on this.

mypy/test/testsamples.py
```diff
@@ -0,0 +1,49 @@
...
+    def test_samples(self) -> None:
+        for f in find_files(os.path.join('test-data', 'samples'), suffix='.py'):
+            mypy_args = ['--no-strict-optional']
+            if f == os.path.join('test-data', 'samples', 'crawl2.py'):
+                # This test requires 3.5 for async functions
+                mypy_args.append('--python-version=3.5')
+            run_mypy(mypy_args + [f])
+
+
+def find_files(base: str, prefix: str = '', suffix: str = '') -> List[str]:
```
ethanhs: Perhaps these should be moved to `mypy.test.helpers` since they are used outside of this file?

elazarg: I don't think so. It's a temporary hack in `runtests.py`, and should be moved here soon.

ethanhs: Ah, that is true. I think this okay then.

msullivan:
Any reason not to merge this now?

gvanrossum:
@msullivan 
> Any reason not to merge this now?

I haven't had the time to review it. If you trust Ethan and/or you've reviewed (and preferably tested) it yourself, go right ahead!

ethanhs:
@elazarg It seems there is a merge conflict. When you fix the merge conflict please also delete 3.3 from the list of tested typeshed directories.

elazarg:
@ethanhs done (I've also removed 3.4)

mypy/test/testsamples.py
```diff
@@ -0,0 +1,49 @@
...
+
+class SamplesSuite(Suite):
+    def test_stubs(self) -> None:
+        # We only test each module in the one version mypy prefers to find.
+        # TODO: test stubs for other versions, especially Python 2 stubs.
+        seen = set()  # type: Set[str]
+        modules = []
+        # TODO: This should also test Python 2, and pass pyversion accordingly.
+        for version in ["2and3", "3", "3.5"]:
+            # FIX: remove 'builtins', this directory does not exist
```
gvanrossum: What does "FIX:" mean? Is it a TODO? Why not do it now?

elazarg: I guess you are right. I think I didn't want to change anything

Issue #5119, opened 2018-05-29
Tracking issue for overloads


Michael0x2a:
This is an issue summarizing all remaining TODOs for the overloads overhaul.

## Done

- [x] Add support for detecting partially-overlapping arities. 

    For example, disallow `f(A, A) -> int` and `f(*A) -> str`. This will likely require modifications to `is_callable_compatible` (or require implementing a new, similar method).

- [x] Improve union math semantics. Related issues:

    - [Support basic union math for overloaded functions](https://github.com/python/mypy/issues/4576)
    - [One more bug with union math logic](https://github.com/python/mypy/issues/5243)
    - [Wrong overload resolution with Any union item in argument](https://github.com/python/mypy/issues/5249) (?)

- [x] Update docs to clarify what the new overload semantics are. Related issues:

    - [Docs for function overloading are made more complex by using __getitem__ as example](https://github.com/python/mypy/issues/4579)

- [x] Improving how overload implementations are checked. Related issues:

    - [Errors when using overloaded functions and type variables](https://github.com/python/mypy/issues/4619)

- [x] Add full support for partially-overlapping types.

    For example, disallow `f(Union[A, B]) -> int` and `f(Union[B, C]) -> str`. We currently support detect partially-overlapping unions and tuples on just the top-level.

    Related issues:

    - [Allow overloading based on TypedDict structure](https://github.com/python/mypy/issues/3618)
    - [Overloaded function signatures 1 and 2 overlap with incompatible return types](https://github.com/python/mypy/issues/4020)
    - [Allow overloads to be distinguished by contained type of a generic](https://github.com/python/mypy/issues/4159)
    - [spurious "Need type annotation" error with overloads](https://github.com/python/mypy/issues/4227)

    Sub-todos:

    - [x] If all arguments in the first overload are partially overlapping with the second, report an error (even if the first return type is narrower).

        We should also probably have a custom error message, since this is technically safe when the first return type is narrower. This should also probably wait until we have more robust support for partially-overlapping types.

        **Edit**: We decided it wasn't worth complicating the code to handle this TODO.

## In-progress

- [x] Add union math for operators. See https://github.com/python/mypy/issues/2128.

- [ ] Error messages need a usability overhaul. Related open issues:

    - [Give better error messages when calling overloaded function with invalid types](https://github.com/python/mypy/issues/672)
    - [Ordering matters for @overload](https://github.com/python/mypy/issues/1270) (error message could be improved)
[...]

PR #5163, opened 2018-06-06
Add support for detecting overloads with overlapping arities

Michael0x2a:
This commit addresses TODO 2 from https://github.com/python/mypy/issues/5119 by adding support for detecting overloads with partially overlapping arities.

It also refactors the `is_callable_compatible` method. Specifically, this pull request...

1. Pulls out a lot of the logic for iterating over formal arguments into a helper method in CallableType.

2. Pulls out logic for handling varargs and kwargs outside of loops.

3. Rearranges some of the logic so we can return earlier slightly more frequently.


mypy/checker.py
```diff
@@ -3648,10 +3642,12 @@ def is_more_precise_or_partially_overlapping(t: Type, s: Type) -> bool:
     return (is_callable_compatible(signature, other,
                                    is_compat=is_more_precise_or_partially_overlapping,
                                    is_compat_return=lambda l, r: not is_subtype(l, r),
-                                   check_args_covariantly=True) or
+                                   check_args_covariantly=True,
+                                   allow_potential_compatibility=True) or
```
ilevkivskyi: TBH, I don't like the term "potential compatibility". Why not "partial overlap"? After all, IIUC, this is the same as `Union[A, B]` vs `Union[B, C]` but for arg counts.

mypy/subtypes.py
```diff
@@ -616,6 +617,30 @@ def g(x: int) -> int: ...
 
         In this case, the first call will succeed and the second will fail: f is a
         valid stand-in for g but not vice-versa.
+
+    allow_potential_compatibility:
```
ilevkivskyi: See my comment above about terminology.

mypy/subtypes.py
```diff
@@ -616,6 +617,30 @@ def g(x: int) -> int: ...
...
+        By default, this function returns True if and only if left is
+        definitely compatible with right.
```
ilevkivskyi: Instead of "definitely" maybe add an explanation in terms of subtyping? This shouldn't be mathematically precise (I could guess why you are mostly using "compatibility" instead of "subtyping", but for historical reasons in mypy code compatibility is checked by `is_subtype`, while subtyping by `is_proper_subtype`). Maybe say something like _each_ call that succeeds for right, will guaranteed to succeed for left callable.

mypy/subtypes.py
```diff
@@ -616,6 +617,30 @@ def g(x: int) -> int: ...
...
+
+        If this flag is set to True, we relax the checks so they return True
+        if the left is *potentially* compatible to right. For example, the
+        following two functions are normally incompatible:
+
+            f(x: int, y: str = "...", *args: bool) -> str
+            g(*args: int) -> int
+
+        However, they would be *potentially* compatible under certain conditions --
+        for example, if the user runs "f_or_g(3)". So, if this flag is
```
ilevkivskyi: Again, this will be probably more clear (at least for me), if you say that partial overlap means that _some_ calls that succeed for right, will also succeed for the left callable.

mypy/types.py
```diff
@@ -662,6 +662,8 @@ class CallableType(FunctionLike):
                  'min_args',  # Minimum number of arguments; derived from arg_kinds
                  'is_var_arg',  # Is it a varargs function?  Derived from arg_kinds
                  'is_kw_arg',  # Is it a **kwargs function?  Derived from arg_kinds
```
ilevkivskyi: Do we need both `is_kw_arg` and `kw_arg` etc. Maybe can make `is_kw_ars` a property that returns `kw_arg is not None`? `CallableType` has already loads of fields.

mypy/types.py
```diff
@@ -818,6 +837,30 @@ def max_possible_positional_args(self) -> int:
...
         return len([kind not in blacklist for kind in self.arg_kinds])
 
+    def formal_arguments(self, include_star_args: bool = False) -> Iterator[FormalArgument]:
+        """Yields the formal arguments corresponding to this callable, ignoring *arg and **kwargs.
+
+        To handle *args and **kwargs, use the 'callable.var_args' and 'callable.kw_args' fields,
+        if they are not None.
+
+        If you really want to include star args in the yielded output, set the
+        'include_star_args' field to 'True'."""
```
ilevkivskyi: set field -> pass argument

mypy/types.py
```diff
@@ -818,6 +837,30 @@ def max_possible_positional_args(self) -> int:
         blacklist = (ARG_NAMED, ARG_NAMED_OPT)
         return len([kind not in blacklist for kind in self.arg_kinds])
 
+    def formal_arguments(self, include_star_args: bool = False) -> Iterator[FormalArgument]:
```
ilevkivskyi: It is a good idea to make this a method.

test-data/unit/check-overloading.test
```diff
@@ -1718,9 +1716,217 @@ def foo1(*args: int) -> int: ...  # E: Overloaded function signatures 1 and 2 ov
 def foo1(**kwargs: int) -> str: ...
 
 @overload
-def foo2(**kwargs: int) -> str: ...
+def foo2(**kwargs: int) -> str: ...  # E: Overloaded function signatures 1 and 2 overlap with incompatible return types
 @overload
-def foo2(*args: int) -> int: ...  # E: Overloaded function signature 2 will never be matched: function 1's parameter type(s) are the same or broader
+def foo2(*args: int) -> int: ...
```
ilevkivskyi: I am actually not so sure now we need to make this an error. This is very similar to empty list situation, maybe we can just pick the first overload?

test-data/unit/check-overloading.test
```diff
@@ -1718,9 +1716,217 @@ def foo1(*args: int) -> int: ...  # E: Overloaded function signatures 1 and 2 ov
...
+
+@overload
+def foo1(*args: int) -> int: ...
+@overload
+def foo1(*args2: int) -> str: ...  # E: Overloaded function signature 2 will never be matched: signature 1's parameter type(s) are the same or broader
+
+@overload
+def foo2(*args: int) -> str: ...  # E: Overloaded function signatures 1 and 2 overlap with incompatible return types
+@overload
+def foo2(*args2: str) -> int: ...
```
ilevkivskyi: Same here, this is similar to empty list.

Michael0x2a:
1. I originally went with `allow_potential_compatibility` because I was thinking about this problem in terms of universal vs existential checks, not overlaps. The term "overlap" also, to me, implied that `is_callable_compatible` would be "symmetric" in a way that it wasn't when the param was set. E.g. the following two functions weren't previously equivalent:

        is_callable_compatible(f, g,
                               is_compat=some_symmetric_check,
                               allow_potential_compatibility=True)
        is_callable_compatible(g, f,
                               is_compat=some_symmetric_check,
                               allow_potential_compatibility=True)

    After some reflection though, I decided it would make the logic more consistent if I just went ahead and added that symmetry.
    
2.  I previously used the term "compatible" mainly because that was how I renamed the function: I previously refactored out all of the subtyping-related logic so that what it means for something to be "compatible" can be customized by the caller by passing in an appropriate `is_compat` function. That way, we could reuse the same logic to perform checks unrelated to subtyping.

    I don't think I was really aware that the term "compatibility" already meant something specific within mypy. So to avoid confusion, maybe I should rename `is_callable_compatible` and its params to not use the word "compatible"? I'm not sure what a good alternative name would be though -- let me know what you think.

3. Regarding the `f(*args)` thing -- I made the change you suggested.

mypy/types.py
```diff
@@ -818,6 +843,30 @@ def max_possible_positional_args(self) -> int:
...
+        """Yields the formal arguments corresponding to this callable, ignoring *arg and **kwargs.
+
+        To handle *args and **kwargs, use the 'callable.var_args' and 'callable.kw_args' fields,
+        if they are not None.
+
+        If you really want to include star args in the yielded output, set the
+        'include_star_args' field to 'True'."""
```
ilevkivskyi: > set field -> pass argument

It seems you missed this comment. Or you think "set field" is an OK formulation for functions?

mypy/subtypes.py
```diff
@@ -677,94 +738,92 @@ def g(x: int) -> int: ...
...
-                assert right_by_name is not None
-                if not are_args_compatible(left_by_name, right_by_name,
-                                           ignore_pos_arg_names, is_compat):
-                    return False
-            continue
 
-        # Left must have some kind of corresponding argument.
+    # Phase 1: Confirm every argument in R has a corresponding argument in L.
+
+    # Phase 1a: If right and right can both accept an infinite number of args,
```
ilevkivskyi: "right and right" -> "left and right"

test-data/unit/check-overloading.test
```diff
@@ -1412,15 +1412,19 @@ reveal_type(f(d))  # E: Revealed type is 'builtins.list[builtins.int]'
 from typing import overload, Any
 
 @overload
-def f(*, x: int = 3, y: int = 3) -> int: ...
+def f(*, x: int = 3, y: int = 3) -> int: ...  # E: Overloaded function signatures 1 and 2 overlap with incompatible return types
```
ilevkivskyi: Hm, I don't think this this should be an error. The only call that is allowed for both is `f()`.

Michael0x2a:
I could have sworn I fixed the field/parameter thing, but I guess not -- thanks for the (re)-catch!

Regarding the test case: I think I'm going to just merge this as-is. I did a bit of poking and I think it's possible to convert that error case back into a pass, but it'll be a bit kludgy. We basically need to special-case the first match an *arg or **kwarg has so we can pretend that args/kwargs is never length 0. I'm thinking we can add in this kludge later if it turns out it's impossible to type certain things w/ the current semantics.

I think this is also more or less consistent with how we handle non-kwarg/non-arg signatures -- for example, this PR considers the following overload to be unsafely overlapping.

    @overload
    def foo(*, x: str = '...', y: str = '...') -> str: ...
    @overload
    def foo(*, x: int = 3, y: int = 3) -> int: ...
    def foo(*args): 
        pass

I think it'd maybe be a bit surprising if swapping the last signature for a `**kwargs: int` were to suddenly make this the check.

Issue #5124, opened 2018-05-30
Bad inference for asyncio.gather with *args


JelleZijlstra:
On the following file:
```python
$ cat bin/asgat.py 
import asyncio

reveal_type(asyncio.gather(*[asyncio.sleep(1), asyncio.sleep(1)]))
```
I expect to get something like `Future[Tuple[Any, ...]]`, but actually with current master I get:
```
$ mypy bin/asgat.py 
bin/asgat.py:3: error: Revealed type is 'asyncio.futures.Future[Tuple[<nothing>]]'
```
(i.e., a single-element tuple).

The asyncio.gather stubs at https://github.com/python/typeshed/blob/master/stdlib/3.4/asyncio/tasks.pyi#L27 look right to me.

I found this in real code that was doing something like `a, b = await asyncio.gather(*[task1, task2])`.

cc @Michael0x2a

ilevkivskyi:
The `<nothing>` part that confused me first is unrelated, it is a known issue:
```python
reveal_type(asyncio.sleep(1))  # Future[<nothing>]
```
see https://github.com/python/mypy/issues/3032

The actual issue is better seen here,
```python
reveal_type(asyncio.gather(*[asyncio.sleep(1, 1.0), asyncio.sleep(1, 2.0)]))  # Future[Tuple[builtins.float*]]
```
A type is correct, but single item tuple is inferred instead of variable length one. I think the problem is that the first overload matches the star signature, because they are overlapping. There already a TODO item to prohibit such unsafe overlaps in https://github.com/python/mypy/issues/5119, second bullet, but I am not sure what should we actually write in typeshed in such cases. Maybe we need to re-think the second bullet. @Michael0x2a your opinion?

JelleZijlstra:
(To be clear this is a regression on master; it didn't happen in 0.590 or 0.600.)

ilevkivskyi:
> To be clear this is a regression on master

This is (kind of) expected. It was a necessary compromise, we decided to split overload overhaul in few PRs to avoid single mega-PR.

Michael0x2a:
I think this is a case of a previously silent bug becoming more visible. If you try running @ilevkivskyi's example on mypy 0.600, the revealed type will be just `Any`.

But my PR removes the "let's have overloads sometimes give up and return Any" behavior which, unfortunately, ends up breaking some code.

[...]

PR #5166, opened 2018-06-07
Refine how overload selection handles *args, **kwargs, and Any

Michael0x2a:
This pull request implements the changes discussed in https://github.com/python/mypy/issues/5124.

Specifically...

1. When two overload alternatives match due to Any, we return the last matching return type if it's a supertype of all of the previous ones. If it's not a supertype, we give up and return 'Any' as before.

2. If a user calls an overload with a starred expression, we try matching alternatives with a starred arg or kwarg first, even if those alternatives do not appear first in the list. If none of the starred alternatives are a valid match, we fall back to checking the other remaining alternatives in order.

### DIFF
--- a/test-data/unit/check-overloading.test
+++ b/test-data/unit/check-overloading.test
@@ -1312,7 +1312,8 @@ def f(x: object, y: int, z: str) -> object: ...
 def f(x): pass
 
 a: Any
-# Any causes ambiguity
+# Any causes ambiguity; we fall back to returning object since it's a
+# supertype of int
 reveal_type(f(a, 1, ''))  # E: Revealed type is 'builtins.object'
 # Any causes no ambiguity
 reveal_type(f(1, a, a))  # E: Revealed type is 'builtins.int'
@@ -1363,9 +1364,15 @@ def f(x: int, y: int, z: int) -> int: ...
 def f(x: object, y: int, z: int) -> object: ...
 def f(x): pass
 
+@overload
+def g(x: int, y: int, z: int) -> int: ...
+@overload
+def g(x: object, y: int, z: str) -> object: ...
+def g(x): pass
+
 a: Any
-# TODO: We could infer 'int' here
-reveal_type(f(1, *a))  # E: Revealed type is 'builtins.object'
+reveal_type(f(1, *a))  # E: Revealed type is 'builtins.int'
+reveal_type(g(1, *a))  # E: Revealed type is 'builtins.object'
 
 [case testOverloadWithOverlappingItemsAndAnyArgument8]
 from typing import overload, Any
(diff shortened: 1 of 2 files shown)


Michael0x2a:
This pull request should be orthogonal to my other pending one -- the two can be reviewed independently.

test-data/unit/check-overloading.test
```diff
@@ -1365,7 +1365,7 @@ def f(x): pass
 
 a: Any
 # TODO: We could infer 'int' here
```
ilevkivskyi: Hm, this TODO looks interesting. Would it be hard to fix?

test-data/unit/check-overloading.test
```diff
@@ -1723,6 +1743,105 @@ def foo2(**kwargs: int) -> str: ...
...
+class A: ...
+class B: ...
+class C: ...
+
+@overload
+def foo(x: int) -> A: ...
+@overload
+def foo(x: int, y: int) -> B: ...
+@overload
+def foo(x: int, y: int, z: int, *args: int) -> C: ...
```
ilevkivskyi: I would add a test with overload like this:
```python
@overload
def f(x: int) -> Tuple[int]: ...
@overload
def f(x: int, y: int) -> Tuple[int, int]: ...
@overload
def f(*xs: int) -> Tuple[int, ...]: ...
```
And check how it various calls with and without star.

test-data/unit/check-overloading.test
```diff
@@ -1723,6 +1782,127 @@ def foo2(**kwargs: int) -> str: ...
...
+def f(*args): pass
+
+i: int
+reveal_type(f(i))           # E: Revealed type is 'Tuple[builtins.int]'
+reveal_type(f(i, i))        # E: Revealed type is 'Tuple[builtins.int, builtins.int]'
+reveal_type(f(i, i, i))     # E: Revealed type is 'builtins.tuple[builtins.int]'
+
+reveal_type(f(*[]))         # E: Revealed type is 'builtins.tuple[builtins.int]'
+reveal_type(f(*[i]))        # E: Revealed type is 'builtins.tuple[builtins.int]'
+reveal_type(f(*[i, i]))     # E: Revealed type is 'builtins.tuple[builtins.int]'
```
ilevkivskyi: What happens if there is a tuple instead of list? Ideally `f(*(i, i))` should infer `Tuple[int, int]`. If this is the case, then I would add this call to this test.

Michael0x2a:
@ilevkivskyi -- I need to run to a thing; I'll check the tuple thing once I get back.

Michael0x2a:
@ilevkivskyi -- ok, I think this is ready for another round of review.

I discovered that `f(*(int, int))` was indeed not being inferred to `Tuple[int, int]`. I decided to fix this by just disabling the "let's move the alternative to the front" thing if the starred arg we're passing in is a type that has a "shape".

Also, I ran into an interesting edge case in the very last test I added (`testOverloadVarargsAndKwargsSelection`) -- do you have any thoughts about how we should handle that case?

test-data/unit/check-overloading.test
```diff
@@ -1723,6 +1782,235 @@ def foo2(**kwargs: int) -> str: ...
...
+a: Tuple[int, int]
+b: Tuple[int, ...]
+c: Dict[str, int]
+
+reveal_type(f(*a, **c))  # E: Revealed type is '__main__.A'
+reveal_type(f(*b, **c))  # E: Revealed type is '__main__.A'
+reveal_type(f(*a))       # E: Revealed type is '__main__.B'
+reveal_type(f(*b))       # E: Revealed type is 'Any'
+
+# TODO: Should this be 'Any' instead?
```
ilevkivskyi: I am OK with the current behaviour. This is exactly one of those cases where the rule "when in doubt -- take first" applies.

PR #5170, opened 2018-06-08
Reorganize and categorize command line flags

Michael0x2a:
This pull request cleans up the message displayed when running `mypy --help` by...

1. Modifying the flag descriptions to consistently start with lowercase letters.
2. Mildly tweaking a few flag descriptions.
3. Categorizing command line flags into distinct groups.
4. Reordering the flags.

Specifically, the flags now arranged in the following order:

1. General flags (`--help`, `--version`)
2. Config file flags
3. Imports handling
4. Flags that influence how code is typechecked/alter strictness
5. Flags that alter mypy's "state" (incremental flags and "internal" flags like `--custom-typeshed-dir`)
6. Reports and error messages
7. Misc flags

Note: this pull request does not change any of the flags themselves -- it only changes the output of `--help`.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -322,23 +322,26 @@ def process_options(args: List[str],
                             default: bool,
                             dest: Optional[str] = None,
                             help: str,
-                            strict_flag: bool = False
+                            strict_flag: bool = False,
+                            group: Optional[argparse._ActionsContainer] = None
                             ) -> None:
         if inverse is None:
             inverse = invert_flag_name(flag)
+        if group is None:
+            group = parser
 
         if help is not argparse.SUPPRESS:
             help += " (inverse: {})".format(inverse)
 
-        arg = parser.add_argument(flag,
-                                  action='store_false' if default else 'store_true',
-                                  dest=dest,
-                                  help=help)
+        arg = group.add_argument(flag,
+                                 action='store_false' if default else 'store_true',
+                                 dest=dest,
+                                 help=help)
         dest = arg.dest
-        arg = parser.add_argument(inverse,
-                                  action='store_true' if default else 'store_false',
-                                  dest=dest,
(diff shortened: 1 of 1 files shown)


Michael0x2a:
Also, a few misc notes:

1. This is what running `--help` looks like now: https://gist.github.com/Michael0x2a/36c5948a7ea571b722686226639b0859
2. The way I organized the flags is honestly a bit arbitrary -- if anybody can think of a better order/groupings, I'm down to change it.
3. I'm planning on reorganizing the command line/config file docs in a similar way in a future PR -- probably once I need another break from overloads :P
4. I was initially planning on having each flag description start with a capital letter because I thought it would make it easier for people to identify where each description starts while scanning the output. But apparently, that goes against argparse's conventions? Anyways, if people want to switch back to capital letters, let me know -- the only blocker is that I'd need to hijack the `--help` flag's description.

mypy/main.py
```diff
@@ -351,125 +354,249 @@ def add_invertible_flag(flag: str,
...
+    add_invertible_flag('--no-implicit-optional', default=False, strict_flag=True,
+                        help="don't assume arguments with default values of None are Optional",
+                        group=none_group)
+    none_group.add_argument(
+        '--strict-optional', action='store_true',
+        help=argparse.SUPPRESS)
+    none_group.add_argument(
+        '--no-strict-optional', action='store_false', dest='strict_optional',
+        help="disable strict Optional checks (inverse: --strict-optional)")
+    none_group.add_argument(
```
msullivan: We should kill this relic of not having per-module configuration

ethanhs: We should probably deprecate it for at least a release.

gvanrossum: [meta] When you comment on a multi-line thing please put your comment on the last line -- the GitHub email and Conversation tab only show code above the line you commented on.

PR #5172, opened 2018-06-08
Fix two option handling bugs in dmypy

### LINKED ISSUE #5168: dmypy issue with `mypy.ini`
In http://github.com/zulip/zulip/, we've been running `dmypy` with this commit from master: https://github.com/python/mypy.git@6a934d32124991a77ea433856cf2d62871ceb632#egg=mypy==0.610+dev.6a934d32124991a77ea433856cf2d62871ceb632

Generally, it's been great, but we've been prevented from turning it on by default because I've been seeing it throw occasional errors like this (generally shows up after a lot of rebasing; I switch branches dozens of times a day as part of merging things):

```
(zulip-py3-venv) tabbott@zaset:~/zulip$ ./tools/run-mypy -d
zproject/test_settings.py:22: error: Need type annotation for 'REALM_HOSTS'
zproject/test_settings.py:43: error: Incompatible types in assignment (expression has type "None", variable has type "str")
zproject/test_settings.py:61: error: Incompatible types in assignment (expression has type "Tuple[str, ...]", variable has type "Tuple[str, str, str, str]")
zproject/test_settings.py:80: error: Dict entry 2 has incompatible type "str": "int"; expected "str": "Union[Dict[str, int], str, None]"
zproject/test_settings.py:80: error: Dict entry 3 has incompatible type "str": "int"; expected "str": "Union[Dict[str, int], str, None]"
zproject/test_settings.py:152: error: Need type annotation for 'PUSH_NOTIFICATION_BOUNCER_URL'
```

but this shouldn't be happening, since we have this in our `mypy.ini`:

```
[mypy-zproject.test_settings]                                                                                                            
ignore_errors = True                                                                                                                     
```

I have it reproduced currently in my laptop, so if there's something I can do to dump data from `dmypy` or otherwise help debug, I'd love to do that.

As a sidenote, why `dmypy` instead of `mypyd`?  I feel like the latter is the traditional UNIX name :)

msullivan:
1. Make sure to use the right options when processing pass3 semanal in triggers
2. Don't throw away `ignore_errors` when processing triggers

Fixes #5168

### DIFF
--- a/mypy/server/update.py
+++ b/mypy/server/update.py
@@ -852,9 +852,6 @@ def reprocess_nodes(manager: BuildManager,
 
     nodes = sorted(nodeset, key=key)
 
-    # TODO: ignore_all argument to set_file_ignored_lines
-    manager.errors.set_file_ignored_lines(file_node.path, file_node.ignored_lines)
-
     targets = set()
     for node in nodes:
         target = target_from_node(module_id, node.node)
--- a/mypy/server/update.py
+++ b/mypy/server/update.py
@@ -852,6 +852,10 @@ def reprocess_nodes(manager: BuildManager,
 
     nodes = sorted(nodeset, key=key)
 
+    options = graph[module_id].options
+    manager.errors.set_file_ignored_lines(
+        file_node.path, file_node.ignored_lines, options.ignore_errors)
+
     targets = set()
     for node in nodes:
         target = target_from_node(module_id, node.node)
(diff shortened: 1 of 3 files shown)


mypy/server/update.py
```diff
@@ -852,9 +852,6 @@ def key(node: DeferredNode) -> int:
 
     nodes = sorted(nodeset, key=key)
 
-    # TODO: ignore_all argument to set_file_ignored_lines
-    manager.errors.set_file_ignored_lines(file_node.path, file_node.ignored_lines)
```
ilevkivskyi: It looks like this change breaks the `# type: ignore` in fine grained mode.

msullivan: Argh I thought I had run the tests >_>

PR #5173, opened 2018-06-08
Various updates to cheat sheets

JukkaL:
Use Python 3.6 variable annotation syntax since it looks nicer and
more modern. Make things generally more consistent.  Add a few
additional examples. Remove Python 3.6 section.

### DIFF
--- a/docs/source/cheat_sheet.rst
+++ b/docs/source/cheat_sheet.rst
@@ -21,32 +21,32 @@ Built-in types
 
    from typing import List, Set, Dict, Tuple, Text, Optional
 
-   # For simple built-in types, just use the name of the type.
-   x = 1 # type: int
-   x = 1.0 # type: float
-   x = True # type: bool
-   x = "test" # type: str
-   x = u"test" # type: unicode
+   # For simple built-in types, just use the name of the type
+   x = 1  # type: int
+   x = 1.0  # type: float
+   x = True  # type: bool
+   x = "test"  # type: str
+   x = u"test"  # type: unicode
 
    # For collections, the name of the type is capitalized, and the
-   # name of the type inside the collection is in brackets.
-   x = [1] # type: List[int]
-   x = set([6, 7]) # type: Set[int]
+   # name of the type inside the collection is in brackets
+   x = [1]  # type: List[int]
+   x = {6, 7}  # type: Set[int]
 
-   # For mappings, we need the types of both keys and values.
-   x = dict(field=2.0) # type: Dict[str, float]
+   # For mappings, we need the types of both keys and values
+   x = {'field': 2.0}  # type: Dict[str, float]
 
-   # For tuples, we specify the types of all the elements.
-   x = (3, "yes", 7.5) # type: Tuple[int, str, float]
+   # For tuples, we specify the types of all the elements
+   x = (3, "yes", 7.5)  # type: Tuple[int, str, float]
    # Empty Tuple types are a bit special
    x = ()  # type: Tuple[()]
 
-   # For textual data, use Text.
-   # This is "unicode" in Python 2 and "str" in Python 3.
-   x = ["string", u"unicode"] # type: List[Text]
+   # For textual data, use Text
+   # ("Text" means  "unicode" in Python 2 and "str" in Python 3)
+   x = ["string", u"unicode"]  # type: List[Text]
 
-   # Use Optional[] for values that could be None.
-   input_str = f() # type: Optional[str]
+   # Use Optional[] for values that could be None
+   input_str = f()  # type: Optional[str]
    if input_str is not None:
       print input_str
 
@@ -56,9 +56,9 @@ Functions
 
 .. code-block:: python
 
-   from typing import Callable, Iterable
+   from typing import Callable, Iterable, Union, Optional, List
 
-   # This is how you annotate a function definition.
+   # This is how you annotate a function definition
    def stringify(num):
        # type: (int) -> str
        """Your function docstring goes here after the type definition."""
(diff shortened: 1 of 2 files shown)


docs/source/cheat_sheet_py3.rst
```diff
@@ -235,113 +235,116 @@ Classes
...
+       while count > 0:
+           print('T-minus {} ({})'.format(count, tag))
+           await asyncio.sleep(0.1)
+           count -= 1
+       return "Blastoff!"
+
+   async def async2(obj: object) -> str:
+       return "placeholder"
+
+   # A generator-based coroutine created with @asyncio.coroutine should have a
```
ilevkivskyi: Python 3.4 EoL is in 9 months, also I doubt there are many Python 3.4 users out there, I would completely remove this part, and instead maybe add an example of async generator (3.6+).

docs/source/cheat_sheet_py3.rst
```diff
@@ -235,113 +235,116 @@ Classes
...
+
+   # A Python 3.5+ coroutine is typed like a normal function
+   async def countdown35(tag: str, count: int) -> str:
+       while count > 0:
+           print('T-minus {} ({})'.format(count, tag))
+           await asyncio.sleep(0.1)
+           count -= 1
+       return "Blastoff!"
+
+   async def async2(obj: object) -> str:
```
ilevkivskyi: I think this example is redundant, the above example makes it already clear.

docs/source/cheat_sheet.rst
```diff
@@ -108,67 +110,71 @@ Functions
...
+       # type: (*str, **str) -> str
+       request = make_request(*args, **kwargs)
+       return self.do_api_query(request)
 
-   # Use `ignore` to suppress type-checking on a given line, when your
-   # code confuses mypy or runs into an outright bug in mypy.
-   # Good practice is to comment every `ignore` with a bug link
+   # Use a "type: ignore" comment to suppress errors on a given line,
+   # when your code confuses mypy or runs into an outright bug in mypy.
+   # Good practice is to comment every "ignore" with a bug link
```
Michael0x2a: Up to you, but maybe we should mention the `--warn-unused-ignores` flag here (and in the Python 3 cheatsheet)?

I remember running into one or two criticisms of mypy that basically boiled down into people not knowing that this flag existed, so it might be worth surfacing it more.

(That said, I think it's also useful to keep the cheatsheets as short as possible, so that would be an argument against doing this.)

docs/source/cheat_sheet.rst
```diff
@@ -108,67 +110,71 @@ Functions
...
    # (in mypy, typeshed, or your own code) or an explanation of the issue.
    x = confusing_function() # type: ignore # https://github.com/python/mypy/issues/1167
 
-   # cast is a helper function for mypy that allows for guidance of how to convert types.
-   # it does not cast at runtime
+   # "cast" is a helper function that lets you override the inferred
+   # type of an expression. It's only for mypy -- there's no runtime check.
```
Michael0x2a: Similar thing with `--warn-redundant-casts`. We'd also probably need to add a brief summary of cast best practices though, similar to what we did up above?

docs/source/cheat_sheet.rst
```diff
@@ -206,43 +217,45 @@ Classes
...
    from typing import Match, AnyStr, IO
-   x = re.match(r'[0-9]+', "15") # type: Match[str]
 
-   # Use AnyStr for functions that should accept any kind of string
-   # without allowing different kinds of strings to mix.
+   # "typing.Match" describes regex matches from the re module
+   x = re.match(r'[0-9]+', "15")  # type: Match[str]
+
+   # You can use "AnyStr" to indicate that any string type will work,
+   # and the return type should match argument types
```
Michael0x2a: I think we should link to the docs for AnyStr or something here. I don't think that it's obvious AnyStr is supposed to be a TypeVar and that you're supposed to use it at least twice in a given signature, but these are probably things a user ought to know.

JukkaL: I decided to leave this section out since `AnyStr` is too hard to explain in a cheat sheet. Besides, most correct uses of `AnyStr` seem to be in library stubs.

docs/source/cheat_sheet.rst
```diff
@@ -21,32 +21,32 @@ Built-in types
...
+   x = (3, "yes", 7.5)  # type: Tuple[int, str, float]
    # Empty Tuple types are a bit special
    x = ()  # type: Tuple[()]
 
-   # For textual data, use Text.
-   # This is "unicode" in Python 2 and "str" in Python 3.
-   x = ["string", u"unicode"] # type: List[Text]
+   # For textual data, use Text
+   # ("Text" means  "unicode" in Python 2 and "str" in Python 3)
+   x = ["string", u"unicode"]  # type: List[Text]
```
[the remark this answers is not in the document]
JukkaL: Updated.

docs/source/cheat_sheet.rst
```diff
@@ -21,32 +21,32 @@ Built-in types
...
 
-   # Use Optional[] for values that could be None.
-   input_str = f() # type: Optional[str]
+   # Use Optional[] for values that could be None
+   input_str = f()  # type: Optional[str]
```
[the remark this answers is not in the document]
JukkaL: Renamed the function to another dummy name not defined anywhere.

docs/source/cheat_sheet.rst
JukkaL: This seems a bit too specialized for the cheat sheet.

docs/source/cheat_sheet.rst
JukkaL: Similar to above, this seems a bit too specialized. Also, `--warn-redundant-casts` is not without problems, since some casts might only make sense on a certain platform or certain Python version, so mypy thinks they are redundant even though they are useful in some contexts.

PR #5175, opened 2018-06-08
docs: fix typos

ceh:


ceh:
Perhaps cherry-pick this as well for 0.610 @msullivan?

PR #5183, opened 2018-06-09
Fix minor typos in comments

ceh:
Minor touchup of existing code comments.

### DIFF
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -269,7 +269,7 @@ class TypeList(Type):
         return visitor.visit_type_list(self)
 
     def serialize(self) -> JsonDict:
-        assert False, "Sythetic types don't serialize"
+        assert False, "Synthetic types don't serialize"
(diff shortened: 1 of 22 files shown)


mypy/types.py
```diff
@@ -2050,7 +2050,7 @@ def callable_type(fdef: mypy.nodes.FuncItem, fallback: Instance,
 
 
 def get_typ_args(tp: Type) -> List[Type]:
-    """Get all type arguments from a parameterizable Type."""
```
ethanhs: I found this quite interesting, as I always have written parameterizable. Apparently both are acceptable and used in literature, but the more accpted spelling is parametrizable.

ceh: I was not sure about this change, being a non-native speaker. I should have looked it up, because changing acceptable spellings is just noise. Thanks for this comment! 😸

PR #5185, opened 2018-06-09
Fix error recovery logic for daemon

ilevkivskyi:
It looks like indentation was wrong, because the try block would _always_ fail with an `UnboundLocalError` if a command is not a string.

As an observation, PEP 572 would prevent this bug.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -172,14 +172,14 @@ class Server:
                             resp = {'error': "Command is not a string"}
                         else:
                             command = data.pop('command')
-                        try:
-                            resp = self.run_command(command, data)
-                        except Exception:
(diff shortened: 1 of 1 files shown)


gvanrossum:
I don't see an UnboundLocalError when command is not a string -- I see a TypeError on line 211 (`key = 'cmd_' + command`). How would inline assignment have saved us here?

ilevkivskyi:
> I see a TypeError on line 211

Ah yes, you are right, somehow I missed that the assignment always happens.

> How would inline assignment have saved us here?

With the inline assignment, the code would be more likely written like this:
```python
if not isinstance(command := data.pop('command'), str):
     ...
else:
    ...
```
But then the else clause would be empty if the try block had the wrong indentation.

Issue #383, opened 2014-08-15
MyPy Logo


mvcisback:
Playing around with logos for another project.....this popped into my head

![mypy](https://cloud.githubusercontent.com/assets/388723/3936084/1df44cae-249d-11e4-8ece-1afc148c0125.png)

Its the normal Python Logo transformed to make an M for mypy

ilevkivskyi:
Here is another idea for the logo:

![mypy](https://user-images.githubusercontent.com/12005495/29493671-1c07a672-859b-11e7-89d1-eef0cf7c922b.png)

@JukkaL I think it is OK to use the Python logo if we have a permission. The "logic" behind this idea is quite simple, it looks like ``x: list[int]``, where ``x == Python`` in our case :-) Also I like simple logos that contain the actual title.

I took the colours (``#1f5082`` and ``#dfdfdf``) from CSS on [mypy-lang.org](http://mypy-lang.org), we could play with other colours, but I think this pair looks quite "solid". I also tried to play with font, it seems to me that monospace looks quite bad (especially letter ``m``), therefore I tried something wider, I think this looks better.

(I have an original SVG, but GitHub doesn't support embedding it in a post.)

gvanrossum:
I like this idea. Something seems wrong with the colors though -- when I click on the image I only see the blue parts. In general the gray part is way too light for my taste, it looks like a "grayed out" menu item.

ilevkivskyi:
Here is a version with a bit darker gray ``#b7b7b7`` (I also put it on a non-transparent white background):

![mypyb](https://user-images.githubusercontent.com/12005495/29498784-43ef5a3c-8603-11e7-97b4-2c29eb9bb01e.png)

gvanrossum:
It's a bit better, but the gray still makes me think "grayed out". Maybe we
need to ask a designer friend... (I could ask my designer brother to have a
look if you want to.)

ilevkivskyi:
> Maybe we need to ask a designer friend... (I could ask my designer brother to have a look if you want to.)

Yes, this is definitely a good idea.

ilevkivskyi:
Here is a balanced version (both colours now have the same lightness):

![mypyc](https://user-images.githubusercontent.com/12005495/29498974-10a607f8-8607-11e7-8bf5-3f7ae38e65bb.png)

gvanrossum:
@justvanrossum What do you think of this logo design? How would you improve it?

[...]

PR #5187, opened 2018-06-09
Add mypy logo to README

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/383

IMO it looks prettier with the logo. You can check how it will look [here](https://github.com/ilevkivskyi/mypy/tree/add-logo).


ilevkivskyi:
I just got an OK from Jukka, so I am merging this now.

PR #5189, opened 2018-06-10
support for PyCharm test debugging, and add tox environment setup

gaborbernat:
Most open source projects use tox as a way of automating project setup. It means the user can just invoke tox and that will setup automatically environments in what to develop and run tests. Additionally, PyCharm is a highly popular IDE, that allows users not that friendly to the command line to quickly become productive. 

This PR makes some changes to ensure that out of box clone of mypy allows running the tests from within PyCharm (and that they pass). Plus, it exposes targets that also run inside the CI locally (e.g. running tests for Python 3.4, 3.5, 3.6, 3.7). 

Once you have tox installed on a system or user level (``pip install tox``):

```bash
cd mypy
tox -av
default environments:
py34 -> run the test driver with python3.4
py35 -> run the test driver with python3.5
py36 -> run the test driver with python3.6
py37 -> run the test driver with python3.7
lint -> check the code style
type -> type check ourselves
docs -> invoke sphinx-build to build the HTML docs

additional environments:
dev  -> generate a DEV environment, that has all project libraries
```

E.g. to generate the documentation locally all one needs to do is ``tox -e docs``, and can then view it by opening the HTML up from a browser inside ``.tox/docs_out`` folder. Tox is kinda like make, but cross-platform (https://tox.readthedocs.org), written in Python and built-in isolation and virtualenv creation.

These are most of the changes, I've did to help me to develop the feature https://github.com/python/mypy/pull/5139. Ideally this would be merged before that, leaving the PR to contain the feature only.

### DIFF
--- a/mypy/test/testcmdline.py
+++ b/mypy/test/testcmdline.py
@@ -58,6 +58,14 @@ def test_python_cmdline(testcase: DataDrivenTestCase) -> None:
     outb = process.stdout.read()
     # Split output into lines.
     out = [s.rstrip('\n\r') for s in str(outb, 'utf8').splitlines()]
+
+    is_running_in_py_charm = "PYCHARM_HOSTED" in os.environ
+    if is_running_in_py_charm:
+        pos = next((p for p, i in enumerate(out) if i.startswith('pydev debugger: ')), None)
+        if pos is not None:
+            del out[pos]  # the attaching debugger message itself
+            del out[pos]  # plus the extra new line added
+
     result = process.wait()
     # Remove temp file.
     os.remove(program_path)
--- a/mypy/test/testcmdline.py
+++ b/mypy/test/testcmdline.py
@@ -59,8 +59,7 @@ def test_python_cmdline(testcase: DataDrivenTestCase) -> None:
     # Split output into lines.
     out = [s.rstrip('\n\r') for s in str(outb, 'utf8').splitlines()]
 
-    is_running_in_py_charm = "PYCHARM_HOSTED" in os.environ
-    if is_running_in_py_charm:
+    if "PYCHARM_HOSTED" in os.environ:
         pos = next((p for p, i in enumerate(out) if i.startswith('pydev debugger: ')), None)
         if pos is not None:
             del out[pos]  # the attaching debugger message itself
--- a/pytest.ini
+++ b/pytest.ini
@@ -19,5 +19,3 @@ python_files = test*.py
 python_classes =
 python_functions =
 
-# always run in parallel (requires pytest-xdist, see test-requirements.txt)
-addopts = -nauto --cov-append --cov-report=
--- a/pytest.ini
+++ b/pytest.ini
@@ -19,3 +19,6 @@ python_files = test*.py
 python_classes =
 python_functions =
 
+# always run in parallel (requires pytest-xdist, see test-requirements.txt)
+addopts = -nauto
+
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ envlist = py34,
 
 [testenv]
 description = run the test driver with {basepython}
-deps = -r ./test-requirements.txt
+deps = -rtest-requirements.txt
 commands = python runtests.py -x lint -x package {posargs}
 
 [testenv:lint]
@@ -27,14 +27,14 @@ commands = python runtests.py testselfcheck -p '-n0' -p '-v'
 [testenv:docs]
 description = invoke sphinx-build to build the HTML docs
 basepython = python3.6
-deps = -r docs/requirements-docs.txt
+deps = -rdocs/requirements-docs.txt
 commands = sphinx-build -d "{toxworkdir}/docs_doctree" docs/source "{toxworkdir}/docs_out" --color -W -bhtml {posargs}
 
 [testenv:dev]
 description = generate a DEV environment, that has all project libraries
 usedevelop = True
 basepython = python3.6
-deps = -r ./test-requirements.txt
-       -r docs/requirements-docs.txt
+deps = -rtest-requirements.txt
+       -rdocs/requirements-docs.txt
 commands = python -m pip list --format=columns
            python -c 'import sys; print(sys.executable)'
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,7 @@ envlist = py34,
 [testenv]
 description = run the test driver with {basepython}
 deps = -rtest-requirements.txt
-commands = python runtests.py -x lint -x package {posargs}
+commands = python runtests.py -x lint -x self-check {posargs}
 
 [testenv:lint]
 description = check the code style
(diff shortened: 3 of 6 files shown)


pytest.ini
```diff
@@ -19,5 +19,3 @@ python_files = test*.py
 python_classes =
 python_functions =
 
-# always run in parallel (requires pytest-xdist, see test-requirements.txt)
-addopts = -nauto --cov-append --cov-report=
```
gvanrossum: Why are you deleting this? It serves a real purpose for *our* typical workflow (running pytest from the command line). We're trying to get rid of runtests.py, so moving it there is not a good development IMO.

gaborbernat: @gvanrossum this is my proposal; feedback is welcome. I feel tox fills in a void that is not solved yet, so I'm not sure what type of configs are you talking about here yet; feel free to enlighten me! 

This config line contains two things set, which I disabled:
- run tests in parallel. Generally, I think tests should only be run in parallel when you want to run the entire test suite. E.g. becomes cumbersome to fix/check a single test with this. Shouldn't this be the other way around? That is single threaded by default, multi-threaded on demand?
- automatically enable coverage report. Again this seems unwarranted when running a single test (e.g. developing a new feature, writing a new test, debugging a test). Also once this is set PyCharm debugger cannot connect anymore, as both use the same mechanism to intercept calls.

gvanrossum:
Since you are in the minority (none of the other devs use PyCharm to
develop mypy AFAIK) don't you think it behooves you to find ways to turn
off the flags you don't want rather than changing the defaults for the rest
of us?

Honestly I don't care about the coverage, we turned that off in Travis
anyways because it's too slow. But I do care about the parallel option.

My problem with supporting tox is that surely once there is a tox file
we'll get PRs from others to change the tox flags and that just adds to the
review load.

Then again this argument also takes time so if one of the other devs speaks
up I'll gladly back off. :-)

ilevkivskyi:
> Then again this argument also takes time so if one of the other devs speaks
up I'll gladly back off. :-)

I am using PyCharm to develop mypy approximately half of time since beginning of this year. _But_ TBH I don't like these changes. IMO our top level directory has already too many config-related things, adding `tox.ini` will make it only worse. Also I don'y see much value in testing on many Python versions, the situations where something doesn't work on a given version are quite rare, and Travis will catch them (I have Travis enabled on my fork, so I see failures before even opening a PR).

gaborbernat:
I had the impression people use the test runner. I made the change so that
with the test runner nothing changes, moved over the config args. I did not
knew people use pytest on its own.

gaborbernat:
Oh and both setup.cfg and pytest.ini can be merged into tox.ini. And soon moved into pyproject.toml file. To help with clutter of files at root level.

tox.ini
```diff
@@ -0,0 +1,40 @@
...
+          py36,
+          py37,
+          lint,
+          type,
+          docs
+
+[testenv]
+description = run the test driver with {basepython}
+deps = -r ./test-requirements.txt
+commands = python runtests.py -x lint -x package {posargs}
```
ethanhs: package has been replaced with self-check.

tox.ini
```diff
@@ -0,0 +1,40 @@
...
+
+[testenv:lint]
+description = check the code style
+basepython = python3.6
+commands = python runtests.py lint {posargs}
+
+[testenv:type]
+description = type check ourselves
+basepython = python3.6
+commands = python runtests.py testselfcheck -p '-n0' -p '-v'
```
ethanhs: `self-check` would be more consistent here.

mypy/test/testcmdline.py
```diff
@@ -58,6 +58,14 @@ def test_python_cmdline(testcase: DataDrivenTestCase) -> None:
     outb = process.stdout.read()
     # Split output into lines.
     out = [s.rstrip('\n\r') for s in str(outb, 'utf8').splitlines()]
+
+    is_running_in_py_charm = "PYCHARM_HOSTED" in os.environ
+    if is_running_in_py_charm:
```
ethanhs: I don't think you need to create a variable here. I'd just do `if "PYCHARM_HOSTED" in os.environ: ...`

gaborbernat:
@ethanhs addressed all requests

gaborbernat:
@ilevkivskyi 
> Also I don'y see much value in testing on many Python versions, the situations where something doesn't work on a given version are quite rare, and Travis will catch them (I have Travis enabled on my fork, so I see failures before even opening a PR).

I don't run all tests either locally. However, when Travis catches something often (especially when you're new to the code base) it helps a lot to be able to reproduce locally quickly, debug it, understand why the problem arises and being able to fix/test it locally. I suppose at the end of the day it flows into the category of different workflow between us. I had a few such instances when creating #5139. 

That being said, I don't want to propose anything that breaks the current core developers workflow. I'm just suggesting to extend things that potentially help new people interested in contributing.

pytest.ini
ethanhs: I think it would be better to document that the tests run in parallel, and how to disable that (I think setting `$PYTEST_ADDOPTS to -n0` does it). Though I do agree that you probably do not want to run coverage in the general case.

gaborbernat:
I've restored the parallel execution of the test suite via ``pytest.ini`` and added a section on the wiki about options on how to make PyCharm play nicely with mypy (https://github.com/python/mypy/wiki/Getting-Started-Developing#pycharm). Is this more agreeable for the core developers? 

I still propose keeping the ``tox.ini`` file around. That could replace the long make files from within ``docs``. E.g. all users need to generate documentation is ``pip install --user tox && tox -e docs`` (which then will go ahead setup a valid environment and generate the documentation).

ethanhs:
Taking another look at this now.

gaborbernat:
cool :sunglasses: is anyone responsible with merging it?

ethanhs:
I think I will defer to @ilevkivskyi since he also uses PyCharm on occasion, and this could affect his workflow.

mypy/test/testcmdline.py
```diff
@@ -58,6 +58,13 @@ def test_python_cmdline(testcase: DataDrivenTestCase) -> None:
     outb = process.stdout.read()
     # Split output into lines.
     out = [s.rstrip('\n\r') for s in str(outb, 'utf8').splitlines()]
+
+    if "PYCHARM_HOSTED" in os.environ:
+        pos = next((p for p, i in enumerate(out) if i.startswith('pydev debugger: ')), None)
+        if pos is not None:
+            del out[pos]  # the attaching debugger message itself
+            del out[pos]  # plus the extra new line added
```
ilevkivskyi: This code looks quite fragile (what if the debugger output format changes?). Is it possible to configure the debugger to just not print these lines?

ilevkivskyi: Also please don't rebase your PR, merge when needed, otherwise it is hard to follow.

gaborbernat: we'll need to follow up then and amend it; I haven't found a way to disable it sadly :cry:

ilevkivskyi: OK.

PR #5197, opened 2018-06-11
Add mypy badge to README

ilevkivskyi:


ilevkivskyi:
I am not sure if we need a separate section showing the badge code verbatim. One can just click on raw version of `README` and copy the code.

PR #5200, opened 2018-06-11
Sync typeshed

gvanrossum:
Needed to get the tests to pass again now that https://github.com/python/typeshed/pull/2128 has landed.


gvanrossum:
Abandoning, we need a newer typeshed due to some issue.

PR #5201, opened 2018-06-11
Sync typeshed

gvanrossum:
Needed for https://github.com/python/mypy/pull/5046, now that https://github.com/python/typeshed/pull/2128 is merged.


gvanrossum:
I hope this magically makes the tests pass.

gvanrossum:
It seems best to combine this with #5046 in a single PR. I'll do that now.

PR #5202, opened 2018-06-11
Combine #5046 with typeshed sync

gvanrossum:
Needed after python/typeshed#2128; typeshed tests will fail until this is merged.


gvanrossum:
Thanks!

PR #5207, opened 2018-06-12
Add a pytest flag to allow saving temp directories from failed tests

msullivan:
This will be helpful for debugging bugs in cache artifacts.
It will also be helpful for experimenting with the generated
code in mypyc.

### DIFF
--- a/mypy/test/data.py
+++ b/mypy/test/data.py
@@ -249,7 +249,19 @@ class DataDrivenTestCase(pytest.Item):  # type: ignore  # inheriting from Any
             pytest.skip()
         suite = self.parent.obj()
         suite.setup()
-        suite.run_case(self)
+        try:
+            suite.run_case(self)
+        except Exception:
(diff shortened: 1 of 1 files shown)


mypy/test/data.py
```diff
@@ -249,7 +249,19 @@ def runtest(self) -> None:
             pytest.skip()
         suite = self.parent.obj()
         suite.setup()
-        suite.run_case(self)
+        try:
+            suite.run_case(self)
+        except Exception:
+            # As a debugging aid, support copying the contents of the tmp directory somewhere
+            target_dir = self.config.getoption('--save-failures-to', None)  # type: Optional[str]
+            print(target_dir)
```
gvanrossum: Is this a debug print? Maybe you want to print a nicely formatted error showing which test's temp dir was saved where instead?

msullivan: Made it do a nice print

Issue #5205, opened 2018-06-12
Support for pyproject.toml


pslacerda:
PEP518 says that `pyproject.toml` is to support the minimum build requirements. I want to consider mypy as part of the minimum build on my package.

For example Pip uses a `[tool.towncrier]` section on `pyproject.toml` to generate a correctly updated `NEWS.rst` as part of its build. One can considers typing correctness as important to the distributed source pakcage as an well documented `NEWS.rst`. Other example is Black that also has it section on `pyproject.toml`.

Maybe you think that this is a silly feature request and that `setup.cfg` must be used instead, Pip or Flit or whatever can setup mypy after all. But an TOML module will be added to the standard library sooner or later, than could mypy accept this feature and antecipate TOML supporting `pyproject.toml` today?

ilevkivskyi:
PEP 518 is accepted and specifies that `pyproject.toml` is preferred over `setup.cfg`, so I think we should support whatever the PEP specifies.

gvanrossum:
Does adding pyproject.toml mean that we can remove other config files?
Until that point I'm not excited.

pslacerda:
I think that removing other config files would simplify config parsing even more. I would be very glad to change `mypy.main.parse_config_file()` accordingly.

If you agree that `mypy==0.620` it is still in its early ages, would be exciting to remove the other config files. Otherwise I'm already coding the unexciting changes to support both legacy and `pyproject.toml`.

There is already lots of deprecated options:

```
--disallow-any
--strict-boolean
--dirty-stub
--use-python-pat
--silent-import
--almost-silen
--fast-parser
--no-fast-parser
``` 

So legacy config files (`mypy.ini`, `setup.cfg` and `~/.mypy.ini`) could also be one of them.

gvanrossum:
Note that mypy.ini serves a completely different purpose from setup.py, setup.cfg, and pyproject.toml. The mypy.ini file is read by mypy when it is about to type-check another program. The other files (IIUC) are used to build mypy itself. I don't think we can change mypy.ini much, though we *can* start removing deprecated flags (after they have been printing a warning for 1-2 release cycles). Please do that in a separate PR.

pslacerda:
Even if `setup.cfg` was created originally for distutils or setuptools, everybody else seems to use it as an generic configuration file. PyTest for example don't care if is running from `setup.py` or whatever, it simply reads `setup.cfg` if it is found; many tools seems to do it (eg flake8, isort and even mypy).

[...]

PR #5208, opened 2018-06-12
Add support to pyproject.toml

pslacerda:
Support for TOML configuration files with sections like `[tool.mypy]` and `[tool.mypy-mypackage.*]`.

Related to #5205.

### DIFF
--- a/mypy/config_parser.py
+++ b/mypy/config_parser.py
@@ -171,12 +171,12 @@ def parse_toml_config_file(options: Options, set_strict_flags: Callable[[], None
     else:
         options.config_file = filename
 
-    if 'mypy' not in table:
-        print("%s: No [mypy] table in config file" % filename, file=stderr)
+    if 'tool' not in table or 'mypy' not in table['tool']:
+        print("%s: No 'tool.mypy' table in config file" % filename, file=stderr)
         return False
 
     # Handle the mypy table.
-    for key, value in table['mypy'].items():
+    for key, value in table['tool']['mypy'].items():
 
         # Is an option.
         if key != 'overrides':
--- a/mypy/config_parser.py
+++ b/mypy/config_parser.py
@@ -138,20 +138,22 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
     if filename is not None:
         filename = os.path.expanduser(filename)
         if os.path.splitext(filename)[1] == '.toml':
-            parse_toml_config_file(options, set_strict_flags, filename, stdout, stderr)
+            parse_toml_config_file(
+                options, set_strict_flags, filename, stdout, stderr, explicit=True)
         else:
-            parse_ini_config_file(options, set_strict_flags, filename, stdout, stderr)
+            parse_ini_config_file(
+                options, set_strict_flags, filename, stdout, stderr, explicit=True)
     else:
         for filename in defaults.CONFIG_FILES:
             filename = os.path.expanduser(filename)
             if not os.path.isfile(filename):
                 continue
             if os.path.splitext(filename)[1] == '.toml':
-                parsed = parse_toml_config_file(options, set_strict_flags,
-                                                filename, stdout, stderr)
+                parsed = parse_toml_config_file(
+                    options, set_strict_flags, filename, stdout, stderr, explicit=False)
             else:
-                parsed = parse_ini_config_file(options, set_strict_flags,
-                                               filename, stdout, stderr)
+                parsed = parse_ini_config_file(
+                    options, set_strict_flags, filename, stdout, stderr, explicit=False)
             if parsed:
                 break
 
@@ -159,7 +161,9 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
 def parse_toml_config_file(options: Options, set_strict_flags: Callable[[], None],
                            filename: str,
                            stdout: Optional[TextIO] = None,
-                           stderr: Optional[TextIO] = None) -> bool:
+                           stderr: Optional[TextIO] = None,
+                           *,
+                           explicit: bool) -> bool:
     stderr = stderr or sys.stderr
 
     # Load the toml config file.
@@ -172,7 +176,8 @@ def parse_toml_config_file(options: Options, set_strict_flags: Callable[[], None
         options.config_file = filename
 
     if 'tool' not in table or 'mypy' not in table['tool']:
-        print("%s: No 'tool.mypy' table in config file" % filename, file=stderr)
+        if explicit:
+            print("%s: No 'tool.mypy' table in config file" % filename, file=stderr)
         return False
 
     # Handle the mypy table.
@@ -228,7 +233,9 @@ def parse_toml_config_file(options: Options, set_strict_flags: Callable[[], None
 def parse_ini_config_file(options: Options, set_strict_flags: Callable[[], None],
                           filename: str,
                           stdout: Optional[TextIO] = None,
-                          stderr: Optional[TextIO] = None) -> bool:
+                          stderr: Optional[TextIO] = None,
+                          *,
+                          explicit: bool) -> bool:
     stderr = stderr or sys.stderr
     parser = configparser.RawConfigParser()
     retv = False
@@ -245,7 +252,7 @@ def parse_ini_config_file(options: Options, set_strict_flags: Callable[[], None]
             os.path.abspath(filename))
 
     if 'mypy' not in parser:
-        if filename not in defaults.SHARED_CONFIG_FILES:
+        if not explicit and filename not in defaults.SHARED_CONFIG_FILES:
             print("%s: No [mypy] section in config file" % filename, file=stderr)
     else:
         retv = True
--- a/mypy/config_parser.py
+++ b/mypy/config_parser.py
@@ -6,7 +6,8 @@ import os
 import re
 import sys
 
-from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple, TextIO
+import toml
+from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple, TextIO, MutableMapping
 from typing_extensions import Final
 
 from mypy import defaults
@@ -50,9 +51,20 @@ def split_and_match_files(paths: str) -> List[str]:
 
     Returns a list of file paths
     """
+
+    return match_files(paths.split(','))
+
+
+def match_files(paths: List[str]) -> List[str]:
+    """Take list of files/directories (with support for globbing through the glob library).
+
+    Where a path/glob matches no file, we still include the raw path in the resulting list.
+
+    Returns a list of file paths
+    """
     expanded_paths = []
 
-    for path in paths.split(','):
+    for path in paths:
         path = expand_path(path.strip())
         globbed_files = fileglob.glob(path, recursive=True)
         if globbed_files:
@@ -77,7 +89,7 @@ def check_follow_imports(choice: str) -> str:
 # sufficient, and we don't have to do anything here.  This table
 # exists to specify types for values initialized to None or container
 # types.
-config_types = {
+ini_type_converters = {
     'python_version': parse_version,
     'strict_optional_whitelist': lambda s: s.split(),
     'custom_typing_module': str,
@@ -103,6 +115,16 @@ config_types = {
 }  # type: Final
 
 
+toml_type_converters = {
+    'python_version': parse_version,
+    'custom_typeshed_dir': expand_path,
+    'mypy_path': lambda l: [expand_path(p) for p in l],
+    'files': match_files,
+    'cache_dir': expand_path,
+    'python_executable': expand_path,
+}  # type: Final
+
+
 def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
                       filename: Optional[str],
                       stdout: Optional[TextIO] = None,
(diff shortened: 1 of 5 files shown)


pslacerda:
All these config files are in fact very much the same, except for the .toml difference. Maybe they all could be combined into a single variable:

```python
CONFIG_FILES = ('pyproject.toml', 'mypy.ini', 'setup.cfg', '~/.mypy.ini')
```

Sorry for the `# type: ignore`, I'm working on it.

Nested TOML tables/sections (like `[tool][mypy-package][module]` == `[tool.mypy-package.module]`) need to be flatten. https://github.com/python/mypy/pull/5208/files#diff-b5e3909c86f7834f70af20bb848acec3R910

Dict that were a SectionProxy that is a Dict again, that is somewhat dirty. Maybe change everything to `.toml` would be saner. https://github.com/python/mypy/pull/5208/files#diff-b5e3909c86f7834f70af20bb848acec3R1008

That library could be packed together. I suppose the authour would be very happy. Or another can also be made or found. https://github.com/python/mypy/pull/5208/files#diff-2eeaed663bd0d25b7e608891384b7298R107

To test I just changed some existing ones.

ethanhs:
It seems the tests timed out or crashed. I'm going to close/reopen to try them one more time. If they still fail @pslacerda can you look into why they are failing?

pslacerda:
I have no idea why it is timing out. Seems really unrelated to this modifications regards functionallity, but related to testing probably. I will look into this soon.

gvanrossum:
Honestly, we should never have caved in to supporting `setup.cfg`, then this whole thing would not have been necessary. It seems just busywork. Unlike the original motivation given in PEP 516 (which is to specify setup-time dependencies), there's nothing wrong with mypy.ini -- this PR simply does not solve a problem we have.

I did say that just submitting the PR wouldn't mean I'd accept it, and having thought it over more I am just going to reject this PR. Sorry!

touilleMan:
Hi,

I'm not complaining about this PR not being merged, but it has raised a question about python project organization:

I would have considered good practice to keep all tool configuration into a single file (keep the root directory cleaner, allow to understand project tooling by just looking at a single place) be it `setup.cfg` in the past and now this fancy `pyproject.toml` ;-)
However you seem to consider supporting configuration from `setup.cfg` was also a mistake, then it would mean good practice for project would be to a configuration file for each tool and avoid setup.cfg totally and use `pyproject.toml` only for setup-time dependency resolution and project building ?

gvanrossum:
Have you seen this? https://xkcd.com/927/

pslacerda:
But sure, jokes aside, there is only setup.cfg and pyproject.toml competing, even if setup.cfg don't really compete, in fact the problem here is just the oposite: lack of standarts.

PgLoLo:
Hi,

I am complaining about this PR not being merged. Absence of pyproject.toml support is just nonsense.

flying-sheep:
Since the corresponding issue #5205 has been accepted, can we reopen this?

JukkaL:
I've now reopened this, but since it's quite a while since this was created, there are a bunch of conflicts that need to be resolved.

pslacerda:
I think it is done but I'm unsure how to test. It was tested with the file that follows and seems to work well. If looks ok I do more rigorous tests. 

```toml
[mypy]
allow_redefinition = true
python_version = '2.7'

[mypy."package.*.a"]
python_version = '3.9'
allow_untyped_globals = true
```

pslacerda:
Maybe instead of deprecate mypy.ini and setup.cfg it should have the following priority.

1. pyproject.toml
2. $XDG_CONFIG_HOME/mypy.toml
3. $HOME/.mypy.toml
4. mypy.ini
5. setup.cfg
6. $XDG_CONFIG_HOME/mypy.ini
7. $HOME/.mypy.ini

layday:
(Though I don't regularly contribute to this project I thought I'd chime in here - let me know if I'm out of line.)

You're not going to want to rejig the priority list if you want this to be backwards compatible - keep everything the way it is now with the addition of pyproject.toml either above or below setup.cfg, i.e.:

1. --config-file
1. $PWD/mypy.ini
1. $PWD/setup.cfg
1. $PWD/pyproject.toml
1. $XDG_CONFIG_HOME/mypy/config
1. $HOME/.config/mypy/config
1. $HOME/.mypy.ini

> `[mypy."package.*.a"]`

You should use a sub-table for per-module configuration - the way it's set up now would have options potentially conflict with module names. To illustrate:

```
>>> toml.loads('''\
... [mypy]
... python_version = "2.7"
...
... [mypy."python_version"]
... allow_untyped_globals = true
... ''')
---------------------------------------------------------------------------
TomlDecodeError                           Traceback (most recent call last)
  ...
TomlDecodeError: What? python_version already exists?{'python_version': '2.7'} (line 4 column 1 char 31)
```

You can work around this easily by defining per-module configuration in e.g. `mypy.overrides`:

```toml
[mypy]
allow_redefinition = true
python_version = "2.7"

[mypy.overrides."package.*.a"]
python_version = "3.9"
allow_untyped_globals = true
```

pslacerda:
Its sunday and I'm specially lay after lunch.

I missed that. Is better `[mypy.overrides."package.*.a"]` or the following is doable?

```toml
[mypy]
allow_untyped_globals = false
[mypy."@python_version"]
allow_untyped_globals = true
```

layday:
Well, TOML has namespacing built into the language, so why would you resort to prefixing modules with at signs?  `.overrides` is easier for users to grok - the at sign is opaque - and easier for developers to operate on in Python.

pslacerda:
I think that it is done for now because `python runtests.py` say it is ok.

Maybe change some tests to use `pyproject.toml` could help.

mypy/config_parser.py
```diff
@@ -88,6 +100,16 @@ def split_and_match_files(paths: str) -> List[str]:
 }  # type: Final
 
 
+toml_type_conversors = {
```
ethanhs: I believe you meant `toml_type_converters` here?

mypy/config_parser.py
```diff
@@ -67,7 +79,7 @@ def split_and_match_files(paths: str) -> List[str]:
 # sufficient, and we don't have to do anything here.  This table
 # exists to specify types for values initialized to None or container
 # types.
-config_types = {
+ini_config_types = {
```
ethanhs: I would change this to `ini_type_converters` or something like that.

mypy/defaults.py
```diff
@@ -6,13 +6,18 @@
 PYTHON3_VERSION = (3, 6)  # type: Final
 PYTHON3_VERSION_MIN = (3, 4)  # type: Final
 CACHE_DIR = '.mypy_cache'  # type: Final
-CONFIG_FILE = 'mypy.ini'  # type: Final
+INI_CONFIG_FILE = 'mypy.ini'  # type: Final
 SHARED_CONFIG_FILES = ['setup.cfg', ]  # type: Final
 USER_CONFIG_FILES = ['~/.config/mypy/config', '~/.mypy.ini', ]  # type: Final
 if os.environ.get('XDG_CONFIG_HOME'):
     USER_CONFIG_FILES.insert(0, os.path.join(os.environ['XDG_CONFIG_HOME'], 'mypy/config'))
+TOML_CONFIG_FILE = 'pyproject.toml'
```
ethanhs: You can add a `# type: Final` annotation here (I believe this will help mypyc optimize things more, and improve correctness of course :^) ).

mypy/config_parser.py
```diff
@@ -97,45 +119,123 @@ def parse_config_file(options: Options, filename: Optional[str],
 
     If filename is None, fall back to default config files.
     """
-    stdout = stdout or sys.stdout
+    if filename:
```
ethanhs:
```suggestion
    if filename is not None:
```

pslacerda:
> One more nit, and I have several test suggestions:

One more nigth of my life hunting nits off my hair.

> 
>     * a test that uses multiple options
> 
>     * a test for the type converters (this can probably be put in one test)
> 

Ok, I got it.

>     * a test for a missing mypy section

This one is pickier. Should an empty `mypy.ini` file (or with no `mypy` section) be skipped like any other config file?

johnthagen:
FWIW, [pytest has recently put pyproject.toml configuration on their next milestone](https://github.com/pytest-dev/pytest/issues/1556#issuecomment-572267167)

mypy/config_parser.py
```diff
@@ -88,6 +100,16 @@ def split_and_match_files(paths: str) -> List[str]:
 }  # type: Final
 
 
+toml_type_converters = {
```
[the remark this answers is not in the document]
ethanhs: Don't worry about it. I don't think it is required after all.

test-data/unit/cmdline.test
```diff
@@ -170,23 +170,34 @@ def f():
...
+[file pyproject.toml]
 \[mypy]
-warn_unused_ignores = True
+warn_unused_ignores = true
+[file main.py]
+# type: ignore
+[out]
+main.py:1: error: unused 'type: ignore' comment
+
+[case testMissingMypySection]
```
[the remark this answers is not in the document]
ethanhs: Hm, I think it may be best not to do this. It would be annoying to have mypy fail all of a sudden on projects with a pyproject.toml just because they don't have a mypy section.

test-data/unit/cmdline.test
```diff
@@ -170,23 +170,34 @@ def f():
...
+# cmd: mypy main.py --config-file=pyproject.toml
+[file pyproject.toml]
 [file main.py]
 # type: ignore
+[out]
+pyproject.toml: No [mypy] table in config file
+== Return code: 0
 
 [case testPerFileConfigSection]
```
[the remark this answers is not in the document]
ethanhs: I would prefer you create a separate test here if possible.

ethanhs:
@pslacerda thank you again for working on this, I have a question about the toml configuration:

Why `mypy.overides.*` as opposed to `mypy.*` for the table names for per-module configuration?

Sorry for the sporadic review, I really want to see this merged soon, and I don't think there is much more to do here.

pslacerda:
@ethanhs in case someone has a module with the same name of an option it is necessary. For instance a module named `python_version` without the `mypy.override.*` would clash in `mypy.python_version`.

An otherway would could be abusing the syntax, eg. `mypy."@python_version"`. But it would require post-processing after the TOML parser, not very friendly.

@layday figured it out.

erikkemperman:
@pslacerda First of all, great work!

I took the liberty of looking into the merge conflicts, let me know if you're interested in how I resolved them. (There is a branch with all tests passing [here](https://github.com/erikkemperman/mypy/commits/pep518-rebase) please feel free to use it if you want!)

Couple of questions though:

- If I understand this PR correctly, it will look for `[mypy]` sections in pyproject.toml -- shouldn't that be `[tool.mypy]` according to the PEP?
- The INI handler does some magic such as boolean option inversion (e.g. transform between allow-x  disallow-x, x and no-x). Maybe I'm missing something, but it seems from looking at the PR superficially that the TOML handler does not do these things?
- To be really rigorous, should there be repeated tests for all the ones that currently use --config-file tmp/mypy.ini but pointing to tmp/pyproject.toml?

pslacerda:
Hi @erikkemperman!

- You are correct about `[tool.mypy]`, my fail.
- The TOML handler doesn't invert booleans TOML is more typed, there is a boolean type. So I deliberately avoided the confusion. Not logically reasonable but more sane anyway.
- To be rigorous it would really be needed to replicate all tests, but the implementation is the same for both handlers. The code coverage will be about the same, replicating or not.

I'll continue the work from your branch. Thank you!

erikkemperman:
> I'll continue the work from your branch

I hope you don't mind that I squashed your commits into a single one... It was just easier to rebase onto the latest upstream master that way. I took care to retain your authorship though.

About the boolean inversions, sorry I think I may have been unclear. I understand that TOML is more strongly typed, but my question was actually whether it should be possible to, for example, add `allow_untyped_defs = True` to the TOML file, which internally would then be interpreted by inverting it to the option as it is specified, e.g. `disallow_untyped_defs = False`.

I think the INI handler does that, and I wonder if the TOML handler should too. Ideally people could more or less just move their configuration verbatim, right?

Adding tests would not be aimed at increasing coverage, it seems to me, but just to assert that the entries people know and love from mypy.ini will work as expected in pyproject.toml.

By the way, I'm still trying to understand the `overrides` thing. Does the INI part have something similar?

pslacerda:
@erikkemperman I just added your remote and hard reset into your branch, but a force push was needed. No idea how git works.

About replicating the tests from INI I'm almost sure it isn't needed. Many command line tests are done through INI files. Command line, INI or TOML are only parsers, the option handling is the same. The TOML parser is currently tested.

About the `disallow`/`allow` thing I'm not sure, my opinion is to keep it off. Are you sure is required?

erikkemperman:
Yes, the history of the branch changed, relative to the one of the PR previously, due to the rebase onto latest master. That requires a force-push. That would be bad for branches other users might be using, but is usually fine for PRs (although reviewers might have to adjust if they had the changes checked out locally).

Personally I would think the processing of the TOML file should allow pretty much for the same content that is currently allowed in INI files, so I would think that includes this inversion. But perhaps that’s just me — I’d be curious what the maintainers have to say about that?

Same for testing extensively, I see what you mean about it not adding that much in the way of coverage — but for example this issue with inversions would have probably showed up if the tests were all duplicated. But I understand it would be a lot of work... I wonder if there’s anything that could be done to automate this, perhaps in parsing the test cases. I might have some time to take a look later today, if you like?

pslacerda:
Some copy & paste and search & replace would alleviate to replicate the tests. I'd love if you @erikkemperman take care of this and help on this boring & interesting PR. But the CLI and INI parsers already share most of the tests, I'm not sure if I'm understanding the need. The only things specific TOML besides general parsing are the bit about strict parsing and the boolean inversion.

Just like the CLI parser I think that boolean inversion is unecessary in the TOML handler. Maybe simply print an error message for any `allow/disallow` option. I may be wrong of course.

Later this week I'll test strict parsing.

erikkemperman:
Sorry for late reply.

I actually managed to hack the test-gathering code to duplicate, on the fly, all the ones with an equivalent toml file. Definitely not something to be included in the actual PR, but it gives some confidence that I could get everything to pass:

https://github.com/erikkemperman/mypy/tree/pep518-tweak

I personally would like the inversion to work, and I can imagine the maintainers might get a lot of question if something that worked in INI doesn't work in TOML. But again, maybe that is just me!

While I was rebasing I actually had to account for some `strict` handling already:

https://github.com/python/mypy/pull/5208/files#diff-09533a8b2bebadfae430681eb15b01c9R178-R179

I just wasn't sure if there needs to be anything for this in the overrides processing as well?

kudlatyamroth:
@erikkemperman @pslacerda 
is there any update on this?
I think that few ppl wait for it ;)

pslacerda:
@kudlatyamroth 
It was working as expected. Maybe it continue to work.

There was some minor disagreement between me and @erikkemperman so I'm awaiting for some regular contributor.

tgy:
Looking forward for this merge 👀

Queuecumber:
@ethanhs can we get some eyes on this? It's the only thing I use that hasn't switched over and one of only a few major tools that doesnt support pyproject.toml

TitaniumHocker:
Are there any updates?

Queuecumber:
Just pointing out that pylint, pytest, and isort all have support for pyproject.toml @ethanhs

msullivan (changes requested):
Alright, I don't really understand why people feel so strongly about pyproject.toml, but they do. I would like to make this PR go away, and it seems the easiest way to do that is probably to get it merged. Sorry for all the feet dragging from the core team here. Let's get this done for 0.800. @pslacerda are you still interested in this?

Here's what I need to merge this:

- [ ] Yes, we absolutely need to support option inversion
- [ ] Error handling; produce an error message on an unknown option and on type errors
- [x] If pyproject.toml is tried implicitly, do not produce a diagnostic if it is missing `tool.mypy`
- [ ] Ideally try to share a lot of code between the two systems; right now there is a totally separate toml path that duplicates things handled in the ini path in different ways. This will get a lot more painful when the previous two points are implemented. I am willing to entertain arguments that this is not possible or that the cure would be worse than the disease.
- [ ] Don't remove any existing tests
- [ ] Tests for everything mentioned above and some other things: wildcard sections, strict, at least one of the ones that uses a list. (always_true, maybe). @erikkemperman's automatic testing scheme is cute but doesn't currently work on the cmdline tests, which are the key ones here.

A note about the type error issue: one reason that this is important is that mypyc-compiled code generates runtime type errors. So with the current patch, if you do something like `disallow_untyped_defs = 0`, you'll get a traceback.

We also need documentation, but I'm OK with that being in a follow-up PR.

msullivan:
I took the liberty of rebasing and implementing "If pyproject.toml is tried implicitly, do not produce a diagnostic if it is missing `tool.mypy`". (I don't plan on picking up any of the other pieces myself.)

pslacerda:
I'll take another path. The current code is the way I expect. In this PR there are 3 systems, don't forget about CLI that share all the features except inversion if not mistaken.

I agree about the tests. You could rebase this PR or start a new one.  Please support pyproject.toml, it is fancier and stronger.

The only drawback I see is the introduction of a new requirement.

msullivan:
Alright, if anybody is excited about picking this up (@erikkemperman?), let me know. Otherwise I'll close the PR in a week or so. (It'll be available for anybody to pick back up still, of course.)

erikkemperman:
@msullivan I would very much like to see support for this, obviously. However I don’t feel entirely comfortable “taking over” the work of @pslacerda  if he is no longer behind it. All I did here was a rebase and playing around with the tests, just hoping to nudge things along.

I would need more time with this, more than I have at the moment, before I’d feel confident submitting anything more than that. I hope you understand.

pslacerda:
@erikkemperman I can hold the stack so you can work on top of the branch.

maxcountryman:
Has anyone taken this on? It would be great to see support for pyproject.toml so that I can define all these things in one place instead of a in dozen-odd disparate, bespoke configurations.

Olegt0rr:
Any news?

ethanhs:
No one has stepped forward to work on this for a while. I am going to close this PR and direct discussion of plans for pyproject.toml support to the issue https://github.com/python/mypy/issues/5205

If you would like to pick things up, please leave a comment on the issue. (Please do *not* ask for updates).

Issue #5210, opened 2018-06-12
map with lambda is broken in Python 2; gives '<Erased>' object is not iterable


gvanrossum:
@Michael0x2a 
This is a regression caused by your PR: f61c2baa Overhaul overload semantics, remove erasure, add union math (#5084).

Example (only works with Python 2 -- it's a syntax error with Python 3):
```py
a = [(1, ''), (2, '')]
b = map(lambda (i, s): str(i) + s, a)
```
Before that PR, this would pass; now it gives
```
_.py:2: error: '<Erased>' object is not iterable
```
Simplifying this further by using a list of ints and a lambda with a non-unpacking argument makes the error going away.

Sorry for being so aggressive with priority etc., this prevents us from using master on our internal code base -- we get a dozen or so of errors due to this. It's not limited to map() -- I see it too for reduce(), imap(), and ifilter(), always with `lambda (x, y): ...`.

gvanrossum:
Actually I'm not sure if it occurs with ifilter() -- the one occurrence is inside an imap() call:
```py
        owners = imap(lambda (k, v): 'owner=' + k,
                ifilter(lambda (k, v): v > SET_SIZE_LOW_CUTOFF, count_by_owner.iteritems()))
```
All the other functions involved are overloaded. Oh, I also saw an occurrence with max(), which is also overloaded.

gvanrossum:
There's also one occurence of
```
'<nothing>' object is not iterable
```
on a map(None, ...) call -- not a lambda in sight.

Also, it's not due to the typeshed changes -- I didn't bother changing typeshed when bisecting.

PR #5211, opened 2018-06-12
Fix leaking error messages during overload selection

Michael0x2a:
Fixes https://github.com/python/mypy/issues/5210 (?)

When mypy tries selecting an appropriate overload alternative, it swaps out the current 'msg' object with a placeholder. If this placeholder does not accumulate any error messages, we know that alternative is a successful match.

Previously, we replaced only `checkexpr.ExpressionChecker`'s msg object with the placeholder. However, since checkexpr can sometimes call `checker.TypeChecker`, we need to replace its msg object as well to prevent error messages that would normally be silenced from leaking.

More broadly, the current overload implementation broke the invariant that both msg fields refer to the same underlying MessageBuilder object. This pull request fixes that oversight.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1238,7 +1238,9 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         for typ in plausible_targets:
             overload_messages = self.msg.clean_copy()
             prev_messages = self.msg
+            assert self.msg is self.chk.msg
             self.msg = overload_messages
+            self.chk.msg = overload_messages
             try:
                 # Passing `overload_messages` as the `arg_messages` parameter doesn't
                 # seem to reliably catch all possible errors.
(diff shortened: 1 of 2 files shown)


Michael0x2a:
@gvanrossum -- I think this fixes the bug you found earlier -- do you mind re-testing w/ these changes?

Also, the way that I'm currently "capturing" the error messages feels pretty janky. Do you know if there's some better way of doing what the code is currently doing? 

(If not, I think we maybe should sometime in the future? It sort of feels like the codebase has several different ad-hoc ways of "capturing" or "conditionally silencing" errors and it might be nice to unify these different approaches in a more flexible way somehow -- maybe as part of a larger refactoring on how we handle error messages in general? /aside)

mypy/checkexpr.py
```diff
@@ -1238,7 +1238,9 @@ def infer_overload_return_type(self,
         for typ in plausible_targets:
             overload_messages = self.msg.clean_copy()
             prev_messages = self.msg
+            assert self.msg is self.chk.msg
```
gvanrossum: I think there's a slightly different pattern to ignore messages during a certain block of code, see e.g. checkmember.py L183. I'll test your code next.

PR #3527, opened 2017-06-12
add support for __path__ attribute on modules

tdyas:
Attempts to fix https://github.com/python/mypy/issues/1422

I am not familiar enough with the mypy code to know how to encode `List[str]` as the type of the `__path__` attribute. Feedback is definitely needed for that. (Then I can uncomment the line in the test that attempts to call `.append` on a `__path__` attribute.)

PR #5212, opened 2018-06-13
Fix name `__path__` not defined

pslacerda:
In #1422 `__path__` implicit module attribute is not defined. This fix is not perfect because `typing.List[str]` would be better.

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -101,7 +101,8 @@ inverse_node_kinds = {_kind: _name for _name, _kind in node_kinds.items()}
 implicit_module_attrs = {'__name__': '__builtins__.str',
                          '__doc__': None,  # depends on Python version, see semanal.py
                          '__file__': '__builtins__.str',
-                         '__package__': '__builtins__.str'}
+                         '__package__': '__builtins__.str',
+                         '__path__': '__builtins__.list'}
 
 
 type_aliases = {
--- a/mypy/semanal_pass1.py
+++ b/mypy/semanal_pass1.py
@@ -26,12 +26,13 @@ from mypy.nodes import (
     TryStmt, OverloadedFuncDef, Lvalue, Context, ImportedName, LDEF, GDEF, MDEF, UNBOUND_IMPORTED,
     MODULE_REF, implicit_module_attrs
 )
-from mypy.types import Type, UnboundType, UnionType, AnyType, TypeOfAny, NoneTyp
+from mypy.types import Type, UnboundType, UnionType, AnyType, TypeOfAny, NoneTyp, function_type
 from mypy.semanal import SemanticAnalyzerPass2, infer_reachability_of_if_statement
 from mypy.semanal_shared import create_indirect_imported_name
 from mypy.options import Options
 from mypy.sametypes import is_same_type
 from mypy.visitor import NodeVisitor
+from mypy.nodes import FuncItem
 
 
 class SemanticAnalyzerPass1(NodeVisitor[None]):
@@ -123,6 +124,11 @@ class SemanticAnalyzerPass1(NodeVisitor[None]):
                     # TODO: Find a permanent solution to this problem.
                     # Maybe add 'bool' to all fixtures?
                     literal_types.append(('True', AnyType(TypeOfAny.special_form)))
+                # TODO: The __path__ implicit module attribute is typed as __builtins__.list
+                # but tests often don't define list.
+                if 'list' not in self.sem.globals:
+                    # literal_types.append(('list', function_type(FuncItem([], ''), None)))
+                    literal_types.append(('list', AnyType(TypeOfAny.special_form)))
 
                 for name, typ in literal_types:
                     v = Var(name, typ)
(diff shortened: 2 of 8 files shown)


pslacerda:
The last commit was to fix the tests.

test-data/unit/check-incomplete-fixture.test
```diff
@@ -12,7 +12,7 @@ import m
 m.x # E: "object" has no attribute "x"
 [file m.py]
 
-[case testListMissingFromStubs]
+[case testListMissingFromStubs-skip]
```
ilevkivskyi: Why do you skip this test?

test-data/unit/check-basic.test
```diff
@@ -232,6 +232,13 @@ a = __file__ # type: A  # E: Incompatible types in assignment (expression has ty
 class A: pass
 [builtins fixtures/primitives.pyi]
 
+[case testModule__path__]
+import typing
+x = __path__ # type: list
+a = __path__ # type: A  # E: Incompatible types in assignment (expression has type "list", variable has type "A")
```
ilevkivskyi: How about just using a `reveal_type` on `__path__`?

mypy/nodes.py
```diff
@@ -101,7 +101,8 @@ def get_column(self) -> int:
 implicit_module_attrs = {'__name__': '__builtins__.str',
                          '__doc__': None,  # depends on Python version, see semanal.py
                          '__file__': '__builtins__.str',
-                         '__package__': '__builtins__.str'}
+                         '__package__': '__builtins__.str',
+                         '__path__': '__builtins__.list'}
```
ilevkivskyi: IIRC `__path__` is only defined in `__init__.py` modules, not in _every_ module.

pslacerda:
#3527 was typing `__path__` as `__builtins__.str`, because of this `__path__.append()` was failing.
In this PR the variable is correctly typed, even if only partially.

I didn't knew that `__path__` was defined only in `__init__.pyi?` files.

The test was skipped because `list` was forced to be defined in every module if not previously defined, very much like `True`.

test-data/unit/check-incomplete-fixture.test
pslacerda: Because I thought that was not that important. I think that it can be fixed adding somethat define `list`.

test-data/unit/check-basic.test
pslacerda: I agree, I was only going with the flow of other similar tests. Should them also use `reveal_type` on `__name__` and etc?

mypy/nodes.py
pslacerda: I thought that `__path__` was defined everywhere. I will fix that.

test-data/unit/check-incomplete-fixture.test
pslacerda: I tried `del list` but in mypy `del` deletes the value and not the variable itself. I would keep as is and mark this test to skip.

test-data/unit/check-basic.test
pslacerda: As `builtins.list` is partially typed, `reveal_type` reveals `builtins.list[Any]` instead of `builtins.list`.

mypy/semanal_pass1.py
```diff
@@ -124,6 +124,11 @@ def visit_file(self, file: MypyFile, fnam: str, mod_id: str, options: Options) -
                     # Maybe add 'bool' to all fixtures?
                     literal_types.append(('True', AnyType(TypeOfAny.special_form)))
 
+                # We are running tests without 'list' in builtins.
+                # TODO: Maybe optionally choose an default fixture stub per test file?
+                if 'list' not in self.sem.globals:
+                    literal_types.append(('list', AnyType(TypeOfAny.special_form)))
```
ilevkivskyi: This is very dangerous. With this addition many tests will pass (of fail in ultra-weird way) if someone forgets to add a fixture with `list`. It is OK with _literals_, but not with a class.

test-data/unit/fixtures/primitives.pyi
```diff
@@ -20,4 +20,5 @@ class str:
 class bytes: pass
 class bytearray: pass
 class tuple(Generic[T]): pass
+class list(Generic[T]): pass
```
ilevkivskyi: The header of this file says `# builtins stub with non-generic primitive types`, so I would avoid adding generic classes here. We have already enough fixtures with `list`, why do you need more?

mypy/nodes.py
ilevkivskyi: If you will fix this, you can then just take care to add `list.pyi` fixture to all tests that include `__init__.py[i]` files. I think there are not many of those.

gvanrossum:
This seems stalled. @pslacerda are you planning to address the issues Ivan mentioned in the code review?

mypy/semanal_pass1.py
pslacerda: Isn't `list` a literal?

mypy/nodes.py
pslacerda: Are you saying that `__path__` shoun't be here because it exists only in `__init__`? So `__init__.py[i]` that uses `__path__` must include an fixture?

ilevkivskyi: Yes, please see the previous comments to this code. You already promised to fix it in https://github.com/python/mypy/pull/5212#discussion_r195136274 two months ago.

It is OK to just close this if you don't have time/desire to work on this. If you still want to do this, then please go through _all_ previous comments and fix mentioned issues.

gvanrossum:
Let's close this, it's stalled.

Issue #4788, opened 2018-03-24
Support default values for collections.namedtuple?


ilevkivskyi:
Python 3.7 adds support for specifying default values in call to `collections.namedtuple`, see https://github.com/python/cpython/pull/4859. So for example this is valid:
```python
from collections import namedtuple

NT = namedtuple('NT', ('x', 'y'), defaults=(0, 0))
NT()
NT(1)
NT(1, 2)
```
Should we support this, or supporting default values for `typing.NamedTuple` (already implemented) is enough?

gvanrossum:
The short form of `collections.namedtuple` is very popular (ask @rhettinger) and I think we should make it as easy as possible to translate its use to `typing.NamedTuple`. So I think we should support it.

JukkaL:
I agree with @gvanrossum. Also, this doesn't look like a lot of work to implement. Tagging as high priority since 3.7 release date is pretty close.

JelleZijlstra:
Fixed by #5215.

Issue #2127, opened 2016-09-13
namedtuple does not support additional arguments


elazarg:
```
$ cat > tmp.py
from collections import namedtuple
T = namedtuple('T', 'a b', verbose=True, rename=False)
$ mypy tmp.py
tmp.py:2: error: Too many arguments for namedtuple()
tmp.py:2: error: "object" not callable
```

Related: [typeshed #541](https://github.com/python/typeshed/pull/533)

gvanrossum:
Thanks for the report -- since you refactored this code yourself recently I presume you know where to fix it!

JelleZijlstra:
I'm working on a fix for this and #2124.

snstanton:
The same error is reported for `typing.NamedTuple`, which supports arbitrary named parameters.  Will the fix for `collections.namedtuple` address `typing.NamedTuple`?

JelleZijlstra:
Fixed by #5215.

PR #5215, opened 2018-06-14
Support additional args to namedtuple()

### LINKED ISSUE #2124: namedtuple should support "module" parameter added in Python 3.6
See http://bugs.python.org/issue17941 and https://github.com/python/typeshed/issues/533

JelleZijlstra:
By letting typeshed do the work.

Fixes #2124, #2127, #4788.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -201,7 +201,8 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         """Type check a call expression."""
         if e.analyzed:
             if isinstance(e.analyzed, NamedTupleExpr) and not e.analyzed.is_typed:
-                # Type check the arguments, but ignore the results.
+                # Type check the arguments, but ignore the results. This relies
+                # on the typeshed stubs to type check the arguments.
                 self.visit_call_expr_inner(e)
             # It's really a special form that only looks like a call.
             return self.accept(e.analyzed, self.type_context[-1])
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -200,8 +200,14 @@ class ExpressionChecker(ExpressionVisitor[Type]):
     def visit_call_expr(self, e: CallExpr, allow_none_return: bool = False) -> Type:
         """Type check a call expression."""
         if e.analyzed:
+            if isinstance(e.analyzed, NamedTupleExpr) and not e.analyzed.is_typed:
+                # Type check the arguments, but ignore the results.
+                self.visit_call_expr_inner(e)
             # It's really a special form that only looks like a call.
             return self.accept(e.analyzed, self.type_context[-1])
+        return self.visit_call_expr_inner(e, allow_none_return=allow_none_return)
+
+    def visit_call_expr_inner(self, e: CallExpr, allow_none_return: bool = False) -> Type:
         if isinstance(e.callee, NameExpr) and isinstance(e.callee.node, TypeInfo) and \
                 e.callee.node.typeddict_type is not None:
             # Use named fallback for better error messages.
--- a/mypy/semanal_namedtuple.py
+++ b/mypy/semanal_namedtuple.py
@@ -158,8 +158,6 @@ class NamedTupleAnalyzer:
             name += '@' + str(call.line)
         if len(defaults) > 0:
             default_items = {
-                # We don't have the actual argument expressions any more, so we
-                # just use Ellipsis for all defaults.
                 arg_name: default
                 for arg_name, default in zip(items[-len(defaults):], defaults)
             }
@@ -174,8 +172,8 @@ class NamedTupleAnalyzer:
         call.analyzed.set_line(call.line, call.column)
         return info
 
-    def parse_namedtuple_args(self, call: CallExpr,
-                              fullname: str) -> Tuple[List[str], List[Type], List[Expression], bool]:
+    def parse_namedtuple_args(self, call: CallExpr, fullname: str
+                              ) -> Tuple[List[str], List[Type], List[Expression], bool]:
         """Parse a namedtuple() call into data needed to construct a type.
 
         Returns a 4-tuple:
@@ -246,9 +244,9 @@ class NamedTupleAnalyzer:
             defaults = defaults[:len(items)]
         return items, types, defaults, ok
 
-    def parse_namedtuple_fields_with_types(
-        self, nodes: List[Expression], context: Context
-    ) -> Tuple[List[str], List[Type], int, bool]:
+    def parse_namedtuple_fields_with_types(self, nodes: List[Expression], context: Context
+                                           ) -> Tuple[List[str], List[Type], List[Expression],
+                                                      bool]:
         items = []  # type: List[str]
         types = []  # type: List[Type]
         for item in nodes:
@@ -268,12 +266,12 @@ class NamedTupleAnalyzer:
                 types.append(self.api.anal_type(type))
             else:
                 return self.fail_namedtuple_arg("Tuple expected as NamedTuple() field", item)
-        return items, types, 0, True
+        return items, types, [], True
 
-    def fail_namedtuple_arg(self, message: str,
-                            context: Context) -> Tuple[List[str], List[Type], int, bool]:
+    def fail_namedtuple_arg(self, message: str, context: Context
+                            ) -> Tuple[List[str], List[Type], List[Expression], bool]:
         self.fail(message, context)
-        return [], [], 0, False
+        return [], [], [], False
 
     def build_namedtuple_typeinfo(self, name: str, items: List[str], types: List[Type],
                                   default_items: Mapping[str, Expression]) -> TypeInfo:
--- a/mypy/semanal_namedtuple.py
+++ b/mypy/semanal_namedtuple.py
@@ -244,8 +244,9 @@ class NamedTupleAnalyzer:
             num_defaults = len(items)
         return items, types, num_defaults, ok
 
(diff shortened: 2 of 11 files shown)


test-data/unit/check-namedtuple.test
```diff
@@ -1,7 +1,7 @@
 [case testNamedTupleUsedAsTuple]
 from collections import namedtuple
 
-X = namedtuple('X', ['x', 'y'])
+X = namedtuple('X', 'x y')
```
JelleZijlstra: My change ends up requiring `list` to exist in all of these tests. It was easier to just change the namedtuple call than to add `[builtins fixtures/list.pyi]` everywhere. I left a few with the list to make sure we test that code path.

test-data/unit/check-namedtuple.test
```diff
@@ -54,31 +54,57 @@ a.y = 5 # E: Property "y" defined in "X" is read-only
...
+B = namedtuple('B', 'a b', rename=1)
+C = namedtuple('C', 'a b', rename='not a bool')  # E: Argument "rename" to "namedtuple" has incompatible type "str"; expected "int"
+# This works correctly but also produces "mypy/test-data/unit/lib-stub/collections.pyi:3: note: "namedtuple" defined here"
+# Don't know how to express that in a test.
+# D = namedtuple('D', 'a b', unrecognized_arg=False)  # E: Unexpected keyword argument "unrecognized_arg" for "namedtuple"
+E = namedtuple('E', 'a b', 0)  # E: Too many positional arguments for "namedtuple"
+
+[builtins fixtures/bool.pyi]
+
+[case testNamedTupleDefaults-skip]
```
JelleZijlstra: will work on enabling this next (#4788)

test-data/unit/check-python2.test
```diff
@@ -13,14 +13,6 @@ s = b'foo'
 from typing import TypeVar
 T = TypeVar(u'T')
 
-[case testNamedTupleUnicode]
```
JelleZijlstra: This no longer worked because my stub for namedtuple just has str. Apparently we don't support `Text` in the test stubs, so I just moved this test case to the pythoneval tests.

test-data/unit/fixtures/args.pyi
```diff
@@ -27,3 +27,4 @@ class int:
 class str: pass
 class bool: pass
 class function: pass
+class ellipsis: pass
```
JelleZijlstra: This was needed because a few tests that import collections didn't have `...` available in their fixtures, and my stub for namedtuple uses it.

test-data/unit/semanal-namedtuple.test
```diff
@@ -138,10 +138,6 @@ MypyFile:1(
 from collections import namedtuple
 N = namedtuple('N') # E: Too few arguments for namedtuple()
 
-[case testNamedTupleWithTooManyArguments]
```
JelleZijlstra: This test stopped working because this error is now caught during type checking rather than semanal.

mypy/semanal_namedtuple.py
```diff
@@ -196,17 +232,27 @@ def parse_namedtuple_args(self, call: CallExpr,
...
             self.fail("namedtuple() field names cannot start with an underscore: "
                       + ', '.join(underscore), call)
-        return items, types, ok
+        if num_defaults > len(items):
+            self.fail("Too many defaults given in call to namedtuple()", call)
+            num_defaults = len(items)
+        return items, types, num_defaults, ok
+
+    def extract_defaults_from_arg(self, call: CallExpr, arg: Expression) -> int:
+        if not isinstance(arg, (ListExpr, TupleExpr)):
```
JelleZijlstra: Technically the defaults argument can be any iterable, but it doesn't seem particularly useful to support that.

gvanrossum:
Did anyone else review this yet? It seems oddly stalled (I wonder what happened around June 13, a number of PRs saw their last activity around that date).

JelleZijlstra:
No. As far as I can see, it's ready for review.

ilevkivskyi:
I can review it later today.

JelleZijlstra:
Thanks Ivan!

mypy/checkexpr.py
```diff
@@ -200,8 +200,14 @@ def analyze_var_ref(self, var: Var, context: Context) -> Type:
     def visit_call_expr(self, e: CallExpr, allow_none_return: bool = False) -> Type:
         """Type check a call expression."""
         if e.analyzed:
+            if isinstance(e.analyzed, NamedTupleExpr) and not e.analyzed.is_typed:
+                # Type check the arguments, but ignore the results.
```
ilevkivskyi: Could you please add a comment here that generating correct errors in call here depends on the typeshed stub?

mypy/semanal_namedtuple.py
```diff
@@ -142,35 +142,71 @@ def check_namedtuple(self,
...
             # Error. Construct dummy return value.
             return self.build_namedtuple_typeinfo('namedtuple', [], [], {})
         name = cast(StrExpr, call.args[0]).value
         if name != var_name or is_func_scope:
             # Give it a unique name derived from the line number.
             name += '@' + str(call.line)
-        info = self.build_namedtuple_typeinfo(name, items, types, {})
+        if num_defaults > 0:
+            default_items = {
+                arg_name: EllipsisExpr()
```
ilevkivskyi: This is a bit hacky so at least deserves a comment why we use `EllipsisExpr()` instead of using actual expressions.

If possible, I would just use the actual expressions here.

mypy/semanal_namedtuple.py
```diff
@@ -142,35 +142,71 @@ def check_namedtuple(self,
...
         if len(args) < 2:
             return self.fail_namedtuple_arg("Too few arguments for namedtuple()", call)
+        num_defaults = 0
         if len(args) > 2:
-            # FIX incorrect. There are two additional parameters
-            return self.fail_namedtuple_arg("Too many arguments for namedtuple()", call)
-        if call.arg_kinds != [ARG_POS, ARG_POS]:
+            # Typed namedtuple doesn't support additional arguments.
+            if fullname == 'typing.NamedTuple':
+                return self.fail_namedtuple_arg("Too many arguments for namedtuple()", call)
```
ilevkivskyi: I would capitalize `namedtuple` -> `NamedTuple` due to the proposed change.

test-data/unit/check-namedtuple.test
```diff
@@ -64,21 +64,48 @@ x = X(1, 2, 3)  # E: Too many arguments for "X"
...
 a, b = x
 a, b, c = x  # E: Need more than 2 values to unpack (3 expected)
 
+[case testNamedTupleAdditionalArgs]
+from collections import namedtuple
+
+A = namedtuple('A', 'a b')
+B = namedtuple('B', 'a b', rename=1)
+C = namedtuple('C', 'a b', rename='not a bool')  # E: Argument "rename" to "namedtuple" has incompatible type "str"; expected "int"
+# This works correctly but also produces "mypy/test-data/unit/lib-stub/collections.pyi:3: note: "namedtuple" defined here"
```
ilevkivskyi: You can add an `[out]` section at the end that lists all error verbatim instead of using `# E:`

test-data/unit/lib-stub/collections.pyi
```diff
@@ -1,3 +1,11 @@
-import typing
+from typing import Any, Iterable, Union, Optional
 
-namedtuple = object()
+def namedtuple(
+    typename: str,
+    field_names: Union[str, Iterable[str]],
+    *,
+    # really int but many tests don't have bool available
```
ilevkivskyi: I think this should read "really bool".

JelleZijlstra:
Thanks for the review! I will push an update soon.

mypy/semanal_namedtuple.py
JelleZijlstra: I think I can get the actual expressions, but I'm not sure I understand why that would be better. It would make the code more complicated and I don't see much of an advantage.

ilevkivskyi: If this is really complex then it maybe doesn't make sense. In general now that we work on mypyc it is simpler if we keep all real info somewhere in the AST.

JelleZijlstra: Oh yes, this makes sense in the context of mypyc. It isn't too bad actually, I'll just make the change.

JelleZijlstra:
Will fix the CI failure later today.

mypy/semanal_namedtuple.py
```diff
@@ -142,35 +142,73 @@ def check_namedtuple(self,
...
             return self.build_namedtuple_typeinfo('namedtuple', [], [], {})
         name = cast(StrExpr, call.args[0]).value
         if name != var_name or is_func_scope:
             # Give it a unique name derived from the line number.
             name += '@' + str(call.line)
-        info = self.build_namedtuple_typeinfo(name, items, types, {})
+        if len(defaults) > 0:
+            default_items = {
+                # We don't have the actual argument expressions any more, so we
+                # just use Ellipsis for all defaults.
```
ilevkivskyi: This comment is now outdated.

JelleZijlstra: Yes, removed.

gvanrossum:
So the initial description said "Fixes #2124, #2127, #4788." And only #2124 got auto-closed (auto-close isn't smart enough :-). Should the other two also be closed now?

JelleZijlstra:
Thanks for catching that! I'll close the other ones too.

JelleZijlstra:
I suppose I should have written "Fixes #2124. Fixes #2127. Fixes #4788."

Issue #5204, opened 2018-06-12
Overload on Optional[str] and bytes fails with arg of type Union[str, bytes]


gvanrossum:
This may well be a known issue. @Michael0x2a. (The real code used `os.listdir()`, but I'm simplifying it here and it fails the same way.)
```py
@overload
def listdir(a: Optional[str] = None): pass
@overload
def listdir(a: bytes): pass
def listdir(a): pass

def f(a: Union[bytes, str]):
  listdir(a)  # error: Argument 1 to "listdir" has incompatible type "Union[bytes, str]"; expected "Optional[str]"
```

ilevkivskyi:
Hm, if I remove the default value then it works:
```python
@overload
def listdir(a: Optional[str]): pass
@overload
def listdir(a: bytes): pass
def listdir(a): pass

def f(a: Union[bytes, str]):
  listdir(a)  # OK
```
So must be some unnecessary restriction on arg kinds in union math and so probably easy to fix.

PR #5222, opened 2018-06-15
Relax union math logic to allow signatures with different arg kinds

Michael0x2a:
Resolves https://github.com/python/mypy/issues/5204.

This commit relaxes the union math logic so that it allows signatures with arg kinds that are nearly identical except that one arg is positional and the other is optional.

This commit also removes the "names must be the same" restriction mostly so that the original example given in https://github.com/python/mypy/issues/4576 will pass. (In retrospect, I think this check didn't really buy us much -- even if the alternatives share the same arg names, there's no guarantee the actual implementation signature will also share the same.)

### DIFF
--- a/test-data/unit/check-overloading.test
+++ b/test-data/unit/check-overloading.test
@@ -2157,7 +2157,7 @@ reveal_type(foo(compat))  # E: Revealed type is 'Union[builtins.int, builtins.st
 not_compat: Union[WrapperCo[A], WrapperContra[C]]
 foo(not_compat)  # E: Argument 1 to "foo" has incompatible type "Union[WrapperCo[A], WrapperContra[C]]"; expected "Union[WrapperCo[B], WrapperContra[B]]"
 
-[case testOverloadInferUnionSkipIfParameterNamesAreDifferent]
+[case testOverloadInferUnionIfParameterNamesAreDifferent]
 from typing import overload, Union
 
 class A: ...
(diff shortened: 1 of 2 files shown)


test-data/unit/check-overloading.test
```diff
@@ -2173,7 +2173,7 @@ def f(x): ...
 x: Union[A, B]
 reveal_type(f(A()))  # E: Revealed type is '__main__.B'
 reveal_type(f(B()))  # E: Revealed type is '__main__.C'
-f(x)                 # E: Argument 1 to "f" has incompatible type "Union[A, B]"; expected "A"
+reveal_type(f(x))    # E: Revealed type is 'Union[__main__.B, __main__.C]'
```
ilevkivskyi: The test case name should be updated (remove "Skip").

test-data/unit/check-overloading.test
```diff
@@ -2191,20 +2191,54 @@ def f(x: A, y: Optional[B] = None) -> C: ...
...
+
+@overload
+def f(x: A) -> Child: ...
+@overload
+def f(x: B, y: B = B()) -> Parent: ...
+def f(*args): ...
+
+x: Union[A, B]
+f(x)       # E: Argument 1 to "f" has incompatible type "Union[A, B]"; expected "A"
+f(x, B())  # E: Argument 1 to "f" has incompatible type "Union[A, B]"; expected "B"
```
ilevkivskyi: I think ideally we could support something like this. Could you please add a TODO here? (I see you have a TODO in code, but I would prefer to have them both).

Michael0x2a:
Thanks! Merging now.

PR #5223, opened 2018-06-15
Require abstract classes in stubs to be explicitly marked

### LINKED ISSUE #3692: Catch accidentally abstract classes
We've had a series of typeshed issues where classes accidentally became abstract because an abstract method in a parent class wasn't overridden (see python/typeshed#1476). It would be nice if mypy could tell us about this sort of issue automatically, since for now we have to rely on bug reports from end users.

I'm not sure if there's a good programmatic way to distinguish between accidentally and intentionally abstract classes. Here are a few ideas:
- Warn if a class is abstract, but doesn't itself have any abstractmethods (all abstractmethods are inherited from a parent class).
- Require all intentionally abstract classes to specify ABCMeta as an explicit metaclass. This would be fairly disruptive.
- Don't do anything. This problem has repeatedly affected typeshed, but it's probably much easier to discover outside stubs, since the type checker will also see code using the class, not just its definition like in stubs.

JelleZijlstra:
Fixes #3692.

In typeshed, we have had many bugs because of classes that were not meant to be abstract, but were inferred by mypy to be abstract because they didn't implement some methods from an abstract base class. I call such classes "accidentally abstract"; python/typeshed#1476 has a list of cases where this bug appeared.

It is hard to catch these bugs because they only show up in user code when somebody tries to instantiate the class, and when reviewing changes to typeshed you have to manually go over all methods to find instances of this bug. Therefore, I'd like to have mypy point out accidentally abstract classes automatically.

This PR throws an error if a class:
- Has any abstract method;
- Does not itself have any method decorated with `@abstractmethod` (i.e., all abstract methods are inherited);
- Does not have `ABCMeta` as an explicit metaclass, or `Protocol` as an explicit base class;
- Is defined in a stub file.
This generates about five false positives in typeshed's Python 3 code, which we can fix by manually adding `ABCMeta` as a metaclass. On the other hand, there have been at least 30 accidentally abstract classes in typeshed that would have been caught by this patch, although I've fixed most of them by now.

This PR is going to fail tests for now because not all typeshed classes are fixed yet. If we're in agreement that this PR is acceptable, I will work on getting typeshed ready for it by adding the right `ABCMeta` metaclasses.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -867,7 +867,8 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
         if abstract and not abstract_in_this_class:
             attrs = ", ".join('"{}"'.format(attr) for attr in sorted(abstract))
             self.fail("Class {} has abstract attributes {}".format(typ.fullname(), attrs), typ)
-            self.note("If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass", typ)
+            self.note("If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass",
+                      typ)
 
     def setup_type_promotion(self, defn: ClassDef) -> None:
         """Setup extra, ad-hoc subtyping relationships between classes (promotion).
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -828,6 +828,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
         """
         concrete = set()  # type: Set[str]
         abstract = []  # type: List[str]
+        abstract_in_this_class = []  # type: List[str]
         for base in typ.mro:
             for name, symnode in base.names.items():
                 node = symnode.node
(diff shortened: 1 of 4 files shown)


ilevkivskyi:
I totally like the idea in principle, but what to do with the test failures (mypy uses many test fixtures)? Maybe limit this to only typeshed stubs (in contrast to all stubs)? There is a corresponding flag in `SemanticAnalyzer`.

JelleZijlstra:
I think the failures in mypy's stubs won't be too hard to fix.

The `is_typeshed_stub_file` flag doesn't work for typeshed's `mypy_test.py`, which would be somewhat beside the point of this feature. Also, in general the feature should be useful for all stub packages, not just for typeshed.

JelleZijlstra:
Going to close and re-open to see if that forces Appveyor to re-run (Travis already passed).

ilevkivskyi:
All tests pass now, but I would wait until tomorrow to see if there any objections from @JukkaL

JelleZijlstra:
@JukkaL do you have any feedback here? (Also, thanks for fixing the merge Ivan.)

ilevkivskyi:
FWIW this PR passes `testpr` cleanly.

ilevkivskyi:
I assume the silence means no.

Issue #328, opened 2014-07-25
Overloaded classmethods don't work


JukkaL:
Overloaded classmethods don't work correctly in the type checker:

```
from typing import overload
class A:
    @overload
    @classmethod
    def f(cls, x: int) -> None: pass
    @overload
    @classmethod
    def f(cls, x: str) -> None: pass
A.f(1)  # No overload variant matches argument types
A.f('')  # No overload variant matches argument types
```

This affects dict.fromkeys.

dmoisset:
Currently the typeshed declares fromkeys as a staticmethod, and I still get the same problem:

```
d = {"greeting": "hello", "name": "Bob"}
print(dict.fromkeys(d, 42))  # No overload variant matches argument types
```

is this the same bug?

Issue #2254, opened 2016-10-14
 No overload variant of "fromkeys" of "dict" matches argument types [typing.Iterable[Any], Union[Any, def (*Any, **Any) -> Any]]


trnsnt:
Hello, 

I have defined a dict wrapper where I override fromkeys method. 
I have this method signature

```
    def fromkeys(cls, iterable, value=None):
        # type: (Iterable, Union[Any, Callable]) -> DictWrapper
```

and this give me the following error :  No overload variant of "fromkeys" of "dict" matches argument types [typing.Iterable[Any], Union[Any, def (_Any, *_Any) -> Any]]

Thanks !

gvanrossum:
Thanks for the question! I'd be happy to help but I'm not sure I understand the context of that definition. Can you show a complete, small, self-contained program that exhibits this behavior?

trnsnt:
Hello, thanks for the reply. Here is what I have done.
I have this class 

`
class DictWrapper(dict):

    @classmethod
    def fromkeys(cls, iterable, value=None):
        # type: (Iterable, Union[Any, Callable]) -> DictWrapper
        if not callable(value):
            return DictWrapper(super(d, cls).fromkeys(iterable, value))

        def _():
            for key in iterable:
                yield key, value(key)

        return DictWrapper(_())
`

Code is working fine (Python 3.5), but when I run mypy, I have the following error : 
No overload variant of "fromkeys" of "dict" matches argument types [typing.Iterable[Any], Union[Any, def (Any, *Any) -> Any]]
I am not sure if it is a problem, or if I have done something wrong.

gvanrossum:
Oh, it seems you've hit two issues! When I run that example I get:
```
__tmp__.py: note: In class "DictWrapper":
__tmp__.py:6: error: Signature of "fromkeys" incompatible with supertype "dict"
__tmp__.py: note: In member "fromkeys" of class "DictWrapper":
__tmp__.py:9: error: No overload variant of "fromkeys" of "dict" matches argument types [typing.Iterable[Any], Union[Any, def (*Any, **Any) -> Any]]
```
(Line 6 is the `def`, line 9 is the `super()` call.)

The first complaint is because in typeshed, `fromkeys()` is defined as a static method, while it is actually a class method. I don't know why this is, but I think I can just fix that.

[...]

PR #5224, opened 2018-06-15
Make overloads support classmethod and staticmethod

Michael0x2a:
This pull request adds support for mixing classmethods, staticmethods, and overloads.

It does so by adding in some extra logic into the semantic analysis phase to detect when an overload is using classmethods or staticmethods and adds that information into OverloadedFuncDef. This allowed me to add `is_class` and `is_static` fields to FuncBase, the common base class for overloads and other function-related things.

This PR does not attempt to handle overloads + other arbitrary decorators.

Resolves https://github.com/python/mypy/issues/328 (which unblocks https://github.com/python/mypy/issues/2254?)

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1290,8 +1290,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             fail = True
 
         if isinstance(original, CallableType) and isinstance(override, CallableType):
-            if (isinstance(original.definition, FuncItem) and
-                    isinstance(override.definition, FuncItem)):
+            if (isinstance(original.definition, FuncBase) and
+                    isinstance(override.definition, FuncBase)):
                 if ((original.definition.is_static or original.definition.is_class) and
                         not (override.definition.is_static or override.definition.is_class)):
                     fail = True
@@ -3923,8 +3923,6 @@ def is_untyped_decorator(typ: Optional[Type]) -> bool:
 def is_static(func: Union[FuncBase, Decorator]) -> bool:
     if isinstance(func, Decorator):
         return is_static(func.func)
-    elif isinstance(func, OverloadedFuncDef):
-        return any(is_static(item) for item in func.items)
-    elif isinstance(func, FuncItem):
+    elif isinstance(func, FuncBase):
         return func.is_static
-    return False
+    raise AssertionError("Unexpected func type: {}".format(type(func)))
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3923,4 +3923,4 @@ def is_static(func: Union[FuncBase, Decorator]) -> bool:
         return is_static(func.func)
     elif isinstance(func, FuncBase):
         return func.is_static
-    raise AssertionError("Unexpected func type: {}".format(type(func)))
+    assert False, "Unexpected func type: {}".format(type(func))
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1289,12 +1289,10 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             # this could be unsafe with reverse operator methods.
             fail = True
 
-        if isinstance(original, CallableType) and isinstance(override, CallableType):
-            if (isinstance(original.definition, FuncBase) and
-                    isinstance(override.definition, FuncBase)):
-                if ((original.definition.is_static or original.definition.is_class) and
-                        not (override.definition.is_static or override.definition.is_class)):
-                    fail = True
+        if isinstance(original, FunctionLike) and isinstance(override, FunctionLike):
+            if ((original.is_classmethod() or original.is_staticmethod()) and
+                    not (override.is_classmethod() or override.is_staticmethod())):
+                fail = True
 
         if fail:
             emitted_msg = False
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -370,6 +370,11 @@ class ImportedName(SymbolNode):
         return 'ImportedName(%s)' % self.target_fullname
 
 
+FUNCBASE_FLAGS = [
+    'is_property', 'is_class', 'is_static',
+]
+
+
 class FuncBase(Node):
     """Abstract base class for function-like nodes"""
 
@@ -377,6 +382,8 @@ class FuncBase(Node):
                  'unanalyzed_type',
                  'info',
                  'is_property',
+                 'is_class',
+                 'is_static',
                  '_fullname',
                  )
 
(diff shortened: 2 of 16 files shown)


mypy/nodes.py
```diff
@@ -481,9 +490,9 @@ def set_line(self, target: Union[Context, int], column: Optional[int] = None) ->
         self.variable.set_line(self.line, self.column)
 
 
-FUNCITEM_FLAGS = [
+FUNCITEM_FLAGS = FUNCBASE_FLAGS + [
```
ilevkivskyi: This change adds `is_property`, is this intentional?

mypy/plugins/attrs.py
```diff
@@ -470,6 +470,16 @@ def _add_init(ctx: 'mypy.plugin.ClassDefContext', attributes: List[Attribute],
...
+        if isinstance(stmt, OverloadedFuncDef) and stmt.is_class:
+            func_type = stmt.type
+            if isinstance(func_type, Overloaded):
+                class_type = ctx.api.class_type(ctx.cls.info)
+                for item in func_type.items():
+                    item.arg_types[0] = class_type
+                if stmt.impl is not None:
+                    assert isinstance(stmt.impl, Decorator)
+                    if isinstance(stmt.impl.func.type, CallableType):
+                        stmt.impl.func.type.arg_types[0] = class_type
```
ilevkivskyi: It's good that you also take care about plugins.

mypy/nodes.py
```diff
@@ -370,13 +370,20 @@ def __str__(self) -> str:
         return 'ImportedName(%s)' % self.target_fullname
 
 
+FUNCBASE_FLAGS = [
+    'is_property', 'is_class', 'is_static',
+]
```
ilevkivskyi: Shouldn't we also update `astdiff.py` according to these flag reshuffling? This may break fine grained incremental mode (in some corner cases).

mypy/semanal.py
```diff
@@ -587,6 +587,37 @@ def _visit_overloaded_func_def(self, defn: OverloadedFuncDef) -> None:
...
+        # We know this is an overload def -- let's handle classmethod and staticmethod
+        class_status = []
+        static_status = []
+        for item in defn.items:
+            if isinstance(item, Decorator):
+                inner = item.func
+            elif isinstance(item, FuncDef):
+                inner = item
+            else:
+                raise AssertionError()
```
ilevkivskyi: Please add a message to all assertion errors. Typically we write `assert False, "Impossible blah-blah-blah"`.

test-data/unit/check-attr.test
```diff
@@ -428,6 +428,38 @@ a = A.new()
...
 [builtins fixtures/classmethod.pyi]
 
+[case testAttrsOtherOverloads]
+import attr
+from typing import overload, Union
+
+@attr.s
+class A:
+    a: int
+    b: str
```
ilevkivskyi: I would rather make these `attr.ib`s, to check that the `cls` signature below is generated correctly by the plugin.

test-data/unit/check-overloading.test
```diff
@@ -2937,3 +2937,322 @@ def add_proxy(x, y):
...
+
+    @overload
+    @classmethod
+    def foo(cls, x: str) -> str: ...
+
+    @classmethod
+    def foo(cls, x): pass
+
+class BadChild(Parent):
+    @overload  # E: Signature of "foo" incompatible with supertype "Parent"
```
ilevkivskyi: Maybe add few more tests to check overriding an overloaded instance method, with overloaded class methods and vice versa?

mypy/checker.py
```diff
@@ -1290,8 +1290,8 @@ def check_override(self, override: FunctionLike, original: FunctionLike,
             fail = True
 
         if isinstance(original, CallableType) and isinstance(override, CallableType):
-            if (isinstance(original.definition, FuncItem) and
-                    isinstance(override.definition, FuncItem)):
+            if (isinstance(original.definition, FuncBase) and
```
ilevkivskyi: Note that this PR (in particular because of this change) may have conflicts with my type aliases refactoring. In turned out that the `type_override` (that I killed) was also used for some overload hacks.

mypy/nodes.py
Michael0x2a: It is. I made this change partly on principle since `is_property` actually is a field of FuncBase -- it felt cleaner to just force all subclasses to preserve that field no matter what.

This change also doesn't actually change the serialized output in practice. FuncItem currently has only two subtypes: FuncDef and LambdaExpr. The former subclass previously explicitly set and serialized the `is_property` field so this change makes no difference there. The latter subclass never really uses `is_property` but also doesn't have any serialize/deserialize methods, which makes this change moot.

mypy/nodes.py
Michael0x2a: Ooh, good point -- I didn't even know that file existed.

I tried making the change + tried adding an test to one of the fine-grained incremental tests. (I'm pretty unfamiliar with fine-grained incremental stuff though, so let me know if I did it incorrectly.)

mypy/semanal.py
Michael0x2a: Done!

test-data/unit/check-attr.test
Michael0x2a: Also done

test-data/unit/check-overloading.test
Michael0x2a: Ah whoops, I completely forgot to handle this case. Fixed.

mypy/checker.py
Michael0x2a: I'm not sure I'm completely following -- is there now going to be some major difference between FuncItem and FuncBase?

(I also ended up streamlining this code to handle the "overriding a classmethod with an instance method" case more cleanly -- not sure if that change makes the potential conflict better or worse.)

test-data/unit/check-overloading.test
```diff
@@ -2937,3 +2937,416 @@ def add_proxy(x, y):
...
+
+[case testOverloadWithClassMethods]
+from typing import overload
+
+class Wrapper:
+    @overload
+    @classmethod
+    def foo(cls, x: int) -> int: ...
+    @overload
+    @classmethod
```
ilevkivskyi: One more suggestion: could you please add to your TODO list adding few tests to check that overloads work well with self-types (both instance, and class methods using `cls: Type[T]`)?

test-data/unit/fine-grained.test
```diff
@@ -2087,6 +2087,33 @@ main:2: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports"
...
+    @overload
+    def foo(cls, x: int) -> int: ...
+    @overload
+    def foo(cls, x: str) -> str: ...
+[builtins fixtures/classmethod.pyi]
+[out]
+main:3: error: Revealed type is 'builtins.int'
+==
+main:3: error: Revealed type is 'Any'
+main:3: error: No overload variant of "foo" of "Wrapper" matches argument type "int"
```
ilevkivskyi: Yes, this test looks good.

Michael0x2a:
@ilevkivskyi -- ok, done. I went ahead and just added some self-type tests to this commit.

PR #5225, opened 2018-06-16
Run each PEP561 test in its virtualenv dir

ethanhs:
When working on partial stub packages I realized that the changes in #5060 were flaking. This change runs mypy in each temporary directory, which avoids parrallelism/cache issues (sometimes the multiple processes would try to read/write the same cache file). To test this thoroughly, I ran the test 25 times in a row, with no failures.

### DIFF
--- a/mypy/test/testpep561.py
+++ b/mypy/test/testpep561.py
@@ -28,14 +28,16 @@ def check_mypy_run(cmd_line: List[str],
     if venv_dir is not None:
         old_dir = os.getcwd()
         os.chdir(venv_dir)
-    if python_executable != sys.executable:
-        cmd_line.append('--python-executable={}'.format(python_executable))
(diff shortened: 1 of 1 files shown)


mypy/test/testpep561.py
```diff
@@ -25,30 +22,38 @@ def check_mypy_run(cmd_line: List[str],
...
+    if venv_dir is not None:
+        old_dir = os.getcwd()
+        os.chdir(venv_dir)
     if python_executable != sys.executable:
         cmd_line.append('--python-executable={}'.format(python_executable))
     out, err, returncode = mypy.api.run(cmd_line)
     assert out == expected_out, err
     assert err == expected_err, out
     assert returncode == expected_returncode, returncode
+    if venv_dir is not None:
```
JelleZijlstra: Put this in a finally block.

PR #5227, opened 2018-06-17
Add support for partial stub packages

ethanhs:
This should conform to PEP 561 as laid out in
https://www.python.org/dev/peps/pep-0561/#partial-stub-packages.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -946,9 +946,7 @@ class FindModuleCache:
                         # 'partial\n' to make the package partial
                         # Partial here means that mypy should look at the runtime
                         # package if installed.
-                        with open(stub_typed_file) as f:
-                            src = f.read()
-                        if src == 'partial\n':
+                        if fscache.read(stub_typed_file).decode().strip() == 'partial':
                             runtime_path = os.path.join(pkg_dir, dir_chain)
                         third_party_inline_dirs.append((runtime_path, True))
                         # if the package is partial, we don't verify the module, as
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -862,6 +862,10 @@ def _get_site_packages_dirs(python_executable: Optional[str]) -> List[str]:
                                 stderr=subprocess.PIPE).decode())
 
 
+# Search paths are a two-tuple of path and whether to verify the module
+SearchPaths = List[Tuple[str, bool]]
+
+
 class FindModuleCache:
     """Module finder with integrated cache.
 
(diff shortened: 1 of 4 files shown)


mypy/build.py
```diff
@@ -933,13 +937,28 @@ def _find_module(self, id: str, lib_path: Tuple[str, ...],
...
                 stub_components = [stub_name] + components[1:]
                 path = os.path.join(pkg_dir, *stub_components[:-1])
                 if fscache.isdir(path):
-                    third_party_stubs_dirs.append(path)
+                    if fscache.isfile(stub_typed_file):
+                        # Stub packages can have a py.typed file, which must include
+                        # 'partial\n' to make the package partial
+                        # Partial here means that mypy should look at the runtime
+                        # package if installed.
+                        with open(stub_typed_file) as f:
```
JelleZijlstra: Does the fscache not show the content of the file? Should we handle the case where the fscache thinks it exists but it actually doesn't?

mypy/build.py
```diff
@@ -933,13 +937,28 @@ def _find_module(self, id: str, lib_path: Tuple[str, ...],
...
+                            src = f.read()
+                        if src == 'partial\n':
```
JelleZijlstra: Although the PEP does explicitly say `partial\n`, would it be better to just check `src.strip() == 'partial'`? It doesn't seem worth worrying about the exact newline representation.

ethanhs: Ah, I should probably use `fscache.read()` here

mypy/build.py
ethanhs: Ah, I should probably use `fscache.read()` here

mypy/build.py
```diff
@@ -933,13 +937,26 @@ def _find_module(self, id: str, lib_path: Tuple[str, ...],
...
                 path = os.path.join(pkg_dir, *stub_components[:-1])
                 if fscache.isdir(path):
-                    third_party_stubs_dirs.append(path)
+                    if fscache.isfile(stub_typed_file):
+                        # Stub packages can have a py.typed file, which must include
+                        # 'partial\n' to make the package partial
+                        # Partial here means that mypy should look at the runtime
+                        # package if installed.
+                        if fscache.read(stub_typed_file).decode().strip() == 'partial':
+                            runtime_path = os.path.join(pkg_dir, dir_chain)
```
ilevkivskyi: PEP 561 is not 100% clear about this, but I think mypy should also use typeshed if a module is not found in a partial stub package. _Unless I am missing something_, the current implementation only captures one use case when a stub package is used to override an inline/runtime package. But there is another use case allowed by PEP 561 -- override typeshed with a stub package, what to do if the latter is declared partial and a module is not found? I think mypy should look in typeshed.

I mean you have this list 1...5 in the PEP, I think a simple rule is that if a module is not found in (3) but the package is partial we should just continue in normal order to (4) inline/runtime packages, _and then_ (5) typeshed.

ethanhs: I agree the PEP text should be amended to be this way, fortunately for mypy, we almost already do what you describe. I *think* if I just re-order the elements of `candidate_base_dirs` to be in the right order as described by the PEP (which it sadly currently is not), it will "just work".

ethanhs: PR is https://github.com/python/mypy/pull/5256

ilevkivskyi: I will merge this one now, and will review #5256 later (it looks independent).

Issue #5226, opened 2018-06-17
Too strict behavior with --ignore-missing-imports


ilevkivskyi:
I think the `--ignore-missing-imports` currently behaves a bit inconsistently. If I have an `existing` directory with only an empty `__init__.py`, then running mypy with `--ignore-missing-imports` gives this:
```python
import whatever.works
import existing.nonexisting

x = whatever.works.f()  # No errors, as expected
y = existing.nonexisting.g()  # Module has no attribute "nonexisting"
```
At the same time both these work (in the same program, otherwise the second fails):
```python
from whatever import works  # no errors
from existing import nonexisting # no errors
```

I think the first example should work as well, since all non-existing modules should be replaced with a dummy `Any` (and apparently it looks easy to fix, unless this is intentional behavior for some reasons).

EDIT: see https://github.com/python/mypy/issues/5226#issuecomment-397851213

gvanrossum:
I believe this has been discussed before but I don't recall the outcome.
:-( Isn't this the reason we don't like partial packages in typeshed?

ilevkivskyi:
> Isn't this the reason we don't like partial packages in typeshed?

Maybe. But what is the value of that error "Module has no attribute ..."?

ilevkivskyi:
Correction: `from existing import nonexisting` also fails. Id however doesn't fail, if there is an `import existing.nonexisting` elsewhere.

gvanrossum:
For me the deciding difference is whether it could reference something other than a module. `import existing.nonexisting` has to refer to a module, so `--ignore-missing-imports` applies. But `from existing import nonexisting` could refer to a class, function or variable defined in `existing` and as such it is not covered by `--ignore-missing-imports`. The presence of `import existing.nonexisting` elsewhere implies that it refers to a submodule, so then the flag applies.

So I don't necessarily believe that "fixing" this will make things better -- it will just trade false positives for false negatives.

ilevkivskyi:
> So I don't necessarily believe that "fixing" this will make things better -- it will just trade false positives for false negatives.

So do you prefer false positives? I prefer false negatives. Also as Jukka mentioned in the PR some users may be surprised that refactoring `import a.b` into `from a import b` turns it into an error.

gvanrossum:
In the end I agree that we need to solve this. I just wanted to explain
that the original behavior was not that irrational but in fact was
internally consistent. (The new solution will be too of course.)

ilevkivskyi:
OK, I updated the title.

PR #5228, opened 2018-06-17
Make --ignore-missing-imports less strict

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5226

Currently, `import existing_package.non_existing_mod` (with `--ignore-missing-imports`) allows subsequent
```python
from existing_package import non_existing_mod
x = non_existing_mod.f()
```
but doesn't allow a direct call
```python
x = existing_package.non_existing_mod.f()
```
With this change both are allowed. This doesn't solve the problem of partial packages, but makes `--ignore-missing-imports` more consistent with the flag name IMO.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -297,6 +297,21 @@ Here are some more useful flags:
 
 - ``--ignore-missing-imports`` suppresses error messages about imports
   that cannot be resolved (see :ref:`follow-imports` for some examples).
+  This doesn't suppress errors about missing names in successfully resolved
+  modules. For example, if one has the following files::
+
+    package/__init__.py
+    package/mod.py
+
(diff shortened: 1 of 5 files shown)


ilevkivskyi:
> Have you verified that doesn't break anything in internal Dropbox codebases?

Not yet, I will check this in next few hours, I will merge if `testpr` passes.

gvanrossum:
Of course testpr will fail, since it already fails with the current master.
:-( Maybe you can compare the errors you get without and with this PR.

msullivan:
Is there any documentation that should be updated along with this?

ilevkivskyi:
> Is there any documentation that should be updated along with this?

This is a good idea, I will add some docs.

ilevkivskyi:
I was finally able to check, this passes `testpr`.

Issue #4579, opened 2018-02-16
Docs for function overloading are made more complex by using __getitem__ as example


gvanrossum:
I've got a feeling that the docs for overloading could be made easier on the eyes and easier to grasp for first-time readers if they didn't use `__getitem__` as the main example. Overloading is quite useful for non-dunder methods and functions!

PR #5229, opened 2018-06-17
Update overload docs

Michael0x2a:
This pull request:

1.  Moves documentation about overloads into the "More types" section.

    (Rationale: although overloads aren't really a type, I feel it makes sense for them to live somewhere in the "Type System Reference" section instead of the "Miscellaneous" section.)

2.  Modifies the docs to start with a non-dunder example (resolves https://github.com/python/mypy/issues/4579) and simplifies the `__dunder__` example.

3.  Adds some documentation about the "pick the first match" rule.

4.  Removes the now out-of-date note about erasure.


docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+must be adjacent in the code: think of them as one indivisible unit.
+
+The variant bodies must all be empty; only the implementation is allowed
+to contain code. This is because at runtime, the variants are completely
+ignored: they're overridden by the final implementation function.
+
+This means that an overloaded function is still an ordinary Python
+function! There is no automatic dispatch handling: you must manually
+handle the different types in the implementation (usually by using
+``if`` statements and ``isinstance`` checks).
```
ilevkivskyi: I would replace "usually" with "e.g."

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+An overloaded function must consist of two or more overload *variants*
+followed by an *implementation*. The variants and the implementations
+must be adjacent in the code: think of them as one indivisible unit.
+
+The variant bodies must all be empty; only the implementation is allowed
+to contain code. This is because at runtime, the variants are completely
+ignored: they're overridden by the final implementation function.
+
+This means that an overloaded function is still an ordinary Python
+function! There is no automatic dispatch handling: you must manually
```
ilevkivskyi: An exclamation sign is not needed here.

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+handle the different types in the implementation (usually by using
+``if`` statements and ``isinstance`` checks).
+
+If you are adding an overload within a stub file, the implementation
+function should be omitted: stubs do not contain runtime logic.
+
+Type checking calls to overloads
+--------------------------------
```
ilevkivskyi: I would place the section about checking definitions first. I think it is more logical, one first writes a function, then one calls it.

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+
+When you call an overloaded function, mypy will infer the correct
+return type using the provided variants. A call is never type checked
+against the implementation signature. This is why mypy will report calls
+like ``ip_address(4)`` as being invalid even though it matches the
+implementation signature.
+
+If multiple variants end up matching a call, mypy will, for the
+most part, select the return type corresponding to the first
+matching call. For example, consider the following program:
```
ilevkivskyi: I would reformulate this slightly. We can say something like: "Mypy will chose the best matching variant (considering argument kinds and types). If there are more than one equally good match, mypy will choose the first one. An exception is when there are multiple matching due to an `Any` type of some of the arguments, in this case mypy will infer `Any` as the result of the call."

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+The ``summarize([])`` call matches both variants: an empty list could
+be either a ``List[str]`` or a ``List[int]``. In this case, mypy
+will break the tie by picking the first matching variant: ``output``
+will have an inferred type of ``str``. The implementor is responsible
+for making sure ``summarize`` breaks ties in the same way at runtime.
+
+There are a few exceptions to the "pick the first match" rule.
+For example, if multiple variants match due to an argument
+being of type ``Any``, mypy will make the inferred type also
+be ``Any``.
```
ilevkivskyi: I would replace this paragraph with an example showing a call with `Any` to an above function (especially if you reword as I suggest above).

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+
+Mypy will also prohibit you from writing overload variants that are
+inherently unsafely overlapping: for example, writing two variants
+that accept the same arguments but return different types.
```
ilevkivskyi: I think this paragraph belongs to the below section.

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+
+Type checking the implementation
+--------------------------------
```
ilevkivskyi: I would rename this to "Type checking the definition" and include more info here. In particular I would show two examples: one showing the "unsafely overlaps error" (e.g. with `int -> int` and `object -> str`), and one with "will be never matched" (briefly). For first example it would be great to explain the motivation with and example `x: object = 42` -> `f(x)` -> `BOOM!`

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
 
         name_by_id(3)  # int is not the same as UserId
 
+.. _function-overloading:
```
ilevkivskyi: I would add a link to this section at the end of the `Callable` section.

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+with ``Union[int, slice]`` and ``Union[T, Sequence]``.
+
+.. note::
+
+   Due to the "pick the first match" rule, changing the order of your
+   overload variants can change how mypy type checks your program.
+
+   To minimize potential issues, we recommend ordering your variants
+   from most to least specific. Your implementation should also
+   perform ``isinstance`` checks and the like in the same order
```
ilevkivskyi: I would replace "``isinstance`` checks" to "runtime dispatch checks" (or something similar).

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+
+Type checking the implementation
+--------------------------------
+
+The body of an implementation is type-checked against the
+type hints provided on the implementation. For example, in the
+``MyList`` example up above, the code in the body is checked with
+``index: Union[int, slice]`` and a return type ``Union[T, Sequence[T]]``.
+If there are no annotations on the implementation, then the body is
+not type checked.
```
gvanrossum: Maybe clarify that this is different from having all args and return annotated with `Any`. That's standard, but somehow a reminder in this context seems helpful.

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+parameter type ``int`` and the return type ``T`` are compatible with
+``Union[int, slice]`` and ``Union[T, Sequence]`` for the
+first variant. For the second variant it verifies the parameter
+type ``slice`` and the return type ``Sequence[T]`` are compatible
+with ``Union[int, slice]`` and ``Union[T, Sequence]``.
+
+.. note::
+
+   Due to the "pick the first match" rule, changing the order of your
+   overload variants can change how mypy type checks your program.
```
gvanrossum: Maybe also mention that older versions of mypy (up to 0.610) used a different rule here? Or that PEP 484 is currently not clear on this?

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+
+   To minimize potential issues, we recommend ordering your variants
+   from most to least specific. Your implementation should also
```
gvanrossum: How should I order them if they are all disjoint? (E.g. the ip_address example.)

docs/source/more_types.rst
```diff
@@ -178,6 +178,211 @@ create subclasses of these objects either.
...
+
+    # The actual *implementation* of 'ip_address'.
+    # The implementation contains the actual runtime logic.
+    #
+    # It may or may not have type hints. If it does, mypy
+    # will check the body of the implementation against the
+    # type hints.
+    #
+    # Mypy will also check and make sure the signature is
+    # consistent with the provided variants.
```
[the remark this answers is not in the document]
ilevkivskyi: > Did I do something wrong?

It looks like a bug. @Michael0x2a could you take a look, seems something simple.

docs/source/more_types.rst
ilevkivskyi: > Is there still a gradient in how "good" a match is?

Yes, we still use some heuristics, we discussed for some time and agreed it would be impossible to specify all corner cases in a way making everyone happy. So we deliberately leave some room for type checkers in ambiguous corner cases (the cases are quite well specified, but the behavior in each is up to type checker).

Also, @Michael0x2a how about you prepare a PR to peps repo using your original proposal on typing tracking and taking into account the things you learned from implementing all this?

> And what about provided context? Does that play a role?

Currently it is used only if there type variables I think. I think in future we can use it to pick up s better overload, but it is more complex that it seems.

docs/source/more_types.rst
ilevkivskyi: > And the examples above already hint at how the definitions are checked.

This way the logic actually jumps IMO. We start from definitions in introduction, then we switch to checking calls, then then back to definitions. But this is not something critical, I can live with current ordering :-)

docs/source/more_types.rst
gvanrossum: > the cases are quite well specified

Remind me where? PEP 484 explains almost entirely through examples. And the mypy docs don't apply to other type checkers. Are the heuristics only relevant to Protocols (PEP 544)? Then that should be called out specifically.

docs/source/more_types.rst
gvanrossum: Such jumping logic is often more readable than an exhaustive treatment per sub-topic -- this is probably different from how you've learned to write mathematical proofs. :-)

docs/source/more_types.rst
ilevkivskyi: > Remind me where?

Original PEP 484 amendment proposal https://github.com/python/typing/issues/253#issuecomment-389262904

Some amendments to amendment discussion is also here https://github.com/python/mypy/issues/5119

> Are the heuristics only relevant to Protocols (PEP 544)?

No, there are several other corner cases: empty collections, star args, union types.

ilevkivskyi: Also beware, the original proposal is _very_ long. It also discusses motivation, but the PEP 484 PR should include mostly the specification part IMO.

gvanrossum: Possibly it can become a separate PEP, and PEP 484 should be amended with a link to the new PEP? That shouldn't be a big deal (it's not like type checking PEPs seem to attract a lot of attention from the usual crowd of armchair critics).

docs/source/more_types.rst
Michael0x2a: Regarding this section/the surrounding text: I wonder if it might be better to just link to PEP 484, once the overload stuff is accepted there.

I'm currently trying to write the `int -> int` and `object -> str` example -- and it gets mired in the weeds pretty quickly. In particular, I'm finding it challenging to concisely explain why mypy flags this example as being unsafe yet accepts the `List[int] -> int` and `List[str] -> str` example despite the fact that we can trigger the exact same kind of error in both cases.

(More specifically, it'd be trivial for me to write an explanation that describe the design decisions behind this behavior in detail; I'm finding it non-trivial to write an explanation that's satisfying yet concise.)

ilevkivskyi: > I'm finding it non-trivial to write an explanation that's satisfying yet concise.

Yes, I know this is the hardest part.

docs/source/more_types.rst
ilevkivskyi: > Possibly it can become a separate PEP, and PEP 484 should be amended with a link to the new PEP?

I have no preferences, I original thought it would be OK to add this to the PEP 484 because it already contains a section on overloads, but having another mini-PEP is also OK.

gvanrossum: I expect evolving and reviewing the doc as a separate PEP will be easier. Otherwise it'll be a *very* large expansion to an already-approved PEP, and that doesn't sound completely fair to people who reviewed it originally.

docs/source/more_types.rst
gvanrossum: > I wonder if it might be better to just link to PEP 484, once the overload stuff is accepted there.

I'd rather be independent from that -- the purpose here is to be the best/clearest user documentation possible, while the goal of the PEP is to be as precise a specification as possible. Those are often conflicting goals.

Michael0x2a:
@ilevkivskyi and @gvanrossum -- I think this is ready for a second look whenever you have time.

Some notes:

1. I edited the overview at the top to include the text as requested
2. I tried switching the ip address example to something that wouldn't require us to talk about positional-only args, but wasn't really able to find anything I was happy with. (E.g. I thought about making an example where we return date vs datetime objects, but it ended up looking messier). I guess on the plus side, talking about positional-only args do hint to the read that types aren't the only thing that matters.
3. I switched to having the variant bodies use `...`, but decided against removing the blank lines between each variant: IMO it made the overloads look a little intimidatingly dense. That said, I don't feel too strongly about this issue, so I'm happy to remove them if people disagree.
4. I ended up making a new section about how overload variants are type checked. I couldn't really find a way to make it concise, unfortunately -- sorry about that.
5. Regarding the "we pick the best match" thing: I didn't really elaborate on it because I think the current behavior is mostly intuitive (?). If people disagree, I can spell out what exactly we mean by "best". I think it mostly boils down to talking about how star-args are handled?
6. I decided to keep the current topic order: I like using the [spiral approach](https://en.wikipedia.org/wiki/Spiral_approach) when teaching, and the topics basically do that: we start with examples and a high-level overview that communicate the gist of how to use overloads, then drill down into the nitty gritty semantics of calls, variant checks, and implementation checks (in that order).

docs/source/more_types.rst
```diff
@@ -178,6 +183,357 @@ create subclasses of these objects either.
...
+   perform runtime checks (e.g. ``isinstance`` checks) in the same
+   order as the listed variants.
+
+   If your variants have no inherent relationship to each other, they
+   can naturally be listed in any arbitrary order.
+
+Type checking the variants
+--------------------------
+
+Although the "pick the first match" algorithm works well in many cases, it can
```
ilevkivskyi: This introductory sentence can make a confusing impression that this is some kind of limitation of mypys implementation. You can just say something simpler like "Not all possible variant combinations are considered safe by mypy. Let us consider an example of unsafe variants:"

docs/source/more_types.rst
```diff
@@ -178,6 +183,357 @@ create subclasses of these objects either.
...
+    def unsafe_func(x: int) -> int: ...
+
+    @overload
+    def unsafe_func(x: object) -> str: ...
+
+    def unsafe_func(x: object) -> Union[int, str]:
+        if isinstance(x, int):
+            return x
+        else:
+            return "Unsafe!"
```
ilevkivskyi: It is not clear for a typical reader why it is unsafe at this point, use a different word here, maybe "Hm, not really a number".

docs/source/more_types.rst
```diff
@@ -178,6 +183,357 @@ create subclasses of these objects either.
...
+.. code-block:: python
+
+    def should_return_str(x: object) -> str:
+        return unsafe_func(x)
+
+    bad_var = should_return_str(42)
+
+If we examine just the annotated types, it seems as if ``bad_var`` ought to be of
+type ``str``. But if we examine the runtime behavior, ``bad_var`` will actually be
+of type ``int``!
```
ilevkivskyi: Instead of saying the reasoning in words, it would be better to just show a code snippet, that will fail with a `TypeError` at runtime. For example, I would replace your snippet and subsequent explanation by this snippet:
```python
yolo: object = 42
unsafe_func(yolo) + "why not?"  # inferred type for the call is "str", but this will
                                # fail at runtime because the function will actually return 42

File: docs/source/more_types.rst [hunk 17]
```diff
@@ -178,6 +183,357 @@ create subclasses of these objects either.
...
+overlapping when both of the following are true:
+
+1. All of the arguments of the first variant are compatible with the second.
+2. The return type of the first variant is *not* compatible with (e.g. is not a
+   subtype of) the second.
+
+So in this example, the ``int`` argument in the first variant is a subtype of
+the ``object`` argument in the second, yet the ``int`` return type is a subtype of
+``str``. Both conditions are true, so mypy will correctly flag ``unsafe_func`` as
+being unsafe.
```
ilevkivskyi: I really like these three paragraphs they nicely put on formal grounds the above example.

docs/source/more_types.rst
```diff
@@ -178,6 +183,357 @@ create subclasses of these objects either.
...
+
+However, it is unfortunately impossible to detect *all* unsafe overloads without
```
ilevkivskyi: All this long part about multiple inheritance is unnecessary I think. It can be replaced with a single short note at the end of this section like: "Note that mypy intentionally allows unsafe overloads related to multiple inheritance and empty containers. This way mypy avoids lots of false positives (that would otherwise make overloads impractical) by having some rare false negatives."

This is really a discussion that belongs to the PEP, not the docs. Remember that you write this mostly for people who have never heard about overloads, while currently it looks like you are talking to a CS prof: "Please forgive us for being so unsafe!" :-)

docs/source/more_types.rst
```diff
@@ -178,6 +183,357 @@ create subclasses of these objects either.
...
+type ``slice`` and the return type ``Sequence[T]`` are compatible
+with ``Union[int, slice]`` and ``Union[T, Sequence]``.
+
+.. note::
+
+   The overload semantics documented above are new as of mypy 0.620.
+
+   Older versions of mypy used different semantics. In particular, overload
+   variants had to different after erasing types and calls to overloaded
+   functions did not use the "pick the first match" rule.
```
ilevkivskyi: You can mention that mypy previously often silently returned `Any` for many overload calls. So that people will be not surprised by new errors when switching to 0.620.

docs/source/more_types.rst
```diff
@@ -178,6 +183,357 @@ create subclasses of these objects either.
...
+checking overloads: it will *not* report any errors with the above program.
+
+Thankfully, these types of situations are relatively rare. What this
+does mean, however, is that you should exercise caution when using an
+overloaded function in scenarios where it can potentially receive values
+that are an instance of two seemingly unrelated types. For example,
+``Combined()`` is an instance of both ``A`` and ``B``; the empty list 
+``[]`` is an instance of both ``List[int]`` and ``List[str]``.
+
+
```
ilevkivskyi: It looks like an example of "...will be never matched" is missing in this section.

docs/source/more_types.rst
```diff
@@ -178,6 +183,357 @@ create subclasses of these objects either.
...
+    # They are ignored at runtime.
+    #
+    # Prefixing an argument with two underscores tells mypy that
+    # those arguments are positional only.
+
+    @overload
+    def ip_address(__a: int, __b: int, __c: int, __d: int) -> IPv4Address: ...
+    @overload
+    def ip_address(__a: int, __b: int, __c: int, __d: int,
+                   __e: int, __f: int, __g: int, __h: int) -> IPv6Address: ...
```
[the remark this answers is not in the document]
Michael0x2a: I think going with `ip_address(*args, **kwargs)` would ruin the narrative that overloads let you "refine" the type of your function to something more precise. Both options are also a kludge -- in that case, I'd rather use the kludge that teaches the reader something new and useful about mypy rather then handwaving the complication away.

That said, I think a better option would be to find a different example that avoids this issue completely. I'm just that I'm having difficulty coming up with a replacement that isn't totally contrived.

docs/source/more_types.rst
Michael0x2a: Mm, I'm ok with just leaving a short snippet like that if we're ok with also linking to the PEP and telling the reader to read that if they want more details.

But I was under the impression that we wanted the docs to be more self-contained? In that case, I think we really do need at least one example of what an unsafe yet allowed overload could look like. I don't think a reader would really understand what we mean if with just that explanation .

One option I could try is re-using the `summarize` example instead -- that would let me cut out all of the setup related to multiple-inheritance which I think would help cut down the text by a decent amount. I still want to keep the gist of recommendations at the end and some of the exposition though. Empty collections are relatively common and some people do like to use mixins -- I think it's worth giving people a more detailed heads up.

docs/source/more_types.rst
Michael0x2a: Will do

docs/source/more_types.rst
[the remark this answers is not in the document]
ilevkivskyi: I just looked at mypy's own code and there is a single example of overload. It looks like this:
```python
@overload
def preferred_currency(c: PaidCustomer) -> Currency: ...
@overload
def preferred_currency(c: Customer) -> Optional[Currency]: ...
```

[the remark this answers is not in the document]
ilevkivskyi: Implementation may be like this:
```python
def preferred_currency(c: Customer) -> Optional[Currency]:
    if isinstance(c, PaidCustomer):
        # Use currency from subscription plan
    elif in_app_purchases(c):
        # Use most commonly used currency
    else:
        return None
```

docs/source/more_types.rst
ilevkivskyi: > But I was under the impression that we wanted the docs to be more self-contained?

Self-contained doesn't mean complete, in particular here too much attention is devoted to motivation why we allow this. You could try re-using `summarize` if this will help cutting the size of this significantly.

docs/source/more_types.rst
```diff
@@ -178,6 +183,357 @@ create subclasses of these objects either.
...
+    output3 = summarize(some_list)
+
+.. note::
+
+   Due to the "pick the first match" rule, changing the order of your
+   overload variants can change how mypy type checks your program.
+
+   To minimize potential issues, we recommend ordering your variants
+   from most to least specific. Your implementation should also
+   perform runtime checks (e.g. ``isinstance`` checks) in the same
```
ilevkivskyi: Hm, now it seems to me it should be vice-versa here. People typically add annotations to existing code, so it is better to say "Your variants should be listed in the same order as runtime checks (e.g. `isinstance` checks) performed in the implementation."

gvanrossum:
I lost track -- who is waiting for whom here?

Michael0x2a:
@gvanrossum -- this PR is currently waiting on me. I've implemented most of Ivan's feedback but haven't had the chance to sit down and do the last bit of polishing yet.

Michael0x2a:
@ilevkivskyi and @gvanrossum -- Sorry for the delay! I think this should be ready for another look now.

Specific substantial changes made:

1. Swapped out the IP address example for something involving "mouse events". 

    (It's still a little contrived, but whatever -- I can showcase overloads w/ different arities, give an example of a call that matches the impl but not the variants, so good enough.)
2. Added an example/explanation of the "X will never be matched" error.
3. Replaced the example with multiple inheritance with an example about 'summarized' and tried making the exposition at the end less wordy/less "doom-and-gloom".
4. Fleshed out the section about old mypy semantics a little
5. Adjusted the section about recommendations for ordering overload variants to place more emphasis on making sure the runtime impl and variant orders mirror each other.

docs/source/more_types.rst
```diff
@@ -178,6 +183,386 @@ create subclasses of these objects either.
...
+    def mouse_event(x1: int, 
+                    y1: int,
+                    x2: Optional[int] = None,
+                    y2: Optional[int] = None) -> Union[ClickEvent, DragEvent]:
+        if x2 is None and y2 is None:
+            return ClickEvent(x1, y1)
+        elif x2 is not None and y2 is not None:
+            return DragEvent(x1, y1, x2, y2)
+        else:
+            raise Exception("Bad arguments")
```
ilevkivskyi: I think either `ValueError` or `TypError` will be better here.

docs/source/more_types.rst
```diff
@@ -178,6 +183,386 @@ create subclasses of these objects either.
...
```
ilevkivskyi: Same here.

docs/source/more_types.rst
```diff
@@ -178,6 +183,386 @@ create subclasses of these objects either.
...
+handle the different types in the implementation (e.g. by using
+``if`` statements and ``isinstance`` checks).
+
+If you are adding an overload within a stub file, the implementation
+function should be omitted: stubs do not contain runtime logic.
+
+.. note::
+
+   While we can leave the variant body empty using the ``pass`` keyword,
+   the more common convention is to instead use the ellipse (``...``) literal.
```
ilevkivskyi: "ellipsis"

docs/source/more_types.rst
```diff
@@ -178,6 +183,386 @@ create subclasses of these objects either.
...
+
+.. code-block:: python
+
+    from typing import List, overload
+
+    @overload
+    def summarize(data: List[int]) -> int: ...
+
+    @overload
+    def summarize(data: List[str]) -> str: ...
```
ilevkivskyi: How about having a signature that can't be expressed without overloads? For example, the first overload may be `List[int] -> float` and the implementation could return an average for integers, and a most common for strings.

docs/source/more_types.rst
```diff
@@ -178,6 +183,386 @@ create subclasses of these objects either.
...
+result in a discrepency between the inferred type and the actual runtime type
+when we try using it like so:
+
+.. code-block:: python
+
+    some_obj: object = 42
+    unsafe_func(some_obj) + " danger danger"  # Type checks, yet crashes at runtime!
+
+This program type checks according to the annotations, but will actually crash
+at runtime! Since ``some_obj`` is of type ``object``, mypy will decide that
```
ilevkivskyi: The sentence "This program..." just repeats the comment above. I would just remove it. (Especially that we have an explanation in the next sentence.)

docs/source/more_types.rst
```diff
@@ -178,6 +183,386 @@ create subclasses of these objects either.
...
+``unsafe_func``:
+
+.. code-block:: python
+
+    some_list: List[str] = []
+    summarize(some_list) + "danger danger"  # Type safe, yet crashes at runtime!
+
+We run into a similar issue here. This program type checks if we look just at the
+annotations on the overloads. But since ``summarize(...)`` is designed to be biased
+towards returning an int when it receives an empty list, this program will actually
```
ilevkivskyi: I would format this as ` ``int`` ` (or rather ` ``float`` ` if you update the signature as suggested).

docs/source/more_types.rst
```diff
@@ -178,6 +183,386 @@ create subclasses of these objects either.
...
+    from typing import List, overload
+
+    @overload
+    def summarize(data: List[int]) -> int: ...
+
+    @overload
+    def summarize(data: List[str]) -> str: ...
+
+    def summarize(data):
+        if len(data) == 0:
```
ilevkivskyi: `if not data:`

PR #5231, opened 2018-06-18
Use __getattr__ to mark partial stub packages

ilevkivskyi:
There is a problem with annotating large frameworks -- they are large. Therefore it is hard to produce good stubs in a single pass. A possible better workflow would be to allow indicating that a given (sub-)package is incomplete.

I propose to use `__getattr__` for the role of such indicator. A motivation is that currently adding a `__getattr__` to `package/__init__.pyi` already makes `from package import mod` work, but `import package.mod` still fails (plus simplicity of implementation). Here are the rules that I propose:
* One can declare a (sub-)package as incomplete by adding a `__getattr__` to its `__init__.pyi`
* If the return type of this function is `types.ModuleType` or `Any`, we assume that all imports from this (sub-)package succeed.
* Incomplete package can contain a complete subpackage:
```python
# file a/__init__.pyi
from types import ModuleType
def __getattr__(attr: str) -> ModuleType: ...

# file a/b/__init__.pyi
# empty (i.e. complete package)

# file main.py
import a.d  # OK
import a.b.c  # Error module not found
```

Note: these rules apply only to stubs (i.e. `.pyi` files). I add several tests to illustrate this behaviour.
This PR shouldn't give any significant performance penalty because the added parsing/loading only happens when an error would be reported (for our internal workflow the penalty will be zero because of the flags we use).

This PR will allow gradually adding stub modules to a large framework package, without generating loads of false positives for user code.

Note: PEP 561 introduces the notion of a partial stub package, implemented in https://github.com/python/mypy/pull/5227. I think however this is a bit different use case that I don't want to mix with this one for two reasons:
* Partial packages in PEP 561 are mainly focused on interaction between stubs and inline/runtime packages.
* The proposed feature may be also used in typeshed, not only for installed stub packages.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1741,7 +1741,7 @@ class State:
                  # If `temporary` is True, this State is beeing created to just
                  # quickly parse/load the tree, without an intention to further
                  # process it. With this flag, any changes to external state as well
-                 # as error reprting should be avoided.
+                 # as error reporting should be avoided.
                  temporary: bool = False,
                  ) -> None:
         assert id or path or source is not None, "Neither id, path nor source given"
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1738,6 +1738,10 @@ class State:
                  caller_line: int = 0,
                  ancestor_for: 'Optional[State]' = None,
                  root_source: bool = False,
+                 # If `temporary` is True, this State is beeing created to just
+                 # quickly parse/load the tree, without an intention to further
+                 # process it. With this flag, any changes to external state as well
+                 # as error reprting should be avoided.
                  temporary: bool = False,
                  ) -> None:
         assert id or path or source is not None, "Neither id, path nor source given"
@@ -2255,6 +2259,7 @@ def find_module_and_diagnose(manager: BuildManager,
       caller_line: the line number of the import
       ancestor_for: the child module this is an ancestor of, if applicable
       root_source: whether this source was specified on the command line
+      quick: skip any error diagnosis and reporting
 
     The specified value of follow_imports for a module can be overridden
     if the module is specified on the command line or if it is a stub,
@@ -2322,12 +2327,18 @@ def find_module_and_diagnose(manager: BuildManager,
 
 
 def in_partial_package(id: str, manager: BuildManager) -> bool:
+    """Check if a missing module can potentially be a part of a package.
+
+    This checks if there is any existing parent __init__.pyi stub that
+    defines a module-level __getattr__ (a.k.a. partial stub package).
+    """
     while '.' in id:
         parent, _ = id.rsplit('.', 1)
         parent_mod = None
         if parent in manager.modules:
             parent_mod = manager.modules[parent]
         else:
+            # Parent is not in build, try quickly if we can find it.
             try:
                 parent_st = State(id=parent, path=None, source=None, manager=manager,
                                   temporary=True)
@@ -2339,6 +2350,7 @@ def in_partial_package(id: str, manager: BuildManager) -> bool:
             if parent_mod.is_partial_stub_package:
                 return True
             if not parent_mod.is_stub:
+                # Bail out soon, only stub package can be partial, not runtime one.
                 return False
         id = parent
     return False
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1738,7 +1738,7 @@ class State:
                  caller_line: int = 0,
                  ancestor_for: 'Optional[State]' = None,
                  root_source: bool = False,
-                 # If `temporary` is True, this State is beeing created to just
+                 # If `temporary` is True, this State is being created to just
                  # quickly parse/load the tree, without an intention to further
                  # process it. With this flag, any changes to external state as well
                  # as error reporting should be avoided.
@@ -1764,7 +1764,7 @@ class State:
             try:
                 path, follow_imports = find_module_and_diagnose(
                     manager, id, self.options, caller_state, caller_line,
-                    ancestor_for, root_source, temporary)
+                    ancestor_for, root_source, skip_diagnose=temporary)
             except ModuleNotFound:
                 if not temporary:
                     manager.missing_modules.add(id)
@@ -2249,9 +2249,10 @@ def find_module_and_diagnose(manager: BuildManager,
                              caller_line: int = 0,
                              ancestor_for: 'Optional[State]' = None,
                              root_source: bool = False,
-                             quick: bool = False) -> Tuple[str, str]:
+                             skip_diagnose: bool = False) -> Tuple[str, str]:
     """Find a module by name, respecting follow_imports and producing diagnostics.
 
+    If the module is not found, then the ModuleNotFound exception is raised.
     Args:
       id: module to find
       options: the options for the module being loaded
@@ -2259,7 +2260,8 @@ def find_module_and_diagnose(manager: BuildManager,
       caller_line: the line number of the import
       ancestor_for: the child module this is an ancestor of, if applicable
       root_source: whether this source was specified on the command line
-      quick: skip any error diagnosis and reporting
+      skip_diagnose: skip any error diagnosis and reporting (but ModuleNotFound is
+          still raised if the module is missing)
 
     The specified value of follow_imports for a module can be overridden
     if the module is specified on the command line or if it is a stub,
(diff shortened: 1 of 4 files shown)


mypy/semanal_pass1.py
```diff
@@ -154,6 +154,17 @@ def visit_func_def(self, func: FuncDef) -> None:
         func.is_conditional = sem.block_depth[-1] > 0
         func._fullname = sem.qualified_name(func.name())
         at_module = sem.is_module_scope()
+        if (at_module and func.name() == '__getattr__' and
+                self.sem.cur_mod_node.is_package_init_file() and self.sem.cur_mod_node.is_stub):
+            if isinstance(func.type, CallableType):
+                ret = func.type.ret_type
+                if isinstance(ret, UnboundType) and not ret.args:
+                    sym = self.sem.lookup_qualified(ret.name, func, suppress_errors=True)
+                    # We only interpred package as partial if __getattr__ return type
```
ethanhs: `We only interpred` -> `We only interpret a`

test-data/unit/check-modules.test
```diff
@@ -2173,6 +2173,103 @@ from c import x
 x = str()
 y = int()
 
+[case testModuleGetattrInit1]
+from a import b
+
+x = b.f()
+[file a/__init__.pyi]
+from typing import Any
+def __getattr__(attr: str) -> Any: pass
```
ethanhs: Don't we usually use an ellipsis for stubs/tests?

mypy/build.py
```diff
@@ -1738,6 +1738,11 @@ def __init__(self,
                  caller_line: int = 0,
                  ancestor_for: 'Optional[State]' = None,
                  root_source: bool = False,
+                 # If `temporary` is True, this State is beeing created to just
```
gvanrossum: beeing -> being

mypy/build.py
```diff
@@ -1759,9 +1764,10 @@ def __init__(self,
             try:
                 path, follow_imports = find_module_and_diagnose(
                     manager, id, self.options, caller_state, caller_line,
-                    ancestor_for, root_source)
+                    ancestor_for, root_source, temporary)
```
gvanrossum: I'd use `quick=temporary` here since it's an optional flag. (Though maybe more of the args should use keyword=value then?

mypy/build.py
```diff
@@ -2252,6 +2259,7 @@ def find_module_and_diagnose(manager: BuildManager,
       caller_line: the line number of the import
       ancestor_for: the child module this is an ancestor of, if applicable
       root_source: whether this source was specified on the command line
+      quick: skip any error diagnosis and reporting
```
gvanrossum: It's kind of odd to have a flag that suppresses the `and_diagnose` part of the function's name. :-) Also `quick` reminds (me) of `--quick-and-dirty` but it has nothing to do with that. Also might want to clarify that `ModuleNotFound` is still raised. (Hm, the docstring should probably mention that that is raised in the first place?)

mypy/build.py
```diff
@@ -2283,7 +2291,8 @@ def find_module_and_diagnose(manager: BuildManager,
                     and not options.follow_imports_for_stubs)  # except when they aren't
                 or id == 'builtins'):  # Builtins is always normal
             follow_imports = 'normal'
-
+        if quick:
+            return (path, follow_imports)
```
gvanrossum: This is a little odd -- it's really just skipping the next if-elif block (up to line 2309 below). Would it be clearer to write it as
```py
if quick:
    pass
elif follow_imports == 'silent':
    ...
elif ...
```
? Alternatively
```py
if not quick:
    if follow_imports == 'silent':
        ...
    elif ...
```

mypy/build.py
```diff
@@ -2315,6 +2326,36 @@ def find_module_and_diagnose(manager: BuildManager,
...
 
+def in_partial_package(id: str, manager: BuildManager) -> bool:
+    """Check if a missing module can potentially be a part of a package.
+
+    This checks if there is any existing parent __init__.pyi stub that
+    defines a module-level __getattr__ (a.k.a. partial stub package).
+    """
+    while '.' in id:
+        parent, _ = id.rsplit('.', 1)
+        parent_mod = None
```
gvanrossum: Maybe move this into the `except` block (L2345)? I was looking whether parent_mod is always defined (because it's not set in the `except` clause) and this felt too far back up. A different way to structure this would be `parent_mod = manager.modules.get(parent)` followed by `if parent_mod is None`.

mypy/build.py
```diff
@@ -2315,6 +2326,36 @@ def find_module_and_diagnose(manager: BuildManager,
...
+        else:
+            # Parent is not in build, try quickly if we can find it.
+            try:
+                parent_st = State(id=parent, path=None, source=None, manager=manager,
+                                  temporary=True)
+            except (ModuleNotFound, CompileError):
+                pass
+            else:
+                parent_mod = parent_st.tree
+        if parent_mod:
```
gvanrossum: `is not None`

mypy/nodes.py
```diff
@@ -256,6 +260,7 @@ def serialize(self) -> JsonDict:
                 'names': self.names.serialize(self._fullname),
                 'is_stub': self.is_stub,
                 'path': self.path,
+                'is_partial_stub_package': self.is_partial_stub_package
```
gvanrossum: Trailing comma.

mypy/semanal_pass1.py
gvanrossum: Also, `if` -> `if the`

test-data/unit/check-modules.test
```diff
@@ -2173,6 +2173,103 @@ from c import x
...
+
+[case testModuleGetattrInit5b]
+from a.b import f
+
+x = f()
+[file a/__init__.pyi]
+def __getattr__(attr: str) -> int: pass
+[out]
+main:1: error: Cannot find module named 'a.b'
+main:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
```
gvanrossum: Would it be too much to ask to add a note pointing to the presence of `__getattr__` with the wrong return type?

test-data/unit/check-modules.test
```diff
@@ -2173,6 +2173,103 @@ from c import x
...
+import a.d  # OK
+[file a/__init__.pyi]
+from typing import Any
+def __getattr__(attr: str) -> Any: pass
+[file a/b/__init__.pyi]
+# empty (i.e. complete subpackage)
+[builtins fixtures/module.pyi]
+[out]
+main:1: error: Cannot find module named 'a.b.c'
+main:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
```
gvanrossum: FWIW when it's about a subpackage, the part of the note about setting MYPYPATH does not apply. I'm in general not a fan of this flag -- does it really help users that much? CC: @JukkaL

mypy/build.py
ilevkivskyi: > Though maybe more of the args should use keyword=value then?

I would do this for `quick` only, because it is a bit "more special" than other.

mypy/build.py
ilevkivskyi: OK, I will rename it to `skip_diagnose` to be more explicit.

test-data/unit/check-modules.test
ilevkivskyi: I can create a follow-up issue after we are done with this PR, unfortunately it will require some efforts to implement (and most likely a new flag on `MypyFile`, that I would rather avoid)

test-data/unit/check-modules.test
ilevkivskyi: > FWIW when it's about a subpackage, the part of the note about setting MYPYPATH does not apply.

This is not something I changed, so I would make it in a separate PR. But I actually also don't like suggesting to set `MYPYPATH`, I think this adds more confusion than help.

ilevkivskyi:
@gvanrossum 
> Could you also add some tests that verify module-specific settings of `ignore_missing_imports` work correctly?

I am not sure what do you mean here. Could you please show an example of what you would like to test?

test-data/unit/check-modules.test
gvanrossum: OK, I filed https://github.com/python/mypy/issues/5241.

mypy/build.py
```diff
@@ -2249,17 +2249,19 @@ def find_module_and_diagnose(manager: BuildManager,
                              caller_line: int = 0,
                              ancestor_for: 'Optional[State]' = None,
                              root_source: bool = False,
-                             quick: bool = False) -> Tuple[str, str]:
+                             skip_diagnose: bool = False) -> Tuple[str, str]:
     """Find a module by name, respecting follow_imports and producing diagnostics.
 
+    If the module is not found, then the ModuleNotFound exception is raised.
     Args:
```
gvanrossum: Conventionally there's a blank line before `Args:`.

test-data/unit/check-modules.test
gvanrossum: OK.

ilevkivskyi: I opened https://github.com/python/mypy/issues/5276 for this.

ilevkivskyi:
@gvanrossum I think I implemented all the requested changes. Is there something here that needs to be done?

PR #5233, opened 2018-06-18
Fix overload interaction with Any

### LINKED ISSUE #5232: Unexpected type 'unicode' for os.path.join() with (Any, str) args
The revealed type for the `os.path.join` call in the program below is `unicode`, even though I'd expect either `Any` or `str`, as the arguments have types `Any` and `str`:

```py
import os
def f(): pass
reveal_type(os.path.join(f(), 'x'))  # builtins.unicode
```

Here's a self-contained example:

```py
# a.pyi
from typing import overload
@overload
def join(__p1: bytes, *p: bytes) -> bytes: ...
@overload
def join(__p1: unicode, *p: unicode) -> unicode: ...
```

```py
# program.py
from a import join
def f(): pass
reveal_type(join(f(), 'x'))  # builtins.unicode
```

Run as `mypy -2 program.py`.

@Michael0x2a Do you have time to have a look at this? This is causing issues in internal Dropbox repos. If you are busy, I can try to fix this. This was apparently introduced by f61c2baac5915ab680c2f1a9858b7f608262a00e.

ilevkivskyi:
Fixes #5232 

This also makes few minor amendments to `pythoneval` tests.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1307,7 +1307,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             if all(is_subtype(ret_type, return_types[-1]) and
                    is_subtype(return_types[-1], ret_type)
                    for ret_type in return_types[:-1]):
-                # The last match is a supertype of all the previous ones, so it's safe
+                # The last match is mutually compatible with all previous ones, so it's safe
                 # to return that inferred type.
                 return return_types[-1], inferred_types[-1]
             else:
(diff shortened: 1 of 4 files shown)


mypy/checkexpr.py
```diff
@@ -1304,7 +1304,9 @@ def infer_overload_return_type(self,
             return None
         elif any_causes_overload_ambiguity(matches, return_types, arg_types, arg_kinds, arg_names):
             # An argument of type or containing the type 'Any' caused ambiguity.
-            if all(is_subtype(ret_type, return_types[-1]) for ret_type in return_types[:-1]):
+            if all(is_subtype(ret_type, return_types[-1]) and
+                   is_subtype(return_types[-1], ret_type)
+                   for ret_type in return_types[:-1]):
                 # The last match is a supertype of all the previous ones, so it's safe
                 # to return that inferred type.
```
Michael0x2a: We should probably update this comment. Other then that, lgtm

ilevkivskyi: OK, will do.

ilevkivskyi:
@JukkaL This reduces the number of errors in internal code bases, only five left in C (and all look like the same error). S however still has lots of errors because of typeshed.

Issue #4619, opened 2018-02-22
Errors when using overloaded functions and type variables


mikez302:
I am using the latest version of mypy from master (mypy 0.570-dev-8f253c2338863c9de5329053890d26a966fc3d7c) and Python 3.6.4. I have a file with the following code:
```python
from typing import overload, List, TypeVar, Union

T = TypeVar('T', bytes, str)

@overload
def make_result(x: T) -> T: ...
@overload
def make_result(x: int) -> List: ...
def make_result(x: Union[bytes, str, int]) -> Union[bytes, str, List]:
    if isinstance(x, int):
        return [x]
    else:
        return x
```
When I run it through mypy, it gives me the following errors:
```
overload_test.py:9: error: Overloaded function implementation does not accept all possible arguments of signature 1
overload_test.py:9: error: Overloaded function implementation cannot produce return type of signature 1
```
What does that mean? I'm guessing that "signature 1" means the first overload, involving the type variable. If I switch around the overloads, I get similar errors about "signature 2".

But what possible arguments does my implementation not accept? The only types that the parameter can have are `bytes` and `str` (from the first overload) and `int` (from the second overload). As far as I know, `Union[bytes, str, int]` covers all of those.

Likewise with the return type. The overloads have possible return types of `bytes`, `str`, and `List`, and `Union[bytes, str, List]` covers all of those.

I am aware that the implementation can accept some combinations that none of my overloads can accept (like `(int) -> str`), but I don't think that should be an error because I'm not actually calling the function with any of those combinations. And if I replace the first signature with two signatures, not involving a generic type (`(bytes) -> bytes` and `(str) -> str`), I have a similar situation, where the implementation can accept combinations that the overloads can't accept, but I get no errors.

I thought that this may be caused by my own misunderstanding of type variables and generics, so I changed the type hints in the function to look like `def make_result(x: Union[T, int]) -> Union[T, List]:` to see what would happen. I got this error:
```
overload_test.py:9: error: Type variable mismatch between overload signature 2 and implementation
```
What type variable mismatch is that referring to?

I know that I have filed some bugs that were duplicates of existing bugs, or errors caused by my own incorrect usage of mypy. I am sorry about that. But this time, I really think I found a new bug.

mikez302:
@gvanrossum thanks for the response. I think I can find a way to deal with it for now.

PR #5236, opened 2018-06-19
Make overload impl checks correctly handle TypeVars and untyped impls

Michael0x2a:
This pull request fixes https://github.com/python/mypy/issues/4619 as well as an unrelated bug where if the overload implementation was untyped, we only check to see if there are unsafe overlaps between the *first* overload alternative and the rest.


mypy/checker.py
```diff
@@ -451,43 +469,36 @@ def check_overlapping_overloads(self, defn: OverloadedFuncDef) -> None:
...
+                else:
+                    impl = impl_type
 
                 # Is the overload alternative's arguments subtypes of the implementation's?
-                if not is_callable_compatible(impl_type, sig1,
+                if not is_callable_compatible(impl, sig1,
                                               is_compat=is_subtype,
-                                              ignore_return=True):
+                                              ignore_return=True,
+                                              unify_generics=False):
```
ilevkivskyi: This function has already too many arguments. Is there a way to avoid `unify_generics`? Like if you already unified above, the unification in `is_callable_compatible` will be a no-op?

Michael0x2a: Huh, that's weird. I could have sworn we needed this parameter, but I tried removing it and everything still worked. Maybe it was a holdover from one of my previous attempts at tackling the bug?

Anyways, I updated the PR to remove this param. Regarding the number of arguments: I'm planning on refactoring and splitting up `is_callable_compatible` in the upcoming PR for improving error messages -- that should help us cut down on the number of params/ad-hoc special casing.

Issue #5234, opened 2018-06-19
Running the PEP 561 tests causes FileNotFoundErrors on Windows


Michael0x2a:
I'm currently getting a few `FileNotFoundError` test failures when I try running the mypy tests from master -- I'm guessing [this PR](https://github.com/python/mypy/pull/5225) is most likely the culprit?

The test outputs are a little long, so I included them in this gist: https://gist.github.com/Michael0x2a/5d022c29a305dab2f028bab84271b213

I'm currently running mypy on Windows 10. Also, I'm not sure if any of this matters or not, but I can also confirm I get the same results whether or not I use powershell or cmd, whether or not I'm running in administrator mode... I'm happy to run more experiments to help narrow down the issue if needed.

@ethanhs -- any ideas?

ethanhs:
Did you pull master since https://github.com/python/mypy/pull/5225 ?

Michael0x2a:
Yeah -- I pulled about 10-15 min ago.

ethanhs:
Ah, I missed that the first time looking at the output. I will take a closer look soon.

ethanhs:
Hm, that is unfortunate. Would either of you be able to drop into pdb before this line https://github.com/python/mypy/blob/master/mypy/test/testpep561.py#L65 and tell me if the path in `python_executable` actually exists on disk?

Michael0x2a:
It seems apparently not. The temp folder itself exists, but appears to be empty. (So, `C:\Users\Michael\AppData\Local\Temp\tmpw5ox88q1` exists, but not `C:\Users\Michael\AppData\Local\Temp\tmpw5ox88q1\Scripts` or `C:\Users\Michael\AppData\Local\Temp\tmpw5ox88q1\Scripts\python`.

Michael0x2a:
Out of curiosity, I tried modifying the `virtualenv` function directly so it prints the output of `run_command` if the return code isn't zero -- it seems the issue was that I didn't have the `virtualenv` package installed. I tried pip-installing it, and everything is working again!

So I guess this is really my fault -- that package is apparently listed in `test-requirements.txt` and I admittedly haven't looked at that file in a long while...

It might be nice if we added a custom error message though or maybe add the missing error code check in the `virtualenv` function.

gvanrossum:
Also (OT) virtualenv is no longer needed, use venv instead, it's in the
stdlib.

ethanhs:
[...]

PR #5237, opened 2018-06-19
Give output of failed venv creation in PEP 561 tests

ethanhs:
This made it harder to realize what was going wrong in #5234 (sorry Michael and Ivan).

I also added a note about why we need virtualenv.


JelleZijlstra:
To make this kind of bug less likely, should we make `run_command` default to throwing an exception if the command fails?

ethanhs:
@JelleZijlstra I do think that it may be a good idea. I think it will be better for that to be in a separate PR.

Issue #5238, opened 2018-06-19
Wrong overload item picked when strict optional disabled


JukkaL:
When type checking the following two files, the revealed type is `unicode` though I'd expect `str` -- but only when using `-2 --no-strict-optional`:

```py
# t.py
from typing import Optional
from m import relpath
a = '' # type: Optional[str]
reveal_type(relpath(a))  # unicode
```

```py
# m.py
from typing import overload
@overload
def relpath(path: str) -> str: ...
@overload
def relpath(path: unicode) -> unicode: ...
```

The issue also happens with `os.path.relpath`.

To reproduce:

```
$ mypy -2 --no-strict-optional t.py
t.py:4: error: Revealed type is 'builtins.unicode'
```

The issue seems to be related to using an optional type in a file which doesn't use strict optional checking.

ilevkivskyi:
I am working on this now.

PR #5239, opened 2018-06-19
Only apply overloading union math to "real" unions

ilevkivskyi:
By "real" unions I mean unions with more than one relevant item.

Fixes https://github.com/python/mypy/issues/5238

(I have found another bug in union math logic while looking at this code, but it is a bit more subtle, so we will fix it separately).

### DIFF
--- a/test-data/unit/python2eval.test
+++ b/test-data/unit/python2eval.test
@@ -401,3 +401,13 @@ def f():  # no annotation
 path = 'test'
 path = os.path.join(f(), 'test.py')
 [out]
+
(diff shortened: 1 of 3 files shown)


test-data/unit/python2eval.test
```diff
@@ -401,3 +401,13 @@ def f():  # no annotation
 path = 'test'
 path = os.path.join(f(), 'test.py')
 [out]
+
+[case testOverloadRelpathRegression]
```
JukkaL: This test case seems unnecessary as we have two test cases that cover this already.

Issue #5240, opened 2018-06-19
Union math is too greedy


ilevkivskyi:
Currently union math can produce excessively wide types for overload calls. Example:
```python
from typing import Union, overload

@overload
def f(x: Union[int, str]) -> int: ...
@overload
def f(x: object) -> object: ...
def f(x):
    pass

x: Union[int, str]
reveal_type(f(x))  # object???
```
A possible solution would be to check if the "normal way" generates a better result ad return it instead of returning soon. But maybe we should actually re-think the union math algorithm. I will spend some time thinking about it, and will make a PR if I will find an easy solution.

Michael0x2a:
Another possible solution would be to take the call argument types, and produce multiple copies of where we vary any arguments that have a union type. We then take each copy and try finding the first match then union those return types together.

So here, we'd try to match `f(int)` and `f(str)` then union the return types together to get `Union[int, int]`, which simplifies to the expected result.

This algorithm would also work with [the example](https://github.com/python/mypy/issues/4063#issuecomment-390857979) that convinced me to do union math first, rather then second.

The only downside is that this would end up being pretty expensive in the case where the call contains many args with unions, or very large unions. I guess it probably wouldn't be too hard to add in some checks to bail out if union math looks expensive though.

ilevkivskyi:
> The only downside is that this would end up being pretty expensive in the case where the call contains many args with unions, or very large unions.

Yes. Choosing best from unioned and normal is also a bit slower than currently. What you propose is exactly what I was thinking about, but it may be actually quadratic in worst case and will only buy us more precise types in some corner cases. I therefore think we should go with the simpler solution for now, if someone will ask we could invent something more sophisticated.

I already have branch ready, will make a PR in few minutes.

Michael0x2a:
> but it may be actually quadratic in worst case and will only buy us more precise types in some corner cases.

Yeah -- I was thinking we could probably steal some of the logic from the current implementation which effectively prevents union math from happening if two or more arguments end up potentially varying. (E.g. we currently mandate the unioned result must have only a single union argument).

We could borrow that same idea to make sure we only destructure a single union argument to avoid entering into the quadratic case (even if the call contains multiple unions) -- the cost would be that we'd have to do an extra linear scan or two.

But in any case, I'm also ok with just going with the simpler solution for now. Union math is supposed to be a convenience, anyways -- if the user is doing something more complicated, the workarounds are easy enough.

PR #5242, opened 2018-06-19
Choose best type when working with unioned overloads

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5240

This is a simple straightforward fix.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1129,6 +1129,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         erased_targets = None  # type: Optional[List[CallableType]]
         unioned_result = None  # type: Optional[Tuple[Type, Type]]
         unioned_errors = None  # type: Optional[MessageBuilder]
+        union_success = False
         if any(isinstance(arg, UnionType) for arg in arg_types):
             erased_targets = self.overload_erased_call_targets(plausible_targets, arg_types,
                                                                arg_kinds, arg_names, context)
@@ -1142,8 +1143,8 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                                                  callable_name=callable_name,
                                                  object_type=object_type)
                 if not unioned_errors.is_errors():
-                    # Success! Stop early.
-                    return unioned_result
+                    # Success! But we need to see maybe normal procedure gives a narrower type.
+                    union_success = True
 
         # Step 3: If the union math fails, or if there was no union in the argument types,
         #         we fall back to checking each branch one-by-one.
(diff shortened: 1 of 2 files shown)


mypy/checkexpr.py
```diff
@@ -1143,17 +1144,23 @@ def check_overload_call(self,
                                                  callable_name=callable_name,
                                                  object_type=object_type)
                 if not unioned_errors.is_errors():
-                    # Success! Stop early.
-                    return unioned_result
+                    # Success! But we need to see maybe normal procedure gives a narrower type.
+                    union_success = True
 
         # Step 3: If the union math fails, or if there was no union in the argument types,
         #         we fall back to checking each branch one-by-one.
```
Michael0x2a: We should probably tweak this comment -- maybe just something simple like

    # Step 3: We try checking each branch one-by-one

?

I don't think it matters too much, as long as we remove the "if the union math fails" bit.

mypy/checkexpr.py
```diff
@@ -1143,17 +1144,23 @@ def check_overload_call(self,
                                                  callable_name=callable_name,
                                                  object_type=object_type)
                 if not unioned_errors.is_errors():
-                    # Success! Stop early.
-                    return unioned_result
+                    # Success! But we need to see maybe normal procedure gives a narrower type.
```
Michael0x2a: "need to see maybe normal procedure" -> "need to see if maybe the normal procedure"

mypy/checkexpr.py
```diff
@@ -1143,17 +1144,23 @@ def check_overload_call(self,
...
+                    union_success = True
```
Michael0x2a: I think we can simplify this if statement to `union_success = not unioned_errors.is_errors()`. We'd also need to change the comment above to `# Record if we succeeded. Next we need to see if maybe...`

mypy/checkexpr.py
```diff
@@ -1143,17 +1144,23 @@ def check_overload_call(self,
...
-            # Success! Stop early.
-            return inferred_result
+            # Success! Stop early by returning the best among normal and unioned.
+            if not union_success:
+                return inferred_result
+            else:
+                assert unioned_result is not None
+                if is_subtype(inferred_result[0], unioned_result[0]):
+                    return inferred_result
+                return unioned_result
```
Michael0x2a: Actually, just had another thought -- I think it would be clearer (and a little more efficient) if we added logic to return early if `union_success` is true somewhere around here. Currently, we return early only if the normal algorithm also succeeds.

(The logic currently does the right thing because step 5 happens to do what we want, but step 4 and down are supposed to be failure cases.)

PR #5245, opened 2018-06-19
fine-grained: Don't strip ForStmt.index_type

msullivan:
index_type is populated during parsing, so stripping it
causes trouble.

Also *do* strip inferred_iterator_type, which is populated during
checking.

### DIFF
--- a/mypy/server/aststrip.py
+++ b/mypy/server/aststrip.py
@@ -233,8 +233,8 @@ class NodeStripVisitor(TraverserVisitor):
         node.imported_names = []
 
     def visit_for_stmt(self, node: ForStmt) -> None:
-        node.index_type = None
         node.inferred_item_type = None
+        node.inferred_iterator_type = None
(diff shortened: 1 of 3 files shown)


mypy/server/aststrip.py
```diff
@@ -233,8 +233,8 @@ def visit_import_all(self, node: ImportAll) -> None:
         node.imported_names = []
 
     def visit_for_stmt(self, node: ForStmt) -> None:
-        node.index_type = None
```
JukkaL: This is not correct, since we mutate `index_type` in the semantic analyzer. Instead, you can add a new attribute called `unanalyzed_index_type` that contains the original type from the parser, and then we can keep this assignment here. This is similar to how we deal with annotations in `AssignmentStmt` nodes.

msullivan: Fixed

PR #5248, opened 2018-06-20
Drop support for Python 3.3

srittau:
See also python/typeshed#2258 and python/typeshed#2260.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -12,14 +12,15 @@ from waiter import Waiter, LazySubprocess
 
 
 def get_versions() -> List[str]:
+    # generates list of python versions to use.
+    # For Python2, this is only [2.7].
+    # Otherwise, it is [3.x, ..., 3.1, 3.0], where x is the version
+    # of the running interpreter.
     major = sys.version_info[0]
     minor = sys.version_info[1]
     if major == 2:
         return ['2.7']
     else:
-        # generates list of python versions to use.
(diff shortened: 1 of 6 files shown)


runtests.py
```diff
@@ -175,7 +176,7 @@ def add_stubs(driver: Driver) -> None:
 
     modules = {'typing'}
     # TODO: This should also test Python 2, and pass pyversion accordingly.
-    for version in ["2and3", "3", "3.3", "3.4", "3.5"]:
```
srittau: I would have added "3.6" and "3.7", but that causes the tests to fail with the following:

```
FAILURE  #3 check stubs

mypy: can't find module 'contextvars'
```

This should also probably use a glob, so adding directories is not forgotten in the future.

runtests.py
```diff
@@ -175,7 +176,7 @@ def add_stubs(driver: Driver) -> None:
...
+    for version in ["2and3", "3", "3.5"]:
```
srittau: I would have added "3.6" and "3.7", but that causes the tests to fail with the following:

```
FAILURE  #3 check stubs

mypy: can't find module 'contextvars'
```

This should also probably use a glob, so adding directories is not forgotten in the future.

JelleZijlstra: I think @ethanhs was working recently on fixing this code.

gvanrossum:
Thanks!

Issue #5246, opened 2018-06-19
Two more issues with overloads and --no-strict-optional


ilevkivskyi:
Here are some _new false positives_ with overloads, based on real code samples with `map` in Python 2.
The bug(s) only appear if the strict optional is disabled:
```python
from typing import TypeVar, Iterable, Tuple, List, overload, Callable
S = TypeVar('S')
T1 = TypeVar('T1')
T2 = TypeVar('T2')

@overload
def map1(func: Callable[[T1], S], iter1: Iterable[T1]) -> List[S]: ...
@overload
def map1(func: None, iter1: Iterable[T1]) -> List[T1]: ...
def map1(*args, **kwargs):
    pass

a: List[int]
b: List[str]

for x in map1(None, a):  # false positive: Need type annotation for 'x'
    pass

@overload
def map2(func: Callable[[T1, T2], S],
        iter1: Iterable[T1],
        iter2: Iterable[T2]) -> List[S]: ...
@overload
def map2(func: None,
        iter1: Iterable[T1],
        iter2: Iterable[T2]) -> List[Tuple[T1, T2]]: ...
def map2(*args, **kwargs):
    pass

for y, z in map2(None, a, b):  # false positive: '<nothing>' object is not iterable
    pass
```

Michael0x2a:
A temporary workaround would be to swap the order of the overloads so the one with `func: None` always comes first.

I'll look into adding a more robust fix.

PR #5252, opened 2018-06-20
Force-enable strict optional only when checking for unsafe overlaps

Michael0x2a:
Resolves https://github.com/python/mypy/issues/5246

Previously, we force-enabled strict optional checks when performing all overload definition checks. This ended up causing mypy to miss some errors in definitions when `--no-strict-optional` mode is enabled.

This commit will now force-enable strict optional only when checking if overloads are unsafely overlapping: we use the existing mode when checking if one variant completely eclipses another or when checking the implementation.


Michael0x2a:
As a note, these tests will currently fail. They should start passing once https://github.com/python/typeshed/pull/2261 is merged and I do the typeshed sync dance.

Issue #5249, opened 2018-06-20
Wrong overload resolution with Any union item in argument


JukkaL:
This fragment derived from production code causes a false positive:

```py
def f(): pass  # No annotation
t = False
t &= f() and True  # Incompatible types in assignment 
                   # (expression has type "int", variable has type "bool")
```

Here's a self-contained repro:

```py
from typing import overload, Union, Any

class C:
    def f(self, other: C) -> C: ...

class D(C):
    @overload  # type: ignore
    def f(self, other: D) -> D: ...
    @overload
    def f(self, other: C) -> C: ...
    def f(self, other): ...

x: D
y: Union[D, Any]
reveal_type(x.f(y))  # C
```

I'd expect the inferred type for `x.f(y)` to be `Union[D, Any]` (or maybe `D`, though it's slightly less correct arguably).

ilevkivskyi:
I think we should switch to a new union math algorithm, I have some ideas how to make it fast enough to replace the current one (that is apparently to dangerous). I will give it a try now.

JukkaL:
Another real-world example where things go wrong:

```py
from typing import Union, Any, List
i: Union[int, Any]
a: List[int]
reveal_type(a[i])  # Union[int, List[int]]
```

JelleZijlstra:
I think that one is actually correct: if `i` is a slice object, it will return `List[int]`.

JukkaL:
`List[int]` is inconsistent with how `Any` works with overloads generally -- if there is ambiguity due to `Any` types in overload resolution, we tend to fall back to `Any` to avoid false positives. False negatives are preferred over false positives when `Any` types are involved. Here is a simple analogy:

```py
from typing import Union, Any
class A:
    x: str
class B:
    x: int
a: Union[A, Any]
reveal_type(a.x)  # Union[str, Any]
```

We don't infer `Union[str, int]` as the type of `a.x` even though `int` is a plausible type for `a.x` -- since it could well be something else. Similarly, the precise type of `i` could actually be just `int` (though we can't tell for certain), and assuming that it can be a slice would be wrong.

ilevkivskyi:
The PR I am working will produce (after some tweaks that I didn't push yet) `Union[int, Any]` for this example:
```python
from typing import Union, Any, List
i: Union[int, Any]
a: List[int]
reveal_type(a[i]) # Union[int, Any]
```
@JukkaL does this make sense to you?

ilevkivskyi:
[...]

Issue #5243, opened 2018-06-19
One more bug with union math logic


ilevkivskyi:
Well, I have found another bug with union math in overloads, this time false negative:
```python
from typing import Union, overload

@overload
def f(x: int, y: int) -> int: ...
@overload
def f(x: object, y: str) -> str: ...
def f(x):
    pass

x: Union[int, str]
y: Union[int, str]
f(x, y) # passes, but should fail because it never matches (str, int)
```
The problem is that `make_simplified_union` is used. On one hand this allows us to do unioning in more scenarios, on other hand it is dangerous for union counting in unioned args. I am not sure what to do. @Michael0x2a could you take care of this? (This is however not urgent, since this is a false negative.)

Michael0x2a:
Will do -- I'll take a look a little later today.

ilevkivskyi:
Sorry, for overtaking, but I went ahead and now have a possible fix for this in a form of a change to a union math logic, so I am un-assigning you here, but I hope you will review my PR that I will open soon :-)

PR #5255, opened 2018-06-20
Alternative algorithm for union math

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5243
Fixes https://github.com/python/mypy/issues/5249

Some comments:
* I went ahead with a slow but very simple recursive algorithm that treats all various complex cases correctly. On one hand it can be quadratic, but on the other hand, the complexity will be bad _only_ if user abuses lots of unions
* This is WIP because I use a hack caused by the fact that currently most function inference functions pass argument _expressions_ instead of types
* It may look like there are many changes in tests, but actually there are not, the differences are because:
  - Error messages now show the _first potentially matching_ overload (which is kind of OK I think)
  - Order of items in many unions turned to the opposite, apparently union `__repr__` is unstable.


mypy/checkexpr.py
```diff
@@ -1130,22 +1135,15 @@ def check_overload_call(self,
...
-                union_success = unioned_result is not None and not unioned_errors.is_errors()
+        if any(self.real_union(arg) for arg in arg_types):
+            unioned_errors = arg_messages.clean_copy()
+            unioned_result = self.union_overload_result(plausible_targets, args, arg_types,
+                                                        arg_kinds, arg_names,
+                                                        callable_name, object_type,
+                                                        context, arg_messages=unioned_errors)
+            # Record if we succeeded. Next we need to see if maybe normal procedure
+            # gives a narrower type.
+            union_success = unioned_result is not None and not unioned_errors.is_errors()
```
Michael0x2a: Given the way you implemented `union_overload_result`, I think it's impossible for `union_errors` to actually ever contain an error. As a consequence, `union_success` will be true if and only if `union_result is not None` is also true.

In that case, I think we can simplify some of the logic by getting rid of `union_success` and using just `union_result`. E.g. here, we replace this line with `if union_errors.is_errors(): unioned_result = None` just in case and get rid of `union_success` entirely.

mypy/checkexpr.py
```diff
@@ -1130,22 +1135,15 @@ def check_overload_call(self,
...
 
         # Step 3: We try checking each branch one-by-one.
         inferred_result = self.infer_overload_return_type(plausible_targets, args, arg_types,
```
Michael0x2a: Github won't let me leave a comment down below after doing the simplification suggested up above, I think we can also simplify the following if statements to something roughly like:

    # Success! Stop early by returning the best among normal and unioned.
    if inferred_result is not None and unioned_result is not None:
        return inferred_result if is_subtype(inferred_result[0], unioned_result[0]) else unioned_result
    elif inferred_result is not None:
        return inferred_result
    elif unioned_result is not None:
        return unioned_result

mypy/checkexpr.py
```diff
@@ -1173,9 +1171,8 @@ def check_overload_call(self,
...
         #         Neither alternative matches, but we can guess the user probably wants the
         #         second one.
-        if erased_targets is None:
-            erased_targets = self.overload_erased_call_targets(plausible_targets, arg_types,
-                                                               arg_kinds, arg_names, context)
+        erased_targets = self.overload_erased_call_targets(plausible_targets, arg_types,
+                                                           arg_kinds, arg_names, context)
 
         # Step 5: We try and infer a second-best alternative if possible. If not, fall back
         #         to using 'Any'.
```
Michael0x2a: Github still won't let me leave a comment below, but I think after doing the above two simplifications, we can get rid of the next if statement entirely.

(It's a bit hard to tell, but based on new error messages in the changed tests, it seems like we're no longer entering into this case at all, so it's redundant either way.)

mypy/checkexpr.py
```diff
@@ -519,7 +520,9 @@ def check_call(self, callee: Type, args: List[Expression],
                    callable_node: Optional[Expression] = None,
                    arg_messages: Optional[MessageBuilder] = None,
                    callable_name: Optional[str] = None,
-                   object_type: Optional[Type] = None) -> Tuple[Type, Type]:
+                   object_type: Optional[Type] = None,
+                   *,
+                   arg_types_override: Optional[List[Type]] = None) -> Tuple[Type, Type]:
```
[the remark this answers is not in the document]
ilevkivskyi: Yes, this is an artefact of a previous attempt to do this "right way".

mypy/checkexpr.py
```diff
@@ -1130,22 +1135,15 @@ def check_overload_call(self,
...
-                                                 object_type=object_type)
-                # Record if we succeeded. Next we need to see if maybe normal procedure
-                # gives a narrower type.
-                union_success = unioned_result is not None and not unioned_errors.is_errors()
+        if any(self.real_union(arg) for arg in arg_types):
+            unioned_errors = arg_messages.clean_copy()
+            unioned_result = self.union_overload_result(plausible_targets, args, arg_types,
+                                                        arg_kinds, arg_names,
+                                                        callable_name, object_type,
+                                                        context, arg_messages=unioned_errors)
```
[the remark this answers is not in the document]
ilevkivskyi: I thought `erased_targets` are calculated using original types (i.e. including full unions). I just wanted to throw away _once_ all the overloads that have no way to match, while keep those that can match after a union destructuring. I think `plausible_targets` is what is needed or am I wrong?

mypy/checkexpr.py
```diff
@@ -1350,91 +1347,56 @@ def overload_erased_call_targets(self,
...
-        Returns None if it is not possible to combine the different callables together in a
-        sound manner.
-
-        Assumes all of the given callables have argument counts compatible with the caller.
+    def union_overload_result(self,
+                              plausible_targets: List[CallableType],
+                              args: List[Expression],
+                              arg_types: List[Type],
+                              arg_kinds: List[int],
+                              arg_names: Optional[Sequence[Optional[str]]],
```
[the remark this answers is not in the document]
ilevkivskyi: Yes, we have many potential refactoring ideas on our internal roadmap, we can add this one to the list.

mypy/checkexpr.py
```diff
@@ -1350,91 +1347,56 @@ def overload_erased_call_targets(self,
...
-        # then we'd incorrectly accept calls like "f(A(), A())" when they really ought to
-        # be rejected.
-        #
-        # However, that means we'll also give up if the original overloads contained
-        # any unions. This is likely unnecessary -- we only really need to give up if
-        # there are more then one *synthesized* union arguments.
-        if union_count >= 2:
-            return None
+        if returns:
+            print('un', returns, UnionType.make_simplified_union(returns, context.line, context.column))
```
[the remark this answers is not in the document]
ilevkivskyi: Oh yes, sorry.

mypy/checkexpr.py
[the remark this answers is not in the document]
Michael0x2a: I think `erased_targets` is computed in a way that it matches a callable if at least one of the union elements, not if all of them match. That allowed it to act as an additional filter that pruned `plausible_targets` down even further.

That said, I think the way we compute the erased targets is currently a bit janky -- pretty much all of that code is from the old overloads implementation and is something I'm planning on refactoring/simplifying in a future PR.  So, I wouldn't be too surprised if it didn't end up working here. (But if it did, it might be a way we could get a minor speedup in the case where the list of overloads is long.)

That said, I guess that's a bit of an edge case, so maybe just using `plausible_targets` is the right thing to do here, since it's more obviously correct.

mypy/checkexpr.py
```diff
@@ -1350,91 +1347,56 @@ def overload_erased_call_targets(self,
...
+                              plausible_targets: List[CallableType],
+                              args: List[Expression],
+                              arg_types: List[Type],
+                              arg_kinds: List[int],
+                              arg_names: Optional[Sequence[Optional[str]]],
+                              callable_name: Optional[str],
+                              object_type: Optional[Type],
+                              context: Context,
+                              arg_messages: Optional[MessageBuilder] = None,
+                              ) -> Optional[Tuple[Type, Type]]:
```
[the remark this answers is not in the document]
ilevkivskyi: Yes, you are right, it can be much worse. Here is some more detailed analysis.

The relevant numbers that we have here are:
* number of (potentially matching) overloads, `p`
* number of unions in args, `k`
* numbers of items in every  union `n_i, i = 1, ..., k`

The worst case scenario is when every check succeeds, in this case we have `p * n_1 * ... * n_k` inference attempts, assuming an average `n` items per union this is roughly `p * n ** k`. This might look very bad, but I have some intuition why in practice average/typical performance may be much better. The key is that in practice it is hard to write an overload that will match so many combinations, plus there is a `return None` that will fully "unwind" the union call stack on first failure. So for any realistic overload the number of union item combinations that _successfully_ match will be limited by number of overloads `p`, for example to match successfully two args with types `Union[A, B]` and `Union[C, D]` one will need four overloads. Thus we will _typically_ make total at most `p ** 2` inference attempts.

Actually, now thinking more about this, we can also bailout soon on success, i.e. after splitting the first union, before we split the next one, we can check, maybe it already matches, for example:
```python
@overload
def f(x: int, y: object, z: object) -> int: ...
@overload
def f(x: str, y: object, z: object) -> str: ...

x: Union[int, str]
f(x, x, x)
```
There is not need to split the second and third overload. This will need just few extra lines, but can be a decent performance gain.

An additional comment in the defence of this algorithm is its simplicity of implementation (especially as compared to the breadth of covered use cases). Anyway, if you still think it is dangerous it is easy to put a limit on the call stack depth.

mypy/checkexpr.py
```diff
@@ -1350,91 +1347,56 @@ def overload_erased_call_targets(self,
...
-        #
-        # However, that means we'll also give up if the original overloads contained
-        # any unions. This is likely unnecessary -- we only really need to give up if
-        # there are more then one *synthesized* union arguments.
-        if union_count >= 2:
-            return None
+        if returns:
+            print('un', returns, UnionType.make_simplified_union(returns, context.line, context.column))
+            return (UnionType.make_simplified_union(returns, context.line, context.column),
+                    UnionType.make_simplified_union(inferred_types, context.line, context.column))
```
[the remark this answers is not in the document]
ilevkivskyi: I am not sure we can use sets (although I really want to), the problem is that we need to have a stable/predictable order of union items to avoid flaky tests.

test-data/unit/check-overloading.test
```diff
@@ -2439,10 +2440,8 @@ def f(x: A) -> Child: ...
 def f(x: B, y: B = B()) -> Parent: ...
 def f(*args): ...
 
-# TODO: It would be nice if we could successfully do union math
-# in this case. See comments in checkexpr.union_overload_matches.
 x: Union[A, B]
-f(x)       # E: Argument 1 to "f" has incompatible type "Union[A, B]"; expected "A"
```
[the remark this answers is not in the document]
ilevkivskyi: Yes, sure, good idea!

test-data/unit/check-overloading.test
```diff
@@ -2519,10 +2518,10 @@ def wrapper() -> None:
 
     obj2: Union[W1[A], W2[B]]
 
-    foo(obj2)  # E: Cannot infer type argument 1 of "foo"
+    foo(obj2)  # OK
     bar(obj2)  # E: Cannot infer type argument 1 of "bar"
```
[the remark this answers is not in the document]
ilevkivskyi: Actually what do you think about the new error messages? They are a bit shorter and match a bit more the logic of "match the first overload", but there is less "symmetry" between overloads and normal functions that have unions in signature.

test-data/unit/check-overloading.test
```diff
@@ -3717,3 +3715,39 @@ def relpath(path: str) -> str: ...
...
+    @overload  # type: ignore
+    def f(self, other: D) -> D: ...
+    @overload
+    def f(self, other: C) -> C: ...
+    def f(self, other): ...
+
+x: D
+y: Union[D, Any]
+# TODO: update after we decide on https://github.com/python/mypy/pull/5254
+reveal_type(x.f(y))  # E: Revealed type is 'Any'
```
ilevkivskyi: Just a comment to myself, we need to fix this before merging this PR, Jukka prefers `Union[D, Any]` here, which actually the result a unioned match, but direct match is `Any`, so:
* #5254 should go in first
* after that we might tweak the choice between direct and unioned match

ilevkivskyi:
@Michael0x2a thanks for a detailed review! I will implement the changes tomorrow probably.

ilevkivskyi:
@Michael0x2a Btw what do you think about the `type_overrides` hack? I am a bit worried actually about its performance implications, it is probably small 1-2%, but it applies to _every_ expression visit. We have something very similar for multiassign from union, but somehow it works only with temporarily modifying `type_map` (btw I should probably also make `type_overrides` a context manager, like in case of multiassign from union).

test-data/unit/check-overloading.test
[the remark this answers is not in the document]
Michael0x2a: I think I'm ok with the new error messages. It does lose symmetry, but idk if the user was expecting that symmetry to begin with. The new messages also require less explanation: if we report back that a union is expected when the overload definitions doesn't contain any unions at all, I could see that confusing a user who wasn't aware of the union math feature.

Michael0x2a:
Regarding the `type_overrides` hack: I agree that it's suboptimal and that something using `arg_types_override` approach would be nice. If there's no other way to make this approach work though, I guess I wouldn't be too upset -- it's at least easy to understand.

mypy/checkexpr.py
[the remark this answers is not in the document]
Michael0x2a: > The key is that in practice it is hard to write an overload that will match so many combinations

I'm not sure if it'll actually be that hard in practice, especially if the overload is intended to match a wide variety of different calls using a mixture of typevars, `Any` types, object types, `*args`, `**kwargs`...

For example, here's a program which calls a simplified version of `map`. It's a bit contrived, but I don't think it's too far off from the sorts of programs some users might write in practice:

```python
from typing import overload, Any, TypeVar, Iterable, List, Dict, Callable, Union

S = TypeVar('S')

@overload
def simple_map() -> None: ...
@overload
def simple_map(func: Callable[..., S], *iterables: Iterable[Any]) -> S: ...
def simple_map(*args): pass

def format_row(*entries: object) -> str: pass

class DateTime: pass
JsonBlob = Dict[str, Any]
Column = Union[List[str], List[int], List[bool], List[float], List[DateTime], List[JsonBlob]]

def print_custom_table() -> None:
    a: Column
    b: Column
    c: Column
    d: Column
    e: Column
    f: Column
    g: Column

    for row in simple_map(format_row, a, b, c, d, e, f):
        print(row)
```

If I try running `simple_map` with 6 columns (a through f), it takes the new algorithm roughly a minute and a half on my computer to type check this program. If I tack on the 7th column (g), it takes mypy significantly longer. (I don't have an exact time unfortunately: I got bored of waiting after 10-15 minutes and killed the thing.)

I think type-checking after each split would probably only aggravate the problem, unfortunately (though I do think it could help in other cases).

> An additional comment in the defence of this algorithm is its simplicity of implementation (especially as compared to the breadth of covered use cases). Anyway, if you still think it is dangerous it is easy to put a limit on the call stack depth.

I also like the simplicity of this algorithm. (It's also not obvious to me how we could robustly solve this problem in any other way, in any case.)

But yeah, I still think we should add a limit of some sort. Even if the pathological cases end up being rare, I think it'd be a poor user experience if mypy starts unexpectedly stalling in those edge cases.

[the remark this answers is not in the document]
ilevkivskyi: > I think type-checking after each split would probably only aggravate the problem, unfortunately (though I do think it could help in other cases).

It actually helped. I added it (simplistic WIP style, need to polish it) and your example passed in a second (even with eight(!) union args) also few other complex examples I tried passed instantly.

> But yeah, I still think we should add a limit of some sort.

I don't want to set a calculated limit, I would add a nesting/split level limit probably 5, and then raise `TooManyUnionsError`, so that we can better diagnose an error and add a note like:
```
main.py: error: No overload variant blah-blah-blah..
main.py: note: Not all union combinations were tried because there are too many unions
```

There are some ideas to avoid hitting the limit, like add some back-tracking add multi-seed (like start splitting the unions from last arg as well). But I will not do this, because this will only add code complexity while gaining some super-rare corner cases.

mypy/checkexpr.py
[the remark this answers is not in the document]
ilevkivskyi: OK, I think I figured out how to avoid multiple intermediate unions, and still have stable order, we can return a set of pairs (indexes of matched overload, return type), at the very end we construct a union once but before sort items by index.

Another important note: I think returning a union of inferred callable types may be wrong here, we should actually return a single unioned callable (or fall back to `Callable[..., Union[<ret_types>]]` if the matched variants are too different to combine). The motivation is that `(Union[int -> None, str -> None])(Union[int, str])` is not a valid call, maybe some code (even future code) relies on the fact that inferred callable type and initial arg types are consistent.

mypy/checkexpr.py
```diff
@@ -1350,91 +1347,62 @@ def overload_erased_call_targets(self,
...
-        for target in callables:
-            # We conservatively end if the overloads do not have the exact same signature.
-            # The only exception is if one arg is optional and the other is positional: in that
-            # case, we continue unioning (and expect a positional arg).
-            # TODO: Enhance the union overload logic to handle a wider variety of signatures.
-            if len(new_kinds) != len(target.arg_kinds):
+        if not any(self.real_union(typ) for typ in arg_types):
+            # No unions in args, just fall back to normal inference
+            for arg, typ in zip(args, arg_types):
+                self.type_overrides[arg] = typ
```
ilevkivskyi: This should be made a context manager, also we need to be sure that the last type for an arg expression (the one that stays in type map after we leave the call expression) is a union.

ilevkivskyi:
@Michael0x2a This should be ready for review now.

mypy/checkexpr.py
```diff
@@ -1358,18 +1374,112 @@ def overload_erased_call_targets(self,
...
+        Return a list of (<return type>, <inferred variant type>) if call succeeds for every
+        item of the desctructured union. Returns None if there is no match.
+        """
+        # Step 1: If we are already too deep, then stop immediately. Otherwise mypy might
+        # hang for long time because of a weird overload call. The caller will get
+        # the exception and generate an appropriate note message, if needed.
+        if level >= MAX_UNIONS:
+            raise TooManyUnions
+
+        # Step 2: Find position of the first union in arguments. Return the normal infered
```
Michael0x2a: Really minor nit: "infered" -> "inferred"

mypy/checkexpr.py
```diff
@@ -1358,18 +1374,112 @@ def overload_erased_call_targets(self,
...
+
+    def union_overload_matches(self, types: Sequence[Type]) -> Union[AnyType, CallableType]:
         """Accepts a list of overload signatures and attempts to combine them together into a
         new CallableType consisting of the union of all of the given arguments and return types.
 
-        Returns None if it is not possible to combine the different callables together in a
-        sound manner.
-
-        Assumes all of the given callables have argument counts compatible with the caller.
+        If there is at least one non-callabe type, return Any (this can happen if there is
```
Michael0x2a: "callabe" -> "callable"

mypy/checkexpr.py
```diff
@@ -1138,41 +1157,42 @@ def check_overload_call(self,
...
+            else:
                 # Record if we succeeded. Next we need to see if maybe normal procedure
                 # gives a narrower type.
-                union_success = unioned_result is not None and not unioned_errors.is_errors()
+                if unioned_return:
+                    returns, inferred_types = zip(*unioned_return)
+                    unioned_result = (UnionType.make_simplified_union(list(returns),
+                                                                      context.line,
+                                                                      context.column),
+                                      self.union_overload_matches(inferred_types))
```
Michael0x2a: Out of curiosity, is there any reason why we switched from using `UnionType.make_simplified_union` to `self.union_overload_matches` for the inferred types?

(fwiw, I don't mind whether we keep it or not -- if we remove it, it simplifies the code; if we keep it, it might help us perform better subtype checks with overloads in the future.)

mypy/checkexpr.py
```diff
@@ -58,6 +62,16 @@
 ArgChecker = Callable[[Type, Type, int, Type, int, int, CallableType, Context, MessageBuilder],
                       None]
 
+# Maximum nesting level for math union in overloads, setting this to large values
+# may cause performance issues.
```
Michael0x2a: Maybe we should include add a blurb + a link telling people to read this PR if they want more details? Up to you.

mypy/checkexpr.py
```diff
@@ -1358,18 +1374,112 @@ def overload_erased_call_targets(self,
...
+        res_items = []
+        for item in first_union.relevant_items():
+            new_arg_types = arg_types.copy()
+            new_arg_types[idx] = item
+            sub_result = self.union_overload_result(plausible_targets, args, new_arg_types,
+                                                    arg_kinds, arg_names, callable_name,
+                                                    object_type, context, arg_messages,
+                                                    level + 1)
+            if sub_result is not None:
+                res_items.append(sub_result)
```
Michael0x2a: Another minor nit: I think it'd look cleaner if we do `res_items.extend(sub_result)` here. We can then simplify the nested loop down below to just a single one. (And we'd need to change the comment to something like "Step 5: If splitting succeeded, filter out duplicate items".)

mypy/checkexpr.py
```diff
@@ -1358,18 +1374,112 @@ def overload_erased_call_targets(self,
...
+            else:
+                # Some item doesn't match, return soon.
+                return None
+
+        # Step 5: If spliting succeeded, then flatten union results into a single
```
Michael0x2a: "spliting" -> "splitting"

mypy/checkexpr.py
```diff
@@ -1389,58 +1499,52 @@ def union_overload_matches(self, callables: List[CallableType]) -> Optional[Call
...
+        too_complex = False
         for target in callables:
-            # We conservatively end if the overloads do not have the exact same signature.
-            # The only exception is if one arg is optional and the other is positional: in that
-            # case, we continue unioning (and expect a positional arg).
-            # TODO: Enhance the union overload logic to handle a wider variety of signatures.
+            # We fall back to Callable[..., Union[<returns>]] if the overloads do not have
+            # the exact same signature. The only exception is if one arg is optional and
+            # the other is positional: in that case, we continue unioning (and expect a
+            # positional arg).
```
Michael0x2a: What about falling back to `Union[callable1, callable2, ..., callableN]` instead? It would let us introduce fewer new ``Any``s. Or does that end up breaking too many things?

(I think this is basically the same question as the previous one -- maybe add a comment here about this?)

mypy/checkexpr.py
ilevkivskyi: The reason is that if I would write
```python
f: Union[Callable[[int], int], Callable[[str], str]]
x: Union[int, str]
```
then mypy would (correctly flagged the call) as an error. Therefore I don't want mypy to infer callable type that is formally inconsistent with argument type(s). I will add a comment about this.

mypy/checkexpr.py
ilevkivskyi: > Or does that end up breaking too many things?

It actually breaks nothing, I just don't want to introduce some internal inconsistencies (as I explained above). Also the `Any`s here don't impact code precision currently, we mostly care about the return type, not inferred callable type.

PR #5256, opened 2018-06-21
Refactor and reorder search path

ethanhs:
As promised in #5227, here is an implementation for refactoring and making the search path compliant with PEP 561. The order is specified in https://www.python.org/dev/peps/pep-0561/#type-checker-module-resolution-order.

This is also my first PR from an airplane :)

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -264,7 +264,7 @@ def compute_search_paths(sources: List[BuildSource],
         if dir not in lib_path:
             python_path.insert(0, dir)
 
-    # Start with a MYPYPATH environment variable to front of library path, if defined.
+    # Start with a MYPYPATH environment variable at the front of the mypy_path, if defined.
     mypypath = mypy_path()
 
     # Add a config-defined mypy path.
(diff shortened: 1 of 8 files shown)


mypy/build.py
```diff
@@ -188,10 +188,43 @@ def default_flush_errors(new_messages: List[str], is_serious: bool) -> None:
         raise
 
 
-def compute_lib_path(sources: List[BuildSource],
+SearchPaths = NamedTuple('SearchPaths',
```
ilevkivskyi: Could you please add a comment above documenting the fields and a typical usage.

mypy/build.py
```diff
@@ -227,20 +259,23 @@ def compute_lib_path(sources: List[BuildSource],
         else:
             dir = os.getcwd()
         if dir not in lib_path:
-            lib_path.appendleft(dir)
+            python_path.insert(0, dir)
 
-    # Prepend a config-defined mypy path.
-    lib_path.extendleft(options.mypy_path)
+    # Start with a MYPYPATH environment variable to front of library path, if defined.
```
ilevkivskyi: This sentence reads a bit weird. Did you mean "prepend"?

ilevkivskyi:
(Hm, also there is a merge conflict now.)

ethanhs:
@ilevkivskyi this should be ready to try on another code base.

I think I will tackle (finally) setuptools support next. It should be faily straightforward.

ethanhs:
Before merging this I also should probably update https://mypy.readthedocs.io/en/latest/command_line.html#how-imports-are-found (which I will do shortly).

PR #5257, opened 2018-06-21
Ignore vscode settings folder in git

ethanhs:
I've started using VS Code for editing things, so it'd be handy to ignore this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ dmypy.json
 # IDEs
 .idea
 *.swp
+.vscode
 
 # Operating Systems
 .DS_Store


gvanrossum:
Hm, `.idea` set a bad example. The right solution here is usually to create a *global* git ignore file and add it there. A projet-specific `.gitignore` should only be used to ignore files created by tools used by the project (e.g. `.pyc`, `/out`).

JukkaL:
I'm not sure if we can win this battle, as many people seem to be expect that editor specific files should be added to `.gitignore`. Supporting a few generated file patterns per each of the top 5 most popular editors or so might be a reasonable compromise. Or else we'd need to add detailed instructions for how to locally ignore editor-specific files in our `.gitignore`.

Issue #4543, opened 2018-02-05
Crash in union with a forward reference to a bad type.


ilevkivskyi:
This code causes a crash:
```python
Alias = Union[Bad, C, D]

Bad = object()

class C:
    pass
class D:
    pass

option: Alias
if isinstance(option, C):
    pass
```
Here is a traceback:
```
tmp.py:3: error: Invalid type "tmp.Bad"
tmp.py:13: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.570-dev-d2c04190b8d407476c535f30d07a9f6712e7b12a-dirty
Traceback (most recent call last):
  File "/Users/ilevkivskyi/mypy/bin/mypy", line 11, in <module>
    load_entry_point('mypy===0.570-dev-d2c04190b8d407476c535f30d07a9f6712e7b12a-dirty', 'console_scripts', 'mypy')()
  File "/Users/ilevkivskyi/mypy/lib/python3.6/site-packages/mypy/__main__.py", line 7, in console_entry
    main(None)
  File "/Users/ilevkivskyi/mypy/lib/python3.6/site-packages/mypy/main.py", line 80, in main
    type_check_only(sources, bin_dir, options, flush_errors)
  File "/Users/ilevkivskyi/mypy/lib/python3.6/site-packages/mypy/main.py", line 129, in type_check_only
    flush_errors=flush_errors)
  File "/Users/ilevkivskyi/mypy/lib/python3.6/site-packages/mypy/build.py", line 172, in build
    result = _build(sources, options, alt_lib_path, bin_dir, saved_cache, flush_errors)
  File "/Users/ilevkivskyi/mypy/lib/python3.6/site-packages/mypy/build.py", line 258, in _build
    graph = dispatch(sources, manager)
  File "/Users/ilevkivskyi/mypy/lib/python3.6/site-packages/mypy/build.py", line 2048, in dispatch
    process_graph(graph, manager)
  File "/Users/ilevkivskyi/mypy/lib/python3.6/site-packages/mypy/build.py", line 2349, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/Users/ilevkivskyi/mypy/lib/python3.6/site-packages/mypy/build.py", line 2525, in process_stale_scc
    graph[id].type_check_first_pass()
  File "/Users/ilevkivskyi/mypy/lib/python3.6/site-packages/mypy/build.py", line 1923, in type_check_first_pass
    self.type_checker().check_first_pass()
[...]

Issue #4987, opened 2018-05-01
Do not leak UnboundTypes to type checking


ilevkivskyi:
Sometimes `UnboundType` leaks to type checking, this is not good because it can lead to crashes like this:
```
File ".../mypy/subtypes.py", line 814, in <listcomp>
if (not (is_proper_subtype(erase_type(item), erased_s) or
File ".../mypy/erasetype.py", line 25, in erase_type
return typ.accept(EraseTypeVisitor())
File ".../mypy/types.py", line 194, in accept
return visitor.visit_unbound_type(self)
File ".../mypy/erasetype.py", line 31, in visit_unbound_type
assert False, 'Not supported'
AssertionError: Not supported
```
It would be better to replace all `UnboundType`s with `Any` (after an error is reported).

ilevkivskyi:
Update: this doesn't crash anymore, but we still to avoid leaking `UnboundType`s.

PR #5258, opened 2018-06-21
Remove an assert triggered by UnboundType leaking from semanal.py

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/4543

This is also a temporary solution for https://github.com/python/mypy/issues/4987. A permanent one would be to turn `UnboundType`s into `Any`s directly in `semanal.py`.


JukkaL:
Fixes #4543.

This is also a temporary solution for #4987. A permanent one would be to 
turn UnboundTypes into Anys directly in semanal.py.

ilevkivskyi:
I think you wanted to merge it, but probably hit the wrong button :-)

PR #5266, opened 2018-06-23
Use a compiled regex when parsing string format

ceh:
Sample timeit shows a 1.4x speedup on my machine.

Before:
```
(venv) ~/src/python/mypy (master) % python3 -m timeit -r 10 -p -s 'from mypy.checkstrformat import StringFormatterChecker; sfc = StringFormatterChecker(None,None,None)' 'sfc.parse_conversion_specifiers("Deferring module to fine-grained update %s (%s)")'
100000 loops, best of 10: 9.32 usec per loop
```

After:
```
(venv) ~/src/python/mypy (str-format-compiled-re) % python3 -m timeit -r 10 -p -s 'from mypy.checkstrformat import StringFormatterChecker; sfc = StringFormatterChecker(None,None,None)' 'sfc.parse_conversion_specifiers("Deferring module to fine-grained update %s (%s)")'
100000 loops, best of 10: 6.65 usec per loop
```

### DIFF
--- a/mypy/checkstrformat.py
+++ b/mypy/checkstrformat.py
@@ -2,7 +2,7 @@
 
 import re
 
-from typing import cast, List, Tuple, Dict, Callable, Union, Optional
+from typing import cast, List, Tuple, Dict, Callable, Union, Optional, Pattern
 
 from mypy.types import (
     Type, AnyType, TupleType, Instance, UnionType, TypeOfAny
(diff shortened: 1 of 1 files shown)


JukkaL:
Can you demonstrate a measurable improvement for any real-world example, such as type checking mypy itself?

gvanrossum:
> Can you demonstrate a measurable improvement for any real-world example

I initially had a similar response, but then realized that it really isn't
very idiomatic to construct such a complex regex out of its parts each time
it is used -- performance or not, it just looks better.

JukkaL:
@gvanrossum Yeah, you are right.

PR #5270, opened 2018-06-24
Move linting to pytest

elazarg:
Use flake8's [legacy api](http://flake8.pycqa.org/en/latest/user/python-api.html) since there's no non-legacy one.

Another item for #1673 (only `stdlibsamples` left).

### DIFF
--- a/mypy/test/testselfcheck.py
+++ b/mypy/test/testselfcheck.py
@@ -1,5 +1,8 @@
 """Self check mypy package"""
 
+import pytest  # type: ignore  # no pytest in typeshed
+from flake8.api import legacy as flake8  # type: ignore  # no falke8 in typeshed
+
 from mypy.test.helpers import Suite, run_mypy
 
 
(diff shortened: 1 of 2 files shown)


gvanrossum:
I'm not sure I like this. Maybe running flake8 should just be a separate step in .travis.yml? That's how it is done currently -- even though runtests.py supports it, .travis.yml excludes lint (and self-check) and runs flake8 separately. I want pytest to run *tests* not other stuff like linters and type-checkers. (FWIW it appears self-check is no long run by .travis.yml -- something has made `python runtests.py package` a no-op.)

ilevkivskyi:
> Maybe running flake8 should just be a separate step in .travis.yml?

I agree with this. I would just remove flake8 from runtests.py

> FWIW it appears self-check is no long run by .travis.yml

One can now run `pytest -k mypy_package`, I am not sure I like it, I would also prefer it to be a separate Travis (and Appveyor) task actually.

elazarg:
Sounds good to me. So this PR is not required at all - after stdlibsamples will be migrated, runtests.py  and waiter.py can simple be removed. Is this correct?

ilevkivskyi:
> after stdlibsamples will be migrated, runtests.py and waiter.py can simple be removed. Is this correct?

I think yes.

elazarg:
Closing then.

gvanrossum:
> One can now run `pytest -k mypy_package`, I am not sure I like it, I would also prefer it to be a separate Travis (and Appveyor) task actually.

Yes, please.

PR #5271, opened 2018-06-25
Migrate stdlibsamples test to pytest

elazarg:
Check as files instead of as modules, without changing cwd.

As far as I understand, this should be the last step of #1673 before removal of `runtests.py` and `waiter.py`.

(completes #5142)

### DIFF
--- a/mypy/test/testsamples.py
+++ b/mypy/test/testsamples.py
@@ -34,6 +34,18 @@ class SamplesSuite(Suite):
                 mypy_args.append('--python-version=3.5')
             run_mypy(mypy_args + [f])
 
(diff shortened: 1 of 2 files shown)


gvanrossum:
I'd like @JukkaL to have a quick look -- he presumably remembers why we have this test in the first place.

JukkaL:
These are regression tests we've had from the early days of mypy. They aren't very useful any more, but I think it's worth keeping them as they provide a sense of historical perspective, at the very least.

gvanrossum:
Thanks Elazar for powering through these issues!

PR #5282, opened 2018-06-26
Support egg/setuptools packages for PEP 561 searching.

### LINKED ISSUE #5007: Support egg installs with PEP 561 searching

ethanhs:
This adds support for setuptool's egg format, which includes support for editable installs (`pip install -e .`/`python setup.py develop`)!

Setuptools creates its own directory to put the package we want to find. These directories are listed in `easy-install.pth` in a site-package directory.

Fixes #5007


mypy/test/testpep561.py
```diff
@@ -67,10 +67,18 @@ def virtualenv(self,
...
         """Context manager to temporarily install a package from test-data/packages/pkg/"""
         working_dir = os.path.join(package_path, pkg)
-        install_cmd = [python_executable, '-m', 'pip', 'install', '.']
+        if use_pip:
+            install_cmd = [python_executable, '-m', 'pip', 'install']
+            if editable:
+                install_cmd.append('-e')
+            install_cmd.append('.')
+        else:
+            install_cmd = [python_executable, 'setup.py', 'install']
```
JelleZijlstra: Should this be `develop` if `editable` is True?

ethanhs: I suppose that would be better, but the test should just use pip and editable (since `pip install -e .` calls `python setup.py develop`). But just to be safe, I will change it.

mypy/build.py
```diff
@@ -199,26 +199,44 @@ def default_flush_errors(new_messages: List[str], is_serious: bool) -> None:
...
 
     This runs a subprocess call, which generates a list of the site package directories.
     To avoid repeatedly calling a subprocess (which can be slow!) we lru_cache the results."""
+    def make_abspath(path: str, root: str) -> str:
+        """Take a path and make it absolute relative to root if not already absolute."""
+        path = os.path.normpath(path)
+        if os.path.abspath(path) == path:
+            return path
+        else:
+            return os.path.join(root, path)
```
gvanrossum: This feels a bit complicated (perhaps in part because you're assigning to a parameter). Is the intention this?
```
if os.path.isabs(path):
    return os.path.normpath(path)
else:
    return os.path.join(root, os.path.normpath(path))
```
While that looks like more code the flow feels a bit clearer. Also it's possible that in one of the branches you don't actually care about normpath, but I can't tell. (Or possibly you might want to call it after calling join.) Finally I note that presumably root is assumed to be absolute itself, else the function name would be misleading.

ethanhs: Yes, when I wrote this I didn't know isabs existed. :) But that is much better than what I was doing. I will use your suggested code.

PR #5284, opened 2018-06-27
Update stubutil.py is_c_module with .pyd check

AlexArcPy:
Added check for .pyd file extension used on Windows

### DIFF
--- a/mypy/stubutil.py
+++ b/mypy/stubutil.py
@@ -87,7 +87,8 @@ def find_unique_signatures(sigs: Sequence[Sig]) -> List[Sig]:
 
 
 def is_c_module(module: ModuleType) -> bool:
-    return '__file__' not in module.__dict__ or module.__dict__['__file__'].endswith('.so')
+    return '__file__' not in module.__dict__ or any(module.__dict__['__file__'].endswith(
+      ext) for ext in ['.so', '.pyd'])
 
 
 def write_header(file: IO[str], module_name: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


mypy/stubutil.py
```diff
@@ -87,7 +87,8 @@ def find_unique_signatures(sigs: Sequence[Sig]) -> List[Sig]:
 
 
 def is_c_module(module: ModuleType) -> bool:
-    return '__file__' not in module.__dict__ or module.__dict__['__file__'].endswith('.so')
+    return '__file__' not in module.__dict__ or any(module.__dict__['__file__'].endswith(
+      ext) for ext in ['.so', '.pyd'])
```
gvanrossum: This is causing the linter to fail:
```
./mypy/stubutil.py:91:7: E121 continuation line under-indented for hanging indent
```
You can run the linter locally: `flake8 mypy/stubutil.py`

Though honestly I would refactor the code a little more -- the `any()` feels like it makes the code more complex than it needs to be. Maybe use `os.path.splitext()`?

AlexArcPy:
Thanks, Guido! I did not dare to refactor, but thanks for the hint, it is done, the linter should be happy now. Please review the latest changes and let me know if there is anything more that should be fixed.

gvanrossum:
Thanks Alex! That was actually exactly the refactoring I was thinking of. :-)

mikiboii:
i dont understand how didi u generate pyi file for pyd python file

PR #5295, opened 2018-06-28
Allow __getattr__ during semantic analysis

ilevkivskyi:
Previously module `__getattr__` was used during type checking (runtime context). But it was only partially allowed in semantic analysis (after `from a import A`, `A` can be used in annotations/base classes). This PR allows the same support for `a.A`, i.e. it can be used in annotations and base classes (provided `__getattr__` has a suitable return type `Any`).

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3101,6 +3101,10 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
 
     def create_getattr_var(self, getattr_defn: SymbolTableNode,
                            name: str, fullname: str) -> Optional[SymbolTableNode]:
+        """Create a dummy global symbol using __getattr__ return type.
+
+        If not possible, return None.
+        """
         if isinstance(getattr_defn.node, (FuncDef, Var)):
             if isinstance(getattr_defn.node.type, CallableType):
                 typ = getattr_defn.node.type.ret_type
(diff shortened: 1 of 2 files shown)


mypy/semanal.py
```diff
@@ -3083,6 +3083,18 @@ def lookup_qualified(self, name: str, ctx: Context,
...
+                            if isinstance(getattr_defn.node, (FuncDef, Var)):
+                                if isinstance(getattr_defn.node.type, CallableType):
+                                    typ = getattr_defn.node.type.ret_type
+                                else:
+                                    typ = AnyType(TypeOfAny.from_error)
+                                name = parts[i]
+                                v = Var(name, type=typ)
+                                v._fullname = name
+                                n = SymbolTableNode(GDEF, v)
+                                names[name] = n
```
msullivan: It might be worth pulling this out into another method, since it is a big chunk of code for an obscure case.

ilevkivskyi:
> Are these bugs run into working on sqlalchemy typing?

Yes. I am slowly moving through existing stubs and trying to combine them with some `__getattr__`s to make a "partial" package.

mypy/semanal.py
ilevkivskyi: Makes sense. I there is also similar code above in `visit_import_from`, it would make sense to combine the two into a helper method. I will do this tomorrow.

ilevkivskyi:
`testpr` is also clean so I am merging now.

PR #5296, opened 2018-06-29
List possible overload variants when none match

Michael0x2a:
Resolves https://github.com/python/mypy/issues/672 and supersedes https://github.com/python/mypy/pull/5054.

Currently, if a user calls an overloaded function and mypy is unable to find a match, mypy will display display a warning like 'No overload variant of "func" matches argument types" with no additional context.

This pull request will list several matching variants in addition to that message. If possible, mypy will attempt to show only overloads that have a compatible number of arguments. The number of overloads shown is always capped at a maximum of 2.

This pull request does *not* attempt to report a custom error when...

1. Union-math fails. Rationale: the pending PR will change how unions are handled dramatically enough to the point where any error handling here would be pointless.

2. Mypy is able to infer what signature the user most likely meant by looking at the erased types. Rationale: I attempted to implement this feature but was unable to make it consistently work without adding several ugly hacks around how mypy records errors so decided to defer implementing this feature.

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -1245,7 +1245,7 @@ class MessageBuilder:
 
     def pretty_overload(self, targets: List[CallableType], func: Overloaded, context: Context,
                         offset: int, max_items: int) -> None:
-        if len(targets) == 0:
+        if not targets:
             targets = func.items()
         for item in targets[:max_items]:
             self.note('@overload', context, offset=2 * offset)
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -771,12 +771,7 @@ class MessageBuilder:
             self.fail('No overload variant{} matches argument types {}'
                       .format(name_str, arg_types_str), context)
 
-        self.note('Possible overload variant(s):', context)
-        self.pretty_overload(plausible_targets,
-                             overload,
-                             context,
-                             offset=2,
-                             max_items=2)
+        self.pretty_overload_matches(plausible_targets, overload, context, offset=2, max_items=2)
 
     def wrong_number_values_to_unpack(self, provided: int, expected: int,
                                       context: Context) -> None:
@@ -1217,13 +1212,13 @@ class MessageBuilder:
                         self.note(self.pretty_callable(exp), context, offset=2 * OFFSET)
                     else:
                         assert isinstance(exp, Overloaded)
-                        self.pretty_overload(exp.items(), exp, context, OFFSET, MAX_ITEMS)
+                        self.pretty_overload(exp, context, OFFSET, MAX_ITEMS)
                     self.note('Got:', context, offset=OFFSET)
                     if isinstance(got, CallableType):
                         self.note(self.pretty_callable(got), context, offset=2 * OFFSET)
                     else:
                         assert isinstance(got, Overloaded)
-                        self.pretty_overload(got.items(), got, context, OFFSET, MAX_ITEMS)
+                        self.pretty_overload(got, context, OFFSET, MAX_ITEMS)
             self.print_more(conflict_types, context, OFFSET, MAX_ITEMS)
 
         # Report flag conflicts (i.e. settable vs read-only etc.)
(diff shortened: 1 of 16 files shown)


ilevkivskyi:
Sorry, it looks like I will not be able to review today :-( Probably on Saturday evening.

Michael0x2a:
@ilevkivskyi -- no worries, take your time :)

If you also want to prioritize landing the other overload PRs first, I'm fine with that. This PR feels like the kind of thing that'll cause a bunch of merge conflicts, so maybe it makes sense to just defer reviewing this so I can fix the conflicts all in one go instead of distributing them through the other PRs.

Michael0x2a:
Up to you though, of course.

Michael0x2a:
@gvanrossum -- before I merge this in, did you want to take a quick look? I know you said that the error messages were too noisy in the previous incarnation of this PR. If you still feel that way, there are probably one or two things I could try doing to make the error message shorter.

mypy/messages.py
```diff
@@ -1233,13 +1243,17 @@ def report_protocol_problems(self, subtype: Union[Instance, TupleType, TypedDict
...
+            targets = func.items()
+        for item in targets[:max_items]:
             self.note('@overload', context, offset=2 * offset)
             self.note(self.pretty_callable(item), context, offset=2 * offset)
-        if len(tp.items()) > max_items:
-            self.note('<{} more overload(s) not shown>'.format(len(tp.items()) - max_items),
+        shown = min(max_items, len(targets))
+        max_available = len(func.items())
+        if shown < max_available:
+            self.note('<{} more overload(s) not shown>'.format(max_available - shown),
```
gvanrossum: Apart form the uglines around plurals, it seems there are two possible reasons why not all overloads are shown:

- because they don't match the number of arguments (or the keywords or whatever)
- because the sheer number exceeds `max_items`

I think it would be good to differentiate between these in the messages somehow.

mypy/messages.py
```diff
@@ -1233,13 +1243,17 @@ def report_protocol_problems(self, subtype: Union[Instance, TupleType, TypedDict
...
         self.print_more(conflict_flags, context, OFFSET, MAX_ITEMS)
 
-    def pretty_overload(self, tp: Overloaded, context: Context,
+    def pretty_overload(self, targets: List[CallableType], func: Overloaded, context: Context,
                         offset: int, max_items: int) -> None:
-        for item in tp.items()[:max_items]:
+        if not targets:
+            targets = func.items()
+        for item in targets[:max_items]:
             self.note('@overload', context, offset=2 * offset)
```
gvanrossum: The extra line with `@overload` just doubles the verbosity without adding any information. I think it's okay to leave it out.

Michael0x2a:
@gvanrossum and/or @ilevkivskyi -- how's this?

The new format should handle plurals correctly and will display a message that looks like this if there are more then 2 overloads that have matching types/arity/whatever:

    Possible overload variants:
        def foo(x: int) -> int
        def foo(x: str) -> str
        <2 more matching overloads not shown, out of 5 total overloads>

...or like this if we're suppressing only non-matching overloads:

    Possible overload variants:
        def foo(x: int) -> int
        def foo(x: str) -> str
        <1 more non-matching overload not shown>

...and of course, like this if we we were able to display all the variants:

    Possible overload variants:
        def foo(x: int) -> int
        def foo(x: str) -> str

(I guess since we cut the number of lines shown in half by removing `@overload`, we could also increase the max number of variants shown to 3 or 4 or something. I don't particularly care either way though.)

mypy/messages.py
```diff
@@ -1233,14 +1238,49 @@ def report_protocol_problems(self, subtype: Union[Instance, TupleType, TypedDict
...
-        if len(tp.items()) > max_items:
-            self.note('<{} more overload(s) not shown>'.format(len(tp.items()) - max_items),
-                      context, offset=2 * offset)
+
+        max_available = len(targets)
+        shown = min(max_items, max_available)
+        if shown < max_available:
+            left = max_available - shown
+            msg = '<{} more overload{} not shown>'.format(left, plural_s(left))
+            self.note(msg, context, offset=2 * offset)
```
ilevkivskyi: All changes in this function are unnecessary now, except `plural_s` usage.

mypy/messages.py
```diff
@@ -1233,14 +1238,49 @@ def report_protocol_problems(self, subtype: Union[Instance, TupleType, TypedDict
...
+        max_available = len(func.items())
+
+        self.note('Possible overload variant{}:'.format(plural_s(shown)), context)
+        for item in targets[:max_items]:
+            self.note(self.pretty_callable(item), context, offset=2 * offset)
+
+        assert shown <= max_matching <= max_available
+        if shown < max_matching <= max_available:
+            left = max_matching - shown
+            msg = '<{} more matching overload{} not shown, out of {} total overloads>'.format(
```
ilevkivskyi: I would rather say this as `{} more matching and {} non-matching overload{} not shown`. But this is still not perfect because the overloads aren't "matching" (as the error message above says). Anyway, I think it is not so important to delay merging this PR, so I will go ahead unless someone proposes better wording.

ilevkivskyi:
@Michael0x2a Will you have time to make minor final updates here before I cut the release branch tomorrow morning? (If not it is also OK.)

Michael0x2a:
@ilevkivskyi -- most likely. I'm planning on leaving work + finishing dinner in roughly 5 hours or so and can start making any necessary changes after that.

I guess the timing might be a little tight, depending on whether you mean morning Dublin time vs Pacific time, but eh, whatever.

Michael0x2a:
@gvanrossum and @ilevkivskyi  -- Mkay, I reverted the changes to `pretty_overload` (apart from some tweaks to make pluralization tidier), swapped out the word "matching" for "similar" for the first error message, and added the tweak Guido suggested (it ended up being pretty simple to implement in the end).

I'll merge this in once the tests pass (unless there are any last-minute objections, I guess).

PR #5299, opened 2018-06-29
Make NamedTuple attributes read-only

### LINKED ISSUE #5293: NamedTuple should not satisfy Protocol with mutable fields
The following code doesn't produce a type error, but will crash at runtime, because the named tuple's fields cannot be set

```python
from typing import NamedTuple

class HasX(Protocol): 
    x: str
    
class A(NamedTuple):
    x: str

a: HasX = A("foo")
a.x = "bar"
```
Since `HasX` requires `x` to be settable, `A` should not be a structural subtype of `HasA`.

ethanhs:
It seems `namedtuple`s are covered already, but `typing.NamedTuple`s are not. This fixes #5293 .

### DIFF
--- a/mypy/subtypes.py
+++ b/mypy/subtypes.py
@@ -524,7 +524,7 @@ def get_member_flags(name: str, info: TypeInfo) -> Set[int]:
         if v.var.is_staticmethod or v.var.is_classmethod:
             return {IS_CLASS_OR_STATIC}
     # just a variable
-    if isinstance(v, Var):
+    if isinstance(v, Var) and not v.is_property:
         flags = {IS_SETTABLE}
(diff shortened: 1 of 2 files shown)


gvanrossum:
Can you add a test to make sure that if the variable is a property not because it's a NamedTuple and if the property defines a setter, this won't incorrectly reject setting such a property?
```py
@property
def foo(self): ...
@foo.setter
@def foo(self, value): ...
```

ethanhs:
I'm not sure if I can write a test that wouldn't exactly duplicate the functionality of  `testSettablePropertyInProtocols` or `testPropertyWithSetter`... but if you have a suggestion I am more than happy to hear it.

PR #5303, opened 2018-07-01
Silence modules in site-packages and typeshed

### LINKED ISSUE #5260: Allow PEP 561 packages to specify mypy settings
[This is a summary of a discussion on the typing gitter, if I missed anything please do mention it]
With the expectation of many new packages being created conforming to PEP 561, @ilevkivskyi suggested that package authors should be able to set strictness settings on their code.

Ivan's original suggestion is to allow setting some flags like `--ignore-missing-imports`, `--no-strict-optional` in the source file(s) themselves.

I suggested per-package, as defining and implementing a new format seemed like a lot of work. Also, I generally am against metadata in source code, as I find it noisy.

@Michael0x2a wanted the ability to override this.

Michael then suggested we mandate PEP 561 packages provide a config file and mypy can read per-module config options from said file.

Ivan suggested amending the suggestion to mypy will *allow* per-module config files in PEP 561 packages.

I concurred with Ivan.

# **tl;dr**

I suggest that mypy read per-module config file settings, if a config file is found.

ethanhs:
Fixes #5260 .


ethanhs:
Huh, this seems to reveal a bug in mypy:

```
from typing import Sequence
from os import fspath
s: Sequence[str]

s = tuple(map(fspath, s))
reveal_type(s)  # error: Revealed type is '<nothing>'
```

ilevkivskyi:
@ethanhs Is this a regression or it was always like this? (There is a chance this is related to a recent overload overhaul.)

ethanhs:
>Is this a regression or it was always like this?

Not sure, I created the repro with https://mypy-play.net, which uses 0.610, I haven't tested it on older versions.

ilevkivskyi:
No, it is a `meet.is_overlapping_types` problem which is still quite simplistic, it considers `typing.Sequence[builtins.str]` and `builtins.tuple[builtins.str]` non-overlapping (surprise).

The fix should be straightforward, but if you don't have time, just open an issue about this.

test-data/unit/fine-grained-modules.test
```diff
@@ -751,15 +751,16 @@ main:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports"
 main:2: error: Incompatible types in assignment (expression has type "int", variable has type "str")
 
 [case testAddFileWhichImportsLibModuleWithErrors]
+# flags: --no-silence-site-packages
```
ilevkivskyi: Why is this necessary here? Is `broken` installed as a PEP 561 package?

mypy/util.py
```diff
@@ -204,3 +207,91 @@ def replace_object_state(new: object, old: object) -> None:
                         setattr(new, attr, getattr(old, attr))
                     elif hasattr(new, attr):
                         delattr(new, attr)
+
+
+# backport commonpath for 3.4
+if sys.platform == 'win32':
+    def commonpath(paths: Sequence[str]) -> str:
```
ilevkivskyi: Do we need the general algorithm here? As I understand we essentially only need something like `is_sub_path(p1, p2)`. Not that I think it is important, but maybe it will allow to simplify code?

test-data/unit/fine-grained-modules.test
ethanhs: Since opening this, I also implemented silencing typeshed, which `broken` is simulated as being part of via `alt_lib_path`.

mypy/util.py
ethanhs: I was originally hoping to eventually use `os.path.commonpath` once we drop support for 3.4, but I think you are right it would be better to do this ourselves. I will change this when I have time later.

ethanhs:
>I think we should add a test that installs a PEP 561 with various errors

I will add some errors to typedpkg to test the functionality.

ilevkivskyi:
> I will add some errors to typedpkg to test the functionality.

OK, thanks! Just note that I would like to cut the release branch tomorrow morning. So it would be better to have the tests by that time.

ethanhs:
>Just note that I would like to cut the release branch tomorrow morning. So it would be better to have the tests by that time.

Sure thing, I will do that momentarily.

PR #5315, opened 2018-07-04
Improve Tuple/Sequence/Iterable overlapping

### LINKED ISSUE #4466: assigning a `Tuple[int, ...]` to an `Iterable[int]` results in `builtins.None` or `<nothing>`
Bug with this code:

```
from typing import Iterable, Tuple
x: Iterable[int] = ()
y: Tuple[int, ...] = (1, 2, 3)
x = y
reveal_type(x)
```

Actual result (on mypy master, Python 3.6):

`tuplebug.py:9: error: Revealed type is 'builtins.None'`

Actual result with `--strict-optional`:

`tuplebug.py:9: error: Revealed type is '<nothing>'`

Expected result:

The revealed type should be `Tuple[int, ...]`.

Via git bisect, I find that this bug was introduced recently, in https://github.com/python/mypy/pull/4238 (cc @elazarg, @JukkaL ).

ethanhs:
Fixes #4466 

@ilevkivskyi hopefully this is similar to what you were expecting?

### DIFF
--- a/mypy/meet.py
+++ b/mypy/meet.py
@@ -142,6 +142,17 @@ def is_overlapping_types(t: Type, s: Type, use_promotions: bool = False) -> bool
     if tup_overlap is not None:
         return tup_overlap
 
+    if isinstance(t, TupleType):
+        if isinstance(s, Instance) and is_subtype(t, s):
+            if all(is_overlapping_types(ti, si, use_promotions)
+                    for ti, si in zip(t.items, s.args)):
+                return True
+    if isinstance(s, TupleType):
+        if isinstance(t, Instance) and is_subtype(s, t):
+            if all(is_overlapping_types(ti, si, use_promotions)
+                    for ti, si in zip(s.items, t.args)):
+                return True
+
     if isinstance(t, Instance):
         if isinstance(s, Instance):
             # Consider two classes non-disjoint if one is included in the mro
@@ -192,9 +203,8 @@ def is_overlapping_tuples(t: Type, s: Type, use_promotions: bool) -> Optional[bo
                 if all(is_overlapping_types(ti, si, use_promotions)
                        for ti, si in zip(t.items, s.items)):
                     return True
-        # TupleType and non-tuples do not overlap
-        return False
-    # No tuples are involved here
+    # TupleType and non-tuples are handled later
+    # Otherwise, no tuples are involved
     return None
 
 
@@ -323,7 +333,8 @@ class TypeMeetVisitor(TypeVisitor[Type]):
             return TupleType(items, t.fallback)
         # meet(Tuple[t1, t2, <...>], Tuple[s, ...]) == Tuple[meet(t1, s), meet(t2, s), <...>].
         elif (isinstance(self.s, Instance) and
-              self.s.type.fullname() == 'builtins.tuple' and self.s.args):
+              (self.s.type.fullname() == 'builtins.tuple' or is_subtype(t, self.s))
+              and self.s.args):
             return t.copy_modified(items=[meet_types(it, self.s.args[0]) for it in t.items])
         elif (isinstance(self.s, Instance) and t.fallback.type == self.s.type):
             # Uh oh, a broken named tuple type (https://github.com/python/mypy/issues/3016).
--- a/test-data/unit/check-tuples.test
+++ b/test-data/unit/check-tuples.test
@@ -1065,3 +1065,27 @@ class CallableTuple(Tuple[str, int]):
         return n
 
 [builtins fixtures/tuple.pyi]
+
+[case testTupleCompatibleWithSequence]
+from typing import Sequence
+s: Sequence[str]
+s = tuple()
+reveal_type(s) # E: Revealed type is 'builtins.tuple[builtins.str]'
+
+[builtins fixtures/tuple.pyi]
+
+[case testTupleInstanceCompatibleWithIterable]
+from typing import Iterable, Tuple
+x: Iterable[int] = ()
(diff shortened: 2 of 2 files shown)


mypy/meet.py
```diff
@@ -192,9 +203,8 @@ def is_overlapping_tuples(t: Type, s: Type, use_promotions: bool) -> Optional[bo
                 if all(is_overlapping_types(ti, si, use_promotions)
                        for ti, si in zip(t.items, s.items)):
                     return True
-        # TupleType and non-tuples do not overlap
-        return False
-    # No tuples are involved here
```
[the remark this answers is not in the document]
ethanhs: I fixed this and added a regression test. Thank you for pointing this out.

mypy/meet.py
```diff
@@ -138,9 +138,11 @@ class C(A, B): ...
...
     # We must check for TupleTypes before Instances, since Tuple[A, ...]
     # is an Instance
-    tup_overlap = is_overlapping_tuples(t, s, use_promotions)
-    if tup_overlap is not None:
-        return tup_overlap
+    if is_overlapping_tuples(t, s, use_promotions):
+        return True
+
+    if is_subtype(t, s) or is_subtype(s, t):
+            return True
```
ilevkivskyi: On one hand I would like to see this general case somewhere at the top of this function as an obvious case of overlap. However the fact that such an obvious case is still not here makes me think of "Chesterton's fence". Is there any danger of putting this here (like performance costs)?

An alternative may be to put `return is_subtype(t, s) or is_subtype(s, t)` instead of `return False` near the end of `is_overlapping_tuples`.

If you think the performance is still fine, then I would keep this check here (or maybe even higher above), but add few more tests that cover the other (non-tuple) situations not covered by existing `if` branches.

Michael0x2a: Yeah, I agree performance is something we might need to think about here, since we're basically traversing the types three times instead of once now. I think the subtype cache might help mitigate this, though of course it's hard to say for certain w/o benchmarking.

I suspect the other reason why the function doesn't already do something like this was because it was originally intended to be super-simple/lightweight -- here's a [snapshot from 2015](https://github.com/python/mypy/blob/081532d1e291206cdf6df7db1c68d4a8667f4adc/mypy/meet.py#L50) and [from 2016](https://github.com/python/mypy/blob/e308f2f786d432577286e0d84abaca505d02ec81/mypy/meet.py#L39), for example. I guess it then just steadily accumulated patches over the years?

ethanhs: I don't particularly have a great idea of how I could benchmark this, but I am happy to move it to the end if you think that is better.

ilevkivskyi: > I don't particularly have a great idea of how I could benchmark this

A good benchmark is self checking `mypy mypy`.

mypy/meet.py
```diff
@@ -192,10 +191,10 @@ def is_overlapping_tuples(t: Type, s: Type, use_promotions: bool) -> Optional[bo
                 if all(is_overlapping_types(ti, si, use_promotions)
                        for ti, si in zip(t.items, s.items)):
                     return True
-        # TupleType and non-tuples do not overlap
-        return False
-    # No tuples are involved here
-    return None
+        return is_subtype(t, s) or is_subtype(s, t)
```
ilevkivskyi: > Anyway, the selfcheck test seems to perform the same within a few hundredths of a second on master and this branch (without cache in both cases), so I am not concerned about performance regressions.

Did you measure performance when this check is here or at the top of the `is_overlapping_types`? If at the top of `is_overlapping_types`, then I would rather add `if is_subtype(t, s) or is_subtype(s, t): return True` at the top of `is_overlapping_types` (after `builtins.object` check) and leave this function completely "intact" then.

ethanhs: It was indeed measured from the top. I just added it after the object check.

mypy/meet.py
```diff
@@ -138,9 +140,8 @@ class C(A, B): ...
 
     # We must check for TupleTypes before Instances, since Tuple[A, ...]
     # is an Instance
-    tup_overlap = is_overlapping_tuples(t, s, use_promotions)
-    if tup_overlap is not None:
-        return tup_overlap
+    if is_overlapping_tuples(t, s, use_promotions):
+        return True
```
ilevkivskyi: I think this change is now unnecessary (and the corresponding change below) after you added the check at the top of this function.

Michael0x2a:
@ilevkivskyi -- sure! I have to go to a thing, but I can re-run these tests a little later tonight.

@ethanhs -- I'm not sure if your PR will run into this case or not, but I remember running into this edge case while working on my other PR yesterday. Could you try checking to see what your PR does on this test case?

```python
from typing import *

# E.g. maybe this was imported from a non-typed module
Parent: Any

class Child(Parent):
    def some_method(self) -> int:
        if self is None:
            return None
        return 3
```

I believe the current behavior of mypy is to report no errors when running in strict-optional mode. If your PR complains about the "return None", you may have to swap out the two calls to `is_subtype` with `is_proper_subtype` or something.

(The actual code that I synthesized this from does `self = super(Child, self).blah()` before the if statement, so the above example is a little less stupid than it looks, if only marginally.)

JelleZijlstra:
Maybe I'm missing something, but shouldn't that code sample produce an error under strict optional? It returns None but is declared to return int.

ilevkivskyi:
@JelleZijlstra mypy infers the branch as unreachable.

ethanhs:
@Michael0x2a thanks! That was indeed an issue, and `is_proper_subtype` fixed it :)

I also reverted the unneeded changes as suggested by Ivan.

PR #5301, opened 2018-06-29
Fail fast if site-packages are in the path

ethanhs:
With my refactoring in #5256, I realized that it would be pretty straightforward to give some nicer warnings if site-packages were in any of the PATHs.

PR #5318, opened 2018-07-04
Fix running mypy from editable install directory

ethanhs:
To fix this we skip checking egg directories, which is the simplest solution to this problem.


gvanrossum:
Would it be hard to add a test?

ethanhs:
>Would it be hard to add a test?

Yes, I think so :(

The tests are designed to run in unique venvs to be isolated. This means running things from the package directory won't really work.

I will look at how the tests can be modified to make that easier. If you'd like, we could simply revert https://github.com/python/mypy/pull/5301 for now, until I have a better idea of how best to test it.

PR #5321, opened 2018-07-04
Remove and refactor old overload selection logic

Michael0x2a:
This commit removes a few remnants of the old overload selection algorithm in checkexpr.

Specifically, this commit:

1.  Modifies `erased_signature_similarity` so it returns a bool instead of an int.

2.  Simplifies how `erased_signature_similarity` handles types like `Type[X]`.

    Note: although this change relaxes and loosens the precision of `erased_signature_similarity`, it will not impact correctness: we now use this method exclusively as a heuristic to help us determine which overloads the user might have meant to select in our error messages.

    As a consequence, loosening this function actually led to a more helpful error message in one or two cases.

3.  Remove the `match_signature_types` method. It didn't appear like anybody was calling it.


Michael0x2a:
As a note, this PR is completely independent and unrelated to my other overload-related PRs.

It's also a pretty low-priority cleanup -- I don't think it's particularly important that we get this merged any time soon.

test-data/unit/check-classes.test
```diff
@@ -1223,11 +1223,8 @@ class D:
     def __get__(self, inst: Base, own: Type[Base]) -> str: pass
 [builtins fixtures/bool.pyi]
 [out]
-main:5: error: Revealed type is 'Any'
-main:5: error: No overload variant of "__get__" of "D" matches argument types "None", "Type[A]"
-main:5: note: Possible overload variants:
-main:5: note:     def __get__(self, inst: None, own: Type[Base]) -> D
-main:5: note:     def __get__(self, inst: Base, own: Type[Base]) -> str
+main:5: error: Revealed type is 'd.D'
```
ilevkivskyi: I am not sure this is a good idea to have `d.D` instead of `Any`. In case of errors we typically return `Any` to avoid other (potentially bogus) errors.

test-data/unit/check-classes.test
```diff
@@ -1223,11 +1223,8 @@ class D:
...
+main:5: error: Argument 2 to "__get__" of "D" has incompatible type "Type[A]"; expected "Type[Base]"
```
ilevkivskyi: Why don't we have the same nicer error for the case right below? (This is very low priority.)

mypy/checkexpr.py
```diff
@@ -3295,101 +3259,68 @@ def visit_uninhabited_type(self, t: UninhabitedType) -> bool:
...
-            (isinstance(actual, Instance) and actual.type.fallback_to_any)):
-        # These could match anything at runtime.
-        return 2
+
+    # Callable or Type[...]-ish types
+
+    def is_typetype_like(typ: Type) -> bool:
+        return (isinstance(typ, TypeType)
+                or (isinstance(typ, FunctionLike) and typ.is_type_obj())
+                or (isinstance(typ, Instance) and typ.type.fullname() == "builtins.object"))
```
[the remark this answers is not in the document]
Michael0x2a: Hmm, good point. I changed that to `builtins.type`.

test-data/unit/check-classes.test
Michael0x2a: I think in this case, what's happening is that there's actually only one plausible matching overload -- when we do `A.f` and `f` is a descriptor, we always end up calling the overload `def __get__(self, inst: None, own: Type[Base]) -> D`.

So, since only one variant is a plausible match, we just infer that variant's return type. (And regarding your comment below -- this is also why we don't bother showing the possible overload variants here).

I think this behavior is consistent with how we handle bad calls to regular functions. E.g. mypy will infer the return type instead of Any in this program:

    def foo(x: int) -> int: 
        return x

    a = foo("asdf")   # Error
    reveal_type(a)  # Revealed type is 'builtins.int', not Any

Granted, I don't think it's obvious that only one variant will be selected here, so I added another test case for this behavior.

Michael0x2a:
@ilevkivskyi -- this PR is also now ready for a second look whenever!

test-data/unit/check-classes.test
ilevkivskyi: OK.

PR #5326, opened 2018-07-06
Remove Python 2 references in Python 3 cheatsheet

JelleZijlstra:
The Python 3 cheatsheet is for people who can use the Python 3 syntax and don't need to care about Python 2 compatibility, so I don't think it's useful to mention Python 2 compatibility issues.

Also removed some unused imports.


gvanrossum:
I was initially hesitant but in the end you convinced me.

PR #5327, opened 2018-07-06
Sync typeshed

ilevkivskyi:
In preparation for the release.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -39,6 +39,8 @@ def daemonize(func: Callable[[], None], log_file: Optional[str] = None) -> int:
     subprocess killed by signal.
     """
     # See https://stackoverflow.com/questions/473620/how-do-you-create-a-daemon-in-python
+    if sys.platform == 'win32':
+        raise ValueError('Mypy daemon is not supported on Windows yet')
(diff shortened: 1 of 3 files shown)


ilevkivskyi:
> Did you check if this necessitates any changes to Dropbox code bases?

Testing now, but apparently we have another problem (see test failures). I am looking into it.

ilevkivskyi:
OK, the failures are because of recently added Windows platform checks.

gvanrossum:
Maybe when we run stdlib-samples we should force the platform to linux.

ilevkivskyi:
All checks were added in https://github.com/python/typeshed/commit/1ae2ba0fbe4257397916a4921305d5a69d67df86

> Maybe when we run stdlib-samples we should force the platform to linux.

Self-check also fails. Because daemon is not supported on Windows yet. `testpr` is not done yet, but I remember there were some problems on C when we last attempted using platform checks in typeshed.

gvanrossum:
LMK if you need another approval.

ilevkivskyi:
> LMK if you need another approval.

Are you OK with the change in `dmypy_server.py`? (Other changes are trivial.)

PR #5333, opened 2018-07-09
Reorganize command line documentation

Michael0x2a:
This commit reorganizes our documentation about the command line along with several miscellaneous fixes.

In short, I moved the stuff about running mypy/managing imports to a separate file and reorganized the command line section.

Specific major changes include:

1.  I split up information about running mypy and information about mypy's command line flags into two separate pages: "Running mypy and managing imports" and "The mypy command line".

    Rationale: I felt the current documentation on running mypy is very dense: it includes a lot of info on managing imports, how mypy discovers files, etc that I think were really only of interest to advanced programmers who are in the middle of trying to get mypy to run on a complex code base.

    I think splitting up the text into two makes it easier for both beginners and experts to get the info they're looking for.

2.  I rewrote the section about how mypy handles the `--ignore-missing-imports` and `--follow-imports` flags (Now located in "Running mypy..." > "How mypy handles imports").

    In particular, it more clearly distinguishes how to handle missing imports vs handling imports to files you didn't want type check: the former is applicable to everybody; the latter is only really applicable to people who are progressively adding type hints.

3.  I moved information about how mypy searches the file system and resolves imports to the bottom of the "Running mypy..." section.

    I think most people don't really need that info to start: running `mypy my_whole_codebase` is generally a good starting point.

4.  I reorganized the command line section to mirror how the flags are grouped when you run `mypy --help`.

Some minor changes:

1.  I switched to using [definition lists][0] instead of bullet points to document flags: I think it looks cleaner.

2.  I removed the `mypy --help` summary from the command line page: I don't really think anybody was reading that at all.

3.  I added documentations about the `--warn-unused-casts`, `-warn-unused-ignores`, `--show-error-contexts`, `--warn-unsed-configs`, and the different report flags. Those flags were apparently undocumented.

    I did *not* document `--cache-fine-grained` and `--memory-xml-report` mostly because I don't actually understand what exactly those do.

4.  I added moved the note about integrating mypy into a separate page, and added a brief note about plugins to that page.

  [0]: https://sphinx-rtd-theme.readthedocs.io/en/latest/demo/lists_tables.html#definition-lists

PR #5334, opened 2018-07-09
Adjust output of --help text to mirror command line changes

Michael0x2a:
This commit adjusts the output of running `mypy --help` to mirror the changes made in https://github.com/python/mypy/pull/5333.

Specifically, it rearranges the order of the flags slightly and replaces the summary of the flags at the top with a short description of how to use mypy along with links for learning more.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -396,17 +396,17 @@ def process_options(args: List[str],
     # https://github.com/python/mypy/wiki/Documentation-Conventions
 
     general_group = parser.add_argument_group(
-            title='Optional arguments')
+        title='Optional arguments')
     general_group.add_argument(
-            '-h', '--help', action='help',
-            help="Show this help message and exit")
+        '-h', '--help', action='help',
+        help="Show this help message and exit")
     general_group.add_argument(
-            '-v', '--verbose', action='count', dest='verbosity',
-            help="More verbose messages")
+        '-v', '--verbose', action='count', dest='verbosity',
+        help="More verbose messages")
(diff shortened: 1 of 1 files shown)


Michael0x2a:
For reference, the new output: https://gist.github.com/Michael0x2a/0bdefea47267b59dd7d3810088f88669

Note that it links to a page introduced in https://github.com/python/mypy/pull/5333.

mypy/main.py
```diff
@@ -297,6 +294,37 @@ def infer_python_version_and_executable(options: Options,
         options.python_executable = special_opts.python_executable
 
 
+HEADER = """%(prog)s [-h] [-v] [-V] [more options; see below]
+            [-m MODULE] [-p PACKAGE] [-c PROGRAM_TEXT] [files [files ...]]"""
```
gvanrossum: Isn't `[[files [files ...]]` redundant? I'd just say `[files ...]`.

Michael0x2a:
@gvanrossum -- I 100% agree with your comments regarding capitalization -- I was about to do it the way you suggested back when I made my first wave of command line improvements, but reluctantly gave up partly because of things like `usage:` and partly because it seemed like it went against argparse's conventions.

Anyways, I'll make the fixes you suggested later tonight.

Michael0x2a:
@gvanrossum -- this is ready for another look. The new help text looks like this: 

https://gist.github.com/Michael0x2a/7220f9a088aac51bbe13f9078d521793

PR #5338, opened 2018-07-09
Support type aliases in builtins

ilevkivskyi:
This fixes a recently discovered internal crash. This should probably go into the release, since this is a regression.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -19,7 +19,7 @@ from mypy.nodes import (
     UnicodeExpr, OpExpr, UnaryExpr, LambdaExpr, TempNode, SymbolTableNode,
     Context, Decorator, PrintStmt, BreakStmt, PassStmt, ContinueStmt,
     ComparisonExpr, StarExpr, EllipsisExpr, RefExpr, PromoteExpr,
-    Import, ImportFrom, ImportAll, ImportBase,
+    Import, ImportFrom, ImportAll, ImportBase, TypeAlias,
     ARG_POS, ARG_STAR, LITERAL_TYPE, MDEF, GDEF,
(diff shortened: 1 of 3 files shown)


mypy/checker.py
```diff
@@ -3171,6 +3171,9 @@ def named_type(self, name: str) -> Instance:
         # Assume that the name refers to a type.
         sym = self.lookup_qualified(name)
         node = sym.node
+        if isinstance(node, TypeAlias):
+            assert isinstance(node.target, Instance)
```
msullivan: It can't ever be a chain of `TypeAlias`es?

ilevkivskyi: It can't, chained aliases are collapsed on sight. An alias can target other types (non-`Instance`) but this function is supposed to be used only with built-in types, that are all instances.

ilevkivskyi:
> How many places do we need to do this sort of check? If it is a lot, we might want a function that we call to collapse out `TypeAlias`es for this sort of thing.

This is actually a quite common pattern. I think I have seen at least three or four other occurrences of the same pattern. I will try to factor them all out.

ilevkivskyi:
> > How many places do we need to do this sort of check? If it is a lot, we might want a function that we call to collapse out `TypeAlias`es for this sort of thing.

> This is actually a quite common pattern. I think I have seen at least three or four other occurrences of the same pattern. I will try to factor them all out.

I have found 11 places where a similar pattern is used, but the problem is that all of them are not like others (mostly we just want to detect unsubscripted aliases in different contexts, but don't rebind the node). I tried to play with it, but it looks like defining such a function will actually complicate the code.

I will then merge this one now, maybe we can revisit this refactoring later.

PR #5339, opened 2018-07-09
Add builtin aliases when updating typing

msullivan:
typing updates can be provoked by triggering protocols from typing.

Failing to do this can manifest as crashes and spurious "invalid type"
messages.

### DIFF
--- a/test-data/unit/fine-grained.test
+++ b/test-data/unit/fine-grained.test
@@ -1233,9 +1233,10 @@ main:2: error: Module 'a' has no attribute 'f'
 [case testRefreshTyping]
 from typing import Sized
 from c import A
+import z
 
 # Force typing to get refreshed by using a protocol from it
-x : Sized = A()
+x: Sized = A()
 
 [file c.py]
 class D:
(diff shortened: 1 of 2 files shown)


test-data/unit/fine-grained.test
```diff
@@ -1230,6 +1230,31 @@ def f() -> None: pass
 ==
 main:2: error: Module 'a' has no attribute 'f'
 
+[case testRefreshTyping]
+from typing import Sized
+from c import A
+
+# Force typing to get refreshed by using a protocol from it
+x : Sized = A()
```
ilevkivskyi: No space before colon.

test-data/unit/fine-grained.test
```diff
@@ -1230,6 +1230,31 @@ def f() -> None: pass
...
+[file c.py.2]
+class C:
+    def __len__(self) -> int: return 0
+A = C
+[file z.py]
+from typing import Tuple, List
+T = List[int]
+[file z.py.2]
+from typing import Tuple, List
+T = List[int]
```
ilevkivskyi: Few questions here:
* Why `Tuple` also needs to be imported
* It looks like there are no changes in `z.py` between runs, is this intentional?
* Should `z` be imported somewhere.

msullivan: `Tuple` doesn't need to be.
Should be a change, got busted while iterating.
Doesn't need to be imported in fine-grained tests.

PR #5346, opened 2018-07-12
Fix linkcheck errors

ceh:
Fix attrs' links

 - www.attrs.org doesn't have a SSL certificate, so don't link to HTTPS.
 - Fix invalid anchor tag for validator decorators.

Use HTTPS scheme for Zulip blog link

Fixes the following linkcheck errors:

```
(line   12) broken    https://www.attrs.org/en/stable - HTTPSConnectionPool(host='www.attrs.org', port=443): Max retries exceeded with url: /en/stable (Caused by SSLError(SSLCertVerificationError("hostname 'www.attrs.org' doesn't match either of '*.readthedocs.org', 'readthedocs.org'")))
(line   78) broken    http://www.attrs.org/en/stable/examples.html#decorator - Anchor 'decorator' not found
```

```
(line  187) redirect  http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/ - permanently to https://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/
```

### DIFF
--- a/docs/source/additional_features.rst
+++ b/docs/source/additional_features.rst
@@ -9,7 +9,7 @@ of the previous sections.
 The attrs package
 *****************
 
-`attrs <https://www.attrs.org/en/stable>`_ is a package that lets you define
+`attrs <http://www.attrs.org/en/stable>`_ is a package that lets you define
 classes without writing boilerplate code. Mypy can detect uses of the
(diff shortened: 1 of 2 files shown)


gvanrossum:
Thanks!

PR #5350, opened 2018-07-13
Add docs for dataclasses

ilevkivskyi:


docs/source/additional_features.rst
```diff
@@ -53,6 +53,65 @@ That enables this to work:
         two: Dict[str, str] = attr.Factory(dict)
         bad: str = attr.ib(16)   # Error: can't assign int to str
 
+In Python 3.7, similar functionality have been added to the standard library
```
JelleZijlstra: have -> has

docs/source/additional_features.rst
```diff
@@ -82,6 +141,9 @@ Caveats/Known Issues
 * Method definitions added by mypy currently overwrite any existing method
   definitions.
 
+* Some functions in ``dataclasses`` module, such as ``replace()`` and ``asdict()``,
```
JelleZijlstra: in *the* dataclasses module

docs/source/additional_features.rst
```diff
@@ -53,6 +53,65 @@ That enables this to work:
...
+
+    @dataclass
+    class Application:
+        name: str
+        plugins: List[str] = field(default_factory=list)
+
+    test = Application("Testing...")  # OK
+    bad = Application("Testing...", "with plugin")  # Error: List[str] expected
+
+Mypy will detect special methods depending on the flags used to define dataclasses.
```
JukkaL: Maybe give examples of these to be more explicit (`__lt__`, for example).

docs/source/additional_features.rst
```diff
@@ -6,8 +6,8 @@ of the previous sections.
 
 .. _attrs_package:
 
-The attrs package
-*****************
+The attrs package and dataclasses
+*********************************
```
JukkaL: Maybe create a separate section for dataclasses? New Python users might not be familiar with attrs but may know dataclasses since they are in the standard library, so I think that it makes sense to give them both equal weight, so that it's possible to learn about dataclass support without having to first read about attrs.

PR #5367, opened 2018-07-17
Add experimental hidden flag to tweak dependencies

ilevkivskyi:
This flag will tweak fine grained dependencies to reflect propagation of type (im)precision instead of actual semantic dependencies. The flag should never be used to run actual type check, _only_ to analyze type coverage in a code base.

The flag could be useful when annotating legacy code, to check which functions (decorators, base classes, variables) are most important to annotate first.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -612,6 +612,11 @@ def process_options(args: List[str],
     # --local-partial-types disallows partial types spanning module top level and a function
     # (implicitly defined in fine-grained incremental mode)
     parser.add_argument('--local-partial-types', action='store_true', help=argparse.SUPPRESS)
+    # --tweaked-deps adds some more dependencies that are not semantically needed, but
+    # may be helpful to determine relative importance of classes and functions for overall
+    # type precision in a code base. It also _removes_ some deps, so this flag should be never
+    # used except for generating code stats.
+    parser.add_argument('--tweaked-deps', action='store_true', help=argparse.SUPPRESS)
     # --bazel changes some behaviors for use with Bazel (https://bazel.build).
     parser.add_argument('--bazel', action='store_true', help=argparse.SUPPRESS)
     # --package-root adds a directory below which directories are considered
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -612,11 +612,12 @@ def process_options(args: List[str],
     # --local-partial-types disallows partial types spanning module top level and a function
     # (implicitly defined in fine-grained incremental mode)
     parser.add_argument('--local-partial-types', action='store_true', help=argparse.SUPPRESS)
-    # --tweaked-deps adds some more dependencies that are not semantically needed, but
+    # --logical-deps adds some more dependencies that are not semantically needed, but
     # may be helpful to determine relative importance of classes and functions for overall
     # type precision in a code base. It also _removes_ some deps, so this flag should be never
-    # used except for generating code stats.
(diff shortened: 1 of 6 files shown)


mypy/server/deps.py
```diff
@@ -201,6 +205,10 @@ def visit_decorator(self, o: Decorator) -> None:
         # generate dependency.
         if not o.func.is_overload and self.scope.current_function_name() is None:
             self.add_dependency(make_trigger(o.func.fullname()))
+        if self.options is not None and self.options.tweaked_deps:
+            for d in o.decorators:
+                if isinstance(d, RefExpr) and d.fullname is not None:
```
JukkaL: Also look for things like `@foo(...)` (i.e. call expressions).

mypy/server/deps.py
```diff
@@ -201,6 +205,10 @@ def visit_decorator(self, o: Decorator) -> None:
...
+                    self.add_dependency(make_trigger(d.fullname), make_trigger(o.func.fullname()))
```
JukkaL: Add comment explaining why we want to have this.

mypy/server/deps.py
```diff
@@ -263,6 +271,9 @@ def process_type_info(self, info: TypeInfo) -> None:
                                         target=make_trigger(info.fullname() + '.' + name))
         for base_info in non_trivial_bases(info):
             for name, node in base_info.names.items():
+                if self.options and self.options.tweaked_deps:
+                    if name not in info.names:
```
JukkaL: Add comment describing why we want to do this (with an example perhaps).

mypy/server/deps.py
```diff
@@ -368,6 +379,21 @@ def visit_assignment_stmt(self, o: AssignmentStmt) -> None:
             if o.type:
                 for trigger in get_type_triggers(o.type):
                     self.add_dependency(trigger)
+        if self.options and self.options.tweaked_deps:
+            if (isinstance(rvalue, CallExpr) and isinstance(rvalue.callee, RefExpr)
```
JukkaL: Add comment explaining why we do this.

mypy/main.py
```diff
@@ -612,6 +612,11 @@ def add_invertible_flag(flag: str,
     # --local-partial-types disallows partial types spanning module top level and a function
     # (implicitly defined in fine-grained incremental mode)
     parser.add_argument('--local-partial-types', action='store_true', help=argparse.SUPPRESS)
+    # --tweaked-deps adds some more dependencies that are not semantically needed, but
```
JukkaL: Mention that this is an experimental, undocumented feature that may go away or be changed at any time.

mypy/main.py
```diff
@@ -612,6 +612,11 @@ def add_invertible_flag(flag: str,
...
+    # may be helpful to determine relative importance of classes and functions for overall
+    # type precision in a code base. It also _removes_ some deps, so this flag should be never
+    # used except for generating code stats.
+    parser.add_argument('--tweaked-deps', action='store_true', help=argparse.SUPPRESS)
```
JukkaL: Maybe rename to something like `--logical-deps` or `--semantic-deps` and make this imply `--cache-fine-grained`? The motivation is that this makes the dependencies more independent of the details of how fine-grained incremental checking works.

ilevkivskyi: I went with `--logical-deps` it sounds better.

Issue #1270, opened 2016-03-04
Ordering matters for @overload


mgeisler:
I am trying to model a generic container (a [Pandas Series](http://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.html) though it shouldn't matter too much for this). The container has a generic "value type" and a generic "index type". When iterating over the container, you get the value type back. So I tried with:

``` python
class Series(Generic[VT, IT], Sequence[VT]):

    @overload
    def __getitem__(self, key: slice) -> List[VT]: ...

    @overload
    def __getitem__(self, key: int) -> VT: ...
```

This gives me an error:

```
stubs/pandas/__init__.pyi: note: In class "Series":
stubs/pandas/__init__.pyi:186: error: Signature of "__getitem__" incompatible with supertype "Sequence"
```

That baffled me for a while and on a whim, I tried reordering the two signatures:

``` python
class Series(Generic[VT, IT], Sequence[VT]):

    @overload
    def __getitem__(self, key: int) -> VT: ...

    @overload
    def __getitem__(self, key: slice) -> List[VT]: ...
```

The error is now gone! :-)

When I first got the error, I tried going into `typing.py` to see exactly what signature Sequence has. That didn't help me since Sequence is defined as

``` python
class Sequence(Sized, Iterable[T_co], Container[T_co],
               extra=collections_abc.Sequence):
    pass
```

I think it could have been helpful if mypy could have written the exact types it was comparing.

This was tested with 0ec0cb4.

JukkaL:
Yeah, mypy should show the signature of the superclass method. 

Also, mypy could perhaps allow you to switch the order of overload variants in case the signatures aren't overlapping, though this is a little questionable as the order could plausibly affect which overload gets picked in a call.

ddfisher:
@JukkaL - what do you mean by "switch the order of overload variants"?  Is there some obvious natural ordering that I'm missing?

JukkaL:
By order of overload variants I mean the textual order of the `@overload` decorated functions. For example, this is not valid right now:

```
class A:
    @overload
    def f(self, x: X) -> int: ...
    @overload
    def f(self, x: Y) -> str: ...

class B(A):
    @overload   # note that Y -> str is now the first variant
    def f(self, x: Y) -> str: ...
    @overload
    def f(self, x: X) -> int: ...
```

[...]

PR #5369, opened 2018-07-17
Improve error message for overload overrides w/ swapped variants

Michael0x2a:
This commit is intended to resolve https://github.com/python/mypy/issues/1270.

Currently, running mypy on the following program:

    from typing import overload

    class Parent:
        @overload
        def foo(self, x: A) -> A: ...
        @overload
        def foo(self, x: B) -> B: ...

    class Child(Parent):
        @overload
        def foo(self, x: B) -> B: ...
        @overload
        def foo(self, x: A) -> A: ...

...will make mypy report the following error -- it considers the fact that we've swapped the order of the two variants to be unsafe:

    test.pyi:10: error: Signature of "foo" incompatible with supertype "Parent"

This error message can be confusing for some users who may not be aware that the order in which your overloads are defined can sometimes matter/is something mypy relies on.

This commit modifies the error message to the following to try
and make this more clear:

    test.pyi:10: error: Signature of "foo" incompatible with supertype "Parent"
    test.pyi:10: note: Overload variants must be defined in the same order as they are in "Parent"

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1349,6 +1349,25 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                     self.msg.return_type_incompatible_with_supertype(
                         name, name_in_super, supertype, node)
                     emitted_msg = True
+            elif isinstance(override, Overloaded) and isinstance(original, Overloaded):
+                # Give a more detailed message in the case where the user is trying to
+                # override an overload, and the subclass's overload is plausible, except
+                # that the order of the variants are wrong.
+                #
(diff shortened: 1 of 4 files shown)


Michael0x2a:
As a misc note, one idea the issue I linked to above mentioned was that we could maybe print out the signature of the parent method. I decided not to do this for now because it felt inconsistent: we don't do this in other cases where the user incorrectly overrides a method. 

If printing out the parent signature seems like a helpful thing to do in general, I'm happy to make a new PR w/ that change.

PR #5370, opened 2018-07-17
Enable --no-silence-site-packages in eval tests

Michael0x2a:
An unintended side-effect of silencing errors in site packages and typeshed by default is that our tests no longer flag changes to mypy that end up breaking typeshed in some way.

(For example, see https://github.com/python/mypy/pulls/5280 which *really* should not be passing, at least as of time of writing.)

This pull request will add the `--no-silence-site-packages` flag to any tests suites that directly or indirectly test typeshed.

Specifically, I believe this ought to add the flag to any tests triggered by testsamples.py, testselfcheck.py, and testpythoneval.py.


JelleZijlstra:
Looks like this is exposing an actual typeshed problem already: `C:\Python36-x64\Lib\mypy\typeshed\stdlib\2and3\shutil.pyi:135: error: Name 'os.terminal_size' is not defined`.

PR #5371, opened 2018-07-18
Create mee

meet-sheth:
do this for me take this one

### DIFF
--- a/mee
+++ b/mee
@@ -1,0 +1,1 @@
+dfkgvnl


Michael0x2a:
Hi -- I'm assuming you're trying to practice using git or Github? 

If so, I recommend you try practicing on a repo that only you own and control.

Whenever you send a PR to a project, you're asking the maintainers of that project to spend some of their time reviewing whatever you submitted. If your proposed change doesn't help improve the project in some way, it's better to just not make it to save everybody some time. 

(Of course, you can make any changes you want to repos you own, which is why it's better to practice there.)

If you *do* want to contribute to mypy, we have [a few starter tasks](https://github.com/python/mypy/labels/pycon-sprints) you could try tackling.

Issue #5383, opened 2018-07-23
Mypy does not recognize aliases of dataclasses.dataclass


wkschwartz:
In Python 3.7.0 and mypy `0.630+dev-3fb16a2b8c5439074e39b75feebc109a439eede3`, if I save the following code to `mypy-test.py`

```python
from dataclasses import dataclass

dataclass_alias = dataclass
dataclass_alias2 = dataclass(order=False)

def dataclass_wrapper(cls):
	return dataclass(cls)

@dataclass
class A: arg1: int

@dataclass_alias
class B: arg1: int

@dataclass_alias2
class C: arg1: int

@dataclass_wrapper
class D: arg1: int

a = A(arg1=1)
b = B(arg1=1)
c = C(arg1=1)
d = D(arg1=1)
```

I get the following mypy output

```
$ mypy mypy-test.py 
mypy-test.py:22: error: Unexpected keyword argument "arg1" for "B"
mypy-test.py:23: error: Unexpected keyword argument "arg1" for "C"
mypy-test.py:24: error: Unexpected keyword argument "arg1" for "D"
```

Notice that mypy has no trouble recognizing the signature of `A`'s constructor. I can understand how mypy would have a hard time with `dataclass_wrapper`, and `dataclass_alias` is probably not critical in terms of use case. However, `dataclass_alias2` is important if you want (as I do) to define many data classes with the same options, especially if you want to make these options available as part of your module's public interface.

JukkaL:
It's unlikely that mypy will support `dataclass_wrapper`, but `dataclass_alias` and `dataclass_alias2` are something that mypy could support. This is low priority since adding support for these use cases would require some pretty non-trivial new machinery. Also, `dataclasses` is still new and we don't know how common this use case is. We can increase the priority if many users encounter this limitation.

wkschwartz:
In the mean time, perhaps a single sentence added to the documentation might be helpful.

gvanrossum:
@wkschwartz 
> In the mean time, perhaps a single sentence added to the documentation might be helpful.

That makes sense. Since you know where you (would have) looked in the docs, can you help us by sending a Pull Request?

edaqa-uncountable:
Support for the `dataclass_wrapper` case would be helpful.  I was trying to create a wrapped class that adds API information to my dataclasses. 

The workaround I've resorted too is double-wrapping:

```
def my_class(desc: Optional[str] = None):
    def decorate(orig_class):
        orig_class.__unc_data = {
[...]

PR #5387, opened 2018-07-24
Refs #5383: Document that aliasing dataclass fails

wkschwartz:
This documentation is a temporary solution. It should be altered or
removed when #5383 is fixed.

I inserted the whole example from #5383. If you think that's too much, feel free to modify. I've granted edit permissions to maintainers.

### DIFF
--- a/docs/source/additional_features.rst
+++ b/docs/source/additional_features.rst
@@ -83,22 +83,27 @@ do **not** work:
     from dataclasses import dataclass
 
     dataclass_alias = dataclass
-    dataclass_alias2 = dataclass(order=False)
     def dataclass_wrapper(cls):
       return dataclass(cls)
 
     @dataclass_alias
-    class BadlyDecoratedClass1: arg1: int
-
-    @dataclass_alias2
-    class BadlyDecoratedClass2: arg1: int
+    class AliasDecorated:
+      """
+      Mypy doesn't recognize this as a dataclass because it is decorated by an
+      alias of `dataclass` rather than by `dataclass` itself.
+      """
+      arg1: int
 
     @dataclass_wrapper
-    class BadlyDecoratedClass3: arg1: int
+    class DynamicallyDecoarted:
+      """
+      Mypy doesn't recognize this as a dataclass because it is decorated by a
+      function returning `dataclass` rather than by `dataclass` itself.
+      """
+      arg1: int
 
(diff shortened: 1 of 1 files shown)


docs/source/additional_features.rst
```diff
@@ -68,9 +68,38 @@ class can be used:
...
+
+    from dataclasses import dataclass
+
+    dataclass_alias = dataclass
+    dataclass_alias2 = dataclass(order=False)
+    def dataclass_wrapper(cls):
+      return dataclass(cls)
+
+    @dataclass_alias
+    class BadlyDecoratedClass1: arg1: int
```
JelleZijlstra: I didn't realize this syntax was legal. I think it's better to put the field on a separate line.

gvanrossum:
@wkschwartz Please respond to the review feedback. (Please use a new local commit, or several, to update your changes -- don't use `git commit --amend; git push -f`.)

docs/source/additional_features.rst
```diff
@@ -68,9 +68,38 @@ class can be used:
...
-**Note:** Some functions in the ``dataclasses`` module, such as ``replace()`` and ``asdict()``,
+Caveats/Known Issues
+====================
+
+Some functions in the ``dataclasses`` module, such as ``replace()`` and ``asdict()``,
 have imprecise (too permissive) types. This will be fixed in future releases.
 
+Mypy does not yet recognize aliases of ``dataclasses.dataclass``, and will
+probably never recognize dynamically computed decorators. The following examples
+do **not** work:
```
ilevkivskyi: I would actually skip the examples, the textual formulation is already clear (except maybe the dynamically computed decorators).

wkschwartz:
@ilevkivskyi I removed one of the three examples in response to https://github.com/python/mypy/pull/5387#discussion_r206927049 in dc742c8

docs/source/additional_features.rst
```diff
@@ -68,9 +68,43 @@ class can be used:
...
+    def dataclass_wrapper(cls):
+      return dataclass(cls)
+
+    @dataclass_alias
+    class AliasDecorated:
+      """
+      Mypy doesn't recognize this as a dataclass because it is decorated by an
+      alias of `dataclass` rather than by `dataclass` itself.
+      """
+      arg1: int
```
ilevkivskyi: Could you please use `attr` (or similar) instead of `arg1`?

PR #5388, opened 2018-07-25
Fix IntEnum iterability

### LINKED ISSUE #5366: `IntEnum` iterator loses type when passed through `enumerate` or `zip`
I've encountered a typing issue with `IntEnum`. Iterating over the `IntEnum` class works fine until you pass it through a generic function, e.g. `enumerate` or `zip`. Mypy then requests a type annotation, but when I provide it, it fails later with `<nothing>`. This problem does not occur with `Enum`, only `IntEnum`.

```
from enum import Enum, IntEnum
from typing import List

class E(IntEnum):
  A = 1

for e0 in E: print(e0) # Works fine.

#e1:E
for i, e1 in enumerate(E): print(i, e1)
```

```
error: Need type annotation for 'e1'
error: Cannot determine type of 'e1'
```

When the type annotation for e1 is uncommented, I then get `error: <nothing> has no attribute "name"`.

setup.cfg:
```
[mypy]
python_version = 3.6
cache_dir = _build/mypy_cache
mypy_path = ~/work/pithy:./lambda
check_untyped_defs = True
disallow_untyped_calls = True
disallow_untyped_defs = False
ignore_missing_imports = False
show_column_numbers = True
show_none_errors = True
strict_boolean = False
strict_optional = True
warn_incomplete_stub = True
warn_no_return = True
warn_redundant_casts = True
warn_return_any = True
warn_unused_configs = True
warn_unused_ignores = True
```

elazarg:
Fix #5366

### DIFF
--- a/mypy/constraints.py
+++ b/mypy/constraints.py
@@ -308,6 +308,8 @@ class ConstraintBuilderVisitor(TypeVisitor[List[Constraint]]):
         res = []  # type: List[Constraint]
         if isinstance(actual, CallableType) and actual.fallback is not None:
             actual = actual.fallback
+        if isinstance(actual, Overloaded) and actual.fallback is not None:
+            actual = actual.fallback
         if isinstance(actual, TypedDictType):
             actual = actual.as_anonymous().fallback
         if isinstance(actual, Instance):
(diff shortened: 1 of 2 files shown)


gvanrossum:
Thanks! Short and sweet. :-)

PR #5396, opened 2018-07-27
used age instead of "a" for ensure proper operation of example.

BTaskaya:
used age instead of a for ensure proper operation of example.


docs/source/cheat_sheet_py3.rst
```diff
@@ -25,15 +25,15 @@ we use it in most examples.
...
 
    # In Python 3.5 and earlier you can use a type comment instead
    # (equivalent to the previous definition)
-   x = 1  # type: int
+   age = 1  # type: int
 
    # You don't need to initialize a variable to annotate it
    a: int  # Ok (no value at runtime until assigned)
-
+
```
gvanrossum: Looks like you accidentally added trailing whitespace to this line. Can you clear that up?

BTaskaya: sure

PR #5397, opened 2018-07-27
 used age instead of "a" for ensure proper operation of example.

BTaskaya:
used age instead of "a" for ensure proper operation of example.

CC: @gvanrossum


JelleZijlstra:
I'm not sure what the point of this change is. Single-character variable names seem fine as examples in this context.

gvanrossum:
> I'm not sure what the point of this change is.

Presumably so that the example further down in the same block works? (`if age < 18:`).

gvanrossum:
Thanks! All set.

PR #5399, opened 2018-07-27
Update versions for pytest and xdist

gvanrossum:
This should fix Travis. Travis is currently failing like this:
```
pkg_resources.VersionConflict: (pytest 3.3.0 (/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages), Requirement.parse('pytest>=3.4'))
```

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,3 +1,4 @@
+attrs>=18.0
 flake8
 flake8-bugbear; python_version >= '3.5'
 flake8-pyi; python_version >= '3.6'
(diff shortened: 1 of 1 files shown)


gvanrossum:
Well, now the 3.6 tests (but those only) fail with 
```
    _marks = attr.ib(converter=list)
TypeError: attrib() got an unexpected keyword argument 'converter'
```
I'll explicitly add `attrs>= 18.0`, maybe that'll fix it.

I'm down to programming by random modifications here, alas. :-(

JelleZijlstra:
Hey, it's not entirely random. :) Sounds like pytest is missing an exact version requirement for attrs.

gvanrossum:
Well, the install logs show that it was using version 17.x and I don't think that `attr.ib(converter=...)` was introduced that recently. But we'll see.

JelleZijlstra:
According to http://www.attrs.org/en/stable/api.html it was added in 17.4.0.

PR #5411, opened 2018-08-01
[Attrs] Add support for attr.converters.optional

euresti:
We hardcode this because it's way too complicated to handle this using the generic type system.  Though I'm open to other ideas.

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -100,11 +100,11 @@ class Attribute:
                 if types:
                     args = UnionType.make_simplified_union(types)
                     init_type = ctx.api.anal_type(args)
+
             if self.converter.is_attr_converters_optional and init_type:
                 # If the converter was attr.converter.optional(type) then add None to
                 # the allowed init_type.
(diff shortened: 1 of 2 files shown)


mypy/plugins/attrs.py
```diff
@@ -90,10 +100,16 @@ def argument(self, ctx: 'mypy.plugin.ClassDefContext') -> Argument:
                 if types:
                     args = UnionType.make_simplified_union(types)
                     init_type = ctx.api.anal_type(args)
+            if self.converter.is_attr_converters_optional and init_type:
+                # If the converter was attr.converter.optional(type) then add None to
+                # the allowed init_type.
+                args = UnionType.make_simplified_union([init_type, NoneTyp()])
+                init_type = ctx.api.anal_type(args)
```
ilevkivskyi: Why do you need the `anal_type` call here?

euresti: Oh I guess by this point the type has been analyzed already.  I'll remove it.

PR #5417, opened 2018-08-02
CI upgrades

gaborbernat:
- Travis test with python3.7 and python3.8-dev
- Travis use tox to setup/run tests
- Bump tox default envs from python3.6 to python3.7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,37 @@
-sudo: false
+if: tag IS present OR type = pull_request OR (branch = master AND type = push)   # we only CI the master, tags and PRs
+
 language: python
 # cache package wheels (1 cache per python version)
 cache: pip
-python:
-  - "3.4"
+dist: xenial
+sudo: required
+
+env:
+  TOXENV=py
+  EXTRA_ARGS=-n 12
+
+jobs:
+  include:
+  - python: 3.4
+    dist: trusty
   # Specifically request 3.5.1 because we need to be compatible with that.
-  - "3.5.1"
-  - "3.6"
-  - "3.7-dev"
+  - python: 3.5.1
+    dist: trusty
+  - python: 3.6    # 3.6.3  pip  9.0.1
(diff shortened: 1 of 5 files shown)


gvanrossum:
This looks great!

.travis.yml
```diff
@@ -1,21 +1,38 @@
...
   # Specifically request 3.5.1 because we need to be compatible with that.
-  - "3.5.1"
-  - "3.6"
-  - "3.7-dev"
+  - python: 3.5.1
+    dist: trusty
+  - python: 3.6    # 3.6.3  pip  9.0.1
+  - python: 3.7    # 3.7.0  pip 10.0.1
+  - python: 3.8-dev
+  - python: 3.7
```
ethanhs: Might you be able to add names to the lint and self check runs? (see https://docs.travis-ci.com/user/customizing-the-build/#naming-jobs-within-matrices). Having 3 "3.7" environments could be a bit confusing :)

.travis.yml
```diff
@@ -1,21 +1,38 @@
-sudo: false
+if: tag IS present OR type = pull_request OR (branch = master AND type = push)   # we only CI the master, tags and PRs
+
 language: python
 # cache package wheels (1 cache per python version)
 cache: pip
-python:
-  - "3.4"
+dist: xenial
```
ethanhs: While *I* know why xenial is needed for some of these runs, it would be nice to leave a short comment for the future.

gaborbernat:
Done requested changes.

gaborbernat:
But somehow it downloads a wheel on other PRS (^^)

ethanhs:
Sorry let me correct myself. lxml doesn't have wheels for *3.7* on Windows yet. It works otherwise because we are targeting 3.6 currently.

gaborbernat:
actually turns out the latest minor has it so bumped to that

gaborbernat:
@ethanhs now should be all ok :+1:

Issue #3645, opened 2017-07-01
Problems using class type variable in class methods


LiraNuna:
Follow up from a discussion on gitter:

```python
from typing import *


T = TypeVar('T')
class Factory(Generic[T]):
    @classmethod
    def instance(cls) -> 'Factory[T]':
        return cls()

    def produce(self) -> T:
        raise NotImplementedError()

    @classmethod
    def get(cls) -> T:
        return cls.instance().produce()


class HelloWorldFactory(Factory[str]):
    def produce(self) -> str:
        return 'Hello World'


# factory.py:25: error: Incompatible return value type (got "T", expected "str")
def error_with_mypy() -> str:
    return HelloWorldFactory.get()
```

@ilevkivskyi mentioned this should not error and that this code is valid. I don't have the necessary context to understand how to classify this issue.

ilevkivskyi:
I would simplify the example to show only the essential part:
```python
class Factory(Generic[T]):
    def produce(self) -> T:
        ...
    @classmethod
    def get(cls) -> T:
        return cls().produce()

class HelloWorldFactory(Factory[str]):
    def produce(self) -> str:
        return 'Hello World'

reveal_type(HelloWorldFactory.get())  # mypy should be able to infer 'str' here
```
On one hand, type variables are bound on instances, not on classes, but in case of a non-generic subclass I think the type variable can be bound on that class. The problem however is that we need to give a clear error message when this is not possible, something like (using above example):
```python
Factory.get()  # E: cannot bind type variable on class (use an instance or a non-generic subclass)
Facrory[int].get()   # E: cannot bind type variable on class (use an instance or a non-generic subclass)
```

JukkaL:
Similar issues have come up previously, so this would perhaps be a reasonable feature to have eventually -- at least assuming the implementation isn't very complex.

gvanrossum:
The `produce()` method is also irrelevant to the example. Shorter yet self-contained:
```py
from typing import Generic, TypeVar
T = TypeVar('T')
class Factory(Generic[T]):
    @classmethod
    def get(cls) -> T: ...
class HelloWorldFactory(Factory[str]): pass
reveal_type(HelloWorldFactory.get())  # mypy should be able to infer 'str' here
```

rowillia:
@JukkaL Do you have any thoughts on what it would take to fix this?  It's blocking us here at Lyft.

JukkaL:
@rowillia My recommendation is to use a regular method instead of a class method. Example:

[...]

Issue #1337, opened 2016-04-06
`@classmethod` of `Generic` class


tharvik:
``` python
from typing import TypeVar, Generic

T = TypeVar('T')


class A(Generic[T]):
    @classmethod
    def func(cls) -> T:
        pass


class B(A['B']):
    pass


def except_b(b: B):
    pass


except_b(B.func())
```

is a valid python3 code, but mypy warns that

```
file:20: error: Argument 1 to "except_b" has incompatible type "T"; expected "B"
```

but the resolved pseudo-code of class B should be

``` python
class B(A['B']):
    @classmethod
    def func(cls) -> 'B':
        pass
```

if I replace it, mypy doesn't complain anymore.

I'm might have overlook some specificities of `@classmethod` but I think that `func` is also bind to `class B` thus should be type checking.

JukkaL:
Actually it seems that the signature of `A.func` should be rejected, since `T` is not defined at class/static method level, since the type variable is specific to an instance, and we have no `self`. Type variables should not be in scope for static or class methods. I'm going to move this to 0.4 since this seems like a pretty rare edge case.

rwbarton:
It also seems consistent to turn the class typevars into function typevars for a static method call, like we do for constructors. We could give `A.func` the type `def [T] () -> T`, and `B.func` the type `def () -> B`. Hard to say whether that is useful without some realistic examples.

gvanrossum:
That would seem useful for constructors. I can't think of any off the bat
but I betcha there are some classes that take AnyStr and have a handy
alternate constructor that's a class or static method.

rwbarton:
This is also an issue for class attributes:
```
class A(Generic[T]):
    a = []  # type: List[T]
class B(A[int]): pass
```
What is the type of `B.a`? For that matter what is the type of `A.a`?

This issue is actually more severe, because on the one hand we cannot give `A.a` any sort of generic type (only functions can be generic in mypy currently, and besides it would not be type safe to give `A.a` a type like `forall T. List[T]`, as it is a mutable variable); and on the other hand the style of declaring class variables in order to specify the types of instance variables is used frequently, at least in mypy itself, and if we want to do the same in generic classes, we need to be able to make declarations like that of `A`.

gvanrossum:
Yeah, but mypy's handling of class attributes is just really poor, and the
meaning of that example isn't clear, because we do type erasure -- at
runtime there's only one A and it has only one a.

[...]

PR #5420, opened 2018-08-05
Specify type when type variables are remained in inherited classmethod.

kitsuyui:
Fix: #3645

~- Fix: 1337, 3645~

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -559,7 +559,19 @@ def analyze_class_attribute_access(itype: Instance,
             msg.fail(messages.GENERIC_INSTANCE_VAR_CLASS_ACCESS, context)
         is_classmethod = ((is_decorated and cast(Decorator, node.node).func.is_class)
                           or (isinstance(node.node, FuncBase) and node.node.is_class))
+        if is_classmethod and not isinstance(original_type, TypeType) and itype.args:
+            msg.fail(messages.GENERIC_CLASSMETHOD_CLASS_ACCESS, context)
+
         result = add_class_tvars(t, itype, is_classmethod, builtin_type, original_type)
+
+        method = itype.type.get_classmethod(name)
+        if method is not None:
+            signature = function_type(method, builtin_type('builtins.function'))
+            signature = freshen_function_type_vars(signature)
+            signature = bind_self(signature, original_type, is_classmethod=True)
+            itype = map_instance_to_supertype(itype, method.info)
+            result = expand_type_by_instance(signature, itype)
+
         if not is_lvalue:
             result = analyze_descriptor_access(original_type, result, builtin_type,
                                                msg, context, chk=chk)
--- a/test-data/unit/check-generics.test
+++ b/test-data/unit/check-generics.test
@@ -1815,3 +1815,269 @@ def g(x: T) -> T: return x
 [out]
 main:3: error: Revealed type is 'def [b.T] (x: b.T`-1) -> b.T`-1'
 main:4: error: Revealed type is 'def [T] (x: T`-1) -> T`-1'
+
+[case testClassMethodOfGenericClass]
+from typing import TypeVar, Generic
+T = TypeVar('T')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> T: ...
+class B(A[str]): pass
+class C(A[int]): pass
+
+reveal_type(B.f)  # E: Revealed type is 'def () -> builtins.str*'
+reveal_type(B.f())  # E: Revealed type is 'builtins.str*'
+reveal_type(C.f)  # E: Revealed type is 'def () -> builtins.int*'
+reveal_type(C.f())  # E: Revealed type is 'builtins.int*'
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOfGenericClassMultipleInheritance]
+from typing import TypeVar, Generic
+T = TypeVar('T')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> T: pass
+
+class B(Generic[T]):
+    @classmethod
+    def g(cls) -> T: pass
+
+class C(A[str], B[int]):
+    @classmethod
+    def f(cls) -> str: pass
+    @classmethod
+    def g(cls) -> int: pass
+
+reveal_type(C.f())  # E: Revealed type is 'builtins.str'
+reveal_type(C.g())  # E: Revealed type is 'builtins.int'
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOfSubcClassOfSubClassOfGenericClass]
+from typing import TypeVar, Generic
+T = TypeVar('T')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> T: ...
+
+class B(A[str]): pass
+class C(B): pass
+class D(C): pass
+
+reveal_type(B.f)  # E: Revealed type is 'def () -> builtins.str*'
+reveal_type(B.f())  # E: Revealed type is 'builtins.str*'
+reveal_type(C.f)  # E: Revealed type is 'def () -> builtins.str*'
+reveal_type(C.f())  # E: Revealed type is 'builtins.str*'
+reveal_type(D.f)  # E: Revealed type is 'def () -> builtins.str*'
+reveal_type(D.f())  # E: Revealed type is 'builtins.str*'
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOfGenericClassMultipleLevelInheritance]
+from typing import TypeVar, Generic
+T = TypeVar('T')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> T: ...
+
+class B(A[str]): pass
+class C(B): pass
+class D(C): pass
+class E(D): pass
+
+reveal_type(C.f)  # E: Revealed type is 'def () -> builtins.str*'
+reveal_type(C.f())  # E: Revealed type is 'builtins.str*'
+reveal_type(D.f)  # E: Revealed type is 'def () -> builtins.str*'
+reveal_type(D.f())  # E: Revealed type is 'builtins.str*'
+reveal_type(E.f)  # E: Revealed type is 'def () -> builtins.str*'
+reveal_type(E.f())  # E: Revealed type is 'builtins.str*'
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOfGenericClassMultipleLevelInheritanceWithChangingTypeVariable]
+from typing import TypeVar, Generic
+T = TypeVar('T')
+S = TypeVar('S')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> T: ...
+class B(A[S]): pass
+class C(B[str]): pass
+class D(B[int]): pass
+
+reveal_type(C.f())  # E: Revealed type is 'builtins.str*'
+reveal_type(D.f())  # E: Revealed type is 'builtins.int*'
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOrderOfMultipleGenericClass]
+from typing import TypeVar, Generic
+T = TypeVar('T')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> T: ...
+class B(Generic[T]):
+    @classmethod
+    def f(cls) -> T: ...
+class C(A[str], B[int]): pass
+class D(A[int], B[str]): pass
+
+reveal_type(C.f)  # E: Revealed type is 'def () -> builtins.str*'
+reveal_type(C.f())  # E: Revealed type is 'builtins.str*'
+reveal_type(D.f)  # E: Revealed type is 'def () -> builtins.int*'
+reveal_type(D.f())  # E: Revealed type is 'builtins.int*'
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOfMultipleGenericClass]
+from typing import Tuple, TypeVar, Generic
+T = TypeVar('T')
+S = TypeVar('S')
+class A(Generic[T]): pass
+class B(Generic[T]): pass
+class C(A[T], B[S]):
+    @classmethod
+    def h(cls) -> Tuple[T, S]: ...
+
+class D(C[str, int]): pass
+class E(C[int, str]): pass
+
+reveal_type(D.h())  # E: Revealed type is 'Tuple[builtins.str*, builtins.int*]'
+reveal_type(E.h())  # E: Revealed type is 'Tuple[builtins.int*, builtins.str*]'
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOfGenericClassOverriding]
+from typing import Generic, TypeVar
+T = TypeVar('T')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> T: pass
+class B(A[str]):
+    @classmethod
+    def f(cls) -> str: pass
+class C(A[str]):
+    @classmethod
+    def f(cls) -> int: pass  # E: Return type of "f" incompatible with supertype "A"
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOfInheritingGenericTypeFromGenericType]
+from typing import Generic, TypeVar
+T = TypeVar('T')
+S = TypeVar('S')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> T: pass
+class B(A[S]): pass
+
+A.f  # E: Access to generic classmethods via class is ambiguous.
+B.f  # E: Access to generic classmethods via class is ambiguous.
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOfGenericClassComplexType]
+from typing import Generic, TypeVar, Iterable, Tuple
+T = TypeVar('T')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> T: pass
+
+class B(A[Iterable[str]]): pass
+
+reveal_type(B.f())  # E: Revealed type is 'typing.Iterable*[builtins.str]'
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOfGenericClassComplexReturnType]
+from typing import Generic, TypeVar, Iterable, Tuple
+T = TypeVar('T')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> Iterable[Tuple[T, T]]: pass
+
+class B(A[str]): pass
+
+reveal_type(B.f())  # E: Revealed type is 'typing.Iterable[Tuple[builtins.str*, builtins.str*]]'
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodOfGenericClassComplexArgumentType]
+from typing import Generic, TypeVar, Iterable, Tuple
+T = TypeVar('T')
+class A(Generic[T]):
+    @classmethod
+    def f(cls, arg: Iterable[Tuple[T, T]]) -> T: pass
+
+class B(A[str]): pass
+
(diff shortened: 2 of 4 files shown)


test-data/unit/typexport-basic.test
```diff
@@ -997,6 +997,18 @@ A.f(ab, o)
 [out]
 MemberExpr(10) : def (self: A[B], y: builtins.object)
 
+[case testClassMethodOfGenericClass]
```
ilevkivskyi: Why are you using a `typexport` test instead of "normal" `check-*.test` tests?

mypy/checkmember.py
```diff
@@ -474,6 +474,9 @@ def analyze_class_attribute_access(itype: Instance,
             msg.fail(messages.GENERIC_INSTANCE_VAR_CLASS_ACCESS, context)
         is_classmethod = ((is_decorated and cast(Decorator, node.node).func.is_class)
                           or (isinstance(node.node, FuncBase) and node.node.is_class))
+        if is_classmethod and not itype.type.tuple_type:
```
ilevkivskyi: Why tuple types are special here? This exception deserves a comment at least.

mypy/checkmember.py
```diff
@@ -474,6 +474,9 @@ def analyze_class_attribute_access(itype: Instance,
...
+            for base_itype in itype.type.bases:
+                t = expand_type_by_instance(t, base_itype)
```
ilevkivskyi: This is not what I would do bind all type variables. For example, it seems to me an extra subclassing will easily break this:
```python
class A(Generic[T]):
    @classmethod
    def m(cls) -> T: ...

class B(A[int]): ...
class C(B): ...
C.m()  # does this work? What about other cases of multiple inheritance?
```

Normally, (i.e. for binding type variables by instances) we use a combination `map_instance_to_supertype` and `expand_type_by_instance`. For this case I would write a similar more universal algorithm.

kitsuyui:
@ilevkivskyi 
Thank you for reviewing my codes.
I will try that you pointed!

test-data/unit/typexport-basic.test
kitsuyui: I have not consciously thought. Sorry. I added it to `typexport-basic.test` because it has some generics tests already.

Is it better if I create new file like `test-data/unit/check-classmethods.test` ?
Or is there some better place?

kitsuyui: I changed my mind. I will put it to `check-generics.test`. It's more narrow than classes.

kitsuyui:
@ilevkivskyi 

> In #1337 there is also a question what to do with class attributes (in addition to class methods), you should address this. Namely,

```python
class C(Generic[T]):
    x: T
class D(C[int]): ...
D.x  # should not error and result in 'builtins.int'
```

I think this sample should be like this

```python
class C(Generic[T]):
    x: List[T] = []
class D(C[int]): pass
D.x # should not error and result in 'builtins.int' 
```

Because `x: T` become type hint for instance attribute. 
So I will test `x: List[T] = []` as class attribute.

kitsuyui:
Sorry, I've tried and I understood it will not be solved for class attributes.

Class attributes are shared with super-class. So subclasses cannot be restricted to narrow type.

```python
from typing import Generic, TypeVar, List

T = TypeVar('T')
class A(Generic[T]):
    x: List[T] = []
class B(A[int]): pass
class C(A[str]): pass
B.x.append(1)
C.x.append('1')
print(B.x)  # => [1, '1']
print(C.x)  # => [1, '1']
```

ilevkivskyi:
Sorry for a delay. I will try to review this PR next week. In the meantime could you please fix the merge conflict?

kitsuyui:
@ilevkivskyi I've updated!

ilevkivskyi:
Hi @kitsuyui! Sorry for a delay with review, are you still interested in working on this?

kitsuyui:
@ilevkivskyi 
Yeah, I'd love to. 
I will challenge updating that it works even in case of #5664 if needed.

test-data/unit/check-generics.test
```diff
@@ -1814,3 +1814,98 @@ def g(x: T) -> T: return x
...
+    @classmethod
+    def f(cls) -> T: ...
+class B(A[str]): pass
+class C(A[int]): pass
+reveal_type(B.f)  # E: Revealed type is 'def () -> builtins.str*'
+reveal_type(B.f())  # E: Revealed type is 'builtins.str*'
+reveal_type(C.f)  # E: Revealed type is 'def () -> builtins.int*'
+reveal_type(C.f())  # E: Revealed type is 'builtins.int*'
+[builtins fixtures/classmethod.pyi]
+[out]
```
ilevkivskyi: The empty `[out]` sections can be left out.

test-data/unit/check-generics.test
```diff
@@ -1814,3 +1814,98 @@ def g(x: T) -> T: return x
...
+class A(Generic[t]):
+    @classmethod
+    def f(cls) -> t: ...
+class B(A[str]): pass
+class C(A[int]): pass
+reveal_type(B.f)  # E: Revealed type is 'def () -> builtins.str*'
+reveal_type(B.f())  # E: Revealed type is 'builtins.str*'
+reveal_type(C.f)  # E: Revealed type is 'def () -> builtins.int*'
+reveal_type(C.f())  # E: Revealed type is 'builtins.int*'
+[builtins fixtures/classmethod.pyi]
```
ilevkivskyi: What is the difference from the tests above here? The name suggests there should be multiple inheritance involved but it is not the case here.

Also we prefer capital letters for type variable names.

test-data/unit/check-generics.test
```diff
@@ -1814,3 +1814,98 @@ def g(x: T) -> T: return x
...
+[case testClassMethodOfGenericClassWithBindingAnotherTypeVariable]
+from typing import Generic, TypeVar
+t = TypeVar('t')
+s = TypeVar('s')
+class A(Generic[t]):
+    @classmethod
+    def f(cls) -> t: pass
+class B(A[s]): pass
+reveal_type(A.f())  # E: Revealed type is '<nothing>'
+reveal_type(B.f())  # E: Revealed type is '<nothing>'
```
ilevkivskyi: What is this supposed to test? I would actually give an error if any variable stays unbound during the access via class object.

test-data/unit/check-generics.test
```diff
@@ -1814,3 +1814,98 @@ def g(x: T) -> T: return x
...
+[builtins fixtures/classmethod.pyi]
+[out]
```
ilevkivskyi: Here suggestions for more tests:
* Overloaded class methods
* More complex bindings, like `T -> Iterable[T] -> Iterable[str]`, etc.
* More complex return types (like `Iterable[Tuple[T, T]]` etc.) and argument types
* Basic multiple inheritance
* Overriding between bases in multiple inheritance

In addition, as I mentioned above:
* Give an error if class method accessed via _class object_ contains type variables after everything is bound. Similar to what we do for variables
* Add tests for binding types of _variables_ and normal methods accessed on class object.

mypy/expandtype.py
```diff
@@ -29,6 +29,17 @@ def expand_type_by_instance(typ: Type, instance: Instance) -> Type:
...
 
+def expand_type_by_instance_bases(typ: Type, instance: Instance) -> Type:
+    if instance.type.tuple_type:
+        # Generic tuple types not supported
+        return typ
+    for base_instance in instance.type.bases:
+        if base_instance.type.bases:
+            typ = expand_type_by_instance_bases(typ, base_instance)
+        typ = expand_type_by_instance(typ, base_instance)
+    return typ
```
ilevkivskyi: This algorithm currently looks wrong, and this is why #5664 and similar more complex type variable bindings don't work. You need to map type variables correctly, look for example in `maptype.py`

mypy/checkmember.py
```diff
@@ -548,6 +551,8 @@ def analyze_class_attribute_access(itype: Instance,
             msg.fail(messages.GENERIC_INSTANCE_VAR_CLASS_ACCESS, context)
         is_classmethod = ((is_decorated and cast(Decorator, node.node).func.is_class)
                           or (isinstance(node.node, FuncBase) and node.node.is_class))
+        if is_classmethod:
+            t = expand_type_by_instance_bases(t, itype)
```
ilevkivskyi: I think binding should also happen for variables and normal methods. As you noticed, accessing variables may be bot safe for mutable containers, so it indeed makes sense to give an error, but the inferred type should not contain type variables, they should be either bound to actual values, or replaced with `Any`.

test-data/unit/check-generics.test
kitsuyui: I supposed I had wanted to check it works correctly when both replacing type variable T to S, and inheriting A from B.

But I didn't know clearly that how to represent. So I added like this ambiguously.

It may what I really wanted to represent like this:

```python
from typing import Generic, TypeVar
T = TypeVar('T')
S = TypeVar('S')
class A(Generic[T]):
    @classmethod
    def f(cls) -> T: pass
class B(A[S]): pass
reveal_type(A.f)  # E: Revealed type is 'def [T] () -> T`1'
reveal_type(B.f)  # E: Revealed type is 'def [S] () -> S`1'
```

ilevkivskyi: Yes, this test makes more sense.

oraluben:
I've tried the latest code and found something that may be added to the test cases, here's the code.
```
from typing import Generic, Callable, TypeVar, Iterator, Type

T = TypeVar('T')
BoxT = TypeVar('BoxT', bound='Box', covariant=True)


class Box(Generic[T]):
    @classmethod
    def wrap(cls, generator: Callable[[], T]) -> 'Box[T]':
        pass


class IteratorBox(Box[Iterator[T]],
                  Generic[T]):
    @classmethod
    def wrap_iter(cls: Type[BoxT],
                  typ: Type[Box[T]]) -> Callable[[Callable[[], Iterator[T]]], BoxT]:
        def inner(generator: Callable[[], Iterator[T]]) -> BoxT:
            pass

        return inner


class Str(Box[str]):
    pass


class StrI(IteratorBox[str]):
    pass


@StrI.wrap_iter(Str)
def g() -> Iterator[str]:
    pass
```
I believe there's no type error in the signature, but the output was
```
test.py:32: error: Argument 1 has incompatible type "Callable[[], Iterator[str]]"; expected "Callable[[], Iterator[Iterator[str]]]"

test.py:32: error: Argument 1 to "wrap_iter" of "IteratorBox" has incompatible type "Type[Str]"; expected "Type[Box[Iterator[str]]]
```
Hope this will help!

kitsuyui:
@oraluben 

Your code works on my local mypy! But I still have not pushed commits yet.
The biggest uncompleting task is about handling error and replacing unbound variable to `Any` with consistency.

kitsuyui:
I have pushed commits that rebased into latest master.

But this is still insufficient yet.

I have been still in the way of this part:  https://github.com/python/mypy/pull/5420#discussion_r222260417
So #5664 is not solved too.

It might be little more time to solve.

ilevkivskyi:
@kitsuyui what is exactly the problem you have? I might be able to give some hints.

kitsuyui:
@ilevkivskyi 

I don't know what is the true behavior on following case;

```python
from typing import Generic, TypeVar
T = TypeVar('T')
class A(Generic[T]):
    @classmethod
    def f(cls, value: T) -> T: pass
reveal_type(A.f(1))  # : Revealed type is 'builtins.int*'
```

T is solved by arguments. But T is not clear on class.
And #5664 seems to happen because of this behavior.

function and type unbounded classmethod's should be checked by different algorithm.
I think it is needed to be fix checkmember.py.

But this fix may be too large. So I'm wondering "doesn't exist Plan B?"

kitsuyui:
I've finally found the answer myself.
I added error message of when accessing classmethod of generic class contains type variables.

```python
from typing import Generic, TypeVar
T = TypeVar('T')
S = TypeVar('S')
class A(Generic[T]):
    @classmethod
    def f(cls) -> T: pass
class B(A[S]): pass
A.f  # E: Access to classmethod of generic type remains type variables T is ambiguous.
B.f  # E: Access to classmethod of generic type remains type variables S is ambiguous.
````

@ilevkivskyi 

Would you please review this once more?

**Comment by ilevkivskyi** [comment]
Sorry for another delay. I will try to review this in the next few days.

**Comment by kitsuyui** [comment]
@ilevkivskyi

Thank you.
And I've just found myself that my branch has false negative like following case;

```python
from typing import Generic, TypeVar


T = TypeVar('T')


class X(Generic[T]):
    @classmethod
    def a(cls) -> T:
        ...

    @classmethod
    def b(cls) -> T:
        return cls.a()
```

This should be treated as valid but `error: Access to classmethod of generic type remains type variables T is ambiguous.` will be happen.

So I'm planning to add more refactoring and more test cases. Please review after this.

kitsuyui:
@ilevkivskyi
I did it. Please review this.

mypy/checkmember.py
```diff
@@ -559,6 +562,15 @@ def analyze_class_attribute_access(itype: Instance,
             msg.fail(messages.GENERIC_INSTANCE_VAR_CLASS_ACCESS, context)
         is_classmethod = ((is_decorated and cast(Decorator, node.node).func.is_class)
                           or (isinstance(node.node, FuncBase) and node.node.is_class))
+        if (is_classmethod and isinstance(itype, Instance)
```
ilevkivskyi: `itype` should be _always_ and `Instance` (see annotation). Have you found a test where it is not?

mypy/expandtype.py
ilevkivskyi: The algorithm is still wrong. For example it will not work for something like this:
```python
class B(Generic[T]):
    @classmethod
    def f(cls) -> T: ...
class C(B[Tuple[T, T]]):
    ...
class D(C[str]):
    ...
reveal_type(D.f())  # Revealed type is 'str', but should be 'Tuple[str, str]'
```
This is not just a small bug there, it is "conceptually" wrong. As I already said twice, you should use something similar to `map_instance_to_supertype` from `maptype.py`. If this is too complex, then I can overtake this PR, it already takes too long.

test-data/unit/check-generics.test
```diff
@@ -1815,3 +1815,214 @@ def g(x: T) -> T: return x
...
+T = TypeVar('T')
+class X(Generic[T]):
+    @classmethod
+    def a(cls) -> T: pass
+    @classmethod
+    def b(cls) -> T:
+        return cls.a()
+X.a  # E: Access to classmethod of generic type remains type variables T is ambiguous.
+reveal_type(X[str].b())  # E: Revealed type is 'builtins.str'
+[builtins fixtures/classmethod.pyi]
```
ilevkivskyi: Thanks for adding many tests! A little note on tests, they will be easier to read if you add few (but only a few) newlines in each separating some logical blocks.

test-data/unit/check-generics.test
```diff
@@ -1815,3 +1815,214 @@ def g(x: T) -> T: return x
...
+from typing import TypeVar, Generic
+T = TypeVar('T')
+class A(Generic[T]):
+    @classmethod
+    def f(cls) -> T: ...
+class B(Generic[T]):
+    @classmethod
+    def f(cls) -> T: ...
+class C(A[str], B[int]): pass
+class D(A[int], B[str]): pass
```
ilevkivskyi: I actually wanted to see a test where a class method return type depends on type variables from two different bases.

mypy/checkmember.py
kitsuyui: @ilevkivskyi 
It seems to be not necessary as you say. I will remove it.

test-data/unit/check-generics.test
kitsuyui: Thank you. I didn't have the confidence. Because I feel that current code bases are using them only fewer.

I will add.

test-data/unit/check-generics.test
kitsuyui: @ilevkivskyi

Does you mean like this?

```python
from typing import Tuple, TypeVar, Generic
T = TypeVar('T')
S = TypeVar('S')


class A(Generic[T]):
    @classmethod
    def f(cls) -> T: ...

class B(Generic[T]):
    @classmethod
    def g(cls) -> T: ...

class C(A[T], B[S]):
    @classmethod
    def h(cls) -> Tuple[T, S]: ...

reveal_type(C[str, int].h())  # E: Revealed type is 'Tuple[builtins.str, builtins.int]'
reveal_type(C[int, str].h())  # E: Revealed type is 'Tuple[builtins.int, builtins.str]'
```

ilevkivskyi: Yes. But I would add one more inheritance:
```python
class D(C[int, str]):
    ...
reveal_type(D.h())
```
I think accessing it on directly on `C[str, int]` should be an error.

kitsuyui: @ilevkivskyi 

> I think accessing it on directly on `C[str, int]` should be an error.

Sorry I don't understand it. Why should `C[str, int]` be an error?
I think it seems to be a valid declaration.

ilevkivskyi: Not `C[str, int]` itself, expression like `C[str, int].h()` should be an error (probably saying something about ambiguous access, etc.)

kitsuyui: @ilevkivskyi 

Thank you. But I still have not understood that it is ambiguous.
I think this case is not ambiguous as same as following case is valid.

```python

T = TypeVar('T')


class A(Generic[T]):
    @classmethod
    def f(cls) -> T: ...


reveal_type(A[str].f())  # E: Revealed type is 'builtins.str'
```

Is this behavior correct?

ilevkivskyi: No, both should be errors. Only subclassing can actually bind a type variable. At runtime, `A[str]` is essentially just an alias to `A` there is no way accessing something on `A[str]` can be different from accessing it on `A`.

On the contrary, this _is_ valid, because there is a new class that (potentially) can add something:
```python
class B(A[str]):
    ...
B.f()  # OK
```
we still can't guarantee it is safe, but I think we should not flag this.

kitsuyui: Oh, thank you. I understand.
And I found that I must fix other test cases that I've added too.

kitsuyui:
I'm going to fix them in a few days.

kitsuyui:
@ilevkivskyi 

I did it. Please review this again?

- Added test-case that class method return type depends on type variables from two different bases.
- Fixed test-cases and error messages calling directly generic class method such like `A[str].f()`.
- Added newlines between logical blocks for readability.

ilevkivskyi:
> I did it. Please review this again?

You didn't fix the main problem (although you marked it as resolved, why?), as I mentioned in two previous reviews, you algorithm is "conceptually" wrong.

It looks like it doesn't make sense to push this forward any more.

kitsuyui:
@ilevkivskyi 
I had closed it by misunderstanding twice.
It is better that if you solve easily and quickly, then it is better than I do any more.
I'm so sorry.

kitsuyui:
@ilevkivskyi Sorry for taking your time again.

I rewrote the code to use `map_instance_to_supertype`.
I read the code that using `map_instance_to_supertype` for ordinary method.

The code for analyze ordinary method uses `TypeInfo.get_method(name)`. This method doesn't suit for classmethod.
So I add `get_classmethod` to `TypeInfo`.

And I added a test for https://github.com/python/mypy/pull/5420#discussion_r228771145

How does this?

ilevkivskyi:
@kitsuyui Thank you for your work! It is a useful starting point for me to try making a better fix for this. Don't worry that this PR didn't make it. You just took a bit too hard issue as a first time contribution.

kitsuyui:
Thank you @ilevkivskyi !
I'm very thankful for mypy. I will take more easier issue next.

PR #5421, opened 2018-08-05
 Don't treat functions with reversible dunder names as such

srittau:
Previously, functions that had the same name as a reversible dunder method (for example, `__rpow__`) were checked as if they were such a method. This could lead to crashes and false positives.

Closes: #5419


mypy/messages.py
```diff
@@ -997,6 +997,7 @@ def overloaded_signatures_ret_specific(self, index: int, context: Context) -> No
     def operator_method_signatures_overlap(
             self, reverse_class: TypeInfo, reverse_method: str, forward_class: Type,
             forward_method: str, context: Context) -> None:
+        assert reverse_class is not None
```
Michael0x2a: I think we can remove this assert -- we already do a check and an assert up above. It'll also be more consistent this way: we don't really perform checks like these in other similar message methods.

(In fact, it actually ought to be impossible for `reverse_class` to be None at this point since we type-check mypy itself under strict-optional mode: the fact that it sometimes *could* be is an unfortunate consequence of [this cast](https://github.com/python/mypy/blob/master/mypy/nodes.py#L401). I'd rather we eventually fix that TODO so we don't have to keep adding asserts -- but that's definitely very much out-of-scope for this PR.)

test-data/unit/check-modules.test
```diff
@@ -2495,3 +2495,6 @@ def __getattr__(attr: str) -> Any: ...
 class Cls: ...
 [builtins fixtures/module.pyi]
 [out]
+
+[case testFunctionWithReversibleDunderName]
+def __rpow__(self) -> int: ...
```
Michael0x2a: While we're at it, maybe leave a test case checking to see what happens if we make this function an overload? And what happens if we define a top-level in-place operator method (like `__iadd__`)?

srittau: Good idea. I will add tests for `__add__`, `__radd__`, and `__iadd__`, so we have covered all three cases.

Michael0x2a:
Thanks!

PR #5426, opened 2018-08-07
Fix two daemon crashes with forward references in decorators

ilevkivskyi:
There are two similar crashes:
* One is caused by using `is_same_type` in `has_type_vars` during third pass of semantic analysis (too soon), I solve this by using equality instead (the only difference between `is_same_type` and equality test is that the former does some union simplification and other type equivalence, while these are not needed in the context of `has_type_vars`).
* Another is caused by in-place changes to callable types done by semantic analysis, to avoid this I take a copy in `defn.unanalyzed_type`.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -396,6 +396,9 @@ class ASTConverter(ast3.NodeTransformer):
                        args,
                        self.as_required_block(n.body, n.lineno),
                        func_type)
+        if isinstance(func_def.type, CallableType):
+            # semanal.py does some in-place modifications we want to avoid
+            func_def.unanalyzed_type = func_def.type.copy_modified()
         if is_coroutine:
             func_def.is_coroutine = True
         if func_type is not None:
--- a/mypy/typevars.py
+++ b/mypy/typevars.py
@@ -3,7 +3,6 @@ from typing import Union, List
 from mypy.nodes import TypeInfo
 
 from mypy.erasetype import erase_typevars
-from mypy.sametypes import is_same_type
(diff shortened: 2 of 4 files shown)


mypy/fastparse.py
```diff
@@ -396,6 +396,9 @@ def do_func_def(self, n: Union[ast3.FunctionDef, ast3.AsyncFunctionDef],
                        args,
                        self.as_required_block(n.body, n.lineno),
                        func_type)
+        if isinstance(func_def.type, CallableType):
+            # semanal.py does some in-place modifications we want to avoid
+            func_def.unanalyzed_type = func_def.type.copy_modified()
```
msullivan: Is a similar change required to fastparse2?

mypy/typevars.py
```diff
@@ -22,4 +21,4 @@ def fill_typevars(typ: TypeInfo) -> Union[Instance, TupleType]:
 
 
 def has_no_typevars(typ: Type) -> bool:
-    return is_same_type(typ, erase_typevars(typ))
+    return typ == erase_typevars(typ)
```
msullivan: Add a comment explaining what is going on here, since doing equality on types is usually the wrong thing, since it isn't overloaded. Might make more sense for it to be `is`?

msullivan: (on the grounds that explicit is better than implicit, so if we are just doing object identity might as well be obvious about it)

msullivan: Oh lurr I actually didn't realize that `Type` subclasses all had overloaded `__eq__`

PR #5428, opened 2018-08-07
Document stub-only packages incompatibility with MYPYPATH

### LINKED ISSUE #5386: MYPYPATH not working with local -stubs package
This is a bug report. However, if the current behavior is not a bug, then this is request to improve documentation.

## Description

Mypy doesn't recognise my local library typing stub.

### Actual Python file

```python
import aiomysql
```

### Directory tree
```
pythonlibtypes/
└── aiomysql-stubs
    ├── __init__.pyi
    ├── connection.pyi
    ├── cursors.pyi
    ├── pool.pyi
    ├── py.typed
    └── utils.pyi
```

### Mypy command

```
MYPYPATH="/path/to/project/pythonlibtypes/" pipenv run -- mypy --config mypy.ini module_to_check
```

`mypy.ini`:

```
[mypy]
warn_redundant_casts = True
warn_unused_ignores = True
incremental = True
follow_imports = normal
```

## Actual behavior

```
path/to/file.py:1: error: Cannot find module named 'aiomysql'
```

## Expected behavior

Import succeeds, Mypy uses the stub 

## Versions

Python 3.6.5

mypy 0.620

ethanhs:
Closes #5386.

### DIFF
--- a/docs/source/installed_packages.rst
+++ b/docs/source/installed_packages.rst
@@ -30,6 +30,9 @@ imports, namespace packages, and custom import hooks.
 If you do not want to use typed packages, use the ``--no-site-packages`` flag
 to disable searching.
 
+Note that stub-only packages cannot be used with ``MYPYPATH``. If you want mypy
+to find the package, it must be installed.
+
 Making PEP 561 compatible packages
 **********************************
 
--- a/docs/source/installed_packages.rst
+++ b/docs/source/installed_packages.rst
@@ -30,8 +30,10 @@ imports, namespace packages, and custom import hooks.
 If you do not want to use typed packages, use the ``--no-site-packages`` flag
 to disable searching.
 
-Note that stub-only packages cannot be used with ``MYPYPATH``. If you want mypy
-to find the package, it must be installed.
+Note that stub-only packages (defined in
+`PEP 561 <https://www.python.org/dev/peps/pep-0561/#stub-only-packages>`_)
+cannot be used with ``MYPYPATH``. If you want mypy to find the package, it must
+be installed.
 
 Making PEP 561 compatible packages
 **********************************
--- a/docs/source/installed_packages.rst
+++ b/docs/source/installed_packages.rst
@@ -33,7 +33,9 @@ to disable searching.
 Note that stub-only packages (defined in
 `PEP 561 <https://www.python.org/dev/peps/pep-0561/#stub-only-packages>`_)
 cannot be used with ``MYPYPATH``. If you want mypy to find the package, it must
-be installed.
+be installed. For a package ``foo``, the name of the stub-only package
+(``foo-stubs``) is not a legal package name, so mypy will not find it, unless
+it is installed.
 
 Making PEP 561 compatible packages
 **********************************
--- a/docs/source/running_mypy.rst
+++ b/docs/source/running_mypy.rst
@@ -306,7 +306,7 @@ This is computed from the following items:
 
 .. note::
 
-    You cannot point to a PEP 561 package via the MYPYPATH, they must be
+    You cannot point to a PEP 561 package via the MYPYPATH, it must be
     installed (see :ref:`PEP 561 support <installed-packages>`)
 
 For sources given on the command line, the path is adjusted by crawling
--- a/docs/source/running_mypy.rst
+++ b/docs/source/running_mypy.rst
@@ -304,6 +304,11 @@ This is computed from the following items:
 - The relevant directories of the
   `typeshed <https://github.com/python/typeshed>`_ repo.
 
+.. note::
+
+    You cannot point to a PEP 561 package via the MYPYPATH, they must be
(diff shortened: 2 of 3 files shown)


docs/source/stubs.rst
```diff
@@ -25,7 +25,9 @@ Here is an overview of how to create a stub file:
     $ export MYPYPATH=~/work/myproject/stubs
 
 Use the normal Python file name conventions for modules, e.g. ``csv.pyi``
-for module ``csv``. Use a subdirectory with ``__init__.pyi`` for packages.
+for module ``csv``. Use a subdirectory with ``__init__.pyi`` for packages. Note
+that PEP 561 stub-only packages must be installed, and may not be pointed at
```
Michael0x2a: Maybe link to the PEP 561 page here? I think this is the first time we mention it on this page.

docs/source/installed_packages.rst
```diff
@@ -30,6 +30,9 @@ imports, namespace packages, and custom import hooks.
 If you do not want to use typed packages, use the ``--no-site-packages`` flag
 to disable searching.
 
+Note that stub-only packages cannot be used with ``MYPYPATH``. If you want mypy
+to find the package, it must be installed.
```
Michael0x2a: Can you clarify either here or somewhere else on this page what it means for a package to be "stub-only"? E.g. is a stub-only package a package that contains only `.pyi` files, or is a stub-only package one that has a name ending in `-stub`, contains only `.pyi` files, and contains a file named `py.typed`? 

(And if you do decide to add the exact definition somewhere else on the page, we should probably link to that here.)

Similarly, maybe clarify what it means for a package to be "installed"?

docs/source/stubs.rst
ethanhs: Sure, that is a good idea.

docs/source/installed_packages.rst
ethanhs: I think ideally the docs should have a glossary, and this would be good to add there. I think linking to PEP 561 https://www.python.org/dev/peps/pep-0561/#stub-only-packages in this case is enough.

ethanhs:
>Do we also need to update anything in this section of the docs?

No, I don't think it is relevant there, that section doesn't talk a lot about MYPYPATH.

>Do you think it might be a good idea to add instructions on using pip to create an editable install somewhere?

That might be a good idea, but I don't think an editable install is necessarily the best solution for everyone. That is why I want to say "install" over "editable install", because for most uses (those in a virtualenv) a regular install is better for general packaging purposes.

Michael0x2a:
> > Do we also need to update anything in this section of the docs?
>
> No, I don't think it is relevant there, that section doesn't talk a lot about MYPYPATH.

Is that what we want, though?  If somebody is looking at that section, I imagine it's because they're running into issues getting their imports to resolve. 

That might be due to a number of different issues -- one of which might include confusion on how exactly MYPYPATH works. In that case, perhaps it would be helpful for those readers if that section *did* talk more about MYPYPATH (or links to a page that does)?

> > Do you think it might be a good idea to add instructions on using pip to create an editable install somewhere?
>
> That might be a good idea, but I don't think an editable install is necessarily the best solution for everyone. That is why I want to say "install" over "editable install", because for most uses (those in a virtualenv) a regular install is better for general packaging purposes.

Mkay, sounds good.

ethanhs:
>Is that what we want, though? If somebody is looking at that section, I imagine it's because they're running into issues getting their imports to resolve.

OK, I will add something there too.

ethanhs:
@gvanrossum when you look at #5517 tomorrow perhaps you could briefly take a look at this as well?

gvanrossum:
Will do!

docs/source/installed_packages.rst
```diff
@@ -30,6 +30,11 @@ imports, namespace packages, and custom import hooks.
 If you do not want to use typed packages, use the ``--no-site-packages`` flag
 to disable searching.
 
+Note that stub-only packages (defined in
+`PEP 561 <https://www.python.org/dev/peps/pep-0561/#stub-only-packages>`_)
+cannot be used with ``MYPYPATH``. If you want mypy to find the package, it must
+be installed.
```
gvanrossum: Maybe explain that the reason is that the stubs for `foo` are in a folder named `foo-stubs`?

docs/source/running_mypy.rst
```diff
@@ -304,6 +304,11 @@ This is computed from the following items:
 - The relevant directories of the
   `typeshed <https://github.com/python/typeshed>`_ repo.
 
+.. note::
+
+    You cannot point to a PEP 561 package via the MYPYPATH, they must be
```
gvanrossum: they -> it (or change the first half to plural :-).

docs/source/stubs.rst
gvanrossum: (This is not done yet.)

PR #5430, opened 2018-08-08
Replaced --silent-imports with --ignore-missing-imports

wilk:


docs/source/running_mypy.rst
```diff
@@ -197,7 +197,7 @@ accepts one of four string values:
     replace the module (and *anything imported from it*) with an
     object of type ``Any``.
 
-    (Note: this option used to be known as ``--silent-imports``.)
+    (Note: this option used to be known as ``--ignore-missing-imports``.)
```
ilevkivskyi: I don't think this is right. `--ignore-missing-imports` is still a valid option and is completely orthogonal to `--follow-imports=skip`

wilk: In fact, I've replaced `--silent-imports` with `--ignore-missing-imports`, due to this message:

screen shot 2018-08-08 at 10 50 15

ilevkivskyi: This message doesn't contradict the current docs, they say _used to be known_. Also as the message says, you should use both flags instead of one.

wilk: Well, then that warning is misleading.
What I understand from that is not to use `--silent-imports` and instead to replace it with two flags, as you said.
Otherwise, if `--silent-imports` is valid why `mypy` should warn me?

ilevkivskyi: It seems to me you just don't understand the meaning of "used to be known". I don't see a reason to continue this discussion.

wilk: Maybe that warning sounds like "deprecated"?
Otherwise is useless pointing out that info and again misleading with a warning, imho.

Issue #4649, opened 2018-02-28
Report of INTERNAL ERROR with minimal reproduction


evhub:
The following code snippet
```python
from functools import reduce
M = Iterable
def f(m1: M, m2):
    ...
def g(ms: T[M]) -> None:
    reduce(f, ms)
T = Iterable
```
when run through `mypy --show-traceback` on version `0.560` produces the following traceback
```
.\break_mypy.py:22: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.560
Traceback (most recent call last):
  File "c:\anaconda3\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\anaconda3\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Anaconda3\Scripts\mypy.exe\__main__.py", line 9, in <module>
    sys.exit(console_entry())
  File "c:\anaconda3\lib\site-packages\mypy\__main__.py", line 7, in console_entry
    main(None)
  File "c:\anaconda3\lib\site-packages\mypy\main.py", line 66, in main
    res = type_check_only(sources, bin_dir, options)
  File "c:\anaconda3\lib\site-packages\mypy\main.py", line 119, in type_check_only
    options=options)
  File "c:\anaconda3\lib\site-packages\mypy\build.py", line 218, in build
    graph = dispatch(sources, manager)
  File "c:\anaconda3\lib\site-packages\mypy\build.py", line 1997, in dispatch
    process_graph(graph, manager)
  File "c:\anaconda3\lib\site-packages\mypy\build.py", line 2299, in process_graph
    process_stale_scc(graph, scc, manager)
  File "c:\anaconda3\lib\site-packages\mypy\build.py", line 2475, in process_stale_scc
    graph[id].type_check_first_pass()
  File "c:\anaconda3\lib\site-packages\mypy\build.py", line 1876, in type_check_first_pass
    self.type_checker().check_first_pass()
  File "c:\anaconda3\lib\site-packages\mypy\checker.py", line 196, in check_first_pass
    self.accept(d)
  File "c:\anaconda3\lib\site-packages\mypy\checker.py", line 286, in accept
    stmt.accept(self)
  File "c:\anaconda3\lib\site-packages\mypy\nodes.py", line 497, in accept
    return visitor.visit_func_def(self)
  File "c:\anaconda3\lib\site-packages\mypy\checker.py", line 534, in visit_func_def
[...]

PR #5433, opened 2018-08-08
Visit resolved forward refs by third pass type visitor

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/4649

This also fixes similar cases of crashes with extreme forward referencing. The idea is quite straightforward, previously forward references just resolved during third pass were not accepted by the third pass visitor itself, this is now fixed.

### DIFF
--- a/mypy/typeanal.py
+++ b/mypy/typeanal.py
@@ -708,28 +708,7 @@ class TypeAnalyserPass3(TypeVisitor[None]):
             self.indicator['synthetic'] = True
         # Check type argument count.
         if len(t.args) != len(info.type_vars):
-            if len(t.args) == 0:
-                any_type = AnyType(TypeOfAny.from_omitted_generics,
-                                   line=t.line, column=t.column)
-                t.args = [any_type] * len(info.type_vars)
-                return
-            # Invalid number of type parameters.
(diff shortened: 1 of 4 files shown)


mypy/typeanal.py
```diff
@@ -824,6 +805,35 @@ def anal_type(self, tp: UnboundType) -> Type:
...
+    act = str(len(t.args))
+    if act == '0':
+        act = 'none'
+    fail('"{}" expects {}, but {} given'.format(
+        t.type.name(), s, act), t)
+    # Construct the correct number of type arguments, as
+    # otherwise the type checker may crash as it expects
+    # things to be right.
+    t.args = [AnyType(TypeOfAny.from_error) for _ in t.type.type_vars]
+    t.invalid = True
```
msullivan: Did anything change here, or is it a pure refactor?

ilevkivskyi: This is a pure refactor. I first tried a different solution, that required calling this manually from other module, but then found a simpler solution. However, looking at it, it seemed to me it actually looks better as a separate function (all visitor methods look compact this way).

msullivan: Agreed, just wanted to check I wasn't missing anything

PR #5436, opened 2018-08-08
Add a `FlexibleAlias` to `mypy_extensions`

msullivan:
Currently there is no way to define generic type aliases that do not
depend on their type arguments or that are just directly a type
variable.

`FlexibleAlias[T, typ]` creates a type that depends on `T` but that is
expanded to `typ` during type alias expansion.

One target use case for this is in creating conditionally defined type
aliases that depend on their argument under certain configurations but
not under others, for example:
```
if BOGUS:
    Bogus = FlexibleAlias[T, Any]
else:
    Bogus = FlexibleAlias[T, T]
```

### DIFF
--- a/test-data/unit/check-type-aliases.test
+++ b/test-data/unit/check-type-aliases.test
@@ -426,3 +426,95 @@ reveal_type(D.meth(1))  # E: Revealed type is 'Union[__main__.D*, builtins.int]'
 reveal_type(D().meth(1))  # E: Revealed type is 'Union[__main__.D*, builtins.int]'
 [builtins fixtures/classmethod.pyi]
 [out]
+
+[case testFlexibleAlias1]
+from typing import TypeVar, List, Tuple
+from mypy_extensions import FlexibleAlias
+
+T = TypeVar('T')
+U = TypeVar('U')
+
+AnInt = FlexibleAlias[T, int]
+
+x: AnInt[str]
+reveal_type(x)  # E: Revealed type is 'builtins.int'
+
+TwoArgs = FlexibleAlias[Tuple[T, U], bool]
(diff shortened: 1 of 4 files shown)


msullivan:
I think this could use some more thought and probably some more sanity checks on its use before we document it as an "official" mypy_extension, but I think it should be find to land something along these lines first.

test-data/unit/check-type-aliases.test
```diff
@@ -426,3 +426,95 @@ reveal_type(D.meth(1))  # E: Revealed type is 'Union[__main__.D*, builtins.int]'
...
+def take_id(x: Id[int]) -> None:
+    reveal_type(x)  # E: Revealed type is 'builtins.int'
+
+def id(x: Id[T]) -> T:
+    return x
+
+# TODO: This doesn't work and maybe it should?
+# Indirection = AnInt[T]
+# y: Indirection[str]
+# reveal_type(y)  # E : Revealed type is 'builtins.int'
```
ilevkivskyi: This doesn't work because all alias expansions happen immediately. When mypy sees `AnInt[T]` wherever, including in the r.h.s. above, it is immediately expanded to just `int`, and then a new alias is created, i.e. aliases are never "chained".

One can argue that this could be similar to how we allow omitting `Generic[T]` when subclassing `Iterable[T]`, but I would rather keep this simple, if someone really wants, one can use the form below.

Also it would be a relatively larger change because we would need to move querying for type variables in the middle of analyzing the alias.

ilevkivskyi:
Could you please also remove the typeshed update from this PR? You can do a separate sync later after the typeshed PR is merged.

PR #5437, opened 2018-08-08
Split the type visitor base classes into a new module

msullivan:
This is to support converting mypy to run under mypyc, since current
mypyc-extension classes can extend interpreted classes but not the
other way around. Separating them out, then, allows us to compile
types before we can compile everything that uses a TypeVisitor.

The visitors are all re-exported from mypyc.types and that is how
other modules refer to them.

### DIFF
--- a/mypy/type_visitor.py
+++ b/mypy/type_visitor.py
@@ -7,7 +7,7 @@ mypyc-extension classes can extend interpreted classes but not the
 other way around. Separating them out, then, allows us to compile
 types before we can compile everything that uses a TypeVisitor.
 
-The visitors are all re-exported from mypyc.types and that is how
+The visitors are all re-exported from mypy.types and that is how
 other modules refer to them.
 """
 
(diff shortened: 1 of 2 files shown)


mypy/type_visitor.py
```diff
@@ -0,0 +1,204 @@
+"""Type visitor classes.
+
+This module defines the type visitors that are intended to be
+subclassed by other code.  They have been separated out into their own
+module to ease converting mypy to run under mypyc, since currently
+mypyc-extension classes can extend interpreted classes but not the
+other way around. Separating them out, then, allows us to compile
+types before we can compile everything that uses a TypeVisitor.
+
+The visitors are all re-exported from mypyc.types and that is how
```
ilevkivskyi: Typo "mypyc.types" -> "mypy.types"

mypy/types.py
```diff
@@ -25,6 +25,12 @@
 
 MYPY = False
 
+# If we import type_visitor in the middle of the file, mypy breaks, and if we do it
```
ilevkivskyi: Do I understand correctly this simply because visitors are used in annotations?

PR #5440, opened 2018-08-09
Sync typeshed

ilevkivskyi:


ilevkivskyi:
OK, I think we can just sync with master.

PR #5450, opened 2018-08-10
Bump the mypy_extensions version number to 0.5.0-dev

msullivan:


JelleZijlstra:
(Rerunning a spurious Travis failure.)

msullivan:
Only @JukkaL and @gvanrossum  have `mypy_extensions` PyPI access, though, so my planned version bump needs to wait for one of them to return from vacation

PR #5458, opened 2018-08-12
Fix descriptor access for dynamic base classes

### LINKED ISSUE #5456: Any base class should not match None
Consider this (oversimplified) code:
```python
from typing import Any

dynamic: Any

class C(dynamic): ...

y: None = C()
```
It passes mypy without errors, but it arguably should not, because `None` is a _non-subclassable_ singleton.

In reality, this causes the following problem: descriptors need to distinguish between access on classes and access on instances, in the former case the instance is `None`. Mypy works well with this pattern that allows to correctly type various ORM patterns. However, if a class has `Any` fallback, then the class and instance access are confused causing weird bugs.

ilevkivskyi:
Fixes #5456

### DIFF
--- a/mypy/subtypes.py
+++ b/mypy/subtypes.py
@@ -141,6 +141,8 @@ class SubtypeVisitor(TypeVisitor[bool]):
 
     def visit_instance(self, left: Instance) -> bool:
         if left.type.fallback_to_any:
+            if isinstance(self.right, NoneTyp):
+                return False
             return True
         right = self.right
         if isinstance(right, TupleType) and right.fallback.type.is_enum:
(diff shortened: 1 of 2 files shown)


mypy/subtypes.py
```diff
@@ -141,6 +141,8 @@ def visit_deleted_type(self, left: DeletedType) -> bool:
 
     def visit_instance(self, left: Instance) -> bool:
         if left.type.fallback_to_any:
+            if isinstance(self.right, NoneTyp):
```
msullivan: Might be worth adding a comment here explaining what is special about `None` here, and why we care, because this is a bit ad-hoc. (Though I suppose most things about `fallback_to_any` are a bit ad-hoc?)

ilevkivskyi: Makes sense, will add a comment.

PR #5459, opened 2018-08-12
Lazy tests: parse using regex first

### LINKED ISSUE #1073: Make short test runs faster
Currently we create a new Python process for each unit test suite, which is wasteful when running just a small number of test cases, as the interpreter startup overhead has to be paid many times. Instead, we could use long-running processes that run multiple unit test tasks that get passed via a pipe or something from the main test runner process.

The goal is to make something like this run faster:

```
$ time ./runtests.py unit-test -a "missingtestname"
PARALLEL 8
SUMMARY  13 tasks selected
SUMMARY  all 13 tasks and 0 tests passed                                        
*** OK ***

real	0m0.584s
user	0m2.556s
sys	0m0.187s
```

If we can get this to ~0.3s on typical hardware I'd be happy.

elazarg:
Fix #1073: parse tests on collection only enough to find the name, so small number of tests run faster

On my machine, `pytest -n0 -k testAttrsSimple` takes at least 2.24 seconds to finish on master, and at most 0.95 seconds to finish with this PR.

- 'skip-cache' is changed to 'only_when_nocache' and similarly 'skip-nocache'
- I have replace while loops with "if True" to make the diff simpler. Further cleanup and optimizations are also possible


mypy/test/data.py
```diff
@@ -615,14 +600,37 @@ def pytest_pycollect_makeitem(collector: Any, name: str,
     return None
 
 
+def split_test_cases(parent: 'DataSuiteCollector', suite: 'DataSuite',
```
msullivan: Add a docstring, with some rationale?

elazarg:
Additional refactoring and cleanup:
* rename variables to be consistent and clear (e.g. `test_items` instead of `p`) 
* extract variables and switch to for-loop
* combine the `out` cases, removing repetition
* rechecked/stale cases reformatted so the repetition is obvious
* fail faster in case `out_section_missing`, with a clearer error message

mypy/test/data.py
```diff
@@ -44,144 +42,111 @@ def parse_test_case(suite: 'DataSuite', p: List['TestItem'], path: str,
 
     # Process the parsed items. Each item has a header of form [id args],
```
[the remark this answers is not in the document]
elazarg: No, it's still plural - items are not test cases. But I should move it to be closer to the loop

elazarg:
Linux 3.4 / 3.5.1 seem to fail spuriously - "Unable to download 3.5.1 archive"?

(The windows tests fail due to path separator on the expected output being reversed when it shouldn't)

msullivan:
I've restarted the linux builds

mypy/test/data.py
```diff
@@ -33,16 +33,14 @@ def parse_test_case(suite: 'DataSuite', test_items: List['TestItem'], file: str,
                     case: 'DataDrivenTestCase') -> None:
     """Parse and prepare a single case from suite with test case descriptions.
 
-    This method is part of the setup phase, just before the test case is run.
+    This method is part of the setup phase, just before the test case is run.s
```
msullivan: Spurious 's'

gvanrossum:
Since this PR landed, `pytest -v mypy/test/testcheck.py` prints a stream of messages like this:
```
[location] PASSED                   [  0%]
[location] PASSED                   [  0%]
[location] PASSED                   [  0%]
```
Previously it would print the test name instead of the string `[location]`, e.g.:
```
mypy/test/testcheck.py::TypeCheckSuite::testEmptyFile <- /Users/guido/src/mypy-master/test-data/unit/check-basic.test PASSED                                              [  0%]
mypy/test/testcheck.py::TypeCheckSuite::testAssignmentAndVarDef <- /Users/guido/src/mypy-master/test-data/unit/check-basic.test PASSED                                    [  0%]
mypy/test/testcheck.py::TypeCheckSuite::testConstructionAndAssignment <- /Users/guido/src/mypy-master/test-data/unit/check-basic.test PASSED                              [  0%]
```
Methinks something went wrong here. @elazarg do you have a hunch as to what caused this? (I feel I've done enough research by bisecting it to this PR.)

elazarg:
Of course. Sorry!

PR #5460, opened 2018-08-12
Move descriptor access logic to checkmember module

### LINKED ISSUE #5455: Descriptor transformation should not be made for __getattr__
Consider this code:
```python
class Descr:
    def __get__(self, inst, owner) -> int: ...

class C:
    def __getattr__(self, attr: str) -> Descr: ...

c = C()
reveal_type(c.dyn_attr)
```
where mypy incorrectly infers an `int`. The point is that descriptor protocol is not invoked for the result of `__getattr__`. In general, I am not sure we cover all the corner cases correctly (like setting a descriptor on `self` etc). Fixing this will probably require moving the descriptor logic to `checkmember.py` (which is actually a plus, I think) otherwise it should be straightforward.

ilevkivskyi:
Fixes #5455 

This does a minor refactoring and also fixes two corner cases:
* Descriptors are not invoked for access via `__getattr__`
* Descriptors are not invoked if they were assigned to `self` (as opposite to defined in _class body_)

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -35,7 +35,10 @@ from mypy.types import (
 from mypy.sametypes import is_same_type, is_same_types
 from mypy.messages import MessageBuilder, make_inferred_type_note
 import mypy.checkexpr
-from mypy.checkmember import map_type_from_supertype, bind_self, erase_to_bound, type_object_type
+from mypy.checkmember import (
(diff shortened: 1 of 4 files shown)


ilevkivskyi:
> Just to make sure I'm understanding this all property, the reason it needs to move to `checkmember` is because otherwise we were doing the descriptor checks after `__getattr__` was resolved with no obvious way to avoid that?

Exactly!

PR #5463, opened 2018-08-12
Support callback protocols

### LINKED ISSUE #5453: Functions can't be passed as callable protocol
This came up in python/typeshed#2379. Consider the following code:

```python
from typing import Protocol

class Caller(Protocol):
    def __call__(self) -> None: ...

def call() -> None:
    pass

def func(caller: Caller) -> None:
    pass

func(call)
```

Currently, mypy 0.620 rejects this:

```
foo.py:12: error: Argument 1 to "func" has incompatible type "Callable[[], None]"; expected "Caller"
```

I believe that mypy should actually allow this, because `call` is actually an object that fulfills the protocol. This would also open up the opportunity to create callback signatures of any complexity (for example, default and keyword-only arguments) without the need for additional features for `Callable`.

ilevkivskyi:
Fixes #5453 

The issue proposed an interesting idea. Allow callables as subtypes of protocols with `__call__`. IMO this is not just reasonable and type safe, but is also more clear that the extended callable syntax in `mypy_extensions`. For example:
```python
class Combiner(Protocol):
    def __call__(self, *vals: bytes, maxlen: Optional[int] = None) -> List[bytes]: ...

def batch_proc(data: Iterable[bytes], cb_results: Combiner) -> bytes:
    ...
```

The callback protocols:
* Have cleaner familiar syntax in contrast to `Callable[[VarArg(bytes), DefaultNamedArg(Optional[int], 'maxlen')], List[bytes]]` (compare to above)
* Allow to be more explicit/flexible about binding of type variables in generic callbacks (see tests for examples)
* Support overloaded callbacks (this is simply impossible with the current extended callable syntax)
* Are easy to implement

If this will get some traction, I would actually propose to deprecate extended callable syntax in favor of callback protocols.

### DIFF
--- a/docs/source/protocols.rst
+++ b/docs/source/protocols.rst
@@ -415,9 +415,10 @@ in ``typing`` such as ``Iterable``.
 Callback protocols
 ******************
 
-Protocols can be used to define flexible callback types that are impossible
-to express using the ``Callable[...]`` syntax, for example variadic, overloaded,
-and complex generic callbacks. They are defined with a special ``__call__`` member:
+Protocols can be used to define flexible callback types that are hard
+(or even impossible) to express using the ``Callable[...]`` syntax, such as variadic,
+overloaded, and complex generic callbacks. They are defined with a special ``__call__``
+member:
 
 .. code-block:: python
 
@@ -441,8 +442,8 @@ and complex generic callbacks. They are defined with a special ``__call__`` memb
                             # different name and kind in the callback
 
 Callback protocols and ``Callable[...]`` types can be used interchangeably.
-To indicate an "anonymous" positional-only argument use double underscore,
-for example:
+Keyword argument names in ``__call__`` methods must be identical, unless
+a double underscore prefix is used. For example:
 
 .. code-block:: python
 
--- a/docs/source/protocols.rst
+++ b/docs/source/protocols.rst
@@ -409,3 +409,53 @@ in ``typing`` such as ``Iterable``.
    ``isinstance()`` with protocols is not completely safe at runtime.
    For example, signatures of methods are not checked. The runtime
    implementation only checks that all protocol members are defined.
+
+.. _callback_protocols:
+
+Callback protocols
+******************
+
+Protocols can be used to define flexible callback types that are impossible
+to express using the ``Callable[...]`` syntax, for example variadic, overloaded,
+and complex generic callbacks. They are defined with a special ``__call__`` member:
+
+.. code-block:: python
+
+   from typing import Optional, Iterable, List
+   from typing_extensions import Protocol
+
+   class Combiner(Protocol):
+       def __call__(self, *vals: bytes, maxlen: Optional[int] = None) -> List[bytes]: ...
+
+   def batch_proc(data: Iterable[bytes], cb_results: Combiner) -> bytes:
+       for item in data:
+           ...
+
+   def good_cb(*vals: bytes, maxlen: Optional[int] = None) -> List[bytes]:
+       ...
+   def bad_cb(*vals: bytes, maxitems: Optional[int]) -> List[bytes]:
+       ...
+
+   batch_proc([], good_cb)  # OK
+   batch_proc([], bad_cb)   # Error! Argument 2 has incompatible type because of
+                            # different name and kind in the callback
+
+Callback protocols and ``Callable[...]`` types can be used interchangeably.
+To indicate an "anonymous" positional-only argument use double underscore,
+for example:
+
+.. code-block:: python
+
+   from typing import Callable, TypeVar
+   from typing_extensions import Protocol
+
+   T = TypeVar('T')
+
+   class Copy(Protocol):
+       def __call__(self, __origin: T) -> T: ...
+
+   copy_a: Callable[[T], T]
+   copy_b: Copy
+
+   copy_a = copy_b  # OK
+   copy_b = copy_a  # Also OK
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1308,10 +1308,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         """
         # Use boolean variable to clarify code.
         fail = False
-        # __call__ is special among dunders, because arguments can be passed
-        # as keyword args (unlike other dunders).
-        ignore_names = name != '__call__'
-        if not is_subtype(override, original, ignore_pos_arg_names=ignore_names):
+        if not is_subtype(override, original, ignore_pos_arg_names=True):
             fail = True
         elif (not isinstance(original, Overloaded) and
               isinstance(override, Overloaded) and
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3170,6 +3170,11 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                 call = find_member('__call__', subtype, subtype)
                 if call:
                     self.msg.note_call(subtype, call, context)
+            if isinstance(subtype, (CallableType, Overloaded)) and isinstance(supertype, Instance):
+                if supertype.type.is_protocol and supertype.type.protocol_members == ['__call__']:
+                    call = find_member('__call__', supertype, subtype)
+                    assert call is not None
+                    self.msg.note_call(supertype, call, context)
             return False
 
     def contains_none(self, t: Type) -> bool:
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1308,7 +1308,10 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         """
         # Use boolean variable to clarify code.
         fail = False
-        if not is_subtype(override, original, ignore_pos_arg_names=True):
+        # __call__ is special among dunders, because arguments can be passed
+        # as keyword args (unlike other dunders).
+        ignore_names = name != '__call__'
+        if not is_subtype(override, original, ignore_pos_arg_names=ignore_names):
             fail = True
         elif (not isinstance(original, Overloaded) and
               isinstance(override, Overloaded) and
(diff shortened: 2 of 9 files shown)


ethanhs:
>If this will get some traction, I would actually propose to deprecate extended callable syntax in favor of callback protocols.

I really like the new syntax, and think we should go ahead with your proposal.

msullivan:
One snag to replacing extended callables is that protocols don't seem to enforce argument name equivalence.

ilevkivskyi:
@msullivan Could you please clarify what do you mean? I think they behave exactly the same way as "normal" callables (including the dunder argument name convention), see the test I added.

msullivan:
I mean that mypy will reject the following:
```
def foo(f: Callable[[NamedArg(int, 'x')], None]) -> None:
    f(x = 10)

def bar(y: int) -> None:
    pass

foo(bar)
```
while the protocol version won't.

msullivan:
Oh which maybe that is a bug in protocols more generally?

ilevkivskyi:
@msullivan maybe I am missing something, but isn't `testCallableImplementsProtocolArgName` exactly testing what you wrote?

ilevkivskyi:
(the error message is cryptic in this case, because arg names are not shown, but we already have a high priority issue about this)

msullivan:
Ah, my confusion was caused by testing that I did using regular protocols, where it *doesn't* seem to check argument names:
```
from typing import Protocol

class Caller(Protocol):
    def foo(self, x: str) -> None: ...

def func(caller: Caller) -> None:
    caller.foo(x='lol')

class Foo:
    def foo(self, y: str) -> None: ...

func(Foo())
```

ilevkivskyi:
Yes, and this is intentional, see this code in `is_protocol_implementation`:
```python
                # Nominal check currently ignores arg names
                is_compat = is_subtype(subtype, supertype, ignore_pos_arg_names=True)
```
the motivation for this is probably very old, I just wanted structural to match nominal as much as possible.

test-data/unit/check-protocols.test
```diff
@@ -2187,3 +2187,123 @@ else:
...
+def call(x: str, *args: int) -> None:
+    pass
+def bad(x: int, *args: str) -> None:
+    pass
+
+def func(caller: Caller) -> None:
+    pass
+
+func(call)
+func(bad)  # E: Argument 1 to "func" has incompatible type "Callable[[int, VarArg(str)], None]"; expected "Caller"
```
JukkaL: Once we deprecate the old syntax, we should come up with a different syntax for "rich" callables in error messages as well.

test-data/unit/check-protocols.test
```diff
@@ -2187,3 +2187,123 @@ else:
...
+
+class Caller(Protocol[T, S]):
+    def __call__(self, x: T, y: S) -> Tuple[T, S]: ...
+
+def call(x: int, y: str) -> Tuple[int, str]: ...
+
+def func(caller: Caller[T, S]) -> Tuple[T, S]:
+    pass
+
+reveal_type(func(call))  # E: Revealed type is 'Tuple[builtins.int*, builtins.str*]'
```
JukkaL: Test also passing a generic function when a function accepts a non-generic protocol-based callable (where the generic function is valid).

test-data/unit/check-protocols.test
```diff
@@ -2187,3 +2187,123 @@ else:
...
+    pass
+
+def anon(caller: CallerAnon) -> None:
+    pass
+
+
+func(call)
+func(bad)  # E: Argument 1 to "func" has incompatible type "Callable[[str], None]"; expected "Caller"
+anon(bad)
+[out]
```
JukkaL: Idea for test cases:

* Pass protocol callable `X` when function accepts protocol callable `Y` (both compatible and incompatible cases). Should these match argument names precisely or not?
* Pass protocol callable when function accepts a normal callable. Should this be okay?

mypy/subtypes.py
```diff
@@ -214,6 +214,13 @@ def visit_callable_type(self, left: CallableType) -> bool:
                                   ignore_pos_arg_names=self.ignore_pos_arg_names)
                        for item in right.items())
         elif isinstance(right, Instance):
+            if right.type.is_protocol and right.type.protocol_members == ['__call__']:
```
JukkaL: Maybe we should also support mixing protocol callables and normal callables in joins and meets?

JukkaL:
Also, there is a merge conflict.

test-data/unit/check-protocols.test
ilevkivskyi: The format is already there, `MessageBuilder.pretty_callable`, but we need to update loads of test cases.

test-data/unit/check-protocols.test
ilevkivskyi: OK, will add this test.

test-data/unit/check-protocols.test
ilevkivskyi: The second is already supported and has test cases, it was part of the original protocols PR. For the first however, the argument names are not checked, but I think they should be (even for nominal), because `__call__` is special, unlike for `__add__` arg names are "visible".

mypy/subtypes.py
ilevkivskyi: Hm, I will try this.

ilevkivskyi:
@JukkaL Thanks for review! Could you please take a look again at this?

mypy/checker.py
```diff
@@ -1308,7 +1308,10 @@ def check_override(self, override: FunctionLike, original: FunctionLike,
         """
         # Use boolean variable to clarify code.
         fail = False
-        if not is_subtype(override, original, ignore_pos_arg_names=True):
+        # __call__ is special among dunders, because arguments can be passed
+        # as keyword args (unlike other dunders).
+        ignore_names = name != '__call__'
+        if not is_subtype(override, original, ignore_pos_arg_names=ignore_names):
```
JukkaL: Hmm, this could break some existing code which assumes that `__call__` won't be called with keyword arguments. Maybe there is a way to restrict this to only protocol classes, where `__call__` is special among dunder methods?

test-data/unit/check-classes.test
```diff
@@ -4636,3 +4636,10 @@ class C:
 reveal_type(C.foo)  # E: Revealed type is 'builtins.int*'
 reveal_type(C().foo)  # E: Revealed type is 'builtins.int*'
 [out]
+
+[case tesBadCallOverride]
+class Base:
+    def __call__(self, arg: int) -> None: ...
+class Sub(Base):
+    def __call__(self, zzz: int) -> None: ...  # E: Signature of "__call__" incompatible with supertype "Base"
```
JukkaL: As discussed above, this might break existing code, and the override doesn't seem special since non-protocols aren't compatible with callable types.

test-data/unit/check-protocols.test
```diff
@@ -2307,3 +2327,63 @@ func(call)
...
+class D(B): ...
+
+class Call(Protocol):
+    def __call__(self, x: B) -> C: ...
+Normal = Callable[[A], D]
+
+a: Call
+b: Normal
+
+reveal_type([a, b])  # E: Revealed type is 'builtins.list[def (__main__.B) -> __main__.B]'
```
JukkaL: Try also `[b, a]`?

test-data/unit/check-protocols.test
```diff
@@ -2307,3 +2327,63 @@ func(call)
...
+class D(B): ...
+
+class Call(Protocol):
+    def __call__(self, __x: C) -> B: ...
+Normal = Callable[[D], A]
+
+def a(x: Call) -> None: ...
+def b(x: Normal) -> None: ...
+
+reveal_type([a, b])  # E: Revealed type is 'builtins.list[def (x: def (__main__.B) -> __main__.B)]'
```
JukkaL: Try also `[b, a]`?

docs/source/protocols.rst
```diff
@@ -409,3 +409,53 @@ in ``typing`` such as ``Iterable``.
    ``isinstance()`` with protocols is not completely safe at runtime.
    For example, signatures of methods are not checked. The runtime
    implementation only checks that all protocol members are defined.
+
+.. _callback_protocols:
+
+Callback protocols
+******************
+
+Protocols can be used to define flexible callback types that are impossible
```
JukkaL: Nit: they are currently possible with the legacy extended callable syntax.

docs/source/protocols.rst
```diff
@@ -409,3 +409,53 @@ in ``typing`` such as ``Iterable``.
...
+to express using the ``Callable[...]`` syntax, for example variadic, overloaded,
```
JukkaL: Style nit: ", for example" -> ", such as" would perhaps read a bit better

docs/source/protocols.rst
```diff
@@ -409,3 +409,53 @@ in ``typing`` such as ``Iterable``.
...
+       ...
+   def bad_cb(*vals: bytes, maxitems: Optional[int]) -> List[bytes]:
+       ...
+
+   batch_proc([], good_cb)  # OK
+   batch_proc([], bad_cb)   # Error! Argument 2 has incompatible type because of
+                            # different name and kind in the callback
+
+Callback protocols and ``Callable[...]`` types can be used interchangeably.
+To indicate an "anonymous" positional-only argument use double underscore,
```
JukkaL: Maybe explicitly mention that keyword argument names in `__call__` methods must be identical, unless a double underscore prefix is used.

docs/source/protocols.rst
ilevkivskyi: Some of them are really impossible, like overloaded callbacks. I will try to clarify here.

ilevkivskyi:
Just not to wait for @JukkaL to return from vacation I am merging this now. All major points have been addressed, if anything minor left, we can fix it later.

PR #5467, opened 2018-08-13
Add machinery for marking places we subvert the type system

msullivan:
We need this for compiling with mypyc, which inserts runtime
typechecks that cause problems when we subvert the type system. So
when compiling with mypyc, we turn those places into Any, while
keeping the types around for normal typechecks.

Since this causes the runtype types to be Any, this is best used
in places where efficient access to properties is not important.
For those cases some other technique should be used.

### DIFF
--- a/mypy/bogus_type.py
+++ b/mypy/bogus_type.py
@@ -1,0 +1,22 @@
+"""A Bogus[T] type alias for marking when we subvert the type system
+
+We need this for compiling with mypyc, which inserts runtime
+typechecks that cause problems when we subvert the type system. So
+when compiling with mypyc, we turn those places into Any, while
+keeping the types around for normal typechecks.
+
+Since this causes the runtype types to be Any, this is best used
+in places where efficient access to properties is not important.
+For those cases some other technique should be used.
+"""
+
+from mypy_extensions import FlexibleAlias
+from typing import TypeVar, Any
+
+T = TypeVar('T')
+
+SUPPRESS_BOGUS_TYPES = False
+if SUPPRESS_BOGUS_TYPES:
+    Bogus = FlexibleAlias[T, Any]
+else:
+    Bogus = FlexibleAlias[T, T]
--- a/mypy/bogus_type.py
+++ b/mypy/bogus_type.py
@@ -5,7 +5,7 @@ typechecks that cause problems when we subvert the type system. So
 when compiling with mypyc, we turn those places into Any, while
 keeping the types around for normal typechecks.
 
-Since this causes the runtype types to be Any, this is best used
+Since this causes the runtime types to be Any, this is best used
 in places where efficient access to properties is not important.
 For those cases some other technique should be used.
 """
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -176,8 +176,11 @@ class SymbolNode(Node):
     # TODO do not use methods for these
 
     @abstractmethod
-    def name(self) -> Bogus[str]: pass
+    def name(self) -> str: pass
 
+    # fullname can often be None even though the type system
+    # disagrees. We mark this with Bogus to let mypyc know not to
+    # worry about it.
     @abstractmethod
     def fullname(self) -> Bogus[str]: pass
 
@@ -197,7 +200,7 @@ class MypyFile(SymbolNode):
     """The abstract syntax tree of a single source file."""
 
     # Module name ('__main__' for initial file)
-    _name = None      # type: Bogus[str]
+    _name = ''      # type: str
     # Fully qualified module name
     _fullname = None  # type: Bogus[str]
     # Path to the file (None if not known)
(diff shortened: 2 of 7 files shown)


msullivan:
This adds a dependency on `mypy_extensions`. Since it needs to use the recently added `FlexibleAlias`, it adds a dependency to a git revision. We probably want to wait until we can do a release of `mypy_extensions` before actually landing this.

mypy/bogus_type.py
```diff
@@ -0,0 +1,22 @@
+"""A Bogus[T] type alias for marking when we subvert the type system
+
+We need this for compiling with mypyc, which inserts runtime
+typechecks that cause problems when we subvert the type system. So
+when compiling with mypyc, we turn those places into Any, while
+keeping the types around for normal typechecks.
+
+Since this causes the runtype types to be Any, this is best used
```
ilevkivskyi: runtype -> runtime?

mypy/nodes.py
```diff
@@ -174,10 +176,10 @@ class SymbolNode(Node):
     # TODO do not use methods for these
 
     @abstractmethod
-    def name(self) -> str: pass
+    def name(self) -> Bogus[str]: pass
```
ilevkivskyi: I would add a comment here (the place `Bogus` first appears in this file) explaining what is going on. My understanding is that there are quite a few places where `name` or `fullname` can be actually `None` while this is not reflected in the type.

Btw, I am surprised you need this for `name` too. I have seen several bugs because of `fullname` being `None`, but I have never encountered a situation where `name` is `None`. Did you see actual problems with it, or you just decided this from analogy with `fullname`?

mypy/nodes.py
```diff
@@ -236,10 +238,10 @@ def __init__(self,
         else:
             self.ignored_lines = set()
 
-    def name(self) -> str:
+    def name(self) -> Bogus[str]:
```
ilevkivskyi: Hm, the above problem is probably only because of `MypyFile`. Can we just fix this instead by always setting `_name` in `__init__`. Also I suppose we can fix `_fullname` for `MypyFile` because the module `id` is always known immediately.

mypy/types.py
```diff
@@ -341,8 +342,8 @@ def accept(self, visitor: 'TypeVisitor[T]') -> T:
         return visitor.visit_any(self)
 
     def copy_modified(self,
-                      type_of_any: TypeOfAny = _dummy,
-                      original_any: Optional['AnyType'] = _dummy,
+                      type_of_any: Bogus[TypeOfAny] = _dummy,
+                      original_any: Bogus[Optional['AnyType']] = _dummy,
```
ilevkivskyi: Maybe add also a _short_ comment about `Bogus` here (just recall that `_dummy` is `Any`)?

mypy/types.py
```diff
@@ -1457,8 +1458,9 @@ class TypeType(Type):
     # a generic class instance, a union, Any, a type variable...
     item = None  # type: Type
 
-    def __init__(self, item: Union[Instance, AnyType, TypeVarType, TupleType, NoneTyp,
-                                   CallableType], *, line: int = -1, column: int = -1) -> None:
+    def __init__(self, item: Bogus[Union[Instance, AnyType, TypeVarType, TupleType, NoneTyp,
+                                         CallableType]], *,
+                 line: int = -1, column: int = -1) -> None:
```
ilevkivskyi: I am a bit surprised the `Bogus` is needed here, and not in `make_normalized`, where the actual hack happens. But most probably I am just missing something.

msullivan: The crash does indeed happen in `make_normalized`, when it tries to coerce the arguments to the type required by `__init__`

mypy/nodes.py
msullivan: I saw actual problems, but they can be easily fixed by setting a default _name.

msullivan: To the empty string, that is. `MypyFile` appears to never have a real name set?

ilevkivskyi: I would prefer something like `id.split('.')[-1]`, if it is possible of course.

ilevkivskyi: (for `MypyFile`s)

mypy/nodes.py
```diff
@@ -195,9 +200,9 @@ class MypyFile(SymbolNode):
     """The abstract syntax tree of a single source file."""
 
     # Module name ('__main__' for initial file)
-    _name = None      # type: str
+    _name = ''      # type: str
     # Fully qualified module name
-    _fullname = None  # type: str
+    _fullname = None  # type: Bogus[str]
```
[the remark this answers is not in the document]
msullivan: It's doable but would require changing some types around and threading the module name into part of the parser where it isn't. Doesn't seem to be an advantage over patching up `_fullname` in semanal like everything else.

`_name`, meanwhile, looks like it never gets set at all, so probably doesn't need to exist.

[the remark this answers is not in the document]
ilevkivskyi: OK.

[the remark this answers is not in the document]
msullivan: Though fixing `name()` to return something useful seems fine also so

PR #5468, opened 2018-08-13
Rework `replace_object_state` to not depend on `__slots__`

msullivan:
Classes from C extension modules (like those generated by mypyc) will
have attributes that are neither from `__dict__` nor `__slots__`.

(`replace_object_state` is used by astmerge to do the merging)

### DIFF
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -188,7 +188,7 @@ def correct_relative_import(cur_mod_id: str,
     return cur_mod_id + (("." + target) if target else ""), ok
 
 
-FIELDS_CACHE = {}  # type: Dict[Type[object], List[str]]
+fields_cache = {}  # type: Dict[Type[object], List[str]]
 
 
 def replace_object_state(new: object, old: object) -> None:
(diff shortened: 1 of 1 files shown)


ilevkivskyi:
I would leave this up to @JukkaL for review (also now there is a merge conflict).

mypy/util.py
```diff
@@ -183,24 +188,39 @@ def correct_relative_import(cur_mod_id: str,
     return cur_mod_id + (("." + target) if target else ""), ok
 
 
+FIELDS_CACHE = {}  # type: Dict[Type[object], List[str]]
```
JukkaL: Style nit: Maybe use lower case name since this the dictionary object gets mutated?

mypy/util.py
```diff
@@ -183,24 +188,39 @@ def correct_relative_import(cur_mod_id: str,
...
-        for base in type(old).__mro__:
-            if '__slots__' in base.__dict__:
-                for attr in getattr(base, '__slots__'):
-                    if hasattr(old, attr):
-                        setattr(new, attr, getattr(old, attr))
-                    elif hasattr(new, attr):
-                        delattr(new, attr)
+
+    cls = old.__class__
+    # Maintain a cache of type -> attributes
```
JukkaL: This doesn't include attributes in `__dict__`. It would be better to explicitly enumerate which attributes are processed here (attributes in `__slots__` and attributes defined in C extension classes?).

PR #5469, opened 2018-08-14
Make FakeInfo a more general-purpose TypeInfo placeholder

msullivan:
There are a lot of places in mypy currently that shove None into
fields expecting TypeInfo. Since mypyc does not appreciate this kind
of behavior, use FakeInfo instances instead of None. FakeInfo is
generalized to allow different assertion messages and to override
`__bool__` so that it can be conveniently checked against.

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -398,10 +398,10 @@ class FuncBase(Node):
         self.unanalyzed_type = None  # type: Optional[mypy.types.Type]
         # If method, reference to TypeInfo
         # TODO: Type should be Optional[TypeInfo]
-        self.info = cast(TypeInfo, None)
-        self.is_property = False
-        self.is_class = False
-        self.is_static = False
+        self.info = FUNC_NO_INFO  # type: TypeInfo
+        self.is_property = False  # type: bool
+        self.is_class = False  # type: bool
+        self.is_static = False  # type: bool
         # Name with module prefix
         # TODO: Type should be Optional[str]
         self._fullname = cast(str, None)
@@ -727,7 +727,7 @@ class Var(SymbolNode):
         # TODO: Should be Optional[str]
         self._fullname = cast(str, None)  # Name with module prefix
         # TODO: Should be Optional[TypeInfo]
-        self.info = cast(TypeInfo, None)  # Defining class (for member variables)
+        self.info = VAR_NO_INFO  # type: TypeInfo
         self.type = type  # type: Optional[mypy.types.Type] # Declared or inferred type, or None
         # Is this the first argument to an ordinary method (usually "self")?
         self.is_self = False
@@ -807,6 +807,7 @@ class ClassDef(Statement):
         self.type_vars = type_vars or []
         self.base_type_exprs = base_type_exprs or []
         self.removed_base_type_exprs = []
+        self.info = CLASSDEF_NO_INFO
         self.metaclass = metaclass
         self.decorators = []
         self.keywords = OrderedDict(keywords or [])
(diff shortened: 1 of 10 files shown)


mypy/nodes.py
```diff
@@ -398,10 +398,10 @@ def __init__(self) -> None:
...
         # If method, reference to TypeInfo
         # TODO: Type should be Optional[TypeInfo]
-        self.info = cast(TypeInfo, None)
-        self.is_property = False
-        self.is_class = False
-        self.is_static = False
+        self.info = FUNC_NO_INFO  # type: TypeInfo
+        self.is_property = False  # type: bool
+        self.is_class = False  # type: bool
+        self.is_static = False  # type: bool
```
ilevkivskyi: Why do you need these three `bool`s here?

mypy/nodes.py
```diff
@@ -2184,6 +2185,7 @@ def __init__(self, names: 'SymbolTable', defn: ClassDef, module_name: str) -> No
         self.assuming_proper = []
         self.inferring = []
         self.add_type_vars()
+        self.replaced = TYPEINFO_NO_REPLACED
```
ilevkivskyi: I would actually make this `Optional`, it is used in few places, so should be easy to fix. Also I think we have an issue about removing `replaced` altogether.

mypy/nodes.py
```diff
@@ -2660,8 +2679,10 @@ def fullname(self) -> Optional[str]:
     @property
     def type(self) -> 'Optional[mypy.types.Type]':
         node = self.node
-        if ((isinstance(node, Var) or isinstance(node, FuncBase))
-                and node.type is not None):
+        if (isinstance(node, Var) and node.type is not None):
+            return node.type
+        # mypy thinks this branch is unreachable but it is wrong
+        elif (isinstance(node, FuncBase) and node.type is not None):
```
ilevkivskyi: Hm, is this a bug then? Do we have an issue for this?

mypy/subtypes.py
```diff
@@ -560,6 +560,7 @@ def find_node_type(node: Union[Var, FuncBase], itype: Instance, subtype: Type) -
         else:
             typ = signature
     itype = map_instance_to_supertype(itype, node.info)
+    assert typ is not None
```
ilevkivskyi: This seems to be redundant, could you please add a comment why this is needed?

mypy/typeanal.py
```diff
@@ -413,7 +413,8 @@ def visit_callable_type(self, t: CallableType, nested: bool = True) -> Type:
                 variables = self.bind_function_type_variables(t, t)
             ret = t.copy_modified(arg_types=self.anal_array(t.arg_types, nested=nested),
                                   ret_type=self.anal_type(t.ret_type, nested=nested),
-                                  fallback=t.fallback or self.named_type('builtins.function'),
+                                  fallback=(t.fallback if t.fallback.type
```
ilevkivskyi: This pattern is a bit confusing, I would add a little remainder here that `FakeInfo` is `False`y. (maybe also below?)

mypy/nodes.py
msullivan: Because referring to `FUNC_NO_INFO`, which is defined later in the file, causes `__init__` to get second-pass typechecked, which causes these to not get inferred right or something.
Annotating the FakeInfos fixes this and the weird `subtype.py` assert also so I'm doing that instead.

mypy/subtypes.py
msullivan: It's a knock-on effect of Var's `__init__` getting typechecked in the second pass I think. But it is pretty bogus and probably a mypy bug, but I don't think I have the cycles right now to track it down. Annotating the FakeInfos fixes the problem, though. SIgh.

mypy/nodes.py
msullivan: It's the "intersection types bug". #3603. I'll point to it in the comments

mypy/nodes.py
msullivan: Oh, that was easy.

PR #5470, opened 2018-08-14
Start moving things mypyc won't support to a new file

msullivan:
There are a handful of features that mypyc is unlikely to properly support soon (if ever), like inheriting from non-object built-in classes and having nontrivial metaclasses. Start moving classes that do this out into a `mypyc_hacks` module.

So far this moves files from `util`, `nodes`, and `types`. (All of which I have working under mypyc!)

### DIFF
--- a/mypy/mypyc_hacks.py
+++ b/mypy/mypyc_hacks.py
@@ -1,0 +1,8 @@
+"""Stuff that we had to move out of its right place because of mypyc limitations."""
+
+# Moved from util.py, because it inherits from Exception
+class DecodeError(Exception):
+    """Exception raised when a file cannot be decoded due to an unknown encoding type.
+
+    Essentially a wrapper for the LookupError raised by `bytearray.decode`
+    """
--- a/mypy/mypyc_hacks.py
+++ b/mypy/mypyc_hacks.py
@@ -35,3 +35,59 @@ class TypeOfAny(Enum):
     from_another_any = 'from_another_any'
     # Does this Any come from an implementation limitation/bug?
     implementation_artifact = 'implementation_artifact'
+
+
+# Moved from nodes.py, because it inherits from dict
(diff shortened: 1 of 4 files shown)


ethanhs:
>There are a handful of features that mypyc is unlikely to properly support soon (if ever), like inheriting from non-object built-in classes and having nontrivial metaclasses. Start moving classes that do this out into a mypyc_hacks module.

This seems... less than ideal. It makes it a lot harder to work on things if I have to remember some nodes aren't actually in `mypy.nodes`.

Perhaps mypyc could work around unsupported items by generating a Python wrapper to the C extension that includes the unsupported items? (Likely replacing the existing module). So nodes.py would included unsupported constructs, while it would import everything else from `_nodes` (the generated C extension)? That way existing imports (theoretically) won't be broken.

msullivan:
It's not great, but I don't think it ought to make things all that much harder when working on the code? All of the classes are reexported from their original locations, and a quick search in those files will find the import from `mypyc_hacks`.

Automatically generating a wrapper is a good idea that I think will probably work, though I'm not sure that the badness of this problem justifies prioritizing as a prereq to bootstrapping

ethanhs:
>It's not great, but I don't think it ought to make things all that much harder when working on the code? All of the classes are reexported from their original locations, and a quick search in those files will find the import from mypyc_hacks.

I was more concerned with code editors, but it seems that mine can handle this, so I don't have as strong objections :)

>Automatically generating a wrapper is a good idea that I think will probably work, though I'm not sure that the badness of this problem justifies prioritizing as a prereq to bootstrapping

Okay, fair enough.

JukkaL:
A potential fix would be to allow defining normal Python classes in mypyc (i.e. not extension classes). These would be slower but could support metaclasses, subclassing built-in types and other such things. This is not a priority, though.

PR #5472, opened 2018-08-14
Always initialize mro in TypeInfo

msullivan:
Currently the mro is left `None` to cause accesses to it before it is
properly initialized to blow up more quickly. This sort of behavior
doesn't play well with mypyc, and leaving it actually uninitialized
causes a lot of blowups.

Just always initialize mro to an empty list.


msullivan:
Looks good.

PR #5473, opened 2018-08-14
Some minor cleanups and fixes to work with mypyc

msullivan:


msullivan:
Hm, I'm somewhat confused about why I'm still having trouble importing `ClassVar` after (I think) telling it to get a more recent typing?

ethanhs:
@msullivan the `typing` module in Python 3.5.1 doesn't have `ClassVar` (which I presume you know). However, the `typing` *package* does not supersede the `typing` *module*.

The typing package readme says:
>NOTE: in Python 3.5 and later, the typing module lives in the stdlib, and installing this package has NO EFFECT. To get a newer version of the typing module in Python 3.5 or later, you have to upgrade to a newer Python (bugfix) version. For example, typing in Python 3.6.0 is missing the definition of ‘Type’ – upgrading to 3.6.2 will fix this.

msullivan:
Oh, durr. Thanks!

(Though also I wonder if we *really* need to keep supporting 3.5.1)

ethanhs:
>(Though also I wonder if we really need to keep supporting 3.5.1)

I for one am in favor of this. Originally it was due to Dropbox using Python 3.5.1. However, I believe that is no longer the case. One significant reason to keep it is IIRC on Ubuntu 16.04 `python3 -V'  is `3.5.1`.

mypy/nodes.py
```diff
@@ -1242,8 +1241,6 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
 class ComplexExpr(Expression):
     """Complex literal"""
 
-    value = 0.0j
-
```
ilevkivskyi: So complex numbers are not yet supported in mypyc? I would actually implement them as an exercise in mypyc. (Next exercise is implementing full SSA :-))

msullivan: They should work, but complex literals aren't implemented yet. Sounds good!

PR #5474, opened 2018-08-14
Generalize subtype caches

Michael0x2a:
Currently, the typestate module keeps track of two separate subtype caches: one for regular subtypes, and another for proper subtypes.

This pattern unfortunately does not generalize well: for example, in the upcoming overloads PR, it ends up being useful to perform subtype checks *with promotions disabled*.

The problem is that while performing a subtype check with promotions disabled, we cannot use or store information to the existing `is_subtype` function without introducing subtle bugs and inconsistencies: if `is_subtype` previously decided that int is a subtype of float due to promotions, it wouldn't make sense to return that cached result with promotions disabled.

One potential solution is to just add two more caches to TypeState. However, this solution does not generalize well, especially if we were to add additional variations of subtype checks in the future.

Instead, this pull request modifies TypeState so it can keep track of an arbitrary number of subtype caches, each of which is characterized by a 'SubtypeKind'.

The 'SubtypeKind' is an int, which we obtain by taking the hash of a tuple containing a unique string per each subtype method, along with whatever additional parameters are active during that subtype check.

So now, instead of hard-coding calls to `record_subtype_cache_entry` or `record_proper_subtype_cache_entry` and friends, we can just call `record_subtype_cache_entry` and pass in the appropriate SubtypeKind to store the type info in the correct cache.

Since manually juggling parameters and SubtypeKinds is a bit of a pain, this pull request also refactors the two subtype visitors so they call a private helper method that takes care of propagating the parameters instead of directly calling the `is_subtype` or `is_proper_subtype` top-level methods.

NOTE: Currently, the mypy codebase does not attempt to perform subtype checks with promotions disabled. That means that this PR should not change the actual behavior of mypy, which is why there are no new tests.

(My subsequent PRs will perform subtype checks w/ promotions disabled.)

### DIFF
--- a/mypy/subtypes.py
+++ b/mypy/subtypes.py
@@ -129,11 +129,11 @@ class SubtypeVisitor(TypeVisitor[bool]):
                            ignore_pos_arg_names: bool = False,
                            ignore_declared_variance: bool = False,
                            ignore_promotions: bool = False) -> SubtypeKind:
-        return hash(('subtype',
-                     type_parameter_checker,
-                     ignore_pos_arg_names,
-                     ignore_declared_variance,
-                     ignore_promotions))
+        return ('subtype',
+                type_parameter_checker,
+                ignore_pos_arg_names,
+                ignore_declared_variance,
+                ignore_promotions)
 
     def _lookup_cache(self, left: Instance, right: Instance) -> bool:
         return TypeState.is_cached_subtype_check(self._subtype_kind, left, right)
@@ -194,7 +194,6 @@ class SubtypeVisitor(TypeVisitor[bool]):
             # was an error somewhere.
             if not self.ignore_promotions and left.type.mro is not None:
                 for base in left.type.mro:
-                    # TODO: Also pass recursively ignore_declared_variance
                     if base._promote and self._is_subtype(base._promote, self.right):
                         self._record_cache(left, right)
                         return True
@@ -252,7 +251,7 @@ class SubtypeVisitor(TypeVisitor[bool]):
         elif isinstance(right, Overloaded):
             return all(self._is_subtype(left, item) for item in right.items())
         elif isinstance(right, Instance):
-            return is_subtype(left.fallback, right)
+            return self._is_subtype(left.fallback, right)
         elif isinstance(right, TypeType):
             # This is unsound, we don't check the __init__ signature.
             return left.is_type_obj() and self._is_subtype(left.ret_type, right.item)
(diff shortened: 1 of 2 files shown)


mypy/subtypes.py
```diff
@@ -150,19 +186,17 @@ def visit_instance(self, left: Instance) -> bool:
...
         if isinstance(right, Instance):
-            if TypeState.is_cached_subtype_check(left, right):
+            if self._lookup_cache(left, right):
                 return True
             # NOTE: left.type.mro may be None in quick mode if there
             # was an error somewhere.
-            if left.type.mro is not None:
+            if not self.ignore_promotions and left.type.mro is not None:
                 for base in left.type.mro:
                     # TODO: Also pass recursively ignore_declared_variance
```
ilevkivskyi: I think this TODO can be removed since `_is_subtype` does this automatically.

mypy/subtypes.py
```diff
@@ -106,11 +110,43 @@ class SubtypeVisitor(TypeVisitor[bool]):
...
+    def build_subtype_kind(*,
+                           type_parameter_checker: TypeParameterChecker = check_type_parameter,
+                           ignore_pos_arg_names: bool = False,
+                           ignore_declared_variance: bool = False,
+                           ignore_promotions: bool = False) -> SubtypeKind:
+        return hash(('subtype',
+                     type_parameter_checker,
+                     ignore_pos_arg_names,
+                     ignore_declared_variance,
+                     ignore_promotions))
```
ilevkivskyi: Is the `hash` call really needed here? As I understand it will be used as a key (tuples can also be keys) so the hash will be called under the hood anyway. Or is it a performance trick?

mypy/subtypes.py
```diff
@@ -204,27 +238,24 @@ def visit_type_var(self, left: TypeVarType) -> bool:
...
                 ignore_pos_arg_names=self.ignore_pos_arg_names)
         elif isinstance(right, Overloaded):
-            return all(is_subtype(left, item, self.check_type_parameter,
-                                  ignore_pos_arg_names=self.ignore_pos_arg_names)
-                       for item in right.items())
+            return all(self._is_subtype(left, item) for item in right.items())
         elif isinstance(right, Instance):
-            return is_subtype(left.fallback, right,
-                              ignore_pos_arg_names=self.ignore_pos_arg_names)
+            return is_subtype(left.fallback, right)
```
ilevkivskyi: Why are you using `is_subtype` instead of `self._is_subtype`? If this is some special case, then it deserves a comment.

mypy/subtypes.py
Michael0x2a: Done.

mypy/subtypes.py
Michael0x2a: It was an attempt at optimization, but I forgot to test it before submitting this PR. After testing, it ended up making little-to-no difference, so I ended up just removing the hash.

mypy/subtypes.py
Michael0x2a: Ah whoops -- that was an oversight. Fixed.

Michael0x2a:
I've tried running https://github.com/python/mypy/pull/5476 against our internal codebases, but not this one specifically. I'll start the run + post an update once it finishes.

Michael0x2a:
@ilevkivskyi -- I've confirmed this PR does not introduce any new regressions to our codebases. (I ran the tests before fixing the merge conflict, but I'm pretty sure that change won't affect anything.)

Issue #5136, opened 2018-06-02
Refactor operator method access and protocol checks to use checkmember.py


ilevkivskyi:
Currently code that checks access to dunder operator methods duplicates code in `checkmember.py`. Apart from code duplication, the code for operators is more simplistic and doesn't cover some corner cases. A similar situation is with accessing attributes for protocol checks.

The common pattern in both cases is that we need almost the same logic as in `checkmember.py`, but with some steps skipped and error reporting suppressed.

JukkaL:
One concrete difference is that operator methods on type objects (as opposed to instances) have different semantics -- they can't generally be defined as class/static methods (and there's `__class_getitem__` in Python 3.7).

Michael0x2a:
Just to add some extra details/clues: I tried removing that call in my recent [operators refactor](https://github.com/python/mypy/pull/5475), and it ended up breaking the `pythoneval.testMetaclassOpAccessAny` test case to fail -- I added a TODO about this to [the relevant section of the code](https://github.com/python/mypy/blob/master/mypy/checkexpr.py#L1874).

PR #5475, opened 2018-08-14
Refactor reversible operators

Michael0x2a:
This pull request refactors and reworks how we handle reversible operators like `__add__`.

Specifically, what our code was previously doing was assuming that given the expression `A() + B()`, we would always try calling `A().__add__(B())` first, followed by `B().__radd__(A())` second (if the `__radd__` method exists).

Unfortunately, it seems like this model was a little too naive, which caused several mismatches/weird errors when I was working on refining how we handle overlaps and TypeVars in a subsequent PR.

Specifically, what actually happens is that...

1. When doing `A() + A()`, we only ever try calling `A.__add__`, never `A.__radd__`. This is the case even if `__add__` is undefined.

2. If `B` is a subclass of `A`, and if `B` defines an `__radd__` method, and we do `A() + B()`, Python will actually try checking `B.__radd__` *first*, then `A.__add__` second.

   This lets a subclass effectively "refine" the desired return type.

   Note that if `B` only *inherits* an `__radd__` method, Python calls `A.__add__` first as usual. Basically, `B` must provide a genuine refinement over whatever `A` returns.

3. In all other cases, we call `__add__` then `__radd__` as usual.

This pull request modifies both checker.py and checkexpr.py to match this behavior, and adds logic so that we check the calls in the correct order.

This ended up slightly changing a few error messages in certain edge cases.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1756,8 +1756,8 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 # are just to verify whether something is valid typing wise).
                 local_errors = self.msg.copy()
                 local_errors.disable_count = 0
-                sub_result, method_type = self.check_op_local('__contains__', right_type,
-                                                          left, e, local_errors)
+                sub_result, method_type = self.check_op_local_by_name('__contains__', right_type,
+                                                                      left, e, local_errors)
                 if isinstance(right_type, PartialType):
                     # We don't really know if this is an error or not, so just shut up.
                     pass
@@ -1808,8 +1808,12 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         else:
             return nodes.op_methods[op]
 
-    def check_op_local(self, method: str, base_type: Type, arg: Expression,
-                       context: Context, local_errors: MessageBuilder) -> Tuple[Type, Type]:
+    def check_op_local_by_name(self,
+                               method: str,
+                               base_type: Type,
+                               arg: Expression,
+                               context: Context,
+                               local_errors: MessageBuilder) -> Tuple[Type, Type]:
         """Type check a binary operation which maps to a method call.
 
         Return tuple (result type, inferred operator method type).
@@ -1817,12 +1821,24 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         method_type = analyze_member_access(method, base_type, context, False, False, True,
                                             self.named_type, self.not_ready_callback, local_errors,
                                             original_type=base_type, chk=self.chk)
+        return self.check_op_local_by_type(method_type, base_type, arg, context, local_errors)
+
+    def check_op_local_by_type(self,
+                               method_type: Type,
+                               base_type: Type,
+                               arg: Expression,
+                               context: Context,
+                               local_errors: MessageBuilder) -> Tuple[Type, Type]:
+        """Type check a binary operation using the (assumed) type of the operator method.
+
+        Return tuple (result type, inferred operator method type).
+        """
         callable_name = None
         object_type = None
         if isinstance(base_type, Instance):
             # TODO: Find out in which class the method was defined originally?
             # TODO: Support non-Instance types.
-            callable_name = '{}.{}'.format(base_type.type.fullname(), method)
+            callable_name = '{}.{}'.format(base_type.type.fullname(), method_type)
             object_type = base_type
         return self.check_call(method_type, [arg], [nodes.ARG_POS],
                                context, arg_messages=local_errors,
@@ -1852,9 +1868,17 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         def lookup_operator(op_name: str, base_type: Type) -> Optional[Type]:
             """Looks up the given operator and returns the corresponding type,
             if it exists."""
+            local_errors = make_local_errors()
+
+            # TODO: Remove this call and rely just on analyze_member_access
+            # Currently, it seems we still need this to correctly deal with
+            # things like plugins and metaclasses?
+            #
+            # E.g. see the check-expressions.testIntPow and pythoneval.testMetaclassOpAccessAny
+            # test cases.
             if not self.has_member(base_type, op_name):
                 return None
-            local_errors = make_local_errors()
+
             member = analyze_member_access(
                 name=op_name,
                 typ=base_type,
@@ -1884,11 +1908,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
             If the attr name is not present in the given class or its MRO, returns None.
             """
-            mro = typ.type.mro
-            if mro is None:
-                return None
-
-            for cls in mro:
+            for cls in typ.type.mro:
                 if cls.names.get(attr_name):
                     return cls.fullname()
             return None
@@ -1904,13 +1924,10 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             any_type = AnyType(TypeOfAny.from_another_any, source_any=right_type)
             return any_type, any_type
 
-        rev_op_name = self.get_reverse_op_method(op_name)
-
         # STEP 1:
         # We start by getting the __op__ and __rop__ methods, if they exist.
 
-        # Records the method type, the base type, and the argument.
-        variants_raw = []  # type: List[Tuple[Optional[Type], Type, Expression]]
+        rev_op_name = self.get_reverse_op_method(op_name)
 
         left_op = lookup_operator(op_name, left_type)
         right_op = lookup_operator(rev_op_name, right_type)
@@ -1919,6 +1936,9 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         # We figure out in which order Python will call the operator methods. As it
         # turns out, it's not as simple as just trying to call __op__ first and
         # __rop__ second.
+        #
+        # We store the determined order inside the 'variants_raw' variable,
+        # which records tuples containing the method, base type, and the argument.
 
         warn_about_uncalled_reverse_operator = False
         bias_right = is_proper_subtype(right_type, left_type)
@@ -1928,10 +1948,10 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             #
             # This is the case even if the __add__ method is completely missing and the __radd__
             # method is defined.
-            #
-            # We report this error message here instead of in the definition checks
 
-            variants_raw.append((left_op, left_type, right_expr))
+            variants_raw = [
+                (left_op, left_type, right_expr)
+            ]
             if right_op is not None:
                 warn_about_uncalled_reverse_operator = True
         elif (is_subtype(right_type, left_type)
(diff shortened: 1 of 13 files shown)


mypy/nodes.py
```diff
@@ -1506,6 +1508,26 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
...
 }
 
+# Suppose we have some class A. When we do A() + A(), Python will only check
+# the output of A().__add__(A()) and skip calling the __radd__ method entirely.
+# This shortcut is used only for the following methods:
+op_methods_that_shortcut = {
+    '__add__',
+    '__sub__',
+    '__mul__',
+    '__truediv__',
```
JelleZijlstra: Should this include `__div__` in Python 2?

mypy/checkexpr.py
```diff
@@ -1840,89 +1828,234 @@ def check_op_local(self, method: str, base_type: Type, arg: Expression,
...
+            local_errors.disable_count = 0
+            if unions_present:
+                local_errors.disable_type_names += 1
+            return local_errors
+
+        def lookup_operator(op_name: str, base_type: Type) -> Optional[Type]:
+            """Looks up the given operator and returns the corresponding type,
+            if it exists."""
+            if not self.has_member(base_type, op_name):
+                return None
```
ilevkivskyi: I would skip this step at all. It caused several bugs in the past, i think we can just use `analyze_member_access `.

mypy/checkexpr.py
```diff
@@ -1840,89 +1828,234 @@ def check_op_local(self, method: str, base_type: Type, arg: Expression,
...
+            So if class A defines foo and class B subclasses A, running
+            'get_class_defined_in(B, "foo")` would return the full name of A.
+
+            However, if B were to override and redefine foo, that method call would
+            return the full name of B instead.
+
+            If the attr name is not present in the given class or its MRO, returns None.
+            """
+            mro = typ.type.mro
+            if mro is None:
```
ilevkivskyi: `mro` can't be `None` anymore.

mypy/checkexpr.py
```diff
@@ -1840,89 +1828,234 @@ def check_op_local(self, method: str, base_type: Type, arg: Expression,
...
+
+        errors = []
+        results = []
+        for method, obj, arg in variants:
+            local_errors = make_local_errors()
+
+            callable_name = None  # type: Optional[str]
+            if isinstance(obj, Instance):
+                # TODO: Find out in which class the method was defined originally?
+                # TODO: Support non-Instance types.
```
ilevkivskyi: Do we have any non-`Instance` types that have a name? Could you please clarify this TODO item?

mypy/checkexpr.py
```diff
@@ -1840,89 +1828,234 @@ def check_op_local(self, method: str, base_type: Type, arg: Expression,
...
+            any_type = AnyType(TypeOfAny.from_another_any, source_any=right_type)
+            return any_type, any_type
+
+        rev_op_name = self.get_reverse_op_method(op_name)
+
+        # STEP 1:
+        # We start by getting the __op__ and __rop__ methods, if they exist.
+
+        # Records the method type, the base type, and the argument.
+        variants_raw = []  # type: List[Tuple[Optional[Type], Type, Expression]]
```
ilevkivskyi: I would actually skip this, and just write
```
variants_raw = [(a, b),
                (c, d)]
```
in each of three branches below. The current pattern makes an impression (this is what happened to me) that the branches are not mutually exclusive (while they are).

mypy/checkexpr.py
```diff
@@ -1840,89 +1828,234 @@ def check_op_local(self, method: str, base_type: Type, arg: Expression,
...
+            cmp_method = nodes.comparison_fallback_method
+            left_cmp_op = lookup_operator(cmp_method, left_type)
+            right_cmp_op = lookup_operator(cmp_method, right_type)
+
+            if bias_right:
+                variants_raw.append((right_cmp_op, right_type, left_expr))
+                variants_raw.append((left_cmp_op, left_type, right_expr))
+            else:
+                variants_raw.append((left_cmp_op, left_type, right_expr))
+                variants_raw.append((right_cmp_op, right_type, left_expr))
```
ilevkivskyi: Here appending is actually good.

mypy/checkexpr.py
```diff
@@ -1840,89 +1828,234 @@ def check_op_local(self, method: str, base_type: Type, arg: Expression,
...
+
+            callable_name = None  # type: Optional[str]
+            if isinstance(obj, Instance):
+                # TODO: Find out in which class the method was defined originally?
+                # TODO: Support non-Instance types.
+                callable_name = '{}.{}'.format(obj.type.fullname(), op_name)
+
+            result = self.check_call(method, [arg], [nodes.ARG_POS],
+                                   context, arg_messages=local_errors,
+                                   callable_name=callable_name, object_type=obj)
```
ilevkivskyi: Can this be refactored to use `check_op_local `, it looks like there is some code (and TODO comments) duplication.

mypy/checkexpr.py
```diff
@@ -1840,89 +1828,234 @@ def check_op_local(self, method: str, base_type: Type, arg: Expression,
...
+                errors.append(local_errors)
+                results.append(result)
+            else:
+                return result
+
+        # STEP 4b:
+        # Sometimes, the variants list is empty. In that case, we fall-back to attempting to
+        # call the __op__ method (even though it's missing).
+
+        if len(errors) == 0:
```
ilevkivskyi: Why checking the length of errors instead of just length of variants?

Also, `if len(seq) == 0` should be `if not seq`.

mypy/checkexpr.py
```diff
@@ -1840,89 +1828,234 @@ def check_op_local(self, method: str, base_type: Type, arg: Expression,
...
+            local_errors = make_local_errors()
+            result = self.check_op_local(op_name, left_type, right_expr, context, local_errors)
+
+            if local_errors.is_errors():
+                errors.append(local_errors)
+                results.append(result)
+            else:
+                return result
```
ilevkivskyi: I am not sure I understand how can this ever succeed, could you please add a comment with an example?

mypy/messages.py
```diff
@@ -994,6 +994,22 @@ def overloaded_signatures_ret_specific(self, index: int, context: Context) -> No
         self.fail('Overloaded function implementation cannot produce return type '
                   'of signature {}'.format(index), context)
 
+    def reverse_operator_method_never_called(self,
+                                             op: str,
+                                             forward_method: str,
+                                             reverse_type: Type,
+                                             reverse_method: str,
+                                             context: Context) -> None:
+        msg = "{rfunc} will not be called when running '{cls} {op} {cls}': must define {ffunc}"
```
ilevkivskyi: I would say "evaluating" instead of "running" here.

test-data/unit/check-classes.test
```diff
@@ -1702,6 +1729,77 @@ class C:
...
+class B(A):
+    def __radd__(self, other: A) -> str: ...  # E: Signatures of "__radd__" of "B" and "__add__" of "A" are unsafely overlapping
+
+# However, if B is a subtype of A, we try calling __radd__ first.
+reveal_type(A() + B())  # E: Revealed type is 'builtins.str'
+
+[case testReverseOperatorOrderingCase5]
+class A:
+    def __add__(self, other: B) -> int: ...
+    def __radd__(self, other: A) -> str: ...
```
ilevkivskyi: Just do double check: this is not an unsafe overlap because only one of the two can be ever called? Could you please add a clarifying comment here?

test-data/unit/check-classes.test
```diff
@@ -1702,6 +1729,77 @@ class C:
...
+
+class B(A):
+    def __radd__(self, other: A) -> str: ...  # E: Signatures of "__radd__" of "B" and "__add__" of "A" are unsafely overlapping
+
+class C(B): pass
+
+# A refinement made by a parent also counts
+reveal_type(A() + C())  # E: Revealed type is 'builtins.str'
+
+
```
ilevkivskyi: Could you please add _few_ tests for operator methods that involve generic classes and overloaded methods (just for completeness).

mypy/checkexpr.py
Michael0x2a: Unfortunately, removing this check ended up breaking a few test cases -- namely testIntPow, which tests our pow plugin, and testMetaclassOpAccessAny, which tests operators + metaclasses. ([Logs here, if you're curious](https://gist.github.com/Michael0x2a/1d5a5bcb7f55c940874808e3712165cb))

I didn't really want to dive into either method in this PR, so I left a TODO for now. (I can also file an issue/add info to an existing issue if you want.)

mypy/checkexpr.py
Michael0x2a: I originally added that out just in case, but it seems we do fall into that case at least once -- for example, adding an assert there makes [this test case fail](https://gist.github.com/Michael0x2a/0822390ad828329d80400eb959b5d7d6).

I added a TODO.

test-data/unit/check-classes.test
Michael0x2a: Yep. I added a comment.

test-data/unit/check-classes.test
Michael0x2a: I added a few tests for overloads. 

I'm not sure if I want to add tests for generics here though, since the way operators handle generics are still broken in this PR. I fixed this in my follow-up PR though, and added a few test cases there.

mypy/checkexpr.py
```diff
@@ -1806,123 +1808,278 @@ def get_operator_method(self, op: str) -> str:
...
+        def lookup_operator(op_name: str, base_type: Type) -> Optional[Type]:
+            """Looks up the given operator and returns the corresponding type,
+            if it exists."""
+            local_errors = make_local_errors()
+
+            # TODO: Remove this call and rely just on analyze_member_access
+            # Currently, it seems we still need this to correctly deal with
+            # things like metaclasses?
+            #
+            # E.g. see the pythoneval.testMetaclassOpAccessAny test case.
```
ilevkivskyi: Could you please add details about this to https://github.com/python/mypy/issues/5136?

mypy/checkexpr.py
```diff
@@ -1806,123 +1808,278 @@ def get_operator_method(self, op: str) -> str:
...
+                results.append(result)
+            else:
+                # In theory, we should never enter this case, but it seems
+                # we sometimes do, when dealing with Type[...]? E.g. see
+                # check-classes.testTypeTypeComparisonWorks.
+                #
+                # This is probably related to the TODO in lookup_operator(...)
+                # up above.
+                #
+                # TODO: Remove this extra case
```
ilevkivskyi: I think this is something deserving a follow-up issue. Could you please open one?

test-data/unit/check-classes.test
```diff
@@ -1702,6 +1729,179 @@ class C:
...
+class B2:
+    def __rtruediv__(self, x):  # E: Signatures of "__rtruediv__" of "B2" and "__truediv__" of "A2" are unsafely overlapping
+        # type: (A2) -> str
+        pass
+
+# That said, mypy currently doesn't handle the actual division operation very
+# gracefully -- it doesn't correctly switch to using __truediv__ when
+# 'from __future__ import division' is included, it doesn't display a very
+# graceful error if __div__ is missing but __truediv__ is present...
+# Also see https://github.com/python/mypy/issues/2048
```
ilevkivskyi: Could you please add these details to the issue and raise its priority to at least normal? This future import is very important in loads of numeric code.

mypy/checkexpr.py
```diff
@@ -1806,123 +1808,278 @@ def get_operator_method(self, op: str) -> str:
...
 
         Return tuple (result type, inferred operator method type).
         """
         method_type = analyze_member_access(method, base_type, context, False, False, True,
                                             self.named_type, self.not_ready_callback, local_errors,
                                             original_type=base_type, chk=self.chk)
+        return self.check_op_local(method, method_type, base_type, arg, context, local_errors)
+
+    def check_op_local(self,
+                       method_name,
```
ilevkivskyi: `method_name` should be `str`.

PR #5476, opened 2018-08-14
Add partial overload checks

Michael0x2a:
This pull request adds more robust support for detecting partially overlapping types. Specifically, it detects overlaps with...

1. TypedDicts
2. Tuples
3. Unions
4. TypeVars
5. Generic types containing variations of the above.

It also swaps out the code for detecting overlaps with operators and removes some associated (and now unused) code.

This pull request builds on top of https://github.com/python/mypy/pull/5474 and https://github.com/python/mypy/pull/5475 -- once those two PRs are merged, I'll rebase this diff if necessary.

This pull request also supercedes https://github.com/python/mypy/pull/5475 -- that PR contains basically the same code as these three PRs, just smushed together.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -471,9 +471,6 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                         if is_unsafe_overlapping_overload_signatures(sig1, sig2):
                             self.msg.overloaded_signatures_overlap(
                                 i + 1, i + j + 2, item.func)
-                        elif is_unsafe_partially_overlapping_overload_signatures(sig1, sig2):
-                            self.msg.overloaded_signatures_partial_overlap(
-                                i + 1, i + j + 2, item.func)
 
             if impl_type is not None:
                 assert defn.impl is not None
@@ -1157,7 +1154,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             first = forward_tweaked
             second = reverse_tweaked
 
-        return is_unsafe_partially_overlapping_overload_signatures(first, second)
+        return is_unsafe_overlapping_overload_signatures(first, second)
 
     def check_inplace_operator_method(self, defn: FuncBase) -> None:
         """Check an inplace operator method such as __iadd__.
@@ -3499,7 +3496,8 @@ def conditional_type_map(expr: Expression,
                 # Expression is always of one of the types in proposed_type_ranges
                 return {}, None
             elif not is_overlapping_types(current_type, proposed_type,
-                                          prohibit_none_typevar_overlap=True):
+                                          prohibit_none_typevar_overlap=True,
+                                          default_return=False):
                 # Expression is never of any type in proposed_type_ranges
                 return None, {}
             else:
@@ -3715,46 +3713,7 @@ def are_argument_counts_overlapping(t: CallableType, s: CallableType) -> bool:
 
 def is_unsafe_overlapping_overload_signatures(signature: CallableType,
                                               other: CallableType) -> bool:
-    """Check if two overloaded signatures are unsafely overlapping, ignoring partial overlaps.
-
-    We consider two functions 's' and 't' to be unsafely overlapping if both
-    of the following are true:
-
-    1.  s's parameters are all more precise or partially overlapping with t's
-    2.  s's return type is NOT a subtype of t's.
-
-    This function will perform a modified version of the above two checks:
-    we do not check for partial overlaps. This lets us vary our error messages
-    depending on the severity of the overlap.
-
-    See 'is_unsafe_partially_overlapping_overload_signatures' for the full checks.
-
-    Assumes that 'signature' appears earlier in the list of overload
-    alternatives then 'other' and that their argument counts are overlapping.
-    """
-    # if "foo" in signature.name or "bar" in signature.name or "chain_call" in signature.name:
-    #    print("in first")
-
-    signature = detach_callable(signature)
-    other = detach_callable(other)
-
-    return (is_callable_compatible(signature, other,
-                                  is_compat=is_more_precise_no_promote,
-                                  is_compat_return=lambda l, r: not is_subtype_no_promote(l, r),
-                                  ignore_return=False,
-                                  check_args_covariantly=True,
-                                  allow_partial_overlap=True) or
-            is_callable_compatible(other, signature,
-                                   is_compat=is_more_precise_no_promote,
-                                   is_compat_return=lambda l, r: not is_subtype_no_promote(r, l),
-                                   ignore_return=False,
-                                   check_args_covariantly=False,
-                                   allow_partial_overlap=True))
-
-
-def is_unsafe_partially_overlapping_overload_signatures(signature: CallableType,
-                                                        other: CallableType) -> bool:
-    """Check if two overloaded signatures are unsafely overlapping, ignoring partial overlaps.
+    """Check if two overloaded signatures are unsafely overlapping or partially overlapping.
 
     We consider two functions 's' and 't' to be unsafely overlapping if both
     of the following are true:
@@ -3765,9 +3724,6 @@ def is_unsafe_partially_overlapping_overload_signatures(signature: CallableType,
     Assumes that 'signature' appears earlier in the list of overload
     alternatives then 'other' and that their argument counts are overlapping.
     """
-    # if "foo" in signature.name or "bar" in signature.name or "chain_call" in signature.name:
-    #    print("in second")
-
     def is_more_precise_or_partially_overlapping(t: Type, s: Type) -> bool:
         return is_more_precise_no_promote(t, s) or is_overlapping_types_no_promote(t, s)
 
@@ -3817,7 +3773,6 @@ def detach_callable(typ: CallableType) -> CallableType:
     The caller can then unify on all type variables whether or not the callable is originally
     from a class or not."""
     type_list = typ.arg_types + [typ.ret_type]
-    # old_type_list = list(type_list)
 
     appear_map = {}  # type: Dict[str, List[int]]
     for i, inner_type in enumerate(type_list):
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3516,8 +3516,7 @@ def conditional_type_map(expr: Expression,
                 # Expression is always of one of the types in proposed_type_ranges
                 return {}, None
             elif not is_overlapping_types(current_type, proposed_type,
-                                          prohibit_none_typevar_overlap=True,
-                                          default_return=True):
+                                          prohibit_none_typevar_overlap=True):
                 # Expression is never of any type in proposed_type_ranges
                 return None, {}
             else:
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3744,9 +3744,6 @@ def is_unsafe_overlapping_overload_signatures(signature: CallableType,
     Assumes that 'signature' appears earlier in the list of overload
     alternatives then 'other' and that their argument counts are overlapping.
     """
-    def is_more_precise_or_partially_overlapping(t: Type, s: Type) -> bool:
-        return is_more_precise_no_promote(t, s) or is_overlapping_types_no_promote(t, s)
-
     # Try detaching callables from the containing class so that all TypeVars
     # are treated as being free.
     #
@@ -3767,13 +3764,13 @@ def is_unsafe_overlapping_overload_signatures(signature: CallableType,
     # This discrepancy is unfortunately difficult to get rid of, so we repeat the
     # checks twice in both directions for now.
     return (is_callable_compatible(signature, other,
-                                  is_compat=is_more_precise_or_partially_overlapping,
+                                  is_compat=is_overlapping_types_no_promote,
                                   is_compat_return=lambda l, r: not is_subtype_no_promote(l, r),
                                   ignore_return=False,
                                   check_args_covariantly=True,
                                   allow_partial_overlap=True) or
             is_callable_compatible(other, signature,
-                                   is_compat=is_more_precise_or_partially_overlapping,
+                                   is_compat=is_overlapping_types_no_promote,
                                    is_compat_return=lambda l, r: not is_subtype_no_promote(r, l),
                                    ignore_return=False,
                                    check_args_covariantly=False,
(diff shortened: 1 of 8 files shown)


Michael0x2a:
@ilevkivskyi -- So, just to summarize the status of all these PRs in one place...

The following PRs are independent from one another and can be reviewed in any order:

1. https://github.com/python/mypy/pull/5474 (Generalizing subtype caches)
2. https://github.com/python/mypy/pull/5475 (Refactoring reversible operators)
3. https://github.com/python/mypy/pull/5321 (Clean up old overload logic -- this one is low priority)

This PR builds on top of code from 1 and 2. Once those two PRs are merged in, I'll rebase this PR to shrink the diff size.

I've left https://github.com/python/mypy/pull/5280, the old PR for adding partial overlaps, open for now since it has some extra context/probably includes some feedback that I've forgotten to incorporate. Feel free to close it whenever though if you want to minimize clutter -- I can re-find it pretty easily.

I've tested this PR on our internal codebases -- so far, I believe this PR introduces one regression that I've as of yet been unable to reproduce and fix. (Basically, the code is asserting some field is None, and mypy gets confused for some reason).

I'll continue working on tracking that final regression and fixing it over the next few days while I'm waiting on your code reviews, but if you have time to take a look/can figure out how to produce a smaller repo, I'd be extremely grateful.

(And sorry for dumping all of this code on you all of a sudden -- if you want me to break up the above PRs into even smaller ones, let me know, and I'll be happy to do it.)

ilevkivskyi:
> I've left #5280, the old PR for adding partial overlaps, open for now since it has some extra context/probably includes some feedback that I've forgotten to incorporate. Feel free to close it whenever though if you want to minimize clutter -- I can re-find it pretty easily.

I would prefer to close it, just to not get lost :-)

> (And sorry for dumping all of this code on you all of a sudden -- if you want me to break up the above PRs into even smaller ones, let me know, and I'll be happy to do it.)

No problem at all, thanks for working on this. The PR sizes are just right for my taste.

(Btw, there is already a merge conflict, you can probably rebase after merging the subtype cache.)

Michael0x2a:
@ilevkivskyi -- the rebase is done now, fyi

ilevkivskyi:
OK, thanks! It is a bit late here. I will finish this tomorrow morning.

mypy/checker.py
```diff
@@ -3710,37 +3714,155 @@ def are_argument_counts_overlapping(t: CallableType, s: CallableType) -> bool:
...
     of the following are true:
 
     1.  s's parameters are all more precise or partially overlapping with t's
     2.  s's return type is NOT a subtype of t's.
 
+    This function will perform a modified version of the above two checks:
+    we do not check for partial overlaps. This lets us vary our error messages
+    depending on the severity of the overlap.
+
+    See 'is_unsafe_partially_overlapping_overload_signatures' for the full checks.
```
ilevkivskyi: Are there two functions only for more tuned error messages? If yes, then I think it is not worth it. This is already complex part of the code and the error messages aren't much cleaner because of this.

mypy/checker.py
```diff
@@ -3710,37 +3714,155 @@ def are_argument_counts_overlapping(t: CallableType, s: CallableType) -> bool:
...
-    # is unsafe: the parameter types are partially overlapping.
-    #
-    # To fix this, we need to either modify meet.is_overlapping_types or add a new
-    # function and use "is_more_precise(...) or is_partially_overlapping(...)" for the is_compat
-    # checks.
-    #
-    # (We already have a rudimentary implementation of 'is_partially_overlapping', but it only
-    # attempts to handle the obvious cases -- see its docstring for more info.)
+    # if "foo" in signature.name or "bar" in signature.name or "chain_call" in signature.name:
+    #    print("in first")
```
ilevkivskyi: Debugging print?

mypy/checker.py
```diff
@@ -3710,37 +3714,155 @@ def are_argument_counts_overlapping(t: CallableType, s: CallableType) -> bool:
...
+    However, if the callable is from a method and that method is using a class type variable,
+    the callable will not keep track of that type variable since it belongs to the class.
+
+    This function will traverse the callable and find all used type vars and add them to the
+    variables field if it isn't already present.
+
+    The caller can then unify on all type variables whether or not the callable is originally
+    from a class or not."""
+    type_list = typ.arg_types + [typ.ret_type]
+    # old_type_list = list(type_list)
```
ilevkivskyi: Stray comment again?

mypy/meet.py
```diff
@@ -51,155 +56,246 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
...
+            return typ.values
+        else:
+            return [typ.upper_bound]
+    elif isinstance(typ, UnionType):
+        return typ.relevant_items()
+    elif isinstance(typ, Overloaded):
+        # Note: doing 'return typ.items()' makes mypy
+        # infer a too-specific return type of List[CallableType]
+        out = []  # type: List[Type]
+        out.extend(typ.items())
```
ilevkivskyi: Just use `return list(typ.iems())` (i.e. make a copy so that mypy is happy about invariance).

mypy/meet.py
```diff
@@ -51,155 +56,246 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
...
+
+        This function mostly exists so we don't have to repeat keyword arguments everywhere.'''
+        return is_overlapping_types(
+            left, right,
+            ignore_promotions=ignore_promotions,
+            prohibit_none_typevar_overlap=prohibit_none_typevar_overlap)
+
+    # We should never encounter these types, but if we do, we handle
+    # them in the same way we handle 'Any'.
+    illegal_types = (UnboundType, PartialType, ErasedType, DeletedType)
```
ilevkivskyi: `PartialType` should be an `assert False, "Something bad about partial types"`, for other it is OK to be `Any`

mypy/meet.py
```diff
@@ -51,155 +56,246 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
...
-        if isinstance(t, NoneTyp) != isinstance(s, NoneTyp):
-            # NoneTyp does not overlap with other non-Union types under strict Optional checking
             return False
-    # We conservatively assume that non-instance, non-union, non-TupleType and non-TypeType types
-    # can overlap any other types.
+
+        if len(left.args) == len(right.args):
+            for left_arg, right_arg in zip(left.args, right.args):
+                if _is_overlapping_types(left_arg, right_arg):
+                    return True
```
ilevkivskyi: Do we need to care about variance here? Maybe add an example why `List[Union[A, B]]` and `List[Union[B, C]]` cause a danger with overloads?

mypy/meet.py
```diff
@@ -51,155 +56,246 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
...
-            if t.length() == s.length():
-                if all(is_overlapping_types(ti, si, use_promotions)
-                       for ti, si in zip(t.items, s.items)):
-                    return True
-        # TupleType and non-tuples do not overlap
+def are_tuples_overlapping(left: Type, right: Type, *,
+                           ignore_promotions: bool = False) -> bool:
+    """Returns true if left and right are overlapping tuples.
+
+    Precondition: is_tuple(left) and is_tuple(right) are both true."""
```
ilevkivskyi: You already have asserts below, better add an assert message there instead.

test-data/unit/check-isinstance.test
```diff
@@ -2005,3 +2005,81 @@ def f(x: Union[A, str]) -> None:
...
+
+T = TypeVar('T')
+
+def foo(x: T) -> T:
+    out = None
+    out = x
+    if out is None:
+        pass
+    return out
+[builtins fixtures/isinstance.pyi]
```
ilevkivskyi: TBH I am not sure what these two tests test. Maybe add a note or somehow modify them so that it is more clear what is a potential failure?

test-data/unit/check-overloading.test
```diff
@@ -2526,34 +2581,374 @@ class C: ...
...
+
+class Wrapper(Generic[T]):
+    @overload
+    def f(self, x: int) -> str: ...   # E: Overloaded function signatures 1 and 2 are partially overlapping: the two signatures may return incompatible types given certain calls
+    @overload
+    def f(self, x: T) -> T: ...
+    def f(self, x): ...
+
+    # TODO: This shouldn't trigger an error message.
+    # Related to testTypeCheckOverloadImplementationTypeVarDifferingUsage2?
```
ilevkivskyi: Could you please create a follow-up issue for this kind of problems?

mypy/meet.py
```diff
@@ -51,155 +56,246 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
...
+
+        if len(left.args) == len(right.args):
+            for left_arg, right_arg in zip(left.args, right.args):
+                if _is_overlapping_types(left_arg, right_arg):
+                    return True
+
+    # We ought to have handled every case by now: we conclude the
+    # two types are not overlapping, either completely or partially.
+
+    return False
```
ilevkivskyi: I think it would be safer to have different defaults for overloads and `isinstance()` checks. For the former, it is better to have `False` to avoid false positives. But for `isinstance()` it is probably better to have `True` for the cases where it is not clear (to avoid inferring some branches as unreachable, as this may cause problems for mypyc).

mypy/meet.py
```diff
@@ -34,7 +39,7 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
     if isinstance(declared, UnionType):
         return UnionType.make_simplified_union([narrow_declared_type(x, narrowed)
                                                 for x in declared.relevant_items()])
-    elif not is_overlapping_types(declared, narrowed, use_promotions=True):
+    elif not is_overlapping_types(declared, narrowed):
```
ilevkivskyi: Do you have a test for this, or you jut noticed the inconsistency?

mypy/meet.py
Michael0x2a: Could I persuade you to keep this a `return False`? Basically the way `is_overlapping_types` works is by steadily working through each pair of types and comparing them against each other.

So, if the left and the right are both TypedDict, we compare those two directly and return either True or False, if the left and the right are both Instance, we do the same...

This final `return False` case is meant to handle the situation where left and right are genuinely two unrelated types -- returning anything other then False here would lead to incorrect results in those cases.

It's possible that there are some overlap cases we've missed (e.g. can CallableType overlap with Type? Can Type overlap with Instance?), but I'd much rather prefer we handle those missing cases directly and fix `is_overlapping_types` as they come up instead of silently letting the function continue to be imprecise.

> to avoid inferring some branches as unreachable, as this may cause problems for mypyc

I think maybe a better way of solving this is to make mypy report branches that are accidentally unreachable -- it's fairly easy to accidentally add an isinstance check that'll never be reached, I think. (Of course, we wouldn't report branches that are *intentionally* unreachable due to system platform guards or things like 'if False' -- we also probably shouldn't report branches that just contain an assert/just throw an exception).

I'm hoping that such a check could just be enabled by default -- I can't really imagine people intentionally adding logic to a no-op branch.

(I'd be willing to work on adding this sort of check if you think it's valuable enough).

mypy/meet.py
Michael0x2a: Oh, what happened here is that that I changed the behavior of `is_overlapping_types` so it does use promotions by default. So, removing this param isn't actually causing any difference in behavior.

Michael0x2a:
@ilevkivskyi -- ok, sorry for the delay! I can confirm that this is now (finally) ready for a second look, and can also confirm that it triggers no new errors in our internal codebases.

Specific changes I made:

1. Removed debug prints and comments
2. Consolidated the two "unsafe overlapping overload" checks into just a single one, and removed the extraneous whitespace.
3. Finally tracked down and fixed a bug related to Nones, isinstance checks, and no-strict-optional. In sort, the bug was that I was inconsistently filtering out `NoneType` when in non-strict-optional mode.
4. Created an issue (https://github.com/python/mypy/issues/5510) to track the extra error message that we generate when turning a generic overload function into a method.
5. Added a comment explaining why we don't care about variance in `is_overlapping_types`.
6. Added a comment to the two extra test cases I added to `check-isinstance.test` explaining what they do.

I've also left a few responses to your comments up above.

mypy/meet.py
ilevkivskyi: > Could I persuade you to keep this a `return False`?

Not just me :-)

Your proposal with the additional error may generate false positives. Consider this example:
```python
class A: ...
class B: ...
if isinstance(x, A):
    ...
    if isinstance(x, B):
        # this is not reachable ...or is it?
```
I think we already discussed some time ago that there are special cases (like multiple inheritance) where we are deliberately unsafe. All such cases should be treated on per-use-case basis. For example, when checking for overloads, we treat `A` and `B` as non-overlapping, while when inferring reachability of a branch we might want to do the opposite.

Even if you will not add the `default` argument for such corner cases, I am quite sure it will be added later (the above example is unreachable on master, but this may be changed soon).

@msullivan what do you think about this?

ilevkivskyi:
Thanks for the updates! I will make another pass of review tomorrow (most likely).

mypy/meet.py
ilevkivskyi: > maybe a better approach might be to add a `allow_multiple_inheritance` flag to this function?
[snip]
> More broadly, I'd prefer whitelisting specific cases where we want to relax the overlap checks -- I think the code would be more maintainable that way

You think it is more maintainable, I think it looks like "premature maintainability optimization" :-) 

Currently all edge cases should be considered non-overlapping for overloads, but overlapping for inferring reachability. For example, empty containers: `List[int]` and `List[Type[int]]` are considered non-overlapping for overloads but should be overlapping for reachability.

Anyway, it looks like this is an independent issue, so could you please add a TODO item above the final `return False`, and add a permalink to it in https://github.com/python/mypy/issues/3603 ?

Michael0x2a: > Currently all edge cases should be considered non-overlapping for overloads, but overlapping for inferring reachability. For example, empty containers: List[int] and List[Type[int]] are considered non-overlapping for overloads but should be overlapping for reachability.

Well, `List[int]` and `List[Type[int]]` are both represented by Instance objects underneath -- so I think special-casing the "left and right are both Instances" tactic would work?

In any case, I'll make a TODO -- we can worry about this later once it comes up again :)

ilevkivskyi: > so I think special-casing the "left and right are both Instances" tactic would work?

It may be more subtle because it will recurse to checking arguments that are `Instance` vs `TypeType`. Also there are other cases if you want non-instances, e.g. `Instance` vs `CallableType` should be _always_ overlapping for the purpose of `isinstance()` and `callable()` checks (that @msullivan implemented a while ago), but should not be overlapping for overloads.

> In any case, I'll make a TODO -- we can worry about this later once it comes up again :)

OK.

ilevkivskyi:
> How big of a concern is this/do you think this is something we should try testing for first in some way?

I think this is OK. I am not too concerned about rare false negatives, we are already aware of the problem, and the issues are filed. I think you can go ahead and merge this.

PR #5478, opened 2018-08-14
Mention exception for generic type aliases

asottile:


docs/source/kinds_of_types.rst
```diff
@@ -546,7 +546,8 @@ assigning the type to a variable:
 .. note::
 
     A type alias does not create a new type. It's just a shorthand notation for
-    another type -- it's equivalent to the target type.
+    another type -- it's equivalent to the target type except for
+    :ref:`generic-type-aliases`.
```
ilevkivskyi: I would rather format this as `` :ref:`generic aliases <generic-type-aliases>` ``

PR #5487, opened 2018-08-15
More tweaks to support mypyc

msullivan:
mypyc doesn't support accepting `*args` or defining methods via assignment

### DIFF
--- a/mypy/indirection.py
+++ b/mypy/indirection.py
@@ -1,4 +1,4 @@
-from typing import Dict, Iterable, List, Optional, Set
+from typing import Dict, Iterable, List, Optional, Set, Union
 
 from mypy.types import SyntheticTypeVisitor
 import mypy.types as types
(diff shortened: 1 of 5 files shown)


mypy/indirection.py
```diff
@@ -22,9 +22,10 @@ def __init__(self) -> None:
         self.cache = {}  # type: Dict[types.Type, Set[str]]
 
     def find_modules(self, typs: Iterable[types.Type]) -> Set[str]:
-        return self._visit(*typs)
+        return self._visit(typs)
 
-    def _visit(self, *typs: types.Type) -> Set[str]:
+    def _visit(self, typ_or_typs: Union[types.Type, Iterable[types.Type]]) -> Set[str]:
```
JelleZijlstra: Why not just change callers passing a single type to do `self._visit([typ])`, so we don't need the union type?

msullivan: That's what I did at first but I felt that it uglified the code more than I liked. I'd be happy to go back to that though if people think it is nicer.

mypy/server/subexpr.py
```diff
@@ -24,21 +24,50 @@ def __init__(self) -> None:
     def _visit_leaf(self, o: Expression) -> None:
         self.add(o)
```
ilevkivskyi: Can we remove `_visit_leaf` and just use `self.add(o)` directly in all methods you add below?

mypy/indirection.py
ilevkivskyi: I am fine with keeping this a union.

PR #5495, opened 2018-08-17
More tweaks to work with mypyc

msullivan:


mypy/fastparse.py
```diff
@@ -205,7 +209,7 @@ def translate_stmt_list(self, l: Sequence[ast3.AST]) -> List[Statement]:
         ast3.BitXor: '^',
         ast3.BitAnd: '&',
         ast3.FloorDiv: '//'
-    }
+    }  # type: ClassVar[Dict[object, str]]
```
ilevkivskyi: Can this be `Dict[ast3.AST, str]`?

mypy/fastparse.py
```diff
@@ -225,7 +229,7 @@ def from_operator(self, op: ast3.operator) -> str:
         ast3.IsNot: 'is not',
         ast3.In: 'in',
         ast3.NotIn: 'not in'
-    }
+    }  # type: ClassVar[Dict[object, str]]
```
ilevkivskyi: Same question here?

mypy/fastparse.py
```diff
@@ -861,7 +865,9 @@ def is_star2arg(k: ast3.keyword) -> bool:
 
     # Num(object n) -- a number as a PyObject.
     @with_line
-    def visit_Num(self, n: ast3.Num) -> Union[IntExpr, FloatExpr, ComplexExpr]:
+    def visit_Num(self, nx: ast3.Num) -> Union[IntExpr, FloatExpr, ComplexExpr]:
+        # XXX: the typeshed stubs are lies
+        n = nx  # type: Any
```
ilevkivskyi: I think this was recently fixed. Did you sync typeshed? Or if stubs are still bad, then maybe we can just fix them.

mypy/build.py
```diff
@@ -1916,6 +1919,11 @@ def __init__(self,
             self.compute_dependencies()
             self.child_modules = set()
 
+    @property
+    def xmeta(self) -> CacheMeta:
+        assert self.meta
```
ilevkivskyi: Maybe add an assert message here?

mypy/semanal.py
```diff
@@ -2241,7 +2247,8 @@ def process_module_assignment(self, lvals: List[Lvalue], rval: Expression,
             # extra elements, so no error will be raised here; mypy will later complain
             # about the length mismatch in type-checking.
             elementwise_assignments = zip(rval.items, *[v.items for v in seq_lvals])
-            for rv, *lvs in elementwise_assignments:
+            for part in elementwise_assignments:
+                rv, lvs = part[0], list(part[1:])
```
ilevkivskyi: Maybe add a TODO here to rewrite this back to non-ugly form when this will be implemented in mypyc?

mypy/fastparse.py
msullivan: Oh oops, I had meant to remove this. I did indeed just fix the typeshed stub for it.

PR #5499, opened 2018-08-17
Eliminate some conditional method defs in fastparse

msullivan:
It seems like they might not be necessary anymore, and mypyc doesn't
support it.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -1003,7 +1003,13 @@ class TypeConverter(ast3.NodeTransformer):
         self.line = line
         self.node_stack = []  # type: List[ast3.AST]
 
-    def _visit_implementation(self, node: Optional[ast3.AST]) -> Optional[Type]:
+    @overload
+    def visit(self, node: ast3.expr) -> Type: ...
+
+    @overload  # noqa
+    def visit(self, node: Optional[ast3.AST]) -> Optional[Type]: ...
+
+    def visit(self, node: Optional[ast3.AST]) -> Optional[Type]:  # noqa
         """Modified visit -- keep track of the stack of nodes"""
         if node is None:
             return None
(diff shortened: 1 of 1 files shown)


mypy/fastparse.py
```diff
@@ -1008,18 +1006,14 @@ def _visit_implementation(self, node: Optional[ast3.AST]) -> Optional[Type]:
         finally:
             self.node_stack.pop()
 
-    if sys.version_info >= (3, 6):
-        @overload
-        def visit(self, node: ast3.expr) -> Type: ...
+    @overload
+    def visit(self, node: ast3.expr) -> Type: ...
```
ilevkivskyi: This will fail on 3.5.1

mypy/fastparse.py
```diff
@@ -1008,18 +1013,14 @@ def _visit_implementation(self, node: Optional[ast3.AST]) -> Optional[Type]:
...
+    @overload  # noqa
+    def visit(self, node: Optional[ast3.AST]) -> Optional[Type]: ...
 
-        def visit(self, node: Optional[ast3.AST]) -> Optional[Type]:  # noqa
-            return self._visit_implementation(node)
-    else:
-        def visit(self, node: Optional[ast3.AST]) -> Any:
-            return self._visit_implementation(node)
+    def visit(self, node: Optional[ast3.AST]) -> Optional[Type]:  # noqa
+        return self._visit_implementation(node)
```
ilevkivskyi: Also I think `_visit_implementation` only existed because of conditional definition, you can kill it now.

mypy/fastparse.py
```diff
@@ -67,6 +67,13 @@
 TYPE_COMMENT_AST_ERROR = 'invalid type comment or annotation'
 
 
+# Older versions of typing don't allow using overload outside stubs,
+# so provide a dummy.
+if sys.version_info < (3, 6):
```
ilevkivskyi: Shouldn't this be guarded with a big `if not MYPY:`?

msullivan: mypy didn't seem to care, but probably ought to anyways

mypy/fastparse.py
msullivan: :heavy_check_mark:

mypy/fastparse.py
ethanhs: On 3.4, the version of typing is newer than on 3.5, so you lose some type safety on 3.4. It works because you are only making the signature less exact, but it is more correct to use `if MYPY`.

PR #5501, opened 2018-08-18
Fixes #1727: constructor after classmethod counts

### LINKED ISSUE #1727: Constructor defined after classmethod is considered to have no args in the classmethod
The type of `cls` within `bar()` in the following code is `def () -> Foo` unless the `__init__` is moved to before the `classmethod`:

``` python
class Foo(object):

  @classmethod
  def bar(cls):
    # type: () -> Foo
    reveal_type(cls)
    return cls("bar")

  def __init__(self, baz):
    # type: (str) -> None
    self.baz = baz
```

VadimPushtaev:
This is a fix for https://github.com/python/mypy/issues/1727.

Instead of prepare classmethod definitions while analyzing class, let's postpone this until the class is fully analyzed.

`dataclass` magic creates `__init__` after the class is already created, so they fix up all the classmethods. This is no longer needed.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -214,6 +214,10 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
     # postpone_nested_functions_stack[-1] == FUNCTION_FIRST_PHASE_POSTPONE_SECOND.
     postponed_functions_stack = None  # type: List[List[Node]]
 
+    # Classmethod definitions that couldn't be prepared yet
+    # until the class analysis is complete
+    postponed_classmethods_stack = None  # type: List[List[FuncDef]]
+
     loop_depth = 0         # Depth of breakable loops
     cur_mod_id = ''        # Current module id (or None) (phase 2)
     is_stub_file = False   # Are we analyzing a stub file?
@@ -246,6 +250,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
         self.missing_modules = missing_modules
         self.postpone_nested_functions_stack = [FUNCTION_BOTH_PHASES]
         self.postponed_functions_stack = []
+        self.postponed_classmethods_stack = []
         self.all_exports = set()  # type: Set[str]
         self.plugin = plugin
         # If True, process function definitions. If False, don't. This is used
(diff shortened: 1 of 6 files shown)


VadimPushtaev:
Possibly, another solution is to do what `dataclass` does for every class instead of postponing something.

VadimPushtaev:
Also, I don't quite like `self.prepare_postponed_classmethods_signature()` being in the `leave_class` method, which seems to be too high-level for this kind of tasks, but I'm not sure. I could use some advice.

test-data/unit/check-classes.test
```diff
@@ -616,6 +616,18 @@ main:3: error: Incompatible types in assignment (expression has type "B", variab
...
+class C:
+    @classmethod
+    def create(cls, arg: int) -> 'C':
+        return cls(arg)
+
+    def __init__(self, arg: 'int') -> None:
+        self._arg: int = arg
+
+reveal_type(C.create)  # E: Revealed type is 'def (arg: builtins.int) -> __main__.C'
+[builtins fixtures/classmethod.pyi]
```
ilevkivskyi: I am not sure this test actually tests the situation enough. I think what you need to add is a `reveal_type` for `cls` and `cls(<expected args>)` in the method body, _and_ as well check that `cls(<unexpected args>)` fails with a reasonable error.

Please be sure to include all the examples that appeared in the original issue. In addition please add at least tests that check your fix works correctly with the following:
* Fine-grained daemon
* Forward references
* Generic classes
* Overloaded methods (there are known issues, but few simple tests would help)

mypy/semanal.py
```diff
@@ -456,6 +461,15 @@ def _visit_func_def(self, defn: FuncDef) -> None:
                     assert ret_type is not None, "Internal error: typing.Coroutine not found"
                     defn.type = defn.type.copy_modified(ret_type=ret_type)
 
+    def prepare_postponed_classmethods_signature(self) -> None:
+        for func in self.postponed_classmethods_stack[-1]:
+            functype = func.type
+            assert self.type
+            assert functype
+            assert isinstance(functype, CallableType)
```
ilevkivskyi: Why do you need these asserts? Please add assert messages.

VadimPushtaev:
Thanks for the review, I'll try to fix all the issues as soon as possible.

mypy/semanal.py
VadimPushtaev: That was basically to make `mypy` happy of this code. Fixed to make more sense.

VadimPushtaev:
I believe that's all tests we reasonably need.

JukkaL:
Hmm is this now superseded by #5646? Thanks for getting the ball rolling with this anyway!

ilevkivskyi:
Yes, I think now that we have a broader fix, this one is not needed. Closing in favor of #5646 

@VadimPushtaev Thank you for working on this!

Issue #4985, opened 2018-04-28
Signatures of overloaded operators are incompatible with super type


HazardousPeach:
I've noticed this problem, very similar to  #3262 , except with overloaded operators.

Originally, #3262 made it so that code like this:
```
from typing import overload

class A:
    def f(self, x : 'A') -> 'A': ...

class B(A):
    @overload
    def f(self, x : 'A') -> 'A': ...
    @overload
    def f(self, x : 'B') -> 'B' : ...
```

Wouldn't typecheck. 

#3263 appears to have fixed this. Unfortunately, while this works for normal method signatures, it doesn't appear to work for overloaded operators. So the very similar code:

```
from typing import overload

class A:
    def __add__(self, x : 'A') -> 'A': ...

class B(A):
    @overload
    def __add__(self, x : 'A') -> 'A': ...
    @overload
    def __add__(self, x : 'B') -> 'B' : ...
```

Still fails to typecheck, with the error message:
```
test.pyi:7: error: Signature of "__add__" incompatible with supertype "A"
```

I can't think of any reason why this should be a problem with operators, although maybe someone else can explain.

I'm on Python 3.6.3 and MyPy 0.600+dev (master branch from git)

ilevkivskyi:
It looks like this is intentional. The point is `__add__` and `__radd__` needs to be mutually compatible. I found this code that explicitly prohibits overloads for special methods:
```python
        if (not isinstance(original, Overloaded) and
              isinstance(override, Overloaded) and
              name in nodes.reverse_op_methods.keys()):
            # Operator method overrides cannot introduce overloading, as
            # this could be unsafe with reverse operator methods.
            fail = True
```
TBH, I don't see an immediate reason why this is unsafe (`__radd__` doesn't even appear in this example), maybe @JukkaL can clarify this.

Michael0x2a:
I think I might understand what's going on here: I believe mypy is correct in saying that overloads don't necessarily interact in a typesafe way with operator dunder methods. Here's an example: it's similar to the original code snippet, except that I changed the signature of B slightly.

```
from typing import overload

class A:
    def __add__(self, x : 'A') -> 'A': 
        if isinstance(x, A):
            return A()
        else:
            return NotImplemented

class B(A):
    @overload
    def __add__(self, x : 'Other') -> 'B' : ...
    @overload
    def __add__(self, x : 'A') -> 'A': ...
    def __add__(self, x):
        if isinstance(x, Other):
            return B()
[...]

PR #5505, opened 2018-08-19
Allow overloaded overrides if they don't extend domain

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/4985

Mypy previously disallowed introducing overloads for operator methods in overrides because this can interact unsafely with reverse methods. However it looks like this is safe for the situations where the domain of the override is not extended.

In fact the same unsafety exists for non-overloaded op-methods (see example in https://github.com/python/mypy/issues/4985#issuecomment-414095444), but it looks like we are fine with this, I have found few existing tests explicitly testing for this.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1341,6 +1341,20 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                 self.msg.signature_incompatible_with_supertype(
                     defn.name(), name, base.name(), context)
 
+    def get_op_other_domain(self, tp: FunctionLike) -> Optional[Type]:
+        if isinstance(tp, CallableType):
+            if tp.arg_kinds and tp.arg_kinds[0] == ARG_POS:
+                return tp.arg_types[0]
+            return None
+        elif isinstance(tp, Overloaded):
+            raw_items = [self.get_op_other_domain(it) for it in tp.items()]
+            items = [it for it in raw_items if it]
+            if items:
+                return UnionType.make_simplified_union(items)
+            return None
+        else:
+            assert False, "Need to check all FunctionLike subtypes here"
+
     def check_override(self, override: FunctionLike, original: FunctionLike,
                        name: str, name_in_super: str, supertype: str,
                        original_class_or_static: bool,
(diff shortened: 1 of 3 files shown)


mypy/checker.py
```diff
@@ -1341,6 +1341,20 @@ def check_method_override_for_base_with_name(
                 self.msg.signature_incompatible_with_supertype(
                     defn.name(), name, base.name(), context)
 
+    def get_op_other_domain(self, tp: FunctionLike) -> Optional[Type]:
+        if isinstance(tp, CallableType):
+            if tp.arg_kinds and tp.arg_kinds[0] == ARG_POS:
```
Michael0x2a: Do we care what happens if the user does `def __add__(self, *args: Foo) -> Foo` or something?

It's technically not illegal, and I think we even have one or two tests existing tests somewhere that check for this sort of thing in operator methods.

mypy/checker.py
```diff
@@ -1357,15 +1371,18 @@ def check_override(self, override: FunctionLike, original: FunctionLike,
         """
         # Use boolean variable to clarify code.
         fail = False
+        dunder_note = False
```
Michael0x2a: Can we rename this variable to `op_method_wider_note` or something else that makes it clear this is related to operator methods, not dunder methods in general? (E.g. we don't display this note for dunder methods like `__len__` or `__init__`.)

test-data/unit/check-overloading.test
```diff
@@ -4125,6 +4125,59 @@ main:11: error: Argument 7 to "f" has incompatible type "Union[int, str]"; expec
 main:11: error: Argument 8 to "f" has incompatible type "Union[int, str]"; expected "int"
 main:11: note: Not all union combinations were tried because there are too many unions
 
+[case testSafeDunderOverlapInSubclass]
+from typing import overload
+
+class A:
+    def f(self, x : 'A') -> 'A': ...
```
Michael0x2a: Maybe we should also add a unit test to see what happens if the parent class is also defining an overloaded operator? Up to you.

mypy/checker.py
```diff
@@ -1357,15 +1371,18 @@ def check_override(self, override: FunctionLike, original: FunctionLike,
...
         if not is_subtype(override, original, ignore_pos_arg_names=True):
             fail = True
-        elif (not isinstance(original, Overloaded) and
-              isinstance(override, Overloaded) and
-              self.is_forward_op_method(name)):
-            # Operator method overrides cannot introduce overloading, as
+        elif isinstance(override, Overloaded) and self.is_forward_op_method(name):
```
Michael0x2a: While we're at it, do you think we also need to perform this same check for reverse operator methods?

(I'm actually not sure myself. I spent a little bit of time trying to come up with an example but couldn't think of anything, but that's not a particularly convincing argument.)

mypy/checker.py
ilevkivskyi: I think we don't care for now, I would say this kind of situations is already rare, and detecting a rare unsafety due to extended domain is essentially `rare ** 2`.

mypy/checker.py
ilevkivskyi: Good point!

mypy/checker.py
ilevkivskyi: I am not 100% sure, but I think we don't need this for `__rX__`. It seems to me that we can check unsafe overlap only for one, and not extending the domain only for other. This however probably only applies for simple cases where the runtime logic is simple: first check `__X__` if it doesn't work, go to `__rX__`. But then again, the `rare ** 2` argument probably applies here.

test-data/unit/check-overloading.test
ilevkivskyi: Oh wait, it looks like I used `f` instead of `__add__`, I will switch to `__add__`. Btw the test case you mention already exists in other file, so I will just keep this one.

ilevkivskyi: Actually this test gets updated in this PR, it was passing but now fails, see `testOverloadedOperatorMethodOverrideWithNewItem`.

ilevkivskyi:
@Michael0x2a Thanks for review! Are you OK with my responses to comments?

Issue #4227, opened 2017-11-08
spurious "Need type annotation" error with overloads


chadrik:
I'm getting "Need type annotation for variable" errors which I think are the result of a bug with overloads.

Stub file `overload_test.pyi`:

```python
from typing import overload, TypeVar, Any

_T = TypeVar('_T')

@overload
def attr(default: _T = ..., blah: int = ...) -> _T: ...
@overload
def attr(default: Any = ...) -> Any: ...
```

Test program:

```python
from overload_test import attr

a = attr()
b = attr()

reveal_type(a)
reveal_type(b)
```

mypy output:
```
test.py:3: error: Need type annotation for variable
test.py:4: error: Need type annotation for variable
test.py:6: error: Revealed type is 'Any'
test.py:6: error: Cannot determine type of 'a'
test.py:7: error: Revealed type is 'Any'
test.py:7: error: Cannot determine type of 'b'
```

It may be interesting to note that if I swap the order of the overloads, the "Need type annotation for variable" and "Cannot determine type" errors go away.

ethanhs:
This happens because mypy is getting confused between the overload and the TypeVar. I believe it is trying to deduce the TypeVar and cannot, but is not considering the alternative signature. It seems we need to consider other valid overloads if typevar deduction fails on one.

Note that removing the other overload entirely has the same output, thus the second overload is not used at all.

gvanrossum:
It would be best to define three different overloads, with 0, 1 and 2
arguments.

chadrik:
> Note that removing the other overload entirely has the same output, thus the second overload is not used at all.

For my own edification, if we ditch the second overload... 

```python
from typing import overload, TypeVar, Any

_T = TypeVar('_T')

def attr(default: _T = ..., blah: int = ...) -> _T: ...
```

...is the "Need type annotation for variable" error valid?   Isn't the unfilled `TypeVar` equivalent to `Any`?

gvanrossum:
> Isn't the unfilled TypeVar equivalent to Any?

No, it's not as simple as that. It becomes similar to
```py
def attr() -> _T: ...
```
and the type of `attr()` then becomes "the uninhabited type" which is printed as `<nothing>`, but which essentially means "there is no value of this type".

chadrik:
> and the type of attr() then becomes "the uninhabited type"

and I assume that anytime you assign "the uninhabited type" to a variable it generates the "Need type annotation for variable" error?

[...]

PR #5506, opened 2018-08-19
Bounded/constrained type variables don't shadow Any in overloads

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/4227

The problem is that after unification with `Any` type variables can also become `Any`. I am aware that this may introduce (rare) cases when we don't detect never-matched overload. However, since this error does not introduce any unsafety, false negatives are clearly better than false positives.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3752,9 +3752,10 @@ def overload_can_never_match(signature: CallableType, other: CallableType) -> bo
 
     Assumes that both signatures have overlapping argument counts.
     """
-    signature = expand_type(signature, {tvar.id: tvar.erase_to_union_or_bound()
-                            for tvar in signature.variables})
-    return is_callable_compatible(signature, other,
+    exp_signature = expand_type(signature, {tvar.id: tvar.erase_to_union_or_bound()
+                                for tvar in signature.variables})
+    assert isinstance(exp_signature, CallableType)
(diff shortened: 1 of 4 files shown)


ilevkivskyi:
Just to clarify, the main complain in the original issue is a valid mypy error, but the subsequently appeared error about never-matching overload is bogus. And it is what I try to fix in this PR.

Michael0x2a:
> Could you please clarify what do you mean by "definitely confirms"? Do you want to find a call to foo that will chose the second overload?

Ah, sorry -- that was poorly worded. I meant to say that I think it's correct that mypy reports no errors with the test cases I included. However, I could see a future PR (probably one that I'd write, TBH) making mypy report errors in cases like those by accident. So, I wanted those test cases added to this PR to guard against that happening.

ilevkivskyi:
> So, I wanted those test cases added to this PR to guard against that happening.

OK, np. I added the test in the last commit above.

ilevkivskyi:
@Michael0x2a OK, so does this mean the PR can be merged now? Could you please take a look again?

Michael0x2a:
Yup, everything looks good to me -- I went ahead and merged.

(I still have vague misgivings about how we're handling Any in general, but it's not obvious to me what to do about it/agree that this PR looks like the best approach so far.)

Issue #4191, opened 2017-11-01
Running mypy with --disallow-untyped-decorators reports an error for overloaded decorators


joshstaiger:
Running mypy --disallow-untyped-decorators with a decorator that defines overloads always seems to report:

Untyped decorator makes function "xxx" untyped

Even though reveal_type(...) calls on the decorated functions seem to indicate type information.

Consider the following code.  

Running mypy  --disallow-untyped-decorators here yields the following:

```
dec_issue.py:19: error: Overloaded function signatures 1 and 2 overlap with incompatible return types
dec_issue.py:35: error: Untyped decorator makes function "default" untyped
dec_issue.py:41: error: Revealed type is 'def (*Any, **Any) -> builtins.int*'
```

The first error seems to be covered by other bug reports here (though, for all I know it may be related).

However, the second error seems to be contradicted by the type_reveal, which indicates at least the return type of the decorated "default" function is being picked up.  

I may be missing something, but at very least, it seems like a more descriptive error may be warranted?  

```
from typing import *

R = TypeVar("R")


def _make_decorator(arg: str) -> Callable[[Callable[..., R]], Callable[..., R]]:
    def dec_(fn: Callable[..., R]) -> Callable[..., R]:
        def wrapped(*args: Tuple[Any, ...],
                    **kwargs: Dict[str, Any]) -> R:
            print(f"Prefix {arg}!")
            return fn(*args, **kwargs)

        return wrapped

    return dec_


@overload
def dec(x: Callable[..., R]) -> Callable[..., R]: ...


@overload
def dec(x: str) -> Callable[[Callable[..., R]], Callable[..., R]]: ...


def dec(x: Any) -> Any:
    if isinstance(x, str):
        return _make_decorator(x)
    elif callable(x):
        return _make_decorator("default")(x)
    else:
        raise Exception("unhandled")


@dec
def default(name: str) -> int:
    print(f"Hello, {name}!")
    return 0


reveal_type(default)


@dec("arg")
def arg(name: str) -> int:
    print(f"Hello, {name}!")
    return 0


if __name__ == "__main__":
    default("Bob")
    arg("Bob")

```

ilevkivskyi:
This still fails on master (the "Untyped decorator makes function "default" untyped"). I tried a more precise signature:
```python
F = TypeVar('F', bound=Callable[..., Any])
@overload
def dec(x: F) -> F: ...
@overload
def dec(x: str) -> Callable[[F], F]: ...
```
[...]

PR #5509, opened 2018-08-20
Fix untyped decorator check for class instances

izquierdo:
Running `mypy --disallow-untyped-decorators` when the decorator is a typed callable instance produces an error even though `reveal_type` shows the correct type.

Related to #4191 but I'll leave that one for a different PR.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -4007,7 +4007,7 @@ def is_untyped_decorator(typ: Optional[Type]) -> bool:
     if not typ:
         return True
     if isinstance(typ, CallableType):
-        return typ.implicit
+        return not is_typed_callable(typ)
     elif isinstance(typ, Instance):
         method = typ.type.get_method('__call__')
         if method and method.type:
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -4014,6 +4014,8 @@ def is_untyped_decorator(typ: Optional[Type]) -> bool:
             return not is_typed_callable(method.type)
         else:
             return False
+    elif isinstance(typ, Overloaded):
+        return any(is_untyped_decorator(item) for item in typ.items())
     return True
 
 
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -4008,7 +4008,7 @@ def is_untyped_decorator(typ: Optional[Type]) -> bool:
         return True
     if isinstance(typ, CallableType):
         return typ.implicit
-    if isinstance(typ, Instance):
+    elif isinstance(typ, Instance):
         method = typ.type.get_method('__call__')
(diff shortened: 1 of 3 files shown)


mypy/checker.py
```diff
@@ -4004,9 +4004,15 @@ def is_typed_callable(c: Optional[Type]) -> bool:
...
+    if not typ:
         return True
-    return typ.implicit
+    if isinstance(typ, CallableType):
+        return typ.implicit
+    if isinstance(typ, Instance):
+        method = typ.type.get_method('__call__')
+        if method and method.type:
+            return not is_typed_callable(method.type)
+    return True
```
ilevkivskyi: If an `Instance` doesn't have a `__call__` method this will return `True`, which will generate an additional spurious error. While I believe it shouldn't.

Could you please add a test case for this?

mypy/checker.py
```diff
@@ -4004,9 +4004,15 @@ def is_typed_callable(c: Optional[Type]) -> bool:
 
 
 def is_untyped_decorator(typ: Optional[Type]) -> bool:
-    if not typ or not isinstance(typ, CallableType):
+    if not typ:
         return True
-    return typ.implicit
+    if isinstance(typ, CallableType):
+        return typ.implicit
+    if isinstance(typ, Instance):
```
ilevkivskyi: `if` -> `elif` would be better here.

mypy/checker.py
```diff
@@ -4004,9 +4004,15 @@ def is_typed_callable(c: Optional[Type]) -> bool:
...
+        method = typ.type.get_method('__call__')
+        if method and method.type:
+            return not is_typed_callable(method.type)
```
ilevkivskyi: This seems a bit inconsistent with the above. I would either use `typ.implicit` in both branches, or `is_typed_callable` in both branches.

test-data/unit/check-flags.test
```diff
@@ -132,6 +132,22 @@ def d3(p) -> Any:
...
+    def __call__(self, c: Callable) -> Callable:
+        return function
+
+class UntypedDecorator:
+    def __call__(self, c):
+        return function
+
+@UntypedDecorator()  # E: Untyped decorator makes function "f" untyped
+@TypedDecorator()
+def f() -> None: pass
```
ilevkivskyi: Could you please add four separate checks that would check only first decorator, only second decorator, and both of them in different order. Each check should also have a `reveal_type` for the function so that it is clear the behaviour is correct.

izquierdo:
Thank you for the review. Updated with requested fixes + fix for the `Overloaded` type.

PR #5511, opened 2018-08-20
Use extras_require over install_requires in mypy_extensions

anowlcalledjosh:
There are still old versions of setuptools hanging around that don't understand environment markers in install_requires (e.g. in Debian Jessie, see common-workflow-language/cwltool#872) and it would be nice if mypy_extensions would be installable on those versions of setuptools, especially since the main mypy setup.py uses extras_require for all its environment markers.

### DIFF
--- a/extensions/setup.py
+++ b/extensions/setup.py
@@ -37,7 +37,7 @@ setup(
     license='MIT License',
     py_modules=['mypy_extensions'],
     classifiers=classifiers,
-    install_requires=[
-        'typing >= 3.5.3; python_version < "3.5"',
-    ],
+    extras_require={
+        ':python_version < "3.5"': 'typing >= 3.5.3',
+    },
 )


ethanhs:
I don't think there is a downside, but I also don't think it is needed if we release `mypy_extensions` as a wheel (which is a good thing to do in general). According to Hynek ([here](https://hynek.me/articles/conditional-python-dependencies/)), only pip 1.4 is needed to install a wheel which uses `install_requires`. However, with a source distribution, a rather recent version of setuptools is needed to build the package. I'd prefer to release a wheel, as using `extras_require` in this way is not really documented.

ilevkivskyi:
> I'd prefer to release a wheel, as using `extras_require` in this way is not really documented.

@anowlcalledjosh Does this solution work for for you? It would be easy to just upload wheels to PyPI.

anowlcalledjosh:
Assuming that:

> only pip 1.4 is needed to install a wheel which uses `install_requires`

then that sounds fine to me - I have no desire to perpetuate the `extras_require` hack if it's not necessary.

ilevkivskyi:
OK, I will make micro-release `mypy_extensions 0.4.1` now.

ilevkivskyi:
OK, I just uploaded the source and wheels distributions for `mypy_extensions 0.4.1`.

Closing this for now. Please re-open or create a new issue if you still have problems.

ilevkivskyi:
FYI https://pypi.org/project/mypy_extensions/0.4.1/#files

PR #5513, opened 2018-08-20
More tweaks to support mypyc

msullivan:


mypy/checkexpr.py
```diff
@@ -720,8 +720,8 @@ def analyze_type_type_callee(self, item: Type, context: Context) -> Type:
                 res = res.copy_modified(from_type_type=True)
             return expand_type_by_instance(res, item)
         if isinstance(item, UnionType):
-            return UnionType([self.analyze_type_type_callee(item, context)
-                              for item in item.relevant_items()], item.line)
+            return UnionType([self.analyze_type_type_callee(x, context)
+                              for x in item.relevant_items()], item.line)
```
ilevkivskyi: I would use a bit more descriptive name like `tp` instead of just `x`.

mypy/options.py
```diff
@@ -191,7 +191,7 @@ def __init__(self) -> None:
         self.use_builtins_fixtures = False
 
         # -- experimental options --
-        self.shadow_file = None  # type: Optional[List[Tuple[str, str]]]
+        self.shadow_file = None  # type: Optional[List[List[str]]]
```
ilevkivskyi: Hm, I am curious why mypy didn't catch any error about this?

msullivan: The field gets populated by argparse, I think?

PR #5518, opened 2018-08-21
Make Options work under mypyc

msullivan:
This requires reworking a bunch of operations to not directly depend
on `__dict__`

### DIFF
--- a/mypy/options.py
+++ b/mypy/options.py
@@ -17,43 +17,43 @@ class BuildType:
     PROGRAM_TEXT = 2  # type: ClassVar[int]
 
 
+PER_MODULE_OPTIONS = {
+    # Please keep this list sorted
+    "always_false",
+    "always_true",
+    "check_untyped_defs",
+    "debug_cache",
+    "disallow_any_decorated",
+    "disallow_any_explicit",
+    "disallow_any_expr",
+    "disallow_any_generics",
+    "disallow_any_unimported",
+    "disallow_incomplete_defs",
+    "disallow_subclassing_any",
+    "disallow_untyped_calls",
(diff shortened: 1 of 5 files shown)


mypy/options.py
```diff
@@ -209,12 +214,16 @@ def __init__(self) -> None:
...
+        d = dict(getattr(self, '__dict__', ()))
+        for k in get_class_descriptors(Options):
+            if hasattr(self, k):
+                d[k] = getattr(self, k)
         del d['per_module_cache']
         return d
 
     def __eq__(self, other: object) -> bool:
-        return self.__class__ == other.__class__ and self.__dict__ == other.__dict__
+        return isinstance(other, Options) and self.snapshot() == other.snapshot()
```
gvanrossum: This looks like this `__eq__` will become much more expensive. Have you tracked where/how it is used?

mypy/options.py
```diff
@@ -209,12 +214,16 @@ def __init__(self) -> None:
 
     def snapshot(self) -> object:
         """Produce a comparable snapshot of this Option"""
-        d = dict(self.__dict__)
+        # Under mypyc, we don't have a __dict__, so we need to do worse things.
+        d = dict(getattr(self, '__dict__', ()))
```
gvanrossum: So IIUC under mypyc d will start out empty?

mypy/options.py
msullivan: I remembered from when I was looking into this while doing the original snapshot stuff that the answer was "not often", but I looked more closely and the answer was: in exactly one test. I just removed `__eq__` and made that test use snapshot directly

Issue #1214, opened 2016-02-11
Implement Const type to indicate non-rebindable variables/attributes


hashstat:
As a Python developer, I would like a _Const_ type so that I can indicate that a variable or attribute may not be rebound to another value. This would be especially useful for module and class constants, but could also be helpful for instance attributes that really shouldn't be modified. Not only does it help prevent accidentally overwriting constants, it would hopefully reduce the need to use `@property` on non-critical attributes. Here is a contrived example:

``` python
GRUNT = 0   # type: Const
MANAGER = 1   # type: Const


class Employee:
    kind = GRUNT   # type: Const

    def __init__(self, idnum: int, name: str):
        self.id = idnum   # type: Const
        self.name = name


class Manager(Employee):
    kind = MANAGER   # type: Const
```

`kind = MANAGER   # type: Const` reqiures the _Const_ type to rebind _kind_ without an error just as a const variable in C requires a cast during reassignment to quiet compiler warnings/errors.

I'm not sure if it makes sense to allow _Const_ to take parameters as the type can be inferred in most cases. One might want other variables, such as those passed into functions, to be const, in which case it makes sense to allow _Const_ to take parameters. Here is an example:

``` python
def frobnicate(kind: Const[int]) -> None:
    kind = 5   # Error
    kind = 6   # type: Const  # Okay
```

JukkaL:
Early in the history of mypy I was actually planning to have something like this, but I guess I later forgot about it. This would likely require a PEP 484 (and potentially `typing`) change.

davidroeca:
Const in Python could have similar guarantees to something like JavaScript ES6's const, to prevent reassignment https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const

A clever programmer would be able to get around this through methods that mutate, or by a shared reference, but the guard rails are still nice. 

Alternatively, it could only be allowed for immutable non-container primitives such as int or str, or maybe even extended to tuples/namedtuples of immutable primitives as well

JukkaL:
We are now considering something like this, perhaps spelled as `Final`. An interesting use case would be with literal types (another potential new feature) -- it would allow mypy to infer things like `if X:` never being taken if there is `X: Final = False`.

PR #5522, opened 2018-08-22
Final names and attributes

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/1214
Fixes https://github.com/python/typing/issues/286
Fixes https://github.com/python/typing/issues/242 (partially, other part is out of scope)

This is a working implementation of final access qualifier briefly discussed at PyCon typing meeting. Final names/attributes can be used to have more static guarantees about semantics of some code and can be used by other tools like mypyc for optimizations.

We can play with this implementation before starting to write an actual PEP.

The basic idea is simple: once declared as final, a name/attribute can't be re-assigned, overridden, or redefined in any other way. For example:
```python
from typing import Final

NO: Final = 0
YES: Final = 255

class BaseEngine:
    RATE: Final[float] = 3000

YES = 1  # Error!

class Engine(BaseEngine):
    RATE = 9000  # Also an error!
```
For more use cases, examples, and specification, see the docs patch.

Here are some comments on decisions made:
* __What can be final?__ It is hard to say what semantic nodes are important, I started from just module and class constants, but quickly realized it is hard to draw the line without missing some use cases (in particular for mypyc). So I went ahead and implemented all of them, everything can be final: module constants, class-level and instance-level attributes, method, and also classes.
* __Two names or one name?__ I currently use two names `Final` for assignments and `@final` for decorators. My PEP8-formatted mind just can't accept `@Final` :-)
* __Should re-exported names keep they const-ness?__ I think yes, this is a very common pattern, so it looks like this is a sane default.
* __What to do with instance-level vs class-level attributes?__ The point here is that mypy has a common namespace for class attributes. I didn't want to complicate things (including the mental model), so I just decided that one can't have, e.g., a name that is constant on class but assignable on instances, etc. Such use cases are relatively rare, and we can implement this later if there will be high demand for this.

...deferred features:
* I didn't implement any constant propagation in mypy _yet_. This can be done later on per use-case 
  basis. For example:
  ```python
  fields: Final = [('x', int), ('y', int)]
  NT = NamedTuple('NT', fields)
  ```
* __Should final classes be like sealed in Scala?__ I think probably no. On one hand it could be be a nice feature, on other hand it complicates the mental model and is less useful for things like mypyc.
* I don't allow `Final` in function argument types. One argument is simplicity, another is I didn't see many bugs related to shadowing an argument in function bodies, finally people might have quite different expectations for this. If people will ask, this would be easy to implement.

...and implementation internals:
* There are two additional safety nets that I don't mention in the docs: (a) there can be no `TypeVar`s in the type of class-level constant, (b) instance-level constant can't be accessed on the class object.
* I generate errors for re-definitions in all subclasses, not only in immediate children. I think this is what most people would want: turning something into a constant will flag most re-assignment points.
* It is hard to always get a single error sometimes (I have two error messages, one for re-assignment, another for overriding) because of how mypy creates new `Var`s. But I think having two errors occasionally is fine.
* We store the `final_value` for constants initialized with a simple literal, but we never use it. This exists only for tools like mypyc that may use it for optimizations.

The PR is open for questions, suggestions, and comments.

cc @ambv @rchen152 @vlasovskikh


mvaled:
@med-merchise you were asking for this, I think.

ilevkivskyi:
@Michael0x2a 
> first, it might be nice if there were some tests + docs about Final and control flow

I would actually prohibit `Final` in loops. Moreover, I would propose to also prohibit `Final` in `if` statements, unless the truthiness of condition is known statically (like `sys.version > 3.6`).

In addition, I just noticed that I forgot about in-place assignments, will fix this soon.

> Second, I expect that this proposal will be rejected immediately for being too unpythonic

I am not sure I understand this comment. Rejected immediately by whom? There were no objections during PyCon typing meeting, moreover many people liked the idea.

> too messy to implemented

Same story here, sorry :-) Do you want to say that the implementation in this PR is messy?

> How do we feel about adding a flag to mypy that makes all variables and attributes Final by default?

I am not sure there is any existing Python program larger that 1k lines that will type-check with such flag. Plus we already have dozens of flags. So I am against this unless people will ask often for this.

Michael0x2a:
> I am not sure I understand this comment. Rejected immediately by whom? There were no objections during PyCon typing meeting, moreover many people liked the idea.

Ah, I meant to say that I expected my proposal to add a flag to make final always-on by default would be rejected immediately by you.

(Which is what just happened -- but it was worth a shot 😄)

msullivan:
> How do we feel about adding a flag to mypy that makes all variables and attributes Final by default?

My response to this is a wistful sigh

msullivan:
Regarding closed classes in mypyc: currently mypyc doesn't allow a mypyc-compiled class to be subclassed except by other classes compiled at the same time, and I expect that to remain the default behavior with a mechanism to allow interpreted classes to subclass. (Doing that requires either deoptimizing or generating substantial glue code, though, so it probably will be opt-in.) So we sort of already assume classes are closed and probably don't need a more official mechanism.

ilevkivskyi:
Thanks for review! I will respond to other comments tomorrow, here is just one clarification.
> Regarding closed classes in mypyc...

My idea was to super-micro-optimise `isinstance()` for final classes. We can just emit `Py_TYPE(obj) == &TargetType`, i.e. a single pointer comparison. Current implementation already does the same when `isinstance()` returns `True`, but the point is that most `isinstance()` checks return `False`. In my experience, `isinstance()` is typically the most called function in a Python program, so optimising even 10 cycles per call, can give a visible performance boost.

We probably can just find which classes are never subclassed in compiled code now, but explicitly marking some classes as final we can not worry about what we will decide in future.

test-data/unit/lib-stub/typing.pyi
```diff
@@ -16,6 +16,7 @@ NamedTuple = 0
 Type = 0
 no_type_check = 0
 ClassVar = 0
+Final = 0
```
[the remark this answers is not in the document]
ilevkivskyi: Yes, we will be able to move to `typing` only with a PEP (can be a short one and/or combined with literal types). I think we can add it to `typing_extensions` even before PEP acceptance. I would propose to follow a strategy similar to what was done for protocols:
* Allow using `Final` in stubs immediately
* After one mypy release cycle fix bugs and design errors and add `Final` to `typing_extensions`. It would be good to have a draft PEP ready at this point.
* After PEP is accepted, move `Final` from `typing_extensions` to `typing`.

If everyone agrees with such schedule, then I can actually also allow using `typing_extensions.Final` (in stubs, because there is no runtime counterpart yet) right in this PR, please let me know. (Otherwise I can do this in a follow-up PR.)

mypy/checker.py
```diff
@@ -217,6 +217,10 @@ def __init__(self, errors: Errors, modules: Dict[str, MypyFile], options: Option
         # If True, process function definitions. If False, don't. This is used
         # for processing module top levels in fine-grained incremental mode.
         self.recurse_into_functions = True
+        # This internal flag is used to indicate whether we a currently type-checking
+        # a final declaration (assignment). Should not be set manually, use
+        # get_final_context/set_final_context instead.
+        self._is_final_def = False
```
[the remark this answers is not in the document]
ilevkivskyi: Yes, will add now.

mypy/checker.py
```diff
@@ -1614,6 +1631,11 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
             for lv in s.lvalues[:-1]:
                 self.check_assignment(lv, rvalue, s.type is None)
 
+        self.check_final(s)
+        if (s.is_final_def and s.type and not has_no_typevars(s.type)
+                and self.scope.active_class() is not None):
+            self.fail("Constant declared in class body can't depend on type variables", s)
```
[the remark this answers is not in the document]
ilevkivskyi: Do you mean class-level constants or instance-level constants? The latter are allowed to depend on type variables, and there is a test case for this, the pattern I could imagine is like:
```python
class C(Generic[T]):
        self.x: Final = x
```
I however can't imagine when the same would be needed for class level constants:
```python
class C(Generic[T]):
    x: Final[T] = ...  # what would I write here? There must be a r.h.s.
```

mypy/checkmember.py
```diff
@@ -262,6 +262,15 @@ def analyze_member_var_access(name: str, itype: Instance, info: TypeInfo,
...
     if isinstance(v, Var):
         implicit = info[name].implicit
+
+        # An assignment to final attribute is always an error,
+        # independently of types.
+        if is_lvalue and not chk.get_final_context():
+            for base in info.mro:
+                sym = base.names.get(name)
+                if sym and isinstance(sym.node, (Var, FuncBase, Decorator)) and sym.node.is_final:
+                    msg.cant_assign_to_final(name, module_level=False, ctx=node)
```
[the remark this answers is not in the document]
ilevkivskyi: Yes, makes sense. I will factor this out into a helper.

mypy/checkmember.py
```diff
@@ -535,6 +544,21 @@ def analyze_class_attribute_access(itype: Instance,
...
+    # can't be accessed on the class object.
+    if node.implicit and isinstance(node.node, Var) and node.node.is_final:
+        msg.fail("Can't access instance constant on class object", context)
+
+    # An assignment to final attribute on class object is also always an error,
+    # independently of types.
+    for base in itype.type.mro:
+        b_node = base.names.get(name)
+        if (b_node and isinstance(b_node.node, (Var, FuncBase, Decorator)) and
+                is_lvalue and b_node.node.is_final
```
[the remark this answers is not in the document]
ilevkivskyi: :+1:

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+
+     class DerivedCls(SomeCls):
+         ...
+
+     DerivedCls.ID = 2  # Error!
+     obj: DerivedCls
+     obj.ID = 2  # Error!
+
+* A final attribute can't be overridden by a subclass (even with another
+  explicit final declaration). Note however, that final attributes can
```
JukkaL: As discussed above, I'm not sure about this restriction. Overriding a class-level final attribute with another class-level final attribute seems like a reasonable thing to do for a per-class constant. This would also allow safe covariant overriding of class-level attributes. Maybe we could even somehow allow covariantly overriding a final attribute in an ABC as a final instance attribute.

Hypothetical example:

```
class Model:
    TABLE_NAME: Final[str]  # It would be better if this was explicitly abstract

class Person(Model):
    TABLE_NAME: Final = 'Persons'  # Inferred type should be 'str'?

class Address(Model):
    TABLE_NAME: Final = 'Address'
```

Finally, should it be possible to override a (possibly abstract) property with a final attribute?

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+----------------
+
+The are two rules that should be always followed when defining a final name:
+
+* There can be *at most one* final declaration per module or class for
+  a given attribute:
+
+  .. code-block:: python
+
+     ID: Final = 1
```
JukkaL: It would be better to explicitly import `Final` at least in the first example that uses it.

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+
+   # 1000 lines later
+
+   class User(Snowflake):
+       id = uuid.uuid4()  # This has valid type, but the meaning
+                          # may be wrong
+
+Some other use cases might be solved by using ``@property``, but note that both
+above use cases can't be solved this way. For such situations, one might want
+to use ``typing.Final``.
```
JukkaL: This won't be in `typing` at least until a new Python version has been released, and even then older Python versions can't use this. Use a name that can be imported now (either `typing_extensions.Final` or `mypy_extensions.Final`). Once some version of Python ships `Final`, we can change this back and add a note about `typing_extensions` for users of earlier Python versions.

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+         ID: Final = 2  # OK
+
+     class Combo(One, Other):  # Error, cannot override final attribute.
+         pass
+
+Final methods
+-------------
+
+Like with attributes, sometimes it is useful to protect a method from
+overriding. In such situations one can use a ``typing.final`` decorator:
```
JukkaL: Like above, this needs to live outside `typing` for now.

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+
+     class One(Base):
+         ID: Final = 1  # OK
+
+     class Other(Base):
+         ID: Final = 2  # OK
+
+     class Combo(One, Other):  # Error, cannot override final attribute.
+         pass
+
```
JukkaL: Maybe also mention that this doesn't turn the object assigned to a final attribute immutable. For example, this would be fine:

```
x: Final = ['a', 'b']
x.append('x')  # Ok
```

Using a read-only ABC or protocol can be used to work around this sometimes (or an immutable concrete class, of course):

```
x: Final[Sequence[str]] = ['a', 'b']
x.append('x')  # Error
y: Final = ('a', 'b')  # Also an option
```

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+     ID = 1
+     ID: Final = 2  # Error!
+
+     class SomeCls:
+         ID = 1
+         ID: Final = 2  # Error!
+
+.. note::
+   Conditional final declarations and final declarations within loops are
+   not supported.
```
JukkaL: Other things that may be worth mentioning:

* `Final` can't be used nested within types. For example, `List[Final[int]]` would be rejected.
* Function arguments can't be declared as `Final` (discuss also why).
* What about final class attributes? Would you use `ClassVar[Final]`, `Final[ClassVar]` or something else? Or are these unnecessary? This may be worth discussing anyway.

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+* There can be *at most one* final declaration per module or class for
+  a given attribute:
+
+  .. code-block:: python
+
+     ID: Final = 1
+     ID: Final = 2  # Error!
+
+     class SomeCls:
+         id: Final = 1
```
JukkaL: What if there is no initializer here. Would `self.id = x` then be valid in `__init__`? I'd argue that it should be valid, to allow all attributes to be declared in the class body.

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+Like with attributes, sometimes it is useful to protect a method from
+overriding. In such situations one can use a ``typing.final`` decorator:
+
+.. code-block:: python
+
+   from typing import final
+
+   class Base:
+       @final
+       def common_name(self) -> None:  # common signature
```
JukkaL: It's unclear what "common signature" means here.

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+           ...
+
+   # 1000 lines later
+
+   class Derived(Base):
+       def common_name(self) -> None:  # Error, this overriding might break
```
JukkaL: Style nit: Using colon after "Error" would be slightly better ("Error: this ...")

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+                                       # invariants in the base class.
+           ...
+
+This ``@final`` decorator can be used with instance methods, class methods,
+static methods, and properties (this includes overloaded methods).
```
JukkaL: An example of using this with an overloaded method would be useful -- it's not entirely clear how to use it. If there are multiple decorators, can `@final` be placed arbitrarily, or does it need to be the outermost decorator, for example?

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+
+Final classes
+-------------
+
+As a bonus, applying a ``typing.final`` decorator to a class indicates to mypy
+that it can't be subclassed. Mypy doesn't provide any additional features for
```
JukkaL: Discuss why users might want to declare a class as final. Here are some ideas:

* A class wasn't designed to be subclassed. Perhaps subclassing does not work as expected, or it's error-prone.
* You want to retain the freedom to arbitrarily change the class implementation in the future, and these changes might break subclasses.
* You believe that subclassing would make code harder to understand or maintain. For example, you may want to prevent unnecessarily tight coupling between base classes and subclasses.

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+
+   # file main.py
+   from lib import Leaf
+
+   class MyLeaf(Leaf):  # Error, library author believes this is unsafe
+       ...
+
+.. note::
+
+   Mypy treats re-exported final names as final. In other words, once declared,
```
JukkaL: This note seems out of place here. This doesn't seem specific to final classes.

JukkaL:
Note that there is a merge conflict.

ilevkivskyi:
@JukkaL Thanks for review! It looks like there are mostly minor comments, but there are two larger questions we need to discuss:
1. Whether we should allow overriding a class level constant in a subclass
2. Whether to allow "delayed initialization"

I don't like the first idea for several reasons:
* This doesn't look right for something called `Final`. It would be OK for something called `Const`, but I don't want to introduce two different features with subtly different semantics.
* This looks like an exception. Why only class-level attributes and not instance-level? It will be both hard to understand and hard to remember for a user.
* This requires further complicating the rules: can we override it with an instance attribute? Or vice versa? What to do with multiple inheritance when one base overrides another? What about multiple inheritance combined with class vs instance?
* This is unsafe (from both mypyc and literal types view) if we allow accessing class level attributes on instances. We can of course prohibit this but then it looks like we need another exception to fix an exception.
```python
class Base:
    MODE: Final = 'r'
class Child(Base):
    MODE: Final = 'w'

def read(file: str, mode: Literal['r', 'rb']) -> None:
    ...
x: Base = Child()
read('somefile', x.MODE)  # I think this is not OK
```

On the other hand I am OK with the second idea. An attribute can be declared as `Final` in the class body without an initial value, but then can be initialized later (exactly once per class). I think we can allow delayed initialization both in `__init__` and in a subclass body. For example:
```python
class Cls:
    x: Final[int]
    def __init__(self, x) -> None:
        self.x = x  # OK, but all subsequent assignments will be errors
                    # in this class, externally, and in subclass methods.

class Base:
    x: Final[int]
class OneChild(Base):
    x = 1
class OtherChild(Base):
    x = 2
```
The only questionable part here is that stub authors should be careful now, because:
```
x: Final[int]
```
and
```
x: Final[int] = ...
```
will mean different things. This difference already appears as significant for protocols, so it is not totally new but still may cause some confusions.

docs/source/more_types.rst
ilevkivskyi: Currently a method is final if at least one overload (or implementation) has `@final`. I will add an example.

docs/source/more_types.rst
ilevkivskyi: Currently initial value is _required_. See my large comment about other option.

docs/source/more_types.rst
ilevkivskyi: `Final[ClassVar[int]]` is redundant, since we anyway know where the constant was initialized, in class body or in `__init__`.

(Moreover `ClassVar[Final[int]]` is an error because `Final` is prohibited in any nested position.)

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+     # file mod.py
+     from typing import Final
+
+     ID: Final = 1
+
+     # file main.py
+     from typing import Final
+
+     import mod
+     mod.ID = 2  # Error, can't assign to constant.
```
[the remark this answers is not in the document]
ilevkivskyi: This is prohibited, I think it is safer to prohibit this than to allow.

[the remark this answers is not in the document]
ilevkivskyi: The note at the very bottom is about this. I will move that note here.

gvanrossum:
@ilevkivskyi 
> 1. Whether we should allow overriding a class level constant in a subclass
> [...]
> I don't like the first idea for several reasons:

I am with Ivan here.

JukkaL:
> Whether we should allow overriding a class level constant in a subclass

I figured this might be useful is since I remember having seen production code that could benefit from it. But I agree that the feature would have to be called `Const` or something similar for this to make sense. So let's not do this.

> Whether to allow "delayed initialization"

I think that, for now, we should only support delayed initialization in the same class (`__init__`). The subclass idea may be useful as well, but it now feels a bit ad hoc to me. We can always add it later if we still like it after further deliberation. Not supporting this in subclasses would enable us to give an error if there is `x: Final` in a class body but the user forgot to initialize `x`.

docs/source/more_types.rst
```diff
@@ -933,3 +933,210 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
 another TypedDict if all required keys in the other TypedDict are required keys in the
 first TypedDict, and all non-required keys of the other TypedDict are also non-required keys
 in the first TypedDict.
+
+Final attributes of classes and modules
```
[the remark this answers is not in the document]
ilevkivskyi: Hm, I just wanted to move this to the section where `ClassVar` is, but it turns out we don't have such section. I will create a new section "Type qualifiers" and add this plus a small stub about `ClassVar`. I will also open an issue to add more docs about `ClassVar` (I have seen several cases where people were confused by `ClassVar` semantics).

mypy/semanal.py
```diff
@@ -610,6 +610,13 @@ def _visit_overloaded_func_def(self, defn: OverloadedFuncDef) -> None:
             # redefinitions already.
             return
 
+        # Check final status, if at least one item or the implementation is marked
+        # as @final, then the whole overloaded definition if @final.
```
JukkaL: Typo: if -> is?

mypy/semanal.py
```diff
@@ -610,6 +610,13 @@ def _visit_overloaded_func_def(self, defn: OverloadedFuncDef) -> None:
             # redefinitions already.
             return
 
+        # Check final status, if at least one item or the implementation is marked
```
JukkaL: I don't like this rule, as it allows a lot of flexibility with no clear benefit. I'd prefer to only allow `@final` in the implementation, and report an error if it's used elsewhere.

mypy/semanal.py
```diff
@@ -1680,6 +1701,95 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
...
+            return
+        assert isinstance(s.type, UnboundType)
+        if len(s.type.args) > 1:
+            self.fail("Final[...] takes at most one type argument", s.type)
+        if not s.type.args:
+            s.type = None
+        else:
+            s.type = s.type.args[0]
+        if len(s.lvalues) != 1 or not isinstance(s.lvalues[0], RefExpr):
+            self.fail("Invalid final declaration, ignoring", s)
```
JukkaL: The ", ignoring" suffix seems isconsistent. Other messages don't use it, so I'd drop it.

mypy/semanal.py
```diff
@@ -1680,6 +1701,95 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
...
+            return
+        lval = s.lvalues[0]
+        assert isinstance(lval, RefExpr)
+        s.is_final_def = True
+        if self.loop_depth > 0:
+            self.fail("Final declarations are prohibited within loops", s)
```
JukkaL: Style nit: The error message feels inconsistent. An arguably better alternative would be "Cannot use Final inside a loop".

mypy/semanal.py
```diff
@@ -1680,6 +1701,95 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
...
+        if self.type and self.type.is_protocol:
+            self.msg.protocol_members_cant_be_final(s)
+        if (isinstance(s.rvalue, TempNode) and s.rvalue.no_rhs and
+                not self.is_stub_file and not self.is_class_scope()):
+            self.fail("Final names outside stubs must have a value", s)
```
JukkaL: Style nits: Use singular for consistency. Maybe also leave out the stub exception since it's kind of specialized. Most users probably never edit stubs, so this could be confusing (for example, a naive user might think that the correct fix is to turn the file into a stub file).  An idea for an altenative: "Final name must be initialized with a value".

mypy/semanal.py
```diff
@@ -1680,6 +1701,95 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
...
+        if isinstance(lval, MemberExpr):
+            if not self.is_self_member_ref(lval):
+                self.fail("Final can be only applied to a name or an attribute on self", s)
+                s.is_final_def = False
+                return
+            else:
+                assert self.function_stack
+                if self.function_stack[-1].name() != '__init__':
+                    self.fail("Can only declare final attributes in class body"
+                              " or __init__ method", s)
```
JukkaL: Really tiny style nits: Leave out "method" for consistency. Write as "... or __init" or "... or '... or "__init__"'.

mypy/semanal.py
```diff
@@ -1806,7 +1916,8 @@ def check_and_set_up_type_alias(self, s: AssignmentStmt) -> None:
 
     def analyze_lvalue(self, lval: Lvalue, nested: bool = False,
                        add_global: bool = False,
-                       explicit_type: bool = False) -> None:
+                       explicit_type: bool = False,
+                       final_cb: Optional[Callable[[], None]] = None) -> None:
```
JukkaL: Document `final_cb` in docstring.

mypy/semanal.py
```diff
@@ -1927,18 +2042,24 @@ def analyze_tuple_or_list_lvalue(self, lval: TupleExpr,
                 star_exprs[0].valid = True
             for i in items:
                 self.analyze_lvalue(i, nested=True, add_global=add_global,
-                                    explicit_type = explicit_type)
+                                    explicit_type=explicit_type)
 
-    def analyze_member_lvalue(self, lval: MemberExpr, explicit_type: bool = False) -> None:
+    def analyze_member_lvalue(self, lval: MemberExpr, explicit_type: bool = False,
+                              final_cb: Optional[Callable[[], None]] = None) -> None:
```
JukkaL: Document `final_cb`.

mypy/server/aststrip.py
```diff
@@ -266,8 +274,15 @@ def visit_member_expr(self, node: MemberExpr) -> None:
...
             self.strip_class_attr(node.name)
             node.def_var = None
+        if isinstance(node.node, Var):
+            self._reset_var_final_flags(node.node)
         super().visit_member_expr(node)
 
+    def _reset_var_final_flags(self, v: Var) -> None:
+        v.is_final = False
+        v.final_unset_in_class = False
+        v.final_set_in_init = False
```
JukkaL: What about the `final_value` attribute?

mypy/test/data.py
```diff
@@ -187,6 +188,14 @@ def parse_test_cases(parent: 'DataSuiteCollector', suite: 'DataSuite',
                     path, p[i0].line))
 
 
+MYPY = False
+if MYPY:
+    if sys.version_info >= (3, 5):
```
JukkaL: This change seems unrelated?

mypy/nodes.py
```diff
@@ -748,6 +760,15 @@ def __init__(self, name: str, type: 'Optional[mypy.types.Type]' = None) -> None:
         # Set to true when this variable refers to a module we were unable to
         # parse for some reason (eg a silenced module)
         self.is_suppressed_import = False
+        # Was this "variable" (rather a constant) defined as Final[...]?
+        self.is_final = False
+        # If constant value is a simple literal,
+        # store the literal value (unboxed) for the benefit of
+        # tools like mypyc.
+        self.final_value = None  # type: Optional[Union[int, float, bool, str]]
```
JukkaL: This doesn't seem to get serialized?

test-data/unit/lib-stub/typing.pyi
```diff
@@ -55,4 +56,6 @@ class Mapping(Generic[T_contra, T_co]):
 
 def runtime(cls: type) -> type: pass
 
+def final(meth: T) -> T: pass
```
JukkaL: Add note mentioning that this is an unofficial extension.

test-data/unit/lib-stub/typing.pyi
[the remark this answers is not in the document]
JukkaL: > After one mypy release cycle fix bugs and design errors and add Final to typing_extensions. It would be good to have a draft PEP ready at this point.

We can't really advertise a feature in mypy documentation if it doesn't work yet (say, if `Final` and `final` are missing from `typing_extensions`). So looks like we need to do one of these:

1. Update `typing_extensions` before the first mypy release that includes this PR. We should probably mention which version of `typing_extensions` is required to use `Final`/`final` in mypy docs.
2. Add the new definitions to `mypy_extensions` and release the changes before the first mypy release that includes this PR. Update documentation to use `mypy_extensions` in all relevant examples, instead of `typing_extensions`.
3. Postpone merging this PR until we've done one of the above two things.

Also, please add a note here mentioning that this is still an unofficial extension.

I'm not sure if it's important to start using `Final` in stubs immediately. The benefit is pretty minor compared to protocols, as this is primarily about fixing false negatives, and the concept of 'final' doesn't really exist in Python as such, so deciding which things should be final is less obvious. Duck typing, on the other hand, has a long history in Python and is pretty essential for certain APIs.

docs/source/type_quals.rst
```diff
@@ -0,0 +1,340 @@
...
+===============
+
+This section describes constructs that do not affect types of variables
+and methods, but affect how they can be accessed, assigned, and overridden.
+
+Class and instance variables
+****************************
+
+By default mypy assumes that a variable declared in the class body is
+an instance variable. One can mark names intended to be used as class variables
```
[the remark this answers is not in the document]
ilevkivskyi: > This is a bit confusing, and arguably not quite true. Mypy assumes that that it can used like either a class or an instance variable.

FWIW, this is the language used in PEP 526. When someone writes:
```python
class C:
    x: int

C.x = 0
```
variable `x` is an instance variable, and its _default_ can be set on class.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,285 @@
...
+     class One(Base):
+         ID: Final = 1  # OK
+
+     class Other(Base):
+         ID: Final = 2  # OK
+
+     class Combo(One, Other):  # Error: cannot override final attribute.
+         pass
+
+* Declaring a name as final only guarantees that the name wll not be re-bound
```
JukkaL: I think that this is good thing to leave, as this could be surprising.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,285 @@
...
+     ID = 2  # Also an error, see note below.
+
+     class DerivedCls(mod.SomeCls):
+         ...
+
+     DerivedCls.ID = 2  # Error!
+     obj: DerivedCls
+     obj.ID = 2  # Error!
+
+* A final attribute can't be overridden by a subclass (even with another
```
JukkaL: This was already mentioned above. I think that the example can be left out and this item can be shortened into "You can override a read-only property with a final attribute".

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,285 @@
...
+   Mypy treats re-exported final names as final. In other words, once declared,
+   the final status can't be "stripped". Such behaviour is typically desired
+   for larger libraries where constants are defined in a separate module and
+   then re-exported.
+
+Final methods
+*************
+
+Like with attributes, sometimes it is useful to protect a method from
+overriding. In such situations one can use a ``typing_extensions.final``
```
JukkaL: Grammar nit: "... use **the** `typing_extensions.final` ..."

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,285 @@
...
+   class Base:
+       @final
+       def common_name(self) -> None:
+           ...
+
+   # 1000 lines later
+
+   class Derived(Base):
+       def common_name(self) -> None:  # Error: this overriding might break
+                                       # invariants in the base class.
```
JukkaL: I think it's best to leave the explanation out, since it doesn't really add much. If the reader understands what "final" is useful for, the explanation adds nothing. If the reader doesn't understand why somebody would use `@final`, the explanation probably isn't sufficient, and may be confusing. I'd write this as `# Error: can't override a final method`.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,285 @@
...
+   class Base:
+       @overload
+       def meth(self) -> None: ...
+       @overload
+       def meth(self, arg: int) -> int: ...
+       @final
+       def meth(self, x=None):
+           ...
+
+    class Derived(Base):
```
JukkaL: I think that the `Derived` class can be left out. It should be obvious that overriding `meth` will be rejected.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,285 @@
...
+        def meth(self, x: Any = None) -> Any:  # Error: can't override final method
+            ...
+
+Final classes
+*************
+
+You can apply a ``typing_extensions.final`` decorator to a class indicates
```
JukkaL: The sentence is not grammatically correct ("indicates").

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,285 @@
...
+
+.. code-block:: python
+
+   from typing_extensions import final
+
+   @final
+   class Leaf:
+       ...
+
+   from lib import Leaf
```
JukkaL: Extra import that can be left out.

test-data/unit/check-final.test
```diff
@@ -0,0 +1,901 @@
...
+[case testFinalReassignModuleVarExternal]
+import mod
+mod.x = 2  # E: Can't assign to constant "x"
+[file mod.pyi]
+from typing import Final
+x: Final[int]
+[out]
+
+[case testFinalReassignInstanceVarClassBody]
+from typing import Final
```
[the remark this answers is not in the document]
ilevkivskyi: I am using both for definition test (cases that test `Final` is correctly identified). I will check and add more is some are missing.

test-data/unit/check-final.test
```diff
@@ -0,0 +1,901 @@
...
+[case testFinalReassignModuleReexport]
+from typing import Final
+
+from lib import X
+from lib.mod import ID
+
+X = 1  # E: Can't assign to constant "X"
+ID: Final = 1  # E: Name 'ID' already defined (possibly by an import) \
+               # E: Final can't redefine an existing name \
+               # E: Can't assign to constant "ID"
```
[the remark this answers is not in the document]
ilevkivskyi: OK, I will try.

test-data/unit/check-final.test
```diff
@@ -0,0 +1,901 @@
...
+
+x: Final = int()
+y: Final[float] = int()
+z: Final[int] = int()
+bad: Final[str] = str()
+
+reveal_type(x)  # E: Revealed type is 'builtins.int'
+reveal_type(y)  # E: Revealed type is 'builtins.float'
+reveal_type(z)  # E: Revealed type is 'builtins.int'
+[out]
```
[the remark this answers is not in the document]
ilevkivskyi: I will keep it because it makes easier to see where the test case ends (there are many other places where it appears).

test-data/unit/check-final.test
```diff
@@ -0,0 +1,901 @@
...
+
+[case testFinalDefiningInstanceVar]
```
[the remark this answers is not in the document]
ilevkivskyi: I like it the current way. Also it is consistent with PEP 526 terminology. Anyway this is a topic of a separate discussion.

test-data/unit/check-final.test
```diff
@@ -0,0 +1,901 @@
...
+
+-- Definitions
+
+[case testFinalDefiningModuleVar]
+from typing import Final
+
+x: Final = int()
+y: Final[float] = int()
+z: Final[int] = int()
+bad: Final[str] = str()
```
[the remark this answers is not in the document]
ilevkivskyi: Oh, I wanted a type mismatch here, will fix.

docs/source/final_attrs.rst
ilevkivskyi: So `can't` or `cannot`? I assume the latter.

docs/source/final_attrs.rst
ilevkivskyi: It is important to mention that final can't redefine final. This question actually appeared in one of feature request issues about this.

ilevkivskyi:
@JukkaL I implemented only some comments in the docs. I will kill it in the next commit, then you or any one else can finish it whenever.

mypy/checker.py
```diff
@@ -1873,6 +1918,106 @@ def check_compatibility_classvar_super(self, node: Var,
             return False
         return True
 
+    def check_compatibility_final_super(self, node: Var,
+                                        base: TypeInfo, base_node: Optional[Node]) -> bool:
+        """Check if an assignment overrides a final attribute in a base class.
+
+        This only check situations where either a node in base class is not a variable
```
JukkaL: Grammar: check -> checks

ilevkivskyi:
OK, all tests passed, we are good to go. Thanks for review!

yxliang01:
From python official documentation for Python3.7, I don't see `Final` being mentioned anywhere. While executing python file with `from typing import Final` under Python3.6 , it raised `ImportError: cannot import name 'Final'`. As a result, this is not really usable? Because this is only valid when we lint using "mypy" and we even need to remove all occurrences of `Final` before execution.

ilevkivskyi:
@yxliang01 It will be available in `typing` starting from Python 3.8, on older versions you should use `typing_extensions`. There is a actually a big blue box saying this in the [docs](https://mypy.readthedocs.io/en/latest/final_attrs.html). Next time please read them first, if you have troubles understanding how to use mypy.

yxliang01:
Thanks for clarification! 
p.s. I don't think I have difficulty on using mypy(besides with this `Final` thing) and I was directly following the repos and found this PR, that's why I missed that doc :(

PR #5533, opened 2018-08-28
Always add object to bases before calculating MRO

### LINKED ISSUE #5527: Mypy fails to calculate MRO in situations with import cycles.
In some situations mypy can't determine MRO, for example consider this test case:
```
[case testMultipleInheritanceCycle]
import b
[file a.py]
from b import B
class A: ...
class C(A, B): ...
class D(C): ...
class Other: ...
[file b.py]
from a import Other
class B: ...
[out]
```
that fails with
```
tmp/a.py:4: error: Cannot determine consistent method resolution order (MRO) for "D"
```
Note that changing order of files in the cycle (for example by importing `a` in `main`) removes the error.

In this case the actual MRO is consistent (works at runtime), while the import cycle is caused by an unrelated class (used for annotations in real stubs).

I think this is high priority because this is just an time bomb waiting to explode in some real code bases.

ilevkivskyi:
Fixes #5527 

The problem is that MRO calculation algorithm may fail if certain corner cases if a base class list is empty (which is the case before class definition is analyzed). This can happen for example in an import cycle. The solution is to always add a temporary `object` base for un-analyzed classes, for them MRO will be anyway recomputed during third pass.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1143,7 +1143,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
             # Give it an MRO consisting of just the class itself and object.
             defn.info.mro = [defn.info, self.object_type().type]
             return
-        calculate_class_mro(defn, self.fail_blocker)
+        calculate_class_mro(defn, self.fail_blocker, self.object_type)
         # If there are cyclic imports, we may be missing 'object' in
         # the MRO. Fix MRO if needed.
         if info.mro and info.mro[-1].fullname() != 'builtins.object':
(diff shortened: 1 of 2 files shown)


mypy/semanal.py
```diff
@@ -1143,7 +1143,7 @@ def analyze_base_classes(self, defn: ClassDef) -> None:
             # Give it an MRO consisting of just the class itself and object.
             defn.info.mro = [defn.info, self.object_type().type]
             return
-        calculate_class_mro(defn, self.fail_blocker)
+        calculate_class_mro(defn, self.fail_blocker, self.object_type)
```
JukkaL: The check just below this call may be redundant now.

ilevkivskyi:
> This is getting a bit complicated. I wonder if we could only calculate the MRO (and perform any MRO-related checks/analysis) in the third pass, avoiding the need to have the optional argument.

I was tempted to just skip the MRO computation in the second pass. But I am afraid to remove things for which I don't understand reasons they exist. MRO may be clearly wrong in the second pass (there is even a comment about this) yet we still do this.

If you are sure there will be no unexpected consequences I cam try just remove it altogether from the second pass. (Btw in this case should we initialize it to `[cls, object]` in the second pass, or just leave empty?)

ilevkivskyi:
(All tests passed long ago when I check on Travis directly, but it is still yellow here, I think we can go ahead.)

gvanrossum:
Yeah, looks like a Travis/GitHub miscommunication. Go ahead!

PR #5543, opened 2018-08-30
Sync typeshed

alexjurkiewicz:
I want this specifically for the 3.7 subprocess.run definition.


ethanhs:
You can `pip install git+https://github.com/python/mypy` and it will install from master. The next version is planned to be released next Friday.

PR #5544, opened 2018-08-31
Fix performance regression caused by #5465: non-daemon part

ilevkivskyi:
The problem with the initial fix in #5465 is that it assumes that all modules that are in `st.suppressed` that now exist in file system are newly added, while in fact they can be not new if one uses `follow_imports = skip`.

My initial guess was that this causes a lot of files become stale. But it actually turns out that the staleness is determined correctly (all the tests I added actually pass on master). However, we still may parse a lot of modules unnecessarily.

This PR fixes the performance regression by:
* Not treating modules as newly added if they were added to `st.suppressed` because of `follow_imports = skip`
* ~~Never processing newly added modules on initial loading during cold daemon run (this is consistent with how it is done in other places where new modules are detected)~~ Moved to a separate PR #5556.

There still some overhead remains -- we need to clone options for module to understand why it got into suppressed dependencies. This can be improved by having to separate suppressed lists, but I think this win will be quite minor and it is not worth the complexity.


mypy/build.py
```diff
@@ -2421,6 +2415,25 @@ def find_module_and_diagnose(manager: BuildManager,
             raise CompileError(["mypy: can't find module '%s'" % id])
 
 
+def find_added_packages(suppressed: List[str],
```
JukkaL: Style nit: The name seems to imply that this returns the added packages. Maybe rename to `exists_added_packages`, for example.

mypy/build.py
```diff
@@ -2421,6 +2415,25 @@ def find_module_and_diagnose(manager: BuildManager,
...
+    """
+    for dep in suppressed:
+        path = manager.find_module_cache.find_module(dep, manager.search_paths,
+                                                     manager.options.python_executable)
+        if not path or dep in manager.source_set.source_modules:
+            continue
+        if (options.follow_imports == 'skip' and
+                (not path.endswith('.pyi') or options.follow_imports_for_stubs)):
+            continue
+        if '__init__.py' in path:
```
JukkaL: This is not 100% correct -- the path could be like `foo/foo__init__.py`, for example (or there could even be a directory named `__init__.py` in the path, though this would almost never happen).

mypy/build.py
```diff
@@ -2421,6 +2415,25 @@ def find_module_and_diagnose(manager: BuildManager,
...
 
 
+def find_added_packages(suppressed: List[str],
+                        manager: BuildManager, options: Options) -> bool:
+    """Find if there are any newly added packages that were previously suppressed.
+
+    Exclude everything not in build for follow-imports=skip.
+    """
+    for dep in suppressed:
+        path = manager.find_module_cache.find_module(dep, manager.search_paths,
```
JukkaL: This same call is performed at least twice. It would be less error-prone to move this in a function and update all identical calls to use the function instead.

mypy/build.py
```diff
@@ -2421,6 +2415,25 @@ def find_module_and_diagnose(manager: BuildManager,
...
+                                                     manager.options.python_executable)
+        if not path or dep in manager.source_set.source_modules:
```
JukkaL: Could we move the `dep in ...` check to happen before the `find_module` call above to speed things up a bit?

mypy/build.py
```diff
@@ -2421,6 +2415,25 @@ def find_module_and_diagnose(manager: BuildManager,
...
+            continue
+        if (options.follow_imports == 'skip' and
```
JukkaL: It's not clear to me what happens when we have follow_imports as 'skip', and we add a pre-existing `__init__.py` file that used to be skipped to the build. Can you add a comment about this somewhere, assuming that it works?

mypy/build.py
```diff
@@ -2695,13 +2708,20 @@ def load_graph(sources: List[BuildSource], manager: BuildManager,
         #   (since direct dependencies reflect the imports found in the source)
         #   but A's cached *indirect* dependency on C is wrong.
         dependencies = [dep for dep in st.dependencies if st.priorities.get(dep) != PRI_INDIRECT]
-        added = [dep for dep in st.suppressed
-                 if manager.find_module_cache.find_module(dep, manager.search_paths,
-                                                          manager.options.python_executable)]
+        if not manager.use_fine_grained_cache():
```
JukkaL: Is this change necessary to fix the performance regression, or is this just a somewhat related cleanup? If this is not required, I'd propose to move this into a separate PR, since the import suppression logic is very delicate and we should do whatever we can to make reasoning about changes easier.

test-data/unit/check-incremental.test
```diff
@@ -4884,3 +4884,60 @@ def f(x: str) -> None: pass
...
+[file other.py.2]
+x = 1
+[stale]
+[rechecked]
+
+[case testFollowImportSkipInvalidatedOnAddedStub]
+# flags: --follow-imports=skip --ignore-missing-imports
+# cmd: mypy -m main
+[file main.py]
+import other
```
JukkaL: Should we have test cases that use `from pkg import foo`, where `pkg` either is a package that exists but is skipped, or gets created? `pkg.foo` is a module.

mypy/build.py
ilevkivskyi: Yes, but this doesn't affect correctness, only performance. Anyway, I will try co find a more precise test.

mypy/build.py
ilevkivskyi: Yes this will save some ms.

test-data/unit/check-incremental.test
ilevkivskyi: There is already a test when `pkg` was actually missing and gets created, see `testAddedMissingStubsPackageFrom`. I will add a test for when it was just skipped.

mypy/build.py
ilevkivskyi: Performance. Cloning options for a lot of files may take time. For example in S repo this could be 5-10 sec, which is negligible for normal incremental run, but is significant for daemon cold start.

mypy/build.py
ilevkivskyi: We don't need to add any special logic for this. If a module is added to build, importers will be invalidated by normal mechanism. The test you proposed to add will check this.

(I will add a comment anyway, after I finish with `Final`.)

mypy/build.py
gvanrossum: Hm, I put a cache in for that long ago. Did the rewrite for mypyc break that again?

ilevkivskyi:
@JukkaL It is hard to make performance test for this, because it is hard to find an old commit that doesn't crash in S (mostly because of SQLAlchemy related code). Here are the results after few attempts:
* This PR: `2m09s`
* Regression pin: `5m03s`
* An older commit: `5m47s` (variance is high here)

This looks surprising, it means that this PR not just brings performance to the pre-regression level, but makes it faster (I think however this is just a measurement effect).

mypy/build.py
```diff
@@ -2363,6 +2357,34 @@ def find_module_and_diagnose(manager: BuildManager,
...
+            # We don't need to add any special logic for this. If a module
+            # is added to build, importers will be invalidated by normal mechanism.
+            continue
+        path = find_module_simple(dep, manager)
+        if not path:
+            continue
+        if (options.follow_imports == 'skip' and
+                (not path.endswith('.pyi') or options.follow_imports_for_stubs)):
+            continue
+        if path.split(os.path.sep)[-1] in ('__init__.py', '__init__.pyi'):
```
JukkaL: Using `str.split` for paths can be problematic, since Windows allows both `/` and `\\` as path separators, and it can be hard to ensure that the same separator is always used. On reflection, the original test seems better, if you add a comment saying that we don't rely on the test for correctness. In my previous review I didn't realize that the test can be more general than strictly necessary. 

The original test was too lenient, which is better than being too strict like this test, and implementing a perfect test doesn't seem worth it.

ilevkivskyi:
(close+re-open to re-trigger AppVeyor)

ilevkivskyi:
OK, it looks like AppVeyor is broken, it fails immediately with a weird single line log
```
{"message":"Value cannot be null.\r\nParameter name: content"}
```

ilevkivskyi:
I am going to merge this when Travis passes (unless there are objections).

PR #5552, opened 2018-09-01
Avoid type-var collisions in attrs and dataclasses

### LINKED ISSUE #5542: attrs plugin error: "The erased type of self '...*' is not a supertype of its class" on class definition with a generic using a constrained type variable, but then seems to work fine after definition
Consider the following example:

```python3
import attr

class V:
    pass

class W:
    pass

T = TypeVar("T", V, W)

@attr.s(auto_attribs=True)
class A(Generic[T]):
    v: T

reveal_type(A(W()))
```

My expectation here would be a revealed type of `A[W*]`, which I *do* get.

But I also get an error on the `@attr.s` definition:
```
The erased type of self 'V*' is not a supertype of its class 'A[T`1]'
The erased type of self 'W*' is not a supertype of its class 'A[T`1]'
```

if I remove the constraints on `T`, (i.e. `T = TypeVar("T")`) then the errors go away.

euresti:
The issue was caused because we were using 1 for the ids of the generated types.
Since by convention TypeVars for methods should have negative numbers and we're creating these methods we're guaranteed to have -1 free.  So use that instead.

Fixes #5542

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -458,7 +458,7 @@ def _add_cmp(ctx: 'mypy.plugin.ClassDefContext', adder: 'MethodAdder') -> None:
     #    AT = TypeVar('AT')
     #    def __lt__(self: AT, other: AT) -> bool
     # This way comparisons with subclasses will work correctly.
-    tvd = TypeVarDef('AT', 'AT', 1, [], object_type)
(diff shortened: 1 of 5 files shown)


ilevkivskyi:
(Also it looks like you still need to update revealed type in one more test.)

euresti:
Ok.  New test added and tests pass!

PR #5556, opened 2018-09-01
Fix performance regression caused by #5465: daemon part

ilevkivskyi:
This is part of the original PR https://github.com/python/mypy/pull/5544 that takes care of the daemon performance regression caused by #5465. (I separated it from original PR as requested by @JukkaL.)

I will add tests I promised in https://github.com/python/mypy/pull/5544 (daemon variants of the tests added in #5465 and #5544).

This change should not affect semantics, only performance by avoiding cloning options for lots of modules in situation with `--follow-imports=skip`.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2369,6 +2369,12 @@ def find_module_and_diagnose(manager: BuildManager,
             raise CompileError(["mypy: can't find module '%s'" % id])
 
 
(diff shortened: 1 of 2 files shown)


ilevkivskyi:
@JukkaL I added the tests that you asked (just fine grained versions of previously added normal incremental tests). It looks like there is only one problematic test but it is not affected by any of the recent changes.

I will make performance measurements later on Monday.

Issue #5534, opened 2018-08-28
Importing packages can poison the `mypy` cache.


TheKevJames:
Ok -- this is a weird one, but I think I've got it down to the smallest possible repro.

I first noticed this issue with the `aiohttp` library as of version `3.4.0` (when they introduced type-hinting), but I'm cross-posting this issue to both projects since the issue seems to run deeper than just a library issue. Importing the `aiohttp` library, even if it is unused, causes the `mypy` cache to occasionally get poisoned.

Here's the repro case:
```
import aiohttp
def trigger() -> dict:
    return 'foo'
```

Running `python -m mypy repro.py` against this file works as expected (eg. returns an incompatible return type error). Replacing the `-> dict` with `-> str` throws the following assertion (attached at bottom of report) on re-running the type check. Note that this holds true in the opposite direction -- if we begin with `-> str`, the first typecheck is successful, but replacing `-> str` with `-> dict` causes the same assertion error. Either way, the situation can only be resolved by clearing the `.mypy_cache` folder, at which point `mypy` acts as expected once more.

Interestingly, removing the `aiohttp` import prevents this poisoning from occurring. In addition, removing the import once the cache is poisoned also fixes the issue.

I speculate there are two issues here, one with `aiohttp` and one with `mypy`:
- `aiohttp` seems to have some weirdness going on in their type hinting
- regardless of what a library does, the above repro case should not cause `mypy` issues when dealing only with stdlib types.

Error Trace:
```
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/kevin/coding/mypy-bug/venv/lib/python3.7/site-packages/mypy/__main__.py", line 11, in <module>
    main(None)
  File "/Users/kevin/coding/mypy-bug/venv/lib/python3.7/site-packages/mypy/main.py", line 91, in main
    res = type_check_only(sources, bin_dir, options, flush_errors, fscache)  # noqa
  File "/Users/kevin/coding/mypy-bug/venv/lib/python3.7/site-packages/mypy/main.py", line 148, in type_check_only
    fscache=fscache)
[...]

PR #5565, opened 2018-09-03
Original node wins in case of a redefinition (especially if it is a class)

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5534

In principle we can be even more strict (see a little exception for class scope) but then we need to work on https://github.com/python/typeshed/issues/2423 first.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -387,10 +387,12 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
             #   a common stack of namespaces. As the 3 kinds of namespaces have
             #   different semantics, this wouldn't always work, but it might still
             #   be a win.
+            #   Also we can re-use some logic in self.add_symbol().
             if self.is_class_scope():
                 # Method definition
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
I submitted python/typeshed#2424 to fix the bad stub. Probably better to get that fix in first so mypy can be more consistent.

mypy/semanal.py
```diff
@@ -3301,13 +3307,15 @@ def add_symbol(self, name: str, node: SymbolTableNode,
...
                     self.name_already_defined(name, context, existing)
+                    return
             self.globals[name] = node
 
     def add_local(self, node: Union[Var, FuncDef, OverloadedFuncDef], ctx: Context) -> None:
         assert self.locals[-1] is not None, "Should not add locals outside a function"
         name = node.name()
         if name in self.locals[-1]:
             self.name_already_defined(name, ctx, self.locals[-1][name])
+            return
```
JukkaL: Is there a test case for this?

PR #5569, opened 2018-09-04
 Call os._exit() for faster shut down in the normal case

JukkaL:
This can speed up self checking by about a second. We still sometimes
call normal `sys.exit()` when performance isn't important or when somebody
might expect a `SystemExit` exception.


gvanrossum:
I worry that we might encounter additional use cases that this breaks. Wouldn't it be safer to have a command-line flag "--hard-exit" that we can turn on in our internal scripts?

JukkaL:
This can have a big performance impact in the future once we migrate to a mypyc-compiled mypy -- around 20% for non-incremental runs. The impact is sufficiently big that I'd rather have this enabled by default. If this breaks some use cases, we can deal with them once they get reported. A 20% performance gain makes it worth it, I think. And if there are rare issues, I'd lean towards providing a `--clean-exit` option instead of making the fast way opt-in.

`mypy.api` is the only supported way of invoking mypy within an application and it already preserves backward compatibility. Normally invoking mypy as a process should have no observable change in semantics, or if there is a difference, it's likely caused by a bug that we can fix (such as not closing a file within mypy).

Some plugin might theoretically be affected, but the plugin API is undocumented and any changes needed to plugins should hopefully be small. If there is a real need for some plugin to do something at process exit, we can provide additional exit hooks that are guaranteed to be run even when doing a hard exit.

gvanrossum:
I still think an unsafe optimization should not be on by default.

I actually have a use case in internal code (I'll point you to it on Slack).

JukkaL:
Ok, let's disable it by default. We can reconsider this in the future once we've been using the option for a while internally.

msullivan:
Maybe a good approach would be to implement the hard exit in `__main__.py`, which nobody should be calling into?

JukkaL:
We can can start with hiding it behind `--fast-exit` (now implemented). The performance difference is pretty minor until we are using a compiled mypy anyway.

Issue #5576, opened 2018-09-05
`-2` and `--python-version 2.7` arguments differ in behaviour


sqwishy:
This is on mypy 0.620 on Python 3.6.

My understanding is that the command arguments `-2` and `--python-version 2.7` are
supposed to do the same thing. But in `infer_python_version_and_executable()`
python_version is different on the given options and I think this affects how it finds a
value for `python_executable`.

If I'm understanding this correctly, I think it might be an issue with the arguments
having different `dest` values? `-2` sets a constant on the wrong namespace?

```
> python3 -m pdb (which mypy) -2 .
> /home/sqwishy/.local/bin/mypy(4)<module>()
-> import re
(Pdb) b mypy/main.py:278
Breakpoint 1 at /home/sqwishy/.local/lib/python3.6/site-packages/mypy/main.py:278
(Pdb) c
> /home/sqwishy/.local/lib/python3.6/site-packages/mypy/main.py(278)infer_python_version_and_executable()
-> if special_opts.python_executable is not None and special_opts.python_version is not None:
(Pdb) pp vars(special_opts)
...
 'python_executable': None,
 'python_version': None,
 'silent_imports': False,
...
```
```
> python3 -m pdb (which mypy) --python-version 2.7 .
> /home/sqwishy/.local/bin/mypy(4)<module>()
-> import re
(Pdb) b mypy/main.py:278
Breakpoint 1 at /home/sqwishy/.local/lib/python3.6/site-packages/mypy/main.py:278
(Pdb) c
> /home/sqwishy/.local/lib/python3.6/site-packages/mypy/main.py(278)infer_python_version_and_executable()
-> if special_opts.python_executable is not None and special_opts.python_version is not None:
(Pdb) pp vars(special_opts)
...
 'python_executable': None,
 'python_version': (2, 7),
 'silent_imports': False,
...
```

I would expect `special_opts.python_version` to be the same in both cases.

gvanrossum:
Why are you digging in mypy's internals? Is there any externally observable behavior that differs?

sqwishy:
I think the site-packages ends up being different. I could be wrong. I'm just not entirely sure how to show this. Sorry for being unclear.

gvanrossum:
OK, I'll wait until you have something more concrete.

sqwishy:
[...]

PR #5578, opened 2018-09-06
Update -2 argument dest to match --python-version.

sqwishy:
And check config for python_version when looking for a python
executable.

---

See issue #5576.

I didn't add any automated tests because I wasn't sure what you guys would have wanted for that. This stuff makes use of subprocess so either that'd be mocked out or it would depend on python2.7 being installed? I'm not sure.

And I'm uncertain that changing `infer_python_version_and_executable()` as I did is a good idea. Or if you'd prefer that `python_executable` on the `special_opts` namespace be set, if it's `None`, to the value of that attribute on the `Options` instance somewhere in `process_options()` before calling `infer_python_version_and_executable()`?

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -277,27 +277,31 @@ def infer_python_version_and_executable(options: Options,
     This function mutates options based on special_opts to infer the correct Python version and
     executable to use.
     """
+    # Check Options in case python_version is set in a config file, but prefer command
+    # line arguments.
+    python_version = special_opts.python_version or options.python_version
+
     # Infer Python version and/or executable if one is not given
 
     # TODO: (ethanhs) Look at folding these checks and the site packages subprocess calls into
     # one subprocess call for speed.
-    if special_opts.python_executable is not None and special_opts.python_version is not None:
+    if special_opts.python_executable is not None and python_version is not None:
         py_exe_ver = _python_version_from_executable(special_opts.python_executable)
-        if py_exe_ver != special_opts.python_version:
+        if py_exe_ver != python_version:
             raise PythonExecutableInferenceError(
                 'Python version {} did not match executable {}, got version {}.'.format(
-                    special_opts.python_version, special_opts.python_executable, py_exe_ver
+                    python_version, special_opts.python_executable, py_exe_ver
                 ))
         else:
-            options.python_version = special_opts.python_version
+            options.python_version = python_version
             options.python_executable = special_opts.python_executable
-    elif special_opts.python_executable is None and special_opts.python_version is not None:
-        options.python_version = special_opts.python_version
+    elif special_opts.python_executable is None and python_version is not None:
+        options.python_version = python_version
         py_exe = None
         if not special_opts.no_executable:
-            py_exe = _python_executable_from_version(special_opts.python_version)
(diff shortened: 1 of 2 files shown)


ethanhs:
I will take a closer look later today on the contents. As for testing, I think the best place to add it is here: https://github.com/python/mypy/blob/master/mypy/test/testargs.py#L26.

sqwishy:
@ethanhs Yeah I had looked at that earlier, for some reason I thought I'd need to test where `options.python_version = (2, 7)` and it would then depend on python 2.7 being installed. But I'm pretty sure I can use sys.version_info like the rest of that test. So I'll try that.

And if you decide that `infer_python_version_and_executable()` probably shouldn't check both objects for the python_version (that the caller should make up its mind and put those objects in a more agreeable state), then we can come up with a different test.

ethanhs:
@sqwishy ah, you probably want to use https://github.com/python/mypy/blob/master/mypy/util.py#L101 which will return `None` if it cannot find a Python 2 interpreter.

sqwishy:
@ethanhs I could. it might simplify the test a bit. But then it wouldn't run if someone didn't have a Python 2. Let me know if you'd prefer that over one of the previous revisions I made for tests.

ethanhs:
If there are no objections I will merge this later today (or someone else can go ahead).

ilevkivskyi:
It looks like this PR broke PEP 561 packages because previously when checking Python 2 code, mypy looked for stub packages installed in the place where mypy is installed. But after this PR it only looks for PEP 561 packages in Python 2 site packages.

I am not sure what PEP 561 says about this, but the previous behavior was intuitive an useful.

sqwishy:
@ilevkivskyi I'm probably misunderstanding but it sounds like you're saying you have mypy stub python packages installed in the system's python 3 site-packages but you want them available in a python 2 virtualenv? I'd imagine you would want to install those typing stubs along side the package they're for in the virtualenv, right?

ilevkivskyi:
A (very) common scenario is where you have a CI server that type-checks some Python 2 repository with mypy. This server might not even have Python 2 available (some containers in Travis CI for example). The only reliable way to install PEP 561 stub packages in such scenario is to install them in the same place where mypy is installed. Before this PR this worked perfectly, but with the PR it doesn't.

In addition to this problem, typeshed CI is currently broken, so they can't merge any PRs (the root cause may however be the same).

I don't want to say the PR is bad, I just want to say we need to carefully discuss what do here. We can re-apply when we will have a clear way forward.

@ethanhs What do you think about this?

JukkaL:
> I just want to say we need to carefully discuss what do here. 

Any changes that break a common use case should be discussed in an issue, not only in a PR. Also, I don't think that the PR title and description quite describe what's changed. It looks to me like there are two separate changes in this PR, and if that's the case, there should also be two PRs (and two issues).

JukkaL:
> I be glad to open a new PR for just the change to argument parsing (a) to address #5576. And create a new issue here for behaviour (b).

Sounds good! (b) clearly is a somewhat tricky issue and worth some more thought.

sqwishy:
@JukkaL I've created issue #5620 for the problem about using Python 2 site-packages. And pull request #5619 to just adjust the argument handling.

PR #5581, opened 2018-09-06
A handful of hacks to appease mypyc

msullivan:
Some of these are kind of unpleasant, since I saved the best for last.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -62,6 +62,8 @@ from mypy.typeanal import make_optional_type
 
 from mypy import experiments
 
+import mypy.type_visitor
+
 # Type of callback user for checking individual function arguments. See
 # check_args() below for details.
 ArgChecker = Callable[[Type, Type, int, Type, int, int, CallableType, Context, MessageBuilder],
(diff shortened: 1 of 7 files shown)


mypy/checkexpr.py
```diff
@@ -62,6 +62,8 @@
 
 from mypy import experiments
 
+import mypy.type_visitor
```
ilevkivskyi: Why do we need this import? (It looks unused, so a comment would help if it is really needed.)

mypy/dmypy_server.py
```diff
@@ -39,7 +39,8 @@ def daemonize(func: Callable[[], None], log_file: Optional[str] = None) -> int:
     subprocess killed by signal.
     """
     # See https://stackoverflow.com/questions/473620/how-do-you-create-a-daemon-in-python
-    if sys.platform == 'win32':
+    # mypyc doesn't like unreachable code, so trick mypy into thinking the branch is reachable
+    if sys.platform == 'win32' or 0 == 1:
```
ilevkivskyi: Ultra minor nit: I would write `bool()` here and below. (Although even better I would switch to checking unreachable code, but this is a separate question.)

mypy/checkexpr.py
msullivan: Actually whatever it was seems fixed soooo

PR #5582, opened 2018-09-06
Support compiling the plugin infrastructure with mypyc

msullivan:
The tricky part here is that fundamentally the plugin infrastructure needs to work
with classes that were not compiled by mypyc. Since mypyc doesn't currently
allow interpreted classes to inherit from mypyc classes, this poses a problem.

The solution is to introduce an `InterpretedPlugin` class that is not compiled by
mypyc that we arrange for interpreted modules to see as `Plugin`. Interpreted
plugins can safely subclass from this module, which arranges to wrap instances
of it in a `WrapperPlugin` class that inherits from `Plugin` and proxies to the
`InterpretedPlugin`.

There is a slightly different approach that would have worked where we do most
of the same stuff but keep `Plugin` interpreted and add an `InternalPlugin` instead
of relying on magic to expose `InterpretedPlugin` to interpreted classes, but I decided 
against that because it exposes this nonsense to more parts of mypy, and I would like
to rip it all out eventually anyways.

This is all kind of unfortunate.

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -1,5 +1,9 @@
 """Plugin system for extending mypy."""
 
+import sys
+import os.path
+import types
+
 from abc import abstractmethod
 from functools import partial
 from typing import Callable, List, Tuple, Optional, NamedTuple, TypeVar, Dict
@@ -16,7 +20,7 @@ from mypy.types import (
 )
 from mypy.messages import MessageBuilder
 from mypy.options import Options
-
+import mypy.interpreted_plugin
 
 @trait
 class TypeAnalyzerPluginInterface:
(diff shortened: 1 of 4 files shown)


mypy/plugin.py
```diff
@@ -1,5 +1,9 @@
 """Plugin system for extending mypy."""
 
+import sys
+import os.path
```
ilevkivskyi: It looks like these two imports are also unused. Is this also for mypyc or you just forgot to remove them?

mypy/plugin.py
```diff
@@ -444,3 +500,14 @@ def int_pow_callback(ctx: MethodContext) -> Type:
         else:
             return ctx.api.named_generic_type('builtins.float', [])
     return ctx.default_return_type
+
+
+# This is an incredibly frumious hack. If this module is compiled by mypyc,
+# set the module 'Plugin' attribute to point to InterpretedPlugin. This means
+# that anything interpreted that imports Plugin will get InterpretedPlugin
+# while anything compiled alongside this module will get the real Plugin.
+if isinstance(int_pow_callback, types.BuiltinFunctionType):
```
ilevkivskyi: Is it possible to write this as something like `if MYPYC: ...` where `MYPYC` is the magic constant that is `False` in reality but treated as `True` by mypyc?


(Not need to do this now but will be useful in future I think.)

mypy/plugin.py
msullivan: Oh, this is from a legacy of trying a few different things for the `if MYPYC` check below, I can remove them

mypy/plugin.py
msullivan: That is something we should have but don't right now

PR #5584, opened 2018-09-07
Sync typeshed

msullivan:


msullivan:
A code fix is needed for this sync to work. #5585 includes both a workaround for the typeshed error and the code fix needed.

PR #5595, opened 2018-09-11
Reformat and reorganize the config file docs

Michael0x2a:
Sorry again for another big-ish docs PR. This time, I'm cleaning up the config file docs.

Here are a list of notable changes made:

1.  I moved the example to come before we list the options, not after.

    I feel this makes the first section of the page more digestible: we have a big block of text, but at least now we can see some of that  information in practice immediately after.

2.  I combined the two examples into one (and rewrote the explanations in list form, to help mitigate the increase in complexity).

3.  I swapped the global-only and per-module sections.

    After auditing the options, it seemed to me that the global-only options were generally less useful then the per-module options. I also think it makes sense to present the more flexible options first, rather then last.

4.  I organized the options into the same categories I introduced when organizing mypy's command line interface.

5.  I switched to using the same definition list format we use in the command line docs for consistency.

6.  I removed the `dump_type_stats`, `dump_inference_stats`, `debug_cache`, and the `strict_boolean` options.

    These are all options we have either suppressed or deprecated.

7.  I added in documentation for the `python_executable`, `no_site_packages`, and `disallow_untyped_decorators` options.

8.  I renamed some links/deleted some now-unnecessary links.

One change I made that was unrelated to the config file was adding docs about the `mypy_path` and the per-module `ignore_missing_imports` options to the "Running mypy" page.

Some interesting inconsistencies I noticed (but took no action on, in a probably-futile attempt to keep this diff small):

1.  The `warn_redundant_casts` section is global-only, not per-module like the other warning options.

    I left this option in the global misc section for now.

2.  There does not appear to be any way of writing a config file that does the same thing as `--no-site-packages`.

    I think we should either add an option for this or allow expressions like `python_executable = None`. (Currently, we try parsing `None` as a string -- this makes mypy promptly crash with a FileNotFoundError). For now, I did nothing.

Final note: I mostly kept the descriptions the same/minimal, apart from some minor wording tweaks here and there -- I wanted to have the command line docs act as the main source of truth.

### DIFF
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -32,7 +32,7 @@ characters.
 - Additional sections named ``[mypy-PATTERN1,PATTERN2,...]`` may be
   present, where ``PATTERN1``, ``PATTERN2``, etc., are comma-separated
   patterns of fully-qualified module names, with some components optionally
-  replaced by `*`s (e.g. ``foo.bar``, ``foo.bar.*``, ``foo.*.baz``).
+  replaced by the '*' character (e.g. ``foo.bar``, ``foo.bar.*``, ``foo.*.baz``).
   These sections specify additional flags that only apply to *modules*
   whose name matches at least one of the patterns.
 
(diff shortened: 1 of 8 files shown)


Michael0x2a:
You can also view a temporary version of these docs here: https://michael0x2a-mypy.readthedocs.io/en/reorganize-config-file-docs/config_file.html

docs/source/config_file.rst
```diff
@@ -65,234 +65,346 @@ unfortunate, and is subject to change in future versions.
...
 
-- ``warn_incomplete_stub`` (Boolean, default False) warns for missing
-  type annotation in typeshed.  This is only relevant in combination
-  with ``disallow_untyped_defs`` or ``disallow_incomplete_defs``.
+    [mypy-somelibrary]
+    ignore_missing_imports = True
 
-- ``warn_redundant_casts`` (Boolean, default False) warns about
-  casting an expression to its inferred type.
+If you place this file at the root of your repo and run mypy, mypy will:
```
gvanrossum: I wonder if each bullet should just also quote the text from the example (e.g. `python_version = 2.7`) which it describes, to make it clear what the descriptions refer to?

docs/source/config_file.rst
```diff
@@ -65,234 +65,346 @@ unfortunate, and is subject to change in future versions.
...
-  (Windows).
+The following config options may be set either globally (in the ``[mypy]`` section)
+or on a per-module basis (in sections like ``[mypy-foo.bar]``).
 
-- ``quick_and_dirty`` (Boolean, default False) enables :ref:`quick
-  mode <quick-mode>`.
+If you set an option both globally and for a specific module, the module configuration
+options take precedence. This lets you set global defaults and override them on a
+module-by-module basis. If multiple pattern sections match a module, the options from the
+most specific section are used where they disagree. See above for more details.
```
gvanrossum: Add a deep link to the section referred to? (`When options conflict, the precedence order for the configuration sections is:`)

docs/source/config_file.rst
```diff
@@ -65,234 +65,346 @@ unfortunate, and is subject to change in future versions.
...
 
-Per-module flags
-****************
+``ignore_missing_imports`` (boolean, default False) 
+    Suppress error messages about imports that cannot be resolved.
 
-The following flags may vary per module.  They may also be specified in
-the global section; the global section provides defaults which are
-overridden by the pattern sections matching the module name.
+    Note that if pattern matching is used, the pattern should match the
```
gvanrossum: The term "pattern matching" seems to be used only here and for the next option. Maybe instead use a phrase like "if this option is used in a per-module section, the module name pattern should match the [etc.]"?

docs/source/config_file.rst
```diff
@@ -65,234 +65,346 @@ unfortunate, and is subject to change in future versions.
...
-- ``ignore_missing_imports`` (Boolean, default False) suppress error
-  messages about imports that cannot be resolved.  Note that if
-  pattern matching is used, the pattern should match the name of the
-  *imported* module, not the module containing the import statement.
+    Used in conjunction with ``follow_imports=error``, this can be used
+    to make any use of a particular ``typeshed`` module an error.
 
-- ``silent_imports`` (Boolean, deprecated) equivalent to
-  ``follow_imports=skip`` plus ``ignore_missing_imports=True``.
+Disallow Any
```
gvanrossum: Maybe give this section a loftier title, e.g. "Disallowing `Any` in various contexts"? Or at least "Disallowing `An`"? I suppose the corresponding section in command_line.rst should also be renamed then, and the label used to link to it.

docs/source/config_file.rst
```diff
@@ -65,234 +65,346 @@ unfortunate, and is subject to change in future versions.
...
+section of the command line docs.
 
-.. _per-module-flags:
+Note: this section describes options that can be used both globally and per-module.
+See below for a list of import discovery options that may be used
+:ref:`only globally <config-file-import-discovery-global>`.
 
-Per-module flags
-****************
+``ignore_missing_imports`` (boolean, default False)
```
gvanrossum: Why'd you choose to change `Boolean` to `boolean`? It's traditionally capitalized because it's named after George Boole. (Though `bool` is not, because it's an abbrev. :-)

docs/source/config_file.rst
```diff
@@ -65,234 +65,346 @@ unfortunate, and is subject to change in future versions.
...
-  defining functions without type annotations or with incomplete type
-  annotations.
+``disallow_untyped_defs`` (boolean, default False)
+    Disallows defining functions without type annotations or with incomplete type
+    annotations.
 
-- ``check_untyped_defs`` (Boolean, default False) type-checks the
-  interior of functions without type annotations.
+``disallow_incomplete_defs`` (boolean, default False)
+    Disallow defining functions with incomplete type annotations.
```
gvanrossum: Can you be consistent about "disallow" vs. "disallows"? I think most places use "disallows", "type-checks", reports" etc., so I'd go with that.

docs/source/config_file.rst
Michael0x2a: I tried doing this, but unfortunately it made the explanations look a little cluttered. I decided instead to group the explanations about the global options into one group and the per-module options into another. This way, only needs to read three bullet points at a time, instead of 6 -- hopefully that helps.

docs/source/config_file.rst
Michael0x2a: Done!

docs/source/config_file.rst
Michael0x2a: Done!

docs/source/config_file.rst
Michael0x2a: I decided to rename this section and the associated section in the command line page to "Disallowing dynamic typing". (I also renamed the section in the `--help` text to "Dynamic typing").

Hopefully this is more loftier? At the very least, it's a little less jargon-y.

docs/source/config_file.rst
Michael0x2a: Oh, mostly because I liked the consistency of having all the types start with a lowercase letter. I went ahead and changed the type to "bool". (The only downside now is that the types are inconsistently abbreviated now -- I guess you can't win them all 🙃)

docs/source/config_file.rst
Michael0x2a: Whoops, thanks for the catch! Fixed.

Michael0x2a:
@gvanrossum -- this should be ready for a second look whenever!

docs/source/config_file.rst
```diff
@@ -102,17 +113,21 @@ If you place this file at the root of your repo and run mypy, mypy will:
...
     ``mycode/foo`` directory.
     
-5.  Selectively *disable* the "function is returning any" warnings within only
-    ``mycode.bar``. This overrides the global default.
-
-6.  Suppress any error messages generated when your codebase tries importing the module 
-    ``somelibrary``. (We assume here that ``somelibrary`` is some 3rd party
-    library missing type hints).
+2.  Selectively *disable* the "function is returning any" warnings within only
+    ``mycode.bar``. This overrides the global default we set earlier.
```
gvanrossum: Nit: IMO the first sentence reads better if you put "only" at the end.

docs/source/config_file.rst
```diff
@@ -124,8 +139,8 @@ or on a per-module basis (in sections like ``[mypy-foo.bar]``).
 
 If you set an option both globally and for a specific module, the module configuration
 options take precedence. This lets you set global defaults and override them on a
-module-by-module basis. If multiple pattern sections match a module, the options from the
-most specific section are used where they disagree. See above for more details.
+module-by-module basis. If multiple pattern sections match a module, the :ref:`options from the
```
gvanrossum: Nit: might as well move `the` into the :ref: link...

docs/source/config_file.rst
```diff
@@ -336,20 +352,21 @@ Incremental mode
 For more information, see the :ref:`incremental mode <incremental>`
 section of the command line docs.
 
-``incremental`` (boolean, default True) 
+``incremental`` (bool, default True) 
     Enables :ref:`incremental mode <incremental>`.
 
 ``cache_dir`` (string, default ``.mypy_cache``) 
-    The default location mypy will store incremental cache info in.
+    Specifies the location mypy will store incremental cache info in.
```
gvanrossum: "Specifies the location where mypy stores incremental cache info."

gvanrossum:
Also, sphinx gave me these errors:
```
/Users/guido/src/mypy/docs/source/command_line.rst:181: WARNING: undefined label: dynamic-typing (if the link has no caption the label must precede a section header)
/Users/guido/src/mypy/docs/source/revision_history.rst:86: WARNING: undefined label: disallow-any (if the link has no caption the label must precede a section header)
/Users/guido/src/mypy/docs/source/revision_history.rst:112: WARNING: undefined label: disallow-any (if the link has no caption the label must precede a section header)
```

Michael0x2a:
@gvanrossum -- ok, I think this is ready for a third look!

Btw, after running `make clean` and rebuilding, it looked like there was also an unrelated error with the "installed packages" doc ("mypy/docs/source/installed_packages.rst:4: WARNING: Duplicate explicit target name: "pep 561") so I went ahead and fixed that too.

Michael0x2a:
Oh lol, should have checked before commenting, whoops

Issue #5514, opened 2018-08-20
Perf regression caused by #5474


msullivan:
It looks likes #5474 caused a perf regression to the tune of 5% on a self check and in our internal repos.

@ilevkivskyi said he had an idea for fixing

PR #5604, opened 2018-09-11
Fix small regression in subtype cache performance

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5514

The root cause is that https://github.com/python/mypy/pull/5474 added `type_parameter_checker` (a function) as one of the keys of subtype kinds. This function is a nested one, so in fact a _new_ function was created on every call to `is_subtype_ignoring_tvars()` causing many thousands parasitic keys created in subtype caches. In addition there are two smaller things listed below.

The performance win is around 5%
This PR:
* Moves the mentioned function to global scope, this causes only 6 subtype kinds present during self-check.
* Prevents similar regressions in future by switching only boolean flags as subtype kind keys.
* Removes two one liners that stay in a very hot code path (cache lookup is hit 140K times during self-check).
* Avoids creation of empty sets on cache look-up, create them only when writing to cache.

### DIFF
--- a/mypy/subtypes.py
+++ b/mypy/subtypes.py
@@ -35,16 +35,21 @@ TypeParameterChecker = Callable[[Type, Type, int], bool]
 
 def check_type_parameter(lefta: Type, righta: Type, variance: int) -> bool:
     if variance == COVARIANT:
-        return is_subtype(lefta, righta, check_type_parameter)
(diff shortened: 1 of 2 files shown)


ilevkivskyi:
I will cherry-pick this into release branch after merging.

PR #5617, opened 2018-09-13
Allow assignment to an empty tuple

### LINKED ISSUE #5594: Can't assign to ()
Mypy incorrectly complains about assignments to `()`. The simplest example of this is
```python
() = []
```
This code executes without a problem, but mypy says `error: can't assign to ()`

A more real world use case might be code like this
```python
(a, b), () = [[1, 2], []]
```
In this case the assignment to `()` is a concise way of asserting that the second list is empty.

`error: can't assign to ()`
No type checking errors
A build from the latest commit as of writing this, 614090b7
* What are the mypy flags you are using?
No flags

If this is a change you decide would be good then I believe it could be implemented by removing the  the check on line 1930 of `mypy/semanal.py` and updating the tests.

tyehle:
Fixes #5594. Simply removes the check in the semantic analyzer. I made a best guess on what to do about the tests.


test-data/unit/semanal-errors.test
```diff
@@ -373,11 +373,6 @@ main:1: error: can't assign to literal
 [out]
 main:1: error: can't assign to literal
 
-[case testInvalidLvalues5]
```
JelleZijlstra: Shouldn't this still be an error saying `1 is not iterable` or similar?

Also, I would like to see a new test that does things like `() = []` or `a, () = list()` and asserts that there are no errors.

test-data/unit/semanal-basic.test
```diff
@@ -47,6 +47,14 @@ MypyFile:1(
       NameExpr(f [__main__.f])
       Args())))
 
+[case testAssignEmptyTuple]
```
JelleZijlstra: Not sure we need this test; these tests that print out internal data structures are harder to maintain and harder to map to user-visible behavior.

tyehle:
Hmm. I did some more tests, and it looks like `() = []` produces a syntax error all the way up through python 3.5. Is I'm not sure how to write tests that check for different behavior on different versions of python. Is this even possible?

JelleZijlstra:
You can pass mypy's `--python-version` flag in the test by putting `# flags: --python-version 3.6` as the first line.

msullivan:
@tyehle do you think you'll have a chance to return to this?

tyehle:
@msullivan I forgot about this. I think I'll take another crack at it :)

tyehle:
I can't figure out how to make these two tests pass

```
() = 1  # E: 'builtins.int' object is not iterable
```
I don't know what is required to make the tests realize that `int` isn't an instance of `Iterable`. It throws `KeyError: 'typing'` which I suspected was related to the way the tests know about the typing module, or possibly the fact that the function `type_is_iterable` is checking `is_subtype(type, self.named_generic_type('typing.Iterable', ...`.

```
# flags: --python-version 3.5
() = []  # E: can't assign to ()
```
The test passing with python 3.5 makes sense because I removed that check. I'm honestly not sure why it correctly produces an error when given `--python-version 2.7`.

msullivan:
Sorry, I missed your comments here and dropped this.

When does it throw `KeyError: 'typing'`? If you try to run that test after your change?

tyehle:
Yeah it throws `KeyError: 'typing'` after the change. I would expect it to throw an error complaining that ints aren't iterable instead.

msullivan:
I apologize that I took a year and a half to look at this again. Adding a `builtins` helped with the stupid typing thing. We run a lot of tests in a pretty broken mypy environment for speed, unfortunately. I moved the tests back into an existing file.

Thanks.

PR #5623, opened 2018-09-16
Add Final annotations to mypy code itself

ilevkivskyi:
This adds `Final` annotations to most module-level and class-level names I have found. Some of these are important to get performance boost after compiling with mypyc, other are just added for completeness. This PR also makes two runtime changes:
* `TypeOfAny` switched from an enum to using small integer constants, this is important for performance
* `binder.BindableTypes` is now inlined in two `isinstance()` checks. I first made it `Final`, but then realised that it will be much more efficient to just inline it, since these `isinstance()` checks are in a hot code path.

Note: after a new version of `typing_extensions` is released on PyPI we can skip all the `if MYPY: ...` checks.

The corresponding mypyc PR is https://github.com/JukkaL/mypyc/pull/439

### DIFF
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 
 MYPY = False
 if MYPY:
-    from typing import DefaultDict, Final
+    from typing import DefaultDict
 
 from mypy.types import Type, AnyType, PartialType, UnionType, TypeOfAny
 from mypy.subtypes import is_subtype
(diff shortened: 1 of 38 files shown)


ilevkivskyi:
> Could you try adding `Final` to `_dummy` in `types.py` also? I think that could give a solid win

Oh yes, thanks, I completely forgot about this one.

Also everyone please suggest anything here if you see some of your favorite class and module final names are missing.

ilevkivskyi:
> I think that could give a solid win

Oh yes, another 50ms, which is amazing for a single constant.

JukkaL:
What about `mypy.nodes.op_methods` (and the various things that immediately follow)?

msullivan:
> Oh yes, another 50ms, which is amazing for a single constant.

In my profiling, the hottest compiled function was `CallableType.copy_modified`, which references `_dummy` *32* times.

PR #5629, opened 2018-09-17
Get rid of Optional when assigning Any in an `is None` branch

### LINKED ISSUE #3526: Assignment of Any does not get rid of Optional (add `AnyUnion` type?)

msullivan:
As a special case, when assigning Any to a variable with a
declared Optional type that has been narrowed to None,
replace all the Nones in the declared Union type with Any.

The result of this is:
```
def f():...
def g(x: Optional[int]) -> int:
    if x is None:
        x = f()
        reveal_type(x)   # Union[int, Any]  # <---- this is new (was Optional[int])
    reveal_type(x)  # Union[int, Any]  # was also Optional[int]
    return x  # Ok  # was an error
```

Fixes #3526.

### DIFF
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -292,7 +292,7 @@ class ConditionalTypeBinder:
             self.put(expr, UnionType(new_items))
         elif (isinstance(type, AnyType)
               and not (isinstance(declared_type, UnionType)
-                       and (any(isinstance(item, AnyType) for item in declared_type.items)))):
+                       and any(isinstance(item, AnyType) for item in declared_type.items))):
             # Assigning an Any value doesn't affect the type to avoid false negatives, unless
             # there is an Any item in a declared union type.
             self.put(expr, declared_type)
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -6,7 +6,7 @@ MYPY = False
 if MYPY:
     from typing import DefaultDict
 
-from mypy.types import Type, AnyType, PartialType, UnionType, TypeOfAny
+from mypy.types import Type, AnyType, PartialType, UnionType, TypeOfAny, NoneTyp
 from mypy.subtypes import is_subtype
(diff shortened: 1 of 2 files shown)


JukkaL:
Great to have this fixed! This has been a pain point for a long time.

PR #5634, opened 2018-09-18
Refactor semantic analysis of assignment statements

JukkaL:
Split a monster function into several smaller functions. Merge some almost
repeated code into a helper method.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1982,89 +1982,7 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
                 overrides an existing name.
         """
         if isinstance(lval, NameExpr):
-            # Top-level definitions within some statements (at least while) are
-            # not handled in the first pass, so they have to be added now.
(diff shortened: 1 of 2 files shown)


mypy/semanal.py
```diff
@@ -2089,6 +2007,105 @@ def analyze_lvalue(self, lval: Lvalue, nested: bool = False,
...
+            lvalue.fullname = v._fullname
+        else:
+            lvalue.fullname = lvalue.name
+        return v
+
+    def make_name_lvalue_point_to_existing_def(
+            self,
+            lval: NameExpr,
+            explicit_type: bool,
+            final_cb: Optional[Callable[[bool], None]]) -> None:
```
ilevkivskyi: I know this is really minor, but I would add a docstring here (jus point to `analyze_lvalue` for args, like you did above).

ilevkivskyi: OK, this is not essential. I am going to merge this just to reduce number of open PRs

PR #5635, opened 2018-09-18
Update imported Vars in the third pass

### LINKED ISSUE #4442: Named tuples in import cycles still fail sometimes
It looks like the issue #3054 was not completely fixed. Namely, in some scenarios named tuples
in import cycles result in ``Invalid type``. For example, checking this file
```python
import posix
```
results in
```
typeshed/stdlib/3/os/__init__.pyi:331: error: Invalid type "posix.uname_result"
typeshed/stdlib/3/os/__init__.pyi:599: error: Invalid type "posix.times_result"
typeshed/stdlib/3/os/__init__.pyi:605: error: Invalid type "posix.waitid_result"
typeshed/stdlib/3/os/__init__.pyi:622: error: Invalid type "posix.sched_param"
typeshed/stdlib/3/os/__init__.pyi:624: error: Invalid type "posix.sched_param"
typeshed/stdlib/3/os/__init__.pyi:625: error: Invalid type "posix.sched_param"
```
Note that the above "invalid" types are all named tuples and there is an import cycle ``os <-> posix``. Interestingly, ``import os`` and ``from os import *`` both typecheck cleanly.

I haven't yet tried to find a simple repro (not involving typrshed).

### LINKED ISSUE #4498: "Invalid type" when using type alias in import cycle

### LINKED ISSUE #5275: Import cycle with type aliasing causes invalid "Invalid type"
Running stub checks on master yielded the following:

```
[ethanhs@towerofart mypy]$ pytest -k testsamples
============================= test session starts ==============================
platform linux -- Python 3.6.5, pytest-3.5.1, py-1.5.3, pluggy-0.6.0
rootdir: /home/ethanhs/mypy, inifile: pytest.ini
plugins: xdist-1.22.2, forked-0.2, cov-2.5.1
gw0 [2] / gw1 [2] / gw2 [2] / gw3 [2] / gw4 [2] / gw5 [2] / gw6 [2] / gw7 [2]
scheduling tests via LoadScheduling
.F                                                                       [100%]
=================================== FAILURES ===================================
___________________________ SamplesSuite.test_stubs ____________________________
[gw2] linux -- Python 3.6.5 /usr/bin/python
Sample check failed
----------------------------- Captured stdout call -----------------------------
typeshed/third_party/2and3/click/core.pyi:337: error: Invalid type "click.types._ConvertibleType"
typeshed/third_party/2and3/click/core.pyi:416: error: Invalid type "click.types._ConvertibleType"
===================== 1 failed, 1 passed in 13.15 seconds ======================
```

Ivan pointed out there is an import cycle, and it seems the cycle and type aliasing is giving mypy trouble.

ilevkivskyi:
Fixes #5275 
Fixes #4498 
Fixes #4442 

This is a simple band-aid fix for `Invalid type` in import cycles where type aliases, named tuples, or typed dicts appear.

Ideally we could refactor semantic analysis to have deferred nodes, but this is _much_ harder.

### DIFF
--- a/mypy/semanal_pass3.py
+++ b/mypy/semanal_pass3.py
@@ -67,11 +67,24 @@ class SemanticAnalyzerPass3(TraverserVisitor, SemanticAnalyzerCoreInterface):
         self.sem.globals = file_node.names
         with experiments.strict_optional_set(options.strict_optional):
             self.scope.enter_file(file_node.fullname())
+            self.update_imported_vars()
             self.accept(file_node)
             self.analyze_symbol_table(file_node.names)
             self.scope.leave()
         del self.cur_mod_node
         self.patches = []
+
+    def update_imported_vars(self) -> None:
+        for sym in self.cur_mod_node.names.values():
+            if sym and isinstance(sym.node, Var):
+                var = sym.node
+                mod_name, _, name = var.fullname().rpartition('.')
+                if mod_name != self.sem.cur_mod_id:  # imported
+                    new_sym = self.sem.modules[mod_name].names.get(name)
+                    if isinstance(new_sym.node, (TypeInfo, TypeAlias)):
+                        # This Var was replaced with a class (like named tuple)
+                        # or alias, update this.
+                        sym.node = new_sym.node
 
     def refresh_partial(self, node: Union[MypyFile, FuncItem, OverloadedFuncDef],
                         patches: List[Tuple[int, Callable[[], None]]]) -> None:
--- a/mypy/semanal_pass3.py
+++ b/mypy/semanal_pass3.py
@@ -85,7 +85,7 @@ class SemanticAnalyzerPass3(TraverserVisitor, SemanticAnalyzerCoreInterface):
                     continue
                 if mod_name != self.sem.cur_mod_id:  # imported
                     new_sym = self.sem.modules[mod_name].names.get(name)
-                    if isinstance(new_sym.node, (TypeInfo, TypeAlias)):
+                    if new_sym and isinstance(new_sym.node, (TypeInfo, TypeAlias)):
                         # This Var was replaced with a class (like named tuple)
                         # or alias, update this.
                         sym.node = new_sym.node
--- a/mypy/semanal_pass3.py
+++ b/mypy/semanal_pass3.py
@@ -77,8 +77,12 @@ class SemanticAnalyzerPass3(TraverserVisitor, SemanticAnalyzerCoreInterface):
     def update_imported_vars(self) -> None:
         for sym in self.cur_mod_node.names.values():
             if sym and isinstance(sym.node, Var):
-                var = sym.node
-                mod_name, _, name = var.fullname().rpartition('.')
+                fullname = sym.node.fullname()
+                if '.' not in fullname:
+                    continue
+                mod_name, _, name = fullname.rpartition('.')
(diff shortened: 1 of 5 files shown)


test-data/unit/check-namedtuple.test
```diff
@@ -691,6 +691,23 @@ my_eval(A([B(1), B(2)])) # OK
...
 
+[case testNamedTupleImportCycle]
+import b
+[file a.py]
+class C:
+    pass
+
+from b import tp
+x: tp
+reveal_type(x.x)  # E: Revealed type is 'builtins.int'
```
JukkaL: What if you have `tp('x')` in this file? Or `reveal_type(tp)`?

ilevkivskyi: Yes, this will not work. I was thinking about the `Invalid type` problem. All references in type context will be correctly re-resolved in the third pass. While there is no such thing as forward reference in runtime context.
I however still think this fix may have some value:
* We have seen many complains about `Invalid type`, but few complains about `Cannot determine type` (the error that appears if you try `tp('x')`)
* The reason for above fact is that many of the import cycle exist _only_ for annotations, `tp('x')` will fail at runtime  anyway
* This error quite often appears in stub files for large libraries, where we only care about type context

It is up to you of course.

ilevkivskyi:
@JukkaL As you requested elsewhere I added a detailed doctring, tests for still not fixed behavior, and updated the PR description.

mypy/semanal_pass3.py
```diff
@@ -67,12 +67,55 @@ def visit_file(self, file_node: MypyFile, fnam: str, options: Options,
...
+    def update_imported_vars(self) -> None:
+        """Update nodes for imported names, if they got updated from Var to TypeInfo or TypeAlias.
+
+        This is a simple _band-aid_ fix for "Invalid type" error in import cycles where type
+        aliases, named tuples, or typed dicts appear. The root cause is that during first pass
+        definitions like:
+
+            A = List[int]
+
+        Are seen by mypy as variables, because it doesn't know yet that `List` refers to a type.
```
JukkaL: Grammar nit: use lower case 'a' in 'are'.

mypy/semanal_pass3.py
```diff
@@ -67,12 +67,55 @@ def visit_file(self, file_node: MypyFile, fnam: str, options: Options,
...
+        In the second pass, such `Var` is replaced with a `TypeAlias`. But in import cycle,
+        import of `A` will still refer to the old `Var` node. Therefore we need to update it.
+
+        Note that this a partial fix that only fixes the "Invalid type" error when a type alias
```
JukkaL: 'this a' -> 'this is a'

JukkaL:
It would also be good to create a follow-up issue for the remaining problems, if one doesn't exist yet.

PR #5638, opened 2018-09-18
Don't duplicate functions for typevar value restrictions under mypyc

msullivan:
mypyc wants one canonical version of a function that is typechecked,
so make that happen.

Adds a per-module mypyc option that should be set by mypyc on modules
that are being compiled.

This allows fixing a mypyc issue that is breaking self-compilation right now.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1243,7 +1243,10 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             if tvar.values:
                 subst.append([(tvar.id, value)
                               for value in tvar.values])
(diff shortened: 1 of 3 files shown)


msullivan:
Making them unions seems like it could lead to false positives, though, since you don't know that two variables are the same

ilevkivskyi:
> Making them unions seems like it could lead to false positives, though, since you don't know that two variables are the same

I mean not for actual checking but for `mypyc` type map. We can get more precise types than `PyObject *` resulting from the `Any`s, or it doesn't really make sense?

Issue #5640, opened 2018-09-19
Mypy infers types of NamedTuple subclass as 'nothing' when using isinstance.


toshihikoyanase:
I encountered an issue that `mypy` failed to infer a type of a variable. This issue occurs when `isinstance()` is applied to a subclass of `NamedTuple`.

Environment:
- Python 3.6.5
- mypy: 0.630

Code
```python
from typing import NamedTuple


class Base:
    pass


class BaseTuple(NamedTuple):
    value: float


class MyTuple(BaseTuple, Base):
    pass


def f(o: Base) -> float:
    if isinstance(o, MyTuple):
        return o.value + 10
    return 0


t = MyTuple(1)
print(f(t))
```

Python 3.6.5 successfully executes this code and returns `11` as expected. But `mypy` 0.630 fails to infer the type of `o` after checking its type using `isinstance` as follows:

```
sample.py:18: error: <nothing> has no attribute "value"
```

Observations:
- Current master head (commit 5751f42ff79265deb4126075b2511b823d006ae0) also has this issue.
- `mypy` 0.620 successfully infer the type and does not show this error message.
- This issue seems to occur from https://github.com/python/mypy/commit/3fb16a2b8c5439074e39b75feebc109a439eede3.
- `mypy` 0.630 successfully infer the type if `BaseTuple` is not a subclass of NamedTuple as follows:
```python
class BaseTuple:

    def __init__(self, value: float) -> None:
        self.value = value
```

ilevkivskyi:
Thanks for reporting! We are sorry for this regression. I will try fixing this shortly.

toshihikoyanase:
This fix solved actual problems in my project. Thank you for your swift action!

PR #5641, opened 2018-09-19
Fix meet for tuple with instance

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5640

Since generic tuple types are not supported, `map_instance_to_supertype` is not needed here. This solution doesn't work for general multiple inheritance, for example the test case will not work with `if isinstance(o, BaseTuple)`, but this is exactly the same as for e.g. two instance types.

### DIFF
--- a/mypy/meet.py
+++ b/mypy/meet.py
@@ -490,16 +490,18 @@ class TypeMeetVisitor(TypeVisitor[Type]):
             # TODO: What if the fallbacks are different?
             return TupleType(items, t.fallback)
         # meet(Tuple[t1, t2, <...>], Tuple[s, ...]) == Tuple[meet(t1, s), meet(t2, s), <...>].
-        elif (isinstance(self.s, Instance) and
-              (self.s.type.fullname() == 'builtins.tuple' or is_proper_subtype(t, self.s))
-              and self.s.args):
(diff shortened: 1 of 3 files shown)


mypy/meet.py
```diff
@@ -489,17 +489,18 @@ def visit_tuple_type(self, t: TupleType) -> Type:
...
-              and self.s.args):
-            return t.copy_modified(items=[meet_types(it, self.s.args[0]) for it in t.items])
+        elif isinstance(self.s, Instance):
+            # meet(Tuple[t1, t2, <...>], Tuple[s, ...]) == Tuple[meet(t1, s), meet(t2, s), <...>].
+            if self.s.type.fullname() == 'builtins.tuple' and self.s.args:
+                return t.copy_modified(items=[meet_types(it, self.s.args[0]) for it in t.items])
+            elif is_proper_subtype(t, self.s):
+                # A named tuple that inherits from a normal class
+                return t
         elif (isinstance(self.s, Instance) and t.fallback.type == self.s.type):
```
JukkaL: It looks like the condition can never be true. Either move this to the above elif block or remove?

ilevkivskyi: Yes, good catch, I think this is not needed anymore.

ilevkivskyi: OK, fixed now!

PR #5643, opened 2018-09-19
 Make NamedTuple provide __new__ instead of __init__

### LINKED ISSUE #1279: Two issues with subclassing namedtuple-generated class and overriding __new__
Python 3.5.1, mypy 2f89c79.

Test code:

```
from typing import NamedTuple

Base = NamedTuple('Base', [('param', int)])


class Child(Base):

    def __new__(cls, param: int = 1) -> 'Child':
        return Base.__new__(cls, param)


print(Child())
```

Python result:

```
% python3 test2.py 
Child(param=1)
```

mypy output:

```
test2.py: note: In member "__new__" of class "Child":
test2.py:9: error: Too many arguments for "__new__" of "object"
test2.py: note: At top level:
test2.py:12: error: Too few arguments for "Child"
```

I poked around in the code a bit and I believe I managed to narrow this down to the following difference:
- The code above depends on the namedtuple-generated class to have `__new__` defined (CPython: https://github.com/python/cpython/blob/9967d6c1053f186f3b3d118ac336a03752efec32/Lib/collections/__init__.py#L316)
- mypy treats namedtuple definitions as if they were overriding `__init__` instead (https://github.com/python/mypy/blob/2f89c79a5b030343879f2f3c7172f0ad2b00c4de/mypy/semanal.py#L1444)

msullivan:
Closes #1279.

Depends on either #5642 or a typeshed sync

### DIFF
--- a/test-data/unit/check-class-namedtuple.test
+++ b/test-data/unit/check-class-namedtuple.test
@@ -587,7 +587,7 @@ class XMethBad(NamedTuple):
 class MagicalFields(NamedTuple):
     x: int
     def __slots__(self) -> None: pass  # E: Cannot overwrite NamedTuple attribute "__slots__"
-    def __new__(cls) -> None: pass  # E: Cannot overwrite NamedTuple attribute "__new__"
+    def __new__(cls) -> None: pass  # E: Name '__new__' already defined (possibly by an import)
     def _source(self) -> int: pass  # E: Cannot overwrite NamedTuple attribute "_source"
     __annotations__ = {'x': float}  # E: NamedTuple field name cannot start with an underscore: __annotations__ \
         # E: Invalid statement in NamedTuple definition; expected "field_name: field_type [= default]" \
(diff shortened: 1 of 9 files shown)


test-data/unit/check-class-namedtuple.test
```diff
@@ -587,7 +587,7 @@ class XMethBad(NamedTuple):
 class MagicalFields(NamedTuple):
     x: int
     def __slots__(self) -> None: pass  # E: Cannot overwrite NamedTuple attribute "__slots__"
-    def __new__(cls) -> None: pass  # E: Cannot overwrite NamedTuple attribute "__new__"
+    def __new__(cls) -> None: pass  # E: Name '__new__' already defined on line 8
```
ilevkivskyi: This error message is not perfect but probably OK. Why don't you want to keep the old error message?

msullivan: I'd prefer the old error message TBH but the interaction between the namedtuple override checks and the method actually existing seemed nontrivial and I didn't think it was worth fighting with.

ilevkivskyi: If it is hard, then I think it is OK.

msullivan: I looked into it a bit and it turns out to be very easy to fix as part of another PR I am putting together. I'm going to land this now with the error message regression (to avoid needing to stack this on top of the other PR, which might require more back-and-forth) and then fix the error message in that PR.

Issue #5195, opened 2018-06-11
Recursive subclass of NamedTuple crashes mypy


toolforger:
Mypy 0.610, Python 3.6.3.
Not tried Mypy from git master (haven't checked how to do that).

Crash vanishes if I replace the recursive type with a nonrecursive one, or use `object` as a parent class.

Code:

```Python
#! /usr/bin/env python3

from typing import List, NamedTuple

class Command(NamedTuple):
    subcommands: List['Command']

class HelpCommand(Command):
    pass

HelpCommand(subcommands=[]) # crashes mypy, works in Python
```

Command: `python3 -m mypy typetest.py --show-traceback`

Output:
```
typetest.py:5: error: Recursive types not fully supported yet, nested types replaced with "Any"
typetest.py: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.610
Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/jo/.local/lib/python3.6/site-packages/mypy/__main__.py", line 11, in <module>
    main(None)
  File "/home/jo/.local/lib/python3.6/site-packages/mypy/main.py", line 91, in main
    res = type_check_only(sources, bin_dir, options, flush_errors, fscache)  # noqa
  File "/home/jo/.local/lib/python3.6/site-packages/mypy/main.py", line 148, in type_check_only
    fscache=fscache)
  File "/home/jo/.local/lib/python3.6/site-packages/mypy/build.py", line 185, in build
    flush_errors, fscache)
  File "/home/jo/.local/lib/python3.6/site-packages/mypy/build.py", line 285, in _build
    graph = dispatch(sources, manager)
  File "/home/jo/.local/lib/python3.6/site-packages/mypy/build.py", line 2398, in dispatch
    process_graph(graph, manager)
  File "/home/jo/.local/lib/python3.6/site-packages/mypy/build.py", line 2691, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/home/jo/.local/lib/python3.6/site-packages/mypy/build.py", line 2820, in process_stale_scc
    graph[id].finish_passes()
  File "/home/jo/.local/lib/python3.6/site-packages/mypy/build.py", line 2067, in finish_passes
[...]

PR #5644, opened 2018-09-19
Fully analyze named tuple subclasses in third pass

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5195

The fix is straightforward, previously `tuple_type` was not analyzed for named tuple subclasses.
I also copy the line number for instance types when transforming them.


ilevkivskyi:
> Is there a similar transformation that ought to be done for `typeddict_type` as well?

I think no. A subclass of a typed dict is also a typed dict (i.e. it becomes itself a "magic" class), while subclassing a named tuple creates a more "normal" class. But I think it makes sense to add a test just for completeness.

PR #5646, opened 2018-09-19
Make classmethod's first argument be Type[...]

### LINKED ISSUE #1727: Constructor defined after classmethod is considered to have no args in the classmethod

### LINKED ISSUE #4190: False positive when calling object.__new__ inside __new__ with args
Consider:

```python
class A:
    def __new__(cls, x: int) -> 'A':
        return object.__new__(cls) # Argument 1 to "__new__" of "object" has incompatible type "Type[A]"; expected "Type[object]"

class B:
    def __new__(cls) -> 'B':
        return object.__new__(cls)
```

The `object.__new__(cls)` call type checks when inside a `__new__` method with a single argument, but not when there are extra arguments.

The code for `A` is taken from `testSuperWithNew`, but because `object.__new__` is unannotated in `fixtures/__new__.pyi`, the type error doesn't show up in the unit test. The discrepancy between the fixture and typeshed was noted in #3111.

### LINKED ISSUE #5263: Incorrect error in factory function on a dataclass
In Python `3.7.0rc1` and mypy `0.620+dev-5dc7454b1acb47dcbb3361c849e6113efcdbaaf2`, if I save the following code to `dc_test.py`

```python
from typing import Type, TypeVar

T = TypeVar('T', bound='A')

@dataclass
class A:
	@classmethod
	def make(cls: Type[T]) -> T:
		return cls()
```

I get the following mypy output

```
$ mypy dc_test.py 
dc_test.py:10: error: Incompatible return value type (got "A", expected "T")
```

If I comment out the `@dataclass` line, I get no error. I would expect to get no error with or without the `@dataclass` line.

msullivan:
Currently the first argument to `__new__` and classmethods is a
callable type that is constructed during semantic analysis by
typechecker code (!) that looks for the `__init__`/`__new__` methods.

This causes a number of problems, including not being able to call
`object.__new__` in a subclass's `__new__` if it took arguments
(#4190) and giving the wrong type if `__init__` appeared after the
class method (#1727).

Taking a `Type` instead lets us solve those problems, and postpone
computing the callable version of the type until typechecking if it is
needed.

This also lets us drop a bunch of plugin code that tries to fix up the
types of its cls arguments post-hoc, sometimes incorrectly (#5263).

Fixes #1727.
Fixes #4190.
Fixes #5263.

### DIFF
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -349,7 +349,11 @@ class A:
 T = TypeVar('T')
 class B(Generic[T]):
     def __new__(cls, foo: T) -> 'B[T]':
-        return object.__new__(cls)
+        x = object.__new__(cls)
+        # object.__new__ doesn't have a great type :(
+        reveal_type(x)  # E: Revealed type is 'Any'
+        return x
+
 [builtins fixtures/__new__.pyi]
 
 [case testInnerFunctionNotOverriding]
@@ -5131,14 +5135,42 @@ class C:
 [builtins fixtures/property.pyi]
 [out]
 
-[case testClassMethodBeforeInit]
-class Foo(object):
+[case testClassMethodBeforeInit1]
+class Foo:
     @classmethod
-    def bar(cls):
-        # type: () -> Foo
+    def bar(cls) -> Foo:
         return cls("bar")
 
-    def __init__(self, baz):
-        # type: (str) -> None
+    def __init__(self, baz: str) -> None:
         self.baz = baz
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodBeforeInit2]
+class Foo:
+    @classmethod
+    def bar(cls) -> Foo:
+        return cls(Bar())
+
+    def __init__(self, baz: 'Bar') -> None:
+        self.baz = baz
+
+class Bar: pass
+[builtins fixtures/classmethod.pyi]
+
+[case testClassMethodBeforeInit3]
+from typing import overload
+class Foo:
+    @classmethod
+    @overload
+    def bar(cls, x: int) -> Foo: ...
+    @classmethod
+    @overload
+    def bar(cls, x: str) -> Foo: ...
+    @classmethod
+    def bar(cls, x: object) -> Foo:
+        return cls(x)
+
+    def __init__(self, baz: object) -> None:
+        self.baz = baz
+
 [builtins fixtures/classmethod.pyi]
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -339,6 +339,19 @@ class B(A):
     def __new__(cls) -> int:
         return 1
 
+[case testOverride__new__AndCallObject]
+from typing import TypeVar, Generic
+
+class A:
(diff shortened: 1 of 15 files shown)


test-data/unit/check-classes.test
```diff
@@ -339,6 +339,19 @@ class B(A):
...
+from typing import TypeVar, Generic
+
+class A:
+    def __new__(cls, x: int) -> 'A':
+        return object.__new__(cls)
+
+T = TypeVar('T')
+class B(Generic[T]):
+    def __new__(cls, foo: T) -> 'B[T]':
+        return object.__new__(cls)
```
JukkaL: Maybe reveal the type of the return value?

test-data/unit/check-dataclasses.test
```diff
@@ -421,6 +421,23 @@ s: str = a.bar()  # E: Incompatible types in assignment (expression has type "in
...
+
+T = TypeVar('T')
+
+@dataclass
+class A(Generic[T]):
+  x: T
+
+  @classmethod
+  def foo(cls) -> None:
+      reveal_type(cls)            # E: Revealed type is 'Type[__main__.A[T`1]]'
```
JukkaL: Maybe try calling `cls` and reveal the type of the result?

test-data/unit/check-dataclasses.test
```diff
@@ -466,3 +483,16 @@ app.rating
...
+from typing import Type, TypeVar
+from dataclasses import dataclass
+
+T = TypeVar('T', bound='A')
+
+@dataclass
+class A:
+    @classmethod
+    def make(cls: Type[T]) -> T:
+        return cls()
```
JukkaL: Maybe reveal the type of `cls()`?

msullivan:
It causes just a little trouble, which I will fix up. Mostly it creates some redundant cast errors.

test-data/unit/check-classes.test
```diff
@@ -5118,6 +5131,18 @@ class C:
...
+class Foo(object):
+    @classmethod
+    def bar(cls):
+        # type: () -> Foo
+        return cls("bar")
+
+    def __init__(self, baz):
+        # type: (str) -> None
+        self.baz = baz
+[builtins fixtures/classmethod.pyi]
```
ilevkivskyi: I would add two more test cases like this:
* One where `__init__` has a forward reference in its signature
* Another with an overloaded classmethod

ilevkivskyi:
@msullivan This looks very good, and simplifies the code. I just have two suggestions for more tests.

test-data/unit/check-classes.test
msullivan: Sure, though it is `Any`, since that's what the stubs have for `object.__new__`. (Which is probably fixable in typeshed/fixtures?)

PR #5654, opened 2018-09-20
Do pass 1 semanatic analysis on the FuncDef of decorated functions

### LINKED ISSUE #5652: Crash when using decorated function as type annotation in import cycle
Discovered from a user report at Dropbox, the following test case crashes:
```
[case testCheckDecoratedFuncAsAnnotWithImportCycle]
import a
[file a.py]
from typing import TypeVar
import b

T = TypeVar('T')
def idf(x: T) -> T: return x

@idf
def Session() -> None: pass

[file b.py]
MYPY = False
if MYPY:
    from a import Session

def f(self, session: Session) -> None:  # E: Invalid type "a.Session"
    pass
[builtins fixtures/bool.pyi]
```

with a traceback ending with:
```
  File "/home/msullivan/src/mypy/mypy/semanal.py", line 442, in _visit_func_def
    defn.type = analyzer.visit_callable_type(defn.type, nested=False)
  File "/home/msullivan/src/mypy/mypy/typeanal.py", line 421, in visit_callable_type
    ret = t.copy_modified(arg_types=self.anal_array(t.arg_types, nested=nested),
  File "/home/msullivan/src/mypy/mypy/typeanal.py", line 639, in anal_array
    res.append(self.anal_type(t, nested))
  File "/home/msullivan/src/mypy/mypy/typeanal.py", line 646, in anal_type
    return t.accept(self)
  File "/home/msullivan/src/mypy/mypy/types.py", line 242, in accept
    return visitor.visit_unbound_type(self)
  File "/home/msullivan/src/mypy/mypy/typeanal.py", line 182, in visit_unbound_type
    typ = self.visit_unbound_type_nonoptional(t)
  File "/home/msullivan/src/mypy/mypy/typeanal.py", line 218, in visit_unbound_type_nonoptional
    tvar_def = self.tvar_scope.get_binding(sym)
  File "/home/msullivan/src/mypy/mypy/tvar_scope.py", line 78, in get_binding
```

msullivan:
This populates fullname properly, fixing a crash where a decorated
function is used as a type annotation in an import cycle,
as well as fixing an obscure bug in sys.platform checks inside
decorated functions.

Fixes #5652.

### DIFF
--- a/mypy/semanal_pass1.py
+++ b/mypy/semanal_pass1.py
@@ -148,6 +148,12 @@ class SemanticAnalyzerPass1(NodeVisitor[None]):
                 self.analyze_lvalue(lval, explicit_type=s.type is not None)
 
     def visit_func_def(self, func: FuncDef, decorated: bool = False) -> None:
+        """Process a func def.
+
+        decorated is true if we are processing a func def in a
+        Decorator that needs a _fullname and to have its body analyzed but
+        does not need to be added to the symbol table.
+        """
         sem = self.sem
         if sem.type is not None:
             # Don't process methods during pass 1.
--- a/mypy/semanal_pass1.py
+++ b/mypy/semanal_pass1.py
@@ -147,14 +147,14 @@ class SemanticAnalyzerPass1(NodeVisitor[None]):
             for lval in s.lvalues:
                 self.analyze_lvalue(lval, explicit_type=s.type is not None)
 
-    def visit_func_def(self, func: FuncDef) -> None:
+    def visit_func_def(self, func: FuncDef, decorated: bool = False) -> None:
         sem = self.sem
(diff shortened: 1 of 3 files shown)


mypy/semanal_pass1.py
```diff
@@ -147,14 +147,14 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
             for lval in s.lvalues:
                 self.analyze_lvalue(lval, explicit_type=s.type is not None)
 
-    def visit_func_def(self, func: FuncDef) -> None:
+    def visit_func_def(self, func: FuncDef, decorated: bool = False) -> None:
```
ilevkivskyi: I would document the argument and the need for it, because it is not obvious.

Issue #5659, opened 2018-09-21
mypy ignores type errors inside `list` and `dict` calls


msullivan:
In the following program:
```
from typing import Union, Iterable, Tuple

class A:
    def foo(self) -> Iterable[Tuple[int, int]]: pass

def bar(x: int) -> Union[A, int]: ...

list(bar('lol').foo())  # No errors!
dict(bar('lol').foo())  # No errors!

tuple(bar('lol').foo())  # Does error
set(bar('lol').foo())  # Does error
```
two errors ought to be generated for each call (one for `int` not having `.foo`, one for `'lol'` being the wrong type of argument). These errors seem to be suppressed while checking `list` and `dict`, which get filled with `Any`s.

msullivan:
This looks like an overloads issue. The following doesn't produce any errors:
```
from typing import Union, Iterable, Tuple, TypeVar, Generic, overload, Any

class A:
    def foo(self) -> Iterable[Tuple[int, int]]: pass

def bar(x: int) -> Union[A, int]: ...

_T = TypeVar('_T')

@overload
def foo() -> None: ...
@overload
def foo(iterable: Iterable[_T]) -> None: ...

def foo(iterable: Any = None) -> None: pass

foo(bar('lol').foo())  # No errors!
```
So this looks like an overloads issue, and it is a regression from 0.610 to 0.620 (I haven't dug any deeper than that.)

@Michael0x2a thoughts?

Issue #4872, opened 2018-04-09
Mypy should be careful when using outer context for inference


ilevkivskyi:
There are several scenarios, when mypy inference leads to false positives. Namely, if the outer type context (return context) is too wide, mypy uses it and fails immediately. Instead it should try the inner context (argument context) before giving up. For example:
```python
T = TypeVar('T', bound=Dict[str, Any])

def f(arg: T) -> T:
    ...
def outer(x: Mapping[str, Any]) -> None:
    ...
d: Dict[str, Any]

m: Mapping[str, Any] = f(d)  # Value of type variable "T" of "f" cannot be "Mapping[str, Any]"
outer(f(d))  # Value of type variable "T" of "f" cannot be "Mapping[str, Any]"
```
There are similar scenarios with union types, for example:
```python
S = TypeVar('S', bound=int)

def g(arg: S) -> Optional[S]:
    ...

x: Optional[int]
x = g(42)  # Value of type variable "S" of "g" cannot be "Optional[int]"
```
etc. What is interesting, the failure _doesn't_ happen if there are only simple types present in the bound and in the context, to reproduce one needs either generics, union types, or other "complex" types.

ilevkivskyi:
I set priority to high because I encountered this twice recently. If there is no simple solution we can downgrade to normal.

JukkaL:
A fix might be easy enough -- if the type inferred from the outer context is not compatible with type variable bounds, we leave the value of the type variable indeterminate (similar to what happens if there's no outer context) and continue type checking.

Another idea would be to remove union items not compatible with the bound from the inferred type variable value, and to map generic types not compatible with the band down to the bound, if possible. This idea feels a bit ad hoc though.

I don't remember ever encountering this issue though.

ilevkivskyi:
#5049 Shows the same problem with unbound type variables which may be harder to fix, because in that case `Optional[int]` is a valid inferred type (because `Optional[Optional[int]] is Optional[int]`), just not the most precise.

jcrada:
Hi,

I  believe I have the same problem. Please find below another example to replicate.

```python
import unittest
from typing import Optional, Type, TypeVar


class A:
    def __init__(self) -> None:
        print("A: Constructing...")


class B:
    def __init__(self) -> None:
        print("B: Constructing...")

    def do_something(self) -> None:
        print("B: Doing something")


class Experimental(object):
    T = TypeVar('T', 'A', 'B')

    def create_component_a(self) -> Optional['A']:
[...]

PR #5660, opened 2018-09-22
Fix bug with inferring bad arguments to overloads

Michael0x2a:
Currently, it seems that when checking overloads, mypy starts by inferring the argument types in an empty context and swallows any errors generated in the process.

Mostly likely what the old overload implementation did next was recheck the arguments against the selected overload variant, regenerating any errors as necessary.

However, in the new overload implementation, it's unclear if we consistently do the same thing, mostly due to how we slice and dice the arguments to do things like union math.

This pull request does the low-effort thing and just removes the error suppression.

This pull request also does some unrelated clean-up -- apparently, the `callee` argument to 'infer_arg_types_in_context` is always 'None'.

Fixes https://github.com/python/mypy/issues/5659.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -627,7 +627,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 callee = self.infer_function_type_arguments(
                     callee, args, arg_kinds, formal_to_actual, context)
 
-            arg_types = self.infer_arg_types_in_context2(
+            arg_types = self.infer_arg_types_in_context(
                 callee, args, arg_kinds, formal_to_actual)
 
             self.check_argument_count(callee, arg_types, arg_kinds,
@@ -655,13 +655,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 callee = callee.copy_modified(ret_type=ret_type)
             return callee.ret_type, callee
         elif isinstance(callee, Overloaded):
-            # Type check arguments in empty context. They will be checked again
-            # later in a context derived from the signature; these types are
-            # only used to pick a signature variant.
-            self.msg.disable_errors()
-            arg_types = self.infer_arg_types_in_context(None, args)
-            self.msg.enable_errors()
-
+            arg_types = self.infer_arg_types_in_empty_context(args)
             return self.check_overload_call(callee=callee,
                                             args=args,
                                             arg_types=arg_types,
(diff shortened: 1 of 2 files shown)


Michael0x2a:
Note: I haven't tested this in our internal repos yet.

mypy/checkexpr.py
```diff
@@ -655,13 +655,7 @@ def check_call(self, callee: Type, args: List[Expression],
...
             return callee.ret_type, callee
         elif isinstance(callee, Overloaded):
-            # Type check arguments in empty context. They will be checked again
-            # later in a context derived from the signature; these types are
-            # only used to pick a signature variant.
-            self.msg.disable_errors()
-            arg_types = self.infer_arg_types_in_context(None, args)
-            self.msg.enable_errors()
-
+            arg_types = self.infer_arg_types_in_empty_context(args)
```
ilevkivskyi: At first I was sceptical about this, but after some thinking and playing it seems to me this is actually OK. Because under the hood all relevant possibilities (even the union math) go through `infer_overload_return_type`, that calls `check_call` with a `CallableType` item that, in turn, will trigger `infer_arg_types_in_context` using current matching overload item arg type as context. So it should be fine.

mypy/checkexpr.py
```diff
@@ -744,38 +738,24 @@ def analyze_type_type_callee(self, item: Type, context: Context) -> Type:
...
-        """Infer argument expression types using a callable type as context.
+    def infer_arg_types_in_empty_context(self, args: List[Expression]) -> List[Type]:
+        """Infer argument expression types in an empty context.
 
-        For example, if callee argument 2 has type List[int], infer the
-        argument expression with List[int] type context.
+        In short, we basically recurse on each argument without considering
+        in what context the argument was called.
         """
         # TODO Always called with callee as None, i.e. empty context.
```
ilevkivskyi: This TODO is now outdated.

test-data/unit/check-overloading.test
```diff
@@ -4873,3 +4873,36 @@ def g(name: str) -> int:
...
+
+@overload
+def foo() -> None: ...
+@overload
+def foo(iterable: Iterable[_T]) -> None: ...
+def foo(iterable = None) -> None: pass
+
+foo(bar('lol').foo())  # E: Argument 1 to "bar" has incompatible type "str"; expected "int" \
+                       # E: Item "int" of "Union[A, int]" has no attribute "foo"
+
```
ilevkivskyi: I would add few tests about situations where absence of context can cause troubles (so that this will not regress). For example something like this:
```python
def g(x: Optional[T] = None) -> List[T]: ...

@overload
def f(x: int) -> int: ...
@overload
def f(x: List[int]) -> List[int]: ...
def f(x):
    pass

reveal_type(f(g()))
```
I would also add something like
```python
@overload
def g(x: List[str]) -> List[str]: ...
@overload
def g(x: List[int]) -> List[int]: ...
def g(x):
    pass

@overload
def f(x: int) -> int: ...
@overload
def f(x: List[int]) -> List[int]: ...
def f(x):
    pass

reveal_type(f(g([])))
```
but this one unfortunately causes an error both on master and with this PR. Probably this is something we should fix, but I am afraid this may depend on https://github.com/python/mypy/issues/4872 (but this is just a guess).

Michael0x2a:
@ilevkivskyi (or maybe @msullivan?) -- this is ready for a second look whenever

Issue #5425, opened 2018-08-07
Overriding definition such as type alias in a subclass causes a crash


Michael0x2a:
The following program crashes on both mypy 0.620 and the latest version on master:

```python
from typing import Any, Callable

class Parent:
    foo = Callable[..., int] 

class Child(Parent):
    def foo(self, val: int) -> int:
        return val * 2
```

(In this case, I meant to do `foo: Callable[..., int]`.)

Here's the traceback:

```
test.py:13: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/
issues version: 0.630+dev-1222c9637b673bcca6218a943cd4b88b8749b0ad
Traceback (most recent call last):
  File "C:\Python3.6\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Python3.6\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\Michael\Dropbox\mypy\mypy\__main__.py", line 11, in <module>
    main(None)
  File "C:\Users\Michael\Dropbox\mypy\mypy\main.py", line 91, in main
    res = type_check_only(sources, bin_dir, options, flush_errors, fscache)  # noqa
  File "C:\Users\Michael\Dropbox\mypy\mypy\main.py", line 148, in type_check_only
    fscache=fscache)
  File "C:\Users\Michael\Dropbox\mypy\mypy\build.py", line 177, in build
    flush_errors, fscache)
  File "C:\Users\Michael\Dropbox\mypy\mypy\build.py", line 350, in _build
    graph = dispatch(sources, manager)
  File "C:\Users\Michael\Dropbox\mypy\mypy\build.py", line 2572, in dispatch
    process_graph(graph, manager)
  File "C:\Users\Michael\Dropbox\mypy\mypy\build.py", line 2866, in process_graph
    process_stale_scc(graph, scc, manager)
  File "C:\Users\Michael\Dropbox\mypy\mypy\build.py", line 2989, in process_stale_scc
    graph[id].type_check_first_pass()
  File "C:\Users\Michael\Dropbox\mypy\mypy\build.py", line 2162, in type_check_first_pass
    self.type_checker().check_first_pass()
  File "C:\Users\Michael\Dropbox\mypy\mypy\checker.py", line 253, in check_first_pass
    self.accept(d)
  File "C:\Users\Michael\Dropbox\mypy\mypy\checker.py", line 352, in accept
    stmt.accept(self)
[...]

Issue #2871, opened 2017-02-15
Multiple inheritance: Cannot determine type of "X" in base class "Y" error


alexpirine:
This is probably more of a question than a real issue.

The following code is considered erroneous (tested in Python 3):

```python
class BaseA:
    class Sub:
        pass

class BaseB:
    class Sub:
        pass

class Test(BaseA, BaseB):
    pass
```

mypy responds with:

```
…: error: Cannot determine type of 'Sub' in base class 'BaseA'
…: error: Cannot determine type of 'Sub' in base class 'BaseB'
```

But it doesn't complain if I don't use multiple inheritance:

```python
class BaseA:
    class Sub:
        pass

class BaseB:
    class Sub:
        pass

class Test(BaseA):
    pass
```

gvanrossum:
I think just the error is a bit confusing; in the M.I. case it is trying to
point out that you are inheriting two unrelated class attributes named
'Sub'.

JukkaL:
Based on the error message, it looks likely that the multiple inheritance validity check doesn't expect to see nested classes and gets confused.

alexpirine:
I have the impression that it gets confused, because the following code passes without errors:

```python
class BaseA:
    sub = 1   

class BaseB:
    sub = 1 

class Test(BaseA, BaseB):
    pass
```

I don't really understand what's wrong with nested classes.

gvanrossum:
Well, for one thing, two integers always have the same type. But two independently declared classes (even if they have the same name and contents) are not considered to have the same type.

JukkaL:
Marked as a bug, because the error message is confusing.

alexpirine:
As a side note, I stumbled upon this issue by trying to use two different mixins in Django:

```python
class CreatedDateMixin(models.Model):
    date_created = models.DateTimeField(_("created on"), auto_now_add=True)

    class Meta:
        abstract = True


class UpdatedDateMixin(models.Model):
    date_updated = models.DateTimeField(_("updated on"), auto_now=True)

    class Meta:
        abstract = True


class Foo(CreatedDateMixin, UpdatedDateMixin, models.Model):
    name = models.CharField(_("name"), max_length=255, blank=True)
```

ilevkivskyi:
(Raising priority to high since this is a very common idiom in Django.)

ilevkivskyi:
@mkurnikov you might want to try fixing this.

ilevkivskyi:
[...]

PR #5662, opened 2018-09-23
Fix crash on overriding class level type alias

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5425

The solution is to just prohibit overriding type aliases in subclass. In some sense they are implicitly final, because we need to know them statically (they can be used as type sin annotations).

The assert that was triggered missed two cases for superclass node: `TypeAlias` and `Var`, for `Var` I just return `Any` since there are no way to determine type if deferring failed (currently there is fixed number of type-checking passes).

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1360,16 +1360,22 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             # it can be checked for compatibility.
             original_type = base_attr.type
             original_node = base_attr.node
+            if isinstance(original_node, TypeAlias):
+                # Overriding a type alias is always an error, it is static by nature.
+                self.msg.cannot_overide_alias(original_node.name(), base.name(), defn)
+                return False
             if original_type is None:
                 if self.pass_num < self.last_pass:
-                    # If there are passes left, defer this node until next pass,
-                    # otherwise try reconstructing the method type from available information.
+                    # If there are passes left, defer this node until next pass...
                     self.defer_node(defn, defn.info)
                     return True
+                # ...otherwise try reconstructing the supertype from available information.
                 elif isinstance(original_node, (FuncDef, OverloadedFuncDef)):
                     original_type = self.function_type(original_node)
                 elif isinstance(original_node, Decorator):
                     original_type = self.function_type(original_node.func)
+                elif isinstance(original_node, Var):
+                    original_type = AnyType(TypeOfAny.implementation_artifact)
                 else:
                     assert False, str(base_attr.node)
             if isinstance(original_node, (FuncDef, OverloadedFuncDef)):
@@ -1845,6 +1851,11 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
 
                 base_type, base_node = self.lvalue_type_from_base(lvalue_node, base)
 
+                if isinstance(base_node, TypeAlias):
+                    # Overriding a type alias is always an error, it is static by nature.
+                    self.msg.cannot_overide_alias(base_node.name(), base.name(), lvalue)
+                    return True
+
                 if base_type:
                     assert base_node is not None
                     if not self.check_compatibility_super(lvalue,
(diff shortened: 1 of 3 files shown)


mypy/checker.py
```diff
@@ -1923,6 +1934,10 @@ def lvalue_type_from_base(self, expr_node: Var,
         if base_var:
             base_node = base_var.node
             base_type = base_var.type
+
+            if isinstance(base_node, TypeAlias):
+                return None, base_node
```
JukkaL: Slightly unrelated: Could you update the docstring to describe the return value? Now it's not immediately clear why this is correct.

mypy/checker.py
```diff
@@ -1360,16 +1360,22 @@ def check_method_override_for_base_with_name(
...
+                    # If there are passes left, defer this node until next pass...
                     self.defer_node(defn, defn.info)
                     return True
+                # ...otherwise try reconstructing the supertype from available information.
                 elif isinstance(original_node, (FuncDef, OverloadedFuncDef)):
                     original_type = self.function_type(original_node)
                 elif isinstance(original_node, Decorator):
                     original_type = self.function_type(original_node.func)
+                elif isinstance(original_node, Var):
+                    original_type = AnyType(TypeOfAny.implementation_artifact)
```
JukkaL: It's not clear to me how this is related to the title of the PR, but the description seems to indicate that this fixes another, related issue with `Var` nodes. Maybe tweak the title of the PR? 

Also, this looks like it could cause false negatives. It's better than crashing, but maybe we should generate a "Cannot determine type" error instead? At least there should be a comment about this, and potentially a follow-up issue.

mypy/checker.py
```diff
@@ -1360,16 +1360,22 @@ def check_method_override_for_base_with_name(
             # it can be checked for compatibility.
             original_type = base_attr.type
             original_node = base_attr.node
+            if isinstance(original_node, TypeAlias):
```
JukkaL: Should we reject other kinds of nodes as well, such as `TypeInfo`, `TypeVarExpr` or module reference? Some of these could cause crashes as well. It would be good to fix other related issues at the same time. 

I think that it would be best to check all combinations of base class and subclass definitions, such as 1) type alias 2) class 3) function 4) variable 5) type variable 6) module reference 7) other magic assignments (NewType or NamedTuple or TypedDict). Some of these are already tested, so they can be left out (like function vs function).

mypy/checker.py
```diff
@@ -1845,6 +1851,11 @@ def check_compatibility_all_supers(self, lvalue: RefExpr, lvalue_type: Optional[
 
                 base_type, base_node = self.lvalue_type_from_base(lvalue_node, base)
 
+                if isinstance(base_node, TypeAlias):
```
JukkaL: Similar to above, should we check for other kinds of things that shouldn't be overridden?

Why do we this check in two places? If we need to have the check in two places, it may be slightly better to move it to a helper method instead of repeating the code, at least if the check becomes more complex.

test-data/unit/check-type-aliases.test
```diff
@@ -611,3 +611,33 @@ except E as e:
...
+[out]
+
+[case testNoOverrideAliasInSubclassVar]
+from typing import Any, Callable
+
+class Parent:
+    foo = Callable[..., int]
+
+class Child(Parent):
+    foo: Callable[..., int]  # E: Cannot override type alias "foo" defined in base class "Parent"
```
JukkaL: I'd merge all these test into a single test case. Having many test cases that test almost the same thing arguably doesn't help maintainability but is kind of expensive to run.

ilevkivskyi:
> Should we reject other kinds of nodes as well, such as `TypeInfo`, `TypeVarExpr` or module reference? Some of these could cause crashes as well. It would be good to fix other related issues at the same time.

> I think that it would be best to check all combinations of base class and subclass definitions, such as 1) type alias 2) class 3) function 4) variable 5) type variable 6) module reference 7) other magic assignments (NewType or NamedTuple or TypedDict). Some of these are already tested, so they can be left out (like function vs function).

Yes, indeed, there are much more problems here:
* There are actually _three_ places where we check Liskov: assignments, function definitions, and multiple inheritance, all with slightly different logic that causes some code duplication and is generally fragile.
*  Indeed other (more "exotic") kinds of nodes are not treated carefully in all these places, causing some other issues, like https://github.com/python/mypy/issues/2871

I don't think however I want to open this can of worms, we can look at it later when we will have time.

Issue #4214, opened 2017-11-06
Unexpected error where Generic return type is None 


neiljp:
I introduced this problem on gitter, and am posting this as an issue after confirmation there.

The code below is a much-simplified piece of code to explore if strict-optional could be valid for some recent improved typing of request.pyi in current zulip.
(https://github.com/zulip/zulip/commit/42f5eea61ff0079fef1b2744cddda9a173596535)

The simplified code below produces the error:
`error: "X" does not return a value`

I thought this was just with strict-optional, but the reduced code gives this issue without too, so I think I must have lost the extra code which caused that difference.

```python
from typing import TypeVar, Optional
T = TypeVar('T')
def X(val: T) -> T: ...
x_in = None
def Y(x: Optional[str] = X(x_in)): ...
```

The last line is not required specifically, for example the following also triggers the error in the same way:
```python
xx: Optional[int] = X(x_in)
```

gvanrossum:
So what you're really saying here is that, while we complain about using the return value of a function explicitly declared as `-> None`, we should not complain if a `None` return type was the result of expanding a type variable that happens to be `None`.

I'm not sure I really like that though. Why should generics be special here? Maybe we should just stop complaining about functions with `-> None` too? At least in strict-optional mode -- after all we'll get other complaints if the `None` is unacceptable for the target.

elazarg:
Dropping this check sounds good, except that `Any` lvalues, which will lose an important error.

neiljp:
The idea that generics aren't special seems more consistent. In a more general sense, my surprise perhaps come down to the asymmetry of:
```python
def none() -> None:
    return None
def string() -> str:
    return "5"
works=string() # type: Optional[str]
fails=none()   # type: Optional[str]
```

gvanrossum:
I'm willing to live without the error if the target/context has type `Any`.

ilevkivskyi:
I don't think we should remove this error completely even with ``--strict-optional``. If a user writes something like this:
```python
def f() -> None:
    # some complex function
    ...
# 1000 lines later
x: Optional[str] = f()
```
Then it is probably a mistake. Why would one write this instead of just ``x: Optional[int] = None``? I think a warning like ``f() always returns None`` makes sense for such cases. But I agree that we don't need this for inferred ``None`` i.e. in cases where ``f`` not _always_ returns ``None``.

gvanrossum:
I still fee like mypy is slightly overreaching here. The code is type-safe.
[...]

PR #5663, opened 2018-09-23
Allow returning inferred None from functions

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/4214

As discussed in the issue, assigning an _inferred_ (from generics, overloads, or lambdas) `None` type should be always allowed, while prohibiting assigning function calls that have an explicit `None` return type may have some value (at least people don't complain much about this).

Note: this error is quite nasty for dataclasses, where it flags a legitimate typical use case (see tests).

### DIFF
--- a/test-data/unit/check-expressions.test
+++ b/test-data/unit/check-expressions.test
@@ -937,7 +937,7 @@ main:3: error: "A" not callable
 a, o = None, None # type: (A, object)
 a = f()         # E: "f" does not return a value
 o = a()
-o = A().g(a)    # E: "g" of "A" does not return a value
+o = A().g(a)
 A().g(f())      # E: "f" does not return a value
 x: A = f()      # E: "f" does not return a value
 f()
--- a/test-data/unit/check-expressions.test
+++ b/test-data/unit/check-expressions.test
@@ -936,7 +936,7 @@ main:3: error: "A" not callable
 
 a, o = None, None # type: (A, object)
 a = f()         # E: "f" does not return a value
-o = a()         # E: Function does not return a value
+o = a()
 o = A().g(a)    # E: "g" of "A" does not return a value
 A().g(f())      # E: "f" does not return a value
 x: A = f()      # E: "f" does not return a value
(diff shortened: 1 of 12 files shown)


ilevkivskyi:
OK, I have rewritten it without a flag and I like the new semantics even more, it looks more consistent.

This however required fixing an unrelated issue: previously all `Var`s at global and local scopes were _unconditionally_ marked as inferred, this is a one line fix however, so I don't want to make a separate PR for this.

test-data/unit/check-expressions.test
```diff
@@ -936,8 +936,8 @@ main:3: error: "A" not callable
 
 a, o = None, None # type: (A, object)
 a = f()         # E: "f" does not return a value
-o = a()         # E: Function does not return a value
-o = A().g(a)    # E: "g" of "A" does not return a value
+o = a()
+o = A().g(a)
```
JukkaL: I think that this should still generate an error, since `A.g` has an explicit `None` return type.

test-data/unit/check-expressions.test
```diff
@@ -936,8 +936,8 @@ main:3: error: "A" not callable
 
 a, o = None, None # type: (A, object)
 a = f()         # E: "f" does not return a value
-o = a()         # E: Function does not return a value
-o = A().g(a)    # E: "g" of "A" does not return a value
+o = a()
```
JukkaL: It would be okay for this to still generate an error, since `__call__` has an explicit `None` return, but since this is an edge case it's not that important. Might be worth adding an issues about this though.

ilevkivskyi:
@JukkaL 
Yes, makes sense, I added some more special-casing. I would however avoid further special-casing as this is (as you mentioned before) not required for correctness.

msullivan:
This seems to have introduced a regression where we will emit the `Function does not return a value` error message inside of functions that shouldn't be type checked:
```
def foo() -> None:
    pass

def lol():
    x = foo()
```
produces an error

ilevkivskyi:
Hm, interesting, I will check this.

PR #5670, opened 2018-09-25
Add a `--permissive-toplevel` flag that avoids partial type errors 

### LINKED ISSUE #5639: More permissive behavior when overriding class variables
The following code fails to typecheck:
```
class A:
    lol = None

class B(A):
    lol = 'b'
```
with
`test.py:5: error: Incompatible types in assignment (expression has type "str", base class "A" defined the type as "None")`

My proposal for fixing this is to add a new flag, on by default, that will cause class variables with the type `None` to instead be given the type `Optional[Any]`.

Not sure what the flag should be called, though.

(I'm investigating top causes of errors when running mypy on an unannotated code base and this issue caused 60% of the 600+ errors generated when checking django.)

msullivan:
A major cause of false positives when checking new codebases is
unannotated top level containers and Nones.

Add a flag that suppresses both the "needs type annotation" errors and
the assigning of NoneTyp to variables at the "toplevel" (not when
nested in functions). To avoid throwing away too much type
information, we give more precise types in these situations than
previously, filling in the arguments to the type with `Any`.

Closes #5639.

I don't love the flag name and would be happy to change it. This just adds an individual flag. Grouping into a `--strictness=0` configuration or something we should spin off as another bug, I think.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -527,6 +527,13 @@ def process_options(args: List[str],
                         help="Disallow decorating typed functions with untyped decorators",
                         group=untyped_group)
 
+    toplevel_group = parser.add_argument_group(
+        title='Top-level hecking',
+        description="Configure how code at the top-level is handled. ")
+    add_invertible_flag('--permissive-toplevel', default=False, strict_flag=False,
+                        help="Suppress toplevel errors caused by missing annotations",
+                        group=toplevel_group)
+
     none_group = parser.add_argument_group(
         title='None and Optional handling',
         description="Adjust how values of type 'None' are handled. For more context on "
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -528,8 +528,8 @@ def process_options(args: List[str],
                         group=untyped_group)
 
     toplevel_group = parser.add_argument_group(
-        title='Top-level hecking',
-        description="Configure how code at the top-level is handled. ")
+        title='Top-level checking',
+        description="Configure how code at the top-level is handled.")
     add_invertible_flag('--permissive-toplevel', default=False, strict_flag=False,
                         help="Suppress toplevel errors caused by missing annotations",
                         group=toplevel_group)
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -527,13 +527,6 @@ def process_options(args: List[str],
                         help="Disallow decorating typed functions with untyped decorators",
                         group=untyped_group)
 
-    toplevel_group = parser.add_argument_group(
-        title='Top-level checking',
-        description="Configure how code at the top-level is handled.")
-    add_invertible_flag('--permissive-toplevel', default=False, strict_flag=False,
-                        help="Suppress toplevel errors caused by missing annotations",
-                        group=toplevel_group)
-
     none_group = parser.add_argument_group(
         title='None and Optional handling',
         description="Adjust how values of type 'None' are handled. For more context on "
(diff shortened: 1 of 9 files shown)


mypy/main.py
```diff
@@ -527,6 +527,13 @@ def add_invertible_flag(flag: str,
                         help="Disallow decorating typed functions with untyped decorators",
                         group=untyped_group)
 
+    toplevel_group = parser.add_argument_group(
+        title='Top-level hecking',
```
gvanrossum: Typo: "hecking".

mypy/main.py
```diff
@@ -527,6 +527,13 @@ def add_invertible_flag(flag: str,
...
+        description="Configure how code at the top-level is handled. ")
```
gvanrossum: Drop the space after the period.

mypy/main.py
```diff
@@ -527,6 +527,13 @@ def add_invertible_flag(flag: str,
                         help="Disallow decorating typed functions with untyped decorators",
                         group=untyped_group)
 
+    toplevel_group = parser.add_argument_group(
```
gvanrossum: Shouldn't this be added to `strictness_group` rather than making a new toplevel group?

test-data/unit/check-inference.test
```diff
@@ -1392,8 +1392,9 @@ class A:
         self.x = [] # E: Need type annotation for 'x'
 
 class B(A):
+    # TODO?: This error is kind of a false positive, unfortunately
     @property
-    def x(self) -> List[int]:
+    def x(self) -> List[int]:  # E: Signature of "x" incompatible with supertype "A"
```
gvanrossum: Perhaps related to #3260?

test-data/unit/check-inference.test
```diff
@@ -2467,3 +2468,115 @@ def f() -> None:
...
+y = {}
+z = y
+
+
+[file a.py]
+from b import x, y
+reveal_type(x)  # E: Revealed type is 'builtins.list[Any]'
+reveal_type(y)  # E: Revealed type is 'builtins.dict[Any, Any]'
+
+[builtins fixtures/dict.pyi]
```
gvanrossum: I'm curious why this doesn't pass:
```py
class C:
    a = []  # Error here
    def f(self) -> None:
        reveal_type(self.a)
```
This still gives an error "Need type annotation for 'a'" on line 2.

Michael0x2a:
Just a few thoughts about how we're categorizing this flag:

1. We should add this flag to the config file docs somewhere -- probably as a per-module option?
2. One idea is maybe to group this under the "untyped definitions and calls" section and name the flag something like `--allow-untyped-globals`. (I'm also not happy with this name, but starting the flag with words like "allow/disallow" feels a little more consistent with what the other flags are doing.)

msullivan:
Re categorization:
The strictness group currently includes only the meta-option `--strict`, which doesn't seem like the right place for this. 
"Untyped definitions and calls" is the closest fit, and I nearly lumped it in there, but it seemed like something of an outlier still. I'd be happy to move it back in there.

gvanrossum:
I still think it belongs better in strictness_group. This looks as follows currently:
```
Other strictness checks:
  --strict                  Strict mode; enables the following flags: --warn-
                            unused-configs, --disallow-subclassing-any,
                            --disallow-untyped-calls, --disallow-untyped-defs,
                            --disallow-incomplete-defs, --check-untyped-defs,
                            --disallow-untyped-decorators, --no-implicit-
                            optional, --warn-redundant-casts, --warn-unused-
                            ignores, --warn-return-any
```

test-data/unit/check-inference.test
msullivan: I think the issue is that `x` in the parent is given the type `List[Any]` which isn't compatible with `List[int]`?

test-data/unit/check-inference.test
msullivan: Because member checking has a totally different partial type path apparently, which I'm now going to merge.

gvanrossum:
(Note that the flag ended up being renamed to `--allow-untyped-globals`, but the commit message wasn't adjusted. :-( )

Issue #604, opened 2015-03-18
Allow __init__ with signature but no return type


JukkaL:
Code:

```
class Visitor:
    def __init__(self, a: int):
        pass
```

Error:

```
x.py: In member "__init__" of class "Visitor":
x.py, line 2: Cannot define return type for "__init__"
```

The return type is `Any` (implicitly), and `Any` should be a valid return type for `__init__`.

This was reported by Guido.

gvanrossum:
Hm, the return type for __init__ should be None.

On Tuesday, March 17, 2015, Jukka Lehtosalo <notifications@github.com>
wrote:

> Code:
>
> class Visitor:
>     def __init__(self, a: int):
>         pass
>
> Error:
>
> x.py: In member "__init__" of class "Visitor":
> x.py, line 2: Cannot define return type for "__init__"
>
> The return type is Any (implicitly), and Any should be a valid return
> type for __init__.
>
> This was reported by Guido.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/JukkaL/mypy/issues/604>.
>


-- 
--Guido van Rossum (on iPad)

JukkaL:
Ah, the issue is that the error message is wrong. The message should say that the return type must be None? The message may be a relict from times before Python-compatible syntax.

gvanrossum:
But why do I have to specify the return type at all? Can't it default to
None here?

On Wednesday, March 18, 2015, Jukka Lehtosalo <notifications@github.com>
wrote:

> Closed #604 <https://github.com/JukkaL/mypy/issues/604> via 182a70b
> <https://github.com/JukkaL/mypy/commit/182a70b80b162b28c74f2b67a2b83ea6f78c8480>
> .
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/JukkaL/mypy/issues/604#event-258931926>.
>


-- 
--Guido van Rossum (on iPad)

JukkaL:
Because of consistency. Consider `__init__` that takes no arguments, other than `self`:

```
def __init__(self):
    ...
```

Would this be dynamically typed or statically typed? The current approach is to treat any function that has no annotation (no argument or return types) as dynamically typed, and statically typed (i.e. type checked) otherwise. The `None` return type marks the function as type checked.

Same goes for functions in general:

```
def f():   # Dynamically typed
    ...
def f() -> None:   # Statically typed
    ...
def g(x: int):  # Statically typed, Any return type
    ...
```

[...]

PR #5677, opened 2018-09-26
Make return type implicitly None for type checked __init__ and __init_subclass__

onlined:
Implements https://github.com/python/mypy/issues/604#issuecomment-348525995.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -406,10 +406,10 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
                             add_symbol = False
                     if add_symbol:
                         self.type.names[defn.name()] = SymbolTableNode(MDEF, defn)
-                if (defn.type is not None and
-                        defn.name() in ('__init__', '__init_subclass__') and
-                        isinstance(defn.type.ret_type, AnyType)):
-                    defn.type = defn.type.copy_modified(ret_type=NoneTyp())
+                if defn.type is not None and defn.name() in ('__init__', '__init_subclass__'):
+                    assert isinstance(defn.type, CallableType)
+                    if isinstance(defn.type.ret_type, AnyType):
+                        defn.type = defn.type.copy_modified(ret_type=NoneTyp())
                 self.prepare_method_signature(defn, self.type)
             elif self.is_func_scope():
                 # Nested function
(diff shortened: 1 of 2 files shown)


mypy/semanal.py
```diff
@@ -407,6 +406,10 @@ def _visit_func_def(self, defn: FuncDef) -> None:
                             add_symbol = False
                     if add_symbol:
                         self.type.names[defn.name()] = SymbolTableNode(MDEF, defn)
+                if defn.type is not None and defn.name() in ('__init__', '__init_subclass__'):
```
gvanrossum: Are you sure this works with overloads and decorated methods? Please add tests to verify that.

Also perhaps it should only work inside a class definition?

mypy/semanal.py
```diff
@@ -68,8 +68,7 @@
 from mypy.types import (
     FunctionLike, UnboundType, TypeVarDef, TupleType, UnionType, StarType, function_type,
     CallableType, Overloaded, Instance, Type, AnyType,
-    TypeTranslator, TypeOfAny, TypeType,
-)
+    TypeTranslator, TypeOfAny, TypeType, NoneTyp)
```
gvanrossum: Please leave the `)` on a line by itself and add a trailing comma, preserving the original formatting.

mypy/semanal.py
ilevkivskyi: I think one needs similar check at least in `visit_overloaded_func_def`.

onlined: @gvanrossum It works only inside a class definition.
@gvanrossum  @ilevkivskyi You are right, I will find a way to extend behavior.

gvanrossum:
Can you also add a test for overloading?

onlined:
I'm on it, but I can say that this change works with overloading too.

onlined:
@gvanrossum It may be good to add tests for explicit `Any` return type, what do you think about it?

gvanrossum:
> @gvanrossum It may be good to add tests for explicit `Any` return type, what do you think about it?

Yes, please.

onlined:
I think this PR is ready to merge.

gvanrossum:
Sorry, I got busy with other stuff. Will review later today.

gvanrossum:
Done (sorry for the delay). Thanks for fixing this long-standing issue!

PR #5683, opened 2018-09-26
Don't report "function does not return a value" in dynamic functions

msullivan:
Fixes a regression (that causes trouble in internal codebases) introduced in #5663.

### DIFF
--- a/test-data/unit/check-functions.test
+++ b/test-data/unit/check-functions.test
@@ -2376,3 +2376,11 @@ def Y(x: Optional[str] = X(x_in)): ...
 
 xx: Optional[int] = X(x_in)
 [out]
+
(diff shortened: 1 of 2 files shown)


test-data/unit/check-functions.test
```diff
@@ -2376,3 +2376,11 @@ def Y(x: Optional[str] = X(x_in)): ...
 
 xx: Optional[int] = X(x_in)
 [out]
+
+[case testNoComplainNoneReturnFromUntyped]
+def foo() -> None:
+    pass
+
+def lol():
+    1+'uh'
```
ilevkivskyi: Maybe add spaces around `+` (totally up to you).

msullivan: Oh actually this should just not be there at all

PR #5685, opened 2018-09-27
Add --disallow-any-generics to --strict

### LINKED ISSUE #5684: Assigning TypeVared generic type to variable loses typing information
If I create an alias for a generic type using a `TypeVar`, Mypy seems to infer the `TypeVar` as `Any` and lose the typing information. For example, this works:
```python

T = TypeVar('T')


def wrapper(func: Callable[[], T]) -> T:
    reveal_type(func)
    reveal_type(func())
    return func()
```
and gives these types:
```
mypy --strict generic_callable.py 
generic_callable.py:7: error: Revealed type is 'def () -> T`-1'
generic_callable.py:8: error: Revealed type is 'T`-1'
```
I would expect this
```python

T = TypeVar('T')
C = Callable[[], T]


def wrapper(func: C) -> T:
    reveal_type(func)
    reveal_type(func())
    return func()
```
to behave in the same way, but it doesn't:
```
mypy --strict generic_callable.py 
generic_callable.py:8: error: Revealed type is 'def () -> Any'
generic_callable.py:9: error: Revealed type is 'Any'
generic_callable.py:10: warning: Returning Any from function declared to return "T"
```
This can be a problem if I am using the same complex generic type multiple times, as in this real-world example:
```python
import functools
import warnings

_ReturnType = TypeVar('_ReturnType')
_CallableType = Callable[..., _ReturnType]


def deprecated(msg: str) -> Callable[[_CallableType], _CallableType]:
    def outer_wrapper(func: _CallableType) -> _CallableType:
        @functools.wraps(func)
        def inner_wrapper(*args: object, **kwargs: object) -> _ReturnType:
            warnings.simplefilter('always', DeprecationWarning)
            warnings.warn(f"{func.__name__} is deprecated: {msg}.",
                          category=DeprecationWarning,
                          stacklevel=2)
            warnings.simplefilter('default', DeprecationWarning)  # reset filter
            return func(*args, **kwargs)

        return inner_wrapper

    return outer_wrapper
```
Having to explicitly type out `Callable[..., _ReturnType]` everywhere I use `_CallableType` would be a little cumbersome in this case.

TV4Fun:
This is a tiny bit hacky as --disallow-any-generics is not defined
through `add_invertible_flag`. Resolves #5684.

@ilevkivskyi

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -498,10 +498,13 @@ def process_options(args: List[str],
     disallow_any_group.add_argument(
         '--disallow-any-explicit', default=False, action='store_true',
         help='Disallow explicit Any in type positions')
-    disallow_any_group.add_argument(
+    dest = disallow_any_group.add_argument(
         '--disallow-any-generics', default=False, action='store_true',
         help='Disallow usage of generic types that do not specify explicit '
-             'type parameters')
+             'type parameters').dest
+    assert dest is not None
+    strict_flag_names.append('--disallow-any-generics')
+    strict_flag_assignments.append((dest, True))
 
     untyped_group = parser.add_argument_group(
         title='Untyped definitions and calls',


TV4Fun:
BTW, I think --disallow-any-generics is not a good name for this flag, as it implies that it doesn't allow generics explicitly parameterized with `Any`. I think a better name would be --disallow-unparameterized-generics.

mypy/main.py
```diff
@@ -498,10 +498,13 @@ def add_invertible_flag(flag: str,
...
         help='Disallow explicit Any in type positions')
-    disallow_any_group.add_argument(
+    dest = disallow_any_group.add_argument(
         '--disallow-any-generics', default=False, action='store_true',
         help='Disallow usage of generic types that do not specify explicit '
-             'type parameters')
+             'type parameters').dest
+    assert dest is not None
+    strict_flag_names.append('--disallow-any-generics')
+    strict_flag_assignments.append((dest, True))
```
ilevkivskyi: Why can't we just use `add_invertible_flag` here? Like, for example, `--disallow-subclassing-any` above.

TV4Fun: Yes, that does make more sense. Fixed to do that now.

TV4Fun:
Added a couple of test cases for `--disallow-any-generics`. Interestingly, though there were already two test cases for `--disallow-any-generics`, neither actually tested for an error condition, so added that. I wanted to add a test for `--strict` as there aren't any now, but that seemed to create some errors unrelated to my test case:
```
----------------------------- Captured stderr call -----------------------------
Expected:
  main:7: error: Missing type parameters for generic type (diff)
Actual:
  tmp/builtins.pyi:8: error: Function is missing a type annotation (diff)
  tmp/builtins.pyi:34: error: Call to untyped function "object" in typed context (diff)
  main:7: error: Missing type parameters for generic type (diff)

  E: main:7: error: Missing type parameters for generic type
  A: tmp/builtins.pyi:8: error: Function is missing a type annotation
```
The test that caused this was:
```
[case testCheckStrictAnyGeneric]
# flags: --strict

T = TypeVar('T')
C = Callable[[], T]

def f(c: C) -> None:  # E: Missing type parameters for generic type
    pass
[builtins fixtures/list.pyi]
[out]
```
Not sure if this is a known issue or not.

JelleZijlstra:
This is probably because the stubs we use for builtins in tests aren't `--strict`-clean. Might be worth fixing.

ilevkivskyi:
(Also would be good to fix the commit history, but it is not essential since we will squash anyway.)

test-data/unit/check-flags.test
```diff
@@ -1124,3 +1124,38 @@ GroupDataDict = TypedDict(  # type: ignore
...
+
+[case testCheckDefaultAllowAnyGeneric]
+from typing import TypeVar, Callable
+
+T = TypeVar('T')
+C = Callable[[], T]
+
+def f(c: C):
+    pass
+[builtins fixtures/list.pyi]
```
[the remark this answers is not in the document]
TV4Fun: You are correct, removed.

test-data/unit/check-flags.test
```diff
@@ -1124,3 +1124,38 @@ GroupDataDict = TypedDict(  # type: ignore
...
+[out]
+
+[case testCheckDisallowAnyGenericsAnyGeneric]
+# flags: --disallow-any-generics
+from typing import TypeVar, Callable
+
+T = TypeVar('T')
+C = Callable[[], T]
+
+def f(c: C):  # E: Missing type parameters for generic type
```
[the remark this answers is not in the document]
TV4Fun: Added, and also tested for `--strict`.

TV4Fun:
Comments addressed. I agree the commit history is a little messed up, but as you said, you are going to squash anyway.

test-data/unit/check-flags.test
```diff
@@ -1124,3 +1124,48 @@ GroupDataDict = TypedDict(  # type: ignore
...
+
+T = TypeVar('T')
+C = Callable[[], T]
+
+def f(c: C):
+    pass
+[out]
+
+[case testCheckAllowAnyGenericAnyGeneric]
+# flags: --allow-any-generics
```
ilevkivskyi: I would instead test that setting both `--strict --allow-any-generics` (with strict being first) still allows them.

TV4Fun: Done.

gvanrossum:
I'm sorry, the 0.640 release train has already closed its doors and is firing up its engine. It's too late too get on board. Take the next one in 3-4 weeks!

PR #5686, opened 2018-09-27
Namespaces refactor

gvanrossum:
In preparation of working on PEP 420 namespaces, here's some initial work. IT DOES NOT IMPLEMENT NAMESPACES YET. What it does do:

- Move nearly 400 lines from build.py to a new file, `modulefinder.py`.  This includes `SearchPaths`, `BuildSource`, `FindModuleCache`, `mypy-path`, `default_lib_path`, `get_site_packages` (now without underscore!), and `compute_search_paths`.
- Slight refactor to `FindModuleCache` so that the `search_paths` are passed to the constructor instead of to `find_module`.
- Removed `search_paths` and `python_executable` from the signature of `find_module` and `find_modules_recursive`, and also from the cache key (**this may be the most controversial change** -- it's not just a refactor).
- Add a (non-functional) `--namespace-packages` flag to `main.py` and add new global config option `namespace_packages`. (These are not used yet.)

I'm presenting this as a separate PR because it's a significant refactor without change of functionality -- everything I plan to do after this will mostly tweak the code in `modulefinder.py`. It seems fairer to reviewers to be able to review this massive code move without having to worry too much about "what else changed".

(Note that this is part of my general plan to move functionality out of `build.py` -- that file is (still) way too bulky.)

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -486,7 +486,7 @@ class BuildManager(BuildManagerBase):
         self.cache_enabled = options.incremental and (
             not options.fine_grained_incremental or options.use_fine_grained_cache)
         self.fscache = fscache
-        self.find_module_cache = FindModuleCache(self.fscache, self.options)
+        self.find_module_cache = FindModuleCache(self.search_paths, self.fscache, self.options)
 
         # a mapping from source files to their corresponding shadow files
         # for efficient lookup
@@ -616,8 +616,7 @@ class BuildManager(BuildManagerBase):
 
     def is_module(self, id: str) -> bool:
         """Is there a file in the file system corresponding to module id?"""
-        return self.find_module_cache.find_module(id, self.search_paths,
-                                                  self.options.python_executable) is not None
+        return self.find_module_cache.find_module(id) is not None
 
     def parse_file(self, id: str, path: str, source: str, ignore_errors: bool) -> MypyFile:
         """Parse the source of a file with the given name.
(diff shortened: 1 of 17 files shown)


mypy/build.py
```diff
@@ -840,197 +648,6 @@ def stats_summary(self) -> Mapping[str, object]:
...
-        dirs = []
-        for pathitem in lib_path:
-            # e.g., '/usr/lib/python3.4/foo/bar'
-            dir = os.path.normpath(os.path.join(pathitem, dir_chain))
-            if self.fscache.isdir(dir):
-                dirs.append((dir, True))
-        return dirs
-
-    def find_module(self, id: str, search_paths: SearchPaths,
-                    python_executable: Optional[str]) -> Optional[str]:
```
ilevkivskyi: FWIW `python_executable` is never actually used by `find_module`, `_find_module`, and `find_module_recursive`, it just gets passed around but I didn't find a place where it is used. As I understand, all the info is stored already in `search_paths` that are calculated using the executable.

mypy/modulefinder.py
```diff
@@ -0,0 +1,397 @@
...
+
+class BuildSource:
+    """A single source file."""
+
+    def __init__(self, path: Optional[str], module: Optional[str],
+                 text: Optional[str], base_dir: Optional[str] = None) -> None:
+        self.path = path  # File where it's found (e.g. 'xxx/yyy/foo/bar.py')
+        self.module = module or '__main__'  # Module name (e.g. 'foo.bar')
+        self.text = text  # Source code, if initially supplied, else None
+        self.base_dir = base_dir    # Directory where the package is rooted (e.g. 'xxx/yyy')
```
ilevkivskyi: Remove extra spaces before this comment?

mypy/build.py
gvanrossum: Yes, that's why I removed it from their signatures in the refactor.

mypy/modulefinder.py
gvanrossum: Will do.

PR #5691, opened 2018-09-29
Namespace packages (PEP 420)

### LINKED ISSUE #1645: MYPYPATH doesn't respect PEP420 namespace packages
Take this directory tree:

```
/path/to/src
└───namespace
    └───package
        │   __init__.py
        │
        └───__pycache__
                __init__.cpython-35.pyc
```

And this MYPYPATH:

```
> $env:MYPYPATH
/path/to/src
```

then running `mypy -p namespace.package` gives:

```
> mypy -p namespace.package
Can't find package 'namespace.package'
```

However, mypy works fine when you add an empty `__init__.py` into `namepace` (turning it into a regular package).

gvanrossum:
Tentative implementation of PEP 420. Fixes #1645.

Clarification of the implementation:

- `candidate_base_dirs` is a list of `(dir, verify)` tuples, laboriously pre-computed.
- The old code just looped over this list, looking for `dir/<module>`, and if found, it would verify that there were `__init__.py` files in all the right places (except if `verify` is false); the first success would be the hit;
- In PEP 420 mode, if the above approach finds no hits, we do something different for those paths that failed due to `__init__` verification; essentially we narrow down the list of candidate paths by checking for `__init__` files from the top down. Hopefully the last test added clarifies this.

### DIFF
--- a/mypy/modulefinder.py
+++ b/mypy/modulefinder.py
@@ -196,12 +196,27 @@ class FindModuleCache:
                         continue
                     return path
 
-        # In namespace mode, re-check those entries that had 'verify'
-        # set using different rules.  This time we allow missing
-        # __init__.py[i] files and we don't look for -stubs.  It's
-        # complicated because if there's a classic subpackage
-        # somewhere it should still be preferred over a namespace
-        # subpackage with the same name earlier on the path.
+        # In namespace mode, re-check those entries that had 'verify'.
+        # Assume search path entries xxx, yyy and zzz, and we're
+        # looking for foo.bar.baz.  Suppose near_misses has:
+        #
+        # - xxx/foo/bar/baz.py
+        # - yyy/foo/bar/baz/__init__.py
+        # - zzz/foo/bar/baz.pyi
+        #
+        # If any of the foo directories has __init__.py[i], it wins.
+        # Else, we look for foo/bar/__init__.py[i], etc.  If there are
+        # none, the first hit wins.  Note that this does not take into
+        # account whether the lowest-level module is a file (baz.py),
+        # a package (baz/__init__.py), or a stub file (baz.pyi) -- for
+        # these the first one encountered along the search path wins.
+        #
+        # The helper function highest_init_level() returns an int that
+        # indicates the highest level at which a __init__.py[i] file
+        # is found; if no __init__ was found it returns 0, if we find
+        # only foo/bar/__init__.py it returns 1, and if we have
+        # foo/__init__.py it returns 2 (regardless of what's un
+        # foo/bar).  It doesn't look higher than that.
         if self.options and self.options.namespace_packages and near_misses:
             levels = [highest_init_level(fscache, id, path) for path in near_misses]
             index = levels.index(max(levels))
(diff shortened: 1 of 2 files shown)


gvanrossum:
(Whoops, accidentally deleted the branch which caused the PR to be closed. I'm not done yet.)

test-data/unit/check-modules.test
```diff
@@ -2525,3 +2526,51 @@ def __radd__(self) -> int: ...
...
+from foo.bar import x
+[file foo/bar.py]
+x = 0
+[out]
+main:1: error: Cannot find module named 'foo.bar'
+main:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
+
+[case testNamespacePackage]
+# flags: --namespace-packages
+from foo.bar import x
```
ilevkivskyi: Just to be on the safe side I would reveal type of `x` here and below.

test-data/unit/check-modules.test
```diff
@@ -2525,3 +2526,51 @@ def __radd__(self) -> int: ...
...
+from foo.baz import z
+[file xx/foo/bax.py]
+x = 0
+[file yy/foo/bay.py]
+y = 0
+[file foo/baz.py]
+z = 0
+[file mypy.ini]
+[[mypy]
+mypy_path = tmp/xx, tmp/yy
```
ilevkivskyi: What happens if there are two files with the same name in different paths? For example, `xx/foo/bar.py` and `yy/foo/bar.py`.

ilevkivskyi:
As I understand this fixes https://github.com/python/mypy/issues/1645. I would add it to the PR description, so that we will not forget to close the issue :)

gvanrossum:
Alas, tests for nesting one type of package inside the other fail (both ways -- it always prefers the decoy file). It's late so I'll just leave it broken, I'll fix it tomorrow.

gvanrossum:
Hey @carljm, IIRC you expressed interest in namespace packages eons ago. Perhaps you can test this PR with your use case?

gvanrossum:
(Hold on, I have an idea for a refinement of the final part of the algorithm.)

gvanrossum:
@ilevkivskyi I think I'm happy now. Are you?

mypy/modulefinder.py
```diff
@@ -178,19 +179,49 @@ def _find_module(self, id: str) -> Optional[str]:
...
+        # none, the first hit wins.  Note that this does not take into
+        # account whether the lowest-level module is a file (baz.py),
+        # a package (baz/__init__.py), or a stub file (baz.pyi) -- for
+        # these the first one encountered along the search path wins.
+        #
+        # The helper function highest_init_level() returns an int that
+        # indicates the highest level at which a __init__.py[i] file
+        # is found; if no __init__ was found it returns 0, if we find
+        # only foo/bar/__init__.py it returns 1, and if we have
+        # foo/__init__.py it returns 2 (regardless of what's un
```
ilevkivskyi: typo: 'un' -> 'in'

gilbsgilbs:
FWIW, I tried this on some real-world projects and it's working great. Thanks!

PR #5694, opened 2018-09-29
Remove unused imports

ceh:


mypy/checkexpr.py
```diff
@@ -18,9 +18,8 @@
 from mypy.types import (
     Type, AnyType, CallableType, Overloaded, NoneTyp, TypeVarDef,
     TupleType, TypedDictType, Instance, TypeVarType, ErasedType, UnionType,
-    PartialType, DeletedType, UnboundType, UninhabitedType, TypeType, TypeOfAny,
-    true_only, false_only, is_named_instance, function_type, callable_type, FunctionLike,
-    get_typ_args, StarType
+    PartialType, DeletedType, UninhabitedType, TypeType, TypeOfAny, true_only,
+    false_only, is_named_instance, function_type, callable_type, FunctionLike, StarType
```
gvanrossum: Can you add a trailing comma? In lists like this that's how we like to format them.

mypy/stats.py
```diff
@@ -1,10 +1,9 @@
 """Utilities for calculating and reporting statistics about types."""
 
 import os.path
-import typing
 
 from collections import Counter
-from typing import Dict, List, cast, Optional
+from typing import cast, Counter, Dict, List, Optional
```
gvanrossum: This doesn't seem right. It overrides the import of `Counter` on the previous line. I think there's a reason why the code isn't doing that.

gvanrossum: This doesn't seem right. It overrides the import of `Counter` on the previous line. I think there's a reason why the code did what it did.

mypy/checkexpr.py
ceh: Fixed.

mypy/stats.py
ceh: Ah, you are correct. I missed this. Thanks for the helpful comment.

gvanrossum:
Thanks again!

PR #5700, opened 2018-10-01
Fix plugin hooks not being called correctly for dunder methods

### LINKED ISSUE #4964: Plugin hooks are not called or receive incorrect names for some dunder methods
(Taken from #4869)

When certain dunder methods are called implicitly through language syntax, the plugin function/method hooks are called incorrectly or not at all. Specifically, the following cases cause problems:

* [ ] For some dunder methods (`__setitem__`, `__iter__`, `__next__`), `get_function_hook` is incorrectly called instead of `get_method_hook`, and the passed in method name is not fully qualified
* [ ] `get_method_signature_hook` is not called for some dunder methods (`__getitem__`, `__setitem__`, `__iter__`, `__next__`)
* [x] When a `Type[...]` expression is called (that is not a simple class name), the class name passed to `get_function_hook` is not fully qualified (this is not strictly related to dunder methods, it might be a separate problem, I'm not sure) (fixed in #5379)

To test this, I've added `print`s to `mypy.plugin.DefaultPlugin`'s methods `get_function_hook`, `get_method_signature_hook`, `get_method_hook`, and `get_attribute_hook` to show when each method is called with what argument. Then I ran `mypy` on the following code:

```python
# fixed
# * class name is not fully qualified ('list') (fixed in #5379)
seq = (list,)[0](("hello",))

# works
# * class name is fully qualified ('builtins.list')
seq = list(("hi",))

# buggy
# * name is not fully qualified ('__iter__ of list')
# * get_function_hook is called instead of get_method_hook
# * get_method_signature_hook is not called
for obj in seq:
    print(obj)

# buggy
# * get_method_signature_hook is not called
print(seq[0])

# buggy
# * name is not fully qualified ('__setitem__ of list')
# * get_function_hook is called instead of get_method_hook
# * get_method_signature_hook is not called
seq[0] = "bye"

# works
# * name is fully qualified ('builtins.list.clear')
# * get_method_signature_hook is called
seq.clear()
```

Output:

```
get_function_hook('builtins.list')
get_function_hook('builtins.list')
get_function_hook('__iter__ of list')
get_attribute_hook('typing.Iterator.__next__')
get_function_hook('__next__ of Iterator')
get_function_hook('builtins.print')
get_method_hook('builtins.list.__getitem__')
get_function_hook('builtins.print')
get_function_hook('__setitem__ of list')
get_method_signature_hook('builtins.list.clear')
get_method_hook('builtins.list.clear')
```

dgelessus:
Fixes #4964. The two commits fix the first and second bullet point, respectively.

Please be aware that I don't the mypy codebase very well. I checked that these changes work for my purposes and pass the test suite, but I don't know if this is the right/best way to fix these issues. (In particular, the second commit does some smallish refactoring that I'm not entirely sure about.) Any feedback is welcome.

Also, I noticed that because of the changes in my first commit, the method signature hooks are now called "too often" for overloaded methods: the hook is first applied to all overloads of the method, then a matching overload is selected, and the hook is applied *again* to the selected overload. Previously the hook was only applied before overload selection. I have no idea if this change matters in practice, or if it's worth fixing.

@ilevkivskyi


mypy/checkexpr.py
```diff
@@ -1781,31 +1784,36 @@ def get_operator_method(self, op: str) -> str:
...
+                                  base_type: Type,
+                                  args: List[Expression],
+                                  arg_kinds: List[int],
+                                  context: Context,
+                                  local_errors: Optional[MessageBuilder] = None,
+                                  ) -> Tuple[Type, Type]:
+        """Type check a call to a named method on an object.
 
-        Return tuple (result type, inferred operator method type).
+        Return tuple (result type, inferred method type).
```
ilevkivskyi: _Maybe_ this is a better place to add hook application? If necessary (after adding support for more types in `visit_call_expr_inner`) you can refactor that logic to a helper method.

mypy/checkexpr.py
```diff
@@ -613,6 +607,15 @@ def check_call(self, callee: Type, args: List[Expression],
...
         """
         arg_messages = arg_messages or self.msg
+
+        if (callable_name is not None and object_type is not None
+                and isinstance(callee, FunctionLike)):
+            # Apply plugin signature hook that may generate a better signature.
+            signature_hook = self.plugin.get_method_signature_hook(callable_name)
+            if signature_hook:
+                callee = self.apply_method_signature_hook(
+                    callee, args, arg_kinds, context, arg_names, object_type, signature_hook)
```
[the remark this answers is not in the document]
dgelessus: I think I understand what you mean - `check_call` is supposed to receive a concrete callable type and check that the given argument list can be applied to that callable type. The code that calls `check_call` has already determined the callable type, so `check_call` shouldn't try to modify it using plugin hooks.

The reason why I put the plugin hook call into `check_call` is because of how method calls are checked in mypy. The usual pattern seems to be:

1. Look up the type of the called method (using `checkexpr.ExpressionChecker.accept` for "real" calls, or `checkmember.analyze_member_access` for implicit calls)
2. Check the actual arguments of the method call against the method type (using `checkexpr.ExpressionChecker.check_call`)

It's not possible to call method signature hooks in the first step - the functions/methods used there are generic to all attribute lookups, not just method calls, so they don't receive the method call-specific information needed by method signature hooks (specifically, information about the actual argument types/values of the call). Since only the second step (`check_call`) has all the necessary information, I put the method signature hook call there.

Maybe it would be better to instead refactor the "look up and apply method signature hook" code into a helper method, and call that manually between steps 1 and 2 where appropriate.

mypy/checkexpr.py
```diff
@@ -297,12 +297,6 @@ def visit_call_expr_inner(self, e: CallExpr, allow_none_return: bool = False) ->
                 if info:
                     fullname = '{}.{}'.format(info.fullname(), e.callee.name)
                     object_type = callee_expr_type
-                    # Apply plugin signature hook that may generate a better signature.
```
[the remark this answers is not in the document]
dgelessus: I'll have a look. Agreed with your other comment below, this code should probably go into a helper method that converts a callee type and a method name into a fully qualified name.

mypy/checkexpr.py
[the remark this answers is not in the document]
ilevkivskyi: > Maybe it would be better to instead refactor the "look up and apply method signature hook" code into a helper method, and call that manually between steps 1 and 2 where appropriate.

This sounds like a promising approach.

dgelessus:
Sorry for the long delay, things were busy for me in the last few weeks.

I've mostly implemented the changes discussed above, but I ran into a couple of issues:

* The "get and apply method signature hook if applicable" code is now in a method called `transform_callee_type`. I'm not sure if it's the best name - `apply_method_signature_hook` was already taken, so I used a more generic name.
  * However, this refactoring has re-broken method signature hooks for `__init__` methods. (One of the tests I added is failing because of this.) I don't see any good way to apply method signature hooks when the `__init__` method is looked up, because that is done in a separate function (`checkmember.type_object_type`) with no access to the `ExpressionChecker`.
* The "object type and method name to fully qualified name" code is in `method_fullname`, and now also supports `TupleType`.
  * I also tried supporting `CallableType`s that have `is_type_obj() == True`, but I couldn't figure out a way to get the fully qualified name of the type that a `CallableType` represents. As far as I can tell, the full name is currently "lost" when a class is converted to a `CallableType`.

ilevkivskyi:
@dgelessus 
> However, this refactoring has re-broken method signature hooks for `__init__` methods

Maybe this is actually OK? We can have some kind of agreement, that class instantiation should be done via function hook applied to the class object. After all "There should be only one way to do it" probably applies here.

> but I couldn't figure out a way to get the fully qualified name of the type that a `CallableType` represents

This is how you do it (not tested but will likely work):
```python
if isinstance(tp, CallableType) and tp.is_type_obj():
    if isinstance(tp.ret_type, Instance):  # Normal class objects
        fullname = tp.ret_type.type.fullname()
    elif isinstance(tp.ret_type, TupleType):  # named tuples and similar type objects
        fullname = tp.ret_type.fallback.type.fullname()
    else:
        assert False, "To the best of my knowledge we don't support other type objects"
```

I will make another pass of review now.

mypy/checkexpr.py
```diff
@@ -1802,40 +1827,44 @@ def get_operator_method(self, op: str) -> str:
...
-    def check_op_local(self,
-                       method_name: str,
-                       method_type: Type,
-                       base_type: Type,
-                       arg: Expression,
-                       context: Context,
-                       local_errors: MessageBuilder) -> Tuple[Type, Type]:
-        """Type check a binary operation using the (assumed) type of the operator method.
+        method_type = self.transform_callee_type(
+            callable_name, method_type, args, arg_kinds, context, object_type=object_type)
```
ilevkivskyi: Again, a short motivational comment would be helpful for a future reader here.

test-data/unit/check-custom-plugin.test
```diff
@@ -229,3 +229,50 @@ def decorator2() -> Callable[..., Callable[..., int]]: pass
...
+
+FullyQualifiedTestNamedTuple = NamedTuple('FullyQualifiedTestNamedTuple', [('foo', str)])
+
+# Check the return types to ensure that the method signature hook is called in each case
+reveal_type(FullyQualifiedTestClass().instance_method()) # E: Revealed type is 'builtins.int'
+reveal_type(FullyQualifiedTestNamedTuple('')._asdict()) # E: Revealed type is 'builtins.int'
+
+[file mypy.ini]
+[[mypy]
+plugins=<ROOT>/test-data/unit/plugins/fully_qualified_test_hook.py
```
ilevkivskyi: I think it would be great to add a test for class objects (assuming the fix I suggested works).

mypy/checkexpr.py
```diff
@@ -321,6 +309,26 @@ def visit_call_expr_inner(self, e: CallExpr, allow_none_return: bool = False) ->
             return AnyType(TypeOfAny.from_error)
         return ret_type
 
+    def method_fullname(self, object_type: Type, method_name: str) -> Optional[str]:
+        """Convert a method name to a fully qualified name, based on the type of the object that
+        it is invoked on. Return `None` if the name of `object_type` cannot be determined.
+        """
+
+        # TODO: Support CallableType (i. e. class methods) and possibly others
```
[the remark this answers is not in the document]
ilevkivskyi: Sorry, not in the issue, I left the comment just above in this PR, need to sleep more :-)

dgelessus:
> We can have some kind of agreement, that class instantiation should be done via function hook applied to the class object.

I agree in general, ideally there shouldn't be two different ways to hook into class instantiation. However I have a case where a function hook is not sufficient: when dealing with the `ctypes.Array` constructor in #4869. This is because of how arrays are created in `ctypes`:

```python
from ctypes import *
array_type = c_int * 4
array_obj = array_type(1, 2, 3, 4)
```

The type of `array_type` here is `Type[Array[c_int]]`. The parameter types of the `array_type` constructor vary depending on the type argument of `array_type` (in this case `c_int`). Since I need to modify the parameter types, a normal function hook is not enough.

Basically I'd need a "function signature hook" here, but since the plugin interface doesn't have that, I wrote a method signature hook on `__init__` instead.

But aside from that case I agree, there shouldn't be any need to call plugin hooks on `__new__` and `__init__`.

> This is how you do it

Ah, I didn't think of looking at the signature's return type. That makes perfect sense, thanks for the hint.

test-data/unit/check-custom-plugin.test
dgelessus: Right, I actually had a test case for that, but removed it while testing. I'll re-add it once I get method signature hooks working on class methods.

dgelessus: Something else I forgot to mention: I also wanted to add a test case for method calls on `TypedDict`s (that's why the `FullyQualifiedTestTypedDict` declaration is there). But it seems that mypy doesn't allow *any* method calls on `TypedDict` instances, even standard methods like `copy`. Am I missing something here?

If method calls on `TypedDict`s are really not allowed by mypy, it doesn't make much sense to support them in `method_fullname` (or at least there should be a comment explaining that the `TypedDict` branch is never called at the moment).

ilevkivskyi:
> Basically I'd need a "function signature hook" here, but since the plugin interface doesn't have that, I wrote a method signature hook on `__init__` instead.

Hm, I see. So basically you just want to type-check that arguments are valid for given callable. It looks like we have only two ways to proceed here:
* Add `function_signature_hook`, that will essentially restore the symmetry between functions/methods. But will be probably very rarely used.
* Try using the current function hook instead. It has all the necessary things to generate mypy errors (`ctx.api.msg.fail("Error message", ctx.context)` will do the things). Then you can get the type argument from `ctx.default_return_type` and type check the arguments "manually" by importing `mypy.subtypes.is_subtype()`.

The second option looks a bit hacky, but TBH I prefer the second option, because although it may be a bit more tedious, I seems to me the current situation is rather an exception to introduce a new hook only for it. What do you think?

dgelessus:
You're right, in many cases a normal function hook is enough, but there are also some use cases for a function signature hook. An example outside of the `ctypes` plugin would be `map(func, *iters)`, where a function signature hook could be used to inter the types of the `iters` based on the type of `func`. (This is just an example, it's really not necessary in practice, because the typeshed stubs have specialized `map` overloads for up to 5 iterables.)

Yes, this could also be implemented using the second option with manual type checking and errors, but as you said that seems a bit hacky 😃 But I think I'll use the second option for now in the `ctypes` plugin, since it's the only option that doesn't require more extra work. If function signature hooks are added to mypy at some point, the `ctypes` plugin can be changed to use them instead.

PR #5702, opened 2018-10-01
Use sys.executable for invoking mypy in runtests if it is Python 3

ethanhs:
I recently installed Python 3.7, which means that the `py` launcher will use it as default (even though I had installed it for testing...). I have Python 3.6 as the default for executing scripts, and the main development environment I use.

This change will use the Python executable used to run the script, instead of the default Python 3 executable (via `py -3`), if possible, or will fall back to the old behavior if the default to run a script is a Python 2 executable.

@Michael0x2a hopefully this doesn't break anything for you?

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -6,7 +6,7 @@ prog, *args = argv
 
 
 # Use the Python provided to execute the script, or fall back to a sane default
-if version_info > (3, 0, 0):
+if version_info >= (3, 4, 0):
         python_name = executable
 else:
     if platform == 'win32':
--- a/runtests.py
+++ b/runtests.py
@@ -1,11 +1,16 @@
 #!/usr/bin/env python3
 from os import system
-from sys import argv, exit, platform
+from sys import argv, exit, platform, executable, version_info
 
 prog, *args = argv
 
 if platform == 'win32':
-    python_name = 'py -3'
+    # use the executable the script was invoked with
+    # or default to Python 3.x
+    if version_info > (3, 0, 0):
+        python_name = executable
+    else:
+        python_name = 'py -3'
 else:
     python_name = 'python3'
 
--- a/runtests.py
+++ b/runtests.py
@@ -4,15 +4,15 @@ from sys import argv, exit, platform, executable, version_info
 
 prog, *args = argv
 
-if platform == 'win32':
-    # use the executable the script was invoked with
-    # or default to Python 3.x
-    if version_info > (3, 0, 0):
+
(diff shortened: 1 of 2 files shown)


ethanhs:
>One question though -- would it maybe be a good idea to just default to using sys.executable for both windows and linux/mac? For example, maybe somebody using Linux might want to try running this script using a version of Python other then whatever python3 is pointing to.

Yes, that makes sense. Done.

gvanrossum:
While you're at it, could you also add a mention of `runtests.py` to the toplevel `README.md`? Probably in the section on Tests.

ethanhs:
>Or did you mean >=?

I did, however now that I think about it, it should be >= 3.4.0 since that is the minimum mypy supports.

runtests.py
```diff
@@ -1,13 +1,18 @@
...
-from sys import argv, exit, platform
+from sys import argv, exit, platform, executable, version_info
 
 prog, *args = argv
 
-if platform == 'win32':
-    python_name = 'py -3'
+
+# Use the Python provided to execute the script, or fall back to a sane default
+if version_info >= (3, 4, 0):
```
gvanrossum: As long as we're nit-picking here, I'd say `if version_info[0] >= 3` -- this speaks to our intention which is that we're really just distinguishing between (supported versions of) Python 2 and 3.

ethanhs: Perhaps you missed my comment [here](https://github.com/python/mypy/pull/5702#issuecomment-426069096), but I think this is not the right thing to do. We want to select only a version of Python 3 that mypy supports.

Michael0x2a:
I think it would also be reasonable to have this script just terminate if somebody tries running it with an unsupported Python version. I think it's reasonable to expect anybody who's hacking on mypy to know that you need to run it using only Python 3 -- and if they're not aware, it's probably easier to correct that misconception sooner rather then later.

(But like I said, I really don't have strong opinions about this script)

runtests.py
gvanrossum: I did see that comment but wasn't sure how it mattered...

So the case where this matters is apparently one where runtests.py is explicitly invoked with an unsupported version of Python 3, e.g. if the user typed `python3.3 runtests.py` while the default `python3` refers to 3.4 or higher. Right? I can dig that, though it's too bad that we can't import mypy.defaults (and please don't try to find a way to be able to import that -- I think it's important that this script is clearly independent from the mypy package).

So now I agree.

PR #5709, opened 2018-10-01
Add mention of `runtests` script to tests readme

TV4Fun:
Helps with issue discussed in #5688.

### DIFF
--- a/test-data/unit/README.md
+++ b/test-data/unit/README.md
@@ -142,6 +142,10 @@ To run the linter:
 
     $ flake8
 
+You can also run all of the above tests together with:
+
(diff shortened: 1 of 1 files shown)


gvanrossum:
Could you also add a mention of this to the toplevel README.md?

Michael0x2a:
> Could you also add a mention of this to the toplevel README.md?

I think this is being handled by [Ethan's PR](https://github.com/python/mypy/pull/5702/files) -- so if we merge both this PR and Ethan's, both READMEs should be updated.

gvanrossum:
Got it. Thanks!

PR #5722, opened 2018-10-03
Optimize mypy.fastparse (Python 3 only)

JukkaL:
This implements various small optimizations to make AST conversion
from the typed_ast AST faster in compiled mypy. This makes compiled
mypy about 4% faster on self-check, and it also makes non-compiled
mypy around 1% faster on self-check.

Notable included optimizations:
* Implement visit methods more efficiently. Don't inherit visitors from
  `typed_ast` visitor classes.
* Refactor decorated methods away, as they are slow with mypyc.
* Refactor away many non-native attribute lookups.
* Refactor away some nested functions as they are slow with mypyc.

Also removed some dead code.

I'll implement the same optimizations for Python 2 AST conversion once
this PR has been accepted to avoid having to update two PRs based
on review.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -1067,6 +1067,13 @@ class TypeConverter:
     def translate_expr_list(self, l: Sequence[ast3.expr]) -> List[Type]:
         return [self.visit(e) for e in l]
 
+    def visit_raw_str(self, s: str) -> Type:
+        # An escape hatch that allows the AST walker in fastparse2 to
+        # directly hook into the Python 3.5 type converter in some cases
+        # without needing to create an intermediary `Str` object.
+        return (parse_type_comment(s.strip(), self.line, self.errors) or
+                AnyType(TypeOfAny.from_error))
+
     def visit_Call(self, e: Call) -> Type:
         # Parse the arg constructor
         f = e.func
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -636,9 +636,9 @@ class ASTConverter:
         else:
             target_type = None
         s = WithStmt([self.visit(i.context_expr) for i in n.items],
-                        [self.visit(i.optional_vars) for i in n.items],
-                        self.as_required_block(n.body, n.lineno),
-                        target_type)
+                     [self.visit(i.optional_vars) for i in n.items],
+                     self.as_required_block(n.body, n.lineno),
+                     target_type)
         s.is_async = True
         return self.set_line(s, n)
 
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -151,13 +151,6 @@ def with_line(f: Callable[['ASTConverter', T], U]) -> Callable[['ASTConverter', 
     return wrapper
 
 
-def find(f: Callable[[V], bool], seq: Sequence[V]) -> Optional[V]:
-    for item in seq:
-        if f(item):
-            return item
-    return None
-
-
 def is_no_type_check_decorator(expr: ast3.expr) -> bool:
     if isinstance(expr, ast3.Name):
         return expr.id == 'no_type_check'
@@ -203,7 +196,7 @@ class ASTConverter:
         res = []  # type: List[Expression]
         for e in l:
             exp = self.visit(e)
-            isinstance(exp, Expression)
+            assert isinstance(exp, Expression)
             res.append(exp)
         return res
 
@@ -211,7 +204,7 @@ class ASTConverter:
         res = []  # type: List[Statement]
         for e in l:
             stmt = self.visit(e)
-            isinstance(stmt, Statement)
+            assert isinstance(stmt, Statement)
             res.append(stmt)
         return res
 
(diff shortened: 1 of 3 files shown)


mypy/fastparse.py
```diff
@@ -179,33 +173,44 @@ def __init__(self,
...
+        node.column = n.col_offset
+        return node
 
-    def translate_expr_list(self, l: Sequence[ast3.AST]) -> List[Expression]:
+    def translate_expr_list(self, l: Sequence[AST]) -> List[Expression]:
         res = []  # type: List[Expression]
         for e in l:
             exp = self.visit(e)
-            isinstance(exp, Expression)
+            # assert isinstance(exp, Expression)
```
ilevkivskyi: Can this be removed? Anyway `visit` returns `Any`.

mypy/fastparse.py
```diff
@@ -179,33 +173,44 @@ def __init__(self,
...
             res.append(exp)
         return res
 
-    def translate_stmt_list(self, l: Sequence[ast3.AST]) -> List[Statement]:
+    def translate_stmt_list(self, l: Sequence[AST]) -> List[Statement]:
         res = []  # type: List[Statement]
         for e in l:
             stmt = self.visit(e)
-            isinstance(stmt, Statement)
+            # assert isinstance(stmt, Statement)
```
ilevkivskyi: This can probably be removed as well.

mypy/fastparse.py
```diff
@@ -324,23 +329,26 @@ def visit_Module(self, mod: ast3.Module) -> MypyFile:
...
+        return f
 
     def do_func_def(self, n: Union[ast3.FunctionDef, ast3.AsyncFunctionDef],
                     is_coroutine: bool = False) -> Union[FuncDef, Decorator]:
         """Helper shared between visit_FunctionDef and visit_AsyncFunctionDef."""
         no_type_check = bool(n.decorator_list and
                              any(is_no_type_check_decorator(d) for d in n.decorator_list))
 
-        args = self.transform_args(n.args, n.lineno, no_type_check=no_type_check)
+        lineno = n.lineno
```
ilevkivskyi: Why not do the same for `n.col_offset`? (Probably a bit below when it is actually used.)

mypy/fastparse.py
```diff
@@ -769,139 +781,140 @@ def visit_UnaryOp(self, n: ast3.UnaryOp) -> UnaryExpr:
...
-                        cast(List[Optional[str]], [None] * len(n.args)) +
-                        [k.arg for k in n.keywords])
+            [a.value if isinstance(a, Starred) else a for a in args] +
+            [k.value for k in keywords])
+        arg_kinds = ([ARG_STAR if type(a) is Starred else ARG_POS for a in args] +
+                     [ARG_STAR2 if arg is None else ARG_NAMED for arg in keyword_names])
+        e = CallExpr(self.visit(n.func),
+                     arg_types,
+                     arg_kinds,
+                     cast('List[Optional[str]]', [None] * len(args)) + keyword_names)
```
ilevkivskyi: Why the quotes are needed around `List[Optional[str]]` now?

mypy/nodes.py
```diff
@@ -742,7 +742,7 @@ def __init__(self, name: str, type: 'Optional[mypy.types.Type]' = None) -> None:
         super().__init__()
         self._name = name   # Name without module prefix
         # TODO: Should be Optional[str]
-        self._fullname = cast(Bogus[str], None)  # Name with module prefix
+        self._fullname = cast('Bogus[str]', None)  # Name with module prefix
```
ilevkivskyi: Why quotes are now needed here?

mypy/fastparse.py
```diff
@@ -179,33 +173,44 @@ def __init__(self,
...
-        return super().visit(node)
+        typeobj = type(node)
+        visitor = self.visitor_cache.get(typeobj)
+        if visitor is None:
+            method = 'visit_' + node.__class__.__name__
+            visitor = getattr(self, method)
+            self.visitor_cache[typeobj] = visitor
+        return visitor(node)
+
+    def set_line(self, node: N, n: Union[ast3.expr, ast3.stmt]) -> N:
```
ilevkivskyi: Just a random idea, maybe it is possible to set line at the end of `visit()` above and avoid "manually" calling it everywhere?

mypy/fastparse.py
JukkaL: It's only used when generating errors, so the code paths will be rare.

mypy/fastparse.py
JukkaL: They are not needed but they speed things up on older Python versions (when not compiled).

mypy/nodes.py
JukkaL: Again, it's a micro-optimization. Profiling hints at spending a lot of time generating temporary type objects.

mypy/fastparse.py
JukkaL: It's like this because it should only be done for certain node types, not always.

PR #5727, opened 2018-10-03
Remove fscache arg from two functions in modulefinder.py

gvanrossum:
In get_site_packages_dirs() the fscache argument was only used to
check for easy-install.pth, and that check appears to be done only
once.  And compute_search_paths() just passes it to
get_site_packages_dirs().  All this required a few places to create a
throwaway FileSystemCache instance.  It seems simpler to just os
os.path.isfile() and drop the fscache argument.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -334,9 +334,7 @@ class Server:
         t0 = time.time()
         self.update_sources(sources)
         changed, removed = self.find_changed(sources)
-        # TODO: Why create a new FileSystemCache rather than using self.fscache?
-        manager.search_paths = compute_search_paths(
-            sources, manager.options, manager.data_dir, FileSystemCache())
+        manager.search_paths = compute_search_paths(sources, manager.options, manager.data_dir)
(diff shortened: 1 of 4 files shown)


ethanhs:
This seems okay to me, but I think @JukkaL had consistency concerns when I used `os.path.*` in some other places.

mypy/dmypy_server.py
```diff
@@ -334,9 +334,7 @@ def fine_grained_increment(self, sources: List[BuildSource]) -> Dict[str, Any]:
         t0 = time.time()
         self.update_sources(sources)
         changed, removed = self.find_changed(sources)
-        # TODO: Why create a new FileSystemCache rather than using self.fscache?
-        manager.search_paths = compute_search_paths(
-            sources, manager.options, manager.data_dir, FileSystemCache())
+        manager.search_paths = compute_search_paths(sources, manager.options, manager.data_dir)
```
gvanrossum: @JukkaL This location drew my attention (since there's also `self.fscache`).

PR #5733, opened 2018-10-04
 Document ClassVar

### LINKED ISSUE #5546: Document ClassVar
We need a section clarifying what exactly is the semantics of `ClassVar`, what is (and is not) allowed. There were some situations where people got confused by this.

nanjekyejoannah:
This PR fixes #5546.

I have added a section in the class basics documentation with details on when, how and when not to use ClassVar.

cc @ilevkivskyi

### DIFF
--- a/docs/source/class_basics.rst
+++ b/docs/source/class_basics.rst
@@ -38,7 +38,7 @@ a type annotation:
    a = A()
    a.x = [1]     # OK
 
-As in Python generally, a variable defined in the class body can used
+As in Python generally, a variable defined in the class body can be used
 as a class or an instance variable.
 
 Type comments work as well, if you need to support Python versions earlier
@@ -77,13 +77,17 @@ to it explicitly using ``self``:
            a = self
            a.x = 1      # Error: 'x' not defined
 
-Class and instance variable annotations
-***************************************
+Class attribute annotations
+***************************
 
 Mypy supports annotations for class and instance 
 variables in class bodies and methods. Use ``ClassVar`` to
 indicate to the static type checker that this variable 
 should not be set on instances. 
+
+A class attribute without the ``ClassVar`` annotation can be used as 
+a class variable. Mypy won't prevent it from being used as an 
+instance variable.
 
 .. code-block:: python
 
@@ -91,14 +95,15 @@ should not be set on instances.
       y: ClassVar[Dict[str, int]] = {}  # class variable
       z: int = 10                       # instance variable
 
+The following are worth noting about ``ClassVar``:
 
-``ClassVar`` accepts only types and cannot be further subscribed.
+- It accepts only types and cannot be further subscribed.
 
-``ClassVar`` is not a class itself, and should not be used with 
+- It is not a class itself, and should not be used with 
 isinstance() or issubclass(). 
 
-``ClassVar`` does not change Python runtime behavior, but 
-it can be used by third-party type checkers. For example, a type checker 
+- It does not change Python runtime behavior, but it can 
+be used by third-party type checkers. For example, a type checker 
 might flag the following code as an error:
 
 .. code-block:: python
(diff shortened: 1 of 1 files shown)


ilevkivskyi:
I think @JukkaL might want to review this.

docs/source/class_basics.rst
```diff
@@ -77,6 +77,53 @@ to it explicitly using ``self``:
...
+  a = A(3000)
+  a.y = {}                # Error, setting class variable on instance
+  a.z = {}                # This is OK
+
+.. note::
+   Note that a ClassVar parameter cannot include any type variables, 
+   regardless of the level of nesting: ``ClassVar[T]`` and ``ClassVar[List[Set[T]]]``
+   are both invalid if ``T`` is a type variable.
+
+For convention, you can annotate instance variables in __init__ or other methods,
```
JukkaL: This part (and the example below) are basically a repeat of parts of section "Instance and class attributes". Note that `y: T` is a valid definition in a class body -- it's only `y: ClassVar[T]` that's rejected.

docs/source/class_basics.rst
```diff
@@ -77,6 +77,53 @@ to it explicitly using ``self``:
...
+      z: int = 10                       # instance variable
+
+
+``ClassVar`` accepts only types and cannot be further subscribed.
+
+``ClassVar`` is not a class itself, and should not be used with 
+isinstance() or issubclass(). 
+
+``ClassVar`` does not change Python runtime behavior, but 
+it can be used by third-party type checkers. For example, a type checker
```
JukkaL: This section is kind of vague. It would be better to explain what mypy actually does rather than what it might do :-)

docs/source/class_basics.rst
```diff
@@ -77,6 +77,53 @@ to it explicitly using ``self``:
            a = self
            a.x = 1      # Error: 'x' not defined
 
+Class and instance variable annotations
+***************************************
+
+Mypy supports annotations for class and instance
```
JukkaL: This section doesn't flow well from the previous section due to different terminology. In the last section we talk about class and instance attributes and here we talk about class and instance variables, and here "class variable" means something different from "class attribute" in the previous section. Not sure what's the best way forward, but the two section should form a coherent discussion.

docs/source/class_basics.rst
```diff
@@ -77,6 +77,53 @@ to it explicitly using ``self``:
            a = self
            a.x = 1      # Error: 'x' not defined
 
+Class and instance variable annotations
+***************************************
+
+Mypy supports annotations for class and instance 
+variables in class bodies and methods. Use ``ClassVar`` to
+indicate to the static type checker that this variable
```
JukkaL: This a small nit, but the convention in the docs is say "the type checker" rather than "the static type checker".

docs/source/class_basics.rst
```diff
@@ -77,6 +77,53 @@ to it explicitly using ``self``:
            a = self
            a.x = 1      # Error: 'x' not defined
 
+Class and instance variable annotations
+***************************************
+
+Mypy supports annotations for class and instance 
+variables in class bodies and methods. Use ``ClassVar`` to
+indicate to the static type checker that this variable 
+should not be set on instances.
```
JukkaL: Maybe contrast this to a class attribute/variable without a ``ClassVar`` annotation: it can used as a class variable, but mypy won't prevent it from being used as an instance variable as well, potentially leading to ambiguity.

docs/source/class_basics.rst
```diff
@@ -77,6 +77,53 @@ to it explicitly using ``self``:
...
+should not be set on instances. 
+
+.. code-block:: python
+
+  class A:
+      y: ClassVar[Dict[str, int]] = {}  # class variable
+      z: int = 10                       # instance variable
+
+
+``ClassVar`` accepts only types and cannot be further subscribed.
```
JukkaL: Maybe instead of having very short paragraphs introduce the relevant rules in a bullet list. It's worth mentioning that `y: ClassVar =0` is also valid (without square brackets) and the type of the variable will become implicitly `Any`. Also there should be a note that this behavior may change in the future.

gvanrossum:
Hey @nanjekyejoannah, do you want to push an updated version in response to Jukka's review? There's still time to get this cherry-picked in the [upcoming 0.640 release](https://github.com/python/mypy/issues/5741).

PR #5734, opened 2018-10-04
Increase the GC threshold 3x

JukkaL:
This seems to speed up compiled mypy by up to about 4%, with marginal
changes to memory use.


mypy/build.py
```diff
@@ -173,7 +173,7 @@ def _build(sources: List[BuildSource],
            fscache: Optional[FileSystemCache],
            ) -> BuildResult:
     # This seems the most reasonable place to tune garbage collection.
-    gc.set_threshold(50000)
+    gc.set_threshold(150 * 1000)
```
ilevkivskyi: If we would support only 3.6+ we could write `150_000` :-)

PR #5735, opened 2018-10-04
Optimize mypy.fastparse2

JukkaL:
These seem to speed up Python 2 type checking runtimes by up to
around 3%, when compiled with mypyc.

This is similar to #5722, but for Python 2. Also added some parser
tests, since Python 2 coverage was spotty. (It's still not quite at full
coverage, though.)

### DIFF
--- a/mypy/fastparse2.py
+++ b/mypy/fastparse2.py
@@ -414,38 +414,11 @@ class ASTConverter:
         converter = TypeConverter(self.errors, line=line)
         decompose_stmts = []  # type: List[Statement]
 
-        def extract_names(arg: ast27.expr) -> List[str]:
-            if isinstance(arg, ast27.Name):
-                return [arg.id]
-            elif isinstance(arg, ast27.Tuple):
-                return [name for elt in arg.elts for name in extract_names(elt)]
-            else:
-                return []
-
-        def convert_arg(index: int, arg: ast27.expr) -> Var:
-            if isinstance(arg, ast27.Name):
-                v = arg.id
-            elif isinstance(arg, ast27.Tuple):
-                v = '__tuple_arg_{}'.format(index + 1)
-                rvalue = NameExpr(v)
-                rvalue.set_line(line)
-                assignment = AssignmentStmt([self.visit(arg)], rvalue)
-                assignment.set_line(line)
-                decompose_stmts.append(assignment)
-            else:
-                raise RuntimeError("'{}' is not a valid argument.".format(ast27.dump(arg)))
-            return Var(v)
-
-        def get_type(i: int) -> Optional[Type]:
-            if i < len(type_comments):
-                comment = type_comments[i]
-                if comment is not None:
-                    return converter.visit_raw_str(comment)
-            return None
-
-        args = [(convert_arg(i, arg), get_type(i)) for i, arg in enumerate(n.args)]
+        args = [(self.convert_arg(i, arg, line, decompose_stmts),
+                 self.get_type(i, type_comments, converter))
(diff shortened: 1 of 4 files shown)


mypy/fastparse2.py
```diff
@@ -872,68 +891,72 @@ def visit_Num(self, new: ast27.Num) -> Expression:
...
         # Hack: assume all string literals in Python 2 stubs are normal
         # strs (i.e. not unicode).  All stubs are parsed with the Python 3
         # parser, which causes unprefixed string literals to be interpreted
         # as unicode instead of bytes.  This hack is generally okay,
         # because mypy considers str literals to be compatible with
         # unicode.
-        if isinstance(s.s, bytes):
-            n = s.s
+        if isinstance(n.s, bytes):
+            s = n.s
```
ilevkivskyi: I would use a longer name here instead of `s`, maybe `string`. Also `s` used in above methods for statements.

mypy/fastparse2.py
```diff
@@ -872,68 +891,72 @@ def visit_Num(self, new: ast27.Num) -> Expression:
...
+    def visit_Attribute(self, n: Attribute) -> Expression:
+        member_expr = MemberExpr(self.visit(n.value), n.attr)
+        obj = member_expr.expr
+        if (isinstance(obj, CallExpr) and
+                isinstance(obj.callee, NameExpr) and
+                obj.callee.name == 'super'):
+            e = SuperExpr(member_expr.name, obj)  # type: Expression
+        else:
+            e = member_expr
+        return self.set_line(e, n)
```
ilevkivskyi: I am curious, does this (using `isinstance()` with native classes in this methid) give a visible perf win?

Also maybe it is worth adding a comment, otherwise it is not clear why we always create a `MemberExpr` instead of implementing this as it was before.

mypy/fastparse2.py
```diff
@@ -469,10 +456,41 @@ def fail_arg(msg: str, arg: None) -> None:
...
+            return [name for elt in arg.elts for name in self.extract_names(elt)]
+        else:
+            return []
+
+    def convert_arg(self, index: int, arg: ast27.expr, line: int,
+                    decompose_stmts: List[Statement]) -> Var:
+        if isinstance(arg, Name):
+            v = arg.id
+        elif isinstance(arg, ast27_Tuple):
+            v = '__tuple_arg_{}'.format(index + 1)
```
ilevkivskyi: Should we add a test for this? Or do we already have one?

test-data/unit/parse-python2.test
```diff
@@ -395,3 +395,416 @@ MypyFile:1(
...
+      Var(g)
+      NameExpr(overload)
+      FuncDef:5(
+        g
+        Block:5(
+          PassStmt:6())))
+    FuncDef:7(
+      g
+      Block:7(
+        PassStmt:8()))))
```
ilevkivskyi: Thanks for adding all these tests!

mypy/fastparse2.py
JukkaL: We already had tests for Python 2 specific features such as this.

mypy/fastparse2.py
JukkaL: I didn't measure the impact of each individual change, but this should be slightly faster since super expressions are much less common than member expressions, and the check for a super expression will be faster in the new implementation since we can do native attribute access (`obj.callee` and such). The impact of this particular change may not be measurable, but the overall reduction in Python attribute accesses gave a measurable aggregate performance increase in the Python 3 AST transformation case at least. Will add a comment.

PR #5739, opened 2018-10-04
Fix infinite loop on corrupted fine-grained cache

msullivan:
This hit me when mypyc was causing crashes after having written out
part of the cache, but subsequent runs would instead infinite loop.


gvanrossum:
I think I may have hit this yesterday. I did something with dmypy like the following:

1. dmypy check mypy
2. dmypy check ../mypyc/mypyc

The daemon was started with `-v` logging to a separate window. It never completed (until I hit `^C`).

gvanrossum:
Should probably be merged before I cut the release branch (#5741).

msullivan:
Turns out that we already had a test case for the case where the cache file was *wrong*, but we only infinite loop when it is *missing*.

Issue #3195, opened 2017-04-19
Let's drop --strict-boolean


gvanrossum:
See discussion in #3185 -- it's not working perfectly and it is encouraging bad style. It also makes `--strict` less useful: it means you can't use `--strict` if you don't believe in `--strict-optional`, else you'd get a ton of false positive from the latter.

gvanrossum:
Or at least start deprecating it and remove it from `--strict`.

pkch:
Before we proceed with this, I just wanted to clear up one thing based on the earlier discussion on gitter.

Would this flag become useful if it was restricted just to the cases where it's actually unsafe? Specifically the usage of non-boolean variable in a boolean context is unsafe only if it has more than 1 values that is interpreted as `False`:

```
def f(x: int) -> None:
  if x:  ... # safe
 
def g(x: int = None) -> None:
  if x:  ... # unsafe; is x 0 or None?
 
def h(x: set) -> None:
  if x: ... # safe

def i(x: set = None) -> None:
  if x: ... # unsafe; is x an empty set or None?

def j(x: Callable[[int], int] = None) -> None:
  if x: ... # safe

def k(x: bool = None) -> None:
  if x: ... # unsafe; is x False or None?

It seems very easy to identify unsafe cases, using the [truth value testing](https://docs.python.org/3/library/stdtypes.html#truth-value-testing) rules. The usage is unsafe if the variable type is a `Union` that includes more than one of the following types:

- `None`
- a numeric type
- a subtype of Sized
- a type that has `__bool__` method defined

Do you want me to check if restricting `--strict-boolean` to just these cases would make sense, perhaps using mypy own source code?

JukkaL:
Agreed -- it's not worth having `--strict` enable it, and the entire option is questionable.

@pkch We can discuss that as a separate feature (which might perhaps share a name with the current option). Feel free to open an issue for it.

JelleZijlstra:
When `--strict-boolean` was added I tried it on my codebase and found it way too pedantic to be useful. Maybe some people really like writing `if len(lst) > 0:` instead of `if lst:`, but I don't think mypy needs to enforce that.

However, there are some related checks that mypy could do that are actually useful, like warning if you do `if some_object.is_foo` instead of `if some_object.is_foo()` (where `is_foo` is a method). But those kinds of things can be discussed in a separate issue.

gvanrossum:
[...]

PR #5740, opened 2018-10-05
Remove deprecated options

gvanrossum:
These will now all fail with `unrecognized arguments: <flag>`
```
--disallow-any (message points to options replacing it)
-f/--dirty-stubs (ignored)
--use-python-path (exited with error message)
-s/--silent-imports (set ignore_missing_imports=True and follow_imports=skip)
--almost-silent (ses follow_imports=error)
--[no-]fast-parser (ignored)
--strict-boolean (actually did something, though questionable)
```
I've put the removal of `--strict-boolean` in a separate commit since it's more controversial -- some people might still be using it, even though it's been deprecated since May 2017.  See discussion in https://github.com/python/mypy/issues/3195.


Michael0x2a:
Regarding the strict-boolean flag -- maybe one idea might be to move that sort of logic into a separate plugin that's independent from mypy.

I've been vaguely thinking about doing this myself, actually -- I can never remember what's supposed to be truthy or falsey from language to language, so most of my personal code is written in a style where this sort of plugin would actually be mildly useful/not very noisy. I'm not sure if our plugin system would support these kinds of extensions yet though.

gvanrossum:
I don't think the plugin system could do that unless there was a way to
request being called for every expression.

ethanhs:
Hm, it seems I did the wrong thing trying to make flake8 happy. Sorry!

gvanrossum:
No worries, I'll fix it.

PR #5745, opened 2018-10-06
Add --add/--remove/--update flags to 'dmypy recheck'

gvanrossum:
Using `dmypy recheck --add <files> --remove <files> --update <files>` will take the previous list of files, apply adds/removes/updates, and then check the resulting list of sources, *without* calling `stat()` on any of the unchanged files. (Useful if the caller is using e.g. Watchman.) Each list of `<files>` may be empty in case there are no files in that category.

For backwards compatibility, if none of `--add`, `--remove` or `--update` is used, `dmypy recheck` will revert to its old behavior (i.e. re-check the exact same list of files, calling `stat()` on each file to determine whether it has changed). Note that even if the list of `<files>` is empty, the presence of the new flags triggers the new behavior.

There are also some cleanups in the diff. (And some loose ends that I will try to address next week.)

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -19,6 +19,7 @@ from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Tuple
 
 import mypy.build
 import mypy.errors
+import mypy.main
 from mypy.find_sources import create_source_list, InvalidSourceList
 from mypy.server.update import FineGrainedBuildManager
 from mypy.dmypy_util import STATUS_FILE, receive
@@ -92,7 +93,6 @@ SOCKET_NAME = 'dmypy.sock'  # type: Final
 
 
 def process_start_options(flags: List[str], allow_sources: bool) -> Options:
-    import mypy.main
     sources, options = mypy.main.process_options(['-i'] + flags,
                                                  require_targets=False,
                                                  server_options=True)
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -1,6 +1,6 @@
 """Server for mypy daemon mode.
 
-Highly experimental!  Only supports UNIX-like systems.
+Only supports UNIX-like systems.
 
 This implements a daemon process which keeps useful state in memory
 to enable fine-grained incremental reprocessing of changes.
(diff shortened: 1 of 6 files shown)


mypy/dmypy_server.py
```diff
@@ -257,11 +262,36 @@ def cmd_check(self, files: Sequence[str]) -> Dict[str, object]:
...
+        """Check the same list of files we checked most recently.
+
+
+        If add/remove/update is given, they modifies the previous list.
+        """
+        t0 = time.time()
         if not self.fine_grained_manager:
             return {'error': "Command 'recheck' is only valid after a 'check' command"}
-        return self.check(self.previous_sources)
+        print(f"\nadd={add}, remove={remove}, update={update}")
```
gvanrossum: This `print()` still needs to be removed.

gvanrossum:
I don’t believe this is right though. It ignores typeshed.

mypy/fswatcher.py
```diff
@@ -85,3 +80,24 @@ def find_changed(self) -> Set[str]:
...
+
+        This only calls self.fs.stat() on added or updated files, not
+        on all files.  It believes all other files are unchanged!
+
+        Implies add_watched_paths() for add and update, and
+        remove_watched_paths() for remove.
+        """
+        self.remove_watched_paths(remove)
+        self.add_watched_paths(add)
+        return self._find_changed(add + update)
```
ilevkivskyi: I didn't do a careful review, but generally this looks good. I just have one question to clarify (and it is probably worth mentioning in the docstring): Is it OK to pass files to `--remove` that were never in the build?

My motivation is to avoid duplication of work in the daemon and in the runner script that filters sources (what we do internally). If only files that were previously in the build are allowed in `--remove`, then we will need to keep full list of the checked files also in the runner script (because there is no way to check whether already deleted file should be filtered or not, if the filtering is content based, like presence of type comments/annotations).

I would prefer therefore to allow passing arbitrary files to `--remove` and explicitly document this.

ilevkivskyi: To clarify, from the current implementation in this PR it looks like it should work, I just would like to make this explicit.

mypy/dmypy_server.py
```diff
@@ -249,21 +252,45 @@ def cmd_run(self, version: str, args: Sequence[str]) -> Dict[str, object]:
...
+        If add/remove/update is given, they modifies the previous list.
+        """
+        t0 = time.time()
+        if not self.fine_grained_manager:
             return {'error': "Command 'recheck' is only valid after a 'check' command"}
-        return self.check(self.last_sources)
+        sources = self.previous_sources
+        if remove:
+            removals = set(remove)
+            print
```
msullivan: Stray `print`

Issue #5741, opened 2018-10-05
Release 0.640 planning


gvanrossum:
I am the release manager for mypy 0.640. I plan to cut the release branch right now (Oct 5) and do the release in a week (Oct 12).

If there are PRs that are low-risk and high-value that ought to go into this release please mention them here and I'll see if they can be cherry-picked.

ilevkivskyi:
I think we should decide what to do about https://github.com/python/mypy/issues/5738, because technically it is a regression. Possible options are:
* Don't care about this for now, it is minor and is caused by an old bug exposed by a fix to another issue.
* Implement a simple fix (unfortunately I don't have one yet, sorry, but it should not be hard) just to avoid the false positive.
* Revert the switch from all to invariant instance contexts.

gvanrossum:
Release branch created: https://github.com/python/mypy/tree/release-0.640

It's currently identical to master -- the last commit is "Fix infinite loop on corrupted fine-grained cache (#5739)".

Anything that we decide should go into this release will be cherry-picked.

gvanrossum:
I went through all of the motions of the release, but within minutes a regression was found (#5784) so I had to withdraw the release. There will be a new one Monday.

msullivan:
0.641 got released a while ago

PR #5749, opened 2018-10-08
Add docs for --namespace-packages

### LINKED ISSUE #5757: Documentation should be updated to for new namespace pkg support
https://github.com/python/mypy/pull/5691 and https://github.com/python/mypy/pull/5591 added support for importing type hints from namespace packages according to pep 561 but the documentation hasn't been updated to reflect it https://mypy.readthedocs.io/en/latest/installed_packages.html#using-pep-561-compatible-packages-with-mypy

Specifically :
> Note that mypy does not support some more advanced import features, such as zip imports, namespace packages, and custom import hooks.

I suggest simply updating the docs to:
> Note that mypy does not support some more advanced import features, such as zip imports, ~~namespace packages,~~ and custom import hooks.

gvanrossum:
Follow-up for #5691. Should be cherry-picked into the release (#5741). Fixes #5757.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -72,6 +72,24 @@ Import discovery
 The following flags customize how exactly mypy discovers and follows
 imports.
 
+``--namespace-packages``
+    This flag enables import discovery to use namespace packages (see
+    `PEP 420`_).  In particular, this allows discovery of imported
+    packages that don't have an ``__init__.py`` (or ``__init__.pyi``)
+    file.
+
+    Namespace packages are found (using the PEP 420 rules, which
+    prefers "classic" packages over namespace packages) along the
+    module search path -- this is primarily set from the source files
+    passed on the command line, the ``MYPYPATH`` environment variable,
+    and the :ref:`mypy_path config option
+    <config-file-import-discovery-global>`.
+
+    Note that this only affects import discovery -- for modules and
+    packages explicitly passed on the command line, mypy still
+    searches for ``__init__.py[i]`` files in order to determine the
+    fully-qualified module/package name.
+
 ``--ignore-missing-imports``
     This flag makes mypy ignore all missing imports. It is equivalent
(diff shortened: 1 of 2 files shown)


docs/source/command_line.rst
```diff
@@ -72,6 +72,24 @@ Import discovery
...
+    prefers "classic" packages over namespace packages) along the
+    module search path -- this is primarily set from the source files
+    passed on the command line, the ``MYPYPATH`` environment variable,
+    and the :ref:`mypy_path config option
+    <config-file-import-discovery-global>`.
+
+    Note that this only affects import discovery -- for modules and
+    packages explicitly passed on the command line, mypy still
+    searches for ``__init__.py[i]`` files in order to determine the
+    fully-qualified module/package name.
```
msullivan: Maybe it would be worth talking about what to do in order to explicitly check code in namespace packages?

JukkaL: What about things like `mypy -m foo.bar` if `foo` is a namespace package? Users may expect  it to work, since the module name is explicit. Similarly, what about `mypy -p foo` if `foo` is a namespace package?

gvanrossum: Hm, none of those currently work. Maybe they should, but I don't have time to make it work in this release (command line files/modules/packages use an entirely different code path than imported modules). I'll open a new issue.

gvanrossum:
Follow-up for #5691. Should be cherry-picked into the release (#5741). Fixes #5757.

docs/source/command_line.rst
msullivan: `mypy -c 'import foo.bar'`, I suppose?

gvanrossum: Hm, there are still problems with that. I'm not sure why. Will try to look into this before the release goes out.

gvanrossum: Maybe #5758?

Issue #5598, opened 2018-09-11
Document final names, methods, and classes.


ilevkivskyi:
The latest WIP version is in https://github.com/python/mypy/pull/5522/commits/66fdbc835c3890201fa21281c24a580c841a1fdc

ilevkivskyi:
This is a follow-up for https://github.com/python/mypy/pull/5522

ilevkivskyi:
I think [docs for final things](https://mypy.readthedocs.io/en/latest/final_attrs.html) are in good shape now, so this can be closed.

Issue #1021, opened 2015-11-29
Infer attributes from __new__


JukkaL:
In this example (from #982) we define an attribute via assignment in `__new__`:
 
```
class C(object):
    def __new__(cls, foo=None):
        obj = object.__new__(cls)
        obj.foo = foo
        return obj

x = C(foo=12)
print(x.foo)
```

Currently mypy doesn't recognize attribute `foo` and complains about `x.foo`. To implement this, we could do these things:

* Infer the fact that `obj` is something similar to `self` because of the way is created via `object.__new__`. This should happen during semantic analysis so that this will work in unannotated method as well.
* Infer attributes from assignments via `obj` because it is classified as similar to `self`.
* Also recognize `object.__new__` calls via `super(...)`.

JukkaL:
This is follow-up to #982.

JukkaL:
Decreased priority since this doesn't seem like users often struggle with this -- it's easy enough to work around by inserting type annotations.

toolforger:
One use case is NamedTuple. It's turning mypy into a noise generator for me.
Not sure how this data point affects priorization.

JelleZijlstra:
@toolforger how are you using NamedTuple? I don't think the example from the original post in this issue is applicable to NamedTuple.

toolforger:
Using a NamedTuple, subclass it with a `__new__` override to provide standard parameters.
Details in #5194, which is a duplicate of #1279.

ilevkivskyi:
> Using a NamedTuple subclass, subclass it with a __new__ override to provide standard parameters.

Use instead:
```python
class Foo(NamedTuple):
    x: int
    y: int = 0

Foo(1)  # OK, same as Foo(1, 0)
Foo(1, 2)  # OK
```
works on Python 3.6+. If you are on Python 2, then wait for this issue to be solved.

toolforger:
@ilevkivskyi your advice works for direct subclasses. My use case is for an indirect subclass. (I already had to move all fields up into the root class because NamedTuples don't properly support subclassing, but I cannot use frozen dataclasses until 3.7 comes out for my distro. I'm on the verge of ripping out mypy again, or switching from Python to Kotlin.)

gvanrossum:
@toolforger I hear your frustration, but please don't take it out on us. We have a lot on our plate and we're only a small team (half of us are volunteers). This will eventually get fixed. (If you want to help by submitting a PR that might speed it up of course.)

[...]

PR #5754, opened 2018-10-09
Draft documentation for final attributes

ilevkivskyi:
Draft work towards #5598

@JukkaL you can either push directly to this branch or merge this and polish it afterwards.

### DIFF
--- a/docs/source/final_attrs.rst
+++ b/docs/source/final_attrs.rst
@@ -1,0 +1,267 @@
+Final names, methods and classes
+================================
+
+You can declare a variable or attribute as final, which means that the variable
+must not be assigned a new value after initialization. This is often useful for
+module and class level constants as a way to prevent unintended modification.
+Mypy will prevent further assignments to final names in type-checked code:
+
+.. code-block:: python
+
+   from typing_extensions import Final
+
+   RATE: Final = 3000
+   class Base:
+       DEFAULT_ID: Final = 0
+
+   # 1000 lines later
+
+   RATE = 300  # Error: can't assign to final attribute
+   Base.DEFAULT_ID = 1  # Error: can't override a final attribute
+
+Another use case for final attributes is where a user wants to protect certain
+instance attributes from overriding in a subclass:
+
+.. code-block:: python
+
+   import uuid
+   from typing_extensions import Final
+
+   class Snowflake:
+       """An absolutely unique object in the database"""
+       def __init__(self) -> None:
+           self.id: Final = uuid.uuid4()
+
+   # 1000 lines later
+
+   class User(Snowflake):
+       id = uuid.uuid4()  # Error: can't override a final attribute
+
+Some other use cases might be solved by using ``@property``, but note that
+neither of the above use cases can be solved with it.
+
+.. note::
+
+   This is an experimental feature. Some details might change in later
+   versions of mypy. The final qualifiers are available in the
+   ``typing_extensions`` package available on PyPI.
+
+Syntax variants
+***************
+
+The ``typing_extensions.Final`` qualifier indicates that a given name or
+attribute should never be re-assigned, re-defined, nor overridden. It can be
+used in one of these forms:
+
+
+* You can provide an explicit type using the syntax ``Final[<type>]``. Example:
+
+  .. code-block:: python
+
+     ID: Final[float] = 1
+
+* You can omit the type: ``ID: Final = 1``. Note that unlike for generic
+  classes this is *not* the same as ``Final[Any]``. Here mypy will infer
+  type ``int``.
+
+* In stub files you can omit the right hand side and just write
+  ``ID: Final[float]``.
+
+* Finally, you can define ``self.id: Final = 1`` (also with a type argument),
+  but this is allowed *only* in ``__init__`` methods.
+
+Definition rules
+****************
+
+The are two rules that should be always followed when defining a final name:
+
+* There can be *at most one* final declaration per module or class for
+  a given attribute:
+
+  .. code-block:: python
+
+     from typing_extensions import Final
+
+     ID: Final = 1
+     ID: Final = 2  # Error: "ID" already declared as final
+
+     class SomeCls:
+         id: Final = 1
+         def __init__(self, x: int) -> None:
+             self.id: Final = x  # Error: "id" already declared in class body
+
+  Note that mypy has a single namespace for a class. So there can't be two
+  class-level and instance-level constants with the same name.
+
+* There must be *exactly one* assignment to a final attribute:
+
+  .. code-block:: python
+
+     ID = 1
+     ID: Final = 2  # Error!
+
+     class SomeCls:
+         ID = 1
+         ID: Final = 2  # Error!
+
+* A final attribute declared in class body without an initializer must
+  be initialized in the ``__init__`` method (you can skip the initializer
+  in stub files):
+
+  .. code-block:: python
+
+     class SomeCls:
+         x: Final[int]
+         y: Final[int]  # Error: final attribute without an initializer
+         def __init__(self) -> None:
+             self.x = 1  # Good
+
+* ``Final`` can be only used as an outermost type in assignments or variable
+  annotations. using it in any other position is an error. In particular,
+  ``Final`` can't be used in annotations for function arguments:
+
+  .. code-block:: python
+
+     x: List[Final[int]] = []  # Error!
+     def fun(x: Final[List[int]]) ->  None:  # Error!
+         ...
+
+* ``Final`` and ``ClassVar`` should not be used together. Mypy will infer
+  the scope of a final declaration automatically depending on whether it was
+  initialized in the class body or in ``__init__``.
+
+Using final attributes
+**********************
+
+As a result of a final declaration mypy strives to provide the
+two following guarantees:
+
+* A final attribute can't be re-assigned (or otherwise re-defined), both
+  internally and externally:
+
+  .. code-block:: python
+
+     # file mod.py
+     from typing_extensions import Final
+
+     ID: Final = 1
+
+     class SomeCls:
+         ID: Final = 1
+
+         def meth(self) -> None:
+             self.ID = 2  # Error: can't assign to final attribute
+
+     # file main.py
+     import mod
+     mod.ID = 2  # Error: can't assign to constant.
+
+     from mod import ID
+     ID = 2  # Also an error, see note below.
+
+     class DerivedCls(mod.SomeCls):
+         ...
+
+     DerivedCls.ID = 2  # Error!
+     obj: DerivedCls
+     obj.ID = 2  # Error!
+
+* A final attribute can't be overridden by a subclass (even with another
+  explicit final declaration). Note however, that final attributes can
+  override read-only properties. This also applies to multiple inheritance:
+
+  .. code-block:: python
+
+     class Base:
+         @property
+         def ID(self) -> int: ...
+
+     class One(Base):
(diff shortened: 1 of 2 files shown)


gvanrossum:
Do you want this cherry-picked in the 0.640 release branch?

JukkaL:
I think that we can cherry-pick this if we have a non-draft version before the release. I'll finish this up tomorrow.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+
+   import uuid
+   from typing_extensions import Final
+
+   class Snowflake:
+       """An absolutely unique object in the database"""
+       def __init__(self) -> None:
+           self.id: Final = uuid.uuid4()
+
+   # 1000 lines later
```
gvanrossum: This comment feels irrelevant.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+   class User(Snowflake):
+       id = uuid.uuid4()  # Error: can't override a final attribute
+
+Some other use cases might be solved by using ``@property``, but note that
+neither of the above use cases can be solved with it.
+
+.. note::
+
+   This is an experimental feature. Some details might change in later
+   versions of mypy. The final qualifiers are available in the
```
gvanrossum: Maybe instead of "the final qualifiers" write "the `final` decorator and `Final` pseudo-type"?

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+   ``typing_extensions`` package available on PyPI.
+
+Syntax variants
+***************
+
+The ``typing_extensions.Final`` qualifier indicates that a given name or
```
gvanrossum: So I'm not sure that "qualifier" is the best term here, though I'm not sure that "pseudo-type" is any better. We probably should use the same word to describe "ClassVar" (see #5733).

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+attribute should never be re-assigned, re-defined, nor overridden. It can be
+used in one of these forms:
+
+
```
gvanrossum: Only one blank line.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+* You can omit the type: ``ID: Final = 1``. Note that unlike for generic
+  classes this is *not* the same as ``Final[Any]``. Here mypy will infer
+  type ``int``.
+
+* In stub files you can omit the right hand side and just write
+  ``ID: Final[float]``.
+
+* Finally, you can define ``self.id: Final = 1`` (also with a type argument),
+  but this is allowed *only* in ``__init__`` methods.
```
gvanrossum: Why the limitation to `__init__` methods? In general we allow inferring or declaring the type of instance variables in any method (though IIRC a declaration must be on the *first* use). Is this because the implementation cannot ensure that there aren't any assignments in other places if `Final` isn't set in `__init__`?

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
+Final names, methods and classes
+================================
```
gvanrossum: I would call out somewhat early in this section that there's both `Final` (a type qualifier or pseudo-type) and `final` (a decorator).

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+
+     ID: Final = 1
+     ID: Final = 2  # Error: "ID" already declared as final
+
+     class SomeCls:
+         id: Final = 1
+         def __init__(self, x: int) -> None:
+             self.id: Final = x  # Error: "id" already declared in class body
+
+  Note that mypy has a single namespace for a class. So there can't be two
```
gvanrossum: Drop "two".

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+     ID = 1
+     ID: Final = 2  # Error!
+
+     class SomeCls:
+         ID = 1
+         ID: Final = 2  # Error!
+
+* A final attribute declared in class body without an initializer must
+  be initialized in the ``__init__`` method (you can skip the initializer
+  in stub files):
```
gvanrossum: Again, why limit this to `__init__`?

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+
+     class SomeCls:
+         x: Final[int]
+         y: Final[int]  # Error: final attribute without an initializer
+         def __init__(self) -> None:
+             self.x = 1  # Good
+
+* ``Final`` can be only used as an outermost type in assignments or variable
+  annotations. using it in any other position is an error. In particular,
+  ``Final`` can't be used in annotations for function arguments:
```
gvanrossum: This is also very similar to `ClassVar`, right?

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+
+  .. code-block:: python
+
+     x: List[Final[int]] = []  # Error!
+     def fun(x: Final[List[int]]) ->  None:  # Error!
+         ...
+
+* ``Final`` and ``ClassVar`` should not be used together. Mypy will infer
+  the scope of a final declaration automatically depending on whether it was
+  initialized in the class body or in ``__init__``.
```
gvanrossum: Hm, interesting. I can understand you don't want `ClassVar[Final]` or  `Final[ClassVar]` (though the latter kind of sounds all right :-).

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+* There must be *exactly one* assignment to a final attribute:
+
+  .. code-block:: python
+
+     ID = 1
+     ID: Final = 2  # Error!
+
+     class SomeCls:
+         ID = 1
+         ID: Final = 2  # Error!
```
gvanrossum: Can you use a different name instead of `ID` for the class variable? Having the name the same distract from the point this example is making regarding duplication. I don't think you need an example showing that a final module global variable and a final class variable don't clash -- it's standard Python semantics (and common knowledge) that those namespaces don't interact.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+     # file mod.py
+     from typing_extensions import Final
+
+     ID: Final = 1
+
+     class SomeCls:
+         ID: Final = 1
+
+         def meth(self) -> None:
+             self.ID = 2  # Error: can't assign to final attribute
```
gvanrossum: Again I think it would be less confusing if the module variable and the class variable had different names.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+         ID: Final = 1  # OK
+
+     class Other(Base):
+         ID: Final = 2  # OK
+
+     class Combo(One, Other):  # Error: cannot override final attribute.
+         pass
+
+* Declaring a name as final only guarantees that the name wll not be re-bound
+  to other value, it doesn't make the value immutable. One can use immutable ABCs
```
gvanrossum: other -> another

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+  and containers to prevent mutating such values:
+
+  .. code-block:: python
+
+     x: Final = ['a', 'b']
+     x.append('c')  # OK
+
+     y: Final[Sequance[str]] = ['a', 'b']
```
gvanrossum: Typo (Sequ*e*nce)

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+Final methods
+*************
+
+Like with attributes, sometimes it is useful to protect a method from
+overriding. In such situations one can use the ``typing_extensions.final``
+decorator:
+
+.. code-block:: python
+
+   from typing_extensions import final
```
gvanrossum: Add a comment emphasizing this uses a lower-case 'f'.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+
+   class Base:
+       @final
+       def common_name(self) -> None:
+           ...
+
+   # 1000 lines later
```
gvanrossum: Drop.

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+  .. code-block:: python
+
+     x: Final = ['a', 'b']
+     x.append('c')  # OK
+
+     y: Final[Sequance[str]] = ['a', 'b']
+     y.append('x')  # Error: Sequance is immutable
+     z: Final = ('a', 'b')  # Also an option
+
+Final methods
```
gvanrossum: Ironic that here "Final" is capitalized even though in the code it uses a lower-case 'f', whereas in previous section headings "final" was not capitalized but the code used `Final`. :-) Maybe add "(the `@final` decorator)" to the title here?

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+   class Base:
+       @overload
+       def meth(self) -> None: ...
+       @overload
+       def meth(self, arg: int) -> int: ...
+       @final
+       def meth(self, x=None):
+           ...
+
+Final classes
```
gvanrossum: Add "(the `@final` class decorator)"?

docs/source/final_attrs.rst
```diff
@@ -0,0 +1,267 @@
...
+
+Here are some situations where using a final class may be useful:
+
+* A class wasn't designed to be subclassed. Perhaps subclassing does not
+  work as expected, or it's error-prone.
+* You want to retain the freedom to arbitrarily change the class implementation
+  in the future, and these changes might break subclasses.
+* You believe that subclassing would make code harder to understand or maintain.
+  For example, you may want to prevent unnecessarily tight coupling between
+  base classes and subclasses.
```
gvanrossum: These bullets all make almost the same point. Maybe we don't need to editorialize and users can figure out for themselves why final classes might be useful (they might even know the concept from another language).

docs/source/final_attrs.rst
ilevkivskyi: We discussed this with Jukka, and IIRC he suggested just "qualifiers".

docs/source/final_attrs.rst
ilevkivskyi: Because the instance attribute should be set only once, when the instance is created. I think we should also allow this in `__new__`, but this is blocked by https://github.com/python/mypy/issues/1021. I will try to reword here to make this clear and add a follow-up issue for `__new__`.

docs/source/final_attrs.rst
ilevkivskyi: > Again, why limit this to `__init__`?

Same as above.

docs/source/final_attrs.rst
ilevkivskyi: Yes.

docs/source/final_attrs.rst
ilevkivskyi: OK, I will update these.

docs/source/final_attrs.rst
ilevkivskyi: I am not sure this is needed, we can add this later.

docs/source/final_attrs.rst
ilevkivskyi: IIRC, Jukka proposed to add some examples, and I think it probably makes sense to keep at least some of them.

PR #5761, opened 2018-10-09
Make Frame stop being a dict subclass

msullivan:
This allows the 'unreachable' flag to not need to be accessed via
GetAttr, and for dict operations to be done via dict primitives
(though I'm going to put up a mypyc change soon that makes that second
part always true.)

Amazingly, this gives a ~200ms speedup on self check (like 6%!).

### DIFF
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -18,7 +18,7 @@ from mypy.nodes import IndexExpr, MemberExpr, NameExpr
 BindableExpression = Union[IndexExpr, MemberExpr, NameExpr]
 
 
-class Frame(Dict[Key, Type]):
+class Frame:
     """A Frame represents a specific point in the execution of a program.
     It carries information about the current types of expressions at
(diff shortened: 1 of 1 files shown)


mypy/binder.py
```diff
@@ -31,14 +31,11 @@ class Frame(Dict[Key, Type]):
...
+        self.types = {}  # type: Dict[Key, Type]
         self.unreachable = False
 
-
-class DeclarationsFrame(Dict[Key, Optional[Type]]):
-    """Same as above, but allowed to have None values."""
-
-    def __init__(self) -> None:
-        self.unreachable = False
+    def __getitem__(self, k: Key) -> Type:
```
gvanrossum: Could you remove the remaining uses of `frame[key]` into `frame.types[key]`? I notice that you did convert *some* operations on frames into operations into the `.types` attribute -- why not all? Or are there too many?

PR #5762, opened 2018-10-09
Fix #5758 -- plain import didn't work with namespace package

### LINKED ISSUE #5758: Importing types from namespace packages emits false positives without `from ... import ...`
Importing type hints from namespace packages only works when using the `from ... import ...` form.

Using the `--namespace-packages` flag works properly when the `typedpkg_namespace-alpha` is not installed but `mypy` is executed from the directory containing the package contents (as expected) but only for imports using the `from ... import ...` style. `import ...` does not work as expected. Likewise the same is true without the `--namespace-packages` flag if `typedpkg_namespace-alpha` is installed.

Specifically, the type hints are correctly imported but a false positive error message is still emitted of `Cannot find module named 'typedpkg_namespace'`.

# Repro steps
- Virtual env state
```
$ pip list
Package         Version
--------------- --------------------------------------------------
mypy            0.640+dev.e12be3ba9e6be3e9242e3a81ffaa9b3a136876fe
mypy-extensions 0.4.1
pip             18.1
setuptools      40.4.3
typed-ast       1.1.0
```
- Repro file structure
```
$ find . \( -path ./.mypy_cache -o -path ./venv \) -prune -o -print
.
./typedpkg_namespace
./typedpkg_namespace/alpha
./typedpkg_namespace/alpha/__init__.py
./typedpkg_namespace/alpha/py.typed
./typedpkg_namespace/alpha/alpha_module.py
./not_from_imp.py
./from_imp.py
./setup.py
```
- Repro files
    - `setup.py`
    ```
    setup(
        name='typedpkg_namespace.alpha',
        version='1.0.0',
        packages=find_namespace_packages(),
        namespace_packages=['typedpkg_namespace'],
        zip_safe=False,
        package_data={'typedpkg_namespace.alpha': ['py.typed']}
    )
    ```
    - `typedpkg_namespace/alpha/alpha_module.py`
    ```
    def alpha_func(a: bool) -> bool:
        return not a
    ```
    - `from_imp.py`
    ```
    from typedpkg_namespace.alpha.alpha_module import alpha_func
    alpha_func(False)
    alpha_func(2)
    ```
    - `not_from_imp.py`
    ```
    import typedpkg_namespace.alpha.alpha_module as am; alpha_func = am.alpha_func
    alpha_func(False)
    alpha_func(2)
    ```
    - `typedpkg_namespace/alpha/__init__.py` and `typedpkg_namespace/alpha/py.typed` are empty
    
# Repro Example
- using `--namespace-packages`
```
$ mypy from_imp.py  --namespace-packages
from_imp.py:3: error: Argument 1 to "alpha_func" has incompatible type "int"; expected "bool"

$ mypy not_from_imp.py --namespace-packages
not_from_imp.py:1: error: Cannot find module named 'typedpkg_namespace'
not_from_imp.py:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
not_from_imp.py:3: error: Argument 1 has incompatible type "int"; expected "bool"
```
- using installed package
```
$ pip install .
Looking in indexes: http://artifactory.factset.com/artifactory/api/pypi/python/simple/
Processing /home/user/cphilip/side_projects/test_namespace/typedpkg_namespace-alpha
Installing collected packages: typedpkg-namespace.alpha
  Running setup.py install for typedpkg-namespace.alpha ... done
Successfully installed typedpkg-namespace.alpha-1.0.0

$ pip list
Package                  Version
------------------------ --------------------------------------------------
mypy                     0.640+dev.e12be3ba9e6be3e9242e3a81ffaa9b3a136876fe
mypy-extensions          0.4.1
pip                      18.1
setuptools               40.4.3
typed-ast                1.1.0
typedpkg-namespace.alpha 1.0.0

$ mypy from_imp.py
from_imp.py:3: error: Argument 1 to "alpha_func" has incompatible type "int"; expected "bool"

$ mypy not_from_imp.py
not_from_imp.py:1: error: Cannot find module named 'typedpkg_namespace'
not_from_imp.py:1: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
not_from_imp.py:3: error: Argument 1 has incompatible type "int"; expected "bool"
```

gvanrossum:


gvanrossum:
So here's a fix for part two of #5758. The way it works is, first, when `find_module()` encounters an installed non-stub package with a `py.typed` marker underneath a namespace parent module, it remembers the namespace ancestor(s) in a special cache dict. Then if a later request for one of those ancestors is received and it is not found in the usual way, the cached ancestor's directory is produced.

But that's not all that's needed! @macbeth322 -- you were spot on with your suggestion that this had something to do with the ancestor dependencies in `all_imported_modules_in_file()`! I ended up moving the "main" dependency to the top so it precedes the ancestors; that way, the first call to `find_module()` is e.g. `find_module('foo.bar.baz')`, which squirrels away the namespace ancestor as I described above.

This order dependency is ugly and caused a bunch of tests to fail (since the errors about missing packages come out in a different order) but it's the best I could come up with.

(As I write this there are some more failures, I have to look into those later.)

macbeth322:
I was able to test 0fa27de1c394a99fa9d89f2e9c2ab38b6e45041b and it looks like it works for all the edge cases I found.

mypy/build.py
```diff
@@ -572,12 +572,12 @@ def correct_rel_imp(imp: Union[ImportFrom, ImportAll]) -> str:
                     pri = import_priority(imp, PRI_MED)
                     ancestor_pri = import_priority(imp, PRI_LOW)
                     for id, _ in imp.ids:
+                        res.append((pri, id, imp.line))
```
msullivan: This somewhat subtle order dependency might deserve some comment.

mypy/modulefinder.py
```diff
@@ -222,7 +235,11 @@ def _find_module(self, id: str) -> Optional[str]:
             index = levels.index(max(levels))
             return near_misses[index]
 
-        return None
+        # Finally, we may be asked to produce an ancestor for an
+        # installed package with a py.typed marker that is a
+        # subpackage of a namespace package.  We only fess up to these
+        # if we would otherwise return "not found".
+        return self.ns_ancestors.get(id)
```
msullivan: Is it possible to add a test for this case?

gvanrossum:
Added a comment. I believe adding a test for this situation would require adding a big chunk of code to testpep561.py. I'd rather focus on other aspects of the release, but I'll file a bug to remind myself.

PR #5764, opened 2018-10-10
Split runtests.py pytest run into 3 parts to improve parallelization

### LINKED ISSUE #5755: Tests not parallelized efficienctly by pytest
When I run `pytest mypy`, towards the end of a test run only 1 or 2 python processes are running and test progress is very slow -- each process is using much less than 100% of a CPU. This happens even though I requested 8 parallel processes. Initially I see 8 processes doing work, but something causes a slowdown towards the end of a test run. 

This doesn't look like a recent regression, but I think that things have gotten worse in the last few months or so.

Some data (Linux, Intel 4-core/8-thread CPU):
* When I check out e3786018ddcdac065a89b2db9889fb04165ee71e from May 2018, tests finish in 66s (6024 tests).
* When I check out 1606f0f9174dd168fa4848b5dafdc27174c67e9e from Oct 2018, tests finish in 133s (6671 tests).

This seems to happen both on Linux and macOS.

Updating to the latest pytest doesn't help.

It would be great if somebody could try to reproduce the issue.

JukkaL:
This can speed up local test runtimes by 15% to 50%. Also this makes it easier
to rerun only a part of the test suite.

Fixes #5755. 

The same issue may also affect travis and appveyor builds, and this doesn't
address them.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -14,11 +14,35 @@ else:
     else:
         python_name = 'python3'
 
+# Slow test suites
+CMDLINE = 'PythonCmdline'
+SAMPLES = 'SamplesSuite'
+TYPESHED = 'TypeshedSuite'
+PEP561 = 'TestPEP561'
+EVALUATION = 'PythonEvaluation'
+
+ALL_NON_FAST = [CMDLINE, SAMPLES, TYPESHED, PEP561, EVALUATION]
+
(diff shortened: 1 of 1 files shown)


JukkaL:
We can revert this change once pytest is able to parallelize tests efficiently without help.

PR #5768, opened 2018-10-11
 Add option to raise exception on internal errors

TV4Fun:
For using external debugging tools like PyCharm, the most helpful thing
to do on an internal error is to just raise an exception, rather than
printing a stacktrace or dropping to PDB.


gvanrossum:
I guess it's arguable what's the right thing to do though. What does your caller want to do with the exception, other than print a traceback?

OTOH maybe the code that *calls* `report_internal_error()` should check this flag instead, and just call `raise` (to re-raise the original exception) so it can be caught and debugged in its full glory. I'd use that occasionally when developing mypy.

PR #5777, opened 2018-10-12
Allow setting python_executable from config file...

### LINKED ISSUE #5620: Cannot get type hints from python 2 packages
In mypy 0.620, there appears to be inconsistent type checking behaviour depending on how you specify the python version. This exists for handling `-2` and `--python-version 2` arguments and is explained well in [issue #5576](https://github.com/python/mypy/issues/5576#issuecomment-418872062) and addressed by pull request #5619. However, this inconsistency also occurs with specifying the python version in the a mypy.ini configuration file.

```
$ mypy --python-version 2.7 pkgb
pkgb/__main__.py:7: error: Argument 1 to "this_is_a_typed_function" has incompatible type "int"; expected "unicode"
```
That above is expected, but when specifying the version in a config file.
```
$ echo -e "[mypy]\npython_version: 2.7" | mypy --config-file /dev/stdin pkgb
pkgb/__main__.py:1: error: Cannot find module named 'pkga'
```

For good measure, I included a python 2 print statement in the test package to ensure it is actually is running under python 2 mode ... [The example is included here](https://github.com/python/mypy/files/2383934/mypy5576.tar.gz). Extract the archive; in a Python 2 virtualenv, install the package in the `a` directory with `pip -e a`; run mypy on `pkgb`.

---

The behaviour seems to come from looking for typing information in different site-packages between the two cases. With `--python-version 2.7`, a Python 2 interpreter's site-packages are used. With the configuration file, the current interpreter's site-packages are used.

The pull request #5578 was an attempt to resolve this inconsistency by using the Python 2 site-packages in both configurations, but it seems like this broke some things for some folks.

ethanhs:
... and let the `python_version` and `python_executable` be different, so that one can install stub packages into the Python used to run mypy.

Fixes #5620.


ethanhs:
I realized this isn't actually correct yet, so I'm not quite ready for review yet.

ethanhs:
It appears the new version of flake8 is very picky. The Travis failure is not related to these changes (its just flake8). I will submit a PR to fix this.

ethanhs:
Okay this should be ready for review.

ethanhs:
>It might be worth adding a test that actually does read the version from a config file

There are several tests for reading the version from the config file already (in `cmdline.test`). I would do the same for `python_executable`, but that is a bit harder to test. I suppose I could add something to testpep561 if such a test is desired.

ethanhs:
I will merge this later today unless someone has suggestions for changes. (@msullivan not sure if you wanted to respond to my above comment).

mypy/main.py
```diff
@@ -264,26 +264,25 @@ def infer_python_version_and_executable(options: Options,
...
-                ))
-        else:
-            options.python_version = special_opts.python_version
-            options.python_executable = special_opts.python_executable
-    elif special_opts.python_executable is None and special_opts.python_version is not None:
-        options.python_version = special_opts.python_version
+
+    # Use the command line specified python_version/executable, or fall back to one set in the
+    # config file
+    python_version = special_opts.python_version or options.python_version
```
gvanrossum: I don't see how this can ever be None -- it is initialized to `sys.version_info[:2]` in options.py. So that means that the final `elif` below cannot be reached. What am I missing?

ethanhs: Ah, good point. I suppose if we are falling back to using the Python version of the running Python that becomes unneeded.

gvanrossum: (Note that in general I'm not keen on this particular if/elif/elif approach to decoding combinations of two booleans -- see also #5855.)

ethanhs: Yeah, I can definitely re-write this to be cleaner/more efficient. Thank you for pointing that out!

ethanhs:
Ah, yes I have it using the default Python executable. I will need to think this over a bit. I believe the issue is that we want to be able to use `options.python_executable`, but it already has a default, so there is no way to tell if the config file has set the value and we don't need to do inference, or if it just is using the default value and we want to infer the correct executable based on the Python version.

I suppose the "correct" thing to do would probably be to make `python_executable` and `python_version` default to `None` when unset and have the entry point set their defaults if all else fails.

gvanrossum:
Hm, it looks like the logic for the Python version could be simplified: just use the default `dest` (don't bother with `special-opts:python_version`). For executable I think you could also do away with `special-opts`: in options.py you set the default to `None` and then in main.py you compute the value dynamically like this:

- if explicitly specified, keep that
- else if `options.no_executable` is set, use `None`
- else compute from version

It looks like everything becomes easier that way.

gvanrossum:
@ethanhs Any change to get this in for the 0.650 release? I think it would be a nice improvement.

gvanrossum:
(If you want to push back on my suggestion to simplify the implementation and just merge the current version I could be okay with that too.)

ethanhs:
Oh, sorry I seem to have missed your message from two weeks back. I will make another PR tomorrow with the simplification.

PR #5780, opened 2018-10-12
Special case None.__bool__()

### LINKED ISSUE #5680: error: Item "None" has no attribute "__bool__"
```python
thing = None
assert not thing.__bool__()
print(thing.__bool__())
```

As expected:
```
>python repro
False
```

Not expected:
```
>mypy repro
test:2: error: Item "None" has no attribute "__bool__"
test:3: error: Item "None" has no attribute "__bool__"
```

cf. python/typeshed#2432

darabos:
Resolves #5680.

Sorry, I don't know if I'm doing this right. We started using Mypy about half a year ago, so I thought I'd try to make some small contribution. The comment from @ilevkivskyi made this sound simple enough, and the test passes, but I'm not entirely confident of this. There are no other hacks like this in `checkmember.py`...

Also I have no idea where this test belongs. Seems like a "basic" thing, so I put it in `check-basic.test`. 😅
Thanks!

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -109,8 +109,10 @@ def analyze_member_access(name: str,
     elif isinstance(typ, NoneTyp):
         if chk.should_suppress_optional_error([typ]):
             return AnyType(TypeOfAny.from_error)
-        if name == '__bool__':
-            # The only attribute of NoneType that is not inherited from object.
+        is_python_2 = chk.options.python_version[0] == 2
+        # In Python 2 "None" has exactly the same attributes as "object". Python 3 adds a single
+        # extra attribute, "__bool__".
+        if not is_python_2 and name == '__bool__':
             return CallableType(arg_types=[],
                                 arg_kinds=[],
                                 arg_names=[],
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -109,10 +109,10 @@ def analyze_member_access(name: str,
     elif isinstance(typ, NoneTyp):
         if chk.should_suppress_optional_error([typ]):
             return AnyType(TypeOfAny.from_error)
-        is_python_2 = chk.options.python_version[0] == 2
+        is_python_3 = chk.options.python_version[0] >= 3
         # In Python 2 "None" has exactly the same attributes as "object". Python 3 adds a single
         # extra attribute, "__bool__".
-        if not is_python_2 and name == '__bool__':
+        if is_python_3 and name == '__bool__':
             return CallableType(arg_types=[],
                                 arg_kinds=[],
                                 arg_names=[],
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -109,10 +109,17 @@ def analyze_member_access(name: str,
     elif isinstance(typ, NoneTyp):
         if chk.should_suppress_optional_error([typ]):
             return AnyType(TypeOfAny.from_error)
-        # The only attribute NoneType has are those it inherits from object
-        return analyze_member_access(name, builtin_type('builtins.object'), node, is_lvalue,
-                                     is_super, is_operator, builtin_type, not_ready_callback, msg,
-                                     original_type=original_type, chk=chk)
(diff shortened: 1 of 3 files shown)


test-data/unit/check-basic.test
```diff
@@ -373,3 +373,10 @@ def foo(
 ):
     # type: (...) -> None
     pass
+
+[case testNoneHasBool]
+none = None
+b = none.__bool__()
+s = 'hi'
+s = b  # E: Incompatible types in assignment (expression has type "bool", variable has type "str")
```
ilevkivskyi: You can just use `reveal_type()` to check that the type is correct.

mypy/checkmember.py
```diff
@@ -109,10 +109,17 @@ def analyze_member_access(name: str,
     elif isinstance(typ, NoneTyp):
         if chk.should_suppress_optional_error([typ]):
             return AnyType(TypeOfAny.from_error)
-        # The only attribute NoneType has are those it inherits from object
-        return analyze_member_access(name, builtin_type('builtins.object'), node, is_lvalue,
-                                     is_super, is_operator, builtin_type, not_ready_callback, msg,
-                                     original_type=original_type, chk=chk)
+        if name == '__bool__':
+            # The only attribute of NoneType that is not inherited from object.
```
ilevkivskyi: The position and content of this comment should be updated.

mypy/checkmember.py
```diff
@@ -109,10 +109,17 @@ def analyze_member_access(name: str,
     elif isinstance(typ, NoneTyp):
         if chk.should_suppress_optional_error([typ]):
             return AnyType(TypeOfAny.from_error)
-        # The only attribute NoneType has are those it inherits from object
-        return analyze_member_access(name, builtin_type('builtins.object'), node, is_lvalue,
-                                     is_super, is_operator, builtin_type, not_ready_callback, msg,
-                                     original_type=original_type, chk=chk)
+        if name == '__bool__':
```
ilevkivskyi: Note that that this special case should be added only on Python 3.

Please also add a test that `None.__bool__` fails on Python 2.

test-data/unit/check-basic.test
darabos: > You can just use `reveal_type()` to check that the type is correct.

Done, thanks!

mypy/checkmember.py
darabos: > The position and content of this comment should be updated.

I've updated the position and content. Hopefully for the better. :sweat_smile:

mypy/checkmember.py
darabos: > Note that that this special case should be added only on Python 3.

Oops, I'd missed that. Thanks, fixed.

> Please also add a test that `None.__bool__` fails on Python 2.

Done.

mypy/checkmember.py
```diff
@@ -109,10 +109,19 @@ def analyze_member_access(name: str,
...
         if chk.should_suppress_optional_error([typ]):
             return AnyType(TypeOfAny.from_error)
-        # The only attribute NoneType has are those it inherits from object
-        return analyze_member_access(name, builtin_type('builtins.object'), node, is_lvalue,
-                                     is_super, is_operator, builtin_type, not_ready_callback, msg,
-                                     original_type=original_type, chk=chk)
+        is_python_2 = chk.options.python_version[0] == 2
+        # In Python 2 "None" has exactly the same attributes as "object". Python 3 adds a single
+        # extra attribute, "__bool__".
+        if not is_python_2 and name == '__bool__':
```
ilevkivskyi: It would read more natural if you use `if is_python_3 ...` here and define it as `is_python_3 = chk.options.python_version[0] >= 3` above.

darabos: > It would read more natural if you use `if is_python_3 ...` here and define it as `is_python_3 = chk.options.python_version[0] >= 3` above.

Done. Though now `is_python_3` will be True on Python 4. 😸 
Thanks!

Issue #5767, opened 2018-10-10
Add a test for #5758 part two


gvanrossum:
See https://github.com/python/mypy/pull/5762#issuecomment-428755257

macbeth322:
I can work on this over the next week unless you were already on planning working on it.

gvanrossum:
> I can work on this over the next week

Please do!

gvanrossum:
Had we had this test, then this bug would (hopefully) have been found before releasing 0.640: https://github.com/python/mypy/issues/5784

PR #5782, opened 2018-10-12
Refactor testpep561 and add test to address #5767

macbeth322:
I took some liberties while refactoring this to more fit my personal style and hopefully make more readable/extensible. I also added test cases for all the edge cases I found and filed in #5758; including a test case for which #5767 was filed.

I understand this is a pretty big change to this file and I am open to redoing it if its not popular.


gvanrossum:
Let's make sure this catches https://github.com/python/mypy/issues/5784

gvanrossum:
I would very much like to use these tests to prove that I've fixed #5784 (rather than just manually running the example shown there). But for that to happen this would probably have to be brought back to minimally invasive form -- or at least it would have to pass the tests on AppVeyor.

macbeth322:
Woops sorry, I'm moving too fast. I just saw I got an error and commited it. I'll work on actually repro'ing now that I've properly read the issue.

macbeth322:
Build failures so far are because of #5784. Right now, I am not really sure what the test should actually look like but for now it reproduces the issue at least, I will make a better test once the issue gets fixed.

gvanrossum:
See if the one-line fix in https://github.com/python/mypy/pull/5785 (latest version) fixes it.

macbeth322:
#5785 worked locally

macbeth322:
@ethanhs could you share a link to the repro that doesn't use c extensions?

gvanrossum:
I've got a hunch that what you need to repro the issue is an installed package with a `py.typed` marker, and an import of a non-existent submodule of that package with a `# type: ignore` on the import. See the repro in https://github.com/python/mypy/issues/5784 -- since mypy doesn't see the C extension it doesn't matter whether it's installed or not.

I think that you have to install at least something like `pkg/mod.py` and inside that do `from pkg.other import blah # type: ignore`.

mypy/test/testpep561.py
```diff
@@ -32,27 +32,100 @@
...
+from typedpkg_c_ext.hello import helloworld
+
+speak("abc")
+speak(1)
+"""
+
+
+class NamespaceProgramImportStyle(Enum):
+    from_import = """\
+from typedpkg_nested.nested_package.nested_module import nested_func
```
gvanrossum: I dislike having those long module names. If you give each of them names of 3-5 letters it will become more readable (and you won't have to break long lines with `\` below). Also a comment may be in order explaining why some things must go together on one line (IIUC it's so that the line numbers in the error messages are consistent?)

mypy/test/testpep561.py
```diff
@@ -102,127 +175,162 @@ def install_package(self, pkg: str,
...
-        self.msg_tuple = \
-            "{}:5: error: Revealed type is 'builtins.tuple[builtins.str]'\n".format(self.tempfile)
-
-        self.namespace_msg_bool_str = (
-            '{0}:8: error: Argument 1 to "nested_func" has incompatible type "bool"; '
-            'expected "str"\n'.format(self.namespace_tempfile))
-        self.namespace_msg_int_bool = (
-            '{0}:9: error: Argument 1 to "alpha_func" has incompatible type "int"; '
-            'expected "bool"\n'.format(self.namespace_tempfile))
+        self.simple_example_program = ExampleProgram(SIMPLE_PROGRAM)
```
gvanrossum: These variable names are all too long -- when skimming all I see is `blah_blah_blah_blah`. I don't think they all need to end in `_example_program`, and you can abbreviate `namespace` as `ns`.

test-data/packages/typedpkg_c_ext/hello.c
```diff
@@ -0,0 +1,35 @@
+/*******************************************************************
+ * This code was copied from the article written by Elliot Forbes at
+ * https://tutorialedge.net/python/python-c-extensions-tutorial/
+ ******************************************************************/
+#include <Python.h>
+
+// ot ForbesFunction 1: A simple 'hello world' function
```
gvanrossum: "ot"?

test-data/packages/typedpkg_c_ext/setup.py
```diff
@@ -0,0 +1,12 @@
+from setuptools import setup, Extension
+
+setup(
+    name='typedpkg_c_ext',
+    version='1.0',
+    packages=['typedpkg_c_ext'],
+    ext_modules=[Extension('typedpkg_c_ext.hello', ['hello.c'])],
```
gvanrossum: Looks like you can just delete this line (and the file hello.c) and the test works just as well. (Verified by removing the `isdir()` check from modulefinder.c line 124.)

gvanrossum:
There's something else missing. I patched this in, and then put an early `return` in `_update_ns_ancestors()` in modulefinder.py. This should simulate *not* fixing part two of #5758, notably https://github.com/python/mypy/pull/5762/commits (with an insightful [review comment](https://github.com/python/mypy/pull/5762/commits/809083878f5678bf114f0c319020729a9076e1d9#r224251059) that I ignored).

But all pep561 tests still pass with that change (except `test_typedpkg_editable()` still fails). So it looks like the requested test still doesn't exist...

macbeth322:
I reorganized the test files so there isn't as much overlap with the test packages and made names in general shorter. I removed the c extensions test package because the repro can be done without it; and I made sure that the #5767 case is covered.

Side Note: I am expecting the tests to fail because it looks like there is repeated warning output in some cases.

gvanrossum:
But most tests are failing in the install phase -- is that also expected?

What warnings are you expecting exactly?

macbeth322:
I was reading the `pytest` output wrong, tests should pass now.

gvanrossum:
I still get the failure on `test_typedpkg_editable()`. Any insight on how I should debug this?

ilevkivskyi:
> I still get the failure on `test_typedpkg_editable()`. Any insight on how I should debug this?

Are you under a virtualenv on a Mac? If yes, then this is not new, it fails since some time.

ilevkivskyi:
I am seeing this:
```
E           AssertionError: 
E           assert "/var/folders...pe is 'Any'\n" == "/var/folders/...ltins.str]'\n"
E             Skipping 61 identical leading characters in diff, use -v to show
E             + simple.py:5: error: Revealed type is 'builtins.tuple[builtins.str]'
E             - simple.py:2: error: Cannot find module named 'typedpkg.sample'
E             - /var/folders/ff/8tg5kw8x3jg3z9cw0359hllxss39wp/T/tmph7q_wh1f/simple.py:2: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
E             - /var/folders/ff/8tg5kw8x3jg3z9cw0359hllxss39wp/T/tmph7q_wh1f/simple.py:3: error: Cannot find module named 'typedpkg'
E             - /var/folders/ff/8tg5kw8x3jg3z9cw0359hllxss39wp/T/tmph7q_wh1f/simple.py:5: error: Revealed type...
E             
E             ...Full output truncated (1 line hidden), use '-vv' to show
```
for few month, but I have never had time to fix it.

gvanrossum:
Yeah, that's what I get too (the [traceback](https://github.com/python/mypy/pull/5782#pullrequestreview-164482850) I reported is about the same). And yes, I am under a venv.

ilevkivskyi:
I think this is an issue with the test infrastructure, since the same scenario works fine with installed mypy. A natural guess is that something goes wrong because this tests creates a new virtual environment that probably conflicts somehow with the current environment.

gvanrossum:
Hm, I should really leave this alone, but I investigated a bit more and found that when an outer venv is already active, [the line](https://github.com/python/mypy/pull/5782/files#diff-38656e5c6aa35a48ebb3f79b483b0122R266) `self.install_package('typedpkg', python_executable, editable=True)` installs the `typedpkg.egg-link` file in the outer venv. And apparently the test cannot find it there. Possibly this is a pip issue -- or possibly the code in `install_package()` needs something extra to convince pip to use the inner venv.

gvanrossum:
One more internal monologue and then I'll let it go. :-)

I added `-v` to the pip command and the crucial bits seem to be this:
```
Installing collected packages: typedpkg
  Found existing installation: typedpkg 0.1
    Not uninstalling typedpkg at /Users/guido/src/mypy/test-data/packages/typedpkg, outside environment /var/folders/63/czkyq6090dd0t157zhx54xvhrdlybt/T/tmp5ov0t1ad/bin/..
    Can't uninstall 'typedpkg'. No files were found to uninstall.
```
There's other verbose output before and after this, but in a successful run, instead of this I see:
```
Installing collected packages: typedpkg
  Found existing installation: typedpkg 0.1
    Uninstalling typedpkg-0.1:
      Created temporary directory: /private/var/folders/63/czkyq6090dd0t157zhx54xvhrdlybt/T/pip-uninstall-5hfyyz5b
      Removing file or directory /Users/guido/src/mypy/vv/lib/python3.7/site-packages/typedpkg.egg-link
      Removing pth entries from /Users/guido/src/mypy/vv/lib/python3.7/site-packages/easy-install.pth:
      Removing entry: /Users/guido/src/mypy/test-data/packages/typedpkg
      Successfully uninstalled typedpkg-0.1
```
Additional notes:

- When I *manually* create two venvs (v1 and v2), activate v1, and run `v2/bin/pip install -e .../typedpkg` in the other, the egg-link file gets created in the right place.
- If I don't activate v1 but run `v1/bin/python3 -m pytest ...` (so effectively v2 is created by the test program) the problem still appears. 
- I added another test, `test_typedpkg_egg_editable()`, which uses `use_pip=False, editable=True` (i.e. invoking `setup.py develop` directly) and that one passed.

mypy/test/testpep561.py
```diff
@@ -102,127 +170,148 @@ def install_package(self, pkg: str,
...
         with self.virtualenv() as venv:
             venv_dir, python_executable = venv
             self.install_package('typedpkg-stubs', python_executable)
-            check_mypy_run(
-                [self.tempfile],
+            self.simple_ep.check_mypy_run(
                 python_executable,
-                expected_out=self.msg_dne + self.msg_list,
+                [SimpleMsg.msg_dne,
+                 SimpleMsg.msg_list],
```
gvanrossum: (I don't care much for splitting this line -- the whole list easily fits.)

mypy/test/testpep561.py
```diff
@@ -21,38 +22,105 @@
...
-        os.chdir(venv_dir)
-    try:
-        if python_executable != sys.executable:
-            cmd_line.append('--python-executable={}'.format(python_executable))
-        out, err, returncode = mypy.api.run(cmd_line)
-        assert out == expected_out, err
-        assert err == expected_err, out
-        assert returncode == expected_returncode, returncode
-    finally:
+class NSImportStyle(Enum):
```
ethanhs: I like moving these into enums, thank you for doing that!

mypy/test/testpep561.py
```diff
@@ -21,38 +22,105 @@
...
+class ExampleProg(object):
+    _fname = 'test_program.py'
+
+    def __init__(self, source_code: str) -> None:
+        self._source_code = source_code
+
+        self._temp_dir = None  # type: Optional[tempfile.TemporaryDirectory[str]]
+        self._full_fname = ''
+
+    def init(self) -> None:
```
ethanhs: Suggestion: rename to `create`? `init` doesn't really tell me what it is doing, `create` implies it is doing setup work and creating the test case.

mypy/test/testpep561.py
ethanhs: I'm glad to see these going into Enums. :)

mypy/test/testpep561.py
```diff
@@ -21,38 +22,105 @@
...
+                '"bool"; expected "str"')
+    to_int_bool = ('{tempfile}:11: error: Argument 1 to "bf" has incompatible type '
+                '"int"; expected "bool"')
+
+
+def create_ns_program_src(import_style: NSImportStyle) -> str:
+    return _NAMESPACE_PROGRAM.format(import_style=import_style.value)
+
+
+class ExampleProg(object):
```
ethanhs: Maybe change this to just `Program` and change the suffixes below to `self.simple_ep` to `self.simple_prog` or something? `ep` is not a straightforward mnemonic (at least to me).

Issue #5784, opened 2018-10-12
No such file or directory for C module


bryanforbes:
Starting with the following using `mypy` version `0.640`:

```
$ python3 -m venv .venv
$ .venv/bin/pip install -U aiohttp mypy
$ echo "import aiohttp" > test.py
```

Running `.venv/bin/mypy test.py` leads to an error:

```
mypy: can't read file '/Users/bryan/Projects/mypy-repro/.venv/lib/python3.7/site-packages/aiohttp/_websocket': No such file or directory
```

This module is a C module that is imported like this:

```python
if NO_EXTENSIONS:
    _websocket_mask = _websocket_mask_python
else:
    try:
        from ._websocket import _websocket_mask_cython  # type: ignore
        _websocket_mask = _websocket_mask_cython
    except ImportError:  # pragma: no cover
        _websocket_mask = _websocket_mask_python
```

gvanrossum:
Yeah, I can repro this too. And it's a regression from 0.630. :-(

PR #5785, opened 2018-10-12
Avoid "No such file or directory" if module ancestor doesn't exist

gvanrossum:
Tentative fix for #5784.

(An alternative fix would be not to add anything to `self.ns_ancestors` unless `self.options.namespace_packages` is set.)

### DIFF
--- a/mypy/modulefinder.py
+++ b/mypy/modulefinder.py
@@ -121,7 +121,7 @@ class FindModuleCache:
         path, verify = match
         for i in range(1, len(components)):
             pkg_id = '.'.join(components[:-i])
-            if pkg_id not in self.ns_ancestors:
+            if pkg_id not in self.ns_ancestors and self.fscache.isdir(path):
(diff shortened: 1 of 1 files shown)


gvanrossum:
Beware: When testing manually, the cache may suppress errors.

gvanrossum:
I am now sufficiently confident that this version of the fix (which leaves all the original code in place but adds an `isdir()` check before putting a path in `ns_ancestors`) is right. I am going to merge this and cherry-pick it into the release-0.641 branch.

(Why am I confident? I added a bunch of `print()` calls and tested this with the repro from #5784 (`import aiohttp`), as well as with the tests from #5758 (`typedpkg_namespace.alpha`), and it gives exactly the expected errors and the output explains why.)

However before I re-cut a release I want #5782 (Refactor testpep561 etc.) to land too. That requires further review. I may or may not merge that into the release-0.641 branch (since I like the idea of having absolutely minimal changes in a "brown bag release" like that).

PR #5789, opened 2018-10-15
check type of module_all earlier

graingert:
currently mypy.stubgen crashes when processing `aiofiles/__init__.py`, because `__all__` is `Tuple[Callable[..., object]]` and doesn't continue to process any further modules

See https://github.com/Tinche/aiofiles/pull/53 for the upstream fix

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -172,6 +172,8 @@ def find_module_path_and_all(module: str, pyversion: Tuple[int, int],
             raise SystemExit(
                 "Can't find module '{}' (consider using --search-path)".format(module))
         module_all = None
+    if not (isinstance(module_all, list) and all(isinstance(v, str) for v in module_all)):
+        raise CantImport(module)
     return module_path, module_all
 
 
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -174,7 +174,7 @@ def find_module_path_and_all(module: str, pyversion: Tuple[int, int],
         module_all = None
     if not (
         module_all is None or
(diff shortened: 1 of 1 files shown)


graingert:
Instead it might be better to convert the `__all__` to a `Option[List[str]]` by dropping non-strs

mypy/stubgen.py
```diff
@@ -172,6 +172,11 @@ def find_module_path_and_all(module: str, pyversion: Tuple[int, int],
             raise SystemExit(
                 "Can't find module '{}' (consider using --search-path)".format(module))
         module_all = None
+    if not (
+        module_all is None or
+        (isinstance(module_all, list) and all(isinstance(v, str) for v in module_all))
+    ):
+        raise CantImport(module)
```
ilevkivskyi: This error might be confusing, because we were able to import the module, but it contains a critical error. Maybe add a message (explaining cause) to this exception class?

mypy/stubgen.py
```diff
@@ -172,6 +172,11 @@ def find_module_path_and_all(module: str, pyversion: Tuple[int, int],
             raise SystemExit(
                 "Can't find module '{}' (consider using --search-path)".format(module))
         module_all = None
+    if not (
+        module_all is None or
+        (isinstance(module_all, list) and all(isinstance(v, str) for v in module_all))
+    ):
```
ilevkivskyi: This statement is a bit hard to read, I think it would be easier if you reformulate it as `if module_all is not None and <bad __all__...>`.

msullivan:
I'm going to close this. Feel free to reopen if you pick it back up.

PR #5791, opened 2018-10-15
Remove fine-grained deps from "__init__" on class attribute access

JukkaL:
The main motivation is to improve the precision of logical
dependencies, as the previously generated dependencies were
unnecessary. Removing extra dependencies also gives a slight
performance improvement.

### DIFF
--- a/mypy/server/deps.py
+++ b/mypy/server/deps.py
@@ -586,7 +586,11 @@ class DependencyVisitor(TraverserVisitor):
         self.process_global_ref_expr(o)
 
     def visit_member_expr(self, e: MemberExpr) -> None:
-        super().visit_member_expr(e)
+        if isinstance(e.expr, RefExpr) and isinstance(e.expr.node, TypeInfo):
+            # Special case class attribute so that we don't depend on "__init__".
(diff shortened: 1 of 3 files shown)


mypy/server/deps.py
```diff
@@ -586,7 +586,11 @@ def visit_name_expr(self, o: NameExpr) -> None:
         self.process_global_ref_expr(o)
 
     def visit_member_expr(self, e: MemberExpr) -> None:
-        super().visit_member_expr(e)
+        if isinstance(e.expr, RefExpr) and isinstance(e.expr.node, TypeInfo):
+            # Special case class attribute so that we don't depend on "__init__".
+            self.add_dependency(make_trigger(e.expr.node.fullname()))
```
ilevkivskyi: Technically it is not a 1:1 refactoring: the new version doesn't get the metaclass attribute dependencies, previously they were added by this code:
```python
        triggers.extend(self.attribute_triggers(typ.fallback, name))
```
I am however not 100% sure they are needed. Could you please add a test case for this scenario?

ilevkivskyi: Oh, wait, it is still called, since we don't return here. I would still add the test just in case.

PR #5793, opened 2018-10-16
Fix fine-grained dependencies for super()

JukkaL:
This fixes both normal and logical dependencies. We now generate
dependencies from potential target attributes in base classes. Also,
we don't generate logical dependencies based on the type object
in Python 2 style `super()` expressions, since they aren't
relevant.

### DIFF
--- a/test-data/unit/deps.test
+++ b/test-data/unit/deps.test
@@ -1329,9 +1329,11 @@ class B(A):
 class C(B):
     pass
 class D(C):
-    def __init__(self): # type: () -> None
+    def __init__(self):
+        # type: () -> None
         super(B, self).__init__()
-    def mm(self): # type: () -> None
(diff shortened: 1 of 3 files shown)


test-data/unit/deps.test
```diff
@@ -1315,3 +1315,35 @@ def h() -> None:
...
+        pass
+class B(A):
+    def m(self):
+        pass
+class C(B):
+    pass
+class D(C):
+    def __init__(self): # type: () -> None
+        super(B, self).__init__()
+    def mm(self): # type: () -> None
```
ilevkivskyi: Why function signature on the same line here and above? This looks unusual.

JukkaL: No particular reason. I'll move them to separate lines.

Issue #1706, opened 2016-06-14
Adding @abstractmethod __init__ to ABC breaks staticmethods


euresti:
I was trying to create an ABC that also worked as a factory.

```
from abc import (
    ABCMeta,
    abstractmethod,
)

class WidgetWithInit(object):
    __metaclass__ = ABCMeta

    @abstractmethod
    def __init__(self, foo, bar):
        # type: (int, int) -> None
        pass

    @staticmethod
    def add(foo, bar):
        # type: (int, int) -> int
        return foo + bar

    @classmethod
    def add_class(cls, foo, bar):
        # type: (int, int) -> int
        return foo + bar

WidgetWithInit.add(1, '2')
WidgetWithInit.add_class(1, '2')

class WidgetNoInit(object):
    __metaclass__ = ABCMeta

    @staticmethod
    def add(foo, bar):
        # type: (int, int) -> int
        return foo + bar

    @classmethod
    def add_class(cls, foo, bar):
        # type: (int, int) -> int
        return foo + bar

WidgetNoInit.add(1, '2')
WidgetNoInit.add_class(1, '2')
```

MyPy will respond:

```
rt6.py:40: error: Argument 2 to "add" of "WidgetNoInit" has incompatible type "str"; expected "int"
rt6.py:41: error: Argument 2 to "add_class" of "WidgetNoInit" has incompatible type "str"; expected "int"
```

So adding the `def __init__` seems to make the staticmethod and classmethod have a type of Any.

Fortunately I can work around it by not defining **init** in the ABC, and having my `def create` enforce the signature.

rwbarton:
In fact mypy thinks the expression `WidgetNoInit` already has type `Any` with the definition
```py
class WidgetWithInit(object):
    @abstractmethod
    def __init__(self, foo, bar):
        # type: (int, int) -> None
        pass
```

rwbarton:
In fact `__init__` having any decorator will result in the class being treated as `Any`, since `type_object_type` starts like this:
```py
    init_method = info.get_method('__init__')
    if not init_method:
        # Must be an invalid class definition.
        return AnyType()
    else: ...
```
and `get_method` doesn't return decorated things.

rwbarton:
Here's a possible test case
```py
[case testAbstractInit]
from abc import abstractmethod, ABCMeta
class A(metaclass=ABCMeta):
    @abstractmethod
    def __init__(self, a: int) -> None:
        pass
class B(A):
    pass
class C(B):
[...]

PR #5798, opened 2018-10-18
Fix decorators on class __init__ methods

### LINKED ISSUE #5398: Decorator on __init__ causes loss of type info
Here's a simple example:

```
from typing import Callable

def decorator(method: Callable[[Foo], None]) -> Callable[[Foo], None]:
    return method

class Foo():
    @decorator
    def __init__(self):
        pass

reveal_type(Foo.__init__)
```

The output of mypy is:

```
% mypy --version
mypy 0.620
% mypy foo.py
foo.py:13: error: Revealed type is 'Any'
```

This seems to be specific to methods as similar type inferences work fine for functions outside of the class scope.

TV4Fun:
This fixes a few issues around decorators, specifically resolves #5398 and #1706. Not sure if this is exactly the way you want to fix this. Probably better would be to get rid of the `Decorator` class altogether, but this provides a workable fix. My reasoning on it is this:
* Many places check whether a `Node` is callable by checking whether it is an instance of `FuncBase`.
* Replace these with a `get_callable` method, which will return a `FuncBase` if it makes sense to construct one for the passed in `Node`, otherwise return `None`.
* This method currently only checks if a `Node` is a `FuncBase` or a callable `Decorator`, but could conceivably be expanded to cover other cases with minimal additional work.
* In the case of a `Decorator` return an instance of the `CallableDecorator` class, which inherits from `FuncItem` iff the decorated function is still callable.
This works, but creates problems in a few places in the codebase that explicitly assume that an instance of `FuncBase` is not decorated. I have fixed the instances of this I could find, and all of the existing tests pass, but we probably want to do more testing on this to be sure.

Stray observations:
* It would be nice if we didn't have to treat `property` and similar descriptor-based decorators as special cases and could work with them more directly. Possibly by having their implementations in typeshed be `Generic`s with appropriate types. This would probably require more robust type annotation than we currently support though.
* It may make sense to either rename `Var.is_staticmethod` and `Var.is_classmethod` to `Var.is_static` and `Var.is_class` or rename `FuncBase.is_static` and `FuncBase.is_class` to `FuncBase.is_staticmethod` and `FuncBase.is_classmethod`. They duplicate the same functionality, and this would help remove some unneeded type checks in a few places.
* I experimented with having properties just return their property type, rather than being treated as a special type of callable. This made a few things nicer, in particular fixing the false positive in `testInferListInitializedToEmptyInClassBodyAndOverriden`, but created problems in other places. This may be worth investigating further.
* This change may make it easier to implement properties with different get/set types as suggested in #3004, I will need to investigate further.
* This change probably means that there is a lot of code related to handling `Decorator`s that is no longer needed, but I am not familiar enough with the codebase to say that for sure.
* This fix requires that the decorator is explicitly annotated to return a `Callable`. It seems like this should be something we could determine through type inference, but I'm not sure how easy that would be to do.

I look forward to your criticisms.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3019,10 +3019,16 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                                                    callable_name=fullname)
         self.check_untyped_after_decorator(sig, e.func)
         sig = set_callable_name(sig, e.func)
-        e.var.type = sig
-        e.var.is_ready = True
         if e.func.is_property:
             self.check_incompatible_property_override(e)
+        e.var.type = sig
+        e.var.is_ready = True
+        if isinstance(sig, CallableType):
+            if e.func.is_property:
+                assert isinstance(sig, CallableType)
+                e.is_callable = isinstance(sig.ret_type, CallableType)
+            else:
+                e.is_callable = True
         if e.func.info and not e.func.is_dynamic():
             self.check_method_override(e)
 
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -21,8 +21,8 @@ from mypy.nodes import (
     Context, Decorator, PrintStmt, BreakStmt, PassStmt, ContinueStmt,
     ComparisonExpr, StarExpr, EllipsisExpr, RefExpr, PromoteExpr,
     Import, ImportFrom, ImportAll, ImportBase, TypeAlias,
-    ARG_POS, ARG_STAR, LITERAL_TYPE, MDEF, GDEF,
-    CONTRAVARIANT, COVARIANT, INVARIANT,
+    ARG_POS, ARG_STAR, LITERAL_TYPE, MDEF, GDEF, CallableDecorator,
+    CONTRAVARIANT, COVARIANT, INVARIANT, get_callable
 )
 from mypy import nodes
 from mypy.literals import literal, literal_hash
@@ -1619,11 +1619,15 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         first = base1[name]
         second = base2[name]
         first_type = first.type
-        if first_type is None and isinstance(first.node, FuncBase):
-            first_type = self.function_type(first.node)
+        if first_type is None:
+            method = get_callable(first.node)
+            if method:
+                first_type = self.function_type(method)
         second_type = second.type
-        if second_type is None and isinstance(second.node, FuncBase):
-            second_type = self.function_type(second.node)
+        if second_type is None:
+            method = get_callable(second.node)
+            if method:
+                second_type = self.function_type(method)
         # TODO: What if some classes are generic?
         if (isinstance(first_type, FunctionLike) and
                 isinstance(second_type, FunctionLike)):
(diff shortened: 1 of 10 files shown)


mypy/nodes.py
```diff
@@ -383,7 +383,18 @@ def __str__(self) -> str:
...
-class FuncBase(Node):
+class FuncBaseMeta(type):
+    def __instancecheck__(self, instance: object) -> bool:
+        if isinstance(instance, Decorator):
+            if instance.is_overload:
+                return issubclass(OverloadedFuncDef, self) and instance.is_callable
+            else:
+                return issubclass(FuncBase, self) and instance.is_callable
+        else:
+            return super().__instancecheck__(instance)
```
ilevkivskyi: Sorry, but this will be a big no from me. I don't like `Decorator` (which is an understatement), but this doesn't look like a solution:
* This will make future issues hard to debug.
* This will not work with `mypyc`.

There should be a better solution, which is probably just fixing several `isinstance()` checks for now to account for `Decorator` (also I am not sure there are many of them to be updated if we just want to fix `__init__`).

mypy/semanal.py
```diff
@@ -2518,6 +2518,8 @@ def visit_decorator(self, dec: Decorator) -> None:
                 self.check_decorated_function_is_method('classmethod', dec)
             elif (refers_to_fullname(d, 'builtins.property') or
                   refers_to_fullname(d, 'abc.abstractproperty')):
+                # TODO: Stop treating properties as special cases and treat as generalized
+                # TODO: descriptors
```
ilevkivskyi: This comment is not needed, we already have an issues for this.

TV4Fun: Removed

mypy/nodes.py
TV4Fun: Have implemented an alternate approach. You may like this one better.

mypy/nodes.py
```diff
@@ -704,6 +705,28 @@ def deserialize(cls, data: JsonDict) -> 'Decorator':
         return dec
 
 
+class CallableDecorator(FuncItem):
+    """A wrapper around a Decorator that allows it to be treated as a callable function"""
+    def __init__(self, decorator: Decorator) -> None:
+        super().__init__(decorator.func.arguments, decorator.func.body,
+                         cast('mypy.types.CallableType', decorator.type))
```
ilevkivskyi: This is better but I think this can be made even simpler. I don't see why this class is necessary, moreover instead of one class that I don't like (`Decorator`), we will have two. Instead of trying to solve everything, why not start from a simple approach that fixes the original issues, and then generalize it to a reasonable extent?

mypy/checker.py
```diff
@@ -1619,11 +1619,15 @@ def check_compatibility(self, name: str, base1: TypeInfo,
...
+            method = get_callable(first.node)
+            if method:
+                first_type = self.function_type(method)
         second_type = second.type
-        if second_type is None and isinstance(second.node, FuncBase):
-            second_type = self.function_type(second.node)
+        if second_type is None:
+            method = get_callable(second.node)
+            if method:
+                second_type = self.function_type(method)
```
ilevkivskyi: I am not sure why you need this change, but please take into account that there are _three_ places in `checker.py` that check Liskov: one here (for "horizontal" overriding in multiple inheritance), one for variables, and one for methods etc. (There is some code duplication, I know our code is not perfect.) So please double check if you need to update other two places.

mypy/nodes.py
TV4Fun: The main reason I do it this way is that `FuncBase` and `FuncItem` are both abstract, so I can't instantiate them directly, and I don't know if it makes sense to return a `FuncDef`, since that's not really what this is. I think it makes sense to have a callable decorator be represented as a `FuncBase`, so that code that is looking for something callable can treat it the same as it would any other callable. The alternative is to create a lot of special cases to handle `Decorator`s, which doesn't seem like it serves the long-term interest of eliminating the `Decorator` class altogether.

ilevkivskyi:
Superseded by https://github.com/python/mypy/pull/6645

PR #5800, opened 2018-10-18
Fix variable name in sample code for Mapping

uolot:


docs/source/cheat_sheet.rst
```diff
@@ -205,7 +205,7 @@ that are common in idiomatic Python are standardized.
    def f(my_mapping):
        # type: (MutableMapping[int, str]) -> Set[str]
        my_dict[5] = 'maybe'
-       return set(my_dict.values())
+       return set(my_mapping.values())
```
ilevkivskyi: Shouldn't it be also `my_mapping` on the line above?

PR #5808, opened 2018-10-19
Explain that you can sometimes safely omit __init__ return type

### LINKED ISSUE #5790: Document that None return type is optional for __init__
Also make it clear that this only applies if at least one argument is annotated. If `__init__` only has the `self` argument, the `None` return type is still needed for type checking.

gvanrossum:
Fixes #5790

### DIFF
--- a/docs/source/class_basics.rst
+++ b/docs/source/class_basics.rst
@@ -8,7 +8,7 @@ rules.
 Instance and class attributes
 *****************************
 
-Mypy type checker detects if you are trying to access a missing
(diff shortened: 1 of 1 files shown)


TV4Fun:
Is there a particular reason we're still required to have the `None` return type annotation for methods without any annotated arguments, or is this just a weakness of the type checking system?

gvanrossum:
We need this because of the rule that the body of an unannotated function isn't checked. We don't want this change to suddenly start checking the bodies of previously unchecked `__init__` methods, because that might call a barrage of errors.

PR #5812, opened 2018-10-19
Don't wait for tests before triggering mypyc wheel build

msullivan:
Triggering a build of mypy_mypyc wheels for a broken revision is not
actually a big deal, and waiting for the tests before starting to
build the wheels unnecessarily lengthens the critical path to
deploying a new version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,6 @@ jobs:
     - TOXENV=lint
     - EXTRA_ARGS=
   - name: "trigger a build of wheels"
-    stage: deploy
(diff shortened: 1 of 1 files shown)


msullivan:
This allows it to run in parallel with the rest, though since the number of jobs is limited it doesn't actually *ensure* that it runs immediately (and there is no way to prioritize a job). In practice so far it seems good though: for this job it finished in 2m, which is a big improvement over the 14m it took for the previous change where it had to wait.

PR #5816, opened 2018-10-22
Fix constraint inference for non-invariant instances

ilevkivskyi:


ilevkivskyi:
Closing in favour of https://github.com/python/mypy/pull/5817

Issue #2035, opened 2016-08-18
Cannot infer type argument when mixing int and float


JukkaL:
Mypy complains about this code, even though it looks fine to me:

```
from typing import Sequence, TypeVar

_T = TypeVar('_T')

def bisect_left(a: Sequence[_T], x: _T) -> int: ...

bisect_left([1, 2, 3], 5.391)  #  Cannot infer type argument 1 of "bisect_left"
```

This doesn't seem to be related to promotions, as it also happens with user-defined types:

```
class A: pass
class B(A): pass

bisect_left([B()], A())   # same error
```

matthiaskramm:
Just out of curiosity: What is mypy *supposed* to do, when having

```
def f(x: _T, y: _T)
```

and a call

```
f(A(), B())
```

Should it try to "promote" either A to B or B to A? Or should it try to find a common base class between the two?

Or would the magic that makes bisect_left work not actually apply for the non-sequence case?

JukkaL:
It looks for a value for `_T` that is compatible for all arguments. There are some additional rules, such as that we generally don't infer `Any` types, even though they would be valid everywhere, and we favor narrower types as results. It tries to look for common base classes and it also considers promotions such as `int` -> `float`. In your example, if `B` is a subclass of `A`, so we'd infer `_T = A`. `object` would also be a valid result, but since `A` is narrower we prefer that.

The non-sequence case should work already -- it's the sequence case that is broken.

graingert:
any way of specifying type arguments to a function?

```python
class A: pass
class B(A): pass

bisect_left[A]([B()], A())   # same error
```

JukkaL:
Not directly, but you can use `cast` within the arguments to make all argument types match exactly, so it's obvious what the value of a type argument will be. Example:

```py
from typing import cast

class A: pass
class B(A): pass

bisect_left([cast(A, B())], A())  # ok
```

graingert:
@JukkaL here's an issue about what I mean https://github.com/python/mypy/issues/3096

JukkaL:
Increasing priority since there was a user question that boils down to this issue and the fix seems simple.

Issue #3816, opened 2017-08-09
Require an annotation if inferred list or dict item type is object


JukkaL:
Mypy infers the type `List[object]` for `x` in this example:

```py
x = [1, 'a']
```

This is often not what the user expects and can be confusing, as there tends to be an error when the program tries to do something with the list. Often the user would prefer the type to be `List[Any]` or `List[Union[int, str]]`, but generally the desired item type could be something else, such as an ABC or protocol (once protocols are supported).

It's perhaps impractical to unambiguously infer the desired type, but it would be easy enough to detect cases like these and require an explicit annotation. For the above example, mypy would require an annotation for `x` (and potentially suggest `List[Any]` or `List[Union[int, str]]` as the potential type), similar to how mypy deals with `x = []`. 

For user-defined generic types `object` might be a reasonable value for a type parameter, so we should probably restrict this to things where the `object` type is likely a problem. For example, we could generate the message for the types `List[object]` and `Dict[<anything>, object]` only. It's less clear if we should do this for dictionaries with`object` keys or for sets.

JukkaL:
#4548 has relevant discussion.

JukkaL:
Marking as high priority since this seems to be a common source of confusion.

jxcl:
Working on this.

gvanrossum:
Should this require a flag to enable it? (Or perhaps allow a flag to disable it?)

jxcl:
I'm in support of a flag to disable. It is a backwards incompatible change, but no heavy refactoring is required to fix existing code, only taking the inferred types that the mypy errors generate and applying them to the codebase. 

Disabling by default would not fix the issue of new users being confused by the inference, since they would also presumably not know about the flag.

JelleZijlstra:
#6805 has a similar example where we unexpectedly infer `Dict[str, int]` for essentially `{"a": 1, "b": True}`.

ilevkivskyi:
It would be great to sometimes suggest using `TypedDict`, but such cases are very hard to identify.

DevilXD:
Is this ever gonna be a thing?

JukkaL:
We might want to also do this for sets (#8947). 

Also, there will need to be a way to disable this, but not enabling this by default would be a bit pointless, as discussed above.

ChetanKhanna:
[...]

PR #5817, opened 2018-10-22
Fix constraint inference for non-invariant instances

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/2035

Note that in two tests we now infer `object` instead of giving an error. This may be not what a user expects, but I think this is still OK, and after https://github.com/python/mypy/issues/3816 mypy will always ask for an annotation in such cases.


ilevkivskyi:
Internal codebase is clean.

PR #5822, opened 2018-10-23
Mark raw strings

### LINKED ISSUE #5820: Deprecation warnings in test files
`test-data/stdlib-samples/3.2/fnmatch.py` and `test-data/stdlib-samples/3.2/test/test_fnmatch.py` contain unmarked raw strings that will produce `DeprecationWarning`s when parsed with `typed-ast` and Python 3.7. See for example here: https://travis-ci.org/python/typeshed/jobs/444824883

These are easy to fix (PR is ready), but I wonder whether they should be, since these files seem to be pristine copies from Python 3.2's standard library. On the other hand not fixing them leads to spurious warning output and might even break tests in future Python versions.

srittau:
Fixes DeprecationWarnings. Closes #5820.

### DIFF
--- a/test-data/stdlib-samples/3.2/fnmatch.py
+++ b/test-data/stdlib-samples/3.2/fnmatch.py
@@ -109,4 +109,4 @@ def translate(pat: str) -> str:
                 res = '%s[%s]' % (res, stuff)
         else:
             res = res + re.escape(c)
-    return res + '\Z(?ms)'
+    return res + r'\Z(?ms)'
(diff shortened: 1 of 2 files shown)


srittau:
I see those two problems as well, but I didn't find the cause. I suspected the raw string in mypy/strconv.py: `r'\u%.4x'`. And the only instances of `<>` I found were in regular expressions.

gvanrossum:
This one's good. If there are more issues we can fix them separately. Thanks!

PR #5826, opened 2018-10-23
Sync typeshed

msullivan:


msullivan:
Don't merge this, https://github.com/python/typeshed/pull/2516 will cause mass breakage for python 2 json code.

JelleZijlstra:
I merged the PR fixing the json issue, so should be OK after you update the submodule again.

PR #5827, opened 2018-10-23
Add path expansion for write_junit_xml output argument

EvanKepner:
Hi - this is a small addition to allow the `--junit-xml` arg to accept a string path as well as a file.

Example:

```
$ mypy file1.py --junit-xml=a.xml

$ mypy file1.py --junit.xml "this/is/a/test.xml"
$ tree this/
this/
└── is
    └── a
        └── test.xml

2 directories, 1 file
```

### DIFF
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -150,7 +150,14 @@ def write_junit_xml(dt: float, serious: bool, messages: List[str], path: str) ->
         xml = FAIL_TEMPLATE.format(text=escape('\n'.join(messages)), time=dt)
     else:
         xml = ERROR_TEMPLATE.format(text=escape('\n'.join(messages)), time=dt)
-    with open(path, 'wb') as f:
+
+    # checks for a directory structure in path and creates folders if needed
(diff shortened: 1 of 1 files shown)


gvanrossum:
You can write `--junit-xml ~/outputs/file.xml`, so there is no need to expand `~/`. Please commit the change.

EvanKepner:
Thanks for the guidance - in my testing I was passing in the argument as a string with quotes which was my error, when I pass in the argument without the string quotes it operates as expected without the expand-user. Commit made, appreciate the review!

msullivan:
I think we probably don't want `expandvars` either. Shells will do this for us, and we don't do it for any other filenames.

gvanrossum:
> I think we probably don't want expandvars either. Shells will do this for us, and we don't do it for any other filenames.

Indeed, that's what I meant, sorry.

EvanKepner:
No problem at all, changes made.

gvanrossum:
Thanks!

EvanKepner:
Absolutely, thanks for the feedback and accepting!

Issue #5667, opened 2018-09-25
Using os.PathLike as an annotation


scolby33:
I am trying to create a typealias for all the types that can be passed as the first argument of `open()`. MyPy reveals the type to be `Union[builtins.str, builtins.bytes, builtins.int, builtins._PathLike[Any]]`.

My first attempt at this is:
```python
Openable = Union[str, bytes, int, os.PathLike[Any]]
```

MyPy is happy with this, but alas Python is not:
```
TypeError: 'ABCMeta' object is not subscriptable
```

Trying this, then:
```python
Openable = Union[str, bytes, int, os.PathLike]
```

But now MyPy is unhappy:
```
error: Missing type parameters for generic type
```

What is the correct way of using `os.PathLike` here? There is no `typing.PathLike` yet. Do I have to create my own generic version of `os.PathLike`?

ilevkivskyi:
Another solution is to use `if TYPE_CHECKING: ...`.

This problem appeared several times. It would be great to create a documentation entry about this.

scolby33:
Oh of course, how could I have forgotten about the string version of annotations!?

+1 to a mention in documentation

ilevkivskyi:
Another example is when the class that is generic in typeshed is not generic at runtime appears as a base class. This case is even more important, since one can't use quotes and `if TYPE_CHECKING: ...` is the only way.

PR #5833, opened 2018-10-24
Document what to do if a class is not generic at runtime

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5667

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -499,10 +499,9 @@ Here's the above example modified to use ``MYPY``:
 Using classes that are generic in stubs but not at runtime
 ----------------------------------------------------------
 
-In some cases a class may be declared as generic in stubs, while
-it is not generic at runtime. For example this is the case for some
-classes in standard library, such as ``os.PathLike`` and ``queue.Queue``.
-Subscripting such class will result in runtime errors:
+Some classes are declared as generic in stubs, but not at runtime. Examples
+in the standard library include ``os.PathLike`` and ``queue.Queue``.
+Subscripting such a class will result in a runtime error:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


docs/source/common_issues.rst
```diff
@@ -496,6 +496,42 @@ Here's the above example modified to use ``MYPY``:
...
+
+   from queue import Queue
+
+   class Tasks(Queue[str]):  # TypeError: 'type' object is not subscriptable
+       ...
+
+   results: Queue[int] = Queue()  # TypeError: 'type' object is not subscriptable
+
+To avoid these errors while still having precise types one can either use
+string literals or ``typing.TYPE_CHECKING``:
```
JukkaL: "string literals" is a bit vague. Maybe reword as "string literal types".

docs/source/common_issues.rst
```diff
@@ -496,6 +496,42 @@ Here's the above example modified to use ``MYPY``:
...
+.. code-block:: python
+
+   from queue import Queue
+
+   class Tasks(Queue[str]):  # TypeError: 'type' object is not subscriptable
+       ...
+
+   results: Queue[int] = Queue()  # TypeError: 'type' object is not subscriptable
+
+To avoid these errors while still having precise types one can either use
```
JukkaL: Style nit: use "you" instead of "one" for consistency.

docs/source/common_issues.rst
```diff
@@ -496,6 +496,42 @@ Here's the above example modified to use ``MYPY``:
...
+string literals or ``typing.TYPE_CHECKING``:
+
+.. code-block:: python
+
+   from queue import Queue
+   from typing import TYPE_CHECKING
+
+   if TYPE_CHECKING:
+       BaseQueue = Queue[str]
```
JukkaL: Maybe add descriptive comment here -- this is only processed by mypy.

docs/source/common_issues.rst
```diff
@@ -496,6 +496,42 @@ Here's the above example modified to use ``MYPY``:
...
+   else:
+       BaseQueue = Queue
```
JukkaL: Another descriptive comment could be helpful here -- this is not seen by mypy but will be executed at runtime.

ilevkivskyi:
@JelleZijlstra @JukkaL thanks for review! I pushed an update.

PR #5840, opened 2018-10-25
Merge --add into --update for daemon

ilevkivskyi:
This simplifies some logic in code and also is easier to use. I tested this manually and it seem to work well.


gvanrossum:
@mjsullivan Wait for this with any pin moves please.

PR #5850, opened 2018-10-28
Improve stub files generated by stubgenc for pybind11 modules

wiktorn:
Following changes fixes stubgenc that generates invalid *pyi files in following cases:

1. Class constructors may sometimes get Any type as return value. After this change - for __init__ None is the default
2. For modules using pybind11, base class is pybind11_object which is non-existent (see #5814). After this change pybind11_object is removed
3. Stubgenc doesn't add any import statements other than referencing some typing.* objects. If module references other types in different modules/packages no import statement is added. This pull requests adds import statements to referenced types

I'm not sure about 2, maybe pybind11 library should provide dummy pybind11_object in its module and annotate types, so dummy object will be assumed as superclass.


gvanrossum:
Please fix the tests (see https://travis-ci.org/python/mypy/jobs/447561624).

Also add some tests for the new functionality. We really need to have more
tests for stubgenc.

mypy/stubgenc.py
```diff
@@ -209,6 +241,9 @@ def generate_c_type_stub(module: ModuleType,
     if all_bases[-1] is object:
         # TODO: Is this always object?
         del all_bases[-1]
+    # remove pybind11_object
```
gvanrossum: Really? If `pybind11` is really that unique, please elaborate more in the comment.

mypy/stubgenc.py
```diff
@@ -217,7 +252,12 @@ def generate_c_type_stub(module: ModuleType,
         if not any(issubclass(b, base) for b in bases):
             bases.append(base)
     if bases:
-        bases_str = '(%s)' % ', '.join(base.__name__ for base in bases)
+        bases_str = '(%s)' % ', '.join(
+            strip_or_import(
+                '%s.%s' % (base.__module__, base.__name__),
+                module,
+                imports
+            ) for base in bases)
```
gvanrossum: I'd move this `)` to the previous line -- somehow for function calls (unlike list/dict displays) we like the close parenthesis right after the last argument. But the `for` clause should stay on its own line.

mypy/test/teststubgen.py
```diff
@@ -239,7 +239,8 @@ def test_infer_unary_op_sig(self) -> None:
     def test_generate_c_type_stub_no_crash_for_object(self) -> None:
         output = []  # type: List[str]
         mod = ModuleType('module', '')  # any module is fine
-        generate_c_type_stub(mod, 'alias', object, output)
+        imports = []  # type: List[str]
+        generate_c_type_stub(mod, 'alias', object, output, imports)
```
gvanrossum: Shouldn't you check the contents of `imports` after the call returns?

mypy/test/teststubgen.py
```diff
@@ -248,6 +249,45 @@ class TestClassVariableCls:
             x = 1
 
         output = []  # type: List[str]
+        imports = []  # type: List[str]
         mod = ModuleType('module', '')  # any module is fine
-        generate_c_type_stub(mod, 'C', TestClassVariableCls, output)
+        generate_c_type_stub(mod, 'C', TestClassVariableCls, output, imports)
```
gvanrossum: (Here too.)

mypy/stubgenc.py
wiktorn: I provided more specific information about this in comment. But this is part that I have doubts if this is the way it should be solved. Maybe better approach would be for pybind11 to point to an type that is actually importable.

wiktorn:
I hope that I correctly identified changes that you've requested. The only thing that may require further discussion is handling of `pybind11_object`.

mypy/stubgenc.py
```diff
@@ -154,23 +154,35 @@ def generate_c_function_stub(module: ModuleType,
 
     if sig:
         sig_types = []
-        for arg in (arg.split(':', 1) for arg in sig.split(',')):
-            if len(arg) == 1:
-                sig_types.append(arg[0].strip())
+        # convert signature in form of "self: TestClass, arg0: str" to list [[self, TestClass], [arg0, str]]
+        for arg_type in (arg.split(':', 1) for arg in sig.split(',')):
```
gvanrossum: Please rewrite this as follows:
```py
# Convert signature in form of "self: testmod.TestClass, arg0: str" to list ["self: testmod.TestClass", "arg0: str"]
# and update imports with relevant modules (e.g. "testmod").
for arg in sig.split(','):
    arg_type = arg.split(':', 1)
    if len(arg_type) == 1:
    # etc.
```

gvanrossum:
Thank you!

Issue #5851, opened 2018-10-29
Assignment to an instance backed by a descriptor doesn't discard optional


gvanrossum:
Universal example:
```py
from typing import Type, TypeVar, Optional
T = TypeVar('T')

class IntDescr:
  def __get__(self, obj: T, typ: Type[T]) -> Optional[int]: ...
  def __set__(self, obj: T, value: Optional[int]) -> None: ...

class C:
  reg = 0
  spec = IntDescr()

  def meth_reg(self) -> int:
    if self.reg is None:
      self.reg = 0
    return self.reg    # OK

  def meth_spec(self) -> int:
    if self.spec is None:
      self.spec = 0
    # assert self.spec is not None
    return self.spec    # Error: Incompatible return value type (got "Optional[int]", expected "int")
```
If I uncomment the `assert` on the line before the error, the error goes away. I suspect that something is missing for assignment to descriptors. Or perhaps this is intentional, because we can't be *sure* that `__set__` doesn't just throw the value away? Then the assert would be an unfortunate workaround needed 99% of the time.

ilevkivskyi:
I don't think this is intentional. I think this is a consequence of the fact that `__set__` is analyzed in `checker`, but it really should be done in `checkmember` (with `is_lvalue=True`) so that all normal mechanisms will kick in.

ilevkivskyi:
I actually always wanted to refactor this part, I will try it now.

ilevkivskyi:
OK, so this may be not so easy. There is a test `DescriptorGetSetDifferentTypes` that will fail if we will always use binder for descriptors. So this requires a bit more care, I will try to think a bit more.

gvanrossum:
What about `__set__` implementations that legitimately normalize some non-`None` values to `None`? E.g.
```py
def __set__(self, obj, value) -> None:
    if not value: value = None  # Get rid of 0, '', (), [], {}
    obj._value = value
```

ilevkivskyi:
The same applies to binder in general. Binder is clearly unsafe, there are several scenarios where it assigns a "too optimistic" type (and there are issues about this). The point is that in real life binder removes lots of false positives by adding few false negatives. I think this is one of such cases.

ilevkivskyi:
(Reducing the priority, since there is not obviously right solution, but anyway, I have a fix, will make a PR shortly.)

PR #5853, opened 2018-10-29
Make descriptor assignments more like normal assignments

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5851

This makes descriptor assignments behave more like normal assignments. This includes:
* The same error message as for normal assignment
* Using binder to restrict types (but under a bit more strict condition)

I was not able to refactor all the logic to `checkmember.py`, but this is probably OK. Also this PR can introduce potential unsafety as discussed in the issue, but this is probably not much different from using binder in some other situations.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1787,7 +1787,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                 elif (isinstance(lvalue, MemberExpr) and
                         lvalue.kind is None):  # Ignore member access to modules
                     instance_type = self.expr_checker.accept(lvalue.expr)
-                    rvalue_type, infer_lvalue_type = self.check_member_assignment(
+                    rvalue_type, lvalue_type, infer_lvalue_type = self.check_member_assignment(
                         instance_type, lvalue_type, rvalue, lvalue)
                 else:
                     rvalue_type = self.check_simple_assignment(lvalue_type, rvalue, lvalue)
@@ -2491,7 +2491,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             return rvalue_type
 
     def check_member_assignment(self, instance_type: Type, attribute_type: Type,
-                                rvalue: Expression, context: Context) -> Tuple[Type, bool]:
+                                rvalue: Expression, context: Context) -> Tuple[Type, Type, bool]:
         """Type member assignment.
 
         This defers to check_simple_assignment, unless the member expression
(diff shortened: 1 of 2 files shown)


test-data/unit/check-classes.test
```diff
@@ -1383,9 +1383,9 @@ class A:
     g = D('10')
 a = A()
 a.f = 1
-a.f = '' # E: Argument 2 to "__set__" of "D" has incompatible type "str"; expected "int"
+a.f = '' # E:  Incompatible types in assignment (expression has type "str", variable has type "int")
```
gvanrossum: Looks like you have an extra space here and below.

mypy/checker.py
```diff
@@ -2491,59 +2491,67 @@ def check_simple_assignment(self, lvalue_type: Optional[Type], rvalue: Expressio
...
             # A message already will have been recorded in check_call
-            return AnyType(TypeOfAny.from_error), False
-
-        return inferred_dunder_set_type.arg_types[1], False
+            return AnyType(TypeOfAny.from_error), get_type, False
+
+        set_type = inferred_dunder_set_type.arg_types[1]
+        # Special case: if the rvalue_type is a subtype of both '__get__' and '__set__' types,
+        # then we invoke the binder to narrow type by this assignment. Technically, this is not
+        # safe, but in practice this is what a user expects.
```
gvanrossum: I'm still worried about edge cases here. E.g. what about this pair:
```py
def __get__(self, obj, typ) -> float: return self._val
def __set__(self, obj, value) -> None: self._val = float(value)
```
and then the usage
```py
if not isinstance(a.foo, int):
    a.foo = int(a.foo)
blah[a.foo]  # Suppose blah is a List, so the index must be int
```
Perhaps I would worry less if we only went after real unions here, not subtypes. (Though in most cases subtypes will also work -- it's really mostly float/int, where casting int to float changes the representation.)

ilevkivskyi: Yes, this is clearly a trade off. @JukkaL what do you think?

mypy/checker.py
```diff
@@ -2491,59 +2491,67 @@ def check_simple_assignment(self, lvalue_type: Optional[Type], rvalue: Expressio
...
+            return AnyType(TypeOfAny.from_error), get_type, False
 
         function = function_type(dunder_set, self.named_type('builtins.function'))
         bound_method = bind_self(function, attribute_type)
         typ = map_instance_to_supertype(attribute_type, dunder_set.info)
         dunder_set_type = expand_type_by_instance(bound_method, typ)
 
         _, inferred_dunder_set_type = self.expr_checker.check_call(
-            dunder_set_type, [TempNode(instance_type), rvalue],
+            dunder_set_type, [TempNode(instance_type), TempNode(AnyType(TypeOfAny.special_form))],
```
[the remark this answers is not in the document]
ilevkivskyi: Initially I thought this is is probably OK, but now I am not so sure. Anyway, to be on the safe side, we can check the call _twice_. Once we use the real expression but turn off errors, just to infer the type (if it is invalid, we will later show a nice "Invalid type in assignment..." error later). Second time we check with the `Any` type to show the remaining `___set__` specific errors (like wrong type of first argument).

mypy/checker.py
```diff
@@ -2491,59 +2491,67 @@ def check_simple_assignment(self, lvalue_type: Optional[Type], rvalue: Expressio
...
 
     def check_member_assignment(self, instance_type: Type, attribute_type: Type,
-                                rvalue: Expression, context: Context) -> Tuple[Type, bool]:
+                                rvalue: Expression, context: Context) -> Tuple[Type, Type, bool]:
         """Type member assignment.
 
         This defers to check_simple_assignment, unless the member expression
         is a descriptor, in which case this checks descriptor semantics as well.
 
         Return the inferred rvalue_type and whether to infer anything about the attribute type.
```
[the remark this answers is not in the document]
ilevkivskyi: It is only for binder, I updated this.

mypy/checker.py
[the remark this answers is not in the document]
ilevkivskyi: (The above is what I already did in the latest commit.)

mypy/checker.py
ilevkivskyi: This is a bit complicated because of the overload case. I propose to instead to apply binder only if "get" type is a subtype of "set" type, and the rvalue type is a subtype of "get" type. This covers the situation when the "get" and "set" types are the same. And in general looks reasonably safe. I added this in latest commit.

ilevkivskyi:
@JukkaL Thanks for review! I addressed your comments.

ilevkivskyi:
@JukkaL Are you happy with the updates I made?

PR #492, opened 2014-11-04
Mention pip in Quick start.

lrem:
It would be beneficial to mention ease of installation in the README.

PR #5859, opened 2018-10-31
Add support for dmypy on Windows

### LINKED ISSUE #5019: Add support for mypy daemon on Windows
There is a major limitation for mypy daemon and fine-grained incremental mode: mypy daemon doesn't  work on Windows, see http://mypy.readthedocs.io/en/latest/mypy_daemon.html#limitations

In particular, `dmypy_server.daemonize` uses some functionalities not available on Windows (such as `os.fork`). It would be great to remove this limitation (and update Appveyor to test the support).

@ethanhs It looks like you have most experience on Windows among all of us, would you like to take this task?

ethanhs:
After a while of banging my head against the Windows APIs today, I finally have most of dmypy running on Windows (wow this makes rechecks fast!).

~~Here is what works and what doesn't~~ 
Everything that is publicly documented should work now:
- [x] `dmypy start`
- [x] `dmypy stop`
- [x] `dmypy restart`
- [x] `dmypy run`
- [x] `dmypy kill`
- [x] `dmypy check`
- [x] `dmypy status`

And here is a photo of it working :)
image

~~This depends on https://github.com/python/typeshed/pull/2571. Until then selfcheck will fail (on Windows).~~

Closes #5019


mypy/dmypy_server.py
```diff
@@ -465,4 +558,17 @@ def get_meminfo() -> Dict[str, Any]:
         meminfo = process.memory_info()
         res['memory_rss_mib'] = meminfo.rss / MiB
         res['memory_vms_mib'] = meminfo.vms / MiB
+        if sys.platform == 'win32':
```
ethanhs: @gvanrossum I wasn't entirely sure what to do here, so I refactored things to be a bit more reliant on psutil existing. I could manually get the max memory used with win32 APIs if psutil isn't installed if you prefer, but I felt this was cleaner.

ethanhs:
Okay, I tested things on 3.4 and 3.7 and it works, so this should be ready for review.

ethanhs:
>Sorry, I don't think I can finish this review.

No problem, glad to get some feedback!

@msullivan happy to talk on Friday about this if you want.

mypy/dmypy.py
```diff
@@ -20,6 +22,17 @@
...
+if sys.platform == 'win32':
+    # This may be private, but it is needed for IPC on Windows, and is basically stable
+    import _winapi
+    import ctypes
+    from ctypes import wintypes
+
+    from mypy.dmypy_util import write_file
+
+    OpenProcess = ctypes.windll.kernel32.OpenProcess
+    GetExitCodeProcess = ctypes.windll.kernel32.GetExitCodeProcess
```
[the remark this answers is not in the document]
ethanhs: I did as best I think I can at the moment. Unfortunately, the ctypes stubs are a bit finnicky, and things like `ctypes.byref` don't do the right thing at the moment, so I stuck with `int`s and `Any, though I used a couple things like `HANDLE` where appropriate.

mypy/dmypy_server.py
```diff
@@ -153,14 +179,84 @@ def __init__(self, options: Options,
 
     def serve(self) -> None:
         """Serve requests, synchronously (no thread or fork)."""
+        if sys.platform == 'win32':
+            self.serve_win()
+        else:
+            self.serve_nix()
+
+    def serve_win(self) -> None:
+        if sys.platform == 'win32':
```
msullivan: Does `assert sys.platform == 'win32'` work? That would avoid all this indenting.

mypy/dmypy.py
```diff
@@ -181,14 +194,19 @@ def start_server(args: argparse.Namespace, allow_sources: bool = False) -> None:
...
+            sys.exit(1)
+    else:
+        if daemonize(Server(start_options,
+                            timeout=args.timeout).serve,
+                    args.log_file) != 0:
+            sys.exit(1)
     wait_for_server()
 
 
-def wait_for_server(timeout: float = 5.0) -> None:
```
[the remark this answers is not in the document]
ethanhs: Ah, I can probably change this back. My initial implementation had a longer timeout because I was doing something incorrectly which caused things to wait for a while, but things should connect in 5 seconds or less now.

ethanhs:
>I think duplicating the serve function will lead to mistakes and bugs done the road (and the windows version getting broken)

Yeah, I can definitely understand that. The reason I chose NamedPipes was twofold: a) it is what Steve Dower recommended and b) its what mostly everyone else uses (at least from what I have seen). I am however not totally married to using them.

>I see two options here
>My inclination is that we should do the latter, unless there is a major advantage of named pipes that I'm missing?

Yeah, that is okay with me. I think moving entirely to TCP sockets makes the most sense, because then we can just share all IPC code across platforms, which makes the odds of things breaking very low.

mypy/dmypy_server.py
msullivan: Does ```
if sys.platform != 'win32':
    return
``` work?

**Comment by msullivan** [comment]
> > I think duplicating the serve function will lead to mistakes and bugs done the road (and the windows version getting broken)
> 
> Yeah, I can definitely understand that. The reason I chose NamedPipes was twofold: a) it is what Steve Dower recommended and b) its what mostly everyone else uses (at least from what I have seen). I am however not totally married to using them.
> 
Yeah, in a very real sense they seem like they are the Right Thing for the job. I think the advantage of NamedPipes (and unix domain sockets) is that they have access control on them. This can matter on multi-user machines, but I think that multi-user windows dev machines aren't really a thing?

> > I see two options here
> > My inclination is that we should do the latter, unless there is a major advantage of named pipes that I'm missing?
> 
> Yeah, that is okay with me. I think moving entirely to TCP sockets makes the most sense, because then we can just share all IPC code across platforms, which makes the odds of things breaking very low.

I am somewhat inclined to keep using unix domain sockets when available, since multi-user Linux machines do still exist? It does mean some more conditional code, though I think a pretty small amount (just starting to listen and establishing the connection; all the communication should be the same).

**Comment by msullivan** [comment]
> I am somewhat inclined to keep using unix domain sockets when available, since multi-user Linux machines do still exist? It does mean some more conditional code, though I think a pretty small amount (just starting to listen and establishing the connection; all the communication should be the same).

Though maybe this is a sign that we should stick with NamedPipe also? Though I think it does need to be abstracted out into a set of communication routines.

**Comment by ethanhs** [comment]
>Though maybe this is a sign that we should stick with NamedPipe also? Though I think it does need to be abstracted out into a set of communication routines.

I think in the interest of not making assumptions about a user's environment, erring on the side of caution makes sense here. While multi-user Windows computers are much less common, I don't think we should preclude them. Changes to the actual IPC is likely going to be pretty rare, so I think a wrapper makes the most sense. It gives the added bonus that we can change the underlying implementation later on without needing to change any of the dmypy code.

**Comment by gvanrossum** [comment]
I didn't read the code, but I do agree that we should stick with unix domain sockets and named pipes, because of the access control they have. It shouldn't be hard to have a simple abstraction on top of these. (Though perhaps there might be a 3rd party package that does this?)

**Comment by ethanhs** [comment]
@msullivan okay so here is the refactored version. I pulled all the IPC into `mypy.ipc`, hopefully that looks reasonable, and hopefully the context managers don't make mypyc _too_ unhappy, they make the code much nicer to read and abstract away all the messy details.

**Comment by msullivan** [comment]
The approach looks good, I'll try to take a more detailed look tomorrow.

One question (and maybe I'm just missing something about either the code or namedpipes): is it possible on windows to run two daemons at once? On unix we generate different temp directories which gives us different paths, but it looks like there is just one path always used. Running multiple daemons at one (on different source trees) is an important use case.

**Comment by ethanhs** [comment]
@msullivan added. I also noticed there could be a race on the options file so I made that use a temp file as well.

**Comment by msullivan** [comment]
I also think that all the 'win32' checks are going to cause mypyc a bunch of grief (on all platforms), since it plays pretty badly with platform checks. I will poke around at that myself and give some advice.

Pickling up `Options` will also break under mypyc, since mypyc-compiled objects aren't pickleable, but I'm happy to say we'll burn that bridge when we get to it.

File: mypy/dmypy.py [hunk 5]
```diff
@@ -93,7 +105,7 @@ def __init__(self, prog: str) -> None:
                help="Server shutdown timeout (in seconds)")
 p.add_argument('flags', metavar='FLAG', nargs='*', type=str,
                help="Regular mypy flags (precede with --)")
-
+p.add_argument('--options-file', help=argparse.SUPPRESS)
```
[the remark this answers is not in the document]
ethanhs: If it is impossible to pass the `Options` as a pickled object anyway, perhaps I could pass an `Options.snapshot` (which is just a dict image of the class) instead of the class itself. As for passing the pickled object via an argument, that seems plausible, I will experiment.

mypy/dmypy_server.py
```diff
@@ -34,62 +37,79 @@
 if MYPY:
     from typing_extensions import Final
 
-
 MEM_PROFILE = False  # type: Final  # If True, dump memory profile after initialization
 
+if sys.platform == 'win32':
+    def daemonize(opts: Options, timeout: Optional[int] = None) -> int:
```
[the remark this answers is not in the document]
ethanhs: Ah, yeah I just need to add this to the pickled data and have the server handle this.

msullivan:
I created #492 to fix most of the mypyc related issues and just changes to this PR to fix the one remaining one (conditional defs not working in classes). So we should be good on mypyc related issues for now.

mypy/ipc.py
```diff
@@ -0,0 +1,181 @@
...
+                break
+            bdata.extend(more)
+        return bytes(bdata)
+
+    def write(self, data: bytes) -> None:
+        """Write bytes to an IPC connection."""
+        if sys.platform == 'win32':
+            try:
+                _winapi.WriteFile(self.connection, data)
+                _winapi.WriteFile(self.connection, b'')
```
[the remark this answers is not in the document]
ethanhs: No, its to emulate sendall from sockets, which sends an empty string when it is at the end of the buffer (see also receive in dmypy_util). I will add a comment.

mypy/ipc.py
```diff
@@ -0,0 +1,181 @@
...
+    """Base class for communication between the dmypy client and server.
+
+    This contains logic shared between the client and server, such as reading
+    and writing.
+    """
+
+    connection = None  # type: _IPCHandle
+
+    def __init__(self, name: str) -> None:
+        self.READ_SIZE = 100000
```
msullivan: Could you do a test where the errors are longer than the read size?

ethanhs: Sure! I should probably add some integration tests for IPC anyway.

ethanhs: I made a test that the IPC works as expected, including for when the message length is over the `READ_SIZE` (which is also over the maximum buffer size for named pipes).

ethanhs: Actually, I seem to have an issue with unique connection names, so I will have to modify the test a bit.

ethanhs:
Hm, I will have to figure out why the Linux tests are failing tomorrow.

mypy/ipc.py
msullivan: I actually just meant to go test it manually once, but this is even better!

mypy/test/testipc.py
```diff
@@ -0,0 +1,36 @@
...
+        server.cleanup()
+
+
+class IPCTests(TestCase):
+    def test_transaction_large(self) -> None:
+        pid = os.getpid()
+        msg = 't' * 100001  # longer than the max read size of 100_000
+        p = Process(target=server, args=(msg, pid), daemon=True)
+        p.start()
+        time.sleep(1)  # wait for server to get set up
```
msullivan: Solve, don't evade, concurrency problems.

Probably the thing to do is have a `Semaphore` or `Event` that gets signaled once the server is ready.

msullivan:
There was a typo that kept the daemon from launching on unix (`Server(options, timeout).serve` instead of `Server(options, timeout).serve()`). I went and fixed that, though in the roundabout way of restoring the old callback-based `daemonize` as `_daemonize_cb` and making `daemonize` call it.

msullivan:
> Hm, I will have to figure out why the Linux tests are failing tomorrow. The server should have created the proper directory and socket...

The server puts the socket in a tmp directory, and it needs to communicate the real path back to the client so that it can connect.

mypy/test/testipc.py
ethanhs: Oh yes, this was just a test to see if that was the root cause, and it was not. What I really should do is pass the connection name from the server process to the client/test process to set up the client correctly.

mypy/ipc.py
ethanhs: Well, I think it is valuable to have tests, and it appears they were worth it :)

ethanhs:
@msullivan okay, got the tests working nicely now. Ended up using a `multiprocessing.Queue` to pass the connection name, which incidentally acts as a nice signal that the client can connect :)

mypy/ipc.py
```diff
@@ -0,0 +1,185 @@
...
+        return self
+
+
+class IPCServer(IPCBase):
+
+    BUFFER_SIZE = 2**16
+
+    def __init__(self, name: str, timeout: Optional[int] = None) -> None:
+        super().__init__(name)
+        if sys.platform != 'win32':
```
msullivan: Let's unify the interfaces by making the server always be responsible for generating a unique socket name. On windows we can probably just grab some bytes from `os.urandom()` and call it a day.

mypy/ipc.py
```diff
@@ -0,0 +1,185 @@
...
+            if self.connection == -1:  # INVALID_HANDLE_VALUE
+                err = _winapi.GetLastError()
+                raise IPCException('Invalid handle to pipe: {err}'.format(err))
+            # NOTE: It is theoretically possible that this will hang forever if the
+            # client never connects, though this can be "solved" by killing the server
+            try:
+                _winapi.ConnectNamedPipe(self.connection, _winapi.NULL)
+            except WindowsError as e:
+                if e.winerror == _winapi.ERROR_PIPE_CONNECTED:
+                    pass  # The client already exists, which is fine.
```
msullivan: Do you mean to throw in all other cases? This does a check but ignores everything regardless.

ethanhs: Thanks I fixed this and a case below.

mypy/ipc.py
```diff
@@ -0,0 +1,191 @@
...
+        if sys.platform == 'win32':
+            # The NamedPipe needs to be recreated every time we wait for a connection
+            self.connection = _winapi.CreateNamedPipe(self.name,
+                _winapi.PIPE_ACCESS_DUPLEX | _winapi.FILE_FLAG_FIRST_PIPE_INSTANCE,
+                _winapi.PIPE_READMODE_MESSAGE | _winapi.PIPE_TYPE_MESSAGE | _winapi.PIPE_WAIT,
+                1,  # one instance
+                self.BUFFER_SIZE,
+                self.BUFFER_SIZE,
+                0,  # PIPE_WAIT, wait for a connection of the client
+                0,  # PIPE_ACCEPT_REMOTE_CLIENTS (this is the default, so I guess?)
```
msullivan: `PIPE_ACCEPT_REMOTE_CLIENTS` seems to actually be a flag for the third arg, not the last one. (And probably we should *not* accept remote clients, I would think?)

The last arg seems to be `lpSecurityAttributes`, and I'm having trouble figuring out if the default if `NULL` is passed is what we want: "The ACLs in the default security descriptor for a named pipe grant full control to the LocalSystem account, administrators, and the creator owner. They also grant read access to members of the Everyone group and the anonymous account."

(Discussed in https://docs.microsoft.com/en-us/windows/desktop/ipc/named-pipe-security-and-access-rights)

It's not completely clear to me what exactly the implications of "read access to everyone" is? Without being to write, it can't cause any typechecking, but if it can connect and hold open a connection it could hang the server?

mypy/dmypy.py
```diff
@@ -7,16 +7,21 @@
...
 import sys
 import time
 
-from typing import Any, Callable, Dict, List, Mapping, Optional, Tuple
+from typing import Any, Callable, Dict, Mapping, Optional, Tuple
 
 from mypy.dmypy_util import STATUS_FILE, receive
+from mypy.dmypy_os import alive, kill
+from mypy.ipc import IPCClient, IPCException
+from mypy.options import Options
```
msullivan: Can this be conditionally imported at its use-site, to speed things up?

mypy/ipc.py
ethanhs: Yes, it could be a DOS, since the maximum clients is 1. Looks like I will need to break out some ctypes for this and `DisconnectNamedPipe`.

ethanhs:
Hm, no idea why the address would not be found in the tests, they are passing locally. I will investigate more tomorrow.

ethanhs:
I'm chasing a flake in the end-to-end daemon tests on Windows, so I will be causing a bunch of commits trying to find what is possibly a race condition, sorry!

ethanhs:
Okay, I have run @msullivan's patch to fix the race condition 9 times on Appveyor and it passed every time. So I think we are finally rid of the flake caused by the race condition 🎉

mypy/dmypy.py
```diff
@@ -254,7 +257,7 @@ def do_status(args: argparse.Namespace) -> None:
     if args.verbose or 'error' in response:
         show_stats(response)
     if 'error' in response:
-        sys.exit("Daemon is stuck; consider %s kill" % sys.argv[0])
+        sys.exit("Daemon is stuck; consider %s kill" % (sys.argv[0]))
```
msullivan: Nit: lose the extra parens that were grown while debugging

mypy/dmypy_server.py
```diff
@@ -189,40 +207,31 @@ def serve(self) -> None:
...
+                server.cleanup()  # try to remove the socket dir on Linux
+            except OSError:
+                pass
+            # If the final command is something other than a clean
+            # stop, remove the status file. (We can't just
+            # simplify the logic and always remove the file, since
+            # that could cause us to remove a future server's
+            # status file.)
+            if command != 'stop':
+                os.unlink(STATUS_FILE)
```
msullivan: I would do this before the cleanup, to avoid leaving a reference to a path that has been deleted?
(Though I doubt it really matters.)

mypy/ipc.py
```diff
@@ -0,0 +1,216 @@
...
+        return False
+
+
+class IPCServer(IPCBase):
+
+    BUFFER_SIZE = 2**16
+
+    def __init__(self, name: str, timeout: Optional[int] = None) -> None:
+        if sys.platform == 'win32':
+            name = r'\\.\pipe\{}-{}.pipe'.format(name, os.getpid())
```
msullivan: Instead of using the pid, it might be better to use some (base64d?) bytes out of `os.urandom`? It doesn't *really* matter for our uses but makes it a nicer bit of library code, at least, since then one process can have multiple servers.

test-data/unit/daemon.test
```diff
@@ -29,15 +29,15 @@ def f(): pass
 $ dmypy run -- foo.py --follow-imports=error
 Daemon started
 $ dmypy run -- foo.py --follow-imports=error
-$ echo '[mypy]' >mypy.ini
-$ echo 'disallow_untyped_defs = True' >>mypy.ini
+$ {python} -c "print('[mypy]')" >mypy.ini
```
msullivan: Sigh.

docs/source/mypy_daemon.rst
```diff
@@ -103,5 +99,3 @@ Limitations
   limitation. This can be defined
   through the command line or through a
   :ref:`configuration file <config-file>`.
-
-* Windows is not supported.
```
msullivan: I think that timeouts don't work on windows, so we should document that.

And file a follow-up bug. I think it is doable, but will require using overlapped IO and WaitForSingleObject. So probably something of a pain. No need to worry about it now.

test-data/unit/daemon.test
ethanhs: Yeah this is unfortunate :(

Sadly I couldn't find an echo call that made both shells happy.

mypy/ipc.py
ethanhs: Yeah, good point. I've thought about the idea of breaking out the ipc module into its own library, I'm sure there are others who might be interested in this code.

PR #5861, opened 2018-10-31
Add missing dependency from missing module attribute

### LINKED ISSUE #4782: Attributes are not reprocessed in fine grained mode with `# type: ignore`
This test case currently fails:
```
[case testIgnoredAttrReprocessedModule]
import a
[file a.py]
import b
x = b.x  # type: ignore
y: int = x
[file b.py]
import c
[file b.py.2]
import c
x = c.x
[file c.py]
x: str
[out]
==
a.py:3: error: Incompatible types in assignment (expression has type "str", variable has type "int")
```
The problem is that since there are no errors in `a.py` on the first run it is not reprocessed, and since `b.x` is invalid at the first run no deps of the form `<c.x> -> <a>` or `<b.x> -> <a>` are generated. Therefore `a.py` is never reprocessed, and the error is not caught, here is the debug output:
```
<a.x>
    a
<a.y>
    a
<a>
    __main__
<b>
    a
<c.x>
    c
<c>
    b

==== update 'b' ====
--- update single 'b' ---
triggered: ['<b.x>', '<b[wildcard]>']
triggered: []
update once: b in 0.001s - 0 left
Expected:
  ==
  a.py:3: error: Incompatible types in assignment (expression has type "str", variable has type "int") (diff)
Actual:
  ==
```

JukkaL:
Fixes #4782.

### DIFF
--- a/mypy/server/deps.py
+++ b/mypy/server/deps.py
@@ -604,6 +604,10 @@ class DependencyVisitor(TraverserVisitor):
                 # No type available -- this happens for unreachable code. Since it's unreachable,
                 # it wasn't type checked and we don't need to generate dependencies.
                 return
+            if isinstance(e.expr, RefExpr) and isinstance(e.expr.node, MypyFile):
+                # Special case: reference to a missing module attribute.
+                self.add_dependency(make_trigger(e.expr.node.fullname() + '.' + e.name))
+                return
             typ = self.type_map[e.expr]
             self.add_attribute_dependency(typ, e.name)
             if self.use_logical_deps() and isinstance(typ, AnyType):
--- a/mypy/server/deps.py
+++ b/mypy/server/deps.py
@@ -599,7 +599,7 @@ class DependencyVisitor(TraverserVisitor):
             # Reference to a module attribute
             self.process_global_ref_expr(e)
         else:
-            # Reference to a non-module attribute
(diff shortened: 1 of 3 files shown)


test-data/unit/deps.test
```diff
@@ -827,6 +827,47 @@ def f(x: A) -> None:
...
+
+[case testIgnoredMissingClassAttribute]
+from a import C
+C.x  # type: ignore
+[file a.py]
+class C: pass
+[out]
+<a.C.x> -> m
+<a.C> -> m
+
```
ilevkivskyi: I would add a fine grained test too, so that it will not regress for other (non dependency caused) reasons.

mypy/server/deps.py
```diff
@@ -604,6 +604,10 @@ def visit_member_expr(self, e: MemberExpr) -> None:
                 # No type available -- this happens for unreachable code. Since it's unreachable,
                 # it wasn't type checked and we don't need to generate dependencies.
                 return
+            if isinstance(e.expr, RefExpr) and isinstance(e.expr.node, MypyFile):
+                # Special case: reference to a missing module attribute.
```
ilevkivskyi: I would also update the comment few lines above (can't comment there):
```
# Reference to a non-module attribute -> # Reference to a non-module (or missing) attribute
```

PR #5862, opened 2018-10-31
Generate fewer redundant fine-grained dependencies for "isinstance"

JukkaL:
This is mostly a problem with logical dependencies, but fewer redundant
dependencies is a small win in general.

Only common cases are handled. Things like tuple arguments still
generate redundant deps, but these are pretty rare.

### DIFF
--- a/mypy/server/deps.py
+++ b/mypy/server/deps.py
@@ -570,8 +570,8 @@ class DependencyVisitor(TraverserVisitor):
 
         # If this is a reference to a type, generate a dependency to its
         # constructor.
-        # TODO: avoid generating spurious dependencies for isinstance checks,
-        # except statements, class attribute reference, etc, if perf problem.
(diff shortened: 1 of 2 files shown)


mypy/server/deps.py
```diff
@@ -659,6 +659,24 @@ def visit_super_expr(self, e: SuperExpr) -> None:
                     break
 
     def visit_call_expr(self, e: CallExpr) -> None:
+        if isinstance(e.callee, RefExpr) and e.callee.fullname == 'builtins.isinstance':
```
ilevkivskyi: Is there a similar situation with `issubclass()`? It is more rare, but it looks like it can be easily added here.

JukkaL: Yeah, we could support it as well, but I don't think that it's worth even the minimal effort. I looked at an internal Dropbox codebase, and `issubclass()` was only used 1/50 as frequently as `isinstance()`.

Issue #5508, opened 2018-08-19
Add a plugin hook to define dynamic base classes


ilevkivskyi:
Some frameworks allow declaring special classes using the following syntax:
```python
Base = some_function(...)
```
It would be convenient to have a plugin hook that would allow analysing such definitions and store them as classes in module symbol table.

In addition, we could later refactor semantic analysis of `namedtuple`, `NewType`, `Enum` etc. to use this unified plugin API.

PR #5875, opened 2018-11-08
Add plugin hook for dynamic class definition

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5508

The new hook allows this:
```python
from some_lib import dynamic_base

Base = dynamic_base()

class C(Base):  # No error, the plugin acts and replaces 'Base' with a TypeInfo
    ...
```
This plugin hook is useful for SQLAlchemy ORM, for user-defined constructs similar to `namedtuple()`, and maybe we can even re-implement `namedtuple()` as a plugin at some point.

Here are some comments:
* I don't add `basic_new_typeinfo()` to semantic analyzer API, because it is hard to find a signature that would satisfy all possible use cases. Instead I add just `add_symbol_table_node()` and `qualified_name()`, so that a user can construct `TypeInfo` manually and add it to the symbol table. We can reconsider this when we will have more experience with this hook, and add a dedicated utility method.
* Semantic analyzer API in `semanal_shared.py` has `add_symbol_table_node()`, but I think it is outdated, and should be replaced with `add_symbol()` which is IMO better (this is unrelated to this PR, but if you agree I will open an issue).
* If you think the added test case is overcomplicated, this is on purpose. The added test case is a PoC implementation of how this would work in SQLAlchemy, I wanted to be sure that new hook will work well also in combination with the other base class related hook (and I would say this may be a typical combination).

### DIFF
--- a/test-data/unit/check-custom-plugin.test
+++ b/test-data/unit/check-custom-plugin.test
@@ -306,4 +306,28 @@ class Instr(Generic[T]): ...
 [file mypy.ini]
 [[mypy]
 plugins=<ROOT>/test-data/unit/plugins/dyn_class.py
-[builtins fixtures/classmethod.pyi]
+
+[case testDynamicClassPluginNegatives]
+# flags: --config-file tmp/mypy.ini
+from mod import declarative_base, Column, Instr, non_declarative_base
+
+Bad1 = non_declarative_base()
+Bad2 = Bad3 = declarative_base()
+
+class C1(Bad1): ...  # E: Invalid base class
+class C2(Bad2): ...  # E: Invalid base class
+class C3(Bad3): ...  # E: Invalid base class
+
+[file mod.py]
+from typing import Generic, TypeVar
+def declarative_base(): ...
+def non_declarative_base(): ...
+
+T = TypeVar('T')
+
+class Column(Generic[T]): ...
+class Instr(Generic[T]): ...
+
+[file mypy.ini]
+[[mypy]
+plugins=<ROOT>/test-data/unit/plugins/dyn_class.py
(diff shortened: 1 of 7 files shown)


test-data/unit/check-custom-plugin.test
```diff
@@ -280,3 +280,30 @@ reveal_type(FullyQualifiedTestNamedTuple('')._asdict()) # E: Revealed type is 'b
 [[mypy]
 plugins=<ROOT>/test-data/unit/plugins/fully_qualified_test_hook.py
 [builtins fixtures/classmethod.pyi]
+
+[case testDynamicClassPlugin]
+# flags: --config-file tmp/mypy.ini
+from mod import declarative_base, Column, Instr
+
+Base = declarative_base()
```
JukkaL: Also test a few different assignment statements and ensure that the plugin doesn't get run them. Maybe things like `A = B = ...` and `C = <other_function>()`. If the plugin doesn't get run, the lvalue shouldn't be a valid base class.

test-data/unit/check-custom-plugin.test
```diff
@@ -280,3 +280,30 @@ reveal_type(FullyQualifiedTestNamedTuple('')._asdict()) # E: Revealed type is 'b
...
+
+T = TypeVar('T')
+
+class Column(Generic[T]): ...
+class Instr(Generic[T]): ...
+
+[file mypy.ini]
+[[mypy]
+plugins=<ROOT>/test-data/unit/plugins/dyn_class.py
+[builtins fixtures/classmethod.pyi]
```
JukkaL: Would it make sense to add an AST diff test case for the base class? E.g. if there is no change, the base class shouldn't be triggered.

ilevkivskyi:
@JukkaL Thanks! I added the tests you suggested.

Issue #3592, opened 2017-06-22
Invalidate cache if plugins change


JukkaL:
If configured plugins change, type checking results may change and the mypy cache should become stale. Currently we don't track this and users may have to delete their cache manually when plugins change, which is a bad user experience.

It's unclear what's the best way to detect that plugins have changed. Here is a simple idea:
* Add a version string/tuple to all plugin objects. Assume that plugin authors update the version each time there is a plugin behavior change.
* Store information about the versions of all active plugins in cache files (except for the default plugin, which is tied to the mypy version).
* If the set of plugins or a version of any plugin changes, a cache file becomes stale.

gvanrossum:
Maybe we should just hash all the plugin source files together and use that as a guard? It's less precise but safer than relying on developers manually updating version strings. We all start with the best of intentions, but it's easy to forget (unless there's a forcing function like PyPI's refusal to accept a new upload with an old version string).

ilevkivskyi:
I think we should proceed with the hashing plugin code idea. I am going to work in this in next few days.

PR #5878, opened 2018-11-08
Invalidate all cache files if plugins change

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/3592

This invalidates _all_ cache files if at least one plugin changes its `__version__` or hash of its content (this will not catch situations when there are changes in modules imported by a plugin, _and_ version is also imported, but such situations are probably very rare).

I was not sure where to place the check to invalidate all caches at the same time, so I invalidate them one by one.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -21,6 +21,7 @@ import stat
 import sys
 import time
 import errno
+import types
 
 from typing import (AbstractSet, Any, Dict, Iterable, Iterator, List,
                     Mapping, NamedTuple, Optional, Set, Tuple, Union, Callable)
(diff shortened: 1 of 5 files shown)


mypy/build.py
```diff
@@ -375,11 +379,21 @@ def plugin_error(message: str) -> None:
                 '(in {})'.format(plugin_path))
         try:
             custom_plugins.append(plugin_type(options))
+            # We record _both_ hash and the version to detect more
```
JukkaL: This function is quite long. It would be better if this would be moved to a helper method to avoid making it even longer.

test-data/unit/check-incremental.test
```diff
@@ -5389,3 +5393,58 @@ E = Enum('E', f())  # type: ignore
 [builtins fixtures/list.pyi]
 [out]
 [out2]
+
+[case testChangedPluginsInvalidateCache]
```
JukkaL: Maybe add another test where only the version changes (e.g. import version from another module so that the text of the plugin doesn't change).

mypy/build.py
```diff
@@ -487,6 +501,9 @@ def __init__(self, data_dir: str,
                                in self.options.shadow_file}
         # a mapping from each file being typechecked to its possible shadow file
         self.shadow_equivalence_map = {}  # type: Dict[str, Optional[str]]
+        self.plugin = plugin
+        self.plugins_snapshot = plugins_snapshot
```
JukkaL: Document the new attributes in the class docstring.

test-data/unit/check-incremental.test
```diff
@@ -5389,3 +5393,58 @@ E = Enum('E', f())  # type: ignore
...
+# flags: --config-file tmp/mypy.ini
+import a
```
JukkaL: Would it make sense to add a fine-grained incremental test case where a plugin changes?

ilevkivskyi: As we discussed, this will have no effect.

JukkaL:
There is a test failure.

ilevkivskyi:
Hm, this is weird, all tests pass locally. Also the error appears on the _first_ run, which is even more weird.

ilevkivskyi:
OK, I also added restart for daemon if plugins change (in addition to cache invalidation) and tested manually that it works. @msullivan does this make sense to you?

ilevkivskyi:
@msullivan @JukkaL I assume this looks good now. Are there any other comments?

mypy/build.py
```diff
@@ -848,6 +901,11 @@ def find_cache_meta(id: str, path: str, manager: BuildManager) -> Optional[Cache
                     manager.trace('    {}: {} != {}'
                                   .format(key, cached_options.get(key), current_options.get(key)))
         return None
+    if manager.old_plugins_snapshot and manager.plugins_snapshot:
```
JukkaL: What about adding a test for this?

ilevkivskyi: We can't yet add tests for daemon. I opened https://github.com/python/mypy/issues/5891 for this.

PR #5892, opened 2018-11-14
Invalidate cache when a plugin is added/removed

ilevkivskyi:
https://github.com/python/mypy/pull/5878 added cache invalidation for situations where a plugin version/hash changes but there is even simpler scenario -- if `plugins` option itself change, the cache should be invalidated as well.


ilevkivskyi:
It is indeed long, but it is better to be on the safe side.

PR #5894, opened 2018-11-14
Skip rest of file upon top-level always-false assert

### LINKED ISSUE #5308: Idea: allow assert with platform or version check at top of file
Some files only work for certain platforms, but sometimes it's inconvenient to move the file to a location where it'll only be picked up for that platform. And yet the file doesn't type check cleanly on another platform (or ditto for Python versions).

We could enclose the entire file in `if sys.platform == 'win32':` but that seems overkill (and adding the indentation makes for an awkward diff). I propose an alternative that would have the same effect: `assert sys.platform == 'win32'`.

gvanrossum:
- A top-level assert has no indentation.

- An always-false condition is a check for `sys.platform` or
  `sys.version_info` or a condition derived from `MYPY` or from a name
  passed to `--always-false`; `assert False` doesn't count (!).
  (This definition is the same used to exclude conditional imports.)

Fixes #5308.

### DIFF
--- a/test-data/unit/check-unreachable-code.test
+++ b/test-data/unit/check-unreachable-code.test
@@ -671,3 +671,11 @@ import sys
 assert sys.platform == 'lol'
 def bar() -> None: pass
 [builtins fixtures/ops.pyi]
+
+[case testUnreachableAfterToplevelAssertNotInsideIf]
+import sys
+if sys.version_info[0] >= 2:
+    assert sys.platform == 'lol'
+    reveal_type('')  # E: Revealed type is 'builtins.str'
+reveal_type('')  # E: Revealed type is 'builtins.str'
+[builtins fixtures/ops.pyi]


docs/source/common_issues.rst
```diff
@@ -382,6 +382,22 @@ More specifically, mypy will understand the use of ``sys.version_info`` and
    else:
        # Other systems
 
+As a special case, you can also use one of these checks in a top-level
+``assert``; this makes mypy skip the rest of the file.  Example:
```
JukkaL: Maybe explicitly mention that the assert must be unconditional and not nested within a statement.

test-data/unit/check-unreachable-code.test
```diff
@@ -623,3 +623,51 @@ class Child(Parent):
         reveal_type(self)       # E: Revealed type is '__main__.Child'
         return 3
 [builtins fixtures/isinstance.pyi]
+
+[case testUnreachableAfterToplevelAssert]
```
JukkaL: Maybe also test the case where the assert is nested within an if statement.

gvanrossum:
OK, updated doc and added test.

PR #5898, opened 2018-11-15
Add hooks for running tests on an installed version of mypy

msullivan:
The main motivation for all of this is to be able to validate
mypy_mypyc wheels.

Because the tests are an integrated subpackage of mypy, I couldn't
figure out a way to run the tests from the source directory against an
installed mypy, and I was reluctant to do a major overhaul of the
source organization, in part because some other projects take
advantage of some of the test infrastructure (mypyc, sqlalchemy-stubs).

So instead we run the tests from where they get installed with the
package. To support this, we allow overriding where the tests look for
test-data using a `MYPY_TEST_PREFIX` environment variable.

### DIFF
--- a/misc/test_installed_version.sh
+++ b/misc/test_installed_version.sh
@@ -1,0 +1,34 @@
+#!/bin/bash -ex
+
+# Usage: misc/test_installed_version.sh [wheel] [python command]
+# Installs a version of mypy into a virtualenv and tests it.
+
+# A bunch of stuff about mypy's code organization and test setup makes
+# it annoying to test an installed version of mypy. If somebody has a
+# better way please let me know.
+
+TO_INSTALL="${1-.}"
+PYTHON="${2-python3}"
+VENV="$(mktemp -d --tmpdir mypy-test-venv.XXXXXXXXXX)"
+trap "rm -rf '$VENV'" EXIT
+
+"$PYTHON" -m virtualenv "$VENV"
+source "$VENV/bin/activate"
+
+pip install -r test-requirements.txt
+pip install $TO_INSTALL
+
+# pytest looks for configuration files in the parent directories of
+# where the tests live. Since we are trying to run the tests from
+# their installed location, we copy those into the venv. Ew ew ew.
+cp pytest.ini conftest.py "$VENV/"
+
+ROOT="$PWD"
+
+# Change directory so we can't pick up any of the stuff in the root
+cd /
+
+# Find the directory that mypy tests were installed into
+MYPY_TEST_DIR="$(python3 -c 'import mypy.test; print(mypy.test.__path__[0])')"
+# Run the mypy tests
+MYPY_TEST_PREFIX="$ROOT" python3 -m pytest "$MYPY_TEST_DIR"/test*.py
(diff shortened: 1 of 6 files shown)


misc/test_installed_version.sh
```diff
@@ -0,0 +1,34 @@
...
+# A bunch of stuff about mypy's code organization and test setup makes
+# it annoying to test an installed version of mypy. If somebody has a
+# better way please let me know.
+
+TO_INSTALL="${1-.}"
+PYTHON="${2-python3}"
+VENV="$(mktemp -d --tmpdir mypy-test-venv.XXXXXXXXXX)"
+trap "rm -rf '$VENV'" EXIT
+
+"$PYTHON" -m virtualenv "$VENV"
```
gvanrossum: Why not use `venv` instead of `virtualenv`? `venv` is in the stdlib, `virtualenv` has to be installed from  PyPI.

misc/test_installed_version.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# pytest looks for configuration files in the parent directories of
+# where the tests live. Since we are trying to run the tests from
+# their installed location, we copy those into the venv. Ew ew ew.
+cp pytest.ini conftest.py "$VENV/"
+
+ROOT="$PWD"
+
+# Change directory so we can't pick up any of the stuff in the root
+cd /
```
gvanrossum: Maybe `$VENV` instead? There's less distraction there...

mypy/stubgen.py
```diff
@@ -884,7 +884,8 @@ def walk_packages(packages: List[str]) -> Iterator[str]:
                 # using the inspect module
                 subpackages = [package.__name__ + "." + name
                                for name, val in inspect.getmembers(package)
-                               if inspect.ismodule(val)]
+                               if inspect.ismodule(val)
+                               if val.__name__ == package.__name__ + "." + name]
```
gvanrossum: I didn't even know you could do this! I'd use `and` to combine the conditions.

mypy/test/testcmdline.py
```diff
@@ -115,7 +115,9 @@ def normalize_file_output(content: List[str], current_abs_path: str) -> List[str
     """Normalize file output for comparison."""
     timestamp_regex = re.compile(r'\d{10}')
     result = [x.replace(current_abs_path, '$PWD') for x in content]
-    result = [re.sub(r'\b' + re.escape(__version__) + r'\b', '$VERSION', x) for x in result]
+    version = mypy.version.__version__
+    result = [re.sub(r'\b' + re.escape(version) + r'\b', '$VERSION', x) for x in result]
+    base_version = getattr(mypy.version, 'base_version', version)
```
gvanrossum: When would `base_version` ever be missing? Or is this another hack to fool mypyc?

msullivan: `base_version` is missing in all installed versions of mypy, due to https://github.com/python/mypy/blob/master/setup.py#L56-L61

gvanrossum: Ah... That's worth a comment here.

PR #5899, opened 2018-11-16
Remove an `if not MYPY` from types.py

msullivan:
`if not MYPY` doesn't /really/ work in mypyc (it works for imports,
but not for anything that uses a variable!), and I intend to change it
to work even less.

The hack here doesn't seem to be necessary, so there is no trouble
removing it.

This is, in a funny way, a prereq for #5859, since it is necessary to make mypy work under a not-yet-posted mypyc PR I have written to make platform checks work under mypyc.

I have an alternate patch I could post that instead merges `type_visitor.py` back into `types.py`, but maybe it is useful to keep them separate?

### DIFF
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -27,9 +27,10 @@ T = TypeVar('T')
 
 JsonDict = Dict[str, Any]
 
-# If we import type_visitor in the middle of the file, mypy breaks, and if we do it
-# at the top, it breaks at runtime because of import cycle issues, so we do it at different
-# times in different places.
+# If we only import type_visitor in the middle of the file, mypy
+# breaks, and if we do it at the top, it breaks at runtime because of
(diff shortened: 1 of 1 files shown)


msullivan:
It isn't that mypyc has a difference in treatment between them, it is that *mypy* believes that `not MYPY` blocks are unreachable and does not semantically analyze them. It turns out that mypyc can compile an import that wasn't semantically analyzed, but that is close to all it can do (most other things it will crash on). I'm going to change mypyc to instead replace the body of unreachable blocks with a `raise RuntimeError`, so that we can compile the conditional code in #5859. But that breaks the `if not MYPY` here, which currently works but mostly by accident.

gvanrossum:
Got it -- it makes sense that mypyc doesn't "see" blocks that are
considered unreachable during type checking.

PR #5900, opened 2018-11-16
Make stubgen less aggressive about walking C extension modules

msullivan:
PR #5814 made stubgen try to walk as a module any attribute of a C
of a C extension module of module type. This is too aggressive: mypyc
generated extension modules will often have attributes of module type
from imports that do not correspond to actual modules.

This causes a stubgen test to fail under mypy_mypyc while trying to
import `mypy.errors.os`, since `mypy.errors` imports `os`.

Only walk such a submodule when its `__name__` matches it being a
submodule.

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -885,7 +885,7 @@ def walk_packages(packages: List[str]) -> Iterator[str]:
                 subpackages = [package.__name__ + "." + name
                                for name, val in inspect.getmembers(package)
                                if inspect.ismodule(val)
-                               if val.__name__ == package.__name__ + "." + name]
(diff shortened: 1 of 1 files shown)


gvanrossum:
Before I approve this, a question for @pkerichang: Can you test this with your pybind11 setup? (This could go awry if pybind11 doesn't set `__name__` to the fully qualified module name.)

msullivan:
I'm going to go ahead and merge this now, so that mypyc tests get fixed. @pkerichang, if this breaks what you were trying to do in #5814, follow up with us

pkerichang:
Sorry for the late response; this works fine on my end.  Thanks for letting me know.

PR #5906, opened 2018-11-16
End to end tests for daemon (dmypy)

### LINKED ISSUE #5891: Add integration tests for daemon
Currently we only have unit tests for fine grained mode, but no integration tests for the daemon itself. This is in particular important for `dmypy run` that has some non-trivial restart logic.

Related https://github.com/python/mypy/issues/5002

gvanrossum:
Fixes #5891

### DIFF
--- a/mypy/test/testdaemon.py
+++ b/mypy/test/testdaemon.py
@@ -30,38 +30,40 @@ def test_daemon(testcase: DataDrivenTestCase) -> None:
     if sys.platform == 'win32':
         return  # These tests don't run on Windows yet.
     assert testcase.old_cwd is not None, "test was not properly set up"
-    for i, cmd in enumerate(parse_script(testcase.input)):
-        input = cmd[0]
-        expected_lines = cmd[1:]
-        assert input.startswith('$')
-        input = input[1:].strip()
-        sts, output = run_cmd(input)
+    for i, step in enumerate(parse_script(testcase.input)):
+        cmd = step[0]
+        expected_lines = step[1:]
+        assert cmd.startswith('$')
+        cmd = cmd[1:].strip()
+        sts, output = run_cmd(cmd)
         output_lines = output.splitlines()
         if sts:
             output_lines.append('== Return code: %d' % sts)
         assert_string_arrays_equal(expected_lines,
                                    output_lines,
                                    "Command %d (%s) did not give expected output" %
-                                   (i + 1, input))
+                                   (i + 1, cmd))
 
 
 def parse_script(input: List[str]) -> List[List[str]]:
-    # Parse testcase.input into commands.
-    # Each command starts with a line starting with '$'.
-    # The first line (less '$') is sent to the shell.
-    # The remaining lines are expected output.
-    commands = []
-    cmd = []  # type: List[str]
+    """Parse testcase.input into steps.
+
+    Each command starts with a line starting with '$'.
+    The first line (less '$') is sent to the shell.
+    The remaining lines are expected output.
+    """
+    steps = []
+    step = []  # type: List[str]
     for line in input:
         if line.startswith('$'):
-            if cmd:
-                assert cmd[0].startswith('$')
-                commands.append(cmd)
-                cmd = []
-        cmd.append(line)
-    if cmd:
-        commands.append(cmd)
-    return commands
+            if step:
+                assert step[0].startswith('$')
+                steps.append(step)
+                step = []
+        step.append(line)
+    if step:
+        steps.append(step)
+    return steps
 
 
 def run_cmd(input: str) -> Tuple[int, str]:
--- a/mypy/test/testdaemon.py
+++ b/mypy/test/testdaemon.py
@@ -1,0 +1,81 @@
+"""End-to-end test cases for the daemon (dmypy).
+
+These are special because they run multiple shell commands.
+"""
+
+import os
+import subprocess
+
+from typing import List, Tuple
+
+from mypy.test.config import test_temp_dir, PREFIX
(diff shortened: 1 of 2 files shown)


mypy/test/testdaemon.py
```diff
@@ -0,0 +1,83 @@
...
+        output = subprocess.check_output(input,
+                                         shell=True,
+                                         stderr=subprocess.STDOUT,
+                                         universal_newlines=True,
+                                         cwd=test_temp_dir,
+                                         env=env)
+        return 0, output
+    except subprocess.CalledProcessError as err:
+        output = err.output
+        if err.stderr:
```
ethanhs: I believe this will always be `None` since you are not using `subprocess.run`. It also doesn't exist until 3.5 anyway.

mypy/test/testdaemon.py
```diff
@@ -0,0 +1,83 @@
...
+    if sys.platform == 'win32':
+        return  # These tests don't run on Windows yet.
+    assert testcase.old_cwd is not None, "test was not properly set up"
+    for i, cmd in enumerate(parse_script(testcase.input)):
+        input = cmd[0]
+        expected_lines = cmd[1:]
+        assert input.startswith('$')
+        input = input[1:].strip()
+        if input.startswith('dmypy '):
+            input = 'python3 -m mypy.' + input
```
ethanhs: `python3` -> `sys.executable` for Windows.

mypy/test/testdaemon.py
gvanrossum: Thanks, I'll get rid of this.

mypy/test/testdaemon.py
gvanrossum: OK. (Though I'd be surprised if that's all that's needed to make this test work on Windows.)

ethanhs: Actually, it is! Though it seems that the last test is failing, but I think that might be an issue on my end, so I will experiment a bit more.

ethanhs: No, I don't think its that, echo exists on Windows and works very similarly to the echo on Linux. Furthermore, the output of the failing test is empty, which is very wrong. Something else is going on, I'll experiment with it on Monday when I have time to get back to things.

test-data/unit/daemon.test
```diff
@@ -0,0 +1,84 @@
...
+$ dmypy restart --  --follow-imports=error
+Daemon started
+$ dmypy recheck
+Command 'recheck' is only valid after a 'check' command
+== Return code: 1
+$ dmypy kill
+Daemon killed
+$ dmypy status
+Daemon has died
+== Return code: 1
```
ilevkivskyi: I am of course biased but I would add two tests:
* Check that `dmypy run` restarts when a plugin version is changed
* Check that `--update`/`--remove` work as expected (including situations like removing non-existent file from the build)

mypy/test/testdaemon.py
```diff
@@ -0,0 +1,81 @@
...
+class DaemonSuite(DataSuite):
+    files = daemon_files
+
+    def run_case(self, testcase: DataDrivenTestCase) -> None:
+        test_daemon(testcase)
+
+
+def test_daemon(testcase: DataDrivenTestCase) -> None:
+    if sys.platform == 'win32':
+        return  # These tests don't run on Windows yet.
```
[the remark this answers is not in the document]
ethanhs: Probably better to land this and keep me honest (also so I can investigate/fix issues if any on my end). I already have a branch with this on top of my branch.

mypy/test/testdaemon.py
```diff
@@ -0,0 +1,81 @@
...
+                                   "Command %d (%s) did not give expected output" %
+                                   (i + 1, input))
+
+
+def parse_script(input: List[str]) -> List[List[str]]:
+    # Parse testcase.input into commands.
+    # Each command starts with a line starting with '$'.
+    # The first line (less '$') is sent to the shell.
+    # The remaining lines are expected output.
+    commands = []
```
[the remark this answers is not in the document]
gvanrossum: Done

mypy/test/testdaemon.py
```diff
@@ -0,0 +1,81 @@
...
+                                   output_lines,
+                                   "Command %d (%s) did not give expected output" %
+                                   (i + 1, input))
+
+
+def parse_script(input: List[str]) -> List[List[str]]:
+    # Parse testcase.input into commands.
+    # Each command starts with a line starting with '$'.
+    # The first line (less '$') is sent to the shell.
+    # The remaining lines are expected output.
```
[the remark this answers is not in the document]
gvanrossum: Done.

test-data/unit/daemon.test
gvanrossum: I've added the former. For the latter I need a little more time to think through all the cases.

gvanrossum: OK, I added a simple version of the latter. Let me know if there's another edge case you'd like me to test.

mypy/test/testdaemon.py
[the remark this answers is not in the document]
ilevkivskyi: OK, makes sense.

gvanrossum:
OK, merged. @ethanhs You can merge this into your branch and delete the win32 test. Hopefully it really does work for you and the test failure was something on your end. (If not we can roll it back or fix it forward.)

gvanrossum:
There's one problem with this. When a test fails while the daemon is running, the daemon keeps running. (A `ps` I ran on my box just revealed this.) It should be simple enough for the test framework to add a `dmypy kill` command, and I'll add that now.

gvanrossum:
See #5913 for the fix.

Issue #5409, opened 2018-07-31
Plugin API feature requests


gvanrossum:
I recently wrote a plugin that used `get_function_hook()` to intercept instantiations of a certain set of classes (recognized by name), and found a bunch of functionality missing. Here's a list.

- I needed the names of the arguments but they were not directly available. I ended up using `rtype.type.names['__init__'].node.arg_names` (with various type and `None` guards) and subtracting 1 to compensate for `self`, which looked very hacky. It would be nice if `FunctionContext` also has an `arg_names` argument.

- `CheckerPluginInterface` has no `fail()` method. It exists on the implementation, and the similar `TypeAnalyzerPluginInterface` does define it.

- I needed to look up a name in the module's global namespace; I ended up calling `ctx.api.lookup_qualified()` but that's not in the interface, and perhaps it's not the thing to use.

- I needed the functionality of `parse_bool()` but it's not there (it only exists on `SemanticAnalyzerPluginInterface`).

- I needed something similar to what's in mypy/plugins/common.py as `_get_bool_argument()` `_get_argument()` but had to write my own because the starting point was a `FunctionContext` instead of a `ClassDefContext`.

- (Not directly a plugin issue, but this happened.) I was also bitten by the error message `The type alias{} is invalid in runtime context` -- this seems a bit inconsistent because expanding the `Union` that the alias references makes it go away. I.e. if `A = Optional[int]`, `foo(A)` gives that error, but `foo(Optional[int])` doesn't.

ilevkivskyi:
Here are two thoughts specifically about `FunctionContext` problems:
* Actual argument kinds are missing, and this causes unsolvable problems (not just inconveniences). For example, if the function is defined as `def func(*x: int) -> Any: ...`, then for a function call `x: List[int]; func(1, 2, *x)` the formal types in the context are simply `[[int, int, List[int]]]`, i.e. it is indistinguishable from `func(1, 2, [1, 2])` (this occurred with `ctypes` plugin). Interestingly, `arg_kinds` are passed to `apply_function_plugin`, but are never used in that function. We should definitely add `arg_kinds: List[List[int]]` (e.g. as `[[ARG_POS, ARG_POS, ARG_STAR]]` in the example above) to the `FunctionContext`.
[...]

PR #5918, opened 2018-11-19
Add arg_names attribute to FunctionContext and MethodContext

mkurnikov:
References https://github.com/python/mypy/issues/5409.

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -43,6 +43,7 @@ from abc import abstractmethod
 from typing import Callable, List, Tuple, Optional, NamedTuple, TypeVar, Dict
 from mypy_extensions import trait
 
+from mypy.lookup import lookup_fully_qualified
 from mypy.nodes import (
     Expression, Context, ClassDef, SymbolTableNode, MypyFile, CallExpr
 )
@@ -50,7 +51,6 @@ from mypy.tvar_scope import TypeVarScope
 from mypy.types import Type, Instance, CallableType, TypeList, UnboundType
 from mypy.messages import MessageBuilder
 from mypy.options import Options
-from mypy.lookup import lookup_fully_qualified
 import mypy.interpreted_plugin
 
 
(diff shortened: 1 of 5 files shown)


mkurnikov:
I'll finish it today or on the weekend. With the recent refactoring, it will be even easier.

mkurnikov:
WIP, missing tests for `arg_kinds`, existing tests should be rewritten and docs for new context attributes. Will try to finish on this weekend.

Explanation of `callee_arg_names`: 
Guido mentioned in the 
https://github.com/python/mypy/issues/5409#issue-346398965
> I needed the names of the arguments but they were not directly available. I ended up using rtype.type.names['\_\_init__'].node.arg_names (with various type and None guards) and subtracting 1 to compensate for self, which looked very hacky. It would be nice if FunctionContext also has an arg_names argument.

I've encountered the same thing in `django-stubs`. 

Five arrays of the same length don't look great, but I couldn't find any other way.

mypy/plugin.py
```diff
@@ -159,7 +162,10 @@ def qualified_name(self, n: str) -> str:
...
     'MethodContext', [
         ('type', Type),                    # Base object type for method call
-        ('arg_types', List[List[Type]]),
+        ('callee_arg_names', List[Optional[str]]),
+        ('arg_names', List[List[Optional[str]]]),
+        ('arg_kinds', List[List[int]]),
+        ('arg_types', List[List[Type]]),  # List of actual caller types for each formal argument
         ('default_return_type', Type),
         ('args', List[List[Expression]]),
         ('context', Context),
```
ilevkivskyi: Could you please also add a comment for every existing attribute here and above?

mkurnikov:
I'll finish it tomorrow.

ilevkivskyi:
Great, thanks!

ilevkivskyi:
Also please never rebase, it is much harder to review, and GitHub only sends a notification when you push a new commit.

ilevkivskyi:
Thanks for the PR!

mkurnikov:
Got it.

Issue #1969, opened 2016-07-31
**kwargs checked against wrong type with defaults or *args


kcr:
```
def a(*args: str, **kw: int) -> None:
    b(**kw)
    b('', **kw)
    c(**kw)
    c('', **kw)

def b(args: str='', **kw: int) -> None:
    pass

def c(*args: str, **kw: int) -> None:
    pass
```
results, with 0.4.3:
```
/tmp/minimize.py: note: In function "a":
/tmp/minimize.py:2: error: Argument 1 to "b" has incompatible type **Dict[str, int]; expected "str"
/tmp/minimize.py:4: error: Argument 1 to "c" has incompatible type **Dict[str, int]; expected "str"
```

gvanrossum:
Can you try again with the latest revision from the repo? I fixed some
things in this area recently.

--Guido (mobile)

gvanrossum:
Never mind. I can still repro this.

gnprice:
The logic the code might be implementing here, at least in the call to `b`, is that it's possible 'args' is a key in `kw`, in which case we'd be passing an `int` to a parameter of type `str`.

That logic doesn't seem to apply in the second case, though, which would make that a bug.

In the `b` case it's possible this is the kind of unsoundness we'd rather allow, because it's probably cumbersome to explicitly prevent it in this code. Even if we do give an error for it, this error message isn't really on point about what the issue is, and if nothing else we should make it clearer.

gnprice:
And thanks, Karl, for the report!

gvanrossum:
In particular, calling `a(args=0)` would end up calling `b(args=0)` which is a bug in your code. But the same reasoning doesn't apply to `c(**kw)`, so that's a bug in mypy.

elazarg:
This can be fixed by adding at [checkexpr.py#L1980](https://github.com/python/mypy/blob/master/mypy/checkexpr.py#L1890) (right below the comment referring to that situation)

    -    if ((callee_names[j] and no_certain_match)
                        or callee_kinds[j] == nodes.ARG_STAR2):
    +    if ((callee_names[j] and no_certain_match and callee_kinds[j] != nodes.ARG_STAR)
                        or callee_kinds[j] == nodes.ARG_STAR2):

However, it raises another error for the following file. I wonder whether this is the expected behavior:
```
def a(**kw: int) -> None:
    a(**kw)
    c(**kw)
    d(**kw)

def b(a: int) -> None: pass
def c(*args: str) -> None: pass
[...]

PR #5925, opened 2018-11-20
Type check typed dict as caller **kwargs

### LINKED ISSUE #5198: TypedDict unpacking should be checked more strictly
I'm not sure if this should be classified as a missing feature or a bug, but the behavior of the following program is surprising to me:

    def foo(x: int, y: int) -> None: pass

    OkDict = TypedDict('OkDict', {'x': int, 'y': int})
    BadDict1 = TypedDict('BadDict1', {'x': int})
    BadDict2 = TypedDict('BadDict2', {'x': int, 'y': str})

    a: OkDict
    b: BadDict1
    c: BadDict2

    foo(**a)  # Type checks, as expected
    foo(**b)  # Type checks; should be an error
    foo(**c)  # Type checks; should be an error

Mypy currently reports no errors with this program; I think we should report an error with the last two calls.

(I'm assuming we're currently ignoring the 'shape' of the TypedDicts and are handling them in the same way we handle Dict?)

JukkaL:
Expand and type check TypedDict types when used as `**kwargs` in calls.

Also refactored the implementation of checking function arguments and
removed some apparently useless code.

Fixes #5198 and another related issue: type checking calls with multiple
`*args` arguments.

### DIFF
--- a/mypy/argmap.py
+++ b/mypy/argmap.py
@@ -76,7 +76,8 @@ def map_actuals_to_formals(caller_kinds: List[int],
                 # We don't exactly know which **kwargs are provided by the
                 # caller. Assume that they will fill the remaining arguments.
                 for j in range(ncallee):
-                    # TODO tuple varargs complicate this
+                    # TODO: If there are also tuple varargs, we might be missing some potential
+                    #       matches if the tuple was short enough to not match everything.
                     no_certain_match = (
                         not map[j] or caller_kinds[map[j][0]] == nodes.ARG_STAR)
                     if ((callee_names[j] and no_certain_match)
--- a/mypy/argmap.py
+++ b/mypy/argmap.py
@@ -105,11 +105,25 @@ def map_formals_to_actuals(caller_kinds: List[int],
     return actual_to_formal
 
 
-class ArgTypeMapper:
-    """Utility class for mapping actual argument types to formal argument types.
+class ArgTypeExpander:
+    """Utility class for mapping actual argument types to formal arguments.
 
-    The main job is to expand tuple *args and typed dict **kwargs in caller, and to
-    keep track of which tuple/typed dict items have already been consumed.
+    One of the main responsibilities is to expand caller tuple *args and TypedDict
+    **kwargs, and to keep track of which tuple/TypedDict items have already been
+    consumed.
+
+    Example:
+
+       . def f(x: int, *args: str) -> None: ...
+       . f(*(1, 'x', 1.1)
+
+    We'd call expand_actual_type twice:
+
+      1. The first call would provide 'int' as the type of 'x'.
+      2. The second call would provide 'str' and 'float' as the types of '*args'.
+
+    Construct a separate instance for each call since instances have per-call
+    state.
     """
 
     def __init__(self) -> None:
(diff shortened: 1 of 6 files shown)


mypy/checkexpr.py
```diff
@@ -1152,47 +1178,28 @@ def check_argument_types(self, arg_types: List[Type], arg_kinds: List[int],
...
-                if (arg_kinds[actual] == nodes.ARG_STAR and
-                        not self.is_valid_var_arg(arg_type)):
-                    messages.invalid_var_arg(arg_type, context)
-                if (arg_kinds[actual] == nodes.ARG_STAR2 and
-                        not self.is_valid_keyword_var_arg(arg_type)):
-                    is_mapping = is_subtype(arg_type, self.chk.named_type('typing.Mapping'))
-                    messages.invalid_keyword_var_arg(arg_type, is_mapping, context)
-                # Get the type of an individual actual argument (for *args
-                # and **args this is the item type, not the collection type).
-                if (isinstance(arg_type, TupleType)
```
JukkaL: This seemed unnecessary so I removed it.

mypy/checkexpr.py
```diff
@@ -1152,47 +1178,28 @@ def check_argument_types(self, arg_types: List[Type], arg_kinds: List[int],
...
+                    actual_type, actual_kind,
+                    callee.arg_names[i], callee.arg_kinds[i])
+                check_arg(expanded_actual, actual_type, arg_kinds[actual],
                           callee.arg_types[i],
                           actual + 1, i + 1, callee, context, messages)
 
-                # There may be some remaining tuple varargs items that haven't
-                # been checked yet. Handle them.
-                tuplet = arg_types[actual]
-                if (callee.arg_kinds[i] == nodes.ARG_STAR and
```
JukkaL: This seemed unnecessary as well so I got rid of it.

test-data/unit/check-typeddict.test
```diff
@@ -1413,3 +1413,81 @@ from mypy_extensions import TypedDict
...
+
+a: A
+b: B
+f(**a) # E: Argument 1 to "f" has incompatible type "**A"; expected "str"
+f(**b)
+g(**a)
+g(**b) # E: Argument "x" to "g" has incompatible type "str"; expected "int"
+g(1, **a) # E: "g" gets multiple values for keyword argument "x"
+g(1, **b) # E: "g" gets multiple values for keyword argument "x" \
+          # E: Argument "x" to "g" has incompatible type "str"; expected "int"
```
ilevkivskyi: Does this PR also fix https://github.com/python/mypy/issues/1969? If yes, then I would update the description and added a test with optional argument.

mypy/checkexpr.py
```diff
@@ -1127,19 +1100,72 @@ def check_argument_count(self, callee: CallableType, actual_types: List[Type],
...
+                                         actual_types: List[Type],
+                                         actual_kinds: List[int],
+                                         actual_names: Optional[Sequence[Optional[str]]],
+                                         all_actuals: List[int],
+                                         context: Context,
+                                         messages: Optional[MessageBuilder]) -> Tuple[bool, bool]:
+        """Check for extra actual arguments.
+
+        Return tuple (was everything ok,
+                      was there an extra keyword argument error).
```
ilevkivskyi: Maybe mention that the latter is used to avoid duplicate errors?

mypy/argmap.py
```diff
@@ -95,35 +105,87 @@ def map_formals_to_actuals(caller_kinds: List[int],
...
+                else:
+                    self.tuple_index += 1
+                return actual_type.items[self.tuple_index - 1]
+            else:
+                return AnyType(TypeOfAny.from_error)
+        elif actual_kind == nodes.ARG_STAR2:
+            if isinstance(actual_type, TypedDictType):
+                if formal_kind != nodes.ARG_STAR2 and formal_name in actual_type.items:
+                    # Lookup type based on keyword argument name.
+                    assert formal_name is not None
```
ilevkivskyi: What if `formal_kind` is `ARG_STAR`? Or this can't happen for some reasons?

mypy/argmap.py
```diff
@@ -95,35 +105,87 @@ def map_formals_to_actuals(caller_kinds: List[int],
...
+                else:
+                    return AnyType(TypeOfAny.from_error)
+            elif isinstance(actual_type, TupleType):
+                # Get the next tuple item of a tuple *arg.
+                if self.tuple_index >= len(actual_type.items):
+                    # Exhausted a tuple -- continue to the next *args.
+                    self.tuple_index = 1
+                else:
+                    self.tuple_index += 1
+                return actual_type.items[self.tuple_index - 1]
```
ilevkivskyi: Maybe I am missing something, but it seems to me the logic is wrong here, if the tuple exhausted, you return the first element of the first tuple, not the second one. It is a bit hard to reason, could you please add a test for something like this:
```python
def f(a: T1, b: T2, c: T3, d: T4) -> Tuple[T1, T2, T3, T4]: ...
x: Tuple[A, B]
y: Tuple[C, D]
reveal_type(f(*x, *y))
```

mypy/argmap.py
```diff
@@ -95,35 +105,87 @@ def map_formals_to_actuals(caller_kinds: List[int],
...
+                           actual_type: Type,
+                           actual_kind: int,
+                           formal_name: Optional[str],
+                           formal_kind: int) -> Type:
+        """Return the actual (caller) type(s) of a formal argument with the given kinds.
+
+        If the actual argument is a tuple *args, return the individual tuple item(s) that
+        map(s) to the formal arg.
+
+        If the actual argument is a TypedDict **kwargs, return the matching typed dict dict
```
ilevkivskyi: Duplicate word "dict"?

mypy/argmap.py
```diff
@@ -65,13 +65,23 @@ def map_actuals_to_formals(caller_kinds: List[int],
...
+                for name, value in argt.items.items():
+                    if name in callee_names:
+                        map[callee_names.index(name)].append(i)
+                    elif nodes.ARG_STAR2 in callee_kinds:
+                        map[callee_kinds.index(nodes.ARG_STAR2)].append(i)
+            else:
+                # We don't exactly which **kwargs are provided by the
+                # caller. Assume that they will fill the remaining arguments.
+                for j in range(ncallee):
+                    # TODO tuple varargs complicate this
```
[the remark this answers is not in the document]
JukkaL: I think that it means that we don't which args a variable-length tuple *args will match, so we make assumptions. I will update the comment -- this is still an issue.

mypy/argmap.py
```diff
@@ -95,35 +105,87 @@ def map_formals_to_actuals(caller_kinds: List[int],
...
+    """Utility class for mapping actual argument types to formal arguments.
+
+    One of the main responsibilities is to expand caller tuple *args and TypedDict
+    **kwargs, and to keep track of which tuple/TypedDict items have already been
+    consumed.
+
+    Example:
+
+       . def f(x: int, *args: str) -> None: ...
+       . f(*(1, 'x', 1.1)
```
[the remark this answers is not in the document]
JukkaL: The dots were an attempt to avoid breaking auto indent in emacs, but apparently they are no longer required.

mypy/argmap.py
```diff
@@ -95,35 +105,87 @@ def map_formals_to_actuals(caller_kinds: List[int],
...
+
+        This is supposed to be called for each formal, in order. Call multiple times per
+        formal if multiple actuals map to a formal.
+        """
+        if actual_kind == nodes.ARG_STAR:
+            if isinstance(actual_type, Instance):
+                if actual_type.type.fullname() == 'builtins.list':
+                    # List *arg.
+                    return actual_type.args[0]
+                elif actual_type.args:
```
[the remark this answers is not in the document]
JukkaL: I'd rather keep it as it is to stay closer to "one issue per PR". This PR already does too much for a single PR, arguably. Piggy packing additional fixes to a PR is a bad habit I'm trying to move away from.

test-data/unit/check-typeddict.test
JukkaL: No

mypy/argmap.py
JukkaL: The `in` check above guarantees that this is not None.

mypy/argmap.py
JukkaL: The logic is correct, but added a test case. We do a -1 afterwards.

mypy/argmap.py
ilevkivskyi: Hm, I though that we narrow optional type to non-optional after an `in` check, if it doesn't work here (assert is still needed) it may be a bug.

Issue #5914, opened 2018-11-19
"Cannot determine type" on inner class inheritance


mthuurne:
I run mypy on the following code, with no arguments other than the file name:
```python
class InnerBase:
    pass

class OuterBase:
    class Inner(InnerBase):
        pass

class OuterMixin:
    class Inner(InnerBase):
        pass

class Outer(OuterBase, OuterMixin):
    pass
```
The output that I get is as follows:
```
inner.py:12: error: Cannot determine type of 'Inner' in base class 'OuterBase'
inner.py:12: error: Cannot determine type of 'Inner' in base class 'OuterMixin'
```
Line 12 is the line that starts with `class Outer`.

I don't understand why the types cannot be determined: `Type[OuterBase.Inner]` and `Type[OuterMixin.Inner]` respectively would be the types for the inner classes, I think.

Note that the errors go away when `Outer` inherits only from `OuterBase` or only from `OuterMixin`.

I'm using mypy 0.641 on Python 3.6.5.

mthuurne:
Sorry for opening a duplicate; neither the manual search I did nor the GitHub "related issues" beta function turned up the existing report.

This minimal test case leaves out some details that the production code uses. Even so, I think it's type-safe as long as the `Outer` class will only use `Inner` according to the `InnerBase` interface.

PR #5926, opened 2018-11-20
Make error messages from multiple inheritance compatibility check more accurate

mkurnikov:
Fixes https://github.com/python/mypy/issues/2871.

Initially, in discussion with @ilevkivskyi in Gitter, he suggested to just remove error, if there are two nested classes in a multiple inheritance with the same name
```
class Mixin1:
    class Meta:
        pass
class Mixin2:
    class Meta:
        pass
class A(Mixin1, Mixin2):
    pass
```
However, later we decided to make it safe and emit a better error message, including for cases with nested class and non-class for obvious cases. Note that for class objects we ignore the `__init__`  method signature of nested class and only check subclassing relationship between them.


mthuurne:
This indeed solves the issue I had in #5914, over 50 fewer false positives in the output now. Thanks!

mypy/checker.py
```diff
@@ -1618,12 +1628,12 @@ def check_compatibility(self, name: str, base1: TypeInfo,
...
         first = base1[name]
         second = base2[name]
-        first_type = first.type
-        if first_type is None and isinstance(first.node, FuncBase):
-            first_type = self.function_type(first.node)
-        second_type = second.type
-        if second_type is None and isinstance(second.node, FuncBase):
-            second_type = self.function_type(second.node)
+        if isinstance(first.node, TypeInfo) and isinstance(second.node, TypeInfo):
+            # allow nested classes with the same name
```
ilevkivskyi: In general I agree it is better to admit this little unsafety and fix a common pain point.

Please add a more detailed comment with motivation/trade-offs. Here are some points to mention:
* This is technically unsafe due to possible errors when accessing members of the nested classes.
* But this creates many false positives for Django.
* Also mypy anyway doesn't flag this for single inheritance.
* A potential solution may be to require at least structural compatibility (since requiring nominal one is impractical).

test-data/unit/check-multiple-inheritance.test
```diff
@@ -240,3 +240,37 @@ def dec(f: Callable[..., T]) -> Callable[..., T]:
...
+class Mixin1:
+    class Nested1:
+        pass
+class Mixin2:
+    def Nested1(self) -> str:
+        pass
+class A(Mixin1, Mixin2):
+    pass
+[out]
+main:7: error: Definition of "Nested1" in base class "Mixin1" is incompatible with definition in base class "Mixin2"
```
ilevkivskyi: Please add test cases where one of the classes defines a variable referring to another class, for example:
```python
class A:
    def __init__(self, arg: str) -> None: ...
class B(A): ...

class Base1:
    Nested = A
class Base2:
    Nested = B
class Combo(Base1, Base1): ...
```
I think this test case (and similar) should pass. Since this is logically (almost) equivalent to "inline" class definitions.

mkurnikov:
@ilevkivskyi I addressed your comments.

ethanhs:
It looks like the Windows test failure is a flake. I am looking into it.

mypy/checker.py
```diff
@@ -1618,19 +1628,23 @@ def check_compatibility(self, name: str, base1: TypeInfo,
...
         if (isinstance(first_type, FunctionLike) and
                 isinstance(second_type, FunctionLike)):
-            # Method override
-            first_sig = bind_self(first_type)
-            second_sig = bind_self(second_type)
-            ok = is_subtype(first_sig, second_sig, ignore_pos_arg_names=True)
+            if ((isinstance(first_type, CallableType)
+                 and first_type.fallback.type.fullname() == 'builtins.type')
+                    and (isinstance(second_type, CallableType)
+                         and second_type.fallback.type.fullname() == 'builtins.type')):
```
ilevkivskyi: This check is unreliable:
* What if the class constructor is overloaded?
* What if there is a custom metaclass?

Maybe you are trying to reinvent `FunctionLike.is_type_obj()`?

mypy/checker.py
```diff
@@ -1618,19 +1628,23 @@ def check_compatibility(self, name: str, base1: TypeInfo,
...
+                # Both members are classes (not necessary nested), check if compatible
+                ok = is_subtype(first_type.ret_type, second_type.ret_type)
+            else:
+                # Method override
```
ilevkivskyi: This comment adds nothing here, I would just remove it.

mypy/checker.py
```diff
@@ -1618,19 +1628,23 @@ def check_compatibility(self, name: str, base1: TypeInfo,
...
                 isinstance(second_type, FunctionLike)):
-            # Method override
-            first_sig = bind_self(first_type)
-            second_sig = bind_self(second_type)
-            ok = is_subtype(first_sig, second_sig, ignore_pos_arg_names=True)
+            if ((isinstance(first_type, CallableType)
+                 and first_type.fallback.type.fullname() == 'builtins.type')
+                    and (isinstance(second_type, CallableType)
+                         and second_type.fallback.type.fullname() == 'builtins.type')):
+                # Both members are classes (not necessary nested), check if compatible
```
ilevkivskyi: This comment is not clear. Probably you want to say something like: `# For class objects only check the subtype relationship of the classes, since we allow incompatible overrides of '__init__'/'__new__'`.

mypy/checker.py
```diff
@@ -1618,19 +1628,23 @@ def check_compatibility(self, name: str, base1: TypeInfo,
...
+                ok = is_subtype(first_type.ret_type, second_type.ret_type)
```
ilevkivskyi: This may cause surprises if the classes are generic, so the return types contain some (independent) type variables. I think you can just erase the type variables here (`erasetype.erase_type_vars` IIRC).

test-data/unit/check-multiple-inheritance.test
```diff
@@ -240,3 +240,127 @@ def dec(f: Callable[..., T]) -> Callable[..., T]:
...
+class Outer(Any):
+    pass
+class Mixin1:
+    Meta = Outer
+class Mixin2:
+    class Meta(Any):
+        pass
+class A(Mixin1, Mixin2):
+    pass
+[out]
```
ilevkivskyi: No newline at the end of file.

mkurnikov:
> are you going to add fallback_to_any by the plugin for all classes nested in models?

Yes, I tried it, it worked. 
https://github.com/mkurnikov/django-stubs/blob/master/mypy_django_plugin/plugins/meta_inner_class.py#L5

In the wild I've encountered only nested classes used to:
1. Somehow specify some properties of an outer class objects. 
2. Create class semantically bound to outer class, like `User.DoesNotExist` for exceptions in Django. 

Both of them are handled by the framework in very specific way, I wouldn't expect mypy to understand them without plugin, and errors are more or less OK there.

ilevkivskyi:
Also please don't force push, GitHub doesn't send e-mail notifications for this either. I will take a look at the updates tomorrow.

test-data/unit/check-multiple-inheritance.test
```diff
@@ -240,3 +240,241 @@ def dec(f: Callable[..., T]) -> Callable[..., T]:
...
+class Generic2(Generic[T]):
+    pass
+class Base1:
+    Nested = Generic1
+class Base2:
+    Nested = Generic2
+class A(Base1, Base2):
+    pass
+[out]
+main:11: error: Definition of "Nested" in base class "Base1" is incompatible with definition in base class "Base2"
```
ilevkivskyi: Please also add a test with a valid use of generics (with a generic subclass of other generic class).

mypy/checker.py
```diff
@@ -1618,19 +1628,21 @@ def check_compatibility(self, name: str, base1: TypeInfo,
...
                 isinstance(second_type, FunctionLike)):
-            # Method override
-            first_sig = bind_self(first_type)
-            second_sig = bind_self(second_type)
-            ok = is_subtype(first_sig, second_sig, ignore_pos_arg_names=True)
+            if first_type.is_type_obj() and second_type.is_type_obj():
+                # For class objects only check the subtype relationship of the classes,
+                # since we allow incompatible overrides of '__init__'/'__new__'
+                ok = is_subtype(left=fill_typevars(first_type.type_object()),
+                                right=fill_typevars(second_type.type_object()))
```
ilevkivskyi: I think using `fill_typevars` here is dangerous, since it may produce some false positives for generic classes (it creates an instance with type variable types). You need to apply `erase_typevars`. Or even better, add a helper function to `typevars.py` that will be very similar to `fill_typevars`, but would insert correct number of `Any` as type arguments.

test-data/unit/check-multiple-inheritance.test
```diff
@@ -240,3 +240,272 @@ def dec(f: Callable[..., T]) -> Callable[..., T]:
...
+
+[case testMultipleInheritance_GenericSubclasses_SuperclassFirst]
+from typing import TypeVar, Generic
+T = TypeVar('T')
+class ParentGeneric(Generic[T]):
+    pass
+class ChildGeneric(ParentGeneric, Generic[T]):
```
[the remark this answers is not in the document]
mkurnikov: No problem, I feel like I need to read pep484 and docs end to end once more, I take too much of your time. I'll fix everything till the end of the day.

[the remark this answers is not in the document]
ilevkivskyi: > I take too much of your time

Don't worry about this :-)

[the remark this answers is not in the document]
mkurnikov: Thanks!=) 
Sorry, haven't found time, but I'm sure I will in a day or two.

ilevkivskyi:
@mkurnikov What is the situation here? Is any input from my side needed?

PR #5931, opened 2018-11-21
Change some protocols into ABCs in test stubs to reflect real stubs

JukkaL:
`Sequence`, `Mapping` and `MutableMapping` are not protocols in
typeshed stubs. They shouldn't be protocols in test stubs.

### DIFF
--- a/test-data/unit/fixtures/typing-full.pyi
+++ b/test-data/unit/fixtures/typing-full.pyi
@@ -122,13 +122,11 @@ class AsyncIterator(AsyncIterable[T], Protocol):
     @abstractmethod
     def __anext__(self) -> Awaitable[T]: pass
 
-@runtime
-class Sequence(Iterable[T_co], Protocol):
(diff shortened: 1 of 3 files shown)


test-data/unit/fixtures/typing-full.pyi
```diff
@@ -122,13 +122,11 @@ class AsyncIterator(AsyncIterable[T], Protocol):
...
 
-@runtime
-class Sequence(Iterable[T_co], Protocol):
+class Sequence(Iterable[T_co]):
     @abstractmethod
     def __getitem__(self, n: Any) -> T_co: pass
 
-@runtime
-class Mapping(Iterable[T], Protocol[T, T_co], metaclass=ABCMeta):
+class Mapping(Generic[T, T_co], Iterable[T], metaclass=ABCMeta):
```
ilevkivskyi: FWIW, usually `Generic` appears as a last base.

PR #5933, opened 2018-11-21
Change TypedDict fallback to `Mapping[str, object]`

### LINKED ISSUE #5927: Unsoundness in subclassing Mappings with TypedDicts
In the course of implementing TypedDicts into Pyre, I came across an issue with how mypy is treating the relationship between TypedDicts and Mappings. 
It seems like the implemented semantics are to treat a `T1 = TypedDict[ {k_1: v_1, k_2: v:2, ... k_n: v:n} ]` as a subclass of `T2 = Mapping[str, Union[v_1, v_2, ... v_n]]`.  However, this causes a problem when combined with the structural subtyping rule that `T3 = TypedDict[ {k_1: v_1, k_2: v:2, ... k_n: v:n, k': v'} ]` is a subclass of `T1`, which implies that `T3` is a subclass of `T2`.  When you call `__getitem__` on a `T2` that's actually a `T3` with key `k'`, you would unexpectedly get a `v'`.  This can lead to an undetected potential AttributeError as in the following example:

```python
from typing import Mapping

class Y():
    def ya(self)-> int:
        return 7

class Z():
    def za(self) -> int:
        return 8

Foo = TypedDict('Foo', {'visible': Y, 'hidden': Z})

Bar = TypedDict('Bar', {'visible': Y})

def access(x:Mapping[str, Y]) -> None:
    if "hidden" in x:
        x["hidden"].ya()

def hide(x:Bar) -> None:
    access(x)

def f(x: Foo) -> None:
    hide(x)

td: Foo = {'visible': Y(), 'hidden': Z()}
f(td)
```

This passes mypy typechecking but when run gives an attribute error.

I think we're going to end up having TypedDicts just subclass `Mapping[str, Any]` for this reason.

JukkaL:
The previous fallback mapping item type was join of the value
types, which is unsafe because of structural subtyping.

Fixes #5927.

### DIFF
--- a/test-data/unit/check-typeddict.test
+++ b/test-data/unit/check-typeddict.test
@@ -6,7 +6,7 @@ Point = TypedDict('Point', {'x': int, 'y': int})
 p = Point(x=42, y=1337)
 reveal_type(p)  # E: Revealed type is 'TypedDict('__main__.Point', {'x': builtins.int, 'y': builtins.int})'
 # Use values() to check fallback value type.
-reveal_type(p.values()) # E: Revealed type is 'typing.Iterable[builtins.int*]'
+reveal_type(p.values()) # E: Revealed type is 'typing.Iterable[builtins.object*]'
 [builtins fixtures/dict.pyi]
 [typing fixtures/typing-full.pyi]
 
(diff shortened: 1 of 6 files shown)


test-data/unit/check-typeddict.test
```diff
@@ -377,25 +377,18 @@ f(ll) # E: Argument 1 to "f" has incompatible type "List[TypedDict({'x': int, 'z
...
 A = TypedDict('A', {'x': int, 'y': int})
 B = TypedDict('B', {'x': int, 'y': str})
 
-def fun(arg: StrIntMap) -> None: ...
+def fun(arg: StrObjectMap) -> None: ...
 a: A
 b: B
 fun(a)
-fun(b)  # Error
+fun(b)
```
ilevkivskyi: I would keep a test that a function that expects a protocol with `__getitem__ -> int` always fails.

PR #5934, opened 2018-11-21
Add LiteralType and stub methods to type visitors

Michael0x2a:
This diff adds a 'LiteralType' class to types.py and adds a corresponding stub method to all of the type visitors.

Most of these stub methods just throw a 'NotImplementedError', though I did pencil in a few implementations that looked obvious.

The reason I'm doing this now instead of later is because I want to start by getting these kind of big, gross, multi-file changes out of the way early and try and have any subsequent pull requests be more focused.

I also want to formally confirm that the correct approach here *is* to create a new 'LiteralType' class before I start on the rest of the implementation work. (Other possible approaches would be to tack on some extra attribute to 'Instance', or make 'LiteralType' a subclass of 'Instance'.)

No tests, sorry. My plan is to work on modifying the parsing and semantic analysis steps next before returning back to these the unimplemented methods and add tests then -- here's the tracking issue for these TODOs: https://github.com/python/mypy/issues/5935

Tests shouldn't be necessary, in any case: everything I added just now is basically dead code.

### DIFF
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -33,12 +33,15 @@ JsonDict = Dict[str, Any]
 # Literals can contain enum-values: we special-case those and
 # store the value as a string.
 #
+# TODO: confirm that we're happy with representing enums (and the
+# other types) in the manner described above.
+#
 # Note: this type also happens to correspond to types that can be
 # directly converted into JSON. The serialize/deserialize methods
 # of 'LiteralType' rely on this, as well 'server.astdiff.SnapshotTypeVisitor'
 # and 'types.TypeStrVisitor'. If we end up adding any non-JSON-serializable
 # types to this list, we should make sure to edit those methods to match.
-LiteralInnerExpr = Union[int, str, bool, None]
+LiteralValue = Union[int, str, bool, None]
 
 # If we only import type_visitor in the middle of the file, mypy
 # breaks, and if we do it at the top, it breaks at runtime because of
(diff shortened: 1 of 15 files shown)


mypy/types.py
```diff
@@ -27,6 +27,19 @@
 
 JsonDict = Dict[str, Any]
 
+# The set of all valid expressions that can currently be contained
```
JukkaL: Add `TODO` comments here about some things? At least enum support is probably not going to happen immediately, so maybe it should be as a `TODO` comment?

mypy/types.py
```diff
@@ -27,6 +27,19 @@
...
+# inside of a Literal[...].
+#
+# Literals can contain enum-values: we special-case those and
+# store the value as a string.
+#
+# Note: this type also happens to correspond to types that can be
```
JukkaL: This comment is very useful, thanks!

mypy/types.py
```diff
@@ -27,6 +27,19 @@
...
+# directly converted into JSON. The serialize/deserialize methods
+# of 'LiteralType' rely on this, as well 'server.astdiff.SnapshotTypeVisitor'
+# and 'types.TypeStrVisitor'. If we end up adding any non-JSON-serializable
+# types to this list, we should make sure to edit those methods to match.
+LiteralInnerExpr = Union[int, str, bool, None]
```
JukkaL: Would we need to be able to distinguish between `unicode` and `str` literals in Python 2? For example, the `open()` mode argument could well be a unicode literal if using `from __future__ import unicode_literals`.

A name like `LiteralValue` would perhaps be a bit more descriptive. The current name implies to me that this is an `Expression` object, which is not the case.

mypy/types.py
```diff
@@ -1240,6 +1253,43 @@ def zipall(self, right: 'TypedDictType') \
             yield (item_name, None, right_item_type)
 
 
+class LiteralType(Type):
+    __slots__ = ('value', 'fallback')
```
JukkaL: Add docstring or at least a `TODO` comment about documenting this.

mypy/types.py
```diff
@@ -1693,6 +1743,9 @@ def item_str(name: str, typ: str) -> str:
                 suffix = ', fallback={}'.format(t.fallback.accept(self))
         return 'TypedDict({}{}{})'.format(prefix, s, suffix)
 
+    def visit_literal_type(self, t: LiteralType) -> str:
+        return 'Literal[{}]'.format(t.value)
```
JukkaL: Maybe use `repr()` here so that string literals that contain unusual characters won't cause trouble. It would also match the source syntax.

JukkaL:
(Feel free to merge once the build is green.)

Issue #5942, opened 2018-11-23
Tabs replaced by '?' in HTML/XML reports


mthuurne:
When I generate an HTML or XML report with mypy, in the annotated code all tabs have been replaced by question marks. It is not just the web browser showing a question mark for an unknown character: there are literal question marks (ASCII 63) in the output.

The progress message says HTML reports are generated via XSLT, so I guess the HTML reports inherit this behavior from the XML reports.

I know that PEP8 states spaces should be used for new code, but it would be nice if mypy could be used on legacy code bases as well.

I'm using a mypy dev snapshot (0.650+) on Python 3.6.5.

mthuurne:
PR created: #5945

gvanrossum:
Thanks for the fix! (Next time put "Fixes #5942" in the commit message, instead of in the title.)

PR #5945, opened 2018-11-23
Preserve tabs when creating XML reports (#5942)

mthuurne:
Control characters are replaced with question marks, to avoid
outputting invalid XML. However, XML does allow tabs in content,
so make an exception for those.


mthuurne:
I added a tab in the test case you mentioned and changed the expected output accordingly.

PR #5965, opened 2018-11-28
Short-circuit `if` expression for always true/always false vars and MYPY/TYPE_CHECKING

### LINKED ISSUE #5963: Small bug: `if` expression doesn't short circuit for --always-false/--always-true
`if` expressions don't correctly short-circuit for variables that are set as "always true/false" constants with the `--always-false` and `--always-true` flags, even though they work for other "always true/false" conditions, like ones that use `sys.platform`.

Here's a test case:
```
# flags: --platform linux --always-false IS_THING
import sys

if sys.platform == 'darwin':
    darwin_only = 1

if sys.platform == 'darwin' and darwin_only:
    # This works because `sys.platform == 'darwin'` is inferred as `ALWAYS_FALSE`, and
    # that causes the short-circuit logic to run.
    pass

IS_THING = False

if IS_THING:
    thing_only = 1

if IS_THING and thing_only:
    # Blows up with the error `error: Name 'thing_only' is not defined` since
    # IS_THING is inferred as `MYPY_FALSE`, so the short-circuiting logic
    # doesn't run
    pass
```

The short-circuit logic is scattered throughout the codebase, but here's an [example](https://github.com/python/mypy/blob/3c311c24323d20bbf7e13084ba16bcb99e8934d7/mypy/semanal.py#L3074):

```
        if expr.op in ('and', 'or'):
            inferred = infer_condition_value(expr.left, self.options)
            if ((inferred == ALWAYS_FALSE and expr.op == 'and') or
                    (inferred == ALWAYS_TRUE and expr.op == 'or')):
                expr.right_unreachable = True
                return
            elif ((inferred == ALWAYS_TRUE and expr.op == 'and') or
                    (inferred == ALWAYS_FALSE and expr.op == 'or')):
                expr.right_always = True
```

Notice that it checks for ALWAYS_FALSE and ALWAYS_TRUE, but not MYPY_FALSE or MYPY_TRUE>

And the --always-true/false flags are translated to MYPY_TRUE/FALSE [here](https://github.com/python/mypy/blob/3c311c24323d20bbf7e13084ba16bcb99e8934d7/mypy/semanal.py#L3856).

There are two potential fixes:

1. Include two MYPY_TRUE/FALSE in all of the `if` short circuit checks.

I think that works, because even when using the MYPY variable directly, it's probably surprising that it doesn't short-circuit.

2. Use ALWAYS_TRUE/FALSE with the --always-true/false flags instead of MYPY_TRUE/FALSE.

@gvanrossum might have more context on why the --always-* flags use the MYPY_* constants instead of the ALWAYS_* constants, since he added them. FWIW, using ALWAYS_* makes more semantic sense to me, since I'm using the flags for variables that will have the same value at runtime. (e.g. I'm setting `--always-true=IS_WINDOWS` to typecheck my app for windows, so IS_WINDOWS will also be true at runtime.) That's distinct from how `MYPY` works, where it's always False at runtime and always True for typechecking.

Thoughts?

python 3.5.2, latest master for mypy

samertm:
- Always short-circuit in the if's expression, instead of only
   short-circuiting for "ALWAYS_TRUE/FALSE" constants (e.g. sys.platform
   checks) and not "MYPY_TRUE/FALSE" constants (e.g. `MYPY`)

 - Make --always-true and --always-false map to ALWAYS_TRUE and
   ALWAYS_FALSE, respectively, instead of MYPY_TRUE and MYPY_FALSE.
   The only difference between the two is a change in import priority,
   but semantically uses of "--always-true" are probably closer to
   "this condition should be true at compile time and runtime" than
   "this condition should be true at compile time but false at
   runtime".

 - Fixes #5963

### DIFF
--- a/mypy/reachability.py
+++ b/mypy/reachability.py
@@ -84,8 +84,8 @@ def infer_condition_value(expr: Expression, options: Options) -> int:
         name = expr.name
     elif isinstance(expr, OpExpr) and expr.op in ('and', 'or'):
         left = infer_condition_value(expr.left, options)
-        if ((left == ALWAYS_TRUE and expr.op == 'and') or
-                (left == ALWAYS_FALSE and expr.op == 'or')):
+        if ((left in (ALWAYS_TRUE, MYPY_TRUE) and expr.op == 'and') or
+                (left in (ALWAYS_FALSE, MYPY_FALSE) and expr.op == 'or')):
             # Either `True and <other>` or `False or <other>`: the result will
             # always be the right-hand-side.
             return infer_condition_value(expr.right, options)
@@ -105,9 +105,9 @@ def infer_condition_value(expr: Expression, options: Options) -> int:
         elif name == 'MYPY' or name == 'TYPE_CHECKING':
             result = MYPY_TRUE
         elif name in options.always_true:
-            result = MYPY_TRUE
+            result = ALWAYS_TRUE
         elif name in options.always_false:
-            result = MYPY_FALSE
+            result = ALWAYS_FALSE
     if negated:
         result = inverted_truth_mapping[result]
     return result
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -90,7 +90,8 @@ from mypy.semanal_typeddict import TypedDictAnalyzer
 from mypy.semanal_enum import EnumCallAnalyzer
 from mypy.semanal_newtype import NewTypeAnalyzer
 from mypy.reachability import (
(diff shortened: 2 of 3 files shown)


samertm:
I haven't run or tested this locally yet, I'll do that tomorrow :o)

samertm:
Just fixed the merge conflicts, should be ready to merge if the tests pass.

msullivan:
Unfortunately they did not

samertm:
Looks like tests are passing now :)

msullivan:
Thanks!

PR #5972, opened 2018-11-28
Provide an "API" for invoking the daemon

msullivan:
Like the existing "api" for invoking mypy, it is just a thin wrapper
around calling the main function.

### DIFF
--- a/mypy/api.py
+++ b/mypy/api.py
@@ -20,6 +20,9 @@ Any pretty formatting is left to the caller.
 The 'run_dmypy' function is similar, but instead mimics invocation of
 dmypy.
 
+Note that these APIs don't support incremental generation of error
+messages.
+
 Trivial example of code using this module:
 
 import sys
@@ -42,6 +45,7 @@ print ('\nExit status:', result[2])
 import sys
 from io import StringIO
 from typing import List, Tuple, Callable
+
 
 def _run(f: Callable[[], None]) -> Tuple[str, str, int]:
     old_stdout = sys.stdout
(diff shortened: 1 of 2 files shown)


mypy/api.py
```diff
@@ -60,3 +62,14 @@ def run(args: List[str]) -> Tuple[str, str, int]:
...
+
+def run(args: List[str]) -> Tuple[str, str, int]:
+    # Lazy import to avoid needing to import all of mypy to call run_dmypy
+    from mypy.main import main
+    return _run(lambda: main(None, args=args))
+
+
+def run_dmypy(args: List[str]) -> Tuple[str, str, int]:
+    from mypy.dmypy import main
+    return _run(lambda: main(args))
```
ilevkivskyi: IIUC this doesn't support incremental flushing of errors (maybe we will do this at some point), but it is the same for normal runs, so probably OK. Maybe add a comment about this.

mypy/dmypy.py
```diff
@@ -114,16 +115,21 @@ class BadStatus(Exception):
...
     else:
         try:
             args.action(args)
         except BadStatus as err:
             sys.exit(err.args[0])
+        except Exception:
+            # We do this explicitly to avoid exceptions percolating up
+            # through mypy.api invocations
+            traceback.print_exc()
+            sys.exit(2)
```
ilevkivskyi: I like this choice, I think there are still some situations where 1 can be returned on a traceback in the server, but I will fix those separately later.

PR #5975, opened 2018-11-29
Support for overloaded functions in stubgenc generated by pybind11

wiktorn:
For overloaded methods pybind generates following docstring:
 ```
__init__(*args, **kwargs)
Overloaded function.
 
1. __init__(self: TestClass, arg0: str) -> None

2. __init__(self: TestClass, arg0: str, arg1: str) -> None
```
For such docstrings stubgenc currently produces following stub:
```
def __init__(self, *args, **kwargs) -> Any: ...
```
With this change stubgenc will generate following stub:
```
@overload
def __init__(self, arg0: str) -> None: ...
@overload
def __init__(self, arg0: str, arg1: str) -> None: ...
```

This pull request introduces:
* class - FunctionSig - representing signature of function/method (function name, arguments, return value)
* class - ArgSig  - representing information about function argument (argument name, type, default value)

`infer_sig_from_docstring` now returns list of `FunctionSig` . List contains one object when function is not overloaded, more objects - when it's overloaded. Each object represents signature of function.

`infer_arg_sig_from_docstring` is a new function that parses argument list for function in docstring. As the type information may contain commas, the code checks, if the comma is outside brackets. I tried to approach that with regex and failed, also tried to use `ast` module, but its not straightforward to move from ast form back to source code form that's why I settled with naive approach, as this is not performance critical path. Tests for edge cases included in `test_infer_sig_from_docstring`

### CODE REVIEW

**Comment by gvanrossum** [comment]
> For such docstrings stubgenc will produce following stub:

I presume you mean that stubgenc *should* produce such a stub, and this PR implements that behavior? (We prefer our commit messages to describe the change rather than the new state of affairs.)

I will try to review now, but I think this has missed the train for the upcoming 0.650 release (see #5960).

**Comment by wiktorn** [comment]
Ok, updated description.

**Comment by gvanrossum** [comment]
Hm... Having looked at this a bit more, it really feels like string partitioning is a pretty poor way of parsing something of this complexity. I realize that all of stubgen looks like a hack, but since you are doing your best to make it better, perhaps you can improve the approach to parsing a bit more? Maybe write a tiny recursive-descent parser, using the `tokenize` module for tokenization?

**Comment by wiktorn** [comment]
Thank you for pointing me to `tokenize` module. Looks like this is something I was looking for. Looks like it is time for me to refresh my knowledge and re-read library reference :-)

I'll start with replacing parsing docstring function declarations. Though quick scan of the code suggests, that some regex usage will remain in the stubgenc/stubutil.

**Comment by gvanrossum** [comment]
Sounds like a plan!

**Comment by gvanrossum** [comment]
Hey @wiktorn, just to let you know, we're planning fairly extensive refactorings of part of stubgen, however, we don't expect the docstring parsing to be affected, so don't worry!

**Comment by wiktorn** [comment]
Thank you for the heads up @gvanrossum . I'll finally have more time next week to polish this PR.

**Comment by wiktorn** [comment]
@gvanrossum Can you trigger another AppVeyor build? My guess that this failure was totally unrelated to my changes (some exceptions in IPC and ast3 module that are not mentioned in this changes)

**Comment by ethanhs** [comment]
@wiktorn if you rebase on master those errors should go away.

File: mypy/stubutil.py [hunk 1]
```diff
@@ -106,32 +123,146 @@ def write_header(file: IO[str], module_name: Optional[str] = None,
...
-        # strip potential white spaces at the right of return type
-        return m.group(1), m.group(2).rstrip()
-
-    # try to not match return type
-    m = re.match(sig_match, docstr)
-    if m:
-        return m.group(1), 'Any'
-    return None
+
+    state = [State.INIT, ]
```
gvanrossum: No need for trailing comma (and space) here.

mypy/stubutil.py
```diff
@@ -106,32 +123,146 @@ def write_header(file: IO[str], module_name: Optional[str] = None,
...
+    ret_type = "Any"
+    found = False
+    args = []  # type: List[TypedArgSig]
+    signatures = []  # type: List[TypedFunctionSig]
+    try:
+        for token in tokenize.tokenize(io.BytesIO(docstr.encode('utf-8')).readline):
+            if token.type == tokenize.NAME and token.string == name and state[-1] == State.INIT:
+                state.append(State.FUNCTION_NAME)
+
+            elif token.type == tokenize.OP and token.string == '(' and state[-1] == \
```
gvanrossum: Please don't use `\` this way (also below). Surround the outermost expression with `()` and break only at `and`.

mypy/stubutil.py
```diff
@@ -106,32 +123,146 @@ def write_header(file: IO[str], module_name: Optional[str] = None,
...
+                accumulator += token.string
+                state.pop()
+
+            elif token.type == tokenize.OP and token.string == ':' and \
+                    state[-1] == State.ARGUMENT_LIST:
+                arg_name = accumulator
+                accumulator = ""
+                state.append(State.ARGUMENT_TYPE)
+
+            elif token.type == tokenize.OP and token.string == '=' and state[-1] in (
```
gvanrossum: I'd rather do the same reformatting here (and again below).

mypy/stubgenc.py
```diff
@@ -123,51 +124,47 @@ def generate_c_function_stub(module: ModuleType,
...
-    else:
-        self_arg = ''
     if (name in ('__new__', '__init__') and name not in sigs and class_name and
             class_name in class_sigs):
-        sig = class_sigs[class_name]
+        inferred = [TypedFunctionSig(
+            name=name,
+            args=infer_arg_sig_from_docstring(class_sigs[class_name]),
+            ret_type=ret_type
+        )]  # type: Optional[List[TypedFunctionSig]]
```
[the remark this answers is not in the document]
wiktorn: done

mypy/stubgenc.py
```diff
@@ -123,51 +124,47 @@ def generate_c_function_stub(module: ModuleType,
...
-                sig_types.append(arg_type[0].strip())
-            else:
-                arg_type_name = strip_or_import(arg_type[1].strip(), module, imports)
-                sig_types.append('%s: %s' % (arg_type[0].strip(), arg_type_name))
-        sig = ", ".join(sig_types)
+                inferred = [TypedFunctionSig(
+                    name=name,
+                    args=infer_arg_sig_from_docstring(sigs.get(name, '(*args, **kwargs)')),
+                    ret_type=ret_type
+                )]
```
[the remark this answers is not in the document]
wiktorn: done

mypy/stubutil.py
```diff
@@ -121,19 +136,81 @@ def infer_sig_from_docstring(docstr: str, name: str) -> Optional[Tuple[str, str]
...
+def infer_arg_sig_from_docstring(docstr: str) -> ArgList:
+    """
+    convert signature in form of "(self: TestClass, arg0: str='ada')" to ArgList
+
+    :param docstr:
+    :return: ArgList with infered argument names and its types
+    """
+    ret = []  # type: ArgList
+    arguments = []
+    right = docstr[1:-1]
```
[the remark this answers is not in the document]
wiktorn: I've refactor this totally to use `infer_sig_from_docstring` instead of own implementation

mypy/stubutil.py
```diff
@@ -121,19 +136,81 @@ def infer_sig_from_docstring(docstr: str, name: str) -> Optional[Tuple[str, str]
...
+    accumulator = ""
+    while right:
+        left, sep, right = right.partition(',')
+        if right.count('[') == right.count(']'):
```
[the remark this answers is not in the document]
wiktorn: Good catch, as above.

mypy/stubutil.py
wiktorn: done

mypy/stubutil.py
wiktorn: done

wiktorn:
I hope that's ok right now. Just let me know how you want to move forward. Shall I till #6256 lands in master and merge master back to this branch?

Sorry for the force push, that's how I understood @ethanhs  and since there was no comments on code yet, I've decided to rebase instead of merging master

mypy/test/teststubgen.py
```diff
@@ -310,3 +462,114 @@ def test(self, arg0: str) -> None:
...
+                                 self_var='self', class_name='TestClass')
+        assert_equal(output, [
+            '@overload',
+            'def __init__(*args, **kwargs) -> Any: ...',
+            '@overload',
+            'def __init__(self, arg0: str) -> None: ...',
+            '@overload',
+            'def __init__(self, arg0: str, arg1: str) -> None: ...',
+        ])
+        assert_equal(set(imports), {
```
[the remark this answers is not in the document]
wiktorn: In generated stub will not be repeated, because when stub is generated (in `generate_stub_for_c_module`), it's made unique:
```
for line in sorted(set(imports)):
```

I could store it in set anyway, though it would introduce some unrelated changes.

mypy/test/teststubgen.py
```diff
@@ -310,3 +462,114 @@ def test(self, arg0: str) -> None:
...
+        ])
+        assert_equal(imports, [
+            'import argparse'
+        ])
+
+    def test_generate_c_function_same_module_arg(self) -> None:
+        """
+        Test that if argument references type from same module but using full path, no module will
+        be imported, and type specification will be striped to local reference
+        """
```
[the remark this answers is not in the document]
wiktorn: I hope I've got all of them

mypy/stubgenc.py
```diff
@@ -307,29 +302,38 @@ def is_skipped_attribute(attr: str) -> bool:
...
+def infer_method_sig(name: str) -> List[TypedArgSig]:
     if name.startswith('__') and name.endswith('__'):
         name = name[2:-2]
         if name in ('hash', 'iter', 'next', 'sizeof', 'copy', 'deepcopy', 'reduce', 'getinitargs',
                     'int', 'float', 'trunc', 'complex', 'bool'):
-            return '()'
+            return []
         if name == 'getitem':
-            return '(index)'
+            return [TypedArgSig(name='index', type=None, default=False)]
```
ilevkivskyi: If you would use normal classes instead of named tuples you could define `__init__` as:
```python
    def __init__(self, name: str, type: Optional[str] = None, default: bool = False) -> None: ...
```
and then this an d a dozen others will be just `TypedArgSig('index')`.

mypy/stubgenc.py
```diff
@@ -307,29 +302,38 @@ def is_skipped_attribute(attr: str) -> bool:
...
+            ]
         if name in ('delattr', 'getattr'):
-            return '(name)'
+            return [TypedArgSig(name='name', type=None, default=False)]
         if name == 'setattr':
-            return '(name, value)'
+            return [
+                TypedArgSig(name='name', type=None, default=False),
+                TypedArgSig(name='value', type=None, default=False)
+            ]
```
ilevkivskyi: For multiline we use this style:
```python
return [TypedArgSig(name='name', type=None, default=False),
        TypedArgSig(name='value', type=None, default=False)]
```
(also many of these will not need to be multiline if you implement the suggestion above).

mypy/test/teststubgen.py
```diff
@@ -103,26 +104,170 @@ def test_find_unique_signatures(self) -> None:
...
+            infer_sig_from_docstring('\nfunc(x, Y_a=3)', 'func'),
+            [TypedFunctionSig(
+                name='func',
+                args=[
+                    TypedArgSig(name='x', type=None, default=False),
+                    TypedArgSig(name='Y_a', type=None, default=True)
+                ],
+                ret_type='Any'
+            )]
+        )
```
ilevkivskyi: As I said we use different style for multiline displays:
```python
        [TypedFunctionSig(name='func',
                          args=[TypedArgSig(name='x', type=None, default=False),
                                TypedArgSig(name='Y_a', type=None, default=True)],
                          ret_type='Any')]
```
this is much more compact.

mypy/stubgenc.py
ilevkivskyi: Also with normal classes you can define `__repr__()`, to simplify test cases significantly. You can just check that `str(<generated signature>)` matches an expected string.

wiktorn:
I've missed your comments on conversation tab on GitHub. As I reviewed changes I found your comments and started to address them.

wiktorn:
And what do you think about this approach:
```
ArgSig = NamedTuple('ArgSig', [
    ('name', str),
    ('type', Optional[str]),
    ('default', bool)
])

ArgSig.__new__.__defaults__ = (None, False)
```

As far as I see, it works in Python3.4, and once oldest supported version will be 3.7, we can move this declaration to NamedTuple call itself.

I'm reluctant to use `repr` strings in tests though.

ilevkivskyi:
> `ArgSig.__new__.__defaults__ = (None, False)`

With this call sites will not type-check by mypy.

> I'm reluctant to use `repr` strings in tests though.

Why? If you just want to be sure it is clear from the string form it is a custom class, use something like `"FunctionSig('method(self, arg: int) -> Any')"` it is still quicker to grasp that the current form. I however don't have strong opinion about this, unlike about the default arguments.

wiktorn:
> 
> 
> > `ArgSig.__new__.__defaults__ = (None, False)`
> 
> With this call sites will not type-check by mypy.

As I'm testing it right now, it rather doesn't pass self-check:
`error: "Callable[[Type[NT], str, Optional[str], bool], NT]" has no attribute "__defaults__"`

Though it properly detects problems in call sites
```
error: Argument "type" to "ArgSig" has incompatible type "int"; expected "Optional[str]"
error: Argument "default" to "ArgSig" has incompatible type "str"; expected "bool"
```

So I'll convert them to normal classes.

> > I'm reluctant to use `repr` strings in tests though.
> 
> Why? If you just want to be sure it is clear from the string form it is a custom class, use something like `"FunctionSig('method(self, arg: int) -> Any')"` it is still quicker to grasp that the current form. I however don't have strong opinion about this, unlike about the default arguments.

We could actually have `__str__` method which would provide stub signature. The only problem I see is that to cover whole API contract, we would need still to check if types in ArgSig's are properly set, as they are used to add necessary imports

ilevkivskyi:
> Though it properly detects problems in call sites

I meant that mypy will flag as errors the calls with less arguments like `ArgSig('index')` if you would use the `__defaults__` pattern.

> So I'll convert them to normal classes.

OK

> The only problem I see is that to cover whole API contract, we would need still to check if types in ArgSig's are properly set, as they are used to add necessary imports

Yes, this could be useful only for testing. Anyway, I don't insist on this. You can just defined `__eq__()` instead and keep your current tests.

wiktorn:
I did not add defaults to `FunctionSig` as it doesn't shorten that much and forces to give a thought, if arguments are properly set.

PR #5976, opened 2018-11-30
Default to wait forever with Windows IPC

ethanhs:
This is to hopefully fix flakes when running tests in slow CI...


msullivan:
Well that didn't work.

ethanhs:
note: merging even though it didn't fix the problem because as Sully said, it makes Windows more in line with Unix. 

I'm currently trying to debug what is actually going wrong.

PR #5982, opened 2018-11-30
Switch away from sys.exit(error_msg) in dmypy

msullivan:
This kills a couple birds with one stone:
 1. It fixes an interaction with the new run_dmypy api where
    that message winds up as the return code instead of in
    stderr.
    This additionally fixes a type unsoundness/mypy_mypyc crash
    caused by SystemExit.code being typed as int.
 2. It allows us to standardize the irregular exit codes of dmypy
    as 2.

### DIFF
--- a/mypy/dmypy.py
+++ b/mypy/dmypy.py
@@ -124,12 +124,17 @@ def main(argv: List[str]) -> None:
         try:
             args.action(args)
         except BadStatus as err:
-            sys.exit(err.args[0])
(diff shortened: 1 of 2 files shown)


JukkaL:
Should we document the exit codes?

ilevkivskyi:
> Should we document the exit codes?

Opened https://github.com/python/mypy/issues/6003 for this.

PR #5985, opened 2018-12-01
Delete dmypy script -- it no longer works and is not used

gvanrossum:


JelleZijlstra:
@msullivan was faster—see #5984.

gvanrossum:
Oops.

PR #5987, opened 2018-12-01
Use exit 2 on crash to signal this is no ordinary error

gvanrossum:
The wrapper script we use to run mypy at Dropbox will greatly benefit from this -- it will be able to distinguish between a crash and type-checking errors.


gvanrossum:
> Use an "abnormal" return value like 255 for crashes.

I'm not a big fan of that -- there's no standard for such error codes, while using 2 for "not even wrong" types of errors is at least an old tradition. See also #5982.

PR #5990, opened 2018-12-03
Make literal exprs have inferred type of 'Literal' based on context

Michael0x2a:
This pull request modifies the type checking logic so that literal expressions will have an inferred type of 'Literal' if the context asks for a literal type. That is, it implements support for this:

    x: Literal[1] = 1
    y = 1

    reveal_type(x)  # E: Revealed type is 'Literal[1]'
    reveal_type(y)  # E: Revealed type is 'builtins.int'

This pull requests also implements the `visit_literal_type` method in the  `constraints.ConstraintBuilderVisitor` and `join.TypeJoinVisitor` methods. Both visitors are exercised indirectly through the "let's use literal types in collection contexts" code, but only the latter is tested directly: I wasn't really sure how to directly test `ConstraintBuilderVisitor`.

The implementation is simple though -- I'm pretty sure literal types count as a "leaf type" so it's fine to return an empty list (no constraints).

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -18,7 +18,7 @@ from mypy.typeanal import (
 from mypy.types import (
     Type, AnyType, CallableType, Overloaded, NoneTyp, TypeVarDef,
     TupleType, TypedDictType, Instance, TypeVarType, ErasedType, UnionType,
-    PartialType, DeletedType, UninhabitedType, TypeType, TypeOfAny,
+    PartialType, DeletedType, UninhabitedType, TypeType, TypeOfAny, LiteralType,
     true_only, false_only, is_named_instance, function_type, callable_type, FunctionLike,
     StarType, is_optional, remove_optional, is_invariant_instance
 )
@@ -210,7 +210,10 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
     def analyze_var_ref(self, var: Var, context: Context) -> Type:
         if var.type:
-            return var.type
+            if is_literal_type_like(self.type_context[-1]) and var.name() in {'True', 'False'}:
+                return LiteralType(var.name() == 'True', self.named_type('builtins.bool'))
+            else:
+                return var.type
         else:
             if not var.is_ready and self.chk.in_checked_function():
                 self.chk.handle_cannot_determine_type(var.name(), context)
@@ -1721,11 +1724,17 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
     def visit_int_expr(self, e: IntExpr) -> Type:
         """Type check an integer literal (trivial)."""
-        return self.named_type('builtins.int')
+        typ = self.named_type('builtins.int')
+        if is_literal_type_like(self.type_context[-1]):
+            return LiteralType(value=e.value, fallback=typ)
+        return typ
 
     def visit_str_expr(self, e: StrExpr) -> Type:
         """Type check a string literal (trivial)."""
-        return self.named_type('builtins.str')
+        typ = self.named_type('builtins.str')
+        if is_literal_type_like(self.type_context[-1]):
+            return LiteralType(value=e.value, fallback=typ)
+        return typ
 
     def visit_bytes_expr(self, e: BytesExpr) -> Type:
         """Type check a bytes literal (trivial)."""
@@ -3583,3 +3592,17 @@ def merge_typevars_in_callables_by_name(
         output.append(target)
 
     return output, variables
+
+
+def is_literal_type_like(t: Optional[Type]) -> bool:
+    """Returns 'true' if the given type context is potentially either a LiteralType,
+    a Union of LiteralType, or something similar.
+    """
+    if t is None:
+        return False
+    elif isinstance(t, LiteralType):
+        return True
+    elif isinstance(t, UnionType):
+        return any(map(is_literal_type_like, t.items))
+    else:
+        return False
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3603,6 +3603,6 @@ def is_literal_type_like(t: Optional[Type]) -> bool:
     elif isinstance(t, LiteralType):
         return True
     elif isinstance(t, UnionType):
-        return any(map(is_literal_type_like, t.items))
+        return any(is_literal_type_like(item) for item in t.items)
     else:
         return False
--- a/mypy/test/testtypes.py
+++ b/mypy/test/testtypes.py
@@ -249,11 +249,11 @@ class TypeOpsSuite(Suite):
         fx = self.fx
 
         lit1 = LiteralType(1, fx.a)
-        lit2 = LiteralType("foo", fx.b)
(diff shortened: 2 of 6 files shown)


Michael0x2a:
Note to self: add tests here making sure inference works as expected with overloads.

I guess maybe with generics as well, while we're at it? Or maybe in a follow-up diff specifically about generics.

mypy/checkexpr.py
```diff
@@ -3583,3 +3592,17 @@ def merge_typevars_in_callables_by_name(
...
+def is_literal_type_like(t: Optional[Type]) -> bool:
+    """Returns 'true' if the given type context is potentially either a LiteralType,
+    a Union of LiteralType, or something similar.
+    """
+    if t is None:
+        return False
+    elif isinstance(t, LiteralType):
+        return True
+    elif isinstance(t, UnionType):
+        return any(map(is_literal_type_like, t.items))
```
JukkaL: Avoid the use of `map()` since mypyc can't compile it efficiently. Use a generator expression instead.

mypy/test/testtypes.py
```diff
@@ -621,6 +621,29 @@ def test_type_type(self) -> None:
...
+        self.assert_join(lit1, self.fx.anyt, self.fx.anyt)
+        self.assert_join(self.fx.anyt, lit1, self.fx.anyt)
+
+        self.assert_simple_join(lit1, lit1, lit1)
+        self.assert_simple_join(lit1, a, a)
+        self.assert_simple_join(a, lit1, a)
+        self.assert_simple_join(lit1, lit2, a)
+        self.assert_simple_join(lit1, lit3, self.fx.o)
+        self.assert_simple_join(lit1, self.fx.anyt, self.fx.anyt)
+        self.assert_simple_join(self.fx.anyt, lit1, self.fx.anyt)
```
JukkaL: Test also joins with unions containing literals types.

test-data/unit/check-literal.test
```diff
@@ -632,6 +669,355 @@ Bar2 = Literal[14]
...
+reveal_type(bool1)  # E: Revealed type is 'Literal[True]'
+reveal_type(bool2)  # E: Revealed type is 'builtins.bool'
+reveal_type(bool3)  # E: Revealed type is 'builtins.bool'
+reveal_type(none1)  # E: Revealed type is 'None'
+reveal_type(none2)  # E: Revealed type is 'None'
+reveal_type(none3)  # E: Revealed type is 'None'
+[builtins fixtures/primitives.pyi]
+[out]
+
+[case testLiteralInferredOnlyForActualLiterals]
```
JukkaL: Add similar tests for unions of literal types.

test-data/unit/check-literal.test
```diff
@@ -632,6 +669,355 @@ Bar2 = Literal[14]
...
+y = g
+z = h
+
+[builtins fixtures/primitives.pyi]
+[out]
+
+[case testLiteralInferredTypeMustMatchExpected]
+from typing_extensions import Literal
+
+a: Literal[1] = 2           # E: Incompatible types in assignment (expression has type "Literal[2]", variable has type "Literal[1]")
```
JukkaL: Add similar tests for unions of literal types.

test-data/unit/check-literal.test
```diff
@@ -197,10 +211,26 @@ from typing_extensions import Literal
...
+def f2(x: Literal["  foo bar  "]) -> Literal["  foo bar  "]: pass
+def f3(x: Literal['  foo bar  ']) -> Literal['  foo bar  ']: pass
+def f4(x: Literal["foo"]) -> Literal["foo"]: pass
+def f5(x: Literal['foo']) -> Literal['foo']: pass
+
+reveal_type(f1)  # E: Revealed type is 'def (x: Literal['']) -> Literal['']'
+reveal_type(f2)  # E: Revealed type is 'def (x: Literal['  foo bar  ']) -> Literal['  foo bar  ']'
+reveal_type(f3)  # E: Revealed type is 'def (x: Literal['  foo bar  ']) -> Literal['  foo bar  ']'
+reveal_type(f4)  # E: Revealed type is 'def (x: Literal['foo']) -> Literal['foo']'
+reveal_type(f5)  # E: Revealed type is 'def (x: Literal['foo']) -> Literal['foo']'
```
JukkaL: Test calls to a function accepting a union of string literal types. Call it with each supported literal and also union argument types.

test-data/unit/check-literal.test
```diff
@@ -632,6 +669,355 @@ Bar2 = Literal[14]
...
+@overload
+def f(x: str, y: Literal['b']) -> str: ...
+def f(x):
+    pass
+
+x: Literal['a', 'b']
+y: Literal['a', 'b']
+f(x, y)  # E: Argument 1 to "f" has incompatible type "Union[Literal['a'], Literal['b']]"; expected "Literal['a']" \
+         # E: Argument 2 to "f" has incompatible type "Union[Literal['a'], Literal['b']]"; expected "Literal['a']" \
+[out]
```
JukkaL: Test overload where there are no literal types and one of the overload items is `T` -> `T`. This shouldn't produce literal types when called with a literal argument.

Test overload where one item accepts a literal `'x'` and another takes `T` -> `T`. When called with an unrelated literal `'y'`, will this produce a literal type or a non-literal type?

test-data/unit/check-literal.test
```diff
@@ -632,6 +669,355 @@ Bar2 = Literal[14]
...
+
+@overload
+def f(x: Literal[1, 2]) -> int: ...
+@overload
+def f(x: int) -> object: ...
+def f(x):
+    pass
+
+x: Literal[1, 2]
+reveal_type(f(x))  # E: Revealed type is 'builtins.int'
```
JukkaL: Also test `f(1)` and `f(2)`, just in case. What about calling with `Literal[1, 2, 3]`?

mypy/test/testtypes.py
```diff
@@ -621,6 +621,29 @@ def test_type_type(self) -> None:
...
+    def test_literal_type(self) -> None:
+        a = self.fx.a
+        d = self.fx.d
+        lit1 = LiteralType(1, a)
+        lit2 = LiteralType(2, a)
+        lit3 = LiteralType("foo", d)
+
+        self.assert_join(lit1, lit1, lit1)
+        self.assert_join(lit1, a, a)
+        self.assert_join(a, lit1, a)
```
JukkaL: Test join of `lit1` and `d`.

test-data/unit/check-literal.test
Michael0x2a: Added to a test named 'testLiteralCallingUnionFunction'.

mypy/test/testtypes.py
Michael0x2a: Done. I also didn't see until now that `assert_join` internally already tests symmetry and calls `assert_simple_join`, so I went ahead and removed some existing duplicate tests.

Michael0x2a:
@JukkaL -- this should be ready for a second look whenever!

PR #5997, opened 2018-12-03
Trigger mypy_mypyc builds on pushes to release branches

msullivan:


msullivan:
I think this needs to be cherry-picked into the release branch in order for it to work for this release

gvanrossum:
(Does this mean there's another pin move and MM update in my future?)

msullivan:
No, this is purely an infra change

PR #6004, opened 2018-12-04
Improve documentation of ABCs

JukkaL:
It's a common error to forget to implement an abstract
method, and mypy doesn't immediately generate an error.
Explain this explicitly in the documentation, since this
behavior can be confusing.

### DIFF
--- a/docs/source/class_basics.rst
+++ b/docs/source/class_basics.rst
@@ -159,19 +159,19 @@ override has a compatible signature:
 
 .. code-block:: python
 
-   class A:
+   class Base:
        def f(self, x: int) -> None:
            ...
 
-   class B(A):
+   class Derived1(Base):
        def f(self, x: str) -> None:   # Error: type of 'x' incompatible
            ...
 
-   class C(A):
+   class Derived2(Base):
        def f(self, x: int, y: int) -> None:  # Error: too many arguments
            ...
 
-   class D(A):
+   class Derived3(Base):
        def f(self, x: int) -> None:   # OK
            ...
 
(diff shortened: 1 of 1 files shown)


docs/source/class_basics.rst
```diff
@@ -194,50 +194,96 @@ effect at runtime:
...
 **********************************************
 
 Mypy supports Python abstract base classes (ABCs). Abstract classes
 have at least one abstract method or property that must be implemented
-by a subclass. You can define abstract base classes using the
-``abc.ABCMeta`` metaclass, and the ``abc.abstractmethod`` and
-``abc.abstractproperty`` function decorators. Example:
+by any *concrete* (non-abstract) subclass. You can define abstract base
+classes using the ``abc.ABCMeta`` metaclass, and the ``abc.abstractmethod``
+and ``abc.abstractproperty`` function decorators. Example:
```
gvanrossum: Actually `@abstractproperty` has been deprecated since 3.3 -- just put`@property` above `@abstractmethod`: https://docs.python.org/3/library/abc.html?highlight=abstractproperty#abc.abstractproperty . (However, in 2.7 you still need `@abstractproperty`.)

docs/source/class_basics.rst
```diff
@@ -194,50 +194,96 @@ effect at runtime:
...
-       def foo(self, x: int) -> None: pass
+       def eat(self, food: str) -> None: pass
 
        @abstractmethod
-       def bar(self) -> str: pass
+       def can_walk(self) -> bool: pass
+
+   class Cat(Animal):
+       def eat(self, food: str) -> None:
+           ...  # Implementation omitted
```
gvanrossum: This comment is confusing, since the point of this example is about whether a method is *implemented* in a subclass or not. If you think a comment is needed, you could use "Body omitted".

docs/source/class_basics.rst
```diff
@@ -194,50 +194,96 @@ effect at runtime:
...
+
+.. note::
+
+   It's a common error to forget to implement an abstract method.
+   As shown above, the class definition will not generate an error
+   in this case, but any attempt to construct an instance will be
+   flagged as an error.
+
 A class can inherit any number of classes, both abstract and
 concrete. As with normal overrides, a dynamically typed method can
```
gvanrossum: At some point (not now) I would like to overhaul the docs and somehow change the terminology regarding statically typed and dynamically typed functions and methods. There are flags (e.g. `--check-untyped-defs`) that blur the difference, and when we talk amongst ourselves or with users we never describe the distinction -- instead we'll use "unannotated" (even though my spell checker doesn't think that's a word :-). It's also not clear to me whether an unannotated method can still override an annotated method if the signatures differ wildly (e.g. no correspondence in argument count).

PR #6007, opened 2018-12-04
Sync typeshed

JukkaL:
In particular, this includes https://github.com/python/typeshed/pull/2670
that is needed for improved TypedDict method support (happening in
a separate PR).


JukkaL:
Appveyor seems stuck -- I'm going to ignore it.

PR #6009, opened 2018-12-04
[daemon] add --status-file arg to dmypy

jhance:
Adds an argument to avoid hardcoding the path to the status file. The intent is to allow for multiple daemons running at the same time with different arguments. This is must for running both py2 and py3 mode during development of straddled code.


msullivan:
It looks like this broke windows, though

jhance:
I think I found the root cause in a typo. However, I can't test the changes on windows locally, so I apologize ahead of time for any noise here from running the tests only in CI.

PR #6011, opened 2018-12-05
Add support for additional TypedDict methods

### LINKED ISSUE #3550: TypedDict and del
Mypy disallows `del td[key]` for typed dicts. However, it should be okay to delete a key from typed dict if the key exists and it's not a required key. `del` with a non-literal key should always be rejected.

### LINKED ISSUE #3843: TypedDict missing many dict methods
mypy isn't recognizing methods of TypedDict.

- mypy 0.521
- mypy_extensions 0.3.0
- typeshed 0.0.1

```python
D = TypedDict('D', {'a': int})

d = {'a': 4}  # type: D
d.copy()
d.get('a')
d.clear()
d.setdefault('b', 7)
d.has_key('a')
d.items()
d.keys()
d.values()
d.iteritems()
d.iterkeys()
d.itervalues()
d.pop('b', None)
d.popitem()
d.update({'f': 11})
d.viewitems()
d.viewkeys()
d.viewvalues()

"""
$ mypy -2 example.py
example.py:7: error: "D" has no attribute "copy"
example.py:9: error: "D" has no attribute "clear"
example.py:10: error: "D" has no attribute "setdefault"
example.py:11: error: "D" has no attribute "has_key"
example.py:18: error: "D" has no attribute "pop"
example.py:19: error: "D" has no attribute "popitem"
example.py:20: error: "D" has no attribute "update"
example.py:21: error: "D" has no attribute "viewitems"
example.py:22: error: "D" has no attribute "viewkeys"
example.py:23: error: "D" has no attribute "viewvalues"
"""
```

JukkaL:
This adds support for these methods through additional plugin hooks:

* `pop`
* `setdefault`
* `update` (positional argument only)
* `__delitem__`

These methods also work and don't need plugin support:

* `copy`
* `has_key` (Python 2 only)
* `viewitems` (Python 2 only)
* `viewkeys` (Python 2 only)
* `viewvalues` (Python 2 only)

The base signatures for all of these methods are defined in 
`mypy_extensions._TypedDict`, which is a stub-only class
only used internally by mypy. It becomes the new fallback
type of all TypedDicts.

Fixes #3843. Fixes #3550.

There's some possible follow-up work that I'm leaving to other PRs,
such as optimizing hook lookup through dictionaries in the default
plugin and documenting the supported methods.


mypy/plugin.py
```diff
@@ -544,6 +563,136 @@ def typed_dict_get_callback(ctx: MethodContext) -> Type:
...
+                ret_type=typ)
+    return signature.copy_modified(arg_types=[str_type, signature.arg_types[1]])
+
+
+def typed_dict_pop_callback(ctx: MethodContext) -> Type:
+    """Type check and infer a precise return type for TypedDict.pop."""
+    if (isinstance(ctx.type, TypedDictType)
+            and len(ctx.arg_types) >= 1
+            and len(ctx.arg_types[0]) == 1):
+        if isinstance(ctx.args[0][0], StrExpr):
```
ilevkivskyi: At some point we should also start using "constant folding", not in this PR however.

mypy/plugin.py
```diff
@@ -544,6 +563,136 @@ def typed_dict_get_callback(ctx: MethodContext) -> Type:
...
+    """Try to infer a better signature type for TypedDict.update."""
+    signature = ctx.default_signature
+    if (isinstance(ctx.type, TypedDictType)
+            and len(signature.arg_types) == 1):
+        arg_type = signature.arg_types[0]
+        assert isinstance(arg_type, TypedDictType)
+        arg_type = arg_type.as_anonymous()
+        arg_type = arg_type.copy_modified(required_keys=set())
+        return signature.copy_modified(arg_types=[arg_type])
+    return signature
```
ilevkivskyi: Maybe add a TODO somewhere for `.update(x=1, y=2)`? Or is this tracked in an issue?

mypy/semanal_typeddict.py
```diff
@@ -276,9 +276,7 @@ def fail_typeddict_arg(self, message: str,
     def build_typeddict_typeinfo(self, name: str, items: List[str],
                                  types: List[Type],
                                  required_keys: Set[str]) -> TypeInfo:
-        fallback = (self.api.named_type_or_none('typing.Mapping',
-                                                [self.api.named_type('__builtins__.str'),
-                                                 self.api.named_type('__builtins__.object')])
+        fallback = (self.api.named_type_or_none('mypy_extensions._TypedDict', [])
```
ilevkivskyi: I would fail soon here if we can't find `mypy_extensions._TypedDict`. There is some logic than may cause a crash/fail that will be harder to debug.

mypy/types.py
```diff
@@ -1172,12 +1172,20 @@ def slice(self, begin: Optional[int], stride: Optional[int],
...
+    are normal dict objects at runtime.
+
+    A TypedDictType can be either named or anonymous. If it's anonymous, its
+    fallback will mypy_extensions._TypedDict (Instance). _TypedDict is a subclass
+    of Mapping[str, object] and defines all non-mapping dict methods that TypedDict
+    supports. Some dict methods are unsafe and not supported. _TypedDict isn't defined
+    at runtime.
+
+    If a TypedDict is named, its fallback will be an Instance of the named type
+    (ex: "Point") whose TypeInfo has a typeddict_type that is anonymous.
```
ilevkivskyi: Maybe mention this is similar to how named tuples work and/or add some motivation for such complex structure?

runtests.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 from os import system
-from sys import argv, exit, platform, executable, version_info
+from sys import argv, exit, platform, executable, version_info, stdout
```
ilevkivskyi: A redundant import?

test-data/unit/check-typeddict.test
```diff
@@ -1481,3 +1481,85 @@ f1(**a, **c) # E: "f1" gets multiple values for keyword argument "x" \
...
+[case testTypedDictNonMappingMethods]
+from typing import List
+from mypy_extensions import TypedDict
+
+A = TypedDict('A', {'x': int, 'y': List[int]})
+a: A
+
+reveal_type(a.copy()) # E: Revealed type is 'TypedDict('__main__.A', {'x': builtins.int, 'y': builtins.list[builtins.int]})'
+a.has_key() # E: "A" has no attribute "has_key"
+a.clear() # E: "A" has no attribute "clear"
```
ilevkivskyi: Another possible TODO is to give a better error message here.

test-data/unit/check-typeddict.test
```diff
@@ -1481,3 +1481,85 @@ f1(**a, **c) # E: "f1" gets multiple values for keyword argument "x" \
...
+d = {'x': 1}
+a.update(d) # E: Argument 1 to "update" of "TypedDict" has incompatible type "Dict[str, int]"; expected "TypedDict({'x'?: int, 'y'?: List[int]})"
+[builtins fixtures/dict.pyi]
+
+[case testTypedDictNonMappingMethods_python2]
+from mypy_extensions import TypedDict
+A = TypedDict('A', {'x': int})
+a = A(x=1)
+reveal_type(a.copy()) # E: Revealed type is 'TypedDict('__main__.A', {'x': builtins.int})'
+reveal_type(a.has_key()) # E: Revealed type is 'builtins.bool'
```
ilevkivskyi: Argument is missing here.

Also we can probably infer `Literal[True/False]` here and for `'s' in SomeTypedDict` @Michael0x2a you might want to add this to your "wishlist" for literal types.

test-data/unit/check-typeddict.test
```diff
@@ -1481,3 +1481,85 @@ f1(**a, **c) # E: "f1" gets multiple values for keyword argument "x" \
...
+from mypy_extensions import TypedDict
+
+A = TypedDict('A', {'x': int, 'y': List[int]}, total=False)
+B = TypedDict('B', {'x': int})
+a: A
+b: B
+
+reveal_type(a.pop('x')) # E: Revealed type is 'builtins.int'
+reveal_type(a.pop('y', [])) # E: Revealed type is 'builtins.list[builtins.int]'
+reveal_type(a.pop('x', '')) # E: Revealed type is 'Union[builtins.int, builtins.str]'
```
ilevkivskyi: Maybe add a line here for non-Instance default value (maybe TupleType) just in case?

mypy/plugin.py
JukkaL: https://github.com/python/mypy/issues/6019

test-data/unit/check-typeddict.test
JukkaL: This wouldn't be safe due to structural subtyping and casts. Users may use a total TypedDict type when some keys are actually missing, and we don't want to break such code even though it's questionable.

PR #6015, opened 2018-12-05
Document that TypedDicts are compatible with Mapping[str, object]

JukkaL:
In particular, they are *not* compatible with any dictionary types.

### DIFF
--- a/docs/source/more_types.rst
+++ b/docs/source/more_types.rst
@@ -820,15 +820,15 @@ arbitrarily complex types. For example, you can define nested
 TypedDicts and containers with TypedDict items.
 Unlike most other types, mypy uses structural compatibility checking
 (or structural subtyping) with TypedDicts. A TypedDict object with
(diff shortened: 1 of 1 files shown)


docs/source/more_types.rst
```diff
@@ -824,6 +824,20 @@ extra items is compatible with a narrower TypedDict, assuming item
 types are compatible (*totality* also affects
 subtyping, as discussed below).
 
+A TypedDict object is not compatible with the regular ``Dict[...]``
+type, since ``Dict`` allows arbitrary keys to be added and removed,
+unlike TypedDict. However, any TypedDict object is compatible with
```
gvanrossum: I don't like the term "compatible". I know we use it all over the place, but I always have to look for clues elsewhere to tell me whether it means "subtype" or "supertype". Maybe add "(i.e., is a subtype of)" at the end of this line?

PR #6020, opened 2018-12-06
Refactor: move the default plugin into a separate module

JukkaL:


JukkaL:
The Appveyor failure is a flake.

PR #6025, opened 2018-12-07
Refactor mypy.checkmember more

JukkaL:
Split a big function into multiple smaller functions plus a few smaller
tweaks. This doesn't change any behavior.

Follow-up to #6022.


mypy/checkmember.py
```diff
@@ -96,155 +96,179 @@ def analyze_member_access(name: str,
...
+        fallback = item.type.metaclass_type or fallback
+    return _analyze_member_access(name, fallback, mx)
 
+
+def analyze_union_member_access(name: str, typ: UnionType, mx: MemberContext) -> Type:
+    mx.msg.disable_type_names += 1
+    results = [_analyze_member_access(name, subtype, mx)
+               for subtype in typ.relevant_items()]
+    mx.msg.disable_type_names -= 1
+    return UnionType.make_simplified_union(results)
```
ilevkivskyi: This one is probably to small to have its own function, but for the sake of consistency it is probably OK.

mypy/checkmember.py
```diff
@@ -96,155 +96,179 @@ def analyze_member_access(name: str,
     return _analyze_member_access(name, typ, mx, override_info)
 
 
+# The several functions that follow implement analyze_member_access for various
+# types and aren't documented individually.
```
ilevkivskyi: I would actually move this comment after `_analyze_member_acces` itself, e.g. to line 135.

PR #6034, opened 2018-12-08
Prohibit some illegal uses of Literal

Michael0x2a:
This pull request checks to make sure we report errors if the user tries using Literal types in invalid places. In particular, this PR...

1. Adds tests to make sure we cannot subclass Literals (mypy already directly supported this)
2. Checks to make sure we cannot use Literals inside `isinstance` and `issubclass` checks

I also wanted to add a check preventing people from attempting to instantiate a Literal (e.g. disallow `Literal[3]()` or `Literal()`), but that might require a little more work and a few changes to `typing_extensions`. (We currently don't raise an error when people try doing things like `Final()`, `Final[int]()` or `Protocol()` either).

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -248,6 +248,9 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                     except KeyError:
                         # Undefined names should already be reported in semantic analysis.
                         pass
+                if is_expr_literal_type(typ):
+                    self.msg.cannot_use_function_with_type(e.callee.name, "Literal", e)
+                    continue
                 if ((isinstance(typ, IndexExpr)
                         and isinstance(typ.analyzed, (TypeApplication, TypeAliasExpr)))
                         or (isinstance(typ, NameExpr) and node and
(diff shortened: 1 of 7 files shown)


ilevkivskyi:
> (We currently don't raise an error when people try doing things like `Final()`, `Final[int]()` or `Protocol()` either)

Hm, maybe this depend on version, but I see this:
```
>>> Protocol()
  File "<stdin>", line 1, in <module>
  File "/home/ivan/mypy_dev/lib/python3.6/site-packages/typing_extensions.py", line 1166, in __new__
    raise TypeError("Type Protocol cannot be instantiated; "
TypeError: Type Protocol cannot be instantiated; it can be used only as a base class
>>> Final()
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/typing.py", line 194, in __call__
    raise TypeError("Cannot instantiate %r" % type(self))
TypeError: Cannot instantiate typing_extensions.Final
>>> Final[int]()
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/typing.py", line 194, in __call__
    raise TypeError("Cannot instantiate %r" % type(self))
TypeError: Cannot instantiate typing_extensions.Final
```

Anyway, I think the fact that the error is emitted at runtime means it is not super-important to also flag this statically.

Michael0x2a:
> (We currently don't raise an error when people try doing things like Final(), Final[int]() or Protocol() either)

Oh, I meant we don't raise an error at type-check-time. (Well, we sort of do for `Final()`, `Protocol()`, and `Literal()` -- but the error message is just "`_SpecialForm not callable`", which I imagine would somewhat puzzle users who aren't aware of mypy/typing internals.)

> Anyway, I think the fact that the error is emitted at runtime means it is not super-important to also flag this statically.

Yeah -- I agree. I thought about making more of an effort to add support for this, but came to the same conclusion you did and just submitted this PR as-is.

PR #6035, opened 2018-12-09
Add tests for literals and generics

Michael0x2a:
This pull request adds some tests verifying that basic interactions between literals and generics work as expected.

It also tweaks `checkexpr` so it adds another exception for type variable return vs a Literal[...] context.


mypy/checkexpr.py
```diff
@@ -3346,6 +3346,10 @@ def narrow_type_from_binder(self, expr: Expression, known_type: Type) -> Type:
                 return ans
         return known_type
 
+    def context_contains_literal_like_type(self) -> bool:
+        """Returns 'true' if the context contains anything that resembles a LiteralType"""
+        return any(is_literal_type_like(item) for item in self.type_context)
```
ilevkivskyi: Why are you checking all items in the type context? I think only `type_context[-1]` should be checked.

Maybe add a test that checks that in this case literal type for `42` is _not_ inferred:
```python
def foo(x: T) -> T: ...
x: Union[List[int], Literal['bad']] = foo([42])
```
(i.e. this example  should type-check without errors).

test-data/unit/check-literal.test
```diff
@@ -385,6 +385,18 @@ b: bt                           # E: Invalid type "__main__.bt"
...
+[case testLiteralDisallowTypeVar]
+from typing import TypeVar
+from typing_extensions import Literal
+
+T = TypeVar('T')
+
+at = Literal[T]  # E: Parameter 1 of Literal[...] is invalid
+a: at
+
+def foo(b: Literal[T]) -> T: pass    # E: Parameter 1 of Literal[...] is invalid
```
ilevkivskyi: Too many spaces before `#`

mypy/checkexpr.py
Michael0x2a: I made that change to support this case:

    def expects_literal(x: Literal[3]) -> None: pass
    def foo(x: T) -> T: ...
    expects_literal(foo(3))  # Should not have an error

I found that without checking the entire context, we got an error on that final line since `type_context[-1]` is `T`, not `Literal[3]`.

Maybe I could modify `context_contains_literal_like_type` so that it starts by checking `type_context[-1]` and only tries moving backwards if `type_context[-1]` happened to be a TypeVar or something?

Or alternatively, modify this method so that it only tries inferring a Literal if the fallback matches -- so since 42 is an int and `Literal['bad']`'s fallback is str, we wouldn't error in your example.

JukkaL: But what if the return type of `foo` is something other than `T`? Then it might not be correct to infer a literal type, right? I suspect that accessing type context below the top item does not really make sense, since we don't know what's "between" the type context items.

ilevkivskyi: @Michael0x2a The outer context doesn't propagate inside in your example because there is special casing for functions that return a plain type variable, for them, only generic instance context is used for inference. If you really want your example to work, then you can add `or isinstance(ctx, LiteralType)` (the current check is anyway pure heuristic). As @JukkaL explained, checking all context stack doesn't really make sense.

Michael0x2a: @JukkaL and @ilevkivskyi -- ok, I updated the PR. I ended up using basically the fix Ivan suggested (and was pleasantly surprised to discover this seemed to fix all but one minor edge case).

Michael0x2a:
Windows test failure appears to be a flake.

PR #6041, opened 2018-12-09
Remove a bunch of unused imports

ilevkivskyi:
I also add one missing `Final` annotation I noticed while cleaning the imports.


ilevkivskyi:
flkae8 still doesn't work. PyCharm works very well, but there is no way to put in CI like Travis as far as I know.

PR #6042, opened 2018-12-09
All protocols and their explicit subclasses are ABCs

### LINKED ISSUE #6036: collections.abc.Sequence considered typing.Sequence
```python
from collections.abc import Sequence

class Spam: pass

Sequence.register(Spam)
```

Produces this error when run through mypy:

```
test.py:5: error: "Type[Sequence[Any]]" has no attribute "register"
```

This is a regression in v0.650. This error was not produced in v0.641.

ilevkivskyi:
Fixes #6036


mypy/semanal.py
```diff
@@ -1339,6 +1339,13 @@ def analyze_metaclass(self, defn: ClassDef) -> None:
             assert isinstance(inst, Instance)
             defn.info.declared_metaclass = inst
         defn.info.metaclass_type = defn.info.calculate_metaclass_type()
+        if any(info.is_protocol for info in defn.info.mro):
+            if (not defn.info.metaclass_type or
+                    defn.info.metaclass_type.type.fullname() == 'builtins.type'):
```
JukkaL: Maybe add a TODO about checking metaclass conflicts If there is another metaclass.

PR #6043, opened 2018-12-09
Implement meet for literal types; add tests for 'is_same_type'

Michael0x2a:
This pull request implements meets for literal types and adds some corresponding tests.

It also adds a test suite for the `is_same_type` method while we're at it. This test suite also lets you test each type's inherent `__eq__` and `__hash__`, since we do also use those throughout the code (especially for literal types).

### DIFF
--- a/mypy/test/testtypes.py
+++ b/mypy/test/testtypes.py
@@ -7,6 +7,7 @@ from mypy.erasetype import erase_type
 from mypy.expandtype import expand_type
 from mypy.join import join_types, join_simple
 from mypy.meet import meet_types
+from mypy.sametypes import is_same_type
 from mypy.types import (
     UnboundType, AnyType, CallableType, TupleType, TypeVarDef, Type, Instance, NoneTyp, Overloaded,
     TypeType, UnionType, UninhabitedType, true_only, false_only, TypeVarId, TypeOfAny, LiteralType
@@ -14,6 +15,7 @@ from mypy.types import (
 from mypy.nodes import ARG_POS, ARG_OPT, ARG_STAR, ARG_STAR2, CONTRAVARIANT, INVARIANT, COVARIANT
 from mypy.subtypes import is_subtype, is_more_precise, is_proper_subtype
 from mypy.test.typefixture import TypeFixture, InterfaceTypeFixture
+from mypy.experiments import strict_optional_set
 
 
 class TypesSuite(Suite):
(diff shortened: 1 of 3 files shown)


mypy/test/testtypes.py
```diff
@@ -846,8 +848,31 @@ def test_type_type(self) -> None:
...
+
+        self.assert_meet(lit1, lit1, lit1)
+        self.assert_meet(lit1, a, lit1)
+        self.assert_meet_uninhabited(lit1, lit3)
+        self.assert_meet_uninhabited(lit1, lit2)
+        self.assert_meet(UnionType([lit1, lit2]), lit1, lit1)
+        self.assert_meet(UnionType([lit1, lit2]), UnionType([lit2, lit3]), lit2)
+        self.assert_meet(UnionType([lit1, lit2]), UnionType([lit1, lit2]), UnionType([lit1, lit2]))
+        self.assert_meet(lit1, self.fx.anyt, lit1)
+        self.assert_meet(lit1, self.fx.o, lit1)
```
ilevkivskyi: Maybe test these two in reverse order too? (i.e. if literal is the second one)

test-data/unit/check-literal.test
```diff
@@ -1240,3 +1240,81 @@ indirect.Literal()
...
+def f3(x: Literal[1], y: int) -> None: pass
+def f4(x: Literal[1], y: object) -> None: pass
+def f5(x: Literal[1], y: Union[Literal[1], Literal[2]]) -> None: pass
+
+reveal_type(unify(f1))  # E: Revealed type is 'Literal[1]'
+reveal_type(unify(f2))  # E: Revealed type is 'None'
+reveal_type(unify(f3))  # E: Revealed type is 'Literal[1]'
+reveal_type(unify(f4))  # E: Revealed type is 'Literal[1]'
+reveal_type(unify(f5))  # E: Revealed type is 'Literal[1]'
+[out]
```
ilevkivskyi: TBH, having both these tests look a bit redundant (especially that you have very good unit tests). I would combine them into a single tests and check meeting some type one way, and some types another way.

mypy/test/testtypes.py
Michael0x2a: The implementation of `assert_meet` already does this for us, thankfully: it's defined as just:

    def assert_meet(self, s: Type, t: Type, meet: Type) -> None:
        self.assert_simple_meet(s, t, meet)
        self.assert_simple_meet(t, s, meet)

...where `assert_simple_meet` is responsible for actually doing the checking.

test-data/unit/check-literal.test
Michael0x2a: I consolidated both tests into one and pruned out some probably-overkill `reveal_type(...)` calls. I think testing the types in both ways looks tidier though, so kept that.

PR #6047, opened 2018-12-10
Explicitly use UTF-8 encoding in tests

JukkaL:
This will make things more consistent in different environments.


JukkaL:
My understanding is that modifying the default encoding is considered a bad idea. We could define a helper function or an encoding constant but it doesn't seem worth it.

PR #6054, opened 2018-12-11
Add support for -posix and -windows test name suffixes

JukkaL:
This allows writing platform-specific test cases, or
two platform-specific variants for a single test case.

A test case with -posix is only run on Linux and macOS, while
tests with a -windows suffix will only be run on Windows.

I want this for a test case I want to add where the output
depends on line ending style. There could be other uses
in the future.

### DIFF
--- a/mypy/test/data.py
+++ b/mypy/test/data.py
@@ -194,7 +194,7 @@ class DataDrivenTestCase(pytest.Item):  # type: ignore  # inheriting from Any
         self.writescache = writescache
         self.only_when = only_when
         if ((platform == 'windows' and sys.platform != 'win32')
-                or (platform == 'posix' and sys.platform not in ('linux', 'darwin'))):
+                or (platform == 'posix' and sys.platform == 'win32')):
             skip = True
         self.skip = skip
         self.data = data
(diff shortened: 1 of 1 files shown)


mypy/test/data.py
```diff
@@ -191,6 +193,9 @@ def __init__(self,
         self.file = file
         self.writescache = writescache
         self.only_when = only_when
+        if ((platform == 'windows' and sys.platform != 'win32')
+                or (platform == 'posix' and sys.platform not in ('linux', 'darwin'))):
```
gvanrossum: But what about BSD? I think the only reasonable test for POSIX is "not Windows".

mypy/test/data.py
```diff
@@ -583,16 +588,19 @@ def split_test_cases(parent: 'DataSuiteCollector', suite: 'DataSuite',
     cases = re.split(r'^\[case ([a-zA-Z_0-9]+)'
                      r'(-writescache)?'
                      r'(-only_when_cache|-only_when_nocache)?'
+                     r'(-posix|-windows)?'
                      r'(-skip)?'
                      r'\][ \t]*$\n', data,
```
gvanrossum: While you're here, can you move `data` to a line by itself? It's hiding here behind the end of the regex.

JukkaL:
AppVeyor failure looks like a flake.

PR #6064, opened 2018-12-13
Fix crash in daemon mode if there is a decode error

JukkaL:
I got a crash after I accidentally inserted a non-ascii
character to a file I was editing.

This also fixes a similar issue when the file can't be
read. I didn't add a test case for this since that
would have required bigger changes to the test support 
code.


test-data/unit/fine-grained-blockers.test
```diff
@@ -407,3 +407,49 @@ a.py:1: error: invalid syntax
...
 ==
+
+[case testDecodeErrorBlocker-posix]
+import a
+a.f(1)
+[file a.py]
+def f(x: int) -> None: ...
+[file a.py.2]
+# coding: ascii
+ä
```
gvanrossum: I'd make this an assignment (which would be legal with the Py3 default encoding).

test-data/unit/fine-grained-blockers.test
```diff
@@ -407,3 +407,49 @@ a.py:1: error: invalid syntax
...
+[file a.py.3]
+def f(x: str) -> None: ...
+[out]
+==
+mypy: can't decode file 'tmp/a.py': 'ascii' codec can't decode byte 0xc3 in position 16: ordinal not in range(128)
+==
+main:2: error: Argument 1 to "f" has incompatible type "int"; expected "str"
+
+[case testDecodeErrorBlocker-windows]
```
gvanrossum: What's the difference between the two tests? Why the platform-specific tags?

JukkaL: The offset in the error message is different due to cr/lf.

PR #6067, opened 2018-12-13
Remove the test of LiteralType with a float

msullivan:
One of the tests of LiteralType does some tests with a float literal,
to "make sure that the semantics are robust against this kind of
thing". Unfortunately, they are not. mypyc raises a TypeError
when the float is passed in.

### DIFF
--- a/mypy/test/testtypes.py
+++ b/mypy/test/testtypes.py
@@ -906,7 +906,6 @@ class SameTypeSuite(Suite):
         self.fx = TypeFixture()
 
     def test_literal_type(self) -> None:
-        a = self.fx.a
         b = self.fx.b  # Reminder: b is a subclass of a
         d = self.fx.d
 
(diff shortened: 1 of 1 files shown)


gvanrossum:
Um... So if I write `Literal[3.0]` and compile with mypyc, the program will crash? Will I at least get an error when type-checking that?

msullivan:
> Um... So if I write `Literal[3.0]` and compile with mypyc, the program will crash? Will I at least get an error when type-checking that?

It gets caught while constructing the type:
```
asdf.py:2: error: Parameter 1 of Literal[...] cannot be of type "float"
```

Michael0x2a:
@gvanrossum -- basically, what happened here was that I got a little over-zealous while constructing a test: I added a test that attempted manually constructing an illegal LiteralType (using some casting shenanigans) and confirmed it still behaved according to the overall Literal semantics.

This crash should never happen while actually running mypy, since the checks prevent an illegal LiteralType from ever being constructed in the first place.

PR #6069, opened 2018-12-13
Fix string parsing errors with literal types

Michael0x2a:
While testing literal types on our internal codebases, I discovered that the current parsing logic for literal types did not correct handle strings inside literal types. In particular, suppose we have the type `Literal["3"]`. Previously, the parser would attempt to parse `"3"` as a regular type and return a `RawLiteralType(3, "builtins.int")`, which caused the type to ultimately become `Literal[3]`.

This diff modifies the string-handled code to guarantee that a string expression will only ever resolve to an UnboundType or a `RawLiteralType(..., "builtins.str")`.


Michael0x2a:
Hmm, good point -- I refactored out the duplicate code (and will land once this goes green).

Issue #5037, opened 2018-05-14
Document `assert x.y is not None` in the cheat sheet


JukkaL:
Asserting when the inferred type is `Optional[X]` but we expect just `X` is a common thing and may not be obvious to users.

sp1rs:
This is already specified in the cheat sheet. This issue can be closed.
From cheat_sheet.rst
```
# Use Optional[] for values that could be None
x = some_function()  # type: Optional[str]
if x is not None:
   print x
```

ilevkivskyi:
This is different. I think the idea of this issue is to document that there are _two_ ways (which is important): `assert x.y is not None` and `if x.y is not None`.

PR #6074, opened 2018-12-15
Complement cheatsheet for Optionals

vincent-prz:
Fixes https://github.com/python/mypy/issues/5037

Add a paragrah to explain there are 2 ways to make the typechecker
understand that an optional is not `None`: use a if condition or an
`assert`.

### DIFF
--- a/docs/source/cheat_sheet.rst
+++ b/docs/source/cheat_sheet.rst
@@ -46,9 +46,13 @@ Built-in types
 
    # Use Optional[] for values that could be None
    x = some_function()  # type: Optional[str]
+   # To indicate to mypy that the value is not None, 2 cases:
+   # If it still may be None sometimes, use an if-statement
    if x is not None:
-      print x
-
+       print x.upper()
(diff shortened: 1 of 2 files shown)


docs/source/cheat_sheet.rst
```diff
@@ -49,6 +49,15 @@ Built-in types
    if x is not None:
       print x
 
+   # There are 2 ways you can make mypy understand that
+   # an Optional value is not None
+   x = some_function()  # type: Optional[str]
+   # First way: with an if condition
+   if x is not None:
+       print(x.upper())
+   # Second way: with an assert statement
```
ilevkivskyi: These are not just two ways. They are semantically different. The cheat sheet should say something like:
```
# If it still may be None sometimes, use an if-statement
...
# If it can never be None (e.g. due to some invariants), use an assert
...
```

vincent-prz:
@ilevkivskyi I have addressed your comment. Does this look good to you ? Thanks!

ilevkivskyi:
Thanks!

PR #6075, opened 2018-12-17
Add tests for Literal types with incremental and fine-grained mode

Michael0x2a:
This pull request adds a variety of tests for Literal types and incremental and fine-grained mode.

Many of these tests are a bit perfunctory: literal types don't really interact with incremental/fine-grained logic in a special way, so it was a little hard for me to think of good test cases.

The only code-related change I ended up needing to make was to remove the assert from the `mypy.server.astmerge.TypeReplaceVisitor`'s `visit_raw_literal_type` method.

This actually worries me a bit: we should have removed all instances of RawLiteralType after phase 2 of semantic analysis. Does the merge step happen before then? If the merge is supposed to happen after phase 2, I can spend more time trying to track down the leak and update this PR.


test-data/unit/fine-grained.test
```diff
@@ -8134,3 +8134,80 @@ x = 'no way'
...
+def foo(x: Literal[4]) -> Literal['foo']: ...
+@overload
+def foo(x: int) -> str: ...
+@overload
+def foo(x: Literal['bar']) -> int: ...
+def foo(x): pass
+[out]
+main:2: error: Revealed type is 'builtins.str'
+==
+main:2: error: Revealed type is 'Literal['foo']'
```
ilevkivskyi: These tests are indeed not very useful. What might be more interesting is to test the situations with "chained" assignments (where you have `y = x`, `z = y`, `w = z` etc, in different targets/modules) and where the type of the initial variable is changed, that causes a type error to appear with the last variable.

A similar idea is when you have a chain of function aliases where the last one provides a context for a function call. You can try playing with this.

Also maybe more interesting tests can be added when you will implement intelligent indexing and interaction with `Final`.

Michael0x2a:
Note: I will rebase on top of master once https://github.com/python/mypy/pull/6121 lands.

Michael0x2a:
As an update, I've added a few more tests: mostly some things testing and string/bytes/unicode stuff. I'll continue adding more tomorrow morning, or in a separate PR if you want.

If we decide to land this mostly as-is, I'll close https://github.com/python/mypy/pull/6121. Or, we could land some form of that PR and I can rebase on top of it.

test-data/unit/fine-grained.test
```diff
@@ -8293,3 +8293,226 @@ B = func
...
+[file mod3.py]
+from typing_extensions import Literal
+qux: Literal[3]
+[file mod3.py.2]
+from typing_extensions import Literal
+qux: Literal[4]
+[out]
+main:2: error: Revealed type is 'Literal[3]'
+==
+main:2: error: Revealed type is 'Literal[4]'
```
ilevkivskyi: As you know, errors can cause an extra pass for a target, so I would remove the reveal type errors on the first run in the tests where it is obvious.

PR #6076, opened 2018-12-17
Ignore cache when producing reports

mthuurne:
I had to add a bit of functionality to the test runner, so command line tests can do multiple runs. That is in the first commit. The second commit contains the actual fix.


mypy/test/testcmdline.py
```diff
@@ -32,10 +32,20 @@ class PythonCmdlineSuite(DataSuite):
...
     def run_case(self, testcase: DataDrivenTestCase) -> None:
-        test_python_cmdline(testcase)
-
-
-def test_python_cmdline(testcase: DataDrivenTestCase) -> None:
+        if testcase.output2:
+            for step in [1] + sorted(testcase.output2):
+                try:
+                    test_python_cmdline(testcase, step)
+                except AssertionError as ex:
```
msullivan: I'd prefer to tweak the message at creation time (like is done in `testcheck.py`), rather than depending on catching it and mucking around with the internals.

Issue #2536, opened 2016-12-07
Making it easier to use unicode_literals


JukkaL:
Currently using `from __future__ import unicode_literals` in Python 2 is kind of painful with mypy. Here are some potential issues:

* Some typeshed stubs declare things to only accept `str` arguments, when in reality they also accept at least ascii-only `unicode` objects.
* Mypy might not accept the implicit `unicode` literals in every context where a literal has a syntactic role (outside normal expressions), and `b'foo'` literals won't work in Python 3. String literal type escaping is an example.

It's fair to argue that using `unicode_literals` is a bad idea, but it's not uncommon to encounter real-world code using that, so getting mypy to work with it might be at least somewhat important.

ddfisher:
FWIW string literal type escaping has already been fixed: https://github.com/python/mypy/pull/2516

After that fix I did a quick grep of the codebase looking for other similar errors, and the only one I found is fixed in https://github.com/python/mypy/pull/2519.

JukkaL:
This is still rejected by mypy when using `--py2 --fast-parser`:

```
from __future__ import unicode_literals

def f(x: 'int') -> int:   # invalid syntax
    return x + 1
```

JukkaL:
This doesn't seem worth fixing any more, since Python 2 reached it's EoL and gradually the amount of code using `unicode_literals` is likely to decrease.

Issue #5098, opened 2018-05-23
Bad typing for Python 2.7 with `unicode_literals`


ikonst:
Consider:
```py
from __future__ import unicode_literals
import sys

print(sys.modules['sys'])
```

`mypy --py2 unicode_literals.py` results in:
```
unicode_literals.py:4: error: Invalid index type "unicode" for "Dict[str, Any]"; expected type "str"
```

However, `python2.7 unicode_literals.py` results in:
```
<module 'sys' (built-in)>
```

gvanrossum:
Did I mention that from __future__ import unicode_literals is a bad idea?

Michael0x2a:
I believe mypy currently doesn't support unicode_literals: there's some related discussion [here](https://github.com/python/mypy/issues/3619) and [here](https://github.com/python/mypy/issues/2536).

Maybe there's an opportunity to do some usability enhancement here? If we're not planning on supporting unicode_literals any time soon, maybe we can have mypy report a warning when it detects some code using it/link to something in the docs.

gvanrossum:
Oh, that is a great idea!

Issue #3619, opened 2017-06-27
Treat ascii-only literals as str, not unicode when using `unicode_literals`


JukkaL:
Many typeshed stubs only accept `str` when ascii-only `unicode` values also work at runtime. This causes a lot of problems with code using `unicode_literals`, since generally most literals will have `unicode` as their type. A quick fix would be to infer `str` as the type for ascii-only string literals when using `unicode_literals`, but only then. This would cause some unsafety, but it may be worth it for users using `unicode_literals`. The type safe solution is to manually check every `str` argument type in the entire typeshed and potentially replace it with `unicode` or similar, but this would be extremely painful and unlikely to happen any time soon. This would have no effect on code that doesn't use `unicode_literals`.

Example of how this would work:

```py
from __future__ import unicode_literals

reveal_type('foo')   # str
reveal_type('ß')   # unicode
```

Some open questions:
* Should `u'foo'` always have type `unicode`?
* What's the inferred type of `x` in `x = 'foo'`?

ilevkivskyi:
I am not sure about the two questions, but it looks like supporting this in ``typed_ast`` will be very simple. I can just add a flag (e.g. ``can_be_bytes``) to the ``Str`` AST node that will be _only_ set if ``unicode_literals`` future import is used and all chars are ASCII (very similar to how I did ``has_b`` flag). This flag can then be used in ``fastparse2`` whether to convert this node to ``StrExpr`` or ``UnicodeExpr``.

manu-chroma:
Hey our project [cwltool](https://github.com/common-workflow-language/cwltool) has been ``mypy`` since some time and it has been really helpful. 

We are trying to adpot use of ``unicode_literals`` in our codebase (https://github.com/common-workflow-language/cwltool/pull/442) but it's breaking ``mypy`` badly. I'm able to handle issues related stdlibs such as ``urllib`` and ``json`` API by slightly modifying stubs and adding into our ``typeshed`` folder.

But I'm not able to figure out how to deal with these issues.
```
cwltool/process.py:436: error: No overload variant of "get" of "Mapping" matches argument types [builtins.unicode, builtins.list[<nothing>]]
cwltool/workflow.py:277: error: No overload variant of "get" of "Mapping" matches argument types [builtins.unicode, builtins.unicode]
cwltool/workflow.py:280: error: No overload variant of "get" of "Mapping" matches argument types [builtins.unicode]
[...]

PR #6087, opened 2018-12-19
Add support for byte and unicode Literal strings

Michael0x2a:
This pull request adds support for byte and unicode Literal strings. I left in some comments explaining some nuances of the implementation; here are a few additional meta-notes:

1.  I reworded several of the comments suggesting that the way we represent bytes as a string is a hack or that we should eventually switch to representing bytes as literally bytes.

    Basically, I tried experimenting with that approach but ultimately rejected it: I ended up having to constantly serialize/deserialize between bytes and strings, which I felt complicated the code.

    As a result, I decided that the solution we had previously is in fact, from a high-level perspective, the best possible approach. (The actual code for translating the output of `typed_ast` into a human-readable string *is* admittedly a bit hacky though.)

    In any case, the phrase "how mypy currently parses the contents of bytes literals" is severely out-of-date anyways. That comment was added about 3 years ago, when we were adding the fast parser for the first time and running it concurrently with the actual parser.

2.  I removed the `is_stub` field from `fastparse2.ASTConverter`: it turned out we were just never using that field.

3.  One complication I ran into was figuring out how to handle forward references to literal strings. For example, suppose we have the type `List["Literal['foo']"]`. Do we treat this as being equivalent to `List[Literal[u'foo']]` or `List[Literal[b'foo']]`?

    If this is a Python 3 file or a Python 2 file with `unicode_literals`, we'd want to pick the former. If this is a standard Python 2 file, we'd want to pick the latter.

    In order to make this happen, I decided to use a heuristic where the type of the "outer" string decides the type of the "inner" string. For example:

    -   In Python 3, `"Literal['foo']"` is a unicode string. So, the inner `Literal['foo']` will be treated as the same as `Literal[u'foo']`.

    -   The same thing happens when using Python 2 with `unicode_literals`.

    -   In Python 3, it is illegal to use a byte string as a forward reference. So, types like `List[b"Literal['foo']"]` are already illegal.

    -   In standard Python 2, `"Literal['foo']"` is a byte string. So the inner `Literal['foo']` will be treated as the same as `Literal[u'foo']`.

4.  I'll add tests validating that all of this stuff works as expected with incremental and fine-grained mode in a separate diff -- probably after fixing and landing https://github.com/python/mypy/pull/6075, which I intend to use as a baseline foundation.

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -1231,10 +1231,12 @@ class StrExpr(Expression):
     """String literal"""
 
     value = ''
+    from_python_2 = False
 
-    def __init__(self, value: str) -> None:
+    def __init__(self, value: str, from_python_2: bool = False) -> None:
         super().__init__()
         self.value = value
+        self.from_python_2 = from_python_2
 
     def accept(self, visitor: ExpressionVisitor[T]) -> T:
         return visitor.visit_str_expr(self)
@@ -1243,7 +1245,16 @@ class StrExpr(Expression):
 class BytesExpr(Expression):
     """Bytes literal"""
 
-    value = ''  # TODO use bytes
+    # Note: we deliberately do NOT use bytes here because it ends up
+    # unnecessarily complicating a lot of the result logic. For example,
+    # we'd have to worry about converting the bytes into a format we can
+    # easily serialize/deserialize to and from JSON, would have to worry
+    # about turning the bytes into a human-readable representation in
+    # error messages...
+    #
+    # It's more convenient to just store the human-readable representation
+    # from the very start.
+    value = ''
 
     def __init__(self, value: str) -> None:
         super().__init__()
@@ -1256,7 +1267,7 @@ class BytesExpr(Expression):
 class UnicodeExpr(Expression):
     """Unicode literal (Python 2.x)"""
 
-    value = ''  # TODO use bytes
+    value = ''
 
     def __init__(self, value: str) -> None:
         super().__init__()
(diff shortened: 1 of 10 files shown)


mypy/nodes.py
```diff
@@ -1231,10 +1231,12 @@ class StrExpr(Expression):
     """String literal"""
 
     value = ''
+    from_python_2 = False
```
ilevkivskyi: Please document this.

mypy/types.py
```diff
@@ -1368,6 +1383,23 @@ def __eq__(self, other: object) -> bool:
         else:
             return NotImplemented
 
+    def value_repr(self) -> str:
```
ilevkivskyi: Please add a docstring.

Michael0x2a:
@ilevkivskyi -- alas, I don't think this PR helps resolve any of those issues. The only real non-literal-types related change this PR makes is modifying the way mypy handles parsing "string inside strings" -- when parsing types like `List["Literal['foo']"]`.

That said, I think we might be able to close https://github.com/python/mypy/issues/2536 though it seems like it's mostly a duplicate of https://github.com/python/mypy/issues/5098 at this point. (And also, the code snippet Jukka included at the end doesn't work because it's using function annotations inside of Python 2 code, not because of anything to do with strings or `unicode_literals`.)

The suggestion in https://github.com/python/mypy/issues/3619 may be eventually superseded by https://github.com/python/typing/issues/208, not sure.

PR #6095, opened 2018-12-20
 Update tests to prepare for variable redefinition with new type

JukkaL:
After the planned change, test cases no longer can rely on code
like this generating an error, since the two assignments will
create two logically separate variables with independent types:

```
x = 0
x = ''  # Currently an error, but likely not in the future
```

Instead, the idea is to use an if statement to prevent redefinition.
This works since redefinition will only happen unconditionally:

```
x = 0
if int():
    x = ''  # Will continue to be an error
```

Even this may stop working in the future, and we may need to
another major test case update then. If we always use the
`if int():` idiom, finding things that need changing will be
easy. (Sometimes `if int():` didn't work and I used `if 1:`
instead.)

In new test cases we should generally avoid using assignment
statements such as the above to generate errors. Instead, we
can use `reveal_type()` and calls to functions with incompatible
argument types, for example.

### DIFF
--- a/test-data/unit/check-dynamic-typing.test
+++ b/test-data/unit/check-dynamic-typing.test
@@ -7,9 +7,12 @@ from typing import Any
 d = None # type: Any
 a = None # type: A
 
-a = d # Everything ok
-d = a
-d = d
+if int():
+    a = d # Everything ok
+if int():
+    d = a
+if int():
+    d = d
 d.x = a
 d.x = d
 
@@ -20,12 +23,17 @@ from typing import Any
 d = None # type: Any
 a, b = None, None # type: (A, B)
 
-d, a = b, b    # E: Incompatible types in assignment (expression has type "B", variable has type "A")
-d, d = d, d, d # E: Too many values to unpack (2 expected, 3 provided)
+if int():
+    d, a = b, b    # E: Incompatible types in assignment (expression has type "B", variable has type "A")
+if int():
+    d, d = d, d, d # E: Too many values to unpack (2 expected, 3 provided)
 
-a, b = d, d
-d, d = a, b
-a, b = d
+if int():
+    a, b = d, d
+if int():
+    d, d = a, b
+if int():
+    a, b = d
 s, t = d
 
 class A: pass
@@ -40,12 +48,17 @@ class B: pass
 from typing import Any
 a, b = None, None # type: (A, B)
 
-b = f(a) # E: Incompatible types in assignment (expression has type "A", variable has type "B")
+if int():
+    b = f(a) # E: Incompatible types in assignment (expression has type "A", variable has type "B")
 
-a = f(a)
-a = f(b)
-a = f(None)
-a = f(f)
+if int():
+    a = f(a)
+if int():
+    a = f(b)
+if int():
+    a = f(None)
+if int():
+    a = f(f)
 
 def f(x: Any) -> 'A':
     pass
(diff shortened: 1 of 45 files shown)


test-data/unit/check-dynamic-typing.test
```diff
@@ -201,10 +252,12 @@ from typing import Any
...
 a = None # type: A
 
-a = d[a()] # E: "A" not callable
+if int():
+    a = d[a()] # E: "A" not callable
 d[a()] = a # E: "A" not callable
 
-a = d[a]
+if int():
+    a = d[a]
```
ilevkivskyi: I am also not sure why it is needed here. There is no error on this line. Will this create a new variable with type `Any` that shows the one with type `A`?

test-data/unit/check-enum.test
```diff
@@ -74,11 +77,11 @@ class N(IntEnum):
     x = 1
     y = 1
 n = 1
-n = N.x  # Subclass of int, so it's okay
+if int():
+    n = N.x  # Subclass of int, so it's okay
```
ilevkivskyi: This one is also not clear. I assume we want to test that there is no error even with the new semantics.

test-data/unit/check-expressions.test
```diff
@@ -33,26 +33,34 @@ class B(A): pass
...
+    a = """foo"""
 class A:
     pass
 
 [case testFloatLiteral]
 a = 0.0
 b = None # type: A
-b = 1.1 # E: Incompatible types in assignment (expression has type "float", variable has type "A")
-a = 1.1
+if str():
```
ilevkivskyi: What is specific about `str()` here so it is used instead of `int()`?

test-data/unit/check-final.test
```diff
@@ -339,22 +339,29 @@ class C:
 from typing import Final
 
 x: Final = 1
+x
```
ilevkivskyi: I strongly prefer to not shadow final names, since this kind of defeats the purpose.

test-data/unit/check-isinstance.test
```diff
@@ -300,6 +307,7 @@ class B(A):
     b = 1
 
 x = A()
+def f(): x  # Prevent redefinition
```
ilevkivskyi: I would create a full list of "hack" patterns used, and place it in one of the test files as a comment.

test-data/unit/check-newtype.test
```diff
@@ -310,17 +310,15 @@ main:4: error: Invalid type "__main__.T"
 from typing import NewType
 
 a = 3
-a = NewType('a', int)
+def f(): a
```
ilevkivskyi: Allowing redefinition with a different kind of symbol looks dangerous from the point of view of fine grained incremental, but maybe it is OK.

test-data/unit/check-final.test
JukkaL: My implementation does some additional checks in a later pass after renaming to prevent the redefinition of final names. It should preserve the existing semantics of final names. Or is there something else you are worried about?

test-data/unit/check-isinstance.test
JukkaL: I'll add a separate file where I describe them, as these aren't specific to any particular test file.

JukkaL:
> First, I would gather all the patterns used

A good idea! I'll add some documentation about these.

> Second, I am worried about shadowing final names.

I think that final semantics should be effectively the same. I don't remember why I modified the test cases -- it may have been due to changes in the details of error messages. I'll double check this before merging.

> Third, interaction with Any is a bit unexpected, it can create both "false" positives, and "false" negatives (depending on the order), again probably we can discuss this separately.

This can be discussed once we have the proposed implementation for the feature. I'd argue that the new behavior is usually an improvement -- they caught many false negatives in Dropbox codebases, and they also fix false positives. There is no _explicit_ interaction with `Any`, since the implementation performs a renaming transformation without any knowledge of types.

> Finally, there is a use case that is problematic with current strategy (allow redefinition only for non-overlapping life-times)

Yes, this could generate new errors. I didn't see many examples of this in the Dropbox S codebase though, so it's probably rare.

test-data/unit/check-final.test
JukkaL: The `x` is reference is used to trigger the renaming machinery. Otherwise the second definition would be trivially rejected, as it would not count as a new definition. My implementation only renames a variable if it has been read, since this shouldn't be treated as a redefinition:

```py
x = None  # type: List[int]
x = []  # Not a redefinition
```

PR #6096, opened 2018-12-20
Don't map actual **kwargs to formal *args

### LINKED ISSUE #1969: **kwargs checked against wrong type with defaults or *args

JukkaL:
The fix was originally proposed by @elazarg.

Fixes #1969.

### DIFF
--- a/test-data/unit/check-kwargs.test
+++ b/test-data/unit/check-kwargs.test
@@ -403,3 +403,16 @@ class A:
 tmp/m.py:1: error: Unsupported operand types for + ("int" and "str")
 main:2: error: Unexpected keyword argument "x" for "A"
 tmp/m.py:3: note: "A" defined here
+
+[case testStarArgsAndKwArgsSpecialCase]
+from typing import Dict, Mapping
+
+def f(*vargs: int, **kwargs: object) -> None:
+	pass
+
+d = {} # type: Dict[str, object]
+f(**d)
+
+m = {} # type: Mapping[str, object]
+f(**m)
+[builtins fixtures/dict.pyi]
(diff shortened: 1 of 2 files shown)


test-data/unit/check-kwargs.test
```diff
@@ -403,3 +403,16 @@ class A:
...
+
+def f(*vargs: int, **kwargs: object) -> None:
+	pass
+
+d = {} # type: Dict[str, object]
+f(**d)
+
+m = {} # type: Mapping[str, object]
+f(**m)
+[builtins fixtures/dict.pyi]
```
ilevkivskyi: Maybe also add the second test case mentioned in the issue? The one with:
```python
def b(arg: str='', **kw: int) -> None:
    pass
```

sobolevn: Three years later we have a similar error: https://github.com/python/mypy/issues/11583

But, calling `b` (in this case) with `b(**{'x': 1})` will work in runtime. It will result in `arg = ''` and `kw = {'x': 1}`.
Not sure why this was explictly forbidden. 

Do you remember why this was a problem?

sobolevn: My PR with the alternative solution: https://github.com/python/mypy/pull/11589

Issue #2055, opened 2016-08-25
Better error message when __eq__ has unexpected signature


JukkaL:
It's pretty common to have `__eq__` with a non-`object` argument type, and mypy will just complain that it's not compatible with `object` (as `object` defines `__eq__` with an `object` argument). It would be better to display a message that shows how to properly implement `__eq__`. Here's an idea (probably needs polish):

```
x.py:23: error: Argument 1 of "__eq__" incompatible with supertype "object"
x.py:23: note: It's recommended for "__eq__" to work with arbitrary objects.
x.py:23: note: This example shows how you can implement "__eq__":
x.py:23: note:
x.py:23: note: class Foo(...):
x.py:23: note:     ...
x.py:23: note:     def __eq__(self, other: object) -> bool:
x.py:23: note:         if not isinstance(other, Foo):
x.py:23: note:             return NotImplemented
x.py:23: note:        return <logic to compare equality of two Foo instances>
```

PR #6106, opened 2018-12-26
Better error message when __eq__ has unexpected signature

David-Wobrock:
Hey there,

This PR tackles an old issue #2055 opened in 2016 - to get started in the mypy source code.

This PR will add a message and an example when the arguments of the dunder methods  `__eq__` or `__ne__` do not match the expected signature. With this PR, the added message looks like this:
```
It is recommended for "{method_name}" to work with arbitrary objects.
The snippet below shows an example of how you can implement "{method_name}":

class Foo(...):
    ...
    def {method_name}(self, other: object) -> bool:
        if not isinstance(other, Foo):
            raise NotImplementedError
        return <logic to compare two Foo instances>
```

The 3 commits can and should be reviewed individually since they represent independent bulks of work:
1. Fix a docstring of errors.render_messages
2. Add an option so that reporting an error/warning/note eventually doesn't strip the message. This was necessary because the new message needs to preserve indentation.
3. The core of the work: add the message when one of these methods is used and add a test

To me, the details that can open to a discussion:
* the message itself. Can it be simpler/better?
* is the `note_multiline`-method approach acceptable for having multiple messages at once
* is the `strip_msg` boolean acceptable?
* did I understand the test framework correctly and is the test sufficient?

Looking forward for some reviews
Cheers,
David

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -182,28 +182,31 @@ class MessageBuilder:
 
     def report(self, msg: str, context: Optional[Context], severity: str,
                file: Optional[str] = None, origin: Optional[Context] = None,
-               offset: int = 0) -> None:
+               offset: int = 0, strip_msg: bool = True) -> None:
         """Report an error or note (unless disabled)."""
         if self.disable_count <= 0:
+            msg = msg.strip() if strip_msg else msg
             self.errors.report(context.get_line() if context else -1,
                                context.get_column() if context else -1,
-                               msg.strip(), severity=severity, file=file, offset=offset,
+                               msg, severity=severity, file=file, offset=offset,
                                origin_line=origin.get_line() if origin else None)
 
     def fail(self, msg: str, context: Optional[Context], file: Optional[str] = None,
-             origin: Optional[Context] = None) -> None:
+             origin: Optional[Context] = None, strip_msg: bool = True) -> None:
         """Report an error message (unless disabled)."""
-        self.report(msg, context, 'error', file=file, origin=origin)
+        self.report(msg, context, 'error', file=file, origin=origin, strip_msg=strip_msg)
 
     def note(self, msg: str, context: Context, file: Optional[str] = None,
-             origin: Optional[Context] = None, offset: int = 0) -> None:
+             origin: Optional[Context] = None, offset: int = 0,
+             strip_msg: bool = True) -> None:
         """Report a note (unless disabled)."""
-        self.report(msg, context, 'note', file=file, origin=origin, offset=offset)
+        self.report(msg, context, 'note', file=file, origin=origin,
+                    offset=offset, strip_msg=strip_msg)
 
     def warn(self, msg: str, context: Context, file: Optional[str] = None,
-             origin: Optional[Context] = None) -> None:
+             origin: Optional[Context] = None, strip_msg: bool = True) -> None:
         """Report a warning message (unless disabled)."""
-        self.report(msg, context, 'warning', file=file, origin=origin)
+        self.report(msg, context, 'warning', file=file, origin=origin, strip_msg=strip_msg)
 
     def quote_type_string(self, type_string: str) -> str:
         """Quotes a type representation for use in messages."""
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -203,6 +203,14 @@ class MessageBuilder:
         self.report(msg, context, 'note', file=file, origin=origin,
                     offset=offset, strip_msg=strip_msg)
 
+    def note_multiline(self, messages: str, context: Context, file: Optional[str] = None,
+             origin: Optional[Context] = None, offset: int = 0,
+             strip_msg: bool = True) -> None:
+        """Report as many notes as lines in the message (unless disabled)."""
+        for msg in messages.splitlines():
+            self.report(msg, context, 'note', file=file, origin=origin,
+                        offset=offset, strip_msg=strip_msg)
+
     def warn(self, msg: str, context: Context, file: Optional[str] = None,
              origin: Optional[Context] = None, strip_msg: bool = True) -> None:
         """Report a warning message (unless disabled)."""
@@ -862,6 +870,21 @@ class MessageBuilder:
         target = self.override_target(name, name_in_supertype, supertype)
         self.fail('Argument {} of "{}" incompatible with {}'
                   .format(arg_num, name, target), context)
+
+        if name in ("__eq__", "__ne__"):
+            multiline_msg = self.comparison_method_example_msg(name)
+            self.note_multiline(multiline_msg, context, strip_msg=False)
+
+    def comparison_method_example_msg(self, method_name: str) -> str:
+        return '''It is recommended for "{method_name}" to work with arbitrary objects.
+The snippet below shows an example of how you can implement "{method_name}":
+
+class Foo(...):
+    ...
+    def {method_name}(self, other: object) -> bool:
+        if not isinstance(other, Foo):
+            raise NotImplementedError
+        return <logic to compare two Foo instances>'''.format(method_name=method_name)
 
     def return_type_incompatible_with_supertype(
             self, name: str, name_in_supertype: str, supertype: str,
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -282,6 +282,33 @@ class B(A):
 main:7: error: Argument 1 of "f" incompatible with supertype "A"
 main:9: error: Return type of "h" incompatible with supertype "A"
 
+[case testEqNeMethodsOverridingWithNonObjects]
+class A:
+  def __eq__(self, other: A) -> bool: pass  # Fail
+  def __ne__(self, other: A) -> bool: pass  # Fail
+[builtins fixtures/attr.pyi]
+[out]
+main:2: error: Argument 1 of "__eq__" incompatible with supertype "object"
+main:2: note: It is recommended for "__eq__" to work with arbitrary objects.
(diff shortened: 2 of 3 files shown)


mypy/messages.py
```diff
@@ -182,28 +182,39 @@ def is_errors(self) -> bool:
...
+            msg = msg.strip() if strip_msg else msg
             self.errors.report(context.get_line() if context else -1,
                                context.get_column() if context else -1,
-                               msg.strip(), severity=severity, file=file, offset=offset,
+                               msg, severity=severity, file=file, offset=offset,
                                origin_line=origin.get_line() if origin else None)
 
     def fail(self, msg: str, context: Optional[Context], file: Optional[str] = None,
-             origin: Optional[Context] = None) -> None:
+             origin: Optional[Context] = None, strip_msg: bool = True) -> None:
```
ilevkivskyi: I would add the new `strip_msg` argument only where it is actually used, i.e. in notes. It is unlikely that we will have multiline errors or warnings.

mypy/messages.py
```diff
@@ -182,28 +182,39 @@ def is_errors(self) -> bool:
...
+             origin: Optional[Context] = None, offset: int = 0,
+             strip_msg: bool = True) -> None:
+        """Report as many notes as lines in the message (unless disabled)."""
+        for msg in messages.splitlines():
+            self.report(msg, context, 'note', file=file, origin=origin,
+                        offset=offset, strip_msg=strip_msg)
 
     def warn(self, msg: str, context: Context, file: Optional[str] = None,
-             origin: Optional[Context] = None) -> None:
+             origin: Optional[Context] = None, strip_msg: bool = True) -> None:
```
ilevkivskyi: Same here.

mypy/messages.py
```diff
@@ -182,28 +182,39 @@ def is_errors(self) -> bool:
...
+             origin: Optional[Context] = None, offset: int = 0,
+             strip_msg: bool = True) -> None:
         """Report a note (unless disabled)."""
-        self.report(msg, context, 'note', file=file, origin=origin, offset=offset)
+        self.report(msg, context, 'note', file=file, origin=origin,
+                    offset=offset, strip_msg=strip_msg)
+
+    def note_multiline(self, messages: str, context: Context, file: Optional[str] = None,
+             origin: Optional[Context] = None, offset: int = 0,
+             strip_msg: bool = True) -> None:
```
ilevkivskyi: I think `strip_msg` is not needed for `note_multiline`. It is quite clear that multiline notes will be often indented.

mypy/messages.py
```diff
@@ -182,28 +182,39 @@ def is_errors(self) -> bool:
...
     def report(self, msg: str, context: Optional[Context], severity: str,
                file: Optional[str] = None, origin: Optional[Context] = None,
-               offset: int = 0) -> None:
+               offset: int = 0, strip_msg: bool = True) -> None:
         """Report an error or note (unless disabled)."""
         if self.disable_count <= 0:
+            msg = msg.strip() if strip_msg else msg
             self.errors.report(context.get_line() if context else -1,
                                context.get_column() if context else -1,
-                               msg.strip(), severity=severity, file=file, offset=offset,
```
ilevkivskyi: After some thinking I am not sure why do we actually need to strip the message here. How many tests fail if you remove the `.strip()` here? Maybe we can instead update our test framework?

mypy/messages.py
```diff
@@ -860,6 +871,21 @@ def argument_incompatible_with_supertype(
         self.fail('Argument {} of "{}" incompatible with {}'
                   .format(arg_num, name, target), context)
 
+        if name in ("__eq__", "__ne__"):
```
ilevkivskyi: I think this note should be applied only for `__eq__`.

mypy/messages.py
```diff
@@ -860,6 +871,21 @@ def argument_incompatible_with_supertype(
...
+            multiline_msg = self.comparison_method_example_msg(name)
+            self.note_multiline(multiline_msg, context, strip_msg=False)
+
+    def comparison_method_example_msg(self, method_name: str) -> str:
+        return '''It is recommended for "{method_name}" to work with arbitrary objects.
```
ilevkivskyi: I would use `textwrap.dedent()` here instead of breaking the indentation.

mypy/messages.py
```diff
@@ -860,6 +871,21 @@ def argument_incompatible_with_supertype(
...
+The snippet below shows an example of how you can implement "{method_name}":
+
+class Foo(...):
```
ilevkivskyi: Can we use the actual class name instead here?

mypy/messages.py
```diff
@@ -860,6 +871,21 @@ def argument_incompatible_with_supertype(
...
+    ...
+    def {method_name}(self, other: object) -> bool:
+        if not isinstance(other, Foo):
```
ilevkivskyi: ...and here

test-data/unit/check-classes.test
```diff
@@ -282,6 +282,33 @@ class B(A):
...
+[case testEqNeMethodsOverridingWithNonObjects]
+class A:
+  def __eq__(self, other: A) -> bool: pass  # Fail
+  def __ne__(self, other: A) -> bool: pass  # Fail
+[builtins fixtures/attr.pyi]
+[out]
+main:2: error: Argument 1 of "__eq__" incompatible with supertype "object"
+main:2: note: It is recommended for "__eq__" to work with arbitrary objects.
+main:2: note: The snippet below shows an example of how you can implement "__eq__":
+main:2: note:
```
ilevkivskyi: This "preamble" is too wordy. I would use shorter `It is recommended for "__eq__" to work with arbitrary objects, for example:` and skip the empty line.

test-data/unit/check-classes.test
```diff
@@ -282,6 +282,33 @@ class B(A):
...
+main:2: note: class Foo(...):
+main:2: note:     ...
```
ilevkivskyi: I would actually skip these two lines (however I would keep the indentation of the next lines).

mypy/messages.py
```diff
@@ -860,6 +871,21 @@ def argument_incompatible_with_supertype(
...
+            raise NotImplementedError
+        return <logic to compare two Foo instances>'''.format(method_name=method_name)
```
ilevkivskyi: And here I would also use the actual class name instead of `Foo`.

ilevkivskyi:
(I also edited the PR description to add the "magic" text for GitHub.)

David-Wobrock:
Hey @ilevkivskyi, thanks for the comments.

The PR is open for review again and I took your changes into account in the last 2 commits:
1. Completely removing the `strip()` call on the message reporting works like charm. No test broke. So my best guess is to go without it.
2. Changing the displayed messaged

mypy/messages.py
```diff
@@ -860,6 +869,20 @@ def argument_incompatible_with_supertype(
         self.fail('Argument {} of "{}" incompatible with {}'
                   .format(arg_num, name, target), context)
 
+        if name == "__eq__":
+            assert isinstance(context, FuncDef)
+            multiline_msg = self.comparison_method_example_msg(class_name=context.info.name())
```
ilevkivskyi: Getting anything except line/column from `context` is a bad idea. First, this kind of defeats the purpose of the type annotation. Second, what if this is an `OverloadedFuncDef`, or a `Decorator`?

I think it is better to add an optional argument `current_type_name` to `argument_incompatible_with_supertype`, and `if name == '__eq__': assert current_type_name is not None` etc. (Or maybe even make it a required argument if there are only couple of call sites to update).

test-data/unit/check-classes.test
```diff
@@ -282,6 +282,18 @@ class B(A):
...
+[case testEqMethodsOverridingWithNonObjects]
+class A:
+  def __eq__(self, other: A) -> bool: pass  # Fail
+[builtins fixtures/attr.pyi]
+[out]
+main:2: error: Argument 1 of "__eq__" incompatible with supertype "object"
+main:2: note: It is recommended for "__eq__" to work with arbitrary objects, for example:
+main:2: note:     def __eq__(self, other: object) -> bool:
+main:2: note:         if not isinstance(other, A):
+main:2: note:             raise NotImplementedError
```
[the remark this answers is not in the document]
JelleZijlstra: Yes, `NotImplemented` is right here. It signals to the runtime to try the other object's `__eq__`.

David-Wobrock:
Hi @ilevkivskyi, thanks again for the feedback, and thank you @JelleZijlstra.

Again, the last 3 commits contain the suggested changes.
The messages method `argument_incompatible_with_supertype` is only called in one place, so that is easy to update. But to get the class/type name, I fall back on a hard-coded  `Foo` string when I don't have access to the function definition.
I don't know the codebase enough to know in what cases this can happen - and if it makes sense in `mypy` to fallback on such a class name for the note. Do you have a cleaner suggestion for this?

Cheers and happy new year!
David

mypy/checker.py
```diff
@@ -1476,6 +1476,10 @@ def check_override(self, override: FunctionLike, original: FunctionLike,
                 # are erased, then it is definitely an incompatibility.
 
                 override_ids = override.type_var_ids()
+                if isinstance(override.definition, FuncDef):
+                    type_name = override.definition.info.name()
+                else:
+                    type_name = "Foo"
```
ilevkivskyi: I think a better alternative is to not show the note at all if `__eq__` is not a function definition. So you can just set `type_name` to `None` here, and add a check in `argument_incompatible_with_supertype()` that will not show the note if the name is `None`.

PR #6107, opened 2018-12-27
[attrs plugin] Support kw_only=True

### LINKED ISSUE #6083: mypy attrs plugin doesn't work with new attrs' `kw_only` argument
Copied from https://github.com/python-attrs/attrs/issues/473 - issue concerns both parties...

```
$ python -m mypy --version
mypy 0.650
>>> attr.__version__
'18.2.0'
```

Use mypy on following file:
```
import typing
import attr


@attr.s(frozen=True, auto_attribs=True, kw_only=True)
class Position:
    line: int
    column: int


@attr.s(frozen=True, auto_attribs=True, kw_only=True)
class Node:
    start: typing.Optional[Position] = None
    end: typing.Optional[Position] = None


@attr.s(frozen=True, auto_attribs=True, kw_only=True)
class Operator(Node):
    symbol: str
    name: str
    precedence: int
    associativity: str  # "left" or "right"
```
It gives following errors:
```
f.py:21: error: Non-default attributes not allowed after default attributes.
f.py:22: error: Non-default attributes not allowed after default attributes.
f.py:23: error: Non-default attributes not allowed after default attributes.
f.py:24: error: Non-default attributes not allowed after default attributes.
```
If I understand correctly, mypy doesn't consider the existence of a new `kw_only=True` .

euresti:
Fixes #6083

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -9,7 +9,7 @@ from mypy.nodes import (
     Context, Argument, Var, ARG_OPT, ARG_POS, TypeInfo, AssignmentStmt,
     TupleExpr, ListExpr, NameExpr, CallExpr, RefExpr, FuncBase,
     is_class_var, TempNode, Decorator, MemberExpr, Expression,
-    SymbolTableNode, MDEF, JsonDict, OverloadedFuncDef
+    SymbolTableNode, MDEF, JsonDict, OverloadedFuncDef, ARG_NAMED_OPT, ARG_NAMED
 )
 from mypy.plugins.common import (
     _get_argument, _get_bool_argument, _get_decorator_bool_argument, add_method
@@ -56,12 +56,13 @@ class Attribute:
     """The value of an attr.ib() call."""
 
     def __init__(self, name: str, info: TypeInfo,
-                 has_default: bool, init: bool, converter: Converter,
+                 has_default: bool, init: bool, kw_only: bool, converter: Converter,
                  context: Context) -> None:
         self.name = name
         self.info = info
         self.has_default = has_default
         self.init = init
+        self.kw_only = kw_only
         self.converter = converter
         self.context = context
 
@@ -132,10 +133,15 @@ class Attribute:
             # Convert type not set to Any.
             init_type = AnyType(TypeOfAny.unannotated)
 
+        if self.kw_only:
+            arg_kind = ARG_NAMED_OPT if self.has_default else ARG_NAMED
+        else:
+            arg_kind = ARG_OPT if self.has_default else ARG_POS
+
         # Attrs removes leading underscores when creating the __init__ arguments.
         return Argument(Var(self.name.lstrip("_"), init_type), init_type,
                         None,
-                        ARG_OPT if self.has_default else ARG_POS)
+                        arg_kind)
 
     def serialize(self) -> JsonDict:
         """Serialize this object so it can be saved and restored."""
@@ -143,6 +149,7 @@ class Attribute:
             'name': self.name,
             'has_default': self.has_default,
             'init': self.init,
+            'kw_only': self.kw_only,
             'converter_name': self.converter.name,
             'converter_is_attr_converters_optional': self.converter.is_attr_converters_optional,
             'context_line': self.context.line,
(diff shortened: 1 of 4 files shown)


test-data/unit/lib-stub/attr.pyi
```diff
@@ -87,8 +95,12 @@ def attrs(maybe_cls: None = ...,
           init: bool = ...,
           slots: bool = ...,
           frozen: bool = ...,
+          weakref_slot: bool = ...,
           str: bool = ...,
-          auto_attribs: bool = ...) -> Callable[[_C], _C]: ...
+          auto_attribs: bool = ...,
+          kw_only: bool = ...,
+          cache_hash: bool = ...,
```
ilevkivskyi: I noticed that some of these are missing in typeshed. Could you please make a typeshed update PR?

test-data/unit/fine-grained.test
```diff
@@ -1001,6 +1001,37 @@ class A:
...
+
+[file b.py.2]
+from a import A
+import attr
+@attr.s()
+class B(A):
+    b = attr.ib("16")   # type: str
+
+B(b="foo", a=7)
+
```
ilevkivskyi: This empty line is not needed.

test-data/unit/fine-grained.test
```diff
@@ -1001,6 +1001,37 @@ class A:
...
+[builtins fixtures/attr.pyi]
+[out]
+==
+b.py:5: error: Non keyword-only attributes are not allowed after a keyword-only attribute.
+
+
```
ilevkivskyi: We prefer single empty line between tests.

test-data/unit/fine-grained.test
```diff
@@ -1001,6 +1001,37 @@ class A:
...
+@attr.s()
+class B(A):
+    b = attr.ib("16")   # type: str
+
+B(b="foo", a=7)
+
+[builtins fixtures/attr.pyi]
+[out]
+==
+b.py:5: error: Non keyword-only attributes are not allowed after a keyword-only attribute.
```
ilevkivskyi: This test doesn't test much of incremental logic. I would also add a test where base class in `a.py` is updated, while subclass is unchanged.

mypy/plugins/attrs.py
```diff
@@ -367,6 +398,11 @@ def _attribute_from_attrib_maker(ctx: 'mypy.plugin.ClassDefContext',
 
     # Read all the arguments from the call.
     init = _get_bool_argument(ctx, rvalue, 'init', True)
+    kw_only |= _get_bool_argument(ctx, rvalue, 'kw_only', False)
+    if kw_only and ctx.api.options.python_version[0] < 3:
+        ctx.api.fail("kw_only is not supported in Python 2", stmt)
```
ilevkivskyi: I would avoid repeating the same error message, and instead define a constant.

mypy/plugins/attrs.py
```diff
@@ -367,6 +398,11 @@ def _attribute_from_attrib_maker(ctx: 'mypy.plugin.ClassDefContext',
 
     # Read all the arguments from the call.
     init = _get_bool_argument(ctx, rvalue, 'init', True)
+    kw_only |= _get_bool_argument(ctx, rvalue, 'kw_only', False)
```
ilevkivskyi: I noticed this (a bit counterintuitive) behavior:
```python
import attr
@attr.s(kw_only=True)
class A:
    a = attr.ib(kw_only=False)

A(1)  # Error
```
The logic here captures it correctly, but I would add a short comment that this is intentional.

mypy/plugins/attrs.py
```diff
@@ -219,13 +231,15 @@ def _get_frozen(ctx: 'mypy.plugin.ClassDefContext') -> bool:
     return False
 
 
-def _analyze_class(ctx: 'mypy.plugin.ClassDefContext', auto_attribs: bool) -> List[Attribute]:
+def _analyze_class(ctx: 'mypy.plugin.ClassDefContext',
+                   auto_attribs: bool,
+                   kw_only: bool) -> List[Attribute]:
```
ilevkivskyi: Could you please document `kw_only` (and `auto_attribs`) in the docstring?

mypy/plugins/attrs.py
euresti: I filed an issue in attrs to see if this is expected: https://github.com/python-attrs/attrs/issues/481

euresti:
Ok I made the requested changes.

Issue #5272, opened 2018-06-25
"Tagged" types


ikonst:
I've been taking a stab at typing [pynamodb](http://pynamodb.readthedocs.io/en/latest/), and haven't been able to type [MapAttribute](http://pynamodb.readthedocs.io/en/latest/attributes.html#map-attributes).

Consider the following code:
```py
from pynamodb.attributes import MapAttribute, NumberAttribute
from pynamodb.models import Model

class MyMapAttribute(MapAttribute):
   foo = NumberAttribute()
   bar = NumberAttribute()

class MyModel(Model):
   my_map = MyMapAttribute()
```

If we try to inspect this, we see:
```py
>>> MyModel.my_map
<__main__.MyMapAttribute object at 0x10adfcb38>
>>> my_model = MyModel()
>>> my_model.my_map = MyMapAttribute(foo=1, bar=2)
>>> my_model.my_map
<__main__.MyMapAttribute object at 0x10adfcd30>
```

In our code, the MapAttribute-derived type has two instances - the first serves as a "descriptor" and the second serves as the instance. Both instances contain "foo" and "bar" members, but for different purpose.

```py
>>> type(my_model.my_map) is type(MyModel.my_map)
True
>>> my_model.my_map is MyModel.my_map
False
>>> my_model.my_map.foo
1
>>> MyModel.my_map.foo
<pynamodb.attributes.NumberAttribute object at 0x10adfccc0>
```

I understand that this is somewhat contrived, but I think I could express this typing mechanism if there was a way to somehow produce a "tagged" variant of an existing type:
```py
class MapAttribute(...):
    @overload
    def __get__(self: _MT, instance: Model, owner: Any) -> Tagged[_MT, "instance"]: ...
    @overload
    def __get__(self: _MT, instance: Type[Model], owner: Any) -> Tagged[_MT, "class"]: ...

class NumberAttribute(...):
    @overload
    def __get__(self, instance: Model, owner: Any) -> float: ...
    @overload
    def __get__(self, instance: Type[Model], owner: Any) -> NumberAttribute: ...
    @overload
    def __get__(self, instance: Tagged[MapAttribute, "instance"], owner: Any) -> float: ...
    @overload
    def __get__(self, instance: Tagged[MapAttribute, "class"], owner: Any) -> float: ...
```

This is, however, just one approach, and I'm not sure how viable or reasonable it is. I'd appreciate any other advice and suggestions.

gvanrossum:
Perhaps NewType will work?

[...]

PR #6111, opened 2018-12-28
Add pynamodb plugin

ikonst:
This plugin will implement aspects of the pynamodb library that cannot be expressed through the stubs alone.

This first version makes attributes with `nullable=True` have their underlying value be optional, i.e. given
```py
class MyModel(Model):
   my_attr = NumberAttribute(null=True)
```
then `MyModel.my_attr` will be typed `_NullableAttribute[NumberAttribute, Number]` and `MyModel().my_attr` will be typed `Optional[Number]`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "typeshed"]
 	path = mypy/typeshed
-	url = https://github.com/python/typeshed
+	url = https://github.com/ikonst/typeshed
+	branch = pynamodb-attr-nullable
--- a/mypy/plugins/pynamodb.py
+++ b/mypy/plugins/pynamodb.py
@@ -1,0 +1,40 @@
+from typing import Callable, Optional
+
+from mypy.nodes import NameExpr, CallExpr
+from mypy.plugin import MethodContext, Plugin
+from mypy.types import Instance, Type
+
+ATTR_FULL_NAME = 'pynamodb.attributes.Attribute'
+
+
+class MypyPlugin(Plugin):
+    def get_function_hook(self, fullname: str) -> Optional[Callable[[MethodContext], Type]]:
+        return _function_hook_callback
+
+
+def _function_hook_callback(ctx: MethodContext) -> Type:
+    if (
+        not isinstance(ctx.default_return_type, Instance) or
+        not ctx.default_return_type.type.has_base(ATTR_FULL_NAME) or
+        # TODO: any better way to tell apart a construction from other functions?
+        not isinstance(ctx.context, CallExpr) or
+        not ctx.context.callee.fullname == ctx.default_return_type.type.fullname()
+    ):
+        return ctx.default_return_type
+
+    attr_type = ctx.default_return_type
+    base_types = [base for base in attr_type.type.bases if base.type.fullname() == 'pynamodb.attributes.Attribute']
+    if not base_types:
+        return ctx.default_return_type
+    base_type = base_types[0]
+    underlying_type = base_type.args[0]
+
+    for arg_name, arg_expr in zip(ctx.context.arg_names, ctx.context.args):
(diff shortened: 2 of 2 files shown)


.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "typeshed"]
 	path = mypy/typeshed
-	url = https://github.com/python/typeshed
+	url = https://github.com/ikonst/typeshed
```
ikonst: 🔒pending on another pull request into the typeshed, which will introduce:
```py
class _NullableAttribute(Generic[_AT, _T], _AT):
    @overload
    def __get__(self, instance: Model, owner: Any) -> Optional[_T]: ...
    @overload
    def __get__(self: _NT, instance: None, owner: Any) -> _NT: ...
```

ikonst:
Following this, I plan to work on supporting `MapAttribute`s as discussed in https://github.com/python/mypy/issues/5272.

mypy/plugins/pynamodb.py
```diff
@@ -0,0 +1,47 @@
...
+        if isinstance(arg_expr, NameExpr) and arg_expr.fullname == 'builtins.True':
+            return ctx.api.named_generic_type('pynamodb.attributes._NullableAttribute', [
+                attr_type,
+                underlying_type,
+            ])
+
+    return ctx.default_return_type
+
+
+def plugin(version: str) -> PynamodbPlugin:
```
ilevkivskyi: The return annotation is wrong, it should be `Type[PynamodbPlugin]`. There are also few other errors in the failed self-check (see Travis logs).

mypy/plugins/pynamodb.py
```diff
@@ -0,0 +1,47 @@
...
+        return _function_hook_callback
+
+
+def _function_hook_callback(ctx: FunctionContext) -> Type:
+    if (
+        not isinstance(ctx.default_return_type, Instance) or
+        not ctx.default_return_type.type.has_base(ATTR_FULL_NAME) or
+        # TODO: any better way to tell apart a construction from other functions?
+        not isinstance(ctx.context, CallExpr) or
+        not isinstance(ctx.default_return_type.type, TypeInfo) or
```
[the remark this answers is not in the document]
ikonst: Thanks, it was a misguided attempt to make mypy happy.

mypy/plugins/pynamodb.py
ikonst: Fixed that one, thanks.

ikonst:
@ilevkivskyi I think the policy of _not_ shoving each and every library's plugin into mypy is sensible and good, so I'll try to move this plugin into the `pynamodb` project. (Along the same line, we should move the pynamodb stubs into the pynamodb project now that we have PEP-561.)

Thanks for the review, hope you don't mind if I keep asking you more questions?

mypy/plugins/pynamodb.py
```diff
@@ -0,0 +1,47 @@
...
+class PynamodbPlugin(Plugin):
+    def get_function_hook(self, fullname: str) -> Optional[Callable[[FunctionContext], Type]]:
+        return _function_hook_callback
+
+
+def _function_hook_callback(ctx: FunctionContext) -> Type:
+    if (
+        not isinstance(ctx.default_return_type, Instance) or
+        not ctx.default_return_type.type.has_base(ATTR_FULL_NAME) or
+        # TODO: any better way to tell apart a construction from other functions?
```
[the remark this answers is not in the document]
ikonst: I'm trying to match instantiation of `pynamodb.attribute.Attribute`-based classes. I'm trying to match:
- "function call" (hence `get_function_hook`)
- the return type of the function call has `Attribute` as a base
- the "function" being called has the same full name as the type it returns (hence its object instantiation)

My matching logic is probably a very awkward way of expressing what I actually want. I'll follow your recommendation w.r.t lookup function.

Issue #6060, opened 2018-12-12
mypy fails to narrow Union[float,int] to float after isinstance(x, float)


DarwinAwardWinner:
Mypy fails to typecheck the following valid code:

```python
from typing import Union
def maybe_convert_to_int(x: Union[float,int]) -> Union[float,int]:
    '''Convert x from float to int if it can be done losslessly.'''
    if isinstance(x, float):
        # x must be a float here, so this is a legal operation, but
        # mypy still thinks its type is Union[float,int]
        assert isinstance(x, float)
        if x.is_integer():
            x = int(x)
    return x
```

When I run mypy on it, I get:

```
$ mypy ~/temp/test_mypy.py
temp/test_mypy.py:8: error: Item "int" of "Union[float, int]" has no attribute "is_integer"
$ python --version
Python 3.7.1
$ mypy --version
mypy 0.650
```

I would expect this program to pass the type checker, since I am using `isinstance` to narrow the type to float before performing the float-specific operation `float.is_integer`. If I replace the assertion with `x = float(x)`, that satisfies the type checker.

ilevkivskyi:
This is because some special-casing in mypy called "type promotions". The point is that `int` is not a subclass of `float` at runtime, but mypy considers `int` a subtype of `float`. Because of this, `int` sneaks into the `if` statement, where mypy complains about the missing method.

We try to ignore the type promotions while working with `isinstance()`, but it looks like there is a bug and type promotions are still used. Another possible workaround (apart from the one you mentioned) for you would be to just annotate the functions as `(float) -> float`. You can probably also change the logic to
```python
def maybe_convert_to_int(x: float) -> float:
    if isinstance(x, int):
        return x
    if x.is_integer():
        x = int(x)
    return x
```

ilevkivskyi:
@Michael0x2a was among the last who worked with this code, so probably he has the most context.

DarwinAwardWinner:
I discovered a workaround that eliminated this code entirely for my specific use case. The reason for this code was to ensure that a float like `1.0` would print as an integer (with no decimal point), but I figured out that `'{:g}'.format(x)` does that without having to check the type of `x`.

However, I guess this is still a bug in the general case.

syastrov:
I can also reproduce this, or a similar issue having to do with statement reachability, both on mypy 0.740 and master (mypy 0.750+dev.e97377c454a1d5c019e9c56871d5f229db6b47b2)

```python
from typing import Union
[...]

PR #6114, opened 2018-12-29
Disable promotions in isinstance checks

Michael0x2a:
This pull request resolves https://github.com/python/mypy/issues/6060 by modifying `conditional_type_map` function in `checker.py` to ignore promotions.


Michael0x2a:
Note: I am somewhat suspicious of this solution. I'll try testing it on our internal codebases later today or tomorrow.

msullivan:
This does actually cause some trouble in internal code bases (with `str` and `unicode`). See #6141

Issue #2417, opened 2016-11-07
Add code to help filtering warning and errors in mypy's output


sametmax:
As mentioned in #2411, it can become difficult to select what you want to see or not in mypy's ouput. This can lead to all-or-nothing behaviors, clutter display/logs, and make CI servers fail on errors we don't wish to check or ignore errors we do want to check.

A possible solution to this is to adopt the behaviour from popular linters such as flake8, pylint, etc.

Each type of error/warning [have a code](http://pylint-messages.wikidot.com/all-codes), and you can pass the ones you wish to ignore as a coma separated list using either the command line, a code comment or [a config file](http://stackoverflow.com/questions/4341746/how-do-i-disable-a-pylint-warning). Each code is also displayed at the beginning of each respective warning/error in the tool output, so that you can grep for it easily.

E.G:

```
$ flake8 tests
tests/test_s.py:8:1: F821 undefined name 'test'
tests/test_s.py:10:1: E402 module level import not at top of file
tests/test_s.py:10:10: E261 at least two spaces before inline comment
tests/test_s.py:12:1: E402 module level import not at top of file
tests/test_s.py:14:1: E402 module level import not at top of file
$ flake8 tests --ignore=E402,E402
tests/test_s.py:8:1: F821 undefined name 'test'
tests/test_s.py:10:10: E261 at least two spaces before inline comment
```

This is a general way of solving the filtering problem:

- it works with one flag instead of many;
- it's flexible enough to cover potential future new use cases;
- with the config file, you can commit it in your VCS and share the checks you wish to apply with your team;
- it plays well with all configuration mechanisms: env var, config file, cmd options, code comments, etc.
- it plays well with the command line, allowing grep on specific code, or code type (in the example, F is a failure, E an error, W a warning, etc).

JukkaL:
Alternatively, we could have a command line option for each warning/error. We already have examples like `--strict-optional` and `--warn-no-return`. We could have two options for each error/warning that can be enabled/disabled: `--<feature-name>` and `--no-<feature-name>`. 

Here are some tradeoffs:
* An error code is more discoverable if we display it on every message.
* If we display an error code on every message the messages become more noisy.
* If we display an error code on every message many test cases need to be updated.
* An error code by itself is mostly meaningless and the meaning has to be looked up from documentation.
* A command-line option is easy to recognize and easy to remember if it has a descriptive name.
[...]

PR #6118, opened 2018-12-30
Migrate remaining failure messages to mypy.messages (part1: mypy.checker)

### LINKED ISSUE #6116: Migrate remaining failure messages to mypy.messages
I'm working on a tool to provide a more gradual mypy adoption path, by providing a fine grained configuration for selecting or ignoring mypy errors.  My current prototype uses regular expressions, but maintaining a comprehensive set of regular expressions is both time consuming and brittle, as existing messages could change, and new ones will be added. 

I noticed that an effort has already been made to move error messages from `mypy.checkexpr` into `mypy.messages`.   Making this set of constants complete would solve my problem, and hopefully be of benefit to you all as well.  Would you accept a pull request to migrate the remaining failure messages from `checkexpr`, `semanal`, `semanal_newtype`, and `fastparse2` to  `messages` constants?

chadrik:
Consolidate all mypy error messages into one place (`mypy.messages`).  This opens the door to some interesting features, such as adding language localization or a more systematic way of working with error messages, akin to error codes (#2417).  It also helps create more consistent messages.  e.g. it's now more apparent that there are a few outliers that use single quotes around type names.

I'm breaking this up change up into several PRs by module, this is the first and simplest.  Once all are complete, this will address #6116.

It's pretty straight-foward.  The biggest challenge is coming up with good constant names.  Let me know what you think.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2542,10 +2542,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             dunder_set_type, [TempNode(instance_type), TempNode(AnyType(TypeOfAny.special_form))],
             [nodes.ARG_POS, nodes.ARG_POS], context)
 
-        if not isinstance(inferred_dunder_set_type, CallableType):
-            self.fail(messages.DESCRIPTOR_SET_NOT_CALLABLE
-                      .format(inferred_dunder_set_type), context)
-            return AnyType(TypeOfAny.from_error), get_type, True
+        # should be handled by get_method above
+        assert isinstance(inferred_dunder_set_type, CallableType)
 
         if len(inferred_dunder_set_type.arg_types) < 2:
             # A message already will have been recorded in check_call
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -432,7 +432,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             # valid overloads.
             return None
         if len(defn.items) == 1:
-            self.fail('Single overload definition, multiple required', defn)
+            self.fail(messages.MULTIPLE_OVERLOADS_REQUIRED, defn)
 
         if defn.is_property:
             # HACK: Infer the type of the property.
@@ -882,16 +882,12 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                             if typ.arg_names[i] in ['self', 'cls']:
                                 if (self.options.python_version[0] < 3
                                         and is_same_type(erased, arg_type) and not isclass):
-                                    msg = ("Invalid type for self, or extra argument type "
-                                           "in function annotation")
+                                    msg = messages.INVALID_SELF_TYPE_OR_EXTRA_ARG
                                     note = '(Hint: typically annotations omit the type for self)'
                                 else:
-                                    msg = ("The erased type of self '{}' "
-                                           "is not a supertype of its class '{}'"
-                                           ).format(erased, ref_type)
+                                    msg = messages.ERASED_SELF_TYPE_NOT_SUPERTYPE.format(erased, ref_type)
                             else:
-                                msg = ("Self argument missing for a non-static method "
-                                       "(or an invalid type for self)")
+                                msg = messages.MISSING_OR_INVALID_SELF_TYPE
                             self.fail(msg, defn)
                             if note:
                                 self.note(note, defn)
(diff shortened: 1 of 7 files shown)


mypy/messages.py
```diff
@@ -109,6 +109,24 @@
     'Implicit generic "Any". Use \'{}\' and specify generic parameters'  # type: Final
 INCOMPATIBLE_TYPEVAR_VALUE = 'Value of type variable "{}" of {} cannot be {}'  # type: Final
 UNSUPPORTED_ARGUMENT_2_FOR_SUPER = 'Unsupported argument 2 for "super"'  # type: Final
+MULTIPLE_OVERLOADS_REQUIRED = 'Single overload definition, multiple required'  # type: Final
+MISSING_OR_INVALID_SELF_TYPE = \
+    "Self argument missing for a non-static method (or an invalid type for self)"  # type: Final
+ERASED_SELF_TYPE_NOT_SUPERTYPE = \
+    "The erased type of self '{}' is not a supertype of its class '{}'"  # type: Final
```
ilevkivskyi: We move towards using double quotes for types, so it would be great to use them here (in the same PR if there are only few tests to update, otherwise in a separate PR).

chadrik: Will do

mypy/checker.py
```diff
@@ -2543,7 +2543,8 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
             [nodes.ARG_POS, nodes.ARG_POS], context)
 
         if not isinstance(inferred_dunder_set_type, CallableType):
-            self.fail(messages.DESCRIPTOR_SET_NOT_CALLABLE, context)
+            self.fail(messages.DESCRIPTOR_SET_NOT_CALLABLE
+                      .format(inferred_dunder_set_type), context)
```
chadrik: The code has two variations of this error message:

1. `"__set__ is not callable"` : used here
2. `"{}.__set__ is not callable"` : used elsewhere

I'd like to standardize on the second, but I'm not familiar enough with the code to know how to get the descriptor name here.  My naive attempt above did not trigger any test errors, but it should have, so I think it's not tested.  How should I proceed?

mypy/checkexpr.py
```diff
@@ -2892,19 +2891,19 @@ def check_super_arguments(self, e: SuperExpr) -> None:
                     if not isinstance(instance_type, (Instance, TupleType)):
                         # Too tricky, give up.
                         # TODO: Do something more clever here.
-                        self.chk.fail(messages.UNSUPPORTED_ARGUMENT_2_FOR_SUPER, e)
+                        self.chk.fail(messages.UNSUPPORTED_ARG_2_FOR_SUPER, e)
```
chadrik: I renamed this constant because I wanted to standardize all of them on the shorthand "ARG" for argument, since the constants can get fairly lengthy.

chadrik:
I added in some of the satellite modules for `checker` since this PR was a bit too easy :)

mypy/messages.py
```diff
@@ -45,6 +44,7 @@
 NO_RETURN_EXPECTED = 'Return statement in function which does not return'  # type: Final
 INVALID_EXCEPTION = 'Exception must be derived from BaseException'  # type: Final
 INVALID_EXCEPTION_TYPE = 'Exception type must be derived from BaseException'  # type: Final
+RETURN_IN_GENERATOR = "'return' with value in async generator is not allowed"  # type: Final
```
JelleZijlstra: This should probably be `RETURN_IN_ASYNC_GENERATOR`, since we plausibly could have a similar message for normal generators in Python 2.

mypy/checker.py
ilevkivskyi: What you are doing is not correct. I think you should use `attribute_type.type.name()` here, and add a test.

chadrik: This particular error is only generated in the case of an inferred `__set__`.  I'm not sure how to trigger this.  I tried a few variations along the lines of the tests in `check-inference.test` but all went through different code paths.  Any suggestions?  Is there a chance this is a dead code path?

ilevkivskyi: Yes, it looks like this is currently dead code: we only allow `__set__` to be a method above, and a method (i.e. function definition) can't have a non-callable type. I think this can be replaced with an assert.

test-data/unit/check-selftype.test
```diff
@@ -259,11 +259,11 @@ from typing import TypeVar, Type
 
 T = TypeVar('T', bound=str)
 class A:
-    def foo(self: T) -> T:   # E: The erased type of self 'builtins.str' is not a supertype of its class '__main__.A'
+    def foo(self: T) -> T:   # E: The erased type of self "builtins.str" is not a supertype of its class "__main__.A"
```
ilevkivskyi: Looking at these errors, I think we should switch them to using "short form", as we do for many other errors (in a _separate_ PR).

mypy/checker.py
chadrik: Addressed in my latest commit

test-data/unit/check-selftype.test
chadrik: Will do.

chadrik:
I think this should be ready to go.

Issue #5989, opened 2018-12-02
Regression: error messages for literal expressions in invalid places are worse


Michael0x2a:
After https://github.com/python/mypy/pull/5947 (basic end-to-end support for literal types) lands, we'll start deferring some of the "invalid type" checks to the semantic analysis layer by wrapping them into a 'RawLiteralType'. This can, in some cases have an unintended side-effect of making some error messages worse. For example, writing `T = TypeVar('T', bound=1)` previously used to report error messages like "TypeVar 'bound' must be a type"; now they report error messages like `E: Invalid type: try using Literal[1] instead?`.

We ran into a similar case with 'NewType' -- in that case, we added some special-casing to restore the original error message. We could do the same thing here, and for other similar "type-alias-like" constructs such as NamedTuple or TypedDict.

Alternatively, we could explore adding `TypeOfAny.invalid_type` as Ivan [suggests here](https://github.com/python/mypy/pull/5947#discussion_r238108162) -- this could also potentially help us resolve https://github.com/python/mypy/issues/4030 at the same time.

Note to self: if it looks like literal types aren't going to make it into the next mypy release, we should at least fix this error and/or change any existing errors to make sure we never recommend using 'Literal[...]'. It would be confusing if error messages started talking about literal types before they're actually officially released.

Issue #4030, opened 2017-09-29
Better error message for "Invalid type"


ilevkivskyi:
There several situations where a quite terse error message is given: ``Invalid type "__main__.T"`` -- this could be an attempt to use a variable as a type, attempt to use function as a type, unsupported forward reference, maybe some other. I propose to make this error message more specific, or add a note:
```
error: Invalid type "__main__.f"
note: Functions can't be used as a type
```
and
```
error: Invalid type "__main__.Alias"
note: Variables can't be used as a type
```
The second can be also useful if someone accidentally created a variable instead of an alias. In this case we can even append a link to docs, if we detect that variable has type ``Type[...]``.

Another similar terse error message is ``Invalid base class``. We can say which base is invalid, this will be useful in case of multiple bases on the same line, and maybe also add a note in some cases explaining why it is invalid, like ``note: Union types can't be subclassed``, so that it is clear what went wrong.

gvanrossum:
Also e.g.
```py
class C:
  A = int

a: C.A  # error: Invalid type "_.C.A"
```

ilevkivskyi:
This case is actually quite important, especially until https://github.com/python/mypy/issues/3494 (document variables vs aliases) is not fixed.

ilevkivskyi:
A rare case where adding "Invalid type" would be actually helpful is the error message for using subscriptable metaclasses in type context, see #6313.

ilevkivskyi:
A similar example where we should give a better error message is (recently closed) https://github.com/python/mypy/issues/2477

ilevkivskyi:
One more example is cast targets, see https://github.com/python/mypy/issues/6862

PR #6121, opened 2018-12-31
Remove RawLiteralType synthetic type

Michael0x2a:
This diff changes how we track raw literal types in the semantic analysis phase. It makes the following changes:

1.  Removes the `RawLiteralType` synthetic type.

2.  Adds a new `TypeOfAny`: `TypeOfAny.invalid_type` as suggested in https://github.com/python/mypy/issues/4030.

3.  Modifies `AnyType` so it can optionally contain a new `RawLiteral` class. This class contains information about the underlying literal that produced that particular `TypeOfAny`.

4.  Adjusts mypy to stop recommending using `Literal[...]` when doing `A = NewType('A', 4)` or `T = TypeVar('T', bound=4)`.

    (The former suggestion is a bad one: you can't create a NewType of a Literal[...] type. The latter suggestion is a valid but stupid one: `T = TypeVar('T', bound=Literal[4])` is basically the same thing as `T = Literal[4]`.)

    This resolves https://github.com/python/mypy/issues/5989.

The net effect of this diff is that:

1.  RawLiteralTypes no longer leak during fine-grained mode, which should partially help unblock https://github.com/python/mypy/pull/6075.

2.  The way mypy handles literal expressions in types is "inverted".

    Previously, we by default assumed literal expressions would belong inside `Literal[...]` and tacked on some logic to make them convert into error `AnyTypes`. Now, we do the reverse: we start with an error `AnyType` and convert those into `Literal[...]`s as needed.

    This more closely mirrors the way mypy *used* to work before we started work on Literal types. It should also hopefully help reduce some of the cognitive burden of working on other parts of the semantic analysis code, since we no longer need to worry about the `RawLiteralType` synthetic type.

3.  We now have more flexibility in how we choose to handle invalid types: since they're just `Anys`, we have more opportunities to intercept and customize the exact way in which we handle errors.

    Also see https://github.com/python/mypy/issues/4030 for additional context. (This diff lays out some of the foundation work for that diff).

### DIFF
--- a/mypy/exprtotype.py
+++ b/mypy/exprtotype.py
@@ -8,7 +8,7 @@ from mypy.nodes import (
 from mypy.fastparse import parse_type_string
 from mypy.types import (
     Type, UnboundType, TypeList, EllipsisType, AnyType, Optional, CallableArgument, TypeOfAny,
-    RawLiteralType,
+    RawLiteral,
 )
 
 
@@ -39,9 +39,19 @@ def expr_to_unanalyzed_type(expr: Expression, _parent: Optional[Expression] = No
     if isinstance(expr, NameExpr):
         name = expr.name
         if name == 'True':
-            return RawLiteralType(True, 'builtins.bool', line=expr.line, column=expr.column)
+            return AnyType(
+                TypeOfAny.invalid_type,
+                raw_literal=RawLiteral(True, 'builtins.bool'),
+                line=expr.line,
+                column=expr.column,
+            )
         elif name == 'False':
-            return RawLiteralType(False, 'builtins.bool', line=expr.line, column=expr.column)
+            return AnyType(
+                TypeOfAny.invalid_type,
+                raw_literal=RawLiteral(False, 'builtins.bool'),
+                line=expr.line,
+                column=expr.column,
+            )
         else:
             return UnboundType(name, line=expr.line, column=expr.column)
     elif isinstance(expr, MemberExpr):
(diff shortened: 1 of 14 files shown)


Michael0x2a:
One meta-note:

I think my "fix" of just removing the assert from the astmerge visitor in https://github.com/python/mypy/pull/6075 might have actually been fine: after analyzing the fine-grained logic, it seemed to me that encountering RawLiteralTypes during that phase of the update logic might have actually been expected behavior.

However, I decided that it would just be easier to rip out RawLiteralType altogether instead of sitting down and formalizing this hunch.

I do feel a little bad that this PR tacks on some additional cruft to AnyType, but we're already tracking a lot of provenance-related info in that class, so I felt this wasn't too extreme of a change.

JukkaL:
This seems too hacky to me. The new responsibility of `AnyType` seems like an implementation shortcut but doesn't really make sense conceptually, at least to me. Is there something that makes it hard to fix the leaking of raw literal types in fine-grained incremental mode using the existing implementation technique?

Adding `TypeOfAny.invalid_type` while also keeping `RawLiteralType` sounds like a more promising approach to me, though I haven't thought about this very carefully.

ilevkivskyi:
I am actually with Jukka here.

Also, if we add `TypeOfAny.invalid_type` we can also easily remove this TODO item in `typeanal.py`:
```python
        # TODO: Would it be better to always return Any instead of UnboundType
        # in case of an error? On one hand, UnboundType has a name so error messages
        # are more detailed, on the other hand, some of them may be bogus.
        return t
```

ilevkivskyi:
Also, it would make sense to use `invalid_type` in some other places like this:
```python
        else:  # sym is None
            if self.third_pass:
                self.fail('Invalid type "{}"'.format(t.name), t)
                return AnyType(TypeOfAny.from_error)  # <- here
```

Michael0x2a:
> Adding TypeOfAny.invalid_type while also keeping RawLiteralType sounds like a more promising approach to me, though I haven't thought about this very carefully.

The main problem with this approach is that I think adding just `TypeOfAny.invalid_type` doesn't actually buy us much. It lets us determine broadly where that `TypeOfAny` came from, but I don't think we'd necessarily have enough context to do anything useful in places that end up using that `Any`.

And if we do add some context to `Any`, I think we'd end up with a solution similar to what this PR does. That was actually what I was hoping to eventual do: follow-up by generalizing `RawLiteral` so it context about any kind of invalid type (and I'd perhaps rename the class to `InvalidTypeContext` or something).

In any case, I'll look into just plugging the leak as well -- but I still think it's worth adding in some variation of this PR if we want to add `TypeOfAny.invalid_type`.

Michael0x2a:
> Is there something that makes it hard to fix the leaking of raw literal types in fine-grained incremental mode using the existing implementation technique?

Ok, after doing more poking around, I'm pretty confident that we're not actually leaking anything and that removing the assert is the from astmerge's `TypeReplaceVisitor.visit_raw_literal_type` is the correct thing to do.

In short, we're invoking this class within [NodeReplaceVisitor.fixup_type](https://github.com/python/mypy/blob/master/mypy/server/astmerge.py#L291). One of the places this function is called is inside [NodeReplaceVisitor.process_base_func](https://github.com/python/mypy/blob/master/mypy/server/astmerge.py#L184), which looks like this:

```python
def process_base_func(self, node: FuncBase) -> None:
    self.fixup_type(node.type)
    node.info = self.fixup(node.info)
    if node.unanalyzed_type:
        # Unanalyzed types can have AST node references
        self.fixup_type(node.unanalyzed_type)
```

It looks like `node.unanalyzed_type` has (intentionally) not been semantically analyzed, which is why we're encountering the RawLiteralTypes. This is also probably why TypeReplaceVisitor [implements visit_unbound_type](https://github.com/python/mypy/blob/master/mypy/server/astmerge.py#L401) (I didn't notice it did until just now).

ilevkivskyi:
@Michael0x2a as I mentioned in #6075, I am fine to merge it as is. So you can keep only the parts necessary to fix #5989 in this PR. Optionally, you can also fix the TODO item I mentioned above.

ilevkivskyi:
(Also I would keep `TypeOfAny.invalid_type` because it will be useful for #4030.)

Michael0x2a:
I think I'll just close this PR and submit a new one with the error message fixes.

PR #6124, opened 2019-01-01
Add intelligent indexing of tuples, NamedTuples, and TypedDict

Michael0x2a:
This pull request adds a preliminary implementation of intelligent indexing of tuples, NamedTuples, and TypedDicts. It uses the first approach we discussed earlier: modifying the existing plugins and special-casing code to also check if the expression has a Literal[...] type.

Once I'm finished with the baseline literal types implementation, I'll look into circling back and seeing how viable the second approach is (writing some sort of plugin that replaces the signatures of methods like `.__getitem__` or `.get()` with overloads that use the appropriate literal types).


mypy/checkexpr.py
```diff
@@ -2421,13 +2421,21 @@ def _get_value(self, index: Expression) -> Optional[int]:
...
-        item_name = index.value
+        if isinstance(index, (StrExpr, UnicodeExpr)):
+            item_name = index.value
+        else:
+            typ = self.accept(index)
+            if isinstance(typ, LiteralType) and isinstance(typ.value, str):
+                item_name = typ.value
+            else:
+                self.msg.typeddict_key_must_be_string_literal(td_type, index)
+                return AnyType(TypeOfAny.from_error)
```
msullivan: Maybe not really a question for this diff, but is there a reason for this to not live in  the typeddict plugin as a `__getitem__` hook?

There is maybe a more relevant question of: could this just use `try_getting_str_literal`, the answer to which would be more obvious if this lived in the plugin.

test-data/unit/check-literal.test
```diff
@@ -2080,3 +2080,138 @@ def func(x: Literal[1], y: Literal[2]) -> None: pass
...
+# TODO: Make the runtime and mypy behaviors here consistent
+#
+# At runtime, all eight of the below operations will successfully return
+# the int because b"key" == u"key" in Python 2.
+#
+# Mypy, in contrast, will accept all the four calls to `some_dict[...]`
+# but will reject `normal_dict.get(unicode_key)` and `unicode_dict.get(unicode_key)`
+# because the signature of `.get(...)` accepts only a str, not unicode.
+#
+# I don't think this has anything to do with Literal types: we had the same
```
msullivan: sentence trails off?

mypy/checkexpr.py
Michael0x2a: I agree -- I think we should move both this and `tuple.__getitem__` out into a plugin, but it felt like it'd be out-of-scope for this PR.

PR #6098, opened 2018-12-22
Make `refresh_top_level()` more like `visit_file()` in the third pass

ilevkivskyi:
Fixes #6032 (first part)

This is a very straightforward fix. I am actually surprised no one found this before (maybe because mypy daemon is not so widely used yet).

In future, it would be great to have a mechanism to avoid such mistakes, by somehow making `refresh_top_level()` to do all the special things `visit_file()` does. I don't have any ideas now, but it probably makes sense to consider this during semantic analyzer refactoring (another possible solution is to avoid any such special things, but this should be clearly documented).

PR #6126, opened 2019-01-02
Second part of #6098

ilevkivskyi:
This PR fixes insufficient traversal of synthetic `TypeInfo`s in fine grained mode.

### DIFF
--- a/test-data/unit/fine-grained.test
+++ b/test-data/unit/fine-grained.test
@@ -8206,6 +8206,35 @@ x = 'no way'
 ==
 main:10: error: Incompatible types in assignment (expression has type "str", variable has type "int")
 
+[case testNamedTupleForwardFunctionDirect]
+# flags: --ignore-missing-imports
+from typing import NamedTuple
+from b import B
+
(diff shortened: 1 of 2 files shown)


test-data/unit/fine-grained.test
```diff
@@ -8206,6 +8206,35 @@ x = 'no way'
 ==
 main:10: error: Incompatible types in assignment (expression has type "str", variable has type "int")
 
+[case testNamedTupleForwardFunctionDirect]
+# flags: --ignore-missing-imports
+from typing import NamedTuple
+from b import B
+
+NT = NamedTuple('NT', [('x', B)])
```
JukkaL: Maybe also try revealing the type of the `x` attribute?

ilevkivskyi: I added separate test (since `reveal_type` is an error and affects how targets are processed). This reminded me about an old question/TODO in `typeanal.py`. We still keep `UnboundType`s in some places instead of replacing them with `Any`. Maybe we should stop doing this?(Note this requires updating many test cases.)

PR #6127, opened 2019-01-02
Speed up mypy startup by avoiding imports

JukkaL:
Move some imports that aren't always needed to functions bodies. This
speeds up mypy startup by up to 120ms on my laptop (when not compiled). 
Tests are a bit faster as well.

### DIFF
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -5,7 +5,6 @@ import pathlib
 import re
 import subprocess
 import sys
-from xml.sax.saxutils import escape
 from typing import TypeVar, List, Tuple, Optional, Dict, Sequence
 
 MYPY = False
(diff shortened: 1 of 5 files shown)


mypy/util.py
```diff
@@ -205,6 +203,7 @@ def correct_relative_import(cur_mod_id: str,
 
 
 def get_class_descriptors(cls: 'Type[object]') -> Sequence[str]:
+    import inspect  # Lazy import for minor startup speed win
```
msullivan: What were you timing? I think `get_class_descriptors` is unfortunately actually on the path for a lot of stuff, even not the daemon, so I suspect this won't speed up real workloads? Though I'm not sure.

mypy/util.py
```diff
@@ -143,7 +141,7 @@ def try_find_python2_interpreter() -> Optional[str]:
 
 
 def write_junit_xml(dt: float, serious: bool, messages: List[str], path: str) -> None:
-    """XXX"""
+    from xml.sax.saxutils import escape
```
msullivan: Huh, I thought I had already made this change.

mypy/util.py
JukkaL: I might have been timing `mypy --help` for this. I timed other changes by type checking a small program. Since this is a very small change I think it's okay to keep it even if it doesn't help in any realistic scenarios.

Issue #6125, opened 2019-01-02
Stdout hijacking in api.py is not thread safe


randomstuff:
See https://github.com/tomv564/pyls-mypy/issues/19, AFAIU mypy `api.py` relying on `sys.stdout` overriding is not thread-safe and causes issues when `sys.stout` and `sys.stderr` are used concurrently.

gvanrossum:
This is a good catch. It may be a bit tricky to fix, but we’re open to PRs. There probably aren’t too many places where stdout or street are used, but passing something in may require a bunch of messy signature changes.

JukkaL:
It would be better to explicitly pass the stdout and stderr objects instead of modifying them in `sys`. Would you like to contribute a fix? This should be a pretty easy one.

Here's a summary of what's involved for anybody interested in helping with this:

* Don't modify `sys.stdout` and `sys.stderr` in `mypy.api`.
* Instead, pass relevant `StringIO` objects to `mypy.main.main` as arguments. These can default to `sys.stdin` and `sys.stderr` in `main`.
* Use these arguments when we generate error messages. 
  * No need to update dmypy, stubgen and logging to use these, since they aren't relevant when using `api.py`.

elkhadiy:
Erm... Well it seems Github happily shows any reference of an issue from any fork whatsoever... that's good to know.

I wanted to wait for tomorrow so my colleague could test it on his setup since he was more "lucky" than me in getting the race conditions to "work" on his machine(tm). I guess nothing's wrong in putting the PR up for review now.

ethanhs:
@elkhadiy, take the time you need to make a good change set. Don't worry we won't jump on you :)

AnOctopus:
What remains to be done about this issue? It causes a lot of problems on my setup (using pyls-mypy and spacemacs) and I'd be willing to do some work to fix it but the changes @elkhadiy made on their fork seem to already cover the needed changes.

ethanhs:
@AnOctopus I believe the fork was committed to but a PR was never opened. If you would like to fix this, please feel free to open a PR! (Or perhaps @elkhadiy is still interested?)

AnOctopus:
PR #6750 opened.

PR #6129, opened 2019-01-02
Fix sys.stdout overriding in mypy.api (#6125)

elkhadiy:
Overriding sys.stdout and sys.stderr in mypy.api is not threadsafe.
This causes problems sometimes when using the api in pyls for example.

I am struggling a little with getting the right type for `f`, was expecting something like `Union[Callable[[TextIO, TextIO], None], Callable[[], None]]` to work.

### DIFF
--- a/mypy/api.py
+++ b/mypy/api.py
@@ -44,36 +44,31 @@ print ('\nExit status:', result[2])
 
 import sys
 from io import StringIO
-from typing import List, Tuple, Callable
+from typing import List, Tuple, Union, TextIO, Callable
+from mypy_extensions import DefaultArg
 
 
-def _run(f: Callable[[], None]) -> Tuple[str, str, int]:
-    old_stdout = sys.stdout
-    new_stdout = StringIO()
-    sys.stdout = new_stdout
+def _run(f: Callable[[TextIO, TextIO], None]) -> Tuple[str, str, int]:
 
-    old_stderr = sys.stderr
-    new_stderr = StringIO()
-    sys.stderr = new_stderr
+    stdout = StringIO()
+    stderr = StringIO()
 
     try:
-        f()
+        f(stdout, stderr)
         exit_status = 0
     except SystemExit as system_exit:
         exit_status = system_exit.code
-    finally:
-        sys.stdout = old_stdout
-        sys.stderr = old_stderr
 
-    return new_stdout.getvalue(), new_stderr.getvalue(), exit_status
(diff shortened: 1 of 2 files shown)


elkhadiy:
> * There are a handful of error message `print`s in `build.py`

Thanks I'll check that and try and look around to see if we missed anything else.

> * In `errors.py`, there are a bunch of prints in `report_internal_error` for displaying a traceback. (I think that a case /could/ be made for having those bypass the API capture, though we'd probably also want to arrange for the failure to bypass that as well? By using a `MypyInternalError` exception that we turn into `sys.exit(2)` in `__main__.py` but let propagate in the api?)

Propagating exceptions through the api sounds like a wise decision. For example in `mypy/api.py:1036`:
```python
def fail(msg: str, stderr: TextIO) -> None:
    stderr.write('%s\n' % msg)
    sys.exit(1)
```
Would it be better to raise an exception instead of `sys.exit(1)`?

> * The `argparse.ArgumentParser` argument processing will print to `sys.stderr`. To avoid this I think we need to create a subclass that overrides the error method.

Ah yes this little guy https://docs.python.org/dev/library/argparse.html#argparse.ArgumentParser.error needs to change.

> It makes me wonder if the stable API ought to just use subprocess and invoke mypy in a fresh interpreter? It would be slower, but easier to make correct.

This is actually what we were first planning to do in https://github.com/tomv564/pyls-mypy/blob/master/pyls_mypy/plugin.py#L56, with something like:
```python
res = subprocess.run(
    ["python", "-m", "mypy", *args],
    stdout=subprocess.PIPE, stderr=subprocess.PIPE
    )
report = res.stdout
errors = res.stderr
```

> (Of course, I added mypy.api.run_dmypy specifically to avoid the performance penalty associated with interpreter startup, but we could make that an unstable API and document its thread safety limitations.)

This could also be a solution, I like it! Have the normal `mypy.api.run` be threadsafe and potentially slower and propose `mypy.api.run_dmypy` as an unstable alternative were we continue hijacking `stdout` and `stderr` while potentially ironing out missing things.

Let me first work with your comments and we can go from there.

gvanrossum:
If you’re just going to use subprocess, why bother with mypy.api at all?

mr-c:
Should this be closed in light of https://github.com/python/mypy/pull/6750 being merged?

gvanrossum:
Superseded by #6750.

PR #6133, opened 2019-01-03
Refactor: move MRO calculation to a new module

JukkaL:


mypy/semanal.py
```diff
@@ -93,7 +93,7 @@
     infer_reachability_of_if_statement, infer_condition_value, ALWAYS_FALSE, ALWAYS_TRUE,
     MYPY_TRUE, MYPY_FALSE
 )
-from mypy.typestate import TypeState
+from mypy.mro import calculate_mro, MroError
```
ilevkivskyi: Currently, third pass does `self.sem.calculate_class_mro(tdef)`, maybe we can also import `calculate_mro` there to reduce coupling between second and third passes?

PR #6137, opened 2019-01-04
Collect timing stats and allow reporting of stats from the daemon 

msullivan:
There are a couple related pieces here:
 * Add a bunch of timing information to the stats tracked by BuildManager
 * Include all the stats in the dmypy server responses
 * Add a --telemetry-file argument to dmypy check et. al that writes out
   the stats information to a file

This will allow the collection of much more detailed internal
profiling information.


mypy/dmypy.py
```diff
@@ -71,12 +71,14 @@ def __init__(self, prog: str) -> None:
 p.add_argument('-v', '--verbose', action='store_true', help="Print detailed status")
 p.add_argument('-q', '--quiet', action='store_true', help=argparse.SUPPRESS)  # Deprecated
 p.add_argument('--junit-xml', help="Write junit.xml to the given file")
+p.add_argument('--telemetry-file', help='write telemetry information to the given file')
```
JukkaL: Maybe reconsider the name of the option. The information will be part of internal telemetry at Dropbox, but telemetry is mostly implemented in internal scripts not included in the mypy project. There is no built-in telemetry support in dmypy which may make this confusing. Random idea: call this `--perf-stats-file` or something.

msullivan: `--perf-stats-file` is good. I wanted to just call it `--stats-file` but decided against that because it is too similar to `--status-file`.

PR #6142, opened 2019-01-04
Disable promotions when restricting types in isinstance checks

### LINKED ISSUE #6141: Else block of `isinstance` check with promotable types has <nothing> for type
This test case
```
from typing import Union
def asdf(val: Union[int, float]):
    if isinstance(val, float):
        reveal_type(val)
    else:
        reveal_type(val)
```
produces:
```
/home/msullivan/lol2.py:4: error: Revealed type is 'builtins.float'
/home/msullivan/lol2.py:6: error: Revealed type is '<nothing>'
```
But the type in the `else` branch ought to be `int`.

Prior to  #6114, it produced
```
/home/msullivan/lol2.py:4: error: Revealed type is 'Union[builtins.int, builtins.float]'
```
which is of course also wrong.
The prior version produced false positives in the then branch and false negatives in the else branch, while the new version produces false positives in the else branch and works correctly in the then branch.

msullivan:
This fixes a bug left behind by #6114, where too much would get
subtracted out of a union type after an isinstance.

Fixes #6141.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3764,7 +3764,8 @@ def conditional_type_map(expr: Expression,
                 proposed_precise_type = UnionType([type_range.item
                                           for type_range in proposed_type_ranges
                                           if not type_range.is_upper_bound])
-                remaining_type = restrict_subtype_away(current_type, proposed_precise_type)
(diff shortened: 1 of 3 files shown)


msullivan:
Argh the test failure is the windows IPC flake, merging

PR #6146, opened 2019-01-05
Fix IPC timeout incoherence

msullivan:
Make IPCClient's timeout parameter be consistently seconds,
rather than seconds on unix and milliseconds on windows.

### DIFF
--- a/mypy/ipc.py
+++ b/mypy/ipc.py
@@ -94,10 +94,10 @@ class IPCBase:
 class IPCClient(IPCBase):
     """The client side of an IPC connection."""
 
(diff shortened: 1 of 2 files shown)


msullivan:
Aha, and this does actually fix the dmypy timeout flake. While *most* dmypy commands default to not specifying a timeout, `stop` uses a 5 second timeout. When interpreted as 5ms, failing is pretty plausible.

I validated that this was the issue by adding `time.sleep(1)` before `ConnectedNamedPipe` in `IPCServer.__enter__`. It failed, and the timeout fix fixes it.

msullivan:
So I think this lays to rest #5983!

PR #6148, opened 2019-01-05
Implement overlapped I/O and timeouts on server side Windows IPC

ethanhs:
This PR does a few things:
- Changes reads/writes/connects to use overlapped I/O on Windows
- Removes the need for special casing `b''` as an `EOF` signal
- Adds support for timing out on connects for the server process (which means dmypy's --timeout flag should work on Windows).
- Make the default timeout infinite (though this doesn't change anything functionally since we always specify alternatives in our usage)

Note that sadly this will fail self type checking since their signatures in typeshed are for the non-overlapped versions (I will be making a PR to typeshed soon).

A lot of the overlapped I/O is based on the docs and https://github.com/python/cpython/blob/master/Lib/multiprocessing/connection.py. The main difference is the timeouts and I used `WaitForSingleObject`, since that seems sufficient for our purposes.

### DIFF
--- a/mypy/ipc.py
+++ b/mypy/ipc.py
@@ -47,9 +47,14 @@ class IPCBase:
 
     connection = None  # type: _IPCHandle
 
-    def __init__(self, name: str) -> None:
+    def __init__(self, name: str, timeout: Optional[float]) -> None:
         self.READ_SIZE = 100000
         self.name = name
+        if sys.platform == 'win32' and timeout is not None:
+            # Windows timeouts are in miliseconds
+            self.timeout = int(timeout * 1000)
+        else:
+            self.timeout = timeout
 
     def read(self) -> bytes:
         """Read bytes from an IPC connection until its empty."""
@@ -59,7 +64,8 @@ class IPCBase:
                 ov, err = _winapi.ReadFile(self.connection, self.READ_SIZE, overlapped=True)
                 try:
                     if err == _winapi.ERROR_IO_PENDING:
-                        res = _winapi.WaitForSingleObject(ov.event, _winapi.INFINITE)
+                        timeout = self.timeout if self.timeout else _winapi.INFINITE
+                        res = _winapi.WaitForSingleObject(ov.event, timeout)
                         assert res == _winapi.WAIT_OBJECT_0
                 except BaseException:
                     ov.cancel()
@@ -88,7 +94,8 @@ class IPCBase:
                 ov, err = _winapi.WriteFile(self.connection, data, overlapped=True)
                 try:
                     if err == _winapi.ERROR_IO_PENDING:
-                        res = _winapi.WaitForSingleObject(ov.event, _winapi.INFINITE)
+                        timeout = self.timeout if self.timeout else _winapi.INFINITE
+                        res = _winapi.WaitForSingleObject(ov.event, timeout)
                         assert res == _winapi.WAIT_OBJECT_0
                 except BaseException:
                     ov.cancel()
@@ -115,9 +122,9 @@ class IPCClient(IPCBase):
     """The client side of an IPC connection."""
 
     def __init__(self, name: str, timeout: Optional[float]) -> None:
-        super().__init__(name)
+        super().__init__(name, timeout)
         if sys.platform == 'win32':
-            timeout = int(timeout * 1000) if timeout else _winapi.NMPWAIT_WAIT_FOREVER
+            timeout = self.timeout if self.timeout else _winapi.NMPWAIT_WAIT_FOREVER
             try:
                 _winapi.WaitNamedPipe(self.name, timeout)
             except FileNotFoundError:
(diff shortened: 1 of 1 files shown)


mypy/ipc.py
```diff
@@ -47,36 +47,63 @@ class IPCBase:
...
+                ov, err = _winapi.ReadFile(self.connection, self.READ_SIZE, overlapped=True)
+                try:
+                    if err == _winapi.ERROR_IO_PENDING:
+                        timeout = self.timeout if self.timeout else _winapi.INFINITE
+                        res = _winapi.WaitForSingleObject(ov.event, timeout)
+                        assert res == _winapi.WAIT_OBJECT_0
+                except BaseException:
+                    ov.cancel()
+                    raise
+                finally:
```
msullivan: I've not fully reviewed this, but one initial thing: I don't immediately understand the logic here of why this is in a finally block?

ethanhs: Yeah I agree it is weird, this is an artifact of the multiprocessing code. I think it should be okay to remove it.

mypy/ipc.py
```diff
@@ -47,36 +47,66 @@ class IPCBase:
...
+                try:
+                    if err == _winapi.ERROR_IO_PENDING:
+                        timeout = self.timeout if self.timeout else _winapi.INFINITE
+                        res = _winapi.WaitForSingleObject(ov.event, timeout)
+                        assert res == _winapi.WAIT_OBJECT_0
+                except BaseException:
+                    ov.cancel()
+                    raise
+                _, err = ov.GetOverlappedResult(True)
+                buff = ov.getbuffer()
```
gvanrossum: It looks like `buff` has the same role as `more` below. If they are also of the same type, consider unifying the variable name?

mypy/ipc.py
```diff
@@ -47,36 +47,66 @@ class IPCBase:
...
+                assert isinstance(err, int)
+                try:
+                    if err == _winapi.ERROR_IO_PENDING:
+                        timeout = self.timeout if self.timeout else _winapi.INFINITE
+                        res = _winapi.WaitForSingleObject(ov.event, timeout)
+                        assert res == _winapi.WAIT_OBJECT_0
+                except BaseException:
+                    ov.cancel()
+                    raise
+                _, err = ov.GetOverlappedResult(True)
```
gvanrossum: I suppose this should be reached when `err` is zero as well, indicating that somehow the `ReadFile` operation returned an instant result? So far so good. But what if `err` has some *other* value? Will `GetOverlappedResult` just pass that through? Or should you check for that?

mypy/ipc.py
```diff
@@ -47,36 +47,66 @@ class IPCBase:
...
+                assert isinstance(err, int)
+                try:
+                    if err == _winapi.ERROR_IO_PENDING:
+                        timeout = self.timeout if self.timeout else _winapi.INFINITE
+                        res = _winapi.WaitForSingleObject(ov.event, timeout)
+                        assert res == _winapi.WAIT_OBJECT_0
+                except BaseException:
+                    ov.cancel()
+                    raise
+                bytes_written, err = ov.GetOverlappedResult(True)
```
gvanrossum: Similar.

mypy/ipc.py
```diff
@@ -47,36 +47,66 @@ class IPCBase:
 
     connection = None  # type: _IPCHandle
 
-    def __init__(self, name: str) -> None:
+    def __init__(self, name: str, timeout: Optional[float]) -> None:
         self.READ_SIZE = 100000
         self.name = name
+        if sys.platform == 'win32':
+            # Windows timeouts are in miliseconds
+            self.timeout = int(timeout * 1000) if timeout is not None else None
```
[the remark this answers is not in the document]
ethanhs: The reason is the APIs use milliseconds, so I guess keeping it near the API calls is best. I'd rather not have separate variables and having to worry about what is defined/what isn't.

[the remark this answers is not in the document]
gvanrossum: SGTM.

mypy/ipc.py
ethanhs: I took a look at the [source](https://github.com/python/cpython/blob/master/Modules/_winapi.c#L202) and the possible values the err value can take (other values raise exceptions):
```
_winapi.ERROR_SUCCESS
_winapi.ERROR_MORE_DATA
_winapi.ERROR_OPERATION_ABORTED
_winapi.ERROR_IO_INCOMPLETE
```
`_winapi.ERROR_IO_INCOMPLETE` can't happen in blocking mode (which we are). More data means we should read again, so I guess I just need to handle operation aborted (it seems CPython does not, for some reason).

gvanrossum: OK, then maybe the code in the try/except should do something like
```
if err != 0:
    assert err == _winapi.ERROR_IO_PENDING
    # rest of the code
```

PR #6149, opened 2019-01-06
Improve error messages related to literal types

Michael0x2a:
This pull request improves how we handle error messages with Literal types in three different ways:

1. When the user tries constructing types like `Literal[3 + a]`, we now report an error message that says 'Invalid type: Literal[...] cannot contain arbitrary expressions'.

2. We no longer recommend using Literal[...] when doing `A = NewType('A', 4)` or `T = TypeVar('T', bound=4)`. This resolves https://github.com/python/mypy/issues/5989.

    (The former suggestion is a bad one: you can't create a NewType of a Literal[...] type. The latter suggestion is a valid but stupid one: `T = TypeVar('T', bound=Literal[4])` is basically the same thing as `T = Literal[4]`.)

3. When the user tries using complex numbers inside Literals (e.g. `Literal[3j]`), we now report an error message that says 'Parameter 1 of Literal[...] cannot be of type "complex"'. This is the same kind of error message we previously used to report when the user tried using floats inside of literals.

In order to accomplish bullet point 1, moved the "invalid type comment or annotation" checks from the parsing layer to the semantic analysis layer. This lets us customize which error message we report depending on whether or not the invalid type appears in the context of a Literal[...] type.

In order to accomplish this, I repurposed RawLiteralType so it can represent any arbitrary expression that does not convert directly into a type (and renamed 'RawLiteralType' to 'RawExpressionType' to better reflect this new usage). I also added an optional "note" field to that class: this lets the parsing layer attach some extra context that would be difficult to obtain up in the semantic analysis layer.

In order to accomplish bullet point 2, I modified the type analyzer so that the caller can optionally suppress the error messages that would otherwise be generated when a RawExpressionType appears outside of a Literal context.

Bullet point 3 only required a minor tweak to the parsing and error handling code.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -1068,7 +1068,16 @@ class TypeConverter:
         self.node_stack = []  # type: List[AST]
         self.assume_str_is_unicode = assume_str_is_unicode
 
-    def _invalid_type(self, node: AST, note: Optional[str] = None) -> RawExpressionType:
+    def invalid_type(self, node: AST, note: Optional[str] = None) -> RawExpressionType:
+        """Constructs a type representing some expression that normally forms an invalid type.
+        For example, if we see a type hint that says "3 + 4", we would transform that
+        expression into a RawExpressionType.
+
+        The semantic analysis layer will report an "Invalid type" error when it
+        encounters this type, along with the given note if one is provided.
+
+        See RawExpressionType's docstring for more details on how it's used.
+        """
         return RawExpressionType(
             None,
             'typing.Any',
@@ -1094,7 +1103,7 @@ class TypeConverter:
             if visitor is not None:
                 return visitor(node)
             else:
-                return self._invalid_type(node)
+                return self.invalid_type(node)
         finally:
             self.node_stack.pop()
 
@@ -1134,7 +1143,7 @@ class TypeConverter:
             note = None
             if constructor:
                 note = "Suggestion: use {0}[...] instead of {0}(...)".format(constructor)
-            return self._invalid_type(e, note=note)
+            return self.invalid_type(e, note=note)
         if not constructor:
             self.fail("Expected arg constructor name", e.lineno, e.col_offset)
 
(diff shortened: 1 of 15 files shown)


Michael0x2a:
Two meta notes:

First, I broke up these changes into 4 separate commits to try and make it a little easier to track down the non-trivial changes during code review.

Second, I experimented with and ultimately decided to reject adding `TypeOfAny.invalid_type`: I found it didn't really help, nor did it seem like it would help us with our existing TODOs and issues.

Specifically, adding that constant wouldn't really help with https://github.com/python/mypy/issues/4030: that's really more of an issue of carefully adding more checks to the type checking or semantic analysis layer.

It also didn't seem like it would help with the [TODO listed here](https://github.com/python/mypy/pull/6121#issuecomment-451163086): returning AnyType instead of UnboundType there ended up making the error messages worse. We would need to attach some additional context to AnyType to preserve the current level of info in the error messages, but we more or less rejected that approach in https://github.com/python/mypy/pull/6121.

We can maybe look into adding `TypeOfAny.invalid_type` in a future PR along with additional changes that actually do end up taking advantage of that new Any kind.

mypy/fastparse.py
```diff
@@ -1069,6 +1068,15 @@ def __init__(self,
         self.node_stack = []  # type: List[AST]
         self.assume_str_is_unicode = assume_str_is_unicode
 
+    def _invalid_type(self, node: AST, note: Optional[str] = None) -> RawExpressionType:
```
ilevkivskyi: No need for this to be private. Also a short docstring would be great.

mypy/semanal_newtype.py
```diff
@@ -114,11 +114,11 @@ def check_newtype_args(self, name: str, call: CallExpr, context: Context) -> Opt
             self.fail(msg, context)
             return None
 
-        old_type = self.api.anal_type(unanalyzed_type)
+        old_type = self.api.anal_type(unanalyzed_type, report_invalid_types=False)
```
ilevkivskyi: This looks a bit ad hoc TBH. I however don't have a better proposal. Could you please add a comment why we pass `False` in this particular case?

PR #6150, opened 2019-01-06
Add docs for Literal types

Michael0x2a:
This pull request adds some documentation on using Literal types.

It focuses mostly on how they can be parameterized, and different ways of declaring variables to be Literal.

These docs deliberately do not mention the "intelligent indexing" feature: we don't yet support intelligently indexing using variables that are declared to be Final, so I think we should leave this feature undocumented for now.

(I'll add in that feature + update the docs once https://github.com/python/mypy/pull/6081 lands.)

### DIFF
--- a/docs/source/more_types.rst
+++ b/docs/source/more_types.rst
@@ -1051,14 +1051,14 @@ type signature for this function using Literal and overloads:
 Parameterizing Literals
 -----------------------
 
-Literal types may contain one or more literal bools, ints, strings, and byte
+Literal types may contain one or more literal bools, ints, strs, and byte
 strings. However, Literal types **cannot** contain arbitrary expressions:
 types like ``Literal[my_string.trim()]``, ``Literal[x > 3]``, or ``Literal[3j + 4]``
 are all illegal.
 
 Literals containing two or more values are equivalent to the union of those values.
 So, ``Literal[-3, b"foo", True]`` is equivalent to
-``Union[Literal[-3], Literal[b"foo"], Literal[True]]``. This can help make writing
+``Union[Literal[-3], Literal[b"foo"], Literal[True]]``. This makes writing
 more complex types involving Literals a little more convenient.
 
 Literal types may also contain ``None``. Mypy will treat ``Literal[None]`` as being
@@ -1089,7 +1089,7 @@ plan on adding support for using enum values inside Literals in an upcoming rele
 Declaring Literal variables
 ---------------------------
 
-You must explicitly add an annotation to a variable to declare that it is
+You must explicitly add an annotation to a variable to declare that it has
 a Literal type:
 
 .. code-block:: python
@@ -1137,7 +1137,7 @@ This is why ``expects_literal(19)`` type-checks despite the fact that ``reveal_t
 reports ``int``.
 
 So while declaring a variable to be final is not quite the same thing as adding
-an explicit Literal annotation, it often leads to the same effect  in practice.
+an explicit Literal annotation, it often leads to the same effect in practice.
 
 Limitations
 -----------
--- a/docs/source/more_types.rst
+++ b/docs/source/more_types.rst
@@ -1009,9 +1009,9 @@ equal to specifically the string ``"foo"``.
 
 This feature is primarily useful when annotating functions that behave
 differently based on the exact value the caller provides. For example,
-suppose we have a function ``fetch_data(...)`` that returns bytes if the
-first argument is True, and str if it's False. We can construct a precise
-type signature for this function using Literal and overloads:
+suppose we have a function ``fetch_data(...)`` that returns ``bytes`` if the
+first argument is ``True``, and ``str`` if it's ``False``. We can construct a
+precise type signature for this function using Literal and overloads:
 
 .. code-block:: python
 
@@ -1033,26 +1033,23 @@ type signature for this function using Literal and overloads:
     def fetch_data(raw: bool) -> Union[bytes, str]: ...
 
     def fetch_data(raw: bool) -> Union[bytes, str]:
-        # (Implementation is omitted)
-        pass
-
-    variable_1: Literal[True] = True
+        # Implementation is omitted
+        ...
 
     reveal_type(fetch_data(True))        # Revealed type is 'bytes'
     reveal_type(fetch_data(False))       # Revealed type is 'str'
-    reveal_type(fetch_data(variable_1))  # Revealed type is 'bytes'
 
     # Variables declared without annotations will continue to have an
     # inferred type of 'bool'.
 
-    variable_2 = True
-    reveal_type(fetch_data(variable_2))  # Revealed type is 'Union[bytes, str]'
+    variable = True
+    reveal_type(fetch_data(variable))    # Revealed type is 'Union[bytes, str]'
 
 Parameterizing Literals
 -----------------------
 
-Literal types may contain one or more literal bools, ints, strs, and byte
-strings. However, Literal types **cannot** contain arbitrary expressions:
+Literal types may contain one or more literal bools, ints, strs, and bytes.
+However, Literal types **cannot** contain arbitrary expressions:
 types like ``Literal[my_string.trim()]``, ``Literal[x > 3]``, or ``Literal[3j + 4]``
 are all illegal.
 
@@ -1074,7 +1071,7 @@ is legal:
     SecondaryColors = Literal["purple", "green", "orange"]
     AllowedColors = Literal[PrimaryColors, SecondaryColors]
 
-    def paint(color: AllowedColors) -> None: pass
+    def paint(color: AllowedColors) -> None: ...
 
     paint("red")        # Type checks!
     paint("turquoise")  # Does not type check
@@ -1106,7 +1103,7 @@ are **not** assumed to be Literals:
     reveal_type(b)          # Revealed type is 'int'
 
 If you find repeating the value of the variable in the type hint to be tedious,
-you can instead declare the variable to be :ref:`Final <final_attrs>`:
+you can instead change the variable to be :ref:`Final <final_attrs>`:
 
 .. code-block:: python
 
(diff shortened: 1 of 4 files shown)


Michael0x2a:
Also, we should probably delay merging this PR until we've released a new version of `typing_extensions` to pypi.

docs/source/more_types.rst
```diff
@@ -985,3 +989,165 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+-----------------------
+
+Literal types may contain one or more literal bools, ints, strings, and byte
+strings. However, Literal types **cannot** contain arbitrary expressions:
+types like ``Literal[my_string.trim()]``, ``Literal[x > 3]``, or ``Literal[3j + 4]``
+are all illegal.
+
+Literals containing two or more values are equivalent to the union of those values.
+So, ``Literal[-3, b"foo", True]`` is equivalent to
+``Union[Literal[-3], Literal[b"foo"], Literal[True]]``. This can help make writing
```
bluetech:
```suggestion
``Union[Literal[-3], Literal[b"foo"], Literal[True]]``. This makes writing
```

docs/source/more_types.rst
```diff
@@ -985,3 +989,165 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+    # Variables declared without annotations will continue to have an
+    # inferred type of 'bool'.
+
+    variable_2 = True
+    reveal_type(fetch_data(variable_2))  # Revealed type is 'Union[bytes, str]'
+
+Parameterizing Literals
+-----------------------
+
+Literal types may contain one or more literal bools, ints, strings, and byte
```
bluetech: If using "ints", maybe use "strs" and "bytes"? Or otherwise, use "integers".

docs/source/more_types.rst
```diff
@@ -985,3 +989,165 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+``Literal[my_instance]``, ``Literal[Any]``, ``Literal[3.14]``, or 
+``Literal[{"foo": 2, "bar": 5}]`` are all illegal.
+
+Future versions of mypy may relax some of these restrictions. For example, we
+plan on adding support for using enum values inside Literals in an upcoming release.
+
+Declaring Literal variables
+---------------------------
+
+You must explicitly add an annotation to a variable to declare that it is
```
bluetech:
```suggestion
You must explicitly add an annotation to a variable to declare that it has
```

docs/source/more_types.rst
```diff
@@ -985,3 +989,165 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+    def expects_literal(x: Literal[19]) -> None: pass
+
+    reveal_type(19)
+    expects_literal(19)
+
+This is why ``expects_literal(19)`` type-checks despite the fact that ``reveal_type(c)``
+reports ``int``.
+
+So while declaring a variable to be final is not quite the same thing as adding
+an explicit Literal annotation, it often leads to the same effect  in practice.
```
bluetech:
```suggestion
an explicit Literal annotation, it often leads to the same effect in practice.
```

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+.. _literal:
+
+Literal
+*******
+
+.. note::
+
+   Literal is an officially supported feature, but is highly experimental
+   and should be considered to be in alpha stage. It is very likely that future
+   releases of mypy will modify the behavior of Literal types, either by adding
```
ilevkivskyi:
```suggestion
   releases of mypy will modify the behavior of literal types, either by adding
```

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+   new features or by tuning or removing problematic ones.
+
+Literal types let you indicate that an expression is equal to some specific
+primitive value. For example, if we annotate a variable with type ``Literal["foo"]``,
+mypy will understand that variable is not only of type ``str``, but is also
+equal to specifically the string ``"foo"``.
+
+This feature is primarily useful when annotating functions that behave
+differently based on the exact value the caller provides. For example,
+suppose we have a function ``fetch_data(...)`` that returns bytes if the
```
ilevkivskyi:
```suggestion
suppose we have a function ``fetch_data(...)`` that returns ``bytes`` if the
```

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+first argument is True, and str if it's False. We can construct a precise
```
ilevkivskyi:
```suggestion
first argument is ``True``, and ``str`` if it's ``False``. We can construct a precise
```

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+
+    # The last overload is a fallback in case the caller
+    # provides a regular bool:
+
+    @overload
+    def fetch_data(raw: bool) -> Union[bytes, str]: ...
+
+    def fetch_data(raw: bool) -> Union[bytes, str]:
+        # (Implementation is omitted)
+        pass
```
ilevkivskyi:
```suggestion
        ...
```

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+    def fetch_data(raw: Literal[False]) -> str: ...
+
+    # The last overload is a fallback in case the caller
+    # provides a regular bool:
+
+    @overload
+    def fetch_data(raw: bool) -> Union[bytes, str]: ...
+
+    def fetch_data(raw: bool) -> Union[bytes, str]:
+        # (Implementation is omitted)
```
ilevkivskyi:
```suggestion
        # Implementation is omitted
```

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+        pass
+
+    variable_1: Literal[True] = True
```
ilevkivskyi: I remove this example, people are unlikely to use it, and a very similar one is discussed in the section below.

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+Literals may also contain aliases of Literal types. For example, the following program
+is legal:
+
+.. code-block:: python
+
+    PrimaryColors = Literal["red", "blue", "yellow"]
+    SecondaryColors = Literal["purple", "green", "orange"]
+    AllowedColors = Literal[PrimaryColors, SecondaryColors]
+
+    def paint(color: AllowedColors) -> None: pass
```
ilevkivskyi:
```suggestion
    def paint(color: AllowedColors) -> None: ...
```

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+In order to preserve backwards-compatibility, variables without this annotation
+are **not** assumed to be Literals:
+
+.. code-block:: python
+
+    b = 19
+    reveal_type(b)          # Revealed type is 'int'
+
+If you find repeating the value of the variable in the type hint to be tedious,
+you can instead declare the variable to be :ref:`Final <final_attrs>`:
```
ilevkivskyi: "Final variable" is an kind of oxymoron. I would rather say "final name", or reformulate this otherwise.

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+    from typing_extensions import Final, Literal
+
+    def expects_literal(x: Literal[19]) -> None: pass
+
+    c: Final = 19
+
+    reveal_type(c)          # Revealed type is 'int'
+    expects_literal(c)      # ...but this type checks!
+
+If we do not provide an explicit type in the Final, the type of ``c`` becomes
```
ilevkivskyi:
```suggestion
If you do not provide an explicit type in the ``Final``, the type of ``c`` becomes
```

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+
+    def expects_literal(x: Literal[19]) -> None: pass
+
+    reveal_type(19)
+    expects_literal(19)
+
+This is why ``expects_literal(19)`` type-checks despite the fact that ``reveal_type(c)``
+reports ``int``.
+
+So while declaring a variable to be Final is not quite the same thing as adding
```
ilevkivskyi: "Final variable" again.

docs/source/more_types.rst
```diff
@@ -985,3 +989,166 @@ and non-required keys, such as ``Movie`` above, will only be compatible with
...
+an explicit Literal annotation, it often leads to the same effect in practice.
```
ilevkivskyi:
```suggestion
an explicit ``Literal[...]`` annotation, it often leads to the same effect in practice.
```

docs/source/more_types.rst
Michael0x2a: I decided to rephrase this to say "you can instead change the variable to be Final" (and did a similar thing below).

docs/source/more_types.rst
Michael0x2a: Just to double-check, did you want me to lowercase every instance where I use "Literal" as a proper noun?

I capitalized this (and every other time I used the phrase "Literal types") because I wanted to refer specifically to the `Literal[...]` type itself. I can lowercase this if there's a good reason for it, but then we should probably lowercase that phrase every other time I use it.

ilevkivskyi: Ideally I would prefer to always use one of two: `literal` or ` ``Literal`` ` (i.e. if it is capitalized, then it is ` ``code`` `).

JukkaL:
Can you create a new top-level file for literal types? `more_types.rst` is kind of long, and I'd like to split it up at some point.

Michael0x2a:
@JukkaL and @ilevkivskyi -- FYI: I made the last few changes you suggested! (I moved the doc to the top level and changed the capitalization).

PR #6152, opened 2019-01-06
Add option to tag known messages with an error code

chadrik:
Adds `--show-error-codes` for tagging known error messages (and their notes) with error codes.  This is a stepping stone toward native message filtering (#2417), but it is useful on its own for users who want to implement their  own basic message filtering without relying on brittle regex against messages.   Also adds `--list-error-codes` to get a list of all known error codes. 

### Design goal

Provide a way of registering message ids that is:
  1. accessible from `main`, so that they can be listed with `--list-error-codes`
  2. concise
  3. able to handle the 3 types of error messages in a fairly consistent way:
    1. static messages, with no args to format:  currently have constants in `mypy.messages`
    2. messages with basic string args to format: currently have constants in `mypy.messages` (mostly)
    3. messages with more complex formatting:  currently have methods on `MessageBuilder`
  4. type safe for the code reporting the error, to avoid mistakes with the necessary arguments for the given message type

### Current Design

The design that I settled on is this:
- convert all messages to methods of `MessageBuilder`
- use a `@tracked` decorator to identify which methods emit messages which should be tracked
  - by default, the name of the method is the message id, but this can be overridden
  - multiple methods can have the same id, which places them all into the same group, which can be filtered together

### Questions
- naming:  I was a bit non-committal on the terminology to use: "message id", "error code"?  Internally I went with "message id", but the cli uses "error code"
- should there be a less verbose way of tracking constant error messages?  I created a descriptor-based prototype of this that I did not include, which allowed simple messages to be added to `MessageBuilder` like this:
  ```python
    no_return_exepected = ErrorReporter('No return value expected')
    invalid_implicit_return = ErrorReporter('Implicit return in function which does not return')
  ```
  this would be called like a normal `MessageBuilder` failure message:  `self.msg.invalid_implicit_return(ctx)`.  I was concerned it was a bit too magical. 

### What's next
This PR is not complete.  My idea is to finish converting message constants to `MessageBuilder` methods, as I already have with `no_return_exepected` and `must_have_none_return_type`, but  I wanted to get feedback on the design before doing this tedious work.

Once this is complete we can begin expanding tracked messages and adding native filtering of tracked messages.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -600,6 +600,9 @@ def process_options(args: List[str],
     add_invertible_flag('--show-error-codes', default=False,
                         help="Show error codes in error messages",
                         group=error_group)
+    error_group.add_argument(
+        '--list-error-codes', action='store_true',
+        help="List known error codes and exit")
 
     strict_help = "Strict mode; enables the following flags: {}".format(
         ", ".join(strict_flag_names))
@@ -701,6 +704,13 @@ def process_options(args: List[str],
     # filename for the config file and know if the user requested all strict options.
     dummy = argparse.Namespace()
     parser.parse_args(args, dummy)
+
+    if dummy.list_error_codes:
+        import mypy.messages
+        for msg_id in sorted(mypy.messages.message_ids):
+            print(msg_id)
+        raise SystemExit(0)
+
     config_file = dummy.config_file
     if config_file is not None and not os.path.exists(config_file):
         parser.error("Cannot find config file '%s'" % config_file)
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -70,6 +70,10 @@ def main(script_path: Optional[str], args: Optional[List[str]] = None) -> None:
     fscache = FileSystemCache()
     sources, options = process_options(args, fscache=fscache)
 
+    if sources is None:
+        # indicates early out
+        sys.exit(0)
+
     messages = []
 
     def flush_errors(new_messages: List[str], serious: bool) -> None:
@@ -299,7 +303,7 @@ def process_options(args: List[str],
                     fscache: Optional[FileSystemCache] = None,
                     program: str = 'mypy',
                     header: str = HEADER,
-                    ) -> Tuple[List[BuildSource], Options]:
+                    ) -> Tuple[Optional[List[BuildSource]], Options]:
     """Parse command line arguments.
 
     If a FileSystemCache is passed in, and package_root options are given,
@@ -704,19 +708,19 @@ def process_options(args: List[str],
     # filename for the config file and know if the user requested all strict options.
     dummy = argparse.Namespace()
     parser.parse_args(args, dummy)
+    options = Options()
 
     if dummy.list_error_codes:
         import mypy.messages
         for msg_id in sorted(mypy.messages.MessageBuilder.get_message_ids()):
             print(msg_id)
-        raise SystemExit(0)
+        return None, options
 
     config_file = dummy.config_file
     if config_file is not None and not os.path.exists(config_file):
         parser.error("Cannot find config file '%s'" % config_file)
 
     # Parse config file first, so command line can override.
-    options = Options()
     parse_config_file(options, config_file)
 
     # Set strict flags before parsing (if strict mode enabled), so other command
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -728,6 +728,7 @@ def process_options(args: List[str],
     # Parse command line for real, using a split namespace.
     special_opts = argparse.Namespace()
     parser.parse_args(args, SplitNamespace(options, special_opts, 'special-opts:'))
+    delattr(options, 'list_error_codes')
 
     # The python_version is either the default, which can be overridden via a config file,
     # or stored in special_opts and is passed via the command line.
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -707,7 +707,7 @@ def process_options(args: List[str],
 
     if dummy.list_error_codes:
         import mypy.messages
-        for msg_id in sorted(mypy.messages.message_ids):
+        for msg_id in sorted(mypy.messages.MessageBuilder.get_message_ids()):
             print(msg_id)
         raise SystemExit(0)
 
@@ -728,6 +728,8 @@ def process_options(args: List[str],
     # Parse command line for real, using a split namespace.
     special_opts = argparse.Namespace()
     parser.parse_args(args, SplitNamespace(options, special_opts, 'special-opts:'))
+
+    # unneeded attribute transfered from parser
     delattr(options, 'list_error_codes')
 
     # The python_version is either the default, which can be overridden via a config file,
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -597,6 +597,9 @@ def process_options(args: List[str],
     add_invertible_flag('--show-column-numbers', default=False,
                         help="Show column numbers in error messages",
                         group=error_group)
+    add_invertible_flag('--show-error-codes', default=False,
+                        help="Show error codes in error messages",
+                        group=error_group)
 
     strict_help = "Strict mode; enables the following flags: {}".format(
         ", ".join(strict_flag_names))
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -235,6 +235,9 @@ class MessageBuilder:
     def get_message_ids(cls) -> Set[str]:
         return _message_ids
 
+    def active_message_id(self) -> Optional[str]:
+        return self.active_msg_ids[-1] if len(self.active_msg_ids) else None
+
     #
     # Helpers
     #
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -16,7 +16,7 @@ import difflib
 from textwrap import dedent
 
 from typing import (
-    cast, List, Dict, Any, Sequence, Iterable, Tuple, Set, Optional, Union, ClassVar, TypeVar,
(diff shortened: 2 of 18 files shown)


mypy/messages.py
```diff
@@ -472,7 +517,16 @@ def format_distinctly(self, type1: Type, type2: Type, bare: bool = False) -> Tup
     # get some information as arguments, and they build an error message based
     # on them.
 
-    def has_no_attr(self, original_type: Type, typ: Type, member: str, context: Context) -> Type:
+    @tracked()
+    def no_return_exepected(self, context: Context) -> None:
+        self.fail('No return value expected', context)
```
ilevkivskyi: Converting all static messages to this format may cause lots of code churn, but in principle I am fine with this.

Another possible option is to have a global dictionary (or JSON file) mapping error message to error code:
```python
message_to_id = {
    NO_RETURN_VALUE_EXPECTED: 'unexpected_return',
    MUST_HAVE_NONE_RETURN_TYPE: 'unexpected_return',
    MISSING_RETURN_STATEMENT: 'missing_return',
    ...
}
```

ilevkivskyi:
> 2. messages with basic string args to format: currently have constants in `mypy.messages` (mostly)

Maybe we should fix the "(mostly)" part first? (I mean finish moving the error message constants to `mypy.messages`.)

mypy/main.py
```diff
@@ -595,6 +595,12 @@ def add_invertible_flag(flag: str,
     add_invertible_flag('--show-column-numbers', default=False,
                         help="Show column numbers in error messages",
                         group=error_group)
+    add_invertible_flag('--show-error-codes', default=False,
+                        help="Show error codes in error messages",
+                        group=error_group)
+    error_group.add_argument(
+        '--list-error-codes', action='store_true',
+        help="List known error codes and exit")
```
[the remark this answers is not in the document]
chadrik: I wanted to get tentative approval of the idea first.  Sounds like I have that, so I'll do that shortly.

mypy/messages.py
chadrik: > Another possible option is to have a global dictionary (or JSON file) mapping error message to error code

I think that could be a good compromise, but it doesn't work for messages that require formatting so we'd have to be very clear about this.  A developer might expect this to work with the `message_to_id` dict:

```python
self.fail(messages.MUST_HAVE_NONE_RETURN_TYPE.format(fdef.name()), item)
```

We can protect against this in a few ways:
- clear comments near `message_to_id` documenting that this is only for static messages
- code that checks `message_to_id` keys for formatting operations `{}`.  
- add a variation of `MessageBuilder.fail`, or a flag, that raises an exception if the message is not in `message_to_id` (I'm assuming here that we're not going to get to 100% message id coverage in the near term)

mypy/main.py
```diff
@@ -696,6 +702,13 @@ def add_invertible_flag(flag: str,
     # filename for the config file and know if the user requested all strict options.
     dummy = argparse.Namespace()
     parser.parse_args(args, dummy)
+
+    if dummy.list_error_codes:
+        import mypy.messages
+        for msg_id in sorted(mypy.messages.MessageBuilder.get_message_ids()):
+            print(msg_id)
+        raise SystemExit(0)
```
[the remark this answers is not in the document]
chadrik: I addressed this by making `List[BuildSource]` optional.

mypy/messages.py
```diff
@@ -174,6 +178,32 @@
 }  # type: Final
 
 
+_message_ids = set()  # type: Set[str]
+
+
+def tracked(id: Optional[str] = None) -> Callable[[T], T]:
```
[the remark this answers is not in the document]
chadrik: fixed

mypy/messages.py
chadrik: I just pushed a commit inspired your `message_to_id` idea, which addresses most of my concerns about type safety.  I used an `Enum`, where the names are ids and the values are the messages.   This is used for messages that are static or require only basic formatting (though I haven transitioned all of the code yet).  If you like it, I'll finish it up.

chadrik: @ilevkivskyi  Sorry to bug you.  Can I get a sign off on the idea of using `Enum`?  If that sounds like a good idea to you I think I can take this PR to completion.

ilevkivskyi: NP, I will think more about this tomorrow and let you know, I was really busy at work last two days.

ilevkivskyi: Eh, it looks like we can't use enums for two reasons:
* Mypyc compiler doesn't support custom metaclasses yet (`Enum` has a complex metaclass), there are plans to support it, but not in near few months.
* More importantly, plugin users should be able to produce custom error messages, which is hard with enums.

I thought more about possible strategies, and it looks like having a global dictionary for all messages is probably inevitable. I like your idea of changing the signature of `MessageBuilder.fail()` to accept "base" message and a list of format parameters. If possible, I would even put `format` before `context` and make it a required argument (caller would pass an empty tuple if the base message is just a constant), but probably we can do this later. Here are some things that I would propose:
* Focus on core logic rather on completeness in this PR
* Add a global dictionary `message_registry: Dict[str, str]` in `messages.py`, add only 3-5 keys to the dictionary in this PR.
* Only add `@register()` to 3-5 methods
* The `fail()` method will first try to get the error code from `message_registry` using base message as a key, if the base message is not there, try getting the active message (btw, why this needs to be a stack, not just a single value?), if no luck, use `'uncategorized_error'` (all user plugin messages will be here).
* In few subsequent PRs gradually add `@register()` to more methods, and more keys to `message_registry`.
* When we will get a bit more experience we might want to migrate messages either to `@register()` or to `message_registry`. But we will probably have both because both have pros and cons. For example static or super simple errors a better for latter, while there are some error messages constructed in a tricky manner, for those the former is better.

What do you think?

JukkaL:
I think that we should discuss how to represent error codes in abstract before looking at merging any code. This is a fairly big change, and it would be good to consider this from a few different viewpoints. 

Here are some questions to consider (some of these were raised by Ivan above):

* How are new static messages defined?
* How are new runtime constructed messages defined?
* What would the error message ids look like in mypy output or in the config file (give examples for a few common messages)?
* How would the message ids be shown in mypy output?
* At what granularity should the message ids be (i.e., should we group similar messages under one id)? 
* How well does this the approach work with mypyc? In particular, mypyc is bad at compiling certain Python idioms such as decorated methods and enums. We should perhaps either use those idioms sparingly or first improve how mypyc handles them. This can be important since there are many error messages, so whatever approach we choose it will probably have many instances in the code.
* How can user plugins define new error messages? 
* Do we need to do something to avoid conflicts between different plugins (e.g. both define a similar message)?
* Does the proposed approach slow down mypy startup? This is probably not an issue, but in case we construct some complex data structures this might plausibly have a measurable effect.

Iterating on a big PR is a lot of work, especially if we need to go through multiple iterations. So I'd suggest creating an issue that describes the current proposed approach in some detail and we can look for answers to the above questions. We can move on from there once we have a clearer picture of the tradeoffs.

mypy/messages.py
chadrik: Everything you've written makes sense.  The main problem is that I can't get over the notion that the `message_registry` dictionary is going to be awfully redundant and tedious to maintain, since it's mostly going to be the message constant followed by a lowercase string of the message constant's name (depending on how broad vs granular we are with grouping) :

```python
message_registry = {
    NO_RETURN_VALUE_EXPECTED: 'no_return_value_expected',
    MISSING_RETURN_STATEMENT: 'missing_return_statement',
    INVALID_IMPLICIT_RETURN: 'invalid_implicit_return',
    INCOMPATIBLE_RETURN_VALUE_TYPE: 'incompatible_return_type',
    RETURN_VALUE_EXPECTED: 'return_value_expected',
    NO_RETURN_EXPECTED: 'no_return_expected',
    INVALID_EXCEPTION: 'invalid_exception',
    INVALID_EXCEPTION_TYPE: 'invalid_exception',
    RETURN_IN_ASYNC_GENERATOR: 'return_in_async_generator',
}
```

I also don't like that the message and the id are stored in different locations.  I think making the constants into basic data classes would be an improvement:

```python
NO_RETURN_VALUE_EXPECTED = Msg('No return value expected', 'no_return_value_expected')  # type: Final
MISSING_RETURN_STATEMENT = Msg('Missing return statement', 'missing_return_statement')  # type: Final
```

But I'd really love to design a system where the constant doubles as the id, *unless* it's overridden by a group definition.  If enums won't work, the only other way I can think of to do this is with a descriptor.  Should I just give up on this idea?

JukkaL: > But I'd really love to design a system where the constant doubles as the id

Why not move all the constants into a dedicated module, and iterate over the module namespace and calculate the id based on the attribute name? This sort of metaprogramming should be okay with mypyc. Something like this:

```py
...
FOO_BAR = Msg('Foo bar')
MSG_B = Msg('Another message')

for name, val in list(globals().items()):
    if isinstance(val, Msg) and val.id is None:
        val.id = name.lower()
```

chadrik: > Gradual transition to the new scheme for existing error messages (I strongly prefer two-three smaller PRs, than one large)

How about I start with a new PR that moves all of the existing constants into `mypy.message_registry` with no other changes?  That seems to be one part of the design we've all come to agreement on.

ilevkivskyi: TBH, I don't see how this is important _now_. Anyway, we agreed that @JukkaL will follow on this and related PRs.

ilevkivskyi:
IIUC this is now replaced by https://github.com/python/mypy/pull/6472, so closing this PR in favour of the latter.

PR #6153, opened 2019-01-07
Fix dmypy run when no target given

ethanhs:
`dmypy run` without a target given will cause the daemon to die unexpectedly on calling `mypy.main.process_options` with improper arguments.

This catches the `SystemExit` the argument parser throws and captures `sys.stderr` for the output of the parser. I had to backport redirect_stderr since that is not available in 3.4, but the backport can be removed in a couple of months when we drop support for 3.4.

I'm not entirely sure if passing the way I'm passing the error messages is a good idea or not, but it certainly is better than dying with just `[WinError 233] No process is on the other end of the pipe`.

### DIFF
--- a/mypy/dmypy.py
+++ b/mypy/dmypy.py
@@ -1,6 +1,6 @@
 """Client for mypy daemon mode.
 
-Highly experimental!  Only supports UNIX-like systems.
+Experimental!
 
 This manages a daemon process which keeps useful state in memory
 rather than having to read it back from disk on each run.
--- a/mypy/dmypy.py
+++ b/mypy/dmypy.py
@@ -1,7 +1,5 @@
 """Client for mypy daemon mode.
 
-Experimental!
-
 This manages a daemon process which keeps useful state in memory
 rather than having to read it back from disk on each run.
 """
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -4,6 +4,7 @@ This implements a daemon process which keeps useful state in memory
 to enable fine-grained incremental reprocessing of changes.
 """
 
+import argparse
 import base64
 import io
 import json
@@ -278,7 +279,9 @@ class Server:
                     ['-i'] + list(args),
                     require_targets=True,
                     server_options=True,
-                    fscache=self.fscache)
+                    fscache=self.fscache,
+                    program='dmypy',
+                    header=argparse.SUPPRESS)
             # Signal that we need to restart if the options have changed
             if self.options_snapshot != options.snapshot():
                 return {'restart': 'configuration changed'}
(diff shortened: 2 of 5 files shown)


mypy/dmypy.py
```diff
@@ -1,6 +1,6 @@
 """Client for mypy daemon mode.
 
-Highly experimental!  Only supports UNIX-like systems.
+Experimental!
```
gvanrossum: Honestly I'd just drop this altogether.

mypy/util.py
```diff
@@ -255,3 +257,49 @@ def hard_exit(status: int = 0) -> None:
     sys.stdout.flush()
     sys.stderr.flush()
     os._exit(status)
+
+
+# The following is a backport of stream redirect utilities from Lib/contextlib.py
```
gvanrossum: Please also add a comment explaining this can be deleted once we stop supporting Python 3.4.

mypy/dmypy_server.py
```diff
@@ -288,6 +295,8 @@ def cmd_run(self, version: str, args: Sequence[str]) -> Dict[str, object]:
                     return {'restart': 'plugins changed'}
         except InvalidSourceList as err:
             return {'out': '', 'err': str(err), 'status': 2}
+        except SystemExit as e:
+            return {'out': '', 'err': stderr.getvalue(), 'status': e.code}
```
gvanrossum: I just stumbled upon another edge case: `dmypy run -- --help` will exit without any output. This is because argparse intercepts `--help` (and `-h` etc.) and prints the full help text to stdout and exits with zero status. I'm not sure what to do about this -- I suppose you could catch stdout too? Or ignore the issue?

mypy/dmypy_server.py
```diff
@@ -270,11 +271,17 @@ def cmd_stop(self) -> Dict[str, object]:
...
+            # Process options can exit on improper arguments, so we need to catch that and
+            # capture stderr so the client can report it
+            stderr = io.StringIO()
+            with redirect_stderr(stderr):
+                sources, options = mypy.main.process_options(
+                    ['-i'] + list(args),
+                    require_targets=True,
+                    server_options=True,
+                    fscache=self.fscache,
+                    program='dmypy',
```
gvanrossum: I'm not sure I agree with this -- it's technically not `dmypy` that cannot parse the argument. Maybe set this to `mypy-daemon` so it's clear that the error comes from the daemon process?

ethanhs: Yeah, good idea. I wanted to somehow signal it wasn't mypy complaining about this.

mypy/dmypy_server.py
ethanhs: It would be just another context manager to capture stdout (since I backported the contextlib stuff). I will play around with this. I should probably also not suppress the help header if we want people to be able to run help from `dmypy run`.

ethanhs:
>Just let's delete the cloned code ASAP.

This won't be a problem, I look forward to taking advantage of some 3.5 features in several places :)

PR #6160, opened 2019-01-08
Add the specified Python version and platform to junit.xml

gvanrossum:
This makes it easier to recognize which build has errors when errors from different versions or platforms may be displayed in the same UI.

(Alas, we can't do this in dmypy because its check command doesn't know the options passed to the start command.)

### DIFF
--- a/mypy/dmypy.py
+++ b/mypy/dmypy.py
@@ -356,7 +356,9 @@ def check_output(response: Dict[str, Any], verbose: bool,
         # Lazy import so this import doesn't slow things down when not writing junit
         from mypy.util import write_junit_xml
         messages = (out + err).splitlines()
-        write_junit_xml(response['roundtrip_time'], bool(err), messages, junit_xml)
+        # NOTE: We don't have access to the selected Python version and platform here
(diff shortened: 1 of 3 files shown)


mypy/dmypy.py
```diff
@@ -356,7 +356,9 @@ def check_output(response: Dict[str, Any], verbose: bool,
         # Lazy import so this import doesn't slow things down when not writing junit
         from mypy.util import write_junit_xml
         messages = (out + err).splitlines()
-        write_junit_xml(response['roundtrip_time'], bool(err), messages, junit_xml)
+        # NOTE: We don't have access to the selected Python version and platform here
+        write_junit_xml(response['roundtrip_time'], bool(err), messages, junit_xml,
+                        "mypy-daemon")
```
msullivan: If we think that collecting this info is important, we could pass it back in the response?

gvanrossum: We could, but for now it sounds like a lot of extra effort. Our internal builds don't direct users to the junit.xml file generated by the daemon.

PR #6166, opened 2019-01-08
Make _RedirectStream._stream a ClassVar to workaround mypyc bug

msullivan:
There is a mypyc bug in which if mypy thinks that an attribute is
"redeclared" in a subclass, mypyc emits it twice in the struct and
then chokes.

Make `_RedirectStream._stream` a `ClassVar` to avoid this.

(I need to add mypyc to our CI.)


msullivan:
It seems like a class variable to me?

Issue #6168, opened 2019-01-08
Inconsistency in Union/promotion interactions


msullivan:
In the following test
```
from typing import Union

class A:
    x: Union[int, float]

def foo(x: Union[int, float], y: A) -> None:
    reveal_type(x)
    reveal_type(y.x)
```
we emit:
```
lol.py:8: error: Revealed type is 'Union[builtins.int, builtins.float]'
lol.py:9: error: Revealed type is 'builtins.float'
```

That is to say, class attributes have unions simplified using promotions while arguments (and locals) do not.

Probably they ought to have the same behavior?
(And probably the behavior ought to be to *not* simplify, especially because of #6114)

ilevkivskyi:
> And probably the behavior ought to be to _not_ simplify

Yes, I think we should never use promotions in union simplification. Also this looks like something that may be easy to fix. We just need to use `ignore_promotions=True` in `UnionType.make_simplified_union()`.

ilevkivskyi:
Added usability label because this can cause confusion, see https://github.com/python/mypy/issues/7241.

hauntsaninja:
These currently have the same type

PR #6169, opened 2019-01-08
Fix mypy_mypyc crash caused by "complex" interaction

msullivan:
The type of Num.n is `complex`, though it can in practice also be an
`int` or a `float`, and we case on the actual type to determine which
sort of mypy ast node to use.

In #6114, promotions were disabled in `isinstance` checks, which
caused mypy to decide that the `isinstance` checks for int and float
won't match. Fix this by giving the value the type object instead of
complex.

Typeshed *does* actually annotate `n` as `Union[float, int, complex]`,
but mypy simplifies that into `complex`. (It does *not* do this
simplification for local variables with such an annotation; see
issue #6168.)

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -928,7 +928,11 @@ class ASTConverter:
 
     # Num(object n) -- a number as a PyObject.
     def visit_Num(self, n: ast3.Num) -> Union[IntExpr, FloatExpr, ComplexExpr]:
-        val = n.n
+        # The n field has the type complex, but complex isn't *really*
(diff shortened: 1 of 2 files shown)


msullivan:
(mypy_mypyc is currently completely busted until this goes in.)

msullivan:
Argh! There's still a failure left that causes some tests to fail.

I've also opened #6170 for running mypyc in CI.

(The way this snuck in was that I ran `runtests.py` on a compiled mypy, it passed the self check, failed linting, and then I forgot to re-run it after.)

PR #6175, opened 2019-01-10
enable isolated build (PEP-517 for tox)

gaborbernat:
https://tox.readthedocs.io/en/latest/example/package.html#packaging 👍

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,6 @@
+[build-system]
+requires = [
+    "setuptools >= 40.6.2",
+    "wheel >= 0.30.0",
+]
+build-backend = "setuptools.build_meta"
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,11 @@ jobs:
   - name: "trigger a build of wheels"
     python: 3.7
     script: if [[ ($TRAVIS_BRANCH = "master" || $TRAVIS_BRANCH =~ release-*) && $TRAVIS_PULL_REQUEST = "false" ]]; then ./misc/trigger_wheel_build.sh; fi
+  - name: "check dev environment builds"
+    python: 3.7
+      env:
+      - TOXENV=dev
+      - EXTRA_ARGS=
 
 install:
 - pip install -U pip setuptools
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,9 @@ if sys.version_info < (3, 4, 0):
     sys.stderr.write("ERROR: You need Python 3.4 or later to use mypy.\n")
     exit(1)
 
+# we'll import stuff from the source tree, let's ensure is on the sys path
+sys.path.insert(0, os.path.dirname(os.path.realpath(__file__)))
+
 # This requires setuptools when building; setuptools is not needed
 # when installing from a wheel file (though it is still neeeded for
 # alternative forms of installing, as suggested by README.md).
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-minversion = 2.9.1
+minversion = 3.4.0
 skip_missing_interpreters = true
 envlist = py34,
           py35,
@@ -8,12 +8,16 @@ envlist = py34,
           lint,
           type,
           docs
+isolated_build = true
 
 [testenv]
 description = run the test driver with {basepython}
 passenv = PYTEST_XDIST_WORKER_COUNT NEWSEMANAL
 deps = -rtest-requirements.txt
 commands = pytest {posargs}
+
+[testenv:.package]
+deps =
 
 [testenv:lint]
 description = check the code style


pyproject.toml
```diff
@@ -3,3 +3,5 @@ requires = [
     "setuptools >= 30.0.2",
     "wheel >= 0.29.0"
 ]
+
+build-backend = "setuptools.build_meta"
```
msullivan: Add a newline?

msullivan:
I don't have a super clear understanding of what this actually changes?

pyproject.toml
gaborbernat: ok

gaborbernat:
@msullivan you would need to read the PEP-517 and PEP-518 to understand clearly, but simply put ensures that the package build dependencies specified in ``pyproject.toml`` are satisfied before packaging the library.

gaborbernat:
Any chance of a merge on this?

gaborbernat:
Without this, the tox packaging build will silently ignore the pyproject.toml and use whatever the tox host environment has, so for users who have an old wheel/setuptools laying around the tox builds might possibly fail.

In parallel, although pip, for now, does default to using setuptools with PEP-517 if the key is not specified we should note that in the PEP-517 this default is not mandated, so it's not guaranteed to work with all build frontends (and pip itself can decide to not activate PEP-517 if this key is missing). This is in the spirit of explicit is better than implicit, in the true spirit of the Python zen. 

Long story short while this might not be a must have, it's definitely a should have.

gvanrossum:
I'm sorry, I still don't follow. If you're asking us to read and understand PEP 517/518, you will have to wait forever, so please bear with us and help us understand, and hopefully we'll feel comfortable that this PR solves a problem for more people than just you yourself.

For context, while I am an avid pip user, I don't understand much about what it does under the covers any more, and I don't understand at all what tox does or why. (At some point someone convinced us to use it in our .travis.yml and appyeyor.yml files, but I don't recall why.)

Given that context, can you explain what a "tox packaging build" is? I see our pyproject.toml but I don't know what it means -- like so much in the packaging world, it looks like pure cargo-cult to me: "don't ask questions, just put this there else 'things' won't work". Also, what is the "key"  you are referring to? And what is a "build frontend"?

Finally, don't go quoting the Zen of Python to me. Following rules without understanding was never meant to be part of it.

gaborbernat:
I'm sorry, my bad, I assumed you would be more familiar with the PEP-517/518. 
I'll write up a ``short``-ish summary of what the problems without PEP-517/518 and get back to you in a few days. Thanks for your patiance and undestanding!

gaborbernat:
Hello, thanks for being patient with me on this, I wrote a blog post explaining what happens more in detail: https://www.bernat.tech/pep-517-518/ Hope this clears up things. Thanks!

msullivan:
Hm, it seems like this will break pip install from the source directory, right?

gaborbernat:
Not if we ammend to prepend the sys path at the start of setup.py 😎

msullivan:
OK, can you update this PR to do that?

gaborbernat:
yeah will do 😊

gaborbernat:
@msullivan addressed, ready for review 👍

msullivan:
@gaborbernat I tried a `pip install .` from the mypy directory and it is still busted.
```
(asdf) msullivan@enki:~/src/mypy [isolated+]
$ pip install .
Processing /home/msullivan/src/mypy
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  Complete output from command /home/msullivan/src/mypy/asdf/bin/python3.7 /home/msullivan/src/mypy/asdf/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmp439qg31b:
    File "/home/msullivan/src/mypy/asdf/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 207, in <module>
      main()
    File "/home/msullivan/src/mypy/asdf/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 197, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/home/msullivan/src/mypy/asdf/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 54, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/tmp/pip-build-env-f4k7nrmc/overlay/lib/python3.7/site-packages/setuptools/build_meta.py", line 130, in get_requires_for_build_wheel
      return self._get_build_requires(config_settings, requirements=['wheel'])
    File "/tmp/pip-build-env-f4k7nrmc/overlay/lib/python3.7/site-packages/setuptools/build_meta.py", line 112, in _get_build_requires
      self.run_setup()
    File "/tmp/pip-build-env-f4k7nrmc/overlay/lib/python3.7/site-packages/setuptools/build_meta.py", line 126, in run_setup
      exec(compile(code, __file__, 'exec'), locals())
    File "setup.py", line 20, in <module>
      from mypy.version import __version__ as version
  ModuleNotFoundError: No module named 'mypy'
  
  ----------------------------------------
Command "/home/msullivan/src/mypy/asdf/bin/python3.7 /home/msullivan/src/mypy/asdf/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /tmp/tmp439qg31b" failed with error code 1 in /tmp/pip-req-build-d22a3xxy
```

gaborbernat:
@msullivan indeed, fixed now 👍

PR #6201, opened 2019-01-16
Refactor class semantic analysis

JukkaL:
Primary changes:

* Remove an unnecessary context manager.
* Split long function.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -826,6 +826,10 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
         self.leave_class()
 
     def analyze_namedtuple_classdef(self, defn: ClassDef) -> bool:
+        """Analyze class-based named tuple if the NamedTuple base class is present.
+
+        Return True only if the class is a NamedTuple class.
+        """
         named_tuple_info = self.named_tuple_analyzer.analyze_namedtuple_classdef(defn)
         if named_tuple_info is None:
             return False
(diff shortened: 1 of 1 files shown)


mypy/semanal.py
```diff
@@ -795,65 +795,69 @@ def check_function_signature(self, fdef: FuncItem) -> None:
...
+        for decorator in defn.decorators:
+            self.analyze_class_decorator(defn, decorator)
+        self.analyze_class_body_common(defn)
+
+    def analyze_class_body_common(self, defn: ClassDef) -> None:
+        """Parts of class body analysis that are common to all kinds of class defs."""
+        self.enter_class(defn.info)
+        defn.defs.accept(self)
+        self.calculate_abstract_status(defn.info)
+        self.setup_type_promotion(defn)
```
ilevkivskyi: This one is not really about class body.

mypy/semanal.py
```diff
@@ -795,65 +795,69 @@ def check_function_signature(self, fdef: FuncItem) -> None:
...
+        self.apply_class_plugin_hooks(defn)
+        self.leave_class()
+
+    def analyze_namedtuple_classdef(self, defn: ClassDef) -> bool:
+        """Analyze class-based named tuple if the NamedTuple base class is present.
+
+        Return True only if the class is a NamedTuple class.
```
ilevkivskyi: Maybe mention that this function is here instead of in `self.named_tuple_analyzer`, because named tuples can have user defined methods, and their bodies needs to be analyzed?

JukkaL: I instead added a TODO comment about moving this to live with the rest of the named tuple semantic analysis code.

PR #6203, opened 2019-01-16
 Refactoring: remove TVAR symbol table node kind 

JukkaL:


mypy/nodes.py
```diff
@@ -68,17 +68,12 @@ def get_column(self) -> int:
...
-# (1) a generic class that uses the type variable as a type argument or
-# (2) a generic function that refers to the type variable in its signature.
-TVAR = 3  # type: Final[int]
 
 # Placeholder for a name imported via 'from ... import'. Second phase of
 # semantic will replace this the actual imported reference. This is
 # needed so that we can detect whether a name has been imported during
 # XXX what?
-UNBOUND_IMPORTED = 5  # type: Final[int]
+UNBOUND_IMPORTED = 4  # type: Final[int]
```
ilevkivskyi: Shouldn't this be 3 now?

PR #6208, opened 2019-01-17
Change version in junit xml to have underscore

jhance:
To make the canonical junit name better


gvanrossum:
I think the typeshed sync doesn't belong in this PR. Otherwise LG.

jhance:
I'm not even sure where a typeshed sync could have came from, I wonder if I
got a bad merge base

gvanrossum:
I'll fix this by redoing it. See #6222.

PR #6213, opened 2019-01-17
Allow star args in ctypes.Array constructor

alanhdu:


mypy/plugins/ctypes.py
```diff
@@ -116,19 +117,27 @@ def array_constructor_callback(ctx: 'mypy.plugin.FunctionContext') -> Type:
...
-            # In a function hook, the type of an *args parameter is the type of the iterable being
-            # unpacked. However, FunctionContext currently doesn't provide a way to differentiate
-            # between normal arguments and *args, so the iterable type is considered invalid.
-            # Once FunctionContext has an API for this, *args should be allowed here if the
-            # iterable's element type is compatible with the array element type.
-            if not is_subtype(arg_type, allowed):
+        for arg_num, (arg_kind, arg_type) in enumerate(zip(ctx.arg_kinds[0], ctx.arg_types[0]), 1):
+            if arg_kind not in {nodes.ARG_POS, nodes.ARG_STAR}:
+                ctx.api.msg.fail(
+                    "Array constructor does not allow keyword arguments",
```
ilevkivskyi: Please add a test that triggers this error message.

mypy/plugins/ctypes.py
```diff
@@ -116,19 +117,27 @@ def array_constructor_callback(ctx: 'mypy.plugin.FunctionContext') -> Type:
...
                     'Array constructor argument {} of type "{}"'
                     ' is not convertible to the array element type "{}"'
                     .format(arg_num, arg_type, et),
                     ctx.context)
+            elif arg_kind == nodes.ARG_STAR:
+                ty = ctx.api.named_generic_type("typing.Iterable", [allowed])
+                if not is_subtype(arg_type, ty):
+                    ctx.api.msg.fail(
+                        'Array constructor argument {} of type "{}"'
+                        ' is not convertible to the array element type "Iterable[{}]"'
```
ilevkivskyi: Please also add a test for this error message.

mypy/plugins/ctypes.py
alanhdu: It looks like this actually isn't ever called -- it looks like the constructor in typeshed `def __init__(self, *args: Any) -> None)` definition takes care of this case.

test-data/unit/check-ctypes.test
```diff
@@ -167,4 +167,16 @@ reveal_type(intarr4(*int_values))  # E: Revealed type is 'ctypes.Array[ctypes.c_
...
 reveal_type(intarr6(1, ctypes.c_int(2), *c_int_values))  # E: Revealed type is 'ctypes.Array[ctypes.c_int]'
+
+float_values = [1.0, 2.0, 3.0, 4.0]
+intarr4(*float_values) # E: Array constructor argument 1 of type "builtins.list[builtins.float*]" is not convertible to the array element type "Iterable[ctypes.c_int]"
+[builtins fixtures/floatdict.pyi]
+
+[case testCtypesArrayConstructorKwargs]
+import ctypes
+intarr4 = ctypes.c_int * 4
+
```
alanhdu: I wanted to add a `intarr4(1, 2, 3, x=4)` test case here, but not sure how to get the tests to pass: the output on my machine is:

```
Expected:
  main:4: error: Unexpected keyword argument "x" for "Array"
  main:7: error: Too many arguments for "Array" (diff)
Actual:
  main:4: error: Unexpected keyword argument "x" for "Array"
  /home/alan/workspace/python/mypy/mypy/typeshed/stdlib/2and3/ctypes/__init__.pyi:273: note: "Array" defined here (diff)
  main:7: error: Too many arguments for "Array" (diff)
```

I added a `# E: ` comment for the error, but don't know how to handle the note (which includes the path, so will be setup specific). Thoughts?

ilevkivskyi: Oh, this note caused some problems before. One option would be to add some very simple mock stubs to `test-data/unit/lib-stub`. But probably even better is just not add this test. The two tests you have added are sufficient.

mypy/plugins/ctypes.py
ilevkivskyi: Oh, wait, so if it is never going to be called, maybe we can remove this code?

alanhdu: Done!

PR #6226, opened 2019-01-20
Expand "getting started" docs to discuss type hints in more detail

Michael0x2a:
This pull request expands our "getting started" docs so they discuss how to use type hints in more detail.

In general, I'm hoping this pull request will accomplish two main things:

1. Give users a better "tour" of the available PEP 484 types.

2. Do a better job of giving users a "heads-up" of common gotchas: for example, how dynamically typed functions aren't type-checked.

Here's a list of specific changes I made:

1. I combined "Installing mypy" and "Running mypy" into one section.

2. I modified the "Function signatures" section to focus more on the distinction between dynamic and static typing: I suspect that's a common tripup new users run into.

   I also changed the examples so they'll actually produce a runtime error if you misuse them.

3. I moved the existing text on annotating no-argument functions or functions with default arguments into a separate "More function signatures" to try and keep the new "Function signatures and dynamic vs static typing" section focused.

4. I expanded the section about using the "typing" module to be more thorough. In particular, I wanted to expose the user to the concept of concrete generic types (List), abstract generic types/protocols (Iterable), and to more non-conventional types like Union.

   I think these three examples are broadly representative of the different "flavors" of PEP 484 types and so should help new users be a little more aware of the kinds of things they can do with types.

   The main thing that's missing is an introduction to custom generic classes and functions: that's probably too complex of a topic to belong in this page.

5. I added a section on local type inference: it's a new concept for people who are more used to languages like Java or C# so felt like it was worth mentioning.

   This section also covers the other stumbling block I think new users are likely to run into: the "Need type annotation for 'variable'" error message.

6. I added a section on customizing mypy via the command line flags. I don't think enough users are aware that you can do this, despite how useful the flags can be.

### DIFF
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -49,18 +49,6 @@ to your code to take full advantage of mypy. See the section below for details.
 
       $ mypy --py2 program.py
 
-
-.. note::
-
-  Depending on how mypy is configured, you may have to run mypy like this:
-
-  .. code-block:: shell
-
-    $ python3 -m mypy program.py
-
-    # Or on Windows
-    $ py -3 -m mypy program.py
-
 Function signatures and dynamic vs static typing
 ************************************************
 
@@ -102,7 +90,7 @@ calls since the arguments have invalid types:
 Note that this is all still valid Python 3 code! The function annotation syntax
 shown above was added to Python `as a part of Python 3.0 <pep3107_>`_.
 
-If you are trying to type-check Python 2 code, you can add type hints
+If you are trying to type check Python 2 code, you can add type hints
 using a comment-based syntax instead of the Python 3 annotation syntax.
 See our section on :ref:`typing Python 2 code <python2>` for more details.
 
@@ -115,15 +103,15 @@ adding type hints to your code rather than adding them all at once. Similarly,
 when you are prototyping a new feature, it may be convenient to initially implement
 the code using dynamic typing and only add type hints later once the code is more stable.
 
-Once you are finished migrating or prototying your code, you can make mypy warn you
+Once you are finished migrating or prototyping your code, you can make mypy warn you
 if you add a dynamic function by mistake by using the ``--disallow-unchecked-defs``
 flag. See :ref:`command-line` for more information on configuring mypy.
 
 .. note::
 
-   The earlier stages of mypy, known as the semantic analysis, may
-   report errors even for dynamically typed functions. However, you
-   should not rely on this, as this may change in the future.
+   The earlier stages of analysis performed by mypy may report errors
+   even for dynamically typed functions. However, you should not rely
+   on this, as this may change in the future.
 
 More function signatures
 ************************
@@ -166,11 +154,13 @@ Arguments with default values can be annotated like so:
 
 .. code-block:: python
 
-   def stars(*names: str, **kwargs: float) -> float:
-       output = 0.0
+   def stars(*args: int, **kwargs: float) -> None:
+       # 'args' has type 'Tuple[int, ...]' (a tuple of ints)
+       # 'kwargs' has type 'Dict[str, float]' (a dict of strs to floats)
        for arg in args:
-           output += kwargs[arg]
-       return output
+           print(name)
+       for key, value in kwargs:
+           print(key, value)
 
 The typing module
 *****************
@@ -229,10 +219,26 @@ ints or strings, but no other types. We can express this using the ``Union`` typ
        else:
            return user_id
 
+Similarly, suppose that we want our function to accept only strings or None. We can
+again use ``Union`` and use ``Union[str, None]`` -- or alternatively, use the type
+``Optional[str]``. These two types are identical and interchangeable: ``Optional[str]``
+is just a shorthand or *alias* for ``Union[str, None]``. It exists mostly as a convenience
+to help our function signatures look a little cleaner:
+
+.. code-block:: python
+
+   from typing import Optional
+
+   def greeting(name: Optional[str] = None) -> str:
+       # Optional[str] means the same thing as Union[str, None]
+       if name is None:
+           name = 'stranger'
+       return 'Hello, ' + name
+
 The ``typing`` module contains many other useful types. You can find a
 quick overview by looking through the :ref:`mypy cheatsheets <overview-cheat-sheets>` 
 and a more detailed overview (including information on how to make your own
-generic types) by looking through the
+generic types or your own type aliases) by looking through the
 :ref:`type system reference <overview-type-system-reference>`.
 
 One final note: when adding types, the convention is to import types 
@@ -240,21 +246,24 @@ using the form ``from typing import Iterable`` (as opposed to doing
 just ``import typing`` or ``import typing as t`` or ``from typing import *``).
 
 For brevity, we often omit these ``typing`` imports in code examples, but
-mypy will give an error if you use definitions such as ``Iterable``
+mypy will give an error if you use types such as ``Iterable``
 without first importing them.
 
 Local type inference
 ********************
 
-Once you have added type hints to a function (e.g. made it statically typed),
+Once you have added type hints to a function (i.e. made it statically typed),
 mypy will automatically type check that function's body. While doing so,
 mypy will try and *infer* as many details as possible. 
 
 We saw an example of this in the ``normalize_id`` function above -- mypy understands
 basic ``isinstance`` checks and so can infer that the ``user_id`` variable was of
-type ``int`` in the if-branch and of type ``str`` in the else-branch.
+type ``int`` in the if-branch and of type ``str`` in the else-branch. Similarly, mypy
+was able to understand that ``name`` could not possibly be ``None`` in the ``greeting``
+function above, based both on the ``name is None`` check and the variable assignment
+in that if statement.
 
-As another example, consider the following function. Mypy can type-check this function
+As another example, consider the following function. Mypy can type check this function
 without a problem: it will use the available context and deduce that ``output`` must be
 of type ``List[float]`` and that ``num`` must be of type ``float``:
 
@@ -280,7 +289,8 @@ of ints to floats, we could annotate it like so:
 
 .. code-block:: python
 
-   # If you are using Python 3.6+, you can use *variable annotations*:
+   # If you are using Python 3.6+, you can use variable annotations.
+   # See https://www.python.org/dev/peps/pep-0526/ for more on variable annotations.
    my_global_dict: Dict[int, float] = {}
 
    # If you are using older versions of Python, you can use the comment-based syntax:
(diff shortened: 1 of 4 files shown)


docs/source/getting_started.rst
```diff
@@ -4,84 +4,132 @@ Getting started
...
 .. code-block:: python
 
    def greeting(name: str) -> str:
-       return 'Hello, {}'.format(name)
+       return 'Hello ' + name
 
-Now mypy will reject the following call, since the argument has an
-incompatible type:
+Our function is now *statically typed*: mypy can use the provided type hints to detect
+incorrect ussages of the ``greeting`` function. For example, it will reject the following
```
bluetech:
```suggestion
incorrect usages of the ``greeting`` function. For example, it will reject the following
```

docs/source/getting_started.rst
```diff
@@ -4,84 +4,132 @@ Getting started
...
+If you are trying to type-check Python 2 code, you can add type hints
+using a comment-based syntax instead of the Python 3 annotation syntax.
+See our section on :ref:`typing Python 2 code <python2>` for more details.
+
+.. _pep3107: https://www.python.org/dev/peps/pep-3107/
+
+Being able to pick whether you want a function to be dynamically or statically
+typed can be very helpful. For example, if you are migrating an existing
+Python codebase to use static types, it's usually easier to migrate by incrementally
+adding type hints to your code rather then adding them all at once. Similarly,
```
bluetech:
```suggestion
adding type hints to your code rather than adding them all at once. Similarly,
```

docs/source/getting_started.rst
```diff
@@ -103,47 +162,130 @@ Arguments with default values can be annotated as follows:
...
+       else:
+           return user_id
+
+The ``typing`` module contains many other useful types. You can find a
+quick overview by looking through the :ref:`mypy cheatsheets <overview-cheat-sheets>` 
+and a more detailed overview (including information on how to make your own
+generic types) by looking through the
+:ref:`type system reference <overview-type-system-reference>`.
+
+One final note: when adding types, the convention is to import types
```
bluetech: Is it also fair to say that it is conventional to put `typing` imports as the first imports? (It seems like it to me, at least).

If it's not actually a common convention, then it shouldn't derail this PR, but if it is I think it will be nice to mention.

docs/source/getting_started.rst
```diff
@@ -103,47 +162,130 @@ Arguments with default values can be annotated as follows:
...
+.. code-block:: python
+
+   def nums_below(numbers: Iterable[float], limit: float) -> List[float]:
+       output = []
+       for num in numbers:
+           if num < limit:
+               output.append(num)
+       return output
+
+Mypy will warn you if it is unable to determine what the type of some variable is.
```
bluetech:
```suggestion
Mypy will warn you if it is unable to determine the type of some variable.
```

docs/source/getting_started.rst
Michael0x2a: I also personally like that style (and I think mypy itself also follows that convention), but I'm thinking it's probably best to not mention this for now -- I'm not sure if other Python linters have caught up on this convention yet. For example, I think isort will sort imports alphabetically?

docs/source/getting_started.rst
```diff
@@ -103,47 +162,130 @@ Arguments with default values can be annotated as follows:
...
+    my_global_dict = {}  # Error: Need type annotation for 'my_global_dict'
+
+We can teach mypy what type ``my_global_dict`` is meant to have by giving it
+a type hint. For example, if we knew this variable is supposed to be a dict
+of ints to floats, we could annotate it like so:
+
+.. code-block:: python
+
+   # If you are using Python 3.6+, you can use *variable annotations*:
+   my_global_dict: Dict[int, float] = {}
```
ilevkivskyi: You can also add a link to PEP 526, like you did above for function annotations.

docs/source/getting_started.rst
ilevkivskyi: I don't have a strong preference, but I also like `typing` imports first. `isort` however has a different opinion (maybe it is configurable, I have very little experience with it).

docs/source/getting_started.rst
```diff
@@ -4,84 +4,132 @@ Getting started
...
-       return 'Hello, {}'.format(name)
+       return 'Hello ' + name
 
+   greeting(3)         # Argument 1 to "greeting" has incompatible type "int"; expected "str"
    greeting(b'Alice')  # Argument 1 to "greeting" has incompatible type "bytes"; expected "str"
 
+Note that this is all still valid Python 3 code! The function annotation syntax
+shown above was added to Python `as a part of Python 3.0 <pep3107_>`_.
+
+If you are trying to type-check Python 2 code, you can add type hints
```
JukkaL: Style nit: Above you wrote "type check" without a dash. I prefer to write it as "type check".

docs/source/getting_started.rst
```diff
@@ -4,84 +4,132 @@ Getting started
...
+.. _pep3107: https://www.python.org/dev/peps/pep-3107/
+
+Being able to pick whether you want a function to be dynamically or statically
+typed can be very helpful. For example, if you are migrating an existing
+Python codebase to use static types, it's usually easier to migrate by incrementally
+adding type hints to your code rather than adding them all at once. Similarly,
+when you are prototyping a new feature, it may be convenient to initially implement
+the code using dynamic typing and only add type hints later once the code is more stable.
+
+Once you are finished migrating or prototying your code, you can make mypy warn you
```
JukkaL: Typo: prototying

docs/source/getting_started.rst
```diff
@@ -4,84 +4,132 @@ Getting started
...
+if you add a dynamic function by mistake by using the ``--disallow-unchecked-defs``
+flag. See :ref:`command-line` for more information on configuring mypy.
+
+.. note::
+
+   The earlier stages of mypy, known as the semantic analysis, may
```
JukkaL: Maybe write as "The earlier stages of analysis performed by mypy" or similar.

docs/source/getting_started.rst
```diff
@@ -103,47 +162,130 @@ Arguments with default values can be annotated as follows:
...
-feature, you may decide to first implement the relevant code using
-dynamic typing and only add type signatures later, when the code is
-more stable.
+``*args`` and ``**kwargs`` arguments can be annotated like so:
 
 .. code-block:: python
 
-   def f():
-       1 + 'x'  # No static type error (dynamically typed)
+   def stars(*names: str, **kwargs: float) -> float:
```
JukkaL: I think that this example is a bit too clever. Maybe instead write too simpler (unrelated) functions, one for demonstrating `*args` and the other for `**kwargs`.

docs/source/getting_started.rst
JukkaL: I don't think that this convention is followed very widely. The `mypy` implementation doesn't consistently put `typing` imports first. I would rather not recommend any particular placement for the imports.

docs/source/getting_started.rst
```diff
@@ -103,47 +162,130 @@ Arguments with default values can be annotated as follows:
...
+just ``import typing`` or ``import typing as t`` or ``from typing import *``).
+
+For brevity, we often omit these ``typing`` imports in code examples, but
 mypy will give an error if you use definitions such as ``Iterable``
 without first importing them.
 
+Local type inference
+********************
+
+Once you have added type hints to a function (e.g. made it statically typed),
```
JukkaL: Should this be "(i.e., made it statically typed)"? "E.g." means "for example".

docs/source/getting_started.rst
```diff
@@ -103,47 +162,130 @@ Arguments with default values can be annotated as follows:
...
+
+    my_global_dict = {}  # Error: Need type annotation for 'my_global_dict'
+
+We can teach mypy what type ``my_global_dict`` is meant to have by giving it
+a type hint. For example, if we knew this variable is supposed to be a dict
+of ints to floats, we could annotate it like so:
+
+.. code-block:: python
+
+   # If you are using Python 3.6+, you can use *variable annotations*:
```
JukkaL: Does `*text*` work in code blocks? If it is included verbatim, maybe leave the emphasis out.

docs/source/getting_started.rst
```diff
@@ -171,7 +313,28 @@ library module that you import. Some modules ship with stubs that mypy
...
+
+Mypy supports many command line options that you can use to tweak how
+mypy behaves.  They are documented in :ref:`command-line`.
+
+For example, suppose you want to make sure *all* functions within your
+codebase are using static typing and make mypy report an error if you
+add a dynamically-typed function by mistake. You can make mypy do this
+by running mypy with the ``--disallow-untyped-defs`` flag.
+
+One particularly useful command is the ``--strict`` flag. Running
```
JukkaL: I'm not sure if we should recommend `--strict` here. Even mypy doesn't use all the options it enables yet. At least I think that there is too much emphasis on it now. I think it might be okay to mention it here as option for people who want to go the extra mile.

JukkaL:
Two more things: 
* `Optional[x]` is very common -- more common than regular union types. Maybe at least add a quick mention?
* Maybe also quickly introduce `Tuple[...]`, since tuple types are common.

Michael0x2a:
@ilevkivskyi and @JukkaL -- ok, this should be ready for a second look whenever!

@JukkaL -- I added in a section about Optional types, as suggested. I also tried adding in some discussion of Tuple, but everything I tried adding felt a little bloated. I think as a compromise, what I'll do is submit a follow-up diff where I expand the "built-in types" page and have the "getting started" page directly link to that.

docs/source/getting_started.rst
```diff
@@ -103,47 +150,152 @@ Arguments with default values can be annotated as follows:
...
+    my_global_dict = {}  # Error: Need type annotation for 'my_global_dict'
+
+We can teach mypy what type ``my_global_dict`` is meant to have by giving it
+a type hint. For example, if we knew this variable is supposed to be a dict
+of ints to floats, we could annotate it like so:
+
+.. code-block:: python
+
+   # If you are using Python 3.6+, you can use variable annotations.
+   # See https://www.python.org/dev/peps/pep-0526/ for more on variable annotations.
```
ilevkivskyi: I think a clickable link would look better. For example, ``...we could `annotate <pep526_>`_ it...`` in the text above. Or maybe just abandon this idea, in any case non-clickable links look bad in the docs.

Also I thought most of the documentation uses "_you_ could..." style, but you have couple of "we" in that sentence above.

docs/source/getting_started.rst
```diff
@@ -103,47 +150,152 @@ Arguments with default values can be annotated as follows:
...
 .. code-block:: python
 
    from typing import Iterable
 
    def greet_all(names: Iterable[str]) -> None:
        for name in names:
-           print('Hello, {}'.format(name))
+           print('Hello ' + name)
+
+As a final example, suppose want to write a function that can accept *either*
```
JukkaL: Nit: this is no longer the final example.

PR #6237, opened 2019-01-22
Enable strictness flags that require no source code changes

Michael0x2a:
This diff tightens up mypy's self-check slightly by enabling some of the `--strict` flags that mypy is currently already passing.

I also thought about enabling `--warn-unused-ignores` since there were only four places where we need to delete a `# type: ignore` comment, but ultimately decided against it: I don't want to complicate typeshed contributions (since typeshed's tests run mypy's self-check) and because one of those ignores looked like it was related to weird mypyc shenanigans (the lxml import in mypy/report.py).

I did go ahead and remove the other three `# type: ignores` though.

### DIFF
--- a/mypy_bootstrap.ini
+++ b/mypy_bootstrap.ini
@@ -1,5 +1,8 @@
 [mypy]
+disallow_untyped_calls = True
 disallow_untyped_defs = True
+disallow_incomplete_defs = True
+check_untyped_defs = True
 disallow_subclassing_any = True
 warn_no_return = True
 strict_optional = True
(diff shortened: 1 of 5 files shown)


mypy_bootstrap.ini
```diff
@@ -15,3 +18,4 @@ platform = mypyc
 # needs py2 compatibility
 [mypy-mypy.test.testextensions]
 disallow_untyped_defs = False
+check_untyped_defs = False
```
ilevkivskyi: Wait, why do we need these tests here? I though `mypy_extensions` were moved to a separate repo, but it looks like the tests were left behind @msullivan will you have time to move tests and setup Travis in the new repo?

msullivan: Yeah this is because I had forgotten about this and assumed we just didn't have any tests for it. I can handle this.

Issue #6204, opened 2019-01-16
Semantic analyzer redesign


JukkaL:
We are planning to do a major update to how the semantic analyzer works. The main idea is to use "deferred nodes" to allow each AST node to be processed an arbitrary number of times until all dependencies are ready. This may be needed to bind names within import cycles, for example. This should allow us to fix various issues with forward references and import cycles. We also hope that it will make it relatively easy to support general recursive types.

While the new design is in development, we'll have two semantic analyzer implementations available. We'll replace the existing semantic analyzer with the new one once the latter works well enough. Before that the new semantic analyzer must be explicitly enabled with a command-line option.

The new design should be strictly better than the current one, but there will likely some backward compatibility issues with existing plugins.

ilevkivskyi:
Another thing (reminded by #3459) that we should do soon is write more docs about the deferral/completeness logic. Some details are missing in the docstrings, it is better not to postpone this until after we make new analyzer default.

JukkaL:
We've now officially released the new semantic analyzer as an experimental feature, enabled through `--new-semantic-analyzer`. There are still a few regressions, some of which may take a while to fix, but we can track those as separate issues.

PR #6240, opened 2019-01-22
Initial work towards new semantic analyzer

JukkaL:
Implement basic deferred nodes functionality and remove most of old 
pass 1 functionality. Also fork `mypy.semanal_shared` since I needed
to tweak the signature of `anal_type` in an incompatible way (there are
not other changes in that module).

A *lot* of things still don't work, including functions and more than two
passes. There are a small number of tests for the new semantic 
analyzer, but I made no attempt at achieving good test coverage yet.

Work towards #6204.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -1147,6 +1147,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 # Special case 'builtins.object'. It was alredy created, just patch the
                 # actual ClassDef object.
                 info = self.globals['object'].node
+                assert isinstance(info, TypeInfo)
                 defn.info = info
                 info.defn = defn
             else:
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -301,6 +301,19 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             del self.cur_mod_node
             del self.globals
 
+    def prepare_file(self, file_node: MypyFile) -> None:
+        file_node.names = SymbolTable()
+        if 'builtins' in self.modules:
+            file_node.names['__builtins__'] = SymbolTableNode(GDEF,
+                                                              self.modules['builtins'])
+        if file_node.fullname() == 'builtins':
+            # Add empty core definitions required for basic operation. These fill be completed
+            # later on.
+            cdef = ClassDef('object', Block([]))  # Dummy ClassDef, will be replaced
+            info = TypeInfo(SymbolTable(), cdef, 'builtins')
+            info._fullname = 'builtins.object'
+            file_node.names['object'] = SymbolTableNode(GDEF, info)
+
     def refresh_partial(self, node: Union[MypyFile, FuncDef, OverloadedFuncDef],
                         patches: List[Tuple[int, Callable[[], None]]]) -> None:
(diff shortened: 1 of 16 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -1726,13 +1765,23 @@ def anal_type(self, t: Type, *,
...
+        and return AnyType.
+        """
         a = self.type_analyzer(tvar_scope=tvar_scope,
                                allow_unbound_tvars=allow_unbound_tvars,
                                allow_tuple_literal=allow_tuple_literal,
                                report_invalid_types=report_invalid_types,
                                third_pass=third_pass)
+        prev_incomplete = self.num_incomplete_refs
         typ = t.accept(a)
+        if self.num_incomplete_refs != prev_incomplete:
```
ilevkivskyi: This way looks a bit tricky, maybe we can get this somehow from the return type (or use an exception)? This is fine for a follow-up PR if you agree.

mypy/newsemanal/semanal.py
```diff
@@ -2531,7 +2575,12 @@ def analyze_types(self, items: List[Expression]) -> List[Type]:
         result = []  # type: List[Type]
         for node in items:
             try:
-                result.append(self.anal_type(expr_to_unanalyzed_type(node)))
+                analyzed = self.anal_type(expr_to_unanalyzed_type(node))
+                if analyzed is not None:
+                    result.append(analyzed)
+                else:
+                    # TODO: Is this the right thing to do?
+                    result.append(AnyType(TypeOfAny.from_error))
```
ilevkivskyi: Should we just return `Optional[List[Type]]` from this function instead?

mypy/newsemanal/semanal_main.py
```diff
@@ -0,0 +1,79 @@
...
+    # Process functions.
+    deferred = []  # type: List[str]
+    for id in scc:
+        tree = graph[id].tree
+        assert tree is not None
+        symtable = tree.names
+        targets = get_all_leaf_targets(symtable)
+        for target in targets:
+            deferred += semantic_analyze_target(target, graph[id])
+    assert not deferred  # There can't be cross-function forward refs
```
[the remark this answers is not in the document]
JukkaL: It's not a limitation as such, since deferring functions should never be necessary. All namespaces need to complete before start processing functions.

test-data/unit/check-newsemanal.test
```diff
@@ -0,0 +1,85 @@
+-- Test cases for the new semantic analyzer
```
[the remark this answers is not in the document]
JukkaL: Yeah, we can start thinking of this once basic stuff works.

mypy/newsemanal/semanal.py
JukkaL: Using return type may be tricky, since this needs to also work with the main AST visitor, which doesn't return anything. Using an exception may be a good idea. Let's revisit this later once we have a more complete implementation.

mypy/newsemanal/semanal.py
JukkaL: Maybe, it's unclear at this point. We can get back to this later.

JukkaL:
I'm going to merge this as-is since I have some other work almost ready that is built on top of this. I'll address some of the comments in another PR, and for others we can create follow-up tasks.

PR #6242, opened 2019-01-23
New semantic analyzer: Basic function support

JukkaL:
Functions are analyzed after all the module top levels are ready.

Also add some missing definitions to builtins such as `None`.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -315,12 +315,34 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             file_node.names['__builtins__'] = SymbolTableNode(GDEF,
                                                               self.modules['builtins'])
         if file_node.fullname() == 'builtins':
-            # Add empty core definitions required for basic operation. These fill be completed
-            # later on.
-            cdef = ClassDef('object', Block([]))  # Dummy ClassDef, will be replaced later
-            info = TypeInfo(SymbolTable(), cdef, 'builtins')
-            info._fullname = 'builtins.object'
-            file_node.names['object'] = SymbolTableNode(GDEF, info)
+            self.prepare_builtins_namespace(file_node)
+
+    def prepare_builtins_namespace(self, file_node: MypyFile) -> None:
+        names = file_node.names
+
+        # Add empty definition for 'object', since it's required for basic operation.
+        # This will be completed later on.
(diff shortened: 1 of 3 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -381,12 +403,30 @@ def file_context(self, file_node: MypyFile, fnam: str, options: Options,
...
+        # TODO: Module-level __getattr__
+        # TODO: Conditional function definitions
+        if self.is_module_scope():
+            if func.name() not in self.globals:
+                self.globals[func.name()] = SymbolTableNode(GDEF, func)
+        elif self.is_class_scope():
+            info = self.type
+            assert info is not None
+            if func.name() not in info.names:
+                info.names[func.name()] = SymbolTableNode(MDEF, func)
```
ilevkivskyi: I think we are making the same mistake again here. We already have several functions and code fragments for adding symbols to symbol table. I would rather have _one_ method that adds symbol node to current namespace. It can have some additional flags and can be called from all similar places like this.

mypy/newsemanal/semanal_main.py
```diff
@@ -44,22 +48,41 @@ def process_functions(graph: 'Graph', scc: List[str]) -> None:
...
+        tree = graph[module].tree
         assert tree is not None
         symtable = tree.names
-        targets = get_all_leaf_targets(symtable)
-        for target in targets:
-            deferred += semantic_analyze_target(target, graph[id])
+        targets = get_all_leaf_targets(symtable, module, None)
+        for target, node, active_type in targets:
+            deferred += semantic_analyze_target(target, graph[module], node, active_type)
     assert not deferred  # There can't be cross-function forward refs
```
ilevkivskyi: But there can be forward references within functions, for example:
```python
def outer():
    def inner():
        print(x)
    x = 1
    inner()
```
In some sense this is similar to the `global x` problem I mentioned before.

mypy/newsemanal/semanal.py
JukkaL: Yes, this will need to be refactored. This is just a placeholder implementation.

mypy/newsemanal/semanal_main.py
JukkaL: I think that this may be handled by another deferral mechanism inside the semantic analyzer, but I can't remember how these work exactly. Added an item about this to our task list.

PR #6246, opened 2019-01-24
New semantic analyzer: Various small updates

JukkaL:
These respond to some review feedback to
https://github.com/python/mypy/pull/6240.

### DIFF
--- a/mypy/newsemanal/semanal_main.py
+++ b/mypy/newsemanal/semanal_main.py
@@ -16,10 +16,13 @@ MAX_ITERATIONS = 10
 
 
 def semantic_analysis_for_scc(graph: 'Graph', scc: List[str]) -> None:
-    # Assume reachability analysis has already been performed.
-    #
-    # Note that functions can't define new module-level attributes,
-    # since module top levels are fully processed before functions.
+    """Perform semantic analysis for all modules in a SCC (import cycle).
+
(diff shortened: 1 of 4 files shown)


mypy/newsemanal/semanal_main.py
```diff
@@ -17,6 +17,9 @@
 
 def semantic_analysis_for_scc(graph: 'Graph', scc: List[str]) -> None:
     # Assume reachability analysis has already been performed.
+    #
+    # Note that functions can't define new module-level attributes,
+    # since module top levels are fully processed before functions.
```
ilevkivskyi: Sorry for being annoying, but maybe explicitly mention `global x` here? (and clarify whether it is a temporary or principal limitation?)

test-data/unit/check-newsemanal.test
```diff
@@ -80,6 +80,14 @@ from b import bad # E: Module 'b' has no attribute 'bad'
 [file b.py]
 from a import bad2 # E: Module 'a' has no attribute 'bad2'
 
+[case testNewAnalyzerTypeAnnotationCycle4]
+# flags: --new-semantic-analyzer
+import b
+[file a.py]
+from b import bad # E: Module 'b' has no attribute 'bad'
```
ilevkivskyi: I would prefer an error in both files, but at least one is also OK.

PR #6247, opened 2019-01-24
 New semantic analyzer: Refactor deferred/incompleteness handling

JukkaL:
The implementation of deferred/incompleteness state remains
the same but now we have a cleaner API for accessing it.


JukkaL:
The appveyor failure looks like a flake.

PR #6255, opened 2019-01-25
New semantic analyzer: Support class-based TypedDicts

JukkaL:
To support deferrals, we only create a TypeInfo once there
are no incomplete references within the definition. This
required some restructuring.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -1248,7 +1248,11 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         return remove_dups(tvars)
 
     def prepare_class_def(self, defn: ClassDef, info: Optional[TypeInfo] = None) -> None:
-        """Prepare for the analysis of a class definition."""
+        """Prepare for the analysis of a class definition.
+
+        Create an empty TypeInfo and store it in a symbol table, or if the 'info'
+        argument is provided, store it instead (used for magic type definitions).
+        """
         if not defn.info:
             defn.fullname = self.qualified_name(defn.name)
             # TODO: Nested classes
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -895,8 +895,16 @@ class NewSemanticAnalyzer(NodeVisitor[None],
 
         base_types, base_error = result
 
+        is_typeddict, info = self.typed_dict_analyzer.analyze_typeddict_classdef(defn)
+        if is_typeddict:
+            if info is None:
+                self.mark_incomplete()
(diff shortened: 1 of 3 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -1241,24 +1247,14 @@ def get_all_bases_tvars(self,
                 tvars.extend(base_tvars)
         return remove_dups(tvars)
 
-    def setup_class_def_analysis(self, defn: ClassDef) -> None:
+    def prepare_class_def(self, defn: ClassDef, info: Optional[TypeInfo] = None) -> None:
```
ilevkivskyi: Documenting this argument would help. (Maybe just explain that some infos are generated by magic defs, so we should use those).

mypy/newsemanal/semanal.py
```diff
@@ -1285,6 +1281,19 @@ def setup_class_def_analysis(self, defn: ClassDef) -> None:
...
 
+    def make_empty_type_info(self, defn: ClassDef) -> TypeInfo:
+        if (self.is_module_scope()
+                and self.cur_mod_id == 'builtins'
+                and defn.name in CORE_BUILTIN_CLASSES):
+            # Special case core built-in classes. A TypeInfo was already
+            # created for it before semantic analysis, but with a dummy
+            # ClassDef. Patch the real ClassDef object.
+            info = self.globals[defn.name].node
+            assert isinstance(info, TypeInfo)
```
ilevkivskyi: Nice idea to move all the special casing to a separate method.

mypy/newsemanal/semanal_typeddict.py
```diff
@@ -45,64 +49,72 @@ def analyze_typeddict_classdef(self, defn: ClassDef) -> bool:
...
-                defn.info.replaced = info
-                defn.info = info
-                node.node = info
+            if (len(defn.base_type_exprs) == 1 and
+                    isinstance(defn.base_type_exprs[0], RefExpr) and
+                    defn.base_type_exprs[0].fullname == 'mypy_extensions.TypedDict'):
+                # Building a new TypedDict
+                fields, types, required_keys = self.analyze_typeddict_classdef_fields(defn)
+                if fields is None:
+                    return True, None  # Defer
```
ilevkivskyi: This logic is different from how classes work. For normal classes we only defer creation of `TypeInfo` if there are unresolved bases. But here we also defer it ob unresolved refs in the "class body". I understand that this is because class syntax for TypedDicts is just syntax, and this logic is right. Nevertheless, I think it makes sense to explicitly document this somewhere.

PR #6256, opened 2019-01-25
Improve CLI, refactor and document stubgen

ilevkivskyi:
This PR does several "infrastructure" changes to `stubgen` tool:
* Update CLI and source discovery/collection to match the `mypy` CLI
* Refactor the logic in `stubgen.main()` into independent functions
* Use semantically analyzed Python ASTs
* Separate inference of signatures from docs to a separate module `stubdoc.py`
* Move some functions from `stubgen.py` to `stubutil.py` and reorganize the latter
* Clean-up the test runner to make more use of `DataSuite`
* Add documentation for the `stubgen` script

This also does few smaller things:
* Minimize number of repeated hard-coded constants
* Removed duplicate TODOs
* Added dosctrings to `stubgenc.py` and (new) `stubdoc.py`
* Avoid mutable defaults

This is not a pure refactoring, turning the semantic analysis on required some (although relatively small) changes in logic (because the sources should be semantically analyzed as a whole). It also required couple minor changes in `semanal.py` and `build.py`.

### DIFF
--- a/mypy/test/teststubgen.py
+++ b/mypy/test/teststubgen.py
@@ -225,8 +225,8 @@ class StubgenUtilSuite(Suite):
         assert_equal(
             infer_arg_sig_from_docstring(
                 "(x: Tuple[int, Tuple[str, int], str]=(1, ('a', 2), 'y'), y: int=4)"),
-                [ArgSig(name='x', type='Tuple[int,Tuple[str,int],str]', default=True),
-                 ArgSig(name='y', type='int', default=True)])
+            [ArgSig(name='x', type='Tuple[int,Tuple[str,int],str]', default=True),
+             ArgSig(name='y', type='int', default=True)])
 
     def test_infer_prop_type_from_docstring(self) -> None:
         assert_equal(infer_prop_type_from_docstring('str: A string.'), 'str')


ilevkivskyi:
(AppVeyor failure is a flake)

mypy/stubgen.py
```diff
@@ -892,86 +843,217 @@ def get_qualified_name(o: Expression) -> str:
...
+
+    return py_modules, c_modules
+
+
+def find_module_paths_using_imports(modules: List[str], packages: List[str],
+                                    interpreter: str,
+                                    pyversion: Tuple[int, int],
+                                    quiet: bool = True) -> Tuple[List[StubSource],
+                                                                 List[StubSource]]:
+    """Find path and runtime value of __all__ (if possible) for modules and packages."""
```
JukkaL: This docstring doesn't make it clear how this is different from `find_module_paths_using_search`. Can you update the two docstrings to explain the difference (the other uses runtime Python imports while the other one just looks for source files at the file system level).

mypy/stubgen.py
```diff
@@ -892,86 +843,217 @@ def get_qualified_name(o: Expression) -> str:
...
+    options.semantic_analysis_only = True
+    options.python_version = stubgen_options.pyversion
+    return options
+
+
+def generate_asts_for_modules(py_modules: List[StubSource],
+                              parse_only: bool, mypy_options: MypyOptions) -> None:
+    """Use mypy to parse (and optionally analyze) source files."""
+    if parse_only:
+        for mod in py_modules:
```
JukkaL: Optional idea for a refactoring: Move this block to a separate function (and maybe move the alternative approach to another function).

mypy/stubgen.py
```diff
@@ -892,86 +843,217 @@ def get_qualified_name(o: Expression) -> str:
...
+def generate_stubs(options: Options,
+                   # additional args for testing
+                   quiet: bool = False, add_header: bool = True) -> None:
+    """Main entry point for the program."""
+    mypy_opts = mypy_options(options)
+    py_modules, c_modules = collect_build_targets(options, mypy_opts)
+
+    # Collect info from docs (if given):
+    sigs = class_sigs = None  # type: Optional[Dict[str, str]]
+    if not options.doc_dir:
```
JukkaL: Why is this condition negated?

mypy/stubgen.py
```diff
@@ -988,14 +1070,15 @@ def parse_options(args: List[str]) -> Options:
...
     parser.add_argument('--ignore-errors', action='store_true',
                         help="ignore errors when trying to generate stubs for modules")
     parser.add_argument('--no-import', action='store_true',
                         help="don't import the modules, just parse and analyze them "
-                             "(doesn't work with C extension modules and doesn't "
+                             "(doesn't work with C extension modules and might not "
                              "respect __all__)")
+    parser.add_argument('--parse-only', action='store_true',
+                        help="don't perform semantic analysis of sources, just parse them "
+                             "(only applies to Python modules, might affect stubs quality)")
```
JukkaL: Minor nit: maybe use "quality of stubs" instead of "stubs quality".

mypy/stubgen.py
```diff
@@ -1009,40 +1092,51 @@ def parse_options(args: List[str]) -> Options:
...
-            continue
-        output = subprocess.check_output([candidate, '--version'],
-                                         stderr=subprocess.STDOUT).strip()
-        if b'Python 2' in output:
-            return candidate
-    raise SystemExit("Can't find a Python 2 interpreter -- please use the -p option")
+def main() -> None:
+    # Make sure that the current directory is in sys.path so that
+    # stubgen can be run on packages in the current directory.
+    if '' not in sys.path:
```
JukkaL: Should you also check for `.` in `sys.path`?

mypy/stubdoc.py
```diff
@@ -0,0 +1,146 @@
...
+
+    result = []
+    for name, name_sigs in sig_map.items():
+        if len(set(name_sigs)) == 1:
+            result.append((name, name_sigs[0]))
+    return sorted(result)
+
+
+def infer_sig_from_docstring(docstr: str, name: str) -> Optional[Tuple[str, str]]:
+    """Look for signature of function with given name in a docstring.
```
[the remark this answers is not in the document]
ilevkivskyi: This and below will be modified by https://github.com/python/mypy/pull/5975 (I will open an issue).

mypy/stubutil.py
```diff
@@ -94,6 +24,7 @@ def is_c_module(module: ModuleType) -> bool:
 
 def write_header(file: IO[str], module_name: Optional[str] = None,
                  pyversion: Tuple[int, int] = (3, 5)) -> None:
+    """Write a header to file indicating this file is auto-generated by stubgen."""
```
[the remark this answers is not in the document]
ilevkivskyi: I think it is mostly useful for Python 2 vs Python 3. I will leave only the major version.

mypy/stubutil.py
```diff
@@ -106,44 +37,126 @@ def write_header(file: IO[str], module_name: Optional[str] = None,
...
+        output = subprocess.check_output([candidate, '--version'],
+                                         stderr=subprocess.STDOUT).strip()
+        if b'Python 2' in output:
+            return candidate
+    raise SystemExit("Can't find a Python 2 interpreter -- "
+                     "please use the --python-executable option")
+
+
+def walk_packages(packages: List[str]) -> Iterator[str]:
+    """Iterates through all packages and sub-packages in the given list.
```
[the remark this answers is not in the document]
ilevkivskyi: > Also, does this work for both Python and C extension modules?

Yes, this uses runtime imports to find also C modules. I will update the docstring.

mypy/stubutil.py
```diff
@@ -106,44 +37,126 @@ def write_header(file: IO[str], module_name: Optional[str] = None,
...
+
+    Optionally report success.
+    """
+    try:
+        yield
+    except Exception as e:
+        if not ignore_errors:
+            raise e
+        else:
+            print("Stub generation failed for", mod, file=sys.stderr)
```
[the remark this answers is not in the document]
ilevkivskyi: The error is ignored only if `--ignore-errors` is passed, I will add a comment.

mypy/stubgen.py
```diff
@@ -248,18 +139,26 @@ def generate_stub(path: str,
...
+# for a given node. These should be manually replaced by a user.
+
+ERROR_MARKER = '<ERROR>'  # type: Final
+
 
 class AnnotationPrinter(TypeStrVisitor):
+    """Visitor used to print existing annotations in a file.
 
+    The main difference from TypeStrVisitor is a better treatment of
+    unbound types.
```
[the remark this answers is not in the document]
ilevkivskyi: This visitor actually just prints practically the same what `reveal_type()` does (i.e. it prints full names for instance types), so it is not very useful on its own. I will update the docstring.

mypy/stubgen.py
```diff
@@ -667,8 +589,14 @@ def is_type_expression(self, expr: Expression, top_level: bool = True) -> bool:
             elif expr.name == 'None':
                 return not top_level
             else:
-                return True
-        elif isinstance(expr, IndexExpr) and isinstance(expr.base, NameExpr):
+                return not self.is_private_name(expr.name)
+        elif isinstance(expr, MemberExpr) and self.analyzed:
+            # Also add function and module aliases.
+            return (top_level and (isinstance(expr.node, (FuncDef, Decorator, MypyFile)) or
+                                   isinstance(expr.node, TypeInfo)) and
```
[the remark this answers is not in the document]
ilevkivskyi: They supposed to be independent, it looks like I messed up parentheses.

mypy/stubgen.py
```diff
@@ -667,8 +589,14 @@ def is_type_expression(self, expr: Expression, top_level: bool = True) -> bool:
             elif expr.name == 'None':
                 return not top_level
             else:
-                return True
-        elif isinstance(expr, IndexExpr) and isinstance(expr.base, NameExpr):
+                return not self.is_private_name(expr.name)
+        elif isinstance(expr, MemberExpr) and self.analyzed:
+            # Also add function and module aliases.
```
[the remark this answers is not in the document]
ilevkivskyi: I just renamed the method.

mypy/stubgen.py
ilevkivskyi: Good question :-)

ilevkivskyi:
OK, playing with real stub generation discovered an issue: we should not add `-> None:` to functions without return statements if it is an abstract method.

docs/source/stubgen.rst
```diff
@@ -0,0 +1,143 @@
...
+By default stubgen will try to import the modules and packages given.
+This has an advantage of possibility to discover and stub also C modules.
+By default stubgen will use mypy to semantically analyze the Python
+sources found. To alter this behavior, you can use following flags:
+
+``--no-import``
+    Don't try to import modules, instead use mypy's normal mechanisms to find
+    sources. This will not find any C extension modules. Stubgen also uses
+    runtime introspection to find actual value of ``__all__``, so with this flag
+    the set of re-expoted names may be incomplete. This flag will be useful if
```
msullivan: `expoted`

docs/source/stubgen.rst
```diff
@@ -0,0 +1,143 @@
...
+    importing the module causes an error.
+
+``--parse-only``
+    Don't perform mypy semantic analysis of source files. This may generate
+    worse stubs, in particular some module, class, and function aliases may
```
msullivan: first comma should be a colon

JukkaL:
The changes in response to my review look good.

test-data/unit/stubgen.test
```diff
@@ -1382,3 +1382,56 @@ from typing import Any, Dict
...
+class A(metaclass=ABCMeta):
+    @abstractmethod
+    def meth(self):
+        pass
+[out]
+from abc import ABCMeta, abstractmethod
+
+class A(metaclass=ABCMeta):
+    @abstractmethod
+    def meth(self) -> None: ...
```
JukkaL: Should the return type be `Any`?

test-data/unit/stubgen.test
```diff
@@ -1382,3 +1382,56 @@ from typing import Any, Dict
...
+class A(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def meth(self):
+        pass
+[out]
+import abc
+
+class A(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def meth(self) -> None: ...
```
JukkaL: Again, should the return type be `Any`?

test-data/unit/stubgen.test
ilevkivskyi: The return is `Any` only if we run semantic analyzer (test name should end in `_semanal`), see a test below. But on the second thought it looks like it is not always necessary, in most common cases (like when we add the decorator here) it is better to return `Any`. I will fix this now.

ilevkivskyi:
Oh, GitHub somehow completely screwed up my merge. I will try to fix it.

ilevkivskyi:
OK, it was actually my fault, I didn't add one file to the commit :P

ilevkivskyi:
Unless there are some objections, I am going to merge this soon after tests pass.

PR #6257, opened 2019-01-25
New semantic analyzer: Fix redefinitions

JukkaL:
When we can't resolve a definition due to some missing dependencies,
we now don't record any additional definitions of the same name. This
is important since the first definition should always win. This applies
to things like assignments, function definitions, imports and class
definitions (anything that defines a name in the current scope).

Example of where things would go wrong:

```
from m import C  # Assume C can't be resolved yet

class C:  # Could be exposed, even though it's not the first def
    pass
```

Star imports are a bit special, since they could define arbitrary names
if the target module is incomplete. We use a special name `*` to
mark this. It disallows the definition of all names.

We don't track the scopes of definitions for simplicity. If definition
of `x` goes wrong, we also refuse to define it in any nested scope.
This may result in some extra deferrals, but otherwise shouldn't be
a problem.

This also refactors symbol table management. The key thing here is
that we want symbol table additions to go through a single function
so that we can skip them if they are defining something that was
already defined, but the previous definition couldn't be resolved. 

Some places still haven't been updated, but these are related to
features that aren't supported yet by the new analyzer. Also removed
some code related function redefinition, but again this isn't supported
yet and will have to be fixed anyway.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -370,7 +370,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         self.patches = patches
         self.deferred = False  # Set to true if another analysis pass is needed
         self.incomplete = False  # Set to true if current module namespace is missing things
-        # These names couldn't be added to the symbol table
+        # These names couldn't be added to the symbol table due to incomplete deps.
         self.missing_names = set()  # type: Set[str]
 
         if isinstance(node, MypyFile):
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -1749,6 +1749,10 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         i_id = self.correct_relative_import(i)
         if i_id in self.modules:
             m = self.modules[i_id]
+            if self.is_incomplete_namespace(i_id):
+                # Any names could be missing from the current namespace if the target module
+                # namespace is incomplete.
+                self.mark_incomplete('*')
             self.add_submodules_to_parent_modules(i_id, True)
             for name, orig_node in m.names.items():
                 node = self.dereference_module_cross_ref(orig_node)
@@ -3649,7 +3653,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         """Mark the current module namespace as incomplete (and defer current analysis target).
 
         Args:
-            name: The name that we weren't able to define
+            name: The name that we weren't able to define (or '*' if the name is unknown)
         """
         self.defer()
         self.incomplete = True
@@ -3830,7 +3834,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         if existing is not None and context is not None:
             if existing.node != symbol.node:
                 self.name_already_defined(name, context, existing)
-        elif name not in self.missing_names:
+        elif name not in self.missing_names and '*' not in self.missing_names:
             names[name] = symbol
 
     def add_module_symbol(self, id: str, as_id: str, module_public: bool,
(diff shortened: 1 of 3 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -3704,8 +3614,9 @@ def lookup_qualified(self, name: str, ctx: Context,
                             gvar = self.create_getattr_var(names['__getattr__'],
                                                            parts[i], parts[i])
                             if gvar:
-                                names[name] = gvar
-                                n = gvar
+                                n = SymbolTableNode(GDEF, gvar)
+                                # TODO: What was the purpose of this?
+                                # names[name] = n
```
ilevkivskyi: TBH, I forgot why. Approximately the idea was that two expressions like `a.A` (if `a` is a module with `__getattr__`) should point to the same "auto-generated" node. Probably I added this just in case to avoid troubles in incremental modes. Also this still sounds consistent to me.

mypy/newsemanal/semanal.py
```diff
@@ -4127,3 +4068,21 @@ def apply_semantic_analyzer_patches(patches: List[Tuple[int, Callable[[], None]]
...
+
+def names_defined_by_lvalue(lvalue: Lvalue) -> List[str]:
+    if isinstance(lvalue, NameExpr):
+        return [lvalue.name]
+    elif isinstance(lvalue, (ListExpr, TupleExpr)):
+        result = []  # type: List[str]
+        for item in lvalue.items:
+            result += names_defined_by_lvalue(item)
+        return result
+    return []
```
ilevkivskyi: What about `StarExpr`?

mypy/newsemanal/semanal.py
```diff
@@ -3889,53 +3805,86 @@ def current_symbol_kind(self) -> int:
             kind = GDEF
         return kind
 
-    def add_symbol(self, name: str, node: SymbolTableNode,
-                   context: Context) -> None:
+    def add_symbol(self, name: str, node: SymbolNode, context: Context,
+                   module_public: bool = True, module_hidden: bool = False) -> None:
         """Add symbol to the currently active symbol table."""
```
ilevkivskyi: Maybe mention in the docstring that all additions to symbol tables should go through this method and adding new ones is discouraged? (Of course, this will not guarantee it will really prevent this, but maybe it will help to at least some extent).

mypy/newsemanal/semanal.py
```diff
@@ -4127,3 +4068,21 @@ def apply_semantic_analyzer_patches(patches: List[Tuple[int, Callable[[], None]]
     patches_by_priority = sorted(patches, key=lambda x: x[0])
     for priority, patch_func in patches_by_priority:
         patch_func()
+
+
+def names_defined_by_assignment(s: AssignmentStmt) -> List[str]:
```
ilevkivskyi: Although this function is trivial, a short docstring mentioning where this is used will be helpful.

test-data/unit/check-newsemanal.test
```diff
@@ -268,3 +268,55 @@ x: T2
...
+x = 0
+
+def y(): pass # E: Name 'y' already defined on line 2
+reveal_type(y) # E: Revealed type is 'builtins.int'
+
+y2 = y
+class y2: pass # E: Name 'y2' already defined on line 7
+reveal_type(y2) # E: Revealed type is 'builtins.int'
+
+y3, y4 = y, y
```
ilevkivskyi: Maybe add the same test for `x, *y, z = [...]`?

test-data/unit/check-newsemanal.test
```diff
@@ -268,3 +268,55 @@ x: T2
...
+x: A
+reveal_type(x) # E: Revealed type is 'b.A'
+
+from b import *
+
+class A: pass # E: Name 'A' already defined (possibly by an import)
+
+[file b.py]
+class A: pass
+from a import x
```
ilevkivskyi: Maybe also add a test where a base class is start-imported in an import cycle? Something like:
```
[file a.py]
from b import *
class A(B): ...
[file b.py]
from a import A  # I am not sure about import here, may we need to try both orders of processing of modules
class B:
    x: A
```

PR #6263, opened 2019-01-29
Associate deps cache files with the depended upon module

msullivan:
Currently the fine-grained deps cache file for a module contains all the
fine-grained dependencies *from* the module (that is, those whose
targets are in the module). This has the advantage that producing fine-grained
dependency cache files is nicely localized (just run it on a module), but
the downside that *loading* the fine-grained deps cache is a global operation,
requiring every cache file to be loaded.

To improve daemon performance, we invert this, and have the deps cache file
for a module contain the fine-grained dependencies *to* the module. This
complicates creating the cache files: we need to explicitly sort out triggers
into the appropriate cache file.

This, then, allows us to postpone loading deps cache files until the module
is updated.

Because protocol dependencies now get sorted out into modules like everything
else, this notionally eliminates the special protocol deps cache. In practice, the
protocol deps cache has morphed in a cache for 'extra' dependencies that weren't
placed somewhere else (either because the module doesn't exist or because they
were added in an incremental run) and the meta file for it has grown additional
metadata to check consistency of the deps cache. (Because deps files are now
written after metadata, we can't store info about the in the metadata files.)

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -72,6 +72,7 @@ DEBUG_FINE_GRAINED = False  # type: Final
 
 Graph = Dict[str, 'State']
 
+
 # TODO: Get rid of BuildResult.  We might as well return a BuildManager.
 class BuildResult:
     """The result of a successful build.
@@ -718,11 +719,18 @@ def write_deps_cache(rdeps: Dict[str, Dict[str, Set[str]]],
                      manager: BuildManager, graph: Graph) -> None:
     """Write cache files for fine-grained dependencies.
 
-    Serialize fine-grained dependencies map for fine grained mode. Also take the snapshot
-    of current sources to later check consistency between the extra cache and individual
-    cache files.
+    Serialize fine-grained dependencies map for fine grained mode.
 
-    XXX: DESCRIBE
+    Dependencies on some module 'm' is stored in the dependency cache
+    file m.deps.json.  This entails some spooky action at a distance:
+    if module 'n' depends on 'm', that produces entries in m.deps.json.
+
+    This means that the validity of the fine-grained dependency caches
+    are a global property, so we store validity checking information
+    fine-grained dependencies in a global cache file.
+    To do this, we take a snapshot of current sources to later check
+    consistency between the extra cache and individual cache files as
+    well as storing the mtime for all of the dependency files.
     """
     metastore = manager.metastore
     meta_path, cache_path = get_deps_cache_name()
@@ -778,7 +786,6 @@ def invert_deps(
     rdeps = {id: {} for id in graph}  # type: Dict[str, Dict[str, Set[str]]]
     extra_deps = {}  # type: Dict[str, Set[str]]
     for trigger, targets in deps.items():
-        assert trigger[0] == '<'
         module = module_prefix(graph, trigger_to_target(trigger))
         if module and graph[module].tree:
             mod_rdeps = rdeps.setdefault(module, {})
@@ -866,7 +873,6 @@ def read_deps_cache(manager: BuildManager,
         return None
 
     module_deps_metas = deps_meta['deps_meta']
-    # XXX: check module metas
     for id, meta in module_deps_metas.items():
         try:
             matched = manager.getmtime(meta['path']) == meta['mtime']
@@ -943,13 +949,6 @@ def get_cache_names(id: str, path: str, manager: BuildManager) -> Tuple[str, str
     if manager.options.cache_fine_grained:
         deps_json = prefix + '.deps.json'
     return (prefix + '.meta.json', prefix + '.data.json', deps_json)
-
-
-# def get_deps_cache_name(id: str, path: str, manager: BuildManager) -> str:
-#     assert manager.options.cache_fine_grained
-#     _, _, deps_json = get_deps_cache_name(id, path, manager)
-#     assert deps_json is not None
-#     return deps_json
 
 
 def get_deps_cache_name() -> Tuple[str, str]:
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -726,8 +726,11 @@ def deps_to_json(x: Dict[str, Set[str]]) -> str:
     return json.dumps({k: list(v) for k, v in x.items()})
 
 
+DEPS_META_FILE = '@deps.meta.json'  # type: Final
+DEPS_EXTRA_FILE = '@extra.meta.json'  # type: Final
+
+
 def write_deps_cache(rdeps: Dict[str, Dict[str, Set[str]]],
-                     extra_deps: Dict[str, Set[str]],
                      manager: BuildManager, graph: Graph) -> None:
     """Write cache files for fine-grained dependencies.
 
@@ -736,6 +739,9 @@ def write_deps_cache(rdeps: Dict[str, Dict[str, Set[str]]],
     Dependencies on some module 'm' is stored in the dependency cache
     file m.deps.json.  This entails some spooky action at a distance:
     if module 'n' depends on 'm', that produces entries in m.deps.json.
+    When there is a dependency on a module that does not exist in the
+    build, it is stored with its first existing parent module. If no
+    such module exists, it is stored with the fake module '@extra'.
 
     This means that the validity of the fine-grained dependency caches
     are a global property, so we store validity checking information for
@@ -744,25 +750,18 @@ def write_deps_cache(rdeps: Dict[str, Dict[str, Set[str]]],
        between the fine-grained dependency cache and module cache metadata
      * We store the mtime of all of the dependency files to verify they
        haven't changed
-
-    There is additionally some "extra" dependency information that
-    isn't stored in any module-specific files. It is treated as
-    belonging to the module '@extra'.
     """
     metastore = manager.metastore
-    meta_path, cache_path = get_deps_cache_name()
 
     error = False
 
     fg_deps_meta = manager.fg_deps_meta.copy()
-    if not metastore.write(cache_path, deps_to_json(extra_deps)):
-        manager.log("Error writing fine-grained extra deps JSON file {}".format(cache_path))
-        error = True
-    else:
-        fg_deps_meta['@extra'] = {'path': cache_path, 'mtime': manager.getmtime(cache_path)}
 
     for id in rdeps:
-        _, _, deps_json = get_cache_names(id, graph[id].xpath, manager)
+        if id != '@extra':
+            _, _, deps_json = get_cache_names(id, graph[id].xpath, manager)
+        else:
+            deps_json = DEPS_EXTRA_FILE
         assert deps_json
         manager.log("Writing deps cache", deps_json)
         if not manager.metastore.write(deps_json, deps_to_json(rdeps[id])):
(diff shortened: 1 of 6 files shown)


mypy/build.py
```diff
@@ -715,46 +721,125 @@ def stats_summary(self) -> Mapping[str, object]:
...
+        if module and graph[module].tree:
+            mod_rdeps = rdeps.setdefault(module, {})
+            mod_rdeps.setdefault(trigger, set()).update(targets)
+        else:
+            extra_deps.setdefault(trigger, set()).update(targets)
+
+    return (rdeps, extra_deps)
+
+
+def process_deps(proto_deps: Dict[str, Set[str]],
```
JukkaL: The name is not very descriptive and docstring doesn't help.

This actually seems to do a few things:
* Load some fine-grained deps from the cache.
* Calculate fine-grained dependencies for files that were processed in the build (if this is an incremental build, this is a subset of all modules).
* Splits/inverts the dependencies per file.

I'd suggest renaming the function (maybe something like `generate_deps_for_cache`?). Also mention all the above things in the docstring (if they are true). Also it's worth mentioning if this is only used in coarse-grained incremental mode.

mypy/build.py
```diff
@@ -715,46 +721,125 @@ def stats_summary(self) -> Mapping[str, object]:
...
+    deps = manager.load_fine_grained_deps('@extra')
+    things = [st.compute_fine_grained_deps() for st in graph.values() if st.tree] + [proto_deps]
+    for st_deps in things:
+        for trigger, targets in st_deps.items():
+            deps.setdefault(trigger, set()).update(targets)
+
+    # If we are operating in non-incremental mode, properly split up
+    # dependencies between all the files. If this is an incremental update,
+    # though, we write all the dependencies into the "extra_deps" file so we
+    # don't need to reload dependency files to update them.
```
JukkaL: Discuss what goes into `extra_deps` here. As far as I understand it, it doesn't cover all fine-grained deps -- only working set that was processed in the build? If everything would go there, it seems like this would make incremental cache generation somewhat inefficient. Anyway, if the extra deps contain a lot of stuff, will this make the optimization in this PR less useful? Would it sometimes be more efficient to split the dependencies in an incremental build even if we'd need to read all fine-grained deps?

mypy/build.py
```diff
@@ -780,19 +865,16 @@ def read_plugins_snapshot(manager: BuildManager) -> Optional[Dict[str, str]]:
...
-    dependencies are read.
-    """
-    proto_meta, proto_cache = get_protocol_deps_cache_name()
-    meta_snapshot = _load_json_file(proto_meta, manager,
-                                    log_sucess='Proto meta ',
-                                    log_error='Could not load protocol metadata: ')
-    if meta_snapshot is None:
+def read_deps_cache(manager: BuildManager,
+                    graph: Graph) -> Optional[Dict[str, FgDepMeta]]:
+    """Read and validate dependencies cache. """
```
JukkaL: The docstring isn't helpful. What is "dependencies cache"?

mypy/build.py
```diff
@@ -873,17 +957,9 @@ def get_cache_names(id: str, path: str, manager: BuildManager) -> Tuple[str, str
...
-    Since these dependencies represent a global state of the program, they
-    are serialized per program, not per module, and the corresponding files
-    live at the root of the cache folder for a given Python version.
-    Return a tuple ('meta file path', 'data file path'), where the meta file
-    contains hashes of all source files at the time the protocol dependencies
-    were written, and data file contains the protocol dependencies.
-    """
-    name = '@proto_deps'
+def get_deps_cache_name() -> Tuple[str, str]:
+    """Return file names for global fine grained dependencies cache."""
```
JukkaL: Here it's not clear what "global fine grained dependencies cache" is. How does it relate to "extra_deps" above?

mypy/build.py
```diff
@@ -989,13 +1064,6 @@ def validate_meta(meta: Optional[CacheMeta], id: str, path: Optional[str],
     if data_mtime != meta.data_mtime:
         manager.log('Metadata abandoned for {}: data cache is modified'.format(id))
         return None
-    deps_mtime = None
-    if manager.options.cache_fine_grained:
```
JukkaL: Why was this removed?

mypy/build.py
```diff
@@ -1184,14 +1250,6 @@ def write_cache(id: str, path: str, tree: MypyFile,
             return interface_hash, None
         data_mtime = manager.getmtime(data_json)
 
-    deps_mtime = None
-    if deps_json:
```
JukkaL: Why was this removed?

mypy/build.py
```diff
@@ -2593,14 +2646,7 @@ def process_graph(graph: Graph, manager: BuildManager) -> None:
 
 def process_fine_grained_cache_graph(graph: Graph, manager: BuildManager) -> None:
```
JukkaL: Can this be removed, since it's empty?

mypy/server/update.py
```diff
@@ -402,6 +405,24 @@ def find_unloaded_deps(manager: BuildManager, graph: Dict[str, State],
     return unloaded
 
 
+def ensure_deps_loaded(module: str,
+                       deps: Dict[str, Set[str]], graph: Dict[str, State]) -> None:
+    """Ensure that the dependencies on a module are loaded.
```
JukkaL: It's worth mentioning in the docstring what state is mutated. In particular, is `deps` modified?

test-data/unit/fine-grained-cache-incremental.test
```diff
@@ -51,6 +51,34 @@ x = 'hi'
 ==
 a.py:3: error: Unsupported operand types for + ("int" and "str")
 
+[case testIncrCacheDoubleChange1]
```
JukkaL: Would it be useful to have a test case that covers the extra dependencies, and in particular reading dependencies from parent packages? Or are there existing test cases that deal with it?

mypy/test/testfinegrained.py
```diff
@@ -86,6 +86,7 @@ def run_case(self, testcase: DataDrivenTestCase) -> None:
         step = 1
         sources = self.parse_sources(main_src, step, options)
         if step <= num_regular_incremental_steps:
+            build_options.incremental = step > 1
```
JukkaL: What's the purpose of this change?

mypy/build.py
msullivan: Oh oops, yeah

test-data/unit/fine-grained-cache-incremental.test
msullivan: There are existing test cases that cover it.

mypy/build.py
msullivan: Because we can't write out the deps file before the meta file anymore, so it can't have the mtime. We store mtime in another cache file.

mypy/build.py
msullivan: I was shoving all of the new stuff into extra_deps because it seemed simpler and unlikely to be that expensive on the daemon side. (Usually not a lot would be generated.) But having different behavior for `--incremental` and not is kind of bad, so I changed it to reload deps and split them in incremental runs.

I also got rid of `extra_deps` basically everywhere in favor of having it be `@root` in the normal map

mypy/test/testfinegrained.py
msullivan: It was so that we'd split into multiple files on the first pass, but I got rid of that behavior

mypy/build.py
```diff
@@ -261,16 +259,15 @@ def default_data_dir() -> str:
 # suppressed contains those reachable imports that were prevented by
 # silent mode or simply not found.
 
+FgDepMeta = TypedDict('FgDepMeta', {'path': str, 'mtime': int})
```
JukkaL: Add comment.

mypy/build.py
```diff
@@ -504,6 +499,7 @@ def __init__(self, data_dir: str,
         self.version_id = version_id
         self.modules = {}  # type: Dict[str, MypyFile]
         self.missing_modules = set()  # type: Set[str]
+        self.fg_deps_meta = {}  # type: Dict[str, FgDepMeta]
```
JukkaL: Document this (in the docstring above).

mypy/build.py
```diff
@@ -715,46 +722,140 @@ def stats_summary(self) -> Mapping[str, object]:
...
 
-def write_protocol_deps_cache(proto_deps: Dict[str, Set[str]],
-                              manager: BuildManager, graph: Graph) -> None:
-    """Write cache files for protocol dependencies.
+def deps_to_json(x: Dict[str, Set[str]]) -> str:
+    return json.dumps({k: list(v) for k, v in x.items()})
+
+
+DEPS_META_FILE = '@deps.meta.json'  # type: Final
+DEPS_ROOT_FILE = '@root.deps.json'  # type: Final
```
JukkaL: Briefly document the purpose of these files.

mypy/build.py
```diff
@@ -715,46 +722,140 @@ def stats_summary(self) -> Mapping[str, object]:
...
     metastore = manager.metastore
-    proto_meta, proto_cache = get_protocol_deps_cache_name()
-    meta_snapshot = {}  # type: Dict[str, str]
+
     error = False
+
+    fg_deps_meta = manager.fg_deps_meta.copy()
+
+    for id in rdeps:
+        if id != '@root':
```
JukkaL: What about using a constant for `'@root'`?

mypy/build.py
```diff
@@ -715,46 +722,140 @@ def stats_summary(self) -> Mapping[str, object]:
...
     if error:
         manager.errors.set_file(_cache_dir_prefix(manager), None)
-        manager.errors.report(0, 0, "Error writing protocol dependencies cache",
+        manager.errors.report(0, 0, "Error writing fine-grained dependencies cache",
                               blocker=True)
 
 
+def invert_deps(deps: Dict[str, Set[str]],
+                graph: Graph) -> Dict[str, Dict[str, Set[str]]]:
+    """Splits fine-grained dependencies based on the module of the trigger
```
JukkaL: Style nit: missing period at the end of first sentence.

mypy/build.py
```diff
@@ -715,46 +722,140 @@ def stats_summary(self) -> Mapping[str, object]:
...
+
+    Returns a dictionary from module ids to all dependencies on that
+    module. Dependencies not associated with a module in the build are
+    associated with the fake module '@root'.
```
JukkaL: Maybe also mention that deps can also be associated with a parent package if it exists.

mypy/build.py
```diff
@@ -780,19 +881,19 @@ def read_plugins_snapshot(manager: BuildManager) -> Optional[Dict[str, str]]:
...
-                        graph: Graph) -> Optional[Dict[str, Set[str]]]:
-    """Read and validate protocol dependencies cache.
+def read_deps_cache(manager: BuildManager,
+                    graph: Graph) -> Optional[Dict[str, FgDepMeta]]:
+    """Read and validate the fine-grained dependencies cache.
 
-    See docstring for write_protocol_cache for details about which kinds of
-    dependencies are read.
+    See the write_deps_cache documentation for more information on
+    the details of the cache.
```
JukkaL: Document `None` return value.

mypy/build.py
```diff
@@ -715,46 +722,140 @@ def stats_summary(self) -> Mapping[str, object]:
...
+        mod_rdeps = rdeps.setdefault(module, {})
+        mod_rdeps.setdefault(trigger, set()).update(targets)
+
+    return rdeps
+
+
+def generate_deps_for_cache(proto_deps: Dict[str, Set[str]],
+                            manager: BuildManager,
+                            graph: Graph) -> Dict[str, Dict[str, Set[str]]]:
+    """Generate fine-grained dependenecies into a form suitable for serializing.
```
JukkaL: Typo: dependenecies

PR #6270, opened 2019-01-30
Now that typed_ast 1.3.0 is released, mandate it

gvanrossum:


mr-c:
Is typed_ast safe or an improvement for mypy 0.660? I noticed some Py3.7 fixes

That is, should I update the typed-ast package in Debian before the final freeze for their upcoming release of Debian Buster? Debian Buster is only shipping Python 3.7 of the Python 3 series..

gvanrossum:
In general you won't notice much of a difference. But for mypy 0.670 we'll require typed_ast 1.3.x, so you might as well get on that train.

ethanhs:
Since Buster is only shipping 3.7, and a goal of typed_ast is to be bug-for-bug compatible, I think it would be better to use typed_ast 1.3.x, since the newer typed_ast has better compatibility with  (and is based on) the 3.7 ast module. Though if you do, you should probably bump mypy to 0.670 when it is released (hopefully this Friday).

PR #6272, opened 2019-01-30
Fixed overlapped I/O when more data needs to be read

ethanhs:
For some reason the tests didn't capture this, but ReadFile, not GetOverlappedResult returns _winapi.ERROR_MORE_DATA, which means that we would not do another read (instead we would fail an assert). This handles the case where there is more data to be read and correctly peeks then reads the remaining data.

I also increased the size of the test message to make sure we don't miss errors like this in the future, and added more information when assertions fail.

### DIFF
--- a/mypy/ipc.py
+++ b/mypy/ipc.py
@@ -77,20 +77,8 @@ class IPCBase:
                     # we are done!
                     break
                 elif err == _winapi.ERROR_MORE_DATA:
-                    left = _winapi.PeekNamedPipe(self.connection)[1]
-                    assert left > 0
-                    ov, err = _winapi.ReadFile(self.connection, left, overlapped=True)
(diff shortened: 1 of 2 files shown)


mypy/ipc.py
```diff
@@ -48,25 +48,24 @@ class IPCBase:
     connection = None  # type: _IPCHandle
 
     def __init__(self, name: str, timeout: Optional[float]) -> None:
-        self.READ_SIZE = 100000
         self.name = name
         self.timeout = timeout
 
-    def read(self) -> bytes:
+    def read(self, size: int = 100000) -> bytes:
```
msullivan: I'm not sure if this is worth putting in the interface, since it doesn't modify any visible behavior.

msullivan: Alright, sure

PR #6280, opened 2019-01-31
New semantic analyzer: Support multiple passes over functions.

ilevkivskyi:
This allows supporting forward references within function namespaces. The logic is very simple, go over and over each top-level function/method until everything inside it is resolved, plus one more pass to report errors.

The big diff in `_visit_func_def()` is because of whitespace, I didn't change any function body analysis logic apart from removing previous two-pass special-casing for functions (GitHub has an option to ignore whitespace in the diff).


mypy/test/testcheck.py
```diff
@@ -150,6 +150,8 @@ def run_case_once(self, testcase: DataDrivenTestCase,
         options.show_traceback = True
         if 'optional' in testcase.file:
             options.strict_optional = True
+        if 'newsemanal' in testcase.file:
+            options.new_semantic_analyzer = True
```
JukkaL: This is a good idea!

mypy/newsemanal/semanal_main.py
```diff
@@ -92,13 +91,29 @@ def process_functions(graph: 'Graph', scc: List[str]) -> None:
...
         assert tree is not None
+        analyzer = graph[module].manager.new_semantic_analyzer
         symtable = tree.names
         targets = get_all_leaf_targets(symtable, module, None)
         for target, node, active_type in targets:
-            deferred, incomplete = semantic_analyze_target(target, graph[module], node,
-                                                           active_type)
-            assert not deferred  # There can't be cross-function forward refs
-            assert not incomplete  # Ditto
+            iteration = 0
```
JukkaL: It would be better to move the new logic to a separate function. I think that `process_functions` is now too big.

mypy/newsemanal/semanal.py
```diff
@@ -630,6 +602,9 @@ def analyze_overload_sigs_and_impl(
         types = []
         non_overload_indexes = []
         impl = None  # type: Optional[OverloadPart]
+        if defn.impl:
```
JukkaL: Ew the way this works is ugly (the existing logic, not this change in particular). Add TODO item about not modifying the `items` list.

test-data/unit/check-newsemanal.test
```diff
@@ -336,6 +312,81 @@ class A: pass # E: Name 'A' already defined (possibly by an import)
...
+                    def meth(self) -> None:
+                        self.t: Test
+                        class Test:
+                            def test(self) -> None:
+                                def one() -> int:
+                                    reveal_type(other())  # E: Revealed type is 'builtins.str'
+                                    return int()
+                                def other() -> str:
+                                    reveal_type(one())  # E: Revealed type is 'builtins.int'
+                                    return str()
```
JukkaL: It would be good to have a test case where we hit the maximum iteration count (cyclic definition).

mypy/newsemanal/semanal.py
```diff
@@ -228,6 +211,8 @@ def __init__(self,
             errors: Report analysis errors using this instance
         """
         self.locals = [None]
+        # Saved namespaces from previous passes.
```
JukkaL: A more detailed explanation would be nice here. Also, avoid the use of the term 'pass' since this means a different thing in the old semantic analyzer (maybe 'iteration'?).

mypy/newsemanal/semanal.py
```diff
@@ -3797,8 +3759,12 @@ def qualified_name(self, n: str) -> str:
             base = self.cur_mod_id
         return base + '.' + n
 
-    def enter(self) -> None:
-        self.locals.append(SymbolTable())
+    def enter(self, function: Optional[FuncItem] = None) -> None:
+        if function:
```
JukkaL: Add docstring and describe the argument. Also discuss when it's okay to leave it out.

mypy/newsemanal/semanal_main.py
```diff
@@ -92,13 +91,29 @@ def process_functions(graph: 'Graph', scc: List[str]) -> None:
...
+            # We need one more pass after incomplete is False.
```
JukkaL: Similar to above, it's probably better to avoid the term 'pass' here. Also consider renaming the `more_passes` variable.

ilevkivskyi:
> I mean things like `x; x = 1` currently generate an error.

Yes, it looks like non-trivial to fix in general, I will open an issue.

PR #6282, opened 2019-02-01
Optimize find_module by enumerating contents of directories on lib_path

msullivan:
In practice, almost all modules can be routed to the correct entry in
the lib_path by looking at just the first component of the module
name.

We take advantage of this by enumerating the contents of all of the
directories on the lib_path and building a map of which entries in
the lib_path could contain each potential top-level module that appears.

The previous cache for find_lib_path_dirs could co-exist with this,
but I removed it for simplicity since it didn't help anymore in my
measurements.

This speeds up module finding in a dmypy cold run on S by 10x.

### DIFF
--- a/mypy/modulefinder.py
+++ b/mypy/modulefinder.py
@@ -70,7 +70,8 @@ class FindModuleCache:
                  options: Optional[Options] = None) -> None:
         self.search_paths = search_paths
         self.fscache = fscache or FileSystemCache()
-        # Cache for get_toplevel_possibilities: dir_chain -> (id -> list(package_dirs))
+        # Cache for get_toplevel_possibilities:
+        # search_paths -> (toplevel_id -> list(package_dirs))
         self.initial_components = {}  # type: Dict[Tuple[str, ...], Dict[str, List[str]]]
         # Cache find_module: id -> result
         self.results = {}  # type: Dict[str, Optional[str]]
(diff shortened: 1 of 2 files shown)


mypy/modulefinder.py
```diff
@@ -70,38 +70,60 @@ def __init__(self,
                  options: Optional[Options] = None) -> None:
         self.search_paths = search_paths
         self.fscache = fscache or FileSystemCache()
-        # Cache find_lib_path_dirs: (dir_chain, search_paths) -> list(package_dirs, should_verify)
-        self.dirs = {}  # type: Dict[Tuple[str, Tuple[str, ...]], PackageDirs]
+        # Cache for get_toplevel_possibilities: dir_chain -> (id -> list(package_dirs))
```
JukkaL: Nit: Maybe document that `dir_chain` contains module search directories. Maybe also explain that `id` is a top-level package name.

mypy/modulefinder.py
```diff
@@ -70,38 +70,60 @@ def __init__(self,
...
-        # to exist. This is shared among different module ids when they differ only
-        # in the last component.
-        # This is run for the python_path, mypy_path, and typeshed_path search paths
-        key = (dir_chain, lib_path)
-        if key not in self.dirs:
-            self.dirs[key] = self._find_lib_path_dirs(dir_chain, lib_path)
-        return self.dirs[key]
-
-    def _find_lib_path_dirs(self, dir_chain: str, lib_path: Tuple[str, ...]) -> PackageDirs:
+    def find_lib_path_dirs(self, id: str, lib_path: Tuple[str, ...]) -> PackageDirs:
```
JukkaL: Add docstring.

mypy/modulefinder.py
```diff
@@ -70,38 +70,60 @@ def __init__(self,
...
+
+        if lib_path in self.initial_components:
+            return self.initial_components[lib_path].get(id, [])
+
+        # Enumerate all the files in the directories on lib_path and produce the map
+        components = {}
+        for dir in lib_path:
+            try:
+                contents = self.fscache.listdir(dir)
+            except FileNotFoundError:
```
JukkaL: What about other possible errors? I guess it's okay to be consistent with how errors are handled elsewhere.

msullivan: Hm I think it should actually be `OSError` for consistency with how fscache works

mypy/modulefinder.py
msullivan: oh oops, `dir_chain` should actually be `search_path`

PR #6302, opened 2019-02-01
New semantic analyzer: Add placeholder types etc.

JukkaL:
This adds the concept of placeholder type (and placeholder `TypeInfo`),
which are types that don't have a complete definition yet. They can be
useful in certain self-referential definition such as this:

```py
class C(Sequence[C]): ...
```

In the above example, we create a placeholder `TypeInfo` for `C`
before analyzing base classes. This allow `C` in `Sequence[C]` to
be bound. Each time there's a reference to a placeholder we defer
the current target so that a proper reference can be created on a
subsequent iteration. Placeholders should not leak semantic 
analysis and will generate an error during type checking.

They aren't that useful yet, but the goal is to eventually expand them
to support general recursive types.

This PR also includes a few additional small updates, including some
refactoring of how protocol types are detected (the refactoring isn't
directly related to the rest of the PR).

### DIFF
--- a/mypy/newsemanal/semanal_main.py
+++ b/mypy/newsemanal/semanal_main.py
@@ -67,10 +67,13 @@ def process_top_levels(graph: 'Graph', scc: List[str]) -> None:
     worklist = scc[:]
     iteration = 0
     while worklist:
+        print(iteration, worklist)
         iteration += 1
         if iteration == MAX_ITERATIONS:
             # Give up. Likely it's impossible to bind all names.
             state.manager.incomplete_namespaces.clear()
+        elif iteration > MAX_ITERATIONS:
+            assert False, 'Max iteration count reached in semantic analysis'
         all_deferred = []  # type: List[str]
         while worklist:
             next_id = worklist.pop()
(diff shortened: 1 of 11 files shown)


mypy/newsemanal/semanal_main.py
```diff
@@ -71,6 +71,8 @@ def process_top_levels(graph: 'Graph', scc: List[str]) -> None:
         if iteration == MAX_ITERATIONS:
             # Give up. Likely it's impossible to bind all names.
             state.manager.incomplete_namespaces.clear()
+        elif iteration > MAX_ITERATIONS:
+            assert False, 'Max iteration count reached in semantic analysis'
```
ilevkivskyi: We should probably add the same to functions after other PR is merged.

mypy/nodes.py
```diff
@@ -2668,6 +2668,43 @@ def deserialize(cls, data: JsonDict) -> 'TypeAlias':
...
+    semantic analyzer. These are created if some dependencies of a type
+    definition are not yet complete. Their purpose is to allow
+    PlaceholderType instances to be created for types that refer to
+    incomplete types. Example where this is useful:
+
+      class C(Sequence[C]): ...
+
+    We create a PlaceholderTypeInfo for C so that the type C in
+    Sequence[C] can be bound. Without a placeholder, the class
+    definition couldn't be analyzed.
```
ilevkivskyi: After some thinking just this use case is not enough motivation for introducing `PlaceholderType` and `PlaceholderTypeInfo`. Since we just need the top-level of every base to have a valid `TypeInfo`, we can just reuse the `ForwardRef` and then resolve it on the subsequent iterations using local lookup of the name `'C'`. The real motivation at this step IMO is not passing local lookup function to `typeanal.py` (and of course recursive named tuples, aliases etc. in future). I agree this is a step in the right direction, but I think this should be explicitly mentioned somewhere in the docstring (if you agree).

PR #6330, opened 2019-02-04
New semantic analyzer: support named tuples

### LINKED ISSUE #6284: New semantic analyzer: class-based named tuple

### LINKED ISSUE #6285: New semantic analyzer: assignment-based named tuple

ilevkivskyi:
Fixes #6284 
Fixes #6285 

This doesn't add support for recursive definitions, but all the forward reference patterns that are supported in old semantic analyzer, should be also supported.

Some notes:
* I moved one function to `semanal_namedtuple.py`
* I never create new `TypeInfo`s in multi-iteration scenario
* I add a dummy type variable `_NT` to the bodies (otherwise type variable scope breaks), but it is probably fine because names starting with underscores are anyway prohibited in named tuples.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -867,15 +867,15 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 self.prepare_class_def(defn, info)
             return
 
-        is_named_tuple, info = self.named_tuple_analyzer.analyze_namedtuple_classdef(defn, defn.info)
+        is_named_tuple, info = self.named_tuple_analyzer.analyze_namedtuple_classdef(defn)
         if is_named_tuple:
             if info is None:
                 self.mark_incomplete(defn.name)
             else:
                 self.prepare_class_def(defn, info)
                 with self.scope.class_scope(defn.info):
-                    self.named_tuple_analyzer.process_namedtuple_body(defn, info,
-                                                                      self.analyze_class_body_common)
+                    with self.named_tuple_analyzer.save_namedtuple_body(info):
+                        self.analyze_class_body_common(defn)
             return
 
         # Create TypeInfo for class now that base classes and the MRO can be calculated.
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -907,12 +907,12 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         if defn.info and defn.info.tuple_type:
             # Don't reprocess everything. We just need to process methods defined
             # in the named tuple class body.
-            is_named_tuple, info = True, defn.info
+            is_named_tuple, info = True, defn.info  # type: bool, Optional[TypeInfo]
         else:
             is_named_tuple, info = self.named_tuple_analyzer.analyze_namedtuple_classdef(defn)
         if is_named_tuple:
             if info is None:
-                self.mark_incomplete(defn.name)
+                self.mark_incomplete(defn.name, defn)
             else:
                 self.prepare_class_def(defn, info)
                 with self.scope.class_scope(defn.info):
@@ -1214,7 +1214,11 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             #       incremental mode and we should avoid it.
             if '@' not in defn.info._fullname:
                 local_name = defn.info._fullname + '@' + str(defn.line)
-                defn.info._fullname = self.cur_mod_id + '.' + local_name
+                if defn.info.is_named_tuple:
+                    # Module already correctly set for named tuples.
+                    defn.info._fullname += '@' + str(defn.line)
+                else:
+                    defn.info._fullname = self.cur_mod_id + '.' + local_name
             else:
                 # Preserve name from previous fine-grained incremental run.
                 local_name = defn.info._fullname
@@ -1410,11 +1414,15 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                               allow_placeholder: bool = False) -> Optional[Type]:
         if isinstance(expr, CallExpr):
             expr.accept(self)
-            info = self.named_tuple_analyzer.check_namedtuple(expr, None, self.is_func_scope())
-            if info is None:
+            is_named_tuple, info = self.named_tuple_analyzer.check_namedtuple(expr, None,
+                                                                              self.is_func_scope())
+            if not is_named_tuple:
                 # Some form of namedtuple is the only valid type that looks like a call
                 # expression. This isn't a valid type.
                 raise TypeTranslationError()
+            elif not info:
+                self.defer()
+                return None
             assert info.tuple_type, "NamedTuple without tuple type"
             fallback = Instance(info, [])
             return TupleType(info.tuple_type.items, fallback=fallback)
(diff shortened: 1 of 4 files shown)


mypy/newsemanal/semanal_namedtuple.py
```diff
@@ -403,8 +431,42 @@ def make_init_arg(var: Var) -> Argument:
...
                          Argument(Var('len'), special_form_any, EllipsisExpr(), ARG_NAMED_OPT)])
+
+        self_tvar_expr = TypeVarExpr(SELF_TVAR_NAME, info.fullname() + '.' + SELF_TVAR_NAME,
+                                     [], info.tuple_type)
+        info.names[SELF_TVAR_NAME] = SymbolTableNode(MDEF, self_tvar_expr)
         return info
 
+    @contextmanager
+    def save_namedtuple_body(self, named_tuple_info: TypeInfo) -> Iterator[None]:
+        """Analyze body of class-based named tuple if the NamedTuple base class is present.
```
JukkaL: The docstring doesn't seem right.

mypy/newsemanal/semanal_namedtuple.py
```diff
@@ -98,7 +118,8 @@ def check_namedtuple_classdef(
                     types.append(AnyType(TypeOfAny.unannotated))
                 else:
                     analyzed = self.api.anal_type(stmt.type)
-                    assert analyzed is not None  # TODO: Handle None values
+                    if analyzed is None:
+                        return None
```
JukkaL: Add comment about this (something is incomplete).

mypy/newsemanal/semanal_namedtuple.py
```diff
@@ -35,35 +35,55 @@
...
-                        node.node = info
-                        defn.info.replaced = info
-                        defn.info = info
-                        defn.analyzed = NamedTupleExpr(info, is_typed=True)
-                        defn.analyzed.line = defn.line
-                        defn.analyzed.column = defn.column
-                        return info
-        return None
+                    if defn.info:
+                        # Don't reprocess everything, potentially deferred methods will be
```
JukkaL: Grammar nit: Run-on sentence, use period instead of comma.

mypy/newsemanal/semanal.py
```diff
@@ -867,6 +867,17 @@ def analyze_class(self, defn: ClassDef) -> None:
                 self.prepare_class_def(defn, info)
             return
 
+        is_named_tuple, info = self.named_tuple_analyzer.analyze_namedtuple_classdef(defn)
+        if is_named_tuple:
+            if info is None:
```
JukkaL: I'd suggest moving the new code to a separate method, as the method is already quite long.

mypy/newsemanal/semanal.py
```diff
@@ -1293,7 +1264,7 @@ def configure_base_classes(self,
 
         for base, base_expr in bases:
             if isinstance(base, TupleType):
-                if info.tuple_type:
+                if info.tuple_type and info.tuple_type != base:
```
JukkaL: Is there any case where `info.tuple_type` and `base` would be equal? I'd expect that at least the fallbacks would be different. Also, we usually use `is_same_type` instead of equality.

mypy/newsemanal/semanal.py
```diff
@@ -1882,7 +1853,12 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
         self.check_and_set_up_type_alias(s)
         self.newtype_analyzer.process_newtype_declaration(s)
         self.process_typevar_declaration(s)
-        self.named_tuple_analyzer.process_namedtuple_definition(s, self.is_func_scope())
+        if isinstance(s.rvalue, CallExpr) and isinstance(s.rvalue.analyzed, NamedTupleExpr):
+            existing_info = s.rvalue.analyzed.info  # type: Optional[TypeInfo]
```
JukkaL: I think that it would be better to do nothing on subsequent iterations if we've already successfully analyzed a named tuple definition. Analysis should only be successful if everything can be bound, and in that case the second iteration should produce the same result. However, in case there is a bug that causes the first iteration to produce an incomplete or invalid type, the second iteration could hide this bug in many cases, making the bug harder to find. Additionally, the logic would be a bit simpler.

We can't use this with normal class definitions though, since base types can be incomplete after the first iteration.

mypy/newsemanal/semanal_namedtuple.py
```diff
@@ -35,35 +35,55 @@
...
+                        # reprocessed later.
+                        return True, defn.info
+                    result = self.check_namedtuple_classdef(defn)
+                    if result is None:
+                        return True, None
```
JukkaL: Add short comment describing what this means.

mypy/newsemanal/semanal_namedtuple.py
```diff
@@ -35,35 +35,55 @@
...
+                    items, types, default_items = result
+                    info = self.build_namedtuple_typeinfo(
+                        defn.name, items, types, default_items)
+                    defn.info = info
+                    defn.analyzed = NamedTupleExpr(info, is_typed=True)
+                    defn.analyzed.line = defn.line
+                    defn.analyzed.column = defn.column
+                    return True, info
+        return False, None
```
JukkaL: Add short comment here.

mypy/newsemanal/semanal_namedtuple.py
```diff
@@ -35,35 +35,55 @@
...
+                    if result is None:
+                        return True, None
+                    items, types, default_items = result
+                    info = self.build_namedtuple_typeinfo(
+                        defn.name, items, types, default_items)
+                    defn.info = info
+                    defn.analyzed = NamedTupleExpr(info, is_typed=True)
+                    defn.analyzed.line = defn.line
+                    defn.analyzed.column = defn.column
+                    return True, info
```
JukkaL: Add short comment here.

mypy/newsemanal/semanal_namedtuple.py
```diff
@@ -115,16 +136,21 @@ def check_namedtuple_classdef(
...
         return items, types, default_items
 
-    def process_namedtuple_definition(self, s: AssignmentStmt, is_func_scope: bool) -> None:
+    def process_namedtuple_definition(self, s: AssignmentStmt, is_func_scope: bool,
+                                      existing_info: Optional[TypeInfo]) -> None:
         """Check if s defines a namedtuple; if yes, store the definition in symbol table.
 
-        Assume that s has already been successfully analyzed.
+        Assume that s has already been successfully analyzed. Use existing_info
+        if given instead of creating a new one.
```
JukkaL: As discussed above, I think that it would be better to avoid calling this method if this has been successfully analyzed before.

mypy/newsemanal/semanal_namedtuple.py
```diff
@@ -403,8 +431,42 @@ def make_init_arg(var: Var) -> Argument:
...
+                                     [], info.tuple_type)
+        info.names[SELF_TVAR_NAME] = SymbolTableNode(MDEF, self_tvar_expr)
         return info
 
+    @contextmanager
+    def save_namedtuple_body(self, named_tuple_info: TypeInfo) -> Iterator[None]:
+        """Analyze body of class-based named tuple if the NamedTuple base class is present.
+
+        Temporarily clear the names dict so we don't get errors about duplicate names
+        that were already set in build_namedtuple_typeinfo.
```
JukkaL: Mention why these shouldn't produce errors.

ilevkivskyi:
@JukkaL Thanks for review!

Because you merged your PR first, it was necessary to update my implementation logic for assignment based named tuples (basically it is much closer to class based one now). I however discovered couple small bugs:
* Wrong full name was set for named tuples nested in methods (this area around names with `@line` is still dark and will need some clean-up later).
* Processing of functions/methods were stopping to soon.

I fixed them and added more tests.

mypy/newsemanal/semanal.py
```diff
@@ -1880,17 +1888,32 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
...
+            return
+        lvalue = s.lvalues[0]
+        name = lvalue.name
+        is_named_tuple, info = self.named_tuple_analyzer.check_namedtuple(s.rvalue, name,
+                                                                          self.is_func_scope())
+        if not is_named_tuple:
+            return
+        # Yes, it's a valid namedtuple, but defer if it is not ready.
+        if not info:
+            self.add_symbol(name, PlaceholderNode(self.qualified_name(name), lvalue, True),
```
JukkaL: `mark_incomplete` adds a `PlaceholderNode`. The logic is a bit delicate so it's important that all additions go through the same implementation.

mypy/newsemanal/semanal_main.py
```diff
@@ -123,7 +123,7 @@ def process_top_level_function(analyzer: 'NewSemanticAnalyzer',
     analyzer.incomplete_namespaces.add(module)
     while deferred and more_iterations:
         iteration += 1
-        if not incomplete or iteration == MAX_ITERATIONS:
+        if not (deferred or incomplete) or iteration == MAX_ITERATIONS:
```
JukkaL: A namespace can be deferred while it's incomplete. I think that we should also remove the namespace from incomplete namespaces if it's complete. Now it looks like that we sometimes don't remove the namespace from `incomplete_namespaces` even though we probably should.

test-data/unit/check-newsemanal.test
```diff
@@ -805,6 +825,61 @@ Out = NamedTuple('Out', [('x', In), ('y', Other)])
 In = NamedTuple('In', [('s', str), ('t', Other)])
 class Other: pass
 
+[case testNewAnalyzerNamedTupleMixed1]
```
JukkaL: Is there a specific reason for adding these tests? It will be impractical to test all combinations of features anyway, as the number of test cases will grow too quickly. For example, is this more useful than a mixed named tuple vs. typed dict test case? It may be better to only test combinations that are known to be risky or have been broken previously.

mypy/newsemanal/semanal.py
```diff
@@ -1214,7 +1214,11 @@ def prepare_class_def(self, defn: ClassDef, info: Optional[TypeInfo] = None) ->
             #       incremental mode and we should avoid it.
             if '@' not in defn.info._fullname:
                 local_name = defn.info._fullname + '@' + str(defn.line)
-                defn.info._fullname = self.cur_mod_id + '.' + local_name
+                if defn.info.is_named_tuple:
+                    # Module already correctly set for named tuples.
+                    defn.info._fullname += '@' + str(defn.line)
```
JukkaL: It's unclear why we need this special case. Even if the module was already correctly set, if it's always `self.cur_mod_id`, I think that it's better to avoid this if statement if it doesn't change behavior.

test-data/unit/check-newsemanal.test
ilevkivskyi: Yes, these two tests cases were broken, but then the breakage had the same cause as for assignment based vs assignment based, so it is probably OK to remove them.

mypy/newsemanal/semanal.py
ilevkivskyi: As discussed elsewhere this is (unfortunately) actually needed.

PR #6335, opened 2019-02-05
Run some tests with new semantic analyzer

ilevkivskyi:


ilevkivskyi:
This is actually a WIP.

ilevkivskyi:
Sorry for the noise, this is now ready for review.

mypy/test/data.py
```diff
@@ -17,6 +17,11 @@
 
 root_dir = os.path.normpath(PREFIX)
 
+# Files to run with new semantic analyzer.
+new_semanal_files = [
```
JukkaL: It would be cleaner to move this definition into a separate module. Also add comment mentioning that this is a temporary hack and will be removed after we've got all test cases working with the new semantic analyzer.

ilevkivskyi:
@JukkaL I refactored as you suggested, and went ahead and switched to blacklist. Currently we have 1532 passed tests. I will make another pass in few days (after adding named tuples and builtin aliases) to see what can be added.

PR #6336, opened 2019-02-05
New semantic analyzer: generalize placeholders to all symbol nodes

### LINKED ISSUE #6299: New semantic analyzer: create dummy symbol table nodes for incomplete things
Currently we don't create any node if a definition has incomplete dependencies. This is a problem, since this can result in incorrectly falling to the a different namespace. Example:

```py
from m import ord  # Assume that this is incomplete, but m.ord is a class

x: ord  # Now this falls back to builtins.ord instead of m.ord, 
        # which may generate a bogus error
```

We should be able to fix this by adding dummy nodes to the symbol table in case we can't bind to the actual node. 

Notes: 
* On subsequent iterations a dummy node may be replaced with a real node.
* If we have a reference to a dummy node, we should always defer the current target. We shouldn't generate an error, however.
* Maybe a namespace can be complete even if some names have dummy nodes.
* Any redefinition or reassignment to a dummy node should not override the dummy node. Looks like we may need to keep track of the statement that defined a dummy node.

JukkaL:
This generalizes the concept of placeholder nodes to also cover
assignment statements and imports. Previously they were only
created for class definitions. The main motivation is to prevent 
access to definitions in outer namespaces if there is an 
incomplete definition that should take precedence.

Fixes #6299.

This also does a few other updates:

* During the final iteration, some references to placeholder nodes
  generate an error instead of producing placeholder nodes. This
  allows the analysis to terminate in certain cases of import
  cycles at least.
* Major refactoring of `analyze_name_lvalue`. The motivation is
  that some early experiments made the old structure unwieldy,
  but the refactoring may not be as important for the final design.
  This seems like a code quality improvement so I'm including it 
  here.
* If a name lvalue was bound during an earlier iteration, we don't 
  rebind it. I'd like to gradually move to this approach more 
  generally.
* Some forward references to names aren't treated as undefined 
  any more. I think that these worked by accident. Now these 
  generally generate "cannot determine type" errors.
* Most definitions won't generate incomplete namespaces any
  more, since placeholders count as definitions in this context. 
  Star imports still generate incomplete namespaces.
* Remove redundant flags from some test cases.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2258,11 +2258,6 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         kind = self.current_symbol_kind()
         names = self.current_symbol_table()
         existing = names.get(name)
-        if (existing
-                and isinstance(existing.node, PlaceholderNode)
-                and existing.node.node != lvalue):
-            # Placeholder associated with another node.
-            return
 
         outer = self.is_global_or_nonlocal(name)
         if (not existing or isinstance(existing.node, PlaceholderNode)) and not outer:
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2323,8 +2323,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             v.is_initialized_in_class = True
         if kind != LDEF:
             v._fullname = self.qualified_name(lvalue.name)
-        if kind == GDEF:
-            v.is_ready = False  # Type not inferred yet
+        v.is_ready = False  # Type not inferred yet
         lvalue.node = v
         lvalue.is_new_def = True
         lvalue.is_inferred_def = True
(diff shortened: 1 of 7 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -2223,57 +2237,45 @@ def analyze_lvalue(self, lval: Lvalue, nested: bool = False,
...
                             explicit_type: bool,
                             is_final: bool) -> None:
         """Analyze an lvalue that targets a name expression.
 
         Arguments are similar to "analyze_lvalue".
         """
-        if self.is_alias_for_final_name(lval.name):
+        if lvalue.node:
+            # This has been bound already in a previous iteration.
+            return
```
ilevkivskyi: Why do we do this for names but not for attributes? Maybe leave a comment/TODO about this.

JukkaL: I'll create an issue about dealing with attribute definitions (through `self.x`).

PR #6339, opened 2019-02-05
New semantic analyzer: clean up index expression analysis

### LINKED ISSUE #6298: New semantic analyzer: fix index expressions on non-first iterations
`visit_index_expr` may return immediately on second and later iterations, which is a problem as some incomplete references may persist. This may be a problem with type applications at least.

JukkaL:
Refactor analysis of index expressions and make two small changes:

1. Don't skip analysis if the `analyzed` attribute is defined (since there
   can be placeholder types that should be fixed).
2. Detect incomplete references when performing ordinary semantic
   analysis.

I couldn't find immediately find test cases that would benefit from the 
above changes, but I believe that it makes to make the changes just in case.

Fixes #6298.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3405,12 +3405,12 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             # We always allow unbound type variables in IndexExpr, since we
             # may be analysing a type alias definition rvalue. The error will be
             # reported elsewhere if it is not the case.
-            typearg2 = self.anal_type(typearg, allow_unbound_tvars=True,
+            analyzed = self.anal_type(typearg, allow_unbound_tvars=True,
                                       allow_placeholder=True)
-            if typearg2 is None:
+            if analyzed is None:
                 self.defer()
                 return None
-            types.append(typearg2)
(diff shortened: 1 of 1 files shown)


ethanhs:
The appveyor failure seems just to be that it didn't run appveyor on the branch, which counts as a "failure".

mypy/newsemanal/semanal.py
```diff
@@ -3347,58 +3347,74 @@ def visit_unary_expr(self, expr: UnaryExpr) -> None:
...
+    def analyze_type_application_args(self, expr: IndexExpr) -> Optional[List[Type]]:
+        """Analyze type arguments (index) in a type application.
+
+        Return None if anything was incomplete.
+        """
+        index = expr.index
+        tag = self.track_incomplete_refs()
+        self.analyze_type_expr(index)
+        if self.found_incomplete_ref(tag):
+            return None
```
ilevkivskyi: Shouldn't we also defer here before returning (or in the caller)?

JukkaL: We defer when we record an incomplete ref in `record_incomplete_ref`.

PR #6342, opened 2019-02-05
Delete pyproject.toml to work around pip bug

msullivan:
We run afoul of https://github.com/pypa/pip/issues/6163, which
prevents directly pip installing mypy from a source directory with pip
19.0 (because our setup.py imports from mypy for version number
related reasons).

Once a bugfix for that is out we can probably restore it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,0 @@
-[build-system]
-requires = [
-    "setuptools >= 30.0.2",
-    "wheel >= 0.29.0"
-]


gvanrossum:
(CC: @gaborbernat who introduced this file.)

msullivan:
Excellent. We'll revert this once there is a pip 19 release that defaults to using the legacy backend.

(We could directly ask to use it ourselves, but that seems dodgy since it has a bunch of underscores in the name and would make us require a very recent setuptools ourselves)

gaborbernat:
The main reasons for making it underscore is because importing from setup.py is considered a bad practice, not necessarily because it's actually dodgy using it.I think if we do want to really on importing from setuptools the version we might as well be upfront about it. I'll link in the discussion about the subject tomorrow.

PR #6347, opened 2019-02-06
New semantic analyzer: minor refactoring in type analyzer

JukkaL:
Use `lookup_qualified` instead of `lookup` for consistency.


ilevkivskyi:
(There is a lint error however -- long line.)

PR #6348, opened 2019-02-06
New semanatic analyzer: fix incomplete namespace with qualified refs

### LINKED ISSUE #6326: New semantic analyzer: all kinds of references correctly work with incomplete namespaces
Ensure that both non-qualified and qualified references look at the incompleteness status of the correct target namespace. This should work both in expression and type context, and in internal lookup functions when relevant.

JukkaL:
For things like `mod.name` use the incompleteness state of `mod`
instead of the current module to determine whether the reference
targets an incomplete namespace.

Also remove some unneeded code that I encountered (MRO check).

Fixes #6326.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3963,7 +3963,6 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 self.name_already_defined(n, ctx, self.globals[n])
 
     def name_not_defined(self, name: str, ctx: Context, namespace: Optional[str] = None) -> None:
-        # TODO: Reference to another namespace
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -3974,9 +3962,9 @@ def check_no_global(self, n: str, ctx: Context,
             else:
                 self.name_already_defined(n, ctx, self.globals[n])
 
-    def name_not_defined(self, name: str, ctx: Context) -> None:
+    def name_not_defined(self, name: str, ctx: Context, namespace: Optional[str] = None) -> None:
         # TODO: Reference to another namespace
```
ilevkivskyi: Is this TODO now irrelevant?

Issue #6297, opened 2019-02-01
New semantic analyzer: dummy alias definitions in typing stub (List etc.)


JukkaL:
Support definitions for built-in type aliases `List`, `Set` in the stub for `typing`. Currently these don't work, which is a problem since typeshed includes these, whereas test fixtures don't.

ilevkivskyi:
FWIW I just whitelisting some test cases to the new semantic analyzer, and it looks like many of them are blocked by this:
```
Actual:
  tmp/builtins.pyi:1: error: Module 'typing' has no attribute 'List' (diff)
  tmp/builtins.pyi:1: error: Module 'typing' has no attribute 'Set' (diff)
```
I will work on this next to unblock the situation.

JukkaL:
Note that my current working branch makes changes to how the aliases are handled.

Issue #6357, opened 2019-02-07
New semantic analyzer: visit decorators that are parts of overloads as a part of module top-level


ilevkivskyi:
Currently, only non-overloaded decorators are visited as a part of module top-level. This causes problems with detecting overloaded abstract methods.

ilevkivskyi:
This was essentially fixed by https://github.com/python/mypy/pull/6423

Issue #6356, opened 2019-02-07
Consider deferring module top level in type checker


ilevkivskyi:
We currently never defer module top levels in type checker. This can cause some `Cannot determine type of x` errors in import cycles. They are real in some sense, because such code would almost certainly fail at runtime. However, this might sometimes be useful in stubs (as practice shows, this is however not important and shows up only in highly entangled import cycles, such as e.g. builtins SCC).

ilevkivskyi:
See #6312 for an example when the error happens.

russelldavis:
> They are real in some sense, because such code would almost certainly fail at runtime.

I think it's import to distinguish the ones that would fail at runtime from the ones that would not. I think there are a lot of [common cases](https://github.com/python/mypy/issues/6312#issuecomment-461495402) that do not fail at runtime that are worth supporting.

ilevkivskyi:
With some more experience, it looks like this can cause troubles.

More generally, the deferral in type checker is different from both semantic analyzer and the daemon: the first two defer either module top level or top-level function and defer until fixed point (our large cut-off), while type checker never defers module top levels, and defers innermost functions, and have hard-coded small number of deferrals.

ilevkivskyi:
(Raising priority to high since this appeared again.)

Issue #6355, opened 2019-02-07
New semantic analyzer: Update type aliases logic


ilevkivskyi:
The old logic for type aliases will create variables in stead of aliases in case the target is class but it is not ready. An example is:
```python
Text = str

class str(Sequence[str]):
    ...
```

JukkaL:
More generally, maybe we should create a `PlaceholderNode` if the rvalue contains some placeholders.

PR #6358, opened 2019-02-07
New semantic analyzer: Support builtin typing aliases

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6297

This adds support for real typeshed stubs that define dummy aliases like `typing.List`, `typing.Dict`, etc. This also fixes couple related issues, so that builtin SCC is almost clean (the two remaining errors are https://github.com/python/mypy/issues/6295 and https://github.com/python/mypy/issues/6357).

Most notably, this PR introduces some re-ordering of targets in builtin SCC, removing this reordering requires some non-trivial work (namely https://github.com/python/mypy/issues/6356, https://github.com/python/mypy/issues/6355, and deferring targets from `named_type()`).

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -842,8 +842,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
 
         bases = defn.base_type_exprs
 
-        # TODO: Support metaclasses
-        # self.update_metaclass(defn)
+        self.update_metaclass(defn)
         bases, tvar_defs, is_protocol = self.clean_up_bases_and_infer_type_variables(bases,
                                                                                      context=defn)
         # TODO: Support keyword arguments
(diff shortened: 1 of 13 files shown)


ilevkivskyi:
Supporting `Final` depends on this, so it is better to review this one first.

mypy/newsemanal/semanal.py
```diff
@@ -348,6 +348,20 @@ def prepare_file(self, file_node: MypyFile) -> None:
                                                               self.modules['builtins'])
         if file_node.fullname() == 'builtins':
             self.prepare_builtins_namespace(file_node)
+        if file_node.fullname() == 'typing':
+            self.prepare_typing_namespace(file_node)
+
+    def prepare_typing_namespace(self, file_node: MypyFile) -> None:
+        """Remove dummy alias definitions.
```
JukkaL: This would be more concrete if with an example, such as `List = TypeAlias(object).`

mypy/newsemanal/typeanal.py
```diff
@@ -285,6 +285,12 @@ def try_analyze_special_unbound_type(self, t: UnboundType, fullname: str) -> Opt
                       " in a variable annotation", t)
             return AnyType(TypeOfAny.from_error)
         elif fullname == 'typing.Tuple':
+            # Tuple is special because it is involved in builtin import cycle
+            # and may be not ready when used.
+            sym = self.api.lookup_fully_qualified_or_none('builtins.tuple')
+            if not sym or isinstance(sym.node, PlaceholderNode):
```
JukkaL: What happens if we don't use the correct fixture that defined `tuple`? Will this still generate a reasonable error message?

ilevkivskyi: I fixed this to report a reasonable error, and added a test.

PR #6361, opened 2019-02-07
Add more time stats collection to parts of load_graph

msullivan:
load_graph has shown up very hot in internal telemetry and I want to
try to break it down a bit more.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -219,7 +219,9 @@ def _build(sources: List[BuildSource],
             TypeState.reset_all_subtype_caches()
         return BuildResult(manager, graph)
     finally:
+        t0 = time.time()
         manager.metastore.commit()
+        manager.add_stats(cache_commit_time=time.time() - t0)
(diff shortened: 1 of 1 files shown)


mypy/build.py
```diff
@@ -1587,7 +1598,9 @@ def __init__(self,
                 self.interface_hash = self.meta.interface_hash
                 self.meta_source_hash = self.meta.hash
         self.add_ancestors()
+        t0 = time.time()
         self.meta = validate_meta(self.meta, self.id, self.path, self.ignore_all, manager)
+        self.manager.add_stats(validate_meta_time=time.time() - t0)
```
ilevkivskyi: I would recommend also checking if the special case below on line 1629 takes some time (just in case).

msullivan: That only fires when not using a fine-grained cache, so I don't think it's a possibility for the cases we care about

ilevkivskyi: Ah yes, indeed, it will be taken care of during the initial update. I remember it caused big troubles before, but it was in normal incremental mode.

Issue #6271, opened 2019-01-30
Release 0.670 planning


gvanrossum:
I am planning to release mypy 0.670 on Friday, February 8. I hope to cut a release branch this Friday (Feb 1) or if that doesn't work, Monday.

Please add comments to this issue if you have any PR you'd like to be included. Ideally those PRs should be landed before the release branch is cut, to minimize cherry-picking.

gvanrossum:
Branch is cut: https://github.com/python/mypy/tree/release-0.670

(In a few hours I will know how much trouble this causes for Dropbox internal repos.)

msullivan:
I think we'll want to cherry-pick #6333

gvanrossum:
CP #6333 is done. (The internal testing still isn't completed.)

ilevkivskyi:
I think we should fix https://github.com/python/mypy/issues/6230 before the release. @JukkaL will you have time for this?

(No need to CP however, since it is a doc issue.)

oremanj:
Any chance that #6334 could make it in? Fixes an issue with plugin invocation of `__call__` that would help with a thing I'm working on.

ilevkivskyi:
I think it is relatively low risk. Although I didn't tests this against our internal codebase (if it works with it, it is definitely safe to CP).

gvanrossum:
OK, unless it causes problems with Dropbox's internal mypy runs I'll CP it.

FWIW we also need to CP #6342.

gvanrossum:
As is #6342. We'll be releasing mypy 0.670 without pyproject.toml.

JukkaL:
I'd like to include these documentation updates:
* #6345 (fixes #6230)
* #6346

gvanrossum:
This is nearly ready to go out, just waiting for the mypy-mypyc wheels.

gvanrossum:
The release is on pypi, but there was a regression (see [Gitter](https://gitter.im/python/typing)) and I've unpublished the blog post.

JukkaL:
The blog post was published again.

PR #6363, opened 2019-02-08
Update pinned mypyc version to 0.0.1+dev.231c69c6bbb3a4a9989c64974de5…

msullivan:
…1e743566fb20

This fixes a crash bug caused by https://github.com/python/typeshed/pull/2782.

It also incorporates a change disabling whole program optimization on Windows
in multi-file mode (which we use for Windows), which should speed up Windows
builds substantially.

Needs to be cherry-picked into #6271.


gvanrossum:
Cherry-picked into release-0.670.

PR #6365, opened 2019-02-09
Fix dmypy on windows when compiled with mypyc

msullivan:
dmypy currently crashes on windows when compiled with mypyc
(since 0.660; it has *never* worked, I believe).

The root cause here is the abusive use of the platform variable,
which we set to 'mypyc', to tell bogus_type.py that mypyc is being used.
This (obviously) interferes with our platform checks, causing mypyc to
spuriously think code is unreachable, but because of the minutae of
how our platform checks were structured and which files are compiled,
this only breaks things for the windows daemon.

Switch to using always_true/always_false instead of abusing platform.

There is a downside to this fix, though (which is the reason I abused platform in the first place instead of using always_true/always_false): mypy will no longer typecheck *without* its config file, since it needs `always_false = MYPYC` to be set. If this is a major problem, I'll need to look for another solution.

### DIFF
--- a/mypy/bogus_type.py
+++ b/mypy/bogus_type.py
@@ -19,7 +19,8 @@ T = TypeVar('T')
 
 # This won't ever be true at runtime, but we consider it true during
 # mypyc compilations.
-if sys.platform == 'mypyc':
+MYPYC = False
+if MYPYC:
     Bogus = FlexibleAlias[T, Any]
 else:
     Bogus = FlexibleAlias[T, T]
--- a/mypy_bootstrap.ini
+++ b/mypy_bootstrap.ini
@@ -12,5 +12,4 @@ disallow_any_unimported = True
 warn_redundant_casts = True
 warn_unused_configs = True
 show_traceback = True
-# Set the platform to something nonsense to signal to make Bogus = Any.
-platform = mypyc
+always_true = MYPYC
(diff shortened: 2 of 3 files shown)


msullivan:
If we don't want to require using `mypy_self_check.ini` for mypy to typecheck, I've got another approach using `--shadow-file` that will work.

We could also make it so that mypy understands `MYPYC` and has it default to being always false without it needing to be specified.

msullivan:
There's also a question of whether this justifies a 0.671. On one hand, it is *really* bad. On the other hand, nobody noticed it in 0.660, where it was also present.

gvanrossum:
In general our policy is, if it's not a regression it doesn't deserve a bugfix release.

I don't mind that it requires a config file to type-check -- in fact I've frequently forgotten to pass that and found that I'd made some typing mistake only when CI called it out.

msullivan:
Thinking about it more I seem to recall there might be some typeshed-related issue? @JelleZijlstra do you remember if that is the case?

JelleZijlstra:
Do you mean whether there is a typeshed issue that might justify releasing 0.671? The only problem I've seen reported is the most recent comment at python/typeshed#2760, which doesn't seem too bad.

msullivan:
No, I mean an issue with the typeshed tests if mypy only typechecks with a configuration file present

JelleZijlstra:
Oh, sorry. I'm not aware of anything specific, but the typeshed tests do run some of the mypy test suite, including the self check. If anything breaks there's it's surely fixable by tweaking the Travis configuration.

msullivan:
All right I think I want to merge this

PR #6368, opened 2019-02-09
Improvements to stubs generated from docstrings

wiktorn:
Few improvements/bugfixes for stub generation from docstrings:
1. Do not generate stubs with improper type names. For now it's just regex check
2. Only generate stubs for methods that have unique argument names
3. Fix missing defaults in stubs generated from docstrings

### DIFF
--- a/mypy/stubdoc.py
+++ b/mypy/stubdoc.py
@@ -22,9 +22,14 @@ Sig = Tuple[str, str]
 
 class ArgSig:
     """Signature info for a single argument."""
+
+    _TYPE_RE = re.compile(r'^[a-zA-Z_][\w\[\], ]*(\.[a-zA-Z_][\w\[\], ]*)*$')
+
     def __init__(self, name: str, type: Optional[str] = None, default: bool = False):
         self.name = name
-        self.type = type
+        if type and not self._TYPE_RE.match(type):
+            raise ValueError("Invalid type: " + type)
+        self.type = type if type and self._TYPE_RE.match(type) else None
         # Does this argument have a default value?
         self.default = default
 
@@ -129,8 +134,13 @@ class DocStringParser:
 
             if token.string == ')':
                 self.state.pop()
-            self.args.append(ArgSig(name=self.arg_name, type=self.arg_type,
-                                    default=bool(self.arg_default)))
+            try:
+                self.args.append(ArgSig(name=self.arg_name, type=self.arg_type,
+                                        default=bool(self.arg_default)))
+            except ValueError:
+                # wrong type, use Any
+                self.args.append(ArgSig(name=self.arg_name, type=None,
+                                        default=bool(self.arg_default)))
             self.arg_name = ""
             self.arg_type = None
             self.arg_default = None
--- a/mypy/stubdoc.py
+++ b/mypy/stubdoc.py
@@ -5,6 +5,7 @@ docstrings and Sphinx docs (.rst files).
 """
 import re
 import io
+import sys
 import contextlib
 import tokenize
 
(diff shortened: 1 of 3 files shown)


wiktorn:
What I'm missing right now in stubgen is infrastructure to emit warnings. Should I just print it on stderr?

msullivan:
Just printing on stderr should be fine for stubgen for now

mypy/stubdoc.py
```diff
@@ -22,9 +22,14 @@
...
     """Signature info for a single argument."""
+
+    _TYPE_RE = re.compile(r'^[a-zA-Z_][\w\[\], ]*(\.[a-zA-Z_][\w\[\], ]*)*$')
+
     def __init__(self, name: str, type: Optional[str] = None, default: bool = False):
         self.name = name
-        self.type = type
+        if type and not self._TYPE_RE.match(type):
+            raise ValueError("Invalid type: " + type)
+        self.type = type if type and self._TYPE_RE.match(type) else None
```
gvanrossum: This smells redundant. Why match twice?

mypy/stubdoc.py
```diff
@@ -193,7 +203,15 @@ def infer_sig_from_docstring(docstr: str, name: str) -> Optional[List[FunctionSi
     with contextlib.suppress(tokenize.TokenError):
         for token in tokenize.tokenize(io.BytesIO(docstr.encode('utf-8')).readline):
             state.add_token(token)
-    return state.get_signatures()
+    ret = state.get_signatures()
+
+    def is_unique_args(sig: FunctionSig) -> bool:
+        """return true if function argument names are unique"""
+        return len(sig.args) == len(set((x.name for x in sig.args)))
```
gvanrossum: `x` is a rather unimaginative variable name. How about `arg`?

mypy/stubdoc.py
```diff
@@ -193,7 +203,15 @@ def infer_sig_from_docstring(docstr: str, name: str) -> Optional[List[FunctionSi
...
+
+    # return only signatures, that have unique argument names. mypy fails on non-uqniue arg names
+    # TODO: emit a warning for duplicate argument names
+    return [x for x in ret if is_unique_args(x)]
```
gvanrossum: Ditto, how about `sig`? (Actually `ret` isn't such a great name either -- maybe `sigs`?)

mypy/test/teststubgen.py
```diff
@@ -218,6 +218,19 @@ def test_infer_sig_from_docstring(self) -> None:
...
+        assert_equal(infer_sig_from_docstring('\nfunc(x: invalid::type<with_template>)', 'func'),
+                     [FunctionSig(name='func', args=[ArgSig(name='x', type=None)],
+                                  ret_type='Any')])
+
+        assert_equal(infer_sig_from_docstring('\nfunc(x: str="")', 'func'),
+                     [FunctionSig(name='func', args=[ArgSig(name='x', type='str', default=True)],
+                                  ret_type='Any')])
+
+    def test_infer_sig_from_docstring_duplicate_args(self) -> None:
+        assert_equal(infer_sig_from_docstring('\nfunc(x, x) -> str\nfunc(x, y) -> int', 'func'),
```
gvanrossum: I take it there's some kind of bug in pybind11 that sometimes generates `func(x, x)`?

wiktorn: There is a bug in pybind11 that sometimes generates signatures using C++ types instead of Python types. As I tried some sanitizations I got errors for duplicate arguments. Though right now those invalid signatures are just dropped, so this is not longer the case

Issue #1271, opened 2016-03-04
Stricter equality checks?


averagehat:
It would be nice if there was a way to enforce a constraint on `__eq__` -- maybe a warning? I almost never want to do the following

``` python
def test_operator(): # type: () -> None
    1 == "1"
# produces no complaints
```

While that check may make sense in a world of vanilla python, I'd argue it doesn't make much sense within mypy's world (especially if the arguments are explicitly typed as different types; i.e., one is not a subtype of the other).

gvanrossum:
Hm, I think you're right this should be flagged if the types don't overlap.

JukkaL:
The operators `!=`, `is` and `is not` are similar. Also, mypy doesn't complain about things like `1 in ['x']` since it doesn't raise a runtime exception.

wcooley:
Attempting to specify a type other than `Any` or `object` with `__eq__` results in an error message like this:

    foo.py: note: In class "Foo":
    foo.py:4: error: Argument 1 of "__eq__" incompatible with supertype "object"

(I'm adding this in case someone else is searching for this error.)

dgoldstein0:
+1 would love to see this - had to troubleshoot a bug today that turned out to a stray `,` making an equality check fail because one side had accidentally become a tuple.

ilevkivskyi:
It looks like this appears often, so I raise priority to high.

nharkins:
+1, would love this feature. just had a bug with a missing parens, i.e. `if func != 0:`

jstasiak:
Kind of similar to the above: I did forget a few times that `obj.thing` is a method, not a property, wrote `if obj.thing:` and got an always-on block of code.

efroemling:
+1 would love to see this too.  I'm replacing a bunch of old string values with more type-safe enum objects and would love to get errors in places where my code needs updating, such as:  if obj.enum_value == 'my_old_hard_coded_string_value'

ilevkivskyi:
@jstasiak Your case is a bit different, it is like https://github.com/python/mypy/issues/2073

PR #6370, opened 2019-02-10
Add flag to prohibit equality checks between non-overlapping checks

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/1271

The new per-file flag `--strict-equality` disables equality, identity, and container checks between non-overlapping types. In general the implementation is straightforward. Here are some corner cases I have found:
* `Any` should be always safe.
* Type promotions should be ignored, `b'abc == 'abc'` should be an error.
* Checks like `if x is not None: ...`, are special cased to not be errors if `x` has non-optional type.
* `Optional[str]` and `Optional[bytes]` should be considered non-overlapping for the purpose of this flag.
* For structural containers and custom `__eq__()` (i.e. incompatible with `object.__eq__()`) I suppress the non-overlapping types error, if there is already an error originating from the method call check.

Note that I updated `typing-full.pyi` so that `Sequence` inherits from `Container` (this is needed by some added tests, and also matches real stubs). This however caused necessary changes in a bunch of builtins fixtures.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1933,7 +1933,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                         self.bool_type(),
                         self.named_type('builtins.function'))
                     if not is_subtype(left_type, itertype):
-                        self.msg.unsupported_operand_types('in', left_type, itertype, e)
+                        self.msg.unsupported_operand_types('in', left_type, right_type, e)
                 # Only show dangerous overlap if there are no other errors.
                 elif (not local_errors.is_errors() and cont_type and
                         self.dangerous_comparison(left_type, cont_type)):
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -44,7 +44,7 @@ from mypy.messages import MessageBuilder
 from mypy import message_registry
 from mypy.infer import infer_type_arguments, infer_function_type_arguments
 from mypy import join
-from mypy.meet import narrow_declared_type
+from mypy.meet import narrow_declared_type, is_overlapping_types
 from mypy.subtypes import (
     is_subtype, is_proper_subtype, is_equivalent, find_member, non_method_protocol_members,
 )
@@ -1914,6 +1914,11 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 _, method_type = self.check_method_call_by_name(
                     '__contains__', right_type, [left], [ARG_POS], e, local_errors)
                 sub_result = self.bool_type()
+                # Container item type for strict type overlap checks. Note: we need to only
+                # check for nominal type, because a usual "Unsupported operands for in"
+                # will be reported for types incompatible with __contains__().
+                # See testCustomContainsCheckStrictEquality for an example.
+                cont_type = self.chk.analyze_container_item_type(right_type)
                 if isinstance(right_type, PartialType):
                     # We don't really know if this is an error or not, so just shut up.
                     pass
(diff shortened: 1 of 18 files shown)


msullivan:
> Interestingly, this flag cause couple dozen errors in mypy itself, vast majority or errors are because of
> 
> ```python
> obj: Instance
> if obj is not None:
>     ...
> ```
> 

In some sense that is a good thing, right, since we are doing a bad thing in those places? On the other hand, there are a lot of them, so...

test-data/unit/check-expressions.test
```diff
@@ -604,7 +604,7 @@ class X:
 class Y:
     def __lt__(self, o: 'Y') -> A: pass
     def __gt__(self, o: 'Y') -> A: pass
-    def __eq__(self, o: 'Y') -> B: pass
+    def __eq__(self, o: 'Y') -> B: pass  # type: ignore
```
msullivan: Why is this needed?

mypy/checkexpr.py
```diff
@@ -1954,6 +1972,26 @@ def visit_comparison_expr(self, e: ComparisonExpr) -> Type:
...
+
+        Rules:
+            * X and None are non-overlapping in strict-optional mode, and
+            overlapping otherwise.
+            * Optional[X] and Optional[Y] are non-overlapping if X and Y are
+            non-overlapping, although technically None is overlap, it is most
+            likely an error.
+            * Any overlaps with everything, i.e. always safe.
+            * Promotions are ignored, so both 'abc' == b'abc' and 1 == 1.0
+            are errors.
```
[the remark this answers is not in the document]
ilevkivskyi: As discussed off-tracker we will keep ignoring promotions, mostly for bytes vs unicode.

mypy/checkexpr.py
```diff
@@ -1954,6 +1972,26 @@ def visit_comparison_expr(self, e: ComparisonExpr) -> Type:
...
+        """
+        if isinstance(left, UnionType) and isinstance(right, UnionType):
```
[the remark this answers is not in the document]
ilevkivskyi: Fixed.

test-data/unit/check-expressions.test
ilevkivskyi: I updated the fixture (to be more like in typeshed) and this now causes incompatible with supertype.

ilevkivskyi:
> This disallows things like `assert foo is not None`

OK, also as discussed of tracker, let's start with allowing this asserts (otherwise even mypy codebase wouldn't be clean with this flag).

ilevkivskyi:
@msullivan
Thanks for review! This is ready for another round.

mypy/meet.py
```diff
@@ -221,9 +221,32 @@ def is_none_typevar_overlap(t1: Type, t2: Type) -> bool:
...
 
     if isinstance(left, TypeType) and isinstance(right, TypeType):
-        # TODO: Can Callable[[...], T] and Type[T] be partially overlapping?
         return _is_overlapping_types(left.item, right.item)
 
+    # Type[C] vs Callable[..., C], where the latter is class object.
+    if isinstance(left, CallableType) and left.is_type_obj() and isinstance(right, TypeType):
+        if _is_overlapping_types(left.ret_type, right.item):
+            return True
+    if isinstance(right, CallableType) and right.is_type_obj() and isinstance(left, TypeType):
```
msullivan: I'd say define a nested function for doing the `Type[C]` vs. whatever checks instead of duplicating the logic for the left and right cases.

JukkaL:
@ilevkivskyi I think that `int`/`float` comparison should be allowed, since they are safe and disallowing them may make this feature much less useful for code that uses lots of floats. Can you create a follow-up issue about this? It may be fine to reject `str` vs. `unicode` comparisons, but even there it's a bit questionable, since `str` and `unicode` can be freely mixed in other contexts due to the promotion. Why do we disallow this only for this particular operation?

ilevkivskyi:
> It may be fine to reject `str` vs. `unicode` comparisons, but even there it's a bit questionable, since `str` and `unicode` can be freely mixed in other contexts due to the promotion. Why do we disallow this only for this particular operation?

This request came many times from internal users, since such comparisons caused actual troubles. One of the problems is that `b'abc' == 'abc'` is `True` in Python 2, but `False` in Python 3.

> I think that `int`/`float` comparison should be allowed, since they are safe and disallowing them may make this feature much less useful for code that uses lots of floats.

As someone who used floats a lot, I think comparing floats with `==` is a bad idea, because of traps like `(1/3)**10 * 3**10 == 1` or `(2**0.1)**10 == 2`, etc. They should be compared with something like `math.isclose()`, `numpy.allclose()` or similar.

As a bottom line, I still think prohibiting promotions with this flag is a good idea. Potentially, we can add a note why this is unsafe.

JukkaL:
Thanks for the detailed explanation -- you convinced me. This is a great feature as it is now and will help catch many bugs.

Issue #5910, opened 2018-11-17
`get_attribute_hook` for unknown attributes


mkurnikov:
How do I intercept type inference of the `settings.ROOT_DIR` here? 
```
from django.conf import settings
settings.ROOT_DIR
```

I wasn't able to get it working neither with `get_attribute_hook`, nor with others. `settings` is specified as 
```
# in django/conf/__init__.pyi
class LazySettings(object):
    pass
settings = LazySettings()
```

From the implementation, it looks like hooks's only called if attribute is present on the object (is a Var)
https://github.com/python/mypy/blob/master/mypy/checkmember.py#L280

Another solution for this problem could be to augment the AST with an additional settings dynamically somewhere in the plugin, but for that I need an initialization hook.

ilevkivskyi:
You can stuff various dummy names in the class symbol table using `get_class_decorator_hook()`, `get_metaclass_hook()` and `get_base_class_hook()` (plus something in stubs). But I think it probably makes sense to call the hook in more cases there (e.g. if it comes from `__getattr__()` etc.)

ilevkivskyi:
Fixed by https://github.com/python/mypy/pull/6371

PR #6371, opened 2019-02-10
Use the get_attribute hook for dynamic classes

### LINKED ISSUE #6259: Call plugin get_attribute_hook() also for attributes that trigger __getattr__()
Currently, `get_attribute_hook()` is only called for attributes that are already present in the class symbol table. However, the typical use case is opposite -- dynamically added attributes. Therefore, the current behavior creates some complications while developing plugins, like every access to dynamic attribute first requires somehow placing it into the class symbol table using other hooks. I propose to widen the scope of `get_attribute_hook()` to at least be called for classes with `__getattr__()`.

lincolnq:
Whenever we find `__getattr__` or `__getattribute__` on an instance, the getattribute hook is called to give the hook a chance to modify the return type of one of those functions.

Closes #6259, #5910

(Note: Currently there are no tests. I can add them if important.)

### DIFF
--- a/docs/source/extending_mypy.rst
+++ b/docs/source/extending_mypy.rst
@@ -184,7 +184,11 @@ For example in this code:
 mypy will call ``get_method_signature_hook("ctypes.Array.__setitem__")``
 so that the plugin can mimic the ``ctypes`` auto-convert behavior.
 
-**get_attribute_hook** overrides instance member field lookups and property access (not assignments, and not method calls). This hook is only called for fields which already exist on the class. *Exception:* if ``__getattr__`` or ``__getattribute__`` is a method on the class, the hook is called for all fields which do not refer to methods.
+**get_attribute_hook** overrides instance member field lookups and property
+access (not assignments, and not method calls). This hook is only called for
+fields which already exist on the class. *Exception:* if ``__getattr__`` or
+``__getattribute__`` is a method on the class, the hook is called for all
+fields which do not refer to methods.
 
 **get_class_decorator_hook()** can be used to update class definition for
 given class decorators. For example, you can add some attributes to the class
--- a/docs/source/extending_mypy.rst
+++ b/docs/source/extending_mypy.rst
@@ -184,11 +184,7 @@ For example in this code:
 mypy will call ``get_method_signature_hook("ctypes.Array.__setitem__")``
 so that the plugin can mimic the ``ctypes`` auto-convert behavior.
 
-**get_attribute_hook** can be used to give more precise type of an instance
-attribute. Note however, that this method is only called for variables that
-already exist in the class symbol table. If you want to add some generated
-variables/methods to the symbol table you can use one of the three hooks
-below.
+**get_attribute_hook** overrides instance member field lookups and property access (not assignments, and not method calls). This hook is only called for fields which already exist on the class. *Exception:* if ``__getattr__`` or ``__getattribute__`` is a method on the class, the hook is called for all fields which do not refer to methods.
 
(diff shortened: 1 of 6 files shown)


mypy/checkmember.py
```diff
@@ -340,7 +340,15 @@ def analyze_member_var_access(name: str,
                     typ = map_instance_to_supertype(itype, method.info)
                     getattr_type = expand_type_by_instance(bound_method, typ)
                     if isinstance(getattr_type, CallableType):
-                        return getattr_type.ret_type
+                        result = getattr_type.ret_type
+
+                        # call the get attribute hook before returning
```
ilevkivskyi:
```suggestion
                        # Call the attribute hook before returning.
```

docs/source/extending_mypy.rst
```diff
@@ -184,11 +184,7 @@ For example in this code:
 mypy will call ``get_method_signature_hook("ctypes.Array.__setitem__")``
 so that the plugin can mimic the ``ctypes`` auto-convert behavior.
 
-**get_attribute_hook** can be used to give more precise type of an instance
-attribute. Note however, that this method is only called for variables that
-already exist in the class symbol table. If you want to add some generated
-variables/methods to the symbol table you can use one of the three hooks
-below.
+**get_attribute_hook** overrides instance member field lookups and property access (not assignments, and not method calls). This hook is only called for fields which already exist on the class. *Exception:* if ``__getattr__`` or ``__getattribute__`` is a method on the class, the hook is called for all fields which do not refer to methods.
```
ilevkivskyi: Please split this huge line into separate lines (around 80 columns per line).

lincolnq:
@ilevkivskyi Thanks for the review! I made the changes you requested.

One of the test cases is that the attribute hook can produce an error. I invented my own error, "Field does not exist", but ideally I would test that you can fallback to the normal mypy error handling -- but I don't think you can access a MemberContext to call `mx.msg.has_no_attr`. Is the way I did it ok?

lincolnq:
Hm, the old docstring was wrong too (it specifies that the hook will be
called with the base class name where the field is defined, but it is
called with the derived class name). I assume the documented behavior is
the desirable one?

test-data/unit/check-custom-plugin.test
```diff
@@ -143,19 +143,49 @@ tmp/mypy.ini:2: error: Return value of "plugin" must be a subclass of "mypy.plug
...
+
+ds: DerivedSignal[Callable[[int], None]] = DerivedSignal()
+ds('') # E: Argument 1 has incompatible type "str"; expected "int"
 [file m.py]
 from typing import TypeVar, Generic, Callable
 T = TypeVar('T', bound=Callable[..., None])
 class Signal(Generic[T]):
     __call__: Callable[..., None]  # This type is replaced by the plugin
+
+class DerivedSignal(Signal, Generic[T]): ...
```
ilevkivskyi: I think this should be `Signal[T]`, this will fix the test.

ilevkivskyi:
> I assume the documented behavior is the desirable one?

Yes, I think giving the name of the base class where `__getattr__()` was defined would be consistent. So you probably need to update your code. (The fix for the old test I posted above.)

test-data/unit/check-custom-plugin.test
ilevkivskyi: Hm, it looks like variables (unlike methods) are passed with non-mapped instance type, so your fix is right. Anyway, it i better to change this to `Signal[T]` (currently it is equivalent to `Signal[Any]`).

Issue #6290, opened 2019-02-01
New semantic analyzer: Final


JukkaL:
In particular, the current implementation of stripping away `Final[...]` may not work if there are multiple iterations.

ilevkivskyi:
This is one of the blockers for mypy self-check so raising priority to high.

PR #6381, opened 2019-02-12
New semantic analyzer: Support Final

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6290

I also fix couple minor bugs necessary to enable existing test file and fix final-related errors in mypy self-check with new analyzer.


mypy/newsemanal/semanal.py
```diff
@@ -1656,11 +1657,6 @@ def visit_import_from(self, imp: ImportFrom) -> None:
                     self.add_symbol(imported_id, gvar, imp)
                     continue
             if node and node.kind != UNBOUND_IMPORTED and not node.module_hidden:
-                if not node:
-                    # Normalization failed because target is not defined. Avoid duplicate
-                    # error messages by marking the imported name as unknown.
-                    self.add_unknown_symbol(imported_id, imp, is_import=True)
-                    continue
```
ilevkivskyi: This is dead code.

test-data/unit/pythoneval.test
```diff
@@ -1365,7 +1365,8 @@ from typing import Dict, List, Tuple
 x: Dict[str, List[int]]
 reveal_type(x['test'][0])
 [out]
-typing.pyi: error: Class typing.Sequence has abstract attributes "__len__"
-typing.pyi: note: If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass
+typing.pyi:251: error: Class typing.Sequence has abstract attributes "__len__"
+typing.pyi:251: note: If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass
+ast.pyi:31: error: Name 'PyCF_ONLY_AST' already defined (possibly by an import)
```
ilevkivskyi: This is an actual bug in typeshed, this is name is indeed already defined by a star import.

mypy/newsemanal/semanal.py
```diff
@@ -3937,7 +3945,7 @@ def is_global_or_nonlocal(self, name: str) -> bool:
                      or name in self.nonlocal_decls[-1]))
 
     def add_symbol_table_node(self, name: str, symbol: SymbolTableNode,
-                              context: Optional[Context] = None) -> None:
+                              context: Optional[Context] = None) -> bool:
         """Add symbol table node to the currently active symbol table."""
```
JukkaL: Document the return value here as well.

mypy/newsemanal/semanal.py
```diff
@@ -4035,18 +4045,23 @@ def name_not_defined(self, name: str, ctx: Context, namespace: Optional[str] = N
...
+        else:
+            node = None
 
         if isinstance(original_ctx, SymbolTableNode) and isinstance(original_ctx.node, MypyFile):
             # Since this is an import, original_ctx.node points to the module definition.
             # Therefore its line number is always 1, which is not useful for this
             # error message.
             extra_msg = ' (by an import)'
-        elif node and node.line != -1:
+        elif node and node.line != -1 and node.fullname().startswith(self.cur_mod_id):
```
JukkaL: A `startswith` check is not reliable. For example, `foo` is a prefix of `foobar`, even though they are unrelated names.

PR #6389, opened 2019-02-13
Add a mechanism for avoiding hashing sources on daemon start

msullivan:
This has two parts:
 * Add an argument to dmypy status (`--fswatcher-dump-file JSONFILE`) that tells the daemon to
   write out the state of fswatcher to a file.
 * Allow mypy/dmypy to use this information to avoid hashing source
   files if the mtime in the cache disagrees with the disk but
   the old fswatcher info agrees. (`--quickstart-file JSONFILE`)

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -910,8 +910,8 @@ def read_quickstart_file(options: Options) -> Optional[Dict[str, Tuple[float, in
             quickstart = {}
             for file, (x, y, z) in raw_quickstart.items():
                 quickstart[file] = (x, y, z)
-        except Exception:
-            pass
+        except Exception as e:
+            print("Warning: Failed to load quickstart file: {}\n".format(str(e)))
     return quickstart
 
 
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -906,11 +906,12 @@ def read_quickstart_file(options: Options) -> Optional[Dict[str, Tuple[float, in
         try:
             with open(options.quickstart_file, "r") as f:
                 raw_quickstart = json.load(f)
+
+            quickstart = {}
+            for file, (x, y, z) in raw_quickstart.items():
+                quickstart[file] = (x, y, z)
         except Exception:
             pass
-        quickstart = {}
-        for file, (x, y, z) in raw_quickstart.items():
-            quickstart[file] = (x, y, z)
     return quickstart
 
 
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -472,6 +472,9 @@ class BuildManager(BuildManagerBase):
                        but is disabled if fine-grained cache loading fails
                        and after an initial fine-grained load. This doesn't
                        determine whether we write cache files or not.
+      quickstart_state:
+                       A cache of filename -> mtime/size/hash info used to avoid
+                       needing to hash source files when using a cache with mismatching mtimes
       stats:           Dict with various instrumentation numbers, it is used
                        not only for debugging, but also required for correctness,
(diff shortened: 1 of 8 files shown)


test-data/unit/daemon.test
```diff
@@ -138,3 +138,22 @@ mypy-daemon: error: Missing target module, package, files, or command.
 == Return code: 2
 $ dmypy stop
 Daemon stopped
+
+-- this is carefully constructed to be able to break if the quickstart system lets
+-- something through incorrectly. in particular, the files need to ahve the same size
```
ethanhs:
```suggestion
-- something through incorrectly. in particular, the files need to have the same size
```

mypy/build.py
```diff
@@ -893,6 +897,23 @@ def read_plugins_snapshot(manager: BuildManager) -> Optional[Dict[str, str]]:
     return snapshot
 
 
+def read_quickstart_file(options: Options) -> Optional[Dict[str, Tuple[float, int, str]]]:
+    quickstart = None  # type: Optional[Dict[str, Tuple[float, int, str]]]
+    if options.quickstart_file:
+        # This is very "best effort". If the file is missing or malformed,
```
ethanhs: I feel like we should at least warn on a bad file. That will also allow our tests to catch bugs.

gvanrossum: Hm, I actually agree: warn but continue.

test-data/unit/daemon.test
```diff
@@ -138,3 +138,22 @@ mypy-daemon: error: Missing target module, package, files, or command.
...
+Daemon is up and running
+$ dmypy stop
+Daemon stopped
+-- sleep guarantees timestamp changes
+$ {python} -c "import time;time.sleep(1)"
+$ {python} -c "print('lol')" >foo.py
+$ dmypy run -- foo.py --follow-imports=error --use-fine-grained-cache --quickstart-file test.json
+Daemon started
+foo.py:1: error: Name 'lol' is not defined
+== Return code: 1
```
gvanrossum: So this test doesn't check that the quickstart does anything, right? It just checks that the behavior is the same.

mypy/build.py
msullivan: mypy isn't very good at warning, though. We can label something as a warning, but it will still cause an error exit code.

We could just do a print, but then that will end up in the log file for the daemon (which is maybe fine?).

msullivan:
> This looks good, though I had to read the comment you added to `validate_meta()` three times before I got it.

Do you have any suggestions on language revision to that comment that might make it more clear?

mypy/build.py
gvanrossum: `sys.stderr.write("warning: Blah\n")` sounds acceptable, assuming that will be visible when we run tests?

gvanrossum:
> Do you have any suggestions on language revision to that comment that might make it more clear?

Alas, not -- it's just complicated.

mypy/build.py
ethanhs: Yeah the daemon tests capture stderr so it should be fine.

test-data/unit/daemon.test
msullivan: Yeah. I've updated the test to be able to make sure that a matching file doesn't have its metadata updated, which is a decent proxy for the thing we really care about

mypy/build.py
msullivan: It can be made visible by having the test dump the log file

msullivan:
@gvanrossum Anything else, or is this good?

PR #6390, opened 2019-02-13
New semantic analyzer: fix corner cases in aliases to not ready classes (and other aliases)

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6355

This does three non-trivial things:
* Creates placeholder with `becomes_typeinfo=True` if r.h.s. is such a placeholder (i.e. the definition is a potential type alias).
* Defers all `Invalid type` and use of corresponding unbound types until final iteration.
* Moves few tests in newly enabled test file to new semantic analyzer only (those where new behaviour is clearly better).

I added a bunch of new tests for things that was broken, plus some variations (although some of them are trivial, replacing `NameExpr` with `MemberExpr` or `IndexExpr`).

I also verified that this PR fixes the crash in mypy self-check.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -1886,16 +1886,24 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             target = self.scope.current_target()
         self.cur_mod_node.alias_deps[target].update(aliases_used)
 
+    def can_be_an_alias(self, rv: Expression) -> bool:
+        if isinstance(rv, NameExpr):
+            n = self.lookup(rv.name, rv)
+            if n and isinstance(n.node, PlaceholderNode) and n.node.becomes_typeinfo:
+                return True
+        return False
+
     def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
         s.is_final_def = self.unwrap_final(s)
         tag = self.track_incomplete_refs()
         s.rvalue.accept(self)
-        if self.found_incomplete_ref(tag):
+        is_alias = self.can_be_an_alias(s.rvalue)
+        if self.found_incomplete_ref(tag) or is_alias:
             # Initializer couldn't be fully analyzed. Defer the current node and give up.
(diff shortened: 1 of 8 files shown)


ilevkivskyi:
There is actually bug in the logic, so marking as WIP.

ilevkivskyi:
OK, this is now ready for review. I will create couple follow-up issues soon (for tricky corner cases, and possibly re-thinking the logic to avoiding intermediate `Var`s).

mypy/newsemanal/semanal.py
```diff
@@ -1856,16 +1859,42 @@ def add_type_alias_deps(self, aliases_used: Iterable[str],
...
+            # Union['ClassA', 'ClassB']
+            self.analyze_alias(s.rvalue, allow_placeholder=True)
+        is_alias = self.is_not_ready_alias(s.rvalue)
+        if self.found_incomplete_ref(tag) or is_alias:
             # Initializer couldn't be fully analyzed. Defer the current node and give up.
             # Make sure that if we skip the definition of some local names, they can't be
             # added later in this scope, since an earlier definition should take precedence.
             for expr in names_modified_by_assignment(s):
-                self.mark_incomplete(expr.name, expr)
+                self.mark_incomplete(expr.name, expr, becomes_typeinfo=is_alias)
```
JukkaL: What if the alias is like `X = List[Y]` or `X = C[Y]` where only `Y` is not ready? It's not clear to me how we'd set `becomes_typeinfo` correctly in that case.

ilevkivskyi: In this case we can turn them into `Var`s, this is what I mentioned in private chat. In general, I don't like the current situation, I will try a bigger change, maybe there is a way to make it more straightforward.

ilevkivskyi:
@JukkaL As discussed, I added the clarifying comments. I will merge this now and open a follow-up issue about refactoring the analysis of assignment statements.

PR #6391, opened 2019-02-13
Add a flag to skip mtime checks on cache files

msullivan:
When using an externally generated sqlite cache in daemon mode,
we should already trust that the cache is valid, and all the mtime
checks are surprisingly slow. (About 1s on S on my MBP.)

Allow skipping these.

An alternative I considered was to just always skip them when using
a sqlite cache, but I think we'd need to be more careful about
when we commit if we want to guarentee cache integrity.


msullivan:
Yeah

PR #6392, opened 2019-02-13
Remove a mypyc hack that became a liability

msullivan:
When removed the sys.platform hack, the dummy `bool()` in
a unreachable code hack became, itself, unreachable.

And while mypyc is much better about unreachable branches that it used
to be, it still crashes on unreachable parts of conditional
expressions.


msullivan:
This will fix OS X mypy-mypyc builds, which I just broke in #6365

PR #6395, opened 2019-02-14
New semantic analyzer: check generic type arguments

### LINKED ISSUE #6288: New semantic analyzer: check generic type arguments
Check type arguments against bounds/value restriction. For example, check that `int` is valid in `C[int]`.

This may need to happen in a distinct pass once we can have incomplete/placeholder types.

JukkaL:
This adds a new pass that checks things like `int` being valid in
`C[int]`. We use a new pass since base classes can be incomplete
during semantic analysis, which breaks subtype checks that we 
rely on here.

The implementation abstracts AST traversal into separate classes,
so that the implementation of the pass is relatively simple.

Fixes #6288.

### DIFF
--- a/mypy/mixedtraverser.py
+++ b/mypy/mixedtraverser.py
@@ -1,0 +1,89 @@
+from typing import Optional
+
+from mypy.nodes import (
+    MypyFile, Var, FuncItem, ClassDef, AssignmentStmt, OperatorAssignmentStmt, ForStmt, WithStmt,
+    CastExpr, TypeApplication, TypeAliasExpr, TypeVarExpr, TypedDictExpr, NamedTupleExpr,
+    EnumCallExpr, PromoteExpr, NewTypeExpr
+)
+from mypy.types import Type
+from mypy.traverser import TraverserVisitor
+from mypy.typetraverser import TypeTraverserVisitor
+
+
+class MixedTraverserVisitor(TraverserVisitor, TypeTraverserVisitor):
+    """Recursive traversal of both Node and Type objects."""
+
+    # Symbol nodes
+
+    def visit_var(self, var: Var) -> None:
+        self.visit_optional_type(var.type)
+
+    def visit_func(self, o: FuncItem) -> None:
+        super().visit_func(o)
+        self.visit_optional_type(o.type)
+
+    def visit_class_def(self, o: ClassDef) -> None:
+        super().visit_class_def(o)
+        info = o.info
+        if info:
+            for base in info.bases:
+                base.accept(self)
+
+    def visit_type_alias_expr(self, o: TypeAliasExpr) -> None:
+        super().visit_type_alias_expr(o)
+        # TODO
+
+    def visit_type_var_expr(self, o: TypeVarExpr) -> None:
+        super().visit_type_var_expr(o)
+        # TODO
+
+    def visit_typeddict_expr(self, o: TypedDictExpr) -> None:
+        super().visit_typeddict_expr(o)
+        # TODO
+
+    def visit_namedtuple_expr(self, o: NamedTupleExpr) -> None:
+        super().visit_namedtuple_expr(o)
+        # TODO
+
+    def visit_enum_call_expr(self, o: EnumCallExpr) -> None:
+        super().visit_enum_call_expr(o)
+        # TODO
+
+    def visit__promote_expr(self, o: PromoteExpr) -> None:
+        super().visit__promote_expr(o)
+        # TODO
+
+    def visit_newtype_expr(self, o: NewTypeExpr) -> None:
+        super().visit_newtype_expr(o)
+        # TODO
+
+    # Statements
+
+    def visit_assignment_stmt(self, o: AssignmentStmt) -> None:
+        super().visit_assignment_stmt(o)
+        self.visit_optional_type(o.type)
+
+    def visit_for_stmt(self, o: ForStmt) -> None:
+        super().visit_for_stmt(o)
+        self.visit_optional_type(o.index_type)
+
+    def visit_with_stmt(self, o: WithStmt) -> None:
+        super().visit_with_stmt(o)
+        self.visit_optional_type(o.target_type)
+
+    # Expressions
+
+    def visit_cast_expr(self, o: CastExpr) -> None:
+        super().visit_cast_expr(o)
+        o.type.accept(self)
+
+    def visit_type_application(self, o: TypeApplication) -> None:
+        super().visit_type_application(o)
+        for t in o.types:
+            t.accept(self)
+
+    # Helpers
+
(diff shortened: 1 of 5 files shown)


mypy/newsemanal/semanal_typeargs.py
```diff
@@ -0,0 +1,79 @@
...
+                                      arg.name, info.name()), t)
+                        continue
+                else:
+                    arg_values = [arg]
+                self.check_type_var_values(info, arg_values, tvar.name, tvar.values, i + 1, t)
+            if not is_subtype(arg, tvar.upper_bound):
+                self.fail('Type argument "{}" of "{}" must be '
+                          'a subtype of "{}"'.format(
+                              arg, info.name(), tvar.upper_bound), t)
+        super().visit_instance(t)
```
ilevkivskyi: The old third pass not just checked the argument types, but also argument counts. For example it replaced `List[int, str]` with `List[Any]` and emitted an error. Why don't we also do this here?

mypy/mixedtraverser.py
```diff
@@ -0,0 +1,89 @@
...
+        assert o.info.tuple_type
+        o.info.tuple_type.accept(self)
+
+    def visit__promote_expr(self, o: PromoteExpr) -> None:
+        super().visit__promote_expr(o)
+        o.type.accept(self)
+
+    def visit_newtype_expr(self, o: NewTypeExpr) -> None:
+        super().visit_newtype_expr(o)
+        self.visit_optional_type(o.old_type)
```
ilevkivskyi: The type that appears in `bases` of the generated `TypeInfo` may be a copy of this one, we had a crash because of this a while ago.

mypy/mixedtraverser.py
```diff
@@ -0,0 +1,89 @@
...
+        super().visit_func(o)
+        self.visit_optional_type(o.type)
+
+    def visit_class_def(self, o: ClassDef) -> None:
+        super().visit_class_def(o)
+        info = o.info
+        if info:
+            for base in info.bases:
+                base.accept(self)
+            self.visit_optional_type(info.typeddict_type)
```
ilevkivskyi: Why this should also be here? This one should be reached as `o.analyzed.info.typeddict_type` from `visit_typeddict_expr()`.

Also if there is any reason for typed dict, why then not named tuple is not here?

mypy/mixedtraverser.py
```diff
@@ -0,0 +1,89 @@
...
+        self.visit_optional_type(var.type)
+
+    def visit_func(self, o: FuncItem) -> None:
+        super().visit_func(o)
+        self.visit_optional_type(o.type)
+
+    def visit_class_def(self, o: ClassDef) -> None:
+        super().visit_class_def(o)
+        info = o.info
+        if info:
```
ilevkivskyi: I think there is something missing here. For various auto-generated `TypeInfo`s there are methods/variables without corresponding syntactic nodes. These will be never reached unless you actually loop through the namespace. We used to have a bunch of crashes because of this.

mypy/mixedtraverser.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def visit_newtype_expr(self, o: NewTypeExpr) -> None:
+        super().visit_newtype_expr(o)
+        self.visit_optional_type(o.old_type)
+
+    # Statements
+
+    def visit_assignment_stmt(self, o: AssignmentStmt) -> None:
+        super().visit_assignment_stmt(o)
+        self.visit_optional_type(o.type)
```
ilevkivskyi: Again, this type is not guaranteed to be the _same_ type as `TypeAlias.target` it technically may be a copy.

mypy/mixedtraverser.py
```diff
@@ -0,0 +1,89 @@
...
+    def visit_type_application(self, o: TypeApplication) -> None:
+        super().visit_type_application(o)
+        for t in o.types:
+            t.accept(self)
+
+    # Helpers
+
+    def visit_optional_type(self, t: Optional[Type]) -> None:
+        if t:
+            t.accept(self)
```
ilevkivskyi: Is `EnumCallExpr` missing, or we don't need it before https://github.com/python/mypy/issues/6289 is fixed?

JukkaL: `EnumCallExpr` doesn't contain any type objects so we don't need to traverse here, I think.

mypy/mixedtraverser.py
JukkaL: If the file and line number is the same, it shouldn't be a problem. The main constraint is that we don't want to visit something that might be defined in another module (line number refer to the wrong file).

mypy/mixedtraverser.py
JukkaL: What sorts of crashes? Since there are generated in code, we can perhaps assume that they have valid type arguments. I'll add a comment here.

mypy/mixedtraverser.py
JukkaL: This looks redundant. I'll remove this.

mypy/mixedtraverser.py
JukkaL: `bases` actually don't seem to be visited at all. This should be the only place where we visit the base type. More generally, generated TypeInfos are handled in a bit ad hoc fashion, but it shouldn't be a problem for type argument checks, as it's somewhat reasonable to assume that generated things are valid. I'll create a follow-up issue about this.

mypy/newsemanal/semanal_typeargs.py
JukkaL: We do this earlier in the type analyzer, since the number of type variables is known earlier. I'll add a comment here.

mypy/mixedtraverser.py
ilevkivskyi: Crashes appeared in something related to this:
```python
class NT(NamedTuple):
    x: List[int, str]
```
Anyway, we have very good test coverage for these, so if there any problems, they will be likely caught when we will start enabling the tests.

mypy/mixedtraverser.py
ilevkivskyi: The problem here might that the bad instance from `A = List[int, str]` will appear in two places, and only one will be fixed if we don't traverse both.

ilevkivskyi: I just noticed your other comment, if these are fixed before, then this should be fine.

mypy/newsemanal/semanal_typeargs.py
ilevkivskyi: > We do this earlier in the type analyzer, since the number of type variables is known earlier.

Well, it looks like we don't do this for type aliases, see
https://github.com/python/mypy/blob/master/mypy/newsemanal/typeanal.py#L350. I will open an issue for this.

Issue #6315, opened 2019-02-04
New semantic analyzer: __name__ and other magic attributes


JukkaL:

ilevkivskyi:
Currently this is one of the blockers for mypy self-check, so raising priority to high.

PR #6401, opened 2019-02-14
New semantic analyzer: Support implicit module attributes such as __name__

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6315

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -284,6 +284,30 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             target = self.scope.current_target()
         self.cur_mod_node.plugin_deps.setdefault(trigger, set()).add(target)
 
+    def add_implicit_module_attrs(self, file_node: MypyFile) -> None:
+        """Manually add implicit definitions of module '__name__' etc."""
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -284,6 +284,30 @@ def add_plugin_dependency(self, trigger: str, target: Optional[str] = None) -> N
...
+                                     UnboundType('__builtins__.unicode')])
+            else:
+                assert t is not None, 'type should be specified for {}'.format(name)
+                typ = UnboundType(t)
+
+            an_type = self.anal_type(typ)
+            if an_type:
+                var = Var(name, an_type)
+                var._fullname = self.qualified_name(name)
+                var.is_ready = True
```
JukkaL: Is this necessary? Looks like we don't set this in the old semantic analyzer.

ilevkivskyi: We do this here https://github.com/python/mypy/blob/master/mypy/semanal.py#L301 the logic used to be split between passes. I am not sure if it is actually useful, but probably it is safer to keep it for 1:1 correspondence with old analyzer.

JukkaL: Makes sense!

PR #6403, opened 2019-02-14
Sync typeshed (and fix typeshed build breakage)

msullivan:
This requires another (wrong, of course) addition to the errors in
testNewAnalyzerBasicTypeshed_newsemanal, since there has been a typeshed
change that causes another newsemanal error.

(This is actually the main motivation for this sync, to allow fixing the
typeshed build breakage.)


msullivan:
I'm gonna just merge this but heads up @ilevkivskyi

JelleZijlstra:
Thanks for fixing this!

PR #6405, opened 2019-02-15
Don't call abspath() on path name when validating cache

msullivan:
abspath() is surprisingly slow. This gave almost a second of speedup
for me on cold run startup.

This does change the cache contents, but I don't think it should
really matter.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -562,6 +562,8 @@ class BuildManager(BuildManagerBase):
         if not self.shadow_map:
             return path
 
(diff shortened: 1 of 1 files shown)


gvanrossum:
Note that normpath doesn’t turn absolute paths into regular paths. That’s probably OK, but some use cases could be affected.

msullivan:
Our `normpath` function doesn't ever actually call normpath, though: it calls relpath in bazel mode and abspath otherwise

PR #6409, opened 2019-02-15
New semantic analyzer: support comprehensions

### LINKED ISSUE #6360: New semantic analyzer: support comprehensions
This is one of the blockers for mypy self-check.

JukkaL:
We now remember the namespaces for comprehensions, since they behave
similarly to nested functions.

Fixes #6360.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -244,7 +244,8 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         # analyzed in several iterations until all names are resolved. We need to save
         # the local namespaces for the top level function and all nested functions between
         # these iterations. See also semanal_main.process_top_level_function().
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -3833,16 +3834,13 @@ def qualified_name(self, n: str) -> str:
             base = self.cur_mod_id
         return base + '.' + n
 
-    def enter(self, function: Optional[FuncItem] = None) -> None:
+    def enter(self, function: Union[FuncItem, GeneratorExpr, DictionaryComprehension]) -> None:
         """Enter the function scope.
 
         The argument can be omitted for temporary scopes (like comprehensions
         and generator expressions) that can't have incomplete definitions.
```
ilevkivskyi: This docstring should be updated.

PR #6414, opened 2019-02-15
Fix mypy-mypyc by making final_iteration a property

msullivan:
There is a mypyc bug where we don't properly support
abstractproperties being implemented by regular attributes
(https://github.com/mypyc/mypyc/issues/531), so avoid that.


msullivan:
mypyc doesn't support setters yet :(
(because mypy didn't use them)

Issue #5664, opened 2018-09-24
type signature in generic subclass


oraluben:
```
from typing import Generic, Callable, TypeVar, Iterator

T = TypeVar('T')


class Box(Generic[T]):
    @classmethod
    def wrap(cls, generator: Callable[[], T]) -> 'Box[T]':
        pass


class IteratorBox(Box[Iterator[T]],
                  Generic[T]):
    pass


def g() -> int:
    return 0


r = IteratorBox.wrap(g)

b: IteratorBox = IteratorBox
reveal_type(b.wrap)
reveal_type(IteratorBox.wrap)
reveal_type(IteratorBox.wrap(g))
```

output of mypy is
```test.py:23: error: Incompatible types in assignment (expression has type "Type[IteratorBox[Any]]", variable h
as type "IteratorBox[Any]")
test.py:24: error: Revealed type is 'def (generator: def () -> typing.Iterator*[Any]) -> test.Box[typing.Iter
ator*[Any]]'
test.py:25: error: Revealed type is 'def [T] (generator: def () -> T`1) -> test.Box[T`1]'
test.py:26: error: Revealed type is 'test.Box[builtins.int*]'
```

I was expecting that mypy will find type mismatch to call `IteratorBox.wrap(g)`, which it didn't.
And `reveal_type(IteratorBox.wrap)` just return the type of the method in base class.

oraluben:
Thanks for your reply! I have tried the PR and found a code may get wrong checking result. I will put my repro in the [PR](https://github.com/python/mypy/pull/5420).

PR #6418, opened 2019-02-16
Fix member access on generic classes

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/3645
Fixes https://github.com/python/mypy/issues/1337
Fixes https://github.com/python/mypy/issues/5664

The fix is straightforward, I just add/propagate the bound type variable values by mapping to supertype.

I didn't find any corner cases with class methods, and essentially follow the same logic as when we generate the callable from `__init__` for generic classes in calls like `C()` or `C[int]()`.

For class attributes there are two things I fixed. First we used to prohibit ambiguous access:
```python
class C(Generic[T]):
    x: T
C.x  # Error!
C[int].x  # Error!
```
but the type variables were leaking after an error, now they are erased to `Any`. Second, I now make an exception and allow accessing attributes on `Type[C]`, this is very similar to how we allow instantiation of `Type[C]` even if it is abstract (because we expect concrete subclasses there), plus this allows accessing variables on `cls` (first argument in class methods), for example:
```python
class C(Generic[T]):
    x: T
    def get(cls) -> T:
        return cls.x  # OK
```

(I also added a bunch of more detailed comments in this part of code.)


mypy/checkmember.py
```diff
@@ -670,23 +707,28 @@ def A(Generic(T)):
...
     B.foo()
 
     original_type is the value of the type B in the expression B.foo()
     """
     # TODO: verify consistency between Q and T
     info = itype.type  # type: TypeInfo
+    if is_classmethod:
+        assert isuper is not None
+        t = expand_type_by_instance(t, isuper)
+    ids = {t.id for t in itype.args if isinstance(t, TypeVarType)}
```
msullivan: Could you add some comments about what is going on here?

ilevkivskyi: I added a detailed comment (and even found a little bug while writing it)/

ilevkivskyi: I double-checked that everything is OK with internal code bases.

Issue #6416, opened 2019-02-16
Inconsistency in metaclass/Type interaction


msullivan:
Since #6370, we reject the following code:
```
from typing import Type
class Meta(type):
    pass
class Thing(metaclass=Meta):
    pass

def foo(x: Type[Thing]) -> Type[Thing]:
    assert isinstance(x, Meta)
    return x
```
with the error `Incompatible return value type (got "Meta", expected "Type[Thing]")`.

Previously this worked because we thought that `Meta` and `Type[Thing]` had no overlap and so the rest of the function after the assert wasn't checked.

This caused one error in S.

The answer might be that we need to consider `Type[Thing]` to be a subtype of `Meta`?

ilevkivskyi:
> The answer might be that we need to consider `Type[Thing]` to be a subtype of `Meta`?

IIRC this is already true, we need to fix _proper_ subtype (which is still a separate visitor, see also https://github.com/python/mypy/issues/3297)

Also even with proper subtype fixed, this may not help since `isinstance()` can widen type, see
https://github.com/python/mypy/issues/6386. Or maybe better proper subtype check will fix both issues.

Issue #6386, opened 2019-02-13
Inheritance broken when testing isinstance?


rldleblanc:
I don't know if I'm running into a bug or I'm just not doing things right.

I was able to minimize my problem to this using Python 3.5 and mypy 0.660:
```
from abc import ABC, abstractmethod                                                
from typing import Any, Optional, Type, Union                                      
                                                                                   
class A(ABC):                                                                      
    pass                                                                           
                                                                                   
class AA(A):                                                                       
    pass                                                                           
                                                                                   
class AB(A):                                                                       
    pass                                                                           
                                                                                   
class M(object):                                                                   
    pass                                                                           
                                                                                   
class MA(M, AA):                                                                   
    pass                                                                           
                                                                                   
class MB(M, AB):                                                                   
    pass                                                                           
                                                                                   
MYTYPES = Union[MA, MB]                                                            
                                                                                   
class T(object):                                                                   
    def __init__(self, newtype:Type[MYTYPES]=MA) -> None:                          
        assert isinstance(newtype, type(M))                                        
        self._new_type = newtype    # type: Type[MYTYPES]                          
                                                                                   
    def test(self) -> A:                                                           
        ret = self._new_type()                                                     
        print(type(ret))                                                           
        return ret                                                                 
                                                                                   
x = T()                                                                            
print(isinstance(x.test(), A)) # True                                              
y = T(MB)                                                                          
print(isinstance(y.test(), A)) # True
```
My results are:
```
$ python3 -m mypy mixin_test.py
mixin_test.py:27: error: Incompatible types in assignment (expression has type "Type[M]", variable has type "Union[Type[MA], Type[MB]]")
$ python3 mixin_test.py                                                                
<class '__main__.MA'>
True
<class '__main__.MB'>
True
```
It seems that the assert is causing mypy to not realize that the object is still of type 'A'. I just want to make sure that the item passed in has the MixIn 'M' so that it can support the functions that I need. It is quite possible that I'm approaching this whole thing wrong.

ilevkivskyi:
Mypy has special treatment of `assert isinstance(obj, ...)`, they work as runtime-checked casts, and are normally used to narrow types. In your case it actually widens the type. I would say, we should not widen type after an assert (unless I am missing some use cases).

Anyway, IIUC you should just do this:
```python
from typing import TYPE_CHECKING

...

if not TYPE_CHECKING:
    assert isinstance(newtype, type(M))

...
```
[...]

Issue #4616, opened 2018-02-21
Binder doesn't work well with Type[...]


ilevkivskyi:
This code:
```python
from typing import Union, Type

class C: ...

x: Union[C, Type[C]]

if isinstance(x, type) and issubclass(x, C):
    reveal_type(x)
```
Shows that the type of `x` is `builtins.type`. Ideally it should be `Type[C]`.

andrewbaxter:
Is this the same as:

```
class A:
    x = 4


i: type = A
assert issubclass(i, A)
print(i.x)
```
which fails with:
```
7: error: "type" has no attribute "x"
```
?

mypy 0.590

It looks like `issubclass` doesn't imply anything AFAICT, unlike `isinstance`.

There's this workaround:
```
i: type = A
j: Type[A] = i
print(j.x)
```

but since this isn't in assertion format it seems more like a static check omission.

ilevkivskyi:
Yes, this is the same bug.

chadrik:
Just got bit by this one.

ilevkivskyi:
(Removing unions tag since there are repros without unions, and increasing priority to high since this appeared several times again.)

PR #6419, opened 2019-02-17
Fix interaction of isinstance() with Type[...]

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/4616
Fixes https://github.com/python/mypy/issues/6416
Fixes https://github.com/python/mypy/issues/6386

Currently `Type[C]` is considered a subtype of metaclass of `C`, and I think this is right. The idea is to fix `ProperSubtypeVisitor.visit_type_type()` to match `SubtypeVisitor.visit_type_type()` (plus couple small updates to make the former used by `isinstance()` and `issubclass()`).


mypy/subtypes.py
```diff
@@ -1210,6 +1210,12 @@ def visit_type_type(self, left: TypeType) -> bool:
                 return True
             if right.type.fullname() == 'builtins.object':
                 return True
+            item = left.item
+            if isinstance(item, TypeVarType):
+                item = item.upper_bound
+            if isinstance(item, Instance):
```
msullivan: There is a "TODO: handle metaclasses?" comment above that I think can be removed now

ilevkivskyi:
> I wonder if there is a long-term way to clean up some of the complicated interactions involving `Type`, though

I was thinking about this, but it looks like this part of the type system is inherently complex. (Especially if we want to keep it usable in practice.) Also there are still many holes in the type visitors, just few lines above is:
```python
    def visit_overloaded(self, left: Overloaded) -> bool:
        # TODO: What's the right thing to do here?
        return False
```
The meet visitor (and related helpers) probably require the most attention.

Issue #6413, opened 2019-02-15
New semantic analyzer: don't crash on malformed named tuples in incremental mode


ilevkivskyi:
Currently malformed named tuples like this one (note name mismatch):
```
sched_param = NamedTuple('sched_priority', [
    ('sched_priority', int),
])
```
Cause a crash in incremental mode with new semantic analyzer.

PR #6421, opened 2019-02-17
New semantic analyzer: Fix (de)serialization of broken named tuples

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6413

In my original PR I missed the fact that "broken" named tuples (those where variable name and named tuple name are different) are stored under both names (one to actually access it, and other for (de)serialization).

I also extend the semantic analyzer API, and clean-up/clarify the symbol adding logic a bit.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1794,8 +1794,9 @@ class State:
         semantic analyzer will perform this patch for us when processing stale
         SCCs.
         """
+        Analyzer = Union[SemanticAnalyzerPass2, NewSemanticAnalyzer]  # noqa
         if self.manager.options.new_semantic_analyzer:
-            analyzer = self.manager.new_semantic_analyzer
+            analyzer = self.manager.new_semantic_analyzer  # type: Analyzer
         else:
             analyzer = self.manager.semantic_analyzer
         for dep in self.dependencies:
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1794,8 +1794,12 @@ class State:
         semantic analyzer will perform this patch for us when processing stale
         SCCs.
         """
+        if self.manager.options.new_semantic_analyzer:
+            analyzer = self.manager.new_semantic_analyzer
+        else:
+            analyzer = self.manager.semantic_analyzer
         for dep in self.dependencies:
-            self.manager.semantic_analyzer.add_submodules_to_parent_modules(dep, True)
+            analyzer.add_submodules_to_parent_modules(dep, True)
 
     def fix_suppressed_dependencies(self, graph: Graph) -> None:
         """Corrects whether dependencies are considered stale in silent mode.
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3959,7 +3959,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         """Same as above, but skipping the local namespace.
 
         This doesn't check for previous definition and is only used
-        for serialization of method-level named tuples.
+        for serialization of method-level classes.
         """
         if self.type is not None:
             names = self.type.names
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3955,6 +3955,21 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                                  module_hidden=module_hidden)
         return self.add_symbol_table_node(name, symbol, context)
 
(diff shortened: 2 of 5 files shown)


mypy/build.py
```diff
@@ -1794,8 +1794,13 @@ def patch_dependency_parents(self) -> None:
         semantic analyzer will perform this patch for us when processing stale
         SCCs.
         """
+        Analyzer = Union[SemanticAnalyzerPass2, NewSemanticAnalyzer]  # noqa
```
JukkaL: What's this change? Is is related to the rest of this PR?

mypy/newsemanal/semanal_namedtuple.py
```diff
@@ -189,15 +189,17 @@ def check_namedtuple(self,
         else:
             default_items = {}
         info = self.build_namedtuple_typeinfo(name, items, types, default_items)
-        # Store it as a global just in case it would remain anonymous.
-        # (Or in the nearest class if there is one.)
-        self.store_namedtuple_info(info, var_name or name, call, is_typed)
+        if var_name:
+            self.store_namedtuple_info(info, var_name, call, is_typed)
+        if name != var_name:
```
JukkaL: This check is too subtle. Maybe also explicitly check whether we are in a function scope?

mypy/newsemanal/semanal_namedtuple.py
```diff
@@ -189,15 +189,17 @@ def check_namedtuple(self,
...
+            # Store function level (or broken) named tuple also in a name
```
JukkaL: Can this also happen if we have an inline named tuple definition in a base class?

Also, can you explain in more detail how the symbol tables look like in this case (there are two symbol table entries, potentially with different names)?

mypy/newsemanal/semanal_shared.py
```diff
@@ -122,7 +123,18 @@ def schedule_patch(self, priority: int, fn: Callable[[], None]) -> None:
...
+
+    @abstractmethod
+    def add_symbol(self, name: str, node: SymbolNode, context: Optional[Context],
+                   module_public: bool = True, module_hidden: bool = False) -> bool:
+        """Add symbol to the current symbol table."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def add_symbol_skip_local(self, name: str, node: SymbolNode) -> None:
+        """Add symbol to the current symbol table, skipping locals."""
```
JukkaL: Explain the purpose of this method, since it's not obvious why we'd need something like this.

mypy/newsemanal/semanal.py
```diff
@@ -3955,6 +3955,21 @@ def add_symbol(self, name: str, node: SymbolNode, context: Optional[Context],
                                  module_hidden=module_hidden)
         return self.add_symbol_table_node(name, symbol, context)
 
+    def add_symbol_skip_local(self, name: str, node: SymbolNode) -> None:
+        """Same as above, but skipping the local namespace.
+
+        This doesn't check for previous definition and is only used
+        for serialization of method-level classes.
```
JukkaL: Should this method be used for other types that can leak, such as regular classes?

Since this whole business is pretty delicate, it would be good to explain this in more detail. Something like this perhaps:

> Classes defined within methods can be exposed through an attribute type, but method-level symbol tables aren't serialized. This method can be used to add such classes to an enclosing, serialized symbol table.

mypy/build.py
ilevkivskyi: This is needed because otherwise incremental tests crash with the new analyzer (I added couple incremental tests) with something like `'BuildManager' doesn't have attribute 'semantic_analyzer'`.

mypy/newsemanal/semanal_namedtuple.py
ilevkivskyi: OK. (although it will be technically redundant, it is probably more robust)

mypy/newsemanal/semanal_namedtuple.py
ilevkivskyi: > Can this also happen if we have an inline named tuple definition in a base class?

Oh yes, forgot to mention this (we have a test case however). I will a detailed comment with some examples.

mypy/newsemanal/semanal.py
ilevkivskyi: Ideally yes, but I decided to separate out the bigger refactoring, see https://github.com/python/mypy/issues/6422

ilevkivskyi: I will also add a TODO with a link to the issue.

PR #6423, opened 2019-02-18
New semantic analyzer: abstract classes

### LINKED ISSUE #6322: New semantic analyzer: abstractness calculations
It's not clear if abstractness related properties can be calculated during the main pass. MROs will be complete, but namespaces of some base classes currently may be incomplete. One option would be to defer subclasses until all base classes have complete namespaces. This takes advantage of the fact that function decorators are analyzed as part of class top levels.

JukkaL:
This fixes abstract classes by moving determination of abstract
attributes to a new pass that happens between the main semantic
analysis pass and type checking. The new pass is added mostly
to simplify the main semantic analysis pass.

This also adds simple type inference for decorated functions from
the old semantic analysis pass 3. This fixes a test case where
we need access to the signature of an abstract method when type
checking the module top level. Since module top levels can't be
deferred during type checking, the simple type inference is needed 
to remain compatible with the old semantic analyzer.

Fixes #6322.

### DIFF
--- a/mypy/newsemanal/semanal_infer.py
+++ b/mypy/newsemanal/semanal_infer.py
@@ -12,9 +12,9 @@ def infer_decorator_signature_if_simple(dec: Decorator,
                                         analyzer: SemanticAnalyzerInterface) -> None:
     """Try to infer the type of the decorated function.
 
-    This lets us resolve references to decorated functions during
-    type checking when there are cyclic imports, as otherwise the
-    type might not be available when we need it.
+    This lets us resolve additional references to decorated functions
+    during type checking. Otherwise the type might not be available
+    when we need it, since module top levels can't be deferred.
 
     This basically uses a simple special-purpose type inference
     engine just for decorators.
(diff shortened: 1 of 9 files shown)


mypy/newsemanal/semanal_main.py
```diff
@@ -56,7 +59,8 @@ def semantic_analysis_for_scc(graph: 'Graph', scc: List[str]) -> None:
     # before functions. This limitation is unlikely to go away soon.
     process_top_levels(graph, scc)
     process_functions(graph, scc)
-    check_type_arguments(graph, scc)
+    check_type_arguments(graph, scc, errors)
+    process_abstract(graph, scc, errors)
```
ilevkivskyi: Bikeshedding: I would give it a longer name, like `process_abstract_status`.

mypy/newsemanal/semanal_infer.py
```diff
@@ -0,0 +1,108 @@
...
+                return typ.ret_type
+            return None
+        elif isinstance(expr.node, Var):
+            return expr.node.type
+    elif isinstance(expr, CallExpr):
+        return calculate_return_type(expr.callee)
+    return None
+
+
+def find_fixed_callable_return(expr: Expression) -> Optional[CallableType]:
```
ilevkivskyi: Maybe add a docstring while you are at it?

PR #6424, opened 2019-02-18
New semantic analyzer: Fix import prios

ilevkivskyi:
Old semantic analyzer first pass is setting whether an import is a top level one or not. New semantic analyzer reachability pass didn't do this. As a result, this screwed up import priorities, and in turn, ordering of modules within SCCs. This caused many `Cannot determine type of 'foo'` errors since we don't defer module top levels.

### DIFF
--- a/mypy/newsemanal/semanal_pass1.py
+++ b/mypy/newsemanal/semanal_pass1.py
@@ -1,8 +1,5 @@
 """Block/import reachability analysis."""
 
-from contextlib import contextmanager
-from typing import Iterator
-
 from mypy.nodes import (
     MypyFile, AssertStmt, IfStmt, Block, AssignmentStmt, ExpressionStmt, ReturnStmt, ForStmt,
     Import, ImportAll, ImportFrom, ClassDef, FuncDef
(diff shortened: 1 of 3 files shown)


mypy/newsemanal/semanal_pass1.py
```diff
@@ -40,16 +44,35 @@ def visit_file(self, file: MypyFile, fnam: str, mod_id: str, options: Options) -
...
             if isinstance(defn, AssertStmt) and assert_will_always_fail(defn, options):
                 # We've encountered an assert that's always false,
                 # e.g. assert sys.platform == 'lol'.  Truncate the
                 # list of statements.  This mutates file.defs too.
                 del file.defs[i + 1:]
                 break
 
+    def visit_import_from(self, node: ImportFrom) -> None:
+        if self.is_global_scope:
+            node.is_top_level = True
```
JukkaL: What about just saying `node.is_top_level = self.is_global_scope` (here and below)?

mypy/newsemanal/semanal_pass1.py
```diff
@@ -1,7 +1,11 @@
 """Block/import reachability analysis."""
 
+from contextlib import contextmanager
+from typing import Iterator
```
JukkaL: These don't seem to be used.

Issue #6420, opened 2019-02-17
New semantic analyzer: Check type argument count for type aliases


ilevkivskyi:
This is a follow-up for https://github.com/python/mypy/pull/6395

Unfixed instances in type aliases cause new semantic analyzer to crash, for example:
```python
class C(Generic[T]):
    ...
def func(x: List[C[T]]) -> T:
    ...
A = List[C]  # Missing type argument for C
x: A

reveal_type(func(x))  # Crash here with list index out of range
```
See also TODO here https://github.com/python/mypy/blob/master/mypy/newsemanal/typeanal.py#L350

ilevkivskyi:
It looks like it is possible to get the crash from unfixed instance even with builtin aliases:
```python
x: Optional[List] = None  # incomplete instance here
y: List[str]

x = ['a', 'b']
x.extend(y)
```

PR #6427, opened 2019-02-18
Use generic tuple in builtin fixtures

ilevkivskyi:
This is needed for #6420

### DIFF
--- a/test-data/unit/fine-grained-modules.test
+++ b/test-data/unit/fine-grained-modules.test
@@ -1271,12 +1271,12 @@ a.py:2: error: Too many arguments for "foo"
 
 [case testAddModuleAfterCache3-only_when_cache]
 # cmd: mypy main a.py
-# cmd2: mypy main a.py b.py c.py d.py e.py f.py g.py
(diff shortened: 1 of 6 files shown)


test-data/unit/fine-grained-modules.test
```diff
@@ -1271,12 +1271,12 @@ a.py:2: error: Too many arguments for "foo"
 
 [case testAddModuleAfterCache3-only_when_cache]
 # cmd: mypy main a.py
-# cmd2: mypy main a.py b.py c.py d.py e.py f.py g.py
-# cmd3: mypy main a.py b.py c.py d.py e.py f.py g.py
+# cmd2: mypy main a.py b.py c.py d.py e.py f.py g.py h.py
+# cmd3: mypy main a.py b.py c.py d.py e.py f.py g.py h.py
```
JukkaL: What's this change?

ilevkivskyi: It tests that we are abandoning cache if more than half of files are missing.

PR #6429, opened 2019-02-18
New semantic analyzer: conditional function definitions

### LINKED ISSUE #6318: New semantic analyzer: conditionally defined functions
Support conditionally defined functions was dropped at some point when implementing the early versions of the new semantic analyzer. It's probably worth implementing this from scratch since the old implementation was pretty messy.

This issue applies to functions where multiple definitions are considered reachable by mypy. Cases involving platform and Python version checks where only one implementation is reachable already work. Example where this applies:

```py
if cond():  # Condition determined at runtime!
    def foo() -> None:
        ... # something; will be checked
else:
    def foo() -> None:
        ... # something else; will also be checked
```

JukkaL:
To semantically analyze conditional function definitions (and function
redefinitions) we add redefinitions to a symbol table with a mangled
name of form `name-redefinition[N]`. This is necessary since 
functions are only analyzed if they can be accessed through a symbol
table.

I checked that this fixes most relevant test cases in 
`check-functions.test`. Using a decorated function in a redefinition
is still not working perfectly. That is also partially implemented in 
the old semantic analyzer. Conditionally defining an overloaded
function is also not supported. I'll create follow-up issues for these
remaining known limitations.

Fixes #6318.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3963,10 +3963,8 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             if existing.node != symbol.node:
                 if isinstance(symbol.node, (FuncDef, Decorator)):
                     self.add_func_redefinition(names, name, symbol)
-                if (isinstance(symbol.node, (FuncDef, Decorator))
+                if not (isinstance(symbol.node, (FuncDef, Decorator))
                         and self.set_original_def(existing.node, symbol.node)):
-                    pass
-                else:
                     self.name_already_defined(name, context, existing)
         elif name not in self.missing_names and '*' not in self.missing_names:
             names[name] = symbol
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3962,9 +3962,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 and not isinstance(existing.node, PlaceholderNode)):
             if existing.node != symbol.node:
                 if isinstance(symbol.node, (FuncDef, Decorator)):
-                    # Redefinition of a function. Must keep track of it in the symbol
-                    # table with a dummy name.
-                    names[name + '!'] = symbol
+                    self.add_func_redefinition(names, name, symbol)
                 if (isinstance(symbol.node, (FuncDef, Decorator))
                         and self.set_original_def(existing.node, symbol.node)):
                     pass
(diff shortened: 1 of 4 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -491,8 +491,14 @@ def file_context(self, file_node: MypyFile, fnam: str, options: Options,
...
 
     def visit_func_def(self, defn: FuncDef) -> None:
-        if not defn.is_decorated and not defn.is_overload:
-            self.add_func_to_symbol_table(defn)
+        defn.is_conditional = self.block_depth[-1] > 0
+
+        # Add module-level function to symbol table on first iteration
+        # (recurse_into_functions == False). Also add nested functions
+        # always, since they won't be processed as separate targets.
+        if not self.recurse_into_functions or len(self.function_stack) > 0:
```
ilevkivskyi: This explanation is not very clear. I would rather say that we don't add function top-level functions when we just enter each of it during second phase of analysis -- processing of functions.

Also I would rather explain _why_ this is needed (I agree that this looks reasonable).

test-data/unit/check-newsemanal.test
```diff
@@ -1417,3 +1417,62 @@ reveal_type(x)  # E: Revealed type is 'Tuple[builtins.int, builtins.int]'
...
+from typing import Callable
+
+def dec(f: Callable[[int], None]) -> Callable[[str], None]:
+    pass
+
+if int():
+    from m import f
+else:
+    @dec
+    def f(x: int) -> None:
```
ilevkivskyi: Do we have a test case for when this causes a type error? (Decorated function signature differs from original one.)

test-data/unit/check-newsemanal.test
```diff
@@ -1417,3 +1417,62 @@ reveal_type(x)  # E: Revealed type is 'Tuple[builtins.int, builtins.int]'
...
+        1()  # E: "int" not callable
+reveal_type(f) # E: Revealed type is 'def (x: builtins.str)'
+[file m.py]
+def f(x: str) -> None: pass
```
ilevkivskyi: I would add a test that a function with same name as a class is still semantically analyzed. In general, do we have enough tests for something like:
```python
x = 1
class C:
    pass
class C:
    x + 'no way'
```
etc?

test-data/unit/check-newsemanal.test
```diff
@@ -1417,3 +1417,62 @@ reveal_type(x)  # E: Revealed type is 'Tuple[builtins.int, builtins.int]'
...
+
+[case testNewAnalyzerConditionallyDefineFuncOverVar]
+from typing import Callable
+
+if int():
+    f: Callable[[str], None]
+else:
+    def f(x: str) -> None:
+        ...
+reveal_type(f) # E: Revealed type is 'def (builtins.str)'
```
ilevkivskyi: Does old semantic analyzer support the opposite order? I mean:
```python
if int():
    def f(x: str) -> None:
        ...
else:
      f: Callable[[str], None]
```
If yes, then the new one should probably support this too. Otherwise, this is probably not important.

test-data/unit/check-newsemanal.test
JukkaL: This doesn't work correctly with the old semantic analyzer either. I'll create an issue about this.

test-data/unit/check-newsemanal.test
JukkaL: This generates an error with the old semantic analyzer as well.

PR #6431, opened 2019-02-18
New semantic analyzer: Fix instances that are type alias targets

### LINKED ISSUE #6420: New semantic analyzer: Check type argument count for type aliases

ilevkivskyi:
Fix #6420 

The fix is straightforward. I don't think this actually requires a separate pass, as it was suggested in the TODO I removed.

I also remove some unused imports and enable `check-generics.test` (without one tricky test).

### DIFF
--- a/mypy/newsemanal/typeanal.py
+++ b/mypy/newsemanal/typeanal.py
@@ -16,7 +16,7 @@ from mypy.types import (
     CallableType, NoneTyp, DeletedType, TypeList, TypeVarDef, TypeVisitor, SyntheticTypeVisitor,
     StarType, PartialType, EllipsisType, UninhabitedType, TypeType, get_typ_args, set_typ_args,
     CallableArgument, get_type_vars, TypeQuery, union_items, TypeOfAny, ForwardRef, Overloaded,
-    LiteralType, RawExpressionType, PlaceholderType
+    LiteralType, RawExpressionType, PlaceholderType, TypeTranslator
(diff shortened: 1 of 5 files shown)


mypy/newsemanal/typeanal.py
```diff
@@ -1281,3 +1286,23 @@ def make_optional_type(t: Type) -> Type:
...
+def fix_instance_types(t: Type, fail: Callable[[str, Context], None]) -> Type:
+    """Recursively fix all instance types (type argument count) in a given type.
+
+    For example 'Union[Dict, List[str, int]]' will be transformed into
+    'Union[Dict[Any, Any], List[Any]]'.
+    """
+    return t.accept(InstanceFixer(fail))
+
+
+class InstanceFixer(TypeTranslator):
```
JukkaL: I think that you can use `TypeTraverserVisitor` here instead, since you modify types in place. It will be more efficient as it doesn't construct new types.

PR #6435, opened 2019-02-19
Update attrs plugin to work with new semantic analyzer

ilevkivskyi:


mypy/plugins/attrs.py
```diff
@@ -43,6 +44,9 @@
     'attr.attr',
 }  # type: Final
 
+# TODO: should we use underscored names like for named tuples?
+SELF_TVAR_NAME = 'AT'  # type: Final
```
JukkaL: I'd suggest using underscore (or double underscore?) here.

mypy/checker.py
```diff
@@ -1614,6 +1614,9 @@ def determine_type_of_class_member(self, sym: SymbolTableNode) -> Optional[Type]
         if isinstance(sym.node, TypeInfo):
             # nested class
             return type_object_type(sym.node, self.named_type)
+        if isinstance(sym.node, TypeVarExpr):
+            # No need to check type vars since they can't be used at runtime.
```
JukkaL: This comment is a bit unclear. Do you mean that their use is rejected in an expression/runtime context by mypy?

test-data/unit/check-attr.test
```diff
@@ -577,7 +577,10 @@ class C:
 
 [builtins fixtures/list.pyi]
 
+-- This is tricky with new analyzer:
+-- We need to know the analyzed type of a function while still processing top-level.
```
JukkaL: Probably worth creating an issue to track this.

mypy/checker.py
ilevkivskyi: Yes.

test-data/unit/check-attr.test
ilevkivskyi: Opened #6438

Issue #6287, opened 2019-02-01
New semantic analyzer: TypeVar bound/values


JukkaL:
Forward references should work in bounds and value restriction.

PR #6437, opened 2019-02-19
Support forward references in upper bounds

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6287

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2636,6 +2636,10 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                         # for anything before main pass of semantic analysis is finished. We will
                         # incrementally populate `TypeVarExpr` if some part is missing during main
                         # pass iterations.
+                        # NOTE: It is safe to not call self.defer() here, because the only way
+                        # we can get None from self.anal_type() is if self.found_incomplete_refs()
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -2634,7 +2634,9 @@ def process_typevar_parameters(self, args: List[Expression],
                     analyzed = self.expr_to_analyzed_type(param_value,
                                                           report_invalid_types=False)
                     if analyzed is None:
-                        return None
+                        # It is fine to simply use a temporary Any because we don't need the bound
+                        # for anything before main pass of semantic analysis is finished.
+                        analyzed = AnyType(TypeOfAny.special_form)
```
JukkaL: Update comment to mention that we will incrementally populate `TypeVarExpr` if some part is missing.

However, do we always defer somewhere if we get here? In particular, there should probably be a test case where this is the only cause of deferral, if possible. 

Previously the caller deferred if there is a `None` return from here -- which actually doesn't look quite right in case of some the above `None` returns. What happens if one of the above checks fails (e.g., `covariant` with invalid value)?

ilevkivskyi: Thanks! I removed the unnecessary `mark_incomplete()` call from the caller, and added a detailed comment explaining why it is safe to not defer here.

PR #6439, opened 2019-02-19
New semantic analyzer: support assignment based TypedDict

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6286

There are no tricks, it mostly matches one-to-one the implementation of named tuples. There is however one temporary hack, that we will be able to remove when we merge analyzers. Currently type checker depends on l.h.s pointing to a (redundant) `Var` that is always created by the old analyzer.

### DIFF
--- a/test-data/unit/check-typeddict.test
+++ b/test-data/unit/check-typeddict.test
@@ -1323,6 +1323,7 @@ reveal_type(x['a']['b']) # E: Revealed type is 'builtins.int'
 [builtins fixtures/dict.pyi]
 
 [case testSelfRecursiveTypedDictInheriting]
+# flags: --no-new-semantic-analyzer
 from mypy_extensions import TypedDict
 
 class MovieBase(TypedDict):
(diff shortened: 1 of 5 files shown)


test-data/unit/check-typeddict.test
```diff
@@ -1323,6 +1323,7 @@ reveal_type(x['a']['b']) # E: Revealed type is 'builtins.int'
 [builtins fixtures/dict.pyi]
 
 [case testSelfRecursiveTypedDictInheriting]
+# flags: --no-new-semantic-analyzer
```
JukkaL: Create a follow-up issue about remaining things that aren't working?

test-data/unit/check-typeddict.test
```diff
@@ -1338,6 +1339,7 @@ reveal_type(m['director']['name']) # E: Revealed type is 'builtins.str'
 [out]
 
 [case testSubclassOfRecursiveTypedDict]
+# flags: --no-new-semantic-analyzer
```
JukkaL: Similar to above.

ilevkivskyi: Opened https://github.com/python/mypy/issues/6445 for both (since this is essentially the same issue).

PR #6441, opened 2019-02-20
Fix missing builtin aliases on Python 2 (and other versions)

ilevkivskyi:
Some builtin aliases (like `typing.ChainMap`) are not available on some Python versions, or in some fixtures. However, if while they were first looked up while target namespace was incomplete, we put a `PlaceholderNode` in `typing`. This PR either doesn't put or removes this placeholder node if we didn't find the target after the target namespace is complete.

I didn't find a way to test this (since we don't have incremental `pythoneval` tests), but I double-checked this fixes the crash with actual Python 2 stubs.


mypy/newsemanal/semanal.py
```diff
@@ -3823,7 +3823,11 @@ def add_builtin_aliases(self, tree: MypyFile) -> None:
                 self.mark_incomplete(name, tree)
             else:
                 # Test fixtures may be missing some builtin classes, which is okay.
-                pass
+                # Also some targets are not available on Python 2, like collections.ChainMap.
+                # Kill the placeholder if there is one.
+                if name in tree.names:
```
JukkaL: This is kind of hacky. Why do we generate the placeholder for things that don't exist? For example, could we include version information in the aliases, and skip the alias definition if the target Python version is too old?

ilevkivskyi: As discussed elsewhere, I keep this for test fixtures, but keep track of Python versions for real typeshed stubs.

Issue #5393, opened 2018-07-26
mypy loses type of optional Tuple variable after setting default in runtime


blazewicz:
Please provide more information to help us understand the issue:

* Are you reporting a bug, or opening a feature request?

reporting a bug

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.

```Python
from typing import Tuple, Optional


def foo(value: Optional[int]) -> None:
    """"Optional argument of type int with default value set in runtime."""
    if value is None:
        value = 42
    reveal_type(value)  # builtins.int
    value.bit_length()  # no error


def bar(value: Tuple[int, list]) -> None:
    """Argument of a tuple type."""
    reveal_type(value)     # Tuple[builtins.int, builtins.list[Any]]
    reveal_type(value[0])  # builtins.int
    reveal_type(value[1])  # builtins.list[Any]
    value[1].sort()  # no errors


def baz(value: Optional[Tuple[int, list]]) -> None:
    """mypy loses actual type of argument after setting default value in runtime."""
    if value is None:
        value = (0, [])  # no errors here, types match
    # unexpected Union type below, notice both elements are identical
    reveal_type(value)     # Union[Tuple[builtins.int, builtins.list[Any]], Tuple[builtins.int, builtins.list[Any]]]
    # now we have only garbage
    reveal_type(value[0])  # Union[builtins.object*, Any]
    reveal_type(value[1])  # Union[builtins.object*, Any]
    value[1].sort()  # error: Item "object" of "Union[object, Any]" has no attribute "sort"
```

* What is the actual behavior/output?

mypy loses actual type of object

* What is the behavior/output you expect?

no errors in code above

* What are the versions of mypy and Python you are using?
  Do you see the same issue after installing mypy from Git master?

fresh mypy from git master

* What are the mypy flags you are using? (For example --strict-optional)

none

gvanrossum:
As a workaround, can you use List instead of list? Does that make things
better? If not, what if you give it an explicit item type, e.g. List[int]?

blazewicz:
changing list -> List[int] doesn't fix anything, I've found this problem with a custom type and a custom class as tuple elements, code above is simplified case

```
mypyerror.py:27: error: Revealed type is 'Union[Tuple[builtins.int, builtins.list[builtins.int]], Tuple[builtins.int, builtins.list[builtins.int]]]'
[...]

Issue #3575, opened 2017-06-20
Use join to calculate better fallback for tuples


JukkaL:
Currently the fallback of named tuple types and tuples is `Tuple[Any, ...]` but we could use join of items to get a more precise fallback type after #3560 has been merged.

Pointers:
* Update `tuple_type` in `mypy.typeanal`.
* Update `build_namedtuple_typeinfo` in `mypy.semanal`.
* Use the fixups to be introduced in #3560 (the PR has not been merged as of writing this).

gvanrossum:
Remind me where the fallback type is used? I already think we're being too strict in some cases, e.g.
```py
a = (1, 2)
b = a + ('', '')
```
give an error on the second line `Unsupported operand types for + ("Tuple[int, int]" and "Tuple[str, str]")` but I think it should just make the type of `b: Tuple[int, int, str, str]`.

Also, this is still an error:
```py
from typing import Tuple
a = (1, 2)
b = a + ('', '')  # type: Tuple[int, int, str, str]
```
There are actually two errors even:
```
__tmp__.py:3: error: Incompatible types in assignment (expression has type Tuple[int, ...], variable has type "Tuple[int, int, str, str]")
__tmp__.py:3: error: Unsupported operand types for + ("Tuple[int, int]" and "Tuple[str, str]")
```

ilevkivskyi:
> Remind me where the fallback type is used?

The most important places are probably in ``subtypes.py`` where subtyping relationships between ``TupleType``s and ``Instance``s are decided (e.g. for named tuples), and in ``checkmember.py`` for look-up of attributes on tuples (like ``__add__`` in your example).

We of course could be more flexible with tuples, and there are issues for this like #224, this could be done by special-casing. A more "broad" solution for such situations may be variadic generics. Anyway, both these ways are mostly unrelated to the joins in fallbacks.

JukkaL:
The fallback `Any` item type seems to used at least for multiplication and addition. It doesn't affect subtyping universally, since tuple subtyping has been special cased, but it might affect some special cases. Other operations might not be affected since they are special cased.

ilevkivskyi:
Hm, actually this issue was fixed by #3952 only _partially_: the first bullet in the first message is still missing. Precise fallbacks are generated only for named tuples, not for "normal" tuples, the latter still have ``Any`` in fallback.

ilevkivskyi:
[...]

PR #6442, opened 2019-02-20
Fixes to tuple fallbacks

### LINKED ISSUE #6400: New semantic analyzer: calculating fallback types
Some fallback types can't be calculated during the main semantic analysis pass, since they require joins. We can either calculate the fallbacks in a later pass before type checking, or we can calculate them on demand during type checking. I'm leaning towards the latter option.

JukkaL:
This changes tuple fallbacks to be calculated on demand during
type checking. This fixes some issues with fallbacks being imprecise.

In the new semantic analyzer, this introduces a new pass just after
the main semantic analysis pass to calculate precise item types for
`tuple` base classes. These can't be calculated on demand since 
the base class is a variable-length tuple type (`Instance`) instead
of `TupleType`. We can't calculate these during the main semantic
analysis pass since base classes can be incomplete.

Fixes #6400.


mypy/newsemanal/semanal_shared.py
```diff
@@ -19,12 +20,8 @@
 # Priorities for ordering of patches within the final "patch" phase of semantic analysis
 # (after pass 3):
```
ilevkivskyi: There is no pass three anymore, so this comment can also be updated.

mypy/newsemanal/semanal_shared.py
```diff
@@ -190,3 +187,14 @@ def set_callable_name(sig: Type, fdef: FuncDef) -> Type:
...
     else:
         return sig
+
+
+def calculate_tuple_fallback(typ: TupleType) -> None:
+    """Calculate a precise item type for the fallback of a tuple type.
+
+    This must be called only after the main semantic analysis pass, since joins
+    aren't available before that.
+    """
```
ilevkivskyi: I would mention the chicken and egg problem about tuple fallbacks and checking upper bounds for instances here, and/or maybe in a comment in `semanal_main()` where the two are called.

test-data/unit/check-newsemanal.test
```diff
@@ -1580,6 +1580,40 @@ class C:
...
 
+[case testNewAnalyzerTupleIteration]
+from typing import Union, Tuple, NamedTuple
+
+class A: pass
+class B(A): pass
+class C(A): pass
+
+class T(Tuple[B, C]):
+    pass
```
ilevkivskyi: Maybe put the classes in reverse order (just to check that forward refs also work, although I am quite they should).

test-data/unit/check-newsemanal.test
```diff
@@ -1580,6 +1580,40 @@ class C:
...
+
+class NTInt(NamedTuple):
+    x: int
+    y: int
+
+class NTStr(NamedTuple):
+    x: str
+    y: str
```
ilevkivskyi: I would add couple non-trivial upper bound checks, just be sure they behave reasonably with the current ordering (fallbacks first, then bounds).

test-data/unit/check-unions.test
```diff
@@ -782,14 +782,16 @@ class NTStr(NamedTuple):
...
+reveal_type(t1.__iter__) # E: Revealed type is 'def () -> typing.Iterator[builtins.int*]'
 nt: Union[NTInt, NTStr]
+reveal_type(nt.__iter__) # E: Revealed type is 'Union[def () -> typing.Iterator[builtins.int*], def () -> typing.Iterator[builtins.str*]]'
 for nx in nt:
     reveal_type(nx) # E: Revealed type is 'Union[builtins.int*, builtins.str*]'
 
 t: Union[Tuple[int, int], Tuple[str, str]]
 for x in t:
-    # TODO(Ivan): This will be OK when tuple fallback patches are added (like above)
-    reveal_type(x) # E: Revealed type is 'Any'
```
ilevkivskyi: There is also an issue about this on tracker https://github.com/python/mypy/issues/3575. Also there is a chance that https://github.com/python/mypy/issues/5393 will be fixed by this PR. Should we close issues when they are fixed in the new analyzer?

mypy/newsemanal/semanal_main.py
```diff
@@ -204,6 +217,11 @@ def semantic_analyze_target(target: str,
         return [], analyzer.incomplete
 
 
+def process_patches(graph: 'Graph', scc: List[str]) -> None:
+    for module in scc:
+        graph[module].semantic_analysis_apply_patches()
```
ilevkivskyi: I would add a motivational docstring explaining why we do mini-passes for one things, and patches for other. My understanding that there will be quite few tuple types in a typical program, but quite a lot instances, so we a make a traverser for the latter.

Also there is an option to combine these two in a single type fixing traverser pass, and then repeat it until we solved the chicken and egg problem (this would require to somehow mark types as ready). I however didn't think much about this, maybe this will not help.

Issue #6434, opened 2019-02-19
New semantic analyzer doesn't work with imports from (and of) missing modules


ilevkivskyi:
For example:
```
[case testNewAnalyzerMissingImport]
# flags: --ignore-missing-imports
from non_existing import stuff
x: C
class C: ...
```
Fails with `Name 'stuff' already defined (possibly by an import)`.

PR #6444, opened 2019-02-20
New semantic analyzer: Fix missing imports

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6434

This makes missing import behave the same way as for old analyzer in situations with multiple iterations.


mypy/newsemanal/semanal.py
```diff
@@ -4041,6 +4041,11 @@ def add_imported_symbol(self, name: str, node: SymbolTableNode, context: Context
     def add_unknown_symbol(self, name: str, context: Context, is_import: bool = False,
                            target_name: Optional[str] = None) -> None:
         """Add symbol that we don't know what it points to (due to error, for example)."""
+        if is_import:
```
JukkaL: Looks like `is_import` is always true. Maybe remove the argument (can happen in a separate PR)?

ilevkivskyi: OK, I will make a separate PR.

PR #6448, opened 2019-02-20
New semantic analyzer: fix class derived from Tuple with __init__

### LINKED ISSUE #6446: New semantic analyzer: overriding __init__ with Tuple base class
This generates a false positive:

```py
from typing import Tuple

c: C
class C(Tuple[int, str]):
    def __init__(self) -> None: pass  # Cannot overwrite NamedTuple attribute "__init__"
```

This error gets generated when processing typeshed builtins import cycle, so this is high priority.

JukkaL:
NamedTuple detection was incorrect.

Fixes #6446.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -947,7 +947,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
 
     def analyze_namedtuple_classdef(self, defn: ClassDef) -> bool:
         """Check if this class can define a named tuple."""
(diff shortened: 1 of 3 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -1277,11 +1277,10 @@ def configure_base_classes(self,
         base_types = []  # type: List[Instance]
         info = defn.info
 
+        info.tuple_type = None
```
JukkaL: This change was needed to get existing tests passing.

PR #6452, opened 2019-02-20
New semantic analyzer: Enable namedtuple tests and fix self-check error

ilevkivskyi:
This fixes the `Name 'cls' already defined error` in self-check with new analyzer. The error is caused by the fact that now we actually analyze generated methods. I solve them simply by using underscored versions of `_self` and `_cls` (this is btw the same what runtime implementation does for some generated methods).

I also enable both named tuple test files. This needs the same temporary redundant variable hack as I did for typed dicts, it can be removed when we will not need to be compatible with old analyzer, I will create an issue so we will not forget about this. 

I skip a bunch of test cases for recursive named tuples (there is already an issue for this). I also skip one test because it is tricky to fix, I will open a follow-up issue.


ilevkivskyi:
I will review the tuple PR later today, so that we can enable one more test here, and then will rebase this one (or make a separate PR).

mypy/newsemanal/semanal.py
```diff
@@ -1915,6 +1915,14 @@ def analyze_namedtuple_assign(self, s: AssignmentStmt) -> bool:
         # Yes, it's a valid namedtuple, but defer if it is not ready.
         if not info:
             self.mark_incomplete(name, lvalue, becomes_typeinfo=True)
+        else:
+            # TODO: This is needed for one-to-one compatibility with old analyzer, otherwise
+            # type checker will try to infer Any for the l.h.s.
+            # Remove this after new analyzer is the default one!
```
JukkaL: This comment is a bit unclear. What would happen if we'd infer `Any`? Can we just remove this afterwards, or does something else need to be changed as well?

ilevkivskyi: This will cause named tuple class object to be inferred as `Any` in runtime context. This is actually also related to https://github.com/python/mypy/issues/6458

I will update the comment.

ilevkivskyi: Just to give a bit more details: since `visit_namedtuple_expr()` returns `AnyType` in type checker, the binder will assign this to the `lvalue` (by its pseudo-hash), so despite the fact that in the symbol table this name points to the `TypeInfo`, it will be "narrowed down" to `Any`.

PR #6455, opened 2019-02-20
Refactoring: remove is_import argument from add_unknown_symbol()

ilevkivskyi:


mypy/newsemanal/semanal.py
```diff
@@ -4062,14 +4062,13 @@ def add_imported_symbol(self, name: str, node: SymbolTableNode, context: Context
                                  module_hidden=module_hidden)
         self.add_symbol_table_node(name, symbol, context)
 
-    def add_unknown_symbol(self, name: str, context: Context, is_import: bool = False,
+    def add_unknown_symbol(self, name: str, context: Context,
                            target_name: Optional[str] = None) -> None:
         """Add symbol that we don't know what it points to (due to error, for example)."""
```
JukkaL: Should the docstring be updated to mention that this is an unknown symbol from an import? Maybe the method name should be renamed as well to reflect this?

ilevkivskyi: Yes, I will make another PR now.

PR #6465, opened 2019-02-22
New semantic analyzer: support class variables

### LINKED ISSUE #6451: New semantic analyzer: ClassVar
Inheriting the class variable bit is a bit tricky. We can't do it fully in the main pass, since base class bodies may not have been fully processed yet. The old semantic analyzer does this in pass 3, which doesn't exist in the new semantic analyzer. The class variable bit needs to be set before type checking. Maybe the best option is to do this similar to how abstract status is calculated, in a pass between semantic analysis and type checking.

JukkaL:
This adds an additional simple pass over all classes that sets
`is_classvar` flags for class variables declared in base classes.

Also refactor to share code used for symbol table iteration.

Fixes #6451.

### DIFF
--- a/mypy/newsemanal/semanal_main.py
+++ b/mypy/newsemanal/semanal_main.py
@@ -27,12 +27,13 @@ will be incomplete.
 from typing import List, Tuple, Optional, Union, Callable
 
 from mypy.nodes import (
-    Node, SymbolTable, SymbolNode, MypyFile, TypeInfo, FuncDef, Decorator, OverloadedFuncDef
+    Node, SymbolTable, SymbolNode, MypyFile, TypeInfo, FuncDef, Decorator, OverloadedFuncDef,
+    Var
 )
 from mypy.newsemanal.semanal_typeargs import TypeArgumentAnalyzer
 from mypy.state import strict_optional_set
 from mypy.newsemanal.semanal import NewSemanticAnalyzer, apply_semantic_analyzer_patches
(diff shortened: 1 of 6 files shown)


mypy/newsemanal/semanal_main.py
```diff
@@ -230,8 +219,26 @@ def check_type_arguments(graph: 'Graph', scc: List[str], errors: Errors) -> None
...
         tree = graph[module].tree
         assert tree
-        calculate_abstract_status(tree, errors)
+        for _, node, _ in tree.local_definitions():
+            if isinstance(node.node, TypeInfo):
+                calculate_class_abstract_status(node.node, tree.is_stub, errors)
+                calculate_class_vars(node.node)
+
+
+def calculate_class_vars(info: TypeInfo) -> None:
```
ilevkivskyi: Should this leave in `semanal_main`? Maybe move it to `semanal_abstract` and rename the latter to `semanal_classes` (or `semanal_util`)?

mypy/newsemanal/semanal_main.py
```diff
@@ -230,8 +219,26 @@ def check_type_arguments(graph: 'Graph', scc: List[str], errors: Errors) -> None
...
+    # Subclass attribute assignments with no type annotation should be
+    # assumed to be classvar if overriding a declared classvar from the base
+    # class.
```
ilevkivskyi: I would propose to turn comment into docstring.

PR #6469, opened 2019-02-22
Fixing error message to match new typeshed stub

mrkmndz:
As requested from https://github.com/python/typeshed/pull/2810.  Not sure how to sync the typeshed commit if that one isn't merged yet.


JelleZijlstra:
Yes, I'll do the merge + typeshed sync for this later today.

PR #6474, opened 2019-02-24
getting started guide refers to non-existant switch 

smcl:
There's a minor typo in the "Getting started" page where the switch `--disallow-unchecked-defs` is mentioned, but using this results in an error saying it's an unrecognized argument. I'm just getting started with the project personally, but I believe this should instead be `--disallow-untyped-defs`


Michael0x2a:
Good catch -- thanks for the fix!

Issue #4286, opened 2017-11-27
Taking a slice that contains unions produces Any


rserbitar:
```
def a() -> typing.Union[typing.Tuple[str, str, str],
                        typing.Tuple[typing.List[str], typing.List[str], typing.List[str]]]:
    return '1', '1', '1'


def b() -> typing.Union[typing.Tuple[str, str, str],
                        typing.Tuple[typing.List[str], typing.List[str], typing.List[str]]]:
    return a()


def c() -> typing.Union[typing.Tuple[str, str, str],
                        typing.Tuple[typing.List[str], typing.List[str], typing.List[str]]]:
    x, y, z = a()
    return x, y, z
```

Result:
`
error: Incompatible return value type (got "Tuple[Union[str, List[str]], Union[str, List[str]], Union[str, List[str]]]", expected "Union[Tuple[str, str, str], Tuple[List[str], List[str], List[str]]]")
`

Why?

ilevkivskyi:
Because after you unpacked, there are three independent variables ``x``, ``y``, and ``z`` each with type ``Union[str, List[str]]``, mypy does not store any cross-variable constraints like if ``x`` is ``str`` then ``y`` should be also ``str``. And I doubt it ever will.

rserbitar:
This however works fine:

```
def a() -> typing.Union[typing.Tuple[str, str, str],
                        typing.Tuple[typing.List[str], typing.List[str], typing.List[str]]]:
    return '1', '1', '1'

def c() -> typing.Union[typing.Tuple[str, str, str],
                        typing.Tuple[typing.List[str], typing.List[str], typing.List[str]]]:
    x, y, z = a()[:]
    return x, y, z
```

ilevkivskyi:
> This however works fine:

And this is a bug!
```python
def c() -> typing.Union[typing.Tuple[str, str, str],
                        typing.Tuple[typing.List[str], typing.List[str], typing.List[str]]]:
    x, y, z = a()[:]
    reveal_type(x)  # Revealed type is Any
    return x, y, z
```
I think however this is not an independent problem, but a problem of interactions between unions and overloads. @JukkaL what do you think?

JukkaL:
Looks like there are two problems. First, the fallback of the tuple types seems to be `Tuple[Any, ...]`, which is not right. Second, we should special case unions in type inference of indexing expressions. Currently it just falls back to calling `__getitem__` which loses the special semantics for tuple types.

ilevkivskyi:
[...]

Issue #2509, opened 2016-11-29
Union of different-length Tuple types should agree for corresponding items


gvanrossum:
See https://github.com/python/typeshed/pull/721#discussion_r90098017. The issue is that after
```python
T = Union[Tuple[int, int], Tuple[int, int, str]]
def f() -> T: pass
a = f()
reveal_type(a[0])
reveal_type(a[2])
```
the item types are considered to be Any. A better approach would be to understand that a[0] is always an int while a[2] is invalid without some check (however we can't do isinstance() checks on tuple lengths so that's problematic).

JukkaL:
We could support `len(x)` checks for tuple unions: #1178

rwbarton:
I'm a little surprised this doesn't go through the path for a method call on a union. For example, this works fine:
```py
class A:
    def f(self) -> int: pass

class B:
    def f(self) -> int: pass

U = Union[A, B]
def g() -> U: pass

b = g()
reveal_type(b.f())   # Revealed type is 'builtins.int'
```
It might just take some code reorganization (translate the IndexExpr into a `__getitem__` method call, then handle `__getitem__` on tuples in `check_op`).

PR #6475, opened 2019-02-24
Support indexing unions containing tuples

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/4286
Fixes https://github.com/python/mypy/issues/2509

The fix is straightforward, map the indexing through union types. I decided to use the existing error message mentioning the full type. IMO `Value of type "Union[int, List[int]]" is not indexable` is better than `Value of type "int" is not idexable`.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2455,7 +2455,9 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         """
         index = e.index
         if isinstance(left_type, UnionType):
-            return UnionType.make_simplified_union([self.visit_index_with_type(typ, e, left_type)
+            original_type = original_type or left_type
+            return UnionType.make_simplified_union([self.visit_index_with_type(typ, e,
+                                                                               original_type)
                                                     for typ in left_type.relevant_items()])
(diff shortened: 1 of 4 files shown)


mypy/checkexpr.py
```diff
@@ -2441,18 +2445,29 @@ def visit_index_expr_helper(self, e: IndexExpr) -> Type:
...
+
+    def visit_index_with_type(self, left_type: Type, e: IndexExpr,
+                              original_type: Optional[Type] = None) -> Type:
+        """Analyze type of an index expression for a given type of base expression.
+
+        The 'original_type' is used for error messages (currently used for union types).
+        """
+        index = e.index
+        if isinstance(left_type, UnionType):
+            return UnionType.make_simplified_union([self.visit_index_with_type(typ, e, left_type)
```
Michael0x2a: Just to check, should we pass `original_type` here as well?

ilevkivskyi: Normally, there should be no nested unions because they are flattened in the constructor, but I agree that it would be more robust to pass it also here.

PR #6481, opened 2019-02-25
Add a SYMBOL_FUNCBASE_TYPES constant to avoid multiple-inheritance bug

msullivan:
A mypy bug (#3603) causes mypy to think that SymbolNodes cannot be
FuncBases, which causes mypy-mypyc to crash in code that tests
that.
Currently we work around this in a couple gross ways (like turning the
SymbolNode into Any). Work around it instead by testing directly
against a tuple of the subtypes instead of FuncBase itself.

This is the better fix promised in #6480.

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -1,6 +1,6 @@
 """Type checking of attribute access"""
 
-from typing import cast, Callable, List, Optional, TypeVar, Any
+from typing import cast, Callable, List, Optional, TypeVar
 
 from mypy.types import (
     Type, Instance, AnyType, TupleType, TypedDictType, CallableType, FunctionLike, TypeVarDef,
@@ -10,7 +10,7 @@ from mypy.types import (
 from mypy.nodes import (
     TypeInfo, FuncBase, Var, FuncDef, SymbolNode, Context, MypyFile, TypeVarExpr,
     ARG_POS, ARG_STAR, ARG_STAR2, Decorator, OverloadedFuncDef, TypeAlias, TempNode,
-    is_final_node
+    is_final_node, SYMBOL_FUNCBASE_TYPES,
 )
 from mypy.messages import MessageBuilder
 from mypy.maptype import map_instance_to_supertype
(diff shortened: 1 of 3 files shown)


mypy/checkmember.py
```diff
@@ -624,12 +624,10 @@ def analyze_class_attribute_access(itype: Instance,
...
         # Find the class where method/variable was defined.
-        # mypyc hack to workaround mypy misunderstanding multiple inheritance (#3603)
-        node_node = node.node  # type: Any
-        if isinstance(node_node, Decorator):
-            super_info = node_node.var.info  # type: Optional[TypeInfo]
-        elif isinstance(node_node, (Var, FuncBase)):
-            super_info = node_node.info
+        if isinstance(node.node, Decorator):
+            super_info = node.node.var.info  # type: Optional[TypeInfo]
+        elif isinstance(node.node, (Var, SYMBOL_FUNCBASE_TYPES)):
```
gvanrossum: So the hack now is that this is a variable and the binder doesn't know whether it's true or false?

mypy/nodes.py
```diff
@@ -403,7 +403,18 @@ def __str__(self) -> str:
...
+
+    N.B: Although this has SymbolNode subclasses (FuncDef,
+    OverloadedFuncDef), avoid calling isinstance(..., FuncBase) on
+    something that is typed as SymbolNode.  This is to work around
+    mypy bug #3603, in which mypy doesn't understand multiple
+    inheritance very well, and will assume that a SymbolNode
+    cannot be a FuncBase.
+
+    Instead, test against SYMBOL_FUNCBASE_TYPES, which enumerates
+    SymbolNode subclasses that are also FuncBase subclasses.
```
gvanrossum: Sounds tedious. :-(

mypy/nodes.py
```diff
@@ -668,6 +679,11 @@ def deserialize(cls, data: JsonDict) -> 'FuncDef':
         return ret
 
 
+# All types that are both SymbolNodes and FuncBases. See the FuncBase
+# docstring for the rationale.
+SYMBOL_FUNCBASE_TYPES = (OverloadedFuncDef, FuncDef)
```
gvanrossum: How do we ensure this stays up to date? If a new class were to be added, would tests for that new class fail clearly?

mypy/nodes.py
```diff
@@ -2857,10 +2873,7 @@ def fullname(self) -> Optional[str]:
     @property
     def type(self) -> 'Optional[mypy.types.Type]':
         node = self.node
-        if (isinstance(node, Var) and node.type is not None):
-            return node.type
-        # mypy thinks this branch is unreachable but it is wrong (#3603)
-        elif (isinstance(node, FuncBase) and node.type is not None):
+        if (isinstance(node, (Var, SYMBOL_FUNCBASE_TYPES)) and node.type is not None):
```
gvanrossum: Style nit: the outer parentheses aren't needed (nor were they before...).

mypy/nodes.py
msullivan: They would fail, I think, though I'm not sure how cleanly.

mypy/checkmember.py
msullivan: There's no hack, anymore, really. We're just explicitly enumerating the overlap between SymbolNode and FuncBase instead of using isinstance on FuncBase

PR #6484, opened 2019-02-26
New semantic analyzer: enable tests and fix to fullnames

JukkaL:
This enables various tests for the new semantic analyzer. It also
fixes an issue with a local definition getting a fully qualified name.

Some tests have different output on new vs old semantic analyzer.
If the output on the new semantic analyzer was reasonable, I forced
the test to always run on it. In a few cases an existing issue had
been fixed on the new semantic analyzer.


test-data/unit/semanal-types.test
```diff
@@ -180,14 +180,16 @@ x = None # type: A[B]
...
+  ImportFrom:2(typing, [TypeVar, Generic, Any])
   AssignmentStmt:3(
     NameExpr(t* [__main__.t])
     TypeVarExpr:3())
   ClassDef:5(
     A
     TypeVars(
       t)
+    BaseType(
+      builtins.object)
```
ilevkivskyi: Maybe we can just update the test runner to remove the `object` base class before dumping?

JukkaL: I initially didn't do this because this would skip some testing of base types, but in retrospect it now seems worth it, as we can run the same test cases with both the old and the new semantic analyzers.

PR #6486, opened 2019-02-27
Fix travis

msullivan:
PR #6175 broke travis with some bogus indentation.
Fix that.


msullivan:
(I'm just gonna merge this once it is green)

Issue #6463, opened 2019-02-21
No type inference for dicts in union


ikonst:
```py
from typing import *

def foo() -> Dict[str, Any]:
    return {}

def bar() -> Union[Dict[str, Any], Dict[int, Any]]:
    return {}
```

In `foo`, `{}` is inferred as `Dict[str, Any]`.
In `bar`, I get:
```
error: Incompatible return value type (got "Dict[<nothing>, <nothing>]", expected "Union[Dict[str, Any], Dict[int, Any]]")
```

gvanrossum:
I’m sure the cause is the union, not the tuple.

ikonst:
@gvanrossum thanks, you're right - corrected the issue title and the sample code

oraluben:
@ikonst you could try this to have a workaround, though the semantics has changed
```
def bar() -> Dict[Union[str, int], Any]:
    return {}
```

ilevkivskyi:
This may be a duplicate of https://github.com/python/mypy/issues/2164

ikonst:
@oraluben - the actual code is more complex:
```py
from typing import *

from flask import Response

JsonDict = Dict[str, Any]

_FlaskResponse = Union[
    str,
    bytes,
    Callable,
    Response,
]
_FlaskRestfulResponse = Union[
    JsonDict,
    _FlaskResponse,
]
FlaskRestfulMethodReturnType = Union[
    _FlaskRestfulResponse,
    Tuple[Optional[_FlaskRestfulResponse], int],  # (response, status_code)
    Tuple[Optional[_FlaskRestfulResponse], int, Dict[str, str]],  # (response, status_code, headers)
    Tuple[Optional[_FlaskRestfulResponse], Dict[str, str]],  # (response, headers)
]
```

and then this works:
```py
class MyResource(Resource):
   def get() -> FlaskRestfulMethodReturnType:
       return {}
```

and this works:
```py
class MyResource(Resource):
   def get() -> FlaskRestfulMethodReturnType:
       return {"foo": "bar"}, 200
```

but this doesn't:
and this works:
```py
class MyResource(Resource):
   def get() -> FlaskRestfulMethodReturnType:
       return {}, 200
```

JukkaL:
This happens with other invariant containers as well, such as lists. I just encountered an instance where mypy couldn't infer the type of a `{}` function argument when calling `string.Template.substitute`, since the type of the argument in the signature is a union.

ruuda:
[...]

Issue #3283, opened 2017-04-29
List[<uninhabited>] error


jcrmatos:
Hello,

Already showed this in Gitter and was told to open the issue here.
My env is Win7P+SP1 x64, Py 3.5.2 32b and mypy 0.501.

I have the following situation:

With the following code mypy doesn't return an error (which is correct):

```
from datetime import datetime, timedelta

data_dic = {idx: []
            for idx in range(len(lcl.msgs['DATA_COL_NAMES']) - 1)}  # type: Dict[int, List[Union[datetime, float]]]

```

But with the following code mypy returns a strange error

```
from datetime import datetime, timedelta

data_dic = {idx: []
            for idx in range(len(lcl.msgs['DATA_COL_NAMES']) - 1)}  # type: Dict[int, Union[List[datetime], List[float]]]

rtg_main.py:197: error: Value expression in dictionary comprehension has incompatible type List[<uninhabited>]; expected type "Union[List[datetime], List[float]]"
```

Best regards,

JM

gvanrossum:
Maybe @ddfisher understands this? IMO it's always an error when `<uninhabited>` shows up in error messages?

ilevkivskyi:
@gvanrossum 
> IMO it's always an error when ``<uninhabited>`` shows up in error messages?

Unfortunately, currently it is quite easy to get ``<uninhabited>`` in an error (TBH I don't like this). For example:
```python
T = TypeVar('T')
def fun(x: List[T]) -> T:
    ...
fun(1) # Error: Argument 1 to "fun" has incompatible type "int"; expected List[<uninhabited>]
```
I think such errors are quite cryptic, so I set priority to high here.

ddfisher:
My understanding is that this is a failure of type variable inference -- the context mechanism isn't working fully here.  Prior to #3024, I think these error messages would reference `List[None]` instead of `List[<uninhabited>]`, but would be similarly cryptic otherwise.  The fact that uninhabited type is inferred here is a bug IMO and is tracked in #3032.

The fact that there's an error here at all is a problem with the context mechanism (perhaps related to Unions?).

JukkaL:
In the original example, it looks like we don't infer a type because it's ambiguous due to the union type context. This is a known issue, though it might only be mentioned in a comment in the code. Here's a simpler repro:

```
from typing import Union, List
a = []  # type: Union[List[int], List[str]]
```

Here I can see a few plausible options:
[...]

Issue #2164, opened 2016-09-21
Error initializing Union of Lists with an empty list


elazarg:
The following test case

```
[case testUnionListInit]
from typing import Union, List
a = []  # type: Union[List[int], List[str]]
[builtins fixtures/list.pyi]
```

Results in

```
Expected:
Actual:
  main:2: error: Incompatible types in assignment (expression has type List[None], variable has type "Union[List[int], List[str]]") (diff)
```

This came up during PR #2154, but I think is an independent issue.

elazarg:
Perhaps related to [binder.py#L189](https://github.com/python/mypy/blob/master/mypy/binder.py#L189)?

gvanrossum:
Doesn't sound like a very important issue, the example has questionable value. (I'm not saying it's not a bug -- just that I don't think it's urgent to fix it.)

JukkaL:
Closing in favor of duplicate #6463 (the latter issue has a slight more interesting example).

PR #6487, opened 2019-02-27
join constraints with Any when constraints are in same shape

oraluben:
Fix [#6463](https://github.com/python/mypy/issues/6463), [#2164](https://github.com/python/mypy/issues/2164) by join Union with same shape with `Any`, for example, `Union[List[int], List[str]]` -> `List[Any]`, I use `Any` because `List[Union[str, int]]` is obviously wrong.

according to [this](https://github.com/python/mypy/issues/3283#issuecomment-298689038), 
> * Infer `Union[List[int], List[str]]` for the type of `[]`. This might be tricky to implement -- basically we'd need to support multiple possible answers from type inference, whereas currently we assume there's just one.

another way to fix that issue is to adjust `any_constraints` and `solve_constraints` to solve Union on any position of a type, currently it seems to work only on the 'leaf node'.

I'm not sure if this is a proper way to fix that, please have a look at your convenience.

### DIFF
--- a/mypy/constraints.py
+++ b/mypy/constraints.py
@@ -183,7 +183,9 @@ def any_constraints(options: List[Optional[List[Constraint]]], eager: bool) -> L
         #       every option, combine the bounds with meet/join.
         elif all(len(option) == len(valid_options[0]) for option in valid_options[1:]):
             # same length, more shape check below
-            to_join_constraint_group = [list(c_t) for c_t in zip(*valid_options)]  # type: List[List[Constraint]]
+            to_join_constraint_group = [
+                list(c_t) for c_t in zip(*valid_options)
+            ]  # type: List[List[Constraint]]
             joined_constraints = []  # type: List[Constraint]
             for to_join_constraint_list in to_join_constraint_group:
                 if any(
@@ -195,7 +197,6 @@ def any_constraints(options: List[Optional[List[Constraint]]], eager: bool) -> L
                 joined_constraints.append(to_join_constraint_list[0])
                 if not all(is_same_constraint(to_join_constraint_list[0], c)
                            for c in to_join_constraint_list[1:]):
-                    # Multiple sets of constraints that are all the same. Just pick any one of them.
                     joined_constraints[-1].target = AnyType(TypeOfAny.explicit)
             else:
                 # not break from loop
--- a/mypy/constraints.py
+++ b/mypy/constraints.py
@@ -174,13 +174,32 @@ def any_constraints(options: List[Optional[List[Constraint]]], eager: bool) -> L
         valid_options = [option for option in options if option is not None]
     if len(valid_options) == 1:
         return valid_options[0]
-    elif (len(valid_options) > 1 and
-          all(is_same_constraints(valid_options[0], c)
-              for c in valid_options[1:])):
-        # Multiple sets of constraints that are all the same. Just pick any one of them.
+    elif len(valid_options) > 1:
+        if all(is_same_constraints(valid_options[0], c)
+               for c in valid_options[1:]):
+            # Multiple sets of constraints that are all the same. Just pick any one of them.
+            return valid_options[0]
         # TODO: More generally, if a given (variable, direction) pair appears in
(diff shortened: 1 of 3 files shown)


oraluben:
this do not fix [this](https://github.com/python/mypy/issues/6463#issuecomment-467373559), so I think a good way is to treat `List[<nothing>]` as subset of `List[<whatever>]`, which currently is not.

ilevkivskyi:
I think this is not a good solution TBH. It just masks the problem instead of solving it. For example, IIUC in the original example in the issue you propose to solve `List[T] <: Union[List[str], List[int]]` with just `T` being `Any`. This is unsafe.

The option of making `List[<nothing>]` a subtype of `List[int]` appeared couple times recently, but it may be risky because we rely on `<nothing>` being a real bottom type at least in some places (e.g. protocol variance checks). It is not clear what we should do with these.

I am leaning towards the tricky option @JukkaL mentions. Essentially we would need to avoid pushing unions on the type context stack (when type-checking r.h.s. of an assignment, and when type-checking arguments in a call). Instead we would need to push every item of a union, type-check the expression in this context, and then union the results. We already do something similar for type-checking overload calls with union arguments.

This latter approach looks potentially promising from the point of view of quality of inference results, however it might be tricky. Also performance implications are not clear.

oraluben:
Yes you'r right, this does not actually solve the problem.

And I have also test if just treat every `<nothing>` as subtype of `anything`, and it do break when inferring `Protocol`s, so I have a version that do this only in `dict`, `list`, `set`, and `tuple`, which will have `<nothing>` when initialed empty.
I don't think this is a good solution neither, for it breaks the consistency when handling `<nothing>`. Just push this to have a reference.

I will see what can be done to infer `Union`, thanks for your reply!

JukkaL:
Note that making `List[<nothing>]` a subtype of `List[int]` would not be safe, since lists are mutable. Assume we have instance of `List[<nothing>]`. We could pass that to a function expecting `List[int]`, which could append an integer. We could also pass this `List[<nothing>]` object to a function expecting `List[str]`, and it could read an integer from a supposed `List[str]`.

ilevkivskyi:
@JukkaL Argument of Pyre people (who actually consider it a subtype) is that if we prohibit this type to appear as a type of _variable_ (btw mypy partially prohibits this), then the unsafety is practically negligible. On one hand this sounds reasonable, but on other hand, I don't like this kind of special-casing. Plus it will break protocol checks.

oraluben:
Personally, I think `List` should not have `<nothing>` as its variable, it's more intuitive that empty `List` should have a `List[Any]` type.

gvanrossum:
@oraluben Do you understand how mypy generally handles empty lists? There are two special cases.

**Case 1:** Type inferred from context:
```
def foo(a: List[T]) -> List[T]:
    ...
x: List[int] = foo([])
```
Mypy does some constraint solving here, where the expected return type of the call to `foo()` is used to infer the right value for `T` (in this example `int`).

A simpler case is:
```
def foo(a: List[int]):
    ...
foo([])
```
Again, the type of `[]` is inferred from the context, which is fixed here (`List[int]`).

You can even check this by writing `foo(reveal_type([]))` -- the constraint solving "passes through" `reveal_type()`.

**Case 2:** Type inferred from peeking at later assignment:
```
a = []
a.append(1)
```
Here the type of `a` is inferred to be `List[int]`. This is done in a very different way -- mypy literally looks ahead in the function for the next use of `a`, and checks whether it is a simple operation like `a.append()`.

**Otherwise:** If neither of these cases holds, the type of `[]` is `List[<nothing>]`, and this is always supposed to give an error:
```
a = []  # E: Need type annotation for 'a'
if a:
    a.append(1)
```
Here `a` is used before the type is determined, and the "peek" operation doesn't work.

You get the same error (though for a different reason) if you try to split the example from case 1 in two:
```
def foo(a: List[T]) -> List[T]:
    ...
a = []  # E:  Need type annotation for 'a'
x: List[int] = foo(a)
```
Here the type of `a` cannot be affected by its use in the context of `foo(a)`.

PR #6489, opened 2019-02-27
Revert "Fixing error message to match new typeshed stub"

ilevkivskyi:
Reverts python/mypy#6469

See https://github.com/python/typeshed/pull/2817 for motivation.


ilevkivskyi:
One test will fail, we will need to re-sync typeshed to get it green again.

PR #6497, opened 2019-03-01
New semantic analyzer: fix issues and enable tests

JukkaL:
Fix various issues that caused failures in the remaining
disabled semantic analyzer tests. Also enable most of the
remaining semantic analyzer tests. A few tests are skipped
because the fix seems nontrivial or is already being worked
on elsewhere.


mypy/build.py
```diff
@@ -1748,7 +1748,7 @@ def check_blockers(self) -> None:
             self.manager.errors.raise_error()
 
     @contextlib.contextmanager
-    def wrap_context(self) -> Iterator[None]:
+    def wrap_context(self, check_blockers: bool = True) -> Iterator[None]:
```
ilevkivskyi: I would add a comment with motivation for this. My understanding is that if a blocker occurs in top-level, we still need to process functions, to maintain backwards compatibility with the old analyzer.

PR #6499, opened 2019-03-01
New semantic analyzer: remove the limit on max number of iterations

### LINKED ISSUE #6485: New semantic analyzer: Finish iterations after no names were added/modified
Currently we defer reporting undefined names until we reach iteration limit. A better solution would be to track whether any new names were added, and if not, make one more iteration where report all errors.

Note that this will also allow a cleaner solution for https://github.com/python/mypy/issues/6412

JukkaL:
Keep track of added names. If an iteration didn't add anything new, 
any further work is meaningless and the next iteration will be the last
one.

Avoid infinite loops by failing if the final iterations defers anything.

Fixes #6485.

Note that this doesn't deal with normal placeholder being changed
into a typeinfo placeholder, since it wasn't needed for any tests.
I can fix this if we can come up with a test case.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -214,13 +214,13 @@ class NewSemanticAnalyzer(NodeVisitor[None],
     # Stack of functions being analyzed
     function_stack = None  # type: List[FuncItem]
 
-    # Made True if semantic analysis defines a name, or makes a definition
-    # more precise. If some iteration makes no progress, the next iteration
-    # is going to be the final one.
+    # Set to True if semantic analysis defines a name, or replaces a
+    # placeholder definition. If some iteration makes no progress,
+    # there can be at most one additional final iteration (see below).
     progress = False
 
-    # Is this the final iteration of semantic analysis  (when we report
-    # unbound names)?
+    # Is this the final iteration of semantic analysis (where we report
+    # unbound names due to cyclic definitions)?
     _final_iteration = False
 
     loop_depth = 0         # Depth of breakable loops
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -4052,15 +4063,19 @@ def add_symbol_table_node(self, name: str, symbol: SymbolTableNode,
         existing = names.get(name)
         if (existing is not None
                 and context is not None
-                and not isinstance(existing.node, PlaceholderNode)):
-            if existing.node != symbol.node:
+                and (not isinstance(existing.node, PlaceholderNode)
+                     or isinstance(symbol.node, PlaceholderNode))):
+            if (existing.node != symbol.node
+                    and not (isinstance(existing.node, PlaceholderNode)
+                             and isinstance(symbol.node, PlaceholderNode))):
```
ilevkivskyi: TBH, this bunch of `if`s is hard to read, I think it makes sense to either refactor it, or add explanatory comments.

mypy/newsemanal/semanal_main.py
```diff
@@ -47,9 +47,6 @@
 Patches = List[Tuple[int, Callable[[], None]]]
 
 
-# Perform up to this many semantic analysis iterations until giving up trying to bind all names.
-MAX_ITERATIONS = 10
-
```
ilevkivskyi: I think we should keep this, but if we ever hit this, just immediately fail with `AssertionError`, without any additional final pass. Just to avoid infinite loops.

PR #6500, opened 2019-03-01
Add support for testing order of processing targets

ilevkivskyi:


mypy/server/update.py
```diff
@@ -764,7 +768,7 @@ def replace_modules_with_new_variants(
 
 
 def propagate_changes_using_dependencies(
-        manager: BuildManager,
+        manager: FineGrainedBuildManager,
```
JukkaL: This couples the function more tightly with `FineGrainedBuildManager` and performs additional mutation that is not visible in the interface of the function. It would be better to pass `process_targets` as an extra argument (and preserve the old `manager` argument`). This might also reduce code churn.

test-data/unit/fine-grained.test
```diff
@@ -8670,6 +8670,36 @@ def f(x: int) -> None: ...
 from typing import Union
 def f(x: Union[int, str]) -> None: ...
 
+[targets c, b]
+[targets2 a]
```
JukkaL: The numbering doesn't seem to be in sync. I think that it would be clearer if `[targets2 ...]` would correspond to change `c.py.2` and `[targets3 ...]` would correspond to `a.py.3` (i.e. increase numbers by one).

test-data/unit/fine-grained.test
```diff
@@ -8670,6 +8670,36 @@ def f(x: int) -> None: ...
...
+class A: ...
+class B(A): ...
+def f() -> int: ...
+
+[file b.py.3]
+class A: ...
+class B(A): ...
+def f() -> str: ...
+
+[targets b, __main__, __main__.C.meth, __main__, __main__.C.meth]
```
JukkaL: Why is `__main__` processed twice?

mypy/server/update.py
```diff
@@ -881,7 +885,7 @@ def find_targets_recursive(
     return result, unloaded_files, stale_protos
 
 
-def reprocess_nodes(manager: BuildManager,
+def reprocess_nodes(manager: FineGrainedBuildManager,
```
JukkaL: This is similar to above.

mypy/server/update.py
ilevkivskyi: OK

test-data/unit/fine-grained.test
ilevkivskyi: OK

test-data/unit/fine-grained.test
ilevkivskyi: This is because of `<C> -> C`. After we first reprocess `C` we detect that MRO changed, so `<C>` is triggered. I agree this is suboptimal, but I don't think there is a simple way to avoid this.

ilevkivskyi:
@JukkaL Thanks for review! I implemented all requested changes.

PR #6502, opened 2019-03-01
New semantic analyzer: Sync aststrip with analyze_member_lvalue()

ilevkivskyi:
This is a follow-up for #6488 

The test I added is already good, because it crashes without the fix. When https://github.com/python/mypy/pull/6500 is merged I will add a check that only necessary targets are processed.

### DIFF
--- a/mypy/server/aststripnew.py
+++ b/mypy/server/aststripnew.py
@@ -86,6 +86,7 @@ class NodeStripVisitor(TraverserVisitor):
         for name, sym in node.info.names.items():
             if isinstance(sym.node, Var) and sym.implicit:
                 explicit_self_type = sym.node.explicit_self_type
+
                 def patch() -> None:
                     existing = node.info.get(name)
                     current_existing = node.info.names.get(name)
--- a/mypy/server/aststripnew.py
+++ b/mypy/server/aststripnew.py
@@ -85,13 +85,15 @@ class NodeStripVisitor(TraverserVisitor):
         """Produce callbacks that re-add attributes defined on self."""
         for name, sym in node.info.names.items():
             if isinstance(sym.node, Var) and sym.implicit:
+                explicit_self_type = sym.node.explicit_self_type
                 def patch() -> None:
                     existing = node.info.get(name)
+                    current_existing = node.info.names.get(name)
                     # This needs to mimic the logic in SemanticAnalyzer.analyze_member_lvalue()
                     # regarding the existing variable in class body or in a superclass.
-                    # TODO: copy the logic better (always create a Var if there is explicit type)
-                    if (not existing or
-                            isinstance(existing.node, Var) and existing.node.is_abstract_var):
+                    if (existing is None or
+                            isinstance(existing.node, Var) and existing.node.is_abstract_var or
+                            current_existing is None and explicit_self_type):
                         node.info.names[name] = sym
 
                 self.patches.append(patch)
--- a/mypy/server/aststripnew.py
+++ b/mypy/server/aststripnew.py
@@ -89,12 +89,16 @@ class NodeStripVisitor(TraverserVisitor):
 
                 def patch() -> None:
                     existing = node.info.get(name)
-                    current_existing = node.info.names.get(name)
+                    defined_in_this_class = name in node.info.names
                     # This needs to mimic the logic in SemanticAnalyzer.analyze_member_lvalue()
-                    # regarding the existing variable in class body or in a superclass.
+                    # regarding the existing variable in class body or in a superclass:
+                    # If the attribute of self is not defined in superclasses, create a new Var.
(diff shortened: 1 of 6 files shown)


mypy/server/aststripnew.py
```diff
@@ -85,13 +85,16 @@ def prepare_implicit_var_patches(self, node: ClassDef) -> None:
...
                 def patch() -> None:
                     existing = node.info.get(name)
+                    current_existing = node.info.names.get(name)
                     # This needs to mimic the logic in SemanticAnalyzer.analyze_member_lvalue()
                     # regarding the existing variable in class body or in a superclass.
-                    # TODO: copy the logic better (always create a Var if there is explicit type)
-                    if (not existing or
-                            isinstance(existing.node, Var) and existing.node.is_abstract_var):
+                    if (existing is None or
+                            isinstance(existing.node, Var) and existing.node.is_abstract_var or
```
JukkaL: This check is not very clear. Maybe also check that `current_existing` is `None` here, i.e. something like this:

```
if (existing is None or
    (current_existing is None and ((isinstance(existing.node, Var) and existing.node.is_abstract_var) or
                                    explicit_self_type))):
```

Style nit: I prefer to use parens to make precedence explicit when combining `and` / `or` in a single expression.

mypy/server/aststripnew.py
```diff
@@ -85,13 +85,16 @@ def prepare_implicit_var_patches(self, node: ClassDef) -> None:
         """Produce callbacks that re-add attributes defined on self."""
         for name, sym in node.info.names.items():
             if isinstance(sym.node, Var) and sym.implicit:
+                explicit_self_type = sym.node.explicit_self_type
+
                 def patch() -> None:
                     existing = node.info.get(name)
+                    current_existing = node.info.names.get(name)
```
JukkaL: The name was a bit confusing. "Current" could mean "current class" or "currently". Maybe rename to something like `defined_in_this_class` make this a boolean: `defined_in_this_class = name in node.info.names`. Actually, it might be better to get rid of this variable and move the check below to the if statement.

mypy/server/aststripnew.py
```diff
@@ -85,13 +85,16 @@ def prepare_implicit_var_patches(self, node: ClassDef) -> None:
...
                     # This needs to mimic the logic in SemanticAnalyzer.analyze_member_lvalue()
                     # regarding the existing variable in class body or in a superclass.
```
JukkaL: Can you update to comment to also explain what is going on, rather than just pointing out to another method?

test-data/unit/fine-grained.test
```diff
@@ -8673,3 +8673,37 @@ def f(x: Union[int, str]) -> None: ...
...
+import a
+[file a.py]
+from b import B
+
+B().x
+
+[file a.py.3]
+from b import B
+
+x = B().x
```
JukkaL: It's not clear to me why we have this update. Maybe add comment? Also, it might be easier to read this if the updates were in order (.2 comes before .3 and all initial versions come first).

test-data/unit/fine-grained.test
```diff
@@ -8673,3 +8673,37 @@ def f(x: Union[int, str]) -> None: ...
...
+        self.x: int
+
+[file c.py]
+def f(x: int) -> None: ...
+
+[file c.py.2]
+from typing import Union
+def f(x: Union[int, str]) -> None: ...
+
+[out]
```
JukkaL: Would it also make sense to test the explicit protocol base class case and the case where we don't have a type annotation but just assign to `self.x`? What about attribute defined in class body of the derived class? These may be covered by other test cases. If so, maybe verify the updated targets in those test cases as well?

mypy/server/aststripnew.py
ilevkivskyi: I didn't check because it is impossible to have `current_exiting` if `existing` is `None`.

test-data/unit/fine-grained.test
ilevkivskyi: OK, I will add couple of test cases.

mypy/server/aststripnew.py
ilevkivskyi: OK, but I think it is important to have a note that these must be kept strictly in sync.

mypy/server/aststripnew.py
ilevkivskyi: TBH, I would rather keep this defined separately. I don't like large conditions, and with the name the intention is clear.

ilevkivskyi:
Oh, I forgot to set `NEWSEMANAL=1` when running locally, I will look at the failure now.

ilevkivskyi:
OK, all green now. @JukkaL I am going to merge soon, unless you wanted to take a look again.

PR #6510, opened 2019-03-06
Add a mention of mypy playground to README

ilevkivskyi:
The playground is quite convenient for quick experimentation and gist sharing. I have been using it sometimes. It makes sense to add a link to README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,9 @@ programs, even if they have type errors:
 
     $ python3 PROGRAM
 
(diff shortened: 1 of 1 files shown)


msullivan:
I like this, but should we check with Yusuke Miyazaki first?

ilevkivskyi:
OK, opened https://github.com/ymyzk/mypy-playground/issues/56

ilevkivskyi:
I have got the approval, so merging now.

Issue #6422, opened 2019-02-17
New semantic analyzer: Clean-up logic for classes in functions


ilevkivskyi:
While working on https://github.com/python/mypy/pull/6421 I looked again at logic for storing symbols for local (function- or method-level) classes in `prepare_class_def()`. The logic is too ad-hoc and needs to be cleaned up. Here are some points that should be fixed:
* `defn.fullname` and `info.fullname()` should always coincide
* `info.fullname()` prefix should coincide with the symbol table where the info is actually stored (for serialization)
* things should work for double nested classes
* things should work in fine-grained mode

I think the logic should be as following. Consider this example:
```python
class Out:
    def meth(self) -> None:
        class In:
            def meth(self) -> None:
                class C: pass
```
The class `C` should be stored in its local namespace under name `'C'`. In addition it should be stored in the namespace of `__main__.Out` under name `'C@5'`. `info.fullname()` and `defn.fullname` should be both `'__main__.Out.C@5'`.

MatthieuDartiailh:
What is the current status of this ? I saw that a couple of PRs addressed part of the issue but https://github.com/python/mypy/issues/7281 is not fixed and I am wondering how much of the initial comment still applies after three PR related to this issue. I added typing to a project but this issue will prevent downstream package to benefit of it so I may be interested in helping.

MatthieuDartiailh:
Also how should the system behaves for a class defined in a function ? Should the symbol be stored on the function or at the global level ?

JukkaL:
I'm not sure if anybody is up to date on what still needs to be done (if anything). To get started with this, you could try to document the current behavior and then we can discuss how it might need to be modified.

MatthieuDartiailh:
Thanks @JukkaL ! Any pointer on the best way to inspect mypy internal after parsing a file, I am new to mypy internals ?

MatthieuDartiailh:
Poking at things after parsing (using mypy at 05571b17, point at which I cloned the repo):
```python
from collections import namedtuple

class A:
    pass


def make_class() -> type:
    class Inner:
        pass

    return Inner


B = make_class()


class Out:
    def meth(self) -> type:
        class In:
            def meth(self) -> type:
                class C(namedtuple("_C", ("a")), A): pass
                return C

        return In

D = Out().meth()

E = Out().meth()().meth()
```
I see that the following is true:
- defn.name is always the class short name
- defn.fullname is always the fully qualified name including the line number for nested classes
- those matches what is found on the node

[...]

PR #6527, opened 2019-03-08
New semantic analyzer: Make assignment analysis logic more straightforward

### LINKED ISSUE #6412: New semantic analyzer: refactor analysis of assignments
This is a follow-up for https://github.com/python/mypy/pull/6390

Assignment statements can create a bunch of different symbol nodes:
* Variable
* Type alias
* Type variable
* Named tuple
* Typed dict

Because of this current implementation is messy and has some hacks. We should refactor this to happen in two steps:
* Classify the kind of assignment
* Perform analysis accordingly for each kind

The tricky corner case is enums: `c = Color['Red']` looks like a type alias, but this should create a variable.

### LINKED ISSUE #6453: New semantic analyzer: Remove temporary Var's for TypedDict and NamedTuple
These are used to be compatible with the old semantic analyzer.

ilevkivskyi:
Fixes #6412 
Fixes #6453 

In this PR:
* Wait until we can classify r.h.s. of an assignment as normal, type alias, named tuple etc.
* Choose the corresponding branch instead of performing all of them (as it used to be before)
* Never create (or rely on creation of) temporary `Var`s
* Add every symbol to symbol table only once (as it is done for classes)
* Don't "patch" existing symbols (with one exception for type variables, see below)
* Add and update docstrings and comments

Notes:
* I allow placeholders for type alias targets and NewTypes targets, they essentially behave as base classes. In general I make logic for type aliases and NewTypes much more similar to logic of class definitions (especially NewTypes), this allows to support a whole bunch of tricky forward references
* I don't store types with placeholders in annotations, and instead if an alias target got updated (e.g. placeholders are resolved), I set `progress=True`

Follow-ups:
* There is still a small hack to keep compatibility with old analyzer in type checker, but now it is basically "don't use binder for l.h.s. of special forms"
* Type variable declarations still have a bug that duplicate type variable definition is not reported as error, second one silently wins, this will be fixed in a separate PR
* This PR doesn't aim to fix aspects of assignment analysis related to https://github.com/python/mypy/issues/6422, that will be also a separate PR

Test updates:
* Add and enable more tests for type aliases and NewTypes
* Add test for cyclic definition error messages
* Change order of error messages on the same line in few tests
* Enable one test file where previously only some tests were opted in (I think it was not enabled by mistake in one of the previous PRs, I enable it here since order of errors in one test in that file is affected by this PR.)


mypy/newsemanal/semanal.py
```diff
@@ -1899,48 +1946,113 @@ def is_not_ready_type_ref(self, rv: Expression) -> bool:
                     return True
         return False
 
+    def should_wait(self, rv: Expression) -> bool:
```
JukkaL: The name is pretty generic. What about renaming to `should_wait_rhs`, for example?

mypy/newsemanal/semanal.py
```diff
@@ -1899,48 +1946,113 @@ def is_not_ready_type_ref(self, rv: Expression) -> bool:
...
+        """Can we already classify this r.h.s. of an assignment or should wait?
```
JukkaL: Grammar nit: "should wait" -> "should we wait".

mypy/newsemanal/semanal.py
```diff
@@ -1899,48 +1946,113 @@ def is_not_ready_type_ref(self, rv: Expression) -> bool:
...
+
+        This returns True if we don't have enough information to decide whether
+        an assignment is just a normal variable definition or a special form.
+        Always return False if this is a final iteration, this will typically cause
```
JukkaL: Grammat nit: run-on sentence (use period or semicolon instead of comma).

mypy/newsemanal/semanal.py
```diff
@@ -1899,48 +1946,113 @@ def is_not_ready_type_ref(self, rv: Expression) -> bool:
...
+        the assignment to be classified as variable plus emmit an error.
```
JukkaL: Typo: emmit -> emit

Maybe change 'the assignment' -> 'the lvalue'.

Grammar nit: 'variable' -> 'a variable'

mypy/newsemanal/semanal.py
```diff
@@ -2624,21 +2755,21 @@ def process_typevar_declaration(self, s: AssignmentStmt) -> None:
...
-        node = self.lookup(name, s)
-        assert node is not None
-        assert node.fullname is not None
-        node.kind = self.current_symbol_kind()
-        if isinstance(node.node, TypeVarExpr):
+        if existing and isinstance(existing.node, TypeVarExpr):
             # Existing definition from previous semanal iteration, use it.
-            type_var = node.node
+            # TODO: this may be confused with a duplicate TypeVar definition.
+            # Fix this and add corresponding tests.
```
JukkaL: Style nits: Capitalize 'this' in 'TODO: this'. Indent the second line?

mypy/newsemanal/semanal.py
```diff
@@ -4065,7 +4206,10 @@ def add_symbol_table_node(self, name: str, symbol: SymbolTableNode,
         if (existing is not None
                 and context is not None
                 and (not isinstance(existing.node, PlaceholderNode)
-                     or isinstance(symbol.node, PlaceholderNode))):
+                     or isinstance(symbol.node, PlaceholderNode) and
+                     # Allow replacing becomes_typeinfo=False with becomes_typeinfo=True,
+                     # this can happend for type aliases and NewTypes.
```
JukkaL: Typo: 'happend' -> 'happen'

Grammar nit: run-on sentence (use period instead of comma).

mypy/newsemanal/semanal_newtype.py
```diff
@@ -26,17 +30,36 @@ def __init__(self,
...
+        # add placeholder as we do for ClassDef.
+
+        fullname = self.api.qualified_name(name)
+        if (not call.analyzed or
+                isinstance(call.analyzed, NewTypeExpr) and not call.analyzed.info):
+            # Start from labeling this as future class with becomes_typeinfo=True,
+            # as we do for normal ClassDefs.
+            self.api.add_symbol(name, PlaceholderNode(fullname, s, True), s)
+
+        old_type, should_defer = self.check_newtype_args(name, call, s)
```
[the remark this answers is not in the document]
ilevkivskyi: Unfortunately `base_type` is already taken in this file.

ilevkivskyi:
@JukkaL Thanks for review! I implemented the suggestions.

test-data/unit/check-newtype.test
```diff
@@ -337,7 +339,7 @@ class C(B): pass  # E: Cannot subclass NewType
 from typing import Protocol, NewType
 
 class P(Protocol):
-    attr: int
+    attr: int = 0
```
JukkaL: What's this change?

test-data/unit/check-newsemanal.test
```diff
@@ -339,14 +339,25 @@ def main() -> None:
             x  # E: Name 'x' is not defined
 
 [case testNewAnalyzerCyclicDefinitions]
-gx = gy  # E: Cannot determine type of 'gy'
+gx = gy  # E: Cannot resolve name "gy", possible cyclic definition
```
JukkaL: Grammar nit: run-on sentence in error message. Suggestion: 

>  Cannot resolve name "gy" (possible cyclic definition)

test-data/unit/check-newsemanal.test
```diff
@@ -339,14 +339,25 @@ def main() -> None:
...
 
+[case testNewAnalyzerCyclicDefinitionCrossModule]
+import b
+[file a.py]
+import b
+x = b.x  # E: Cannot determine type of 'x'
+[file b.py]
+import a
+x = a.x  # E: Cannot resolve attribute "x", possible cyclic definition \
+         # E: Module has no attribute "x"
```
JukkaL: Would it be easy to get rid of the second error, as it seems redundant and a bit confusing.

mypy/newsemanal/semanal.py
```diff
@@ -1877,14 +1881,57 @@ def add_type_alias_deps(self, aliases_used: Iterable[str],
...
+                    isinstance(node.args[0], NameExpr)):
+                call = self.lookup_qualified(node.callee.name, node.callee)
+                arg = self.lookup_qualified(node.args[0].name, node.args[0])
+                if (call is not None and call.node and call.node.fullname() == 'builtins.type' and
+                        arg is not None and arg.node and arg.node.fullname() == 'builtins.None'):
+                    return True
+        return False
+
+    def is_type_ref(self, rv: Expression, bare: bool = False) -> bool:
+        """Does this expression refer to a type?
```
JukkaL: The docstring doesn't make it clear whether `C[int]` would be considered a type reference. It would be good to be explicit about this. Also, probably worth mentioning that this assumes a type alias context. It's important since some of the error messages generated below only make sense when defining a type alias.

mypy/newsemanal/semanal.py
```diff
@@ -1899,48 +1946,113 @@ def is_not_ready_type_ref(self, rv: Expression) -> bool:
...
-                # we proceed forward and create a Var. The latter will be replaced with
-                # a type alias it r.h.s. is a valid alias.
-                self.mark_incomplete(expr.name, expr, becomes_typeinfo=top_level_not_ready)
-            return
-        if self.analyze_namedtuple_assign(s):
+                self.mark_incomplete(expr.name, expr)
             return
-        if self.analyze_typeddict_assign(s):
+
+        # The r.h.s. is now ready to be classified, first check if it is a special form:
```
JukkaL: This is much cleaner than what we had previously! Thanks for cleaning this up.

mypy/newsemanal/semanal.py
```diff
@@ -3379,7 +3515,12 @@ def visit_member_expr(self, expr: MemberExpr) -> None:
                 n = self.rebind_symbol_table_node(n)
                 if n:
                     if isinstance(n.node, PlaceholderNode):
-                        self.defer()
+                        if self.final_iteration:
```
JukkaL: This is duplicated above. Maybe move this if statement to a helper method?

test-data/unit/check-newtype.test
ilevkivskyi: As I explained offline, there was previously a false negative in this test (the NewType is abstract otherwise).

test-data/unit/check-newsemanal.test
ilevkivskyi: Opened https://github.com/python/mypy/issues/6551 for this.

ilevkivskyi:
@JukkaL Thanks! I implemented all comments except one (for that I opened a follow-up issue).

PR #6529, opened 2019-03-09
New semantic analyzer: fix error reporting in class_props and enable protocol tests

ilevkivskyi:


mypy/newsemanal/semanal_classprop.py
```diff
@@ -63,6 +64,17 @@ def report(message: str, severity: str) -> None:
                    'note')
 
 
+def check_protocol_status(info: TypeInfo, errors: Errors) -> None:
+    """Check that all classes in MRO of a protocol are protocols"""
+    for type in info.bases:
+        if info.is_protocol:
```
JukkaL: Move this outside the for loop? It would probably be clearer and also a bit more efficient.

PR #6532, opened 2019-03-11
Sync typeshed

gleb-chipiga:


msullivan:
(Closing because we merged a more recent one)

PR #6539, opened 2019-03-12
In Python 3.8, use the stdlib ast instead of typed_ast

gvanrossum:
This required mostly changes to fastparse.py, plus a smattering of other things. The tests still don't all pass, but I believe all remaining errors are due to differences in error messages for `SyntaxError` between typed_ast and Python 3.8's ast (where a bunch of improvements and spelling changes were made, including "can't" -> "cannot").

To test, you need https://github.com/python/cpython/pull/12295 (else it crashes due to `ast.Constant` not having a `kind` field). **UPDATE:** That landed today, March 13.

**UPDATE:** This also requires a typeshed sync to incorporate https://github.com/python/typeshed/pull/2859; I've send a PR to typeshed as https://github.com/python/mypy/pull/6540. **UPDATE^2:** Those have landed.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -61,11 +61,13 @@ try:
             UnaryOp,
             USub,
         )
+
         def ast3_parse(source: Union[str, bytes], filename: str, mode: str,
                        feature_version: int = sys.version_info[1]) -> AST:
             return ast3.parse(source, filename, mode,
                               type_comments=True,  # This works the magic
                               feature_version=feature_version)
+
         NamedExpr = ast3.NamedExpr
         Constant = ast3.Constant
     else:
@@ -87,10 +89,12 @@ try:
             UnaryOp,
             USub,
         )
+
         def ast3_parse(source: Union[str, bytes], filename: str, mode: str,
                        feature_version: int = sys.version_info[1]) -> AST:
             return ast3.parse(source, filename, mode, feature_version=feature_version)
-        # These doesn't exist before 3.8
+
+        # These don't exist before 3.8
         NamedExpr = Any
         Constant = Any
 except ImportError:
@@ -1173,7 +1177,7 @@ class TypeConverter:
     def visit(self, node: ast3.expr) -> Type: ...
 
     @overload  # noqa
-    def visit(self, node: Optional[AST]) -> Optional[Type]: ...
+    def visit(self, node: Optional[AST]) -> Optional[Type]: ...  # noqa
 
     def visit(self, node: Optional[AST]) -> Optional[Type]:  # noqa
         """Modified visit -- keep track of the stack of nodes"""
@@ -1311,7 +1315,6 @@ class TypeConverter:
             return RawExpressionType(contents, 'builtins.bytes', self.line, column=n.col_offset)
         # Everything else is invalid.
         return self.invalid_type(n)
-
 
     # UnaryOp(op, operand)
     def visit_UnaryOp(self, n: UnaryOp) -> Type:
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -42,7 +42,8 @@ try:
     # Check if we can use the stdlib ast module instead of typed_ast.
     if sys.version_info >= (3, 8):
         import ast as ast3
-        assert 'kind' in ast3.Constant._fields, "This 3.8.0 alpha is too old"
+        assert 'kind' in ast3.Constant._fields, \
+               "This 3.8.0 alpha (%s) is too old; 3.8.0a3 required" % sys.version.split()[0]
         from ast import (
             AST,
             Call,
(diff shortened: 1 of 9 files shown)


mypy/fastparse.py
```diff
@@ -926,6 +979,30 @@ def visit_Call(self, n: Call) -> CallExpr:
                      cast('List[Optional[str]]', [None] * len(args)) + keyword_names)
         return self.set_line(e, n)
 
+    # Constant(object value) -- a constant, in Python 3.8.
+    def visit_Constant(self, n: Constant) -> Any:
+        val = n.value
+        e = None  # type: Any
+        if val is None:
+            e = NameExpr(str(val))
```
ilevkivskyi: Wouldn't it be safer and faster to write `NameExpr('None')`?

mypy/fastparse.py
```diff
@@ -926,6 +979,30 @@ def visit_Call(self, n: Call) -> CallExpr:
...
+        elif isinstance(val, int):
+            e = IntExpr(val)
+        elif isinstance(val, float):
+            e = FloatExpr(val)
+        elif isinstance(val, complex):
+            e = ComplexExpr(val)
+        elif val is Ellipsis:
+            e = EllipsisExpr()
+        else:
+            raise RuntimeError('num not implemented for ' + str(type(val)))
```
ilevkivskyi: What is `'num'`, do you mean`'Constant'`?

gvanrossum:
At some point tomorrow I hope the Python 3.8-dev build will include https://github.com/python/cpython/pull/12295 and then the tests should all pass.

mypy/fastparse.py
```diff
@@ -1205,6 +1283,35 @@ def visit_NameConstant(self, n: NameConstant) -> Type:
         else:
             return UnboundType(str(n.value), line=self.line)
 
+    # Only for 3.8 and newer
+    def visit_Constant(self, n: Constant) -> Type:
+        val = n.value
+        if val is None:
+            # None is a type.
+            return UnboundType(str(val), line=self.line)
```
ilevkivskyi: Maybe also `'None'` here? Or we can add a top-level constant `NONE_NAME: Final = str(None)`.

ilevkivskyi:
Oh, my review crossed with some updates and this is no more WIP.

mypy/fastparse.py
```diff
@@ -568,7 +609,15 @@ def visit_ClassDef(self, n: ast3.ClassDef) -> ClassDef:
                         metaclass=dict(keywords).get('metaclass'),
                         keywords=keywords)
         cdef.decorators = self.translate_expr_list(n.decorator_list)
-        self.set_line(cdef, n)
+        if n.decorator_list and sys.version_info >= (3, 8):
+            # Before 3.8, n.lineno points to the first decorator; in
+            # 3.8, it points to the 'class' statement.  We always make
+            # it point to the first decorator.  (The node structure
```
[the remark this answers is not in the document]
gvanrossum: Yeah, I just followed the guidance of what was done before: for decorated functions, the line number was adjusted to point to `def`; for decorated classes, this wasn't done. So what I did in both cases was simply expedient to keep the tests passing without changing them.

gvanrossum:
After a job restart, the tests passed (meaning python/cpython#12295 made it into Travis's "3.8-dev"). Maybe @jukka can greenlight this?

gvanrossum:
@ilevkivskyi Can you review this one more time? @JukkaL told me offline he's fine with it.

mypy/fastparse.py
```diff
@@ -1094,7 +1177,7 @@ def invalid_type(self, node: AST, note: Optional[str] = None) -> RawExpressionTy
     def visit(self, node: ast3.expr) -> Type: ...
 
     @overload  # noqa
-    def visit(self, node: Optional[AST]) -> Optional[Type]: ...
+    def visit(self, node: Optional[AST]) -> Optional[Type]: ...  # noqa
```
ilevkivskyi: It's sad this `#noqa` is still needed. The issue https://github.com/PyCQA/pyflakes/issues/434 was fixed two weeks ago, so we will be able to update `flake8` and remove these soon.

bluetech:
Does it makes sense to add a `; python_version < "3.8"` [environment marker](https://www.python.org/dev/peps/pep-0508/#environment-markers) to the `typed_ast` requirement in `setup.py`? This way mypy won't pull `typed_ast` on Python >= 3.8.

gvanrossum:
No, it’s still needed for 2.7.

bluetech:
2.7 is included in `< "3.8"`.

The idea is to avoid needlessly installing a package when it's not needed, similar to how the `typing` package is only pulled by mypy for Python < 3.5.3.

ethanhs:
@bluetech, I believe Guido means "It's needed to parse 2.7 on Python 3". Mypy only runs under Python 3, which means to type check Python 2 code we still need typed_ast.

bluetech:
Ha, I misunderstood. That makes sense.

PR #6553, opened 2019-03-15
Fix six.with_metaclass in the new semantic analyzer

### LINKED ISSUE #6350: New semantic analyzer: support six.with_metaclass etc.
#6349 added support for metaclasses, but `six.with_metaclass` and
`@six.add_metaclass` still have not been tested.

msullivan:
Fixes 10 of the 16 errors we have trying to run check-classes.test.
Fixes #6350.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -1371,7 +1371,6 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                         and len(base_expr.args) >= 1
                         and all(kind == ARG_POS for kind in base_expr.arg_kinds)):
                     with_meta_expr = base_expr.args[0]
-                    defn.removed_base_type_exprs.append(defn.base_type_exprs[0])
                     defn.base_type_exprs = base_expr.args[1:]
 
         # Look for @six.add_metaclass(M)
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -855,12 +855,13 @@ class NewSemanticAnalyzer(NodeVisitor[None],
 
         tag = self.track_incomplete_refs()
 
-        bases = defn.base_type_exprs
-
-        self.update_metaclass(defn)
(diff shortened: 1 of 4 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -1370,6 +1371,7 @@ def update_metaclass(self, defn: ClassDef) -> None:
                         and len(base_expr.args) >= 1
                         and all(kind == ARG_POS for kind in base_expr.arg_kinds)):
                     with_meta_expr = base_expr.args[0]
+                    defn.removed_base_type_exprs.append(defn.base_type_exprs[0])
```
ilevkivskyi: Add a short comment clarifying this is needed for multi-iteration logic.

test-data/unit/check-newsemanal.test
```diff
@@ -995,6 +995,32 @@ class C(type):
...
+
+@six.add_metaclass(B)
+class A:
+    pass
+
+class B(type):
+    def f(cls) -> int:
+        return 0
+
+reveal_type(A.f()) # E: Revealed type is 'builtins.int'
```
ilevkivskyi: I would add couple more tests with more tricky deferrals. For example with a base class that also gets deferred.

Also since there are now only 6 tests fail in `check-classes.test`, maybe just explicitly skip those with `# flags: --no-new-semantic-analyzer` and a short motivational comment or a link to issue?

test-data/unit/check-newsemanal.test
```diff
@@ -995,6 +995,32 @@ class C(type):
...
+
+
```
ilevkivskyi: Extra newline.

mypy/newsemanal/semanal.py
msullivan: OK, I think this is not actually needed for this change. I feel like it *ought* to be needed for fine-grained incremental (both for this and for the old analyzer) but I haven't been able to write a test case to prove it.

Issue #6170, opened 2019-01-09
Build and run with mypyc in CI


msullivan:
We've had a lot of issues that this would catch, and mypyc is I think in a good enough state that we can justify exposing it to people.

PR #6554, opened 2019-03-15
Fix mypyc build by removing conditional method definition in fastparse

msullivan:
Remove the conditional method definition added in #6539, which mypyc
doesn't support.


msullivan:
Nobody needs to ever apologize for breaking things under mypyc until I set up CI for it >_>
(#6170)

PR #6557, opened 2019-03-17
Add info about classes and types to 'getting started' docs

Michael0x2a:
While helping somebody use mypy, it came to my attention that our docs never actually quite explain how classes and types interact.

This PR adds a new section named "Types and classes" to the "Getting started" docs that tries to fix this.

This section focuses on introducing two ideas:

1. That you can use any Python class as a type
2. That if a variable has type T, it can also accept any instances of *subclasses* of type T.

The examples also give us an opportunity to sneak in a brief mini-lesson on how to add type hints to classes, and what sorts of things do and do not get inferred.

I also rewrote the next section about stubs/typeshed partly so it would dovetail better with this new section but also partly because I always felt that section was a little too dense and rushed.

Three other minor changes made:

- I added an example about sqlalchemy and sqlalchemy-stubs to one of the other pages.

- I removed a line directing people to the mypy issue tracker if they have questions.

- I dropped the term "library stubs" in favor of just "stub files".


docs/source/getting_started.rst
```diff
@@ -300,34 +300,144 @@ syntax like so:
...
 .. code-block:: python
 
-  x = chr(4)
+    audited = AuditedBankAccount('Charlie', 300)
+    transfer(account_1, audited, 100)   # Type checks!
+
+This behavior is actually a fundamental aspect of the PEP 484 type system: when 
+we annotate some variable with a type ``T``, we are actually telling mypy that
+variable can be assigned an instance of ``T``, or an instance of a *subclass* of ``T``.
+The same rule applies to type hints on parameters or fields.
```
ilevkivskyi: Should we mention that mypy enforces LSP to make sure this is actually safe (maybe in a footnote)?

docs/source/getting_started.rst
```diff
@@ -300,34 +300,144 @@ syntax like so:
...
+:ref:`dynamically typed <dynamic-typing>` and report an error
+whenever you import the library.
+
+We discuss strategies for handling this category of errors in
+:ref:`ignore-missing-imports`, but in short, there are three things you can do:
+
+1. Look for a separate package containing stubs for that library. For example,
+   if you want type hints for the `sqlalchemy <https://www.sqlalchemy.org/>`_ 
+   library, you can install
+   `sqlalchemy-stubs <https://pypi.org/project/sqlalchemy-stubs/>`_.
```
ilevkivskyi: Should we also mention other notable stubs projects? This link appears twice (here and in the section on running mypy), maybe replace one of them with e.g. `django-stubs`?

ilevkivskyi:
@JukkaL are you going to review this PR? If no, I can do a more thorough review.

msullivan:
Hey, @Michael0x2a, if you rebase this, I'll review it.

97littleleaf11:
Any progress on this PR?

JukkaL:
These still look useful improvements, and I can review this if the merge conflicts get resolved.

docs/source/getting_started.rst
```diff
@@ -320,35 +320,140 @@ syntax like so:
...
+
+Mypy can understand third party libraries in the same way as long as those
+libraries come bundled with stub files or have inline annotations that mypy
+can automatically find. (see :ref:`fix-missing-imports` and
+:ref:`installed-packages` for the details). However, if the third party library
+does *not* come bundled with type hints, mypy will not try and guess what
+the types are: it'll assume the entire library is
+:ref:`dynamically typed <dynamic-typing>` and report an error
+whenever you import the library.
+
```
97littleleaf11, line 451: @JukkaL I tried to merge the conflicts. But I am not sure about this parts.

docs/source/getting_started.rst
```diff
@@ -320,35 +320,140 @@ syntax like so:
...
-modules, including the Python standard library. A library stub defines
-a skeleton of the public interface of the library, including classes,
-variables and functions, and their types. Mypy ships with stubs for
-the standard library from the `typeshed
-<https://github.com/python/typeshed>`_ project, which contains library
-stubs for the Python builtins, the standard library, and selected
-third-party packages.
+So far, we've only seen examples of pre-existing types like the ``int``
+or ``float`` builtins, or the ``List`` and ``Dict`` types from the typing
+module. However, these aren't the only types you can use: in fact, you can
```
AlexWaygood, line 331: This is now out of date; the previous section has been updated to use PEP 585 syntax.

```suggestion
or ``float`` builtins, or generic types from ``collections.abc`` and
``typing``, such as ``Iterable``. However, these aren't the only types you can
use: in fact, you can
```

docs/source/getting_started.rst
```diff
@@ -320,35 +320,140 @@ syntax like so:
...
+.. code-block:: python
+
+    class AuditedBankAccount(BankAccount):
+        def __init__(self, account_name: str, initial_balance: int = 0) -> None:
+            super().__init__(account_name, initial_balance)
+
+            # In this case, mypy can't infer the exact type of this field
+            # based on the information available in this constructor, so we
+            # need to add a type annotation.
+            self.audit_log: List[str] = []
```
AlexWaygood, line 383: I'm not massively keen on an example that requires a three-line comment, especially since this is a beginner's tutorial. Can we think of a different example here, that doesn't involve having to instantiate an empty list? (I'm also trying to think of one.)

In any event, we should use PEP 585 syntax here :)

```suggestion
            # In this case, mypy can't infer the exact type of this field
            # based on the information available in this constructor, so we
            # need to add a type annotation.
            self.audit_log: list[str] = []
```

docs/source/getting_started.rst
AlexWaygood: I think this section (from lines 452-487) should be moved to `running_mypy.rst`, or maybe just cut altogether. With the changes made in this PR, it's a very different topic to the rest of the article, and I don't think we need it here since it's covered in depth elsewhere.

AlexWaygood (commented):
This is a really nice PR! Some thoughts below :)

docs/source/getting_started.rst
AlexWaygood: I don't think we need to mention that in this article, since this is a beginner's tutorial.

AlexWaygood:
(BTW @97littleleaf11, I'm only a triager so I can review, but I can't edit the PR :)

docs/source/getting_started.rst
JelleZijlstra: Maybe a transaction count as an int?

AlexWaygood: I like that idea!

docs/source/getting_started.rst
97littleleaf11: I removed the changes by Michael. But I keep the current example since it's quite useful for beginners IMHO.

97littleleaf11:
@AlexWaygood Thanks for your reviews! It's quite an old PR and let's make it merged soon.

I fixed the conflicts and added back some changes made after this PR. Could you take another round of check and merge it if possible?

AlexWaygood:
> Could you take another round of check and merge it if possible?

I don't have merge privileges, but I'd be happy to do another review in the next few days! :)

docs/source/getting_started.rst
97littleleaf11: IMHO it's not really a big deal since typing an empty list is a frequent case (personal feeling).

hauntsaninja: mypy would be able to infer an int

docs/source/getting_started.rst
hauntsaninja: I'd vote for cutting this section entirely. At this point, typeshed is good enough that users don't need to know at all about typeshed and beginner users don't really need to know about stubs to get started

docs/source/getting_started.rst
```diff
@@ -319,37 +319,136 @@ syntax like so:
 
 .. _stubs-intro:
```
hauntsaninja, line 320: Can remove this anchor

docs/source/getting_started.rst
97littleleaf11: I'm going to keep the `list[str]` but remove the comments since the rule has been introduced in previous section

docs/source/getting_started.rst
97littleleaf11: Done!

hauntsaninja (approved):

hauntsaninja:
Thanks everyone for getting this over the line!

Issue #6262, opened 2019-01-28
Using `Literal` typed variables as keys for TypedDict causes "TypedDict key must be a string literal"


chungwu:
Please provide more information to help us understand the issue:

* Are you reporting a bug, or opening a feature request?
Bug

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.

```
from mypy_extensions import TypedDict
from typing_extensions import Literal

class Test(TypedDict):
  a: str
  b: str

TestKey = Literal['a', 'b']

def read(test: Test, key: TestKey):
  return test[key]
```

* What is the actual behavior/output?
mypy.py:13: error: TypedDict key must be a string literal; expected one of ('a', 'b')

* What is the behavior/output you expect?
No type check errors

* What are the versions of mypy and Python you are using?
mypy -- 0.660
python -- 3.6.3

  Do you see the same issue after installing mypy from Git master?
Yes

* What are the mypy flags you are using? (For example --strict-optional)
None

* If mypy crashed with a traceback, please paste
  the full traceback below.

(You can freely edit this text, please remove all the lines
you believe are unnecessary.)

Michael0x2a:
Thanks for the bug report!

What's happening in this case is that mypy doesn't yet understand how to handle indexing TypedDicts using Unions of Literals -- it can currently handle indexing using only Literals. It probably should, though. I'll look into making a fix when I find some time later this week/next week.

chungwu:
Thanks!  This is very useful for typing JSON blobs, and having functions that extract information from the JSON blobs using specific keys.

It would also be nice to extract a keys type from a `TypedDict`, like so :-)

```
class Test(TypedDict):
  a: str
  b: str

TestKey = Test.keys()  # same as Literal['a', 'b']
```

b0g3r:
I have a similar problem, but I try to use a variable as a key:

```python
from mypy_extensions import TypedDict
from typing_extensions import Literal, Final


class MyDict(TypedDict):
    field_name: int

d: MyDict
d = {'field_name': 1}

name_in_var = 'field_name'
d = {name_in_var: 1}

name_in_literal2: Literal['field_name'] = 'field_name'
d = {name_in_literal2: 1}

name_in_final: Final = 'field_name'
d = {name_in_final: 1}
```

And then get errors:
```
Dima Boger, [04.10.19 21:51]
test.py:13: error: Expected TypedDict key to be string literal
[...]

Issue #6123, opened 2019-01-01
TypedDict constructor first argument types are handled inconsistently in Python 2


Michael0x2a:
I noticed an minor inconsistency in the way mypy handles the following program in Python 2:

    from __future__ import unicode_literals
    from mypy_extensions import TypedDict
    
    MyDict = TypedDict('MyDict', {'foo': int})

Mypy type-checks this problem without errors. However, if you actually try running this program at runtime, you get the following exception:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\site-packages\mypy_extensions.py", line 40, in _typeddict_new
    '__total__': total})
  File "C:\Python27\lib\site-packages\mypy_extensions.py", line 52, in __new__
    tp_dict = super(_TypedDictMeta, cls).__new__(cls, name, (dict,), ns)
TypeError: type() argument 1 must be string, not unicode
```

You have to do `MyDict = TypedDict(b'MyDict', {'foo': int})` to get this program to both typecheck and run.

I'm not sure whether the correct thing to do is modify mypy to report an error or to relax the runtime implementation. 

On one hand, `type(name, bases, dict)` legitimately won't accept unicode names in Python 2 so perhaps we should add a check to mypy. On the other, we don't get the same kind of runtime error when creating `NamedTuples`, so perhaps it's better to try and convert the unicode name to str inside `TypedDict` for consistency.

Michael0x2a:
Hmm, it also seems like the keys are also handled inconsistently at well.

Mypy and the Python runtime both let you have unicode keys. They also both accept `some_typed_dict["blah"]` and `some_typed_dict[u"blah"]` in Python 2, regardless of how the typed dict is defined.

However, mypy will report an error when doing `some_typed_dict.get(u"blah")` since the `get(...)` method is defined to only accept strings. This operation succeeds at runtime though.

gvanrossum:
Why isn’t this a typeshed issue?

JukkaL:
I think that unicode literals should be accepted everywhere, since using `b'foo'` won't work in Python 3. It would be good if TypedDict was compatible with `unicode_literals`.

JukkaL:
Closing due to Python 2 feature freeze (#12237).

PR #6558, opened 2019-03-17
Add union math for intelligent indexing

Michael0x2a:
This pull request hacks on supports for union math when using Unions of Literals to index into tuples, NamedTuples, and TypedDicts.

It also fixes a bug I apparently introduced. Currently, mypy correctly reports an error with this code:

    class Test(TypedDict):
        foo: int

    t: Test

    # Error: int can't be assigned a str value
    t.setdefault("foo", "unrelated value")

...but does not report an error with:

    key: Literal["foo"]
    t.setdefault(key, "unrelated value")

This diff should make mypy report an error in both cases.

Resolves https://github.com/python/mypy/issues/6262.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2,6 +2,7 @@
 
 from collections import OrderedDict
 from contextlib import contextmanager
+import itertools
 from typing import (
     cast, Dict, Set, List, Tuple, Callable, Union, Optional, Sequence, Iterator
 )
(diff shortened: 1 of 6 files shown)


ilevkivskyi:
Now that https://github.com/python/mypy/pull/6560 is merged, could you please add tests to check interactions between the two features? Then I will take a look at this.

Michael0x2a:
@ilevkivskyi -- ok, I added in a test testing a few basic interactions with TypedDicts. It didn't seem like this PR and https://github.com/python/mypy/pull/6560 really interfered with one another (as we hoped, I didn't have to make any code changes), so I stuck to just adding some basic sanity checks. LMK if you want something more in-depth.

mypy/checkexpr.py
```diff
@@ -2607,40 +2617,59 @@ def nonliteral_tuple_index_helper(self, left_type: TupleType, index: Expression)
...
+            key_names = []
+            for key_type in key_types:
+                if isinstance(key_type, Instance) and key_type.last_known_value is not None:
+                    key_type = key_type.last_known_value
+
+                if isinstance(key_type, LiteralType) and isinstance(key_type.value, str):
+                    key_names.append(key_type.value)
+                else:
+                    self.msg.typeddict_key_must_be_string_literal(td_type, index)
+                    return AnyType(TypeOfAny.from_error)
```
ilevkivskyi: Would it be practical to refactor this, so that this uses `try_getting_str_literals()`? It looks like there is some code duplication. (If yes, the function should be probably moved to `nodes.py` and imported both here and in the plugin.)

ilevkivskyi:
@Michael0x2a The release branch will be cut tomorrow morning (SF time). Is this PR ready?

Michael0x2a:
@ilevkivskyi -- sorry, it took me a bit to work out some quirks with my setup so I could start testing this against our internal codebases.

But I think (?) I've confirmed this PR introduces no new errors, so I went ahead and merged it in.

mypy/checkexpr.py
Michael0x2a: Unfortunately, no -- this segment of code permits unicode keys but the `try_getting_str_literals` section doesn't.

IMO the correct thing to do is (a) modify `try_getting_str_literals` so it also allows unicode keys and (b) convert this entire function into a plugin.

But I think making this change fully correct is partly blocked on https://github.com/python/mypy/issues/6123, so I decided to defer the refactor for now.

Issue #6117, opened 2018-12-30
Several plugin hooks don't work on unions


ilevkivskyi:
Some plugin hooks (in particular for `TypedDict`s) are not called if the underlying "base" type is a union. For example:
```python
from typing import Union
from mypy_extensions import TypedDict

class TDA(TypedDict):
    a: int

class TDB(TypedDict):
    a: int

td: Union[TDA, TDB]

reveal_type(td['a'])  # Revealed type is 'builtins.object*'
reveal_type(td.pop('a'))  # Revealed type is 'builtins.object'
                          # Argument 1 to "pop" of "TypedDict" has incompatible type "str"; expected "NoReturn"
```
This affects both "builtin" plugins and user defined plugins.

ilevkivskyi:
It actually looks like the only plugin hook that is broken is `get_method_signature_hook()`, other hooks seem to work, but this one is still important to fix, I will make a PR now.

ilevkivskyi:
It is a bit more tricky than I thought (also `get_method_hook()` may be also broken for unions). I will continue next weekend.

Issue #5930, opened 2018-11-21
Unions of TypedDicts don't work as expected


JukkaL:
Operations on unions of typed dicts seem to use the fallback types, resulting in unexpected types:

```py
from mypy_extensions import TypedDict
from typing import Union

class A(TypedDict):
    x: int
    y: str

class B(TypedDict):
    x: str

d: Union[A, B]
reveal_type(d['x'])  # builtins.object (expected: Union[int, str])
reveal_type(d.get('x'))  # builtins.object (expected: Union[int, str, None])
```

grantwwu:
Here is an additional example, noting that you don't need conflicting types to hit this issue:

```
from mypy_extensions import TypedDict
from typing import Union


class A(TypedDict):
    x: int
    y: str


class B(TypedDict):
    x: int


d: Union[A, B]
reveal_type(d["x"]) # Revealed type is 'builtins.object*' (expected: int)
```

PR #6560, opened 2019-03-18
Support method plugin hooks on unions

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6117
Fixes https://github.com/python/mypy/issues/5930

Currently both our plugin method hooks don't work with unions. This PR fixes this with three things:
* Moves a bit of logic from `visit_call_expr_inner()` (which is a long method already) to `check_call_expr_with_callee_type()` (which is a short method).
* Special-cases unions in `check_call_expr_with_callee_type()` (normal method calls) and `check_method_call_by_name()` (dunder/operator method calls).
* Adds some clarifying comments and a docstring.

The week point is interaction with binder, but IMO this is the best we can have for now. I left a comment mentioning that check for overlap should be consistent in two functions.

In general, I don't like special-casing, but I spent several days thinking of other solutions, and it looks like special-casing unions in couple more places is the only reasonable way to fix unions-vs-plugins interactions.

This PR may interfere with https://github.com/python/mypy/pull/6558 that fixes an "opposite" problem, hopefully they will work together unmodified, so that accessing union of literals on union of typed dicts works. Whatever PR lands second, should add a test for this.


mypy/checkexpr.py
```diff
@@ -645,15 +649,48 @@ def check_call_expr_with_callee_type(self,
...
                                          e: CallExpr,
                                          callable_name: Optional[str],
-                                         object_type: Optional[Type]) -> Type:
+                                         object_type: Optional[Type],
+                                         member: Optional[str] = None) -> Type:
         """Type check call expression.
 
         The given callee type overrides the type of the callee
         expression.
+
```
JukkaL: The above docstring part was a bit confusing. It looks like we aren't actually overriding the callee type but to reusing a previously calculated callee type. This looks like an optimization. Maybe update the docstring to mention that? Otherwise it's not clear whether recalculating the callee type is correct.

mypy/checkexpr.py
```diff
@@ -645,15 +649,48 @@ def check_call_expr_with_callee_type(self,
...
+        The 'callable_name' and 'object_type' are used to call plugin hooks.
```
JukkaL: Also `'member'` is used for plugin hooks only? Maybe be more explicit and say that they are *only* used for this purpose.

mypy/checkexpr.py
```diff
@@ -645,15 +649,48 @@ def check_call_expr_with_callee_type(self,
...
+            res = []  # type: List[Type]
+            for typ in object_type.items:
+                # Member access errors are already reported when visiting the member expression.
+                self.msg.disable_errors()
+                item = analyze_member_access(member, typ, e, False, False, False,
+                                             self.msg, original_type=object_type, chk=self.chk,
+                                             in_literal_context=self.is_literal_context())
+                self.msg.enable_errors()
+                item = self.narrow_type_from_binder(e.callee, item)
+                # TODO: This check is not one-to-one with non-special-cased behavior.
```
JukkaL: Do you know what issues this may cause? Currently it's not clear if this is a serious problem or not.

mypy/checkexpr.py
```diff
@@ -645,15 +649,48 @@ def check_call_expr_with_callee_type(self,
...
+        if callable_name is None and member is not None:
+            assert object_type is not None
+            callable_name = self.method_fullname(object_type, member)
+        if callable_name:
+            # Try to refine the call signature using plugin hooks before checking the call.
+            callee_type = self.transform_callee_type(
+                callable_name, callee_type, e.args, e.arg_kinds, e, e.arg_names, object_type)
+        # Unions are special-cased to allow plugins to act on each item in the union.
+        elif member is not None and isinstance(object_type, UnionType):
+            res = []  # type: List[Type]
```
JukkaL: Can you move the elif block to a separate method? Now the largest part of this function handles a rare special case, obscuring the main flows.

mypy/checkexpr.py
```diff
@@ -2021,6 +2058,22 @@ def check_method_call_by_name(self,
         """
         local_errors = local_errors or self.msg
         original_type = original_type or base_type
+        # Unions are special-cased to allow plugins to act on each element of the union.
+        if isinstance(base_type, UnionType):
+            res = []  # type: List[Type]
```
JukkaL: Similar to above, it would be better to move this block to a separate function.

mypy/checkexpr.py
ilevkivskyi: Actually I found a cleaner solution, I will push an update soon.

ilevkivskyi:
@Michael0x2a @JukkaL I addressed the comments from last round, do you have any more comments here?

test-data/unit/check-custom-plugin.test
```diff
@@ -585,6 +585,67 @@ reveal_type(instance(2))  # N: Revealed type is 'builtins.float*'
...
+class Bar:
+    def meth(self, x: int) -> str: ...
+class Other:
+    meth: int
+
+x: Union[Foo, Bar, Other]
+if isinstance(x.meth, int):
+    reveal_type(x.meth)  # N: Revealed type is 'builtins.int'
+else:
+    reveal_type(x.meth(int()))  # N: Revealed type is 'builtins.int'
```
JukkaL: Since the plugin switches types between `int`/`str` and the existing types are also `int`/`str`, it's a bit hard to see what is going on. What about using more than two separate types?

test-data/unit/check-custom-plugin.test
```diff
@@ -585,6 +585,67 @@ reveal_type(instance(2))  # N: Revealed type is 'builtins.float*'
...
 plugins=<ROOT>/test-data/unit/plugins/callable_instance.py
 
+[case testGetMethodHooksOnUnions]
+# flags: --config-file tmp/mypy.ini --no-strict-optional
+from typing import Union
+
+class Foo:
+    def meth(self, x: str) -> str: ...
+class Bar:
+    def meth(self, x: int) -> str: ...
```
JukkaL: If each `meth` variant would have a distinct return type, the test would perhaps catch more issues.

test-data/unit/check-custom-plugin.test
```diff
@@ -585,6 +585,67 @@ reveal_type(instance(2))  # N: Revealed type is 'builtins.float*'
...
+class Other:
+    meth: int
+
+x: Union[Foo, Bar, Other]
+if isinstance(x.meth, int):
+    reveal_type(x.meth)  # N: Revealed type is 'builtins.int'
+else:
+    reveal_type(x.meth(int()))  # N: Revealed type is 'builtins.int'
```
JukkaL: This is similar to above -- maybe use more than two separate types.

test-data/unit/check-custom-plugin.test
ilevkivskyi: OK, I now have different return types for two `meth` methods, and correspondingly different type conversions for the two hooks: `str` -> `int`, and `float -> int`, this way the call would fail if a wrong hook would act, and hopefully this is now easier to read.

PR #6562, opened 2019-03-18
Add --no-implicit-reexport flag

jhance:
Adds a --no-implicit-reexport flag, intended for giving modules the same behavior as stubs for re-exporting their imports.

The goal is to make it possible to prevent bad transitive imports from causing dependency issues. And to make it easier to move things to a different module.

### DIFF
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -294,6 +294,18 @@ Miscellaneous strictness flags
     Allows variables to be redefined with an arbitrary type, as long as the redefinition
     is in the same block and nesting level as the original definition.
 
+``implicit-reexport`` (bool, default True)
+    By default, imported values to a module are treated as exported and mypy allows
+    other modules to import them. When false, mypy will not re-export unless
+    the item is imported using from-as. Note that mypy treats stub files as if this
+    is always disabled. For example:
+
+    .. code-block:: python
+    # This won't re-export the value
+    from foo import bar
+    # This will re-export it as bar and allow other modules to import it
+    from foo import bar as bar
+
 ``strict_equality``  (bool, default False)
    Prohibit equality checks, identity checks, and container checks between
    non-overlapping types.
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -302,10 +302,10 @@ Miscellaneous strictness flags
 
     .. code-block:: python
 
-    # This won't re-export the value
-    from foo import bar
-    # This will re-export it as bar and allow other modules to import it
-    from foo import bar as bar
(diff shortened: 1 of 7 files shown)


mypy/semanal.py
```diff
@@ -1545,7 +1545,8 @@ def visit_import_from(self, imp: ImportFrom) -> None:
                             imported_id, existing_symbol, node, imp):
                         continue
                 # 'from m import x as x' exports x in a stub file.
-                module_public = not self.is_stub_file or as_id is not None
+                module_public = not self.is_stub_file and not self.options.no_implicit_reexport \
+                        or as_id is not None
```
gvanrossum: Style nit: use `(...)` instead of `\`.

jhance:
There may be poor behavior if this is enabled without `--no-ignore-missing-imports` as it will just turn all the stuff to `Any`

msullivan:
I've thought about it some and the interactions with ignore_missing_imports seems fine? Ignore missing imports ignores missing *modules*, not missing attributes.

I'm going to try to get mypy to check clean with this option and report back on that.

msullivan:
https://github.com/python/mypy/pull/6570 fixes mypy to check clean with this.

The error output was intimidating at first (because of all the knock-on effects), but when I just looked at the import failures it was very simple to tackle.

(It may be a good idea to treat imports of non-public names differently: produce an error on the import, but still resolve the name to the private symbol, to reduce the error volume and to provide better type checking if the import error gets ignored. That's neither here nor there for this bug.)

test-data/unit/check-flags.test
```diff
@@ -1137,3 +1137,34 @@ strict_equality = True
 [[mypy-b]
 strict_equality = False
 [builtins fixtures/bool.pyi]
+
+[case testNoImplicitReexport]
+# flags: --no-implicit-reexport
+from other_module_2 import a
+
```
gvanrossum: I'd remove all the blank lines from the test cases except the blank line *between* test cases (so no blank lines between files within a test case).

docs/source/config_file.rst
```diff
@@ -294,6 +294,19 @@ Miscellaneous strictness flags
...
+    other modules to import them. When false, mypy will not re-export unless
+    the item is imported using from-as. Note that mypy treats stub files as if this
+    is always disabled. For example:
+
+    .. code-block:: python
+
+    # This won't re-export the value
+    from foo import bar
+    # This will re-export it as bar and allow other modules to import it
+    from foo import bar as bar
```
ilevkivskyi: Indentation is still wrong here.

Issue #6300, opened 2019-02-01
New semantic analyzer: only create each SymbolNode once


JukkaL:
Currently we can create a symbol node multiple times (once per iteration), which is risky since we don't want to risk having different references to a particular name pointing to different nodes. For example, class-based TypedDict definitions may create multiple `TypeInfo` objects per definition.

I think that for most things we can treat the definition as complete once we've successfully created a symbol table node, and further iterations would skip the definition, as there's nothing more to do. 

Normal class definitions should not have this problem, but we should verify that these work as expected:

* `Var` nodes
* TypedDict (as discussed above)
* Named tuples
* Type variables
* Type aliases
* NewTypes
* Enums

ilevkivskyi:
@JukkaL the problem with `AnyStr` I told you about was caused by this. I fixed this for type vars, but we still need to check other node kinds in this list.

ilevkivskyi:
Just a note to self that we need to fix `TypeVar` redefinition by preserving a link to defining assignment (I will make a PR).

JukkaL:
Lowering priority since I believe that we've addresses most issues (but not necessarily all).

JukkaL:
We currently allow submodules to be redefined as various different things (see #6828) which breaks these assumptions. It looks like we need another approach for dealing with storing submodules. The current approach is problematic in other ways as well.

ilevkivskyi:
I think we have fixed all the cases I am aware. We still do some patching/updating of the _existing_ nodes (replace placeholder types in base classes, alias targets, and variable upper bounds), but we don't re-add symbol table nodes. If we will discover more, we can open dedicated issues.

PR #6563, opened 2019-03-18
New semantic analyzer: Prohibit duplicate type variable definitions and fix stale upper bounds and values

ilevkivskyi:
This is a follow up for #6527.

Fixes https://github.com/python/mypy/issues/7037
I think this also essentially concludes https://github.com/python/mypy/issues/6300

This PR prohibits duplicate type variable definitions, previously the second one was silently winning. Also now the first one wins. Note that it didn't require additional attribute on `TypeVarExpr` as I initially though, it looks like we can simple use `s.rvalue.analyzed` for the purpose of linking assignments to variables.

This PR also naturally fixes an issue where stale upper bound or value appears in a class that uses the type variable.


mypy/newsemanal/semanal.py
```diff
@@ -2675,10 +2675,15 @@ def process_typevar_declaration(self, s: AssignmentStmt) -> bool:
...
             return False
+
+        assert isinstance(s.rvalue, CallExpr)
         name = lvalue.name
         names = self.current_symbol_table()
         existing = names.get(name)
-        if existing and not isinstance(existing.node, (TypeVarExpr, PlaceholderNode)):
+        if existing and not (isinstance(existing.node, PlaceholderNode) or
+                             # Also give error for another type variable with the same name.
+                             isinstance(existing.node, TypeVarExpr) and
```
JukkaL: Style nit: Use parentheses to make precedence explicit when mixing and/or.

mypy/newsemanal/semanal.py
```diff
@@ -2721,8 +2726,15 @@ def process_typevar_declaration(self, s: AssignmentStmt) -> bool:
...
             # Existing definition from previous semanal iteration, use it.
-            # TODO: This may be confused with a duplicate TypeVar definition.
-            #       Fix this and add corresponding tests.
+            # Note that unlike in most other cases, we patch an existing symbol, instead of
+            # adding the type variable symbol only when the types it depends on are ready.
+            # This is necessary to avoid a "deadlock" in this typical situation:
+            #     T = TypeVar('T', bound=C[Any])
+            #     class C(Generic[T]):
+            #         ...
+            # We can use allow_placeholder=True for upper bound and values to avoid this,
```
JukkaL: This doesn't make it clear whether we are actually doing this. Instead, write something like "We could use ..., but we don't do this, because ...".

What happens if the type variable definition and the class definition are in different parts of an import cycle? Do we ensure that once the type variable has been completed, the new bound will be used in the class definition?

mypy/type_visitor.py
```diff
@@ -282,7 +282,7 @@ def visit_deleted_type(self, t: DeletedType) -> T:
         return self.strategy([])
 
     def visit_type_var(self, t: TypeVarType) -> T:
-        return self.strategy([])
+        return self.query_types([t.upper_bound] + t.values)
```
ilevkivskyi: This is strictly speaking unrelated, but it made one new test fail (because `has_praceholder()` was wrong).

mypy/newsemanal/semanal.py
```diff
@@ -2719,20 +2731,18 @@ def process_typevar_declaration(self, s: AssignmentStmt) -> bool:
...
             type_var = TypeVarExpr(name, self.qualified_name(name),
                                    values, upper_bound, variance)
             type_var.line = call.line
             call.analyzed = type_var
-            self.add_symbol(name, type_var, s)
+        else:
+            assert isinstance(call.analyzed, TypeVarExpr)
+            call.analyzed.upper_bound = upper_bound
+            call.analyzed.values = values
+            self.progress = True
```
JukkaL: Is this always progress, or should we check for changes in the bounds?

mypy/newsemanal/semanal.py
```diff
@@ -2807,18 +2817,13 @@ def process_typevar_parameters(self, args: List[Expression],
...
-                        # incrementally populate `TypeVarExpr` if some part is missing during main
-                        # pass iterations.
-                        # NOTE: It is safe to not call self.defer() here, because the only way
-                        # we can get None from self.anal_type() is if self.found_incomplete_refs()
-                        # returned True. In turn, the only way it can happen is if someone called
-                        # self.record_incomplete_ref(), and the latter unconditionally calls
-                        # self.defer().
-                        analyzed = AnyType(TypeOfAny.special_form)
+                        # Type variables are special: we need to place them in the symbol table
+                        # soon, even if upper bound is not ready yet.
```
JukkaL: Can you explain why they need to be added to the symbol table soon?

mypy/newsemanal/semanal.py
```diff
@@ -2872,18 +2877,19 @@ def basic_new_typeinfo(self, name: str, basetype_or_fallback: Instance) -> TypeI
...
-                    result.append(analyzed)
-                else:
-                    # It is fine to simply use temporary Anys because we don't need values
-                    # for anything before main pass of semantic analysis is finished.
-                    result.append(AnyType(TypeOfAny.special_form))
+                analyzed = self.anal_type(expr_to_unanalyzed_type(node),
+                                          allow_placeholder=True)
+                if analyzed is None:
+                    # Type variables are special: we need to place them in the symbol table
+                    # soon, even if some value is not ready yet.
```
JukkaL: Similar to above: can you explain why?

JukkaL:
Oh and feel free to merge when you've addressed my comments.

ilevkivskyi:
Travis hook seems to be stuck, all tests have passed, so I am merging this now.

Issue #6292, opened 2019-02-01
New semantic analyzer: casts


JukkaL:
In particular, support `cast(t, e)` when `t` contains a forward reference.

msullivan:
This got closed, messed up the syntax

PR #6566, opened 2019-03-18
New semantic analyzer: add tests for casts and type application

### LINKED ISSUE #6291: New semantic analyzer: type application
Support type applications such as in `x = C[int]()`.

msullivan:
It seems like they already work fine?

Closes #6291, #6292.

### DIFF
--- a/test-data/unit/check-newsemanal.test
+++ b/test-data/unit/check-newsemanal.test
@@ -1859,7 +1859,7 @@ x: A
 reveal_type(x)  # E: Revealed type is '__main__.G[Tuple[builtins.int, fallback=__main__.C]]'
 [builtins fixtures/list.pyi]
 
-[case testNewAnalyzerCastForward]
+[case testNewAnalyzerCastForward1]
 from typing import cast
 
 x = cast('C', None)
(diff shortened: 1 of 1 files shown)


test-data/unit/check-newsemanal.test
```diff
@@ -1856,3 +1858,47 @@ reveal_type(y[0])  # E: Revealed type is 'builtins.int'
...
+
+x = cast('C', None)
+class A:
+    def foo(self) -> None:
+        self.x = cast('C', None)
+
+reveal_type(x)  # E: Revealed type is '__main__.C'
+reveal_type(A().x)  # E: Revealed type is '__main__.C'
+
+class C(A): ...
```
ilevkivskyi: Maybe add couple of tests that use special forms (type aliases, named tuples, etc)?

test-data/unit/check-newsemanal.test
```diff
@@ -1856,3 +1858,47 @@ reveal_type(y[0])  # E: Revealed type is 'builtins.int'
...
+
+x = C['A']()
+reveal_type(x)  # E: Revealed type is '__main__.C[__main__.A*]'
+
+class A: ...
+
+[case testNewAnalyzerApplicationForward3]
+from typing import Generic, TypeVar
+
+x = C[A]()
```
ilevkivskyi: Maybe add a test where the upper bound (also forward one) is violated?

test-data/unit/check-newsemanal.test
```diff
@@ -1476,6 +1476,8 @@ strict_optional = True
 [[mypy-b]
 strict_optional = False
 
+-- ]]
```
ilevkivskyi: Btw, why is this needed? We a bunch of unbalanced `[` in other test files.

test-data/unit/check-newsemanal.test
```diff
@@ -1856,3 +1858,78 @@ reveal_type(y[0])  # E: Revealed type is 'builtins.int'
...
+reveal_type(x)  # E: Revealed type is 'builtins.int'
+
+C = int
+
+[case testNewAnalyzerCastForward2]
+from typing import cast, NamedTuple
+
+x = cast('C', None)
+
+reveal_type(x)  # E: Revealed type is '__main__.C'
```
ilevkivskyi: This should be a tuple, the test failure is likely because of this.

test-data/unit/check-newsemanal.test
msullivan: Yeah, and I usually fix them when I run into them :P 
They mess up editor indentation, which I find super annoying

PR #6567, opened 2019-03-19
New semantic analyzer: support __all__

### LINKED ISSUE #6319: New semantic analyzer: __all__
It's not clear whether `__all__` works with the new semantic analyzer.

msullivan:
A bunch of relevant test cases are in check-classes, which we remove
from the blacklist and manually opt out of remaining failures.

Closes #6319.

### DIFF
--- a/test-data/unit/check-modules.test
+++ b/test-data/unit/check-modules.test
@@ -324,9 +324,9 @@ main:2: error: Name 'y' is not defined
 
 [case testUnknownModuleRedefinition]
 # Error messages differ with the new analyzer
-# flags: --no-new-semantic-analyzer
+# flags: --new-semantic-analyzer
 import xab  # E: Cannot find module named 'xab'  # N: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
-def xab(): pass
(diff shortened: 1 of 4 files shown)


JukkaL:
It looks like a test needs to be updated.

test-data/unit/check-modules.test
```diff
@@ -323,11 +323,10 @@ main:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missin
 main:2: error: Name 'y' is not defined
 
 [case testUnknownModuleRedefinition]
-import xab
+# Error messages differ with the new analyzer
+# flags: --no-new-semantic-analyzer
```
JukkaL: We generally have preferred to use `--new-semantic-analyzer` if test case passes with both semantic analyzers, but error messages differ.

test-data/unit/check-modules.test
```diff
@@ -520,6 +519,8 @@ x = 1
 x = 1
 
 [case testAssignToFuncDefViaImport]
+# Error messages differ with the new analyzer
+# flags: --no-new-semantic-analyzer
```
JukkaL: Similar to above.

Issue #6575, opened 2019-03-20
NamedTuples can't match Protocols with Callable fields


FuegoFro:
This looks like a bug that is a combination of how `NamedTuple` fields are created and how node types are determined.

If a `Protocol` expects a field with type `Callable`, passing a `NamedTuple` that has a `Callable` field won't work. A minimal reproduction is below. This behavior happens for me with `Python 3.7.2` and `mypy 0.680+dev.4e0a1583aeb00b248e187054980771f1897a1d31` using only the `--strict` flag.

```py3
from typing import Callable, NamedTuple

from typing_extensions import Protocol


class A(NamedTuple):
    my_func: Callable[[], str]


class WithFunc(Protocol):
    # This is a property so it matches the read-only `my_func` in the NamedTuple
    @property
    def my_func(self) -> Callable[[], str]:
        ...


def call_func(with_func: WithFunc) -> None:
    print(with_func.my_func())


a = A(lambda: "hi")
call_func(a)  # <-- PROBLEM: This says that `A#my_func` has type `str`

MYPY = False
if MYPY:
    reveal_type(a.my_func)  # This correctly identifies `A#my_func` as a function
```

It looks like this is due to an interaction between the [`find_node_type` code](https://github.com/python/mypy/blob/master/mypy/subtypes.py#L625-L627) that handles properties and the [`build_namedtuple_typeinfo` code](https://github.com/python/mypy/blob/master/mypy/semanal_namedtuple.py#L330-L331) which implicitly marks all `NamedTuple` fields as properties. This causes the `find_node_type` code to think that the field has the type of the `Callable`'s return type, rather than the whole `Callable`.

I'm happy to submit a PR to try to fix this, but it's unclear to me if the right solution here is to no longer assume that all `NamedTuple` are properties or to add an explicit check for `NamedTuple`s to the property-resolving code in  `find_node_type`. Or something else entirely! If you have suggestions for the best way to resolve this I can submit a PR 🙂

PR #6576, opened 2019-03-20
Don't resolve Callable NamedTuple fields to their return type

FuegoFro:
`NamedTuple` fields are always marked an properties. In certain situations, this causes their type to be determined incorrectly when they hold `Callable`s, resolving to the `Callable`'s return type rather than the `Callable` itself. This prevents that and fixes https://github.com/python/mypy/issues/6575.

### DIFF
--- a/mypy/subtypes.py
+++ b/mypy/subtypes.py
@@ -619,10 +619,13 @@ def find_node_type(node: Union[Var, FuncBase], itype: Instance, subtype: Type) -
     if typ is None:
         return AnyType(TypeOfAny.from_error)
     # We don't need to bind 'self' for static methods, since there is no 'self'.
-    if isinstance(node, FuncBase) or isinstance(typ, FunctionLike) and not node.is_staticmethod:
+    if (isinstance(node, FuncBase)
+            or (isinstance(typ, FunctionLike)
+                and node.is_initialized_in_class
+                and not node.is_staticmethod)):
         assert isinstance(typ, FunctionLike)
         signature = bind_self(typ, subtype)
-        if node.is_property and not node.info.is_named_tuple:
+        if node.is_property:
             assert isinstance(signature, CallableType)
             typ = signature.ret_type
         else:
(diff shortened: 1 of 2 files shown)


msullivan:
I am going to look at the bug you filed and think a bit about the whether this is the best approach (or whether, as you suggested, we should change something on the namedtuple side setting properties), but separate from that: could you please add a test.

mypy/subtypes.py
```diff
@@ -622,7 +622,7 @@ def find_node_type(node: Union[Var, FuncBase], itype: Instance, subtype: Type) -
     if isinstance(node, FuncBase) or isinstance(typ, FunctionLike) and not node.is_staticmethod:
         assert isinstance(typ, FunctionLike)
         signature = bind_self(typ, subtype)
-        if node.is_property:
+        if node.is_property and not node.info.is_named_tuple:
```
msullivan: Hm, it looks like this will still do `bind_self` on the type, which I think might cause trouble, since it will modify the type of the field in an unexpected way. Make sure to test with multiple arguments.

FuegoFro:
Thanks for the response! Yeah, I'll take a stab at adding some tests for various scenarios, including where the `Callable` takes multiple arguments 🙂

mypy/subtypes.py
msullivan: Looking at https://github.com/python/mypy/blob/master/mypy/checkmember.py#L495, it seems that only  `is_initialized_in_class` vars get the property/self binding treatment, so probably the right approach is to match that.

FuegoFro:
Thanks for the pointers here! I changed the 'self' binding check to match the code/condition you linked to 🙂 I also added some integration tests. I haven't really worked with this codebase much before, so I wasn't sure if there's a way I should try to make these work in the [`testsubtypes.py`](https://github.com/python/mypy/blob/master/mypy/test/testsubtypes.py) file. Let me know if there's a better way to add tests, or any other types of tests you'd like to see!

I did notice that a `TestPEP561.test_typedpkg_editable` test was failing locally, but it failed for me even on master and it seems unrelated to this change. However, if that does seem relevant and you have some pointers I can dig into it more 🙂

FuegoFro:
Thank you for you help and quick responses/turnaround! Don't think I was in a `venv`, but it sounds like this wasn't entirely unexpected, so I'll hold off on looking into my local setup until/unless I have more PRs to send 😉

PR #6582, opened 2019-03-22
New semantic analyzer: store import deps generated during analysis

JukkaL:
Previously some module dependencies genererated from `from m import *`
went missing, and this could result in crashes when deserialization,
since the target module wasn't serialized yet.

This fixes one incremental mode test case.

### DIFF
--- a/mypy/newsemanal/semanal_main.py
+++ b/mypy/newsemanal/semanal_main.py
@@ -281,7 +281,9 @@ def semantic_analyze_target(target: str,
                 infer_decorator_signature_if_simple(node, analyzer)
     for dep in analyzer.imports:
         state.dependencies.append(dep)
-        state.priorities[dep] = mypy.build.PRI_LOW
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal_main.py
```diff
@@ -278,6 +279,9 @@ def semantic_analyze_target(target: str,
             analyzer.refresh_partial(refresh_node, patches, final_iteration)
             if isinstance(node, Decorator):
                 infer_decorator_signature_if_simple(node, analyzer)
+    for dep in analyzer.imports:
+        state.dependencies.append(dep)
+        state.priorities[dep] = mypy.build.PRI_LOW
```
msullivan: There is no checking against what is already a dependency, so can this result in things ending up in the dependency list multiple times or having the priority clobbered?
(Though I suppose we hope that the priority does not matter with the new semantic analyzer, right?)

PR #6584, opened 2019-03-22
implement @no_type_check for class defs

### LINKED ISSUE #607: Support no_type_check and no_type_check_decorator
These are defined in the PEP 484 draft.

lunixbochs:
Fixes #607 (which is incorrectly titled).

This adds `TypeInfo.is_no_type_check`, and enables `TypeInfo.fallback_to_any` for classes decorated with `@no_type_check`, which is checked in a couple of places to abort the type checker early and stub method types as `Any`.

I'm not sure confident about my use of `decorator.name in ('no_type_check', 'typing.no_type_check'` in `semanal.py`. I see `RefName.fullname = 'typing.no_type_check'` was used elsewhere but it wasn't a RefName in the AST during `semanal` and I didn't look too far into that.

This is my current test:
```
from typing import no_type_check, Any

@no_type_check
class TestClass:
    bar: {'x': 1} = 1

    def meth1(): pass
    def meth2(a: {'x': 1}) -> {'x': 1}:
        def garbage_inner(a: {'x': 1}):
            pass

    def test(a: int): pass

TestClass().test(1)
TestClass().doesntexist()

class Test(TestClass):
    def test2(self, a: int): pass

t = Test()
t.meth1()
t.doesntexist2()
t.test(1)
t.test2(1)

x: int = TestClass().bar
```

With `@no_type_check`, mypy throws a `Missing return statement` for the nested `garbage_inner` definition, but suppresses the rest of the type errors for this class and subclass.

Without `@no_type_check`, we get this nice pile of errors:
```
repro.py:4: error: Invalid type comment or annotation
repro.py:6: error: Method must have at least one argument
repro.py:7: error: Invalid type comment or annotation
repro.py:8: error: Invalid type comment or annotation
repro.py:11: error: Self argument missing for a non-static method (or an invalid type for self)
repro.py:13: error: Too many arguments for "test" of "TestClass"
repro.py:14: error: "TestClass" has no attribute "doesntexist"
repro.py:21: error: "Test" has no attribute "doesntexist2"
repro.py:22: error: Too many arguments for "test" of "TestClass"
```

TODO (I'd like advice from core contributors on these):

- [ ] I think I should write in-tree tests for this, but I could use guidance on that.
- [ ] confirm my `decorator.name` check is correct.
- [ ] is it possible to suppress the `no return` error?
- [ ] am I missing any additional obvious class type behavior in my test that should be suppressed/stubbed by the decorator?

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -162,6 +162,9 @@ def analyze_instance_member_access(name: str,
     if override_info:
         info = override_info
 
+    if info.is_no_type_check:
+        return AnyType(TypeOfAny.special_form)
+
     if (state.find_occurrences and
             info.name() == state.find_occurrences[0] and
             name == state.find_occurrences[1]):
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -2219,6 +2219,7 @@ class TypeInfo(SymbolNode):
     metaclass_type = None  # type: Optional[mypy.types.Instance]
 
     names = None  # type: SymbolTable      # Names defined directly in this type
+    is_no_type_check = False               # Was the class annotated with @no_type_check?
     is_abstract = False                    # Does the class have any abstract attributes?
     is_protocol = False                    # Is this a protocol class?
     runtime_protocol = False               # Does this protocol support isinstance checks?
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -487,6 +487,9 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
         """Check basic signature validity and tweak annotation of self/cls argument."""
         # Only non-static methods are special.
         functype = func.type
+        if info.is_no_type_check:
+            func.type = AnyType(TypeOfAny.special_form)
+            return
         if not func.is_static:
             if not func.arguments:
                 self.fail('Method must have at least one argument', func)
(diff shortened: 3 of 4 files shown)


lunixbochs:
Can a maintainer look at this / comment?

ilevkivskyi:
Few comments:
* This is still a draft, please make it non-draft and resolve merge conflicts if you want to get a review
* Your `decorator.name` check is wrong, you should check for a `RefExpr` and its `fullname` instead.
* I don't think we need a new flag for this. I think `fallback_to_any` plus replicating effect of `@no_type_check` for every method should be enough.
* Please add tests for this, I think https://github.com/python/mypy/blob/master/test-data/unit/check-classes.test is the best place.

lunixbochs:
- At the point of `SemanticAnalyzer.analyze_class`, all `defn.decorators` are still of type `NameExpr`. How/where can I get `RefExpr` decorators instead at that point?
- I used the special flag because I needed to propagate the behavior to semanal3 and analyze_instance_member_access, which `fallback_to_any` isn't expressive enough to do. I don't see how to propagate the behavior to methods at all at the point where we're just skipping analyze_class.

msullivan:
I'm going to close this because it is against the old semantic analyzer which is totally gone now. Feel free to redo this against the current code base and I'll review it.

PR #6585, opened 2019-03-24
Loosen base class attribute compatibility checks when attribute is redefined

### LINKED ISSUE #2619: Base class function signature compatibility checks overly strict
Actual example of this code can be found in Python 3.6's enum with `Flag` and `IntFlag`.

**Minimal repro:**:
```py3
from typing import TypeVar

_T = TypeVar('_T')


class Base:
    def __or__(self: _T, other: _T) -> _T: ...
    def __and__(self: _T, other: _T) -> _T: ...
    def __xor__(self: _T, other: _T) -> _T: ...


class Other(int, Base):  # <-- errors here
    def __or__(self: _T, other: Union[int, _T]) -> _T: ...  # type: ignore
    def __and__(self: _T, other: Union[int, _T]) -> _T: ...  # type: ignore
    def __xor__(self: _T, other: Union[int, _T]) -> _T: ...  # type: ignore
```

**Expected behavior**: no errors.
**Actual behavior**:
```
f:11: error: Definition of "__and__" in base class "int" is incompatible with definition in base class "Base"
f:11: error: Definition of "__or__" in base class "int" is incompatible with definition in base class "Base"
f:11: error: Definition of "__xor__" in base class "int" is incompatible with definition in base class "Base"
```

The errors would be fine if `Other` didn't redefine the methods itself. In this case the incompatibility in base classes is irrelevant.

syastrov:
This eliminates the error `Definition of "Nested" in base class "Base1" is incompatible with definition in base class "Base2"` in the following code, where `Nested` is redefined in `A`:
```
T = TypeVar('T', covariant=True)
class GenericBase(Generic[T]):
    pass
class Base1:
    Nested: GenericBase['Base1']
class Base2:
    Nested: GenericBase['Base2']
class A(Base1, Base2):
    Nested: GenericBase['A']
```

- In the case of multiple inheritance, don't give errors about definitions of an
  attribute in base classes being incompatible when the attribute is redefined.
  The redefinition must itself be compatible with all (non-type-ignored)
  definitions of the attribute in all base classes.
  This is achieved by making the following change to checking of incompatible
  types in assignments.
- Don't stop checking after the first base where the attribute is defined
  when checking for incompatible types in assignments.
  There is still a maximum of one "Incompatible type in assignment" error
  per assignment.

Resolves #2619

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1870,6 +1870,9 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                         # Show only one error per variable
                         break
 
+            direct_bases = lvalue_node.info.direct_base_classes()
+            last_immediate_base = direct_bases[-1] if direct_bases else None
+
             for base in lvalue_node.info.mro[1:]:
                 # Only check __slots__ against the 'object'
                 # If a base class defines a Tuple of 3 elements, a child of
@@ -1893,6 +1896,9 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                         # Only show one error per variable; even if other
                         # base classes are also incompatible
                         return True
+                    if base == last_immediate_base:
+                        # At this point, the attribute was found to be compatible with all immediate parents,
+                        break
         return False
 
     def check_compatibility_super(self, lvalue: RefExpr, lvalue_type: Optional[Type],
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1896,8 +1896,9 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                         # Only show one error per variable; even if other
                         # base classes are also incompatible
                         return True
-                    if base == last_immediate_base:
-                        # At this point, the attribute was found to be compatible with all immediate parents,
+                    if base is last_immediate_base:
+                        # At this point, the attribute was found to be compatible with all
+                        # immediate parents.
                         break
         return False
 
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1602,7 +1602,10 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         # Verify that inherited attributes are compatible.
         mro = typ.mro[1:]
         for i, base in enumerate(mro):
-            for name in base.names:
+            # Attributes defined in both the type and base are skipped.
+            # Normal checks for attribute compatibility should catch any problems elsewhere.
+            non_overridden_attrs = base.names.keys() - typ.names.keys()
+            for name in non_overridden_attrs:
                 for base2 in mro[i + 1:]:
                     # We only need to check compatibility of attributes from classes not
                     # in a subclass relationship. For subclasses, normal (single inheritance)
@@ -1890,7 +1893,6 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                         # Only show one error per variable; even if other
                         # base classes are also incompatible
                         return True
-                    break
         return False
 
     def check_compatibility_super(self, lvalue: RefExpr, lvalue_type: Optional[Type],
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -4020,7 +4020,6 @@ class C(B):
 [out]
 main:6: error: Incompatible types in assignment (expression has type "str", base class "A" defined the type as "int")
 
-
 [case testClassIgnoreType_RedefinedAttributeTypeIgnoredInChildren]
 class A:
     x = 0
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -3725,7 +3725,6 @@ class C(B):
     a = "a"
 [out]
 main:4: error: Incompatible types in assignment (expression has type "str", base class "A" defined the type as "int")
-main:6: error: Incompatible types in assignment (expression has type "str", base class "A" defined the type as "int")
 
 [case testVariableSubclassTypeOverwriteImplicit]
 class A:
@@ -4018,7 +4017,6 @@ class B(A):
 class C(B):
     x = ''
 [out]
-main:6: error: Incompatible types in assignment (expression has type "str", base class "A" defined the type as "int")
 
 [case testClassIgnoreType_RedefinedAttributeTypeIgnoredInChildren]
 class A:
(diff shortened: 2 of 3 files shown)


test-data/unit/check-multiple-inheritance.test
```diff
@@ -228,7 +228,7 @@ class B(A, int): pass
 from typing import Callable, TypeVar
 T = TypeVar('T')
 class A(B, C):
-    def f(self): pass
+    pass
```
syastrov: Just wanted to note that this change was necessary so that the original error would still be triggered (since redefining the method `f` would mean that the compatibility checks which may trigger the error would be skipped, due to my other changes)

syastrov:
Does anyone have any comments about the change in logic or code nitpicks? :)

ilevkivskyi:
I just returned from vacation. I will likely take a look at this later this week.

test-data/unit/check-classes.test
```diff
@@ -3725,6 +3725,7 @@ class C(B):
     a = "a"
 [out]
 main:4: error: Incompatible types in assignment (expression has type "str", base class "A" defined the type as "int")
+main:6: error: Incompatible types in assignment (expression has type "str", base class "A" defined the type as "int")
```
ilevkivskyi: This looks like a false positive. It can cause errors in code that already passes mypy.

I assume this is related to the `break` statement you removed.

test-data/unit/check-classes.test
```diff
@@ -4009,14 +4010,25 @@ class B(A):
...
-[case testClassIgnoreType]
+[case testClassIgnoreType_RedefinedAttributeAndGrandparentAttributeTypesNotIgnored]
 class A:
     x = 0
 class B(A):
     x = ''  # type: ignore
 class C(B):
     x = ''
 [out]
+main:6: error: Incompatible types in assignment (expression has type "str", base class "A" defined the type as "int")
```
ilevkivskyi: Same as above. A user might want to ignore the error _once_ in the base class, instead of ignoring it in _every_ subclass.

test-data/unit/check-multiple-inheritance.test
```diff
@@ -461,6 +461,128 @@ class Combo(Base2, Base1): ...
...
+class A(object):
+    def x(self) -> 'A':
+        return self
+
+class B(object):
+    def x(self) -> 'B':
+        return self
+
+class C(A, B): # E: Definition of "x" in base class "A" is incompatible with definition in base class "B"
+    pass
```
ilevkivskyi: Thanks for adding all the test cases! 👍

test-data/unit/check-multiple-inheritance.test
```diff
@@ -461,6 +461,128 @@ class Combo(Base2, Base1): ...
...
+class Base1:
+    Nested: GenericBase['Base1']
+class Base2:
+    Nested: GenericBase['Base1']
+class A(Base1, Base2):
+    Nested: GenericBase['Base1']
+[out]
+
+
+
```
ilevkivskyi: Keep only one empty line between tests.

test-data/unit/check-multiple-inheritance.test
```diff
@@ -461,6 +461,128 @@ class Combo(Base2, Base1): ...
 [out]
 main:10: error: Definition of "NestedVar" in base class "Base2" is incompatible with definition in base class "Base1"
 
+[case testMultipleInheritance_NestedVariableOverriddenWithCompatibleType]
+
```
ilevkivskyi: No need for empty line after test case header (here and everywhere below).

mypy/checker.py
```diff
@@ -1890,7 +1893,6 @@ def check_compatibility_all_supers(self, lvalue: RefExpr, lvalue_type: Optional[
                         # Only show one error per variable; even if other
                         # base classes are also incompatible
                         return True
-                    break
```
ilevkivskyi: I think in order to maintain consistency with the current behavior, you still need this break after last _non-immediate_ base.

ilevkivskyi: EDIT: (non-immediate -> immediate).

test-data/unit/check-classes.test
syastrov: This is slightly different than the above situation, as here, there was no error before.

I think it depends on how you want to interpret the `# type: ignore`.
In the current behavior, because `B.x` has an ignore, you are no longer warned when you redefine `x` in `C.x` that that redefinition is actually incompatible with ancestor's `A.x`.

A library author may have defined `B.x` with the ignore comment, and let us say that you are writing class `C` yourself. I would find it unexpected that because the library author chose to add an ignore comment on `B.x` that that would suppress a certain type of error when defining my own class.
I may not know in detail how `B` is defined/type-annotated, so I may reasonably expect that the type checker would perform its normal checks for my definition of class `C` (in this case, that it is compatible with `A`).

Are there other places in mypy where `# type: ignore` comments "spread" in that sense?

test-data/unit/check-classes.test
syastrov: Yes, it's related to the `break` statement being removed.

I wouldn't say that it is a false positive. `C.a` is incompatible with `A.a`, just as `B.a` is incompatible with `A.a`.

syastrov: Mypy simply would not show you the error before. If you fixed the compatibility problem in `B.a`, then you would get an error about `C.a`. Isn't it unexpected to start getting new errors when mypy could have reported it the first time?

test-data/unit/check-multiple-inheritance.test
syastrov: 👍

test-data/unit/check-multiple-inheritance.test
syastrov: 👍

test-data/unit/check-multiple-inheritance.test
syastrov: Thank you for reviewing.

mypy/checker.py
syastrov: See my replies below.

test-data/unit/check-classes.test
syastrov: Oops:
~(in this case, that it is compatible with `A`)~
should be:
(in this case, that `C.x` is actually compatible with `A.x`)

syastrov:
Thank you for the review.
I am not 100% convinced that the we should change the current behavior regarding incompatible types in assignment checks, although I wanted to challenge it and get your feedback, since it is a similar issue. Perhaps that part belongs in a separate issue though, and we can retain the original behavior here. What do you think?

syastrov:
Having understood the code a bit more and reading your comments, I agree with you that this behavior should be kept, so I have restored it. I followed your suggestion, so it should still handle the case of multiple inheritance.

Thanks for taking the time to respond!

Issue #5967, opened 2018-11-28
Type ignore comment has no effect after argument type comment


JukkaL:
This program generates an error on the first line even though there is a `# type: ignore` comment:

```py
def f(x,  # type: x  # type: ignore
      ):
    # type: (...) -> None
    pass
```

Similarly, this generates an error even though it shouldn't:

```py
def f(x=y,  # type: int  # type: ignore  # Name 'y' not defined
      ):
    # type: (...) -> None
    pass
```

ethanhs:
For your first example, here is the AST as parsed by typed_ast:
```
Module(
    body=[
        FunctionDef(
            lineno=2,
            col_offset=0,
            name='f',
            args=arguments(
                args=[arg(lineno=2, col_offset=6, arg='x', annotation=None, type_comment='x  # type: ignore')],
                vararg=None,
                kwonlyargs=[],
                kw_defaults=[],
                kwarg=None,
                defaults=[],
            ),
            body=[Pass(lineno=5, col_offset=4)],
            decorator_list=[],
            returns=None,
            type_comment='(...) -> None',
        ),
    ],
    type_ignores=[],
)
```
Notice the `type_comment='x  # type: ignore'`.
We generally ignore comments after the first type comment, but it should be easy to make an exception for `#\s+type:\s+ignore` and add that line to the list of lines to ignore.

JukkaL:
This happens also with variable type comments:

```py
x = 1 # type: str # type: ignore
```

JukkaL:
Increasing priority since it's not at all obvious how to work this around in general.

ethanhs:
I believe I have hacked together a general implementation in fastparse. I just do a regex search for a `# type: ignore` when parsing the type comment and keep track of these additional type ignores in the Ast converter class, then add them to the set of ignores when visiting the `Module` node.

Though this likely should be implemented in typed_ast itself...

PR #6591, opened 2019-03-26
 Allow type ignores after type comments 

ethanhs:
This is a hack in fastparse, but allows for the following to pass typechecking:

`x = 1  # type: str  # type: ignore`

This also handles the edge case where there is a `# type: ignore` in a comment, which we don't want to pick up. See the tests for more examples.

Fixes https://github.com/python/mypy/issues/5967

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -1,3 +1,4 @@
+import re
 import sys
 
 from typing import (
@@ -124,6 +125,8 @@ _dummy_fallback = Instance(MISSING_FALLBACK, [], -1)  # type: Final
 
 TYPE_COMMENT_SYNTAX_ERROR = 'syntax error in type comment'  # type: Final
 
+TYPE_IGNORE_PATTERN = re.compile(r'#\s*type:\s*ignore')
+
 
 # Older versions of typing don't allow using overload outside stubs,
 # so provide a dummy.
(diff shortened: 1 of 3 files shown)


mypy/fastparse.py
```diff
@@ -393,11 +408,11 @@ def translate_module_id(self, id: str) -> str:
 
     def visit_Module(self, mod: ast3.Module) -> MypyFile:
         body = self.fix_function_overloads(self.translate_stmt_list(mod.body))
-
+        ignores = [ti.lineno for ti in mod.type_ignores] + self.extra_type_ignores
```
gvanrossum: That's an extra copy -- maybe use `.extend()`?

mypy/fastparse.py
```diff
@@ -184,18 +187,28 @@ def parse_type_comment(type_comment: str,
...
     else:
+        # We need to handle the case where an
+        # extra # type: ignore is in a comment, like this line.
+        # Therefore we split on comment delineations and check
+        # only the first possibility for a type ignore.
+        parts = type_comment.split('#')
+        if len(parts) > 1:
+            _, possible_ignore, *rest = parts
+        else:
+            possible_ignore = type_comment
```
gvanrossum: You don't need to check the type comment itself for looking like `ignore` -- typed_ast (or ast in 3.8) has already done that.

I also wonder if you could omit splitting on `#` at all and just use a slightly more complex regex. I think something like this would do it:
```
extra_ignore = re.match(r'[^#]*#\s*type:\s*ignore\s*($|#)', type_comment) is not None
```

mypy/fastparse.py
```diff
@@ -216,7 +229,7 @@ def parse_type_string(expr_string: str, expr_fallback_name: str,
     code with unicode_literals...) and setting `assume_str_is_unicode` accordingly.
     """
     try:
-        node = parse_type_comment(expr_string.strip(), line=line, errors=None,
+        _, node = parse_type_comment(expr_string.strip(), line=line, errors=None,
                                   assume_str_is_unicode=assume_str_is_unicode)
```
gvanrossum: Reindent here.

mypy/fastparse.py
```diff
@@ -1215,11 +1243,11 @@ def visit_raw_str(self, s: str) -> Type:
         # An escape hatch that allows the AST walker in fastparse2 to
         # directly hook into the Python 3.5 type converter in some cases
         # without needing to create an intermediary `Str` object.
-        return (parse_type_comment(s.strip(),
+        _, typ = parse_type_comment(s.strip(),
                                    self.line,
```
gvanrossum: Indent (one more space).

mypy/fastparse2.py
```diff
@@ -547,8 +549,10 @@ def visit_Delete(self, n: ast27.Delete) -> DelStmt:
     def visit_Assign(self, n: ast27.Assign) -> AssignmentStmt:
         typ = None
         if n.type_comment:
-            typ = parse_type_comment(n.type_comment, n.lineno, self.errors,
+            extra_ignore, typ = parse_type_comment(n.type_comment, n.lineno, self.errors,
                                      assume_str_is_unicode=self.unicode_literals)
```
gvanrossum: Reindent (and several more times below).

Issue #6564, opened 2019-03-18
Stop supporting Python 3.4 to run mypy


gvanrossum:
Python 3.4 has reached its end of life: https://mail.python.org/pipermail/python-dev/2019-March/156705.html

We can still support targeting it with `--python-version 3.4`, but we should no longer feel bound to support *running* mypy with 3.4.

ilevkivskyi:
+1 from me. Some PRs fail on 3.4 because it has some syntax limitations, I think the latest example is Michael's union PR.

JukkaL:
Let's do this. 

I think that mypyc doesn't support 3.4. This means that once we ship a mypyc-compiled mypy by default, we should either drop 3.4 support or add Python 3.4 support to mypyc. The latter seems pointless because of the EoL, so dropping 3.4 support seems to make the most sense from this point of view as well.

ethanhs:
Apologies, fat fingered. I am about ready with a PR.

PR #6592, opened 2019-03-26
Drop support for running with 3.4

ethanhs:
Fixes https://github.com/python/mypy/issues/6564

This PR was a bit larger than I expected, but after the test failure in https://github.com/python/mypy/pull/6591 I am keen to move forward with dropping support for 3.4.


mypy/fastparse.py
```diff
@@ -98,21 +98,17 @@ def ast3_parse(source: Union[str, bytes], filename: str, mode: str,
         NamedExpr = Any
         Constant = Any
 except ImportError:
-    if sys.version_info.minor > 2:
```
ethanhs: I realized this version check doesn't make much sense anymore, so I just removed it. Everyone getting to importing fastparse should already have checked they are on a compatible version of Python, especially with the extra checks I added to stubgen and dmypy.

mypy/util.py
```diff
@@ -257,3 +257,15 @@ def hard_exit(status: int = 0) -> None:
...
+
+def check_python_version(program: str) -> None:
+    """Report issues with the Python used to run mypy, dmypy, or stubgen"""
+    # Check for known bad Python versions.
+    if sys.version_info[:2] < (3, 5):
+        sys.exit("Running {name} with Python 3.4 or lower is not supported; "
+                 "please upgrade to 3.5 or newer".format(name=program))
+    # this can be deleted once we drop support for 3.5
+    if sys.version_info[:3] == (3, 5, 0):
+        sys.exit("Running stubgen with Python 3.5.0 is not supported; "
```
msullivan: This should be `program`, I think

ethanhs:
>Have you tried to run it with 3.4? Does it even make it to that error message? For all entry points?

Yes I checked this in a docker container with 3.4.10, it does indeed cause the error to be shown.

I tested `mypy`,`dmypy`, and `stubgen`.

PR #6596, opened 2019-03-27
Produce a distribution named 'mypy' when built in mypyc mode

msullivan:
This will allow us to distribute compiled packages by default for mypy.

This will require some changes to our dropbox internal infrastructure for using mypy_mypyc wheels but it shouldn't be a big deal.


msullivan:
The breakage is due to something with the new version of tox asdf

Issue #6588, opened 2019-03-25
Release 0.700 planning


JukkaL:
I plan to release mypy 0.700 next Wed (Apr 3) and cut the release branch for 0.700 this Wed (Mar 27). This will be the first release with mypyc-compiled mypy available by default (on supported platforms).

Please add comments to this issue if you have any PR you'd like to be included. Ideally those PRs should be landed before the release branch is cut, to minimize cherry-picking.

ethanhs:
I would appreciate it if https://github.com/python/mypy/pull/6592 was merged, and https://github.com/python/mypy/pull/6591 should be pretty low risk as well.

JukkaL:
I've cut the release branch.

msullivan:
I think we might want to write and cherry-pick some plugin hooks to support the Django plugin, which currently monkeypatches `load_graph` and `State` and so doesn't work with mypyc.

msullivan:
Two cherry-pick requests:
 * #6597 to fix CI on the release branch
 * #6596 to make compiled mypy packages

msullivan:
And another, though it still needs some finalization:
 * #6598 to allow the django plugin to stop monkey patching

msullivan:
I am sorry about the relentless cherry-pick requests but we also might want to cherry-pick:
 * #6601 - update mypyc - previously mypyc had a specified dependency on the wrong version of typed_ast. ... though it seemed to work in practice OK, so maybe it doesn't really matter?

gvanrossum:
1.3.0 introduced the new typed_ast synced with Python 3.7. Presumably mypyc itself doesn't use Python 3.7 features, so assuming this is only about mypyc's own dependencies, not those of the binary it builds, it's not the end of the world. It's also possible that somehow magically mypy's typed_ast dependency (which *was* updated) overrides here?

msullivan:
> 1.3.0 introduced the new typed_ast synced with Python 3.7. Presumably mypyc itself doesn't use Python 3.7 features, so assuming this is only about mypyc's own dependencies, not those of the binary it builds, it's not the end of the world. It's also possible that somehow magically mypy's typed_ast dependency (which _was_ updated) overrides here?

mypyc includes a vendored copy of mypy, and has to duplicate its typed_ast dependency. But we forgot, so it will use typed_ast 1.2.0 and a mypy that expects to be using 1.3.1. But since mypy doesn't use python 3.7, this might be fine?

msullivan:
[...]

PR #6597, opened 2019-03-27
Pin tox version to 3.7.0

msullivan:
tox 3.8.0 broke something about how we specified dependencies.
I've filed a bug with tox, but let's just pin versions for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ jobs:
 
 install:
 - pip install -U pip setuptools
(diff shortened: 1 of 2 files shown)


msullivan:
This needs to be cherry-picked into #6588 if we want CI to work on the release branch.

msullivan:
https://github.com/tox-dev/tox/issues/1207 is the bug I filed against tox

PR #6598, opened 2019-03-27
Add a get_additional_deps hook to Plugin to support django-stubs

msullivan:
django-stubs needs to be able to add in additional dependencies to
modules and is currently using monkeypatching to do it.

That is 1) not great and 2) breaks under mypyc.
Add a hook to support generating additional dependencies from a plugin.

One thing I'm unsure about with this PR is the handling of priorities. I'm not sure that being able to specify priorities is necessary, but I am reluctant to leave them out of the API and then need to either have two hooks or a breaking change when it turns out that something does need them. The priority constants are all defined in mypy.build, which plugins probably shouldn't be importing, but moving them to a new file seems like overkill for a super marginal plugin use case.

This should allow a fix for https://github.com/mkurnikov/django-stubs/issues/48. @mkurnikov could you try updating django-stubs to work with this PR so that we can validate it and cherry-pick it for the release?

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -374,10 +374,14 @@ class Plugin(CommonPluginApi):
     def get_additional_deps(self, file: MypyFile) -> List[Tuple[int, str, int]]:
         """Customize dependencies for a module.
 
-        This hook allows adding in new dependencies for a module. It is called
-        after parsing a file but before analysis.
+        This hook allows adding in new dependencies for a module. It
+        is called after parsing a file but before analysis. This can
+        be useful if a library has dependencies that are dynamic based
+        on configuration information, for example.
 
         Returns a list of (priority, module name, line number) tuples.
+
+        The line number can be -1 when there is not a known real line number.
 
         Priorities are defined in mypy.build (but maybe shouldn't be).
         10 is a good choice for priority.
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -371,6 +371,19 @@ class Plugin(CommonPluginApi):
         assert self._modules is not None
         return lookup_fully_qualified(fullname, self._modules)
 
+    def get_additional_deps(self, file: MypyFile) -> List[Tuple[int, str, int]]:
+        """Customize dependencies for a module.
+
+        This hook allows adding in new dependencies for a module. It is called
+        after parsing a file but before analysis.
+
+        Returns a list of (priority, module name, line number) tuples.
+
+        Priorities are defined in mypy.build (but maybe shouldn't be).
+        10 is a good choice for priority.
+        """
+        return []
+
     def get_type_analyze_hook(self, fullname: str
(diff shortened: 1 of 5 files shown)


test-data/unit/check-custom-plugin.test
```diff
@@ -585,3 +584,17 @@ reveal_type(instance(2))  # E: Revealed type is 'builtins.float*'
 [file mypy.ini]
 [[mypy]
 plugins=<ROOT>/test-data/unit/plugins/callable_instance.py
+
+
```
JukkaL: Style nit: Extra empty line

mypy/plugin.py
```diff
@@ -371,6 +371,19 @@ def lookup_fully_qualified(self, fullname: str) -> Optional[SymbolTableNode]:
         assert self._modules is not None
         return lookup_fully_qualified(fullname, self._modules)
 
+    def get_additional_deps(self, file: MypyFile) -> List[Tuple[int, str, int]]:
+        """Customize dependencies for a module.
+
+        This hook allows adding in new dependencies for a module. It is called
+        after parsing a file but before analysis.
+
+        Returns a list of (priority, module name, line number) tuples.
```
JukkaL: Document that line number can be -1 if there's no specific line or the line is unknown.

mypy/plugin.py
```diff
@@ -371,6 +371,19 @@ def lookup_fully_qualified(self, fullname: str) -> Optional[SymbolTableNode]:
         assert self._modules is not None
         return lookup_fully_qualified(fullname, self._modules)
 
+    def get_additional_deps(self, file: MypyFile) -> List[Tuple[int, str, int]]:
+        """Customize dependencies for a module.
+
+        This hook allows adding in new dependencies for a module. It is called
+        after parsing a file but before analysis.
```
JukkaL: Maybe also discuss why you'd want to add new module dependencies.

mkurnikov:
I've checked in https://github.com/mkurnikov/django-stubs/pull/60, all tests pass with new hook and disabled monkeypatches 
(CI fails, it is configured for stock mypy).

mypy/plugin.py
```diff
@@ -371,6 +371,23 @@ def lookup_fully_qualified(self, fullname: str) -> Optional[SymbolTableNode]:
...
+        be useful if a library has dependencies that are dynamic based
+        on configuration information, for example.
+
+        Returns a list of (priority, module name, line number) tuples.
+
+        The line number can be -1 when there is not a known real line number.
+
+        Priorities are defined in mypy.build (but maybe shouldn't be).
+        10 is a good choice for priority.
+        """
```
ilevkivskyi: This (+ an example) should be also added to `mypy/docs/source/extending_mypy.rst`.

@msullivan Will you have time to make a PR? I suppose this will be announced in the release blog post, it would be great to have some docs to link to.

PR #6602, opened 2019-03-28
Ignore flake8 warnings about assert False

msullivan:


msullivan:
(Maybe we should pin a flake8 version?)

setup.cfg
```diff
@@ -41,7 +41,8 @@ exclude =
 #   B3??: Python 3 compatibility warnings
 #   B006: use of mutable defaults in function signatures
 #   B007: Loop control variable not used within the loop body.
-ignore = E251,E128,F401,W601,E701,W503,W504,E704,E402,B3,B006,B007
+#   B011: Don't use assert False
```
gvanrossum: What does flake8 want you to use???

gvanrossum:
> (Maybe we should pin a flake8 version?)
Pinning deps seems to be a trend. :-) Though we should also strive to upgrade our dependencies regularly, which may become yet more busywork.

Also, the flake errors that this failed on look reasonable. Just poorly timed during our release cycle. :-)

setup.cfg
msullivan: `Do not call assert False since python -O removes these calls. Instead callers should raise AssertionError().`

msullivan:
> Also, the flake errors that this failed on look reasonable. Just poorly timed during our release cycle. :-)

Yeah. I am going to suppress them in config anyways, because of release cycle concerns, and then go back and actually fix them.

msullivan:
(The new errors come from flake8-bugbear, it seems)

msullivan:
> > Also, the flake errors that this failed on look reasonable. Just poorly timed during our release cycle. :-)
> 
> Yeah. I am going to suppress them in config anyways, because of release cycle concerns, and then go back and actually fix them.

Actually, in going to fix them, I've changed my mind. Basically every fix is adding either `# type: ignore` or `# noqa`. Might as well just keep it ignored.

gvanrossum:
> Actually, in going to fix them, I've changed my mind. Basically every fix is adding either # type: ignore or # noqa. Might as well just keep it ignored.

Hm, the ones about `getattr(x, 'name')` being in all ways inferior to `x.name` are correct, so those should be fixed using that substitution.

PR #6606, opened 2019-04-01
New semantic analyzer: support module-level __getattr__

### LINKED ISSUE #6314: New semantic analyzer: module-level __getattr__

JukkaL:
This is slightly different from how it's implemented in the old
semantic analyzer. We don't try to check the return type, since
we can't generally do this during the initial pass. The approach
used in the old semantic samalyzer is hacky and does not always
work.

Fixes #6314.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -1628,6 +1628,8 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             # If it is still not resolved, check for a module level __getattr__
             if (module and not node and (module.is_stub or self.options.python_version >= (3, 7))
                     and '__getattr__' in module.names):
+                # We use the fullname of the orignal definition so that we can
+                # detect whether two imported names refer to the same thing.
                 fullname = import_id + '.' + id
                 gvar = self.create_getattr_var(module.names['__getattr__'], imported_id, fullname)
                 if gvar:
@@ -3940,6 +3942,12 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         """Create a dummy variable using module-level __getattr__ return type.
 
         If not possible, return None.
+
+        Note that multiple Var nodes can be created for a single name. We
+        can use the from_module_getattr and the fullname attributes to
+        check if two dummy Var nodes refer to the same thing. Reusing Var
+        nodes would require non-local mutable state, which we prefer to
+        avoid.
         """
         if isinstance(getattr_defn.node, (FuncDef, Var)):
             if isinstance(getattr_defn.node.type, CallableType):
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -4187,7 +4187,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                     and not is_same_var_from_getattr(old_node, new_node)):
                 if isinstance(new_node, (FuncDef, Decorator)):
                     self.add_func_redefinition(names, name, symbol)
-                if not (isinstance(symbol.node, (FuncDef, Decorator))
+                if not (isinstance(new_node, (FuncDef, Decorator))
                         and self.set_original_def(old_node, new_node)):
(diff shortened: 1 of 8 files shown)


JukkaL:
The failure is a real one. I'm working on a fix.

mypy/newsemanal/semanal_pass1.py
```diff
@@ -44,17 +47,35 @@ def visit_file(self, file: MypyFile, fnam: str, mod_id: str, options: Options) -
...
+        super().visit_func_def(node)
+        self.is_global_scope = old_global_scope
+        file_node = self.cur_mod_node
+        if (self.is_global_scope
+                and file_node.is_stub
+                and node.name() == '__getattr__'
+                and file_node.is_package_init_file()):
+            # __init__.pyi with __getattr__ means that any submodules are assumed
+            # to exist, even if there is no stub. Note that we can't verify that the
+            # return type is compatible, since we haven't bound types yet.
```
ilevkivskyi: This is OK, but I think it is something worth documenting.

mypy/newsemanal/semanal.py
```diff
@@ -1628,10 +1628,7 @@ def visit_import_from(self, imp: ImportFrom) -> None:
             # If it is still not resolved, check for a module level __getattr__
             if (module and not node and (module.is_stub or self.options.python_version >= (3, 7))
                     and '__getattr__' in module.names):
-                if self.type:
-                    fullname = self.type.fullname() + "." + imported_id
-                else:
-                    fullname = self.qualified_name(imported_id)
+                fullname = import_id + '.' + id
```
ilevkivskyi: This is a bit subtle, I would add a comment saying that we use the original full name, so that multiple imports refer to the same "original definition".

mypy/newsemanal/semanal.py
```diff
@@ -4508,3 +4507,12 @@ def names_modified_in_lvalue(lvalue: Lvalue) -> List[NameExpr]:
...
     return []
+
+
+def is_same_var_from_getattr(n1: Optional[SymbolNode], n2: Optional[SymbolNode]) -> bool:
+    """Do n1 and n2 refer to the same Var derived from module-level __getattr__?"""
+    return (isinstance(n1, Var)
+            and n1.from_module_getattr
+            and isinstance(n2, Var)
+            and n2.from_module_getattr
+            and n1.fullname() == n2.fullname())
```
ilevkivskyi: Can we just insert the newly created `Var` in the namespace we import from instead of using this function? Then we will not need also the new flag and all imports will actually point to the same `Var`.

mypy/newsemanal/semanal.py
```diff
@@ -3874,13 +3871,11 @@ def lookup_qualified(self, name: str, ctx: Context,
...
                                 and not self.is_incomplete_namespace(namespace)):
+                            fullname = namespace + '.' + '.'.join(parts[i:])
                             gvar = self.create_getattr_var(names['__getattr__'],
-                                                           parts[i], parts[i])
+                                                           parts[i], fullname)
                             if gvar:
                                 n = SymbolTableNode(GDEF, gvar)
-                                # TODO: Enable this (the purpose is likely to preserve the
-                                #       identity of the Var node if there are multiple imports)
-                                # names[name] = n
```
ilevkivskyi: It looks like my suggestion about inserting the `Var` is about this. If it doesn't work for some reason, this should be explained in a comment.

mypy/newsemanal/semanal.py
JukkaL: Modifying the ASTs of other modules makes things harder to reason about. For example, the other symbol table may already have been serialized, and now the serialized symbol table doesn't match what's in memory, which could be confusing.

mypy/newsemanal/semanal_pass1.py
JukkaL: This could be mentioned in mypy release notes when we announce the new semantic analyzer.

PR #6615, opened 2019-04-03
 New semantic analyzer: fix "with" statement with type annotation

JukkaL:
Keep the original unanalyzed type around instead of modifying
the type in-place. This fixes a crash.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3139,6 +3139,8 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                         new_types.append(analyzed)
                         self.store_declared_types(n, analyzed)
 
+        s.analyzed_types = new_types
+
         self.visit_block(s.body)
 
     def visit_del_stmt(self, s: DelStmt) -> None:
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3102,18 +3102,18 @@ class NewSemanticAnalyzer(NodeVisitor[None],
     def visit_with_stmt(self, s: WithStmt) -> None:
         types = []  # type: List[Type]
 
-        if s.target_type:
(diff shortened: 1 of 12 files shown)


mypy/treetransform.py
```diff
@@ -285,7 +285,7 @@ def visit_with_stmt(self, node: WithStmt) -> WithStmt:
         return WithStmt(self.expressions(node.expr),
                         self.optional_expressions(node.target),
                         self.block(node.body),
-                        self.optional_type(node.target_type))
+                        self.optional_type(node.unanalyzed_type))
```
ilevkivskyi: Interestingly, we do the opposite for assignments, i.e. copy only the analyzed type. I think that we should copy both analyzed and unanalyzed types both here and for assignments.

mypy/newsemanal/semanal.py
```diff
@@ -3139,12 +3139,7 @@ def visit_with_stmt(self, s: WithStmt) -> None:
                         new_types.append(analyzed)
                         self.store_declared_types(n, analyzed)
 
-        # Reverse the logic above to correctly reassign target_type
-        if new_types:
-            if len(s.target) == 1:
-                s.target_type = new_types[0]
-            elif isinstance(s.target_type, TupleType):
-                s.target_type = s.target_type.copy_modified(items=new_types)
+        s.analyzed_types = new_types
```
ilevkivskyi: It looks like `analyzed_types` is only used in `deps.py`. Maybe we can update the logic there, so that we don't need this additional attribute? (Individual items can be already accessed on corresponding target `Var`s.)

JukkaL: I considered this but the logic to support different kinds of lvalues would be more complicated. Also, this is kind of consistent with how assignment statements work.

JukkaL: That is, the approach in the PR is consistent.

Issue #6617, opened 2019-04-03
Announcement issue for plugin API changes


msullivan:
The mypy plugin interface is experimental, unstable, and prone to change. In particular, there are no guarantees about backwards compatibility. Backwards incompatible changes may be made without a deprecation period.

We will, however, attempt to announce breaking changes in this issue, so that plugin developers can subscribe to this issue and be notified.

Breaking changes fall into three broad categories:
1. Changes to the actual plugin API itself
2. Changes to parts of the "implicit plugin API"---that is, internals that plugins are likely to use (representation of types, etc).
3. Changes to things that really aren't plausibly part of the plugin API (but, of course, that some plugins might be using anyway...)

Issues in category 1 will be consistently announced here, issues in category 3 will probably be announced here only if problems are reported, and issues in category 2 will be somewhere in the middle.

msullivan:
And we'll start this off with a belated category 3 announcement:

The default wheels for mypy 0.700 are compiled with mypyc. This breaks monkey-patching of mypy internals.

If you are the author of a mypy plugin that relies on monkey-patching mypy internals, get in touch with us and we can probably find a better approach. (#6598 added hooks needed by the django plugin.)

#6598

gvanrossum:
(Sorry for the accidental unpin. Fixed now.)

JukkaL:
The new semantic analyzer requires changes to some plugins, especially those that modify classes. In particular, hooks may be executed multiple times for the same definitions. PR #7135 added documentation about how to support the new semantic analyzer. 

Note that mypy 0.720 (to be released soon) will enable the semantic analyzer by default, and the next release after that will remove the old semantic analyzer.

PRs #7136, #7132, #7096, #6987, #6984, #6724 and #6515 contain examples of changes that may be needed to plugins.

To test that a plugin works with the semantic analyzer, you should have test cases that cause mypy to analyze things twice. The easiest way to achieve is to add a forward reference to a type at module top level:

```py
forwardref: C   # Forward reference to C causes deferral
class C: pass

# ... followed by whatever you want to test
```

msullivan:
PR #7829 makes all `name` and `fullname` methods in mypy into properties. This will unfortunately require changes to many plugins. We've decided that it is worth removing a long-standing pain point and that it is better to do it sooner than later.

sed can be used to update code to the new version with something like `sed -i -e 's/\.name()/.name/g' -e 's/\.fullname()/.fullname/g'`

[...]

PR #6618, opened 2019-04-03
Add some more plugin docs

msullivan:
* Link to plugin changes announcement issue (#6617)
 * Document get_additional_deps()

### DIFF
--- a/docs/source/extending_mypy.rst
+++ b/docs/source/extending_mypy.rst
@@ -62,7 +62,7 @@ new first class kinds of types.
    no guarantees about backwards compatibility.
 
    Backwards incompatible changes may be made without a deprecation period,
-   but we will attempt to announce them in
+   but we will announce them in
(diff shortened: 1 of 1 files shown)


docs/source/extending_mypy.rst
```diff
@@ -59,8 +59,11 @@ new first class kinds of types.
    The plugin system is experimental and prone to change. If you want to write
    a mypy plugin, we recommend you start by contacting the mypy core developers
    on `gitter <https://gitter.im/python/typing>`_. In particular, there are
-   no guarantees about backwards compatibility. Backwards incompatible changes
-   may be made without a deprecation period.
+   no guarantees about backwards compatibility.
+
+   Backwards incompatible changes may be made without a deprecation period,
+   but we will attempt to announce them in
```
ilevkivskyi: I think we should just say "we will announce" (and actually do this :-))

msullivan: "WP:WEASEL" :P

PR #6624, opened 2019-04-04
Fix typo in types.py

mickeypash:
There was a typo where `NoneType` in a doc string was missing an `e`.
Apologies if this comes across as overkill.


mypy/types.py
```diff
@@ -453,7 +453,7 @@ class UninhabitedType(Type):
     This type is the bottom type.
     With strict Optional checking, it is the only common subtype between all
     other types, which allows `meet` to be well defined.  Without strict
-    Optional checking, NoneTyp fills this role.
+    Optional checking, NoneType fills this role.
```
ilevkivskyi: This is not a typo. There is a `NoneTyp` class used for internal representation of `None` type.

PR #6627, opened 2019-04-04
New semantic analyzer: fix deserialization of generated classes

ilevkivskyi:
In new semantic analyzer, `qualified_name()` returns a bare name if called inside a top-level (non-method) function. I am not sure for motivation and can't say it is generally wrong. Here is the hot-fix for a case where this is definitely wrong: generated classes.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2803,6 +2803,10 @@ class NewSemanticAnalyzer(NodeVisitor[None],
     def basic_new_typeinfo(self, name: str, basetype_or_fallback: Instance) -> TypeInfo:
         class_def = ClassDef(name, Block([]))
         class_def.fullname = self.qualified_name(name)
+        if self.is_func_scope() and not self.type:
+            # Full names of generated classes should always be prefixed with the module names
+            # even if they are nested in a function, since these classes will be (de-)serialized.
+            class_def.fullname = self.cur_mod_id + '.' + class_def.fullname
 
         info = TypeInfo(SymbolTable(), class_def, self.cur_mod_id)
         class_def.info = info
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2802,11 +2802,14 @@ class NewSemanticAnalyzer(NodeVisitor[None],
 
     def basic_new_typeinfo(self, name: str, basetype_or_fallback: Instance) -> TypeInfo:
         class_def = ClassDef(name, Block([]))
-        class_def.fullname = self.qualified_name(name)
         if self.is_func_scope() and not self.type:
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -2803,6 +2803,10 @@ def process_typevar_parameters(self, args: List[Expression],
     def basic_new_typeinfo(self, name: str, basetype_or_fallback: Instance) -> TypeInfo:
         class_def = ClassDef(name, Block([]))
         class_def.fullname = self.qualified_name(name)
+        if self.is_func_scope() and not self.type:
```
JukkaL: This would be better as an if/else statement.

mypy/newsemanal/semanal.py
```diff
@@ -2803,6 +2803,10 @@ def process_typevar_parameters(self, args: List[Expression],
...
+            # Full names of generated classes should always be prefixed with the module names
+            # even if they are nested in a function, since these classes will be (de-)serialized.
+            class_def.fullname = self.cur_mod_id + '.' + class_def.fullname
```
JukkaL: It looks like this can collide with another class name. Maybe you should use the `@line` suffix that we use elsewhere?

test-data/unit/check-incremental.test
```diff
@@ -4949,6 +4949,25 @@ NT = NamedTuple('BadName', [('x', int)])
...
+x: C
+[file a.py.2]
+from b import C
+x: C
+# touch
+[file b.py]
+class C: ...
+from collections import namedtuple
+def test() -> None:
+    NT = namedtuple('BadName', ['x', 'y'])
```
JukkaL: I wonder if we could instead not serialize the class, since it's only visible within the function body.

Another idea would be to ignore the `'BadName'` argument here and treat it as `NT`. Would this fix the issue?

mypy/newsemanal/semanal.py
ilevkivskyi: OK.

mypy/newsemanal/semanal.py
ilevkivskyi: This should be already done by class generation code. I will add a comment.

ilevkivskyi:
The issue for this is essentially https://github.com/python/mypy/issues/6422, but this is only partial fix.

test-data/unit/check-incremental.test
ilevkivskyi: Actually, this is a good idea. A type can leak to outer scope only from a method, so we can just don't serialize the class for non-method functions.

ilevkivskyi: As we discussed privately, we will stick to just fixing the full name for now.

PR #6630, opened 2019-04-04
New semantic analyzer: allow replacing submodule with a variable

ilevkivskyi:
Although this looks like an arbitrary exception from general rules, this:
* is an actual pattern used by some libraries
* matches old analyzer in most cases
* matches runtime semantics

### DIFF
--- a/test-data/unit/check-newsemanal.test
+++ b/test-data/unit/check-newsemanal.test
@@ -2049,3 +2049,20 @@ from pp import x as y
 def __getattr__(attr): ...
 [out2]
 tmp/a.py:2: error: Revealed type is 'Any'
+
+[case testNewAnanlyzerTrickyImportPackage]
+from lib import config
+import lib
+
+reveal_type(lib.config.x)  # E: Revealed type is 'builtins.int'
+reveal_type(config.x)  # E: Revealed type is 'builtins.int'
+
+[file lib/__init__.py]
+from lib.config import config
+
+[file lib/config.py]
+class Config:
+    x: int
+
+config = Config()
+[builtins fixtures/module.pyi]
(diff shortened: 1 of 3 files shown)


test-data/unit/check-newsemanal.test
```diff
@@ -2049,3 +2049,20 @@ from pp import x as y
 def __getattr__(attr): ...
 [out2]
 tmp/a.py:2: error: Revealed type is 'Any'
+
+[case testNewAnanlyzerTrickyImportPackage]
+from lib import config
+import lib
+
+reveal_type(lib.config.x)  # E: Revealed type is 'builtins.int'
+reveal_type(config.x)  # E: Revealed type is 'builtins.int'
```
JukkaL: What about trying to import `lib.config` as a module? Will it work? What about `import lib.config as c`?

test-data/unit/check-newsemanal.test
```diff
@@ -2049,3 +2049,20 @@ from pp import x as y
 def __getattr__(attr): ...
 [out2]
 tmp/a.py:2: error: Revealed type is 'Any'
+
+[case testNewAnanlyzerTrickyImportPackage]
```
JukkaL: Would it be useful to an incremental mode test case as well?

ilevkivskyi: Yes, I will add an incremental test.

test-data/unit/check-newsemanal.test
ilevkivskyi: Hm, OK, I will try adding these tests.

ilevkivskyi: OK so these two have different behavior in mypy:
* `import lib.config` actually imports `lib`, so when one writes `lib.config` later, the variable takes precedence.
* `import lib.config as tmp` finds module `lib.config` and binds it to `tmp`, so when one writes `tmp` later it refers to the module.

At runtime variable always wins. Fixing the second one is non-trivial, because we need to check whether there is a "shadowing" variable at every submodule level. Also this is kind of corner case to a corner case.

PR #6638, opened 2019-04-05
Eliminate all mypyc warnings when compiling mypy

msullivan:
This was a big obstacle that made me reluctant to set up CI.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3735,8 +3735,7 @@ def any_causes_overload_ambiguity(items: List[CallableType],
     return False
 
 
-def all_same_types(types: Iterable[Type]) -> bool:
-    types = list(types)
+def all_same_types(types: List[Type]) -> bool:
     if len(types) == 0:
         return True
     return all(is_same_type(t, types[0]) for t in types[1:])
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1495,7 +1495,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             # We try returning a precise type if we can. If not, we give up and just return 'Any'.
             if all_same_types(return_types):
                 return return_types[0], inferred_types[0]
-            elif all_same_types(erase_type(typ) for typ in return_types):
+            elif all_same_types([erase_type(typ) for typ in return_types]):
                 return erase_type(return_types[0]), erase_type(inferred_types[0])
             else:
                 return self.check_call(callee=AnyType(TypeOfAny.special_form),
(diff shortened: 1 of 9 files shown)


mypy/checkexpr.py
```diff
@@ -1495,7 +1495,7 @@ def infer_overload_return_type(self,
             # We try returning a precise type if we can. If not, we give up and just return 'Any'.
             if all_same_types(return_types):
                 return return_types[0], inferred_types[0]
-            elif all_same_types(erase_type(typ) for typ in return_types):
+            elif all_same_types([erase_type(typ) for typ in return_types]):
```
gvanrossum: Note that `all_same_types()` calls `list()` on its argument. Maybe skip that if it's already a `list`?

mypy/messages.py
```diff
@@ -921,7 +921,7 @@ def incompatible_conditional_function_def(self, defn: FuncDef) -> None:
     def cannot_instantiate_abstract_class(self, class_name: str,
                                           abstract_attributes: List[str],
                                           context: Context) -> None:
-        attrs = format_string_list("'%s'" % a for a in abstract_attributes)
+        attrs = format_string_list(["'%s'" % a for a in abstract_attributes])
```
gvanrossum: Same here.

mypy/report.py
```diff
@@ -419,7 +419,8 @@ def __init__(self, reports: Reports, output_dir: str) -> None:
     # XML doesn't like control characters, but they are sometimes
     # legal in source code (e.g. comments, string literals).
     # Tabs (#x09) are allowed in XML content.
-    control_fixer = str.maketrans(''.join(chr(i) for i in range(32) if i != 9), '?' * 31)
+    control_fixer = str.maketrans(
+        ''.join(chr(i) for i in range(32) if i != 9), '?' * 31)  # type: Final
```
gvanrossum: How does this silence the warning in mypyc? Does `Final` make it compute the value at compile time?

mypy/stubdoc.py
```diff
@@ -24,7 +24,7 @@
 class ArgSig:
     """Signature info for a single argument."""
 
-    _TYPE_RE = re.compile(r'^[a-zA-Z_][\w\[\], ]*(\.[a-zA-Z_][\w\[\], ]*)*$')
+    _TYPE_RE = re.compile(r'^[a-zA-Z_][\w\[\], ]*(\.[a-zA-Z_][\w\[\], ]*)*$')  # type: Final
```
gvanrossum: Just curious, what was the warning here?

mypy/report.py
msullivan: `Final` makes it compute the value once at module load time, like expected, instead of every time the object is instantiated.

For annoying cpython internals reasons, there is (as far as we have been able to find) no good way to accurately and efficiently reproduce the interactions between instance variables and class variables (where the class variable is read if the instance variable is missing). So what we do is we treat the value in a class definition as a *default* that gets evaluated and placed in the instance when the instance is created, and we warn if it is something nontrivial.

mypy/stubdoc.py
msullivan: "Unsupported default attribute value", like for the other one

PR #6640, opened 2019-04-06
Add tests for new semantic analyzer on older Python

ilevkivskyi:
Hopefully, some tests will fail

### DIFF
--- a/mypy/server/aststripnew.py
+++ b/mypy/server/aststripnew.py
@@ -10,7 +10,7 @@ from mypy.nodes import (
     FuncDef, NameExpr, MemberExpr, RefExpr, MypyFile, ClassDef, AssignmentStmt,
     ImportFrom, CallExpr, Decorator, OverloadedFuncDef, Node, TupleExpr, ListExpr,
     SuperExpr, IndexExpr, ImportAll, ForStmt, Block, CLASSDEF_NO_INFO, TypeInfo,
-    StarExpr, Var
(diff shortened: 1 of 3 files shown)


ilevkivskyi:
The situation here is hilarious: there was a bug caused by Python "late binding" but masked by order preserving dictionaries in Python 3.6+.

ilevkivskyi:
Unless there are some objections, I will merge this soon (this blocks some things internally).

mypy/server/aststripnew.py
```diff
@@ -87,7 +87,9 @@ def prepare_implicit_var_patches(self, node: ClassDef) -> None:
             if isinstance(sym.node, Var) and sym.implicit:
                 explicit_self_type = sym.node.explicit_self_type
 
-                def patch() -> None:
+                # These arguments should not be passed, we just want to capture
+                # the names in closure at current iteration in the for-loop.
+                def patch(name: str = name, sym: SymbolTableNode = sym) -> None:
```
msullivan: This is tragically a bug under mypyc, which doesn't properly bind default variables at function creation time (https://github.com/mypyc/mypyc/issues/336).

There is a bad workaround for this, which is to have another nested function take them as arguments and have them get bound that way. I thought that I had needed to do this somewhere in mypy but I can't find it.

(As an aside: I noticed this because mypyc generates a warning on it, and it is now the only warning, so I actually noticed it.)

PR #6645, opened 2019-04-08
Support decorated constructors

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/1706
Fixes https://github.com/python/mypy/issues/5398

Previously, decorated constructors either resulted in silent `Any` types for class objects (in case of a decorated `__init__()`) or incorrect signatures (in case of a decorated `__new__()`).

This PR adds support for decorated constructors, implementation is straightforward.

This PR supersedes https://github.com/python/mypy/pull/5798.

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -817,6 +817,11 @@ def type_object_type(info: TypeInfo, builtin_type: Callable[[str], Instance]) ->
 
 
 def is_valid_constructor(n: Optional[SymbolNode]) -> bool:
+    """Does this node represents a valid constructor method?
+
+    This includes nomral functions, overloaded functions, and decorators
+    that return a callable type.
+    """
     if isinstance(n, FuncBase):
         return True
     if isinstance(n, Decorator):
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -813,14 +813,13 @@ def type_object_type(info: TypeInfo, builtin_type: Callable[[str], Instance]) ->
     else:
         assert isinstance(method.type, FunctionLike)  # is_valid_constructor() ensures this
         t = method.type
-    signature = bind_self(t)
-    return type_object_type_from_function(signature, info, method.info, fallback)
+    return type_object_type_from_function(t, info, method.info, fallback)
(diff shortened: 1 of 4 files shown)


gvanrossum:
I'm sorry, I don't think I can review this.

mypy/checkmember.py
```diff
@@ -797,17 +804,35 @@ def type_object_type(info: TypeInfo, builtin_type: Callable[[str], Instance]) ->
...
+        t = method.type
+    signature = bind_self(t)
+    return type_object_type_from_function(signature, info, method.info, fallback)
+
 
+def is_valid_constructor(n: Optional[SymbolNode]) -> bool:
+    """Does this node represents a valid constructor method?
 
-def type_object_type_from_function(init_or_new: FuncBase,
+    This includes nomral functions, overloaded functions, and decorators
```
msullivan: typo - "nomral"

mypy/checkmember.py
```diff
@@ -797,17 +804,35 @@ def type_object_type(info: TypeInfo, builtin_type: Callable[[str], Instance]) ->
...
+        method = init_method.node
     # Construct callable type based on signature of __init__. Adjust
     # return type and insert type arguments.
-    return type_object_type_from_function(method, info, fallback)
+    if isinstance(method, FuncBase):
+        t = function_type(method, fallback)
+    else:
+        assert isinstance(method.type, FunctionLike)  # is_valid_constructor() ensures this
+        t = method.type
+    signature = bind_self(t)
```
msullivan: Any particular reason for moving this `bind_self` out of `type_object_type_from_function`? I don't have any objection, I'm just wondering if there is a technical reason I'm missing or you just like it more.

ilevkivskyi: No particular reason. I think I started refactoring by pulling `function_type()` out of there, and pulled whole line out of there. But looking again at it I think it makes sense to keep bind and map together, so I will put it back.

ilevkivskyi:
> How much damage does it cause for internal codebases?

Not so much actually, around 60 errors S+C.

PR #6648, opened 2019-04-09
Expression-scoped ignores in Python 3.8.

### LINKED ISSUE #1032: Placement of # type: ignore on multi-line expression is iffy
Consider (this stands in for a longer example :-):

```
vars = {"table": "Users"}
sql = dict(vars,
           cols="username")
```

This currently gives an error due to #984. So I want to add `# type: ignore`. Unfortunately this doesn't work:

```
sql = dict(vars,
           cols="username")  # type: ignore
```

Instead I must use this:

```
sql = dict(vars,  # type: ignore
           cols="username")
```

which is a little unintuitive.

brandtbucher:
Fixes #1032. On Python 3.8, we use `end_lineno` information to scope `# type: ignore` comments to enclosing expressions. Auto-formatter users, rejoice!

The `--warn-unused-ignores` semantics should be clear from the test cases. Basically, ignores in inner scopes take precedence over ignores in enclosing scopes, and the first of multiple ignores in a scope "wins".

A few notes:
 - ~This changes the type of `Errors.ignored_lines` from a `set` to a `dict`, which maps ignored lines to the line of the comment they originated from. This structure is necessary in order to support `--warn-unused-ignores`.~
 - This adds a new slot, `Context.end_line`. It defaults to `None` in the absence of an `end_lineno` or if `Context` is not an expression.
 - `ErrorInfo.origin` now has a third member, representing the end line of the error context. It is used to determine the range of lines to search for `# type: ignore` comments. If unavailable, it defaults to the same value as the origin line. 
 - Because this uses 3.8-only AST features, a new `check-38.test` file has been created, and is hidden behind a version guard in `testcheck.py`.

### DIFF
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -54,8 +54,9 @@ class ErrorInfo:
     # Only report this particular messages once per program.
     only_once = False
 
-    # Actual origin of the error message as tuple (path, line number)
-    origin = None  # type: Tuple[str, int]
+    # Actual origin of the error message as tuple (path, line number, end line number)
+    # If end line number is unknown, use line number.
+    origin = None  # type: Tuple[str, int, int]
 
     # Fine-grained incremental target where this was reported
     target = None  # type: Optional[str]
@@ -72,7 +73,7 @@ class ErrorInfo:
                  message: str,
                  blocker: bool,
                  only_once: bool,
-                 origin: Optional[Tuple[str, int]] = None,
+                 origin: Optional[Tuple[str, int, int]] = None,
                  target: Optional[str] = None) -> None:
         self.import_ctx = import_ctx
         self.file = file
@@ -85,7 +86,7 @@ class ErrorInfo:
         self.message = message
         self.blocker = blocker
         self.only_once = only_once
-        self.origin = origin or (file, line)
+        self.origin = origin or (file, line, line)
         self.target = target
 
 
@@ -233,7 +234,8 @@ class Errors:
                file: Optional[str] = None,
                only_once: bool = False,
                origin_line: Optional[int] = None,
-               offset: int = 0) -> None:
+               offset: int = 0,
+               end_line: Optional[int] = None) -> None:
         """Report message at the given line using the current error context.
 
         Args:
@@ -244,6 +246,7 @@ class Errors:
             file: if non-None, override current file as context
             only_once: if True, only report this exact message once per build
             origin_line: if non-None, override current context as origin
+            end_line: if non-None, override current context as end
         """
         if self.scope:
             type = self.scope.current_type_name()
@@ -260,10 +263,17 @@ class Errors:
             file = self.file
         if offset:
             message = " " * offset + message
+
+        if origin_line is None:
+            origin_line = line
+
+        if end_line is None:
+            end_line = origin_line
+
         info = ErrorInfo(self.import_context(), file, self.current_module(), type,
                          function, line, column, severity, message,
                          blocker, only_once,
-                         origin=(self.file, origin_line) if origin_line else None,
+                         origin=(self.file, origin_line, end_line),
                          target=self.current_target())
         self.add_error_info(info)
 
@@ -274,12 +284,17 @@ class Errors:
         self.error_info_map[file].append(info)
 
     def add_error_info(self, info: ErrorInfo) -> None:
-        file, line = info.origin
+        file, line, end_line = info.origin
         if not info.blocker:  # Blockers cannot be ignored
-            if file in self.ignored_lines and line in self.ignored_lines[file]:
-                # Annotation requests us to ignore all errors on this line.
-                self.used_ignored_lines[file].add(line)
-                return
+            if file in self.ignored_lines:
+                # Check each line in this context for "type: ignore" comments.
+                # For anything other than Python 3.8 expressions, line == end_line,
+                # so we only loop once.
+                for scope_line in range(line, end_line + 1):
+                    if scope_line in self.ignored_lines[file]:
+                        # Annotation requests us to ignore all errors on this line.
+                        self.used_ignored_lines[file].add(scope_line)
+                        return
             if file in self.ignored_files:
                 return
         if info.only_once:
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -113,9 +113,8 @@ class Errors:
     # Path to current file.
     file = ''  # type: str
 
-    # Ignore errors on these lines of each file. Keys are modules, values are mappings from:
-    # ignored line -> line of "type: ignore" comment it is scoped to within that module.
-    ignored_lines = None  # type: Dict[str, Dict[int, int]]
+    # Ignore errors on these lines of each file.
+    ignored_lines = None  # type: Dict[str, Set[int]]
 
     # Lines on which an error was actually ignored.
     used_ignored_lines = None  # type: Dict[str, Set[int]]
@@ -200,7 +199,7 @@ class Errors:
         self.scope = scope
 
     def set_file_ignored_lines(self, file: str,
-                               ignored_lines: Dict[int, int],
+                               ignored_lines: Set[int],
                                ignore_all: bool = False) -> None:
         self.ignored_lines[file] = ignored_lines
         if ignore_all:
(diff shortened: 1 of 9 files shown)


test-data/unit/check-38.test
```diff
@@ -0,0 +1,21 @@
+[case testScopedIgnore]
+def f(a: int): ...
+f(
+    'BAD'
+)  # type: ignore
```
ilevkivskyi: I would add few more tests with expressions that are not function calls. For example list and dictionary displays, operators (like + and *), etc.

mypy/treetransform.py
```diff
@@ -60,7 +60,7 @@ def __init__(self) -> None:
     def visit_mypy_file(self, node: MypyFile) -> MypyFile:
         # NOTE: The 'names' and 'imports' instance variables will be empty!
         new = MypyFile(self.statements(node.defs), [], node.is_bom,
-                       ignored_lines=set(node.ignored_lines))
+                       ignored_lines=node.ignored_lines)
```
ilevkivskyi: I think this should make a copy instead of passing the original reference.

mypy/test/testcheck.py
```diff
@@ -85,6 +85,9 @@
     'check-newsemanal.test',
 ]
 
+if (3, 8) < sys.version_info:
```
ilevkivskyi: I would flip this.
```suggestion
if  sys.version_info >= (3, 8):
```
Also please add explanatory comment.

mypy/errors.py
```diff
@@ -114,7 +114,7 @@ class Errors:
     file = ''  # type: str
 
     # Ignore errors on these lines of each file.
-    ignored_lines = None  # type: Dict[str, Set[int]]
+    ignored_lines = None  # type: Dict[str, Dict[int, int]]
```
ilevkivskyi: Please document the meaning of this dictionary.

mypy/errors.py
```diff
@@ -300,7 +300,7 @@ def clear_errors_in_targets(self, path: str, targets: Set[str]) -> None:
             self.error_info_map[path] = new_errors
 
     def generate_unused_ignore_notes(self, file: str) -> None:
-        ignored_lines = self.ignored_lines[file]
+        ignored_lines = set(self.ignored_lines[file].values())
```
ilevkivskyi: Add a comment reminding that this is a set of actual `# type: ignore`s in the file.

mypy/nodes.py
```diff
@@ -227,7 +227,7 @@ class MypyFile(SymbolNode):
     # All import nodes within the file (also ones within functions etc.)
     imports = None  # type: List[ImportBase]
     # Lines to ignore when checking
-    ignored_lines = None  # type: Set[int]
+    ignored_lines = None  # type: Dict[int, int]
```
ilevkivskyi: Please add a documenting comment also here.

mypy/fastparse.py
```diff
@@ -251,7 +251,7 @@ def __init__(self,
         self.is_stub = is_stub
         self.errors = errors
 
-        self.extra_type_ignores = []  # type: List[int]
+        self.type_ignores = {}  # type: Dict[int, int]
```
ilevkivskyi: Also a comment here would be helpful (the one reading the code should not guess what is the direction of the mapping).

mypy/fastparse.py
```diff
@@ -271,7 +271,20 @@ def visit(self, node: Optional[AST]) -> Any:
...
+        return result
+
+    def scope_ignores(self, node: ast3.expr) -> None:
+        end_lineno = getattr(node, "end_lineno", None)
+        if end_lineno is None:
+            return
+        node_lines = range(node.lineno, end_lineno + 1)
+        for line in node_lines:
+            if line in self.type_ignores.values():
+                self.type_ignores = {**dict.fromkeys(node_lines, line), **self.type_ignores}
```
ilevkivskyi: @msullivan does mypyc support this?

mypy/fastparse.py
```diff
@@ -271,7 +271,20 @@ def visit(self, node: Optional[AST]) -> Any:
...
+            self.scope_ignores(node)
+        return result
+
+    def scope_ignores(self, node: ast3.expr) -> None:
+        end_lineno = getattr(node, "end_lineno", None)
+        if end_lineno is None:
+            return
+        node_lines = range(node.lineno, end_lineno + 1)
+        for line in node_lines:
+            if line in self.type_ignores.values():
```
ilevkivskyi: I am not sure how efficient is this. I think it would be better to assign `mod.type_ignores` to an attribute on `self` and use it for this check instead of recomputing this same set every time.

Also please add a comment with an example explaining what is going on here.

ilevkivskyi:
Please let me know when this is ready for another review.

mypy/fastparse.py
brandtbucher: This has been refactored not to use unpacking, either way!

brandtbucher:
Thanks for the feedback @ilevkivskyi! I've gone ahead and addressed your concerns, and am ready for another review.

I've chosen to maintain a set of source comment lines as you suggested, in addition to the mapping of ignore scopes. One nice benefit is, with a change in the order of traversal, we can now avoid expensive dictionary copies at each ignored expression node.

mypy/fastparse.py
```diff
@@ -394,13 +414,13 @@ def translate_module_id(self, id: str) -> str:
         return id
 
     def visit_Module(self, mod: ast3.Module) -> MypyFile:
+        self.type_ignore_comments = {ti.lineno for ti in mod.type_ignores}
         body = self.fix_function_overloads(self.translate_stmt_list(mod.body))
-        ignores = [ti.lineno for ti in mod.type_ignores]
-        ignores.extend(self.extra_type_ignores)
+        self.type_ignores.update({line: line for line in self.type_ignore_comments})
```
ilevkivskyi: I think it is less error-prone to reset `type_ignores` when visiting each file. I.e. here I would just use `self.type_ignores = {line: line for ...}`.

test-data/unit/check-38.test
```diff
@@ -0,0 +1,82 @@
...
+)  # type: ignore  # N: unused 'type: ignore' comment
+
+[case testNestedScopedIgnores3]
+# flags: --warn-unused-ignores
+def f(a: int) -> int: ...
+f(  # type: ignore  # N: unused 'type: ignore' comment
+    f(
+        'BAD'
+    )  # type: ignore
+)  # type: ignore  # N: unused 'type: ignore' comment
```
ilevkivskyi: I actually meant more tests without `--warn-unused-ignores`. I would keep only couple dedicated tests for `--warn-unused-ignores`, and remove it from the rest of the test cases.

mypy/fastparse.py
```diff
@@ -271,8 +276,23 @@ def visit(self, node: Optional[AST]) -> Any:
...
+    def scope_ignores(self, node: ast3.expr) -> None:
+        end_lineno = getattr(node, "end_lineno", None)
+        if end_lineno is None:
+            return
+        node_lines = range(node.lineno, end_lineno + 1)
+        # Check to see if this expression overlaps with any "type: ignore" comments.
+        # If so, take the first one and grow its scope to cover the whole node:
+        for line in node_lines:
+            if line in self.type_ignore_comments:
+                self.type_ignores.update(dict.fromkeys(node_lines, line))
```
ilevkivskyi: It seems to me now there is a flaw in this logic. Imagine this test case:
```python
def func(x: int) -> None: ...
lst = [func(1),
       func('no'),  # type: ignore
       func('yes'),
       func(2)]
```
I think only the error for `func('no')` should be ignored, while we still should show the error for `func('yes')`.
It looks like the actual solution should be a bit more complex.

In some sense the logic should be reversed, for every error you should check whether there is a `# type: ignore` comment in the span of error context expression for that particular error.

I am not sure what is the best way, but maybe we would need to copy the end position for every expression in the generated mypy AST.

brandtbucher: I agree now, when you put it like that. Another clear example would be:

```py
def f(a: int): ...
[f("error"), f(
    "ignored"  # type: ignore
    ), f("error")]
```

...which isn’t really possible by predefining the line scope of each ignore. Really, it’s expressions that have scopes, not the ignores. This fix is exactly backwards.

I think that having the expression context grow an `end_lineno` slot and checking for ignores when reporting is both correct and more efficient. I’ll back out the current implementation and shift to that approach, if that seems right to you @ilevkivskyi!

ilevkivskyi: OK, go ahead, and let me know when the PR is ready for another review.

brandtbucher:
I'm ready for a new review @ilevkivskyi.

I've updated my implementation to use a new `Context.end_line` slot. It's much cleaner than my first attempt, and works as expected. I've also updated the tests, as you requested.

brandtbucher:
Thanks for the guidance @ilevkivskyi! Happy to contribute.

Issue #4134, opened 2017-10-19
Constrained TypeVar fails to narrow on left hand side of conditional expression


alanhdu:
When I run `mypy --python-version 3.6` on the following code:

```python
from typing import AnyStr
def f(x: AnyStr) -> bytes:
    return x.encode("ascii") if isinstance(x, str) else x
```

I get:

```
test.py:3: error: "bytes" has no attribute "encode"; maybe "decode"?
```

But as far as I can tell, this is backwards! `x.encode("ascii")` is only called when `isinstance(x, str)`, not in the `else` case (where x is a `bytes` object).

This is with the latest `mypy==0.530`

JelleZijlstra:
I think this is because you're using `AnyStr` incorrectly. As a TypeVar, it really makes sense only when used multiple times in a signature (or in a few other cases not relevant here). I believe we're going to introduce a warning for incorrect usages like yours, although I can't find the relevant issue now.

Your code should have been written as:
```
from typing import Union
def f(x: Union[bytes, str]) -> bytes:
    return x.encode("ascii") if isinstance(x, str) else x
```
which passes mypy with no errors for me.

I don't know why your code gives an error talking about `bytes` though, and maybe there's a real bug there. I think it's related to the ternary expression, because rewriting the code to use an `if-else` block removes the error.

alanhdu:
Ah... I see. Yup, `Union[bytes, str]` works perfectly!

I'll leave this open because I think the inference that `x` is a `bytes` object probably represents a bug somewhere, but if you disagree then feel free to close it!

gvanrossum:
Hm, I do think there's a bug somewhere, since this gives no errors -- the key difference being that it uses an `if` statement instead of an `if` expression:
```py
def f(xs: List[AnyStr]) -> List[bytes]:
    res = []
    for x in xs:
        if isinstance(x, str):
            res.append(x.encode('ascii'))
        else:
            res.append(x)
    return res
```

ilevkivskyi:
I don't think there is any mypy bug here, although this may be a usability problem. What is going on here consists of two parts:
1) For constrained type variables (like ``AnyStr``) the function body is checked for all constraints (``str`` and ``bytes`` in this case).
2) For ternary expression, mypy can't mark first part as "unreachable", but it doesn't perform any type narrowing either if the types are non-overlapping, this can be seen form following:
```python
x: int
y = reveal_type(x) if isinstance(x, str) else ''  # Revealed type is 'int'
```

[...]

PR #6649, opened 2019-04-09
document ternary/conditional type inference limitations

petergaultney:
As per the discussion in #4134 , it seems worth documenting that conditional expressions don't always work smoothly with type inference.  This PR makes sure to include the word ternary and conditional so that page/doc searches will turn up an explanation for this behavior in case anyone is searching for it.

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -339,6 +339,24 @@ style anyway).  We can write the above code without a cast by using
            g(o + 1)        # Okay; type of o is inferred as int here
            ...
 
+Note that MyPy cannot always narrow a type based on ``isinstance()``
+used within a ternary/conditional expression, e.g.:
+
+.. code-block:: python
+
+	def f(x: AnyStr, y: AnyStr) -> bytes:
+		x += y
+
+		r = x.encode('ascii') if isinstance(x, str) else x
+		# results in 'error: "bytes" has no attribute "encode"; maybe "decode"?'
+
+		if isinstance(x, str):
+			r = x.encode('ascii')  # whereas MyPy can narrow within an if statement
+		else:
+			r = x
+		...
+
+
 Type inference in mypy is designed to work well in common cases, to be
 predictable and to let the type checker give useful error
 messages. More powerful type inference strategies often have complex
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -339,23 +339,21 @@ style anyway).  We can write the above code without a cast by using
            g(o + 1)        # Okay; type of o is inferred as int here
            ...
 
-Note that MyPy cannot always narrow a type based on ``isinstance()``
+Note that Mypy cannot always narrow a type based on ``isinstance()``
 used within a ternary/conditional expression, e.g.:
 
 .. code-block:: python
 
-	def f(x: AnyStr, y: AnyStr) -> bytes:
-		x += y
-
+	def f(x: AnyStr) -> bytes:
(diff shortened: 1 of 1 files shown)


docs/source/common_issues.rst
```diff
@@ -339,6 +339,24 @@ style anyway).  We can write the above code without a cast by using
            g(o + 1)        # Okay; type of o is inferred as int here
            ...
 
+Note that MyPy cannot always narrow a type based on ``isinstance()``
```
ilevkivskyi:
```suggestion
Note that Mypy cannot always narrow a type based on ``isinstance()``
```

docs/source/common_issues.rst
```diff
@@ -339,6 +339,24 @@ style anyway).  We can write the above code without a cast by using
...
+.. code-block:: python
+
+	def f(x: AnyStr, y: AnyStr) -> bytes:
+		x += y
+
+		r = x.encode('ascii') if isinstance(x, str) else x
+		# results in 'error: "bytes" has no attribute "encode"; maybe "decode"?'
+
+		if isinstance(x, str):
+			r = x.encode('ascii')  # whereas MyPy can narrow within an if statement
```
ilevkivskyi:
```suggestion
			r = x.encode('ascii')  # whereas Mypy can narrow within an if statement
```

docs/source/common_issues.rst
```diff
@@ -339,6 +339,24 @@ style anyway).  We can write the above code without a cast by using
...
+		else:
+			r = x
+		...
+
+
```
ilevkivskyi: Only keep one empty line.

docs/source/common_issues.rst
```diff
@@ -339,6 +339,24 @@ style anyway).  We can write the above code without a cast by using
            g(o + 1)        # Okay; type of o is inferred as int here
            ...
 
+Note that MyPy cannot always narrow a type based on ``isinstance()``
+used within a ternary/conditional expression, e.g.:
+
+.. code-block:: python
+
+	def f(x: AnyStr, y: AnyStr) -> bytes:
```
ilevkivskyi: The example with `AnyStr` is not the best one, since it combines two tricky points:
* the one you are explaining
* the fact that functions generic in variables with restrictions are type checked multiple times.

I would rather use a single argument function with a `Union[str, bytes]` argument.

petergaultney: Actually, if I try to simplify this down to ```def f(x: Union[str, bytes]) -> bytes:
    r = x.encode('ascii') if isinstance(x, str) else x
    # results in 'error: "bytes" has no attribute "encode"; maybe "decode"?'

    if isinstance(x, str):
        r = x.encode('ascii')  # whereas Mypy can narrow within an if statement
    else:
        r = x
    return r```

Mypy doesn't error at all, which is quite interesting. Whereas if I use AnyStr on that same argument, there is an error.

It may be that I have oversimplified the explanation for what is going on?

ilevkivskyi: To clarify, mypy _can_ restrict types using `isinstance()` in ternary expressions, but it cannot consider sub-expressions of ternary expression _unreachable_. Here is even simpler example:
```python
x: bytes

if isinstance(x, str):
    r = x.encode()  # OK, this is unreachable
else:
    r = x

r = x.encode() if isinstance(x, str) else x  # Error here
```

JukkaL:
Closing since there has been no activity in a long time. Feel free to repopen if you want to continue working on this.

PR #6655, opened 2019-04-10
Fix __getattr__ and operators related infinite recursion

### LINKED ISSUE #6579: segmentation fault on indexing object with __getattr__

Bug.
```python
class C:
    def __getattr__(self, name: str) -> 'C':
        ...


def f(v: C) -> None:
    v[0]
```
segfault
not crashing
python 3.6.7 on Ubuntu 18.04.2 x86_64
mypy 0.670 installed march 20 2019 via pip

yes
0.680+dev.4e0a1583aeb00b248e187054980771f1897a1d31
none

onlined:
Fixes #6579.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -704,8 +704,8 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         elif isinstance(callee, UnionType):
             return self.check_union_call(callee, args, arg_kinds, arg_names, context, arg_messages)
         elif isinstance(callee, Instance):
-            call_function = analyze_member_access('__call__', callee, context,
-                                                  False, False, True, self.msg,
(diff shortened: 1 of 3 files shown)


mypy/checkexpr.py
```diff
@@ -705,7 +705,7 @@ def check_call(self,
             return self.check_union_call(callee, args, arg_kinds, arg_names, context, arg_messages)
         elif isinstance(callee, Instance):
             call_function = analyze_member_access('__call__', callee, context,
-                                                  False, False, False, self.msg,
+                                                  False, False, True, self.msg,
```
JelleZijlstra: Not new to your diff, but it would be nice if we could use keyword args here so it's clear what all these bools mean.

onlined: Did it, thanks!

PR #6657, opened 2019-04-10
Fix memprofile crashes and make it work with mypyc

msullivan:
memprofile crashes when trying to call isinstance on FakeInfos, since isinstance 
will sometimes try to access `__class__`. Make `__class__` on FakeInfo work.

Also use `get_class_descriptors` instead of traversing `__slots__` directly so that
mypyc objects can be traversed.
(Though some fixes are needed to mypyc for this to work that I haven't pushed yet.)


msullivan:
Yeah, looks good.

PR #6664, opened 2019-04-11
Allow disabling active config files.

brandtbucher:
With this change, passing `--config-file=` from the command line disables any active config files.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -57,6 +57,9 @@ Config file
     Settings override mypy's built-in defaults and command line flags
     can override settings.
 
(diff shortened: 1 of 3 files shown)


gvanrossum:
Why not use /dev/null?

brandtbucher:
```sh
$ mypy --config-file /dev/null .
/dev/null: No [mypy] section in config file
```

gvanrossum:
OK, this sounds like a good idea. But yeah, please update the docs. Also please add a unittest to test-data/unit/check-flags.test, e.g. show that options in mypy.ini are ignored with `--config-file=`.

brandtbucher:
Ok, added documentation and testing!

Issue #6531, opened 2019-03-10
Support ellipsis default arguments in runtime functions missing a body


Michael0x2a:
Mypy (and PEP 484) currently allows using ellipsis as the default argument for functions defined inside stubs:

    # Legal inside *.pyi files, but illegal inside *.py files -- we get the following error:
    # "Incompatible default for argument "some_arg" (default has type "ellipsis", argument has type "int")"
    def foo(some_arg: int = ...) -> None: pass

I propose generalizing the way mypy handles ellipsis default arguments so that the above becomes permitted for *both* stub files and regular files as long as the body of the function is either empty or contains just a `raise SomeException(...)` statement.

The main reason I want this is because I was trying to define a custom Callback protocol where one of the arguments accepted a mutable argument, like so:

    class MyCallback(Protocol):
        def __call__(self, foo: List[int] = ???) -> None: ...

Unfortunately, there doesn't seem to be an entirely clean way of supporting this:

1. I could do `foo: List[int] = []`, but then I get warnings from my IDE and linter about having a mutable default argument.
2. I could do `foo: Optional[List[int]] = None`, but I don't actually want to allow `None` arguments.
3. I could do `foo: List[int] = bogus`, where `bogus` has type `Any`, but that causes another error when using the `--disallow-any-explicit` flag.

You run into similar issues when trying to define overloads variants or abstract methods -- basically, function definitions that effectively exist only at "type-check time". 

The restriction that you can only use ellipsis default arguments when the function body is missing/just throws an exception should hopefully help prevent the ellipsis from appearing in unexpected places and causing runtime errors.

I'd be willing to implement this myself (especially since this seems like a pretty low-priority feature, all things considered), but want to make sure the idea seems like a reasonable one first before investing the effort.

ilevkivskyi:
I like this idea, this can be useful for both protocols and ABCs.

PR #6665, opened 2019-04-11
Allow ellipsis default args in "trivial" non-stub functions

Michael0x2a:
Resolves https://github.com/python/mypy/issues/6531

This diff modifies the checker to allow default arguments to be used in function definitions in non-stub files as long as the body of that function is either empty or contains just a `raise NotImplementedError`. (This is after ignoring docstrings.)

For rationale on why this is useful, see the issue linked above.

One note on function bodies containing only a `raise Blah`. I was worried that letting mypy ignore the ellipsis when the function raises *any* Exception would lead to unsafe behavior -- I didn't want mypy to start allowing code that looks like this:

    def halt(self, reason: str = ...) -> NoReturn:
        raise MyCustomError("Fatal error: " + reason, self.line, self.current_context)

Whitelisting only NotImplementedError seemed like a good enough compromise -- it's unlikely users will use the arguments in non-trivial ways when raising that particular exception. It should also be easy enough to relax this restriction in the future if necessary.

### DIFF
--- a/docs/source/stubs.rst
+++ b/docs/source/stubs.rst
@@ -62,20 +62,81 @@ Stub file syntax
 
 Stub files are written in normal Python 3 syntax, but generally
 leaving out runtime logic like variable initializers, function bodies,
-and default arguments, or replacing them with ellipses.
+and default arguments.
 
-In this example, each ellipsis ``...`` is literally written in the
+If it is not possible to completely leave out some piece of runtime
+logic, the recommended convention is to replace or elide them with ellipsis
+expressions (``...``). Each ellipsis below is literally written in the
 stub file as three dots:
 
 .. code-block:: python
 
+    # Variables with annotations do not need to be assigned a value.
+    # So by convention, we omit them in the stub file.
     x: int
 
-    def afunc(code: str) -> int: ...
-    def afunc(a: int, b: int = ...) -> int: ...
+    # Function bodies cannot be completely removed. By convention,
+    # we replace them with `...` instead of the `pass` statement.
+    def func_1(code: str) -> int: ...
+
+    # We can do the same with default arguments.
+    def func_2(a: int, b: int = ...) -> int: ...
 
 .. note::
 
     The ellipsis ``...`` is also used with a different meaning in
     :ref:`callable types <callable-types>` and :ref:`tuple types
     <tuple-types>`.
+
+.. note::
+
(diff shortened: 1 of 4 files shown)


mypy/checker.py
```diff
@@ -1012,7 +1015,29 @@ def is_trivial_body(self, block: Block) -> bool:
             return True
         elif len(body) > 1:
             return False
+
         stmt = body[0]
+
+        # Bodies that just raise "NotImplementedError()" count as trivial.
```
msullivan: Add a doc string for this function, maybe by moving some of the inline comments to it?

JukkaL:
Maybe mention this somewhere in the mypy docs?

test-data/unit/fixtures/exception.pyi
```diff
@@ -13,5 +13,10 @@ class unicode: pass
...
 class ellipsis: pass
 
-class BaseException: pass
+# Note: this is a slight simplification. In Python 2, the inheritance hierarchy
+# is actually Exception -> StandardError -> RuntimeError -> ...
+class BaseException:
+    def __init__(self, *args: object) -> None: ...
 class Exception(BaseException): pass
+class RuntimeError(Exception): pass
+class NotImplementedError(RuntimeError): pass
```
ilevkivskyi: No new line at the end of file.

Michael0x2a:
Sorry, returning to this only now. I added docs.

I did find writing them to be a bit difficult though: I wasn't sure exactly where to add the docs or how to best present the new info. I ended up shoving them into the page about stubs for now.

So, I'm a little leery of just landing this now. If it's not too much bother, would somebody mind doing one last sanity check over the docs I added?

docs/source/stubs.rst
```diff
@@ -62,20 +62,81 @@ Stub file syntax
...
+
+The recommended style is to use ellipses to do so, just like in
+stub files. It is also considered stylistically acceptable to
+throw a ``NotImplementedError`` in cases where the user of the
+code may accidentally call functions with no actual logic.
+
+You can also elide default arguments as long as the function body
+also contains no runtime logic: the function body only contains
+a single ellipsis, the pass statement, or a ``raise NotImplementedError()``.
+It is also ok for the function body to contain a docstring.
```
JelleZijlstra:
```suggestion
It is also acceptable for the function body to contain a docstring.
```
"ok" is a bit too informal

Issue #6667, opened 2019-04-12
Semantic analyzers do not correctly handle qualified type aliases


Michael0x2a:
Consider the following program:

    # test.py
    class Outer:
        class Inner: pass

    Alias = Outer
    x: Alias.Inner
    reveal_type(x)

Mypy thinks that the type of `x` is `test.Outer` instead of `test.Outer.Inner`. We get this behavior with both the new and old semantic analyzers.

---

It seems the root cause has to do with the [`lookup_qualified` function](https://github.com/python/mypy/blob/master/mypy/semanal.py#L3466) -- when the loop encounters a SymbolNode of type TypeAlias, it effectively stops processing the `parts` variable, ignoring any subsequent information in that list. So, we drop everything to the right of the first type alias. 

There's actually a TODO about this in the code, though it points out the issue for only `Var` and `FuncDef`. The new semantic analyzer's `lookup_qualified` uses the same algorithm + has the same TODO.

In order to fix this correctly, we'll maybe need to either transfer the unhandled parts back to the caller, or move the [type alias expansion logic](https://github.com/python/mypy/blob/master/mypy/typeanal.py#L243) from `typeanal` into this function? I briefly experimented with both approaches, but they both seemed very clunky.

Probably the right thing to do is to just fix this as a part of the long-awaited lookup-functions-refactor (https://github.com/python/mypy/issues/4157).

(Hopefully this isn't a dupe -- I'm mostly filing this so I can link to it in a PR I'm preparing.)

ilevkivskyi:
This is not a duplicate, and this is indeed about the [TODO](https://github.com/python/mypy/blob/master/mypy/semanal.py#L3508) you mention. I am not sure if there is any "graceful" fix for this. Ideally, the fix should take care of all currently missing node cases (including `FuncDef`, `Var`, `TypeAlias`, maybe also `PlaceholderNode` if necessary).

I would also defer this until the fix of #4157

ilevkivskyi:
Interestingly on master this now gives:
```
test.py:5: error: Name 'Alias.Inner' is not defined
test.py:6: note: Revealed type is 'Any'
```
Which is probably a bit less bad.

ilevkivskyi:
This was fixed by https://github.com/python/mypy/pull/8014

PR #6668, opened 2019-04-12
Add basic support for enum literals

Michael0x2a:
This pull request adds in basic support for enum literals. Basically, you can construct them and use them as types, but not much else.

(I originally said I was going to wait until the new semantic analyzer work was finished, but I'm bored.)

Some notes:

1. I plan on submitting incremental and fine-grained-mode tests in a future PR.

2. I added two copies of each test -- one using the old semantic analyzer and one using the new one. Let me know if we prefer not doing this and if I should pick one over the other.

3. I wanted to add support for aliases to enums, but ran into some difficulty doing so. See https://github.com/python/mypy/issues/6667 for some analysis on the root cause.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3989,10 +3989,10 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         module namespace is ignored.
         """
         parts = name.split('.')
-        n = self.modules[parts[0]]
+        n = self.modules[parts[0]]  # type: Union[MypyFile, TypeInfo]
         for i in range(1, len(parts) - 1):
             next_sym = n.names[parts[i]]
-            assert isinstance(next_sym.node, MypyFile)
+            assert isinstance(next_sym.node, (MypyFile, TypeInfo))
             n = next_sym.node
         return n.names[parts[-1]]
 
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3989,10 +3989,10 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         module namespace is ignored.
         """
         parts = name.split('.')
-        n = self.modules[parts[0]]  # type: Union[MypyFile, TypeInfo]
+        n = self.modules[parts[0]]
         for i in range(1, len(parts) - 1):
             next_sym = n.names[parts[i]]
-            assert isinstance(next_sym.node, (MypyFile, TypeInfo))
+            assert isinstance(next_sym.node, MypyFile)
             n = next_sym.node
         return n.names[parts[-1]]
 
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -73,7 +73,7 @@ from mypy.nodes import implicit_module_attrs
 from mypy.typeanal import (
     TypeAnalyser, analyze_type_alias, no_subscript_builtin_alias,
     TypeVariableQuery, TypeVarList, remove_dups, has_any_from_unimported_type,
-    check_for_explicit_any
+    check_for_explicit_any, expand_type_alias
 )
 from mypy.exprtotype import expr_to_unanalyzed_type, TypeTranslationError
 from mypy.sametypes import is_same_type
@@ -3590,10 +3590,10 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
         module namespace is ignored.
         """
         parts = name.split('.')
-        n = self.modules[parts[0]]
+        n = self.modules[parts[0]]  # type: Union[MypyFile, TypeInfo]
         for i in range(1, len(parts) - 1):
             next_sym = n.names[parts[i]]
-            assert isinstance(next_sym.node, MypyFile)
+            assert isinstance(next_sym.node, (MypyFile, TypeInfo))
             n = next_sym.node
         return n.names[parts[-1]]
 
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3590,10 +3590,10 @@ class SemanticAnalyzerPass2(NodeVisitor[None],
         module namespace is ignored.
         """
         parts = name.split('.')
-        n = self.modules[parts[0]]  # type: Union[MypyFile, TypeInfo]
+        n = self.modules[parts[0]]
         for i in range(1, len(parts) - 1):
             next_sym = n.names[parts[i]]
-            assert isinstance(next_sym.node, (MypyFile, TypeInfo))
+            assert isinstance(next_sym.node, MypyFile)
             n = next_sym.node
         return n.names[parts[-1]]
 
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -73,7 +73,7 @@ from mypy.nodes import implicit_module_attrs
 from mypy.typeanal import (
     TypeAnalyser, analyze_type_alias, no_subscript_builtin_alias,
     TypeVariableQuery, TypeVarList, remove_dups, has_any_from_unimported_type,
-    check_for_explicit_any, expand_type_alias
+    check_for_explicit_any
 )
 from mypy.exprtotype import expr_to_unanalyzed_type, TypeTranslationError
 from mypy.sametypes import is_same_type
--- a/mypy/typeanal.py
+++ b/mypy/typeanal.py
@@ -397,24 +397,25 @@ class TypeAnalyser(SyntheticTypeVisitor[Type], TypeAnalyzerPluginInterface):
             return t
 
         # Option 3:
-        # Enum value. Note: Enum values are not real types, so we return
-        # RawExpressionType only when this function is being called by
-        # one of the Literal[...] handlers -- when `defining_literal` is True.
+        # Enum value. Note: we only want to return a LiteralType when
+        # we're using this enum value specifically within context of
+        # a "Literal[...]" type. So, if `defining_literal` is not set,
+        # we bail out early with an error.
         #
-        # It's unsafe to return RawExpressionType in any other case, since
-        # the type would leak out of the semantic analysis phase.
-        if isinstance(sym.node, Var) and sym.node.info and sym.node.info.is_enum:
+        # If, in the distant future, we decide to permit things like
+        # `def foo(x: Color.RED) -> None: ...`, we can remove that
+        # check entirely.
+        if isinstance(sym.node, Var) and not t.args and sym.node.info and sym.node.info.is_enum:
             value = sym.node.name()
             base_enum_short_name = sym.node.info.name()
-            base_enum_qualified_name = sym.node.info.fullname()
             if not defining_literal:
                 msg = message_registry.INVALID_TYPE_RAW_ENUM_VALUE.format(
                     base_enum_short_name, value)
                 self.fail(msg, t)
                 return AnyType(TypeOfAny.from_error)
-            return RawExpressionType(
-                literal_value=value,
-                base_type_name=base_enum_qualified_name,
+            return LiteralType(
+                value=value,
+                fallback=Instance(sym.node.info, [], line=t.line, column=t.column),
                 line=t.line,
                 column=t.column,
             )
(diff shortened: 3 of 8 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -3989,10 +3989,10 @@ def lookup_fully_qualified(self, name: str) -> SymbolTableNode:
         module namespace is ignored.
         """
         parts = name.split('.')
-        n = self.modules[parts[0]]
+        n = self.modules[parts[0]]  # type: Union[MypyFile, TypeInfo]
         for i in range(1, len(parts) - 1):
             next_sym = n.names[parts[i]]
-            assert isinstance(next_sym.node, MypyFile)
+            assert isinstance(next_sym.node, (MypyFile, TypeInfo))
```
ilevkivskyi: I think there is some misunderstanding/bug somewhere. This function should never be used for things like local lookup. It should be only used to construct a known type like `builtins.str`. For local lookup one needs to use `lookup_qualified()`. Why do you need this modification?

mypy/semanal.py
```diff
@@ -73,7 +73,7 @@
 from mypy.typeanal import (
     TypeAnalyser, analyze_type_alias, no_subscript_builtin_alias,
     TypeVariableQuery, TypeVarList, remove_dups, has_any_from_unimported_type,
-    check_for_explicit_any
+    check_for_explicit_any, expand_type_alias
```
ilevkivskyi: Unused import.

mypy/semanal.py
```diff
@@ -3590,10 +3590,10 @@ def lookup_fully_qualified(self, name: str) -> SymbolTableNode:
         module namespace is ignored.
         """
         parts = name.split('.')
-        n = self.modules[parts[0]]
+        n = self.modules[parts[0]]  # type: Union[MypyFile, TypeInfo]
         for i in range(1, len(parts) - 1):
             next_sym = n.names[parts[i]]
-            assert isinstance(next_sym.node, MypyFile)
+            assert isinstance(next_sym.node, (MypyFile, TypeInfo))
```
ilevkivskyi: Same question as for new analyzer here.

mypy/typeanal.py
```diff
@@ -245,11 +245,14 @@ def visit_unbound_type_nonoptional(self, t: UnboundType) -> Type:
                 all_vars = node.alias_tvars
                 target = node.target
                 an_args = self.anal_array(t.args)
-                return expand_type_alias(target, all_vars, an_args, self.fail, node.no_args, t)
+                out = expand_type_alias(target, all_vars, an_args, self.fail, node.no_args, t)
+                if 'RED' in str(t):
+                    print('...')
```
ilevkivskyi: Debug print?

mypy/typeanal.py
```diff
@@ -245,11 +245,14 @@ def visit_unbound_type_nonoptional(self, t: UnboundType) -> Type:
...
+                return out
```
ilevkivskyi: Debug variable?

test-data/unit/check-literal.test
```diff
@@ -2643,3 +2643,306 @@ c_wrap: Literal[4, c]  # E: Invalid type "__main__.c" \
...
+    def func(self) -> int: pass
+
+r: Literal[Color.RED]
+g: Literal[Color.GREEN]
+b: Literal[Color.BLUE]
+bad: Literal[Color]   # E: Parameter 1 of Literal[...] is invalid
+
+def expects_color(x: Color) -> None: pass
+def expects_red(x: Literal[Color.RED]) -> None: pass
+def bad_func(x: Color.RED) -> None: pass             # E: Invalid type: try using Literal[__main__.Color.RED] instead?
```
ilevkivskyi: I would use short name for errors here and below `Literal[Color.RED]`.

test-data/unit/check-literal.test
```diff
@@ -2643,3 +2643,306 @@ c_wrap: Literal[4, c]  # E: Invalid type "__main__.c" \
...
+[case testLiteralWithEnumsBasic]
+from typing_extensions import Literal
+from enum import Enum
+
+class Color(Enum):
+    RED = 1
+    GREEN = 2
+    BLUE = 3
+
+    def func(self) -> int: pass
```
ilevkivskyi: Maybe add a test that errors on `Literal[Color.func]`?

test-data/unit/check-literal.test
```diff
@@ -2643,3 +2643,306 @@ c_wrap: Literal[4, c]  # E: Invalid type "__main__.c" \
...
+expects_red(g)  # E: Argument 1 to "expects_red" has incompatible type "Literal[__main__.Color.GREEN]"; expected "Literal[__main__.Color.RED]"
+expects_red(b)  # E: Argument 1 to "expects_red" has incompatible type "Literal[__main__.Color.BLUE]"; expected "Literal[__main__.Color.RED]"
+
+reveal_type(expects_red)  # E: Revealed type is 'def (x: Literal[__main__.Color.RED])'
+reveal_type(r)            # E: Revealed type is 'Literal[__main__.Color.RED]'
+reveal_type(r.func())     # E: Revealed type is 'builtins.int'
+[out]
+
+[case testLiteralWithEnumsBasicNewSemanticAnalyzer]
+# flags: --new-semantic-analyzer
```
ilevkivskyi: This test file (in fact all test files except one temporary blacklisted, see `test/hacks.py`) is already run with both analyzers, so no need to duplicate tests.

test-data/unit/check-literal.test
```diff
@@ -2643,3 +2643,306 @@ c_wrap: Literal[4, c]  # E: Invalid type "__main__.c" \
...
+        GREEN = 2
+        BLUE = 3
+
+def foo(x: Literal[Wrapper.Color.RED]) -> None: pass
+
+r: Literal[Wrapper.Color.RED]
+g: Literal[Wrapper.Color.GREEN]
+
+foo(r)
+foo(g)  # E: Argument 1 to "foo" has incompatible type "Literal[__main__.Wrapper.Color.GREEN]"; expected "Literal[__main__.Wrapper.Color.RED]"
```
ilevkivskyi: Here I would however use `Wrapper.Color.RED`, so right thing would be probably to just skip the module name. I think this is already what we do for instances in cases like `x: int = str()`.

test-data/unit/check-literal.test
```diff
@@ -2643,3 +2643,306 @@ c_wrap: Literal[4, c]  # E: Invalid type "__main__.c" \
...
+
+reveal_type(foo)  # E: Revealed type is 'def (x: Literal[__main__.Wrapper.Color.RED])'
+reveal_type(r)    # E: Revealed type is 'Literal[__main__.Wrapper.Color.RED]'
+[out]
+
+[case testLiteralWithEnumsDefinedInClassNewSemanticAnalyzer]
+# flags --new-semantic-analyzer
```
ilevkivskyi: Same as above, no need for this here and below.

test-data/unit/check-literal.test
```diff
@@ -2643,3 +2643,306 @@ c_wrap: Literal[4, c]  # E: Invalid type "__main__.c" \
...
+import mod_b
+
+def f(x: Literal[mod_a.Test.FOO]) -> None: pass
+
+a: Literal[mod_a.Test.FOO]
+b: Literal[mod_a.Test2.FOO]
+c: Literal[mod_b.Test.FOO]
+
+f(a)
+f(b)  # E: Argument 1 to "f" has incompatible type "Literal[mod_a.Test2.FOO]"; expected "Literal[mod_a.Test.FOO]"
```
ilevkivskyi: Even here I would skip the module name. Maybe you can somehow hack this with `MessageBuilder.format_distinctly()`?

test-data/unit/check-literal.test
```diff
@@ -2643,3 +2643,306 @@ c_wrap: Literal[4, c]  # E: Invalid type "__main__.c" \
...
+
+class Test(Enum):
+    FOO = 1
+    BAR = 2
+
+Alias = Test
+
+x: Literal[Alias.FOO]
+reveal_type(x)  # E: Revealed type is 'Literal[__main__.Test.FOO]'
+[out]
```
ilevkivskyi: No newline at the end of file.

mypy/newsemanal/semanal.py
Michael0x2a: I made this change in order to get this example working:

    from enum import Enum

    class Outer:
        class Inner(Enum):
            FOO = 1
            BAR = 2

    x: Literal[Outer.Inner.FOO]
    reveal_type(x)

Basically, the literal-handling code in `typeanal` receives back the string "module.Outer.Inner". We then call `typeanal.named_type_with_normalized_str` which in turn calls `typeanal.named_type` which in turn calls `semanal.lookup_fully_qualified`. That function then ends up crashing because "Outer" is a TypeInfo, not a MypyFile.

FWIW I was under the impression that:

1. `lookup_qualified` is for looking up fully-qualified types, and has a bunch of error handling for when the thing we're trying to look up doesn't exist.
2. `lookup_fully_qualified` is a simpler version of `lookup_qualified`, except without error handling.
3. `lookup` is for looking up some unqualified name across all namespaces (local and global).

I also noticed that `lookup_qualified` handles both MypyFile and TypeInfo, and so figured that the fact that `lookup_fully_qualified` doesn't was an oversight.

test-data/unit/check-literal.test
Michael0x2a: Ooh, I forgot about `MessageBuilder.format_distinctly`! It turned out tweaking that function was sufficient to simplify most of the error messages above in the way you suggested.

mypy/newsemanal/semanal.py
ilevkivskyi: > FWIW I was under the impression that

This is a wrong impression. Which is normal, we have more than dozen lookup functions in the code-base (likely caused by the fact that many people who are not familiar with mypy contributed some reinvented bicycles, which is again normal). Largely, they can be separated in two classes:
* Local lookup: what given name or member expression points to using Python name resolution
* Global lookup: find a symbol using its "full address" used to construct some standard known types.

The `named_type()` and friends are from the second group. While the first group is roughly two methods `lookup()` and `lookup_qualified()`.

Essentially what is going on here looks like a hack (and TBH I don't like it). You already have the `TypeInfo` needed to construct the literal type, but instead you get its full name and pass it to `RawExpressionType` that later passes the name to find the same `TypeInfo` again.

Initially, string name was passed to `RawLiteralExpression` because it is was constructed at very early stage. Here however, it looks unnecessary. Why can't we just immediately return `LiteralType` there with a fallback instance constructed from a known `TypeInfo`?

Michael0x2a: > The `named_type()` and friends are from the second group. While the first group is roughly two methods `lookup()` and `lookup_qualified()`.

Hmm, ok. In that case, what's the correct way of looking up a nested type via a global lookup?

It seems all of the "global lookup" functions assume that all types are defined within modules only, which I'm not so sure is correct.

> Why can't we just immediately return `LiteralType` there with a fallback instance constructed from a known `TypeInfo`?

Sure, we can do that. Is it safe to just construct new whenever we want Instances though? E.g. there isn't a need to register them in some global registry?

Michael0x2a:
> I also noticed a long comment and a TODO in visit_literal_type() in the third pass, is it still relevant? 

I think probably not? I'll make sure to double-check this and see if we can delete the comment while preparing the incremental mode and fine-grained mode tests PR.

mypy/newsemanal/semanal.py
ilevkivskyi: > It seems all of the "global lookup" functions assume that all types are defined within modules only, which I'm not so sure is correct.

It is not correct, but this is kind of optimization. After the refactoring we will have two methods, one fast for builtin things like `builtins.str`, and another one that can be used in plugins etc., with full semantics.

> E.g. there isn't a need to register them in some global registry?

What kind of registry do you mean? I think one can just call `Instance(info, args, line, column)` or am I missing something?.

Michael0x2a: > What kind of registry do you mean? I think one can just call Instance(info, args, line, column) or am I missing something?.

Honestly, I have no idea. Instances and TypeInfos are stuffed with so much functionality that I always get a little worried that touching either of them or creating new ones will break something or interfere with something related to incremental or fine-grained mode in some weird way.

Issue #6432, opened 2019-02-19
New semantic analyzer: crash when function has multiple definitions


JukkaL:
Conditionally redefining a function as an overloaded function results in a crash. Example:

```py
from typing import overload

if int():
    def f(x: int) -> None:
        pass
else:
    @overload
    def f(x: int) -> None: ...
    @overload
    def f(x: str) -> None: ...
    def f(x) -> None:
        yadsf
```

Traceback:

```
  File "/Users/jukka/src/mypy/mypy/build.py", line 162, in build
    result = _build(sources, options, alt_lib_path, flush_errors, fscache)
  File "/Users/jukka/src/mypy/mypy/build.py", line 217, in _build
    graph = dispatch(sources, manager)
  File "/Users/jukka/src/mypy/mypy/build.py", line 2433, in dispatch
    process_graph(graph, manager)
  File "/Users/jukka/src/mypy/mypy/build.py", line 2738, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/Users/jukka/src/mypy/mypy/build.py", line 2836, in process_stale_scc
    semantic_analysis_for_scc(graph, scc, manager.errors)
  File "/Users/jukka/src/mypy/mypy/newsemanal/semanal_main.py", line 62, in semantic_analysis_for_scc
    check_type_arguments(graph, scc, errors)
  File "/Users/jukka/src/mypy/mypy/newsemanal/semanal_main.py", line 214, in check_type_arguments
    state.tree.accept(analyzer)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/contextlib.py", line 99, in __exit__
    self.gen.throw(type, value, traceback)
  File "/Users/jukka/src/mypy/mypy/build.py", line 1754, in wrap_context
    yield
  File "/Users/jukka/src/mypy/mypy/newsemanal/semanal_main.py", line 214, in check_type_arguments
    state.tree.accept(analyzer)
  File "/Users/jukka/src/mypy/mypy/nodes.py", line 249, in accept
    return visitor.visit_mypy_file(self)
  File "/Users/jukka/src/mypy/mypy/newsemanal/semanal_typeargs.py", line 28, in visit_mypy_file
    super().visit_mypy_file(o)
  File "/Users/jukka/src/mypy/mypy/traverser.py", line 33, in visit_mypy_file
    d.accept(self)
  File "/Users/jukka/src/mypy/mypy/nodes.py", line 1102, in accept
[...]

PR #6681, opened 2019-04-16
New semantic analyzer: fix redefining symbols as overloaded functions or classes

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6432

This is a hot fix that simply extends the solution in PR https://github.com/python/mypy/pull/6429 to more situations. We agreed with @JukkaL that he will clean this up when he is back from PTO.

Note that we still give the errors for redefinitions that we consider invalid (see tests), but we don't crash now.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -4162,7 +4162,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         return False
 
     def add_redefinition(self, names: SymbolTable, name: str,
-                              symbol: SymbolTableNode) -> None:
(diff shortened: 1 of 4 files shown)


ilevkivskyi:
> Is there a cleaner solution planned?

The idea of name mangling looks like a hack, but it works OK so far (we also use this for `--allow-redefinition`). The problem is that all alternative cleaner solutions will be big changes. At least we will need to be sure that the name mangling works consistently and covers all corner cases. We can discuss this when we all will be in SF. There also few other topics highlighted by new analyzer work that might be worth discussing, e.g. tracking order of definitions, and SSA-like semantics for assignments.

PR #6683, opened 2019-04-16
New semantic analyzer: fix a crash with imported tricky alias

ilevkivskyi:
This fixes a crash discovered internally, see tests for a simplified repro.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2338,7 +2338,15 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 self.progress = True
                 # We need to defer so that this change can get propagated to base classes.
                 self.defer()
-            existing.node = alias_node
+            if isinstance(existing.node, TypeAlias):
+                # Copy expansion to the existing alias, this matches how we update base classes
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -2338,7 +2338,15 @@ def check_and_set_up_type_alias(self, s: AssignmentStmt) -> bool:
                 self.progress = True
                 # We need to defer so that this change can get propagated to base classes.
                 self.defer()
-            existing.node = alias_node
+            if isinstance(existing.node, TypeAlias):
```
msullivan: What exactly goes wrong without this?

ilevkivskyi: In the test case the alias initially has `<placeholder C>` nested inside, it is deferred and re-analyzed to correctly have `<instance C>`. But before that, the importer already bounds name to the old (partially analyzed) alias node. Therefore we need to update existing node instead of creating a new `TypeAlias` node.

ilevkivskyi: Then this causes a crash because the placeholder type leaks out of semantic analysis.

PR #6684, opened 2019-04-16
Add case sensitive directory checks

ilevkivskyi:
This removes one TODO item that used to cause subtle bugs internally.

The test failed on Mac without the fix, I didn't test on Windows, but I think it should be good as well.

### DIFF
--- a/mypy/fscache.py
+++ b/mypy/fscache.py
@@ -185,10 +185,8 @@ class FileSystemCache:
         """Return whether path exists and is a file.
 
         On case-insensitive filesystems (like Mac or Windows) this returns
-        False if the case of the path's last component does not exactly
-        match the case found in the filesystem.
-        TODO: We should maybe check the case for some directory components also,
-        to avoid permitting wrongly-cased *packages*.
+        False if the case of any path's component does not exactly match
+        the case found in the filesystem.
         """
         if path in self.isfile_case_cache:
             return self.isfile_case_cache[path]
@@ -198,9 +196,21 @@ class FileSystemCache:
         else:
             try:
                 names = self.listdir(head)
+                # This allows to check file name case sensitively in
+                # case-insensitive filesystems.
                 res = tail in names and self.isfile(path)
             except OSError:
                 res = False
+
+        # Also check the other path components in case sensitive way.
+        head, dir = os.path.split(head)
+        while res and head and dir:
+            try:
+                res = dir in self.listdir(head)
+            except OSError:
+                res = False
+            head, dir = os.path.split(head)
+
         self.isfile_case_cache[path] = res
         return res
 
--- a/mypy/fscache.py
+++ b/mypy/fscache.py
@@ -209,7 +209,7 @@ class FileSystemCache:
 
         # Also check the other path components in case sensitive way.
         head, dir = os.path.split(head)
-        while res and head and dir and prefix in head:
+        while res and head and dir and head.startswith(prefix):
             try:
                 res = dir in self.listdir(head)
             except OSError:
--- a/mypy/fscache.py
+++ b/mypy/fscache.py
@@ -181,12 +181,17 @@ class FileSystemCache:
             return False
         return stat.S_ISREG(st.st_mode)
 
-    def isfile_case(self, path: str) -> bool:
+    def isfile_case(self, path: str, prefix: str) -> bool:
         """Return whether path exists and is a file.
 
         On case-insensitive filesystems (like Mac or Windows) this returns
-        False if the case of any path's component does not exactly match
+        False if the case of path's last component does not exactly match
         the case found in the filesystem.
+
+        We check also the case of other path components up to prefix.
+        For example, if path is 'user-stubs/pack/mod.pyi' and prefix is 'user-stubs',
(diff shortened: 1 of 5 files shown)


msullivan:
There is a danger with this approach: it checks case sensitivity for the entire path, not just parts of the path that correspond to a python module. This means that if we have something specified in the mypy search path that does not match the canonical casing, we won't be able to find anything in it.

I'm not sure if this is actually a problem? It could be, though. Worth documenting at least.

 Fixing this means changing the API to include a root part that isn't checked, I guess.

ilevkivskyi:
> Fixing this means changing the API to include a root part that isn't checked, I guess.

Unfortunately this causes some code churn, because I need to "thread" this root part (prefix) through callers. Otherwise this looks a simple straightforward fix.

mypy/fscache.py
```diff
@@ -198,9 +201,21 @@ def isfile_case(self, path: str) -> bool:
...
                 names = self.listdir(head)
+                # This allows to check file name case sensitively in
+                # case-insensitive filesystems.
                 res = tail in names and self.isfile(path)
             except OSError:
                 res = False
+
+        # Also check the other path components in case sensitive way.
+        head, dir = os.path.split(head)
+        while res and head and dir and prefix in head:
```
gvanrossum: Why `prefix in head`? That does a substring check and I can think of so many edge cases where this goes wrong (e.g. if prefix is `'/'` or `''`). I would implement a more robust check that checks `prefix` is actually a filesystem prefix of `head`. Or you could remove `prefix` from `head` before you even enter the loop. Or you could use `os.path.relpath()`. Or the caller could be requested to pass the prefix and whatever follows separately (so you'd have pass `os.path.join(prefix, head)` to `listdir()`. Or...

test-data/unit/check-modules-case.test
```diff
@@ -0,0 +1,57 @@
...
+[file a/__init__.py]
+[file a/b/__init__.py]
+[file FuNkY_CaSe/other.py]
+x = 1
+
+[file mypy.ini]
+[[mypy]
+mypy_path = tmp/funky_case
+
+[case testPreferPackageOverFileCase]
```
gvanrossum: Looks like the test name is backwards. But don't we also need a test for the reverse? (Where the package has the right case and so is preferred.)

ilevkivskyi: There is a similar test case `testPreferPackageOverFile` in `check-modules.test`. I will add however another one here with `mypy_path` just in case.

mypy/fscache.py
ilevkivskyi: Indeed.

mypy/modulefinder.py
```diff
@@ -215,28 +215,31 @@ def _find_module(self, id: str) -> Optional[str]:
         near_misses = []  # Collect near misses for namespace mode (see below).
         for base_dir, verify in candidate_base_dirs:
             base_path = base_dir + seplast  # so e.g. '/usr/lib/python3.4/foo/bar/baz'
+            dir_prefix = base_dir
+            for _ in range(len(components) - 1):
```
msullivan: Might be better to strip off `dir_chain` directly with `base_dir[:len(dir_chain)]`?

mypy/fscache.py
ilevkivskyi: Actually after playing more with these, it looks like `head.startswith(prefix)` is the only one that behaves correctly, while other options I tried fail in some corner cases. Essentially, if the prefix is `''` or `'/'` we should check the whole path in case sensitive manner.

Also could you please show a tricky test case, so that I can add it to the tests?

mypy/modulefinder.py
ilevkivskyi: This doesn't work (also it looks like you forgot minus there). I also tried with plus or minus one and some tests fail.

mypy/fscache.py
```diff
@@ -209,7 +209,7 @@ def isfile_case(self, path: str, prefix: str) -> bool:
 
         # Also check the other path components in case sensitive way.
         head, dir = os.path.split(head)
-        while res and head and dir and prefix in head:
+        while res and head and dir and head.startswith(prefix):
```
gvanrossum: This seems good enough. Perhaps you could see if adding `assert head.startswith(prefix)` before line 211 always passes? OTOH perhaps that's risking too much.

I have a faint worry that e.g. `head = '/foo/bar'` and `prefix = '/foo/b'` but that's not possible given how `prefix` is constructed by the caller right?

ilevkivskyi: > Perhaps you could see if adding `assert head.startswith(prefix)` before line 211 always passes? OTOH perhaps that's risking too much.

I added the assert, and all tests still passed. I think however it is better to not put it there, just in case.

> I have a faint worry that e.g. `head = '/foo/bar'` and `prefix = '/foo/b'` but that's not possible given how prefix is constructed by the caller right?

Yes. I will probably just clarify this in the docstring.

ilevkivskyi:
This PR is still labeled as "Changes required", but IIUC the last "LGTM" means it is good to go. So I am merging now.

PR #6686, opened 2019-04-17
Error when a class member is redefined

### LINKED ISSUE #6571: No error when type annotation for `self.foo` comes after type is already inferred
(Not sure if this is a dupe.)

This is not an error:
```
class C:
    def __init__(self) -> None:
        self.foo = 12
        self.foo: int = 12  # Expected: self.foo is already defined on line 4
```
Even though this is:
```
foo = 12
foo: int = 12   # E: Name 'foo' already defined on line 1
```
It's the same if the annotation has the form of a `# type:` comment.

It's particularly obnoxious if the inferred type is wider than the type annotation, e.g. `Any` -- the inferred type silently wins.

onlined:
Fixes #6571.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2563,6 +2563,8 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             if cur_node and is_final:
                 # Overrides will be checked in type checker.
                 self.fail("Cannot redefine an existing name as final", lval)
+            # On first encounter with this definition, if this attribute was defined before
+            # with an inferred type and it's marked with an explicit type now, give an error.
             if (not lval.node and cur_node and isinstance(cur_node.node, Var) and
                     cur_node.node.is_inferred and explicit_type):
                 self.attribute_already_defined(lval.name, lval, cur_node)
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2555,7 +2555,8 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             if cur_node and is_final:
                 # Overrides will be checked in type checker.
                 self.fail("Cannot redefine an existing name as final", lval)
-            if cur_node and explicit_type:
+            if (not lval.node and cur_node and isinstance(cur_node.node, Var) and
+                    cur_node.node.is_inferred and explicit_type):
                 self.name_already_defined('self.' + lval.name, lval, cur_node)
             # If the attribute of self is not defined in superclasses, create a new Var, ...
             if ((node is None or isinstance(node.node, Var) and node.node.is_abstract_var) or
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2555,6 +2555,8 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             if cur_node and is_final:
                 # Overrides will be checked in type checker.
                 self.fail("Cannot redefine an existing name as final", lval)
+            if cur_node and explicit_type:
+                self.name_already_defined('self.' + lval.name, lval, cur_node)
             # If the attribute of self is not defined in superclasses, create a new Var, ...
             if ((node is None or isinstance(node.node, Var) and node.node.is_abstract_var) or
                     # ... also an explicit declaration on self also creates a new Var.
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2565,7 +2565,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 self.fail("Cannot redefine an existing name as final", lval)
             if (not lval.node and cur_node and isinstance(cur_node.node, Var) and
                     cur_node.node.is_inferred and explicit_type):
-                self.name_already_defined('self.' + lval.name, lval, cur_node)
+                self.attribute_already_defined(lval.name, lval, cur_node)
             # If the attribute of self is not defined in superclasses, create a new Var, ...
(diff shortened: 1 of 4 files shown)


ilevkivskyi:
Note that your fix broke a whole bunch of new analyzer tests.

onlined:
Fixed.

mypy/newsemanal/semanal.py
```diff
@@ -2563,6 +2563,9 @@ def analyze_member_lvalue(self, lval: MemberExpr, explicit_type: bool, is_final:
             if cur_node and is_final:
                 # Overrides will be checked in type checker.
                 self.fail("Cannot redefine an existing name as final", lval)
+            if (not lval.node and cur_node and isinstance(cur_node.node, Var) and
+                    cur_node.node.is_inferred and explicit_type):
```
ilevkivskyi: TBH I don't follow the logic here, maybe add a comment explaining this?

test-data/unit/check-classes.test
```diff
@@ -5837,3 +5837,9 @@ def test() -> None:
         return
     reveal_type(x)  # E: Revealed type is 'Union[Type[__main__.One], Type[__main__.Other]]'
 [builtins fixtures/isinstancelist.pyi]
+
+[case testMemberRedefinition]
+class C:
+    def __init__(self) -> None:
+        self.foo = 12
+        self.foo: int = 12  # E: Name 'self.foo' already defined on line 3
```
ilevkivskyi: Could you please also add a test for something like this:
```python
class C:
    foo = 12
    def __init__(self) -> None:
        self.foo: int = 12  # E: Attribute 'foo' already defined on line 2
```

mypy/semanal.py
```diff
@@ -2288,6 +2288,9 @@ def analyze_member_lvalue(self, lval: MemberExpr, explicit_type: bool, is_final:
             if cur_node and is_final:
                 # Overrides will be checked in type checker.
                 self.fail("Cannot redefine an existing name as final", lval)
+            if (cur_node and isinstance(cur_node.node, Var) and cur_node.node.is_inferred and
+                    explicit_type):
+                self.name_already_defined('self.' + lval.name, lval, cur_node)
```
ilevkivskyi: I wouldn't use this error message because `self.foo` is not a name, it is an attribute. I would better use `Attribute "foo" already defined ...`.

mypy/newsemanal/semanal.py
```diff
@@ -2563,6 +2563,9 @@ def analyze_member_lvalue(self, lval: MemberExpr, explicit_type: bool, is_final:
...
+                self.name_already_defined('self.' + lval.name, lval, cur_node)
```
ilevkivskyi: Using this in the new analyzer is dangerous, it not not only shows the error, but also has some special code for deferring unknown names.

mypy/newsemanal/semanal.py
onlined: Yes, I can write a comment. But before, let me explain: Fail if passing first time, the node exists and it's a `Var` node, it's type is inferred, and new assignment has an explicit type.

mypy/newsemanal/semanal.py
onlined: I couldn't understand what you meant. Could you explain it a little more?

test-data/unit/check-newsemanal.test
```diff
@@ -2179,6 +2179,12 @@ else:
     def f(x) -> None:
         y()  # E: "str" not callable
 
+[case testNewAnalyzerClassMemberRedefinition]
+class C:
+    def __init__(self) -> None:
+        self.foo = 12
+        self.foo: int = 12  # E: Name 'self.foo' already defined on line 3
+
```
onlined: I am not very sure about this. Should I really add same tests to new semantic analyzer tests?

mypy/semanal.py
onlined: Fixed.

test-data/unit/check-newsemanal.test
ilevkivskyi: You don't need to duplicate tests, most test are currently run with both analyzers.

test-data/unit/check-classes.test
ilevkivskyi: You didn't add this test.

mypy/newsemanal/semanal.py
ilevkivskyi: I actually confused with with `name_not_defined()`. This one is OK. Anyway, did you check that in the situation where an error is emitted, no _new_ variable created, and the l.h.s. points to an existing one?

mypy/newsemanal/semanal.py
ilevkivskyi: Basically yes. But please try to avoid this:
```
x += 1  # Increment x
```
Describe the intent and the code achieves it.

ilevkivskyi:
Thanks for updates! I left few more comments.

mypy/newsemanal/semanal.py
onlined: yes, basically if `cur_node` is not `None`, l.h.s. points to an existing variable.

test-data/unit/check-newsemanal.test
onlined: Removed redundant tests.

mypy/newsemanal/semanal.py
ilevkivskyi: @onlined Are you going to add this comment (also in the old analyzer)?

onlined: Yes, I will.

PR #6690, opened 2019-04-18
Delete build-requirements.txt

gaborbernat:
With the move to pyproject.toml and PEP-517 this is no longer needed.


gaborbernat:
Nope we don't need this file anymore.

PR #6691, opened 2019-04-18
ignore pip-wheel-metadata

gaborbernat:
pip 19 generates this if installing current folder https://github.com/pypa/pip/issues/6213


msullivan:
Sure but please add a newline

gaborbernat:
Done.

PR #6694, opened 2019-04-18
add coverege report to tox run

gaborbernat:
Whenever we run with tox we automatically collect coverage data. The users can use the coverage environment to generate the report (stdout, xml, html). This also works merging outcomes from multiple environments. While at it normalized the format (indent by 4 space), and bump to tox ``3.9.0``.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,22 +1,45 @@
 [tox]
-minversion = 3.5.1
+minversion = 3.8.0
 skip_missing_interpreters = true
-envlist = py35,
-          py36,
-          py37,
-          lint,
-          type,
-          docs
+envlist =
+    py35,
+    py36,
+    py37,
+    lint,
+    type,
+    docs,
 isolated_build = true
 
 [testenv]
 description = run the test driver with {basepython}
+setenv = cov: COVERAGE_FILE={toxworkdir}/.coverage.{envname}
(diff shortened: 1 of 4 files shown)


gaborbernat:
@msullinvan can you take a look at this too? thanks!

tox.ini
```diff
@@ -1,22 +1,47 @@
...
 [testenv]
 description = run the test driver with {basepython}
+setenv = COVERAGE_FILE={toxworkdir}/.coverage.{envname}
 passenv = PYTEST_XDIST_WORKER_COUNT NEWSEMANAL
 deps = -rtest-requirements.txt
-commands = pytest {posargs}
+commands =
+    python -m pytest {posargs} \
+    --cov mypy \
+    --cov-config setup.cfg
```
msullivan: Does this incur a performance penalty? If so, I don't think I want it as the default

gaborbernat: Decided to profile it:

With:
      630.19 real      2068.13 user       129.78 sys
Without:
     307.89 real       942.10 user       108.35 sys

gaborbernat:
@msullivan with the profiling above decided to instead do ``tox -e py37-cov`` - that is what will run with coverage report, ``tox -e py37`` will run without coverage report (default fast, but give the option to generate report).

PR #6695, opened 2019-04-18
[WIP] Ensure stub and sources are in sync

gaborbernat:
Resurrect #5139.

I would like to have a discussion about how source files and stub files work together. Currently, if the stub file is specified the source is blankly ignored. This is troublesome as a stub and the source can easily get out of sync. I would like to improve on this. Ideally having a stub and the source file should be the same as inline type annotations with source file only.

What I would propose:
- prefer the stubs as the source location; if the stub file is present disallow any type annotation in the source file (check for this and error if they do),
- ensure that source file signature and stub signature/type matchup (report when they don't),
- report source elements without type data (present in the source but not in tab),
- report type stubs without source data (present in the stub but no source - type aliases need to be exempt).

What makes this really tricky at the moment is that now 2 files can make up a module. Our current cache is 1 module - 1 file based. This will need to be broken sadly.


gvanrossum:
@gaborbernat Did you intentionally close this?

gaborbernat:
By error, but I do plan to reopen it during the sprint.

Issue #6303, opened 2019-02-01
Detect and flag as errors forward references to variables


ilevkivskyi:
With the old semantic analyzer this code is an error:
```python
def func() -> None:
    x
    x = 1
```
We should also flag it as an error in the new semantic analyzer (ideally also at module scope but it doesn't work currently and there is a separate issue).

This is a follow-up for https://github.com/python/mypy/pull/6280

ilevkivskyi:
This looks like a part of a more general problem with the new analyzer. In the old analyzer definitions are always processed in the order they appear in the source file. In the new analyzer it is no longer true and we need a (robust) way to track order of definitions. So far this caused at least three hard to solve issues:
* Forward references work to good, like in the original example.
* Scopes can be confused for the same names, for example:
  ```python
  X = 1
  class C:
      X = X  # Cannot determine type of 'X'
  ```
* An original definition may not take precedence, see e.g. https://github.com/python/mypy/pull/6696

I therefore raise priority to high.

erictraut:
This appears to no longer be a bug. The latest version of mypy handles these cases correctly.

PR #6696, opened 2019-04-18
New semantic analyzer: When defining an alias first definition wins

ilevkivskyi:
In the old analyzer the first definition always wins. We should try to always preserve this logic. This is a hotfix for a problem discovered internally, we should try to find a more robust and general solution.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2270,8 +2270,12 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         #     A = float  # OK, but this doesn't define an alias
         #     B = int
         #     B = float  # Error!
-        if existing and (isinstance(existing.node, Var) or
-                         isinstance(existing.node, TypeAlias) and not s.is_alias_def):
+        # Don't create an alias in these cases:
+        if existing and (isinstance(existing.node, Var) or  # existing variable
+                isinstance(existing.node, TypeAlias) and not s.is_alias_def or  # existing alias
+                (isinstance(existing.node, PlaceholderNode) and
+                # TODO: find a more robust way to track the order of definitions.
+                 existing.node.node.line < s.line)):  # or previous incomplete definition
             # Note: if is_alias_def=True, this is just a node from previous iteration.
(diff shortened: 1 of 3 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -2270,8 +2270,12 @@ def check_and_set_up_type_alias(self, s: AssignmentStmt) -> bool:
...
         #     B = int
         #     B = float  # Error!
-        if existing and (isinstance(existing.node, Var) or
-                         isinstance(existing.node, TypeAlias) and not s.is_alias_def):
+        # Don't create an alias in these cases:
+        if existing and (isinstance(existing.node, Var) or  # existing variable
+                isinstance(existing.node, TypeAlias) and not s.is_alias_def or  # existing alias
+                (isinstance(existing.node, PlaceholderNode) and
+                # TODO: find a more robust way to track the order of definitions.
+                 existing.node.node.line < s.line)):  # or previous incomplete definition
```
gaborbernat: I believe this to be robust enough in practice, but on equal one would want to also check the column, not?

ilevkivskyi:
> I worry that maybe we overcorrected a bit in the design of the new analyzer...

Yes, now forward references work "too good", I commented on the other issue https://github.com/python/mypy/issues/6303. This is however something we already discussed with @JukkaL for some time, and there are some ideas, so things are not so bad, but I think we will need to devote another couple weeks for more work on the new analyzer. Time when we are all in SF seems ideal for this.

mypy/newsemanal/semanal.py
ilevkivskyi: It is more subtle than this. First of all our tracking of columns is bad (for historical reasons); second, using placeholder sources for anything that affects semantics is dangerous because we didn't assign them carefully before.; finally, we need to be sure this works correctly in incremental and daemon modes when files are edited and processed only partially.

gaborbernat: Is there a plan to fix the columns bug? The daemon thing probably makes it dealbreaker though to go down that path, maybe we should document the currently limited behaviour then, just so that user knows what to expect.

Issue #6548, opened 2019-03-15
Daemon issues with retargeting re-exported names


msullivan:
Consider the following test:
```
[case testRetargetReexport]
from m import M

def f(x: M) -> None: ...

f(M())

[file m.py]
from n import M
[file m.py.2]
from q import M

[file n.py]
class M: ...

[file q.py]
class M: ...
[out]
==
```

The daemon fails in step two with the error
`Argument 1 to "f" has incompatible type "q.M"; expected "n.M"`

msullivan:
There is a related seeming bug with NamedTuples whose names don't match their line numbers:
```
[case testRetargetReexportNamedTuple]
from m import M

def f(x: M) -> None: ...

f(M(0))

[file m.py]
from n import M

[file n.py]
from typing import NamedTuple
M = NamedTuple('_N', [('x', int)])

[file n.py.2]
# oh no
from typing import NamedTuple
M = NamedTuple('_N', [('x', int)])

[out]
==
```
Fails with `Argument 1 to "f" has incompatible type "_N@3"; expected "_N@2" `

Here it is also arguably a bug that the names get messed up like that, but it works if there is not a re-export...

ilevkivskyi:
I think I was also hit by this recently. I raise the priority to high, since this is a relatively large area where daemon's behavior is problematic.

PR #6698, opened 2019-04-18
Use variable name to determined NamedTuple class name

msullivan:
This lets us avoid inserting line numbers into the name when the
variable name and argument to NamedTuple disagree. This is good,
because the line numbers are ugly and when combined with bug #6548
causes problems when a namedtuple is reexported.

### DIFF
--- a/mypy/semanal_namedtuple.py
+++ b/mypy/semanal_namedtuple.py
@@ -179,14 +179,20 @@ class NamedTupleAnalyzer:
             info = self.build_namedtuple_typeinfo(name, [], [], {}, node.line)
             self.store_namedtuple_info(info, name, call, is_typed)
             return True, info
-        name = cast(Union[StrExpr, BytesExpr, UnicodeExpr], call.args[0]).value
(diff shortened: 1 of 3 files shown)


ilevkivskyi:
As we discussed offline, it may be better to do this as part of a larger clean-up proposed in https://github.com/python/mypy/issues/6422, unless this is urgent.

msullivan:
This issue got reported again at Dropbox and since #6422 seems moribund, I've rebased this. I don't remember what the objections we discussed were. @ilevkivskyi, what are your thoughts about merging this now?

JukkaL:
I didn't do a full review, but I think that this is a reasonable incremental improvement. I agree that it would be even better to fix #6422, but it could take a long time before it happens, unfortunately.

ilevkivskyi:
@msullivan Yes, go ahead, it is unlikely I will ever work on #6422.

msullivan:
Apparently after rebasing this I then forgot about it, so I'm rebasing again and am actually going to land it.

Issue #5909, opened 2018-11-17
Specify files/directories in `mypy.ini`


samuelhwilliams:
I'm posting this issue to gather some feedback around whether a feature I'd like would be accepted/well-received.

I'd like to be able to specify target files/directories as part of the `mypy.ini` config file, and I can't see a way to do this at the moment (if I'm missing something, hopefully this can be cleared up quickly 😄 ). This would, in my opinion, reduce the need to have a separate script that runs `mypy` with the correct arguments, as it should be easier to just rely on the settings in the config file.

I would propose something along the lines of a `files` key in the main `mypy` section, but I'm not sure if this is in line with how it works under the `special_opts` namespace. Open to any recommendation around the best section for this to sit in, if something else makes more sense.

Would a PR implementing this feature be merged/are there any reasons why this feature doesn't exist? I asked quickly in the `typing` slack and it wasn't shot down, so hoping to get more of a discussion here.

Cheers.

gvanrossum:
You can’t expect a promise that your PR will be accepted. But a ‘files’
option in mypy.ini, only used when no files are given on the command line,
makes sense. Would it also include globbing?
-- 
--Guido (mobile)

samuelhwilliams:
Wasn't expecting a promise, definitely - more just clarification that there wasn't a specific reason it *wouldn't* be accepted.

I'd intend to replicate the same treatment as the `files` argument on the command line (overridden by the command-line, as you say). The CLI does globbing, so yes, I would intend to make the config file do the same.

gvanrossum:
Technically the CLI does no globbing,the shell does. (Except on Windows
IIRC.) But yes, this makes sense.
-- 
--Guido (mobile)

samuelhwilliams:
Yep, I was a bit loose with my terminology there.

I will start work on putting a PR together.

Cheers.

Michael0x2a:
I remembered this only recently, but we have a [`mypy_path`](https://mypy.readthedocs.io/en/stable/config_file.html#config-file-import-discovery-global) config file option that's intended to serve as a replacement for editing the MYPYPATH environment variable.

Would specifying the files you want to type check there solve your use case? If so, we should probably do a better job of documenting it. If not, it might at least serve as a good foundation for your PR.

hmoravec:
[...]

PR #6707, opened 2019-04-21
Allow files to be specified via ini config

samuelhwilliams:
## Summary
Allow target paths to be specified through config files as an alternative to the command line.

## Reason
To simplify the enforcement of which files/directories should be validated when working in a team. Rather than having to write a separate script that specifies the files, you can simply declare them in the existing `mypy.ini` (or alternative) config file.

## Behaviour
I've tried to copy the functionality that you might expect from the typical command line, so this includes recursive file globbing. There are two main differences:

1) paths are comma- or semicolon-separated rather than just using spaces
2) by using the `glob` module we get `**` resolving to zero or more intermediate directories. In bash (at least), `*/*.py` or `**/*.py` will only match python files in the directory below the current one.

If a path (wildcarded or not) doesn't match at least one file, it bubbles up and errors in the same way as when given on the command line. If any files *are* given on the command line, the value in the config file is ignored.

## Addendum
Closes: #5909

### DIFF
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -333,6 +333,11 @@ a list of import discovery options that may be used
     Specifies the paths to use, after trying the paths from ``MYPYPATH`` environment
     variable.  Useful if you'd like to keep stubs in your repo, along with the config file.
 
+``files`` (string)
+    Specifies the paths which should be checked by mypy if none are given on the command
+    line. Supports recursive file globbing with `*` (eg `*.py` for files in the same directory)
+    and `**/` (eg `**/*.py`) for files in any directory below the current one.
+
 
 Platform configuration
 ----------------------
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -335,7 +335,7 @@ a list of import discovery options that may be used
 
 ``files`` (string)
     Specifies the paths which should be checked by mypy if none are given on the command
-    line. Supports file globbing through
+    line. Supports recursive file globbing through
     [the glob library](https://docs.python.org/3/library/glob.html).
 
 
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -334,9 +334,11 @@ a list of import discovery options that may be used
     variable.  Useful if you'd like to keep stubs in your repo, along with the config file.
 
 ``files`` (string)
-    Specifies the paths which should be checked by mypy if none are given on the command
-    line. Supports recursive file globbing through
-    [the glob library](https://docs.python.org/3/library/glob.html).
+    A comma-separated list of paths which should be checked by mypy if none are given on the command
+    line. Supports recursive file globbing using
+    [the glob library](https://docs.python.org/3/library/glob.html), where `*` (eg `*.py`) matches
+    files in the current directory and `**/` (eg `**/*.py`) matches files in any directories below
+    the current one.
 
 
 Platform configuration
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -335,8 +335,8 @@ a list of import discovery options that may be used
 
 ``files`` (string)
     Specifies the paths which should be checked by mypy if none are given on the command
-    line. Supports recursive file globbing with `*` (eg `*.py` for files in the same directory)
-    and `**/` (eg `**/*.py`) for files in any directory below the current one.
+    line. Supports file globbing through
+    [the glob library](https://docs.python.org/3/library/glob.html).
 
 
 Platform configuration
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -336,8 +336,8 @@ a list of import discovery options that may be used
 ``files`` (string)
     A comma-separated list of paths which should be checked by mypy if none are given on the command
     line. Supports recursive file globbing using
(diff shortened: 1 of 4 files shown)


mypy/main.py
```diff
@@ -866,6 +872,26 @@ def process_cache_map(parser: argparse.ArgumentParser,
...
+def split_and_match_files(paths: str) -> List[str]:
+    """Take a string representing a list of files/directories (including globbing with */**).
+
+    Where a path/glob matches no file, we still include the raw path in the resulting list.
+
+    Returns a list of file paths
+    """
+    expanded_paths = []
+
+    for path in re.split('[,:]', paths):
```
ethanhs: Is the colon here due it being the path separator on Unix? If so, it might be good to also add semicolon, as that is the path separator on Windows.

docs/source/config_file.rst
```diff
@@ -333,6 +333,11 @@ a list of import discovery options that may be used
     Specifies the paths to use, after trying the paths from ``MYPYPATH`` environment
     variable.  Useful if you'd like to keep stubs in your repo, along with the config file.
 
+``files`` (string)
+    Specifies the paths which should be checked by mypy if none are given on the command
+    line. Supports recursive file globbing with `*` (eg `*.py` for files in the same directory)
```
ethanhs: I think it might be good to add we pass globbing onto `glob`  (with `recursive=True`) and link https://docs.python.org/3/library/glob.html as glob supports more than `*` and `**`, though those are likely the most used constructs.

mypy/main.py
samuelhwilliams: I copied this re.split from the lambda on mypy_path a few lines down. Happy to add the semicolon as well; would it make sense to add it there too?

mypy/main.py
```diff
@@ -730,6 +731,11 @@ def add_invertible_flag(flag: str,
     if special_opts.no_executable:
         options.python_executable = None
 
+    # Paths listed in the config file will be ignored if any paths are passed on
+    # the command line.
+    if not special_opts.files and options.files:
```
gvanrossum: I'd write this as `if options.files and not special_opts.files:` -- just because in English the priority of `not` vs. `and` is not so clear.

docs/source/config_file.rst
```diff
@@ -333,6 +333,11 @@ a list of import discovery options that may be used
     Specifies the paths to use, after trying the paths from ``MYPYPATH`` environment
     variable.  Useful if you'd like to keep stubs in your repo, along with the config file.
 
+``files`` (string)
+    Specifies the paths which should be checked by mypy if none are given on the command
+    line. Supports recursive file globbing through
+    [the glob library](https://docs.python.org/3/library/glob.html).
```
gvanrossum: Also this should mention the separators used (comma, colon, semicolon), or whatever we end up deciding in the end.

mypy/options.py
```diff
@@ -179,6 +179,10 @@ def __init__(self) -> None:
         # source hashes as often.
         self.quickstart_file = None  # type: Optional[str]
 
+        # A comma/semicolon-separated list of files/directories for mypy to type check;
```
gvanrossum: Update comment (it's comma, colon, semicolon), unless we change our mind.

mypy/main.py
gvanrossum: Honestly I wonder if it should just be comma, or space and comma? This isn't a search path like MYPYPATH, and when I tried to test this the very first mistake I made was using a space for a separator.

(I should add that for mypy_path, the behavior for `$MYPYPATH` is different than that for the config `mypy_path` variable -- the latter always uses `[,:]`, while the former uses `os.pathsep`, which is only `:` on *NIX and only `;` on Windows. So I don't think that's a great model.)

docs/source/config_file.rst
```diff
@@ -335,7 +335,7 @@ a list of import discovery options that may be used
 
 ``files`` (string)
     Specifies the paths which should be checked by mypy if none are given on the command
-    line. Supports file globbing through
+    line. Supports recursive file globbing through
```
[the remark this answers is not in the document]
ethanhs: Oh yes, I liked the original text too, I should have mentioned to keep it.

mypy/main.py
samuelhwilliams: I'm a smidge nervous about using spaces as separators as that seems the most likely of all to be legitimately used in a path (albeit hopefully unlikely), and managing those within paths would introduce a lot more complexity. I think probably just commas are sufficient? It will always be processed slightly differently than by a shell, so making a clear distinction feels reasonable.

docs/source/config_file.rst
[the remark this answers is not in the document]
samuelhwilliams: Sure, makes sense. I'll add those details back in.

mypy/main.py
samuelhwilliams: Yep, that's how I did it the first time and then decided to flip it. I'll change it back.

mypy/main.py
gvanrossum: I guess that might be the case on Windows... Agreed though, so let's do just comma. (I suppose colon is also sketchy on Windows, as in `C:\`.)

samuelhwilliams:
The latest should address all of your comments so far :)

gvanrossum:
Thanks! I fixed a spelling nit, and I plan to merge this once the tests pass.

PR #6711, opened 2019-04-22
Fixes #6743

### LINKED ISSUE #6473: mypy does not recognize __qualname__
1. Given Python file `test1.py`
```Python
class A(object):
    NAME = __qualname__
```
2. run mypy checking

       mypy test1.py

3. **_BUG:_** mypy reports `not defined`

       test1.py:2: error: Name '__qualname__' is not defined

The Python interpreter runs the code without error.

`__qualname__` has [been available since Python 3.3](https://docs.python.org/3/library/stdtypes.html#definition.__qualname__).

Tested on _Python 3.7.1_, using _mpypy 0.670_.

aaalex5:
Mypy will type check '__qualname__' attribute without throwing error. Important for Python versions 3.3 and up.

Fixes #6473

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -95,6 +95,7 @@ inverse_node_kinds = {_kind: _name for _name, _kind in node_kinds.items()}  # ty
 
 
 implicit_module_attrs = {'__name__': '__builtins__.str',
+                         '__qualname__': '__builtins__.str',
                          '__doc__': None,  # depends on Python version, see semanal.py
                          '__file__': '__builtins__.str',
                          '__package__': '__builtins__.str'}  # type: Final
--- a/test-data/unit/check-qualname.test
+++ b/test-data/unit/check-qualname.test
@@ -1,0 +1,10 @@
+[case testQualNameTypeCheck]
+
+# flags: --python-version 2.7
+class C:
+    NAME = __qualname__ # E: NameError: name '__qualname__' is not defined
+
+# flags: --python-version 3.3
+class A:
+    NAME = __qualname__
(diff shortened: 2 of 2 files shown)


mypy/nodes.py
```diff
@@ -95,6 +95,7 @@ def get_column(self) -> int:
 
 
 implicit_module_attrs = {'__name__': '__builtins__.str',
+                         '__qualname__': '__builtins__.str',
```
ilevkivskyi: I don't think this is a right solution since this way mypy will not complain about `__qualname__` at module scope. The latter is defined only at a class scope.

test-data/unit/check-qualname.test
```diff
@@ -0,0 +1,10 @@
+[case testQualNameTypeCheck]
+
+# flags: --python-version 2.7
+class C:
+    NAME = __qualname__ # E: NameError: name '__qualname__' is not defined
+
+# flags: --python-version 3.3
```
ilevkivskyi: You can't have two separate frag sets in the same test case, you need to have a separate one.

test-data/unit/check-qualname.test
```diff
@@ -0,0 +1,10 @@
+[case testQualNameTypeCheck]
```
ilevkivskyi: There is no need to add a separate test file for this special case. Look where other tests for `__name__` etc. are defined and add your test cases there.

msullivan:
I'm going to close this as part of a PR tidy. Feel free to reopen if you address Ivan's comments and fix the tests.

Issue #5603, opened 2018-09-11
Per-file strict optional doesn't play well with subtype cache


ilevkivskyi:
Consider this test case:
```
[case testPerFileStrictOptionalTricky]
# flags: --config-file tmp/mypy.ini
import a
[file a.py]
from typing import List, Optional
import b
x: List[int]
y: List[Optional[int]]
x = y  # E: Incompatible types in assignment (expression has type "List[Optional[int]]", variable has type "List[int]")
[file b.py]
from typing import List, Optional
x: List[int]
y: List[Optional[int]]
x = y

[file mypy.ini]
[[mypy]
strict_optional = True
[[mypy-b]
strict_optional = False
[builtins fixtures/list.pyi]
[out]
```
The error in `a.py` is not detected because the _same_ two types are first compared in `b.py` and the  positive result is stored in the subtype cache.

After studying a bit the code around per-file strict optional, it looks quite problematic. There is `TypeChecker.should_suppress_optional_error()` with some questionable code, but it actually doesn't work because `self.suppress_none_errors` is `False` even if strict optional is disabled. In fact, the strict optional errors are suppressed because `SubtypeVisitor` has code that depends on `experiments.STRICT_OPTIONAL`.

In real world code, the false negatives caused by this are rare, but I think this is worth fixing because the code is fragile.

JukkaL:
Would it help if we added the value of `experiments.STRICT_OPTIONAL` (the name of which is a misnomer) in the subtype cache key?

ilevkivskyi:
Yes, but I think we already have many flags there, so I would avoid adding them (plus it may affect performance).

JukkaL:
Have you a better way to fix the issue?

I don't see adding a new flag to a cache key as a problem. It will be used only in a few places, and we just need to document what it means and why we include it.

The performance impact is easy enough to measure/estimate once we have an implementation.

PR #6714, opened 2019-04-23
Track strict optional mode in typestate cache

msullivan:
Otherwise subtype relationships that are valid with strict optional
off (like List[str] <: List[Optional[str]]) can leak into strict
optional code via the subtype cache.

### DIFF
--- a/mypy/typestate.py
+++ b/mypy/typestate.py
@@ -12,6 +12,7 @@ if MYPY:
 from mypy.nodes import TypeInfo
 from mypy.types import Instance
 from mypy.server.trigger import make_trigger
+from mypy import state
 
 # Represents that the 'left' instance is a subtype of the 'right' instance
 SubtypeRelationship = Tuple[Instance, Instance]
@@ -22,7 +23,7 @@ SubtypeKind = Tuple[bool, ...]
 
 # A cache that keeps track of whether the given TypeInfo is a part of a particular
 # subtype relationship
-SubtypeCache = Dict[TypeInfo, Dict[SubtypeKind, Set[SubtypeRelationship]]]
+SubtypeCache = Dict[TypeInfo, Dict[Tuple[bool, SubtypeKind], Set[SubtypeRelationship]]]
 
 
 class TypeState:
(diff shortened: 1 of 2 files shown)


test-data/unit/check-optional.test
```diff
@@ -751,3 +751,28 @@ def g(x: Optional[int]) -> int:
...
+
+[file mypy.ini]
+[[mypy]
+[[mypy-a]
+strict_optional = False
+[out]
+main:4: error: Argument 1 to "asdf" has incompatible type "List[str]"; expected "List[Optional[str]]"
+main:4: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
+main:4: note: Consider using "Sequence" instead, which is covariant
+[builtins fixtures/list.pyi]
```
ilevkivskyi: There is another (but very similar, so I am not sure if it is worth adding) test case in the original issue https://github.com/python/mypy/issues/5603 (don't forget to update PR description before merging to auto-close).

mypy/typestate.py
```diff
@@ -104,15 +106,16 @@ def is_cached_subtype_check(kind: SubtypeKind, left: Instance, right: Instance)
         if info not in TypeState._subtype_caches:
             return False
         cache = TypeState._subtype_caches[info]
-        if kind not in cache:
+        key = (state.strict_optional, kind)
```
ilevkivskyi: `kind` is itself a tuple of bools, why don't just add a one more bool there? (Essentially this will be equivalent to `(state.strict_optional, *kind)`.)

ilevkivskyi:
Btw, there was a concern about how this would affect performance, did you make any measurements?

msullivan:
> Btw, there was a concern about how this would affect performance, did you make any measurements?

Just did. Didn't seem noticeable.

msullivan:
(I didn't find #5603 when I was looking because I checked topic-affects-strict-optional and not topic-depends-on-strict-optional!)

gvanrossum:
We should unify those two labels into topic-strict-optional.
-- 
--Guido (mobile)

PR #6719, opened 2019-04-24
Rename NoneTyp to NoneType

msullivan:
Every other type gets an e in its name. Might as well fix this one.
Keeps a NoneTyp alias to NoneType to avoid breaking external plugins.

### DIFF
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -6,7 +6,7 @@ MYPY = False
 if MYPY:
     from typing import DefaultDict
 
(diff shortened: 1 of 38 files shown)


msullivan:
This is a pretty mechanical sed PR, so I'm happy to postpone it to whenever if there are some big PRs people want to avoid a merge conflict in, but I think it is worth doing.

gvanrossum:
I suppose it was because it would shadow the builtin `NoneType`, which is the type of `None`, aka `type(None)`. But I agree we don't need to reference that much, and `type(None)` is fine. So +1.

Issue #5973, opened 2018-11-29
Definition of "__iter__" in "ItemsView" is incompatible with definition "Sequence"


althonos:
Hi there !

Trying to annotate a class inheriting from both a sequence and an items view will raise an error, although I believe both definitions are compatible in the following case:

```python
from typing import *

K = TypeVar('K')
V = TypeVar('V')

class OrderedItemsView(Generic[K, V], ItemsView[K, V], Sequence[Tuple[K, V]]):
  def __iter__(self) -> Iterator[Tuple[K, V]]:
    return iter([])
```

This example raises:
```
test.py:7: error: Definition of "__iter__" in base class "ItemsView" is incompatible with definition in base class "Sequence"
```
although, for the two base classes, I  believe `__iter__` to be defined as `Iterator[Tuple[K, V]]`.

PR #6720, opened 2019-04-25
Fix multiple inheritance check for generic classes

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5973

The fix is straightforward, I just copy a bit of logic (bind self and map type) from normal override check
to the multiple inheritance one (I also factored it out in a small method to avoid duplication).

Using this opportunity I also added/expanded some docstrings.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1647,6 +1647,20 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         Assume base1 comes before base2 in the MRO, and that base1 and base2 don't have
         a direct subclass relationship (i.e., the compatibility requirement only derives from
         multiple inheritance).
+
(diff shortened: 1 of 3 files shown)


mypy/checker.py
```diff
@@ -1628,22 +1641,35 @@ def determine_type_of_class_member(self, sym: SymbolTableNode) -> Optional[Type]
...
+            x: A[int] = C()
+            x.foo  # ...runtime type is (str) -> None, while static type is (int) -> None
         """
         if name in ('__init__', '__new__', '__init_subclass__'):
             # __init__ and friends can be incompatible -- it's a special case.
             return
-        first = base1[name]
-        second = base2[name]
+        first = base1.names[name]
+        second = base2.names[name]
```
ilevkivskyi: This change is not needed, but it makes the intent of this method more evident.

PR #6723, opened 2019-04-25
Improvements to dmypy suggest

msullivan:
* Add option to require that a suggestion produce no errors
* Add option to require that a suggestion not contain Any
* Handle `__init__`
* Cap the number of suggestions considered
* Improve output formatting a bunch

### DIFF
--- a/mypy/suggestions.py
+++ b/mypy/suggestions.py
@@ -512,7 +512,8 @@ class TypeFormatter(TypeStrVisitor):
         if mod == 'builtins':
             return obj
         else:
-            return "{}:{}".format(mod, obj)
+            delim = '.' if '.' not in obj else ':'
+            return mod + delim + obj
 
     def visit_tuple_type(self, t: TupleType) -> str:
         if t.partial_fallback and t.partial_fallback.type:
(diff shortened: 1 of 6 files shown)


mypy/suggestions.py
```diff
@@ -135,19 +142,28 @@ def is_implicit_any(typ: Type) -> bool:
 class SuggestionEngine:
     """Engine for finding call sites and suggesting signatures."""
 
-    def __init__(self, fgmanager: FineGrainedBuildManager):
+    def __init__(self, fgmanager: FineGrainedBuildManager,
+                 json: bool,
+                 no_errors: bool = False,
+                 no_any: bool = False) -> None:
```
gvanrossum: It seems inconsistent to have defaults for the new flags but not for json. It seems that at runtime all flags are always passed by the caller; OTOH it may be more consistent with the spirit of `**kwds` to have a default for `json` too.

mypy/suggestions.py
```diff
@@ -224,7 +241,9 @@ def get_args(self, is_method: bool,
                 types.append(generate_type_combinations(all_arg_types))
             else:
                 # If we don't have anything, we'll try Any and object
-                types.append([AnyType(TypeOfAny.explicit), self.builtin_type('builtins.object')])
+                # (Actually object usually is bad for downstream consumers...)
+                # types.append([AnyType(TypeOfAny.explicit), self.builtin_type('builtins.object')])
```
gvanrossum: Maybe commit to not doing object rather than commenting it out and making the comment contradict itself?

PR #6724, opened 2019-04-25
New semantic analyzer: Fix --disallow-untyped-defs with attrs plugin and forward references

ilevkivskyi:
The fix essentially mirrors the one for dataclasses. Note I temporary add `final_iteration` to the old analyzer, because plugin files import everything from the old one.

The non-trivial part is removing previously generated functions in plugins to make them re-entrant.

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -39,7 +39,7 @@ Every value stored there must be JSON-serializable.
 
 import types
 
-from abc import abstractmethod
+from abc import abstractmethod, abstractproperty
 from typing import Any, Callable, List, Tuple, Optional, NamedTuple, TypeVar, Dict
 from mypy_extensions import trait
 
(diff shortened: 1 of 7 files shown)


ilevkivskyi:
Actually, it looks like the guard against final iteration is not necessary, I am going to remove it for now.

ilevkivskyi:
Oh, it is actually needed, I just run the tests without `NEWSEMANAL=1`.

ilevkivskyi:
The test failures are actually hard to fix. They highlight a problem with all plugins that use `add_method()` -- they are not re-entrant. The `add_method()` calls `.append()` on class body that causes multiple functions with same name added if there is any deferral in the file. I am surprised no tests failed because of this before.

ilevkivskyi:
> Explicitly checking for the new analyzer and explicitly checking final_iteration are pretty nitty-gritty details that we don't want to have to inflict everywhere.

Yes. One of this will go after the new one is the default. For the second one, possible option is to have `defer_if_possible()`. It will be actually useful in general, I think we already use this pattern in several places in semantic analyzer. I will create an issue for this.

ilevkivskyi:
@msullivan Opened https://github.com/python/mypy/issues/6725

PR #6731, opened 2019-04-27
Fix mypyc build

msullivan:
PR #6645 introduced a mypyc failure due to the irritating #3603
intersection type issue.

Fix it by testing again SYMBOL_FUNCBASE_TYPES instead of FuncBase


msullivan:
And, of course, I should add mypyc to our CI: #6170

Issue #6516, opened 2019-03-06
Incorrect "Attributes without a default cannot follow attributes with one" in dataclasses


reinhrst:
Mypy incorrectly (in my understanding) complains about members in dataclasses having default values, even though many `dataclasses.field` assignments do not actually assign a default value.

I feel the correct behaviour should be to only detect an error if a field assignment that leads to a default value being assigned in the init method is followed by a default value.

This results (incorrectly) in an error:
```
import dataclasses

@dataclasses.dataclass
class A:
    x: int = dataclasses.field(metadata={"doc": "foo"})
    y: str
```

So does this:
```
import dataclasses

@dataclasses.dataclass
class A:
    x: int = dataclasses.field(init=False, default=1)
    y: str
```

Whereas this does not emit an error (where it should):
```
import dataclasses

@dataclasses.dataclass
class A:
    x: int = dataclasses.field(default=1)
    y: str = dataclasses.field(metadata={"doc": "foo"})
```

I understand that there may be quite some complexities to fix this correctly, but I would suggest in that case just removing this check altogether until it works correctly.

Tested on python 3.7.2 with mypy 0.670 and master, and python 3.6.6 with mypy 0.670

ilevkivskyi:
Sorry for late answer, these examples indeed look like a bug. I think the fix can be simple, just use `RefExpr` in `_collect_field_args`.

ilevkivskyi:
Here is the fix https://github.com/python/mypy/pull/6739

PR #6739, opened 2019-04-29
Fix dataclass plugin interaction with member expressions

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6516

It also fixes minor corner case that affects tests cases proposed in the issue.

### DIFF
--- a/mypy/plugins/dataclasses.py
+++ b/mypy/plugins/dataclasses.py
@@ -300,7 +300,7 @@ class DataclassTransformer:
                     Context(line=attr.line, column=attr.column),
                 )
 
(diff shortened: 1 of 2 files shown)


mypy/plugins/dataclasses.py
```diff
@@ -300,7 +300,7 @@ def collect_attributes(self) -> List[DataclassAttribute]:
                     Context(line=attr.line, column=attr.column),
                 )
 
-            found_default = found_default or attr.has_default
+            found_default = found_default or attr.has_default and attr.is_in_init
```
JukkaL: Style nit: add parentheses to make precedence explicit.

PR #6743, opened 2019-04-29
Fix mypyc build

msullivan:
Which doesn't support **kwargs


msullivan:
And of course, since not even I can remember to not use features mypyc doesn't support, I need to add mypyc to our CI: #6170

PR #6744, opened 2019-04-30
Support TypedDict imported from typing_extensions

ilevkivskyi:
This updates a bunch of places to work with both `mypy_extensions.TypedDict` and `typing_extensions.TypedDict`. There is one place where we create a fallback, there I prefer the latter if available.

See PR https://github.com/python/typing/pull/628 for corresponding runtime change.


msullivan:
Would it work to instead make mypy_extensions.TypedDict just be a reexport of typing_extensions.TypedDict in typeshed, and then just update things to be typing_extensions.TypedDict instead of handling both?

ilevkivskyi:
> Would it work to instead make mypy_extensions.TypedDict just be a reexport of typing_extensions.TypedDict in typeshed, and then just update things to be typing_extensions.TypedDict instead of handling both?

It would definitely work, but this makes the transition smoother, we don't need to worry about the order of the PRs, and people who have wrong version of typeshed will not see weird errors. Moreover, the next transition when it will be moved to `typing` will be just updating couple constants, instead of same sized PRs here and in typeshed. Finally, I like a single constant more than same hardcoded string everywhere.

PR #6750, opened 2019-05-01
Fix sys.stdout overriding in mypy.api

### LINKED ISSUE #6125: Stdout hijacking in api.py is not thread safe

AnOctopus:
These are the changes made by @elkhadiy to resolve the problems with stdout not being thread-safe in the api.

Fixes #6125.

### DIFF
--- a/mypy/api.py
+++ b/mypy/api.py
@@ -48,13 +48,13 @@ from typing import List, Tuple, Union, TextIO, Callable
 from mypy_extensions import DefaultArg
 
 
-def _run(f: Callable[[TextIO, TextIO], None]) -> Tuple[str, str, int]:
+def _run(main_wrapper: Callable[[TextIO, TextIO], None]) -> Tuple[str, str, int]:
 
     stdout = StringIO()
     stderr = StringIO()
 
     try:
-        f(stdout, stderr)
+        main_wrapper(stdout, stderr)
         exit_status = 0
     except SystemExit as system_exit:
         exit_status = system_exit.code
--- a/mypy/api.py
+++ b/mypy/api.py
@@ -44,36 +44,31 @@ print('\nExit status:', result[2])
 
 import sys
 from io import StringIO
-from typing import List, Tuple, Callable
+from typing import List, Tuple, Union, TextIO, Callable
+from mypy_extensions import DefaultArg
 
 
-def _run(f: Callable[[], None]) -> Tuple[str, str, int]:
-    old_stdout = sys.stdout
-    new_stdout = StringIO()
-    sys.stdout = new_stdout
+def _run(f: Callable[[TextIO, TextIO], None]) -> Tuple[str, str, int]:
 
-    old_stderr = sys.stderr
-    new_stderr = StringIO()
-    sys.stderr = new_stderr
+    stdout = StringIO()
+    stderr = StringIO()
 
     try:
-        f()
+        f(stdout, stderr)
         exit_status = 0
     except SystemExit as system_exit:
         exit_status = system_exit.code
-    finally:
-        sys.stdout = old_stdout
-        sys.stderr = old_stderr
 
-    return new_stdout.getvalue(), new_stderr.getvalue(), exit_status
+    return stdout.getvalue(), stderr.getvalue(), exit_status
 
 
 def run(args: List[str]) -> Tuple[str, str, int]:
     # Lazy import to avoid needing to import all of mypy to call run_dmypy
     from mypy.main import main
-    return _run(lambda: main(None, args=args))
+    return _run(lambda stdout, stderr: main(None, args=args,
+                                            stdout=stdout, stderr=stderr))
 
 
 def run_dmypy(args: List[str]) -> Tuple[str, str, int]:
     from mypy.dmypy import main
-    return _run(lambda: main(args))
+    return _run(lambda stdout, stderr: main(args))
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -128,8 +128,8 @@ def build(sources: List[BuildSource],
           alt_lib_path: Optional[str] = None,
           flush_errors: Optional[Callable[[List[str], bool], None]] = None,
           fscache: Optional[FileSystemCache] = None,
-          stdout: TextIO = sys.stdout,
-          stderr: TextIO = sys.stderr,
+          stdout: Optional[TextIO] = None,
+          stderr: Optional[TextIO] = None,
           ) -> BuildResult:
     """Analyze a program.
 
@@ -161,6 +161,8 @@ def build(sources: List[BuildSource],
         messages.extend(new_messages)
 
     flush_errors = flush_errors or default_flush_errors
+    stdout = stdout or sys.stdout
+    stderr = stderr or sys.stderr
 
     try:
         result = _build(sources, options, alt_lib_path, flush_errors, fscache, stdout, stderr)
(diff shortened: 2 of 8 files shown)


gvanrossum:
Sadly this is not enough. When `-v` is specified, `build.py` writes directly to `sys.stderr`. There are probably also other places where `sys.stderr` is written to directly. (And possibly even places where `sys.stdout` is implicitly used via `print()` calls without `file=...`.)

For example, I have this simple test program:
```py
import mypy.api

stdout, stderr, exitcode = mypy.api.run(['-v', '_.py'])
print('=== stdout ===')
print(stdout)
print('=== stderr ===')
print(stderr)
print('=== exit %d ===' % exitcode)
```
Note that without your patch, the LOG entries are printed under the `=== stderr ===` heading. But with your patch, they precede the `=== stdout ===` header.

gvanrossum:
PS. While a PR is under review, please don't force-push. This makes the life difficult for the reviewer. We always squash commits when we merge them into master.

Is this now ready for another review?

AnOctopus:
I realized the issue with it when I read through the contributing guidelines a bit more thoroughly and saw it said to not squash them on my end. I believe it's ready for review. It's possible that I missed something in my fixes or that things were messed up when I rebased to the current master, but I think I got all the code reachable from the api, and the simple test program you provided has the expected output.

mypy/api.py
```diff
@@ -44,36 +44,31 @@
...
-from typing import List, Tuple, Callable
+from typing import List, Tuple, Union, TextIO, Callable
+from mypy_extensions import DefaultArg
 
 
-def _run(f: Callable[[], None]) -> Tuple[str, str, int]:
-    old_stdout = sys.stdout
-    new_stdout = StringIO()
-    sys.stdout = new_stdout
+def _run(f: Callable[[TextIO, TextIO], None]) -> Tuple[str, str, int]:
```
gvanrossum: While you're at it, maybe rename `f` to something more memorable? E.g. `main_wrapper` or `call_main` or something?

mypy/build.py
```diff
@@ -128,6 +128,8 @@ def build(sources: List[BuildSource],
           alt_lib_path: Optional[str] = None,
           flush_errors: Optional[Callable[[List[str], bool], None]] = None,
           fscache: Optional[FileSystemCache] = None,
+          stdout: TextIO = sys.stdout,
+          stderr: TextIO = sys.stderr,
```
gvanrossum: This fixes the stdout/stderr streams to whatever they are when the module is imported. That doesn't sound ideal. Maybe leave them `None` and dynamically choose, e.g. `file=(stdout or sys.stdout)`. (Note that `print()` lets you say `print(..., file=None)` and it will use sys.stdout. There's no such shortcut for stderr, but it seems it's used less.) Of course a bunch of types will need to be changed from `TextIO` to `Optional[TextIO]`.

Alternatively, if there's a single "toplevel" place where stdout/stderr are passed in, you could set the default to `None` only there, and add `if not stdout: stdout = stdout` and ditto for `stderr`.

(Basically, there are two options -- turn `None` into `sys.stdout` as early as possible (but not at module load time, i.e. not as a default value), or do it as late as possible (in the `print()` calls). Your choice.

mypy/build.py
```diff
@@ -319,7 +325,11 @@ def import_priority(imp: ImportBase, toplevel_priority: int) -> int:
     return toplevel_priority
 
 
-def load_plugins(options: Options, errors: Errors) -> Tuple[Plugin, Dict[str, str]]:
+def load_plugins(
+        options: Options,
+        errors: Errors,
+        stdout: TextIO = sys.stdout
```
gvanrossum: Same comment as above -- or perhaps it's possible to make this a mandatory parameter?

Also, add a trailing comma.

mypy/build.py
```diff
@@ -904,7 +918,9 @@ def read_plugins_snapshot(manager: BuildManager) -> Optional[Dict[str, str]]:
     return snapshot
 
 
-def read_quickstart_file(options: Options) -> Optional[Dict[str, Tuple[float, int, str]]]:
+def read_quickstart_file(options: Options,
+                         stdout: TextIO = sys.stdout
```
gvanrossum: DItto on both counts. :-)

mypy/build.py
```diff
@@ -2429,7 +2446,10 @@ def log_configuration(manager: BuildManager) -> None:
 # The driver
 
 
-def dispatch(sources: List[BuildSource], manager: BuildManager) -> Graph:
+def dispatch(sources: List[BuildSource],
+             manager: BuildManager,
+             stdout: TextIO = sys.stdout
```
gvanrossum: And again.

mypy/errors.py
```diff
@@ -585,8 +585,14 @@ def remove_path_prefix(path: str, prefix: Optional[str]) -> str:
...
 
-def report_internal_error(err: Exception, file: Optional[str], line: int,
-                          errors: Errors, options: Options) -> None:
+def report_internal_error(err: Exception,
+                          file: Optional[str],
+                          line: int,
+                          errors: Errors,
+                          options: Options,
+                          stdout: TextIO = sys.stdout,
+                          stderr: TextIO = sys.stderr
```
gvanrossum: Again. :-)

mypy/errors.py
```diff
@@ -597,7 +603,7 @@ def report_internal_error(err: Exception, file: Optional[str], line: int,
         for msg in errors.new_messages():
             print(msg)
     except Exception as e:
-        print("Failed to dump errors:", repr(e), file=sys.stderr)
+        print("Failed to dump errors:", repr(e), file=stderr)
```
gvanrossum: Note that if you change the default to `None` above, you have to be explicit here, e.g. `file=(stderr or sys.stderr)`.

mypy/main.py
```diff
@@ -46,7 +47,9 @@ def stat_proxy(path: str) -> os.stat_result:
         return st
 
 
-def main(script_path: Optional[str], args: Optional[List[str]] = None) -> None:
+def main(script_path: Optional[str], args: Optional[List[str]] = None,
+         stdout: TextIO = sys.stdout,
+         stderr: TextIO = sys.stderr) -> None:
```
gvanrossum: Again... (Maybe this is the toplevel where you should check for `None`?)

mypy/mypyc_hacks.py
```diff
@@ -1,24 +1,26 @@
...
 
-from typing import Dict, Any
+from typing import Dict, Any, TextIO
 import sys
 
 
 # Extracted from build.py because we can't handle *args righit
 class BuildManagerBase:
-    def __init__(self) -> None:
+    def __init__(self, stdout: TextIO, stderr: TextIO) -> None:
```
gvanrossum: I like how here it's a mandatory argument, no default, so no premature locking in of `sys.stdout/stderr`.

mypy/test/testgraph.py
```diff
@@ -42,6 +44,8 @@ def _make_manager(self) -> BuildManager:
         options = Options()
         fscache = FileSystemCache()
         search_paths = SearchPaths((), (), (), ())
+        stdout = StringIO()
+        stderr = StringIO()
```
gvanrossum: This essentially silences the output from the test, which may not be what's desired. (At least I don't see any code that captures the contents of `stdout/stderr`.) Why not use `sys.stdout/stderr`?

AnOctopus: I think this was accidentally left over from my initial attempt to make the arguments mandatory and change all the callsites before I noticed that was so much stuff. I'll remove it.

mypy/main.py
AnOctopus: This is not quite the top level, because a fair few things call `build` directly (mostly tests but adding this parameter they don't care about just so they keep working exactly as they currently do seems dumb). I tried doing the `None` checking here but mypy couldn't infer that `f` in `flush_errors` couldn't be `None` after that, so I'm having to pass in actual values in `__main__.py` and instead do the checking in `build`

PR #6753, opened 2019-05-01
Fix class/def line reporting and ignoring for 3.8+.

### LINKED ISSUE #3871: Function line number information is not always correct
Currently, the line number of the function definition is not correct leading to issues such as #3869.

In the Python ast (as well as type-ast) the `FunctionDef` node we use to, for example, report missing return annotations gives a line number of the beginning of the first decorator (if any). Therefore it can incorrectly report the line number information we need to report errors.

Our currently logic increments the line number of any decorators, however, we cannot depend on this to be consistent. Consider this problematic (but completely valid!) example:

```python
@decorate(
        "multi",
        "line"
        )
def foo(
        a: int,
        b: str,
        ):

        pass
```

The `FunctionDef.lineno` would be 1, so mypy will report `test.py:2: error: Function is missing a return type annotation`. However, the return should be on line 8! Note that blanks space between the body and end of the function definition node means we cannot decrement based on the first item in the body of the function.

As a first possible solution, we could modify `typed_ast` to give the line number of the `returns` if the function is not typed (currently it just puts the attribute as None), so that the information is known, but it clearly indicates that the node is empty. This is certainly not an ideal solution.

Corresponding issue opened in https://github.com/python/typed_ast/issues/50

brandtbucher:
Fixes #3871. This change fixes a few things concerning class/function definition line numbers.

For users running mypy with Python 3.8+:
- Function definitions are now reported on the correct line, rather than the line of the first decorator plus the number of decorators.
- Class definitions are now reported on the correct line, rather than the line of the first decorator.
- **These changes are fully backward compatible with respect to `# type: ignore` lines**. In other words, existing comments will _not_ need to be moved. This is accomplished by defining an ignore "scope" just like we added to expressions with #6648. See the recent discussion at #3871 for reasoning.

For other users:
- Class definition lines are reported by using the same decorator-offset estimate as functions. This is both more accurate, and necessary to get 15 or so tests to pass for both pre- and post-3.8 versions. This seems fine, since there [doesn't appear](https://github.com/python/mypy/pull/6539#discussion_r265376127) to be a compelling reason why it was different in the first place.
- **This change is also backward-compatible with existing ignores, as above.**

About 15 tests had to have their error messages / nodes moved down one line, and 4 new regression tests have been added to `check-38.test`.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -522,13 +522,8 @@ class ASTConverter:
                 lineno += len(n.decorator_list)
                 end_lineno = None
             else:
-                # It's okay that end_lineno is *before* lineno here.
-                # lineno is where the error is reported, and end_lineno defines
-                # the opposite end of the range to search for "# type: ignore"
-                # comments in. This reports the correct line in the event of an
-                # error, but keeps old pre-3.8 "# type: ignore" comments working
-                # correctly.
-                lineno = n.lineno
+                # Set end_lineno to the old pre-3.8 lineno, in order to keep
+                # existing "# type: ignore" comments working:
                 end_lineno = n.decorator_list[0].lineno + len(n.decorator_list)
 
             var = Var(func_def.name())
@@ -639,11 +634,8 @@ class ASTConverter:
                         metaclass=dict(keywords).get('metaclass'),
                         keywords=keywords)
         cdef.decorators = self.translate_expr_list(n.decorator_list)
-        # It's okay if end_lineno is *before* lineno here.
-        # lineno is where the error is reported, and end_lineno defines the
-        # opposite end of the range to search for "# type: ignore" comments in.
-        # This reports the correct line in the event of an error, but keeps old
-        # pre-3.8 "# type: ignore" comments working correctly.
+        # Set end_lineno to the old mypy 0.700 lineno, in order to keep
+        # existing "# type: ignore" comments working:
         if sys.version_info < (3, 8):
             cdef.line = n.lineno + len(n.decorator_list)
             cdef.end_line = n.lineno
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -518,13 +518,23 @@ class ASTConverter:
                 # Before 3.8, [typed_]ast the line number points to the first decorator.
                 # In 3.8, it points to the 'def' line, where we want it.
                 lineno += len(n.decorator_list)
+                end_lineno = None
+            else:
+                # It's okay that end_lineno is *before* lineno here.
+                # lineno is where the error is reported, and end_lineno defines
+                # the opposite end of the range to search for "# type: ignore"
+                # comments in. This reports the correct line in the event of an
+                # error, but keeps old pre-3.8 "# type: ignore" comments working
+                # correctly.
+                lineno = n.lineno
+                end_lineno = n.decorator_list[0].lineno + len(n.decorator_list)
 
             var = Var(func_def.name())
             var.is_ready = False
             var.set_line(lineno)
 
             func_def.is_decorated = True
-            func_def.set_line(lineno, n.col_offset)
+            func_def.set_line(lineno, n.col_offset, end_lineno)
             func_def.body.set_line(lineno)  # TODO: Why?
 
             deco = Decorator(func_def, self.translate_expr_list(n.decorator_list), var)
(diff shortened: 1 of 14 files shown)


brandtbucher:
Bump. Hopefully this doesn't get too stale from all of the new sprint PRs!

gvanrossum:
Alas, it's already stale. Could you merge the conflicts?

brandtbucher:
@gvanrossum Merged, tests are passing.

mypy/fastparse.py
```diff
@@ -520,13 +520,23 @@ def do_func_def(self, n: Union[ast3.FunctionDef, ast3.AsyncFunctionDef],
...
                 # In 3.8, it points to the 'def' line, where we want it.
                 lineno += len(n.decorator_list)
+                end_lineno = None
+            else:
+                # It's okay that end_lineno is *before* lineno here.
+                # lineno is where the error is reported, and end_lineno defines
+                # the opposite end of the range to search for "# type: ignore"
+                # comments in. This reports the correct line in the event of an
+                # error, but keeps old pre-3.8 "# type: ignore" comments working
+                # correctly.
```
gvanrossum: This comment confused me. The key thing to note here is that we want to point `lineno` to the line that has the `def` keyword, while we want to point `end_lineno` to the line that would have been used for Python 3.7 and before. And we must compute `end_lineno` (i.e. the "old start line") using the crufty way given here since the old way of computing it (i.e. `n.lineno + len(n.decorator_list)` won't have the same result in 3.8.

Sorry that this is still so confusing. The code is right!

mypy/fastparse.py
```diff
@@ -520,13 +520,23 @@ def do_func_def(self, n: Union[ast3.FunctionDef, ast3.AsyncFunctionDef],
...
+                lineno = n.lineno
```
gvanrossum: Do you even need this? It seems this was already set on line 432 above.

mypy/fastparse.py
```diff
@@ -629,15 +639,18 @@ def visit_ClassDef(self, n: ast3.ClassDef) -> ClassDef:
...
-            # 3.8, it points to the 'class' statement.  We always make
-            # it point to the first decorator.  (The node structure
-            # here is different than for decorated functions.)
-            cdef.line = n.decorator_list[0].lineno
-            cdef.column = n.col_offset
+        # It's okay if end_lineno is *before* lineno here.
+        # lineno is where the error is reported, and end_lineno defines the
+        # opposite end of the range to search for "# type: ignore" comments in.
+        # This reports the correct line in the event of an error, but keeps old
+        # pre-3.8 "# type: ignore" comments working correctly.
```
gvanrossum: Similar comment (except I wasn't really confused here, since I had just figured out the previous one :-).

test-data/unit/check-attr.test
```diff
@@ -74,8 +74,8 @@ A(1, [2], '3', 4, 5)  # E: Too many arguments for "A"
 [case testAttrsUntypedNoUntypedDefs]
 # flags: --disallow-untyped-defs
 import attr
-@attr.s    # E: Function is missing a type annotation for one or more arguments
-class A:
+@attr.s
+class A:  # E: Function is missing a type annotation for one or more arguments
```
gvanrossum: Hm... Do I understand correctly that this is a backwards incompatibility? If the user had a `# type: ignore` on the decorator it would no longer work, right? I can't even figure out which part of this PR moved the line where the error occurs. :-(

(FWIW I agree that this is a better place for the error to occur. But I am concerned about compatibility.)

brandtbucher: It shouldn't be a backwards incompatibility. Even though the `lineno` of the error is different (648 and 651 of `fastparse.py`, since you were curious), we use the same `end_lineno` trick that's used for functions to respect `# type: ignore` comments occurring at the old location, _even for versions prior to 3.8_. This is done in `fastparse.py`, line 649 (for < 3.8) and line 652 (for >= 3.8).

mypy/fastparse.py
brandtbucher: Nice catch. I believe that's left over from an earlier iteration of this change.

mypy/fastparse.py
gvanrossum: Are you going to update these comments? I hope so!

brandtbucher: I'll try to reword this to be more concise. Maybe something like:

```py
# Set end_lineno to the old pre-3.8 lineno, in order to keep
# existing "# type: ignore" comments working:
```

mypy/fastparse.py
brandtbucher: As above:

```py
# Set end_lineno to the old mypy 0.700 lineno, in order to keep
# existing "# type: ignore" comments working:
```

mypy/fastparse.py
gvanrossum: Sounds good -- also for the other location!

brandtbucher:
New changes are passing, ready for another review.

gvanrossum:
Thanks! Congratulations on getting this thorny PR in. Excellent work.

brandtbucher:
Thanks, that really means a lot coming from you. Always happy to help!

PR #6756, opened 2019-05-02
Fix the mypy self check under 3.8

msullivan:


mypy/fastparse.py
```diff
@@ -1384,7 +1384,12 @@ def visit_Str(self, n: Str) -> Type:
         # If we're analyzing Python 3, that function will translate 'builtins.unicode'
         # into 'builtins.str'. In contrast, if we're analyzing Python 2 code, we'll
         # translate 'builtins.bytes' in the method below into 'builtins.str'.
-        if 'u' in n.kind or self.assume_str_is_unicode:
+
+        # Do an ignore because the field doesn't exist in 3.8 (where
+        # this function doesn't actually ever run.)
```
ilevkivskyi:
```suggestion
        # this method doesn't actually ever run.)
```
(this is extremely minor)

PR #6763, opened 2019-05-04
Fix incorrect tracking of "final" Instances

Michael0x2a:
This diff makes three changes: it fixes a bug where we incorrectly track "final" Instances, does some related refactoring, and finally modifies tuple indexing to be aware of literal contexts.

Specifically, here is an example of the bug. Note that mypy ignores the mutable nature of `bar`:

    def expect_3(x: Literal[3]) -> None: ...

    foo: Final = 3
    bar = foo
    for i in range(10):
        bar = i

    # Currently type-check; this PR makes mypy correctly report an error
    expect_3(bar)

To fix this bug, I decided to adjust the variable assignment logic: if the variable is non-final, we now scan the inferred type we try assigning and recursively erase all set `instance.final_value` fields.

This change ended up making the `in_final_declaration` field redundant -- after all, we're going to be actively erasing types on non-final assignments anyways. So, I decided to just remove this field.

I suspect this change will also result in some nice dividends down the road: defaulting to preserving the underlying literal when inferring expression types will probably make it easier to add more sophisticated
literal-related inference down the road.

In the process of implementing the above two, I discovered that "nested" Instance types are effectively ignored. So, the following program does not type check, despite the `Final` and despite that tuples are
immutable:

    bar: Final = (3, 2, 1)

    # 'bar[0] == 3' is always true, but we currently report an error
    expect_3(bar[0])

This is mildly annoying, and also made it slightly harder for me to verify my changes above, so I decided to modify `visit_index_expr` to also examine the literal context.

(Actually, I found I could move this check directly into the 'accept' method instead of special-casing things within `visit_index_expr` and `analyze_var_ref`. But I decided against this approach: the special-casing feels less intrusive, easier to audit, and slightly more efficient.)

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -53,7 +53,7 @@ from mypy.maptype import map_instance_to_supertype
 from mypy.typevars import fill_typevars, has_no_typevars, fill_typevars_with_any
 from mypy.semanal import set_callable_name, refers_to_fullname
 from mypy.mro import calculate_mro
-from mypy.erasetype import erase_typevars, remove_literal_metadata
+from mypy.erasetype import erase_typevars, remove_instance_last_known_values
 from mypy.expandtype import expand_type, expand_type_by_instance
 from mypy.visitor import NodeVisitor
 from mypy.join import join_types
@@ -1870,7 +1870,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             if inferred:
                 rvalue_type = self.expr_checker.accept(rvalue)
                 if not inferred.is_final:
-                    rvalue_type = remove_literal_metadata(rvalue_type)
+                    rvalue_type = remove_instance_last_known_values(rvalue_type)
                 self.infer_variable_type(inferred, lvalue, rvalue_type, rvalue)
 
     def check_compatibility_all_supers(self, lvalue: RefExpr, lvalue_type: Optional[Type],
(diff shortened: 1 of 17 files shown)


Michael0x2a:
@ilevkivskyi -- I renamed the field as suggested.

One complication of the rename is that it potentially invalidates any existing mypy caches -- we now look for the "last_known_value" entry in the cache, instead of "final_value".

Maybe this is too disruptive, especially during the pycon sprints? If you want, I can partially revert the changes I made to the Instance deserialization/serialization methods so they continue using "final_value" and submit a PR for those two methods closer to the next mypy release.

Regarding invariance -- I don't think it'll cause an issue because we always erase this field when binding Instances to a TypeVar. I think the existing [`testLiteralFinalPropagatesThroughGenerics`](https://github.com/python/mypy/blob/master/test-data/unit/check-literal.test#L2570) test checks for this?

I also added `testLiteralWithFinalPropagationIsNotLeaking` which checks this for lists, dicts, and sets, and modified `testLiteralFinalCollectionPropagation` just now to do something similar.

Or maybe you were thinking of something else regarding invariance?

Regarding mypyc -- I'll start work on a PR to make any necessary changes now.

ilevkivskyi:
> One complication of the rename is that it potentially invalidates any existing mypy caches -- we now look for the "last_known_value" entry in the cache, instead of "final_value".

I don't think it is something worth worrying about, mypy normally ignores the cache from another mypy version anyway.
> Or maybe you were thinking of something else regarding invariance

Nothing specific, just wanted to double-check that we have a good coverage for this.
> Regarding mypyc -- I'll start work on a PR to make any necessary changes now.

OK, I think this can be merged now, because mypyc uses a pinned mypy commit hash. So your PR to mypyc should combine a pin move plus the related updates to mypyc code.

Issue #4547, opened 2018-02-07
Base class attribute type does not provide type context in subclass


JukkaL:
In this example there is an unnecessary message about needing a type annotation:

```py
from typing import List

class A:
    x: List[str]

class B(A):
    x = []  # Need type annotation
```

Here we should we able to use the base class type for `x` as type context for the `x` initializer.

This is closely related to #3208.

ilevkivskyi:
https://github.com/python/mypy/issues/6511 provides another use case for this.

ashb:
Minimal repro case:

```python
from typing import Iterable

class A:
    a: Iterable[str] = ("a", )

class B(A):
    a = ("a", "b")

class C(B):
    a = ("a", "b", "c")
```

And the error:

> ```a.py:10: error: Incompatible types in assignment (expression has type "Tuple[str, str, str]", base class "B" defined the type as "Tuple[str, str]")```

The error is reported on the assignment in `C`.

I would have expected that B.a would have the same type as A.a since B.a doesn't have an explicit type.

(This is the case from #6511 copied here in case it is useful)

ilevkivskyi:
(Raising priority to high since this keeps coming.)

onlined:
This example from https://github.com/python/mypy/pull/6766#discussion_r283309925, which is related to partial list type, doesn't give an error while it should:
```py
class A:
    x = ['a', 'b']

class B(A):
    x = []
    x.append(2)
```

ilevkivskyi:
It looks like the situation is even worse here. Here we have no partial types and we still infer wrong type (the one that violates Liskov):
```python
from typing import List, Optional

class B:
    x: List[Optional[int]]

class C(B):
    x = [1]

reveal_type(C().x)  # Revealed type is "builtins.list[builtins.int]"
```
The problem is that we use supertype as type context when checking compatibility. But then again when storing the variable type we use empty context.

PR #6766, opened 2019-05-05
Don't crash when partial types are used in inherited attribute

### LINKED ISSUE #4552: Crash related to partial type in subclass
Mypy crashes when type checking this program:

```py
class A:
    x: str

class B(A):
    x = None
    x = ''
```

Here's the traceback:

```
  File "/Users/jukka/src/mypy/scripts/mypy", line 6, in <module>
    main(__file__)
  File "/Users/jukka/src/mypy/mypy/main.py", line 80, in main
    type_check_only(sources, bin_dir, options, flush_errors)
  File "/Users/jukka/src/mypy/mypy/main.py", line 129, in type_check_only
    flush_errors=flush_errors)
  File "/Users/jukka/src/mypy/mypy/build.py", line 173, in build
    result = _build(sources, options, alt_lib_path, bin_dir, saved_cache, flush_errors)
  File "/Users/jukka/src/mypy/mypy/build.py", line 259, in _build
  File "/Users/jukka/src/mypy/mypy/build.py", line 2094, in dispatch
    process_graph(graph, manager)
  File "/Users/jukka/src/mypy/mypy/build.py", line 2395, in process_graph
  File "/Users/jukka/src/mypy/mypy/build.py", line 2579, in process_stale_scc
  File "/Users/jukka/src/mypy/mypy/build.py", line 1951, in type_check_first_pass
    self.type_checker().check_first_pass()
  File "/Users/jukka/src/mypy/mypy/checker.py", line 229, in check_first_pass
    self.accept(d)
  File "/Users/jukka/src/mypy/mypy/checker.py", line 324, in accept
    stmt.accept(self)
  File "/Users/jukka/src/mypy/mypy/nodes.py", line 704, in accept
  File "/Users/jukka/src/mypy/mypy/checker.py", line 1252, in visit_class_def
    self.accept(defn.defs)
  File "/Users/jukka/src/mypy/mypy/checker.py", line 324, in accept
    stmt.accept(self)
  File "/Users/jukka/src/mypy/mypy/nodes.py", line 766, in accept
  File "/Users/jukka/src/mypy/mypy/checker.py", line 1372, in visit_block
    self.accept(s)
  File "/Users/jukka/src/mypy/mypy/checker.py", line 324, in accept
    stmt.accept(self)
  File "/Users/jukka/src/mypy/mypy/nodes.py", line 810, in accept
  File "/Users/jukka/src/mypy/mypy/checker.py", line 1379, in visit_assignment_stmt
  File "/Users/jukka/src/mypy/mypy/checker.py", line 1412, in check_assignment
    if self.check_compatibility_all_supers(lvalue, lvalue_type, rvalue):
  File "/Users/jukka/src/mypy/mypy/checker.py", line 1513, in check_compatibility_all_supers
    base_node):
  File "/Users/jukka/src/mypy/mypy/checker.py", line 1571, in check_compatibility_super
    'base class "%s" defined the type as' % base.name())
  File "/Users/jukka/src/mypy/mypy/checker.py", line 2961, in check_subtype
    if is_subtype(subtype, supertype):
  File "/Users/jukka/src/mypy/mypy/subtypes.py", line 83, in is_subtype
    ignore_declared_variance=ignore_declared_variance))
  File "/Users/jukka/src/mypy/mypy/types.py", line 1292, in accept
    return visitor.visit_partial_type(self)
  File "/Users/jukka/src/mypy/mypy/subtypes.py", line 349, in visit_partial_type
    raise RuntimeError
RuntimeError:
```

We can fix this issue by checking the compatibility of attributes with base classes at the end of type checking a class.

onlined:
Fix #4552.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1786,7 +1786,6 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                                                       infer_lvalue_type)
         else:
             lvalue_type, index_lvalue, inferred = self.check_lvalue(lvalue)
-
             # If we're assigning to __getattr__ or similar methods, check that the signature is
             # valid.
             if isinstance(lvalue, NameExpr) and lvalue.node:
@@ -1803,7 +1802,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                         else:
                             self.check_getattr_method(signature, lvalue, name)
 
-            if isinstance(lvalue, RefExpr):
+            # Defer PartialType's super type checking.
+            if isinstance(lvalue, RefExpr) and not isinstance(lvalue_type, PartialType):
                 if self.check_compatibility_all_supers(lvalue, lvalue_type, rvalue):
                     # We hit an error on this line; don't check for any others
                     return
@@ -1833,6 +1833,11 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                         # Try to infer a partial type. No need to check the return value, as
                         # an error will be reported elsewhere.
                         self.infer_partial_type(lvalue_type.var, lvalue, rvalue_type)
+                    # Handle PartialType's super type checking here, after it's resolved.
+                    if (isinstance(lvalue, RefExpr) and
+                            self.check_compatibility_all_supers(lvalue, lvalue_type, rvalue)):
+                        # We hit an error on this line; don't check for any others
+                        return
                 elif (is_literal_none(rvalue) and
                         isinstance(lvalue, NameExpr) and
                         isinstance(lvalue.node, Var) and
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1803,7 +1803,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                             self.check_getattr_method(signature, lvalue, name)
 
             # Defer PartialType's super type checking.
-            if isinstance(lvalue, RefExpr) and not isinstance(lvalue_type, PartialType):
+            if (isinstance(lvalue, RefExpr) and
+                    not (isinstance(lvalue_type, PartialType) and lvalue_type.type is None)):
                 if self.check_compatibility_all_supers(lvalue, lvalue_type, rvalue):
                     # We hit an error on this line; don't check for any others
                     return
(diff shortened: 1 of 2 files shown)


test-data/unit/check-inference.test
```diff
@@ -2689,3 +2689,13 @@ reveal_type(x)  # E: Revealed type is 'builtins.list[Any]'
 reveal_type(y)  # E: Revealed type is 'builtins.dict[Any, Any]'
 
 [builtins fixtures/dict.pyi]
+
+[case testInheritedAttributeNoStrictOptional]
+# flags: --no-strict-optional
```
ilevkivskyi: I would also add a test case with `--strict-optional` as well (to avoid regressions).

test-data/unit/check-inference.test
```diff
@@ -2689,3 +2689,13 @@ reveal_type(x)  # E: Revealed type is 'builtins.list[Any]'
...
+class A:
+    x: str
+
+class B(A):
+    x = None
+    x = ''
```
ilevkivskyi: Could you please also add a test where this causes an incompatible override error (an `int`)?

test-data/unit/check-inference.test
```diff
@@ -2689,3 +2689,13 @@ reveal_type(x)  # E: Revealed type is 'builtins.list[Any]'
...
+    reveal_type(x)  # E: Revealed type is 'builtins.str'
```
ilevkivskyi: No new line at the end of file.

mypy/checker.py
```diff
@@ -1833,6 +1833,11 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
                         # Try to infer a partial type. No need to check the return value, as
                         # an error will be reported elsewhere.
                         self.infer_partial_type(lvalue_type.var, lvalue, rvalue_type)
+                    # Handle PartialType's super type checking here, after it's resolved.
+                    if (isinstance(lvalue, RefExpr) and
+                            self.check_compatibility_all_supers(lvalue, lvalue_type, rvalue)):
```
ilevkivskyi: The exception you added above removes the check for all kinds of partial types, while this only re-adds the check for partial `None` types. I don't say it is necessarily wrong, but I would at least add test cases with partial list and dict types (e.g. with an empty list in subclass that is subsequently gets an `.append()`).

onlined: Yes, you are right. I will change the code, actually.

ilevkivskyi: Are you going to add the tests with an empty list in the subclass?

onlined: I don't think I should add them anymore, because now I make the exception for only `None` partial types.

ilevkivskyi: But there may be a similar crash (or a regression later), I think it is better to have one in any case.

onlined: I tried this on both master and this branch, it didn't give an error:

```py
from typing import List

class A:
    x = ['a', 'b']

class B(A):
    x = []
    x.append(2)
```

I think it should give an error, but I could not be sure. What do you think? @ilevkivskyi

ilevkivskyi: Yes, this should be an error. It looks like another consequence of https://github.com/python/mypy/issues/4547. Let's then merge this issue as is, and consider adding this test when https://github.com/python/mypy/issues/4547 is fixed. Could you please add a comment in that issue?

PR #6769, opened 2019-05-06
Don't add self argument if nested function in method 

### LINKED ISSUE #5806: No error for missing argument in type comment if function nested within method
There is no error reported for `g` even though the type comment is clearly missing an argument type:

```py
class A:
    def f(self): 
        # type: () -> None
        def g(x): 
            # type: () -> None  # No error
            pass

def f2(): 
    # type: () -> None
    def g2(x): 
        # type: () -> None  # Correctly reported
        pass
```

onlined:
Fix #5806.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -7,7 +7,7 @@ from typing import (
 MYPY = False
 if MYPY:
     import typing  # for typing.Type, which conflicts with types.Type
-    from typing_extensions import Final
+    from typing_extensions import Final, Literal
 
 from mypy.sharedparse import (
     special_function_elide_names, argument_elide_name,
@@ -249,6 +249,8 @@ class ASTConverter:
                  is_stub: bool,
                  errors: Errors) -> None:
         self.class_nesting = 0
+        # 'C' for class, 'F' for function
+        self.class_and_function_stack = []  # type: List[Literal['C', 'F']]
         self.imports = []  # type: List[ImportBase]
 
         self.options = options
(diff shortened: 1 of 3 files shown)


mypy/fastparse.py
```diff
@@ -513,6 +516,7 @@ def do_func_def(self, n: Union[ast3.FunctionDef, ast3.AsyncFunctionDef],
             func_type.definition = func_def
             func_type.line = lineno
 
+        retval: Union[FuncDef, Decorator]
```
gvanrossum: Because mypy needs to run with Python 3.5, you can't use this PEP-526 syntax here yet (which is 3.6 or later only). To fix this, use a type comment on the first line that assigns to `retval`.

onlined: Sorry, fixing now.

mypy/fastparse.py
```diff
@@ -249,6 +249,8 @@ def __init__(self,
                  is_stub: bool,
                  errors: Errors) -> None:
         self.class_nesting = 0
```
gvanrossum: Shouldn't you delete this variable? It's no longer updated.

test-data/unit/check-functions.test
```diff
@@ -737,6 +737,14 @@ class A:
 main:6: error: Incompatible types in assignment (expression has type "int", variable has type "A")
 main:8: error: Argument 1 to "g" has incompatible type "A"; expected "int"
 
+[case testNestedFunctionInMethodWithTooFewArgumentsInTypeComment]
```
gvanrossum: Perhaps you could also add a test for a few other configurations of the stack, e.g. a method inside a class inside a method (i.e. `['C', 'F', 'C', 'F']`)?

mypy/fastparse.py
onlined: Yes, I forgot it.

gvanrossum:
Thanks! Ready to be merged once the tests pass.

PR #6770, opened 2019-05-06
Include a note when user forgets to import something from the typing module

### LINKED ISSUE #4317: Add import hints for errors like 'Any' undefined
If new users forget to import `Any`, `List` or similar, it could be useful to add a hint such as this:

```
x.py:20: error: Name "Any" undefined
x.py:20: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Any")
```

We could do this for some common stdlib functionality as well, such as `collections`, `subprocess`, `re`, etc.

Even better, mypy could also give import hints for some user-defined classes and functions (that have unique definitions).

cs-cordero:
Fixes #4317

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -166,8 +166,6 @@ SUGGESTED_TEST_FIXTURES = {
     'builtins.classmethod': 'classmethod.pyi',
 }  # type: Final
 
-# Map from the lowercased name of a missing definition to a tuple containing
-# the name of the likely module definition and a suggested import string.
 TYPES_FOR_UNIMPORTED_HINTS = {
     'typing.Any',
     'typing.Callable',
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -169,18 +169,18 @@ SUGGESTED_TEST_FIXTURES = {
 # Map from the lowercased name of a missing definition to a tuple containing
 # the name of the likely module definition and a suggested import string.
 TYPES_FOR_UNIMPORTED_HINTS = {
-    'any': ('typing', 'from typing import Any'),
-    'callable': ('typing', 'from typing import Callable'),
-    'dict': ('typing', 'from typing import Dict'),
-    'iterable': ('typing', 'from typing import Iterable'),
-    'iterator': ('typing', 'from typing import Iterator'),
-    'list': ('typing', 'from typing import List'),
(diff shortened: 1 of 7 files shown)


onlined:
I think you should do the same thing in the new semantic analyzer.

mypy/semanal.py
```diff
@@ -3734,13 +3751,23 @@ def name_not_defined(self, name: str, ctx: Context) -> None:
...
             fullname = 'builtins.{}'.format(name)
             if self.lookup_fully_qualified_or_none(fullname) is None:
                 # Yes. Generate a helpful note.
                 self.add_fixture_note(fullname, ctx)
 
+        if name.lower() in TYPES_FOR_UNIMPORTED_HINTS:
+            # User probably forgot to import these types.
+            unimported_module, suggestion = TYPES_FOR_UNIMPORTED_HINTS[name.lower()]
+            hint = (
+                f'Did you forget to import it from "{unimported_module}"?'
```
JukkaL: Unfortunately we can't use f-strings since mypy still supports Python 3.5.

mypy/semanal.py
```diff
@@ -166,6 +166,23 @@
...
+    'dict': ('typing', 'from typing import Dict'),
+    'iterable': ('typing', 'from typing import Iterable'),
+    'iterator': ('typing', 'from typing import Iterator'),
+    'list': ('typing', 'from typing import List'),
+    'optional': ('typing', 'from typing import Optional'),
+    'set': ('typing', 'from typing import Set'),
+    'tuple': ('typing', 'from typing import Tuple'),
+    'typevar': ('typing', 'from typing import TypeVar'),
+    'union': ('typing', 'from typing import Union'),
+    'cast': ('typing', 'from typing import cast'),
```
JukkaL: Since all of these follow the same pattern, we could perhaps avoid some redundancy. One idea would be to just have a list/set of full names (`typing.Any`, `typing.Callable`, ...) and we can derive the other information (lower case version, `from typing import Foo`) from them.

mypy/semanal.py
cs-cordero: :+1:

mypy/semanal.py
cs-cordero: :+1:

PR #6771, opened 2019-05-06
Ignore virtualenv for commits and tests

### LINKED ISSUE #6768: Fix tests when development virtualenv is under mypy/
It would be good to allow having the dev/test virtualenv under the mypy repository, such as in `mypy/venv`. Currently at least lint fails in this case, and the virtualenv needs to outside the mypy repository.

viourr:
Fixes #6768

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,5 +35,11 @@ htmlcov
 # pytest cache
 .pytest_cache/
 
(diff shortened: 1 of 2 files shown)


gvanrossum:
Thanks! Works like a charm.

Issue #6741, opened 2019-04-29
When there is a duplicate module, show path of other module


JukkaL:
It's easy to accidentally have two modules with the same name if a `__init__.py` file missing, for example. Mypy currently complains about one of the modules but it may not be obvious where the other conflicting module lives. It would be better to show the paths of both/all conflicting modules. Maybe also include a hint about possible missing `__init__.py` files.

Example:

```
foo.py
pkg/__init__.py
pkg/subpkg/foo.py
```

Note that `pkg/subpkg/__init__.py` is missing, which causes mypy to think that there are two instances of module `foo`.

Users get confused about this quite frequently, so we should perhaps prioritize this.

jxcl:
Looking at this issue

PR #6772, opened 2019-05-06
Display both instances of duplicate modules in error message (#6741)

jxcl:


gvanrossum:
Thanks!

PS. Next time put "Fixes #XXXX" in the body of your commit message, don't put "(#XXXX)" in the first line -- this is confusing because GitHub adds a similar notation for the *PR*.

PR #6773, opened 2019-05-06
Improve message when function is missing return type

### LINKED ISSUE #5952: Improve message when function is missing return type
If a function takes no arguments (or only `self`) and `--disallow-untyped-defs` is enabled, the error message "Function is missing a type annotation" is not very helpful. Here's an idea for a better message:

```
foo.py:2: error: Function must have a return type annotation ('disallow_untyped_defs' enabled)
foo.py:2: note: Use "-> None" if function does not return a value
```

The error message could be further specialized for `__init__`.

See #5943 for a related issue.

allisonking:
Fixes #5952


mypy/checker.py
```diff
@@ -984,7 +984,12 @@ def is_unannotated_any(t: Type) -> bool:
         check_incomplete_defs = self.options.disallow_incomplete_defs and has_explicit_annotation
         if show_untyped and (self.options.disallow_untyped_defs or check_incomplete_defs):
             if fdef.type is None and self.options.disallow_untyped_defs:
-                self.fail(message_registry.FUNCTION_TYPE_EXPECTED, fdef)
+                if (not len(fdef.arguments) or
```
ilevkivskyi: You don'y need `len(...)` here, just `not fdef.arguments`.

mypy/checker.py
```diff
@@ -984,7 +984,12 @@ def is_unannotated_any(t: Type) -> bool:
...
+                        (len(fdef.arg_names) == 1 and 'self' in fdef.arg_names)):
```
ilevkivskyi: `fdef.arg_names[0] == 'self'` would be more straightforward. Also maybe you should check for both `self` and `cls`.

mypy/checker.py
```diff
@@ -984,7 +984,12 @@ def is_unannotated_any(t: Type) -> bool:
...
+                    self.fail(message_registry.RETURN_TYPE_EXPECTED_DISALLOW_UNTYPED, fdef)
+                    self.note('Use "-> None" if function does not return a value', fdef)
```
ilevkivskyi: I would only show this note if the function body doesn't have non-trivial return statements (i.e. `return x` not just plain `return`).

mypy/message_registry.py
```diff
@@ -74,6 +74,8 @@
 FUNCTION_TYPE_EXPECTED = "Function is missing a type annotation"  # type: Final
 ONLY_CLASS_APPLICATION = "Type application is only supported for generic classes"  # type: Final
 RETURN_TYPE_EXPECTED = "Function is missing a return type annotation"  # type: Final
+RETURN_TYPE_EXPECTED_DISALLOW_UNTYPED = \
+    "Function must have a return type annotation ('disallow_untyped_defs' enabled)"  # type: Final
```
ilevkivskyi: Although it is mentioned in the issue, I would not mention the name of option of the flag in the error message. It may easily get out of sync. So I would remove the part in `(...)`.

mypy/checker.py
gvanrossum: @JukkaL or @ilevkivskyi Is there a more reliable way to tell whether something's a method? Some (unusual) code bases use a different name instead of `self`, and class methods also aren't covered by this test.

It's also possible to scan the body to see whether occurrences of `return` have a value or not on them (though I don't know how practical that is from this context).

ilevkivskyi: I think for `self` there is actually one: `Var.is_self`. I am not sure we have the same for `cls`.

mypy/checker.py
allisonking: okay sounds good-- could you point me to how to find if there's a return value?

gvanrossum: You probably have to create a new `StatementVisitor` that traverses `fdef.body` looking for any `ReturnStmt` instances whose `expr` is not `None`.

(And don't take my word for it. :-)

ilevkivskyi: You can use `srubgen.has_return_statement()` (Maybe also move it to some common file like `nodes.py`?)

ilevkivskyi: (Fix: srubgen -> stubgen)

mypy/checker.py
allisonking: Thanks! I tried checking for `fdef.arguments[0].variable.is_self` here but it seems like I always get `False`. I couldn't find where `is_self` was being set-- for now I have this checking for the strings `self` and `cls` but agree it's not great

mypy/checker.py
allisonking: I took a stab at implementing this in `nodes.py`... let me know if it's reasonable! 😬

mypy/checker.py
ilevkivskyi: It may be because if a function is missing type annotation, then we don't analyze it at all during semantic analysis. I think it is OK to rely on the argument name, since this is needed only to tweak the error message, and _not_ to decide whether we need to emit the error.

mypy/checker.py
```diff
@@ -984,7 +984,14 @@ def is_unannotated_any(t: Type) -> bool:
         check_incomplete_defs = self.options.disallow_incomplete_defs and has_explicit_annotation
         if show_untyped and (self.options.disallow_untyped_defs or check_incomplete_defs):
             if fdef.type is None and self.options.disallow_untyped_defs:
-                self.fail(message_registry.FUNCTION_TYPE_EXPECTED, fdef)
+                if (not fdef.arguments or (len(fdef.arguments) == 1 and 
+                        #(fdef.arguments[0].variable.is_self or fdef.arg_names[0] == 'cls'))):
```
ilevkivskyi: This comment is not needed.

mypy/nodes.py
```diff
@@ -1101,6 +1111,19 @@ def __init__(self, expr: Optional[Expression]) -> None:
     def accept(self, visitor: StatementVisitor[T]) -> T:
         return visitor.visit_return_stmt(self)
 
+class ReturnSeeker(StatementVisitor):
```
ilevkivskyi: `StatementVisitor[None]`

mypy/nodes.py
```diff
@@ -1101,6 +1111,19 @@ def __init__(self, expr: Optional[Expression]) -> None:
...
+        self.found = False
+
+    def visit_statement(self, o: Statement) -> None:
+        if isinstance(o, ReturnStmt):
+            self.visit_return_stmt(o)
+
+    def visit_return_stmt(self, o: ReturnStmt) -> None:
+        if (o.expr is None or isinstance(o.expr, NameExpr) and o.expr.name == 'None'):
+            return
+        self.found = True
```
ilevkivskyi: This causes a bunch of code duplication with `stubgen.py`. Why don't you remove this class (and corresponding helper) from there and also use the added method there?

test-data/unit/cmdline.test
```diff
@@ -1063,7 +1063,8 @@ disallow_any_generics = True
 def get_tasks(self):
     return 'whatever'
 [out]
-a.py:1: error: Function is missing a type annotation
+a.py:1: error: Function is missing a return type annotation
+a.py:1: note: Use "-> None" if function does not return a value
```
ilevkivskyi: This should not be here because method returns a string (see CI failures).

mypy/stubgen.py
```diff
@@ -97,6 +97,7 @@ class Options:
 
     This class is mutable to simplify testing.
     """
+
```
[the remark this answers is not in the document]
allisonking: whoops-- turned off my autoformatter :)

mypy/nodes.py
```diff
@@ -2181,6 +2191,97 @@ def accept(self, visitor: ExpressionVisitor[T]) -> T:
...
+        pass
+
+    def visit_try_stmt(self, o: TryStmt) -> T:
+        pass
+
+    def visit_print_stmt(self, o: PrintStmt) -> None:
+        pass
+
+    def visit_exec_stmt(self, o: ExecStmt) -> None:
+        pass
```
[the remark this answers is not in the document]
allisonking: I was having trouble importing the `mypy.traverser.TraverserVisitor` class, I believe because `traverser.py` also imports from `nodes.py`?
```
mypy/test/testcheck.py:9: in <module>
    from mypy import build
mypy/build.py:36: in <module>
    from mypy.nodes import MypyFile, ImportBase, Import, ImportFrom, ImportAll, SymbolTable
mypy/nodes.py:20: in <module>
    from mypy.traverser import TraverserVisitor
mypy/traverser.py:4: in <module>
    from mypy.nodes import (
E   ImportError: cannot import name 'Block' from 'mypy.nodes' (mypy/nodes.py)
```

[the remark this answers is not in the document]
ilevkivskyi: Hm, yes, this would cause an import cycle. What if you instead keep this a function (not a method) and move it (together with the visitor class) to `mypy.traverser` and then import it from this common place in both cases (in `stubgen.py` and in `checker.py`)? I think this would be a reasonable solution.

Issue #6205, opened 2019-01-16
Show used config settings in verbose mode


ezyang:
Steps to reproduce:
1. Create `mypy.ini` with:
```
[mypy]
python_version = 2.7
```
2. Create a test file that uses Python 3 only type syntax, e.g.:
```
def f(x : int) -> int:
  return x
```
3. Run `python -mmypy test.py` where `python` in PATH is Python 3.

What I expected was for Python 3 to be used to typecheck the file. What actually happened was mypy somehow found a Python 2 executable on my system, and used that instead to typecheck the file. I thought this was unexpected, because I specifically ran mypy as `python -m mypy` so that it knew to use that python interpreter. Now, I know why this happened; because mypy is respecting the requested python version when it runs the typechecking. It would have been easier to debug if mypy by default emitted information about which Python it was using when it wasn't using the default.

Note: in the actual sequence of events, I didn't do (1), and didn't realize the file existed until I started digging.

ilevkivskyi:
Mypy never "selects" an interpreter to be interpreted (what would that even mean). Also the files are never executed by mypy (e.g. you don't need Python 2 interpreter to check Python 2 code).

However, mypy parses your code according to the syntax/grammar of Python version specified in the config file. And your example is not a valid Python 2.7.

ilevkivskyi:
Actually on the second thought, it seems to me we should show more info in verbose mode (currently it mostly useful for incremental mode).

ezyang:
I don't know too much about how mypy is implemented internally, so I was mostly guessing based on behavior. What I noticed is that if I run `python -mmypy test.py` it respects `python_version` in the config file, but when I run `python -mmypy --python-executable python3` it ignores `python_version` and parses the file as Python 3.

JelleZijlstra:
Yes, that's expected behavior; command-line flags override config file settings.

ezyang:
Sure. What's unexpected/confusing is the fact that they're different flags. `--python-executable` is documented to set `--python-version` if it is not already set, but a config file python-version doesn't count for th epurposes of determining "if it's already set."

gvanrossum:
Is that still true in mypy 0.660? AFAIK we rationalized the behavior here. Perhaps the docs need tweaking?

PR #6774, opened 2019-05-06
Correct --python-executable documentation

armooo:
Remove mention of --python-version being set based on
--python-executable. This functionality was removed in
d2a9f9d9e8111877621eb2d2a24104f9787ae553 and --python-executable is now
only used to find PEP 561 packages. This is a documentation fix for #6205.


armooo:
I am going to send another PR to add the parsed options to the trace output per the title of #6205. It seems that that would still be useful when debugging.

PR #6776, opened 2019-05-06
Output the parsed command line and file options

### LINKED ISSUE #6205: Show used config settings in verbose mode

armooo:
This is generally useful when debugging interactions between command line
flags and options in mypy.ini.

Fixes #6205.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -212,6 +212,7 @@ def _build(sources: List[BuildSource],
                            errors=errors,
                            flush_errors=flush_errors,
                            fscache=fscache)
+    manager.trace(repr(options))
 
     reset_global_state()
     try:


mypy/build.py
```diff
@@ -212,6 +212,7 @@ def _build(sources: List[BuildSource],
                            errors=errors,
                            flush_errors=flush_errors,
                            fscache=fscache)
+    manager.trace(options)
```
gvanrossum: Use `repr(options)` to avoid the test failure on self-type-check.

PR #6779, opened 2019-05-06
Ouput bad comment during syntax error

### LINKED ISSUE #4103: When there's a syntax error, print the comment used
Given the Python code:

```python
def big_function(
        x1, # type: Optional[int]
        x2, # type: Optional[int]
        x3, # type: Optional[int]
        x4, # type: Optional[int]
        x5, # type: Optional[int]]
        x6, # type: Optional[int]
        x7, # type: Optional[int]
):
    # type: (...) -> int
    return 123
```

Sometimes it's difficult to spot which comment has a syntax error. It would be nicer if mypy could print the comment it's looking at, to make it more obvious.

```
$ mypy --py2 bar.py
bar.py:1: error: syntax error in type comment
```

kcaebe:
fixes #4103

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -472,7 +472,9 @@ class ASTConverter:
                 if self.in_method_scope() and len(arg_types) < len(args):
                     arg_types.insert(0, AnyType(TypeOfAny.special_form))
             except SyntaxError:
-                self.fail(TYPE_COMMENT_SYNTAX_ERROR, lineno, n.col_offset)
+                stripped_type = n.type_comment.split("#", 2)[0].strip()
+                err_msg = "{} '{}'".format(TYPE_COMMENT_SYNTAX_ERROR, stripped_type)
+                self.fail(err_msg, lineno, n.col_offset)
                 if n.type_comment and n.type_comment[0] != "(":
                     self.note('Suggestion: wrap argument types in parentheses',
                               lineno, n.col_offset)
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -192,7 +192,8 @@ def parse_type_comment(type_comment: str,
         typ = ast3_parse(type_comment, '<type_comment>', 'eval')
     except SyntaxError as e:
         if errors is not None:
-            err_msg = "{} '{}'".format(TYPE_COMMENT_SYNTAX_ERROR, type_comment)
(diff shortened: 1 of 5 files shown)


ilevkivskyi:
You need to update existing tests. They are currently failing.

gvanrossum:
This is a good idea, but I'd like to see some tests (and the existing tests that are failing fixed). You also ought to add the same thing to `fastparse2.py`.

mypy/fastparse.py
```diff
@@ -192,7 +192,8 @@ def parse_type_comment(type_comment: str,
         typ = ast3_parse(type_comment, '<type_comment>', 'eval')
     except SyntaxError as e:
         if errors is not None:
-            errors.report(line, e.offset, TYPE_COMMENT_SYNTAX_ERROR, blocker=True)
+            err_msg = "{} '{}'".format(TYPE_COMMENT_SYNTAX_ERROR, type_comment)
```
ilevkivskyi: I would strip everything in the `type_comment` after `#`, so that only the interesting part is shown.

test-data/unit/parse-errors.test
```diff
@@ -113,25 +113,43 @@ file:1: error: invalid syntax
...
+file:2: error: syntax error in type comment ''
 
 [case testInvalidTypeComment4]
 0
 x = 0 # type: *
 [out]
-file:2: error: syntax error in type comment
+file:2: error: syntax error in type comment '*'
+
+[case testInvalidTypeComment5
```
ilevkivskyi: You have missing closing `]` here in in couple tests below.

mypy/fastparse2.py
```diff
@@ -337,6 +337,7 @@ def visit_FunctionDef(self, n: ast27.FunctionDef) -> Statement:
             return_type = None
         elif type_comment is not None and len(type_comment) > 0:
             try:
+                print(type_comment)
```
ilevkivskyi: Stray debugging print left.

ilevkivskyi:
Your last commit will likely cause many failures in CI (because at least some existing tests need to be updated). It is much easier to run tests locally first, since now the line in CI is really long, your PR might wait for an hour there.

kcaebe:
I wanted to get the function annotations in the same PR so I'm okay with it taking a bit. Thanks for all the help @ilevkivskyi.

Issue #6544, opened 2019-03-14
Verbose diagnostics can be improved


jolaf:
Even in verbose mode, `mypy` diagnostic output misses a lot of very useful things:
- Path to Python interpreter that is actually being used;
- Path to config file that is actually being used;
- Path to cache location that is actually being used.
- Actual path to files to check, gathered from `MYPYPATH` variable and config file.

Also for each file being checked, it would be useful to know:
- Full path to the file;
- Why `mypy` decided to check this particular file? Was it specified in the command line? By filename? By package name? By module name? Was it found in `MYPYPATH`? Was it `import`ed by some other file? Maybe there are other ways?
- If module is `import`ed, was it found in the current directory? In `MYPYPATH`? In `dist-packages` of the python interpreter?
- Is it a stub or an actual source file?

If file gets `import`ed, but import fails, it would be extremely useful to know, why particularly.
- Is module really not found?
- Is module found, but stubs are not?
- Maybe stubs were found, but they're incorrect?

I'm having serious trouble debugging my `mypy` configuration in CI/CD system, for example at some points `mypy` stops seeing my stubs I've created for a 3rd-party library I use, and than at some points starts seeing them again. The cause is eluding me for many hours, and tracking it is really difficult without detailed logging.

I understand most of my questions are silly, but I'm really having trouble finding answers to them, and by the way, verbose/debug mode IS really a way to get answers to a lot of silly questions that nobody thought somebody would ever need, but at some points someone does, isn't it?

msullivan:
There are a lot of good ideas here. Would you like to submit a PR implementing some of them?

jolaf:
Hardly, for now.
I'm just starting working with `mypy`, writing my first stubs etc.
I know nothing about the `mypy` internal structure.

jxcl:
Working on this issue

jxcl:
The first section has a PR now, the other parts are harder to implement.

SwagatSBhuyan:
Is this issue still open? Some of the later suggestions may not have been implemented due to complexity. I myself am really new to the mypy repository, but I'd like to help with the later implementations. It'd really be kind of you guys to maybe nudge me in the right direction @ilevkivskyi @akerami @jxcl @jolaf 
Thanks!

ilevkivskyi:
TBH I kind of lost track of what is the situation here. But other people can probably help.

CarlFK:
There are 11 bullet points in the OP.  
[...]

PR #6780, opened 2019-05-06
Add startup debug information

jxcl:
Implements some ideas from #6544

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2402,15 +2402,16 @@ def log_configuration(manager: BuildManager) -> None:
     """Output useful configuration information to LOG and TRACE"""
 
     manager.log()
-    manager.log("Mypy version %s" % __version__)
-    manager.log("Python interpreter located at: %s" % manager.options.python_executable)
+    configuration_vars = (
+        ("Mypy Version", __version__),
+        ("Config File", (manager.options.config_file or "Default")),
+        ("Configured Executable", manager.options.python_executable),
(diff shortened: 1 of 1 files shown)


mypy/build.py
```diff
@@ -2398,12 +2398,38 @@ def skipping_ancestor(manager: BuildManager, id: str, path: str, ancestor_for: '
                           severity='note', only_once=True)
 
 
+def log_configuration(manager: BuildManager) -> None:
+    """Output useful configuration information to LOG and TRACE"""
+
+    manager.log()
+    manager.log("Mypy version %s" % __version__)
+    manager.log("Python interpreter located at: %s" % manager.options.python_executable)
```
gvanrossum: That's the *configured* interpreter. It may or may not match `sys.executable`. Maybe print both?

mypy/build.py
```diff
@@ -2398,12 +2398,38 @@ def skipping_ancestor(manager: BuildManager, id: str, path: str, ancestor_for: '
                           severity='note', only_once=True)
 
 
+def log_configuration(manager: BuildManager) -> None:
+    """Output useful configuration information to LOG and TRACE"""
+
+    manager.log()
+    manager.log("Mypy version %s" % __version__)
```
gvanrossum: I'd recommend aligning the various fields -- we do that also in `dump_stats()` near line 561.

gvanrossum: I'd add a `:` to the message (like the others).

PR #6781, opened 2019-05-06
Update third party modules

### LINKED ISSUE #6767: Update set of recognized third party modules
`mypy/moduleinfo.py` contains a hard-coded list of third party modules that mypy uses to generate better error messages in some cases. This list hasn't been updated in a while. Update the list based on up-to-date package popularity information.

matys18:
Fixes #6767 🔧

Updated the third party module list with the top 200 packages of the past 365 days from [here](https://hugovk.github.io/top-pypi-packages/). Not entirely sure how reliable the data is, however, the [old source](http://python3wos.appspot.com/) has been deprecated and has not been updated in a while.


mypy/moduleinfo.py
```diff
@@ -21,198 +21,205 @@
...
     'pip',
-    'virtualenv',
-    'boto',
-    'setuptools',
-    'dateutil',
-    'pytz',
+    'urllib3',
+    'six',
+    'botocore',
+    'python-dateutil',
```
JukkaL: This is the PyPI distribution name, but the actual package name is different, and the package name is required in this file. In this case the package name is `dateutil`. Unfortunately, some manual tweaking seems to be required to get the package name to import.

mypy/moduleinfo.py
```diff
@@ -21,198 +21,205 @@
...
+    'simplejson',
+    'boto3',
+    'pytz',
+    'numpy',
+    'markupsafe',
+    'jinja2',
+    'cffi',
+    'cryptography',
+    'protobuf',
+    'awscli-cwlogs',
```
JukkaL: This is another PyPI distribution name that isn't a real Python package name. There are some other examples below.

matys18:
@JukkaL My bad. I've updated the list to contain the actual package names instead.

PR #6782, opened 2019-05-06
Improve message when an annotation is expected

### LINKED ISSUE #2463: Improve message when an annotation is expected
In cases like this, the current error message isn't as helpful as it could be:

```
x = Queue()  # Need type annotation for variable
```

Here's an idea for a better message:

```
x = Queue()  # Need type annotation for variable (hint: # type: Queue[<type>])
x = []  # Need type annotation for variable (hint: # type: List[<type>])
```

We could recognize the common case where the initializer constructs a generic instance and there is not enough information to infer all the type arguments. For other cases we can fall back to the current message.

The idea came from @aleksanb on gitter.

rafaelcaricio:
Fixes #2463

Improve error message when partial types are found for built-in types `Dict`, `List`, `Set`, `FrozenSet`.

Messages used to be like:
```
main:3: error: Need type annotation for 'x'
```
This PR modifies it to look like:
```
main:3: error: Need type annotation for 'x' (hint: "x: List[<type>] = ...")
```

In case the variable is not of a built-in type the hints are not shown.

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -1079,16 +1079,17 @@ class MessageBuilder:
                   ctx)
 
     def need_annotation_for_var(self, node: SymbolNode, context: Context,
-                                python_version: Tuple[int, int]) -> None:
-        hint = ''  # Only gives hint if it's a variable declaration and the partial type is a builtin type
-        if (isinstance(node, Var) and isinstance(node.type, PartialType) and
+                                python_version: Optional[Tuple[int, int]]) -> None:
+        hint = ''
+        # Only gives hint if it's a variable declaration and the partial type is a builtin type
+        if (python_version and isinstance(node, Var) and isinstance(node.type, PartialType) and
                 node.type.type and node.type.type.fullname() in reverse_builtin_aliases):
             alias = reverse_builtin_aliases[node.type.type.fullname()]
             alias = alias.split('.')[-1]
             type_dec = '<type>'
             if alias == 'Dict':
                 type_dec = '{}, {}'.format(type_dec, type_dec)
-            if python_version < PYTHON3_VERSION:
+            if python_version < (3, 6):
                 hint = ' (hint: "{} = ...  # type: {}[{}]")'.format(node.name(), alias, type_dec)
             else:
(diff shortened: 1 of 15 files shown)


mypy/messages.py
```diff
@@ -1077,8 +1078,21 @@ def unimported_type_becomes_any(self, prefix: str, typ: Type, ctx: Context) -> N
...
+                                python_version: Tuple[int, int]) -> None:
+        hint = ''  # Only gives hint if it's a variable declaration and the partial type is a builtin type
+        if (isinstance(node, Var) and isinstance(node.type, PartialType) and
+                node.type.type and node.type.type.fullname() in reverse_builtin_aliases):
+            alias = reverse_builtin_aliases[node.type.type.fullname()]
+            alias = alias.split('.')[-1]
+            type_dec = '<type>'
+            if alias == 'Dict':
+                type_dec = '{}, {}'.format(type_dec, type_dec)
+            if python_version < PYTHON3_VERSION:
```
JukkaL: I'd prefer `... < (3, 6)` here since `PYTHON3_VERSION` might get changed to `(3, 7)` in the future and the would then become incorrect.

mypy/plugins/attrs.py
```diff
@@ -133,7 +134,8 @@ def argument(self, ctx: 'mypy.plugin.ClassDefContext') -> Argument:
                 # assignment, which is where you would fix the issue.
                 node = self.info[self.name].node
                 assert node is not None
-                ctx.api.msg.need_annotation_for_var(node, self.context)
+                python_version: Tuple[int, int] = sys.version_info[:2]
```
JukkaL: This is slightly questionable, since the target Python version could be different from the current Python version we are running on. If we can't easily get the target Python version, I'd prefer to either not provide a Python version at all (such as giving `None` as the version -- we'd omit the hint in this case), or to provide a constant Python version for consistency (say, `(3, 7)`).

JukkaL:
The changes look good -- thanks for the updates! This is ready to merge if the builds are green.

rafaelcaricio:
@JukkaL Thank for the review! Glad to help.

PR #6785, opened 2019-05-06
Ask user about __init__ on duplicate packages

### LINKED ISSUE #5743: Add a note suggesting to add __init__.py in case of duplicate modules
For example:
```
dir/lib.py: error: Duplicate module named 'lib'
dir/lib.py: note: Are you missing '__init__.py'?
```
Probably we should use some heuristic to add this note only when it makes sense (e.g. there is at least one directory level between the file and the place where mypy is running).

jxcl:
Fixes #5743

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2602,8 +2602,8 @@ def load_graph(sources: List[BuildSource], manager: BuildManager,
                 -1, -1,
                 "Duplicate module named '%s' (also at '%s')" % (st.id, graph[st.id].xpath)
             )
-            p1 = len(pathlib.PurePath(st.xpath).parents)  # type: int
-            p2 = len(pathlib.PurePath(graph[st.id].xpath).parents)  # type: int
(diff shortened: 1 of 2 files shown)


mypy/build.py
```diff
@@ -2601,6 +2602,15 @@ def load_graph(sources: List[BuildSource], manager: BuildManager,
                 -1, -1,
                 "Duplicate module named '%s' (also at '%s')" % (st.id, graph[st.id].xpath)
             )
+            p1 = len(pathlib.PurePath(st.xpath).parents)  # type: int
+            p2 = len(pathlib.PurePath(graph[st.id].xpath).parents)  # type: int
```
ilevkivskyi: Do you really need these type comments? Mypy should be able to infer that `len()` gives an `int`.

ilevkivskyi:
I think you might need to add/update some test cases.

jxcl:
No test cases broke as far as I could tell, but I'll add a case for triggering the reminder.

PR #6787, opened 2019-05-07
Fix mypyc build

msullivan:
mypyc doesn't support non-default arguments following default arguments,
even if they are keyword only.


gvanrossum:
Have you filed a mypyc issue to fix this?

msullivan:
https://github.com/mypyc/mypyc/pull/607 is a PR to fix it

PR #6791, opened 2019-05-07
Clean up last traces of 'final_value' field

Michael0x2a:
This pull request is a follow-up to https://github.com/python/mypy/pull/6763: it renames some parameter names and variables that got missed earlier.

One interesting side-note: it seems like 'Var' notes also contain a `final_value` field that does almost the same thing as this `last_known_value` field, but is ultimately unrelated: it was introduced back in https://github.com/python/mypy/pull/5522. I decided to leave this field alone.

(I discovered this while updating mypyc and discovered (to my surprise) that nothing broke.)

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1804,7 +1804,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         if self.is_literal_context():
             return LiteralType(value=value, fallback=typ)
         else:
-            return typ.copy_modified(final_value=LiteralType(
+            return typ.copy_modified(last_known_value=LiteralType(
                 value=value,
                 fallback=typ,
                 line=typ.line,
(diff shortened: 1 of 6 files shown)


Michael0x2a:
As a tangent, the presence of this new [`Var.final_value`](https://github.com/python/mypy/blob/2f99e09dadcddf794d7c36019e8e988b732ef190/mypy/nodes.py#L824) field makes me wonder if it might be worth supporting Literal floats after all.

It'd be nice if we could unify this field with the existing Literal types work, or even remove it altogether in favor of using the inferred types. But that would mean dropping support for floats/dropping support for optimizing code involving floats in mypyc, which feels bad.

ilevkivskyi:
> As a tangent, ...

I would leave it as is for now, but in long term perspective I see a value in simplifying these things.

PR #6792, opened 2019-05-07
Require explicit annotations for List[object] and Dict[*, object]

jxcl:
This currently breaks mypy typechecking. Including instances like 

```
def get_reachable_graph(root: object) -> Tuple[Dict[int, object],
                                               Dict[int, Tuple[int, object]]]:
    parents = {}
    seen = {id(root): root}
```

Should `seen` require explicit annotation, or presumably mypy should be smart enough to figure this out. 

(#3816)

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1901,7 +1901,19 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                 rvalue_type = self.expr_checker.accept(rvalue)
                 if not inferred.is_final:
                     rvalue_type = remove_instance_last_known_values(rvalue_type)
+                self.check_forbidden_inference_types(rvalue_type, rvalue)
                 self.infer_variable_type(inferred, lvalue, rvalue_type, rvalue)
+
+    def check_forbidden_inference_types(self, rvalue_type: Type, rvalue: Context) -> None:
+        if isinstance(rvalue_type, Instance):
+            if rvalue_type.type.fullname() == 'builtins.list':
+                list_arg_type = rvalue_type.args[0]
+                if isinstance(list_arg_type, Instance) and list_arg_type.type.fullname() == 'builtins.object':
+                    self.msg.forbidden_inference_of_object(rvalue_type, rvalue)
+            elif rvalue_type.type.fullname() == 'builtins.dict':
+                dict_value_type = rvalue_type.args[1]
+                if isinstance(dict_value_type, Instance) and dict_value_type.type.fullname() == 'builtins.object':
+                    self.msg.forbidden_inference_of_object(rvalue_type, rvalue)
 
     def check_compatibility_all_supers(self, lvalue: RefExpr, lvalue_type: Optional[Type],
                                        rvalue: Expression) -> bool:
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -14,7 +14,7 @@ from mypy.nodes import (
     SymbolTable, Statement, MypyFile, Var, Expression, Lvalue, Node,
     OverloadedFuncDef, FuncDef, FuncItem, FuncBase, TypeInfo,
     ClassDef, Block, AssignmentStmt, NameExpr, MemberExpr, IndexExpr,
-    TupleExpr, ListExpr, ExpressionStmt, ReturnStmt, IfStmt,
+    TupleExpr, ListExpr, DictExpr, ExpressionStmt, ReturnStmt, IfStmt,
     WhileStmt, OperatorAssignmentStmt, WithStmt, AssertStmt,
     RaiseStmt, TryStmt, ForStmt, DelStmt, CallExpr, IntExpr, StrExpr,
     UnicodeExpr, OpExpr, UnaryExpr, LambdaExpr, TempNode, SymbolTableNode,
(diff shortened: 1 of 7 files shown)


mypy/checker.py
```diff
@@ -1901,8 +1901,20 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
                 rvalue_type = self.expr_checker.accept(rvalue)
                 if not inferred.is_final:
                     rvalue_type = remove_instance_last_known_values(rvalue_type)
+                self.check_forbidden_inference_types(rvalue_type, rvalue)
```
ilevkivskyi: I think this should be triggered only if `rvalue` is a list or dict literal, like `[x, y, z]` or `{1: 2, 3: 4}`.

mypy/messages.py
```diff
@@ -1035,6 +1035,12 @@ def operator_method_signatures_overlap(
                       forward_method, self.format(forward_class)),
                   context)
 
+    def forbidden_inference_of_object(self, expr: Type, context: Context) -> None:
+        self.fail('Variable inferred to {}, which is forbidden.'.format(
```
ilevkivskyi: I think the error message should be just `Need type annotation for variable`. I think you can re-use the existing method for this error.

mypy/checker.py
```diff
@@ -1901,8 +1901,20 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
...
+    def check_forbidden_inference_types(self, rvalue_type: Type, rvalue: Context) -> None:
+        if isinstance(rvalue_type, Instance):
+            if rvalue_type.type.fullname() == 'builtins.list':
+                list_arg_type = rvalue_type.args[0]
+                if isinstance(list_arg_type, Instance) and list_arg_type.type.fullname() == 'builtins.object':
+                    self.msg.forbidden_inference_of_object(rvalue_type, rvalue)
+            elif rvalue_type.type.fullname() == 'builtins.dict':
+                dict_value_type = rvalue_type.args[1]
+                if isinstance(dict_value_type, Instance) and dict_value_type.type.fullname() == 'builtins.object':
+                    self.msg.forbidden_inference_of_object(rvalue_type, rvalue)
```
ilevkivskyi: I think we should not emit these errors if `--allow-untyped-globals` is set to true and we are not at function scope (especially if we change the message to be just `Need type annotation ...`).

ilevkivskyi:
Please don't force-push and/or rebase, this makes reviewing much harder, just push more commits and merge when necessary.

jxcl:
Sorry, didn't see that until my last force push. I will stop.

mypy/checker.py
```diff
@@ -1901,8 +1901,47 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
...
+
+            elif isinstance(rvalue, DictExpr):
+                dict_value_type = rvalue_type.args[1]
+                if isinstance(dict_value_type, Instance):
+                    if dict_value_type.type.fullname() != 'builtins.object':
+                        # The values in this dict have a specific type, no further checks
+                        # are needed
+                        return
+
+                    if self.iterable_has_object_type(map(lambda x: x[1], rvalue.items)):
```
jxcl: I'm pretty sure this is not optimal, but I couldn't think how to do this a better way

mypy/messages.py
jxcl: Would there be any use to modify this method to add a "reason" field to perhaps make the error message more friendly?

ilevkivskyi: I think it may be only confusing. A better idea would be to add a note with a link to docs under such error explaining why annotations are needed. I think this idea already appeared in one of the issues. This can probably be done in a separate PR.

mypy/checker.py
```diff
@@ -1901,8 +1901,47 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
                 rvalue_type = self.expr_checker.accept(rvalue)
                 if not inferred.is_final:
                     rvalue_type = remove_instance_last_known_values(rvalue_type)
+                self.check_forbidden_inference_types(lvalue, rvalue, rvalue_type)
                 self.infer_variable_type(inferred, lvalue, rvalue_type, rvalue)
 
+    def check_forbidden_inference_types(self, lvalue: Lvalue, rvalue: Context, rvalue_type: Type) -> None:
+        if (isinstance(rvalue_type, Instance) and isinstance(lvalue, RefExpr)):
```
ilevkivskyi: Parentheses are redundant here.

mypy/checker.py
```diff
@@ -1901,8 +1901,47 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
                 rvalue_type = self.expr_checker.accept(rvalue)
                 if not inferred.is_final:
                     rvalue_type = remove_instance_last_known_values(rvalue_type)
+                self.check_forbidden_inference_types(lvalue, rvalue, rvalue_type)
                 self.infer_variable_type(inferred, lvalue, rvalue_type, rvalue)
 
+    def check_forbidden_inference_types(self, lvalue: Lvalue, rvalue: Context, rvalue_type: Type) -> None:
```
ilevkivskyi: Please add a docstring to this function.

mypy/checker.py
```diff
@@ -1901,8 +1901,47 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
...
+                        # The values in this dict have a specific type, no further checks
+                        # are needed
+                        return
+
+                    if self.iterable_has_object_type(map(lambda x: x[1], rvalue.items)):
+                        return
+
+                    self.msg.need_annotation_for_var(lvalue.node, rvalue)
+
+    def iterable_has_object_type(self, items: Iterator[Expression]) -> bool:
```
ilevkivskyi: A docstring here would be also helpful.

ilevkivskyi: Why can't you just accept an `Iterable` here?

mypy/checker.py
ilevkivskyi: You can just pass a list here `[v for (k, v) in rvalue.items]`.

mypy/checker.py
```diff
@@ -1904,16 +1904,22 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
...
                 if isinstance(list_arg_type, Instance):
                     if list_arg_type.type.fullname() != 'builtins.object':
                         # This list has a specific type, no further checks needed.
                         return
 
-                    if self.iterable_has_object_type(iter(rvalue.items)):
+                    if self.iterable_has_object_type(rvalue.items):
                         return
 
                     self.msg.need_annotation_for_var(lvalue.node, rvalue)
```
jxcl: This currently throws a mypy error:
```
mypy/checker.py:1919: error: Argument 1 to "need_annotation_for_var" of "MessageBuilder" has incompatible type "Optional[SymbolNode]"; expected "SymbolNode"
```

I guess I need to check if lvalue.node is none, but what do I do if it is?

ilevkivskyi: I think the best way to solve this, is to use `inferred` as the first argument of this method, which is a non-optional `Var` at the place where the method is called  (instead of `lvalue`). You only use `lvalue` once in an `if` at the top, instead you can move this `if` to the call site (this will also save one indentation level here).

jxcl:
Re: Guido's comment on the issue, should I add another option to disable this behavior to preserve some backwards compatibility?

JukkaL:
It may be best to first only enable this through a flag, and once we've successfully used this with a few big codebases (for example, with internal Dropbox codebases) in addition to mypy we can make this enabled by default (or always).

ilevkivskyi:
> It may be best to first only enable this through a flag, and once we've successfully used this with a few big codebases (for example, with internal Dropbox codebases) in addition to mypy we can make this enabled by default (or always).

Why can't we just try it with Dropbox code bases _now_, even before merging? Adding new flag is still some effort, so I would propose to first address existing comments, add tests, make CI build green (this may already highlight some problems in mypy self-check, as I mentioned in my review), and then I will try this PR against our internal code bases to see how well it works.

mypy/checker.py
jxcl: According to my read, `inferred` is set by `check_lvalue` which returns a `Tuple[Optional[Type], Optional[IndexExpr], Optional[Var]]`, so we still have to check if the `Var` is there. Perhaps I am missing something?

For moving the `if`, you're talking about `if isinstance(rvalue_type, Instance) and isinstance(lvalue, RefExpr):`, correct?

ilevkivskyi: > According to my read, inferred is set by `check_lvalue` which returns a `Tuple[Optional[Type], Optional[IndexExpr], Optional[Var]]`

Well your reading is wrong. Please read more carefully, there is an unpacking.
> For moving the `if`, you're talking about `if isinstance(rvalue_type, Instance) and isinstance(lvalue, RefExpr):`, correct?

Yes.

jxcl:
Other test fixes still incoming

ilevkivskyi:
What is the status here? This has a merge conflict and some tests still fail.

msullivan:
Closing this in favor of #9051. (Which I'm not sure if it will get merged, but.)

PR #6796, opened 2019-05-07
Update error message for incompatible args

### LINKED ISSUE #5025: Argument X of "Y" incompatible with supertype "Z" - needs more detail

feature request. Presumably mypy has the type information from the supertype available and at hand. It would be extremely helpful if that information was included in the diagnostics so that I don't have to go hunting for it.

```
import datetime
class Tzoffset ( datetime.tzinfo ):
	_delta = _tzname = None
	def __init__ ( self, delta, tzname='' ):
		# type: ( datetime.timedelta, str ) -> None
		assert isinstance ( delta, datetime.timedelta )
		self._delta = delta
		self._tzname = tzname
	def utcoffset ( self, dt ):
		# type: ( datetime.datetime ) -> datetime.timedelta
		return self._delta
	def dst ( self, dt ):
		# type: ( datetime.datetime ) -> datetime.timedelta
		return self._delta
	def tzname ( self, dt ):
		# type: ( datetime.datetime ) -> str
		return self._tzname
```

error: Argument 1 of "tzname" incompatible with supertype "tzinfo"

error: Argument 1 of "tzname" incompatible with supertype "tzinfo"; expected "foo"

Python 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:19:30) [MSC v.1500 32 bit (Intel)] on win32
Python 3.6.4 (v3.6.4:d48eceb, Dec 19 2017, 06:04:45) [MSC v.1900 32 bit (Intel)] on win32
mypy 0.600
dunno how to check git master

mypy --py2 %1 --ignore-missing-imports --disallow-untyped-defs
mypy %1 --ignore-missing-imports --disallow-untyped-defs

  the full traceback below.

(n/a)

matys18:
Closes #5025 🛠 

Updated the error message that is displayed when the argument types of a method in a subclass are incompatible with the arguments of the same method in the superclass.

Since the output message has changed there are quite a few broken tests. I would like to review and agree on the wording before changing them.

```python
# sample.py

class A:

    def kek(self, lol: int = 0) -> None:
        print("Kek in A")


class B(A):

    def kek(self, haha: str = "") -> None:
        super().kek(1)
        print("Kek in B")
```

Output before:
```sh
mypy sample.py
sample.py:11: error: Argument 1 of "kek" incompatible with supertype "A"
```

Output now:
```
mypy sample.py
sample.py:11: error: Argument 1 of "kek" of type "str" is incompatible with supertype "A"; supertype defines the argument type as "int"
```

As pointed out by @ilevkivskyi incompatible return type error reporting could also benefit from the same treatment. Will enter a followup issue once this is merged.


mypy/messages.py
```diff
@@ -796,11 +796,15 @@ def signature_incompatible_with_supertype(
...
+            name_in_supertype: str, arg_type_in_supertype: Type, supertype: str,
+            context: Context) -> None:
         target = self.override_target(name, name_in_supertype, supertype)
-        self.fail('Argument {} of "{}" incompatible with {}'
-                  .format(arg_num, name, target), context)
+        arg_type_f = self.format_bare(arg_type)
+        arg_type_in_supertype_f = self.format_bare(arg_type_in_supertype)
+        self.fail('Argument {} of "{}" of type "{}" is incompatible with {}; '
+                  'supertype defines the argument type as "{}"'
+                  .format(arg_num, name, arg_type_f, target, arg_type_in_supertype_f), context)
```
ilevkivskyi: I don't think you need `arg_type_f` in this function. The third format argument is the name of the superclass (it was already OK). You only need `arg_type_in_supertype_f`.

matys18: Fair enough. I've updated the code 👍

PR #6800, opened 2019-05-08
Sync typeshed

ilevkivskyi:
Syncing just because we didn't do this for a while (plus to test the recent `str.__contains__()` signature updates).

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -717,7 +717,7 @@ class StubGenerator(mypy.traverser.TraverserVisitor):
             typename = self.print_annotation(annotation)
             if (isinstance(annotation, UnboundType) and not annotation.args and
                     annotation.name == 'Final' and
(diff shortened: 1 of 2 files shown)


mypy/stubgen.py
```diff
@@ -717,7 +717,7 @@ def get_init(self, lvalue: str, rvalue: Expression,
             typename = self.print_annotation(annotation)
             if (isinstance(annotation, UnboundType) and not annotation.args and
                     annotation.name == 'Final' and
-                    self.import_tracker.module_for.get('Final') in ('typing, typing_extensions')):
+                    self.import_tracker.module_for.get('Final') in ('typing', 'typing_extensions')):
```
JelleZijlstra: Nice that we were able to find this with the `__contains__` change.

ilevkivskyi: Yes, exactly!

PR #6804, opened 2019-05-08
Improve import error message for stubgen Issue 6794

### LINKED ISSUE #6794: stubgen -m/-p "Failed to import xx skipping it" because of ModuleNotFoundError
`stubgen -m my_module`
`stubgen -p my_package `
output: "Failed to import my_module/my_package skipping it"  Error without details.
Because the environment actually missing packages which generate ModuleNotFoundError.

mypy 0.701
python3.7

xbian0403:
Since the error message is being printing out, any good ideas on how to test it?

`pytest -n0 --mypy-verbose -k testImportErrorMessage_import -s`
currently error message looks like this:
Failed to import a with error: ("No module named 'b'",); skipping it

Fixes #6794

### DIFF
--- a/mypy/stubutil.py
+++ b/mypy/stubutil.py
@@ -14,8 +14,9 @@ from typing import Optional, Tuple, List, IO, Iterator
 
 
 class CantImport(Exception):
-    pass
-
+    def __init__(self, module: str, message: str):
+        self.module = module
+        self.message = message
 
 def is_c_module(module: ModuleType) -> bool:
     return ('__file__' not in module.__dict__ or
@@ -104,7 +105,7 @@ def find_module_path_and_all_py2(module: str,
     try:
         output_bytes = subprocess.check_output(cmd_template % code, shell=True)
     except subprocess.CalledProcessError:
-        raise CantImport(module)
+        raise CantImport(module, e.args)
     output = output_bytes.decode('ascii').strip().splitlines()
     module_path = output[0]
     if not module_path.endswith(('.py', '.pyc', '.pyo')):
@@ -125,8 +126,8 @@ def find_module_path_and_all_py3(module: str) -> Optional[Tuple[str, Optional[Li
     # TODO: Support custom interpreters.
     try:
         mod = importlib.import_module(module)
-    except Exception:
-        raise CantImport(module)
+    except Exception as e:
+        raise CantImport(module, e.args)
     if is_c_module(mod):
         return None
     module_all = getattr(mod, '__all__', None)
@@ -155,8 +156,8 @@ def generate_guarded(mod: str, target: str,
             print('Created %s' % target)
 
 
-def report_missing(mod: str) -> None:
-    print('Failed to import {}; skipping it'.format(mod))
+def report_missing(mod: str, message: str) -> None:
+    print('Failed to import {} with error: {}; skipping it'.format(mod, message))
 
 
 def fail_missing(mod: str) -> None:
--- a/mypy/stubutil.py
+++ b/mypy/stubutil.py
@@ -104,8 +104,8 @@ def find_module_path_and_all_py2(module: str,
             "print(mod.__file__); print(json.dumps(getattr(mod, '__all__', None)))") % module
     try:
         output_bytes = subprocess.check_output(cmd_template % code, shell=True)
-    except subprocess.CalledProcessError:
-        raise CantImport(module, e.args)
+    except subprocess.CalledProcessError as e:
+        raise CantImport(module, str(e))
     output = output_bytes.decode('ascii').strip().splitlines()
     module_path = output[0]
     if not module_path.endswith(('.py', '.pyc', '.pyo')):
@@ -127,7 +127,7 @@ def find_module_path_and_all_py3(module: str) -> Optional[Tuple[str, Optional[Li
     try:
         mod = importlib.import_module(module)
     except Exception as e:
-        raise CantImport(module, e.args)
+        raise CantImport(module, str(e))
     if is_c_module(mod):
         return None
     module_all = getattr(mod, '__all__', None)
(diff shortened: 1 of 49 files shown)


mypy/test/teststubgen.py
```diff
@@ -266,6 +266,7 @@ def run_case(self, testcase: DataDrivenTestCase) -> None:
             with open(file, 'w') as f:
                 f.write(content)
 
+        sys.path.append('.')
```
ilevkivskyi: Why is this change needed? Is it here so that tests mimic the logic of `mypy.stubgen.main()`? Please add a comment.

test-data/unit/stubgen.test
```diff
@@ -1435,3 +1435,12 @@ from typing import Any
...
     def other(self) -> Any: ...
+
+[case testImportErrorMessage_import]
+# flags: --module a
+
+[file a.py]
+import b
+
+
+[out]
```
ilevkivskyi: Missing newline at the end of file.

test-data/unit/stubgen.test
```diff
@@ -1435,3 +1435,12 @@ from typing import Any
...
     @abstractmethod
     def other(self) -> Any: ...
+
+[case testImportErrorMessage_import]
+# flags: --module a
+
+[file a.py]
+import b
+
+
```
ilevkivskyi: These empty lines are not needed.

test-data/unit/stubgen.test
```diff
@@ -1435,3 +1435,12 @@ from typing import Any
 class C(Base, metaclass=abc.ABCMeta):
     @abstractmethod
     def other(self) -> Any: ...
+
+[case testImportErrorMessage_import]
```
ilevkivskyi: What exactly this test tests? Does it fail for the current version? (I assume it doesn't). Ideally I would add a test that checks the error formatting (or at least something that fails on master, but passes on your branch).

mypy/stubutil.py
```diff
@@ -155,8 +157,11 @@ def generate_guarded(mod: str, target: str,
             print('Created %s' % target)
 
 
-def report_missing(mod: str) -> None:
-    print('Failed to import {}; skipping it'.format(mod))
+def report_missing(mod: str, message: Optional[str] = None) -> None:
+    if message:
+        print('Failed to import {} with error: {}; skipping it'.format(mod, message))
+    else:
+        print('Failed to import {}; skipping it'.format(mod))
```
ilevkivskyi: I would avoid the duplication by making the default value for `message` an empty string (in the function signature) and writing this as:
```python
if message:
    message = ' with error: ' + message
print('Failed to import {}{}; skipping it'.format(mod, message))
```

mypy/test/teststubgen.py
xbian0403: current folder is not in sys path, when I test the function generate_stubs with `--module` option which should call `importlib.import_module` can't find the module. Thus I add it to sys.path.

test-data/unit/stubgen.test
xbian0403: Actually I need help on testing it. The error message is being print out. Any suggestion for testing it?
should I capture stdout and check it for only this test?
`
          capturedOutput = io.StringIO()
          sys.stdout = capturedOutput
`

ilevkivskyi: > should I capture stdout and check it for only this test?

I think this is OK, just don't forget to put the old value back in a `try ... finally: ...`. You can make this testing feature more generic for stubgen tests. For example you can run each test with captured output and add an additional check that stdout matches some regexp if the test case contains a `[stdout <some regexp here>]` line.

Alternatively you can simply make it a non data-driven test case similar to `mypy.teststubgen.StubgenCmdLineSuite.test_packages_found()` (this is probably even better).

mypy/test/teststubgen.py
```diff
@@ -64,6 +64,23 @@ def test_packages_found(self) -> None:
             finally:
                 os.chdir(current)
 
+    def test_module_not_found(self) -> None:
+        current = os.getcwd()
+        with tempfile.TemporaryDirectory() as tmp:
+            try:
+                import io
+                capturedOutput = io.StringIO()
```
ilevkivskyi: This is a non-typical case for Python variables, I would propose to use `captured_output`.

mypy/test/teststubgen.py
```diff
@@ -64,6 +64,23 @@ def test_packages_found(self) -> None:
             finally:
                 os.chdir(current)
 
+    def test_module_not_found(self) -> None:
+        current = os.getcwd()
+        with tempfile.TemporaryDirectory() as tmp:
+            try:
+                import io
```
ilevkivskyi: Please put this import at the top of file.

mypy/test/teststubgen.py
```diff
@@ -64,6 +64,23 @@ def test_packages_found(self) -> None:
...
+                capturedOutput = io.StringIO()
+                sys.stdout = capturedOutput
```
ilevkivskyi: I would move the parts that should not fail (the two above lines) above the `try`.

mypy/test/teststubgen.py
```diff
@@ -64,6 +64,23 @@ def test_packages_found(self) -> None:
...
+                os.chdir(tmp)
+                self.make_file(tmp, 'mymodule.py', content='import a')
+                opts = parse_options(['-m', 'mymodule'])
+                py_mods, c_mods = collect_build_targets(opts, mypy_options(opts))
+                sys.stdout = sys.__stdout__
```
ilevkivskyi: Please put this line in the `finally` block, since this is a clean-up step that should be always taken.

mypy/test/teststubgen.py
```diff
@@ -64,6 +64,23 @@ def test_packages_found(self) -> None:
...
+                match = re.search('No module named \'a\'', capturedOutput.getvalue())
```
ilevkivskyi: Use double quotes `"` for outer position to avoid `\`.

mypy/test/teststubgen.py
```diff
@@ -64,6 +64,23 @@ def test_packages_found(self) -> None:
...
+                self.assertTrue(match)
```
ilevkivskyi: You can probably simply use `self.assertRegex()` instead of the two above lines.

gvanrossum:
Thanks for the fix. Very useful!

Issue #2500, opened 2016-11-28
For incompatible attribute assignment, show class that defines attribute


JukkaL:
It would be nice to display the class that defines the attribute when we assign an incompatible value to an attribute:

```
class A:
    a = 1

class B(A):
    a = ''  # Show that 'a' was defined in 'A' (in addition to the current message)

class C(A): pass

c = C()
c.a = ''   # Again, maybe show that 'a' was defined in 'A'
```

Maybe also show the file and line on which the attribute was originally defined. However, we probably don't want these verbose messages in the body of a class if the attribute definition is in the same class, as the error is probably clear enough based on immediate context:

```
class A:
    a = 1

    def f(self) -> None:
        self.a = ''   # The current message is probably just fine, as the context is clear
```

viourr:
working on this.

jhabarsingh:
hi @JukkaL i want to work on can you please assign it to me

pn320:
So just looking at the issue, and doing some local testing I can see that mypy does identify and report the base class if the attribute of the class is overridden with an incorrect type, i.e it reports this error for line 5
`Incompatible types in assignment (expression has type "str", base class "A" defined the type as "int)`
and has no base class reporting for `c.a = ''`, just an ` Incompatible types in assignment (expression has type "str", variable has type "int")` message.

Just so I'm clear on what we'd like the PR to do, you want the base class where the attribute is defined to be reported for the second case and for the what about the first case (the error seems pretty clear to me)?

pn320:
I will be working on this, just to clarify the current error message is this,
```bash
t.py:6: error: Incompatible types in assignment (expression has type "str", base class "A" defined the type as "int")  [assignment]
t.py:14: error: Incompatible types in assignment (expression has type "str", variable has type "int")  [assignment]
Found 2 errors in 1 file (checked 1 source file)
```
It seems to me the error is fairly well defined, what is the output you'd like to see instead?

JukkaL:
This is still open. #6806 can be used for context, though it was closed in an incomplete state (and is now quite old). Marking this as "good second issue" instead of "good first issue" since this has been open for quite a long time and apparently the implementation is not quite trivial.

PR #6806, opened 2019-05-08
Show class that originally defined attribute in output

ucodery:
If a subclass or its instance attempts to assign an incompatible type
display the class that originally defined the type.

Fixes issue #2500


mypy/checker.py
```diff
@@ -2599,7 +2599,8 @@ def check_simple_assignment(self, lvalue_type: Optional[Type], rvalue: Expressio
             return rvalue_type
 
     def check_member_assignment(self, instance_type: Type, attribute_type: Type,
-                                rvalue: Expression, context: Context) -> Tuple[Type, Type, bool]:
+                                rvalue: Expression, context: MemberExpr
```
ilevkivskyi: I would rename `context` to `lvalue`. Context is typically something that is used mostly for error line/column.

mypy/checker.py
```diff
@@ -2612,9 +2613,17 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
         care about interaction between binder and __set__().
         """
         # Descriptors don't participate in class-attribute access
-        if ((isinstance(instance_type, FunctionLike) and instance_type.is_type_obj()) or
-                isinstance(instance_type, TypeType)):
-            rvalue_type = self.check_simple_assignment(attribute_type, rvalue, context)
+        if (isinstance(instance_type, CallableType) and instance_type.is_type_obj() and
+                isinstance(instance_type.ret_type, Instance)):
```
ilevkivskyi: Named tuple class object have a `TupleType` return type, you can "convert" it to instance by using its fallback.

mypy/checker.py
```diff
@@ -2674,6 +2684,52 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
         infer = is_subtype(rvalue_type, get_type) and is_subtype(get_type, set_type)
         return rvalue_type if infer else set_type, get_type, infer
 
+    def check_parent_member_assignment(self, lvalue_type: Type,
+                                       lvalue_attribute_type: Type, rvalue_type: Type,
+                                       rvalue: Expression, context: MemberExpr) -> Type:
```
ilevkivskyi: Also `context` -> `lvalue` here.

mypy/checker.py
```diff
@@ -2674,6 +2684,52 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
         infer = is_subtype(rvalue_type, get_type) and is_subtype(get_type, set_type)
         return rvalue_type if infer else set_type, get_type, infer
 
+    def check_parent_member_assignment(self, lvalue_type: Type,
+                                       lvalue_attribute_type: Type, rvalue_type: Type,
```
ilevkivskyi: The naming here is also confusing. `lvalue_type` is actually `lvalue_base_type`, while `lvalue_attribute_type` is just `lvalue_type`.

mypy/checker.py
```diff
@@ -2674,6 +2684,52 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+                                       rvalue: Expression, context: MemberExpr) -> Type:
+        """Type inherited member assignment.
+
+        This defers to check_simple_assignment if the attribute is local to the instance_type,
+        otherwise it will inform the user where the attribute type was originally defined.
+
+        Return the inferred rvalue_type.
+        """
+        if isinstance(lvalue_type, (Instance)):
```
ilevkivskyi: Parentheses are redundant here.

mypy/checker.py
```diff
@@ -2674,6 +2684,52 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+            attribute_is_defined = lvalue_type.has_readable_member(context.name)
+            if isinstance(lvalue_type, UnionType):
```
ilevkivskyi: This branch is unreachable (see `if` just above). Also I think we can't give any better error message for unions, so I would just remove this branch.

mypy/checker.py
```diff
@@ -2674,6 +2684,52 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+                direct_lvalue_vars = list(itertools.chain.from_iterable(
+                                          i.type.names for i in lvalue_type.items))
+                lvalue_bases = list(itertools.chain.from_iterable(
+                                    i.type.bases for i in lvalue_type.items))
+            else:
+                direct_lvalue_vars = lvalue_type.type.names
+                lvalue_bases = lvalue_type.type.bases
+                if '__metaclass__' in lvalue_type.type.names:
+                    # if the assigned metaclass is not class itself we have other problems
+                    metacls = lvalue_type.type.names['__metaclass__'].type.ret_type  # type: ignore
```
ilevkivskyi: This is a bed way of getting the metaclass. You should just use `metaclass_type` of the `TypeInfo` object.

mypy/checker.py
```diff
@@ -2674,6 +2684,52 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+                    lvalue_bases.append(metacls)
+                    attribute_is_defined = (attribute_is_defined or
+                                            metacls.has_readable_member(context.name))
+            if attribute_is_defined and context.name not in direct_lvalue_vars:
```
ilevkivskyi: Why do you need to check for `attribute_is_defined`, to avoid this extra message if the type is from a `__getattr__()`? Please add a comment. Naively, if an attribute is not defined at all, we can't know its type.

mypy/checker.py
```diff
@@ -2674,6 +2684,52 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+                for base in lvalue_bases:
+                    if context.name in base.type.names:
+                        parent_context = base.type.names[context.name].node
+                        if isinstance(parent_context, Var) and parent_context.info.is_generic():
+                            # attribute may have been defined in this class but its type is
```
```suggestion
                            # Attribute may have been defined in this class but its type is
```

mypy/checker.py
```diff
@@ -2630,7 +2639,8 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
             # the return type of __get__. This doesn't match the python semantics,
             # (which allow you to override the descriptor with any value), but preserves
             # the type of accessing the attribute (even after the override).
-            rvalue_type = self.check_simple_assignment(get_type, rvalue, context)
+            rvalue_type = self.check_parent_member_assignment(instance_type, get_type,
+                                                              attribute_type, rvalue, context)
```
ilevkivskyi: Why do you need to pass _both_ `attribute_type` and `get_type` here? This looks wrong, I think you should only pass `get_type`.

mypy/checker.py
```diff
@@ -2674,6 +2684,52 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+            if attribute_is_defined and context.name not in direct_lvalue_vars:
+                for base in lvalue_bases:
+                    if context.name in base.type.names:
+                        parent_context = base.type.names[context.name].node
+                        if isinstance(parent_context, Var) and parent_context.info.is_generic():
+                            # attribute may have been defined in this class but its type is
+                            # probably defined elsewhere, possibly in the instance itself.
+                            continue
+                        always_allow_any = (rvalue_type is not None and
+                                            not isinstance(rvalue_type, AnyType))
```
ilevkivskyi: This should be `lvalue_attribute_type` (in your notations) here and one line above. I understand you are trying to duplicate the logic in `check_simple_assignment()`. Maybe instead modify the latter to accept an optional extra message?

mypy/checker.py
```diff
@@ -2674,6 +2684,52 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+                                    i.type.bases for i in lvalue_type.items))
+            else:
+                direct_lvalue_vars = lvalue_type.type.names
+                lvalue_bases = lvalue_type.type.bases
+                if '__metaclass__' in lvalue_type.type.names:
+                    # if the assigned metaclass is not class itself we have other problems
+                    metacls = lvalue_type.type.names['__metaclass__'].type.ret_type  # type: ignore
+                    lvalue_bases.append(metacls)
+                    attribute_is_defined = (attribute_is_defined or
+                                            metacls.has_readable_member(context.name))
```
ilevkivskyi: Using the same message (`base class "..." defined ... as ...`) for both cases (base class and metaclass) is not good.

mypy/checker.py
ilevkivskyi: On a second thought, it seems to me, we may not show this additional info if descriptor protocol is involved, it may only add confusion, because the definition a user will see in the base class, is the descriptor type. But this is not a strong opinion.

mypy/checker.py
```diff
@@ -2612,31 +2612,39 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
-            self.msg, context, chk=self)
+            self.msg, lvalue, chk=self)
         if not attribute_type.type.has_readable_member('__set__'):
             # If there is no __set__, we type-check that the assigned value matches
             # the return type of __get__. This doesn't match the python semantics,
             # (which allow you to override the descriptor with any value), but preserves
             # the type of accessing the attribute (even after the override).
-            rvalue_type = self.check_simple_assignment(get_type, rvalue, context)
+            rvalue_type = self.check_parent_member_assignment(instance_type, get_type,
+                                                              lvalue, attribute_type, rvalue)
```
ilevkivskyi: Do you have a test case for a situation where `__get__` is involved?

mypy/checker.py
```diff
@@ -2645,18 +2653,18 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
         dunder_set_type = expand_type_by_instance(bound_method, typ)
 
         # Here we just infer the type, the result should be type-checked like a normal assignment.
-        # For this we use the rvalue as type context.
+        # For this we use the rvalue as type lvalue.
```
ilevkivskyi: The mass replace context -> lvalue went wrong.

mypy/checker.py
```diff
@@ -2670,10 +2678,53 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
         # what a user expects.
-        rvalue_type = self.check_simple_assignment(set_type, rvalue, context)
+        rvalue_type = self.check_simple_assignment(set_type, rvalue, lvalue)
         infer = is_subtype(rvalue_type, get_type) and is_subtype(get_type, set_type)
         return rvalue_type if infer else set_type, get_type, infer
 
+    def check_parent_member_assignment(self, lvalue_base_type: Type,
+                                       lvalue_type: Type, lvalue: MemberExpr,
+                                       rvalue_type: Type, rvalue: Expression) -> Type:
+        """Type inherited member assignment.
```
ilevkivskyi: I can't parse this as English sentence, could you please reformulate?

mypy/checker.py
```diff
@@ -2670,10 +2678,53 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+
+        This defers to check_simple_assignment if the attribute is local to the instance_type,
+        otherwise it will inform the user where the attribute type was originally defined.
```
ilevkivskyi: I would say explicitly that it does exactly the same as `check_simple_assignment()`, but adds better error message if possible.

mypy/checker.py
```diff
@@ -2670,10 +2678,53 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+
+        Return the inferred rvalue_type.
+        """
+        if (isinstance(lvalue_base_type, Instance) and
+                lvalue.name not in lvalue_base_type.type.names):
```
ilevkivskyi: Could you invert this check and put the last return here? This will save one indent level for the whole function body.

mypy/checker.py
```diff
@@ -2670,10 +2678,53 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+            metaclass = lvalue_base_type.type.metaclass_type
+            if metaclass and metaclass.has_readable_member(lvalue.name):
+                if lvalue.name in metaclass.type.names:
+                    meta_attribute = metaclass.type.names[lvalue.name].node
+                    if not (isinstance(meta_attribute, Var) and meta_attribute.info.is_generic()):
+                        return self.check_simple_assignment(
+                            lvalue_type, rvalue, lvalue,
+                            msg=message_registry.INCOMPATIBLE_TYPES_IN_ASSIGNMENT,
+                            lvalue_name=message_registry.METACLASS_MISMATCH.format(
+                                metaclass.type.defn.name))
```
ilevkivskyi: This whole part is duplicated below. Can you try factoring it out?

mypy/checker.py
```diff
@@ -2670,10 +2678,53 @@ def check_member_assignment(self, instance_type: Type, attribute_type: Type,
...
+                else:
+                    # attribute is defined further up in the metaclass MRO
+                    lvalue_warn = 'base class "{{}}" of {}'.format(
+                        message_registry.METACLASS_MISMATCH.format(
+                            metaclass.type.defn.name))
+                    lvalue_base_type = metaclass
+            if lvalue_base_type.has_readable_member(lvalue.name):
```
ilevkivskyi: Is this just a performance optimization? If yes, I would remove it. It will not give much speed-up and the logic in this method is already hard to follow.

test-data/unit/check-functions.test
```diff
@@ -2542,3 +2542,108 @@ def f() -> int: ...
...
+a = A(True)
+a().x = 9
+
+A.x = "" # E: Incompatible types in assignment (expression has type "str", variable has type "int")
+A().x = "" # E: Incompatible types in assignment (expression has type "str", variable has type "int")
+A(False).x = "" # E: Incompatible types in assignment (expression has type "str", variable has type "int")
+A()().x = "" # E: Incompatible types in assignment (expression has type "str", variable has type "int")
+A(False)().x = "" # E: Incompatible types in assignment (expression has type "str", variable has type "int")
+A()()().x = "" # E: Incompatible types in assignment (expression has type "str", variable has type "int")
+A(False)()().x = "" # E: Incompatible types in assignment (expression has type "str", variable has type "int")
```
ilevkivskyi: Why do you need al these test cases? What do they test?

test-data/unit/check-functions.test
```diff
@@ -2542,3 +2542,108 @@ def f() -> int: ...
...
+A.b = True
+A(True)
+A(True).b = False
+A(True)(False)
+A(True)(False)(True).b = False
+A(True)(None)
+A(True)().b = False
+A(True)(None).b = True
+A(True)(None).b = 1 # E: Incompatible types in assignment (expression has type "int", variable has type "bool")
+A(True)()(False).b = "" # E: Incompatible types in assignment (expression has type "str", variable has type "bool")
```
ilevkivskyi: Same question here.

test-data/unit/check-functions.test
```diff
@@ -2542,3 +2542,108 @@ def f() -> int: ...
...
+f().x = 0 # E: Incompatible types in assignment (expression has type "int", base class "A" defined the type as "str")
+f()().x = "baz"
+f()().x = 1 # E: Incompatible types in assignment (expression has type "int", base class "A" defined the type as "str")
+g().x = ""
+g().x = 2 # E: Incompatible types in assignment (expression has type "int", base class "A" defined the type as "str")
+
+[case testRecursiveCallable]
+from typing import Optional
+class A:
+    x = 0 # Type: int
```
ilevkivskyi: * You don't need to use type comments in tests.
* This is not a type comment because of a typo (uppercase `T`)
* It is not needed here

hauntsaninja:
Closing this PR, since it's gotten stale and there are a number of merge conflicts. But you're most welcome to pick it back up :-)

Issue #6798, opened 2019-05-07
testpep561.py fails locally when run from a venv


gvanrossum:
There's one test that always fails for me when I run pytest locally. Other people also see this, and it is apparently due to the nesting of virtual environments. In particular, I run pytest in a venv, and then the test creates a _nested_ virtual environment. It's been speculated that it would pass if I created my own virtual environment using `virtualenv` rather than `venv`. (The testpep561.py file uses `virtualenv`, and it has a comment explaining why -- it needs to create a Python 2 virtual environment.)

If this can't be fixed, maybe the offending tests can detect that they're being run inside of a `venv` and then be skipped? Note that Travis-CI and AppVeyor are unaffected.

ethanhs:
Could you give a snippet of output? Also I presume you mean that it would work if `venv` was used instead of `virtualenv`?

gvanrossum:
I don't know for sure what combination would work. I chatted a bit with @msullivan and he seems to think that `venv` doesn't nest properly but `virtualenv` does. I haven't tested this theory yet.

Output:
```
(v37) ~/src/mypy$ pytest -n0 mypy/test/testpep561.py 
==================================================================== test session starts ====================================================================
platform darwin -- Python 3.7.2, pytest-4.4.1, py-1.5.3, pluggy-0.9.0
rootdir: /Users/guido/src/mypy, inifile: pytest.ini
plugins: xdist-1.22.2, forked-0.2, cov-2.5.1
collected 12 items                                                                                                                                          

mypy/test/testpep561.py ......F....F                                                                                                                  [100%]

========================================================================= FAILURES ==========================================================================
_____________________________________________________________ TestPEP561.test_typedpkg_editable _____________________________________________________________

self = <mypy.test.testpep561.TestPEP561 testMethod=test_typedpkg_editable>

    def test_typedpkg_editable(self) -> None:
        self.simple_prog.create()
        with self.virtualenv() as venv:
            venv_dir, python_executable = venv
            self.install_package('typedpkg', python_executable, editable=True)
            self.simple_prog.check_mypy_run(
                python_executable,
                [SimpleMsg.msg_tuple],
>               venv_dir=venv_dir,
            )

mypy/test/testpep561.py:269: 
[...]

PR #6822, opened 2019-05-12
Skip tests in testpep561.py

PattenR:
Added logic to temporarily skip tests in testpep561.py that fail on Mac due to nesting of ```virtualenv ``` inside a ```venv```. This PR could be updated to only skip tests on Mac but I've left that out for now. Once this bug is fixed these tests should be added back in. 

Fixes [#6798](https://github.com/python/mypy/issues/6798)

### DIFF
--- a/mypy/test/testpep561.py
+++ b/mypy/test/testpep561.py
@@ -188,7 +188,9 @@ class TestPEP561(TestCase):
         dirs = get_site_packages_dirs(sys.executable)
         assert dirs
 
-    @pytest.mark.skipif(hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix, reason="Temporarily skip to avoid having a virtualenv within a venv.")
+    @pytest.mark.skipif(sys.platform == 'darwin' and hasattr(sys, 'base_prefix') and \
+    sys.base_prefix != sys.prefix, \
+    reason="Temporarily skip to avoid having a virtualenv within a venv.")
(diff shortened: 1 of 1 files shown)


PattenR:
Thanks! I've updated the PR with those changes.

mypy/test/testpep561.py
```diff
@@ -188,7 +188,9 @@ def test_get_pkg_dirs(self) -> None:
         dirs = get_site_packages_dirs(sys.executable)
         assert dirs
 
-    @pytest.mark.skipif(hasattr(sys, 'base_prefix') and sys.base_prefix != sys.prefix, reason="Temporarily skip to avoid having a virtualenv within a venv.")
+    @pytest.mark.skipif(sys.platform == 'darwin' and hasattr(sys, 'base_prefix') and \
+    sys.base_prefix != sys.prefix, \
+    reason="Temporarily skip to avoid having a virtualenv within a venv.")
```
gvanrossum: You don't need the backslashes, because it's inside `(...)`. Also, please reindent the continuation lines to line up (i.e. `sys` and `reason` in the 2nd and 3rd lines should start in the same column as `sys` on the previous line).

PattenR: Fixed, cheers!

gvanrossum:
Thanks again for figuring this out. I personally don't think we ever need to fix this better (if it's even possible).

PR #6829, opened 2019-05-14
Only honor --warn-unused-configs in non-incremental mode

### LINKED ISSUE #5957: Unused section warning with warn_unused_configs and cache
With mypy 0.641 or the latest git master, and a config file like:

```ini
[mypy]
warn_unused_configs = True
[mypy-bdb]
ignore_missing_imports = True
```

checking a file like

```python3
import bdb
```

suppresses the `error: No library stub file for standard library module 'bdb'` error as expected.

However, when running mypy a second time, I assume results are retrieved from its cache, and it shows:

```
Warning: unused section(s) in mypy.ini: [mypy-bdb]
```

which means I can only use `--warn-unused-configs` selectively (like after editing the config), as otherwise, I'd get a lot of bogus warnings in day-to-day usage.

This is probably somewhat similar to #2960, which is the same kind of issue for `--warn-unused-ignores`.

gvanrossum:
(In incremental mode it doesn't work right, and it isn't worth fixing.)

Fixes #5957

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -65,6 +65,7 @@ Config file
 ``--warn-unused-configs``
     This flag makes mypy warn about unused ``[mypy-<pattern>]`` config
     file sections.
+    (This requires turning off incremental mode using ``--no-incremental``.)
 
 
 .. _import-discovery:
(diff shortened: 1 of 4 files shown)


JukkaL:
Looks like some test case needs to be updated to use `--no-incremental`.

PR #6839, opened 2019-05-16
Support '# mypy: ' comments for inline configuration

### LINKED ISSUE #2938: ability to configure flags for a module from a comment within it
I would like to be able to use some syntax like

```
# MYPY: strict
```

to turn on all the useful stricter-checking flags, like

--disallow-subclassing-any
--warn-return-any
--disallow-untyped-calls
--strict-optional

Not totally sure what each one does.  Ideally we'd turn them all on globally, but ideally all our code would already be typed, and we don't live in that world.  I'd prefer this to our current mypy.ini for a few reasons:

- makes it really easy to turn on - no need to edit a separate file to add the flags
- the setting follows the file around during renames and moves
- it eliminates the need to know all the different flags

I suppose we'll still want to introduce new flags over time - to make stricter checking opt in so large codebases don't have to migrate all at once.  So either mypy standardizes ways to refer to a few flags (e.g. the first 3 could be called --strict-any? or --no-implicit-any like typescript?) or we could allow configuring these aliases in mypy.ini.  I think I'm a fan of the former - provides more standardization across codebases.

msullivan:
Implements line configuration using '# mypy: ' comments, following the
blueprint I proposed in #2938.

It currently finds them just using a regex which means it is possible to pick
up a directive spuriously in a string literal or something but honestly I am
just not worried about that in practice.

Examples of what it looks like in the tests.
Still needs documentation.

Fixes #2938.

Thoughts?

~(This is stacked on #6838 so can't be merged yet)~


JukkaL:
General idea looks solid to me. Probably worth adding some tests for invalid `# mypy: ...` comments. And I guess we should make sure that `# mypy: skip-file` either generates an error or does something reasonable. Maybe it should skip processing of the file?

I didn't do a proper review (yet).

mypy/build.py
```diff
@@ -1937,6 +1945,16 @@ def parse_file(self) -> None:
...
 
+    def parse_inline_configuration(self, source: str) -> None:
+        # Check for inline mypy: options directive and parse them.
+        flags = get_mypy_comments(source)
+        if flags:
+            changes, config_errors = parse_mypy_comments(flags, self.options)
+            self.options = self.options.apply_changes(changes)
+            self.manager.errors.set_file(self.xpath, self.id)
+            for error in config_errors:
+                self.manager.errors.report(-1, 0, error)
```
gvanrossum: Too bad the line number isn't preserved. If someone accidentally has a `# mypy:` comment in their file, how would they investigate the error if they didn't know about such comments? The error message doesn't even mention `# mypy:` comments.

At the same time, I do understand this is complicated (esp. since the flag parsing is separated from the flag extraction, and shared with `mypy.ini` file parsing).

Perhaps you can parse each `# mypy:` comment separately, and associate the comments with line numbers?

mypy/config_parser.py
```diff
@@ -198,29 +196,98 @@ def parse_section(prefix: str, template: Options,
...
+    parts = []
+    cur = []  # type: List[str]
+    i = 0
+    while i < len(s):
+        if s[i] == ',':
+            parts.append(''.join(cur).strip())
+            cur = []
+        elif s[i] == '"':
+            i += 1
+            while i < len(s) and s[i] != '"':
```
gvanrossum: So this doesn't gripe about unmatched quotes. :-(

mypy/util.py
```diff
@@ -18,6 +18,9 @@
 ENCODING_RE = \
     re.compile(br'([ \t\v]*#.*(\r\n?|\n))??[ \t\v]*#.*coding[:=][ \t]*([-\w.]+)')  # type: Final
 
+MYPY_RE = \
```
gvanrossum: That's not a great name -- maybe `MYPY_COMMENT_RE`?

mypy/build.py
```diff
@@ -1937,6 +1945,16 @@ def parse_file(self) -> None:
 
         self.check_blockers()
 
+    def parse_inline_configuration(self, source: str) -> None:
+        # Check for inline mypy: options directive and parse them.
```
ilevkivskyi: Turn this comment into a docstring?

test-data/unit/check-inline-config.test
```diff
@@ -0,0 +1,108 @@
...
+[file b.py.2]
+# no changes to a.py and we want to make sure it isn't rechecked
+[out]
+[out2]
+[rechecked b]
+
+[case testInlineError1]
+# mypy: invalid-whatever
+[out]
+main: error: Unrecognized option: invalid_whatever = True
```
ilevkivskyi: I would add more tests for the cases when user did something wrong:
* what if there is extra spaces around `=`?
* what if `;` used as a separator?
* what if a user forgot to quote a flag value that has commas?
etc.

mypy/build.py
msullivan: Yeah we should do this.

msullivan:
I am happy with how this all looks now.

Currently we produce an error on a skip-file directive. We could make it actually skip in a follow-up PR.

PR #6845, opened 2019-05-16
New semantic analyzer: fix crash caused by leaking placeholders

JukkaL:
This fixes a crash when processing the latest `unittest` stubs.

Previously we didn't always defer a target when placeholder nodes
were added to a symbol table. This could result in crashes.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3905,7 +3905,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
     def defer(self) -> None:
         """Defer current analysis target to be analyzed again.
 
-        This must called if something in the current target is
+        This must be called if something in the current target is
         incomplete or has a placeholder node.
         """
         self.deferred = True
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -4196,13 +4196,12 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         """
         names = self.current_symbol_table()
         existing = names.get(name)
-        is_placeholder = isinstance(symbol.node, PlaceholderNode)
-        if is_placeholder and can_defer:
+        if isinstance(symbol.node, PlaceholderNode) and can_defer:
             self.defer()
         if (existing is not None
                 and context is not None
                 and (not isinstance(existing.node, PlaceholderNode)
-                     or (is_placeholder and
+                     or (isinstance(symbol.node, PlaceholderNode) and
                          # Allow replacing becomes_typeinfo=False with becomes_typeinfo=True.
                          # This can happen for type aliases and NewTypes.
                          not symbol.node.becomes_typeinfo))):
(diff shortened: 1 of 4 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -3899,7 +3903,11 @@ def lookup_qualified(self, name: str, ctx: Context,
             return None
 
     def defer(self) -> None:
-        """Defer current analysis target to be analyzed again."""
+        """Defer current analysis target to be analyzed again.
+
+        This must called if something in the current target is
```
ilevkivskyi:
```suggestion
        This must be called if something in the current target is
```

mypy/newsemanal/semanal.py
```diff
@@ -848,7 +847,12 @@ def visit_class_def(self, defn: ClassDef) -> None:
     def analyze_class(self, defn: ClassDef) -> None:
         fullname = self.qualified_name(defn.name)
         if not defn.info and not self.is_core_builtin_class(defn):
-            self.add_symbol(defn.name, PlaceholderNode(fullname, defn, True), defn)
+            # Add placeholder so that self-references in base classes can be
+            # resolved.  We don't want this to cause a deferral, since if there
+            # are no incomplete references, we'll replace this with a TypeInfo
+            # before returning.
+            self.add_symbol(defn.name, PlaceholderNode(fullname, defn, True), defn,
+                            can_defer=False)
```
ilevkivskyi: OK, this makes sense, in the test case `Iterable` will trigger `Name not defined` instead of deferring on final iteration. So we shouldn't defer here either.

I am just a bit worried that this may cause placeholder to leak to later stages, so I would add couple more tests, for example with a start import in the file where the class is defined.

Do we also need the same for `A = NewType('A', List['A'])`?

test-data/unit/check-newsemanal.test
```diff
@@ -2200,3 +2201,28 @@ x = 1
...
+import unittest
+[file unittest/__init__.pyi]
+from unittest.suite import *
+
+def load_tests() -> TestSuite: ...
+[file unittest/suite.pyi]
+from typing import Union # Iterable not imported
+import unittest.case
+
+_TestType = Union[unittest.case.TestCase]
```
ilevkivskyi: Maybe also add a test with the same alias, but without the union?

mypy/newsemanal/semanal.py
JukkaL: This actually isn't required to fix the new test case. Without this, we do some extra iterations, and some other tests would fail. The test failures are likely caused by other bugs triggered by the extra pass. I'll create a separate issue about them.

I can add tests if you have specific ideas, but I'd rather not add test cases just in case, as they can give a false sense of security. A test case needs to do very particular things to trigger this error case, and small changes can turn it into something that doesn't test anything useful. The existing test case was a result of some trial an error.

`NewType` was a good catch -- it requires a similar treatment. I added a test case that relies on another bug, which is somewhat unfortunate. It would be better to have a test case that tests the number of semantic analysis iterations, but that's not possible right now.

test-data/unit/check-newsemanal.test
JukkaL: I'm not sure if it will be provide useful coverage. The union may not be necessary, but it's here because this was derived from the real stubs for `unittest`, and they happen to contain a union :-)

PR #6848, opened 2019-05-16
Renable flake8 warning about getattr with constant string

msullivan:
Most of the `getattr` calls with constant strings were to shut up mypy, so add a bunch of
`# type: ignore`s for those.

A couple were for other things (something dumb with mypyc, using an invalid identifier),
so those get `# noqa`.

Not totally sure if this is actually better, which is why I didn't put the PR up when I first wrote it,
but pushing now to see what other people think so I can clean up the branch.
(I'm leaning towards it being better, but only marginally.)

### DIFF
--- a/mypy/fscache.py
+++ b/mypy/fscache.py
@@ -146,7 +146,7 @@ class FileSystemCache:
         tpl = tuple(seq)
         # FIXME: this works around typeshed claiming stat_result is from posix
         # (typeshed #2683)
-        st = getattr(os, 'stat_result')(tpl)
+        st = getattr(os, 'stat_result')(tpl)  # noqa
         self.stat_cache[path] = st
         # Make listdir() and read() also pretend this file exists.
         self.fake_package_cache.add(dirname)
--- a/mypy/fscache.py
+++ b/mypy/fscache.py
@@ -144,9 +144,7 @@ class FileSystemCache:
         seq[stat.ST_NLINK] = 1
         seq[stat.ST_SIZE] = 0
         tpl = tuple(seq)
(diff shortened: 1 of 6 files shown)


mypy/fscache.py
```diff
@@ -146,7 +146,7 @@ def _fake_init(self, path: str) -> os.stat_result:
         tpl = tuple(seq)
         # FIXME: this works around typeshed claiming stat_result is from posix
         # (typeshed #2683)
-        st = getattr(os, 'stat_result')(tpl)
+        st = getattr(os, 'stat_result')(tpl)  # noqa
```
JelleZijlstra: https://github.com/python/typeshed/pull/2683 was merged, so seems like this can just use `os.stat_result`?

In general I would have preferred for this to use `# type: ignore`, which is more greppable, and that way a run with `--warn-unused-ignores` could have found that this hack is no longer needed.

msullivan: Fixed. It was working around an issue that showed up in mypyc, not in mypy, so the getattr was needed for annoying reasons.

mypy/main.py
```diff
@@ -700,7 +700,7 @@ def add_invertible_flag(flag: str,
 
     # Set strict flags before parsing (if strict mode enabled), so other command
     # line options can override.
-    if getattr(dummy, 'special-opts:strict'):
+    if getattr(dummy, 'special-opts:strict'):  # noqa
```
gvanrossum: Curious that this requires a `# noqa` -- apparently flake8 doesn't realize that the literal string is not a valid identifier. :-)

PR #6859, opened 2019-05-17
Make parse errors always blockers in fine-grained mode

msullivan:
Since fine-grained mode doesn't re-run the parser, non-blocking errors get lost.
There are a bunch of different approaches that could fix this but for now just
disable non-blocking errors from the parser in fine-grained mode.

Partially disables #6813.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -267,7 +267,7 @@ class ASTConverter:
         self.errors.report(line, column, msg, severity='note')
 
     def fail(self, msg: str, line: int, column: int, blocker: bool = True) -> None:
-        # Fine-grained mode doesn't support non-blocking parse errors yet
+        # Fine-grained mode doesn't support non-blocking parse errors yet.
         blocker = blocker or self.options.fine_grained_incremental
         if blocker or not self.options.ignore_errors:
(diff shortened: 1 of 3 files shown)


mypy/fastparse.py
```diff
@@ -267,6 +267,8 @@ def note(self, msg: str, line: int, column: int) -> None:
         self.errors.report(line, column, msg, severity='note')
 
     def fail(self, msg: str, line: int, column: int, blocker: bool = True) -> None:
+        # Fine-grained mode doesn't support non-blocking parse errors yet
```
ilevkivskyi:
```suggestion
        # Fine-grained mode doesn't support non-blocking parse errors yet.
```

mypy/fastparse2.py
```diff
@@ -166,6 +166,8 @@ def __init__(self,
         self.type_ignores = set()  # type: Set[int]
 
     def fail(self, msg: str, line: int, column: int, blocker: bool = True) -> None:
+        # Fine-grained mode doesn't support non-blocking parse errors yet
```
ilevkivskyi:
```suggestion
        # Fine-grained mode doesn't support non-blocking parse errors yet.
```

PR #6860, opened 2019-05-18
Make expression checking always produce Any once a node has been deferred

msullivan:
This fixes an issue where the Any that is inferred temporarily when a
node is deferred could cause a bad overload to be picked and produce
a spurious error even though it checked properly the *second* time.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3276,7 +3276,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 has_any_type(typ)):
             self.msg.disallowed_any_type(typ, node)
 
-        if not self.chk.in_checked_function():
+        if not self.chk.in_checked_function() or self.chk.current_node_deferred:
             return AnyType(TypeOfAny.unannotated)
         else:
             return typ
--- a/test-data/unit/check-overloading.test
+++ b/test-data/unit/check-overloading.test
@@ -5001,3 +5001,31 @@ def f(x):
 
 reveal_type(f(g([])))  # E: Revealed type is 'builtins.list[builtins.int]'
 [builtins fixtures/list.pyi]
+
+[case testOverloadDeferredNode]
+from typing import Callable, TypeVar, Generic, Any, overload
+
+_S = TypeVar('_S')
+_T = TypeVar('_T')
+_R = TypeVar('_R')
+_F = TypeVar('_F', bound=Callable[..., Any])
+
+@overload
+def partial(__func: Callable[[_T], _S], __arg: _T) -> Callable[[], _S]: ...
+@overload
+def partial(__func: Callable[[_T, _S], _S], __arg: _T) -> Callable[[_S], _R]: ...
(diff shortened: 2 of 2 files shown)


test-data/unit/check-overloading.test
```diff
@@ -5001,3 +5001,31 @@ def f(x):
...
 reveal_type(f(g([])))  # E: Revealed type is 'builtins.list[builtins.int]'
 [builtins fixtures/list.pyi]
+
+[case testOverloadDeferredNode]
+from typing import Callable, TypeVar, Generic, Any, overload
+
+_S = TypeVar('_S')
+_T = TypeVar('_T')
+_R = TypeVar('_R')
+_F = TypeVar('_F', bound=Callable[..., Any])
```
ilevkivskyi: This type variable is not used in the test.

test-data/unit/check-overloading.test
```diff
@@ -5001,3 +5001,31 @@ def f(x):
...
+
+@overload
+def partial(__func: Callable[[_T], _S], __arg: _T) -> Callable[[], _S]: ...
+@overload
+def partial(__func: Callable[[_T, _S], _S], __arg: _T) -> Callable[[_S], _R]: ...
+def partial(*args: Any) -> Any:
+    pass
+
+
```
ilevkivskyi: One space would be enough.

mypy/checkexpr.py
```diff
@@ -3276,7 +3276,7 @@ def accept(self,
                 has_any_type(typ)):
             self.msg.disallowed_any_type(typ, node)
 
-        if not self.chk.in_checked_function():
+        if not self.chk.in_checked_function() or self.chk.current_node_deferred:
```
ilevkivskyi: I would return `TypeOfAny.special_form` for the latter case.

Michael0x2a:
@ilevkivskyi -- I did a bit of digging, and it seems the breach you were talking about is rooted in how we're computing the `arg_contains_any` variable here: https://github.com/python/mypy/blob/master/mypy/checkexpr.py#L1445

When you try running the test case in this PR, the args are of type `Any` and `int` on the first iteration, but this variable ended up being set to False. (So, the usual "if an arg is Any, return Any" logic ended up not running.)

It turned out the reason for *this* is that `has_any_type` is designed to ignore `special_form` Anys: https://github.com/python/mypy/blob/master/mypy/checkexpr.py#L3518.

Not sure if/how this should be fixed -- I'm not very familiar with what the node-deferring semantics are supposed to be. Maybe we could add a new TypeOfAny for deferred nodes, or maybe we could add a check to the overload logic to return early if a node is deferred? Or maybe Sully's fix is still the best approach overall.

JukkaL:
I'm going to merge this as is, since we need this at Dropbox urgently. The improvements discussed above may make sense as separate PRs.

ilevkivskyi:
@Michael0x2a Coming back to this I think we can live with this fix as is. I will only submit a PR with minor tests cleanups I suggested.

Issue #5599, opened 2018-09-11
Make enum attributes implicitly final


ilevkivskyi:
This is currently allowed but fails at runtime:
```python
from enum import Enum

class Other(Enum):
    test = 1

Other.test = 2
```
but it fails at runtime.

This is a follow-up for https://github.com/python/mypy/pull/5522

PR #6867, opened 2019-05-20
Add more precise inference for enum attributes

Michael0x2a:
This pull request makes two changes to enum attributes.

First, this PR refines type inference for expressions like `MyEnum.FOO` and `MyEnum.FOO.name`. Those two expressions will continue to evaluate to `MyEnum` and `str` respectively under normal conditions, but will evaluate to `Literal[MyEnum.FOO]` and `Literal["FOO"]` respectively when used in Literal contexts.

Second, the type of `MyEnum.FOO.value` will be more precise when possible: mypy will evaluate that expression to the type of whatever FOO was assigned in the enum definition, falling back to `Any` as a
default.

Somewhat relatedly, this diff adds a few tests confirming we handle enum.auto() correctly.

Two additional notes:

1. The changes I made to the `name` and `value` fields up above are strictly speaking unsafe. While those files are normally read-only (doing `MyEnum.FOO.name = blah` is a runtime error), it's actually possible to change those fields anyway by altering the  `_name_` and `_value_` fields which are *not* protected.

   But I think this use case is probably rare -- I'm planning on investigating the feasibility of just having mypy just disallow modifying these attributes altogether after I investigate how enums are used in some internal codebases in a little more detail.

2. I would have liked to make `MyEnum.FOO.value` also return an even more precise type when used in literal contexts similar to `MyEnum.FOO.name`, but I think our plugin system needs to be a bit more flexible first.


Michael0x2a:
I guess this is kinda-sorta a follow-up to https://github.com/python/mypy/issues/5599?

It makes some enum attributes implicitly Final for the purposes of type inference, but doesn't disallow assignment to them. (I'm planning on tackling that in a separate PR.)

test-data/unit/check-protocols.test
```diff
@@ -2427,7 +2427,7 @@ from typing import Protocol
 class P(Protocol): ...
 class C(P): ...
 
-reveal_type(C.register(int))  # E: Revealed type is 'def () -> builtins.int'
+reveal_type(C.register(int))  # E: Revealed type is 'def (x: builtins.object =, base: builtins.int =) -> builtins.int'
```
Michael0x2a: Not sure why the revealed type is the constructor signature -- when I try directly running mypy on this, I get `Type[builtins.int]` instead. I guess this is just some test-related artifact?

mypy/plugins/enums.py
```diff
@@ -0,0 +1,69 @@
+from typing import Optional
```
JukkaL: Add docstring explaining the purpose of the plugin and the implementation approach (how this uses `last_known_value` at least).

mypy/plugins/enums.py
```diff
@@ -0,0 +1,69 @@
...
+    ['{}.name'.format(prefix) for prefix in ENUM_PREFIXES]
+    + ['{}._name_'.format(prefix) for prefix in ENUM_PREFIXES]
+)
+ENUM_VALUE_ACCESS = (
+    ['{}.value'.format(prefix) for prefix in ENUM_PREFIXES]
+    + ['{}._value_'.format(prefix) for prefix in ENUM_PREFIXES]
+)
+
+
+def enum_name_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
```
JukkaL: Add docstring explaining what this does and how.

mypy/plugins/enums.py
```diff
@@ -0,0 +1,69 @@
...
+    enum_field_name = extract_underlying_field_name(ctx.type)
+    if enum_field_name is None:
+        return ctx.default_attr_type
+    else:
+        str_type = ctx.api.named_generic_type('builtins.str', [])
+        literal_type = LiteralType(enum_field_name, fallback=str_type)
+        return str_type.copy_modified(last_known_value=literal_type)
+
+
+def enum_value_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
```
JukkaL: Add docstring explaining what this does and how.

Michael0x2a:
Thanks for the review! I'll work on making the changes you suggested later today.

Just to quickly answer the two questions you asked though...

1. I wanted to have more precise inference for things like `A.x.name` mostly for the sake of completeness. The general impression I got was that enums have been sort of neglected for a while in mypy and was interested in working on trying to polish them up.

2. Things like `Literal[MyEnum.FOO]` is currently supported! It snuck in here: https://github.com/python/mypy/pull/6668

mypy/plugins/enums.py
```diff
@@ -1,23 +1,49 @@
...
 from mypy.types import Type, Instance, LiteralType
-from mypy.nodes import Var, MDEF
 
 # Note: 'enum.EnumMeta' is deliberately excluded from this list. Classes that directly use
 # enum.EnumMeta do not necessarily automatically have the 'name' and 'value' attributes.
-ENUM_PREFIXES = ['enum.Enum', 'enum.IntEnum', 'enum.Flag', 'enum.IntFlag']
-ENUM_NAME_ACCESS = (
-    ['{}.name'.format(prefix) for prefix in ENUM_PREFIXES]
-    + ['{}._name_'.format(prefix) for prefix in ENUM_PREFIXES]
+ENUM_PREFIXES: Final = {'enum.Enum', 'enum.IntEnum', 'enum.Flag', 'enum.IntFlag'}
```
JukkaL: Sadly, you must use a type comment since we support Python 3.5.

mypy/plugins/enums.py
```diff
@@ -27,7 +53,27 @@ def enum_name_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
...
+            BAR = B()
+
+    By default, mypy will infer that 'SomeEnum.FOO.value' and
+    'SomeEnum.BAR.value' both are of type 'Any'. This plugin refines
+    this inference so that mypy understands the expressions are
+    actually of types 'A' and 'B' respectively. This better reflects
+    the actual runtime behavior.
+
+    This plugin works simply by looking up the original value assigned
+    to the enum. For example,
```
JukkaL: Something is missing here?

test-data/unit/check-enum.test
```diff
@@ -456,3 +456,154 @@ main:3: error: Revealed type is 'm.F'
...
+F2.x.name      # E: "F2" has no attribute "name"
+F2.x._name_    # E: "F2" has no attribute "_name_"
+F2.x.value     # E: "F2" has no attribute "value"
+F2.x._value_   # E: "F2" has no attribute "_value_"
+F3.x.name      # E: "F3" has no attribute "name"
+F3.x._name_    # E: "F3" has no attribute "_name_"
+F3.x.value     # E: "F3" has no attribute "value"
+F3.x._value_   # E: "F3" has no attribute "_value_"
+
+[case testEnumAttributeChangeIncremental]
```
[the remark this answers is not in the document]
Michael0x2a: Hmm, I'm not sure if we have one. Do you know which file I should add the test to? (I don't remember where we keep the deserialization tests.)

mypy/plugins/enums.py
```diff
@@ -0,0 +1,69 @@
+from typing import Optional
+import mypy.plugin  # To avoid circular imports.
+from mypy.types import Type, Instance, LiteralType
+from mypy.nodes import Var, MDEF
+
+# Note: 'enum.EnumMeta' is deliberately excluded from this list. Classes that directly use
+# enum.EnumMeta do not necessarily automatically have the 'name' and 'value' attributes.
+ENUM_PREFIXES = ['enum.Enum', 'enum.IntEnum', 'enum.Flag', 'enum.IntFlag']
+ENUM_NAME_ACCESS = (
```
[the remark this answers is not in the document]
Michael0x2a: It turns out it is indeed faster, at least based on some microbenchmarking I did. I thought the list would be small enough that overhead would be about the same either way, but that was wrong.

(In retrospect, I guess doing on average 4 to 8 `some_str.__eq__(...)` calls per containment check is always going to be noticeably more expensive then doing a `__hash__(...)` followed by maybe an `__eq__(...)`, at least in Python.)

test-requirements.txt
```diff
@@ -4,6 +4,7 @@ flake8-bugbear; python_version >= '3.5'
 flake8-pyi; python_version >= '3.6'
 lxml==4.2.4
 mypy_extensions>=0.4.0,<0.5.0
+typing_extensions>=3.7.0,<4.0.0
```
Michael0x2a: I don't really understand why I needed to add this in as an explicit dependency.

All of the other pull requests seemed fine? But I wasn't able to import `typing_extensions` within the new `enums` plugin module without it...

ethanhs: @Michael0x2a All the other files do:
```
MYPY=False
if MYPY:
```
I believe you don't have a guard for the import? I think doing the same guard would fix the issue.

Michael0x2a: Argh, that would do it, thanks!

(I guess just blindly grepping for "typing_extensions" might not have quite given me the full picture...)

PR #6869, opened 2019-05-20
Add a CI build that uses mypyc

msullivan:
This probably doesn't work yet but our CI only triggers when there is a PR up


msullivan:
Closes #6170!

Issue #6495, opened 2019-02-28
New semantic analyzer: crash on cyce in inheritance hierarchy


JukkaL:
This generates a crash (max iteration count reached):
```
class A(A): pass 
```

ilevkivskyi:
Actually I wanted to open a more general issue a while ago about ensuring that we are gracefully handling various kinds of "deadlocks": another thing I discovered involved circular star imports in some situations (I forgot a simple repro however).

PR #6872, opened 2019-05-21
Fix reporting of likely hangs in the new semantic analyzer

JukkaL:
Previously things like `class A(A): pass` resulted in a crash
instead of an error message. Now we generate an error message
with some context (but still not enough).

Partially addresses #6495.


JukkaL:
> Should we do the same also for report_hang() in process_top_level_function() below?

Probably yes? I'll create another PR to fix that.

PR #6876, opened 2019-05-21
When strict optional is disabled, anything can be falsey

### LINKED ISSUE #3601: Optional[NamedTuple] makes mypy think a branch is unreachable
In the following example, the `else` branch is not typechecked. As far as I've been able to narrow it down, it seems like it has to do with NamedTuple.

```py
from typing import Optional, NamedTuple, List

NamedTup = NamedTuple('NamedTup', [('a', str)])


class Foo:
    def __init__(self) -> None:
        self.x = get_optional()
        if self.x:
            # this is typechecked
            f: str = 2  # error: Incompatible types in assignment
        else:
            # this is not typechecked
            j: str = 2  # no error
            self.mypy_does_not_typecheck_this()  # no error
            reveal_type(j)  # no error


def get_optional() -> Optional[NamedTup]:
    return None


Foo()
```

Here is the output from `mypy` and `python`.
```
$ mypy n.py
n.py:11: error: Incompatible types in assignment (expression has type "int", variable has type "str")
$ python3 n.py
  File "n.py", line 23, in <module>
    Foo()
  File "n.py", line 15, in __init__
    self.mypy_does_not_typecheck_this()  # no error
AttributeError: 'Foo' object has no attribute 'mypy_does_not_typecheck_this'
```

### LINKED ISSUE #6827: Block incorrectly treated as unreachable when strict optional disabled
This program doesn't generate an error when using `--no-strict-optional`, even though every line is reachable:

```py
from typing import NamedTuple, Dict

class C(NamedTuple):
    a: int
    b: str

x: Dict[int, C]
y = x.get(int())
if not y:
    1 + ''  # No error (considered unreachable)
```

Apparently mypy thinks that the tuple object is always truthy, even though it could also be `None` when strict optional checking is disabled.

JukkaL:
Previously some code was treated incorrectly as unreachable, since
some checks where treated as always true even though they could
be false given a None value.

Fixes  #3601.

### DIFF
--- a/test-data/unit/check-namedtuple.test
+++ b/test-data/unit/check-namedtuple.test
@@ -859,6 +859,6 @@ a: C
 if not a:
     1()  # E: "int" not callable
 b = (1, 2)
-if not a:
+if not b:
     ''()  # E: "str" not callable
 [builtins fixtures/tuple.pyi]
--- a/test-data/unit/check-namedtuple.test
+++ b/test-data/unit/check-namedtuple.test
@@ -846,3 +846,19 @@ class Child(Parent):
 
 reveal_type(Child.class_method())  # E: Revealed type is 'Tuple[builtins.str, fallback=__main__.Child]'
 [builtins fixtures/classmethod.pyi]
+
+[case testNamedTupleAsConditionalStrictOptionalDisabled]
+# flags: --no-strict-optional
+from typing import NamedTuple
+
+class C(NamedTuple):
+    a: int
+    b: str
+
+a: C
+if not a:
+    1()  # E: "int" not callable
+b = (1, 2)
+if not a:
+    ''()  # E: "str" not callable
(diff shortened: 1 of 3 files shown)


test-data/unit/check-namedtuple.test
```diff
@@ -846,3 +846,19 @@ class Child(Parent):
...
+
+class C(NamedTuple):
+    a: int
+    b: str
+
+a: C
+if not a:
+    1()  # E: "int" not callable
+b = (1, 2)
+if not a:
```
ilevkivskyi: Shouldn't this be `b` here?

JukkaL:
Update the target issue; it was pointing to an obsolete duplicate.

test-data/unit/check-namedtuple.test
JukkaL: Yes!

Issue #5895, opened 2018-11-15
Error feedback for unions of lists that should be unions of sequences is not as helpful as it could be


toolness:
Sorry this issue has such an odd title, but I was about to report a mypy bug based on the following code:

```py
from typing import List, Union
from enum import Enum


class Boop(Enum):
    FOO = 1
    BAR = 2


def do_thing_with_enums(enums: Union[List[Enum], Enum]) -> None:
    if isinstance(enums, Enum):
        enums = list([enums])
    for thing in enums:
        thing.value


class Funky:
    boop: List[Boop]

    def boop_it_up(self) -> None:
        do_thing_with_enums(self.boop)
```

The mypy error output for the following code is:

```
boop.py:21: error: Argument 1 to "do_thing_with_enums" has incompatible type "List[Boop]"; expected "Union[List[Enum], Enum]"
```

I found this quite confusing since `List[Boop]` is also a `List[Enum]`, which I thought was a mypy bug.

However, after a lot of tinkering, I simplified the `do_thing_with_enums` signature to be:

```py
def do_thing_with_enums(enums: List[Enum]) -> None:
```

at this point mypy still failed, but it gave me _much_ better error feedback:

```
boop.py:21: error: Argument 1 to "do_thing_with_enums" has incompatible type "List[Boop]"; expected "List[Enum]"
boop.py:21: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
boop.py:21: note: Consider using "Sequence" instead, which is covariant
```

After converting the type annotation to use `Sequence` instead of `List`, everything worked fine.  I just wished it had given me such excellent advice for my original type annotation, rather than requiring me to simplify it, if that makes sense.

remilapeyre:
I made a very naive PR to display a note when this happen. I will be happy to improve it when you have time to review :)

PR #6877, opened 2019-05-21
Display note when having invariants in Union

remilapeyre:
Fix https://github.com/python/mypy/issues/5895

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -628,8 +628,13 @@ class MessageBuilder:
             msg = 'Argument {} {}has incompatible type {}; expected {}'.format(
                 arg_label, target, self.quote_type_string(arg_type_str),
                 self.quote_type_string(expected_type_str))
-            if isinstance(arg_type, Instance) and isinstance(expected_type, Instance):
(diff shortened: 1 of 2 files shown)


mypy/messages.py
```diff
@@ -628,8 +628,13 @@ def incompatible_argument(self, n: int, m: int, callee: CallableType, arg_type:
             msg = 'Argument {} {}has incompatible type {}; expected {}'.format(
                 arg_label, target, self.quote_type_string(arg_type_str),
                 self.quote_type_string(expected_type_str))
-            if isinstance(arg_type, Instance) and isinstance(expected_type, Instance):
-                notes = append_invariance_notes(notes, arg_type, expected_type)
+            expected_types = (
+                expected_type.items if isinstance(expected_type, UnionType)
+                else [expected_type]
+            )
```
gvanrossum: Please rewrite this as an if/else *statement*. The long-form conditional expression is less clear than that.

remilapeyre: Thanks, it should be better now.

PR #6881, opened 2019-05-22
New semantic analyzer: fix issues with looking up qualified names

### LINKED ISSUE #6866: Using qualified type on un-imported module in stubs causes false positive
Consider this test case:
```
[case testMissingSubmoduleClash]
# flags: --ignore-missing-imports
import other
import p.u

[file p/__init__.py]

[file p/u.pyi]
from . import c
x: c.C

[file other.py]
from p.c import B
```
Currently it fails with `tmp/p/u.pyi:2: error: Name 'c.C' is not defined`. This is a confluence of several problems, the central one is the TODO in `lookup_qualified()`. It causes `module_hidden` (note no `as id` import in stub) to be applied to `C` instead of `c`.

Also for unknown reasons, new semantic analyzer is more prone to this issue.

JukkaL:
Various fixes to `lookup_qualified`:

1. Allow qualified types that refer to unimported modules. Fixes #6866.
2. Flag module private references in the middle of a qualified name.
3. Don't allow module-level `__getattr__` in Python 3.6 and earlier
   outside stubs.
4. Better error message when qualified name includes a function/variable
   component.

Also includes a major refactor of `lookup_qualified`.

More cleanup is possible, but this PR seems big enough as is.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3870,7 +3870,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         if sym:
             for i in range(1, len(parts)):
                 node = sym.node
-                error_type = None
+                source_type = None
                 if isinstance(node, TypeInfo):
                     nextsym = node.get(parts[i])
                 elif isinstance(node, MypyFile):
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -3862,46 +3862,70 @@ def check_for_obsolete_short_name(self, name: str, ctx: Context) -> None:
...
+
+    def get_module_symbol(self, node: MypyFile, parts: List[str]) -> Optional[SymbolTableNode]:
+        """Look up a symbol from the module symbol table."""
+        # TODO: Use this logic in more places?
+        module = node.fullname()
+        names = node.names
+        # Rebind potential references to old version of current module in
+        # fine-grained incremental mode.
+        if module == self.cur_mod_id:
+            names = self.globals
```
ilevkivskyi: Is this actually needed in the new analyzer?

JukkaL: I don't think that anything has changed here. During a fine-grained update we have both the old and the new module object alive during semantic analysis.

ilevkivskyi: @JukkaL do we have a test for this? I tried to just remove these two lines of code and all tests passed. (See also a TODO in `visit_member_expr()`). Moreover, I removed also removed `rebind_symbol_table_node()` completely and all tests still passed.

Should we just remove it from the new analyzer?

PR #6882, opened 2019-05-22
New semantic analyzer: remove remains of pass 3

### LINKED ISSUE #6316: New semantic analyzer: remove redundant pass 3 / patching code
Much of the code of pass 3 is not required, but there may be checks that should be preserved. Some of the checks can probably be moved to the main pass.

JukkaL:
Nothing in pass 3 should need to be migrated any more.

Fixes #6316.


JukkaL:
> I think we can also remove TypeAnalyser.third_pass (and related code) as well 

A good point! I'm happy to do this in a follow-up PR.

Issue #6880, opened 2019-05-22
Request to update readme: replace vim-py by syntastic


robinchm:
Request to update Readme.

Currently the Readme features [vim-mypy](https://github.com/Integralist/vim-mypy) as the IDE integration for vim. As I just set up mypy with vim, this repo is dead. Its current HEAD is incompatible with the latest version of mypy (refer to its open issues) and it has not been updated for 2 years.

Instead I successfully integrate mypy with vim using [syntastic](https://github.com/vim-syntastic/syntastic). Putting the following in the `.vimrc` after installing syntastic works:
```
let g:syntastic_python_checkers=['mypy']
```

Could you consider updating the Readme so that vim users look into the working plugins?

gvanrossum:
Sounds like a good idea. Can you provide a PR?

gvanrossum:
Closed by #6886

PR #6886, opened 2019-05-23
Update readme: replace vim-mypy by syntastic

robinchm:
Vim-mypy is not maintained and does not work with the current version.
    Syntastic is a well-maintained static grammer checking plugin for
    vim that supports multiple languages.

    Addressing (https://github.com/python/mypy/issues/6880)


gvanrossum:
Thanks! IIRC none of the mypy core team are vim users, so we're dependent on contributors to help out here.

PR #6888, opened 2019-05-23
New semantic analyzer: refactor file_context

JukkaL:
Modify the main entry point `refresh_partial` to call `file_context`
to make things less error-prone.

Also remove arguments that are not needed.


mypy/newsemanal/semanal.py
```diff
@@ -407,17 +411,17 @@ def adjust_public_exports(self) -> None:
                     g.module_public = False
 
     @contextmanager
-    def file_context(self, file_node: MypyFile, fnam: str, options: Options,
-                     active_type: Optional[TypeInfo] = None,
-                     scope: Optional[Scope] = None) -> Iterator[None]:
-        # TODO: Use this above in visit_file
-        scope = scope or self.scope
+    def file_context(self,
```
ilevkivskyi: Would be great to also add a docstring for this one.

PR #6889, opened 2019-05-23
New semantic analyzer: refactor

JukkaL:
Group related methods together. Add some additional headings.
Rename a method.

There are no non-trivial changes.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3445,21 +3445,6 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                     expr.fullname = n.fullname
                     expr.node = n.node
 
-    def process_placeholder(self, name: str, kind: str, ctx: Context) -> None:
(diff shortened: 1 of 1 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -3810,7 +3665,7 @@ def tvar_scope_frame(self, frame: TypeVarScope) -> Iterator[None]:
 
     def lookup(self, name: str, ctx: Context,
                suppress_errors: bool = False) -> Optional[SymbolTableNode]:
-        """Look up an unqualified name in all active namespaces."""
+        """Look up an unqualified (no dots) name in all active namespaces."""
```
ilevkivskyi: Just another random idea: split the `Helpers` "section" into `Lookup methods` (and move everything related there), and `Misc helpers`.

JukkaL: A good idea. I'll do this in my next refactoring PR.

PR #6894, opened 2019-05-23
New semantic analyzer: fix issues with bad base classes

### LINKED ISSUE #6495: New semantic analyzer: crash on cyce in inheritance hierarchy

### LINKED ISSUE #6892: New semantic analyzer: max iteration count reached when base class is bad
This code causes max iteration count error when using the new semantic analyzer:

```py
class C:
    class E: pass

    def f(self) -> None:
        class D(self.E): # Should generate error here
            pass
```

JukkaL:
Now we detect additional problems in base classes and generate a
proper error instead of deferring indefinitely and generating an
internal error.

There are two fixes:

1. Don't treat replacing a placeholder node that may be a type with
   another placeholder that may be type as progress.
2. Don't generate a placeholder type for base class during the final
   iteration.

Also did some refactoring.

Fixes #6495.
Fixes #6892.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -405,9 +405,11 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 var = Var(name, an_type)
                 var._fullname = self.qualified_name(name)
                 var.is_ready = True
-                self.add_symbol(name, var, None)
+                self.add_symbol(name, var, dummy_context())
             else:
-                self.add_symbol(name, PlaceholderNode(self.qualified_name(name), file_node), None)
+                self.add_symbol(name,
+                                PlaceholderNode(self.qualified_name(name), file_node),
+                                dummy_context())
 
     def add_builtin_aliases(self, tree: MypyFile) -> None:
         """Add builtin type aliases to typing module.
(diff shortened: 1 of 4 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -405,9 +405,11 @@ def add_implicit_module_attrs(self, file_node: MypyFile) -> None:
                 var = Var(name, an_type)
                 var._fullname = self.qualified_name(name)
                 var.is_ready = True
-                self.add_symbol(name, var, None)
+                self.add_symbol(name, var, dummy_context())
```
JukkaL: The idea is to use `dummy_context()` instead of `None` to avoid some special cases.

mypy/newsemanal/semanal.py
```diff
@@ -4149,6 +4144,9 @@ def defer(self) -> None:
 
         This must be called if something in the current target is
         incomplete or has a placeholder node.
+
+        This must not be called during the final analysis iteration!
+        Instead, an error should be generated.
```
msullivan: As a follow-up to this, it might be worth adding an assert to this effect and seeing if anything more shakes out.

JukkaL: A good idea! We already have a check for this later, but generating an error here would produce a useful traceback.

JukkaL:
> We might need to make some separate dedicated effort to make plugins work well with the new analyzer

Agreed, but it doesn't seem to be to be the most urgent thing just yet.

PR #6899, opened 2019-05-25
Support PEP 572

JelleZijlstra:
A couple of tricky aspects:
- We skip various parts of semanal on expressions that we know won't affect the symbol table, but with PEP 572 pretty much anything can contain an assignment, so we can no longer do that. I fixed a few cases but there may be more.
- The special semantics in https://www.python.org/dev/peps/pep-0572/#scope-of-the-target required some extra work.


test-data/unit/check-python38.test
```diff
@@ -181,3 +181,122 @@ def f(p1: bytes, p2: float, /, p_or_kw: int) -> None:
...
+    reveal_type(y7)  # E: Revealed type is 'builtins.int'
+
+    reveal_type((lambda: (y8 := 3) and y8)())  # E: Revealed type is 'builtins.int'
+
+    y7 = 1.0  # E: Incompatible types in assignment (expression has type "float", variable has type "int")
+    if y7 := "x":  # E: Incompatible types in assignment (expression has type "str", variable has type "int")
+        pass
+
+    # Just make sure we don't crash on this sort of thing.
+    if (NT := NamedTuple("NT", [("x", int)])):  # E: "int" not callable
```
JelleZijlstra: This is probably because we define special forms as `NamedTuple = 0` in the stub. If we think users are likely to actually try this, I can work on a better error message.

mypy/newsemanal/semanal.py
```diff
@@ -2396,16 +2410,22 @@ def check_and_set_up_type_alias(self, s: AssignmentStmt) -> bool:
...
+                       is_final: bool = False,
+                       escape_comprehensions: bool = False) -> None:
         """Analyze an lvalue or assignment target.
 
         Args:
             lval: The target lvalue
             nested: If true, the lvalue is within a tuple or list lvalue expression
             explicit_type: Assignment has type annotation
+            escape_comprehensions: If we are inside a comprehension, set the variable
+            in the enclosing scope instead. This implements
```
JukkaL: Style nit: indent the second and following lines in an argument specification like this:

```
    """
    ...

    Args:
        argument: Something something ...
            explanation continues here ...
```

mypy/semanal.py
```diff
@@ -2107,9 +2116,15 @@ def analyze_lvalue(self, lval: Lvalue, nested: bool = False,
             nested: If true, the lvalue is within a tuple or list lvalue expression
             add_global: Add name to globals table only if this is true (used in first pass)
             explicit_type: Assignment has type annotation
+            escape_comprehension: If we are inside a comprehension, set the variable
+            in the enclosing scope instead. This implements
```
JukkaL: Style nit: indent second and third lines, similar to above.

test-data/unit/check-python38.test
```diff
@@ -181,3 +181,122 @@ def f(p1: bytes, p2: float, /, p_or_kw: int) -> None:
 def f(p1: bytes, p2: float, /) -> None:
     reveal_type(p1)  # E: Revealed type is 'builtins.bytes'
     reveal_type(p2)  # E: Revealed type is 'builtins.float'
+
+[case testWalrus]
+from typing import NamedTuple
+
+if (a := 2):
```
JukkaL: I think that the parentheses are optional here. Maybe leave out all optional parentheses?

test-data/unit/check-python38.test
```diff
@@ -181,3 +181,122 @@ def f(p1: bytes, p2: float, /, p_or_kw: int) -> None:
...
+    if (Alias := int):
+        z3: Alias  # E: Invalid type "Alias"
+
+    return (y8 := 3) + y8
+
+reveal_type(c)  # E: Revealed type is 'builtins.int'
+
+[builtins fixtures/f_string.pyi]
+
+[case testWalrusNewSemanal]
```
JukkaL: All the tests are also ran with the new semantic analyzer in CI, so duplicating the test case is probably not necessary. If there are some differences between the old and the new semantic analyzer, it's better to isolate the differences into additional test cases and share most of the test case.

test-data/unit/check-python38.test
```diff
@@ -181,3 +181,122 @@ def f(p1: bytes, p2: float, /, p_or_kw: int) -> None:
...
+        pass
+
+    # Just make sure we don't crash on this sort of thing.
+    if (NT := NamedTuple("NT", [("x", int)])):  # E: "int" not callable
+        z2: NT  # E: Invalid type "NT"
+
+    if (Alias := int):
+        z3: Alias  # E: Invalid type "Alias"
+
+    return (y8 := 3) + y8
```
JukkaL: Is this resusing the name `y8` from the lambda above by design? What about testing whether a variable defined inside a lambda using `:=` escapes to the outer scope?

test-data/unit/check-python38.test
JelleZijlstra: I didn't realize that we run all tests with the new semanal. I'll remove the duplicate test case.

test-data/unit/check-python38.test
JelleZijlstra: Good point; I'll use a different variable name here and check that the one from the lambda doesn't escape.

JelleZijlstra:
Thanks for the review! I'll address your comments soon. For the two bigger issues you flag, I think it's better to handle them in separate PRs.

mypy/checkexpr.py
```diff
@@ -2421,6 +2421,11 @@ def check_list_multiply(self, e: OpExpr) -> Type:
         e.method_type = method_type
         return result
 
+    def visit_assignment_expr(self, e: AssignmentExpr) -> Type:
+        value = self.accept(e.value)
+        self.chk.check_assignment(e.target, e.value)
```
ilevkivskyi: It seems to me this direct call will result in allowing assignments to final variables.

test-data/unit/check-python38.test
```diff
@@ -181,3 +181,122 @@ def f(p1: bytes, p2: float, /, p_or_kw: int) -> None:
...
+
+reveal_type(c)  # E: Revealed type is 'builtins.int'
+
+[builtins fixtures/f_string.pyi]
```
ilevkivskyi: I would like to see more checker related tests here, namely:
* Interactions with final names/attributes (see the comment above)
* Interactions with binder (this will probably work as is, but without a test factoring out binder code from `check_assignment()` to containing method will cause the same problem as with `Final` now)
* Interactions with partial types and deferred nodes (mostly for the same reason as above)

mypy/checkexpr.py
JelleZijlstra: Good catch, I'll add `self.chk.check_final`.

test-data/unit/check-python38.test
JelleZijlstra: Adding this now. I'm omitting final attributes because only names are allowed in the LHS of :=.

```
$ python -c 'if self.x := 2: pass'
  File "<string>", line 1
SyntaxError: cannot use named assignment with attribute
```
Also, what would be a good test for a deferred node? Perhaps a variable assigned in an assignment expression with a type that is only defined later in the file?

test-data/unit/check-python38.test
```diff
@@ -240,4 +242,26 @@ def f(x: int = (c := 4)) -> int:
...
+        pass
+
+def check_binder(x: Optional[int], y: Optional[int]) -> None:
+    reveal_type(x)  # E: Revealed type is 'Union[builtins.int, None]'
+
+    (x := 1)
+    reveal_type(x)  # E: Revealed type is 'builtins.int'
+
+    if x or (y := 1):
+        reveal_type(y)  # E: Revealed type is 'Union[builtins.int, None]'
```
ilevkivskyi: Maybe also add a test that the same `if` where `or` is replaced with `and` will result in `y` being `int`?

test-data/unit/check-python38.test
ilevkivskyi: > Perhaps a variable assigned in an assignment expression with a type that is only defined later in the file?

Yes, r.h.s. can be an inferred attribute of some class that is initialized with a function call.

test-data/unit/check-python38.test
```diff
@@ -240,4 +242,26 @@ def f(x: int = (c := 4)) -> int:
...
+
+def check_partial() -> None:
+    x = None
+    if bool() and (x := 2):
+        pass
```
ilevkivskyi: Maybe add a similar test where you first assign `(x := [])`, and then append something to `x`.

test-data/unit/check-python38.test
JelleZijlstra: Interesting, it actually infers the type as Optional[int] if I do that. I'll take a deeper look.

ilevkivskyi:
What is the status here? Is there a chance this will be ready by the end of the week?

JelleZijlstra:
I realized your request in https://github.com/python/mypy/pull/6899#discussion_r290539995 is somewhat hard to resolve. Would it be OK to handle that in a followup PR?

ilevkivskyi:
> Would it be OK to handle that in a followup PR?

If it is tricky then OK. @JukkaL will you have time for a second round of review this week? If not I will review this on the weekend.

JelleZijlstra:
OK, tests pass again now.

Known issues that aren't covered:
- The binder interaction doesn't work properly for assignments in the right-hand side of an `and` (after `if x and (y := 1):`, the type of `y` isn't narrowed to `int`).
- Interaction with `--allow-redefinition` isn't tested
- Interaction with `mypy.server.deps` isn't tested

I'll create followups for these if this PR is approved.

gvanrossum:
Thank you! I don't think I can review this (my mypy internals are too rusty), but I tried this on some real code I've written recently, and stumbled upon an issue.
```py
from typing import Optional

class C:
    def foo(self):
        pass

def good(b: Optional[C]) -> None:
    a = b
    if a:
        a.foo()

def bad(b: Optional[C]) -> None:
    if a := b:
        a.foo()  # E:  Item "None" of "Optional[C]" has no attribute "foo"
```
Since the two functions are equivalent, the second should not get that error.

gvanrossum:
Oh I guess that's your first bullet...

JelleZijlstra:
No, that's separate from the issue I was aware of, since there's no `and` involved. I'll take a look to see if I can fix it.

JelleZijlstra:
Yes, sorry for not getting around to fix the other remaining issues. I'll fix the merge conflict and merge this PR over the next few days, and open issues for the remaining problems. (I have a feeling we'll continue to run into PEP 572/binder edge cases for a while, so it would make sense to have multiple issues we can organize under a topic.)

PR #6900, opened 2019-05-26
PEP 570: Positional-Only Arguments

### LINKED ISSUE #6878: Support 3.8 positional-only arguments
PEP 570 (https://www.python.org/dev/peps/pep-0570/) introduces positional-only arguments for 3.8, which we should support before the 3.8 release.

brandtbucher:
Fixes #6878. This was actually much more straightforward than I anticipated; since we already support `__some` forms of positional-only arguments, it was just a matter of adding a couple of lines to properly pick up the new `posonlyargs` attribute from the AST.

The tricky part is the tests. Travis doesn't currently have any build environments for 3.8.0a4 (even the `python: "nightly"` image is only 3.8.0a3+... :neutral_face:), so these tests fail with syntax errors. I've skipped them for now, but they pass fine on my local Python build. I added a small check in `testcheck.py` that will emit a warning when it's safe to unskip these tests.

### DIFF
--- a/test-data/unit/check-38.test
+++ b/test-data/unit/check-38.test
@@ -106,3 +106,69 @@ def g(x: int): ...
     /
     0  # type: ignore
 )  # type: ignore  # E: unused 'type: ignore' comment
+
+[case testPEP570ArgTypesMissing]
+# flags: --disallow-untyped-defs
+def f(arg, /) -> None: ...  # E: Function is missing a type annotation for one or more arguments
+
+[case testPEP570ArgTypesBadDefault]
+def f(arg: int = "ERROR", /) -> None: ...  # E: Incompatible default for argument "arg" (default has type "str", argument has type "int")
+
+[case testPEP570ArgTypesDefault]
+def f(arg: int = 0, /) -> None:
+    reveal_type(arg)  # E: Revealed type is 'builtins.int'
+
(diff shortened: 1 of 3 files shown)


test-data/unit/check-38.test
```diff
@@ -106,3 +106,71 @@ def g(x: int): ...
...
+[case testPEP570Default-skip]
+def f(arg: int = 0, /) -> None: ...  # N: "f" defined here
+f()
+f(1)
+f("ERROR")  # E: Argument 1 to "f" has incompatible type "str"; expected "int"
+f(arg=1)  # E: Unexpected keyword argument "arg" for "f"
+f(arg="ERROR")  # E: Unexpected keyword argument "arg" for "f"
+
+[case testPEP570Signatures1-skip]
+def f(p1: bytes, p2: float, /, p_or_kw: int, *, kw: str) -> None:
```
JukkaL: Try also calling `f` with valid and invalid arguments. For example, make sure `p1` is not a valid keyword argument but `p_or_kw` is both a valid positional and keyword argument, and `kw` must be a keyword argument.

brandtbucher:
Thanks for the suggestion @JukkaL! I've included the new test in my patch.

JukkaL:
I verified manually that the tests pass on Python 3.8a4.

PR #6908, opened 2019-05-29
Allow non-blocking parse errors in fine-grained mode

msullivan:
We do this by storing the errors generated at parse time and replaying
them when the module is reprocessed.

This restores consistency between the daemon and non-daemon for
type comments with the wrong number of arguments.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -42,7 +42,7 @@ from mypy.newsemanal.semanal import NewSemanticAnalyzer
 from mypy.newsemanal.semanal_main import semantic_analysis_for_scc
 from mypy.checker import TypeChecker
 from mypy.indirection import TypeIndirectionVisitor
-from mypy.errors import Errors, CompileError, report_internal_error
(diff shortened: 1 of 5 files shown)


test-data/unit/fine-grained.test
```diff
@@ -8901,16 +8901,29 @@ def bar(x):
     # type: () -> None
     pass
 
+def baz(x):
+    # type: () -> None
+    pass
+
+[file c.py]
+def bar(x):
+    # type: () -> None
```
JukkaL: What if the error is generated inside a nested function (i.e. not at module top level)?

mypy/build.py
```diff
@@ -1654,6 +1654,10 @@ class State:
     # Whether the module has an error or any of its dependencies have one.
     transitive_error = False
 
+    # Errors reported before semantic analysis, to allow fine-grained
+    # mode to keep reporting them
```
blueyed:
```suggestion
    # mode to keep reporting them.
```

PR #6912, opened 2019-05-29
Disable travis builds on 3.8

msullivan:
The new argument to PyCode_New means that lxml doesn't work on 3.8
right now, so we're out of business until there's a new lxml release
that does. (If that is taking too long, we can disable the tests that
want lxml.)

There are also some other minor issues: we fail some tests with
`DeprecationWarning: "@coroutine" decorator is deprecated since Python 3.8, use "async def" instead` while running some of the eval tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,8 +41,8 @@ jobs:
     - TOXENV=py
     - EXTRA_ARGS="-n 12"
     - TEST_MYPYC=1
-  - name: "run test suite with python 3.8-dev"
-    python: 3.8-dev
+  # - name: "run test suite with python 3.8-dev"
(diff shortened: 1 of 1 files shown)


blueyed:
How about skipping those tests when lxml is not installed, and making it an optional requirement for tests in general?
Building lxml can take quite a while, in case no wheel is used, so that would help with making tests more lightweight in that case also.

JukkaL:
> How about skipping those tests when lxml is not installed, and making it an optional requirement for tests in general?

This is a good idea. We just need to be careful to run at least some lxml tests in CI, and we probably need to run them on all major platforms (but not every platform/Python version combination probably).

PR #6916, opened 2019-05-31
New semantic analyzer: fix daemon crash during AST strip 

### LINKED ISSUE #6914: New semantic analyzer: crash in AST strip
This daemon test case crashes with a KeyError:

```
[case testStripNewAnalyzer]
# flags: --ignore-missing-imports
import a
[file a.py]
from typing import List
from b import B

class A:
    def __init__(self) -> None:
        self.x: List[int] = []

    def method(self) -> None:
        B()
        self.x = []

[file b.py]
class B: ...
[delete b.py.2]

[builtins fixtures/list.pyi]
[out]
==
```

Here's the traceback:

```
data: /Users/jukka/src/mypy/test-data/unit/fine-grained.test:8932:
/Users/jukka/src/mypy/mypy/test/testfinegrained.py:122: in run_case
    new_messages = self.run_check(server, sources)
/Users/jukka/src/mypy/mypy/test/testfinegrained.py:194: in run_check
    response = server.check(sources)
/Users/jukka/src/mypy/mypy/dmypy_server.py:366: in check
    res = self.fine_grained_increment(sources)
/Users/jukka/src/mypy/mypy/dmypy_server.py:463: in fine_grained_increment
    messages = self.fine_grained_manager.update(changed, removed)
/Users/jukka/src/mypy/mypy/server/update.py:241: in update
    result = self.update_one(changed_modules, initial_set, removed_set)
/Users/jukka/src/mypy/mypy/server/update.py:307: in update_one
    result = self.update_module(next_id, next_path, next_id in removed_set)
/Users/jukka/src/mypy/mypy/server/update.py:389: in update_module
    targets_with_errors=set(), processed_targets=self.processed_targets)
/Users/jukka/src/mypy/mypy/server/update.py:835: in propagate_changes_using_dependencies
    triggered |= reprocess_nodes(manager, graph, id, nodes, deps, processed_targets)
/Users/jukka/src/mypy/mypy/server/update.py:953: in reprocess_nodes
    new_patches = strip_target_new(deferred.node)
/Users/jukka/src/mypy/mypy/server/aststripnew.py:35: in strip_target_new
    node.accept(visitor)
/Users/jukka/src/mypy/mypy/nodes.py:655: in accept
    return visitor.visit_func_def(self)
/Users/jukka/src/mypy/mypy/server/aststripnew.py:122: in visit_func_def
    super().visit_func_def(node)
/Users/jukka/src/mypy/mypy/traverser.py:52: in visit_func_def
    self.visit_func(o)
/Users/jukka/src/mypy/mypy/traverser.py:49: in visit_func
    o.body.accept(self)
/Users/jukka/src/mypy/mypy/nodes.py:978: in accept
/Users/jukka/src/mypy/mypy/server/aststripnew.py:60: in visit_block
    super().visit_block(b)
/Users/jukka/src/mypy/mypy/traverser.py:37: in visit_block
    s.accept(self)
/Users/jukka/src/mypy/mypy/nodes.py:1036: in accept
/Users/jukka/src/mypy/mypy/server/aststripnew.py:152: in visit_assignment_stmt
    self.process_lvalue_in_method(lvalue)
/Users/jukka/src/mypy/mypy/server/aststripnew.py:203: in process_lvalue_in_method
    del self.type.names[lvalue.name]
E   KeyError: 'x'
```

JukkaL:
Fix occasional crash when method and top level are updated at 
the same time.

Stripping the top level removes attributes from class symbol
table. Previously we could process a method after stripping and
infer that initialization happens in the wrong place, resulting
in two initializers for a single attribute. This could result in the 
same attribute being stripped twice, which causes a KeyError.

This fixes the issue by re-applying removed attributes before
processing any methods. There is an extra complication in that
if we re-process a method that defines an attribute, we can't
re-apply that attribute as the type could change. Added logic 
to handle that by filtering out saved attributes when their 
initialization assignment is stripped. I changed the data 
structure used for this away from a list of patch functions to 
something that allows us to examine the data about saved 
attributes.

Fixes #6914.


JukkaL:
Test coverage is a bit limited. I can add more tests in a separate PR after this PR has been merged, or to this PR next week if this hasn't been merged yet.

test-data/unit/fine-grained.test
```diff
@@ -8928,3 +8928,26 @@ class A(Generic[T]): ...
...
+        B()
+        self.x = []
+
+[file b.py]
+class B: ...
+[delete b.py.2]
+
+[builtins fixtures/list.pyi]
+[out]
+==
```
ilevkivskyi: Could you please add the test for the scenario you mentioned in the PR description (when the type of variable changes)?

mypy/server/aststripnew.py
```diff
@@ -18,8 +18,11 @@
 from mypy.server.aststrip import nothing
 
 
-def strip_target_new(node: Union[MypyFile, FuncDef, OverloadedFuncDef]
-                     ) -> List[Callable[[], None]]:
+SavedAttributes = Dict[Tuple[ClassDef, str], SymbolTableNode]
+
+
+def strip_target_new(node: Union[MypyFile, FuncDef, OverloadedFuncDef],
+                     saved_attrs: SavedAttributes) -> None:
```
ilevkivskyi: Could you please update the docstring with the new argument?

ilevkivskyi:
I am going to merge this now, so that we can use this internally, you can address comments later.

PR #6917, opened 2019-06-01
tests: re-enable py38, skip with missing/broken lxml

blueyed:


mypy/report.py
```diff
@@ -35,7 +35,7 @@
     import lxml.etree  # type: ignore
     from lxml import etree  # type: ignore
     LXML_INSTALLED = True
-except ImportError:
+except Exception:  # ImportError, or SystemError (py38-dev currently)
```
blueyed: Not really sure about this one.. could/should at least be `(ImportError, SystemError)` then maybe.
Fixes https://travis-ci.org/python/mypy/jobs/540201447#L358

blueyed:
Now it appears to fail because of py38 itself: https://travis-ci.org/python/mypy/jobs/540209205#L357

blueyed:
Mostly/only:

> DeprecationWarning: "@coroutine" decorator is deprecated since Python 3.8, use "async def" 

But still at least one related to lxml:

> AttributeError: module 'lxml.etree' has no attribute 'Element'

(https://travis-ci.org/python/mypy/jobs/540193652#L635)

mypy/report.py
ethanhs: A system error implies something is very broken (often an internal interpreter error). We should not be catching SystemError. I think this implies we should just not install lxml under 3.8 until it is fixed. I'd recommend modifying `test-requirements.txt` so that it looks like `lxml; python_version < '3.8'`. Then lxml will not be installed and we won't have to deal with it being broken on 3.8.

blueyed:
> I think it would be best not to install lxml on 3.8 for the time being

Sounds good - just noticed that it was pinned to an older version, trying an upgrade there first.

> We should not be catching SystemError. 

:+1: reverted

blueyed:
Ok, also broken with lxml==4.3.3.
Disabled now, and rebased to include 580e0e44.

blueyed:
Needs more skipping...

What to do about the following?

> DeprecationWarning: "@coroutine" decorator is deprecated since Python 3.8, use "async def"

blueyed:
FWIW, locally typed-ast fails to build for me (current cpython master), so that might be a next issue here then maybe also.

setup.py
```diff
@@ -178,7 +178,7 @@ def run(self):
       classifiers=classifiers,
       cmdclass=cmdclass,
       # When changing this, also update test-requirements.txt.
-      install_requires=['typed-ast >= 1.3.5, < 1.4.0; python_version < "3.8"',
+      install_requires=['typed-ast >= 1.3.5, < 1.4.0',
```
blueyed: I wonder if that does not make sense though in general?
Even when assuming that it gets used with `--python-version` / for versions before py38, it could still be an optional dependency then, being checked during runtime?!

ethanhs:
Looks like you had a bad rebase or something?

blueyed:
> Looks like you had a bad rebase or something?

Confused by running only a single job?  (see the TEMP commits)

msullivan:
https://github.com/python/typed_ast/pull/119 should fix compilation on 3.8, then we'll do a typed-ast release. No way around requiring typed_ast, I think. (We *could* make python2 support optional but I don't like that idea.)

blueyed:
@msullivan 
Thanks, will update this after https://github.com/python/mypy/pull/6937 then.

What should we do about the DeprecationWarnings?

msullivan:
we've reenabled 3.8 without trouble a while ago

PR #6918, opened 2019-06-01
pytest.ini: add `-ra` to `addopts`

blueyed:
This displays a short test summary, which is very useful.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -13,12 +13,11 @@ python_files = test*.py
 # http://doc.pytest.org/en/latest/writing_plugins.html#collection-hooks
 
 # Both our plugin and unittest provide their own collection logic,
-# So we can disable the default python collector by giving it empty
+# so we can disable the default python collector by giving it empty
 # patterns to search for.
 # Note that unittest requires that no "Test*" classes exist.
 python_classes =
 python_functions =
 
 # always run in parallel (requires pytest-xdist, see test-requirements.txt)
-addopts = -nauto
-
+addopts = -nauto -ra


pytest.ini
```diff
@@ -13,12 +13,11 @@ python_files = test*.py
...
+# so we can disable the default python collector by giving it empty
 # patterns to search for.
 # Note that unittest requires that no "Test*" classes exist.
 python_classes =
 python_functions =
 
 # always run in parallel (requires pytest-xdist, see test-requirements.txt)
-addopts = -nauto
-
+addopts = -nauto -ra
```
blueyed: I am also not a fan of `-n auto` here really, which gets in your way when running a single test etc (but `-n0` can be used then, so it is not too bad).

ethanhs: Yeah, we have `-n0` in the developer docs IIRC, it is just more often that you want to run with `-n auto`, so that is why it is the config.

JukkaL:
It's not clear to me how the test summary is useful. Can you give an example from the mypy test suite where this provides useful information?

JukkaL:
In particular, when I tried running the tests with `-k Enum -ra` I got this which doesn't provide any useful context such as test name:

```
================================================ short test summary info ================================================
SKIP [1] /Users/jukka/src/mypy/mypy/test/data.py:220: <Skipped instance>
```

blueyed:
It is more useful with failed/errored tests, where it has nodeids.
E.g. it would be useful to see here: https://travis-ci.org/python/mypy/builds/540634510#L608

msullivan:
I'm going to close this since I don't think we get why this helps

blueyed:
Fair enough - it (or rather the core of it) might become default anyway (https://github.com/pytest-dev/pytest/issues/6454).

Issue #6901, opened 2019-05-26
reveal_type / reveal_locals should not use "error" but "note"


blueyed:
For integration of mypy's output (and in general) it would be better to have them treated as notes, not errors.

```
app.py:336: note: Revealed type is 'Any'
app.py:338: note: Revealed local types are:
app.py:338: note: content: Any
app.py:338: note: data: Any
```

While at it, it would be nice to have e.g. a `" - "` prefix (i.e. list) with the continuation lines after "Revealed local types are:" (also related: https://github.com/python/mypy/issues/6526).

```diff
diff --git i/mypy/messages.py w/mypy/messages.py
index 92b0b6ce..45d66afb 100644
--- i/mypy/messages.py
+++ w/mypy/messages.py
@@ -1068,15 +1068,15 @@ def invalid_signature_for_special_method(
         self.fail('Invalid signature "{}" for "{}"'.format(func_type, method_name), context)

     def reveal_type(self, typ: Type, context: Context) -> None:
-        self.fail('Revealed type is \'{}\''.format(typ), context)
+        self.note('Revealed type is \'{}\''.format(typ), context)

     def reveal_locals(self, type_map: Dict[str, Optional[Type]], context: Context) -> None:
         # To ensure that the output is predictable on Python < 3.6,
         # use an ordered dictionary sorted by variable name
         sorted_locals = OrderedDict(sorted(type_map.items(), key=lambda t: t[0]))
-        self.fail("Revealed local types are:", context)
+        self.note("Revealed local types are:", context)
         for line in ['{}: {}'.format(k, v) for k, v in sorted_locals.items()]:
-            self.fail(line, context)
+            self.note(line, context)

     def unsupported_type_type(self, item: Type, context: Context) -> None:
         self.fail('Unsupported type Type[{}]'.format(self.format(item)), context)
```

ilevkivskyi:
In principle I like this idea, but how will you update hundreds of existing test cases?

blueyed:
I assume the tests could be updated once, e.g using `sed`.
Will give it a try.

What about indenting following lines with `reveal_locals`?

ilevkivskyi:
> What about indenting following lines with `reveal_locals`?

I think this would make sense.

kylebebak:
Hey all, I also think `note` is more appropriate than `errors` to denote output from `reveal_locals` and `reveal_type`, but we should be aware that this was a breaking change.

[...]

PR #6919, opened 2019-06-01
Use level "note" for reveal_type and reveal_locals

blueyed:
Fixes https://github.com/python/mypy/issues/6901.

TODO:

- [x] indentation with `reveal_locals`

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -456,8 +456,8 @@ to see the types of all local variables at once. Example:
    b = 'one'
    reveal_locals()
    # Revealed local types are:
-   # a: builtins.int
-   # b: builtins.str
+   #     a: builtins.int
+   #     b: builtins.str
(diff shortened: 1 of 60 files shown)


blueyed:
Great.

> so I think this requires approval from most of core team members.

Ok, no hurry - I can see that this might cause trouble with pending PRs etc.

blueyed:
> * indentation with `reveal_locals`

As for this: do you think prepending " - " is good here?

ilevkivskyi:
> As for this: do you think prepending " - " is good here?

I think just four spaces offset is enough (like in protocol error message notes)

blueyed:
Ok, updated and rebased.

msullivan:
Sure! If you rebase it again I'll merge it.

blueyed:
@msullivan 
Ok, hope it stays green.

PR #6920, opened 2019-06-01
tox.ini: passenv: TRAVIS: for -n auto

blueyed:
This is required/useful for pytest-xdist's workaround to only use 2
workers on Travis [1].

Currently it uses 12 workers there [2].

1: https://github.com/pytest-dev/pytest-xdist/blob/9fcf8fa636bc69ee6cac9348a6ec20c87f2bb5e4/xdist/plugin.py#L11-L13
2: https://travis-ci.org/python/mypy/jobs/540201314#L256


blueyed:
Continued in https://github.com/python/mypy/pull/6921 - `-n 12` was passed explicitly..

PR #6921, opened 2019-06-01
ci: Travis: use PYTEST_ADDOPTS for xdist workers (2)

blueyed:
Only use 2 xdist workers.

Followup to https://github.com/python/mypy/pull/6920.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ dist: xenial
 
 env:
   TOXENV=py
-  EXTRA_ARGS="-n 12"
+  PYTEST_ADDOPTS="-n 2"
   TEST_MYPYC=0
 
 jobs:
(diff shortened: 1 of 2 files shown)


blueyed:
So total build time on Travis is a bit better:

> Total time 1 hr 15 min 32 sec

vs two other builds (one PR, previous master)

>  Total time 1 hr 24 min 40 sec
>  Total time 1 hr 23 min 29 sec

JukkaL:
Previously when we experimented with this, `-n 12` gave better performance than `-n 2`, even though Travis documentation implied that it shouldn't be the case. Things may have changed since then, however, and it's a good idea to try different things.

> Total time 1 hr 15 min 32 sec

I'm not sure if this is the right metric to optimize. The interesting metric is the wall clock time from creating a PR to a finished build, which I believe is different from this since we run multiple things in parallel. Can you check if this change has an effect on the observed wall clock time (long pole)?

blueyed:
I've not tried to optimize that specifc metric, but it is a good indicator for improved (total) performance (or rather at least no obvious regression).
It includes the flake8 job etc.

I do not really feel like looking more into details here, but I've seen that creating the 12 workers alone adds noticable overhead already while looking at some live build log.

ethanhs:
@JukkaL I believe Travis switched from Docker to only VMs so we no longer have the possibility of other workers sharing thread.

PR #6927, opened 2019-06-03
Add back a few lint checks now that flake8 is better

asottile:


mypy/build.py
```diff
@@ -499,7 +499,7 @@ def __init__(self, data_dir: str,
                  search_paths: SearchPaths,
                  ignore_prefix: str,
                  source_set: BuildSourceSet,
-                 reports: Optional['Reports'],
+                 reports: 'Optional[Reports]',
```
asottile: this one is subtle, `pyflakes` knows how to look at string-annotations but hasn't yet figured out string-generics, there's an open issue for this however (will need to handle `Literal['foo']` differently, and a few other things (why it wasn't tackled quite yet)) https://github.com/PyCQA/pyflakes/issues/447

mypy/fastparse.py
```diff
@@ -105,7 +105,7 @@ def ast3_parse(source: Union[str, bytes], filename: str, mode: str,
         Constant = Any
 except ImportError:
     try:
-        from typed_ast import ast35  # type: ignore
+        from typed_ast import ast35  # type: ignore  # noqa: F401
```
asottile: I considered changing this to `__import__('typed_ast').ast35` to avoid the lint error but decided to one-off exclude

mypy/messages.py
```diff
@@ -294,7 +294,7 @@ def format_bare(self, typ: Type, verbosity: int = 0) -> str:
                         arg_strings.append(
                             self.format_bare(
                                 arg_type,
-                                verbosity = max(verbosity - 1, 0)))
+                                verbosity=max(verbosity - 1, 0)))
```
asottile: this was the only thing keeping back E251 -- since this isn't a function definition I went ahead and fixed it

setup.cfg
```diff
@@ -31,20 +31,16 @@ exclude =
   .Python
 
 # Things to ignore:
-#   E251: spaces around default arg value (against our style)
 #   E128: continuation line under-indented (too noisy)
-#   F401: unused identifiers (useless, as it doesn't see inside # type: comments)
 #   W601: has_key() deprecated (false positives)
 #   E701: multiple statements on one line (colon) (we use this for classes with empty body)
-#   W503: line break before binary operator
-#   W504: line break after binary operator
```
asottile: by switching from `ignore` to `extend-ignore` (new in flake8 3.7) you no longer have to re-enable the default-ignored things

tox.ini
```diff
@@ -44,7 +44,7 @@ parallel_show_output = True
 [testenv:lint]
 description = check the code style
 basepython = python3.7
-commands = flake8 -j0 {posargs}
+commands = flake8 {posargs}
```
asottile: `-j0` is the default so I removed it

asottile:
I also wrote a flake8 plugin that might be of use to `mypy`: https://github.com/asottile/flake8-typing-imports (since I noticed you test against 3.5.1 -- presumably to check the typing imports -- if that's the only reason you may be able to lean on linting instead and speed up your test matrix 🤷‍♂)

mypy/fastparse.py
ethanhs: I think that (or even better `importlib.import_module`) would give mypy issues as well, so perhaps it is better to do it this way.

mypy/messages.py
ethanhs: The comment for ignoring E251 indicates that this is against our style, so perhaps this should be left as-is?

setup.cfg
```diff
@@ -31,20 +31,16 @@ exclude =
   .Python
 
 # Things to ignore:
-#   E251: spaces around default arg value (against our style)
 #   E128: continuation line under-indented (too noisy)
-#   F401: unused identifiers (useless, as it doesn't see inside # type: comments)
```
ethanhs: It's great to see this usable with type comments now :)

mypy/messages.py
JelleZijlstra: I think the comment was about spaces in typed function definitions (`def f(x : int = 0)`); I don't think mypy's style requires parentheses around `=` in calls.

ethanhs: Ah that is definitely possible.

JukkaL:
I'm particularly glad that flake8 can now find unused imports!

mypy/messages.py
asottile: and pycodestyle fixed the false-positive there pretty quickly :) (it used to get angry about type-annotated default-arg functions)

PR #6929, opened 2019-06-03
test-requirements: unpin lxml

blueyed:
It appears to have been pinned initially without a specific reason (in
79bce277).


blueyed:
Amended, see commit message.
Changed to `>=4.2.4`.

PR #6932, opened 2019-06-03
Use new semantic analyzer in self check

JukkaL:
It's good to dogfood new features before release.

### DIFF
--- a/mypy_self_check.ini
+++ b/mypy_self_check.ini
@@ -13,3 +13,4 @@ warn_redundant_casts = True
 warn_unused_configs = True
 show_traceback = True
 always_false = MYPYC
+new_semantic_analyzer = True


JelleZijlstra:
We should probably use both in typeshed; if we use only the new semanal errors in the stubs could creep in and cause issues for mypy users who are still using the old semanal.

ilevkivskyi:
> We should probably use both in typeshed; if we use only the new semanal errors in the stubs could creep in and cause issues for mypy users who are still using the old semanal.

Indeed, makes sense!

JukkaL:
> We should probably use both in typeshed

Agreed. Hopefully we can drop the old semantic analyzer in not-too-distant future.

PR #6934, opened 2019-06-03
Suggest trying master branch on crash

ethanhs:
I noticed that several recent crash reports were from people on older mypy versions. I thought it would be good to suggest people try the latest master as that may have needed fixes. Here is a sample of the output:

```
test.py:5: error: INTERNAL ERROR -- Please try using master
https://mypy.rtfd.io/en/latest/common_issues.html#using-pre-release-mypy
If this crash persists, please report a bug at https://github.com/python/mypy/issues
version: 0.710+dev.bde68e77bc6fd209fce765f3ddcc9cb2b4c43e77.dirty
test.py:5: : note: please use --show-traceback to print a traceback when reporting a bug
```

I also edited the docs section so the title/link wouldn't be so long.

### DIFF
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -618,11 +618,16 @@ def report_internal_error(err: Exception,
 
     # Print "INTERNAL ERROR" message.
     print('{}error: INTERNAL ERROR --'.format(prefix),
-          'Please try using master\n'
-          'https://mypy.rtfd.io/en/latest/common_issues.html#using-pre-release-mypy',
+          'Please try using mypy master on Github:\n'
+          'https://mypy.rtfd.io/en/latest/common_issues.html#using-development-mypy-build',
           file=stderr)
-    print('If this crash persists, please report a bug at https://github.com/python/mypy/issues',
-          file=stderr)
+    if options.show_traceback:
+        print('Please report a bug at https://github.com/python/mypy/issues',
+            file=stderr)
+    else:
+        print('If this issue continues with mypy master, '
+              'please report a bug at https://github.com/python/mypy/issues',
+            file=stderr)
     print('version: {}'.format(mypy_version),
           file=stderr)
 
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -618,8 +618,12 @@ def report_internal_error(err: Exception,
 
     # Print "INTERNAL ERROR" message.
     print('{}error: INTERNAL ERROR --'.format(prefix),
-          'please report a bug at https://github.com/python/mypy/issues',
-          'version: {}'.format(mypy_version),
+          'Please try using master\n'
+          'https://mypy.rtfd.io/en/latest/common_issues.html#using-pre-release-mypy',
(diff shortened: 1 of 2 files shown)


mypy/errors.py
```diff
@@ -618,8 +618,12 @@ def report_internal_error(err: Exception,
 
     # Print "INTERNAL ERROR" message.
     print('{}error: INTERNAL ERROR --'.format(prefix),
-          'please report a bug at https://github.com/python/mypy/issues',
-          'version: {}'.format(mypy_version),
+          'Please try using master\n'
+          'https://mypy.rtfd.io/en/latest/common_issues.html#using-pre-release-mypy',
```
JukkaL: This link doesn't seem to point to the correct section (it goes to the top of the page).

mypy/errors.py
```diff
@@ -618,8 +618,12 @@ def report_internal_error(err: Exception,
...
+          file=stderr)
+    print('If this crash persists, please report a bug at https://github.com/python/mypy/issues',
```
JukkaL: I'd like to hear about all crashes with a traceback, even if they are not reproducible. Sometimes a traceback is sufficient to figure out what went wrong. What about reverting to the original message if we are displaying a traceback?

mypy/errors.py
```diff
@@ -618,8 +618,12 @@ def report_internal_error(err: Exception,
 
     # Print "INTERNAL ERROR" message.
     print('{}error: INTERNAL ERROR --'.format(prefix),
-          'please report a bug at https://github.com/python/mypy/issues',
-          'version: {}'.format(mypy_version),
+          'Please try using master\n'
```
JukkaL: Nit: it may not be clear what master means in this context. Maybe use 'mypy master on GitHub' or something.

mypy/errors.py
ethanhs: Ah, I forgot to commit a change to shorten the title of that section so the link isn't egregious.

ethanhs:
Hopefully the new wording changes are more in line with what you want.
Here are updated examples of what things look like now:

without `--show-traceback`
```
test.py:5: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-development-mypy-build
If this issue continues with mypy master, please report a bug at https://github.com/python/mypy/issues
version: 0.710+dev.7edc5c91f811468d0e5a4d98e3ea5eac59549698.dirty
test.py:5: : note: please use --show-traceback to print a traceback when reporting a bug
```

with
```
test.py:5: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 0.710+dev.7edc5c91f811468d0e5a4d98e3ea5eac59549698.dirty
... (snip)
test.py:5: : note: use --pdb to drop into pdb
```

PR #6937, opened 2019-06-04
Update typed_ast dep to 1.4.0+

gvanrossum:
Now that typed_ast 1.4.0 is released (whose main new feature is support for extra text after `# type: ignore`).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -178,7 +178,7 @@ setup(name='mypy',
       classifiers=classifiers,
       cmdclass=cmdclass,
       # When changing this, also update test-requirements.txt.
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -178,7 +178,7 @@ def run(self):
       classifiers=classifiers,
       cmdclass=cmdclass,
       # When changing this, also update test-requirements.txt.
-      install_requires=['typed-ast >= 1.3.5, < 1.4.0',
+      install_requires=['typed_ast >= 1.4.0, < 1.5.0',
```
msullivan: Update test-requirements.txt also?

PR #6938, opened 2019-06-05
New semantic analyzer: fix falling back to outer scope in class body

JukkaL:
This fixes scoping issue where code like this was rejected:

```py
c: C  # Forces second semantic analysis pass

class X: pass

class C:
    X = X   # Should be no error here
```

The approach is to fall back to the outer scope if the definition
in the class body is not ready yet, based on the line numbers of
the definitions.

Additionally generate an error for references to names defined 
later in a class body.

This only addresses class bodies. Local scopes have a similar 
issue with references to undefined names. I'm planning to 
address that in a separate PR.

The basic fix is very simple, but various other changes were
required to support multi-line definitions and certain edge cases.

Work towards #6303.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3228,6 +3228,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             self.bind_name_expr(expr, n)
 
     def bind_name_expr(self, expr: NameExpr, sym: SymbolTableNode) -> None:
+        """Bind name expression to a symbol table node."""
         if isinstance(sym.node, TypeVarExpr) and self.tvar_scope.get_binding(sym):
             self.fail("'{}' is a type variable and only valid in type "
                       "context".format(expr.name), expr)
@@ -3751,9 +3752,17 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         if self.type and not self.is_func_scope() and name in self.type.names:
             node = self.type.names[name]
             if not node.implicit:
-                # Only allow access to most class attributes after the definition
-                # so that it's possible to fall back to the outer scope
-                # (except for types).
+                # Only allow access to class attributes textually after
+                # the definition, so that it's possible to fall back to the
+                # outer scope. Example:
+                #
+                #     class X: ...
+                #
+                #     class C:
+                #         X = X  # Initializer refers to outer scope
+                #
+                # Nested classes are an exception, since we want to support
+                # arbitrary forward references in type annotations.
                 if (node.node is None
                         or node.node.line < self.statement.line
                         or isinstance(node.node, TypeInfo)):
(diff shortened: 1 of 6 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -222,6 +222,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
     # not be found in phase 1, for example due to * imports.
     errors = None  # type: Errors     # Keeps track of generated errors
     plugin = None  # type: Plugin     # Mypy plugin for special casing of library features
+    statement = None  # type: Node    # Statement/definition being analyzed
```
ilevkivskyi: This is probably bikesheding, but I would call this `cur_statement`, similar to `cur_mod_id` above.

Also, should the type be actually `Statement`?

mypy/newsemanal/semanal.py
```diff
@@ -2403,7 +2409,9 @@ def check_and_set_up_type_alias(self, s: AssignmentStmt) -> bool:
             alias_node.normalized = rvalue.node.normalized
         return True
 
-    def analyze_lvalue(self, lval: Lvalue, nested: bool = False,
+    def analyze_lvalue(self,
+                       lval: Lvalue,
+                       nested: bool = False,
```
ilevkivskyi: Please add this to the list of args in the docstring below.

mypy/newsemanal/semanal.py
```diff
@@ -2984,7 +2984,7 @@ def process_module_assignment(self, lvals: List[Lvalue], rval: Expression,
                     # respect explicitly annotated type
                     if (isinstance(lval.node, Var) and lval.node.type is not None):
                         continue
-                    lnode = self.lookup(lval.name, ctx)
+                    lnode = self.current_symbol_table().get(lval.name)
```
ilevkivskyi: Why is this limitation needed? A comment would help here.

mypy/newsemanal/semanal.py
```diff
@@ -3731,9 +3752,25 @@ def lookup(self, name: str, ctx: Context,
...
+                #     class X: ...
+                #
+                #     class C:
+                #         X = X  # Initializer refers to outer scope
+                #
+                # Nested classes are an exception, since we want to support
+                # arbitrary forward references in type annotations.
+                if (node.node is None
+                        or node.node.line < self.statement.line
+                        or isinstance(node.node, TypeInfo)):
```
ilevkivskyi: Node may be also `TypeAlias`, for example:
```python
class C:
    x: A
    A = List[int]
```
In general this looks a bit ad-hoc. Would it be better instead to add a flag to lookup function that indicates whether lookup happens from type context or from runtime context? Or will this cause problems in situations like this:
```python
class A: ...
class C:
    A = List[A]  # A looked-up in type context on r.h.s.
```

ilevkivskyi:
(Also couple tests failed.)

JukkaL:
I'll address the feedback in a separate PR so that we can deploy this PR sooner internally.

PR #6942, opened 2019-06-06
Print deferral debug trace in case of a hang

ilevkivskyi:


ilevkivskyi:
@JukkaL Should we move towards merging this? It gets merge conflicts really quick.

JukkaL:
Yes, I'll review this soon.

mypy/newsemanal/semanal.py
```diff
@@ -4544,8 +4553,11 @@ def schedule_patch(self, priority: int, patch: Callable[[], None]) -> None:
         self.patches.append((priority, patch))
 
     def report_hang(self) -> None:
+        print('Deferral trace:')
+        for mod, line in self.deferral_debug_context:
+            print('    {}: {}'.format(mod, line))
```
JukkaL: Maybe leave out the space after `:`, for consistency with mypy error message formatting.

mypy/newsemanal/semanal_main.py
```diff
@@ -168,10 +168,12 @@ def process_top_levels(graph: 'Graph', scc: List[str], patches: Patches) -> None
         worklist += list(reversed(core_modules)) * CORE_WARMUP
     final_iteration = False
     iteration = 0
+    analyzer = state.manager.new_semantic_analyzer
+    analyzer.deferral_debug_context.clear()
```
JukkaL: This doesn't work correctly inside import cycles, since we only clear one of the debug contexts.

mypy/newsemanal/semanal_main.py
```diff
@@ -168,10 +168,12 @@ def process_top_levels(graph: 'Graph', scc: List[str], patches: Patches) -> None
...
+
     while worklist:
         iteration += 1
         if iteration > MAX_ITERATIONS:
-            analyzer = state.manager.new_semantic_analyzer
             # Just pick some module inside the current SCC for error context.
             assert state.tree is not None
             with analyzer.file_context(state.tree, state.options):
```
JukkaL: Hang reporting doesn't necessarily report the correct information here, since we only report the deferral context for some module in the SCC, which might not even be participating in the final iterations.

ilevkivskyi: But we should set _some_ context. The trace printed above makes it clear (assuming the import cycle bug is fixed, I am doing this now).

mypy/newsemanal/semanal_main.py
JukkaL: I was mistaken. There is only one semantic analyzer instance.

mypy/newsemanal/semanal_main.py
JukkaL: I was mistaken about this. I was confused by the code accessing the semantic analyzer through the state object. It might be worth it to refactor it -- I can look at it.

PR #6958, opened 2019-06-08
Fix mypy tests for contextlib change in typeshed PR 3083

RafiB:
https://github.com/python/typeshed/pull/3083 fixed an issue in the `contextlib` stub file, which broke an existing mypy test that was written based off the incorrect stub file.

*NOTE*:
This commit will have to synced with that change to fix the tests + add new `decorator` tests.

### DIFF
--- a/test-data/unit/pythoneval.test
+++ b/test-data/unit/pythoneval.test
@@ -1030,8 +1030,8 @@ reveal_type(g)
 with f('') as s:
     reveal_type(s)
 [out]
-_program.py:13: note: Revealed type is 'def (x: builtins.int) -> contextlib.GeneratorContextManager[builtins.str*]'
-_program.py:14: note: Revealed type is 'def (*x: builtins.str) -> contextlib.GeneratorContextManager[builtins.int*]'
+_program.py:13: note: Revealed type is 'def (x: builtins.int) -> contextlib._GeneratorContextManager[builtins.str*]'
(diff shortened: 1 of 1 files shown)


test-data/unit/pythoneval.test
```diff
@@ -1367,3 +1367,49 @@ x: Dict[str, List[int]]
 reveal_type(x['test'][0])
 [out]
 _testNewAnalyzerBasicTypeshed_newsemanal.py:4: error: Revealed type is 'builtins.int*'
+
+[case testDecoratorDecorator]
```
ilevkivskyi: The existing `pythoneval` tests typically test some tricky but important corner cases where mypy used to not work. We only try to add `pythoneval` tests when they are really necessary, because they are very slow. I would propose to just keep the above fix in this PR.

JelleZijlstra: Also, `decorator` is a third-party library; I don't think it makes sense for mypy to test the stubs for third-party libraries.

srittau:
@JelleZijlstra @ilevkivskyi I'd like to get the corresponding typeshed fix (python/typeshed#3083) merged. What's the status of this PR?

JelleZijlstra:
Merging this will require a complicated dance of merging the typeshed PR + syncing the typeshed submodule in mypy + merging this PR. I can do it at some point but don't have time right now.

ilevkivskyi:
I can do the typeshed dance now.

ilevkivskyi:
I have restarted all builds, should be green now.

Issue #6952, opened 2019-06-07
Literal + Final don't work well on new semantic analyzer


ilevkivskyi:
This test case fails
```
[case testLiteralFinalInferredAsLiteral]
from typing_extensions import Final, Literal

defer: Yes

var: Final = 42
def force(x: Literal[42]) -> None: pass
force(var)

class Yes: ...
```
with
```
Argument 1 to "force" has incompatible type "int"; expected "Literal[42]"
```
on new semantic analyzer

cc @Michael0x2a

Michael0x2a:
This seems to be related to how we handle the type context in some way.

Mypy seems to correctly remember that the type of `var` has a `last_known_value` of 42 in each iteration of the new semantic analyzer, but the type context seems to contain just Nones in the final iteration. So, `var` is treated as just an `int` on the final iteration, which is why we get the type error.

I'll try digging a bit deeper.

Michael0x2a:
@ilevkivskyi -- actually, should we keep this one open? Both of our PRs are only kind of fixes this. 

Or maybe it's sufficient to just keep this closed and track just #6458 instead.

ilevkivskyi:
#6458 is sufficient.

Issue #6458, opened 2019-02-21
New semantic analyzer: Refactor handling of is_inferred_def & co


ilevkivskyi:
There are several this that hold information about whether a type of variable should be inferred of not:
* `RefExpr.is_new_def`
* `RefExpr.is_inferred_def`
* `Var.is_ready`
* `Var.is_inferred`
* `TypeChecker.is_definition()`

Such a variety of things is partially caused by the fact that special definitions like named tuples etc. used to have a `Var()` to which the `lvalue` was pointing. With the new analyzer we don't have this, so it would be good to clean-up/refactor the logic around inference after we merge the two analyzers.

See also https://github.com/python/mypy/issues/6453

PR #6961, opened 2019-06-08
Make `narrow_declared_type` correctly handle Literal types

Michael0x2a:
This diff fixes the `narrowed_declared_type` function to correctly handle cases where the declared type is actually a `Literal[...]`.

This fixes https://github.com/python/mypy/issues/6952. 

In short, it seems like using the new semantic analyzer somehow made mypy call [the `narrow_type_from_binder(...)` function][0] where it previously did not.

Specifically, mypy passed in the `var` expression and a `known_type` of `Literal[42]`, we got a restriction of `builtins.int` (with a `last_known_value` of 42), and attempted to narrow the `Literal[42]` on line 3512.

But due to the missing case in `narrow_declared_type`, we ended up returning just the int.

  [0]: https://github.com/python/mypy/blob/master/mypy/checkexpr.py#L3504


Michael0x2a:
> The difference in behavior is likely related to #6458, and the fact that Final gets "stripped" in the process.

I'm not too familiar with the behavior of the new semantic analyzer, but fwiw it didn't seem as if this were the case. More precisely, it seemed we did correctly remember that the type of the variable was `Literal[int]` throughout the whole process -- we just ended up inferring `builtins.int` instead of `Literal[int]` when we tried narrowing `Literal[int]` against a `builtin.int`.

> Although this fix works for the issue, I would like to actually understand why it happens before deciding whether we should use this as a fix.

I do think this fix would be useful independently from the linked issue. IIUC doing `narrow_declared_type(ChildType, ParentType)` yields just `ChildType`, so I think it makes sense to make `narrow_declared_type(Literal[42], int)` behave in the same way.

The fact that change fixes #6458 is almost a side-effect IMO.

ilevkivskyi:
> I'm not too familiar with the behavior of the new semantic analyzer, but fwiw it didn't seem as if this were the case.

I mean this https://github.com/python/mypy/pull/6988.

I still think this PR has value, but it might also mask an actual flaw in the logic in new semantic analyzer.

PR #6965, opened 2019-06-10
Keep track of fstring conversions and format-specs

SanjitKal:
This pull request modifies fastparse.py to keep track of format specifiers and conversions for f-strings. Currently, braces values within f-strings are just converted to calls to format, ignoring format specifiers and conversions ('{}'.format(...)). Now, the information regarding format specifiers and conversions are passed along within the brackets to format (ex: '{!r<30}.format(...)). Note that format specifiers can contain expressions that are evaluated at runtime. The newly generated AST for an fstring expression containing a format specifier that references a named expression looks (a little) cumbersome. I am open to suggestions regarding how to better handle this situation if needed.

Keeping track of this information is needed in order to provide mypyc with enough information to support fstrings with format specifiers and conversions.


mypy/fastparse.py
```diff
@@ -1111,15 +1114,32 @@ def visit_JoinedStr(self, n: ast3.JoinedStr) -> Expression:
...
+            empty_string = StrExpr('')
+            empty_string.set_line(n.lineno, n.col_offset)
+            join_method = MemberExpr(empty_string, 'join')
+            join_method.set_line(empty_string)
+            args = ListExpr([StrExpr('{' + conversion_str + ':'), format_spec_exp, StrExpr('}')])
+            args.set_line(join_method)
+            format_string = CallExpr(join_method,
+                                     [args],
+                                     [ARG_POS],
+                                     [None])
```
msullivan: I have an idea for another way to handle the nested computation thing: format lets you do `'{:{}}.format(x, y)` where `x` is the thing to format and `y` is the format specifier.

mypy/fastparse.py
```diff
@@ -1111,15 +1114,17 @@ def visit_JoinedStr(self, n: ast3.JoinedStr) -> Expression:
     def visit_FormattedValue(self, n: ast3.FormattedValue) -> Expression:
         # A FormattedValue is a component of a JoinedStr, or it can exist
         # on its own. We translate them to individual '{}'.format(value)
-        # calls -- we don't bother with the conversion/format_spec fields.
-        exp = self.visit(n.value)
-        exp.set_line(n.lineno, n.col_offset)
-        format_string = StrExpr('{}')
+        # calls, passing along specified format specifiers and conversions.
```
msullivan: Add a quick comment indicating that mypyc is the motivation for caring about this

PR #6966, opened 2019-06-10
Use new semantic analyzer for bootstrap

ilevkivskyi:
I am not sure we should do this _right now_. But maybe it is at least we should use it for mypy CI build (because e.g. making this work required couple small updates in mypyc)?

(IIUC, this should initially fail, until I update the mypyc pin.)


ilevkivskyi:
OK, now that the mypyc build indeed failed, I will move the pin.

ilevkivskyi:
> We can revert it if there are problems.

OK, lets do this!

PR #6970, opened 2019-06-11
Enable additional tests on new semantic analyzer

JukkaL:
Some tests are only enabled for the new semantic analyzer. For a few
I decided to create duplicate tests, but only if the test seemed
worthwhile enough, as the old semantic analyzer isn't going to change
much any more.

### DIFF
--- a/test-data/unit/check-final.test
+++ b/test-data/unit/check-final.test
@@ -485,12 +485,12 @@ class C:
 
 [case testFinalReassignInstanceVarClassVsInit]
 # flags: --new-semantic-analyzer
-# Methods are processed after top-level in new analyzer.
 from typing import Final
 
 class C:
     y: Final = 1
     def __init__(self) -> None:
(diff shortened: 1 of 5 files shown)


test-data/unit/check-final.test
```diff
@@ -487,16 +484,16 @@ class C:
...
-# flags: --no-new-semantic-analyzer
+# flags: --new-semantic-analyzer
 # Methods are processed after top-level in new analyzer.
 from typing import Final
 
 class C:
     y: Final = 1
     def __init__(self) -> None:
-        self.x: Final = 1
+        self.x: Final = 1  # E: Cannot redefine an existing name as final
```
ilevkivskyi: The error here is a bit confusing. Maybe move the comment at the top of the test case, right above the error, so that the former is more prominent?

JukkaL:
Opening and closing to retrigger CI builds. Travis may be flaky.

PR #6972, opened 2019-06-11
New analyzer: resolve name clashes between generated and existing nodes

### LINKED ISSUE #6454: New semantic analyzer: should we collect functions from AST instead of symbol tables?
Currently, functions are collected for processing from symbol tables, not form ASTs. This causes some issues, for example, if there is method name clash in class-based named tuple, then the user provided method will be not analyzed, which will cause a crash later, see `testNewNamedTupleIllegalNames`.

This is a follow-up for https://github.com/python/mypy/pull/6452

ilevkivskyi:
Fixes #6454
Fixes https://github.com/python/mypy/issues/6973

Note that the solution doesn't guarantee the order of numbers `N` in `foo-redefinitionN` coincides with the textual order of nodes. But it looks like we don't need this for anything.

### DIFF
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -4,7 +4,7 @@ import pathlib
 import re
 import subprocess
 import sys
-from typing import TypeVar, List, Tuple, Optional, Dict, Sequence, Iterable
+from typing import TypeVar, List, Tuple, Optional, Dict, Sequence, Iterable, Container
 
 MYPY = False
 if MYPY:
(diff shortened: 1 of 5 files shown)


mypy/util.py
```diff
@@ -290,6 +290,22 @@ def unmangle(name: str) -> str:
     return name.rstrip("'")
 
 
+def get_unique_redefinition_name(name: str, existing: Container[str]) -> str:
```
JukkaL: I wonder if the mangled names can be shown in error messages (say, when showing error context). If you believe that it might be the case, can you create a follow-up issue about it?

ilevkivskyi: Out of top of my head I found only one case (multiple inheritance) where we use names from symbol table instead of `.name()` or `.fullname()`, but there may be other. I will open an issue.

PR #6979, opened 2019-06-12
Specify return types for the return types incompatible with supertype error

### LINKED ISSUE #2797: Return type incompatible with supertype error should specify the types involved
I'm a relatively new user to mypy, and I have a quick bit of feedback / suggestion:

(Tangentially related to #2783 and/or #2782)

I see this error message:

```
error: Return type of "__iter__" incompatible with supertype "Iterable"
```

It would be really helpful if the message specified what the return type is, and more helpfully, what mypy thinks it should be.

This is particularly helpful in the context of generic/templated types like `Iterable`, where reading the definition of the type doesn't immediately give you the answer.  In other words: "Oh no!  I have to think!" 😃

joejuzl:
fixes #2797

### DIFF
--- a/test-data/unit/check-abstract.test
+++ b/test-data/unit/check-abstract.test
@@ -326,7 +326,7 @@ class A(I, J):
     def f(self, x: str) -> int: pass \
         # E: Argument 1 of "f" is incompatible with supertype "I"; supertype defines the argument type as "int"
     def g(self, x: str) -> int: pass \
-        # E: Return type "int" of "g" incompatible with return type "str" of supertype "J"
+        # E: Return type "int" of "g" incompatible with return type "str" in supertype "J"
     def h(self) -> int: pass # Not related to any base class
 [out]
 
@@ -372,7 +372,7 @@ class A(I):
         pass
 [out]
 main:11: error: Argument 1 of "h" is incompatible with supertype "I"; supertype defines the argument type as "I"
-main:11: error: Return type "I" of "h" incompatible with return type "A" of supertype "I"
+main:11: error: Return type "I" of "h" incompatible with return type "A" in supertype "I"
 
 
 -- Accessing abstract members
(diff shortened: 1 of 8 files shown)


test-data/unit/check-abstract.test
```diff
@@ -326,7 +326,7 @@ class A(I, J):
     def f(self, x: str) -> int: pass \
         # E: Argument 1 of "f" is incompatible with supertype "I"; supertype defines the argument type as "int"
     def g(self, x: str) -> int: pass \
-        # E: Return type of "g" incompatible with supertype "J"
+        # E: Return type "int" of "g" incompatible with return type "str" of supertype "J"
```
ilevkivskyi: I think it is better to say:
```
Return type "int" of "g" incompatible with return type "str" in supertype "J"
```
i.e. replace `of` -> `in`.

joejuzl: the `of` here refers to the function. I.e. the function "g" has a return type `of` "int".
Whereas the `in` that comes later refers to the superclass i.e. return type "str" of the function "g" `in` class "J".
Do you still think it makes sense to change the `in` to `of`? I haven't heard people refer to a return type `in` a function... (although I may be wrong!)

joejuzl: you are totally correct - sorry I misread!

joejuzl: There is a slight edge case where the function name doesn't match the superclass function name. You end up with this message:

`Return type "A" of "__iadd__" incompatible with return type "B" in "__add__" of supertype "A"` 

This due to how `override_target` is implemented. Do you think this is OK? Or should I try and handle it?

ilevkivskyi: > This due to how `override_target` is implemented. Do you think this is OK? Or should I try and handle it?

It is a bit suboptimal, but I wouldn't bother, it is already much better than it used to be.

PR #6983, opened 2019-06-13
New semantic analyzer: track line numbers of placeholders

### LINKED ISSUE #6949: New semantic analyzer: cannot resolve assignment to imported value from outer scope
This test case fails with `Cannot resolve name "A" (possible cyclic definition)`:

```py
[case testNewAnalyzerTrickyX]
import b
[file a.py]
import b
class A(B): ...
class B: ...

[file b.py]
from a import A

class C:
    A = A
```

JukkaL:
Line numbers are used sometimes to decide when a name should be looked
up from an outer scope, so they should be set for placeholders.

This actually breaks some forward references to assignment-based named
tuples. The reason is that we don't set `becomes_typeinfo=True` for
the placeholders. I'm going to open a separate issue about this, as
it's an existing issue and this change only exposes it.

Fixes #6949.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -410,7 +410,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 self.add_symbol(name, var, dummy_context())
             else:
                 self.add_symbol(name,
-                                PlaceholderNode(self.qualified_name(name), file_node),
+                                PlaceholderNode(self.qualified_name(name), file_node, -1),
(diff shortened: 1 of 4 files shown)


JukkaL:
> Would it be better to first merge fix the named tuple issue to avoid changing the test?

I think that the named tuple issue is lower priority, since forward references to named tuples aren't supported on the old semantic analyzer, so it's not a regression. Also, the fix may be quite involved and may involve changing how other things such as type aliases and NewTypes are processed.

PR #6984, opened 2019-06-13
New semantic analyzer: fix crash related to dataclasses.InitVar

### LINKED ISSUE #6955: dataclasses.InitVar crashes on new analyzer
Currently this test case crashes on the new semantic analyzer:
```
[case testDataclassesInitVars]
from dataclasses import InitVar, dataclass

defer: Yes

@dataclass
class Application:
  name: str
  database_name: InitVar[str]

class Yes: ...
[builtins fixtures/list.pyi]
```

JukkaL:
Previously InitVar attributes were missing on the second semantic
analysis pass, since the dataclasses plugin removes those attributes
from the class symbol table. This caused a crash. The fix is to reset
the attribute declarations so that they will be re-added to the symbol
table on successive semantic analysis passes.

Fixes #6955.

### DIFF
--- a/test-data/unit/check-dataclasses.test
+++ b/test-data/unit/check-dataclasses.test
@@ -520,6 +520,26 @@ app.database_name  # E: "SpecializedApplication" has no attribute "database_name
 
 [builtins fixtures/list.pyi]
 
+[case testDataclassesInitVarsAndDefer]
+# flags: --new-semantic-analyzer
+from dataclasses import InitVar, dataclass
+
+defer: Yes
+
+@dataclass
+class Application:
+  name: str
+  database_name: InitVar[str]
+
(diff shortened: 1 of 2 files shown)


test-data/unit/check-dataclasses.test
```diff
@@ -520,6 +520,26 @@ app.database_name  # E: "SpecializedApplication" has no attribute "database_name
 
 [builtins fixtures/list.pyi]
 
+[case testDataclassesInitVarsAndDefer]
+# flags: --new-semantic-analyzer
```
ilevkivskyi: Is this actually needed? Or it doesn't work well with the old analyzer?

JukkaL: No, but I added it since the test case doesn't test anything useful on the old analyzer and I didn't want to add it to `check-newsemanal.test`. We may want to move test cases out of `check-newsemanal.test` to other files once we've removed the old semantic analyzer, and this is anticipating that move.

PR #6988, opened 2019-06-13
 New semantic analyzer: Keep explicit type and inferred status in sync

ilevkivskyi:
This is (kind of) an actual fix for #6952

The problem is that previously `unwrap_final()` unconditionally set `s.type` to `None` for bare `Final`, even if the type was inferred from a simple literal r.h.s. on a previous iteration. At the same time `lvalue.is_inferred_def` remained `False`, so that the type was not inferred on the second iteration.

This way `s.type` and `lvalue.is_inferred_def` may get out of sync depending on the number of iterations, thus causing weird behavior in type checker.

I generally can say the current way of storing the inferred status is error-prone, but we already have https://github.com/python/mypy/issues/6458 to track this.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2116,6 +2116,8 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             return False
         lval = s.lvalues[0]
         assert isinstance(lval, RefExpr)
+        # Reset inferred status if it was set due to simple literal rvalue on previous iteration.
+        lval.is_inferred_def = s.type is None
         if self.loop_depth > 0:
             self.fail("Cannot use Final inside a loop", s)
         if self.type and self.type.is_protocol:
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -2118,6 +2118,8 @@ def unwrap_final(self, s: AssignmentStmt) -> bool:
             return False
         lval = s.lvalues[0]
         assert isinstance(lval, RefExpr)
+        # Reset inferred status if it was set due to simple literal rvalue on previous iteration.
+        lval.is_inferred_def = s.type is None
```
JukkaL: This seems error-prone and hard to reason about. For example, what happens if there are multiple assignments to a final name? How hard it would be to avoid the need to reset this (by not storing the bad value perhaps)?

ilevkivskyi: What do you mean by "bad" value? The value of `is_inferred_def` is set correctly, but just few lines above we reset `s.type` to `None`, but not the flag. If we don't reset it together, `s.type` will stay `None`.

> For example, what happens if there are multiple assignments to a final name?

There will be an inconsistency with old analyzer. But we anyway give an error. Also this can be fixed only doing the reset if `lvalue.is_new_def`.

Anyway, if you have a better fix, go ahead.

ilevkivskyi:
After all we decided to go with a quick-fix.

ilevkivskyi:
@JukkaL does this look acceptable?

PR #6996, opened 2019-06-14
dmypy suggest: add a flag to try using unicode

msullivan:
This allows preferring unicode in annotations generated for Python 2
code. pyannotate will spell unicode as Text, so this helps faciliate
porting to Python 3.

I've already successfully used this code in generating annotations at Dropbox.

### DIFF
--- a/test-data/unit/fine-grained-suggest.test
+++ b/test-data/unit/fine-grained-suggest.test
@@ -175,6 +175,19 @@ Foo('lol')
 (str) -> None
 ==
 
+[case testSuggestTryText]
+# flags: --py2
+# suggest: --try-text foo.foo
+[file foo.py]
+def foo(s):
+    return s
+[file bar.py]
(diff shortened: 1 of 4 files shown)


test-data/unit/fine-grained-suggest.test
```diff
@@ -175,6 +175,19 @@ Foo('lol')
...
+# flags: --py2
+# suggest: --try-text foo.foo
+[file foo.py]
+def foo(s):
+    return s
+[file bar.py]
+from foo import foo
+foo('lol')
+[out]
+(unicode) -> unicode
```
gvanrossum: Huh, why is this not Text?

msullivan: I think because the real underlying typing is called `unicode`, not `Text`. `pyannotate` will print it as `Text`, though, so it is fine in practice.

PR #6997, opened 2019-06-14
dmypy suggest: handful of bug fixes/improvements

msullivan:
* Fix --no-any to not spuriously reject all methods
 * Support inferring types of decorated methods if the decorator
   doesn't change the type
 * Refuse to infer functions with then or more arguments to avoid
   spending forever calculating an exponentially large number of
   possibilities before deciding there are too many possibilities.

I've already successfully used this code in generating annotations at Dropbox.

### DIFF
--- a/mypy/suggestions.py
+++ b/mypy/suggestions.py
@@ -29,13 +29,14 @@ from typing import (
 from mypy.state import strict_optional_set
 from mypy.types import (
     Type, AnyType, TypeOfAny, CallableType, UnionType, NoneType, Instance, TupleType, is_optional,
+    TypeVarType, FunctionLike,
     TypeStrVisitor,
 )
 from mypy.build import State, Graph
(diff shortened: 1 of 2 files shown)


mypy/suggestions.py
```diff
@@ -388,11 +394,39 @@ def find_node(self, key: str) -> Tuple[str, str, FuncDef]:
...
+                raise SuggestionFailure("Object %s is a decorator we can't handle" % key)
+
         if not isinstance(node, FuncDef):
             raise SuggestionFailure("Object %s is not a function" % key)
 
         return (modname, tail, node)
 
+    def extract_from_decorator(self, node: Decorator) -> Optional[FuncDef]:
+        for dec in node.decorators:
+            print(dec)
```
gvanrossum: Debug print.

PR #7005, opened 2019-06-17
New semantic analyzer: don't add submodules to symbol tables

### LINKED ISSUE #6828: Uncertainty between submodule and local definition in new semantic analyzer
Consider this test case:
```
[case testNewAnalyzerImportFromTopLevelAlias]
import a.b
[file a/__init__.py]
from .b import B
from . import b as c
b = int
x: Forward
class Forward:
    ...

[file a/b.py]
class B: ...
```
It currently fails with `Name 'c' is already defined`, because on the first iteration mypy sees `a.b` as module, while on the second one it is a type alias. There is also a similar but different test case:
```
[case testNewAnalyzerImportFromTopLevelFunction]
import a.b
[file a/__init__.py]
from .b import B
from . import b as c
def b(): ...  # can be also b: int
x: Forward
class Forward:
    ...

[file a/b.py]
class B: ...
```
that fails with `Name 'b' already defined` for a similar reason. Note that we already added some special casing for a similar scenario, see https://github.com/python/mypy/pull/6630, to match old semantic analyzer (and also Python runtime). Setting priority to high since this causes problems with internal deployment of new analyzer. @JukkaL I assign this to you since this may be related to tracking the order of definitions.

JukkaL:
Previously we added each submodule implicitly to the the symbol table of
the parent package. This PR removes this and instead we look up names from
the modules dictionary if they aren't found in the symbol table. The change
only affects the new semantic analyzer.

This provides the foundation that should make it easier to fix #6828. It also 
arguably cleans up the code.

Also refactor some related code to avoid duplication.

This isn't a pure refactor since some error messages are slightly different.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3803,6 +3803,10 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         """
         module = node.fullname()
         names = node.names
+        # Rebind potential references to old version of current module in
+        # fine-grained incremental mode.
+        if module == self.cur_mod_id:
+            names = self.globals
         part = parts[0]
         sym = names.get(part, None)
         if not sym:
@@ -3826,8 +3830,6 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 sym = SymbolTableNode(GDEF, v)
         elif sym.module_hidden:
             sym = None
-        else:
-            sym = self.rebind_symbol_table_node(sym)
         return sym
 
     def is_missing_module(self, module: str) -> bool:
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -99,9 +99,7 @@ from mypy.plugin import (
     Plugin, ClassDefContext, SemanticAnalyzerPluginInterface,
     DynamicClassDefContext
 )
-from mypy.util import (
-    get_prefix, correct_relative_import, unmangle, module_prefix
-)
+from mypy.util import correct_relative_import, unmangle, module_prefix
 from mypy.scope import Scope
 from mypy.newsemanal.semanal_shared import (
     SemanticAnalyzerInterface, set_callable_name, calculate_tuple_fallback, PRIORITY_FALLBACKS
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -3813,16 +3813,16 @@ class NewSemanticAnalyzer(NodeVisitor[None],
             submodule = module + '.' + part
             if submodule in self.modules:
                 sym = SymbolTableNode(GDEF, self.modules[submodule])
+            elif self.is_incomplete_namespace(module):
+                self.record_incomplete_ref()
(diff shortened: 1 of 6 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -3459,59 +3418,15 @@ def visit_member_expr(self, expr: MemberExpr) -> None:
...
-                        typ = AnyType(TypeOfAny.from_error)
-                else:
-                    typ = AnyType(TypeOfAny.from_error)
-                expr.kind = GDEF
-                expr.fullname = '{}.{}'.format(file.fullname(), expr.name)
-                expr.node = Var(expr.name, type=typ)
-            else:
-                if self.is_incomplete_namespace(file.fullname()):
-                    self.record_incomplete_ref()
+            # Handle 'module.foo'.
```
ilevkivskyi: Doesn't this comment just duplicates the one two lines above?

mypy/newsemanal/semanal.py
```diff
@@ -3881,27 +3797,42 @@ def lookup_type_node(self, expr: Expression) -> Optional[SymbolTableNode]:
...
+        """Look up a symbol from a module.
+
+        Return None if no matching symbol could be bound.
+        """
         module = node.fullname()
         names = node.names
         # Rebind potential references to old version of current module in
         # fine-grained incremental mode.
         if module == self.cur_mod_id:
             names = self.globals
```
ilevkivskyi: Since we are here again, could you please add a test that would fail without these two lines. Currently all tests pass if I remove them.

mypy/newsemanal/semanal.py
```diff
@@ -3881,27 +3797,42 @@ def lookup_type_node(self, expr: Expression) -> Optional[SymbolTableNode]:
...
+                    and (node.is_stub
+                         or self.options.python_version >= (3, 7))):
+                gvar = self.create_getattr_var(names['__getattr__'], name, fullname)
+                if gvar:
+                    sym = SymbolTableNode(GDEF, gvar)
+            elif self.is_missing_module(fullname):
+                var_type = AnyType(TypeOfAny.from_unimported_type)
+                v = Var(name, type=var_type)
+                v._fullname = fullname
+                sym = SymbolTableNode(GDEF, v)
```
ilevkivskyi: Maybe add a comment explaining why we create a new `Var` every time for technically the same missing module? (IIUC the logic is the same as explained in the docstring for `create_getattr_var()`).

mypy/newsemanal/semanal.py
JukkaL: I couldn't come up with a test that would require this. It's possible that fine-grained dependencies will deal with all the cases that this was supposed to handle. I've removed this now.

mypy/newsemanal/semanal.py
JukkaL: Removed one of the comments.

mypy/newsemanal/semanal.py
JukkaL: Added a comment.

PR #7013, opened 2019-06-18
Sync mypy with recent runtime updates in typing

ilevkivskyi:
This introduces the following updates:
* Allow using `typing.TypedDict` (still support all the old forms)
* Add a test for `typing.Literal` (it was already supported, but there were no tests)
* Rename `@runtime` to `@runtime_checkable`, while keeping the alias in `typing_extensions` for backwards compatibility.

(Note that `typing.Final` and `typing.Protocol` were already supported and there are tests.)
See also https://github.com/python/typeshed/pull/3070

### DIFF
--- a/docs/source/more_types.rst
+++ b/docs/source/more_types.rst
@@ -740,11 +740,6 @@ and one
 TypedDict
 *********
 
-.. note::
-
-   TypedDict is an officially supported feature, but it is still experimental.
-
-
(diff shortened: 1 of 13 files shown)


ilevkivskyi:
Actually, it looks like this depends on https://github.com/python/typeshed/pull/3070, so that should be merged first.

PR #7014, opened 2019-06-18
Sync typeshed

ilevkivskyi:
Needed for https://github.com/python/mypy/pull/7013


ilevkivskyi:
It is too late here, I will not wait for tests, I can continue tomorrow morning.

ilevkivskyi:
Hm, it looks like this should be made part of #7013. So closing this, and will update the latter.

PR #7017, opened 2019-06-18
New semantic analyzer: fix ambiguity between submodule and local def

### LINKED ISSUE #6828: Uncertainty between submodule and local definition in new semantic analyzer

JukkaL:
Now imports are biased to target the submodule of current package,
but elsewhere a local definition with the same name as a submodule 
takes precedence.  Even though this doesn't quite reflect what 
happens at runtime, this seems to cover the typical use cases well, 
and the implementation is simple.

Fixes #6828.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -1628,10 +1628,20 @@ class NewSemanticAnalyzer(NodeVisitor[None],
         import_id = self.correct_relative_import(imp)
         module = self.modules.get(import_id)
         for id, as_id in imp.names:
-            node = module.names.get(id) if module else None
+            possible_module_id = import_id + '.' + id
+            if module is None:
+                node = None
+            elif import_id == self.cur_mod_id and possible_module_id in self.modules:
+                # Submodule takes precedence over definition in surround package, for
+                # compatibility with runtime semantics in typical use cases. This
+                # could more precisely model runtime semantics by taking into account
+                # the line number beyond which the local definition should take
+                # precedence, but doesn't seem to be important in most use cases.
+                node = SymbolTableNode(GDEF, self.modules[possible_module_id])
+            else:
+                node = module.names.get(id)
 
             missing = False
-            possible_module_id = import_id + '.' + id
             imported_id = as_id or id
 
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -1628,10 +1628,20 @@ def visit_import_from(self, imp: ImportFrom) -> None:
...
+                node = None
+            elif import_id == self.cur_mod_id and possible_module_id in self.modules:
+                # Submodule takes precedence over definition in surround package, for
+                # compatibility with runtime semantics in typical use cases. This
+                # could more precisely model runtime semantics by taking into account
+                # the line number beyond which the local definition should take
+                # precedence, but doesn't seem to be important in most use cases.
+                node = SymbolTableNode(GDEF, self.modules[possible_module_id])
+            else:
+                node = module.names.get(id)
```
ilevkivskyi: Do we still need the special case couple dozen lines below (can't comment there) that checks for submodules added to parent module? (The one that has `# Special case: allow replacing submodules with variables.`)

It seems to be it should be safe to remove it now.

JukkaL: I'll look at it later and create a separate PR if it makes sense.

PR #7022, opened 2019-06-19
Expand environment variables in cache_dir config

hetmankp:
This further extends parsing of the 'cache_dir' configuration option (as originally introduced by #6651) to do environment variable expansion in addition to user home directory expansion. This makes it much easier to integrate mypy with other tools as well as making it possible to create more generic configuration files.


hetmankp:
Hi. Just pinging this thread to get some feedback as per the contributors documentation. Is it preferable to submit an issue linked to this pull request for the discussion?

PR #7023, opened 2019-06-19
Document --new-semantic-analyzer

JukkaL:
The "next mypy release" in the text refers to the next release after 
0.710.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -624,14 +624,14 @@ Miscellaneous
 
 ``--new-semantic-analyzer``
     This flag switches to an improved, experimental implementation of
-    the *semantic analyzer*, which mypy uses internally to bind Python
-    names to definitions. The old and the new semantic analyzers
-    mostly behave identically.  The new semantic analyzer is better at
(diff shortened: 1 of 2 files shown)


docs/source/command_line.rst
```diff
@@ -622,6 +622,19 @@ Miscellaneous
...
+    the *semantic analyzer*, which mypy uses internally to bind Python
+    names to definitions. The old and the new semantic analyzers
+    mostly behave identically.  The new semantic analyzer is better at
+    handling import cycles and forward references to definitions. It
+    also fixes inconsistencies between the daemon and non-daemon modes,
+    and it detects additional error conditions.
+
+    The next mypy release will use the new semantic analyzer by
+    default, and the old semantic analyzer will be removed in the next
+    release after that.
```
ilevkivskyi: Maybe use the actual expected release numbers instead of "next" and "next after that"?

gvanrossum: I'd rather not, since we may change our mind about that. However I would like to hedge a bit by adding "Likely, ...".

PR #7027, opened 2019-06-20
Use some more dogfooding

ilevkivskyi:
This PR does essentially two (related) things:
* `TYPE_CHECKING` is the official way for type-checking-only imports to avoid import cycles. This PR removes all uses of `if MYPY: ...` except for couple absolutely necessary files that are imported from `setup.py`.
* `typing_extensions` was designed in part as a drop-in replacement for `typing` on Python versions where some names are not yet available in `typing`.

I propose that mypy code adopts these itself. IMO this makes code look tidier (also it is 170 negative lines).

Are there any objections?

### DIFF
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -1,10 +1,8 @@
-from typing import Dict, List, Set, Iterator, Union, Optional, Tuple, cast
 from contextlib import contextmanager
 from collections import defaultdict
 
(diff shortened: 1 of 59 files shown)


ilevkivskyi:
> I presume this works with mypyc?

We now have mypyc in CI :-)

PR #7032, opened 2019-06-21
Fix MYPYC_BLACKLIST on Windows

### LINKED ISSUE #7031: [Regression] dmypy run broken on Windows with 0.710
This is a regression in 0.710
My environment:
Windows 10
mypy (from pypi) 0.710
Python 3.7.2

I finally got to testing the release on Windows and it seems that dmypy run is broken on Windows. This seems to be an issue with how mypyc is handling the dmypy directory.

A sample of the behavior:
```
$ dmypy run -- test.py --follow-imports=error
Timed out waiting for daemon to start.
```

The work-around for now is to install mypy without mypyc via `pip install mypy --no-binary mypy`.

Not satisfied with that, I decided to start removing binary extensions until things worked. I was able to further narrow things down to the offending file, which is (oddly) `dmypy/__main__.py`. I'm not at all sure why deleting the extension module corresponding to that file should fix things, but it does. The corresponding PR that introduced this change is #6851.

cc @msullivan

ethanhs:
This simple error has an unfortunately far reaching effect :(

The compiled `__main__.py` file doesn't work, so dmypy is broken with mypy 0.710 on Windows, which this fixes.

Closes #7031

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,7 +89,7 @@ if USE_MYPYC:
         # Need to be runnable as scripts
         '__main__.py',
         'sitepkgs.py',
-        'dmypy/__main__.py',
+        os.path.join('dmypy', '__main__.py'),
 
         # Needs to be interpreted to provide a hook to interpreted plugins
         'interpreted_plugin.py',


ilevkivskyi:
It is of course up to @gvanrossum but I would say it makes sense to make a point release.

JukkaL:
Yeah a point release seems worth it.

JelleZijlstra:
If we do do a point release, can I request that python/typeshed#3077 be included?

gvanrossum:
Yes, I agree we should also include python/typeshed#3077. (I think this would have to be done by branching typeshed to cherry-pick just that change, and then setting the mypy release branch to use that branch as its submodule.

Can this wait until Monday?

PR #7046, opened 2019-06-23
Use mypyc during appveyor tests

ethanhs:


ethanhs:
@msullivan the CI passes and things build fine, but there appear to be several warnings that may be worth looking at for the sake of mypyc.

msullivan:
This looks like a reasonable setup for a mypyc build, but maybe we should also still have non-mypyc builds?

ethanhs:
@msullivan sure, I think we can run two concurrent jobs, so it won't take any longer

ethanhs:
@msullivan done! Though it looks like we only get a single job, so it takes twice as long :/

JukkaL:
I'm not sure this is worth that much slowdown. Would it be possible to only run both builds after a commit has landed? I.e., in PRs we'd only run one of the builds. The two builds will likely agree with each other (or one of the non-Windows builds) 98% of the time or so.

ethanhs:
It appears I can have it run uncompiled mypy on PRs, then run the mypyc version on master, if that works?

JukkaL:
Yeah, that would work for me. I guess we can run the version that is faster on PRs.

msullivan:
Sure, that seems like a reasonable strategy for now.

The *right* solution is probably to run the tests as part of `mypy_mypyc-wheels`, but currently the mechanism for testing a built wheel of mypy is a really unfortunate shell script: https://github.com/python/mypy/blob/master/misc/test_installed_version.sh (I am sure there must be a way that is not that bad, possibly involving some modification of mypy's test organization, but I timed out trying to figure it out and went with brute force instead.)

ethanhs:
@msullivan perhaps then it would be better if I were to write an equivalent script for Windows?

msullivan:
If you want to do that, that would be the best approach probably.

msullivan:
Are you planning to write a testing script for windows or should we move forward with merging some variant of this?

ethanhs:
Ah sorry I got really busy and forgot about this. I probably won't have time to work on it until mid August. Perhaps we can talk at PyBay about it?

ethanhs:
FWIW this is the follow up https://github.com/python/mypy/pull/7488

Issue #2395, opened 2016-11-03
Warn about always true/false isinstance tests


JukkaL:
These may result in code that mypy considers unreachable and silently doesn't type check. However, they may happen quite frequently in innocent things like `assert isinstance(x, str)` so finding a good heuristic for what to report can be tricky.

rwbarton:
Initial reactions:

* Don't warn about an assertion that always passes

* Do warn about an assertion that always fails

* Do warn about a branch of an `if` that is unreachable

ddfisher:
To add to Reid's comment: but don't warn about an assertion that always fails if it's an explicit `assert False`.

kirbyfan64:
@ddfisher Or any false-y builtin constant, since I've seen `assert 0` and `assert None` before.

gvanrossum:
@ilinum This issue might be interesting given your recent quest for flagging unreachable code.

ilevkivskyi:
I have found five other issues on the tracker that are some variations of this, so I think this is high priority.

ilevkivskyi:
Also from my experience there is definitely no need in a warning about an always True asserts. So 100% with @rwbarton [above](https://github.com/python/mypy/issues/2395#issuecomment-258235424).

Issue #7008, opened 2019-06-17
Detect impossible condition caused by user defined __bool__


graingert:
Eg with the following:

```python
from typing import *

class Foo:
    def __bool__(self) -> Literal[True]:
        return True

def some_foo() -> Foo:
    ...

def throw_error():
    raise Exception


if __name__ == '__main__':
    print(some_foo() or throw_error())
    #                   ^^^^^^^^^^^^^  impossible condition, unreachable
```

mypy should be able to detect that the call to throw_error is unreachable

ilevkivskyi:
Well mypy doesn't detect even simpler unreachable function calls like `True or print()`. It may be a useful feature to warn about unreachable code.

Related to https://github.com/python/mypy/issues/2395

ilevkivskyi:
https://github.com/python/mypy/issues/6990 is another example where this may be useful.

Michael0x2a:
FWIW I've started sort of looking into adding in some kind of flag for this during the past week (I'm already poking around this corner of the codebase for literal and enum reachability stuff anyways).

I think in principle it should be straightforward: to avoid false-positives with things like `sys.platform` checks, we don't alert on blocks that were marked as unreachable as a part of semantic analysis and only look at results from the type-checking phase.

We'll also probably want to adjust mypy so it distinguishes between intentionally and unintentionally unreachable blocks. E.g. if somebody does `if False: ...` that was probably intentional so we'll probably need to do something different there. Probably we'll need to refactor `find_isinstance_check` a bit to make this all work cleanly, but I was maybe thinking of doing that anyways.

TBH I think the hardest/most annoying part will be constructing an understandable error message for subexpressions that were skipped due to being unreachable, just due to the "line-by-line" nature of error reporting in mypy.

Possibly adding in a "fyi this expression is always True/always False" error message might be sufficient instead, as opposed to indicating specific ignored subexpressions. Not sure, still need to do more poking around.

ilevkivskyi:
> TBH I think the hardest/most annoying part will be constructing an understandable error message for subexpressions that were skipped due to being unreachable, just due to the "line-by-line" nature of error reporting in mypy.

On Python 3.8 there is a simple solution for that: `ast.get_source_segment()` gives exact source code for every AST node. But maybe a more generic error message would be sufficient.

> E.g. if somebody does `if False: ...` that was probably intentional

[...]

Issue #5818, opened 2018-10-22
Preferred idiom for exhaustiveness checking of unions


bluetech:
* Are you reporting a bug, or opening a feature request?

A question...

* Please insert below the code you are checking with mypy.

I have some union, e.g.

```python
from typing import Union

MyUnion = Union[int, str]
```

and I have a function which takes the union and matches on it:

```python
def handle_my_union(x: MyUnion) -> None:
    if isinstance(x, int): ...
    elif isinstance(x, str): ...
```

Now suppose that the union gets added a new member:

```diff
- MyUnion = Union[int, str]
+ MyUnion = Union[int, str, bytes]
```

I have to remember to add it to all of the places which match on the union, for example `handle_my_union`. As written above, mypy doesn't warn (and has no reason to). So what I use is this function (stolen mostly verbatim [from TypeScript](https://www.typescriptlang.org/docs/handbook/advanced-types.html#exhaustiveness-checking)):

```python
from typing import NoReturn

def assert_never(x: NoReturn) -> NoReturn:
    assert False, "Unhandled type: {}".format(type(x).__name__)
```

and use it as follows:

```diff
 def handle_my_union(x: MyUnion) -> None:
     if isinstance(x, int): ...
     elif isinstance(x, str): ...
+    else:
+        assert_never(x)
```

* What is the actual behavior/output?

Now mypy warns:

```
error: Argument 1 to "assert_never" has incompatible type "bytes"; expected "NoReturn"
```

When I add a branch to handle `bytes`:

```diff
 def handle_my_union(x: MyUnion) -> None:
     if isinstance(x, int): ...
     elif isinstance(x, str): ...
+    elif isinstance(x, bytes): ...
     else:
         assert_never(x)
```

the error is gone.

* What is the behavior/output you expect?

I am happy with this solution, I just wonder if this is something that is considered appropriate to rely on, or if there's another preferable way to achieve exhaustiveness checking?

* What are the versions of mypy and Python you are using?

Python 3.5.5, mypy 0.630.

* What are the mypy flags you are using? (For example --strict-optional)

`no_implicit_optional, check_untyped_defs, warn_unused_ignores`.

srittau:
Typescript also has a `never` type that is not compatible to anything, not even itself. I wonder whether it makes sense to have something like this in `typing`, which could be used in place of the argument type here.

bluetech:
Isn't `NoReturn` equivalent to `never`? I'm just concluding by analogy here, specifically

[...]

PR #7050, opened 2019-06-24
Add flag to warn about unreachable branches and exprs

Michael0x2a:
This diff adds a `--warn-unreachable` flag that reports an error if:

1. Any statement is inferred to be unreachable
2. Any subexpression in boolean expressions, inline if statements, and list/set/generator/dict comprehensions are always unreachable or redundant.

This only takes into account the results of *type* analysis. We do not report errors for statements that are inferred to be unreachable in the semantic analysis phase (e.g. due to `sys.platform` checks and the like).

Those checks are all intentional/deliberately flag certain branches as unreachable, so error messages would be annoying in those cases.

A few additional notes:

1. I didn't spend a huge amount of time trying to come up with error messages. They could probably do with some polishing/rewording.

2. I thought about enabling this flag for default with mypy, but unfortunately that ended up producing ~40 to 50 (apparently legitimate) error messages.

   About a fourth of them were due to runtime checks checking to see if some type is None (despite not being declared to be Optional), about a fourth seemed to be due to mypy not quite understanding how to handle things like traits and Bogus[...], and about a fourth seemed to be legitimately unnecessary checks we could safely remove.

   The final fourth was a mixture of typeshed bugs and misc errors with the reachability checks. (e.g. see https://github.com/python/mypy/pull/7048)

3. For these reasons, I decided against adding this flag to `--strict` and against documenting it yet. We'll probably need to flush out a few bugs in mypy first/get mypy to the state where we can at least honestly dogfood this.

Resolves https://github.com/python/mypy/issues/2395; partially addresses https://github.com/python/mypy/issues/7008.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1777,12 +1777,35 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
 
     def visit_block(self, b: Block) -> None:
         if b.is_unreachable:
+            # This block was marked as being unreachable during semantic analysis.
+            # It turns out any blocks marked in this way are *intentionally* marked
+            # as unreachable -- so we don't display an error.
             self.binder.unreachable()
             return
         for s in b.body:
             if self.binder.is_unreachable():
+                if self.options.disallow_inferred_unreachable and not self.is_raising_or_empty(s):
+                    self.msg.unreachable_branch(s)
                 break
             self.accept(s)
+
+    def is_raising_or_empty(self, s: Statement) -> bool:
+        if isinstance(s, AssertStmt) and is_false_literal(s.expr):
+            return True
+        elif isinstance(s, (RaiseStmt, PassStmt)):
+            return True
+        elif isinstance(s, ExpressionStmt):
+            if isinstance(s.expr, EllipsisExpr):
+                return True
+            elif isinstance(s.expr, CallExpr):
+                self.expr_checker.msg.disable_errors()
+                typ = self.expr_checker.accept(
+                    s.expr, allow_none_return=True, always_allow_any=True)
+                self.expr_checker.msg.enable_errors()
+
+                if isinstance(typ, UninhabitedType):
+                    return True
+        return False
 
     def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
         """Type check an assignment statement.
@@ -2956,7 +2979,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             # type checks in both contexts, but only the resulting types
             # from the latter context affect the type state in the code
             # that follows the try statement.)
-            self.accept(s.finally_body)
+            if not self.binder.is_unreachable():
+                self.accept(s.finally_body)
 
     def visit_try_without_finally(self, s: TryStmt, try_frame: bool) -> None:
         """Type check a try statement, ignoring the finally block.
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -783,7 +783,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
     def check_func_def(self, defn: FuncItem, typ: CallableType, name: Optional[str]) -> None:
         """Type check a function definition."""
         # Expand type variables with value restrictions to ordinary types.
-        for item, typ in self.expand_typevars(defn, typ):
+        expanded = self.expand_typevars(defn, typ)
+        for item, typ in expanded:
             old_binder = self.binder
             self.binder = ConditionalTypeBinder()
             with self.binder.top_frame_context():
@@ -932,6 +933,14 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             # Type check body in a new scope.
             with self.binder.top_frame_context():
                 with self.scope.push_function(defn):
+                    # We suppress reachability warnings when we use TypeVars with value
+                    # restrictions: we only want to report a warning if a certain statement is
+                    # marked as being suppressed in *all* of the expansions, but we currently
+                    # have no good way of doing this.
+                    #
+                    # TODO: Find a way of working around this limitation
+                    if len(expanded) >= 2:
+                        self.binder.suppress_unreachable_warnings()
                     self.accept(item.body)
                 unreachable = self.binder.is_unreachable()
 
@@ -1784,12 +1793,22 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             return
         for s in b.body:
             if self.binder.is_unreachable():
-                if self.options.warn_unreachable and not self.is_raising_or_empty(s):
-                    self.msg.unreachable_branch(s)
+                if (self.options.warn_unreachable
+                        and not self.binder.is_unreachable_warning_suppressed()
+                        and not self.is_raising_or_empty(s)):
+                    self.msg.unreachable_statement(s)
                 break
             self.accept(s)
 
     def is_raising_or_empty(self, s: Statement) -> bool:
+        """Returns 'true' if the given statement either throws an error of some kind
+        or is a no-op.
+
+        We use this function mostly while handling the '--warn-unreachable' flag. When
+        that flag is present, we normally report an error on any unreachable statement.
+        But if that statement is just something like a 'pass' or a just-in-case 'assert False',
+        reporting an error would be annoying.
+        """
         if isinstance(s, AssertStmt) and is_false_literal(s.expr):
             return True
         elif isinstance(s, (RaiseStmt, PassStmt)):
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1784,7 +1784,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             return
         for s in b.body:
             if self.binder.is_unreachable():
-                if self.options.disallow_inferred_unreachable and not self.is_raising_or_empty(s):
+                if self.options.warn_unreachable and not self.is_raising_or_empty(s):
                     self.msg.unreachable_branch(s)
                 break
             self.accept(s)
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -492,6 +492,10 @@ def process_options(args: List[str],
                         help="Treat imports as private unless aliased",
                         group=strictness_group)
 
+    add_invertible_flag('--disallow-inferred-unreachable', default=False, strict_flag=False,
+                        help="Disallow branches inferred to be unreachable after type analysis",
+                        group=strictness_group)
+
     incremental_group = parser.add_argument_group(
         title='Incremental mode',
         description="Adjust how mypy incrementally type checks and caches modules. "
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -464,7 +464,8 @@ def process_options(args: List[str],
                              " from non-Any typed functions",
                         group=lint_group)
     add_invertible_flag('--warn-unreachable', default=False, strict_flag=False,
-                        help="Disallow branches inferred to be unreachable after type analysis",
+                        help="Warn about statements or expressions inferred to be"
+                             " unreachable or redundant",
                         group=lint_group)
 
     # Note: this group is intentionally added here even though we don't add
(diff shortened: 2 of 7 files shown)


mypy/main.py
```diff
@@ -492,6 +492,10 @@ def add_invertible_flag(flag: str,
                         help="Treat imports as private unless aliased",
                         group=strictness_group)
 
+    add_invertible_flag('--disallow-inferred-unreachable', default=False, strict_flag=False,
```
JukkaL: Maybe we can make the flag name a bit clearer. What about `--warn-unreachable`? We still allow some unreachable code, but I think that we don't need to make this distinction in the flag name. I think that it's sufficient to describe the semantics in the documentation.

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
...
+    reveal_type(x)
+[builtins fixtures/exception.pyi]
+
+[case testUnreachableFlagExpressions]
+# flags: --disallow-inferred-unreachable
+def foo() -> bool: ...
+
+lst = [1, 2, 3, 4]
+
+a = True or foo()                        # E: The right operand of this 'or' expression is never evaluated
```
JukkaL: Idea for message: `Right operand of 'or' is never evaluated`

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
...
+b = False or foo()                       # E: The left operand of this 'or' expression is always false
```
JukkaL: Similar to above, maybe rephrase as `Left operand of 'or' is always false`

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
...
+c = True and foo()                       # E: The left operand of this 'and' expression is always true
+d = False and foo()                      # E: The right operand of this 'and' expression is never evaluated
+e = True or (True or (True or foo()))    # E: The right operand of this 'or' expression is never evaluated
+f = (True or foo()) or (True or foo())   # E: The right operand of this 'or' expression is never evaluated
+g = 3 if True else 4                     # E: The 'else' expression in this inline if expression is never evaluated
```
JukkaL: Idea for message: `If condition is always true`

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
...
+h = 3 if False else 4                    # E: The 'if' expression in this inline if expression is never evaluated
```
JukkaL: Idea for message: `If condition is always false`

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
...
+i = [x for x in lst if True]             # E: The conditional check in this comprehension is always true
```
JukkaL: Idea for message: `If condition in comprehension is always true`

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
...
+if False:
+    assert False
+    reveal_type(x)
+
+if False:
+    raise Exception()
+    reveal_type(x)
+
+if False:
+    assert_never(x)
```
JukkaL: Random thought (no need to address in this PR): What do you think about providing a magic utility function such as `unreachable()` for easily silencing errors about unreachable code?

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
...
+[builtins fixtures/exception.pyi]
+
+[case testUnreachableFlagIgnoresSemanticAnalysisUnreachable]
+# flags: --disallow-inferred-unreachable --python-version 3.7 --platform win32 --always-false FOOBAR
+import sys
+from typing import TYPE_CHECKING
+
+x: int
+if TYPE_CHECKING:
+    reveal_type(x)  # N: Revealed type is 'builtins.int'
```
JukkaL: Also test that any code is allowed in an else block?

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
...
+
+if sys.platform == 'darwin':
+    reveal_type(x)
```
JukkaL: Similar to above, test that any code is allowed in an else block if the platform check is true?

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
...
+
+if sys.version_info == (2, 7):
+    reveal_type(x)
```
JukkaL: Similar to above, test that any code is allowed in an else block if the version info check is true?

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
...
 reveal_type('')  # N: Revealed type is 'builtins.str'
 [builtins fixtures/ops.pyi]
+
+[case testUnreachableFlagWithBadControlFlow]
+# flags: --disallow-inferred-unreachable
+a: int
+if isinstance(a, int):
+    reveal_type(a)  # N: Revealed type is 'builtins.int'
+else:
+    reveal_type(a)  # E: This branch is inferred to be unreachable
```
JukkaL: Idea for message: `Statement is unreachable`

Even better, maybe we can somehow detect that this is caused by an always-true `isinstance` check and generate an error on the line with the `isinstance` check.

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
     reveal_type('')  # N: Revealed type is 'builtins.str'
 reveal_type('')  # N: Revealed type is 'builtins.str'
 [builtins fixtures/ops.pyi]
+
+[case testUnreachableFlagWithBadControlFlow]
+# flags: --disallow-inferred-unreachable
+a: int
+if isinstance(a, int):
```
JukkaL: Add test with a function that accepts a type variable with value restriction and performs an `isinstance` check that is always true/false for some value of the type variable. It may be reasonable to disable unreachability checks in a function with a type variable restriction (at least for now), since the correct implementation may be tricky and such type variables are relatively rare.

mypy/checker.py
```diff
@@ -1777,13 +1777,36 @@ def check_import(self, node: ImportBase) -> None:
...
             self.binder.unreachable()
             return
         for s in b.body:
             if self.binder.is_unreachable():
+                if self.options.disallow_inferred_unreachable and not self.is_raising_or_empty(s):
+                    self.msg.unreachable_branch(s)
                 break
             self.accept(s)
 
+    def is_raising_or_empty(self, s: Statement) -> bool:
```
JukkaL: Add docstring which describes the purpose of this function.

mypy/messages.py
```diff
@@ -1228,6 +1228,30 @@ def note_call(self, subtype: Type, call: Type, context: Context) -> None:
         self.note('"{}.__call__" has type {}'.format(self.format_bare(subtype),
                                                      self.format(call, verbosity=1)), context)
 
+    def unreachable_branch(self, context: Context) -> None:
```
JukkaL: What do you think about renaming this to `unreachable_statement`? Would it be clearer (and correct)?

mypy/messages.py
```diff
@@ -1228,6 +1228,30 @@ def note_call(self, subtype: Type, call: Type, context: Context) -> None:
...
+        self.fail("This branch is inferred to be unreachable", context)
+
+    def always_same_truth_value_left_operand(self, op_name: str, context: Context) -> None:
```
JukkaL: Document supported op names.

test-data/unit/check-unreachable-code.test
```diff
@@ -715,3 +715,179 @@ if sys.version_info[0] >= 2:
     reveal_type('')  # N: Revealed type is 'builtins.str'
 reveal_type('')  # N: Revealed type is 'builtins.str'
 [builtins fixtures/ops.pyi]
+
+[case testUnreachableFlagWithBadControlFlow]
```
JukkaL: What about something like `DEBUG: Final = False` followed by `if DEBUG: print('foo')`? Is there a test for this use case?

Michael0x2a: Our reachability checks currently don't special-case Literal bools (or Finals that are bools) at all -- we'd actually just end up checking both branches in the `if DEBUG: ... else: ...` block.

I'm not sure exactly what behavior we want though, which is why I refrained from adding a test case.

(On one hand, it'd be nice to make mypy smarter about reachability in general; on the other improving mypy's understanding would lead to actual regressions in type inference if users don't opt-in to using this new flag.)

mypy/messages.py
Michael0x2a: Good idea! I think it'd be more correct, especially since we're also reporting these errors after things like asserts and returns. (I also added a test case for the latter).

test-data/unit/check-unreachable-code.test
Michael0x2a: I experimented a bit with this idea. I think switching to reporting an error on the check might be a bit messy -- it'd actually be much easier to report *both* error message.

Do you think that's ok? If so, I can work on submitting a separate PR with this change after this one is landed. (I'll probably need to do some mild refactoring so we can reuse the "should we report a warning with this line?" logic, and I'd prefer to do that in a separate branch.)

test-data/unit/check-unreachable-code.test
Michael0x2a: That could be handy, though possibly annoying to use for people with linters (since they don't understand mypy magic functions). We could work around this by adding the function to `mypy_extensions`? Or maybe that's overkill?

Another idea: once the "tag all error messages with a code" PR is landed, the user could maybe add a `# type: ignore` for just the "Statement is unreachable" warning?

Michael0x2a:
@JukkaL -- Thanks for the CR! I updated the PR; this should be ready for a second look.

test-data/unit/check-unreachable-code.test
Michael0x2a: Or another idea -- what if we instead modified mypy so that we *mandate* that any calls to `unreachable()` are actually unreachable?

We could also modify this function so that optionally take in any number of arguments: if the checker runs into this function, it'll run `reveal_type(...)` on the provided args before reporting an error that the statement was actually reachable. This would also let us supplant this "assert never" pattern (and providing a runtime implementation of the same logic in `mypy_extensions` or something would feel less like overkill).

bluetech:
I tried this PR on some of our code. It found a bunch of useless `is None` checks (and a [typeshed bug](https://github.com/python/typeshed/pull/3093)), which is nice.

I was interested to check whether this would "break" the `assert_never` trick from #5818, which is like the opposite of this feature ("warn if this is *not* unreachable"), but the warning is not triggered in that case, which is good. I used this:

```py
from typing import Union, NoReturn


def assert_never(value: NoReturn) -> NoReturn:
    assert False, f'Unhandled value: {value} ({type(value).__name__})'


def f(x: Union[int, str]) -> None:
    if isinstance(x, int):
        pass
    elif isinstance(x, str):
        pass
    else:
        # Triggers "Statement is unreachable".
        # print('xx')
        # Doesn't trigger.
        assert_never(x)
```

test-data/unit/check-unreachable-code.test
JukkaL: Yeah, that would be how I imagined it to work. Not sure about the arguments though. My idea would be to do something like this:

```py
X = Union[A, B]

def f(x: X) -> None:
    if isinstance(x, A):
        ...
    elif isinstance(x, B):
        ...
    else:
        unreachable()  # Error only if we forgot to handle some case
```

Maybe we could automatically recognize that pattern and print out the inferred type of `x` if we hit an error.

test-data/unit/check-unreachable-code.test
JukkaL: The current behavior seems reasonable to me.

ilevkivskyi:
Nice feature! Would be great to add docs for this.

PR #7051, opened 2019-06-24
[wip] Basic PoC of Expand type for TypedDict-kwargs

untitaker:
This implements `Expand` in kwarg positions as per #4441. TODO:

- [ ] Forbid usage of `Expand` outside of `*args` and `**kwargs` (what does `x: Expand[...] = 42` even mean?)
- [ ] Implement `Expand` properly s.t one can import it at runtime
- [ ] Commit the changes to mypy_extensions
- [ ] tests

Right now this PR only exists as a discussion basis for the approach I should take when implementing this properly. Adding `expand_formal_type` seems to be the best way to me right now but I don't have good oversight over the entire codebase to know any better.


mypy/argmap.py
```diff
@@ -193,3 +193,21 @@ def expand_actual_type(self,
         else:
             # No translation for other kinds -- 1:1 mapping.
             return actual_type
+
+
+def expand_formal_type(formal_type: Type, actual_name: Optional[str], tuple_idx: int) -> Type:
+    if isinstance(formal_type, Instance) and formal_type.type.fullname() == 'mypy_extensions.Expand':
```
untitaker: This is ugly. How do I get something like `TypedDictType` (`ExpandType`) and make mypy map `mypy_extensions.Expand` to `mypy.types.ExpandType`?

ilevkivskyi:
Thanks for PR!

I think it should be done in a different way: it will be simpler if most of things happen during semantic analysis rather than type checking. Note that we currently have two semantic analyzers (things that construct symbol tables and bind reference expressions to symbols). The old one will be deprecated soon, so I propose to only support this with the new one (located in `mypy/newsemanal`). Here are some comments:
* `Expand` should be prohibited everywhere except in `**kwargs`.
* This should not be a proper type but rather a name qualifier (like `Final` or `ClassVar`)
* Type analyzer should modify function in place during semantic analysis using the expanded version (see `mypy.newsemanal.typeanal.TypeAnalyzer.visit_callable_type`)
* At the same time, the analyzer should insert a `kwargs` variable (having the actual explicit `TypedDict` type) in the local symbol table, see [around here](https://github.com/python/mypy/blob/master/mypy/newsemanal/semanal.py#L864).

Let me know if you have more questions.

untitaker:
I think I don't have the motivation to work on this anymore, sorry.

PR #7054, opened 2019-06-24
Support flexible option specification in config file and inline configuration

msullivan:
Support inverting options based on `no_`, and adding or removing a `dis` prefix
in the config file (and therefore in inline configuration also).

This moves the handling of `no_` prefixes from inline configuration to the config file.
As a result, the config file is brought closer into alignment with the command line options.

### DIFF
--- a/mypy/config_parser.py
+++ b/mypy/config_parser.py
@@ -209,7 +209,8 @@ def parse_section(prefix: str, template: Options,
                     v = not v
             elif callable(ct):
                 if invert:
-                    print("%sCan not invert non-boolean key %s" % (prefix, options_key), file=stderr)
+                    print("%sCan not invert non-boolean key %s" % (prefix, options_key),
+                          file=stderr)
                     continue
                 try:
                     v = ct(section.get(key))
--- a/mypy/config_parser.py
+++ b/mypy/config_parser.py
@@ -163,6 +163,8 @@ def parse_section(prefix: str, template: Options,
     results = {}  # type: Dict[str, object]
     report_dirs = {}  # type: Dict[str, str]
     for key in section:
+        invert = False
+        options_key = key
         if key in config_types:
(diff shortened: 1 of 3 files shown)


mypy/config_parser.py
```diff
@@ -177,21 +179,38 @@ def parse_section(prefix: str, template: Options,
...
             ct = type(dv)
         v = None  # type: Any
         try:
             if ct is bool:
                 v = section.getboolean(key)  # type: ignore  # Until better stub
+                if invert:
+                    v = not v
             elif callable(ct):
+                if invert:
+                    print("%sCan not invert non-boolean key %s" % (prefix, options_key), file=stderr)
```
ilevkivskyi: Should this kind of errors be `sys.exit(2)`?

msullivan: 1. Probably
2. Similar errors are not

ilevkivskyi: @gvanrossum What do you think?

gvanrossum: When the mypy.ini file was new, I decided to make errors in the config file just print warnings. I can't recall why, possibly because the control flow to make them real errors was complex, or possibly because I was worried that people would have junk in their config files.

I'd stick to this tradition for now and do a separate sweep to clean it up later. (Maybe sections we don't recognize should not be fatal but in recognized sections, flags we don't understand should be fatal?)

msullivan:
> This will invalidate some people's mypy.ini files, right? So it should be messaged carefully with the next release.

I don't think it really will? Certain configurations that currently produce a warning will instead have meaning, but I don't think that counts.

mypy/config_parser.py
ilevkivskyi: > I'd stick to this tradition for now and do a separate sweep to clean it up later.

SGTM.

PR #7063, opened 2019-06-25
New semantic analyzer: fix false positives from TypedDict in stub

### LINKED ISSUE #7028: Using TypedDict in stub files with new analyzer produces error
mypy version: 0.710
Python version: 3.7.7

Given the following stub file (test.pyi):

```python
from typing_extensions import TypedDict

class StuffDict(TypedDict):
    foo: str
    bar: int

def thing(stuff: StuffDict) -> int: ...
```

Running `mypy --new-semantic-analyzer test.pyi` produces the following output:

```
test.pyi: error: Class test.StuffDict has abstract attributes "__getitem__", "__iter__", "__len__"
test.pyi: note: If it is meant to be abstract, add 'abc.ABCMeta' as an explicit metaclass
```

JukkaL:
Fixes #7028.

### DIFF
--- a/mypy/newsemanal/semanal_classprop.py
+++ b/mypy/newsemanal/semanal_classprop.py
@@ -51,6 +51,8 @@ def calculate_class_abstract_status(typ: TypeInfo, is_stub_file: bool, errors: E
     abstract attribute.  Also compute a list of abstract attributes.
     Report error is required ABCMeta metaclass is missing.
     """
+    if typ.typeddict_type:
+        return  # TypedDict can't be abstract
     concrete = set()  # type: Set[str]
     abstract = []  # type: List[str]
     abstract_in_this_class = []  # type: List[str]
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal_classprop.py
```diff
@@ -95,7 +95,7 @@ def calculate_class_abstract_status(typ: TypeInfo, is_stub_file: bool, errors: E
     if is_stub_file:
         if typ.declared_metaclass and typ.declared_metaclass.type.fullname() == 'abc.ABCMeta':
             return
-        if typ.is_protocol:
+        if typ.is_protocol or typ.typeddict_type:
```
ilevkivskyi: I am curious why this is not needed in the old analyzer. Is it an old analyzer bug?

JukkaL: I think that the old semantic analyzer just doesn't run this code at all for TypedDicts.

JukkaL: I updated the code to be closer to the logic in the old semantic analyzer.

PR #7066, opened 2019-06-25
Fix slicing tuples with nonliterals

### LINKED ISSUE #7056: Tuple slice by variable results in incorrect type
```python
from typing import Tuple

sz = 2
x: Tuple[int, ...] = (1, 2)[:sz]
```

```console
$ mypy --version
mypy 0.711
$ mypy t.py
mt.py:4: error: Incompatible types in assignment (expression has type "int", variable has type "Tuple[int, ...]")
```

ethanhs:
This fixes #7056.

The idea is that we don't know what types a tuple slice will result in, other than a tuple of some sort.

For example, `(1,'', 3.3)[:a]` could result in `Tuple[()]`, `Tuple[int]`, `Tuple[int, str]`, or `Tuple[int, str, float]`.

I also fixed a typo I noticed in TypeOfAny.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2520,7 +2520,13 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                                       'actual type', 'expected type'):
             return AnyType(TypeOfAny.from_error)
         else:
-            return UnionType.make_simplified_union(left_type.items)
+            if isinstance(index, SliceExpr):
+                # if we index (1,2,3)[:name], we don't know what the return type is,
+                # beyond that it is a tuple of some sort.
+                anys = (AnyType(TypeOfAny.implementation_artifact),)
+                return self.chk.named_generic_type('builtins.tuple', anys)
+            else:
+                return UnionType.make_simplified_union(left_type.items)
 
     def _get_value(self, index: Expression) -> Optional[int]:
         if isinstance(index, IntExpr):
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2520,13 +2520,18 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                                       'actual type', 'expected type'):
             return AnyType(TypeOfAny.from_error)
         else:
+            union = UnionType.make_simplified_union(left_type.items)
+            # If we are indexing, we return the union of elements. In a slice,
+            # we return Tuple[Any, ...], unless it is homogeneous, where we return
+            # Tuple[T, ...]
             if isinstance(index, SliceExpr):
-                # if we index (1,2,3)[:name], we don't know what the return type is,
-                # beyond that it is a tuple of some sort.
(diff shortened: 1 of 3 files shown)


mypy/checkexpr.py
```diff
@@ -2520,7 +2520,13 @@ def nonliteral_tuple_index_helper(self, left_type: TupleType, index: Expression)
                                       'actual type', 'expected type'):
             return AnyType(TypeOfAny.from_error)
         else:
-            return UnionType.make_simplified_union(left_type.items)
+            if isinstance(index, SliceExpr):
+                # if we index (1,2,3)[:name], we don't know what the return type is,
+                # beyond that it is a tuple of some sort.
+                anys = (AnyType(TypeOfAny.implementation_artifact),)
```
ethanhs: I wasn't sure which variant to use here for TypeOfAny, so guidance would be appreciated.

asottile: hmmm, I was thinking this would end up being `Tuple[Union[the tuple types], ...]` -- that way for homogenous tuples you'd end up with a consistent type

asottile: or maybe only that specialization for homogenous tuples? 🤔

ethanhs: Yeah, I was thinking that too, but then I thought about code like:
```
a = 2
s = (1, 3.3, '')[:a]
s[0] + s[1]
```
Though perhaps I am being too cautious.

ethanhs: I could also special case homogeneous tuples to return `Tuple[T, ...]`, and for everything else fall back to `Tuple[Any, ...]`. Though for a decision like this I will defer to @JukkaL or someone else who wants to chime in here.

asottile: that sounds fine to me 👍

mypy/checkexpr.py
```diff
@@ -2520,7 +2520,18 @@ def nonliteral_tuple_index_helper(self, left_type: TupleType, index: Expression)
...
-            return UnionType.make_simplified_union(left_type.items)
+            union = UnionType.make_simplified_union(left_type.items)
+            # If we are indexing, we return the union of elements. In a slice,
+            # we return Tuple[Any, ...], unless it is homogeneous, where we return
+            # Tuple[T, ...]
+            if isinstance(index, SliceExpr):
+                if union == left_type.items[0]:
+                    t = union
+                else:
+                    t = AnyType(TypeOfAny.implementation_artifact)
```
JukkaL: I think that it would be better to produce `Tuple[Union[...], ...]` instead of `Tuple[Any, ...]`. The rationale is that the bug has been around for quite a long time so the rate of false positives is expected to low. On the other hand, producing `Any` can produce imprecise typing in an otherwise fully typed program.

ethanhs: Ok, then I will do that.

PR #7072, opened 2019-06-26
Add more tests for column numbers in error messages

JukkaL:
Added tests for many of the most common error messages. These should
cover maybe 95% of error messages (based on user logs).

This only updates tests and doesn't change any behavior.

### DIFF
--- a/test-data/unit/check-columns.test
+++ b/test-data/unit/check-columns.test
@@ -115,6 +115,7 @@ def f(x, y): pass
 
 [case testColumnListOrDictItemHasIncompatibleType]
 from typing import List, Dict
+# TODO: Point to the actual item since a list/dict literal can span many lines
 x: List[int] = [  # E:16: List item 0 has incompatible type "str"; expected "int"
     'x']
 y: Dict[int, int] = {  # E:21: Dict entry 0 has incompatible type "str": "int"; expected "int": "int"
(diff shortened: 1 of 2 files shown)


test-data/unit/check-columns.test
```diff
@@ -65,4 +45,175 @@ def f(x: object, n: int, s: str) -> None:
...
+
+[case testColumnValueOfTypeVariableCannotBe]
+from typing import TypeVar, Generic
+
+T = TypeVar('T', int, str)
+
+class C(Generic[T]):
+    pass
+
+# TOOD: Column number missing
```
ilevkivskyi:
```suggestion
# TODO: Column number missing
```

test-data/unit/check-columns.test
```diff
@@ -65,4 +45,175 @@ def f(x: object, n: int, s: str) -> None:
...
+
+[case testColumnInvalidArguments]
+def f(x, y): pass
+(f()) # E:2: Too few arguments for "f"
+(f(y=1)) # E:2: Missing positional argument "x" in call to "f"
+
+[case testColumnListOrDictItemHasIncompatibleType]
+from typing import List, Dict
+x: List[int] = [  # E:16: List item 0 has incompatible type "str"; expected "int"
+    'x']
```
ilevkivskyi: This may break some `# type: ignore`s, but I would set the context to the item expression, not the whole list expression.

Add a TODO item also about this?

JukkaL:
Travis integration seems to be having issues; ignoring the pending build.

PR #7074, opened 2019-06-26
Write docs for inline configuration

msullivan:


docs/source/inline_config.rst
```diff
@@ -0,0 +1,40 @@
...
+Multiple flags can be separated by commas or placed on separate
+lines. To include a comma as part of an option's value, place the
+value inside quotes.
+
+Like in the configuration file, options that take a boolean value may be
+inverted by adding ``no-`` to their name or by (when applicable)
+swapping their prefix from ``disallow`` to ``allow`` (and vice versa).
+
+
+Examples
```
gvanrossum: I'd prefer to show the relevant example to each feature described, e.g. omitting `=`, multiple flags, `no-` and `dis`, rather than a single example block that shows all possibilities.

gvanrossum:
Once this lands we should just update the blog post.

msullivan:
It looks like I have permission to create posts but not to edit somebody else's posts on the blog

PR #7078, opened 2019-06-27
Fixes to column number reporting

JukkaL:
Previously many types and AST nodes didn't have 
(good) column numbers. This fixes several such issues.

In particular, this gives correct column number of simple 
types like `Foo` and `List[Foo]` in Python 3 type 
annotations.

For type comments we cheat a bit and point them to the
surrounding statement or definition, which is still a
small improvement.

For some types we may still be missing column numbers,
but this covers the most common cases.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -566,7 +566,8 @@ class ASTConverter:
             func_def.body.set_line(lineno)  # TODO: Why?
 
             deco = Decorator(func_def, self.translate_expr_list(n.decorator_list), var)
-            deco.set_line(n.decorator_list[0].lineno)
+            first = n.decorator_list[0]
+            deco.set_line(first.lineno, first.col_offset)
             retval = deco  # type: Union[FuncDef, Decorator]
         else:
             # FuncDef overrides set_line -- can't use self.set_line
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -517,11 +517,13 @@ class ASTConverter:
         if any(arg_types) or return_type:
             if len(arg_types) != 1 and any(isinstance(t, EllipsisType) for t in arg_types):
                 self.fail("Ellipses cannot accompany other argument types "
-                          "in function type signature.", lineno, 0)
+                          "in function type signature", lineno, n.col_offset)
             elif len(arg_types) > len(arg_kinds):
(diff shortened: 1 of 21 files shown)


mypy/fastparse.py
```diff
@@ -1207,13 +1222,21 @@ class TypeConverter:
...
+                 override_column: int = -1,
                  assume_str_is_unicode: bool = True,
                  ) -> None:
         self.errors = errors
         self.line = line
+        self.override_column = override_column
         self.node_stack = []  # type: List[AST]
         self.assume_str_is_unicode = assume_str_is_unicode
 
+    def convert_column(self, column: int) -> int:
```
ilevkivskyi: Add a docstring what is the intended use of this method?

mypy/newsemanal/semanal.py
```diff
@@ -3516,6 +3517,7 @@ def analyze_type_application(self, expr: IndexExpr) -> None:
         base = expr.base
         expr.analyzed = TypeApplication(base, types)
         expr.analyzed.line = expr.line
+        expr.analyzed.column = expr.column
```
ilevkivskyi: There are couple more similar places: `process_typevar_declaration()` and `process_newtype_declaration()` not sure if they are important.

test-data/unit/check-classes.test
```diff
@@ -5180,8 +5180,8 @@ A = G
 x: A[B[int]]
 B = G
 [out]
-main:7: error: Type argument "builtins.int" of "G" must be a subtype of "builtins.str"
 main:7: error: Type argument "__main__.G[builtins.int]" of "G" must be a subtype of "builtins.str"
+main:7: error: Type argument "builtins.int" of "G" must be a subtype of "builtins.str"
```
ilevkivskyi: Maybe add `--show-column-numbers` in this test? (IIUC columns caused the change of order.)

test-data/unit/fixtures/property_py2.pyi
```diff
@@ -0,0 +1,21 @@
...
+
+class object:
+    def __init__(self) -> None: pass
+
+class type:
+    def __init__(self, x: typing.Any) -> None: pass
+
+class function: pass
+
+property = object() # Dummy definition.
```
ilevkivskyi: Two spaces before comment?

Issue #6998, opened 2019-06-15
Avoid redundant extra errors after "Need type annotation for variable"


ilevkivskyi:
Consider this (oversimplified) example:
```python
from typing import TypeVar, Optional

T = TypeVar('T')

def f(x: Optional[T] = None) -> T: ...

x = f()  # Need type annotation for 'x'
y = x  # Cannot determine type of 'x'

def g() -> None:
    x = f()  # Need type annotation for 'x'
    y = x  # Cannot determine type of 'x'
```
The subsequent errors look redundant (and maybe even annoying). This is more important with the new semantic analyzer because the old one doesn't show the error if the problematic variable was defined at the function scope.

Potentially, we can replace `<nothing>` with `Any` after we show the error. Also the code that detects failed inference can be improved. For example, `is_valid_inferred_type()` should be replaced with a proper type visitor.

PR #7079, opened 2019-06-27
Make type inference failures more consistent

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6998

This PR makes several related changes:
* Reserve `Cannot determine type` error for cases where node deferral failed etc (for example with runtime recursive definitions); don't show it after failed (ambiguous) type inference.
* Always set reasonable variable type after failed inference (like `List[Any]` for an empty list).
* Always give `Need type annotation for variable` error, not just for instances and tuples (that was ad-hoc IMO)

This may actually give some _new_ errors as compared to status quo, but as one can see from the diff, the result is negative in number of (redundant) errors.

There are couple changes in tests that look unrelated, these are because `is_same_type(..., UnboundType())` returns `True` for everything, including `UninhabitedType`. I would say we should actually avoid leaking unbound types from semantic analysis and replace the, with `Any`, but this is a separate issue.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -4303,6 +4303,7 @@ def is_valid_inferred_type(typ: Type) -> bool:
 
 class NothingSeeker(TypeQuery[bool]):
     """Find any <nothing> types resulting from failed (ambiguous) type inference."""
+
     def __init__(self) -> None:
         super().__init__(any)
 
@@ -4312,6 +4313,7 @@ class NothingSeeker(TypeQuery[bool]):
 
 class SetNothingToAny(TypeTranslator):
     """Replace all ambiguous <nothing> types with Any (to avoid spurious extra errors)."""
+
     def visit_uninhabited_type(self, t: UninhabitedType) -> Type:
         if t.ambiguous:
             return AnyType(TypeOfAny.from_error)
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -4302,6 +4302,7 @@ def is_valid_inferred_type(typ: Type) -> bool:
 
 
 class NothingSeeker(TypeQuery[bool]):
+    """Find any <nothing> types resulting from failed (ambiguous) type inference."""
     def __init__(self) -> None:
         super().__init__(any)
 
@@ -4310,6 +4311,7 @@ class NothingSeeker(TypeQuery[bool]):
 
 
 class SetNothingToAny(TypeTranslator):
+    """Replace all ambiguous <nothing> types with Any (to avoid spurious extra errors)."""
     def visit_uninhabited_type(self, t: UninhabitedType) -> Type:
         if t.ambiguous:
             return AnyType(TypeOfAny.from_error)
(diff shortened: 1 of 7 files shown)


test-data/unit/check-generics.test
```diff
@@ -996,7 +996,7 @@ IntNode[int](1, 1)
 IntNode[int](1, 'a')  # E: Argument 2 to "Node" has incompatible type "str"; expected "int"
 
 SameNode = Node[T, T]
-ff = SameNode[T](1, 1)  # E: Need type annotation for 'ff'
+ff = SameNode[T](1, 1)
```
JukkaL: I think that this should generate an error, since `T` is not in scope here. If this seems unrelated, you can create an issue about this (and add a TODO comment here).

test-data/unit/check-newsemanal.test
```diff
@@ -2716,3 +2716,34 @@ class N(NamedTuple):
         )
         b
 [builtins fixtures/tuple.pyi]
+
+[case testNewAnalyzerLessErrorsNeedAnnotation]
```
JukkaL: What about adding a test case for `x = []  # type: ignore` followed by `reveal_type(x)`?

mypy/checker.py
```diff
@@ -4300,26 +4298,24 @@ def is_valid_inferred_type(typ: Type) -> bool:
...
-            if not is_valid_inferred_type_component(item):
-                return False
-    return True
+    def visit_uninhabited_type(self, t: UninhabitedType) -> bool:
+        return t.ambiguous
+
+
+class SetNothingToAny(TypeTranslator):
+    """Replace all ambiguous <nothing> types with Any (to avoid spurious extra errors)."""
+    def visit_uninhabited_type(self, t: UninhabitedType) -> Type:
```
JukkaL: Style nit: add empty line after class docstring.

mypy/checker.py
```diff
@@ -4300,26 +4298,24 @@ def is_valid_inferred_type(typ: Type) -> bool:
...
         # leave_partial_types when we pop a partial types scope.
         return False
-    return is_valid_inferred_type_component(typ)
+    return not typ.accept(NothingSeeker())
 
 
-def is_valid_inferred_type_component(typ: Type) -> bool:
-    """Is this part of a type a valid inferred type?
+class NothingSeeker(TypeQuery[bool]):
+    """Find any <nothing> types resulting from failed (ambiguous) type inference."""
```
JukkaL: Style nit: add empty line after class docstring.

test-data/unit/fine-grained-cycles.test
```diff
@@ -205,8 +205,7 @@ def h() -> None:
 [out]
 ==
 a.py:3: error: Invalid type "b.C"
-b.py:6: error: Need type annotation for 'c'
-b.py:7: error: Cannot determine type of 'c'
+b.py:7: error: C? has no attribute "g"
```
JukkaL: Hmm do we now propagate unbound types in type inference? Not sure if that's a good or bad thing.

ilevkivskyi: > Not sure if that's a good or bad thing.

This is rather bad, but I believe the right solution is https://github.com/python/mypy/issues/4987.

ilevkivskyi:
> I still don't fully understand what caused the different behavior in new semantic analyzer. Could you explain that in some detail?

See the discussion in PR https://github.com/python/mypy/pull/6450 and the change in `[case testUnderspecifiedInferenceResult]` there.

Issue #6953, opened 2019-06-07
New semantic analyzer: support default decorators in attrs


ilevkivskyi:
Currently this test case fails
```
[case testAttrsDefaultDecorator]
defer: Yes

import attr
@attr.s
class C(object):
    x: int = attr.ib(default=1)
    y: int = attr.ib()
    @y.default
    def inc(self):
        return self.x + 1

class Yes: ...
[builtins fixtures/list.pyi]
```
with
```
main:7: error: Non-default attributes not allowed after default attributes.
```

See also #6438

PR #7083, opened 2019-06-27
New semantic analyser: Fix special attrs decorators

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6953

I first thought about actually not modifying them in `attrs` plugin, but this is non-trivial, plus there are few places apart from the plugin where we modify decorators. So for now I just restore them after each iteration (in the same way we do this for overloads and other things).

### DIFF
--- a/test-data/unit/check-attr.test
+++ b/test-data/unit/check-attr.test
@@ -1049,3 +1049,18 @@ class B:  # E: Function is missing a type annotation for one or more arguments
 
 reveal_type(B)  # N: Revealed type is 'def (x: Any) -> __main__.B'
 [builtins fixtures/list.pyi]
+
+[case testAttrsDefaultDecoratorDeferred]
+defer: Yes
+
+import attr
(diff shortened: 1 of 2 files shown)


test-data/unit/check-attr.test
```diff
@@ -1049,3 +1049,18 @@ class B:  # E: Function is missing a type annotation for one or more arguments
 
 reveal_type(B)  # N: Revealed type is 'def (x: Any) -> __main__.B'
 [builtins fixtures/list.pyi]
+
```
JukkaL: Can you check that there are tests that run multiple semantic passes with various kinds of decorators? There is a small risk that this might break something.

ilevkivskyi: I run all test cases where top-levels processed twice (second time gives an extra iteration). I have found no failures related to this change. I however added another `attrs` specific decorator test case just in case.

Issue #6445, opened 2019-02-20
New semantic analyzer: support recursive TypedDicts and NamedTuples


ilevkivskyi:
These have (limited) support in old analyzer, with new semantic analyzer they just crash.

ilevkivskyi:
There are some tests skipped with `--no-new-semantic-analyzer` that needs to be fixed.

JukkaL:
Recursive type aliases are similar.

ilevkivskyi:
I will take care of this issue.

JukkaL:
I have a work-in-progress branch that turns these into blocking errors.

JukkaL:
I think that this is now mostly resolved. Added #7111 as a follow-up issue.

ilevkivskyi:
@JukkaL 
> I think that this is now mostly resolved. Added #7111 as a follow-up issue.

There are 7 tests that refer to this issue as an excuse, should we enable them first?
```
test-data/unit/check-classes.test:4358:# FIXME: #6445
test-data/unit/check-classes.test:4367:# FIXME: #6445
test-data/unit/check-classes.test:4375:# FIXME: #6445
test-data/unit/check-classes.test:4390:# FIXME: #6445
test-data/unit/check-unions.test:960:# https://github.com/python/mypy/issues/6445
test-data/unit/check-flags.test:1056:# Crashes on new semantic analyzer: https://github.com/python/mypy/issues/6445
test-data/unit/check-flags.test:1065:# Crashes on new semantic analyzer: https://github.com/python/mypy/issues/6445
```

JukkaL:
Yeah, let's go through all of those issues firs.

PR #7091, opened 2019-06-28
New semantic analyzer: fix recursive type aliases

JukkaL:
This fixes maximum iteration count errors.

Recursive type aliases still don't work and the behavior is a
regression from the old semantic analyzer. Previously they were
truncated, but now they aren't supported at all. We could perhaps hack
something similar to what we used to have, but I don't think that it's
urgent.

This also breaks some use cases where "semi-recursive" type aliases
were supported previously. The support mechanism caused infinite
expansion for actual recursive type aliases. I couldn't come up with a
simple way to support both, so I decided to prioritize fixing infinite
expansion, as they have a worse impact on user experience.

Work towards #6445.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2347,6 +2347,8 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 self.analyze_alias(rvalue, allow_placeholder=True)
             if not res:
                 return False
+            # TODO: Maybe we only need to reject top-level placeholders, similar
+            #       to base classes.
             if self.found_incomplete_ref(tag) or has_placeholder(res):
                 # Since we have got here, we know this must be a type alias (incomplete refs
                 # may appear in nested positions), therefore use becomes_typeinfo=True.
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -2347,7 +2347,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 self.analyze_alias(rvalue, allow_placeholder=True)
             if not res:
                 return False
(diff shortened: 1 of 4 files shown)


test-data/unit/check-newsemanal.test
```diff
@@ -1383,9 +1383,15 @@ x: B
...
-reveal_type(x[0][0])  # N: Revealed type is '__main__.C*'
+reveal_type(x)
+reveal_type(x[0][0])
 [builtins fixtures/list.pyi]
+[out]
+main:3: error: Cannot resolve name "B" (possible cyclic definition)
+main:4: error: Cannot resolve name "B" (possible cyclic definition)
+main:4: error: Cannot resolve name "C" (possible cyclic definition)
+main:7: note: Revealed type is 'Any'
+main:8: note: Revealed type is 'Any'
```
ilevkivskyi: It looks like all new failures _in this file_ are because of the problem with nested placeholders. Please add a comment in the existing issue mentioning this, and add links to the issue in these tests.

JukkaL: Added new issues about this: #7111

mypy/newsemanal/semanal.py
```diff
@@ -2347,7 +2347,7 @@ def check_and_set_up_type_alias(self, s: AssignmentStmt) -> bool:
                 self.analyze_alias(rvalue, allow_placeholder=True)
             if not res:
                 return False
-            if self.found_incomplete_ref(tag) or isinstance(res, PlaceholderType):
+            if self.found_incomplete_ref(tag) or has_placeholder(res):
```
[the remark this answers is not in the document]
JukkaL: Added a comment.

Issue #7042, opened 2019-06-21
`--no-implicit-reexport` should treat names in  `__all__` as exported


Zac-HD:
From HypothesisWorks/hypothesis#2017, passing `--no-implicit-reexport` with Mypy 0.710 causes `error: Module 'hypothesis' has no attribute 'given'` - except that it *does*, and [`given` is listed in the `__all__` attribute](https://github.com/HypothesisWorks/hypothesis/blob/master/hypothesis-python/src/hypothesis/__init__.py).

IMO when `__all__` is present *only* those names should be considered explicitly re-exported, though taking the union of those and the from-import-x-as-x form would be reasonable too.  Ignoring `__all__` is pretty weird though!

bluetech:
FWIW, I agree. I ran into this when enabling `--no-implicit-reexport` and importing from the `sentry-sdk` package (this is the file: https://github.com/getsentry/sentry-python/blob/0.9.2/sentry_sdk/__init__.py, cc @untitaker).

I also intuitively think `__all__` should define the export list when present. [Personally, I found the "`import as`-rexports" rule surprising, so I'd go further and make `__all__` the *only* way to reexport. But I realize it may be too heavy handed.]

`no-implicit-reexport` is a great feature, BTW.

untitaker:
I didn't know about this issue. Is there a workaround we can add to the SDK?

bluetech:
@untitaker Like @Zac-HD suggested in the Hypothesis issue, I just turned the feature off for `sentry-sdk` for now. I think we should wait to see whether the Mypy devs consider this a bug or not, before exploding the `import *` in `sentry-sdk`.

JukkaL:
Agreed that this looks like a bug.

JukkaL:
Anybody want to try to fix this? This doesn't sound very tricky.

bluetech:
I started writing tests for the PR but there is a behavior I'm not sure about. Currently, star imports always reexport. That is, given

```py
# private_module.py
a = 1
# public_module.py
from private_module import *
```

then `from public_module import a` is legal. This is considered an explicit reexport, so `--no-implicit-reexport` doesn't affect it.

This behavior is documented for stub files in [PEP 484](https://www.python.org/dev/peps/pep-0484/#stub-files):

> Modules and variables imported into the stub are not considered exported from the stub unless the import uses the `import ... as ... `form or the equivalent `from ... import ... as ...` form.
>
[...]

PR #7099, opened 2019-06-28
Fix variables in __all__ being flagged by --no-implicit-reexport

bluetech:
The natural expectation is that when a variable is explicitly added to a
module's `__all__`, the module explicitly means to export it.

Previously, this was not the case, and a `from [..] import [..] as [..]`
reexport was required by `--no-implicit-export` even if the variable is
listed in `__all__`.

Fixes https://github.com/python/mypy/issues/7042

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -429,8 +429,8 @@ of the above sections.
 ``--no-implicit-reexport``
     By default, imported values to a module are treated as exported and mypy allows
     other modules to import them. This flag changes the behavior to not re-export unless
-    the item is imported using from-as. Note this is always treated as enabled for
-    stub files. For example:
+    the item is imported using from-as or is included in ``__all__``. Note this is
+    always treated as enabled for stub files. For example:
 
     .. code-block:: python
 
@@ -438,6 +438,9 @@ of the above sections.
        from foo import bar
        # This will re-export it as bar and allow other modules to import it
        from foo import bar as bar
+       # This will also re-export bar
+       from foo import bar
+       __all__ = ['bar']
 
 
 ``--strict-equality``
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -310,8 +310,8 @@ Miscellaneous strictness flags
 ``implicit_reexport`` (bool, default True)
     By default, imported values to a module are treated as exported and mypy allows
     other modules to import them. When false, mypy will not re-export unless
-    the item is imported using from-as. Note that mypy treats stub files as if this
-    is always disabled. For example:
+    the item is imported using from-as or is included in ``__all__``. Note that mypy
+    treats stub files as if this is always disabled. For example:
 
     .. code-block:: python
 
@@ -319,6 +319,9 @@ Miscellaneous strictness flags
        from foo import bar
        # This will re-export it as bar and allow other modules to import it
        from foo import bar as bar
+       # This will also re-export bar
+       from foo import bar
+       __all__ = ['bar']
 
 ``strict_equality``  (bool, default False)
    Prohibit equality checks, identity checks, and container checks between
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -463,10 +463,16 @@ class SemanticAnalyzer(NodeVisitor[None],
                     del tree.names[name]
 
     def adjust_public_exports(self) -> None:
-        """Make variables not in __all__ not be public"""
+        """Adjust the module visibility of globals due to __all__."""
         if '__all__' in self.globals:
             for name, g in self.globals.items():
-                if name not in self.all_exports:
+                # Being included in __all__ explicitly exports and makes public.
+                if name in self.all_exports:
+                    g.module_public = True
+                    g.module_hidden = False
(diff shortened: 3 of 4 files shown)


bluetech:
Following @JukkaL suggestion in https://github.com/python/mypy/issues/7042#issuecomment-506690942, I tried to look at the code for a couple of hours and this is what I came up with. I am not sure it is correct at all, or if it has any unexpected consequences, so please review carefully (once/if it passes CI) :)

mypy/newsemanal/semanal.py
```diff
@@ -445,10 +445,13 @@ def add_builtin_aliases(self, tree: MypyFile) -> None:
...
 
     def adjust_public_exports(self) -> None:
-        """Make variables not in __all__ not be public"""
+        """Make variables in __all__ be exported, and variables not in
+        __all__ not be public"""
         if '__all__' in self.globals:
             for name, g in self.globals.items():
-                if name not in self.all_exports:
+                if name in self.all_exports:
+                    g.module_hidden = False
```
msullivan: Looking at the docs for `module_public` and `module_hidden`:
 * public means that it will be picked up by an `import *`
 * hidden means that it can't be imported at all

So I think we need to set `public = True` when the name does appear to prevent it from being spurious excluded from `import *`.

bluetech:
@msullivan Yes, testing with `*`-imports I see some unexpected things.

If I understand `module_public` and `module_hidden` correctly, this is what they mean (with `--no-implicit-reexport` enabled):

`module_public` | `module_hidden` | meaning
-----|-----------|-----------
false | false | cannot be imported directly, not included in `import *`
false | true | can be imported directly, not included in `import *`
true | false | cannot be imported directly, included in `import *`
true | true | can be imported directly, included in `import *`

But the "cannot be imported directly, included in import *" case, I think makes no sense? That is, I think that if a symbol is imported with `import *`, it should always be possible to import it directly.

If true, this suggests that instead of two booleans, an enum should be used, maybe something like this:

```python
class SymbolModuleVisibility:
    """The visibility of the symbol in a module, i.e. whether and how another module can import the symbol from this module.

    Only relevant for module symbol tables.
    """
    # Another module cannot import this symbol.
    hidden = 1
    # Another module can import this symbol directly.
    exported = 2
    # Another module can import this symbol directly, and with `from <module> import *`.
    exported_public = 3
```

Things which affect visibility (AFAICS) are:

- Name starts with `_`.
- Name appears in `__all__`.
- Name imported from another module with `as` (`from another_module import x as y`).
- Name is a module import without `as` (`import another_module`).
- Is stub file (per rules in [PEP 484 Stub Files section](https://www.python.org/dev/peps/pep-0484/#stub-files)).
- Implicit reexports are enabled.

Do you think it makes sense to get into that, or just do the fix you suggested and be done with it?

msullivan:
Removing the potential invalid state by using defined constants seems reasonable. (Unfortunately we can't use an `Enum` right now because of limitations in `mypyc`, a compiler we use to compile `mypy`)

bluetech:
I applied @msullivan's suggestion and also adjusted the docs.

I also added a second commit which switches the code to use an enum as proposed in the comment above.

I want to add more test cases and probably adjust the terminology in the second commit, so marking as WIP.

mypy/nodes.py
```diff
@@ -2824,6 +2824,20 @@ def accept(self, visitor: NodeVisitor[T]) -> T:
...
 
 
+class SymbolModuleVisibility:
+    """The visibility of the symbol in a module, i.e. whether and how
+    another module can import the symbol from this module.
+
+    Only relevant for module symbol tables.
+    """
+    # Another module cannot import this symbol.
+    hidden = 1  # type: Final
```
msullivan: Make these names all caps

bluetech:
Left a comment in the issue https://github.com/python/mypy/issues/7042#issuecomment-509005322 on what's blocking this.

ilevkivskyi:
@bluetech 
What is the status of this PR? It fixes a high-priority issue and looks mostly ready apart from a merge conflict caused by moving semantic analyzer files.

bluetech:
@ilevkivskyi Sorry for not following up. This needs some work/research, in particular spec'ing out the rules for each of these:

- runtime imports
- type checking regular files with implicit exports
- type checking regular files without implicit exports
- type checking stub files

per each type of import and export (partial list?)

- `import x`
- `import x as z`
- `import x.y` (what happens to `x`)
- `import x.y as z`
- `from x import y`
- `from x import y as z`
- `from x import *`
- `__all__` defined?
- symbol starts with `_`?

and I haven't had enough patience to do it. For now I'd like to close it, but instead I'll just try to go back to the initial commit and rebase which at least makes the `__all__` exports explicit, and you can decide if you want to merge it or close it.

bluetech:
Updated as described above.

ilevkivskyi:
I think we can keep this open (after https://github.com/python/mypy/pull/7361 is merged) as a place to discuss possible follow-up refactoring/clean-up that will replace `module_public`/`module_hidden` with an enum and clarifies the corner cases.

bluetech:
This was squashed in to e9cf4f2e797d76d68e895d9ecc39b1baa56dfa81 (PR #7361).

Issue #7077, opened 2019-06-27
--disallow-any-generics generates spurious messages


msullivan:
In certain cases, `--disallow-any-generics` will generate an error message in the file where a type alias is used with the line number at which the type alias is defined.

For example
```
[case testDisallowAnyGenericsMessages]
# mypy: disallow-any-generics
from a import B
x: B

[file a.py]
from typing import TypeVar, List

T = TypeVar('T')

A = List[T]
B = A

[builtins fixtures/list.pyi]
```
generates the error `main:6: error: Missing type parameters for generic type`. `main:6` is not a line that appears.

The error here is line 6 of `a.py`, `B = A`, though that is not an error in this case because `disallow-any-generics` is not specified for that file. It is triggered by line 3 of main, but reporting an error *there* would also be wrong because there is nothing wrong with line 3 of main, just with the definition of `B`. (And there is nothing that can be fixed at that call site, because `B` is not generic!).

I'm marking this as high priority because `disallow-any-generics` is a super important option that we I think we should make a default once it is fixed.

ilevkivskyi:
In general it seems to me we should prohibit using types as contexts for errors in general (or make file name part of error context).

JukkaL:
I think that types are probably fine as error contexts in the semantic analyzer.

happy2mylife:
Sorry, I have problem.

```
from typing import TypeVar, Generic

T = TypeVar('T', bound='Test')
class Test(Generic[T]):

    def __str__(self) -> str:
        return ""
```

test.py:3: error: Missing type parameters for generic type

I can't understand this error.

res0nanz:
I also have same problem with @happy2mylife .

happy2mylife:
Is this problem not solved yet? @msullivan

JelleZijlstra:
You probably need to put `Test[Any]` in your bound. This example looks rather different from the original one in this bug.

res0nanz:
If do the following,  cause `[flake8] F401:'typing.Any' imported but unused`.
```python
from typing import Any
T = TypeVar('T', bound='Test[Any]')
```
But if not import, `[mypy] error:Name 'Any' is not defined`.  How can I solve it?

I'm sorry in clumsy sentence.

PR #7100, opened 2019-06-28
Fix error reporting context for missing generic type arguments

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7077

The logic of the fix is quite straightforward: emit the error immediately when we fix an instance type, and not in a traversal after the main pass. Note that I fix this only for the new analyser, because the old one will be anyway deprecated soon.

This causes a bit of code churn because we need to pass the flags to several functions, but the logic should not change much. I didn't add many new tests, since we have a bunch of existing tests. It would probably make sense to play with this, suggestions for more tests are very welcome.


msullivan:
I wasn't able to fully test this out internally because I ran into some other issues but one thing I noticed is this makes aliases like `Foo = List` illegal. (Even though if you `# type: ignore` the error they do work correctly)

mypy/newsemanal/semanal.py
```diff
@@ -2369,7 +2369,9 @@ def check_and_set_up_type_alias(self, s: AssignmentStmt) -> bool:
         # so we need to replace it with non-explicit Anys.
         res = make_any_non_explicit(res)
         no_args = isinstance(res, Instance) and not res.args
-        fix_instance_types(res, self.fail)
+        fix_instance_types(res, self.fail,
```
msullivan: I think this needs to be done conditionally only when `no_args`, maybe?

ilevkivskyi: Yes, aliases like `A = List` should be allowed even with the flag, will fix it now.

ilevkivskyi:
@msullivan I fixed the errors you reported, could you please try again?

PR #7107, opened 2019-06-29
Remove miscellaneous 'is None' checks

Michael0x2a:
I'm embarking on a (probably futile) quest to get mypy to type-check cleanly under the new `--warn-unreachable` flag.

This pull request starts off by tackling some of the low-hanging fruit. Specifically, I make three changes:

1. I removed some unnecessary 'is None' checks for the MypyFile.path field, which is annotated to be of just type `str`, not `Optional[str]`.

    As a side-effect, this ends up fixing one itty-bitty bug: previously, we would not generate reports when running something like `mypy --any-exprs-report report_dir -c 'print("foobar")`.

2. I removed unnecessary "is this line or column None" checks from errors.py: those variables are only ever ints; we use "-1" as placeholders if those values are missing from various error messages.

3. I swapped a `is False:` check in visitors.py with a `is TYPE_CHECKING:` check. (As a simplification, mypy considers the former to be unreachable when type-checking.

Each of these three changes are split out into separate commits, in case that'll help simplify code reviewing.

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -232,7 +232,7 @@ class MypyFile(SymbolNode):
 
     # Fully qualified module name
     _fullname = None  # type: Bogus[str]
-    # Path to the file (None if not known)
+    # Path to the file (empty string if not known)
     path = ''
     # Top-level definitions and statements
     defs = None  # type: List[Statement]
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -290,8 +290,7 @@ class MypyFile(SymbolNode):
         return visitor.visit_mypy_file(self)
 
     def is_package_init_file(self) -> bool:
-        return not (self.path is None) and len(self.path) != 0 \
-            and os.path.basename(self.path).startswith('__init__.')
+        return len(self.path) != 0 and os.path.basename(self.path).startswith('__init__.')
 
     def serialize(self) -> JsonDict:
         return {'.class': 'MypyFile',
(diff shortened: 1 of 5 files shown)


mypy/nodes.py
```diff
@@ -290,8 +290,7 @@ def accept(self, visitor: NodeVisitor[T]) -> T:
         return visitor.visit_mypy_file(self)
 
     def is_package_init_file(self) -> bool:
-        return not (self.path is None) and len(self.path) != 0 \
-            and os.path.basename(self.path).startswith('__init__.')
+        return len(self.path) != 0 and os.path.basename(self.path).startswith('__init__.')
```
ilevkivskyi: Maybe also update the comment on line 235? Now it mentions it can be `None`.

mypy/visitor.py
```diff
@@ -1,10 +1,10 @@
 """Generic abstract syntax tree node visitor"""
 
 from abc import abstractmethod
-from typing import TypeVar, Generic
+from typing import TypeVar, Generic, TYPE_CHECKING
```
ilevkivskyi: Well, this should be imported from `typing_extensions`, see imports in other modules (because of Python 3.5.1).

Michael0x2a: Argh, Python 3.5 strikes yet again...

I really wish we could drop support for it early. It sometimes feels like I get bitten by some Python 3.5 related issue in every other PR I try submitting...

Issue #7097, opened 2019-06-28
The export_map field seems to be unused in NewSemanticAnalyzer


bluetech:
While looking at the code, I noticed this field is not used or filled anywhere.

 https://github.com/python/mypy/blob/efd68dd752410f4d987ce407354623938ff8e7f3/mypy/newsemanal/semanal.py#L258-L259

Grepping for `export_map` shows such a field is defined and filled in the (old?) `SemanticAnalyzerPass2` class, and used by stubgen, which if I am reading correctly is not using the new semantic analyzer.

It looks "suspicious" so thought I'd mention it.

ilevkivskyi:
Thanks for reporting! This is actually very important. We need to be sure `stubgen` works with the new analyser.

Issue #7118, opened 2019-07-01
Couple report tests show deviation on new analyzer


ilevkivskyi:
It looks like some `Any` kinds may be different on the new analyzer, see `testAnyExprReportHigherKindedTypesAreNotAny` and `testAnyExprReportIgnoresSpecialForms`.

Issue #7117, opened 2019-07-01
SamplesSuite.test_samples crashes on new semantic analyzer


ilevkivskyi:
Traceback ends in something like:
```
  File "/Users/ilevkivskyi/src/mypy/mypy/checker.py", line 1801, in visit_block
    self.accept(s)
  File "/Users/ilevkivskyi/src/mypy/mypy/checker.py", line 392, in accept
    stmt.accept(self)
  File "/Users/ilevkivskyi/src/mypy/mypy/nodes.py", line 1112, in accept
    return visitor.visit_for_stmt(self)
  File "/Users/ilevkivskyi/src/mypy/mypy/checker.py", line 3113, in visit_for_stmt
    self.analyze_index_variables(s.index, item_type, s.index_type is None, s)
  File "/Users/ilevkivskyi/src/mypy/mypy/checker.py", line 3167, in analyze_index_variables
    self.check_assignment(index, self.temp_node(item_type, context), infer_lvalue_type)
  File "/Users/ilevkivskyi/src/mypy/mypy/checker.py", line 1874, in check_assignment
    infer_lvalue_type)
  File "/Users/ilevkivskyi/src/mypy/mypy/checker.py", line 2270, in check_assignment_to_multiple_lvalues
    self.check_multi_assignment(lvalues, rvalue, context, infer_lvalue_type)
  File "/Users/ilevkivskyi/src/mypy/mypy/checker.py", line 2312, in check_multi_assignment
    context, undefined_rvalue, infer_lvalue_type)
  File "/Users/ilevkivskyi/src/mypy/mypy/checker.py", line 2413, in check_multi_assignment_from_tuple
    assert isinstance(reinferred_rvalue_type, TupleType)
AssertionError:
```

ilevkivskyi:
This one is tricky. A simple repro is:
```python
from typing import Tuple

def f() -> None:
    t: Tuple[str, Tuple[str, str, str]]
    x, (y, *z) = t
```
Moreover the reason this doesn't fail with old analyzer is because there are two bugs that cancel each other. One of them is fixed in the new analyzer (`Var.is_ready` is set consistently), so that e.g. this error is caught by the new analyzer:
```python
def g() -> None:
    def f() -> None:
        x + 'no way'  # Function level forward refs turn into silent Any in old analyzer.
    x = int()
    f()
```
I will make a PR soon.

PR #7119, opened 2019-07-01
New semantic analyser: enable more tests and fix stubgen

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7097

Previously some test files didn't use the new analyser even with `NEWSEMANAL=1`. This PR switches to more robust way of enforcing it.

I fix the tests that are easy to fix, and skip the rest (there are already two follow-up issues open: https://github.com/python/mypy/issues/7117 and https://github.com/python/mypy/issues/7118).


mypy/newsemanal/semanal.py
```diff
@@ -445,11 +445,16 @@ def add_builtin_aliases(self, tree: MypyFile) -> None:
...
+        """Make variables not in __all__ not be public.
+
+        Also record and erase value of __all__ for the current module.
+        """
         if '__all__' in self.globals:
             for name, g in self.globals.items():
                 if name not in self.all_exports:
                     g.module_public = False
+        self.export_map[self.cur_mod_id] = self.all_exports
+        self.all_exports = []
```
JukkaL: I don't like this side effect. Since this is only called in one place, why not move this to the calling function?

ilevkivskyi: OK.

PR #7123, opened 2019-07-02
Sync typeshed

ilevkivskyi:
This is expected fail, but https://github.com/python/mypy/pull/6958 will fix the name issue.


ilevkivskyi:
There are only expected failures, so I am merging this now.

PR #7124, opened 2019-07-02
New semantic analyzer: fix bogus messages and update tests

### LINKED ISSUE #6445: New semantic analyzer: support recursive TypedDicts and NamedTuples

JukkaL:
Various recursive type definitions no longer cause crashes. 

Fix bogus messages about cyclic definitions when definition is 
acyclic when there are also recursive type definitions in the same
target.

Closes #6445.

### DIFF
--- a/test-data/unit/check-flags.test
+++ b/test-data/unit/check-flags.test
@@ -1065,8 +1065,8 @@ n: N
 from typing import Dict, Any, Optional
 from mypy_extensions import TypedDict
 
-VarsDict = Dict[str, Any]  # type: ignore
-HostsDict = Dict[str, Optional[VarsDict]]  # type: ignore
+VarsDict = Dict[str, Any]
+HostsDict = Dict[str, Optional[VarsDict]]
 
 GroupDataDict = TypedDict(
     "GroupDataDict", {"children": "GroupsDict",  # type: ignore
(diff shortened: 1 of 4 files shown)


test-data/unit/check-flags.test
```diff
@@ -1053,25 +1053,23 @@ always_false = BLAH, BLAH1
...
-# Crashes on new semantic analyzer: https://github.com/python/mypy/issues/6445
-# flags: --disallow-any-generics --no-new-semantic-analyzer
+# flags: --disallow-any-generics --new-semantic-analyzer
 from typing import Dict, Any, Optional
 from mypy_extensions import TypedDict
 
-VarsDict = Dict[str, Any]
-HostsDict = Dict[str, Optional[VarsDict]]
+VarsDict = Dict[str, Any]  # type: ignore
+HostsDict = Dict[str, Optional[VarsDict]]  # type: ignore
```
ilevkivskyi: Why these two ignores are needed? Both these aliases are not recursive and don't even have forward refs.

JukkaL: Good catch! Yes, there is something suspicious going on.

Issue #7126, opened 2019-07-02
Crash on unpacking to member expression


ilevkivskyi:
This test case crashes on master:
```
[case testDeferralOfMemberNested]
from typing import Tuple

def f() -> None:
    c: C
    t: Tuple[str, Tuple[str, str]]
    x, (y, c.a) = t

class C:
    def __init__(self, a: str) -> None:
        self.a = a
```
This is likely a regression caused by https://github.com/python/mypy/pull/6860. This PR turns everything into `Any` in a deferred node, while checking the member lvalue triggers the deferral, causing then `assert isinstance(reinferred_rvalue_type, TupleType)` to trigger.

This equally affects both analysers.

PR #7127, opened 2019-07-02
Fix new semantic analyser crash on unpacking to star lvalue and enable sample tests

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7117

As explained in the issue this didn't crash with old analyser because it was compensated by another bug: `Var.is_ready` was set to `True` in some conditions thus silently inferring an `Any` type instead of deferring the node, see added test.

The actual crash was caused by the fact that star lvalue was previously checked as a generic reference expression, instead of processing it as name definition, thus triggering a bogus unexpected deferral.


ilevkivskyi:
Just to clarify, this is not the same as https://github.com/python/mypy/issues/7126. Even with the regression fixed, https://github.com/python/mypy/issues/7117 would still produce bogus `Cannot determine type of '...'` on unpacking to star lvalues that are definitions.

Issue #7071, opened 2019-06-26
New semantic analyser: Establish and document the defer() calling convention


ilevkivskyi:
(adding documentation label because it is also important for external plugin contributors)

The most important point is probably to decide on which side of `anal_type()` (and all other similar functions) we should call `defer()`. I think there are still some calls that are not needed (and may be even harmful during final iteration, note that the `defer()` calls in `typeanal.py` are correctly guarded for final iteration).

See also https://github.com/python/mypy/issues/6725

PR #7130, opened 2019-07-02
New semantic analyzer: add comments and refactor

JukkaL:
Document deferral logic in some detail and remove redundant
calls to `defer()`.

Work towards #7071.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -4513,9 +4513,10 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 analyzer
 
         Return None only if some part of the type couldn't be bound *and* it
-        referred to an incomplete namespace or definition. In this also defer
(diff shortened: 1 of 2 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -4507,9 +4505,22 @@ def anal_type(self,
...
-        and return AnyType.
+        Args:
+            typ: Type to analyze (if already analyzed, this is a no-op)
+            allow_placeholder: If True, may return PlaceholderType if
+                encountering an incomplete definition
+            third_pass: Unused; only for compatibility with old semantic
+                analyzer
+
+        Return None only if some part of the type couldn't be bound *and* it
+        referred to an incomplete namespace or definition. In this also defer
```
ilevkivskyi: Something is missing here, should it be "In this case also defer" or "In this situation also defer"?

Issue #7098, opened 2019-06-28
New analyser: Crash in attrs plugin on bad or deferred name


ilevkivskyi:
Both of these crash with the new analyser:
```
[case testTypeInAttrBad]
import attr

@attr.s
class C:
    total = attr.ib(type=Bad)

[builtins fixtures/bool.pyi]
```
and
```
[case testTypeInAttrBad2]
import attr

@attr.s
class C:
    total = attr.ib(default=func())

def func(): ...
[builtins fixtures/bool.pyi]
```
Both crash with an assertion error because plugin doesn't expect a placeholder in class symbol table. I didn't check dataclasses, but likely there is a similar issue.

PR #7132, opened 2019-07-02
New analyser: fix crashes when plugins encounter a placeholder node

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7098

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -13,7 +13,7 @@ from mypy.nodes import (
     TupleExpr, ListExpr, NameExpr, CallExpr, RefExpr, FuncDef,
     is_class_var, TempNode, Decorator, MemberExpr, Expression,
     SymbolTableNode, MDEF, JsonDict, OverloadedFuncDef, ARG_NAMED_OPT, ARG_NAMED,
-    TypeVarExpr
+    TypeVarExpr, PlaceholderNode
 )
 from mypy.plugins.common import (
     _get_argument, _get_bool_argument, _get_decorator_bool_argument, add_method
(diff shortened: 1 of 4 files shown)


mypy/plugins/attrs.py
```diff
@@ -270,7 +270,9 @@ def _analyze_class(ctx: 'mypy.plugin.ClassDefContext',
         # instance level assignments.
         if attribute.name in ctx.cls.info.names:
             node = ctx.cls.info.names[attribute.name].node
-            assert isinstance(node, Var)
+            if not isinstance(node, Var):
+                # This node may be not ready yet (i.e. a PlaceholderNode).
```
JukkaL: Should we instead check for `PlaceholderNode`? If there's something unexpected there, weird things might happen.

test-data/unit/check-attr.test
```diff
@@ -1107,3 +1107,47 @@ C(43)
...
+
+[case testTypeInAttrBad]
+import attr
+
+@attr.s
+class C:
+    total = attr.ib(type=Bad)  # E: Name 'Bad' is not defined
+[builtins fixtures/bool.pyi]
+
+[case testTypeInAttrBad2]
```
JukkaL: Maybe rename the test case? There seems nothing bad here, just a forward reference.

test-data/unit/check-attr.test
```diff
@@ -1107,3 +1107,47 @@ C(43)
...
+@attr.s
+class C:
+    total = attr.ib(type=Forward)
+
+reveal_type(C.total)  # N: Revealed type is '__main__.Forward'
+C('no')  # E: Argument 1 to "C" has incompatible type "str"; expected "Forward"
+class Forward: ...
+[builtins fixtures/bool.pyi]
+
+[case testTypeInAttrBad3]
```
JukkaL: Similar to above, the test case name doesn't seem quite right.

ilevkivskyi:
Travis looks stuck. I am going to merge this now.

PR #7134, opened 2019-07-02
Don't do any tricky inference in deferred nodes

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7126

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2410,6 +2410,10 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                                                            reinferred_rvalue_type, context,
                                                            infer_lvalue_type)
                     return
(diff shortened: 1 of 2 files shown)


mypy/checker.py
```diff
@@ -2410,6 +2410,10 @@ def check_multi_assignment_from_tuple(self, lvalues: List[Lvalue], rvalue: Expre
                                                            reinferred_rvalue_type, context,
                                                            infer_lvalue_type)
                     return
+                if isinstance(reinferred_rvalue_type, AnyType) and self.current_node_deferred:
+                    # Doing more inference in deferred nodes can be hard, so give up for now.
+                    # TODO: should we try harder?
```
JukkaL: I don't think that we need the TODO, since doing inference in deferred nodes is pretty pointless.

Issue #7076, opened 2019-06-27
New Semantic Analyzer: INTERNAL ERROR on aiohttp


sk-:
This is only a potential bug report, as I found that upgrading the triggering library solves the issue.

When analyzing code with the new semantic analyzer, `mypy` fails processing the `client` module of `aiohttp` (v3.4.4).

The difference between the affected version and the latest code is how they annotate the dataclasses using `attr.ib`. In v3.4.4 they have `total = attr.ib(type=float, default=None)`, while on v3.5.4 they use `total = attr.ib(type=Optional[float], default=None)`.

Compare https://github.com/aio-libs/aiohttp/blob/v3.4.4/aiohttp/client.py#L49-L53 and https://github.com/aio-libs/aiohttp/blob/v3.5.4/aiohttp/client.py#L132-L136

Below is the output of the run with the latest development version
```
/Users/skreft/.virtualenvs/zapship/lib/python3.6/site-packages/aiohttp/client.py:49: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 0.720+dev.262fe3dc4d5d14492c6fd4009d91c555c406ac04
Traceback (most recent call last):
  File "/Users/skreft/.virtualenvs/zapship/bin/mypy", line 10, in <module>
    sys.exit(console_entry())
  File "/Users/skreft/.virtualenvs/zapship/lib/python3.6/site-packages/mypy/__main__.py", line 8, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "/Users/skreft/.virtualenvs/zapship/lib/python3.6/site-packages/mypy/main.py", line 83, in main
    res = build.build(sources, options, None, flush_errors, fscache, stdout, stderr)
  File "/Users/skreft/.virtualenvs/zapship/lib/python3.6/site-packages/mypy/build.py", line 164, in build
    result = _build(sources, options, alt_lib_path, flush_errors, fscache, stdout, stderr)
  File "/Users/skreft/.virtualenvs/zapship/lib/python3.6/site-packages/mypy/build.py", line 224, in _build
    graph = dispatch(sources, manager, stdout)
  File "/Users/skreft/.virtualenvs/zapship/lib/python3.6/site-packages/mypy/build.py", line 2567, in dispatch
    process_graph(graph, manager)
[...]

PR #7136, opened 2019-07-02
New analyser: Fix crash in plugins when star import blocks adding a symbol

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7076

The fix is pretty straightforward. I also add a comment about `missing_names` being per-module, not per namespace.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -205,6 +205,8 @@ class NewSemanticAnalyzer(NodeVisitor[None],
     # a missing name at global scope will block adding same name at a class scope.
     # This should not affect correctness and is purely a performance issue,
     # since it can cause unnecessary deferrals.
+    # Note that a star import adds a special name '*' to the set, this blocks
+    # adding _any_ names in the current file.
     missing_names = None  # type: Set[str]
     # Callbacks that will be called after semantic analysis to tweak things.
     patches = None  # type: List[Tuple[int, Callable[[], None]]]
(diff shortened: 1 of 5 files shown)


test-data/unit/check-dataclasses.test
```diff
@@ -665,3 +665,19 @@ def func() -> int: ...
 c: C
 c.x = 1  # E: Property "x" defined in "C" is read-only
 [builtins fixtures/bool.pyi]
+
+[case testTypeInDataclassDeferredStar]
+import lib
+[file lib.py]
+from dataclasses import dataclass
+from other import *
```
JukkaL: Similar to above, maybe we should be able to process this import without adding anything to `missing_names`.

mypy/plugins/attrs.py
```diff
@@ -212,7 +212,10 @@ def attr_class_maker_callback(ctx: 'mypy.plugin.ClassDefContext',
     if ctx.api.options.new_semantic_analyzer:
         # Check if attribute types are ready.
         for attr in attributes:
-            if info[attr.name].type is None and not ctx.api.final_iteration:
+            node = info.get(attr.name)
+            if node is None:
+                return
```
JukkaL: Add comment. Discuss why we don't defer here.

mypy/newsemanal/semanal.py
```diff
@@ -201,6 +201,10 @@ class NewSemanticAnalyzer(NodeVisitor[None],
     # unbound names due to cyclic definitions and should not defer)?
     _final_iteration = False
     # These names couldn't be added to the symbol table due to incomplete deps.
+    # Note that missing names are per module, _not_ per namespace. This means that e.g.
+    # a missing name at global scope will block adding same name at a class scope.
+    # This should not affect correctness and is purely a performance issue,
+    # since it can cause unnecessary deferrals.
```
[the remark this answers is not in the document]
ilevkivskyi: It is hard to tell how easy it is. I can try it after landing this.

test-data/unit/check-dataclasses.test
ilevkivskyi: Opened https://github.com/python/mypy/issues/7141 for these two.

PR #7139, opened 2019-07-03
Add some more telemetry to graph loading

msullivan:
Internal telemetry shows it to be a hot spot in practice and I want to
dig a little deeper.

Gate some potentially expensive logging code on this path, while I'm
at it.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1016,12 +1016,15 @@ def read_deps_cache(manager: BuildManager,
 def _load_json_file(file: str, manager: BuildManager,
                     log_sucess: str, log_error: str) -> Optional[Dict[str, Any]]:
     """A simple helper to read a JSON file with logging."""
+    t0 = time.time()
     try:
         data = manager.metastore.read(file)
     except IOError:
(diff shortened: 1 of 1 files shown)


mypy/build.py
```diff
@@ -1016,12 +1016,15 @@ def read_deps_cache(manager: BuildManager,
...
     """A simple helper to read a JSON file with logging."""
+    t0 = time.time()
     try:
         data = manager.metastore.read(file)
     except IOError:
         manager.log(log_error + file)
         return None
-    manager.trace(log_sucess + data.rstrip())
+    manager.add_stats(metastore_read_time=time.time() - t0)
+    if manager.verbosity() >= 2:
```
ilevkivskyi: This `if` is not needed, `trace()` already has it.

msullivan: The idea is to avoid stripping and concatenating the strings

ilevkivskyi: Oh, I see! I would then add a comment about this.

ilevkivskyi:
I am going to include this in todays pin move.

PR #7144, opened 2019-07-03
New semantic analyzer: fix issues with 'X = X' assignments

### LINKED ISSUE #6404: New semantic analyzer: Support idem forward aliases
This is a follow-up for #6390

This tricky corner case is used in some real code and is supported by old semantic analyzer
```python
C = C
class C:  # type: ignore
    pass

x: C
```
But with new analyzer this results in `Invalid type` since the first definition always wins. Probably we just need to special-case this somehow.

JukkaL:
We special case these since a more general fix, such as looking
at line numbers of definitions, would break a ton of test cases.

Fixes #6404.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -1899,6 +1899,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                 return False
             else:
                 self.defer()
+                return True
         else:
             if sym.node is None:
                 # Something special -- fall back to normal assignment analysis.
@@ -1911,7 +1912,7 @@ class NewSemanticAnalyzer(NodeVisitor[None],
                     node.node = sym.node
                     node.kind = GDEF
                     node.fullname = sym.node.fullname()
-        return True
+            return True
 
     def should_wait_rhs(self, rv: Expression) -> bool:
         """Can we already classify this r.h.s. of an assignment or should we wait?
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -1824,6 +1824,11 @@ class NewSemanticAnalyzer(NodeVisitor[None],
 
     def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
         self.statement = s
+
+        # Special case assignment like X = X.
+        if self.analyze_identity_global_assignment(s):
+            return
+
         tag = self.track_incomplete_refs()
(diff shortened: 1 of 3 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -1868,6 +1873,46 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
...
+            return False
+        # It's an assignment like 'X = X' in the global scope.
+        name = lvalue.name
+        sym = self.lookup(name, s)
+        if sym is None:
+            if self.final_iteration:
+                # Fall back to normal assignment analysis.
+                return False
+            else:
+                self.defer()
```
ilevkivskyi: `name_not_defined()` currently already calls `defer()`, is this additional defer actually needed?

Also this returns `True`, I would make it more clear by putting the return directly here.

mypy/newsemanal/semanal.py
```diff
@@ -1868,6 +1873,46 @@ def visit_assignment_stmt(self, s: AssignmentStmt) -> None:
...
+                return False
+            if name not in self.globals:
+                # The name is from builtins. Add an alias to the current module.
+                self.add_symbol(name, sym.node, s)
+            if not isinstance(sym.node, PlaceholderNode):
+                for node in s.rvalue, lvalue:
+                    node.node = sym.node
+                    node.kind = GDEF
+                    node.fullname = sym.node.fullname()
+        return True
```
ilevkivskyi: With the above this return can be indented one level more.

test-data/unit/check-newsemanal.test
```diff
@@ -2886,3 +2886,85 @@ def f() -> None:
...
+[case testNewAnalyzerIdentityAssignment7]
+C = C # E: Name 'C' is not defined
+
+reveal_type(C) # N: Revealed type is 'Any' \
+               # E: Name 'C' is not defined
+
+[case testNewAnalyzerIdentityAssignment8]
+from typing import Final
+x: Final = 0
+x = x # E: Cannot assign to final name "x"
```
ilevkivskyi: Great to see so detailed test cases!

mypy/newsemanal/semanal.py
JukkaL: The fact that `name_not_defined` calls defer requires some pretty subtle reasoning so I'd rather keep make it explicit here.

PR #7148, opened 2019-07-04
New analyser: Fix Any kinds for special forms

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7118

Previously left hand sides of special forms were inferred as `TypeOfAny.from_error`. Note that _total_ number of expressions in one of the tests is now smaller because of the subtle logic in visiting definitions of special forms, but I don't think it is import to keep it one-to-one with old analyser.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -156,6 +156,11 @@ class ExpressionChecker(ExpressionVisitor[Type]):
     def analyze_ref_expr(self, e: RefExpr, lvalue: bool = False) -> Type:
         result = None  # type: Optional[Type]
         node = e.node
+
+        if isinstance(e, NameExpr) and e.is_special_form:
+            # A special form definition, nothing to check here.
+            return AnyType(TypeOfAny.special_form)
+
         if isinstance(node, Var):
             # Variable reference.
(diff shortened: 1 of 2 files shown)


mypy/checkexpr.py
```diff
@@ -156,6 +156,11 @@ def visit_name_expr(self, e: NameExpr) -> Type:
     def analyze_ref_expr(self, e: RefExpr, lvalue: bool = False) -> Type:
         result = None  # type: Optional[Type]
         node = e.node
+
+        if isinstance(e, NameExpr) and e.is_special_form:
```
JukkaL: Do you know how did this work with the old semantic analyzer (i.e. why do we need this now)?

ilevkivskyi: Yes, this is again the story from https://github.com/python/mypy/issues/6458. Old analyzer generates a spurious `Var` as an lvalue node for special form assignments, and thus the type of lvalue is actually inferred (but still `Any` because this is what rvalue returns). The new analyzer doesn't do this, so that we fall through to the end in the `if` just below, thus getting `TypeOfAny.from_error`.

Issue #7149, opened 2019-07-04
Crash when using Optional[TypedDictType]


JukkaL:
Mypy crashes when type checking this code:

```py
from typing import Union
from mypy_extensions import TypedDict

class A(TypedDict):
    x: int

d: Union[A, None]
d.update({'x': 1})
```

Here's the traceback:

```
...
  File "/Users/jukka/src/mypy/mypy/checker.py", line 281, in check_first_pass
    self.accept(d)
  File "/Users/jukka/src/mypy/mypy/checker.py", line 392, in accept
    stmt.accept(self)
  File "/Users/jukka/src/mypy/mypy/nodes.py", line 1002, in accept
    return visitor.visit_expression_stmt(self)
  File "/Users/jukka/src/mypy/mypy/checker.py", line 2788, in visit_expression_stmt
    self.expr_checker.accept(s.expr, allow_none_return=True, always_allow_any=True)
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 3351, in accept
    typ = self.visit_call_expr(node, allow_none_return=True)
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 234, in visit_call_expr
    return self.visit_call_expr_inner(e, allow_none_return=allow_none_return)
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 307, in visit_call_expr_inner
    object_type, member)
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 660, in check_call_expr_with_callee_type
    return self.check_union_call_expr(e, object_type, member)
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 682, in check_union_call_expr
    item_object_type))
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 657, in check_call_expr_with_callee_type
    callable_name, callee_type, e.args, e.arg_kinds, e, e.arg_names, object_type)
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 630, in transform_callee_type
    callee, args, arg_kinds, context, arg_names, object_type, signature_hook)
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 594, in apply_method_signature_hook
    MethodSigContext(object_type, formal_arg_exprs, callee, context, self.chk))
  File "/Users/jukka/src/mypy/mypy/plugins/default.py", line 325, in typed_dict_update_signature_callback
    assert isinstance(arg_type, TypedDictType)
[...]

PR #7152, opened 2019-07-04
Fix binding of self-types on unions

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7149

The fix is quite straightforward, self-types should be bound to each individual component, so I just added only basic tests. Please let me know if you want better test coverage (properties, class-methods) etc.

### DIFF
--- a/test-data/unit/check-selftype.test
+++ b/test-data/unit/check-selftype.test
@@ -503,3 +503,17 @@ class C:
 ab: Union[A, B, C]
 reveal_type(ab.x)  # N: Revealed type is 'builtins.int'
 [builtins fixtures/property.pyi]
+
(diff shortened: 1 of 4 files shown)


test-data/unit/check-selftype.test
```diff
@@ -503,3 +503,17 @@ class C:
...
+T = TypeVar('T')
+
+class A:
+    same: int
+
+class C:
+    def same(self: T) -> T: ...
+
+x: Union[A, C]
+reveal_type(x.same)  # N: Revealed type is 'Union[builtins.int, def () -> __main__.C*]'
```
JukkaL: Maybe add a test for class method with a self type?

PR #7154, opened 2019-07-05
Add missing magic methods.

kamahen:


kamahen:
This is a generalization of #4770

mypy/sharedparse.py
```diff
@@ -6,36 +6,52 @@
...
     "__int__",
     "__invert__",
     "__iter__",
     "__len__",
+    "__length_hint__",
     "__long__",
+    "__missing__",
     "__neg__",
     "__new__",
+    "__not__",
```
JukkaL: As far as I know, this method is not special. It's included in the `operator` module though but some of the names are specific to the module and aren't really special.

test-data/unit/check-classes.test
```diff
@@ -5930,6 +5930,90 @@ class B:
...
+    def __enter__(self, other: Any) -> None: return NotImplemented  # E: Returning Any from function declared to return "None"
+    def __exit__(self, other: Any) -> None: return NotImplemented  # E: Returning Any from function declared to return "None"
+    def __float__(self) -> 'A': return NotImplemented  # E: Returning Any from function declared to return "A"
+    def __floor__(self) -> 'A': return NotImplemented  # E: Returning Any from function declared to return "A"
+    def __get__(self, instance: Any, other: Any) -> str: return NotImplemented  # E: Returning Any from function declared to return "str"
+    def __getitem__(self, other: Any) -> 'A': return NotImplemented  # E: Returning Any from function declared to return "A"
+    def __int__(self) -> 'A': return NotImplemented  # E: Returning Any from function declared to return "A"
+    def __index__(self) -> 'A': return NotImplemented  # E: Returning Any from function declared to return "A"
+    def __inv__(self) -> 'A': return NotImplemented  # E: Returning Any from function declared to return "A"
```
JukkaL: Again, I think that this isn't a magic method (but `__invert__` is)?

test-data/unit/check-classes.test
```diff
@@ -5930,6 +5930,90 @@ class B:
...
+    def __rmod__(self, other: Any) -> 'A': return NotImplemented
+    def __rdivmod__(self, other: Any) -> 'A': return NotImplemented
+    def __rpow__(self, other: Any) -> 'A': return NotImplemented
+    def __rlshift__(self, other: Any) -> 'A': return NotImplemented
+    def __rrshift__(self, other: Any) -> 'A': return NotImplemented
+    def __rand__(self, other: Any) -> 'A': return NotImplemented
+    def __rxor__(self, other: Any) -> 'A': return NotImplemented
+    def __ror__(self, other: Any) -> 'A': return NotImplemented
+[builtins fixtures/notimplemented.pyi]
+=======
```
JukkaL: This seems extra. Replace with an empty line?

mypy/sharedparse.py
```diff
@@ -50,27 +66,33 @@
     "__add__",
     "__and__",
     "__cmp__",
-    "__divmod__",
+    "__concat__",
```
JukkaL: This doesn't look like a special method.

mypy/sharedparse.py
```diff
@@ -50,27 +66,33 @@
...
+    "__contains__",
     "__div__",
+    "__divmod__",
     "__eq__",
     "__floordiv__",
     "__ge__",
     "__gt__",
     "__iadd__",
     "__iand__",
+    "__iconcat__",
```
JukkaL: Similar to above, this doesn't seem to be special.

kamahen:
I re-read the documentation for magic methods and `NotImplemented` and realized that `NotImplemented` only applies to _binary_ operators. (Which seems a bit strange; for example, it means that `__bool__` can't return `NotImplemented` to force the use of `__len__`.)

So, I need to review my changes, because there might be some additional details that both @JukkaL and I have missed. I'll add a comment when I've finished this review (and have responded to @JukkaL's other comments).

test-data/unit/check-classes.test
kamahen: https://docs.python.org/3/library/operator.html
seems to show `__inv__` and `__invert__` as synonyms.
But by experiment, `__inv__` doesn't work (either in Python2.7 or Python3.x).

So, is this a bug in the documentation or is there some use of `__inv__` that I'm not aware of?

JukkaL: It looks like `__inv__` is specific to the `operator` module, so it doesn't need any special casing.

JukkaL:
I think that you accidentally added a typeshed pin update. It's causing a conflict.

kamahen:
I am not very good at github workflow, so it's very possible that I've messed things up (and I have no idea what a github "pin update" is).
I won't be able to do anything more with this until about Sept 5.

kamahen:
Obsolete

Issue #7035, opened 2019-06-21
--strict-equality doesn't work if __eq__ is decorated


ilevkivskyi:
For example this test fails, while it shouldn't:
```
[case testCustomEqDecoratedStrictEquality]
# flags: --strict-equality
from typing import TypeVar, Callable, Any

F = TypeVar('F', bound=Callable[..., Any])

def deco(f: F) -> F: ...

class Custom:
    @ deco
    def __eq__(self, other: object) -> bool: ...

Custom() == int()
[builtins fixtures/bool.pyi]
```
The problem is that `TypeInfo.get_method()` ignores `Decorator`, it only checks for `FuncBase`. I really wish we get rid of `Decorator`.

PR #7156, opened 2019-07-05
Fix --strict-equality for decorated and aliased __eq__ methods

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7035

The fix is straightforward: we use `get()` instead of `get_method()` and check for more symbol kinds.


ilevkivskyi:
> Looks good! I wonder if we should deprecate `get_method`, since it seems pretty error-prone.

I think yes, but it may require some work because it is still used in 12 places (and in some plugins).

PR #7158, opened 2019-07-05
New analyzer: fix issues with order-dependent results

### LINKED ISSUE #7157: New analyzer: module processing order may affect import resolution
These two test cases should be equivalent (with new semantic analyzer), but the second generates an error. The only difference is the order in which modules are processed within an import cycle.

```
[case testNewAnalyzerImportStarForwardRef1]
import a

[file a.py]
x: A

from b import * # No error

class A: pass

[file b.py]
MYPY = False
if MYPY:
    from a import x

[case testNewAnalyzerImportStarForwardRef2]
import a

[file a.py]
x: A

MYPY = False
if MYPY:
    from b import * # Unexpected error: Name 'x' already defined on line 1

class A: pass

[file b.py]
from a import x
```

JukkaL:
When we add a placeholder, we should do it through `mark_incomplete`,
since the name needs to added to the set of missing names. Otherwise
a later definition can take precedence over an earlier definition,
if the first definition initially creates a placeholder.

Fixes #7157 and probably another issue related to type alises, which
was exposed by the fix.

### DIFF
--- a/mypy/newsemanal/semanal.py
+++ b/mypy/newsemanal/semanal.py
@@ -204,7 +204,10 @@ class NewSemanticAnalyzer(NodeVisitor[None],
     # Note that missing names are per module, _not_ per namespace. This means that e.g.
     # a missing name at global scope will block adding same name at a class scope.
     # This should not affect correctness and is purely a performance issue,
-    # since it can cause unnecessary deferrals.
+    # since it can cause unnecessary deferrals. These are represented as
+    # PlaceholderNodes in the symbol table. We use this to ensure that the first
+    # definition takes precedence even if it's incomplete.
+    #
     # Note that a star import adds a special name '*' to the set, this blocks
     # adding _any_ names in the current file.
     missing_names = None  # type: Set[str]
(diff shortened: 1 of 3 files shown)


mypy/newsemanal/semanal.py
```diff
@@ -1699,7 +1702,9 @@ def process_imported_symbol(self,
             if self.final_iteration:
                 self.report_missing_module_attribute(module_id, id, imported_id, context)
                 return
-            self.record_incomplete_ref()
+            else:
+                # This might become a type.
+                self.mark_incomplete(imported_id, node.node, becomes_typeinfo=True)
```
ilevkivskyi: I think this requires an update to the `becomes_typeinfo` docs, currently it says "will", not "might".

JukkaL: Good point. I'll update this in a separate PR, since I want to merge this now.

JukkaL:
Once #7160 has landed, I'll add some tests for order of iterations.

Issue #6725, opened 2019-04-26
Add a method for conditional deferral and use it in the new analyzer


ilevkivskyi:
This idea appeared in https://github.com/python/mypy/pull/6724. It will be useful for plugins and in general to have something like this in `SemanticAnalyzer`:
```python
def defer_if_possible(self) -> bool:
    if not self.final_iteration:
        self.defer()
        return True
    return False
```
Then it can be used like this:
```python
if ctx.api.defer_if_possible():
    ...
else:
    ... # probably give some error
```
Maybe we can also add it to the old analyzer as always returning `False`, then we can completely hide the old vs new analyzer details in plugin API.

JukkaL:
`process_placeholder` is enough for now, as discussed in #7159.

PR #7135, opened 2019-07-02
Document how plugins can support the new semantic analyzer

JukkaL:
Work towards #7071 (and more).

PR #7159, opened 2019-07-05
New analyzer: improve documentation of defer()

### LINKED ISSUE #7071: New semantic analyser: Establish and document the defer() calling convention

JukkaL:
This concludes documentation that started in #7130 and #7135.

Closes #7071.


JukkaL:
Let's close #6725, since `process_placeholder` seems mostly good enough.

Issue #6341, opened 2019-02-05
New semantic analyzer: tests for number of iterations


JukkaL:
It would be nice to be able to test things like "the top level of module `m` should be analyzed exactly twice". I've had cases where there were more iterations than expected, and these kinds of errors are hard to detect without tests.

PR #7160, opened 2019-07-05
Add infrastructure for testing order of processing nodes in new analyser

ilevkivskyi:
See https://github.com/python/mypy/issues/6341

This PR doesn't add actual tests, just the infra. I simply expose the existing machinery we use in the fine grained mode also to the normal tests.

Using this opportunity I also refactor `assert_module_equivalence()` and `assert_target_equivalence()` by moving `is None` checks to callers.

### DIFF
--- a/mypy/test/testcheck.py
+++ b/mypy/test/testcheck.py
@@ -16,9 +16,10 @@ from mypy.test.data import (
 from mypy.test.helpers import (
     assert_string_arrays_equal, normalize_error_messages, assert_module_equivalence,
     retry_on_error, update_testcase_output, parse_options,
-    copy_and_fudge_mtime
+    copy_and_fudge_mtime, assert_target_equivalence
 )
 from mypy.errors import CompileError
+from mypy.newsemanal.semanal_main import core_modules
 
 
 # List of files that contain test case descriptions.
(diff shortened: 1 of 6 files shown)


mypy/test/testcheck.py
```diff
@@ -222,6 +223,15 @@ def run_case_once(self, testcase: DataDrivenTestCase,
...
 
+            if options.new_semantic_analyzer:
+                name = 'targets'
+                if incremental_step:
+                    name += str(incremental_step + 1)
+                expected = testcase.expected_fine_grained_targets.get(incremental_step + 1)
+                actual = res.manager.processed_targets
+                # Skip the initial builtin cycle.
+                actual = [t for t in actual if not any(t.startswith(mod) for mod in core_modules)]
+                assert_target_equivalence(name, expected, actual)
```
JukkaL: It's pretty confusing that this does nothing if `expected` is None. Can you refactor this so that the `None` check happens in the caller instead?

ilevkivskyi: FWIW this matches the behaviour of a similar helper `assert_module_equivalence()` for `[rechecked]` and `[stale]`. I would rather either change them both, or leave them as is.

JukkaL: Sure, it's fine to modify both (or neither).

ilevkivskyi: OK, I removed it from both and moved to the callers.

Issue #7036, opened 2019-06-21
Dict and TypedDict are considered non-overlapping


ilevkivskyi:
For example this code doesn't type-check with `--strict-equality`:
```python
class Config(TypedDict):
    a: str
    b: str

x: Dict[str, object]
y: Config

x == y
``` 
I think we should consider typed dicts overlapping with `Dict[str, <sufficiently wide type>]`, and with `Dict[<nothing>, <nothing>]` (empty dict) only for the case where all keys are optional.

Also we should probably flag the corresponding example with overloads (but I am less convinced):
```python
@overload
def f(x: Config) -> int: ...
@overload
def f(x: Dict[str, object]) -> str: ...
def f(x):
    pass

z: Dict[str, object] = {'a': 'oh', 'b': 'no'}
f(z)  # Is this int or str?
```
@Michael0x2a what do you think?

Michael0x2a:
I agree with your analysis -- the last example should probably be flagged, yeah.

PR #7164, opened 2019-07-06
Fix overlap between TypedDict and Mapping

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7036

This adds more "fine grained" logic for overlap between `TypedDict` and `Mapping` used by overload checks and `--strict-equality`. Basically the rules are:
* A `TypedDict` with some required keys is overlapping with `Mapping[str, <some type>]` if and only if every key type is overlapping with `<some type>`.
* A `TypedDict` with no required keys overlaps with `Mapping[str, <some type>]` if and only if at least one of key types overlaps with `<some type>`.
* Empty dictionaries are as usual in gray area, so we follow the same logic as in other places: an empty dictionary can't cause an overlap (i.e. `TypedDict` with no required keys doesn't overlap with `Mapping[str, <some type>]`), but is itself (i.e. `Mapping[<nothing>, <nothing>]`) overlapping with a `TypedDict` with no required keys.

### DIFF
--- a/mypy/meet.py
+++ b/mypy/meet.py
@@ -204,6 +204,8 @@ def is_overlapping_types(left: Type,
 
     if isinstance(left, TypedDictType) and isinstance(right, TypedDictType):
         return are_typed_dicts_overlapping(left, right, ignore_promotions=ignore_promotions)
+    elif typed_dict_mapping_pair(left, right):
+        return typed_dict_mapping_overlap(left, right)
     elif isinstance(left, TypedDictType):
         left = left.fallback
     elif isinstance(right, TypedDictType):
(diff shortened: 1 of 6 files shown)


silviogutierrez:
I checked this out eagerly, as I thought it covers a big use case of @cs-cordero and mine's. But now that I read it closely, it seems *related* but not the actual solution. That is, this still won't work:

```python3
from typing import Mapping

def mapping_test(thing: Mapping[str, str]) -> bool:
    return True


class Foo(TypedDict):
    foo: str


a = Foo(foo="a")
b: Foo = {'foo': 'a'}


mapping_test(a)  # Errors out still, with below.
mapping_test(b)  # Errors out still, with below.
# Argument 1 to "mapping_test" has incompatible type "Foo"; expected "Mapping[str, str]"
```

You mention "overlap" but in this case, it's referring strictly to equality and overloads, not general types matching. Is that right?

Is our use case a big stretch?

Either way, thanks for all the work on mypy!

silviogutierrez:
Adding a more thorough snippet that demonstrates what works and what doesn't as of this PR, in case it's helpful. All running with `--strict-equality`

```python3
from typing import Mapping

def mapping_test(thing: Mapping[str, str]) -> bool:
    return True


class Foo(TypedDict):
    foo: str


a = Foo(foo="a")
b: Foo = {"foo": "a"}
c: Mapping[str, str] = {"foo": "a"}


a == b == c  # Works


mapping_test(c)  # Works, of course
mapping_test(a)  # Does not work
mapping_test(b)  # Does not work
```

ilevkivskyi:
@silviogutierrez Your issue is completely unrelated. The code is your example is unsafe because typed dicts use structural subtyping, you can use `Mapping[str, object]` instead.

PR #7166, opened 2019-07-06
Add more details for "Invalid type" errors

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/4030

This adds some more details to various `Invalid type` errors, and similar such as `Invalid type alias`, and `Invalid base class`. Unfortunately, `MessageBuilder` is not available in `typeanal.py`, so I put some error message formatting logic there.

Please add ideas about how to improve these errors.

### DIFF
--- a/test-data/unit/check-incremental.test
+++ b/test-data/unit/check-incremental.test
@@ -2196,9 +2196,11 @@ from b import x
 1 + 1
 [out]
 [out2]
-tmp/b.py:2: error: Invalid type "c.C": Cannot use function as type, use Callable[...] or a protocol instead
+tmp/b.py:2: error: Function "c.C" is not valid as a type
+tmp/b.py:2: note: Perhaps you need "Callable[...]" or a callback protocol?
 [out3]
-tmp/b.py:2: error: Invalid type "c.C": Cannot use function as type, use Callable[...] or a protocol instead
+tmp/b.py:2: error: Function "c.C" is not valid as a type
+tmp/b.py:2: note: Perhaps you need "Callable[...]" or a callback protocol?
 
 [case testCacheDeletedAfterErrorsFound3]
 import a
--- a/test-data/unit/check-incremental.test
+++ b/test-data/unit/check-incremental.test
@@ -2180,6 +2180,7 @@ tmp/b.py:1: error: Module 'c' has no attribute 'x'
 tmp/b.py:1: error: Module 'c' has no attribute 'x'
 
 [case testCacheDeletedAfterErrorsFound2]
+# flags: --new-semantic-analyzer
 import a
 [file a.py]
 from b import x
@@ -2195,9 +2196,9 @@ from b import x
 1 + 1
 [out]
 [out2]
-tmp/b.py:2: error: Invalid type "c.C"
+tmp/b.py:2: error: Invalid type "c.C": Cannot use function as type, use Callable[...] or a protocol instead
 [out3]
-tmp/b.py:2: error: Invalid type "c.C"
+tmp/b.py:2: error: Invalid type "c.C": Cannot use function as type, use Callable[...] or a protocol instead
 
 [case testCacheDeletedAfterErrorsFound3]
 import a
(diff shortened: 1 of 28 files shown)


test-data/unit/check-incremental.test
```diff
@@ -2195,9 +2196,9 @@ from b import x
 1 + 1
 [out]
 [out2]
-tmp/b.py:2: error: Invalid type "c.C"
+tmp/b.py:2: error: Invalid type "c.C": Cannot use function as type, use Callable[...] or a protocol instead
```
JukkaL: Again, the first part could be shortened to `Function "c.C" is not valid as a type`. I'd move the final part to a separate note and rephrase it, since it's not clear of `Callable[...]` or a protocol is actually what the user wants. Maybe something like this:

```
b.py:2: error: Function "c.C" is not a valid type
b.py:2: note: Perhaps you need Callable[...] or a callback protocol?
```

test-data/unit/check-literal.test
```diff
@@ -914,28 +916,31 @@ c: {"a": 1, "b": 2}             # E: Invalid type comment or annotation
 d: {1, 2, 3}                    # E: Invalid type comment or annotation
 
 [case testLiteralDisallowCollections2]
+# flags: --new-semantic-analyzer
 from typing_extensions import Literal
 a: (1, 2, 3)                    # E: Syntax error in type annotation \
                                 # N: Suggestion: Use Tuple[T1, ..., Tn] instead of (T1, ..., Tn)
 b: Literal[[1, 2, 3]]           # E: Parameter 1 of Literal[...] is invalid
-c: [1, 2, 3]                    # E: Invalid type
+c: [1, 2, 3]                    # E: Invalid type: Did you want to use "List[...]"?
```
JukkaL: When we give a suggestion, we typically give it as a separate line. For example:

```
x.py:2: error: Expression "[...]" is not a valid type
x.py:2 note: Did you mean "List[t]"? 
```

We could do the same thing for `x: (int, str)` (`Tuple[t1, ..., tN]`), `x: {int}` (`Set[t]`) and `x: {int: str}` (`Dict[kt, vt]`).

test-data/unit/check-newtype.test
```diff
@@ -295,15 +295,16 @@ a = Foo(type(3))
...
 A = NewType('A', T)
 B = NewType('B', List[T])
 [builtins fixtures/list.pyi]
 [out]
-main:3: error: Argument 2 to NewType(...) must be subclassable (got T?)
-main:3: error: Invalid type "__main__.T"
-main:4: error: Invalid type "__main__.T"
+main:4: error: Argument 2 to NewType(...) must be subclassable (got T?)
+main:4: error: Invalid type "__main__.T": Can only use bound type variables as types
+main:5: error: Invalid type "__main__.T": Can only use bound type variables as types
```
JukkaL: Bikeshedding: `Type variable "__main__.T" is unbound`. Maybe also add a note about what this means, since unbound/bound can be confusing for new users. Idea:

```
main:4: error: Type variable "__main__.T" is unbound
main:4:note: (Hint: Use "Generic[T]" or "Protocol[T]" base class to bind "T" inside a class)
main:4:note: (Hint: Use "T" in function signature to bind "T" inside a function)
```

test-data/unit/check-statements.test
```diff
@@ -1351,13 +1351,13 @@ with A() as (a, b):
...
 [case testWithStmtTypeComment]
-
+# flags: --new-semantic-analyzer
 from typing import Union
 class A:
     def __enter__(self) -> int: pass
     def __exit__(self, x, y, z): pass
 
-with A():  # type: int  # E: Invalid type comment
+with A():  # type: int  # E: Invalid type comment: `with` statement has no targets
```
JukkaL: Style nit: use double quotes (`... "with" statement ...`).

test-data/unit/check-literal.test
ilevkivskyi: We already do this for tuples, dict and set are more tricky (and I think more rare) so I would leave this for a separate PR.

PR #7167, opened 2019-07-08
Avoid telling the user to use type annotations on pre-3.5

### LINKED ISSUE #6673: "Need type annotation for '<variable>'" in Python < 3.6
mypy is complaining about missing variable type annotations even if the Python version is explicitly set.

This is the code I have:
```
i_am_a_set = set()
```
I then run mypy with the command:
```
mypy --python-version 3.5 test.py
```
This results in an error message saying `error: Need type annotation for 'i_am_a_set'`. However, since variable annotations were introduced in Python 3.6, this is clearly a nonsense error for Python 3.5.

Also please note that if I change the code to `i_am_a_set: Set[int] = set()`, and run mypy again with `--python-version 3.5`, I get: `error: Variable annotation syntax is only supported in Python 3.6 and greater`. So mypy knows that this annotation is invalid in Python 3.5, but still requests it.

I used the latest release version of mypy (0.700).

refi64:
Closes #6673.

A year and a half since my last PR here... Turns out I haven't died after all :rofl:

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -1087,6 +1087,7 @@ class MessageBuilder:
     def need_annotation_for_var(self, node: SymbolNode, context: Context,
                                 python_version: Optional[Tuple[int, int]] = None) -> None:
         hint = ''
+        has_variable_annotations = not python_version or python_version >= (3, 6)
         # Only gives hint if it's a variable declaration and the partial type is a builtin type
         if (python_version and isinstance(node, Var) and isinstance(node.type, PartialType) and
                 node.type.type and node.type.type.fullname() in reverse_builtin_aliases):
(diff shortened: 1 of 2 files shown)


mypy/messages.py
```diff
@@ -1095,11 +1096,17 @@ def need_annotation_for_var(self, node: SymbolNode, context: Context,
...
-        self.fail("Need type annotation for '{}'{}".format(unmangle(node.name()), hint), context)
+            else:
+                hint = ' (hint: "{} = ...  # type: {}[{}]")'.format(node.name(), alias, type_dec)
+
+        if has_variable_annotations:
+            needed = 'annotation'
+        else:
+            needed = 'comment'
+
+        self.fail("Need type {} for '{}'{}".format(needed, unmangle(node.name()), hint), context)
```
blueyed: Note: this make it impossible to grep for "Need type annotation" in the source to find this.

blueyed:
I've wondered about making this a note instead of an error - for the use case of `check_untyped_defs = True`, where you then get "real"/other errors still, but those would become just notes.

mypy/messages.py
ethanhs: "impossible"? not at all! You can grep for "Need type .* for" or even pass -e with multiple expressions or do grep -E with a `|` in the expression.

blueyed: Yes, I've found it after all.. ;)
This was just meant as feedback that this pattern makes it harder to grep for (log) messages.

ethanhs: That is fair and well received feedback, but I think that will get solved by Jukka's error code work anyway. Thank you for raising this point all the same.

blueyed: Ok, sorry for the exaggeration.
My main question is the one below, i.e. if this should be a note instead.

PR #7168, opened 2019-07-08
Don't define a duplicate __init__ for dataclasses

### LINKED ISSUE #6399: Wrong inference of dataclasses with custom __init__ and varargs.
Small case to reproduce:

```python
from typing import Tuple

@dataclass
class Foo:
    spams: Tuple[int, ...]

    def __init__(self, *spams: int) -> None:
        self.spams = spams

def build_foo(*spams: int) -> Foo:
    return Foo(*spams)
```

Checking that program with mypy 0.670 (Python 3.6), results in:

```
bug.py:13: error: Argument 1 to "Foo" has incompatible type "*Tuple[int, ...]"; expected "Tuple[int, ...]"
```

Without the `@dataclass` decorator the program passes.  It seems that the dataclass is hiding the actual ` __init__`.

refi64:
Closes #6399. Pretty simple fix but I ran into it so figured might as well fix it. :man_shrugging:

### DIFF
--- a/mypy/plugins/dataclasses.py
+++ b/mypy/plugins/dataclasses.py
@@ -92,7 +92,11 @@ class DataclassTransformer:
             'frozen': _get_decorator_bool_argument(self._ctx, 'frozen', False),
         }
 
-        if decorator_arguments['init']:
+        # If there are no attributes, it may be that the new semantic analyzer has not
+        # processed them yet. In order to work around this, we can simply skip generating
+        # __init__ if there are no attributes, because if the user truly did not define any,
+        # then the object default __init__ with an empty signature will be present anyway.
+        if decorator_arguments['init'] and '__init__' not in info.names and attributes:
             add_method(
                 ctx,
                 '__init__',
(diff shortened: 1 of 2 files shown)


mypy/plugins/dataclasses.py
```diff
@@ -92,7 +92,7 @@ def transform(self) -> None:
             'frozen': _get_decorator_bool_argument(self._ctx, 'frozen', False),
         }
 
-        if decorator_arguments['init']:
+        if decorator_arguments['init'] and info.get('__init__') is None:
```
ilevkivskyi: This can get `__init__()` from `object`, this is why many tests failed. I think this should be simply `and '__init__' not in info.names`.

refi64: @ilevkivskyi ugh not my brightest moment, fixed.

refi64:
Welp I broke the new semantic analyzer, everything is complicated :rofl:

refi64:
This is interesting, `testTypeInDataclassDeferredStar` is failing because under the new semantic analyzer, attributes is empty on the first round. Before this worked because __init__ was unconditionally overwritten every time, so the newer __init__ that actually contained attributes would override the empty one that was generated the first time. To work around this, I just skip generating __init__ if the attribute list is empty, since in that case the autogenerated __init__ would be identical to object's anyway.

refi64:
Okay it's all green now!

Issue #4168, opened 2017-10-27
binder is too optimistic about refining attribute types in methods


carljm:
This example typechecks in strict-optional mode, but it is not type safe:

```
from typing import Optional 
 
 
class Foo: 
    bar: Optional[int] 
 
    def thing(self) -> int: 
        if self.bar is None: 
            return 0 
        self.set_bar_to_none() 
        return self.bar 
 
    def set_bar_to_none(self) -> None: 
        self.bar = None 
```

elazarg:
Looks like a deep problem. I assume there's a similar problem with any "literal" that is not a local variable.

gvanrossum:
The binder should probably assume that any method call on an object might modify any attributes of that object (within their declared types).

carljm:
I don't know what the desired balance is between paranoid type-safety vs probably-correct-in-most-cases, but only watching for direct method calls on the object is not fully safe either, `set_bar_to_none` could just as well be a standalone function that takes a `Foo` instance.

Next step would be to consider any method call, or other call taking `self`, as unsafe. But even this probably leaves loopholes where there might be other references available to the current instance. Ultimately probably any intervening call is not safe.

JukkaL:
This is basically the binder working as designed. As @carljm explained well, doing this safely would have to be overly restrictive. 

The realistic alternatives we have is to either only use the binder for local variables, or use the current, a-little-too-optimistic approach. I think that here convenience wins over absolute safety. I suspect that if we'd restrict the binder to local variables only, many users would use (unsafe) casts instead of refactoring their code to safely conform to this restriction, actually resulting in more unsafety than we have now.

carljm:
I suppose in theory (given someone who cares enough to implement it) this is the kind of behavior that could be selected by a strictness flag? On a brand-new project I could imagine preferring to opt in to the strict behavior. Typically the workaround is just `bar = self.bar` and use the local variable, which isn't too onerous.

JukkaL:
Yeah, we could use a strictness flag for this -- but this doesn't feel like a high priority feature. Let's keep the issue open to see if there seems to be additional interest in this.

ilevkivskyi:
Just a minor stylistic comment. What happens here (the side effect) is typically absent in languages with strict typing discipline (i.e. they typically only allow pure functions). This is not the case for Python, but I think it will not be a surprise to see such behaviour in relation with static types.

Also related https://github.com/python/mypy/issues/4019.

msullivan:
[...]

PR #7169, opened 2019-07-08
Make reachability code understand chained comparisons

Michael0x2a:
Currently, our reachability code does not understand how to parse comparisons like `a == b == c`: the `find_isinstance_check` method only attempts to analyze comparisons that contain a single `==`, `is`, or `in` operator.

This pull request generalizes that logic so we can support any arbitrary number of comparisons.

It also along the way unifies the logic we have for handling `is` and `==` checks: the latter check is now just treated a weaker variation of the former. (Expressions containing `==` may do arbitrary things if the
underlying operands contain custom `__eq__` methods.)

As a side-effect, this PR adds support for the following:

    x: Optional[str]
    if x is 'some-string':
        # Previously, the revealed type would be Union[str, None]
        # Now, the revealed type is just 'str'
        reveal_type(x)
    else:
        reveal_type(x)  # N: Revealed type is 'Union[builtins.str, None]'

We previously supported this narrowing logic when doing equality checks (e.g. when doing `if x == 'some-string'`).

As a second side-effect, this PR adds support for the following:

    class Foo(Enum):
        A = 1
        B = 2

    y: Foo
    if y == Foo.A:
        reveal_type(y)  # N: Revealed type is 'Literal[Foo.A]'
    else:
        reveal_type(y)  # N: Revealed type is 'Literal[Foo.B]'

We previously supported this kind of narrowing only when doing identity checks (e.g. `if y is Foo.A`).

Note that to avoid any bad interactions with custom `__eq__` methods, we enable this narrowing check only if both operands do not define custom `__eq__` methods.


Michael0x2a:
A few meta notes:

- This PR is a follow-up to https://github.com/python/mypy/pull/7000 and builds on top of the work I did there. Once that other PR is merged, I'll rebase this one to remove the duplicate commits: only the last commit in this PR is new.
- I'm thinking we do *not* try landing this before the mypy 0.720 release? Unlike the other PR, these changes are somewhat intrusive, and I'm not sure if I'll have time to follow-up on any weirdness this PR might uncover for the next few days.

test-data/unit/check-enum.test
```diff
@@ -854,3 +902,81 @@ def process(response: Union[str, Reason] = '') -> str:
...
+x: Foo
+y: Foo
+if x is Foo.A is y:
+    reveal_type(x)  # N: Revealed type is 'Literal[__main__.Foo.A]'
+    reveal_type(y)  # N: Revealed type is 'Literal[__main__.Foo.A]'
+else:
+    reveal_type(x)  # N: Revealed type is '__main__.Foo'
+    reveal_type(y)  # N: Revealed type is '__main__.Foo'
+
+if x == Foo.A == y:
```
msullivan: Does x == y == Foo.A work? (It feels like we ought to support this or I'm not sure how much this helps?)

mypy/checker.py
```diff
@@ -3719,6 +3713,78 @@ def find_isinstance_check(self, node: Expression
...
+        producing TypeMaps refining the types of either the left or right exprs (or neither,
+        if we can't learn anything from the comparison).
+
+        For more details about what TypeMaps are, see the docstring in find_isinstance_check.
+
+        If 'assume_identity' is true, assume that this comparison was done using an
+        identity comparison (left_expr is right_expr), not just an equality comparison
+        (left_expr == right_expr). Identity checks are not overridable, so we can infer
+        more information in that case.
+        """
```
ilevkivskyi: Also update docstring style here.

mypy/checker.py
```diff
@@ -4615,6 +4681,32 @@ def is_private(node_name: str) -> bool:
     return node_name.startswith('__') and not node_name.endswith('__')
 
 
+def uses_default_equality_checks(typ: Type) -> bool:
+    """Returns 'true' if we know for certain that the given type is using
```
```suggestion
    """Returns 'True' if we know for certain that the given type is using
```

mypy/checker.py
```diff
@@ -3622,67 +3622,61 @@ def find_isinstance_check(self, node: Expression
...
-                    other_type = operand_types[1 - i]
-
-                    if literal(expr) == LITERAL_TYPE and is_singleton_type(other_type):
-                        # This should only be true at most once: there should be
-                        # exactly two elements in node.operands and if the 'other type' is
-                        # a singleton type, it by definition does not need to be narrowed:
-                        # it already has the most precise type possible so does not need to
-                        # be narrowed/included in the output map.
-                        #
-                        # TODO: Generalize this to handle the case where 'other_type' is
```
ilevkivskyi: Was this TODO moved somewhere or is it now fixed? Do we have regression test in the latter case?

mypy/checker.py
```diff
@@ -3719,6 +3713,78 @@ def find_isinstance_check(self, node: Expression
...
+        For more details about what TypeMaps are, see the docstring in find_isinstance_check.
+
+        If 'assume_identity' is true, assume that this comparison was done using an
+        identity comparison (left_expr is right_expr), not just an equality comparison
+        (left_expr == right_expr). Identity checks are not overridable, so we can infer
+        more information in that case.
+        """
+
+        # For the sake of simplicity, we currently attempt inferring a more precise type
+        # for just one of the two variables.
```
ilevkivskyi: How important is this? If we actually do this, I think it is better to use the type of right one to restrict the type of left one.

mypy/checker.py
```diff
@@ -3719,6 +3713,78 @@ def find_isinstance_check(self, node: Expression
...
+            # Note: This check is actually strictly speaking unsafe: stripping away the 'None'
+            # would be unsound in the case where A defines an '__eq__' method that always
+            # returns 'True', for example.
+            #
+            # We implement this check partly for backwards-compatibility reasons and partly
+            # because those kinds of degenerate '__eq__' implementations are probably rare
+            # enough that this is fine in practice.
+            #
+            # We could also probably generalize this block to strip away *any* singleton type,
+            # if we were fine with a bit more unsoundness.
```
ilevkivskyi: Prepend the last sentence with `TODO:`?

ilevkivskyi:
> or binder being too optimistic

Just to clarify I mean something like this https://github.com/python/mypy/issues/4168#issuecomment-522031920

Michael0x2a:
Superseded by https://github.com/python/mypy/pull/8148.

PR #7173, opened 2019-07-08
New analyzer: add more tests for order of processing

### LINKED ISSUE #6341: New semantic analyzer: tests for number of iterations

JukkaL:
Closes #6341.


ilevkivskyi:
@JukkaL The test fails likely because dictionaries (i.e. symbol tables in our case) are not ordered on Python 3.5.

PR #7175, opened 2019-07-08
Sync typeshed

ilevkivskyi:


ilevkivskyi:
The sync is in preparation for 0.720 release.

Issue #7044, opened 2019-06-22
New semantic analyzer: crash on assignment to sqlalchemy @hybrid_property


Deimos:
I'm testing out the new semantic analyzer, but it crashes on my codebase. Trying to assign to a [SQLAlchemy `@hybrid_property`](https://docs.sqlalchemy.org/en/13/orm/extensions/hybrid.html)) results in a `Cannot assign to a method` error followed by a crash.

Here's a fairly minimal reproduction for it:

```python3
from sqlalchemy import Base, Column, String
from sqlalchemy.ext.hybrid import hybrid_property

class FirstNameOnly(Base):
    first_name = Column(String)

    @hybrid_property
    def name(self) -> str:
        return self.first_name

    @name.setter  # type: ignore
    def name(self, value: str) -> None:
        self.first_name = value

    def __init__(self, name: str):
        self.name = name
```

The `# type: ignore` comment on the setter is a workaround for this issue I reported last year: https://github.com/python/mypy/issues/4430

Here's the full output:

```
(tildes) vagrant@ubuntu-xenial:/opt/tildes$ mypy --new-semantic-analyzer --show-traceback test_mypy.py 
test_mypy.py:16: error: Cannot assign to a method
test_mypy.py:16: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 0.720+dev.48916e63403645730a584d6898fbe925d513a841
Traceback (most recent call last):
  File "/opt/venvs/tildes/bin/mypy", line 10, in <module>
    sys.exit(console_entry())
  File "/opt/venvs/tildes/lib/python3.7/site-packages/mypy/__main__.py", line 8, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "/opt/venvs/tildes/lib/python3.7/site-packages/mypy/main.py", line 83, in main
    res = build.build(sources, options, None, flush_errors, fscache, stdout, stderr)
  File "/opt/venvs/tildes/lib/python3.7/site-packages/mypy/build.py", line 164, in build
    result = _build(sources, options, alt_lib_path, flush_errors, fscache, stdout, stderr)
  File "/opt/venvs/tildes/lib/python3.7/site-packages/mypy/build.py", line 224, in _build
    graph = dispatch(sources, manager, stdout)
[...]

PR #7181, opened 2019-07-09
New semantic analyzer: fix crash on broken overload at class scope

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7044

The crash is caused by the fact that `.type` for broken overloads can be `None`. There is another example where we deal with this in `analyze_ref_expr()`. The crash doesn't happen on the old analyzer because it doesn't add the broken overload to the symbol table.

I solve the crash by generating a dummy `Overloaded()` type in this case. An alternative solution would be to re-analyze a broken overload as independent (re-)definitions, but I think this is more error-prone to keep an overload in AST, but a bunch of functions in symbol table.

Using this opportunity I also remove a redundant `add_symbol()` call (we always add the overload before even processing it).

Note that the behavior of the new analyzer is still different from the old one, but IMO it is more consistent in some sense.

### DIFF
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -2181,6 +2181,7 @@ def function_type(func: mypy.nodes.FuncBase, fallback: Instance) -> FunctionLike
             return callable_type(func, fallback)
         else:
             # Broken overloads can have self.type set to None.
+            # TODO: should we instead always set the type in semantic analyzer?
             assert isinstance(func, mypy.nodes.OverloadedFuncDef)
             any_type = AnyType(TypeOfAny.from_error)
             dummy = CallableType([any_type, any_type],
@@ -2188,6 +2189,8 @@ def function_type(func: mypy.nodes.FuncBase, fallback: Instance) -> FunctionLike
                                  [None, None], any_type,
                                  fallback,
                                  line=func.line, is_ellipsis_args=True)
+            # Return an Overloaded, because some callers may expect that
+            # an OverloadedFuncDef has an Overloaded type.
             return Overloaded([dummy])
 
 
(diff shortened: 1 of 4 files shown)


mypy/types.py
```diff
@@ -2177,9 +2177,18 @@ def function_type(func: mypy.nodes.FuncBase, fallback: Instance) -> FunctionLike
...
+        else:
+            # Broken overloads can have self.type set to None.
+            assert isinstance(func, mypy.nodes.OverloadedFuncDef)
+            any_type = AnyType(TypeOfAny.from_error)
+            dummy = CallableType([any_type, any_type],
+                                 [ARG_STAR, ARG_STAR2],
+                                 [None, None], any_type,
+                                 fallback,
+                                 line=func.line, is_ellipsis_args=True)
+            return Overloaded([dummy])
```
JukkaL: Why can't we just return `dummy`? Add a comment explaining why.

mypy/types.py
```diff
@@ -2177,9 +2177,18 @@ def function_type(func: mypy.nodes.FuncBase, fallback: Instance) -> FunctionLike
...
     else:
         # Implicit type signature with dynamic types.
-        # Overloaded functions always have a signature, so func must be an ordinary function.
-        assert isinstance(func, mypy.nodes.FuncItem), str(func)
-        return callable_type(func, fallback)
+        if isinstance(func, mypy.nodes.FuncItem):
+            return callable_type(func, fallback)
+        else:
+            # Broken overloads can have self.type set to None.
+            assert isinstance(func, mypy.nodes.OverloadedFuncDef)
```
JukkaL: I wonder if it would be better to always set the overload type in the semantic analyzer. No need to do anything about this now, but consider adding a TODO comment.

PR #7184, opened 2019-07-09
Support selecting TypedDicts from unions

ilevkivskyi:
It is a relatively common pattern to narrow down typed dicts from unions with non-dict types using `isinstance(x, dict)`. Currently mypy infers `Dict[Any, Any]` after such checks which is suboptimal.

I propose to special-case this in `narrow_declared_type()` and `restrict_subtype_away()`. Using this opportunity I factored out special cases from the latter in a separate helper function.

Using this opportunity I also fix an old type erasure bug in `isinstance()` checks (type should be erased after mapping to supertype, not before).

### DIFF
--- a/mypy/meet.py
+++ b/mypy/meet.py
@@ -54,6 +54,12 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
         return TypeType.make_normalized(narrow_declared_type(declared.item, narrowed.item))
     elif isinstance(declared, (Instance, TupleType, TypeType, LiteralType)):
         return meet_types(declared, narrowed)
+    elif isinstance(declared, TypedDictType) and isinstance(narrowed, Instance):
+        # Special case useful for selecting TypedDicts from unions using isinstance(x, dict).
+        if (narrowed.type.fullname() == 'builtins.dict' and
+                all(isinstance(t, AnyType) for t in narrowed.args)):
+            return declared
+        return narrowed
     return narrowed
 
 
--- a/mypy/meet.py
+++ b/mypy/meet.py
@@ -59,7 +59,7 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
         if (narrowed.type.fullname() == 'builtins.dict' and
                 all(isinstance(t, AnyType) for t in narrowed.args)):
             return declared
-        return narrowed
+        return meet_types(declared, narrowed)
     return narrowed
 
 
@@ -484,6 +484,8 @@ class TypeMeetVisitor(TypeVisitor[Type]):
             return meet_types(t, self.s)
         elif isinstance(self.s, LiteralType):
             return meet_types(t, self.s)
+        elif isinstance(self.s, TypedDictType):
+            return meet_types(t, self.s)
         return self.default(self.s)
 
     def visit_callable_type(self, t: CallableType) -> Type:
@@ -561,6 +563,8 @@ class TypeMeetVisitor(TypeVisitor[Type]):
             fallback = self.s.create_anonymous_fallback(value_type=mapping_value_type)
             required_keys = t.required_keys | self.s.required_keys
             return TypedDictType(items, required_keys, fallback)
+        elif isinstance(self.s, Instance) and is_subtype(t, self.s):
+            return t
         else:
             return self.default(self.s)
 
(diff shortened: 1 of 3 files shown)


mypy/meet.py
```diff
@@ -54,6 +54,12 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
         return TypeType.make_normalized(narrow_declared_type(declared.item, narrowed.item))
     elif isinstance(declared, (Instance, TupleType, TypeType, LiteralType)):
         return meet_types(declared, narrowed)
+    elif isinstance(declared, TypedDictType) and isinstance(narrowed, Instance):
+        # Special case useful for selecting TypedDicts from unions using isinstance(x, dict).
+        if (narrowed.type.fullname() == 'builtins.dict' and
+                all(isinstance(t, AnyType) for t in narrowed.args)):
+            return declared
+        return narrowed
```
JukkaL: Hmm, I'm not sure this is the best thing to do, but it's consistent with what we did previously. One option would be to call recursively and narrow the fallback type instead. It's okay to leave like this for now, but maybe at least add a TODO comment.

mypy/meet.py
```diff
@@ -54,6 +54,12 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
         return TypeType.make_normalized(narrow_declared_type(declared.item, narrowed.item))
     elif isinstance(declared, (Instance, TupleType, TypeType, LiteralType)):
         return meet_types(declared, narrowed)
+    elif isinstance(declared, TypedDictType) and isinstance(narrowed, Instance):
+        # Special case useful for selecting TypedDicts from unions using isinstance(x, dict).
+        if (narrowed.type.fullname() == 'builtins.dict' and
```
JukkaL: What about `Mapping` as the narrowed type? Consider at least adding a TODO comment.

mypy/subtypes.py
```diff
@@ -1007,32 +1007,41 @@ def unify_generic_callable(type: CallableType, target: CallableType,
...
+    supertype = erase_type(supertype)
+    if is_proper_subtype(erase_type(item), supertype, ignore_promotions=ignore_promotions):
+        return True
+    if isinstance(supertype, Instance) and supertype.type.is_protocol:
+        # TODO: Implement more robust support for runtime isinstance() checks, see issue #3827.
+        if is_proper_subtype(item, supertype, ignore_promotions=ignore_promotions):
+            return True
+    if isinstance(item, TypedDictType) and isinstance(supertype, Instance):
+        # Special case useful for selecting TypedDicts from unions using isinstance(x, dict).
+        if supertype.type.fullname() == 'builtins.dict':
```
JukkaL: What about `Mapping` (and other ABCs)? Maybe at least add a TODO item.

mypy/subtypes.py
```diff
@@ -1007,32 +1007,41 @@ def unify_generic_callable(type: CallableType, target: CallableType,
...
         return t
 
 
+def covers_at_runtime(item: Type, supertype: Type, ignore_promotions: bool) -> bool:
+    """Will isinstance(item, supertype) always return True at runtime?"""
+    # Since runtime type checks will ignore type arguments, erase the types.
+    supertype = erase_type(supertype)
+    if is_proper_subtype(erase_type(item), supertype, ignore_promotions=ignore_promotions):
+        return True
+    if isinstance(supertype, Instance) and supertype.type.is_protocol:
```
JukkaL: Why do we need the `is_protocol` check? It wasn't there previously.

ilevkivskyi: This whole branch was initially added as a crutch to support `isinstance()` with protocols (see the TODO I moved). The point is that `erase_type()` only erases type arguments but not attribute types. (Well, the check is still unsafe, this is why we have the issue).

Also for non-protocol classes this branch should be a no-op (unless there is a bug in `is_proper_subtype()`).

mypy/meet.py
ilevkivskyi: I can try adding the recursive call unless this will cause troubles in internal code bases.

mypy/meet.py
ilevkivskyi: It looks like ABCs (here and in your other comment below) should be supported by normal mechanisms, but there is an unrelated bug that blocks this. I have a fix, but I want to check it against internal code bases again.

PR #7188, opened 2019-07-10
Honor return type of `__new__`

### LINKED ISSUE #1020: Honor return type of `__new__`
Currently the return type of `A(...)` is implicitly `A` even if `A.__new__` return something else. If the return type is `Any`, maybe `A(...)` should have type `Any`, and if it's a subclass, we should also use that. Currently this is a little difficult to implement due to implementation limitations (we derive the type object identity from the return type of a type object). Maybe we should introduce a new attribute to `Callable` for this purpose.

msullivan:
This basically follows the approach Jukka laid out in #1020 four years ago:
 * If the return type is Any, ignore that and keep the class type as
   the return type
 * Otherwise respect `__new__`'s return type
 * Produce an error if the return type is not a subtype of the class.

The main motivation for me in implementing this is to support
overloading `__new__` in order to select type variable arguments,
which will be useful for subprocess.Popen.

Fixes #1020.


mypy/checkmember.py
```diff
@@ -868,25 +872,32 @@ def type_object_type_from_function(signature: FunctionLike,
...
 def class_callable(init_type: CallableType, info: TypeInfo, type_type: Instance,
-                   special_sig: Optional[str]) -> CallableType:
+                   special_sig: Optional[str],
+                   is_new: bool = False) -> CallableType:
     """Create a type object type based on the signature of __init__."""
     variables = []  # type: List[TypeVarDef]
     variables.extend(info.defn.type_vars)
     variables.extend(init_type.variables)
 
+    is_new = True
```
ilevkivskyi: This looks suspicious for two reasons:
* You unconditionally override an argument with an opposite default.
* Just below it appears in an `if`

msullivan: Oh hm. That was testing code that survived because it works without it, since NoneTyp won't trigger the isinstance check...

PR #7195, opened 2019-07-11
Reject fake TypeInfos for TypedDict fallbacks

ilevkivskyi:
This avoids possible crashes during calculations of meets and joins on a cold update. 

We do this by rejecting a fake ad-hoc `TypeInfo` created by `fixup.py` if a file containing the typed dict fallback was deleted, and instead using the generic `typing._TypedDict` fallback.

### DIFF
--- a/mypy/fixup.py
+++ b/mypy/fixup.py
@@ -209,6 +209,12 @@ class TypeFixer(TypeVisitor[None]):
             for it in tdt.items.values():
                 it.accept(self)
         if tdt.fallback is not None:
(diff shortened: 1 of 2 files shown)


ilevkivskyi:
@JukkaL if you think this is relatively safe, then we can merge this in release branch, otherwise I think it is not critical because this is kind of an old issue (see e.g. the second test).

ilevkivskyi:
I played more with the named tuples and they seem safe (mostly because fallbacks are computed on the fly, but also several places are guarded with `if foo.tuple_type is not None`).

PR #7210, opened 2019-07-14
Add support for future.utils.with_metaclass

### LINKED ISSUE #7193: Add support for future.utils.with_metaclass
I'm working on a codebase that uses [`future.utils.with_metaclass`](https://github.com/PythonCharmers/python-future/blob/master/src/future/utils/__init__.py#L109). 
 The interface is the same as [`six.with_metaclass`](https://github.com/benjaminp/six/blob/master/six.py#L820) so perhaps this is as simple as updating `semanal` to add `'future.utils.with_metaclass'` everywhere that `'six.with_metaclass'` is present.

chadrik:
This util function has the same signature and basic behavior as `six.with_metaclass`. 

Fixes #7193.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1541,6 +1541,8 @@ class SemanticAnalyzer(NodeVisitor[None],
         * __metaclass__ attribute in Python 2
         * six.with_metaclass(M, B1, B2, ...)
         * @six.add_metaclass(M)
+        * future.utils.with_metaclass(M, B1, B2, ...)
+        * past.utils.with_metaclass(M, B1, B2, ...)
         """
 
         # Look for "__metaclass__ = <metaclass>" in Python 2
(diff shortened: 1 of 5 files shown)


chadrik:
> A test might be good, but otherwise LGTM.

Ah yeah, I missed the tests for `six.with_metaclass`.  I'll just copy those.

chadrik:
So this is going to require stubs for `future`, which is going to be pretty complex, since there are a lot of very ambiguous cases to solve for, like backports of python3-compatible builtins like `dict` and `list`.  I'll check back when it's ready.

JelleZijlstra:
For the purposes of mypy tests, you can probably put a basic stub in mypy/test-data somewhere with just what you need.

ilevkivskyi:
There is also a merge conflict now.

chadrik:
No worries. It’s a very simple PR. I’ve just about got the stubs ready for
typeshed, so I’ll be revisiting this soon.

chadrik:
Added tests.  This should be all ready to go!

PR #7220, opened 2019-07-16
For --disallow-any-generic, print the name of the type

msullivan:
This makes the message easier to understand for complex types.  Only
do this for the new semantic analyzer, since the old analyzer doesn't
have the information available to do it.

### DIFF
--- a/test-data/unit/check-flags.test
+++ b/test-data/unit/check-flags.test
@@ -1168,6 +1168,6 @@ main:2: error: Module 'other_module_2' has no attribute 'a'
 from typing import List
 
 A = List  # OK
-B = List[A]  # E:10: Missing type parameters for generic type "list"
+B = List[A]  # E:10: Missing type parameters for generic type "A"
 x: A  # E:4: Missing type parameters for generic type "A"
 [builtins fixtures/list.pyi]
--- a/test-data/unit/check-flags.test
+++ b/test-data/unit/check-flags.test
@@ -1105,7 +1105,7 @@ from typing import TypeVar, Callable
 T = TypeVar('T')
 C = Callable[[], T]
 
-def f(c: C):  # E: Missing type parameters for generic type "Callable[[], T?]"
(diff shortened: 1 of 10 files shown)


JukkaL:
This is helpful especially for beginners that don't understand generic types very well yet. A test case is failing, otherwise looks good.

test-data/unit/check-flags.test
```diff
@@ -1099,26 +1099,26 @@ def f(c: C) -> None:
...
 [case testCheckDisallowAnyGenericsAnyGeneric]
-# flags: --disallow-any-generics
+# flags: --new-semantic-analyzer --disallow-any-generics
 from typing import TypeVar, Callable
 
 T = TypeVar('T')
 C = Callable[[], T]
 
-def f(c: C):  # E: Missing type parameters for generic type
+def f(c: C):  # E: Missing type parameters for generic type "Callable[[], T?]"
```
ilevkivskyi: A tiny bit of bikeshedding: it would be better to show original alias name (if available at this point) instead of the expansion. But this is totally optional, since this is already much better than the current state.

test-data/unit/check-flags.test
```diff
@@ -1164,10 +1164,10 @@ implicit_reexport = False
...
 [case testImplicitAnyOKForNoArgs]
-# flags: --disallow-any-generics --show-column-numbers
+# flags: --new-semantic-analyzer --disallow-any-generics --show-column-numbers
 from typing import List
 
 A = List  # OK
-B = List[A]  # E:10: Missing type parameters for generic type
-x: A  # E:4: Missing type parameters for generic type
+B = List[A]  # E:10: Missing type parameters for generic type "list"
+x: A  # E:4: Missing type parameters for generic type "A"
```
ilevkivskyi: Here I would ideally expect "A" for both errors, otherwise the first error is a bit confusing. But again, not really important.

msullivan: Indeed, which is half of why that test was failing

Issue #7209, opened 2019-07-14
Crashes when @overload is used but not defined


xuhdev:
Please provide more information to help us understand the issue:

* Are you reporting a bug, or opening a feature request?

Bug.

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.

```python
import typing

@overload
def func(x: int) -> int: ...

def func(x):
    return x
```

* What is the actual behavior/output?

Crashes.

* What is the behavior/output you expect?

Report `overload` is not defined.

* What are the versions of mypy and Python you are using?

 0.720+dev.dc6da54f50d5443fe46308046719043eba03c84d (master at the time of report)

* What are the mypy flags you are using? (For example --strict-optional)

None.

* If mypy crashed with a traceback, please paste
  the full traceback below.
```

Traceback (most recent call last):
  File "/home/hong/src/mypy/.venv/bin/mypy", line 11, in <module>
    load_entry_point('mypy', 'console_scripts', 'mypy')()
  File "/home/hong/src/mypy/mypy/__main__.py", line 8, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "/home/hong/src/mypy/mypy/main.py", line 83, in main
    res = build.build(sources, options, None, flush_errors, fscache, stdout, stderr)
  File "/home/hong/src/mypy/mypy/build.py", line 164, in build
    result = _build(sources, options, alt_lib_path, flush_errors, fscache, stdout, stderr)                                                                                          
  File "/home/hong/src/mypy/mypy/build.py", line 224, in _build
    graph = dispatch(sources, manager, stdout)
  File "/home/hong/src/mypy/mypy/build.py", line 2579, in dispatch
    process_graph(graph, manager)
  File "/home/hong/src/mypy/mypy/build.py", line 2892, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/home/hong/src/mypy/mypy/build.py", line 2990, in process_stale_scc
    semantic_analysis_for_scc(graph, scc, manager.errors)
  File "/home/hong/src/mypy/mypy/newsemanal/semanal_main.py", line 78, in semantic_analysis_for_scc                                                                                 
    process_functions(graph, scc, patches)
  File "/home/hong/src/mypy/mypy/newsemanal/semanal_main.py", line 238, in process_functions                                                                                        
    patches)
[...]

PR #7223, opened 2019-07-16
Fix crash when broken overload appears in stub file

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7209

The fix is to never create an "empty" overload that has neither items nor implementation.

### DIFF
--- a/test-data/unit/check-overloading.test
+++ b/test-data/unit/check-overloading.test
@@ -5031,18 +5031,30 @@ def lol(x: int, y: int) -> int:
 
 [case testVeryBrokenOverload]
 import lib
+reveal_type(lib.func)
 [file lib.pyi]
-@overload  # E: Name 'overload' is not defined
+@overload
 def func(x: int) -> int: ...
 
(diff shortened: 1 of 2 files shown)


test-data/unit/check-overloading.test
```diff
@@ -5028,3 +5028,21 @@ def asdf() -> None:
 @dec
 def lol(x: int, y: int) -> int:
     pass
+
+[case testVeryBrokenOverload]
+import lib
```
JukkaL: Maybe also reveal the type of `func`.

test-data/unit/check-overloading.test
```diff
@@ -5028,3 +5028,21 @@ def asdf() -> None:
...
+[file lib.pyi]
+@overload  # E: Name 'overload' is not defined
+def func(x: int) -> int: ...
+
+def func(x):  # E: Name 'func' already defined on line 1
+    return x
+
+[case testVeryBrokenOverload2]
+import lib
```
JukkaL: Like above, reveal the type of `func` here?

PR #7226, opened 2019-07-16
Make type formatting use fullnames whenever there is ambiguity

msullivan:
When pretty printing types, find all groups of distinct appearing
instances that have the same short name and format them using their
fullnames.

This should eliminate error messages such as `got "List[A]", expected "Sequence[A]"`
(and a bunch of even less tricky ones we were messing up...).

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -13,7 +13,6 @@ from collections import OrderedDict
 import re
 import difflib
 from textwrap import dedent
-import itertools
 
 from typing import cast, List, Dict, Any, Sequence, Iterable, Tuple, Set, Optional, Union
 from typing_extensions import Final
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -24,8 +24,8 @@ from mypy.types import (
     Type, CallableType, Instance, TypeVarType, TupleType, TypedDictType, LiteralType,
     UnionType, NoneType, AnyType, Overloaded, FunctionLike, DeletedType, TypeType,
     UninhabitedType, TypeOfAny, ForwardRef, UnboundType, PartialType,
-    TypeQuery,
 )
+from mypy.typetraverser import TypeTraverserVisitor
 from mypy.nodes import (
     TypeInfo, Context, MypyFile, op_methods, FuncDef, reverse_builtin_aliases,
     ARG_POS, ARG_OPT, ARG_NAMED, ARG_NAMED_OPT, ARG_STAR, ARG_STAR2,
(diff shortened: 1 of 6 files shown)


test-data/unit/check-basic.test
```diff
@@ -390,3 +390,61 @@ y = x # E: Incompatible types in assignment (expression has type "Dict[str, int]
...
+    return A()  # E: Incompatible return value type (got "A", expected "Tuple[A, A]")
+
+def types() -> Sequence[Type[A]]:
+    x = [a.A]
+    return x  # E: Incompatible return value type (got "List[Type[a.A]]", expected "Sequence[Type[b.A]]")
+
+a = (a.A(), A())
+a.x  # E: "Tuple[a.A, b.A]" has no attribute "x"
+
+[builtins fixtures/list.pyi]
```
gvanrossum: I didn't read the whole diff yet, but seeing this set of tests makes me wonder, what if one of the types is `a.A` (i.e. not the locally defined `A`) and the other is, say, `int`? Will it correctly show `a.A`, or might it show just `A`? (Which would be confusing because in scope, `A` refers to `b.A`.)

msullivan: We would just show `A`, yeah. This is a bummer but fixing it would be much more invasive than this diff, since type formatting will need to be much more contextual.

mypy/messages.py
```diff
@@ -1396,27 +1385,109 @@ def format_type_bare(typ: Type, verbosity: int = 0) -> str:
...
+    def visit_literal_type(self, t: LiteralType) -> List[Instance]:
+        return self.query_types([t.fallback])
+
+    def visit_tuple_type(self, t: TupleType) -> List[Instance]:
+        return self.query_types(t.items + [t.partial_fallback])
+
+    def visit_typeddict_type(self, t: TypedDictType) -> List[Instance]:
+        return self.query_types(list(t.items.values()) + [t.fallback])
+
+    def query_types(self, types: Iterable[Type]) -> List[Instance]:
```
JukkaL: Overriding `query_types` makes this hard to reason about. It took me a while to figure out why this works at all. This also seems to remove the cyclic type avoidance in the base class. I'd suggest not overriding this (as it's not really designed to be overridden) but to add additional visit methods for `Instance` and `Callable` at least (there may be others). You could also simplify `collect_all_instances` to just use `accept`, which is the normal entry point, instead of calling `query_types`, which is meant to be an internal helper.

mypy/messages.py
```diff
@@ -1396,27 +1385,109 @@ def format_type_bare(typ: Type, verbosity: int = 0) -> str:
...
+def format_type(typ: Type, verbosity: int = 0) -> str:
+    """
+    Convert a type to a relatively short string suitable for error messages.
+
+    `verbosity` is a coarse grained control on the verbosity of the type
+
+    This function returns a string appropriate for unmodified use in error
+    messages; this means that it will be quoted in most cases.  If
+    modification of the formatted string is required, callers should use
+    .format_bare.
```
JukkaL: Should `.format_bare` be replaced with `format_type_bare`?

mypy/messages.py
```diff
@@ -1396,27 +1385,109 @@ def format_type_bare(typ: Type, verbosity: int = 0) -> str:
...
+                     verbosity: int = 0,
+                     fullnames: Optional[Set[str]] = None) -> str:
+    """
+    Convert a type to a relatively short string suitable for error messages.
+
+    `verbosity` is a coarse grained control on the verbosity of the type
+    `fullnames` specifies a set of names that should be printed in full
+
+    This function will return an unquoted string.  If a caller doesn't need to
+    perform post-processing on the string output, .format should be used
```
JukkaL: `.format` -> `format_type`

mypy/messages.py
```diff
@@ -1396,27 +1385,109 @@ def format_type_bare(typ: Type, verbosity: int = 0) -> str:
...
+    instead.  (The caller may want to use .quote_type_string after
```
JukkaL: `.quote_type_string` -> `quote_type_string`

mypy/messages.py
```diff
@@ -1396,27 +1385,109 @@ def format_type_bare(typ: Type, verbosity: int = 0) -> str:
...
+
 def format_type_distinctly(type1: Type, type2: Type, bare: bool = False) -> Tuple[str, str]:
     """Jointly format a pair of types to distinct strings.
 
-    Increase the verbosity of the type strings until they become distinct.
+    Increase the verbosity of the type strings until they become distinct
+    while also requiring that distinct types with the same short name are
+    formatted distinctly.
 
     By default, the returned strings are created using .format() and will be
```
JukkaL: `.format` -> `format_type`?

mypy/type_visitor.py
```diff
@@ -316,7 +316,7 @@ def visit_overloaded(self, t: Overloaded) -> T:
         return self.query_types(t.items())
 
     def visit_type_type(self, t: TypeType) -> T:
-        return t.item.accept(self)
+        return self.query_types([t.item])
```
JukkaL: If you implement the change I requested above, this change won't be necessary.

mypy/messages.py
msullivan: I'm going to rework it to use `TypeTraverserVisitor`

PR #7232, opened 2019-07-17
Analyze super() expressions more precisely

### LINKED ISSUE #5794: Too many arguments for "__setattr__" of "object" on valid code
This code is a reasonable pattern for initializing "immutable" objects:

```python
class Foo:
    def __new__(cls, bar: int):
        ret = super().__new__(cls)
        super().__setattr__(ret, "bar", bar)
        return ret
```

But mypy complains

    foo.py:4: error: Too many arguments for "__setattr__" of "object"

tavianator:
The previous implementation assumed that all super(T, x) expressions
were for instances of the current class.  This new implementation uses
the precise types of T and x to do a more accurate analysis.
    
Fixes #5794

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3108,111 +3108,140 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
     def visit_super_expr(self, e: SuperExpr) -> Type:
         """Type check a super expression (non-lvalue)."""
-        self.check_super_arguments(e)
-        t = self.analyze_super(e, False)
-        return t
 
-    def check_super_arguments(self, e: SuperExpr) -> None:
-        """Check arguments in a super(...) call."""
-        if ARG_STAR in e.call.arg_kinds:
+        # We have an expression like super(T, var).member
+
+        # First compute the types of T and var
+        types = self._super_arg_types(e)
+        if isinstance(types, tuple):
+            type_type, instance_type = types
+        else:
+            return types
+
+        # Now get the MRO
+        type_info = type_info_from_type(type_type)
+        if type_info is None:
+            self.chk.fail(message_registry.UNSUPPORTED_ARG_1_FOR_SUPER, e)
+            return AnyType(TypeOfAny.from_error)
+
+        instance_info = type_info_from_type(instance_type)
+        if instance_info is None:
+            self.chk.fail(message_registry.UNSUPPORTED_ARG_2_FOR_SUPER, e)
+            return AnyType(TypeOfAny.from_error)
+
+        mro = instance_info.mro
+
+        # The base is the first MRO entry *after* type_info that has a member
+        # with the right name
+        try:
+            index = mro.index(type_info)
+        except ValueError:
+            self.chk.fail(message_registry.SUPER_ARG_2_NOT_INSTANCE_OF_ARG_1, e)
+            return AnyType(TypeOfAny.from_error)
+
+        for base in mro[index+1:]:
+            if e.name in base.names or base == mro[-1]:
+                if e.info and e.info.fallback_to_any and base == mro[-1]:
+                    # There's an undefined base class, and we're at the end of the
+                    # chain.  That's not an error.
+                    return AnyType(TypeOfAny.special_form)
+
+                return analyze_member_access(name=e.name,
+                                             typ=instance_type,
+                                             is_lvalue=False,
+                                             is_super=True,
+                                             is_operator=False,
+                                             original_type=instance_type,
+                                             override_info=base,
+                                             context=e,
+                                             msg=self.msg,
+                                             chk=self.chk,
+                                             in_literal_context=self.is_literal_context())
+
+        assert False, 'unreachable'
+
+    def _super_arg_types(self, e: SuperExpr) -> Union[Type, Tuple[Type, Type]]:
+        """
+        Computes the types of the type and instance expressions in super(T, instance), or the
+        implicit ones for zero-argument super() expressions.  Returns a single type for the whole
+        super expression when possible (for errors, anys), otherwise the pair of computed types.
+        """
+
+        if not self.chk.in_checked_function():
+            return AnyType(TypeOfAny.unannotated)
+        elif len(e.call.args) == 0:
+            if self.chk.options.python_version[0] == 2:
+                self.chk.fail(message_registry.TOO_FEW_ARGS_FOR_SUPER, e)
+                return AnyType(TypeOfAny.from_error)
+            elif not e.info:
+                # This has already been reported by the semantic analyzer.
+                return AnyType(TypeOfAny.from_error)
+            elif self.chk.scope.active_class():
+                self.chk.fail(message_registry.SUPER_OUTSIDE_OF_METHOD_NOT_SUPPORTED, e)
+                return AnyType(TypeOfAny.from_error)
+
+            # Zero-argument super() is like super(<current class>, <self>)
+            current_type = fill_typevars(e.info)
+            type_type = TypeType(current_type)  # type: Type
+
+            # Use the type of the self argument, in case it was annotated
+            method = self.chk.scope.top_function()
+            assert method is not None
+            if method.arguments:
+                instance_type = method.arguments[0].variable.type or current_type  # type: Type
+            else:
+                self.chk.fail(message_registry.SUPER_ENCLOSING_POSITIONAL_ARGS_REQUIRED, e)
+                return AnyType(TypeOfAny.from_error)
+        elif ARG_STAR in e.call.arg_kinds:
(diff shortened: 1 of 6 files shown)


test-data/unit/check-super.test
```diff
@@ -204,6 +204,17 @@ class C(B, Generic[T, S]):
...
 class D(C): pass
 
+[case testSuperInClassMethod]
+class A:
+    def f(self, i: int) -> None: pass
+
+class B(A):
+    @classmethod
+    def g(cls, i: int) -> None:
+        super().f(B(), i)
```
JukkaL: Can you also test invalid calls to make sure the arguments are checked properly. Another idea would be to have a narrower argument type in the base class method, and to make sure the correct signature is being used.

mypy/checkexpr.py
```diff
@@ -3199,7 +3199,7 @@ def analyze_super(self, e: SuperExpr, is_lvalue: bool) -> Type:
                         return AnyType(TypeOfAny.from_error)
                     declared_self = args[0].variable.type or fill_typevars(e.info)
                     return analyze_member_access(name=e.name,
-                                                 typ=fill_typevars(e.info),
+                                                 typ=declared_self,
```
JukkaL: I'm not sure about this. This looks like it might look up the signature from the current class, not the base class?

test-data/unit/check-super.test
JukkaL: The narrower arguments would only be relevant when you have a method with the same name in the base class.

mypy/checkexpr.py
tavianator: I think that's what `override_info` is for below, which makes the checkmember machinery use that instead of `typ` as the context to look up members.

Actually I think `override_info` could be removed, and instead just pass the base type as `typ`.  I'll try that out.

tavianator:
I've updated the PR to do much more complete typing of `super()` expressions.

mypy/checkexpr.py
```diff
@@ -3108,112 +3108,135 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+                # Needed for named tuples and other Tuple[...] subclasses.
+                simple_type = tuple_fallback(simple_type)
+            mro = simple_type.type.mro
+        elif isinstance(instance_type, FunctionLike) and instance_type.is_type_obj():
+            mro = instance_type.type_object().mro
+        elif isinstance(instance_type, TypeType):
+            item = instance_type.item
+            if isinstance(item, AnyType):
+                # Could be anything.
+                # XXX: TODO: Same as elif isinstance(instance_type, AnyType) below
```
tavianator: Oops, didn't do this yet

JukkaL: No worries, just let us know when this ready for review.

tavianator: Should be ready now, thanks!

mypy/checkexpr.py
tavianator: I looked into this in more detail, the `override_info` is necessary to handle things like generics.  If we just pass the base type as `typ` it throws away too much information.

test-data/unit/check-super.test
tavianator: Done!

tavianator:
@JukkaL Good call, I split the implementation up a bit so it should be easier to follow.

tavianator:
The CI failure is a socket timeout in mypy/test/testipc.py which I assume is unrelated to this

mypy/checkexpr.py
```diff
@@ -3108,112 +3108,134 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+        if len(e.call.args) == 0:
+            if self.chk.options.python_version[0] == 2:
+                self.chk.fail(message_registry.TOO_FEW_ARGS_FOR_SUPER, e)
+                return AnyType(TypeOfAny.from_error)
+            elif self.chk.scope.active_class():
+                self.chk.fail(message_registry.SUPER_OUTSIDE_OF_METHOD_NOT_SUPPORTED, e)
+                return AnyType(TypeOfAny.from_error)
+
+            # Use the type of the self argument, in case it was annotated
+            method = self.chk.scope.top_function()
```
JukkaL: What if this was a nested function within a method?

mypy/checkexpr.py
```diff
@@ -3108,112 +3108,134 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+
+        # Imprecisely assume that the instance is of the current class
+        if isinstance(instance_type, AnyType):
+            instance_type = current_instance
+        elif isinstance(instance_type, TypeType) and isinstance(instance_type.item, AnyType):
+            instance_type = current_type
+
+        return type_type, instance_type
+
+    def _type_info(self, typ: Type) -> Optional[TypeInfo]:
```
JukkaL: Style nit: since this doesn't use `self` in the body, this could be changed to a module-level function. If you do this, the name would have to be changed, since we don't generally use `_` prefix at module level. Suggestion: `type_info_from_type`.

mypy/checkexpr.py
```diff
@@ -3108,112 +3108,134 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+        except ValueError:
+            self.chk.fail(message_registry.SUPER_ARG_2_NOT_INSTANCE_OF_ARG_1, e)
+            return AnyType(TypeOfAny.from_error)
+
+        for base in mro[index+1:]:
+            if e.name in base.names or base == mro[-1]:
+                if e.info and e.info.fallback_to_any and base == mro[-1]:
+                    # There's an undefined base class, and we're at the end of the
+                    # chain.  That's not an error.
+                    return AnyType(TypeOfAny.special_form)
```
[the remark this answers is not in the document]
tavianator: Probably yeah, this is just what the old code did.  Is `source_any=mro[-1]` fine in this case?

mypy/checkexpr.py
```diff
@@ -3108,112 +3108,134 @@ def infer_lambda_type_using_context(self, e: LambdaExpr) -> Tuple[Optional[Calla
...
+                                             original_type=instance_type,
+                                             override_info=base,
+                                             context=e,
+                                             msg=self.msg,
+                                             chk=self.chk,
+                                             in_literal_context=self.is_literal_context())
+
+        assert False, 'unreachable'
+
+    def _super_arg_types(self, e: SuperExpr) -> Union[Type, Tuple[Type, Type]]:
```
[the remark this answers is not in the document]
tavianator: Good call.  It returns a single `Type` when that should be the overall return type of the super expression (e.g. errors, anys), otherwise it returns the type_type, instance_type pair.  Maybe there's a nicer way to do that?

mypy/checkexpr.py
tavianator: Also copied from the old code, but I wouldn't be surprised if this is wrong in that case.  I guess we need to walk the scopes to find the current method?

mypy/checkexpr.py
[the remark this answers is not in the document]
tavianator: Looks like not, it's just `<TypeInfo builtins.object>` in that case.  I guess that's why it's a `special_form`.

mypy/checkexpr.py
tavianator: Oh actually I think this code is right.  The Python interpreter uses the innermost function even if it's not a method.  Something like

```
class A:
    def f(self): pass

class B(A):
    def f(self):
        def g():
            super().f()
        g()

B().f()
```

gives

```
RuntimeError: super(): no arguments
```

`testSuperInMethodWithNoArguments` already checks for this.

mypy/checkexpr.py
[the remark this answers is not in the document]
JukkaL: Let's keep as it is since that's how it was previously.

tavianator:
Ah okay, I'll keep that in mind for next time.  Thanks for merging!

msullivan:
@tavianator This rejects calls where the instance is of type `Any`, which seems wrong. Could you produce a follow-up that fixes that? Thanks!

tavianator:
@msullivan Do you have an example?  There is e.g. the test case

```python
[case testSuperWithAny]
class B:
    def f(self) -> None: pass
class C(B):
    def h(self, x) -> None:
        reveal_type(super(x, x).f) # N: Revealed type is 'def ()'
        reveal_type(super(C, x).f) # N: Revealed type is 'def ()'
        reveal_type(super(C, type(x)).f) # N: Revealed type is 'def (self: __main__.B)'
```

which passes.

msullivan:
I'm going to withdraw my report. The basic case seems to work fine. I tripped over a small number cases in internal code but they had a lot of additional complications and to be honest are all pretty heinous, so I'm not going to worry about it.

PR #7240, opened 2019-07-18
Prevent duplicates in module dependency lists

msullivan:
Currently in some cases we will write out a lot of duplicate entries
in the dependency lists of modules. Refactor dependency management to
prevent this.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2185,7 +2185,7 @@ class State:
             if dep not in self.suppressed_set and dep not in self.manager.missing_modules:
                 self.add_dependency(dep)
                 self.priorities[dep] = PRI_INDIRECT
-            elif dep in self.manager.missing_modules:
+            elif dep not in self.suppressed_set and dep in self.manager.missing_modules:
                 self.suppress_dependency(dep)
 
     def compute_fine_grained_deps(self) -> Dict[str, Set[str]]:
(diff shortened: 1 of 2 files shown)


mypy/build.py
```diff
@@ -2157,11 +2182,11 @@ def _patch_indirect_dependencies(self,
...
                 continue
-            if dep not in self.suppressed and dep not in self.manager.missing_modules:
-                self.dependencies.append(dep)
+            if dep not in self.suppressed_set and dep not in self.manager.missing_modules:
+                self.add_dependency(dep)
                 self.priorities[dep] = PRI_INDIRECT
-            elif dep not in self.suppressed and dep in self.manager.missing_modules:
-                self.suppressed.append(dep)
+            elif dep in self.manager.missing_modules:
+                self.suppress_dependency(dep)
```
ilevkivskyi: TBH it is a bit hard to follow logic here. Why is it OK to remove the `dep not in self.suppressed` from the check for this branch?

msullivan: Because the check is pushed down into `suppress_dependency`; I can add it back if it seems clearer, though

ilevkivskyi: I would keep it (just in case if anything is changed that method). Also I noticed you keep other similar checks, so it will add some consistency (or are these are checks actually needed, for example the one on line 2777, green side).

PR #7254, opened 2019-07-23
Report precision stats for imports and add new precision report

JukkaL:
Previously when reporting per-line type checking precision, imports were 
treated as empty lines, which seems inconsistent. Now treat an import as
precise if the target module exists, and as imprecise if the target module is 
missing from the build.

Also add a new report, `lineprecision`, which is mostly designed to be used 
in tests. The output is more compact and cleaner than the existing 
XML-based reports.

Make it possible to run report generation tests without using full stubs,
since using full stubs is slow.

This is still not quite perfect. If using module-level `__getattr__`, the import
should probably be treated as imprecise. Also, if a missing name is ignored
using `# type: ignore`, we may want to treat that as imprecise. For multi-line
imports, we only report precision for the first line. These would be a bit tricky 
to fix in the report generator so I decided to skip these for now.

### DIFF
--- a/mypy/report.py
+++ b/mypy/report.py
@@ -74,9 +74,13 @@ class Reports:
         self.named_reporters[report_type] = reporter
         return reporter
 
-    def file(self, tree: MypyFile, type_map: Dict[Expression, Type], options: Options) -> None:
+    def file(self,
+             tree: MypyFile,
+             modules: Dict[str, MypyFile],
+             type_map: Dict[Expression, Type],
+             options: Options) -> None:
         for reporter in self.reporters:
-            reporter.on_file(tree, type_map, options)
+            reporter.on_file(tree, modules, type_map, options)
 
     def finish(self) -> None:
         for reporter in self.reporters:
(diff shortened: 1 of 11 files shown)


mypy/report.py
```diff
@@ -108,6 +116,23 @@ def alias_reporter(source_reporter: str, target_reporter: str) -> None:
     reporter_classes[target_reporter] = reporter_classes[source_reporter]
 
 
+def should_skip_path(path: str) -> bool:
+    if stats.is_special_module(path):
+        return True
+    if path.startswith('..'):
+        return True
+    if 'stubs' in path.split('/'):
```
msullivan: Should this be `os.sep`?

mypy/report.py
```diff
@@ -749,6 +788,73 @@ def on_finish(self) -> None:
...
+        visitor = stats.StatisticsVisitor(inferred=True,
+                                          filename=tree.fullname(),
+                                          modules=modules,
+                                          typemap=type_map,
+                                          all_nodes=True)
+        tree.accept(visitor)
+
+        file_info = FileInfo(path, tree._fullname)
+        for lineno, _ in iterate_python_lines(path):
+            print(repr((lineno, _)))
```
msullivan: Stray debug code?

mypy/report.py
```diff
@@ -749,6 +788,73 @@ def on_finish(self) -> None:
...
+            status = visitor.line_map.get(lineno, stats.TYPE_EMPTY)
+            file_info.counts[status] += 1
+
+        self.files.append(file_info)
+
+    def on_finish(self) -> None:
+        output_files = sorted(self.files, key=lambda x: x.module)
+        report_file = os.path.join(self.output_dir, 'lineprecision.txt')
+        width = max(4, max(len(info.module) for info in output_files))
+        fmt = '{:%d}  {:5}  {:7}  {:9}  {:3}  {:5}  {:10}\n' % width
```
msullivan: Maybe generate the width numbers here and below from the list of column headers instead of putting them in as magic numbers?

PR #7256, opened 2019-07-23
Allow forcing mypyc multi-file config in setup.py

msullivan:


msullivan:
This will allow users doing their own builds to set this flag, which is useful for people using gcc.
(I am setting it for some dropbox internal stuff.)

PR #7261, opened 2019-07-24
Improve type precision stats

JukkaL:
This makes various improvements to type precision reporting:

* Consider literals precise in type checked code.
* Consider class headers precise (I'll improve this later).
* Consider pass/break/continue precise in type checked code.
* Consider call imprecise if target function has imprecise argument
  types, and the caller provides one of these arguments.

Some of these are still incomplete (for example, overloaded functions
aren't supported). I'll deal with these issues in follow-up PRs.

### DIFF
--- a/mypy/stats.py
+++ b/mypy/stats.py
@@ -225,7 +225,7 @@ class StatisticsVisitor(TraverserVisitor):
             self.record_call_target_precision(o)
 
     def record_call_target_precision(self, o: CallExpr) -> None:
-        """Record precision of used formal argument types."""
+        """Record precision of formal argument types used in a call."""
         if not self.typemap or o.callee not in self.typemap:
             # Type not availabe.
             return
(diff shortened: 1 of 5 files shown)


mypy/stats.py
```diff
@@ -179,6 +222,35 @@ def visit_call_expr(self, o: CallExpr) -> None:
...
+        if not self.typemap or o.callee not in self.typemap:
+            # Type not availabe.
+            return
+        callee_type = self.typemap[o.callee]
+        if isinstance(callee_type, CallableType):
+            self.record_callable_target_precision(o, callee_type)
+        else:
+            pass  # TODO: Handle overloaded functions, etc.
+
+    def record_callable_target_precision(self, o: CallExpr, callee: CallableType) -> None:
```
msullivan: Document the rationale for how we count this, since it is arguable maybe?

Issue #7239, opened 2019-07-18
Support error codes and ignoring only specific errors


JukkaL:
It would be handy to support error codes to allow ignoring only specific error messages on each line. For example, here we only ignore an error about an incompatible override in class `B`:

```py
class A:
    def f(self) -> None: 
        ...

class B(A):
    def f(self, x: int) -> None:  # type: ignore[override]
        ...
```

If there are any other errors on that line (such as "invalid type"), they will still be reported. Using the existing blanket `# type: ignore` comment risks ignoring too many errors, resulting in false negatives.

For this we need these changes:

1. Make it possible to display error codes with errors (either optionally or by default).
2. Support ignoring specific errors using `# type: ignore[code1, ..., codeN]`.
3. Implement error codes for all common errors (at least).

This is potentially a pretty disruptive change, so we need to be careful about how to do this. Here are some ideas. There is an existing PR that implements some of this (#6472). My proposal differs in various details, but the end goal is the same.

**Format of error codes.** There are various options such as numeric codes, strings derived from messages (as in #6472), and short error identifiers that are chosen manually. I'm leaning towards the final option. I'd like error codes to be pretty short but still understandable, and perhaps even memorable. I'm proposing that most error codes are of the form `foo`, `foo-bar` or `foo-bar-lulz` (one to three words separated by dashes). 

**Granularity of error codes.** I think that not every error message needs a separate code, since this could mean hundreds of different codes. The number of codes should small enough that the list of all codes is easy enough to browse. I'm thinking of maybe having 20-50 distinct error codes in the long term. We don't need a unique error code for every rare message. We can either use a category code (such as `indexing` for various errors related to `x[y]` expressions) or a fallback misc/other category for all unclassified errors. The misc/other category also makes adoption easy -- we can initially have lots of errors in this category, and gradually migrate additional errors to more specific codes.

**Displaying error codes.** We could display error codes like this in mypy output:

```
foo.py:124: error: Some bad stuff going on here [bad-stuff]
```

The rationale is that the error code is *usually* not important, so we shouldn't make it too prominent. This is also one of the reasons why the error codes should be pretty short. This way we could probably enable error codes by default, as they are perhaps not too distracting.

**Implementing error codes.** My idea would be to have error codes totally separate from error messages. Each error reporting call would optionally include the error code (it would default to a generic other/misc category code). Error codes would defined with some metadata in a new error code registry module. The metadata could be used for grouping and help output. Here's a rough idea (this will need some iteration):

```py
[...]

PR #7267, opened 2019-07-26
Add beginnings of error code support

JukkaL:
This PR adds a foundation for error codes and implements a few error
codes. It also adds support for `# type: ignore[code1, ...]` which 
ignores only specific error codes on a line.

Only a few errors include interesting error codes at this point. I'll add 
support for more error codes in additional PRs. Most errors will implicitly 
fall back to a `misc` error code.

Error codes are only shown if `--show-error-codes` is used.

The error codes look like this in mypy output:

```
t.py:3: error: "str" has no attribute "trim"  [attr-defined]
```

Error codes are intended to be short but human-readable. The name
of an error code refers to the check that produces this error. In the above
example we generate a "no attribute" error when we check whether an 
attribute is defined.

Work towards #7239.


mypy/fastparse.py
```diff
@@ -322,6 +345,21 @@ def translate_stmt_list(self,
 
         return res
 
+    def translate_type_comment(self,
```
msullivan: This is a good refactor :+1:

mypy/errorcodes.py
```diff
@@ -0,0 +1,50 @@
...
+
+# All created error codes are implicitly stored in this list.
+all_error_codes = []  # type: List[ErrorCode]
+
+
+class ErrorCode:
+    def __init__(self, code: str, description: str, category: str) -> None:
+        self.code = code
+        self.description = description
+        self.category = category
```
msullivan: What are the plans for `category`?

msullivan:
I'm going to look at the mypyc crash in CI (other two failures are old semantic analyzer)

mypy/messages.py
```diff
@@ -119,12 +127,18 @@ def report(self, msg: str, context: Optional[Context], severity: str,
...
-
-    def fail(self, msg: str, context: Optional[Context], file: Optional[str] = None,
+                               end_line=end_line,
+                               code=code)
+
+    def fail(self,
+             msg: str,
+             context: Optional[Context],
+             *,
+             code: Optional[ErrorCode] = None,
```
msullivan: We should add the `code` argument to the versions of this method in `SemanticAnalyzerCoreInterface` and maybe in `TypeAnalyzerPluginInterface`.

mypy/newsemanal/semanal.py
```diff
@@ -4529,7 +4530,9 @@ def is_local_name(self, name: str) -> bool:
     def fail(self,
              msg: str,
              ctx: Context,
-             serious: bool = False, *,
+             serious: bool = False,
+             *,
+             code: Optional[ErrorCode] = None,
```
msullivan: We should add the code argument to the versions of this method in SemanticAnalyzerCoreInterface and maybe in TypeAnalyzerPluginInterface.

JukkaL: Yes. I'd rather do this only after we've removed the old semantic analyzer, since changing this would require changes in it.

mypy/errorcodes.py
JukkaL: My idea is to use it for grouping the available error codes in the output from `--show-error-codes`, once that's implemented. This way error codes defined by plugins, for example, can be shown in a separate section.

mypy/errorcodes.py
```diff
@@ -0,0 +1,50 @@
...
+RETURN_VALUE = ErrorCode(
+    'return-value', "Check that return value is compatible with signature",
+    'General')  # type: Final
+ASSIGNMENT = ErrorCode(
+    'assignment', "Check that assigned value is compatible with target", 'General')  # type: Final
+
+SYNTAX = ErrorCode(
+    'syntax', "Report syntax errors", 'General')  # type: Final
+
+MISC = ErrorCode(
```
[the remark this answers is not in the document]
JukkaL: My goal is to categorize all common errors so that the misc code should not be needed that often, making the backwards compatibility issue less of a problem. For example, if error codes could cover ~98% of errors, all the remaining errors in the misc category would each be relatively rare, so changing some of them to non-misc codes should not have a big impact.

mypy/fastparse.py
```diff
@@ -173,39 +173,58 @@ def parse(source: Union[str, bytes],
...
             stripped_type = type_comment.split("#", 2)[0].strip()
             err_msg = "{} '{}'".format(TYPE_COMMENT_SYNTAX_ERROR, stripped_type)
             errors.report(line, e.offset, err_msg, blocker=True)
-            return False, None
+            return None, None
         else:
             raise
     else:
-        extra_ignore = TYPE_IGNORE_PATTERN.match(type_comment) is not None
+        extra_ignore = TYPE_IGNORE_PATTERN.match(type_comment)
```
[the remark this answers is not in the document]
JukkaL: If it can be understood as a bug, maybe it can still be fixed? I'm not sure.

mypy/fastparse.py
```diff
@@ -124,7 +124,7 @@ def ast3_parse(source: Union[str, bytes], filename: str, mode: str,
 
 TYPE_COMMENT_SYNTAX_ERROR = 'syntax error in type comment'  # type: Final
 
-TYPE_IGNORE_PATTERN = re.compile(r'[^#]*#\s*type:\s*ignore\s*($|#)')
+TYPE_IGNORE_PATTERN = re.compile(r'[^#]*#\s*type:\s*ignore\s*(\[[^[#]*\]\s*)?($|#)')
```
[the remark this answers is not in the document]
JukkaL: I'm going to make this less permissive in another PR. I'm not quite sure what the behavior should be, but likely something different from what we have here.

mypy/errors.py
```diff
@@ -81,12 +87,23 @@ def __init__(self,
...
+        self.code = code
         self.blocker = blocker
         self.only_once = only_once
         self.origin = origin or (file, line, line)
         self.target = target
 
 
+# Type used internally to represent errors:
+#   (path, line, column, severity, message, code)
+ErrorTuple = Tuple[Optional[str],
```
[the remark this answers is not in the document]
JukkaL: We could do it, but since we never access the items by index, the benefit would be minor, I think, so I'd rather keep it as it is.

mypy/errorcodes.py
JukkaL: I actually meant a separate command such as `--list-error-codes` that shows all the available error codes.

JukkaL:
The build is finally green.

Issue #5903, opened 2018-11-16
Stubgen: Wrong placement of decorators for methods with instance field declarations


hoefling:
Consider the following snippet:

```python
class A:
    @property
    def x(self):
        self.y = 'y'
        return 'x'

class B:
    @property
    def x(self):
        return 'x'

    @x.setter
    def x(self, value):
        self.y = 'y'
```

The stub generated via `stubgen` contains:

```python

# Stubs for tests.m (Python 3.6)
#
# NOTE: This dynamically typed stub was automatically generated by stubgen.

from typing import Any

class A:
    @property
    y: str = ...
    def x(self): ...

class B:
    @property
    def x(self): ...
    @x.setter
    y: str = ...
    def x(self, value: Any) -> None: ...
```

Notice the instance field type declarations are being placed between the decorator and the function declaration lines.

I can reproduce this issue using the latest `mypy` installed from the `master` branch:
```sh
$ pip show mypy | grep Version
Version: 0.650+dev.9459c7544f9adedb509d92e211f300e573cb6844
```

ilevkivskyi:
Oh yes, this is a weird bug, `stubgen` definitely needs more attention (and we are going to devote some time to it next month). Also you are welcome to submit a PR!

BuongiornoTexas:
I think the following is either a symptom of the same bug or is closely related (and had me going in circles for a quite a while):

```python
from typing import Dict

class Circus:

    _db_config: Dict[str, str]

    @property
    def input_file(self) -> str:
        return self._db_config.setdefault("Input", "")

    @input_file.setter
    def input_file(self, value: str) -> None:
        self._db_config["Input"] = value

    @property
    def user_id(self) -> str:
        return self._db_config.setdefault("User", "")

    @user_id.setter
    def user_id(self, value: str) -> None:
        # the next line triggers useful side effects in my code.
        self.player_profile = "" 
        self._db_config["User"] = value

    @property
    def player_profile(self) -> str:
        return self._db_config.setdefault("Player", "")

    @player_profile.setter
    def player_profile(self, value: str) -> None:
        self._db_config["Player"] = value

    def __init__(self) -> None:
        self._db_config = dict()

        # this looks like a non-op, but triggers side effects in my code
        self.user_id = self.user_id

    def sell_cheese(self) -> None:
        self.input_file = 'test.xml'
```

Sorry for the long sample, but I'm not sure which parts are important (I'm pretty  new at python).

Stubgen yields:

```python
[...]

PR #7269, opened 2019-07-27
Fix 5903: wrong placement of decorators for methods with instance field declarations

gantsevdenis:
pretty old issue https://github.com/python/mypy/issues/5903, not sure if was worth it, but looked simple enough


mypy/stubgen.py
```diff
@@ -401,6 +406,8 @@ def visit_func_def(self, o: FuncDef, is_abstract: bool = False) -> None:
             name = var.name()
             annotated_type = (o.unanalyzed_type.arg_types[i]
                               if isinstance(o.unanalyzed_type, CallableType) else None)
+            # I think the name check is incorrect: there are libraries which
+            #   name their 0th argument other than self/cls
```
msullivan: I'd drop the extra spaces

gantsevdenis:
thank you for your review! I have corrected accordingly

PR #7271, opened 2019-07-27
New analyzer: don't fill ImportAll.imported_names

bluetech:
It is used by aststrip.py, but not by aststripnew.py, so the new
semantic analyzer doesn't need it. Ignoring it makes some future changes
easier.


bluetech:
Sorry for being cryptic. I was trying to understand better how imports flow through the code for #7042. This is one less to worry about.

PR #7273, opened 2019-07-29
Support expansion of all path configuration options

### LINKED ISSUE #7272: Feature: Do expansion on all path options in configuration
Previously issues #6508 and #7259 added home directory and environment variable expansion to the `cache_dir` option in the configuration file. It might make sense to the the same for all the other options also referring to paths, these would include: `files`, `mypy_path`, `custom_typeshed_dir` and `python_executable`.

Some review may be necessary to determine if this is not appropriate for any of these options.

hetmankp:
This closes #7272; all options in the configuration file that deal with paths will be expanded for the user home directory and environment variables.

### DIFF
--- a/mypy/config_parser.py
+++ b/mypy/config_parser.py
@@ -34,6 +34,14 @@ def parse_version(v: str) -> Tuple[int, int]:
     return major, minor
 
 
+def expand_path(path: str) -> str:
+    """Expand the user home directory and any environment variables contained within
+    the provided path.
+    """
+
+    return os.path.expandvars(os.path.expanduser(path))
+
+
 def split_and_match_files(paths: str) -> List[str]:
     """Take a string representing a list of files/directories (with support for globbing
     through the glob library).
@@ -45,8 +53,7 @@ def split_and_match_files(paths: str) -> List[str]:
     expanded_paths = []
 
     for path in paths.split(','):
-        path = path.strip()
-        path = os.path.expandvars(os.path.expanduser(path))
+        path = expand_path(path.strip())
         globbed_files = fileglob.glob(path, recursive=True)
         if globbed_files:
             expanded_paths.extend(globbed_files)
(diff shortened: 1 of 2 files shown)


mypy/config_parser.py
```diff
@@ -63,8 +64,9 @@ def split_and_match_files(paths: str) -> List[str]:
     'python_version': parse_version,
     'strict_optional_whitelist': lambda s: s.split(),
     'custom_typing_module': str,
-    'custom_typeshed_dir': str,
-    'mypy_path': lambda s: [p.strip() for p in re.split('[,:]', s)],
+    'custom_typeshed_dir': lambda s: os.path.expandvars(os.path.expanduser(s)),
```
msullivan: How about defining a function that does the two expansions together?

hetmankp:
Ok, let me know if that's what you had in mind.

msullivan:
Yeah!

mypy/config_parser.py
```diff
@@ -63,8 +71,8 @@ def split_and_match_files(paths: str) -> List[str]:
     'python_version': parse_version,
     'strict_optional_whitelist': lambda s: s.split(),
     'custom_typing_module': str,
-    'custom_typeshed_dir': str,
-    'mypy_path': lambda s: [p.strip() for p in re.split('[,:]', s)],
+    'custom_typeshed_dir': expand_path,
+    'mypy_path': lambda s: [expand_path(p.strip()) for p in re.split('[,:]', s)],
     'files': split_and_match_files,
     'quickstart_file': str,
     'junit_xml': str,
```
Mattwmaster58: Would enabling environment variable interpolation for this config key be as simple as this?
```suggestion
    'junit_xml': expand_path
```

hetmankp: > Would enabling environment variable interpolation for this config key be as simple as this?

Yes. Probably should have done this for `junit_xml` and `quickstart_file` as well.

Mattwmaster58: Alright, I'll include that in my PR #8479

PR #7274, opened 2019-07-29
Fix XML attribute order in tests

Michael0x2a:
Our tests recently started breaking -- it turns out this was because LXML 4.4.0 was released this Saturday, and LXML apparently respects now respects the insertion order of dicts/keyword arguments when setting XML attributes.

See the changelog for more details: https://github.com/lxml/lxml/blob/master/CHANGES.txt

I didn't feel like tweaking a bunch of tests, so decided to fix this in a relatively lazy way: I just went through every place we instantiated an "Element" or "SubElement" object in report.py and manually sorted the attributes.

The only exception was the precision info in stats.py -- I discovered to my mild surprise that the order of those stats apparently has semantic meaning (the higher the number, the less "precise" that line is?), so decided it might be more useful to preserve that ordering.

And for good measure, I pinned the LXML test requirement to 4.4.0 so we won't need to remember to keep attributes sorted or whatever when adding new tests/reports in the future.


JelleZijlstra:
Unfortunately this breaks 3.5 in CI, probably because dicts don't preserve insertion order in 3.5.  @srittau and I talked about this issue briefly in python/typeshed#3156. Maybe we'll have to use DOM comparison instead in the tests.

msullivan:
Sure. Thanks for fixing this!

Issue #5159, opened 2018-06-06
Separate syntactic and semantic nodes?


ilevkivskyi:
There is something that I was thinking about for a year or so that became more apparent while working on type alias refactoring (and planning the module refactoring). We have two different things mixed together in mypy code: syntactic nodes and "semantic" nodes. This is most obvious in `fixup.py` where we use `NodeVisitor` interface (intended to traverse syntactic tree) to patch symbol tables (something that carries purely semantic info). As a result, there are various weird things like `visit_var`, `visit_type_info`, etc. The idea is to have clear separation between syntactic and semantic nodes and have separate visitors for them. We currently already have this partially, we have `TypeInfo` vs `ClassDef`, and `Var` vs `AssignmentStmt`.

Here is a _short (and approximate)_ summary of the proposal:
* There are following semantic nodes: `Var`, `Function`, `Class` (non-leaf, has symbol table), `Module` (non-leaf, ditto), `TypeVar`, `TypeAlias`, `ConditionalNode` (see below).
* The above nodes inherit from `SymbolNode`, while syntactic nodes inherit from `Node`, these two both inherit from `Context`.
* Only the above nodes (and types) are serialized in cache and deserialized in incremental runs.
* Semantic nodes can have attributes that point to the relevant (defining) syntactic nodes (for example variable can have defining assignment, or function statement if property), but we should limit this to minimize cache size.
* The above nodes will have their separate visitor, so that in total we have three: for AST, for symbol tables, and for types.
* `ConditionalNode` exists to avoid having `.nodes` instead of `.node` in `SymbolTableNode`s (the latter are thin wrappers around semantic nodes), while still supporting certain conditional definitions like conditional imports. This node will have an attribute that is a list of other semantic nodes.

This is a very large refactoring, but IMO this will add robustness and clarity, and will simplify addition of new features (e.g. conditional imports). This is probably a low priority (long term) thing. We just discussed this with @JukkaL, he likes the idea but is concerned about the size of this refactoring. A possible way to go forward with this is to split this in several separate PRs.

elazarg:
I tried to do this several times without success - the changes were unmanageable for me. I really hope this will be done.

ilevkivskyi:
Some ideas to split this into smaller parts:
* Kill `Decorator`, it isn't present in native Python AST, and currently this node appears in both syntactic and semantic trees (intersection that we want to get rid of).
[...]

PR #7280, opened 2019-08-02
Free typechecker state and ASTs when they are no longer needed

msullivan:
Instead of hanging around to all of them for the entire run of the
process, free ASTs and typechecker state (especially the type map) as
soon as a module is finished being processing.

In order to have this work when generating fine-grained dependencies,
we need to produce fine-grained dependencies much earlier, so
BuildManager now has an `fg_deps` field.

In the daemon, only free typechecker state, since we want to keep ASTs
around to increase recheck speed. (A future change might use an LRU
cache to keep only some around.)

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -59,6 +59,7 @@ from mypy.metastore import MetadataStore, FilesystemMetadataStore, SqliteMetadat
 from mypy.typestate import TypeState, reset_global_state
 from mypy.renaming import VariableRenameVisitor
 from mypy.config_parser import parse_mypy_comments
+from mypy.freetree import free_tree
 
 
 # Switch to True to produce debug output related to fine-grained incremental
@@ -471,6 +472,7 @@ class BuildManager:
       missing_modules: Set of modules that could not be imported encountered so far
       stale_modules:   Set of modules that needed to be rechecked (only used by tests)
       fg_deps_meta:    Metadata for fine-grained dependencies caches associated with modules
+      fg_deps:         A fine-grained dependency map
       version_id:      The current mypy version (based on commit id when possible)
       plugin:          Active mypy plugin(s)
       plugins_snapshot:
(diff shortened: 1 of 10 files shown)


msullivan:
This cuts mypy's memory usage when doing a full check of S in half.

mypy/build.py
```diff
@@ -2172,6 +2171,16 @@ def finish_passes(self) -> None:
                                 typemap=self.type_map())
             manager.report_file(self.tree, self.type_map(), self.options)
 
+            self.update_fine_grained_deps(self.manager.fg_deps)
+            self.free_state()
+            if not manager.options.fine_grained_incremental and not manager.options.preserve_asts:
+                free_tree(self.tree)
```
Michael0x2a: Is there a reason why we can't do `self.tree = None`? Or maybe doing `self.tree.defs.clear()` or something?

mypy/freetree.py
```diff
@@ -0,0 +1,14 @@
+"""Generic node traverser visitor"""
+
+from mypy.traverser import TraverserVisitor
+from mypy.nodes import Block, MypyFile
+
+
+class TreeFreer(TraverserVisitor):
+    def visit_block(self, block: Block) -> None:
+        super().visit_block(block)
```
Michael0x2a: Would clearing just the top-level blocks be sufficient here, instead of recursively clearing?

Michael0x2a:
Also, it could maybe be worth adding one or two tests that make sure reports aren't disrupted by this? I don't really remember when we accumulate stats for those/whether we end up needing full access to the ASTs after they're freed. (Not sure if the test failures are related to this potential failure mode -- they mostly just reminded me that reports are a thing.)

mypy/freetree.py
msullivan: We need to go recursively or we'll miss nested structures that appear in the symbol table.

mypy/build.py
msullivan: The reason we need to go deeper than just clearing `defs` is that the symbol table contains references to functions and classes and the like, and we need to clear out their bodies.

It would be possible to drive this from the symbol table, and I considered that, but it is more complex for not much gain. (Probably a bit faster, but probably doesn't matter)

(We actually probably should also do `self.tree.defs.clear()`, which would free a little bit more memory, though probably not a substantial amount.)

msullivan:
I'm going to go ahead and merge this so that I can catch it in an internal release, but I'm happy to make any additional revisions if Ivan wants next week

JukkaL:
This is a nice win!

mypy/freetree.py
```diff
@@ -0,0 +1,21 @@
...
+        block.body.clear()
+
+
+def free_tree(tree: MypyFile) -> None:
+    """Free all the ASTs associated with a module.
+
+    This needs to be done recursively, since symbol tables contain
+    references to definitions, so those won't be freed but we want their
+    contents to be.
```
ilevkivskyi: We might fix this in long term, see https://github.com/python/mypy/issues/5159, but definitely not right now. I don't have any additional comments here, thanks for improving this!

PR #7283, opened 2019-08-03
scripts/stubtest: make it run again and type-check it

bluetech:
This script broke some time ago. Update it and type-check it so at least it keeps "compiling".

The script depends on `scripts/dumpmodule` so add type annotations there as well.

See individual commits for easier review.

Refs: https://github.com/python/mypy/issues/3329#issuecomment-515668522


msullivan:
I don't really have any opinions about `stubtest` but this seems pretty reasonable. Thanks.

PR #7286, opened 2019-08-04
Config option to ignore file parse errors

indygreg:
**RFC quality. Not anywhere close to landing.**

The Mercurial project has some .py files that are Python 2 or
Python 3 only. Mypy's current behavior is to find all .py files
and attempt to parse them. If the parse fails, this is a blocking
error and execution aborts immediately. This behavior is obviously
not desirable for projects containing mixed Python version sources.

This commit teaches Mypy to make parse errors non-fatal/blocking.

It teaches the option system a new `ignore_parse_errors` boolean
to control whether parse errors are ignored. This option is
consulted at the low-level parser to determine whether to make
the parse error a blocker.

Open issues:

* Ignoring parse errors is a special case of "ignore file." Perhaps
  the more general feature would be better?
* The option doesn't seem to work on per-module entries in the
  mypy.ini file. I'm not sure why.
* Missing tests.
* Missing documentation updates.
* Is inserting an empty ast tree on parse failure the proper failure
  mode? I'm not sure what the implications are for using an empty
  tree versus ignoring the module completely.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -173,7 +173,8 @@ def parse(source: Union[str, bytes],
         tree.path = fnam
         tree.is_stub = is_stub_file
     except SyntaxError as e:
-        errors.report(e.lineno, e.offset, e.msg, blocker=True)
+        errors.report(e.lineno, e.offset, e.msg,
+                      blocker=not options.ignore_parse_errors)
         tree = MypyFile([], [], False, set())
 
     if raise_on_error and errors.is_errors():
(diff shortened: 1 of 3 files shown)


msullivan:
I think I agree with the hypothesis that a more general feature of skipping files would be better.

A common approach to using mypy tends to be to have it driven by an external script that handles such logic and passes mypy a list of files, but I think it would be better if mypy had better support for this.

I think that adding a `skip_file` directive would be the right approach. It could then be specified internal to files as well with `# mypy: skip-file`.

msullivan:
I'm going to close this. I'd review a PR to add a `# mypy: skip-file` directive

PR #7294, opened 2019-08-06
Remove all references to the old semantic analyzer.

msullivan:
Follow-up diffs will delete the code and move the new analyzer

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -29,11 +29,8 @@ from typing_extensions import ClassVar, Final, TYPE_CHECKING
 from mypy_extensions import TypedDict
 
 from mypy.nodes import MypyFile, ImportBase, Import, ImportFrom, ImportAll, SymbolTable
-from mypy.semanal_pass1 import SemanticAnalyzerPass1
 from mypy.newsemanal.semanal_pass1 import SemanticAnalyzerPreAnalysis
-from mypy.semanal import SemanticAnalyzerPass2, apply_semantic_analyzer_patches
-from mypy.semanal_pass3 import SemanticAnalyzerPass3
-from mypy.newsemanal.semanal import NewSemanticAnalyzer
+from mypy.newsemanal.semanal import SemanticAnalyzer
 from mypy.newsemanal.semanal_main import semantic_analysis_for_scc
 from mypy.checker import TypeChecker
 from mypy.indirection import TypeIndirectionVisitor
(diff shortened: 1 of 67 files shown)


mypy/typeops.py
```diff
@@ -8,11 +8,12 @@
 # TODO: Move more type operations here
 
 from mypy.types import TupleType, Instance
-from mypy.join import join_type_list
 
 
 def tuple_fallback(typ: TupleType) -> Instance:
     """Return fallback type for a tuple."""
+    from mypy.join import join_type_list
```
JukkaL: What's this change?

test-data/unit/check-incremental.test
```diff
@@ -3338,7 +3340,7 @@ class C:
     d: int = attr.ib(converter=parse)
 [builtins fixtures/attr.pyi]
 [out1]
-main:6: note: Revealed type is 'def (a: Union[builtins.float, builtins.str], b: Union[builtins.str, builtins.bytes, builtins.int], c: builtins.str, d: Union[builtins.int, builtins.str]) -> a.C'
+5main:6: note: Revealed type is 'def (a: Union[builtins.float, builtins.str], b: Union[builtins.str, builtins.bytes, builtins.int], c: builtins.str, d: Union[builtins.int, builtins.str]) -> a.C'
```
JukkaL: What's this?

test-data/unit/check-incremental.test
```diff
@@ -4474,8 +4476,8 @@ def outer() -> None:
 [out]
 [out2]
 
-[case testRecursiveAliasImported]
-# flags: --no-new-semantic-analyzer
+[case testRecursiveAliasImported-skip]
```
JukkaL: It looks like this test case can be removed, as the next case covers is roughly the same for the new semantic analyzer. Here the behavior in the new analyzer is different by design.

test-data/unit/check-incremental.test
msullivan: A typo

mypy/typeops.py
msullivan: There is some fragility in mypy about import cycles, where certain modules need to be imported before others for things to work. Removing references to the old analyzer caused one of these places to surface when trying to run tests, so I fixed it. There are a couple more that I will try to fix in a separate PR.

JukkaL: One of these days I'm going to try to clean up the import cycles (though it might not help with this in particular).

mypy/build.py
```diff
@@ -2024,31 +1989,22 @@ def semantic_analysis_pass1(self) -> None:
...
-            # import statements as unreachable.
-            first = SemanticAnalyzerPass1(self.manager.semantic_analyzer)
-            with self.wrap_context():
-                first.visit_file(self.tree, self.xpath, self.id, options)
+        # Do the first pass of semantic analysis: analyze the reachability
+        # of blocks and import statements. We must do this before
+        # processing imports, since this may mark some import statements as
+        # unreachable.
+        #
+        # TODO: Once we remove the old semantic analyzer, this no longer should
```
ilevkivskyi: This TODO item might be updated.

mypy/newsemanal/typeanal.py
```diff
@@ -210,6 +210,7 @@ def visit_unbound_type_nonoptional(self, t: UnboundType, defining_literal: bool)
                 an_args = self.anal_array(t.args)
                 disallow_any = self.options.disallow_any_generics and not self.is_typeshed_stub
                 res = expand_type_alias(target, all_vars, an_args, self.fail, node.no_args, t,
+                                        unexpanded_type=t,
```
ilevkivskyi: This change looks unrelated, or am I missing something?

test-data/unit/check-literal.test
```diff
@@ -694,7 +694,7 @@ y: Foo[Foo]     # E: Literal[...] must have at least one parameter
 [out]
 
 [case testLiteralBadRawExpressionWithBadType]
-# flags: --new-semantic-analyzer
+
```
ilevkivskyi: Are these extra empty lines hard to avoid?

msullivan: They are easy to remove but they cause a lot of test failures due to line numbers changing and
1. Fixing them would have caused a lot more churn in the diff
2. I didn't want to fix them :P

mypy/build.py
msullivan: Yeah

mypy/newsemanal/typeanal.py
msullivan: `expand_type_alias` is used by the type checker and switching the checker to using the new analyzer required some tweaks

Issue #7275, opened 2019-07-30
Unexpected error comparing set and frozenset for equality with --strict-equality


neiljp:
I just upgraded a local mypy installation to 0.720 (python 3.6.5) and was exploring `--strict-equality`.

My interpretation is that sets and frozensets can be compared just fine; as such the following error was rather unexpected:
```bash
$mypy --strict-equality /tmp/blah.py
blah.py:3: error: Non-overlapping equality check (left operand type: "FrozenSet[int]", right operand type: "Set[int]")
```
for
```python3
# blah.py
x = frozenset([1,2,3])
y = set([1,2,3])
if x == y:
    print("equal")
```
As expected, `python3 blah.py` gives `equal`.

msullivan:
That is cute and frustrating. The error trips because the types aren't considered overlapping since neither is a superclass of the other, but nevertheless `__eq__` returns true because they are specifically built to handle that.

A related failure mode is `{1: 2}.keys() == {1}`, which fails due to non-overlapping equality of `Set` and `KeysView`.

If this is relatively uncommon, we can maybe handle it with some special casing (of `AbstractSet`, perhaps) in `dangerous_comparison`. If this turns out to happen more than I would guess, we might need to do something smart. (Looking at `__eq__` types, for example)

ilevkivskyi:
@msullivan I think we can special-case both cases.

PR #7302, opened 2019-08-08
Whitelist some equality checks under --strict-equality

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7275

As I mentioned in the issue, we can just whitelist some pairs of classes.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -71,6 +71,15 @@ ArgChecker = Callable[[Type, Type, int, Type, int, int, CallableType, Context, M
 MAX_UNIONS = 5  # type: Final
 
 
+# Types considered safe for comparisons with --strict-equality due to known behaviour of __eq__.
+OVERLAPPING_TYPES_WHITELIST = [('builtins.set', 'builtins.frozenset'),
+                               ('typing.KeysView', 'builtins.set'),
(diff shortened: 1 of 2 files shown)


mypy/checkexpr.py
```diff
@@ -71,6 +71,15 @@
 MAX_UNIONS = 5  # type: Final
 
 
+# Types considered safe for comparisons with --strict-equality due to known behaviour of __eq__.
+OVERLAPPING_TYPES_WHITELIST = [('builtins.set', 'builtins.frozenset'),
+                               ('typing.KeysView', 'builtins.set'),
+                               ('typing.KeysView', 'builtins.frozenset')]  # type: Final
```
msullivan: Is it worth doing Values also? (And Items?) (Genuinely not sure.)

ilevkivskyi: Interesting:
```
>>> {1: 2}.values() == {2}
False
>>> {1: 2}.items() == {(1, 2)}
True
```
So it works for keys and items, but not values.

JelleZijlstra: There was actually a discussion about this on the python-ideas mailing list recently. Even comparing `d.values() == d.values()` (where d is a dict) returns False.

PR #7306, opened 2019-08-08
Analyze and check default arguments to lambdas

msullivan:
This will allow us to report errors in them and also is needed to
support them in mypyc.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -577,9 +577,6 @@ class SemanticAnalyzer(NodeVisitor[None],
                 if isinstance(defn, FuncDef):
                     assert isinstance(defn.type, CallableType)
                     defn.type = set_callable_name(defn.type, defn)
-            for arg in defn.arguments:
-                if arg.initializer:
-                    arg.initializer.accept(self)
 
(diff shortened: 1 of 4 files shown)


mypy/semanal.py
```diff
@@ -871,6 +868,11 @@ def add_function_to_symbol_table(self, func: Union[FuncDef, OverloadedFuncDef])
     def analyze_function_body(self, defn: FuncItem) -> None:
         is_method = self.is_class_scope()
         with self.tvar_scope_frame(self.tvar_scope.method_frame()):
+            # Analyze default arguments
+            for arg in defn.arguments:
+                if arg.initializer:
+                    arg.initializer.accept(self)
```
ilevkivskyi: I assume you move this here because this method is already called on lambdas. I however see two problems with this:
* The comment on line 560 above becomes outdated (can't comment so far above).
* This can potentially cause problem with scopes if make some refactorings in future, so I would add a test like this (if we don't have one already):
```python
def f(x=i):  # Should give a 'Name not defined' error
    i = 42
```

Issue #6207, opened 2019-01-17
mypy fails suddenly with No module named '_sqlite3'


vseguin:
Please provide more information to help us understand the issue:

* Are you reporting a bug, or opening a feature request?
Bug

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.

Very minimal mypy test:
```
import unittest

from mypy import api


class TypeChecker(unittest.TestCase):
    def test_check_type_for_all_modules(self):
        stdout, stderr, exit_code = api.run(
            ['--strict', 'api', 'producer', 'i18n_extractor', 'services', 'database', 'main']
        )
        self.assertEqual(exit_code, 0, f'mypy validation failed\r\nType errors:\r\n{stdout}\r\nErrors:\r\n{stderr}')

```

* What is the actual behavior/output?
It just explodes with the following stack trace:

```
Traceback (most recent call last):
  File "/tmp/pythonproject/tests/test_mypy.py", line 8, in test_check_type_for_all_modules
    stdout, stderr, exit_code = mypy.api.run(['--strict', 'api', 'main', 'services', 'session', 'system_test'])
  File "/tmp/pythonproject/venv/lib/python3.6/site-packages/mypy/api.py", line 73, in run
    from mypy.main import main
  File "/tmp/pythonproject/venv/lib/python3.6/site-packages/mypy/main.py", line 14, in <module>
    from mypy import build
  File "/tmp/pythonproject/venv/lib/python3.6/site-packages/mypy/build.py", line 55, in <module>
    from mypy.metastore import MetadataStore, FilesystemMetadataStore, SqliteMetadataStore
  File "/tmp/pythonproject/venv/lib/python3.6/site-packages/mypy/metastore.py", line 12, in <module>
    import sqlite3
  File "/usr/local/lib/python3.6/sqlite3/__init__.py", line 23, in <module>
    from sqlite3.dbapi2 import *
  File "/usr/local/lib/python3.6/sqlite3/dbapi2.py", line 27, in <module>
    from _sqlite3 import *
ModuleNotFoundError: No module named '_sqlite3'

```

* What is the behavior/output you expect?

That it works as before :)

* What are the versions of mypy and Python you are using?
  Do you see the same issue after installing mypy from Git master?

mypy 0.660, Python 3.6.0

* What are the mypy flags you are using? (For example --strict-optional)

--strict

* If mypy crashed with a traceback, please paste
[...]

PR #7307, opened 2019-08-08
Only import sqlite3 when using the sqlite metadata store

msullivan:
Every so often I am trying to test on a semi-broken python that
I built from source on a machine where I hadn't installed sqlite
and it fails because of this and doesn't need to.


msullivan:
And this "fixes" #6207, which was correctly closed as not a bug but we have fixed anyways :P

Issue #730, opened 2015-07-27
Catch TypeVisitor subclasses with unimplemented abstract methods


JamesGuthrie:
Issue #638 was a pain to find because the root cause is an unimplemented abstract method which silently returns `None` and somewhere down the line causes type checking to fail. It would be better to check this explicitly. The following shows that this is done for one of the abstract methods:
https://github.com/JukkaL/mypy/blob/bbb2dd4f07f995a4b28b8044fa91aa97960fe192/mypy/types.py#L532. There are two potential solutions: 1) use abc's @abstractmethod decorator, 2) raise an error directly in TypeVisitor.

1) Ensures that future implementations and changes catch subclasses which do not implement the abstract methods, but requires adding implementations for all of the abstract methods to existing subclasses which are missing them. As there are a few, it would probably be simplest to raise an error there instead of providing the complete implementation, which raises the question: why not just use 2) straight off the bat?

@JukkaL Do you have a preferred solution?

o11c:
I am suspicious of any visitors that can meaningfully work without full implementations.

Perhaps there is a need for more specific visitors?

JamesGuthrie:
It appears as though there is a common pattern of unimplemented abstract methods (visit_star_type, visit_overloaded, visit_error_type, visit_erased_type), I figured that maybe they had been added after the initial implementation of all visitors, and don't appear in the test cases and as such haven't been caught. In total there are about 25 subclasses which do not implement all abstract methods, mostly it is either visit_star_type or visit_overloaded which are unimplemented.

JukkaL:
The current implementation is just sloppy, and is inherited from early pre-mypy days when there were no abstract methods.

Using `@abstractmethod` is probably the best option, as missing implementations can then be caught statically. Trying to catch all of these errors using test cases is an uphill battle.

JamesGuthrie:
I've opened a PR to get the ball rolling.

As mentioned in the PR, Subclasses with incomplete implementations are:

~~~
EraseTypeVisitor (mypy/erasetype.py: 26)
    visit_ellipsis_type()
    visit_star_type()

SameTypeVisitor(mypy/sametype.py: 29)
    visit_overloaded()
    visit_ellipsis_type()
    visit_star_type()

ExpandTypeVisitor(mypy/expandtype.py: 35)
    visit_ellipsis_type()
    visit_star_type()

SubtypeVisitor(mypy/subtypes.py: 44)
    visit_ellipsis_type()
    visit_star_type()

[...]

PR #7312, opened 2019-08-10
Make TypeVisitor and SyntheticTypeVisitor (almost) fully abstract

Michael0x2a:
This pull request almost resolves https://github.com/python/mypy/issues/730 -- it makes all TypeVisitor and SyntheticTypeVisitor methods with the exception of TypeAliasType abstract.

This thankfully required fairly minimal changes. I ended up needing to make only three non-trivial changes:

1. The visitor for PlaceholderType was moved out of TypeVisitor into SyntheticTypeVisitor. My understanding is that PlaceholderType is a semantic analysis only type, so this ought to be safe.

2. As a consequence of 1, I ended up removing the existing PlaceholderType visitor method from TypeTranslator. (It seems that method was added about 6 months ago when PlaceholderType was first introduced, and was untouched since then).

3. TypeIndirectionVisitor was turned from a SyntheticTypeVisitor into a TypeVisitor. I believe we call this visitor only in the final pass, after the type-checking phase.


mypy/fixup.py
```diff
@@ -185,6 +185,9 @@ def visit_overloaded(self, t: Overloaded) -> None:
         for ct in t.items():
             ct.accept(self)
 
+    def visit_erased_type(self, o: Any) -> None:
+        pass  # Nothing to descend into.
```
ilevkivskyi: I think ideally this should raise a `RuntimeError` as `visit_partial_type()` below, because this type only exists temporarily during type inference.

mypy/server/astmerge.py
```diff
@@ -373,6 +373,9 @@ def visit_overloaded(self, t: Overloaded) -> None:
         if t.fallback is not None:
             t.fallback.accept(self)
 
+    def visit_erased_type(self, t: ErasedType) -> None:
```
ilevkivskyi: Same here. Does it actually break some tests when you add a raise here?

mypy/server/astmerge.py
```diff
@@ -429,6 +432,10 @@ def visit_union_type(self, typ: UnionType) -> None:
         for item in typ.items:
             item.accept(self)
 
+    def visit_placeholder_type(self, t: PlaceholderType) -> None:
```
ilevkivskyi: I think this also should raise, since this should be called after semantic analysis is done. But maybe I am wrong because there are many visit methods for synthetic types. Or maybe it was mistakenly made `SyntheticTypeVisitor` instead of just `TypeVisitor`? Anyway, this is not important for this PR.

mypy/server/deps.py
```diff
@@ -901,6 +901,9 @@ def visit_overloaded(self, typ: Overloaded) -> List[str]:
             triggers.extend(self.get_type_triggers(item))
         return triggers
 
+    def visit_erased_type(self, t: ErasedType) -> List[str]:
```
ilevkivskyi: And this one too.

mypy/typeanal.py
```diff
@@ -485,6 +488,9 @@ def visit_callable_type(self, t: CallableType, nested: bool = True) -> Type:
                                   variables=self.anal_var_defs(variables))
         return ret
 
+    def visit_overloaded(self, t: Overloaded) -> Type:
+        return t
```
ilevkivskyi: Why this is so different from `visit_callable_type()` above?

mypy/typeanal.py
```diff
@@ -451,6 +451,9 @@ def visit_none_type(self, t: NoneType) -> Type:
     def visit_uninhabited_type(self, t: UninhabitedType) -> Type:
         return t
 
+    def visit_erased_type(self, t: ErasedType) -> Type:
```
ilevkivskyi: I think this too should raise, see `visit_partial_type()` below.

mypy/server/astmerge.py
Michael0x2a: I think it was intentional that this visitor is a SyntheticTypeVisitor -- we apparently try running this visitor on `FuncBase.unanalyzed_type` in the `process_base_func` method up above. Not entirely sure why, but I figured it'd be safest to just leave this alone.

Michael0x2a:
Thanks for the review! I added exceptions/asserts to the `visit_erased_types` methods as suggested, and added a comment about Overloaded.

Issue #7214, opened 2019-07-15
Bug: false positive with --warn-unreachable


jcugat:
Please provide more information to help us understand the issue:

* Are you reporting a bug, or opening a feature request?
Bug

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.

```
from contextlib import suppress
from typing import Dict


def foo(blah: Dict[str, str]):
    with suppress(KeyError):
        return blah["blah"]
    return blah["default"]


print(foo({"default": "default"}))
```

* What is the actual behavior/output?
```
$ mypy --warn-unreachable test.py
test.py:8: error: Statement is unreachable
```

* What is the behavior/output you expect?
No error, since line 8 is reachable.

* What are the versions of mypy and Python you are using?

```
$ pip freeze | grep mypy
mypy==0.720
mypy-extensions==0.4.1
$ python -V
Python 3.6.8
```

* What are the mypy flags you are using? (For example --strict-optional)
`--warn-unreachable`

Maybe related to https://github.com/python/mypy/issues/7204 and https://github.com/python/mypy/issues/7207

JukkaL:
This seems like a bug in how mypy calculates reachability of with statements.

cc @Michael0x2a

Michael0x2a:
@JukkaL -- I'm not sure if there's a simple solution for this one either. Basically, we could try adjusting mypy so it (correctly) treats all with statements as if they were the body of a try-finally block, which would mean statements after the with would start to be considered reachable.

But that ends up causing "Missing return statement" errors for code like the following:

```python
def foo() -> int:
    with open("foobar.txt", "r") as f:
        return 3
```

More generally, there doesn't seem to be a clean way of distinguishing between context managers that suppress exceptions vs ones that don't.

It'd be nice to find some way of encoding this distinction at the type-system level (for example, by taking advantage of the proposed `Annotated` type?), but maybe for now we could get by just hard-coding a list of these exception-swallowing context managers within mypy?

JelleZijlstra:
You could in theory get this out of the return type of the context manager's `__exit__`: if it returns `None`, you know that it doesn't swallow exceptions, and if it returns `bool` or `Optional[bool]`, you know that it may. Of course that's not sufficient, because you don't in general know the concrete contextmanager class involved (`typing.ContextManager.__exit__` is annotated as returning `Optional[bool]`).

[...]

PR #7317, opened 2019-08-10
Make warn-unreachable understand exception-swallowing contextmanagers

Michael0x2a:
This pull request fixes https://github.com/python/mypy/issues/7214: it makes mypy treat any contextmanagers where the `__exit__` returns `bool` or `Literal[True]` as ones that can potentially swallow exceptions.

Contextmanagers that return `Any`, `Optional[bool]`, None, or `Literal[False]` continue to be treated as non-exception-swallowing ones.

This distinction helps the `--warn-unreachable` flag do the right thing in this example program:

```python
from contextlib import suppress

def should_warn() -> str:
    with contextlib.suppress(IndexError):
        return ["a", "b", "c"][0]

def should_not_warn() -> str:
    with open("foo.txt") as f:
        return "blah"
```

This pull request needs the typeshed changes I made in https://github.com/python/typeshed/pull/3179 before the tests and the mypy self-check will pass. Once that one gets merged, I'll update typeshed and rebase this pull request.


ilevkivskyi:
I am going to close/re-open to trigger builds.

Michael0x2a:
FYI, the tests for this PR aren't going to pass -- https://github.com/python/typeshed/pull/3179 needs to be landed first.

ilevkivskyi:
Oh, I somehow thought it was already merged. We can make then a dedicated sync when this will be ready. Also there is now a bonus self-check failure (easy to fix however).

JelleZijlstra:
I merged the typeshed PR.

test-data/unit/check-unreachable-code.test
```diff
@@ -963,3 +963,229 @@ class Test3(Generic[T2]):
...
+        return 3
+    noop()                         # E: Statement is unreachable
+
+def f_bad_2() -> int:
+    with DoesNotSuppress1():  # E: "DoesNotSuppress1" has no attribute "__enter__"; maybe "__aenter__"? \
+                              # E: "DoesNotSuppress1" has no attribute "__exit__"; maybe "__aexit__"?
+        return 3
+    noop()                    # E: Statement is unreachable
+
+[typing fixtures/typing-full.pyi]
```
ilevkivskyi: Keep newline.

test-data/unit/pythoneval.test
```diff
@@ -1405,3 +1405,23 @@ _testStrictEqualityWhitelist.py:5: error: Non-overlapping equality check (left o
 _testStrictEqualityWhitelist.py:11: error: Non-overlapping equality check (left operand type: "KeysView[int]", right operand type: "Set[str]")
 _testStrictEqualityWhitelist.py:12: error: Non-overlapping equality check (left operand type: "ValuesView[int]", right operand type: "Set[int]")
 _testStrictEqualityWhitelist.py:13: error: Non-overlapping equality check (left operand type: "KeysView[int]", right operand type: "List[int]")
+
+[case testUnreachableWithOpen]
+# mypy: warn-unreachable
+
+class Suppresses:
+    def __enter__(self) -> int: ...
+    def __exit__(self, exctype: object, excvalue: object, traceback: object) -> bool: ...
```
ilevkivskyi: Maybe instead use `from contextlib import suppress`, which is IMO another important case in stdlib.

test-data/unit/check-unreachable-code.test
```diff
@@ -963,3 +963,229 @@ class Test3(Generic[T2]):
...
+def f_suppress_3() -> int:  # E: Missing return statement
+    with Suppresses2():
+        return 3
+    noop()
+
+def f_mix() -> int:         # E: Missing return statement
+    with DoesNotSuppress1(), Suppresses1(), DoesNotSuppress1():
+        return 3
+    noop()
+[typing fixtures/typing-full.pyi]
```
ilevkivskyi: This test and the below one are quite long. Maybe split each one into three: suppresses, doesn't suppress, and mixed?

mypy/types.py
```diff
@@ -2405,6 +2405,19 @@ def remove_optional(typ: Type) -> ProperType:
         return typ
 
 
+def is_literal_type(typ: Type, fallback_fullname: str, value: LiteralValue) -> bool:
+    """Returns 'true' if this type is a LiteralType with the given value
+    and underlying base fallback type.
+    """
```
ilevkivskyi: It would be easy to fit this on a single line if you phrase this as `Check if this type is a LiteralType with the given value and fallback type.`

mypy/checker.py
```diff
@@ -3341,12 +3341,24 @@ def check_incompatible_property_override(self, e: Decorator) -> None:
...
+                exit_ret_type = self.check_with_item(expr, target, s.unanalyzed_type is None)
+            if is_literal_type(exit_ret_type, "builtins.bool", False):
+                continue
+            if is_literal_type(exit_ret_type, "builtins.bool", True):
+                exceptions_maybe_suppressed = True
+            elif (isinstance(exit_ret_type, Instance)
+                    and exit_ret_type.type.fullname() == 'builtins.bool'):
+                exceptions_maybe_suppressed = True
+        if exceptions_maybe_suppressed:
+            with self.binder.frame_context(can_skip=True, try_frame=True):
```
ilevkivskyi: Although it is almost clear, please add a short comment here attracting attention to `try_frame=True`.

mypy/checker.py
```diff
@@ -3341,12 +3341,24 @@ def check_incompatible_property_override(self, e: Decorator) -> None:
...
-                self.check_with_item(expr, target, s.unanalyzed_type is None)
-        self.accept(s.body)
+                exit_ret_type = self.check_with_item(expr, target, s.unanalyzed_type is None)
+            if is_literal_type(exit_ret_type, "builtins.bool", False):
+                continue
+            if is_literal_type(exit_ret_type, "builtins.bool", True):
+                exceptions_maybe_suppressed = True
+            elif (isinstance(exit_ret_type, Instance)
+                    and exit_ret_type.type.fullname() == 'builtins.bool'):
+                exceptions_maybe_suppressed = True
```
ilevkivskyi: Can these two be merged with an `or`?

mypy/checker.py
```diff
@@ -3341,12 +3341,24 @@ def check_incompatible_property_override(self, e: Decorator) -> None:
...
+        exceptions_maybe_suppressed = False
         for expr, target in zip(s.expr, s.target):
             if s.is_async:
-                self.check_async_with_item(expr, target, s.unanalyzed_type is None)
+                exit_ret_type = self.check_async_with_item(expr, target, s.unanalyzed_type is None)
             else:
-                self.check_with_item(expr, target, s.unanalyzed_type is None)
-        self.accept(s.body)
+                exit_ret_type = self.check_with_item(expr, target, s.unanalyzed_type is None)
+            if is_literal_type(exit_ret_type, "builtins.bool", False):
```
ilevkivskyi: Are these rules/conventions documented somewhere? I think they should be, otherwise we will just forget them (either in mypy docs, btw do we have a section about context managers?, or in typeshed). Maybe also add a link here.

Michael0x2a: I did add some notes about these conventions to the typeshed contribution page (at the bottom of the [conventions][0] section), but this conventions currently aren't otherwise documented.

We do mention context managers once on the [Protocols page](https://mypy.readthedocs.io/en/stable/protocols.html#context-manager-protocols) -- I'll add a note about this there in a follow-up PR.

  [0]: https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#conventions

Michael0x2a:
As an FYI, I ended up making one more change to this PR: basically, I made it so we no longer assume context managers suppress exceptions when `__exit__` returns bool when strict-optional mode is *disabled*. I found while playing around with `open(...)` that we didn't necessarily preserve whether `__exit__` returns bool vs Optional[bool] in that mode, so decided it'd probably be safer to just give up and bias towards retaining the old behavior.

LMK if you see a better way of handling this or if handled the no-strict-optional check incorrectly, and I can submit a follow-up PR to fix.

ilevkivskyi:
> We do mention context managers once on the [Protocols page](https://mypy.readthedocs.io/en/stable/protocols.html#context-manager-protocols) -- I'll add a note about this there in a follow-up PR.

Thanks! Sounds good to me.

Issue #7284, opened 2019-08-03
`assert` second (error) expression should be type checked as conditional on the failed predicate


gwk:
`assert` has two forms; the two-expression form could be handled better by the type checker. The second expression only evaluates when the predicate (first expression) fails, and should therefore be type checked under that assumption.

For example, here we have a regex that validates a string. When the regex matches, the assertion fails; `m` was just shown not to be None, so the f-string is ok to evaluate `m[0]`.
```
import re
s = 'a!b'
m = re.search('\W', s)
assert not m, f'illegal char: {m[0]!r}'
```

Mypy currently rejects this code, but should accept it:
```
t.py:6: error: Value of type "Optional[Match[str]]" is not indexable
```

msullivan:
I think it should be pretty simple to fix this using logic similar to how `and` and `or` are handled

gantsevdenis:
May I take this?

msullivan:
Sure! Thanks!

PR #7318, opened 2019-08-10
Second expression in assert statement should be checked as conditional

gantsevdenis:
Fixes https://github.com/python/mypy/issues/7284


test-data/unit/check-expressions.test
```diff
@@ -2303,3 +2303,9 @@ def f(x: any) -> None:  # E: Name 'any' is not defined \
 def f(x: Optional[str]) -> None:  # E: Name 'Optional' is not defined \
                                   # N: Did you forget to import it from "typing"? (Suggestion: "from typing import Optional")
     pass
+
+[case testAssertionLazily]
+from typing import Optional, List
+li: Optional[List] = []
+assert li is None, li[0]
+[builtins fixtures/list.pyi]
```
ilevkivskyi: Please keep newline at the end of file.

mypy/checker.py
```diff
@@ -2964,14 +2964,16 @@ def visit_operator_assignment_stmt(self,
...
-            self.expr_checker.accept(s.msg)
-
         if isinstance(s.expr, TupleExpr) and len(s.expr.items) > 0:
             self.fail(message_registry.MALFORMED_ASSERT, s)
 
+        true_map, else_map = self.find_isinstance_check(s.expr)
+        if s.msg is not None:
+            with self.binder.frame_context(can_skip=False, fall_through=1):
+                self.push_type_map(else_map)
+                self.expr_checker.accept(s.msg)
```
ilevkivskyi: It looks like there is a code duplication with `ExpressionChecker.analyze_cond_branch()` why can't you re-use it here? (If there is a reason I am missing please add a comment).

gantsevdenis: I must have missed it!  I ll check that

ilevkivskyi: It looks like there is either a misunderstanding, or you did something wrong in git. I just asked if it is possible to re-use the existing helper here, but you just removed the context manger instead, which is wrong. Namely, if the `else_map` is `None` (i.e. the condition is statically always true), then binder will mark the _current_ frame unreachable.

For example, in this test the missing return statement will be no more caught:
```
[case testAssertNew]
def f() -> int:  # Correct error here on master, no error on your branch.
    x: int
    assert isinstance(x, int), 'Hm...'
```
Btw, since this was not caught by our test suite, could you please add such test? (Maybe also copy few other basic assert tests adding an assert message to them.)

gantsevdenis:
Actually I wasn't sure how to use `analyze_cond_branch()` (I think its 3rd argument, context, is None here), so I went with what I thought, was the safest decision.. turned out to be wrong

So I have added the test that you have suggested, I also added another test with `isinstance` because with first attempt that test would fail while it shoudn't
Thank you for your review time

mypy/checker.py
```diff
@@ -2979,14 +2979,14 @@ def visit_operator_assignment_stmt(self,
...
-
         if isinstance(s.expr, TupleExpr) and len(s.expr.items) > 0:
             self.fail(message_registry.MALFORMED_ASSERT, s)
 
         # If this is asserting some isinstance check, bind that type in the following code
-        true_map, _ = self.find_isinstance_check(s.expr)
+        true_map, else_map = self.find_isinstance_check(s.expr)
+        if s.msg is not None:
+            with self.binder.frame_context(can_skip=False, fall_through=1):
+                self.expr_checker.analyze_cond_branch(else_map, s.msg, None)
```
ilevkivskyi: Hm, sorry for being annoying, but it looks like something is still wrong, now _two_ frames are being pushed, one manually, and one in the helper. Why can't you just use the helper here?

The `None` context is indeed right, because we have no expectations about what should be the assertion "message" (it can be arbitrary object, not just a string).

gantsevdenis: > Why can't you just use the helper here?

Mostly because I misunderstood the binder and its `fall_through` argument

PR #7323, opened 2019-08-12
Add error code arguments to plugin API

JukkaL:
Update some `fail()` methods to accept an error code argument.


JukkaL:
> Do we need a test case for how this is used in a plugin?

This would be nice, but I don't think that it's urgent. I'll create a follow-up issue.

Issue #6759, opened 2019-05-03
Subclasses cannot use properties to override super-class attributes


OddBloke:
When subclassing, overriding a class attribute with a property with the same type causes type checking failures. This is (I think?) a fairly common way for subclasses to introduce dynamic behaviour transparently to consumers, so my expectation is that this would not cause failures.

#### Python Code

```py
from typing import List


class Super:

    x: List[str] = []
    reveal_type(x)


class Sub(Super):

    @property
    def x(self) -> List[str]:
        return []
    reveal_type(x)
```

#### mypy Output

```
$ mypy reproducer.py
reproducer.py:7: error: Revealed type is 'builtins.list[builtins.str]'
reproducer.py:13: error: Signature of "x" incompatible with supertype "Super"
reproducer.py:15: error: Revealed type is 'def (self: reproducer.Sub) -> builtins.list[builtins.str]'
```

#### Versions

```
$ python3 --version  # This is the packaged version in Ubuntu 19.04
Python 3.7.2+
$ mypy --version
mypy 0.710+dev.b29a4330d2d7d1032c30a1990e7f1c477db070fd
```

JelleZijlstra:
This is unsafe because the superclass allows writing to `x` (`y = Super(); y.x = []` works), but the subclass doesn't (`y = Sub(); y.x = []` is an AttributeError).

OddBloke:
Aha, right, that makes sense. Thanks!

ilevkivskyi:
Mypy however still complains if I make the property settable. I looked at the code and it is indeed buggy (several corner cases with `is_property` and `is_settable_property` are not considered).

Generally, the way how we check incompatibility with supertype is fragile and uses a lot of code duplication with `checkmember.py`. I would say we should simplify this significantly by just re-using `checkmember.py`. See also https://github.com/python/mypy/issues/5136

ilevkivskyi:
Oh, this is actually a duplicate of https://github.com/python/mypy/issues/4125

PR #7330, opened 2019-08-13
Implementing background infrastructure for recursive types: Part 1

ilevkivskyi:
During planning discussions one of the main concerns about recursive types was the fact that we have hundreds of places where certain types are special-cased using `isinstance()`, and fixing all of them will take weeks.

So I did a little experiment this weekend, to understand how bad it _actually_ is. I wrote a simple mypy plugin for mypy self-check, and it discovered 800+ such call sites. This looks pretty bad, but it turns out that fixing half of them (roughly 400 plugin errors) took me less than 2 days. This is kind of a triumph of our tooling :-) (i.e. mypy plugin + PyCharm plugin).

Taking into account results of this experiment I propose to actually go ahead and implement recursive types. Here are some comments:

* There will be four subsequent PRs: second part of `isinstance()` cleanup, implementing visitors and related methods everywhere, actual core implementation, adding extra tests for tricky recursion patterns.
* The core idea of implementation stays the same as we discussed with @JukkaL: `TypeAliasType` and `TypeAlias` node will essentially match logic between `Instance` and `TypeInfo` (but structurally, as for protocols)
* I wanted to make `PlaceholderType` a non-`ProperType`, but it didn't work immediately because we call `make_union()` during semantic analysis. If this seems important, this can be done with a bit more effort.
* I make `TypeType.item` a proper type (following PEP 484, only very limited things can be passed to `Type[...]`). I also make `UnionType.items` proper types, mostly because of `make_simplified_union()`. Finally, I make `FuncBase.type` a proper type, I think a type alias can never appear there.
* It is sometimes hard to decide where exactly is to call `get_proper_type()`, I tried to balance calling them not too soon and not too late, depending of every individual case. Please review, I am open to modifying logic in some places.

### DIFF
--- a/mypy/plugins/common.py
+++ b/mypy/plugins/common.py
@@ -6,7 +6,10 @@ from mypy.nodes import (
 )
 from mypy.plugin import ClassDefContext
 from mypy.semanal import set_callable_name
-from mypy.types import CallableType, Overloaded, Type, TypeVarDef, LiteralType, Instance, UnionType
+from mypy.types import (
+    CallableType, Overloaded, Type, TypeVarDef, LiteralType, Instance, UnionType,
+    get_proper_type, get_proper_types
+)
 from mypy.typevars import fill_typevars
 from mypy.util import get_unique_redefinition_name
 
@@ -55,7 +58,7 @@ def _get_argument(call: CallExpr, name: str) -> Optional[Expression]:
     callee_node = call.callee.node
     if (isinstance(callee_node, (Var, SYMBOL_FUNCBASE_TYPES))
             and callee_node.type):
-        callee_node_type = callee_node.type
+        callee_node_type = get_proper_type(callee_node.type)
         if isinstance(callee_node_type, Overloaded):
             # We take the last overload.
             callee_type = callee_node_type.items()[-1]
@@ -141,18 +144,20 @@ def try_getting_str_literals(expr: Expression, typ: Type) -> Optional[List[str]]
     2. 'typ' is a LiteralType containing a string
     3. 'typ' is a UnionType containing only LiteralType of strings
     """
+    typ = get_proper_type(typ)
+
     if isinstance(expr, StrExpr):
         return [expr.value]
 
     if isinstance(typ, Instance) and typ.last_known_value is not None:
         possible_literals = [typ.last_known_value]  # type: List[Type]
     elif isinstance(typ, UnionType):
-        possible_literals = typ.items
+        possible_literals = list(typ.items)
     else:
         possible_literals = [typ]
 
     strings = []
-    for lit in possible_literals:
+    for lit in get_proper_types(possible_literals):
         if isinstance(lit, LiteralType) and lit.fallback.type.fullname() == 'builtins.str':
             val = lit.value
             assert isinstance(val, str)
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -2155,7 +2155,6 @@ def true_or_false(t: Type) -> ProperType:
 
 def function_type(func: mypy.nodes.FuncBase, fallback: Instance) -> FunctionLike:
     if func.type:
-        assert isinstance(func.type, ProperType)  # TODO: can this ever be an alias?
         assert isinstance(func.type, FunctionLike)
         return func.type
     else:
@@ -2333,7 +2332,7 @@ def get_proper_types(it: Iterable[Optional[Type]]) -> List[Optional[ProperType]]
 names = globals().copy()  # type: Final
 names.pop('NOT_READY', None)
 deserialize_map = {
-    key: obj.deserialize  # type: ignore
+    key: obj.deserialize
     for key, obj in names.items()
     if isinstance(obj, type) and issubclass(obj, Type) and obj is not Type
 }  # type: Final
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -2360,6 +2360,9 @@ def get_type_vars(typ: Type) -> List[TypeVarType]:
 
 def flatten_nested_unions(types: Iterable[Type]) -> List[ProperType]:
     """Flatten nested unions in a type list."""
+    # This and similar functions on unions can cause infinite recursion
+    # if passed a "pathological" alias like A = Union[int, A] or similar.
+    # TODO: ban such aliases in semantic analyzer.
     flat_items = []  # type: List[ProperType]
     for tp in get_proper_types(types):
         if isinstance(tp, UnionType):
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -2399,11 +2399,11 @@ def remove_optional(typ: Type) -> ProperType:
 
 @overload
 def get_proper_type(typ: None) -> None: ...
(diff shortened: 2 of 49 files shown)


mypy/types.py
```diff
@@ -144,6 +144,97 @@ def deserialize(cls, data: JsonDict) -> 'Type':
         raise NotImplementedError('Cannot deserialize {} instance'.format(cls.__name__))
 
 
+class TypeAliasType(Type):
+    """
+    A type alias to another type.
+
+    To support recursive type aliases we don't immediately expand a type alias
+    during semantic analysis, but create an in stance of this type that records the target alias
```
JukkaL: Typo: "in stance"

mypy/types.py
```diff
@@ -144,6 +144,97 @@ def deserialize(cls, data: JsonDict) -> 'Type':
...
+        super().__init__(line, column)
+        self.alias = alias
+        self.args = args
+        self.type_ref = None  # type: Optional[str]
+
+    def expand_once(self) -> Type:
+        """Expand to the target type exactly once.
+
+        This doesn't do full expansion, i.e. the result can contain another
+        (or even this same) type alias.
```
JukkaL: Maybe this should be considered an internal helper (perhaps prefixed with `_`) and we could suggest using `get_proper_type` instead.

mypy/types.py
```diff
@@ -144,6 +144,97 @@ def deserialize(cls, data: JsonDict) -> 'Type':
...
+    def expand_all_if_possible(self) -> Optional['ProperType']:
+        """Attempt a full expansion of the type alias (including nested aliases).
+
+        If the expansion is not possible, i.e. the alias is (mutually-)recursive,
+        return None.
+        """
+        raise NotImplementedError('TODO')
+
+    @property
+    def can_be_true(self) -> bool:  # type: ignore
```
JukkaL: The `# type: ignore` may break mypyc is strange ways, so it would be better not to use it.

mypy/types.py
```diff
@@ -144,6 +144,97 @@ def deserialize(cls, data: JsonDict) -> 'Type':
...
+        assert self.alias is not None
+        return self.alias.target.can_be_true
+
+    @property
+    def can_be_false(self) -> bool:  # type: ignore
```
JukkaL: Similar to above.

mypy/types.py
```diff
@@ -144,6 +144,97 @@ def deserialize(cls, data: JsonDict) -> 'Type':
         raise NotImplementedError('Cannot deserialize {} instance'.format(cls.__name__))
 
 
+class TypeAliasType(Type):
+    """
+    A type alias to another type.
```
JukkaL: Style nit: move the first line to the same line with `"""` (`"""A type alias ...`).

mypy/types.py
```diff
@@ -144,6 +144,97 @@ def deserialize(cls, data: JsonDict) -> 'Type':
...
+
+    To support recursive type aliases we don't immediately expand a type alias
+    during semantic analysis, but create an in stance of this type that records the target alias
+    definition node (mypy.nodes.TypeAlias) and type arguments (for generic aliases).
+
+    This is very similar to how TypeInfo vs Instance interact.
+    """
```
JukkaL: Style nit: Add empty line after `"""`, for consistency.

mypy/types.py
```diff
@@ -144,6 +144,97 @@ def deserialize(cls, data: JsonDict) -> 'Type':
         raise NotImplementedError('Cannot deserialize {} instance'.format(cls.__name__))
 
 
+class TypeAliasType(Type):
+    """
+    A type alias to another type.
+
+    To support recursive type aliases we don't immediately expand a type alias
```
JukkaL: Add a prominent note that this is not being used yet, and the implementation is still incomplete?

mypy/types.py
```diff
@@ -868,7 +959,7 @@ def __init__(self,
         assert len(arg_types) == len(arg_kinds) == len(arg_names)
         if variables is None:
             variables = []
-        self.arg_types = arg_types
+        self.arg_types = list(arg_types)
```
JukkaL: What about making sure that these are expanded enough so that they are proper types (but may contain type alias references)? Not sure if this would be worth it, but it might let us avoid a bunch of `get_proper_type` calls elsewhere.

mypy/types.py
```diff
@@ -1216,7 +1307,7 @@ def deserialize(cls, data: JsonDict) -> 'Overloaded':
         return Overloaded([CallableType.deserialize(t) for t in data['items']])
 
 
-class TupleType(Type):
+class TupleType(ProperType):
```
JukkaL: Similar to above, I can imagine that expanding items could be useful (or not).

mypy/plugins/common.py
```diff
@@ -55,7 +58,7 @@ def _get_argument(call: CallExpr, name: str) -> Optional[Expression]:
     callee_node = call.callee.node
     if (isinstance(callee_node, (Var, SYMBOL_FUNCBASE_TYPES))
             and callee_node.type):
-        callee_node_type = callee_node.type
+        callee_node_type = get_proper_type(callee_node.type)
```
JukkaL: Since this change may affect plugins, it needs to be documented in plugin docs and the plugins changelog issue before the next mypy release.

ilevkivskyi: Yes, I actually added this expansion to minimize possible breakages (at least few of our plugins used `isinstance()` with these types).

mypy/types.py
ilevkivskyi: There is actually a mypy false positive https://github.com/python/mypy/issues/6759 that doesn't allow fixing this. I will leave a comment here with a link to issue.

mypy/types.py
ilevkivskyi: It looks like this is actually not possible. Actually even what I did for unions is problematic, for example if we have (not very meaningful, since technically `A = int`) `A = Union[int, A]`. To represent this we must have somewhere a union that contains an item that is `A`. I will try to fix this now.

ilevkivskyi: Actually these kind of "trivial" recursive unions are the most problematic, maybe we should actually ban them. Another solution would be to actually remove aliases being defined from immediate union, but this can silently hide an error, so I don't like this.

ilevkivskyi: After some more thinking I think I convinced myself that aliases like `A = Union[int, A]` are really bad. Essentially they are not much different from `A = A`, for example all situations like this:
```python
x: A
x.to_bytes()
```
cause infinite recursion. Also, it is very hard to imagine any practical use for such aliases. Finally, they can be easily detected during semantic analysis and banned.

If we ban them, then we can keep `UnionType.items` and `TypeType.item` proper. I however didn't find any simple consistent representation schema where we could also make `TupleType.items` and `CallableType.arg_types` (and `.ret_type`) proper, mostly because of things like `A = Union[int, Callable[[], A]]`.

@JukkaL If you don't have objections, I would merge this to get things moving (sorry for pinging during vacation :-)).

PR #7335, opened 2019-08-13
Silence deprecation warning from typed-ast in tests

JukkaL:
This fixes a deprecation warning about `<>` in Python 2 that came up every
time the full test suite was run.

### DIFF
--- a/mypy/fastparse2.py
+++ b/mypy/fastparse2.py
@@ -107,6 +107,7 @@ def parse(source: Union[str, bytes],
     is_stub_file = fnam.endswith('.pyi')
     try:
         assert options.python_version[0] < 3 and not is_stub_file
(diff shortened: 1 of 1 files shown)


mypy/fastparse2.py
```diff
@@ -106,7 +107,9 @@ def parse(source: Union[str, bytes],
     is_stub_file = fnam.endswith('.pyi')
     try:
         assert options.python_version[0] < 3 and not is_stub_file
-        ast = ast27.parse(source, fnam, 'exec')
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", category=DeprecationWarning)
```
JelleZijlstra: Is it worth adding a comment that this is to suppress warnings about <>?

Issue #6662, opened 2019-04-11
Enable --fast-exit by default


JukkaL:
If we'd enable `--fast-exit` by default, mypy would run a bit faster.

ilevkivskyi:
In general I am fine with this. What are the possible downsides? Also do you have some numbers to compare?

msullivan:
We'd want to make sure that it didn't get turned on by default when mypy is invoked through the API, but other than that it seems like a basically good idea.

JukkaL:
There are some edge cases where this could make a difference, such a plugin that relies on files being flushed at process exit, but these should be rare. I don't have recent performance measurements, but if I remember correctly, the speedup was in the 5% range for big mypy runs.

gvanrossum:
I'm not too keen on this. People with big code bases should use the daemon
anyway. And if they don't, they can use this flag explicitly. (We probably
should do so in our internal scripts that don't use the daemon.)

PR #7337, opened 2019-08-14
--fast-exit enabled by default

devbihari:
closes: #6662

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -552,7 +552,8 @@ def process_options(args: List[str],
         dest='shadow_file', action='append',
         help="When encountering SOURCE_FILE, read and type check "
              "the contents of SHADOW_FILE instead.")
-    add_invertible_flag('--fast-exit', default=True, group=internals_group)
+    add_invertible_flag('--fast-exit', default=True, help=argparse.SUPPRESS,
+                        group=internals_group)
 
     error_group = parser.add_argument_group(
         title='Error reporting',
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -552,8 +552,8 @@ def process_options(args: List[str],
         dest='shadow_file', action='append',
         help="When encountering SOURCE_FILE, read and type check "
              "the contents of SHADOW_FILE instead.")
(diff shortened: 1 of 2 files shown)


mypy/main.py
```diff
@@ -552,7 +552,7 @@ def add_invertible_flag(flag: str,
         dest='shadow_file', action='append',
         help="When encountering SOURCE_FILE, read and type check "
              "the contents of SHADOW_FILE instead.")
-    add_invertible_flag('--fast-exit', default=False, help=argparse.SUPPRESS,
+    add_invertible_flag('--fast-exit', default=True, help=argparse.SUPPRESS,
```
ilevkivskyi: I would say if we enable this by default we should at least make this publicly visible. Also Ideally this not affect runs made through `api.run()` calls.

cc @gvanrossum

devbihari: Forgive me if you find my question stupid but I think I need a bit of clarification here. What exactly do you mean by "make this publicly visible"? As in, should the user be warned that "--fast-exit" is now enabled by default? or is there something I'm unable to understand?
As I've only started contributing recently, any guidance will be much appreciated.

ilevkivskyi: > What exactly do you mean by "make this publicly visible"?

Two things:
* Remove `help=argparse.SUPPRESS`
* Describe it in the docs.

devbihari: --fast-exit was never described in the documentation as an experimental feature or otherwise. I've added it :smile:

mypy/main.py
```diff
@@ -552,8 +552,8 @@ def add_invertible_flag(flag: str,
         dest='shadow_file', action='append',
         help="When encountering SOURCE_FILE, read and type check "
              "the contents of SHADOW_FILE instead.")
-    add_invertible_flag('--fast-exit', default=False, help=argparse.SUPPRESS,
-                        group=internals_group)
+    add_invertible_flag('--fast-exit', default=True, help="makes mypy exit faster",
```
ilevkivskyi: I would add a bit more info here, explaining the down sides (for example files not being flushed at exit).

msullivan:
I'm going to close this absent some consensus about whether we want it

PR #7354, opened 2019-08-16
Always erase last known value from instances when using binder or partial types

ilevkivskyi:
See the test cases for simplified example of currently problematic behaviour when using `--strict-equality`.


ilevkivskyi:
Hm, this is one more case when binder and partial types share bugs. Looks like another argument in favour of trying to combine them.

PR #7355, opened 2019-08-16
semanal: Treat __init_subclass__ as an implicit classmethod

### LINKED ISSUE #7352: super().__init_subclass__() fails
```
$ mypy bin/insubcl.py 
bin/insubcl.py:3: error: Too few arguments for "__init_subclass__" of "object"
$ cat bin/insubcl.py 
class X:
    def __init_subclass__(cls) -> None:
        super().__init_subclass__()
$ mypy --version
mypy 0.730+dev.36fcf1d5aa0f5656ac367bc66aba6e390f82ea9a.dirty
```

This code works correctly at runtime and the latest release accepts it.

tavianator:
Fixes #7352

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -439,7 +439,7 @@ class FuncBase(Node):
                  'unanalyzed_type',
                  'info',
                  'is_property',
-                 'is_class',        # Uses "@classmethod"
+                 'is_class',        # Uses "@classmethod" (explicit or implicit)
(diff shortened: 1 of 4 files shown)


tavianator:
From my reading of https://www.python.org/dev/peps/pep-0487/, we should treat `__init_subclass__` as if it were decorated with `@classmethod`, so this PR does that.  Not sure the code is in the right place.

PR #7358, opened 2019-08-16
Consider fallback instance promotions in is_overlapping_types()

ilevkivskyi:
Using this opportunity I also made a tiny refactoring to avoid code duplication (that would grow even bigger if I would also add logic for promotions).

### DIFF
--- a/mypy/meet.py
+++ b/mypy/meet.py
@@ -276,7 +276,13 @@ def is_overlapping_types(left: Type,
         right = right.fallback
 
     if isinstance(left, LiteralType) and isinstance(right, LiteralType):
-        return left == right
+        if left.value == right.value:
+            # If values are the same, we still need to check if fallbacks are overlapping,
+            # this is done below.
+            left = left.fallback
+            right = right.fallback
+        else:
+            return False
     elif isinstance(left, LiteralType):
         left = left.fallback
     elif isinstance(right, LiteralType):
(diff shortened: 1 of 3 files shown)


mypy/meet.py
```diff
@@ -285,15 +288,13 @@ def _type_object_overlap(left: ProperType, right: ProperType) -> bool:
     # Finally, we handle the case where left and right are instances.
 
     if isinstance(left, Instance) and isinstance(right, Instance):
-        if left.type.is_protocol and is_protocol_implementation(right, left):
-            return True
-        if right.type.is_protocol and is_protocol_implementation(left, right):
+        # First we need to handle promotions and structural compatibility for instances
+        # that came as fallbacks, so simply call is_subtype() to avoid code duplication.
+        if (is_subtype(left, right, ignore_promotions=ignore_promotions)
+                or is_subtype(right, left, ignore_promotions=ignore_promotions)):
```
Michael0x2a: It's somewhat unfortunate we end up needing to perform a subtype check again here (we already call `is_proper_subtype` earlier in the method).

But as you said, it doesn't seem like there's a clean way of working around this without repeating some of the promotion logic anyways...

test-data/unit/check-expressions.test
```diff
@@ -2303,6 +2303,17 @@ if f == 0:  # E: Non-overlapping equality check (left operand type: "FileId", ri
...
 
+[case testStrictEqualityPromotionsLiterals]
+# flags: --strict-equality --py2
+from typing import Final
+
+U_FOO = u'foo'  # type: Final
+
+if str() == U_FOO:
+    pass
+assert u'foo' == 'foo'
```
Michael0x2a: Maybe we can also add a line ensuring something like `assert u'foo' == u'bar'` is still flagged as an error?

PR #7366, opened 2019-08-18
Implementing background infrastructure for recursive types: Part 2

ilevkivskyi:
This completes the addition of expansion points before every `isinstance()` checks on types.

This essentially does the same as https://github.com/python/mypy/pull/7330. Couple comments:
* Many checks (probably more than half) in type checker are against `AnyType`, `NoneType`, and `UninhabitedType`, but technically these are valid targets for aliases, so I still do expansions there.
* I added couple hooks to the plugin to flag redundant `get_proper_type()` and `get_proper_types()` calls, however sometimes when they trigger they add some additional spurious errors about non-matching overloads. I tried to fix this, but it looks like this requires calling `get_function_hook()` once on a whole overload (now it is called for every item).

Next part will contain implementation of the visitors.


misc/proper_plugin.py
```diff
@@ -40,21 +42,40 @@ def isinstance_proper_hook(ctx: FunctionContext) -> Type:
...
+                                              'mypy.types.ErasedType'):
+            # Special case: these are not valid targets for a type alias and thus safe.
+            # TODO: introduce a SyntheticType base to simplify this?
+            return True
+    elif isinstance(right, TupleType):
+        return all(is_special_target(t) for t in get_proper_types(right.items))
+    return False
+
+
 def is_improper_type(typ: Type) -> bool:
```
Michael0x2a: Should we also consider 'Any' to be an improper type? (Could be overkill/out-of-scope for this PR)

misc/proper_plugin.py
```diff
@@ -40,21 +42,40 @@ def isinstance_proper_hook(ctx: FunctionContext) -> Type:
...
-                                                      'mypy.types.ProperType',
-                                                      'mypy.types.TypeAliasType'):
-                    # Special case: things like assert isinstance(typ, ProperType) are always OK.
-                    return ctx.default_return_type
-                if right.type_object().fullname() in ('mypy.types.UnboundType',
-                                                      'mypy.types.TypeVarType'):
-                    # Special case: these are not valid targets for a type alias and thus safe.
-                    return ctx.default_return_type
+            if is_special_target(right):
+                return ctx.default_return_type
```
Michael0x2a: Idea: we could also maybe add a check for when the left operand is Any and the right operand is some Type. Could be overkill/out-of-scope for this PR.

mypy/checkexpr.py
```diff
@@ -227,9 +229,10 @@ def analyze_ref_expr(self, e: RefExpr, lvalue: bool = False) -> Type:
...
-            if isinstance(var.type, Instance):
-                if self.is_literal_context() and var.type.last_known_value is not None:
-                    return var.type.last_known_value
+            var_type = get_proper_type(var.type)
+            if isinstance(var_type, Instance):
+                if self.is_literal_context() and var_type.last_known_value is not None:
+                    return var_type.last_known_value
                 if var.name() in {'True', 'False'}:
                     return self.infer_literal_expr_type(var.name() == 'True', 'builtins.bool')
             return var.type
```
Michael0x2a: Do things end up breaking if we return `var_type` and make the return value be ProperType?

misc/proper_plugin.py
```diff
@@ -40,21 +42,40 @@ def isinstance_proper_hook(ctx: FunctionContext) -> Type:
...
+            # Special case: things like assert isinstance(typ, ProperType) are always OK.
+            return True
+        if right.type_object().fullname() in ('mypy.types.UnboundType',
+                                              'mypy.types.TypeVarType',
+                                              'mypy.types.EllipsisType',
+                                              'mypy.types.StarType',
+                                              'mypy.types.TypeList',
+                                              'mypy.types.CallableArgument',
+                                              'mypy.types.PartialType',
+                                              'mypy.types.ErasedType'):
```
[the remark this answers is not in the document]
ilevkivskyi: I agree.

mypy/checker.py
```diff
@@ -999,6 +1012,8 @@ def is_reverse_op_method(self, method_name: str) -> bool:
     def check_for_missing_annotations(self, fdef: FuncItem) -> None:
         # Check for functions with unspecified/not fully specified types.
         def is_unannotated_any(t: Type) -> bool:
+            if not isinstance(t, ProperType):
+                return False
```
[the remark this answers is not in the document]
ilevkivskyi: Because we know that an alias can't target unannotated `Any`.

mypy/checker.py
```diff
@@ -1040,8 +1055,9 @@ def check___new___signature(self, fdef: FuncDef, typ: CallableType) -> None:
         self_type = fill_typevars_with_any(fdef.info)
         bound_type = bind_self(typ, self_type, is_classmethod=True)
         # Check that __new__ (after binding cls) returns an instance
-        # type (or any)
-        if not isinstance(bound_type.ret_type, (AnyType, Instance, TupleType)):
+        # type (or any).
+        if not isinstance(get_proper_type(bound_type.ret_type),
```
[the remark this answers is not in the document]
ilevkivskyi: > Perhaps this can come later as we convert our various helper functions to work with ProperType instead of Type?

Yes, we can try to put more order here after get some experience, currently it is pure intuition.

mypy/checker.py
```diff
@@ -2999,7 +3031,7 @@ def visit_raise_stmt(self, s: RaiseStmt) -> None:
 
     def type_check_raise(self, e: Expression, s: RaiseStmt,
                          optional: bool = False) -> None:
-        typ = self.expr_checker.accept(e)
+        typ = get_proper_type(self.expr_checker.accept(e))
```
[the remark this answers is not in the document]
ilevkivskyi: Answering here for all three places. I was thinking about this, but I want to preserve actual type as more as possible to keep the possibility to mention original alias name in error messages. Also putting it in such hot places as `accept()` may cause performance penalty.

mypy/checkexpr.py
ilevkivskyi: No, but again I would like to keep the original type for error message (if we will decide this at some point, I think there is even an issue about this).

misc/proper_plugin.py
ilevkivskyi: I think this and below actually makes sense. I will try and if it is not hard, I will add to this PR.

ilevkivskyi:
@msullivan Thanks!

> Any sort of unification based improvements to the type inference system would need something like this also.

I am sorry but I am not sure what do you mean by "unification based improvements to the type inference system", do you mean something like `TypeRangeType(lower, upper)` or `RefinedType(user_defined, refiner_1, refiner_2, ...)`?

Issue #2276, opened 2016-10-18
Add more test cases for --warn-no-return


JukkaL:
Here are some ideas for additional test cases (as follow-up to #1748):
- A try statement that returns almost everywhere but can fall through.
- Empty function with a literal `...` body, with or without a docstring.
-  For loops + missing return statements.
- `--warn-no-return` and `raise` or `assert False` as the last thing in a function.
- Nested try statements or loops.

(Sorry if some of these are already there.)

msullivan:
It's been 3 years and it seems to work

PR #7373, opened 2019-08-20
test(No return rule): add test - looping over generator case

yxliang01:
For unconditionally looping over non-exhaustive generators, it should be acceptable to have no `return` statement. For unconditionally looping over exhaustive generators, the function should have `return` statement.

Partially addressing #2276

This should be merged after fixing https://github.com/python/mypy/issues/7374 .


msullivan:
We can reopen if #7374 is fixed

PR #7378, opened 2019-08-21
Merge mypyc into the mypy repo

### LINKED ISSUE #7308: Merge mypyc into mypy
There are a number of annoying problems caused by mypyc being separated from mypy that essentially stem from mypy and mypyc forming one self-hosted compiler that is split between two git repos.

We should merge mypyc into mypy. My plan is to have `mypyc` as a top-level source directory in the `mypy` repo. mypyc's `test-data`, `doc`, and `external` directories would get sunk into `mypyc`, and configuration files would be merged.

mypyc would then become part of the mypy distribution.

I will try to do a history rewrite of mypyc to fix up issue/PR references to point to the mypyc repo.

I would like to continue to use the mypyc project's separate issue tracker, however.

I plan to do this around August 20 or so.

msullivan:
*DO NOT MERGE.* This will be an actual merge commit that I will do manually.

The general approach here is that `mypyc` becomes a top-level package
and mypyc-specific files are sunk down into its subdirectories. This includes
README, test-data, docs, and our vendored copy of googletest.

The existing CI test builds all now run the mypyc tests as well. New builds are added
to run mypyc run tests on OS X, linux with a debug build of python, and 32-bit windows.

A side effect of this is that we now test a fully bootstrapped mypyc, which is pretty cool.

There will probably need to be some follow-up work to improve docs, test experience,
etc, but I think we are ready to pull the trigger on this.

The mypyc history was rewritten before this merge to fix up
PR and issue references to be cross-repo references to the old mypyc repo. 

Closes #7308.


msullivan:
I merged this manually!

Issue #7384, opened 2019-08-23
Error codes in # type: ignore don't suppress related notes


ilevkivskyi:
For example:
```python
import nostubs  # type: ignore[import]
```
Causes
```
mypy/report.py:30: note: (Stub files are from https://github.com/python/typeshed)
```
and error code 1. I didn't check carefully, but I think this affects all error messages that also emit additional notes (there are quite many of them).

JukkaL:
Looks like each note should have the same error code as the corresponding error.

PR #7385, opened 2019-08-23
Use some more dogfooding

ilevkivskyi:
This PR enables `--strict-equality` and `--show-error-codes` for self-check, and adds error codes in some `# type: ignore`s. So far this revealed two issues:
* Both PyCharm and flake8 get crazy about error codes, generating many false positives
* Error codes in `# type: ignore`s don't always work, see https://github.com/python/mypy/issues/7384


msullivan:
I had a PR merged with pyflakes to fix the lint issues (https://github.com/PyCQA/pyflakes/pull/455) but it hasn't been released yet I think

ilevkivskyi:
@msullivan This is great! We can then remove `F821` from exception list when the new version is released.

ilevkivskyi:
I think this is quite uncontroversial, so I am going to merge this now.

PR #7386, opened 2019-08-24
docs/final_attrs.rst: Add missing "be".

pfalcon:
"*Final names* ... should not reassigned" misses "be" after "should".


pfalcon:
Made just via github online editor, so feel free to treat as a bugreport.

PR #7392, opened 2019-08-26
Accept empty list unpacking for function without parameter

### LINKED ISSUE #4997: Erroneous 'Too many arguments' error
It appears as though mypy treats argument unpacking as containing at least one element. Here's an example:

```python
from typing import List

class Foo:
    pass

args: List[str] = []
packet = Foo(*args)
```

```
tbug.py:7: error: Too many arguments for "Foo"
```

I guess Mypy is assuming that you would never unpack zero arguments.

In my particular case I have a registry of 'packet classes' and a dynamic number of arguments parsed from the network. Some of those packet classes legitimately have 0 parameters, and Mypy gives me an error about each of them.

Running `mypy 0.600` with no other flags.

onlined:
Fixes #4997.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1233,6 +1233,8 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         for i, kind in enumerate(actual_kinds):
             if i not in all_actuals and (
                     kind != nodes.ARG_STAR or
+                    # We accept the other iterables than tuple (including Any)
+                    # as star arguments because they could be empty, resulting no arguments.
                     is_non_empty_tuple(actual_types[i])):
                 # Extra actual: not matched by a formal argument.
                 ok = False
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1233,7 +1233,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         for i, kind in enumerate(actual_kinds):
             if i not in all_actuals and (
                     kind != nodes.ARG_STAR or
-                    is_not_empty_tuple(actual_types[i])):
+                    is_non_empty_tuple(actual_types[i])):
                 # Extra actual: not matched by a formal argument.
                 ok = False
                 if kind != nodes.ARG_NAMED:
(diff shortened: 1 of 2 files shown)


mypy/checkexpr.py
```diff
@@ -3833,9 +3833,9 @@ def is_async_def(t: Type) -> bool:
     return isinstance(t, Instance) and t.type.fullname() == 'typing.Coroutine'
 
 
-def is_empty_tuple(t: Type) -> bool:
+def is_not_empty_tuple(t: Type) -> bool:
```
JukkaL: I was a little confused by the name at first. What about renaming this to `is_non_empty_tuple`?

test-data/unit/check-varargs.test
```diff
@@ -448,6 +447,15 @@ def f(x: int, *, y: str) -> None: pass
 f(y='x', *(1,))
 [builtins fixtures/list.pyi]
 
+[case testVarArgsEmptyList]
```
JukkaL: I assume that there's an existing test case that covers the empty tuple case?

mypy/checkexpr.py
```diff
@@ -1233,7 +1233,7 @@ def check_for_extra_actual_arguments(self,
         for i, kind in enumerate(actual_kinds):
             if i not in all_actuals and (
                     kind != nodes.ARG_STAR or
-                    not is_empty_tuple(actual_types[i])):
+                    is_not_empty_tuple(actual_types[i])):
```
JukkaL: Here it would be nice to add a comment about list and `Any` star args potentially being empty, so we don't complain about them.

test-data/unit/check-varargs.test
onlined: Actually, there is not. I thought that a tuple test is not in the scope of this PR.

mypy/checkexpr.py
onlined: Fixed.

mypy/checkexpr.py
onlined: I added the comment.

test-data/unit/check-varargs.test
JukkaL: Since this changes the code related to handling empty/non-empty tuples as `*args`, it would be good to add a test case for this in case one doesn't exist already. This would validate that the change hasn't introduced a regression.

onlined: Added.

PR #7396, opened 2019-08-26
[mypyc] Don't swallow KeyboardInterrupt in test_run

msullivan:
Currently we use distutil's `run_setup` function to wrap the
invocation of `setup.py` in test_run, which speeds things up a little
bit. Unfortunately `run_setup` swallows KeyboardInterrupt (and all
other failures), which prevents ctrl-c from terminating tests (the
tests will fail instead with -n0; with parallel testing, pytest will
die but it will take several seconds, I think while it waits for tests
to finish.)

Make a clone of run_setup that doesn't have these problems.

### DIFF
--- a/mypyc/test/test_run.py
+++ b/mypyc/test/test_run.py
@@ -7,7 +7,7 @@ import subprocess
 import contextlib
 import shutil
 import sys
-from typing import Iterator, Optional, List
+from typing import Any, Iterator, Optional, List, cast
 
 from mypy import build
 from mypy.test.data import DataDrivenTestCase
@@ -54,6 +54,8 @@ def run_setup(script_name: str, script_args: List[str]) -> bool:
     and KeyboardInterrupt with no way to recover them (!).
     The real version has some extra features that we removed since
     we weren't using them.
+
+    Returns whether the setup succeeded.
     """
     save_argv = sys.argv.copy()
     g = {'__file__': script_name}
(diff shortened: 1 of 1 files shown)


mypyc/test/test_run.py
```diff
@@ -46,6 +44,39 @@
...
+        finally:
+            sys.argv = save_argv
+    except SystemExit as e:
+        # distutils converts KeyboardInterrupt into a SystemExit with
+        # "interrupted" as the argument. Convert it back so that
+        # pytest will exit instead of just failing the test.
+        if e.args == ("interrupted",):
+            raise KeyboardInterrupt
+
+        return False
```
JukkaL: What if the exit code is 0? Should we return True in that case?

mypyc/test/test_run.py
```diff
@@ -46,6 +44,39 @@
...
+def run_setup(script_name: str, script_args: List[str]) -> bool:
+    """Run a setup script in a somewhat controlled environment.
+
+    This is adapted from code in distutils and our goal here is that is
+    faster to not need to spin up a python interpreter to run it.
+
+    We had to fork it because the real run_setup swallows errors
+    and KeyboardInterrupt with no way to recover them (!).
+    The real version has some extra features that we removed since
+    we weren't using them.
```
JukkaL: Document the return value.

Issue #7367, opened 2019-08-19
Plugins and overloads don't interact well


ilevkivskyi:
While working on our own plugin for self-checking proper types and `isinstance()` I have tried to use `get_function_hook()` to emit some additional errors for overloaded functions. However this didn't work for three reasons:
* The hook is called for every item in the overload, not for the overload as a whole.
* The overload logic relies on empty `Messages` as a sign of successful match
* The errors emitted during overload selection are discarded afterwards, thus discarding the plugin generated errors

This issue can be fixed in three ways:
* Calling the hook on the overload as a whole after selecting the overload variant and type checking it (this however will be a breaking API change)
* Use newly added error codes to only abandon a variant if the error is due to bad argument type (this is however still bad because `check_call()` may have an `accept()` call possibly triggering almost arbitrary error including a nested function call with bad argument type).
* Use a boolean attribute somewhere that will be set by `check_argument_types()` and will be reset after return from every `accept()`.

I am leaning towards the last option, since it is most principled IMO.

cc @Michael0x2a

tapaswenipathak:
Hi @ilevkivskyi: I would like closing the issue. Can I PR?

ilevkivskyi:
Sure, you can try, but note that this may not be the best issue for a new contributor. If you would like to try few simpler issues first we have a special label https://github.com/python/mypy/issues?q=is%3Aopen+is%3Aissue+label%3Agood-first-issue

PR #7400, opened 2019-08-27
Remove mypy.messages from the typeops tangle

msullivan:
mypy.applytype needs to be able to report errors, so have it
take an error reporting callback.

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -33,7 +33,7 @@ from mypy.nodes import (
     CallExpr
 )
 from mypy.subtypes import (
-    is_subtype, find_member, is_subtype, get_member_flags,
+    is_subtype, find_member, get_member_flags,
     IS_SETTABLE, IS_CLASSVAR, IS_CLASS_OR_STATIC,
 )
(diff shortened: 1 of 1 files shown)


mypy/subtypes.py
```diff
@@ -1008,10 +1007,15 @@ def unify_generic_callable(type: CallableType, target: CallableType,
...
-    applied = mypy.applytype.apply_generic_arguments(type, non_none_inferred_vars, msg,
+    had_errors = False
+
+    def report(*args: Any) -> None:
+        nonlocal had_errors
+        had_errors = True
+
+    applied = mypy.applytype.apply_generic_arguments(type, non_none_inferred_vars, report,
                                                      context=target)
-    if msg.is_errors():
```
ilevkivskyi: This pattern is unfortunately used quite often in other contexts (in particular in overloads, see https://github.com/python/mypy/issues/7367). It would be great to get rid of those (but in case of methods I would probably prefer `self.flag` rather that `nonlocal`).

PR #7409, opened 2019-08-29
Fixed length tuple concatenation

### LINKED ISSUE #224: Concatenation of tuples
Concatenation of fixed-length tuples should be supported by the type checker. 

Arbitrary tuples could be concatenated. `Tuple[int, str] + Tuple[bool]` results in `Tuple[int, str, bool]`.

Concatenating two homogeneous tuples (`Tuple[T, ...]`) already works.

onlined:
Fixes #224.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1886,6 +1886,13 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 column=typ.column,
             ))
 
+    def concat_tuples(self, left: TupleType, right: TupleType) -> TupleType:
+        """Concatenate two fixed length tuples."""
+        result = TupleType(items=left.items + right.items,
+                           fallback=self.named_type('builtins.tuple'))
+        result.partial_fallback = tuple_fallback(result)
+        return result
+
     def visit_int_expr(self, e: IntExpr) -> Type:
         """Type check an integer literal (trivial)."""
         return self.infer_literal_expr_type(e.value, 'builtins.int')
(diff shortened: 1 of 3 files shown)


mypy/checkexpr.py
```diff
@@ -1886,6 +1886,13 @@ def infer_literal_expr_type(self, value: LiteralValue, fallback_name: str) -> Ty
                 column=typ.column,
             ))
 
+    def concat_tuples(self, left: TupleType, right: TupleType) -> TupleType:
+        """Concatenate two fixed length tuples."""
+        result = TupleType(items=left.items + right.items,
+                           fallback=self.named_type('builtins.tuple'))
+        result.partial_fallback = tuple_fallback(result)
```
JukkaL: This line seems unnecessary. Anybody who needs the fallback type should just call `tuple_fallback` instead of assuming that `partial_fallback` has the correct type.

mypy/checkexpr.py
```diff
@@ -1945,6 +1952,12 @@ def visit_op_expr(self, e: OpExpr) -> Type:
                     return self.strfrm_checker.check_str_interpolation(e.left, e.right)
         left_type = self.accept(e.left)
 
+        proper_left_type = get_proper_type(left_type)
+        if isinstance(proper_left_type, TupleType) and e.op == '+':
+            proper_right_type = get_proper_type(self.accept(e.right))
+            if isinstance(proper_right_type, TupleType):
+                return self.concat_tuples(proper_left_type, proper_right_type)
```
JukkaL: There is a slight complication here: it's possible that the fallback type has a custom `__add__` override that does something different from what's defined in `tuple`. In that case we can't safely use this special casing. So there should be a check that checks whether the `__add__` method in the left partial fallback comes from `builtins.tuple`. If not, we use the old code path below instead of using `concat_tuples`. Let me know if you need more hints about how to do this.

onlined: Isn't the same thing applicable for `__radd__` and right tuple?

JukkaL: Yes, you are correct.

mypy/checkexpr.py
onlined: Thanks! Removed the line.

JukkaL:
To fix the build, you need to add `tuple.__add__` to `test-data/unit/fixtures/tuple.pyi`.

onlined:
Thanks!

mypy/checkexpr.py
onlined: Added the check.

onlined:
@JukkaL The change in stubs broke a test, I think I should change the test too. What do you think?

JukkaL:
Yeah, you should change the tests to match the new output.

onlined:
Fixed the tests!

PR #7414, opened 2019-08-29
Support suggesting types for class and static methods

msullivan:


test-data/unit/fine-grained-suggest.test
```diff
@@ -456,3 +456,31 @@ def bar() -> None:
...
+        return x
+
+[file bar.py]
+from typing import Any
+any: Any
+
+from foo import F
+def bar() -> None:
+    F.bar(0, any)
+    F.baz(0, any)
```
JukkaL: Maybe call with different types here, to make sure that separate signatures can be inferred? What about calling a static/class method through an instance instead of a type object?

PR #7415, opened 2019-08-29
Format suggestions using the "signature" pyannotate format instead

msullivan:
This bypasses pyannotate's parsing and processing of annotations
and allows us to just specify them directly. This is useful because
pyannotate doesn't support Callable.

### DIFF
--- a/mypy/suggestions.py
+++ b/mypy/suggestions.py
@@ -650,9 +650,6 @@ class TypeFormatter(TypeStrVisitor):
         else:
             return super().visit_union_type(t)
 
-    def visit_type_var(self, t: TypeVarType) -> str:
-        return t.name
-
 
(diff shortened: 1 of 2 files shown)


mypy/suggestions.py
```diff
@@ -247,7 +253,11 @@ def get_args(self, is_method: bool,
             if default:
                 all_arg_types.append(default)
 
-            if all_arg_types:
+            if len(all_arg_types) == 1 and isinstance(get_proper_type(all_arg_types[0]), NoneType):
```
ilevkivskyi: Generally this is good idea, this is consistent to what `stubgen` does. But why do we need to check that length is 1? This is not a set, how do we know all the types are unique there?

PR #7416, opened 2019-08-29
Support suggesting callable types

msullivan:


mypy/suggestions.py
```diff
@@ -580,6 +580,24 @@ def visit_tuple_type(self, t: TupleType) -> str:
         s = self.list_str(t.items)
         return 'Tuple[{}]'.format(s)
 
+    def visit_callable_type(self, t: CallableType) -> str:
+        args = []
+        for kind, name, typ in zip(t.arg_kinds, t.arg_names, t.arg_types):
+            arg = typ.accept(self)
+            if kind == ARG_STAR:
+                arg = '*' + arg
```
JukkaL: This doesn't actually work? `Callable[[*int], int]` is an error. You could "extended callable types" instead.

mypy/suggestions.py
```diff
@@ -580,6 +580,24 @@ def visit_tuple_type(self, t: TupleType) -> str:
...
+            elif kind == ARG_STAR2:
+                arg = '**' + arg
```
JukkaL: Similar to above.

mypy/suggestions.py
```diff
@@ -580,6 +580,24 @@ def visit_tuple_type(self, t: TupleType) -> str:
...
+            # ARG_NAMED we just mishandle, whatever
```
JukkaL: Extended callable types would work with these as well, but this seems pretty low-priority.

test-data/unit/fine-grained-suggest.test
```diff
@@ -324,6 +324,24 @@ def bar() -> None:
...
+def foo(f):
+    return f(0, "lol")
+[file bar.py]
+from foo import foo
+
+def whatever(x: int, y: str) -> int:
+    return 0
+
+def bar() -> None:
+    foo(whatever)
```
JukkaL: Test more complex argument callable, such as one with default argument values or `*args`.

PR #7421, opened 2019-08-29
Fix tests on 3.8 and enable it in CI

msullivan:
A couple things were needed to fix the tests:
 * mypy failed because ast.Str.kind doesn't exist
 * flake8 run with python3.8 now reports decorated function
   redefinitions on the line of the function definition and not the
   first decorator which means we would need to noqa overloads *twice*.
   I already found it really annoying so I disabled the check.
   (But it looks like the next version of pyflakes will fix the overload
   issue.)
 * pythoneval tests failed due to a bunch of deprecation warnings for
   `@coroutine` which I silenced.
 * One walrus test had broken because it hasn't been run
 * The docstring of `object` changed and we were printing it in an eval test
 * The column of something changed in one test

Pleasantly no actually code issues needed to be fixed.

### DIFF
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -1670,12 +1670,12 @@ class UnionType(ProperType):
     @overload
     @staticmethod
     def make_union(items: List[ProperType], line: int = -1, column: int = -1) -> ProperType: ...
-    @overload  # noqa
+    @overload
     @staticmethod
-    def make_union(items: List[Type], line: int = -1, column: int = -1) -> Type: ...  # noqa
+    def make_union(items: List[Type], line: int = -1, column: int = -1) -> Type: ...
 
-    @staticmethod  # noqa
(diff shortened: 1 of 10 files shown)


mypy/types.py
```diff
@@ -1672,10 +1672,10 @@ def __eq__(self, other: object) -> bool:
     def make_union(items: List[ProperType], line: int = -1, column: int = -1) -> ProperType: ...
     @overload  # noqa
     @staticmethod
-    def make_union(items: List[Type], line: int = -1, column: int = -1) -> Type: ...
+    def make_union(items: List[Type], line: int = -1, column: int = -1) -> Type: ...  # noqa
```
JelleZijlstra: These noqas can be removed, right?

ilevkivskyi:
A random idea: What about also running mypyc compiled tests on Python 3.8 (In addition to 3.7)?

msullivan:
I'm not sure if we want to add another full run build?

Issue #7410, opened 2019-08-29
Use colors in mypy output


JukkaL:
Many development tools use color in their output these days, and I think that it's time that mypy did that as well. Arguments for using colors:

* It's faster to scan error messages to find the relevant part.
* Colored messages may subjectively feel/look better even if there is no concrete utility.
* This is becoming a standard, like it or not.

My personal preference is to not go overboard and to use colors in a somewhat restrained manner. Also, I'd prefer to mostly stick to a few colors instead of using all the colors available.

There should be an option to explicitly disable colors. We may also want to disable colors automatically if the output is piped to another process. If we do this, we'd probably need an option to also force colors to be used.

(Originally discussed in #3439 but moved to a separate issue.)

brandtbucher:
I would definitely appreciate a three-color scheme:

- Red, for errors.
- Yellow, for notes.
- Green, for a success message (#7368).

Perhaps something would be needed for actual mypy errors/failure though...

JukkaL:
I think that we don't want to color the entire message to maintain good readability. I think that some tools only color the status code, which may be a reasonable trade-off.

Surveying how/if various popular tools use colors would be interesting (and not only Python tools). Ideas:

* Python linters
* Cython
* Compilers for various languages: TypeScript, Go, Rust, Java, C++
* Linters for other languages

ilevkivskyi:
I like the idea of highlighting only error code (plus the summary line at the end). IIRC this is what Rust does (or used to do) apart from showing the source snippet with some notes. Actually it highlights both the status (i.e. word `error`), but it is different from mypy because the error code follows the status, like `error[EXXXX]`.

Also I would like to highlight the `note:` in light blue or other light color.

gvanrossum:
I'm all for it, as long as it turns itself off if stdout is not a tty, and also if $TERM is `dumb` (or perhaps any value that doesn't have a defined termcap entry).

In terms of color scheme, let's make sure that it's readable both for dark themes and light themes.

I don't have strong preferences about what parts to highlight, just let's not invent the wheel and follow the lead of some other popular tools in the Python ecosystem. Maybe `pylint -f colorized`?

gvanrossum:
Also, let's not get into bikeshedding. Let's just do it.

Issue #7368, opened 2019-08-19
Make mypy more positive


ilevkivskyi:
Maybe someone already noticed this, but IMO mypy is a bit "grumpy". Currently mypy _never_ says anything good to a user. If there are no errors, then the output is just empty. This can even cause some confusion for a new user trying it on a simple clean file.

What do you think about emitting a simple message like `Success: no issues found` if there are no errors/notes emitted during type checking?

Btw, we can also optionally add a summary line if there are errors saying something like `N issues found in M files`. I often want to check how many errors there are and this forces me to re-run mypy with `| grep error | wc -l`.

msullivan:
My gut inclination is for silence in the success case. (Although certainly our internal wrapper scripts are all very noisy?)

ilevkivskyi:
@msullivan 
* Yes, our internal wrappers add a success message, and I didn't hear anyone complaining about this.
* I think an important part of reality that we need to accept is that although mypy was intended as a (relatively) advanced tool, it is in fact used a lot by novice programmers.

JukkaL:
I don't have a strong opinion about this, but at least printing something when there is nothing to report might help new users. It could also be useful to print something like "no issues found in 35 files". One possible failure mode is accidentally checking only a small number of files, when imports aren't being followed. 

I think it would make sense to do a quick survey of what other popular developer tools report on success (both Python and non-Python tools).

JukkaL:
Another reason for preferring a success/summary message is that otherwise multiple mypy runs can be hard to distinguish in a terminal. Here's an actual example:

```
(mypy) jukka mypy $ ./chk.sh
mypy/mypy/checker.py:3797: error: Unexpected keyword argument "code" for "report_protocol_problems" of "MessageBuilder"  [call-arg]
mypy/mypy/messages.py:1135: note: "report_protocol_problems" of "MessageBuilder" defined here
mypy/mypy/checker.py:3801: error: Unexpected keyword argument "code" for "note_call" of "MessageBuilder"  [call-arg]
mypy/mypy/messages.py:1102: note: "note_call" of "MessageBuilder" defined here
mypy/mypy/checker.py:3806: error: Unexpected keyword argument "code" for "note_call" of "MessageBuilder"  [call-arg]
mypy/mypy/messages.py:1102: note: "note_call" of "MessageBuilder" defined here
(mypy) jukka mypy $ ./chk.sh
[...]

PR #7425, opened 2019-08-30
Add support for nicer error output

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7368
Fixes https://github.com/python/mypy/issues/7410

This adds some basic support for output styling on Mac and Linux. The formatter does some basic error message parsing because passing them forward in a structured form would be a bigger change, however if you think this is worth doing I can refactor the PR in a more principled way.

To illustrate the result this is how the output would look:

Screenshot from 2019-08-30 11:22:32

### DIFF
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -400,6 +400,6 @@ class FancyFormatter:
         return self.style('Success: no issues found', 'green', bold=True)
 
     def format_error(self, n_errors: int, n_files: int) -> str:
-        msg = 'Found {} error{} in {} file{}'.format(n_errors, 's' if n_errors > 1 else '',
-                                                     n_files, 's' if n_files > 1 else '')
+        msg = 'Found {} error{} in {} file{}'.format(n_errors, 's' if n_errors != 1 else '',
+                                                     n_files, 's' if n_files != 1 else '')
         return self.style(msg, 'red', bold=True)
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -8,6 +8,12 @@ import sys
 from typing import TypeVar, List, Tuple, Optional, Dict, Sequence, Iterable, Container, IO
 from typing_extensions import Final, Type, Literal
 
+try:
+    import curses
+    CURSES_ENABLED = True
+except ImportError:
+    CURSES_ENABLED = False
+
 T = TypeVar('T')
 
 ENCODING_RE = \
@@ -339,9 +345,7 @@ class FancyFormatter:
             return
 
         # We in a human-facing terminal, check if it supports enough styling.
-        try:
-            import curses
-        except ImportError:
+        if not CURSES_ENABLED:
             self.dummy_term = True
             return
         try:
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -4,7 +4,6 @@ import pathlib
 import re
 import subprocess
 import sys
-import curses
 
 from typing import TypeVar, List, Tuple, Optional, Dict, Sequence, Iterable, Container, IO
 from typing_extensions import Final, Type, Literal
@@ -340,7 +339,17 @@ class FancyFormatter:
             return
 
         # We in a human-facing terminal, check if it supports enough styling.
-        curses.setupterm()
+        try:
+            import curses
+        except ImportError:
+            self.dummy_term = True
+            return
+        try:
+            curses.setupterm()
+        except curses.error:
+            # Most likely terminfo not found.
+            self.dummy_term = True
+            return
         bold = curses.tigetstr('bold')
         setaf = curses.tigetstr('setaf')
         self.dummy_term = not (bold and setaf)
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -332,7 +332,7 @@ def count_stats(errors: List[str]) -> Tuple[int, int]:
 class FancyFormatter:
     """Apply color and bold font to terminal output.
 
-    This currently only works on Linus an Mac.
+    This currently only works on Linux and Mac.
     """
     def __init__(self, f_out: IO[str], f_err: IO[str],
                  show_error_codes: bool) -> None:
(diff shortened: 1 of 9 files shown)


ilevkivskyi:
I just realized that these changes don't affect the daemon. When we will agree here, I will make a separate PR to update the daemon.

JukkaL:
Nice! I'd like to play around a little with this before merging to get a feel of how this works. 

A few quick comments:

* The error code looks too prominent to me. Maybe it should be in gray or cyan, for example? It's probably the least interesting thing in the output.
* It would be interesting to try making the error message bolded/more bright to make it stand out a bit more from the path.
* Maybe "error" should be in brighter red? Now it's a bit dark.
* What about using an emoji such as ✨ for success? Something like "✨ All clean" or something. Emojis in error messages seem to be all the rage these days. (No need to do this in this PR, as this could be a bit tricky, depending on the default encoding etc.)

ilevkivskyi:
@JukkaL OK, I made the changes we discussed privately. Also here is how new output looks on white background.

Screenshot 2019-08-30 16 28 17

gvanrossum:
I like the new look on a light theme. (What does it look now on black theme? Still the same?)

I don't recognize Jukka's emoji; let's not go there (error messages are formal speech, where emoji aren't as accepted as in informal speech).

ilevkivskyi:
> What does it look now on black theme? Still the same?

I want to check on my home Linux machine (Mac and Linux may have slightly different colors). I will post a new screenshot when I am back home.

ilevkivskyi:
Btw I opened https://github.com/mypyc/mypyc/issues/696 for mypyc compilation error.

mypy/util.py
```diff
@@ -313,3 +320,120 @@ def check_python_version(program: str) -> None:
...
+    """Count total number of errors and files in error list."""
+    errors = [e for e in errors if ': error:' in e]
+    files = {e.split(':')[0] for e in errors}
+    return len(errors), len(files)
+
+
+class FancyFormatter:
+    """Apply color and bold font to terminal output.
+
+    This currently only works on Linus an Mac.
```
msullivan: Linux and Mac

mypy/util.py
```diff
@@ -313,3 +320,120 @@ def check_python_version(program: str) -> None:
...
+    """
+    def __init__(self, f_out: IO[str], f_err: IO[str],
+                 show_error_codes: bool) -> None:
+        self.show_error_codes = show_error_codes
+        # Check if we are in a human-facing terminal on a supported platform.
+        if sys.platform not in ('linux', 'darwin'):
+            self.dummy_term = True
+            return
+        if not f_out.isatty() or not f_err.isatty():
```
msullivan: I think we need a flag to suppress the isatty checking, since our internal scripts all wrap mypy for logging purposes.

mypy/util.py
```diff
@@ -313,3 +320,120 @@ def check_python_version(program: str) -> None:
...
+        out = ''
+        for i, part in enumerate(parts):
+            if i % 2 == 0:
+                out += self.style(part, 'none')
+            else:
+                out += self.style('"' + part + '"', 'none', bold=True)
+        return out
+
+    def underline_link(self, note: str) -> str:
+        match = re.search(r'http://\S*', note)
```
msullivan: probably want https too

msullivan:
I love this

mypy/util.py
ilevkivskyi: Yes, indeed. I however added just an environment variable, it is probably relatively rarely needed thing.

ilevkivskyi:
@gvanrossum I think the new version still works well on dark background:

Screenshot from 2019-08-31 00:35:23

gvanrossum:
IMO the contrast for "note" is too low. But presumably it's easy enough to tweak later, so don't worry about it. Or just leave it black?

mypy/util.py
```diff
@@ -5,8 +5,15 @@
 import subprocess
 import sys
 
-from typing import TypeVar, List, Tuple, Optional, Dict, Sequence, Iterable, Container
-from typing_extensions import Final, Type
+from typing import TypeVar, List, Tuple, Optional, Dict, Sequence, Iterable, Container, IO
+from typing_extensions import Final, Type, Literal
+
+try:
+    import curses
```
ethanhs: Instead of using curses which doesn't support Windows well, perhaps we could use [colorama](https://github.com/tartley/colorama) instead? If you don't want to change things in this PR I can open one to redo this in colorama if you want (maybe after the daemon changes).

ilevkivskyi:
@gvanrossum I actually like it this way. I would keep it and see what people say.

@ethanhs I wanted to start from `curses` because it is what is in stdlib. But yeah at some point we can switch to something more advanced (if this is required for Windows). Btw, I have heard `curses` may be available on Windows too but I can't check this. Can you try this PR on Windows (after removing the platform check)?

gvanrossum:
No worries. It turns out my love for this is mostly theoretical, since I usually look at mypy errors through various tools (e.g. the PyCharm plugin or Dropbox CI).

IIUC curses on Windows is a 3rd party package: https://pypi.org/project/windows-curses/.

ethanhs:
So I tested the PR removing the platform check on Windows:
image

I went ahead and tried the windows-curses package and it returned `None` for the `tigetstr` calls, and so it doesn't seem to work, which didn't really surprise me, as I've heard that curses on Windows is rather buggy.

ilevkivskyi:
I opened some follow-up issues. I would propose to merge this soon (today) and then iterate later (if there are no objections).

ssbarnea:
What is the trick to enable colors for mypy as I am running on MacOS? All the other python tools are able to use colors but mypy is still plain.  I did an `import curses` and I got no error, so I doubt is that. Funny, all the CLI options seem to be about how to disable it not how to enforce it.

ethanhs:
@ssbarnea please open a new issue describing what terminal you are using and any other relevant information.

ssbarnea:
On MacOS, I am using iterm2 like most developers and have these vars defined. Shell is zsh. I also tried the default terminal. I doubt any environment variable affects the outcome because I did also run with `env -i /path/to/mypy .` and got the same result.
```
COLORTERM=truecolor
TERM=xterm-256color  # <-- clearly not affecting because same works on Linux
```

PR #7429, opened 2019-08-30
Add error codes to notes to allow ignoring them

### LINKED ISSUE #7384: Error codes in # type: ignore don't suppress related notes

JukkaL:
Associated note should have the same error code as the related
error, so that both will be ignored when the error is ignored.

Also make the default error code for notes to be 'misc', similar
to errors.

As a side effect, error codes for notes are now not shown. I
didn't find any notes which would benefit from having a specific
error code.

Fixes #7384.

Added tests for various notes, but they are not exhaustive.

### DIFF
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -262,7 +262,7 @@ class Errors:
             line: line number of error
             column: column number of error
             message: message to report
-            code: error code (defaults to 'misc' for 'error' severity)
+            code: error code (defaults to 'misc'; not shown for notes)
             blocker: if True, don't continue analysis after this error
             severity: 'error' or 'note'
             file: if non-None, override current file as context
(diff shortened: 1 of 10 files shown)


mypy/errors.py
```diff
@@ -262,7 +262,7 @@ def report(self,
             line: line number of error
             column: column number of error
             message: message to report
-            code: error code (defaults to 'misc' for 'error' severity)
+            code: error code (defaults to 'misc'; not shown for notes)
```
ilevkivskyi: This is a very good choice! Could you please add test (for example with protocols, they generate long notes) checking that codes are not shown on notes? Maybe also mention this in docs.

ilevkivskyi:
> Could you please add test (for example with protocols, they generate long notes) checking that codes are not shown on notes? 

I just found there are already few existing tests for this, so this is not important.

PR #7430, opened 2019-08-30
Be more flexible with spaces in "# type: ignore[code]" comments

JukkaL:
In particular, allow `# type: ignore [code]` (space after `ignore`).

Add tests for various cases (some of which already worked).

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -186,7 +186,7 @@ def parse_type_ignore_tag(tag: Optional[str]) -> List[str]:
     # TODO: Implement proper parsing and error checking
     if not tag:
         return []
-    m = re.match(r'\[([^#]*)\]', tag)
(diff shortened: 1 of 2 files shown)


mypy/fastparse.py
```diff
@@ -186,7 +186,7 @@ def parse_type_ignore_tag(tag: Optional[str]) -> List[str]:
     # TODO: Implement proper parsing and error checking
     if not tag:
         return []
-    m = re.match(r'\[([^#]*)\]', tag)
+    m = re.match(r'\s*\[([^#]*)\]', tag)
```
ilevkivskyi: Ideally we should give a warning if user makes a mistake like `# type: ignore [one-code][other-code]` or something else we don recognize, but I see there is already a TODO about this.

JukkaL: Yes, I'll add better error checking in a separate PR.

Issue #7411, opened 2019-08-29
Print source line with error and underline target of error


JukkaL:
Now that mypy can usually report the column number of errors in a useful manner, we could print out the contents of the source line for each error and underline the token the error points to (based on line/column information). This can be either enabled by default or through a flag -- I'm not sure which option I'd prefer yet.

Arguments for doing this:
* This makes it easier to see what is wrong, especially if the error message happens to be terse, or if there is a lot of things going on in a single line.
* Consistency with Python: Python tracebacks show the source line.
* Not everybody has configured their editor to highlight the location of the error.
* If a user edits the file with the error, the line number will get out of alignment. Having the source line in the output makes it easier to see where the error points to even if there have been edits.
* Many other tools can do this, so many users are used to this behavior.

Arguments for not doing this (at least by default):
* The length of mypy output is increased significantly. If we show a source line and a separate line with underlining, the number of lines may increase by a factor of three.
* The output may be harder to scan or look noisy, since there will be more stuff. Using colors should help with this (#7410).
* If the reported line/column number is not logically correct, this may be confusing. We may want to fix the line/column numbers of most errors first. (I've already done some of this relatively recently.)

Thoughts?

ilevkivskyi:
I think reporting reliably correct line/column numbers is a blocker for this. And it is still not perfect unfortunately.

Also `end_lineno` and `end_col_offset` will simplify finding better spans, but they are Python 3.8+ only (there is even nice [helper](https://docs.python.org/3.8/library/ast.html#ast.get_source_segment)).

gvanrossum:
Well, even if it only shows the line that can sometimes help (e.g. if you're looking at slightly stale mypy output and have trouble matching up the line number with the actual source line in your editor). So I'm not so hung up on correct column numbers as Ivan. Having this will probably motivate users to report cases where the column number (or even the line number) is off.

I would not like this by default, because I often look at long lists of mypy errors where it would just be noise, and in an Emacs `*compilation*` buffer going to the corresponding line is easy.

Interestingly, this is an area where compilers and linters seem to differ -- I know of no linter that shows the text of the line (at least not by default), while all compilers I know of do that. I think mypy is still more like a linter than like a compiler (mypyc notwithstanding :-).

Also, let's not get into bikeshedding. Let's just do it.

ilevkivskyi:
[...]

PR #7440, opened 2019-08-31
Show code snippets on demand

ilevkivskyi:
This essentially fixes https://github.com/python/mypy/issues/7411

Couple ideas I think may make the much larger amount of output more readable are:
* Use dim color for the source code snippet
* Wrap long error messages so that they are not messed up with file names etc.

An example output (dark terminal on Linux):

Screenshot from 2019-08-31 18:26:49

### DIFF
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -416,7 +416,7 @@ class Errors:
                         source_lines: Optional[List[str]]) -> List[str]:
         """Return a string list that represents the error messages.
 
-        Use a form suitable for displaying to the user. If add_snippets
+        Use a form suitable for displaying to the user. If self.pretty
         is True also append a relevant trimmed source code line (only for
         severity 'error').
         """
@@ -443,17 +443,20 @@ class Errors:
             if self.pretty:
                 # Add source code fragment and a location marker.
                 if severity == 'error' and source_lines and line > 0:
+                    source_line = source_lines[line - 1]
+                    if column < 0:
+                        # Something went wrong, take first non-empty column.
+                        column = len(source_line) - len(source_line.lstrip())
                     # Note, currently coloring uses the offset to detect source snippets,
                     # so these offsets should not be arbitrary.
-                    a.append(' ' * DEFAULT_SOURCE_OFFSET + source_lines[line - 1])
+                    a.append(' ' * DEFAULT_SOURCE_OFFSET + source_line)
                     a.append(' ' * (DEFAULT_SOURCE_OFFSET + column) + '^')
         return a
 
     def file_messages(self, path: str) -> List[str]:
         """Return a string list of new error messages from a given file.
 
-        Use a form suitable for displaying to the user. If add_snippets is True,
-        then one must also pass the source code and Python version.
+        Use a form suitable for displaying to the user.
         """
         if path not in self.error_info_map:
             return []
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -3,7 +3,7 @@ import sys
 import traceback
 from collections import OrderedDict, defaultdict
 
-from typing import Tuple, List, TypeVar, Set, Dict, Optional, TextIO
+from typing import Tuple, List, TypeVar, Set, Dict, Optional, TextIO, Callable
 from typing_extensions import Final
 
 from mypy.scope import Scope
@@ -11,11 +11,7 @@ from mypy.options import Options
 from mypy.version import __version__ as mypy_version
 from mypy.errorcodes import ErrorCode
 from mypy import errorcodes as codes
-from mypy.util import (
-    decode_python_encoding, DecodeError, trim_source_line, DEFAULT_SOURCE_OFFSET,
-    WIGGLY_LINE, DEFAULT_COLUMNS, MINIMUM_WIDTH
-)
-from mypy.fscache import FileSystemCache
+from mypy.util import trim_source_line, DEFAULT_SOURCE_OFFSET, DEFAULT_COLUMNS, MINIMUM_WIDTH
 
 T = TypeVar('T')
 allowed_duplicates = ['@overload', 'Got:', 'Expected:']  # type: Final
@@ -158,14 +154,18 @@ class Errors:
     target_module = None  # type: Optional[str]
     scope = None  # type: Optional[Scope]
 
-    def __init__(self, fscache: Optional[FileSystemCache] = None,
-                 options: Optional[Options] = None) -> None:
-        self.show_error_context = options.show_error_context if options else False
-        self.show_column_numbers = options.show_column_numbers if options else False
-        self.show_error_codes = options.show_error_codes if options else False
+    def __init__(self,
+                 show_error_context: bool = False,
+                 show_column_numbers: bool = False,
+                 show_error_codes: bool = False,
+                 show_source_code: bool = False,
+                 read_source: Optional[Callable[[str], Optional[List[str]]]] = None) -> None:
+        self.show_error_context = show_error_context
+        self.show_column_numbers = show_column_numbers
+        self.show_error_codes = show_error_codes
+        self.show_source_code = show_source_code
         # We use fscache to read source code when showing snippets.
-        self.fscache = fscache
-        self.options = options
+        self.read_source = read_source
         self.initialize()
 
     def initialize(self) -> None:
@@ -185,7 +185,11 @@ class Errors:
         self.initialize()
 
     def copy(self) -> 'Errors':
-        new = Errors(self.fscache, self.options)
+        new = Errors(self.show_error_context,
+                     self.show_column_numbers,
+                     self.show_error_codes,
+                     self.show_source_code,
+                     self.read_source)
         new.file = self.file
         new.import_ctx = self.import_ctx[:]
         new.type_name = self.type_name[:]
@@ -419,7 +423,6 @@ class Errors:
         a = []  # type: List[str]
         errors = self.render_messages(self.sort_messages(error_info))
         errors = self.remove_duplicates(errors)
-        add_snippets = self.options and self.options.show_source_code
         for file, line, column, severity, message, code in errors:
             s = ''
             if file is not None:
@@ -432,21 +435,18 @@ class Errors:
                 s = '{}: {}: {}'.format(srcloc, severity, message)
             else:
                 s = message
-            # If we show source code snippet, the error code is shown under it
-            # (with a pointer to error location), instead of after the error message.
-            if self.show_error_codes and code and not add_snippets and severity != 'note':
+            if self.show_error_codes and code and severity != 'note':
                 # If note has an error code, it is related to a previous error. Avoid
                 # displaying duplicate error codes.
                 s = '{}  [{}]'.format(s, code.code)
             a.append(s)
-            if add_snippets:
+            if self.show_source_code:
                 if severity == 'error' and source_lines and line > 0:
                     source_line, offset = trim_source_line(source_lines[line - 1],
                                                            DEFAULT_COLUMNS, column, MINIMUM_WIDTH)
                     a.append(' ' * DEFAULT_SOURCE_OFFSET + source_line)
                     # Also append a marker pointing to the error start location.
-                    a.append(' ' * (DEFAULT_SOURCE_OFFSET + column - offset) + WIGGLY_LINE +
-                             ' [{}]'.format(code.code) if code else '')
+                    a.append(' ' * (DEFAULT_SOURCE_OFFSET + column - offset) + '^')
         return a
 
     def file_messages(self, path: str) -> List[str]:
@@ -459,18 +459,9 @@ class Errors:
             return []
         self.flushed_files.add(path)
         source_lines = None
-        if self.options and self.options.show_source_code:
-            if self.fscache:
-                try:
-                    source = self.fscache.read(path)
-                except OSError:
-                    pass
-                else:
-                    try:
-                        source_lines = decode_python_encoding(
-                            source, self.options.python_version).splitlines()
-                    except DecodeError:
-                        pass
+        if self.show_source_code:
+            assert self.read_source
+            source_lines = self.read_source(path)
         return self.format_messages(self.error_info_map[path], source_lines)
 
     def new_messages(self) -> List[str]:
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -11,7 +11,7 @@ from mypy.options import Options
 from mypy.version import __version__ as mypy_version
 from mypy.errorcodes import ErrorCode
 from mypy import errorcodes as codes
-from mypy.util import trim_source_line, DEFAULT_SOURCE_OFFSET, get_terminal_width, MINIMUM_WIDTH
+from mypy.util import DEFAULT_SOURCE_OFFSET
 
 T = TypeVar('T')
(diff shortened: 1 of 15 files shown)


ilevkivskyi:
I just discovered that this doesn't work well with blocking errors and/or daemon. I have a fix but it requires `mypy.errors` depend on `mypy.fscache`. Which is probably not so bad.

JukkaL:
Some general comments, most of which are copied from #7411:

I think that we should fix at least the most obvious problems with column numbers before documenting this as a feature (it's okay to have it as a hidden option though). I have a half-finished PR that fixes column numbers of arguments and may be able to finish that up soon. Also, assignments should probably point to the rvalue.

Other comments:

*   I don't like the new location for the error code. I'd keep that in the original location after the message.
*   It might look better if we'd only underline the target token, i.e. the length of the underlining would vary. Maybe the underlining should cover a simple member expression (say, underline `foo.bar` in `foo.bar.zar`).
*   If the column number is missing, we should perhaps point to the first non-space token on the line. There should be a test for this.
* The indentation when splitting error message may be too much, since the path can be very long. Maybe use a fixed level of indent (8 spaces?).

JukkaL:
> I have a fix but it requires mypy.errors depend on mypy.fscache.

Interesting, why does is need `fscache`?

mypy/errors.py
```diff
@@ -11,6 +11,11 @@
 from mypy.version import __version__ as mypy_version
 from mypy.errorcodes import ErrorCode
 from mypy import errorcodes as codes
+from mypy.util import (
+    decode_python_encoding, DecodeError, trim_source_line, DEFAULT_SOURCE_OFFSET,
+    WIGGLY_LINE, DEFAULT_COLUMNS, MINIMUM_WIDTH
+)
+from mypy.fscache import FileSystemCache
```
JukkaL: I think it would be nice if there was no import from `mypy.fscache` in this file, and instead we'd accept a callback that reads file contents, for example (dependency injection). We could probably also get rid of the `decode_python_encoding` and `DecodeError` imports.

mypy/errors.py
```diff
@@ -153,13 +158,14 @@ class Errors:
...
 
-    def __init__(self,
-                 show_error_context: bool = False,
-                 show_column_numbers: bool = False,
-                 show_error_codes: bool = False) -> None:
-        self.show_error_context = show_error_context
-        self.show_column_numbers = show_column_numbers
-        self.show_error_codes = show_error_codes
+    def __init__(self, fscache: Optional[FileSystemCache] = None,
+                 options: Optional[Options] = None) -> None:
```
JukkaL: This adds a big dependency, which we previously avoided by providing the required information explicitly. It would arguably be cleaner if you'd just add the required options as arguments here instead of passing the whole `Options` object.

mypy/errors.py
```diff
@@ -422,22 +432,46 @@ def format_messages(self, error_info: List[ErrorInfo]) -> List[str]:
...
+        then one must also pass the source code and Python version.
         """
         if path not in self.error_info_map:
             return []
         self.flushed_files.add(path)
-        return self.format_messages(self.error_info_map[path])
+        source_lines = None
+        if self.options and self.options.show_source_code:
+            if self.fscache:
+                try:
```
JukkaL: This try statement could be provided as a callback to avoid a depending on `fscache`.

mypy/util.py
```diff
@@ -20,6 +21,15 @@
 ENCODING_RE = \
     re.compile(br'([ \t\v]*#.*(\r\n?|\n))??[ \t\v]*#.*coding[:=][ \t]*([-\w.]+)')  # type: Final
 
+PLAIN_ANSI_DIM = '\x1b[2m'  # type: Final
```
JukkaL: Is this portable? Could we get this from `curses` or something?

mypy/util.py
```diff
@@ -110,6 +119,24 @@ def decode_python_encoding(source: bytes, pyversion: Tuple[int, int]) -> str:
     return source_text
 
 
+def trim_source_line(line: str, max_len: int, col: int, min_width: int) -> Tuple[str, int]:
```
JukkaL: This function is a great candidate for regular unit tests.

mypy/util.py
```diff
@@ -329,14 +356,64 @@ def count_stats(errors: List[str]) -> Tuple[int, int]:
     return len(errors), len(files)
 
 
+def split_words(msg: str) -> List[str]:
```
JukkaL: This is another function that would benefit from plain unit tests.

ilevkivskyi:
@JukkaL 
> I don't like the new location for the error code. I'd keep that in the original location after the message.

I actually like the new position more, since we are already taking much more vertical space, why not save a bit of horizontal space? Otherwise the line with wiggly arrow will have few information on it.
> It might look better if we'd only underline the target token, i.e. the length of the underlining would vary.

It is hard to define what is "token". It seems to me you don't realize how hard would be implementing this to a reasonable degree of reliability and how misleading the positions can be if we will do this in ad-hoc manner. Like we may end up underlining the first item of a tuple while the problem is actually in the type of the second one. I would rather focus on making the start column better first.
> If the column number is missing, we should perhaps point to the first non-space token on the line. There should be a test for this.

OK.
> The indentation when splitting error message may be too much, since the path can be very long. Maybe use a fixed level of indent (8 spaces?).

OK.
> Interesting, why does is need `fscache`?

I want to pass the cache as an argument to the `Errors` constructor (literally once), rather that "threading" the error line content to the every place it is needed. Therefore I need to annotate the constructor argument. I however just realized I can use a tiny protocol.

mypy/util.py
```diff
@@ -329,14 +356,64 @@ def count_stats(errors: List[str]) -> Tuple[int, int]:
...
+
+def get_term_columns() -> int:
+    try:
+        cols, _ = os.get_terminal_size()
+        return cols
+    except OSError:
+        return DEFAULT_COLUMNS
+
+
+def soft_wrap(msg: str, max_len: int, pad: int) -> str:
```
JukkaL: Maybe rename `pad` to `indent` or `num_indent`?

mypy/util.py
```diff
@@ -384,13 +467,21 @@ def style(self, text: str, color: Literal['red', 'green', 'blue', 'yellow', 'non
...
         return start + self.colors[color] + text + self.NORMAL
 
     def colorize(self, error: str) -> str:
         """Colorize an output line by highlighting the status and error code."""
         if ': error:' in error:
             loc, msg = error.split('error:', maxsplit=1)
-            if not self.show_error_codes:
+            if self.show_source_code:
+                # Improve readability by wrapping lines when showing source code.
+                pad = len(loc) + len('error: ')
```
JukkaL: This can be too high if the path the long. A fixed number such 8 might be best, or falling back to a fixed smaller number if the path is long enough (not sure which would be better).

mypy/util.py
```diff
@@ -384,13 +467,21 @@ def style(self, text: str, color: Literal['red', 'green', 'blue', 'yellow', 'non
...
+            start += self.DIM
         return start + self.colors[color] + text + self.NORMAL
 
     def colorize(self, error: str) -> str:
         """Colorize an output line by highlighting the status and error code."""
         if ': error:' in error:
             loc, msg = error.split('error:', maxsplit=1)
-            if not self.show_error_codes:
+            if self.show_source_code:
+                # Improve readability by wrapping lines when showing source code.
```
JukkaL: It's can be surprising that showing source code also affects how long messages are wrapped. Maybe the option should be named `--pretty` or something, as it controls a few things.

mypy/util.py
```diff
@@ -401,10 +492,18 @@ def colorize(self, error: str) -> str:
         elif ': note:' in error:
             loc, msg = error.split('note:', maxsplit=1)
             return loc + self.style('note:', 'blue') + self.underline_link(msg)
+        elif self.show_source_code and error.startswith(' ' * DEFAULT_SOURCE_OFFSET):
```
JukkaL: Detecting source code highlights through an indent prefix can be surprising. I wonder if there's a more general way to do this. If there's no alternative simple approach, this should be documented carefully somewhere.

mypy/util.py
```diff
@@ -329,14 +356,64 @@ def count_stats(errors: List[str]) -> Tuple[int, int]:
...
+            next_word = ''
+            continue
+        if c == '"':
+            allow_break = not allow_break
+        next_word += c
+    res.append(next_word)
+    return res
+
+
+def get_term_columns() -> int:
```
JukkaL: What about renaming this to `get_terminal_width`?

mypy/errors.py
```diff
@@ -422,22 +432,46 @@ def format_messages(self, error_info: List[ErrorInfo]) -> List[str]:
...
+            # (with a pointer to error location), instead of after the error message.
+            if self.show_error_codes and code and not add_snippets and severity != 'note':
                 # If note has an error code, it is related to a previous error. Avoid
                 # displaying duplicate error codes.
                 s = '{}  [{}]'.format(s, code.code)
             a.append(s)
+            if add_snippets:
+                if severity == 'error' and source_lines and line > 0:
+                    source_line, offset = trim_source_line(source_lines[line - 1],
+                                                           DEFAULT_COLUMNS, column, MINIMUM_WIDTH)
```
JukkaL: Should this use the terminal width?

mypy/util.py
```diff
@@ -384,13 +467,21 @@ def style(self, text: str, color: Literal['red', 'green', 'blue', 'yellow', 'non
...
 
     def colorize(self, error: str) -> str:
         """Colorize an output line by highlighting the status and error code."""
         if ': error:' in error:
             loc, msg = error.split('error:', maxsplit=1)
-            if not self.show_error_codes:
+            if self.show_source_code:
+                # Improve readability by wrapping lines when showing source code.
+                pad = len(loc) + len('error: ')
+                max_len = get_term_columns() - pad - 1  # compensate for space after 'error:'
```
JukkaL: This is kind of suspicious -- how will this work in the daemon? Should the client pass the terminal width to the server as part of each request, since the terminal may change between invocations?

Anyway, it might be better if this was provided by the caller.

mypy/errors.py
ilevkivskyi: OK, I was thinking about a protocol, but callback is even better.

mypy/errors.py
ilevkivskyi: OK, I will split this back.

mypy/util.py
ilevkivskyi: I tried it in couple terminals and it works (because it is ANSI standard). The problem is that although it is a basic ANSI "feature", terminfo files for most default terminals don't have `dim` termcap entry, so `curses` doesn't report it. I was thinking about choosing a grey color that would look good on both white and black background, but it is actually not easy, and again most default terminals are 8-color, not 256-color, so we can't get the color code from `curses`.

mypy/util.py
ilevkivskyi: > This is kind of suspicious -- how will this work in the daemon?

I was going to fix this after the other PR lands. Essentially yes, this should be passed together with `is_tty` from the other PR.

ilevkivskyi:
@JukkaL thanks for review, I agree with all comments I didn't reply above.

JukkaL:
> I actually like the new position more, since we are already taking much more vertical space, why not save a bit of horizontal space? Otherwise the line with wiggly arrow will have few information on it.

I think it's okay that the wiggly arrow is on its own line without anything else. The savings in horizontal space will be pretty minor, since error codes tend to be short.

Here's some reasoning why I'd rather not add the error code after the wiggly arrow (besides that it subjectively looks odd to me):
* It makes the error code more prominent, even though most of the time it's the least interesting bit of information in the output. Now it kind of jumps at you, since it's the only alpha-numeric information on the line with the wiggly arrow.
* It arguably changed the error formatting more than is necessary from the default mode. If we keep the error code in the original location, the new mode primarily adds some extra lines to the output but otherwise doesn't change things, which I like since it's simple and consistent. (Okay, it also does line wrapping, but that can be argued for on the basis of making it easier to see where the source line starts.)
* Other tools tend to have the underlining/caret alone on a separate line, at least most of the time, so this seems to be a popular way of doing things. We don't need to follow what others do, but if we are unsure about something, falling back to common conventions may be a reasonable default.
* We might want to put some other information in the location after/below the underlining in the future. See https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html for some relevant Rust examples. Clang can also add some notes near the underlining (https://clang.llvm.org/diagnostics.html). I'd prefer to leave the space empty for future experiments. We could always move the error code back in the future, but it's a bit awkward if we have to move back and forth on an issue.

> It is hard to define what is "token". [...]

Yes, there is a risk of doing something confusing. However, I think that the current approach is also confusing sometimes, especially if the error is at the end of the line, or if the target is a name that is much longer/shorter than the squiggle. I initially thought that it underlines a span of code, and was confused when this wasn't the case. A conservative option would be to only show a caret (`^`) and no squiggles, similar to what gcc does (or did), in the first implementation. We can always iterate on this later.

Here are other heuristics that could be built on top of the "caret only" approach that seem reasonably safe bets to me (but I haven't thought about this very carefully):
* If the target is an identifier or a keyword, underline the identifier/keyword.
* If the target starts an int/float/string/bytes literal, underline the literal.
* Otherwise just show a caret.

As long as we can't always find the end span of an AST node, we may want to allow some extra options to be passed along with error messages that may affect how the length of the underlining is determined. For example, if the target expression is an attribute expression `foo.bar`, we can pass this as a flag, and the highlighter can try to find such an expression at the location using some heuristics. Yes, this will get increasingly hard to do if we want this to work always.

mypy/util.py
JukkaL: Ok, sounds reasonable. Maybe add a comment about this?

Since this is not on by default, we can iterate on this afterwards even if some things don't work in every possible environment.

ilevkivskyi:
After addressing some comments it looks like this. I will continue tomorrow.

Screenshot 2019-09-02 22 07 47

JukkaL:
Thanks for the updates! I like how it looks now.

mypy/util.py
ilevkivskyi: After some more experimentation it seems to me having zero indent here is the safest bet. Often there will be just one or few wrapped lines, and having them indented by some non-zero fixed value doesn't create a visual vertical line, so it looks a bit random. The wrapping however still makes sense so that words and types are not accidentally split.

ethanhs:
Wow! This looks really good. Thank you for working on it Ivan!

ilevkivskyi:
@ethanhs I am glad you like it :-)

@JukkaL This is ready for another review.

ilevkivskyi:
Opened follow-up issues https://github.com/python/mypy/issues/7453 and https://github.com/python/mypy/issues/7454

mypy/util.py
```diff
@@ -329,13 +396,75 @@ def count_stats(errors: List[str]) -> Tuple[int, int]:
...
+
+    For example:
+               first_offset
+        ------------------------
+        path/to/file: error: 58: Some very long error message
+            that needs to be split in separate lines.
+            "Long[Type, Names]" are never split.
+        ^^^^--------------------------------------------------
+        num_indent           max_len
+    """
```
JukkaL: I like the detailed comment!

PR #7444, opened 2019-09-02
Handle class with tuple base as TypeVar bound

### LINKED ISSUE #7427: Using a class inherited from tuple as a bound in TypeVar
I think I've encountered mypy bug. Here's the code I'm trying to check with mypy (sorry if it's big, I think additional code helps debugging):
```python
from typing import NamedTuple, Type, TypeVar

# case 1 - inheriting from custom class, mypy check passes.
T1 = TypeVar("T1", bound="B")

class A:
    pass

class B(A):
    @classmethod
    def method(cls) -> None:
        print("B class method")

    @classmethod
    def factory(cls: Type[T1]) -> T1:
        cls.method()
        return cls()

# case 2 - inheriting from named tuple, mypy check fails.
T2 = TypeVar("T2", bound="C")

class C(NamedTuple):
    @classmethod
    def method(cls) -> None:
        print("C class method")

    @classmethod
    def factory(cls: Type[T2]) -> T2:
        cls.method()  # error: "Type[T2]" has no attribute "method"
        return cls()
```

```bash
$ mypy test.py
test.py:29: error: "Type[T2]" has no attribute "method"
```
  **no errors**
  **python 3.7.3, mypy both `0.720` and `0.730+dev.811c201551ecb151f83e982da3ad8f46fa2ebe84`**
* What are the mypy flags you are using?
  **no flags**

I've been able to track the error till [this line](https://github.com/python/mypy/blob/811c201551ecb151f83e982da3ad8f46fa2ebe84/mypy/checkmember.py#L261) `mypy/checkmember.py#L261`. In first case `upper_bound` is `test.B`, but in second it's `Tuple[, fallback=test.C]` and fallback is not checked. I'm not sure how to proceed from here though because it's probably not safe to always check a fallback.

ateraz:
Fixes #7427 
PR fixes using classes with tuple base as bound of `TypeVar`

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -261,6 +261,8 @@ def analyze_type_type_member_access(name: str,
         upper_bound = get_proper_type(typ.item.upper_bound)
         if isinstance(upper_bound, Instance):
             item = upper_bound
+        elif isinstance(upper_bound, TupleType):
+            item = upper_bound.partial_fallback
     elif isinstance(typ.item, TupleType):
         item = tuple_fallback(typ.item)
     elif isinstance(typ.item, FunctionLike) and typ.item.is_type_obj():
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -262,7 +262,7 @@ def analyze_type_type_member_access(name: str,
         if isinstance(upper_bound, Instance):
             item = upper_bound
         elif isinstance(upper_bound, TupleType):
-            item = upper_bound.partial_fallback
+            item = tuple_fallback(upper_bound)
     elif isinstance(typ.item, TupleType):
         item = tuple_fallback(typ.item)
     elif isinstance(typ.item, FunctionLike) and typ.item.is_type_obj():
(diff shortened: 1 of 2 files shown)


mypy/checkmember.py
```diff
@@ -261,6 +261,8 @@ def analyze_type_type_member_access(name: str,
         upper_bound = get_proper_type(typ.item.upper_bound)
         if isinstance(upper_bound, Instance):
             item = upper_bound
+        elif isinstance(upper_bound, TupleType):
+            item = upper_bound.partial_fallback
```
ilevkivskyi: I think it will be safer to use `tuple_fallback(upper_bound)` here.

test-data/unit/check-bound.test
```diff
@@ -169,6 +169,18 @@ def f(x: T) -> int:
...
 
+[case testBoundClassMethodWithNamedTupleBase]
+from typing import NamedTuple, Type, TypeVar
+class A(NamedTuple):
+    @classmethod
+    def foo(cls) -> None: ...
+
+T = TypeVar('T', bound=A)
+def f(x: Type[T]) -> None:
+    x.foo()
```
ilevkivskyi: Could you please use `reveal_type(x.foo)` to check that the type is correct?

ateraz:
@ilevkivskyi Thanks for your feedback, I've updated PR to incorporate it. Could you please check once again?

PR #7446, opened 2019-09-02
Fix crash related with __getattr__ and __call__

### LINKED ISSUE #7243: __getattr__ returning self type results in RecursionError in overloaded contexts
The following code (`foo.pyi`):

```python
from typing import Any, Callable, overload

class T:
    def __getattr__(self, item: Any) -> T: ...

@overload
def f(f: Callable[..., Any], t: str = ...) -> Callable[...,  T]: ...
@overload
def f(f: str = ...) -> Callable[[Callable[..., Any]], Callable[..., T]]: ...
```

when used with current master

```
$ mypy --version                      
mypy 0.730+dev.aaad048cf6f88280f4295865882888b7721dacdb
```

with default options 

```
mypy --show-traceback foo.py
```
results in `RecursionError`

```
  File "/home/zero323/anaconda3/bin/mypy", line 10, in <module>
    sys.exit(console_entry())
...
RecursionError: maximum recursion depth exceeded while calling a Python object
foo.pyi:6: : note: use --pdb to drop into pdb
```

with repeated following calls:

```
  File "/path/to/anaconda3/lib/python3.7/site-packages/mypy/subtypes.py", line 152, in _is_subtype
    ignore_promotions=self.ignore_promotions)
  File "/path/to/anaconda3/lib/python3.7/site-packages/mypy/subtypes.py", line 95, in is_subtype
    ignore_promotions=ignore_promotions))
  File "/path/to/anaconda3/lib/python3.7/site-packages/mypy/types.py", line 667, in accept
  File "/path/to/anaconda3/lib/python3.7/site-packages/mypy/subtypes.py", line 232, in visit_instance
```

The problem disappears if:

- We remove `__getattr__`. Other dunders with the same signature, for example

      class T:
          def __getitem__(self, item: Any) -> T: ...

  seem to work fine.

- We change return type of `__getattr__`, for example:

      class T:
           def __getattr__(self, item: Any) -> str: ...

- We remove any of the overloaded `f`. Either 

      def f(f: Callable[..., Any], t: str = ...) -> Callable[...,  T]: ...

  or

      def f(f: str = ...) -> Callable[[Callable[..., Any]], Callable[..., T]]: ...

  alone, works just fine.

- We change return type of the first overloaded `f`, for example to:

      def f(f: Callable[..., Any], t: str = ...) -> Callable[...,  int]: ...

  or even other `Generic` of `T`:

      def f(f: Callable[..., Any], t: str = ...) -> List[T]: ..

- We make `t` obligatory in both overloaded definitions, or remove it completely from the second one.

The same code causes segmentation fault with the latest PyPi release:

```
$ mypy --show-traceback foo.pyi
[1]    1530 segmentation fault  mypy --show-traceback foo.pyi
```

so it might be (?) related to #6579.

onlined:
Fixes #7243.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3801,12 +3801,12 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                 isinstance(subtype, (Instance, TupleType, TypedDictType))):
             self.msg.report_protocol_problems(subtype, supertype, context, code=code)
         if isinstance(supertype, CallableType) and isinstance(subtype, Instance):
-            call = find_member('__call__', subtype, subtype)
+            call = find_member('__call__', subtype, subtype, is_operator=True)
             if call:
                 self.msg.note_call(subtype, call, context, code=code)
         if isinstance(subtype, (CallableType, Overloaded)) and isinstance(supertype, Instance):
             if supertype.type.is_protocol and supertype.type.protocol_members == ['__call__']:
-                call = find_member('__call__', supertype, subtype)
+                call = find_member('__call__', supertype, subtype, is_operator=True)
                 assert call is not None
                 self.msg.note_call(supertype, call, context, code=code)
         return False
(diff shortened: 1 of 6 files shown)


onlined:
Thanks for the review! I changed the logic as you proposed. I will add tests, but I couldn't think of some simple snippet which can call `find_member` with `__call__`. Could you give a suggestion?

ilevkivskyi:
I think the simplest way to trigger the crash is:
```python
from typing import Callable, Any

class C:
    def __getattr__(self, attr: str) -> C: ...

def do(cd: Callable[..., Any]) -> None:
    ...

do(C())
```
Using this I think you get the idea how to trigger similar crashes in `constraints.py`, `join.py`, etc. There is no need to add a test for every call site, just couple test cases will be enough.

onlined:
Added.

Issue #7439, opened 2019-08-31
Document how to suppress colorized output and summary line


ilevkivskyi:
We need to document `--no-coror-output` and `--no-error-summary`.

This is a follow-up for https://github.com/python/mypy/pull/7425

tapaswenipathak:
Hi @ilevkivskyi: If you don't mind, can I PR?

ilevkivskyi:
Yes, please go ahead!

PR #7448, opened 2019-09-03
Document how to suppress colorized output and summary line

tapaswenipathak:
Fixes https://github.com/python/mypy/issues/7439.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -449,6 +449,22 @@ in error messages.
 
         main.py:12:9: error: Unsupported operand types for / ("int" and "str")
 
+``--no-color-output``
+    This flag will disable color output in error messages, enabled by default.
+
+``--error-summary``
+    This flag will show error messages with stats summary,
+    for example, the following indicates an error summary::
+        from typing import List
+        x = []  # type: List[float]
+        y = []  # type: List[int]
+        x = y
+        bad.py:4: error: Incompatible types in assignment (expression has type
+        "List[int]", variable has type "List[float]")
+        bad.py:4: note: "List" is invariant -- see
+        http://mypy.readthedocs.io/en/latest/common_issues.html#variance
+        bad.py:4: note: Consider using "Sequence" instead, which is covariant
+        Found 1 error in 1 file (checked 1 source file)
 
 .. _incremental:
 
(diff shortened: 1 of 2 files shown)


tapaswenipathak:
@ilevkivskyi: done.

docs/source/config_file.rst
```diff
@@ -401,6 +401,11 @@ section of the command line docs.
 ``show_column_numbers`` (bool, default False)
     Shows column numbers in error messages.
 
+``color_output`` (bool, default True)
+    Shows error messages with color enabled.
+
+``error_summary`` (bool, default True)
+    Shows error messages with any stats summary.
```
ilevkivskyi:
```suggestion
    Shows a short summary line after error messages.
```

docs/source/command_line.rst
```diff
@@ -449,6 +449,22 @@ in error messages.
 
         main.py:12:9: error: Unsupported operand types for / ("int" and "str")
 
+``--no-color-output``
+    This flag will disable color output in error messages, enabled by default.
+
+``--error-summary``
```
ilevkivskyi: I would document `--no-error-summary` (as you did above), since it is also enabled by default.

docs/source/command_line.rst
```diff
@@ -449,6 +449,22 @@ in error messages.
...
+        from typing import List
+        x = []  # type: List[float]
+        y = []  # type: List[int]
+        x = y
+        bad.py:4: error: Incompatible types in assignment (expression has type
+        "List[int]", variable has type "List[float]")
+        bad.py:4: note: "List" is invariant -- see
+        http://mypy.readthedocs.io/en/latest/common_issues.html#variance
+        bad.py:4: note: Consider using "Sequence" instead, which is covariant
+        Found 1 error in 1 file (checked 1 source file)
```
ilevkivskyi: Instead of this long snippet,  I would just say something like "By default mypy shows a summary line including total number of errors, number of files with errors, and number of files checked".

docs/source/config_file.rst
ilevkivskyi: This was not applied, because after committing this you didn't pull and instead force pushed.

tapaswenipathak: Uh, sorry.

PR #7458, opened 2019-09-04
Be more aggressive about site-packages in MYPYPATH

ilevkivskyi:
Fixes https://github.com/dropbox/sqlalchemy-stubs/issues/77

Alternatively (if this looks too aggressive), we can at least remove `os.path.sep` when performing the check.

### DIFF
--- a/mypy/modulefinder.py
+++ b/mypy/modulefinder.py
@@ -493,8 +493,10 @@ def compute_search_paths(sources: List[BuildSource],
     egg_dirs, site_packages = get_site_packages_dirs(options.python_executable)
     for site_dir in site_packages:
         assert site_dir not in lib_path
-        if site_dir in mypypath:
+        if any(p.startswith(site_dir) for p in mypypath):
             print("{} is in the MYPYPATH. Please remove it.".format(site_dir), file=sys.stderr)
+            print("See https://mypy.readthedocs.io/en/latest/running_mypy.html"
+                  "#how-mypy-handles-imports for more info", file=sys.stderr)
             sys.exit(1)
         elif site_dir in python_path:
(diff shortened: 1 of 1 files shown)


mypy/modulefinder.py
```diff
@@ -493,8 +493,10 @@ def compute_search_paths(sources: List[BuildSource],
     egg_dirs, site_packages = get_site_packages_dirs(options.python_executable)
     for site_dir in site_packages:
         assert site_dir not in lib_path
-        if site_dir in mypypath:
+        if any(p.startswith(site_dir) for p in mypypath):
```
JukkaL: If site_dir is `/foo/bar`, should we allow things like `/foo/bar2` in the mypy path? I.e., maybe we should require that there is a path separator after the site_dir.

ilevkivskyi: OK, makes sense.

ilevkivskyi: The new version disables `/foo/bar`, `/foo/bar/`, `/foo/bar/baz`, but allows `/foo/bar2`.

mypy/modulefinder.py
```diff
@@ -493,8 +493,10 @@ def compute_search_paths(sources: List[BuildSource],
     egg_dirs, site_packages = get_site_packages_dirs(options.python_executable)
     for site_dir in site_packages:
         assert site_dir not in lib_path
-        if site_dir in mypypath:
+        if site_dir in mypypath or any(p.startswith(site_dir + os.path.sep) for p in mypypath):
```
JukkaL: Hmm on Windows we may want to also support `os.path.altsep`.

Issue #7412, opened 2019-08-29
Exception while typechecking: KeyError: 'typing'


corydolphin:
Hello,

First, thank you all for your work on MyPy, adding static typing to Python is a huge boon.

MyPy is hitting an exception while typechecking our project, both with version `0.720` and with master `0.730+dev.8954045912e8d7f00af5d7d6289085dd8eb8c64a` on Python 3.6.5

The minimal test case which triggers the exception is: 
```Python
import contextlib


@contextlib.contextmanager
def do_nothing(*args, **kwargs):
    yield
```

When evaluated with a config file which contains `follow_imports_for_stubs = True`, as well as passing `--follow-imports=skip` in the commandline. Are these two incompatible? Their names almost suggest as much. 

E.g.
```
[mypy]
follow_imports_for_stubs = True
```

```bash
python -m mypy --show-traceback --follow-imports=skip --config-file=example.cfg example.py
example.py:4: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 0.730+dev.8954045912e8d7f00af5d7d6289085dd8eb8c64a
Traceback (most recent call last):
  File "/var/lib/conda/envs/py35/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/var/lib/conda/envs/py35/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/var/lib/conda/envs/py35/lib/python3.6/site-packages/mypy/__main__.py", line 12, in <module>
    main(None, sys.stdout, sys.stderr)
  File "/var/lib/conda/envs/py35/lib/python3.6/site-packages/mypy/main.py", line 83, in main
    res = build.build(sources, options, None, flush_errors, fscache, stdout, stderr)
  File "/var/lib/conda/envs/py35/lib/python3.6/site-packages/mypy/build.py", line 163, in build
    result = _build(sources, options, alt_lib_path, flush_errors, fscache, stdout, stderr)
  File "/var/lib/conda/envs/py35/lib/python3.6/site-packages/mypy/build.py", line 225, in _build
    graph = dispatch(sources, manager, stdout)
  File "/var/lib/conda/envs/py35/lib/python3.6/site-packages/mypy/build.py", line 2562, in dispatch
    process_graph(graph, manager)
[...]

PR #7459, opened 2019-09-04
Fix 7412: Exception while typechecking: KeyError: 'typing'

gantsevdenis:
fix for https://github.com/python/mypy/issues/7412 looks pretty simple... not sure if this solution is general enough


mypy/build.py
```diff
@@ -2303,6 +2303,7 @@ def find_module_and_diagnose(manager: BuildManager,
         if (root_source  # Honor top-level modules
                 or (not path.endswith('.py')  # Stubs are always normal
                     and not options.follow_imports_for_stubs)  # except when they aren't
+                or id == 'typing'  # typing is not a stub, and is always normal
```
ilevkivskyi: I think this makes sense, to make it more general may be use `id not in mypy.semanal_main.core_modules` instead of this line and the below one?

PR #7460, opened 2019-09-04
Check syntax errors in "# type: ignore[code]" comments

JukkaL:


test-data/unit/check-errorcodes.test
```diff
@@ -593,3 +642,19 @@ class A:
...
+main:1: error: Name 'y' is not defined  [name-defined]
+main:2: error: Name 'ignored' is not defined  [name-defined]
+main:2: error: Name 'y' is not defined  [name-defined]
+
+[case testErrorCodeTypeIgnoreMisspelled2]
+x = y  # type: int  # type: ignored[foo]
+x = y  # type: int  # type: ignored [foo]
+[out]
+main:1: error: syntax error in type comment 'int'  [syntax]
+main:2: error: syntax error in type comment 'int'  [syntax]
```
ilevkivskyi: TBH this error is confusing. Is it hard to fix this?

ilevkivskyi:
Also wow, you just segfaulted gcc, we should probably report a bug.

JukkaL:
It's not clear if the gcc segfault is reproducible. It happened in code that I didn't touch in this PR.

PR #7463, opened 2019-09-04
Enable output color and effects on Windows

### LINKED ISSUE #7437: Support colorized output on Windows
This is a follow-up for https://github.com/python/mypy/pull/7425

ethanhs:
image

It turns out it is relatively simple to do this on Windows 10 without dependencies. Note that we can get color output on Windows 7/8 but not really underlining or other VT100 features. This should be supported on all versions of Windows 10 people are running. I'm not sure Windows 7/8 support is worth it here. Windows 7 is EOL at the end of the year, and Windows 8.1 has only 5% market share.

Fixes #7437

### DIFF
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -338,38 +338,61 @@ class FancyFormatter:
                  show_error_codes: bool) -> None:
         self.show_error_codes = show_error_codes
         # Check if we are in a human-facing terminal on a supported platform.
-        if sys.platform not in ('linux', 'darwin'):
+        if sys.platform not in ('linux', 'darwin', 'win32'):
             self.dummy_term = True
             return
         force_color = int(os.getenv('MYPY_FORCE_COLOR', '0'))
         if not force_color and (not f_out.isatty() or not f_err.isatty()):
             self.dummy_term = True
             return
+        if sys.platform == 'win32':
+            # Windows ANSI escape sequences are only supported on Threshold 2 and above.
+            if sys.getwindowsversion().build < 10586:
+                self.dummy_term = True
+                return
+            self.dummy_term = False
+            import ctypes
+            kernel32 = ctypes.windll.kernel32
+            ENABLE_PROCESSED_OUTPUT = 0x1
+            ENABLE_WRAP_AT_EOL_OUTPUT = 0x2
+            ENABLE_VIRTUAL_TERMINAL_PROCESSING = 0x4
+            STD_OUTPUT_HANDLE = -11
+            kernel32.SetConsoleMode(kernel32.GetStdHandle(STD_OUTPUT_HANDLE),
+                                                                ENABLE_PROCESSED_OUTPUT
+                                                                | ENABLE_WRAP_AT_EOL_OUTPUT
+                                                                | ENABLE_VIRTUAL_TERMINAL_PROCESSING)
+            self.BOLD = '\033[1m'
+            self.UNDER = '\033[4m'
+            self.BLUE = '\033[94m'
+            self.GREEN = '\033[92m'
+            self.RED = '\033[91m'
+            self.YELLOW = '\033[93m'
+            self.NORMAL = '\033[0m'
+        else:
+            # We in a human-facing terminal, check if it supports enough styling.
+            if not CURSES_ENABLED:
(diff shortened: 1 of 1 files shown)


mypy/util.py
```diff
@@ -467,43 +467,66 @@ class FancyFormatter:
...
-        self.DIM = dim.decode() if dim else PLAIN_ANSI_DIM
-
-        self.BLUE = curses.tparm(set_color, curses.COLOR_BLUE).decode()
-        self.GREEN = curses.tparm(set_color, curses.COLOR_GREEN).decode()
-        self.RED = curses.tparm(set_color, curses.COLOR_RED).decode()
-        self.YELLOW = curses.tparm(set_color, curses.COLOR_YELLOW).decode()
-
+        if sys.platform == 'win32':
+            # Windows ANSI escape sequences are only supported on Threshold 2 and above.
+            if sys.getwindowsversion().build < 10586:
```
ilevkivskyi: I would move this constant (10586) to the top of file, and add a bit more info about this.

mypy/util.py
```diff
@@ -467,43 +467,66 @@ class FancyFormatter:
...
+            self.BOLD = bold.decode()
+            self.UNDER = under.decode()
+            dim = curses.tigetstr('dim')
+            # TODO: more reliable way to get gray color good for both dark and light schemes.
+            self.DIM = dim.decode() if dim else PLAIN_ANSI_DIM
+
+            self.BLUE = curses.tparm(set_color, curses.COLOR_BLUE).decode()
+            self.GREEN = curses.tparm(set_color, curses.COLOR_GREEN).decode()
+            self.RED = curses.tparm(set_color, curses.COLOR_RED).decode()
+            self.YELLOW = curses.tparm(set_color, curses.COLOR_YELLOW).decode()
```
ilevkivskyi: An idea for refactoring: move the blocks inside `if` and `else` to two methods: `initialize_win_colors()` and `initialize_unix_colors()`?

ethanhs:
Here is the daemon in action. I also tested and the terminal width information does indeed work.

image

I've gone ahead and followed your suggestions, hopefully the background I gave is enough (I also added a check for major version). Let me know if you have further feedback/suggestions.

ethanhs:
Huh, that's concerning. I'm not sure why the daemon timeouts happened. I triggered a new build so hopefully that was a fluke (but I will think about why that could happen).

ilevkivskyi:
The problem may be because if `initialize_unix_colors()` exits early the color attributes are not initialized actually, so it will crash right after in `__init__()`. I think both these methods should return `bool`, and then the called should set `dummy_terminal` to the result (and return soon if `False`).

mypy/util.py
```diff
@@ -467,30 +474,63 @@ class FancyFormatter:
...
+        if sys.platform not in ('linux', 'darwin', 'win32'):
             self.dummy_term = True
             return
         force_color = int(os.getenv('MYPY_FORCE_COLOR', '0'))
         if not force_color and (not f_out.isatty() or not f_err.isatty()):
             self.dummy_term = True
             return
-
+        if sys.platform == 'win32':
+            self.dummy_term = self.initialize_win_colors()
```
ilevkivskyi: I would rather make the bool value opposite for both methods and document them as "Return True if initialization was successful, False otherwise".

mypy/util.py
```diff
@@ -467,30 +474,63 @@ class FancyFormatter:
...
+        else:
+            self.dummy_term = self.initialize_unix_colors()
+        if not self.dummy_term:
+            self.colors = {'red': self.RED, 'green': self.GREEN,
+                           'blue': self.BLUE, 'yellow': self.YELLOW,
+                           'none': ''}
+
+    def initialize_win_colors(self) -> bool:
+        # Windows ANSI escape sequences are only supported on Threshold 2 and above.
+        if sys.platform == 'win32':
```
ilevkivskyi: FWIW I would prefer the assert, instead of silently doing nothing on wrong platform.

ethanhs: Unfortunately the assert does not correctly narrow the platform. I'll use both for runtime and compile time checks (and leave a note explaining why).

PR #7470, opened 2019-09-05
Point error to incompatible argument instead of call expression

JukkaL:
Previously we pointed to the beginning of the call expression if there
was an incompatible argument, which is not great, particularly if there 
are multiple arguments.

Also point to incompatible list/dict/set item instead of the beginning
of a list/dict/set expression.

Note that this may require some `# type: ignore` comments to be
updated.

I had to update the output of many test cases, mostly because line
ordering was changed.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -64,7 +64,16 @@ import mypy.errorcodes as codes
 
 # Type of callback user for checking individual function arguments. See
 # check_args() below for details.
-ArgChecker = Callable[[Type, Type, int, Type, int, int, CallableType, Context, MessageBuilder],
+ArgChecker = Callable[[Type,
+                       Type,
+                       int,
+                       Type,
+                       int,
+                       int,
+                       CallableType,
+                       Context,
+                       Context,
+                       MessageBuilder],
                       None]
 
 # Maximum nesting level for math union in overloads, setting this to large values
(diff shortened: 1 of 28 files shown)


mypy/checkexpr.py
```diff
@@ -64,7 +64,16 @@
...
+ArgChecker = Callable[[Type,
+                       Type,
+                       int,
+                       Type,
+                       int,
+                       int,
+                       CallableType,
+                       Context,
+                       Context,
+                       MessageBuilder],
```
ilevkivskyi: Are you trying black? :-)

mypy/checkexpr.py
```diff
@@ -1169,7 +1178,8 @@ def check_argument_count(self,
         if messages:
             assert context, "Internal error: messages given without context"
         elif context is None:
-            context = TempNode(AnyType(TypeOfAny.special_form))  # Avoid "is None" checks
+            # Avoid "is None" checks
+            context = TempNode(AnyType(TypeOfAny.special_form), context=context)
```
ilevkivskyi: `context=context` is redundant since it is anyway `None`.

ilevkivskyi:
It looks like you also need to update the `--pretty` test.

ilevkivskyi:
Also note that on Python 3.8 `# type: ignore` spans expression, i.e. a `# type: ignore` on a line between `ctx.line` and `ctx.end_line` suppresses the error for given `ctx`.

PR #7475, opened 2019-09-05
Work around crash caused by mypyc/mypyc#695

msullivan:
mypyc has some issues when shadowing variables in a list comprehension
so work around it for now. This only affects mypyc when compiled with
mypyc.


msullivan:
I think I'm going to put up a diff to actually fix the bug instead of merging this, though.

msullivan:
(The bug is in mypy itself actually)

ilevkivskyi:
OK, as you prefer.

ilevkivskyi:
Superseded by https://github.com/python/mypy/pull/7476

PR #7476, opened 2019-09-05
Fix interaction of list comprehension shadowing and binders

msullivan:
Binders are keyed by "literal_hash" which uses the name of the
variable as part of the key. Use the Var node instead so that
shadowing works.

Fixes mypyc/mypyc#695 but also was a bug separate from mypyc.

### DIFF
--- a/mypy/literals.py
+++ b/mypy/literals.py
@@ -113,7 +113,10 @@ class _Hasher(ExpressionVisitor[Optional[Key]]):
         return ('Star', literal_hash(e.expr))
 
     def visit_name_expr(self, e: NameExpr) -> Key:
-        return ('Var', e.name)
+        # N.B: We use the node itself as the key, and not the name,
+        # because using the name causes issues when there is shadowing
+        # (for example, in list comprehensions).
+        return ('Var', e.node)
 
     def visit_member_expr(self, e: MemberExpr) -> Key:
         return ('Member', literal_hash(e.expr), e.name)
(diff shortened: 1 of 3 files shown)


mypy/literals.py
```diff
@@ -113,7 +113,10 @@ def visit_star_expr(self, e: StarExpr) -> Key:
         return ('Star', literal_hash(e.expr))
 
     def visit_name_expr(self, e: NameExpr) -> Key:
-        return ('Var', e.name)
+        # N.B: We use the node itself as the key, and not the name,
+        # because using the name causes issues when there is shadowing
+        # (for example, in list comprehensions).
+        return ('Var', e.node)
```
ilevkivskyi: Shouldn't we instead push a new binber frame when accepting comprehension to match what Python does?

I mean binder works fine with:
```python
def foo(x: object) -> None:
    if isinstance(x, str):
        def nested() -> None:
            x = 42
            reveal_type(x)
```

msullivan: That would discard other bindings, right? Which we wouldn't want.

ilevkivskyi: Indeed, this may cause tons of new false positives (unlike for normal functions where this is questionable, comprehensions are functions executed immediately after definition so always safe).

OK, although this somehow looks dangerous on some unconscious level, I can't really object, so go ahead.

msullivan: Yeah, it is a little perilous seeming, but the `Var` node is fundamentally how we represent a variable binding, and having separate `Var` nodes is how we implement the new scope in a comprehension, so I think it is OK.

PR #7481, opened 2019-09-07
Support pickling of extension classes

### LINKED ISSUE #697: Is there a way to type-alias a generic type?
Can I write a generic type-aliase so that:

```MyBundle[t]``` is exactly equivalent to ```Tuple[List[t], t, List[t]]```

### PR DESCRIPTION
**Pull request by msullivan**
This operates by providing default implementations of `__getstate__`
and `__setstate__` for extension classes. Our implementations work by
storing a `__mypyc_attrs__` tuple in each class that we generate and
collecting all of the attributes in it into a dict.

This is is the last big obstacle to supporting black well.

Fixes mypyc/mypyc#697.

### DIFF
--- a/mypyc/genops.py
+++ b/mypyc/genops.py
@@ -533,7 +533,7 @@ def prepare_class_def(path: str, module_name: str, cdef: ClassDef,
 
     for name, node in info.names.items():
         if isinstance(node.node, Var):
-            assert node.node.type, "Class member missing type"
+            assert node.node.type, "Class member %s missing type" % name
             if not node.node.is_classvar and name != '__slots__':
                 ir.attributes[name] = mapper.type_to_rtype(node.node.type)
         elif isinstance(node.node, (FuncDef, Decorator)):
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: mypyc/test-data/run-classes.test [hunk 1]
```diff
@@ -1031,3 +1031,70 @@ try:
...
+assert not hasattr(D, '__mypyc_attrs__')
+assert E.__mypyc_attrs__ == ('y', '__dict__')
+
+b = B(10, '20', False)
+assert b.__getstate__() == {'z': False, 'x': 10, 'y': '20'}
+b2 = copy.copy(b)
+assert b != b2 and b.y == b2.y
+
+b3 = pickle.loads(pickle.dumps(b))
+assert b != b3 and b.y == b3.y
```
ilevkivskyi: I think `is not` instead of `!=` will make intention more clear here and above.

mypyc/genops.py
```diff
@@ -595,6 +595,8 @@ def prepare_class_def(path: str, module_name: str, cdef: ClassDef,
                 ir.inherits_python = True
             continue
         base_ir = mapper.type_to_ir[cls]
+        if not base_ir.is_ext_class:
+            ir.inherits_python = True
```
ilevkivskyi: Shouldn't this be "transitive"? I.e what if all `base_ir` are all extension classes, but some inherit from Python? I would add a test for this (like add a `class F(E): pass` and check it has `__dict__`).

msullivan: This is already searching the whole mro, but there is a check that isn't part of that search that I'll move in.

PR #7483, opened 2019-09-07
Allow passing path/to/file.py:line instead of fully.qualified.name to dmypy suggest

ilevkivskyi:
This should help with integrations with editors and IDEs, where getting line number of a function might be much easier than the full name.

The implementation in mostly straightforward. I also did a little refactoring to reuse the existing logic efficiently. The only important difference that I introduced after few experiments is that I force reloading of the file before trying to search the function, since even the smallest edit can offset the line number and lead to some weird results.

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -247,6 +247,8 @@ class MypyFile(SymbolNode):
     # If the value is empty, ignore all errors; otherwise, the list contains all
     # error codes to ignore.
     ignored_lines = None  # type: Dict[int, List[str]]
+    # This map allow find quickly a top level function/method by its line number.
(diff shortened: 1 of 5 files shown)


ilevkivskyi:
Couple observations:
* This is still a bit slow, takes about 7.5s (interpreted) for a method in `mypy/types.ty`, while providing full name takes 6.5s.
* Suggestion logic doesn't work for `yield`.

ilevkivskyi:
> Is the slowdown mostly caused by the extra reload?

It looks like yes, I tried to play with this a bit more, and it looks like penalty is typically a bit lower, around 500ms (interpreted) for mypy self-suggest with some removed signatures. For compiled this would mean something like 1100ms vs 1200ms, so probably not worth it to try to optimize now (it affects only `suggest`, not other commands). We can come back to this later if it turns out o be important.

PR #7488, opened 2019-09-09
Run mypyc on Windows on tags

ethanhs:
Note this isn't ready for review yet, I want to test running mypyc actually works.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,7 @@ install:
 build: off
 
 test_script:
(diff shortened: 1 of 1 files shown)


ethanhs:
Ugh, the run is just erroring with "Command executed with exception: ". Guess I will need to figure out how to debug this...

@msullivan if you have any hints/suggestions they are more than welcome :P

appveyor.yml
```diff
@@ -20,6 +20,7 @@ install:
 build: off
 
 test_script:
+    - ps: "If ($env:APPVEYOR_REPO_TAG -Match \"false\") { Set-Alias -Name python3 -Value \"$env:PYTHON\\python.exe\"; python3 -m pip install -r mypy-requirements.txt; $env:CC=\"clang\"; $env:MYPYC_OPT_LEVEL=\"0\"; python3 setup.py --use-mypyc build_ext --inplace }"
```
ethanhs: Note to self: change this to `-Match "true"` when done testing.

msullivan:
What if you change the optimization level to "d" from "0"? That seems to be what it is complaining about the most.

msullivan:
we don't use appveyor anymore so I'm gonna close this one

Issue #7095, opened 2019-06-28
Overloaded function signatures don't fully qualify types


dirn:
This is a request for a future that will provide more useful output when calling overloaded functions.

When I check the following code with mypy

```
from a import MyType as MyTypeA
from b import MyType as MyTypeB

def f(a: MyTypeA) -> MyTypeB:
    return a
```

I get

```
error: Incompatible return value type (got "a.MyType", expected "b.MyType")
```

The message includes the fully qualified name for the types that don't match. When I check this code with mypy, though

```
from typing import overload, Union

from a import MyType as MyTypeA
from b import MyType as MyTypeB

@overload
def f(a: MyTypeA) -> MyTypeB:
    pass

@overload
def f(a: MyTypeB) -> MyTypeB:
    pass

def f(a: Union[MyTypeA, MyTypeB]) -> MyTypeB:
    return a

def g():
    f(MyTypeA())
```

I get

```
error: Incompatible return value type (got "Union[MyType, MyType]", expected "MyType")
```

This message only includes the names of the types, not the fully qualified name. (This also applies to the `No overload variant` message that is sometimes output.) The information is really useful for understanding why the type mismatch occurs.

The above output was generated using mypy 0.711 and CPython 3.7.2, but I can also reproduce the error using 46aa00d8a8d5c870c6ef24b82261b6b257e2ea03 against code written for CPython 2.7.x.

ilevkivskyi:
(Setting priority to high since the error is really obscure, and is probably not hard to fix.)

nkashy1:
@ilevkivskyi : I made a PR which adds a test for this issue: https://github.com/python/mypy/pull/7490

That's as far as I got, because the test passes.

Not sure if you want to merge in that test, but it's there if you want it.

msullivan:
This was fixed by #7226. Thanks for triaging this, @nkashy1!

PR #7490, opened 2019-09-10
Added a test issue #7095

nkashy1:
https://github.com/python/mypy/issues/7095

I wrote a test to verify the behavior observed in https://github.com/python/mypy/issues/7095

Was pleasantly surprised to find the test passing.

Using mypy version `0.730+dev` and Python version `3.7.1`.


msullivan:
Actually, on closer inspection, I don't think this had anything to do with overloads, just with unions, and there is already a test it works with unions that I included in the PR that fixed this (#7226). Thanks, though!

PR #7496, opened 2019-09-10
[mypyc] Fix using values from other modules that were reexported

### LINKED ISSUE #393: Support Callable[..., t] (with literal ellipsis, unspecified argument types and specified return type)
It would be handy to be able to specify a callable whose argument types are not checked but whose return type is, e.g., by specifying the argument types as `None` instead of a list. 

For example, consider `map`:

    def map(function, *iterables):

How would you annotate that? Trying to actually specify the argument types to `function` to match the element types of the iterables in `iterables` would require something horrible like C++11's parameter packs and variable-length type tuples. But just specifying `Function` means there's no way to know that the result iterates values of the same type the function returns, which seems like a pity.

If you could just use `None` or `Any` instead of a list, you could do this:

    def map(function: Function[None, T], *iterables: Iterable[Iterable]) -> Iterator[T]:

msullivan:
There are a couple parts to this:
 * Compile module attribute accesses by compiling the LHS on its own
   instead of by loading its fullname (which will be what mypy thinks
   its source is)
 * Only use the static modules if they have actually been imported

Closes mypyc/mypyc#393.

I ran into this again while finishing up my separate compilation PR and so got annoyed and fixed it.

### DIFF
--- a/mypyc/genops.py
+++ b/mypyc/genops.py
@@ -1050,8 +1050,9 @@ class IRBuilder(ExpressionVisitor[Value], StatementVisitor[None]):
 
         self.errors = errors
         self.mapper = mapper
-        # We use an OrderedDict for this so we can maintain insertion order
-        # and do quick lookups
+        # Notionally a list of all of the modules imported by the
+        # module being compiled, but stored as an OrderedDict so we
+        # can also do quick lookups.
         self.imports = OrderedDict()  # type: OrderedDict[str, None]
 
     def visit_mypy_file(self, mypyfile: MypyFile) -> None:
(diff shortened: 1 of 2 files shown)


mypyc/genops.py
```diff
@@ -1050,7 +1050,9 @@ def __init__(self,
 
         self.errors = errors
         self.mapper = mapper
-        self.imports = []  # type: List[str]
+        # We use an OrderedDict for this so we can maintain insertion order
+        # and do quick lookups
+        self.imports = OrderedDict()  # type: OrderedDict[str, None]
```
JukkaL: Document this attribute. Maybe this would benefit from renaming -- it took me a while to figure out what it's used for.

mypyc/test-data/run.test
```diff
@@ -4645,3 +4645,18 @@ from native import foo
 
 assert foo(None) == None
 assert foo([1, 2, 3]) == ((1, 2, 3), [1, 2, 3])
+
+[case testReexport]
```
JukkaL: Maybe also test re-exporting a module or a class. What about re-exporting a global variable?

PR #7530, opened 2019-09-18
Give self argument the correct type when using check_untyped_defs

msullivan:
This is accomplished by updating the helper that constructs a callable type
from a func def to fill in the self argument.

This will make check_untyped_defs a lot more useful.

### DIFF
--- a/mypy/typeops.py
+++ b/mypy/typeops.py
@@ -10,7 +10,7 @@ from typing import cast, Optional, List, Sequence, Set
 from mypy.types import (
     TupleType, Instance, FunctionLike, Type, CallableType, TypeVarDef, Overloaded,
     TypeVarType, TypeType, UninhabitedType, FormalArgument, UnionType, NoneType,
-    AnyType, TypeOfAny, ProperType, get_proper_type, get_proper_types, copy_type
+    AnyType, TypeOfAny, TypeType, ProperType, get_proper_type, get_proper_types, copy_type
 )
 from mypy.nodes import (
     FuncBase, FuncItem, OverloadedFuncDef, TypeInfo, TypeVar, ARG_STAR, ARG_STAR2,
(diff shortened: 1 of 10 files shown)


test-data/unit/check-classes.test
```diff
@@ -6178,3 +6178,15 @@ class B(A):
...
+# flags: --check-untyped-defs
+class Foo:
+    def foo(self):
+        self.bar()
+        self.baz()  # E: "Foo" has no attribute "baz"
+
+    @classmethod
+    def bar(cls):
+        cls.baz()  # E: "Type[Foo]" has no attribute "baz"
+[builtins fixtures/classmethod.pyi]
```
ilevkivskyi: I would add also tests for two other reported problems:
* An attribute initialized with `0` in untyped `__init__` has revealed type `int` in another untyped method
* Descriptors also work as expected in untyped methods.

mypy/typeops.py
```diff
@@ -369,3 +370,50 @@ def erase_to_union_or_bound(typ: TypeVarType) -> ProperType:
...
+                                 fallback,
+                                 line=func.line, is_ellipsis_args=True)
+            # Return an Overloaded, because some callers may expect that
+            # an OverloadedFuncDef has an Overloaded type.
+            return Overloaded([dummy])
+
+
+def callable_type(fdef: FuncItem, fallback: Instance,
+                  ret_type: Optional[Type] = None) -> CallableType:
+    # FIXME? somewhat unfortunate duplication with prepare_method_signature in semanal
```
ilevkivskyi: I would make this a TODO (without question mark).

mypy/typeops.py
```diff
@@ -369,3 +370,50 @@ def erase_to_union_or_bound(typ: TypeVarType) -> ProperType:
...
+    if fdef.info and not fdef.is_static:
+        self_type = fill_typevars(fdef.info)  # type: Type
```
ilevkivskyi: Another test idea inspired by `fill_typevars()`:
```python
class C(Generic[T]):
    x: T
    def meth(self):
        self.x + 1  # Should be an error with --check-untyped-defs
```

test-data/unit/check-classes.test
JukkaL: Also, if an attribute is initialized to `None` in `__init__` the inferred type should probably be either `Optional[Any]` or `Any` (the first one may generate many false positives). If an attribute is initialized to `[]`, the inferred type should probably be `List[Any]`, and I think that we shouldn't require a type annotation.

The-Compiler:
As a data point: In [my project](https://www.qutebrowser.org/) I just introduced `check_untyped_defs` a week ago - after the upgrade to 0.740 I now see 469 new errors :laughing: 

Thanks for this, though! I have some work ahead of me now, but it indeed seems to make mypy a lot more useful for codebases which aren't fully annotated yet.

Issue #3903, opened 2017-08-31
untyped defs on derived classes


ostcar:
Currently all arguments and return values on untyped defs are set to ```Any```.

It would be nice, if arguments and return values on derived classes would take the type annotations from the parent class. For example:

```
class A:
    def foo(self, a: str) -> None: pass


class B(A):
    def foo(self, a):
        reveal_type(a)
```
In this example the revealed type is ```Any``` but I would expect it to be ```str```.

Of cause, if the entire project has type annotations, than this would not be an issue, but when you add type annotations on a big project and start with the base classes, it would be nice, if you could benefit from that.

ethanhs:
I believe we probably don't want to support this unless `--check-untyped-defs` is enabled. Otherwise it would be inconsistent with how things work outside the class. If we restrict it to with that flag I believe this could be a nice feature to add, however this would complicate things and slow mypy.

gvanrossum:
I think this should be a separate, new flag. Then it sounds worth
considering.

ilevkivskyi:
+1 from me for a _separate_ flag.

ilevkivskyi:
Also for reference, a very similar discussion happened in the typing tracker, see https://github.com/python/typing/issues/269.

JukkaL:
This can't be enabled by default, since a method override can have a different signature from the base class, and blindly inheriting the signature can generate false positives. A separate flag is worth considering, since similar things have been proposed several times.

ethanhs:
Thinking more about it, I suppose an override should require a new type signature, as the method body is not inherited with an override, thus it is likely a signature not to be. So Im not sure I think this is such a good feature after all. If you are overriding, it is a different method body, thus inheriting the type signature seems wrong (at least to me).

However, if we do support this, I think a flag is too global, as these changes are likely to be only wanted on a single class or method even, but that is my personal experience on this.

gvanrossum:
You could inherit the type signature (which I do think is the common case
for method overrides) but still check the new body against that signature.

JukkaL:
Yes, I think that this only makes sense if we check the body against the inherited signature. This is another reason why this needs to be opt-in -- unannotated methods are not type checked according to PEP 484.

OrDuan:
I want to point out that PyCharm added this to their code analyzer.
So far this is really helpful for me.
Usually I have a base class and child class that should have the same signatures, if it is different, I override it with different types. 

[...]

PR #7548, opened 2019-09-21
Fix 3903: untyped defs on derived classes

gantsevdenis:
fix https://github.com/python/mypy/issues/3903
Looked simple enough, but not sure about the part:
 ```python
if not defn.info:
            return None
```
If removed, produces AssertionErrors

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -277,9 +277,8 @@ definitions or calls.
     as the return type.
 
 ``--inherit-signatures``
-    This flag must be used along with `--check-untyped-defs` flag. It allows
-    to use base class function signature in derived classes, without explicitly
-    annotating the derived class
+    This allows to use base class function signature in derived classes, without explicitly
+    annotating the derived class.
 
 ``--disallow-untyped-decorators``
     This flag reports an error whenever a function with type annotations
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -276,6 +276,11 @@ definitions or calls.
     It will assume all arguments have type ``Any`` and always infer ``Any``
     as the return type.
 
+``--inherit-signatures``
+    This flag must be used along with `--check-untyped-defs` flag. It allows
+    to use base class function signature in derived classes, without explicitly
+    annotating the derived class
+
 ``--disallow-untyped-decorators``
     This flag reports an error whenever a function with type annotations
     is decorated with a decorator without annotations.
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -87,8 +87,8 @@ from mypy.types import (
     FunctionLike, UnboundType, TypeVarDef, TupleType, UnionType, StarType,
     CallableType, Overloaded, Instance, Type, AnyType, LiteralType, LiteralValue,
     TypeTranslator, TypeOfAny, TypeType, NoneType, PlaceholderType, TPDICT_NAMES, ProperType,
-    get_proper_type, get_proper_types
-)
+    get_proper_type, get_proper_types,
+    TypeList)
 from mypy.typeops import function_type
 from mypy.type_visitor import TypeQuery
 from mypy.nodes import implicit_module_attrs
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

**Comment by ilevkivskyi** [comment]
I think most of the questions are answered in https://github.com/python/mypy/issues/3903#issuecomment-408833797 (there are also couple additional points).

**Comment by gantsevdenis** [comment]
Thank you for guiding!
I have followings question points:
- I added support for generics, but not sure if something exists already which could be used, instead of my own `substitute..` function?
- I had to do a `cast`, because I couldn't solve typing problem..
- I am not sure to understand the difference between `incremental` mode, and `daemon` mode, but I ve tried using feature with daemon, and could not encounter any dependency tracking issues (I used a very simple setup though, only 2 modules, one import each other, one of them defines a`class Base`, and the other imports it: the typing issues in the second module were caught )

File: mypy/semanal.py [hunk 1]
```diff
@@ -160,6 +160,45 @@
 Tag = int
 
 
+def substitute_unbounds(t: UnboundType, names_to_types: Mapping[str, Instance]) -> ProperType:
```
ilevkivskyi: Sorry for misleading suggestion, doing the copy during semantic analyzis is still bad. I was thinking about superclass being in a different module that is type-checked before the module with subclass, but it obviously doesn't work always.

Also this function is just one big hack. I was thinking about something along the lines of `mypy.checker.TypeChecker.bind_and_map_method()`.

For the latter to work well I see a possible strategy:

0. Add two flags (or maybe better one three state flag) to functions indicating whether this function can be "improved", i.e. it has some untyped arguments, while a supertype has those typed, and whether the "copying" has already been performed.
1. The first flag should be set during semantic analysis, this should be thought carefully to work well in import cycles.
2. During type checking, whenever we encounter a reference to a method `Sub.meth` for which we know it can be "improved", but wasn't yet, we trigger a deferral.
3. Finally, we perform the copying while analyzing the method itself (in a place where you did initially, but with a more sophisticated logic like currently)
4. Be sure that `aststrip.py` resets the flags, and `deps.py` generates dependencies so that subclass method callsites are all rechecked when the signature in supertype changes (this may be already the case, at least partially).

mypy/semanal.py
```diff
@@ -160,6 +160,45 @@
...
+            def foo(self) -> U:pass
+            def bar(self,*args)->Callable[[U],T]:pass
+        class B(A[int,str]):
+            def foo(self):pass
+            def bar(self,*args):pass
+
+    Then class B will be transformed to:
+         class B(A[int,str]):
+            def foo(self) -> str:pass
+            def bar(self,*args)->Callable[[str],int]:pass
```
ilevkivskyi: Please use PEP 8 everywhere including in docstring and in tests! (You can however skip empty lines in tests whenever it makes sense).

mypy/semanal.py
```diff
@@ -477,6 +516,74 @@ def adjust_public_exports(self) -> None:
...
+                if isinstance(parent, Decorator):
+                    parent = parent.func
+                break
+        if not parent:
+            return None
+        assert base_1 is not None
+        assert isinstance(parent, FuncDef), "got %s" % type(parent)
+        pnb_args, cnb_args = len(parent.arg_kinds), len(func.arg_kinds)
+        if cnb_args < pnb_args:
+            self.fail("Signature of \"%s\" incompatible with supertype \"%s\"" %
```
ilevkivskyi: All the message generating logic should live `messages.py`. Also I think we should emit a more specific error message, explaining that we can't copy the signature, and mentioning what went wrong (not enough arguments, etc).

Also in Python you can use `'single quotes to avoid "escaping" double quotes'` (also below).

test-data/unit/check-classes.test
```diff
@@ -6223,3 +6223,97 @@ class Foo:
...
+    def return_KT(self) -> KT: pass
+    def return_VT(self) -> VT: pass
+    def return_VT_wrong(self) -> VT: pass
+    def return_Tuple(self)->Tuple[VT,KT]:pass
+class StrDict(MyDict[str, int]):
+    def return_KT(self,*args,**kwargs): return "5"
+    def return_VT(self, *args,**kwargs): return 5
+    def return_VT_wrong(self,*args,**kwargs): return ""  # E: Incompatible return value type (got "str", expected "int")
+    def return_Tuple(self,*args,**kwargs): return ("", 5)  # E: Incompatible return value type (got "Tuple[str, int]", expected "Tuple[int, str]")
+[builtins fixtures/dict.pyi]
```
ilevkivskyi: Several important scenarios are still missing, for example:
* Methods with self-types
* Generic inheritance where subclass has new type variables with non-trivial mapping to old ones
* Chained inheritance where `C` inherits `B` that inherits `A`
* Multiple inheritance
* Incremental mode tests, see `check-incremental.test`
* Daemon mode tests, see `fine-grained.test`
* Import cycles
* Tests that inherited signature is used on external and internal access (including situations with deferrals)

ilevkivskyi:
Also please note there are some merge conflicts.

docs/source/command_line.rst
```diff
@@ -276,6 +276,10 @@ definitions or calls.
     It will assume all arguments have type ``Any`` and always infer ``Any``
     as the return type.
 
+``--inherit-signatures``
+    This allows to use base class function signature in derived classes, without explicitly
```
rmorshea: Slight typo - "This allows the use of..." instead of "This allows to use..."

rmorshea:
This is awesome work and would go a long way in resolving all my use cases for https://github.com/python/typing/issues/270. Thanks for this!

robertwb:
I would love to see this as well, would eliminate a lot of DRY violations that are currently required for typing.

test-data/unit/check-classes.test
gantsevdenis: I am sorry this looks daunting, especially the daemon / incremental part, I didn't expect that...
Could you please give an example of what is a signature with "self-type"? I don't quite understand that term

mypy/semanal.py
ilevkivskyi: @gantsevdenis I can't find your comment that I have got by e-mail, but here is the response. Self-types is something like this:
```python
T = TypeVar('T')
class Base:
    def copy(self: T) -> T: ...

class Sub(Base): ...
x = Sub().copy()  # inferred type of 'x' is 'Sub'
```

msullivan:
I'm going to close this because it has been open with changes requested for quite some time. Anyone should feel free to pick it back up, though!

PR #7550, opened 2019-09-21
Switch ret. type of asyncio.gather() to list

bsolomon1124:
This is a companion commit dependent on the change
introduced in typeshed PR 3248:

https://github.com/python/typeshed/pull/3248


JelleZijlstra:
I'm traveling for the next week so I won't have time to do the merge dance for a little bit, but I'll take care of it after I come back unless someone else gets to it first.

PR #7569, opened 2019-09-27
Point incompatible assignment errors to rvalue instead of lvalue

JukkaL:
This seems more logical. If there is a tuple lvalue, we point to the
incompatible tuple item.

Note that this may break some `# type: ignore` statements since
the reported line number may change (but it usually doesn't).


JukkaL:
> Do we need a similar change to augmented assignments

They work like this already.

PR #7573, opened 2019-09-27
Fix some notes not being ignored when using an error code

### LINKED ISSUE #7562: Ignoring a specific error code still outputs "notes" and returns exit status 1
Here's a minimal repro:

```python
from datetime import datetime
from typing import Optional

a: Optional[datetime] = None
b = datetime.now()

def some_function() -> None:
    if a <= b:
        pass
```

Running mypy on that file with `--show-error-codes` and `--pretty` outputs:

```
mypy_test.py:8: error: Unsupported operand types for >= ("datetime" and "None")  [operator]
        if a <= b:
           ^
mypy_test.py:8: note: Left operand is of type "Optional[datetime]"
```

If I add `# type: ignore` on the `if a <= b:` line, there is no error output, as expected.

However, if I use `# type: ignore[operator]` instead, the output is:

```
mypy_test.py:8: note: Left operand is of type "Optional[datetime]"
```

This also gives an exit status of `1`, making it seem like it returned an error.

As one step more, if I also add the `--show-error-context` flag, output is now:

```
mypy_test.py: note: In function "some_function":
mypy_test.py:8: note: Left operand is of type "Optional[datetime]"
```

JukkaL:
Either change the error code to match the related error, so that
both will always be ignored at the same time, or switch a note to
an error (in case the relevant error code is hard to predict).

Fixes #7562.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1472,8 +1472,8 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                                  callable_name=callable_name,
                                  object_type=object_type)
         if union_interrupted:
-            self.chk.msg.note("Not all union combinations were tried"
-                              " because there are too many unions", context)
+            self.chk.fail("Not all union combinations were tried"
+                          " because there are too many unions", context)
         return result
 
     def plausible_overload_call_targets(self,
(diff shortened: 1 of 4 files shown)


msullivan:
`note: Right operand is of type ...` still seems to slip through

msullivan:
`note: Right operand is of type ...` still seems to slip through

Issue #7053, opened 2019-06-24
Make column/line numbers more precise for common errors


JukkaL:
In some cases mypy error messages don't point to the expected location, especially when using `--show-column-numbers`. Example:

```py
def f(x: int) -> None: pass

# Mypy message points to 'f' at the beginning of the line
f(
    'foo')     
  # ^ I'd expect the error point to 'foo' here
```

This is a significant usability issue when running mypy from an editor, especially if error locations are highlighted in the code. Having the correct line number would help usability in general, however.

It would be good to go through selected common error messages, such as invalid argument, undefined name, missing attribute, and incompatible assignment, and make sure that the line/column numbers point to the correct part of the expression/statement.

Note that this may break existing `# type: ignore` comments. I don't think that there's a way around this, other than perhaps providing a tool that can automatically move `# type: ignore` comments to the new locations.

This is basically a subset of existing issues #4276 and #4829. The original issues are too wide-ranging to be actionable in the short term, but fixing a few common messages would solve perhaps 80% of the problem without much effort.

JukkaL:
Some of this has been implemented in #7078 and #7081. I'm also working on another PR, but it turned out to be a bit harder than expected.

JukkaL:
#7569 and #7470 improved error locations for return statement and incompatible arguments.

JukkaL:
This seems good enough now. We can create additional issues about specific contexts where line/column numbers are still imprecise.

PR #7578, opened 2019-09-27
Make reported line and column numbers more precise

JukkaL:
Most notably, report more precise error locations for incompatible
return value types and incompatible default argument values.

Also add some tests.

Work towards #7053.


test-data/unit/check-columns.test
```diff
@@ -326,6 +336,12 @@ if int():
 main:2:11: error: Syntax error in type annotation
 main:2:11: note: Suggestion: Is there a spurious trailing comma?
 
+[case testColumnSyntaxErrorInTypeAnnotation2]
+if int():
+        x = 0  # type: x y
+[out]
+main:2:9: error: syntax error in type comment 'x y'
```
ilevkivskyi: It is not clear what is 9th column here? Maybe keep the intent as 4, otherwise there are two 8-column offsets, the indent itself and also the length of `# type: `.

test-data/unit/check-columns.test
```diff
@@ -133,9 +133,11 @@ def foobar(): pass
 
 [builtins fixtures/module.pyi]
 
-[case testColumnUnexpectedKeywordArg]
+[case testColumnUnexpectedOrMissingKeywordArg]
 def f(): pass
 (f(x=1)) # E:2: Unexpected keyword argument "x" for "f"
```
ilevkivskyi: In future we might want to point to `x` here.

JukkaL: Yeah, added a TODO. I don't think that we can easily find the location of `x` currently.

test-data/unit/check-columns.test
JukkaL: A good point, updated.

PR #7590, opened 2019-09-30
Fix #7570: Include mypyc/lib-rt in source distributions

### LINKED ISSUE #7570: Add complete tarballs to releases on GitHub or extend MANIFEST.in to include mypyc/lib-rt
Hello,

I'm updating the Debian Package for mypy and very happy to have mypyc included.

When using the sdist from PyPI we are unable to build with `MYPY_USE_MYPYC=1` as the `mypyc/lib-rt` directory is not present.

Switching to the autogenerated source code tarballs from the GitHub releases doesn't work because they lack the Git submodules, specifically `mypy/typeshed`

https://github.com/python/mypy/releases/tag/v0.730

We'd love to include the much faster and compiled version of mypy in Debian. 

Options:
1. Tarballs that contain `mypy/typeshed` are added to the GitHub releases (manually at first, and then automatically)
2. `MANIFEST.in` is augmented to include the missing paths, such at `mypyc/lib-rt` and `misc/proper_plugin.py` (referenced from `mypy_self_check.ini` which we run)

`mypyc/lib-rt/` is about 29K gzip'd, so I don't think that is too much to add to the sdist.

I'm happy to send a PR for option 2, just let me know.

Thanks, as always, for a fantastic project!

jreese:
Adds mypyc/lib-rt to MANIFEST.in so that it gets included in source distributions for building with mypyc.

Verified by:
* applying this change to the v0.730 tag
* running `python setup.py sdist` to generate `mypy-0.730.tar.gz`
* creating and activating a fresh venv via `python3.7 -m venv`
* extracting the tarball
* running `pip install -U pip mypy_extensions typing_extensions typed_ast`
* running `MYPY_USE_MYPYC=1 python setup.py bdist_wheel`

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,6 +4,7 @@ recursive-include extensions *
 recursive-include docs *
 recursive-include mypy/typeshed *.py *.pyi
 recursive-include mypy/xml *.xsd *.xslt *.css
+recursive-include mypyc/lib-rt *.c *.h
 include mypy_bootstrap.ini
 include mypy_self_check.ini
 include LICENSE


msullivan:
Thanks!

mr-c:
Yes, thank you @jreese !

jreese:
@msullivan thanks for merging. Is there any chance of a point release being made available on PyPI, or do we just need to continue building from a git checkout until next month?

PR #7619, opened 2019-10-03
[attrs] Support attr.s(eq=..., order=...)

### LINKED ISSUE #7615: attrs plugin doesn't support new `eq` and `order` arguments
attrs 19.2.0 has added `eq` and `order` arguments to `attr.ib`, and deprecated `cmp`. I think mypy needs to be updated to take this into account.

```python
@attr.s(auto_attribs=True, frozen=True)
class Context:
    _original: Dict[str, Any] = attr.ib(order=False)
```

```
_context.py:37: error: No overload variant matches argument type "bool"
_context.py:37: note: Possible overload variants:
_context.py:37: note:     def (default: None = ..., validator: None = ..., repr: bool = ..., cmp: bool = ..., hash: Optional[bool] = ..., init: bool = ..., convert: None = ..., metadata: Optional[Mapping[Any, Any]] = ..., type: None = ..., converter: None = ..., factory: None = ..., kw_only: bool = ...) -> Any
_context.py:37: note:     def [_T] (default: None = ..., validator: Union[Callable[[Any, Attribute[_T], _T], Any], Sequence[Callable[[Any, Attribute[_T], _T], Any]], None] = ..., repr: bool = ..., cmp: bool = ..., hash: Optional[bool] = ..., init: bool = ..., convert: Optional[Callable[[Any], _T]] = ..., metadata: Optional[Mapping[Any, Any]] = ..., type: Optional[Type[_T]] = ..., converter: Optional[Callable[[Any], _T]] = ..., factory: Optional[Callable[[], _T]] = ..., kw_only: bool = ...) -> _T
_context.py:37: note:     <2 more similar overloads not shown, out of 4 total overloads>
```

mypy 0.711 and 0.730

I expect this to pass type checking, as attrs 19.2.0 adds `eq` and `order` parameters to attr.ib: http://www.attrs.org/en/19.2.0/changelog.html#deprecations

Without this, using `cmp=` generates a DeprecationWarning. 

I haven't tried installing mypy from Git master.

```
[mypy]
python_version = 3.7
strict_optional = true
warn_redundant_casts = true
namespace_packages = true
```

euresti:
attrs 19.2.0 added support for eq and order and deprecated to attr.s.
It also deprecated cmp.

Fixes #7615


ilevkivskyi:
Can this cause false positives for people who have an older version of `attrs`? If yes, this is probably something we should mention in the release notes probably.

euresti:
Yes this will cause false positives for older versions of attrs. I don't know what the process for that normally is.

belm0:
Following up on PR comment https://github.com/python/mypy/commit/47093eb2fb1b0112b5bc76dc4ea98088a2f7e843#r35428812

Failing on deprecated API use doesn't seem in line with mypy's scope, and defeats Python's tried and true depreciation mechanisms.

The timing of say, when a project is forced to upgrade attrs due to a transitive dependency and when it's able to migrate it's own code off a deprecated API is often different due to policy and other constraints.  The runtime deprecation mechanism and supporting tools allows projects to deal with this sanely.

euresti:
Yep. It's been fixed.

belm0:
@euresti 

> Yep. It's been fixed. #7675

Well, but unfortunately the fix didn't make it into the new 0.740 release  😿 

```
foo.py:23: error: cmp is deprecated, use eq and order
```

please have it patched...

PR #7620, opened 2019-10-04
Add example of non-mutable mapping to cheat sheet

### LINKED ISSUE #7617: Mapping and MutableMapping use the same example argument in docs
The [cheat sheet from the docs](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html) shows the following:

```
# Mapping describes a dict-like object (with "__getitem__") that we won't
# mutate, and MutableMapping one (with "__setitem__") that we might
def f(my_dict: Mapping[int, str]) -> List[int]:
    return list(my_dict.keys())

f({3: 'yes', 4: 'no'})

def f(my_mapping: MutableMapping[int, str]) -> Set[str]:
    my_mapping[5] = 'maybe'
    return set(my_mapping.values())

f({3: 'yes', 4: 'no'})
```

In both cases, the argument passed to `f` is a `dict`, though in one case the argument is annotated as `Mapping` and in the other `MutableMapping`.

From this example, it might not be clear to a beginner when to use `Mapping` over `MutableMapping`.

MarcoGorelli:
closes #7617

### DIFF
--- a/docs/source/cheat_sheet_py3.rst
+++ b/docs/source/cheat_sheet_py3.rst
@@ -201,6 +201,7 @@ that are common in idiomatic Python are standardized.
 
 .. code-block:: python
 
+   from types import MappingProxyType
    from typing import Mapping, MutableMapping, Sequence, Iterable, List, Set
 
    # Use Iterable for generic iterables (anything usable in "for"),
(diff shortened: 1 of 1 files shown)


msullivan:
Thinking about this a bit, this example is correct but I think fails to capture the main reason you usually want to use `Mapping` over `MutableMapping`: to ensure that even *if* the mapping happens to be mutable (which it usually is, since it is usually just a dict), that the function *doesn't* mutate it.

This is useful in two ways:
 1. Just in a general correctness/reasoning sense. Making it immutable means the caller can't mutate it, which makes reasoning easier and might prevent someone introducing a later change that mutates it.
 2. From a type system perspective, `Mapping` is covariant in its second argument, which means that you can, for example, pass a `Dict[int, str]` to a function expecting a `Mapping[int, Optional[str]]`, for example. (Which you can't do if the function took a `Dict` or a `MutableMapping`)

MarcoGorelli:
Ah, I understand - so the difference isn't in the object you pass, but in what you do with that object. I've tried updating the PR so this is more explicit, but perhaps it's clear enough as is, in which case, feel free to close this :)

PR #7624, opened 2019-10-04
[Do not merge] Adding crossrefs to standard library

hoefling:
This is the second part of adding referencing to `mypy` docs (first part in #7623, this one is based on the first part). Replaced verbatim texts with crossrefs to standard library where applicable. Example: a passage
```rst
The ``typing`` module defines various protocol classes that correspond
to common Python protocols, such as ``Iterable[T]``.  If a class
defines a suitable ``__iter__`` method, mypy understands that it
implements the iterable protocol and is compatible with ``Iterable[T]``.
For example, ``IntList`` below is iterable, over ``int`` values:
```
image
becomes
```
The :py:mod:`typing` module defines various protocol classes that
correspond to common Python protocols, such as
:py:class:`Iterable[T] <typing.Iterable>`.
If a class defines a suitable :py:meth:`__iter__ <object.__iter__>`
method, mypy understands that it implements the iterable protocol
and is compatible with :py:class:`Iterable[T] <typing.Iterable>`.
For example, ``IntList`` below is iterable, over :py:class:`int`
values:
```
image
where each "bold verbatim" block references to the relevant spot in the standard library. IMO this has  several advantages:

1. When reading `mypy` docs, one doesn't need to keep a separate tab with [`typing` docs](https://docs.python.org/3/library/typing.html) opened, each term is immediately available for lookup.
2. Sometimes the highlighting is already enough - I can e.g. hover the link over `SupportsRound` and immediately get the hint - aha, this is part of the stdlib, `typing` module, can read it up later. Otherwise, especially an unexperienced reader will first have to google `SupportsRound` in separate tab.

### DIFF
--- a/docs/source/additional_features.rst
+++ b/docs/source/additional_features.rst
@@ -26,8 +26,8 @@ They can be defined using the :py:func:`@dataclasses.dataclass
     test = Application("Testing...")  # OK
     bad = Application("Testing...", "with plugin")  # Error: List[str] expected
 
-Mypy will detect special methods (such as ``__lt__``) depending on the flags used to
-define dataclasses. For example:
+Mypy will detect special methods (such as :py:meth:`__lt__ <object.__lt__>`)
+depending on the flags used to define dataclasses. For example:
 
 .. code-block:: python
 
@@ -72,10 +72,11 @@ and :pep:`557`.
 Caveats/Known Issues
 ====================
 
-Some functions in the ``dataclasses`` module, such as ``replace()`` and ``asdict()``,
+Some functions in the :py:mod:`dataclasses` module, such as
+:py:func:`~dataclasses.replace` and :py:func:`~dataclasses.asdict`,
 have imprecise (too permissive) types. This will be fixed in future releases.
 
-Mypy does not yet recognize aliases of ``dataclasses.dataclass``, and will
+Mypy does not yet recognize aliases of :py:func:`dataclasses.dataclass`, and will
 probably never recognize dynamically computed decorators. The following examples
 do **not** work:
 
@@ -140,9 +141,9 @@ If you're using ``auto_attribs=True`` you must use variable annotations.
         three: int = attr.ib(8)
 
 Typeshed has a couple of "white lie" annotations to make type checking
-easier. ``attr.ib`` and ``attr.Factory`` actually return objects, but the
-annotation says these return the types that they expect to be assigned to.
-That enables this to work:
+easier. :py:func:`attr.ib` and :py:class:`attr.Factory` actually
+return objects, but the annotation says these return the types that
+they expect to be assigned to. That enables this to work:
 
 .. code-block:: python
 
@@ -175,7 +176,7 @@ Caveats/Known Issues
 
 * Currently, ``converter`` only supports named functions.  If mypy finds something else it
   will complain about not understanding the argument and the type annotation in
-  ``__init__`` will be replaced by ``Any``.
+  :py:meth:`__init__ <object.__init__>` will be replaced by :py:data:`~typing.Any`.
 
 * :ref:`Validator decorators <attrs:examples_validators>`
   and `default decorators <http://www.attrs.org/en/stable/examples.html#defaults>`_
@@ -249,7 +250,7 @@ Your CI script might work like this:
 * Create a tarball from the ``.mypy_cache`` directory.
 
 * Determine the current git master branch commit id (say, using
-  ``git rev-parse HEAD``).
+  :code:`git rev-parse HEAD`).
 
 * Upload the tarball to the shared repository with a name derived from the
   commit id.
@@ -348,13 +349,14 @@ Extended Callable types
    This feature is deprecated.  You can use
    :ref:`callback protocols <callback_protocols>` as a replacement.
 
-As an experimental mypy extension, you can specify ``Callable`` types
-that support keyword arguments, optional arguments, and more.  When
-you specify the arguments of a Callable, you can choose to supply just
-the type of a nameless positional argument, or an "argument specifier"
-representing a more complicated form of argument.  This allows one to
-more closely emulate the full range of possibilities given by the
-``def`` statement in Python.
+As an experimental mypy extension, you can specify
+:py:data:`~typing.Callable` types that support keyword arguments,
+optional arguments, and more. When you specify the arguments of a
+``Callable``, you can choose to supply just the type of a nameless
+positional argument, or an "argument specifier" representing a
+more complicated form of argument.  This allows one to more closely
+emulate the full range of possibilities given by the ``def`` statement
+in Python.
 
 As an example, here's a complicated function definition and the
 corresponding ``Callable``:
@@ -434,7 +436,7 @@ purpose:
 In all cases, the ``type`` argument defaults to ``Any``, and if the
 ``name`` argument is omitted the argument has no name (the name is
 required for ``NamedArg`` and ``DefaultNamedArg``).  A basic
-``Callable`` such as
+:py:data:`~typing.Callable` such as
 
 .. code-block:: python
 
@@ -446,7 +448,7 @@ is equivalent to the following:
 
    MyFunc = Callable[[Arg(int), Arg(str), Arg(int)], float]
 
-A ``Callable`` with unspecified argument types, such as
+A :py:data:`~typing.Callable` with unspecified argument types, such as
 
 .. code-block:: python
 
--- a/docs/source/builtin_types.rst
+++ b/docs/source/builtin_types.rst
@@ -22,19 +22,22 @@ Type                   Description
 ``Any``                dynamically typed value with an arbitrary type
 ====================== ===============================
 
-The type ``Any`` and type constructors such as ``List``, ``Dict``,
-``Iterable`` and ``Sequence`` are defined in the ``typing`` module.
+The type :py:data:`~typing.Any` and type constructors such as :py:class:`~typing.List`,
+:py:class:`~typing.Dict`, :py:class:`~typing.Iterable` and :py:class:`~typing.Sequence`
+are defined in the :py:mod:`typing` module.
 
 The type ``Dict`` is a *generic* class, signified by type arguments within
 ``[...]``. For example, ``Dict[int, str]`` is a dictionary from integers to
 strings and ``Dict[Any, Any]`` is a dictionary of dynamically typed
 (arbitrary) values and keys. ``List`` is another generic class. ``Dict`` and
-``List`` are aliases for the built-ins ``dict`` and ``list``, respectively.
+``List`` are aliases for the built-ins :py:class:`dict` and :py:class:`list`,
+respectively.
 
-``Iterable``, ``Sequence``, and ``Mapping`` are generic types that
-correspond to Python protocols. For example, a ``str`` object or a
-``List[str]`` object is valid
-when ``Iterable[str]`` or ``Sequence[str]`` is expected. Note that even though
-they are similar to abstract base classes defined in ``collections.abc``
-(formerly ``collections``), they are not identical, since the built-in
-collection type objects do not support indexing.
+``Iterable``, ``Sequence``, and :py:class:`~typing.Mapping` are
+generic types that correspond to Python protocols. For example,
+a ``str`` object or a ``List[str]`` object is valid when
+``Iterable[str]`` or ``Sequence[str]`` is expected.
+Note that even though they are similar to abstract base classes
+defined in :py:mod:`collections.abc` (formerly ``collections``),
+they are not identical, since the built-in collection type objects
+do not support indexing.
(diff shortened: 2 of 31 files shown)


hoefling:
Open points:

1. First of all, I'm not sure whether you find the suggestion in this PR worth it at all. While #7623 is IMO important, this one is not. Especially that it will probably become a hassle to review as I have read and edited through the whole documentation.
2. I have not replaced all references to basic stuff like `int`, `float`, `str`, `bytes` etc as I'm unsure whether this wouldn't be an overkill - however, this can easily be done with `sed`, e.g. for `int`:
```find docs/source -name "*.rst" | xargs sed -i 's/\`\`int\`\`/:py:class:`int`/g'```
3. I haven't added references to expressions with types, like `Mapping[str, int]`, would want to ask for permission first. Sometimes a reference is context-dependent, e.g. in the text:
   ```
   the parameter type ``slice`` and the return type ``Sequence[T]`` are compatible with ``Union[int, slice]`` and ``Union[T, Sequence]``
   ```
   linking both `Union[int, slice]` and `Union[T, Sequence]` to https://docs.python.org/3/library/typing.html#typing.Union doesn't add to readability IMO.

   However, in another example:
   ```
   result of calling an ``async def`` function *without awaiting* will be a value of type ``typing.Coroutine[Any, Any, T]``, which is a subtype of ``Awaitable[T]``
   ```
   linking both `Coroutine` and `Awaitable` would increase the readability.
4. I have used the Python 3.7 docs (current release) for referencing. With Python 3.8, the `typing` module will be extended with more features from `typing_extensions` (`TypedDict`, `Literal`, `Protocol` etc). When using Python 3.8 docs already, I could reference those as well. Also, right now the description of `TypedDict`, `Protocol` and `Literal` suggests to install `typing_extensions` - I could add notes mentioning their availability in 3.8, similar to the one I found under [Simple user-defined protocols](https://mypy.readthedocs.io/en/latest/protocols.html#simple-user-defined-protocols).

**Comment by gvanrossum** [comment]
Also, I merged your first PR, so this one now has conflicts to resolve. (Probably just remove the first commit.)

**Comment by hoefling** [comment]
@gvanrossum first of all, thank you for reviewing the other PR! I'm not happy with this first iteration myself - played too much with `sed` at the beginning and often inserted the refs just because they _can_ be inserted, not because they _should_ be inserted. Let me reiterate this until Monday. I would propose the following rules:

1. Most common terms (`int`, `None`, `Any` etc) are referenced once per document, first occurence only.
2. Less common terms (`list`, `typing.Dict`, third-party libraries etc) are referenced once per chapter/section, first occurence only. Ideally, the same term is referenced once per each screen no matter what page anchor was selected. This way, one can easily spot a needed reference if coming to the docs from google search.

Also, I have discovered [`doc8`](https://github.com/PyCQA/doc8) for checking documentation, would fix some errors it emits.

**Comment by gvanrossum** [comment]
Thanks, those rules seem reasonable, except I would personally categorize `list` as "most common terms". (But not `typing.Dict`.)

**Comment by hoefling** [comment]
I think I have it ready for a review now; sadly, these are still a lot of changes. However, all refs are valid as Sphinx doesn't emit any warnings when building docs. To verify locally, install Sphinx via `pip install sphinx` and issue `sphinx-build -Ean docs/source docs/build` to force a complete rebuild and run checks in nitpicky mode.

I have also double-checked the changes to existing formatting are minimal. For example, to reference `isinstance()` I used
```
:py:func`isinstance`
```
which retains the parentheses. While for `isinstance`, a more complex ref
```
:py:func:`isinstance <isinstance>`
```
writes the referenced text as-is, so no parentheses are added.

I have also discovered and fixed minor formatting issue like [using markdown instead of ReST](https://github.com/python/mypy/pull/7624/files#diff-1602a5635cd0bfc20012571b8c1b0892L188-L194). This adds up to the total diffs, sadly.

I have decided not to reference `None` at all, as advised by @gvanrossum; common classes like `int`, `str`, `Any` etc. are referenced once per `.rst` document. The rest of entities in `typing` module and third-party entities, if occurring often, are referenced once per chapter or enough to be visible once per screen. Example:

image

### Last but not least: refs to py38 features

I have also referenced the new features in Python 3.8 (`TypedDict`, `Protocol`, `Literal`, `Final` etc) against the 3.8 docs. Thus:

1. The imports from `typing_extensions` in examples had to be updated too ([example](https://github.com/python/mypy/pull/7624/files#diff-6671e4e194a310d82efeca7d2f55670cL426-L429)). I have tested the changed examples out, but the error messages often differ from the ones in the examples - I hope this is no big deal. If it is, I would propose to revisit the examples in another PR.
2. I have added or updated the notes reflecting the 3.8 changes. Relevant diffs:
  * [`final_attrs.rst`](https://github.com/python/mypy/pull/7624/files#diff-d2fe6c83ea4dc73425c0ea808faa535fL16-L21),
  * [`literal_types.rst`](https://github.com/python/mypy/pull/7624/files#diff-8aad34cb1befce58fd1c49da98551da4L6-L12),
  * [`more_types.rst`](https://github.com/python/mypy/pull/7624/files#diff-8cf30e02200fbb72d0cccb51520db332L835-L838)

Is this too soon / should this be rather done in a separate PR?

**Comment by hoefling** [comment]
Additional ideas:

### doctest code examples?

In our company's docs, we require that each example is implemented in a [`.. doctest::` directive](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-doctest) (using `.. code-block:: python` is actually prohibited). This way, we execute doctests with Sphinx on code changes and verify the examples in the docs still work. Marrying doctests with `mypy`, however, would require non-trivial changes, probably best would be adding a custom doctest parser that will parse code in doctest blocks and lint them with `mypy` - e.g. with [Sybil](https://sybil.readthedocs.io/en/latest/parsers.html#developing-your-own-parsers). This may be not worth it, depending on whether you have to edit the examples often when the code changes.

### API docs?

I saw `mypy` mentions its own code in the [Extending and integrating mypy](https://mypy.readthedocs.io/en/latest/extending_mypy.html) chapter. Sphinx can generate an API docs section from code (Sphinx has a built-in [`autodoc` extension](http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) for that), so `mypy.api.run` or `mypy.plugin.Plugin` can be referenced to internal API and the docstrings from code will be automatically inserted in the docs. This will e.g. cover typos like `mypy.plugins.Plugin` in the [High-level overview](https://mypy.readthedocs.io/en/latest/extending_mypy.html#high-level-overview) section (fixed that already in this PR). However, this suggestion may be too soon since `mypy`s API is not stable yet.

**Comment by hoefling** [comment]
Here's [an example documentation](http://pypi-mirror.herokuapp.com/root/pypi/mypy/0.730+dev.c610a312853921d09083f7525c183810e4232056.dirty/+doc/extending_mypy.html#api) showing how the rendered API could look like, with a minimal subset of the codebase made available.

**Comment by gvanrossum** [comment]
Before I go review this in detail, let me suggest that you try to focus on a single issue per PR. If you have additional ideas on which you'd like to receive feedback (e.g. API docs) please open a new issue. If you have fixes that are unrelated to the core change here (making things link to the Python docs), submit them as a separate PR (we can land those immediately).

I would also skip linking to Python 3.8 -- even after that's released, most people will not immediately upgrade, so `typing_extensions` still remains relevant.

Finally, you have a few links (e.g. `https://github.com/python/mypy/pull/7624/files#diff-6671e4e194a310d82efeca7d2f55670cL426-L429`) that don't point anywhere interesting -- this doesn't scroll to anything that looks like L426-L429.

**Comment by gvanrossum** [comment]
Another suggestion: don't be too clever with not linking to the docs more than once per page / section / file / whatever. The non-link markup actually looks more hideous (red) than the link markup (black bold). And if you look through Python's own docs, in general when something is linked, there's no fear to link to it multiple times. (Example: in `Doc/library/dataclasses.rst`, every occurrence of `__init__` is spelled as
```
:meth:`__init__`
```
I haven't checked if the Python docs are consistent in this (it's all done manually and I'm sure different authors use slightly different conventions) but I'd say stick to the KISS principle (Keep It Simple, Stupid), which in this case I interpret as "for a particular term, either never link or always link".

File: docs/source/additional_features.rst [hunk 1]
```diff
@@ -140,9 +141,9 @@ If you're using ``auto_attribs=True`` you must use variable annotations.
         three: int = attr.ib(8)
 
 Typeshed has a couple of "white lie" annotations to make type checking
-easier. ``attr.ib`` and ``attr.Factory`` actually return objects, but the
-annotation says these return the types that they expect to be assigned to.
-That enables this to work:
+easier. :py:func:`attr.ib` and :py:class:`attr.Factory` actually
+return objects, but the annotation says these return the types that
+they expect to be assigned to. That enables this to work:
```
gvanrossum: Diff chunks like this one look worse than they are because the paragraph was reformatted. We don't particularly care about line lengths in the doc sources. Perhaps the diff would be simpler if you just let the lines become longer, and kept the line breaks where they were? (GitHub has a nice feature where it uses two different shares of red and green to indicate text that was just moved vs. text that was actually changed, but reflowing a paragraph is considered to be an actual change. :-( )

docs/source/additional_features.rst
```diff
@@ -348,13 +349,14 @@ Extended Callable types
...
-that support keyword arguments, optional arguments, and more.  When
-you specify the arguments of a Callable, you can choose to supply just
-the type of a nameless positional argument, or an "argument specifier"
-representing a more complicated form of argument.  This allows one to
-more closely emulate the full range of possibilities given by the
-``def`` statement in Python.
+As an experimental mypy extension, you can specify
+:py:data:`~typing.Callable` types that support keyword arguments,
+optional arguments, and more. When you specify the arguments of a
+``Callable``, you can choose to supply just the type of a nameless
```
gvanrossum: As an example of the KISS principle I mentioned, this second occurrence of `Callable` could easily be linked as well, even though it's in the same paragraph as the first.

docs/source/builtin_types.rst
```diff
@@ -22,19 +22,22 @@ Type                   Description
 ``Any``                dynamically typed value with an arbitrary type
 ====================== ===============================
 
-The type ``Any`` and type constructors such as ``List``, ``Dict``,
-``Iterable`` and ``Sequence`` are defined in the ``typing`` module.
+The type :py:data:`~typing.Any` and type constructors such as :py:class:`~typing.List`,
```
gvanrossum: I am thinking that all the types mentioned here are just too common to bother linking to their docs. I would only link the `typing` module itself.

Another idea: just above here is a table showing the most common built-in types. Maybe the items in the table should be linked instead?

docs/source/casts.rst
```diff
@@ -34,8 +34,8 @@ quite understand what is going on.
           assert isinstance(o, int)
           print(o + 5)  # OK: type of 'o' is 'int' here
 
-You don't need a cast for expressions with type ``Any``, or when
-assigning to a variable with type ``Any``, as was explained earlier.
+You don't need a cast for expressions with type :py:data:`~typing.Any`,
```
gvanrossum: Again, `Any` should probably not be linked. It's just too common. Also, the [explanation of this type in the Python docs](https://docs.python.org/3/library/typing.html#typing.Any) is rather lame. If anything, it should probably link to an explanation of the concept in the mypy docs (if there is one), or perhaps in [PEP 483](https://www.python.org/dev/peps/pep-0483/#id8).

docs/source/class_basics.rst
```diff
@@ -24,8 +25,8 @@ initialized within the class. Mypy infers the types of attributes:
    a.y = 3  # Error: 'A' has no attribute 'y'
 
 This is a bit like each class having an implicitly defined
-``__slots__`` attribute. This is only enforced during type
-checking and not when your program is running.
+:py:data:`__slots__ <object.__slots__>` attribute. This is only
```
gvanrossum: This is an example of one where I *do* think the link is useful.

hoefling:
Thank you for guiding me through this - I have never contributed anything more significant than a bugfix, so I'm still learning to find the right balance between _"this change should be included no matter what"_ and _"any sane person should be able to review it in the end"_. I suggest to keep this PR open while I land more related PRs and once I verified each change has found its place, I will simply close this one (and open new issues for proposals that I'm not sure whether or how to solve).

docs/source/builtin_types.rst
hoefling: I thought about linking the table items instead, but the classes listed contain types, e.g. `Tuple[int, int]`. The `typing` module docs do not link those, e.g. [`Optional` docs](https://docs.python.org/3/library/typing.html#typing.Optional). I guess it counts as an example, which are not linked in the stdlib docs.

docs/source/additional_features.rst
hoefling: This makes the job a lot easier, knowing this!

hoefling:
Closing this since the changes were reworked and landed in linked PRs.

PR #7634, opened 2019-10-05
[mypyc] Reorganization of headers and declarations

msullivan:
This splits the headers we generate into internal headers (containing
type declarations) and external headers (containing function and
object declarations) and refactors declarations to be much more driven
by HeaderDeclaration objects. This will allow us to use
HeaderDeclaration information to populate structures for dynamic
linking in separate compilation.

### DIFF
--- a/mypyc/emit.py
+++ b/mypyc/emit.py
@@ -29,7 +29,8 @@ class HeaderDeclaration:
       decl: C source code for the declaration.
       defn: Optionally, C source code for a definition.
       dependencies: The names of any objects that must be declared prior.
-      is_type: Whether the declaration is of a type.
+      is_type: Whether the declaration is of a C type. (C types will be declared in
+               external header files and not marked 'extern'.)
     """
 
     def __init__(self,
--- a/mypyc/emit.py
+++ b/mypyc/emit.py
@@ -20,13 +20,29 @@ from mypyc.sametype import is_same_type
 
 
 class HeaderDeclaration:
+    """A representation of a declaration in C.
+
+    This is used to generate declarations in header files and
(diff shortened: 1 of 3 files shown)


mypyc/emitmodule.py
```diff
@@ -82,9 +82,11 @@ def compile_modules_to_c(result: BuildResult, module_names: List[str],
 
 
 def generate_function_declaration(fn: FuncIR, emitter: Emitter) -> None:
-    emitter.emit_line('{};'.format(native_function_header(fn.decl, emitter)))
+    emitter.context.declarations[emitter.native_function_name(fn.decl)] = HeaderDeclaration(
+        '{};'.format(native_function_header(fn.decl, emitter)))
     if fn.name != TOP_LEVEL_NAME:
-        emitter.emit_line('{};'.format(wrapper_function_header(fn, emitter.names)))
+        emitter.context.declarations[PREFIX+fn.cname(emitter.names)] = HeaderDeclaration(
```
JukkaL: Style nit: use spaces around `+`.

mypyc/emit.py
```diff
@@ -20,13 +20,29 @@
...
+    """A representation of a declaration in C.
+
+    This is used to generate declarations in header files and
+    (optionally) definitions in source files.
+
+    Attributes:
+      decl: C source code for the declaration.
+      defn: Optionally, C source code for a definition.
+      dependencies: The names of any objects that must be declared prior.
+      is_type: Whether the declaration is of a type.
```
JukkaL: Be more explicit about what we mean by 'type' here, since it's a fuzzy concept.

PR #7636, opened 2019-10-05
[mypyc] Support separate compilation

msullivan:
This adds support for separate compilation to mypyc.

A `separate` argument is added to `mypycify`. If `True`, mypyc places every
module its own separate shared library. Otherwise it can take a list of groups
of source files that should be placed in a shared library together.

The shared libraries communication with each other using the C API's intended
mechanism for communicating between C extension modules: Capsules.
Each library creates a table of pointers to all of its exported data and stores a pointer
to it in a capsule stored as a module attribute. When a library is loaded, it
loads the linking table capsules from all its dependencies and copies them
into a local copy of the table. (To eliminate the need for a pointer indirection when
accessing it.)

This adds a test mode that will run all multi-module run tests in separate compilation mode.
I also manually tested mypy itself compiled in separate compilation mode.

This supports a limited form of incremental compilation already: only modules
that changed or had a header they depend on change will be recompiled by the
C compiler. The entire project still needs to go through the entire mypy/mypyc
front-end and middle-end, however. I expect to have a PR that adds support for 
hooking into incremental mode next week.

This is progress on mypyc/mypyc#682.

### DIFF
--- a/mypyc/build.py
+++ b/mypyc/build.py
@@ -188,19 +188,11 @@ def generate_c(sources: List[BuildSource], options: Options,
 
     errors = Errors()
 
-    mapper = emitmodule.prepare_groups(result, all_module_names,
-                                       compiler_options=compiler_options,
-                                       errors=errors)
-
     ops = []  # type: List[str]
-    ctext = []  # type: List[List[Tuple[str, str]]]
-    for group_sources, shared_lib_name in groups:
-        module_names = [source.module for source in group_sources]
-        ctext.append(emitmodule.compile_modules_to_c(result, module_names, shared_lib_name,
-                                                     mapper,
-                                                     compiler_options=compiler_options,
-                                                     errors=errors, ops=ops,
-                                                     groups=groups))
+    ctext = emitmodule.compile_modules_to_c(result,
+                                            compiler_options=compiler_options,
+                                            errors=errors, ops=ops,
+                                            groups=groups)
     if errors.num_errors:
         errors.flush_errors()
         sys.exit(1)
@@ -288,14 +280,15 @@ def write_file(path: str, contents: str) -> None:
             f.write(contents)
 
 
-def mypycify(paths: List[str],
-             mypy_options: Optional[List[str]] = None,
-             opt_level: str = '3',
-             multi_file: bool = False,
-             skip_cgen: bool = False,
-             verbose: bool = False,
-             strip_asserts: bool = False,
-             separate: Union[bool, List[Tuple[List[str], Optional[str]]]] = False,
+def mypycify(
+    paths: List[str],
+    mypy_options: Optional[List[str]] = None,
+    opt_level: str = '3',
+    multi_file: bool = False,
+    skip_cgen: bool = False,
+    verbose: bool = False,
+    strip_asserts: bool = False,
+    separate: Union[bool, List[Tuple[List[str], Optional[str]]]] = False,
 ) -> List[Extension]:
     """Main entry point to building using mypyc.
 
@@ -336,7 +329,9 @@ def mypycify(paths: List[str],
     use_shared_lib = len(sources) > 1 or any('.' in x.module for x in sources)
 
     if separate is True:
-        groups = [([source], None) for source in sources]  # type: List[Tuple[List[BuildSource], Optional[str]]]
+        groups = [  # type: List[Tuple[List[BuildSource], Optional[str]]]
+            ([source], None) for source in sources
+        ]
     elif isinstance(separate, list):
         groups = []
         for files, name in separate:
--- a/mypyc/build.py
+++ b/mypyc/build.py
@@ -280,12 +280,40 @@ def write_file(path: str, contents: str) -> None:
             f.write(contents)
 
 
+def construct_groups(
+    sources: List[BuildSource],
+    separate: Union[bool, List[Tuple[List[str], Optional[str]]]],
+    use_shared_lib: bool,
+) -> List[Tuple[List[BuildSource], Optional[str]]]:
+    if separate is True:
+        groups = [
+            ([source], None) for source in sources
+        ]  # type: List[Tuple[List[BuildSource], Optional[str]]]
+    elif isinstance(separate, list):
+        groups = []
+        for files, name in separate:
+            groups.append((
+                [src for src in sources if src.path in files],
+                name))
+    else:
+        groups = [(sources, None)]
+
+    # Generate missing names
+    groups = [
+        (group,
+         name or (
+             shared_lib_name([source.module for source in group]) if use_shared_lib else None))
+        for group, name in groups]
+
+    return groups
+
+
 def mypycify(
     paths: List[str],
     mypy_options: Optional[List[str]] = None,
     opt_level: str = '3',
     multi_file: bool = False,
-    skip_cgen: bool = False,
+    skip_cgen_input: Optional[Any] = None,
     verbose: bool = False,
     strip_asserts: bool = False,
     separate: Union[bool, List[Tuple[List[str], Optional[str]]]] = False,
@@ -328,46 +356,28 @@ def mypycify(
     # around with making the single module code handle packages.)
     use_shared_lib = len(sources) > 1 or any('.' in x.module for x in sources)
 
-    if separate is True:
-        groups = [  # type: List[Tuple[List[BuildSource], Optional[str]]]
-            ([source], None) for source in sources
-        ]
-    elif isinstance(separate, list):
-        groups = []
-        for files, name in separate:
-            groups.append((
-                [src for src in sources if src.path in files],
-                name))
-    else:
-        groups = [(sources, None)]
+    groups = construct_groups(sources, separate, use_shared_lib)
 
-    # Generate missing names
-    groups = [
-        (group,
-         name or (
-             shared_lib_name([source.module for source in group]) if use_shared_lib else None))
-        for group, name in groups]
-
-    # We let the test harness make us skip doing the full compilation
+    # We let the test harness just pass in the c file contents instead
     # so that it can do a corner-cutting version without full stubs.
-    # TODO: Be able to do this based on file mtimes?
-    if not skip_cgen:
+    if not skip_cgen_input:
         group_cfiles, ops_text = generate_c(sources, options, groups,
                                             compiler_options=compiler_options)
         # TODO: unique names?
         with open(os.path.join(build_dir, 'ops.txt'), 'w') as f:
             f.write(ops_text)
-        group_cfilenames = []  # type: List[List[str]]
-        for cfiles in group_cfiles:
-            group_cfilenames.append([])
-            for cfile, ctext in cfiles:
-                cfile = os.path.join(build_dir, cfile)
-                write_file(cfile, ctext)
-                if os.path.splitext(cfile)[1] == '.c':
-                    group_cfilenames[-1].append(cfile)
     else:
-        assert len(groups) == 1
-        group_cfilenames = [glob.glob(os.path.join(build_dir, '*.c'))]
+        group_cfiles = skip_cgen_input
+
+    # Write out the generated C and collect the files for each group
+    group_cfilenames = []  # type: List[List[str]]
+    for cfiles in group_cfiles:
+        group_cfilenames.append([])
+        for cfile, ctext in cfiles:
+            cfile = os.path.join(build_dir, cfile)
+            write_file(cfile, ctext)
+            if os.path.splitext(cfile)[1] == '.c':
+                group_cfilenames[-1].append(cfile)
 
     cflags = []  # type: List[str]
     if compiler.compiler_type == 'unix':
--- a/mypyc/build.py
+++ b/mypyc/build.py
@@ -162,7 +162,7 @@ def include_dir() -> str:
 def generate_c(sources: List[BuildSource], options: Options,
                groups: List[Tuple[List[BuildSource], Optional[str]]],
                compiler_options: Optional[CompilerOptions] = None
-               ) -> Tuple[List[Tuple[str, str]], str]:
+               ) -> Tuple[List[List[Tuple[str, str]]], str]:
     """Drive the actual core compilation step.
 
     Returns the C source code and (for debugging) the pretty printed IR.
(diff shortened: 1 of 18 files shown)


msullivan:
(I haven't squashed the commit history yet because it was useful for me in revising this PR and might be useful to me in addressing feedback. But it probably isn't useful to anybody reviewing this, because I made some changes in early commits that I undid in later commits and because github bizarrely insists on sorting commits by their timestamp.)

mypyc/build.py
```diff
@@ -160,14 +161,13 @@ def include_dir() -> str:
...
 
 def generate_c(sources: List[BuildSource], options: Options,
-               shared_lib_name: Optional[str],
+               groups: List[Tuple[List[BuildSource], Optional[str]]],
                compiler_options: Optional[CompilerOptions] = None
-               ) -> Tuple[List[Tuple[str, str]], str]:
+               ) -> Tuple[List[List[Tuple[str, str]]], str]:
     """Drive the actual core compilation step.
 
     Returns the C source code and (for debugging) the pretty printed IR.
```
JukkaL: Document `groups`.

mypyc/build.py
```diff
@@ -276,13 +283,57 @@ def write_file(path: str, contents: str) -> None:
...
-             opt_level: str = '3',
-             multi_file: bool = False,
-             skip_cgen: bool = False,
-             verbose: bool = False,
-             strip_asserts: bool = False) -> List[Extension]:
+def construct_groups(
+    sources: List[BuildSource],
+    separate: Union[bool, List[Tuple[List[str], Optional[str]]]],
+    use_shared_lib: bool,
+) -> List[Tuple[List[BuildSource], Optional[str]]]:
```
JukkaL: Add docstring.

mypyc/build.py
```diff
@@ -276,13 +283,57 @@ def write_file(path: str, contents: str) -> None:
...
+                name))
+    else:
+        groups = [(sources, None)]
+
+    # Generate missing names
+    groups = [
+        (group,
+         name or (
+             shared_lib_name([source.module for source in group]) if use_shared_lib else None))
+        for group, name in groups]
```
JukkaL: I found this a bit hard to understand. It might be easier to read if it used for and if statements.

mypyc/build.py
```diff
@@ -276,13 +283,57 @@ def write_file(path: str, contents: str) -> None:
...
+
+    return groups
+
+
+def get_header_deps(cfiles: List[Tuple[str, str]]) -> List[str]:
```
JukkaL: Document the `cfiles` argument.

mypyc/build.py
```diff
@@ -276,13 +283,57 @@ def write_file(path: str, contents: str) -> None:
...
+def mypycify(
+    paths: List[str],
+    mypy_options: Optional[List[str]] = None,
+    opt_level: str = '3',
+    multi_file: bool = False,
+    skip_cgen_input: Optional[Any] = None,
+    verbose: bool = False,
+    strip_asserts: bool = False,
+    separate: Union[bool, List[Tuple[List[str], Optional[str]]]] = False,
+) -> List[Extension]:
```
JukkaL: Document additional arguments -- at least `separate`.

mypyc/emit.py
```diff
@@ -30,27 +31,36 @@ class HeaderDeclaration:
...
                  ) -> None:
         self.decl = [decl] if isinstance(decl, str) else decl
         self.defn = defn
         self.dependencies = dependencies or set()
         self.is_type = is_type
+        self.needs_export = needs_export
 
 
 class EmitterContext:
     """Shared emitter state for an entire compilation unit."""
```
JukkaL: Maybe discuss what a compilation unit means in this context?

mypyc/emit.py
```diff
@@ -125,6 +135,28 @@ def new_label(self) -> str:
         self.context.temp_counter += 1
         return '__LL%d' % self.context.temp_counter
 
+    def get_lib_prefix(self, obj: Union[str, ClassIR, FuncDecl],
```
JukkaL: Document what a `str` `obj` argument means. Document `is_variable`.

mypyc/emitmodule.py
```diff
@@ -104,17 +124,26 @@ def encode_bytes_as_c_string(b: bytes) -> Tuple[str, int]:
     return '"{}"'.format(escaped), len(b)
 
 
+def pointerize(s: str, name: str) -> str:
```
JukkaL: Add docstring.

mypyc/emitmodule.py
```diff
@@ -104,17 +124,26 @@ def encode_bytes_as_c_string(b: bytes) -> Tuple[str, int]:
...
+        return s.replace(name, '*{}'.format(name))
+
+
 class ModuleGenerator:
     def __init__(self,
                  literals: LiteralsMap,
                  modules: List[Tuple[str, ModuleIR]],
                  source_paths: Dict[str, str],
                  shared_lib_name: Optional[str],
+                 group_map: Dict[str, Optional[str]],
```
JukkaL: Document `group_map` (and maybe other arguments).

mypyc/emitmodule.py
```diff
@@ -217,16 +257,44 @@ def generate_c_for_modules(self) -> List[Tuple[str, str]]:
...
-                                ('__native.h', ''.join(ext_declarations.fragments)),
+        return file_contents + [('__native{}.c'.format(self.lib_suffix),
+                                 ''.join(emitter.fragments)),
+                                ('__native_internal{}.h'.format(self.lib_suffix),
+                                 ''.join(declarations.fragments)),
+                                ('__native{}.h'.format(self.lib_suffix),
+                                 ''.join(ext_declarations.fragments)),
                                 ]
 
+    def generate_linking_struct(self, decl_emitter: Emitter, code_emitter: Emitter) -> None:
```
JukkaL: Add docstring. Document the linking table somewhere (both the purpose and the structure).

mypyc/emitmodule.py
```diff
@@ -244,34 +312,66 @@ def generate_shared_lib_init(self, emitter: Emitter) -> None:
...
             '',
         )
 
+        emitter.emit_lines(
+            'capsule = PyCapsule_New(&exports, "{}.exports", NULL);'.format(
+                self.shared_lib_name),
+            'if (!capsule) {',
+            'goto fail;',
+            '}',
+            'res = PyObject_SetAttrString(module, "exports", capsule);',
```
JukkaL: Does this get added to the module namespace? What if there's an attribute with name `exports`?

mypyc/emitmodule.py
```diff
@@ -244,34 +312,66 @@ def generate_shared_lib_init(self, emitter: Emitter) -> None:
...
+            'return module;',
+            '}',
+            'module = PyModule_Create(&def);',
             'if (!module) {',
             'goto fail;',
             '}',
             '',
         )
 
+        emitter.emit_lines(
```
JukkaL: Document the whole capsule business in detail somewhere.

mypyc/test/test_run.py
```diff
@@ -251,3 +245,13 @@ class TestRunMultiFile(TestRun):
         'run-multimodule.test',
         'run-mypy-sim.test',
     ]
+
+
+# Run the main multi-module tests in separate compliation mode
+class TestRunSeparate(TestRun):
```
JukkaL: Would it be useful to add some new test cases specific to separate compilation mode, or are the existing tests good enough?

mypyc/emitmodule.py
```diff
@@ -76,16 +87,25 @@ def compile_modules_to_c(result: BuildResult, module_names: List[str],
     # Generate C code.
     source_paths = {module_name: result.files[module_name].path
                     for module_name in module_names}
-    generator = ModuleGenerator(literals, modules, source_paths, shared_lib_name,
-                                compiler_options.multi_file)
-    return generator.generate_c_for_modules()
+
+    # Generate all the groups
```
JukkaL: Explain this in a bit more detail?

mypyc/emitmodule.py
```diff
@@ -42,16 +42,27 @@ def parse_and_typecheck(sources: List[BuildSource], options: Options,
...
-def compile_modules_to_c(result: BuildResult, module_names: List[str],
-                         shared_lib_name: Optional[str],
-                         compiler_options: CompilerOptions,
-                         errors: Errors,
-                         ops: Optional[List[str]] = None) -> List[Tuple[str, str]]:
+def compile_modules_to_c(
+    result: BuildResult,
+    compiler_options: CompilerOptions,
+    errors: Errors,
+    groups: List[Tuple[List[BuildSource], Optional[str]]],
```
JukkaL: Document the `groups` argument (and maybe some other arguments as well).

mypyc/emitmodule.py
msullivan: This is part of the module namespace for the shared library module, not the actual module.

mypyc/test/test_run.py
msullivan: I'm adding two cases I remember hitting to one of the tests, but otherwise they were pretty good.

It is probably worth adding testing of manually specifying group assignments, but that will require some infrastructure so I'll leave it to a follow-up diff I think. (I've tested that manually.)

mypyc/emitmodule.py
```diff
@@ -216,62 +304,151 @@ def generate_c_for_modules(self) -> List[Tuple[str, str]]:
...
+        stored as an attribute named 'exports' on the group's shared
+        library's python module.
+
+        On load, a group's init function will import all of its
+        dependencies' exports tables using the capsule mechanism and
+        copy the contents into a local copy of the table (to eliminate
+        the need for a pointer indirection when accessing it).
+
+        Then, all calls to functions in another group and accesses to statics
+        from another group are done indirectly via the export table.
```
JukkaL: Maybe give a concrete example of what the export table looks like, and another group can access and item from it (using C fragments)?

mypyc/emitmodule.py
```diff
@@ -216,62 +304,151 @@ def generate_c_for_modules(self) -> List[Tuple[str, str]]:
...
+                code_emitter.emit_line('&{},'.format(name))
+
+        code_emitter.emit_line('};')
+
     def generate_shared_lib_init(self, emitter: Emitter) -> None:
         """Generate the init function for a shared library.
 
-        A shared library contains all of the actual code for a set of modules.
+        A shared library contains all of the actual code for a
+        compilation group.
```
JukkaL: A more concrete explanation would be helpful. For each module in a group, we generate a shim with a name derived from the module name (?) and a shared library with a name that is not derived in a straightforward way from the module name, right? But if the group only has one module, do we need a separate shared lib?

msullivan: We need the separate shared lib for an annoying and perhaps somewhat subtle reason: the reading of export tables from dependencies is done as a separate phase from actually executing module top levels, because I wanted to make sure that the export tables were always initialized and couldn't get mixed up trouble due to import cycles.

But:
 1. I think I have figured out a way to keep that property while allowing just one extension module per python module, though it is somewhat unpleasant
 2. Maybe it isn't actually important anyway

I'll document what's going on here a bit more and consider one of those alternatives in a follow-up maybe

mypyc/namegen.py
```diff
@@ -36,18 +36,22 @@ class NameGenerator:
     The generated should be internal to a build and thus the mapping is
     arbitrary. Just generating names '1', '2', ... would be correct,
     though not very usable.
+
+    FIXME: Currently when performing seperate compilation we just
+    totally give up on shortening names.
```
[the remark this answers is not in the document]
msullivan: Mostly just a little more verbose.

mypyc/emit.py
```diff
@@ -126,6 +136,30 @@ def new_label(self) -> str:
...
+        If the module lives is in the current compilation group, there is
+        no prefix.  But if it lives in a different group (and hence a separate
+        extension module), we need to access objects from it indirectly via an
+        export table.
+        """
+        groups = self.context.group_map
+        target_group_name = groups.get(module_name)
+        if target_group_name and target_group_name != self.context.group_name:
+            self.context.group_deps.add(target_group_name)
+            return 'exports_{}.'.format(target_group_name)
```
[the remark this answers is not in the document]
msullivan: Yeah, I will change the names to be based on the real names for the auto-generated single-module groups.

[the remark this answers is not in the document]
msullivan: Aha, and if I do that and also fix NameGenerator to work at a group level, I think everything works out pretty nicely here.

msullivan:
OK I think I have addressed all the feedback. There was also some function renaming.

One major pair of substantive changes went in, which is to generate group names from single-module groups in a legible manner and to restore namegen module name shortening. This depends on https://github.com/python/mypy/pull/7676 now, though.

mypyc/emitmodule.py
```diff
@@ -73,17 +131,32 @@ def compile_modules_to_c(result: BuildResult, module_names: List[str],
...
                     for module_name in module_names}
-    generator = ModuleGenerator(literals, modules, source_paths, shared_lib_name,
-                                compiler_options.multi_file)
-    return generator.generate_c_for_modules()
+
+    names = NameGenerator([[source.module for source in sources] for sources, _ in groups])
+
+    # Generate C code for each compilation group. Each group will be
+    # compiled into a separate extension module, so we produce
+    # a separate group of C source for each of them.
```
JukkaL: "Group of C source" is not entirely clear. Do we generate a separate file for each file in a group?

mypyc/emitmodule.py
```diff
@@ -104,32 +177,62 @@ def encode_bytes_as_c_string(b: bytes) -> Tuple[str, int]:
     return '"{}"'.format(escaped), len(b)
 
 
-class ModuleGenerator:
+def pointerize(s: str, name: str) -> str:
+    """Given C type and a variable name, return a declaration of a pointer to it."""
```
JukkaL: I'm not sure if the docstring matches the implementation. It looks like `s` should have both a type and a variable name, since we are replacing `name` in it. Also, is it possible that there will be multiple matches to replace? I assume that the variable name needs to be sufficiently unique.

mypyc/emitmodule.py
```diff
@@ -104,32 +177,62 @@ def encode_bytes_as_c_string(b: bytes) -> Tuple[str, int]:
...
                  modules: List[Tuple[str, ModuleIR]],
                  source_paths: Dict[str, str],
-                 shared_lib_name: Optional[str],
+                 group_name: Optional[str],
+                 group_map: Dict[str, Optional[str]],
+                 names: NameGenerator,
                  multi_file: bool) -> None:
+        """Generator for C source for a compilation group.
+
+        We do this by just regexping the source, which is a bit simpler than
```
JukkaL: Explain what the generated C source includes. This can be understood (based on just this docstring) as being some grepped subset of some other source. Also this would be clearer if the distinction between the "source source" and generated source was more explicit. Maybe a few more sentences would help here?

JukkaL:
Oh and there is a build failure.

mypyc/emitmodule.py
msullivan: Oh lurr that regexping comment got copy-pasted from elsewhere and is just wrong

PR #7637, opened 2019-10-05
Updates attrs.py to correct an error message for attrs converter resolves #6172

### LINKED ISSUE #6172: "Unsupported converter" for completely unanotated function
A slightly different case with same error message just popped up with mypy==0.650:

```py
import attr


def foo(value):
    return value.split()


@attr.s
class Bar:
    field = attr.ib(default=None, converter=foo)
```

```bash
> mypy test_mypy.py
test_mypy.py:10: error: Unsupported converter, only named functions and types are currently supported
```

samyak-jn:
This commit is in reference to the #6172 issue, which corrects information given by the error at the time of attrs converters.

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -559,7 +559,7 @@ def _parse_converter(ctx: 'mypy.plugin.ClassDefContext',
 
         # Signal that we have an unsupported converter.
         ctx.api.fail(
-            "Unsupported converter, only types and named functions with an annotated return type are currently supported",
+            "Unsupported converter, only types and named functions with annotated return type are supported",
             converter
         )
         return Converter('')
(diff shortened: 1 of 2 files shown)


samyak-jn:
@anyone, can help me with why the check is failing?

JelleZijlstra:
Looks like there's a few issues, all of which should be easy to fix: a line of code is too long and a few tests need to be updated to reflect the new error message. You can see the exact test failures by clicking through the "Details" link.

mypy/plugins/attrs.py
```diff
@@ -559,7 +559,7 @@ def _parse_converter(ctx: 'mypy.plugin.ClassDefContext',
 
         # Signal that we have an unsupported converter.
         ctx.api.fail(
-            "Unsupported converter, only named functions and types are currently supported",
```
msullivan: Could we go witih "Unsupported converter; only types and annotated named functions are currently supported"

PR #7639, opened 2019-10-05
[mypyc] Invoke __init__ functions inherited from non-ext classes

msullivan:
Currently if a class inherits from a nonext class or an interpreted
class and doesn't define an __init__, the parent's __init__ is not
invoked. Fix this.


msullivan:
(This helps lay groundwork for more efficient dataclasses, also, since those need to use the generated `__init__`)

PR #7641, opened 2019-10-05
Avoid creating partial annotations on attrs-classes __init__ method

### LINKED ISSUE #5954: Bogus "Function is missing a type annotation for one or more arguments" with attrs
When checking this file with mypy (0.641 or git master) and `--disallow-incomplete-defs`:

```python
import attr

@attr.s
class A:
    a = attr.ib()
```

I get an unexpected:

```
at.py:3: error: Function is missing a type annotation for one or more arguments
```

(line 3 being the `@attr.s` decorator). If I'm interpreting [what `--disallow-incomplete-defs` does](https://mypy.readthedocs.io/en/latest/command_line.html#untyped-definitions-and-calls) correctly, it should only warn if there's a partial type annotation (as opposed to none at all, which `--disallow-untyped-defs` would warn about).

Also, the error message is confusing - I don't define a function at all there. With `--show-error-context`, it says:

```
at.py: note: In member "__init__" of class "A":
```

FWIW, I (finally) started using mypy with a rather big project ([qutebrowser](https://qutebrowser.org/)) and I'm quite happy with it so far - the approach of gradual typing and the awesome documentation make it very easy to get a benefit without having to annotate the entire codebase at once.

Zac-HD:
This is important because `--disallow-incomplete-defs` is basically unusable otherwise, and I want to use it for e.g. HypothesisWorks/hypothesis#2071.

I'm well aware that this is a terrible hack, but it fixes #5954 for now.  @euresti had sketched out the proper fix but I just got lost in the codebase - hopefully someone will come back and fix it properly later, but for now this is good enough for me.

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -639,11 +639,21 @@ def _make_frozen(ctx: 'mypy.plugin.ClassDefContext', attributes: List[Attribute]
 def _add_init(ctx: 'mypy.plugin.ClassDefContext', attributes: List[Attribute],
               adder: 'MethodAdder') -> None:
     """Generate an __init__ method for the attributes and add it to the class."""
-    adder.add_method(
-        '__init__',
-        [attribute.argument(ctx) for attribute in attributes if attribute.init],
-        NoneType()
-    )
+    args = [attribute.argument(ctx) for attribute in attributes if attribute.init]
+    if all(
+        # We use getattr rather than instance checks because the variable.type
+        # might be wrapped into a Union or some other type, but even non-Any
+        # types reliably track the fact that the argument was not annotated.
+        getattr(arg.variable.type, "type_of_any", None) == TypeOfAny.unannotated
+        for arg in args
+    ):
+        # This workaround makes --disallow-incomplete-defs usable with attrs,
+        # but is definitely suboptimal as a long-term solution.
+        # See https://github.com/python/mypy/issues/5954 for discussion.
+        for a in args:
+            a.variable.type = AnyType(TypeOfAny.implementation_artifact)
+            a.type_annotation = AnyType(TypeOfAny.implementation_artifact)
+    adder.add_method('__init__', args, NoneType())
 
 
 class MethodAdder:
--- a/test-data/unit/check-attr.test
+++ b/test-data/unit/check-attr.test
@@ -75,7 +75,7 @@ A(1, [2], '3', 4, 5)  # E: Too many arguments for "A"
 # flags: --disallow-untyped-defs
 import attr
 @attr.s
-class A:  # E: Function is missing a type annotation for one or more arguments
+class A:
     a = attr.ib()  # E: Need type annotation for 'a'
     _b = attr.ib()  # E: Need type annotation for '_b'
     c = attr.ib(18)  # E: Need type annotation for 'c'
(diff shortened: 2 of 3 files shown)


mypy/plugins/attrs.py
```diff
@@ -640,11 +640,13 @@ def _make_frozen(ctx: 'mypy.plugin.ClassDefContext', attributes: List[Attribute]
 def _add_init(ctx: 'mypy.plugin.ClassDefContext', attributes: List[Attribute],
               adder: 'MethodAdder') -> None:
     """Generate an __init__ method for the attributes and add it to the class."""
-    adder.add_method(
-        '__init__',
-        [attribute.argument(ctx) for attribute in attributes if attribute.init],
-        NoneType()
-    )
+    args = [attribute.argument(ctx) for attribute in attributes if attribute.init]
+    if all(getattr(a.variable.type, "type_of_any", None) == TypeOfAny.unannotated for a in args):
```
euresti: You'll probably want to use isinstance checks and the like instead of getattr.

mypy/plugins/attrs.py
```diff
@@ -640,11 +640,13 @@ def _make_frozen(ctx: 'mypy.plugin.ClassDefContext', attributes: List[Attribute]
...
+        # This is a total hack , but makes --disallow-incomplete-defs usable with
```
euresti: Maybe I wouldn't say hack. Maybe something about how this is a workaround but that there is a more correct solution.

mypy/plugins/attrs.py
```diff
@@ -640,11 +640,18 @@ def _make_frozen(ctx: 'mypy.plugin.ClassDefContext', attributes: List[Attribute]
...
               adder: 'MethodAdder') -> None:
     """Generate an __init__ method for the attributes and add it to the class."""
-    adder.add_method(
-        '__init__',
-        [attribute.argument(ctx) for attribute in attributes if attribute.init],
-        NoneType()
-    )
+    args = [attribute.argument(ctx) for attribute in attributes if attribute.init]
+    if all(
+        isinstance(arg.variable.type, get_proper_type(AnyType))
```
euresti: Did you mean `isinstance(get_proper_type(arg.variable.type), AnyType)`

euresti:
Ok.  Maybe the mypy devs can comment on using `getattr`.  (Also, I don't know if I'm allowed to approve this code). That being said, your tests are failing.

mypy/plugins/attrs.py
```diff
@@ -640,11 +640,20 @@ def _make_frozen(ctx: 'mypy.plugin.ClassDefContext', attributes: List[Attribute]
...
+        # might be wrapped into a Union or some other type, but even non-Any
+        # types reliably track the fact that the argument was not annotated.
+        getattr(arg.variable.type, "type_of_any", None) == TypeOfAny.unannotated
+        for arg in args
+    ):
+        # This workaround makes --disallow-incomplete-defs usable with attrs,
+        # but is definitely suboptimal as a long-term solution.
+        # See https://github.com/python/mypy/issues/5954 for discussion.
+        for a in args:
+            a.variable.type = AnyType(TypeOfAny.implementation_artifact)
```
ilevkivskyi: I think test fails because you also need to reset `a.type_annotation`.

Zac-HD:
@ilevkivskyi - done!  Thanks for the suggestion, that made it *way* easier to fix 🙂

PR #7645, opened 2019-10-07
Allow TypedDict key with literal type during construction

### LINKED ISSUE #7644: Literal type not accepted as key when constructing TypedDict
@b0g3r reported an issue with TypedDict keys not being accepted with literal types. Example code from https://github.com/python/mypy/issues/6262#issuecomment-538520768:

```py
from typing_extensions import Literal, Final


class MyDict(TypedDict):
    field_name: int

d: MyDict
d = {'field_name': 1}  # OK

name_in_var = 'field_name'
d = {name_in_var: 1}  # Correct error

name_in_literal2: Literal['field_name'] = 'field_name'
d = {name_in_literal2: 1}  # Unexpected error

name_in_final: Final = 'field_name'
d = {name_in_final: 1}  # Unexpected error
```

JukkaL:
Also refactored an operation away from `mypy.plugins.common` so that it
is more cleanly reusable. Also kept an alias in the original location
to avoid breaking existing plugins.

Fixes #7644.

### DIFF
--- a/mypy/typeops.py
+++ b/mypy/typeops.py
@@ -10,10 +10,12 @@ from typing import cast, Optional, List, Sequence, Set
 from mypy.types import (
     TupleType, Instance, FunctionLike, Type, CallableType, TypeVarDef, Overloaded,
     TypeVarType, TypeType, UninhabitedType, FormalArgument, UnionType, NoneType,
-    AnyType, TypeOfAny, TypeType, ProperType, get_proper_type, get_proper_types, copy_type
+    AnyType, TypeOfAny, TypeType, ProperType, LiteralType, get_proper_type, get_proper_types,
+    copy_type
 )
 from mypy.nodes import (
-    FuncBase, FuncItem, OverloadedFuncDef, TypeInfo, TypeVar, ARG_STAR, ARG_STAR2,
+    FuncBase, FuncItem, OverloadedFuncDef, TypeInfo, TypeVar, ARG_STAR, ARG_STAR2, Expression,
+    StrExpr
 )
 from mypy.maptype import map_instance_to_supertype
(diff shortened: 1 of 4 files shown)


test-data/unit/check-typeddict.test
```diff
@@ -1883,3 +1883,25 @@ assert isinstance(u2, Mapping)
...
+v = {num: ''}  # E: Incompatible types (expression has type "str", TypedDict item "num" has type "int")
+
+bad: Final = 2
+v = {bad: 3}  # E: Expected TypedDict key to be string literal
+union: Literal['num', 'foo']
+v = {union: 2} # E: Expected TypedDict key to be string literal
+num2: Literal['num']
+v = {num2: 2}
+bad2: Literal['bad']
+v = {bad: 2}  # E: Expected TypedDict key to be string literal
```
ilevkivskyi: I suppose this should be `bad2`, otherwise `bad2` would be unused in this test.

mypy/typeops.py
```diff
@@ -417,3 +419,38 @@ def callable_type(fdef: FuncItem, fallback: Instance,
...
+    elif isinstance(typ, UnionType):
+        possible_literals = list(typ.items)
+    else:
+        possible_literals = [typ]
+
+    strings = []
+    for lit in get_proper_types(possible_literals):
+        if isinstance(lit, LiteralType) and lit.fallback.type.fullname() == 'builtins.str':
+            val = lit.value
+            assert isinstance(val, str)
```
b0g3r: It's not a review comment, just my curiosity: why is this needed?

JukkaL: Mypy thinks that the value could be an integer, for example, since it doesn't understand the `fullname()` check above.

PR #7647, opened 2019-10-07
Treat __class_getitem__ as an implicit classmethod like __init_subclass__

### LINKED ISSUE #7319: __class_getitem__ is not assumed to be a classmethod
mypy (version 0.720) gives confusing errors like this:

    error: The erased type of self "Type[_Type]" is not a supertype of its class "_Type"

for the following method:

```python
_TypeT = TypeVar("_TypeT", bound="_Type")
class _Type:
    def __class_getitem__(cls: Type[_TypeT], item: Any) -> _TypeT:
        raise NotImplementedError
```

even though this is [apparently the correct way](https://github.com/python/typing/issues/58#issuecomment-326240794) to type-hint factory functions.

Adding the `@classmethod` decorator makes the error go away, but [according to PEP 560](https://www.python.org/dev/peps/pep-0560/#class-getitem), `__class_getitem__` is automatically a class method, so the decorator shouln't be required.

AFanaei:
Fixes #7319


test-data/unit/fixtures/implicit_class_method.pyi
```diff
@@ -0,0 +1,13 @@
+# builtins stub with object.__init_subclass__ and __class_getitem__
+
+class object:
+    def __init_subclass__(cls) -> None: pass
+
+    def __class_getitem__(cls, item) -> None: pass
```
JukkaL: I think that `object` doesn't define `__class_getitem__`, so this method should probably be in some other class. Perhaps add a base class in your test case? If you do this, it looks like you don't need to touch the fixtures at all.

PR #7650, opened 2019-10-07
Fix ReST linters warnings

hoefling:
* single backticks don't produce verbatim text in ReST (reported by `rst-backticks` hook)
* cleaned up whitespaces reported by `doc8`

### DIFF
--- a/docs/source/class_basics.rst
+++ b/docs/source/class_basics.rst
@@ -78,8 +78,8 @@ to it explicitly using ``self``:
            a = self
            a.x = 1      # Error: 'x' not defined
 
-Annotating `__init__` methods
-*****************************
+Annotating ``__init__`` methods
+*******************************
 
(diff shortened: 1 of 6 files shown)


JelleZijlstra:
Is there something we could do in CI to make sure we don't reintroduce warnings like this later?

hoefling:
@JelleZijlstra For linting docs, we use the [`pre-commit`](https://pre-commit.com/) hooks, so the subtle errors like trailing lines/spaces or wrong backticks are caught by committing locally. Then, the CI runs `pre-commit run --all-files` to execute the same hooks once in the CI job. Since `mypy` repo doesn't use `pre-commit`, I suppose the `testenv:docs` job could be extended with a `doc8`/rst-backticks checks run?

gvanrossum:
I recall looking into the "pre-commit" tool and being baffled by its complexity.

Let's start by adding a script to lint docs and run it in from .travis.yml.

PR #7651, opened 2019-10-07
Fixes for syntax errors in code examples

hoefling:
Extracted all code samples from the `.. code-block::` directives and ran them against `tokenize`, `ast.parse` and `exec(compile(...))`. Some samples contain subtle typos that produced `SyntaxError`s, fixed them.


docs/source/generics.rst
```diff
@@ -457,8 +457,6 @@ Java, where the return type would be ``S``. The way mypy implements
 this is correct for ``concat``, since ``concat`` actually returns a
 ``str`` instance in the above example:
 
-.. code-block:: python
-
```
gvanrossum: Why'd you delete this code-block directive?

hoefling: This is a [doctest block](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#doctest-blocks) already because of its format (mimicking the Python interactive shell input/output), Sphinx knows to render them as `code-block`s. The wrapping `.. code-block::` directive thus has no effect. However, this has a bad side-effect that the text inside `.. code-block:: python` may be interpreted as pure Python code by other tools (e.g. the doctest runner I mentioned in #7624 will raise a `SyntaxError`).

gvanrossum: That honestly sounds a little too clever of Sphinx. And is that doctest runner an open source project or just a script you hacked together?

In the Python docs IIUC we use `::` to indicate non-Python code blocks, and there are a few places where that's also used in the mypy docs. I'd feel better if we used that here, or `code-block:: text`.

hoefling: It's the [Sybil codeblock runner](https://sybil.readthedocs.io/en/latest/parsers.html#codeblock) which is not that official, in the sense that it's not related to Sphinx in any way :-) I used that to extract and execute the code examples to check for more syntax issues. I didn't think the `code-block` directive is an intention here; will revert that.

gvanrossum: IMO it's still a code block, and explicit markup is better, so thanks for fixing. Presumably colorizing tools would do the right thing. :-)

PR #7652, opened 2019-10-07
Crossreferences to standard library in mypy docs

hoefling:

* `additional_features.rst`
* `casts.rst`
* `class_basics.rst`
* `command_line.rst`

### DIFF
--- a/docs/source/additional_features.rst
+++ b/docs/source/additional_features.rst
@@ -75,7 +75,7 @@ Caveats/Known Issues
 Some functions in the :py:mod:`dataclasses` module, such as :py:func:`~dataclasses.replace` and :py:func:`~dataclasses.asdict`,
 have imprecise (too permissive) types. This will be fixed in future releases.
 
-Mypy does not yet recognize aliases of :py:func:`dataclasses.dataclass`, and will
+Mypy does not yet recognize aliases of :py:func:`dataclasses.dataclass <dataclasses.dataclass>`, and will
 probably never recognize dynamically computed decorators. The following examples
 do **not** work:
 
--- a/docs/source/additional_features.rst
+++ b/docs/source/additional_features.rst
@@ -139,7 +139,8 @@ If you're using ``auto_attribs=True`` you must use variable annotations.
         two: int = 7
         three: int = attr.ib(8)
 
-Typeshed has a couple of "white lie" annotations to make type checking easier. :py:func:`attr.ib` and :py:class:`attr.Factory` actually return objects, but the
+Typeshed has a couple of "white lie" annotations to make type checking
+easier. :py:func:`attr.ib` and :py:class:`attr.Factory` actually return objects, but the
 annotation says these return the types that they expect to be assigned to.
 That enables this to work:
 
--- a/docs/source/additional_features.rst
+++ b/docs/source/additional_features.rst
@@ -26,7 +26,7 @@ They can be defined using the :py:func:`@dataclasses.dataclass
     test = Application("Testing...")  # OK
     bad = Application("Testing...", "with plugin")  # Error: List[str] expected
 
-Mypy will detect special methods (such as ``__lt__``) depending on the flags used to
+Mypy will detect special methods (such as :py:meth:`__lt__ <object.__lt__>`) depending on the flags used to
 define dataclasses. For example:
 
 .. code-block:: python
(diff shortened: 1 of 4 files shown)


docs/source/class_basics.rst
```diff
@@ -111,7 +111,7 @@ annotation, it is considered an untyped method:
 Class attribute annotations
 ***************************
 
-You can use a ``ClassVar[t]`` annotation to explicitly declare that a
```
hoefling: I'm unsure whether a reference is allowed or necessary here. Also, shouldn't the type be a capital `T`?

docs/source/class_basics.rst
```diff
@@ -232,10 +231,10 @@ effect at runtime:
 Abstract base classes and multiple inheritance
 **********************************************
 
-Mypy supports Python abstract base classes (ABCs). Abstract classes
+Mypy supports Python :doc:`abstract base classes <library/abc>` (ABCs). Abstract classes
```
hoefling: This is a new reference that was added by me, IMO it would be nice to guide the reader to the module document.

docs/source/command_line.rst
```diff
@@ -129,17 +129,17 @@ imports.
 ``--python-executable EXECUTABLE``
     This flag will have mypy collect type information from :pep:`561`
     compliant packages installed for the Python executable ``EXECUTABLE``.
-    If not provided, mypy will use :pep:`561` compliant packages installed for
```
hoefling: There is a link to PEP 561 already one line above, the other ones were too much blue in one paragraph...?

docs/source/command_line.rst
```diff
@@ -230,9 +230,9 @@ The following options are available:
 
 ``--disallow-any-generics``
     This flag disallows usage of generic types that do not specify explicit
-    type parameters. Moreover, built-in collections (such as ``list`` and
-    ``dict``) become disallowed as you should use their aliases from the typing
-    module (such as ``List[int]`` and ``Dict[str, str]``).
+    type parameters. Moreover, built-in collections (such as :py:class:`list` and
```
hoefling: Please advise whether this change is necessary at all.

docs/source/command_line.rst
```diff
@@ -292,7 +292,7 @@ For more details, see :ref:`no_strict_optional`.
 
 ``--no-implicit-optional``
     This flag causes mypy to stop treating arguments with a ``None``
-    default value as having an implicit ``Optional[...]`` type.
```
hoefling: Is the referencing necessary here? It may be confusing when combined with the ellipsis in type brackets.

docs/source/additional_features.rst
```diff
@@ -72,10 +72,10 @@ and :pep:`557`.
 Caveats/Known Issues
 ====================
 
-Some functions in the ``dataclasses`` module, such as ``replace()`` and ``asdict()``,
+Some functions in the :py:mod:`dataclasses` module, such as :py:func:`~dataclasses.replace` and :py:func:`~dataclasses.asdict`,
 have imprecise (too permissive) types. This will be fixed in future releases.
 
-Mypy does not yet recognize aliases of ``dataclasses.dataclass``, and will
+Mypy does not yet recognize aliases of :py:func:`dataclasses.dataclass`, and will
```
gvanrossum: This one causes is a regression in rendering IMO. The `:py:func:` version gets rendered with a trailing `()` pair, while the original gets rendered without it. Since an alias to a function does not involve calling it, IMO it should be rendered without `()`. (Maybe one of the other `:py:` things link the same way but render without `()`?)

docs/source/class_basics.rst
gvanrossum: I think it's useful. The `t` is intentionally lowercase, since it's not a type variable -- in this context (and occasionally elsewhere) it's just any type.

docs/source/class_basics.rst
gvanrossum: +1

docs/source/command_line.rst
gvanrossum: OK.

docs/source/command_line.rst
gvanrossum: I think it's fine.

docs/source/command_line.rst
gvanrossum: I think the link is useful here. For someone who doesn't know `Optional` but is looking up this flag (maybe they inherited a script to run mypy), the link may be a lifeline. But let's remove the `[...]` part.

docs/source/additional_features.rst
hoefling: Dang, missed that. Sure, the proper ref here would be
```
:py:func:`dataclasses.dataclass <dataclasses.dataclass>`
```
Will fix that!

hoefling:
The tilde removes the dotted path in the rendered text. E.g.
```
:py:class:`foo.bar.baz.Spam`
```
will render as `foo.bar.baz.Spam`, while
```
:py:class:`~foo.bar.baz.Spam`
```
will render as just `Spam`. However, the link will be identical in both cases.

PR #7655, opened 2019-10-08
Give error if __exit__ returns False but is declared to return bool

### LINKED ISSUE #7577: "Missing return statement" caused by context manager?
```
""" Unit tests. """

from types import TracebackType
from typing import Optional, Type


class TestManager:
    def __enter__(self) -> None:
        ...

    def __exit__(self, exc_type: Optional[Type[BaseException]], 
                 exc_val: Optional[BaseException],
                 exc_tb: Optional[TracebackType]) -> bool:
        ...


def func() -> int:
    with TestManager():
        return 1
```

Mypy will complain about `func()`, saying it doesn't have a return statement.

```
tests/__init__.py:16:1: error: Missing return statement
```

Is this intended?

_Note: I also tried `class TestManager(ContextManager[None]):` just to see if that would help mypy but it didn't change anything._

JukkaL:
Mypy can give false positives about missing return statements
if `__exit__` that always returns `False` is annotated to return
`bool` instead of `Literal[False]`.

Add new error code and documentation for the error code since
this error condition is not very obvious.

Fixes #7577.

There are two major limitations:

1. This doesn't support async context managers.
2. This won't help if a stub has an invalid `__exit__` return type.

I'll create a follow-up issues about the above.

### DIFF
--- a/docs/source/error_code_list.rst
+++ b/docs/source/error_code_list.rst
@@ -593,6 +593,63 @@ To work around the issue, you can either give mypy access to the sources
 for ``acme`` or create a stub file for the module.  See :ref:`ignore-missing-imports`
 for more information.
 
+Check the return type of __exit__ [exit-return]
+-----------------------------------------------
+
+If mypy can determine that ``__exit__`` always returns ``False``, mypy
+checks that the return type is *not* ``bool``.  The boolean value of
+the return type affects which lines mypy thinks are reachable after a
+``with`` statement, since any ``__exit__`` method that can return
+``True`` may swallow exceptions. An imprecise return type can result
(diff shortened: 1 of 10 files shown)


Michael0x2a:
Maybe we could also suggest returning `Optional[bool]`? I believe that'll also make mypy assume the function will not swallow exceptions.

JukkaL:
> Maybe we could also suggest returning Optional[bool]? I believe that'll also make mypy assume the function will not swallow exceptions.

That doesn't sound correct, though. If the return type is `Optional[bool]`, the function could still return `True` and swallow exceptions.

PR #7656, opened 2019-10-08
Only return a declared type from __new__ if it is a subtype of the class

### LINKED ISSUE #7597: Unexpected error when inheriting a class from Popen and overriding some method in it
Bug

```python
import subprocess
from typing import List, Tuple


class MyPopen(subprocess.Popen):

    def communicate(self, _input=None, _timeout=None) -> Tuple[bytes, bytes]:
        return super().communicate(_input, _timeout)

def interactive_spawn(cmd: List[str], **kwargs) -> MyPopen:
    process = MyPopen(cmd, **kwargs)
    process.communicate()
    return process
```

```
$ mypy mypy_subprocess.py
mypy_subprocess.py:13: error: Incompatible return value type (got "Popen[Any]", expected "MyPopen")
````

```
$ mypy mypy_subprocess.py
```

```
$ mypy --version
mypy 0.730
```

```
(env) $ mypy --version
mypy 0.740+dev.583b99e5096c979043a63dcf92bf3f0d5993d03a
```

**NOTE: it's not reproducing with `0.720`**

```
$ python --version
Python 3.7.4
```

-

  the full traceback below.
-

### PR DESCRIPTION
**Pull request by msullivan**
Fixes #7597.

### DIFF
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -6031,7 +6031,7 @@ class A:
     def __new__(cls) -> int:  # E: Incompatible return type for "__new__" (returns "int", but must return a subtype of "A")
         pass
 
-reveal_type(A())  # N: Revealed type is 'builtins.int'
+reveal_type(A())  # N: Revealed type is '__main__.A'
 
 [case testNewReturnType4]
 from typing import TypeVar, Type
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test-data/unit/check-classes.test [hunk 1]
```diff
@@ -6031,7 +6031,7 @@ class A:
     def __new__(cls) -> int:  # E: Incompatible return type for "__new__" (returns "int", but must return a subtype of "A")
         pass
 
-reveal_type(A())  # N: Revealed type is 'builtins.int'
+reveal_type(A())  # N: Revealed type is '__main__.A'
```
ilevkivskyi: I think we should keep the declared return type for the class itself. The original PR closed an issue called "Honor return type of `__new__`", so this change would be a step back.

msullivan: We report an error at the `__new__` declaration site, and there are a bunch of other situations where we wouldn't honor it before (like if it is a union or something other than an instance or tuple), so I think this actually makes thing a little more consistent.

actionless:
thanks! it worked however i found another problem with popen inheritance, i am posting an issue right now

PR #7658, opened 2019-10-08
attrs: note variable annotation preference over .ib type= arguments

scop:
http://www.attrs.org/en/stable/api.html#attr.ib

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -505,6 +505,8 @@ def _attribute_from_attrib_maker(ctx: 'mypy.plugin.ClassDefContext',
     # If the type isn't set through annotation but is passed through `type=` use that.
     type_arg = _get_argument(rvalue, 'type')
     if type_arg and not init_type:
+        if ctx.api.options.python_version >= (3, 6):
+            ctx.api.note('Variable annotation syntax is preferred over type= arguments', rvalue)
         try:
             un_type = expr_to_unanalyzed_type(type_arg)
         except TypeTranslationError:
--- a/test-data/unit/check-attr.test
+++ b/test-data/unit/check-attr.test
@@ -102,7 +102,7 @@ class C:
         return self.x  # E: Incompatible return value type (got "int", expected "str")
 @attr.s
 class D:
-    x = attr.ib(8, type=int)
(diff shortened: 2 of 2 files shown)


brandtbucher:
Thanks for your time @scop, and welcome to mypy!

scop:
There are still some errors remaining, both just with the expected test output (dunno at least yet how to express two messages from a single line), as well as `note` not being part of `SemanticAnalyzerInterface`.

Before doing more work to fix these, it would be good to hear that the added message would be an acceptable and welcome addition to mypy in the first place :)

Also, instructions/opinions welcome whether `note` is the right thing to use for this message or if something else should be used, and if it is, would it be ok to add it to `SemanticAnalyzerInterface`.

mypy/plugins/attrs.py
```diff
@@ -505,6 +505,8 @@ def _attribute_from_attrib_maker(ctx: 'mypy.plugin.ClassDefContext',
     # If the type isn't set through annotation but is passed through `type=` use that.
     type_arg = _get_argument(rvalue, 'type')
     if type_arg and not init_type:
+        if ctx.api.options.python_version >= (3, 6):
+            ctx.api.note('Variable annotation syntax is preferred over type= arguments', rvalue)
```
JukkaL: I think that this too noisy to be enabled by default. Some users might want to type check their codebase using multiple Python versions, and this could result in many notes when using 3.6 or later as the target Python version. I think that this would work better as an optional check in the future, once we support enabling and disabling individual checks (I can't give a timeline for this feature though).

scop: I tend to agree, however I would still appreciate this included in nearish future. How about if the note would be emitted just once for the first occurrence in a run, and message changed to note the Python version, something like:

> Variable annotation syntax is preferred over type= arguments with Python 3.6 and later

Do you think that would be acceptable already now?

JelleZijlstra: Any `note` makes mypy exit with a 1 exit code, so I don't think it's realistically possible to have this note at all.

97littleleaf11:
I am going to close this due to inactive status. Feel free to reopen it if you have further progress!

PR #7660, opened 2019-10-08
Crossreferences to standard library in mypy docs, part 2

hoefling:

* `builtin_types.rst`
* `common_issues.rst`
* `config_file.rst`
* `dynamic_typing.rst`

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -69,7 +69,7 @@ flagged as an error.
   :ref:`reveal_type() <reveal-type>` might come in handy.
 
   Note that sometimes library stubs have imprecise type information,
-  e.g. the ``pow()`` builtin returns ``Any`` (see `typeshed issue 285
+  e.g. the :py:func:`pow` builtin returns ``Any`` (see `typeshed issue 285
   <https://github.com/python/typeshed/issues/285>`_ for the reason).
 
 - **Some imports may be silently ignored**.  Another source of
@@ -143,7 +143,7 @@ Another option is to explicitly annotate values with type ``Any`` --
 mypy will let you perform arbitrary operations on ``Any``
 values. Sometimes there is no more precise type you can use for a
 particular value, especially if you use dynamic Python features
-such as ``__getattr__``:
+such as :py:meth:`__getattr__ <object.__getattr__>`:
 
 .. code-block:: python
 
@@ -326,7 +326,7 @@ above example:
 Complex type tests
 ------------------
 
-Mypy can usually infer the types correctly when using ``isinstance()``
+Mypy can usually infer the types correctly when using :py:func:`isinstance <isinstance>`
 type tests, but for other kinds of checks you may need to add an
 explicit type cast:
 
@@ -342,17 +342,17 @@ explicit type cast:
 
 .. note::
 
-    Note that the ``object`` type used in the above example is similar
+    Note that the :py:class:`object` type used in the above example is similar
     to ``Object`` in Java: it only supports operations defined for *all*
-    objects, such as equality and ``isinstance()``. The type ``Any``,
+    objects, such as equality and :py:func:`isinstance`. The type ``Any``,
     in contrast, supports all operations, even if they may fail at
     runtime. The cast above would have been unnecessary if the type of
     ``o`` was ``Any``.
 
-Mypy can't infer the type of ``o`` after the ``type()`` check
-because it only knows about ``isinstance()`` (and the latter is better
+Mypy can't infer the type of ``o`` after the :py:class:`type() <type>` check
+because it only knows about :py:func:`isinstance` (and the latter is better
 style anyway).  We can write the above code without a cast by using
-``isinstance()``:
+:py:func:`isinstance`:
 
 .. code-block:: python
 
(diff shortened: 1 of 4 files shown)


docs/source/common_issues.rst
```diff
@@ -417,14 +417,14 @@ Example:
    # The rest of this file doesn't apply to Windows.
 
 Some other expressions exhibit similar behavior; in particular,
-``typing.TYPE_CHECKING``, variables named ``MYPY``, and any variable
+:py:data:`~typing.TYPE_CHECKING`, variables named ``MYPY``, and any variable
```
hoefling: The document mentions the `TYPE_CHECKING` constant sometimes as `TYPE_CHECKING` and sometimes as `typing.TYPE_CHECKING`. The occurrence of `typing.XXX` throughout the docs is marginal and the link leads to the `typing` module docs anyway, so I decided to drop the `typing.` prefix.

docs/source/config_file.rst
```diff
@@ -254,8 +254,8 @@ For more information, see the :ref:`None and optional handling <none-and-optiona
 section of the command line docs.
 
 ``no_implicit_optional`` (bool, default False)
-    Changes the treatment of arguments with a default value of None by not implicitly
-    making their type Optional.
+    Changes the treatment of arguments with a default value of ``None`` by not implicitly
```
hoefling: I suppose the verbatim formatting was forgotten here, so added the `Optional` ref as well.

docs/source/config_file.rst
```diff
@@ -377,8 +377,9 @@ a list of import discovery options that may be used
     User home directory and environment variables will be expanded.
 
 ``files`` (string)
+
     A comma-separated list of paths which should be checked by mypy if none are given on the command
-    line. Supports recursive file globbing using :doc:`library/glob`, where ``*`` (e.g. ``*.py``) matches
+    line. Supports recursive file globbing using :py:mod:`glob`, where ``*`` (e.g. ``*.py``) matches
```
hoefling: Please advise here. Both links lead to the same target: the [`glob`](https://docs.python.org/3/library/glob.html) module docs, but are rendered differently:

image

vs

image

docs/source/config_file.rst
```diff
@@ -418,7 +419,7 @@ section of the command line docs.
     variable.
 
     Note that the cache is only read when incremental mode is enabled
-    but is always written to, unless the value is set to ``/dev/nul``
+    but is always written to, unless the value is set to ``/dev/null``
```
hoefling: This isn't related to the crossreferencing (and it's the only one unrelated change I included in this PR), but is a simple typo, so I decided to include it.

docs/source/dynamic_typing.rst
```diff
@@ -80,7 +80,7 @@ operations:
         n = 1  # type: int
         n = o         # Error!
 
-You can use ``cast()`` (see chapter :ref:`casts`) or ``isinstance`` to
-go from a general type such as ``object`` to a more specific
-type (subtype) such as ``int``.  ``cast()`` is not needed with
+You can use :py:func:`~typing.cast` (see chapter :ref:`casts`) or :py:func:`isinstance` to
```
hoefling: I have altered the rendering by including the parentheses to `isinstance`. Before:

image

After:

image

Both are functions - shouldn't both have parentheses in this case?

docs/source/common_issues.rst
gvanrossum: +1

docs/source/config_file.rst
gvanrossum: +1

docs/source/config_file.rst
gvanrossum: I like the second better. So +1 to the change. (Do we use this `:doc:` style elsewhere, and does it get rendered so longwindedly there too? I think it's not needed.)

docs/source/config_file.rst
gvanrossum: +1

docs/source/dynamic_typing.rst
gvanrossum: +1. The only time it's better not to add `()` to a function is when we're talking about the function object (e.g. in the context of creating an alias or accessing attributes).

docs/source/config_file.rst
hoefling: There are a few `:doc:` roles that refer to standard library modules: 

https://github.com/python/mypy/blob/c95ecc1952b4f804662e37b36c0779b32226f88b/docs/source/additional_features.rst#L69

https://github.com/python/mypy/blob/c95ecc1952b4f804662e37b36c0779b32226f88b/docs/source/class_basics.rst#L235

https://github.com/python/mypy/blob/c95ecc1952b4f804662e37b36c0779b32226f88b/docs/source/config_file.rst#L32

https://github.com/python/mypy/blob/c95ecc1952b4f804662e37b36c0779b32226f88b/docs/source/error_code_list.rst#L545-L546

Do you suggest to replace them with the appropriate `:mod:` role?

gvanrossum: It doesn't seem necessary, since those other places all specify link text.

We cannot hope to strive for 100% consistency everywhere.

PR #7662, opened 2019-10-09
Produce error when NamedTuple is assigned to attribute

### LINKED ISSUE #7531: NamedTuple triggered mypy error inside a class
```python
from typing import NamedTuple


class A:
    def __init__(self) -> None:
        self.tuple2 = NamedTuple('x', [('y', str), ('z', bool)])

    def run2(self) -> NamedTuple:
        return self.tuple2(y='y', z=True)


print(A().run2())
```
The script will trigger `"NamedTuple" not callable  mypy(error)`, not sure if this is as expected

Reference:
https://mypy.readthedocs.io/en/latest/kinds_of_types.html#named-tuples

---

Tested with Python37 and mypy==0.701

AFanaei:
Fixes #7531

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -2120,8 +2120,7 @@ class SemanticAnalyzer(NodeVisitor[None],
         """Check if s defines a namedtuple."""
         if isinstance(s.rvalue, CallExpr) and isinstance(s.rvalue.analyzed, NamedTupleExpr):
             return True  # This is a valid and analyzed named tuple definition, nothing to do here.
-        if len(s.lvalues) != 1 or not (isinstance(s.lvalues[0], NameExpr) or
-                                       isinstance(s.lvalues[0], MemberExpr)):
+        if len(s.lvalues) != 1 or not isinstance(s.lvalues[0], (NameExpr, MemberExpr)):
             return False
         lvalue = s.lvalues[0]
         name = lvalue.name
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -2129,7 +2129,7 @@ class SemanticAnalyzer(NodeVisitor[None],
         if not is_named_tuple:
             return False
         if isinstance(s.lvalues[0], MemberExpr):
-            self.fail("NamedTuple type as an attribute is not supported", s.lvalues[0])
+            self.fail("NamedTuple type as an attribute is not supported", lvalue)
             return False
         # Yes, it's a valid namedtuple, but defer if it is not ready.
         if not info:
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -2120,7 +2120,8 @@ class SemanticAnalyzer(NodeVisitor[None],
         """Check if s defines a namedtuple."""
         if isinstance(s.rvalue, CallExpr) and isinstance(s.rvalue.analyzed, NamedTupleExpr):
             return True  # This is a valid and analyzed named tuple definition, nothing to do here.
-        if len(s.lvalues) != 1 or type(s.lvalues[0]) not in [NameExpr, MemberExpr]:
+        if len(s.lvalues) != 1 or not (isinstance(s.lvalues[0], NameExpr) or
+                                       isinstance(s.lvalues[0], MemberExpr)):
             return False
         lvalue = s.lvalues[0]
         name = lvalue.name
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -2120,13 +2120,16 @@ class SemanticAnalyzer(NodeVisitor[None],
         """Check if s defines a namedtuple."""
         if isinstance(s.rvalue, CallExpr) and isinstance(s.rvalue.analyzed, NamedTupleExpr):
             return True  # This is a valid and analyzed named tuple definition, nothing to do here.
-        if len(s.lvalues) != 1 or not isinstance(s.lvalues[0], NameExpr):
+        if len(s.lvalues) != 1 or type(s.lvalues[0]) not in [NameExpr, MemberExpr]:
             return False
         lvalue = s.lvalues[0]
         name = lvalue.name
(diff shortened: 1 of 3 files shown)


mypy/semanal.py
```diff
@@ -2120,14 +2120,17 @@ def analyze_namedtuple_assign(self, s: AssignmentStmt) -> bool:
         """Check if s defines a namedtuple."""
         if isinstance(s.rvalue, CallExpr) and isinstance(s.rvalue.analyzed, NamedTupleExpr):
             return True  # This is a valid and analyzed named tuple definition, nothing to do here.
-        if len(s.lvalues) != 1 or not isinstance(s.lvalues[0], NameExpr):
+        if len(s.lvalues) != 1 or type(s.lvalues[0]) not in [NameExpr, MemberExpr]:
             return False
         lvalue = s.lvalues[0]
         name = lvalue.name
```
AFanaei: here a CI error is raised `"error: Expression" has no attribute "name" `,   
its confusing because lvalue is either NameExpr or MemberExpr both of which have `name` attribute.

mypy/semanal.py
```diff
@@ -2120,14 +2120,17 @@ def analyze_namedtuple_assign(self, s: AssignmentStmt) -> bool:
...
         is_named_tuple, info = self.named_tuple_analyzer.check_namedtuple(s.rvalue, name,
                                                                           self.is_func_scope())
         if not is_named_tuple:
             return False
+        if isinstance(s.lvalues[0], MemberExpr):
+            self.fail("NamedTuple type as an attribute is not supported", s.lvalues[0])
+            return False
```
AFanaei: This hole block could be moved at the start of function before `check_namedtuple` is called.   
but then strangely a `error: "int" not callable` error happens during test-case.   
i believe the error message is totally miss-leading.

mypy/semanal.py
JelleZijlstra: Mypy's type inference isn't smart enough to figure that out. You should probably move the `isinstance()` check to operate on the `lvalue` variable.

AFanaei: I managed to get the CI green but the code isn't the best.

AFanaei:
hi @JukkaL , could you review this PR?

mypy/semanal.py
```diff
@@ -2120,14 +2120,18 @@ def analyze_namedtuple_assign(self, s: AssignmentStmt) -> bool:
         """Check if s defines a namedtuple."""
         if isinstance(s.rvalue, CallExpr) and isinstance(s.rvalue.analyzed, NamedTupleExpr):
             return True  # This is a valid and analyzed named tuple definition, nothing to do here.
-        if len(s.lvalues) != 1 or not isinstance(s.lvalues[0], NameExpr):
+        if len(s.lvalues) != 1 or not (isinstance(s.lvalues[0], NameExpr) or
+                                       isinstance(s.lvalues[0], MemberExpr)):
```
JukkaL: Style nit: The latter part could be written a bit more compactly as `isinstance(s.values[0], (NameExpr, MemberExpr))`.

mypy/semanal.py
JukkaL: The `"int" not callable` error is due to how test stubs cheat a bit and use too simple definitions for some things such as `NamedTuple`. I'd recommend moving the new test case to `test-data/unit/pythoneval.test` which uses the full typeshed stubs. This should allow you to move this to the start of the function.

test-data/unit/check-namedtuple.test
```diff
@@ -853,3 +853,10 @@ b = (1, 2)
 if not b:
     ''()  # E: "str" not callable
 [builtins fixtures/tuple.pyi]
+
+[case testNamedTupleAtRunTime]
+from typing import NamedTuple
+
+class A:
+    def __init__(self) -> None:
+        self.b = NamedTuple('x', [('s', str), ('n', int)])  # E: NamedTuple type as an attribute is not supported
```
JukkaL: As I mentioned above, it would make sense to move this to `pythoneval.test`, since the behavior may depend on how `NamedTuple` is defined in stubs, and typeshed has a different definition than the simplified stubs used in most tests.

Also, add something like `reveal_type(A().b)  # N: ...` to the end of the test case.

mypy/semanal.py
AFanaei: I moved the block to the start, but `NamedTuple type as a ...` error was generated on every line with an attribute assignment statment.  
i think this error should be created only if RHS is a NamedTuple so the if is better at the end.

test-data/unit/pythoneval.test
```diff
@@ -1453,3 +1453,15 @@ def f_suppresses() -> int:
...
+from typing import NamedTuple
+
+class A:
+    def __init__(self) -> None:
+        self.b = NamedTuple('x', [('s', str), ('n', int)])
+
+reveal_type(A().b)
+[out]
+_testNamedTupleAtRunTime.py:5: error: NamedTuple type as an attribute is not supported
+_testNamedTupleAtRunTime.py:7: note: Revealed type is 'Any'
```
AFanaei: Shouldn't it be `NamedTuple` instead of `Any`

mypy/semanal.py
```diff
@@ -2120,14 +2120,17 @@ def analyze_namedtuple_assign(self, s: AssignmentStmt) -> bool:
...
             return False
         lvalue = s.lvalues[0]
         name = lvalue.name
         is_named_tuple, info = self.named_tuple_analyzer.check_namedtuple(s.rvalue, name,
                                                                           self.is_func_scope())
         if not is_named_tuple:
             return False
+        if isinstance(s.lvalues[0], MemberExpr):
+            self.fail("NamedTuple type as an attribute is not supported", lvalue)
+            return False
```
AFanaei: I moved this block to start but then a `NamedTuple type as an attrib ...`  error was generated on every attribute assignment statement.   
as the error should raise only if RHS is NamedTuple, i think this block is better at the end.

PR #7675, opened 2019-10-09
[attrs] Remove error message about deprecated cmp

euresti:
It still works it just emits a warning so there's no need for mypy to freak out about it.

### DIFF
--- a/test-data/unit/check-attr.test
+++ b/test-data/unit/check-attr.test
@@ -278,15 +278,15 @@ A(1) != 1
 
 [case testAttrsCmpEqOrderValues]
 from attr import attrib, attrs
-@attrs(cmp=True)  # E: cmp is deprecated, use eq and order
+@attrs(cmp=True)
 class DeprecatedTrue:
    ...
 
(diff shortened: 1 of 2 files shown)


test-data/unit/check-attr.test
```diff
@@ -278,15 +278,15 @@ A(1) != 1
...
 class DeprecatedTrue:
    ...
 
-@attrs(cmp=False)  # E: cmp is deprecated, use eq and order
+@attrs(cmp=False)
 class DeprecatedFalse:
    ...
 
-@attrs(cmp=False, eq=True)  # E: Don't mix `cmp` with `eq' and `order`  # E: cmp is deprecated, use eq and order
+@attrs(cmp=False, eq=True)  # E: Don't mix `cmp` with `eq' and `order`
```
ilevkivskyi: I just noticed the old-school quotes here :-) I mean `` `...' `` is great, but we are now moving towards unifying all error messages on using plain boring double quotes. So it would be great to update this (possibly in a separate PR)

PR #7676, opened 2019-10-09
[mypyc] Make namegen not need to use suffixes to disambiguate

msullivan:
Accomplish this by using a mapping that doesn't collide on valid
python identifiers. We map `.` to `___` and the (rare and unlikely,
but possible) `___` to `_3___`. (This collides `___` with `.3`,
which is fine.)

This makes naming not reliant on the order of processing things
which is important for separate compilation, since it will turn out
that most of our "internal" names actually do need to be exported
(but we would still like to shorten the module name part).

I also like the output better, even though it is a little more
verbose, since it makes it obvious what is a module separator and what
is an underscore.

### DIFF
--- a/mypyc/ops.py
+++ b/mypyc/ops.py
@@ -1416,11 +1416,12 @@ class FuncDecl:
             else:
                 self.bound_sig = FuncSignature(sig.args[1:], sig.ret_type)
 
+    @property
+    def shortname(self) -> str:
+        return self.class_name + '.' + self.name if self.class_name else self.name
+
     def cname(self, names: NameGenerator) -> str:
-        name = self.name
-        if self.class_name:
-            name += '_' + self.class_name
-        return names.private_name(self.module_name, name)
+        return names.private_name(self.module_name, self.shortname)
 
 
 class FuncIR:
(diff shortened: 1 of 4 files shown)


mypyc/test/test_namegen.py
```diff
@@ -7,33 +7,34 @@
 
 class TestNameGen(unittest.TestCase):
     def test_candidate_suffixes(self) -> None:
-        assert candidate_suffixes('foo') == ['', 'foo_']
-        assert candidate_suffixes('foo.bar') == ['', 'bar_', 'foo_bar_']
+        assert candidate_suffixes('foo') == ['', 'foo.']
```
JukkaL: Would it be useful to have an end-to-end test case that has things like triple underscores in names and possible name clashes, such as having names `C.foo` and `C___foo` in the same namespace?

mypyc/ops.py
```diff
@@ -1416,11 +1416,12 @@ def __init__(self,
             else:
                 self.bound_sig = FuncSignature(sig.args[1:], sig.ret_type)
 
+    @property
+    def shortname(self) -> str:
+        return self.class_name + '.' + self.name if self.class_name else self.name
```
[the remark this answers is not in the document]
msullivan: Nothing about mypyc works with nested classes

mypyc/test/test_namegen.py
msullivan: I'll add that to the name clash issue test

PR #7677, opened 2019-10-09
Crossreferences to standard library in mypy docs, part 3

hoefling:

* `error_code_list.rst`
* `error_code_list2.rst`
* `extending_mypy.rst`
* `faq.rst`
* `final_attrs.rst`
* `getting_started.rst`

### DIFF
--- a/docs/source/error_code_list.rst
+++ b/docs/source/error_code_list.rst
@@ -83,7 +83,7 @@ definition in an active scope, such as an assignment, function
 definition or an import. This can catch missing definitions, missing
 imports, and typos.
 
-This example accidentally calls ``sort()`` instead of ``sorted()``:
+This example accidentally calls ``sort()`` instead of :py:func:`sorted`:
 
 .. code-block:: python
 
@@ -181,7 +181,7 @@ This example incorrectly uses the function ``log`` as a type:
        for x in objs:
            f(x)
 
-You can use ``Callable`` as the type for callable objects:
+You can use :py:data:`~typing.Callable` as the type for callable objects:
 
 .. code-block:: python
 
@@ -415,11 +415,11 @@ Example:
     # Error: Dict entry 0 has incompatible type "str": "str"; expected "str": "int"  [dict-item]
     d: Dict[str, int] = {'key': 'value'}
 
-Check TypedDict items [typeddict-item]
---------------------------------------
+Check ``TypedDict`` items [typeddict-item]
+------------------------------------------
 
-When constructing a TypedDict object, mypy checks that each key and value is compatible
-with the TypedDict type that is inferred from the surrounding context.
+When constructing a ``TypedDict`` object, mypy checks that each key and value is compatible
+with the ``TypedDict`` type that is inferred from the surrounding context.
 
 Example:
 
@@ -542,8 +542,7 @@ Check instantiation of abstract classes [abstract]
 
 Mypy generates an error if you try to instantiate an abstract base
 class (ABC). An abtract base class is a class with at least one
-abstract method or attribute. (See also :doc:`Python
-abc module documentation <python:library/abc>`)
+abstract method or attribute. (See also :py:mod:`abc` module documentation)
 
 Sometimes a class is made accidentally abstract, often due to an
 unimplemented abstract method. In a case like this you need to provide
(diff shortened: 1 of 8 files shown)


docs/source/error_code_list.rst
```diff
@@ -415,11 +415,11 @@ Example:
     # Error: Dict entry 0 has incompatible type "str": "str"; expected "str": "int"  [dict-item]
     d: Dict[str, int] = {'key': 'value'}
 
-Check TypedDict items [typeddict-item]
---------------------------------------
+Check ``TypedDict`` items [typeddict-item]
```
hoefling: Please advise whether verbatim texts in section headers are useful. This renders as:

image

The side effect is also verbatim text in the table of contents:

image

docs/source/extending_mypy.rst
```diff
@@ -185,12 +185,12 @@ For example in this code:
    x[0] = 42
 
 mypy will call ``get_method_signature_hook("ctypes.Array.__setitem__")``
-so that the plugin can mimic the ``ctypes`` auto-convert behavior.
+so that the plugin can mimic the :py:mod:`ctypes` auto-convert behavior.
 
-**get_attribute_hook** overrides instance member field lookups and property
+**get_attribute_hook()** overrides instance member field lookups and property
```
hoefling: All other paragraphs use parentheses in hook names, I suppose they were omitted by mistake here.

docs/source/getting_started.rst
```diff
@@ -160,16 +160,16 @@ Arguments with default values can be annotated like so:
...
+*********************
 
 So far, we've added type hints that use only basic concrete types like
 ``str`` and ``float``. What if we want to express more complex types,
 such as "a list of strings" or "an iterable of ints"?
 
-You can find many of these more complex static types inside of the ``typing``
+You can find many of these more complex static types inside of the :py:mod:`typing`
 module. For example, to indicate that some function can accept a list of
-strings, use the ``List`` type from the ``typing`` module:
```
hoefling: Second change unrelated to referencing. IMO with the `List` reference, there's no need to mention the module again - it's also referenced one line above that. What do you think?

docs/source/faq.rst
```diff
@@ -164,7 +163,7 @@ monkey patching of methods.
 How is mypy different from Cython?
 **********************************
 
-`Cython :doc:<cython:index>` is a variant of Python that supports
+:doc:`Cython <cython:index>` is a variant of Python that supports
```
hoefling: Fixing my own errors here. No idea why Sphinx didn't catch this on build.

docs/source/extending_mypy.rst
```diff
@@ -98,7 +98,7 @@ High-level overview
 *******************
 
 Every entry point function should accept a single string argument
-that is a full mypy version and return a subclass of ``mypy.plugins.Plugin``:
```
hoefling: One of two changes in this PR that are not related to referencing, typo fix.

docs/source/error_code_list.rst
gvanrossum: I think it looks better without the heavy formatting. TBH I find most CSS styling for code blocks over the top, and red is particularly bad in a heading. I also don't care about the box (or shading in other CSS setups).

docs/source/extending_mypy.rst
gvanrossum: +1

docs/source/getting_started.rst
gvanrossum: +1

(PS. When commenting on a diff chunk, please attach the comment to the last line, so the context in email or in the Conversation view is clearer. I think you can now also shift-click to attach a comment to a range of lines.)

docs/source/faq.rst
gvanrossum: Didn't you have to run it with special flags to check errors?

docs/source/error_code_list.rst
hoefling: Agree. IMO the best solution is simply use monospaced font for verbatim formatting like Python stdlib docs do that (and afaik Javadoc too). If you don't mind, I will also replace other occurences of verbatim text in section titles in this PR.

docs/source/getting_started.rst
hoefling: Got it!

docs/source/faq.rst
hoefling: I did (using `sphinx-build -WEan`), however Sphinx ignores markup errors and just renders the text as-is. Neither does `doc8` nor `restructuredtext-lint`, sadly.

PR #7683, opened 2019-10-10
Add __qualname__ to implicit attrs

### LINKED ISSUE #6473: mypy does not recognize __qualname__

forrestchang:
Fix #6473.

Todo:
- [ ] Add tests.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -75,7 +75,7 @@ from mypy.nodes import (
     nongen_builtins, get_member_expr_fullname, REVEAL_TYPE,
     REVEAL_LOCALS, is_final_node, TypedDictExpr, type_aliases_target_versions,
     EnumCallExpr, RUNTIME_PROTOCOL_DECOS, FakeExpression, Statement, AssignmentExpr,
-)
+    implicit_class_attrs)
 from mypy.tvar_scope import TypeVarScope
 from mypy.typevars import fill_typevars
 from mypy.visitor import NodeVisitor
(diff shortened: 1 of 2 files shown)


mypy/semanal.py
```diff
@@ -393,9 +393,10 @@ def refresh_top_level(self, file_node: MypyFile) -> None:
         self.export_map[self.cur_mod_id] = self.all_exports
         self.all_exports = []
 
-    def add_implicit_module_attrs(self, file_node: MypyFile) -> None:
-        """Manually add implicit definitions of module '__name__' etc."""
-        for name, t in implicit_module_attrs.items():
+    def add_implicit_attrs(self, file_node: MypyFile) -> None:
+        """Manually add implicit definitions of module and class '__name__' etc."""
+        implicit_attrs = {**implicit_module_attrs, **implicit_class_attrs}
+        for name, t in implicit_attrs.items():
```
JukkaL: Note that `__qualname__` is not defined at module scope. It should only be added in class bodies. Also, it's only defined in Python 3.

forrestchang: Ok, I will work on that.

msullivan:
I'm going to close this as part of a PR tidy up. Feel free to reopen if you pick it back up.

PR #7694, opened 2019-10-11
Update references to new/old semantic analyzers

JukkaL:
Since we only have a single semantic analyzer now, generally only talk
about "the semantic analyzer" instead of "the new semantic analyzer"
in documentation and docstrings.

Plugin documentation still explains the differences between the old
and the new semantic analyzers, since some plugins might not support
the new semantic analyzer yet.

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -104,10 +104,9 @@ analyzer is no longer available):
    another module, such as the builtins, that is outside the current module or
    import cycle, you can safely assume that you won't receive a placeholder.
 
-When testing your plugin with the semantic analyzer, you should have a test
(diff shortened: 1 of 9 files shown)


mypy/plugin.py
```diff
@@ -104,7 +104,7 @@
    another module, such as the builtins, that is outside the current module or
    import cycle, you can safely assume that you won't receive a placeholder.
 
-When testing your plugin with the new semantic analyzer, you should have a test
+When testing your plugin with the semantic analyzer, you should have a test
```
ilevkivskyi: TBH this sound a bit awkward. I would just say it is recommended to have a test like below.

```suggestion
When testing your plugin, you should have a test
```

Issue #4542, opened 2018-02-05
Import handling is confusing and misleading


nedbat:
I tried mypy and was alarmed to see it unable to find imports that are clearly available.  Reading the explanations in #1293 and #1339, I now understand why it doesn't merrily parse third-party code that will only add to a growing pile of unannotated code.

But this is a really bad first experience.  I install something that claims to understand Python code, and it's confused by imports? Then it gives me two suggestions, neither of which is seems like the right advice: 
```
note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
```
I shouldn't try to set MYPYPATH, that's silly busy-work that if successful will only lead me into the unannotated third-party module trap.  The imports aren't missing, so why would `--ignore-missing-imports` help?  NOTE: using `--ignore-missing-imports` silences the warnings, but the option is misnamed: the imports are not missing.

Reading the help, I see `--follow-imports`, so I try those options. I try `--follow-imports=silent`, but that still complains about the imports. I try `--follow-imports=skip`, and that also still complains about the imports.  Both of these sound like exactly what I need, and neither changes the behavior much:

```
$ mypy --version
mypy 0.560
$ mypy census.py
census.py:16: error: Cannot find module named 'aiohttp'
census.py:16: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
census.py:17: error: Cannot find module named 'async_timeout'
census.py:18: error: Cannot find module named 'attr'
census.py:20: error: Cannot find module named 'opaque_keys'
census.py:21: error: Cannot find module named 'opaque_keys.edx'
census.py:21: error: Cannot find module named 'opaque_keys.edx.keys'
helpers.py:3: error: No library stub file for module 'lxml'
helpers.py:3: note: (Stub files are from https://github.com/python/typeshed)
helpers.py:4: error: No library stub file for module 'lxml.html'
$ mypy --follow-imports=silent census.py
census.py:16: error: Cannot find module named 'aiohttp'
census.py:16: note: (Perhaps setting MYPYPATH or using the "--ignore-missing-imports" flag would help)
census.py:17: error: Cannot find module named 'async_timeout'
census.py:18: error: Cannot find module named 'attr'
census.py:20: error: Cannot find module named 'opaque_keys'
census.py:21: error: Cannot find module named 'opaque_keys.edx'
census.py:21: error: Cannot find module named 'opaque_keys.edx.keys'
$ mypy --follow-imports=skip census.py
[...]

PR #7698, opened 2019-10-11
Improve missing import message

Lewiscowles1986:
It can be difficult to understand the missing module error message at times.

Changes 

> Cannot find module named '{}'

to

> Cannot find implementation or library stub for module named '{}'

This improves https://github.com/python/mypy/issues/4542

Great thanks to all that participated in that discussion, this is all of our work.

@JukkaL


JukkaL:
Looks good!

PR #7699, opened 2019-10-11
Crossreferences to standard library in mypy docs, part 7

hoefling:
Added references to the `protocols.rst`. This is the last part of splitting up the changes in #7624 into more readable PRs.

### DIFF
--- a/docs/source/protocols.rst
+++ b/docs/source/protocols.rst
@@ -10,7 +10,7 @@ inherits class ``C``, it's also a subtype of ``C``, and instances of
 ``D`` can be used when ``C`` instances are expected. This form of
 subtyping is used by default in mypy, since it's easy to understand
 and produces clear and concise error messages, and since it matches
-how the native ``isinstance()`` check works -- based on class
+how the native :py:func:`isinstance <isinstance>` check works -- based on class
 hierarchy. *Structural* subtyping can also be useful. Class ``D`` is
 a structural subtype of class ``C`` if the former has all attributes
 and methods of the latter, and with compatible types.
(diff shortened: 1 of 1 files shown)


docs/source/protocols.rst
```diff
@@ -66,18 +66,18 @@ Iteration protocols
 The iteration protocols are useful in many contexts. For example, they allow
 iteration of objects in for loops.
 
-``Iterable[T]``
----------------
+:py:class:`Iterable[T] <typing.Iterable>`
+-----------------------------------------
```
hoefling: Please advise whether the refs in the section titles are useful.

gvanrossum: I prefer not to have them (consistent with the earlier policy against `code` in titles. But then we need to edit the body text to include a ref, since it's still useful to have it.

hoefling: Maybe a generic line like "See also `ProtocolName` documentation." to each section? The stdlib docs don't shed more light on protocol types TBH; PEP 544 is the reading source if one needs details (and it's not even mentioned in the `typing` documentation!).

Or maybe a line like "The `ProtocolName` type is defined in `typing` module.", just to point out where the type can be imported from?

gvanrossum: Let's put a line like
```
See also :py:class:`Iterable`.
```
at the end of each section.

ilevkivskyi: > (and it's not even mentioned in the `typing` documentation!)

It is in the relevant version (i.e. Python 3.8): https://docs.python.org/3.8/library/typing.html

hoefling: I have removed the remote branch by accident, didn't mean to close this PR.

Issue #7266, opened 2019-07-26
get_dynamic_class_hook for method calls


mkurnikov:
In Django, there's a method `from_queryset()`, which creates a new subclass of `Manager` based on passed `QuerySet` class
https://docs.djangoproject.com/en/2.2/topics/db/managers/#from-queryset

`get_dynamic_class_hook` is a hook to use it this case, but unfortunately it only called for functions, not for methods.

msullivan:
Yeah, this sounds like something we need. Would you like to submit a PR?

mkurnikov:
Sorry, I won't be able to do it anytime soon. Maybe @aleksanb would like to do it, as he was the one who pointed out the bug in `django-stubs`.

PR #7704, opened 2019-10-11
Invoke get_dynamic_class_hook on method calls

mkurnikov:
Fixes https://github.com/python/mypy/issues/7266

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -2175,9 +2175,16 @@ class SemanticAnalyzer(NodeVisitor[None],
         if not isinstance(lval, NameExpr) or not isinstance(s.rvalue, CallExpr):
             return
         call = s.rvalue
-        if not isinstance(call.callee, RefExpr):
-            return
-        fname = call.callee.fullname
+        fname = None
+        # check if method call
+        if isinstance(call.callee, MemberExpr):
+            callee_expr = call.callee.expr
+            if isinstance(callee_expr, NameExpr) and callee_expr.fullname:
+                method_name = call.callee.name
+                fname = callee_expr.fullname + '.' + method_name
+        # else check if function call
+        elif isinstance(call.callee, RefExpr):
+            fname = call.callee.fullname
         if fname:
             hook = self.plugin.get_dynamic_class_hook(fname)
             if hook:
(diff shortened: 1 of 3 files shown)


mypy/semanal.py
```diff
@@ -2175,9 +2175,16 @@ def apply_dynamic_class_hook(self, s: AssignmentStmt) -> None:
...
             return
         call = s.rvalue
-        if not isinstance(call.callee, RefExpr):
-            return
-        fname = call.callee.fullname
+        fname = None
+        # check if method call
+        if isinstance(call.callee, MemberExpr):
+            callee_expr = call.callee.expr
+            if isinstance(callee_expr, NameExpr) and callee_expr.fullname:
```
```suggestion
            if isinstance(callee_expr, RefExpr) and callee_expr.fullname:
```

mypy/semanal.py
```diff
@@ -2175,9 +2175,16 @@ def apply_dynamic_class_hook(self, s: AssignmentStmt) -> None:
...
+                method_name = call.callee.name
+                fname = callee_expr.fullname + '.' + method_name
+        # else check if function call
+        elif isinstance(call.callee, RefExpr):
+            fname = call.callee.fullname
```
ilevkivskyi: I believe the order is wrong here: a `RefExpr` (including `MemberExpr`) with a non-`None` full name should always take precedence.

test-data/unit/check-custom-plugin.test
```diff
@@ -516,6 +516,29 @@ class Instr(Generic[T]): ...
...
 
+[case testDynamicClassHookFromClassMethod]
+# flags: --config-file tmp/mypy.ini
+
+from mod import QuerySet, Manager
+
+MyManager = Manager.from_queryset(QuerySet)
+
+reveal_type(MyManager())  # N: Revealed type is '__main__.MyManager'
+reveal_type(MyManager().attr)  # N: Revealed type is 'builtins.str'
```
ilevkivskyi: Please add a check here that `MyManager` can be used as an annotation for a function argument.

ilevkivskyi:
@mkurnikov Note a competing PR appeared https://github.com/python/mypy/pull/7990, will you have time to update your PR soon?

Issue #2393, opened 2016-11-03
Improve googleability by using real error messages in documentation


JukkaL:
For example, http://mypy.readthedocs.io/en/latest/common_issues.html could have the actual mypy error messages for examples to make it easier to Google for common errors.

nathanielksmith:
I think this is something I can take on. I'll run through the examples and add actual error messages where appropriate.

gvanrossum:
There's more than that particular chapter of the docs right?

nathanielksmith:
Yup, and I'm reading the docs through now to look for more stuff. I figured I'd eagerly PR though to make sure my approach is ok.

I'll update my PR description though to be more clear that it's incremental.

PR #7706, opened 2019-10-13
Update mypy error messages in code examples, part 1

hoefling:
First part of changes for #2393.

### DIFF
--- a/docs/source/additional_features.rst
+++ b/docs/source/additional_features.rst
@@ -24,7 +24,7 @@ They can be defined using the :py:func:`@dataclasses.dataclass
         plugins: List[str] = field(default_factory=list)
 
     test = Application("Testing...")  # OK
-    bad = Application("Testing...", "with plugin")  # Error: List[str] expected
+    bad = Application("Testing...", "with plugin")  # error: Argument 2 to "Application" has incompatible type "str"; expected "List[str]"
 
 Mypy will detect special methods (such as :py:meth:`__lt__ <object.__lt__>`) depending on the flags used to
 define dataclasses. For example:
@@ -44,7 +44,7 @@ define dataclasses. For example:
         y: int
 
     OrderedPoint(1, 2) < OrderedPoint(3, 4)  # OK
-    UnorderedPoint(1, 2) < UnorderedPoint(3, 4)  # Error: Unsupported operand types
+    UnorderedPoint(1, 2) < UnorderedPoint(3, 4)  # error: Unsupported left operand type for < ("UnorderedPoint")
 
 Dataclasses can be generic and can be used in any other way a normal
 class can be used:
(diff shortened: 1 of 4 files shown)


docs/source/common_issues.rst
```diff
@@ -270,7 +269,8 @@ unexpected errors when combined with type inference. For example:
 
    lst = [A(), A()]  # Inferred type is List[A]
    new_lst = [B(), B()]  # inferred type is List[B]
-   lst = new_lst  # mypy will complain about this, because List is invariant
+   # mypy will complain about this, because List is invariant
+   lst = new_lst  # error: Incompatible types in assignment (expression has type "List[B]", variable has type "List[A]")
```
hoefling: Please advise whether keeping the explanation comments is useful, and whether the selected format is acceptable:
`<code>  # original explanation` becomes

```
# original explanation
<code>  # mypy error
```

docs/source/common_issues.rst
```diff
@@ -582,7 +582,7 @@ the same line as the import:
 
    # to silence complaints about unused imports
    from typing import List  # noqa
-   a = None  # type: List[int]
+   a = []  # type: List[int]
```
hoefling: In the original example, `mypy` reports `error: Incompatible types in assignment (expression has type "None", variable has type "List[int]")`.

gvanrossum:
I'm not actually 100% sure that this is always an improvement. IIRC the messages shown in the doc examples aren't meant (in most cases) to represent exactly what mypy will say, but are an indication of what's wrong with the line in terms that make sense to the narrative.

So I'll leave it up to @JukkaL to review this.

docs/source/additional_features.rst
```diff
@@ -24,7 +24,7 @@ They can be defined using the :py:func:`@dataclasses.dataclass
         plugins: List[str] = field(default_factory=list)
 
     test = Application("Testing...")  # OK
-    bad = Application("Testing...", "with plugin")  # Error: List[str] expected
+    bad = Application("Testing...", "with plugin")  # error: Argument 2 to "Application" has incompatible type "str"; expected "List[str]"
```
JukkaL: I'd try to keep lines a bit shorter to avoid horizontal scrolling in most cases. Maybe keep them under 100 characters? I think that it's fine to shorten messages. For example leaving out `to "Application"` above seems reasonable. Also, it would be fine to split he comment into multiple lines. Maybe like this:

```
    bad = Application("Testing...", "with plugin")  # error: Argument 2 has incompatible type "str"; 
                                                    # expected "List[str]"
```

docs/source/additional_features.rst
```diff
@@ -44,7 +44,7 @@ define dataclasses. For example:
         y: int
 
     OrderedPoint(1, 2) < OrderedPoint(3, 4)  # OK
-    UnorderedPoint(1, 2) < UnorderedPoint(3, 4)  # Error: Unsupported operand types
+    UnorderedPoint(1, 2) < UnorderedPoint(3, 4)  # error: Unsupported left operand type for < ("UnorderedPoint")
```
JukkaL: Again, this is a bit long. Maybe leave out `("UnorderedPoint")`?

docs/source/common_issues.rst
JukkaL: This looks fine to me. Though here the explanation is pretty informal, and it may be better to be more explicit about what this refers to. Here is one idea:

```
    # Mypy rejects the following assignment, because List is invariant
    lst = new_lst  # error: ...
```
```

hoefling:
After some evaluating the remaining docs, I'm not sure I can tackle this - I'd agree with @gvanrossum that not all spots deserve to have the formal error description. Let me close this for now.

Issue #7539, opened 2019-09-19
`OrderedDict` fails disallow_any_generics, but `OrderedDict[Any, Any]` doesn't run


l0b0:
I've been trying to introduce `disallow_any_generics = true` to an already extensively typed code base of ~48kSLOC, and the only remaining issue is typing `OrderedDict`s.

# Actual behaviour

Given this configuration:

```
[mypy]
check_untyped_defs = true
disallow_any_generics = true
disallow_untyped_defs = true
ignore_missing_imports = true
no_implicit_optional = true
warn_redundant_casts = true
warn_return_any = true
warn_unused_ignores = true
```

Code with implicit `Any`:

```python
from collections import OrderedDict
from typing import Any

def name(value: OrderedDict) -> None:
    pass
```

This fails validation because of `disallow_any_generics = true`:

```text
$ mypy example.py
example.py:4: error: Missing type parameters for generic type
```

Code with typed `OrderedDict`:

```python
from collections import OrderedDict
from typing import Any

def name(value: OrderedDict[Any, Any]) -> None:
    pass
```

This passes validation, but doesn't run:

```text
$ python example.py 
Traceback (most recent call last):
  File "example.py", line 4, in <module>
    def name(value: OrderedDict[Any, Any]) -> None:
TypeError: 'type' object is not subscriptable
```

Versions:

```
$ python --version
Python 3.6.8
$ mypy --version
mypy 0.720
```

(Also failed on mypy 0.711.)

# Expected behaviour

I'd expect either the first snippet to be valid or the second snippet to run. As it is I'm stuck for how to proceed.

JukkaL:
You can use string literal escapes, such as `value: 'OrderedDict[Any, Any]'`. Another option is `from __future__ import annotations` (introduced in Python 3.7).

dms-cat:
@JukkaL Is there no way to do this without quoting the type? Because even if I `if TYPE_CHECKING: from collections import OrderedDict` that triggers an issue in `flake8`:

> F401 'collections.OrderedDict' imported but unused

JelleZijlstra:
You can use `from __future__ import annotations` in 3.7+.

asottile:
do you want `typing.OrderedDict` and not `collections.OrderedDict` for that annotation?

JelleZijlstra:
I was going to say that too but it turns out there is no `typing.OrderedDict`.

asottile:
there is in 3.7 iirc

[...]

PR #7707, opened 2019-10-13
Add recommendation to use string literal escaping

Coronon:
This commit introduces a recommendation for classes that don't support indexing at runtime to be escaped as string literals on missing required generic type args (BARE_GENERIC) error.

It should resolve https://github.com/python/mypy/issues/7539

This is my first time contributing here, so please let me know if there is anything I could have done better. :)

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -569,11 +569,15 @@ string literal types or :py:data:`~typing.TYPE_CHECKING`:
 
    results: 'Queue[int]' = Queue()  # OK
 
-If you are running Python 3.7+ simply activate annotations. Read more about them in `PEP563 <https://www.python.org/dev/peps/pep-0563/>`_.
+If you are running Python 3.7+ you can use ``from __future__ import annotations``
+as a (nicer) alternative to string quotes, read more in :pep:`563`.  For example:
 
 .. code-block:: python
 
    from __future__ import annotations
+   from queue import Queue
+
+   results: Queue[int] = Queue()  # This works at runtime
 
 .. _silencing-linters:
 
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -569,6 +569,11 @@ string literal types or :py:data:`~typing.TYPE_CHECKING`:
 
    results: 'Queue[int]' = Queue()  # OK
 
+If you are running Python 3.7+ simply activate annotations. Read more about them in `PEP563 <https://www.python.org/dev/peps/pep-0563/>`_.
+
+.. code-block:: python
+
+   from __future__ import annotations
 
 .. _silencing-linters:
 
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -533,6 +533,7 @@ Here's the above example modified to use ``MYPY``:
    def listify(arg: 'bar.BarClass') -> 'List[bar.BarClass]':
        return [arg]
 
+.. _not-generic-runtime:
 
 Using classes that are generic in stubs but not at runtime
 ----------------------------------------------------------
--- a/mypy/typeanal.py
+++ b/mypy/typeanal.py
@@ -57,7 +57,7 @@ ARG_KINDS_BY_CONSTRUCTOR = {
 
 GENERIC_STUB_NOT_AT_RUNTIME_TYPES = {
     'queue.Queue',
-    'os.Pathlike',
+    'builtins._PathLike',
 }  # type: Final
 
 
--- a/mypy/typeanal.py
+++ b/mypy/typeanal.py
@@ -929,9 +929,7 @@ def get_omitted_any(disallow_any: bool, fail: MsgCallback, note: MsgCallback,
                 # (string literal escaping) for classes not generic at runtime
                 note(
                     "Subscripting classes that are not generic at runtime may require "
-                    "escaping. If you are running Python 3.7+ you can simply use "
-                    "`from __future__ import annotations`, otherwise please "
-                    "see https://mypy.readthedocs.io/"
+                    "escaping, see https://mypy.readthedocs.io/"
                     "en/latest/common_issues.html#not-generic-runtime",
                     typ,
                     code=codes.TYPE_ARG)
--- a/mypy/typeanal.py
+++ b/mypy/typeanal.py
@@ -56,7 +56,8 @@ ARG_KINDS_BY_CONSTRUCTOR = {
 }  # type: Final
 
 GENERIC_STUB_NOT_AT_RUNTIME_TYPES = {
-    'Queue'
+    'queue.Queue',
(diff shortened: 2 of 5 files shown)


mypy/typeanal.py
```diff
@@ -907,8 +911,13 @@ def get_omitted_any(disallow_any: bool, fail: FailCallback,
             typ = unexpanded_type or typ
             type_str = typ.name if isinstance(typ, UnboundType) else format_type_bare(typ)
 
-            fail(message_registry.BARE_GENERIC.format(quote_type_string(type_str)), typ,
-                 code=codes.TYPE_ARG)
+            if type_str in GENERIC_STUB_NOT_AT_RUNTIME_TYPES:
```
gvanrossum: I'm not sure I understand what's special about OrderedDict. The same thing would happen with any type that's defined both in collections and in typing. And in general I'm not sure that this particular hint belongs in the error message.

Coronon: JukkaL wanted this little reminder in the error message:
https://github.com/python/mypy/issues/7539#issuecomment-534124529

Indeed OrderedDicts are not unique that way, Queue would also have to go into the list.
This error message is useful because OderedDict[Any, Any] would pass mypy checks atm. but wouldn’t run. Many users would probably have difficulties understanding the problem and finding a fix.

Coronon:
Couldn’t we just check this dynamically with
```py
getattr(typ, "__getitem__", None)
```

gvanrossum:
But I believe you can use ‘from typing import OrderedDict’ and the problem will go away. Is that not so?

Coronon:
No, this would indeed run(runtime), but raises a mypy error:
```py
from typing import OrderedDict
from typing import Any

def name(value: OrderedDict[Any, Any]) -> None:
    pass
```
mypy error: 
```
error: Variable "typing.OrderedDict" is not valid as a type
```

ilevkivskyi:
> `Variable "typing.OrderedDict" is not valid as a type`

This is just a bug that should be fixed by updating `mypy.nodes.type_aliases` and `mypy.nodes.type_aliases_target_versions` (as you can see `OrderedDict` is not there yet).

Coronon:
Should this little reminder still be added or not, I just saw the 'good-first-issue' by JukkaL and wanted to contribute. I think what this boils down to is just usability for new users, like me.

Coronon:
> > `Variable "typing.OrderedDict" is not valid as a type`
> 
> This is just a bug that should be fixed by updating `mypy.nodes.type_aliases` and `mypy.nodes.type_aliases_target_versions` (as you can see `OrderedDict` is not there yet).

I could make a new pr with this fix :)

ilevkivskyi:
> Should this little reminder still be added or not, I just saw the 'good-first-issue' by JukkaL and wanted to contribute. I think what this boils down to is just usability for new users, like me.

I think it still makes sense, but:
* This should be made a note on a separate line saying something like `Subscripting classes that are not generic at runtime may require escaping, see https://mypy.readthedocs.io/en/latest/common_issues.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime`. Not an alternative error message.
* We should probably update the docs to make this link shorter
* The note should be shown only for classes in a blacklist that are known to cause problems, and `OrderedDict` should not be in this list.

Coronon:
Would you add a new doc page or just make the name shorter?

ilevkivskyi:
> I could make a new pr with this fix, or just add it to this one :)

I think a separate PR is better
> Would you add a new doc page or just make the name shorter?

I am not a big expert in Sphinx, so I hope there is a way to generate a shorter link without changing the name. If not then you can try finding a better shorter title for that section.

Coronon:
> > I could make a new pr with this fix, or just add it to this one :)
> 
> I think a separate PR is better
> 
> > Would you add a new doc page or just make the name shorter?
> 
> I am not a big expert in Sphinx, so I hope there is a way to generate a shorter link without changing the name. If not then you can try finding a better shorter title for that section.

This has the note working and a shorter link to the doc `https://mypy.readthedocs.io/en/latest/common_issues.html#not-generic-runtime`

I have tested this with `OrderedDict` and then removed it from the blacklist, the note is displayed after the error with the note function I routet to `mypy.typeanal.get_omitted_any` from all callers I could find

docs/source/common_issues.rst
```diff
@@ -533,6 +533,7 @@ Here's the above example modified to use ``MYPY``:
    def listify(arg: 'bar.BarClass') -> 'List[bar.BarClass]':
        return [arg]
 
+.. _not-generic-runtime:
```
ilevkivskyi: Reading now through this section I think it makes sense to add couple sentences at the end about using `from __future__ import annotations` as a (better) alternative to string quotes on Python 3.7+.

mypy/typeanal.py
```diff
@@ -892,9 +902,10 @@ def tuple_type(self, items: List[Type]) -> TupleType:
 
 # Mypyc doesn't support callback protocols yet.
 FailCallback = Callable[[str, Context, DefaultNamedArg(Optional[ErrorCode], 'code')], None]
+NoteCallback = Callable[[str, Context, DefaultNamedArg(Optional[ErrorCode], 'code')], None]
```
ilevkivskyi: This alias is identical to the above one. You can rename it to `MsgCallback` and use it to annotate both `fail` and `note`.

Coronon:
> Thanks for the updates! This looks almost ready, I have few more comments.

Implemented all things as requested (hopefully) :)

mypy/typeanal.py
```diff
@@ -907,17 +918,33 @@ def get_omitted_any(disallow_any: bool, fail: FailCallback,
...
+                typ,
+                code=codes.TYPE_ARG)
+
+            if fullname in GENERIC_STUB_NOT_AT_RUNTIME_TYPES:
+                # Recommend `from __future__ import annotations` or to put type in quotes
+                # (string literal escaping) for classes not generic at runtime
+                note(
+                    "Subscripting classes that are not generic at runtime may require "
+                    "escaping. If you are running Python 3.7+ you can simply use "
+                    "`from __future__ import annotations`, otherwise please "
```
ilevkivskyi: I actually wanted to add this to the docs themselves, not here. Please keep the note as it was, and instead add this sentence (plus maybe also a link to PEP 563) at the end of the docs section.

Coronon: Ok, will do once I landed :)

ilevkivskyi:
@Coronon 
> I could make a new pr with this fix, or just add it to this one :)

Do you still want to make a PR for the `Variable "typing.OrderedDict" is not valid as a type` issue?

Coronon:
I can make it tomorrow if you want :)

ilevkivskyi:
> I can make it tomorrow if you want :)

OK, good.

PR #7708, opened 2019-10-13
Record special Any as precise in report generation

### LINKED ISSUE #6784: Type aliases become red in html report.
Take the following code:
```python
SourceToDisplay = Callable[[int], int]
DisplayToSource = Callable[[int], int]
```
This becomes red, while it shouldn't.
image

Thanks!

AFanaei:
Fixes #6784

### DIFF
--- a/mypy/stats.py
+++ b/mypy/stats.py
@@ -332,7 +332,8 @@ class StatisticsVisitor(TraverserVisitor):
             return
 
         if isinstance(t, AnyType) and is_special_form_any(t):
-            # This is not a real Any type, so don't collect stats for it.
+            # If there is an error in special form definition. it will override TYPE_PRECISE
+            self.record_line(self.line, TYPE_PRECISE)
             return
 
         if isinstance(t, AnyType):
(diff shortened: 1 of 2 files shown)


mypy/stats.py
```diff
@@ -332,7 +332,8 @@ def type(self, t: Optional[Type]) -> None:
             return
 
         if isinstance(t, AnyType) and is_special_form_any(t):
-            # This is not a real Any type, so don't collect stats for it.
+            # If there is an error in special form definition. it will override TYPE_PRECISE
```
ilevkivskyi:
```suggestion
            # If there is an error in special form definition, it will override TYPE_PRECISE
```

Do we actually have a test for this?

AFanaei: no we don't and actually after checking, i found out the comment is wrong.

msullivan: Is there an update that should be made?

AFanaei: i checked this for a while, and i have a couple of questions:  
1. what is the desired behavior? (right now typeAlias lines are marked as unanalysed and thus its rendered Red)
2. this PR will fix the original issue but i'm afraid it may add other undesirable side effects that we don't have tests for them yet.    

if some one could help me to create test cases for both correct and incorrect typeAlias exprs. i may be able to fix this PR.  
thanks.

JukkaL: This PR is fine to land, if this comment is updated. Can you replace the comment with a TODO comment like this: 

`# TODO: What if there is an error in special form definition?`

I don't think that errors is special forms are super important to test, since it's reasonable to expect that there is typically a clean mypy run before generating a report.

PR #7712, opened 2019-10-14
Document how to integrate `mypy` into Vim with ALE

bobwhitelock:
ALE is another popular plugin for integrating linting into Vim; this commit documents how this can also be configured for `mypy` linting (in addition to the existing documentation for Syntastic).


bobwhitelock:
Thanks for the very quick review!

Issue #7631, opened 2019-10-04
Using mypy with mixins and abstract properties


Xyene:
* Are you reporting a bug, or opening a feature request?

Unclear, I'm assuming it's not a bug. It's possible (very likely) that this is a duplicate issue, but I haven't been able to find the likely dupe.

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.

```python
from abc import *


class SomeABC(metaclass=ABCMeta):
    @property
    @abstractmethod
    def foo(cls) -> str:
        pass


class SomeMixin:
    foo = "foo"


class SomeClass(SomeMixin, SomeABC):
    pass
```

* What is the actual behavior/output?
```
a.py:15: error: Definition of "foo" in base class "SomeMixin" is incompatible with definition in base class "SomeABC"
```

Using a mixin to satisfy abstract properties doesn't seem like _too_ uncommon of a use-case, so I'm erring on the side of me doing something wrong, but I haven't been able to get the code above to type-check.

* What is the behavior/output you expect?

I would hope to see no errors while typechecking. Currently a workaround is to make the property not-abstract, but that trades runtime safety for mypy checking on that file.

* What are the versions of mypy and Python you are using?
```
$ mypy --version
mypy 0.730
$ python3 --version
Python 3.7.3
```
* Do you see the same issue after installing mypy from Git master?
Yes, with `mypy 0.740+dev.cc7667a2008ef4cdcf89b06ab4e171bf478b0734`.

* What are the mypy flags you are using? (For example --strict-optional)
None.

Thanks!

Xyene:
This horribly-formatted and probably-broken PoC patch resolves the issue I'm seeing without breaking any tests (except for the typechecking ones, but I'm happy to fix that + clean up the code if support for this should exist in mypy):

```patch
From 5f6ef19792c217c2e8bacb231b0a760e10c04eaa Mon Sep 17 00:00:00 2001
From: Tudor Brindus <me@tbrindus.ca>
Date: Fri, 4 Oct 2019 20:10:40 -0400
Subject: [PATCH] Hack for abstract properties and mixins

---
 mypy/checker.py | 6 ++++++
 1 file changed, 6 insertions(+)

diff --git a/mypy/checker.py b/mypy/checker.py
index bdfda80f..2a2862eb 100644
--- a/mypy/checker.py
+++ b/mypy/checker.py
[...]

PR #7713, opened 2019-10-15
Correctly process properties provided in mixins

Xyene:
This is an attempt at fixing the issue reported in #7631. I'm happy to address any issues that might be present in this first approach.

Fixes https://github.com/python/mypy/issues/7631

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1886,7 +1886,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                 ok = is_subtype(first_sig, second_sig, ignore_pos_arg_names=True)
         elif first_type and second_type:
             ok = is_equivalent(first_type, second_type)
-            if not ok and name in base2.abstract_attributes:
+            if not ok:
                 second_node = base2[name].node
                 if isinstance(second_node, Decorator) and second_node.func.is_property:
                     ok = is_subtype(first_type, cast(CallableType, second_type).ret_type)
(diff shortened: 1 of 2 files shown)


mypy/checker.py
```diff
@@ -1886,6 +1886,10 @@ class C(B, A[int]): ...  # this is unsafe because...
                 ok = is_subtype(first_sig, second_sig, ignore_pos_arg_names=True)
         elif first_type and second_type:
             ok = is_equivalent(first_type, second_type)
+            if not ok and name in base2.abstract_attributes:
```
ilevkivskyi: Maybe I am missing something but I think this should be OK even if `name` is not abstract, so I would remove the right part of `and`.

test-data/unit/check-abstract.test
```diff
@@ -871,6 +871,17 @@ class A(metaclass=ABCMeta):
...
+from abc import ABCMeta, abstractproperty
+class A(metaclass=ABCMeta):
+    @abstractproperty
+    def foo(cls) -> str:
+        pass
+class Mixin:
+    foo = "foo"
+class C(Mixin, A):
+    pass
+[out]
```
ilevkivskyi: I would add couple more tests with a non-abstract property that check:
* the error is still shown if you switch the order of bases in `C`.
* the error is not shown if type in `Mixin` is a strict subtype of type in property (it is safe since it is read-only)

mypy/checker.py
Xyene: I think you're right; I had considered only the abstract case. Fixed.

test-data/unit/check-abstract.test
Xyene: Added some more tests.

PR #7717, opened 2019-10-15
Promote memoryview to bytes

### LINKED ISSUE #7714: Mypy not treating bytes as typing.ByteString
According to the [docs](https://docs.python.org/3/library/typing.html#typing.ByteString), an argument typed as `bytes` should also accept `bytearray` and `memoryview`, but this doesn't seem to be the case.

The following example demonstrates this:

```python
def process(b: bytes) -> None:
    pass

process(memoryview(b"foo"))
```

Mypy produces the following error:

```
error: Argument 1 to "process" has incompatible type "memoryview"; expected "bytes"
```

I found https://github.com/python/mypy/issues/4871 which is essentially the same issue. If there hasn't been any relevant changes since April 2008, perhaps its a docs issue only?

srittau:
Fixes #7714 

Currently this fails a lot of seemingly unrelated tests. I will investigate later.

### DIFF
--- a/mypy/semanal_classprop.py
+++ b/mypy/semanal_classprop.py
@@ -30,6 +30,7 @@ TYPE_PROMOTIONS = {
 TYPE_PROMOTIONS_PYTHON3 = TYPE_PROMOTIONS.copy()  # type: Final
 TYPE_PROMOTIONS_PYTHON3.update({
     'builtins.bytearray': 'bytes',
+    'builtins.memoryview': 'bytes',
 })
 
 # Hard coded type promotions for Python 2.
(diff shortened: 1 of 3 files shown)


srittau:
When running the tests locally with `./runtest.py`, I get a total of 41 of the following errors during `pytest -k "PythonCmdline or PythonEvaluation or StubgenCmdLine or StubgenPythonSuite"`:

```
Expected:
Actual:
  __builtin__.pyi:812: error: Overloaded function signature 3 will never be matched: signature 2's parameter type(s) are the same or broader (diff)
```

Are the Python 2 tests not run during CI?

ilevkivskyi:
> Are the Python 2 tests not run during CI?

Good point! It looks like at least `pythoneval` tests are all skipped if a Python 2 interpreter is not found. This may be the case for newer distros that Travis now has. Would you like to make a separate PR to make `python2` required for 1-2 builds?

ilevkivskyi:
We can land this after the typeshed PR is landed an synced.

Issue #7722, opened 2019-10-16
Regression: false positive when creating a TypedDict with total=False


ilevkivskyi:
On current master this code:
```python
class B(TypedDict, total=False):
    x: int
```
fails with
```
test.py:3: error: Unexpected keyword argument "total" for "__init_subclass__" of "object"
/Users/ilevkivskyi/venvs/mypy_dev/lib/python3.7/site-packages/mypy/typeshed/stdlib/2and3/builtins.pyi:69: note: "__init_subclass__" of "object" defined here
```
I think this may be a release blocker @msullivan

PR #7723, opened 2019-10-16
Fix __init_subclass__ type check

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7722

This skips checking class keywords against `__init_subclass__` for:
* Classes with custom metaclasses, because they can pop the keywords before calling `super()`.
* Classes that are special forms, like `TypedDict` etc.


msullivan:
Yeah, good catch. This will get cherry picked.

PR #7730, opened 2019-10-16
[mypyc] Fix multi-file linker error on macOS

msullivan:
Final variables were being declared twice, once via HeaderDeclaration
and once manually. Get rid of the manual one and always go through
HeaderDeclaration.

Fixes mypyc/mypyc#703.

### DIFF
--- a/mypyc/emitmodule.py
+++ b/mypyc/emitmodule.py
@@ -253,8 +253,6 @@ class GroupGenerator:
             self.declare_module(module_name, emitter)
             self.declare_internal_globals(module_name, emitter)
             self.declare_imports(module.imports, emitter)
-            # Finals must be last (types can depend on declared above)
-            self.define_finals(module_name, module.final_names, emitter)
 
             for cl in module.classes:
                 if cl.is_ext_class:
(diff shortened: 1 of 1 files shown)


mypyc/emitmodule.py
```diff
@@ -721,20 +719,19 @@ def declare_finals(
             static_name = emitter.static_name(name, module)
             emitter.context.declarations[static_name] = HeaderDeclaration(
                 '{}{};'.format(emitter.ctype_spaced(typ), static_name),
+                [self.final_definition(module, name, typ, emitter)],
```
JukkaL: I don't understand how this fixes the issue. It looks like `generate_c_for_modules` can still emit things twice, since there are two loops over declarations, both of which can emity stuff?

msullivan:
It fixed the issue by moving the definition and the extra declaration into a single file, so the compiler resolved it without giving the linker a chance to get upset.

This was not actually a great way to solve it so I merged the two loops and now we don't emit extra declarations.

I'm going to go ahead and merge this because having broken tests on master is pretty bad but I am happy to do any follow-up work you want on this.

JukkaL:
Thanks, it looks cleaner now after the refactor.

PR #7734, opened 2019-10-17
[mypyc] Implement serialization and deserialization of mypyc IR

msullivan:
This is work towards incremental compilation (mypyc/mypyc#682).

Serialization format is JSON blobs similar to mypy itself.

The serialization and deserialization is not actually connected to
anything yet, except that run tests now validate that the
serialization round trips.

### DIFF
--- a/mypyc/ops.py
+++ b/mypyc/ops.py
@@ -159,6 +159,7 @@ class RPrimitive(RType):
     These often have custom ops associated with them.
     """
 
+    # Map from primitive names to primitive types and is used by deserialization
     primitive_map = {}  # type: ClassVar[Dict[str, RPrimitive]]
 
     def __init__(self,
@@ -1757,6 +1758,7 @@ class ClassIR:
         # We populate method_types with the signatures of every method before
         # we generate methods, and we rely on this information being present.
         self.method_decls = OrderedDict()  # type: OrderedDict[str, FuncDecl]
+        # Map of methods that are actually present in an extension class
         self.methods = OrderedDict()  # type: OrderedDict[str, FuncIR]
         # Glue methods for boxing/unboxing when a class changes the type
         # while overriding a method. Maps from (parent class overrided, method)
@@ -1899,8 +1901,8 @@ class ClassIR:
             'ctor': self.ctor.serialize(),
             # We serialize dicts as lists to ensure order is preserved
             'attributes': [(k, t.serialize()) for k, t in self.attributes.items()],
-            # We serialize decls out if the decl isn't in methods otherwise
-            # we refer to the whatever
+            # We try to serialize a name reference, but if the decl isn't in methods
+            # then we can't be sure that will work so we serialize the whole decl.
             'method_decls': [(k, d.fullname if k in self.methods else d.serialize())
                              for k, d in self.method_decls.items()],
             # We serialize method fullnames out and put methods in a separate dict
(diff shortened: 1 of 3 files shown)


mypyc/test/test_serialization.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+
+    assert type(x) is type(y), ("Type mismatch at {}".format(trail), type(x), type(y))
+    if isinstance(x, (FuncDecl, FuncIR, ClassIR)):
+        assert x.fullname == y.fullname
+    elif isinstance(x, OrderedDict):
+        assert len(x.keys()) == len(y.keys()), "Keys mismatch at {}".format(trail)
+        for (xk, xv), (yk, yv) in zip(x.items(), y.items()):
+            assert_blobs_same(xk, yk, trail + ("keys",))
+            assert_blobs_same(xv, yv, trail + (xv,))
```
r3m0t: `trail + (xk,)` ?

mypyc/ops.py
```diff
@@ -1696,6 +1885,94 @@ def concrete_subclasses(self) -> List['ClassIR']:
...
+            'is_abstract': self.is_abstract,
+            'is_generated': self.is_generated,
+            'inherits_python': self.inherits_python,
+            'has_dict': self.has_dict,
+            'builtin_base': self.builtin_base,
+            'ctor': self.ctor.serialize(),
+            # We serialize dicts as lists to ensure order is preserved
+            'attributes': [(k, t.serialize()) for k, t in self.attributes.items()],
+            # We serialize decls out if the decl isn't in methods otherwise
+            # we refer to the whatever
```
JukkaL: Something seems to be missing in the comment: "... to the whatever". Also, this kind of explains what the code does -- maybe this can be written in a more abstract or purpose oriented way?

Also, it would be good to document `methods` in `__init__`.

mypyc/ops.py
```diff
@@ -1733,6 +2010,45 @@ def __init__(
...
+        return ModuleIR(
+            data['fullname'],
+            data['imports'],
+            [ctx.functions[FuncDecl.get_name_from_json(f['decl'])] for f in data['functions']],
+            [ClassIR.deserialize(c, ctx) for c in data['classes']],
+            [(k, deserialize_type(t, ctx)) for k, t in data['final_names']],
+        )
+
+
+def deserialize_scc(data: Dict[str, JsonDict], ctx: DeserMaps) -> Dict[str, ModuleIR]:
```
JukkaL: Add docstring. Explain that this populates `ctx`, and that this expects that `ctx` contains information from all dependencies as a pre-condition?

mypyc/test/test_run.py
```diff
@@ -182,6 +183,9 @@ def run_case(self, testcase: DataDrivenTestCase) -> None:
                     print(line)
                 assert False, 'Compile error'
 
+            # Check that serialization works on this IR
+            check_serialization_roundtrip(ir)
```
JukkaL: Nice, this is a clever way to test serialization -- without writing any new tests!

Have you checked that the existing tests cover all the interesting cases? What about things like multiple SCCs that depend on each other in different ways.

mypyc/test/test_serialization.py
msullivan: good catch

mypyc/test/test_run.py
msullivan: Well, it catches all of the interesting cases that I know we had bugs in ;).
(My original take only ran this test on genops tests, since run wasn't set up in a way that was conducive to it, but I hacked it in to see if it would help and it found bugs that testing with genops didn't, so I did #7731 to enable the test on run.)

Our run tests are pretty decent I think at covering most of what mypyc does.

It definitely doesn't really test any SCC stuff, because the code that drives deserialization an SCC at a time hasn't been written yet (well, it has, but not as part of this PR) so it is definitely not tested :P. This test just deserializes everything at once.

mypyc/ops.py
msullivan: ... In grad school I once accidentally submitted a fellowship application that listed "AND STUFF" as one of my skills.

PR #7754, opened 2019-10-19
Add option to print absolute file paths

maksbotan:
It is a common use case to integrate mypy into an IDE or other
environment where it's run automatically and its output is processed by
other programs instead of being read by a human.

IDEs usually are able to parse paths, lines and messages printed by mypy
and show those messages directly in their code editor right under
affected lines.

However, some IDEs, like PyCharm, support reading only absolute file
paths from tools' output.

For example, there is an issue in PyCharm's issue tracker asking for a
way to read relative paths from tools:
https://youtrack.jetbrains.com/issue/IDEA-48163

Sadly, that issue did not receive any action for 11 years already,
apparently. And there is a comment from someone saying that mypy is
affected by this problem, not having any way to force absolute paths in
the output.

This commit adds a new option, --show-absolute-path, to fix exactly
this, with a straightforward implementation. I believe that this simple
fix will bring considerable benefits to PyCharm users and potentially
other people as well.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -203,7 +203,8 @@ def _build(sources: List[BuildSource],
                     options.show_column_numbers,
                     options.show_error_codes,
                     options.pretty,
-                    lambda path: read_py_file(path, cached_read, options.python_version))
+                    lambda path: read_py_file(path, cached_read, options.python_version),
+                    options.show_absolute_path)
     plugin, snapshot = load_plugins(options, errors, stdout)
 
     # Construct a build manager object to hold state during the build.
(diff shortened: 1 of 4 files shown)


ilevkivskyi:
It is however trivial to work around this problem, since PyCharm knows project root itself, see e.g. https://github.com/dropbox/mypy-PyCharm-plugin. I am however not against this if other people would like to have this flag.

maksbotan:
Well, I would say that a third-party plugin with a complicated installation procedure for Python beginners with no Java experience is not exactly a good solution to the such a simple task.

Moreover, your plugin shows a list of errors, but with an "Output Filter" PyCharm will show errors on lines with them, so that developer does not have to look away from the code to see errors. In my opinion, this is better user experience.

maksbotan:
So, what is your opinion. I still think this feature is useful and does no harm.

By the way, PyCharm is not the only tool out there. I suppose that this could be also in some CI workflows to simplify things.

PR #7784, opened 2019-10-23
Mypy command line options reference anchors

hoefling:
Also added references in ``command_line.rst``. This is another suggestion for adding crossrefs to the docs. If you like this suggestion, I will add more PRs that replace verbatim option texts with references in chunks (same approach as with the stdlib crossrefs).

### Motivation

When reading the docs in the other PR I made earlier, I noticed the command line options are referenced throughout the docs. For that, a label is added above the command line paragraph which is linked by using the `ref` role. Although this does generate a valid link, e.g. in [Import discovery](https://mypy.readthedocs.io/en/latest/config_file.html#import-discovery):

image

this doesn't look nice (the verbatim formatting is missing and the double hyphens are merged to an em dash) and the overall construct (adding an anchor and using `ref` for referencing) can be simplified by using the `:option:` directive. When rendered, it will add an anchor automatically:

image

The anchor can then be referenced by using the `:option:` role:

image

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -1,5 +1,7 @@
 .. _command-line:
 
+.. program:: mypy
+
 The mypy command line
 =====================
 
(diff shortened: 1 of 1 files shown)


hoefling:
Another side effect would be implicitly fixing (rare) typos in command args, i.e. `--disallow-untype-defs` instead of `--disallow-untyped-defs` in `error_codes.rst`.

PR #7794, opened 2019-10-25
Remove open plugin, not needed with overloaded signatures on typeshed

ilai-deutel:
python/typeshed#3371 adds overloaded signatures for `open()`, `Path.open()`, etc.
Consequently:
- Plugins that analyze the signatures of these functions can be removed
- Tests have to be changed. For instance, `reveal_type(open())` now produces `error: All overload variants of "open" require at least one argument` instead of `Too few arguments for "open"`.

This pull request should probably not be merged before python/typeshed#3371. However, we have a circular dependency: `mypy_selftest` in the typeshed PR [fails](https://travis-ci.org/python/typeshed/jobs/602655894) because of the issues with `mypy` tests mentioned above. We can use CI on this PR to make sure the tests will pass on typeshed after merging both PRs.

After python/typeshed#3371 is merged, the typeshed submodule will need to be updated.


test-data/unit/pythoneval.test
```diff
@@ -321,8 +325,8 @@ reveal_type(p.open(errors='replace', mode='rb'))
 mode = 'rb'
 reveal_type(p.open(mode=mode, errors='replace'))
 [out]
-_program.py:3: note: Revealed type is 'typing.BinaryIO'
-_program.py:4: note: Revealed type is 'typing.BinaryIO'
+_program.py:3: note: Revealed type is 'typing.IO[Any]'
```
JelleZijlstra: These seem like regressions. Why can mypy no longer infer BinaryIO here?

ilai-deutel: `p.open(mode='rb', errors='replace')` raises an exception (`ValueError: binary mode doesn't take an errors argument`), so its type is not really `BinaryIO`.

`mypy` infers it is a `IO[Any]` because of the [fallback overload](https://www.python.org/dev/peps/pep-0586/#interactions-with-overloads) with `mode: str`, but ideally `open` should have a return type of `NoReturn` in this case. I wonder if adding an overload variant with a return type of `NoReturn` would solve the problem.

JukkaL: `NoReturn` is not really intended for error cases. Falling back to the default return type on error case seems fine to me.

ilai-deutel: To be clear, when you say "default return type", do you mean `IO[Any]` or `BinaryIO`?

ilevkivskyi: Looking at this it seems to me it is better to improve this plugin instead of removing it. For example, the `p.open(mode='rb', errors='replace')` call will silently fall back to the default overload, but with the plugin we can detect the error and give a nice error message (matching the runtime one).

ilai-deutel: I modified the plugin to detect use of text-only arguments in binary mode.

ilevkivskyi:
The CI failure is caused by https://github.com/python/typeshed/issues/3446.

ilai-deutel:
Fixed upstream, AppVeyor tests now pass

mypy/plugins/default.py
```diff
@@ -107,55 +107,65 @@ def get_class_decorator_hook(self, fullname: str
...
-        if isinstance(mode_expr, StrExpr):
-            mode = mode_expr.value
-    if mode is not None:
-        assert isinstance(default_return_type, Instance)  # type: ignore
-        if 'b' in mode:
-            return api.named_generic_type('typing.BinaryIO', [])
-        else:
-            return api.named_generic_type('typing.TextIO', [])
-    return default_return_type
+        return None
```
JukkaL: Style nit: we prefer a bare `return` with no `None` when the return type is annotated as `None`.

mypy/plugins/default.py
```diff
@@ -107,55 +107,65 @@ def get_class_decorator_hook(self, fullname: str
...
+
+    mode_str = try_getting_str_literals(
+        args[mode_arg_index][0], arg_types[mode_arg_index][0]
+    )
+    if mode_str is None or len(mode_str) != 1 or 'b' not in mode_str[0]:
+        # mode cannot be found, or is not binary
+        return None
```
JukkaL: Similar to above.

test-data/unit/python2eval.test
```diff
@@ -189,7 +189,6 @@ f.write(u'foo')
 f.write('bar')
 f.close()
 [out]
-_program.py:3: error: Argument 1 to "write" of "IO" has incompatible type "unicode"; expected "str"
```
JukkaL: This seems like a regression. `f` should be opened as a binary file in Python 2.

ilevkivskyi:
Since Jukka made previous reviews, I think it would make sense if he will review.

ilai-deutel:
Hi @JukkaL, is there anything else I need to do before we can merge this PR?

msullivan:
I'm sorry about this, but I'm going to close this in favor of #9275. (Which I'm also not sure if it is going to get merged)

Issue #7301, opened 2019-08-08
Plugin add_classmethod


dmontagu:
**Feature request:**
A function like `mypy.plugins.common.add_method`, but with support for classmethods.

-----

More context:

I am working on a plugin, and would like to be able to detect and raise errors when an inherited (factory) classmethod is called, but *only* if certain conditions hold for the type (which I am computing in the `base_model_hook` and storing on `ctx.cls.info.metadata`).

The method is inherited from a parent class, and when the method hook is called it only contains references to the parent class, not the calling class (on which I've stored the metadata). (The parent does not have the appropriate type-specific metadata I want to access.)

In order to get around this, I had the idea to directly add a classmethod to the subclass so that it could trigger the method_hook itself. (Note: I'm not sure whether this would work since I'd be generating the method in the `base_class_hook`; I'm not sure whether it would get picked up by the `get_method_hook` or not.)

**(Any insight into how I can create a conditionally-erroring classmethod, where the error occurs based on data computed in the `base_model_hook`, would be appreciated.)**

Whether or not the above approach would work for this purpose, I didn't see an easy way to add a classmethod (similarly to how `add_method` works), and I think this might be a nice capability to suppport.

(This came up in a discussion with @ilevkivskyi on gitter, and he suggested I create an issue.)

ilevkivskyi:
I think we can just add an optional argument `is_classmethod=False` (maybe also `is_staticmethod=False`?) to the existing function (and essentially copy the corresponding part from `semanal_namedtuple.py`).

msullivan:
Would you like to submit a PR, @dmontagu

dmontagu:
@msullivan I can take a look tonight and give it a shot, but am still very new to mypy. So as long as you don't mind it being an "attempt", I can open a PR. (But I haven't put any effort into this yet if someone else is able to do it very quickly/easily.)

PR #7796, opened 2019-10-25
Add support to add_method for classmethods, __new__, and staticmethods

dmontagu:
Adds keyword arguments to `add_method` to create classmethods, staticmethods, and `__new__`. Hopefully this makes it easier to add such methods in mypy plugins.

This is based on the implementation in semanal_namedtuple.py,  as suggested by @ilevkivskyi in #7301.

I still need to add tests, but I wanted to make sure this was along the right path before figuring that out.

### DIFF
--- a/mypy/plugins/common.py
+++ b/mypy/plugins/common.py
@@ -125,6 +125,8 @@ def add_method(
     func = FuncDef(name, args, Block([PassStmt()]))
     func.info = info
     func.type = set_callable_name(signature, func)
+    func.is_class = is_classmethod
+    func.is_static = is_staticmethod
     func._fullname = info.fullname() + '.' + name
     func.line = info.line
 
--- a/mypy/plugins/common.py
+++ b/mypy/plugins/common.py
@@ -2,11 +2,11 @@ from typing import List, Optional
 
 from mypy.nodes import (
     ARG_POS, MDEF, Argument, Block, CallExpr, Expression, SYMBOL_FUNCBASE_TYPES,
-    FuncDef, PassStmt, RefExpr, SymbolTableNode, Var
+    Decorator, FuncDef, NameExpr, PassStmt, RefExpr, SymbolTableNode, Var
 )
 from mypy.plugin import ClassDefContext
 from mypy.semanal import set_callable_name
-from mypy.types import CallableType, Overloaded, Type, TypeVarDef, get_proper_type
+from mypy.types import CallableType, Overloaded, Type, TypeType, TypeVarDef, get_proper_type
 from mypy.typevars import fill_typevars
 from mypy.util import get_unique_redefinition_name
 from mypy.typeops import try_getting_str_literals  # noqa: F401  # Part of public API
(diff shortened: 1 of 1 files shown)


dmontagu:
@ilevkivskyi is this what you had in mind? Is there a good way for me to go about adding tests for this even though it isn't used by any of the provided plugins?

mypy/plugins/common.py
```diff
@@ -86,6 +86,9 @@ def add_method(
         return_type: Type,
         self_type: Optional[Type] = None,
         tvar_def: Optional[TypeVarDef] = None,
```
dmontagu: Does this mean that you can currently only include a single TypeVar when using this method? Is there a good reason for that limitation?

ilevkivskyi:
I think it would be fair if @msullivan will review this. Fro myself I can add two large scale comments:
* This definitely needs tests
* IIRC `@classmethod` and `@staticmethod` are removed from the decorator lists by the semantic analyzer, so we need to be sure to match this here (you can try inspecting generated and actual class methods).

dmontagu:
@ilevkivskyi Thanks for taking the time to give feedback!

------

I'm using essentially this code in https://github.com/samuelcolvin/pydantic/blob/97c9b5cdbff9b83bed38ef243488eed01b070e93/pydantic/mypy.py (minus staticmethod handling) and noticed something weird that I think might be related to the point about the difference between generated and actual class methods.

Calls to a method added with `is_classmethod=True` type check properly when called via the class, but don't behave properly when called via an instance of the class. In particular, if I call `reveal_type(instance.classmethod)`, I get the following difference:

For an actual classmethod:
```
note: Revealed type is 'def (obj: Any) -> tests.mypy.plugin-success.Model*'
```

For a classmethod added via `add_method` as in this PR
```
note: Revealed type is 'def [Model] (_cls: Type[Model`-1], obj: Any) -> Model`-1'
```

Any ideas about where to look to handle this? (In particular, is this handled properly now in `semanal_namedtuple.py` and I am doing something wrong here?)

Or, perhaps related, any suggestions about what to pay attention to if I "[inspect] generated and actual class methods"?

--------

Separately, are there any tests in the codebase now that test plugin-oriented functionality? As far as I can see, for example, the `add_method` function as it exists currently is only used in the attrs and dataclasses plugins, and presumably only tested indirectly through tests of the output with those plugins enabled?

I'm happy to write tests for the added functionality, but I don't know what would be idiomatic in the mypy codebase, or even where to start looking to begin to answer that question. Should I create a "test" plugin? Is there a way to *unit*-test a function like this? Are there any reasonably analogous examples in existing tests?

ilevkivskyi:
> Or, perhaps related, any suggestions about what to pay attention to if I "[inspect] generated and actual class methods"?

I mean have your plugin activated, then manually write a method that should like a method you want generated. The set a breakpoint somewhere in `mypy/checker.py` and in the debugger inspect both methods (e.g. corresponding `FuncDef`s and their attributes) to see that they are identical.
> Separately, are there any tests in the codebase now that test plugin-oriented functionality?

There are (relatively) many tests, see e.g. https://github.com/python/mypy/blob/master/test-data/unit/check-custom-plugin.test. The corresponding test plugins live in https://github.com/python/mypy/tree/master/test-data/unit/plugins

dmontagu:
Thank you! This is very helpful!

97littleleaf11:
Hi @dmontagu , any updates on this PR?

Issue #7363, opened 2019-08-18
Statement is unreachable when checking if `Match[str]`.group(...) type is `None`


jack1142:
Hi, I am not entirely sure if this is the right place to report it so please tell me if I should submit the issue somewhere else.

**Issue type:** Bug
**Problematic code:**
```py
from typing import Match, Optional
import re


def some_function(some_match: Match[str]) -> bool:
    maybe_match_group = some_match.group(1)
    if maybe_match_group is None:
        group_matched = True
    else:
        group_matched = False
    return group_matched


pattern = re.compile("(str)?")
match = pattern.fullmatch("")
if match is not None:
    print(some_function(match))
```
**Error message:**
```
main.py:8: error: Statement is unreachable
```
**Expected behaviour:** 
No error as the statement is reachable (and it's used when you run this code). I tried using `Match[Optional[str]]` to fix this but it's not valid argument for `Match`.
Another thing to note is that this code will not error if I'll just use `Match` without specifying type of string.

**Python version:**: 3.7.4
**Mypy version:** 0.730, occurs on master as well
**Mypy flags:** `--warn-unreachable`

Michael0x2a:
This is a legitimate bug/limitation in mypy. In short, what's happening is that the `Match.group(...)` function is currently [defined in typeshed](https://github.com/python/typeshed/blob/master/stdlib/3/typing.pyi#L528) to only ever return a str or a byte, depending on whether you have a `Match[str]` or a `Match[byte]`.

This is obviously incorrect, as you noted. And due to this incorrect annotation, the `--warn-unreachable` flag ends up assuming that the "if the group is None" branch will never be run.

We actually did experiment with having the return be an Optional, but it ended up [causing regressions in various codebases we were testing on](https://github.com/python/typeshed/pull/3190). E.g. some snippet of code was using a regex where there would *always* be N groups and so rightfully skipped the 'is this group None?' checks.

The reason why not specifying the type of the string works is because such a type is equivalent to doing `Match[Any]`. So, `group()` ends up also returning Any.

cc @ilevkivskyi, since we were just talking about this the other day. Maybe writing a plugin is the right approach after all? Doing `re.compile(...).groups` will at least get you the upper bound on the number of groups there'll be, and doing `sre_parse.parse(...)` seems to get us an AST of the regex we could perhaps try analyzing.

ilevkivskyi:
OK, it looks like we need some compromise here, possible plan is:
[...]

PR #7803, opened 2019-10-27
Add plugin to infer more precise regex match types

Michael0x2a:
This pull request adds a plugin to make mypy infer more precise types when grabbing regex groups: the plugin will when possible analyze original regex to deduce whether a given group is required or not.

```
from typing_extensions import Final, Literal
import re

pattern: Final = re.compile("(a)(b)*")
match: Final = pattern.match("")
if match:
    reveal_type(match.groups())  # Revealed type is Tuple[str, Optional[str]]
    reveal_type(match.group(0))  # Revealed type is str
    reveal_type(match.group(1))  # Revealed type is str
    reveal_type(match.group(2))  # Revealed type is Optional[str]

    index: int
    reveal_type(match.group(index))  # Revealed type is Optional[str]

    # Error: Regex has 3 total groups, given group number 5 is too big
    match.group(5)
```

To track this information, I added in an optional 'metadata' dict field to the Instance class, similar to the metadata dict for plugins in TypeInfos. We skip serializing this dict if it does not contain any data.

A limitation of this plugin is that both the pattern and the match variables must be declared to be final. Otherwise, we just default to using whatever types are defined in typeshed.

This is because we set and erase the metadata field in exactly the same way we set and erase the `last_known_value` field in Instances: both kinds of info are "transient" and are unsafe to keep around if the variable reference is mutable.

This limitation *does* end up limiting the usefulness of this plugin to some degree: it won't support common patterns like the below, since variables aren't allowed to be declared final inside loops:

```
for line in file:
    match = pattern.match(line)
    if match:
        ...
```

Possibly we can remove this limitation by making mypy less aggressive about removing this transient info by tracking the "lifetime" of this sort of data in some way?

This pull request should mostly address https://github.com/python/mypy/issues/7363, though it's unclear if it really fully resolves it: we might want to do something about the limitation described above and re-tune typeshed first.

The other mostly unrelated change this PR makes is to refactor some of the helper functions in checker.py into typeops.py so I could use them more cleanly in the plugin.

### DIFF
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -10,7 +10,7 @@ from mypy.types import (
 from mypy.subtypes import is_subtype
 from mypy.join import join_simple
 from mypy.sametypes import is_same_type
-from mypy.erasetype import remove_instance_last_known_values
+from mypy.erasetype import remove_instance_transient_info
 from mypy.nodes import Expression, Var, RefExpr
 from mypy.literals import Key, literal, literal_hash, subkeys
 from mypy.nodes import IndexExpr, MemberExpr, NameExpr
@@ -251,7 +251,7 @@ class ConditionalTypeBinder:
                     restrict_any: bool = False) -> None:
         # We should erase last known value in binder, because if we are using it,
         # it means that the target is not final, and therefore can't hold a literal.
-        type = remove_instance_last_known_values(type)
+        type = remove_instance_transient_info(type)
 
         type = get_proper_type(type)
         declared_type = get_proper_type(declared_type)
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2,7 +2,6 @@
 
 import itertools
 import fnmatch
-import sys
 from contextlib import contextmanager
 
 from typing import (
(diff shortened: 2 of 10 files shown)


Michael0x2a:
One other note: it's possible that this plugin is a bit too intrusive to include in mypy itself. While the actual regex-related code is pretty lightweight, it does require us to now have to worry about interfacing with the `sre_parse` module and the (undocumented?) Python regex internals.

So maybe this is better off being spun out into a separate plugin? Not sure.

ilevkivskyi:
> So maybe this is better off being spun out into a separate plugin? Not sure.

FWIW I think everything related to stdlib (like `builtins`, `ctypes`, `contextlib`, and `dataclasses`) should be bundled with mypy and all plugins for third party libs (like `django` or `sqlalchemy`) should be separate (`attrs` is the only historic exception form this pattern).

JukkaL:
I think that the requirement to use `Final` unfortunately limits the usefulness of this, and may make it a bit too marginal to be added to mypy proper. The main problem I see is that this doesn't seem very discoverable -- users would need to learn that for return values from these specific functions, they should use `Final`.  I wonder if it would be better to detect locals that are only ever assigned once, and effectively make them implicitly final. This way the new feature would work for most existing code without changes, making it much more generally useful.

A few other random notes:
* Instead of adding a new attribute to `Instance`, what about merging the attribute with `last_known_value` to save memory? Dictionary might be too big in terms of memory use for this, however. Something like a linked list of "extra attribute objects" would be pretty memory-efficient. Not sure if the latter optimization is worth it though.
* The PR could be made smaller by isolating the refactoring which moves code to `typeops` into a separate PR.

97littleleaf11:
`(` or `)` inside comments of multiline regex causes false positive of mypy_primer.

PR #7804, opened 2019-10-28
Add Capturable* argparse structures to mypy.main

### LINKED ISSUE #7800: stderr is not always captured by mypy.api.run
Sometimes, stderr is printed instead of being returned from `mypy.api.run`.

Here is an SSCCE:
``` python
import mypy.api
print(mypy.api.run(['--foo']))
```
mypy 0.701:
```
('', 'usage: mypy [-h] [-v] [-V] [more options; see below]\n            [-m MODULE] [-p PACKAGE] [-c PROGRAM_TEXT] [files ...]\nmypy: error: unrecognized arguments: --foo\n', 2)
```
mypy 0.710:
```
usage: mypy [-h] [-v] [-V] [more options; see below]
            [-m MODULE] [-p PACKAGE] [-c PROGRAM_TEXT] [files ...]
mypy: error: unrecognized arguments: --foo
('', '', 2)
```

dmtucker:
Fix #7800 

Most of this code is yanked from the standard library since the same behaviors are desired, just with plumbing for `stdout`/`stderr`. One difference of note is that `version` is mandatory in `CapturableVersionAction.__init__` to get around this error:
```
mypy/main.py:339: error: "ArgumentParser" has no attribute "version"  [attr-defined]
                version = parser.version
                          ^
```
I'm unaware of a situation where [that line](https://github.com/python/cpython/blob/955f96f6aae7c1a54d88c3f7a51c2e142ac7e4d4/Lib/argparse.py#L1067) wouldn't raise an `AttributeError` 🤔 
Anyways, I didn't want to patch `sys.stdout`/`sys.stderr` in order to avoid reintroducing #6125.


mypy/test/testapi.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_capture_bad_opt(self) -> None:
+        """stderr should be captured when a bad option is passed."""
+        _, stderr, _ = mypy.api.run(['--some-bad-option'])
+        assert stderr != ''
+
+    def test_capture_empty(self) -> None:
+        """stderr should be captured when a bad option is passed."""
+        _, stderr, _ = mypy.api.run([])
+        assert stderr != ''
```
JukkaL: Similar to above.

mypy/test/testapi.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_capture_help(self) -> None:
+        """stdout should be captured when --help is passed."""
+        stdout, _, _ = mypy.api.run(['--help'])
+        assert stdout != ''
```
JukkaL: Similar to above.

mypy/test/testapi.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def test_capture_version(self) -> None:
+        """stdout should be captured when --version is passed."""
+        stdout, _, _ = mypy.api.run(['--version'])
+        assert stdout != ''
```
JukkaL: Similar to above.

mypy/main.py
```diff
@@ -252,6 +257,89 @@ def infer_python_executable(options: Options,
...
+
+        If you override this in a subclass, it should not return -- it
+        should either exit or raise an exception.
+        """
+        self.print_usage(self.stderr)
+        args = {'prog': self.prog, 'message': message}
+        self.exit(2, gettext('%(prog)s: error: %(message)s\n') % args)
+
+
+class CapturableVersionAction(argparse.Action):
```
JukkaL: Add docstring (maybe refer to the docstring that I requested for `CapturableArgumentParser`).

mypy/main.py
```diff
@@ -252,6 +257,89 @@ def infer_python_executable(options: Options,
   Define MYPY_CACHE_DIR to override configuration cache_dir path."""  # type: Final
 
 
+class CapturableArgumentParser(argparse.ArgumentParser):
```
JukkaL: Add docstring that explains what this does and why we need to do this.

mypy/main.py
```diff
@@ -252,6 +257,89 @@ def infer_python_executable(options: Options,
...
+
+    def __init__(self, *args: Any, **kwargs: Any):
+        try:
+            self.stdout = kwargs.pop('stdout')
```
JukkaL: Maybe you can avoid the try statements by using `kwargs.pop('stdout', sys.stdout)` (and similarly for stderr).

Issue #7763, opened 2019-10-21
False positive [str-format] for Union[Tuple[], Tuple[]]


jolaf:
The following code:
```python3
from typing import Tuple, Union

t: Union[Tuple[str, int], Tuple[int, str]] = ('A', 5)
print("%s %s" % t)
```
produces the following output:
```python3
$ python3 Test.py 
A 5
$ mypy Test.py 
Test.py:4: error: Not enough arguments for format string
Found 1 error in 1 file (checked 1 source file)
$ mypy --version
mypy 0.740
$ python3 --version
Python 3.6.8
```
Note that `t` is in any case a two-member tuple, but `mypy` still treats it as a single value, which is clearly wrong.

If only one `Tuple[]` is left inside the `Union[]` or `Union[]` is removed altogether, the problem disappears.

msullivan:
I guess we should do something like, if it is a union, take the join of the elements and then look at that. The relevant code here is in `check_simple_str_interpolation`.

But really it would be nice if we could come up with some sort of good generic interface for this sort of destructuring of types. Basically every time we implement some new check we have to go back and fix some weird union behavior, or something similar.

vbarbaresi:
The fix was merged, I suppose this can be closed

PR #7809, opened 2019-10-28
handle Union type in string interpolation checker #7763

vbarbaresi:
The problematic case was

```
t: Union[Tuple[str, int], Tuple[int, str]] = ('A', 1)
"%s %s" % t
```

The Union was treated as a single value instead of looking at the types inside
The fix consists in checking each of the union types

Two questions:
- I created a `TempNode` for each of the sub types and added the current line to keep some context, not sure it's the right way to do it.
- I'm running the check for all of the types in the union, but I've seen in the code base that it's usually checked for `any()`.
I suppose it shouldn't return an error if one of the types matches the formatting arguments, for instance with `Union[Tuple[str, int], Tuple[str, str, str]]`
If you confirm, I will have to refactor part of the check in a sub-method

### DIFF
--- a/test-data/unit/check-expressions.test
+++ b/test-data/unit/check-expressions.test
@@ -1332,7 +1332,7 @@ def f(t: Tuple[int, ...]) -> None:
 
 [case testStringInterpolationUnionType]
 from typing import Tuple, Union
-a: Union[Tuple[int, str], Tuple[str,int]] = ('A', 1)
+a: Union[Tuple[int, str], Tuple[str, int]] = ('A', 1)
 '%s %s' % a
 '%s' % a  # E: Not all arguments converted during string formatting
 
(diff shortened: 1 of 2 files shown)


test-data/unit/check-expressions.test
```diff
@@ -1330,6 +1330,16 @@ def f(t: Tuple[int, ...]) -> None:
...
 
+[case testStringInterpolationUnionType]
+from typing import Tuple, Union
+a: Union[Tuple[int, str], Tuple[str,int]] = ('A', 1)
+'%s %s' % a
+'%s' % a  # E: Not all arguments converted during string formatting
+
+b: Union[Tuple[int, str], Tuple[int, int], Tuple[str, int]] = ('A', 1)
+'%s %s' % b
+'%s %s %s' % b  # E: Not enough arguments for format string
```
JukkaL: Add test case where one union item is valid and another is invalid. It should be an error if any item is invalid.

vbarbaresi: done ✅

PR #7810, opened 2019-10-28
[suggest] Fix inference of returned collection

msullivan:
Do this by suppressing the use of Any contexts, which cause arguments
to be misinferred. This can cause certain things to be inferred as
`<nothing>` instead of `Any`, but that's seems fine.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3593,6 +3593,11 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         """
         if node in self.type_overrides:
             return self.type_overrides[node]
+        # Don't use an Any type context, since that will cause more
+        # anys than we want for things that consume the inferred
+        # types.
+        if isinstance(get_proper_type(type_context), AnyType):
+            type_context = None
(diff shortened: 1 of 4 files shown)


mypy/checkexpr.py
```diff
@@ -3593,6 +3593,11 @@ def accept(self,
         """
         if node in self.type_overrides:
             return self.type_overrides[node]
+        # Don't use an Any type context, since that will cause more
+        # anys than we want for things that consume the inferred
+        # types.
+        if isinstance(get_proper_type(type_context), AnyType):
+            type_context = None
```
ilevkivskyi: This may actually be a big change. I am curious how this behaves w.r.t. to things like
```python
def foo(x: Any) -> None: ...
T = TypeVar('T', bound=int)
def g(x: T) -> T: # also -> Optional[T] and -> List[T], IIRC all three are special cased
    ...
foo(g('no'))
```
If these example don't change behaviour, then I think this is good to go.

msullivan: Doesn't seem to change anything there

PR #7811, opened 2019-10-28
[suggest] Allow requesting line numbers *in* the function

msullivan:
Do this by grabbing the last function that starts on or above the
line specified.

### DIFF
--- a/mypy/suggestions.py
+++ b/mypy/suggestions.py
@@ -478,7 +478,7 @@ class SuggestionEngine:
         # We must be sure about any edits in this file as this might affect the line numbers.
         tree = self.ensure_loaded(self.fgmanager.graph[modname], force=True)
         node = None  # type: Optional[SymbolNode]
-        closest_line = None
+        closest_line = None  # type: Optional[int]
         # TODO: Handle nested functions.
         for _, sym, _ in tree.local_definitions():
             if isinstance(sym.node, (FuncDef, Decorator)):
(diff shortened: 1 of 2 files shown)


mypy/suggestions.py
```diff
@@ -476,14 +478,20 @@ def find_node_by_file_and_line(self, file: str, line: int) -> Tuple[str, SymbolN
...
+            elif isinstance(sym.node, Decorator):
+                sym_line = sym.node.func.line
             # TODO: add support for OverloadedFuncDef.
+            else:
+                continue
+
+            # We want the closest function above the specified line
+            if sym_line <= line and (closest_line is None or sym_line > closest_line):
+                closest_line = sym_line
+                node = sym.node
```
ilevkivskyi: This has one downside, nested functions. I mean currently it doesn't work either, but now it will instead add an annotation to enclosing function (that may be already annotated). I have hit this several times last week.

I am not sure what to do here. If there is a simple way to support nested functions, then it would make sense to add it. Otherwise maybe open an issue or add a TODO here.

test-data/unit/fine-grained-suggest.test
```diff
@@ -641,6 +641,25 @@ class C:
...
+[file foo.py]
+def foo():
+    # hi
+    return 1
+
+def bar():
+    return 1.0
+
+def baz():
+    return 'test'
```
ilevkivskyi: Just in case I would add a method and/or a decorator here, not only functions.

PR #7813, opened 2019-10-29
[suggest] Allow making suggestions even when there are errors

msullivan:


test-data/unit/fine-grained-suggest.test
```diff
@@ -807,3 +807,16 @@ def bar():
 () -> typing.Dict[str, int]
 () -> typing.Dict[Any, Any]
 ==
+
+[case testSuggestWithErrors]
```
JukkaL: Do we have a test case where we have  blocking error such as parse error?

PR #7817, opened 2019-10-29
[mypyc] Support generating extension classes for dataclasses

### LINKED ISSUE #671: Make tuple types subtypes of Sized, Container, Reversible, etc.
Currently `(1, 1)` is compatible with `Sequence[int]` and `Iterable[int]` but not with `Sized` or `Container[int]`, which is inconsistent.

This code should be fine, but currently mypy complains about it:

```
from typing import Sized, Container
def f(x: Sized) -> None: pass
f((1, 2)) # Argument 1 to "f" has incompatible type "Tuple[int, int]"; expected "Sized"
def g(x: Container[int]) -> None: pass
g((1, 2)) # Argument 1 to "g" has incompatible type "Tuple[int, int]"; expected Container[int]
```

Once this has been implemented, also simplify the signature of `len()` to not have `tuple` explicitly (`Sized` should be enough for the argument).

msullivan:
The approach here is to generate an extension class while also
populating a dictionary with the appropriate fields for the dataclass
constructor. Once the extension class is created, we swap out the
contents of its `__dict__` with the `__dict__` that `dataclass`
expects, invoke `dataclass`, and then put back all the original stuff.

Fixes #671.

### DIFF
--- a/mypyc/genops.py
+++ b/mypyc/genops.py
@@ -86,7 +86,8 @@ from mypyc.ops_misc import (
     py_call_op, py_call_with_kwargs_op, py_method_call_op,
     fast_isinstance_op, bool_op, new_slice_op,
     type_op, pytype_from_template_op, import_op, get_module_dict_op,
-    ellipsis_op, method_new_op, type_is_op, type_object_op, py_calc_meta_op
+    ellipsis_op, method_new_op, type_is_op, type_object_op, py_calc_meta_op,
+    dataclass_sleight_of_hand,
 )
 from mypyc.ops_exc import (
     raise_exception_op, raise_exception_with_tb_op, reraise_exception_op,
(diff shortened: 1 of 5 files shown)


mypyc/lib-rt/CPy.h
```diff
@@ -1385,6 +1393,59 @@ CPyPickle_GetState(PyObject *obj)
     return NULL;
 }
 
+/* Support for our partial built-in support for dataclasses.
+ *
+ * Take a class we want to make a dataclass, swap in conventional
+ * looking attributes instead of our descriptors, invoke dataclass,
+ * then swap everything back.
```
JukkaL: Add a short explanation of why we do this?

mypyc/genops.py
```diff
@@ -202,8 +203,32 @@ def build_ir(modules: List[MypyFile],
     return result
 
 
+def is_trait_decorator(d: Expression) -> bool:
+    return isinstance(d, NameExpr) and d.fullname == 'mypy_extensions.trait'
```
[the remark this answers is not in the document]
msullivan: ... yup.

mypyc/genops.py
```diff
@@ -1459,6 +1460,12 @@ def dataclass_finalize(
         (which will be descriptors) with whatever they would be in a
         non-extension class, calling dataclass, then switching them back.
 
+        The resulting class is an extension class and instances of it do not
+        does not have a __dict__ (unless something else requires it).
```
JelleZijlstra: Remove "does not"

Issue #7807, opened 2019-10-28
Cannot dump parse tree with mypy.parse


goodmami:
This is either a feature request or a documentation bug.

I found in the [Getting Started Developing](https://github.com/python/mypy/wiki/Getting-Started-Developing) wiki the following bullet point:

> * You can parse a file and dump the parse tree by running mypy/parse.py as a script. This can make it easier to get familiar with the internals.

However that does not seem to be true; `mypy/parse.py` does not contain any code for running it as a script. In fact it used to have such code, but it was removed along with thousands of other lines when an old parser was taken out (see a8c7947974b1b4bd1d205f6b92a936393085df2a). There was no mention of removing the script functionality in the PR (#2977) so I suspect it was an accident. Nobody seems to have missed it enough the past 2.5 years to raise an issue, but I'm just getting acquainted with mypy's internals and would like to have such easy access to the parse tree to help with debugging.

I propose one of the following fixes:

1. Re-add the script functionality to `mypy.parse` (I already have a commit ready: https://github.com/goodmami/mypy/commit/dddabd94762c5f762e3299b318b0d74e5cd57c10)
2. Add it under `scripts/` (e.g., `scripts/dumpast.py`)
3. Update the wiki

If people agree to (1) above, I'll go ahead and create the PR, or I can take care of (2) or (3).

JukkaL:
My preference would be to add the script under `misc/` (and update the wiki).

goodmami:
@JukkaL Sure, I can do that. But what's the difference between the `scripts/` and `misc/` directories?

JukkaL:
`misc/` mostly has things useful for mypy developers, whereas `scripts/` has things for regular users.

goodmami:
Perfect, thanks.

I now have a commit adding `misc/dump-ast.py`: https://github.com/goodmami/mypy/commit/e572eb3ffdec4a5c8309edbacd28e2a2505b2faf

Unless there're other objections I'll submit the PR. Also I won't edit the wiki until the change is accepted, but here's a minimal change.

```diff
- * You can parse a file and dump the parse tree by running mypy/parse.py as a script. This can make it easier to get familiar with the internals.
+ * You can parse a file and dump the parse tree with the `misc/dump-ast.py` script. This can make it easier to get familiar with the internals.
```

goodmami:
Wiki is now updated. I'll close this issue. Thanks for your help!

PR #7820, opened 2019-10-30
Fix 7807

goodmami:
Commit a8c7947 removed the old parser after the new one had taken
over, but the ability to use `mypy.parse` as a script for dumping
parse trees was also removed, perhaps by accident. This commit adds
that functionality back as the `misc/dump-ast.py` script.

This partially resolves issue #7807. What remains is editing the wiki accordingly.

### DIFF
--- a/misc/dump-ast.py
+++ b/misc/dump-ast.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python
+
+"""
+Parse source files and print the abstract syntax trees.
+"""
+
+from typing import Tuple
+import sys
+import argparse
+
+from mypy.errors import CompileError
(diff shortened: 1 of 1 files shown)


dhilst:
@JukkaL  I just came across this. The wiki is outdated yet, I'm willing to update it, is there any review process to edit the wiki? I never contributed to a wiki inside github yet, only PRs.

JukkaL:
@dhilst We don't have a specific review process for the wiki (the wiki has been somewhat neglected). One way would be to create an issue here with the proposed changes. Or maybe you can just edit a wiki page, and create an issue where you link to your changes, so that somebody can check them.

Issue #7806, opened 2019-10-28
Internal Error: maptype.py: AttributeError: attribute 'type_vars' of 'TypeInfo' undefined


sebastian-philipp:
* Are you reporting a bug, or opening a feature request?

Bug

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.

Hints on how to simply create a minimal example are appreciated.

* What is the actual behavior/output?

```
dashboard/plugins/plugin.py: note: In member "get_options" of class "SimplePlugin":
dashboard/plugins/plugin.py:20: error: "SimplePlugin" has no attribute "OPTIONS"
dashboard/plugins/plugin.py:24: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 0.740
Traceback (most recent call last):
  File "mypy/checkexpr.py", line 3603, in accept
  File "mypy/nodes.py", line 1484, in accept
  File "mypy/checkexpr.py", line 1887, in visit_member_expr
  File "mypy/checkexpr.py", line 1899, in analyze_ordinary_member_access
  File "mypy/checkmember.py", line 117, in analyze_member_access
  File "mypy/checkmember.py", line 134, in _analyze_member_access
  File "mypy/checkmember.py", line 210, in analyze_instance_member_access
  File "mypy/checkmember.py", line 359, in analyze_member_var_access
  File "mypy/checkmember.py", line 517, in analyze_var
  File "mypy/maptype.py", line 20, in map_instance_to_supertype
AttributeError: attribute 'type_vars' of 'TypeInfo' undefined
dashboard/plugins/plugin.py:24: : note: use --pdb to drop into pdb
```

* What are the versions of mypy and Python you are using?

0.740

*  Do you see the same issue after installing mypy from Git master?

Not checked yet.

* What are the mypy flags you are using? (For example --strict-optional)

```
[mypy]
strict_optional = True
no_implicit_optional = True
ignore_missing_imports = True
warn_incomplete_stub = True
check_untyped_defs = True
show_error_context = True
```

* Reproducable with

```
git clone --depth 1 --single-branch --branch mypy-mgr_module https://github.com/sebastian-philipp/ceph.git
cd ceph/src 
script/run_mypy.sh
```

JukkaL:
[...]

PR #7828, opened 2019-10-30
ISSUE-7806: Add missing attributes to FakeInfo

vemel:
### Goal

Fix [ISSUE-7806](https://github.com/python/mypy/issues/7806).

Add `fullname` and `type_vars` to `FakeInfo` so `mypy.checkmember` will not crash on member with  `FakeInfo`.

### Changes

- Handle `fullname`, `_fullname`, `type_vars` as regular attributes of `FakeInfo`
- Add `FakeInfo._fullname = 'FakeInfo'` attribute (probably should be `FakeMember`)

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -2619,20 +2619,16 @@ class FakeInfo(TypeInfo):
     # TypeInfo defines a __bool__ method that returns False for FakeInfo
     # so that it can be conveniently tested against in the same way that it
     # would be if things were properly optional.
+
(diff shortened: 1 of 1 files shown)


JukkaL:
As discussed in https://github.com/python/mypy/issues/7806#issuecomment-548182880, the most fundamental issue is the leaking of a `FakeInfo` to type checking, which should never happen.

JukkaL:
Closing as the fix doesn't address the root cause and there has been no activity in a long time.

PR #7829, opened 2019-10-30
Make all names and fullnames into properties

### LINKED ISSUE #7640: Make all names and fullnames in mypy properties
^

msullivan:
SymbolNode and FuncBase had name and fullname as methods while everything else has them as attributes or properties. Turn them into properties.

Fixes #7640.

This was done with `sed -i -e 's/\.name()/.name/g' -e 's/\.fullname()/.fullname/g' mypy/*.py mypy/*/*.py mypyc/*.py mypyc/*/*.py misc/proper_plugin.py test-data/unit/plugins/*.py`.

This is an annoying plugin compatibility break unfortunately but the name/fullname situation is very annoying so it is probably worth it.

Plugins that want to work with old and new versions can do something like:
```
from typing import Union
from mypy.nodes import FuncBase, SymbolNode


def fullname(x: Union[FuncBase, SymbolNode]) -> str:
    fn = x.fullname
    if callable(fn):
        return fn()
    return fn


def name(x: Union[FuncBase, SymbolNode]) -> str:
    fn = x.name
    if callable(fn):
        return fn()
    return fn
```

I expect the odds that this will need to be refreshed before I merge it as approximately 100%.

### DIFF
--- a/misc/proper_plugin.py
+++ b/misc/proper_plugin.py
@@ -51,20 +51,24 @@ def is_special_target(right: ProperType) -> bool:
         if right.type_object().fullname == 'builtins.tuple':
             # Used with Union[Type, Tuple[Type, ...]].
             return True
-        if right.type_object().fullname in ('mypy.types.Type',
-                                              'mypy.types.ProperType',
-                                              'mypy.types.TypeAliasType'):
+        if right.type_object().fullname in (
+            'mypy.types.Type',
+            'mypy.types.ProperType',
+            'mypy.types.TypeAliasType'
+        ):
             # Special case: things like assert isinstance(typ, ProperType) are always OK.
             return True
-        if right.type_object().fullname in ('mypy.types.UnboundType',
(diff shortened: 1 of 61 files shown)


ilevkivskyi:
> SymbolNode and FuncBase had name and fullname as methods while everything else has them as attributes or properties.

This is not very fair comparison, as in my experience `.fullname()` appears in code three-four times more often than `.fullname` (this is also what grep says). I somehow thought the idea was to make the opposite (i.e. make them all methods). Of course properties are more natural and more usable, but this change will break literally every plugin I am aware of.

msullivan:
I wasn't really trying to be fair, I guess :P. The main reason for doing it this way is that it is much nicer (and it seems pretty silly to turn simple attributes on ast nodes into methods).

Many plugins would be broken doing it the other way, too. I'd rather break all the plugins (instead of just *most* of them :P) and end up with code we'd rather maintain in the long run, I think.

PR #7831, opened 2019-10-31
Can understand functools.total_ordering

### LINKED ISSUE #4610: @total_ordering support
Please consider the following code:

```python
from functools import total_ordering
from typing import Any


@total_ordering
class Ord:

    def __eq__(self, other: Any) -> bool:
        return False

    def __lt__(self, other: "Ord") -> bool:
        return False


Ord() <= Ord()
```

mypy 0.560 (using Python 3.6.4, default flags) complains:

`test.py:15: error: Unsupported left operand type for <= ("Ord")`

It would be helpful if mypy would support `@total_ordering`.

AWhetter:
This change adds support for understanding the additional methods that are added by decorating a class with `functools.total_ordering`.

## Limitations
There's still some potential improvements to make to this before merging:

* There is no support for overloaded definitions. `mypy/plugins/functools.py:_analyze_class` looks specifically for `FuncItem`s. I wasn't sure how to get the type of the arguments off of an overload?
* `mypy/plugins/functools.py:_find_other_type`: I don't know if there is an existing function that I can use to find the value of the first argument instead?
* The type of the input argument isn't really correct. For the input argument, it needs to support both what the argument in the implemented comparison supports and what the `__eq__` supports (depending on which comparison method was implemented. For example if `__ge__` was implemented, the argument to `__le__` needs to be compatible with `__ge__` and `__eq__` (https://github.com/python/cpython/blob/bdac32e9fe25fdb97a7172a93aabd1ffead89462/Lib/functools.py#L146).
* The return types also have room for improvement. Currently the generated methods inherit the return type of the implemented comparison method, but this is usually a `Union` of the `__eq__` return type and the return type of the implemented method. Sometimes it's the return type of the `__bool__` of those types. How can I condense the `Union` of the two potential return types if possible (eg how could the code know to use `bool` instead of `Union[bool,bool]`?

Closes #4610


aecay:
Without realizing it, I made #7848 one day later to do exactly the same thing.  Great minds think alike I guess 😝 

I think the code in this PR is a better base to build on than the code in #7848, so once the questions above are answered and anything from my PR that isn't here is migrated over (see below), I think the way forward is to land this one and close mine.

What I see missing from this version of the change is the checking for the existence of the `__eq__` function.  It's not a runtime error to fail to define `__eq__`, because `object` has an implementation that will be used as a fallback, which uses reference equality.  But I can't think of a case where this semantics would make sense with `total_ordering`.  What I chose to do was emit a warning when we detect that there's no definition of `__eq__` in the mro (besides `object.__eq__`).  I wasn't super confident that was TRT to do, but it's definitely better than not looking at `__eq__` at all.

mypy/plugins/functools.py
```diff
@@ -0,0 +1,78 @@
...
+        ctx.api.fail('must define at least one ordering operation: < > <= >=', ctx.reason)
+        return
+
+    root = max(comparison_methods)  # prefer __lt__ to __le__ to __gt__ to __ge__
+    root_method = comparison_methods[root]
+    other_type = _find_other_type(root_method)
+    if isinstance(root_method.type, CallableType):
+        ret_type = root_method.type.ret_type
+    else:
+        ret_type = ctx.api.named_type('__builtins__.bool')
```
aecay: Shouldn't it just be an error if `__lt__` (etc.) is not callable?

mypy/plugins/functools.py
```diff
@@ -0,0 +1,78 @@
...
+    for additional_op in _ORDERING_METHODS - {root}:
+        if additional_op not in comparison_methods:
```
aecay: `for additional_op in _ORDERING_METHODS - set(comparison_methods.keys())` is an alternative.  I guess it's debatable whether it's clearer, but it does avoid a level of indentation...

mypy/plugins/functools.py
```diff
@@ -0,0 +1,78 @@
...
+    """Analyze the class body, its parents, and return the comparison methods found."""
+    # Traverse the MRO and collect ordering methods.
+    comparison_methods = {}  # type: Dict[str, FuncItem]
+    # Skip object because total_ordering does not use methods from object
+    for cls in ctx.cls.info.mro[:-1]:
+        for name in _ORDERING_METHODS:
+            if name in cls.names and name not in comparison_methods:
+                node = cls.names[name].node
+                if isinstance(node, FuncItem):
+                    comparison_methods[name] = node
```
JukkaL: What if the method is implemented using a variable (e.g. `__lt__ = lambda ...`)? More generally, if the dunder method exists as any kind of node in the symbol table (not just `FuncItem`), we probably don't want to generate a method. 

Random idea: use `Optional[FuncItem]` as the dictionary value type. If it's `None`, the method is defined, but it's something unexpected. In this case it's okay to fall back to default types for the signature of a generated method.

test-data/unit/check-functools.test
```diff
@@ -0,0 +1,250 @@
...
+Ord() <= Ord()
+Ord() == Ord()
+Ord() > Ord()
+Ord() >= Ord()
+
+Ord() < 1  # E: Unsupported operand types for < ("Ord" and "int")
+Ord() <= 1  # E: Unsupported operand types for <= ("Ord" and "int")
+Ord() == 1
+Ord() > 1  # E: Unsupported operand types for > ("Ord" and "int")
+Ord() >= 1  # E: Unsupported operand types for >= ("Ord" and "int")
```
JukkaL: Add `reveal_type(Ord() <= Ord())` or similar to verify the type of the operation is as expected. Maybe use a non-`bool` return type in some test case.

test-data/unit/check-functools.test
```diff
@@ -0,0 +1,250 @@
...
+[builtins fixtures/ops.pyi]
+[builtins fixtures/dict.pyi]
+
+[case testTotalOrderingEqLe]
```
JukkaL: The remaining tests are a bit verbose. Maybe simplify some of them and only test cases that are expected to be materially different from other test cases? Another option is to vary the test cases in some aspects. Here are some ideas (some of these could be separate tests as well):

* Omit the function type annotation in some test case (fall back to `Any` types).
* Use a different argument type in some test case.
* Use an unusual return type in some test case, such as `str`.
* Test non-callable `__lt__`.
* Test method defined as a variable (`__lt__ = lambda: ...`, for example).

mypy/plugins/functools.py
```diff
@@ -0,0 +1,78 @@
...
+    comparison_methods = _analyze_class(ctx)
+    if not comparison_methods:
+        ctx.api.fail('must define at least one ordering operation: < > <= >=', ctx.reason)
+        return
+
+    root = max(comparison_methods)  # prefer __lt__ to __le__ to __gt__ to __ge__
+    root_method = comparison_methods[root]
+    other_type = _find_other_type(root_method)
+    if isinstance(root_method.type, CallableType):
+        ret_type = root_method.type.ret_type
```
JukkaL: The generated methods seem to always return `bool` values at runtime. Copying of the return type thus seems incorrect. However, having inconsistent return types is problematic as well. Maybe use `bool` as the return type if the existing methods use `bool`, and otherwise fall back to `Any` to avoid trouble.

mypy/plugins/functools.py
JukkaL: If `__lt__` is not callable, it's okay to either generate an error or to fall back to some default behavior.

mypy/plugins/functools.py
JukkaL: I like this better than using a set operation.

mypy/plugins/functools.py
AWhetter: I've implemented this but it wasn't totally simple. The return type of "bool" is unbound when the plugin sees it so I've had to do some matching that feel a bit hacky. Let me know if you think there's a better way of doing it.

AWhetter:
All the comments are addressed so this should be ready for another round of review.

mypy/plugins/functools.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+_MethodInfo = NamedTuple('_MethodInfo', [('is_static', bool), ('type', CallableType)])
+
+
+def functools_total_ordering_maker_callback(ctx: mypy.plugin.ClassDefContext,
+                                            auto_attribs_default: bool = False) -> None:
+    """Add dunder methods to classes decorated with functools.total_ordering."""
+    if ctx.api.options.python_version < (3, 2):
+        ctx.api.fail('"functools.total_ordering" is not supported in Python 2 or 3.1', ctx.reason)
```
JelleZijlstra, line 29: Or 3.0 I guess. It doesn't really matter since mypy doesn't support any of these early Python 3 versions, so this could just check for Python 2 too.

mypy/plugins/functools.py
```diff
@@ -0,0 +1,105 @@
...
+        # None of the defined comparison methods can be analysed
+        return
+
+    other_type = _find_other_type(root_method)
+    bool_type = ctx.api.named_type('__builtins__.bool')
+    ret_type = bool_type  # type: Type
+    if root_method.type.ret_type != ctx.api.named_type('__builtins__.bool'):
+        proper_ret_type = get_proper_type(root_method.type.ret_type)
+        if not (isinstance(proper_ret_type, UnboundType)
+                and proper_ret_type.name.endswith('bool')):
```
JelleZijlstra, line 52: endswith seems wrong, what if I make a type called "notabool"?

mypy/plugins/functools.py
```diff
@@ -0,0 +1,105 @@
...
+    for additional_op in _ORDERING_METHODS:
+        # Either the method is not implemented
+        # or has an unknown signature that we can now extrapolate.
+        if not comparison_methods.get(additional_op):
+            args = [Argument(Var('other', other_type), other_type, None, ARG_POS)]
+            add_method_to_class(ctx.api, ctx.cls, additional_op, args, ret_type)
+
+
+def _find_other_type(method: _MethodInfo) -> Type:
+    """Find the type of the ``other`` argument in a comparision method."""
```
JelleZijlstra, line 63:
```suggestion
    """Find the type of the ``other`` argument in a comparison method."""
```

JelleZijlstra (approved):
I have a couple of minor comments, but otherwise this looks good to go. Sorry for the long wait.

belm0:
@AWhetter are you available to follow up on the PR?

AWhetter:
Sorry, this fell of my radar. I've made the requested changes.

PR #7832, opened 2019-10-31
Add note about mypy_path on Windows

### LINKED ISSUE #7805: Colon character ':' in Windows path is a separator for mypy_path
How can I avoid the colon in my mypy_path to be considered as a separator ?
Thanks

Djedouas:
Add note about mypy_path on Windows in file config_file.rst.

Fixes #7805

### DIFF
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -172,9 +172,11 @@ section of the command line docs.
     variable.  Useful if you'd like to keep stubs in your repo, along with the config file.
     Multiple paths are always separated with a ``:`` or ``,`` regardless of the platform.
     User home directory and environment variables will be expanded.
-
+    
     This option may only be set in the global section (``[mypy]``).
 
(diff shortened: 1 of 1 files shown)


docs/source/config_file.rst
```diff
@@ -175,6 +175,8 @@ section of the command line docs.
 
     This option may only be set in the global section (``[mypy]``).
 
+    **Note:** On Windows, use UNC paths to avoid using ``:`` (e.g. ``\\127.0.0.1\X$\MyDir`` where ``X`` is the drive letter).
```
ethanhs: Thank you for adding this! I think it may be good to use the RST note marker like this:

```suggestion
.. note::
    On Windows, use UNC paths to avoid using ``:`` (e.g. ``\\127.0.0.1\X$\MyDir`` where ``X`` is the drive letter).
```

Djedouas: I frankly hesitated, but because it's already in a specific section (the description of `mypy_path`) it could be mistaken for a global note and not a specific `mypy_path` note don't you think?
But OK for me if you prefer 👍

ethanhs: Yeah in that case I think it is good as is.

Issue #3194, opened 2017-04-19
We need a better story about Any


ilevkivskyi:
While working on protocols, I noticed that stubs often use ``Any`` where ``object`` will be more appropriate and vice versa. Because of how protocol inference works and how current mypy solver works, this leads to problems since sometimes mypy infers ``Any`` instead of giving an error (maybe we could improve this by "tightening" solver?)

I think the root of the problem is that wee don't have a good story about ``Any``. I believe the rules for interaction between ``Any`` and other things should be like this (where ``C`` is a normal class like ``int`` or ``str``):
```python
(1) join(Any, object) == object
(2) meet(Any, object) == Any
(3) join(Any, C) == C
(4) meet(Any, C) == C
(5) join(Any, NoReturn) == Any
(6) meet(Any, NoReturn) == NoReturn
```
I use ``NoReturn`` for uninhabited type, since it is how it is currently called in PEP 484. The above rules could be summarized in a simple diagram:
```
        object
          |
         Any (if it appears in a meet)
        / | \
... all other types ...
        \ | /
         Any (if it appear in a join)
          |
       NoReturn
```
Currently, mypy behaves in a completely different way, and I believe this is not right, let me explain why. We could focus on rules (3) and (4). Currently, mypy returns ``Any`` in both these cases. Here is why it is not safe:
```python
from typing import Any

x = 'a'
y: Any = 1

lst = [x, y]
lst[0] >> 1  # not detected by mypy, fails at runtime, will be fixed by rule (3)

def f(x: int) -> None:
    x >> 1
def g(x: Any) -> None:
    pass

funs = [f, g]
funs[0]('a')  # not detected by mypy, fails at runtime, will be fixed by rule (4)
```
@JukkaL I will be grateful for your opinion.

gvanrossum:
Hm, I have to think about this. I'm not sure I agree with everything in the first diagram either. I thought joins with Any always gave Any? (IOW why is rule (2) not the same as rule (4)?)

Off-topic there's also something with `y: Any = 1` where I'd like mypy to behave differently: IMO it should infer that y is temporarily an int but allow assigning anything to it later.

ilevkivskyi:
@gvanrossum 
> I thought joins with Any always gave Any?

The reasoning why I don't like this is similar to why we recently prohibited simplifying ``Union[int, Any]`` to ``Any``, this is not safe.

> (IOW why is rule (2) not the same as rule (4)?)

[...]

PR #7838, opened 2019-10-31
[suggest] Support refining existing type annotations

msullivan:


mypy/suggestions.py
```diff
@@ -778,6 +787,25 @@ def count_errors(msgs: List[str]) -> int:
...
+    This won't do perfectly with complex compound types (like
+    callables nested inside), but it does pretty well.
+    """
+
+    # We don't want to ever squash away optionals while doing this, so set
+    # strict optional to be true always
+    with strict_optional_set(True):
+        arg_types = []  # type: List[Type]
+        for i in range(len(t.arg_types)):
+            arg_types.append(meet_types(t.arg_types[i], s.arg_types[i]))
```
ilevkivskyi: I think this may be a bad idea, technically you should use join here (except `join(C, Any)` is `Any`, while I used to think it should be `C`, see https://github.com/python/mypy/issues/3194).

I could imagine many situations where attempting to improve type for an untyped argument this will cause bad types for existing precise types (like dropping items from a union etc).

A potential solution is to add a flag to `join_types` like `precise=True`. This should be pretty easy, but may cause a bit of churn (every call to `join_types()` in `TypeJoinVisitor` should be replaced with `self.join_types()`).

(Also a related note we should fix `mypy.subtypes.is_more_precise()` at some point.)

test-data/unit/fine-grained-suggest.test
```diff
@@ -829,3 +829,40 @@ Command 'suggest' is only valid after a 'check' command (that produces no parse
...
+    return x + y
+
+def eggs(x: int) -> List[Any]:
+    a = [x]
+    return a
+
+def take_l(x: List[Any]) -> Any:
+    return x[0]
+
+
```
ilevkivskyi: I think one new line would be fine.

test-data/unit/fine-grained-suggest.test
```diff
@@ -829,3 +829,40 @@ Command 'suggest' is only valid after a 'check' command (that produces no parse
...
+spam(bar(), 20)
+test = [10, 20]
+take_l(test)
+[builtins fixtures/isinstancelist.pyi]
+[out]
+(int, int) -> int
+(int, int) -> int
+(int) -> foo.List[int]
+(foo.List[int]) -> int
+==
```
ilevkivskyi: I would add few more examples for situations with two arguments where one is precise, and another one imprecise to be sure the precise one isn't "spoiled" (like union items are not dropped).

mypy/suggestions.py
msullivan: Oh that's a good catch. I'm going to need to rework the merging a bunch.

msullivan:
I've rewritten the merge logic completely to not use either meet or join

msullivan:
Oh wow this one that is failing only under mypyc is going to be exciting. That is frankly super rare!

mypy/suggestions.py
```diff
@@ -840,6 +850,103 @@ def count_errors(msgs: List[str]) -> int:
...
+      refine(Dict[int, Any], Dict[Any, int]) = Dict[int, int]
+      refine(Tuple[int, Any], Tuple[Any, int]) = Tuple[int, int]
+
+      refine(Callable[[Any], Any], Callable[[int], int]) = Callable[[int], int]
+      refine(Callable[..., int], Callable[[int, float], Any]) = Callable[[int, float], int]
+
+      refine(Optional[Any], int) = Optional[int]
+      refine(Optional[Any], Optional[int]) = Optional[int]
+      refine(Optional[Any], Union[int, str]) = Optional[Union[int, str]]
+      refine(Optional[List[Any]], List[int]) = List[int]
```
ilevkivskyi:
```suggestion
      refine(Optional[List[Any]], List[int]) = Optional[List[int]]
```

mypy/suggestions.py
```diff
@@ -840,6 +850,103 @@ def count_errors(msgs: List[str]) -> int:
...
+        for rhs in rhs_items:
+            new = refine_type(lhs, rhs)
+            if new != lhs:
+                new_items.append(new)
+                refined = True
+        if not refined:
+            new_items.append(lhs)
+
+    # We don't ever want to drop None while making these things and
+    # make_simplified_union calls join which cases
```
ilevkivskyi: It looks like something is wrong with grammar here.

mypy/suggestions.py
```diff
@@ -840,6 +850,103 @@ def count_errors(msgs: List[str]) -> int:
...
+    ):
+        return t.copy_modified(items=[refine_type(ta, sa) for ta, sa in zip(t.items, s.items)])
+
+    if isinstance(t, CallableType) and isinstance(s, CallableType):
+        return refine_callable(t, s)
+
+    if isinstance(t, UnionType):
+        return refine_union(t, s)
+
+    return t
```
ilevkivskyi: I would add couple special cases (not important however) for `type` and `tuple`. These two are represented by instances, but can be refined by `Type[C]` and `Tuple[X, Y, ...]`.

msullivan:
The failure with mypyc is due to https://github.com/python/mypy/pull/7854

PR #7840, opened 2019-10-31
[suggest] Support a simple system for inferring argument types from uses

msullivan:
This is extremely simple, only taking types from function calls, but
hopefully will be pretty valuable anyway.

### DIFF
--- a/mypy/suggestions.py
+++ b/mypy/suggestions.py
@@ -48,7 +48,7 @@ from mypy.find_sources import SourceFinder, InvalidSourceList
 from mypy.modulefinder import PYTHON_EXTENSIONS
 from mypy.plugin import Plugin, FunctionContext, MethodContext
 from mypy.traverser import TraverserVisitor
-from mypy.checkexpr import has_any_type
+from mypy.checkexpr import has_any_type, map_actuals_to_formals
 
 from mypy.join import join_type_list
 from mypy.sametypes import is_same_type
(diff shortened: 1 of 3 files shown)


mypy/suggestions.py
```diff
@@ -136,6 +137,47 @@ def get_return_types(typemap: Dict[Expression, Type], func: FuncDef) -> List[Typ
...
+    """
+    def __init__(self, func: FuncDef, typemap: Dict[Expression, Type]) -> None:
+        self.typemap = typemap
+        self.arg_types = {
+            arg.variable: [] for arg in func.arguments
+        }  # type: Dict[SymbolNode, List[Type]]
+
+    def visit_func_def(self, o: FuncDef) -> None:
+        # Skip nested functions
+        pass
```
ilevkivskyi: Do we really need to skip nested functions here? I could imagine something like:
```python
def g(x: int) -> int: ...

def f(x):
    def nested():
        g(x)
```
Since you are matching my the `Var` node, this seems safe.

mypy/suggestions.py
```diff
@@ -136,6 +137,47 @@ def get_return_types(typemap: Dict[Expression, Type], func: FuncDef) -> List[Typ
...
+
+        for i, args in enumerate(formal_to_actual):
+            for arg_idx in args:
+                arg = o.args[arg_idx]
+                if isinstance(arg, RefExpr) and arg.node in self.arg_types:
+                    self.arg_types[arg.node].append(typ.arg_types[i])
+
+
+def get_arg_uses(typemap: Dict[Expression, Type], func: FuncDef) -> List[List[Type]]:
+    """Find all the types of arguments that each arg is passed to."""
```
ilevkivskyi: Instead of repeating the class docstring, I would just add a simple example here:
```
For situation like:

    def foo(x: int) -> None: ...
    def bar(x: str) -> None: ...
    def test(x, y):
        foo(x)
        bar(y)

This will return [[int], [str]].
```

mypy/suggestions.py
```diff
@@ -179,7 +221,7 @@ def __init__(self, fgmanager: FineGrainedBuildManager,
         if no_any:
             self.flex_any = 1.0
 
-        self.max_guesses = 16
+        self.max_guesses = 64
```
ilevkivskyi: An unrelated note: maybe we should make this configurable?

test-data/unit/fine-grained-suggest.test
```diff
@@ -287,6 +287,66 @@ def bar() -> None:
...
+    def spam(self, x, y):
+        take_star(x, y)
+
+def method(x):
+    b = B()
+    b.test(x)
+
+def meet(x):
+    take_a(x)
+    take_b(x)
```
ilevkivskyi: Maybe add examples of meet with `Any` and/or optional types?

mypy/suggestions.py
msullivan: Yeah I'll do that in a follow-up soon

mypy/suggestions.py
msullivan: Yeah, good point.

PR #7842, opened 2019-11-01
Fix crashes in astdiff when sorting union items

### LINKED ISSUE #7834: Daemon crash: comparing tuple and None
In encountered a mypy daemon crash (when using `dmypy suggest`, not sure if it's related):

```
  File "mypy/suggestions.py", line 212, in restore_after
  File "mypy/suggestions.py", line 187, in suggest
  File "mypy/suggestions.py", line 355, in get_suggestion
  File "mypy/suggestions.py", line 325, in find_best
  File "mypy/suggestions.py", line 524, in try_type
  File "mypy/server/update.py", line 275, in trigger
  File "mypy/server/update.py", line 804, in propagate_changes_using_dependencies
  File "mypy/server/update.py", line 951, in reprocess_nodes
  File "mypy/server/astdiff.py", line 160, in snapshot_symbol_table
  File "mypy/server/astdiff.py", line 174, in snapshot_definition
  File "mypy/server/astdiff.py", line 231, in snapshot_type
  File "mypy/types.py", line 1061, in accept
  File "mypy/server/astdiff.py", line 302, in visit_callable_type
  File "mypy/server/astdiff.py", line 242, in snapshot_types
  File "mypy/server/astdiff.py", line 231, in snapshot_type
  File "mypy/types.py", line 1681, in accept
  File "mypy/server/astdiff.py", line 325, in visit_union_type
TypeError: '<' not supported between instances of 'tuple' and 'NoneType'
```

It seems possible that type snapshots contain `None` values, which breaks sorting.

I don't have a repro, but the traceback may contain enough hints to fix the issue.

JukkaL:
Previously the snapshots could be unsortable, since the snapshot
tuples could have inconsistent types. This attemps to make the types
in tuples predictable based on the tuple prefix.

More formally, consider these two snapshot tuples:

```
   (x1, ..., xn, xn+1, ...)
   (y1, ..., yn, yn+1, ...)
```

If `(x1, ..., xn)` == `(y1, ..., yn)`, `xn+1` and `yn+1` must be
comparable.

Attempt to fix #7834.

### DIFF
--- a/mypy/server/astdiff.py
+++ b/mypy/server/astdiff.py
@@ -246,6 +246,13 @@ def snapshot_simple_type(typ: Type) -> SnapshotItem:
     return (type(typ).__name__,)
 
 
+def encode_optional_str(s: Optional[str]) -> str:
+    if s is None:
+        return '<None>'
+    else:
+        return s
+
+
 class SnapshotTypeVisitor(TypeVisitor[SnapshotItem]):
     """Creates a read-only, self-contained snapshot of a type object.
 
@@ -256,6 +263,9 @@ class SnapshotTypeVisitor(TypeVisitor[SnapshotItem]):
     - Has no references to mutable or non-primitive objects.
     - Two snapshots represent the same object if and only if they are
       equal.
+    - Results must be sortable. It's important that tuples have
+      consistent types and can't arbitrarily mix str and None values,
+      for example, since they can't be compared.
     """
 
     def visit_unbound_type(self, typ: UnboundType) -> SnapshotItem:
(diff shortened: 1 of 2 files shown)


mypy/server/astdiff.py
```diff
@@ -282,9 +292,9 @@ def visit_deleted_type(self, typ: DeletedType) -> SnapshotItem:
 
     def visit_instance(self, typ: Instance) -> SnapshotItem:
         return ('Instance',
-                typ.type.fullname(),
+                encode_optional_str(typ.type.fullname()),
```
ilevkivskyi: Doesn't this mean there is a bug elsewhere? Maybe we shouldn't mask it (especially if there is no repro)?

JukkaL: I think that this is not a great place to detect the bug, since the traceback won't tell much.

PR #7848, opened 2019-11-01
Add plugin to typecheck functools.total_ordering

aecay:
This PR adds a plugin to typecheck classes decorated with `functools.total_ordering`, as requested in #4610 .

- The plugin validates the preconditions of `@total_ordering` (that the class must define `__eq__` and one other dunder inequality methods)
- Unit test cases are added

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -260,6 +260,11 @@ class SemanticAnalyzerPluginInterface:
         raise NotImplementedError
 
     @abstractmethod
+    def note(self, msg: str, ctx: Context, code: Optional[ErrorCode] = None) -> None:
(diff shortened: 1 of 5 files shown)


msullivan:
I'm going to review this, but also note that it will need a rebase on top of  #7829

ilevkivskyi:
@msullivan Note there is a competing PR that is 1 day older. I am not sure what to do, but it looks like we will need to close one in favor of another, or maybe somehow combine best parts of both. Up to you.

msullivan:
Closing this because the author felt it was a worse base to work from than #7831. (Which I will probably also close.) Feel free to pick it back up though.

PR #7859, opened 2019-11-02
References for config file values

hoefling:
This is another suggestion for adding crossrefs to the docs, a follow-up of #7652 (crossrefs to standard library items) and #7784 (crossrefs for command line options), only now for the config file values.

Since Sphinx doesn't define any config-specific role, I have added the special `confval` directive and role. The `.. confval::` directive accepts two optional info fields, `:type:` and `:default:`. For example,

```
.. confval:: verbosity

    :type: integer
    :default: 0

    Controls how much debug output will be generated.  Higher numbers are more verbose.
```
renders as

image

The `:confval:` role adds a crossref to the config value in a similar fashion as the `:option:` role for command line options. Example:

```
See the :confval:`verbosity` configuration option for more details.
```
renders as
image

If you like this suggestion, I will add another PR that adds crossreferences to the config file values throughout the docs.


docs/source/config_file.rst
```diff
@@ -57,6 +57,7 @@ section names in square brackets and flag settings of the form
   .. _config-precedence:
 
   When options conflict, the precedence order for configuration is:
+
```
hoefling: This newline is not related to this PR's topic, but it fixes a small rendering issue. Right now on current master, the list is rendered like this:

image

This is because the list items are too deeply indented. With a proper separation of the list and the preceding paragraph, it becomes

image

hoefling:
@JukkaL awesome, thanks! I will then prepare (or better to say, blow the dust off of) an another PR that will add crossrefs for config values in the current docs.

Issue #3625, opened 2017-06-28
Metaclass type/instance difficulties


rtpg:
Consider the following code from my hypothetical ORM that has three different ways of building an object

```
from typing import TypeVar, Type

T = TypeVar('T')


class ModelMeta(type):
    def build(cls: Type[T]) -> T:
        return cls()

    def build_other(cls):
        return cls()


class Model(metaclass=ModelMeta):

    @classmethod
    def new(cls: Type[T]) -> T:
       return cls()
```

I'm having trouble defining `build` in a way that type checks. In this example, I get an error on `build` about how `Type[T]` is not a supertype of the class `ModelMeta`. 

`build_other` passes the type checker, but usages don't seem to catch anything (implicit Any perhaps?)

```
class Dog(Model):
    def __init__(self):
        print("Built Dog")


class Cat(Model):
    def __init__(self):
        print("Built Cat")


c: Cat = Dog.new()  # type check fails
c2: Cat = Dog.build()  # type check fails
c3: Cat = Dog.build_other()  # passes
```

Considering class methods work, it _feels_ like metaclass methods should equally be able to "unwrap" the instance type for a class, but there might be subtleties I'm missing here.

My wish list here would be:
  -  have the inverse of `Type` (something like `Instance`) so I could write something like `def build(cls: T) -> Instance[T]`. This would be (I think) a _bit_ clearer.
  - Have an intersection type, with which I could declare something like 
  `def build(cls: (Type[T] & ModelMeta)) -> T`. This would help with a lot of other problems as well. This seems like it would be a useful shortcut to writing subclasses 
  - Have a magic hook in `subtypes.is_subtype` that would let me write a custom subtype check in some types
 
Full context: I'm writing some stubs for Django's ORM, and was hitting issues around the type of `Model.objects.create`. So in my case I end up needing a "class property" (hence doing things through the metaclass), which is why the simple `new()` call strategy hasn't been working for me.

I think this is related to #3438, but I'm having a hard time identifying the link.

elazarg:
Related to #3346 and some other discussions which I can't seem to find right now.

`build_other` is not checked at all, since it is not annotated. Also that `cls()` is not checked in your code, since `Type[T]` is considered instantiable with arbitrary arguments.

I think your first error message does not show up on master. However, It does show up if you constrain your variable "properly":
```python
T = TypeVar('T', bound='Model')
```
[...]

PR #7860, opened 2019-11-02
Allow fancy self-types

ilevkivskyi:
So, lately I was noticing many issues that would be fixed by (partially) moving the check for self-type from definition site to call site.

This morning I found that we actually have such function `check_self_arg()` that is applied at call site, but it is almost not used. After more reading of the code I found that all the patterns for self-types that I wanted to support should either already work, or work with minimal modifications. Finally, I discovered that the root cause of many of the problems is the fact that `bind_self()` uses wrong direction for type inference! All these years it expected actual argument type to be _supertype_ of the formal one.

After fixing this bug, it turned out it was easy to support following patterns for explicit self-types:
* Structured match on generic self-types
* Restricted methods in generic classes (methods that one is allowed to call only for some values or type arguments)
* Methods overloaded on self-type
* (Important case of the above) overloaded `__init__` for generic classes
* Mixin classes (using protocols)
* Private class-level decorators (a bit hacky)
* Precise types for alternative constructors (mostly already worked)

This PR cuts few corners, but it is ready for review (I left some TODOs). Note I also add some docs, I am not sure this is really needed, but probably good to have.

### DIFF
--- a/docs/source/more_types.rst
+++ b/docs/source/more_types.rst
@@ -561,6 +561,125 @@ with ``Union[int, slice]`` and ``Union[T, Sequence]``.
    to returning ``Any`` only if the input arguments also contain ``Any``.
 
 
+.. _advanced_self:
+
+Advanced uses of self-types
+***************************
+
+Normally, mypy doesn't require annotation for first argument of instance and
+class methods. However, they may be needed to have more precise static typing
+for certain programming patterns.
+
+Restricted methods in generic classes
+-------------------------------------
+
+In generic classes some methods may be allowed to call only
+for certain values of type arguments:
+
+.. code-block:: python
+
+   T = TypeVar('T')
+   class Tag(Generic[T]):
+       item: T
+       def uppercase_item(self: C[str]) -> str:
+           return self.item.upper()
+
+   def label(ti: Tag[int], ts: Tag[str]) -> None:
+       ti.uppercase_item()  # E: Invalid self argument "Tag[int]" to attribute function
+                            # "uppercase_item" with type "Callable[[Tag[str]], str]"
+       ts.uppercase_item()  # This is OK
+
+This pattern also allows extraction of items in situations where type
+argument is itself generic:
+
+.. code-block:: python
+
+  T = TypeVar('T')
+  S = TypeVar('S')
+
+   class Node(Generic[T]):
+       def __init__(self, content: T) -> None:
+           self.content = content
+       def first_item(self: Node[Sequence[S]]) -> S:
+           return self.content[0]
+
+   page: Node[List[str]]
+   page.get_first_item()  # OK, type is "str"
+
+   Node(0).get_first_item()  # Error: Invalid self argument "Node[int]" to attribute function
+                             # "first_item" with type "Callable[[Node[Sequence[S]]], S]"
+
+Finally, one can use overloads on self-type to express precise types of
+some tricky methods:
+
+.. code-block:: python
+
+   T = TypeVar('T')
+
+   class Tag(Generic[T]):
+       @overload
+       def export(self: Tag[str]) -> str: ...
+       @overload
+       def export(self, converter: Callable[[T], str]) -> T: ...
+
+       def export(self, converter=None):
+           if isinstance(self.item, str):
+               return self.item
+           return converter(self.item)
+
+Mixin classes
+-------------
+
+Using host class protocol as a self-type in mixin methods allows
+static typing of mixin class patter:
+
+.. code-block:: python
+
+   class Resource(Protocol):
+       def close(self) -> int: ...
+
+   class AtomicClose:
+       def atomic_close(self: Resource) -> int:
+           with Lock():
+               return self.close()
+
+   class File(AtomicClose):
+       def close(self) -> int:
+          ...
+
+   class Bad(AtomicClose):
+       ...
+
+   f: File
+   b: Bad
+   f.atomic_close()  # OK
+   b.atomic_close()  # Error: Invalid self type for "atomic_close"
+
+Precise typing of alternative constructors
+------------------------------------------
+
+Some classes may define alternative constructors. If these
+classes are generic, self-type allows giving them precise signatures:
+
+.. code-block:: python
+
+   T = TypeVar('T')
+   Q = TypeVar('Q', bound=Base[Any])
+
+   class Base(Generic[T]):
+       def __init__(self, item: T) -> None:
+           self.item = item
+       @classmethod
+       def make_pair(cls: Type[Q], item: T) -> Tuple[Q, Q]:
+           return cls(item), cls(item)
+
+   class Sub(Base[T]):
+       ...
+
+   pair = Sub.make_pair('yes')  # Type is "Tuple[Sub[str], Sub[str]]"
+   bad = Sub[int].make_pair('no')  # Error: Argument 1 to "make_pair" of "Base"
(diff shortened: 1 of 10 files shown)


ilevkivskyi:
Couple extra comments:
* I just updated the PR description with a list of issues this will close, I didn't do a thorough search however.
* Potentially this PR can be split in few parts, if needed (I would prefer to keep it single however).
* I am going to push couple more tests that mimic tricky examples in one issue (just to be sure they are fixed). If you think it is too much I can remove them.

ilevkivskyi:
(Also note that we can count https://github.com/python/mypy/issues/3625 as fixed only if remove all checks at the definition site).

mypy/typeops.py
```diff
@@ -130,6 +141,12 @@ def map_type_from_supertype(typ: Type,
     return expand_type_by_instance(typ, inst_type)
 
 
+def instance_or_var(typ: ProperType) -> bool:
+    # TODO: use more principled check for non-trivial self-types.
```
Michael0x2a: Maybe we should handle `typ` being a TypeType in here as well. That would let us clean up the if statement down below a little more.

test-data/unit/check-selftype.test
```diff
@@ -468,16 +468,225 @@ class B(A[Q]):
...
+[case testMixinAllowedWithProtocol]
+from typing import TypeVar
+from typing_extensions import Protocol
+
+class Resource(Protocol):
+    def close(self) -> int: ...
+
+class AtomicClose:
+    def atomic_close(self: Resource) -> int:
+        return self.close()
```
Michael0x2a: What happens if we add a second method to AtomicClose and try calling it inside here?

Do we end up treating the type of `self` as the intersection of AtomicClose and Resource and allow that to type-check? Or do we assume that `self` is just of type Resource and report an error?

test-data/unit/check-selftype.test
```diff
@@ -468,16 +468,225 @@ class B(A[Q]):
...
+        ...
+
+a: Lnk[str, Tuple[str, int]] = Lnk()
+b: Lnk[str, int] = Lnk()
+c: Lnk[int, float] = Lnk()
+
+d: Lnk[str, float] = b >> c  # OK
+e: Lnk[str, Tuple[int, float]] = a >> (b, c)  # OK
+f: Lnk[str, Tuple[float, int]] = a >> (c, b) # E: Unsupported operand types for >> ("Lnk[str, Tuple[str, int]]" and "Tuple[Lnk[int, float], Lnk[str, int]]")
+
```
Michael0x2a: I'm sure this is already handled, but maybe it could be worth checking to make sure we correctly report errors with custom self-types within the actual methods?

For example, maybe we could add a test case where we have a class `Foo[T]` that defines two methods `f1(self: Foo[int]) -> None` and `f2(self: Foo[str]) -> None`. Then, try calling f2 from inside f1 and make sure that reports an error?

test-data/unit/check-selftype.test
ilevkivskyi: > Do we end up treating the type of `self` as the intersection of AtomicClose and Resource and allow that to type-check?

No, I mentioned in one of the issues that I don't really like the implicit intersections. I think if there is an explicit annotation, it should be just it. Later when we add intersections people will be able to put an explicit intersection there. Currently one can just add a minimal bunch of needed methods to the protocol.

docs/source/more_types.rst
```diff
@@ -561,6 +561,125 @@ with ``Union[int, slice]`` and ``Union[T, Sequence]``.
...
+       def export(self, converter=None):
+           if isinstance(self.item, str):
+               return self.item
+           return converter(self.item)
+
+Mixin classes
+-------------
+
+Using host class protocol as a self-type in mixin methods allows
+static typing of mixin class patter:
```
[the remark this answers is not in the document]
ilevkivskyi: Yes, typically people want to add mixins (or some people call these traits) to many classes. The protocol defines the minimal required functionality of the host classes for a given mixin. I will try to be precise here.

mypy/typeops.py
```diff
@@ -89,6 +98,8 @@ def class_callable(init_type: CallableType, info: TypeInfo, type_type: Instance,
         and is_subtype(init_ret_type, default_ret_type, ignore_type_params=True)
     ):
         ret_type = init_ret_type  # type: Type
+    elif orig_self_type is not None:
+        ret_type = orig_self_type
     else:
         ret_type = default_ret_type
```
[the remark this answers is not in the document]
ilevkivskyi: I tried to do this but there appeared some complications mostly caused by the fact that this helpers gets the callable already after `bind_self()` is called. But in principle yes, there shouod be no difference I think.

ilevkivskyi:
> (Also note that we can count https://github.com/python/mypy/issues/3625 as fixed only if remove all checks at the definition site).

I changed my mind about this. With a tiny extra fix this issue will be now also closed by this PR.

docs/source/more_types.rst
```diff
@@ -561,6 +561,140 @@ with ``Union[int, slice]`` and ``Union[T, Sequence]``.
    to returning ``Any`` only if the input arguments also contain ``Any``.
 
 
+.. _advanced_self:
+
+Advanced uses of self-types
```
msullivan: Is it worth documenting the use in `__init__`?

mypy/typeops.py
```diff
@@ -41,37 +41,49 @@ def type_object_type_from_function(signature: FunctionLike,
                                    def_info: TypeInfo,
                                    fallback: Instance,
                                    is_new: bool) -> FunctionLike:
-    # The __init__ method might come from a generic superclass
-    # (init_or_new.info) with type variables that do not map
-    # identically to the type variables of the class being constructed
-    # (info). For example
+    # We first need to record all non-trivial (explicit) self types in __init__,
+    # since they will not be available after we bind them. Note, we use explicit
+    # self-types only in the defining class, similar to __new__ (see below).
```
msullivan: `__new__` can use explicit returns in subclasses, but kind of as a technicality, since it only works using bound type vars, I think.

mypy/typeops.py
```diff
@@ -130,6 +145,18 @@ def map_type_from_supertype(typ: Type,
...
 
+def supported_self_type(typ: ProperType) -> bool:
+    """Is this a supported kind of explicit self-types?
+
+    Currently, this means a X or Type[X], where X is an instance or
+    a type variable with an instance upper bound.
+    """
+    if isinstance(typ, TypeType):
+        return supported_self_type(typ.item)
+    return (isinstance(typ, TypeVarType) or
```
msullivan: This doesn't seem to check that there is an instance upper bound? Should there be?

Also my inclination would be to throw parens around the and clause.

docs/source/more_types.rst
ilevkivskyi: I noticed it is mostly useful for typeshed stubs, but didn't find this patter in user code, I will add a short sentence at the end, where I discuss overloads.

mypy/typeops.py
ilevkivskyi: Yes, I will make the comment more clear.

mypy/typeops.py
ilevkivskyi: > This doesn't seem to check that there is an instance upper bound? Should there be?

I would say it is safe to keep it as is. We can restrict this later if we will find problems.
> Also my inclination would be to throw parens around the and clause.

OK.

Issue #7846, opened 2019-11-01
Unsafe use of a generic class method of a generic class allowed


JukkaL:
Mypy associates class type variables to generic class methods in some contexts, but this is inconsistent, which results in type unsafety. Here is an example that mypy is okay with but that is actually unsafe:

```py
from typing import TypeVar, Generic

T = TypeVar('T')
T2 = TypeVar('T2')

class C(Generic[T]):
    @classmethod
    def f(cls, x: T) -> T:
        return x

    @classmethod
    def g(cls, x: T2) -> T2:
        cls.f(x)  # This is accepted
        return x

class D(C[int]):
    @classmethod
    def f(cls, x: int) -> int:  # This override is accepted
        return x + 1

D.g('')   # Runtime error!
```

If the methods are instance methods, `T2` is not treated as compatible with `T`, however.

ilevkivskyi:
> Mypy associates class type variables to generic class methods in some contexts

This is done to support alternative class constructors (a popular feature request in the past). I think this idea is not unsafe by itself but the implementation likely have some bugs.

I actually remember hitting something like this while working on some class method issue. I don't remember what was exactly the problem, but most likely it is related to what type `cls` is given. I vaguely remember this is because `cls` looks like `C`, not like `C[T]` at the point where the helper adding type variables is called. So that in your example `cls.f` has type `def [T] (T) -> T` instead of `def (T) -> T`.

There is a test case `testGenericClassMethodUnboundOnClassNonMatchingIdNonGeneric`. From it's name it is clear that the method should be non-generic but it doesn't work (see revealed type for `cls.foo`). I wanted to add a TODO here but probably forgot.

JukkaL:
Yeah, the underlying feature looks useful. The unsafety is a bit worrying though, since it recently hid quite a few bad annotations in some code is was annotating. Increasing priority to high since the false negatives can be hard to track down.

ilevkivskyi:
While working on this I discovered there is the same issue for normal constructors:
```python
class C(Generic[T]):
    def __init__(self, item: T) -> None:
        self.item = item
    @classmethod
    def f(cls) -> None:
        cls(1)  # This is in fact unsafe, cls is def (T) -> T, not def [T] (T) -> T

class D(C[str]):
    def __init__(self, item: str) -> None:
        self.item = item + 'no way'

D('hm').f()
```
I can prohibit this to (for the same reason described above), but I feel like people may  be not happy about this. Probably we can test this against internal code bases to estimate fallout.

PR #7862, opened 2019-11-03
Fix class method type variables

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/7846

The fix is mostly straightforward, instead of ad-hoc guessing of type variables from the return type, we pass the them all the way through intermediate helper functions.

The `ctypes` plugin change is because it used the bare representation of types (the one used by `reveal_type()`) for user errors. So after my change arrangement of stars changed, instead of adjusting it, I switched to the proper type formatting.


mypy/checkmember.py
```diff
@@ -240,7 +240,8 @@ def analyze_type_callable_member_access(name: str,
             # This check makes sure that when we encounter an operator, we skip looking up
             # the corresponding method in the current instance to avoid this edge case.
             # See https://github.com/python/mypy/pull/1787 for more info.
-            result = analyze_class_attribute_access(ret_type, name, mx)
+            result = analyze_class_attribute_access(ret_type, name, mx,
+                                                    original_vars=typ.items()[0].variables)
```
JukkaL: I wonder if picking the first item may cause problems with overloaded methods. If so, it would be good to add a comment and maybe create a follow-up issue (unless the fix is trivial).

mypy/checkmember.py
```diff
@@ -649,7 +650,9 @@ def f(self: S) -> T: ...
 def analyze_class_attribute_access(itype: Instance,
                                    name: str,
                                    mx: MemberContext,
-                                   override_info: Optional[TypeInfo] = None) -> Optional[Type]:
+                                   override_info: Optional[TypeInfo] = None,
+                                   original_vars: Optional[List[TypeVarDef]] = None
+                                   ) -> Optional[Type]:
```
JukkaL: Document the new argument.

mypy/checkmember.py
```diff
@@ -799,7 +803,6 @@ class B(A[str]): pass
     component in case if a union (this is used to bind the self-types).
```
JukkaL: Also discuss `original_vars`.

test-data/unit/check-generics.test
```diff
@@ -1928,6 +1928,89 @@ reveal_type(E.make_one)  # N: Revealed type is 'def [S] (x: Tuple[S`1, builtins.
...
+    Q = TypeVar('Q', bound=Base[T])
+
+    def __init__(self, item: T) -> None: ...
+
+    @classmethod
+    def make_pair(cls: Type[Q], item: T) -> Tuple[Q, Q]:
+        if bool():
+            return cls(0), cls(0)  # E: Argument 1 to "Base" has incompatible type "int"; expected "T"
+        return cls(item), cls(item)
+[builtins fixtures/classmethod.pyi]
```
JukkaL: What about testing an overloaded generic class method?

ilevkivskyi: This may interfere with #7926, if yes, then I will add a test in the other PR fixing that issue.

mypy/checkmember.py
ilevkivskyi: I think I use thus because I noticed overloaded constructors have "uniform" type variables, but maybe it is not guaranteed, I will add a test case and a TODO here.

test-data/unit/check-dataclasses.test
```diff
@@ -484,9 +484,12 @@ class A(Generic[T]):
   @classmethod
   def foo(cls) -> None:
       reveal_type(cls)  # N: Revealed type is 'Type[__main__.A[T`1]]'
-      reveal_type(cls(1))  # N: Revealed type is '__main__.A[builtins.int*]'
-      reveal_type(cls('wooooo'))  # N: Revealed type is '__main__.A[builtins.str*]'
+      reveal_type(cls(cls.x))  # N: Revealed type is '__main__.A[T`1]'
```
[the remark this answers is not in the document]
ilevkivskyi: There is exclusion to specifically support a common pattern like this:
```python
class Base(Generic[T]):
    impl_class: Type[T]  # to be defined on subclasses
    @classmethod
    def make(cls) -> T:
        return cls.impl_class()
```
IIUC you want to prohibit this:
```python
class C(Generic[T]):
        self.x = x
    @classmethod
    def meth(cls) -> None:
        cls.x  # This should be banned as ambiguous
```
If yes, then I can try this (although may require a bit of special-casing for dataclasses where the variable appears in class body).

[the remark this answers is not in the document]
JukkaL: Yeah, the first example looks reasonable, while the second one is not. Not sure what's the best way to deal with it.

mypy/checkmember.py
ilevkivskyi: Funny enough, when adding a test for this (that seem to work well as I said) I discovered a bug because dozen lines above we use `ret_type = typ.items()[0].ret_type` and now _this_ is clearly wrong.

test-data/unit/check-dataclasses.test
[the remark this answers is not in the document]
ilevkivskyi: OK, I added the check for this an no tests failed, if people will complain about this it will be easy to revert, the change is very simple.

PR #7865, opened 2019-11-04
Generate an error if trying to use @final with TypedDict

### LINKED ISSUE #7849: Generate an error if trying to use @final with TypedDict
`@final` doesn't work as expected with TypedDicts, but mypy doesn't complain about it:

```py
from typing_extensions import final, TypedDict

@final  # No error here
class C(TypedDict):
    x: float
    y: int
    z: str

class D(C):  # No error here
    pass

```

This is confusing, since users would expect that `@final` prevents structural subtyping, but that's not the case. The simplest fix would be to generate an error if `@final` is used in a TypedDict definition.

Hints: This should be a pretty simple thing to fix.  Look at what happens in `analyze_class` in `mypy/semanal.py`.

Originally reported in #7845.

TH3CHARLie:
Fixes #7849 

This fix modifies function `analyze_class` in `mypy/semanal.py`, if the analyzer identifies a `TypedDict`, it then traverses all decorators and checks if `@final` exists, and if so, generate an error.

The code is very straightforward, but I think it has room for potential improvement in terms of clarity.


mypy/semanal.py
```diff
@@ -1067,6 +1067,12 @@ def analyze_class(self, defn: ClassDef) -> None:
 
         is_typeddict, info = self.typed_dict_analyzer.analyze_typeddict_classdef(defn)
         if is_typeddict:
+            for decorator in defn.decorators:
+                decorator.accept(self)
+                if isinstance(decorator, RefExpr):
+                    if decorator.fullname in ('typing.final',
+                                            'typing_extensions.final'):
```
```suggestion
                                              'typing_extensions.final'):
```

TH3CHARLie:
@ilevkivskyi I add a test case called `testCannotUseFinalDecoratorWithTypedDict`, please take a look

test-data/unit/check-typeddict.test
```diff
@@ -1924,3 +1924,16 @@ v = {bad2: 2}  # E: Extra key 'bad' for TypedDict "Value"
...
+from typing_extensions import final
+
+@final  # E: @final cannot be used with TypedDict
+class DummyTypedDict(TypedDict):
+    int_val: int
+    float_val: float
+    str_val: str
+
+[builtins fixtures/dict.pyi]
+[typing fixtures/typing-full.pyi]
```
ilevkivskyi: Please add a newline at the end of file.

Issue #5206, opened 2018-06-12
type refinement in function call


dfroger:
This is ok for mypy:

```
from typing import Union

def foo(x: Union[int, str]):
    if isinstance(x, int):
        return x + 10
```

but how can I make it work if the `isinstance` test is in a function call?

```
from typing import Union

def check(x: Union[int, str]) -> bool:
    return isinstance(x, int)

def foo(x: Union[int, str]):
    if check(x):
        return x + 10
```

I get:

```
foo.py:8: error: Unsupported operand types for + ("Union[int, str]" and "int")
```

JelleZijlstra:
This could perhaps be done with literal types in the future, something like this:
```
@overload
def check(x: int) -> Literal[True]: ...
@overload
def check(x: str) -> Literal[False]: ...
```

gvanrossum:
That's pretty verbose though..

gwerbin:
I've had this issue recently too. I definitely think Mypy would benefit from some way to refine types by following conditional paths. Unfortunately implementing that kind of logic is well beyond my expertise.

elazarg:
Related: #1203, #2357 #4063, #3062

ilevkivskyi:
Here I copy the relevant link from one of the above issues I just closed https://github.com/Microsoft/TypeScript/issues/1007

dbrgn:
I've found the TypeScript [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types) to be a very useful concept, especially when dealing with structured data coming in over the network (e.g. a JSON API).

This is the TS syntax:

```typescript
function isNumber(x: any): x is number {
    return typeof x === "number";
}
function isFish(pet: Fish | Bird): pet is Fish {
    return (pet as Fish).swim !== undefined;
}
```

That syntax is probably not feasible in Python (I don't think `pet is Fish` can be used as return type annotation, but can't find the definition for `TYPE_COMMENT` in [the grammar](https://docs.python.org/3/reference/grammar.html)).

Maybe a placeholder type (`def is_fish(pet: Union[Fish, Bird]): TypeAssertion`) could be used, which would tell the type checker to "inline" the type hint function when determining the type refinement in an if statement (or another similar control structure)?

Maybe `TypeAssertion[pet=Fish]` could work as well? Depends on the grammar of course.

JelleZijlstra:
Grammatically type annotations are just expressions, so something like this:
```
[...]

Issue #4063, opened 2017-10-05
Narrow types using overload information


elazarg:
mypy can use `NoReturn` to narrow types of variables, thus enabling us to express some kinds of guards, in the sense of #1203 / #2357, without using any dedicated mechanism. For example
```python
from typing import Union, NoReturn, overload

@overload
def assert_int(x: int) -> None: ...
@overload
def assert_int(x: str) -> NoReturn: ...

def assert_int(x):
    assert isinstance(x, int)

y: Union[int, str]
assert_int(y)
reveal_type(y)  # should be: int; currently Union[int, str]
```
The same technique could be used to narrow types in conditionals, using only "always true" / "always false" types (do we already have such things? maybe it requires singleton `True` `False` literal-types). 

Note that unlike simple `isinstance`, this can also be used to correlate two different arguments, test for generic compatibility, etc; it is the responsibility of the programmer to have an actual implementation.

ethanhs:
Another use case brought up by @timabbott in #5088 is for `assertisNotNone`, which, like many methods of `TestCase`, could be useful for type narrowing. I think this could be quite useful, especially once we have literal types.

 @Michael0x2a since you have been working on overloads recently, do you have thoughts on this?

Michael0x2a:
@ethanhs: I agree this feature could be pretty helpful -- I think we might actually be pretty close to implementing this, albeit with some limitations.

For example, we can make @elazarg's original example to work if we use my (pending) PR and modify the example a little bit:

    from typing import Union, NoReturn, overload

    @overload
    def assert_int(x: int) -> int: ...
    @overload
    def assert_int(x: str) -> NoReturn: ...
    def assert_int(x):
        assert isinstance(x, int)
        return x

    y: Union[int, str]
    y = assert_int(y)  # Type is narrowed!
    reveal_type(y)  # Revealed type is 'int'

Basically, this works as an unintended side-effect of the union-math feature: `Union[int, str]` doesn't match either arm of the overload but *does* match the union of the two arms. So, my PR will infer the return type is `Union[int, NoReturn]` which simplifies to `int`, narrowing the type. 

The main caveat is that support for union-math is currently extremely simple which means it's unclear if we can support more sophisticated type guards. We can probably fix this in the future by just implementing better heuristics, though.

[...]

PR #7870, opened 2019-11-04
Custom type narrowing with a special decorator

gantsevdenis:
This is surely not meant to be merged.. but just a small proposal?
I think the `narrow_cast` (or whatever it's called) should be inside `typing` and not `mypy`. 
Having that decorator, type narrowing with user-defined functions is trivial  (cf tests) and looks something like 
```
@narrow_cast(str)  # mypy will narrow type to 'str' if this returns True
def is_fizz_buzz(foo):
    return foo in ['fizz', 'buzz']

def foobar(foo: Union[str, float]):
    if foo in ['fizz', 'buzz']:
        reveal_type(foo)  # N: Revealed type is 'Union[builtins.str, builtins.float]'
    if is_fizz_buzz(foo):
        reveal_type(foo)  # N: Revealed type is 'builtins.str'
```

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3721,7 +3721,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             for name, expr_ in self.ctns:
                 if refers_to_fullname(node.callee, name):
                     expr = node.args[0]
-                    if literal(expr_) == LITERAL_TYPE:
+                    if literal(expr) == LITERAL_TYPE:
                         vartype = type_map[node.args[0]]
                         type = get_isinstance_type(expr_, type_map)
                         break  # name is unique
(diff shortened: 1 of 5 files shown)


ilevkivskyi:
This looks too ad-hoc TBH. I would rather implement support for type-guards using overloads, see https://github.com/python/mypy/issues/4063 and https://github.com/python/mypy/issues/5206. This will not require any new special forms and will be more flexible.

DustinWehr:
@ilevkivskyi so is your concern that a more-general feature that comes later (but how much later?) would make this (seemingly) simple one obsolete? That doesn't seem like a big deal, especially if the documentation for `@narrow_cast` expressed that expectation, or if we later provided a little converter script that converts source code containing `@narrow_cast` to use the more-general mechanism. 

@gantsevdenis if you got this working, I'd be interested enough to help maintain a fork.

PR #7871, opened 2019-11-04
Add a plugin hook for specifying per-module configuration data

msullivan:
This is useful in general for plugins that might have per-module
configuration and we will be able to use it to manage the interaction
between mypyc caches and mypy caches without needing to add custom
mypyc hooks.

### DIFF
--- a/docs/source/extending_mypy.rst
+++ b/docs/source/extending_mypy.rst
@@ -236,11 +236,13 @@ module. It is called before semantic analysis. For example, this can
 be used if a library has dependencies that are dynamically loaded
 based on configuration information.
 
-**get_config_data()** can be used if the plugin has some sort of
+**report_config_data()** can be used if the plugin has some sort of
 per-module configuration that can affect typechecking. In that case,
 when the configuration for a module changes, we want to invalidate
-mypy's cache for that module so that it can be rechecked. This hook,
-which should return data encodable as JSON, allows doing this.
+mypy's cache for that module so that it can be rechecked. This hook
+should be used to report to mypy any relevant configuration data,
+so that mypy knows to recheck the module if the configuration changes.
+The hooks hould return data encodable as JSON.
 
 Notes about the semantic analyzer
 *********************************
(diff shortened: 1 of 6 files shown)


test-data/unit/plugins/config_data.py
```diff
@@ -0,0 +1,18 @@
...
+import json
+
+from typing import Any
+
+from mypy.plugin import Plugin
+
+
+class ConfigDataPlugin(Plugin):
+    def get_config_data(self, id: str, path: str) -> Any:
+        path = os.path.join('tmp/test.json')
```
ilevkivskyi: This looks suspicious.

docs/source/extending_mypy.rst
```diff
@@ -236,6 +236,12 @@ module. It is called before semantic analysis. For example, this can
 be used if a library has dependencies that are dynamically loaded
 based on configuration information.
 
+**get_config_data()** can be used if the plugin has some sort of
+per-module configuration that can affect typechecking. In that case,
+when the configuration for a module changes, we want to invalidate
+mypy's cache for that module so that it can be rechecked. This hook,
+which should return data encodable as JSON, allows doing this.
```
ilevkivskyi: I think this name may be a bit confusing, I would propose something like `report_config_data()`. Maybe also emphasize more that plugins being normal Python modules can read their custom config files (without this method), but since mypy heavily uses caching, the plugin should report whenever cache for given module should be invalidated because a plugin config changed.

test-data/unit/plugins/config_data.py
msullivan: Yeah but a lot about our test framework is suspicious :P

ilevkivskyi: OK, I give up, how does this work on Windows? Is this some kind of `pytest` magic? I would not use it and rather write `os.path.join('tmp', 'test.json')`

ethanhs: Windows allows `/` in paths as a separator. Problems with Windows paths usually happen when you read a path value, like `path.split('/')` (since it likely uses `\`) or comparing output to expected values (again because the paths usually are rendered with `\`).

ilevkivskyi: Interesting, TIL.

PR #7874, opened 2019-11-05
Add an is_check parameter to report_config_data

msullivan:
This allows plugins to implement more elaborate checking where they do some checks
themselves while producing config data.

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -351,6 +351,15 @@ class SemanticAnalyzerPluginInterface:
         raise NotImplementedError
 
 
+# A context for querying for configuration data about a module for
+# cache invalidation purposes.
+ReportConfigContext = NamedTuple(
+    'DynamicClassDefContext', [
+        ('id', str),        # Module name
+        ('path', str),      # Module file path
+        ('is_check', bool)  # Is this invocation for checking whether the config matches
+    ])
+
 # A context for a function hook that infers the return type of a function with
 # a special signature.
 #
@@ -459,7 +468,7 @@ class Plugin(CommonPluginApi):
         assert self._modules is not None
         return lookup_fully_qualified(fullname, self._modules)
 
-    def report_config_data(self, id: str, path: str, is_check: bool) -> Any:
+    def report_config_data(self, ctx: ReportConfigContext) -> Any:
         """Get representation of configuration data for a module.
 
         The data must be encodable as JSON and will be stored in the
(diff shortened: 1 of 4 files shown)


mypy/plugin.py
```diff
@@ -459,16 +459,23 @@ def lookup_fully_qualified(self, fullname: str) -> Optional[SymbolTableNode]:
         assert self._modules is not None
         return lookup_fully_qualified(fullname, self._modules)
 
-    def report_config_data(self, id: str, path: str) -> Any:
+    def report_config_data(self, id: str, path: str, is_check: bool) -> Any:
```
ilevkivskyi: Looking at the call sites maybe it is better to call this argument `context` and pass it two predefined constants like `mypy.plugin.LOAD` and `mypy.plugin.STORE`? I think this would be more user-friendly.

mypy/plugin.py
```diff
@@ -459,16 +459,23 @@ def lookup_fully_qualified(self, fullname: str) -> Optional[SymbolTableNode]:
...
 
+        If is_check is true, then the return of this call will be
+        checked against the cached version. Otherwise the call is
+        being made to determine what to put in the cache. This can be
+        used to allow consulting extra cache files in certain complex
+        situations.
+
         This can be used to incorporate external configuration information
         that might require changes to typechecking.
+
```
ilevkivskyi: I don't think this extra empty line is needed.

mypy/plugin.py
```diff
@@ -351,6 +351,15 @@ def final_iteration(self) -> bool:
         raise NotImplementedError
 
 
+# A context for querying for configuration data about a module for
+# cache invalidation purposes.
+ReportConfigContext = NamedTuple(
+    'DynamicClassDefContext', [
+        ('id', str),        # Module name
+        ('path', str),      # Module file path
+        ('is_check', bool)  # Is this invocation for checking whether the config matches
```
ilevkivskyi: I still think it is not the best name, because it explains for what _we_ need this, while a plugin writer may not be fluent in mypy internals. I would say `on_load` or `on_loading` or similar would be already better. But up to you.

Maybe add a longer comment explaining that this hook is called twice: on loading cache and on storing cache.

msullivan: The fact that we are loading doesn't seem like the most relevant part to me, since the hook isn't doing any loading, while it might do some checking

PR #7876, opened 2019-11-05
Make some stuff in mypy.build take Options instead of Manager

msullivan:
This is a bit cleaner and I want to call some of these functions from
mypyc in a place where we don't have a BuildManager.


msullivan:
Yeah, good call

PR #7878, opened 2019-11-05
Support Negative Int Literal

### LINKED ISSUE #7844: Literal[negative int]
Python 3.7.4
typing_extensions 3.7.4
mypy git tip ( 0.750+dev.78f2df9d46198563106dd32cb4494678e97d71eb):

mypy returns false positives for negative integer literals.
errors inlined for convenience:
```python


a: Literal[-2] = -2  # error: Incompatible types in assignment (expression has type "int", variable has type "Literal[-2]")
b: Literal[-1] = -1  # error: Incompatible types in assignment (expression has type "int", variable has type "Literal[-1]")
c: Literal[0] = 0
d: Literal[1] = 1
e: Literal[2] = 2
```

TH3CHARLie:
This PR aims to support negative int Literal, resolves #7844 

~Currently, the newly added test will broke and show inconsistent output with directly invoking mypy from command line~

Commit 278ec69 shows the inconsistent error output between running mypy from command line and running from tests

Commit 7b3d1e8 now behaves ideally for testcase like #7844. But it breaks current codebase and tests.


test-data/unit/check-literal.test
```diff
@@ -3093,3 +3093,13 @@ def test() -> None:
...
+
+[case testNegativeIntLiteral]
+from typing_extensions import Literal
+
+a: Literal[-2] = -2
+b: Literal[-1] = -1
+c: Literal[0] = 0
+d: Literal[1] = 1
+e: Literal[2] = 2
+[out]
```
ilevkivskyi: You need to add a test fixture here, probably patch `fixtures/float.pyi` by adding `__neq__` there, and use it here.

test-data/unit/check-literal.test
```diff
@@ -3093,3 +3093,13 @@ def test() -> None:
     if x == 2:  # OK
         ...
 [builtins fixtures/bool.pyi]
+
+[case testNegativeIntLiteral]
```
ilevkivskyi: I would also a test like:
```python
ONE: Final = 1
x: Literal[-1] = -ONE
```

mypy/plugins/default.py
```diff
@@ -417,3 +419,17 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
     return ctx.default_return_type
+
+
+def int_neg_callback(ctx: MethodContext) -> Type:
+    """Infer a more precise return type for int.__neg__.
+
+    This is mainly used to infer the return type as LiteralType
+    if the original underlying object is a LiteralType object
+    """
+    if isinstance(ctx.type, Instance) and isinstance(ctx.type.last_known_value, LiteralType):
```
ilevkivskyi:
```suggestion
    if isinstance(ctx.type, Instance) and ctx.type.last_known_value is not None:
```

mypy/plugins/default.py
```diff
@@ -417,3 +419,17 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+        value = ctx.type.last_known_value.value
+        fallback = ctx.type.last_known_value.fallback
+        if isinstance(value, int):
+            return LiteralType(-value, fallback)
```
ilevkivskyi: You should return an `Instance` here with the last known value set to this literal, instead of the literal directly. This is what screws type inference and makes many tests fail.

mypy/plugins/default.py
```diff
@@ -417,3 +419,17 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+    return ctx.default_return_type
```
ilevkivskyi: For consistency, you should also check `if isinstance(ctx.type, LiteralType)` as Michael said in the issue.

mypy/plugins/default.py
TH3CHARLie: I summary the message as below:
- add `type_context` to `CheckerPluginInterface`
- fill the missing logic so that `TypeChecker ` correctly has a `type_context`
- copy logic from `ExpressionChecker.infer_literal_expr_type()` to complete this task
If my understanding is correct, I will start working on it ASAP

ilevkivskyi: Yes.

(about the last bullet, copy the logic to your plugin)

TH3CHARLie: It's my misunderstanding. Now I think it's going to work, please check 32a99a2

mypy/checker.py
```diff
@@ -219,6 +219,7 @@ def __init__(self, errors: Errors, modules: Dict[str, MypyFile], options: Option
         self.msg = MessageBuilder(errors, modules)
         self.plugin = plugin
         self.expr_checker = mypy.checkexpr.ExpressionChecker(self, self.msg, self.plugin)
+        self.type_context = self.expr_checker.type_context
```
ilevkivskyi: Could you please make this a property, to make it clear it is not writeable.

mypy/plugin.py
```diff
@@ -215,6 +215,8 @@ class CheckerPluginInterface:
     msg = None  # type: MessageBuilder
     options = None  # type: Options
     path = None  # type: str
+    # Type context for type inference
+    type_context = None  # type: List[Optional[Type]]
```
ilevkivskyi: Here it should be an abstract property.

mypy/plugins/default.py
```diff
@@ -417,3 +420,30 @@ def int_pow_callback(ctx: MethodContext) -> Type:
...
+                    value=-value,
+                    fallback=ctx.type,
+                    line=ctx.type.line,
+                    column=ctx.type.column,
+                ))
+    elif isinstance(ctx.type, LiteralType):
+        value = ctx.type.value
+        fallback = ctx.type.fallback
+        if isinstance(value, int):
+            return LiteralType(value=-value, fallback=fallback)
```
ilevkivskyi: Could you please also add this test:
```python
ONE: Final = 1
TWO: Final = 2

err_code = -ONE
if bool():
    err_code = -TWO
```

TH3CHARLie: Should I add it as a new test or append to `testNegativeIntLiteralWithFinal` in `check-literal.test` @ilevkivskyi

ilevkivskyi: I think appending to the existing test is better.

TH3CHARLie:
Thanks for the guidance! It’s a great education

PR #7882, opened 2019-11-05
Make sure type inference works for literals against protocols

ilevkivskyi:
This is mostly important for builtin functions like `round()`, `abs()`, etc., that currently don't work well when a literal is passed to it.


ilevkivskyi:
Looks like an AppVeyor bug, I will try closing and re-opening.

ilevkivskyi:
It didn't really help, but at least I can see that one of the builds on AppVeyor passed, I will merge now since this looks low risk.

PR #7885, opened 2019-11-05
Implementing background infrastructure for recursive types: Part 3

ilevkivskyi:
This is the last part of plumbing for recursive types (previous https://github.com/python/mypy/pull/7366 and https://github.com/python/mypy/pull/7330). Here I implement visitors and related functions.

I convinced myself that we need to only be more careful when a recursive type is checked against another recursive one, so I only special-case these. Logic is similar to how protocols behave, because very roughly type alias can be imagined as a protocol with single property:
```python
A = Union[T, Tuple[A[T], ...]]

class A(Protocol[T]):
    @property
    def __base__(self) -> Union[T, Tuple[A[T], ...]]: ...
```
but where `TypeAliasType` plays role of `Instance` and `TypeAlias` plays role of `TypeInfo`.

Next two pull requests will contain some non-trivial implementation logic.


mypy/join.py
```diff
@@ -454,6 +468,27 @@ def object_from_instance(instance: Instance) -> Instance:
     return res
 
 
+def object_or_any_from_type(typ: ProperType) -> ProperType:
```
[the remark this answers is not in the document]
msullivan: Literal is missing and could be done easily.

NoneType, UnboundType, ErasedType, DeletedType don't have an obvious way to do them, though. Not sure which of those matter.

mypy/meet.py
```diff
@@ -21,6 +21,19 @@
 
 def meet_types(s: Type, t: Type) -> ProperType:
     """Return the greatest lower bound of two types."""
+    if (isinstance(s, TypeAliasType) and isinstance(t, TypeAliasType) and
+            s.is_recursive and t.is_recursive):
+        # This case can trigger an infinite recursion, general support for this will be
```
msullivan: Is it worth breaking this logic out into a function? (Not sure.) Similar question join I guess

mypy/nodes.py
```diff
@@ -2735,6 +2738,9 @@ def __init__(self, target: 'mypy.types.Type', fullname: str, line: int, column:
         self.alias_tvars = alias_tvars
         self.no_args = no_args
         self.normalized = normalized
+        self.assuming = []  # type: List[Tuple[TypeAliasType, TypeAliasType]]
```
msullivan: Document these?

Is the reason why these "algorithmic" data structures need to be stored on the object themselves (instead of some external map, I guess) because it would require a context being threaded through the whole of the typeops tangle, basically?

mypy/subtypes.py
```diff
@@ -63,6 +63,33 @@ def is_subtype(left: Type, right: Type,
     between the type arguments (e.g., A and B), taking the variance of the
     type var into account.
     """
+    if (isinstance(left, TypeAliasType) and isinstance(right, TypeAliasType) and
+            left.is_recursive and right.is_recursive):
+        # This case requires special care because it may cause infinite recursion.
```
msullivan: Probably worth documenting this in more detail.

To make sure I understand this, the algorithm is basically "under the assumption that left <: right, check that get_proper_type(left) <: get_proper_type(right)"?

(Might be worth splitting out some of the logic into another function again, but your call)

mypy/type_visitor.py
```diff
@@ -232,6 +233,14 @@ def visit_overloaded(self, t: Overloaded) -> Type:
     def visit_type_type(self, t: TypeType) -> Type:
         return TypeType.make_normalized(t.item.accept(self), line=t.line, column=t.column)
 
+    @abstractmethod
+    def visit_type_alias_type(self, t: TypeAliasType) -> Type:
+        # This method doesn't have a default implementation for type translators,
+        # because type aliases are special: some information is contained in the
+        # TypeAlias node, and we normally don't generate new nodes. Every subclass
+        # must implement this depending on its semantics.
```
msullivan: No reasonable way around this?

mypy/join.py
```diff
@@ -454,6 +468,27 @@ def object_from_instance(instance: Instance) -> Instance:
...
+        return object_from_instance(typ.partial_fallback)
+    elif isinstance(typ, TypeType):
+        return object_or_any_from_type(typ.item)
+    elif isinstance(typ, TypeVarType) and isinstance(typ.upper_bound, ProperType):
+        return object_or_any_from_type(typ.upper_bound)
+    elif isinstance(typ, UnionType):
+        for it in typ.items:
+            o = object_or_any_from_type(it)
+            if isinstance(o, Instance):
+                return object_from_instance(o)
```
JukkaL: This seems to pick one union item arbitrarily. It seems non-deterministic, since union item order shouldn't affect semantics. 

Random idea: Maybe a join of results of all items instead?

ilevkivskyi:
> Would it make sense to at least add some non-data-driven unit tests for recursive types at this point?

Yes, I think it would make sense to add some unit tests.

mypy/type_visitor.py
ilevkivskyi: Unfortunately no. This is kind of a fundamental limitation. Technically we can put something moderately reasonable, but then there is a high chance we will forget to override this when needed in some visitors we will add in future.

ilevkivskyi:
> Yes, I think it would make sense to add some unit tests.

Actually I just discovered I am missing one implementation to be able to run unit tests, I will add it and add some tests tomorrow.

mypy/subtypes.py
ilevkivskyi: > To make sure I understand this, the algorithm is basically "under the assumption that left <: right, check that get_proper_type(left) <: get_proper_type(right)"?

Yes. I will add some docs for this.

mypy/nodes.py
ilevkivskyi: > Document these?

Yes.
> Is the reason why these "algorithmic" data structures need to be stored on the object themselves (instead of some external map, I guess) because it would require a context being threaded through the whole of the typeops tangle, basically?

Yes, the canonic implementation of `is_subtype()` takes three arguments: `left`, `right`, and `assumptions`, but last time I tried to do this, it required adding `assumptions` to like two dozen functions, and consequently every call site. This is also the reason why we have `TypeInfo.assuming` and friends.

mypy/indirection.py
```diff
@@ -20,14 +20,19 @@ class TypeIndirectionVisitor(TypeVisitor[Set[str]]):
...
 
     def find_modules(self, typs: Iterable[types.Type]) -> Set[str]:
+        self.seen.clear()
         return self._visit(typs)
 
     def _visit(self, typ_or_typs: Union[types.Type, Iterable[types.Type]]) -> Set[str]:
         typs = [typ_or_typs] if isinstance(typ_or_typs, types.Type) else typ_or_typs
         output = set()  # type: Set[str]
         for typ in typs:
+            if any(typ is s for s in self.seen):
```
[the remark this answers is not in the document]
ilevkivskyi: Actually I think it is fine to use equality here, I will add unit test to double-check.

ilevkivskyi:
@msullivan @JukkaL Could one of you please take a look again?

Summary of changes:
* Address review comments
* Add some comments and docstrings
* Added basic unit tests (basics, some visitors, subtyping)
* Replaced ad-hoc type alias expansion with proper visitors
* Shifted boundary of `get_proper_type()` around unions and `expand_types()` so that it happens later (previous location caused some conceptual difficulties)
* Moved the shared state to `TypeState` where it belongs

Unfortunately the typeshed test is going to fail, because the type alias expansion clean-up discovered two errors in typeshed (previously `Type[...]` was not visited). Looking at it, it seems to me the check for class variables there is overly strict, we can move it from definition site to use site as we do for class methods. I will make a separate PR, and will merge this one on top.

ilevkivskyi:
> What's next? Hooking it up in the semantic analyzer?

Yes, exactly!

Issue #3215, opened 2017-04-21
Occasional errors on AppVeyor due to writing .mypy_cache


gvanrossum:
Since https://github.com/python/mypy/pull/3133 was merged I see occasional test failures in AppVeyor (i.e. on Windows) with the traceback below. I suspect this is due to some background process on Windows (e.g. a virus checker or another container?) still having the file open, but a simple (?) solution would be to add `--cache-dir=nul` to the mypy invocations on Windows. (Though that will need to be done somewhere in the test framework I suspect.)

For now I'm just going to ignore such failures.

(**UPDATE:** example taken from https://ci.appveyor.com/project/gvanrossum/mypy/build/1.0.1227/job/9q6qgru4f6ig6ar8 )

```
Traceback (most recent call last):
  File "C:\projects\mypy\scripts\mypy", line 6, in <module>
    main(__file__)
  File "C:\projects\mypy\mypy\main.py", line 46, in main
    res = type_check_only(sources, bin_dir, options)
  File "C:\projects\mypy\mypy\main.py", line 93, in type_check_only
    options=options)
  File "C:\projects\mypy\mypy\build.py", line 188, in build
    graph = dispatch(sources, manager)
  File "C:\projects\mypy\mypy\build.py", line 1570, in dispatch
    process_graph(graph, manager)
  File "C:\projects\mypy\mypy\build.py", line 1813, in process_graph
    process_stale_scc(graph, scc, manager)
  File "C:\projects\mypy\mypy\build.py", line 1921, in process_stale_scc
    graph[id].write_cache()
  File "C:\projects\mypy\mypy\build.py", line 1551, in write_cache
    self.manager)
  File "C:\projects\mypy\mypy\build.py", line 904, in write_cache
    os.replace(data_json_tmp, data_json)
PermissionError: [WinError 5] Access is denied: '.mypy_cache\\3.\\enum.data.json.d2782ac765091066' -> '.mypy_cache\\3.6\\enum.data.json'
```

pkch:
This probably has the same root cause as [this](http://bugs.python.org/issue29982); TLDR: Windows file system inherently leads to intermittent race conditions on write access attempts, even without any other processes, antivirus, etc. running.

If this is a risk in production, retrying the `os.replace` is the [right solution](https://mail.python.org/pipermail/python-dev/2013-September/128353.html)

For AppVeyor, the simplest solution seems to be to make this change in `mypy.defaults`:

[...]

PR #7887, opened 2019-11-05
[mypyc] Support incremental compilation

msullivan:
This works by reworking IR generation to proceed a SCC at a time and
writing out caches of serialized IR information so that we can
generated code that calls into a module without compiling the module
in full. A mypy plugin is used to ensure cache validity by checking
that a hash of the metadata matches and that all of the generated source
is present and matches.

Closes mypyc/mypyc#682.

### DIFF
--- a/mypyc/build.py
+++ b/mypyc/build.py
@@ -35,7 +35,7 @@ from mypy.build import BuildSource
 from mypyc.namegen import exported_name
 from mypyc.options import CompilerOptions
 from mypyc.errors import Errors
-from mypyc.common import shared_lib_name
+from mypyc.common import BUILD_DIR, shared_lib_name
 from mypyc.ops import format_modules
 
 from mypyc import emitmodule
@@ -79,7 +79,8 @@ def fail(message: str) -> NoReturn:
 
 
 def get_mypy_config(paths: List[str],
-                    mypy_options: Optional[List[str]]) -> Tuple[List[BuildSource], Options]:
+                    mypy_options: Optional[List[str]],
+                    compiler_options: CompilerOptions) -> Tuple[List[BuildSource], Options]:
     """Construct mypy BuildSources and Options from file and options lists"""
     # It is kind of silly to do this but oh well
     mypy_options = mypy_options or []
@@ -99,8 +100,7 @@ def get_mypy_config(paths: List[str],
     options.show_traceback = True
     # Needed to get types for all AST nodes
     options.export_types = True
-    # TODO: Support incremental checking
-    options.incremental = False
+    options.incremental = compiler_options.separate
     options.preserve_asts = True
 
     for source in sources:
@@ -184,7 +184,7 @@ def generate_c(sources: List[BuildSource],
     # Do the actual work now
     t0 = time.time()
     try:
-        result = emitmodule.parse_and_typecheck(sources, options)
+        result = emitmodule.parse_and_typecheck(sources, options, groups)
     except CompileError as e:
         for line in e.messages:
             print(line)
@@ -283,14 +283,17 @@ def write_file(path: str, contents: str) -> None:
     want to write, skip writing so as to preserve the mtime
     and avoid triggering recompilation.
     """
+    # We encode it ourselves and open the files as binary to avoid windows
+    # newline translation
+    encoded_contents = contents.encode('utf-8')
     try:
-        with open(path, 'r', encoding='utf-8') as f:
-            old_contents = f.read()  # type: Optional[str]
+        with open(path, 'rb') as f:
+            old_contents = f.read()  # type: Optional[bytes]
     except IOError:
         old_contents = None
-    if old_contents != contents:
-        with open(path, 'w', encoding='utf-8') as f:
-            f.write(contents)
+    if old_contents != encoded_contents:
+        with open(path, 'wb') as f:
+            f.write(encoded_contents)
 
         # Fudge the mtime forward because otherwise when two builds happen close
         # together (like in a test) setuptools might not realize the source is newer
@@ -400,8 +403,12 @@ def mypycify(
     """
 
     setup_mypycify_vars()
-    compiler_options = CompilerOptions(strip_asserts=strip_asserts,
-                                       multi_file=multi_file, verbose=verbose)
+    compiler_options = CompilerOptions(
+        strip_asserts=strip_asserts,
+        multi_file=multi_file,
+        verbose=verbose,
+        separate=separate is not False,
+    )
 
     # Create a compiler object so we can make decisions based on what
     # compiler is being used. typeshed is missing some attribues on the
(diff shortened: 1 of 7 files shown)


msullivan:
Excitingly I can not reproduce the windows CI failure on my local windows machine

msullivan:
I think I have a bead on the Windows issues and I think it can be hacked around in the test harness. I don't think it is blocking the review

mypyc/build.py
```diff
@@ -283,14 +283,17 @@ def write_file(path: str, contents: str) -> None:
...
+        with open(path, 'rb') as f:
+            old_contents = f.read()  # type: Optional[bytes]
     except IOError:
         old_contents = None
-    if old_contents != contents:
-        with open(path, 'w', encoding='utf-8') as f:
-            f.write(contents)
+    if old_contents != encoded_contents:
+        with open(path, 'wb') as f:
+            f.write(encoded_contents)
```
ilevkivskyi: I am not sure if this is relevant, but long time ago we had problems on Windows with incremental mode (see for example https://github.com/python/mypy/issues/3215). This is why we have some non-trivial logic in and around `FilesystemMetadataStore.write()`.  Maybe you can use the same here?

msullivan:
I switched to using setuptools in the run tests and that seems to fix it. setuptools changes how it copies files into place with `--inplace` so I think that fixed it.

mypyc/emitmodule.py
```diff
@@ -63,57 +74,122 @@ def __init__(self, declaration: HeaderDeclaration, mark: bool) -> None:
         self.mark = False
 
 
+class MypycPlugin(Plugin):
```
JukkaL: Maybe move this to a separate module, since a mypy plugin seems like its own thing.

mypyc/test/test_run.py
```diff
@@ -99,6 +99,15 @@ def chdir_manager(target: str) -> Iterator[None]:
         os.chdir(dir)
 
 
+def fudge_dir_mtimes(dir: str, delta: int) -> None:
+    for dirpath, _, filenames in os.walk(dir):
+        for name in filenames:
+            # if name.endswith(('.c', '.h')): continue
```
JukkaL: Should this commented-out line be removed?

mypyc/test/test_run.py
```diff
@@ -99,6 +99,15 @@ def chdir_manager(target: str) -> Iterator[None]:
         os.chdir(dir)
 
 
+def fudge_dir_mtimes(dir: str, delta: int) -> None:
```
JukkaL: Maybe move this to `testutil.py`?

mypyc/emitmodule.py
msullivan: The plugin is pretty closely tied into other logic in this module, since what it does is enforce the caching and group settings from the rest of the driver logic, so I would prefer to keep it in the same module as that code. (Splitting it out in the most obvious way would introduce a cycle, though that is avoidable without much pain.)

I do think that it will be worth splitting and maybe renaming `emitmodule` to separate `GroupGenerator` from the rest. My inclination is to do that in a follow-up PR to keep down the diff noise for this diff and to avoid needing to rebase some follow-up code I've already written, though I could do it now too if you think it's important.

mypyc/build.py
```diff
@@ -99,8 +100,7 @@ def get_mypy_config(paths: List[str],
     options.show_traceback = True
     # Needed to get types for all AST nodes
     options.export_types = True
-    # TODO: Support incremental checking
-    options.incremental = False
+    options.incremental = compiler_options.separate
```
JukkaL: Maybe add comment?

mypyc/emitmodule.py
JukkaL: Sounds reasonable.

mypyc/emitmodule.py
```diff
@@ -63,57 +74,122 @@ def __init__(self, declaration: HeaderDeclaration, mark: bool) -> None:
...
+            return self.group_map[id]
+
+        # Load the metadata and IR cache
+        meta_path, _, _ = get_cache_names(id, path, self.options)
+        ir_path = get_ir_cache_name(id, path, self.options)
+        try:
+            meta_json = self.metastore.read(meta_path)
+            ir_json = self.metastore.read(ir_path)
+        except FileNotFoundError:
+            return None
```
JukkaL: Add comment about when this happens (this is not an error condition as such, right?).

mypyc/emitmodule.py
```diff
@@ -63,57 +74,122 @@ def __init__(self, declaration: HeaderDeclaration, mark: bool) -> None:
...
+
+        ir_data = json.loads(ir_json)
+
+        # Check that the IR cache matches the metadata cache
+        if compute_hash(meta_json) != ir_data['meta_hash']:
+            return None
+
+        # Check that all of the source files are present and as expected
```
JukkaL: Maybe elaborate a bit about when they may not be present or as expected (i.e. is this a normal condition or has something been corrupted).

mypyc/emitmodule.py
```diff
@@ -127,23 +203,172 @@ def compile_modules_to_c(
...
             compiler_options.multi_file
         )
-        ctext.append(generator.generate_c_for_modules())
-    return modules, ctext
+        ctext[group_name] = generator.generate_c_for_modules()
+
+    return ctext
+
+
+def get_ir_cache_name(id: str, path: str, options: Options) -> str:
```
JukkaL: Are the contents of the new cache files documented somewhere? If not, it would be good to document them. Also mention what they are used for.

mypyc/emitmodule.py
```diff
@@ -127,23 +203,172 @@ def compile_modules_to_c(
...
+    return get_ir_cache_name(state.id, state.xpath, state.options)
+
+
+def write_cache(
+    modules: ModuleIRs,
+    result: BuildResult,
+    group_map: Dict[str, Optional[str]],
+    ctext: Dict[Optional[str], List[Tuple[str, str]]],
+) -> None:
+    """Write out the cache information for modules."""
```
JukkaL: Explain in some detail what cache information is written, and how this relates to mypy cache information.

mypyc/genops.py
```diff
@@ -2851,11 +2869,14 @@ def visit_bytes_expr(self, expr: BytesExpr) -> Value:
         value = bytes(expr.value, 'utf8').decode('unicode-escape').encode('raw-unicode-escape')
         return self.load_static_bytes(value)
 
+    def is_native_module(self, module: str) -> bool:
+        return module in self.mapper.group_map
```
JukkaL: Add docstring that explains what a native module means?

mypyc/test-data/run-multimodule.test
```diff
@@ -554,4 +554,180 @@ print(y)
...
+[file other_a.py]
+def foo() -> int: return 10
+
+
+[delete build/__native_other_a.c.2]
+
+[file driver.py]
+import native
+
+[rechecked native, other_a]
```
JukkaL: Is there a test case where only the native module needs to be rechecked? What about test cases where a change in a native module causes another native/Python module to become stale?

mypyc/emitmodule.py
msullivan: Will do. Main situation is if the user deleted `build/` without deleting `.mypy_cache`, which is sort of corruption but also reasonable to expect that it will happen.

mypyc/test-data/run-multimodule.test
msullivan: `other` modules are native, so we already test most of those. I'll make a couple that test the interaction with interpreted modules.

Issue #824, opened 2015-08-19
Show closest candidates for misspellings


JukkaL:
Instead of just saying `"foobar" undefined` try to look for closest candidates (using edit distance, for example) and give suggestions. For example: `"get_stats" undefined, did you mean "getstats"?`.

JukkaL:
This would be useful for at least these things:

* Attribute lookups
* Keyword argument names
* Module imports

JukkaL:
There is basic support for attribute lookups in the `hack` branch.

JelleZijlstra:
This is done for attribute lookups, but not for the other things Jukka mentions.

marcodoordash:
#4934 adds more details and use cases to the **Module imports** part of this issue, pasting those here and closing it as a duplicate of this one.

`Mypy can suggest correct spellings for some attributes, but it doesn't seem to work for module attributes:

```
from subprocess import popen  # Module 'subprocess' has no attribute 'popen'

import subprocess
subprocess.popen  # Module has no attribute "popen"

class Foobar:
    def do_it(self): pass
foobar  # Name 'foobar' is not defined

Foobar().doit()  # "Foobar" has no attribute "doit"; maybe "do_it"?
```
It would be better if mypy would suggest corrections for all of the cases in the above example.`

marcodoordash:
looking at this one

Wuisch:
I took a shot at the keyword argument name case. Comments are welcome.

ilevkivskyi:
This is now fully implemented.

PR #7888, opened 2019-11-05
Show closest candidates for misspellings : keyword argument case

Wuisch:
This fixes the following case outlined in #824
```python
def f(other: A) -> None: pass
f(otter: A())
```
`Unexpected keyword argument "otter" for "f"`
is now 
`Unexpected keyword argument "otter" for "f";  did you mean "other"?`

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -573,17 +573,19 @@ class MessageBuilder:
     def unexpected_keyword_argument(self, callee: CallableType, name: str, arg_type: Type,
                                     context: Context) -> None:
         msg = 'Unexpected keyword argument "{}"'.format(name) + for_function(callee)
-        # Suggest intended keyword, look for same type match else check for any match.
-        matches = []
+        # Suggest intended keyword, look for type match else fallback on any match.
         matching_type_args = []
+        not_matching_type_args = []
         for i, kwarg_type in enumerate(callee.arg_types):
             callee_arg_name = callee.arg_names[i]
-            if (callee_arg_name is not None and
-                    (is_same_type(arg_type, kwarg_type) or is_subtype(arg_type, kwarg_type))):
-                matching_type_args.append(callee_arg_name)
-        matches.extend(best_matches(name, matching_type_args))
+            if callee_arg_name is not None and callee.arg_kinds[i] != ARG_STAR:
+                if is_subtype(arg_type, kwarg_type):
+                    matching_type_args.append(callee_arg_name)
+                else:
+                    not_matching_type_args.append(callee_arg_name)
+        matches = best_matches(name, matching_type_args)
         if not matches:
-            matches.extend(best_matches(name, list(filter(None, callee.arg_names))))
+            matches = best_matches(name, not_matching_type_args)
         if matches:
             suggestion = ";  did you mean {}?".format(pretty_or(matches[:3]))
             msg += "{}".format(suggestion)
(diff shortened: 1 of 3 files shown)


mypy/messages.py
```diff
@@ -570,9 +570,23 @@ def too_many_positional_arguments(self, callee: CallableType,
...
+    def unexpected_keyword_argument(self, callee: CallableType, name: str, arg_type: Type,
                                     context: Context) -> None:
         msg = 'Unexpected keyword argument "{}"'.format(name) + for_function(callee)
+        # Suggest intended keyword, look for same type match else check for any match.
+        matches = []
+        matching_type_args = []
+        for i, kwarg_type in enumerate(callee.arg_types):
+            callee_arg_name = callee.arg_names[i]
+            if (callee_arg_name is not None and
+                    (is_same_type(arg_type, kwarg_type) or is_subtype(arg_type, kwarg_type))):
```
ilevkivskyi: You don't need the `is_same_type()` here,`is_subtype(X, X)` returns `True`.

test-data/unit/check-kwargs.test
```diff
@@ -86,6 +86,35 @@ def f(a: 'A') -> None: pass # N: "f" defined here
...
+def f(atter: 'A', btter: 'B', ctter: 'C') -> None: pass # N: "f" defined here
+f(otter=B()) # E: Unexpected keyword argument "otter" for "f";  did you mean "btter" or "atter"?
+class A: pass
+class B(A): pass
+class C: pass
+
+[case testKeywordMisspellingFloatInt]
+def f(atter: float, btter: int) -> None: pass # N: "f" defined here
+x: int = 5
+f(otter=x) # E: Unexpected keyword argument "otter" for "f";  did you mean "btter" or "atter"?
```
ilevkivskyi: Could you please add a test or two involving `*args` to be sure they don't interfere with the logic?

mypy/messages.py
Wuisch: Done

test-data/unit/check-kwargs.test
Wuisch: Added some tests. 
And the *args did interfere with the logic. --> improved logic

mypy/messages.py
```diff
@@ -570,9 +570,25 @@ def too_many_positional_arguments(self, callee: CallableType,
...
+            if callee_arg_name is not None and callee.arg_kinds[i] != ARG_STAR:
+                if is_subtype(arg_type, kwarg_type):
+                    matching_type_args.append(callee_arg_name)
+                else:
+                    not_matching_type_args.append(callee_arg_name)
+        matches = best_matches(name, matching_type_args)
+        if not matches:
+            matches = best_matches(name, not_matching_type_args)
+        if matches:
+            suggestion = ";  did you mean {}?".format(pretty_or(matches[:3]))
```
ilevkivskyi: Could you please leave only one space after semicolon `;`

mypy/messages.py
```diff
@@ -570,9 +570,25 @@ def too_many_positional_arguments(self, callee: CallableType,
...
+            msg += "{}".format(suggestion)
```
ilevkivskyi: Instead of above two lines you can just write:
```python
msg += "; did you mean {}?".format(pretty_or(matches[:3]))
```

PR #7889, opened 2019-11-05
Increment lambda_counter

jag426:
Previously, with multiple lambdas, the second one would be named like `__mypyc_lambda__0_f_obj_0`, since the counter didn't increment. Now, the second one is named like `__mypyc_lambda__1_f_obj`.


msullivan:
Eh whatever

msullivan:
Oh right my request made no sense, also, since it is fundamentally a genops thing. tbh maybe it isn't worth testing though

PR #7891, opened 2019-11-06
Make `dump_stats` under its own flag

msullivan:
I find it irritatingly noisy as part of `--verbose`, especially when
using verbose to debug incremental tests.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -584,9 +584,10 @@ class BuildManager:
         self.processed_targets = []  # type: List[str]
 
     def dump_stats(self) -> None:
-        self.log("Stats:")
-        for key, value in sorted(self.stats_summary().items()):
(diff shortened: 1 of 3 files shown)


mypy/build.py
```diff
@@ -584,9 +584,10 @@ def __init__(self, data_dir: str,
         self.processed_targets = []  # type: List[str]
 
     def dump_stats(self) -> None:
-        self.log("Stats:")
-        for key, value in sorted(self.stats_summary().items()):
-            self.log("{:24}{}".format(key + ":", value))
+        if self.options.dump_build_stats:
+            print("Stats:")
+            for key, value in sorted(self.stats_summary().items()):
+                print("{:24}{}".format(key + ":", value))
```
ilevkivskyi: Just curious, what is the reason to replace `self.log()` with `print()`?

PR #7892, opened 2019-11-06
Fix error when assigning a ternary with an empty collection to a union type variable 

### LINKED ISSUE #7780: Error when assigning a ternary with an empty collection to a union type variable
Hello,

the following code:
```python
from typing import List, Union
import sys
p : Union[int, List[int]] = 1 if len(sys.argv) else []
```

incorrectly generates the following error:

    t.py:3: error: Incompatible types in assignment (expression has type "Union[int, List[<nothing>]]", variable has type "Union[int, List[int]]")

Adding an integer in the list makes the error disappear. I first observed this issue with a dictionary:
```python
from typing import Dict, Union
import sys
p : Union[int, Dict[int, int]] = 1 if len(sys.argv) else {}
```
Error:

    t.py:3: error: Incompatible types in assignment (expression has type "Union[int, Dict[<nothing>, <nothing>]]", variable has type "Union[int, Dict[int, int]]")

It seems `<nothing>` should be matched with any type in type parameters in unions.

TH3CHARLie:
resolves #7780

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3538,7 +3538,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         full_context_else_type = self.analyze_cond_branch(else_map, e.else_expr, context=ctx)
         if not mypy.checker.is_valid_inferred_type(if_type):
             # Analyze the right branch disregarding the left branch.
-            else_type = self.analyze_cond_branch(else_map, e.else_expr, context=ctx)
+            else_type = full_context_else_type
 
             # If it would make a difference, re-analyze the left
             # branch using the right branch's type as context.
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3534,6 +3534,8 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
         if_type = self.analyze_cond_branch(if_map, e.if_expr, context=ctx)
 
+        # Analyze the right branch using full type context and store the type
+        full_context_else_type = self.analyze_cond_branch(else_map, e.else_expr, context=ctx)
         if not mypy.checker.is_valid_inferred_type(if_type):
             # Analyze the right branch disregarding the left branch.
             else_type = self.analyze_cond_branch(else_map, e.else_expr, context=ctx)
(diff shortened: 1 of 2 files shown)


mypy/checkexpr.py
```diff
@@ -3534,6 +3534,8 @@ def visit_conditional_expr(self, e: ConditionalExpr) -> Type:
 
         if_type = self.analyze_cond_branch(if_map, e.if_expr, context=ctx)
 
+        # Analyze the right branch using full type context and store the type
+        full_context_else_type = self.analyze_cond_branch(else_map, e.else_expr, context=ctx)
         if not mypy.checker.is_valid_inferred_type(if_type):
             # Analyze the right branch disregarding the left branch.
             else_type = self.analyze_cond_branch(else_map, e.else_expr, context=ctx)
```
ilevkivskyi: It looks like can replace this with `else_type = full_context_else_type` to avoid making exactly the same call twice.

test-data/unit/check-unions.test
```diff
@@ -1004,3 +1004,9 @@ do_thing_with_enums(boop)  # E: Argument 1 to "do_thing_with_enums" has incompat
                            # N: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance \
                            # N: Consider using "Sequence" instead, which is covariant
 [builtins fixtures/isinstancelist.pyi]
+
+[case testAssignUnionWithTenaryExprWithEmptyCollection]
+from typing import Dict, List, Union
+x: Union[int, List[int]] = 1 if "" else []
+y: Union[int, Dict[int, int]] = 1 if "" else {}
```
ilevkivskyi: Could you please also add the opposite conditional expressions to the test (since the current inference logic seems to be a bit asymmetric):
```
x: Union[int, List[int]] = [] if bool() else 1
y: Union[int, Dict[int, int]] = {} if bool() else 1
```

test-data/unit/check-unions.test
```diff
@@ -1004,3 +1004,9 @@ do_thing_with_enums(boop)  # E: Argument 1 to "do_thing_with_enums" has incompat
                            # N: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance \
                            # N: Consider using "Sequence" instead, which is covariant
 [builtins fixtures/isinstancelist.pyi]
+
+[case testAssignUnionWithTenaryExprWithEmptyCollection]
+from typing import Dict, List, Union
+x: Union[int, List[int]] = 1 if "" else []
```
ilevkivskyi: I think it is safer to use a function call like `bool()` as the condition, so that mypy will not infer the condition as always false if we add such feature in future.

test-data/unit/check-unions.test
TH3CHARLie: Very reasonable to make it symmetric. See 47f9f25

PR #7916, opened 2019-11-09
Added --no-follow-imports-for-stubs to command line arguments

### LINKED ISSUE #7774: Command line options for suppressing errors are not available
`mypy v0.740`

`show_none_errors` and `ignore_errors`, `follow_imports_for_stubs` config file options are not available as command line options, while it seems to be no reason for that.

`show_none_errors` WONT_FIX
`ignore_errors` WONT_FIX
`follow_imports_for_stubs` PR NEEDED

See https://mypy.readthedocs.io/en/latest/config_file.html#suppressing-errors

jolaf:
Fixes #7774.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -151,6 +151,11 @@ imports.
     all modules. For more information on what the other options do,
     see :ref:`Following imports <follow-imports>`.
 
+.. option:: --no-follow-imports-for-stubs
+
+    This flag will have mypy to ignore the ``--follow-imports`` setting
+    for stub (``.pyi``) files.
+
 .. option:: --python-executable EXECUTABLE
 
     This flag will have mypy collect type information from :pep:`561`
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -151,10 +151,17 @@ imports.
     all modules. For more information on what the other options do,
     see :ref:`Following imports <follow-imports>`.
 
-.. option:: --no-follow-imports-for-stubs
+.. option:: --follow-imports-for-stubs
 
-    This flag will have mypy to ignore the ``--follow-imports`` setting
-    for stub (``.pyi``) files.
+    This flag will cause mypy to respect the ``--follow-imports``
+    setting even for stub (``.pyi``) files.
+
+    Used in conjunction with ``--follow-imports=skip``, this can be used
+    to suppress the import of modules from ``typeshed``, replacing them
+    with ``Any``.
+
+    Used in conjunction with ``--follow-imports=error``, this can be used
+    to make any use of a ``typeshed`` module an error.
 
 .. option:: --python-executable EXECUTABLE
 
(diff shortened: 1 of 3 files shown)


docs/source/config_file.rst
```diff
@@ -211,7 +211,7 @@ section of the command line docs.
     match the name of the *imported* module, not the module containing the
     import statement.
 
-``follow_imports_for_stubs`` (bool, default False)
+``follow_imports_for_stubs`` (bool, default True)
```
ilevkivskyi: This is not true. Mypy by default ignores `--follow-import` for stub files (so that they are always `'normal'`).

docs/source/command_line.rst
```diff
@@ -151,6 +151,11 @@ imports.
     all modules. For more information on what the other options do,
     see :ref:`Following imports <follow-imports>`.
 
+.. option:: --no-follow-imports-for-stubs
```
ilevkivskyi: It is better to document the flag without `no-` prefix.

mypy/main.py
```diff
@@ -459,6 +459,10 @@ def add_invertible_flag(flag: str,
     imports_group.add_argument(
         '--follow-imports', choices=['normal', 'silent', 'skip', 'error'],
         default='normal', help="How to treat imports (default normal)")
+    add_invertible_flag(
+        '--no-follow-imports-for-stubs', dest='follow_imports_for_stubs', default=True,
```
ilevkivskyi: Could you please add a test for this? (Probably in `check-flags.test`)

docs/source/config_file.rst
jolaf: I'm amazed but from my persperctive THIS is not true, and by default `mypy` uses `--follow-import` setting for stubs.

jolaf: Oh, now I see what's going on.
`follow_imports_for_stubs=True` was specified in my config file. :(
I'll fix it, surely.

jolaf: Fixed.

docs/source/command_line.rst
jolaf: Fixed.

mypy/main.py
jolaf: Sorry I have no idea how to do it. :(
The tests file looks too cryptic to me.

jolaf: I performed basic manual functional checks, it looks like it works as it should.

ilevkivskyi: > I performed basic manual functional checks, it looks like it works as it should.

I trust you did this, but the test case is still needed. It is not that hard, try making something like `[case testNoImplicitReexport]` in that file.

jolaf: The problem is not adding a section, the problem is finding out the proper contents of it.
There's a ton of tests for ``--follow-imports`` there, so the proper content seems like not something trivial.

Unfortunately, I don't feel that I have the resources to dive into this cryptic test engine.

I didn't expect this PR would get this complex, I thought it would be limited to a simple formal change.

jolaf: We can just leave this issue and PR open until someone qulified finds time to complete it.

Or we can close it as WONTFIX.

hauntsaninja:
Author isn't interested in writing a test, and I don't think there's that much interest in having this in the CLI

Issue #1533, opened 2016-05-15
Can't pass a value of type Union[str, bytes] to a generic function typed with AnyStr


gvanrossum:
Repro:
```
from typing import AnyStr, Union
def f() -> Union[str, bytes]:
    pass
def g(a: AnyStr) -> AnyStr:
    pass
x = f()
y = g(x)
```
We get an error on the last line:
```
c.py:7: error: Type argument 1 of "g" has incompatible value "Union[str, bytes]"
```
I would have expected that it would just deduce that y is also a Union[str, bytes].

JukkaL:
Yeah, it would be nice if the example would work. Generally substituting `Union[str, bytes]` for `AnyStr` would be wrong, though. Instead, the union could be decomposed into the components and these could be checked separately.

The reason why can't substitute a union is a little subtle. First, consider this example:

```
from typing import AnyStr, Union, Sequence

def g(a: Sequence[AnyStr]) -> Sequence[AnyStr]:
    if isinstance(a[0], str):
        for s in a:
            print(s + 'x')

def f(a: Sequence[Union[str, bytes]]):
    g(a)   # should be rejected
```

Now the call should be rejected, since `g` expects a uniform sequence and can fail on a mixed sequence.

Here is another example:

```
from typing import AnyStr, Union, List

def g(a: AnyStr, b: AnyStr) -> None:
    if isinstance(a, str):
        print(a + b)

def f(a: Union[str, bytes], b: Union[str, bytes]):
    g(a, b)  # should be rejected 
```

The latter example is similar but doesn't use generics. Here we could type check the call `g(a, b)` with all 4 possible argument type combinations.

Substituting a union type might be safe if `AnyStr` is only used once in an argument type and it's used as a bare type (not within a generic or a composite type). Not 100% about this, though.

rwbarton:
We do already perform this kind of separate type checking of each case when type checking a method call on a union, so it does seem natural to extend that to function calls as @JukkaL described. Another place this would come up is in calls to overloaded functions.

ilevkivskyi:
Btw note this issue is not specific to type variables with values. This example should be supported as well:
```python
def foo[T: int | str](x: list[T]) -> T: ...
x: list[str] | list[int]
foo(x)
```
[...]

PR #7921, opened 2019-11-11
Various improvements to stubgen

JukkaL:
This improves various things in stubgen. Here are some notable changes:

* Filter out some bad type comments (or comments that look like them) since
  these generate parse errors, and stubgen can't recover from them.
* Remove the initial generated comment from all stubs, since it gets easily
  out of date and seems redundant, especially for packages with hundreds of
  submodules.
* Don't generate a blocking error for inconsistent MROs.
* Make docstring parser more robust.
* Tweak the level of logging. Add `--verbose` and `--quiet` flags.
* Export imported names that aren't references in the module.
* Fix handling of unreachable code.
* Fix signatures of various C extension methods.
* Skip modules that are not useful or can cause trouble, such as vendored 
  modules and tests.
* Preserve `@abstractproperty`.
* Fix clashes between `typing.Any` and a class named `Any` (etc.).
* Translate imports to vendored `six` to refer to normal `six`.
* Add ad-hoc whitelisting of specific definitions that should be exported.

(Even though this is a big PR, most of changes are small and hopefully
this won't be too difficult to review.)


mypy/options.py
```diff
@@ -262,6 +262,8 @@ def __init__(self) -> None:
         self.cache_map = {}  # type: Dict[str, Tuple[str, str]]
         # Don't properly free objects on exit, just kill the current process.
         self.fast_exit = False
+        # Used to transform source code before parsing if not None
+        self.transform_source = None  # type: Optional[Callable[[Any], Any]]
```
ilevkivskyi: Maybe add a TODO to make this more precise when the mypy issue is fixed?

mypy/parse.py
```diff
@@ -18,6 +18,11 @@ def parse(source: Union[str, bytes],
     The python_version (major, minor) option determines the Python syntax variant.
     """
     is_stub_file = fnam.endswith('.pyi')
+    if options.transform_source is not None:
+        if isinstance(source, str):  # Work around mypy issue
+            source = options.transform_source(source)
+        else:
+            source = options.transform_source(source)
```
ilevkivskyi: Is this because of https://github.com/python/mypy/issues/1533? It looks like this should not be needed with the `Any` above.

mypy/test/teststubgen.py
```diff
@@ -319,32 +556,55 @@ def parse_flags(self, program_text: str, extra: List[str]) -> Options:
             flag_list = flags.group(1).split()
         else:
             flag_list = []
-        return parse_options(flag_list + extra)
-
-    def add_file(self, path: str, result: List[str]) -> None:
+        options = parse_options(flag_list + extra)
+        if '--verbose' not in flag_list:
```
ilevkivskyi: Maybe it makes sense to add a test for this flag?

mypy/test/teststubgen.py
```diff
@@ -270,8 +281,228 @@ def test_infer_prop_type_from_docstring(self) -> None:
...
+        assert is_blacklisted_path('foo/vendored/bar.py')
+        assert is_blacklisted_path('foo/vendored/bar/thing.py')
+        assert is_blacklisted_path('foo/six.py')
+
+    def test_is_test_module(self) -> None:
+        assert not is_test_module('foo')
+        assert not is_test_module('foo.bar')
+
+        # The following could be test modules, but we are very conservative and
+        # don't treat them as such since they could plausibly be real modules.
```
ilevkivskyi: :+1:

mypy/stubgen.py
```diff
@@ -91,15 +92,49 @@
...
+# Common ways of naming package containing vendored modules.
+VENDOR_PACKAGES = [
+    'packages',
+    'vendor',
+    'vendored',
+]
+
+# Avoid some file names that are unnecessary or likely to cause trouble (\n for end of path).
+BLACKLIST = [
+    '/six.py\n',  # Likely vendored six; too dynamic for us to handle
```
ilevkivskyi: Maybe make these regexps and use `$` instead of `\n`?

mypy/stubgen.py
```diff
@@ -453,56 +571,71 @@ def visit_func_def(self, o: FuncDef, is_abstract: bool = False) -> None:
...
+            is_abstract = True
+        return is_abstract
+
+    def refers_to_fullname(self, name: str, fullname: str) -> bool:
+        module, short = fullname.rsplit('.', 1)
+        return (self.import_tracker.module_for.get(name) == module and
+                (name == short or
+                 self.import_tracker.reverse_alias.get(name) == short))
+
+    def process_member_expr_decorator(self, expr: MemberExpr, context: Decorator) -> bool:
```
ilevkivskyi: Maybe add a short docstring for this function (and maybe also `process_name_expr_decorator`)?

mypy/stubgen.py
```diff
@@ -1043,8 +1275,6 @@ def generate_stub_from_ast(mod: StubSource,
     if subdir and not os.path.isdir(subdir):
         os.makedirs(subdir)
     with open(target, 'w') as file:
-        if add_header:
-            write_header(file, mod.module, pyversion=pyversion)
```
ilevkivskyi: Maybe we can keep this as opt-in, but make it more informative, like add a creation date, mypy version, etc.?

mypy/parse.py
JukkaL: Yeah, it's caused by that issue, and this is not actually needed.

mypy/stubgen.py
JukkaL: I'll deal with this in a follow-up PR, since the change would cause a merge conflict.

mypy/stubgen.py
JukkaL: TBH, I don't think that the header is useful, since it will likely get stale once the stub is updated.

srittau:
@JukkaL Since `@abstractproperty` is deprecated, would it make sense to split this into `@property @abstractmethod` in stubgen?

JukkaL:
> Since @abstractproperty is deprecated, would it make sense to split this into @property @abstractmethod in stubgen?

A good idea! Can you create an issue to track this? The change itself should be pretty simple.

Issue #6417, opened 2019-02-16
Union[Generic[T], T]] does not work as expected as a parameter in the signature of a generic function


haikuginger:
* Are you reporting a bug, or opening a feature request?

Bug

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.
```python3
from typing import Generic, TypeVar, Union

F = TypeVar("F")

class Z(Generic[F]):
    y: F
    def __init__(self, y: F):
        self.y = y


def q2(x: Union[Z[F], F]) -> F:
    if isinstance(x, Z):
        return x.y
    else:
        return x

def q3(x: Z[F]) -> F:
    return x.y

p = Z('hi')
reveal_type(p) # Z[builtins.str*]
reveal_type(q2)  # def [F] (x: Union[Z[F`-1], F`-1]) -> F`-1
q2(p)  # error: Argument 1 to "q2" has incompatible type "Z[str]"; expected "Z[<nothing>]"
q3(p)  # No error

```
* What is the actual behavior/output?

When creating a generic function that has a signature where one parameter is a union on a generic of a typevar and that typevar, calling with the generic causes an error.

* What is the behavior/output you expect?

`Z[F]` satisfies `Union[Z[F], F]`

* What are the versions of mypy and Python you are using?
  Do you see the same issue after installing mypy from Git master?

Python 3.6.6, using mypy 0.660; same issue happens with master.

* What are the mypy flags you are using? (For example --strict-optional)

No flags

ilevkivskyi:
Yeah, inferring against unions is tricky. We have couple similar issues, but there is no simple solution.

Just to illustrate why it is tricky, in your case `F = str` and `F = Z[str]` are both valid solutions for `F` in that function call. Instead  of guessing which one to take, mypy computes a meet of them (common subtype), which is `<nothing>` (bottom type) in this case.

Adding an upper bound to `F` might be a possible solution, but it looks like mypy checks the bound after computing the meet, which is probably a bug.

So the only option for you is to use overloads:
```python
@overload
def q2(x: Z[F]) -> F: ...
@overload
def q2(x: F) -> F: ...
def q2(x):
    if isinstance(x, Z):
        return x.y
    else:
        return x

reveal_type(q2(p)) # Revealed type is "str"
```
which is a bit verbose but removes the ambiguity I explained above, because for overloads the variants are chosen in the order of appearance.

haikuginger:
[...]

PR #7922, opened 2019-11-11
Tweak constraint inference against unions to exclude more unsatisfiable items

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/6417

Currently constraint inference against unions excludes trivially unsatisfiable items, but a corner case of a plain type variable was missing, so I add one.

(Note: this may help with constraint inference for recursive types.)

### DIFF
--- a/mypy/constraints.py
+++ b/mypy/constraints.py
@@ -174,6 +174,7 @@ def infer_constraints_if_possible(template: Type, actual: Type,
         return None
     if (direction == SUPERTYPE_OF and isinstance(template, TypeVarType) and
             not mypy.subtypes.is_subtype(actual, erase_typevars(template.upper_bound))):
+        # This is not caught by the above branch because of the erase_typevars() call.
         return None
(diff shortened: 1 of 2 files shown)


mypy/constraints.py
```diff
@@ -172,6 +172,10 @@ def infer_constraints_if_possible(template: Type, actual: Type,
     if (direction == SUPERTYPE_OF and
             not mypy.subtypes.is_subtype(actual, erase_typevars(template))):
         return None
+    if (direction == SUPERTYPE_OF and isinstance(template, TypeVarType) and
+            not mypy.subtypes.is_subtype(actual, erase_typevars(template.upper_bound))):
+        # This is not caught by the above branch because of the erase_typevars() call.
```
JukkaL: This comment was initially confusing. It wasn't clear to which `erase_typevars()` call this refers to. Maybe make it more explicit, such as saying that `erase_typevars` would return `Any` for a type variable.

PR #7929, opened 2019-11-12
Fix compiled dmypy on Windows

msullivan:
The issue was that the mechanism for transmitting Options to the daemon
on Windows used snapshot/apply_changes and those don't play well with
read-only attributes when compiled.

I've fixed this in the most expedient way for now but will do some
follow ups to clean up this mechanism and hopefully test this on
Windows in CI.


msullivan:
This would be worth a point release except that the paucity of reports makes me suspect nobody would care? I don't know.

Issue #7863, opened 2019-11-03
abstractmethod + generic method seems to disturb inference results


beezee:
I believe this would qualify as a bug.

When a method takes generic input types that are also present in the return type, and that method is defined as abstract, the revealed type of the method appears correct, but the inferred type of the returned value is not consistent with the inferred method type. This is resolved when the method is not declared abstract.

I have a very broad use case for this code, but I've minimized for the example here. I hope this is clear and please let me know if there's anything else I can add or help with.

```python
from abc import ABC, abstractmethod
from typing import Generic, TypeVar

A = TypeVar('A')
B = TypeVar('B')

class TwoTypes(Generic[A, B]): pass

class MakeTwoSurprise(ABC, Generic[A]):

  @abstractmethod
  def __call__(self, a: B) -> TwoTypes[A, B]: pass

class MakeTwoExpected(Generic[A]):

  def __call__(self, a: B) -> TwoTypes[A, B]: pass


class Test():
  
  def make_two(self, 
                mts: MakeTwoSurprise[A], 
                mte: MakeTwoExpected[A]) -> None:
    reveal_type(mts.__call__) # Revealed type is 'def [B] (a: B`-1) -> repro.TwoTypes[A`-1, B`-1]'
    reveal_type(mts(2)) # Revealed type is 'repro.TwoTypes[builtins.int*, builtins.int*]'
    reveal_type(mte.__call__) # Revealed type is 'def [B] (a: B`2) -> repro.TwoTypes[A`-1, B`2]'
    reveal_type(mte(2)) # Revealed type is 'repro.TwoTypes[A`-1, builtins.int*]'
```

I'd expect both of these implementations to typecheck identically, but more importantly you can see `reveal_type(mts(2))` has inferred `int` for both type arguments in `TwoTypes`, even though the first argument, while still unknown to the implementation of `make_two`, would have previously been determined by a caller of this method, based on the instance of `MakeTwoSurprise` that they had passed.

Any and all feedback/assistance is greatly appreciated.

ilevkivskyi:
Thanks for reporting, this definitely looks like a bug. Unfortunately we have a bunch of code duplication for decorated and non-decorated methods (for historical reasons), so I assume some corner cases are missing in the decorated part.

In particular, ``TwoTypes[A`-1, B`-1]`` is clearly wrong, two type variables should not have same ids. Just to clarify, is this specific to `__call__()` or any abstract methods in your code?

beezee:
[...]

PR #7933, opened 2019-11-12
Fix issue 7863

beezee:
See https://github.com/python/mypy/issues/7863 for background context.

A few small notes.

1- Assignment to `t` on line 529 of checkmember.py was not being referenced til 535. Unless `get_proper_type(expand_type_by_instance(typ, itype))` is mutating `typ` or `itype` then this created ambiguity for me around whether lines 530-534 were going to be impacted by my overall change. Since it appears not to be the case, I figured it's clearer to introduce `t` no earlier than it is used, (which in hindsight with the rest of the change could probably be even later in the function.)

2- Changes to check-selftype.test appear to be a result of calling `bind_self` before `get_proper_type(expand_type_by_instance(signature, itype))` - my understanding is that `bind_self` does not register inference by susbtitution of type var, and so the inference of self types are no longer marked as such.

3- #TODO in my added tests - I'm unable to successfully run when importing ABC (error: module abc has no value ABC,) if anyone has guidance on how I might fix that I'd appreciate it, will gladly update accordingly.

I haven't thoroughly grokked the work in freshen_function_type_vars, I arrived at this solution mostly by stepping through debugger between the diverging code paths in my original issue to isolate when the type became incorrect, and comparing the work being done between the two paths.

I have a few qs that I'd love any insight into.

1- What is the difference between negative and positive generic type identifiers? I had suspected contravariant vs covariant, but then the results of freshen_function_type_vars would seem to contradict (it flips negative to positive in contravariant position.)

2- Why do we track generic type vars solely by the internally assigned numeric identifiers and not by the names assigned by programmer? The original issue came up by conflating `A-1` and `B-1` during inference, but I can't ever think of a case where a programmer would intend `A` to have reference equality to `B`, vs a particular instantiation of `Generic[A, B]` that happens to choose the same concrete type for both arguments.

I hope this is on the right track. Thanks in advance for review, to @ilevkivskyi for the initial guidance, and for any insight offered on the questions above.

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -554,9 +554,9 @@ def analyze_var(name: str,
                 # In `x.f`, when checking `x` against A1 we assume x is compatible with A
                 # and similarly for B1 when checking agains B
                 dispatched_type = meet.meet_types(mx.original_type, itype)
-                functype = check_self_arg(functype, dispatched_type, var.is_classmethod,
-                                          mx.context, name, mx.msg)
                 signature = freshen_function_type_vars(functype)
+                signature = check_self_arg(signature, dispatched_type, var.is_classmethod,
+                                          mx.context, name, mx.msg)
                 signature = bind_self(signature, mx.self_type, var.is_classmethod)
                 expanded_signature = get_proper_type(expand_type_by_instance(signature, itype))
                 if var.is_property:
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -533,6 +533,7 @@ def analyze_var(name: str,
             mx.msg.cant_assign_to_classvar(name, mx.context)
         t = get_proper_type(expand_type_by_instance(typ, itype))
         result = t  # type: Type
+        typ = get_proper_type(typ)
         if var.is_initialized_in_class and isinstance(typ, FunctionLike) and not typ.is_type_obj():
             if mx.is_lvalue:
                 if var.is_property:
@@ -557,12 +558,13 @@ def analyze_var(name: str,
                                           mx.context, name, mx.msg)
                 signature = freshen_function_type_vars(functype)
                 signature = bind_self(signature, mx.self_type, var.is_classmethod)
+                expanded_signature = get_proper_type(expand_type_by_instance(signature, itype))
                 if var.is_property:
                     # A property cannot have an overloaded type => the cast is fine.
-                    assert isinstance(signature, CallableType)
-                    result = signature.ret_type
+                    assert isinstance(expanded_signature, CallableType)
+                    result = expanded_signature.ret_type
                 else:
-                    result = get_proper_type(expand_type_by_instance(signature, itype))
+                    result = expanded_signature
     else:
         if not var.is_ready:
             mx.not_ready_callback(var.name(), mx.context)
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -526,12 +526,12 @@ def analyze_var(name: str,
     if typ:
         if isinstance(typ, PartialType):
             return mx.chk.handle_partial_var_type(typ, mx.is_lvalue, var, mx.context)
-        t = get_proper_type(expand_type_by_instance(typ, itype))
         if mx.is_lvalue and var.is_property and not var.is_settable_property:
             # TODO allow setting attributes in subclass (although it is probably an error)
             mx.msg.read_only_property(name, itype.type, mx.context)
         if mx.is_lvalue and var.is_classvar:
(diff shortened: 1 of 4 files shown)


beezee:
Well I can already see I'm failing to typecheck.

I'm going to have a look later to see if this indicates a better path to resolution or requires some additional changes.

JelleZijlstra:
>  I'm unable to successfully run when importing ABC

This is because the mypy tests run with special, small stubs from the `test-data/unit/lib-stubs/` directory, and the `abc` stub there doesn't include `abc.ABC`, presumably because no previous test has needed it.

beezee:
I thought I could smash this thing to green by adjusting my brittle tests to match the output of travis, and completely missed appveyor, which just highlights the comment I had prepared for once it got there:

At this point it looks like the build passes with the exception of the tests I've added as part of the PR.

The tests currently cannot reliably pass because of the call to `freshen_function_type_vars` I added in this PR, which is key to the bugfix.

Given that this function is already being called in a separate related code path (`analyze_instance_member_access`) - is there something I can do to preserve the value of the new tests without the overfit assertions? Is there precedent for data-driven tests that contain generic types in their output?

Best case, is there a way to dynamically get the next increment for assignment of ids to type arguments, and interpolate the correct value into the comments that make up the assertions in my test?

Any suggestions greatly appreciated.

test-data/unit/check-generics.test
```diff
@@ -2128,3 +2128,50 @@ class B(A):
...
+
+A = TypeVar('A')
+B = TypeVar('B')
+C = TypeVar('C')
+
+class TwoTypes(Generic[A, B]):
+
+  def __call__(self) -> B: pass
+
+# TODO - This should extend ABC but test errors that ABC not in abc on import
```
msullivan: Comment stale

msullivan:
There is not a lot of precedent for type variables like this in tests I think. We could probably hack together something if needed, but how much value would be lost if we just removed that part of the test?

msullivan:
For bound type variables, negative indicates that they were bound by a function and positive that they were bound by a class. For "type inference metavariables" that are used during type inference to solve constraints, they are always positive. Type variables track which of these they are but that is not reflected in the string formatting of them...

We don't use the names because names can be reused in bindings for different functions (or even the same function invoked multiple times in an expression with different type parameters)

mypy/checkmember.py
```diff
@@ -555,13 +556,15 @@ def analyze_var(name: str,
                 dispatched_type = meet.meet_types(mx.original_type, itype)
                 functype = check_self_arg(functype, dispatched_type, var.is_classmethod,
                                           mx.context, name, mx.msg)
-                signature = bind_self(functype, mx.self_type, var.is_classmethod)
+                signature = freshen_function_type_vars(functype)
```
msullivan: I think we should freshen the function type vars before the check_self_arg, to match how it is done elsewhere

test-data/unit/check-generics.test
beezee: Removed thanks.

mypy/checkmember.py
beezee: Done, thanks for the pointer.

beezee:
I appreciate your q around the tests. I adjusted them to demonstrate the independence of type var bound at class level vs bound at method level, and I think this reasonably covers the focus of the PR against regression, if not less generally.

Re pos/neg type vars - is that association something that freshen_function_type_vars would/should respect, or does that all fall under inference metavariables?

Re using names - it sounds like the identifiers are meant to subsume tracking of lexical scope? I'm sure the scope of a change at that level would be huge, but I can't help wondering if assigning prefixes at certain levels of scope would produce just enough disambiguation without needing to discard the programmers input. To that end, does the internal id management break down if it is used _in addition_ to the names assigned by programmer? I think if the latter approach were taken, we'd still have binding distinction _at least_ at the granularity we have now, and then some. The additional distinction would have prevented the original issue in this PR, namely that `A-1` is considered the same as `B-1`, regardless what concrete types are inhabiting those parameters.

Sorry for the side thread, mostly trying to catch myself up with the underlying design and implementation. I'm having a lot of fun pushing what I can get mypy to prove, and hoping in the process I'll be able to continue contributing back more as I find dark corners.

Thanks for the review, please lmk if there's anything else I can improve here to get merged.

msullivan:
All the variables freshen_function_type_vars inserts are inference metavariables.

Using the names in addition to the IDs could have prevented this bug but would have allowed other similar bugs, I think. It would still have been wrong to not replace the variables with inference metavariables.

PR #7946, opened 2019-11-13
Ignore invalid self types in constructors

ilevkivskyi:
This previously caused small troubles in Dropbox internal codebase C (because of weird annotations).

For example if an `__init__()` of class `C` had its `self` argument annotated with one of its superclasses `B`, this made all calls to `C()` inferred as `B`, causing spurious errors.

This PR also makes explicit types in `__new__()` and `__init__()` treated in a more similar way.


ilevkivskyi:
> Maybe also make the commit message explain the issue this fixes in more detail, and replace 'C' with "Dropbox internal codebase C" or something.

👍

PR #7951, opened 2019-11-14
More improvements to stubgen

JukkaL:
This includes many improvements to stubgen and a related mypy fix. 
Here are the most useful ones:

* Use a separate process to do introspection of modules so that we can
  recover if a module kills the current process on import, for example.
* Export all names imported from the current package by default.
 Add `--export-less` stubgen flag to disable this behavior.
* Avoid a crash in semantic analysis if there's a bad property definition
  (stubgen can generate these).
* Fix various issues with bad Python code being generated by stubgen.
* Ignore bad signatures in docstrings (this is still very ad-hoc, but it's a
  bit more robust now).
* Try to find a module using `sys.path` if we can't import it.
* Skip some additional modules that may be runnable since they can cause
  trouble when we try to introspect them.

This is again a big PR, but the commit history should be reasonably clean.


mypy/moduleinspect.py
```diff
@@ -0,0 +1,160 @@
...
+
+        Return the value read from the queue, or None if the process unexpectedly died.
+        """
+        max_iter = 100
+        n = 0
+        while True:
+            if n == max_iter:
+                raise RuntimeError('Timeout waiting for subprocess')
+            try:
+                return self.q2.get(timeout=0.05)
```
ilevkivskyi: Maybe make the timeout configurable in constructor?

mypy/semanal.py
```diff
@@ -865,19 +865,26 @@ def analyze_property_with_multi_part_definition(self, defn: OverloadedFuncDef) -
...
+                            first_item.var.is_settable_property = True
+                            # Get abstractness from the original definition.
+                            item.func.is_abstract = first_item.func.is_abstract
+                else:
+                    self.fail("Decorated property not supported", item)
                 item.func.accept(self)
+            else:
+                self.fail('Unexpected definition for property "{}"'.format(first_item.func.name),
+                          item)
+                deleted_items.append(i + 1)
```
ilevkivskyi: What if the extra item is also decorated (with some unrelated decorator)? Maybe add a test for this?

mypy/stubdoc.py
```diff
@@ -129,6 +139,10 @@ def add_token(self, token: tokenize.TokenInfo) -> None:
                 self.state.pop()
             elif self.state[-1] == STATE_ARGUMENT_LIST:
                 self.arg_name = self.accumulator
+                if not re.match(r'\**[A-Za-z_][A-Za-z0-9_]*$', self.arg_name):
```
ilevkivskyi: Maybe also move this regex to the top of file and make it compiled?

mypy/stubgen.py
```diff
@@ -90,28 +90,63 @@
...
     'packages',
     'vendor',
     'vendored',
+    '_vendor',
+    '_vendored_packages',
 ]
 
 # Avoid some file names that are unnecessary or likely to cause trouble (\n for end of path).
 BLACKLIST = [
     '/six.py\n',  # Likely vendored six; too dynamic for us to handle
```
ilevkivskyi: Replace `\n` with `$`?

mypy/stubgen.py
```diff
@@ -90,28 +90,63 @@
...
+METHODS_WITH_RETURN_VALUE = {
+    '__ne__',
+    '__eq__',
+    '__lt__',
+    '__le__',
+    '__gt__',
+    '__ge__',
+    '__hash__',
+    '__iter__',
 }
```
ilevkivskyi: Not really important but I think all these top-level constants can be made `Final`.

mypy/stubgen.py
```diff
@@ -863,24 +940,18 @@ def visit_import_from(self, o: ImportFrom) -> None:
...
-            # Include import from targets that import from a submodule of a package.
-            if relative:
-                sub_names = [name for name, alias in o.names
-                             if alias is None]
-                exported_names.update(sub_names)
-                if module:
-                    for name in sub_names:
-                        self.import_tracker.require_name(name)
 
     def translate_module_name(self, module: str, relative: int) -> Tuple[str, int]:
```
ilevkivskyi: This technically doesn't need to be a method, but up to you.

mypy/stubgen.py
```diff
@@ -1041,6 +1105,21 @@ def is_recorded_name(self, name: str) -> bool:
         return self.is_top_level() and name in self._toplevel_names
 
 
+def find_method_names(defs: List[Statement]) -> Set[str]:
+    # TODO: Traverse into nested definitions
+    # TODO: Overloads
```
ilevkivskyi: There is an `elif` for `OverloadedFuncDef` is this TODO outdated?

mypy/stubgen.py
```diff
@@ -1385,6 +1481,9 @@ def parse_options(args: List[str]) -> Options:
     parser.add_argument('--include-private', action='store_true',
                         help="generate stubs for objects and members considered private "
                              "(single leading underscore and no trailing underscores)")
+    parser.add_argument('--export-less', action='store_true',
+                        help=("don't implicitly export all names imported from other modules "
+                              "in the same package"))
```
ilevkivskyi: Maybe update the docs with the new flag? (In a separate PR if you prefer.)

test-data/unit/stubgen.test
```diff
@@ -1248,6 +1262,24 @@ class F:
 @t.coroutine
 def g(): ...
 
+[case testCoroutineSpecialCase_import]
+import asyncio
+
+__all__ = ['C']
+
+@asyncio.coroutine
+def f():
```
ilevkivskyi: Why don't we include `f()` in the output?

ilevkivskyi:
(I think you noticed, but tests are still failing.)

test-data/unit/stubgen.test
JukkaL: It's not included in `__all__`.

JukkaL: (This is something we may want to change, but it may require the generation of an explicit `__all__`.)

mypy/stubgen.py
JukkaL: A good idea. I'll create a separate PR.

mypy/moduleinspect.py
JukkaL: I don't think that making it configurable is useful.

PR #7953, opened 2019-11-14
Fix typechecking on Python 3.8

msullivan:
The Python 3.8 ast stubs now (correctly!) indicate that a key in Dict
can be None. Fix fastparse to deal with this. Really the typed_ast stubs
should say this also.

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -321,12 +321,15 @@ class ASTConverter:
         node.end_line = getattr(n, "end_lineno", None) if isinstance(n, ast3.expr) else None
         return node
 
-    def translate_expr_list(self, l: Sequence[AST]) -> List[Expression]:
-        res = []  # type: List[Expression]
+    def translate_opt_expr_list(self, l: Sequence[Optional[AST]]) -> List[Optional[Expression]]:
+        res = []  # type: List[Optional[Expression]]
         for e in l:
             exp = self.visit(e)
             res.append(exp)
         return res
+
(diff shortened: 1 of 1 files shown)


mypy/fastparse.py
```diff
@@ -321,13 +321,16 @@ def set_line(self, node: N, n: Union[ast3.expr, ast3.stmt]) -> N:
...
-    def translate_expr_list(self, l: Sequence[AST]) -> List[Expression]:
-        res = []  # type: List[Expression]
+    def translate_opt_expr_list(self, l: Sequence[Optional[AST]]) -> List[Optional[Expression]]:
+        res = []  # type: List[Optional[Expression]]
         for e in l:
             exp = self.visit(e)
             res.append(exp)
         return res
 
+    def translate_expr_list(self, l: Sequence[AST]) -> List[Expression]:
```
ilevkivskyi: Maybe you can use an overload instead?
```python
@overload
def translate_expr_list(l: Sequence[AST]) -> List[Expression]: ...
@overload
def translate_expr_list(l: Sequence[Optional[AST]]) -> List[Optional[Expression]]: ...
```

msullivan: I tried that first and it complained that they overlapped

ilevkivskyi: Hm, I think I remember having similar troubles with `get_proper_types()`. @Michael0x2a is this actually unsafe or maybe we can relax this somehow (I mean invariant generics in return types)?

PR #7955, opened 2019-11-15
Make revealed type of Final vars distinct from non-Final vars

Michael0x2a:
This diff changes how we format Instances with a last known value when displaying them with `reveal_type`.

Previously, we would always ignore the `last_known_value` field:

```python
x: Final = 3
reveal_type(x)  # N: Revealed type is 'builtins.int'
```

Now, we format it like `Literal[3]?`. Note that we use the question mark suffix as a way of distinguishing the type from true Literal types.

```python
x: Final = 3
y: Literal[3] = 3
reveal_type(x)  # N: Revealed type is 'Literal[3]?'
reveal_type(y)  # N: Revealed type is 'Literal[3]'
```

While making this change and auditing our tests, I also discovered we were accidentally copying over the `last_known_value` in a few places by accident. For example:

```python

a = []
a.append(1)
a.append(2)     # Got no error here?
reveal_type(a)  # Incorrect revealed type: got builtins.list[Literal[1]?]

b = [0, None]
b.append(1)     # Got no error here?
reveal_type(b)  # Incorrect revealed type: got builtins.list[Union[Literal[0]?, None]]
```

The other code changes I made were largely cosmetic.

Similarly, most of the remaining test changes were just due to places where we were doing something like `reveal_type(0)` or `reveal_type(SomeEnum.BLAH)`.

The main motivation behind this diff is that once this lands, it should become much simpler for me to write some tests I'll need while revamping https://github.com/python/mypy/pull/7169. It also helps make a somewhat confusing and implicit part of mypy internals more visible.

### DIFF
--- a/docs/source/literal_types.rst
+++ b/docs/source/literal_types.rst
@@ -125,9 +125,12 @@ you can instead change the variable to be ``Final`` (see :ref:`final_attrs`):
     expects_literal(c)      # ...and this type checks!
 
 If you do not provide an explicit type in the ``Final``, the type of ``c`` becomes
-context-sensitive: mypy will basically try "substituting" the original assigned
-value whenever it's used before performing type checking. So, mypy will type-check
-the above program almost as if it were written like so:
+*context-sensitive*: mypy will basically try "substituting" the original assigned
+value whenever it's used before performing type checking. This is why the revealed
+type of ``c`` is ``Literal[19]?``: the question mark at the end reflects this
+context-sensitive nature.
+
+For example, mypy will type check the above program almost as if it were written like so:
 
 .. code-block:: python
 
@@ -138,9 +141,13 @@ the above program almost as if it were written like so:
     reveal_type(19)
     expects_literal(19)
 
-In other words, the type of ``c`` is *context-dependent*: It could be either ``int``
-or ``Literal[19]`` depending on where it's used. For example, here is an example of
-where mypy will decide to use ``int`` over ``Literal[19]``:
+This means that while changing a variable to be ``Final`` is not quite the same thing
+as adding an explicit ``Literal[...]`` annotation, it often leads to the same effect
+in practice.
+
+The main cases where the behavior of context-sensitive vs true literal types differ are
+when you try using those types in places that are not explicitly expecting a ``Literal[...]``. 
+For example, compare and contrast what happens when you try appending these types to a list:
 
 ..code-block:: python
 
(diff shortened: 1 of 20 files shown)


Michael0x2a:
Note: I wasn't 100% sure what the new reveal_type format ought to look like. I was initially thinking of doing things like `builtins.int(fallback=Literal[1])` (more explicit) or `builtins.int(1)` (concise, mimics a runtime call), but eventually tentatively settled on `Literal[1]?` mostly because it seemed to best balance explicitness with concision.

I'm open to other ideas though.

docs/source/literal_types.rst
```diff
@@ -138,8 +138,27 @@ the above program almost as if it were written like so:
     reveal_type(19)
     expects_literal(19)
 
-This is why ``expects_literal(19)`` type-checks despite the fact that ``reveal_type(c)``
-reports ``int``.
+In other words, the type of ``c`` is *context-dependent*: It could be either ``int``
```
ilevkivskyi: This sentence is a bit repetitive, there is almost the same one paragraph above, but it uses term "context-sensitive" instead of "context-dependent". It is better to be more consistent with terms here.

mypy/checkmember.py
```diff
@@ -691,7 +691,12 @@ def analyze_class_attribute_access(itype: Instance,
 
     if info.is_enum and not (mx.is_lvalue or is_decorated or is_method):
         enum_literal = LiteralType(name, fallback=itype)
-        return itype.copy_modified(last_known_value=enum_literal)
+        # When we analyze enums, the corresponding Instance is always considered to be erased
+        # due to how the signature of Enum.__new__ is `(cls: Type[_T], value: object) -> _T`
+        # in Typeshed. However, this is really more of an implementation detail of how Enums
```
ilevkivskyi:
```suggestion
        # in typeshed. However, this is really more of an implementation detail of how Enums
```

mypy/erasetype.py
```diff
@@ -140,9 +140,12 @@ class LastKnownValueEraser(TypeTranslator):
...
     def visit_instance(self, t: Instance) -> Type:
-        if t.last_known_value:
-            return t.copy_modified(last_known_value=None)
-        return t
+        if not t.last_known_value and not t.args:
+            return t
+        return t.copy_modified(
+            args=[a.accept(self) for a in t.args],
+            last_known_value=None,
+        )
```
ilevkivskyi: Good catch! Did you add a test for this?

docs/source/literal_types.rst
Michael0x2a: I tried editing this section of the docs more heavily. If you have time, could you give the new version a second sanity check?

mypy/erasetype.py
Michael0x2a: Good point -- I initially didn't since we already had several tests unrelated to literal types that were breaking because of this.

But I think it'd be good to exercise this more directly, so I added a test to `check-literals.test`.

docs/source/literal_types.rst
```diff
@@ -138,11 +141,33 @@ the above program almost as if it were written like so:
...
+
+..code-block:: python
+
+    from typing_extensions import Final, Literal
+
+    a: Final = 19
+    b: Literal[19] = 19
+
+    # Mypy will chose to infer List[int] instead of List[Literal[19]?] or
+    # List[Literal[19]] since the former is most likely more useful.
```
ilevkivskyi: Shorten this comment to say just `# Mypy will chose to infer List[int] here`

PR #7960, opened 2019-11-16
Prevent dmypy from restarting if the configuration contains globs.

### LINKED ISSUE #7576: dmypy run keeps rerunning when using certain glob styles
With a configfile like this:

```
...
[mypy-abc.*]
...
```

dmypy run works fine, and only reruns when the configfile changes.

However, when I change the module configuration to:
```
...
[mypy-abc.*.abc]
...
```

dmypy keeps rerunning, saying:
```
Restarting: configuration changed
Daemon stopped
Daemon started
```

even without further changes to the config file.

I tried to track the error down, and changed the error message in`dmypy_server.py` to:
```
    def cmd_run(self, version: str, args: Sequence[str],
                is_tty: bool, terminal_width: int) -> Dict[str, object]:
        """Check a list of files, triggering a restart if needed."""
            ...
            # Signal that we need to restart if the options have changed
            if self.options_snapshot != options.snapshot():
                return {'restart': 'configuration changed' + str(self.options_snapshot) + '===' + str(options.snapshot())}
            ...
```

Comparing `self.options_snapshot` with `options.snapshot()`, I can see that with the second config file, `self.options_snapshot` contains this:
```
'glob_options':
[('abc.*.abc',
re.compile('abc(\\..*)?\\.abc\\Z'))],
```
while `options.snapshot()` contains this:
```
'glob_options':
[],
```

So I believe something is wrong in this case...

syastrov:
If a configuration contains globs, then dmypy incorrectly detected
that the configration changed even though it did not.
This caused the daemon to always restart if a configuration contained globs.

Fixes #7576.

### DIFF
--- a/test-data/unit/daemon.test
+++ b/test-data/unit/daemon.test
@@ -101,18 +101,17 @@ def plugin(version): return Dummy
 
 [case testDaemonRunRestartGlobs]
 -- Ensure dmypy is not restarted if the configuration doesn't change and it contains globs
-$ dmypy run -- foo --follow-imports=error
+-- Note: Backslash path separator in output is replaced with forward slash so the same test succeeds on Windows as well
+$ dmypy run -- foo --follow-imports=error --python-version=3.6 | {python} -c 'import sys; sys.stdout.write(sys.stdin.read().replace("\\", "/"))'
 Daemon started
 foo/lol.py:1: error: Name 'fail' is not defined
 Found 1 error in 1 file (checked 3 source files)
-== Return code: 1
-$ dmypy run -- foo --follow-imports=error
+$ dmypy run -- foo --follow-imports=error --python-version=3.6 | {python} -c 'import sys; sys.stdout.write(sys.stdin.read().replace("\\", "/"))'
 foo/lol.py:1: error: Name 'fail' is not defined
 Found 1 error in 1 file (checked 3 source files)
-== Return code: 1
 $ {python} -c "print('[mypy]')" >mypy.ini
 $ {python} -c "print('ignore_errors=True')" >>mypy.ini
-$ dmypy run -- foo --follow-imports=error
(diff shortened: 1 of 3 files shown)


syastrov:
I don't really think this is a "pretty" fix: it seems like `self.glob_options` is derived from `per_module_options`, so it should not be part of the snapshot. However, perhaps there is a better way to indicate which items should be part of the snapshot or not? (Perhaps all attributes starting with `_` should be excluded, and we could rename it to `self._glob_options`).
But, since there was precedent in deleting `per_module_cache` from the snapshot, I just implemented it this way.

By the way, this bug is really nasty, because it is quite useful to ignore files in mypy.ini using globs and by doing that, it means the daemon always restarts.

syastrov:
I did a little refactor to make the attributes private as I commented above. You can choose to include the last commit or not depending on whether you think it makes sense.
With a little hackery, tests should now pass on Windows. I couldn't find a prettier way to do that.

syastrov:
OK, test is not passing on Windows. I don't have time to look at it right now, but am open to suggestions.

test-data/unit/daemon.test
```diff
@@ -99,6 +99,37 @@ from mypy.plugin import Plugin
 class Dummy(Plugin): pass
 def plugin(version): return Dummy
 
+[case testDaemonRunRestartGlobs]
+-- Ensure dmypy is not restarted if the configuration doesn't change and it contains globs
+-- Note: Backslash path separator in output is replaced with forward slash so the same test succeeds on Windows as well
+$ dmypy run -- foo --follow-imports=error --python-version=3.6 | {python} -c 'import sys; sys.stdout.write(sys.stdin.read().replace("\\", "/"))'
```
msullivan: `'` isn't a quote character in windows command line parsing, I think.

I would ditch the pipe stuff, though and do the slash normalization in the test case driver. We have a `normalize_error_messages` function that fixes path slashes in test output that we use in most of our test drivers but don't seem to call in `testdaemon`.

syastrov:
It should be ready now

PR #7962, opened 2019-11-16
[mypyc] Add basic support for protocols

ilevkivskyi:
Fixes https://github.com/mypyc/mypyc/issues/586

This is more like WIP because I am not sure what are the dangerous corners to test and whether this is a viable approach in general, I will be grateful for suggestions.

Currently I essentially use python operations on protocols everywhere. It looks like just marking the protocol class as non extension already triggers this.


ilevkivskyi:

PR #7970, opened 2019-11-18
[suggest] Make the max number of guesses/arguments configurable

msullivan:


msullivan:
Actually I'm going to just get rid of the max arguments thing and do something smarter on that side

PR #7971, opened 2019-11-18
Implements module attribute suggestions

theodoretliu:
Addresses the remainder of issue #824

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1903,15 +1903,17 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             # This is a reference to a non-module attribute.
             original_type = self.accept(e.expr)
             base = e.expr
-            module_members = None
+            module_symbol_table = None
 
             if isinstance(base, RefExpr) and isinstance(base.node, MypyFile):
-                module_members = [key for key in base.node.names.keys() if not base.node.names.get(key).module_hidden]
+                module_symbol_table = base.node.names
 
             member_type = analyze_member_access(
                 e.name, original_type, e, is_lvalue, False, False,
                 self.msg, original_type=original_type, chk=self.chk,
-                in_literal_context=self.is_literal_context(), module_members=module_members)
+                in_literal_context=self.is_literal_context(),
+                module_symbol_table=module_symbol_table)
+
             return member_type
 
     def analyze_external_member_access(self, member: str, base_type: Type,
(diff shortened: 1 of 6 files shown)


mypy/checkexpr.py
```diff
@@ -1902,10 +1902,16 @@ def analyze_ordinary_member_access(self, e: MemberExpr,
         else:
             # This is a reference to a non-module attribute.
             original_type = self.accept(e.expr)
+            base = e.expr
+            module_members = None
+
+            if isinstance(base, RefExpr) and isinstance(base.node, MypyFile):
+                module_members = [key for key in base.node.names.keys() if not base.node.names.get(key).module_hidden]
```
JukkaL: This can be expensive to calculate, in case the module has many attributes. Instead of passing the members, what about passing a `SymbolTable` object, and calculating the members only later if we generate an error message?

test-data/unit/check-columns.test
```diff
@@ -126,7 +126,7 @@ import m
 if int():
     from m import foobaz # E:5: Module 'm' has no attribute 'foobaz'; maybe "foobar"?
 (1).x # E:2: "int" has no attribute "x"
-(m.foobaz()) # E:2: Module has no attribute "foobaz"
+(m.foobaz()) # E:2: Module has no attribute "foobaz"; maybe "foobar"?
```
JukkaL: Maybe add a test case where there are multiple possible suggestions?

mypy/messages.py
```diff
@@ -175,7 +175,12 @@ def note_multiline(self, messages: str, context: Context, file: Optional[str] =
     # get some information as arguments, and they build an error message based
     # on them.
 
-    def has_no_attr(self, original_type: Type, typ: Type, member: str, context: Context) -> Type:
+    def has_no_attr(self,
+                    original_type: Type,
```
JukkaL: If you pass around a symbol table, one option may be to change the type of this argument to `Union[Type, SymbolTable]` (and rename it).

JukkaL:
Also, self check seems to be failing (Travis CI).

theodoretliu:
Hey, I addressed the comments! I wasn't sure what you meant by the `Union` type though

ilevkivskyi:
(GitHub did something very weird and I have no idea how to undo this.)

theodoretliu:
Hm, I never received an email for that comment either

ilevkivskyi:
Anyway, I think Jukka meant that you can use the same argument (say `extra_info`) instead of two separate ones like `original_type` and `module_symbol_table`, and adjust the logic in the method. I don't think this is really important. I think however it makes sense in any case update the docstrings for `analyze_member_access()` and `has_no_attr()` to mention the args you added.

PR #7972, opened 2019-11-18
[suggest] Rework max guesses and eliminate max args

msullivan:
Rework max guesses to be configurable and to throw out guesses past
the limit instead of just hard failing. This might allow succeeding in
more situations. Eliminate the maximum argument restriction since we
no longer have to compute all of the options even if we don't try
that.

### DIFF
--- a/test-data/unit/fine-grained-suggest.test
+++ b/test-data/unit/fine-grained-suggest.test
@@ -44,6 +44,23 @@ bar.py:4: (arg=str)
 bar.py:6: (*typing.List[str])
 bar.py:8: (**typing.Dict[str, str])
 
+[case testMaxGuesses]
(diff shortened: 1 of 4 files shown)


test-data/unit/fine-grained-suggest.test
```diff
@@ -44,6 +44,23 @@ bar.py:4: (arg=str)
 bar.py:6: (*typing.List[str])
 bar.py:8: (**typing.Dict[str, str])
 
+[case testMaxGuesses]
+# suggest: foo.foo
+# suggest: --max-guesses=2 foo.foo
+[file foo.py]
+# This idea here is that we can only find the union type with more guesses.
```
ilevkivskyi: This idea -> The idea?

PR #7975, opened 2019-11-19
fixes #6513 change "Cannot find module" message

### LINKED ISSUE #6513: Cannot find module that is 100% for sure installed
I'm encountering a bug with mypy telling me that it can't find a module that I know is installed.

The module in question is an internal library called `mango`, and the project I'm trying to use it in is called `raptor`. For `mango`, all functions are annotated and mypy passes with flying colors.

I am on macOS, Python 3.7, and using tox to run my tests. Below is the part of my `tox.ini` file that has the environment I've defined for mypy:

```
[testenv:type]
deps =
    vendor/mango-0.4.3.tar.gz
    mypy
commands =
    python -c "import mango; print(mango.__file__)"
    pip list
    mypy --package raptor
```

Normally, under commands I only have the mypy executable line, but for this post I've added the other two lines just to prove that there should be no issues importing `mango`. Below is the error output from tox:

```
type installed: raptor==0.2.0,mango==0.4.3,mpmath==1.1.0,mypy==0.670,mypy-extensions==0.4.1,numpy==1.16.2,scipy==1.2.1,six==1.12.0,sympy==1.3,typed-ast==1.3.1,unyt==1.0.7
type run-test-pre: PYTHONHASHSEED='3632297771'
type runtests: commands[0] | python -c 'import mango; print(mango.__file__)'
/Users/username/projects/raptor/.tox/type/lib/python3.7/site-packages/mango/__init__.py
type runtests: commands[1] | pip list
Package         Version
--------------- -------
raptor          0.2.0  
mango           0.4.3  
mpmath          1.1.0  
mypy            0.670  
mypy-extensions 0.4.1  
numpy           1.16.2 
pip             19.0.3 
scipy           1.2.1  
setuptools      40.8.0 
six             1.12.0 
sympy           1.3    
typed-ast       1.3.1  
unyt            1.0.7  
wheel           0.33.1 
type runtests: commands[2] | mypy --package raptor
raptor/__init__.py:5: error: Cannot find module named 'mango'
raptor/__init__.py:5: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
...
ERROR: InvocationError for command '/Users/username/projects/raptor/.tox/type/bin/mypy --package raptor' (exited with code 1)
commands failed
```

As you can see, Python is able to import my package, and pip also shows that it is there. pytest and sphinx are both able to import `mango` in their respective virtual environments. It's only mypy that is broken. 

I'm not sure how to fix this issue because mypy is lying to me telling me that `mango` isn't installed. Since I can import it just fine, and it is installed into the virtualenv, I feel like I shouldn't have to modify MYPYPATH or anything.

rooterkyberian:
An attempt to make message less confusing.
Right it seems quite common for new users to be
1) get confused, try to figure out why mypy doesn't see the installed module
2) google for a problem, get a hit on a mypy website FAQ and only then
learn what this error is really about

Hopefully adding info that is not about module, but its type information
will speedup this process.
It also should make message itself more googlable that "cannot find
module" which does not return relevant sites unless `mypy` is added to
keywords.

### DIFF
--- a/mypy/suggestions.py
+++ b/mypy/suggestions.py
@@ -489,7 +489,7 @@ class SuggestionEngine:
         else:
             target = split_target(self.fgmanager.graph, key)
             if not target:
-                raise SuggestionFailure("Cannot find module for %s" % (key,))
+                raise SuggestionFailure("Cannot find module type information for %s" % (key,))
             modname, tail = target
             node = self.find_node_by_module_and_name(modname, tail)
 


mypy/suggestions.py
```diff
@@ -489,7 +489,7 @@ def find_node(self, key: str) -> Tuple[str, str, FuncDef]:
         else:
             target = split_target(self.fgmanager.graph, key)
             if not target:
-                raise SuggestionFailure("Cannot find module for %s" % (key,))
+                raise SuggestionFailure("Cannot find module type information for %s" % (key,))
```
JukkaL: This is actually not relevant for #6513, since this change only applies to `dmypy suggest`, which is still an undocumented feature. Feel free to open a new issue if you still think that this is worth changing.

rooterkyberian: oops, I blindly changed the only matching line in the code - sorry for wasting your time!

it seems the "fix" was already done in 70f2857f2ef867ce6d238838367406bc3e8e14de ;) never mind me then!

PR #7978, opened 2019-11-19
mypyc: Fix C string encoding

saleemrashid:
Using `repr()` to encode C strings relied on implementation-defined behaviour and had a number of bugs, for example:

* The C string could inadvertently contain trigraphs.

* Valid hexdecimal characters following a hexadecimal escape sequence would be parsed as part of the escape sequence.

Octal escape sequences are used for unprintable characters because they do not have the same issue as hexadecimal escape sequences.

It would be possible to mitigate the issue with hexadecimal escape sequences by emitting multiple string literals. However, the complexity is not worth it, especially as the output is not designed to be human-readable.

As an ad hoc test, I generated a Python module containing string literals for every possible character:

```python
with open("literals.py", "w") as f:
    f.write(
        "LITERALS = [{}]".format(
            ", ".join([repr(bytes([x, ord("0")])) for x in range(256)])
        )
    )
```

Then I compiled the module with Mypyc and checked it produced the correct byte strings:

```python
import literals

assert literals.__file__.endswith(".so")

for x in range(256):
    assert literals.LITERALS[x] == bytes([x, ord("0")])
```

### DIFF
--- a/mypyc/cstring.py
+++ b/mypyc/cstring.py
@@ -1,0 +1,49 @@
+"""Encode valid C string literals from Python strings.
+
+If a character is not allowed in C string literals, it is either emitted
+as a simple escape sequence (e.g. '\\n'), or an octal escape sequence
+with exactly three digits ('\\oXXX'). Question marks are escaped to
+prevent trigraphs in the string literal from being interpreted. Note
+that '\\?' is an invalid escape sequence in Python.
+
+Consider the string literal "AB\\xCDEF". As one would expect, Python
+parses it as ['A', 'B', 0xCD, 'E', 'F']. However, the C standard
+specifies that all hexadecimal digits immediately following '\\x' will
+be interpreted as part of the escape sequence. Therefore, it is
+unexpectedly parsed as ['A', 'B', 0xCDEF].
+
+Emitting ("AB\\xCD" "EF") would avoid this behaviour. However, we opt
+for simplicity and use octal escape sequences instead. They do not
+suffer from the same issue as they are defined to parse at most three
+octal digits.
+"""
+
+import string
+from typing import Tuple
+
+CHAR_MAP = ['\\{:03o}'.format(i) for i in range(256)]
+
+# It is safe to use string.printable as it always uses the C locale.
+for c in string.printable:
+    CHAR_MAP[ord(c)] = c
+
+# These assignments must come last because we prioritize simple escape
+# sequences over any other representation.
+for c in ('\'', '"', '\\', 'a', 'b', 'f', 'n', 'r', 't', 'v'):
+    escaped = '\\{}'.format(c)
+    decoded = escaped.encode('ascii').decode('unicode_escape')
+    CHAR_MAP[ord(decoded)] = escaped
+
+# This escape sequence is invalid in Python.
+CHAR_MAP[ord('?')] = r'\?'
+
+
+def encode_as_c_string(s: str) -> Tuple[str, int]:
+    """Produce a quoted C string literal and its size, for a UTF-8 string."""
+    return encode_bytes_as_c_string(s.encode('utf-8'))
+
+
+def encode_bytes_as_c_string(b: bytes) -> Tuple[str, int]:
+    """Produce a quoted C string literal and its size, for a byte string."""
+    escaped = ''.join([CHAR_MAP[i] for i in b])
+    return '"{}"'.format(escaped), len(b)
--- a/mypyc/emitmodule.py
+++ b/mypyc/emitmodule.py
@@ -23,6 +23,7 @@ from mypyc import genops
 from mypyc.common import (
     PREFIX, TOP_LEVEL_NAME, INT_PREFIX, MODULE_PREFIX, shared_lib_name,
 )
+from mypyc.cstring import encode_as_c_string, encode_bytes_as_c_string
 from mypyc.emit import EmitterContext, Emitter, HeaderDeclaration
 from mypyc.emitfunc import generate_native_function, native_function_header
 from mypyc.emitclass import generate_class_type_decl, generate_class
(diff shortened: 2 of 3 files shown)


mypyc/emitmodule.py
```diff
@@ -421,8 +449,7 @@ def encode_as_c_string(s: str) -> Tuple[str, int]:
 
 def encode_bytes_as_c_string(b: bytes) -> Tuple[str, int]:
     """Produce a single-escaped, quoted C string and its size from a bytes"""
-    # This is a kind of abusive way to do this...
-    escaped = repr(b)[2:-1].replace('"', '\\"')
+    escaped = ''.join(map(C_CHAR_MAP.__getitem__, b))
```
msullivan: I'd write this as a list comprehension instead of a map

mypyc/emitmodule.py
```diff
@@ -67,6 +68,33 @@
 # A list of (file name, file contents) pairs.
 FileContents = List[Tuple[str, str]]
 
+# The C standard specifies that any number of valid hexadecimal characters are
```
msullivan: I think it's worth moving this out of `emitmodule` at this point. `mypyc.cstring`, maybe?

mypyc/emitmodule.py
```diff
@@ -67,6 +68,33 @@
...
+# digits.
+C_CHAR_MAP = ["\\{:03o}".format(x) for x in range(256)]
+# Most printable characters do not need to be escaped in string literals. We
+# can safely use string.printable here because it always uses the C locale.
+for x in string.printable:
+    C_CHAR_MAP[ord(x)] = x
+# These assignments must be done after string.printable because they are
+# overrides for the printable characters that need to be escaped in string
+# literals.
+C_CHAR_MAP[ord('\'')] = '\\\''
```
msullivan: These are all generated the same way (adding a backslash), so I think it would be an improvement to generate them all mechanically from a list of characters to escape.

mypyc/emitmodule.py
msullivan: (And then encode_as_c_string and encode_bytes_as_c_string too)

saleemrashid:
I've force-pushed some changes, i.e. rewrote the comments, and refactored the escaped strings to raw string literals.

I will try and implement your suggestions shortly.

> I'm also sad about the octal thing but oh well.

@msullivan My original prototype used `itertools.groupby` to group blocks of printable characters and unprintable characters so you could safely use hexadecimal escape sequences. It worked fine but it seemed like unnecessary complexity.

mypyc/emitmodule.py
saleemrashid: I considered this, but the neatest way I can think to do this is:

```python
for x in ("'", '"', '\\', 'a', 'b', 'f', 'n', 'r', 't', 'v'):
    escaped = '\\{}'.format(x)
    value = escaped.encode('ascii').decode('unicode_escape')
    C_CHAR_MAP[ord(value)] = escaped
```

I am not sure whether this is an improvement. What do you think?

msullivan:
> @msullivan My original prototype used `itertools.groupby` to group blocks of printable characters and unprintable characters so you could safely use hexadecimal escape sequences. It worked fine but it seemed like unnecessary complexity.

I agree that using octal is preferable to writing complex code that avoids it.

mypyc/cstring.py
```diff
@@ -0,0 +1,53 @@
...
+
+CHAR_MAP = ['\\{:03o}'.format(i) for i in range(256)]
+
+# It is safe to use string.printable as it always uses the C locale.
+for c in string.printable:
+    CHAR_MAP[ord(c)] = c
+
+# These assignments must come last because we prioritize the escape
+# sequences in the C standard over any other representation.
+CHAR_MAP[ord('\'')] = r'\''
```
saleemrashid: @msullivan This is the simplest method I have been able to devise to generate these escape sequences:

```python
for x in ("'", '"', '\\', 'a', 'b', 'f', 'n', 'r', 't', 'v'):
    escaped = '\\{}'.format(x)
    value = escaped.encode('ascii').decode('unicode_escape')
    C_CHAR_MAP[ord(value)] = escaped
```

While I am not a huge fan of manually listing each escape sequence, I dislike this method more.

msullivan: The encode/decode is a little hokey but I think I like this more

msullivan:
Could you also remove `-Wno-trigraph` from our C flags in `build.py` <_<

mypyc/cstring.py
```diff
@@ -0,0 +1,53 @@
...
+CHAR_MAP[ord('\t')] = r'\t'
+CHAR_MAP[ord('\v')] = r'\v'
+
+
+def encode_as_c_string(s: str) -> Tuple[str, int]:
+    """Produce a quoted C string literal and its size, for a UTF-8 string."""
+    return encode_bytes_as_c_string(s.encode('utf-8'))
+
+
+def encode_bytes_as_c_string(b: bytes) -> Tuple[str, int]:
```
saleemrashid: This is what an implementation with hexadecimal escape sequences, instead of octal, would look like:

```diff
diff --git i/mypyc/cstring.py w/mypyc/cstring.py
index 37f3ab38..0557a216 100644
--- i/mypyc/cstring.py
+++ w/mypyc/cstring.py
@@ -19,9 +19,11 @@ octal digits.
 """
 
 import string
+import itertools
 from typing import Tuple
 
-CHAR_MAP = ['\\{:03o}'.format(i) for i in range(256)]
+HEX_DIGITS = frozenset(string.hexdigits)
+CHAR_MAP = {}
 
 # It is safe to use string.printable as it always uses the C locale.
 for c in string.printable:
@@ -49,5 +51,13 @@ def encode_as_c_string(s: str) -> Tuple[str, int]:
 
 def encode_bytes_as_c_string(b: bytes) -> Tuple[str, int]:
     """Produce a quoted C string literal and its size, for a byte string."""
-    escaped = ''.join([CHAR_MAP[i] for i in b])
+    escaped = ''
+    for printable, group in itertools.groupby(b, key=CHAR_MAP.__contains__):
+        if printable:
+            s = ''.join([CHAR_MAP[i] for i in group])
+            if len(escaped) and s[0] in HEX_DIGITS:
+                escaped += '" "'
+            escaped += s
+        else:
+            escaped += ''.join(['\\x{:02X}'.format(i) for i in group])
     return '"{}"'.format(escaped), len(b)
```

mypyc/cstring.py
saleemrashid: The main reason I dislike this is that we're calling into Python's Unicode APIs. There's also `codecs.escape_decode` that handles ASCII only, but it's undocumented.

msullivan:
Thanks!

PR #7979, opened 2019-11-20
[suggest] Fix inference from collection literals

msullivan:
This is another take at #7810 which was reverted because it caused
internal trouble.

Do this by ignoring constraints with Any when the Any comes from the
suggestion engine. The Any context causes assigned types to be Any
overaggressively.  Restricting it only to Anys from the suggestion
engine avoids the issues seen last time. Doing it at a fine-grained
level in constraints allows us to do better refining things like `List[Any]`.

### DIFF
--- a/mypy/suggestions.py
+++ b/mypy/suggestions.py
@@ -23,14 +23,14 @@ Other things:
 """
 
 from typing import (
-    List, Optional, Tuple, Dict, Callable, Union, NamedTuple, TypeVar, Iterator,
+    List, Optional, Tuple, Dict, Callable, Union, NamedTuple, TypeVar, Iterator, cast,
 )
 from typing_extensions import TypedDict
 
 from mypy.state import strict_optional_set
 from mypy.types import (
     Type, AnyType, TypeOfAny, CallableType, UnionType, NoneType, Instance, TupleType,
-    TypeVarType, FunctionLike,
+    TypeVarType, FunctionLike, UninhabitedType,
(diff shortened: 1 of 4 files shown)


mypy/suggestions.py
```diff
@@ -426,7 +430,7 @@ def get_suggestion(self, mod: str, node: FuncDef) -> PyAnnotateSignature:
                 callsites,
                 uses,
             )
-        guesses = self.filter_options(guesses, is_method)
+        guesses = self.filter_options(guesses, is_method, ignore_return=True)
```
JukkaL: How is the `ignore_return=True` change related to the rest of the PR?

mypy/types.py
```diff
@@ -103,6 +103,8 @@ class TypeOfAny:
     from_another_any = 7  # type: Final
     # Does this Any come from an implementation limitation/bug?
     implementation_artifact = 8  # type: Final
+    # Does this Any come from use in the suggestion engine?
```
JukkaL: Maybe also explain that this allows these `Any` types to be ignored when inferring suggestion, resulting in more precise types.

mypy/suggestions.py
msullivan: I was abusing `special_form` because `has_any_type` didn't consider it an any. I didn't want to extend it to know about `suggestion_engine` also (and doing so would have interfered some with the use of `suggestion_engine` when refining)

PR #7980, opened 2019-11-20
[suggest] Use the type of a method in a parent class as a guess

msullivan:


mypy/suggestions.py
```diff
@@ -404,6 +404,22 @@ def find_best(self, func: FuncDef, guesses: List[CallableType]) -> Tuple[Callabl
...
+            return []
+
+        for parent in node.info.mro[1:]:
+            pnode = parent.names.get(node.name)
+            if pnode and isinstance(pnode.node, (FuncDef, Decorator)):
+                typ = pnode.node.type
+                if isinstance(typ, CallableType) and len(typ.arg_types) == len(node.arguments):
+                    # Return the first thing we find, since it probably doesn't make sense
+                    # to grab things further up in the chain if an earlier parent has it.
+                    return [pnode.node.type]
```
JukkaL: This doesn't work as expected if a base class is generic. Maybe fix it (if it's easy to fix) or add a TODO comment.

JukkaL:
Also Travis CI is failing.

PR #7987, opened 2019-11-20
[mypyc] Handle non-extension classes with built-in bases

saleemrashid:
For both of these class definitions, Mypyc will generate the equivalent of `globals()["int"]`, which will raise `KeyError: 'int'` (as `int` is a built-in, not a global).

```python
from enum import Enum, IntEnum


class A(IntEnum):
    pass


class B(int, Enum):
    pass
```

I'm incredibly unfamiliar with the Mypy codebase, but I think this is the correct solution.

---

Consider the following class definition:

```python
class MyClass(IntEnum):
    ...
```

CPython would call `EnumMeta.__new__(bases=[IntEnum])`. On the other hand, Mypyc would use the MRO (with `MyClass` and `builtins.object` removed) and call `EnumMeta.__new__(bases=[IntEnum, int, Enum])`.

Additionally, due to the use of `load_global_str` instead of `load_global`, it would attempt to retrieve `globals()["int"]`, which would raise a `KeyError` at runtime.

This is fixed by using the base type expressions in the class definition instead of the MRO, and using `load_global` instead of `load_global_str` (which correctly handles built-ins).

### DIFF
--- a/mypyc/genops.py
+++ b/mypyc/genops.py
@@ -1373,7 +1373,6 @@ class IRBuilder(ExpressionVisitor[Value], StatementVisitor[None]):
         for non-extension classes.
         """
         ir = self.mapper.type_to_ir[cdef.info]
-        bases = []
         for cls in cdef.info.mro[1:]:
             if cls.fullname == 'builtins.object':
                 continue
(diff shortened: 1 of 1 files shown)


msullivan:
Could you add a test? The relevant tests are driven by data files and live in `mypyc/test-data/run*.test`. There might be an existing test in `run-classes.test` that it could be added to, or make a new one.

saleemrashid:
@msullivan Hi, I'll busy with school at the moment but will work on this as soon as I'm free to.

saleemrashid:
@msullivan Hi, I'm busy with school at the moment but will work on this as soon as I'm free to.

msullivan:
All right, I'm going to close this PR but feel free to reopen it as soon as you have a chance to approach it again

PR #7989, opened 2019-11-21
[mypyc] Fix some deserialization related issues

msullivan:
Certain parts of the mypy AST aren't present and we were relying on
them anyways. This caused a crash when subclassing and classes not
making it into the type table, preventing using fast paths.

### DIFF
--- a/mypyc/test/test_run.py
+++ b/mypyc/test/test_run.py
@@ -224,8 +224,10 @@ class TestRun(MypycDataSuite):
                 print(line)
             assert False, 'Compile error'
 
-        # Check that serialization works on this IR
-        check_serialization_roundtrip(ir)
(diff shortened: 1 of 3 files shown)


mypyc/test/test_run.py
```diff
@@ -224,8 +224,11 @@ def run_case_step(self, testcase: DataDrivenTestCase, incremental_step: int) ->
                 print(line)
             assert False, 'Compile error'
 
-        # Check that serialization works on this IR
-        check_serialization_roundtrip(ir)
+        # Check that serialization works on this IR. (Only on the first
+        # step because the test doesn't track the contexts and ir only
```
JukkaL: It's unclear to me what "track the contexts" means here.

PR #8004, opened 2019-11-23
[mypyc] Add a mypyc_attr to support interpreted children

### LINKED ISSUE #296: Tuple[...] should be compatible with Iterable[...] and Sequence[...]
Currently, Tuple[int, int] is not compatible with Iterable[int] and Sequence[int], but it should obviously be. Also, update type join to support this (join of Tuple[int, int] and Tuple[int] should be Sequence[int]).

For example, this program is rejected:

```
from typing import Iterable
def f(x: Iterable[int]) -> None: pass
f([1,2])  # OK
f((1,2))  # Argument 1 to "f" has incompatible type "Tuple[int, int]"
```

See also #184 (it's very closely related).

msullivan:
This operates by generating a "shadow vtable" containing pointers to
glue methods that dispatch to the appropriate method via the C API. We
then install those shadow vtables in interpreted subclasses so that
overridden methods will be called.

This does not support directly inheriting from traits, which I think
will require generating vtables dynamically (and maybe some more
nonsense too.)

Closes #296. (Though I will file a follow-up for traits.)

### DIFF
--- a/mypyc/genops.py
+++ b/mypyc/genops.py
@@ -392,6 +392,8 @@ def compute_vtable(cls: ClassIR) -> None:
             # TODO: don't generate a new entry when we overload without changing the type
             if fn == cls.get_method(fn.name):
                 cls.vtable[fn.name] = len(entries)
+                # If the class contains a glue method referring to itself, that is a
+                # shadow glue method to support interpreted subclasses.
                 shadow = cls.glue_methods.get((cls, fn.name))
                 entries.append(VTableMethod(t, fn.name, fn, shadow))
 
@@ -1772,6 +1774,14 @@ class IRBuilder(ExpressionVisitor[Value], StatementVisitor[None]):
                  *,
                  do_py_ops: bool = False
                  ) -> FuncIR:
+        """Generate glue methods that mediate between different method types in subclasses.
+
+        Works on both properties and methods. See gen_glue_methods below for more details.
+
+        If do_py_ops is True, then the glue methods should use generic
+        C API operations instead of direct calls, to enable generating
+        "shadow" glue methods that work with interpreted subclasses.
+        """
         if fdef.is_property:
             return self.gen_glue_property(sig, target, cls, base, fdef.line, do_py_ops)
         else:
@@ -1802,6 +1812,9 @@ class IRBuilder(ExpressionVisitor[Value], StatementVisitor[None]):
         we need to generate glue methods that mediate between the
         different versions by coercing the arguments and return
         values.
+
+        If do_pycall is True, then make the call using the C API
+        instead of a native call.
         """
         self.enter(FuncInfo())
         self.ret_types[-1] = sig.ret_type
@@ -1836,9 +1849,15 @@ class IRBuilder(ExpressionVisitor[Value], StatementVisitor[None]):
     def gen_glue_property(self, sig: FuncSignature, target: FuncIR, cls: ClassIR, base: ClassIR,
                           line: int,
                           do_pygetattr: bool) -> FuncIR:
-        """Similarly to methods, properties of derived types can be covariantly subtyped. Thus,
+        """Generate glue methods for properties that mediate between different subclass types.
+
+        Similarly to methods, properties of derived types can be covariantly subtyped. Thus,
         properties also require glue. However, this only requires the return type to change.
-        Further, instead of a method call, an attribute get is performed."""
+        Further, instead of a method call, an attribute get is performed.
+
+        If do_pygetattr is True, then get the attribute using the C
+        API instead of a native call.
+        """
         self.enter(FuncInfo())
 
         rt_arg = RuntimeArg(SELF_NAME, RInstance(cls))
--- a/mypyc/genops.py
+++ b/mypyc/genops.py
@@ -239,10 +239,55 @@ def is_dataclass(cdef: ClassDef) -> bool:
     return any(is_dataclass_decorator(d) for d in cdef.decorators)
 
 
-def is_extension_class(cdef: ClassDef) -> bool:
+def get_mypyc_attr_literal(e: Expression) -> Any:
+    """Convert an expression from a mypyc_attr decorator to a value.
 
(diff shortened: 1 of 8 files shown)


mypyc/genops.py
```diff
@@ -346,7 +392,8 @@ def compute_vtable(cls: ClassIR) -> None:
             # TODO: don't generate a new entry when we overload without changing the type
             if fn == cls.get_method(fn.name):
                 cls.vtable[fn.name] = len(entries)
-                entries.append(VTableMethod(t, fn.name, fn))
+                shadow = cls.glue_methods.get((cls, fn.name))
```
JukkaL: Add comment since this seems pretty key to how this works.

mypyc/genops.py
```diff
@@ -1164,25 +1215,28 @@ def handle_ext_method(self, cdef: ClassDef, fdef: FuncDef) -> None:
...
+                class_ir.glue_methods[(base, name)] = f
                 self.functions.append(f)
 
+        # If the class allows interpreted children, create glue
+        # methods that dispatch via the Python API. These will go in a
+        # "shadow vtable" that will be assigned to interpreted
+        # children.
+        if class_ir.allow_interpreted_children:
+            f = self.gen_glue(func_ir.sig, func_ir, class_ir, class_ir, fdef, do_py_ops=True)
+            class_ir.glue_methods[(class_ir, name)] = f
```
JukkaL: What if a class that allow interpreted children has a native subclass? Will these glue methods cause trouble?

I wonder if putting the python glue methods in a separate data structure would be clearer.

mypyc/genops.py
```diff
@@ -1707,8 +1766,20 @@ def visit_import_all(self, node: ImportAll) -> None:
             return
         self.gen_import(node.id, node.line)
 
+    def gen_glue(self, sig: FuncSignature, target: FuncIR,
+                 cls: ClassIR, base: ClassIR, fdef: FuncItem,
+                 *,
+                 do_py_ops: bool = False
+                 ) -> FuncIR:
+        if fdef.is_property:
```
JukkaL: Add docstring. At least explain `do_py_ops`.

mypyc/genops.py
```diff
@@ -1707,8 +1766,20 @@ def visit_import_all(self, node: ImportAll) -> None:
...
+            return self.gen_glue_property(sig, target, cls, base, fdef.line, do_py_ops)
+        else:
+            return self.gen_glue_method(sig, target, cls, base, fdef.line, do_py_ops)
+
     def gen_glue_method(self, sig: FuncSignature, target: FuncIR,
-                        cls: ClassIR, base: ClassIR, line: int) -> FuncIR:
+                        cls: ClassIR, base: ClassIR, line: int,
+                        do_pycall: bool,
+                        ) -> FuncIR:
         """Generate glue methods that mediate between different method types in subclasses.
```
JukkaL: Document `do_pycall`.

mypyc/genops.py
```diff
@@ -1758,7 +1833,8 @@ def f(self, x: object) -> int: ...
             blocks, env)
 
     def gen_glue_property(self, sig: FuncSignature, target: FuncIR, cls: ClassIR, base: ClassIR,
-                          line: int) -> FuncIR:
+                          line: int,
+                          do_pygetattr: bool) -> FuncIR:
```
JukkaL: Document `do_pygetattr`.

mypyc/test-data/commandline.test
```diff
@@ -170,3 +174,11 @@ def f(l: List[object]) -> None:
     for i in l:
         if x is None:
             x = i
+
+@mypyc_attr(allow_interpreted_children=True)
```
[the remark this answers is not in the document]
msullivan: I think the `allow_` prefix is important because otherwise it could be read as causing all subclasses to be interpreted?

mypyc/ops.py
```diff
@@ -1677,15 +1677,24 @@ def deserialize(cls, data: JsonDict, ctx: DeserMaps) -> 'FuncIR':
 # The arrow points to the "start" of the vtable (what vtable pointers
 # point to) and the bars indicate which parts correspond to the parent
 # class A's vtable layout.
+#
+# Classes that allow interpreted code to subclass them also have a
+# "shadow vtable" that contains implementations that delegate to
+# making a pycall, so that overridden methods in interpreted children
+# will be called. (A better strategy could dynamically generate these
+# vtables based on which methods are overridden in the children.)
```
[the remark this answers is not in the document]
msullivan: I think probably. Monkey patching without incurring a dict lookup in the non-monkey-patched case is gonna require some heroics anyway (I think a metaclass that catches writes and updates the vtable or something similar) and I think they will all apply here too

mypyc/test-data/commandline.test
```diff
@@ -170,3 +174,11 @@ def f(l: List[object]) -> None:
...
+class AllowInterp1(Concrete1):  # E: Parents must allow interpreted children also
```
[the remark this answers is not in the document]
msullivan: Yeah

mypyc/genops.py
msullivan: Shouldn't cause trouble

mypyc/test-data/run-classes.test
```diff
@@ -1198,3 +1198,85 @@ test(b, 20)
...
+assert y.bar(10) == "child bar: 20"
+assert y.foo() == "parent foo: child bar: 20"
+assert foo(y) == "parent foo: child bar: 20"
+assert bar(y, 30) == "child bar: 40"
+assert spam(y) == "child"
+assert y.read_property == "child prop"
+assert prop(x) == "parent prop"
+assert prop(y) == "child prop"
+
+assert y.trait_method() == "child"
```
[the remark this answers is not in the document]
msullivan: "Interpreted class overrides an attribute with a settable property" and "Multiple inheritance (multiple native base classes)" will definitely not work. I'll add tests for the rest and file issues for that one. (Multiple inheritance was on my radar, overriding attr with property was not. 

I'll add tests for the rest

PR #8011, opened 2019-11-26
Fix false positive: return Any from a function declared to return object

### LINKED ISSUE #8008: False positive on returning Any from function declared to return object
mypy with `--warn-return-any` gives a warning for this function:
```python
def f() -> object:
    x: Any = 1
    return x
```
This code is correct in fact: any type is `object`, so we can safely return `Any`.
It can be fixed like that:
```python
def f() -> object:
    x: Any = 1
    return cast(object, x)
```
but `cast(object, ...)` should be a no-op for mypy IMO.

TH3CHARLie:
resolves #8008

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3038,7 +3038,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                         and not (defn.name in BINARY_MAGIC_METHODS and
                                  is_literal_not_implemented(s.expr))
                         and not (isinstance(return_type, Instance) and
-                                return_type.type.fullname == 'builtins.object')):
+                                 return_type.type.fullname == 'builtins.object')):
                         self.msg.incorrectly_returning_any(return_type, s)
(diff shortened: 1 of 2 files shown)


mypy/checker.py
```diff
@@ -3036,7 +3036,9 @@ def check_return_stmt(self, s: ReturnStmt) -> None:
                         and not self.current_node_deferred
                         and not is_proper_subtype(AnyType(TypeOfAny.special_form), return_type)
                         and not (defn.name in BINARY_MAGIC_METHODS and
-                                 is_literal_not_implemented(s.expr))):
+                                 is_literal_not_implemented(s.expr))
+                        and not (isinstance(return_type, Instance) and
+                                return_type.type.fullname == 'builtins.object')):
```
```suggestion
                                 return_type.type.fullname == 'builtins.object')):
```

TH3CHARLie:
Thanks for the review, the indentation has been updated

PR #8014, opened 2019-11-26
Support type alias as Literal

### LINKED ISSUE #7996: Variant of alias for enum type cannot be used as Literal
```python
class Foo(Enum):
    A = 1

F = Foo

# works
x: Literal[Foo.A]
# name-defined error
y: Literal[F.A]
```

Is there a simple workaround here? I also tried `F: Final = Foo` which doesn't work either

TH3CHARLie:
resolves #7996

I'm not very sure about how effective the current testcase is, suggestions are welcomed.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3971,8 +3971,7 @@ class SemanticAnalyzer(NodeVisitor[None],
                 elif isinstance(node, TypeAlias) and node.no_args:
                     assert isinstance(node.target, ProperType)
                     if isinstance(node.target, Instance):
-                        node = node.target.type
-                        nextsym = node.get(part)
+                        nextsym = node.target.type.get(part)
                 else:
                     if isinstance(node, Var):
                         typ = get_proper_type(node.type)
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3621,7 +3621,6 @@ class SemanticAnalyzer(NodeVisitor[None],
                         type_info = self.type
             elif isinstance(base.node, TypeAlias) and base.node.no_args:
                 assert isinstance(base.node.target, ProperType)
-                # TODO: support chained aliases.
                 if isinstance(base.node.target, Instance):
                     type_info = base.node.target.type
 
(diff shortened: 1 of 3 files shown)


test-data/unit/check-literal.test
```diff
@@ -3149,3 +3149,18 @@ err_code = -TWO
...
+    class In:
+        class Inner:
+            pass
+
+O = Out
+I = Out.In
+OI = O.In
+x: O.In
+y: I.Inner
+z: OI.Inner
```
ilevkivskyi: Please add a `reveal_type()` for all three variables. (Also please keep newline at the end of file.)

test-data/unit/check-literal.test
```diff
@@ -3149,3 +3149,18 @@ err_code = -TWO
 if bool():
     err_code = -THREE
 [builtins fixtures/float.pyi]
+
+[case testAliasForEnumTypeAsLiteral]
```
ilevkivskyi: Could you please actually add a test with enum and literal (as originally reported in the issue)?

mypy/semanal.py
```diff
@@ -3969,6 +3969,12 @@ def lookup_qualified(self, name: str, ctx: Context,
                     namespace = node.fullname
                 elif isinstance(node, PlaceholderNode):
                     return sym
+                elif (isinstance(node, TypeAlias) and
+                      isinstance(get_proper_type(node.target), Instance)):
```
ilevkivskyi: I think this should be supported only for `no_args` aliases, i.e. you can almost literally copy the code from `visit_member_expr()`. Btw, it seems to me the TODO there is no more relevant, you can also try adding a test for situation like `A = Out; B = A; x: B.In`.

TH3CHARLie:
Updated:
- `reveal_type()` for all declared vars
- rename original test to  `testAliasAsLiteral` (I forgot to change the name when I decided to exclude the enum test)
- add enum testcase `testAliasForEnumTypeAsLiteral`
- update `lookup_qualified` logic, remove TODO in `visit_member_expr`

test-data/unit/check-literal.test
```diff
@@ -3149,3 +3149,40 @@ err_code = -TWO
 if bool():
     err_code = -THREE
 [builtins fixtures/float.pyi]
+
+[case testAliasAsLiteral]
```
ilevkivskyi:
```suggestion
[case testNestedClassOnAliasAsType]
```

mypy/semanal.py
```diff
@@ -3969,6 +3968,11 @@ def lookup_qualified(self, name: str, ctx: Context,
                     namespace = node.fullname
                 elif isinstance(node, PlaceholderNode):
                     return sym
+                elif isinstance(node, TypeAlias) and node.no_args:
+                    assert isinstance(node.target, ProperType)
+                    if isinstance(node.target, Instance):
+                        node = node.target.type
+                        nextsym = node.get(part)
```
ilevkivskyi: I don't think you need to introduce the temporary `node` here.

test-data/unit/check-literal.test
```diff
@@ -3149,3 +3149,40 @@ err_code = -TWO
 if bool():
     err_code = -THREE
 [builtins fixtures/float.pyi]
+
+[case testNestedClassOnAliasAsType]
+from typing_extensions import Literal
```
ilevkivskyi: Actually one more thing, sorry for not noticing this before. This import is unused, and the test case can be moved to `check-type-aliases.test`.

TH3CHARLie: Wow, I didn't notice that unused Literal import either, great catch!

PR #8017, opened 2019-11-26
[suggest] Fix crash when refining type with unresolved import type name

msullivan:
This would trip an assert in the `__init__` of AnyType. I fixed the
crash by properliy handling it.

### DIFF
--- a/mypy/suggestions.py
+++ b/mypy/suggestions.py
@@ -937,7 +937,7 @@ def refine_type(ti: Type, si: Type) -> Type:
     s = get_proper_type(si)
 
     if isinstance(t, AnyType):
-        # If s is also an any, we return t in case it is a missing_import any
+        # If s is also an Any, we return t in case it is a missing_import Any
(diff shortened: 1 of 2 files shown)


mypy/suggestions.py
```diff
@@ -928,7 +937,8 @@ def refine_type(ti: Type, si: Type) -> Type:
     s = get_proper_type(si)
 
     if isinstance(t, AnyType):
-        return s
+        # If s is also an any, we return t in case it is a missing_import any
```
ilevkivskyi:
```suggestion
        # If s is also an Any, we return t in case it is a missing_import Any
```

mypy/suggestions.py
```diff
@@ -928,7 +937,8 @@ def refine_type(ti: Type, si: Type) -> Type:
...
+        return s if not isinstance(s, AnyType) else t
```
ilevkivskyi: But what if `t` was a "plain" `Any` and `s` is a missing import? Is such case possible?

PR #8019, opened 2019-11-27
Fix incorrect type inference for inherited from Any class

### LINKED ISSUE #8001: Float multiplication of numpy.ndarray in lambda is incorrectly analysed

Bug.

```python
from typing import NewType, Tuple, Deque

import numpy as np


class Point(np.ndarray):
    def __new__(cls, x: float, y: float) -> np.ndarray:
        return np.array((x, y), dtype=np.float32)


Nose = NewType("Nose", Tuple[Point, ...])


def average_position(positions: Deque) -> Nose:
    factor = 1.0 / len(positions)
    point_sum = [Point(0.0, 0.0) for _ in range(10)]
    for nose in positions:
        for i, point in enumerate(nose):
            point_sum[i] += point
    avg_pos = Nose(tuple(map(lambda p: factor * p, point_sum)))
    return avg_pos


def average_position_with_multiply(positions: Deque) -> Nose:
    factor = 1.0 / len(positions)
    point_sum = [Point(0.0, 0.0) for _ in range(10)]
    for nose in positions:
        for i, point in enumerate(nose):
            point_sum[i] += point
    avg_pos = Nose(tuple(map(lambda p: np.multiply(factor, p), point_sum)))
    return avg_pos

```

Mypy gives the errors below:
```python
bug.py:20: error: Argument 1 to "map" has incompatible type "Callable[[Point], float]"; expected "Callable[[Point], Point]"
bug.py:20: error: Incompatible return value type (got "float", expected "Point")

```

In the lambda in `average_position()`, mypy incorrectly (?) determines the type as `Callable[[Point], float]`, while using the NumPy function `np.multiply()` (which does the same thing) gives no error. Both functions output the same correct value.

There should be no error.

```plain
mypy==0.740
Python 3.8.0
```

```ini
[mypy]
python_version = 3.8

[mypy-cv2,dlib,numpy]
ignore_missing_imports = True
```

I am unsure if this is actually an error, or if I've done something wrong somewhere, but both functions run and output correctly.

TH3CHARLie:
resolves #8001


mypy/checkexpr.py
```diff
@@ -2360,6 +2360,12 @@ def lookup_definer(typ: Instance, attr_name: str) -> Optional[str]:
             any_type = AnyType(TypeOfAny.from_another_any, source_any=right_type)
             return any_type, any_type
 
+        # If either the LHS or the RHS is Instance and fallbacks to Any, we also return Any
+        if ((isinstance(left_type, Instance) and left_type.type.fallback_to_any) or
+                (isinstance(right_type, Instance) and right_type.type.fallback_to_any)):
+            any_type = AnyType(TypeOfAny.special_form)
+            return any_type, any_type
```
ilevkivskyi: Honestly this looks a bit too radical. But I am not a big expert here, @Michael0x2a maybe you can review this (feel free to un-assign if you don't have time)?

tavianator: I think this would be fine only after the below checks for the presence of `__op__`/`__rop__`.  E.g. this should still infer a precise type:

```
from typing import Any
C: Any
class D(C):
    def __add__(self, rhs: float) -> D:
        return self

reveal_type(D() + 0.5)  # N: Revealed type is 'D'
```

mypy/checkexpr.py
```diff
@@ -2451,6 +2451,14 @@ def lookup_definer(typ: Instance, attr_name: str) -> Optional[str]:
             else:
                 return result
 
+        # We finish invoking above operators and no early return happens. Therefore,
+        # we check if either the LHS or the RHS is Instance and fallbacks to Any,
+        # if so, we also return Any
+        if ((isinstance(left_type, Instance) and left_type.type.fallback_to_any) or
+                (isinstance(right_type, Instance) and right_type.type.fallback_to_any)):
+            any_type = AnyType(TypeOfAny.special_form)
```
msullivan: I'm not sure if `special_form` is the right thing here but I'm not totally sure what else to do. `from_another_any` is correct but digging out the other any is almost certainly not worth bothering.

TH3CHARLie: I am also not sure about whether `special_form` is the right thing to go, but clearly directly using `from_another_any` with either `left_type` or `right_type` here is incorrect. To proper use `from_another_any`, we need a better way to transform the Instance to an Any

msullivan:
Oh, I guess please fix the merge conflict

TH3CHARLie:
> Oh, I guess please fix the merge conflict

fixed

PR #8023, opened 2019-11-27
Recommend typing_extensions instead of mypy_extensions for TypedDict

ilevkivskyi:
Now that `TypedDict` is an official feature I think this would make sense.


ilevkivskyi:
> What about also documenting that in Python 3.8 `TypedDict` can be imported from typing?

OK, I will add a short note about this (Python 3.8 is still new so I think we can use `typing_extensions` in the examples).

PR #8026, opened 2019-11-27
Treat --python-version X.Y and --python-version=X.Y in tests equally

ilevkivskyi:
While working on https://github.com/python/mypy/pull/8024 I noticed that `# flags: --python-version X.Y` and `# flags: --python-version=X.Y` are not treated equally by the test harness, namely the latter is actually ignored. This PR should fix this, because both forms are actually used in tests.

@Michael0x2a there is some suspicious change in literal types, is it expected?


ilevkivskyi:
OK, I opened https://github.com/python/mypy/issues/8062 to track the stub unicode literal alias issue.

Issue #7724, opened 2019-10-16
Refactor all member access to go through checkmember.py (meta issue)


ilevkivskyi:
There are bunch of places where we duplicate some parts of code for special logic in member access (such as properties, class methods, descriptors, `__getattr__()`, etc):
* The `checkmember.py` itself
* Override checks for variables
* Override checks for methods
* Multiple inheritance checks
* Checks for special methods via `check_op()` and `has_member()`
* Protocol implementation check

Here are some issues that will be solved (or significantly simplified) by refactoring this to go through the same place (probably `checkmember.py`):
* https://github.com/python/mypy/issues/4125
* https://github.com/python/mypy/issues/5425
* https://github.com/python/mypy/issues/5136
* https://github.com/python/mypy/issues/3832
* https://github.com/python/mypy/issues/7565
* https://github.com/python/mypy/issues/7153
* https://github.com/python/mypy/issues/5836
* https://github.com/python/mypy/issues/5803
* https://github.com/python/mypy/issues/5491

Also I think having member access cleaned-up, centralized, and documented will open the way to finally solving https://github.com/python/mypy/issues/708 (currently the oldest high priority issue), and to having general support for descriptor protocol (for historical reasons, several things like properties are supported via some special-casing).

sobolevn:
Plugin management is also really important here, because right now protocols do not support objects with custom `get_attribute_hook` / `get_method_sig_hook` items. Quick example:

```python
# ex.py
class A(object):
    attr: str  # is set via plugin

    def method(self) -> str:  # is set via plugin
        ...

a = A()
reveal_type(a.attr)
# Revealed type is "builtins.float"
reveal_type(a.method())
# Revealed type is "builtins.int"
```

Plugin:

```python
from mypy.plugin import Plugin

class MyPlugin(Plugin):
    def get_attribute_hook(self, fullname):
        if fullname == 'ex.A.attr':
            def test(ctx):
                return ctx.api.named_type('builtins.float')
            return test
    def get_method_signature_hook(
        self,
        fullname: str,
    ) -> Optional[Callable[[MethodSigContext], CallableType]]:
        """Here we fix the calling method types to accept only valid types."""
        if fullname == 'ex.A.method':
            def test(ctx):
                return ctx.default_signature.copy_modified(
                    ret_type=ctx.api.named_type('builtins.int'),
                )
            return test
```

sobolevn:
[...]

PR #8030, opened 2019-11-28
Apply generic class fix also to non-callable types

ilevkivskyi:
This is a follow up for https://github.com/python/mypy/pull/8021, that applies the fix also to non-callable types. Currently non-callable types are still wrong:
```python
R = TypeVar('R')
T = TypeVar('T')
# Can be any decorator that makes type non-callable.
def classproperty(f: Callable[..., R]) -> R: ...

class C(Generic[T]):
    @classproperty
    def test(self) -> T: ...

x: C[int]
y: Type[C[int]]
reveal_type(x.test)  # Revealed type is 'int', OK
reveal_type(y.test)  # Revealed type is 'T' ???
```

So, https://github.com/python/mypy/issues/7724 strikes again. It turns out there is not only duplicated logic for attribute kinds (decorators vs normal methods), but also for callable vs non-callable types. In latter case we still need to expand the type (like in other places, e.g., `analyze_var`).

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -855,7 +855,9 @@ def add_class_tvars(t: ProperType, itype: Instance, isuper: Optional[Instance],
                                                               builtin_type, original_type,
                                                               original_vars=original_vars))
                            for item in t.items()])
-    return expand_type_by_instance(t, isuper)
+    if isuper is not None:
+        t = cast(ProperType, expand_type_by_instance(t, isuper))
+    return t
 
 
 def type_object_type(info: TypeInfo, builtin_type: Callable[[str], Instance]) -> ProperType:
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -823,13 +823,13 @@ def add_class_tvars(t: ProperType, isuper: Optional[Instance],
     B.foo()
 
     Args:
-        t: Declared type of the method (or property);
+        t: Declared type of the method (or property)
         isuper: Current instance mapped to the superclass where method was defined, this
-            is usually done by map_instance_to_supertype();
-        is_classmethod: True if this method is decorated with @classmethod;
+            is usually done by map_instance_to_supertype()
+        is_classmethod: True if this method is decorated with @classmethod
         original_type: The value of the type B in the expression B.foo() or the corresponding
-            component in case of a union (this is used to bind the self-types);
-        original_vars: Type variables of the class callable on which the method was accessed.
+            component in case of a union (this is used to bind the self-types)
+        original_vars: Type variables of the class callable on which the method was accessed
     Returns:
         Expanded method type with added type variables (when needed).
     """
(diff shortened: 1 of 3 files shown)


mypy/checkmember.py
```diff
@@ -855,6 +855,8 @@ class B(A[str]): pass
                                                               builtin_type, original_type,
                                                               original_vars=original_vars))
                            for item in t.items()])
+    if isuper is not None:
+        t = cast(ProperType, expand_type_by_instance(t, isuper))
```
JukkaL: It would be nice to have a comment here, or maybe add a case to the above comment where the type is not callable.

Also, I have trouble convincing myself how we can be sure that the type arguments from `isuper` can always be applied to `t`. Maybe explain this is briefly (in the above comment perhaps, since this doesn't seem specific to this case).

test-data/unit/check-generics.test
```diff
@@ -2337,3 +2337,50 @@ class Test():
...
+U = TypeVar('U')
+
+def classproperty(f: Callable[..., U]) -> U: ...
+
+class C(Generic[T]):
+    @classproperty
+    def test(self) -> T: ...
+
+class D(C[str]): ...
+class G(C[List[T]]): ...
```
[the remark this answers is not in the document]
ilevkivskyi: I can add few more tests, but essentially this will be testing `map_instance_to_supertype()` which is already quite well tested.

mypy/checkmember.py
ilevkivskyi: > Also, I have trouble convincing myself how we can be sure that the type arguments from `isuper` can always be applied to `t`.

I actually have trouble understanding how this can _not_ work :-) My guess is confusion comes from missing docstring entry for `isuper`, which is current instance mapped to the method definition superclass. I was also thinking about an assert, but this would be probably too much.

ilevkivskyi: Btw, while adding docstring items I noticed there are two unused arguments in this function.

mypy/checkmember.py
```diff
@@ -821,12 +820,18 @@ class A(Generic[T]):
...
 
     class B(A[str]): pass
-
     B.foo()
 
-    original_type is the value of the type B in the expression B.foo() or the corresponding
-    component in case if a union (this is used to bind the self-types); original_vars are type
-    variables of the class callable on which the method was accessed.
+    Args:
+        t: Declared type of the method (or property);
```
JukkaL: Style nit: We don't use `;` after each description of an argument in docstrings.

mypy/checkmember.py
JukkaL: Yeah, I was confused about `isuper`. Now it all makes sense.

test-data/unit/check-generics.test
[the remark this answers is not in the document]
JukkaL: Yes, the extra tests don't seem useful.

PR #8032, opened 2019-11-28
Improve daemon docs; add docs for dmypy suggest

ilevkivskyi:


docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
 options not discussed here, and ``dmypy <command> --help`` for help on
 command-specific options.
 
+Additional daemon flags
+***********************
+
+.. option:: --status-file FILE
+
+   Status file to retrieve daemon details. This is normally a JSON file
```
JukkaL: Suggested edit: "Use `FILE` as the status file for storing daemon runtime state. ..."

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+
+.. option:: --log-file FILE
+
+   Direct daemon stdout/stderr to ``FILE``. This is useful for debugging daemon
+   crashes, since the server traceback may be not printed to client stderr. Only
+   available for ``start``, ``restart``, and ``run`` commands.
+
+.. option:: --timeout TIMEOUT
+
+   Server shutdown timeout (in seconds). Only available for ``start``,
```
JukkaL: Suggested edit: Automatically shut down server after ``TIMEOUT`` seconds of inactivity.

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+   ``restart``, and ``run`` commands.
+
+.. option:: --fswatcher-dump-file FILE
+
+   Collect information about the current file state. Only available for
+   ``status`` command.
```
JukkaL: I think that this description is not detailed enough to be useful. I think that it's fine to leave this undocumented for now. Altarnatively, at least the purpose of this information and the structure/contents of it should probably be documented.

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+
+.. option:: --perf-stats-file FILE
+
+   Write performance telemetry information to the given file. Only available
```
JukkaL: Nit: Tthis is not telemetry, since it's only written to a local file. 

Suggested edit: "Write performance profiling information to ``FILE``.

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+   for ``check``, ``recheck``, and ``run`` commands.
+
+.. option:: --update FILE
+
+   Files in the run to add or check again, may be repeated. Default: all
+   files from the previous run. Only available for ``recheck`` command.
```
JukkaL: Give some motivation for why somebody might want to use this option. Something about speeding up runs if there are thousands of files, and maybe suggest using file events to find the modified files.

Maybe also mention that this option is never required and is only available for speeding things up.

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+
+.. option:: --remove FILE
+
+   Files to remove from the run, may be repeated. Only available for
```
JukkaL: This suggests that this is needed whenever files are removed, which is not the case? Similar to above, document that this can be used as an optimization to avoid looking at all source files for deletions.

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+   ``recheck`` command.
+
+Static inference of annotations
+*******************************
+
+Mypy daemon supports (as experimental feature) statically inferring draft type
```
JukkaL: Grammar nits: "as **an** experimental feature"; "inferring **a** draft type"

It's probably worth mentioning that this is a low-level feature that is usually driven by other tools such as editor integrations. Maybe add link to the PyCharm plugin as an example.

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+annotation for a given function or method. For example, given this program:
```
JukkaL: Incomplete sentence "For example, ..." (no verb). Suggested edit: "In this example, the function ``format_id`` has no annotation:"

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+
+   root = format_id(0)
+
+Mypy can infer that ``format_id()`` takes an ``int`` and returns a ``str``.
+To get a suggested signature for a function, use ``dmypy suggest FUNCTION``,
+where the function may be specified in either of two forms:
+
+* By its fully qualified name, i.e. ``[package.]module.[class.]function``
+
+* By its textual location, i.e. ``/path/to/file.py:line``
```
JukkaL: Does a relevative path also work? I assume that the line can refer to any line within a function; maybe mention this as well.

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+
+   Files in the run to add or check again, may be repeated. Default: all
+   files from the previous run. Only available for ``recheck`` command.
+
+.. option:: --remove FILE
+
+   Files to remove from the run, may be repeated. Only available for
+   ``recheck`` command.
+
+Static inference of annotations
```
JukkaL: Maybe move this to a separate rst file, since this is quite distinct from other features provided by dmypy, and mostly relevant for users that are going to write integrations (since the low-level UX is kind of hard to use).

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+To get a suggested signature for a function, use ``dmypy suggest FUNCTION``,
+where the function may be specified in either of two forms:
+
+* By its fully qualified name, i.e. ``[package.]module.[class.]function``
+
+* By its textual location, i.e. ``/path/to/file.py:line``
+
+Running this command will produce a suggested signature in the format
+``(param_type_1, param_type_2, ...) -> ret_type``. This may be used by IDEs
+or similar tools to propose to user and/or insert the annotation into file.
```
JukkaL: As I mentioned earlier, this information is key and should probably be presented earlier.

Maybe discuss what happens to arguments with default values, `*args` and `**kwargs`.

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+
+This command can also be used to find an improvement for an existing (imprecise)
+annotation. The following flags customize various aspects of the ``dmypy suggest``
+command.
+
+.. option:: --json
+
+   Use JSON format to output the signature, so that `PyAnnotate`_ can use it
```
JukkaL: Suggested edit: "Output the signature as JSON, so that ..."

Maybe also give an example of the JSON?

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+.. option:: --callsites
+
+   Only find call sites for a given function instead of suggesting a type.
+   This will produce a list including textual locations and types of actual
+   arguments for each call: ``/path/to/file.py:line: (arg_type_1, arg_type_2, ...)``.
+
+.. option:: --use-fixme NAME
+
+   A dummy name to use instead of plain ``Any`` for types that cannot
+   be inferred.
```
JukkaL: Maybe add some motivation for this feature.

docs/source/mypy_daemon.rst
```diff
@@ -94,6 +94,118 @@ Use ``dmypy --help`` for help on additional commands and command-line
...
+annotation for a given function or method. For example, given this program:
+
+.. code-block:: python
+
+   def format_id(user):
+       return "User: {}".format(user)
+
+   root = format_id(0)
+
+Mypy can infer that ``format_id()`` takes an ``int`` and returns a ``str``.
```
JukkaL: Maybe explain the algorithm in little detail, such as mentioning that mypy looks at call sites and the function definition to infer types.

docs/source/mypy_daemon.rst
ilevkivskyi: I think we can keep it in the same file for now. It is anyway at the very end and it is probably OK to attract some attention to it.

Also I can't find a good name for a section that would look good in table of contents with the daemon. I think we can move it to a separate file when we will add `dmypy reveal`, then it will look more natural.

Issue #1055, opened 2015-12-06
Infer variable type if initializer is an empty collection


JukkaL:
Code like this is one of the most common cases where we need a variable annotation (**DONE**):

```
def f() -> None:
    x = []   # mypy insists on annotation (such as List[int]) here
    for i in range(5):
        x.append(i * 2)
```

Mypy could do a better job of inferring types in cases like these:
- Variable is initialized with an empty collection (or more generally, a generic type where the type argument values are indeterminate using only local statement context).
- The first reference to the variable after initialization (in the same scope) mutates the collection and allows us to infer the item types.

These cases would still require an annotation:
- Collection is read after initialization and before mutation (this happens sometimes but not too often).
- The variable is not mutated in the scope that initializes it. For example, an attribute is initialized in `__init__` and mutated in another method.

This is related to #254.

More common examples where this should work:

(1) **DONE**

```
x = {}
x['key'] = 2
```

(2) **DONE**

```
x = set()
x.add(2)
```

Additional, less common examples (not done):

(3) **[postponed -- too rare and complicated]**

```
x = {}
x.setdefault('key', []).append(1)   # this is harder than most examples
```

(4) **DONE**

```
x = {}
x.update({'x': 1})
```

(5) **[rejected -- too rare]**

```
x = set()
x |= {1, 2}
```

(6) **DONE** (#8036)

```
if cond():
    x = []
else:
    x = [1]
```

(7) **DONE** (#8167 and #8039)

Similar to above but with a user-defined generic type or a standard library generic type that is not list, dict or set.

**Update:** After some more analysis, only `defaultdict` and `OrderedDict` seem common enough to support.

Not sure about this:

(8) **[rejected idea]**

```
x = []
foo(x)  # Should we infer type of x from the argument type? Probably not.
```

kirbyfan64:
FWIW, this can get slow really quickly: http://crystal-lang.org/2013/11/14/good-bye-ruby-thursday.html#compile-times-were-too-big

JukkaL:
We'll use a simpler form of type inference that doesn't slow things much (if at all). We aren't going to do whole-program type inference.

JukkaL:
Here's the general idea for how this would work:

* If we detect an assignment such as `x = []` (and we haven't seen `x` before) we'd initially give `x` a 'partial' type `List[?]`.
* When you perform a mutation operation such as `append`, we infer the missing bits for the type and replace the type with a precise type (for example, `List[int]`).
[...]

PR #8036, opened 2019-11-29
Infer type for partial generic type from assignment

JukkaL:
Code like this no longer requires a type annotation:

```
a = []
if foo():
    a = [1]
```

Work towards #1055.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2042,6 +2042,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             self.check_assignment_to_multiple_lvalues(lvalue.items, rvalue, rvalue,
                                                       infer_lvalue_type)
         else:
+            self.try_infer_partial_generic_type_from_assignment(lvalue, rvalue)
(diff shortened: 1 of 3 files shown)


mypy/checker.py
```diff
@@ -2141,6 +2142,37 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
...
+                                                       rvalue: Expression) -> None:
+        """Try to infer a precise type for partial generic type from assignment.
+
+        Example where this happens:
+
+            x = []
+            if foo():
+                x = [1]  # Infer List[int] as type of 'x'
+        """
+        if (isinstance(lvalue, NameExpr)
```
ilevkivskyi: Should this actually be a `NameExpr` or `RefExpr` is fine? I am thinking about `self.x = []`, please add tests if yes.

JukkaL: Partial types are not properly supported with attributes in general, so this is consistent with how things work elsewhere.

Issue #8043, opened 2019-11-30
Refactor parial types to avoid code duplication


ilevkivskyi:
Currently there is some duplication around partial types:
* Logic in `try_infer_partial_generic_type_from_assignment()` essentially duplicates the logic inlined in `check_assignment()` for `None` partial types. These two probably can be refactored to the same method if we also update `handle_partial_var_type()` to not special-case partial `None` types, and instead consistently return a partial type in lvalue context.
* Logic in `try_infer_partial_type_from_indexed_assignment()` duplicates that in `try_infer_partial_type()` (there is already a TODO item). This one can be refactored by either pushing the latter a bit down the call stack (closer to `check_call()`), or by generating a synthetic `CallExpr` with `__setitem__` and passing it to `try_infer_partial_type()`.

The second item may be something to watch out when implementing support for these:
```python
a = defaultdict(list)
a[0].append('yes')

b = {}
b.setdefault(0, []).append('yes')

c = defaultdict(set)
c[0].add('yes')

d = {}
d.setdefault(0, set()).add('yes')
```

@JukkaL this is probably not something important, but maybe it makes sense to fix this while we are at it?

JukkaL:
These refactorings sounds like good ideas. I'd prefer not creating temporary node objects if possible, as they tend to increase complexity and sometimes cause trouble with error location reporting.

LeoLomardo:
Hi, i just started trying to contribute on open source projects, and i thought mypy would be a great project to start. I preferred to start working on refactoring issues and i saw this one. I had a question related to `import typing_extensions`, does mypy still support python <=3.9? its a good refactoring to update this to `typing` ? Can i work on this issue? 
One example is on the file `/mypy/misc/incremental_checker.py` that i changed to do this : 

`
    from typing import Any, Final, Dict
    
    try:
    
      from typing import TypeAlias as _TypeAlias
    
    except ImportError:  
    
      from typing_extensions import TypeAlias as _TypeAlias
 `

PR #8044, opened 2019-11-30
Support generic partial types for attributes

ilevkivskyi:
Work towards https://github.com/python/mypy/issues/1055

Currently partial types are supported for local variables. However, only partial `None` types are supported for `self` attributes. This PR adds the same level of support to generic partial types. They follow mostly the same rules:
* A partial type can be refined in the _same_ method where it is defined.
* But a partial type from class body can not be refined in a method, as if `local_partial_types = True`.

The logic is pretty simple: the `.node` attribute for `self.attr` expressions is set to `None`, so I added a little helper to get it from the class symbol table instead.

I add tests for various scenarios, including for partial `None` types. It may happen they are already tested elsewhere, but this way it is easy to compare side-by-side behavior is consistent for both kinds of partial types. I can remove them if you think they are redundant.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -537,6 +537,19 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
         return callee
 
+    def get_partial_self_var(self, expr: MemberExpr) -> Optional[Var]:
+        if not (isinstance(expr.expr, NameExpr) and
+                isinstance(expr.expr.node, Var) and expr.expr.node.is_self):
+            return None
+        info = self.chk.scope.enclosing_class()
+        if not info or expr.name not in info.names:
+            return None
+        sym = info.names[expr.name]
+        # TODO: check implicit (add tests for defined in class body, both orders)?
+        if isinstance(sym.node, Var) and isinstance(sym.node.type, PartialType):
+            return sym.node
+        return None
+
     # Types and methods that can be used to infer partial types.
     item_args = {'builtins.list': ['append'],
                  'builtins.set': ['add', 'discard'],
@@ -550,6 +563,8 @@ class ExpressionChecker(ExpressionVisitor[Type]):
     def try_infer_partial_type(self, e: CallExpr) -> None:
         if isinstance(e.callee, MemberExpr) and isinstance(e.callee.expr, RefExpr):
             var = e.callee.expr.node
+            if var is None and isinstance(e.callee.expr, MemberExpr):
+                var = self.get_partial_self_var(e.callee.expr)
             if not isinstance(var, Var):
                 return
             partial_types = self.chk.find_partial_types(var)
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -538,14 +538,20 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         return callee
 
     def get_partial_self_var(self, expr: MemberExpr) -> Optional[Var]:
+        """Get variable node for a partial self attribute.
+
+        If the expression is not a self attribute, or attribute is not variable,
+        or variable is not partial, return None.
+        """
         if not (isinstance(expr.expr, NameExpr) and
                 isinstance(expr.expr.node, Var) and expr.expr.node.is_self):
+            # Not a self.attr expression.
             return None
         info = self.chk.scope.enclosing_class()
         if not info or expr.name not in info.names:
+            # Don't mess with partial types in superclasses.
             return None
         sym = info.names[expr.name]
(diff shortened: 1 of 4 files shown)


test-data/unit/check-inference.test
```diff
@@ -1585,6 +1585,145 @@ oo.update(d)
...
+    def __init__(self) -> None:
+        self.a.append(1)
+reveal_type(C().a)  # N: Revealed type is 'builtins.list[builtins.int]'
+[builtins fixtures/list.pyi]
+
+[case testInferAttributeInitializedToEmptyAndAssignedItemClassBody]
+class C:
+    a = {}
+    def __init__(self) -> None:
+        self.a[0] = 'yes'
```
JukkaL: Similar to above.

test-data/unit/check-inference.test
```diff
@@ -1585,6 +1585,145 @@ oo.update(d)
...
+        self.a = None
+    def meth(self) -> None:
+        self.a = 1  # E: Incompatible types in assignment (expression has type "int", variable has type "None")
+reveal_type(C().a)  # N: Revealed type is 'None'
+
+[case testInferAttributeInitializedToEmptyAndAssignedClassBody]
+class C:
+    a = []
+    def __init__(self) -> None:
+        self.a = [1]
```
JukkaL: I'd rather not add support for this, since I want to deprecate not using `--local-partial-types` in the not too distant future (to make daemon and non-daemon modes uniform). If we further increase the semantic differences of `--local-partial-types`, the deprecation is going to be more painful.

test-data/unit/check-inference.test
```diff
@@ -1585,6 +1585,145 @@ oo.update(d)
...
+reveal_type(C().a)  # N: Revealed type is 'builtins.list[builtins.int*]'
+[builtins fixtures/list.pyi]
+
+[case testInferAttributeInitializedToEmptyAndAppendedClassBody]
+class C:
+    a = []
+    def __init__(self) -> None:
+        self.a.append(1)
```
JukkaL: Similar to above.

test-data/unit/check-inference.test
```diff
@@ -1585,6 +1585,145 @@ oo.update(d)
...
+        self.a[0] = 'yes'
+reveal_type(C().a)  # N: Revealed type is 'builtins.dict[builtins.int, builtins.str]'
+[builtins fixtures/dict.pyi]
+
+[case testInferAttributeInitializedToNoneAndAssignedClassBody]
+# flags: --strict-optional
+class C:
+    a = None
+    def __init__(self) -> None:
+        self.a = 1
```
JukkaL: This worked before this PR already, so this needs to be supported for now.

test-data/unit/fine-grained.test
```diff
@@ -2714,6 +2714,7 @@ class C:
         class D:
             def __init__(self) -> None:
                 self.x = {}
+            def meth(self) -> None:
```
JukkaL: What about also testing that the type inference works correctly in fine-grained mode (within one method)?

test-data/unit/check-inference.test
```diff
@@ -1585,6 +1585,145 @@ oo.update(d)
 reveal_type(oo) # N: Revealed type is 'collections.OrderedDict[builtins.int*, builtins.str*]'
 [builtins fixtures/dict.pyi]
 
+[case testInferAttributeInitializedToEmptyAndAssigned]
+class C:
+    def __init__(self) -> None:
+        self.a = []
+        if bool():
+            self.a = [1]
```
JukkaL: Test using a non-self attribute -- it shouldn't affect type inferece. For example:

```py
...
self.a = []
if bool():
    a = self
    a.a = [1]  # Should not infer type
    a.append(1)  # Should not infer type
```

mypy/checkexpr.py
```diff
@@ -537,6 +537,25 @@ def check_typeddict_call_with_kwargs(self, callee: TypedDictType,
 
         return callee
 
+    def get_partial_self_var(self, expr: MemberExpr) -> Optional[Var]:
+        """Get variable node for a partial self attribute.
+
+        If the expression is not a self attribute, or attribute is not variable,
+        or variable is not partial, return None.
+        """
+        if not (isinstance(expr.expr, NameExpr) and
```
JukkaL: Does this duplicate any logic used for `None` partial type handling? Could they be merged?

ilevkivskyi: > Does this duplicate any logic used for `None` partial type handling? Could they be merged?

Short answers are no and no. Some things may be however simplified after (or while) fixing https://github.com/python/mypy/issues/8043 (at least the first bullet).

PR #8063, opened 2019-12-03
Better error message when using pyd file as stubgen

### LINKED ISSUE #8035: stubgen pyd file utf-8 error
Hello,

I have a file `_core.pyd` that comes from [QGIS](http://www.qgis.org) installation (file enclosed) that I can't use with stubgen:

```
Critical error during semantic analysis: mypy: can't decode file '_core.pyd': 'utf-8' codec can't decode byte 0x90 in position 2: invalid start byte
```
[_core.zip](https://github.com/python/mypy/files/3905454/_core.zip)

TH3CHARLie:
resolves #8035

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2004,9 +2004,14 @@ class State:
                             self.path, os.strerror(ioerr.errno))],
                         module_with_blocker=self.id)
                 except (UnicodeDecodeError, DecodeError) as decodeerr:
-                    raise CompileError([
-                        "mypy: can't decode file '{}': {}".format(self.path, str(decodeerr))],
-                        module_with_blocker=self.id)
+                    if self.path.endswith('.pyd'):
+                        raise CompileError([
+                            "mypy: no support for .pyd files: '{}'".format(self.path)],
+                            module_with_blocker=self.id)
+                    else:
+                        raise CompileError([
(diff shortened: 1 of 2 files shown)


mypy/build.py
```diff
@@ -2004,9 +2004,14 @@ def parse_file(self) -> None:
                             self.path, os.strerror(ioerr.errno))],
                         module_with_blocker=self.id)
                 except (UnicodeDecodeError, DecodeError) as decodeerr:
-                    raise CompileError([
-                        "mypy: can't decode file '{}': {}".format(self.path, str(decodeerr))],
-                        module_with_blocker=self.id)
+                    if self.path.endswith('.pyd'):
```
ethanhs: I think it would be less repetitive to have the string selected based on the file type and raise the error at the end:
```python
                    if self.path.endswith('.pyd'):
                        err = "mypy: stubgen does not support .pyd files: '{}'".format(self.path)
                    else:
                        err = "mypy: can't decode file '{}': {}".format(self.path, str(decodeerr))
                    raise CompileError([err], module_with_blocker=self.id)
```

TH3CHARLie: very reasonable

TH3CHARLie:
Thanks for the review

PR #8066, opened 2019-12-03
Stubgen: split @abstractproperty

### LINKED ISSUE #7952: stubgen: split @abstractproperty
From #7921: stubgen should split `@abstractproperty` into `@property @abstractmethod`, since the former has been deprecated since Python 3.3.

TH3CHARLie:
resolves #7952

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -647,10 +647,13 @@ class StubGenerator(mypy.traverser.TraverserVisitor):
                                                           'asyncio.coroutines',
                                                           'types'):
             self.add_coroutine_decorator(context.func, name, name)
-        elif any(self.refers_to_fullname(name, target)
-                 for target in ('abc.abstractmethod', 'abc.abstractproperty')):
+        elif self.refers_to_fullname(name, 'abc.abstractmethod'):
             self.add_decorator(name)
             self.import_tracker.require_name(name)
+            is_abstract = True
+        elif self.refers_to_fullname(name, 'abc.abstractproperty'):
+            self.add_decorator('property')
+            self.add_decorator('abc.abstractmethod')
             is_abstract = True
         return is_abstract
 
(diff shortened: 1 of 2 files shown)


mypy/stubgen.py
```diff
@@ -647,11 +647,14 @@ def process_name_expr_decorator(self, expr: NameExpr, context: Decorator) -> boo
...
-        elif any(self.refers_to_fullname(name, target)
-                 for target in ('abc.abstractmethod', 'abc.abstractproperty')):
+        elif self.refers_to_fullname(name, 'abc.abstractmethod'):
             self.add_decorator(name)
             self.import_tracker.require_name(name)
             is_abstract = True
+        elif self.refers_to_fullname(name, 'abc.abstractproperty'):
+            self.add_decorator('property')
+            self.add_decorator('abc.abstractmethod')
+            is_abstract = True
```
TH3CHARLie: I'd like to be more specific about here.
I intentionally did not require_name in import_tracker. This is because `is_abstract` will be set and `abc` will be imported, so directly add `abc.abstractmethod` would work. And as shown in testcase: `testSplitAbstractPropertyUsingAlias`, if the imported `abstractproperty` is aliased, to properly split would have to modify code in import_tracker, which IMO is not necessary since using `abc.abstractmethod` already achieves purposes

TH3CHARLie: I'd like to be more specific about here. (The previous commit has incorrect tests)
I intentionally did not require_name in import_tracker. This is because `is_abstract` will be set and `abc` will be imported, so directly add `abc.abstractmethod` would work. And as shown in testcase: `testAbstractProperty3_semanal`, if the imported abstractproperty is aliased, to properly split would have to modify code in import_tracker, which IMO is not necessary since using `abc.abstractmethod` already achieves purposes

TH3CHARLie:
@JukkaL Would you please review this since it may be a followup of #7921

PR #8067, opened 2019-12-03
Stubgen: normalize input paths when finding common parent directory for out files

### LINKED ISSUE #8065: Stubgen stuck in an endless loop in 0.750
* Are you reporting a bug, or opening a feature request? **Bug**

Only reproducible on Windows b/c of different path separator. On a windows machine, issue:

```cmd
> mkdir foo bar
> touch foo\__init__.py foo\spam.py bar\__init__.py bar\eggs.py
> stubgen foo\spam.py bar\eggs.py
```
Result: `stubgen` never finishes, because caught in the endless while loop: `out\foo/` doesn't start with `out\bar/`, then doesn't start with `out/`, then with `/` repeatedly.
https://github.com/python/mypy/blob/4e021d908a8298c99c280f9cef20534c22c77d42/mypy/stubutil.py#L245-L249

Since `mypy` supports only Python 3.5 and onwards anyway - may I suggest using `pathlib` instead of string manipulation? With `pathlib`, the task of finding the common parent directory is a walk in the park and works on all OSes. For example, the `common_dir_prefix` method could look like this:

```py
from pathlib import Path

def common_dir_prefix(paths: List[str]) -> str:
    if not paths:
        return '.'
    cur = Path(paths[0])
    for parent in cur.parents:
        # if the parent is parent of all paths, we're good to go
        if all(parent in Path(path).parents for path in paths[1:]):
            return str(parent)
    return '.'
```
In addition, if an extensive path manipulation is done in `mypy` (didn't check that), I would also plead switching to `pathlib` exclusively.

hoefling:
Close #8065.

### DIFF
--- a/mypy/stubutil.py
+++ b/mypy/stubutil.py
@@ -2,13 +2,12 @@
 
 import sys
 import os.path
-import itertools
 import json
 import subprocess
 import re
 from contextlib import contextmanager
 
-from typing import Optional, Tuple, List, Iterator, Sequence, Union
+from typing import Optional, Tuple, List, Iterator, Union
 from typing_extensions import overload
 
 from mypy.moduleinspect import ModuleInspect, InspectError
@@ -249,10 +248,11 @@ def common_dir_prefix(paths: List[str]) -> str:
     if not paths:
         return '.'
 
-    def common_part(parts: Sequence[str]) -> bool:
-        return all(x == parts[0] for x in parts)
-
-    dirs = [os.path.dirname(p) for p in paths]
-    parts = zip(*(path.split(os.path.sep) for path in dirs))
-    common_parts = (p[0] for p in itertools.takewhile(common_part, parts))
-    return os.path.join(*common_parts) or '.'
+    cur = os.path.dirname(os.path.normpath(paths[0]))
+    for path in paths[1:]:
+        while True:
+            path = os.path.dirname(os.path.normpath(path))
+            if (cur + os.sep).startswith(path + os.sep):
+                cur = path
+                break
+    return cur or '.'
--- a/mypy/stubutil.py
+++ b/mypy/stubutil.py
@@ -2,11 +2,11 @@
 
 import sys
 import os.path
+import itertools
 import json
 import subprocess
 import re
 from contextlib import contextmanager
-from pathlib import Path
 
 from typing import Optional, Tuple, List, Iterator, Union
 from typing_extensions import overload
@@ -248,8 +248,11 @@ def remove_misplaced_type_comments(source: Union[str, bytes]) -> Union[str, byte
 def common_dir_prefix(paths: List[str]) -> str:
     if not paths:
         return '.'
-    cur = Path(paths[0])
-    for parent in cur.parents:
-        if all(parent in Path(path).parents for path in paths[1:]):
-            return str(parent)
-    return '.'
+
+    def common_part(parts):
+        return all(x == parts[0] for x in parts)
+
+    dirs = [os.path.dirname(p) for p in paths]
+    parts = zip(*(path.split(os.path.sep) for path in dirs))
+    common_parts = (p[0] for p in itertools.takewhile(common_part, parts))
+    return os.path.join(*common_parts) or '.'
--- a/mypy/stubutil.py
+++ b/mypy/stubutil.py
@@ -247,7 +247,6 @@ def remove_misplaced_type_comments(source: Union[str, bytes]) -> Union[str, byte
 def common_dir_prefix(paths: List[str]) -> str:
     if not paths:
         return '.'
-
     cur = os.path.dirname(os.path.normpath(paths[0]))
     for path in paths[1:]:
         while True:
--- a/mypy/stubutil.py
+++ b/mypy/stubutil.py
@@ -6,6 +6,7 @@ import json
 import subprocess
 import re
 from contextlib import contextmanager
+from pathlib import Path
 
 from typing import Optional, Tuple, List, Iterator, Union
 from typing_extensions import overload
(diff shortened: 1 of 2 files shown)


mypy/stubutil.py
```diff
@@ -240,11 +241,8 @@ def remove_misplaced_type_comments(source: Union[str, bytes]) -> Union[str, byte
...
-            path = os.path.dirname(path)
-            if (cur + '/').startswith(path + '/'):
-                cur = path
-                break
-    return cur or '.'
+    cur = Path(paths[0])
+    for parent in cur.parents:
+        if all(parent in Path(path).parents for path in paths[1:]):
+            return str(parent)
+    return '.'
```
JukkaL: I'd rather not introduce code using `pathlib`, since there is a lot of code using `os.path` elsewhere. I know that `pathlib` has various improvements over `os.path`, but I think that consistency is more important. Could you implement the fix using `os.path`?

mypy/test/teststubgen.py
```diff
@@ -447,15 +449,15 @@ def test_common_dir_prefix(self) -> None:
...
-                                  'foo/bar/y.pyi']) == 'foo/bar'
+                                  'foo/bar/y.pyi']) == str(Path('foo/bar'))
         assert common_dir_prefix(['foo/bar/x.pyi', 'foo/y.pyi']) == 'foo'
         assert common_dir_prefix(['foo/x.pyi', 'foo/bar/y.pyi']) == 'foo'
         assert common_dir_prefix(['foo/bar/zar/x.pyi', 'foo/y.pyi']) == 'foo'
         assert common_dir_prefix(['foo/x.pyi', 'foo/bar/zar/y.pyi']) == 'foo'
-        assert common_dir_prefix(['foo/bar/zar/x.pyi', 'foo/bar/y.pyi']) == 'foo/bar'
-        assert common_dir_prefix(['foo/bar/x.pyi', 'foo/bar/zar/y.pyi']) == 'foo/bar'
+        assert common_dir_prefix(['foo/bar/zar/x.pyi', 'foo/bar/y.pyi']) == str(Path('foo/bar'))
+        assert common_dir_prefix(['foo/bar/x.pyi', 'foo/bar/zar/y.pyi']) == str(Path('foo/bar'))
```
JukkaL: Update test case to cover the case that triggered the endless loop?

mypy/stubutil.py
hoefling: Sure, can do. However, do you find the idea of switching to `pathlib` useful in general? If yes, I could replace all the `os.path` usages with appropriate `pathlib` calls throughout the codebase in a separate PR.

JukkaL: Thanks for the offer, but I'm worried that it might introduce subtle bugs since some of the path handling code is pretty delicate. (It looks like we have a few references to `pathlib` already, though.)

hoefling: Was worth a shot :-)

mypy/test/teststubgen.py
hoefling: The case is already covered, but only for Unix paths. The best solution I came with is to add another test case that does the tests with windows paths and runs on windows only (controlled via `pytest` marks). Not sure whether you agree on this as now both tests `test_common_dir_prefix_unix` and `test_common_dir_prefix_windows` mostly duplicate each other, the only difference being the separator used. I have tried monkeypatching to mimick windows behaviour, but found it to be pretty tedious to do and introduce a lot of boilerplate code that also doesn't add to test code readability.

mypy/stubutil.py
```diff
@@ -240,11 +241,11 @@ def remove_misplaced_type_comments(source: Union[str, bytes]) -> Union[str, byte
...
-                break
-    return cur or '.'
+
+    def common_part(parts: Sequence[str]) -> bool:
+        return all(x == parts[0] for x in parts)
+
+    dirs = [os.path.dirname(p) for p in paths]
+    parts = zip(*(path.split(os.path.sep) for path in dirs))
+    common_parts = (p[0] for p in itertools.takewhile(common_part, parts))
+    return os.path.join(*common_parts) or '.'
```
JukkaL: This doesn't seem to work if there is a mix of `/` and `\` path separators, which can happen on Windows.

Also, I find the new implementation to be harder to understand than the original one. To make reviewing easier, it would be better to incrementally update the original implementation. For example, apply `os.path.normpath` on all the paths before operating on them and use `os.sep` instead of `/`, but otherwise preserver the original logic.

mypy/test/teststubgen.py
```diff
@@ -457,6 +460,21 @@ def test_common_dir_prefix(self) -> None:
...
+        assert common_dir_prefix([r'.\x.pyi']) == '.'
+        assert common_dir_prefix([r'foo\bar\x.pyi']) == r'foo\bar'
+        assert common_dir_prefix([r'foo\bar\x.pyi',
+                                  r'foo\bar\y.pyi']) == r'foo\bar'
+        assert common_dir_prefix([r'foo\bar\x.pyi', r'foo\y.pyi']) == 'foo'
+        assert common_dir_prefix([r'foo\x.pyi', r'foo\bar\y.pyi']) == 'foo'
+        assert common_dir_prefix([r'foo\bar\zar\x.pyi', r'foo\y.pyi']) == 'foo'
+        assert common_dir_prefix([r'foo\x.pyi', r'foo\bar\zar\y.pyi']) == 'foo'
+        assert common_dir_prefix([r'foo\bar\zar\x.pyi', r'foo\bar\y.pyi']) == r'foo\bar'
+        assert common_dir_prefix([r'foo\bar\x.pyi', r'foo\bar\zar\y.pyi']) == r'foo\bar'
```
JukkaL: Test mix of `/` and `\` path separators.

mypy/test/teststubgen.py
```diff
@@ -443,7 +444,9 @@ def h():
 
         assert_equal(remove_misplaced_type_comments(original), dest)
 
-    def test_common_dir_prefix(self) -> None:
+    @pytest.mark.skipif(sys.platform == 'win32',
```
JukkaL: We use `@unittest.skipIf` elsewhere and it would be better to use it here for consistency.

hoefling:
Done, sorry for the delay - caught the flu.

hoefling:
@JukkaL may I kindly ask about reviewing the latest changes?

hauntsaninja (approved):
Thanks! Hope you've recovered from your flu :-)

PR #8078, opened 2019-12-04
Refactor: remove unused attribute from PartialType

JukkaL:
The `inner_types` attribute seems to have no effect.


JukkaL:
> Btw, do we have an issue about allowing the above behavior (specifically for an empty initializer)?

I couldn't find one, but I also don't think that this is important to support this (and perhaps not even desirable).

PR #8091, opened 2019-12-05
Fix type inference regression with partial types

### LINKED ISSUE #8090: Regression: type inference can fall back to List[<nothing>]
Previously if mypy couldn't infer a type for a variable, it would fall back to a type such as `List[Any]` instead of `List[<nothing>]`. However, #8036 accidentally changed this, as can be seen from this example:

```py
x = []  # type: ignore
print(x)
x = []
reveal_type(x)    # List[<nothing>] (used to be List[Any])
```

We should revert to the old behavior.

JukkaL:
Don't infer types such as `List[<nothing>]`; infer `List[Any]` instead.

Fixes #8090.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2174,7 +2174,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             rvalue_type = get_proper_type(rvalue_type)
             if isinstance(rvalue_type, Instance):
                 if rvalue_type.type == typ.type:
-                    var.type = rvalue_type
+                    self.infer_variable_type(var, lvalue, rvalue_type, rvalue)
                     del partial_types[var]
             elif isinstance(rvalue_type, AnyType):
                 var.type = fill_typevars_with_any(typ.type)
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2174,7 +2174,10 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             rvalue_type = get_proper_type(rvalue_type)
             if isinstance(rvalue_type, Instance):
                 if rvalue_type.type == typ.type:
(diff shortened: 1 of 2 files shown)


JukkaL:
Not ready to merge -- there seems to a problem with this PR.

mypy/checker.py
```diff
@@ -2174,7 +2174,10 @@ def try_infer_partial_generic_type_from_assignment(self,
             rvalue_type = get_proper_type(rvalue_type)
             if isinstance(rvalue_type, Instance):
                 if rvalue_type.type == typ.type:
-                    var.type = rvalue_type
+                    if is_valid_inferred_type(rvalue_type):
+                        var.type = rvalue_type
+                    else:
+                        var.type = self.inference_error_fallback_type(rvalue_type)
```
ilevkivskyi: What if one just assigns the variable twice (without reading it), would it silence the error on first assignment? Not that it is too bad, but still may be surprising.

mypy/checker.py
```diff
@@ -2785,6 +2788,7 @@ def infer_variable_type(self, name: Var, lvalue: Lvalue,
         init_type = get_proper_type(init_type)
         if isinstance(init_type, DeletedType):
             self.msg.deleted_as_rvalue(init_type, context)
+            name.type = AnyType(TypeOfAny.from_error)
```
ilevkivskyi: Is there a test for this?

mypy/checker.py
JukkaL: Yeah, a good catch! This behavior is inconsistent. I'll try to fix it.

mypy/checker.py
JukkaL: No, and this is kind of an unrelated thing that I encountered when trying a different fix. I might take this out of this PR.

Issue #8093, opened 2019-12-05
Type inference regression involving filter(None, ...)


JukkaL:
This example started generating errors after #8077: 

```py
from typing import Optional, TypeVar, Iterable, Iterator, List

_T = TypeVar('_T')

def filter(__function: None, __iterable: Iterable[Optional[_T]]) -> List[_T]: ...

x: Optional[str]

# Need type annotation for 'pagelets'
# List item 0 has incompatible type "Optional[str]"; expected "None"
pagelets = filter(None, [x])
```

The issue also happens with Python 2 `filter()`, at least.

cc @ilevkivskyi

PR #8095, opened 2019-12-05
Don't simplify out ErasedType from unions during type inference

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/8093

The fix is kind of simple, we really need to keep all `ErasedType`s intact to preserve the existing type inference logic, but it is also long because I need to thread the flag down like dozen function calls.

The issue itself btw highlight how fragile is current type inference logic (note that `expand_type()` implicitly calls `make_simplified_union()` too, but I think that one doesn't need to be updated because it shouldn't be called with any erased components).

### DIFF
--- a/test-data/unit/check-inference.test
+++ b/test-data/unit/check-inference.test
@@ -2950,5 +2950,6 @@ def filter(__function: None, __iterable: Iterable[Optional[_T]]) -> List[_T]: ..
 
 x: Optional[str]
 
-pagelets = filter(None, [x])
+y = filter(None, [x])
+reveal_type(y)  # N: Revealed type is 'builtins.list[builtins.str*]'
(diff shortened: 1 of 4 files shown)


test-data/unit/check-inference.test
```diff
@@ -2939,3 +2939,16 @@ X = Union[Cov[T], Inv[T]]
...
+# flags: --strict-optional
+from typing import Optional, TypeVar, Iterable, Iterator, List
+
+_T = TypeVar('_T')
+
+def filter(__function: None, __iterable: Iterable[Optional[_T]]) -> List[_T]: ...
+
+x: Optional[str]
+
+pagelets = filter(None, [x])
```
JukkaL: Test that the type of `pagelets` is `List[str]`?

PR #8096, opened 2019-12-06
Inferencing type reference of abstract class inferred from concrete type values

### LINKED ISSUE #8050: Inferencing type reference of abstract class inferred from concrete type values
```python3
import abc

class MyAbstractType(metaclass=abc.ABCMeta):
  @abc.abstractmethod
  def do(self): pass

class MyConcreteA(MyAbstractType):
  def do(self):
    print('A')

class MyConcreteB(MyAbstractType):
  def do(self):
    print('B')

my_types = {
  'A': MyConcreteA,
  'B': MyConcreteB,
}
reveal_type(my_types)  # Revealed type is 'builtins.dict[builtins.str*, def () -> test-abc.MyAbstractType]'

a = my_types['A']()  # Cannot instantiate abstract class 'MyAbstractType' with abstract attribute 'do'
a.do()
b = my_types['B']()  # Cannot instantiate abstract class 'MyAbstractType' with abstract attribute 'do'
b.do()
```
The above example does not pass the type check as-is.
I need to change it to pass the type check as follows:

```python3
import abc
from typing import Mapping, Type

class MyAbstractType(metaclass=abc.ABCMeta):
  @abc.abstractmethod
  def do(self): pass

class MyConcreteA(MyAbstractType):
  def do(self):
    print('A')

class MyConcreteB(MyAbstractType):
  def do(self):
    print('B')

my_types: Mapping[str, Type[MyAbstractType]] = {  # explicit annotation
  'A': MyConcreteA,
  'B': MyConcreteB,
}
reveal_type(my_types)  # Revealed type is 'typing.Mapping[builtins.str, Type[test-abc.MyAbstractType]]'

a = my_types['A']()
a.do()
b = my_types['B']()
b.do()
```

Since mypy is able to infer `MyAbstractType` from the dictionary in the first snippet, I believe that it should be also able to infer `Type[MyAbstractType]` since the dictionary values are not instances but class names.

My setup: Python 3.8.0 + mypy 0.750

Related issues/PRs:
 - #5135 looks like a similar issue as the author did not use explicit annotation for the dictionaries.
 - #2853 allowed using explicit annotation above.

TH3CHARLie:
resolves #8050, following the purposed solution 2 in  https://github.com/python/mypy/issues/8050#issuecomment-560322273 and using the implementation purposed in https://github.com/python/mypy/issues/8050#issuecomment-560326219

Currently, no testcase with regard to meet of callables is available, since I'm not sure what a proper testcase would be.

### DIFF
--- a/mypy/join.py
+++ b/mypy/join.py
@@ -187,8 +187,8 @@ class TypeJoinVisitor(TypeVisitor[ProperType]):
             if is_equivalent(t, self.s):
                 return combine_similar_callables(t, self.s)
             result = join_similar_callables(t, self.s)
-            # unless some or all items represent abstract objects,
-            # we set the flag: from_type_type to True
+            # We set the from_type_type flag to suppress error when a collection of 
+            # concrete class objects gets inferred as their common abstract superclass.
             if not ((t.is_type_obj() and t.type_object().is_abstract) or
                     (self.s.is_type_obj() and self.s.type_object().is_abstract)):
                 result.from_type_type = True
--- a/mypy/join.py
+++ b/mypy/join.py
@@ -187,6 +187,11 @@ class TypeJoinVisitor(TypeVisitor[ProperType]):
             if is_equivalent(t, self.s):
                 return combine_similar_callables(t, self.s)
             result = join_similar_callables(t, self.s)
+            # unless some or all items represent abstract objects,
+            # we set the flag: from_type_type to True
+            if not ((t.is_type_obj() and t.type_object().is_abstract) or
+                (self.s.is_type_obj() and self.s.type_object().is_abstract)):
+                result.from_type_type = True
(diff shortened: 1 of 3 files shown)


mypy/join.py
```diff
@@ -187,6 +187,11 @@ def visit_callable_type(self, t: CallableType) -> ProperType:
             if is_equivalent(t, self.s):
                 return combine_similar_callables(t, self.s)
             result = join_similar_callables(t, self.s)
+            # unless some or all items represent abstract objects,
+            # we set the flag: from_type_type to True
```
ilevkivskyi: This and below comment are not very helpful, because they are just repeating what happens in code. Maybe instead explain the _motivation_. Something like: "We set the from_type_type flag to suppress error when a collection of concrete class objects gets inferred as their common abstract superclass."

The same applies below. Also please start comments with an uppercase letter.

PR #8099, opened 2019-12-06
Couple obvious optimizations for type aliases

ilevkivskyi:
This PR:
* Removes three unused slots
* Makes recursivity test once per alias
* Documents an existing optimization (accidentally made before)

The effect of this it at the limit of precision on my laptop, but it looks like this gives around 1% speed-up (for cold runs).

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -2741,7 +2741,7 @@ class TypeAlias(SymbolNode):
     line and column: Line an column on the original alias definition.
     """
     __slots__ = ('target', '_fullname', 'alias_tvars', 'no_args', 'normalized',
-                 'line', 'column', 'assuming', 'assuming_proper', 'inferring')
+                 'line', 'column', '_is_recursive')
 
     def __init__(self, target: 'mypy.types.Type', fullname: str, line: int, column: int,
                  *,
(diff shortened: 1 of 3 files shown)


mypy/nodes.py
```diff
@@ -2755,6 +2755,7 @@ def __init__(self, target: 'mypy.types.Type', fullname: str, line: int, column:
         self.alias_tvars = alias_tvars
         self.no_args = no_args
         self.normalized = normalized
+        self._is_recursive = None  # type: Optional[bool]
```
JukkaL: Add comment (mention that if non-`None`, this the cached value). Maybe also mention that this is manipulated by `TypeAliasType`, as otherwise a reader may be confused as not much is done here with this.

mypy/types.py
```diff
@@ -212,10 +211,11 @@ def expand_all_if_possible(self) -> Optional['ProperType']:
...
     def is_recursive(self) -> bool:
-        if self._is_recursive is not None:
-            return self._is_recursive
-        is_recursive = self.expand_all_if_possible() is None
-        self._is_recursive = is_recursive
+        assert self.alias is not None, 'Unfixed type alias'
+        is_recursive = self.alias._is_recursive
+        if is_recursive is None:
+            is_recursive = self.expand_all_if_possible() is None
+            self.alias._is_recursive = is_recursive
```
JukkaL: Add comment (we cache the value as an optimization).

PR #8102, opened 2019-12-07
Support silencing only some error codes

ilevkivskyi:
This PR adds support for silencing particular _error codes_ either on command line, in config file or in inline mypy options. The implementation is mostly straightforward, here are however two important comments:
* I initially wanted to also add an option to _opt-in_ selected error codes, but this turns out to be much harder, because there are many call sites scattered in the code that check for whether a given _flag_ is enabled. We can later migrate some of them to error codes. Also in my experience people often ask about opting out for a specific error code rather than opting in.
* If there are two sections in config file that match a given module, the ignore lists are _not_ merged, the more narrow match acts as an override. This is similar to how other options work (like `--always-true` and `--always-false`), I am not sure this is the most obvious behavior, but it looks like it is most flexible. We can reconsider this later if people will find it confusing.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -593,6 +593,41 @@ in error messages.
 
     Show absolute paths to files.
 
+.. option:: --ignore-error-codes CODES
+
+   This flag makes mypy ignore all errors with given error codes. Flag accepts
+   error codes as a comma separated list (there should be no spaces after commas).
+   For example, by default mypy would produce the following errors:
+
+   .. code-block:: python
+
+       class Dynamic:
+           def __init__(self, attr: str, value: object) -> None:
+               setattr(self, attr, value)
+
+       magic_builtin  # error: Name "magic_builtin" is not defined
+       Dynamic("test", 0).test  # error: "Dynamic" has no attribute "test"
+       x: int = "no"  # error: Incompatible types in assignment
+                      # (expression has type "str", variable has type "int")
+
+   But when used as ``mypy --ignore-error-codes=attr-defined,name-defined test.py``
+   it will produce the following errors:
+
+   .. code-block:: python
+
+       class Dynamic:
+           def __init__(self, attr: str, value: object) -> None:
+               setattr(self, attr, value)
+
+       magic_builtin  # No error
+       Dynamic("test", 0).test  # No error
+       x: int = "no"  # error: Incompatible types in assignment
+                      # (expression has type "str", variable has type "int")
+
+   To make mypy show error codes in error messages use :option:`--show-error-codes`.
+   See also the lists of :ref:`default error codes <error-code-list>` and
+   :ref:`optional error codes <error-codes-optional>`.
+
 
 .. _incremental:
 
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -595,38 +595,38 @@ in error messages.
 
 .. option:: --ignore-error-codes CODES
 
-   This flag makes mypy ignore all errors with given error codes. Flag accepts
-   error codes as a comma separated list (there should be no spaces after commas).
-   For example, by default mypy would produce the following errors:
+    This flag makes mypy ignore all errors with given error codes. Flag accepts
+    error codes as a comma separated list (there should be no spaces after commas).
+    For example, by default mypy would produce the following errors:
 
-   .. code-block:: python
+    .. code-block:: python
 
-       class Dynamic:
-           def __init__(self, attr: str, value: object) -> None:
-               setattr(self, attr, value)
+        class Dynamic:
+            def __init__(self, attr: str, value: object) -> None:
+                setattr(self, attr, value)
 
-       magic_builtin  # error: Name "magic_builtin" is not defined
-       Dynamic("test", 0).test  # error: "Dynamic" has no attribute "test"
-       x: int = "no"  # error: Incompatible types in assignment
-                      # (expression has type "str", variable has type "int")
+        magic_builtin  # error: Name "magic_builtin" is not defined
+        Dynamic("test", 0).test  # error: "Dynamic" has no attribute "test"
+        x: int = "no"  # error: Incompatible types in assignment
+                       # (expression has type "str", variable has type "int")
 
-   But when used as ``mypy --ignore-error-codes=attr-defined,name-defined test.py``
-   it will produce the following errors:
+    But when used as ``mypy --ignore-error-codes=attr-defined,name-defined test.py``
+    it will produce the following errors:
 
-   .. code-block:: python
+    .. code-block:: python
 
-       class Dynamic:
-           def __init__(self, attr: str, value: object) -> None:
-               setattr(self, attr, value)
+        class Dynamic:
+            def __init__(self, attr: str, value: object) -> None:
+                setattr(self, attr, value)
 
-       magic_builtin  # No error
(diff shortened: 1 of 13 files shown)


docs/source/command_line.rst
```diff
@@ -593,6 +593,41 @@ in error messages.
 
     Show absolute paths to files.
 
+.. option:: --ignore-error-codes CODES
+
+    This flag makes mypy ignore all errors with given error codes. Flag accepts
+    error codes as a comma separated list (there should be no spaces after commas).
```
JukkaL: It might be better for usability if spaces around commas were ignored. Error codes can't have spaces.

Grammar nit: "comma separated" -> "comma-separated" (we use the latter spelling elsewhere)

docs/source/command_line.rst
```diff
@@ -593,6 +593,41 @@ in error messages.
...
+    For example, by default mypy would produce the following errors:
+
+    .. code-block:: python
+
+        class Dynamic:
+            def __init__(self, attr: str, value: object) -> None:
+                setattr(self, attr, value)
+
+        magic_builtin  # error: Name "magic_builtin" is not defined
```
JukkaL: The comments make this example look kind of busy, since some of the messages are long. Also, it would be nice to include the error codes in the message, as now it's unclear where the error codes come from (`--show-error-codes` is introduced only later).

Here's a suggested change:
* Remove the comments from the example.
* Include full output from a mypy run, without `--ignore-error-codes` flag, but with `--show-error-codes`.
* Show output from another mypy run, now with the the various error codes ignored.

docs/source/command_line.rst
```diff
@@ -593,6 +593,41 @@ in error messages.
...
+                setattr(self, attr, value)
+
+        magic_builtin  # No error
+        Dynamic("test", 0).test  # No error
+        x: int = "no"  # error: Incompatible types in assignment
+                       # (expression has type "str", variable has type "int")
+
+    To make mypy show error codes in error messages use :option:`--show-error-codes`.
+    See also the lists of :ref:`default error codes <error-code-list>` and
+    :ref:`optional error codes <error-codes-optional>`.
```
JukkaL: I think that we should have a warning here, as ignoring some errors could be quite confusing and dangerous, especially `misc`.

docs/source/config_file.rst
```diff
@@ -439,6 +439,11 @@ These options may only be set in the global section (``[mypy]``).
 ``show_absolute_path`` (bool, default False)
     Show absolute paths to files.
 
+``ignore_error_codes`` (comma-separated list of strings)
+    Ignore errors with these error codes in given files or directories.
```
JukkaL: The part "in given files or directories" seems redundant. It's not included in other descriptions, and so it raises the question why this option applies to given files or directories, while other options don't apply to given files or directories?

docs/source/config_file.rst
```diff
@@ -439,6 +439,11 @@ These options may only be set in the global section (``[mypy]``).
...
+    See :option:`--ignore-error-codes <mypy --ignore-error-codes>` for
+    more details.
```
JukkaL: Really tiny consistency nit: we use "for more information" about and "more details here". Maybe use "for more information" here as well.

docs/source/command_line.rst
```diff
@@ -593,6 +593,41 @@ in error messages.
 
     Show absolute paths to files.
 
+.. option:: --ignore-error-codes CODES
+
+    This flag makes mypy ignore all errors with given error codes. Flag accepts
```
JukkaL: I think that we should include a link to documentatino about error codes here instead of later. The text "error codes" could link to the relevant topic.

Grammar nit: "Flag" -> "This flag"

docs/source/error_codes.rst
```diff
@@ -37,8 +37,8 @@ Silencing errors based on error codes
 You can use a special comment ``# type: ignore[code, ...]`` to only
 ignore errors with a specific error code (or codes) on a particular
 line.  This can be used even if you have not configured mypy to show
-error codes. Currently it's only possible to disable arbitrary error
-codes on individual lines using this comment.
+error codes. Also you can use :option:`--ignore-error-codes <mypy --ignore-error-codes>`
```
JukkaL: Style nit: "Also" seems redundant.

docs/source/error_codes.rst
```diff
@@ -37,8 +37,8 @@ Silencing errors based on error codes
...
+for more coarse-grained (per file or per directory) silencing of errors.
```
JukkaL: The actual command line flag only allows global silencing of errors, but this implies that it's per file or per directory. Either leave the parenthetical remark out, or somehow make it clear that per-file/directory options are only available through the config file or `# mypy: ...` comments. For consistency, I'd leave it out, since we probably don't want to discuss this separately for every reference to every option.

mypy/main.py
```diff
@@ -640,6 +641,9 @@ def add_invertible_flag(flag: str,
     add_invertible_flag('--show-absolute-path', default=False,
                         help="Show absolute paths to files",
                         group=error_group)
+    error_group.add_argument(
+        '--ignore-error-codes', metavar='CODES', type=split_commas,
+        help="Ignore errors with these error codes (comma separated)")
```
JukkaL: Spelling nit: use "comma-separated" (similar to above)

mypy/options.py
```diff
@@ -132,6 +133,9 @@ def __init__(self) -> None:
         # Files in which to ignore all non-fatal errors
         self.ignore_errors = False
 
+        # Ignore errors with these error codes in a given file.
```
JukkaL: Nit: "in a given files." seems redunant. Also omit period at the end for consistency?

test-data/unit/check-flags.test
```diff
@@ -1232,3 +1232,37 @@ A = List  # OK
...
+not_defined
+x: int = 'no'
+'no'.no_way
+def test(x: str) -> None: ...
+test(0)
+[file mypy.ini]
+\[mypy]
+ignore_error_codes = attr-defined, arg-type
+\[mypy-b]
+ignore_error_codes = name-defined, assignment
```
JukkaL: I'd like to eventually move various strictness flags to be enabled/disabled through error codes. To do this, it seems that per-file options need to be a delta over global settings, instead of replacing them. Also, command-line options probably should be a delta over settings in the ini file. I.e., the ignored error codes would be the union of error codes ignored in the config file, on the command line, and in per-file options. 

The logic gets a bit more involved once we allow enabling error codes as well, but at least we should be able to design the semantic for ignoring error codes in a way that we don't need to introduce a compatibility break.

test-data/unit/check-inline-config.test
```diff
@@ -157,3 +157,12 @@ main:4: error: Unterminated quote in configuration comment
...
 [out]
 main:1: error: Unrecognized option: skip_file = True
+
+[case testIgnoreErrorCodesInline]
+# mypy: ignore-error-codes="attr-defined,arg-type"
+
+not_defined  # E: Name 'not_defined' is not defined
+x: int = 'no'  # E: Incompatible types in assignment (expression has type "str", variable has type "int")
+'no'.no_way
+def test(x: str) -> None: ...
```
JukkaL: Also add daemon test case (to make sure the daemon applies these after updates).

ilevkivskyi:

PR #8107, opened 2019-12-08
Produce error when assigning Enum or TypedDict as attribute

### LINKED ISSUE #8009: "Enum" has no attribute "__iter__" on instance attributes defined with functional API
**Steps to reproduce**
Run mypy against the following code:

```
from enum import Enum


class OuterEnum(Enum):
    foo = "bar"


class Foo:
    def __init__(self,) -> None:
        self.OuterEnum = OuterEnum
        list(self.OuterEnum)
        self.InnerEnum = Enum("InnerEnum", [("foo", "bar")])
        list(self.InnerEnum)  # Error!
        [i for i in self.InnerEnum]  # Error!
        self.DictEnum = Enum("DictEnum", {"foo": "bar"})
        list(self.DictEnum)  # Error!


OtherEnum = Enum("OtherEnum", [("foo", "bar")])
list(OtherEnum)
```

**Observed behavior**
When trying to iterate over an enumeration defined as an instance attribute using the [functional API](https://docs.python.org/3.7/library/enum.html#functional-api), mypy raises an error when attempting to iterate over said attribute.

The above snippet raises three mypy errors where marked:

```
$ poetry run mypy foo.py 
foo.py:13: error: No overload variant of "list" matches argument type "Enum"
foo.py:13: note: Possible overload variant:
foo.py:13: note:     def [_T] __init__(self, iterable: Iterable[_T]) -> List[_T]
foo.py:13: note:     <1 more non-matching overload not shown>
foo.py:14: error: "Enum" has no attribute "__iter__"; maybe "__str__" or "__dir__"? (not iterable)
foo.py:16: error: No overload variant of "list" matches argument type "Enum"
foo.py:16: note: Possible overload variant:
foo.py:16: note:     def [_T] __init__(self, iterable: Iterable[_T]) -> List[_T]
foo.py:16: note:     <1 more non-matching overload not shown>
Found 3 errors in 1 file (checked 1 source file)
```

**Expected behavior**
When testing the above code, iteration over the instance attribute Enum works fine, and mypy does not raise errors when iterating over similar Enums if they're defined using the class syntax or aren't instance attributes. I would not expect mypy to raise an error here.

**Environment**

```
$ poetry run mypy --version
mypy 0.740
$ poetry run python --version
Python 3.7.4
```
Also tested against master with the same behavior
```
$ poetry run mypy --version
mypy 0.750+dev.2af0ac021335ec4a3e84a0ba739b4aea7e659c43
```

Thank you for your work on this tool!

TH3CHARLie:
resolves #8009, simply following solution by #7662 

Additionally, when I was finishing the first version of the PR I noticed that some of #7662 could be refined:
https://github.com/python/mypy/blob/e8ae12293cf1a8e32769c2149f4c8c2ca043b790/mypy/semanal.py#L2152-L2154
Here using `lvalue` instead of `s.lvalues[0]` would be more concise.
https://github.com/python/mypy/blob/e8ae12293cf1a8e32769c2149f4c8c2ca043b790/test-data/unit/pythoneval.test#L1482-L1492
The test is in `pythoneval.test`, but it makes more sense in `check-namedtuple.test` and the name could be also changed accordingly.

@ilevkivskyi What's your opinion on this? Should these changes be made? And if so, should they be made in this PR or a sperate one.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -2161,13 +2161,16 @@ class SemanticAnalyzer(NodeVisitor[None],
         """Check if s defines a typed dict."""
         if isinstance(s.rvalue, CallExpr) and isinstance(s.rvalue.analyzed, TypedDictExpr):
             return True  # This is a valid and analyzed typed dict definition, nothing to do here.
-        if len(s.lvalues) != 1 or not isinstance(s.lvalues[0], NameExpr):
+        if len(s.lvalues) != 1 or not isinstance(s.lvalues[0], (NameExpr, MemberExpr)):
             return False
         lvalue = s.lvalues[0]
         name = lvalue.name
         is_typed_dict, info = self.typed_dict_analyzer.check_typeddict(s.rvalue, name,
                                                                        self.is_func_scope())
         if not is_typed_dict:
+            return False
+        if isinstance(lvalue, MemberExpr):
+            self.fail("TypedDict type as attribute is not supported", lvalue)
             return False
         # Yes, it's a valid typed dict, but defer if it is not ready.
         if not info:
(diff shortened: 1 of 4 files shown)


ilevkivskyi:
> Should these changes be made? And if so, should they be made in this PR or a sperate one.

Yes, I think in a separate PR.

msullivan:
I would like https://github.com/python/mypy/pull/8134 to make it in (not so much because I care if it is in the release itself, but I don't want to wait until after our testing period to deploy it internally)

Issue #8005, opened 2019-11-23
Extending a Protocol leads to unsafe behavior


jingw:
* Are you reporting a bug, or opening a feature request?
bug

* Please insert below the code you are checking with mypy, or a mock-up repro if the source is private. We would appreciate if you try to simplify your case to a minimal repro.
```
from typing import Protocol

class P(Protocol):
    def foo(self) -> int: ...

class A(P):  # mypy does not flag this
    pass

class B:
    pass

test: P = A()  # mypy does not flag this
test = B()  # mypy flags this

print(A().foo() + 5)  # both error at runtime
print(B().foo() + 5)
```

* What is the actual behavior/output?
`A().foo() + 5` crashes at runtime with `TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'`
`B` is correctly type checked and rejected.

* What is the behavior/output you expect?
According to https://mypy.readthedocs.io/en/latest/protocols.html#defining-subprotocols-and-subclassing-protocols, "Explicitly including a protocol as a base class ... forces mypy to verify that your class implementation is actually compatible with the protocol."
I would expect extending a Protocol to be similar to implementing an interface in Java.
In this particular example, mypy should ideally detect that `A` does not fill in the stub method from `P`, like it does for `B`. Alternatively, if that is not possible, perhaps extending a protocol should not be allowed.

* What are the versions of mypy and Python you are using? Do you see the same issue after installing mypy from Git master?
Python 3.8.0
mypy 0.750+dev.d9dea5f3ad30bb449e4167b3e8476684ded0482e

* What are the mypy flags you are using? (For example --strict-optional)
`mypy --strict test.py`

ilevkivskyi:
The real source of unsafety here is https://github.com/python/mypy/issues/2350. A quick fix in the context of protocols (assuming we are not fixing the root cause soon) is to treat methods with empty bodies as implicitly abstract.

thomkeh:
FWIW, `pyright` now handles this correctly as of version [1.1.213](https://github.com/microsoft/pyright/releases/tag/1.1.213). (The relevant commit: https://github.com/microsoft/pyright/commit/44c98fe74734e692683b9b71ae743c82f2072154)

nickgaya:
If you add an `@abstractmethod` annotation to `P.foo` then MyPy will behave as expected.  I recommend closing this bug as invalid.

PR #8111, opened 2019-12-09
Handle empty bodies safely

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/2350
Fixes https://github.com/python/mypy/issues/8005

The implementation is mostly simple, except overloads are a bit tricky and there is still little unsafety can slip through (we can fix it later with another function flag if it will look like worth it, see TODO item in a test). The actual diff is quite small but there are tons of tests, because there are many possible combinations and corner cases, and I wanted to test as much of them as possible.

I am following the logic I previously proposed in #2350. I copy it here FTR:
* Properly check empty bodies as a default
* Add an `--allow-empty-bodies` flag, we need it for tests (hundreds would fail otherwise, also in other projects like `sqlalchemy-stubs` and `django-stubs`)
* Allow empty bodies in stub files (obviously)
* Allow empty bodies if method is decorated with `@abstractmethod`
* In _protocols_ make functions with empty bodies implicitly abstract (unless this is a stub)
* Add a flag for empty-bodied abstract methods to prohibit calling them via `super()` (unless they come from a stub)

Note that the stub exceptions in two last points can be summarized as: since we don't know the actual body for something that comes from a stub file, we choose to not give a false positive and stay silent.

Another note is that after https://github.com/python/mypy/pull/8102 is merged we can turn the hidden flag into an error code, but TBH I am not sure it is necessary. It looks like the flag is more handy, and it makes sense (and feels natural) to use the existing error code `[return]` (missing return statement) for the unsafe empty bodies.

I am going to try this soon against internal codebases to see if there are some problematic corner cases missing and make any adjustment + tests if needed. Note that `mypy/visitor.py` has an existing unsafety that this new logic uncovers (there is an existing TODO about this). This requires turning `warn_no_return` flag off for this file.

Finally, there were couple tiny bugs discovered by tests I added, so I just fixed them in this same PR.


ilevkivskyi:

TH3CHARLie:
I am going to try to work on this if no one else is currently doing the same.

Issue #8088, opened 2019-12-05
dmypy start doesn't work when files is defined in mypy.ini


rggjan:
I have a mypy.ini with some default files to be checked like:

```
[mypy]
files = path/to/**/*.py
```

Running `mypy` with this config works fine. Running `dmypy run` also works fine.

However, running
`dmypy start`
gives the following error:
`dmypy: start/restart does not accept sources`

Wouldn't it make sense to just ignore the files configuration in this case? Or do I need a second mypy.ini with exactly the same content and just the `files` line deleted to be able to start mypy with it?

JukkaL:
Yeah, this this clearly doesn't make much sense. Ignoring the files in configuration seems like the right thing to do.

PR #8133, opened 2019-12-11
dmypy start doesn't work when files is defined in mypy.ini (#8088)

unoyx:
I'm trying to fix issue #8088 , current change is just ignore the file fields in config when start from dmypy start.

I'm still not familier with dmypy, please review the PR and any advice is welcome.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -809,6 +809,9 @@ def process_options(args: List[str],
     # Parse config file first, so command line can override.
     options = Options()
     parse_config_file(options, config_file, stdout, stderr)
+    # the 'dmypy start' will just ignore files in config
+    if server_options:
+        options.files = None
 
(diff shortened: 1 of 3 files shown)


mypy/main.py
```diff
@@ -809,6 +809,9 @@ def add_invertible_flag(flag: str,
     # Parse config file first, so command line can override.
     options = Options()
     parse_config_file(options, config_file, stdout, stderr)
+    # the 'dmypy start' will just ignore files in config
```
msullivan: Have you tested this? I think this might still fail.

In any case, I think the best place to do make the change here is in `dmypy_server.py` where the error message is generated.

unoyx: ok, thank you. I'll enhance the fix first.

PR #8135, opened 2019-12-12
stubgen: fix return type of annotated __init__

### LINKED ISSUE #8110: stubgen generates __init__ with return type Any for partially annotated input
From this code, which typechecks
 
```python
class X:
  def __init__(self, x: str):
     pass
```
`stubgen` generates:

```python
from typing import Any

class X:
  def __init__(self, x: str) -> Any: ...
```
for which `mypy` reports:
```
out/repro.pyi:4: error: The return type of "__init__" must be None
```

`stubgen` already has knowledge of `__init__` return value special case, when converting `__init__` without type annotations, the return value is annotated as `None`, but when `__init__` is partially annotated and lacks a type annotation for returned value, it is annotated as `Any`.

I don't know how much this practice of omitting type annotation for return value of `__init__` is common, but I noticed that there also are cases in some examples from pep048 (
https://www.python.org/dev/peps/pep-0484/#forward-references ).

If that makes sense, it seems easy to annotate `__init__`  with `None` also in the case where there are partial type annotations. It would also prevent the weird behaviour of generating stubs which do not typecheck from code which typechecks.

The selection of return type is implemented here https://github.com/python/mypy/blob/bbb192d9aab3803cb172ce4674c026e20606fcb0/mypy/stubgen.py#L604-L613

we could move the special casing of `__init__` around to also use `None` in that case. Please let me know if this looks correct and if you want me to submit a pull request for this.

perrinjerome:
When __init__ had partial type annotations, for parameters but not for
the return type, the return type was annotated as Any, which is reported
as an error by mypy when checking the generated stubs.

Make sure to annotate the return value as None also in the case of
existing type annotations for

fixes #8110

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -602,14 +602,14 @@ class StubGenerator(mypy.traverser.TraverserVisitor):
                 arg = name + annotation
             args.append(arg)
         retname = None
-        if isinstance(o.unanalyzed_type, CallableType):
+        if o.name != '__init__' and isinstance(o.unanalyzed_type, CallableType):
             retname = self.print_annotation(o.unanalyzed_type.ret_type)
         elif isinstance(o, FuncDef) and (o.is_abstract or o.name in METHODS_WITH_RETURN_VALUE):
             # Always assume abstract methods return Any unless explicitly annotated. Also
             # some dunder methods should not have a None return type.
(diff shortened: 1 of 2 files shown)


perrinjerome:
@msullivan That's for #8110 , thank you !

msullivan:
Great!

PR #8140, opened 2019-12-13
Exclude "__order__" attribute from Enum Union expansion

lazytype:
The `__order__` attribute is used to determine the definition order of enum values and does not define an enum value itself.

### DIFF
--- a/test-data/unit/check-enum.test
+++ b/test-data/unit/check-enum.test
@@ -659,6 +659,40 @@ else:
     reveal_type(y)  # No output here: this branch is unreachable
 [builtins fixtures/bool.pyi]
 
+[case testEnumReachabilityChecksWithOrdering]
+from enum import Enum
+from typing_extensions import Literal
+
+class Foo(Enum):
+    _order_ = "A B"
+    A = 1
+    B = 2
+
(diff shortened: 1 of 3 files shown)


lazytype:
Actually, don't review this yet. There's some more nuanced behavior for Enums that I need to look into, e.g. https://github.com/python/cpython/blob/0b41a922f95f62b620d5a197b9f2ed8e4bb70730/Lib/enum.py#L90

lazytype:
Ok, ready for review.

test-data/unit/check-enum.test
```diff
@@ -659,6 +659,40 @@ else:
...
+[case testEnumReachabilityChecksWithOrdering]
+from enum import Enum
+from typing_extensions import Literal
+
+class Foo(Enum):
+    _order_ = "A B"
+    A = 1
+    B = 2
+
+Foo._order_  # E: "Type[Foo]" has no attribute "_order_"
```
lazytype: I feel like we could give a better explanation, but the wording is a bit too tricky for me.

msullivan: Yeah this is fine

msullivan:
Thanks!

PR #8146, opened 2019-12-14
Flatten TypeAliasType when it is aliased as a Union

### LINKED ISSUE #8125: Binary arithmetic on nested Unions of numeric types fails to typecheck
### Are you reporting a bug, or opening a feature request?

bug (I think)

### Please insert below the code you are checking with mypy, or a mock-up repro if the source is private. We would appreciate if you try to simplify your case to a minimal repro.

``` python
from typing import Union

T1 = Union[int, float]
T2 = Union[T1, complex]

def foo(a: T2, b: T2) -> T2:
    return a + b
```

### What is the actual behavior/output?

```
$ mypy ~/tmp/foo.py
tmp/foo.py:7: error: Unsupported operand types for + ("int" and "Union[int, float]")
tmp/foo.py:7: note: Both left and right operands are unions
```

### What is the behavior/output you expect?

To typecheck without errors.

### What are the versions of mypy and Python you are using? Do you see the same issue after installing mypy from Git master?

python 3.6.9
mypy 0.750 and master produce the above error.
mypy 0.740 does not produce an error.

### What are the mypy flags you are using? (For example --strict-optional)

The issue reproduces when running mypy both without any flags and when running with `--strict`.

### Additional info

In case it's helpful, here are some possibly/loosely-related issues I found while searching the issue tracker:

https://github.com/python/mypy/issues/2128
https://github.com/python/mypy/issues/3196

TH3CHARLie:
Resolves #8125 

@appleby has some really good thoughts on the issue and I tried some of them. 

The main problem is not about flattening unions inside variants since the following code generates no error
```python
from typing import Union

T1 = Union[int, float]
T2 = Union[Union[Union[int, float], float], Union[float, complex], complex]

def foo(a: T2, b: T2) -> T2:
    return a + b
```

The problem, however, is because when using `TypeAliasType` to alias a Union, the `TypeAliasType` will not get flattened, so this PR fixes this.

I haven't added testcase yet, since no proper fixture file seems to support the add ops in the test code of #8125 , so lots of `error: Unsupported left operand type for + ` will be reported (running mypy using command line will pass the test code)
- [x] add testcase

cc @msullivan

### DIFF
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -2239,12 +2239,11 @@ def flatten_nested_unions(types: Iterable[Type],
     # TODO: ban such aliases in semantic analyzer.
     flat_items = []  # type: List[Type]
     if handle_type_alias_type:
-        types = [get_proper_type(item) if isinstance(item, TypeAliasType)
-                 else item for item in types]
+        types = get_proper_types(types)
     for tp in types:
         if isinstance(tp, ProperType) and isinstance(tp, UnionType):
             flat_items.extend(flatten_nested_unions(tp.items,
-                                        handle_type_alias_type=handle_type_alias_type))
+                              handle_type_alias_type=handle_type_alias_type))
         else:
             flat_items.append(tp)
     return flat_items
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -2231,15 +2231,20 @@ def has_type_vars(typ: Type) -> bool:
     return typ.accept(HasTypeVars())
 
 
-def flatten_nested_unions(types: Iterable[Type]) -> List[Type]:
+def flatten_nested_unions(types: Iterable[Type],
+                          handle_type_alias_type: bool = False) -> List[Type]:
     """Flatten nested unions in a type list."""
     # This and similar functions on unions can cause infinite recursion
     # if passed a "pathological" alias like A = Union[int, A] or similar.
     # TODO: ban such aliases in semantic analyzer.
     flat_items = []  # type: List[Type]
+    if handle_type_alias_type:
+        types = [get_proper_type(item) if isinstance(item, TypeAliasType)
+                 else item for item in types]
     for tp in types:
         if isinstance(tp, ProperType) and isinstance(tp, UnionType):
-            flat_items.extend(flatten_nested_unions(tp.items))
+            flat_items.extend(flatten_nested_unions(tp.items,
+                                        handle_type_alias_type=handle_type_alias_type))
         else:
             flat_items.append(tp)
     return flat_items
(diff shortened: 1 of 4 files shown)


appleby:
IIUC, this change only handles a single level of nesting?  A single level is all I currently need so I'm not complaining, but, for example, the following still fails to typecheck for me.

```python
from typing import Union

T1 = int
T2 = Union[T1, float]
T3 = Union[T2, complex]
T4 = Union[T3, int]

def foo(a: T4, b: T4) -> T4:
    return a + b
```

TH3CHARLie:
> IIUC, this change only handles a single level of nesting? A single level is all I currently need so I'm not complaining, but, for example, the following still fails to typecheck for me.
> 
> ```python
> from typing import Union
> 
> T1 = int
> T2 = Union[T1, float]
> T3 = Union[T2, complex]
> T4 = Union[T3, int]
> 
> def foo(a: T4, b: T4) -> T4:
>     return a + b
> ```

Yeah you're right, I'll fix it up

appleby:
I can't comment on the suitability of the fix, but I can confirm that the changes here resolve both the toy minimal example in #8125, and also the Real World type checking issue I was having that prompted me to open #8125 in the first place.

mypy/types.py
```diff
@@ -2231,15 +2231,20 @@ def has_type_vars(typ: Type) -> bool:
...
     # TODO: ban such aliases in semantic analyzer.
     flat_items = []  # type: List[Type]
+    if handle_type_alias_type:
+        types = [get_proper_type(item) if isinstance(item, TypeAliasType)
+                 else item for item in types]
     for tp in types:
         if isinstance(tp, ProperType) and isinstance(tp, UnionType):
-            flat_items.extend(flatten_nested_unions(tp.items))
+            flat_items.extend(flatten_nested_unions(tp.items,
+                                        handle_type_alias_type=handle_type_alias_type))
```
ilevkivskyi: Please fix indentation here to align this with `(` on the line above (I am not sure why flake8 is happy with this).

mypy/types.py
```diff
@@ -2231,15 +2231,20 @@ def has_type_vars(typ: Type) -> bool:
...
+def flatten_nested_unions(types: Iterable[Type],
+                          handle_type_alias_type: bool = False) -> List[Type]:
     """Flatten nested unions in a type list."""
     # This and similar functions on unions can cause infinite recursion
     # if passed a "pathological" alias like A = Union[int, A] or similar.
     # TODO: ban such aliases in semantic analyzer.
     flat_items = []  # type: List[Type]
+    if handle_type_alias_type:
+        types = [get_proper_type(item) if isinstance(item, TypeAliasType)
+                 else item for item in types]
```
ilevkivskyi: This can be expressed simply as `types = get_proper_types(types)` (note `s` at the end of the function name).

PR #8148, opened 2019-12-15
Make reachability code understand chained comparisons (v2)

Michael0x2a:
This pull request is v2 (well, more like v10...) of my attempts to make our reachability code better understand chained comparisons.

Unlike https://github.com/python/mypy/pull/7169, this diff focuses exclusively on adding support for chained operation comparisons and deliberately does not attempt to change any of the semantics of how identity and equality operations are performed.

Specifically, mypy currently only examines the first two operands within a comparison expression when refining types. That means the following expressions all do not behave as expected:

```python
x: MyEnum
y: MyEnum
if x is y is MyEnum.A:
    # x and y are not narrowed at all

if x is MyEnum.A is y:
    # Only x is narrowed to Literal[MyEnum.A]
```

This pull request fixes this so we correctly infer the literal type
for x and y in both conditionals.

Some additional notes:

1. While analyzing our codebase, I found that while comparison expressions involving two or more `is` or `==` operators were somewhat common, there were almost no comparisons involving chains of `!=` or `is not` operators, and no comparisons involving "disjoint chains" -- e.g. expressions like `a == b < c == b` where there are multiple "disjoint" chains of equality comparisons.

   So, this diff is primarily designed to handle the case where a comparision expression has just one chain of `is` or `==`. For all other cases, I fall back to the more naive strategy of evaluating each comparision individually and and-ing the inferred types together without attempting to propagate any info.

2. I tested this code against one of our internal codebases. This ended up making mypy produce 3 or 4 new errors, but they all seemed legitimate, as far as I can tell.

3. I plan on submitting a follow-up diff that takes advantage of the work done in this diff to complete support for tagged unions using any Literal key, as previously promised.

   (I tried adding support for tagged unions in this diff, but attempting to simultaneously add support for chained comparisons while overhauling the semantics of `==` proved to be a little too overwhelming for me. So, baby steps.)

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -5368,7 +5368,7 @@ def group_comparison_operands(pairwise_comparisons: Iterable[Tuple[str, Expressi
     }  # type: Dict[str, DisjointDict[Key, int]]
 
     simplified_operator_list = []  # type: List[Tuple[str, List[int]]]
-    last_operator = None
+    last_operator = None  # type: Optional[str]
     current_indices = set()  # type: Set[int]
     current_hashes = set()  # type: Set[Key]
     for i, (operator, left_expr, right_expr) in enumerate(pairwise_comparisons):
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3788,67 +3788,109 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                     vartype = type_map[expr]
                     return self.conditional_callable_type_map(expr, vartype)
         elif isinstance(node, ComparisonExpr):
-            operand_types = [coerce_to_literal(type_map[expr])
-                             for expr in node.operands if expr in type_map]
+            # Step 1: Obtain the types of each operand and whether or not we can
+            # narrow their types. (For example, we shouldn't try narrowing the
+            # types of literal string or enum expressions).
 
-            is_not = node.operators == ['is not']
-            if (is_not or node.operators == ['is']) and len(operand_types) == len(node.operands):
-                if_vars = {}  # type: TypeMap
-                else_vars = {}  # type: TypeMap
+            operands = node.operands
+            operand_types = []
+            narrowable_operand_indices = set()
+            for i, expr in enumerate(operands):
+                if expr not in type_map:
+                    return {}, {}
+                expr_type = type_map[expr]
+                operand_types.append(expr_type)
 
-                for i, expr in enumerate(node.operands):
-                    var_type = operand_types[i]
-                    other_type = operand_types[1 - i]
+                if (literal(expr) == LITERAL_TYPE
+                        and not is_literal_none(expr)
+                        and not is_literal_enum(type_map, expr)):
+                    narrowable_operand_indices.add(i)
 
-                    if literal(expr) == LITERAL_TYPE and is_singleton_type(other_type):
-                        # This should only be true at most once: there should be
-                        # exactly two elements in node.operands and if the 'other type' is
-                        # a singleton type, it by definition does not need to be narrowed:
-                        # it already has the most precise type possible so does not need to
-                        # be narrowed/included in the output map.
-                        #
-                        # TODO: Generalize this to handle the case where 'other_type' is
-                        # a union of singleton types.
+            # Step 2: Group operands chained by either the 'is' or '==' operands
+            # together. For all other operands, we keep them in groups of size 2.
+            # So the expression:
+            #
+            #   x0 == x1 == x2 < x3 < x4 is x5 is x6 is not x7 is not x8
+            #
+            # ...is converted into the simplified operator list:
+            #
+            #  [("==", [0, 1, 2]), ("<", [2, 3]), ("<", [3, 4]),
+            #   ("is", [4, 5, 6]), ("is not", [6, 7]), ("is not", [7, 8])]
+            #
+            # We group identity/equality expressions so we can propagate information
+            # we discover about one operand across the entire chain. We don't bother
+            # handling 'is not' and '!=' chains in a special way: those are very rare
+            # in practice.
 
-                        if isinstance(other_type, LiteralType) and other_type.is_enum_literal():
-                            fallback_name = other_type.fallback.type.fullname
-                            var_type = try_expanding_enum_to_union(var_type, fallback_name)
+            simplified_operator_list = []  # type: List[Tuple[str, List[int]]]
+            last_operator = node.operators[0]
+            current_group = set()  # type: Set[int]
+            for i, (operator, left_expr, right_expr) in enumerate(node.pairwise()):
+                if current_group and (operator != last_operator or operator not in {'is', '=='}):
+                    simplified_operator_list.append((last_operator, sorted(current_group)))
+                    last_operator = operator
+                    current_group = set()
 
-                        target_type = [TypeRange(other_type, is_upper_bound=False)]
-                        if_vars, else_vars = conditional_type_map(expr, var_type, target_type)
-                        break
+                # Note: 'i' corresponds to the left operand index, so 'i + 1' is the
+                # right operand.
+                current_group.add(i)
+                current_group.add(i + 1)
 
-                if is_not:
-                    if_vars, else_vars = else_vars, if_vars
-                return if_vars, else_vars
-            # Check for `x == y` where x is of type Optional[T] and y is of type T
-            # or a type that overlaps with T (or vice versa).
-            elif node.operators == ['==']:
-                first_type = type_map[node.operands[0]]
-                second_type = type_map[node.operands[1]]
-                if is_optional(first_type) != is_optional(second_type):
-                    if is_optional(first_type):
-                        optional_type, comp_type = first_type, second_type
-                        optional_expr = node.operands[0]
+            simplified_operator_list.append((last_operator, sorted(current_group)))
+
+            # Step 3: Analyze each group and infer more precise type maps for each
+            # assignable operand, if possible. We combine these type maps together
+            # in the final step.
+
+            partial_type_maps = []
+            for operator, expr_indices in simplified_operator_list:
+                if operator in {'is', 'is not'}:
+                    if_map, else_map = self.refine_identity_comparison_expression(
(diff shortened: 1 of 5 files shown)


test-data/unit/check-enum.test
```diff
@@ -951,3 +951,119 @@ class A:
...
+if x is Foo.A and x is Foo.B:
+    reveal_type(x)  # E: Statement is unreachable
+else:
+    reveal_type(x)  # N: Revealed type is '__main__.Foo'
+reveal_type(x)      # N: Revealed type is '__main__.Foo'
+
+# TODO: This should behave in the same way as above.
+# However, unlike the above, we currently don't progressively update the type of 'x' as
+# we check each individual comparison. So, when we do 'x is Foo.B', mypy still thinks
+# 'x' is of type 'Foo', which is why we get the below faulty result.
```
ilevkivskyi: Would it be possible to fix this by merging the groups if they contain expressions with the same literal hash and literal level `LITERAL_TYPE`? I mean my guess is that this fails because there are two groups in `simplified_operator_list` that contain `x`.

test-data/unit/check-enum.test
```diff
@@ -951,3 +951,119 @@ class A:
...
+
+final_a: Final = Foo.A
+final_b: Final = Foo.B
+if x is final_a is final_b:
+   reveal_type(x)   # E: Statement is unreachable
+else:
+   reveal_type(x)   # N: Revealed type is '__main__.Foo'
+reveal_type(x)      # N: Revealed type is '__main__.Foo'
+
+[builtins fixtures/primitives.pyi]
```
ilevkivskyi: Maybe one test with a tricky grouping, like two triples connected with `<`? Ideally you could just add a unit test for the grouping algorithm, but it seems to me this is not easy.

mypy/checker.py
```diff
@@ -3788,67 +3788,109 @@ def find_isinstance_check_helper(self, node: Expression) -> Tuple[TypeMap, TypeM
...
-                right_ok = right_type and (not is_optional(right_type) and
-                                           (not isinstance(right_type, Instance) or
-                                            right_type.type.fullname != 'builtins.object'))
-                if (right_type and right_ok and is_optional(left_type) and
-                        literal(expr) == LITERAL_TYPE and not is_literal_none(expr) and
-                        is_overlapping_erased_types(left_type, right_type)):
-                    if node.operators == ['in']:
-                        return {expr: remove_optional(left_type)}, {}
-                    if node.operators == ['not in']:
-                        return {}, {expr: remove_optional(left_type)}
```
ilevkivskyi: I can't find where the this whole chunk moved. Do we now support `a in b in c`? If yes, is there a test?

mypy/checker.py
```diff
@@ -4053,6 +4095,120 @@ def replay_lookup(new_parent_type: ProperType) -> Optional[Type]:
...
+
+        if_map = {}
+        for i in narrowable_operand_indices:
+            expr_type = operand_types[i]
+            if not is_optional(expr_type):
+                continue
+            if any(is_overlapping_erased_types(expr_type, t) for t in non_optional_types):
+                if_map[operands[i]] = remove_optional(expr_type)
+
+        return if_map, {}
```
ilevkivskyi: Should this refine both `x` and `y` if I have:
```python
x: Optional[int]
y: Optional[int]
if x == y == 1:
    ...
```
If yes, please add a test for this.

mypy/checker.py
```diff
@@ -4053,6 +4095,120 @@ def replay_lookup(new_parent_type: ProperType) -> Optional[Type]:
...
+        # We skip refining the type of the singleton below, so ideally we'd
+        # want to pick an expression we were going to skip anyways.
+        singleton_index = -1
+        for i in possible_singleton_indices:
+            if i not in narrowable_operand_indices:
+                singleton_index = i
+
+        # Oh well, give up and just arbitrarily pick the last item.
+        if singleton_index == -1:
+            singleton_index = possible_singleton_indices[-1]
```
ilevkivskyi: Do we have a test for this situation?

mypy/checker.py
```diff
@@ -4496,6 +4652,26 @@ def is_false_literal(n: Expression) -> bool:
             or isinstance(n, IntExpr) and n.value == 0)
 
 
+def is_literal_enum(type_map: Mapping[Expression, Type], n: Expression) -> bool:
```
ilevkivskyi: Docstring is missing here (ideally with an example).

mypy/checker.py
```diff
@@ -4587,6 +4763,75 @@ def or_conditional_maps(m1: TypeMap, m2: TypeMap) -> TypeMap:
     return result
 
 
+def or_partial_conditional_maps(m1: TypeMap, m2: TypeMap) -> TypeMap:
```
ilevkivskyi: This name doesn't really reflect what this does. Maybe use `combine_conditional_maps()` would be better? I however don't have strong feelings here.

mypy/checker.py
```diff
@@ -4587,6 +4763,75 @@ def or_conditional_maps(m1: TypeMap, m2: TypeMap) -> TypeMap:
...
+
+    ...we'd return the output:
+
+        (
+            {x: TypeIfX,   y: TypeIfY,   shared: PseudoIntersection[TypeIfShared1, TypeIfShared2]},
+            {x: TypeElseX, y: TypeElseY, shared: Union[TypeElseShared1, TypeElseShared2]},
+        )
+
+    ...where "PseudoIntersection[X, Y] == Y" because mypy actually doesn't understand intersections
+    yet, so we settle for just arbitrarily picking the right expr's type.
```
ilevkivskyi: Do we have a test whether this would actually make a difference?

test-data/unit/check-enum.test
Michael0x2a: I tried implementing this suggestion, and it seems to work!

It did make the grouping algorithm more complicated though -- the line-count roughly doubled in size, I think. I also tried implementing a more direct/naive "merge the groups after making them" approach, and that ended up being similarly complex.

Not sure if that's something we're ok with or not: this feels like a lot of code for what I suspect is ultimately a very rare edge case. Maybe it might be better to remove the changes I made and keep this TODO to try and help minimize the maintenance burden of these changes? LMK what you think.

test-data/unit/check-enum.test
Michael0x2a: Done!

I also refactored out the grouping algorithm into a standalone function and added several unit tests for it to `testinfer.py`. (Not sure is this is the right home for these tests though. LMK if you want me to move them.)

mypy/checker.py
Michael0x2a: This chunk of code starts on line 3853. I guess the diff ended up being a little messy because I renamed the variables ("left_type" -> "item_type" and "right_type" -> "collection_type") and moved the logic around a bit.

But we don't do anything really special for this case: we treat `a in b in c` as if it were `a in b and b in c` and only narrow away Optionals from the LHS item type if it overlaps with whatever's inside the RHS collection.

So in the end, I don't think we've really improved support for this pattern.

mypy/checker.py
Michael0x2a: It does! I added a test to `check-optional.test`.

mypy/checker.py
Michael0x2a: Currently, no -- but I did spend some time trying to construct one and eventually ended up convincing myself that we'll always get the same result no matter which index we pick. I updated the comment to include the reasoning.

mypy/checker.py
Michael0x2a: I think using the word "combine" would be a bit ambiguous because it'd be unclear whether the combining step will end up "and"-ing or "or"-ing the the two maps.

Basically, the way I was thinking about this change is that there are now two kinds of TypeMaps -- full ones, which represent all info we know about types within a certain "context", and partial ones which contain only some of the info.

Partial TypeMaps also only exist as an implementation detail of the narrowing logic: it always ends up returning full TypeMaps.

So if `and_conditional_maps` and `or_conditional_maps` functions are for full TypeMaps, I was thinking it made sense to add `and_partial_conditional_maps` and `or_partial_conditional_maps` functions for the partial ones.

But it *also* turned out that `and_conditional_maps` and `and_partial_conditional_maps` behave in the exact same way/would share the same implementation, so I didn't bother defining the latter.

The other approach I could take is to *not* define the `or_partial_conditional_maps` and instead just rewrite `reduce_partial_conditional_maps` function to produce the output maps more directly and do the (pseudo)-intersecting and unioning itself.

That works just as well (and is actually probably slightly more efficient), but does end up making it harder to see how the function relates to the existing conditional map logic.

mypy/checker.py
Michael0x2a: It seems yes -- if I modify `and_conditional_maps` to bias towards picking elements from the left map, this ends up breaking the `testEnumReachabilityWithMultipleEnums` test. Basically, when we do:

```python
class Foo(Enum):
    A = 1
    B = 2
class Bar(Enum):
    A = 1
    B = 2

x3: Union[Foo, Bar]
if x3 is Foo.A or x3 is Bar.A:
    reveal_type(x3)
else:
    reveal_type(x3)
```

...we end up inferring the less precise type of `Union[Literal[Foo.B], Bar]` instead of `Union[Literal[Foo.B], Literal[Bar.B]]` in the else case -- the less precise narrowing we get from `x3 is Foo.A` overrode anything we learned in the next clause.

mypy/checker.py
```diff
@@ -4953,6 +5233,192 @@ def nothing() -> Iterator[None]:
...
+TKey = TypeVar('TKey')
+TValue = TypeVar('TValue')
+
+
+class DisjointDict(Generic[TKey, TValue]):
+    """An variation of the union-find algorithm/data structure where instead of
+    keeping track of just disjoint sets, we keep track of disjoint dicts.
+
+    That is, we keep track of multiple Set[Keys] -> Set[Values] mappings, where each
+    mapping's keys are guaranteed to be disjoint.
```
ilevkivskyi: Add short motivation here for why the naive algorithm is not enough.

Issue #7944, opened 2019-11-13
Narrowing unions with Literal and TypedDict


porglezomp:
I expected the following code to narrow the type of the argument, but it is unable to.
```python
from mypy_extensions import TypedDict
from typing_extensions import Literal

TaggedStr = TypedDict('TaggedStr', {'tag': Literal['str'], 's': str})
TaggedInt = TypedDict('TaggedInt', {'tag': Literal['int'], 'i': int})

def do_tagged(x: Union[TaggedStr, TaggedInt]):
    if x['tag'] == 'str':
        reveal_type(x)
    else:
        reveal_type(x)
```

* What is the actual behavior/output?
No narrowing occurs:
```
example.py:10: note: Revealed type is 'Union[TypedDict('example.TaggedStr', {'tag': Literal['str'], 's': builtins.str}), TypedDict('example.TaggedInt', {'tag': Literal['int'], 'i': builtins.int})]'
example.py:12: note: Revealed type is 'Union[TypedDict('example.TaggedStr', {'tag': Literal['str'], 's': builtins.str}), TypedDict('example.TaggedInt', {'tag': Literal['int'], 'i': builtins.int})]'
```

* What is the behavior/output you expect?
I'd expect the comparisons against literal keys to allow narrowing the types.
```
example.py:10: note: Revealed type is 'TypedDict('example.TaggedStr', {'tag': Literal['str'], 's': builtins.str})'
example.py:12: note: Revealed type is 'TypedDict('example.TaggedInt', {'tag': Literal['int'], 'i': builtins.int})'
```

* What are the versions of mypy and Python you are using?
mypy 0.740, Python 3.7.4

msullivan:
I feel like this issue must be a duplicate but I can't find what it is a duplicate of. I can, however, find a PR that should fix it: #7169

msullivan:
Also as of mypy master, this works with enums but not yet with literals:
```
from typing import Union
from mypy_extensions import TypedDict
from typing_extensions import Literal
from enum import Enum

class Key(Enum):
    Str = "str"
    Int = "int"


TaggedStr = TypedDict('TaggedStr', {'tag': Literal[Key.Str], 's': str})
TaggedInt = TypedDict('TaggedInt', {'tag': Literal[Key.Int], 'i': int})

def do_tagged(x: Union[TaggedStr, TaggedInt]):
    if x['tag'] is Key.Str:
        reveal_type(x)
    else:
        reveal_type(x)
```

PR #8151, opened 2019-12-16
Add support for narrowing Literals using equality

Michael0x2a:
This pull request (finally) adds support for narrowing expressions using Literal types by equality, instead of just identity. For example, the following "tagged union" pattern is now supported:

```python
class Foo(TypedDict):
    key: Literal["A"]
    blah: int

class Bar(TypedDict):
    key: Literal["B"]
    something: str

x: Union[Foo, Bar]
if x.key == "A":
    reveal_type(x)  # Revealed type is 'Foo'
else:
    reveal_type(x)  # Revealed type is 'Bar'
```

Previously, this was possible to do only with Enum Literals and the `is` operator, which is perhaps not very intuitive.

The main limitation with this pull request is that it'll perform narrowing only if either the LHS or RHS contains an explicit Literal type somewhere. If this limitation is not present, we end up breaking a decent amount of real-world code -- mostly tests -- that do something like this:

```python
def some_test_case() -> None:
    worker = Worker()

    # Without the limitation, we narrow 'worker.state' to
    # Literal['ready'] in this assert...
    assert worker.state == 'ready'

    worker.start()

    # ...which subsequently causes this second assert to narrow
    # worker.state to <uninhabited>, causing the last line to be
    # unreachable.
    assert worker.state == 'running'
    worker.query()
```

I tried for several weeks to find a more intelligent way around this problem, but everything I tried ended up being either insufficient or super-hacky, so I gave up and went for this brute-force solution.

The other main limitation is that we perform narrowing only if both the LHS and RHS do not define custom `__eq__` or `__ne__` methods, but this seems like a more reasonable one to me.

Resolves https://github.com/python/mypy/issues/7944.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -6,7 +6,7 @@ from contextlib import contextmanager
 
 from typing import (
     Dict, Set, List, cast, Tuple, TypeVar, Union, Optional, NamedTuple, Iterator, Iterable,
-    Sequence, Mapping, Generic, AbstractSet
+    Sequence, Mapping, Generic, AbstractSet, Callable
 )
 from typing_extensions import Final
 
@@ -50,7 +50,8 @@ from mypy.typeops import (
     erase_def_to_union_or_bound, erase_to_union_or_bound, coerce_to_literal,
     try_getting_str_literals_from_type, try_getting_int_literals_from_type,
     tuple_fallback, is_singleton_type, try_expanding_enum_to_union,
-    true_only, false_only, function_type, TypeVarExtractor,
+    true_only, false_only, function_type, TypeVarExtractor, custom_special_method,
+    is_literal_type_like,
 )
 from mypy import message_registry
 from mypy.subtypes import (
@@ -3890,20 +3891,59 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
 
             partial_type_maps = []
             for operator, expr_indices in simplified_operator_list:
-                if operator in {'is', 'is not'}:
-                    if_map, else_map = self.refine_identity_comparison_expression(
-                        operands,
-                        operand_types,
-                        expr_indices,
-                        narrowable_operand_index_to_hash.keys(),
-                    )
-                elif operator in {'==', '!='}:
-                    if_map, else_map = self.refine_equality_comparison_expression(
-                        operands,
-                        operand_types,
-                        expr_indices,
-                        narrowable_operand_index_to_hash.keys(),
-                    )
+                if operator in {'is', 'is not', '==', '!='}:
+                    # is_valid_target:
+                    #   Controls which types we're allowed to narrow exprs to. Note that
+                    #   we cannot use 'is_literal_type_like' in both cases since doing
+                    #   'x = 10000 + 1; x is 10001' is not always True in all Python impls.
+                    #
+                    # coerce_only_in_literal_context:
+                    #   If true, coerce types into literal types only if one or more of
+                    #   the provided exprs contains an explicit Literal type. This could
+                    #   technically be set to any arbitrary value, but it seems being liberal
+                    #   with narrowing when using 'is' and conservative when using '==' seems
+                    #   to break the least amount of real-world code.
+                    #
+                    # should_narrow_by_identity:
+                    #   Set to 'false' only if the user defines custom __eq__ or __ne__ methods
+                    #   that could cause identity-based narrowing to produce invalid results.
+                    if operator in {'is', 'is not'}:
+                        is_valid_target = is_singleton_type    # type: Callable[[Type], bool]
+                        coerce_only_in_literal_context = False
+                        should_narrow_by_identity = True
+                    else:
+                        is_valid_target = is_exactly_literal_type
+                        coerce_only_in_literal_context = True
+
+                        def has_no_custom_eq_checks(t: Type) -> bool:
+                            return not custom_special_method(t, '__eq__', check_all=False) \
+                                   and not custom_special_method(t, '__ne__', check_all=False)
+                        expr_types = [operand_types[i] for i in expr_indices]
+                        should_narrow_by_identity = all(map(has_no_custom_eq_checks, expr_types))
+
+                    if_map = {}    # type: TypeMap
+                    else_map = {}  # type: TypeMap
+                    if should_narrow_by_identity:
+                        if_map, else_map = self.refine_identity_comparison_expression(
+                            operands,
+                            operand_types,
+                            expr_indices,
+                            narrowable_operand_index_to_hash.keys(),
+                            is_valid_target,
+                            coerce_only_in_literal_context,
+                        )
+
+                    # Strictly speaking, we should also skip this check if the objects in the expr
+                    # chain have custom __eq__ or __ne__ methods. But we (maybe optimistically)
+                    # assume nobody would actually create a custom objects that considers itself
+                    # equal to None.
+                    if if_map == {} and else_map == {}:
+                        if_map, else_map = self.refine_away_none_in_comparison(
+                            operands,
+                            operand_types,
+                            expr_indices,
+                            narrowable_operand_index_to_hash.keys(),
+                        )
                 elif operator in {'in', 'not in'}:
                     assert len(expr_indices) == 2
                     left_index, right_index = expr_indices
@@ -4146,8 +4186,10 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                                               operand_types: List[Type],
                                               chain_indices: List[int],
                                               narrowable_operand_indices: AbstractSet[int],
+                                              is_valid_target: Callable[[ProperType], bool],
+                                              coerce_only_in_literal_context: bool,
                                               ) -> Tuple[TypeMap, TypeMap]:
-        """Produces conditional type maps refining expressions used in an identity comparison.
+        """Produces conditional type maps refining exprs used in an identity/equality comparison.
 
         The 'operands' and 'operand_types' lists should be the full list of operands used
         in the overall comparison expression. The 'chain_indices' list is the list of indices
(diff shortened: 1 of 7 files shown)


Michael0x2a:
Also -- this PR was written on top of https://github.com/python/mypy/pull/8148. I can rebase this once the other PR lands.

I also ran this on one of our internal codebases and confirmed there were no new errors, with the exception of one or two (legitimate) "unnecessary cast" errors.

fluggo:
I'm glad to see this landing.

My comment on your second type-narrowing sample, where checking the state limits the operations available, I would argue that the `Worker` type shouldn't be written to narrow its attributes based on the state. The `query()` method almost certainly doesn't cease to exist when `worker.state == 'ready'`. I feel like that's abusing the type system; type narrowing is there to clarify the type of `worker` as of its last assignment, and not to try to keep up with how the type might change itself as time goes by.

TypeScript doesn't support that use case, and I don't feel like you should have to, either.

Michael0x2a:
@fluggo -- ah, sorry. To clarify, the last example isn't attempting to narrow the type of `worker` in any way, just `worker.state`.

Basically, the core problem is that just blindly applying the way mypy normally narrows variables ends up causing too many issues. For example, here's how mypy currently narrows the following program:

```python
class A: pass
class B: pass

def test(x: object) -> None:
    assert isinstance(x, A)
    reveal_type(x)  # Revealed type is 'A'

    assert isinstance(x, B)
    reveal_type(x)  # No output: stmt is unreachable; mypy thinks the assert is always false
```

In this case, the behavior is mostly reasonable. While 'x' could theoretically be a type that multiply inherits from both A and B, it usually isn't, so concluding that the final assert will always be False and the last line to be unreachable is a reasonable heuristic. (And when/if mypy starts understanding the notion of intersection types, it could maybe infer x is an intersection of A and B, but whatever, that's a separate issue.)

So, since Literal types are supposed to be treated more or less as regular subtypes of the underlying type, it might initially make sense to do the same thing there:

```python
def test(x: int) -> None:
    assert x == 10
    reveal_type(x)  # Suppose we narrowed so the revealed type is Literal[10]

    assert x == 20  # Then, mypy would conclude x can't be both 10 and 20
    reveal_type(x)  # ...and since the assert is always false, this is unreachable
```

Attempting to narrow is reasonable in this specific case, since the last assert genuinely *will* always be false, since there's no way for the value of 'x' to abruptly change in after the first assert.

But the expression we're narrowing *can* abruptly change if it's, say, some object attribute instead of just a variable. The 'worker' example I had was just trying to show a realistic example of where this happens: a method changes some internal field, and we check the new value is what we expected in some test case.

This "what if a method changes an attribute and causes our narrowing to be bogus" problem also appears when narrowing regular types with `isinstance(...)` as well. It's just that it's (probably) pretty uncommon for real-world objects to have some field of type `Union[A, B]` and keep dynamically switching from one to another.

That is, while our current heuristics for type-level narrowing are sometimes mildly problematic, attempting apply them for value-level narrowing ends up being actually problematic.

So we need to adjust how we do value-level narrowing in some way (disable them for asserts? change the behavior of asserts so they override instead of narrowing? do the narrowing only if the attribute is `Final`? infer a pseudo instance/Literal hybrid? give up and tell users to use enums if they want this sort of narrowing?), and I ended up going with the "require the LHS or RHS to explicitly use Literals" approach.

And in practice, this ends up meaning that the thing you're narrowing needs to be a Union of Literals. (This conveniently happens to be just what you need if you want tagged unions w/o having to use enum everywhere.)

That said, I'm not really sure whether this is good enough either. Sure, it'll ensure that the 'worker' example doesn't end up with any unreachable statements if `worker.state` was originally defined to be type `str`. But if it were defined to be type `Literal["ready", "running", "paused", "done"]` or something, we'd end up right back where we started with the last assert being inferred to be always False.

¯\_(ツ)_/¯

fluggo:
I see what you're saying now. I ran the same sample through TypeScript, and I was surprised to find that the same example fails there:

```typescript
class Worker {
  state: 'ready' | 'running';

  start() {
    this.state = 'running';
  }

  query() {
  }
}

function someTestCase(): void {
  const worker = new Worker();

  if(worker.state !== 'ready')
    throw new Error('');

  worker.start();

  if(worker.state !== 'running')   // Error: This condition will always return 'true' since the types '"ready"' and '"running"' have no overlap. ts(2367)
    throw new Error('');

  worker.query();
}
```

It seems the solution in TypeScript is exactly as you proposed: don't define a "status"-like field as type Literal. TypeScript itself follows this principle for (XmlHttpRequest)[https://github.com/microsoft/TypeScript/blob/master/src/lib/dom.generated.d.ts#L18829], which defines the `readyState` as type `number` instead of the literals it could be. TypeScript only uses literals in static data types or for union discrimination.

ilevkivskyi:
I will review this right after https://github.com/python/mypy/pull/8148 is merged (currently it is a bit too large to review, sorry)

Michael0x2a:
@ilevkivskyi -- fyi, I finished rebasing this PR

mypy/checker.py
```diff
@@ -3890,20 +3891,59 @@ def find_isinstance_check_helper(self, node: Expression) -> Tuple[TypeMap, TypeM
...
+                        is_valid_target = is_exactly_literal_type
+                        coerce_only_in_literal_context = True
+
+                        def has_no_custom_eq_checks(t: Type) -> bool:
+                            return not custom_special_method(t, '__eq__', check_all=False) \
+                                   and not custom_special_method(t, '__ne__', check_all=False)
+                        expr_types = [operand_types[i] for i in expr_indices]
+                        should_narrow_by_identity = all(map(has_no_custom_eq_checks, expr_types))
+
+                    if_map = {}    # type: TypeMap
```
ilevkivskyi: Two spaces before `#`
```suggestion
                    if_map = {}  # type: TypeMap
```

mypy/checker.py
```diff
@@ -4146,8 +4186,10 @@ def refine_identity_comparison_expression(self,
                                               operand_types: List[Type],
                                               chain_indices: List[int],
                                               narrowable_operand_indices: AbstractSet[int],
+                                              is_valid_target: Callable[[ProperType], bool],
+                                              coerce_only_in_literal_context: bool,
                                               ) -> Tuple[TypeMap, TypeMap]:
-        """Produces conditional type maps refining expressions used in an identity comparison.
+        """Produces conditional type maps refining exprs used in an identity/equality comparison.
```
```suggestion
        """Produce conditional type maps refining expressions by an identity/equality comparison.
```

mypy/checker.py
```diff
@@ -5540,3 +5597,9 @@ def has_bool_item(typ: ProperType) -> bool:
         return any(is_named_instance(item, 'builtins.bool')
                    for item in typ.items)
     return False
+
+
+# TODO: why can't we define this as an inline function?
+# Does mypyc not support them?
```
ilevkivskyi: Do you mean lambda by "inline function"? What exactly goes wrong when you try?

mypy/typeops.py
```diff
@@ -655,7 +674,7 @@ def coerce_to_literal(typ: Type) -> ProperType:
             enum_values = get_enum_values(typ)
             if len(enum_values) == 1:
                 return LiteralType(value=enum_values[0], fallback=typ)
-    return typ
+    return original_type
```
ilevkivskyi: This is right, thanks!

test-data/unit/check-narrowing.test
```diff
@@ -1,3 +1,86 @@
...
+else:
+    reveal_type(x4)         # N: Revealed type is 'Tuple[Literal['B'], builtins.str]'
+    reveal_type(x4[0])      # N: Revealed type is 'Literal['B']'
+
+x5: Union[TypedDict1, TypedDict2]
+if x5["key"] == "A":
+    reveal_type(x5)         # N: Revealed type is 'TypedDict('__main__.TypedDict1', {'key': Literal['A'], 'foo': builtins.int})'
+else:
+    reveal_type(x5)         # N: Revealed type is 'TypedDict('__main__.TypedDict2', {'key': Literal['B'], 'foo': builtins.str})'
+[builtins fixtures/primitives.pyi]
```
ilevkivskyi: Maybe add couple more tests with typed dicts (for example nested typed dicts), since this is pretty important use case.

test-data/unit/check-narrowing.test
```diff
@@ -445,3 +528,225 @@ if y["model"]["key"] is Key.C:
...
+from typing_extensions import Literal, Final
+
+A_final: Final = "A"
+A_literal: Literal["A"]
+
+# Neither the LHS nor the RHS are explicit literals, so regrettably nothing
+# is narrowed here -- see 'testNarrowingEqualityFlipFlop' for an example of
+# why more precise inference here is problematic.
+x_str: str
+if x_str == A_final:
```
ilevkivskyi: I don't think I understand why literal type is "better" than a final name. IMO final names should behave exactly as if they were replaced by their values.

test-data/unit/check-narrowing.test
```diff
@@ -445,3 +528,225 @@ if y["model"]["key"] is Key.C:
...
+    reveal_type(x)   # N: Revealed type is 'Union[Literal[1], Literal[2], None]'
+    reveal_type(y)   # N: Revealed type is '__main__.Custom'
+
+# No contamination here
+if 1 == x == z:      # E: Non-overlapping equality check (left operand type: "Union[Literal[1], Literal[2], None]", right operand type: "Default")
+    reveal_type(x)   # E: Statement is unreachable
+    reveal_type(z)
+else:
+    # TODO: Why do we narrow away 'Literal[1]' here?
+    # Even if the equality comparison is bogus, we should try and do better here.
```
ilevkivskyi: This is a bit worrying, so I think this deserves a follow-up issue.

test-data/unit/check-narrowing.test
Michael0x2a: If it's any consolation, we also won't narrow anything away if we do `if x_str == "A"` -- I updated the test case with an example.

Basically, I found trying to narrow when we do `x_str == "A"` often ended up causing too many false positives, which is why I settled on the heuristic of requiring either the LHS or the RHS to be a `Literal[...]` of some sort before we attempt narrowing.

test-data/unit/check-narrowing.test
Michael0x2a: Well, this is a bit embarrassing. I did some more digging to try and it turned out that the whole concept of "partial TypeMaps" I introduced in my earlier diff was incorrect and that several of the narrowing test cases I added were too optimistic.

Basically, I incorrectly made mypy narrow the else case in programs like:

```
x: Literal[1, 2]
y: Literal[1, 2]
if x == 2 == y:
    reveal_type(x)  # Revealed type is Literal[2]
    reveal_type(y)  # Revealed type is Literal[2]
else:
    reveal_type(x)  # Revealed type is incorrectly Literal[1]!
    reveal_type(y)  # Revealed type is incorrectly Literal[1]!
```

But actually, we could enter the else case if x == 2 and y == 1 or vice versa, so narrowing was incorrect.

Going back to using just `or_conditional_maps` instead of `or_partial_conditional_maps` ended up fixing these bogus narrowings. I also added a few more test cases + fixed some earlier enum ones.

mypy/checker.py
Michael0x2a: Hmm, I actually don't remember. I'll try experimenting with this a little later today and see if I can repro.

Michael0x2a: Hmm, yeah, I don't seem to be able to repro this anymore? Maybe it was fixed after I rebased or maybe I just doing something wrong before, but defining an inline function seems to be working fine now.

Defining a lambda unfortunately causes flake8 to complain -- it doesn't like it when you assign a lambda to a variable.

Anyways, I moved this function back to where it's being used.

test-data/unit/check-narrowing.test
Michael0x2a: Forgot to mention this earlier, but I added a few more tests, mostly copying some of the existing ones I had for enums and objects. LMK if there are more test cases you'd like to see.

Michael0x2a:
This should be ready for a second look whenever fyi.

test-data/unit/check-enum.test
```diff
@@ -978,30 +978,41 @@ class Foo(Enum):
...
+    reveal_type(x)   # N: Revealed type is 'Literal[__main__.Foo.A]'
+    reveal_type(y)   # N: Revealed type is 'Literal[__main__.Foo.A]'
+elif x is y is Foo.B:
+    reveal_type(x)   # N: Revealed type is 'Literal[__main__.Foo.B]'
+    reveal_type(y)   # N: Revealed type is 'Literal[__main__.Foo.B]'
 else:
-   reveal_type(x)   # N: Revealed type is 'Literal[__main__.Foo.B]'
-   reveal_type(y)   # N: Revealed type is 'Literal[__main__.Foo.B]'
+    reveal_type(x)      # N: Revealed type is '__main__.Foo'
+    reveal_type(y)      # N: Revealed type is '__main__.Foo'
```
ilevkivskyi: Spacing before comments is kind of inconsistent here and below.

PR #8152, opened 2019-12-16
Update docs for Literal types

Michael0x2a:
This pull request is a long-overdue update of the Literal type docs. It:

1. Removes the "this is alpha" warning we have at the top.

2. Mentions Literal enums are a thing (and works in a brief example of one).

3. Adds a section about "intelligent indexing".

4. Adds a section about the "tagged union" pattern (see https://github.com/python/mypy/pull/8151). I made the example focus on the TypedDict/JSON use case -- IMO that's really the only realistically useful use case for the pattern.

5. Cross-references the "tagged union" docs with the TypedDicts docs.

I also thought about making the "Unions of TypedDict" section I added to the TypedDicts doc mention using [pydantic][0] as an alternative to the "tagged union" pattern.

I personally prefer using libraries like this which handle validation and let me use regular classes (and `isinstance`) instead of dicts, but I wasn't sure whether we want to be recommending 3rd party libraries so held off for now. LMK if you think it's worth including it, and I can update the PR.

  [0]: https://pydantic-docs.helpmanual.io

### DIFF
--- a/docs/source/literal_types.rst
+++ b/docs/source/literal_types.rst
@@ -16,8 +16,6 @@ precise type signature for this function using ``Literal[...]`` and overloads:
 
 .. code-block:: python
 
-    # Note: if you are using Python 3.7 or earlier, you will need to import
-    # Literal from the typing_extensions module.
     from typing import overload, Union, Literal
 
     # The first two overloads use Literal[...] so we can
@@ -46,6 +44,13 @@ precise type signature for this function using ``Literal[...]`` and overloads:
 
     variable = True
     reveal_type(fetch_data(variable))    # Revealed type is 'Union[bytes, str]'
+
+.. note::
+
+    The examples in this page import ``Literal`` as well as ``Final`` and
+    ``TypedDict`` from the ``typing`` module. These types were added to
+    ``typing`` in Python 3.8, but are also available for use in Python 2.7
+    and 3.4 - 3.7 via the ``typing_extensions`` package.
 
 Parameterizing Literals
 ***********************
@@ -106,8 +111,6 @@ you can instead change the variable to be ``Final`` (see :ref:`final_attrs`):
 
 .. code-block:: python
 
-    # Note: if you are using Python 3.7 or earlier, you will need to import
-    # Literal and Final from the typing_extensions module.
     from typing import Final, Literal
 
     def expects_literal(x: Literal[19]) -> None: pass
@@ -144,7 +147,7 @@ For example, compare and contrast what happens when you try appending these type
 
 .. code-block:: python
 
-    from typing_extensions import Final, Literal
+    from typing import Final, Literal
 
     a: Final = 19
     b: Literal[19] = 19
@@ -174,6 +177,8 @@ corresponding to some particular index, we can use Literal types like so:
 
 .. code-block:: python
 
+    from typing import TypedDict
+
     tup = ("foo", 3.4)
 
     # Indexing with an int literal gives us the exact type for that index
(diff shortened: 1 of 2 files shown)


docs/source/literal_types.rst
```diff
@@ -23,8 +16,9 @@ precise type signature for this function using ``Literal[...]`` and overloads:
 
 .. code-block:: python
 
-    from typing import overload, Union
-    from typing_extensions import Literal
+    # Note: if you are using Python 3.7 or earlier, you will need to import
+    # Literal from the typing_extensions module.
```
ilevkivskyi: I would make it a proper `.. note::` (for example above at the start), we do this for `Protocol` for example.

docs/source/literal_types.rst
```diff
@@ -115,7 +106,9 @@ you can instead change the variable to be ``Final`` (see :ref:`final_attrs`):
 
 .. code-block:: python
 
-    from typing_extensions import Final, Literal
+    # Note: if you are using Python 3.7 or earlier, you will need to import
+    # Literal and Final from the typing_extensions module.
```
ilevkivskyi: If you add the above one, this one would be unnecessary.

docs/source/literal_types.rst
```diff
@@ -168,6 +161,89 @@ For example, compare and contrast what happens when you try appending these type
...
+    # know exactly what the index is and so will return a less precise type:
+    int_index = 1
+    reveal_type(tup[int_index])  # Revealed type is 'Union[str, float]'
+
+    # But if we use either Literal types or a Final int, we can gain back
+    # the precision we originally had:
+    lit_index: Literal[1] = 1
+    fin_index: Final = 1
+    reveal_type(tup[lit_index])  # Revealed type is 'str'
+    reveal_type(tup[fin_index])  # Revealed type is 'str'
```
ilevkivskyi: Maybe add an example with TypedDict? Maybe also an example with union of literals, something along these lines:
```python
i: Literal[1, 2]
t: Tuple[str, int, int]
reveal_type(t[i])  # This is "int"
```

docs/source/literal_types.rst
```diff
@@ -168,6 +161,89 @@ For example, compare and contrast what happens when you try appending these type
...
+possible to use ``isinstance`` to distinguish between two different TypedDicts since
+at runtime, your variable will simply be just a dict.
+
+Instead, what you can do is *label* or *tag* your TypedDicts with a distinct Literal
+type. Then, you can discriminate between each kind of TypedDict by checking the label:
+
+.. code-block:: python
+
+    # Note: if you are using Python 3.7 or earlier, you will need to import
+    # Literal and TypedDict from the typing_extensions module.
```
ilevkivskyi: Again, it is not necessary to repeat everywhere if there is a dedicated note at the top.

docs/source/literal_types.rst
```diff
@@ -168,6 +161,89 @@ For example, compare and contrast what happens when you try appending these type
...
+        #
+        # This in turns narrows the type of 'event' to either NewJobEvent 
+        # or CancelJobEvent.
+        if event["tag"] == "new-job":
+            print(event["job_name"])
+        else:
+            print(event["job_id"])
+
+While this feature is mostly useful when working with TypedDicts, you can also
+use the same technique wih regular objects, tuples, or namedtuples.
```
ilevkivskyi: Mention that tags can also be proper enum values (using `is`) and classes (using `isinstance()`), not just strings.

PR #8153, opened 2019-12-16
[mypyc]Implement str.split primitive

TH3CHARLie:
relates mypyc/mypyc#644

Generate three `method_op`s corresponding to three different ways of calling `str.split`


TH3CHARLie:
It's weird that my custom test can generate for str.split can generate correct result while the mypy compiled using this modified-version mypyc is clearly broken because of str.split

TH3CHARLie:
> It's weird that my custom test can generate for str.split can generate correct result while the mypy compiled using this modified-version mypyc is clearly broken because of str.split

I think I've located the bug, it's because of incorrect int literal value, for example:
``` python
s.split(' ', 1)
```
directly passing `int_rprimitive` into `simple_emit` would generate the follwoing C code:

```C
CPyL12: ;
    cpy_r_r30 = CPyStatic_unicode_4; /* ' ' */
    cpy_r_r31 = 2;
    cpy_r_r32 = PyUnicode_Split(cpy_r_r29, cpy_r_r30, cpy_r_r31);
```
clearly, the value of the `CPyTagged` is doubled compared to the target size to be passed in, so instead I write an ad-hoc function to generate code for this case.

Now I think it is review-ready, @msullivan

mypyc/ops_str.py
```diff
@@ -31,6 +32,27 @@
...
+    emitter.emit_lines(
+        '%s = CPyTagged_AsSsize_t(%s);' % (temp, args[2]),
+        '%s = PyUnicode_Split(%s, %s, %s);' % (dest, args[0], args[1], temp))
+
+
+str_split_types = [str_rprimitive, str_rprimitive, int_rprimitive]  # type: List[RType]
+str_split_emits = [simple_emit('{dest} = PyUnicode_Split({args[0]}, NULL, -1);'),
+                   simple_emit('{dest} = PyUnicode_Split({args[0]}, {args[1]}, -1);'),
+                   emit_str_split]  # type: List[EmitCallback]
+for i in range(3):
```
msullivan: Nit: `len(str_split_types)`

mypyc/ops_str.py
```diff
@@ -31,6 +32,27 @@
...
+    method_op(
+        name='split',
+        arg_types=str_split_types[0:i+1],
+        result_type=object_rprimitive,
```
msullivan: Making this `list_rprimitive` will let us generate slightly better code. Otherwise we'll emit a cast on the result to make sure it is actually a list.

mypyc/ops_str.py
```diff
@@ -31,6 +32,27 @@
     error_kind=ERR_MAGIC,
     emit=simple_emit('{dest} = PyUnicode_Join({args[0]}, {args[1]});'))
 
+
+def emit_str_split(emitter: EmitterInterface, args: List[str], dest: str) -> None:
+    temp = emitter.temp_name()
+    emitter.emit_declaration('Py_ssize_t %s;' % temp)
+    emitter.emit_lines(
+        '%s = CPyTagged_AsSsize_t(%s);' % (temp, args[2]),
+        '%s = PyUnicode_Split(%s, %s, %s);' % (dest, args[0], args[1], temp))
```
msullivan: I think I'd prefer to define a helper function in `CPy.h` and then call that for this case. (The helper function could more easily handle the error case where the number is too big, also).

TH3CHARLie:
Thanks for the review, I've updated it in the newest commit but I'm not quite sure about the helper function, especially about its error message.

TH3CHARLie:
@msullivan One more thing, do you think the coding pattern in this PR can be extend to more primitives(like `join`, `replace`, `startswith`, `endswith`)? If so, I'd be happy to submit a PR finishing all these four primitives.

JukkaL:
I would be happy to get additional native implementations of common primitives, at least if the implementations are relatively simple. Beyond the most obvious ones (that have direct C API counterparts and are used all over the place) it may be worth discussing these in an issue first to determine which should have the highest priority.

`join`, `replace`, `startswith` and `endswith` seem no-brainers to me as they are very common. However, less common calling conventions such as `x.startswith(('x', 'y'))` may not be worth it yet (not sure).

TH3CHARLie:
> I would be happy to get additional native implementations of common primitives, at least if the implementations are relatively simple. Beyond the most obvious ones (that have direct C API counterparts and are used all over the place) it may be worth discussing these in an issue first to determine which should have the highest priority.

Good point, please raise an issue in either mypy or mypyc to track this, I'd be happy to contribute as much as I can.

JukkaL:
We can add more primitives to https://github.com/mypyc/mypyc/issues/644 for now. If it becomes too big, we can split it into smaller issues.

PR #8157, opened 2019-12-17
Fix some daemon crashes involving classes becoming generic

### LINKED ISSUE #3279: Fine grained crash when class is turned into a generic class
`a.py`:

```py
import b
def f() -> b.C: pass
```

Original `b.py`:

```py
import a
class C: pass
```

Updated `b.py`:

```py
import a

T = TypeVar('T')
class C(Generic[T]): pass

reveal_type(a.f)
c: C[int]
d = a.f()
d = c
c = d
```

Now if I run `mypy --quick a.py` first with the original files, then with the updated `b.py`, I get a crash:

```
$ mypy --quick --show-traceback a.py
b.py:7: error: Revealed type is 'def () -> b.C'
/Users/jukka/src/mypy/b.py:11: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.510-dev-c856a53fe06c0ad9193d86ab35fb86693940f273-dirty
  File "/Users/jukka/src/mypy/scripts/mypy", line 6, in <module>
    main(__file__)
  File "/Users/jukka/src/mypy/mypy/main.py", line 46, in main
  File "/Users/jukka/src/mypy/mypy/main.py", line 93, in type_check_only
    options=options)
  File "/Users/jukka/src/mypy/mypy/build.py", line 188, in build
  File "/Users/jukka/src/mypy/mypy/build.py", line 1570, in dispatch
    process_graph(graph, manager)
  File "/Users/jukka/src/mypy/mypy/build.py", line 1813, in process_graph
  File "/Users/jukka/src/mypy/mypy/build.py", line 1912, in process_stale_scc
  File "/Users/jukka/src/mypy/mypy/build.py", line 1485, in type_check_first_pass
    self.type_checker.check_first_pass()
  File "/Users/jukka/src/mypy/mypy/checker.py", line 177, in check_first_pass
    self.accept(d)
  File "/Users/jukka/src/mypy/mypy/checker.py", line 262, in accept
    stmt.accept(self)
  File "/Users/jukka/src/mypy/mypy/nodes.py", line 853, in accept
  File "/Users/jukka/src/mypy/mypy/checker.py", line 1177, in visit_assignment_stmt
  File "/Users/jukka/src/mypy/mypy/checker.py", line 1239, in check_assignment
    rvalue_type = self.check_simple_assignment(lvalue_type, rvalue, lvalue)
  File "/Users/jukka/src/mypy/mypy/checker.py", line 1698, in check_simple_assignment
    rvalue_type = self.expr_checker.accept(rvalue, lvalue_type)
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 2056, in accept
    typ = node.accept(self)
  File "/Users/jukka/src/mypy/mypy/nodes.py", line 1213, in accept
    return visitor.visit_name_expr(self)
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 123, in visit_name_expr
    return self.narrow_type_from_binder(e, result)
  File "/Users/jukka/src/mypy/mypy/checkexpr.py", line 2271, in narrow_type_from_binder
    ans = narrow_declared_type(known_type, restriction)
  File "/Users/jukka/src/mypy/mypy/meet.py", line 46, in narrow_declared_type
    return meet_types(declared, narrowed)
  File "/Users/jukka/src/mypy/mypy/meet.py", line 25, in meet_types
    return t.accept(TypeMeetVisitor(s))
  File "/Users/jukka/src/mypy/mypy/types.py", line 381, in accept
  File "/Users/jukka/src/mypy/mypy/meet.py", line 198, in visit_instance
    args.append(self.meet(t.args[i], si.args[i]))
IndexError: list index out of range
/Users/jukka/src/mypy/b.py:11: note: use --pdb to drop into pdb
```

msullivan:
Fixes #3279. Also fixes another related crash.

### DIFF
--- a/mypy/constraints.py
+++ b/mypy/constraints.py
@@ -347,29 +347,33 @@ class ConstraintBuilderVisitor(TypeVisitor[List[Constraint]]):
                     template.type.has_base(instance.type.fullname)):
                 mapped = map_instance_to_supertype(template, instance.type)
                 tvars = mapped.type.defn.type_vars
-                for i in range(len(instance.args)):
+                # N.B: We use zip instead of indexing because the lengths might have
+                # mismatches during daemon reprocessing.
+                for tvar, mapped_arg, instance_arg in zip(tvars, mapped.args, instance.args):
                     # The constraints for generic type parameters depend on variance.
                     # Include constraints from both directions if invariant.
-                    if tvars[i].variance != CONTRAVARIANT:
+                    if tvar.variance != CONTRAVARIANT:
                         res.extend(infer_constraints(
-                            mapped.args[i], instance.args[i], self.direction))
-                    if tvars[i].variance != COVARIANT:
+                            mapped_arg, instance_arg, self.direction))
+                    if tvar.variance != COVARIANT:
                         res.extend(infer_constraints(
-                            mapped.args[i], instance.args[i], neg_op(self.direction)))
+                            mapped_arg, instance_arg, neg_op(self.direction)))
                 return res
             elif (self.direction == SUPERTYPE_OF and
                     instance.type.has_base(template.type.fullname)):
                 mapped = map_instance_to_supertype(instance, template.type)
                 tvars = template.type.defn.type_vars
-                for j in range(len(template.args)):
+                # N.B: We use zip instead of indexing because the lengths might have
+                # mismatches during daemon reprocessing.
+                for tvar, mapped_arg, template_arg in zip(tvars, mapped.args, template.args):
                     # The constraints for generic type parameters depend on variance.
                     # Include constraints from both directions if invariant.
-                    if tvars[j].variance != CONTRAVARIANT:
+                    if tvar.variance != CONTRAVARIANT:
                         res.extend(infer_constraints(
-                            template.args[j], mapped.args[j], self.direction))
-                    if tvars[j].variance != COVARIANT:
+                            template_arg, mapped_arg, self.direction))
+                    if tvar.variance != COVARIANT:
                         res.extend(infer_constraints(
-                            template.args[j], mapped.args[j], neg_op(self.direction)))
+                            template_arg, mapped_arg, neg_op(self.direction)))
                 return res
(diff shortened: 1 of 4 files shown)


mypy/constraints.py
```diff
@@ -347,29 +347,33 @@ def visit_instance(self, template: Instance) -> List[Constraint]:
                     template.type.has_base(instance.type.fullname)):
                 mapped = map_instance_to_supertype(template, instance.type)
                 tvars = mapped.type.defn.type_vars
-                for i in range(len(instance.args)):
+                # N.B: We use zip instead of indexing because the lengths might have
+                # mismatches during daemon reprocessing.
+                for tvar, mapped_arg, instance_arg in zip(tvars, mapped.args, instance.args):
```
JukkaL: It's kind of unfortunate that we have to do this. Another option might be to make `args` a property and have it adjust the number of args dynamically if the `TypeInfo` has changed. This could have performance impliciations, however.

msullivan:
I didn't do join_instances because I couldn't manage to make it crash and because I had convinced myself from a glance that it couldn't break because of some other side condition, but I don't think that is actually true.

I don't love the idea of adjusting things as an `args` property, which feels too magic and has performance implications.

mypy/meet.py
```diff
@@ -486,13 +486,15 @@ def visit_type_var(self, t: TypeVarType) -> ProperType:
     def visit_instance(self, t: Instance) -> ProperType:
         if isinstance(self.s, Instance):
             si = self.s
-            if t.type == si.type:
+            if t.type == si.type and len(t.args) == len(si.args):
```
ilevkivskyi: Btw why do we need the `and ...` here if we anyway use `zip()` below?

msullivan: Oh, no, we definitely do not. The `and` was my first approach for fixing this, before I decided I liked the zip version more. (This check does work, though, so that could be an approach?)

PR #8167, opened 2019-12-18
Support type inference for defaultdict()

JukkaL:
This allows inferring type of `x`, for example:

```
from collections import defaultdict

x = defaultdict(list)  # Infer defaultdict[str, List[int]]
x['foo'].append(1)
```

The implemention is not pretty and we have probably
reached about the maximum reasonable level of special
casing in type inference now.

There is a hack to work around the problem with leaking
type variable types in nested generics calls (I think).
This will break some (likely very rare) use cases.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2813,7 +2813,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             partial_type = PartialType(None, name)
         elif isinstance(init_type, Instance):
             fullname = init_type.type.fullname
-            if (isinstance(lvalue, (NameExpr, MemberExpr)) and
+            is_ref = isinstance(lvalue, (NameExpr, MemberExpr))
+            if (is_ref and
                     (fullname == 'builtins.list' or
                      fullname == 'builtins.set' or
                      fullname == 'builtins.dict' or
@@ -2821,6 +2822,15 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                     all(isinstance(t, (NoneType, UninhabitedType))
                         for t in get_proper_types(init_type.args))):
                 partial_type = PartialType(init_type.type, name)
+            elif is_ref and fullname == 'collections.defaultdict':
+                arg0 = get_proper_type(init_type.args[0])
+                arg1 = get_proper_type(init_type.args[1])
+                if (isinstance(arg0, (NoneType, UninhabitedType)) and
+                        isinstance(arg1, Instance) and
+                        self.is_valid_ordereddict_partial_value_type(arg1)):
+                    partial_type = PartialType(init_type.type, name, arg1)
+                else:
+                    return False
             else:
                 return False
         else:
@@ -2828,6 +2838,15 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         self.set_inferred_type(name, lvalue, partial_type)
         self.partial_types[-1].map[name] = lvalue
         return True
+
+    def is_valid_ordereddict_partial_value_type(self, t: Instance) -> bool:
+        if len(t.args) == 0:
+            return True
+        if len(t.args) == 1:
+            arg = get_proper_type(t.args[0])
+            if isinstance(arg, (TypeVarType, UninhabitedType)):  # TODO: This is too permissive
+                return True
+        return False
 
     def set_inferred_type(self, var: Var, lvalue: Lvalue, type: Type) -> None:
         """Store inferred variable type.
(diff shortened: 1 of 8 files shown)


mypy/checker.py
```diff
@@ -2813,14 +2813,24 @@ def infer_partial_type(self, name: Var, lvalue: Lvalue, init_type: Type) -> bool
...
                      fullname == 'collections.OrderedDict') and
                     all(isinstance(t, (NoneType, UninhabitedType))
                         for t in get_proper_types(init_type.args))):
                 partial_type = PartialType(init_type.type, name)
+            elif is_ref and fullname == 'collections.defaultdict':
+                arg0 = get_proper_type(init_type.args[0])
+                arg1 = get_proper_type(init_type.args[1])
+                if (isinstance(arg0, (NoneType, UninhabitedType)) and
+                        isinstance(arg1, Instance) and
+                        self.is_valid_ordereddict_partial_value_type(arg1)):
```
ilevkivskyi: This name is really weird:
* Why `ordereddict` and not `defaultdict`?
* I don't think it should be called `valid`, maybe `empty`?

For example `is_empty_defaultdict_value_type()`.

mypy/checker.py
```diff
@@ -2813,14 +2813,24 @@ def infer_partial_type(self, name: Var, lvalue: Lvalue, init_type: Type) -> bool
...
                      fullname == 'builtins.dict' or
                      fullname == 'collections.OrderedDict') and
                     all(isinstance(t, (NoneType, UninhabitedType))
                         for t in get_proper_types(init_type.args))):
                 partial_type = PartialType(init_type.type, name)
+            elif is_ref and fullname == 'collections.defaultdict':
+                arg0 = get_proper_type(init_type.args[0])
+                arg1 = get_proper_type(init_type.args[1])
+                if (isinstance(arg0, (NoneType, UninhabitedType)) and
+                        isinstance(arg1, Instance) and
```
ilevkivskyi: It would be cleaner to move this check to the below helper.

mypy/checker.py
```diff
@@ -2813,14 +2813,24 @@ def infer_partial_type(self, name: Var, lvalue: Lvalue, init_type: Type) -> bool
             partial_type = PartialType(None, name)
         elif isinstance(init_type, Instance):
             fullname = init_type.type.fullname
-            if (isinstance(lvalue, (NameExpr, MemberExpr)) and
+            is_ref = isinstance(lvalue, (NameExpr, MemberExpr))
```
ilevkivskyi: While you are at it maybe change this to `RefExpr`?

mypy/checker.py
```diff
@@ -2829,6 +2839,19 @@ def infer_partial_type(self, name: Var, lvalue: Lvalue, init_type: Type) -> bool
         self.partial_types[-1].map[name] = lvalue
         return True
 
+    def is_valid_ordereddict_partial_value_type(self, t: Instance) -> bool:
```
ilevkivskyi: Add a short docstring with an example (also the name needs fixing, see above).

mypy/checker.py
```diff
@@ -2829,6 +2839,19 @@ def infer_partial_type(self, name: Var, lvalue: Lvalue, init_type: Type) -> bool
...
+        if len(t.args) == 0:
+            return True
+        if len(t.args) == 1:
+            arg = get_proper_type(t.args[0])
+            # TODO: This is too permissive -- we only allow TypeVarType since
+            #       they leak in cases like defaultdict(list) due to a bug.
+            #       This can result in incorrect types being inferred, but only
+            #       in rare cases.
+            if isinstance(arg, (TypeVarType, UninhabitedType)):
```
ilevkivskyi: Should this also include `NoneType`? Otherwise it may not work if `strict_optional=False` (see the check for key type). Add a test for this.

mypy/types.py
```diff
@@ -1763,13 +1763,21 @@ class PartialType(ProperType):
     # None for the 'None' partial type; otherwise a generic class
     type = None  # type: Optional[mypy.nodes.TypeInfo]
     var = None  # type: mypy.nodes.Var
+    # For partial defaultdict[K, V], the type V (K is unknown). If V is generic,
+    # the type argument is not final and will be replaced later (it's TypeVarType
+    # or UninhabitedType).
+    #
+    # TODO: The type argument can only be TypeVarType due to leaking type variables
```
ilevkivskyi: Should we just store an erased instance here? Then we can remove this TODO and only keep the one in `checker.py`.

mypyc/test-data/fixtures/ir.py
```diff
@@ -26,6 +26,9 @@ class ellipsis: pass
 # Primitive types are special in generated code.
 
 class int:
+    @overload
+    def __init__(self) -> None: pass
+    @overload
```
ilevkivskyi: Why the change to mypyc needed here?

test-data/unit/check-inference.test
```diff
@@ -2976,3 +2976,72 @@ x: Optional[str]
 y = filter(None, [x])
 reveal_type(y)  # N: Revealed type is 'builtins.list[builtins.str*]'
 [builtins fixtures/list.pyi]
+
+[case testPartialDefaultDict]
+from collections import defaultdict
+x = defaultdict(int)
+x[''] = 1
+reveal_type(x) # N: Revealed type is 'collections.defaultdict[builtins.str, builtins.int]'
```
ilevkivskyi: Add a similar test case where value type is generic, e.g.:
```python
x = defaultdict(list)  # No error
x['a'] = [1, 2, 3]
```
and 
```python
x = defaultdict(list)  # Error here
x['a'] = []
```

mypy/checker.py
JukkaL: Yeah. It's here since we rely on the narrowed down type below, so we'd need a type check anyway.

mypy/checker.py
JukkaL: Oops, I'll rename it.

mypy/checker.py
JukkaL: I'll add a test.

mypy/types.py
JukkaL: A good point -- I'll try this out.

mypy/checker.py
JukkaL: Makes sense.

JukkaL:
The typeshed change is accidental, I'll revert it.

PR #8175, opened 2019-12-19
Fix incorrect name lookup for decorated methods

### LINKED ISSUE #8161: mypy incorrectly resolves names inside decorated methods

A bug

```
from typing import Type

class A:
    pass

class B:
    @staticmethod
    def A() -> Type[A]:
        return A

assert A is B.A()
```

```
$ mypy mypy_fail.py
mypy_fail.py:8: error: Function "mypy_fail.B.A" is not valid as a type
mypy_fail.py:8: note: Perhaps you need "Callable[...]" or a callback protocol?
```

I expect this to be accepted - it works fine in python

Probably of interest: regular methods work like I expect. `classmethod` and `property` complain the same way, so I assume this has to do with decorators. For functions, I can't reproduce syntax like this that's also valid python (I get `UnboundLocalError`s), so that's as far as I got with debugging.

mypy `0.750` and python 3.6/3.8. I also get it on git master.

No flags required

TH3CHARLie:
Resolves #8161 

According to comments of `lookup` in `mypy/semanal.py`, when we look up a class attribute, we require that it is defined textually before the reference statement, thus line number is used for comparison.

When function has decorators, its line number is determined by the top decorator instead of the `def`. That's why #8161's code fails because on line 8, the `A` in `Type[A]` has the line number of 8 while the @staticmethod function `A` has the line number of 7 due to the decorator. 

Thus we need to properly handle this by introducing the number of decorators when deciding textural precedence.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3949,7 +3949,9 @@ class SemanticAnalyzer(NodeVisitor[None],
         """
         assert self.statement
         line_diff = self.statement.line - node.line
-        if isinstance(node, Decorator):
+        if isinstance(node, OverloadedFuncDef) and isinstance(self.statement, FuncDef):
+            return node.name != self.statement.name and line_diff > 0
+        elif isinstance(node, Decorator) and not node.is_overload:
             return line_diff > len(node.original_decorators)
         else:
             return line_diff > 0
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3953,16 +3953,20 @@ class SemanticAnalyzer(NodeVisitor[None],
         # The first branch handles reference an overloaded function variant inside itself,
         # this is a corner case where mypy technically deviates from runtime name resolution,
         # but it is fine because we want an overloaded function to be treated as a single unit.
-        if isinstance(node, OverloadedFuncDef) and isinstance(self.statement, FuncDef):
-            return self.is_overloaded_item(node, self.statement) and line_diff > 0
+        if self.is_overloaded_item(node, self.statement):
+            return False
         elif isinstance(node, Decorator) and not node.is_overload:
             return line_diff > len(node.original_decorators)
         else:
             return line_diff > 0
 
-    def is_overloaded_item(self, overload: OverloadedFuncDef, func: FuncDef) -> bool:
+    def is_overloaded_item(self, node: SymbolNode, statement: Statement) -> bool:
         """Check whehter the function belongs to the overloaded variants"""
-        return func in overload.items
+        if isinstance(node, OverloadedFuncDef) and isinstance(statement, FuncDef):
+            return (statement in
+                    [item.func if isinstance(item, Decorator) else item for item in node.items]
+                    or (node.impl is not None and statement is node.impl))
+        return False
 
     def is_defined_in_current_module(self, fullname: Optional[str]) -> bool:
         if fullname is None:
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3942,9 +3942,17 @@ class SemanticAnalyzer(NodeVisitor[None],
                 or (isinstance(node, PlaceholderNode) and node.becomes_typeinfo))
 
     def is_textually_before_statement(self, node: SymbolNode) -> bool:
+        """Check if a node is defined textually before a statement
+
+        Note that decorated functions' line number are the same as
+        the topest decorator.
+        """
         assert self.statement
         line_diff = self.statement.line - node.line
-        return line_diff > 1 or (line_diff == 1 and not isinstance(node, Decorator))
+        if isinstance(node, Decorator):
+            return line_diff > len(node.original_decorators)
+        else:
+            return line_diff > 0
 
     def is_defined_in_current_module(self, fullname: Optional[str]) -> bool:
         if fullname is None:
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3936,10 +3936,15 @@ class SemanticAnalyzer(NodeVisitor[None],
         #       caught.
         assert self.statement  # we are at class scope
         return (node is None
-                or node.line < self.statement.line
+                or self.is_textually_before_statement(node)
                 or not self.is_defined_in_current_module(node.fullname)
                 or isinstance(node, TypeInfo)
                 or (isinstance(node, PlaceholderNode) and node.becomes_typeinfo))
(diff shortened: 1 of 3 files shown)


mypy/semanal.py
```diff
@@ -3936,11 +3936,24 @@ class C:
...
         assert self.statement  # we are at class scope
         return (node is None
-                or node.line < self.statement.line
+                or self.is_textually_before_statement(node)
                 or not self.is_defined_in_current_module(node.fullname)
                 or isinstance(node, TypeInfo)
                 or (isinstance(node, PlaceholderNode) and node.becomes_typeinfo))
 
+    def is_textually_before_statement(self, node: SymbolNode) -> bool:
+        """Check if a node is defined textually before a statement
```
```suggestion
        """Check if a node is defined textually before the current statement
```

test-data/unit/check-classes.test
```diff
@@ -6596,3 +6596,22 @@ reveal_type(D() + "str")  # N: Revealed type is 'Any'
...
+[case testRefMethodWithDecorator]
+from typing import Type
+
+class A:
+    pass
+
+class B:
+    @staticmethod
+    def A() -> Type[A]:
+        return A
```
ilevkivskyi: Could you also add a negative test? Something like:
```python
class A:
    pass

class B:
    @staticmethod
    def A() -> Type[A]: ...
    @staticmethod
    def B() -> Type[A]: ...  # This should be an error
```

test-data/unit/check-classes.test
```diff
@@ -6596,3 +6596,22 @@ reveal_type(D() + "str")  # N: Revealed type is 'Any'
...
+
+class C:
+    @property
+    @staticmethod
+    def A() -> Type[A]:
+        return A
```
ilevkivskyi: Maybe also add a test with an overloaded method?

TH3CHARLie:
oops, I made some stupid mistakes

TH3CHARLie:
I add new logic in `semanal.py` to check if the class attribute we are accessing is an overloaded function if so, we check whether we are referencing it from a function with the same name, i.e. a variant of these overloads

test-data/unit/check-classes.test
```diff
@@ -6596,3 +6596,45 @@ reveal_type(D() + "str")  # N: Revealed type is 'Any'
...
+[case testRefMethodWithDecorator]
+from typing import Type
+
+class A:
+    pass
+
+class B:
+    @staticmethod
+    def A() -> Type[A]:
+        return A
```
ilevkivskyi: Please put `return` on the same line as the `def` (or replace with `...` altogether, this doesn't really test anything). I had it like this in my example for a reason, I wanted to be sure there is no off-by-one error.

mypy/semanal.py
```diff
@@ -3936,11 +3936,26 @@ class C:
...
+    def is_textually_before_statement(self, node: SymbolNode) -> bool:
+        """Check if a node is defined textually before the current statement
+
+        Note that decorated functions' line number are the same as
+        the top decorator.
+        """
+        assert self.statement
+        line_diff = self.statement.line - node.line
+        if isinstance(node, OverloadedFuncDef) and isinstance(self.statement, FuncDef):
+            return node.name != self.statement.name and line_diff > 0
```
ilevkivskyi: Comparing names looks like a hack as it can confuse two totally unrelated definitions. You can instead directly check if function definition is one of the overload items (the latter are stored as `.items` and `.impl`). This can be written as a separate helper method.

Also add a comment that this is a corner case where mypy technically deviates from runtime name resolution, but it is fine because we want an overloaded function to be treated as a single unit.

test-data/unit/check-classes.test
TH3CHARLie: Should I replace all the `return`s?

ilevkivskyi: Other returns are up to you, their positions are not really relevant

mypy/semanal.py
```diff
@@ -3936,11 +3936,34 @@ class C:
...
+        if isinstance(node, OverloadedFuncDef) and isinstance(self.statement, FuncDef):
+            return self.is_overloaded_item(node, self.statement) and line_diff > 0
+        elif isinstance(node, Decorator) and not node.is_overload:
+            return line_diff > len(node.original_decorators)
+        else:
+            return line_diff > 0
+
+    def is_overloaded_item(self, overload: OverloadedFuncDef, func: FuncDef) -> bool:
+        """Check whehter the function belongs to the overloaded variants"""
+        return func in overload.items
```
ilevkivskyi: This will return `False` in almost all cases, because in most cases items are `Decorator`s, not `FuncDef`. You also need to check `.impl`.

TH3CHARLie:
Good catch and thanks for the careful review! I check the logic again and indeed there are some mistakes, here's the update:

- update logic in `is_textually_before_statement` to ensure if we are referencing an overloaded item, the function return `False`
- add negative test
- For the logic in `is_overloaded_item`, I have to point that neither my previous solution nor the solution purposed works without modification. In this case, we need to check if a `FuncDef` is one of the overloaded variants and these variants are expressed in either `Decorator` or `FuncDef`. The former one is tricky as we need one more step to extract its decorated function for checking.

mypy/semanal.py
```diff
@@ -3936,11 +3936,38 @@ class C:
...
+        elif isinstance(node, Decorator) and not node.is_overload:
+            return line_diff > len(node.original_decorators)
+        else:
+            return line_diff > 0
+
+    def is_overloaded_item(self, node: SymbolNode, statement: Statement) -> bool:
+        """Check whehter the function belongs to the overloaded variants"""
+        if isinstance(node, OverloadedFuncDef) and isinstance(statement, FuncDef):
+            return (statement in
+                    [item.func if isinstance(item, Decorator) else item for item in node.items]
```
ilevkivskyi: Make this a set.

mypy/semanal.py
```diff
@@ -3936,11 +3936,38 @@ class C:
...
+                    or (node.impl is not None and statement is node.impl))
```
ilevkivskyi: `node.impl` also can be a `Decorator`.

TH3CHARLie:
To address that `node.impl` can also be a `Decorator`, the logic would be long enough to make it hard to understand, so I deliberately separate single return into two booleans to make the code easy to understand.

PR #8177, opened 2019-12-19
Allow 'in' operations with partial types

JukkaL:
This improves type inference in code like this:

```
d = {}
for k in f():
    if k not in d:
        d[k] = 0
    else:
        d[k] += 1
```

Unfortunately, this change breaks strict equality checking in 'in' operations
involving partial types. I'm not 100% sure this is a net improvement. I couldn't
find a clean way of getting strict equality working together with this.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2211,11 +2211,17 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         return result
 
     def find_partial_type_ref_fast_path(self, expr: Expression) -> Optional[Type]:
+        """If expression has a partial generic type, return it without additional checks.
+
+        In particular, this does not generate an error about a missing annotation.
+
+        Otherwise, return None.
+        """
         if not isinstance(expr, RefExpr):
             return None
         if isinstance(expr.node, Var):
             result = self.analyze_var_ref(expr.node, expr)
-            if isinstance(result, PartialType):
+            if isinstance(result, PartialType) and result.type is not None:
                 self.chk.store_type(expr, result)
                 return result
         return None
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2125,7 +2125,11 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             method_type = None  # type: Optional[mypy.types.Type]
 
             if operator == 'in' or operator == 'not in':
-                right_type = self.accept(right)  # always validate the right operand
+                # If the right operand has partial type, look it up without triggering
+                # a "Need type annotation ..." message, as it would be noise.
+                right_type = self.find_partial_type_ref_fast_path(right)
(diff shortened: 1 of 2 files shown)


mypy/checkexpr.py
```diff
@@ -2206,6 +2210,22 @@ def visit_comparison_expr(self, e: ComparisonExpr) -> Type:
...
+        In particular, this does not generate an error about a missing annotation.
+
+        Otherwise, return None.
+        """
+        if not isinstance(expr, RefExpr):
+            return None
+        if isinstance(expr.node, Var):
+            result = self.analyze_var_ref(expr.node, expr)
+            if isinstance(result, PartialType) and result.type is not None:
+                self.chk.store_type(expr, result)
```
ilevkivskyi: TBH I am comfortable with putting some extra partial types in type map. A while ago we had a bunch of crashes caused by this. Should we instead use `fixup_partial_type()` here?

mypy/checkexpr.py
```diff
@@ -2125,7 +2125,11 @@ def visit_comparison_expr(self, e: ComparisonExpr) -> Type:
             method_type = None  # type: Optional[mypy.types.Type]
 
             if operator == 'in' or operator == 'not in':
-                right_type = self.accept(right)  # always validate the right operand
+                # If the right operand has partial type, look it up without triggering
+                # a "Need type annotation ..." message, as it would be noise.
+                right_type = self.find_partial_type_ref_fast_path(right)
+                if right_type is None:
+                    right_type = self.accept(right)  # Validate the right operand
```
ilevkivskyi: What if we still defer (when possible) the current function, will it fix the `--strict-equality` problem?

mypy/checkexpr.py
JukkaL: Yeah, probably! I will try it.

mypy/checkexpr.py
JukkaL: I tried this but it didn't help. I think that the reason was that we don't infer more precise types after deferral.

test-data/unit/check-inference.test
```diff
@@ -1738,6 +1738,34 @@ def g() -> None:
...
+if 'x' in d:  # E: "None" has no attribute "__iter__" (not iterable)
+    pass
+reveal_type(d)  # N: Revealed type is 'None'
+[builtins fixtures/dict.pyi]
+
+[case testInferFromEmptyListWhenUsingInWithStrictEquality]
+# flags: --strict-equality
+def f() -> None:
+    a = []
+    if 1 in a:  # TODO: This should be an error
```
ilevkivskyi: From the point of view of bookkeeping, maybe this deserves a follow-up issue? Otherwise I am afraid we will forget about this.

JukkaL: Created #8182 to track this.

PR #8183, opened 2019-12-19
Support partial type inference with += and |=

JukkaL:
Code like this no longer requires a type annotation:

```
x = []
x += [1, 2]
```

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2042,7 +2042,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             self.check_assignment_to_multiple_lvalues(lvalue.items, rvalue, rvalue,
                                                       infer_lvalue_type)
         else:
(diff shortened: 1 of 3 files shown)


JukkaL:
I was originally not planning to support this, but decided to do this anyway since the implementation was easy. The list += case is quite frequent in the Dropbox internal S codebase.

mypy/checker.py
```diff
@@ -2142,11 +2142,21 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
                     rvalue_type = remove_instance_last_known_values(rvalue_type)
                 self.infer_variable_type(inferred, lvalue, rvalue_type, rvalue)
 
+    # (type, operator) tuples for augmented assignments supported with partial types
+    partial_type_augmented_ops = {
+        ('builtins.list', '+'),
+        ('builtins.set', '|'),
+    }
```
ilevkivskyi: This can be made final (maybe also move it to the top of file).

PR #8189, opened 2019-12-20
Tests: run self check and lint in parallel

JukkaL:
I've seen this speed up `runtests.py` by about 20s when multiple cores are
available.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python3
+import subprocess
+from subprocess import Popen
 from os import system
 from sys import argv, exit, platform, executable, version_info
-
-prog, *args = argv
 
 
 # Use the Python provided to execute the script, or fall back to a sane default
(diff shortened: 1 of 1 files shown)


runtests.py
```diff
@@ -63,23 +63,71 @@
...
+    if not set(args).issubset(cmds):
+        print("usage:", prog, " ".join('[%s]' % k for k in cmds))
+        exit(1)
+
+    if not args:
+        args = list(cmds)
+
+    status = 0
+
+    if args[:2] == ['self', 'lint']:
```
ilevkivskyi: Maybe instead check if both self and lint are somewhere in list? Otherwise it would be sad to accidentally switch the order and get a slower run.

runtests.py
```diff
@@ -63,23 +63,71 @@
...
+        if name in FAST_FAIL:
             exit(status)
+    return status
+
+
+def start_background_cmd(name: str) -> Popen:
+    cmd = cmds[name]
+    proc = subprocess.Popen(cmd,
+                            shell=True,
+                            stderr=subprocess.STDOUT,
```
ilevkivskyi: Could this cause mixing output of both processes? Maybe pipe both and then print when done.

JukkaL: I think that this is already what happens. Both are piped to the same pipe.

PR #8192, opened 2019-12-22
Allow strict in config, explicitly disallow inline

scop:


msullivan:
What's the rationale for explicitly disallowing in inline configuration?

scop:
Mainly to fix the current situation: specifying it inline doesn't work, and ends up displaying a misleading error message about configuration files (not inline config), currently:
```
$ cat /tmp/x.py 
# mypy: strict
$ python mypy /tmp/x.py 
/tmp/x.py:1: error: Strict mode is not supported in configuration files: specify individual flags instead (see 'mypy -h' for the list of flags enabled in strict mode)
```
After:
```
$ python mypy /tmp/x.py 
/tmp/x.py:1: error: Setting 'strict' not supported in inline configuration
```

If I remember correctly, making it work for inline config would not have been a quick fix. The details escape me at the moment.

Anyway, the rubber duck tells me that I should improve that message somewhat though; rebase and improvement coming in a jiffy.

PR #8193, opened 2019-12-22
Automatically write .gitignore to cache dir, ignoring everything

scop:
Kudos to pytest for the idea.

https://github.com/pytest-dev/pytest/issues/3286#issuecomment-393142058

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -601,8 +601,10 @@ class BuildManager:
                               and not has_reporters)
         self.fscache = fscache
         self.find_module_cache = FindModuleCache(self.search_paths, self.fscache, self.options)
+        # add ignore-all .gitignore to cache dir if we created it
+        cache_dir_existed = os.path.isdir(options.cache_dir)
         self.metastore = create_metastore(options)
-        if fscache.isdir(options.cache_dir):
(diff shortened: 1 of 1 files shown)


mypy/build.py
```diff
@@ -602,6 +602,12 @@ def __init__(self, data_dir: str,
         self.fscache = fscache
         self.find_module_cache = FindModuleCache(self.search_paths, self.fscache, self.options)
         self.metastore = create_metastore(options)
+        if fscache.isdir(options.cache_dir):
```
ethanhs: I would suggest adding the `.gitignore` file only on creating the cache directory, as I wouldn't be surprised if some people store it in git intentionally. With this they would have to delete the .gitignore file every run.

scop: Sure, why not, 90013eb should take care of it. Also using `os.path.isdir` instead of `fscache.isdir` as metastore creates dirs without updating fscache's stat cache.

scop:
Done. I get an unrelated flake8 failure in my Travis tests, but I guess that might go away once rebased. Would you like me to do that BTW, and maybe squash the commits while at it? Not sure about the preferred PR flow here.

msullivan:
I'll squash it when I merge the PR, which I will do now. Thank you!

PR #8198, opened 2019-12-23
Run 3.5.1 tests compiled with mypyc

JukkaL:
Python 3.5.1 tests were the slowest job, and this makes them faster
(roughly from 20min to 15min).


JukkaL:
cc @msullivan

PR #8199, opened 2019-12-23
Make runtests.py skip some slow mypyc tests by default

JukkaL:
The skipped mypyc run tests are very slow and I believe that they only
rarely fail for changes that don't touch mypyc.

The skipped tests can be run via `runtests.py mypyc-extra`.

Also explain running tests in some more detail in the the docs.


JukkaL:
cc @msullivan

Issue #8187, opened 2019-12-20
Cannot use Literal variable inside Literal type


intgr:
Using mypy 0.761, Python 3.7.5 here. The `ANIMALS2` value is highlighted as incorrect by Mypy:
```python
COW: Literal['cow'] = 'cow'
PIG: Literal['pig'] = 'pig'
ANIMALS1: Literal['cow', 'pig']  # This is OK
ANIMALS2: Literal[COW, PIG]
#         ^^^^^^^ Mypy: Parameter 1 of Literal[...] is invalid
```

But it seems Mypy should be able to deduce that ANIMALS1 and ANIMALS2 are the same thing?

I'm not sure if the more lax version should be allowed as well, but seems useful:

```python
OX = 'ox'
ANIMALS = Literal[OX]
```

TH3CHARLie:
This fails too,
```python
from typing import Union
from typing_extensions import Literal
Foo: Literal[1] = 1
Bar: Literal[2] = 2

Test: Union[Foo, Bar]   # E: Variable "__main__.Foo" is not valid as a type \
                                         # E: Variable "__main__.Bar" is not valid as a type
Test1: Union[Literal[1], Literal[2]]
```
cc @ilevkivskyi

TH3CHARLie:
Well, I find #3494 maybe related to this

ilevkivskyi:
> Well, I find #3494 maybe related to this

Essentially yes, variables and type aliases should be never mixed. If you want an alias, use `COW = Literal['cow']`.

TH3CHARLie:
@ilevkivskyi indeed and I misunderstand this, too. However, I found a `TODO`:
https://github.com/python/mypy/blob/fe8309aaed79a714c3e5267c25ad7812f909259e/mypy/typeanal.py#L425-L427
which I think it would be useful to add a `note` here to point to the docs to help the anyone encounters this error message in the future

ilevkivskyi:
Yes, but fixing this TODO depends on fixing #3494 first.

Issue #3500, opened 2017-06-06
Better types for modules


ilevkivskyi:
In situations with dynamically/conditionally imported modules:
```python
if random() > 0.5:
    import mod1
else:
    import mod2 as mod1
```
mypy can infer only ``types.ModuleType`` for the ``mod1`` node. Another (quite widespread) example is something like:
```python
try:
    import fastmath
except ImportError:
    import math as fastmath
```
In such case all symbol table information for ``math`` and ``coolmath`` is "lost". We could support (at least internally) some special type that will hold symbol table information and will be allowed to appear in unions (or support structural joins). For example, in the above case type of ``fastmath`` would be ``Union[Module['/path/to/fastmath'], Module['math']]``. Then mypy will be able to infer ``fastmath.sin`` as ``Callable[[float], float]``.

gvanrossum:
IIRC you get an error "Name 'mod1' already defined" on the second. The
solution is to put a `# type: ignore` there, then mypy will assume the type
of the first import. Honestly that seems reasonable to me.

But if you really want the Module[] notation, the argument should probably
the full module name, not the filename, i.e. Union[Module['fastmath'],
Module['math']].

ilevkivskyi:
> then mypy will assume the type of the first import

Actually mypy assumes the type of the (textually) last encountered import

> Honestly that seems reasonable to me.

Except that there is a dozen issues on the tracker about ``try ... except ImportError`` and similar.

> But if you really want the Module[] notation

I am more interested in having a better internal handling of modules first. Having something conceptually equivalent to "a union of symbol tables" would solve many of the existing issues. Then, depending on experience we could consider adding syntactic support for such types.

ilevkivskyi:
Related #1153 and #1297 (plus another issues mentioned in the second).

gvanrossum:
OK, you're right.

PR #8200, opened 2019-12-24
Update docs about variables and aliases

### LINKED ISSUE #3494: Clarify aliases vs variables
We have both type aliases and variables with types like ``Type[...]``, also there are module aliases and variables of type ``ModuleType``. I think we should document the difference in semantics and provide examples. (Also I think there are few minor things that should be fixed in mypy.) Here how I see things:

1. Variables with type ``Type[...]`` should be created by assignments with an explicit type annotations:
   ```python
   class A: ...
   tp: Type[A] = A
   ```
2. Aliases are created by assignments without an explicit type:
   ```python
   class A: ...
   Alias = A
   ```
3. The difference is that aliases are completely known statically and can be used in type context (annotations):
   ```python
   class A: ...
   class B: ...

   if random() > 0.5:
       Alias = A
   else:
       Alias = B  # I think this should be prohibited (there are two TODOs in semanal.py)

   tp: Type[object]
   if random() > 0.5:
       tp = A
   else:
       tp = B  # This is OK

   def fun1(x: Alias) -> None: ... # This is OK
   def fun2(x: tp) -> None: ... # Error
   ```
4. There is similar distinction between module aliases and variables with type ``types.ModuleType``:
   ```python
   import math, cmath
   from types import ModuleType

   mod: ModuleType = math  # variable
   m = math # Alias

   mod.sin # Error. ModuleType has no attribute 'sin'
   m.sin # OK
   ```
   and also
   ```python
   if random() > 0.5:
       m = math
   else:
       m = cmath  # Should be prohibited

   mod: ModuleType
   if random() > 0.5:
       mod = math
   else:
       mod = cmath # OK, note explicit annotation above
   ```
5. For modules, we could add better support later (a particular module is a literal type as in https://github.com/python/mypy/issues/3062 and binder can bind ``mod`` to ``Union[Literal[math], Literal[cmath]]`` after the above code).

@JukkaL Do we agree on semantics here? If yes, then I will make some minor changes in mypy and add this to documentation (maybe we also need to clarify this in PEP 484).

TH3CHARLie:
resolves #3494(Since module types are tracked in #3500)

Following https://github.com/python/mypy/issues/8187#issuecomment-567957575 and #3494, if I understand correctly and the semantics in #3494's example code has been fixed(report error on re-assign `Alias = B`, the remaining work of #3494 is to update the docs, which is the main focus of this PR.

Newly added docs are in `common issues and solutions` section, with the content mostly adapted from Ivan's example in #3494. And a note point to the docs is also added.

When reviewing the PR, I think there are two things to tweak:
- The section of the docs: `common issues` or `alias type` ?
- The wording: is current example and explantion enough for users to understand the difference


TH3CHARLie:
Many tests are broken due to the newly added note, I'd like to first confirm the wording and availability of this PR before pervasively modifying the tests

TH3CHARLie:
I almost forget about this PR.....Now I modify the test suite to apply the new change and pass the tests. But I am worried about that this PR may be too pervasive

docs/source/common_issues.rst
```diff
@@ -710,3 +710,44 @@ You can install the latest development version of mypy from source. Clone the
     git clone --recurse-submodules https://github.com/python/mypy.git
     cd mypy
     sudo python3 -m pip install --upgrade .
+
+Differentiate variables and aliases
```
JukkaL: Brainstorming: "aliases" is an overloaded term. This might be a slightly clearer title: "Variables vs type aliases" (also similar to the existing section "Invariance vs covariance").

gvanrossum:
I was just pointed to the docs introduced by this PR ([Variables vs type aliases](https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases)) and for a few moments I was very confused.

Perhaps this was because I was trying to create a type alias for `Annotated`, which apparently doesn't work (though I think it ought to).

But the docs (copied verbatim from Ivan's text in #3494) did not help. I think what the docs are trying to point out is that both a variable with type `Type[A]` and an unannotated "variable" initialized with `A` hold the class object `A` at runtime, but the type checker sees them differently: the variable `tp` is a variable holding a class object (that must subclass the given class), and (being a variable) cannot be used in annotations; while the alias is a type definition to the type checker, usable in annotations, and (like other type definitions) is not allowed to be changed.

I'm still not sure how to best write this down, but I feel that the docs as they currently exist could still use some improvement.

TH3CHARLie:
@gvanrossum Thanks for the feedback. The docs here definitely needs refinement since #9238 also points out its incompleteness. I'll file a PR to update that and will request for your review there.

PR #8201, opened 2019-12-24
Add docs for --local-partial-types option

### LINKED ISSUE #8046: dmypy and mypy behave differently with same settings
Consider this example:

```
class Config:
    url = None

    def __init__(self) -> None:
        self.url = 'xxx'
```

`mypy` says there are no errors, but `dmypy` with the same settings finds one:

```
$ mypy example.py

$ dmypy run -- --follow-imports=skip example.py
Daemon started
example.py:3: error: Need type annotation for 'url'

```

my package versions:

```

$ python --version
Python 3.6.5 :: Anaconda, Inc.
$ mypy --version
mypy 0.750
```

TH3CHARLie:
resolves #8046, document the `--local-partial-types` command line option

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -484,21 +484,27 @@ of the above sections.
 
 .. option:: --local-partial-types
 
-    By default, mypy won't check partial types spanning module top level or class top level.
-    This flag changes the behavior to only allow partial types at local level. The most common
-    cases are variables initialized with ``None`` but without explicit ``Optional`` annotation,
-    therefore their types can not be fully inferred and specified. For example:
+    In mypy, the most common cases for partial types are variables initialized using ``None``,
+    but without explicit ``Optional`` annotations. By default, mypy won't check partial types
+    spanning module top level or class top level. This flag changes the behavior to only allow
+    partial types at local level, therefore it disallows inferring variable type for ``None``
+    from two assignments in different scopes. For example:
 
     .. code-block:: python
 
         from typing import Optional
 
-        a = None  # error
+        a = None  # Need type annotation here if using --local-partial-types
         b = None  # type: Optional[int]
 
         class Foo:
-            bar = None  # error
-            boo = None  # type: Optional[int]
+            bar = None  # Need type annotation here if using --local-partial-types
+            baz = None  # type: Optional[int]
+
+            def __init__(self) -> None
+                self.bar = 1
+
+        reveal_type(Foo().bar)  # Union[int, None] without --local-partial-types
 
     Note: this option is always implicitly enabled in mypy daemon and will become
     enabled by default for mypy in a future release.
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -482,6 +482,23 @@ of the above sections.
            # 'items' now has type List[List[str]]
            ...
 
+.. option:: --local-partial-types
+
+    By default, mypy won't check partial types spanning module top level or class top level.
+    This flag changes the behavior to only allow partial types at local level. For example:
+
+    .. code-block:: python
+
+        from typing import Optional
+
+        a = None  # error
+        b = None  # type: Optional[str]
+
+        class Foo:
+            bar = None  # error
+            boo = None  # type: Optional[int]
+
+
 .. option:: --no-implicit-reexport
 
     By default, imported values to a module are treated as exported and mypy allows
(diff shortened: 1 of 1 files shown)


docs/source/command_line.rst
```diff
@@ -482,6 +482,23 @@ of the above sections.
            # 'items' now has type List[List[str]]
            ...
 
+.. option:: --local-partial-types
+
+    By default, mypy won't check partial types spanning module top level or class top level.
+    This flag changes the behavior to only allow partial types at local level. For example:
```
JukkaL: We don't mention partial types anywhere else in the documentation, so they are essentially a low-level implementation concept. It would be better if the explanation would explain what the option does without referring to the partial types concept.

Also mention that this is always implicitly enabled in the mypy daemon and will become enabled by default in a future mypy release.

TH3CHARLie: Good catch, what do you suggest as an alternative for partial types that is more user-friendly to non-devs?

TH3CHARLie: My current thoughts are to keep the name 'partial types' unchanged in the docs, in correspondence with the option name, and provide some description for partial types, maybe like
`Types such as List[?] where type arguments are unknown or None` so the readers can connect the name with some concrete examples

docs/source/command_line.rst
```diff
@@ -482,6 +482,32 @@ of the above sections.
            # 'items' now has type List[List[str]]
            ...
 
+.. option:: --local-partial-types
+
+    By default, mypy won't check partial types spanning module top level or class top level.
+    This flag changes the behavior to only allow partial types at local level. The most common
+    cases are variables initialized with values such as ``[]``, ``{}`` where the explicit type
+    arguments ``T`` for ``List[T]``, ``Dict[T]`` is missing, or ``None`` without explicit
+    ``Option`` annotation, therefore their types can not be fully inferred and specified. For example:
```
JukkaL: `Option` -> `Optional`

docs/source/command_line.rst
```diff
@@ -482,6 +482,32 @@ of the above sections.
            # 'items' now has type List[List[str]]
            ...
 
+.. option:: --local-partial-types
+
+    By default, mypy won't check partial types spanning module top level or class top level.
```
JukkaL: This sentence is pretty vague. It would be better to also introduce the concept "partial types" before discussing it.

This is an important case where this option makes a difference:

```py
class A:
    x = None  # Need type annotation here if using --local-partial-types

    def __init__(self) -> None:
        self.x = 1

reveal_type(A().x)  # Union[int, None] without --local-partial-types
```

This example also makes it easier to see why this is called "local partial types" -- it disallows inferring variable type from two assignments in different scopes (for example, in class body and inside a method).

docs/source/command_line.rst
```diff
@@ -482,6 +482,32 @@ of the above sections.
...
+    This flag changes the behavior to only allow partial types at local level. The most common
+    cases are variables initialized with values such as ``[]``, ``{}`` where the explicit type
```
JukkaL: This option only affects partial `None` types, and discussing `[]`, `{}`, etc. could be confusing here, since they aren't affected by the option.

docs/source/command_line.rst
```diff
@@ -482,6 +482,32 @@ of the above sections.
...
+    ``Option`` annotation, therefore their types can not be fully inferred and specified. For example:
+
+    .. code-block:: python
+
+        from typing import Optional, List, Dict
+
+        a = None  # error
+        b = None  # type: Optional[int]
+        l1 = []   # error
+        l2 = []   # type: List[int]
```
JukkaL: The list and dict examples are not relevant, since this option has no effect on them.

docs/source/command_line.rst
```diff
@@ -482,6 +482,32 @@ of the above sections.
...
+
+        class Foo:
+            bar = None  # error
+            boo = None  # type: Optional[int]
+            d1 = {}  # error
+            d2 = {}  # type: Dict[int, float]
```
JukkaL: Here is the dict example which can be left out.

docs/source/command_line.rst
```diff
@@ -482,6 +482,32 @@ of the above sections.
...
+
+        from typing import Optional, List, Dict
+
+        a = None  # error
+        b = None  # type: Optional[int]
+        l1 = []   # error
+        l2 = []   # type: List[int]
+
+        class Foo:
+            bar = None  # error
```
JukkaL: Similar to above -- `--local-partial-types` causes this to be an error.

TH3CHARLie:
Thanks for the review!

PR #8205, opened 2019-12-24
Add symbols for member variables to the symbol table when a subclass of TypedDict with the same name exists

### LINKED ISSUE #8097: Forward reference to namespaced type with same name breaks with TypedDict

I am reporting a bug

```
from typing import Union, Any
from typing_extensions import TypedDict


class T(TypedDict):
    b: b.T


class b:
    T = Union[Any]
```

```
failure2.py:7: error: Name 'b.T' is not defined
```

No type error

Python 3.7.3, mypy 0.750. I get the same results after installing mypy from Git master.

No particular flag.

  the full traceback below.

No crash

Additional notes:
The following examples *do* typecheck properly:
* Name the type differently:
```
from typing import Union, Any
from typing_extensions import TypedDict


class T(TypedDict):
    b: b.U


class b:
    U = Union[Any]
```
* Without TypedDict:
```
from typing import Union, Any


class T:
    b: b.T


class b:
    T = Union[Any]
```
* Without forward reference:
```
from typing import Union, Any
from typing_extensions import TypedDict


class b:
    T = Union[Any]


class T(TypedDict):
    b: b.T
```

PattenR:
PR title a little long - suggestions welcome.

Fixes #8097


test-data/unit/check-newsemanal.test
```diff
@@ -3175,3 +3175,13 @@ class User:
...
+
+[case testNewAnalyzerMemberNameMatchesTypeDict]
+from typing import Union, Any
+from typing_extensions import TypedDict
+
+class T(TypedDict):
+    b: b.T
+
+class b:
+    T: Union[Any]
```
TH3CHARLie: Could you please also check the `NamedTuple` case as I mentioned in the issue's comment to prevent regression on both cases

PattenR: Added a test for this

TH3CHARLie:
Travis breaks with `mypyc runtime test with Python3.6 on OSX` with logs:
```
........................................................................ [ 30%]
..........................................
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
```
Could you please submit a new empty commit to retrigger Travis? @PattenR even though it may not be caused by your commits.

TH3CHARLie:
ping @JukkaL for review

mypy/semanal.py
```diff
@@ -4295,7 +4295,8 @@ def add_symbol_table_node(self,
                 if not (isinstance(new, (FuncDef, Decorator))
                         and self.set_original_def(old, new)):
                     self.name_already_defined(name, context, existing)
-        elif name not in self.missing_names and '*' not in self.missing_names:
+        elif (not (name in self.missing_names and symbol.kind != MDEF)
```
JukkaL: Hmm I'm not sure if this change is safe. Why are you skipping the check for `MDEF`?

PattenR: In https://github.com/python/mypy/pull/6257 you ensure that symbol table additions are skipped if they are defining something that was already defined, but the previous definition couldn't be resolved. 

You mentioned that you don't track the scopes for simplicity, and that extra deferrals should occur in order to correct this. I'm assuming in this case that these deferrals aren't happening as we'd expect.

My reasoning here was that because in the case of Typeddicts (and NamedTuples too?) can only have members that collisions wouldn't be able to take place in the scope of the class. It seems like we either want to treat this as a special case and only have this condition for these types of class or investigate further the reason for these deferrals not happening (any ideas you have for that one would be great).

This is a tricky one! Any thoughts?

JukkaL: It's possible to have missing names as MDEFs, and they should be skipped, so I'm now convinced that this is not the right fix. A more promising fix would be to separate `missing_names` per scope, since sharing it across scopes seems to be the underlying issue -- an incomplete definition in the module namespace blocks the addition of a thing to a class namespace. We have a TypedDict `T` and a nested `b.T`, and the common short name `T` is the key here. 

We want to be able to define `b.T` while the module-level `T` is incomplete, and we could do this if we'd recognize that `T` is not actually missing in the class scope/namespace (but it might be, so we can't just remove the check). This was a real head-scratcher!

Does this make sense to you? Would you like to try splitting `missing_names` per scope?

PattenR: Yeah that would make a lot more sense, I've had a go at implementing it but it wasn't super obvious. I wasn't totally sure on the best way to get the current scope to split on so any suggestions you have there would be interesting. 

I also looked into adding further tests to catch cases where this succeeds where the previous one failed. I thought a nested TypedDicts class could be a good way forward, but found that nested TypedDicts weren't being marked as incomplete. Any ideas you have for further testing would be great.

PattenR: Ping @JukkaL, any thoughts?

mypy/semanal.py
```diff
@@ -4170,6 +4171,12 @@ def lookup_current_scope(self, name: str) -> Optional[SymbolTableNode]:
         else:
             return self.globals.get(name)
 
+    def current_scope_identifier(self) -> Optional[str]:
```
JukkaL: This is a little inconsistent with how we track scopes elsewhere. It would be better to mirror the logic used for the `locals` attribute (also `block_depth` a few others). We'd maintain a stack and append a fresh set on each `enter()` or `enter_class()` call, and pop an item on each `leave()` or `leave_class()` call. We can then use `missing_names[-1]` to access the innermost scope.

mypy/semanal.py
JukkaL: I left a comment below. Your new approach works, but I have suggestion about how to make it more consistent with how other similar things are handled. And sorry for the slow response!

mypy/semanal.py
PattenR: That makes a lot more sense, cheers for the hint!

mypy/semanal.py
PattenR: No worries, thanks for the review!

TH3CHARLie:
@JukkaL ping

PR #8212, opened 2019-12-28
Typeddict context returned correctly when unambiguous

### LINKED ISSUE #8156: Dictionaries not accepted for implicit cast to TypedDict Union

* Are you reporting a bug, or opening a feature request? **Bug**

```python
from typing import TypedDict, Literal, Union, Mapping, Any, cast

A = TypedDict('A', {'@type': Literal['a-type'], 'a': str})
B = TypedDict('B', {'@type': Literal['b-type'], 'b': int})

bad_cast: Union[A, B] = {'@type': 'a-type', 'a': 'Test'}
good_cast_1: Union[A, B] = cast(Union[A, B], {'@type': 'a-type', 'a': 'Test'})
good_cast_2: Union[A, B] = cast(A, {'@type': 'a-type', 'a': 'Test'})
```

```test_mypy_discriminate.py:6:25: error: Incompatible types in assignment (expression has type "Dict[str, str]", variable has type "Union[A, B]")```

Since the given type matches one of the types in the Union, it should be accepted.

* What are the versions of mypy and Python you are using? **Python 3.8, mypy 0.750**
* Do you see the same issue after installing mypy from Git master? **Don't know how to do that**

### PR DESCRIPTION
**Pull request by PattenR**
Fixes #8156 

This is a suggestion as to how we could handle getting the Typeddict context from a Union with more than one Typeddict. 

The idea is that we check each of the Typeddict types to find one who's keys match that of the expression, if more than one match it's then ambiguous. The slightly weird thing is that this matching is only applied in the Union case, if there is only a single Typeddict it's returned as before. 

I'm not sure that I've put the tests in the most appropriate place, and suggestions would be great.

### CODE REVIEW

File: mypy/checkexpr.py [hunk 1]
```diff
@@ -3294,19 +3313,25 @@ def visit_dict_expr(self, e: DictExpr) -> Type:
         assert rv is not None
         return rv
 
-    def find_typeddict_context(self, context: Optional[Type]) -> Optional[TypedDictType]:
+    def find_typeddict_context(self, context: Optional[Type],
+                              dict_expr: DictExpr) -> Optional[TypedDictType]:
```
msullivan: (Has the linter broken??). Indent `dict_expr` another space

mypy/messages.py
```diff
@@ -1111,6 +1111,14 @@ def typeddict_key_not_found(
         else:
             self.fail("TypedDict {} has no key '{}'".format(format_type(typ), item_name), context)
 
+    def typeddict_context_ambiguous(
+            self,
+            types: List[TypedDictType],
+            context: Context) -> None:
+        self.fail('Type of TypedDict is Ambiguous, could be any of {}'.format(
+                  ([(format_type(typ), format_item_name_list(typ.items.keys())) for typ in types])
```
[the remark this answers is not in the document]
PattenR: Yeah I wasn't quite sure what to do about printing the types here. Thanks a lot for that PR, I've updated to use ```format_type_distinctly``` and it's much cleaner! Also if you can think of any slightly nicer wording do let me know.

mypy/checkexpr.py
PattenR: Yeah this was weird, my vscode showed this being inline! I've added the space in but yeah weird this wasn't picked up.

PattenR:
Thanks for the review @msullivan! Any more comments would be great

test-data/unit/check-typeddict.test
```diff
@@ -822,6 +822,27 @@ reveal_type(u(c, m_i_i)) # N: Revealed type is 'Union[typing.Mapping*[builtins.i
...
+[builtins fixtures/tuple.pyi]
+
+[case testTypedDictUnionAmbiguousCase]
+from typing import Union, Mapping, Any, cast
+from typing_extensions import TypedDict, Literal
+
+A = TypedDict('A', {'@type': Literal['a-type'], 'a': str})
+B = TypedDict('B', {'@type': Literal['a-type'], 'a': str})
+
+c: Union[A, B] = {'@type': 'a-type', 'a': 'Test'} # E: Type of TypedDict is ambiguous, could be any of ('"A"', '"B"') \
```
JukkaL: Nit: Both single and double quotes in the error message (`'"A"'` etc.) look strange. It would be better to only use double quotes.

PattenR: Thanks for the review! I've updated to have double quotes although I could also have used format_type_distinctly directly with the ```bare``` argument set to true, would have had the same result except for single quotes (maybe that would have been cleaner?).

Issue #8244, opened 2020-01-06
add_method() should be usable in the get_dynamic_class_hook() too


mkurnikov:
No reason for it to accept `ClassDefContext` as a first parameter.

PR #8245, opened 2020-01-06
add add_method_to_class() function for plugins.common

mkurnikov:
* move `mypy.plugins.common.add_method()` to `SemanticAnalyzerPluginInterface` and to `SemanticAnalyzer`
* change old implementation to call new one, to preserve backwards compatibility
* replace first `ctx` parameter with `cls: ClassDef`

Fixes https://github.com/python/mypy/issues/8244

### DIFF
--- a/mypy/plugins/common.py
+++ b/mypy/plugins/common.py
@@ -1,7 +1,7 @@
 from typing import List, Optional, Union
 
 from mypy.nodes import (
-    ARG_POS, MDEF, Argument, Block, CallExpr, Expression, SYMBOL_FUNCBASE_TYPES,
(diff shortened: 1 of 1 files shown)


mkurnikov:
The only reason was to make the fact that it could only be used for `SemanticAnalyzer` phase clearer. Does it need to be another function, or could I change a signature?

msullivan:
I'd prefer to add a new function, I think, to avoid breaking compatibility unnecessarily. It is kind of unfortunate, though, since it was a mistake really.

mkurnikov:
I've updated, not sure about naming, let me know what you think

msullivan:
Thanks!

Issue #4975, opened 2018-04-26
Type inference of Tuples returns object instead


eddieschoute:
While interacting with `zip` I encountered a variant of the following error:
```
error: No overload variant of "zip" matches argument types [builtins.list[builtins.int*], builtins.object*]
```
which is emitted by mypy on the case of
```py
out = list(
    list(zip([1, 2], ordered_item))
    for item in [(1, 2), (4, 5)]
    for ordered_item in [item, tuple(reversed(item))]
    )
print(out)
```
This code runs normally on python 3.6.5, outputting:
```
[[(1, 1), (2, 2)], [(1, 2), (2, 1)], [(1, 4), (2, 5)], [(1, 5), (2, 4)]]
```

## Expected Behavior
Mypy should infer that `ordered_item` is a tuple, or at least an iterable so that `zip` can be applied on it.

It may also be interesting to consider the case where `ordered_item` is heterogenous over different `Iterables` (i.e. not only tuples). Would Iterable be inferred then?

## Actual Behavior
Mypy infers the `object` type on `ordered_item` which you understandably cannot zip.

By explicitly casting `item` to tuple the error message is resolved:
```py
out = list(
    list(zip([1, 2], ordered_item))
    for item in [(1, 2), (4, 5)]
    for ordered_item in [tuple(item), tuple(reversed(item))]
    )
print(out)
```

## Environment
OS: macOS 10.13
Python: 3.6.5
mypy: 0.590

beamerblvd:
Is this fix available in a release?

ethanhs:
@beamerblvd I'm not sure I understand. I don't think there is a fix for this yet

beamerblvd:
@ethanhs, my mistake. I saw a red `Closed` and thought it was for this issue, but it was for another issue having reference this.

ilevkivskyi:
This already appeared five times, so I am raising priority to high (especially since IIUC we are not going to infer unions for ternary expressions). This should be not hard to fix if someone wants to try.

intgr:
Another minimal reproducer:
```python
tup1: tuple = ()
tup2 = tup1 if input() else ()
reveal_type(tup2)   # N: Revealed type is 'builtins.object'
```
What is even happening here? 😄

JelleZijlstra:
[...]

PR #8253, opened 2020-01-07
[WIP] Fix issue #4975

theodoretliu:
Addresses issue #4975. See three test cases, first is passing, unclear on what the behavior should be on second two

### DIFF
--- a/mypy/join.py
+++ b/mypy/join.py
@@ -272,7 +272,7 @@ class TypeJoinVisitor(TypeVisitor[ProperType]):
             return TupleType(items, fallback)
         elif isinstance(self.s, Instance) and self.s.type.fullname == 'builtins.tuple':
             typ = self.s.args[0]
-            for item in t.items[1:]:
+            for item in t.items:
                 typ = self.join(item, typ)
 
             return Instance(self.s.type, [typ])
--- a/mypy/join.py
+++ b/mypy/join.py
@@ -179,6 +179,8 @@ class TypeJoinVisitor(TypeVisitor[ProperType]):
             return join_types(t, self.s)
         elif isinstance(self.s, LiteralType):
             return join_types(t, self.s)
+        elif isinstance(self.s, TupleType):
+            return join_types(t, self.s)
         else:
             return self.default(self.s)
 
@@ -268,6 +270,12 @@ class TypeJoinVisitor(TypeVisitor[ProperType]):
                                       mypy.typeops.tuple_fallback(t))
             assert isinstance(fallback, Instance)
             return TupleType(items, fallback)
+        elif isinstance(self.s, Instance) and self.s.type.fullname == 'builtins.tuple':
+            typ = self.s.args[0]
+            for item in t.items[1:]:
+                typ = self.join(item, typ)
+
+            return Instance(self.s.type, [typ])
         else:
             return self.default(self.s)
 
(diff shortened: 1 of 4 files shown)


theodoretliu:
@ilevkivskyi @Michael0x2a appears to be working if I hardcode a file `test.py` and run `mypy` on it
```python
from typing import Tuple

x: Tuple[str, float]
y: Tuple[int, ...]

lst = [x, y]
reveal_type(lst[0])
```
produces
```
test.py:7: note: Revealed type is 'builtins.tuple*[builtins.object]'
```
with no errors.

However, running the same in the test suite seems to cause issues.

Similar with the following
```python
from typing import Tuple

x: Tuple[int, str]
y: Tuple[int, ...]

lst = [x, y]
reveal_type(lst[0])
```
produces
```
test.py:7: note: Revealed type is 'builtins.tuple*[builtins.object]'
```

mypy/join.py
```diff
@@ -268,6 +270,12 @@ def visit_tuple_type(self, t: TupleType) -> ProperType:
                                       mypy.typeops.tuple_fallback(t))
             assert isinstance(fallback, Instance)
             return TupleType(items, fallback)
+        elif isinstance(self.s, Instance) and self.s.type.fullname == 'builtins.tuple':
```
ilevkivskyi: Couple more things:
* What if the instance is a _subclass_ of `tuple`?
* What if we have a named tuple on either side?

theodoretliu: I pushed some updates to try and handle the subclass case. I'm afraid I don't really understand the subclassing system. I know to look in .type.mro to see the hierarchy of types, but was unclear how to get the original "args" of the `builtins.tuple` type. For example
```python
class A(Tuple[int, ...]):
    pass

class B(A):
    pass

x: B
y: Tuple[int]
```
How do we access the `builtins.int` part when joining?

theodoretliu: Once this is resolved, I can work on the NamedTuple case as well

ilevkivskyi:
Sorry, I will not have time for this, hopefully @JukkaL or @msullivan can help.

JukkaL:
I can finish the review. The next steps would be fixing the merge conflict and the failing test cases. If you need help with some of these, just let me know.

intgr:
Doh, sorry :( I was under the impression that nobody was working on this issue (made that assumption when JelleZijlstra commented in #4975). It didn't help that this PR is titled "Initial attempt"

So meanwhile I submitted two pull requests for the two behaviors implemented here: #8333 and #8335.

The good news is that I happened to find a cleaner implementation, my approach can join `Tuple[x, x] + List[x]` into a `Sequence[x]` type, which seems to not be the case here.

I now tested my code with the test cases in this PR and my code passes these cases.

What do you think is the best way forward?

theodoretliu:
@intgr well, you were quick on the uptake, so I don't have a problem if they merge yours instead of mine. In the future, I can title my PR better, but I did reference the issue in this PR and it does show up on the Issue tracker... Nevertheless, taking a quick look at your PR shows that we had approximately the same idea, it's up the maintainers @JukkaL @ilevkivskyi to make the final call.

There was an issue with my PR that was raised regarding types that inherit from Tuple: if we go with yours, we might try to fix that too?

intgr:
Thanks, for the heads-up, I hadn't considered Tuple subclasses. I'll have to test that behavior.

JukkaL:
Hmm, I think I prefer #8333 and #8335, since they are more recent and have no conflicts. Supporting tuple subclasses would be a nice addition.

theodoretliu:
Closing in favor of @intgr work in #8333 and #8335

PR #8256, opened 2020-01-07
Speed up tests by simplifying default builtins

### LINKED ISSUE #6428: Speed-up tests by reorganizing fixtures to not import typing when not needed
#6427 added typing import to `lib-stub/builtins.pyi` and `lib-stub/__builtins__.pyi`. This causes some slow down for tests (up to 9% for fast tests). We need to figure out how to speed-up tests without losing correctness (like `tuple` should be generic).

JukkaL:
This removes `tuple` from the default builtins test stubs, which seems to 
speed up fast tests by about 7%.

Also improved error messages that are generated when a test case is
missing the right fixture.

I didn't update Python 2 test stubs since they aren't used in many tests.

Fixes #6428 (except for Python 2 which doesn't seem important).

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -53,6 +53,23 @@ ARG_CONSTRUCTOR_NAMES = {
 }  # type: Final
 
 
+# Map from the full name of a missing definition to the test fixture (under
+# test-data/unit/fixtures/) that provides the definition. This is used for
+# generating better error messages when running mypy tests only.
+SUGGESTED_TEST_FIXTURES = {
+    'builtins.list': 'list.pyi',
+    'builtins.dict': 'dict.pyi',
+    'builtins.set': 'set.pyi',
+    'builtins.tuple': 'tuple.pyi',
+    'builtins.bool': 'bool.pyi',
+    'builtins.Exception': 'exception.pyi',
+    'builtins.BaseException': 'exception.pyi',
+    'builtins.isinstance': 'isinstancelist.pyi',
+    'builtins.property': 'property.pyi',
+    'builtins.classmethod': 'classmethod.pyi',
+}  # type: Final
+
+
 class MessageBuilder:
     """Helper class for reporting type checker error messages with parameters.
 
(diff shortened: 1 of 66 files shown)


test-data/unit/fine-grained.test
```diff
@@ -1717,6 +1717,8 @@ N = NamedTuple('N', [('x', int)])
 [file a.py]
 def f() -> None: pass
 [file a.py.2]
+[builtins fixtures/tuple.pyi]
+[builtins fixtures/tuple.pyi]
```
ilevkivskyi: Why this appears twice here and in a couple dozen other tests below?

test-data/unit/semanal-types.test
```diff
@@ -147,6 +147,8 @@ class A: pass
 class B: pass
 a, b = None, None # type: (A, B)
 x = a, b
+[builtins fixtures/tuple.pyi]
+[builtins fixtures/tuple.pyi]
```
ilevkivskyi: Same question in this file.

test-data/unit/fine-grained-blockers.test
```diff
@@ -405,6 +405,8 @@ import blocker
 [file a.py.4]
 import sys
 1()
+[builtins fixtures/tuple.pyi]
+[builtins fixtures/tuple.pyi]
```
ilevkivskyi: It looks like many files have this problem actually.

test-data/unit/check-incomplete-fixture.test
```diff
@@ -58,17 +58,24 @@ main:1: error: Name 'isinstance' is not defined
...
+x: Tuple[int, str]
 [out]
--- These errors are pretty bad, but keeping this test anyway to
--- avoid things getting worse.
-main:3: error: Value of type "Tuple[int, ...]" is not indexable
-main:4: error: "Tuple[int, ...]" has no attribute "__iter__" (not iterable)
+main:1: error: Name 'tuple' is not defined
+main:1: note: Maybe your test fixture does not define "builtins.tuple"?
+main:1: note: Consider adding [builtins fixtures/tuple.pyi] to your test description
+main:1: note: Did you forget to import it from "typing"? (Suggestion: "from typing import Tuple")
```
ilevkivskyi: We can probably skip this note when the above one is added (not very important however).

mypy/messages.py
```diff
@@ -53,6 +53,23 @@
...
+    'builtins.list': 'list.pyi',
+    'builtins.dict': 'dict.pyi',
+    'builtins.set': 'set.pyi',
+    'builtins.tuple': 'tuple.pyi',
+    'builtins.bool': 'bool.pyi',
+    'builtins.Exception': 'exception.pyi',
+    'builtins.BaseException': 'exception.pyi',
+    'builtins.isinstance': 'isinstancelist.pyi',
+    'builtins.property': 'property.pyi',
+    'builtins.classmethod': 'classmethod.pyi',
```
ilevkivskyi: Potentially we can have a unit test for this (check that every file defines a corresponding class).

test-data/unit/fine-grained.test
JukkaL: I probably ran the script that added these twice for this file, or I did some these manually and the script replicated them.

Issue #8236, opened 2020-01-03
Type guard in list comprehension with assignment expression does not constrain type


fluggo:
* Are you reporting a bug, or opening a feature request? **Bug**

```python
from typing import List, Iterator
import os
import re
    
def get_schema_versions(self) -> List[int]:
    return [int(match.group(0))
        for filename in os.listdir('schema')
        if (match := re.fullmatch(r'(\d+)\.sql', filename)) is not None]

def get_schema_versions2(self) -> Iterator[int]:
    for filename in os.listdir('schema'):
        match = re.fullmatch(r'(\d+)\.sql', filename)

        if match is not None:
            yield int(match.group(0))
```

* What is the actual behavior/output?

In the first function, mypy interprets the `match` variable as `Optional[Match[str]]` and flags the use of the possibly-`None` object. In the second function, `match` under the `if` statement is considered `Match[str]`.

* What is the behavior/output you expect?

Since both statements guard the `match` variable against being `None`, both uses should be of type `Match[str]`.

* What are the versions of mypy and Python you are using? **mypy 0.761, Python 3.8**
  Do you see the same issue after installing mypy from Git master? **Don't know how**
* What are the mypy flags you are using? (For example --strict-optional) **Nothing relevant**

fluggo:
This _might_ be the same bug as #7316.

msullivan:
Closing in favor its duplicate #8447 which is about to be fixed anyway

hauntsaninja:
@fluggo I was wondering if the code you're working on is open source? My interest is in adding projects to https://github.com/hauntsaninja/mypy_primer that use newer Python features.

fluggo:
@hauntsaninja Sorry, this project wasn't. I appreciate the ask, though.

PR #8258, opened 2020-01-08
Narrow types for walrus assignment in if statements

### LINKED ISSUE #7313: PEP 572: Assignments not narrowed in right-hand side of and
The binder interaction doesn't work properly for assignments in the right-hand side of an `and` (after `if x and (y := 1):`, the type of `y` isn't narrowed to `int`).

Test case:
```

[case testAndRHS]
from typing import Optional

def f(x: Optional[int], y: Optional[int]) -> None:
    if x and (y := 1):
        reveal_type(y)  # N: Revealed type is 'builtins.int'
```
(Currently it produces `Revealed type is 'Union[builtins.int, None]'` instead.)

Followup from #6899

### LINKED ISSUE #7316: PEP 572: Type not narrowed correctly in assignment
Example from @gvanrossum in #6899:
```
from typing import Optional

class C:
    def foo(self):
        pass

def good(b: Optional[C]) -> None:
    a = b
    if a:
        a.foo()

def bad(b: Optional[C]) -> None:
    if a := b:
        a.foo()  # E:  Item "None" of "Optional[C]" has no attribute "foo"
```

hauntsaninja:
Fixes #7313
Fixes #7316

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3937,6 +3937,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                 partial_type_maps.append((if_map, else_map))
 
             return reduce_partial_conditional_maps(partial_type_maps)
+        elif isinstance(node, AssignmentExpr):
+            return self.find_isinstance_check_helper(node.target)
         elif isinstance(node, RefExpr):
             # Restrict the type of the variable to True-ish/False-ish in the if and else branches
             # respectively
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2772,6 +2772,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         value = self.accept(e.value)
         self.chk.check_assignment(e.target, e.value)
         self.chk.check_final(e)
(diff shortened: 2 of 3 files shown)


hauntsaninja:
Let me know if there's a better way to handle the partial types that can result! Or if there are more tests I should be adding. There's a lot that I don't know :-)

I guess one thing I was wondering if there's a way to improve the hint resulting from the partial type (see line 296 on `test-data/unit/check-python38.test`). Worried that it might be hard for users to realise that you would have to put the annotation on another line.

JelleZijlstra:
I doubt that a lot of users are going to write `if (x := []):` (why would you do that except to confuse a typechecker?), so I don't think it's worth it to invest too much in improving the error message. We can revisit that if we get user complaints.

TH3CHARLie:
Will you have an interest in testing against #7313 to see if the PR also happens to fix that issue?

hauntsaninja:
Yes, it fixes #7313 too (updated the PR to mention that). A test exists for that issue on line 269 of `test-data/unit/check-python38.test`.

Looking through other issues, it does not fix #8236, but I can maybe work on that (in a separate PR).

PR #8259, opened 2020-01-08
Fix module alias as instance attribute

### LINKED ISSUE #4291: Module alias as instance attribute does not work
The error generated by the following program is unexpected:

```py
import os

class C:
    def __init__(self) -> None:
        self.os = os
        self.os.chdir('/')  # Error: Module has no attribute "chdir"
        os.chdir('/')  # Ok

C()  # No runtime error
```

xhochy:
Fixes #4291

This is only a partial fix. As I'm new to the codebase, I'm unaware of the way an assignment correctly works. Thus also one of the newly added tests still fail. See the inline comment on the code change for the problem.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -3120,12 +3120,14 @@ class SemanticAnalyzer(NodeVisitor[None],
             rnode = self.lookup_type_node(rval)
             if rnode and isinstance(rnode.node, MypyFile):
                 for lval in lvals:
-                    if not isinstance(lval, NameExpr):
+                    if not isinstance(lval, RefExpr):
                         continue
                     # respect explicitly annotated type
                     if (isinstance(lval.node, Var) and lval.node.type is not None):
                         continue
                     lnode = self.current_symbol_table().get(lval.name)
+                    if lnode is None and self.type is not None:
+                        lnode = self.type.names.get(lval.name)
                     if lnode:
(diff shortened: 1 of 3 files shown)


mypy/semanal.py
```diff
@@ -3142,12 +3142,14 @@ def process_module_assignment(self, lvals: List[Lvalue], rval: Expression,
...
             if rnode and isinstance(rnode.node, MypyFile):
                 for lval in lvals:
-                    if not isinstance(lval, NameExpr):
+                    if not isinstance(lval, RefExpr):
                         continue
                     # respect explicitly annotated type
                     if (isinstance(lval.node, Var) and lval.node.type is not None):
                         continue
                     lnode = self.current_symbol_table().get(lval.name)
+                    if lnode is None and self.type is not None:
```
xhochy: The symbol table selection doesn't work correctly. In the case of the the `self.<var> = x` definitions, the result of `current_symbol_table` returns the `warnings` variable the `import` statement generated when run in the context of the `__init__` function. But we actually want to have `lnode = self.type.names.get(lval.name)`. Is there a better way to get the correct symbol table for the `lval`?

xhochy: Would `lval.node.info.names.get(lval.name)` be a better one?

xhochy:
@msullivan Thanks!

PR #8271, opened 2020-01-10
Fix some crashes in dataclasses

### LINKED ISSUE #6809: AssertionError: All arguments must be fully typed
* Are you reporting a bug, or opening a feature request? **bug**

```

@dataclass(frozen=True)
class Modifiers:
    pass

@dataclass(frozen=True)
class Buff:
    modifiers: Modifiers

@dataclass(frozen=True)
class HB(Buff):
    modifiers = Modifiers()
```

```
bug.py:11: error: INTERNAL ERROR -- please report a bug at https://github.com/python/mypy/issues version: 0.701
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/bin/mypy", line 10, in <module>
    sys.exit(console_entry())
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/__main__.py", line 7, in console_entry
    main(None)
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/main.py", line 84, in main
    res = build.build(sources, options, None, flush_errors, fscache)
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/build.py", line 163, in build
    result = _build(sources, options, alt_lib_path, flush_errors, fscache)
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/build.py", line 218, in _build
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/build.py", line 2461, in dispatch
    process_graph(graph, manager)
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/build.py", line 2761, in process_graph
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/build.py", line 2862, in process_stale_scc
    graph[id].semantic_analysis()
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/build.py", line 1997, in semantic_analysis
    self.manager.semantic_analyzer.visit_file(self.tree, self.xpath, self.options, patches)
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/semanal.py", line 309, in visit_file
    self.accept(d)
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/semanal.py", line 3791, in accept
    node.accept(self)
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/nodes.py", line 896, in accept
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/semanal.py", line 807, in visit_class_def
    self.analyze_class(defn)
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/semanal.py", line 825, in analyze_class
    self.analyze_class_body_common(defn)
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/semanal.py", line 833, in analyze_class_body_common
    self.apply_class_plugin_hooks(defn)
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/semanal.py", line 895, in apply_class_plugin_hooks
    hook(ClassDefContext(defn, decorator, self))
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/plugins/dataclasses.py", line 330, in dataclass_class_maker_callback
    transformer.transform()
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/plugins/dataclasses.py", line 100, in transform
    return_type=NoneTyp(),
  File "/Users/ezyang/Dev/oss/dl-hmc-calc-env/lib/python3.7/site-packages/mypy/plugins/common.py", line 97, in add_method
    assert arg.type_annotation, 'All arguments must be fully typed.'
AssertionError: All arguments must be fully typed.
bug.py:11: : note: use --pdb to drop into pdb
```

No crash

```
(/Users/ezyang/Dev/oss/dl-hmc-calc-env) MacBook-Pro-116:dl-hmc-calc ezyang$ python --version
Python 3.7.3
(/Users/ezyang/Dev/oss/dl-hmc-calc-env) MacBook-Pro-116:dl-hmc-calc ezyang$ mypy --version
mypy 0.701
```

*  Do you see the same issue after installing mypy from Git master?

I didn't try.

None

msullivan:
Store the type of an attribute in `DataclassAttribute` so we don't
need to pull it out of `TypeInfo` in a fragile way (since the child
might try to override it in a way that breaks things). This also
allows us to get rid of some pretty dodgy code having to do with
InitVars (that could cause an incremental mode crash.)

Fixes #6809 and an unreported incremental bug

### DIFF
--- a/mypy/plugins/dataclasses.py
+++ b/mypy/plugins/dataclasses.py
@@ -55,6 +55,7 @@ class DataclassAttribute:
         return Var(self.name, self.type)
 
     def serialize(self) -> JsonDict:
+        assert self.type
         return {
             'name': self.name,
             'is_in_init': self.is_in_init,
@@ -62,7 +63,7 @@ class DataclassAttribute:
             'has_default': self.has_default,
             'line': self.line,
             'column': self.column,
-            'type': self.type.serialize() if self.type else None,
+            'type': self.type.serialize(),
         }
 
     @classmethod
@@ -70,8 +71,7 @@ class DataclassAttribute:
         cls, info: TypeInfo, data: JsonDict, api: SemanticAnalyzerPluginInterface
     ) -> 'DataclassAttribute':
         data = data.copy()
-        typ_serialized = data.pop('type')
-        typ = deserialize_and_fixup_type(typ_serialized, api) if typ_serialized else None
+        typ = deserialize_and_fixup_type(data.pop('type'), api)
         return cls(type=typ, **data)
 
 
--- a/mypy/plugins/dataclasses.py
+++ b/mypy/plugins/dataclasses.py
@@ -69,6 +69,7 @@ class DataclassAttribute:
     def deserialize(
         cls, info: TypeInfo, data: JsonDict, api: SemanticAnalyzerPluginInterface
     ) -> 'DataclassAttribute':
+        data = data.copy()
         typ_serialized = data.pop('type')
         typ = deserialize_and_fixup_type(typ_serialized, api) if typ_serialized else None
         return cls(type=typ, **data)
--- a/mypy/plugins/dataclasses.py
+++ b/mypy/plugins/dataclasses.py
@@ -287,7 +287,7 @@ class DataclassTransformer:
                 has_default=has_default,
                 line=stmt.line,
                 column=stmt.column,
-                type=cls.info[lhs.name].type,
+                type=sym.type,
             ))
 
         # Next, collect attributes belonging to any class in the MRO
--- a/mypy/plugins/dataclasses.py
+++ b/mypy/plugins/dataclasses.py
@@ -5,12 +5,14 @@ from typing_extensions import Final
 
 from mypy.nodes import (
     ARG_OPT, ARG_POS, MDEF, Argument, AssignmentStmt, CallExpr,
-    Context, Expression, FuncDef, JsonDict, NameExpr, RefExpr,
+    Context, Expression, JsonDict, NameExpr, RefExpr,
     SymbolTableNode, TempNode, TypeInfo, Var, TypeVarExpr, PlaceholderNode
 )
-from mypy.plugin import ClassDefContext
-from mypy.plugins.common import add_method, _get_decorator_bool_argument
-from mypy.types import Instance, NoneType, TypeVarDef, TypeVarType, get_proper_type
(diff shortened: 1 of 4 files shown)


mypy/plugins/dataclasses.py
```diff
@@ -282,6 +287,7 @@ def collect_attributes(self) -> Optional[List[DataclassAttribute]]:
                 has_default=has_default,
                 line=stmt.line,
                 column=stmt.column,
+                type=cls.info[lhs.name].type,
```
JukkaL: Could this be replaced with `sym.type`?

mypy/plugins/dataclasses.py
```diff
@@ -81,12 +91,7 @@ def transform(self) -> None:
             # Some definitions are not ready, defer() should be already called.
             return
         for attr in attributes:
-            node = info.get(attr.name)
-            if node is None:
-                # Nodes of superclass InitVars not used in __init__ cannot be reached.
-                assert attr.is_init_var and not attr.is_in_init
-                continue
-            if node.type is None:
+            if attr.type is None:
```
JukkaL: What if the some attribute has no type? We seem to handle this case during deserialization, so maybe it can happen here as well. Could this result in always deferring for some attribute? If not, maybe we should require that the serialized type is not `None`?

msullivan: We only consider something as being an attribute if it has a type annotation, so the type missing is (should be) always due to resolution issues that can be fixed by deferring.

I think we can only serialize things with non-None types, yeah.

PR #8288, opened 2020-01-16
stubtest: various fixes

hauntsaninja:
This is enough to get stubtest working again, e.g., with `python3 stubtest.py zipfile`.

I'm not familiar enough with the code to know precisely why we need to disable incremental, but from tracing the code we seemed to run into issues with `mypy.Build.State.meta` – let me know if there's a better fix.


hauntsaninja:
Yes, I'm working on an overhaul of stubtest that implements the things you mentioned in https://github.com/python/typeshed/issues/754 and a couple more, with the goal of improving tests for typeshed. I'll open up a couple issues when I think it's ready, but WIP is at https://github.com/hauntsaninja/mypy/tree/stubtest2/scripts

PR #8290, opened 2020-01-16
Include --strict-equality in --strict

### LINKED ISSUE #7910: Consider adding --strict-equality to --strict
This doesn't report any errors with `--strict-equality`:

```python
def foo() -> bool:
    x: int = 5
    y: str = "foo"
    return x == y


def bar() -> bool:
    x = 5
    y = "bar"
    return x == y


def is_magic(x: bytes) -> bool:
    # Error: Non-overlapping equality check (left operand type: "bytes",
    #        right operand type: "str")  [comparison-overlap]
    return x == "magic"


if __name__ == "__main__":
    print(foo())
    print(bar())
    print(is_magic(b"magic"))
```

even though the third example [is from the docs](https://github.com/python/mypy/blob/d7268ee3336a7d2bba894794a6193872631414a5/docs/source/error_code_list2.rst#check-that-comparisons-are-overlapping-comparison-overlap).

Versions:
- Python: 3.7.4
- mypy: 0.750+dev.59617e8f895ca12a6fb0ec46f378ce384a757118

JukkaL:
Fixes #7910.

The stub changes are needed for clean output when using `--strict`.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -602,7 +602,7 @@ def process_options(args: List[str],
                         help="Treat imports as private unless aliased",
                         group=strictness_group)
 
-    add_invertible_flag('--strict-equality', default=False, strict_flag=False,
+    add_invertible_flag('--strict-equality', default=False, strict_flag=True,
                         help="Prohibit equality, identity, and container checks for"
                              " non-overlapping types",
                         group=strictness_group)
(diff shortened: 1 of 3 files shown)


JukkaL:
I'm not going to wait for AppVeyor.

PR #8302, opened 2020-01-17
Don't make underscored attributes as enum members

### LINKED ISSUE #5312: Private property in subclass of Enum is not indexable
Hi,

Reporting a minor bug in `--strict` mode when adding private property to Enum subclass.

```python
import operator
import enum
from typing import Callable

@enum.unique
class Comparator(enum.Enum):
    LessThan = "<"
    LessThanOrEqualTo = "<="
    EqualTo = "=="
    NotEqualTo = "!="
    GreaterThanOrEqualTo = ">="
    GreaterThan = ">"

    __operator__ = {
        LessThan: operator.lt,
        LessThanOrEqualTo: operator.le,
        EqualTo: operator.eq,
        NotEqualTo: operator.ne,
        GreaterThanOrEqualTo: operator.ge,
        GreaterThan: operator.gt
    }

    @property
    def operator(self) -> Callable[[object, object], bool]:
        return Comparator.__operator__[self.value]

Comparator.LessThan.operator(1,2)
Comparator("<").operator(1,2)
```
The output is the following, where the first line is a missing stub for `operator`, and the second is the error in this report:
```
<string>:26: warning: Returning Any from function declared to return "Callable[[object, object], bool]"
<string>:26: error: Value of type "Comparator" is not indexable
```
`<string>:26: error: Value of type "Comparator" is not indexable` should not appear.

mypy: 0.610
python: 3.6.5
Could not run mypy from master

`--strict`

TH3CHARLie:
fixes #5312

From @JelleZijlstra 's description, for underscored attributes, we don't make them as enum members


bluetech:
A bit silly, but the output of this program

```py
import enum

class E(enum.Enum):
    _ = 1
    __ = 2
    ___ = 3
    ____ = 4
    _____ = 5
    __a__ = 6

print([e.value for e in E])
```

is `[1, 2, 3, 4, 5]`, but your check `if name.startswith('__') and name.endswith("__")` rejects 2, 3, 4 and 5 :smile:

TH3CHARLie:
> ```python
> print([e.value for e in E])
> ```

Good catch, I wouldn't think anyone would write code like this but it is definitely a valid case, thanks!

PR #8304, opened 2020-01-19
Allow Travis for other than python/mypy repos, e.g. contributor private

scop:
Whenever I want to submit something, I find myself needing to temporarily comment out the .travis.yml `if` condition, commit to get CI tests done and pass in my private fork and Travis, then rebase to remove that commit before I submit the PR. This should eliminate that extra annoyance.


Michael0x2a:
Thanks!

PR #8305, opened 2020-01-20
Make isinstance/issubclass generate ad-hoc intersections

Michael0x2a:
This diff makes `isinstance(...)` and `issubclass(...)` try generating ad-hoc intersections of Instances when possible.

For example, we previously concluded the if-branch is unreachable in the following program. This PR makes mypy infer an ad-hoc intersection instead.

    class A: pass
    class B: pass

    x: A
    if isinstance(x, B):
        reveal_type(x)  # N: Revealed type is 'test.<subclass of "A" and "B">'

If you try doing an `isinstance(...)` that legitimately is impossible due to conflicting method signatures or MRO issues, we continue to declare the branch unreachable. Passing in the `--warn-unreachable` flag will now also report an error about this:

    # flags: --warn-unreachable
    x: str

    # E: Subclass of "str" and "bytes" cannot exist: would have
    #    incompatible method signatures
    if isinstance(x, bytes):
        reveal_type(x)  # E: Statement is unreachable

This error message has the same limitations as the other `--warn-unreachable` ones: we suppress them if the isinstance check is inside a function using TypeVars with multiple values.

However, we *do* end up always inferring an intersection type when possible -- that logic is never suppressed.

I initially thought we might have to suppress the new logic as well (see https://github.com/python/mypy/issues/3603#issuecomment-506996850), but it turns out this is a non-issue in practice once you add in the check that disallows impossible intersections.

For example, when I tried running this PR on the larger of our two internal codebases, I found about 25 distinct errors, all of which were legitimate and unrelated to the problem discussed in the PR.

(And if we don't suppress the extra error message, we get about 100-120 errors, mostly due to tests repeatdly doing `result = blah()` followed by `assert isinstance(result, X)` where X keeps changing.)

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -4804,7 +4804,11 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                                                expr_type: Type,
                                                type_ranges: Optional[List[TypeRange]],
                                                ) -> Tuple[TypeMap, TypeMap]:
-        yes_map, no_map = conditional_type_map(expr, expr_type, type_ranges)
+        # For some reason, doing "yes_map, no_map = conditional_type_maps(...)"
+        # doesn't work: mypyc will decide that 'yes_map' is of type None if we try.
+        initial_maps = conditional_type_map(expr, expr_type, type_ranges)
+        yes_map = initial_maps[0]  # type: TypeMap
+        no_map = initial_maps[1]  # type: TypeMap
 
         if yes_map is not None or type_ranges is None:
             return yes_map, no_map
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -38,7 +38,7 @@ from mypy.types import (
     get_proper_types, is_literal_type, TypeAliasType)
 from mypy.sametypes import is_same_type
 from mypy.messages import (
-    MessageBuilder, make_inferred_type_note, append_invariance_notes,
+    MessageBuilder, make_inferred_type_note, append_invariance_notes, pretty_seq,
     format_type, format_type_bare, format_type_distinctly, SUGGESTED_TEST_FIXTURES
 )
 import mypy.checkexpr
@@ -63,7 +63,7 @@ from mypy.constraints import SUPERTYPE_OF
 from mypy.maptype import map_instance_to_supertype
 from mypy.typevars import fill_typevars, has_no_typevars, fill_typevars_with_any
 from mypy.semanal import set_callable_name, refers_to_fullname
-from mypy.mro import calculate_mro
+from mypy.mro import calculate_mro, MroError
 from mypy.erasetype import erase_typevars, remove_instance_last_known_values, erase_type
 from mypy.expandtype import expand_type, expand_type_by_instance
 from mypy.visitor import NodeVisitor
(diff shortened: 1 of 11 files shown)


Michael0x2a:
Not really sure what's up with the tests though -- they seem to be crashing for unrelated reasons?

ethanhs:
@Michael0x2a I think https://github.com/python/mypy/pull/8308 fixes the failing tests.

ilevkivskyi:
A quick question before I will do a full review: last time I was thinking about this my main worry was about how the fake intersections would interact with incremental mode. Did you try this in incremental mode? Maybe add few tests where generated `TypeInfo` "leaks" (for example via assignment to `self`) and therefore gets serialized and subsequently deserialized.

msullivan:
I'm also interested in how this interacts with mypyc---though making it work with mypyc I don't think is actually a requirement to merge this, but would be great to tackle if you are interested.

(These intersection issues can cause runtime errors under mypyc, which is very frustrating.)

Michael0x2a:
Ok, I added some incremental mode tests.

Just as a disclaimer though, I'm pretty rusty on anything to do with incremental mode, so I'm not actually sure if the new tests are high quality. In particular:

1. I'm a little suspicious of the new test results in deps.test. If I'm understanding the results correctly, any change to either of the original classes should end up triggering a reanalysis of whatever scope happens to be creating the ad-hoc intersection -- but it's a little surprising to me that I don't see any triggers along the lines of `<m.A> -> <m.<subclass of "A" and "B">>`.
2. I copied the incremental mode tests into the fine-grained mode tests, but I'm not actually sure if that's actually a useful thing to do.

If it's any consolation, I'm creating the ad-hoc intersection using the same code we were previously using in `intersect_callable`, which was added to mypy ~2 years ago. Basically, that method was generating a new classdef + typeinfo then adding the new typeinfo to the global scope.

I can try looking into adding support for this in mypyc, sure.

msullivan:
I know that `intersect_callable` had some issue with `astmerge` where it wouldn't necessarily get  merged correctly. Since it was pretty obscure anyway (it was my starter project on mypy! :P) I don't think we worried about it too much and I'm not sure if we ever figured out how to make a test that failed. But I think it *did* fail the astmerge consistency checker. (Which actually might just be completely broken because I don't think we've run it in a year and a half and it never passed on everything anyway)

msullivan:
I'm hacking on the consistency checker to get it back into shape and while there are some failures it looks like none of them are related to the `callable` test. So I think that the new semantic analyzer must have fixed that issue, which makes me a lot less nervous about these changes.

mypy/checker.py
```diff
@@ -4726,10 +4795,52 @@ def infer_issubclass_maps(self, node: CallExpr,
...
         yes_map, no_map = map(convert_to_typetype, (yes_map, no_map))
         return yes_map, no_map
 
+    def conditional_type_map_with_intersection(self,
+                                               expr: Expression,
+                                               expr_type: Type,
+                                               type_ranges: Optional[List[TypeRange]],
+                                               ) -> Tuple[TypeMap, TypeMap]:
+        # For some reason, doing "yes_map, no_map = conditional_type_maps(...)"
+        # doesn't work: mypyc will decide that 'yes_map' is of type None if we try.
```
msullivan: weird

mypy/checker.py
```diff
@@ -4726,10 +4795,52 @@ def infer_issubclass_maps(self, node: CallExpr,
...
+        initial_maps = conditional_type_map(expr, expr_type, type_ranges)
+        yes_map = initial_maps[0]  # type: TypeMap
+        no_map = initial_maps[1]  # type: TypeMap
+
+        if yes_map is not None or type_ranges is None:
+            return yes_map, no_map
+
+        # If we couldn't infer anything useful, try again by trying to compute an intersection
```
msullivan: Elaborate a bit more maybe?

mypy/checker.py
```diff
@@ -3636,6 +3638,78 @@ def visit_continue_stmt(self, s: ContinueStmt) -> None:
...
+        info.bases = bases
+        calculate_mro(info)
+        info.calculate_metaclass_type()
+        return cdef, info
+
+    def intersect_instances(self,
+                            instances: Sequence[Instance],
+                            ctx: Context,
+                            ) -> Optional[Instance]:
+        curr_module = self.scope.stack[0]
```
msullivan: Document what we're doing here and why?

Issue #8306, opened 2020-01-21
Pretty print option breaks when tabs are in source


dosisod:
Please provide more information to help us understand the issue:

* Are you reporting a bug, or opening a feature request?
Bug + PR
* Please insert below the code you are checking with mypy:
```python
# 4spaces.py
def test_4spaces() -> str:
    return None
test_4spaces()

# 8spaces.py
def test_8spaces() -> str:
        return None
test_8spaces()

# tabs.py
def test_tabs() -> str:
	return None
test_tabs()
```
* What is the actual behavior/output?
```
$ mypy *.py --pretty
tabs.py:2: error: Incompatible return value type (got "None", expected "str")
    	return None
            ^
8spaces.py:2: error: Incompatible return value type (got "None", expected "str")
            return None
                   ^
4spaces.py:2: error: Incompatible return value type (got "None", expected "str")
        return None
               ^
```
* What is the behavior/output you expect?
The `^` indicator to be in correct position in all 3 examples. Fixed in PR.
* What are the versions of mypy and Python you are using?
mypy 0.770+dev.861f01c3be99b9faf2e8d42e316aee4ec6e571c4.dirty
Python 3.6.9
* Do you see the same issue after installing mypy from Git master?
Yes.
* What are the mypy flags you are using? (For example --strict-optional)
--pretty

In most terminals the `tab` character will be expanded to the nearest tab column, that being every 8 characters.

When mypy is displaying the source, lines with tabs gobble up space, and the column attribute treats tabs as a single character and not 8. This causes many problems when displaying source lines with tabs.

ilevkivskyi:
Fixed by https://github.com/python/mypy/pull/8307

PR #8307, opened 2020-01-21
fixed indicator position in code with tabs

dosisod:
PR for issue #8306.

### DIFF
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -451,12 +451,17 @@ class Errors:
                 # Add source code fragment and a location marker.
                 if severity == 'error' and source_lines and line > 0:
                     source_line = source_lines[line - 1]
+                    source_line_expanded = source_line.replace('\t', ' ' * 8)
                     if column < 0:
                         # Something went wrong, take first non-empty column.
                         column = len(source_line) - len(source_line.lstrip())
+
+                    # Shifts column after tab expansion
+                    column = len(source_line[:column].replace('\t', ' ' * 8))
+
                     # Note, currently coloring uses the offset to detect source snippets,
                     # so these offsets should not be arbitrary.
-                    a.append(' ' * DEFAULT_SOURCE_OFFSET + source_line)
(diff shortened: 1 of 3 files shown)


dosisod:
Now that the tests are actually failing for code related reasons, what is the suggested change for the 2 failing tests?

https://github.com/python/mypy/blob/5d8c0cd28f8c038e853731d2cd12578412945beb/test-data/unit/fine-grained-blockers.test#L27

The cursor is off by 1 character, though I would think that the new position of the cursor makes more sense. Also, where should I put new tests to check for consistency between tabs and spaces when pretty printing?

mypy/errors.py
```diff
@@ -451,12 +451,17 @@ def format_messages(self, error_info: List[ErrorInfo],
                 # Add source code fragment and a location marker.
                 if severity == 'error' and source_lines and line > 0:
                     source_line = source_lines[line - 1]
+                    source_line_expanded = source_line.replace('\t', ' ' * 8)
```
ilevkivskyi: Why not `source_line.expandtabs()`? (Also 8 is default for the latter.)

mypy/errors.py
```diff
@@ -451,12 +451,17 @@ def format_messages(self, error_info: List[ErrorInfo],
...
                     if column < 0:
                         # Something went wrong, take first non-empty column.
                         column = len(source_line) - len(source_line.lstrip())
+
+                    # Shifts column after tab expansion
+                    column = len(source_line[:column].replace('\t', ' ' * 8))
```
ilevkivskyi: Same question here.

ilevkivskyi:
> The cursor is off by 1 character, though I would think that the new position of the cursor makes more sense.

This doesn't look important, you can just update the existing tests.

> Also, where should I put new tests to check for consistency between tabs and spaces when pretty printing?

I would put them at the end of `cmdline.test`.

dosisod:
Did not know that `expandtabs()` existed. Changed that, then added and modified the tests.

PR #8312, opened 2020-01-22
Add new error code for unreachable statement

### LINKED ISSUE #8190: Unreachable Warning Code
There are sometimes cases where I'd like deliberately put something in a statically-unreachable section---maybe some asserts, maybe some debug info, etc.---in case it does hit that case at runtime.  I currently silence them with a `# type: ignore[misc]` which isn't particularly informative to someone reading the code.  Could we give `unreachable` its own dedicated error/warning code?

EDIT: just so it's clear, I do want to keep the warning on to detect legitimate issues.

DenisOH:
Closes #8190.
This introduces a new error for `unreachable_statement`.

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -1249,7 +1249,8 @@ class MessageBuilder:
         it does not change the truth value of the entire condition as a whole.
         'op_name' should either be the string "and" or the string "or".
         """
-        self.fail("Right operand of '{}' is never evaluated".format(op_name), context)
+        self.fail("Right operand of '{}' is never evaluated".format(op_name),
+                  context, code=codes.UNREACHABLE)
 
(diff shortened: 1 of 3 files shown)


mypy/messages.py
```diff
@@ -1235,7 +1235,7 @@ def note_call(self,
                   context, code=code)
 
     def unreachable_statement(self, context: Context) -> None:
-        self.fail("Statement is unreachable", context)
+        self.fail("Statement is unreachable", context, code=codes.UNREACHABLE)
 
     def redundant_left_operand(self, op_name: str, context: Context) -> None:
```
Michael0x2a: Can you also include the new error code for the `redundant_blah` error messages listed below here?

These error messages use the same reachability inference logic and are also reported only when the `--warn-unreachable` flag is set.

DenisOH:
@Michael0x2a Sure!

PR #8313, opened 2020-01-22
Make the fine-grained mergechecker work again

msullivan:
This fixes some stuff in the mergechecker that breaks with current
mypy's and also some bugs that the mergechecker caught. The entire
fine-grained test suite now passes with the merge checker on, which I
don't think has ever been true before.

This means that we could turn it on by default, but it doubles the
runtime of the fine-grained tests (from 90s CPU time to 180s CPU time
on my laptop), so I've left it off for now.

The motivation here is that I knew intersect_callable's creation of
types during typechecking used to run afoul of the consistency checker
and so I was nervous that #8305 would cause more problems by adding
more logic of that kind. It no longer does, probably as a result of
the semantic analyzer rewrite, so I think we are in the clear on that.

### DIFF
--- a/mypy/semanal_main.py
+++ b/mypy/semanal_main.py
@@ -335,6 +335,12 @@ def semantic_analyze_target(target: str,
         priority = mypy.build.PRI_LOW
         if priority <= state.priorities.get(dep, priority):
             state.priorities[dep] = priority
+
+    # Clear out some stale data to avoid memory leaks and astmerge
+    # validity check confusion
+    analyzer.statement = None
+    delattr(analyzer, 'cur_mod_node')
+
     if analyzer.deferred:
         return [target], analyzer.incomplete, analyzer.progress
     else:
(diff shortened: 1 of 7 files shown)


mypy/semanal_main.py
```diff
@@ -335,6 +335,12 @@ def semantic_analyze_target(target: str,
         priority = mypy.build.PRI_LOW
         if priority <= state.priorities.get(dep, priority):
             state.priorities[dep] = priority
+
+    # Clear out some stale data to avoid memory leaks and astmerge
+    # validity check confusion
+    analyzer.statement = None
+    delattr(analyzer, 'cur_mod_node')
```
ilevkivskyi: Why not `del analyzer.cur_mod_node`?

msullivan: Because I forgot how to write python apparently

PR #8325, opened 2020-01-24
stubtest: rewrite

### LINKED ISSUE #3329: Document and improve stubtest
See https://github.com/python/typeshed/issues/1243#issuecomment-299497346

@JukkaL seems to be volunteering. Also calling out to @jgarst who seems to have worked on this with Jukka during the hack week.

hauntsaninja:
Hello,

I recently embarked on a rewrite of stubtest with the goal of improving our testing of typeshed. For some more context, please refer to https://github.com/python/typeshed/issues/754 That issue also links examples of output and of typeshed issues this has detected.

Some notes about what this change contains:
- Previously stubtest.py used dumpmodule.py to generate a JSON representation of runtime objects. This abandons that approach in favour of directly using runtime objects, allowing us easier and more dynamic introspection.
- Much of stubtest.py was unimplemented, most notably, examining functions. This provides new and fairly detailed implementations of verify_funcitem, verify_overloadedfuncdef, verify_var, etc.
- In particular, we can now do everything (and more!) @JukkaL mentioned in the original post at https://github.com/python/typeshed/issues/754 
- There's a more complete CLI, with features to make testing typeshed easier.
- There's more documentation.

My goal with this draft PR is to solicit code review and feedback from folks familiar with mypy internals. Some notes:
- I'm still working on it!
- The code doesn't match mypy's style, eg, line length is shorter than 99 and the script currently doesn't support py35
- I haven't written any tests yet; I've just been running this against typeshed.
- There's some discussion about where this script should live over at https://github.com/python/typeshed/issues/754 -- curious if mypy has opinions on this.
- The previous stubtest.py has been broken by changes to mypy internals multiple times before. If we can merge it, is the extra maintenance something mypy is okay with? We'd also have to come up with good CI tests that aren't checking typeshed, to avoid horrible cross-project build dependencies.

Anyway, let me know your thoughts!

Resolves #3329

### DIFF
--- a/scripts/stubtest.py
+++ b/scripts/stubtest.py
@@ -978,14 +978,11 @@ def get_whitelist_entries(whitelist_file: Optional[str]) -> Iterator[str]:
 def main() -> int:
     assert sys.version_info >= (3, 5), "This script requires at least Python 3.5"
 
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(
+        description="Compares stubs to objects introspected from the runtime."
+    )
     parser.add_argument("modules", nargs="*", help="Modules to test")
-    parser.add_argument(
-        "--check-typeshed", action="store_true", help="Check all stdlib modules in typeshed"
-    )
-    parser.add_argument(
-        "--custom-typeshed-dir", metavar="DIR", help="Use the custom typeshed in DIR"
-    )
+    parser.add_argument("--concise", action="store_true", help="Make output concise")
     parser.add_argument(
         "--ignore-missing-stub",
         action="store_true",
@@ -997,6 +994,12 @@ def main() -> int:
         help="Ignore errors for whether an argument should or shouldn't be positional-only",
     )
     parser.add_argument(
+        "--custom-typeshed-dir", metavar="DIR", help="Use the custom typeshed in DIR"
+    )
+    parser.add_argument(
+        "--check-typeshed", action="store_true", help="Check all stdlib modules in typeshed"
+    )
+    parser.add_argument(
         "--whitelist",
         action="append",
         metavar="FILE",
@@ -1006,7 +1009,6 @@ def main() -> int:
             "whitelists. Whitelist can be created with --generate-whitelist"
         ),
     )
-    parser.add_argument("--concise", action="store_true", help="Make output concise")
     parser.add_argument(
         "--generate-whitelist",
         action="store_true",
--- a/scripts/stubtest.py
+++ b/scripts/stubtest.py
@@ -179,9 +179,7 @@ def verify(
 
 @verify.register(nodes.MypyFile)
 def verify_mypyfile(
-    stub: nodes.MypyFile,
-    runtime: MaybeMissing[types.ModuleType],
-    object_path: List[str]
+    stub: nodes.MypyFile, runtime: MaybeMissing[types.ModuleType], object_path: List[str]
 ) -> Iterator[Error]:
     if isinstance(runtime, Missing):
         yield Error(object_path, "is not present at runtime", stub, runtime)
@@ -198,9 +196,7 @@ def verify_mypyfile(
     )
     # Check all things declared in module's __all__
     to_check.update(getattr(runtime, "__all__", []))
-    to_check.difference_update(
-        {"__file__", "__doc__", "__name__", "__builtins__", "__package__"}
-    )
+    to_check.difference_update({"__file__", "__doc__", "__name__", "__builtins__", "__package__"})
     # We currently don't check things in the module that aren't in the stub, other than things that
     # are in __all__, to avoid false positives.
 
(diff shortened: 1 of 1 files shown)


hauntsaninja:
Oh huh, didn't know about `proper_plugin.py`. That's neat.

scripts/stubtest.py
```diff
@@ -1,234 +1,1008 @@
...
+                    and len(dec.decorators) == 1
+                    and isinstance(dec.decorators[0], nodes.NameExpr)
+                    and dec.decorators[0].fullname == "typing.overload"
+                ):
+                    yield dec.func
+                else:
+                    raise ValueError
+
+        # For all dunder methods other than __init__, just assume all args are positional-only
+        assume_positional_only = stub.name != "__init__" and stub.name.startswith("__")
```
JelleZijlstra: Should check that is also endswith `__`

scripts/stubtest.py
```diff
@@ -1,234 +1,1008 @@
...
+) -> Iterator[str]:
+    # Check positional arguments match up
+    for stub_arg, runtime_arg in zip(stub.pos, runtime.pos):
+        yield from _verify_arg_name(stub_arg, runtime_arg, function_name)
+        yield from _verify_arg_default_value(stub_arg, runtime_arg)
+        if (
+            runtime_arg.kind == inspect.Parameter.POSITIONAL_ONLY
+            and not stub_arg.variable.name.startswith("__")
+            and not stub_arg.variable.name.strip("_") == "self"
+            and not function_name.startswith("__")  # noisy for dunder methods
```
JelleZijlstra: Same here

ilevkivskyi:
@JelleZijlstra thanks for reviewing this! Please feel free to merge this when you think this is ready.

scripts/stubtest.py
```diff
@@ -1,234 +1,1008 @@
...
+
+        :param concise: Whether to return a concise, one-line description
+
+        """
+        if concise:
+            return _style(self.object_desc, bold=True) + " " + self.message
+
+        stub_line = None
+        stub_file = None
+        if not isinstance(self.stub_object, Missing):
```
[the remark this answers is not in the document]
hauntsaninja: Ideally yes, but then mypy can't narrow the Union.

hauntsaninja:
I'm done making changes; this is ready for review 😄

hauntsaninja:
Okay, I lied. But only because I didn't expect to run into a Python bug. (Up till now I'd just been testing on my Mac, bug is Linux-only because it manifests with `select.epoll`)

hauntsaninja:
(This comment contains a patch that detects issues with base classes. Unfortunately, in practice it's noisy. Sticking this here, in case I need to find it again)
<details>

```
--- a/scripts/stubtest.py
+++ b/scripts/stubtest.py
@@ -213,6 +213,31 @@ def verify_mypyfile(
         )
 
 
+def _verify_base_classes(
+    stub: nodes.TypeInfo, runtime: Type[Any], object_path: List[str]
+) -> Iterator[Error]:
+    stub_bases = [(t.type.module_name, t.type.name) for t in stub.bases]
+    runtime_bases = [(t.__module__, t.__name__) for t in runtime.__bases__]
+
+    def should_filter(module_name: str, class_name: str) -> bool:
+        return (
+            module_name in ("typing", "collections.abc", "abc", "contextlib")
+            or (module_name, class_name) == ("builtins", "object")
+        )
+
+    filtered_stub_bases = [t[1] for t in stub_bases if not should_filter(t[0], t[1])]
+    filtered_runtime_bases = [t[1] for t in runtime_bases if not should_filter(t[0], t[1])]
+    if filtered_stub_bases != filtered_runtime_bases:
+        yield Error(
+            object_path,
+            "is inconsistent, base classes differ.",
+            stub,
+            runtime,
+            stub_desc=repr(stub) + ": " + ", ".join(".".join(t) for t in stub_bases),
+            runtime_desc=str(runtime) + ": " + ", ".join(".".join(t) for t in runtime_bases)
+        )
+
+
 @verify.register(nodes.TypeInfo)
 def verify_typeinfo(
     stub: nodes.TypeInfo, runtime: MaybeMissing[Type[Any]], object_path: List[str]
@@ -224,6 +249,8 @@ def verify_typeinfo(
         yield Error(object_path, "is not a type", stub, runtime, stub_desc=repr(stub))
         return
 
+    yield from _verify_base_classes(stub, runtime, object_path)
+
     to_check = set(stub.names)
     to_check.update(m for m in vars(runtime) if not m.startswith("_"))
```
</details>

hauntsaninja:
Oops, didn't notice that one of those broke CI

JelleZijlstra:
Before I merge this, I'd like to make sure that there's something in CI that makes sure stubtest doesn't get broken by changes in other areas of mypy. From https://travis-ci.org/python/mypy/jobs/641260424?utm_medium=notification&utm_source=github_status it looks like the mypy self-check already covers stubtest, which is good.

Would it be possible to add a test that invokes stubtest on either the typeshed submodule or on some special test file? The latter would have the advantage that you could also test for error cases, but the former might be easier to set up.

hauntsaninja:
I can do that. Currently we don't package stubtest, so tox knows nothing about it. I think it would make sense to treat this like stubgen, so move it into mypy/ and add a console script entry_point in setup.py? (I'd also prefer to do it in a separate diff, but can keep stacking on to this too)

JelleZijlstra:
Sure, that sounds good. I'll merge this PR now so that you can do the testing in a followup PR.

PR #8332, opened 2020-01-25
Report an error if final class has abstract attributes

### LINKED ISSUE #8316: Feature request: Disallow final classes from having abstract methods or properties
Currently, mypy checks whether a class is abstract when it's instantiated. For classes with the `@final` decorator, it would make sense for there to be a typing error if they define any abstract methods or properties, since those could never be implemented.

It would also make sense for classes that explicitly have the `ABCMeta` metaclass and also have the `@final` decorator to be a typing error if the class contains any instance methods or properties, since there could never be an instance created to call them with.

Note that `abstract final` classes are an idiom that is sometimes used in Java to define a class that is just a grouping of static methods, but I believe that's mostly a result of Java not having modules and top-level functions. In Python, one could organize top-level functions and group them with a module. This could be an argument to disallow a class to have both `@final` and `ABCMeta` all-together.

DenisOH:
Closes #8316.

### DIFF
--- a/docs/source/class_basics.rst
+++ b/docs/source/class_basics.rst
@@ -321,3 +321,17 @@ class, including an abstract method defined in an abstract base class.
 
 You can implement an abstract property using either a normal
 property or an instance variable.
+
+An abstract class that defines at least one abstract method or
+property and has `@final` decorator will generate an error from
+mypy, since those attributes could never be implemented.
+
+.. code-block:: python
+
+    from abc import ABCMeta, abstractmethod
+    from typing import final
+
+    @final
+    class A(metaclass=ABCMeta):  # Error: Final class A has abstract attributes "f"
+        @abstractmethod
+        def f(self, x: int) -> None: pass
--- a/docs/source/class_basics.rst
+++ b/docs/source/class_basics.rst
@@ -321,17 +321,3 @@ class, including an abstract method defined in an abstract base class.
 
 You can implement an abstract property using either a normal
 property or an instance variable.
-
-An abstract class that defines at least one abstract method or
-property and has `@final` decorator will generate an error from
-mypy, since those attributes could never be implemented.
-
-.. code-block:: python
-
-    from abc import ABCMeta, abstractmethod
-    from typing import final
-
-    @final
-    class A(metaclass=ABCMeta):  # Error: Final class A has abstract attributes "f"
-        @abstractmethod
-        def f(self, x: int) -> None: pass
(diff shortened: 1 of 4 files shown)


DenisOH:
@ilevkivskyi Sure!
Since I'm new to contributing to mypy, could you give me a little hint on where the test would go?

ilevkivskyi:
Probably in `test-data/unit/check-classes.test`

docs/source/class_basics.rst
```diff
@@ -321,3 +321,17 @@ class, including an abstract method defined in an abstract base class.
 
 You can implement an abstract property using either a normal
 property or an instance variable.
+
+An abstract class that defines at least one abstract method or
```
ilevkivskyi: I would rather put this at the end of https://github.com/python/mypy/blob/master/docs/source/final_attrs.rst#final-classes

docs/source/class_basics.rst
```diff
@@ -321,3 +321,17 @@ class, including an abstract method defined in an abstract base class.
...
+property and has `@final` decorator will generate an error from
```
ilevkivskyi:
```suggestion
property and has ``@final`` decorator will generate an error from
```

docs/source/class_basics.rst
```diff
@@ -321,3 +321,17 @@ class, including an abstract method defined in an abstract base class.
...
+mypy, since those attributes could never be implemented.
+
+.. code-block:: python
+
+    from abc import ABCMeta, abstractmethod
+    from typing import final
+
+    @final
+    class A(metaclass=ABCMeta):  # Error: Final class A has abstract attributes "f"
```
ilevkivskyi:
```suggestion
    class A(metaclass=ABCMeta):  # error: Final class A has abstract attributes "f"
```

test-data/unit/check-classes.test
```diff
@@ -6677,3 +6677,20 @@ class B:
                                         # N: Perhaps you need "Callable[...]" or a callback protocol?
 
 [builtins fixtures/classmethod.pyi]
+
+[case testFinalClassWithAbstractAttributes]
+from abc import abstractmethod, ABCMeta
+from typing import final
+
+@final
+class A(metaclass=ABCMeta):  # E: Final class __main__.A has abstract attributes "bar", "foo"
```
ilevkivskyi: Please add also the same test without `ABCMeta`, and a test where abstract attribute is inherited (while the body of final class itself doesn't define abstract attributes).

DenisOH: I've added 2 additional tests, let me know if they correspond to what you meant.

PR #8333, opened 2020-01-26
Fix joining of fixed-length tuples with mismatching lengths

intgr:
For example: `Tuple[bool, int] + Tuple[bool]` becomes `Tuple[int, ...]`

Previously Mypy simply punted and returned `object`.

This solves part of #4975. The other missing handling of fixed tuple + variadic tuple will be implemented separately, if this PR goes well. :)

### DIFF
--- a/mypy/join.py
+++ b/mypy/join.py
@@ -265,6 +265,8 @@ class TypeJoinVisitor(TypeVisitor[ProperType]):
         # When given two fixed-length tuples:
         # * If they have the same length, join their subtypes item-wise:
         #   Tuple[int, bool] + Tuple[bool, bool] becomes Tuple[int, bool]
+        # * If lengths do not match, return a variadic tuple:
+        #   Tuple[bool, int] + Tuple[bool] becomes Tuple[int, ...]
         #
         # Otherwise, `t` is a fixed-length tuple but `self.s` is NOT:
         # * Joining with a variadic tuple returns variadic tuple:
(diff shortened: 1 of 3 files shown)


intgr:
Will add tests once I figure out how to run pytest locally; currently fails with:

> INTERNALERROR>   File "/home/marti/.local/share/virtualenvs/mypy-CahTNRog/lib/python3.8/site-packages/_pytest/nodes.py", line 465, in location
> INTERNALERROR>     assert isinstance(location[0], py.path.local), location[0]
> INTERNALERROR> AssertionError: /home/marti/own/mypy/test-data/unit/check-basic.test

JukkaL:
Thanks! Looks good. Maybe also add simple data-driven test (in `test-data/unit/check-tuples.test`), just in case?

intgr:
I discovered some awkward behavior when attempting to create a sub-subclass of `NamedTuple`. Seems that `NamedTuple` never supported this, but doesn't raise an error either :)

```python
from typing import Tuple, NamedTuple

class NTup1(NamedTuple):
    a: bool
class NTup2(NTup1):
    b: bool

tup1 = None  # type: NTup1
tup2 = None  # type: NTup2

tup1 = tup2  # This should be an error, a 2-tuple type is not compatible with 1-tuple

def bla(x: NTup1): ...
bla(tup2)  # This should be an error, a 2-tuple type is not compatible with 1-tuple
```

intgr:
Pushed additional tests.

intgr:
Rebased after conflicts from merging #8335.

JukkaL:
> I discovered some awkward behavior when attempting to create a sub-subclass of NamedTuple. 

Could you file a separate issue about this?

Issue #8074, opened 2019-12-04
Type of conditional expression is object


cjolowicz:
The type of the following conditional expression is `object`:

```
s: str = "key=value"
a, b = s.split("=", 1) if "=" in s else ("key", "default")
# main.py:2: error: 'builtins.object' object is not iterable
```

It should be a union type instead, such as `Union[List[str], Tuple[str, str]]`.

(Note that `str.split` returns `List[str]` while the `else` branch has the type `Tuple[str, str]`.)

- [Gist](https://gist.github.com/4d8ef543c0a86d6966fa5722c792e345)
- [Playground](https://mypy-play.net/?mypy=latest&python=3.8&gist=4d8ef543c0a86d6966fa5722c792e345)

This is a follow-up to https://github.com/python/mypy/issues/3487.

JukkaL:
Alternatively, inferring `Sequence[str]` for the conditional expression would help in this case.

Inferring a union type would break backward compatibility. If the fallout is not too bad, it may still make sense.

ilevkivskyi:
Note that a good fix for https://github.com/python/mypy/issues/4975 may automatically lead to inferring `Sequence` here.

cjolowicz:
Here is a simpler reproduction:

```
a, b = [1, 2] if bool() else (1, 2)
```

intgr:
This issue is addressed in PR #8335.

ethanhs:
Closing as this was fixed by #8335. Thanks @intgr !

carlosporta:
I am also facing a similar problem. 

    from typing import Protocol


    class Shape(Protocol):
        def area(self) -> str:
            ...


    class Circle:
        def area(self) -> str:
            return 'π * r²'


    class Square:
        def area(self) -> str:
            return 'w * l'


    class ShapeFactory:
        def create(self, shape_type: str) -> Shape:
            return Circle() if shape_type == 'circle' else Square()  # Here

Mypy shows this error message:
> error: Incompatible return value type (got "object", expected "Shape")

intgr:
@carlosporta That's a separate issue. Mypy tries to find the common base class of `Circle` and `Square`; since those two don't subclass `Shape` directly, the common base class happens to be `object`.

carlosporta:
@intgr, I understood what is happening, but this only happens using when using a conditional expression. If I change it to a conditional statement mypy identifies the shape protocol correctly as the return type.

This works perfectly, no error messages on mypy:

    class ShapeFactory:
[...]

PR #8335, opened 2020-01-26
Fix joining of Sequence (e.g. variadic tuple) and fixed-length tuple

intgr:
For example:
* `Tuple[int] + Tuple[bool, ...]` becomes `Tuple[int, ...]`
* `List[int] + Tuple[bool, ...]` becomes `Sequence[int]`

This solves the other part of issue #4975 (along with PR #8333) and also issue #8074. Previously Mypy simply punted and returned `object`.

### DIFF
--- a/mypy/join.py
+++ b/mypy/join.py
@@ -177,6 +177,8 @@ class TypeJoinVisitor(TypeVisitor[ProperType]):
             return join_types(t, self.s)
         elif isinstance(self.s, TypedDictType):
             return join_types(t, self.s)
+        elif isinstance(self.s, TupleType):
+            return join_types(t, self.s)
         elif isinstance(self.s, LiteralType):
             return join_types(t, self.s)
         else:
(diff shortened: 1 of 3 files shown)


JukkaL:
As discussed elsewhere, also testing subclasses of tuple would be a good idea.

intgr:
I added test `testTupleSubclassJoin`, does that cover what you had in mind?

theodoretliu:
Quick question: why is subclassing "automatically" handled by this PR? This is more just for my curiosity, does the "default fallback" function automatically find the ancestor of the class?

intgr:
@theodoretliu The `tuple_fallback` function returns a variadic tuple in place of a fixed-length tuple. But most tuple subclasses only support fixed-length operation, it cannot construct `NamedTuple[int, ...]` so it has to fall back to `builtin.tuple` class for tuple subclasses.

Issue #3157, opened 2017-04-12
Support function decorators excellently


sixolet:
Decorators currently stress mypy's support for functional programming to the point that many decorators are impossible to type.  I'm intending this issue as a project plan and exploration of the issue of decorators and how to type them. It's a collection point for problems or incompletenesses that keep decorators from being fully supported, plans for solutions to those problems, and fully-implemented solutions to those problems.

# Decorators that can only decorate a function of a fixed signature

You can define your decorator's signature explicitly, and it's completely supported:

```python
from typing import Callable, Tuple

def with_strlen(f: Callable[[int], str]) -> Callable[[int], Tuple[str, int]]:
    def ret(__i: int) -> Tuple[str, int]:
        r = f(__i)
        return r, len(r)
    return ret

@with_strlen
def lol(x: int) -> str:
    return "lol"*x

reveal_type(lol)  # E: Revealed type is 'def (builtins.int) -> Tuple[builtins.str, builtins.int]
 ```
 Notes:
* You can even use type variables to get some flexibility in argument and return types, but this falls over as soon as you don't know the exact number of arguments to expect.
* Your decorated function's arguments can't be called by name.
  - [ ] https://github.com/python/mypy/pull/2607 enables that

# Decorators that do not change the signature of the function

Nearly fully supported. Here's how you do it:

```python
from typing import TypeVar, Callable, cast

T = TypeVar('T')

def print_callcount(f: T) -> T:
    x = 0
    def ret(*args, **kwargs):
        nonlocal x
        x += 1
        print("%d calls so far" % x)
        return f(*args, **kwargs)

    return cast(T, ret)

@print_callcount
def lol(x: int) -> str:
    return "lol"*x

reveal_type(lol)  # E: Revealed type is 'def (x: builtins.int) -> builtins.str'
```

Notes: 
* Mypy trusts that you can call `f`.  You can set a bound on `T` to be `Callable`, but you don't need to for it to typecheck.
* This business doesn't typecheck without the cast. That's a symptom of the fact that mypy doesn't understand the function nature of the argument to `print_callcount`, and there's no way to declare that argument as a `Callable` explicitly without losing the argument type of `lol` later. We'd like to minimize the places where casts are required. Doing so requires something along the lines of "variadic argument variables", discussed below.

# Decorators that take arguments ("second-order"?)

Plenty of decorators "take arguments" by *actually* being functions that return a decorator. For example, we'd like to be able to do this:

```python
[...]

PR #8336, opened 2020-01-26
Improve decorators documentation (add decorator factories)

intgr:
As mentioned in https://github.com/python/mypy/issues/3157#issuecomment-578532179 I found the documentation for decorator usage to be lacking:

* Previously only bare decorators were explained, but decorator
  factories are also fairly common in the real world.
* Also added decorator examples to cheat sheet page.
* Explained difference in behavior for class decorators.
* Shortened (IMO) excessive example for bare decorators.

### DIFF
--- a/docs/source/cheat_sheet.rst
+++ b/docs/source/cheat_sheet.rst
@@ -113,7 +113,6 @@ Functions
        # type: (...) -> bool
        <code>
 
-
 When you're puzzled or when things are complicated
 **************************************************
 
@@ -256,3 +255,22 @@ Miscellaneous
            return sys.stdin
        else:
            return sys.stdout
+
+
+Decorators
+**********
+
+Decorator functions can be expressed via generics. See
+:ref:`declaring-decorators` for the more details.
+
+.. code-block:: python
+
+    from typing import Any, Callable, TypeVar
+
+    F = TypeVar('F', bound=Callable[..., Any])
+
+    def bare_decorator(func):  # type: (F) -> F
+        ...
+
+    def decorator_args(url):  # type: (str) -> Callable[[F], F]
+        ...
--- a/docs/source/cheat_sheet_py3.rst
+++ b/docs/source/cheat_sheet_py3.rst
@@ -127,7 +127,6 @@ Python 3 supports an annotation syntax for function declarations.
    quux(3)  # Fine
    quux(__x=3)  # Error
 
-
 When you're puzzled or when things are complicated
 **************************************************
 
@@ -311,3 +310,22 @@ Miscellaneous
    # class of that name later on in the file
    def f(foo: 'A') -> int:  # Ok
        ...
+
+
+Decorators
+**********
+
+Decorator functions can be expressed via generics. See
+:ref:`declaring-decorators` for the more details.
+
+.. code-block:: python
+
+    from typing import Any, Callable, TypeVar
+
+    F = TypeVar('F', bound=Callable[..., Any])
+
+    def bare_decorator(func: F) -> F:
+        ...
+
+    def decorator_args(url: str) -> Callable[[F], F]:
+        ...
--- a/docs/source/generics.rst
+++ b/docs/source/generics.rst
@@ -522,14 +522,19 @@ Declaring decorators
 
 One common application of type variable upper bounds is in declaring a
 decorator that preserves the signature of the function it decorates,
-regardless of that signature. Here's a complete example:
+regardless of that signature.
+
+Note that class decorators are handled differently than function decorators in
+mypy: decorating a class does not erase its type, even if the decorator has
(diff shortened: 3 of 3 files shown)


docs/source/generics.rst
```diff
@@ -522,14 +522,19 @@ Declaring decorators
 
 One common application of type variable upper bounds is in declaring a
 decorator that preserves the signature of the function it decorates,
-regardless of that signature. Here's a complete example:
+regardless of that signature.
+
+Note that class decorators are handled differently than function decorators in
+Mypy: decorating a class does not erase its type, even if the decorator has
```
JukkaL: Nit: we write "mypy" with a lower case initial 'm' in the middle of a sentence.

docs/source/generics.rst
```diff
@@ -522,14 +522,19 @@ Declaring decorators
...
+
+Here's a complete example of a function decorator:
 
 .. code-block:: python
 
    from typing import Any, Callable, TypeVar, Tuple, cast
 
-   FuncType = Callable[..., Any]
-   F = TypeVar('F', bound=FuncType)
+   F = TypeVar('F', bound=Callable)
```
JukkaL: I'd prefer to use `Callable[..., Any]` to be explicit.

docs/source/generics.rst
```diff
@@ -566,6 +564,59 @@ functions are typically small enough that this is not a big
...
+-------------------
+
+Functions that take arguments and return a decorator (also called second-order decorators), are
+similarly supported via generics:
+
+.. code-block:: python
+
+    from typing import Any, Callable, TypeVar
+
+    F = TypeVar('F', bound=Callable)
```
JukkaL: Again, I prefer `Callable[..., Any]`.

docs/source/generics.rst
```diff
@@ -566,6 +564,59 @@ functions are typically small enough that this is not a big
...
+        return 'Hello world'
+
+Sometimes the same decorator supports both bare calls and calls with arguments. This can be
+achieved by combining with :py:func:`@overload <typing.overload>`:
+
+.. code-block:: python
+
+    from typing import Callable, TypeVar, overload
+
+    F = TypeVar('F', bound=Callable)
```
JukkaL: Similar to above.

docs/source/cheat_sheet.rst
```diff
@@ -113,6 +113,14 @@ Functions
        # type: (...) -> bool
        <code>
 
+   # Decorator functions can be expressed via generics
```
JukkaL: I'd argue that this is too advanced to be included in the cheat sheet. Some of the existing things probably go a bit too deep, to be honest. My reasoning is that we don't even explain type variables in the cheat sheet, so this could be pretty confusing by itself.

docs/source/cheat_sheet_py3.rst
```diff
@@ -127,6 +127,14 @@ Python 3 supports an annotation syntax for function declarations.
    quux(3)  # Fine
    quux(__x=3)  # Error
 
+   # Decorator functions can be expressed via generics
```
JukkaL: Similar to above, maybe this doesn't need to be in the cheat sheet.

docs/source/cheat_sheet.rst
intgr: From my personal experience, I don't think we should worry about explaining anything at all here. Often the first time I'm faced with a problem, I just want a copy-paste-able example that I can bang together and see if it works. That seems exactly the point of the cheat sheet.

The fact that it's nontrivial is more reason to include it IMO.

docs/source/generics.rst
intgr: Like this?

Should I keep the Callable unchanged in the function definition?:
```python
    def atomic(__func: Callable = None, *, savepoint: bool = True):
        def decorator(func: Callable):
```

JukkaL: I'd use `Callable[..., Any]` everywhere instead of bare `Callable`, since it's just a more explicit way of saying the same thing. This is similar how we discourage users from writing `list` instead of `List[Any]`.

docs/source/cheat_sheet.rst
JukkaL: Okay, that's fair. I'd prefer to move this to the end of the cheat sheet, since it's more specialized than the rest of the things in this section. I like to have a narrative, where we go from basic things to less common or more advanced things.

intgr: Sure, that makes sense. I inserted it before "Coroutines and asyncio" because it seemed less advanced than that. But maybe that's just me, I'm open to moving it to the end. :)

intgr: Actually, I reconsidered and moved it to the end.

Issue #5152, opened 2018-06-05
Add function signature hooks for dataclasses functions: replace, asdict, astuple


ilevkivskyi:
Currently, the plugin only supports dataclass creation (i.e. generation of dunder methods with correct types, including `__init__`). However, some functions in the `dataclass` modules have quite broad types in typeshed, for example:
```python
def replace(obj: _T, **changes: Any) -> _T: ...
```
so that arbitrary arguments can be give for `changes`, etc. The plugin should be able to give more precise type for these functions. In particular, `asdict` could return a `TypedDict`.

danr:
How would one go about to fix the type signature for `replace`? In TypeScript it is possible to give a replace function this type:

```typescript
function replace<X extends Record<string, any>>(base: X, changes: Partial<X>): X {
  return {...base, ...changes}
}
```

It's not a 100% correspondence, I'm using TS' `Record` type instead of `dataclass` and just one single `changes` argument instead of a `**changes` keyword argument. But in essence it's the type I want to give `replace` from the `dataclasses` module.

edaqa-uncountable:
This hole is allowing errors to leak through in some of my production code.  Lacking a 'typeof' operator as well, it's difficult to ensure the types are correct.

OlegAlexander:
Hello and thank you for creating mypy!! 

If I may, I'd like to make a case for this issue. `dataclasses.replace()` is equivalent to the [`with` keyword in F#](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/copy-and-update-record-expressions). For anyone wishing to do typed functional programming in Python, having `replace()` checked at compile time is critical. This issue was opened 5 years ago and even the mypy [documentation says:](https://mypy.readthedocs.io/en/stable/additional_features.html#caveats-known-issues)

> Some functions in the [dataclasses](https://docs.python.org/3/library/dataclasses.html#module-dataclasses) module, such as [replace()](https://docs.python.org/3/library/dataclasses.html#dataclasses.replace) and [asdict()](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict), have imprecise (too permissive) types. This will be fixed in future releases.

I can only assume that this isn't a trivial issue to fix. 

It seems I haven't made a strong case for this issue after all--I've only managed to express my frustration 😢. It's just that the next time someone asks me if it's possible to do typed functional programming in Python, I'd love to say "Yes!" instead of "Almost!"

ikonst:
[...]

PR #8339, opened 2020-01-28
DRAFT: Returning TypedDict for dataclasses.asdict

syastrov:
I wanted to implement dataclasses.asdict returning TypedDicts. I ran into issues supporting the recursion that it does. Does anyone have suggestions as to the problems I commented in the tests? Thanks

Edit: Link to comment: https://github.com/python/mypy/pull/8339/files#diff-5b6d315f9321d8b1943195ed39e18f14R1084

Edit: Relates to #5152

### DIFF
--- a/mypy/plugins/common.py
+++ b/mypy/plugins/common.py
@@ -134,23 +134,24 @@ def add_method(
 
 
 def deserialize_and_fixup_type(
-    data: Union[str, JsonDict], api: SemanticAnalyzerPluginInterface
+        data: Union[str, JsonDict], api: SemanticAnalyzerPluginInterface
 ) -> Type:
     typ = deserialize_type(data)
     typ.accept(TypeFixer(api.modules, allow_missing=False))
     return typ
 
 
-def make_typeddict(api: CheckerPluginInterface, fields: 'OrderedDict[str, Type]',
-                   required_keys: Set[str]) -> TypedDictType:
-    # Try to resolve the TypedDict fallback type
-    anonymous_typeddict_type: Optional[Instance] = None
+def get_anonymous_typeddict_type(api: CheckerPluginInterface) -> Instance:
     for type_fullname in TPDICT_FB_NAMES:
         try:
             anonymous_typeddict_type = api.named_generic_type(type_fullname, [])
             if anonymous_typeddict_type is not None:
-                break
+                return anonymous_typeddict_type
         except KeyError:
             continue
-    assert anonymous_typeddict_type is not None
-    return TypedDictType(fields, required_keys=required_keys, fallback=anonymous_typeddict_type)
+    raise RuntimeError("No TypedDict fallback type found")
+
+
+def make_anonymous_typeddict(api: CheckerPluginInterface, fields: 'OrderedDict[str, Type]',
+                             required_keys: Set[str]) -> TypedDictType:
+    return TypedDictType(fields, required_keys=required_keys, fallback=get_anonymous_typeddict_type(api))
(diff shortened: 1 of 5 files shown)


intgr:
Maybe this would be expanding the scope too much, but there have been many times I've wanted something similar. If you could implement this as a generic type, such as `DictOf[SomeClass]`, this opens up potential new use cases, for example:

1. There are many serialization libraries that work on type info besides `asdict`, and could be supported with this. Pydantic and djangorestframework-dataclasses are ones that I'm personally familiar with.
2. For any class `A`, the attribute `A().__dict__` could be inspected to have type `DictOf[A]`, except properties and class variables. Maybe a future PR could implement `NonTotalDicfOf[A, total=False]` (similar to `TypedDict(total=False)`)
3. There are times when you want to handle a subset of an object's fields as a dict. One example is Django's `Model.objects.create_or_update(field=value, defaults={'field2': value})`. The `defaults` argument could be annotated as `NonTotalDictOf[Model]`

Something like [TypeScript's mapped types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types) would be even better but that's rocket science compared to this.

syastrov:
@intgr It's out of scope for this PR, but I have to agree with you and I've also really wanted to make it easier to type these kinds of things. It deserves an issue where it can be discussed and maybe on https://github.com/python/typing/ instead?

The problem I encountered is basically that at runtime `isinstance` succeeds for the newly `asdict`ed type (e.g. if it's a NamedTuple), yet the fields don't have the same types (because the dataclasses were converted to `TypedDict`s) as the original NamedTuple.

JukkaL:
You should be able to support `list` subclasses by using `map_instance_to_supertype`.

For named tuples and tuple subclasses, you can perhaps use the `tuple_type` attribute of `TypeInfo`.

Recursion is currently not possible, since general recursive types aren't supported. One idea would be to detect recursive references and replace them with `Any`, and to perhaps generate an error about recursive types not being supported (which could be ignored using `# type: ignore` if the user is fine with this).

syastrov:
Hi @JukkaL. Thanks for the pointers.
I did manage to support subclasses of list/dict (hopefully correctly) using `map_instance_to_supertype`: so only the TypeVars used by the supertype (list or dict) are transformed according to `asdict` logic.

There is a problem I ran into with this, though. The type vars on the list/dict can be constrained or have variance such that, once the type var is converted from a dataclass to a TypedDict, the resultant type violates the constraints/variance of the original TypeVar. I am not sure how I can get mypy to throw errors in that situation (or if we should just fallback to `Any` instead?). Currently, it just allows it:
https://github.com/python/mypy/pull/8339/commits/0111e3646f174246bebedc979f5785d89ef73c05#diff-5b6d315f9321d8b1943195ed39e18f14R1181

I am still not sure how to support subclasses of `NamedTuple`. It seems that I need to produce a new `TypeInfo` (somehow) which is essentially a totally new subclass of `NamedTuple` where the items are (recursively) transformed according to the `asdict` logic? I am not sure which APIs I could use to achieve this. I would be afraid I would have to duplicate a lot of the code in `semanal_namedtuple.py`.

I also managed to implement detection of recursion, where an error is thrown and it falls back to producing `Dict[str, Any]` for the first detected recursive reference.


TODO before ready for review:
- [ ] Support subclasses of `Tuple` (should be straightforward).
- [ ] Support subclasses of `NamedTuple` (hopefully doable depending on if I can figure out how to do it)
- [ ] Cleanup / type-check / reformat code

syastrov:
Going to make a new PR

PR #8350, opened 2020-01-31
Allow covariance for read-only attributes

### LINKED ISSUE #5836: Decorated method type-incompatibility with superclass
I'm raising a bug.

It's also posted on SO at https://stackoverflow.com/q/52744713/5986907

I get 

    $ mypy file.py
    python.py:34: error: Signature of "fn" incompatible with supertype "B"

for file.py

    from typing import Callable, TypeVar, cast

    T = TypeVar('T')

    def dec(f: Callable[..., T]) -> T:
        def p(self):
            return getattr(self, '_' + f.__name__)

        return cast(T, property(p))

    class X:
        pass

    class Y(X):
        pass

    class A:
        _fn = Y()

    class B(A):
        @dec
        def fn(self) -> X:
            return X()

    class C(B):
        @dec
        def fn(self) -> Y:
            return Y()

I'd expect there to be no error as Y is a subclass of X.

I've tried this with mypy 0.630 and Python 3.5.2, 3.6.6 and 3.7.0. Also with mypy git master for python 3.5.2 and 3.7.0

It's possible these are related but I couldn't quite connect them
https://github.com/python/mypy/issues/220
https://github.com/python/mypy/issues/4430

TH3CHARLie:
resolves #5836

For now, I've not added any tests yet. I've tested against code samples in #5836 and #3260 and the patch is working but adding either of them seems to require modifying a lot in test fixtures

The new helper function `is_writable_attribute` is actually borrowed from `check_no_writable`. I'd also like to refactor the latter function's name and implement it using the new helper in this PR, since I think its name is kind of misleading considering the comments
https://github.com/python/mypy/blob/ea3c65cc377fcb6e810dec9ed6314b25113dcda2/mypy/checker.py#L2391-L2402


JukkaL:
> but adding either of them seems to require modifying a lot in test fixtures

What's missing in text fixtures? Maybe you can simplify the examples to not rely on so many features.

TH3CHARLie:
> > but adding either of them seems to require modifying a lot in test fixtures
> 
> What's missing in text fixtures? Maybe you can simplify the examples to not rely on so many features.

Good point. I've simplified the case from #5836 and added it. I've also done some refactor work to reuse the new helper added and modify the function's name to make it less misleading. But it is not the main focus of the PR and I'd like to know your thoughts on it.

PR #8365, opened 2020-02-04
Analyze descriptor methods as if they're methods, not functions

janverb:
When `__get__` and `__set__` were implicitly called they were analyzed without taking the type of the descriptor into account. That meant that they were only visible to plugins as function calls with unclear
names.


msullivan:
This seems reasonable and I'm happy to merge it. Can a test be written, though?

janverb:
Of course. Thank you for reminding me, I ended up finding an issue with how I did it.

PR #8375, opened 2020-02-05
Make mypy.api.run_dmypy actually capture the output

msullivan:
When the main run API was made threadsafe this broke output capturing
from run_dmypy. I don't really care about the threadsafety of
run_dmypy but I do care about output capture ever working, so I am
restoring the old sys.stdout swapping behavior for run_dmypy. I'd take
a patch to thread stdout through the client if anybody really cares.


msullivan:
Yeah. When I went to do that, I discovered that our wrapper scripts were already *trying* to silence those messages, it just didn't work.

Issue #7994, opened 2019-11-21
Better message if override violates Liskov substitution principle


JukkaL:
Some users don't understand why mypy complains about incompatible overrides. We could add an extra note in cases where the argument type is narrower in an override or the return type is wider (or all incompatible overrides). The note could mention Liskov substitution principle and add a link to mypy documentation (add a new section to https://mypy.readthedocs.io/en/latest/common_issues.html perhaps). We can also mention that `# type: ignore[override]` can be used to silence this error.

#7992 is an example where this came up, but several users have been confused about this.

ilevkivskyi:
Another example is overriding writeable attribute with read-only, see https://github.com/python/mypy/issues/8185. We can probably just add a note to the standard error  in such cases.

rahulrj:
Can i submit an PR for this if no one else is working on this?

JelleZijlstra:
@rahulrj please do, all contributions are appreciated!

SwagatSBhuyan:
Is this issue still open? I see it has been merged as well, so I don't know if I should proceed here.

sanchiagarwal0:
Hi! I would like to work on this issue as my first open-source contribution. I would like to improve the documentation/error message as described in the issue. Please let me know if this is okay.

JukkaL:
Improving the error messages would still be useful. If you want to work on this, one option would be to first propose what the new error messages would look like here in this issue. In particular, it would be useful to have a specific note given only if the return type is wider that the superclass return type, or the opposite for parameter types. This can be implemented in multiple steps (e.g. first add a note specific to wider return types).

sanchiagarwal0:
Thanks! That makes sense. For the first step, I propose adding a specific note when the overridden method's return type is wider than the superclass return type, explaining that the subclass return type must be compatible with the superclass method's return type.

If this approach looks good, I can implement the return-type case first and then handle the parameter-type case separately.

JukkaL:
Starting with the return type case is a good approach, since this makes the change easier to review and iterate on, and a follow-up PR can then apply similar logic to parameter types.

PR #8377, opened 2020-02-06
Add a section for incompatible overrides

rahulrj:
This is according to changes suggested in https://github.com/python/mypy/issues/7994
I will work further to update the code and point to this documentation when mypy throws error.


msullivan:
Is there a simpler example that could be used, without TypedDicts? Ideally using built-in types?

rahulrj:
@msullivan  updated the code to include a simpler example. Would that work?

PR #8384, opened 2020-02-09
mypy: require Python 3.5.3 or greater

hauntsaninja:
See #8380 for motivation, we run into some issue with the version of typing.py in 3.5.1 (same as in https://github.com/dropbox/sqlalchemy-stubs/issues/63). It's also hard to find workarounds, because I can't get a working version of that Python going.

3.5.1 is over four years old, and EOL for 3.5 is in seven months.
Obviously understand if we can't/don't want to merge this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,8 @@ env:
 
 jobs:
   include:
-  # Specifically request 3.5.1 because we need to be compatible with that.
-  - name: "run test suite with python 3.5.1 (compiled with mypyc)"
-    python: 3.5.1
+  - name: "run test suite with python 3.5.3 (compiled with mypyc)"
+    python: 3.5.3
     dist: trusty
     env:
     - TOXENV=py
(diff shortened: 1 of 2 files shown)


hauntsaninja:
We could also a) just change the version that runs CI, in which case users with 3.5.1 or 3.5.2 can continue to run mypy, but would run into issues if they attempted to use stubtest, b) skip stubtest tests in CI for all of 3.5.

msullivan:
We could drop 3.5.1 support but I dropping 3.5.2 support is a nonstarter. We still (unfortunately) run mypy with 3.5.2 in some places. Ubuntu 16.04.6 has 3.5.2. I don't really mind dropping stubtest support though in 3.5.2?

hauntsaninja:
Hmm okay! If 3.5.2 support is important, then changing what we run in CI isn't great (option a, above). Let me try harder to workaround the Python bug (I would have done that first, but I can't seem to build a working version of 3.5.1 or 3.5.2... which leaves test-debugging through Travis as my option).

PR #8387, opened 2020-02-11
mypy: attempt to fix build

hauntsaninja:
virtualenv had a new release today that seems to break various things.


msullivan:
Thanks for looking at this. That is very annoying.

PR #8408, opened 2020-02-15
Point incompatible override error to argument rather than method header

### LINKED ISSUE #8298: Better error location for incompatible override due to argument incompatibility
An incompatible override gets reported on the first line of a method header, even though it would be better to point it at the line with an argument that is incompatible, when possible. Example:

```py
class A:
    def f(self, x: int, y: str, z: float) -> None: pass

class B(A):
    def f(  # Argument 3 incompatible with supertype "A" <-- reported here
            self,
            x: int,
            y: str,
            z: bool,  # <-- would be better to report it here
    ) -> None:
        pass
```

This would improve usability, especially for methods with many arguments on many lines. This would also make it cleaner to ignore these errors, as the location of the `# type: ignore` comment would document which argument is incompatible (if there is one argument per line, which is a common convention).

DenisOH:
Possibly closes #8298.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1643,6 +1643,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                     if not is_subtype(original.arg_types[i],
                                       erase_override(override.arg_types[i])):
                         arg_type_in_super = original.arg_types[i]
+                        arg_node = override.definition.arguments[i+1]
                         self.msg.argument_incompatible_with_supertype(
                             i + 1,
                             name,
@@ -1650,7 +1651,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                             name_in_super,
                             arg_type_in_super,
                             supertype,
-                            node
+                            arg_node
                         )
                         emitted_msg = True
 
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -697,7 +697,9 @@ class ASTConverter:
                 arg_type = TypeConverter(self.errors, line=arg.lineno).visit(annotation)
             else:
                 arg_type = self.translate_type_comment(arg, type_comment)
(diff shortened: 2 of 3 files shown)


DenisOH:
@JukkaL 

While I was working on #8298, I've stumbled upon a different issue: arguments always had the same `line` as the `FuncDef`.

I've tracked it down to this method in `FuncItem`:
https://github.com/python/mypy/blob/83012da090b1b1b8f082663778acb44b2cd7ecb0/mypy/nodes.py#L627-L633

Which gets called while parsing `FuncDef`:
https://github.com/python/mypy/blob/83012da090b1b1b8f082663778acb44b2cd7ecb0/mypy/fastparse.py#L621

This was basically always overwriting `arg`'s line.

However, removing 
https://github.com/python/mypy/blob/83012da090b1b1b8f082663778acb44b2cd7ecb0/mypy/nodes.py#L632-L633
didn't completely solve the problem. At this point, the argument's `line` was always `-1`.
I've modified `make_argument` (in fastparse.py) by creating the `argument` variable and calling `set_line` on it before returning it.

mypy/checker.py
```diff
@@ -1643,14 +1643,15 @@ def erase_override(t: Type) -> Type:
                     if not is_subtype(original.arg_types[i],
                                       erase_override(override.arg_types[i])):
                         arg_type_in_super = original.arg_types[i]
+                        arg_node = override.definition.arguments[i+1]
```
ethanhs: This doesn't seem quite right. It seems that you are passing this to forward the correct context, but 1) you are passing the next argument as context? 2) arguments aren't contexts. You probably want to make sure the context is set correctly to begin with.

97littleleaf11:
Thanks for your PR! But I'm going to close it due to inactive status. Feel free to reopen it if you still work on this!

PR #8419, opened 2020-02-20
[mypyc] Refactor: extract low-level genops to a new module

msullivan:
This is the beginning of providing a lower-level IRBuilder
abstraction. Extract most of the IRBuilder methods that operate only
on IR (and not on the mypy AST) into a LowLevelIRBuilder class in a
new module.

For now, IRBuilder subclasses from LowLevelIRBuilder, but we'll want
to change that and make IRBuilder have a more standalone interface.


JukkaL:
Thanks, this is a nice code quality improvement!

PR #8425, opened 2020-02-21
Use fully qualified names for ambiguous class names resembling builtins.

### LINKED ISSUE #8372: Use fully qualified name for class with confusing name such as "Any"
The following example generates a confusing error message, since it's unclear which `Any` type it is referring to:

```py
class Any: pass

x = Any()
x = 1  # Incompatible types in assignment 
       # (expression has type "int", variable has type "Any")
```

(It would be more confusing if the `Any` type is inferred from the return type of a function.)

It would be better to always use a fully qualified name for a class named `Any` in error messages (similarly also for `None`, `Union`, etc.). 

This can also happen with classes that have names resembling built-in types:

```py
class float: pass

x = float()
x = 0  # Incompatible types in assignment 
       # (expression has type "int", variable has type "float")
```

The relevant code seems to be in `mypy.messages.format_type`.

Muks14x:
Resolves #8372.

Is there a better way to get the types from `typing` and `builtins`?

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -13,7 +13,6 @@ from collections import OrderedDict
 import re
 import difflib
 from textwrap import dedent
-import builtins
 
 from typing import cast, List, Dict, Any, Sequence, Iterable, Tuple, Set, Optional, Union
 from typing_extensions import Final
@@ -1733,13 +1732,10 @@ def find_type_overlaps(*types: Type) -> Set[str]:
     with their fullname.
     """
     d = {}  # type: Dict[str, Set[str]]
-    builtin_types = set(dir(builtins))
     for type in types:
         for inst in collect_all_instances(type):
             d.setdefault(inst.type.name, set()).add(inst.type.fullname)
     for shortname in d.keys():
-        if shortname in builtin_types:
-            d[shortname].add('builtins.{}'.format(shortname))
         if 'typing.{}'.format(shortname) in TYPES_FOR_UNIMPORTED_HINTS:
             d[shortname].add('typing.{}'.format(shortname))
 
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -13,7 +13,9 @@ from collections import OrderedDict
 import re
 import difflib
 from textwrap import dedent
+import builtins
 
+import typing
 from typing import cast, List, Dict, Any, Sequence, Iterable, Tuple, Set, Optional, Union
 from typing_extensions import Final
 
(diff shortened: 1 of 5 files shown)


JelleZijlstra:
> Is there a better way to get the types from typing and builtins?

Probably we should get them from the stubs for the appropriate Python version mypy is targeting, but I don't know off hand how to do that.

Muks14x:
@msullivan Thanks for the suggestions! I've fixed older outputs that needed changing, I'll try and add a couple of new test cases too.

> Probably we should get them from the stubs for the appropriate Python version mypy is targeting, but I don't know off hand how to do that.

@JelleZijlstra Thank you! I don't know how to load version-specific stubs though, could you tell me where to start looking?

Muks14x:
> To figure out what names to worry about, I'd just use the list from TYPES_FOR_UNIMPORTED_HINTS in `semanal.py` (which can be moved out of it).

Oh, sorry, I overlooked this. I'll look for the names there, thanks!

Muks14x:
I've added a test case, but I think I might've added it to the wrong place. Does it belong somewhere else?

mypy/messages.py
```diff
@@ -1717,9 +1733,15 @@ def find_type_overlaps(*types: Type) -> Set[str]:
     with their fullname.
     """
     d = {}  # type: Dict[str, Set[str]]
+    builtin_types = set(dir(builtins))
     for type in types:
         for inst in collect_all_instances(type):
             d.setdefault(inst.type.name, set()).add(inst.type.fullname)
+    for shortname in d.keys():
+        if shortname in builtin_types:
```
msullivan: I would drop the `builtins_types` part of this

msullivan:
Test is in the right place!

Muks14x:
I've removed the overlap-check for builtins. Is there a better way to do the check? I guess dir(builtins) is hacky and could cause incorrect messages if the target python version is very different.

Thanks for the help!

test-data/unit/check-basic.test
```diff
@@ -75,6 +75,14 @@ if int():
     x = 1
 
 
+[case testIncompatibleAssignmentAmbiguousShortnames]
+
+class Any: pass
+
+x = Any()
+x = 1  # E: Incompatible types in assignment (expression has type "int", variable has type "__main__.Any")
```
JukkaL: It would be good to test a few other type names as well, such as `list` / `List`. (Lower-case `list` might not work, since it's not in the `TYPES_FOR_UNIMPORTED_HINTS` set.)

Muks14x: I've added a few more types from `TYPES_FOR_UNIMPORTED_HINTS`.

PR #8426, opened 2020-02-22
Make IRBuilder not inherit from LowLevelIRBuilder.

msullivan:
Instead we make IRBuilder do all of its building through a
LowLevelIRBuilder. A new LowLevelIRBuilder is created for each
function that is compiled (which lets LowLevelIRBuilder lose its stack
of block lists and environments and only has one of each).

A collection of the most commonly used methods in LowLevelIRBuilder
are given passthrough methods in IRBuilder for convenience.

Work on mypyc/mypyc#714.

### DIFF
--- a/mypyc/genfunc.py
+++ b/mypyc/genfunc.py
@@ -117,7 +117,7 @@ class BuildFuncIR:
         if expr.expr:
             retval = self.builder.accept(expr.expr)
         else:
-            retval = self.builder.none()
+            retval = self.builder.builder.none()
         return self.emit_yield(retval, expr.line)
 
     def visit_yield_from_expr(self, o: YieldFromExpr) -> Value:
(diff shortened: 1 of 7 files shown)


TH3CHARLie:
Out of curiosity, I am working on mypyc/mypyc#709 should I wait the ongoing work of mypyc/mypyc#714 be completed since I  believe they have pretty much overlap

JukkaL:
@TH3CHARLie It makes sense to wait until the most disruptive refactorings (step 6) are finished.

Also, I'd strongly suggest that you split your work into many smaller PRs and create PRs as soon as you've made some progress. Work-in-progress PRs are fine if you mark them as using WIP in the title. If the small PRs are merged as they are ready, you will likely spend much less time dealing with merge conflicts, and you'll get feedback sooner rather than later. Early feedback may save a lots of time (both your time and the reviewers' time).

TH3CHARLie:
Thanks for the response. I'll keep track of the refactoring. Splitting large PRs into smaller ones is very reasonable to me, I'll keep that in mind.

PR #8427, opened 2020-02-22
Added Note in README

kmrakash:
Many new contributors like me do these mistakes. So I added a note to remind them


hauntsaninja:
Thanks, but I don't think this is really an improvement. Refer to https://devguide.python.org/documenting/#economy-of-expression and https://devguide.python.org/documenting/#audience

PR #8436, opened 2020-02-24
Update semanal_enum.py

Arshaan-256:
Relaxed the checks for IntEnum arguments.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3423,9 +3423,6 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         """Type check a super expression (non-lvalue)."""
 
         # We have an expression like super(T, var).member
-
-        print(SuperExpr)
-
         # First compute the types of T and var
         types = self._super_arg_types(e)
         if isinstance(types, tuple):
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3424,6 +3424,8 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
         # We have an expression like super(T, var).member
 
+        print(SuperExpr)
+
         # First compute the types of T and var
         types = self._super_arg_types(e)
         if isinstance(types, tuple):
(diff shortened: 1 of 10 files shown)


Michael0x2a:
Thanks for the PR!

It looks like some of the tests are broken -- can you fix them and add a few more? You can find more information on how to run and write tests at https://github.com/python/mypy#tests.

mypy/semanal_enum.py
```diff
@@ -107,7 +107,8 @@ def parse_enum_call_args(self, call: CallExpr,
         if len(args) < 2:
             return self.fail_enum_call_arg("Too few arguments for %s()" % class_name, call)
         if len(args) > 2:
-            return self.fail_enum_call_arg("Too many arguments for %s()" % class_name, call)
+            if call.arg_kinds != [ARG_POS, ARG_POS, ARG_POS]:
```
JelleZijlstra: This doesn't make much sense in the light of the next `if` condition: if this one passes the next one (on line 112) will still fail.

Arshaan-256: @JelleZijlstra Thanks for your review. Could you tell me what does `ARG_POS` mean?

mypy/semanal_enum.py
```diff
@@ -107,9 +107,11 @@ def parse_enum_call_args(self, call: CallExpr,
         if len(args) < 2:
             return self.fail_enum_call_arg("Too few arguments for %s()" % class_name, call)
         if len(args) > 2:
-            return self.fail_enum_call_arg("Too many arguments for %s()" % class_name, call)
-        if call.arg_kinds != [ARG_POS, ARG_POS]:
-            return self.fail_enum_call_arg("Unexpected arguments to %s()" % class_name, call)
+            if call.arg_kinds != [ARG_POS, ARG_POS, ARG_NAMED_OPT]:
```
Michael0x2a: Enum can accept up to [four named arguments](https://docs.python.org/3/library/enum.html#functional-api), not just one.

You should also try checking `call.arg_names` to make sure the arguments are what we expect.

Arshaan-256: @Michael0x2a Sure! Will do it.

mypy/checkexpr.py
```diff
@@ -3424,6 +3424,8 @@ def visit_super_expr(self, e: SuperExpr) -> Type:
 
         # We have an expression like super(T, var).member
 
+        print(SuperExpr)
```
ethanhs: I think you left this debug print in on accident.
```suggestion

```

Issue #5491, opened 2018-08-16
Clean up kludges for handling metaclasses and Type[...] in check_op_reversible


Michael0x2a:
The [current implementation for checking reversible operators][0] contains two kludges that seem to be related to operators, metaclasses, and Type[...]

1. This call to [has_member][1] feels like it ought to be unnecessary/redundant due to the following `analyze_member_access` -- but removing it causes pythoneval's [OpMetaclassAccess](https://github.com/python/mypy/blob/master/test-data/unit/pythoneval.test#L1159) test case to fail.

    (Note: if you remove just that call, you also need to add in a `elif isinstance(member, AnyTyp): return None` after calling `analyze_member_access`)

2. If the reversible operators logic was unable to find a matching operator, it just [tries calling `check_call` on the name directly][2] to get back some generic error message. It seems as if this call ought to always return an error message, but sometimes it surprisingly will not. E.g. see the [TypeTypeComparisonWorks](https://github.com/python/mypy/blob/master/test-data/unit/check-classes.test#L3251) tests in check-classes -- adding an `assert local_errors.is_errors()` causes that test case to crash.

    This feels related to the first TODO in some way -- either we're not handling operators + metaclasses + Type directly, or there are some weird shenanigans happening due to `Any`...

For additional context, see https://github.com/python/mypy/issues/5136 and https://github.com/python/mypy/pull/1806.

  [0]:https://github.com/python/mypy/blob/master/mypy/checkexpr.py#L1848
  [1]: https://github.com/python/mypy/blob/master/mypy/checkexpr.py#L1879
  [2]: https://github.com/python/mypy/blob/master/mypy/checkexpr.py#L2020

elazarg:
I should work on this, if I will find the time.

Michael0x2a:
@elazarg -- sounds good to me! 

(I added myself mostly out of a sense of obligation, since the TODOs were a fallout of my operator refactors. But I also don't know when I'll find the time to do this, so feel free to tackle this whenever.)

elazarg:
@Michael0x2a - sorry, I can't seem to find the time to do that.

Michael0x2a:
@elazarg -- no worries, I'm also in the same boat.

PR #8438, opened 2020-02-25
mypy: remove has_member

hauntsaninja:
In particular:
- The test case mentioned in the code passes without it
- The test case changed seems to have more desirable behaviour now,
  consider:

```
from typing import Any

"""
class C:
    def __radd__(self, other) -> float:
        return 1.234
"""
C: Any
class D(C):
    pass

reveal_type("str" + D())
```


Michael0x2a:
If all the tests are now passing/the metaclass thing no longer seems to be an issue, I agree merging this seems like a good idea.

I do remember filing an issue related to this in https://github.com/python/mypy/issues/5491 -- it could be worth double-checking to see if we also still need to keep that open.

PR #8444, opened 2020-02-26
Update test_commandline.py

dubesar:
1. Made changes to test_commandline.py to make it faster using regex
2. Used iter_replace method of regex to make the replace faster.

### DIFF
--- a/mypyc/test/test_commandline.py
+++ b/mypyc/test/test_commandline.py
@@ -30,6 +30,15 @@ class TestCommandLine(MypycDataSuite):
     files = files
     base_path = test_temp_dir
     optional_out = True
+    
+    def finditer_replace(self, text, reg, rep) -> str:
+        cursor_pos = 0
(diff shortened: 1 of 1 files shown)


dubesar:
@JukkaL please review the pull request and I am not able to understand why is it failing CI travis.

JelleZijlstra:
Why would this be faster? I haven't timed it but I'm pretty sure calling `.replace` directly is going to be faster than a bunch of hand-rolled Python code and regex execution.

dubesar:
@JelleZijlstra  Sorry I didn't time it but it approximately the same and doesn't have any much effect. 

For above regex is  0.0001814s
and .replace is 0.000141s

JelleZijlstra:
I timed it too (using `%timeit` in IPython, which runs the code many times to get more reliable numbers) and found that the new code was more than 10x slower.

Issue #8453, opened 2020-02-28
Filenames with dots


tallesl:
Calling `mypy .` finds and checks successfully the files `foobar.py`, `foo-bar.py` and `foo_bar.py`, but it doesn't look up `foo.bar.py`.

Is this a bug?

At first I thought that it could be interpreting `.bar.py` as its extension, but `os.path.splitext('foo.bar.py')` gives `.py`. Tested with bash, python 3.7.6 and mypy 0761.

ethanhs:
Huh, I seem unable to reproduce this. What OS are you on?

hauntsaninja:
Repros on MacOS for me:
```
~/delete λ mkdir test      
~/delete λ echo 'x = 1; x += "asdf"' > test/test.lol.py
~/delete λ mypy test       
There are no .py[i] files in directory 'test'
```

ethanhs:
Hm, thinking about this more, I'm not actually sure that this should be supported? Names with `.` in their name aren't valid modules even to CPython (if you try running them with -m). And you can't import them without using `importlib`.

@tallesl, what is your usecase?

pr4k:
As far as i remember , its also mentioned in the official pep guidelines that using periods should be avoided in the names of module,but the question is if someone is just developing a standalone script or maybe something which is not developed as a module to be imported, should we allow it for that?
@ethanhs @tallesl @hauntsaninja

ethanhs:
If they are developing a standalone script they should run `mypy script.dot.py`, which works in my testing.

pr4k:
Yeah valid point! So should we then raise a error if someone tries to use period in names for a project, because right now what i have observed is that mypy ignores the file with `dot` in names when we use `mypy .`, So a response stating the ignored file could be useful i think

tallesl:
I have a repo with configuration files by environment, like "
configuration.production.py". My deployment script then picks the desired
one and yields "configuration.py". I know python has INI support, but
they're much more painful than using pure python (but that's off-topic
anyway).

Considering that is not a valid module name, that's fair that it doesn't
find the file.

msullivan:
An error might be a little too aggressive in some situations, though. I think maybe we say this isn't a bug?

ethanhs:
Yeah I think it's fine to close as won'tfix.

smheidrich:
[...]

PR #8456, opened 2020-02-28
Added Fix: #8453

pr4k:
Added Fix for Issue #8453

### DIFF
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -92,7 +92,7 @@ class SourceFinder:
                 base, suffix = os.path.splitext(name)
                 if base == '__init__':
                     continue
-                if base not in seen and '.' not in base and suffix in PY_EXTENSIONS:
+                if base not in seen and suffix in PY_EXTENSIONS:
                     seen.add(base)
                     src = BuildSource(path, mod_prefix + base, None, base_dir)
                     sources.append(src)


mypy/find_sources.py
```diff
@@ -92,7 +92,7 @@ def expand_dir(self, arg: str, mod_prefix: str = '') -> List[BuildSource]:
                 base, suffix = os.path.splitext(name)
                 if base == '__init__':
                     continue
-                if base not in seen and '.' not in base and suffix in PY_EXTENSIONS:
+                if base not in seen and suffix in PY_EXTENSIONS:
                     seen.add(base)
                     src = BuildSource(path, mod_prefix + base, None, base_dir)
```
hauntsaninja: Passing in `mod_prefix + base` to `BuildSource` here is probably incorrect when there's a period in the `base` name. It might make sense to pass in `None`. Someone who knows build better would know for sure :-)

hauntsaninja: Although when you explicitly list the file we pass that in as module name to `BuildSource` even if it contains a period, so maybe it's fine: https://github.com/python/mypy/blob/master/mypy/find_sources.py#L34

pr4k: @hauntsaninja Thanks for pointing this out, I think you are right, as I think it is also mentioned in pep guideline that using period is not allowed in file names, but then i tried it on the tests and it was working, so I am not sure how the build is configured. So yes someone with better experience with build would be knowing.

msullivan:
I'm going to close this because I think that either this is not a bug or the resolution should be emitting an error?

PR #8458, opened 2020-02-28
Support narrowing of walrus in most cases

### LINKED ISSUE #8447: PEP 572 Walrus Operator Regex Assignment in Condition
Mypy version: 0.761
Python version: 3.8.1

 #6899 implements `PEP 572`, or at least parts of it, as mentioned in the PR assignments can occur anywhere. Fails with an example similar to [here](https://www.python.org/dev/peps/pep-0572/#id9) with `regex`.

Minimal Example:
```python
import re


TXT = "b"
if (a := re.search(r"a", TXT)) is not None:
    string = a.string
    print(f"{string}")
else:
    print("not in there") #=> not in there
```
mypy message:
```sh
~ λ mypy test.py
test.py:8: error: Item "None" of "Optional[Match[str]]" has no attribute "string"
```
but the type of `a` is not `Optional[Match[str]]` it's `Match[str]` as a result of the if check.

msullivan:
It is a pretty simple matter of pulling out the assignment target from
the walrus. We don't bother handling things like `x := (y := z)` since
I can't imagine they are common enough to be worth bothering but we
could in the future if anyone cares.

Fixes #8447.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3923,7 +3923,9 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             return {}, None
         elif is_false_literal(node):
             return None, {}
-        elif isinstance(node, CallExpr) and len(node.args) > 0:
+        elif isinstance(node, CallExpr):
+            if len(node.args) == 0:
+                return {}, {}
             expr = collapse_walrus(node.args[0])
             if refers_to_fullname(node.callee, 'builtins.isinstance'):
                 if len(node.args) != 2:  # the error will be reported elsewhere
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3923,11 +3923,11 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             return {}, None
         elif is_false_literal(node):
             return None, {}
-        elif isinstance(node, CallExpr):
+        elif isinstance(node, CallExpr) and len(node.args) > 0:
+            expr = collapse_walrus(node.args[0])
             if refers_to_fullname(node.callee, 'builtins.isinstance'):
                 if len(node.args) != 2:  # the error will be reported elsewhere
                     return {}, {}
-                expr = node.args[0]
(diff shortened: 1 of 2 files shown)


mypy/checker.py
```diff
@@ -3923,11 +3923,11 @@ def find_isinstance_check_helper(self, node: Expression) -> Tuple[TypeMap, TypeM
             return {}, None
         elif is_false_literal(node):
             return None, {}
-        elif isinstance(node, CallExpr):
+        elif isinstance(node, CallExpr) and len(node.args) > 0:
```
Michael0x2a: Maybe instead add a separate 'if' statement under here that does `if len(node.args) == 0: return {}, {}`? 

So that way, we don't have to do a bunch of extra `isinstance(...)` checks if the node happens to be an unrelated CallExpr.

test-data/unit/check-python38.test
```diff
@@ -288,10 +288,23 @@ def check_partial() -> None:
...
     reveal_type(x)  # N: Revealed type is 'Union[builtins.int, None]'
 
-def check_narrow(x: Optional[int]) -> None:
+def check_narrow(x: Optional[int], s: List[int]) -> None:
     if (y := x):
         reveal_type(y)  # N: Revealed type is 'builtins.int'
-[builtins fixtures/f_string.pyi]
+
+    if (y := x) is not None:
+        reveal_type(y)  # N: Revealed type is 'builtins.int'
```
Michael0x2a: Will 'x' also end up being narrowed here?

If it doesn't, I think it's probably fine for now, but we should probably file a follow-up task.

msullivan: No, but I think that's probably fine? I don't think it's actually idiomatic or makes sense to be using a walrus operator when the right hand side is a value.

Michael0x2a: Hmm, that's true. I guess we can just not worry about it unless somebody brings it up.

PR #8460, opened 2020-02-29
[mypyc]add str.__getitem__ primitive op

TH3CHARLie:
PR related to mypyc/mypyc#644

Implement a specialized `str.__getitem__` (which is often used as `str[n]`) op instead of relying on the original general `PyObject_GetItem` solution.

For testing, compile the code below:
```python
def foo(s: str, index: int) -> str:
    return s[index]
```
produces
```
>>>import foo from test
>>> foo("abc", 0)
'a'
>>> foo("abc", 1)
'b'
>>> foo("abc", 2)
'c'
>>> foo("abc", -1)
'c'
>>> foo("abc", -2)
'b'
>>> foo("abc", -3)
'a'
>>> foo("abc", -4)
  File "<stdin>", line 1, in <module>
  File "test.py", line 2, in foo
    return s[index]
IndexError: string index out of range
>>> foo("abc", 4)
  File "<stdin>", line 1, in <module>
  File "test.py", line 2, in foo
    return s[index]
IndexError: string index out of range
>>> 
```
I'd like to add it into the test suite but there's a CompileError `Value of type "str" is not indexable`. I'm not sure why it happens but I suspect it is a fixture issue.

Please review this PR, thanks! @msullivan


TH3CHARLie:
- [ ] resolve conflicts once mypyc/mypyc#716 completes

JukkaL:
Sorry, my refactorings caused some conflicts. They should be easy to resolve once I've finished the big refactorings.

mypyc/lib-rt/CPy.h
```diff
@@ -930,6 +930,42 @@ static PyObject *CPyDict_FromAny(PyObject *obj) {
...
+            if (n < 0)
+                n += size;
+            enum PyUnicode_Kind kind = (enum PyUnicode_Kind)PyUnicode_KIND(str);
+            void *data = PyUnicode_DATA(str);
+            Py_UCS4 ch = PyUnicode_READ(kind, data, n);
+            PyObject *unicode = PyUnicode_New(1, ch);
+            if (unicode == NULL)
+                return NULL;
+
+            assert(PyUnicode_KIND(unicode) != PyUnicode_1BYTE_KIND);
```
msullivan: This seems wrong to me. Why can't it be 1BYTE? Looking at the code in `unicodeobject.c` it seems like it should be able to be?

TH3CHARLie:
> Also this is enough code, and fiddly enough code, that I think it is worth making sure that there is actually a performance improvement. Could you try out some simple microbenchmark that verifies this can help?

Agreed. I will do some benchmark and try to come out with some numbers/figures to show whether the change has helped.

TH3CHARLie:
> Also this is enough code, and fiddly enough code, that I think it is worth making sure that there is actually a performance improvement. Could you try out some simple microbenchmark that verifies this can help?

Agreed. I will do some benchmark and try to come out with some numbers/figures to show whether the change has helped.

TH3CHARLie:
I did some benchmark that randomly accessing the string using `str[n]` syntax. I've run two versions of compiled code using the mypyc command and the first one is this PR's version and the second is the original one.

Here's a figure showing the number of accesses/total runtime:

https://photos.app.goo.gl/UmSwsNXyUoKKoCJX8

Generally, the updated version helps, but only notable when the number of accesses is large enough.

Can we do better? For example, since we know this is a str-primitive, can we ignore the `PyUnicode_Check` calls?

msullivan:
I think we could drop the `PyUnicode_Check`, yeah.
Could you share your benchmark code?

TH3CHARLie:
> Could you share your benchmark code?

Sure, that's how I do it. First, I write a wrapper function
```python
def foo(s:str, index: int) -> str:
    return s[index]
```
Then compile it using the new/original mypyc into `new.*.so` and `original.*.so`. Finally, use the following script:
```python
from new import foo as nf
from original import foo as of
from random import randint
from time import perf_counter
import matplotlib.pyplot as plt
import numpy as np

ACCESS_LOWER_BOUND = 100
ACCESS_UPPER_BOUND = 50000
ACCESS_STEP = 100
ITERATIONS = 10

def run(s):
    nf_times = []
    of_times = []
    length = len(s)
    for accesses in range(ACCESS_LOWER_BOUND, ACCESS_UPPER_BOUND, ACCESS_STEP):
        # new version
        nf_start = perf_counter()
        for it in range(ITERATIONS):
            for i in range(accesses):
                idx = randint(0, length - 1)
                x = nf(s, idx)
        nf_end = perf_counter()

        # original version
        of_start = perf_counter()
        for it in range(ITERATIONS):
            for i in range(accesses):
                idx = randint(0, length - 1)
                x = of(s, idx)
        of_end = perf_counter()

        nf_times.append(nf_end - nf_start)
        of_times.append(of_end - of_start)
        print(f"access {accesses} completed")

    return nf_times, of_times

def main():
    s = "a" * 50000
    x = np.arange(ACCESS_LOWER_BOUND, ACCESS_UPPER_BOUND, ACCESS_STEP)
    y1, y2 = run(s)

    plt.plot(x, y1)
    plt.plot(x, y2)
    plt.legend(['new', 'original'], loc='upper left')
    plt.show()

if __name__ == "__main__":
    main()
```

I've removed the `PyUnicode_Check` but observe no obvious improvement. I am also uncertain about the benchmark that it may be biased or incorrect.

msullivan:
Oh, aha. The code under test here is going to be dwarfed by everything else:
 * randint
 * globals lookup to do the call
 * parsing the arguments out of the argument tuple in the wrapper

I tried the following quick-and-dirty benchmark and got a substantial speedup:
```
import time

def foo(s: str) -> None:
    t0 = time.time()
    for i in range(100000):
        for j in range(len(s)):
            x = s[j]
    t1 = time.time()
    print("{:.3f}".format(t1-t0))
```

In general, for microbenchmarks, you want to do as little as possible in the timed portion. So ideally you'll just have some loops and the code under test.

TH3CHARLie:
Aha now I see, thanks!

So you confirm that this PR is helping, right?

msullivan:
Yes

TH3CHARLie:
Thanks!

Issue #8166, opened 2019-12-17
Recognize system.platform() as well as sys.platform is


brettcannon:
Note: if you are reporting a wrong signature of a function or a class in
the standard library, then the typeshed tracker is better suited
for this report: https://github.com/python/typeshed/issues

Please provide more information to help us understand the issue:

* Are you reporting a bug, or opening a feature request? feature request
* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.
```python
    if platform.system() != "Windows":  # sys.platform != "win32":
        try:
            # os.confstr("CS_GNU_LIBC_VERSION") returns a string like "glibc 2.17".
            version_string = os.confstr("CS_GNU_LIBC_VERSION")
            assert version_string is not None
            _, version = version_string.split()
        except (AssertionError, OSError, ValueError):
            # os.confstr() or CS_GNU_LIBC_VERSION not available (or a bad value)...
            return None
        return version
    return None
```
* What is the actual behavior/output? "error: Module has no attribute "confstr""
* What is the behavior/output you expect? No error when run under Windows
* What are the versions of mypy and Python you are using? mypy 0.750, Python 3.7.4

Basically it would be nice to get to use `platform.system()` instead of having to use `sys.platform` to make mypy happy with code that is conditional on the platform.

ilevkivskyi:
On one hand this is pretty niche, but on other hand should be not hard to implement.

neesara:
Hi @ilevkivskyi I am interested in working on this issue, please guide me on where I can start.

ilevkivskyi:
Look at `consider_sys_platform()` in `mypy/reachability.py`.

neesara:
Which option would you recommend? - Write a new function similar to 'consider_sys_platform()' for handling platform.system() or modify 'consider_sys_platform()'  to accomodate platform.system()

joybhallaa:
@ilevkivskyi , I was reading the function `consider_sys_platform`, and from what I could understand, we want to check Comaprision Expression, and call expressions using the `platform.system()` module.
In addition to the existing `sys.platform()`. Am I on the right track?

ilevkivskyi:
Yes.

joybhallaa:
@ilevkivskyi  I was going through the code, and I think I should:
1. Change function name from `consider_sys_platform` to `consider_platform` 
[...]

PR #8461, opened 2020-02-29
Added supprot for platform.system()

joybhallaa:
**What is this PR about ?**
This PR is related to #8166 . 
@brettcannon asked to include support for `platform.system()` with the existing `sys.platform`

**Things I've Changed**
1. Renamed `platform` to `sys_platform` to avoid confusion
1. I've added the `platform = platform.system()` in **options.py**
2. Added param `platform` in `consider_sys_platform` function

**Tests**
   I've checked the changes with `pep8(pycodestyle)` 

I think changing the `platform` to `sys_platform` may cause some errors if it is being used somewhere else.

### DIFF
--- a/mypy/options.py
+++ b/mypy/options.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 import re
 import pprint
 import sys
+import platform
 
 from typing_extensions import Final
 from typing import Dict, List, Mapping, Optional, Pattern, Set, Tuple, Callable, Any
@@ -70,7 +71,8 @@ class Options:
         # The executable used to search for PEP 561 packages. If this is None,
         # then mypy does not search for PEP 561 packages.
         self.python_executable = sys.executable  # type: Optional[str]
-        self.platform = sys.platform
+        self.sys_platform = sys.platform
+        self.platform = platform.system()
         self.custom_typing_module = None  # type: Optional[str]
         self.custom_typeshed_dir = None  # type: Optional[str]
         self.mypy_path = []  # type: List[str]
--- a/mypy/reachability.py
+++ b/mypy/reachability.py
@@ -1,5 +1,5 @@
 """Utilities related to determining the reachability of code (in semantic analysis)."""
-
+import platform
 from typing import Tuple, TypeVar, Union, Optional
 from typing_extensions import Final
 
@@ -92,7 +92,7 @@ def infer_condition_value(expr: Expression, options: Options) -> int:
     else:
         result = consider_sys_version_info(expr, pyversion)
         if result == TRUTH_VALUE_UNKNOWN:
-            result = consider_sys_platform(expr, options.platform)
+            result = consider_sys_platform(expr, options.sys_platform)
     if result == TRUTH_VALUE_UNKNOWN:
         if name == 'PY2':
             result = ALWAYS_TRUE if pyversion[0] == 2 else ALWAYS_FALSE
@@ -150,7 +150,7 @@ def consider_sys_version_info(expr: Expression, pyversion: Tuple[int, ...]) -> i
     return TRUTH_VALUE_UNKNOWN
 
 
-def consider_sys_platform(expr: Expression, platform: str) -> int:
+def consider_sys_platform(expr: Expression, sys_platform: str, platform: str) -> int:
     """Consider whether expr is a comparison involving sys.platform.
 
     Return ALWAYS_TRUE, ALWAYS_FALSE, or TRUTH_VALUE_UNKNOWN.
@@ -166,7 +166,7 @@ def consider_sys_platform(expr: Expression, platform: str) -> int:
         op = expr.operators[0]
         if op not in ('==', '!='):
             return TRUTH_VALUE_UNKNOWN
-        if not is_sys_attr(expr.operands[0], 'platform'):
+        if not is_sys_attr(expr.operands[0], 'sys_platform', 'platform'):
             return TRUTH_VALUE_UNKNOWN
         right = expr.operands[1]
         if not isinstance(right, (StrExpr, UnicodeExpr)):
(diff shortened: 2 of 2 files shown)


mypy/options.py
```diff
@@ -71,6 +72,7 @@ def __init__(self) -> None:
         # then mypy does not search for PEP 561 packages.
         self.python_executable = sys.executable  # type: Optional[str]
         self.platform = sys.platform
```
ethanhs: It seems you didn't rename this to be consistent with your usage below. You should also change any usage of this elsewhere.
```suggestion
        self.sys_platform = sys.platform
```

mypy/reachability.py
```diff
@@ -92,7 +92,7 @@ def infer_condition_value(expr: Expression, options: Options) -> int:
     else:
         result = consider_sys_version_info(expr, pyversion)
         if result == TRUTH_VALUE_UNKNOWN:
-            result = consider_sys_platform(expr, options.platform)
+            result = consider_sys_platform(expr, options.sys_platform)
```
ethanhs: You seem not to be passing `options.platform_system` here, but you added it as an argument below.

mypy/reachability.py
```diff
@@ -150,8 +150,8 @@ def consider_sys_version_info(expr: Expression, pyversion: Tuple[int, ...]) -> i
     return TRUTH_VALUE_UNKNOWN
 
 
-def consider_sys_platform(expr: Expression, platform: str) -> int:
-    """Consider whether expr is a comparison involving sys.platform.
+def consider_sys_platform(expr: Expression, platform: str, platform_system: str) -> int:
+    """Consider whether expr is a comparison involving sys.platform
```
ethanhs: Please update the docstring to add that this also considers `platform.system` (see below).

mypy/reachability.py
```diff
@@ -261,6 +266,16 @@ def is_sys_attr(expr: Expression, name: str) -> bool:
     return False
 
 
+def is_platform_attr(expr: Expression, name: str) -> bool:
+    # Platform lib calls methods, it differs from sys lib
+    # sys.platform is of str class and platform.system is of
+    # function class
+    if isinstance(expr, MemberExpr) and expr.name == name:
+        if isinstance(expr.expr, CallExpr) and expr.expr.name == 'system':
```
ethanhs: Shouldn't this be `if isinstance(expr.expr, CallExpr) and expr.expr.name == 'platform'`?

mypy/reachability.py
```diff
@@ -166,9 +166,14 @@ def consider_sys_platform(expr: Expression, platform: str) -> int:
...
-            return TRUTH_VALUE_UNKNOWN
-        right = expr.operands[1]
+        #check if either platform or platform_system is being used
+        if platform_system:
+            if not is_platform_attr(expr.operands[0], 'plaform_system'):
+                return TRUTH_VALUE_UNKNOWN
+        elif platform:
+            if not is_sys_attr(expr.operands[0], 'platform'):
+                return TRUTH_VALUE_UNKNOWN
+            right = expr.operands[1]
```
ethanhs: This seems over-indented, it gets used below.

mypy/reachability.py
```diff
@@ -1,5 +1,5 @@
 """Utilities related to determining the reachability of code (in semantic analysis)."""
-
+import platform
```
ethanhs: I don't think this import is needed?

test-data/unit/check-unreachable-code.test
```diff
@@ -302,6 +302,8 @@ else:
 import sys
 if sys.platform == 'fictional':
     def foo() -> int: return 0
+if platform.system():
```
ethanhs:
```suggestion
if platform.system() == 'fake':
```

I think this test should test comparison.

joybhallaa:
@ethanhs  thanks for reviewing this commit, will work on it ASAP

joybhallaa:
@ethanhs I've added the requested changes, please let me know what else needs to be done, and then I'll start to fix the checks!

mypy/options.py
ethanhs: It looks like you missed updating this based on the test failure. In your call to `consider_sys_platform` you use `options.sys_platform`, but here you are assigning to `options.platform` those two need to be consistently named.

joybhallaa:
@ethanhs I've renamed the variable, but I still think there is some logic inconsistencies

ethanhs:
@joybhallaa I will take a look at this later.

mypy/reachability.py
```diff
@@ -166,9 +165,14 @@ def consider_sys_platform(expr: Expression, platform: str) -> int:
         op = expr.operators[0]
         if op not in ('==', '!='):
             return TRUTH_VALUE_UNKNOWN
-        if not is_sys_attr(expr.operands[0], 'platform'):
-            return TRUTH_VALUE_UNKNOWN
-        right = expr.operands[1]
+        # check if either platform or platform_system is being used
+        if platform_system:
```
ethanhs: Right now you are saying the truth value is unknown if the expression is not a platform attribute, but you don't want to do that. Both `platform` and `platform_system` should be truthy, since they will both be set in options.

ethanhs:
Also, user configuration/command line input should override this. I'm not certain what exactly to do here, but I think mapping `sys.platform` values to `platform.system()` values would be best. (@JukkaL  thoughts?)

Looking at the [`sys.platform`](https://docs.python.org/3/library/sys.html#sys.platform) and [`platform.system()`](https://docs.python.org/3/library/platform.html#platform.system) docs, the following mapping makes sense:

```
{
     'win32': 'Windows',
     'linux': 'Linux',
     'darwin': 'Darwin',
}
```

There is some awkwardness with cygwin https://stackoverflow.com/a/32612402 and I would be surprised if anyone uses mypy on AIX.

test-data/unit/check-unreachable-code.test
```diff
@@ -302,6 +302,8 @@ else:
 import sys
 if sys.platform == 'fictional':
     def foo() -> int: return 0
+if platform.system() == 'fake':
+    def foo() -> int: return 0
```
JukkaL: There should also be a test that has a successful platform check.

Also test that `--platform` affects a `platform.system()` check.

hauntsaninja:
PR is pretty stale, feel free to re-open if you want to pick this back up and address feedback / broken tests

PR #8474, opened 2020-03-02
Change dict.fromkeys() test to be compatible with stricter type

bluetech:
`dict.fromkeys([1,2])` has type `Dict[int, Any]`. The `Any` was probably intended and is exercised in this test. But this stands in the way of (maybe) making it stricter in typeshed, to be `Dict[int, None]`.

Refs: https://github.com/python/typeshed/pull/3799

### DIFF
--- a/test-data/unit/pythoneval.test
+++ b/test-data/unit/pythoneval.test
@@ -210,7 +210,7 @@ True
 [case testDictFromkeys]
 import typing
 d = dict.fromkeys('foo')
-d['x'] = 2
+d['x'] = None
 d2 = dict.fromkeys([1, 2], b'')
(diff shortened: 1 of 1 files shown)


ilevkivskyi:
TBH I don't really like this idea. I think it should be `Dict[int, Optional[Any]]` instead.

Even better until PEP 585 is implemented I think mypy could use partial types to infer value type from first assignment (like it does for empty lists, dicts, etc). But this is too much work for a rare case. So `Optional[Any]` seems like the best way forward to me.

ilevkivskyi:
Just to clarify, I think the existing test should pass.

bluetech:
OK, that's understandable.

PR #8476, opened 2020-03-02
Trying to use dictionaries other than if-elif-else 

GantaChaitra:
mostly many files in mypy folder have their functions written in if-elif-else forms , rather we can use dictionaries .


JelleZijlstra:
Your description doesn't match the code in your PR, and it's questionable to me that the change in your PR is an improvement.

PR #8494, opened 2020-03-05
Fix fine-grained dependency to "__call__"

### LINKED ISSUE #8489: Mypy daemon dependency problem with __call__
This fine-grained incremental mode test case fails, probably because of missing dependencies to `__call__`:

```
[case testDunderCall]
from a import C

c = C()
c(1)

[file a.py]
class C:
    def __call__(self, x: int) -> None: ...

[file a.py.2]
class C:
    def __call__(self, x: str) -> None: ...

[out]
==                               
main:4: error: Argument 1 to "__call__" of "C" has incompatible type "int"; expected "str"
```

No error is currently reported when running the test case.

JukkaL:
Also fix the inferred type of a callee when we use `__call__`.
Previously we stored the callable type inferred from `__call__`,
which is incorrect.

Fixes #8489.

### DIFF
--- a/test-data/unit/fine-grained.test
+++ b/test-data/unit/fine-grained.test
@@ -9594,3 +9594,38 @@ class N(p.util.Test):
 [builtins fixtures/list.pyi]
 [out]
 ==
+
+[case testDunderCall1]
+from a import C
+
+c = C()
+c(1)
+
+[file a.py]
+class C:
+    def __call__(self, x: int) -> None: ...
+
+[file a.py.2]
+class C:
+    def __call__(self, x: str) -> None: ...
+
+[out]
+==
+main:4: error: Argument 1 to "__call__" of "C" has incompatible type "int"; expected "str"
+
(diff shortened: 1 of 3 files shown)


test-data/unit/fine-grained.test
```diff
@@ -9594,3 +9594,38 @@ class N(p.util.Test):
...
+class C:
+    def __call__(self, x: int) -> None: ...
+
+[file a.py.2]
+class C:
+    def __call__(self, x: str) -> None: ...
+
+[out]
+==
+main:3: error: Argument 1 to "__call__" of "C" has incompatible type "int"; expected "str"
```
ilevkivskyi: Maybe add a test where a `__call__` method is added/removed? For example:
```
from a import C

c = C()
x = c()  # type: ignore
x + 42  # error should appear here after an update

[file a.py]
class C: ...

[file a.py.2]
class C:
    def __call__(self) -> str: ...
```

PR #8496, opened 2020-03-05
[mypyc] Implement specialized int(str)

TH3CHARLie:
Relates mypyc/mypyc#644. Implement a specialized version of `int(str)`

@msullivan @JukkaL Please take a look, thanks!


mypyc/lib-rt/CPy.h
```diff
@@ -1042,6 +1042,21 @@ static PyObject *CPyLong_FromFloat(PyObject *o) {
     }
 }
 
+static PyObject *CPyLong_FromStrWithBase(PyObject *o, CPyTagged base) {
+    if (PyLong_Check(o)) {
```
msullivan: Do we need this check? It should always be a str, right?

TH3CHARLie: Yes. We should remove it.

Issue #8344, opened 2020-01-30
Too many arguments for IntEnum reported


Spenhouet:
If I define the following `IntEnum`:

    Numbers = IntEnum('Numbers', ['ZERO', 'ONE', 'TWO'], start=0)

mypy throws the following lint error: `Too many arguments for IntEnum()`.

The `start` parameter is an official methods:

> use the start parameter to specify a different starting value

https://docs.python.org/3/library/enum.html

Michael0x2a:
Yeah, it looks like the checks in https://github.com/python/mypy/blob/master/mypy/semanal_enum.py#L99 are too strict.

If you're feeling up for submitting a PR, I think it would probably be good enough relax the checks so they make sure there are only two *positional* arguments, not just two arguments in total, and that the keyword arguments all have the right names -- see `call.arg_kinds`, `call.arg_names`, and https://github.com/python/mypy/blob/master/mypy/nodes.py#L1501.

And if you also feel like checking the types of the keyword arguments, we can probably get away with insisting the user always uses int and str literals for start/qualname/module. (Supporting arbitrary variables is tricky, since we special-case handling enums before the type checking phase).

rahulrj:
@JukkaL is this a feature or a mypy bug? Can i submit a PR for this?

mridullpandey:
hi bro @rahulrj  are you working on it...

swetavooda:
If nobody is assigned for this issue, I would like to work on it.

rahulrj:
I am not working on this @mridullpandey

mridullpandey:
hi @swetavooda i have did some sort of modifications on it,will you please take another one ..pls let me know asap thanks :-)

mridullpandey:
@Michael0x2a  will you please elaborate it  what to actually do

swetavooda:
@mridullpandey If you are really willing to work on this, you may continue. I'll find another issue to work on.

Arshaan-256:
#8436

Arshaan-256:
Merge pull request #8436. Hopefully that would fix the issue.

mridullpandey:
then why this is still alive
@Arshaan-256

Arshaan-256:
@mridullpandey Because it has been reviewed yet. Once the PR is reviewed by the collaborator, only then can the issue be closed. The PR has passed the tests as well.

momohatt:
This seems to have been already fixed in #9123

JelleZijlstra:
[...]

PR #8499, opened 2020-03-05
Added a test file and updated semanal_enum file.

Arshaan-256:
Issue: #8344.
Added a check for valid IntEnum arguments. Wrote the test file.


test-data/unit/check-intEnum-arguments.test
```diff
@@ -0,0 +1,7 @@
+[case testIntEnumArguments]
+# flags: --python-version 3.6
+
+from enum import intEnum
+a = IntEnum('a', ['ZERO', 'ONE', 'TWO'], _start=10) # E: Unexpected keyword _start arguments for IntEnum()
+b = IntEnum('b', ['ZERO', 'ONE', 'TWO'], start=10, qualname='SomeClass.b')
+c = IntEnum('c', ['A', 'B', 'C'], start=17, module=__name__, qual_name='SomeClass.c') # E: Unexpected keyword qual_name arguments for IntEnum()
```
TH3CHARLie: Add a newline at the end of test

mypy/semanal_enum.py
```diff
@@ -106,10 +106,26 @@ def parse_enum_call_args(self, call: CallExpr,
...
+                    return self.fail_enum_call_arg("Too many arguments for %s()" % class_name, call)
+            elif len(args) == 5:
+                if call.arg_kinds != [ARG_POS, ARG_POS, ARG_NAMED_OPT, ARG_NAMED_OPT, ARG_NAMED_OPT]:
+                    return self.fail_enum_call_arg("Too many arguments for %s()" % class_name, call)
+            elif len(args) == 6:
+                if call.arg_kinds != [ARG_POS, ARG_POS, ARG_NAMED_OPT, ARG_NAMED_OPT, ARG_NAMED_OPT, ARG_NAMED_OPT]:
+                    return self.fail_enum_call_arg("Too many arguments for %s()" % class_name, call)
+            for arg in call.arg_names:
+                if arg not in ['module', 'qualname', 'type', 'start']:
+                    return self.fail_enum_call_arg("Unexpected keyword %s arguments for %s()" % (arg, class_name), call)
```
TH3CHARLie: fix lint

Arshaan-256: Ok! Thanks for your review :)

mypy/semanal_enum.py
```diff
@@ -106,10 +106,33 @@ def parse_enum_call_args(self, call: CallExpr,
...
             return self.fail_enum_call_arg("Too few arguments for %s()" % class_name, call)
-        if len(args) > 2:
-            return self.fail_enum_call_arg("Too many arguments for %s()" % class_name, call)
-        if call.arg_kinds != [ARG_POS, ARG_POS]:
-            return self.fail_enum_call_arg("Unexpected arguments to %s()" % class_name, call)
+        elif len(args) == 2:
+            if call.arg_kinds != [ARG_POS, ARG_POS]:
+                return self.fail_enum_call_arg("Unexpected arguments to %s()" % class_name, call)
+        elif len(args) > 2:
+            if len(args) == 3:
```
msullivan: This all seems correct but could you please refactor it to be a little more concise and remove some of the repetition? Instead of enumerating all possible cases, I'd suggest checking that there are between 2 and 6 arguments, that the first two are ARG_POS, and that the rest are ARG_NAMED_OPT.

Thanks!

Arshaan-256: Alright :)

mypy/semanal_enum.py
```diff
@@ -110,27 +110,14 @@ def parse_enum_call_args(self, call: CallExpr,
...
-                    return self.fail_enum_call_arg("Too many arguments for %s()" % class_name,
-                        call)
-            elif len(args) == 6:
-                if call.arg_kinds != [ARG_POS, ARG_POS,
-                        ARG_NAMED_OPT, ARG_NAMED_OPT, ARG_NAMED_OPT, ARG_NAMED_OPT]:
-                    return self.fail_enum_call_arg("Too many arguments for %s()" % class_name,
-                        call)
+            if call.arg_kinds[:2] != [ARG_POS, ARG_POS]:
+                return self.fail_enum_call_arg("Unexpected arguments to %s()" % class_name, call)
+            if all([True if arg == ARG_NAMED_OPT else False for arg in call.arg_kinds[2:]]):
```
JelleZijlstra:
```suggestion
            if all(arg == ARG_NAMED_OPT for arg in call.arg_kinds[2:]):
```

Arshaan-256: Thank you.

mypy/semanal_enum.py
```diff
@@ -106,10 +106,23 @@ def parse_enum_call_args(self, call: CallExpr,
...
+            if call.arg_kinds != [ARG_POS, ARG_POS]:
+                return self.fail_enum_call_arg("Unexpected arguments to %s()" % class_name, call)
+        elif len(args) > 2:
+            if call.arg_kinds[:2] != [ARG_POS, ARG_POS]:
+                return self.fail_enum_call_arg("Too many arguments for %s()" % class_name, call)
+            if all(arg == ARG_NAMED_OPT for arg in call.arg_kinds[2:]):
+                return self.fail_enum_call_arg("Too many arguments for %s()" % class_name, call)
+
+            for arg in call.arg_names:
+                if arg is None:
```
msullivan: I'm confused about this case? Won't this fire because of the positional arguments?

mypy/semanal_enum.py
```diff
@@ -106,10 +106,23 @@ def parse_enum_call_args(self, call: CallExpr,
...
-            return self.fail_enum_call_arg("Too many arguments for %s()" % class_name, call)
-        if call.arg_kinds != [ARG_POS, ARG_POS]:
-            return self.fail_enum_call_arg("Unexpected arguments to %s()" % class_name, call)
+        elif len(args) == 2:
+            if call.arg_kinds != [ARG_POS, ARG_POS]:
+                return self.fail_enum_call_arg("Unexpected arguments to %s()" % class_name, call)
+        elif len(args) > 2:
+            if call.arg_kinds[:2] != [ARG_POS, ARG_POS]:
+                return self.fail_enum_call_arg("Too many arguments for %s()" % class_name, call)
+            if all(arg == ARG_NAMED_OPT for arg in call.arg_kinds[2:]):
```
msullivan: I think all of these calls to `fail_enum_call_arg` could be merged

mypy/semanal_enum.py
Arshaan-256: I am worried about that too. Should I just not check for the first two elements of the list `arg_names`?

test-data/unit/check-intEnum-arguments.test
```diff
@@ -0,0 +1,7 @@
+[case testIntEnumArguments]
```
JukkaL, line 1: Instead of adding a new test file, can you add the test case to `test-data/unit/check-enum.test` instead? Currently the test case won't be run, I think.

Arshaan-256: Ok, will do!

PR #8500, opened 2020-03-06
Revert "mypy: remove has_member (#8438)"

msullivan:
It turns out that the has_member check is an important (accidental?)
performance optimization. Removing this caused a major (30+%?)
slowdown at dropbox. There might be a better way to optimize this but
I'm just going to revert it for now at least.

This reverts commit 09cdab4ffd34f89cc02f512fc35fdef57cb4c267.


msullivan:
Heads up @hauntsaninja

PR #8502, opened 2020-03-06
stubtest: various improvements

hauntsaninja:
- refine verify_signature, fixing a false negative and improving documentation
- ignore "this", another annoying module
- support using regexes in whitelist
- add --ignore-unused-whitelist
- handle name mangling

### DIFF
--- a/mypy/stubtest.py
+++ b/mypy/stubtest.py
@@ -9,6 +9,7 @@ import copy
 import enum
 import importlib
 import inspect
+import re
 import sys
 import types
 import warnings
@@ -1019,6 +1020,7 @@ def test_stubs(args: argparse.Namespace) -> int:
         for whitelist_file in args.whitelist
         for entry in get_whitelist_entries(whitelist_file)
     }
+    whitelist_regexes = {entry: re.compile(entry) for entry in whitelist}
 
     # If we need to generate a whitelist, we store Error.object_desc for each error here.
     generated_whitelist = set()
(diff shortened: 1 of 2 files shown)


mypy/stubtest.py
```diff
@@ -1058,7 +1072,9 @@ def test_stubs(args: argparse.Namespace) -> int:
 
     # Print unused whitelist entries
     for w in whitelist:
-        if not whitelist[w]:
+        # Don't consider an entry unused if it regex-matches the empty string
+        # This allows us to whitelist errors that don't manifest at all on some systems
+        if not whitelist[w] and not whitelist_regexes[w].fullmatch(""):
```
JelleZijlstra: I'm not sure I understand what's going on here. Could you give an example of a situation where this would apply?

hauntsaninja: The motivation here is to create a way to provide a better experience for cases like https://github.com/python/typeshed/issues/3782 where the whitelisted errors don't match up on some user's system.

For instance, on my system running stubtest gives me a number of (unfixable within current type system) errors like `termios.B921600 is not present at runtime`, but in CI we don't have any errors from `termios` at all. If we added a whitelist entry for `termios`, it would complain about an unused whitelist entry in CI. This allows us to add an entry for `(termios.[A-Z0-9_]+)?` so things work both on CI and on user's systems.

This is tested by the `unused_missing` part of `test_whitelist`

PR #8512, opened 2020-03-08
Add `--pep561-override` option to declare package names that are always considered typed

ntninja:
Title says it all: Packages declared using this option will not be ignored as *found module but no type hints or library stubs*.

This is intended as a last resort if one depends on a library that neither provides type hints *and* there are no external typing stubs available. In my specific use-case I depend on a library that provides some type hints in the code base, but they are incomplete/spotty and so the library authors prefer to not mark the package as PEP-561 compatible until they are more usable. They are still quite helpful when used from `mypy` however when one keeps their limitations in mind.

This still lacks docs and tests, but I'd like to hear what people here think about this before going down that road.

### DIFF
--- a/mypy/config_parser.py
+++ b/mypy/config_parser.py
@@ -85,6 +85,7 @@ config_types = {
     'package_root': lambda s: [p.strip() for p in s.split(',')],
     'cache_dir': expand_path,
     'python_executable': expand_path,
+    'pep561_override': lambda s: [p.strip() for p in s.split(',')],
 }  # type: Final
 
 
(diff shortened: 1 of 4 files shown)


ethanhs:
>I depend on a library that provides some type hints in the code base, but they are incomplete/spotty and so the library authors prefer to not mark the package as PEP-561 compatible until they are more usable.

Ideally in this situation, a stub package should be created I would argue. This allows for iteration on types (and opting into them while they are unstable) while keeping the main package stable.

I will think about this more.

ethanhs:
>the library authors prefer to not mark the package as PEP-561 compatible until they are more usable.

I have opened https://github.com/python/mypy/issues/8545 to discuss this further.

97littleleaf11:
Hello! Any progress on this PR?

ethanhs:
I think this will be superseded by an implementation of https://github.com/python/mypy/issues/8545

PR #8524, opened 2020-03-11
Add no_site_packages setting

### LINKED ISSUE #7768: Expose `--no-site-packages` to configuration file.
`mypy 0.730` currently supports a `--no-site-packages` flag, which disables searching for PEP561 compliant packages. Unfortunately, it does not seem to be exposed via a configuration file (I did not find it in https://mypy.readthedocs.io/en/latest/config_file.html and several variations in my config file raised an unused config error).

Would it be possible this option configurable via the config file?

davidzwa:
Fixes #7768 
Adds main.py logic for the `no_site_packages` flag. Tested as being subjective to CLI argument (override).

- [x] Add field to Options constructor
- [x] Add Bool key to config_parser.py
- [x] Adjust 2 places in main.py when checking for python executable.
- [x] Add test(s) in PEP 561 test file
- [x] Rebase onto master
- [x] Remove tests in `testpep561.py`
- [x] Merged new args and mypy.ini parse options for `pep561.test`
- [x] Add test to `pep561.test` file
- [x] Run test and add proper messages

**Note**: run test with the following command
> python -m pytest -k testTypedPkg_nositepackages ./mypy


mypy/config_parser.py
```diff
@@ -78,6 +78,7 @@ def split_and_match_files(paths: str) -> List[str]:
     'junit_xml': expand_path,
     # These two are for backwards compatibility
     'silent_imports': bool,
+    'no_site_packages': bool,
```
ethanhs: I think you stuck this between two configs the comment above points to, so you should probably put it below `almost_silent`.

davidzwa: Thanks for responding so fast. I have applied the suggestion. Is it possible to add a test somewhere? I performed 2 manual test-cases (both testing combination of CLI and config version of the flag).

ethanhs: Good question! The best place for a test would probably be in https://github.com/python/mypy/blob/0e8e135548a4d2fde42f93e2a2a56802af019263/test-data/unit/cmdline.test.

davidzwa: I have an idea that the following is not monkey-proof:
```
[case testNoSitePackagesFlag]
# cmd: mypy emtpy_test.py --no-site-packages --verbose
[file emtpy_test.py]
pass
[out]
LOG:  Configured Executable:  None
```

And the test with pytest also fails, as it is not the right line, so this is just to indicate what I'd like to achieve. I'd like the test the fact that the options class ends up with the right value, so this is one of the way I thought it might be achievable.
Any thoughts on how to make this more robust would be appreciated as my knowledge of the test syntax is limited.

ethanhs: Ah, the best way to test this may be to add to https://github.com/python/mypy/blob/master/mypy/test/testpep561.py

PEP 561 tests are somewhat special.

davidzwa: I find line [Line 235](https://github.com/python/mypy/blob/1677efdfd7c37ee91d64bc018b12d6bfda042655/mypy/test/testpep561.py#L235) on building a config very handy. That's definitely the way to go.

But [Line 249](https://github.com/python/mypy/blob/1677efdfd7c37ee91d64bc018b12d6bfda042655/mypy/test/testpep561.py#L249) about accessing the mypy module(s), makes me worried that I still have to scan `stdout` (see [main.py line 63](https://github.com/python/mypy/blob/1677efdfd7c37ee91d64bc018b12d6bfda042655/mypy/main.py#L63)). Is that acceptible in your view? It is prone to false-positives.

ethanhs: Yes, you will have to match against stdout, which isn't ideal, but how most of our testing works. You can see how that is done for other tests based on this: https://github.com/python/mypy/blob/master/mypy/test/testpep561.py#L55. If the test is static there should be no false positive.

davidzwa: I have added a static test like previously discussed by extending the ExampleProgram with args. Any idea why it's failing on importing `typedpkg`? I've copied the other tests, so this is unexpected.

mypy/test/testpep561.py
```diff
@@ -250,6 +277,31 @@ def test_mypy_path_is_respected(self) -> None:
...
+                    mypy_file.write('[mypy]\n')
+                    mypy_file.write('no_site_packages=True\n')
+
+                with self.virtualenv() as venv:
+                    venv_dir, python_executable = venv
+                    self.install_package('typedpkg', python_executable)
+                    self.install_package('typedpkg-stubs', python_executable)
+                    self.simple_prog.check_mypy_run_args(
+                        python_executable,
+                        expected_out=[SimpleMsg.msg_skip_executable],
```
ethanhs: The expected output is supposed to be the entire expected output. I would not use -vv and instead use the warning mypy gives for not finding the package.

ethanhs:
>Any idea why it's failing on importing typedpkg?

That is the point of no-site-packages, it disables mypy from searching the site-packages directory for installed packages, failing to import typedpkg is the expected result. I think you should change the expected output to be the failure to find typedpkg warning.

mypy/test/testpep561.py
```diff
@@ -124,6 +125,32 @@ def check_mypy_run(self,
         finally:
             if venv_dir is not None:
                 os.chdir(old_dir)
+    
+    def check_mypy_run_args(self,
```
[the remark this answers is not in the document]
davidzwa: Applied this.

mypy/test/testpep561.py
davidzwa: Applied! I had some trouble changing the line numbers (f.e. some had different line numbers) and had to make a new message for that. Wouldn't it be better to inject line numbers on the spot where the list `expected_out` is put together?

davidzwa:
Not to happy with the ':4' and ':2' messages in the `testpep561.py` file. Thinking about formatting the message with a line number in the `expected_out` list. Sound good?

mypy/test/testpep561.py
```diff
@@ -56,11 +56,18 @@ class SimpleMsg(Enum):
     msg_dne = "{tempfile}:3: error: Module 'typedpkg' has no attribute 'dne'"
     msg_list = "{tempfile}:5: note: Revealed type is 'builtins.list[builtins.str]'"
     msg_tuple = "{tempfile}:5: note: Revealed type is 'builtins.tuple[builtins.str]'"
+    msg_any = "{tempfile}:5: note: Revealed type is 'Any'"
 
 
 class NamespaceMsg(Enum):
     cfm_beta = ("{tempfile}:4: error: Cannot find implementation or library stub for module named "
                 "'typedpkg_ns.ns.dne'")
+    missing_typedpkg_module = ("{tempfile}:3: error: Cannot find implementation or library stub"
```
ethanhs: Instead of messing with string formatting, I would make your own Enum for these values.

davidzwa: Wow glad to be back. I am not too sure what too change, since the messages are currently added to the proper class without formatting (that was a different idea).
I've added:
- missing_typedpkg_module
- missing_typedpkg_sample_module
- help_note2 (copy of help_note)

Anything specific worth improving?

davidzwa: @ethanhs I'd gladly hear your thoughts.

ethanhs: Hi @davidzwa, I actually restructured the tests to be a bit more data driven. Sorry for making extra work for you. If you rebase on master and add your test to  https://github.com/python/mypy/blob/master/test-data/unit/pep561.test, you should just be able to put the test there and then put the messages in the `[out]` section (please also put `# pkg: typedpkg` at the beginning of your test, so that we can check it is not finding the installed package). Hopefully this is a bit clearer than before.

davidzwa: I am currently adding the test. You can see my progress in the check list at the top.

davidzwa: @ethanhs in hindsight: how do I add the configuration file now? I remember from earlier that I can add a line `# cmd: mypy --no-site-packages` to trigger the flag, but the `mypy.conf` I cannot create.

I am ok with just testing the no-site-packages command line flag and leaving the Options class untested, if that's ok?

davidzwa:
@ethanhs I added a test, but I think this is not working correctly. I tried the following variants:
- `# flags: no-site-packages`
- `# mypy: no-site-packages`
- `# mypy: no-site-packages=true` 
- `# cmd: mypy --no-site-packages`

But all give the "testTypedPkg_nositepackages.py:6: note: Revealed type is 'builtins.tuple[builtins.str]'" note instead of an import error. I hope you can take over to find the solution.

ethanhs:
@davidzwa hm, yeah it looks like the test harness doesn't support flags like that. You can probably re-use this: https://github.com/python/mypy/blob/master/mypy/test/testcmdline.py#L100, move it to `mypy/test/helpers.py` and call it [here](https://github.com/python/mypy/blob/master/mypy/test/testpep561.py#L103) and have it parse `testcase.input[1]` (Since the first line already handles packages). Let me know if that doesn't make sense, or I can do it myself if you wish.

ethanhs:
@davidzwa you should be able to pull my changes from the `no-sp-fix`.

You can pull the changes with something like:
```
$ git remote add ethanhs https://github.com/ethanhs/mypy.git
$ git fetch ethanhs
$ git merge ethanhs/no-sp-fix
```
Let me know if you have any problems.

davidzwa:
Merged, tomorrow I will look at how to adapt test.

ethanhs:
@davidzwa if you look at the test I added, you have to escape the `[`

So:
```
\[mypy]
no_site_packages = True
```
should work. Let me know if it doesn't.

davidzwa:
I think I should learn to read more carefully. You'll find the the two tests sufficient, I'm sure of it!

davidzwa:
If you can point me to a related issue including this test-framework, that'd be awesome

PR #8536, opened 2020-03-13
Path to source files added in verbose mode

akerami:
Implementation of parts of #6544 , namely of printing the path of the source files given to be processed by mypy. Tested using the command line as well as a configuration file.


msullivan:
There are a bunch of test failures. I'll take a look if you fix them.

mypy/build.py
```diff
@@ -2559,14 +2559,30 @@ def log_configuration(manager: BuildManager) -> None:
...
+    ]
+
+    src_pth_str = "Source Path File"
+    src_pths = manager.source_set.source_paths
+
+    if len(src_pths) > 1:
+        src_pth_str += "s"
+        configuration_vars.append((src_pth_str, ""))
+        for source_path in src_pths:
+            configuration_vars.append(("." * len(src_pth_str), source_path))
```
JukkaL: This looks odd to me. For consistency with other verbose, I think that it would be better to show all source paths on a single line, separated by spaces.

mypy/build.py
```diff
@@ -2559,14 +2559,30 @@ def log_configuration(manager: BuildManager) -> None:
...
+    elif len(src_pths) == 1:
+        configuration_vars.append((src_pth_str, src_pths.copy().pop()))
```
JukkaL: This case will go away once you implement the suggestion above.

mypy/build.py
```diff
@@ -2559,14 +2559,30 @@ def log_configuration(manager: BuildManager) -> None:
...
 
     manager.log()
-    configuration_vars = (
+    configuration_vars = [
         ("Mypy Version", __version__),
         ("Config File", (manager.options.config_file or "Default")),
-        ("Configured Executable", manager.options.python_executable),
+    ]
+
+    src_pth_str = "Source Path File"
```
JukkaL: This would clearer if it would be just "Source Path".

akerami:
Thanks! I will make the changes.

PR #8549, opened 2020-03-17
Fix invalid type causing namedtuple errors in wrong files

### LINKED ISSUE #8511: Wrong file blamed for a mixture of NamedTuple and passing a function instead of type.
bug

  * Create an empty dir/__init__.py.
  * Create dir/file1.py with:
```
import typing

def foo():
  pass


Type1 = typing.NamedTuple('Type1', [
    ('foo', foo),
])
```
  * Create file2.py (in the root dir) with:
```
from typing import NamedTuple

from dir.file1 import Type1

Type2 = NamedTuple('Type2', [
    ('x', Type1),
])
```
  * Run: mypy file2.py --follow-imports=silent

The error message is:
file2.py:8: error: Name 'foo' is not defined

-> Note that file2.py has no line 8 (it the problematic line in file1.py) and no mention of `foo`.

  * the error message should point to file1.py or say something about Type1 in file2.py being invalid.
  * foo is well defined in this case, making the error message more cryptic.

When reducing this case I've found that removing --follow-import-silent gives an error message about file1.py which is actionable, but the other message is still present and is wrong, since the line number refers to a line in file1.py.

mypy 0.761 on Python 3.5.2

--follow-imports=silent

msullivan:
Fixes #8511.

This is a workaround of #4987, where UnboundTypes can leak.

### DIFF
--- a/test-data/unit/check-namedtuple.test
+++ b/test-data/unit/check-namedtuple.test
@@ -959,6 +959,6 @@ from typing import NamedTuple
 def foo():
     pass
 
-Type1 = NamedTuple('Type1', [('foo', foo)])  # type: ignore
+Type1 = NamedTuple('Type1', [('foo', foo)])  # E: Function "b.foo" is not valid as a type  # N: Perhaps you need "Callable[...]" or a callback protocol?
 
 [builtins fixtures/tuple.pyi]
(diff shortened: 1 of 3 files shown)


test-data/unit/check-namedtuple.test
```diff
@@ -948,3 +948,17 @@ class A:
...
+from typing import NamedTuple
+from b import Type1
+Type2 = NamedTuple('Type2', [('x', Type1)])
+[file b.py]
+from typing import NamedTuple
+
+def foo():
+    pass
+
+Type1 = NamedTuple('Type1', [('foo', foo)])  # type: ignore
```
Michael0x2a: Maybe instead we check the error message that's generated here to make sure it's something sensible?

PR #8552, opened 2020-03-17
Don't free too much before doing fast_exit

msullivan:


mypy/main.py
```diff
@@ -129,6 +128,9 @@ def flush_errors(new_messages: List[str], serious: bool) -> None:
     elif code:
         sys.exit(code)
 
+    # HACK: keep res alive so that mypyc won't free it before the hard_exit
+    res2 = res  # noqa
```
JukkaL: If mypyc would be a bit smarter, this could be optimized away. What about doing something a bit more involved like `list([res])` which likely would not get optimized away even by a clever compiler?

PR #8559, opened 2020-03-18
Work towards following imports in mypy daemon

JukkaL:
This implements some basic use cases, such as adding files, removing files,
and basic package support.

The implementation is separate from what is used for coarse-grained 
incremental mode, since it seemed too difficult (and risky) to try to
merge them.

Many special cases are not handled correctly yet, and code is in need of some
cleanup. Some functionality is still missing.

Following imports is only enabled in tests for now. I'll enable the feature 
generally after it's more fully implemented.


mypy/server/update.py
```diff
@@ -1108,3 +1109,51 @@ def target_from_node(module: str,
...
+    that 'a.b' is not a module but a regular name.
+
+    This is only relevant when following imports normally.
+
+    Args:
+        module: target package in which to look for submodules
+        path: path of the module
+    """
+    # TODO: Windows paths
+    if path is None or not path.endswith('/__init__.py'):
```
ethanhs:
```suggestion
    if path is None or not path.endswith(os.sep + '__init__.py'):
```

mypy/dmypy_server.py
```diff
@@ -486,8 +513,179 @@ def fine_grained_increment(self,
...
+        for module_id, state in graph.items():
+            refresh_suppressed_submodules(module_id, state.path, fine_grained_manager.deps, graph,
+                                          self.fscache)
+
+        # There may be new files that became available, currently treated as
+        # suppressed imports. Process them.
+        seen_suppressed = set()  # type: Set[str]
+        while True:
+            # TODO: Merge seen and seen_suppressed?
+            new_suppressed, seen_suppressed = self.find_added_suppressed(
```
msullivan: This isn't really `new_suppressed`, right? This is stuff that is becoming *un*suppressed?

mypy/server/update.py
```diff
@@ -1108,3 +1109,51 @@ def target_from_node(module: str,
...
+        module: str,
+        path: Optional[str],
+        deps: Dict[str, Set[str]],
+        graph: Graph,
+        fscache: FileSystemCache) -> None:
+    """Look for submodules that are now suppressed in target package.
+
+    If a submodule a.b gets added, we need to mark it as suppressed
+    in modules that contain "from a import b". Previously we assumed
+    that 'a.b' is not a module but a regular name.
```
msullivan: Why does it need to be marked suppressed? We want it to be a real dep, right? Is this just an implementation artifact of how this is handled in a loop of processing suppressed imports?

mypy/server/update.py
```diff
@@ -1108,3 +1109,51 @@ def target_from_node(module: str,
...
+
+    Args:
+        module: target package in which to look for submodules
+        path: path of the module
+    """
+    # TODO: Windows paths
+    if path is None or not path.endswith(os.sep + '__init__.py'):
+        # Only packages have submodules.
+        return
+    # Find and submodules present in the directory.
```
msullivan: and -> any

mypy/server/update.py
```diff
@@ -1108,3 +1109,51 @@ def target_from_node(module: str,
...
+        if (not fnam.endswith(('.py', '.pyi'))
+                or fnam.startswith("__init__.")
+                or fnam.count('.') != 1):
+            continue
+        shortname = fnam.split('.')[0]
+        submodule = module + '.' + shortname
+        trigger = make_trigger(submodule)
+        if trigger in deps:
+            for dep in deps[trigger]:
+                # TODO: <...> deps, imports in functions, etc.
```
msullivan: Do we think this approach will scale to handling this stuff?

mypy/dmypy_server.py
```diff
@@ -361,8 +361,12 @@ def cmd_recheck(self,
         t1 = time.time()
         manager = self.fine_grained_manager.manager
         manager.log("fine-grained increment: cmd_recheck: {:.3f}s".format(t1 - t0))
-        res = self.fine_grained_increment(sources, is_tty, terminal_width,
-                                          remove, update)
+        if self.options.follow_imports != 'normal':
```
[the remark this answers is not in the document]
JukkaL: I'm not yet sure whether it's worth supporting silent imports at this point. The function is a good idea.

mypy/dmypy_server.py
```diff
@@ -486,8 +513,179 @@ def fine_grained_increment(self,
...
+                assert nxt.path is not None  # TODO
+                changed.append((nxt.module, nxt.path))
+            elif nxt.module in graph:
+                state = graph[nxt.module]
+                for dep in state.dependencies:
+                    if dep not in seen:
+                        seen.add(dep)
+                        worklist.append(BuildSource(graph[dep].path,
+                                                    graph[dep].id))
+        return seen, changed, [], new_files
```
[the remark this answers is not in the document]
JukkaL: Yeah, I'll fix it.

mypy/dmypy_server.py
```diff
@@ -486,8 +513,179 @@ def fine_grained_increment(self,
...
+                                                       List[Tuple[str, str]],
+                                                       List[BuildSource]]:
+        """Follow imports within graph from given sources.
+
+        Args:
+            sources: roots of modules to search
+            graph: module graph to use for the search
+            seen: modules we've seen before (that won't be visited)
+            changed_paths: which paths have changed (stop search here and return any found)
+            sources_set: set of sources (TODO: relationship with seen)
```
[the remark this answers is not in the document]
JukkaL: I'll clean it up in a follow-up PR.

mypy/server/update.py
JukkaL: I hope so, with some further optimizations perhaps. Following imports might not be very performant for code bases with more than (say) 500 kLOC, but even if that's the case, it would still probably be fine for a significant majority of users. Users with very large codebases can continue to disable following imports, and we just need to document this clearly.

 It sure would be better to have this perform well with large code bases, but I'm not sure how much extra work that would take.

mypy/server/update.py
JukkaL: This just arguably makes the logic a bit simpler, as these can be handled via the same code as "real" suppressed modules. If this turns out to be bad for performance, for example, we could turn these into real deps and process these separately.

mypy/dmypy_server.py
```diff
@@ -486,8 +513,179 @@ def fine_grained_increment(self,
...
+        changed_paths = self.fswatcher.find_changed()
+
+        assert self.fine_grained_manager is not None
+        fine_grained_manager = self.fine_grained_manager
+        graph = fine_grained_manager.graph
+        manager = fine_grained_manager.manager
+
+        orig_modules = list(graph.keys())
+
+        self.update_sources(sources)
```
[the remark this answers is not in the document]
JukkaL: Not sure. I'll add a TODO comment about these.

mypy/dmypy_server.py
JukkaL: Yes, this is confusing. I'll rename the variable.

PR #8569, opened 2020-03-21
[mypyc] Improve documentation in mypyc.primitives

JukkaL:
The aim is to give at least a short comment for all non-trivial ops.

Add docstrings to all modules that don't have them.

Also minor fix to `py_call_with_kwargs_op`.


JukkaL:
@msullivan Ping

Issue #7864, opened 2019-11-04
TypeVars should be allowed to be the same


twoertwein:
Two unbound TypeVars `A` and `B` can potentially be the same:

```py
from typing import Callable, Iterable, TypeVar

A = TypeVar("A")
B = TypeVar("B")

def identity(x: A) -> A:
    return x

# ok
def _map(queue: Iterable[B], function: Callable[[B], A]) -> Iterable[A]:
    return map(function, queue)

# fails
def _map2(queue: Iterable[B], function: Callable[[B], A] = identity) -> Iterable[A]:
    return map(function, queue)
```

The introduced default keyword argument leads to the following unexpected error:
`error: Incompatible default for argument "function" (default has type "Callable[[A], A]", argument has type "Callable[[B], A]")`

I'm using mypy 0.740.

ilevkivskyi:
> TypeVars should be allowed to be the same

No, they shouldn't. The default value isn't part of the function API, so calls with omitted second argument will be either unsafe or underspecified. For example, imagine you have an entry in a stub:
```python
def _map2(queue: Iterable[B], function: Callable[[B], A] = ...) -> Iterable[A]: ...
```
assuming this type, what should be the inferred type of `_map2([1, 2, 3])`? Instead you should just use a correct signature for your function:
```python
@overload
def _map2(queue: Iterable[A]) -> Iterable[A]: ...
@overload
def _map2(queue: Iterable[B], function: Callable[[B], A]) -> Iterable[A]: ...
def _map2(queue, function=identity):
    # implementation goes here
    ...
```
Surprisingly, this often comes as a source of confusion (I have seen half a dozen issues about this), so I think we should document this (probably in the section "Common issues").

twoertwein:
thank you for the clarifications!

It might be helpful if mypy/the typing module provide examples of some more 'complex' functions (map, zip, function decorators, callables which have a few typed arguments and arbitrarily many Any arguments).

tz-earl:
First time contributor here. If no one else is expanding the docs, I'd like to take a stab at it.

ethanhs:
@tz-earl Welcome, please feel free to get started on it! Let me know if you have any questions.

tz-earl:
@ethanhs Thanks a lot for the go-ahead.

Just to be sure, it seems that there is an older package **overload** that has been superceded by the package **overloading**. Is that right?

JelleZijlstra:
[...]

PR #8572, opened 2020-03-22
Fixes:  #7864 TypeVars should be allowed to be the same

tz-earl:
For the **Common Issues and Solutions** doc, added a section **TypeVars are not the same**.

I'm not entirely sure about the example I concocted to show code that does pass muster with Mypy.

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -846,12 +846,8 @@ Likewise, this error will occur for constrained and bound TypeVars as well.
 
 .. code-block:: python
 
-    A = TypeVar("A", bound=int)
-    B = TypeVar("B", bound=int)
-
-or
-
-.. code-block:: python
-
-    A = TypeVar("A", int, float, complex)
-    B = TypeVar("B", int, float, complex)
+    A = TypeVar("A", int, str)
+    B = TypeVar("B", int, str)
+    # or
+    A = TypeVar("A", bound=complex)
+    B = TypeVar("B", bound=complex)
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -777,7 +777,6 @@ This example demonstrates both safe and unsafe overrides:
 
     class A:
         def test(self, t: Sequence[int]) -> Sequence[str]:
-<<<<<<< HEAD
             ...
 
     class GeneralizedArgument(A):
@@ -851,3 +850,36 @@ Likewise, this error will occur for constrained and bound TypeVars as well.
     # or
     A = TypeVar("A", bound=complex)
     B = TypeVar("B", bound=complex)
+
+
+In some cases, the ``@overload`` decorator can be used to provide the desired type checking. Here's an example.
+
+.. code-block:: python
+
+    from typing import Callable, Iterable, TypeVar, overload, Tuple
+
+    A = TypeVar("A")
+    B = TypeVar("B")
+
+    def identity(x: A) -> A:
+        return x
+
+    def second(tup: Tuple[int, str]) -> str:
+        return tup[1]
+
+    @overload
+    def _map2(queue: Iterable[A]) -> Iterable[A]: ...
+    @overload
+    def _map2(queue: Iterable[A], function: Callable[[A], A]) -> Iterable[A]: ...
+    @overload
+    def _map2(queue: Iterable[B], function: Callable[[B], A]) -> Iterable[A]: ...
+    def _map2(queue, function=identity):
+        return map(function, queue)
+
+    list_1 = [2, 4, 6, 8]
+    list_2 = ["hello", "world"]
+    list_3 = list(enumerate(["fazzle", "baz", "rompl" ]))
+
+    mapped_1 = [n for n in _map2(list_1)]
+    mapped_2 = [s for s in _map2(list_2, identity)]
+    mapped_3 = [t for t in _map2(list_3, second)]
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -777,6 +777,7 @@ This example demonstrates both safe and unsafe overrides:
 
     class A:
         def test(self, t: Sequence[int]) -> Sequence[str]:
+<<<<<<< HEAD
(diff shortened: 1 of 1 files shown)


docs/source/common_issues.rst
```diff
@@ -808,3 +808,78 @@ not necessary:
     class NarrowerArgument(A):
         def test(self, t: List[int]) -> Sequence[str]:  # type: ignore[override]
             ...
+
+TypeVars are not the same
+------------------------------------------------
+
+As discussed in the issue
+`TypeVars should be allowed to be the same <https://github.com/python/mypy/issues/7864>`_,
```
JukkaL: Style nit: we don't add links to mypy issues in the documentation, since the discussions in issues may be too technical for many readers (further information is likely needed by beginners). In this case it's okay to not have any link at all, or you can try to find a source which is more geared towards general developer population.

docs/source/common_issues.rst
```diff
@@ -808,3 +808,78 @@ not necessary:
...
+TypeVars are *not* equivalent even if they are identically defined.
```
JukkaL: I'd add one or two sentences about why this is the case. I'm not sure what's the best way express this, but each type variable can vary independently.

docs/source/common_issues.rst
```diff
@@ -808,3 +808,78 @@ not necessary:
...
+
+Here is an example using two TypeVars that are unconstrained and unbound:
```
JukkaL: Nit: I'd leave out "unconstrained and unbound" since most type variables are like this in my experience, and not all users are familiar with constraints and bounds, possibly causing some confusion.

docs/source/common_issues.rst
```diff
@@ -808,3 +808,78 @@ not necessary:
...
+    from typing import Callable, Iterable, TypeVar
+
+    A = TypeVar("A")
+    B = TypeVar("B")
+
+    def identity(x: A) -> A:
+        return x
+
+    # ok
+    def _map(queue: Iterable[B], function: Callable[[B], A]) -> Iterable[A]:
```
JukkaL: Nit: leave out the underscore prefixes, since they serve no function in a code example.

docs/source/common_issues.rst
```diff
@@ -808,3 +808,78 @@ not necessary:
...
+        return map(function, queue)
+
+    # fails
+    def _map2(queue: Iterable[B], function: Callable[[B], A] = identity) -> Iterable[A]:
+        return map(function, queue)
+
+
+The signature of ``_map2()`` triggers
+``error: Incompatible default for argument "function" (default has type "Callable[[A], A]",``
+``argument has type "Callable[[B], A]")``
```
JukkaL: The error message would look nicer if formatted in a code block.

docs/source/common_issues.rst
```diff
@@ -808,3 +808,78 @@ not necessary:
...
+.. code-block:: python
+
+    A = TypeVar("A", int, str)
+    B = TypeVar("B", int, str)
+    # or
+    A = TypeVar("A", bound=complex)
+    B = TypeVar("B", bound=complex)
+
+
+In some cases, the ``@overload`` decorator can be used to provide the desired type checking. Here's an example.
```
JukkaL: Style nit: write "... an example:" (with a colon)

A simpler workaround would be to cast the default value to the desired type using `cast`.

docs/source/common_issues.rst
```diff
@@ -808,3 +808,78 @@ not necessary:
...
+    def identity(x: A) -> A:
+        return x
+
+    def second(tup: Tuple[int, str]) -> str:
+        return tup[1]
+
+    @overload
+    def _map2(queue: Iterable[A]) -> Iterable[A]: ...
+    @overload
+    def _map2(queue: Iterable[A], function: Callable[[A], A]) -> Iterable[A]: ...
```
JukkaL: The second overload item doesn't seem useful, as it's just a special case of the third.

docs/source/common_issues.rst
```diff
@@ -808,3 +808,78 @@ not necessary:
     class NarrowerArgument(A):
         def test(self, t: List[int]) -> Sequence[str]:  # type: ignore[override]
             ...
+
+TypeVars are not the same
```
JukkaL: Suggestion for an alternative title: "Each type variable is distinct"

tz-earl:
@JukkaL Thank you for your careful review of this PR. I will be adopting the minor changes and addressing the more substantive ones.

Question: what's the process I should follow to make further changes to the branch for this PR? I see that there are a number of different opinions about how to do this.

JelleZijlstra:
@tz-earl you should make new local commits and push them to your remote branch. Don't force-push or rebase.

tz-earl:
@JelleZijlstra Thank you for the helpful clarification about more commits

docs/source/common_issues.rst
tz-earl: I was able to eliminate the second overload def:

`def map_2(queue: Iterable[A], function: Callable[[A], A]) -> Iterable[A]: ...`

by using a cast like so:

`map_2(list_2, cast(Callable[[B], A], identity))`

Is this what you meant?

If so, I would prefer the original approach of keeping the extra overload def so that the actual call is easier to read and to understand.

97littleleaf11:
Hi! Could you please fix the conflicts?

docs/source/common_issues.rst
mixed-source: Hi, I'm not sure whether it is valid to say like this. 
`TypeVar' is not one of the types such as `int`, `str`. Hence each type variable is unique even though they have identical signatures since it is not meant to be compared with each other.

docs/source/common_issues.rst
mixed-source: Hi, I'm not sure whether this is the workaround that @JukkaL suggested. 
```
def map2(queue: Iterable[B], function: Callable[[B], A] = cast(Callable[[B], A], identity)) -> Iterable[A]: ...
```

PR #8573, opened 2020-03-22
Added example in the doc for TypeVar in using generic classes

### LINKED ISSUE #8523: Error when preserving generic type behaviour in subclasses (when not supported at runtime)
This is a bug report.
* mypy = 0.770
* python = 3.8.2
* I did not try running mypy from git master
----
The docs for "Using classes that are generic in stubs but not at runtime" ([here](https://mypy.readthedocs.io/en/stable/common_issues.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime)) clearly indicate how to subclass classes that are defined to be generic in their stubs but don't actually support generics in their runtime code.

The shown method (conditionally defining a typed subclass) works great when you want to lock the type down in your subclass. However, **if you want to maintain the subclass as a generic it does not work** (mypy reports an error).

Here is failing case (with `mypy --strict`), where the only variation from the canonical [example](https://mypy.readthedocs.io/en/stable/common_issues.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime) is using a generic `TypeVar` instead of a `str` as the type for the Queue:

```python
from queue import Queue
from typing import TYPE_CHECKING, TypeVar

_T = TypeVar("_T")
if TYPE_CHECKING:
    MyQueueBase = Queue[_T]  # change _T to str here and it works perfectly (per the docs)
else:
    MyQueueBase = Queue
class MyQueue(MyQueueBase): pass  # mypy error is on this line (9)
```

mypy reports an error as follows (when the base is `Queue[_T]`):

```
$ python --version
Python 3.8.2
$ mypy --version
mypy 0.770
$ mypy --strict ./generic_subclass_example.py 
generic_subclass_example.py:9: error: Missing type parameters for generic type "MyQueueBase"
```

Note that without `--strict` there is no error (not sure which of the flags causes it).

ashutosh1919:
Fixes #8523 . 
@JukkaL and @ethanhs, Please review this PR.

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -595,9 +595,12 @@ string literal types or :py:data:`~typing.TYPE_CHECKING`:
 
    results: 'Queue[int]' = Queue()  # OK
 
-But while using :py:data:`~typing.TypeVar`, observe here that it requires parameters
-while using it as a type. And if we implement it in above way it will give error as
-shown in below example.
+If you use a type variable in the type alias that is used as a base class and
+implement it in above way, it will give error. Note that type arguments are only
+required when using `--disallow-any-generics` (or `--strict`). However, this will
+still be a problem even if these options aren't enabled, since we want to make the
+subclass generic (otherwise it would be pretty pointless to use a type variable),
+and this would result in a runtime error without the suggested trick.
 
 .. code-block:: python
 
@@ -612,6 +615,7 @@ shown in below example.
 
    class MyQueue(MyQueueBase): pass  # error: Missing type parameters for generic type
 
+Note here that this error only comes when some configuration options are enabled.
 To avoid this problem, we need to create another class in the hierarchy which makes
 use of :py:data:`~typing.Generic` when :py:data:`~typing.TYPE_CHECKING` is disabled
 and behaves normally when :py:data:`~typing.TYPE_CHECKING` is enabled.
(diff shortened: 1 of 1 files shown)


docs/source/common_issues.rst
```diff
@@ -595,6 +595,40 @@ string literal types or :py:data:`~typing.TYPE_CHECKING`:
...
+   from queue import Queue
+   from typing import TYPE_CHECKING, TypeVar
+
+   _T = TypeVar("_T")
+   if TYPE_CHECKING:
+       MyQueueBase = Queue[_T]
+   else:
+       MyQueueBase = Queue
+
+   class MyQueue(MyQueueBase): pass  # error: Missing type parameters for generic type
```
JukkaL: As mentioned above, the error only comes when some configuration options are enabled.

docs/source/common_issues.rst
```diff
@@ -595,6 +595,40 @@ string literal types or :py:data:`~typing.TYPE_CHECKING`:
 
    results: 'Queue[int]' = Queue()  # OK
 
+But while using :py:data:`~typing.TypeVar`, observe here that it requires parameters
+while using it as a type. And if we implement it in above way it will give error as
```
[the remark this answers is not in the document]
ashutosh1919: done

docs/source/common_issues.rst
ashutosh1919: done

ashutosh1919: done

docs/source/common_issues.rst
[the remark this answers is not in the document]
ashutosh1919: done

sbdchd:
Any update on this? Took me some spelunking to find it

hauntsaninja:
Subsumed by https://github.com/python/mypy/pull/9763

Issue #7968, opened 2019-11-18
Set MYPY_CONFIG_FILE_DIR environment variable


blueyed:
In addition to / as an alternative to https://github.com/python/mypy/issues/7967 mypy could set an environment variable of the config file being used.

This could be used with e.g. `mypy_path` then (`mypy_path = $MYPY_CONFIGFILE/src`), and also with defaults like `cache_dir`.

JukkaL:
I like this better than #7967, since this preserves backward compatibility. My proposal for the name of the variable is `MYPY_CONFIG_FILE_DIR`.

mridullpandey:
hi @blueyed  i want work on it will  you please elaborate  it

blueyed:
@mridullpandey 
What is unclear here?
It should set `MYPY_CONFIG_FILE_DIR` in `os.environ`, so that it can be expanded / is available while parsing the config.

blueyed:
btw: I think `$MYPY_CONFIG_FILE(_PATH)` might be more flexible: it would provide the actual file (which might be `mypy.ini`, `setup.cfg` etc), and could also be used as `$MYPY_CONFIG_FILE/../…` to have its directory then.
While `$MYPY_CONFIG_FILE_DIR` might be more explicit, we might end up with adding `$MYPY_CONFIG_FILE_PATH` additionally afterwards.

JukkaL:
@blueyed Do you have a use case for using the path to the config file, instead of the directory where it's located? Using `..` after a path to a file looks odd to me (though it's not hard to support), and I'd rather avoid it.

@mridullpandey Config file parsing is implemented in `mypy.config_parser`. In particular, `expand_path` seems relevant.

blueyed:
@JukkaL 
Not really, but thought about it when adding `PYTEST_CONFIG_{FILE,DIR}` to pytest, and figured it might be more flexible, since you cannot determine the ini file being used from the dir (due to different names).
As for pytest I've thought that processes/tests being run by pytest then might be able to get the used config file location, but have not thought much about a use case.

`$file/..` is a bit weird, but gets resolved fine via `os.path.{normpath,realpath}`.

So I am +0 myself on using the file - mainly to avoid having to add another env var later, and given that it is more explicit/precise.  But it has the downside of having to append "/.." for the more typical use case (this issue).

joybhallaa:
i'm taking this issue

gvanrossum:
PR was reverted (it broke CI).

PR #8574, opened 2020-03-22
Modified expand_path in config_parser to add environment var for mypy_path

### LINKED ISSUE #7967: Treat relative paths in config file relative to config file, not cwd
When using `mypy_path = src` it should be treated relative to the config file (if any), not the current working directory.
This also applies to the default of `cache_dir` (`.cache_dir`), which currently gets created in the current working directory always.

I.e. a detected/found (project) config file could be treated as the root for options with relative defaults also.

As for `mypy_path` this could be achieved in the type callback, which then however would need the config file / root as a new argument:
https://github.com/python/mypy/blob/84126ab9985844d8cda3809688034af895fdaf7c/mypy/config_parser.py#L75

As for global user config files, it should treat relative files therein relative to it, but it should not be used as a root for default settings like cache_dir.

ashutosh1919:
Possibly fixes #7967 and #7968 .
@JukkaL, @ethanhs and @blueyed , Please review this PR.

### DIFF
--- a/mypy/config_parser.py
+++ b/mypy/config_parser.py
@@ -40,7 +40,7 @@ def expand_path(path: str) -> str:
     """
 
     expanded_path = os.path.expandvars(os.path.expanduser(path))
-    if os.environ.get('MYPY_CONFIG_FILE_DIR')=='False':
+    if 'MYPY_CONFIG_FILE_DIR' not in os.environ:
         os.environ['MYPY_CONFIG_FILE_DIR'] = expanded_path
     else:
         os.environ['MYPY_CONFIG_FILE_DIR'] += (os.pathsep + expanded_path)
--- a/mypy/config_parser.py
+++ b/mypy/config_parser.py
@@ -113,9 +113,9 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
             continue
         try:
             if 'MYPY_CONFIG_FILE_DIR' not in os.environ:
-                os.environ['MYPY_CONFIG_FILE_DIR'] = config_file
+                os.environ['MYPY_CONFIG_FILE_DIR'] = os.path.dirname(config_file)
             else:
-                os.environ['MYPY_CONFIG_FILE_DIR'] += os.pathsep + config_file
+                os.environ['MYPY_CONFIG_FILE_DIR'] += os.pathsep + os.path.dirname(config_file)
             parser.read(config_file)
         except configparser.Error as err:
             print("%s: %s" % (config_file, err), file=stderr)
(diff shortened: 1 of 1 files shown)


mypy/config_parser.py
```diff
@@ -39,7 +39,12 @@ def expand_path(path: str) -> str:
     the provided path.
     """
 
-    return os.path.expandvars(os.path.expanduser(path))
+    expanded_path = os.path.expandvars(os.path.expanduser(path))
+    if 'MYPY_CONFIG_FILE_DIR' not in os.environ:
+        os.environ['MYPY_CONFIG_FILE_DIR'] = expanded_path
+    else:
+        os.environ['MYPY_CONFIG_FILE_DIR'] += (os.pathsep + expanded_path)
+    return expanded_path
```
blueyed: That looks wrong to me: `expand_path` appears to be a generic method.
It should probably be done here instead: https://github.com/python/mypy/pull/8574/files#diff-09533a8b2bebadfae430681eb15b01c9R120 - i.e. setting it to the current file before parsing it.

ashutosh1919: The link you gave above points to changed code itself. I don't understand your comment. Please, can you elaborate about where should the code be changed ?

blueyed: @ashutosh1919 
Needs to be expanded via clicking, sorry.
https://github.com/python/mypy/blob/00e8024ffd187bdbddeb71fcced53cd2912a9410/mypy/config_parser.py#L120
But that is just a guess.
The idea would be to define it before parsing a config file, then it gets / can be expanded.

ashutosh1919: @blueyed , Please take a look at it now.

ashutosh1919:
@blueyed , I am not able to re-request review here. Please review.

mypy/config_parser.py
```diff
@@ -112,6 +112,10 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
         if not os.path.exists(config_file):
             continue
         try:
+            if 'MYPY_CONFIG_FILE_DIR' not in os.environ:
+                os.environ['MYPY_CONFIG_FILE_DIR'] = config_file
```
ilevkivskyi: Why are you setting the _file_ name as a directory? Shouldn't you use something like `os.path.dirname()`?

ashutosh1919: Yes you are right. Thanks for noticing. I will change it in next commit.

ashutosh1919: done

ashutosh1919:
@ilevkivskyi , please review.

ilevkivskyi:
TBH, I don't totally understand the logic/reasoning behind this so I will leave it to someone else to review.

ashutosh1919:
cc @JukkaL

hauntsaninja:
Closing, since we merged #9414 to address the same issues

PR #8578, opened 2020-03-23
Add __dataclass_fields__ and __attrs_attrs__ to dataclasses

### LINKED ISSUE #6568: Recognizing wrapped classes as supporting a Protocol
```python
from typing import Dict

@dataclass
class A:
    pass

class IsDataclass(Protocol):
    __dataclass_fields__: Dict

def dataclass_only(x: IsDataclass):
    pass

dataclass_only(A())
```
When running mypy on this code, I'd expect no warning, but I get 

> test.py:16: error: Argument 1 to "dataclass_only" has incompatible type "A"; expected "IsDataclass". 

While checking for `__dataclass_fields__` might not look like the best thing to go for, it's what `dataclasses.is_dataclass` checks for as well, so this seems like the most reliable condition right now.

I'd assume it happens because the dataclass-wrapper adds this attribute, and mypy only checks the class definition itself. Can this be fixed?

tkukushkin:
Fixes #6568.


mypy/plugins/attrs.py
```diff
@@ -656,6 +658,20 @@ def _add_init(ctx: 'mypy.plugin.ClassDefContext', attributes: List[Attribute],
     adder.add_method('__init__', args, NoneType())
 
 
+def _add_attrs_magic_attribute(ctx: 'mypy.plugin.ClassDefContext') -> None:
+    attr_name = '__attrs_attrs__'
+    attr_type = ctx.api.named_type('__builtins__.tuple', [
+        ctx.api.named_type('attr.Attribute'),
```
tkukushkin: @ilevkivskyi `ctx.api.named_type` is for builtin types, but what i should use instead of it?

ilevkivskyi: It is not for builtin types, it is for types that are guaranteed to be present (i.e. loaded).

tkukushkin: @ilevkivskyi, and how to ensure that `attr.Attribute` is loaded? Or maybe need to use `Any` if it is not loaded?

ilevkivskyi: They should be loaded, because the `attr` module must be analyzed if this plugin is called. The tests fail because they run with partial stubs, not the real ones, see for example https://github.com/python/mypy/blob/master/test-data/unit/lib-stub/attr.pyi, `Attribute` is not present there.

Also the latter is generic, so I would explicitly add a type argument here (even if it is `Any`).

mypy/plugins/dataclasses.py
```diff
@@ -361,6 +363,21 @@ def _freeze(self, attributes: List[DataclassAttribute]) -> None:
                 var._fullname = info.fullname + '.' + var.name
                 info.names[var.name] = SymbolTableNode(MDEF, var)
 
+    def _add_dataclass_fields_magic_attribute(self) -> None:
+        attr_name = '__dataclass_fields__'
+        dataclass_field_type = self._ctx.api.named_type('dataclasses.Field')
```
ilevkivskyi: Same comment here.

mypy/plugins/attrs.py
tkukushkin: > They should be loaded, because the attr module must be analyzed if this plugin is called.

I have a problem with different styles of import of `attr`.

This example works fine:
```python
import attr

@attr.s
class Foo:
    pass

class AttrsProtocol(Protocol):
    __attrs_attrs__: Tuple[attr.Attribute, ...]

def foo(x: AttrsProtocol) -> None:
    pass

foo(Foo())
```
But if I change import to:
```python
from attr import attrs, Attribute

@attrs
class Foo:
    pass
```
I get exception:
```
  File "/Users/tkukushkin/Projects/mypy/mypy/plugins/attrs.py", line 664, in _add_attrs_magic_attribute
    ctx.api.named_type('attr.Attribute'),
  File "/Users/tkukushkin/Projects/mypy/mypy/semanal.py", line 4143, in named_type
    assert sym, "Internal error: attempted to construct unknown type"
AssertionError: Internal error: attempted to construct unknown type
```
Some tests fail because of this.

> The tests fail because they run with partial stubs, not the real ones, see for example https://github.com/python/mypy/blob/master/test-data/unit/lib-stub/attr.pyi, `Attribute` is not present there.

Yes, I already added `Attribute` to stub, but not pushed yet.

ilevkivskyi: Well, I just looked at the code of `named_type()` and it is horrible, it uses `lookup_qualified()` (which is just a normal symbol lookup following Python name resolution rules) so I wonder how it ever worked. You can try using `named_type_or_none()` instead, that one looks reasonable.

Another option is to actually fix `named_type()`, but it may be not the best task as a first contribution.

tkukushkin:
@ilevkivskyi Thanks for your suggestions. `named_type_or_none` works fine, but `ClassDefContext.api` has type `SemanticAnalyzerPluginInterface` which hasn't member `named_type_or_none`. Should I add this method to interface? 

And second question: how do you recommend to fix tests `testNarrowingParentWithEnumsBasic` and `testNarrowingParentWithIsInstanceBasic`. They use `fixtures/tuple.pyi` and `fixtures/isinstance.pyi` respectively. These partial stubs don't contain `dict`, which I use in `_add_dataclass_fields_magic_attribute`. Should I create special partial stubs for this tests? Or may be need to extend existing stubs by simple dict definition (`class dict(Generic[_KT, _VT]): pass`)?

test-data/unit/check-dataclasses.test
```diff
@@ -390,9 +389,10 @@ app1 = Application("example-1", 5)
 app2 = Application("example-2", 5)
 app1 == app2
 app1 != app2
-app1 == None  # E: Unsupported operand types for == ("Application" and "None")
+app1 == None
+reveal_type(app1.__eq__)  # N: Revealed type is 'def (builtins.object) -> builtins.bool'
```
tkukushkin: @ilevkivskyi maybe this test is not valid. Possible it passes on master because of incomplete `object` definition in fixtures/list.pyi stub, in dataclasses.pyi `object` has `__eq__` and `__ne__` methods.

Example:
```python

@dataclass
class A:
    pass

reveal_type(A.__eq__)
```
On latest version of mypy output is:
```
test.py:7: note: Revealed type is 'def (builtins.object, builtins.object) -> builtins.bool'
```
Maybe this code doesn't work: https://github.com/python/mypy/blob/master/mypy/plugins/dataclasses.py#L127
And I don't understand, why `__eq__` has this type: `(self: T, other: T) -> bool`. Maybe need to create new issue for this?

ilevkivskyi:
@tkukushkin Thanks for updates! Here are answers to your questions:

> Should I add this method to interface?

I think it would make sense to add it.
> Should I create special partial stubs for this tests, or split these tests on several small tests with different objects, or extend existing stubs by simple dict definition

I think creating dedicated partial stubs (fixtures) is fine.
> maybe this test is not valid.

Good catch! Yeah, I think this test passes because of bad partial stubs. In real stubs `object` always has `__eq__()` so the code you linked is wrong, it should read `info.names.get('__eq__') is None` instead of `info.get('__eq__') is None` (to avoid MRO lookup and only check current class).

You can open a separate issue/PR for this.

tkukushkin:
Fixed `__eq__` in #8642

mypy/plugin.py
```diff
@@ -252,6 +252,13 @@ def named_type(self, qualified_name: str, args: Optional[List[Type]] = None) ->
         """Construct an instance of a builtin type with given type arguments."""
         raise NotImplementedError
 
+    @abstractmethod
+    def named_type_or_none(self,
+                           qualified_name: str,
+                           args: Optional[List[Type]] = None) -> Optional[Instance]:
+        """Construct an instance of a builtin type with given type arguments."""
```
JelleZijlstra, line 259: Docstring is the name as `named_type`, can you say what's different about this method?

mypy/plugins/attrs.py
```diff
@@ -709,6 +711,22 @@ def _add_init(ctx: 'mypy.plugin.ClassDefContext', attributes: List[Attribute],
     adder.add_method('__init__', args, NoneType())
 
 
+def _add_attrs_magic_attribute(ctx: 'mypy.plugin.ClassDefContext') -> None:
+    attr_name = '__attrs_attrs__'
+    any_type = AnyType(TypeOfAny.explicit)
+    attribute_type = ctx.api.named_type_or_none('attr.Attribute', [any_type]) or any_type
+    var = Var(name=attr_name, type=ctx.api.named_type('__builtins__.tuple', [
```
JelleZijlstra, line 718: Looks like this generates `Tuple[attr.Attribute[Any], ...]`. Since you know what the fields are, shouldn't it be possible to generate the precise type for each attribute?

JelleZijlstra (changes requested):
A few small comments, and CI is currently failing.

mypy/plugin.py
tkukushkin: I'm afraid I myself do not know what the difference between these methods is, @ilevkivskyi advised (https://github.com/python/mypy/pull/8578#discussion_r399578066) to use `named_type_or_none` and I had to add it to `SemanticAnalyzerPluginInterface`.

mypy/plugins/attrs.py
tkukushkin: Maybe it is possible, but I don’t think that it is necessary to solve the original problem. It can be implemented later in another MR without breaking changes.

mypy/plugin.py
EdwardLarson: suggest docstring:

```
"""Construct an instance of a type with given type arguments, or None if a type could not be constructed for the qualified type name. This is possible when the qualified name includes a module name and the module has not been imported."""
```
From looking through the code it seems `named_type_or_none` is useful when `qualified_name` is NOT a built-in type. Things like `typing._TypedDict`, `abc.ABCMeta`, etc. whereas `named_type` is almost exclusively used for `__builtins__.*` or `builtins.*` (the one exception is `m.Signal`). So my assumption is that `named_type_or_none` is useful when the qualified type name is includes a module name which may not be known. This matches with the reason why `named_type_or_none` was added to `SemanticAnalyzerPluginInterface` in this PR in the first place - exception when importing as `from attr import attrs, Attribute` but not when `import attrs` and referring to `attrs.Attribute`.

However since I am not an expert on mypy internals, and the `named_type_or_none` method does not have a docstring in `SemanticAnalyzerInterface` I'm afraid I'm rather underqualified to make the docstring and can't be certain about my interpretations.

mypy/plugins/attrs.py
EdwardLarson: Does it generate `Tuple[attr.Attribute[Any], ...]` or `Tuple[attr.Attribute[Any]]`? I don't see how the difference is specified in the mypy API, though there obviously is a difference and mypy does handle ellipses. 

Anyway, I believe the precise attribute types could be generate by changing the following lines:
```
308: _add_attrs_magic_attribute(ctx, [node.type for attr in attributes])
714: def _add_attrs_magic_attribute(ctx: 'mypy.plugin.ClassDefContext', raw_attr_types: 'List[mypy.types.Type]') -> None:
718: attribute_types = [attr_type if attr_type is not None else any_type for attr_type in raw_attr_types]
719: var = Var(name=attr_name, type=ctx.api.named_type('__builtins__.tuple', attribute_types))
```

Sort of sloppy formatting but hopefully gets the picture across. The checks for `ctx.api.final_iteration` and `ctx.api.defer` calls are completed by the time the call to `_add_attrs_magic_attribute`, so it looks safe to check for `None` and substitute `any_type` at that point.

tkukushkin: @EdwardLarson I changed the method to generate the exact types for the attributes. Thank you for your help.

test-data/unit/foo.py
```diff
@@ -0,0 +1,22 @@
+from typing import Tuple
```
JelleZijlstra, line 1: Is this a stray file?

JelleZijlstra (changes requested):
Looks good apart from the stray file

test-data/unit/foo.py
tkukushkin: Oh, sorry, deleted.

JelleZijlstra (approved):
Thanks! I'll merge once CI passes (if I don't, feel free to remind me).

eviltwin:
I really don't want to be "that guy" and ask when this will see a release... but do you have published release cadences? This has been merged for a month now and the last release was in June. I'm not trying to phrase this in an "omg what's the hold up?" kind of way - I totally get that other things can be blocking the release.

It's just that I happened across an issue this morning whilst consuming [djangorestframework-dataclasses](https://github.com/oxan/djangorestframework-dataclasses) that would be solved by this feature, so I'm a little itchy to see it released :)

JukkaL:
@eviltwin Tracking issue for the next release: #11158

Tinche:
@tkukushkin Thanks a lot for this, I think this work is going to be very valuable going forward and I can close my failed attempt at https://github.com/python/mypy/pull/10467. How difficult would it be to have `attr.fields(A)` actually return `A.__attrs_attrs__`? That API is much nicer than the dunder (although the dunder should stay for protocol checking etc).

tkukushkin:
@Tinche I spent some time and got such a proof of concept:

```python
class AttrsFieldsPlugin(Plugin):

    def get_function_signature_hook(self, fullname: str) -> Optional[Callable[[FunctionSigContext], FunctionLike]]:
        if fullname == 'attr.fields':
            return self._attrs_fiels_hook
        return None

    def _attrs_fiels_hook(self, context: FunctionSigContext) -> FunctionLike:
        return CallableType(
            arg_types=[AnyType(TypeOfAny.explicit)],
            arg_kinds=[ArgKind.ARG_POS],
            arg_names=['cls'],
            ret_type=context.args[0][0].node.names['__attrs_attrs__'].type,
            fallback=context.api.named_type('builtins.function'),
        )
```

Unfortunately, I don't have time to finalize this solution (need to correct the type of the function argument and deal with the situation when not attrs class passed to the `attrs.fields`)

I also understand attrs.fields returns named tuple, in this solution the function returns regular tuple, and I'm not sure if there is a way to fix that.

Tinche:
@tkukushkin 

Hello again! Thanks for looking into this.

I think `fields` doesn't need to be handled by the plugin at all. I've created a PR for attrs adding a protocol and making fields work out of the box: https://github.com/python-attrs/attrs/pull/890

The problem isn't fields, but as you said the fact the plugin generates tuples instead of namedtuples for `__attrs_attrs__`. So the Mypy plugin needs to be improved to generate a namedtuple subclass for each attrs class it sees and assign that to `__attrs_attrs__`. I will attempt this but since I'm not super knowledgeable about Mypy would appreciate pointers ;)

In any case, thanks for getting us this far.

tkukushkin:
@Tinche I looked at your pull request – beautiful solution, didn't think of that. But yes, I don't really know how to make `__attrs_attrs__` become namedtuple :-(.

PR #8583, opened 2020-03-24
Implement support for returning TypedDict for dataclasses.asdict

syastrov:
There are a couple of limitations that would be nice to remove, but I am hoping this is enough for an initial version.

Subclasses of list/dict/tuple are transformed into the base class rather than a new version of the subclass.
This is because the transformation happens in the `get_function_hook` plugin hook which only has access to the `CheckerPluginInterface` and so can't (without hacking) add things to the symbol table.
Besides that, there could be violations of variance/constraints if the new type is just constructed without being checked.

Also, `NamedTuple`s found within dataclasses are transformed into `Any`.
This is because the new namedtuple needs a partial_fallback generated/added to the symbol table (e.g. same problem as the above).
Supporting generating new `NamedTuple`s properly would probably require refactoring `NamedTupleAnalyzer.build_namedtuple_typeinfo` to make it reusable.

As far as I know, I was looking for some way of somehow adding new types to the symbol table from `CheckerPluginInterface` and then hoping that I can trigger a new type-checking pass so that variance/constraints are checked on the newly added types. But I'm sure there's a better way to do it.

Relates to #5152


syastrov:
I had previously made a "draft" PR at #8339. This supersedes that.

This should be ready for review now.

mypy/plugins/dataclasses.py
```diff
@@ -386,3 +395,100 @@ def _collect_field_args(expr: Expression) -> Tuple[bool, Dict[str, Expression]]:
...
+        if isinstance(typ, UnionType):
+            return UnionType([_asdictify_inner(item, seen_dataclasses) for item in typ.items])
+        if isinstance(typ, Instance):
+            info = typ.type
+            if is_type_dataclass(info):
+                if info.fullname in seen_dataclasses:
+                    api.fail(
+                        "Recursive types are not supported in call to asdict, so falling back to "
+                        "Dict[str, Any]",
+                        context)
```
ilevkivskyi: I think there is no need to show an error here, just silently give an imprecise type.

mypy/plugins/dataclasses.py
```diff
@@ -386,3 +395,100 @@ def _collect_field_args(expr: Expression) -> Tuple[bool, Dict[str, Expression]]:
...
+                    # At least for now, typeshed stubs for asdict don't allow you to pass in
+                    # `dict` as dict_factory, so we can't special-case that.
+                    return ctx.default_return_type
+
+    ctx.api.fail("'dataclasses.asdict' should be called on dataclass instances",
+                 ctx.context)
+    return ctx.default_return_type
+
+
+def _transform_type_args(*, typ: Instance, transform: Callable[[Instance], Type]) -> List[Type]:
```
ilevkivskyi: It seems to me all these dances with callbacks and two mutually recursive functions can be avoided (and the whole logic made more straightforward) with a small `TypeTransformVisitor`.

mypy/plugins/dataclasses.py
```diff
@@ -386,3 +395,100 @@ def _collect_field_args(expr: Expression) -> Tuple[bool, Dict[str, Expression]]:
...
+                    attr = DataclassAttribute.deserialize(info, data, api)
+                    sym_node = info.names[attr.name]
+                    attr_type = sym_node.type
+                    assert attr_type is not None
+                    fields[attr.name] = _asdictify_inner(attr_type, seen_dataclasses)
+                return make_anonymous_typeddict(api, fields=fields,
+                                                required_keys=set())
+            elif info.has_base('builtins.list'):
+                supertype_instance = map_instance_to_supertype(typ, api.named_generic_type(
+                    'builtins.list', []).type)
```
ilevkivskyi: Creating instance with wrong type argument count (event for very short time) is a bad idea. Use dummy `Any` here.

mypy/plugins/dataclasses.py
```diff
@@ -386,3 +395,100 @@ def _collect_field_args(expr: Expression) -> Tuple[bool, Dict[str, Expression]]:
...
+                new_args = _transform_type_args(
+                    typ=supertype_instance,
+                    transform=lambda arg: _asdictify_inner(arg, seen_dataclasses))
+                return api.named_generic_type('builtins.list', new_args)
+            elif info.has_base('builtins.dict'):
+                supertype_instance = map_instance_to_supertype(typ, api.named_generic_type(
+                    'builtins.dict', []).type)
```
ilevkivskyi: Same here.

test-data/unit/check-dataclasses.test
```diff
@@ -1032,3 +1032,275 @@ class B(A):
...
+class Course:
+    staff: Staff
+
+instance = Course(staff=Staff(teacher=Person("Joe", 32), assistant=Person("John", 23)))
+result = asdict(instance)
+# Due to implementation limitations, namedtuples are transformed to Any
+reveal_type(result['staff']) # N: Revealed type is 'Any'
+
+[typing fixtures/typing-full.pyi]
+[builtins fixtures/tuple.pyi]
```
ilevkivskyi: I would add couple fine grained (daemon) test cases. The tricky part is that when type of a dataclass attribute changes, then the call to `asdict()` should be reprocessed (put it in a different module). I am not 100% sure we already have enough extra dependencies added by the plugin to guarantee this, so such tests will be really useful.

mypy/plugins/common.py
```diff
@@ -155,8 +156,26 @@ def add_method_to_class(
...
     return typ
+
+
+def get_anonymous_typeddict_type(api: CheckerPluginInterface) -> Instance:
+    for type_fullname in TPDICT_FB_NAMES:
+        try:
+            anonymous_typeddict_type = api.named_generic_type(type_fullname, [])
+            if anonymous_typeddict_type is not None:
+                return anonymous_typeddict_type
+        except KeyError:
```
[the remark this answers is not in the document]
syastrov: You're right that it's not needed.

mypy/plugins/dataclasses.py
```diff
@@ -386,3 +395,100 @@ def _collect_field_args(expr: Expression) -> Tuple[bool, Dict[str, Expression]]:
...
         return True, args
     return False, {}
+
+
+def asdict_callback(ctx: FunctionContext) -> Type:
+    positional_arg_types = ctx.arg_types[0]
+
+    if positional_arg_types:
+        dataclass_instance = positional_arg_types[0]
+        dataclass_instance = get_proper_type(dataclass_instance)
```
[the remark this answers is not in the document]
syastrov: Makes sense.

mypy/plugins/dataclasses.py
syastrov: 👍

mypy/plugins/dataclasses.py
syastrov: I agree that it was too complex for what it did: I removed the function with the callback.
I think it's reasonable now, just one recursive function. I'm not sure how I would use `TypeTransformVisitor` here though. If you think it would help and could give some pointers, then I'd try and use it.

mypy/plugins/dataclasses.py
syastrov: Makes sense. Out of curiosity, is there any reason why this is not checked in an assertion? 
Looks like the code for `named_generic_type` has a TODO comment:
```
        # TODO: assert len(args) == len(info.defn.type_vars)
```
Is it because of performance reasons?

mypy/plugins/common.py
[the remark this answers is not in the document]
syastrov: I guess if `warn_unreachable` was enabled on mypy self-check, this would have been caught. But that's another issue.

test-data/unit/check-dataclasses.test
syastrov: I added one. It revealed that there was no dependency on the fallback type for TypedDict. I tried to add that. Since it depends on python version which TypedDict fallback to use (I think?), I had to add a dependency on all modules (e.g. typing/typing_extensions/mypy_extensions) to the default plugin. Doesn't seem very nice, but I'm not sure of the right way to do it. Do you have any ideas?

Do you have any ideas for more fine-grained tests to add?

syastrov:
Thanks for the review @ilevkivskyi. I've addressed your comments.

Also, I'm not sure if it makes sense to make the TypedDicts total or not (I've left them non-total because it's more lenient).

syastrov:
Oops, looks like tests got messed up.. I'll have a look

mypy/plugins/dataclasses.py
syastrov: Ah, I see it can be done with `TypeTranslator`. I've made the refactor. It does seem more straightforward once I understood how this TypeVisitor stuff worked.

syastrov:
I decided to make `asdict` return `TypedDict` only for Python >= 3.8, since `TypedDict` has only become standard since that version. I think it makes sense, but it also solves this problem: If I were to try to include a dependency on `mypy_extensions` or `typing_extensions` in `get_additional_deps`, but they were not installed, then the user would get an error about a missing dependency (besides that, there was an issue with using `mypy_extensions` that could have meant I would have had to add more fixtures to a lot of unit tests, since it was indirectly using `tuple`).

I also changed the `TypedDict` to be `total`. I figure that most use-cases would not be deleting keys from the `TypedDict` returned by `asdict`, so it's useful to know that the keys are there (as they will be to start with). But if the user wants to modify the dict, of course it's possible with a type-ignore or cast to ordinary `dict`. I hope this makes sense.

Pending your re-review of course, as far as I can see, the only thing missing is some more fine-grained tests (though I'm unsure what makes sense to test there).

Please let me know if there are any other issues :)

test-data/unit/fine-grained-dataclasses.test
```diff
@@ -0,0 +1,39 @@
...
+@dataclass
+class MyDataclass:
+    attr: AttributeInOtherModule
+
+[file b.py]
+AttributeInOtherModule = str
+[file c.py]
+from dataclasses import asdict
+from a import MyDataclass
+reveal_type(asdict(MyDataclass('John')))
```
ilevkivskyi, line 21: Remove `reveal_type()` from the first pass, it must not emit messages. Instead add an operation that would pass with the initial type, but would fail with the updated type. For example, `asdict(my_var)['attr'] + "foo"`.

test-data/unit/fine-grained-dataclasses.test
```diff
@@ -0,0 +1,39 @@
...
+
+@dataclass
+class MyDataclass:
+    attr: AttributeInOtherModule
+
+[file b.py]
+AttributeInOtherModule = str
+[file c.py]
+from dataclasses import asdict
+from a import MyDataclass
```
ilevkivskyi, line 20: Don't import the class itself, import a variable with this type from `a`, say `my_var: MyDataclass`.

test-data/unit/fine-grained-dataclasses.test
```diff
@@ -0,0 +1,39 @@
...
+reveal_type(asdict(MyDataclass('John')))
+
+[file b.py.2]
+from typing import List
+class MyList(List[int]):
+    pass
+AttributeInOtherModule = MyList
+[file c.py.2]
```
ilevkivskyi, line 28: This change is not needed.

test-data/unit/fine-grained-dataclasses.test
```diff
@@ -0,0 +1,39 @@
+-- Test cases for fine-grained incremental checking of dataclasses
```
ilevkivskyi, line 1: Maybe make this file more generic, like `fine-grained-plugins.test`? Or just move this test case to the main `fine-grained.test` file.

mypy/plugins/default.py
```diff
@@ -20,16 +20,26 @@
 class DefaultPlugin(Plugin):
     """Type checker plugin that is enabled by default."""
 
+    def get_additional_deps(self, file: MypyFile) -> List[Tuple[int, str, int]]:
+        if self.python_version >= (3, 8):
+            # Add module needed for anonymous TypedDict (used to support dataclasses.asdict)
+            return [(10, "typing", -1)]
```
ilevkivskyi, line 26: Why this is needed? I think `typing` module is always loaded. It may be needed to add only dependency on `typing_extensions`, which is available on all versions.

mypy/plugins/default.py
```diff
@@ -20,16 +20,26 @@
...
 
         if fullname == 'contextlib.contextmanager':
             return contextmanager_callback
         elif fullname == 'builtins.open' and self.python_version[0] == 3:
             return open_callback
         elif fullname == 'ctypes.Array':
             return ctypes.array_constructor_callback
+        elif fullname == 'dataclasses.asdict':
+            return partial(dataclasses.asdict_callback,
+                           return_typeddicts=self.python_version >= (3, 8))
```
ilevkivskyi, line 42: I think this is not needed, see my comment about `typing_extensions`.

mypy/plugins/common.py
```diff
@@ -155,8 +157,24 @@ def add_method_to_class(
...
 ) -> Type:
     typ = deserialize_type(data)
     typ.accept(TypeFixer(api.modules, allow_missing=False))
     return typ
+
+
+def get_anonymous_typeddict_type(api: CheckerPluginInterface) -> Instance:
+    for type_fullname in TPDICT_FB_NAMES:
+        try:
+            return api.named_generic_type(type_fullname, [])
```
ilevkivskyi, line 171: This is actually not the best method for this purpose, it will fail if `typing_extensions` is not imported in the current module. Instead, you need to use something like `named_type_or_none()`, see e.g. fallback code in `build_typeddict_typeinfo()` in `semanal_typeddict.py`. It is however not available in `TypeChecker`, so maybe you can try to copy it from `SemanticAnalyzer`.

docs/source/additional_features.rst
```diff
@@ -72,9 +72,13 @@ and :pep:`557`.
 Caveats/Known Issues
 ====================
 
-Some functions in the :py:mod:`dataclasses` module, such as :py:func:`~dataclasses.replace` and :py:func:`~dataclasses.asdict`,
+Some functions in the :py:mod:`dataclasses` module, such as :py:func:`~dataclasses.replace`,
 have imprecise (too permissive) types. This will be fixed in future releases.
 
+In Python version 3.8 and above, calls to :py:func:`~dataclasses.asdict` will return a ``TypedDict`` based on the
```
ilevkivskyi, line 78: Again, the version limitation is not needed, see my comment about `typing_extensions`.

syastrov:
Thanks @ilevkivskyi for the re-review.

I tried to address your comments (thanks, I now think I understand how the fine-grained test works). I have no idea if the code copied from SemanticAnalyzer regarding `named_type_or_none` is logically correct in the TypeChecker context, but I'll take your word for it.

I could not get the new fine-grained test of `asdict` to pass. I tried to debug it a bit and the `asdict` function hook is called twice. But it's also called twice when I remove the modification to c.py from the test. So it seems like it should be called at least 3 times, and there's some missing dependency that prevents it from being called again? Any tips how to proceed?

mypy/plugins/default.py
syastrov: It seems adding that dependency on `typing_extensions` means a lot of test fixtures need updating (it's the `tuple` builtin missing, which I believe is transitively imported from `typing_extensions`).

JukkaL:
If something doesn't get reported properly in fine-grained incremental mode, this is often related to missing fine-grained dependencies. We have test cases in `test-data/unit/deps.test` that deal with these. You can try adding a test case there and somebody from the core team can review the dependencies if it's not clear if they are correct.

mypy/checker.py
```diff
@@ -4538,6 +4538,21 @@ def named_type(self, name: str) -> Instance:
         any_type = AnyType(TypeOfAny.from_omitted_generics)
         return Instance(node, [any_type] * len(node.defn.type_vars))
 
+    def named_type_or_none(self, qualified_name: str,
+                           args: Optional[List[Type]] = None) -> Optional[Instance]:
+        sym = self.lookup_qualified(qualified_name)
+        if not sym or isinstance(sym.node, PlaceholderNode):
```
ilevkivskyi, line 4544: No need to check for placeholders, they should never appear at this stage.

syastrov:
Thanks for all your time reviewing -- II'm going to try to pick this up again.

@JukkaL wrote:
> If something doesn't get reported properly in fine-grained incremental mode, this is often related to missing fine-grained dependencies. We have test cases in `test-data/unit/deps.test` that deal with these. You can try adding a test case there and somebody from the core team can review the dependencies if it's not clear if they are correct.

Thanks for the pointer. I am not sure from which module it makes sense to try and output dependencies for.

I tried with this example:

```python
[case testDataclassAsdictDeps]
# flags: --python-version 3.8
from b import AttributeInOtherModule

@dataclass
class MyDataclass:
    attr: AttributeInOtherModule

my_var: MyDataclass

[file b.py]
AttributeInOtherModule = str

[typing fixtures/typing-typeddict.pyi]
[builtins fixtures/fine_grained.pyi]

[out]
```

Output is:
```
  <m.MyDataclass> -> <m.my_var>, m, m.MyDataclass (diff)
  <m.my_var> -> m                               (diff)
  <b.AttributeInOtherModule> -> m               (diff)
  <b> -> m                                      (diff)
  <dataclasses.dataclass> -> m                  (diff)
  <dataclasses> -> m                            (diff)
```

Shouldn't b.AttributeInOtherModule affect m.MyDataclass? Or am I misinterpreting how this should work?

mypy/checker.py
syastrov: Removed the check

mypy/checker.py
```diff
@@ -4538,6 +4538,21 @@ def named_type(self, name: str) -> Instance:
         any_type = AnyType(TypeOfAny.from_omitted_generics)
         return Instance(node, [any_type] * len(node.defn.type_vars))
 
+    def named_type_or_none(self, qualified_name: str,
+                           args: Optional[List[Type]] = None) -> Optional[Instance]:
+        sym = self.lookup_qualified(qualified_name)
```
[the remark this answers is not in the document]
syastrov: I guess I misunderstood your comment. Not really knowing how the checker or semantic analyzer work internally, I don't really think I am qualified to make this change. However, I will try.

I have copied the lookup_fully_qualified_or_none function to checker (and only modified it to raise KeyError if the result is None) and used it in `named_type_or_none`.
I am not sure if this makes sense and whether the docstring/TODO comments should be altered/removed.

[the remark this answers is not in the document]
syastrov: Oops, late-night coding mistake :) I removed the part about raising KeyError, and it returns None now instead if it can't find the name of course.

syastrov:
I made another experiment with outputting deps.

```python
[case testDataclassesAsdictDeps]
# flags: --python-version 3.8
from dataclasses import asdict
from a import my_var
x = my_var
x['attr'] + "foo"

[file a.py]
from b import AttributeInOtherModule

@dataclass
class MyDataclass:
    attr: AttributeInOtherModule

my_var: MyDataclass

[file b.py]
AttributeInOtherModule = str

[typing fixtures/typing-typeddict.pyi]
[builtins fixtures/fine_grained.pyi]

[out]
```

Output:
```
Actual:
  main:5: error: Value of type "MyDataclass" is not indexable (diff)
  <m.x> -> m                                    (diff)
  <a.MyDataclass.__getitem__> -> m              (diff)
  <a.MyDataclass> -> <m.x>                      (diff)
  <a.my_var> -> m                               (diff)
  <a> -> m                                      (diff)
  <dataclasses.asdict> -> m                     (diff)
  <dataclasses> -> m                            (diff)
```

Here, I think it makes sense that `<a.MyDataclass> -> <m.x>` appears.

But when changing the code to
```python
x = asdict(my_var)
```

the output is:

```
  <m.x> -> m                                    (diff)
  <a.my_var> -> m                               (diff)
  <a> -> m                                      (diff)
  <dataclasses.asdict> -> m                     (diff)
  <dataclasses> -> m                            (diff)
```

After trying to add an `add_plugin_dependency` method to the checker, I seem to get one of the fine-grained tests to pass, but not the other (FineGrainedCacheSuite is the one passing now).

My guess is that because the dataclass attributes are loaded from the cache, and the failing test is testing incremental checking when there is no cache, which fails of course because the cache is empty. However, I'm not sure what to do about it. Populating the dataclass attribute metadata is done during semantic analysis by the dataclass plugin as part of a class decorator hook. I would like to somehow add a dependency to get that phase to run again. I tried adding a dependency on both the dataclass itself and a wildcard trigger on the module defining the dataclass, but that does not seem to help.

I feel like I am getting very close to getting this working.
Any pointers would be much appreciated if you @ilevkivskyi or @JukkaL have the time.
Thanks :)

syastrov:
@JukkaL or @ilevkivskyi 
After a lot of head-banging and re-reading your comments, I finally got this working! The fine-grained test passes both with/without cache. I had to add a plugin dependency on each attr in the dataclass. A wildcard trigger on the dataclass did not help.

Would one of you be able to review again?


---

I do have some reservations about the impact changing this might have, and wonder what you think about it.

For example, a lot of code might do something like:
```python
from dataclasses import dataclass, asdict

@dataclass
class Foo:
  bar: str

def serialize_foo(foo: Foo) -> dict:
  asdict(foo)
```

It'll fail since TypedDict is not compatible with dict:
> error: Incompatible return value type (got "Foo", expected "Dict[Any, Any]")  [return-value]

It's easily fixed by changing it to return `Mapping`.

It's probably more annoying if you mutate the returned dict.

```python
x = asdict(foo)
x['some_other_attr'] = 123 # error
```

You'd have to cast it to `dict` or `MutableMapping` I'd assume to get this working.

Do you see this as an issue? Is it enough to warrant making a flag for asdict returning TypedDict, disabled by default?

It's just that this kind of churn can be annoying for users who don't actually want to take advantage of this feature.

Thanks again! :)

syastrov:
Ping

syastrov:
To reduce false-positives, I've removed the part about raising an error if the instance is not detected as a dataclass by the plugin.

97littleleaf11:
I fixed the test errors caused by `lookup` functions, which had been refactored several month ago. However the mypy primer results is confusing. I just skip the unfound sym node as a workaround.

97littleleaf11:
The incompatible cases with `TypedDict` looks like false-positive.

syastrov:
@ilevkivskyi Thank you for the review 👍 
@97littleleaf11 Thanks so much for bringing this up to date/fixing issues :)

The stack trace in the mypy primer result mentions the key `upnp` which seems to be used in the class `_UpnpServiceDescription` which is inherited from by another dataclass (`SsdpServiceInfo`) along with other classes. Perhaps cause is that the (multiple) inheritance is not handled correctly? (https://github.com/home-assistant/core/blob/67c35652f0d49dc56e1e934fab3b2c51f5f82592/homeassistant/components/ssdp/__init__.py#L94)?
Do you think making a proper fix for this is a blocker for this PR to be merged?

Regarding false positives: Despite providing more strictness, this PR introduces some false-positives to existing code, which either uses an overly-precise type `Dict` or mutates the resulting dict. In the case of code which mutates the resulting dict, you could `cast` to `dict` or `MutableMapping` as a workaround, or rework the code to construct a new `dict` based on the result from `asdict` (optionally, using a separately defined `TypedDict` for more precision). It's arguably very few false positives, though.

97littleleaf11:
> Perhaps cause is that the (multiple) inheritance is not handled correctly?

Could you please add a minimal reproduce if possible?

> Do you see this as an issue? Is it enough to warrant making a flag for asdict returning TypedDict, disabled by default?

> cast to dict or MutableMapping as a workaround

Yeah it's a typical solution. `TypedDict` is designed to be not comparible with `Dict[str, Any]` in mypy. My concern is that `asdict` currently is already typed as `Dict[str, Any]` as well as other real-world projects. cc @JukkaL

zerlok:
I think it's a good feature for transforming one dataclass to another with simple line `DataclassTwo(**asdict(DataclassOne(...)))` that still support typing.

P.S. In my case, I would like to tell DataclassTwo constructor to ignore unknown fields from DataclassOne. I'll try to make another feature request for that.

Issue #6406, opened 2019-02-15
Disallow tuple unpacking for strings


msullivan:
I think it would be a good idea to disallow tuple unpacking from strings, so that the following is caught:
```
d: Dict[str, str]
for a, b in d:
    stuff
```

There are a lot of cases involving one or the other of string iteration and tuple unpacking that we probably can't solve without risking too many false positives, but this one should be pretty harmless.

lladhibhutall:
Hi @msullivan I would like to take a stab at it, I am new to mypy so if you could give a little guidance.

msullivan:
I think you'll want to add some sort of special case to `check_multi_assignment` in `checker.py`

lladhibhutall:
@msullivan can you explain the example? I get it that mypy won't catch this exception but couldn't understand the tuple unpacking of string part.

JukkaL:
Here is a another example:

```py
s = "foo"
a, b = s
```

The tuple unpacking refers to the lvalue (`a, b` in the above example). If this is still unclear, it would probably be useful to study how assignment statements generally work in Python in some detail.

asmeurer:
What is the mypy policy on false positives? This rule seems to restrict what is allowed for no good reason. Something like `a, b, c = 'abc'` is perfectly fine, and much better than `a, b, c = list('abc')` or `a, b, c = 'a', 'b', 'c'` in terms of readability.

pearu:
Notice that if one has
```
d: Dict[Tuple[int, int], str]
for a, b in d:
    stuff
```
then the same reasoning would apply as in the description of this issue for disallowing tuple unpacking of tuples, that is,
```
a, b = 1, 2
```
would raise an exception. I think this would be undesired.

I think the fix to this issue is wrong because a Python string is a sequence and it should be treated as any other sequence in Python when applying indexing operations (that unpacking is).
The correct fix would have been to reject the idea of disallowing unpacking of strings as it deviates from standard Python behaviour, IMHO.

AustinScola:
Here is an example I just ran into where mypy reporting this error was helpful to me:
```Python3
head, *tail = 'abc'
```
I expected to get `head == 'a'` and `tail == 'bc'`. Instead, I was surprised to find that `tail == ['b', 'c']`.

(Another error for tail is also reported here).

The reason I bring up this example is just because I don't see one like it on this thread.

asmeurer:
[...]

PR #8589, opened 2020-03-25
Implementation for disallowing tuple unpacking of strings

akerami:
Implementation of #6406. 

Tests are failing, but /i believe the fix isn't too far from the implementation I have so far.

I would really appreciate suggestions on a different way of checking the type of a string so that errors do not occur with the tests.

Thanks!

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2553,7 +2553,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         elif isinstance(rvalue_type, UnionType):
             self.check_multi_assignment_from_union(lvalues, rvalue, rvalue_type, context,
                                                    infer_lvalue_type)
-        elif rvalue_type.type.fullname == 'builtins.str':
+        elif isinstance(rvalue_type, Instance) and rvalue_type.type.fullname == 'builtins.str':
             self.msg.unpacking_strings_disallowed(context)
         else:
             self.check_multi_assignment_from_iterable(lvalues, rvalue_type,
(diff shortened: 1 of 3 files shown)


mypy/checker.py
```diff
@@ -2553,6 +2553,8 @@ def check_multi_assignment(self, lvalues: List[Lvalue],
         elif isinstance(rvalue_type, UnionType):
             self.check_multi_assignment_from_union(lvalues, rvalue, rvalue_type, context,
                                                    infer_lvalue_type)
+        elif rvalue_type.type.fullname == 'builtins.str':
```
JukkaL: You need to protect this with an `isinstance` check.

akerami: Yeah it was my original thought to match on `isinstance(rvalue_type, str)` but then this conditional branch will not be triggered even though `rvalue_type.type.fullname == 'builtins.str' ` does get triggered.

msullivan: Protected with an `isinstance(rvalue_type, Instance)` check, probably. Note that checking mypy with itself produces a type error now.

akerami: I see thanks, I made the change, it made most of the tests pass for me, except a few cases where the actual output of a test included the error message which made the tests fail due to a mismatch.

akerami:
So now that the checks have passed should I write unit tests in test-data?

msullivan:
Yes please.

akerami:
Not sure where to place my tests, should I place it right under `[case testUnionOfNonIterableUnpacking]` in `check-unions.test`?

msullivan:
That sounds good, yeah.

akerami:
Okay are there any further points of improvements that should be made?

davidzwa:
@JukkaL we're eager to hear your thoughts.

PR #8590, opened 2020-03-25
Various fixes to following imports in mypy daemon

JukkaL:
Included fixes:

- filter out bogus suppressed modules to speed things up
- fix issue with import inside function
- fix stub packages
- also support `/` path separators on Windows
- update search path based on paths given on the command line
- support changes to files passed on the command line
- add logging
- minimal namespace package support

Namespace package support is not complete, but the issue I found isn't
specific to following imports in mypy daemon, so I didn't try to fix it
in this PR. I'll create a follow-up issue.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -671,6 +671,10 @@ class Server:
         for module, state in graph.items():
             all_suppressed |= state.suppressed_set
 
+        # Filter out things that shouldn't actually be considered suppressed.
+        # TODO: Figure out why these are treated as suppressed
+        all_suppressed = {module for module in all_suppressed if module not in graph}
+
         # TODO: Namespace packages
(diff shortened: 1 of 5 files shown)


JukkaL:
The travis build seems to have passed, but it's not reflected here.

mypy/dmypy_server.py
```diff
@@ -720,6 +741,9 @@ def pretty_messages(self, messages: List[str], n_sources: int,
 
     def update_sources(self, sources: List[BuildSource]) -> None:
         paths = [source.path for source in sources if source.path is not None]
+        if self.following_imports():
```
msullivan: I don't follow this really?

JukkaL: If there are directories in paths, we'll get exceptions inside `fswatcher`.

PR #8593, opened 2020-03-25
[mypyc] Move a bunch of things to being call_emit

msullivan:


mypyc/lib-rt/CPy.h
```diff
@@ -1277,6 +1312,10 @@ static void CPy_Reraise(void) {
     PyErr_Restore(p_type, p_value, p_traceback);
 }
 
+static int CPy_NoErrOccured(void) {
```
JukkaL: I was a bit worried about the extra call required, but turns out that this is rarely called so the performance impact will be negligible.

PR #8606, opened 2020-03-30
Fix bug by removing fix_suppressed_dependencies

msullivan:
We found a bug internally caused by deleting an indirect dependency
with --follow-imports=skip and --warn-unused-ignores. This is caused
by the indirect dependency being moved from deps to suppressed by
fix_suppressed_dependencies, which loses the knowledge that it is an
indirect dependency, causing verify_dependencies to take it much more
seriously than it should be.

fix_suppressed_dependencies was added as a hotfix for an incremental
bug discovered internally in 2016 that we never found a test case
for. I hypothesized during an earlier refactor that it had been made
redundant. Since fix_suppressed_dependencies is causing trouble now,
let's put that hypothesis to the test and remove it. If this causes a
regression, then maybe we'll be able to finally extract the test case
we want.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2247,11 +2247,9 @@ class State:
             all_deps = self.suppressed
         else:
             # Strip out indirect dependencies. See comment in build.load_graph().
+            dependencies = [dep for dep in self.dependencies
+                            if self.priorities.get(dep) != PRI_INDIRECT]
             all_deps = dependencies + self.suppressed + self.ancestors
-            all_deps = [dep for dep in all_deps
-                        if self.priorities.get(dep) != PRI_INDIRECT]
-            #all_deps = dependencies + self.suppressed + self.ancestors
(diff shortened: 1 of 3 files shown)


mypy/build.py
```diff
@@ -2286,9 +2247,11 @@ def verify_dependencies(self, suppressed_only: bool = False) -> None:
             all_deps = self.suppressed
         else:
             # Strip out indirect dependencies. See comment in build.load_graph().
-            dependencies = [dep for dep in self.dependencies
-                            if self.priorities.get(dep) != PRI_INDIRECT]
             all_deps = dependencies + self.suppressed + self.ancestors
+            all_deps = [dep for dep in all_deps
+                        if self.priorities.get(dep) != PRI_INDIRECT]
+            #all_deps = dependencies + self.suppressed + self.ancestors
```
Michael0x2a: Can we remove this commented line? Or alternatively include a comment explaining why we're keeping it/linking to this PR/describing the experiment-in-process.

msullivan: Oh, yeah, oops

PR #8642, opened 2020-04-06
Remove __eq__ and __ne__ overrides for dataclasses and attrs

### LINKED ISSUE #8638: Invalid definition of __eq__ on dataclasses
[This code](https://github.com/python/mypy/blob/master/mypy/plugins/dataclasses.py#L127) does not work. It defines `__eq__` method like this:
```python
def __eq__(self: T, other: T) -> bool: pass
```
But on this example
```python

@dataclass
class A: pass

reveal_type(A.__eq__)
```
latest version of mypy outputs
```
test.py:6: note: Revealed type is 'def (builtins.object, builtins.object) -> builtins.bool'
```

And it is right, I think dataclasses shouldn't override `__eq__` signature of base `object` type and we should remove this broken code completely.

tkukushkin:
Fixes #8638.


JukkaL:
The Travis failure is a flake.

PR #8643, opened 2020-04-06
Resolves #8627

### LINKED ISSUE #8627: Update fastparse for 3.9 AST changes
In 3.9.0a5, some changes landed to the AST for subscripting (https://github.com/python/cpython/pull/9605). We need to support them.

This first got reported in #8614 as "syntax error in type comment", which is pretty bad, also! We should clean up that error message when we fix this.

CyberSaxosTiGER:


CyberSaxosTiGER:
Sorry, I don't understand what you mean with "expressions in Python 3"

davidzwa:
Just a thought, I think he means Python 3.5, 3.6 etc (not only 3.9 and up)

hauntsaninja:
Hello!

I wrote up my own version of this fix, before I found that this PR existed. I believe Sully's referring to making sure anywhere else we visit_Subscript we do the right thing, which I think happens to work out.

You'll still want CI to pass though, feel free to look at my fix. Also happy to open a PR directly: https://github.com/hauntsaninja/mypy/commit/3aec92b9222b09aafa1385377c00860528223f9f

Separately, some stuff still needs to happen to get CI green for Python 3.9, e.g. see https://bugs.python.org/issue40349. We should probably start tracking stuff come beta.

hauntsaninja:
This should be closed, since we merged https://github.com/python/mypy/pull/8716

JelleZijlstra:
Thanks for your contribution @CyberSaxosTiGER.

PR #8660, opened 2020-04-11
[mypyc] Use struct access instead of native getters/setters

ilevkivskyi:
After some discussions with Sully we agreed that essentially inlining all attribute access may be a good idea. For trait attribute access, this requires setting up offset tables.

I didn't add many tests, relying on existing ones, since this PR doesn't add any new functionality. Anyway, please review carefully (ignore whitespace option may be useful).

### DIFF
--- a/mypyc/codegen/emitclass.py
+++ b/mypyc/codegen/emitclass.py
@@ -345,7 +345,8 @@ def generate_vtables(base: ClassIR,
             max(1, len(vtable))))
         # Trait attributes entry.
         emitter.emit_line('static size_t {}[{}];'.format(
-            trait_offset_table_name, max(1, len(trait.attributes)))
+            trait_offset_table_name(trait),
+            max(1, len(trait.attributes)))
         )
 
     # Emit vtable setup function
@@ -369,7 +370,7 @@ def generate_vtables(base: ClassIR,
     emitter.emit_line('return 1;')
     emitter.emit_line('}')
 
-    return vtable_name if not subtables else "{} + {}".format(vtable_name, len(subtables) * 2)
+    return vtable_name if not subtables else "{} + {}".format(vtable_name, len(subtables) * 3)
 
 
 def generate_offset_table(trait_offset_table_name: str,
--- a/mypyc/codegen/emitclass.py
+++ b/mypyc/codegen/emitclass.py
@@ -117,7 +117,6 @@ def generate_class(cl: ClassIR, module: str, emitter: Emitter) -> None:
     members_name = '{}_members'.format(name_prefix)
     getseters_name = '{}_getseters'.format(name_prefix)
     vtable_name = '{}_vtable'.format(name_prefix)
-    offset_table_name = '{}_offset_table'.format(name_prefix)
     traverse_name = '{}_traverse'.format(name_prefix)
     clear_name = '{}_clear'.format(name_prefix)
     dealloc_name = '{}_dealloc'.format(name_prefix)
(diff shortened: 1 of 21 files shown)


ilevkivskyi:
It looks like Windows build failed because of empty initializers. Btw @msullivan if you have suggestions for tricky corner cases worth testing (interactions with interpreted classes, separate compilation, etc) please let me know.

ilevkivskyi:
OK, all green now.

ilevkivskyi:
In other news: this PR gives a 15%-17% perf boost (unless I messed up something).

ilevkivskyi:
Unfortunately, the above result, only holds for gcc, for clang the perf win is only 2%.

mypyc/codegen/emitclass.py
```diff
@@ -260,6 +267,82 @@ def emit_line() -> None:
...
+
+    These will look similar to trait vtables, but contain memory
+    offsets for trait attributes in class structs:
+        {
+            CPyType_T1,
+            {offsetof(native__C, _x1), offsetof(native__C, _y1), ...},
+            CPyType_T2,
+            {offsetof(native__C, _x2), offsetof(native__C, _y2), ...},
+            ...
+        }
```
msullivan, line 282: Could we included the offset tables as part of the trait vtables?

mypyc/codegen/emitfunc.py
```diff
@@ -199,13 +191,50 @@ def visit_load_error_value(self, op: LoadErrorValue) -> None:
...
+
+        This either has a form like obj->attr_name for attributes defined in non-trait
+        classes, and *(obj + attr_offset) for attributes defined by traits. We also
+        insert all necessary C casts here.
+        """
+        cast = '({} *)'.format(op.class_type.struct_name(self.emitter.names))
+        if decl_cl.is_trait and op.class_type.class_ir.is_trait:
+            # For pure trait access find the offset first, offsets
+            # are ordered by attribute position in the cl.attributes dict.
+            trait_attr_index = list(decl_cl.attributes).index(op.attr)
```
msullivan, line 205: Put a `TODO` here or something so we can find this easy if we ever want to improve the perf

ilevkivskyi:
@msullivan 
Thanks for the suggestion, merging offset tables in the trait vtables indeed removed the memory overhead. I tried to play with this a bit and I decided to ditch `VTableAttr`, it is probably subjective, but it seems simpler to me to have the 3-layer table layout than mix attributes and methods in trait vtables. Also if we will need to make this backend-agnostic, then it should be easy to adjust the logic depending on particular architecture we will choose.

mypyc/codegen/emitfunc.py
```diff
@@ -216,26 +246,31 @@ def visit_get_attr(self, op: GetAttr) -> None:
...
+                    attr_rtype, attr_expr, '==', unlikely=True
+                )
+                exc_class = 'PyExc_AttributeError'
+                self.emitter.emit_lines(
+                    'PyErr_SetString({}, "attribute {} of {} undefined");'.format(
+                        exc_class, repr(op.attr), repr(cl.name)),
+                    '} else {')
+                self.emitter.emit_inc_ref(attr_expr, attr_rtype)
+                self.emitter.emit_line('}')
+            self.emitter.emit_line('{} = {};'.format(dest, attr_expr))
```
msullivan, line 262: Let's do the assignment to the destination before we do the checks. I think that will actually save us a load, since the compiler won't be able to assume that the value stays the same across the incref call.

PR #8661, opened 2020-04-12
semanal: populate module_public even for missing modules

### LINKED ISSUE #8649: stubtest reports imported name as not present at runtime even though it's not reexported

A bug, unless this is somehow intended behavior.

https://github.com/henribru/stubtest-bug

```
error: stubtest_bug.foo.Bar is not present at runtime
Stub: at line -1
Any
Runtime:
MISSING

error: stubtest_bug.foo.Baz is not present at runtime
Stub: at line -1
Any
Runtime:
MISSING
```

Given that Baz is reexported while Bar isn't, I would only expect an error for Baz. Importing a name that's not present at runtime in the stub to use in type annotations is a common use case and shouldn't produce errors, since Mypy won't let you import that name anyway without it being reexported.

Note that this seems to work as expected if there are annotations for the module you import from. The error only occurs if you import from a module without annotations and stick a type ignore comment at the import (the latter is needed to make stubtest run at all).

Python 3.7.3 and Mypy 0.770. I haven't tested master, but seems like there haven't been any changes to stubtest since 0.770 anyway.

hauntsaninja:
Fixes #8649

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1731,8 +1731,15 @@ class SemanticAnalyzer(NodeVisitor[None],
                 # Target module exists but the imported name is missing or hidden.
                 self.report_missing_module_attribute(module_id, id, imported_id, imp)
             else:
+                module_public = (
+                    not self.is_stub_file
+                    and self.options.implicit_reexport
+                    or as_id is not None
+                )
                 # Import of a missing (sub)module.
-                self.add_unknown_imported_symbol(imported_id, imp, target_name=fullname)
+                self.add_unknown_imported_symbol(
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
Could you add a test?

The Travis failure doesn't look real (it's the docs build). I hit the "restart job" button but not sure if that will help.

hauntsaninja:
I was a little unsure about where to add a test, since module_public isn't widely used and directly checking symbol tables isn't something current unit tests do. But since the issue was originally filed on stubtest, I figure this is okay.

hauntsaninja:
Gentle bump!

Issue #8219, opened 2019-12-30
mypy reports error when Enum 2nd arg is variable instead of literal


cppljevans:
Please provide more information to help us understand the issue:

* Are you reporting a bug, or opening a feature request?

Reporting a bug.

* Please insert below the code you are checking with mypy,

from enum import Enum
Cs = 'C0 C1 C2'
Cn = Enum('Cn', Cs)

* What is the actual behavior/output?

simplest.py:3: error: Enum() expects a string, tuple, list or dict literal as the second argument

* What is the behavior/output you expect?

No errors

* What are the versions of mypy and Python you are using?

python3 --version
Python 3.6.9
mypy --version
mypy 0.560

* Do you see the same issue after installing mypy from Git
  master?
  
Did not try with Git master.  
  
* What are the mypy flags you are using? (For example --strict-optional)

No flags.

* If mypy crashed with a traceback

No crash.

DevilXD:
Pretty sure this is due to enums already being special-treated by MyPy, as they're classes that are generated dynamically, so static analysis for those is pretty much already impossible. The reason a literal (not variable) is expected there, is because with a literal, the "special treatment" of enums can somewhat check what members you have defined, and type check based on that. When you put in a variable there (even if it's a non-changeable literal at runtime), MyPy can't tell if the string you're passing in there has changed dynamically before or not (since all it knows is that "it's a string"), so type-checking it accurately is impossible.

I'd say this is working as intended.

cppljevans:
On 3/26/20 9:34 AM, DevilXD wrote:
> Pretty sure this is due to enums already being special-treated by MyPy,
> as they're classes that are generated dynamically, so static analysis
> for those is pretty much already impossible. The reason a literal (not
> variable) is expected there, is because with a literal, the "special
> treatment" of enums can somewhat check what members you have defined,
> and type check based on that. When you put in a variable there (even if
> it's a non-changeable literal at runtime), MyPy can't tell if the string
> you're passing in there has changed dynamically before or not (since all
> it knows is that "it's a string"), so type-checking it accurately is
> impossible.

IIUC, c++ can tell if a variable is a constexpr:

  https://en.cppreference.com/w/cpp/language/constexpr

and propagate that conexpr value to all the places that variable
is used. Now, I realize python is an interpreted language,
but surely something like mypy can analyze the code to
determine that a value of a variable has not changed since
it was initially assigned.  There's even a way to specify
that a python variable can never change it's value:


[...]

PR #8664, opened 2020-04-12
Enum now accepts String literals and final ...

vincent-prz:
This aims to solve https://github.com/python/mypy/issues/8219

This PR is still a draft, but am I going in the right direction ? I am doing lot of if refinements from line 150, in order to be compliant with mypy linting, maybe there is a less clunky way to achieve the same result ? Also, should the following cases typecheck ?

```python
class A:
    x: Literal['x y'] = 'x y'

Coordinates = Enum('Coordinates', A.x)
```

```python
def f() -> Literal['x y']
    return 'x y'

Coordinates = Enum('Coordinates', f())
```

Moreover, I had to add `[builtins fixtures/tuple.pyi]` in my tests, they would otherwise crash, but I am not sure to understand why.


mypy/semanal_enum.py
```diff
@@ -146,6 +147,25 @@ def parse_enum_call_args(self, call: CallExpr,
...
+                fields = proper_type.value
+                for field in fields.replace(',', ' ').split():
+                    items.append(field)
+            elif args[1].node.is_final and isinstance(args[1].node.final_value, str):
+                fields = args[1].node.final_value
+                for field in fields.replace(',', ' ').split():
+                    items.append(field)
+            else:
+                return self.fail_enum_call_arg(
+                    "%s() expects a string, tuple, list or dict literal as the second argument" %
```
97littleleaf11, line 166: Could you please merge the same error messages into messages_registry.py

97littleleaf11 (commented):
@sobolevn Do you have any suggestions on this PR?

test-data/unit/check-newsemanal.test
```diff
@@ -2058,6 +2058,24 @@ from enum import Enum
...
+Animal = Enum('Animal', x)
+
+[builtins fixtures/tuple.pyi]
+
+[case testNewAnalyzerEnumAcceptsFinalValue]
+from enum import Enum
+from typing_extensions import Final
+
+x: Final['str'] = 'ANT BEE CAT DOG'
+Animal = Enum('Animal', x)
```
sobolevn, line 2075: I would also check that `Animal.ANT`, `Animal.BEE`, etc are available in both test cases.

sobolevn (commented):
Sorry, I've missed it! Thanks for the ping 🙂

97littleleaf11:
cc @sobolevn

Issue #8663, opened 2020-04-12
Unpacking a NamedTuple subclass uses superclass' types


brettcs:
If you subclass an existing NamedTuple to provide more specific type declarations for some of the fields, and then unpack an instance of that NamedTuple, mypy 0.770 gets its type information from the superclass, not the subclass. Minimal reproduction:

```python
from typing import NamedTuple, Optional

class BaseInventory(NamedTuple):
    number: Optional[int]
    name: str


class Inventory(BaseInventory):
    number: int


item = Inventory(42, 'parrot')
item.number % 10  # mypy allows this
number, _ = item
number % 10  # mypy does not allow this
```

The last line reports:

```
error: Unsupported operand types for % ("None" and "int")  [operator]
note: Left operand is of type "Optional[int]"
```

I expected no error, since accessing the field by name works fine.

PR #8665, opened 2020-04-12
[mypyc] Translate attribute access to index access for named tuples

ilevkivskyi:
Fixes https://github.com/mypyc/mypyc/issues/638

Previously `PyObject_GetAttr` was used for named tuples, which is slower.

### DIFF
--- a/mypyc/test-data/irbuild-tuple.test
+++ b/mypyc/test-data/irbuild-tuple.test
@@ -162,3 +162,35 @@ L4:
     r8 = None
     return r8
 
+[case testNamedTupleAttribute]
+from typing import NamedTuple
+
+NT = NamedTuple('NT', [('x', int), ('y', int)])
+
+def f(nt: NT, b: bool) -> int:
+    if b:
+        return nt.x
+    return nt.y
+[out]
+def f(nt, b):
+    nt :: tuple
+    b :: bool
+    r0 :: short_int
+    r1 :: object
+    r2 :: int
+    r3 :: short_int
+    r4 :: object
+    r5 :: int
+L0:
+    if b goto L1 else goto L2 :: bool
+L1:
+    r0 = 0
+    r1 = nt[r0] :: tuple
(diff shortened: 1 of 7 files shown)


mypyc/test-data/irbuild-tuple.test
```diff
@@ -162,3 +162,35 @@ L4:
     r8 = None
     return r8
 
+[case testNamedTupleAttribute]
+from typing import NamedTuple
+
+NT = NamedTuple('NT', [('x', int), ('y', int)])
```
JelleZijlstra, line 168: Might be worth adding a unit test for a NamedTuple that inherits from another, something like the example from #8663.

ilevkivskyi:
@msullivan 
> I agree with Jelle's suggestion

I was trying to add a subclass in compiled code in different ways, but I was always getting `Inheriting from most builtin types is unimplemented`. I therefore only added a test with a subclass in interpreted code.

While doing this, I actually noticed that using index access is unsafe. Imagine someone overrides the accessor in an interpreted subclass:
```
class NT(NamedTuple):
    x: int
    y: int

def get_y(nt: NT) -> int:
    return nt.y

[file driver.py]
from native import NT, get_y

class C(NT):
    @property
    def y(self):
        return 2 * self[1]
assert get_y(C(1, 1)) == 2  # Works in Python but fails if we compile code
```
Should we worry about such overrides? Or people who write such overrides deserve this?

ilevkivskyi:
Also the perf win from this is pretty minor, around 1% for mypy self-check.

msullivan:
We have the same problem for attribute access and property overrides for all inherited subclasses of compiled classes, as it turns out, and we don't worry about it too much there...

ilevkivskyi:
> We have the same problem for attribute access and property overrides for all inherited subclasses of compiled classes, as it turns out, and we don't worry about it too much there...

OK, I will merge this one then.

PR #8672, opened 2020-04-14
[mypyc] Fix for loop over range mutating start variable

JukkaL:
Don't modify the range start variable in a loop like this:

```
def f() -> None:
    n = 0
    for x in range(n, 5):
        pass
    print(n)  # Should be 0, not 5
```


JukkaL:
I encountered this when trying to compile a benchmark.

Issue #8678, opened 2020-04-16
Add Code_of_Conduct.md


csharp489:
No Code_of_Conduct.md file is present.
As a open sourec Contributor, I would like to add a Code_of_conduct.md file in this project.

ethanhs:
Hi, as a project under the Python organization I believe we abide by the PSF code of conduct, I'd be happy to review a PR that adds a CODE_OF_CONDUCT.md file pointing to the PSF code of conduct.

EkanshMangal:
Hello Sir, If issue is still open
Can i fix this issue?

hauntsaninja:
Closing as per discussion in https://github.com/python/mypy/pull/8690

PR #8690, opened 2020-04-17
Add Code_of_Conduct.md file and link it to README.md file

csharp489:
PR #8678
 @ethanhs I have added Code_of__conduct file and also link it to readme . Please check it.
Thanks.

### DIFF
--- a/Code_of_Conduct.md
+++ b/Code_of_Conduct.md
@@ -1,0 +1,39 @@
+## Code of Conduct
+>The code of conduct helps to welcome all people to contribute and pledges in return to value them as whole human beings and to foster an atmosphere of kindness, cooperation, and understanding.
+### Our Pledge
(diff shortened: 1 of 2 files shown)


TH3CHARLie:
I may miss some context here, I think since mypy is under the PSF code of conduct should reflect this, the code of conduct file from cpython would be a good and sufficient example: https://github.com/python/cpython/blob/master/CODE_OF_CONDUCT.md. 

Also, the conduct presented in your PR seems pretty unrelated to PSF code of conduct but is similar to https://www.contributor-covenant.org/version/1/4/code-of-conduct/ instead.

JukkaL:
We already have a code of conduct specified in https://github.com/python/mypy/blob/master/CONTRIBUTING.md. Maybe it could be somewhere where it's easier to discover.

hauntsaninja:
Closing as per above

PR #8702, opened 2020-04-21
mypy tests: sidestep AST col_offset changes

### LINKED ISSUE #8701: mypy tests are broken on Python 3.8.2
https://github.com/python/cpython/pull/18491 changes col_offsets for some parenthesised expressions, of which mypy has a few in its test suite. This affects Python 3.9 as well. Note Travis currently seems to run with 3.8.0, I believe they update their Python images whenever they see fit.

https://bugs.python.org/issue40349
https://bugs.python.org/issue39474

hauntsaninja:
Fixes #8701


msullivan:
Sigh, yeah.

PR #8716, opened 2020-04-23
Fix parser when using Python 3.9

### LINKED ISSUE #8627: Update fastparse for 3.9 AST changes

gvanrossum:
The structure of the AST for subscripts was changed in
https://github.com/python/cpython/pull/9605

Fixes #8627

Also fix the majority of the errors produced when running pytest with Python 3.9, but not all.

### DIFF
--- a/mypy/test/helpers.py
+++ b/mypy/test/helpers.py
@@ -233,6 +233,8 @@ def clean_up(a: List[str]) -> List[str]:
     remove trailing carriage returns.
     """
     res = []
+    pwd = os.getcwd()
+    driver = pwd + '/driver.py'
     for s in a:
         prefix = os.sep
         ss = s
@@ -241,6 +243,8 @@ def clean_up(a: List[str]) -> List[str]:
                 ss = ss.replace(p, '')
         # Ignore spaces at end of line.
         ss = re.sub(' +$', '', ss)
+        # Remove pwd from driver.py's path
+        ss = ss.replace(driver, 'driver.py')
         res.append(re.sub('\\r$', '', ss))
     return res
 
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -49,6 +49,8 @@ try:
         import ast as ast3
         assert 'kind' in ast3.Constant._fields, \
                "This 3.8.0 alpha (%s) is too old; 3.8.0a3 required" % sys.version.split()[0]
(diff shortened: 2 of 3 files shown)


gvanrossum:
Hm, there are some errors in the 3.9 test run. Let me look into these.

gvanrossum:
Okay, there are some remaining Python 3.9 test failures, but they seem to be just cases where Python 3.9 somehow produces different output.  Examples:
```
Expected:
  main:3: error: keyword argument repeated      (diff)
Actual:
  main:3: error: "f" gets multiple values for keyword argument "a" (diff)
```
(The new parser produces a better error message for duplicate keyword args.)

Also a few parse trees are different, e.g.:
```
  E:       TupleExpr:-1(
  A:       TupleExpr:1(
                     ^
```
(The old parser didn't have a line or column number for certain Tuples.)

Finally:
```
Expected:
  Daemon started                                (diff)
  mypy-daemon: error: Missing target module, package, files, or command. (diff)
  == Return code: 2
Actual:
  mypy-daemon: error: Missing target module, package, files, or command. (diff)
  Daemon started                                (diff)
  == Return code: 2
```
Seems the two error lines are produced in a different order.

mypy/test/helpers.py
```diff
@@ -233,6 +233,8 @@ def clean_up(a: List[str]) -> List[str]:
     remove trailing carriage returns.
     """
     res = []
+    pwd = os.getcwd()
```
msullivan, line 236: What was going on here?

gvanrossum: Many tests suddenly showed the full pathname in the traceback instead of simply `File "driver.py", line NNN`. I don't actually know why, but I could imagine that somehow an absolute path was computed by Python where it previously believed the pathname on the command line? It's always about "driver.py" which I presume is the synthetic script created by the test harness.

gvanrossum: Playing around, this just seems new behavior in Python 3.9, so I think my fix is the best we can do.

gvanrossum:
Merging now, we can iterate later.

hauntsaninja:
Opened https://github.com/python/mypy/pull/8743 for fixing parse trees.
The full path name in the traceback is probably caused by https://bugs.python.org/issue20443.

PR #8717, opened 2020-04-24
Sync typeshed

mxr:


msullivan:
Sure, why not; what are you trying to get synced?

mxr:
@msullivan nothing urgent but it seemed like the next logical step after committing to typeshed 😄

PR #8725, opened 2020-04-25
[mypyc] Faster dict iteration

ilevkivskyi:
Fixes https://github.com/mypyc/mypyc/issues/167

The implementation is pretty straightforward and follows the idea proposed in the issue. The perf impact is actually pretty small, around 1%

### DIFF
--- a/mypyc/irbuild/for_helpers.py
+++ b/mypyc/irbuild/for_helpers.py
@@ -7,20 +7,23 @@ such special case.
 
 from typing import Union, List, Optional, Tuple, Callable, Type
 
-from mypy.nodes import Lvalue, Expression, TupleExpr, CallExpr, RefExpr, GeneratorExpr, ARG_POS, MemberExpr
+from mypy.nodes import (
+    Lvalue, Expression, TupleExpr, CallExpr, RefExpr, GeneratorExpr, ARG_POS, MemberExpr
+)
 from mypyc.ir.ops import (
-    Value, BasicBlock, LoadInt, Branch, Register, AssignmentTarget,
-    TupleGet, AssignmentTargetTuple, TupleSet)
+    Value, BasicBlock, LoadInt, Branch, Register, AssignmentTarget, TupleGet,
+    AssignmentTargetTuple, TupleSet
+)
 from mypyc.ir.rtypes import (
     RType, is_short_int_rprimitive, is_list_rprimitive, is_sequence_rprimitive,
-    RTuple, is_dict_rprimitive)
-from mypyc.primitives.dict_ops import dict_next_rtuple, dict_next_pair_op, dict_check_size_op
+    RTuple, is_dict_rprimitive
+)
+from mypyc.primitives.dict_ops import dict_next_pair_op, dict_check_size_op
 from mypyc.primitives.int_ops import unsafe_short_add
 from mypyc.primitives.list_ops import new_list_op, list_append_op, list_get_item_unsafe_op
 from mypyc.primitives.generic_ops import iter_op, next_op
 from mypyc.primitives.exc_ops import no_err_occurred_op
 from mypyc.irbuild.builder import IRBuilder
-from mypyc.primitives.tuple_ops import new_tuple_op
 
 GenFunc = Callable[[], None]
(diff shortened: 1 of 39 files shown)


ilevkivskyi:
Oh, tons of tests fail on Python 3.5 because this apparently changes dict iteration order there. It is probably a good opportunity to fix them (i.e. make them stable).

ilevkivskyi:
@msullivan I think this should be now ready for review. Some notes:
* Python 3.5 turned out to be a pain (because of dict ordering), so I just skip this on Python 3.5.
* I squeeze some ms by having dedicated ops for `keys`/`values`/`items`, if you think it is not worth it, I can return to just one.
* Would it make sense to also add ops for `keys`/`values`/`items` when the appear in a non-loop context (there are pretty many of these in mypy)? If yes I can either cp them here or make a separate PR.

JukkaL:
FWIW, I believe that the most common uses of `keys`/`values`/`items` outside of a loop context in general Python code is as arguments to `list` (and maybe `set` and/or `dict`), such as `list(foo.values())`. If special casing helps these use cases (at least for small dictionaries), I think that it's worth supporting non-loop contexts. Separate PR would be fine.

mypyc/test-data/run.test
```diff
@@ -1065,6 +1065,36 @@ assert d == dict(object.__dict__)
...
+        print(k)
+        print(v)
+    for v in d3.values():
+        print(v)
+
+class Custom(Dict[int, int]): pass
+[file driver.py]
+from native import print_dict_methods, Custom
+print_dict_methods({}, {}, {})
+print_dict_methods({1: 2}, {3: 4}, {5: 6})
```
JukkaL, line 1085: Also test dictionaries with multiple items.

Maybe test what happens if we change the length of the dictionary during iteration, such as by clearing it?

mypyc/test/test_run.py
```diff
@@ -219,12 +219,6 @@ def run_case_step(self, testcase: DataDrivenTestCase, incremental_step: int) ->
                 errors=errors,
                 groups=groups,
             )
-            if VERBOSE_DEBUG_DUMP:
-                for group in cfiles:
-                    for name, content in group:
-                        print(name)
-                        print('-' * len(name))
-                        print(content + '\n')
```
JukkaL, line 227: This seems unrelated. Maybe move it to a separate PR (not a big deal though)?

mypyc/irbuild/for_helpers.py
```diff
@@ -170,6 +177,16 @@ def make_for_loop_generator(builder: IRBuilder,
         for_list.init(expr_reg, target_type, reverse=False)
         return for_list
 
+    # Unfortunately fast dict iteration breaks too many tests on Python 3.5.
```
JukkaL, line 180: This is not a big problem, since Python 3.5 will reach EoL relatively soon.

ilevkivskyi: There is actually a bigger issue Sully found: even on newer version this is not safe for some dict subclasses. For example if someone passes an `OrderedDict`, after `.move_to_end()` was called on it, it may not always work properly w.r.t. items order.

I will need to investigate this more on weekend (this applies to other dict PR too).

mypyc/test-data/run.test
ilevkivskyi: > Maybe test what happens if we change the length of the dictionary during iteration, such as by clearing it?

There is an existing (huge) test that changes dict size and verifies that `RuntimeError` is raised. I can add a small dedicated test case, since this is probably important.

ilevkivskyi:
@msullivan So I disable the fast path for subclasses now (and enable on Python 3.5). It looks pretty complicated, but not too bad probably. Please let me know if it looks OK.

JukkaL:
This looks good to me. Let's merge this tomorrow unless @msullivan has some additional feedback.

ilevkivskyi:
> This looks good to me. Let's merge this tomorrow unless @msullivan has some additional feedback.

I am assuming Sully doesn't have any comments here, so will merge now.

Issue #3796, opened 2017-08-03
Heavily nested structures are very slow to proceed


yedpodtrzitko:
Hi,
the dictionary below takes more than 30s to be processed. If the codebase contains more of such structures, it takes painfully long time to check it. 

For now we use #3789 as a workaround to skip such slow parts.

Cheers,
yed_

testing env:
MB Pro 2013
Python 3.6
Mypy 0.521
 
testing file:
```
input = {
    'body': {
        'token': {
            'type': 'string',
            'required': False
        },
        'payloads': {
            'type': 'list',
            'required': True,
            'schema': {
                'type': 'dict',
                'schema': {
                    'message': {
                        'type': 'string',
                        'required': True
                    },
                    'notification': {
                        'type': 'dict',
                        'required': True,
                        'schema': {
                            'body': {
                                'type': 'dict',
                                'schema': {
                                    'type': {
                                        'type': 'string',
                                        'required': True
                                    },
                                    'data': {
                                        'type': 'dict',
                                        'required': True,
                                        'schema': {
                                            'bid': {
                                                'type': 'string',
                                                'required': False
                                            },
                                            'url': {
                                                'type': 'string',
                                                'required': False
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

gvanrossum:
Wow. Confirmed -- on my Mac it takes 21s to process that file.

kirbyfan64:
I wonder if this is somehow similar to:

https://spin.atomicobject.com/2016/04/26/swift-long-compile-time/

kirbyfan64:
(I mean algorithm-wise.)

ilevkivskyi:
It looks like we have two problems here. This code typechecks quickly:
```python
def f(x: int) -> int:
    return x

y = f(f(f(f(f(f(f(f(f(f(f(f(f(f(f(f(f(f(1))))))))))))))))))
```
while this takes around 41 seconds on my desktop:
```python
@overload
def f(x: str) -> str: ...
@overload
def f(x: int) -> int: ...
def f(x):
    return x

y = f(f(f(f(f(f(f(f(f(f(f(f(f(f(f(f(f(f(1))))))))))))))))))
```
Independently, this takes even more, around 1 minute 12 seconds:
```python
T = TypeVar('T')
def f(x: T) -> T:
    return x

[...]

PR #8729, opened 2020-04-26
dont process ignored modules

JulianSza:
dont touch ignored files

otherwise https://github.com/python/mypy/issues/3796 cannot be avoided by using `ignore_modules` in mypy.ini

work in progress, does not support `*` in module path

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -970,7 +970,8 @@ def process_options(args: List[str],
             targets = [
                 t
                 for t in targets
-                if explicitly_needed_regex.fullmatch(t.module) or not ignored_regex.fullmatch(t.module)]
+                if explicitly_needed_regex.fullmatch(t.module)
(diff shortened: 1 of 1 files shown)


JulianSza:
Hmm... that was meant to hit my personal repo, not the official one... oh well:)

I was struggling with heavily nested dicts in our unit tests. Since we are just starting with mypy I was hoping I can just disable type checks for unit tests. Sadly `ignore_errors` in mypy.ini only prevents the errors from being printed, instead of being run (and therefore taking longer to execute)

Was there any reason for this choice? (this fix is working for us and its passing `./runtests` but not sure if this impacts other use cases somehow

JukkaL:
If mypy is not behaving as expected, please file an issue first. (You can add a link to this PR in the issue.)

hauntsaninja:
Closing as per Jukka's comment.

PR #8742, opened 2020-04-28
[mypyc] Add primitive ops for dictionary methods

ilevkivskyi:
This covers both view and list versions (it looks like both are used relatively often, at least in mypy). This accompanies https://github.com/python/mypy/pull/8725 to account for cases where `keys`/`values`/`items` appear in non-loop contexts.

### DIFF
--- a/mypyc/test-data/run.test
+++ b/mypyc/test-data/run.test
@@ -989,7 +989,7 @@ update(s, [5, 4, 3])
 assert s == {1, 2, 3, 4, 5}
 
 [case testDictStuff]
-from typing import Dict, Any, List, Tuple
+from typing import Dict, Any, List, Set, Tuple
 from defaultdictwrap import make_dict
 
 def f(x: int) -> int:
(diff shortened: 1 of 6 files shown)


ilevkivskyi:
...and as usual Python 3.5 is totally busted. I will probably just disable this on Python 3.5. PR #8725 adds the corresponding `CompilerOptions` flag, I will just reuse it here.

ilevkivskyi:
@msullivan @JukkaL I switched to disabling the optimization for subclasses here too. Now tests pass on Python 3.5.

JukkaL:
I measured that this can make the ops at least 2x faster for small dictionaries. This is a great improvement!

mypyc/test-data/run.test
```diff
@@ -1035,14 +1035,17 @@ def u(x: int) -> int:
     d.update(x=x)
     return d['x']
 
+def get_content(d: Dict[int, int]) -> Tuple[List[int], List[int], List[Tuple[int, int]]]:
+    return list(d.keys()), list(d.values()), list(d.items())
```
JukkaL, line 1039: Add also tests for the "plain" primitive ops that don't create a list, such as by doing `set(d.keys())`?

Might be worth testing dict subclasses as well.

PR #8743, opened 2020-04-29
fastparse: add line and column information for extended slices

hauntsaninja:
Helps with Python 3.9 test failures

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -1205,7 +1205,10 @@ class ASTConverter:
     def visit_Subscript(self, n: ast3.Subscript) -> IndexExpr:
         e = IndexExpr(self.visit(n.value), self.visit(n.slice))
         self.set_line(e, n)
-        if isinstance(e.index, SliceExpr):
+        if (
+            isinstance(e.index, SliceExpr) or
+            (sys.version_info <= (3, 9) and isinstance(e.index, TupleExpr))
+        ):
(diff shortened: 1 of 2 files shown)


mypy/fastparse.py
```diff
@@ -1205,7 +1205,10 @@ def visit_Attribute(self, n: Attribute) -> Union[MemberExpr, SuperExpr]:
     def visit_Subscript(self, n: ast3.Subscript) -> IndexExpr:
         e = IndexExpr(self.visit(n.value), self.visit(n.slice))
         self.set_line(e, n)
-        if isinstance(e.index, SliceExpr):
+        if (
+            isinstance(e.index, SliceExpr) or
+            (sys.version_info <= (3, 9) and isinstance(e.index, TupleExpr))
```
JukkaL, line 1210: Does this mean that the version check is only true for versions earlier than 3.9.0 (since `version_info` has extra components beyond the first three) or up to 3.9.0? Please add a comment about this.

Or could you perhaps use `< (3, 9)` as the condition?

hauntsaninja: Thanks! I did mean that. Changed a couple things and added comments; hopefully the result is more understandable.

JukkaL:
Thanks!

gvanrossum:
Oh, thanks for this fix @hauntsaninja! I had been wondering about those diff lines in the CI logs for 3.9.

PR #8748, opened 2020-04-29
Fixes error that callable with None argument

### LINKED ISSUE #8720: 'if callable(fn)' should infer that 'fn' is not None
Reporting a bug.

Code.
```
fn = None
if callable(fn):
  fn()
```

Actual output.
```
$ mypy  xyz.py 
xyz.py:3: error: "None" not callable
```

Expected behavior.
I would expect `if callable(fn)` to infer a non-None type.

This is a workaround, but shouldn't be necessary:
```
fn = None
if fn and callable(fn):
  fn()
```

mypy version.
```
$ mypy --version
mypy 0.770
$ python --version
Python 3.8.2
```

LiuYuHui:
Fixes #8720

### DIFF
--- a/test-data/unit/check-callable.test
+++ b/test-data/unit/check-callable.test
@@ -449,3 +449,11 @@ if callable():  # E: Too few arguments for "callable"
     pass
 
 [builtins fixtures/callable.pyi]
+
+[case testCallableWithNoneArgs]
+
+fn = None
+if callable(fn):
+  fn()
+
+[builtins fixtures/callable.pyi]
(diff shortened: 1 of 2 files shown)


test-data/unit/check-callable.test
```diff
@@ -449,3 +449,11 @@ if callable():  # E: Too few arguments for "callable"
     pass
 
 [builtins fixtures/callable.pyi]
+
+[case testCallableWithNoneArgs]
+
+fn = None
+if callable(fn):
+  fn()
```
JukkaL, line 457: Can you also add a test case that narrows down from a union of `None` and a callable type, and use `reveal_type(fn)` to display type of the callable in a few different places?

PR #8777, opened 2020-05-04
[daemon] Don't do updates of builtins modules

msullivan:
We hit a bug internally involving this. I'm going to put a bit more
effort into tracking those in case they are more general, but in
general, nothing good can come of it, I think.

### DIFF
--- a/mypy/server/update.py
+++ b/mypy/server/update.py
@@ -342,6 +342,12 @@ class FineGrainedBuildManager:
         self.manager.log_fine_grained('--- update single %r ---' % module)
         self.updated_modules.append(module)
 
+        # builtins and friends could potentially get triggered because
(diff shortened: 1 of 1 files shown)


mypy/server/update.py
```diff
@@ -342,6 +342,12 @@ def update_module(self,
         self.manager.log_fine_grained('--- update single %r ---' % module)
         self.updated_modules.append(module)
 
+        # builtins and friends could potentially get triggered because
+        # of protocol stuff, but nothing good could possibly come from
+        # actually updating them.
+        if module in ('builtins', 'typing', 'mypy_extensions', 'typing_extensions'):
```
ilevkivskyi, line 348: Do we also need to add `abc` here? Also maybe you can re-use `mypy.semanal_main.core_modules` or define a new constant?

Issue #5354, opened 2018-07-13
type alias to union is invalid in runtime context


glyph:
As of MyPy 0.620:
```python
# aliases.py
from typing import Union
x = Union[int, str]
def y() -> None:
    print(x)
```
This produces the error:

```
aliases.py:5: error: The type alias to Union is invalid in runtime context
```

However, type variables are quite useful in "runtime context" in order to implement things like serialization; for example, https://github.com/Tinche/cattrs makes heavy use of introspecting Union objects to determine which types are valid.

Those of us hacking around code like this have known that the current API for enumerating the members of a Union was private & internal, and not for public consumption, so we've been keeping pace with implementation details while waiting for a public API to get this information to emerge.  These errors are super annoying because they imply the only future-looking direction for this type of code will involve manually re-typing the same list of types for every union alias defined anywhere.

Can this change be backed out?  If not, is there any guidance for how to define a Union such that it is introspectable?

ilevkivskyi:
From mypy's point of view it is hard to distinguish whether this is intentional or a mistake. The point is that this change discovered actual errors in real code, so I am not so keen to revert this change even though it was unintentional.

On the other hand, such aliases will likely fail at runtime in most scenarios where it is an error quite soon. So the value of discovering them statically is not so big.

One possible way is to turn the type aliases into variables by giving them explicit types:
```python
X: Type[Any] = Union[A, B, C]
print(X)  # OK
```
but then `X` can't be used in annotations. I would like to see more input here before making a decision here.

glyph:
I can see your point as far as this being a useful thing to check most of the time.  And given that we're relying on private implementation details of typing I know it's my own responsibility to keep up with such changes :).  Mainly what I'm looking for here is a non-horrible way forward that translates to something that will be close to a one-day public API.  For example, I'm very curious what you mean by 

> such aliases will likely fail at runtime in most scenarios where it is an error quite soon

what will make this the case?

Your suggestion to use `Type[Any]` is an interesting one, but it's somewhat stymied by the fact that your example

```
X: Type[Any] = Union[A, B, C]
```

works, but only *exactly* as written; in a more practical context, like

```
x = Union[int, str]
z: Type[Any] = x
```

we are bitten by

```
"Incompatible types in assignment (expression has type "Type alias to Union", variable has type "Type[Any]")
```

glyph:
OK, now that I'm actually testing on the correct version:

```
x = Union[int, str]
z: Type[Any] = x
```

[...]

PR #8779, opened 2020-05-05
Allow type aliases in runtime context

ilevkivskyi:
Fixes https://github.com/python/mypy/issues/5354

This PR allows using type aliases in runtime context and instead gives the an `object` type (which is the best type that we know). The new error messages are less specific, but the old logic caused annoying false positives for various ORMs that use non-annotation syntax, and for runtime type validation tools.

While working on this I found a minor bug in `object_or_any_from_type()` that returned `Any` for `Union[int, Any]` while it should be `object`.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3126,7 +3126,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             if alias_definition:
                 return AnyType(TypeOfAny.special_form)
             # This type is invalid in most runtime contexts, give it an 'object' type.
-            return join.object_or_any_from_type(item)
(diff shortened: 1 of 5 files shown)


JukkaL:
I like this! Now inferred types of runtime types are inconsistent, however:

```py
from typing import Union

A = Union[int, str]

reveal_type(A)  # object
reveal_type(Union[int, str])  # Any
```

Can you also update the type of `Union[int, str]` (and other similar) to be `object`?

ilevkivskyi:
@JukkaL opened https://github.com/python/typeshed/pull/3966 for this.

Tinche:
Thank you!

PR #8780, opened 2020-05-05
Sync typeshed

ilevkivskyi:
This is to get https://github.com/python/typeshed/pull/3966


ilevkivskyi:
NB: mypyc build fails because of https://github.com/python/typeshed/pull/3830

PR #8786, opened 2020-05-07
Drop appveyor

ethanhs:
Fixes #8786

I believe @gvanrossum will need to delete the project from Appveyor.


JukkaL:
Let's see what happens when we merge this.

PR #8789, opened 2020-05-08
[mypyc] fix syntax

hauntsaninja:
Somehow, this only broke on 3.9's tests, but passed everywhere else.

### DIFF
--- a/mypyc/lib-rt/CPy.h
+++ b/mypyc/lib-rt/CPy.h
@@ -1399,7 +1399,7 @@ static PyObject *CPyDict_ItemsView(PyObject *dict) {
 }
 
 static PyObject *CPyDict_Keys(PyObject *dict) {
-    if PyDict_CheckExact(dict) {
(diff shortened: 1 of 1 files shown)


TH3CHARLie:
I am curious why this syntax did not trigger compile errors(I reproduce this with my local mypyc with no errors as well)

hauntsaninja:
Looks like it's because the `PyDict_CheckExact` is a macro.

Its definition was changed in 3.9 from `#define PyDict_CheckExact(op) (Py_TYPE(op) == &PyDict_Type)` to `#define PyDict_CheckExact(op) Py_IS_TYPE(op, &PyDict_Type)`, which you'll notice no longer has brackets around it.

The parentheses are required in the old version to prevent horrible operator precedence text substitution bugs, but the new version ends up substituting a function call so they're no longer needed.

hauntsaninja:
cc @gvanrossum, since you were interested in 3.9 test failures

PR #8793, opened 2020-05-08
dmypy: fix test failure on py39

hauntsaninja:
There's a fun Python 3.9 only test case failure, where "Daemon started" is
output after the error message from parsing options.  I ended up needing to
bisect to figure out why this consistently breaks on Python 3.9. The issue is
caused by https://github.com/python/cpython/pull/17646/ (which fixes a
surprisingly old BPO).


hauntsaninja:
cc @gvanrossum

Issue #7489, opened 2019-09-09
Don't warn about non keyword-only attributes after keyword-only attributes in attrs


oakkitten:
This behavior was recently [changed](https://github.com/python-attrs/attrs/pull/559) in `attrs`. [Relevant code](https://github.com/python/mypy/blob/master/mypy/plugins/attrs.py#L320-L324) in mypy.

ilevkivskyi:
We will not have time to update implementation soon, but PRs are welcome.

markusschmaus:
Here is a PR: https://github.com/python/mypy/pull/8803

PR #8803, opened 2020-05-12
Allow attrs kw_only arguments at any position

markusschmaus:
This PR fixes https://github.com/python/mypy/issues/7489

```python
import inspect
from typing import Optional

import attr


@attr.s()
class Valid:
    kw_only: Optional[str] = attr.ib(default=None, kw_only=True)
    param: int = attr.ib()
    optional: bool = attr.ib(default=False)


valid1 = Valid(1)
print(valid1)
valid2 = Valid(2, kw_only='something')
print(valid2)
invalid2 = Valid(2, 'something')
print(invalid2)
invalid3 = Valid('something')
print(invalid3)


@attr.s()
class Valid2:
    param: int = attr.ib()
    kw_only: Optional[str] = attr.ib(default=None, kw_only=True)


valid3 = Valid2(3)
print(valid3)
valid4 = Valid2(4, kw_only='something')
print(valid4)

try:
    invalid = Valid2(4, 'something')
except TypeError:
    pass
```

produces the expected output:
```
mypy my_example.py
my_example.py:18: error: Argument 2 to "Valid" has incompatible type "str"; expected "bool"
my_example.py:20: error: Argument 1 to "Valid" has incompatible type "str"; expected "int"
my_example.py:36: error: Too many positional arguments for "Valid2"
Found 3 errors in 1 file (checked 1 source file)
```

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -2,17 +2,9 @@
 import subprocess
 from subprocess import Popen
 from os import system
-from sys import argv, exit, platform, executable, version_info
+from sys import argv, exit
 
-
-# Use the Python provided to execute the script, or fall back to a sane default
-if version_info >= (3, 5, 0):
-    python_name = executable
-else:
-    if platform == 'win32':
-        python_name = 'py -3'
-    else:
-        python_name = 'python3'
+python_name = 'python'
 
 # Slow test suites
 CMDLINE = 'PythonCmdline'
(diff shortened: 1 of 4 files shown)


runtests.py
```diff
@@ -2,17 +2,9 @@
...
-
-# Use the Python provided to execute the script, or fall back to a sane default
-if version_info >= (3, 5, 0):
-    python_name = executable
-else:
-    if platform == 'win32':
-        python_name = 'py -3'
-    else:
-        python_name = 'python3'
+python_name = 'python'
```
TH3CHARLie, line 7: I assume modifications of this file is for local testing on your machine. However, this generally should not be committed to the PR.

markusschmaus: Of course. Sorry I let this slip through.

oakkitten:
i tested this and found that it perfectly resolves the issue for me. thanks!

PR #8811, opened 2020-05-13
Setup pre-commit hook with flake8 and black formatting

### LINKED ISSUE #8799: Addition of Pre-Commit hooks
**Implement Pre-Commit Hooks**

We can setup `pre-commit` hooks to rectify the linting issues so as to keep the code base consistent and prevent travis build fails. This would reduce the time of testing on travis and getting back.

jsparmani:
Fix #8799 
`pre-commit` has been added to maintain code standards and linting. This will perform formatting and lint checks on all staged files using `flake8` and `black` formatting on commit. This would decrease the number of travis build fails due to linting issues. 
`pre-commit install` needs to be run once the requirements have been installed.

We need to perform flake8 and black formatting on the whole repository once to prevent build fails for previous files that are present.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,10 @@
+repos:
+    - repo: https://gitlab.com/pycqa/flake8
+      rev: 3.7.9
+      hooks:
+        - id: flake8
+    - repo: https://github.com/psf/black
+      rev: stable
+      hooks:
+        - id: black
+          language_version: python3
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,3 +4,21 @@ requires = [
     "wheel >= 0.30.0",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.black]
+line-length = 200
+target-version = ['py36', 'py37', 'py38']
+skip-string-normalization = true
+include = '\.pyi?$'
+exclude = '''
+/(
+    \.*
+  | \.git
+  | \.mypy_cache
+  | \.venv
+  # The following are specific to Black, you probably don't want those.
+  | blib2to3
+  | tests/data
+  | profiling
+)/
+'''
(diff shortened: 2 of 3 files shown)


msullivan:
Reformatting the code-base with black is a major proposal that should be advocated for separately. The last I've discussed it with anybody, the sentiment of the core time is sightly negative towards doing this, though I don't think anything is hard or fast there.

What is the flow for using the pre-commit hooks here. Do users need to opt into them before they will be run?

ethanhs:
Yeah I am not very keen on black, I just don't like some of the choices it makes, though that's just my personal opinion. 

On the other hand, I do very much like pre-commit. I generally `pre-commit install` (this runs the hooks before you commit) when I see a project has a pre-commit setup. So to answer your question, it is opt in to run the hooks before committing. It has a bunch of autofixers I quite like the `end-of-file-fixer` and `trailing-whitespace` checks as well.

jsparmani:
@msullivan @ethanhs We can then probably setup `pre-commit` with some other formatting e.g. PEP8, whichever is better. Let me know, what can be added to the `pre-commit` for formatting. I will work further on this.

msullivan:
If you want to update this to just run the linter (with the settings we have) and any reasonable built in autofixes like end-of-file-fixer and trailing-whitespace, then I'll probably merge it.

Getting any autoformatter used on the whole code base will require someone to advocate for it totally separately and I don't think it is a good use of the time of a new contributor :P.

ssbarnea:
I would totally support introduction of pre-commit tool for linting, especially as mypy itself can be orchestrated by pre-commit tool.

Regarding black, I would postpone introduction of black due to being more intrusive. Best part is that pre-commit can be slowly introduced, first with line checking as mentioned by @ethanhs and flake8 after.

With this occasion we maybe can also merge the hook definition from https://github.com/pre-commit/mirrors-mypy and avoid the need to use a mirror for enabling mypy use from pre-commit.

PS. Someone please do a `s/hook/tool/` on the title as this is about using pre-commit tool, the hook part being something optional that needs to be manually installed (something I do not personally use as I like calling it manually).

PR #8700, opened 2020-04-21
Minor documentation fixes

cjolowicz:
Some minor fixes for the Sphinx documentation:

- two simple typos
- unintended word repetition/copy-paste error
- invalid markup (`:py:meth:` nested inside bold markup)

PR #8818, opened 2020-05-13
fix rendering in common_issues.rst

JelleZijlstra:
Current rendering looks messed up (screenshot below). I'm not too familiar with rst, but just not linking in this one place might be fine.

image


JelleZijlstra:
Marking as a draft until I can figure out how to render the RST and confirm that this fixes it.

hauntsaninja:
You can render mypy docs with `tox -e docs`.
This was already fixed by #8700, it seems to just not have been deployed yet.

JelleZijlstra:
Oh right, so we don't need this :)

PR #8827, opened 2020-05-15
fix error that tuple with star expr

### LINKED ISSUE #7779: Unpacking starred expressions in tuples
(migrated from [a comment](https://github.com/python/mypy/issues/1178#issuecomment-545327224) in #1178 at @ethanhs's request)

There seems to an issue with unpacking things when they are part of a tuple on the right side of assignment. #1178 is related, but specifically deals with variable-length tuples (and using `len()` to distinguish them). The current issue occurs even for fixed-length and type-annotated tuples.

```python
# file: ex.py
from typing import Tuple
points = (1, 2)  # type: Tuple[int, int]
x, y, z = *points, 0
```

This results in incorrect inference about the tuple size (but Python does not complain, only mypy):

```console
$ mypy ex.py 
ex.py:4: error: Need more than 2 values to unpack (3 expected)
```

A workaround is to use cast(), as @JukkaL [suggested](https://github.com/python/mypy/issues/1178#issuecomment-176185607):

```python
x, y, z = cast(Tuple[int, int, int], (*points, 0))
```

And here's my versions of Python and mypy:

```console
$ mypy --version
mypy 0.740
$ python --version
Python 3.7.3
```

LiuYuHui:
fixes #7779

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2483,36 +2483,47 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             # control in cases like: a, b = [int, str] where rhs would get
             # type List[object]
             rvalues = []  # type: List[Expression]
-            lhs_len = len(lvalues)
-            idx_of_iterable = []
-            item_type_of_iterable = []  # type: List['mypy.types.Type']
-            idx = 0
-            for rval in rvalue.items:
+            iterable_type = None  # type: Optional[Type]
+            last_idx = None  # type: Optional[int]
+            for idx_rval, rval in enumerate(rvalue.items):
                 if isinstance(rval, StarExpr):
                     typs = get_proper_type(self.expr_checker.visit_star_expr(rval).type)
                     if isinstance(typs, TupleType):
                         rvalues.extend([TempNode(typ) for typ in typs.items])
-                        lhs_len -= len(typs.items)
-                        idx += len(typs.items)
                     elif self.type_is_iterable(typs) and isinstance(typs, Instance):
-                        item_type_of_iterable.append(self.iterable_item_type(typs))
-                        idx_of_iterable.append(idx)
+                        if (iterable_type is not None
+                                and iterable_type != self.iterable_item_type(typs)):
+                            self.fail("Contiguous iterable with same type expected", context)
+                        else:
+                            if last_idx is None or last_idx + 1 == idx_rval:
+                                rvalues.append(rval)
+                                last_idx = idx_rval
+                                iterable_type = self.iterable_item_type(typs)
+                            else:
+                                self.fail("Contiguous iterable with same type expected", context)
                     else:
-                        self.fail("StarExpr should not be a '{}'".format(typs), context)
+                        self.fail("Invalid type '{}' for *expr (iterable expected)".format(typs),
+                             context)
                 else:
                     rvalues.append(rval)
-                    lhs_len -= 1
-                    idx += 1
-            num_every_iterable = 0
-            num_last_iterable = 0
-            if len(idx_of_iterable):
-                num_every_iterable = int(lhs_len / len(idx_of_iterable))
-                num_last_iterable = lhs_len - (len(idx_of_iterable) - 1) * int(num_every_iterable)
-            for i, (idx, item_type) in enumerate(zip(idx_of_iterable, item_type_of_iterable)):
-                if i == (len(idx_of_iterable) - 1):
-                    rvalues[idx:idx] = [TempNode(item_type) for _ in range(num_last_iterable)]
-                else:
-                    rvalues[idx:idx] = [TempNode(item_type) for _ in range(num_every_iterable)]
+            iterable_start = None  # type: Optional[int]
+            iterable_end = None  # type: Optional[int]
+            for i, rval in enumerate(rvalues):
+                if isinstance(rval, StarExpr):
+                    typs = get_proper_type(self.expr_checker.visit_star_expr(rval).type)
+                    if self.type_is_iterable(typs) and isinstance(typs, Instance):
+                        if iterable_start is None:
+                            iterable_start = i
+                        iterable_end = i
+            if (iterable_start is not None
+                    and iterable_end is not None
+                    and iterable_type is not None):
+                iterable_num = iterable_end - iterable_start + 1
+                rvalue_needed = len(lvalues) - (len(rvalues) - iterable_num)
+                if rvalue_needed > 0:
+                    rvalues = rvalues[0: iterable_start] + [TempNode(iterable_type)
+                        for i in range(rvalue_needed)] + rvalues[iterable_end + 1:]
+
             if self.check_rvalue_count_in_assignment(lvalues, len(rvalues), context):
                 star_index = next((i for i, lv in enumerate(lvalues) if
                                    isinstance(lv, StarExpr)), len(lvalues))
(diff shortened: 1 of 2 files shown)


mypy/checker.py
```diff
@@ -2483,8 +2483,16 @@ def check_assignment_to_multiple_lvalues(self, lvalues: List[Lvalue], rvalue: Ex
...
-            rvalues = rvalue.items
-
+            rvalues = []  # type: List[Expression]
+            for rval in rvalue.items:
+                if isinstance(rval, StarExpr):
+                    typs = get_proper_type(self.expr_checker.visit_star_expr(rval).type)
+                    if isinstance(typs, TupleType):
+                        rvalues.extend([TempNode(typ) for typ in typs.items])
+                    else:
+                        rvalues.append(TempNode(typs))
```
JukkaL, line 2493: If the type is an iterable, for example, the correct thing to do here would be to expand the right number of iterable item types, depending on the number of lvalues. It's okay if you don't want to implement that in this PR, however. The alternative would be to give an error about not being able to check the number of rvalues. The current behavior propagates an incorrect type, which I'd prefer to avoid.

test-data/unit/check-tuples.test
```diff
@@ -1421,3 +1421,10 @@ t6: Tuple[int, int, int, int, int, int, int, int, int, int, int, int] = (1, 2, 3
                                                                     # E: Incompatible types in assignment (expression has type Tuple[int, int, ... <15 more items>], variable has type Tuple[int, int, ... <10 more items>])
 
 [builtins fixtures/tuple.pyi]
+
+[case testTupleWithStarExpr]
+from typing import Tuple
+points = (1, 2)  # type: Tuple[int, int]
+x, y, z = *points, 0
```
JukkaL, line 1428: Also reveal the types of `x`, `y`, and `z` to check that we inferred the correct types.

test-data/unit/check-tuples.test
```diff
@@ -1421,3 +1421,10 @@ t6: Tuple[int, int, int, int, int, int, int, int, int, int, int, int] = (1, 2, 3
                                                                     # E: Incompatible types in assignment (expression has type Tuple[int, int, ... <15 more items>], variable has type Tuple[int, int, ... <10 more items>])
 
 [builtins fixtures/tuple.pyi]
+
+[case testTupleWithStarExpr]
+from typing import Tuple
+points = (1, 2)  # type: Tuple[int, int]
```
JukkaL, line 1427: Instead of using all `int` types, use different types so that each of `x`, `y` and `z` will get distinct types. This way we can test that the types are propagated in the correct order.

mypy/checker.py
LiuYuHui: thanks for your review, i would like to implement the correct thing, so the basic idea is that just get the right number of iterable item types, instead of constructing a new `TempNode` type, right?

JukkaL: The more correct thing would be figure out the number of rvalues needed here so that the number of rvalues matches the number of lvalues. Then assuming the rvalue type is an iterable, look at the iterable item type and append a suitable number of `TempNode` instances with the iterable item type. Here we can assume that the length of the iterable is correct, as we don't know how many items there will actually be.

There will be some interesting edge cases, I assume. If the type is not iterable and not a tuple, you can probably fall back to `AnyType` (make sure that an error gets reported either here or elsewhere if the rvalue type is bogus).

mypy/checker.py
```diff
@@ -2482,9 +2482,37 @@ def check_assignment_to_multiple_lvalues(self, lvalues: List[Lvalue], rvalue: Ex
...
+                        self.fail("StarExpr should not be a '{}'".format(typs), context)
+                else:
+                    rvalues.append(rval)
+                    lhs_len -= 1
+                    idx += 1
+            num_every_iterable = 0
+            num_last_iterable = 0
+            if len(idx_of_iterable):
+                num_every_iterable = int(lhs_len / len(idx_of_iterable))
+                num_last_iterable = lhs_len - (len(idx_of_iterable) - 1) * int(num_every_iterable)
```
JukkaL, line 2510: This seems kind of arbitrary. We assume that the items from multiple `*args` rvalues are split between the values in a specific way, but we can't really predict this. Instead, I'd suggest this:
* If all the item types of iterable `*args` rvalues are the same and next to each other, treat them as a single iterable.
* If the item types vary or they aren't next to each other generate an error and fall back to some conservative approximation, such as assuming all their types are `Any`.

I think that the current behavior can be unsafe. Consider this:

```py
a = [1]
b = ['x', 'y', 'z']
x1, x2, x3, x4 = *a, *b
x2 + 5  # Runtime error
```

mypy/checker.py
```diff
@@ -2482,9 +2482,37 @@ def check_assignment_to_multiple_lvalues(self, lvalues: List[Lvalue], rvalue: Ex
...
+                    typs = get_proper_type(self.expr_checker.visit_star_expr(rval).type)
+                    if isinstance(typs, TupleType):
+                        rvalues.extend([TempNode(typ) for typ in typs.items])
+                        lhs_len -= len(typs.items)
+                        idx += len(typs.items)
+                    elif self.type_is_iterable(typs) and isinstance(typs, Instance):
+                        item_type_of_iterable.append(self.iterable_item_type(typs))
+                        idx_of_iterable.append(idx)
+                    else:
+                        self.fail("StarExpr should not be a '{}'".format(typs), context)
```
JukkaL, line 2501: Nits: StarExpr is an internal name that we'd rather not use in an error message. It would be better to use something more descriptive, such as `Invalid type 'x' for *expr (iterable expected)`

If you pass `rval` as the context to `fail`, the error message should point to the relevant subexpression, making it easier to figure out what it going on.

It would be good to have a test case for this as well.

mypy/checker.py
```diff
@@ -2482,9 +2482,37 @@ def check_assignment_to_multiple_lvalues(self, lvalues: List[Lvalue], rvalue: Ex
...
+            num_every_iterable = 0
+            num_last_iterable = 0
+            if len(idx_of_iterable):
+                num_every_iterable = int(lhs_len / len(idx_of_iterable))
+                num_last_iterable = lhs_len - (len(idx_of_iterable) - 1) * int(num_every_iterable)
+            for i, (idx, item_type) in enumerate(zip(idx_of_iterable, item_type_of_iterable)):
+                if i == (len(idx_of_iterable) - 1):
+                    rvalues[idx:idx] = [TempNode(item_type) for _ in range(num_last_iterable)]
+                else:
+                    rvalues[idx:idx] = [TempNode(item_type) for _ in range(num_every_iterable)]
```
JukkaL, line 2515: I think that the indices can become out of date as this modifies the length of `rvalues`?

test-data/unit/check-tuples.test
```diff
@@ -1421,3 +1421,28 @@ t6: Tuple[int, int, int, int, int, int, int, int, int, int, int, int] = (1, 2, 3
...
+reveal_type(y2) # N: Revealed type is 'builtins.int*'
+reveal_type(z2) # N: Revealed type is 'builtins.str'
+
+x3, x4, y3, y4, z3 = *points, *points2, "test"
+
+reveal_type(x3) # N: Revealed type is 'builtins.int'
+reveal_type(x4) # N: Revealed type is 'builtins.str'
+reveal_type(y3) # N: Revealed type is 'builtins.int*'
+reveal_type(y4) # N: Revealed type is 'builtins.int*'
+reveal_type(z3) # N: Revealed type is 'builtins.str'
```
JukkaL, line 1446: Add test case for two `*expr` that are both variable-length.

PR #8838, opened 2020-05-17
mypy: remove subprocess samples

hauntsaninja:
They depend heavily on _subprocess, which is no longer a module that
exists: https://github.com/python/typeshed/pull/4022#issuecomment-629720860


hauntsaninja:
Seems like we should add in some newer samples. Current subprocess.py and test_subprocess.py don't type check off the bat, but I can twist them until they do

JelleZijlstra:
I'm not sure that's necessary. I think these samples were added as a way to run mypy on a largish codebase when the project was getting started, but now mypy is run on huge codebases that are more useful to find regressions (like Dropbox's codebases). @JukkaL what do you think?

JukkaL:
I don't think that the stdlib samples are very useful any more. We can keep those that can be maintained without much effort, but non-trivial investments into them don't seem worth it.

PR #8845, opened 2020-05-18
Look at arguments when generating constraints for an overload

msullivan:
This improves a lot of cases. I'm not totally sure why we skipped the
arguments in the past.


mypy/constraints.py
```diff
@@ -486,12 +486,11 @@ def infer_against_overloaded(self, overloaded: Overloaded,
                                  template: CallableType) -> List[Constraint]:
         # Create constraints by matching an overloaded type against a template.
         # This is tricky to do in general. We cheat by only matching against
-        # the first overload item, and by only matching the return type. This
+        # the first overload item that is callable compatible. This
```
ilevkivskyi, line 489: This comment is 8 years old. I think this answers your question in the PR description.

PR #8847, opened 2020-05-19
Follow .pth files in site-packages to find additional imports

### LINKED ISSUE #8824: Cannot find imports from .pth files in site-packages folder
Packages can use .pth files to extend locations used for importing. This works at runtime, but breaks mypy's import logic. Mypy is not recognizing PEP 561 packages that are installed into site packages. I tried working around this with MYPYPATH, but it seems like that enters a different import code path that doesn't recognize a PEP 561 package. As a result, I'm getting lots of these errors:

> Cannot find implementation or library stub for module named X

Root cause might be simple. If I manually merge all .pth files in site-packages into one file named easy-install.pth, imports work correctly. Unfortunately I can't easily do this in my development workflow because it was designed to use symlinks to construct site-packages, and merging into one file would be very complicated.

#### Proposal
If I understand correctly, mypy's module finder is only looking for easy-install.pth. According to [Python's docs](https://docs.python.org/3/library/site.html), the .pth file can be named anything. Could this code scan for any file in site-packages that ends with ".pth"?
https://github.com/python/mypy/blob/c90b405934cbf2142b8a575ddd89dbedf762106c/mypy/modulefinder.py#L476-L480

It seems to me this would match how Python implements the .pth import hook: https://github.com/python/cpython/blob/8510f430781118d9b603c3a2f06945d6ebc5fe42/Lib/site.py#L206

natemcmaster:
### What
This will parse _any_ \*.pth file in site-packages, not just easy-install.pth. This matches how Python's stdlib 'site' module handles .pth files in site-packages directories (see https://docs.python.org/3/library/site.html)

### Why
Resolves #8824 

### Testing
Added more test packages to testmodulefinder.py


JukkaL:
I'll review this next week when I have more time. This is non-trivial enough that I need to look at this carefully.

natemcmaster:
@JukkaL thanks for the update!

natemcmaster:
Hey there, anything I can do to help with this? Does this need more test coverage or more documentation? I'm interested in fixing this to improve the mypy coverage on some of my projects, and don't have another way to workaround #8824.

natemcmaster:
Friendly ping! Anything I can do to help here?

gvanrossum:
Not really -- the project is just really behind on PR review.

mypy/modulefinder.py
```diff
@@ -472,15 +466,53 @@ def make_abspath(path: str, root: str) -> str:
...
+
+def expand_site_packages(site_packages: List[str]) -> Tuple[List[str], List[str]]:
+    """Expands .pth imports in site-packages directories"""
+    egg_dirs = []  # type: List[str]
     for dir in site_packages:
-        pth = os.path.join(dir, 'easy-install.pth')
-        if os.path.isfile(pth):
-            with open(pth) as f:
-                egg_dirs.extend([make_abspath(d.rstrip(), dir) for d in f.readlines()])
+        if not os.path.exists(dir):
```
gvanrossum, line 476: I'd use `isdir` here, on the odd chance that somehow it exists as a file, since the `listdir` operation below would raise in that case.

mypy/modulefinder.py
```diff
@@ -472,15 +466,53 @@ def make_abspath(path: str, root: str) -> str:
...
+        for pth_filename in pth_filenames:
+            egg_dirs.extend(_parse_pth_file(dir, pth_filename))
+
     return egg_dirs, site_packages
 
 
+def _parse_pth_file(dir: str, pth_filename: str) -> Iterator[str]:
+    """
+    Mimics a subset of .pth import hook from Lib/site.py
+    cref https://github.com/python/cpython/blob/3.5/Lib/site.py#L146-L185
```
gvanrossum, line 488: "cref" is an obscure abbrev -- I'd just use "See". Also, why reference the 3.5 branch? It's so old!

mypy/modulefinder.py
```diff
@@ -472,15 +466,53 @@ def make_abspath(path: str, root: str) -> str:
...
+        f = open(pth_file, "r")
+    except OSError:
+        return
+    with f:
+        for line in f.readlines():
+            if line.startswith("#"):
+                # Skip comment lines
+                continue
+            if line.startswith(("import ", "import\t")):
+                # import statements in .pth files are not supported"
```
gvanrossum, line 502: Stray trailing `"`.

mypy/test/testmodulefinder.py
```diff
@@ -143,14 +148,15 @@ def setUp(self) -> None:
...
             "modulefinder-site-packages",
         ))
+
+        egg_dirs, site_packages = expand_site_packages([self.package_dir])
+
         self.search_paths = SearchPaths(
             python_path=(),
             mypy_path=(
                 os.path.join(data_path, "pkg1"),
             ),
```
gvanrossum, line 158: Might as well clean up this line (It really doesn't need to be split across three lines).

gvanrossum (commented):
Sorry for the slow review. This is good work; I am happy you included such thorough tests. I only have a few nits. Please fix those and I'll land it.

mypy/modulefinder.py
natemcmaster: I picked the 3.5 branch b/c that's the min version supported by mypy.

mypy/modulefinder.py
natemcmaster: Good catch. I'll update that!

natemcmaster:
@gvanrossum thanks for the review! I've updated per your feedback, and CI checks are passing.

gvanrossum (approved):
Thanks! Let’s wait until CI passes.

gvanrossum:
I think there’s a GitHub or Travis bug that prevents GitHub from picking up status of one of the builds. If I click through, it passed on Travis. So, landing.

gvanrossum:
And thanks for your work!

PR #8866, opened 2020-05-20
Fixes #5705

### LINKED ISSUE #5705: Provide more context about why incompatible with supertype is an error
Users often don't realize that overriding a method with an incompatible signature is unsafe from a type checking perspective; see #1237 for an example. We could generate some more descriptive notes with the error message, perhaps with a link to mypy documentation what contains a detailed discussion of Liskov Substitution Principle. We could also discuss how to work around the issue.

#5704 and #5025 are related issues.

ChetanKhanna:
Added notes as suggested in the issue description after error from
incompatible arguments in subtype is raised.

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -772,6 +772,14 @@ class MessageBuilder:
                   .format(arg_num, name, target, arg_type_in_supertype_f),
                   context,
                   code=codes.OVERRIDE)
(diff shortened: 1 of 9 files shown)


Eric-Arellano:
As a MyPy user, this is awesome! Thank you for doing this!

PR #8870, opened 2020-05-22
[mypyc] Add a primitive for formatting an int as a str

msullivan:
This is partially just shameless cheesing on an example kmod used in a
blog post (http://blog.kevmod.com/2020/05/python-performance-its-not-just-the-interpreter/comment-page-1/#comment-55896),
but it is generally useful. Probably it will be more useful as part of
more general string formatting optimizations, but it is a start.

This gives a 3x speedup on the example in the blog post.

I had been hoping to just allocate a unicode string and call sprintf
into it but that turned out to be quite a bit slower than boxing the
int and calling PyObject_Str on it! So I just implemented it.

### DIFF
--- a/mypyc/lib-rt/CPy.h
+++ b/mypyc/lib-rt/CPy.h
@@ -642,13 +642,12 @@ static CPyTagged CPyTagged_Id(PyObject *o) {
 // using snprintf or PyUnicode_FromFormat was way slower than
 // boxing the int and calling PyObject_Str on it, so we implement our own
 static int fmt_ssize_t(char *out, Py_ssize_t n) {
-	Py_ssize_t in = n;
 	bool neg = n < 0;
 	if (neg) n = -n;
-	char buf[MAX_INT_CHARS];
-	int i = 0;
 
 	// buf gets filled backward and then we copy it forward
+	char buf[MAX_INT_CHARS];
+	int i = 0;
 	do {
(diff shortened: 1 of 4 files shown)


msullivan:
Oh lurr I need to make it *not* trigger for bools. That's kind of silly. Tomorrow.

JukkaL:
I checked this with another benchmark that uses smaller ints, and it was about 4.5x faster than CPython 3.7. The new version also was 2x faster than mypyc without this optimization. Nice!

We could perhaps do even better by avoiding relatively the slow % operations (https://stackoverflow.com/questions/4351371/c-performance-challenge-integer-to-stdstring-conversion has many examples), but let's not get carried away :-)

hauntsaninja:
From your comment:
> we trade it off for increased I$ pressure!
I was unable to google this, is this register pressure?

msullivan:
I$ is an silly abbreviation for "instruction cache". (Likewise, D$ for "data cache".)

msullivan:
Ha, there was another issue which is that when doing `str(0)` I didn't null terminate the string! If you don't null terminate the buffer in a python string, you get a super cursed string where almost everything works except that there is plenty of C code that expects nul termination, including str-to-int conversion!

(And it only happened for 0 because I apparently forgot that `do ... while` exists.)

Issue #8625, opened 2020-04-03
False positive with attrs converter


tsx:
Looks like mypy can't infer generic parameter properly when used in attrs converter.

```python
from attr import attrs, attrib
from typing import FrozenSet, Text


@attrs
class Test(object):
    values = attrib(type=FrozenSet[Text], converter=frozenset)


Test(values={'a'})
```

**Actual**: reporting error
```
error: Argument 1 to <set> has incompatible type "str"; expected "_T"
```

**Expected**: no errors.

mypy version 0.770, python 3.5.2, no flags are used.
Latest master has basically the same error message except generic param has a different name:
```
error: Argument 1 to <set> has incompatible type "str"; expected "_T_co"
```

Possibly related: https://github.com/python/mypy/issues/5313

msullivan:
Yeah, that does look wrong. Not sure how easy fixing it would be.

markusschmaus:
This also happens in case of a generic function.
```python
from attr import attrs, attrib
from typing import FrozenSet, Text, Iterable, TypeVar

_T = TypeVar('_T')
def converter(ar: Iterable[_T]) -> FrozenSet[_T]:
    return frozenset(ar)

@attrs
class Test(object):
    values = attrib(type=FrozenSet[Text], converter=converter)


Test(values={'a'})
```

results in an error:

```
error: Type variable "scratch_11._T" is unbound
```


I looked into how this could be fixed and here is what I found out for reference:
The attrs plugin gets called during 

https://github.com/python/mypy/blob/41927cf6dbbc87728eab632486372971c45999cb/mypy/semanal_main.py#L78

At this point the type of the `converter` isn't fully processed so when attrs looks at its type at 

https://github.com/python/mypy/blob/41927cf6dbbc87728eab632486372971c45999cb/mypy/plugins/attrs.py#L98 

`converter.is_generic` returns `False` even though the function is generic. The full type of this function is only established during 

https://github.com/python/mypy/blob/41927cf6dbbc87728eab632486372971c45999cb/mypy/semanal_main.py#L79

Which the attrs plugin has already completed.

markusschmaus:
I opened a PR which fixes this issue for builtin types, including `frozenset`. https://github.com/python/mypy/pull/8876/files

It doesn't fix the issue for generic functions, since their type still contains `UnboundType`s when the attr plugin is run.

ryansobol:
Just ran into this bug. Thank you so much for submitting a PR!

PR #8876, opened 2020-05-23
Partially fixes issue 8625

markusschmaus:
This fixes https://github.com/python/mypy/issues/8625 for instances when the converter is a builtin type like `typle`, `list`, or `dict`. 

This is only a partial fix, since for user defined generic converters a similar issue still persists.


markusschmaus:
Is there anything I can do to get this PR reviewed?

mypy/plugins/attrs.py
```diff
@@ -57,6 +60,43 @@ def __init__(self,
...
+
+    if ret_type == target_type or isinstance(ret_type, AnyType):
+        # If the callable has the exact same return type as the target
+        # we can directly return the type of the first argument.
+        # This is also the case if the return type is `Any`.
+        return arg_type
+
+    # Find the constraints on the type vars given that the result must be a
+    # subtype of the target type.
+    constraints = infer_constraints(ret_type, target_type, SUBTYPE_OF)
```
JukkaL, line 82: The relevant plugin hook gets called during semantic analysis, and solving constraints is unfortunately not supported during semantic analysis, since some type definitions may be incomplete. (This is pretty confusing since most simple cases work just fine, but this will fail in some edge cases.)

Basically this would need to be implemented without using any non-trivial type operations, perhaps by hard coding the behavior for a set of builtin collections.

markusschmaus: I am aware that when this gets called with a user defined converter with type variables the type definition is incomplete and doesn't contain enough information to actually solve the issue. However for builtin collections there is enough information available. 

I started out by hard coding the behavior for a set of common builtin collections as you are suggesting, but soon realized that the code I was writing was identical to `infer_constraints` and that for builtin collections it works as long as I manually construct the list of type variables (`type_var_ids = list({const.type_var for const in constraints})`).

That's why this is a partial fix. But it since this works for builtin collections, it does address the most important use case. And for all other instances it fails just as before.

markusschmaus: @JukkaL Would this PR be acceptable for you if I duplicated the code of `infer_constraints` in a separate function to avoid any code coupling? As you say, we cannot solve this issue in general because when the plugin hook gets called type definitions may be incomplete, but we can solve this for this for builtin collections, and this PR does exactly that.

PR #8880, opened 2020-05-24
[mypyc]Implement CallC IR

TH3CHARLie:
relates mypyc/mypyc#709

This PR adds a new IR op `CallC` to replace some `PrimitiveOp` that simply calls a C function. To demonstrate this prototype, `str.join` primitive is now switched from `PrimitiveOp` to `CallC`, with identical generated C code:

Test driver code
```python
from typing import List

def test_str_join_helper(helper_arg: List[str]) -> str:
    helper_base = "#"
    return helper_base.join(helper_arg)

def main():
    l = ["a", "bb", "cc"]
    print(test_str_join_helper(l))

main()
```

generated C code(for the helper function, identical before/after):
```C
PyObject *CPyDef_test_str_join_helper(PyObject *cpy_r_helper_arg) {
    PyObject *cpy_r_r0;
    PyObject *cpy_r_helper_base;
    PyObject *cpy_r_r1;
    PyObject *cpy_r_r2;
CPyL0: ;
    cpy_r_r0 = CPyStatic_unicode_3; /* '#' */
    CPy_INCREF(cpy_r_r0);
    cpy_r_helper_base = cpy_r_r0;
    cpy_r_r1 = PyUnicode_Join(cpy_r_helper_base, cpy_r_helper_arg);
    CPy_DecRef(cpy_r_helper_base);
    if (unlikely(cpy_r_r1 == NULL)) {
        CPy_AddTraceback("foo.py", "test_str_join_helper", 5, CPyStatic_globals);
        goto CPyL2;
    } else
        goto CPyL1;
CPyL1: ;
    return cpy_r_r1;
CPyL2: ;
    cpy_r_r2 = NULL;
    return cpy_r_r2;
}
```

Some objectives need to be completed after some discussion.

- [x] figure out the differences between`OpDescrption` and new `LLOpDescrption`, do we need fields including `is_var_arg`, `steals`, `is_borrowed`?(Not supported in this PR yet, but we will gradually update the design when we encounter ops that require such changes)
- [x] support void functions(through a flag?)
- [x] textual IR testcase

### DIFF
--- a/mypyc/ir/ops.py
+++ b/mypyc/ir/ops.py
@@ -1150,7 +1150,6 @@ class CallC(RegisterOp):
         super().__init__(line)
         self.function_name = function_name
         self.args = args
-        # TODO: handle void C function by adding a new flag and make this Optional
         self.type = ret_type
 
     def to_str(self, env: Environment) -> str:
--- a/mypyc/ir/ops.py
+++ b/mypyc/ir/ops.py
@@ -1138,7 +1138,7 @@ class RaiseStandardError(RegisterOp):
         return visitor.visit_raise_standard_error(self)
 
 
-class CFunctionCall(RegisterOp):
+class CallC(RegisterOp):
     """ret = func_call(arg0, arg1, ...)
 
     A call to a C function
@@ -1155,14 +1155,13 @@ class CFunctionCall(RegisterOp):
 
     def to_str(self, env: Environment) -> str:
         args_str = ', '.join(env.format('%r', arg) for arg in self.args)
-        # TODO: comment: c_function_call to distinguish from cast, box, etc
-        return env.format('%r = %s(%s) :: c_function_call', self, self.function_name, args_str)
+        return env.format('%r = %s(%s)', self, self.function_name, args_str)
 
     def sources(self) -> List[Value]:
         return self.args
 
     def accept(self, visitor: 'OpVisitor[T]') -> T:
-        return visitor.visit_c_function_call(self)
+        return visitor.visit_call_c(self)
 
 
 @trait
@@ -1256,7 +1255,7 @@ class OpVisitor(Generic[T]):
         raise NotImplementedError
 
     @abstractmethod
-    def visit_c_function_call(self, op: CFunctionCall) -> T:
+    def visit_call_c(self, op: CallC) -> T:
         raise NotImplementedError
 
 
--- a/mypyc/ir/ops.py
+++ b/mypyc/ir/ops.py
@@ -1138,6 +1138,33 @@ class RaiseStandardError(RegisterOp):
         return visitor.visit_raise_standard_error(self)
 
 
+class CFunctionCall(RegisterOp):
+    """ret = func_call(arg0, arg1, ...)
+
+    A call to a C function
+    """
+
+    error_kind = ERR_MAGIC
+
+    def __init__(self, function_name: str, args: List[Value], ret_type: RType, line: int) -> None:
+        super().__init__(line)
+        self.function_name = function_name
+        self.args = args
+        # TODO: handle void C function by adding a new flag and make this Optional
+        self.type = ret_type
+
+    def to_str(self, env: Environment) -> str:
+        args_str = ', '.join(env.format('%r', arg) for arg in self.args)
+        # TODO: comment: c_function_call to distinguish from cast, box, etc
+        return env.format('%r = %s(%s) :: c_function_call', self, self.function_name, args_str)
+
+    def sources(self) -> List[Value]:
+        return self.args
+
+    def accept(self, visitor: 'OpVisitor[T]') -> T:
+        return visitor.visit_c_function_call(self)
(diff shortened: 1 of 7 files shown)


mypyc/ir/ops.py
```diff
@@ -1138,6 +1138,33 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
         return visitor.visit_raise_standard_error(self)
 
 
+class CFunctionCall(RegisterOp):
```
JukkaL, line 1141: Bikeshedding: for consistency with the existing `Call` primitive, a name such as `CallC` would arguably be more consistent. Later on we may want to merge `Call` and `CallC`.

mypyc/ir/ops.py
```diff
@@ -1138,6 +1138,33 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
...
+        super().__init__(line)
+        self.function_name = function_name
+        self.args = args
+        # TODO: handle void C function by adding a new flag and make this Optional
+        self.type = ret_type
+
+    def to_str(self, env: Environment) -> str:
+        args_str = ', '.join(env.format('%r', arg) for arg in self.args)
+        # TODO: comment: c_function_call to distinguish from cast, box, etc
+        return env.format('%r = %s(%s) :: c_function_call', self, self.function_name, args_str)
```
JukkaL, line 1159: I think that the `:: c_function_call` suffix is too verbose. We can leave it out for now, even though this results in some ambiguity. f/when we merge `Call` and this new class, we can figure out how to disambiguate different names.

mypyc/primitives/registry.py
```diff
@@ -35,13 +35,21 @@
...
 
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, NamedTuple
 
 from mypyc.ir.ops import (
     OpDescription, EmitterInterface, EmitCallback, StealsDescription, short_name
 )
 from mypyc.ir.rtypes import RType,  bool_rprimitive
 
+# TODO: comment, we don't need error_kind since C functions are all ERR_MAGIC
```
JukkaL, line 45: We'll want to have C functions that are not ERR_MAGIC. Some things will never return an error, for example.

mypyc/ir/ops.py
```diff
@@ -1228,6 +1255,10 @@ def visit_unbox(self, op: Unbox) -> T:
     def visit_raise_standard_error(self, op: RaiseStandardError) -> T:
         raise NotImplementedError
 
+    @abstractmethod
+    def visit_c_function_call(self, op: CFunctionCall) -> T:
```
JukkaL, line 1259: If you rename the class, also rename everything else that has a name derived from the old name.

mypyc/primitives/registry.py
```diff
@@ -35,13 +35,21 @@
...
+# however, other fields may need further investigation
+LLOpDescription = NamedTuple(
```
JukkaL, line 47: This name doesn't fit the current implementation. Can you propose a new name that communicates that this specifically describes how to call a C function?

TH3CHARLie: I am using `CFunctionDescription` in the updated commits.

mypyc/primitives/registry.py
```diff
@@ -35,13 +35,22 @@
...
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, NamedTuple
 
 from mypyc.ir.ops import (
     OpDescription, EmitterInterface, EmitCallback, StealsDescription, short_name
 )
 from mypyc.ir.rtypes import RType,  bool_rprimitive
 
+# TODO: comment, we don't need error_kind since C functions are all ERR_MAGIC
+# however, other fields may need further investigation
```
JukkaL, line 46: This comment is out of date.

mypyc/primitives/registry.py
```diff
@@ -58,6 +67,8 @@
 # Primitive ops for reading module attributes (key is name such as 'builtins.None')
 name_ref_ops = {}  # type: Dict[str, OpDescription]
 
+call_c_ops = {}  # type: Dict[str, List[CFunctionDescription]]
```
JukkaL, line 70: Since we store a method call here, the name is the short name of a method (i.e, not a fully qualified name such as `builtins.ord`). For consistency with the existing registries, a better name would be `c_method_call_ops` or similar. We can then have a separate dictionary for module-level functions.

mypyc/primitives/registry.py
```diff
@@ -312,6 +323,18 @@ def custom_op(arg_types: List[RType],
                          emit, steals, is_borrowed, 0)
 
 
+def call_c_op(name: str,
```
JukkaL, line 326: Similar to above, something with 'method' in it would be a better name. For example, `c_method_op`. (These names will be temporary as we'll eventually remove `method_op`.)

TH3CHARLie:
Thanks for the review!

PR #8886, opened 2020-05-25
stubtest: fix false negative with dunder methods, small changes

hauntsaninja:
If a stub was completely missing a dunder method, we wouldn't check it against the runtime. It's noisy to check all dunder methods, but checking init, new and call catches a lot of false negatives on typeshed without much noise.

### DIFF
--- a/mypy/stubtest.py
+++ b/mypy/stubtest.py
@@ -662,15 +662,6 @@ def verify_funcitem(
 def verify_none(
     stub: Missing, runtime: MaybeMissing[Any], object_path: List[str]
 ) -> Iterator[Error]:
-    if isinstance(runtime, Missing):
-        try:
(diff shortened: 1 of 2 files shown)


JelleZijlstra:
The Appveyor CI failure isn't real. Looks like we may have disabled the Appveyor project but we still try to trigger it.

Out of curiosity, what are the noisy results from enabling this for other dunders?

hauntsaninja:
I wasn't too thorough, but there were a number of complaints about `__hash__` and `__doc__`. Using @functools.total_ordering causes complaints about the various comparators. One or two other things. Maybe a blacklist would be better than a whitelist.

hauntsaninja:
#8786 just got merged, so let's try CI again

hauntsaninja:
cc @JelleZijlstra not sure how to request review, but everything is green now

PR #8888, opened 2020-05-26
Fix type expected by stubgenc

### LINKED ISSUE #7692: stubgen and C Extensions (pyd)
I work on a project called python-openzwave. I have rewritten the build system for this library. The library makes use of a CPP library called OpenZWave. This library gets compiled by the build system. Cython is what is used for the "bridge" between python code and the cpp code in OpenZWave. The output of this process is a single pyd file. I would like to be able to create a stub file for this pyd file. I have encountered an issue when creating documentation for the library where if the library is already installed on the system and the build system imports the newly created pyd file python crashes. So a direct import of the new file cannot be done. Using the code example below I am able to import the new library and avoid the application crash.

```
import os
import imp

mod_file = os.path.join(build_dir, 'libopenzwave.pyd')
mod = imp.load_dynamic(
    'libopenzwave',
    mod_file
)
    
sys.modules['libopenzwave'] = mod
```

I am unsure of how stubgen handles the importation of c extensions. I am needing to build the stub file at build time so I have subclassed setuptools.Command to acomplish this task,

the code below is what is used in the `run` method of the class

```
        from mypy import stubgen
        import sys

        builder = self.distribution.get_command_obj('build')
        build_path = builder.build_lib
        build_ext = self.distribution.get_command_obj('build_ext')
        extension = self.distribution.ext_modules[0]
        ext_path = build_ext.get_ext_fullpath(extension.name)

        if sys.version_info[0] < 3:
            args = ['--py2']
        else:
            args = []

        args += [
            '--include-private',
            '--output',
            build_path,
            ext_path
        ]

        options = stubgen.parse_options(args)
        stubgen.generate_stubs(options)
```

when this runs the program exits with the following error

```
Critical error during semantic analysis: mypy: can't decode file 'build\lib.win-amd64-3.7\libopenzwave.pyd': 'utf-8' codec can't decode byte 0x90 in position 2: invalid start byte
```

using the positional file is the only way to be able to direct the program to load the proper pyd file. Is there a special way to go about passing the path and filename of the c extension that I want to create the stub file for? or do I have to use the imp module as outlines in the code block above and set the module into sys.modules and then pass the module name to stubgen?

henryiii:
In investigating a crash when running on a PyBind11 module, I noticed a mismatch in the types; `infer_prop_type_from_docstring` can clearly get None, but it has a type of plain `str`. Fix the annotation to match the expectation. Now, when rebuilding Mypy, the error goes away and this now works without crashing, and can completely generate stubs.

Here's one of the uses where None is obviously possible:

https://github.com/python/mypy/blob/d1048867ce27f607a7f48edf33a1888789b187e7/mypy/stubgenc.py#L220-L221

Fixes #7692.

Example printout when this error is seen:

```
    sys.exit(main())
  File "mypy/stubgen.py", line 1564, in main
  File "mypy/stubgen.py", line 1457, in generate_stubs
  File "mypy/stubgenc.py", line 51, in generate_stub_for_c_module
  File "mypy/stubgenc.py", line 273, in generate_c_type_stub
  File "mypy/stubgenc.py", line 221, in generate_c_property_stub
TypeError: str object expected; got None
```

### DIFF
--- a/mypy/stubdoc.py
+++ b/mypy/stubdoc.py
@@ -339,7 +339,7 @@ def find_unique_signatures(sigs: Sequence[Sig]) -> List[Sig]:
     return sorted(result)
 
 
-def infer_prop_type_from_docstring(docstr: str) -> Optional[str]:
+def infer_prop_type_from_docstring(docstr: Optional[str]) -> Optional[str]:
     """Check for Google/Numpy style docstring type annotation for a property.
 
     The docstring has the format "<type>: <descriptions>".


JelleZijlstra:
Looks good. I wonder why mypy itself didn't catch that—are we running with strict-optional off perhaps?

t-fi:
Hi, will this PR be merged? It looks quite simple and finished, or is there some work to be done still? I am interested since I have this issue when running `stubgen` on several different modules.

PR #8902, opened 2020-05-27
[mypyc] Support top level function ops via CallC

TH3CHARLie:
relates mypyc/mypyc#709

This PR supports top-level function ops via recently added `CallC` IR. To demonstrate the idea, it transform `to_list` op from `PrimitiveOp` to `CallC`. It also refines `CallC` with arguments coercing and support of `steals`.

### DIFF
--- a/mypyc/irbuild/builder.py
+++ b/mypyc/irbuild/builder.py
@@ -43,7 +43,7 @@ from mypyc.ir.rtypes import (
 )
 from mypyc.ir.func_ir import FuncIR, INVALID_FUNC_DEF
 from mypyc.ir.class_ir import ClassIR, NonExtClassInfo
-from mypyc.primitives.registry import func_ops
+from mypyc.primitives.registry import func_ops, CFunctionDescription, c_function_ops
 from mypyc.primitives.list_ops import list_len_op, to_list, list_pop_last
 from mypyc.primitives.dict_ops import dict_get_item_op, dict_set_item_op
 from mypyc.primitives.generic_ops import py_setattr_op, iter_op, next_op
@@ -228,6 +228,9 @@ class IRBuilder:
 
     def load_module(self, name: str) -> Value:
         return self.builder.load_module(name)
+
+    def call_c(self, desc: CFunctionDescription, args: List[Value], line: int) -> Value:
+        return self.builder.call_c(desc.c_function_name, args, line, desc.result_type)
 
     @property
     def environment(self) -> Environment:
@@ -506,7 +509,7 @@ class IRBuilder:
         # Assign the starred value and all values after it
         if target.star_idx is not None:
             post_star_vals = target.items[split_idx + 1:]
-            iter_list = self.primitive_op(to_list, [iterator], line)
+            iter_list = self.call_c(to_list, [iterator], line)
             iter_list_len = self.primitive_op(list_len_op, [iter_list], line)
             post_star_len = self.add(LoadInt(len(post_star_vals)))
             condition = self.binary_op(post_star_len, iter_list_len, '<=', line)
@@ -723,6 +726,10 @@ class IRBuilder:
 
         # Handle data-driven special-cased primitive call ops.
         if callee.fullname is not None and expr.arg_kinds == [ARG_POS] * len(arg_values):
+            call_c_ops_candidates = c_function_ops.get(callee.fullname, [])
+            target = self.builder.matching_call_c(call_c_ops_candidates, arg_values, expr.line)
+            if target:
+                return target
             ops = func_ops.get(callee.fullname, [])
             target = self.builder.matching_primitive_op(
                 ops, arg_values, expr.line, self.node_type(expr)
--- a/mypyc/irbuild/builder.py
+++ b/mypyc/irbuild/builder.py
@@ -230,7 +230,7 @@ class IRBuilder:
         return self.builder.load_module(name)
 
     def call_c(self, desc: CFunctionDescription, args: List[Value], line: int) -> Value:
-        return self.builder.call_c(desc.c_function_name, args, line, desc.result_type)
+        return self.builder.call_c(desc, args, line)
 
     @property
     def environment(self) -> Environment:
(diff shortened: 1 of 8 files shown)


TH3CHARLie:
Matching `func_ops`(where `to_list` op originally resides) happens in two places: `ll_builder/builtin_call` and `builder/call_refexpr_with_args`. I changed them accordingly, but I also observe that for `to_list` case(at least for simple cases I tested), only the latter one is doing the work.

JukkaL:
Let me know when you are ready with the PR (it now says WIP).

TH3CHARLie:
> Let me know when you are ready with the PR (it now says WIP).

Yes, I'd say it's WIP because I intend to include `list_set_item_op` as well and there are some other things may need to discuss.

TH3CHARLie:
I think it's now ready for review(assuming all tests passed).

mypyc/primitives/list_ops.py
```diff
@@ -83,7 +84,6 @@ def emit_new(emitter: EmitterInterface, args: List[str], dest: str) -> None:
     error_kind=ERR_FALSE,
     emit=call_emit('CPyList_SetItem'))
 
-
```
TH3CHARLie, line 86: I delete a blank line here and I find that one/two blank line(s) is used interchangeably. I am surprised that flake8 is OK with this

mypyc/irbuild/ll_builder.py
```diff
@@ -768,8 +775,7 @@ def translate_special_method_call(self,
         """
         ops = method_ops.get(name, [])
         call_c_ops_candidates = c_method_call_ops.get(name, [])
-        call_c_op = self.matching_call_c(call_c_ops_candidates, [base_reg] + args, line,
-                                         result_type=result_type)
+        call_c_op = self.matching_call_c(call_c_ops_candidates, [base_reg] + args, line)
```
JukkaL, line 778: Result type seems important to preserve -- we may need it to box/unbox the result?

mypyc/irbuild/builder.py
```diff
@@ -723,6 +726,10 @@ def call_refexpr_with_args(
 
         # Handle data-driven special-cased primitive call ops.
         if callee.fullname is not None and expr.arg_kinds == [ARG_POS] * len(arg_values):
+            call_c_ops_candidates = c_function_ops.get(callee.fullname, [])
+            target = self.builder.matching_call_c(call_c_ops_candidates, arg_values, expr.line)
```
JukkaL, line 730: We should probably pass the expected type here, similar to `matching_primitive_op` below (`self.node_type(expr)`).

mypyc/irbuild/ll_builder.py
```diff
@@ -649,20 +653,23 @@ def add_bool_branch(self, value: Value, true: BasicBlock, false: BasicBlock) ->
...
-               function_name: str,
+               desc: CFunctionDescription,
                args: List[Value],
-               line: int,
-               result_type: Optional[RType]) -> Value:
+               line: int) -> Value:
         # handle void function via singleton RVoid instance
-        ret_type = void_rtype if result_type is None else result_type
-        target = self.add(CallC(function_name, args, ret_type, line))
+        ret_type = void_rtype if desc.result_type is None else desc.result_type
```
JukkaL, line 660: Maybe simplify this by requiring an explicit `void_rtype` instead of `None` as the `result_type`?

mypyc/irbuild/ll_builder.py
```diff
@@ -649,20 +653,23 @@ def add_bool_branch(self, value: Value, true: BasicBlock, false: BasicBlock) ->
         self.add(Branch(value, true, false, Branch.BOOL_EXPR))
 
     def call_c(self,
-               function_name: str,
+               desc: CFunctionDescription,
                args: List[Value],
-               line: int,
-               result_type: Optional[RType]) -> Value:
```
JukkaL, line 655: I think that this is still useful to have. This would be the type the result should be coerced into, which can be different from the return type of the called function.

mypyc/irbuild/ll_builder.py
```diff
@@ -649,20 +653,23 @@ def add_bool_branch(self, value: Value, true: BasicBlock, false: BasicBlock) ->
...
         # handle void function via singleton RVoid instance
-        ret_type = void_rtype if result_type is None else result_type
-        target = self.add(CallC(function_name, args, ret_type, line))
+        ret_type = void_rtype if desc.result_type is None else desc.result_type
+        coerced = []
+        for i, arg in enumerate(args):
+            formal_type = desc.arg_types[i]
+            arg = self.coerce(arg, formal_type, line)
+            coerced.append(arg)
+        target = self.add(CallC(desc.c_function_name, coerced, ret_type, desc.steals, line))
```
JukkaL, line 666: Also coerce the result into `result_type`? Maybe rename `desc.result_type` to `desc.return_type` or something to avoid confusion. Return type would be the type of the value returned by the function, and result type would be the type that the return value needs to be coerced to, based on the context.

mypyc/irbuild/ll_builder.py
```diff
@@ -679,7 +686,7 @@ def matching_call_c(self,
                 else:
                     matching = desc
         if matching:
-            target = self.call_c(matching.c_function_name, args, line, result_type)
+            target = self.call_c(matching, args, line)
```
JukkaL, line 689: Again, propagating the result type seems important.

mypyc/primitives/registry.py
```diff
@@ -326,11 +331,27 @@ def c_method_op(name: str,
...
     desc = CFunctionDescription(name, arg_types, result_type,
-                                c_function_name, error_kind, priority)
+                                c_function_name, error_kind, steals, priority)
     ops.append(desc)
+    return desc
+
+
+def c_function_op(name: str,
+                  arg_types: List[RType],
+                  result_type: Optional[RType],
```
JukkaL, line 345: As discussed above, maybe rename this to `return_type`.

mypyc/irbuild/ll_builder.py
TH3CHARLie: Yes, you are right. I misunderstand this, I thought that return type in the description provided enough information but it's clearly not the case.

mypyc/irbuild/ll_builder.py
TH3CHARLie: This is super reasonable and I clearly miss that distinction.

mypyc/irbuild/ll_builder.py
TH3CHARLie: `desc.return_type`(after naming) is of type `RType`, for void functions, its description should be set to `void_rtype`

TH3CHARLie:
Thanks for the review!

PR #8903, opened 2020-05-27
[mypyc] Optimize two-argument super() for the simple case

msullivan:


mypyc/irbuild/expression.py
```diff
@@ -241,8 +241,28 @@ def translate_method_call(builder: IRBuilder, expr: CallExpr, callee: MemberExpr
 
 
 def translate_super_method_call(builder: IRBuilder, expr: CallExpr, callee: SuperExpr) -> Value:
-    if callee.info is None or callee.call.args:
+    if callee.info is None or (callee.call.args and len(callee.call.args) != 2):
```
ilevkivskyi, line 244: Can `args` ever be `None`? If not, you can just check the `len()`.

PR #8904, opened 2020-05-27
Don't use OrderedDict on Python 3.6 and greater

msullivan:
OrderedDict is written in Python and maintains its own ordering on top
of the ordering the dict provides in recent Pythons. This ordering can
differ if certain methods are called, but we don't need that. Just
always use dict when it is ordered.

### DIFF
--- a/mypy/ordered_dict.py
+++ b/mypy/ordered_dict.py
@@ -1,0 +1,9 @@
+# OrderedDict is kind of slow, so for most of our uses in Python 3.6
+# and later we'd rather just use dict
+
+import sys
+
+if sys.version_info <= (3, 5):
+    from collections import OrderedDict as OrderedDict
+else:
+    OrderedDict = dict
--- a/mypy/ordered_dict.py
+++ b/mypy/ordered_dict.py
@@ -3,7 +3,7 @@
 
 import sys
 
-if sys.version_info <= (3, 5):
+if sys.version_info < (3, 6):
     from collections import OrderedDict as OrderedDict
 else:
     OrderedDict = dict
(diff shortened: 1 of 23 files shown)


msullivan:
Oh wow that is a spicy failure:
```
build/__native_4a7a5cbf27265f591fac.h:1146:5: error: unknown type name 'tuple_T2II'

    tuple_T2II _python_version;
```

mypy/ordered_dict.py
```diff
@@ -0,0 +1,9 @@
+# OrderedDict is kind of slow, so for most of our uses in Python 3.6
+# and later we'd rather just use dict
+
+import sys
+
+if sys.version_info <= (3, 5):
```
JelleZijlstra, line 6: This won't work right because in reality version_info is something like `(3, 6, 10, "final", 0)`, which for any kind of 3.5 is `> (3, 5)`. You need to write `< (3, 6)` instead.

msullivan: Yeah so it seems. Fun fact: I think mypy doesn't understand this.

msullivan:
OK I fixed the obvious major issue and hopefully that fixes the compile error. I wasn't able to reproduce that error with 3.5.5, and 3.5.1 is hard to get running because of SSL stuff...

mypy/ordered_dict.py
JelleZijlstra: We put checks against this sort of thing in flake8-pyi (https://github.com/ambv/flake8-pyi/blob/master/pyi.py#L230) but obviously that doesn't help here since this isn't a stub. Maybe mypy itself should have the same checks.

PR #8907, opened 2020-05-28
Debugging code to log PyObject_GetAttr and python method calls

msullivan:


TH3CHARLie:
> Note that this will need to be rewritten for the low-level IR that @TH3CHARLie is working on, since we won't be able to use C macros. We'd have to deal with this in the IR builder, and provide some other way of enabling this (can't use a `#ifdef`).

I am not sure that I understand why we won't be able to use C macros, at least for now. `CallC` literally generates a call to a function represented by a string. It has no knowledge whether it's a function or macro. However, I assume this is not semantically correct, since we want to have a one-to-one correspondence between an actually C function and a `CallC`, maybe introducing macro contradicts this idea.

JukkaL:
You are correct that macros will continue to work as long as we generate C, so this isn't urgent as long as we only have a C back end. However, one of the objectives is to enable non-C backends, and they won't be able to use macros. Anything that currently uses a macro will need to be either converted to a function (in cases where the performance impact will be minimal) or replaced with some `irbuild` logic that replicates what the macro does.

We'll also want to replace some inline C functions with `irbuild` logic, I think, at least eventually. I don't think that there are very many inline functions that we use, but there are maybe a few dozen? I'm not sure if we'll have time migrate all of them this summer.

msullivan:
Sure. If this only works with the C backend that isn't anything like the end of the world, though.

PR #8911, opened 2020-05-28
Update .gitignore to include _build/ folder from docs/source

ChetanKhanna:
While sending a PR for a documentation update I noticed an untracked ``docs/source/_build/`` folder, containing html files for the documentation. I added the same to gitignore so to avoid them getting pushed mistakenly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ __pycache__
 /build
 /env*/
 docs/build/
+docs/source/_build
 *.iml
 /out/
 .venv*/


TH3CHARLie:
I am curious about how you compile/generate the HTML files, would you mind sharing the commands you run?

ChetanKhanna:
I didn't run any particular command other than the usual git commands after making the changes. The only thing unusual  was probably a sphinx popup on my vs code editor while opening the rst file which I might have clicked :thinking:
Unfortunately I don't have any knowledge of sphinx so I don't have any further details

TH3CHARLie:
> I didn't run any particular command other than the usual git commands after making the changes. The only thing unusual was probably a sphinx popup on my vs code editor while opening the rst file which I might have clicked 🤔
> Unfortunately I don't have any knowledge of sphinx so I don't have any further details

What I originally mean is that, for usual cases, people build docs by:
```
cd docs
make html
```
Following this workflow usually doesn't generate anything that hasn't been git-ignored. In your cases, I bet the build folder is generated by vscode plugin.

ChetanKhanna:
> What I originally mean is that, for usual cases, people build docs by:
> 
> ```
> cd docs
> make html
> ```
> 
> Following this workflow usually doesn't generate anything that hasn't been git-ignored. In your cases, I bet the `docs/source/_build` folder is generated by vscode plugin.

Agreed since I never executed these commands.

JukkaL:
Ignoring this directory seems no big deal even if it's not necessary when using the official workflow.

PR #8914, opened 2020-05-28
[mypyc] Basic support for protocols

msullivan:
The idea is to represent a protocol type P as Union[P, object], so
that we'll make fast calls if a class explictly inherits from the
protocol and fall back to generic ops otherwise.


sobolevn:
@msullivan great timing! 😄 

Thanks a lot!

sobolevn:
Oh, this didn't make its path into `0.780` 😞

Issue #7405, opened 2019-08-29
Explain that mypy should be run in the package root directory


JukkaL:
If I have a package `pkg/` and I run mypy while the current working directory is `pkg`, mypy can't resolve imports from `pkg`. This is consistent with how Python works (see #7389 for discussion), but it may be surprising for some users. Mypy could detect this issue and generate a note that would tell the user to switch to another directory.

More details: If an imported module can't be resolved, but the current working directory contains a `__init__.py[i]` file, see if the import could be resolved if some parent directory was added to the search path. If yes, add a note suggesting that the current working directory should perhaps be changed to a parent directory.

Shashankjain12:
hello @JukkaL  can you assign me the issue so that I can start working on this paricular issue?

JelleZijlstra:
@Shashankjain12 feel free to just start working on this.

Shashankjain12:
hello @JelleZijlstra  i am new to open source can you suggest me which file to edit in order to prompt users that they are in wrong directory while executing the code?

JukkaL:
@Shashankjain12 You'll probably need to edit `mypy/build.py` (and/or maybe some other module imported there). One way to move forward is to find the code that generates the error message about the import, and look around there.

PR #8927, opened 2020-05-31
Improve missing module error for subdirectories (#7405)

ethan-leba:
Fixes https://github.com/python/mypy/issues/7405

### DIFF
--- a/mypy/modulefinder.py
+++ b/mypy/modulefinder.py
@@ -49,10 +49,18 @@ class ModuleNotFoundReason(Enum):
     # corresponding *-stubs package.
     FOUND_WITHOUT_TYPE_HINTS = 1
 
+    # The module was not found in the current working directory, but
+    # was able to be found in the parent directory.
+    WRONG_WORKING_DIRECTORY = 2
+
     def error_message_templates(self) -> Tuple[str, str]:
         if self is ModuleNotFoundReason.NOT_FOUND:
             msg = "Cannot find implementation or library stub for module named '{}'"
             note = "See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports"
+        elif self is ModuleNotFoundReason.WRONG_WORKING_DIRECTORY:
+            msg = "Cannot find implementation or library stub for module named '{}'"
+            note = ("You may be running mypy in a subpackage, "
+                    "mypy should be run on the package root")
         elif self is ModuleNotFoundReason.FOUND_WITHOUT_TYPE_HINTS:
             msg = "Skipping analyzing '{}': found module but no type hints or library stubs"
             note = "See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports"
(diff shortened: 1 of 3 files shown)


ethan-leba:
Thanks, would be happy to take any suggestions on the wording of the message!

mypy/modulefinder.py
```diff
@@ -49,10 +49,18 @@ class ModuleNotFoundReason(Enum):
...
+    # was able to be found in the parent directory.
+    WRONG_WORKING_DIRECTORY = 2
+
     def error_message_templates(self) -> Tuple[str, str]:
         if self is ModuleNotFoundReason.NOT_FOUND:
             msg = "Cannot find implementation or library stub for module named '{}'"
             note = "See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports"
+        elif self is ModuleNotFoundReason.WRONG_WORKING_DIRECTORY:
+            msg = "Cannot find implementation or library stub for module named '{}'"
+            note = ("You may be running mypy in the wrong directory, "
```
TH3CHARLie, line 62: Some brainstorm: `You may be running mypy in a subdirectory, mypy should be run in the package root directory`. @JukkaL may have better suggestions.

ethan-leba: Any suggestions on this message, or other changes that need to be made in order for this PR to be ready for merge @JukkaL?

mypy/test/testcmdline.py
```diff
@@ -26,6 +26,14 @@
     'reports.test',
 ]
 
+# Runs the test in the specified working directory
+custom_working_dir = {
+    "testFailedImportCanBeFoundInParent": "main/subdir",
+    "testFailedImportCannotBeFoundInParent": "main/subdir",
+    "testImportInParentButNoInit": "main/subdir",
+    "testFailedImportCanBeFoundInParentsParent": "main/subdir1/subdir2",
+}
```
JukkaL, line 35: I'd prefer having a way of specifying this in the test case, as this is more fragile (consider what happens if one of the test cases is renamed). For example, this could be specified via `# cwd: main/subdir` in the test case description, similar to how we have `# cmd: ...` and similar.  This would also make it easier to understand what the test cases are testing exactly.

This is not absolutely essential, but it would be nice to have. If you decide not to implement this, please add a comment to each test case that describes the working directory and mentions `custom_working_dir`.

test-data/unit/cmdline.test
```diff
@@ -476,6 +476,50 @@ ignore_missing_imports = True
...
+
+[case testImportInParentButNoInit]
+# cmd: mypy main.py
+[file main/subdir/main.py]
+import wrongdir
+[file main/wrongdir.py]
+[file main/__init__.py]
+[out]
+main.py:1: error: Cannot find implementation or library stub for module named 'wrongdir'
+main.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```
JukkaL, line 521: Command line tests are quite expensive to run. Could you reduce the number of test cases to 2, for example? Maybe you can combine tested features.

mypy/test/testcmdline.py
ethan-leba: Added `# cwd: ...` to `testcmdline.py` 👍

ethan-leba:
Thanks for the review @JukkaL! I addressed your comments, and used @TH3CHARLie's suggestion for the error message. Let me know if there's any more work that needs to be done on this PR!

PR #8931, opened 2020-06-02
Update README.md

Mukulbaid63:


JukkaL:
I'd rather not add colons to headings.

PR #8932, opened 2020-06-02
Add docs for no_site_packages config option

ilevkivskyi:


docs/source/config_file.rst
```diff
@@ -230,10 +230,15 @@ section of the command line docs.
 
     This option may only be set in the global section (``[mypy]``).
 
+``no_site_packages`` (bool, default False)
+    Disables searching for :pep:`561` compliant packages. This will also
```
JukkaL, line 234: Regular users won't recognize most PEP numbers, so it would be better to give a short summary of what this means here, as otherwise most readers would have to open the link to understand what is being talked about. Maybe something like "Disables using type information in installed packages (PEP 561)".

docs/source/config_file.rst
```diff
@@ -230,10 +230,15 @@ section of the command line docs.
...
+    disable searching for a usable Python executable. This acts the same as
+    :ref:`--no-site-packages <no-site-packages>` command line flag.
+
 ``no_silence_site_packages`` (bool, default False)
-    Enables reporting error messages generated within :pep:`561` compliant packages.
-    Those error messages are suppressed by default, since you are usually
-    not able to control errors in 3rd party code.
+    Enables reporting error messages generated within :pep:`561` compliant
```
JukkaL, line 239: Optionally also add a summary of what PEP 561 compliant packages are here.

PR #8933, opened 2020-06-02
Support unary op via CallC

TH3CHARLie:
related mypyc/mypyc#709, mypyc/mypyc#734

Support unary ops and provide `int_neg_op` as an example.

### DIFF
--- a/mypyc/primitives/int_ops.py
+++ b/mypyc/primitives/int_ops.py
@@ -6,14 +6,14 @@ representation (CPyTagged).
 See also the documentation for mypyc.rtypes.int_rprimitive.
 """
 
-from mypyc.ir.ops import OpDescription, ERR_NEVER, ERR_MAGIC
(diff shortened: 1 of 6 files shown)


mypyc/primitives/int_ops.py
```diff
@@ -131,13 +131,12 @@ def int_compare_op(op: str, c_func_name: str) -> None:
     emit=simple_emit('{dest} = {args[0]} + {args[1]};'))
 
 
-def int_unary_op(op: str, c_func_name: str) -> OpDescription:
```
TH3CHARLie, line 134: Some thoughts: I think the original design somehow reflects or predicts what we are doing now

JukkaL: Yeah, the new approach fits in very nicely here.

TH3CHARLie:
Thanks for the review!

PR #8936, opened 2020-06-03
Report note about binary operation on the same line as error

JukkaL:
This way we also suppress the note when ignoring the error.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2680,12 +2680,16 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
             if msg.is_errors():
                 self.msg.add_errors(msg)
+                # Point any notes to the same location as an existing message.
+                recent_context = msg.most_recent_context()
                 if len(left_variants) >= 2 and len(right_variants) >= 2:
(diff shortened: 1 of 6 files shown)


mypy/checkexpr.py
```diff
@@ -2683,7 +2683,7 @@ def check_op(self, method: str, base_type: Type,
                 if len(left_variants) >= 2 and len(right_variants) >= 2:
                     self.msg.warn_both_operands_are_from_unions(context)
                 elif len(left_variants) >= 2:
-                    self.msg.warn_operand_was_from_union("Left", base_type, context=right_expr)
+                    self.msg.warn_operand_was_from_union("Left", base_type, context=context)
                 elif len(right_variants) >= 2:
                     self.msg.warn_operand_was_from_union("Right", right_type, context=right_expr)
```
JelleZijlstra, line 2688: Should we make the same change here?

JukkaL: I updated the PR to do this more consistently.

JukkaL:
@ilevkivskyi I updated the logic for finding the note location since a bunch of tests were failing.

PR #8939, opened 2020-06-03
Fixed stubgen parsing generics from C extensions

AWhetter:
pybind11 is capable of producing type signatures that use generics (for example https://github.com/pybind/pybind11/blob/4e3d9fea74ed50a042d98f68fa35a3133482289b/include/pybind11/stl.h#L140). A user may also opt to write a signature in the docstring that uses generics.
Currently when stubgen parses one of these generics, it attempts to import a part of it. For example if a docstring had `my_func(str, int) -> List[mypackage.module_being_parsed.MyClass]`, the resulting stub file tries to import `List[mypackage.module_being_parsed`.
This change fixes this behaviour by breaking the found type down into the multiple types around `[], ` characters, adding any imports from those types that are needed, and then stripping out the name of the module being parsed.

### DIFF
--- a/mypy/stubgenc.py
+++ b/mypy/stubgenc.py
@@ -201,7 +201,16 @@ def strip_or_import(typ: str, module: ModuleType, imports: List[str]) -> str:
         imports: list of import statements (may be modified during the call)
     """
     stripped_type = typ
-    if module and typ.startswith(module.__name__ + '.'):
+    if any(c in typ for c in '[,'):
+        for subtyp in re.split(r'[\[,\]]', typ):
+            strip_or_import(subtyp.strip(), module, imports)
+        if module:
+            stripped_type = re.sub(
+                r'(^|[\[, ]+)' + re.escape(module.__name__ + '.'),
+                r'\1',
+                typ,
+            )
+    elif module and typ.startswith(module.__name__ + '.'):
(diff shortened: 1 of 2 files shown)


mypy/stubgenc.py
```diff
@@ -201,7 +201,16 @@ def strip_or_import(typ: str, module: ModuleType, imports: List[str]) -> str:
...
-    if module and typ.startswith(module.__name__ + '.'):
+    if any(c in typ for c in '[,'):
+        for subtyp in re.split(r'\[|,\]', typ):
+            strip_or_import(subtyp.strip(), module, imports)
+        if module:
+            stripped_type = re.sub(
+                r'(^|[\[, ]+)' + re.escape(module.__name__ + '.'),
+                r'\1',
+                typ,
+            )
```
JukkaL, line 212: It would be nice to have some unit tests for the new functionality, since this logic is non-trivial. You can add them to `mypy/test/teststubgen.py`.

(Maybe also modify this so that the module name is passed instead of a module object.)

AWhetter: Ah I couldn't find the tests for stubgen. I'll get those added. Thanks for the review!

AWhetter:
Is there anything else that I can do to help progress this pull request?

PR #8945, opened 2020-06-03
Change some sequence components in types from lists to tuples

msullivan:
This cuts down on memory a bit by avoiding teh separate allocation
inside lists and by allowing all empty sequences to share the empty
tuple.


JelleZijlstra:
I restarted the Travis build and now it's green.

PR #8948, opened 2020-06-04
[mypyc] Support var arg in CallC, replace new_dict_op

TH3CHARLie:
related mypyc/mypyc#709 mypyc/mypyc#734

Summary:
- introduce variable arguments in `CallC`
- replace old `new_dict_op`(which relies on a specialized emit callback) with two `CallC` ops: `dict_new_op` and `dict_build_op`, which handles create an empty dict and a dict from k-v pairs.
- fix related IR tests, especially separate the `testDel` case into three subcases which test list, dict and attributes respectively.


TH3CHARLie:
The current codegen for `CallC`
https://github.com/python/mypy/blob/348a9d4774b229b96c30a27ed2ca93992da5bd4f/mypyc/codegen/emitfunc.py#L418-L421

Seeing this, I start to think, does the newly introduced `var_arg_idx` really necessary? Since we join the args anyway(and there's no difference for fixed/var args), if we handle all the semantics properly before make a `CallC`(mainly via `xxx.call_c`), then maybe the `var_arg_idx` is unnecessary.

mypyc/ir/ops.py
```diff
@@ -1151,13 +1151,15 @@ def __init__(self,
...
-                 line: int) -> None:
+                 line: int,
+                 var_arg_idx: int = -1) -> None:
         self.error_kind = error_kind
         super().__init__(line)
         self.function_name = function_name
         self.args = args
         self.type = ret_type
         self.steals = steals
+        self.var_arg_idx = var_arg_idx
```
JukkaL, line 1162: Document this attribute.

mypyc/irbuild/ll_builder.py
```diff
@@ -592,7 +601,12 @@ def make_dict(self, key_value_pairs: Sequence[DictEntry], line: int) -> Value:
                 )
 
         if result is None:
-            result = self.primitive_op(new_dict_op, initial_items, line)
+            items_size = len(initial_items)
+            if items_size > 0:
+                size = self.add(LoadInt(items_size // 2, -1, c_int_rprimitive))
+                result = self.call_c(dict_build_op, [size] + initial_items, line)
+            else:
+                result = self.call_c(dict_new_op, [], line)
```
JukkaL, line 609: This same code is repeated above around line 589. Maybe extract these lines into a helper method and share the code?

mypyc/primitives/dict_ops.py
```diff
@@ -108,6 +108,19 @@ def emit_new_dict(emitter: EmitterInterface, args: List[str], dest: str) -> None
     error_kind=ERR_MAGIC,
     emit=emit_new_dict)
 
+dict_new_op = c_custom_op(
```
JukkaL, line 111: For consistency with other ops, add a single-line comment (e.g. "Create an empty dictionary").

Can we remove the old op, or is it still needed for something?

mypyc/primitives/dict_ops.py
```diff
@@ -108,6 +108,19 @@ def emit_new_dict(emitter: EmitterInterface, args: List[str], dest: str) -> None
...
+    arg_types=[],
+    return_type=dict_rprimitive,
+    c_function_name='PyDict_New',
+    error_kind=ERR_MAGIC)
+
+dict_build_op = c_custom_op(
```
JukkaL, line 117: Add a single-line comment that documents what this does and explains the arguments.

mypyc/primitives/registry.py
```diff
@@ -337,10 +338,11 @@ def c_method_op(name: str,
                 return_type: RType,
                 c_function_name: str,
                 error_kind: int,
+                var_arg_type: Optional[RType] = None,
                 steals: StealsDescription = False,
                 priority: int = 1) -> CFunctionDescription:
     ops = c_method_call_ops.setdefault(name, [])
```
JukkaL, line 344: Add a docstring and document at least `var_arg_type`.

mypyc/primitives/registry.py
```diff
@@ -351,10 +353,11 @@ def c_function_op(name: str,
                   return_type: RType,
                   c_function_name: str,
                   error_kind: int,
+                  var_arg_type: Optional[RType] = None,
                   steals: StealsDescription = False,
                   priority: int = 1) -> CFunctionDescription:
     ops = c_function_ops.setdefault(name, [])
```
JukkaL, line 359: Add a docstring here as well.

mypyc/primitives/registry.py
```diff
@@ -365,15 +368,26 @@ def c_binary_op(name: str,
...
     return desc
 
 
+def c_custom_op(arg_types: List[RType],
+                return_type: RType,
+                c_function_name: str,
+                error_kind: int,
+                var_arg_type: Optional[RType] = None,
+                steals: StealsDescription = False) -> CFunctionDescription:
+    return CFunctionDescription('<custom>', arg_types, return_type, var_arg_type,
```
JukkaL, line 387: Add a docstring, similar to above. You can use the docstring of `custom_op` as a starting point and tweak it (and similarly for the other docstrings).

mypyc/primitives/str_ops.py
```diff
@@ -73,6 +73,7 @@
             return_type=str_rprimitive,
             c_function_name='CPyStr_Append',
             error_kind=ERR_MAGIC,
+            var_arg_type=None,
```
JukkaL, line 76: Minor nit: I think that this argument can be left out and we can rely on the default; (almost) all binary ops will likely use positional arguments.

mypyc/test-data/irbuild-statements.test
```diff
@@ -765,18 +761,32 @@ L0:
...
+    r14 = unicode_1 :: static  ('one')
+    r15 = unicode_4 :: static  ('four')
     r16 = d.__delitem__(r14) :: object
-    r17 = None
-    return r17
+    r17 = d.__delitem__(r15) :: object
+    r18 = None
+    return r18
+
+[case testDelAttribute]
```
JukkaL, line 778: Thanks for splitting this test case! It's much better like this.

mypyc/primitives/dict_ops.py
TH3CHARLie: The old op is actually a miss, it should be removed.

mypyc/primitives/str_ops.py
TH3CHARLie: This is also a miss. It shouldn't be explicitly set

PR #8949, opened 2020-06-04
include mypyc subdirectories in the sdist

mr-c:
Needed for the Debian package of `mypy` as we use `mypyc` during building.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,6 +5,15 @@ recursive-include docs *
 recursive-include mypy/typeshed *.py *.pyi
 recursive-include mypy/xml *.xsd *.xslt *.css
 recursive-include mypyc/lib-rt *.c *.h *.tmpl
+recursive-include mypyc/ir *.py
+recursive-include mypyc/codegen *.py
+recursive-include mypyc/irbuild *.py
+recursive-include mypyc/primitives *.py
+recursive-include mypyc/transform *.py
+recursive-include mypyc/test *.py
+recursive-include mypyc/test-data *.test
+recursive-include mypyc/test-data/fixtures *.py *.pyi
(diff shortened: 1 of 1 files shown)


mr-c:
Originally I was going to patch in the missing files, when I thought it was just one directory. Now it is a lot of directories, so with my Debian hat on I would like to request a 0.781 release with this fix.

Thanks, as always, for such a great project!

mr-c:
I can contribute a test for this error, if that would be welcome.

msullivan:
A test with that error would be fantastic, because we mess this up a lot >_>

@ilevkivskyi what do you think about a point release?

ilevkivskyi:
Yes, I will release 0.781 tomorrow. What exactly should I cherry-pick? Just this PR?

mr-c:
FYI, one cannot `pip install mypy` in Python 3.9 and have `mypyc` working as there is no binary wheel for mypy 0.780 for Python 3.9 yet

msullivan:
> FYI, one cannot `pip install mypy` in Python 3.9 and have `mypyc` working as there is no binary wheel for mypy 0.780 for Python 3.9 yet

There's still a `none` wheel, though, right?

mr-c:
> There's still a `none` wheel, though, right?

The `none` wheel for 0.780 is missing the same sub-directories under `mypyc/` as the sdist is :pleading_face:

mr-c:
@ilevkivskyi 0.781 should also include https://github.com/python/mypy/pull/8963/commits/4d9a07dd2c0fba3eb6a53dc5ca7c9f9c618b3aa1

ilevkivskyi:
I already released 0.781 without this one. We are however thinking about 0.782 (for couple other fixes), so we will include it there.

PR #8955, opened 2020-06-05
[mypyc] Introduce low level integer type

TH3CHARLie:
closes mypyc/mypyc#735

This PR introduces a `c_int_rprimitive` RType which represents a low level, plan integer(corresponds to C's `int`). It also allows `LoadInt` to select its `rtype` to generate tagged/plain integer code accordingly.

### DIFF
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -174,7 +174,7 @@ class RPrimitive(RType):
         self.is_unboxed = is_unboxed
         self._ctype = ctype
         self.is_refcounted = is_refcounted
-        if ctype == 'CPyTagged':
+        if ctype in {'CPyTagged', 'int'}:
             self.c_undefined = 'CPY_INT_TAG'
         elif ctype == 'PyObject *':
             # Boxed types use the null pointer as the error value.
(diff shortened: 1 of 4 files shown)


mypyc/ir/rtypes.py
```diff
@@ -174,7 +174,7 @@ def __init__(self,
         self.is_unboxed = is_unboxed
         self._ctype = ctype
         self.is_refcounted = is_refcounted
-        if ctype == 'CPyTagged':
+        if ctype in {'CPyTagged', 'int'}:
```
JukkaL, line 177: Minor performance suggestion: `x in (y, z)` is generally faster than `x in {y, z}` (at least when there are only a few items in the set).

mypyc/ir/rtypes.py
```diff
@@ -234,6 +234,10 @@ def __repr__(self) -> str:
 short_int_rprimitive = RPrimitive('short_int', is_unboxed=True, is_refcounted=False,
                                   ctype='CPyTagged')  # type: Final
 
+# low level integer (corresponds to C's 'int').
+c_int_rprimitive = RPrimitive('c_int', is_unboxed=True, is_refcounted=False,
+                              ctype='int')  # type: Final
```
JukkaL, line 239: The ctype should be `Py_ssize_t`, since we want the integer to have the same size as a pointer.

PR #8960, opened 2020-06-07
Better error message for incompatible unpacked **kwargs

### LINKED ISSUE #8874: Generate better error message for incompatible **kwargs
This output is from #8772:

```
$ mypy test.py 
test.py:6: error: Argument 2 to "dumps" has incompatible type "**Dict[str, int]"; expected "bool"
test.py:6: error: Argument 2 to "dumps" has incompatible type "**Dict[str, int]"; expected "Optional[Type[JSONEncoder]]"
test.py:6: error: Argument 2 to "dumps" has incompatible type "**Dict[str, int]"; expected "Optional[Tuple[str, str]]"
test.py:6: error: Argument 2 to "dumps" has incompatible type "**Dict[str, int]"; expected "Optional[Callable[[Any], Any]]"
Found 4 errors in 1 file (checked 1 source file)
```

The error messages are pretty confusing. We should special case this and produce a customized error message that explains the situation better.

Here's the code that triggers the errors:

```py
import json

json_kwargs = dict(indent=2)

json.dumps({}, **json_kwargs)
```

Some ideas about what to do:
* Generate a single error message instead of multiple ones.
* Add a note suggesting the use of `Dict[str, Any]` or a TypedDict type for the `**kwargs` argument.
* Maybe also explain that mypy matches the value type against all arguments, or add a link to documentation (add something to common issues, for example).

rohitkg98:
Closes #8874 
Added a note for unpacked **kwargs.
I believe there should be multiple error messages, to inform of all callable argument type mismatch.
Hence, I have added just a note.
Also, added in Common Issues.


docs/source/common_issues.rst
```diff
@@ -808,3 +808,34 @@ not necessary:
...
         def test(self, t: List[int]) -> Sequence[str]:  # type: ignore[override]
             ...
+
+Unpacking Dictionary in keyword arguments
+-----------------------------------
+If you are unpacking a dictionary into a Callable's arguments, then
+you need to ensure the types of dictionary members by providing either
+a TypedDict or a variable with type Dict[str, Any].
+
+For example, the following function calls are valid:
```
mental32, line 818: Just a suggestion - maybe an example that shows some offending cases or invalid function calls would help drive the point home? (otherwise good job!)

rohitkg98: Sure, will add them.

mypy/messages.py
```diff
@@ -528,6 +528,7 @@ def incompatible_argument(self,
                 arg_type_str = '*' + arg_type_str
             elif arg_kind == ARG_STAR2:
                 arg_type_str = '**' + arg_type_str
+                notes.append("Use TypedDict or Dict[str, Any] for unpacked **kwargs")
```
ChetanKhanna, line 531: Hey! You will also have to modify the tests affected by this change :) Cheers!

rohitkg98: yea, its a work in progress

rohitkg98:
Any updates on this? Updated PR with respect to master

gvanrossum:
I'm not sure adding this note is actually helpful. The root problem seems to be that the current error is just confusing -- it says (for example) _argument has incompatible type "**Dict[str, float]"; expected "int"_ where it should explain that the dict's value type is not compatible with all of the function's argument type, or something like that. It is easy to come up with examples where the note is just wrong, e.g.
```
def f(a: int): ...
f(**dict(a=3.14))
```

rohitkg98:
The note is needed in cases where the Dict is untyped and inferred type becomes `**Dict[str, object]`. In such cases, the dict needs to explicitly typed. e.g.

```
def f(a: int, b: str): ...
f(**dict(a=1, b="some_str"))
```
Even though types are correct, it fails.

In cases where inferred dict type is not `**Dict[str, object]`, we can modify the error message to explain the dict's value type is not compatible with functions arguments. Would that work?

gvanrossum:
> The note is needed in cases where the Dict is untyped and inferred type becomes `**Dict[str, object]`. In such cases, the dict needs to explicitly typed. e.g.
> 
> ```
> def f(a: int, b: str): ...
> f(**dict(a=1, b="some_str"))
> ```
> 
> Even though types are correct, it fails.

In this case we should do what Jukka suggested in his first bullet in #8874: "Generate a single error message instead of multiple ones."

> In cases where inferred dict type is not `**Dict[str, object]`, we can modify the error message to explain the dict's value type is not compatible with functions arguments. Would that work?

I'm thinking there are several cases.

- The key type is not compatible with `str` -- that should be a separate error message
- The value type is not compatible with one or more argument types -- that should be another error message

In the example you give, maybe the error should be
```
error: first **kwargs argument to "f" has value type "object", incompatible with formal parameter 1 ("int") and 2 ("str")
```
(However, I don't know if all the information needed to format it that way is actually available.)

hauntsaninja (changes requested):
Requesting changes as per gvanrossum's comment

97littleleaf11:
Hi, any progress on this PR?

JukkaL:
Closing due to inactivity.

PR #8963, opened 2020-06-07
add release tester

mr-c:
As requested in https://github.com/python/mypy/pull/8949#issuecomment-640097714

Adapted from https://github.com/common-workflow-language/cwltool/blob/814bd0405a7701efc7d63e8f0179df394c7766f7/release-test.sh which was based off of https://github.com/dib-lab/khmer/blob/6fde87ef565cece87306ea14971fac3c23954e77/doc/dev/release.rst

If there is a manual release process then I suggest adding `misc/release-test.sh` to it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,6 +80,10 @@ jobs:
   #   env:
   #   - TOXENV=dev
   #   - EXTRA_ARGS=
+  - name: "check MANIFEST.in completeness"
+    python: 3.7
+    install: ""
+    script: RELEASE_SKIP=head ${TRAVIS_BUILD_DIR}/misc/release-test.sh
 
 install:
 - pip install -U pip setuptools
--- a/misc/release-test.sh
+++ b/misc/release-test.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+
+set -e
+set -x
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." >/dev/null && pwd )"
+
+export LC_ALL=C
+export TEST_MYPYC=1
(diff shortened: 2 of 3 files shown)


msullivan:
Could you look into the CI failure?

mr-c:
@msullivan Fixing it now, thanks for the reminder!

msullivan:
Still broken

mr-c:
@msullivan Commit https://github.com/python/mypy/pull/8963/commits/52d84f6974dbc40b56914ad672bba1a718e1d876 works, now rebasing one last time.

ilevkivskyi:
We had a short offline discussion about this with @JukkaL: it looks like this is too slow to run in every CI build, to speed up 0.782 release, I include the MANIFEST changes as a separate PR https://github.com/python/mypy/pull/9033. In the meantime we can discuss what is the best way to run this test.

mr-c:
Thanks for looking at this. It is a beast of a gauntlet, but it did find many MANIFEST issues.

Happy to modify this further.A travis mode could skip all the test runs besides the last one.

mr-c:
Can we get a version of this merged? It would have detected https://github.com/python/mypy/pull/9587 before the release ..

hauntsaninja:
A couple months ago we made a lot of MANIFEST improvements, including running check manifest in our release builds (https://github.com/mypyc/mypy_mypyc-wheels/pull/14). The last couple releases seem to have gone well, ending a streak where we messed our sdists up pretty regularly. Given this, I'm going to close, but thank you for helping improve and test mypy releases! :-)

PR #8969, opened 2020-06-08
Add union narrowing using in/not in

### LINKED ISSUE #8940: Narrowing unions with `in`/`__contains__`
It seems like you should be able to narrow a union of literals to a subset, or a union of unions of literals to just one of the unions, by checking if a value with such a union is `in` a container of one of the subset/sub unions (see the code example for something hopefully more coherent).

This is really useful for, say, a deserialized JSON object that has different subsets of keys that map to differently-shaped objects, and you want to pass them to functions that take the appropriate parameters.

Really could be either - this is behavior I'm surprised by, but it could be not implemented rather than working incorrectly

* Minimal repro showing the issue:

Playground gist link: https://mypy-play.net/?mypy=latest&python=3.8&gist=18262b0bbd541a22d4f48c5bb628f00e
Code:
```python
from typing import Union, Dict, List, Literal

firstset = Union[Literal['a'], Literal['b'], Literal['c']]
secondset = Union[Literal['d'], Literal['e'], Literal['f']]

firstdict: Dict[firstset, int] = {'a': 2}
seconddict: Dict[secondset, float] = {'d': 3.0}

firstlist: List[firstset] = ['a', 'b', 'c']
secondlist: List[secondset] = ['d', 'e', 'f']

def printit(key: Union[Literal['a'], Literal['b'], Literal['c'],
                       Literal['d'], Literal['e'], Literal['f']]):
    if key in firstdict:
        reveal_type(key) # Union[Literal['a'], Literal['b'], Literal['c'], Literal['d'], Literal['e'], Literal['f']]'
    if key in seconddict:
        reveal_type(key) # Union[Literal['a'], Literal['b'], Literal['c'], Literal['d'], Literal['e'], Literal['f']]'

    if key in firstlist:
        reveal_type(key) # Union[Literal['a'], Literal['b'], Literal['c'], Literal['d'], Literal['e'], Literal['f']]'
    if key in secondlist:
        reveal_type(key) # Union[Literal['a'], Literal['b'], Literal['c'], Literal['d'], Literal['e'], Literal['f']]'

    if key == 'a' or key == 'b' or key == 'c':  # this one works
        reveal_type(key)  # 'Union[Literal['a'], Literal['b'], Literal['c']]' 
```

```
test.py:16: note: Revealed type is 'Union[Literal['a'], Literal['b'], Literal['c'], Literal['d'], Literal['e'], Literal['f']]'
test.py:18: note: Revealed type is 'Union[Literal['a'], Literal['b'], Literal['c'], Literal['d'], Literal['e'], Literal['f']]'
test.py:21: note: Revealed type is 'Union[Literal['a'], Literal['b'], Literal['c'], Literal['d'], Literal['e'], Literal['f']]'
test.py:23: note: Revealed type is 'Union[Literal['a'], Literal['b'], Literal['c'], Literal['d'], Literal['e'], Literal['f']]'
test.py:26: note: Revealed type is 'Union[Literal['a'], Literal['b'], Literal['c']]'
```
```
test.py:16: note: Revealed type is 'Union[Literal['a'], Literal['b'], Literal['c']]
test.py:18: note: Revealed type is 'Union[Literal['d'], Literal['e'], Literal['f']]
test.py:21: note: Revealed type is 'Union[Literal['a'], Literal['b'], Literal['c']]
test.py:23: note: Revealed type is 'Union[Literal['d'], Literal['e'], Literal['f']]
test.py:26: note: Revealed type is 'Union[Literal['a'], Literal['b'], Literal['c']]'
```
Happens on 3.7 and 3.8, mypy 0.770 and whatever playground has as "latest"
Defaults

(FYI if you were reading this shortly after I posted it and were annoyed by me editing it a lot - I'm really sorry, I accidentally posted it early and needed to actually fill it out)

sfoster1:
mypy can already narrow unions in this case:

def b(a: Union[int, str]):
    if a == 2:
         reveal_type(a)  # int

and it can narrow optionals in this case:

def c(d: Optional[int], e: List[int]):
    if d in e:
        reveal_type(d)  # int

This pr allows it to narrow unions in that case:

def f(g: Union[int, str, bool], h: List[Union[int, str]]):
    if g in h:
        reveal_type(g)  # Union[int, str]
    else:
        reveal_type(g)  # bool

This is useful for (in my case) picking structures out of a tagged
union.

In terms of what still needs to be done
- This is my first time contributing to mypy so I'm sure I've missed
some edge cases
- I'm sure I haven't added quite enough tests - some pointers in that
direction would be quite welcome
- I would really like this to also work if the collection items are
scalar types like List[int] - I'm not sure if that's necessary for this
PR though

And any and all comments on the details or broad structure of the PR are
welcome of course.

Closes #8940

### DIFF
--- a/test-data/unit/check-unions.test
+++ b/test-data/unit/check-unions.test
@@ -1048,3 +1048,17 @@ def foo(a: T2, b: T2) -> T2:
 def bar(a: T4, b: T4) -> T4:  # test multi-level alias
     return a + b
 [builtins fixtures/ops.pyi]
+
+
+[case testNarrowingUnionFromInContainer]
+from typing import Union, NewType, Set
+MyCoolInt = NewType('MyCoolInt', int)
+T1 = Union[int, float, MyCoolInt, str]
+
+my_cool_set: Set[Union[int, str]] = set(2, 'hi')
+
+def my_cool_function(my_cool_arg: T1):
+    if my_cool_arg in my_cool_set:
+        reveal_type(my_cool_arg)  # N: Revealed type is Union[int, str]
+    else:
+        reveal_type(my_cool_arg)  # N: Revealed type is Union[float, MyCoolint]
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -4076,18 +4076,37 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                     item_type = operand_types[left_index]
                     collection_type = operand_types[right_index]
 
-                    # We only try and narrow away 'None' for now
-                    if not is_optional(item_type):
+                    proper_item = get_proper_type(item_type)
+                    collection_item_type = get_proper_type(builtin_item_type(collection_type))
+                    # nothing to narrow
+                    if not isinstance(proper_item, UnionType):
                         continue
 
-                    collection_item_type = get_proper_type(builtin_item_type(collection_type))
-                    if collection_item_type is None or is_optional(collection_item_type):
+                    if collection_item_type is None:
+                        continue
+                    if is_optional(collection_item_type) and is_optional(item_type):
                         continue
                     if (isinstance(collection_item_type, Instance)
                             and collection_item_type.type.fullname == 'builtins.object'):
                         continue
-                    if is_overlapping_erased_types(item_type, collection_item_type):
-                        if_map, else_map = {operands[left_index]: remove_optional(item_type)}, {}
+
+                    # optionals need type erasure
+                    if is_optional(item_type):
+                        if is_overlapping_erased_types(item_type, collection_item_type):
+                            if_map = {operands[left_index]: remove_optional(item_type)}
+                            else_map = {}
+                        else:
+                            continue
+                    elif isinstance(collection_item_type, UnionType):
+                        # generic unions produce their intersection
+                        items_set = frozenset(proper_item.items)
+                        collection_set = frozenset(collection_item_type.items)
+                        intersect = items_set & collection_set
+                        if not intersect:
+                            continue
+                        if_map = {operands[left_index]: UnionType.make_union(list(intersect))}
+                        else_map = {operands[left_index]: UnionType.make_union(
+                            list(items_set - collection_set))}
                     else:
                         continue
                 else:


test-data/unit/check-unions.test
```diff
@@ -1048,3 +1048,17 @@ def foo(a: T2, b: T2) -> T2:
...
+MyCoolInt = NewType('MyCoolInt', int)
+T1 = Union[int, float, MyCoolInt, str]
+
+my_cool_set: Set[Union[int, str]] = set(2, 'hi')
+
+def my_cool_function(my_cool_arg: T1):
+    if my_cool_arg in my_cool_set:
+        reveal_type(my_cool_arg)  # N: Revealed type is Union[int, str]
+    else:
+        reveal_type(my_cool_arg)  # N: Revealed type is Union[float, MyCoolint]
```
JelleZijlstra, line 1064: This is incorrect; `my_cool_arg` could be an int or str that happens not to be a member of `my_cool_set`.

test-data/unit/check-unions.test
```diff
@@ -1048,3 +1048,17 @@ def foo(a: T2, b: T2) -> T2:
...
+[case testNarrowingUnionFromInContainer]
+from typing import Union, NewType, Set
+MyCoolInt = NewType('MyCoolInt', int)
+T1 = Union[int, float, MyCoolInt, str]
+
+my_cool_set: Set[Union[int, str]] = set(2, 'hi')
+
+def my_cool_function(my_cool_arg: T1):
+    if my_cool_arg in my_cool_set:
+        reveal_type(my_cool_arg)  # N: Revealed type is Union[int, str]
```
JelleZijlstra, line 1062: This is not quite safe; for example, `2.0 in {2}` returns True, so it could still be a float.

test-data/unit/check-unions.test
sfoster1: Ah totally yeah. So only the positive results are safe to do this, that's too bad.

test-data/unit/check-unions.test
sfoster1: Yeah and now that I think about it, it wouldn't necessarily be safe if the class's `__contains__` or `__eq__` was doing something special. Is that something mypy tries to help with?

sfoster1: Oh I think I see, this should borrow some of the same precondition logic that the `==`/`is` checks use like ten lines above it

sfoster1:
As I suspected this ends up being more complex than I thought! Who could see it coming (not me, apparently).

It seems like the check should use pretty much the same rules as the `==` checks, does that seem reasonable?

JelleZijlstra:
> It seems like the check should use pretty much the same rules as the `==` checks, does that seem reasonable?

That makes sense since `in` is normally implemented using `__eq__`.

sfoster1:
As you can tell by me not touching it for a week, I think doing this right is going to take a bit more time than I have, so I'm going to close this - if I come up with something better, I'll reopen. Sorry to waste your time!

PR #8973, opened 2020-06-09
[mypyc] Add LoadGlobal IR

TH3CHARLie:
In this PR we introduce a `LoadGlobal` IR that loads a name in global space.

It's part of mypyc/mypyc#736 and mypyc/mypyc#737, it replaces literal `LoadStatic` with `LoadGlobal` and moves the name generation logic to caller in irbuild. It will eventually replace `LoadStatic`.

`LoadGlobal` simply reads a global **value** via a given name(its counterpart for reading an **address** will be a future `LoadAddress`)

This PR also introduces a registry and a description for making `LoadGlobal` worked with more customized loading ops(some `name_ref_ops` like the True and False)


TH3CHARLie:
> Should this just be merged with LoadStatic?

Yes, this comes to my mind when I was implementing this. One problem is how to properly handle the current registry-driven ops. Jukka and I will have a video meeting today discussing this(and I expect this prototype design will change a lot once we had the discussion.

runtests.py
```diff
@@ -65,7 +65,7 @@
 # time to run.
 cmds = {
     # Self type check
-    'self': python_name + ' -m mypy --config-file mypy_self_check.ini -p mypy',
+    'self': python_name + ' -m mypy --config-file mypy_self_check.ini -p mypyc',
```
TH3CHARLie, line 68: remove this file from commit

mypyc/irbuild/ll_builder.py
```diff
@@ -432,34 +433,34 @@ def literal_static_name(self, value: Union[int, float, complex, str, bytes]) ->
     def load_static_int(self, value: int) -> Value:
         """Loads a static integer Python 'int' object into a register."""
         if abs(value) > MAX_LITERAL_SHORT_INT:
-            static_symbol = self.literal_static_name(value)
-            return self.add(LoadStatic(int_rprimitive, static_symbol, ann=value))
+            identifier = STATIC_PREFIX + self.literal_static_name(value)
```
JukkaL, line 436: Can you move the `STATIC_PREFIX + ...` calculation to `literal_static_name` so that it doesn't need to be duplicated several times?

mypyc/primitives/registry.py
```diff
@@ -52,6 +52,14 @@
                               ('steals', StealsDescription),
                               ('priority', int)])
 
+# A description for C load operations including LoadGlobal and LoadAddress
+CLoadDescription = NamedTuple(
+    'CLoadDescription',     [('name', str),
+                            ('return_type', RType),
+                            ('identifier', str),
+                            ('cast_str', str),
+                            ('load_address', bool)])
```
JukkaL, line 61: Document the attributes.

PR #8998, opened 2020-06-12
Fix exception causes all over the codebase

cool-RR:
I recently went over [Matplotlib](https://github.com/matplotlib/matplotlib/pull/16706), [Pandas](https://github.com/pandas-dev/pandas/pull/32322) and [NumPy](https://github.com/numpy/numpy/pull/15731), fixing a small mistake in the way that Python 3's exception chaining is used. If you're interested, I can do it here too. I've done it on just one file right now. 

The mistake is this: In some parts of the code, an exception is being caught and replaced with a more user-friendly error. In these cases the syntax `raise new_error from old_error` needs to be used.

Python 3's exception chaining means it shows not only the traceback of the current exception, but that of the original exception (and possibly more.) This is regardless of `raise from`. The usage of `raise from` tells Python to put a more accurate message between the tracebacks. Instead of this: 

    During handling of the above exception, another exception occurred:

You'll get this: 

    The above exception was the direct cause of the following exception:

The first is inaccurate, because it signifies a bug in the exception-handling code itself, which is a separate situation than wrapping an exception.

Let me know what you think!

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1988,13 +1988,13 @@ class State:
                     raise CompileError([
                         "mypy: can't read file '{}': {}".format(
                             self.path, os.strerror(ioerr.errno))],
-                        module_with_blocker=self.id)
+                        module_with_blocker=self.id) from ioerr
                 except (UnicodeDecodeError, DecodeError) as decodeerr:
                     if self.path.endswith('.pyd'):
                         err = "mypy: stubgen does not support .pyd files: '{}'".format(self.path)
                     else:
                         err = "mypy: can't decode file '{}': {}".format(self.path, str(decodeerr))
-                    raise CompileError([err], module_with_blocker=self.id)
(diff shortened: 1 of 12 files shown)


cool-RR:
Yup, done.

cool-RR:
LOL, I did this exercise to a few dozen projects by now, and you're the first one to ask me that.

Yes, I have a script. It's too ugly and buggy to share, but basically it takes a GitHub repo and outputs a list of modules and lines in which this mistake occurs. The next step would be a script that automatically opens PRs :)

I wrote a more elegant version of it [as a PyLint rule here](https://github.com/PyCQA/pylint/pull/3695), hopefully it gets accepted soon. It's likely to make a more long-term impact than me going through repos one-by-one.

cool-RR:
I just made a blog post about this change here: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with

PR #8999, opened 2020-06-12
Improve stubgenc property type detection

### LINKED ISSUE #8995: Improve stubgenc type detection on PyBind11 properties
Hello,

Thank you for the work done here on mypy. I don't know much about it yet but I believe the following would improve stub files generation for [Pybind11](https://pybind11.readthedocs.io/en/stable/) users.

## Current behavior
PyBind11 has the ability to bind C++ classes to Python objects.
While `stubgen` properly detects types for most functions/methods from the docstrings, types in properties (get, set) created from PyBind11, are reported as `Any`.

For instance, the following C++ code defines a class `Dog` and bind it to Python in a module named `python_example`.
```cpp
#include <pybind11/pybind11.h>

struct Dog {
  int age = 0;
  int n_friends = 0;
};

PYBIND11_MODULE(python_example, m) {
  pybind11::class_<Dog>(m, "Dog")
    .def(pybind11::init<>())  // Bind an __init__ method
    .def("get_age", [](Dog const& self){return self.age; })  // Bind a get_age method
    .def_readonly("age", &Dog::age)  // Bind a readonly property
    .def_readwrite("n_friends", &Dog::n_friends); // Bind a read and write property
}
```

Running `stubgen -m python_example` generates the following stub file
```python
from typing import Any

class Dog:
    def __init__(self) -> None: ...
    def get_age(self) -> int: ...
    @property
    def age(self) -> Any: ...
    @property
    def n_friends(self) -> Any: ...
    @n_friends.setter
    def n_friends(self, val: Any) -> None: ...
```
We can seed that while `stubgen` properly detect type for the `get_name` method, all properties are being created with `Any`.

## Expected behavior
It would be preferable if `stubgen` generated the following stub file
```python
from typing import Any

class Dog:
    def __init__(self) -> None: ...
    def get_age(self) -> int: ...
    @property
    def age(self) -> int: ...
    @property
    def n_friends(self) -> int: ...
    @n_friends.setter
    def n_friends(self, val: int) -> None: ...
```

## Docstring information
Pybind11 sets the following information in the docstrings
```python
In [1]: import python_example

In [2]: python_example.Dog.get_age.__doc__
Out[2]: 'get_age(self: python_example.Dog) -> int\n'

In [3]: python_example.Dog.age.__doc__
Out[3]: ''

In [4]: python_example.Dog.age.fget.__doc__
Out[4]: '(self: python_example.Dog) -> int\n'

In [5]: python_example.Dog.age.fset.__doc__

In [6]: python_example.Dog.n_friends.fget.__doc__
Out[6]: '(self: python_example.Dog) -> int\n'

In [7]: python_example.Dog.n_friends.fset.__doc__
Out[7]: '(self: python_example.Dog, arg0: int) -> None\n'

```

## Technical information
* mypy version: mypy 0.790+dev. 068594eb38fc550d834b7e61745177c423f2b224
* Python 3.8.2
* MacOs 10.15.5
* Compiler Apple clang version 11.0.3 (clang-1103.0.32.62) Target: x86_64-apple-darwin19.5.0

AntoinePrv:
Close #8995

There are two things at play here.
- Read the docstring signature in the first line of C type properties;
- Read the docstring on `property.fget` if the type cannot be inferred from the property doc itself.

The solution is a bit tailored to PyBind, in particular it can read signature that do not have names like `(self: TestType) -> PropType`.

### DIFF
--- a/mypy/stubdoc.py
+++ b/mypy/stubdoc.py
@@ -247,6 +247,14 @@ def infer_arg_sig_from_docstring(docstr: str) -> List[ArgSig]:
     return []
 
 
+def infer_ret_type_sig_from_docstring(docstr: str) -> Optional[str]:
+    """Convert signature in form of "(self: TestClass, arg0) -> int" to their return type."""
+    ret = infer_sig_from_docstring("stub" + docstr.strip(), "stub")
+    if ret:
+        return ret[0].ret_type
+    return None
+
+
 def parse_signature(sig: str) -> Optional[Tuple[str,
                                                 List[str],
                                                 List[str]]]:
--- a/mypy/stubdoc.py
+++ b/mypy/stubdoc.py
@@ -239,7 +239,7 @@ def infer_sig_from_docstring(docstr: str, name: str) -> Optional[List[FunctionSi
     return [sig for sig in sigs if is_unique_args(sig)]
 
 
-def infer_arg_sig_from_docstring(docstr: str) -> List[ArgSig]:
+def infer_arg_sig_from_anon_docstring(docstr: str) -> List[ArgSig]:
     """Convert signature in form of "(self: TestClass, arg0: str='ada')" to List[TypedArgList]."""
     ret = infer_sig_from_docstring("stub" + docstr, "stub")
     if ret:
@@ -247,7 +247,7 @@ def infer_arg_sig_from_docstring(docstr: str) -> List[ArgSig]:
     return []
 
 
-def infer_ret_type_sig_from_docstring(docstr: str) -> Optional[str]:
+def infer_ret_type_sig_from_anon_docstring(docstr: str) -> Optional[str]:
     """Convert signature in form of "(self: TestClass, arg0) -> int" to their return type."""
     ret = infer_sig_from_docstring("stub" + docstr.strip(), "stub")
     if ret:
--- a/mypy/test/teststubgen.py
+++ b/mypy/test/teststubgen.py
@@ -25,7 +25,7 @@ from mypy.stubgenc import (
 from mypy.stubdoc import (
     parse_signature, parse_all_signatures, build_signature, find_unique_signatures,
     infer_sig_from_docstring, infer_prop_type_from_docstring, FunctionSig, ArgSig,
-    infer_arg_sig_from_docstring, is_valid_type
+    infer_arg_sig_from_anon_docstring, is_valid_type
 )
 from mypy.moduleinspect import ModuleInspect, InspectError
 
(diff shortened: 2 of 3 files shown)


mypy/test/teststubgen.py
```diff
@@ -778,6 +780,20 @@ def test(arg0: str) -> None:
         assert_equal(output, ['def test(arg0: str) -> Action: ...'])
         assert_equal(imports, [])
 
+    def test_generate_c_property_with_pybind11(self) -> None:
+        """Signatures included by PyBind11 inside property.fget are read."""
+        class TestClass:
+            def get_attribute(self) -> str:
```
JukkaL, line 786: Now the type seems to be also available through the return type here. Maybe leave out the return type to make sure that we can figure out the property type from the docstring only? Or use a different return type in the annotation to check that docstring takes precedence.

mypy/test/teststubgen.py
```diff
@@ -778,6 +780,20 @@ def test(arg0: str) -> None:
...
+                """
+                (self: TestClass) -> str
+                """
+                pass
+            attribute = property(get_attribute, doc="")
```
JukkaL, line 791: Would it make sense to also have a test case which provides the docstring here as the `doc` argument?

mypy/stubdoc.py
```diff
@@ -247,6 +247,14 @@ def infer_arg_sig_from_docstring(docstr: str) -> List[ArgSig]:
     return []
 
 
+def infer_ret_type_sig_from_docstring(docstr: str) -> Optional[str]:
```
JukkaL, line 250: What about renaming the function to indicate that this parses a return type using an alternative docstring syntax (`infer_arg_sig_from_docstring` is also similar and could be renamed)? Maybe add an `_alt` suffix or something. The main different seems to be that this syntax is missing the function name in the docstring.

mypy/test/teststubgen.py
AntoinePrv: Good point!

mypy/test/teststubgen.py
AntoinePrv: So this is not the behavior of PyBind, which only document `fget` `fset` given that they have separate signature.
That being said, `generate_c_property_stub` does check for that docstring (to avoid managing individually too many conditions).
It could be another test, but I believe the question really is what is the expected behavior here?

mypy/stubdoc.py
AntoinePrv: `infer_ret_type_sig_from_partial_docstring`? `infer_ret_type_sig_from_anon_docstring`?

AntoinePrv:
@JukkaL this is what seems most reasonable to me. Please let me know what you think.

>Also, have you tested that this works as expected with some module that uses pybind11?

I have tested with the Pybind example of the associated issue #8995 and it produce the desired output.

PR #9004, opened 2020-06-15
[mypyc] Speed up in operations for list/tuple

jdahlin:
When right hand side of a in/not in operation is a literal
list/tuple, simplify it into simpler direct equality comparision
expressions and use binary and/or to join them. 

Part of mypyc/mypyc#726, but this only speeds up list/tuple.

This is my first contribution to mypy/mypyc, please let me know if there's anything I can do to improve the pull request. I ended up create new tree nodes (OpExpr/ComparisionExpr) inside IRBuilder which is probably not generating as efficient as it can be. If that needs to change let me know and please provide me with some pointers on how to build a more efficient ir. Happy to add tests for the specific IR generated as well if needed.

I didn't do any macro benchmarks on mypy itself, would be happy to know what's normally benchmarked and instructions on how to do so.

```
# before (x = 10)
2000000 loops, best of 5: 113 nsec per loop  # x in [1]
2000000 loops, best of 5: 116 nsec per loop  # x in [1, 2]
2000000 loops, best of 5: 128 nsec per loop  # x in [1, 2, 3]
2000000 loops, best of 5: 136 nsec per loop  # x in [1, 2, 3, 4]
2000000 loops, best of 5: 145 nsec per loop  # x in [1, 2, 3, 4, 5]

5000000 loops, best of 5: 88.7 nsec per loop # x in (1)
5000000 loops, best of 5: 97.6 nsec per loop # x in (1, 2)
2000000 loops, best of 5: 108 nsec per loop  # x in (1, 2, 3)
2000000 loops, best of 5: 118 nsec per loop  # x in (1, 2, 3, 4)
2000000 loops, best of 5: 129 nsec per loop  # x in (1, 2, 3, 4, 5)

# after (x = 10)
5000000 loops, best of 5: 54.8 nsec per loop  # x in [1]
5000000 loops, best of 5: 55.9 nsec per loop  # x in [1, 2]
5000000 loops, best of 5: 56 nsec per loop  # x in [1, 2, 3]
5000000 loops, best of 5: 55.5 nsec per loop  # x in [1, 2, 3, 4]
5000000 loops, best of 5: 55.1 nsec per loop  # x in [1, 2, 3, 4, 5]

5000000 loops, best of 5: 55 nsec per loop  # x in (1)   
5000000 loops, best of 5: 55.1 nsec per loop    # x in (1, 2)
5000000 loops, best of 5: 54.9 nsec per loop  # x in (1, 2, 3) 
5000000 loops, best of 5: 55.8 nsec per loop  # x in (1, 2, 3, 4) 
5000000 loops, best of 5: 55.6 nsec per loop  # x in (1, 2, 3, 4, 5) 

For reference, using CPython 3.8.2:

5000000 loops, best of 5: 57 nsec per loop  # x in (1)   
5000000 loops, best of 5: 61.6 nsec per loop    # x in (1, 2)
5000000 loops, best of 5: 72 nsec per loop  # x in (1, 2, 3) 
5000000 loops, best of 5: 76.8 nsec per loop  # x in (1, 2, 3, 4) 
5000000 loops, best of 5: 84.6 nsec per loop  # x in (1, 2, 3, 4, 5) 
```


mypyc/irbuild/expression.py
```diff
@@ -352,8 +352,32 @@ def transform_conditional_expr(builder: IRBuilder, expr: ConditionalExpr) -> Val
 
 
 def transform_comparison_expr(builder: IRBuilder, e: ComparisonExpr) -> Value:
-    # TODO: Don't produce an expression when used in conditional context
+    # x in (...)/[...]
+    if e.operators[0] in ['in', 'not in'] and isinstance(e.operands[1], (TupleExpr, ListExpr)):
```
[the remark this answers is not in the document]
jdahlin: Good point, < 16 is way more likely to see in real code, n_items == 1 is probably rare and n_items == 0 is just wrong, should perhaps be an error.

mypyc/irbuild/expression.py
```diff
@@ -352,8 +352,32 @@ def transform_conditional_expr(builder: IRBuilder, expr: ConditionalExpr) -> Val
...
+        # # x not in [y]/(y) -> x != y
+        elif n_items == 1:
+            e.operators = [cmp_op]
+            e.operands[1] = items[0]
+        # x in y -> x == y[0] or ... or x == y[n]
+        # 16 is arbitrarily chosen to limit code size
+        elif n_items < 16:
+            bin_op = 'or' if e.operators[0] == 'in' else 'and'
+            lhs = e.operands[0]
+            exprs = (ComparisonExpr([cmp_op], [lhs, item]) for item in items)
```
msullivan, line 374: I /think/ that since these expressions don't appear in the type table, they will get coerced to object, leading to some kind of pointless boxing. I guess that isn't actually that expensive for bools, but it's worth cleaning up. Fine to just put a TODO in for now if you don't feel like cleaning it up now.

jdahlin: Thanks for the review @msullivan, I'd be happy to fix this but I would probably need some more specific pointers of what needs to be modified for that to work.

jdahlin: For what it's worth, `x in [1, 2, 3]` generates:

```
def f(i):
    i :: int
    r0, r1 :: object
    r2 :: short_int
    r3 :: bool
    r4 :: object
    r5 :: bool
    r6 :: short_int
    r7 :: bool
    r8 :: object
    r9 :: bool
    r10 :: short_int
    r11 :: bool
    r12 :: object
    r13 :: bool
L0:
    r2 = 1
    r3 = i == r2 :: int
    r4 = box(bool, r3)
    r5 = bool r4 :: object
    if r5 goto L1 else goto L2 :: bool
L1:
    r1 = r4
    goto L3
L2:
    r6 = 2
    r7 = i == r6 :: int
    r8 = box(bool, r7)
    r1 = r8
L3:
    r9 = bool r1 :: object
    if r9 goto L4 else goto L5 :: bool
L4:
    r0 = r1
    goto L6
L5:
    r10 = 3
    r11 = i == r10 :: int
    r12 = box(bool, r11)
    r0 = r12
L6:
    r13 = unbox(bool, r0)
    return r13
```

jdahlin: @msullivan I ended up, perhaps somewhat hacky to shortcut all OpExpr/ComparisonExpr without types as bool_primitive. That removed the excessive box/unboxing, with these changes it's significantly faster, I measured somewhere between 46%-78% for micro benchmarks. Seems like it triggers some happy path in the C compiler, as the length of the tuple/list is no longer relevant for performance. (tested up to 16 items)

msullivan: Happy to hear it is a lot faster! I don't like this hack, though.

I think there are two ways forward:
 * Directly generate the code without creating an AST for it first. One way to do this would involve using `shortcircuit_helper`, though also it could probably just be done directly.
 * Put all of the generated expressions into the type table. This could probably be done ergonomically by adding a helper method that takes an expression and a type, adds it to the table, and returns the expression.

Historically though we haven't really done AST generation as part of compilation (I think mostly because it would require populating the type table), but it seems fine to allow if it is done tastefully.
(@JukkaL, do you agree?)

jdahlin: No problem, I'll populate the type table with these types then, `shortcut_helper` scared me a bit, it seemed more straight forward to me to generate AST nodes for this issue.

msullivan:
Looks good. Check out the lint failure though

mypyc/irbuild/expression.py
```diff
@@ -352,8 +352,55 @@ def transform_conditional_expr(builder: IRBuilder, expr: ConditionalExpr) -> Val
 
 
 def transform_comparison_expr(builder: IRBuilder, e: ComparisonExpr) -> Value:
-    # TODO: Don't produce an expression when used in conditional context
+    # x in (...)/[...]
+    # x not in (...)/[...]
+    if e.operators[0] in ['in', 'not in'] and isinstance(e.operands[1], (TupleExpr, ListExpr)):
```
r3m0t, line 357:
```suggestion
    if e.operators[0] in ['in', 'not in'] and len(e.operators) == 1 and isinstance(e.operands[1], (TupleExpr, ListExpr)):
```

jdahlin: Sure, but I'd like to understand when will the number of operators in a ComparisonExpr != 1 ?

TH3CHARLie: > Sure, but I'd like to understand when will the number of operators in a ComparisonExpr > 1 ?

From `checkexpr.py`->`visit_comparison_expr`'s comment, expression like `a < b > c == d` would have multiple operators

JukkaL:
@msullivan @jdahlin I wonder what needs to be done (beyond fixing merge conflicts) to get this merged? This would be really nice to have.

jdahlin:
I believe nothing else is needed apart from the merge conflicts. I will see
if I can fix these in the next couple of days.

jdahlin:
@JukkaL I've finished merging it with latest master, all tests pass now.

mypyc/test-data/irbuild-tuple.test
```diff
@@ -181,3 +181,67 @@ L2:
...
+    r15 = i == 6
+    r12 = r15
+    goto L14
+L13:
+    r16 = CPyTagged_IsEq_(i, 6)
+    r12 = r16
+L14:
+    r0 = r12
+L15:
+    return r0
```
TH3CHARLie, line 247:
```suggestion
    return r0

```
add an extra newline

TH3CHARLie (approved):
LGTM! Thanks for speeding up mypyc!

JukkaL:
Thanks! I expect to see some nice improvements in benchmark results after the next nightly run (https://github.com/mypyc/mypyc-benchmarks).

jdahlin:
Thanks for merging it @JukkaL! Is there a way to see the results for the nightly build somewhere? I'm also curious about the effect of this on larger pieces of code.

TH3CHARLie:
> Thanks for merging it @JukkaL! Is there a way to see the results for the nightly build somewhere? I'm also curious about the effect of this on larger pieces of code.

https://github.com/mypyc/mypyc-benchmark-results/blob/master/reports/summary-microbenchmarks.md would be the place to find out the related results. The performance boost would be less significant on larger code though.

JukkaL:
For many optimizations microbenchmarks work well to estimate the level of improvement. Even for a very good optimization the impact to most major benchmarks can be below the measurement noise floor, or we might have no major benchmarks that happen to use the targeted feature in sufficient volume to be affected.

This should get gradually better as we add more benchmarks. If we have an optimization that isn't reflected in any existing benchmarks, it may be a good idea to add another (micro)benchmark to catch regressions in the future.

TH3CHARLie:
From the results of microbenchmark `in_list` and `in_tuple`, the performance boost of this PR is huge.

PR #9018, opened 2020-06-18
Exclude cache directories from backups using CACHEDIR.TAG

jstasiak:
This helps to prevent bloating backups with caches.

See https://bford.info/cachedir/ for more information about the
specification – it's supported by Borg, restic, GNU Tar and other
solutions.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -267,6 +267,7 @@ def _build(sources: List[BuildSource],
             reports.finish()
         if not cache_dir_existed and os.path.isdir(options.cache_dir):
             add_catch_all_gitignore(options.cache_dir)
+            exclude_from_backups(options.cache_dir)
 
 
 def default_data_dir() -> str:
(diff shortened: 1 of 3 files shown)


jstasiak:
I'm wondering if this maybe should always (re)create the `CACHEDIR.TAG` file in `.mypy_cache` to avoid situations where Mypy is interrupted after creating the directory and before the file is created, leaving the directory `CACHEDIR.TAG`-less forever. I estimate the chance of this to be slim, but given enough people using it there'll be a handful who'll encounter that condition I image and the code would not be much more complicated, the relevant lines would just read

```python
if os.path.isdir(options.cache_dir):
    exclude_from_backups(options.cache_dir)
```

jstasiak:
Following up on the concern mentioned above, we could attempt to "create" the directory + gitignore + CACHEDIR.TAG atomically by creating a temporary directory, creating files in it and then renaming the temporary directory to the desired path (like I did in a [patch to Cargo](https://github.com/rust-lang/cargo/pull/8378/commits/2793d7bf62aae0212c0c31430069445966aa2c26)) but this would have to happen somewhere deeper down the stack, in the place where the actual cache directory is created now (I'm not sure where is it).

jstasiak:
FWIW I had a quick look around and the only other open source backup solution I found (I only looked at open source ones) that supports this out of the box is attic, others (like rsync and bacula) support this in a exclude-directory-if-a-file-named-like-this-exists-in-it way – so while my initial commit message may be technically correct I wanted the above to be known, I don't want to oversell this solution on false premises.

msullivan (approved):
Yeah, sure.

It might be worth handling the interrupted case, yeah. It would be fine to just check every time. If you're excited about that, submit a follow-up.

jstasiak:
Cheers.

PR #9022, opened 2020-06-18
Add changes to ignore all vim temporary files

asheux:
Vim creates a number of temporary files but the current settings in .gitignore only ignores ```.swp``` files.
- with this change, all vim created temporary files prefixed by ```sw```  will be ignored
- examples of potential files; swn, swo, swp, etc.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,8 +23,11 @@ dmypy.json
 
 # IDEs
 .idea
-*.swp
 .vscode
+
+# vim temporary files
+.*.sw?
+*.sw?
 
 # Operating Systems
 .DS_Store


.gitignore
```diff
@@ -23,9 +23,11 @@ dmypy.json
 
 # IDEs
 .idea
-*.swp
 .vscode
 
+# vim temporary files
+.*.sw?
```
ethanhs, line 29: As I understand it, this will only work for vim swap files where the file starts with a dot, because in git ignore syntax, `*` need not be preceded by a `.`.

asheux: Yes. You are absolutely right. What my change actually does is ignore ```.tmp.sw?``` files. I should specify that in the PR description. Also, I'll add the correct syntax to ignore ```.sw?``` files. Thanks.

asheux: Did some digging and according to https://github.com/github/gitignore/blob/master/Global/Vim.gitignore, these changes should do the trick recursively.

asheux:
> Alright, I think this looks good.

@ethanhs, is this ready to go?

PR #9030, opened 2020-06-21
[mypyc] Introduce subcommand to mypyc command line interface

TH3CHARLie:
related mypyc/mypyc#739

### DIFF
--- a/scripts/mypyc
+++ b/scripts/mypyc
@@ -124,7 +124,7 @@ def mypyc_clean(options: Options) -> None:
         # TODO: this simply hardcodes what generates now
         # a better solution is to use some configuration files
         # maybe use a `.mypyc_cache` folder to store them
-        with os.scandir(options.build_dir) as entries:
+        with os.scandir(options.build_dir) as entries:  # type: ignore
             for entry in entries:
                 if entry.is_dir() and entry.path.find("temp") != -1:
                     directories.append(entry.path)
--- a/scripts/mypyc
+++ b/scripts/mypyc
@@ -41,7 +41,7 @@ mode_mapping = {
     'build': MODE_BUILD,
     'run': MODE_RUN,
     'clean': MODE_CLEAN
-}  # type: Dict[str, strß]
+}  # type: Dict[str, str]
 
 
 class Options:
--- a/scripts/mypyc
+++ b/scripts/mypyc
@@ -19,6 +19,7 @@ import tempfile
 import time
 import argparse
 from typing import Dict, List
+from typing_extensions import Final
 import shutil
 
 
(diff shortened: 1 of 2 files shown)


scripts/mypyc
```diff
@@ -27,29 +30,128 @@ from mypyc.build import mypycify
...
+    import_command = '"import {}"'.format(module_name)
+    # TODO: well this is dumb, we'd come up with a better way
+    env = os.environ.copy()
+    base_path = os.path.join(os.path.dirname(__file__), '..')
+    env['PYTHONPATH'] = base_path + os.pathsep + env.get('PYTHONPATH', '')
+    # TODO: How do we guarantee that we run the compiled version when both .py and .so files
+    # are in the same directory?
+    cmd = subprocess.run("python -c " + import_command, shell=True)
+    # TODO: I don't know why this is not working...
+    # cmd = subprocess.run([sys.executable, 'cmd', import_command], env=env, capture_output=True)
```
JukkaL, line 116: Should `'cmd'` be `'-c'` instead? We don't want to use `shell=True`, since who knows which shell the user might use, and how it might be configured.

scripts/mypyc
```diff
@@ -27,29 +30,128 @@ from mypyc.build import mypycify
...
+                if entry.is_file() and (entry.path.endswith('.c') or entry.path.endswith('.h')
+                                        or entry.path.find('ops.txt') != -1):
+                    files.append(entry.path)
+        for file in files:
+            try:
+                os.remove(file)
+            except OSError as e:
+                print('Error "{}" occurred when removing {}'.format(e.strerror, file))
+        for directory in directories:
+            shutil.rmtree(directory)
```
JukkaL, line 140: Can you also delete the `build` directory, if it's empty now? Also, we should delete the shared libraries outside the build directory (but somehow only those what have been compiled with mypyc using this build directory).

scripts/mypyc
```diff
@@ -27,29 +30,128 @@ from mypyc.build import mypycify
...
     sys.exit(cmd.returncode)
 
+
+def mypyc_clean(options: Options) -> None:
+    if os.path.exists(options.build_dir):
+        files = []
+        directories = []
+        # TODO: this simply hardcodes what generates now
+        # a better solution is to use some configuration files
+        # maybe use a `.mypyc_cache` folder to store them
```
JukkaL, line 126: Brainstorming: maybe the `build` directory should be renamed to something with `mypyc` in the name, perhaps `__mypyc__` or something, and then we can assume that everything in the directory are owned by us can be deleted.

TH3CHARLie: `__mypyc__` would be a decent choice, but I am worried about the generated shared object files? Currently, we place them under cwd. To clean them, we probably need to search all `*.so` in cwd and delete them, which IMO is not a good approach.

JukkaL: Deleting the `.so` files would make the command much more useful, I think. 

We should be able to maintain the paths to all generated `.so` files in a file under `build`/`__mypyc__` and delete only the files we have generated. If we are still worried that we may deleting the wrong files, we can also record the mtime and size of the files, and only delete files if the mtime and size match what we have recorded.

TH3CHARLie: This approach looks way better. Let's try go with `__mypyc__` to place all the `*.c, *.h, ops.txt, setup.py`(or any other intermediate scripts) and `__mypyc__/.mypyc_meta`(the name can be tweak) to record such file information.

TH3CHARLie:
I've lost the branch for this both locally and on Github, so I am creating a new PR for this.

PR #9035, opened 2020-06-23
[mypyc] new error_kind and branch variant to handle call_negative_bool_emit

TH3CHARLie:
related mypyc/mypyc#734 and mypyc/mypyc#741.

Introducing `ERR_NEG_INT` error_kind and `NEG_INT_EXPR` branch variant to support checking the return value of a call is non-negative. `set.discard` is used as an example. With some modifications, this would also support `negative_int_emit`.

### DIFF
--- a/mypyc/codegen/emitfunc.py
+++ b/mypyc/codegen/emitfunc.py
@@ -13,7 +13,7 @@ from mypyc.ir.ops import (
     BasicBlock, Value, MethodCall, PrimitiveOp, EmitterInterface, Unreachable, NAMESPACE_STATIC,
     NAMESPACE_TYPE, NAMESPACE_MODULE, RaiseStandardError, CallC, LoadGlobal
 )
-from mypyc.ir.rtypes import RType, RTuple, is_c_int_rprimitive
+from mypyc.ir.rtypes import RType, RTuple, is_int32_rprimitive, is_int64_rprimitive
 from mypyc.ir.func_ir import FuncIR, FuncDecl, FUNC_STATICMETHOD, FUNC_CLASSMETHOD
 from mypyc.ir.class_ir import ClassIR
 
@@ -182,7 +182,7 @@ class FunctionEmitterVisitor(OpVisitor[None], EmitterInterface):
 
     def visit_load_int(self, op: LoadInt) -> None:
         dest = self.reg(op)
-        if is_c_int_rprimitive(op.type):
+        if is_int32_rprimitive(op.type) or is_int64_rprimitive(op.type):
             self.emit_line('%s = %d;' % (dest, op.value))
         else:
             self.emit_line('%s = %d;' % (dest, op.value * 2))
--- a/mypyc/codegen/emitfunc.py
+++ b/mypyc/codegen/emitfunc.py
@@ -105,6 +105,9 @@ class FunctionEmitterVisitor(OpVisitor[None], EmitterInterface):
         if op.op == Branch.BOOL_EXPR:
             expr_result = self.reg(op.left)  # right isn't used
             cond = '{}{}'.format(neg, expr_result)
+        elif op.op == Branch.NEG_INT_EXPR:
+            expr_result = self.reg(op.left)
+            cond = '{} < 0'.format(expr_result)
         elif op.op == Branch.IS_ERROR:
             typ = op.left.type
             compare = '!=' if op.negated else '=='
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -20,13 +20,12 @@ mypyc.irbuild.mapper.Mapper.type_to_rtype converts mypy Types to mypyc
 RTypes.
 """
 
-import sys
 from abc import abstractmethod
 from typing import Optional, Union, List, Dict, Generic, TypeVar
 
 from typing_extensions import Final, ClassVar, TYPE_CHECKING
 
-from mypyc.common import JsonDict, short_name
+from mypyc.common import JsonDict, short_name, IS_32_BIT_PLATFORM
 from mypyc.namegen import NameGenerator
 
 if TYPE_CHECKING:
@@ -175,6 +174,8 @@ class RPrimitive(RType):
         self.is_unboxed = is_unboxed
         self._ctype = ctype
         self.is_refcounted = is_refcounted
+        # TODO: For low-level integers, they actually don't have undefined values
+        #       we need to figure out some way to represent here.
         if ctype in ('CPyTagged', 'int32_t', 'int64_t'):
             self.c_undefined = 'CPY_INT_TAG'
         elif ctype == 'PyObject *':
@@ -242,7 +243,7 @@ int64_rprimitive = RPrimitive('int64', is_unboxed=True, is_refcounted=False,
                               ctype='int64_t')  # type: Final
 # integer alias
 c_int_rprimitive = int32_rprimitive
-if sys.maxsize < (1 << 31):
+if IS_32_BIT_PLATFORM:
     c_pyssize_t_rprimitive = int32_rprimitive
 else:
     c_pyssize_t_rprimitive = int64_rprimitive
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -20,6 +20,7 @@ mypyc.irbuild.mapper.Mapper.type_to_rtype converts mypy Types to mypyc
 RTypes.
 """
 
+import sys
 from abc import abstractmethod
 from typing import Optional, Union, List, Dict, Generic, TypeVar
 
@@ -241,7 +242,10 @@ int64_rprimitive = RPrimitive('int64', is_unboxed=True, is_refcounted=False,
                               ctype='int64_t')  # type: Final
 # integer alias
 c_int_rprimitive = int32_rprimitive
-c_pyssize_t_rprimitive = int64_rprimitive
+if sys.maxsize < (1 << 31):
+    c_pyssize_t_rprimitive = int32_rprimitive
+else:
+    c_pyssize_t_rprimitive = int64_rprimitive
 
 # Floats are represent as 'float' PyObject * values. (In the future
 # we'll likely switch to a more efficient, unboxed representation.)
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -174,7 +174,7 @@ class RPrimitive(RType):
         self.is_unboxed = is_unboxed
         self._ctype = ctype
         self.is_refcounted = is_refcounted
-        if ctype in ('CPyTagged', 'int32_t', 'int64_t'):
+        if ctype in ('CPyTagged', 'int', 'Py_ssize_t'):
             self.c_undefined = 'CPY_INT_TAG'
         elif ctype == 'PyObject *':
             # Boxed types use the null pointer as the error value.
@@ -236,9 +236,9 @@ short_int_rprimitive = RPrimitive('short_int', is_unboxed=True, is_refcounted=Fa
 
 # low level integer (corresponds to C's 'int's).
 c_int32_rprimitive = RPrimitive('c_int32', is_unboxed=True, is_refcounted=False,
-                              ctype='int32_t')  # type: Final
+                              ctype='int')  # type: Final
 c_int64_rprimitive = RPrimitive('c_int64', is_unboxed=True, is_refcounted=False,
-                              ctype='int64_t')  # type: Final
+                              ctype='Py_ssize_t')  # type: Final
 # integer alias
 c_int_rprimitive = c_int32_rprimitive
 c_pyssize_t_rprimitive = c_int64_rprimitive
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -235,13 +235,13 @@ short_int_rprimitive = RPrimitive('short_int', is_unboxed=True, is_refcounted=Fa
                                   ctype='CPyTagged')  # type: Final
 
 # low level integer (corresponds to C's 'int's).
-c_int32_rprimitive = RPrimitive('c_int32', is_unboxed=True, is_refcounted=False,
+int32_rprimitive = RPrimitive('int32', is_unboxed=True, is_refcounted=False,
                               ctype='int')  # type: Final
-c_int64_rprimitive = RPrimitive('c_int64', is_unboxed=True, is_refcounted=False,
+int64_rprimitive = RPrimitive('int64', is_unboxed=True, is_refcounted=False,
                               ctype='Py_ssize_t')  # type: Final
 # integer alias
-c_int_rprimitive = c_int32_rprimitive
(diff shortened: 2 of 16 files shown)


TH3CHARLie:
Generated code:

before:
```c
    PyObject *cpy_r_r0;
    char cpy_r_r1;
    char cpy_r_r2;
    PyObject *cpy_r_r3;
CPyL0: ;
    CPyTagged_IncRef(cpy_r_y);
    cpy_r_r0 = CPyTagged_StealAsObject(cpy_r_y);
    cpy_r_r1 = PySet_Discard(cpy_r_x, cpy_r_r0) >= 0;
    CPy_DecRef(cpy_r_r0);
    if (unlikely(!cpy_r_r1)) {
        CPy_AddTraceback("foo.py", "fossss", 4, CPyStatic_globals);
        goto CPyL2;
    } else
        goto CPyL1;
```

after:
```c
    PyObject *cpy_r_r0;
    Py_ssize_t cpy_r_r1;
    char cpy_r_r2;
    PyObject *cpy_r_r3;
CPyL0: ;
    CPyTagged_IncRef(cpy_r_y);
    cpy_r_r0 = CPyTagged_StealAsObject(cpy_r_y);
    cpy_r_r1 = PySet_Discard(cpy_r_x, cpy_r_r0);
    CPy_DecRef(cpy_r_r0);
    if (unlikely(cpy_r_r1 < 0)) {
        CPy_AddTraceback("foo.py", "fossss", 4, CPyStatic_globals);
        goto CPyL2;
    } else
        goto CPyL1;
```

TH3CHARLie:
And for the remaining `negative_int_emit` case(the other one `call_negative_magic_emit` depends on it), for a simple isinstance statement, it generates following C code:
```c
    int __tmp1 = PyObject_IsInstance(cpy_r_x, cpy_r_r0);
    if (__tmp1 < 0)
        cpy_r_r1 = 2;
    else
        cpy_r_r1 = __tmp1;
        CPy_AddTraceback("foo.py", "check", 8, CPyStatic_globals);
        goto CPyL2;
    } else
        goto CPyL1;
```
It actually checks if the temp is < 0 and then check the error value, which I believe may be replaced with a single < 0 check in this PR? @JukkaL

JukkaL:
The overall approach looks good. This also simplifies the generated code, which is nice.

One detail that I noticed is that it would be better to use the `int` type instead of `Py_ssize_t` for the return value of `PySet_Discard`. In the low-level IR converting an integer to a different width should be made explicit. It's fine to do this in a separate PR, since we don't have a way to represent `int` yet.

> It actually checks if the temp is < 0 and then check the error value(ERR_MAGIC), which I believe may be replaced with a single < 0 check in this PR?

Yeah, the `PyObject_IsInstance` case can be modified to only do a single check, and after that we can narrow down the result to a `bool`. (Getting rid of the conversion to bool in simple cases would be nice, but it's out of scope for now.)

TH3CHARLie:
Well, some strange errors running `run.test`
```
running build_ext
building 'native' extension
creating build/temp.macosx-10.15-x86_64-3.7
creating build/temp.macosx-10.15-x86_64-3.7/build
clang -Wno-unused-result -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk -I/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include -I/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/System/Library/Frameworks/Tk.framework/Versions/8.5/Headers -I/Users/th3charlie/Documents/mypy/mypyc/lib-rt -I/usr/local/include -I/usr/local/opt/openssl@1.1/include -I/usr/local/opt/sqlite/include -I/Users/th3charlie/Documents/venv/include -I/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/include/python3.7m -c build/__native.c -o build/temp.macosx-10.15-x86_64-3.7/build/__native.o -O3 -Werror -Wno-unused-function -Wno-unused-label -Wno-unreachable-code -Wno-unused-variable -Wno-unused-command-line-argument -Wno-unknown-warning-option
creating build/lib.macosx-10.15-x86_64-3.7
clang -bundle -undefined dynamic_lookup -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk build/temp.macosx-10.15-x86_64-3.7/build/__native.o -L/usr/local/lib -L/usr/local/opt/openssl@1.1/lib -L/usr/local/opt/sqlite/lib -o build/lib.macosx-10.15-x86_64-3.7/native.cpython-37m-darwin.so
copying build/lib.macosx-10.15-x86_64-3.7/native.cpython-37m-darwin.so -> 

*** Exit status: -11
```
compile these testcases using mypyc via command line would work, though.

mypyc/ir/rtypes.py
```diff
@@ -234,9 +234,14 @@ def __repr__(self) -> str:
 short_int_rprimitive = RPrimitive('short_int', is_unboxed=True, is_refcounted=False,
                                   ctype='CPyTagged')  # type: Final
 
-# low level integer (corresponds to C's 'int').
-c_int_rprimitive = RPrimitive('c_int', is_unboxed=True, is_refcounted=False,
+# low level integer (corresponds to C's 'int's).
+c_int32_rprimitive = RPrimitive('c_int32', is_unboxed=True, is_refcounted=False,
```
JukkaL, line 238: I have a few nits.

I think that we can leave the `c_` prefix out, since "32-bit integer" is a fairly generic concept not specific to C and not likely to be confused with Python types.

Also, the correspondence to C `int` is a bit of an accident (and not generally true, on all possible platforms). It would be better to use `int32_t`, which happens to be the same as `int` on all supported platforms.

mypyc/ir/rtypes.py
```diff
@@ -234,9 +234,14 @@ def __repr__(self) -> str:
...
+                              ctype='int')  # type: Final
+c_int64_rprimitive = RPrimitive('c_int64', is_unboxed=True, is_refcounted=False,
                               ctype='Py_ssize_t')  # type: Final
```
JukkaL, line 241: See the above comment. (In particular, use `int64_t` as the type.)

mypyc/ir/rtypes.py
```diff
@@ -234,9 +234,14 @@ def __repr__(self) -> str:
...
+# integer alias
+c_int_rprimitive = c_int32_rprimitive
```
JukkaL, line 243: Here it's fine to use a `c_` prefix.

mypyc/ir/rtypes.py
```diff
@@ -234,9 +234,14 @@ def __repr__(self) -> str:
...
+c_pyssize_t_rprimitive = c_int64_rprimitive
```
JukkaL, line 244: The definition should depend on the platform. If we are on a 64-bit platform, it's expected to be `int64_rprimitive`, but on a 32-bit platform it's `int32_rprimitive`. (We don't care about exotic platforms such as 32-bit addresses with 64-bit registers.)

mypyc/ir/rtypes.py
```diff
@@ -279,10 +284,22 @@ def is_short_int_rprimitive(rtype: RType) -> bool:
...
+
+def is_c_int64_rprimitive(rtype: RType) -> bool:
+    return rtype is c_int64_rprimitive
+
+
 def is_c_int_rprimitive(rtype: RType) -> bool:
     return rtype is c_int_rprimitive
 
 
+def is_c_pyssize_t_rprimitive(rtype: RType) -> bool:
```
JukkaL, line 299: I wonder if we can leave out the `is_<type>` functions for the aliased types? Maybe only having them defined for `int32` and `int64` would be sufficient?

mypyc/ir/rtypes.py
TH3CHARLie: Actually at commit f3011ad, I've tried to use `int32_t` and `int64_t` just as we discussed. But somehow the above error comes up, I am actually debugging that and once I find out which of this modification leads to that error and resolve it, I'll make sure we have `int32_t` and `int64_t` as the corresponding C types.

mypyc/ir/rtypes.py
TH3CHARLie: for now at least we need `is_c_int_rprimitive` as we've used that in `emitfunc.py`.

mypyc/primitives/set_ops.py
```diff
@@ -70,13 +72,12 @@ def emit_len(emitter: EmitterInterface, args: List[str], dest: str) -> None:
...
 # set.discard(obj)
-method_op(
+c_method_op(
     name='discard',
     arg_types=[set_rprimitive, object_rprimitive],
-    result_type=bool_rprimitive,
-    error_kind=ERR_FALSE,
-    emit=call_negative_bool_emit('PySet_Discard')
-)
+    return_type=c_int_rprimitive,
```
TH3CHARLie, line 78: The problem above(the magic error -11) is caused by this line of change, if we alias the `c_int_rprimtive` to the type represents 64 bit int(both `int64_t` or `Py_ssize_t` on a 64bit machine) would work. But changing it to `int32_t`  or int would fail.

mypyc/codegen/emitfunc.py
```diff
@@ -179,7 +182,7 @@ def visit_assign(self, op: Assign) -> None:
 
     def visit_load_int(self, op: LoadInt) -> None:
         dest = self.reg(op)
-        if is_c_int_rprimitive(op.type):
+        if is_int32_rprimitive(op.type) or is_int64_rprimitive(op.type):
```
TH3CHARLie, line 185: I've located the key to fix this. I guess that although I have not create any `LoadInt` with `int64_rprimitive` explicitly. And therefore all loaded int64 value would double and fail eventually.

mypyc/ir/rtypes.py
TH3CHARLie: Do we have ways of detecting the current platform?

TH3CHARLie:
For the `negative_int_emit` cases, as we discussed yesterday it's wise to introduce a truncate op, therefore, we can coerce the int return value of the called function to the real boolean value we want for the primitive to have. To achieve that, we'd modify the description and also `CallC` accordingly. But what confuses me is where should we place the coerce/truncate. I'd think of two options if we follow the style we do with `call_negative_bool_emit:

1. add this coercion as part of CallC so that CallC with coercion has two processes. Call a function and then truncate. The IR would be like
```
r0 :: int32_t
r1 :: bool
r0 = PyObject_IsInstance(obj1, obj2)
r1 = r0
```
But the problem is clear that direct coercion provides no help. To keep the current semantics, we need an additional branch to finally know whether it should be r0 or error_value to set into r1.

2. add this coercion in exception handling. We can introduce a more complicated error check logic for `negative_int_emit`, like firstly check and truncate the int return value to bool and then check it for error. But this would make the primitive has no reasonable value until all this completes(recall that we want a bool value from this primitive, but until all this comparison->truncate->error-check logic completes we can only have an int value returned from the c function).

@JukkaL

JukkaL:
> add this coercion as part of CallC so that CallC with coercion has two processes. Call a function and then truncate. The IR would be like

I think that `call_c` can generate the CallC op followed by a truncate op (the truncate op would be a new RegisterOp subclass that never generates an error). We can rely on the exception handling transform to add a negative check after the call. The final IR would look something like this (without refcount handling):

```
r0 :: int32_t
r1 :: bool
r0 = PyObject_IsInstance(obj1, obj2)
if r0 < 0 goto ...
r1 = trunc r0 int32_t to bool
```

mypyc/ir/rtypes.py
JukkaL: You can use something like `sys.maxsize < (1 << 31)` to check if we are on a 32-bit platform.

mypyc/ir/rtypes.py
JukkaL: It looks like you figured this out, good!

TH3CHARLie:
> ```
> r0 :: int32_t
> r1 :: bool
> r0 = PyObject_IsInstance(obj1, obj2)
> if r0 < 0 goto ...
> r1 = trunc r0 int32_t to bool
> ```

If we have python code like and the change you mentioned:
```python
result = isinstance(x, str)
```
How should we make sure(which part to change) that result gets `r1` instead of `r0` if we generate two ops: `CallC` and `Truncate`, we actually need to reference the `Truncate`'s value in python code like this.

TH3CHARLie:
Ohh that's bad, we have both 32/64 bit CI platforms but the IR tests text are fixed(`int64` for now).

mypyc/ir/rtypes.py
```diff
@@ -241,7 +242,10 @@ def __repr__(self) -> str:
                               ctype='int64_t')  # type: Final
 # integer alias
 c_int_rprimitive = int32_rprimitive
-c_pyssize_t_rprimitive = int64_rprimitive
+if sys.maxsize < (1 << 31):
```
JukkaL, line 245: You can store the result of this check into a constant in `mypyc.common` so that it can be easily reused.

mypyc/ir/rtypes.py
```diff
@@ -174,7 +175,7 @@ def __init__(self,
         self.is_unboxed = is_unboxed
         self._ctype = ctype
         self.is_refcounted = is_refcounted
-        if ctype in ('CPyTagged', 'Py_ssize_t'):
+        if ctype in ('CPyTagged', 'int32_t', 'int64_t'):
             self.c_undefined = 'CPY_INT_TAG'
```
JukkaL, line 179: For low-level integers we actually don't have undefined values, since all values values are possible (at least in some contexts). We don't have a good way to represent this right now, so maybe add a TODO comment about this?

JukkaL:
Hmm some tests are failing because the types are different on 32-bit platforms. To work around this, you can make it possible to represent an integer type that varies by the register width in test cases (e.g. `native_int`); this can be replaced (in the test case description) by the test runner with `int32` or `int64`, depending on the platform. Alternatively, we can skip some of the tests on 32-bit platforms, but this is not as good, since testing these types is actually useful.

The test runner is in `test_irbuild.py` (and there are others like that, such as `test_exceptions.py`, that may also need to be changed).

mypyc/test/test_irbuild.py
```diff
@@ -42,7 +42,9 @@ def run_case(self, testcase: DataDrivenTestCase) -> None:
         """Perform a runtime checking transformation test case."""
         with use_custom_builtins(os.path.join(self.data_prefix, ICODE_GEN_BUILTINS), testcase):
             expected_output = remove_comment_lines(testcase.output)
-
+            # replace native_int with platform specific ints
+            int_format_str = 'int32' if IS_32_BIT_PLATFORM else 'int64'
+            expected_output = [s.replace('native_int', int_format_str) for s in expected_output]
```
JukkaL, line 47: Since the same code is repeated twice, it would make sense to move this into a helper function. This can be done in a separate PR and isn't really necessary, since it's only two lines of code.

PR #9037, opened 2020-06-23
pythoneval.test: update for typeshed change

hauntsaninja:
Change this so we can land https://github.com/python/typeshed/pull/4258


hauntsaninja:
The test failures are the expected ones.

PR #9050, opened 2020-06-26
[mypyc] handle negative_int_emit and Truncate op

TH3CHARLie:
related mypyc/mypyc#734

### DIFF
--- a/mypyc/ir/ops.py
+++ b/mypyc/ir/ops.py
@@ -1183,6 +1183,33 @@ class CallC(RegisterOp):
         return visitor.visit_call_c(self)
 
 
+class Truncate(RegisterOp):
+
+    error_kind = ERR_NEVER
+
+    def __init__(self,
+                 src: Value,
+                 src_type: RType,
+                 dst_type: RType,
+                 line: int = -1) -> None:
+        super().__init__(line)
+        self.src = src
+        self.src_type = src_type
+        self.type = dst_type
+
+    def sources(self) -> List[Value]:
+        return [self.src]
+
+    def stolen(self) -> List[Value]:
+        return [self.src]
+
+    def to_str(self, env: Environment) -> str:
+        return env.format("truncate %r: %r to %r", self.src, self.src_type, self.type)
+
+    def accept(self, visitor: 'OpVisitor[T]') -> T:
+        return visitor.visit_truncate(self)
+
+
 class LoadGlobal(RegisterOp):
     """Load a global variable/pointer"""
 
@@ -1305,6 +1332,10 @@ class OpVisitor(Generic[T]):
 
     @abstractmethod
     def visit_call_c(self, op: CallC) -> T:
+        raise NotImplementedError
+
+    @abstractmethod
+    def visit_truncate(self, op: Truncate) -> T:
         raise NotImplementedError
 
     @abstractmethod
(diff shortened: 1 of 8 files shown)


TH3CHARLie:
There will be auto-generated `Box`, `Unbox`, `DecRef` sequence right after the `CallC`
Screen Shot 2020-06-26 at 2 30 54 PM
in the following code's case, this sequence will try to check whether the return value from CallC is a boolean.
```python
def foo(x: object) ->bool:
    z = isinstance(x, str)
    return z
```

TH3CHARLie:
With this PR, for the following python function:
```python
def foo(x: object) -> bool:
    z = isinstance(x, str)
    return z
```
The C code before:
```c
char CPyDef_foo(PyObject *cpy_r_x) {
    PyObject *cpy_r_r0;
    char cpy_r_r1;
    char cpy_r_z;
    char cpy_r_r2;
CPyL0: ;
    cpy_r_r0 = (PyObject *)&PyUnicode_Type;
    int __tmp1 = PyObject_IsInstance(cpy_r_x, cpy_r_r0);
    if (__tmp1 < 0)
        cpy_r_r1 = 2;
    else
        cpy_r_r1 = __tmp1;
        CPy_AddTraceback("../foo.py", "foo", 2, CPyStatic_globals);
        goto CPyL2;
    } else
        goto CPyL1;
CPyL1: ;
    cpy_r_z = cpy_r_r1;
    return cpy_r_z;
CPyL2: ;
    cpy_r_r2 = 2;
    return cpy_r_r2;
}
```
after:
```c
char CPyDef_foo(PyObject *cpy_r_x) {
    PyObject *cpy_r_r0;
    int32_t cpy_r_r1;
    char cpy_r_r2;
    char cpy_r_z;
    char cpy_r_r3;
CPyL0: ;
    cpy_r_r0 = (PyObject *)&PyUnicode_Type;
    cpy_r_r1 = PyObject_IsInstance(cpy_r_x, cpy_r_r0);
    if (unlikely(cpy_r_r1 < 0)) {
        CPy_AddTraceback("../foo.py", "foo", 2, CPyStatic_globals);
        goto CPyL2;
    } else
        goto CPyL1;
CPyL1: ;
    cpy_r_r2 = cpy_r_r1;
    cpy_r_z = cpy_r_r2;
    return cpy_r_z;
CPyL2: ;
    cpy_r_r3 = 2;
    return cpy_r_r3;
}
```
We actually reduce the comparison by 1 even though we introduce a new op in its IR.

mypyc/ir/ops.py
```diff
@@ -1183,6 +1183,33 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
         return visitor.visit_call_c(self)
 
 
+class Truncate(RegisterOp):
```
JukkaL, line 1186: Add docstring (similar to other ops).

Mention that this only works between non-reference counted integer types or bool (not `int_rprimitive`).

mypyc/primitives/registry.py
```diff
@@ -367,7 +369,7 @@ def c_method_op(name: str,
         priority: if multiple ops match, the one with the highest priority is picked
```
JukkaL, line 369: Document `truncated_type`. Mention that if defined, it and return type must be non-reference counted integer types (or bool).

TH3CHARLie:
Once this is merged, I'd start to merge the remaining ops during the weekends.

PR #9051, opened 2020-06-26
Enforce annotation when inferred item in list, dict or set is object type

### LINKED ISSUE #3816: Require an annotation if inferred list or dict item type is object

ChetanKhanna:
Fixes #3816 

- Fixes failing tests from PR 6792
- Adds suggested annotation note
- Adds annotation checker for sets
- Adds tests

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -15,7 +15,7 @@ from mypy.nodes import (
     SymbolTable, Statement, MypyFile, Var, Expression, Lvalue, Node,
     OverloadedFuncDef, FuncDef, FuncItem, FuncBase, TypeInfo,
     ClassDef, Block, AssignmentStmt, NameExpr, MemberExpr, IndexExpr,
-    TupleExpr, ListExpr, ExpressionStmt, ReturnStmt, IfStmt,
+    TupleExpr, ListExpr, DictExpr, SetExpr, ExpressionStmt, ReturnStmt, IfStmt,
     WhileStmt, OperatorAssignmentStmt, WithStmt, AssertStmt,
     RaiseStmt, TryStmt, ForStmt, DelStmt, CallExpr, IntExpr, StrExpr,
     UnicodeExpr, OpExpr, UnaryExpr, LambdaExpr, TempNode, SymbolTableNode,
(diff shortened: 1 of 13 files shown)


ChetanKhanna:
Thanks to the original creator of https://github.com/python/mypy/pull/6792 ! :)

I haven't looked into the flag to disable this check as discussed in the issue as of now. Any suggestions as to where the code for it should go and/or how should I go about it?

test-data/unit/check-inference.test
```diff
@@ -3176,3 +3176,27 @@ reveal_type(f(lambda: 1))  # N: Revealed type is 'builtins.int*'
...
 reveal_type(f(g))  # N: Revealed type is 'None'
+
+-- Tests for checking if annotation is needed
+-- --------------
+
+[case testexplicitAnnotationNeededForLists]
+x = [1, 'two', 3.5]  # E: Need type annotation for 'x' \
+                     # N: Suggested annotation: List[Any] or List[Union[float, int, str]]
+
+y = [2, 'three', object()]
```
ChetanKhanna, line 3187: I wanted to add a similar test for ``set`` but I guess ``/fixtures/set.pyi`` does not supprt ``float``. So the test was failing. Is there a reason that this is missing from the set fixture?

JelleZijlstra: Not really, we tend to just put things in fixtures when we need them. Feel free to add it, but adding things to fixtures incurs a slight performance penalty, so if possible it may be better to switch to another fixture that does have float.

Or just use a different type than float.

ChetanKhanna: Oh okay! I wasn't sure about fixtures and hence the confusion. I'll redo the test without floats, it wouldn't make much of a difference. Thank you!

JukkaL:
> I haven't looked into the flag to disable this check as discussed in the issue as of now. Any suggestions as to where the code for it should go and/or how should I go about it?

Maybe #6562 serves as an example of how to add a flag?

Also, can you rename the PR to describe the new feature, and include "Fixes #3816." in the description?

ChetanKhanna:
> Maybe #6562 serves as an example of how to add a flag?

This was helpful, thanks! I added the flag and a corresponding test for it.

ChetanKhanna:
Hey there! Sorry I didn't keep track of this. Is this still needed? Should I rebase it?

mypy/checker.py
```diff
@@ -2147,8 +2147,120 @@ def check_assignment(self, lvalue: Lvalue, rvalue: Expression, infer_lvalue_type
...
+        >>> x = [1, 'a', object()]  # OK
+        Dict keys may be inferred to type `object` without annotation:
+        >>> x = ['a': 1, 2: 2]  # OK
+        """
+        if not self.scope.top_function() and self.options.allow_untyped_globals:
+            return
+
+        rvalue_type = get_proper_type(rvalue_type)
+        if isinstance(rvalue_type, Instance):
+            if isinstance(rvalue, ListExpr):
```
msullivan, line 2171: I think it should be possible to reduce most of the duplicate between the three cases with an appropriate helper function.

msullivan:
Yeah, I'd appreciate that! I've left what I think is my main comment, so if you address that and rebase it, we should be able to get this merged.

Thanks.

hauntsaninja:
Not saying we shouldn't do this, but since this has the potential to be a disruptive change, I ran mypy_primer on it: `mypy_primer --repo https://github.com/ChetanKhanna/mypy.git --new issue-3816 --old 6ee562a8f3e69 -o concise --clear`. It looks like it's around 325 new errors in 4M lines of code, which is concerning for a fix meant to improve usability. Fwiw, a quick skim indicates most of the new errors are for dicts.

msullivan:
> Not saying we shouldn't do this, but since this has the potential to be a disruptive change, I ran mypy_primer on it: `mypy_primer --repo https://github.com/ChetanKhanna/mypy.git --new issue-3816 --old 6ee562a8f3e69 -o concise --clear`. It looks like it's around 325 new errors in 4M lines of code, which is concerning for a fix meant to improve usability. Fwiw, a quick skim indicates most of the new errors are for dicts.

Yeah, I would definitely not take this if it wasn't behind a flag. But that does raise a question about whether it is worth adding a flag for this. @JukkaL @ilevkivskyi thoughts?

msullivan (changes requested):

mypy/options.py
```diff
@@ -136,6 +136,9 @@ def __init__(self) -> None:
         # Apply strict None checking
         self.strict_optional = True
 
+        # Testing new flag
```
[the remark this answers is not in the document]
ChetanKhanna: Yes I must have left it there while debugging or something. I removed it.

mypy/main.py
```diff
@@ -612,6 +612,14 @@ def add_invertible_flag(flag: str,
         '--strict', action='store_true', dest='special-opts:strict',
         help=strict_help)
 
+    forbidden_inference_group = parser.add_argument_group(
+        title='Handling forbidden inference related type checks',
+        description="Warn when infered type is List[object], Set[Object] or Dict[*, Object].")
+    add_invertible_flag('--suppress-need-type-annotation-errors', default=True,
```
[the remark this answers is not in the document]
ChetanKhanna: Changed now. I was unsure of the naming so I had left it to whatever first popped in my mind

mypy/checker.py
ChetanKhanna: I tried this. The left over redundancy is due to the fact that the `Dict` case is handled a little differently from the `List` and `Set` case. Let me know if you think more could be done here.

ChetanKhanna:
> Not saying we shouldn't do this, but since this has the potential to be a disruptive change, I ran mypy_primer on it: `mypy_primer --repo https://github.com/ChetanKhanna/mypy.git --new issue-3816 --old 6ee562a8f3e69 -o concise --clear`. It looks like it's around 325 new errors in 4M lines of code, which is concerning for a fix meant to improve usability.

I did not get a chance to look into this. But I will once I set up `mypy_primer` (I wasn't aware of it, thanks :smile:).

>  Fwiw, a quick skim indicates most of the new errors are for dicts.

Maybe `DictExpr` part needs changes? :thinking:

JukkaL:
Closing due to inactivity.

PR #9059, opened 2020-06-27
Use absolute path when checking source duplication error

### LINKED ISSUE #9058: "Source file found twice under different module names" doesn't work in some cases
I found that "Source file found twice under different module names" error doesn't happen in some cases.

### Files
Same as `case testDuplicateModules` in `test-data/unit/cmdline.test` ([source](https://github.com/python/mypy/blob/8219564d365ba29dda8c4383594d9db91f26feec/test-data/unit/cmdline.test#L1071-L1084)).
```
.
├── mypy.ini
└── src
    ├── a.py
    ├── foo
    │   ├── bar.py
    │   └── __init__.py
    └── __init__.py
```

### Actual behavior
```sh
$ mypy src
src/foo/bar.py: error: Source file found twice under different module names: 'src.foo.bar' and 'foo.bar'
Found 1 error in 1 file (checked 4 source files)

$ mypy `pwd`/src
src/foo/bar.py:1: error: Unsupported operand types for + ("int" and "str")
src/foo/bar.py:1: error: Unsupported operand types for + ("int" and "str")
Found 2 errors in 1 file (checked 4 source files)
```

### Expected behavior

```sh
$ mypy src
src/foo/bar.py: error: Source file found twice under different module names: 'src.foo.bar' and 'foo.bar'
Found 1 error in 1 file (checked 4 source files)

$ mypy `pwd`/src
src/foo/bar.py: error: Source file found twice under different module names: 'src.foo.bar' and 'foo.bar'
Found 1 error in 1 file (checked 4 source files)
```

### Files
Same as `case testDuplicateModules` in `test-data/unit/cmdline.test` except `mypy.ini`.

```
.
├── mypy.ini
└── src
    ├── a.py
    ├── foo
    │   ├── bar.py
    │   └── __init__.py
    └── __init__.py
```

- mypy.ini
```
[mypy]
mypy_path = /absolute/path/to/src
```

### Actual behavior
```sh
$ mypy src
src/foo/bar.py:1: error: Unsupported operand types for + ("int" and "str")
src/foo/bar.py:1: error: Unsupported operand types for + ("int" and "str")
Found 2 errors in 1 file (checked 4 source files)

$ mypy `pwd`/src
src/foo/bar.py: error: Source file found twice under different module names: 'src.foo.bar' and 'foo.bar'
Found 1 error in 1 file (checked 4 source files)
```

### Expected behavior

```sh
$ mypy src
src/foo/bar.py: error: Source file found twice under different module names: 'src.foo.bar' and 'foo.bar'
Found 1 error in 1 file (checked 4 source files)

$ mypy `pwd`/src
src/foo/bar.py: error: Source file found twice under different module names: 'src.foo.bar' and 'foo.bar'
Found 1 error in 1 file (checked 4 source files)
```

## Versions
```sh
$ mypy --version
mypy 0.790+dev.8219564d365ba29dda8c4383594d9db91f26feec
$ python --version
Python 3.6.10
```

bonprosoft:
closes #9058 

I updated both `seen_files` and `newst` to use an absolute path to fix the problem.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -1699,6 +1699,7 @@ class State:
     order = None  # type: int  # Order in which modules were encountered
     id = None  # type: str  # Fully qualified module name
     path = None  # type: Optional[str]  # Path to module source
+    abspath = None  # type: Optional[str]  # Absolute path to module source
     xpath = None  # type: str  # Path or '<string>'
     source = None  # type: Optional[str]  # Module source code
     source_hash = None  # type: Optional[str]  # Hash calculated based on the source code
@@ -1800,6 +1801,8 @@ class State:
             if follow_imports == 'silent':
                 self.ignore_all = True
         self.path = path
+        if path:
+            self.abspath = os.path.abspath(path)
         self.xpath = path or '<string>'
         if path and source is None and self.manager.fscache.isdir(path):
             source = ''
@@ -2758,7 +2761,7 @@ def load_graph(sources: List[BuildSource], manager: BuildManager,
 
     # Note: Running this each time could be slow in the daemon. If it's a problem, we
     # can do more work to maintain this incrementally.
-    seen_files = {os.path.abspath(st.path): st for st in graph.values() if st.path}
+    seen_files = {st.abspath: st for st in graph.values() if st.path}
 
     # Collect dependencies.  We go breadth-first.
     # More nodes might get added to new as we go, but that's fine.
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2811,11 +2811,16 @@ def load_graph(sources: List[BuildSource], manager: BuildManager,
 
                         if seen_state is None:
                             if os.path.isabs(newst_path):
-                                newst_path = os.path.relpath(newst_path)
+                                try:
+                                    newst_path = os.path.relpath(newst_path)
+                                except ValueError:
+                                    # Ignore when newst_path does not start with os.curdir
+                                    pass
+                                else:
+                                    seen_state = seen_files.get(newst_path)
                             else:
                                 newst_path = os.path.abspath(newst_path)
-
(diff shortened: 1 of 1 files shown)


mypy/build.py
```diff
@@ -2754,7 +2754,7 @@ def load_graph(sources: List[BuildSource], manager: BuildManager,
 
     # Note: Running this each time could be slow in the daemon. If it's a problem, we
     # can do more work to maintain this incrementally.
-    seen_files = {st.path: st for st in graph.values() if st.path}
+    seen_files = {os.path.abspath(st.path): st for st in graph.values() if st.path}
```
JukkaL, line 2757: This code is very performance critical, and `os.path.abspath` is too slow to call here since it can make a syscall. We'd need a way to do this without introducing much overhead (graph may have 100k+ values).

bonprosoft: Thank you for your review!
I understand. Then, how about looking up both absolute path and relative path of `newst.path` in `seen_files` at this part:
https://github.com/python/mypy/blob/d9c5c65d3421af67fe8ecc598430f85ad87c8319/mypy/build.py#L2805-L2807

bonprosoft: I have updated my PR: https://github.com/python/mypy/pull/9059/commits/c1f7d1e4c458cf2bdb57e2ef23e9ed0ca5c6dc24
Could you please review if it is reasonable?

msullivan: How about going back to the original approach, but adding an `abspath` field to `State` that is computed in State's constructor, instead of recomputing it on each load_graph?

(In the daemon, load_graph gets called a lot on a big graph but with only a few sources changed. We do some things that are linear in the graph but we feel bad about it and try to make sure that they aren't *expensive* things.)

msullivan: OK actually I just went and did that thing. Thank you!

Issue #8986, opened 2020-06-11
Bug: INTERNAL ERROR from := assignment expression within list comprehension


Jasha10:
I've encountered a bug when running mypy on a file with an assignment expression contained within a list comprehension.
```# mwe.py
# mwe.py
lst = [1, 2, 3, 4]
processed = [e for elt in lst if (e := elt - 1) > 1]
print(processed)
```
I am running the following:
mypy version: 0.790+dev.068594eb38fc550d834b7e61745177c423f2b224
CPython version: 3.8.3, installed with anaconda

Here is the traceback:
```
$ mypy --show-traceback mwe.py
mwe.py:3: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 0.790+dev.068594eb38fc550d834b7e61745177c423f2b224
Traceback (most recent call last):
  File "/home/jbss/anaconda3/envs/tmp/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/home/jbss/anaconda3/envs/tmp/lib/python3.8/site-packages/mypy/__main__.py", line 8, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "/home/jbss/anaconda3/envs/tmp/lib/python3.8/site-packages/mypy/main.py", line 89, in main
    res = build.build(sources, options, None, flush_errors, fscache, stdout, stderr)
  File "/home/jbss/anaconda3/envs/tmp/lib/python3.8/site-packages/mypy/build.py", line 180, in build
    result = _build(
  File "/home/jbss/anaconda3/envs/tmp/lib/python3.8/site-packages/mypy/build.py", line 252, in _build
    graph = dispatch(sources, manager, stdout)
  File "/home/jbss/anaconda3/envs/tmp/lib/python3.8/site-packages/mypy/build.py", line 2626, in dispatch
    process_graph(graph, manager)
  File "/home/jbss/anaconda3/envs/tmp/lib/python3.8/site-packages/mypy/build.py", line 2949, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/home/jbss/anaconda3/envs/tmp/lib/python3.8/site-packages/mypy/build.py", line 3041, in process_stale_scc
    mypy.semanal_main.semantic_analysis_for_scc(graph, scc, manager.errors)
[...]

Issue #8379, opened 2020-02-07
Walrus operators unrecognised in dictionary initialisation of a class attribute


masasakano:
It seems Walrus operators (in Python 3.8+) are not recognised by mypy when used in initialising a dictionary of a class attribute. Here is a test code.

```python
class TryWalrus:
    CKEY = { 
        'KY': (p:='a_'),
         'ANOTHER': p+'content',
        }

print(TryWalrus.CKEY['ANOTHER']) # => a_content
```

This outputs the following:

    $ mypy try_walrus.py
    try_walrus.py:4: error: Name 'p' is not defined
    Found 1 error in 1 file (checked 1 source file)

If the variable `CKEY` is out of the class definition, i.e. if directly in Main, it does not raise an error.

I tested it with Python 3.8.1 and mypy version `0.770+dev.8888b1aedf6d4ba543ae12f538f190ce5e65c20d` (yesterday's head).

dhood:
This will also happen with a list e.g. `l = [x2 := 1, x2 + 2, x2 + 3]` (for this case also it will only happen if it is in the class scope).

I have noticed, however, that it only happens on RE-USE of the variable from the assignment expression within the initialisation. Otherwise the walrus operator is ok.

```
class Example:
    # l = [x2 := 1, x2 + 2, x2 + 3] # not OK
    l = [x2 := 1, 2, 3] # OK
```

bunny-therapist:
I just (2023) ran into this exact problem with
```
mypy 1.4.1 (compiled: yes)
Python 3.11.0
```

it176131:
Any proposed fix—besides `# type: ignore[name-defined]`—for this? I ran into it with the following versions:
```text
Python 3.12.2
mypy   1.10.0
```

JelleZijlstra:
Anyone impacted should feel free to work on a PR to fix this. I haven't looked at the code but I imagine it's a relatively easy fix.

PR #9062, opened 2020-06-28
Fix internal error on list/dict comprehension with walrus operator in global scope

dhood:
Fixes https://github.com/python/mypy/issues/8986

The reported issue is for list comprehensions involving assignment expressions, though I noticed that dictionary comprehensions are also impacted.

```
# mwe.py
lst = [1, 2, 3, 4]
processed_list = [e for elt in lst if (e := elt - 1) > 1]

dct = {'a': 1, 'b': 2, 'c': 3, 'd': 4}
processed_dict = {k: new_v for k, v in dct.items() if (new_v := v - 1) > 1
```

On the other hand, the same declarations _inside_ a function pass:
```
# valid_comprehensions.py
def calling_function() -> None:
    lst = [1, 2, 3, 4]
    processed_list = [e for elt in lst if (e := elt - 1) > 1]
    reveal_type(processed_list)

    dct = {'a': 1, 'b': 2, 'c': 3, 'd': 4}
    processed_dict = {k: new_v for k, v in dct.items() if (new_v := v - 1) > 1}
    reveal_type(processed_dict)
```

This tipped me off to it being related to the special logic added in #6899 to [store the variables in the scope of the comprehension's caller](https://github.com/python/mypy/pull/6899/files#diff-ebbe3135b7d21c5fa8783a4d0d0ccd48R4448), which I believe can benefit from special treatment for the edge case of the caller being in the global scope. That's the approach that I took, though it's my first contribution to MyPy so another approach may be more appropriate.

I added regression tests for these cases, but I also added some tests that were already passing. I've kept them in separate commits (with CI triggered separately) to be clear about which ones are being addressed by this PR!

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -4550,7 +4550,10 @@ class SemanticAnalyzer(NodeVisitor[None],
                             # The caller of the comprehension is in the global space.
                             names = self.globals
                         else:
-                            names = cast(SymbolTable, self.locals[-1 - i])
+                            names_candidate = self.locals[-1 - i]
+                            assert names_candidate is not None, \
+                                "Escaping comprehension from invalid scope"
+                            names = names_candidate
                         break
                 else:
                     assert False, "Should have at least one non-comprehension scope"
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -4542,9 +4542,15 @@ class SemanticAnalyzer(NodeVisitor[None],
         if self.is_func_scope():
             assert self.locals[-1] is not None
             if escape_comprehensions:
+                assert len(self.locals) == len(self.is_comprehension_stack)
+                # Retrieve the symbol table from the enclosing non-comprehension scope.
                 for i, is_comprehension in enumerate(reversed(self.is_comprehension_stack)):
                     if not is_comprehension:
-                        names = self.locals[-1 - i]
+                        if i == len(self.locals) - 1:  # The last iteration.
+                            # The caller of the comprehension is in the global space.
+                            names = self.globals
+                        else:
+                            names = cast(SymbolTable, self.locals[-1 - i])
                         break
                 else:
                     assert False, "Should have at least one non-comprehension scope"
--- a/test-data/unit/check-python38.test
+++ b/test-data/unit/check-python38.test
@@ -218,6 +218,19 @@ def f(x: int = (c := 4)) -> int:
     reveal_type(filtered_data)  # N: Revealed type is 'builtins.list[builtins.int*]'
     reveal_type(y3)  # N: Revealed type is 'builtins.int'
 
+    d = {'a': (a2 := 1), 'b': a2 + 1, 'c': a2 + 2}
+    reveal_type(d)  # N: Revealed type is 'builtins.dict[builtins.str*, builtins.int*]'
+    reveal_type(a2)  # N: Revealed type is 'builtins.int'
+
+    d2 = {(prefix := 'key_') + 'a': (start_val := 1), prefix + 'b': start_val + 1, prefix + 'c': start_val + 2}
+    reveal_type(d2)  # N: Revealed type is 'builtins.dict[builtins.str*, builtins.int*]'
+    reveal_type(prefix)  # N: Revealed type is 'builtins.str'
+    reveal_type(start_val)  # N: Revealed type is 'builtins.int'
+
+    filtered_dict = {k: new_v for k, v in [('a', 1), ('b', 2), ('c', 3)] if (new_v := v + 1) == 2}
+    reveal_type(filtered_dict)  # N: Revealed type is 'builtins.dict[builtins.str*, builtins.int*]'
+    reveal_type(new_v)  # N: Revealed type is 'builtins.int'
+
     # https://www.python.org/dev/peps/pep-0572/#exceptional-cases
     (y4 := 3)
     reveal_type(y4)  # N: Revealed type is 'builtins.int'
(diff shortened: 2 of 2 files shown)


test-data/unit/check-python38.test
```diff
@@ -198,6 +198,27 @@ if a := 2:
...
+d = {'a': (a2 := 1), 'b': a2 + 1, 'c': a2 + 2}
+reveal_type(d)  # N: Revealed type is 'builtins.dict[builtins.str*, builtins.int*]'
+reveal_type(a2)  # N: Revealed type is 'builtins.int'
+
+d2 = {(prefix := 'key_') + 'a': (start_val := 1), prefix + 'b': start_val + 1, prefix + 'c': start_val + 2}
+reveal_type(d2)  # N: Revealed type is 'builtins.dict[builtins.str*, builtins.int*]'
+reveal_type(prefix)  # N: Revealed type is 'builtins.str'
+reveal_type(start_val)  # N: Revealed type is 'builtins.int'
+
+filtered_dict = {k: new_v for k, v in [('a', 1), ('b', 2), ('c', 3)] if (new_v := v + 1) == 2}
```
dhood, line 218: If reviewers are wondering why I didn't just use `d.items()` here instead of the list of tuples, it's because it's causing `error: "Dict[str, int]" has no attribute "items"`.
I couldn't figure out why that was happening, I tried different flags and casting and whatnot but something in the test environment must be different... in a regular python script `filtered_dict = {k: new_v for k, v in d.items() if (new_v := v + 1) == 2}` can have its type revealed without such an error. Open to suggestions if you have ideas what's causing that!

JelleZijlstra: This is almost certainly because the test fixture you're using happens not to define `dict.items`. What you're doing here is fine.

mypy/semanal.py
```diff
@@ -4542,9 +4542,15 @@ def current_symbol_table(self, escape_comprehensions: bool = False) -> SymbolTab
...
+                assert len(self.locals) == len(self.is_comprehension_stack)
+                # Retrieve the symbol table from the enclosing non-comprehension scope.
                 for i, is_comprehension in enumerate(reversed(self.is_comprehension_stack)):
                     if not is_comprehension:
-                        names = self.locals[-1 - i]
+                        if i == len(self.locals) - 1:  # The last iteration.
+                            # The caller of the comprehension is in the global space.
+                            names = self.globals
+                        else:
+                            names = cast(SymbolTable, self.locals[-1 - i])
```
JelleZijlstra, line 4553: Why is this `cast()` needed? Could you use a `# type:` comment on the first assignment instead?

dhood:
> Sorry for missing it when I wrote the code. 

No, not at all. Certainly no anger or frustration from me, I am grateful of the initial implementation that you wrote! (which worked great in most cases, this is just an edge case). Actually I was excited to stumble upon this bug... I am very happy to have something small to contribute towards the support of the walrus operator! 😄

Will add those class-scope tests and ping you again.

mypy/semanal.py
dhood: the cast here was because `self.locals` is a list of `Optional[SymbolTable]`, and so without it, you would get:
```
mypy\semanal.py:4552: error: Incompatible types in assignment (expression has type "Optional[SymbolTable]", variable has type "SymbolTable")  [assignment]
                                names = self.locals[-1 - i]
                                        ^
```

I reasoned (at the time) that it is a safe cast as long as we are not retrieving the `None` that is the first element in the list.
https://github.com/python/mypy/blob/ed9ba4f87fadefa9de335246bf2c12135dec735c/mypy/semanal.py#L226

However, now that you have mentioned to check the class scope, I have realised that there may be other elements in the list that are None: https://github.com/python/mypy/blob/ed9ba4f87fadefa9de335246bf2c12135dec735c/mypy/semanal.py#L1166
Perhaps once I deal with the class case properly this cast will not be necessary. Either way, for now, my original assumption that _only_ the first entry is None seems invalid! So I will find an alternative approach.

dhood: Since comprehensions in a class scope e.g. 
```
class Example:
    [(j := i) for i in range(5)]
```
can also result in the retrieved `self.locals[-1 -i]` symbol table being None, I added extra logic for that case, which removed the need for the cast, in https://github.com/python/mypy/pull/9062/commits/a68bd68958cfa37bfc1a4d9d9946c07b3a63e518.

Because this specific case is called out in https://www.python.org/dev/peps/pep-0572/#scope-of-the-target as being invalid syntax (I know that you know this, I am just being verbose), the specific logic that's added is actually just an assertion that we don't end up in that case.

dhood:
@JelleZijlstra, I added tests for the class scope in https://github.com/python/mypy/pull/9062/commits/8585a4f11a929b56b5f83379fc6192a62f9a09bd

In doing so, I noticed some cases of what I consider inappropriate mypy behaviour:
1. The invalid syntax highlighted in https://www.python.org/dev/peps/pep-0572/#scope-of-the-target where a comprehension is used in a class scope does not raise a Syntax error. I opened https://github.com/python/mypy/issues/9095
1. Use of assignment operator within a list and subsequent re-use of the variable in the same list e.g. `l = [x2 := 1, x2 + 2]` from a class scope does not work. This is another variation of https://github.com/python/mypy/issues/8379, so I have commented there.
1. Re-use of variables defined in comprehensions (in any scope) is not permitted. Opened https://github.com/python/mypy/issues/9096

Since I consider these to be separate issues (they happen on master also), and because I didn't plan time to address those this weekend, I have spun them off to not block this PR.

dhood:
thanks for the review, @JelleZijlstra - I'll try to make progress on the spin-off issues this weekend

PR #9066, opened 2020-06-28
added project example to the Readme.md 

MedAmineFouzai:
you can see that mypy has taken a lot of attention when it comes to the future of python and creating a new web frameworks that uses it under the hood.


JukkaL:
This doesn't seem relevant in a readme.

PR #9067, opened 2020-06-29
[mypyc] Support argument reordering in CallC

TH3CHARLie:
related mypyc/mypyc#734.

Support argument reordering via adding a new field to the description. It will solve the difference in args ordering between python syntax and C calling requirement(mostly with `in` ops). It should never be used together with variable args.

### DIFF
--- a/mypyc/primitives/registry.py
+++ b/mypyc/primitives/registry.py
@@ -51,6 +51,7 @@ CFunctionDescription = NamedTuple(
                               ('c_function_name', str),
                               ('error_kind', int),
                               ('steals', StealsDescription),
+                              ('ordering', Optional[List[int]]),
                               ('priority', int)])
 
 # A description for C load operations including LoadGlobal and LoadAddress
@@ -352,6 +353,7 @@ def c_method_op(name: str,
                 error_kind: int,
                 var_arg_type: Optional[RType] = None,
                 truncated_type: Optional[RType] = None,
+                ordering: Optional[List[int]] = None,
                 steals: StealsDescription = False,
                 priority: int = 1) -> CFunctionDescription:
     """Define a c function call op that replaces a method call.
@@ -368,12 +370,15 @@ def c_method_op(name: str,
         truncated_type: type to truncated to(See Truncate for info)
                         if it's defined both return_type and it should be non-referenced
                         integer types or bool type
+        ordering: optional ordering of the arguments, if defined,
+                  reorders the arguments accordingly.
+                  should never be used together with var_arg_type
         steals: description of arguments that this steals (ref count wise)
         priority: if multiple ops match, the one with the highest priority is picked
     """
     ops = c_method_call_ops.setdefault(name, [])
     desc = CFunctionDescription(name, arg_types, return_type, var_arg_type, truncated_type,
-                                c_function_name, error_kind, steals, priority)
+                                c_function_name, error_kind, steals, ordering, priority)
     ops.append(desc)
     return desc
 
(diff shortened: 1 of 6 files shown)


TH3CHARLie:
for python code:
```python
def foo(x: Dict[int, int], y: int) -> bool:
    return y in x
```

before:
```c
char CPyDef_foo(PyObject *cpy_r_x, CPyTagged cpy_r_y) {
    PyObject *cpy_r_r0;
    char cpy_r_r1;
    char cpy_r_r2;
CPyL0: ;
    CPyTagged_IncRef(cpy_r_y);
    cpy_r_r0 = CPyTagged_StealAsObject(cpy_r_y);
    int __tmp1 = PyDict_Contains(cpy_r_x, cpy_r_r0);
    if (__tmp1 < 0)
        cpy_r_r1 = 2;
    else
        cpy_r_r1 = __tmp1;
    CPy_DecRef(cpy_r_r0);
        CPy_AddTraceback("../foo.py", "foo", 5, CPyStatic_globals);
        goto CPyL2;
    } else
        goto CPyL1;
CPyL1: ;
    return cpy_r_r1;
CPyL2: ;
    cpy_r_r2 = 2;
    return cpy_r_r2;
}
```

after:
```c
char CPyDef_foo(PyObject *cpy_r_x, CPyTagged cpy_r_y) {
    PyObject *cpy_r_r0;
    int32_t cpy_r_r1;
    char cpy_r_r2;
    char cpy_r_r3;
CPyL0: ;
    CPyTagged_IncRef(cpy_r_y);
    cpy_r_r0 = CPyTagged_StealAsObject(cpy_r_y);
    cpy_r_r1 = PyDict_Contains(cpy_r_x, cpy_r_r0);
    CPy_DecRef(cpy_r_r0);
    if (unlikely(cpy_r_r1 < 0)) {
        CPy_AddTraceback("../foo.py", "foo", 5, CPyStatic_globals);
        goto CPyL2;
    } else
        goto CPyL1;
CPyL1: ;
    cpy_r_r2 = cpy_r_r1;
    return cpy_r_r2;
CPyL2: ;
    cpy_r_r3 = 2;
    return cpy_r_r3;
}
```

mypyc/test/test_emitfunc.py
```diff
@@ -235,12 +235,7 @@ def test_new_dict(self) -> None:
     def test_dict_contains(self) -> None:
         self.assert_emit_binary_op(
             'in', self.b, self.o, self.d,
-            """int __tmp1 = PyDict_Contains(cpy_r_d, cpy_r_o);
-               if (__tmp1 < 0)
-                   cpy_r_r0 = 2;
-               else
-                   cpy_r_r0 = __tmp1;
-            """)
+            """cpy_r_r0 = PyDict_Contains(cpy_r_o, cpy_r_d);""")
```
JukkaL, line 238: This looks wrong. Should `assert_emit_binary_op` do the ordering as well?

mypyc/primitives/registry.py
```diff
@@ -368,12 +370,15 @@ def c_method_op(name: str,
         truncated_type: type to truncated to(See Truncate for info)
                         if it's defined both return_type and it should be non-referenced
                         integer types or bool type
+        ordering: optional ordering of the arguments, if defined,
+                  reorders the arguments accordingly.
+                  should never be used together with var_arg_type
```
JukkaL, line 375: Also mention that all the other arguments, such as `arg_types`, are in the order accepted by the called function (i.e., after reordering).

mypyc/test/test_emitfunc.py
TH3CHARLie: Good catch, this indeed looks like a bug

mypyc/primitives/registry.py
TH3CHARLie: I am a bit confused about this. `arg_types` are actually accepted by the python scope function(opposed to the called C function) and `ordering` is actually reordering `arg_types`

JukkaL: Sorry, I was confused. The order of all the other arguments are *before* reordering (Python ordering), right? Anyway, this illustrates that this is useful to document!

PR #9071, opened 2020-06-30
Add gray color ANSI escape sequence 

asheux:
Including ANSI escape sequence for gray color that is visible in both light and dark schemed terminal. The idea is to elimate the dim color attribute and replace it with gray color that is good for both light and dark scheme terminal to make the output  much more predictable. Below screenshots show how these changes improves the error view.

- Before on dark theme using dim capname

before_on_dark_theme

- After on dark theme using gray color

after_on_dark_theme

- Before on light theme using dim capname

before_on_light_theme

- After on light theme gray color

after_on_light_theme

### DIFF
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -26,13 +26,6 @@ T = TypeVar('T')
 ENCODING_RE = \
     re.compile(br'([ \t\v]*#.*(\r\n?|\n))??[ \t\v]*#.*coding[:=][ \t]*([-\w.]+)')  # type: Final
 
-# This works in most default terminals works (because it is ANSI standard). The problem
-# this tries to solve is that although it is a basic ANSI "feature", terminfo files
-# for most default terminals don't have dim termcap entry, so curses doesn't report it.
-# Potentially, we can choose a grey color that would look good on both white and black
-# background, but it is not easy, and again most default terminals are 8-color, not 256-color,
-# so we can't get the color code from curses.
-
 DEFAULT_SOURCE_OFFSET = 4  # type: Final
 
 # At least this number of columns will be shown on each side of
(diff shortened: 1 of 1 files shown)


JukkaL:
The Windows build is broken. 

It would be good to see some before/after screenshots to see how this improves things.

asheux:
@JukkaL, fix the tests. Added screenshots for confirmation.

asheux:
Any thoughts on this PR?

mypy/util.py
```diff
@@ -32,7 +32,6 @@
 # Potentially, we can choose a grey color that would look good on both white and black
 # background, but it is not easy, and again most default terminals are 8-color, not 256-color,
 # so we can't get the color code from curses.
```
msullivan, line 34: Drop this comment which is now commenting nothing?

msullivan:
This seems fine to me. I'll merge it unless @JukkaL or @ilevkivskyi object.

mypy/util.py
```diff
@@ -476,6 +468,13 @@ def hash_digest(data: bytes) -> str:
     return hashlib.sha256(data).hexdigest()
 
 
+def parse_gray_color(cup: bytes) -> str:
+    """Reproduce a gray color in ANSI escape sequence"""
+    set_color = ''.join([cup[:-1].decode(), 'm'])
+    gray = curses.tparm(set_color.encode('utf-8'), 1, 89).decode()
+    return gray
```
blueyed, line 475: This seems quite odd: "cup" is for "move to row/column", replacing its `H` with `m` in the end seems quite hackish.
Why not use `\e[90m` directly?
Ref: https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_parameters

PR #9073, opened 2020-06-30
[mypyc] Support ERR_ALWAYS

TH3CHARLie:
related mypyc/mypyc#734, with a focus on exceptions related ops.

This PR adds a new error kind: `ERR_ALWAYS`, which indicates the op always fail.

It adds temporary false value to ensure such behavior and makes the raise op void.

### DIFF
--- a/mypyc/transform/exceptions.py
+++ b/mypyc/transform/exceptions.py
@@ -13,7 +13,7 @@ from typing import List, Optional
 
 from mypyc.ir.ops import (
     BasicBlock, LoadErrorValue, Return, Branch, RegisterOp, ERR_NEVER, ERR_MAGIC,
-    ERR_FALSE, ERR_NEG_INT, NO_TRACEBACK_LINE_NO,
+    ERR_FALSE, ERR_NEG_INT, ERR_ALWAYS, NO_TRACEBACK_LINE_NO,
 )
 from mypyc.ir.func_ir import FuncIR
 
@@ -77,12 +77,16 @@ def split_blocks_at_errors(blocks: List[BasicBlock],
                 elif op.error_kind == ERR_NEG_INT:
                     variant = Branch.NEG_INT_EXPR
                     negated = False
+                elif op.error_kind == ERR_ALWAYS:
+                    variant = Branch.ALWAYS_FALSE
+                    negated = False
                 else:
                     assert False, 'unknown error kind %d' % op.error_kind
 
                 # Void ops can't generate errors since error is always
                 # indicated by a special value stored in a register.
-                assert not op.is_void, "void op generating errors?"
+                if op.error_kind != ERR_ALWAYS:
+                    assert not op.is_void, "void op generating errors?"
 
                 branch = Branch(op,
                                 true_label=error_label,
--- a/mypyc/transform/exceptions.py
+++ b/mypyc/transform/exceptions.py
@@ -12,10 +12,11 @@ only be placed at the end of a basic block.
 from typing import List, Optional
 
 from mypyc.ir.ops import (
-    BasicBlock, LoadErrorValue, Return, Branch, RegisterOp, ERR_NEVER, ERR_MAGIC,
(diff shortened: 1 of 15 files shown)


TH3CHARLie:
I find the way we deal with `negative_int_emit` may not work here.

Suppose we build an assignment op following the CallC via the call_c function(just like we build a truncate after a CallC)
```
r0 = call_c()
r1 = assign 0
```
However, the exception branch is generated directly following the CallC, so after the transform, it would be:
```
r0 = call_c()
if !r0 goto .. else goto...
r1 = assign 0
```
which clearly does not meet our goals.

JukkaL:
Hmm, the proposal we discussed today doesn't seem to work.

Here's another idea. Update exception transform to generate the Op for `tmp = 0`, followed by a BOOL_EXPR branch, when seeing `ERR_ALWAYS`. The temporary assignment will be optimized away. This would happen in `split_blocks_at_errors`, I think.

mypyc/transform/exceptions.py
```diff
@@ -77,14 +79,21 @@ def split_blocks_at_errors(blocks: List[BasicBlock],
                 elif op.error_kind == ERR_NEG_INT:
                     variant = Branch.NEG_INT_EXPR
                     negated = False
+                elif op.error_kind == ERR_ALWAYS:
+                    variant = Branch.BOOL_EXPR
+                    negated = True
+                    tmp = LoadInt(0, rtype=bool_rprimitive)
+                    cur_block.ops.append(tmp)
+                    target = tmp
```
TH3CHARLie, line 87: here's my attempt to add a tmp between the CallC and the generated branch. I've made the branch to reference this tmp's value when doing the comparison. However, it still fails at refcount transform pass:
```
  File "/Users/th3charlie/Documents/mypy/mypyc/transform/refcount.py", line 181, in insert_branch_inc_and_decrefs
    source_borrowed, source_live_regs, env, omitted),
  File "/Users/th3charlie/Documents/mypy/mypyc/transform/refcount.py", line 210, in after_branch_decrefs
    for reg in sorted(decref, key=lambda r: env.indexes[r])
  File "/Users/th3charlie/Documents/mypy/mypyc/transform/refcount.py", line 210, in <lambda>
    for reg in sorted(decref, key=lambda r: env.indexes[r])
KeyError: <mypyc.ir.ops.LoadInt object at 0x10700cdd0>
```
The keyerror is due to not adding the tmp op into the environment.

TH3CHARLie, line 87: here's my attempt to add a tmp between the CallC and the generated branch. I've made the branch to reference this tmp's value when doing the comparison. However, it still fails at refcount transform pass:
```
  File "/Users/th3charlie/Documents/mypy/mypyc/transform/refcount.py", line 181, in insert_branch_inc_and_decrefs
    source_borrowed, source_live_regs, env, omitted),
  File "/Users/th3charlie/Documents/mypy/mypyc/transform/refcount.py", line 210, in after_branch_decrefs
    for reg in sorted(decref, key=lambda r: env.indexes[r])
  File "/Users/th3charlie/Documents/mypy/mypyc/transform/refcount.py", line 210, in <lambda>
    for reg in sorted(decref, key=lambda r: env.indexes[r])
KeyError: <mypyc.ir.ops.LoadInt object at 0x10700cdd0>
```
The keyerror is due to not adding the tmp op into the environment.

TH3CHARLie:
I think this is ready for its first-round review @JukkaL

mypyc/transform/exceptions.py
```diff
@@ -77,14 +80,22 @@ def split_blocks_at_errors(blocks: List[BasicBlock],
                 elif op.error_kind == ERR_NEG_INT:
                     variant = Branch.NEG_INT_EXPR
                     negated = False
+                elif op.error_kind == ERR_ALWAYS:
+                    variant = Branch.BOOL_EXPR
+                    negated = True
+                    tmp = LoadInt(0, rtype=bool_rprimitive)
```
msullivan, line 86: Add a comment that this is kind of a hack?

JukkaL:
It's nice to finally have this!

PR #9088, opened 2020-07-02
stubgen can import Iterable and Iterator from typing

AWhetter:
pybind just gained support for defining the types of `Iterator`s and `Iterable`s (https://github.com/pybind/pybind11/commit/57070fb0a084e40ceab8a9f80b9e4436a48b9e53). This change makes stubgen capable of importing in generated stubs when used in a module being analysed.

### DIFF
--- a/mypy/stubgenc.py
+++ b/mypy/stubgenc.py
@@ -18,6 +18,19 @@ from mypy.stubdoc import (
 )
 
 
+_DEFAULT_TYPING_IMPORTS = (
+    'Any'
+    'Dict',
+    'Iterable',
+    'Iterator',
+    'List',
+    'Optional',
+    'Tuple',
+    'Union',
+)
+"""Members of the typing module to consider for importing by default."""
+
+
 def generate_stub_for_c_module(module_name: str,
                                target: str,
                                sigs: Optional[Dict[str, str]] = None,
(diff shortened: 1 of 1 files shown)


mypy/stubgenc.py
```diff
@@ -18,6 +18,19 @@
...
+    'Any'
+    'Dict',
+    'Iterable',
+    'Iterator',
+    'List',
+    'Optional',
+    'Tuple',
+    'Union',
+)
+"""Members of the typing module to consider for importing by default."""
```
JukkaL, line 31: Style nit: We don't use docstrings for variables. Please instead move this to a comment before the definition.

AWhetter:
Is there anything else that I can do to help this pull request progress?

mypy/stubgenc.py
```diff
@@ -18,6 +18,19 @@
 )
 
 
+# Members of the typing module to consider for importing by default.
+_DEFAULT_TYPING_IMPORTS = (
+    'Any'
```
dalcinl, line 23: @AWhetter There is a missing comma here, without it you get `'AnyDict'` as the first entry.

AWhetter: I've addressed this here: https://github.com/python/mypy/pull/9900

PR #9090, opened 2020-07-03
Use [arg-type] code for some one-off argument type error messages

### LINKED ISSUE #9083: "Argument has incompatible type" is [arg-type] and [misc]?
When running mypy on some code I get this error message:
`tesla_api/vehicle.py:70: error: Argument 2 to "setdefault" of "TypedDict" has incompatible type "None"; expected "int"  [misc]`

If i add a `type: ignore[misc]` to the line, then I get this error message:
`tesla_api/vehicle.py:70: error: Argument 2 to "setdefault" of "TypedDict" has incompatible type "None"; expected "int"  [arg-type]`

Seems that this error is under both misc and arg-type categories.

Excepted behaviour: The error should only appear under the arg-type category, and not require ignoring 2 categories.

JukkaL:
Fixes #9083.

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -854,7 +854,7 @@ class MessageBuilder:
                 suffix = ', not {}'.format(format_type(typ))
             self.fail(
                 'Argument after ** must be a mapping{}'.format(suffix),
-                context)
+                context, code=codes.ARG_TYPE)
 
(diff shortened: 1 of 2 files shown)


TH3CHARLie:
Please note that since commit the master has been failing.

PR #9098, opened 2020-07-05
[mypyc] Various documentation updates

JukkaL:
Includes many changes (this list is not exhaustive):

* Document how to pass mypy options.
* Document some additional primitives.
* Protocols are now supported.
* General wordsmithing.

### DIFF
--- a/mypyc/doc/differences_from_python.rst
+++ b/mypyc/doc/differences_from_python.rst
@@ -262,12 +262,6 @@ Descriptors
 Native classes can't contain arbitrary descriptors. Properties, static
 methods and class methods are supported.
 
-Defining protocols
-******************
-
-Protocols can't be defined in compiled code. You can use protocols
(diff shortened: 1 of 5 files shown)


mypyc/doc/differences_from_python.rst
```diff
@@ -262,12 +262,6 @@ Descriptors
 Native classes can't contain arbitrary descriptors. Properties, static
 methods and class methods are supported.
 
-Defining protocols
-******************
-
-Protocols can't be defined in compiled code. You can use protocols
-defined elsewhere, however.
-
```
TH3CHARLie, line 270: Does this mean protocols can be used now?

JukkaL: Yeah, they should work now. Operations on them will fall back to generic operations, though.

PR #9100, opened 2020-07-06
Sync typeshed

JukkaL:


JelleZijlstra:
Looks like the multiprocessing changes from python/typeshed#4282 or python/typeshed#4289 are causing some issues.

JukkaL:
Yeah, `multiprocessing.Queue` is no longer a type in the stubs. At runtime it's not type, but it's documented as a class, which makes things confusing: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Queue

Also, `multiprocessing.queues` seems to be undocumented, so for a user to annotate code using queues, they need to either read the stubs or the stdlib implementation, which is unfortunate.

Finally, the return type of `multiprocessing.Queue` is `Queue[Any]`, so the change loses some typing precision. Since `multiprocessing.queues.Queue` is not very discoverable (as it's undocumented), for many users this may effectively remove the ability to type check queue items.

The above considerations probably explain why `multiprocessing.Queue` used to be defined as a class, even thought it's not quite true.

JukkaL:
Perhaps more worryingly, the change breaks existing code that uses `multiprocessing.Queue` as a type. I'll create a typeshed issue about this.

JukkaL:
Let's move the discussion to https://github.com/python/typeshed/issues/4313.

PR #9102, opened 2020-07-06
Adds get_function_signature_hook

### LINKED ISSUE #9101: New plugin hook: get_function_signature_hook

I am proposing a new feature: `get_function_signature_hook`
It works the same way as `get_method_signature_hook`, but for functions.

I needed this while working on `HKT` for `dry-python/returns`: https://github.com/dry-python/returns/pull/446

sobolevn:
This PR introduces `get_function_signature_hook` that behaves the similar way as `get_method_signature_hook`.

Closes #9101

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -686,7 +686,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             self, callee: FunctionLike, args: List[Expression],
             arg_kinds: List[int], context: Context,
             arg_names: Optional[Sequence[Optional[str]]],
-            signature_hook: Callable[[MethodSigContext], CallableType]) -> FunctionLike:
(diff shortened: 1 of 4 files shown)


mypy/checkexpr.py
```diff
@@ -675,6 +679,34 @@ def try_infer_partial_value_type_from_call(
 
         return None
 
+    def apply_function_signature_hook(
```
msullivan, line 682: Would it be possible to merge this with apply_method_signature_hook? Maybe with a joint helper that takes formal args, callee, context, and the checker, and then pass different versions of that to the joint helper?

msullivan (commented):
This looks good, one request.

mypy/checkexpr.py
msullivan: Actually I just did this myself

PR #9107, opened 2020-07-07
mypyc: ignore deprecation

hauntsaninja:
PyUnicode_AsUnicodeAndSize has been deprecated since 3.3

Python 3.9 has compiler warnings for this deprecated function, and we
compile with Werror, causing Python 3.9 builds to fail.

I've just copied over the relevant deprecation ignoring code from the
original getargs.c (including the TODO, but I can remove that)

### DIFF
--- a/mypyc/lib-rt/getargs.c
+++ b/mypyc/lib-rt/getargs.c
@@ -18,6 +18,29 @@
  *    and is responsible for decrefing them.
  */
 
+// These macro definitions are copied from pyport.h in Python 3.9 and later
+// https://bugs.python.org/issue19569
+#if defined(__clang__)
+#define _Py_COMP_DIAG_PUSH _Pragma("clang diagnostic push")
+#define _Py_COMP_DIAG_IGNORE_DEPR_DECLS \
+    _Pragma("clang diagnostic ignored \"-Wdeprecated-declarations\"")
+#define _Py_COMP_DIAG_POP _Pragma("clang diagnostic pop")
+#elif defined(__GNUC__) \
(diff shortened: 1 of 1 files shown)


TH3CHARLie:
I think this would solve mypyc/mypyc#738

hauntsaninja:
Thanks for pointing that out! It would definitely solve mr-c's problem. We're still using the deprecated APIs, so not sure whether we should close the issue (but we can probably remove priority-0-high). 
I figured the most maintainable thing to do was just copy over code from CPython master, and we'll fix our usage of deprecated APIs whenever CPython's is.

TH3CHARLie:
> Thanks for pointing that out! It would definitely solve mr-c's problem. We're still using the deprecated APIs, so not sure whether we should close the issue (but we can probably remove priority-0-high).
> I figured the most maintainable thing to do was just copy over code from CPython master, and we'll fix our usage of deprecated APIs whenever CPython's is.

Good point, last time I check the CPython documentation about it, it recommends some APIs to replace the deprecated ones(I recall this would create longer code since things are handled more explicitly), maybe that would worth a try as well.

JelleZijlstra:
Not sure if you guys saw, but there's currently active discussions in CPython around removing Py_UNICODE-related stuff. This is in PEP 623 and in some mailing list threads. I haven't read in too much detail, but it sounds like tools like mypyc shouldn't be using any Py_UNICODE-like APIs, and they're likely to be removed in 3.10 or 3.11.

Silencing the deprecation warning for now might still be the best short-term solution though.

mr-c:
I think this is ready for review & merge :-)

PR #9108, opened 2020-07-07
[mypyc] Add BinaryIntOp for low-level integer operations

TH3CHARLie:
relates mypyc/mypyc#741

This PR introduces `BinaryIntOp` to represent all low-level integer binary operations.

The block-like logic described in mypyc/mypyc#743 would be handled differently, `BinaryIntOp` would be the building block of it.

### DIFF
--- a/mypyc/ir/ops.py
+++ b/mypyc/ir/ops.py
@@ -1252,6 +1252,43 @@ class LoadGlobal(RegisterOp):
         return visitor.visit_load_global(self)
 
 
+class BinaryIntOp(RegisterOp):
+    """Binary operations on integer types
+
+    These ops are low-level and will be eventually generated to simple x op y form.
+    The left and right values should be of low-level integer types that support those ops
+    """
+    error_kind = ERR_NEVER
+
+    ADD = 0  # type: Final
+    SUB = 1  # type: Final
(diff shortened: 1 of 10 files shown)


TH3CHARLie:
One improvement here is to support immediate value(without using registers), so `BinaryIntOp` would need to not inherited from `RegisterOp`.

mypyc/ir/ops.py
```diff
@@ -1252,6 +1252,70 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
...
+        LEQ: '<=',
+        GEQ: '>=',
+        AND: '&',
+        OR: '|',
+        XOR: '^',
+        LEFT_SHIFT: '<<',
+        RIGHT_SHIFT: '>>',
+    }  # type: Final
+
+    def __init__(self, type: RType, lhs: Value, rhs: Value, op: int, line: int = -1) -> None:
```
JukkaL, line 1302: It would be good to store the integer type also here, since different code may be generated for 32 bit and 64 bit adds, for example, on non-C back ends.

JukkaL: Actually, it's better to do this in a separate PR, and also update things to more generally store the result type.

PR #9110, opened 2020-07-08
[mypyc] Merge more primitive ops

TH3CHARLie:
relates mypyc/mypyc#734. This PR completes **ALL** ops of dict, str, list, tuple, set that are supported using current design. The remaining ones would rather need to split into multiple ops(via specializers) or using pointers.


TH3CHARLie:
oops, something broke mypyc. Had to do a binary search to find out....

TH3CHARLie:
Turn out this is much more complicated than I assumed...

Here's what I find:

using mypyc to compile mypy itself, the first assertion fail would happen during codegen the function: `SymbolTable.__str__` in `mypy/nodes.py`:
```python
    def __str__(self) -> str:
        a = []  # type: List[str]
        for key, value in self.items():
            # Filter out the implicit import of builtins.
            if isinstance(value, SymbolTableNode):
                if (value.fullname != 'builtins' and
                        (value.fullname or '').split('.')[-1] not in
                        implicit_module_attrs):
                    a.append('  ' + str(key) + ' : ' + str(value))
            else:
                a.append('  <invalid item>')
        a = sorted(a)
        a.insert(0, 'SymbolTable(')
        a[-1] += ')'
        return '\n'.join(a)
```

The ` a.append('  <invalid item>')` method call will have a result_type of `object_rprimitive` if you checked the generated IR, so it will coerce and generate box op following the CallC in IR. However, if you write python code like this:
```python
from typing import List
def f(x: List[str], y: str) -> List[str]:
    a = []  # type: List[str]
    a.append('  <invalid item>')
    return a
```
There would be a result_type of `none_rprimitive` and therefore no box and everything works as expected.

So what causes such differences? It seems that the `builder.node_type` fail to find the method call's type in builder's types in the first case, therefore leading to an `object_rprimitive` as the expression result type. However, according to python's semantics, `foo.append(x)` should return none.
https://github.com/python/mypy/blob/ffdbeb3d47ccb2d9691b36993be0a18e0718d997/mypyc/irbuild/expression.py#L234-L240

the builder's types come from the mypy check, I am not sure why these two code snippets have different types on this expression. @JukkaL

JukkaL:
I suspect that `a.append('  <invalid item>')` is considered unreachable by mypy, and it doesn't type check unreachable code, so types will be incomplete. You can verify that by adding something that should generate an error to it (e.g. `1 + 'x'`). If there's no error, the code is treated as unreachable.

A reasonable way forward would be to ensure that the coercion to `object` works. I'm not sure why we are triggering an assertion, but it may be that we need to add an extra coercion somewhere, or that the coercion doesn't work for short integer types. Each of these should not be hard to fix.

Here's a possible simplified test case (didn't verify that it results in the same issue):

```py
def f(x: List[str], y: str) -> None:
    if isinstance(y, int):  # Since int and str are distinct types, mypy things that this will always be false
        x.append('x')
```

If you are blocked on this, you can start working on some other issue while we investigate this, such as inlined integer operations.

TH3CHARLie:
> I suspect that `a.append(' <invalid item>')` is considered unreachable by mypy, and it doesn't type check unreachable code, so types will be incomplete. You can verify that by adding something that should generate an error to it (e.g. `1 + 'x'`). If there's no error, the code is treated as unreachable.
> 
> A reasonable way forward would be to ensure that the coercion to `object` works. I'm not sure why we are triggering an assertion, but it may be that we need to add an extra coercion somewhere, or that the coercion doesn't work for short integer types. Each of these should not be hard to fix.
> 
> Here's a possible simplified test case (didn't verify that it results in the same issue):
> 
> ```python
> def f(x: List[str], y: str) -> None:
>     if isinstance(y, int):  # Since int and str are distinct types, mypy things that this will always be false
>         x.append('x')
> ```
> 
> If you are blocked on this, you can start working on some other issue while we investigate this, such as inlined integer operations.

Yes exactly it is caused by mypy's report on unreachable code. I haven't checked the detail but in a `Dict[str, SymbolTableNode]` with `if isinstance(value, SymbolTableNode)`, it makes sense mypy thinks the else branch is unreachable(I wonder if it gets called at all). I add a hack to box int32/int64, since they are not supposed to be boxed and the boxed result will not be used anywhere else, I used a NULL to achieve that.

mypyc/codegen/emit.py
```diff
@@ -695,6 +695,11 @@ def emit_box(self, src: str, dest: str, typ: RType, declare_dest: bool = False,
             self.emit_lines('{}{} = Py_None;'.format(declaration, dest))
             if not can_borrow:
                 self.emit_inc_ref(dest, object_rprimitive)
+        # TODO: This is a hack to handle mypy's false negative on unreachable code.
+        #       All ops returning int32/int64 should not be coerced into object.
+        #       Since their result will not be used elsewhere, it's safe to use NULL here
+        elif is_int32_rprimitive(typ) or is_int64_rprimitive(typ):
+            self.emit_lines('{}{} = NULL;'.format(declaration, dest))
```
TH3CHARLie, line 702: here's the hack

JukkaL: This is fine for now, but can you create a follow-up issue to properly box these as Python `int` objects? It should be easy enough to support, and it could be useful in the future.

PR #9116, opened 2020-07-09
[mypyc] low-level integer operations: integer equal

TH3CHARLie:
relates mypyc/mypyc#741, mypyc/mypyc#743

### DIFF
--- a/mypyc/irbuild/builder.py
+++ b/mypyc/irbuild/builder.py
@@ -35,12 +35,11 @@ from mypyc.ir.ops import (
     AssignmentTargetAttr, AssignmentTargetTuple, Environment, LoadInt, Value,
     Register, Op, Assign, Branch, Unreachable, TupleGet, GetAttr, SetAttr, LoadStatic,
     InitStatic, PrimitiveOp, OpDescription, NAMESPACE_MODULE, RaiseStandardError,
-    BinaryIntOp
 )
 from mypyc.ir.rtypes import (
     RType, RTuple, RInstance, int_rprimitive, dict_rprimitive,
     none_rprimitive, is_none_rprimitive, object_rprimitive, is_object_rprimitive,
-    str_rprimitive, c_pyssize_t_rprimitive, bool_rprimitive
+    str_rprimitive,
 )
 from mypyc.ir.func_ir import FuncIR, INVALID_FUNC_DEF
 from mypyc.ir.class_ir import ClassIR, NonExtClassInfo
@@ -48,7 +47,6 @@ from mypyc.primitives.registry import func_ops, CFunctionDescription, c_function
 from mypyc.primitives.list_ops import list_len_op, to_list, list_pop_last
 from mypyc.primitives.dict_ops import dict_get_item_op, dict_set_item_op
 from mypyc.primitives.generic_ops import py_setattr_op, iter_op, next_op
-from mypyc.primitives.int_ops import int_logical_op_mapping
 from mypyc.primitives.misc_ops import true_op, false_op, import_op
 from mypyc.crash import catch_errors
 from mypyc.options import CompilerOptions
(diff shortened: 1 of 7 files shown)


mypyc/irbuild/expression.py
```diff
@@ -382,6 +382,8 @@ def transform_basic_comparison(builder: IRBuilder,
                                left: Value,
                                right: Value,
                                line: int) -> Value:
+    if is_int_rprimitive(left.type) and is_int_rprimitive(right.type) and op in ('=='):
```
JukkaL, line 385: The latter part should be `op == '=='` or `op in ('==',)`, as now it's equivalent to `op in '=='`.

TH3CHARLie:
```python
def foo(x: int, y: int) -> bool:
     return x == y
```

before:
```C
char CPyDef_foo(CPyTagged cpy_r_x, CPyTagged cpy_r_y) {
    char cpy_r_r0;
CPyL0: ;
    cpy_r_r0 = CPyTagged_IsEq(cpy_r_x, cpy_r_y);
    return cpy_r_r0;
}
```

after:
```c
char CPyDef_foo(CPyTagged cpy_r_x, CPyTagged cpy_r_y) {
    char cpy_r_r0;
    int64_t cpy_r_r1;
    int64_t cpy_r_r2;
    int64_t cpy_r_r3;
    char cpy_r_r4;
    char cpy_r_r5;
    char cpy_r_r6;
CPyL0: ;
    cpy_r_r1 = 1;
    cpy_r_r2 = cpy_r_x & cpy_r_r1;
    cpy_r_r3 = 0;
    cpy_r_r4 = cpy_r_r2 == cpy_r_r3;
    if (cpy_r_r4) {
        goto CPyL1;
    } else
        goto CPyL2;
CPyL1: ;
    cpy_r_r5 = cpy_r_x == cpy_r_y;
    cpy_r_r0 = cpy_r_r5;
    goto CPyL3;
CPyL2: ;
    cpy_r_r6 = CPyTagged_IsEq_(cpy_r_x, cpy_r_y);
    cpy_r_r0 = cpy_r_r6;
    goto CPyL3;
CPyL3: ;
    return cpy_r_r0;
}
```

JukkaL:
The code looks good to me. It's kind of verbose, but that's to be expected.

If we implement the optimizations discussed in mypyc/mypyc#749, the generated code would look like this:

```c
char CPyDef_foo(CPyTagged cpy_r_x, CPyTagged cpy_r_y) {
    char cpy_r_r0;
CPyL0: ;
    if (cpy_r_x & 1 == 0) {
        goto CPyL1;
    } else
        goto CPyL2;
CPyL1: ;
    cpy_r_r0 = cpy_r_x == cpy_r_y;
    goto CPyL3;
CPyL2: ;
    cpy_r_r0 = CPyTagged_IsEq_(cpy_r_x, cpy_r_y);
    goto CPyL3;
CPyL3: ;
    return cpy_r_r0;
}
```

If we'd also simplify redundant gotos and labels, it could look like this:

```c
char CPyDef_foo(CPyTagged cpy_r_x, CPyTagged cpy_r_y) {
    char cpy_r_r0;
    if (cpy_r_x & 1 != 0)
        goto CPyL2;
    cpy_r_r0 = cpy_r_x == cpy_r_y;
    goto CPyL3;
CPyL2: ;
    cpy_r_r0 = CPyTagged_IsEq_(cpy_r_x, cpy_r_y);
CPyL3: ;
    return cpy_r_r0;
```

For now, the verbose generated code is fine. We can look at simplifying it in later PRs.

TH3CHARLie:
> 
> If we implement the optimizations discussed in [mypyc/mypyc#749](https://github.com/mypyc/mypyc/issues/749)

Yes this would be an interesting next step.

I've updated this PR with some dispatch logic to abstract the verbose logic and make the generalization of more ops easier.

mypyc/irbuild/builder.py
```diff
@@ -240,6 +242,31 @@ def environment(self) -> Environment:
...
     # Helpers for IR building
+    def check_tagged_short_int(self, val: Value, line: int) -> Value:
+        int_tag = self.add(LoadInt(1, line, rtype=c_pyssize_t_rprimitive))
+        bitwise_and = self.binary_int_op(c_pyssize_t_rprimitive, val,
+                                         int_tag, BinaryIntOp.AND, line)
+        zero = self.add(LoadInt(0, line, rtype=c_pyssize_t_rprimitive))
+        check = self.binary_int_op(bool_rprimitive, bitwise_and, zero, BinaryIntOp.EQ, line)
+        return check
+
+    def binary_comparison_op(self, lhs: Value, rhs: Value, op: str, line: int) -> Value:
```
JukkaL, line 253: Maybe rename this to `compare_tagged`? This is comparing tagged integers, specifically. Also I think that this should live in `ll_builder.py`, since this doesn't seem to depend on any high-level things such as AST nodes.

Add a docstring.

mypyc/primitives/int_ops.py
```diff
@@ -129,6 +130,12 @@ def int_compare_op(name: str, c_function_name: str) -> None:
     format_str='{dest} = {args[0]} + {args[1]} :: short_int',
     emit=simple_emit('{dest} = {args[0]} + {args[1]};'))
 
+int_equal = c_custom_op(
```
JukkaL, line 133: Add comment. Maybe rename this to something like `int_equal_` to highlight that this is just a part of an implementation.

mypyc/primitives/int_ops.py
```diff
@@ -139,3 +146,7 @@ def int_unary_op(name: str, c_function_name: str) -> CFunctionDescription:
 
 
 int_neg_op = int_unary_op('-', 'CPyTagged_Negate')
+
+int_logical_op_mapping = {
+    '==': (BinaryIntOp.EQ, int_equal)
+}  # type: Dict[str, Tuple[int, CFunctionDescription]]
```
JukkaL, line 152: Add comment saying that these are not complete implementations, but assume that one of operands is a long integer.

TH3CHARLie: Actually there's no such assumption I think, the first item of the tuple provides op variant for the case when lhs of the operands is short, therefore not assume at least one of them is long.

JukkaL: Ah yeah, you are correct.

PR #9121, opened 2020-07-09
[mypyc] Merge exception ops

TH3CHARLie:
relates mypyc/mypyc#734

Should we also consider building exceptions using the approach we treat int binary ops now? @JukkaL


JukkaL:
> Should we also consider building exceptions using the approach we treat int binary ops now?

Can you give an example of this? We can discuss this offline.

PR #9123, opened 2020-07-10
Add keyword arguments for Functional Enum API

### LINKED ISSUE #9079: Bug: Functional Enum API Unexpected arguments to Enum()
I get error 
```
enum-repro.py:3: error: Unexpected arguments to Enum()
```

When mypy checking for example 
```python
from enum import Enum

enum_name = Enum(
    value="enum_name",
    names={
        "a": 1,
        "b": 2
    }
)
print(enum_name)
```

I'm using mypy 0.782 and Python 3.8.3. Also tried latest master (0.790+dev.eae1860bef0a6fe06753459bf97633a41e789ed7)

Since this is valid python3 syntax, I would expect mypy to at least let it through, maybe with a warning about dynamic types being impossible to infer at static typecheck time.

For my actual use case, I could try using the class EnumName(Enum)): form but It doesn't fully support my use case, because I would like to use names containing spaces and special symbols.

LiuYuHui:
fix #9079

### DIFF
--- a/mypy/semanal_enum.py
+++ b/mypy/semanal_enum.py
@@ -8,7 +8,7 @@ from typing import List, Tuple, Optional, Union, cast
 from mypy.nodes import (
     Expression, Context, TypeInfo, AssignmentStmt, NameExpr, CallExpr, RefExpr, StrExpr,
     UnicodeExpr, TupleExpr, ListExpr, DictExpr, Var, SymbolTableNode, MDEF, ARG_POS,
-    EnumCallExpr, MemberExpr
+    ARG_NAMED, EnumCallExpr, MemberExpr
 )
 from mypy.semanal_shared import SemanticAnalyzerInterface
 from mypy.options import Options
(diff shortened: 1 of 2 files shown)


mypy/semanal_enum.py
```diff
@@ -104,23 +104,37 @@ def parse_enum_call_args(self, call: CallExpr,
...
+        valid_name = [None, 'value', 'names', 'module', 'qualname', 'type', 'start']
+        for arg_name in call.arg_names:
+            if arg_name not in valid_name:
+                self.fail_enum_call_arg("Unexpected keyword argument '{}'".format(arg_name), call)
+        value, names = None, None
+        for arg_name, arg in zip(call.arg_names, args):
+            if arg_name == 'value':
+                value = arg
+            if arg_name == 'names':
+                names = arg
```
JukkaL, line 122: These additional improvements would be nice: 

* Make sure that only at most the first two arguments are given as positional arguments (the others are keyword-only).
* If some argument is given as a keyword argument, ensure that they aren't also given as positional arguments.

Issue #9054, opened 2020-06-26
INTERNAL ERROR when using `:=` assignment expression


Apakottur:
To reproduce the bug, run `mypy` with the following command:
```bash
mypy --show-traceback test.py
```

Where the contents of `test.py` are:

```python
def func() -> None:
    obj = MyClass()
    if result := obj.var_name:
        pass


class MyClass:
    def __init__(self):
        self.var_name = "abc"
```

Here is the traceback that I get:

```
test.py:3: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 0.790+dev.6ee562a8f3e69ac134f8c501b79b3712c9e24bbd
Traceback (most recent call last):
  File "***/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "***/python3.8/site-packages/mypy/__main__.py", line 8, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "***/python3.8/site-packages/mypy/main.py", line 89, in main
    res = build.build(sources, options, None, flush_errors, fscache, stdout, stderr)
  File "***/python3.8/site-packages/mypy/build.py", line 180, in build
    result = _build(
  File "***/python3.8/site-packages/mypy/build.py", line 252, in _build
    graph = dispatch(sources, manager, stdout)
  File "***/python3.8/site-packages/mypy/build.py", line 2626, in dispatch
    process_graph(graph, manager)
  File "***/python3.8/site-packages/mypy/build.py", line 2949, in process_graph
    process_stale_scc(graph, scc, manager)
  File "***/python3.8/site-packages/mypy/build.py", line 3047, in process_stale_scc
    graph[id].type_check_first_pass()
  File "***/python3.8/site-packages/mypy/build.py", line 2107, in type_check_first_pass
    self.type_checker().check_first_pass()
  File "***/python3.8/site-packages/mypy/checker.py", line 294, in check_first_pass
    self.accept(d)
  File "***/python3.8/site-packages/mypy/checker.py", line 401, in accept
    stmt.accept(self)
  File "***/python3.8/site-packages/mypy/nodes.py", line 677, in accept
    return visitor.visit_func_def(self)
  File "***/python3.8/site-packages/mypy/checker.py", line 726, in visit_func_def
    self._visit_func_def(defn)
[...]

Issue #8975, opened 2020-06-09
Add support for enabling/disabling individual error codes


ofek:
From https://github.com/python/mypy/issues/7511#issuecomment-640665509

This would also entail giving a specific error code to most error messages

JukkaL:
Mypy can already produce error codes for many of the most common errors (`--show-error-codes`). We'd probably want to add some more error codes for messages that would be good candidates for disabling (or enabling), however.

#7599 is very closely related, but it's a slightly different use case, so let's keep both issues open.

Open questions (we only need to resolve the first one before implementation work can start):
1. How would we configure error codes (command line flag / config file option)?
2. Should it be possible to set these on a per module basis? (Maybe initially these could be global flags, for simplicity.)
3. Which additional error codes would be useful to add?
4. Should some of the existing strictness flags be retrofitted to use this approach?
5. Should we also display error codes by default, as otherwise the feature may be confusing to use?

Brainstorming for (1): We could call the options `--enable-error-code code` and `--disable-error-code code` (these can be repeated, similar to `--always-true`). In the config file this could be `disable_error_code = code1, code2, ...`, etc.

b-thebest:
Can I start working on this?

ofek:
@b-thebest That would be awesome, yes please!

jsar3004:
I am new to open source. I want to work on this issue. Please tell how can I fix this issue or what technologies should I learn to fix this issue.

JukkaL:
@jsar3004 To get started, I'd recommend looking at https://github.com/python/mypy/blob/master/README.md and the linked documents.

cph-w:
Hi @JukkaL, I took a first crack at this in the linked MR. Please let me know if you have any feedback or additional work to be completed for this based on that implementation

m-vdb:
@JukkaL hi! is it possible to have a `mypy` release on pypi that includes this change? I'd love to use it 😄 

And thanks a lot @cph-w for making it 💯

hauntsaninja:
See #9290

PR #9125, opened 2020-07-10
Distinguish redundant-expr warnings from unreachable warnings

### LINKED ISSUE #8711: Unreachable with always true/false operands to and/or
It appears that an always true/false operand into and/or is marked as unreachable, even when it doesnt cause the other operand to be short circuited away (that scenario is correct, and emits a different "is never evaluated" message).  The examples below are obviously very contrived; a more real-world example is something like

```python
assert isinstance(x, float) and x > 0.0
```
---

* Are you reporting a bug, or opening a feature request? bug
* Please insert below the code you are checking with mypy.
    ```python
    True and True
    ```
    ```
    main.py: note: In function "foo":
    main.py:1:1: error: Left operand of 'and' is always true  [unreachable]
            True and True
            ^
    ```
* What is the behavior/output you expect?  Success
* What are the versions of mypy and Python you are using? master (0.770+dev.833fc2caec6773683d691a72058293ae87139210) with CPython 3.7.7
* What are the mypy flags you are using?  Custom, but effectively `--strict`

llchan:
There are redundant expression warnings such as "Left operand of 'or' is always false" and "Left operand of 'and' is always true" that are previously categorized as `unreachable`.  This is a bit of a miscategorization because in these cases all parts of the expression are still evaluated at runtime, compared to situations when the opposite condition is true and the right operand is short-circuited away (which is a stronger hint that something is wrong).

This PR introduces a new `redundant-expr` code and splits the redudant-but-not-short-circuiting cases out from the `unreachable` code.

Fixes #8711


llchan:
Note to reviewer: I'm not familiar with the `mypyc`/`dmypy` side of things, so if any corresponding changes need to happen there let me know.

JukkaL:
Thanks for the PR!

Since we already have lots of command line flags, are okay with waiting until #8975 has been implemented (#9172 is the PR) and then making the new error code disabled by default, instead of using a command-line flag? Our longer-term plan is to remove several of the existing command-line flags and to use error codes that can be individually disabled and enabled instead.

llchan:
Sure, can wait on it. I'll subscribe to that PR, but if I overlook the notification feel free to ping me here.

llchan:
@JukkaL I've rebased off master and updated this to use the new `--enable-error-code` functionality

test-data/unit/check-redundant-expr.test
```diff
@@ -0,0 +1,17 @@
+-- Type checker test cases for conditional checks that result in some
+-- expressions classified as redundant.
+
+[case testRedundantExpressions]
+# flags: --enable-error-code redundant-expr
```
JukkaL, line 5: Also use `--show-error-codes`.

test-data/unit/check-redundant-expr.test
```diff
@@ -0,0 +1,17 @@
+-- Type checker test cases for conditional checks that result in some
```
JukkaL, line 1: Can you move the contents of this file to `check-errorcodes.test`, since it looks like the logical place for this?

mypy/errorcodes.py
```diff
@@ -112,6 +112,8 @@ def __str__(self) -> str:
     'General')  # type: Final
 UNREACHABLE = ErrorCode(
     'unreachable', "Warn about unreachable statements or expressions", 'General')  # type: Final
+REDUNDANT_EXPR = ErrorCode(
+    'redundant-expr', "Warn about redundant expressions", 'General')  # type: Final
```
JukkaL, line 116: Add `default_enabled=False` when constructing `ErrorCode` object.

llchan:
Done

llchan:
@JukkaL + maintainers: anything else you need from me here?

gvanrossum:
Mostly time to get to this. I’m not likely the best reviewer myself.--
--Guido (mobile)

gvanrossum (commented):
LGTM, two minor nits.

IIRC Jukka asked for documentation; I don’t see any documentation diffs in your PR. How/where are the error codes documented?

gvanrossum:
(OT) Hey @llchan, would you be interested in taking a crack at https://github.com/python/mypy/issues/9054 ?

mypy/checkexpr.py
```diff
@@ -2740,6 +2740,17 @@ def check_boolean_op(self, e: OpExpr, context: Context) -> Type:
             restricted_left_type = true_only(left_type)
             result_is_left = not left_type.can_be_false
 
+        # If left_map is None then we know mypy considers the left expression
+        # to be reundant.
+        #
+        # Note that we perform these checks *before* we take into account
+        # the analysis from the semanal phase below. We assume that nodes
+        # marked as unreachable during semantic analysis were done so intentionally.
+        # So, we shouldn't report an error.
```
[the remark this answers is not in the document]
llchan: I don't know what exactly that comment means.  When I split the conditions into two, I erred on the side of "explicit is better than implicit" and made sure the comment was copied to both places, so that a future change doesnt inadvertently lead to a dangling comment reference.

gvanrossum:
Okay, let's leave that comment as is then. But what about documentation?

llchan:
> (OT) Hey @llchan, would you be interested in taking a crack at #9054 ?

Is that related to this PR somehow, or are you asking for general help?  To be quite honest I'm not sure I'm familiar enough with the mypy internals to fix that in a time-efficient way (this PR was a little more straightforward since I was taking an existing error code and splitting it into two).  I can experiment with it, since I'm sure we will hit that at some point now that we're using 3.8, but it may be better to assign it to an expert.

llchan:
Documentation coming in the next push, working on it now

gvanrossum:
> > (OT) Hey @llchan, would you be interested in taking a crack at #9054 ?
> 
> Is that related to this PR somehow, or are you asking for general help? To be quite honest I'm not sure I'm familiar enough with the mypy internals to fix that in a time-efficient way (this PR was a little more straightforward since I was taking an existing error code and splitting it into two). I can experiment with it, since I'm sure we will hit that at some point now that we're using 3.8, but it may be better to assign it to an expert.

It's only related to this PR because the problem is (likely) in the same area of mypy, and I believe you have shown that you know your way around that part. But if you feel that you're not up to it that's totally fine!

gvanrossum (approved):
LGTM. Thanks!

PR #9127, opened 2020-07-11
[mypyc] Generates inline compare statement on short ints and fix sieve performance regression

TH3CHARLie:
relates mypyc/mypyc#750

Generate almost identical code as before https://github.com/python/mypy/commit/be01236bcdb9a9da66e68dd0d45ff0f9a604e44a

before:
```c
CPyL1: ;
    cpy_r_r4 = (Py_ssize_t)cpy_r_r3 < (Py_ssize_t)cpy_r_r2;
    if (cpy_r_r4) {
        goto CPyL2;
    } else
        goto CPyL5;
```

now with this PR:
```c
CPyL1: ;
    cpy_r_r4 = cpy_r_r3 < cpy_r_r2;
    if (cpy_r_r4) {
        goto CPyL2;
    } else
        goto CPyL5;
```

### DIFF
--- a/mypyc/primitives/int_ops.py
+++ b/mypyc/primitives/int_ops.py
@@ -150,8 +150,15 @@ int_equal_ = c_custom_op(
     c_function_name='CPyTagged_IsEq_',
     error_kind=ERR_NEVER)
 
+int_less_than_ = c_custom_op(
+    arg_types=[int_rprimitive, int_rprimitive],
+    return_type=bool_rprimitive,
+    c_function_name='CPyTagged_IsLt',
+    error_kind=ERR_NEVER)
+
 # provide mapping from textual op to short int's op variant and boxed int's description
 # note these are not complete implementations
 int_logical_op_mapping = {
-    '==': (BinaryIntOp.EQ, int_equal_)
+    '==': (BinaryIntOp.EQ, int_equal_),
+    '<': (BinaryIntOp.LT, int_less_than_)
 }  # type: Dict[str, Tuple[int, CFunctionDescription]]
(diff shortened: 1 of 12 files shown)


TH3CHARLie:
IR test changes are a bit of messy because the integer < got expanded. To see the change in list for, check irbuild-list and irbuild-statement.

This should also fix the int_lists regression as well.

mypyc/primitives/int_ops.py
```diff
@@ -150,8 +150,15 @@ def int_unary_op(name: str, c_function_name: str) -> CFunctionDescription:
     c_function_name='CPyTagged_IsEq_',
     error_kind=ERR_NEVER)
 
+int_less_than_ = c_custom_op(
+    arg_types=[int_rprimitive, int_rprimitive],
+    return_type=bool_rprimitive,
+    c_function_name='CPyTagged_IsLt',
```
JukkaL, line 156: Should this have a `_` suffix?

mypyc/test/test_exceptions.py
```diff
@@ -32,7 +32,9 @@ def run_case(self, testcase: DataDrivenTestCase) -> None:
         """Perform a runtime checking transformation test case."""
         with use_custom_builtins(os.path.join(self.data_prefix, ICODE_GEN_BUILTINS), testcase):
             expected_output = remove_comment_lines(testcase.output)
-
+            # replace native_int with platform specific ints
+            int_format_str = 'int32' if IS_32_BIT_PLATFORM else 'int64'
+            expected_output = [s.replace('native_int', int_format_str) for s in expected_output]
```
JukkaL, line 37: Since we do this same dance in multiple test files, maybe you can move this to a helper function? This way if we modify this, we only have to change it in one place.

mypyc/primitives/int_ops.py
TH3CHARLie: It should. It's like the `int_equal_` several lines above it, which provides a CallC description for a logical op but not complete and should always be used within ll_builder's function.

TH3CHARLie: yeah, it definitely should! Thanks for catching this potentially costly bug!

PR #9131, opened 2020-07-11
[mypyc] Use C optimization level 0 for tests by default

JukkaL:
This speeds up run tests significantly, but it might miss some checks
only performed on higher optimization levels.

On my Linux desktop, this speeds up `pytest mypyc` from about 18s to
about 12s. It should also speed up CI builds, but I haven't measured
the impact.

Let's try this out. We can always revert this back if this turns out
to cause problems.

The environment variable MYPYC_OPT_LEVEL can be used to override
the optimization level. Example:

    MYPYC_OPT_LEVEL=3 pytest mypyc

Closes mypyc/mypyc#745.

### DIFF
--- a/mypyc/test/test_run.py
+++ b/mypyc/test/test_run.py
@@ -56,7 +56,7 @@ from mypyc.build import mypycify
 
 setup(name='test_run_output',
       ext_modules=mypycify({}, separate={}, skip_cgen_input={!r}, strip_asserts=False,
-                           multi_file={}),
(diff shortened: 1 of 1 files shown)


TH3CHARLie:
Maybe I should finish #9030 quickly... so we can make opt_level a -O{0-3} option

JukkaL:
There's no hurry, since we'll need a different way to pass options to pytest anyway.

PR #9132, opened 2020-07-11
[mypyc] Split CPy.h into multiple C files

JukkaL:
Move functions from CPy.h to C files corresponding to various modules that define 
primitives. For example, the new file `int_ops.c` corresponds to 
`mypyc.primitives.int_ops`. Also group related things together in CPy.h

The motivation is to clean things up by not having function implementations in a 
header (for non-inline functions), and to improve the organization of the code. CPy.h 
used to have a lot of only loosely related functionality.

As a side effect, many of the functions are no longer declared as static. I left all inline 
functions in CPy.h to avoid regressing performance.

Closes mypyc/mypyc#733.


JukkaL:
I'm not going to wait for a code review, since this PR will get stale quickly.

PR #9143, opened 2020-07-13
Clean up terminal width handling

The-Compiler:
This PR does three related cleanups (cc'ing @ilevkivskyi who seems to have implemented most of the underlying code):

---

Use `shutil.get_terminal_size()`

The [documentation](https://docs.python.org/3/library/os.html#os.get_terminal_size) for Python's `os.get_terminal_size()` says:

> `shutil.get_terminal_size()` is the high-level function which should normally be used, `os.get_terminal_size` is the low-level implementation.

`shutil.get_terminal_size()` already [takes care of various corner cases](https://github.com/python/cpython/blob/v3.8.3/Lib/shutil.py#L1298-L1341) such as:

- Providing a fallback (to 80 columns, like the current `DEFAULT_COLUMNS`)
- Returning that fallback if `os.get_terminal_size()` raises `OSError`
- Doing the same if it returns a size of 0

In addition to that:

- It takes care of some more corner cases ("stdout is None, closed, detached, or not a terminal, or `os.get_terminal_size()` is unsupported")
- It adds support for the `COLUMNS` environment variable (rather than the non-standard `MYPY_FORCE_TERMINAL_WIDTH`)

---

Move `MYPY_FORCE_TERMINAL_WIDTH` handling into `get_terminal_width()`
    
Those two callers are the only two, and both check `MYPY_FORCE_TERMINAL_WIDTH` before calling `get_terminal_width()`.

---

Remove `fixed_terminal_width` from `utils.colorize` docstring
    
This was added by 91adf615f98b110089bff842aafcdd7eb657093f, yet `colorize()` doesn't actually have a `fixed_terminal_width` argument.

### DIFF
--- a/mypy/dmypy/client.py
+++ b/mypy/dmypy/client.py
@@ -475,8 +475,7 @@ def request(status_file: str, command: str, *, timeout: Optional[int] = None,
     # Tell the server whether this request was initiated from a human-facing terminal,
     # so that it can format the type checking output accordingly.
     args['is_tty'] = sys.stdout.isatty() or int(os.getenv('MYPY_FORCE_COLOR', '0')) > 0
-    args['terminal_width'] = (int(os.getenv('MYPY_FORCE_TERMINAL_WIDTH', '0')) or
-                              get_terminal_width())
+    args['terminal_width'] = get_terminal_width()
     bdata = json.dumps(args).encode('utf8')
     _, name = get_status(status_file)
     try:
--- a/mypy/util.py
+++ b/mypy/util.py
@@ -424,6 +424,10 @@ def split_words(msg: str) -> List[str]:
 
 def get_terminal_width() -> int:
     """Get current terminal width if possible, otherwise return the default one."""
+    forced_width = int(os.getenv('MYPY_FORCE_TERMINAL_WIDTH', '0'))
+    if forced_width:
+        return forced_width
+
     try:
         cols, _ = os.get_terminal_size()
     except OSError:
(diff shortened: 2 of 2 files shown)


asottile:
this regressed https://github.com/python/mypy/pull/8145/commits/9dccccaf68924a89f5d01d2a76a671a5dc7f3918

The-Compiler:
@asottile Can you elaborate? `shutil.get_terminal_size` [handles this case](https://github.com/python/cpython/blob/v3.9.0/Lib/shutil.py#L1344-L1354) and the test you added there seems to pass as well.

asottile:
> handles this case

look carefully, it handles when `os.get_terminal_size` *raises* but not when it returns (0, 0)

> and the test you added there seems to pass as well

nope

```
=================================== FAILURES ===================================
_______ TestGetTerminalSize.test_get_terminal_size_in_pty_defaults_to_80 _______
[gw0] linux -- Python 3.8.5 /tmp/mypy/venv/bin/python

self = <t.TestGetTerminalSize testMethod=test_get_terminal_size_in_pty_defaults_to_80>

    def test_get_terminal_size_in_pty_defaults_to_80(self) -> None:
        # when run using a pty, `os.get_terminal_size()` returns `0, 0`
        ret = os.terminal_size((0, 0))
        with mock.patch.object(os, 'get_terminal_size', return_value=ret):
>           assert get_terminal_width() == 80
E           AssertionError: assert 0 == 80
E            +  where 0 = get_terminal_width()

t.py:12: AssertionError
=========================== short test summary info ============================
FAILED t.py::TestGetTerminalSize::test_get_terminal_size_in_pty_defaults_to_80
============================== 1 failed in 0.50s ===============================
```

asottile:
note that the test never made it into master because it was "not useful" #8145

The-Compiler:
Well, I agree the test would've been useful :wink: I didn't catch that it indeed didn't end up being committed...

Seems to me like this should really be handled correctly in the stdlib by `shutil.get_terminal_size`. I can re-add the workaround to mypy, unless you're already on it.

asottile:
feel free

The-Compiler:
Sorry for the trouble - here's a fix: #9651

PR #9148, opened 2020-07-14
[mypyc] Check both operands when int logical op is not EQ or NEQ

TH3CHARLie:
This a bug I find during merging int logical ops.

Here's the less-than's inline function we have now:
```C
static inline bool CPyTagged_IsLt(CPyTagged left, CPyTagged right) {
    if (CPyTagged_CheckShort(left) && CPyTagged_CheckShort(right)) {
        return (Py_ssize_t)left < (Py_ssize_t)right;
    } else {
        return CPyTagged_IsLt_(left, right);
    }
}
```
Different from the equal's inline function, it checks both operands. So this PR completes that support and fix the potential bug.

### DIFF
--- a/mypyc/irbuild/ll_builder.py
+++ b/mypyc/irbuild/ll_builder.py
@@ -576,7 +576,14 @@ class LowLevelIRBuilder:
         op_type, c_func_desc = int_logical_op_mapping[op]
         result = self.alloc_temp(bool_rprimitive)
         short_int_block, int_block, out = BasicBlock(), BasicBlock(), BasicBlock()
-        check = self.check_tagged_short_int(lhs, line)
+        check_lhs = self.check_tagged_short_int(lhs, line)
+        if op in ("==", "!="):
+            check = check_lhs
+        else:
+            # for non-equal logical ops(less than, greater than, etc.), need to check both side
+            check_rhs = self.check_tagged_short_int(rhs, line)
+            check = self.binary_int_op(bool_rprimitive, check_lhs,
+                                       check_rhs, BinaryIntOp.AND, line)
         branch = Branch(check, short_int_block, int_block, Branch.BOOL_EXPR)
         branch.negated = False
         self.add(branch)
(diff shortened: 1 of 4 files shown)


TH3CHARLie:
There are two things to discuss also:

1. LT, GT, GE, LE rely on the same C function `CPyTagged_IsLt_` with different parameter order(swapping lhs and rhs) and use of `!` operator, should we implement this logic case by case in the `ll_builder`(three if-else branches within the `compare_tagged` function) or should we simply implement three C helpers like `CPyTagged_IsGt_`?
2. The IR tests for each of this logical op is actually huge(200 lines per each), although they are pretty trivial, I lean to merge these ops one per PR, what do you think?

cc @JukkaL

JukkaL:
> Different from the equal's inline function, it checks both operands

This is correct, and we need to check both operands. If only one of the operands is a short integer, we can't use an integer comparison, as a long integer could have either a high or low value (depending on the sign). For equality comparisons we can only check one of the operands -- no long integer can be equal to a short integer, since the ranges don't overlap.

TH3CHARLie:
> > Different from the equal's inline function, it checks both operands
> 
> This is correct, and we need to check both operands. If only one of the operands is a short integer, we can't use an integer comparison, as a long integer could have either a high or low value (depending on the sign). For equality comparisons we can only check one of the operands -- no long integer can be equal to a short integer, since the ranges don't overlap.

Exactly, that's why I submitted this before merging any other remaining ops, cause they will depend on this to produce the correct IR.

TH3CHARLie:
Also, an (expected) side-effect of this is that IR of <, >, <=, >= would be even verbose. So we'd raise the priority of the integer optimization.

JukkaL:
> LT, GT, GE, LE rely on the same C function CPyTagged_IsLt_ with different parameter order(swapping lhs and rhs) and use of ! operator, should we implement this logic case by case

Let's implement the parameter swapping logic in the IR builder, to remain consistent with how things work now.

PR #9149, opened 2020-07-14
[mypyc] Support swapping operands and negating result and merge int NEQ

TH3CHARLie:
This PR implements support for swapping operands and negating result when building logical ops mentioned in https://github.com/python/mypy/pull/9148#issuecomment-658123381. To demonstrate, NEQ is merged. Since it has no IR test, I build one `irbuild-int.test`.

### DIFF
--- a/mypyc/test-data/irbuild-basic.test
+++ b/mypyc/test-data/irbuild-basic.test
@@ -1358,7 +1358,7 @@ def lst(x):
 L0:
     r0 = len x :: list
     r1 = 0
-    r2 = CPyTagged_IsNe(r0, r1)
+    r2 = r0 != r1
     if r2 goto L1 else goto L2 :: bool
 L1:
     r3 = 1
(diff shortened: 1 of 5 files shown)


TH3CHARLie:
The following merges will base on this branch.

mypyc/test-data/irbuild-basic.test
```diff
@@ -1358,7 +1358,7 @@ def lst(x):
 L0:
     r0 = len x :: list
     r1 = 0
-    r2 = CPyTagged_IsNe(r0, r1)
+    r2 = r0 != r1
```
TH3CHARLie, line 1361: This is caused by https://github.com/python/mypy/commit/358522e28cd58d95daf36256c46eb7ffcc55eea4

mypyc/test-data/irbuild-int.test
```diff
@@ -0,0 +1,25 @@
+[case testIntEq]
```
JukkaL, line 1: Should the test case name be `testIntNeq`?

PR #9163, opened 2020-07-17
[mypyc] Fix signed integer comparison

TH3CHARLie:
https://github.com/python/mypy/commit/358522e28cd58d95daf36256c46eb7ffcc55eea4 generates inline comparison between short ints, explicit conversion to signed is missing, though, causing negative cases to fail.

This PR add explicit type casts(although the name truncate here is a little bit misleading).

This PR will fix microbenchmark `int_list`.

### DIFF
--- a/mypyc/codegen/emitfunc.py
+++ b/mypyc/codegen/emitfunc.py
@@ -14,7 +14,7 @@ from mypyc.ir.ops import (
     NAMESPACE_TYPE, NAMESPACE_MODULE, RaiseStandardError, CallC, LoadGlobal, Truncate,
     BinaryIntOp
 )
-from mypyc.ir.rtypes import RType, RTuple
+from mypyc.ir.rtypes import RType, RTuple, is_tagged
 from mypyc.ir.func_ir import FuncIR, FuncDecl, FUNC_STATICMETHOD, FUNC_CLASSMETHOD
 from mypyc.ir.class_ir import ClassIR
 
@@ -438,7 +438,15 @@ class FunctionEmitterVisitor(OpVisitor[None], EmitterInterface):
         dest = self.reg(op)
         lhs = self.reg(op.lhs)
         rhs = self.reg(op.rhs)
-        self.emit_line('%s = %s %s %s;' % (dest, lhs, op.op_str[op.op], rhs))
+        lhs_cast = ""
+        rhs_cast = ""
+        if (is_tagged(op.lhs.type) and is_tagged(op.rhs.type)
+                and op.op in (BinaryIntOp.SLT, BinaryIntOp.SGT,
+                              BinaryIntOp.SLE, BinaryIntOp.SGE)):
+            lhs_cast = "(Py_ssize_t)"
+            rhs_cast = "(Py_ssize_t)"
+        self.emit_line('%s = %s%s %s %s%s;' % (dest, lhs_cast, lhs,
+                                               op.op_str[op.op], rhs_cast, rhs))
 
     # Helpers
 
--- a/mypyc/codegen/emitfunc.py
+++ b/mypyc/codegen/emitfunc.py
@@ -14,7 +14,9 @@ from mypyc.ir.ops import (
     NAMESPACE_TYPE, NAMESPACE_MODULE, RaiseStandardError, CallC, LoadGlobal, Truncate,
     BinaryIntOp
 )
-from mypyc.ir.rtypes import RType, RTuple, is_tagged
+from mypyc.ir.rtypes import (
+    RType, RTuple, is_tagged, is_int32_rprimitive, is_int64_rprimitive
+)
 from mypyc.ir.func_ir import FuncIR, FuncDecl, FUNC_STATICMETHOD, FUNC_CLASSMETHOD
 from mypyc.ir.class_ir import ClassIR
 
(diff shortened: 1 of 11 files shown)


mypyc/irbuild/ll_builder.py
```diff
@@ -588,7 +588,17 @@ def compare_tagged(self, lhs: Value, rhs: Value, op: str, line: int) -> Value:
...
         self.add(branch)
         self.activate_block(short_int_block)
-        eq = self.binary_int_op(bool_rprimitive, lhs, rhs, op_type, line)
+        # for now equal op, explicitly convert tagged(unsigned) to signed to include negative
+        # situtations
+        if op not in ("==", "!="):
+            lhs_short = self.add(Truncate(lhs, short_int_rprimitive,
+                                          c_pyssize_t_rprimitive, line))
+            rhs_short = self.add(Truncate(rhs, short_int_rprimitive,
+                                          c_pyssize_t_rprimitive, line))
```
JukkaL, line 597: This is not very principled. The truncate op is redundant, and also confusingly the sizes of source and target types are the same, so there's no truncation actually going on. 

Instead, we should generate the correct C code for short int comparison (add the cast to a signed integer type). Maybe you can look at the `type` attribute of `BinaryIntOp` to determine whether a signed comparison is needed when emitting C.

JukkaL:
Note that there's a merge conflict.

mypyc/irbuild/ll_builder.py
TH3CHARLie: Yeah, the truncate also looks weird to me. So when we emit a BinaryIntOp, we check if both operands are short_int(and probably check if the op is not EQ or NEQ as well since they do not require casts to have the correct result) to cast them into `py_ssize_t`

JukkaL: More generally, also plain tagged integers need to use a cast since they are signed. 

Probably the cleanest way to check the operand types would be to update every `Value` to contain the result type, but I'm not sure if this is necessary.

mypyc/codegen/emitfunc.py
```diff
@@ -438,7 +438,15 @@ def visit_binary_int_op(self, op: BinaryIntOp) -> None:
...
         rhs = self.reg(op.rhs)
-        self.emit_line('%s = %s %s %s;' % (dest, lhs, op.op_str[op.op], rhs))
+        lhs_cast = ""
+        rhs_cast = ""
+        if (is_tagged(op.lhs.type) and is_tagged(op.rhs.type)
+                and op.op in (BinaryIntOp.SLT, BinaryIntOp.SGT,
+                              BinaryIntOp.SLE, BinaryIntOp.SGE)):
+            lhs_cast = "(Py_ssize_t)"
+            rhs_cast = "(Py_ssize_t)"
+        self.emit_line('%s = %s%s %s %s%s;' % (dest, lhs_cast, lhs,
```
JukkaL, line 448: We'll also need similar checks for unsigned comparisons -- if the types aren't tagged, unsigned comparisons require a cast to the corresponding unsigned type.

Since we aren't using unsigned comparisons yet, it would be good to add unit tests for these.

mypyc/ir/ops.py
```diff
@@ -1291,10 +1296,14 @@ class BinaryIntOp(RegisterOp):
...
-        LEQ: '<=',
-        GEQ: '>=',
+        SLT: '<',
+        SGT: '>',
+        SLE: '<=',
+        SGE: '>=',
+        ULT: '<',
+        UGT: '>',
+        ULE: '<=',
+        UGE: '>=',
```
JukkaL, line 1306: The pretty printed IR should distinguish between signed and unsigned comparisons. They look identical now, I think.

mypyc/codegen/emitfunc.py
TH3CHARLie: I agree that we should add these checks when the types are int32/int64 and op is unsigned, though we don't have that kind of usages right now so maybe leave it out from this PR would be OK.

I am not sure what kind of unit tests we are adding here. My guess is to ensure this kind of cast is enforced in the generated C code? Probably in `test_emitfunc.py`?

JukkaL: `test_emitfunc.py` would be a reasonable location for unit tests.

mypyc/test/test_emitfunc.py
```diff
@@ -245,6 +247,10 @@ def test_dict_contains(self) -> None:
             'in', self.b, self.o, self.d,
             """cpy_r_r0 = PyDict_Contains(cpy_r_d, cpy_r_o);""")
 
+    def test_binary_int_op(self) -> None:
+        self.assert_emit(BinaryIntOp(bool_rprimitive, self.s1, self.s2, BinaryIntOp.SLT, 1),
+                         """cpy_r_r0 = (Py_ssize_t)cpy_r_s1 < (Py_ssize_t)cpy_r_s2;""")
```
JukkaL, line 252: Also test unsigned comparison with `int64` and `int32`. These need a cast to the corresponding unsigned type (`uint64_t` or `uint32_t`, respectively). Finally, test that unsigned comparison with `short_int_rprimitive` and signed comparison with `int32` or `int64` doesn't use a cast.

TH3CHARLie: Thanks! This actually catches a bug in ops.py that constant's definition are overlapped.

PR #9166, opened 2020-07-17
Validate `follow_imports` values in mypy.ini

msullivan:
Currently we only validate the values on the commandline, which allows
bogus values to show up which will cause us to take the `else` branch
anywhere we consider follow_imports, which isn't likely to do anything
particularly correct.


msullivan:
Whoops

PR #9172, opened 2020-07-19
Add command line support to enable/disable error codes

cph-w:
Fixes https://github.com/python/mypy/issues/8975 

Add the `--disable-error-code` flag to disable error codes globally.
Add the `--enable-error-code` flag to enable error codes globally

This is my first time working on this repo, so please let me know anything additional that needs completing.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -553,6 +553,36 @@ of the above sections.
     Note: the exact list of flags enabled by running :option:`--strict` may change
     over time.
 
+.. option:: --disable-error-code
+
+    This flag allows disabling one or multiple error codes globally.
+
+    .. code-block:: python
+
+        # no flag
+        x = 'a string'
+        x.trim()  # error: "str" has no attribute "trim"  [attr-defined]
+
+        # --disable-error-code attr-defined
+        x = 'a string'
+        x.trim()
+
+.. option:: --enable-error-code
+
+    This flag allows enabling one or multiple error codes globally.
+
+    Note: This flag will override disabled error codes from the --disable-error-code
+    flag
+
+    .. code-block:: python
+
+        # --disable-error-code attr-defined
+        x = 'a string'
+        x.trim()
+
+        # --disable-error-code attr-defined --enable-error-code attr-defined
+        x = 'a string'
+        x.trim()  # error: "str" has no attribute "trim"  [attr-defined]
 
 .. _configuring-error-messages:
 
--- a/mypy/errorcodes.py
+++ b/mypy/errorcodes.py
@@ -3,19 +3,26 @@
 These can be used for filtering specific errors.
 """
 
-from typing import List
+from typing import Dict, List
 from typing_extensions import Final
 
 
 # All created error codes are implicitly stored in this list.
 all_error_codes = []  # type: List[ErrorCode]
 
+error_codes = {}  # type: Dict[str, ErrorCode]
+
 
 class ErrorCode:
-    def __init__(self, code: str, description: str, category: str) -> None:
+    def __init__(self, code: str,
+                 description: str,
+                 category: str,
+                 default_enabled: bool = True) -> None:
         self.code = code
         self.description = description
         self.category = category
+        self.default_enabled = default_enabled
+        error_codes[code] = self
 
     def __str__(self) -> str:
         return '<ErrorCode {}>'.format(self.code)
--- a/mypy/errors.py
+++ b/mypy/errors.py
@@ -164,7 +164,9 @@ class Errors:
                  show_error_codes: bool = False,
                  pretty: bool = False,
                  read_source: Optional[Callable[[str], Optional[List[str]]]] = None,
-                 show_absolute_path: bool = False) -> None:
+                 show_absolute_path: bool = False,
+                 enabled_error_codes: Optional[Set[ErrorCode]] = None,
+                 disabled_error_codes: Optional[Set[ErrorCode]] = None) -> None:
         self.show_error_context = show_error_context
         self.show_column_numbers = show_column_numbers
         self.show_error_codes = show_error_codes
@@ -172,6 +174,8 @@ class Errors:
         self.pretty = pretty
         # We use fscache to read source code when showing snippets.
         self.read_source = read_source
+        self.enabled_error_codes = enabled_error_codes or set()
+        self.disabled_error_codes = disabled_error_codes or set()
         self.initialize()
 
     def initialize(self) -> None:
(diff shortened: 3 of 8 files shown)


mypy/errorcodes.py
```diff
@@ -3,19 +3,25 @@
 These can be used for filtering specific errors.
 """
 
-from typing import List
+from typing import Dict, List, Set
 from typing_extensions import Final
 
 
 # All created error codes are implicitly stored in this list.
 all_error_codes = []  # type: List[ErrorCode]
```
cph-w, line 11: This list seemed so far unused, and for my use case was not the ideal data structure, so I added the below dict. If it's preferred I can change to adding the error codes here and iterating through.

JukkaL:
Thanks for the PR! It would be nice to also have `--enable-error-code` available. 

This is a sketch of how it could work:

* Add a boolean attribute to `ErrorCode` called `enabled`, for example, that controls whether the error code is enabled by default. It defaults to true. If it's false, the error code is disabled by default.
* Initially error codes that are disabled by default are in the disabled set.
* `--enable-error-code` will remove an error code from the disabled set.

Also,  `--disable-error-code X` followed by `--enable-error-code X` on the command line results in `X` being enabled.

Initially we wouldn't have any error codes that are disabled by default, but you can test this manually by making some error code disabled by default. Tests can use `--disable..` followed by `--enable..` to test this.

cph-w:
@JukkaL Thanks for getting back to me. Your implementation plan is clear and clean, I'll implement this tomorrow.

cph-w:
@JukkaL I have what looks like a working implementation for this now, but I have failing tests. When I disable error codes in a test it looks like possibly that state is persisting between cases that run subsequently. Do you have an established/preferred way way to clean something like this up?

cph-w:
@JukkaL Implemented a reset for the error codes that get changed, and `--enable-error-code`. Let me know any feedback or required changes :)

cph-w:
Hey @JukkaL , no rush but looking for feedback on this PR when somebody is free or direction for the next steps to contributing this.

mypy/errorcodes.py
```diff
@@ -3,19 +3,27 @@
...
-    def __init__(self, code: str, description: str, category: str) -> None:
+    def __init__(self, code: str,
+                 description: str,
+                 category: str,
+                 default_enabled: bool = True) -> None:
         self.code = code
         self.description = description
         self.category = category
+        self.default_enabled = default_enabled
+        self.enabled = default_enabled
```
JukkaL, line 25: I don't like having mutable state here. I think that it's better to filter error codes based on the enabled/disabled sets (my other comments contain additional context).

mypy/errors.py
```diff
@@ -351,7 +351,9 @@ def add_error_info(self, info: ErrorInfo) -> None:
         self._add_error_info(file, info)
 
     def is_ignored_error(self, line: int, info: ErrorInfo, ignores: Dict[int, List[str]]) -> bool:
-        if line not in ignores:
+        if info.code and info.code.enabled is False:
+            return True
```
JukkaL, line 355: You can pass the enabled and disabled error codes to `Errors`, and we can use them together with the `default_enabled` attribute to decide whether to ignore them or not.

This way there will be less global mutable state, and it will be easier to have per-file enabled/disabled sets in the future.

mypy/main.py
```diff
@@ -860,6 +869,25 @@ def set_strict_flags() -> None:
         parser.error("You can't make a variable always true and always false (%s)" %
                      ', '.join(sorted(overlap)))
 
+    # Process disabled (--disable-error-code) codes
+    disabled_codes = set(options.disable_error_code)
+    if disabled_codes:
+        invalid_error_codes = process_error_codes(disabled_codes, enable=False)
+
+        if invalid_error_codes:
+            parser.error("Disabled error codes provided invalid values: (%s)" %
```
JukkaL, line 878: This error message is a bit verbose. What about `"Invalid error code(s): %s"`?

mypy/main.py
```diff
@@ -860,6 +869,25 @@ def set_strict_flags() -> None:
...
+                         ', '.join(sorted(invalid_error_codes)))
+
+    # Process enabled (--enable-error-code) codes
+    # If a flag is both enabled and disabled the result is enabled
+    enabled_codes = set(options.enable_error_code)
+    if enabled_codes:
+        invalid_error_codes = process_error_codes(enabled_codes, enable=True)
+
+        if invalid_error_codes:
+            parser.error("Enabled error codes provided invalid values: (%s)" %
```
JukkaL, line 888: Similar to above.

mypy/test/testcheck.py
```diff
@@ -253,6 +254,15 @@ def run_case_once(self, testcase: DataDrivenTestCase,
...
             check_test_output_files(testcase, incremental_step, strip_prefix='tmp/')
 
+        # Post run cleanup step to reset the state of the ErrorCodes that may
+        # have been enabled or disabled by flags
+        if options.disable_error_code or options.enable_error_code:
+            altered_error_codes = set(
+                options.disable_error_code).union(set(options.enable_error_code))
+            for code in altered_error_codes:
+                error_code = errorcodes.error_codes[code]
+                error_code.enabled = error_code.default_enabled
```
JukkaL, line 264: Once error code objects aren't mutated, this can be removed.

test-data/unit/check-flags.test
```diff
@@ -1501,3 +1501,48 @@ class ShouldNotBeFine(x): ...  # E: Class cannot subclass 'x' (has type 'Any')
 disallow_subclassing_any = True
 \[mypy-m]
 disallow_subclassing_any = False
+
+[case testDisableErrorCode]
```
JukkaL, line 1505: Add tests for invalid error codes?

docs/source/command_line.rst
```diff
@@ -554,6 +554,36 @@ of the above sections.
...
+
+        # --disable-error-code attr-defined
+        x = 'a string'
+        x.trim()
+
+.. option:: --enable-error-code
+
+    This flag allows enabling one or multiple error codes globally.
+
+    NOTE: Will override disabled error codes from the --disable-error-code
```
JukkaL, line 575: Nist: For consistency, don't use all caps (above we use "Note: ..."). Use a full sentence, such as "This will ...".

mypy/main.py
```diff
@@ -984,6 +1012,20 @@ def process_cache_map(parser: argparse.ArgumentParser,
         options.cache_map[source] = (meta_file, data_file)
 
 
+def process_error_codes(codes: Iterable[str], enable: bool) -> List[str]:
```
JukkaL, line 1015: When error code code objects aren't mutated, this can go away.

ofek:
Need rebase btw

cph-w:
@JukkaL Switched back to passing sets of error codes to Errors, addressed your other comments. Let me know if there's anything else. Sorry for the short delay getting back to you, was moving house at the weekend and running without internet, will try to respond quickly to any further required iteration.

PR #9177, opened 2020-07-20
Show all overloads in error message

### LINKED ISSUE #9175: show more than two overloads in error message
* feature request
* 

```
from typing import overload
@overload
def foo(x: int) -> None: ...
@overload
def foo(x: str) -> None: ...
@overload
def foo(x: bytes) -> None: ...
@overload
def foo(x: bytearray) -> None: ...
def foo(x): return None
def main() -> None: foo(None)
```

* Even with verbose, the error message shows only two overloads

```
$ mypy foo.py 
foo.py:9: error: Overloaded function signature 4 will never be matched: signature 3's parameter type(s) are the same or broader
foo.py:11: error: No overload variant of "foo" matches argument type "None"
foo.py:11: note: Possible overload variants:
foo.py:11: note:     def foo(x: int) -> None
foo.py:11: note:     def foo(x: str) -> None
foo.py:11: note:     <2 more similar overloads not shown, out of 4 total overloads>
$ mypy -vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv foo.py 2>&1 | grep 'def foo'
foo.py:11: note:     def foo(x: int) -> None
foo.py:11: note:     def foo(x: str) -> None
```

* I would expect increased verbosity to increase the number of overloads shown, in case someone wants to see them all
* 

```
$ mypy --version
mypy 0.782
$ head -1 `which mypy`
#!/home/akuli/.local/bin/python3.8
```

haven't tried mypy from master yet

* verbosity flags used as shown above

Akuli:
closes #9175


JukkaL:
Thanks for the PR!

I think that we can just always, unconditionally show all overloads. It may be bit verbose, but probably better since otherwise we'd be hiding important information. `-v` is mostly used for debugging issues with mypy or configuration, and I'd rather not recommend users to use it for basic use cases.

Could you simplify the PR to just make the change unconditional?

Akuli:
What should I do in `pretty_overload_matches`? Show the matching overloads but not the non-matching overloads?

Akuli:
I was quite confused about what that method does, so let me ask differently: should it show all overloads, or should it show all overloads that have the correct number of arguments but the arguments have wrong types?

Akuli:
I made it show all the overloads

Akuli:
Is there something else I need to do to get this merged?

JukkaL:
Sorry, I lost track of this PR. If you can fix the merge conflict, I can do the final review round now.

Akuli:
Fixing the unit tests will take some time, many are failing :)

Akuli:
@JukkaL Ready for review

JukkaL (approved):
Looks good! This makes things more consistent and also simplifies the code.

PR #9158, opened 2020-07-16
[mypyc] Constant integer registers optimization

TH3CHARLie:
Recent changes to mypyc introduce low-level, inline integer ops, which makes both IR and generated C code verbose and potentially hurting performance. This PR introduces an on-demand optimization pass that find out all the registers with constant integer values.

Relates mypyc/mypyc#749 mypyc/mypyc#746

PR #9181, opened 2020-07-21
[mypyc] Optimize const int regs during pretty IR printing

TH3CHARLie:
follow up of #9158, make IR less verbose


TH3CHARLie:
the main changes lie in `func_ir.py` and `ops.py` and I used re to replace reg names with values during the IR generation.

mypyc/ir/func_ir.py
```diff
@@ -11,6 +12,7 @@
 )
 from mypyc.ir.rtypes import RType, deserialize_type
 from mypyc.namegen import NameGenerator
+from mypyc.analysis.const_int import find_constant_integer_registers
```
JukkaL, line 15: I'd prefer if `mypyc.ir` didn't depend on `mypyc.analysis`, to keep import structures less tangled. I think that we can move `mypyc.analysis.const_int` to `mypyc.ir.const_int`, since the analysis it performs is actually quite trivial. If we implement a fancier constant analysis, we'd add it to `mypyc.analysis`, but we wouldn't use it for IR pretty printing.

mypyc/ir/func_ir.py
```diff
@@ -244,9 +248,13 @@ def format_blocks(blocks: List[BasicBlock], env: Environment) -> List[str]:
                 and ops[-1].label == blocks[i + 1]):
             # Hide the last goto if it just goes to the next basic block.
             ops = ops[:-1]
+        regex = re.compile('r[0-9]+')
```
JukkaL, line 251: This needs to match whole words only, as otherwise we might be matching components of identifiers. `\b` does the trick (both at the beginning and the end).

mypyc/test-data/analysis.test
```diff
@@ -10,23 +10,17 @@ def f(a: int) -> None:
 [out]
 def f(a):
     a :: int
-    r0 :: short_int
     x :: int
     r1 :: bool
-    r2, r3, r4 :: native_int
+    r3 :: native_int
```
JukkaL, line 15: Unfortunately, this leaves gaps in the register numbering. A relatively easy way to fix this would be to give `LoadInt` values different names with independent numbering, such as `i0`, `i1`, etc., instead of `r0`, `r1`, and so on. This way we'd just filter out all `iX` names, and `rX` would remain without gaps.

mypyc/ir/ops.py
```diff
@@ -232,17 +232,23 @@ def format(self, fmt: str, *args: Any) -> str:
                 i = n
         return ''.join(result)
 
-    def to_lines(self) -> List[str]:
+    def to_lines(self, const_regs: Dict[str, int] = {}) -> List[str]:
```
JukkaL, line 235: Nit: It's considered error-prone to have a mutable default value. It's better to use `None` as the default and construct an empty dictionary if needed in the body. Some linters complain about this. (I know it's semantically fine here, but I think it's better to avoid this for style reasons.)

PR #9187, opened 2020-07-22
[mypyc] Expand int ops when operation contains at least one tagged int

TH3CHARLie:
Follow-up of #9108 and #9127, generates the new inlined style ops when there is at least one tagged integer presented(The previous two PRs actually specialized two ints and two short_ints cases). After this and the remaining merge, we will get rid of `CPyTagged_IsEq`, `CPyTagged_IsNe`, `CPyTagged_IsLt`, `CPyTagged_IsLe`, `CPyTagged_IsGt` and `CPyTagged_IsGe`


TH3CHARLie:
I'll do the IR test updates once const_int has been applied to pretty IR printing, to reduce conflicts.

TH3CHARLie:
should we remove `CPyTagged_IsEq` from `CPy.h` and its related test in `test_capi.cc`? and for `CPyTagged_IsNe` and `CPyTagged_IsLt` as well?

JukkaL:
> should we remove CPyTagged_IsEq from CPy.h and its related test in test_capi.cc? and for CPyTagged_IsNe and CPyTagged_IsLt as well?

Yeah, we should remove them from CPy.h, since they are no longer used. However, we should consider updating the tests to use the alternative functions with an underscore suffix.

mypyc/test-data/analysis.test
```diff
@@ -197,31 +241,63 @@ def f(n: int) -> None:
...
 L0:
 L1:
-    r0 = CPyTagged_IsLt(n, 10)
-    if r0 goto L2 else goto L3 :: bool
+    r1 = n & 1
+    r2 = r1 == 0
+    r3 = 10 & 1
+    r4 = r3 == 0
+    r5 = r2 & r4
+    if r5 goto L2 else goto L3 :: bool
```
JukkaL, line 257: Not directly related to this diff: we could simplify this sequence of ops like this:

```
r1 = n | 10
r2 = r1 & 1
r3 = r2 == 0
if r3 goto ...
```
```

TH3CHARLie:
> > should we remove CPyTagged_IsEq from CPy.h and its related test in test_capi.cc? and for CPyTagged_IsNe and CPyTagged_IsLt as well?
> 
> Yeah, we should remove them from CPy.h, since they are no longer used. However, we should consider updating the tests to use the alternative functions with an underscore suffix.

then it would make more sense we remove them and update the C++ tests after all merges in the python end have completed.

Issue #9169, opened 2020-07-18
Regression from 0.750 - Literal types cause significant slowdown


CommanderPrS:
### Are you reporting a bug, or opening a feature request?

This is a bug. It is a regression from the 0.750 version.

### Please insert below the code you are checking with mypy, or a mock-up repro if the source is private. We would appreciate if you try to simplify your case to a minimal repro.

Here is the code that triggers this bug. It is essentially a list of `TypedDict` objects created in some nested fashion. The definitions for these objects are coming from the `boto3-stubs` library, which is a typed wrapper around the `boto3` library, which is a Python SDK for talking to AWS REST APIs. The same behavior could probably be reproduced with just custom made `TypeDict` objects, but, given, the number of them and the relationship between them, it is easier just to reuse the ones from the library.

To install it, you need to have this line in `requirements.txt` (the exact version probably does not matter, but this is what I have):

```
boto3-stubs[autoscaling,cloudwatch,ec2,lambda,logs] == 1.10.35
```

**IMPORTANT:** After running `pip install` for the above line, one must also run this command to make `boto3-stubs` initialize itself:

```
python -m mypy_boto3
```

If doing this from `tox.ini`, you need to have this line there:

```
commands_pre = python -m mypy_boto3
```

Here is the offending code:

```
import typing as t

from mypy_boto3.ec2.type_defs import (
    DescribeInstancesResultTypeDef,
    FilterTypeDef,
    InstanceNetworkInterfaceAttachmentTypeDef,
    InstanceNetworkInterfaceTypeDef,
    InstanceStateTypeDef,
    InstanceTypeDef,
    PlacementTypeDef,
    ReservationTypeDef,
    TagTypeDef,
)

x: t.List[DescribeInstancesResultTypeDef] = [
    DescribeInstancesResultTypeDef(
        Reservations=[
            ReservationTypeDef(
                Instances=[
                    InstanceTypeDef(
                        InstanceId="az1-instance1-id1",
                        Placement=PlacementTypeDef(
                            AvailabilityZone="az1",
                        ),
                        State=InstanceStateTypeDef(
                            Name="running",
                        ),
                        NetworkInterfaces=[
                            InstanceNetworkInterfaceTypeDef(
                                Attachment=InstanceNetworkInterfaceAttachmentTypeDef(
                                    DeviceIndex=0,
                                ),
                                NetworkInterfaceId="az1-instance1-nic0-id1",
                                PrivateDnsName="az1-instance1-nic0-dns1",
[...]

PR #9192, opened 2020-07-22
Speed up make_simplified_union for Literal[string]s (issue #9169)

Akuli:
`make_simplified_union` no longer runs for 0.5 seconds every time it's called with a Union containing 270 literals of strings.

like I explained in #9169, this only fixes half of the problem and I'm not capable of fixing the other half. This function is still called 1098 times for the "reduced" example code, and IMO it should be called only once.

### DIFF
--- a/mypy/typeops.py
+++ b/mypy/typeops.py
@@ -344,21 +344,33 @@ def make_simplified_union(items: Sequence[Type],
     from mypy.subtypes import is_proper_subtype
 
     removed = set()  # type: Set[int]
-    for i, ti in enumerate(items):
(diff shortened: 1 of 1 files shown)


Akuli:
travis has actually passed, it just doesn't show up like that on github for whatever reason

Akuli:
>Changing mypy so that the simplification happens less often would definitely be nice, but it will be non-trivial to implement.

mypy used to be like that, but 25b1c4ddb349da2e79a6f1f30a38a5bc3b0f6509 broke it

PR #9196, opened 2020-07-23
Check __setattr__ when property is not settable

### LINKED ISSUE #9160: class with __setattr__ reports properties are read-only
When a class implements `__setattr__`, mypy still reports properties as read-only.

Minimum example:

````
class Foo:
    def __init__(self) -> None:
        self._d = {'a': 1}

    def __setattr__(self, name: str, value: Any) -> None:
        if hasattr(self, '_d'):
            self._d[name] = value
        else:
            super(self.__class__, self).__setattr__(name, value)

    @property
    def a(self) -> int:
        return self._d['a']

f = Foo()
f.a = 2
````

The error points to the last line in the above code: `error: Property "a" defined in "Foo" is read-only`.

Expectation: Since `Foo` implements `__setattr__`, mypy should allow the attribute to be set, even though it overlaps with a class property.

There are no issues identified if the property `a` is removed, and I am using version 0.781.

### PR DESCRIPTION
**Pull request by kamilturek**
Fixes #9160

### CODE REVIEW

File: mypy/checkmember.py [hunk 1]
```diff
@@ -550,7 +550,8 @@ def analyze_var(name: str,
             return mx.chk.handle_partial_var_type(typ, mx.is_lvalue, var, mx.context)
         if mx.is_lvalue and var.is_property and not var.is_settable_property:
             # TODO allow setting attributes in subclass (although it is probably an error)
-            mx.msg.read_only_property(name, itype.type, mx.context)
+            if info.names.get('__setattr__') is None:
```
Akuli, line 553: how about `if '__setattr__' not in info.names:`

JukkaL: The best thing would be to use `info.get`, since that also looks at base classes that might define `__setattr__`.

mypy/checkmember.py
```diff
@@ -560,7 +561,8 @@ def analyze_var(name: str,
             if mx.is_lvalue:
                 if var.is_property:
                     if not var.is_settable_property:
-                        mx.msg.read_only_property(name, itype.type, mx.context)
+                        if info.names.get('__setattr__') is None:
```
JukkaL, line 564: Similar to above.

test-data/unit/check-setattr.test
```diff
@@ -0,0 +1,12 @@
+[case testSetAttrWithProperty]
+from typing import Any
+class Foo:
+    _a = 1
+    def __setattr__(self, name: str, value: Any) -> None:
```
JukkaL, line 5: Can you add these additional test cases:
* `__setattr__` inherited from a base class.
* The value argument of `__setattr__` has a specific type such as `int` (different from the type of the property). Check that assigning a non-`int` value to the attribute gets flagged by mypy.

rhshadrach:
@JukkaL friendly ping

hauntsaninja (approved):
This looks good to me, and it looks like Jukka's concerns have been addressed. Thanks @kamilturek for your series of PRs! :-)

hauntsaninja:
Actually, I think I'm going to revert this, see #9474

Issue #9193, opened 2020-07-22
stubtest with config file


pikachu:
I want to pass in some plugins to `stubtest.py`, but it [calls](https://github.com/python/mypy/blob/65186ae1e23fd44f1d7e6aa2c4458bdf55640742/mypy/stubtest.py#L956) `build` without passing in any `extra_plugins` (you can see `extra_plugins` is a [param for`build`](https://github.com/python/mypy/blob/65186ae1e23fd44f1d7e6aa2c4458bdf55640742/mypy/build.py#L143)). Is this an in-progress feature?

More generally, can we make stubtest take in a config file the same way static type checking does? Is it true that most users would like to run runtime and static type checking with the same configuration?

Thanks in advance!

hauntsaninja:
My primary goal with stubtest was to improve tests for typeshed stubs, so I just took the path of least resistance with the build (stubgen, a much older script, also kind of does its own thing for its build).

> Is it true that most users would like to run runtime and static type checking with the same configuration?

Plugins aside, I'm not sure how many mypy config options would be relevant to what stubtest does (since they do pretty different things...), so I'd be a little wary of changes purporting to make their configuration the same. Anyway, at the very least supporting plugins in stubtest sounds like something that's useful and should be done.

Btw, I might have missed something, but I'm not sure that anything currently passes `extra_plugins` to `build` (I think existing use is all from options), although refactoring some logic from `load_plugins_from_config` plus passing `extra_plugins` is probably the way to go here if you're looking to implement this.

pikachu:
Sweet - I ended up creating a [PR](https://github.com/python/mypy/pull/9203) for this. I didn't modify `load_plugins_from_config` at all - was able to isolate this all in `stubtest.py`. We can continue discussion there.

PR #9203, opened 2020-07-24
Adding config file for stubtest.py

pikachu:
This fixes [9193](https://github.com/python/mypy/issues/9193).

Before: You could not use plugins during the stub loading step.

Problem: If you need plugins for regular static type checking, you may need them when loading stubs for runtime type checking as well.

Solution: Add support for loading plugins from a config file. The easiest way to do this is to just load the config file and let `build` take care of it.

### DIFF
--- a/mypy/stubtest.py
+++ b/mypy/stubtest.py
@@ -1166,15 +1166,15 @@ def parse_options(args: List[str]) -> argparse.Namespace:
         help="Ignore unused whitelist entries",
     )
     config_group = parser.add_argument_group(
-        title='Config file',
+        title='mypy config file',
         description="Use a config file instead of command line arguments. "
-                    "Plugins are the only supported configurations as of right "
-                    "now.",
+                    "Plugins and mypy path are the only supported "
+                    "configurations.",
     )
     config_group.add_argument(
-        '--config-file',
-        help="Configuration file, must have a [mypy] section "
-             "(defaults to {})".format(', '.join(defaults.CONFIG_FILES)))
+        '--mypy-config-file',
+        help="Configuration file, must have a [mypy] section",
+    )
 
     return parser.parse_args(args)
 
--- a/mypy/stubtest.py
+++ b/mypy/stubtest.py
@@ -16,7 +16,7 @@ import warnings
 from functools import singledispatch
 from pathlib import Path
 from typing import (
-    Any, Dict, Generic, Iterator, List, Optional, Sequence, Tuple, TypeVar, Union,
+    Any, Dict, Generic, Iterator, List, Optional, Tuple, TypeVar, Union,
     cast,
 )
 
@@ -27,11 +27,8 @@ import mypy.modulefinder
 import mypy.types
 from mypy import nodes
 from mypy.config_parser import parse_config_file
-from mypy.errors import Errors
-from mypy.fscache import FileSystemCache
 from mypy.options import Options
-from mypy.plugin import Plugin
-from mypy.util import read_py_file, FancyFormatter
+from mypy.util import FancyFormatter
 
 
 class Missing:
@@ -931,7 +928,6 @@ def build_stubs(
     modules: List[str],
     options: Options,
     find_submodules: bool = False,
-    extra_plugins: Optional[Sequence[Plugin]] = None,
 ) -> List[str]:
     """Uses mypy to construct stub objects for the given modules.
 
@@ -965,7 +961,7 @@ def build_stubs(
             all_modules.extend(s.module for s in found_sources if s.module not in all_modules)
 
     try:
-        res = mypy.build.build(sources=sources, options=options, extra_plugins=extra_plugins)
+        res = mypy.build.build(sources=sources, options=options)
     except mypy.errors.CompileError as e:
         output = [_style("error: ", color="red", bold=True), "failed mypy compile.\n", str(e)]
         print("".join(output))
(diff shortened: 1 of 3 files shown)


mypy/stubtest.py
```diff
@@ -1040,9 +1043,23 @@ def test_stubs(args: argparse.Namespace) -> int:
     options = Options()
     options.incremental = False
     options.custom_typeshed_dir = args.custom_typeshed_dir
+    options.config_file = args.config_file
+
+    plugins = []
+    if options.config_file:
+        def set_strict_flags() -> None: # not needed yet
```
pikachu, line 1050: not sure if there's a better way to deal with this - seems like we just don't really need that param but I didn't want to pass in `lambda _: None`. open to suggestions here

mypy/stubtest.py
```diff
@@ -1133,6 +1150,15 @@ def parse_options(args: List[str]) -> argparse.Namespace:
         action="store_true",
         help="Ignore unused whitelist entries",
     )
+    config_group = parser.add_argument_group(
```
pikachu, line 1153: copied directly from `main.py`.

pikachu: modified the description, though

pikachu:
@hauntsaninja Thanks for the comment.

I don't have a great idea of all the cases where you want to use different settings (not just a subset, but actually **different**). I wonder if there are other ways we could make it more clear? Maybe we rename the arg to `plugin-config-file`, or have a _big print statement_ ™ that tells the user that we are only processing plugins and mypy_path (if you pass in a config file).

In my use case, it's nice to have the config file because it automatically picks up on all of the following settings:

```wml
[mypy]
mypy_path = path
plugins = plugin1, plugin2, plugin3

[mypy.plugins.django-stubs]
setting1 = "value1"
setting2 = "value2"
```

mypy/stubtest.py
```diff
@@ -1040,9 +1051,28 @@ def test_stubs(args: argparse.Namespace) -> int:
...
     options.incremental = False
     options.custom_typeshed_dir = args.custom_typeshed_dir
+    options.config_file = args.config_file
+
+    plugins = None # type: Optional[Sequence[Plugin]]
+    if options.config_file:
+        def set_strict_flags() -> None:  # not needed yet
+            return
+        parse_config_file(options, set_strict_flags, options.config_file, sys.stdout, sys.stderr)
+        errors = Errors(options.show_error_context,
```
hauntsaninja, line 1063: Since we don't ever set any of these options, the defaults match the defaults of Options, and because pretty is False we never use read_source, maybe we can just pass in `Errors()`?
That way you also no longer need to worry about cached_read etc not existing

mypy/stubtest.py
```diff
@@ -1133,6 +1163,16 @@ def parse_options(args: List[str]) -> argparse.Namespace:
...
+    config_group = parser.add_argument_group(
+        title='Config file',
+        description="Use a config file instead of command line arguments. "
+                    "Plugins are the only supported configurations as of right "
+                    "now.",
+    )
+    config_group.add_argument(
+        '--config-file',
+        help="Configuration file, must have a [mypy] section "
+             "(defaults to {})".format(', '.join(defaults.CONFIG_FILES)))
```
hauntsaninja, line 1177: We don't actually set this default anywhere (and maybe we shouldn't?)

mypy/stubtest.py
```diff
@@ -1133,6 +1140,16 @@ def parse_options(args: List[str]) -> argparse.Namespace:
...
+    config_group = parser.add_argument_group(
+        title='mypy config file',
+        description="Use a config file instead of command line arguments. "
+                    "Plugins and mypy path are the only supported "
+                    "configurations.",
+    )
+    config_group.add_argument(
+        '--mypy-config-file',
+        help="Configuration file, must have a [mypy] section",
+    )
```
hauntsaninja, line 1152: minor: a wording suggestion
```suggestion
    config_group.add_argument(
        '--mypy-config-file',
        help="An existing mypy configuration file, currently used by stubtest to help determine mypy path and plugins",
    )
```

pikachu:
@hauntsaninja Could you elaborate on how this test would work? I spent some time today trying to get one running with `named_callable.py` which uses `get_function_hook`. I defined a test case that looked like:

```
def test_config_file(self) -> None:
    runtime = (
        "from typing import Callable\n"
        "def decorator1() -> Callable[..., Callable[..., int]]: pass\n"
        "def decorator2() -> Callable[..., Callable[..., int]]: pass\n"
        "@decorator1()\n"
        "def f() -> None: pass\n"
        "@decorator2()\n"
        "def g() -> None: pass\n"
        "reveal_type(f)\n"
    )
    stub = (
        "import builtins\n"
        "def f(*args: Any, **kwargs: Any) -> builtins.int: ...\n"
        "def g(*args: Any, **kwargs: Any) -> builtins.int: ...\n"
    )
    config_file = (
        "[mypy]\n"
        "plugins={}/test-data/unit/plugins/named_callable.py\n".format(root_dir)
    )
    output = run_stubtest(
        stub=stub,
        runtime=runtime,
        options=[],
        config_file=config_file,
    )
```

If you take a look at [named_callable.py](https://github.com/pikachu/mypy/blob/81ec44b0cd1030cef0d884281d872aac7cf7bc08/test-data/unit/plugins/named_callable.py#L10), it looks for a `fullname` of `m.decorator1`, and converts that function's return type to a function with a return type of `str`. Using the debugger, I found out that `get_function_hook` is not being called for the `decorator1` or `2` functions at all - it's only being called with `builtins.object` and `typing.runtime_checkable`. 

I tried similar approaches with `testTypeAnalyzeHookPlugin` (trying to modify the test case to fit the stubtest input format) to use the `get_type_analyze_hook` (as you mentioned), but still couldn't get it to work. 

Is there an example test case you could point me to? Thanks for all your help!

hauntsaninja:
stubtest uses mypy to build the stubs. The runtime is only imported and introspected using `inspect` and friends; stubtest doesn't use the runtime source code at all (and for C extensions, it isn't even available to us). You'll need to use the decorators you define in the stub.

hauntsaninja:
Thanks, looks good!

PR #9211, opened 2020-07-26
[mypyc] Introduce LoadMem

TH3CHARLie:
relates mypyc/mypyc#741

Introduce `LoadMem` IR to read a memory address and convert it into a designated-type value. (The address would mostly be in `py_ssize_t` for now)

Part of efforts to implement `len` primitives: `*(Py_ssize_t*)(ob + size_offset)`

### DIFF
--- a/mypyc/codegen/emitfunc.py
+++ b/mypyc/codegen/emitfunc.py
@@ -12,7 +12,7 @@ from mypyc.ir.ops import (
     LoadStatic, InitStatic, TupleGet, TupleSet, Call, IncRef, DecRef, Box, Cast, Unbox,
     BasicBlock, Value, MethodCall, PrimitiveOp, EmitterInterface, Unreachable, NAMESPACE_STATIC,
     NAMESPACE_TYPE, NAMESPACE_MODULE, RaiseStandardError, CallC, LoadGlobal, Truncate,
-    BinaryIntOp
+    BinaryIntOp, PtrDeref
 )
 from mypyc.ir.rtypes import (
     RType, RTuple, is_tagged, is_int32_rprimitive, is_int64_rprimitive
@@ -463,6 +463,13 @@ class FunctionEmitterVisitor(OpVisitor[None], EmitterInterface):
             rhs_cast = self.emit_unsigned_int_cast(op.rhs.type)
         self.emit_line('%s = %s%s %s %s%s;' % (dest, lhs_cast, lhs,
                                                op.op_str[op.op], rhs_cast, rhs))
+
+    def visit_ptr_deref(self, op: PtrDeref) -> None:
+        dest = self.reg(op)
+        src = self.reg(op.src)
+        # TODO: we shouldn't dereference to type that are pointer type so far
+        type = self.ctype(op.type)
+        self.emit_line('%s = *(%s*)(%s)' % (dest, type, src))
 
     # Helpers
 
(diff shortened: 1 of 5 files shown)


mypyc/ir/ops.py
```diff
@@ -1347,6 +1347,28 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
         return visitor.visit_binary_int_op(self)
 
 
+class PtrDeref(RegisterOp):
```
JukkaL, line 1350: For consistency with `LoadInt` and `LoadStatic`, what about calling this `LoadMem`? We can then explain this in terms of loading from memory, instead of pointer dereference. Pointer dereference sounds like a C concept, and for a low-level IR it's better to move closer to machine semantics.

mypyc/ir/ops.py
```diff
@@ -1347,6 +1347,28 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
...
+    def __init__(self, type: RType, src: Value, line: int = -1) -> None:
+        super().__init__(line)
+        self.type = type
+        self.src = src
+
+    def sources(self) -> List[Value]:
+        return [self.src]
+
+    def to_str(self, env: Environment) -> str:
+        return env.format("%r = ptr_deref %r :: %r*", self, self.src, self.type)
```
JukkaL, line 1366: Also rename the string representation if the op gets renamed.

mypyc/ir/ops.py
```diff
@@ -1347,6 +1347,28 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
...
+
+    def accept(self, visitor: 'OpVisitor[T]') -> T:
+        return visitor.visit_ptr_deref(self)
```
JukkaL, line 1369: Similarly for the visit method name.

mypyc/codegen/emitfunc.py
```diff
@@ -464,6 +464,13 @@ def visit_binary_int_op(self, op: BinaryIntOp) -> None:
         self.emit_line('%s = %s%s %s %s%s;' % (dest, lhs_cast, lhs,
                                                op.op_str[op.op], rhs_cast, rhs))
 
+    def visit_ptr_deref(self, op: PtrDeref) -> None:
+        dest = self.reg(op)
+        src = self.reg(op.src)
+        # TODO: we shouldn't dereference to type that are pointer type so far
+        type = self.ctype(op.type)
+        self.emit_line('%s = *(%s*)(%s)' % (dest, type, src))
```
JukkaL, line 472: The second set of parentheses are redundant?

TH3CHARLie:
Update:

I've added the assertion on `src` type and a codegen test.

mypyc/codegen/emitfunc.py
```diff
@@ -464,6 +464,13 @@ def visit_binary_int_op(self, op: BinaryIntOp) -> None:
         self.emit_line('%s = %s%s %s %s%s;' % (dest, lhs_cast, lhs,
                                                op.op_str[op.op], rhs_cast, rhs))
 
+    def visit_load_mem(self, op: LoadMem) -> None:
+        dest = self.reg(op)
+        src = self.reg(op.src)
+        # TODO: we shouldn't dereference to type that are pointer type so far
+        type = self.ctype(op.type)
+        self.emit_line('%s = *(%s*)%s;' % (dest, type, src))
```
JukkaL, line 472: A tiny nit: in C it's more common to add a space before `*` in a pointer type (e.g., `char *` instead of `char*`).

PR #9215, opened 2020-07-27
[mypyc] Introduce RStruct

TH3CHARLie:
relates #9211 

This PR introduces `RStruct` to represent CPython structures.

`StructInfo` is used to hold information about a unique struct-like type.

This PR also introduces several functions to compute aligned offsets and size of structure types.

### DIFF
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -515,8 +515,12 @@ def compute_aligned_offsets_and_size(types: List[RType]) -> Tuple[List[int], int
     return offsets, final_size
 
 
-class RStruct(RType):
-    """Represent CPython structs"""
+class StructInfo:
+    """Struct-like type Infomation
+
+    StructInfo should work with registry to ensure constraints like the unique naming
+    constraint for struct type
+    """
     def __init__(self,
                  name: str,
                  names: List[str],
@@ -529,7 +533,18 @@ class RStruct(RType):
             for i in range(len(self.types) - len(self.names)):
                 self.names.append('_item' + str(i))
         self.offsets, self.size = compute_aligned_offsets_and_size(types)
-        self._ctype = self.name
+
+
+class RStruct(RType):
+    """Represent CPython structs"""
+    def __init__(self,
+                 info: StructInfo) -> None:
+        self.name = info.name
+        self.names = info.names
+        self.types = info.types
+        self.offsets = info.offsets
+        self.size = info.size
+        self._ctype = info.name
 
     def accept(self, visitor: 'RTypeVisitor[T]') -> T:
         return visitor.visit_rstruct(self)
@@ -557,7 +572,8 @@ class RStruct(RType):
     @classmethod
     def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RStruct':
         types = [deserialize_type(t, ctx) for t in data['types']]
-        return RStruct(data['name'], data['names'], types)
+        info = StructInfo(data['name'], data['names'], types)
+        return RStruct(info)
 
 
 class RInstance(RType):
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -488,9 +488,8 @@ def compute_aligned_offsets_and_size(types: List[RType]) -> Tuple[List[int], int
     Note that the types argument are types of values that are stored
     sequentially with platform default alignment.
     """
-    platform_alignment = PLATFORM_SIZE
     unaligned_sizes = [compute_rtype_size(typ) for typ in types]
-    alignment = [compute_rtype_alignment(typ) for typ in types]
+    alignments = [compute_rtype_alignment(typ) for typ in types]
 
     current_offset = 0
     offsets = []
@@ -499,19 +498,15 @@ def compute_aligned_offsets_and_size(types: List[RType]) -> Tuple[List[int], int
         offsets.append(current_offset)
         if i + 1 < len(unaligned_sizes):
             cur_size = unaligned_sizes[i]
-            next_size = unaligned_sizes[i + 1]
-            if next_size > cur_size:
-                current_offset += cur_size
-                next_alignment = alignment[i + 1]
-                padding = (next_alignment - (current_offset % next_alignment)) % next_alignment
-                current_offset += padding
-            else:
-                current_offset += cur_size
+            current_offset += cur_size
+            next_alignment = alignments[i + 1]
+            # compute aligned offset,
+            # check https://en.wikipedia.org/wiki/Data_structure_alignment for more infomation
+            current_offset = (current_offset + (next_alignment - 1)) & -next_alignment
         else:
+            struct_alignment = max(alignments)
             final_size = current_offset + unaligned_sizes[i]
-            final_padding = ((platform_alignment - (final_size % platform_alignment))
-                             % platform_alignment)
-            final_size += final_padding
+            final_size = (final_size + (struct_alignment - 1)) & -struct_alignment
     return offsets, final_size
 
 
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -529,10 +529,6 @@ class StructInfo:
                 self.names.append('_item' + str(i))
         self.offsets, self.size = compute_aligned_offsets_and_size(types)
 
-    def __eq__(self, other: object) -> bool:
-        return (isinstance(other, StructInfo) and self.name == other.name
-                and self.names == other.names and self.types == other.types)
-
 
 class RStruct(RType):
     """Represent CPython structs"""
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -371,7 +371,8 @@ class TupleNameVisitor(RTypeVisitor[str]):
         return 'T{}{}'.format(len(parts), ''.join(parts))
 
     def visit_rstruct(self, t: 'RStruct') -> str:
-        return "S"
+        assert False
+        return ""
 
     def visit_rvoid(self, t: 'RVoid') -> str:
         assert False, "rvoid in tuple?"
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -529,17 +529,34 @@ class StructInfo:
                 self.names.append('_item' + str(i))
         self.offsets, self.size = compute_aligned_offsets_and_size(types)
 
+    def __eq__(self, other: object) -> bool:
+        return (isinstance(other, StructInfo) and self.name == other.name
+                and self.names == other.names and self.types == other.types)
+
 
 class RStruct(RType):
     """Represent CPython structs"""
     def __init__(self,
                  info: StructInfo) -> None:
-        self.name = info.name
-        self.names = info.names
-        self.types = info.types
-        self.offsets = info.offsets
-        self.size = info.size
-        self._ctype = info.name
+        self.info = info
+        self.name = self.info.name
+        self._ctype = self.info.name
+
+    @property
+    def names(self) -> List[str]:
+        return self.info.names
+
+    @property
+    def types(self) -> List[RType]:
+        return self.info.types
+
+    @property
+    def offsets(self) -> List[int]:
+        return self.info.offsets
+
+    @property
+    def size(self) -> int:
+        return self.info.size
 
     def accept(self, visitor: 'RTypeVisitor[T]') -> T:
         return visitor.visit_rstruct(self)
@@ -554,21 +571,17 @@ class RStruct(RType):
                                                           in zip(self.names, self.types)))
 
     def __eq__(self, other: object) -> bool:
-        return (isinstance(other, RStruct) and self.name == other.name
-                and self.names == other.names and self.types == other.types)
+        return isinstance(other, RStruct) and self.info == other.info
 
     def __hash__(self) -> int:
-        return hash((self.name, self.names, self.types))
+        return hash(self.info)
 
     def serialize(self) -> JsonDict:
-        types = [x.serialize() for x in self.types]
-        return {'.class': 'RStruct', 'name': self.name, 'names': self.names, 'types': types}
+        assert False
 
     @classmethod
     def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RStruct':
-        types = [deserialize_type(t, ctx) for t in data['types']]
-        info = StructInfo(data['name'], data['names'], types)
-        return RStruct(info)
+        assert False
 
 
 class RInstance(RType):
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -21,7 +21,7 @@ RTypes.
 """
 
 from abc import abstractmethod
-from typing import Optional, Union, List, Dict, Generic, TypeVar
+from typing import Optional, Union, List, Dict, Generic, TypeVar, Tuple
 
 from typing_extensions import Final, ClassVar, TYPE_CHECKING
 
@@ -443,28 +443,88 @@ dict_next_rtuple_single = RTuple(
 )
 
 
+def compute_rtype_size(typ: RType) -> int:
+    """Compute unaligned size of rtype"""
+    platform_size = 4 if IS_32_BIT_PLATFORM else 8
+    if isinstance(typ, RPrimitive):
+        if is_int32_rprimitive(typ):
+            return 4
+        elif is_int64_rprimitive(typ):
+            return 8
+        elif is_none_rprimitive(typ):
+            return 1
+        elif is_bool_rprimitive(typ):
+            return 1
+        else:
+            return platform_size
+    elif isinstance(typ, RTuple):
+        return compute_aligned_offsets_and_size(list(typ.types))[1]
+    elif isinstance(typ, RUnion):
+        return compute_aligned_offsets_and_size(typ.items)[1]
+    elif isinstance(typ, RStruct):
+        return compute_aligned_offsets_and_size(typ.types)[1]
+    elif isinstance(typ, RInstance):
+        return platform_size
+    else:
+        assert False, "invalid rtype for computing size"
+
+
+def compute_aligned_offsets_and_size(types: List[RType]) -> Tuple[List[int], int]:
+    """Compute offsets and total size of a list types after alignment"""
+    platform_alignment = 4 if IS_32_BIT_PLATFORM else 8
+    unaligned_sizes = [compute_rtype_size(typ) for typ in types]
+
+    current_offset = 0
+    offsets = []
+    final_size = 0
+    for i in range(len(unaligned_sizes)):
+        offsets.append(current_offset)
+        if i + 1 < len(unaligned_sizes):
+            cur_size = unaligned_sizes[i]
+            next_size = unaligned_sizes[i + 1]
+            if next_size > cur_size:
+                current_offset += cur_size
+                padding = (next_size - (current_offset % next_size)) % next_size
+                current_offset += padding
+            else:
+                current_offset += cur_size
+        else:
+            final_size = current_offset + unaligned_sizes[i]
+            final_padding = ((platform_alignment - (final_size % platform_alignment))
+                             % platform_alignment)
+            final_size += final_padding
+    return offsets, final_size
+
+
 class RStruct(RType):
-    """
-    """
+    """Represent CPython structs"""
     def __init__(self,
                  name: str,
                  names: List[str],
-                 types: List[RType],
-                 offsets: List[int]) -> None:
+                 types: List[RType]) -> None:
         self.name = name
         self.names = names
         self.types = types
-        self.offsets = offsets
+        self.offsets, self.size = compute_aligned_offsets_and_size(types)
         self._ctype = self.name
 
     def accept(self, visitor: 'RTypeVisitor[T]') -> T:
         return visitor.visit_rstruct(self)
 
     def __str__(self) -> str:
-        return '%s{%s}' % (self.name, ', '.join(str(typ) for typ in self.types))
+        # if not tuple(unamed structs)
+        if len(self.names) > 0:
+            return '%s{%s}' % (self.name, ', '.join(name + ":" + str(typ)
+                                                    for name, typ in zip(self.names, self.types)))
+        else:
+            return '%s{%s}' % (self.name, ', '.join(str(typ) for typ in self.types))
 
     def __repr__(self) -> str:
-        return '<RStruct %s{%s}>' % (self.name, ', '.join(repr(typ) for typ in self.types))
+        if len(self.names) > 0:
+            return '<RStruct %s{%s}>' % (self.name, ', '.join(name + ":" + repr(typ) for name, typ
+                                                              in zip(self.names, self.types)))
+        else:
+            return '<RStruct %s{%s}>' % (self.name, ', '.join(repr(typ) for typ in self.types))
 
     def __eq__(self, other: object) -> bool:
         return (isinstance(other, RStruct) and self.name == other.name
@@ -480,7 +540,7 @@ class RStruct(RType):
     @classmethod
     def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RStruct':
         types = [deserialize_type(t, ctx) for t in data['types']]
-        return RStruct(data['name'], data['names'], types, data['offsets'])
+        return RStruct(data['name'], data['names'], types)
 
 
 class RInstance(RType):
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -480,7 +480,7 @@ class RStruct(RType):
     @classmethod
     def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RStruct':
         types = [deserialize_type(t, ctx) for t in data['types']]
-        return RStruct(data['name'], data['names'], types)
+        return RStruct(data['name'], data['names'], types, data['offsets'])
 
 
 class RInstance(RType):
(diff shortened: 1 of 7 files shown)


TH3CHARLie:
I'd mark this as ready to review for some first-round suggestions @JukkaL , but this is not completed for the PR. I tried to mimic the usages of `RTuple` when implementing this `RStruct` but clearly most of them are not directly related to our current goal of using struct and pointer to handle some primitives. So probably guide me what's the essential part I've missing. Another thing that I've designed a prototype registry for structs, the `len` primitive case requires the knowledge of `PyVarObject` which is
```
typedef struct {
    PyObject ob_base;
    Py_ssize_t ob_size; /* Number of items in variable part */
} PyVarObject;
```
The PyObject has a marco `_PyObject_HEAD_EXTRA` which can be expanded to two pointers (for reference counting) or nothing. I am not sure what's the best way here. Also, we don't have the `PyTypeObject` for building the PyObject struct yet.

JukkaL:
> The PyObject has a marco _PyObject_HEAD_EXTRA which can be expanded to two pointers (for reference counting) or nothing. I am not sure what's the best way here.

We can assume that no non-default build options are enabled, which means that this macro will expand to nothing. It's sufficient to work with stock CPython builds.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +443,46 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
 
 
+class RStruct(RType):
+    """
+    """
+    def __init__(self,
+                 name: str,
+                 names: List[str],
+                 types: List[RType],
+                 offsets: List[int]) -> None:
```
JukkaL, line 453: The offsets can be inferred from types of fields. First, we can calculate the size of each simple RType easily (e.g. 32-bit integer -> 4 bytes). Second, we need to align each field based on platform alignment rules. The information in https://en.wikipedia.org/wiki/Data_structure_alignment should do the trick. 

Finally, we need to be able to calculate the alignment and the size of the struct, since structs can be placed inside other structs. The previous link will explain this as well.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +443,46 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+        self.names = names
+        self.types = types
+        self.offsets = offsets
+        self._ctype = self.name
+
+    def accept(self, visitor: 'RTypeVisitor[T]') -> T:
+        return visitor.visit_rstruct(self)
+
+    def __str__(self) -> str:
+        return '%s{%s}' % (self.name, ', '.join(str(typ) for typ in self.types))
```
JukkaL, line 464: Also include names of fields in the string representation.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +443,46 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+
+    def __repr__(self) -> str:
+        return '<RStruct %s{%s}>' % (self.name, ', '.join(repr(typ) for typ in self.types))
```
JukkaL, line 467: Include names of fields.

mypyc/rt_subtype.py
```diff
@@ -51,5 +51,12 @@ def visit_rtuple(self, left: RTuple) -> bool:
                 is_runtime_subtype(t1, t2) for t1, t2 in zip(left.types, self.right.types))
         return False
 
+    def visit_rstruct(self, left: RStruct) -> bool:
+        if isinstance(self.right, RStruct):
+            # TODO: should we consider names?
```
JukkaL, line 56: If we use this to allow assignments from one struct to another in generated C (I'm not sure but I think we might), we should consider names of both the struct and the fields, since C structs don't support structural subtyping.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +443,46 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
 )
 
 
+class RStruct(RType):
+    """
```
JukkaL, line 447: Also add docstring.

mypyc/ir/rtypes.py
```diff
@@ -443,28 +443,88 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
 )
 
 
+def compute_rtype_size(typ: RType) -> int:
```
JukkaL, line 446: We need a separate function for calculating alignment, since alignment is different from the size for struct types.

mypyc/ir/rtypes.py
```diff
@@ -443,28 +443,88 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+
+def compute_aligned_offsets_and_size(types: List[RType]) -> Tuple[List[int], int]:
+    """Compute offsets and total size of a list types after alignment"""
+    platform_alignment = 4 if IS_32_BIT_PLATFORM else 8
+    unaligned_sizes = [compute_rtype_size(typ) for typ in types]
+
+    current_offset = 0
+    offsets = []
+    final_size = 0
+    for i in range(len(unaligned_sizes)):
```
JukkaL, line 480: I think that this needs changes to work with nested structs. You should probably calculate the alignment of each item type using the new function that I'm proposing above.

mypyc/ir/rtypes.py
```diff
@@ -443,28 +443,88 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+    elif isinstance(typ, RStruct):
+        return compute_aligned_offsets_and_size(typ.types)[1]
+    elif isinstance(typ, RInstance):
+        return platform_size
+    else:
+        assert False, "invalid rtype for computing size"
+
+
+def compute_aligned_offsets_and_size(types: List[RType]) -> Tuple[List[int], int]:
+    """Compute offsets and total size of a list types after alignment"""
```
JukkaL, line 473: "list types" -> "list of types"? Make it explicit that the types argument are types of values that are stored sequentially with platform default alignment.

mypyc/test/test_struct.py
```diff
@@ -0,0 +1,33 @@
...
+class TestStruct(unittest.TestCase):
+    def test_struct_offsets(self) -> None:
+        # test per-member alignment
+        r1 = RStruct("", [], [bool_rprimitive, int32_rprimitive, int64_rprimitive])
+        assert r1.size == 16
+        assert r1.offsets == [0, 4, 8]
+
+        # test final alignment
+        r2 = RStruct("", [], [int32_rprimitive, bool_rprimitive])
+        r3 = RStruct("", [], [int64_rprimitive, bool_rprimitive])
```
JukkaL, line 18: Add another test for, say, three bools followed by an int32.

mypyc/test/test_struct.py
```diff
@@ -0,0 +1,33 @@
...
+            assert r3.size == 12
+        else:
+            assert r2.size == 8
+            assert r3.size == 16
+
+    def test_struct_str(self) -> None:
+        r = RStruct("Foo", ["a", "b"],
+                    [bool_rprimitive, object_rprimitive])
+        assert str(r) == "Foo{a:bool, b:object}"
+        assert repr(r) == "<RStruct Foo{a:<RPrimitive builtins.bool>, " \
```
JukkaL, line 32: Also test structs with no items and 1 item.

mypyc/ir/rtypes.py
```diff
@@ -364,6 +368,9 @@ def visit_rtuple(self, t: 'RTuple') -> str:
         parts = [elem.accept(self) for elem in t.types]
         return 'T{}{}'.format(len(parts), ''.join(parts))
 
+    def visit_rstruct(self, t: 'RStruct') -> str:
+        return "S"
```
JukkaL, line 374: This is incorrect. Since all structs are mapped to a single identifier, tuples with different types will get the same name. For now it's sufficient to add `assert False` here, to not support structs in tuples.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +445,123 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
 )
 
 
+def compute_rtype_alignment(typ: RType) -> int:
+    """Compute alignment of a given type based on platform alignment rule"""
+    platform_alignment = 4 if IS_32_BIT_PLATFORM else 8
```
JukkaL, line 450: `4 if IS_32_BIT_PLATFORM else 8` gets repeated a few times, so it's better to define it as a named constant.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +445,123 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+    if isinstance(typ, RPrimitive):
+        return typ.size
+    elif isinstance(typ, RInstance):
+        return platform_alignment
+    else:
+        if isinstance(typ, RTuple):
+            items = list(typ.types)
+        elif isinstance(typ, RStruct):
+            items = typ.types
+        elif isinstance(typ, RUnion):
```
JukkaL, line 460: Union types are always boxed, so they actually always use the platform alignment.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +445,123 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+
+def compute_rtype_size(typ: RType) -> int:
+    """Compute unaligned size of rtype"""
+    platform_size = 4 if IS_32_BIT_PLATFORM else 8
+    if isinstance(typ, RPrimitive):
+        return typ.size
+    elif isinstance(typ, RTuple):
+        return compute_aligned_offsets_and_size(list(typ.types))[1]
+    elif isinstance(typ, RUnion):
+        return compute_aligned_offsets_and_size(typ.items)[1]
```
JukkaL, line 476: This is always the pointer size, since unions are boxed.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +445,123 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+
+    current_offset = 0
+    offsets = []
+    final_size = 0
+    for i in range(len(unaligned_sizes)):
+        offsets.append(current_offset)
+        if i + 1 < len(unaligned_sizes):
+            cur_size = unaligned_sizes[i]
+            next_size = unaligned_sizes[i + 1]
+            if next_size > cur_size:
```
JukkaL, line 503: This check doesn't seem correct, at least for nested structs. Alignment and size may be unrelated (consider e.g. a struct with 27 bools).

mypyc/ir/rtypes.py
```diff
@@ -436,6 +445,123 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+                current_offset += cur_size
+                next_alignment = alignment[i + 1]
+                padding = (next_alignment - (current_offset % next_alignment)) % next_alignment
+                current_offset += padding
+            else:
+                current_offset += cur_size
```
JukkaL, line 509: Can you just remove this special case? Alignment will always be checked for.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +445,123 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+    final_size = 0
+    for i in range(len(unaligned_sizes)):
+        offsets.append(current_offset)
+        if i + 1 < len(unaligned_sizes):
+            cur_size = unaligned_sizes[i]
+            next_size = unaligned_sizes[i + 1]
+            if next_size > cur_size:
+                current_offset += cur_size
+                next_alignment = alignment[i + 1]
+                padding = (next_alignment - (current_offset % next_alignment)) % next_alignment
```
JukkaL, line 506: Minor optimization idea: alignment is always a power of 2, so we can use bitwise operations that are much faster than %.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +445,123 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+                current_offset += padding
+            else:
+                current_offset += cur_size
+        else:
+            final_size = current_offset + unaligned_sizes[i]
+            final_padding = ((platform_alignment - (final_size % platform_alignment))
+                             % platform_alignment)
```
JukkaL, line 513: This doesn't look right. Structs sizes don't need to be a multiple of platform alignment. Playing around with a C compiler, it looks like the size should be rounded up to a multiple of struct alignment.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +445,123 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+        self.types = types
+        self.offsets, self.size = compute_aligned_offsets_and_size(types)
+        self._ctype = self.name
+
+    def accept(self, visitor: 'RTypeVisitor[T]') -> T:
+        return visitor.visit_rstruct(self)
+
+    def __str__(self) -> str:
+        # if not tuple(unamed structs)
+        if len(self.names) > 0:
```
JukkaL, line 535: I think that we can always require names to be defined, for consistency. It's always possible to use some dummy names like `item0`, `item1`, etc. if there are no better names.

mypyc/primitives/struct_regsitry.py
```diff
@@ -0,0 +1,24 @@
...
+             names: List[str],
+             types: List[RType]) -> CStructDescription:
+    """Define a known C struct for generating IR to manipulate it
+
+    name: The name of the C struct
+    types: type of each field
+    names: name of each field
+           TODO: the names list can be empty in the future when we merge Tuple as part of Struct
+    """
+    return CStructDescription(name, names, types)
```
JukkaL, line 22: This doesn't actually add anything to a registry, it just constructs a description?

mypyc/rt_subtype.py
```diff
@@ -51,5 +51,12 @@ def visit_rtuple(self, left: RTuple) -> bool:
                 is_runtime_subtype(t1, t2) for t1, t2 in zip(left.types, self.right.types))
         return False
 
+    def visit_rstruct(self, left: RStruct) -> bool:
+        if isinstance(self.right, RStruct):
+            return len(self.right.types) == len(left.types) and all(
+                is_runtime_subtype(t1, t2) for t1, t2 in zip(left.types, self.right.types)) and (
+                self.right.names == left.names)
```
JukkaL, line 58: Also check that the names of the structs are the same.

mypyc/sametype.py
```diff
@@ -52,5 +52,11 @@ def visit_rtuple(self, left: RTuple) -> bool:
             and len(self.right.types) == len(left.types)
             and all(is_same_type(t1, t2) for t1, t2 in zip(left.types, self.right.types)))
 
+    def visit_rstruct(self, left: RStruct) -> bool:
+        return (isinstance(self.right, RStruct)
+            and self.right.names == left.names
```
JukkaL, line 57: On second thought, I think the best we can do it require that the names of the structs are identical, since in C two different structs can't be compatible with each other.

mypyc/rt_subtype.py
```diff
@@ -51,5 +51,12 @@ def visit_rtuple(self, left: RTuple) -> bool:
                 is_runtime_subtype(t1, t2) for t1, t2 in zip(left.types, self.right.types))
         return False
 
+    def visit_rstruct(self, left: RStruct) -> bool:
```
JukkaL, line 54: Add a unit test for this.

mypyc/subtype.py
```diff
@@ -56,5 +56,11 @@ def visit_rtuple(self, left: RTuple) -> bool:
                 is_subtype(t1, t2) for t1, t2 in zip(left.types, self.right.types))
         return False
 
+    def visit_rstruct(self, left: RStruct) -> bool:
+        if isinstance(self.right, RStruct):
+            return len(self.right.types) == len(left.types) and all(
+                is_subtype(t1, t2) for t1, t2 in zip(left.types, self.right.types))
```
JukkaL, line 62: A struct should only be compatible with the same struct, similar to above.

mypyc/test/test_struct.py
```diff
@@ -0,0 +1,52 @@
...
+
+class TestStruct(unittest.TestCase):
+    def test_struct_offsets(self) -> None:
+        # test per-member alignment
+        r1 = RStruct("", [], [bool_rprimitive, int32_rprimitive, int64_rprimitive])
+        assert r1.size == 16
+        assert r1.offsets == [0, 4, 8]
+
+        # test final alignment
+        r2 = RStruct("", [], [int32_rprimitive, bool_rprimitive])
```
JukkaL, line 17: Also test a struct with two bools.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +445,123 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+    """Represent CPython structs"""
+    def __init__(self,
+                 name: str,
+                 names: List[str],
+                 types: List[RType]) -> None:
+        self.name = name
+        self.names = names
+        self.types = types
+        self.offsets, self.size = compute_aligned_offsets_and_size(types)
+        self._ctype = self.name
```
JukkaL, line 528: On reflection, we should ensure that a struct with a given name has a unique definition, since we'll have a global namespace for structs. This doesn't enforce that constraint.

One way to do this would be to have a separate struct info field that `RStruct` refers to. So it would be defined like this:

```
class RStruct(RType):
    def __init__(self, info: StructInfo) -> None:
        ...
```

`StructInfo` would be similar to the current `RStruct`. We'd maintain a registry of `StructInfo` objects.

We can then later update `RTuple` to also contain a `StructInfo` reference. I'm starting to think that we'll continue to need a separate `RTuple` type, but it can share much of the implementation with `RStruct`.

mypyc/primitives/struct_regsitry.py
TH3CHARLie: yes, for now, I haven't figure out what's the best way to use the description, should we use it as a one-shot(like custom op) or add to a registry for matching. I think it's OK to postpone that to when we finally implement the `len` primitive

TH3CHARLie:
Thanks for the detailed review and thanks for picking out the bugs! I agree with most of them and only have a single minor confusion with the last point. The only reason we move the current struct definition into `StructInfo` is to enforce the naming constraint since we cannot ensure that within the constructor without extra complexity but introduced `StructInfo` can guarantee the unique naming simply with a registry, is this understanding correct?

JukkaL:
> The only reason we move the current struct definition into StructInfo is to enforce the naming constraint since we cannot ensure that within the constructor without extra complexity but introduced StructInfo can guarantee the unique naming simply with a registry, is this understanding correct?

That's a key point, though we could achieve a similar result by requiring that RStructs are not directly constructed but looked up via a registry. The other point is that this allows us to share StructInfo with `RTuple`.

TH3CHARLie:
I've addressed all the proposed changes except the unfold one, I'd like to add structinfo registry when we are finally implementing the len primitive.

mypyc/test/test_struct.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # test nested struct
+        info5 = StructInfo("", [], [bool_rprimitive, r])
+        r5 = RStruct(info5)
+        assert r5.offsets == [0, 8]
+        assert r5.size == 24
+        info6 = StructInfo("", [], [int32_rprimitive, r5])
+        r6 = RStruct(info6)
+        assert r6.offsets == [0, 8]
+        assert r6.size == 32
```
JukkaL, line 47: Test nested struct with smaller alignment than 8.

mypyc/test/test_struct.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def test_runtime_subtype(self) -> None:
+        # right type to check with
+        info = StructInfo("Foo", ["a", "b"],
+                    [bool_rprimitive, int_rprimitive])
+        r = RStruct(info)
+
+        # second type is subtype of r's second type
+        info1 = StructInfo("Foo", ["a", "b"],
+                    [bool_rprimitive, short_int_rprimitive])
```
JukkaL, line 73: Since we plan to ensure that for each name there will only be a single struct, this doesn't make sense as a test case. It's sufficient to test that a struct is a runtime subtype only of the same struct.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +446,131 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+class RStruct(RType):
+    """Represent CPython structs"""
+    def __init__(self,
+                 info: StructInfo) -> None:
+        self.name = info.name
+        self.names = info.names
+        self.types = info.types
+        self.offsets = info.offsets
+        self.size = info.size
+        self._ctype = info.name
```
JukkaL, line 542: This can be simplified by just holding a reference to `info` (e.g. `self.info = info`) and defining read-only properties corresponding to the info attributes. I.e., we'd only have a single `info` attribute.

This has a few benefits: 

* Serialized representation will be smaller. Currently we duplicate the entire structure of the struct on each serialization, which can be expensive for large structs that are used in many places.
* Constructing an object is faster. 
* We can compare struct equality quickly via comparing the identity of `info` objects (using `is`). 
* We 'd use a bit less memory.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +446,131 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+        # if not tuple(unamed structs)
+        return '%s{%s}' % (self.name, ', '.join(name + ":" + str(typ)
+                                                for name, typ in zip(self.names, self.types)))
+
+    def __repr__(self) -> str:
+        return '<RStruct %s{%s}>' % (self.name, ', '.join(name + ":" + repr(typ) for name, typ
+                                                          in zip(self.names, self.types)))
+
+    def __eq__(self, other: object) -> bool:
+        return (isinstance(other, RStruct) and self.name == other.name
```
JukkaL, line 557: It's sufficient to compare the identity of the `info` objects if you implement the change I'm suggesting above.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +446,131 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+                and self.names == other.names and self.types == other.types)
+
+    def __hash__(self) -> int:
+        return hash((self.name, self.names, self.types))
```
JukkaL, line 561: Again, if you implement my suggestion, it's enough to base the hash on the identity of `info`.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +446,131 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+
+    def serialize(self) -> JsonDict:
+        types = [x.serialize() for x in self.types]
+        return {'.class': 'RStruct', 'name': self.name, 'names': self.names, 'types': types}
```
JukkaL, line 565: With the above suggestion it's sufficient to serialize the name of the struct. However, this means that we'll need to serialize the struct registry somewhere else. It's okay to just `assert False` here for now, since we won't need to serialize struct types yet, I think.

mypyc/ir/rtypes.py
```diff
@@ -436,6 +446,131 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+
+    @classmethod
+    def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RStruct':
+        types = [deserialize_type(t, ctx) for t in data['types']]
+        info = StructInfo(data['name'], data['names'], types)
+        return RStruct(info)
```
JukkaL, line 571: You can leave this method out for now.

mypyc/rt_subtype.py
```diff
@@ -51,5 +51,12 @@ def visit_rtuple(self, left: RTuple) -> bool:
...
+        if isinstance(self.right, RStruct):
+            return (self.right.name == left.name and self.right.names == left.names and
+                    len(self.right.types) == len(left.types) and all(is_runtime_subtype(t1, t2)
+                    for t1, t2 in zip(left.types, self.right.types)))
```
JukkaL, line 58: I think that this should actually only compare equal if the other struct is the same as the first one, since there must only be one definition for each struct with a particular name, and C structs don't support structural compatibility.

mypyc/ir/rtypes.py
TH3CHARLie: Nice idea, although I tried to make all these attributes read-only properties, `name` and `_ctype` will report `Signature of "name" incompatible with supertype "RType"`

mypyc/ir/rtypes.py
```diff
@@ -436,6 +446,144 @@ def deserialize(cls, data: JsonDict, ctx: 'DeserMaps') -> 'RTuple':
...
+        self.types = types
+        # generate dummy names
+        if len(self.names) < len(self.types):
+            for i in range(len(self.types) - len(self.names)):
+                self.names.append('_item' + str(i))
+        self.offsets, self.size = compute_aligned_offsets_and_size(types)
+
+    def __eq__(self, other: object) -> bool:
+        return (isinstance(other, StructInfo) and self.name == other.name
+                and self.names == other.names and self.types == other.types)
```
JukkaL, line 534: There's no need to compare StructInfos for value equality, since they should be unique. Thus this doesn't seem to server a useful purpose, and can be misleading, so I'd rather remove it. (Also, if you define `__eq__`, you should also define `__hash__`, but this won't be relevant once this is removed.)

PR #9227, opened 2020-07-28
Update Travis CI OSX image to 10.3

TH3CHARLie:
xcode8.3 is pretty old, this PR updates OSX image to a more recent version 10.3


hauntsaninja:
Unclear to me why this caused things to break, but you might have some luck setting `TOXENV=py`

TH3CHARLie:
> Unclear to me why this caused things to break, but you might have some luck setting `TOXENV=py`

Thanks! this works!

TH3CHARLie:
We have the final builds are created in https://github.com/mypyc/mypy_mypyc-wheels/blob/master/.travis.yml
so maybe the xcode version we use for tests doesn't matter, as if something goes wrong, the wheel build will fail

PR #9237, opened 2020-07-31
Inform about no explicit export instead of attribute suggestions

### LINKED ISSUE #9084: "has no attribute" then suggests another missing attribute
Output in a mypy run:
```
vaults.py:14: error: Module 'datatypes' has no attribute 'VaultDataResponse'; maybe "VaultSummaryResponse"?  [attr-defined]
vaults.py:14: error: Module 'datatypes' has no attribute 'VaultSummaryResponse'; maybe "VaultDataResponse"?  [attr-defined]
```

Hopefully the error in the above output is self explanatory, it's contradicting itself.

kamilturek:
Fixes #9084

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1803,7 +1803,8 @@ class SemanticAnalyzer(NodeVisitor[None],
         module = self.modules.get(import_id)
         if module:
             if not self.options.implicit_reexport and source_id in module.names.keys():
-                message += "; implicit reexport disabled"
+                message = ("Module '{}' does not explicitly export attribute '{}'"
+                           "; implicit reexport disabled".format(import_id, source_id))
             else:
                 alternatives = set(module.names.keys()).difference({source_id})
                 matches = best_matches(source_id, alternatives)[:3]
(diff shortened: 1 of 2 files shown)


bluetech:
This is a helpful improvement.

The message in your first version of the patch is a bit more understandable than the second IMO (the "Append to message instead of overwriting" commit). Interested why you changed it.

kamilturek:
> This is a helpful improvement.
> 
> The message in your first version of the patch is a bit more understandable than the second IMO (the "Append to message instead of overwriting" commit). Interested why you changed it.

I didn't want to so radically change the message, but I may bring it back as it is more understable.

bluetech:
Yes, I can see this.

Perhaps the right thing to do it give it an entirely new error code, like `attr-exported`, instead of `attr-defined`. Not sure what is the policy regarding the granularity of these though.

JukkaL:
> Not sure what is the policy regarding the granularity of these though.

The policy on error codes is that if there's a realistic use case where having a separate error code would help, or if some error message can be confusing and providing more context via an error code (which can be looked up from the docs) can be helpful, splitting an error code may be reasonable.

kamilturek:
I have changed back the message and appended "implicit reexport disabled".

hauntsaninja:
Thanks!

PR #9243, opened 2020-08-02
messages: clarify bytes repr errors

### LINKED ISSUE #9236: Please include a "how to fix it" example in the error message

Feature request. A slightly more clear error message, which might be appreciated by folks newer to Python who aren't as familiar with the mini formatting language. If wanted, I'll contribute a pull request to fix this.

```
$ cat tmp.py
print("{}".format(b'abc'))
```

```
$ mypy tmp.py
tmp.py:1: error: On Python 3 '{}'.format(b'abc') produces "b'abc'"; use !r if this is a desired behavior
```

I'd love an example of what exactly to change. Something like:

```
$ mypy tmp.py
tmp.py:1: error: On Python 3 '{}'.format(b'abc') produces "b'abc'"; use !r if this is a desired behavior. Example: '{!r}'.format(b'abc'
```

```
$ python3 --version
Python 3.7.6
```

```
$ mypy --version
mypy 0.770
```

The error source code is still the same in Git master

No flags

  the full traceback below.

(You can freely edit this text, please remove all the lines
you believe are unnecessary.)

hauntsaninja:
Resolves #9236

### DIFF
--- a/mypy/checkstrformat.py
+++ b/mypy/checkstrformat.py
@@ -384,9 +384,10 @@ class StringFormatterChecker:
             if self.chk.options.python_version >= (3, 0):
                 if (has_type_component(actual_type, 'builtins.bytes') and
                         not custom_special_method(actual_type, '__str__')):
-                    self.msg.fail("On Python 3 '{}'.format(b'abc') produces \"b'abc'\";"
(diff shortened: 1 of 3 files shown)


JelleZijlstra:
Looks like some error messages in the tests need updating.

PR #9250, opened 2020-08-03
[very much a wip] pep 612

hauntsaninja:


JukkaL:
Thanks for working on this! This addresses a common user pain point.

I haven't had time to look at this in detail, but here are some general thoughts. The implementation doesn't need to be complete before merging -- maintaining a big PR over time gets tedious. Tests need to pass, of course. One option would be to add a command-line flag for enabling experimental features. Another option would be to document the feature as experimental until it's fully implemented and enable it always (this is what we've done previously).

PR #9251, opened 2020-08-03
[mypyc] Allow extra integer constant as the last argument to a C call

TH3CHARLie:
relates mypyc/mypyc#734 and mypyc/mypyc#753

### DIFF
--- a/mypyc/ir/ops.py
+++ b/mypyc/ir/ops.py
@@ -1173,8 +1173,7 @@ class CallC(RegisterOp):
                  steals: StealsDescription,
                  error_kind: int,
                  line: int,
-                 var_arg_idx: int = -1,
-                 extra_int_constant: Optional[int] = None) -> None:
+                 var_arg_idx: int = -1) -> None:
         self.error_kind = error_kind
         super().__init__(line)
         self.function_name = function_name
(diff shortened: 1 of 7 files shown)


mypyc/primitives/registry.py
```diff
@@ -354,6 +355,7 @@ def c_method_op(name: str,
                 var_arg_type: Optional[RType] = None,
                 truncated_type: Optional[RType] = None,
                 ordering: Optional[List[int]] = None,
+                extra_int_constant: Optional[int] = None,
```
JukkaL, line 358: Include also the type of the integer. For non-C backends in particular it will be useful to know whether it's a 32-bit or 64-bit integer, for example. Sometimes we'll also want to pass null pointers, and it would be nice to be able to specify their types. A reasonable way would be to make this argument `Optional[Tuple[int, RType]]`.

Update docstring to document this.

mypyc/ir/ops.py
```diff
@@ -1173,17 +1173,21 @@ def __init__(self,
                  steals: StealsDescription,
                  error_kind: int,
                  line: int,
-                 var_arg_idx: int = -1) -> None:
+                 var_arg_idx: int = -1,
+                 extra_int_constant: Optional[int] = None) -> None:
```
JukkaL, line 1177: It's unnecessary to modify `CallC`. It's better for IR builder to generate a `LoadInt` op for the extra argument.

mypyc/irbuild/ll_builder.py
```diff
@@ -762,7 +762,7 @@ def call_c(self,
                 arg = self.coerce(arg, desc.var_arg_type, line)
                 coerced.append(arg)
         target = self.add(CallC(desc.c_function_name, coerced, desc.return_type, desc.steals,
-                                desc.error_kind, line, var_arg_idx))
+                                desc.error_kind, line, var_arg_idx, desc.extra_int_constant))
```
JukkaL, line 765: As discussed in other comments, it would be better for consistency to create a `LoadInt` for the constant value and append it to `coerced`.

mypyc/ir/ops.py
TH3CHARLie: I like this approach!

TH3CHARLie:
should we include primitives using `NULL` as extra argument in this PR as well?

Issue #9147, opened 2020-07-13
`is_subtype` function does not work correctly for two overloads


sobolevn:
* Are you reporting a bug, or opening a feature request?

I am reporting a bug.

In mypy `curry` plugin I generate types like this for curried functions:

```python
from returns.curry import curry

@curry
def curried(a: int, b: int, c: str) -> float:
    ...

reveal_type(curried)  
# Revealed type is 'Overload(
#   def (a: builtins.int) -> Overload(
#     def (b: builtins.int, c: builtins.str) -> builtins.float, 
#     def (b: builtins.int) -> def (c: builtins.str) -> builtins.float
#   ), 
#   def (a: builtins.int, b: builtins.int) -> def (c: builtins.str) -> builtins.float, 
#   def (a: builtins.int, b: builtins.int, c: builtins.str) -> builtins.float,
# )'
```

When I try to manipulate this type in other places like:

```python
from returns.maybe import Maybe

Maybe.from_value(1).apply(Maybe.from_value(curried))
```

It raises an error:

```
Argument 1 to "Maybe" has incompatible type overloaded function; expected overloaded function
```

Original issue: https://github.com/dry-python/returns/issues/459

So, I have started from here: https://github.com/python/mypy/blob/358522e28cd58d95daf36256c46eb7ffcc55eea4/mypy/checkexpr.py#L1470-L1481

And checked what are the types I am dealing with:

```python
print('----')
print('--->', caller_type)
print('---<', callee_type)
print(is_subtype(caller_type, callee_type))
print(is_subtype(
                caller_type, callee_type,
                ignore_type_params=True,
                ignore_pos_arg_names=True,
                ignore_declared_variance=True,
                ignore_promotions=True))  # I have tried with all possible ignores set
print(str(caller_type) == str(callee_type))
print('----')
```

Outputs:

```
----
---> Overload(def (a: builtins.int) -> Overload(def (b: builtins.int, c: builtins.str) -> builtins.float, def (b: builtins.int) -> def (c: builtins.str) -> builtins.float), def (a: builtins.int, b: builtins.int) -> def (c: builtins.str) -> builtins.float, def (a: builtins.int, b: builtins.int, c: builtins.str) -> builtins.float)
[...]

Issue #8978, opened 2020-06-09
Error to infer types from a curried function with more than two args


thepabloaguilar:
This is possibly a **BUG**

I'm using [returns](https://returns.readthedocs.io/en/latest/) lib that provide fully typed functions, it has a [curry decorator](https://returns.readthedocs.io/en/latest/pages/curry.html#id3). I'm trying to use it, but `mypy` has got an error when we have a curried function with more than two args and try to create a [Result container](https://returns.readthedocs.io/en/latest/pages/result.html) with the function.

[Result.from_value(...) source code](https://returns.readthedocs.io/en/latest/_modules/returns/result.html#Result.from_value)
[@curry source code](https://returns.readthedocs.io/en/latest/_modules/returns/curry.html#curry)

Working example:
```python
from returns.curry import curry
from returns.result import Result

@curry
def first(a: int, b: int) -> int:
    ...

reveal_type(
    Result.from_value(first),
)
# Revealed type is 'returns.result.Result[Overload(def (a: builtins.int) -> def (b: builtins.int) -> builtins.int, def (a: builtins.int, b: builtins.int) -> builtins.int), Any]'
```

Nonworking example:
```python
from returns.curry import curry
from returns.result import Result

@curry
def first(a: int, b: int, c: int) -> int:
    ...

reveal_type(
    Result.from_value(first),
)
```
mypy output:
```sh
ex.py:9: note: Revealed type is 'returns.result.Result[Overload(def (a: builtins.int) -> Overload(def (b: builtins.int, c: builtins.int) -> builtins.int, def (b: builtins.int) -> def (c: builtins.int) -> builtins.int), def (a: builtins.int, b: builtins.int) -> def (c: builtins.int) -> builtins.int, def (a: builtins.int, b: builtins.int, c: builtins.int) -> builtins.int), Any]'
ex.py:9: error: Argument 1 to "from_value" of "Result" has incompatible type overloaded function; expected overloaded function
Found 1 error in 1 file (checked 1 source file)
```
---

**Versions**
* mypy: 0.770
* returns: 0.0.14
* Python: 3.8.2

My config file:
```cfg
[mypy]
# suppress errors about unsatisfied imports
ignore_missing_imports=True

# Custom plugins
plugins =
    returns.contrib.mypy.returns_plugin

# be strict
warn_return_any = True
strict_optional = True
warn_no_return = True
warn_redundant_casts = True
warn_unused_ignores = True
disallow_any_generics = True

[...]

PR #9259, opened 2020-08-04
Fixes corner case for comparing nested overloads

sobolevn:
We have this really nasty bug in `returns` with our `@curry` plugin:
- Closes https://github.com/python/mypy/issues/9147
- Closes https://github.com/python/mypy/issues/8978

So, this little fix solves the problem for us with very little effort.
In the ideal world, I would have to rewrite the whole thing to make the nested structures work.

Current problem is that we have a local state in `matched_overloads` and `possible_invalid_overloads` variables that is not syncronyzed with the nested calls.
I can also add a `TODO` note for later.

Refs https://github.com/dry-python/returns/issues/410

### DIFF
--- a/mypy/subtypes.py
+++ b/mypy/subtypes.py
@@ -399,6 +399,12 @@ class SubtypeVisitor(TypeVisitor[bool]):
                     return True
             return False
         elif isinstance(right, Overloaded):
+            if left == self.right:
+                # When it is the same overload, then the types are equal.
+                # This was originally introduced as a fix / hack for:
+                # https://github.com/python/mypy/issues/9147
+                return True
+
             # Ensure each overload in the right side (the supertype) is accounted for.
             previous_match_left_index = -1
             matched_overloads = set()
--- a/mypy/subtypes.py
+++ b/mypy/subtypes.py
@@ -401,8 +401,6 @@ class SubtypeVisitor(TypeVisitor[bool]):
         elif isinstance(right, Overloaded):
             if left == self.right:
                 # When it is the same overload, then the types are equal.
-                # This was originally introduced as a fix / hack for:
-                # https://github.com/python/mypy/issues/9147
                 return True
 
             # Ensure each overload in the right side (the supertype) is accounted for.


sobolevn:
Testing that this solves our issue:

```python
from returns.curry import curry
from returns.io import IO

@curry
def test(a: int, b: int, c: int) -> int:
    ...

reveal_type(IO(1).apply(IO(test)))
```


Without this fix:

```
» mypy ex.py
ex.py:8: note: Revealed type is 'returns.io.IO[Overload(def (b: builtins.int, c: builtins.int) -> builtins.int, def (b: builtins.int) -> def (c: builtins.int) -> builtins.int)]'
ex.py:8: error: Argument 1 to "IO" has incompatible type overloaded function; expected overloaded function
```

With this fix:

```
» python -m mypy /Users/sobolev/Documents/github/returns/ex.py --config-file=/Users/sobolev/Documents/github/returns/setup.cfg
/Users/sobolev/Documents/github/returns/ex.py:8: note: Revealed type is 'returns.io.IO[Overload(def (b: builtins.int, c: builtins.int) -> builtins.int, def (b: builtins.int) -> def (c: builtins.int) -> builtins.int)]'
```

It works! 🎉

sobolevn:
This failure does not seem related:

```
----------------------------- Captured stdout call -----------------------------

mypy: Failed to import, skipping

----------------------------- Captured stderr call -----------------------------

Inconsistency detected by ld.so: dl-open.c: 231: dl_open_worker: Assertion `_dl_debug_initialize (0, args->nsid)->r_state == RT_CONSISTENT' failed!
```

gvanrossum:
Let me close and reopen the issue to see if re-running the tests fares better.

mypy/subtypes.py
```diff
@@ -399,6 +399,12 @@ def visit_overloaded(self, left: Overloaded) -> bool:
                     return True
             return False
         elif isinstance(right, Overloaded):
+            if left == self.right:
+                # When it is the same overload, then the types are equal.
+                # This was originally introduced as a fix / hack for:
+                # https://github.com/python/mypy/issues/9147
```
gvanrossum, line 405: There is no need to add a comment in the source referring to the bug (people can find that information using `git blame`).

sobolevn: 👍

gvanrossum:
LGTM. Would it be feasible to add a unit test?

sobolevn:
@gvanrossum I don't know how to easily replicate this. Because creating a nested `Overloaded` is not possible without a plugin AFAIK. 

I can write one if that's required. But, if you ask me I would probably leave this as-is.

gvanrossum:
Okay, that's fine then.

sobolevn:
@gvanrossum thanks a lot! 👍

PR #9260, opened 2020-08-04
[mypyc] Introduce GetElementPtr

TH3CHARLie:
relates mypyc/mypyc#741

This PR introduces `GetElementPtr`, which computes the address of an element in an aggregate type.

Part of efforts to support the `len` and other macro-related primitives.

### DIFF
--- a/mypyc/codegen/emitfunc.py
+++ b/mypyc/codegen/emitfunc.py
@@ -12,10 +12,10 @@ from mypyc.ir.ops import (
     LoadStatic, InitStatic, TupleGet, TupleSet, Call, IncRef, DecRef, Box, Cast, Unbox,
     BasicBlock, Value, MethodCall, PrimitiveOp, EmitterInterface, Unreachable, NAMESPACE_STATIC,
     NAMESPACE_TYPE, NAMESPACE_MODULE, RaiseStandardError, CallC, LoadGlobal, Truncate,
-    BinaryIntOp, LoadMem
+    BinaryIntOp, LoadMem, GetElementPtr
 )
 from mypyc.ir.rtypes import (
-    RType, RTuple, is_tagged, is_int32_rprimitive, is_int64_rprimitive
+    RType, RTuple, is_tagged, is_int32_rprimitive, is_int64_rprimitive, RStruct
 )
 from mypyc.ir.func_ir import FuncIR, FuncDecl, FUNC_STATICMETHOD, FUNC_CLASSMETHOD
 from mypyc.ir.class_ir import ClassIR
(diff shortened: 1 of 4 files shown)


mypyc/ir/ops.py
```diff
@@ -1372,6 +1373,28 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
         return visitor.visit_load_mem(self)
 
 
+class GetElementPtr(RegisterOp):
+    """Get the address of a struct element"""
+    error_kind = ERR_NEVER
+
+    def __init__(self, src: Value, src_type: RType, index: int, line: int = -1) -> None:
```
JukkaL, line 1380: It would be more user friendly to take the name of the struct field instead of the index.

mypyc/ir/ops.py
```diff
@@ -1372,6 +1373,28 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
...
+        self.src = src
+        self.src_type = src_type
+        self.index = index
+
+    def sources(self) -> List[Value]:
+        return [self.src]
+
+    def to_str(self, env: Environment) -> str:
+        return env.format("%r = get_element_ptr %r %r :: %r", self, self.src,
+                          self.index, self.src_type)
```
JukkaL, line 1392: Display the name of the field instead of the integer index.

mypyc/codegen/emitfunc.py
```diff
@@ -471,6 +471,15 @@ def visit_load_mem(self, op: LoadMem) -> None:
...
         self.emit_line('%s = *(%s *)%s;' % (dest, type, src))
 
+    def visit_get_element_ptr(self, op: GetElementPtr) -> None:
+        dest = self.reg(op)
+        src = self.reg(op.src)
+        # TODO: support tuple type
+        assert isinstance(op.src_type, RStruct)
+        assert op.index < len(op.src_type.offsets), "Invalid element index."
+        offset = op.src_type.offsets[op.index]
+        self.emit_line('%s = (char *)%s + %s;' % (dest, src, offset))
```
JukkaL, line 481: Generate `<dest> = &<src>.<field_name>` instead (by default). This produces more readable C. 

We'll use the offsets in non-C backends. It's still important that we can calculate the offsets, but not for the C back end. 

To test the offset calculation for real, it would be good to have some configuration option to use pointer arithmetic in tests, even when using the C back end. You can do this in a separate PR to keep this one simpler.

mypyc/ir/ops.py
```diff
@@ -1372,6 +1373,28 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
         return visitor.visit_load_mem(self)
 
 
+class GetElementPtr(RegisterOp):
+    """Get the address of a struct element"""
+    error_kind = ERR_NEVER
+
+    def __init__(self, src: Value, src_type: RType, field: str, line: int = -1) -> None:
+        super().__init__(line)
+        self.type = c_pyssize_t_rprimitive
```
JukkaL, line 1382: Later on, this can be switched to a dedicated pointer rtype (which is still treated as an integer).

PR #9263, opened 2020-08-04
Use pytest Node.from_parent if available

llchan:
The pytest Node construction API has changed to prefer a `Node.from_parent` factory over the raw constructor [1].  This PR updates a few points to conform to the recommended API to avoid the `PytestDeprecationWarning` that arises when tests are being collected with pytest 6+.

[1] https://docs.pytest.org/en/stable/deprecations.html#node-construction-changed-to-node-from-parent

### DIFF
--- a/mypy/test/data.py
+++ b/mypy/test/data.py
@@ -10,18 +10,7 @@ from abc import abstractmethod
 import sys
 
 import pytest  # type: ignore  # no pytest in typeshed
-from typing import (
-    Any,
-    Callable,
-    Dict,
-    Iterator,
-    List,
-    NamedTuple,
-    Optional,
-    Set,
-    Tuple,
-    Union,
-)
+from typing import List, Tuple, Set, Optional, Iterator, Any, Dict, NamedTuple, Union
 
 from mypy.test.config import test_data_prefix, test_temp_dir, PREFIX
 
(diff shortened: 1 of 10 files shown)


llchan:
I have no objection to that, if everyone's on board with requiring a recent pytest for static analysis tests.  Would certainly keep this cleaner.

Also I force-pushed with a minor fix to shift a variable annotation to a type comment, since it looks like 3.5 is still part of the CI suite.

gvanrossum:
Okay, can you make it so? Hopefully pytest 5.4 still supports 3.5...

llchan:
Done.  Pushed it up as a separate commit, but we can squash before merging to keep the diff cleaner (I unwound some other misc changes to cancel out when we squash).

gvanrossum:
Thanks. Please don't squash anything, we do that on merging. Squashing earlier confuses the code review.

llchan:
Since we're already bumping test requirements from pytest 5.3.x to 5.4+, I figured I would try bumping up to the latest pytest 6.x.x (which is where I ran into the PytestDeprecationWarning becoming an error in the first place).  Most notably, this release adds types to the pytest package, which allows us to remove some type ignores in the mypy test code.  Waiting to see what CI looks like but if that's green and there are no objections, I think we can just bump this up for the additional type safety.

gvanrossum:
Awesome! I am anxiously awaiting the test results. :)

PR #9269, opened 2020-08-05
Reject bound covariant type variable in List argument

### LINKED ISSUE #734: Reject bound covariant type variable in an argument type
We should reject a bound covariant type variable in an argument type (i.e., in a contravariant position). For example, this should be rejected:

```
T = TypeVar('T', covariant=True)
class A(Generic[T]):
    def foo(self, x: T) -> None: ...  # Can't use as argument type! Return type is okay.
```

Similarly, a contravariant type variable should not be used in a return type.

The actual rule is probably a bit more involved than what I gave above, but this would probably cover most cases.

LiuYuHui:
Fixes #734

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -947,6 +947,14 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                             if ctx.line < 0:
                                 ctx = typ
                             self.fail(message_registry.FUNCTION_PARAMETER_CANNOT_BE_COVARIANT, ctx)
+                    elif is_named_instance(arg_type, 'builtins.list'):
+                        arg_typ = get_proper_type(arg_type)
+                        if isinstance(arg_typ, Instance):
+                            item_type = self.iterable_item_type(arg_typ)
+                            if (isinstance(item_type, TypeVarType)
+                                    and item_type.variance == COVARIANT):
+                                message = "Cannot use a covariant type variable as a parameter"
+                                self.fail(message, arg_type)
                     if typ.arg_kinds[i] == nodes.ARG_STAR:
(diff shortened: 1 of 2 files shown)


mypy/checker.py
```diff
@@ -947,6 +947,14 @@ def check_func_def(self, defn: FuncItem, typ: CallableType, name: Optional[str])
                             if ctx.line < 0:
                                 ctx = typ
                             self.fail(message_registry.FUNCTION_PARAMETER_CANNOT_BE_COVARIANT, ctx)
+                    elif is_named_instance(arg_type, 'builtins.list'):
```
sobolevn, line 950: What about `Dict` and (possibly) other types? 

```python
from typing import Dict, TypeVar

T_co = TypeVar('T_co', covariant=True)
T_contra = TypeVar('T_contra', contravariant=True)

def cov(a: Dict[T_co, T_co]) -> Dict[T_co, T_co]:
    ...

def contra(a: Dict[T_contra, T_contra]) -> Dict[T_contra, T_contra]:
    ...

reveal_type(cov({'a': 1}))
reveal_type(contra({'a': 1}))
```

Outputs:

```
ex.py:12: error: Cannot infer type argument 1 of "cov"
ex.py:12: note: Revealed type is 'builtins.dict[Any, Any]'
ex.py:13: error: Cannot infer type argument 1 of "contra"
ex.py:13: note: Revealed type is 'builtins.dict[Any, Any]'
```

msullivan:
I don't think we want to take a PR that does this only for List, and not other things. Happy to discuss how to generalize it.

(I'm closing the PR as part of trying to tidy up, but I'm happy to work with you on moving forward if you're still interested)

PR #9271, opened 2020-08-05
[mypyc] Implement builtins.len primitive for list

TH3CHARLie:
In this PR,

- `pointer_rprimitive` is introduced to represent pointer type
-  `PyVarObject` is registerd
- `size_t_to_short_int` custom op is added
- `GetElementPtr`'s codegen is **largely** changed.

I am marking this as a draft now and I expect tests to fail. Tests will be fixed once we decide the final design.


TH3CHARLie:
for this function:
```python
def f(a: List[int]) -> int:
    return len(a)
```

originally:
```c
CPyTagged CPyDef_f(PyObject *cpy_r_a) {
    CPyTagged cpy_r_r0;
    Py_ssize_t __tmp1;
CPyL0: ;
    __tmp1 = PyList_GET_SIZE(cpy_r_a);
    cpy_r_r0 = CPyTagged_ShortFromSsize_t(__tmp1);
    return cpy_r_r0;
}
```

now:
```c
CPyTagged CPyDef_f(PyObject *cpy_r_a) {
    int64_t cpy_r_r0;
    int64_t cpy_r_r1;
    CPyTagged cpy_r_r2;
CPyL0: ;
    cpy_r_r0 = (int64_t)&((PyVarObject *)cpy_r_a)->ob_size;
    cpy_r_r1 = *(int64_t *)cpy_r_r0;
    cpy_r_r2 = CPyTagged_ShortFromSsize_t(cpy_r_r1);
    return cpy_r_r2;
}
```

mypyc/ir/rtypes.py
```diff
@@ -254,6 +254,16 @@ def __repr__(self) -> str:
 else:
     c_pyssize_t_rprimitive = int64_rprimitive
 
+# low level pointer, represented as integer in C backends
+int32_ptr_rprimitive = RPrimitive('int32 *', is_unboxed=True, is_refcounted=False,
+                              ctype='int32_t *', size=4)  # type: Final
+int64_ptr_rprimitive = RPrimitive('int64 *', is_unboxed=True, is_refcounted=False,
+                              ctype='int64_t *', size=8)  # type: Final
```
JukkaL, line 261: I don't see the value of having the pointer target 32-bit integers on 32-bit platforms and 64-bit integers on 64-bit platforms. Since we also want to use integer operations on these (pointer arithmetic), the most reasonable choices for the type would be `char *` (for pointer arithmetic) or an unsigned integer type. 

I'm leaning towards using an unsigned integer type, so that we'll always have to explicitly cast to the desired pointer type in generated C. I think that this seems like a logical way to do this, as the pointers are untyped. Thus a plain pointer value should never be deferenced without a cast. (`void *` is not a good option since it doesn't support pointer arithmetic.)

So the pointer type could use `size_t` as the C type, similar to `CPyTagged`. However, `size_t` would look a bit strange in generated code, so we can add a typedef that resolves to it, for example, `CPyPtr`, defined in `mypyc_util.h` (similar to `CPyTagged`).

mypyc/irbuild/specialize.py
```diff
@@ -18,14 +18,17 @@
...
+    GetElementPtr
 )
 from mypyc.ir.rtypes import (
     RType, RTuple, str_rprimitive, list_rprimitive, dict_rprimitive, set_rprimitive,
-    bool_rprimitive, is_dict_rprimitive
+    bool_rprimitive, is_dict_rprimitive, is_list_rprimitive, c_pyssize_t_rprimitive
 )
 from mypyc.primitives.dict_ops import dict_keys_op, dict_values_op, dict_items_op
 from mypyc.primitives.misc_ops import true_op, false_op
+from mypyc.primitives.struct_regsitry import struct_types
```
JukkaL, line 30: Type in module name: `struct_regsitry` -> `struct_registry`

mypyc/primitives/struct_regsitry.py
```diff
@@ -1,24 +1,30 @@
...
+    typ = RStruct(info)
+    struct_types[name] = typ
+    return info
 
-# TODO: create PyVarObject, to do which we probably need PyObject
+
+c_struct(
+    name='PyVarObject',
+    names=['ob_refcnt', 'ob_type', 'ob_size'],
+    types=[c_pyssize_t_rprimitive, pointer_rprimitive, c_pyssize_t_rprimitive])
```
JukkaL, line 30: `PyVarObject` is defined like this in Python 3.7:

```c
typedef struct {
    PyObject ob_base;
    Py_ssize_t ob_size; /* Number of items in variable part */
} PyVarObject;
```

For consistency, we should define the `PyObject` struct and use `PyObject` as part of the definition of `PyVarObject` struct, instead of expanding the fields.

TH3CHARLie:
@JukkaL I find a potential problem when fixing the IR tests, though we implement `builtins.len` for list in the specializers but `list_len_op` is exported and used in some places directly. We should overcome that, I am thinking about adding a `list_len` to  LowLevelBuilder so that both specializer and other places can use it.

JukkaL:
>  I am thinking about adding a list_len to LowLevelBuilder so that both specializer and other places can use it. What do you think?

That's a good idea.

mypyc/ir/rtypes.py
```diff
@@ -182,7 +182,7 @@ def __init__(self,
         self.size = size
         # TODO: For low-level integers, they actually don't have undefined values
         #       we need to figure out some way to represent here.
-        if ctype in ('CPyTagged', 'int32_t', 'int64_t'):
+        if ctype in ('CPyTagged', 'int32_t', 'int64_t', 'CPyPtr'):
```
JukkaL, line 185: `CPY_INT_TAG` is kind of confusing here since the real empty value is a null pointer (`0`). Please change the initializer for `CPyPtr` to 0 (why not also change for `int32_t` and `int64_t` as well). `CPyTagged` should still have `CPY_INT_TAG`.

mypyc/ir/rtypes.py
```diff
@@ -571,10 +551,11 @@ def __repr__(self) -> str:
                                                           in zip(self.names, self.types)))
 
     def __eq__(self, other: object) -> bool:
-        return isinstance(other, RStruct) and self.info == other.info
+        return (isinstance(other, RStruct) and self.name == other.name
+                and self.names == other.names and self.types == other.types)
 
     def __hash__(self) -> int:
-        return hash(self.info)
+        return hash((self.name, self.names, self.types))
```
JukkaL, line 558: Python doesn't let you take a hash of a list. Convert to tuples, for example (e.g. `tuple(self.names)`). Add unit test for equality and hashing.

mypyc/irbuild/specialize.py
```diff
@@ -75,6 +78,11 @@ def translate_len(
             # though we still need to evaluate it.
             builder.accept(expr.args[0])
             return builder.add(LoadInt(len(expr_rtype.types)))
+        elif is_list_rprimitive(expr_rtype):
+            obj = builder.accept(expr.args[0])
+            elem_address = builder.add(GetElementPtr(obj, PyVarObject, 'ob_size'))
+            size_value = builder.add(LoadMem(c_pyssize_t_rprimitive, elem_address))
+            return builder.call_c(size_t_to_short_int, [size_value], -1)
```
JukkaL, line 85: We don't want to call a function here, since the reason we are doing this the hard way here instead of a primitive is to avoid a function call. Add a low-level IR builder helper for doing this. You can use `BinaryIntOp` to shift left by 1.

mypyc/primitives/int_ops.py
```diff
@@ -70,6 +70,12 @@
     error_kind=ERR_MAGIC,
     priority=3)
 
+size_t_to_short_int = c_custom_op(
+    arg_types=[c_pyssize_t_rprimitive],
+    return_type=short_int_rprimitive,
+    c_function_name='CPyTagged_ShortFromSsize_t',
+    error_kind=ERR_NEVER)
```
JukkaL, line 77: As mentioned above, we don't want to use this in this PR.

JukkaL: This can be removed, since it's unused.

TH3CHARLie: sure, I must have forgotten to do that.

TH3CHARLie:
I just updated that, I'll merge that once all tests go green.

Issue #9270, opened 2020-08-05
"raise from" type checking is broken if multiple except blocks


aviau:
This should not pass type checking:

```python

def test() -> None:
    try:
        raise Exception()
    except NotImplementedError as e:
        raise Exception("notimplemented") from e
    except Exception:
        raise Exception("test") from e

test()
```

``e`` does not exist on the second ``except`` block, so we cannot raise from it:
```
UnboundLocalError: local variable 'e' referenced before assignment
```

gvanrossum:
Oh, pretty subtle. I’m not even sure mypy has the internal data structures to keep track of this. Does it give you an error if you use a local variable after deleting it with ‘del’?

aviau:
Yes.

code:
```python
def test() -> None:
    a = "aa"
    del a
    print(a)

test()
```
mypy output:
````
test.py:4: error: Trying to read deleted variable 'a'
````

**Comment by isra17** [comment]
Note that this check successfully:

```
def test() -> None:
    a = "aa"
    del a
    raise Exception from a
test()
```

and 

```
def test() -> None:
    try:
        pass
    except NotImplementedError as e:
        pass
    except Exception:
        reveal_type(e)

test()
```

Does reveal `e` being deleted. So this might be due to `raise from` not checking if the source value is deleted?

**Comment by gvanrossum** [comment]
Oh, very interesting. Does the source for raise-from checking reveal anything interesting? For that matter, is the from clause type checked at all?

**Comment by aviau** [comment]
> is the from clause type checked at all?

I think yes.

The following code raises a mypy error:
```python
def test() -> None:
    try:
        raise Exception("test")
    except Exception as e:
        raise Exception("test2") from some_other_variable
````

```
test.py:5: error: Name 'some_other_variable' is not defined
```

Support this was apparently tracked in this [issue](https://github.com/python/mypy/issues/337) and merged in [this Pull Request](https://github.com/python/mypy/pull/463)

**Comment by isra17** [comment]
Had some fun looking at mypy internals for the first time. The checker only check for the source (And the exception too actually) to match the subtypes: https://github.com/python/mypy/blob/781caffa5c325b197e2c206a7322e671d2fee442/mypy/checker.py#L3332 , but the subtype see `DeletedType` as a subtype of everything.

This fix the issue:

```
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3317,6 +3317,9 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
[...]

PR #9272, opened 2020-08-05
check for deleted vars in 'raise from' (#9270)

aviau:
Thanks to @isra17

This fixes an issue where the following code would raise no error:

```python

def test() -> None:
    try:
        raise Exception()
    except NotImplementedError as e:
        raise Exception("notimplemented") from e
    except Exception:
        raise Exception("test") from e

test()
```

``e`` does not exist on the second ``except`` block, so we cannot raise from it. It would result in a runtime error:
```
UnboundLocalError: local variable 'e' referenced before assignment
```

See #9270 for details.


gvanrossum:
Thanks for the fix, @aviau!

PR #9275, opened 2020-08-07
plugins: delete open plugin

hauntsaninja:
Improvements to typeshed types using literals should have rendered this
unnecessary.


hauntsaninja:
Draft PR because we'll need https://github.com/python/typeshed/pull/4407 to be merged and synced before tests will pass.

test-data/stdlib-samples/3.2/test/test_genericpath.py
```diff
@@ -114,26 +114,21 @@ def test_getsize(self) -> None:
...
-            self.assertEqual(d, b"foobar")
-
-            self.assertLessEqual(
-                self.pathmodule.getctime(support.TESTFN),
-                self.pathmodule.getmtime(support.TESTFN)
-            )
-        finally:
-            if not f.closed:
-                f.close()
-            support.unlink(support.TESTFN)
```
gvanrossum, line 136: Is it safe to drop the finally clause with unlink?

hauntsaninja: It wouldn't, if this code were ever executed! We've talked about entirely getting rid of the stdlib samples before (e.g., in https://github.com/python/mypy/pull/8838), so I didn't try very hard to preserve them.

JukkaL:
> @JukkaL what do you think? Would it be bad for the Dropbox build?

It's unclear, but I wouldn't be surprised if this requires quite a few changes, as there are thousands of calls to `open`. I can test this against Dropbox internal repos once the build is green.

mypy/report.py
```diff
@@ -243,10 +243,10 @@ def _write_out_report(self,
             f.write(separator + '\n')
             for row_values in rows:
                 r = ("{:>{}}" * len(widths)).format(*itertools.chain(*zip(row_values, widths)))
-                f.writelines(r + '\n')
+                f.write(r + '\n')
```
JelleZijlstra, line 246: What made this change necessary?

gvanrossum: That looks like it fixes a bug! writelines takes a sequence of strings, not a single string (though because a string is a sequence of strings and writelines just writes the concatenation of the items it worked :-).

Though I'm not sure how the change found this bug -- does the TextIoWriter class (or whatever is used) not have writelines?

hauntsaninja: We now infer f as TextIOWrapper. TextIOWrapper and TextIOBase take `List[str]` for `writelines` in typeshed. Maybe that signature should be loosened though.

JelleZijlstra: Yes, I think that should be `Iterable[str]`. I just tried in 3.6 and TextIOWrapper.writelines accepts at least tuples and generator expressions.

hauntsaninja: And it looks like the perennial "str is a iterable of str" bug is benign in this case, so there's not really any downside to the looser signature.

hauntsaninja:
I wrote a script that democratises the ability to evaluate changes like this: https://github.com/hauntsaninja/mypy_primer

hauntsaninja:
Here are the results for this PR:

```
Fixed a false positive:
- zerver/lib/cache.py:97: error: Incompatible types in assignment (expression has type "TextIO", variable has type "TextIOWrapper")  [assignment]

Exposes an incorrect type in typeshed:
+ sphinx/ext/coverage.py:250:38: error: Argument 1 to "writelines" of "TextIOWrapper" has incompatible type "Generator[str, None, None]"; expected "List[str]"
+ sphinx/ext/coverage.py:279:46: error: Argument 1 to "writelines" of "TextIOWrapper" has incompatible type "Generator[str, None, None]"; expected "List[str]"
+ sphinx/ext/coverage.py:297:30: error: Argument 1 to "writelines" of "TextIOWrapper" has incompatible type "Generator[str, None, None]"; expected "List[str]"

Finds a "bug" (but also would go away by loosening the above type in typeshed):
+ mypy/report.py:246: error: Argument 1 to "writelines" of "TextIOWrapper" has incompatible type "str"; expected "List[str]"  [arg-type]
+ mypy/report.py:249: error: Argument 1 to "writelines" of "TextIOWrapper" has incompatible type "str"; expected "List[str]"  [arg-type]
+ tools/lib/provision_inner.py:79: error: Argument 1 to "writelines" of "TextIOWrapper" has incompatible type "str"; expected "List[str]"  [arg-type]
+ tools/lib/provision_inner.py:82: error: Argument 1 to "writelines" of "TextIOWrapper" has incompatible type "str"; expected "List[str]"  [arg-type]

Presumed false positive (although intentional on mypy's part, in the absence of --allow-redefinition):
+ src/pip/_internal/req/req_uninstall.py:644: error: Incompatible types in assignment (expression has type "BufferedWriter", variable has type "BufferedReader")
+ src/pip/_internal/operations/install/wheel.py:126: error: Incompatible types in assignment (expression has type "BufferedWriter", variable has type "BufferedReader")
+ mypyc/build.py:304: error: Incompatible types in assignment (expression has type "BufferedWriter", variable has type "BufferedReader")  [assignment]
+ tools/setup/generate_integration_bots_avatars.py:52: error: Incompatible types in assignment (expression has type "BufferedWriter", variable has type "BufferedReader")  [assignment]
+ pandas/io/stata.py:1063: error: Incompatible types in assignment (expression has type "BytesIO", variable has type "BufferedReader")  [assignment]
```

If we loosen the type of writelines in typeshed (as we should), this brings the count to five errors against and one error four out of a couple million lines of code. The five errors against are arguably "by design", since `--allow-redefinition` is not the default.

msullivan:
There's a competing "delete open plugin" from earlier: https://github.com/python/mypy/pull/7794.

Do you have any opinions on this one versus that, @hauntsaninja

hauntsaninja:
Yeah, I noticed the duplication a couple weeks ago when cleaning up some old PRs.

That PR no longer gets rid of the open plugin, because of https://github.com/python/mypy/pull/7794/files#discussion_r342519954
My opinion is that raising an error just for the `errors='replace'` + binary` case doesn't pull the extra weight of having a plugin.

It's also not clear to me what that PR doesn't necessitate some of the changes that I've made here. It's possible merging in a newer typeshed into that PR would surface those complaints (although I'll note that I changed the writelines type in typeshed in https://github.com/python/typeshed/pull/4642)

msullivan:
@JukkaL @ilevkivskyi what is the damage like for this PR at Dropbox?

hauntsaninja:
Issues that this would fix are reported fairly regularly. Given how long this PR has been lying around, and that the mypy_primer output doesn't look unworkable, I'm going to try asking for forgiveness on this merge :-)

PR #9286, opened 2020-08-11
Fix up mypyc to work for thirdparty users

untitaker:
This is the minimal set of changes I had to make to get mypyc working
for my own usecases:

* Identifiers ending with single quotation marks are not valid C
* A module was missing and caused import errors when executing mypyc

### DIFF
--- a/mypy/renaming.py
+++ b/mypy/renaming.py
@@ -265,7 +265,7 @@ class VariableRenameVisitor(TraverserVisitor):
 
     def rename_refs(self, names: List[NameExpr], index: int) -> None:
         name = names[0].name
-        new_name = name + "'" * (index + 1)
+        new_name = name + str(index)
         for expr in names:
             expr.name = new_name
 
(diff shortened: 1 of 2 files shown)


mypy/renaming.py
```diff
@@ -265,7 +265,7 @@ def flush_refs(self) -> None:
 
     def rename_refs(self, names: List[NameExpr], index: int) -> None:
         name = names[0].name
-        new_name = name + "'" * (index + 1)
+        new_name = name + str(index)
```
JukkaL, line 268: The new name may now conflict with an existing name. That's why we use a character that's not valid as an identifier.

Since we also generate names that aren't valid in C elsewhere, it would be better to have a centralized facility for dealing with problematic names within mypyc. If you want to pursue this, I can give some pointers.

untitaker: It seems easier to generate a hyphenless UUID, paste it into the sourcecode and use it as prefix for all `rename_refs` return values. I am not entirely sure how this can then work to get anything done for mypyc but I don't really have the motivation for this anyway, sorry.

PR #9287, opened 2020-08-11
[mypyc] Implement LoadAddress

TH3CHARLie:
This PR introduces `LoadAddress`, an op for taking the address of a given name. 

Currently, in mypyc we only need to take the address of a name when we are using the name_ref_ops, so the op takes a string(CPython name) as its argument. We can revisit the design later if we want to take the address of arbitrary name later.

### DIFF
--- a/mypyc/irbuild/expression.py
+++ b/mypyc/irbuild/expression.py
@@ -33,12 +33,12 @@ from mypyc.irbuild.builder import IRBuilder
 from mypyc.irbuild.for_helpers import translate_list_comprehension, comprehension_helper
 
 
-# Name and attribute references
 builtin_names = {
     'builtins.dict': (object_rprimitive, 'PyDict_Type')
 }  # type: Dict[str, Tuple[RType, str]]
 
 
+# Name and attribute references
 def transform_name_expr(builder: IRBuilder, expr: NameExpr) -> Value:
     assert expr.node, "RefExpr not resolved"
     fullname = expr.node.fullname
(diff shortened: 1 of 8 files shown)


TH3CHARLie:
Please note that there are no related IR tests in the suite using the `name_ref_op` for dict, we should probably introduce one.

mypyc/irbuild/expression.py
```diff
@@ -33,12 +33,18 @@
 from mypyc.irbuild.for_helpers import translate_list_comprehension, comprehension_helper
 
 
-# Name and attribute references
+builtin_names = {
+    'builtins.dict': (object_rprimitive, 'PyDict_Type')
+}  # type: Dict[str, Tuple[RType, str]]
 
 
+# Name and attribute references
```
JukkaL, line 41: Style nit: move the commend to inside the function or replace with a docstring.

mypyc/irbuild/expression.py
```diff
@@ -33,12 +33,18 @@
 from mypyc.irbuild.for_helpers import translate_list_comprehension, comprehension_helper
 
 
-# Name and attribute references
+builtin_names = {
+    'builtins.dict': (object_rprimitive, 'PyDict_Type')
+}  # type: Dict[str, Tuple[RType, str]]
```
JukkaL, line 38: Did you consider storing these inside a registry under `mypyc.primitives`? By declaring these there, we'd have the benefit of having most things related to a built-in type defined in a single place. The declaration of a name reference could be simplified like this, for example:

```py
load_address_op('builtins.dict', object_rprimitive, 'PyDict_Type')
```

Issue #9220, opened 2020-07-27
Incomplete or inconsistent Literal and Union behavior on conditional checks


kprzybyla:
Used mypy version: `0.782`

I stumbled upon some incomplete behavior related to `Literal` and `Union`. In most cases `Union` is not properly expanded upon conditional check when dealing with literals. Below example should describe clearly the situation.

```python
from typing import Union
from typing_extensions import Literal

class A:
    def __bool__(self) -> Literal[True]:
        return True

    @property
    def is_ok(self) -> Literal[True]:
        return True

    def ok(self) -> Literal[True]:
        return True

class B:
    def __bool__(self) -> Literal[False]:
        return False

    @property
    def is_ok(self) -> Literal[False]:
        return False

    def ok(self) -> Literal[False]:
        return False


def get_a_or_b() -> Union[A, B]: ...

thing = get_a_or_b()

if thing:
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'
else:
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'

if bool(thing):
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'
else:
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'

if thing.is_ok:
    reveal_type(thing)  # Revealed type is 'main.A'
else:
    reveal_type(thing)  # Revealed type is 'main.B'

if thing.ok():
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'
else:
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'
```

I would expect that all examples will give the same result as the one with `is_ok` property. Maybe the `bool()` example would require some additional thought but it would seem wise to make it also work. Also, walrus operator makes it so that even the `is_ok` property example does not work. Other examples behave the same as before:

```python
if thing := get_a_or_b():
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'
else:
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'

if bool(thing := get_a_or_b()):
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'
else:
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'

if (thing := get_a_or_b()).is_ok:
    reveal_type(thing)  # Revealed type is 'Union[main.A, main.B]'
else:
[...]

Issue #9229, opened 2020-07-29
Type fails to get "refined" on branch when `is None`-check happens in variable assignment


davidmrdavid:
* Are you reporting a bug, or opening a feature request? **Bug**
* Please insert below the code you are checking with mypy:

```python
from typing import Optional

def bar(var: str):
    pass

def foo(var: Optional[str]=None):
    var_not_None = not (var is None)
    if var_not_None:
        # var can only be a `str` at this point,
        # but `mypy` complains that
        # Argument 1 to "bar" has incompatible type "Optional[str]"; expected "str"
        bar(var)

def foo2(var: Optional[str]=None):
    if not(var is None):
        # no complaints from mypy
        bar(var)
```

* What is the actual behavior/output?

Running `mypy` on this reproducer returns:

`error: Argument 1 to "bar" has incompatible type "Optional[str]"; expected "str"
Found 1 error in 1 file (checked 1 source file)`

* What is the behavior/output you expect?

No errors, both `foo2` and `foo` are equivalent.

* What are the versions of mypy and Python you are using?

`mypy 0.782` and `Python 3.6.8`

* Do you see the same issue after installing mypy from Git master?

Yup, unless I did something wrong 😅 

**Summary:**

When you have an `Optional[str]`-typed variable `var`, one would expect that an `not (var is None)` check in an `if-statement` should tell `mypy` that all code within that branch has that the variable has type `str`. Indeed, `mypy` shows the expected behaviour for this, as shown in function `foo2` above.

However, there are _many_ instances where you want to check if a variable is `None` outside of the if-statement. For instance, if your if-statement were to check many boolean conditions at once. In this situation, one may save the result of the check into a variable. This is the case in the `foo` function above, where `mypy` complains.

I suspect this problem may not be solvable in the general case, but perhaps we can improve the status-quo nonetheless to catch this error.

If this is a beginner-friendly bug, I would love to help fix it! I've been meaning to get involved with the `mypy` project; just some guidance would do. Thanks!

Akuli:
using `:=` operator doesn't help here either

```
from typing import Optional
def foo(var: Optional[str]=None):
    if (var_not_None := not (var is None)):
        bar(var)
```

why is `var_not_None := bla` treated differently than plain `bla` here?

JukkaL:
This feature has been requested before, and we've closed the issues as too complicated to fix (or at least some of them). Since this is a somewhat common request, let's keep this open, but this looks low priority, since there may be no easy fix.

[...]

PR #9288, opened 2020-08-11
Improve type narrowing for walrus operator in conditional statements

kprzybyla:
This improves type narrowing for walrus operator (AssignmentExpr) in conditional statements. Information about the both assignment and condition will be now passed correctly in the if and else map to the ConditionalTypeBinder for type narrowing. Also, literal checks are now correctly interpreted for AssignmentExpr.

Partially fixes https://github.com/python/mypy/issues/9220 and https://github.com/python/mypy/issues/9229


JukkaL:
Thanks for the PR! This addresses some important issues. 

I can review this later this week -- but if somebody else wants to review this, please go ahead!

mypy/join.py
```diff
@@ -355,9 +355,15 @@ def default(self, typ: Type) -> ProperType:
 def join_instances(t: Instance, s: Instance) -> ProperType:
     """Calculate the join of two instance types."""
     if t.type == s.type:
+        # In some cases t will have empty mro list and
```
davidmrdavid, line 358: some minor typos here :)

mypy/checker.py
```diff
@@ -4143,7 +4143,26 @@ def has_no_custom_eq_checks(t: Type) -> bool:
...
+
+            if_condition_map, else_condition_map = self.find_isinstance_check_helper(node.value)
+
+            if if_condition_map is not None:
+                if_map.update(if_condition_map)
+
+            if else_condition_map is not None:
+                else_map.update(else_condition_map)
+
+            return if_map, else_map
```
JukkaL, line 4165: None values are special, but now we always return a non-`None` value. Should we sometimes still return a `None` value? The comment for `TypeMap` explains the distinction.

test-data/unit/check-python38.test
```diff
@@ -362,6 +362,36 @@ reveal_type(z2)  # E: Name 'z2' is not defined  # N: Revealed type is 'Any'
...
+else:
+    reveal_type(thing)  # N: Revealed type is '__main__.Bad'
+[builtins fixtures/property.pyi]
+
+[case testWalrusConditionalTypeCheck]
+from typing import Optional
+
+maybe_str: Optional[str]
+
+if (is_str := maybe_str is not None):
```
JukkaL, line 390: Does this PR affect simple cases, such as `if x := f()`? If yes, it's worth testing these separately.

What about `if (x := f()) is not None` (note that parentheses are different from this test case)?

JukkaL:
@kprzybyla Do you have time to respond to the feedback above? If you are busy, maybe somebody else could finish this PR.

kprzybyla:
@JukkaL Sorry, I had a really busy last week, I should be able to take care of this this week or maybe even today.

mypy/checker.py
kprzybyla: Yes, we should return `None` value in case when `condition_map` is `None`. Basically there are three possibilities here:
- `assignment_map is None` and `condition_map is None`
```python
def always_false() -> Literal[False]: ...

if x := always_false():
    reveal_type(x)  # E: Statement is unreachable
```
- `assignment_map is not None` and `condition_map is None`

```python
if (x := 0):
    reveal_type(x)  # E: Statement is unreachable
```
- `assignment_map is not None` and `condition_map is not None`
```python
if (is_str := maybe_str is not None):
    reveal_type(is_str)  # N: Revealed type is 'builtins.bool'
    reveal_type(maybe_str)  # N: Revealed type is 'builtins.str'
```

I couldn't find any use case where `assignment_map is None` and `condition_map is not None`. I guess that there isn't actually such scenarios where this condition would be true. I added test cases for all above cases and corrected the return statement here.

Also, I modified slightly the behavior of `find_isinstance_check_helper` method to have a default mechanism for creating type maps (which was basically cut and pasted from the `RefExpr` case). This was actually needed to make use cases in `testWalrusAssignmentAndConditionScopeForFunction` test case work properly.

test-data/unit/check-python38.test
kprzybyla: I added tests for both `if x := f()` and `if (x := f()) is not None` cases.

Also, I found two odd behaviors:

- Type narrowing without `--strict-optional` flags seems odd because it can narrows the `Union` outside of conditional statement even though it is not known which value the variable actually holds. With `--strict-optional` the last `reveal_type` in below example emits `Union[builtins.str, None]`.

```python
from typing import Optional

maybe_str: Optional[str]

if (x := maybe_str is not None):
    reveal_type(maybe_str)  # N: Revealed type is 'builtins.str'
else:
    reveal_type(maybe_str)  # N: Revealed type is 'None'

reveal_type(maybe_str)  # N: Revealed type is 'builtins.str'
```

- Below case is not handled properly and does not reflect the runtime behavior. Note that this is actually a different use case than the one I added in `testWalrusAssignmentAndConditionScopeForLiteral` test case because it does not fall into `is_true_literal` or `is_false_literal` checks in `find_isinstance_check_helper` method.

```python
if (x := 5):
    reveal_type(x)  # N: Revealed type is 'builtins.int'
else:
    reveal_type(x)  # E: Statement is unreachable

reveal_type(x)  # N: Revealed type is 'builtins.int'
```

gvanrossum: > if (x := maybe_str is not None):

Note that this is parsed as “x := (maybe_str is not None)“

kprzybyla:
@JukkaL Please tell me if there's anything else I should do or correct. Once you have time of course :smiley:

hauntsaninja (commented):
Thanks, and these new test cases are great!

The AssignmentExpr logic looks good to me, although technically I think if either if_assignment_map or if_condition_map is None, you'll want to return None. It's a little contrived, but e.g.:
```
x: None 
if x := returns_any():
   x = None 
```
doesn't get flagged currently by `--warn-unreachable`. (Maybe a little less contrived thanks to your previous PR, since mypy is now smarter about classes whose `__bool__` is a literal :-).

I don't have a good sense of potential consequences of now falling back to what was previously just the RefExpr case (having things under explicit if statements made reasoning about things somewhat easier for me), and you now have a small merge conflict, but otherwise looks good!

cazador481:
Any word on this?  I can fix the merge conflict and put it up for review if that helps.

hauntsaninja:
Not sure why Github didn't like that. I opened a new PR at #11202

PR #9294, opened 2020-08-12
[mypyc] Build True, False and None op in irbuild

TH3CHARLie:
This PR builds three specialized op directly in irbuild, instead of using name_ref_ops.

### DIFF
--- a/mypyc/irbuild/ll_builder.py
+++ b/mypyc/irbuild/ll_builder.py
@@ -28,7 +28,8 @@ from mypyc.ir.rtypes import (
     RType, RUnion, RInstance, optional_value_type, int_rprimitive, float_rprimitive,
     bool_rprimitive, list_rprimitive, str_rprimitive, is_none_rprimitive, object_rprimitive,
     c_pyssize_t_rprimitive, is_short_int_rprimitive, is_tagged, PyVarObject, short_int_rprimitive,
-    is_list_rprimitive, is_tuple_rprimitive, is_dict_rprimitive, is_set_rprimitive, PySetObject
+    is_list_rprimitive, is_tuple_rprimitive, is_dict_rprimitive, is_set_rprimitive, PySetObject,
+    none_rprimitive
 )
 from mypyc.ir.func_ir import FuncDecl, FuncSignature
 from mypyc.ir.class_ir import ClassIR, all_concrete_classes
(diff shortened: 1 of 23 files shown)


mypyc/test-data/irbuild-set.test
```diff
@@ -202,11 +202,12 @@ def update(s, x):
     s :: set
     x :: list
     r0 :: int32
-    r1, r2 :: None
+    r1 :: None
+    r2 :: None
```
TH3CHARLie, line 206: Although it's not something that important, I wonder how this happened.

JukkaL: This seems worth investigating. The underlying issue causing might be important.

mypyc/irbuild/ll_builder.py
```diff
@@ -424,7 +425,24 @@ def call_union_item(value: Value) -> Value:
 
     def none(self) -> Value:
         """Load unboxed None value (type: none_rprimitive)."""
-        return self.add(PrimitiveOp([], none_op, line=-1))
+        res = self.alloc_temp(none_rprimitive)
+        value = self.add(LoadInt(1, -1, bool_rprimitive))
```
JukkaL, line 429: Use `none_rprimitive` as the type?

mypyc/primitives/misc_ops.py
```diff
@@ -19,23 +19,23 @@
...
+# true_op = name_ref_op('builtins.True',
+#                       result_type=bool_rprimitive,
+#                       error_kind=ERR_NEVER,
+#                       emit=simple_emit('{dest} = 1;'))
+
+# # Get an unboxed False value
+# false_op = name_ref_op('builtins.False',
+#                        result_type=bool_rprimitive,
+#                        error_kind=ERR_NEVER,
+#                        emit=simple_emit('{dest} = 0;'))
```
JukkaL, line 38: Remove commented-out code.

mypyc/test-data/analysis.test
```diff
@@ -37,7 +37,7 @@ L4:
 L5:
     z = 2
 L6:
-    r5 = None
+    r5 = 1
     return r5
```
JukkaL, line 41: Could this be simplified to `return 1`, similar to how we simplify constant value returns in other contexts?

PR #9298, opened 2020-08-13
[mypyc] New style new_tuple_op

TH3CHARLie:
This PR supports new-style `new_tuple_op`, which completes the support of all tuple ops.

### DIFF
--- a/mypyc/irbuild/classdef.py
+++ b/mypyc/irbuild/classdef.py
@@ -8,11 +8,11 @@ from mypy.nodes import (
 )
 from mypyc.ir.ops import (
     Value, Call, LoadErrorValue, LoadStatic, InitStatic, TupleSet, SetAttr, Return,
-    BasicBlock, Branch, MethodCall, NAMESPACE_TYPE, LoadAddress
+    BasicBlock, Branch, MethodCall, NAMESPACE_TYPE, LoadAddress, LoadInt
 )
 from mypyc.ir.rtypes import (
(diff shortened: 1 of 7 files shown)


mypyc/test-data/irbuild-basic.test
```diff
@@ -1840,7 +1840,8 @@ def h():
     r2, r3 :: object
     r4, r5 :: dict
     r6 :: str
-    r7, r8 :: object
+    r7 :: object
+    r8 :: object
```
TH3CHARLie, line 1844: I have located the source cause of this problem. I'll submit a separate PR for that since it's going to change the IR tests in a lot of places(originally when we introduce const_int_regs into pretty IR printing, we didn't find out this happening)

mypyc/irbuild/classdef.py
```diff
@@ -165,7 +165,8 @@ def allocate_class(builder: IRBuilder, cdef: ClassDef) -> Value:
     base_exprs = cdef.base_type_exprs + cdef.removed_base_type_exprs
     if base_exprs:
         bases = [builder.accept(x) for x in base_exprs]
-        tp_bases = builder.primitive_op(new_tuple_op, bases, cdef.line)
+        load_size_op = builder.add(LoadInt(len(bases), -1, c_pyssize_t_rprimitive))
+        tp_bases = builder.call_c(new_tuple_op, [load_size_op] + bases, cdef.line)
```
JukkaL, line 169: Since these two lines are duplicated several times in this PR, it would be better to move these to a helper function and use it in all call sites (looks like there are 5 in this PR).

PR #9299, opened 2020-08-13
[mypyc] Don't free target of LoadMem too early

JukkaL:
Add optional reference to the object from which we are reading from
to `LoadMem` so that the object won't be freed before we read memory.

Fixes mypyc/mypyc#756.

### DIFF
--- a/mypyc/ir/ops.py
+++ b/mypyc/ir/ops.py
@@ -1348,25 +1348,42 @@ class BinaryIntOp(RegisterOp):
 
 
 class LoadMem(RegisterOp):
-    """Reading a memory location
+    """Read a memory location.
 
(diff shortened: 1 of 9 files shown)


JukkaL:
I checked that this fixes the crash with Dropbox internal repositories.

TH3CHARLie:
This commit caused https://github.com/mypyc/mypyc-benchmark-results/blob/master/reports/benchmarks/nqueens.md a 6.4% performance downgrade.

JukkaL:
> ...a 6.4% performance downgrade.

I will look at the generated code to see if there's anything suspicious that might have caused this. Since the benchmark is pretty small, 6.4% could also be just noise.

PR #9302, opened 2020-08-13
Don't infinite loop on self deps in --follow-imports=normal

msullivan:
There are situations in complex SCCs where the semantic analyzer will
infer self dependencies, which will cause an infinite loop in dmypy
--follow-imports=normal.

It's probably a bug that we do that, and that should be fixed too, but
fixing a graph algorithm to not infinite loop on self edges seems like
a reasonable thing to do in any case.

I don't have a minimized test case yet, and am submitting this without
one because it should be harmless and because I want it to get into
the release.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -557,6 +557,10 @@ class Server:
             if module[0] not in graph:
                 continue
             sources2 = self.direct_imports(module, graph)
+            # Filter anything already seen before. This prevents
+            # infinite looping if there are any self edges. (Self
+            # edges are maybe a bug, but...)
+            sources2 = [source for source in sources2 if source.module not in seen]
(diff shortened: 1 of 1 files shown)


mypy/dmypy_server.py
```diff
@@ -557,6 +557,10 @@ def fine_grained_increment_follow_imports(self, sources: List[BuildSource]) -> L
             if module[0] not in graph:
                 continue
             sources2 = self.direct_imports(module, graph)
+            # Filter anything already seen before. This prevents
+            # infinite looping if there are any self edges. (Self
+            # edges are maybe a bug, but...)
+            sources2 = [source for source in sources2 if source.module not in seen]
```
JukkaL, line 563: Would it be better to do this filtering in `direct_imports`?

PR #9303, opened 2020-08-14
[mypyc] Merge keep_propagating_op and remove unused assert_err_occurred_op

TH3CHARLie:
This PR merges keep_propagating_op and removes unused assert_err_occurred_op. To keep the design unified, we turn to C wrapper function for now.

This completes all exception-related ops.

### DIFF
--- a/mypyc/lib-rt/CPy.h
+++ b/mypyc/lib-rt/CPy.h
@@ -407,11 +407,7 @@ static int CPy_NoErrOccured(void) {
     return PyErr_Occurred() == NULL;
 }
 
-static void CPy_AssertErrOccurred(void) {
-    assert(PyErr_Occurred() != NULL && "failure w/o err!");
-}
-
-static bool CPy_KeepPropagating(void) {
+static inline bool CPy_KeepPropagating(void) {
     return 0;
 }
 // We want to avoid the public PyErr_GetExcInfo API for these because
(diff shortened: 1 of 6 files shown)


mypyc/lib-rt/CPy.h
```diff
@@ -407,6 +407,13 @@ static int CPy_NoErrOccured(void) {
     return PyErr_Occurred() == NULL;
 }
 
+static void CPy_AssertErrOccurred(void) {
+    assert(PyErr_Occurred() != NULL && "failure w/o err!");
+}
+
+static bool CPy_KeepPropagating(void) {
+    return 0;
+}
```
JukkaL, line 416: If you define a function in this header, they should be marked `static inline` (we don't do this consistently, but I'd like to make this better).

mypyc/primitives/exc_ops.py
```diff
@@ -44,21 +42,19 @@
     error_kind=ERR_FALSE)
 
 # Assert that the error indicator has been set.
-assert_err_occured_op = custom_op(
+assert_err_occurred_op = c_custom_op(
```
JukkaL, line 45: This seems to be unused, so it can actually be removed.

JukkaL:
I approved this with comments, which means that you can merge this yourself after you've made the changes and the build is green, without waiting for another review pass.

mypyc/primitives/exc_ops.py
TH3CHARLie: yeah, that's actually what I prepared to discuss today. Then I'll just simply remove this.

PR #9310, opened 2020-08-14
[mypyc] Obsolete several old-style registry functions

TH3CHARLie:
This PR merges the generic `in` and `op` as well as misc `builtins.bool` op, and obsoletes unused old-style registry functions: `call_void_emit`, `call_and_fail_emit`, `call_negative_bool_emit`, `negative_int_emit`, `call_negative_magic_emit`.

A test for generic `in` and misc `builtins.bool` is added.

### DIFF
--- a/mypyc/irbuild/ll_builder.py
+++ b/mypyc/irbuild/ll_builder.py
@@ -623,12 +623,12 @@ class LowLevelIRBuilder:
                  lreg: Value,
                  expr_op: str,
                  line: int) -> Value:
-        call_c_ops_candidates = c_unary_ops.get(expr_op, [])
-        target = self.matching_call_c(call_c_ops_candidates, [lreg], line)
-        if target:
-            return target
         ops = unary_ops.get(expr_op, [])
         target = self.matching_primitive_op(ops, [lreg], line)
+        if target:
+            return target
+        call_c_ops_candidates = c_unary_ops.get(expr_op, [])
+        target = self.matching_call_c(call_c_ops_candidates, [lreg], line)
         assert target, 'Unsupported unary operation: %s' % expr_op
         return target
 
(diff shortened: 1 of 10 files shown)


mypyc/irbuild/ll_builder.py
```diff
@@ -623,12 +623,12 @@ def unary_op(self,
...
-        call_c_ops_candidates = c_unary_ops.get(expr_op, [])
-        target = self.matching_call_c(call_c_ops_candidates, [lreg], line)
-        if target:
-            return target
         ops = unary_ops.get(expr_op, [])
         target = self.matching_primitive_op(ops, [lreg], line)
+        if target:
+            return target
+        call_c_ops_candidates = c_unary_ops.get(expr_op, [])
+        target = self.matching_call_c(call_c_ops_candidates, [lreg], line)
```
TH3CHARLie, line 631: this adjustment is to ensure the only left old-style unary op(boolean `not`) is always used first then the generic one. 

The boolean `not` should eventually be built using a `ComparsionOp`(or maybe `UnaryComparisonOp` so we differentiate the unary/binary property?) Then this hack can be removed.

JukkaL: Unary comparison op doesn't seem worth it.

TH3CHARLie: > Unary comparison op doesn't seem worth it.

Then probably a single `ComparisonOp` class for all ops producing boolean, with arg_types and op_type to ensure each operator gets the exact number of argument it needs?

JukkaL: I meant that having binary comparison operations seems sufficient to me. The unary operations can be built using the binary operations.

TH3CHARLie: I see, a `NOR` operation should be sufficient to represent the `not` here.

PR #9313, opened 2020-08-15
[mypyc] Split BinaryIntOp and introduce ComparisonOp, implements is/is not op

TH3CHARLie:
`BinaryIntOp` used to represent arithmetic, bitwise and comparison operations on integer operations. However, this design prevents us to compare pointer types and all comparison operations should be of boolean return type while manually specifying this in `BinaryIntOp` is both verbose and error-prone.

This PR splits `BinaryIntOp` and moves the comparison functionalities to `ComparsionOp`.

Based on the new op, this PR also implements `is` and `is not` op.

### DIFF
--- a/mypyc/ir/ops.py
+++ b/mypyc/ir/ops.py
@@ -1316,7 +1316,7 @@ class BinaryIntOp(RegisterOp):
 
 
 class ComparisonOp(RegisterOp):
-    """Binary Comparsion ops
+    """Comparison ops
 
     The result type will always be boolean.
 
--- a/mypyc/ir/ops.py
+++ b/mypyc/ir/ops.py
@@ -1263,7 +1263,7 @@ class LoadGlobal(RegisterOp):
 
 
 class BinaryIntOp(RegisterOp):
-    """Binary operations on integer types
+    """Binary arithmetic and bitwise operations on integer types
 
     These ops are low-level and will be eventually generated to simple x op y form.
     The left and right values should be of low-level integer types that support those ops
@@ -1276,18 +1276,7 @@ class BinaryIntOp(RegisterOp):
     MUL = 2  # type: Final
     DIV = 3  # type: Final
     MOD = 4  # type: Final
-    # logical
-    # S for signed and U for unsigned
-    EQ = 100  # type: Final
-    NEQ = 101  # type: Final
-    SLT = 102  # type: Final
-    SGT = 103  # type: Final
-    SLE = 104  # type: Final
-    SGE = 105  # type: Final
-    ULT = 106  # type: Final
-    UGT = 107  # type: Final
-    ULE = 108  # type: Final
-    UGE = 109  # type: Final
+
     # bitwise
     AND = 200  # type: Final
     OR = 201  # type: Final
@@ -1301,16 +1290,6 @@ class BinaryIntOp(RegisterOp):
         MUL: '*',
         DIV: '/',
         MOD: '%',
-        EQ: '==',
-        NEQ: '!=',
-        SLT: '<',
-        SGT: '>',
-        SLE: '<=',
-        SGE: '>=',
-        ULT: '<',
-        UGT: '>',
-        ULE: '<=',
-        UGE: '>=',
         AND: '&',
         OR: '|',
         XOR: '^',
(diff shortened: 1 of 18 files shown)


TH3CHARLie:
travis is abnormally slow today...

JukkaL:
There are merge conflicts. Also, I can't see the definition of `ComparisonOp` in this PR. The first commit is somehow missing from this PR.

mypyc/ir/ops.py
```diff
@@ -1301,6 +1290,53 @@ class BinaryIntOp(RegisterOp):
...
+    def to_str(self, env: Environment) -> str:
+        return env.format('%r = %r %s %r', self, self.lhs,
+                          self.op_str[self.op], self.rhs)
+
+    def accept(self, visitor: 'OpVisitor[T]') -> T:
+        return visitor.visit_binary_int_op(self)
+
+
+class ComparisonOp(RegisterOp):
+    """Binary Comparsion ops
```
JukkaL, line 1319: Typo: Comparsion.

Also I'd suggest remove the word 'binary' as redundant, since comparisons are always binary operations.

TH3CHARLie:
@JukkaL Looks like the github review fail to represent to diff clearly, so here's the definition of two classes:
```python
class ComparisonOp(RegisterOp):
    """Comparison ops

    The result type will always be boolean.

    Support comparison between integer types and pointer types
    """
    error_kind = ERR_NEVER

    # S for signed and U for unsigned
    EQ = 100  # type: Final
    NEQ = 101  # type: Final
    SLT = 102  # type: Final
    SGT = 103  # type: Final
    SLE = 104  # type: Final
    SGE = 105  # type: Final
    ULT = 106  # type: Final
    UGT = 107  # type: Final
    ULE = 108  # type: Final
    UGE = 109  # type: Final

    op_str = {
        EQ: '==',
        NEQ: '!=',
        SLT: '<',
        SGT: '>',
        SLE: '<=',
        SGE: '>=',
        ULT: '<',
        UGT: '>',
        ULE: '<=',
        UGE: '>=',
    }  # type: Final

    def __init__(self, lhs: Value, rhs: Value, op: int, line: int = -1) -> None:
        super().__init__(line)
        self.type = bool_rprimitive
        self.lhs = lhs
        self.rhs = rhs
        self.op = op

    def sources(self) -> List[Value]:
        return [self.lhs, self.rhs]

    def to_str(self, env: Environment) -> str:
        if self.op in (self.SLT, self.SGT, self.SLE, self.SGE):
            sign_format = " :: signed"
        elif self.op in (self.ULT, self.UGT, self.ULE, self.UGE):
            sign_format = " :: unsigned"
        else:
            sign_format = ""
        return env.format('%r = %r %s %r%s', self, self.lhs,
                          self.op_str[self.op], self.rhs, sign_format)

    def accept(self, visitor: 'OpVisitor[T]') -> T:
        return visitor.visit_comparison_op(self)
```


```python
class BinaryIntOp(RegisterOp):
    """Binary arithmetic and bitwise operations on integer types

    These ops are low-level and will be eventually generated to simple x op y form.
    The left and right values should be of low-level integer types that support those ops
    """
    error_kind = ERR_NEVER

    # arithmetic
    ADD = 0  # type: Final
    SUB = 1  # type: Final
    MUL = 2  # type: Final
    DIV = 3  # type: Final
    MOD = 4  # type: Final

    # bitwise
    AND = 200  # type: Final
    OR = 201  # type: Final
    XOR = 202  # type: Final
    LEFT_SHIFT = 203  # type: Final
    RIGHT_SHIFT = 204  # type: Final

    op_str = {
        ADD: '+',
        SUB: '-',
        MUL: '*',
        DIV: '/',
        MOD: '%',
        AND: '&',
        OR: '|',
        XOR: '^',
        LEFT_SHIFT: '<<',
        RIGHT_SHIFT: '>>',
    }  # type: Final

    def __init__(self, type: RType, lhs: Value, rhs: Value, op: int, line: int = -1) -> None:
        super().__init__(line)
        self.type = type
        self.lhs = lhs
        self.rhs = rhs
        self.op = op

    def sources(self) -> List[Value]:
        return [self.lhs, self.rhs]

    def to_str(self, env: Environment) -> str:
        return env.format('%r = %r %s %r', self, self.lhs,
                          self.op_str[self.op], self.rhs)

    def accept(self, visitor: 'OpVisitor[T]') -> T:
        return visitor.visit_binary_int_op(self)
```

mypyc/irbuild/ll_builder.py
```diff
@@ -984,6 +990,21 @@ def translate_eq_cmp(self,
             line
         )
 
+    def translate_is_op(self,
+                        lreg: Value,
+                        rreg: Value,
+                        expr_op: str,
+                        line: int) -> Value:
+        """Create equality comparison operation between object types
```
JukkaL, line 998: "object types" -> "object identities"

mypyc/test/test_emitfunc.py
```diff
@@ -247,21 +247,51 @@ def test_dict_contains(self) -> None:
             """cpy_r_r0 = PyDict_Contains(cpy_r_d, cpy_r_o);""")
 
     def test_binary_int_op(self) -> None:
+        self.assert_emit(BinaryIntOp(short_int_rprimitive, self.s1, self.s2, BinaryIntOp.ADD, 1),
```
JukkaL, line 250: Thanks for adding more test coverage!

PR #9316, opened 2020-08-16
[mypyc] Merge generic call ops

TH3CHARLie:
This PR merges three generic call ops.

### DIFF
--- a/mypyc/irbuild/ll_builder.py
+++ b/mypyc/irbuild/ll_builder.py
@@ -241,7 +241,7 @@ class LowLevelIRBuilder:
         """
         # If all arguments are positional, we can use py_call_op.
         if (arg_kinds is None) or all(kind == ARG_POS for kind in arg_kinds):
-            return self.primitive_op(py_call_op, [function] + arg_values, line)
+            return self.call_c(py_call_op, [function] + arg_values, line)
 
         # Otherwise fallback to py_call_with_kwargs_op.
         assert arg_names is not None
@@ -278,7 +278,7 @@ class LowLevelIRBuilder:
 
         kw_args_dict = self.make_dict(kw_arg_key_value_pairs, line)
 
-        return self.primitive_op(
+        return self.call_c(
             py_call_with_kwargs_op, [function, pos_args_tuple, kw_args_dict], line)
 
     def py_method_call(self,
@@ -291,7 +291,7 @@ class LowLevelIRBuilder:
         """Call a Python method (non-native and slow)."""
         if (arg_kinds is None) or all(kind == ARG_POS for kind in arg_kinds):
             method_name_reg = self.load_static_unicode(method_name)
-            return self.primitive_op(py_method_call_op, [obj, method_name_reg] + arg_values, line)
+            return self.call_c(py_method_call_op, [obj, method_name_reg] + arg_values, line)
         else:
             method = self.py_get_attr(obj, method_name, line)
             return self.py_call(method, arg_values, line, arg_kinds=arg_kinds, arg_names=arg_names)
--- a/mypyc/primitives/generic_ops.py
+++ b/mypyc/primitives/generic_ops.py
@@ -12,7 +12,7 @@ check that the priorities are configured properly.
 from mypyc.ir.ops import ERR_NEVER, ERR_MAGIC, ERR_NEG_INT
 from mypyc.ir.rtypes import object_rprimitive, int_rprimitive, bool_rprimitive, c_int_rprimitive
 from mypyc.primitives.registry import (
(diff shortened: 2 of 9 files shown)


TH3CHARLie:
Ummm, this is not the first time merging this call ops leading to segfault on mypy tests using the compiled version, log here:
https://travis-ci.org/github/python/mypy/jobs/718439274

However, looking at both the definition of registrys and the generated C code, I can't find anything suspicious

mypyc/primitives/generic_ops.py
```diff
@@ -200,32 +200,31 @@
...
-    is_var_arg=True,
+py_method_call_op = c_custom_op(
+    arg_types=[],
+    return_type=object_rprimitive,
+    c_function_name='CPyObject_CallMethodObjArgs',
     error_kind=ERR_MAGIC,
-    format_str='{dest} = py_method_call({comma_args})',
-    emit=simple_emit('{dest} = CPyObject_CallMethodObjArgs({comma_args}, NULL);'))
+    var_arg_type=object_rprimitive,
+    extra_int_constant=(0, c_int_rprimitive))
```
JukkaL, line 227: It would be better for `NULL` to be represented as the pointer type, or at least an unsigned integer, not `c_int`.

mypyc/primitives/generic_ops.py
```diff
@@ -200,32 +200,31 @@
...
-    is_var_arg=True,
+py_call_op = c_custom_op(
+    arg_types=[],
+    return_type=object_rprimitive,
+    c_function_name='PyObject_CallFunctionObjArgs',
     error_kind=ERR_MAGIC,
-    format_str='{dest} = py_call({comma_args})',
-    emit=simple_emit('{dest} = PyObject_CallFunctionObjArgs({comma_args}, NULL);'))
+    var_arg_type=object_rprimitive,
+    extra_int_constant=(0, c_int_rprimitive))
```
JukkaL, line 209: This is similar to below.

JukkaL:
Another idea: Diff the generated C when compiling something non-trivial, such as a benchmark or mypy. Try to minimize the differences.

TH3CHARLie:
> Another idea: Diff the generated C when compiling something non-trivial, such as a benchmark or mypy. Try to minimize the differences.

Thanks for the advice. Currently I only test against the py_call_op, and I compile several benchmarks from mypyc-benchmark, the only difference in the generated code is:
```
# old
cpy_r_r42 = PyObject_CallFunctionObjArgs(cpy_r_r41, cpy_r_r18, NULL);
# new
cpy_r_r42 = PyObject_CallFunctionObjArgs(cpy_r_r41, cpy_r_r18, 0);
```
It happens several times because call is a often used op, besides this, everything is the same.

TH3CHARLie:
I am more and more leaning the cause of this is the `0` and `NULL`, I did some research, looks like directly passing 0 to a function receiving a pointer and expected it to be `NULL` without any type cast is UB.

mypyc/irbuild/ll_builder.py
```diff
@@ -779,8 +779,12 @@ def call_c(self,
...
         if desc.extra_int_constant is not None:
             val, typ = desc.extra_int_constant
-            extra_int_constant = self.add(LoadInt(val, line, rtype=typ))
-            coerced.append(extra_int_constant)
+            if is_pointer_rprimitive(typ) and val == 0:
+                casted_constant = self.add(LoadGlobal(object_rprimitive, "NULL"))
+                coerced.append(casted_constant)
+            else:
+                extra_int_constant = self.add(LoadInt(val, line, rtype=typ))
+                coerced.append(extra_int_constant)
```
TH3CHARLie, line 787: Here's the special case. @JukkaL what do you think? If we decide this is a workable approach, I'll first fix the `to_str` of `LoadGlobal` since there's a hack there to ignore the prefix. I'll remove that hack and have all the related IR changes in a separate PR and merge this PR with master.

TH3CHARLie: I'm requesting a review for this even the build will fail.

JukkaL: This looks good.

TH3CHARLie: > This looks good.

Glad to here that. I've submitted #9331 before updating the IR tests in this PR.

JukkaL: Hmm I would actually prefer to do the translation to NULL in `mypy.codegen`, since it's only relevant for the C back end. We can represent null pointers as integer 0 values in the IR, but we should use NULL in the generated C.

TH3CHARLie:
Now I think it's ready for a full review @JukkaL

PR #9324, opened 2020-08-18
mention py.typed in missing imports doc

### LINKED ISSUE #9323: docs: mention py.typed files
* documentation bug
* ctrl+f for `py.typed`: https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
* it's not found
* I would expect it to be mentioned so that library authors know how to make mypy recognize the packages as having type hints
* latest documentation from the above link

Akuli:
fixes #9323

### DIFF
--- a/docs/source/running_mypy.rst
+++ b/docs/source/running_mypy.rst
@@ -182,7 +182,7 @@ corresponding type hints.
 
 Mypy will not try inferring the types of any 3rd party libraries you have installed
 unless they either have declared themselves to be
-:ref:`PEP 561 compliant stub package with a py.typed file <installed-packages>` or have registered
+:ref:`PEP 561 compliant stub package <installed-packages>` with a ``py.typed`` file or have registered
 themselves on `typeshed <https://github.com/python/typeshed>`_, the repository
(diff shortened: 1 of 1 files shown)


Akuli:
I wish there was a way to put `code` inside a link

docs/source/running_mypy.rst
```diff
@@ -182,7 +182,7 @@ corresponding type hints.
 
 Mypy will not try inferring the types of any 3rd party libraries you have installed
 unless they either have declared themselves to be
-:ref:`PEP 561 compliant stub package <installed-packages>` or have registered
+:ref:`PEP 561 compliant stub package <installed-packages>` with a ``py.typed`` file or have registered
```
gvanrossum, line 185: Actually not all installed packages need a `py.typed` file -- this marker is only needed when the `.pyi` files are in the same directory as the `.py` files. For packages containing only stubs, to be installed separately, no `py.typed` file is needed; instead such packages are recognized by the name `...-stubs`.

Read PEP 561 about it: https://www.python.org/dev/peps/pep-0561/#stub-only-packages

Akuli:
Is it better now?

PR #9331, opened 2020-08-20
[mypyc] Make LoadGlobal print full name during pretty IR printing

TH3CHARLie:
`LoadGlobal` used to have this hack to print partial name to keep consistent with the old `LoadStatic` behavior, now I've decided to make it print fullname since #9316 would need to print simple names like `NULL`.


JukkaL (approved):
This is a nice improvement and makes the IR output better match what's really happening. 

However, I actually don't think that we should use this for NULL, since NULL is a macro, not a global.

PR #9334, opened 2020-08-21
[mypyc] Support list of extra constants and str.split op

TH3CHARLie:
This PR changes the `CallC` description so that a list of extra constants can be used to a call. With this added functionality, `str.split` is implemented in new style IR.


JukkaL (approved):
Looks good!

PR #9343, opened 2020-08-23
[mypyc] Fast tuple equality checks

TH3CHARLie:
closes mypyc/mypyc#728


TH3CHARLie:
The tests failures are due to the runtime error described in the issue.

TH3CHARLie:
some performance numbers(running on my mac locally, very imprecise)

On master
```
interpreted: 0.180333s (avg of 5 iterations; stdev 0.61%)
compiled:    0.210294s (avg of 5 iterations; stdev 0.9%)

compiled is 0.858x faster
```

with this PR:
```
interpreted: 0.182221s (avg of 6 iterations; stdev 0.51%)
compiled:    0.055916s (avg of 6 iterations; stdev 0.35%)

compiled is 3.259x faster
```

mypyc/irbuild/ll_builder.py
```diff
@@ -656,10 +656,14 @@ def compare_tuples(self,
             lhs_item = lhs_items[i]
             rhs_item = rhs_items[i]
             compare = self.binary_op(lhs_item, rhs_item, op, line)
+            # Cast to bool if necessary since most types uses comparison returning a object type
+            # See generic_ops.py for more information
+            if not is_bool_rprimitive(compare.type):
+                comapre = self.coerce(compare, bool_rprimitive, line)
```
JelleZijlstra: typo ("comapre")

TH3CHARLie: Thanks! But I've already addressed that in e34c025. The problem now is the `Tuple[T, ...]` nested within a regular Tuple

TH3CHARLie:
@JukkaL I think it's ready for review now.

mypyc/irbuild/ll_builder.py
```diff
@@ -622,6 +626,56 @@ def compare_tagged(self, lhs: Value, rhs: Value, op: str, line: int) -> Value:
...
+                       rhs: Value,
+                       op: str,
+                       line: int = -1) -> Value:
+        """Compare two tuples item by item"""
+        # type cast to pass mypy check
+        assert isinstance(lhs.type, RTuple) and isinstance(rhs.type, RTuple)
+        equal = True if op == '==' else False
+        result = self.alloc_temp(bool_rprimitive)
+        # empty tuples
+        if (equal and len(lhs.type.types) == 0 and len(rhs.type.types) == 0):
```
msullivan, line 640: Why a special case for just equality?

TH3CHARLie: Nice catch!

msullivan:
No need for this to be in this PR, but it would be pretty straightforward to do `<` and friends too, right?

TH3CHARLie:
> No need for this to be in this PR, but it would be pretty straightforward to do `<` and friends too, right?

Sure, only need to do some minor modifications. But they appear much less than these two ops, so probably it's OK to lower their priority IMO.

mypyc/irbuild/ll_builder.py
```diff
@@ -622,6 +626,56 @@ def compare_tagged(self, lhs: Value, rhs: Value, op: str, line: int) -> Value:
...
+        for i in range(len(lhs.type.types)):
+            if i != 0:
+                self.activate_block(check_blocks[i])
+            lhs_item = lhs_items[i]
+            rhs_item = rhs_items[i]
+            compare = self.binary_op(lhs_item, rhs_item, op, line)
+            # Cast to bool if necessary since most types uses comparison returning a object type
+            # See generic_ops.py for more information
+            if not is_bool_rprimitive(compare.type):
+                compare = self.coerce(compare, bool_rprimitive, line)
```
JukkaL, line 663: Coerce only does a type check. I think that the normal CPython semantics would be to allow arbitrary objects that have a truth value, i.e. it may be better to use `bool_op` instead. Consider what happens if a user-defined class defines `__eq__` that returns a non-bool value such as `1` or `'x'`.

JukkaL (approved):
Looks good!

PR #9347, opened 2020-08-24
[mypyc] Merge dict get op and generic power op

TH3CHARLie:
This PR merges `get` op with `None` as default value for dict, and a power op for generic numbers, which completes the support of `generic_ops.py` and `dict_ops.py`

### DIFF
--- a/mypyc/lib-rt/dict_ops.c
+++ b/mypyc/lib-rt/dict_ops.c
@@ -63,6 +63,18 @@ PyObject *CPyDict_Get(PyObject *dict, PyObject *key, PyObject *fallback) {
     return res;
 }
 
+PyObject *CPyDict_GetWithNone(PyObject *dict, PyObject *key) {
+    PyObject *res = PyDict_GetItemWithError(dict, key);
+    if (!res) {
+        if (PyErr_Occurred()) {
+            return NULL;
+        }
+        res = Py_None;
+    }
+    Py_INCREF(res);
+    return res;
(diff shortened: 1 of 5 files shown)


mypyc/lib-rt/dict_ops.c
```diff
@@ -63,6 +63,18 @@ PyObject *CPyDict_Get(PyObject *dict, PyObject *key, PyObject *fallback) {
     return res;
 }
 
+PyObject *CPyDict_GetWithNone(PyObject *dict, PyObject *key) {
```
JukkaL, line 66: What about just calling `CPyDict_Get` in the body with an extra `Py_None` argument? The extra call should have minimal performance impact, and this would reduce some code repetition?

Declare the function in `CPy.h`.

mypyc/lib-rt/generic_ops.c
```diff
@@ -35,3 +35,8 @@ PyObject *CPyIter_Next(PyObject *iter)
 {
     return (*iter->ob_type->tp_iternext)(iter);
 }
+
+PyObject *CPyNumber_Power(PyObject *base, PyObject *index)
```
JukkaL, line 39: Declare this in `CPy.h` (it's unfortunate that this isn't caught by our tests).

JukkaL (approved):
Looks good, just a minor comments. Feel free to merge after you've addressed the feedback.

PR #9353, opened 2020-08-25
[mypyc] Boolean not op

TH3CHARLie:
This PR implements boolean not op with a XOR with 1.


mypyc/ir/ops.py
```diff
@@ -1281,6 +1281,10 @@ class BinaryIntOp(RegisterOp):
     LEFT_SHIFT = 203  # type: Final
     RIGHT_SHIFT = 204  # type: Final
 
+    # logical
+    NOR = 300  # type: Final
```
JukkaL, line 1285: NOR is somewhat unusual choice here, since it's not directly supported in hardware or C. It would be more standard to implement boolean not using xor, as `value XOR 1`

mypyc/irbuild/ll_builder.py
```diff
@@ -622,14 +622,17 @@ def compare_tagged(self, lhs: Value, rhs: Value, op: str, line: int) -> Value:
         self.goto_and_activate(out)
         return result
 
+    def unary_not(self,
+                  value: Value,
+                  line: int) -> Value:
+        return self.binary_int_op(bool_rprimitive, value, value, BinaryIntOp.NOR, line)
```
JukkaL, line 628: This is problematic, since `x NOR x` is `~x` (bitwise negation), whereas we want to only touch the lowest order bit.

mypyc/ir/ops.py
```diff
@@ -1305,6 +1309,8 @@ def sources(self) -> List[Value]:
         return [self.lhs, self.rhs]
 
     def to_str(self, env: Environment) -> str:
+        if self.lhs is self.rhs and self.op == BinaryIntOp.NOR:
+            return env.format('%r = !%r', self, self.lhs)
```
JukkaL, line 1313: This is confusing, since `!r` in C means a different thing than NOR. `~r` would match the semantics, but it would seem unexpected to me.

JukkaL (commented):
I like how we can drop support for the old-style unary ops.

NOR doesn't seem right here; I think that we'd need `value XOR 1` .

TH3CHARLie:
> I like how we can drop support for the old-style unary ops.
> 
> NOR doesn't seem right here; I think that we'd need `value XOR 1` .

Thanks for catching this. My initial thought was to use the XOR 1, I must have forgotten that when I was writing this.

JukkaL (approved):
Looks good now, thanks for the updates!

PR #9359, opened 2020-08-26
[mypyc] Merge pytype_from_template_op

TH3CHARLie:
This PR merges `pytype_from_template_op` op. Since other type-related ops are more-complicated, to make each PR more clear, I'll split them into separate PRs


TH3CHARLie:
on d0b381b
The following code
```python
from typing import Optional
class X: pass
class A(X): pass
class B(A): pass

def isa(x: object) -> bool:
    return isinstance(x, A)
```

compiles to
```c
char CPyDef_isa(PyObject *cpy_r_x) {
    PyObject *cpy_r_r0;
    CPyPtr cpy_r_r1;
    PyObject *cpy_r_r2;
    char cpy_r_r3;
    char cpy_r_r4;
    PyObject *cpy_r_r5;
    CPyPtr cpy_r_r6;
    PyObject *cpy_r_r7;
    char cpy_r_r8;
    char cpy_r_r9;
CPyL0: ;
    cpy_r_r0 = (PyObject *)CPyType_B;
    cpy_r_r1 = (CPyPtr)&((PyObject *)cpy_r_x)->ob_type;
    cpy_r_r2 = (PyObject *)cpy_r_r1;
    if (unlikely(cpy_r_r2 == NULL)) {
        CPy_AddTraceback("../foo2.py", "isa", 7, CPyStatic_globals);
        goto CPyL6;
    } else
        goto CPyL1;
CPyL1: ;
    cpy_r_r3 = cpy_r_r2 == cpy_r_r0;
    CPy_DecRef(cpy_r_r2);
    if (cpy_r_r3) {
        goto CPyL2;
    } else
        goto CPyL3;
CPyL2: ;
    cpy_r_r4 = cpy_r_r3;
    goto CPyL5;
CPyL3: ;
    cpy_r_r5 = (PyObject *)CPyType_A;
    cpy_r_r6 = (CPyPtr)&((PyObject *)cpy_r_x)->ob_type;
    cpy_r_r7 = (PyObject *)cpy_r_r6;
    if (unlikely(cpy_r_r7 == NULL)) {
        CPy_AddTraceback("../foo2.py", "isa", 7, CPyStatic_globals);
        goto CPyL6;
    } else
        goto CPyL4;
CPyL4: ;
    cpy_r_r8 = cpy_r_r7 == cpy_r_r5;
    CPy_DecRef(cpy_r_r7);
    cpy_r_r4 = cpy_r_r8;
    goto CPyL5;
CPyL5: ;
    return cpy_r_r4;
CPyL6: ;
    cpy_r_r9 = 2;
    return cpy_r_r9;
}

```

mypyc/test-data/irbuild-classes.test
```diff
@@ -460,18 +460,20 @@ def f(x: A) -> B:
...
-    r2, r3 :: __main__.B
+    r1 :: ptr
+    r2 :: bool
+    r3, r4 :: __main__.B
 L0:
     r0 = __main__.B :: type
-    r1 = type_is x, r0
-    if r1 goto L1 else goto L2 :: bool
+    r1 = get_element_ptr x ob_type :: PyObject
+    r2 = r1 == r0
```
JukkaL, line 469: You should read the value of the struct item before comparing it here.

TH3CHARLie: > You should read the value of the struct item before comparing it here.

I don't think so. The `Py_TYPE` macro is just getting the struct item(a pointer) and compare with another pointer.

JukkaL: Yeah, but this code gets a pointer to the struct item, not the actual struct item? We need to `LoadMem` to read the value.

Here's how it's defined:
```
#define Py_TYPE(ob)             (((PyObject*)(ob))->ob_type)
```

Note that it reads the value of `ob_type`, it doesn't just take a pointer to it.

mypyc/lib-rt/misc_ops.c
```diff
@@ -294,6 +294,12 @@ PyObject *CPyType_FromTemplate(PyTypeObject *template_,
     return NULL;
 }
 
+PyObject *CPyType_FromTemplateWarpper(PyObject *template_,
+                                      PyObject *orig_bases,
+                                      PyObject *modname) {
+    return CPyType_FromTemplate((PyTypeObject *)template_, orig_bases, modname);
+}
```
JukkaL, line 301: There's no need for a wrapper function, since the function we are wrapping is also defined here. You can just move the cast to the implementation of `CPyType_FromTemplate`, and this wrapper functions becomes redundant.

JukkaL (approved):
Thanks for the updates! Looks mostly good, but we can drop the wrapper function to simplify the code a bit. Feel free to merge after you've addressed by comment.

mypyc/lib-rt/misc_ops.c
TH3CHARLie: Unfortunately, that would break one other existing code from `generate_module_def` in `emitmodule.py` , it uses:
```python
emitter.emit_lines(
  '{t} = (PyTypeObject *)CPyType_FromTemplate({t}_template, NULL, modname);'.
  format(t=type_struct))
```
if we remove the wrapper function and change the signature, we'll need to change the emit here.

JukkaL: Hmm maybe we should change that emit then (add a cast?). This way we'd only need a single function.

TH3CHARLie: OK I'll patch that also in this PR.

Issue #9358, opened 2020-08-26
Don't simplify tuple types in error messages


JukkaL:
A user reported message like `Argument 1 to "method" of "Class" has incompatible type "Collection[Tuple[Gid, Gid, Dict[str, Any]]]"; expected "Collection[<tuple: 3 items>]"`. This isn't helpful. We should show the full tuple type always, at least in cases like this where there are two tuple types with the same number of items.

This would be a good issue for new contributors. The relevant code is in `mypy.messages.format_type_inner`.

JukkaL:
Fixed by #9360.

PR #9360, opened 2020-08-26
Don't simplify tuple/union types in error messages

XD-DENG:
This PR is addressing issue https://github.com/python/mypy/issues/9358.

- We have the same question for `union` type (other than `tuple` mentioned in https://github.com/python/mypy/issues/9358, so I made the same change for `union` type (please let me know if it's not desired)
- I hesitated if it's necessary to keep intermediate var `s`. But eventually I kept it to follow the existing practice in line https://github.com/python/mypy/blob/a08bbbb2fa37b8d2784dd3b194bcb1cd111bbcac/mypy/messages.py#L1643-L1644

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -1626,10 +1626,7 @@ def format_type_inner(typ: Type,
         for t in typ.items:
             items.append(format(t))
         s = 'Tuple[{}]'.format(', '.join(items))
-        if len(s) < 400:
-            return s
-        else:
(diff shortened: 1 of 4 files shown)


XD-DENG:
Hi @JukkaL , mind taking a look? Cheers!

JukkaL (approved):
Looks good, thanks for the PR! I agree that the simplification can be removed from both tuple and union types.

XD-DENG:
Thanks :)

PR #9363, opened 2020-08-27
[mypyc] Merge string equality ops

TH3CHARLie:
This PR merges str `==` and `!=` by directly building them in irbuild.

The old primitive relies on `ERR_MAGIC`  to handle the exception, now we use an `err_occurred_op` with `keep_propagating_op` to represent the same semantics.

Actually, our first several commits to replace `PrimitiveOp` with `CallC` cause an incorrect primitive lookup logic that at the point the generic compare is merged, string primitives would just use the generic ones. This PR will also fix this since we can obsolete the old binary op registry.


TH3CHARLie:
for this function:
```python
def f(x: str, y:str) -> bool:
    return x == y
```

master:
```c
char CPyDef_f(PyObject *cpy_r_x, PyObject *cpy_r_y) {
    char cpy_r_r0;
    char cpy_r_r1;
    int __tmp1;
CPyL0: ;
    __tmp1 = PyUnicode_Compare(cpy_r_x, cpy_r_y);
    if (__tmp1 == -1 && PyErr_Occurred())
        cpy_r_r0 = 2;
    else
        cpy_r_r0 = (__tmp1 == 0);
    if (unlikely(cpy_r_r0 == 2)) {
        CPy_AddTraceback("../foo1.py", "f", 2, CPyStatic_globals);
        goto CPyL2;
    } else
        goto CPyL1;
CPyL1: ;
    return cpy_r_r0;
CPyL2: ;
    cpy_r_r1 = 2;
    return cpy_r_r1;
}

```

this PR
```C
char CPyDef_f(PyObject *cpy_r_x, PyObject *cpy_r_y) {
    int32_t cpy_r_r0;
    char cpy_r_r1;
    char cpy_r_r2;
    char cpy_r_r3;
    char cpy_r_r4;
CPyL0: ;
    cpy_r_r0 = PyUnicode_Compare(cpy_r_x, cpy_r_y);
    cpy_r_r1 = cpy_r_r0 == -1;
    if (cpy_r_r1) {
        goto CPyL1;
    } else
        goto CPyL2;
CPyL1: ;
    cpy_r_r2 = CPy_ErrOccured();
    if (unlikely(!cpy_r_r2)) {
        CPy_AddTraceback("../foo1.py", "f", 2, CPyStatic_globals);
        goto CPyL3;
    } else
        goto CPyL2;
CPyL2: ;
    cpy_r_r3 = cpy_r_r0 == 0;
    return cpy_r_r3;
CPyL3: ;
    cpy_r_r4 = 2;
    return cpy_r_r4;
}
```

mypyc/irbuild/ll_builder.py
```diff
@@ -622,6 +624,24 @@ def compare_tagged(self, lhs: Value, rhs: Value, op: str, line: int) -> Value:
...
+        compare_result = self.call_c(unicode_comapre, [lhs, rhs], line)
+        error_constant = self.add(LoadInt(-1, line, c_int_rprimitive))
+        compare_error_check = self.add(ComparisonOp(compare_result,
+                                                    error_constant, ComparisonOp.EQ, line))
+        exception_check, final_compare = BasicBlock(), BasicBlock()
+        branch = Branch(compare_error_check, exception_check, final_compare, Branch.BOOL_EXPR)
+        branch.negated = False
+        self.add(branch)
+        self.activate_block(exception_check)
+        self.call_c(no_err_occurred_op, [], line)
```
JukkaL, line 638: Call `PyErr_Occurred` directly to avoid an extra function call in case the function is not inlined in a non-C back end?

JukkaL (commented):
Looks good -- nice to be able to get rid of more old-style registry code.

I only have one comment. The new implementation adds an extra function call that should be easy to avoid.

mypyc/irbuild/ll_builder.py
TH3CHARLie: I am not sure what does "call `PyErr_Occurred`" mean in your comment. The no_err_occurred_op here, if you look at its definition, it's just calling the function and checking the return value.

TH3CHARLie: The problem with directly calling `PyErr_Occurred` is that it returns `NULL` on normal exit and non-NULL on error, while `ERR_MAGIC` handles the exception when `NULL`, which means we'd add extra logic to negate the check on this single op.

JukkaL: Maybe you can can call `PyErr_Occurred`, compare the return against NULL (`ComparisonOp`), and if it's NULL, use `CPy_KeepPropagating`?

TH3CHARLie: That should work. Let me do a quick update.

TH3CHARLie:
Good news, I tried `PyErr_Occurred` with `is_borrowed=True`. No redundant incref/decref in the generated code.

mypyc/primitives/str_ops.py
```diff
@@ -80,30 +80,8 @@
...
-          error_kind=ERR_MAGIC,
-          emit=emit_str_compare('== 0'))
-
-# str1 != str2
-binary_op(op='!=',
-          arg_types=[str_rprimitive, str_rprimitive],
-          result_type=bool_rprimitive,
-          error_kind=ERR_MAGIC,
-          emit=emit_str_compare('!= 0'))
+unicode_comapre = c_custom_op(
```
JukkaL, line 83: Typo: unicode_cmapre -> unicode_compare.

JukkaL (approved):
Looks good, just one comment about a typo. Feel free to merge after you've addressed the typo.

PR #9364, opened 2020-08-27
[mypyc] Implement SetMem

TH3CHARLie:
This PR implements `SetMem` op to set value to a memory address.

### DIFF
--- a/mypyc/analysis/dataflow.py
+++ b/mypyc/analysis/dataflow.py
@@ -9,7 +9,7 @@ from mypyc.ir.ops import (
     BasicBlock, OpVisitor, Assign, LoadInt, LoadErrorValue, RegisterOp, Goto, Branch, Return, Call,
     Environment, Box, Unbox, Cast, Op, Unreachable, TupleGet, TupleSet, GetAttr, SetAttr,
     LoadStatic, InitStatic, PrimitiveOp, MethodCall, RaiseStandardError, CallC, LoadGlobal,
-    Truncate, BinaryIntOp, LoadMem, GetElementPtr, LoadAddress, ComparisonOp
+    Truncate, BinaryIntOp, LoadMem, GetElementPtr, LoadAddress, ComparisonOp, SetMem
 )
 
 
@@ -151,6 +151,10 @@ class BaseAnalysisVisitor(OpVisitor[GenAndKill]):
     def visit_assign(self, op: Assign) -> GenAndKill:
         raise NotImplementedError
 
+    @abstractmethod
+    def visit_set_mem(self, op: SetMem) -> GenAndKill:
+        raise NotImplementedError
+
     def visit_call(self, op: Call) -> GenAndKill:
         return self.visit_register_op(op)
 
@@ -248,6 +252,13 @@ class DefinedVisitor(BaseAnalysisVisitor):
         else:
             return {op.dest}, set()
 
+    def visit_set_mem(self, op: SetMem) -> GenAndKill:
+        # Loading an error value may undefine the register.
+        if isinstance(op.src, LoadErrorValue) and op.src.undefines:
+            return set(), {op.dest}
+        else:
+            return {op.dest}, set()
+
 
 def analyze_maybe_defined_regs(blocks: List[BasicBlock],
                                cfg: CFG,
@@ -308,6 +319,11 @@ class BorrowedArgumentsVisitor(BaseAnalysisVisitor):
             return set(), {op.dest}
         return set(), set()
 
+    def visit_set_mem(self, op: SetMem) -> GenAndKill:
+        if op.dest in self.args:
+            return set(), {op.dest}
+        return set(), set()
+
 
 def analyze_borrowed_arguments(
         blocks: List[BasicBlock],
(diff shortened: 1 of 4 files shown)


mypyc/ir/ops.py
```diff
@@ -1412,6 +1412,55 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
...
 
+class SetMem(Op):
+    """Write a memory location.
+
+    *(type *)src = val
+
+    Attributes:
+      type: Type of the read value
+      dest: Pointer to memory to write
+      src: source value
```
JukkaL, line 1423: Nit: Capitalize 'source' for consistency.

mypyc/ir/ops.py
```diff
@@ -1412,6 +1412,55 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
         return visitor.visit_load_mem(self)
 
 
+class SetMem(Op):
+    """Write a memory location.
+
+    *(type *)src = val
```
JukkaL, line 1418: This is a bit confusing, since `src` means a different thing compared tot he comment below. I'd suggest writing it like this instead:
```
*(type *)dest = src
```

mypyc/ir/ops.py
```diff
@@ -1412,6 +1412,55 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
...
+                 line: int = -1) -> None:
+        super().__init__(line)
+        self.type = type
+        self.src = src
+        self.dest = dest
+        self.base = base
+
+    def sources(self) -> List[Value]:
+        if self.base:
+            return [self.src, self.base]
```
JukkaL, line 1446: I think that `self.dest` should also be a source, since it's a value we access here.

mypyc/ir/ops.py
```diff
@@ -1412,6 +1412,55 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
...
+        else:
+            return [self.src]
```
JukkaL, line 1448: Similar to above.

mypyc/analysis/dataflow.py
```diff
@@ -381,6 +400,9 @@ def visit_register_op(self, op: RegisterOp) -> GenAndKill:
     def visit_assign(self, op: Assign) -> GenAndKill:
         return set(op.sources()), {op.dest}
 
+    def visit_set_mem(self, op: SetMem) -> GenAndKill:
+        return set(op.sources()), {op.dest}
```
JukkaL, line 404: `op.dest` is a pointer so we don't do anything to it, just the memory pointed to by it. Since we don't track properties of memory locations, the second item of the return value should be empty.

mypyc/analysis/dataflow.py
```diff
@@ -342,6 +358,9 @@ def visit_register_op(self, op: RegisterOp) -> GenAndKill:
     def visit_assign(self, op: Assign) -> GenAndKill:
         return set(), {op.dest}
 
+    def visit_set_mem(self, op: SetMem) -> GenAndKill:
+        return set(), {op.dest}
```
JukkaL, line 362: We don't modify the pointer, so the the second return value item should be empty. We don't track memory locations in data flow analyses.

mypyc/analysis/dataflow.py
```diff
@@ -308,6 +319,11 @@ def visit_assign(self, op: Assign) -> GenAndKill:
             return set(), {op.dest}
         return set(), set()
 
+    def visit_set_mem(self, op: SetMem) -> GenAndKill:
+        if op.dest in self.args:
+            return set(), {op.dest}
+        return set(), set()
```
JukkaL, line 325: This should be similar to `visit_register_op`, since we don't modify `op.dest`, but the value pointed to by it.

mypyc/analysis/dataflow.py
```diff
@@ -248,6 +252,13 @@ def visit_assign(self, op: Assign) -> GenAndKill:
         else:
             return {op.dest}, set()
 
+    def visit_set_mem(self, op: SetMem) -> GenAndKill:
+        # Loading an error value may undefine the register.
+        if isinstance(op.src, LoadErrorValue) and op.src.undefines:
+            return set(), {op.dest}
```
JukkaL, line 258: Since we don't modify `op.dest` but a value pointed to by it, this case can be removed.

mypyc/analysis/dataflow.py
```diff
@@ -248,6 +252,13 @@ def visit_assign(self, op: Assign) -> GenAndKill:
...
+        else:
+            return {op.dest}, set()
```
JukkaL, line 260: The return value should be `set(), set()`, since we don't modify `op.dest`.

JukkaL (commented):
The data flow analysis visit methods need work, since data flow analyses don't track memory locations and thus `SetMem` doesn't actually set the destination in terms of data flow. Otherwise looks good.

mypyc/analysis/dataflow.py
JukkaL: This should be `set(op.sources()), set()`, since is this essentially same as a void register op.

JukkaL (approved):
Thanks for the updates! I found one more small issue with dataflow analysis, otherwise looks good. Feel free to merge after you've addressed the feedback.

Issue #9342, opened 2020-08-23
Addition of pull_request template and issue template for mypy


iamrajiv:
Adding ISSUE_TEMPLATE.md and PULL_REQUEST_TEMPLATE.md file will enhance the workflow of the repository.

I will try to use the new form of ISSUE_TEMPLATE i.e. making a folder ISSUE_TEMPLATE in `.github` and adding different types of `.md ` files for different issues like `bug.md`, `documentation.md`, `proposal.md`, `feature.md` etc. A contributor can choose the type of templates whenever they make an issue or PR.

Reference: https://docs.github.com/en/github/building-a-strong-community/about-issue-and-pull-request-templates

gvanrossum:
Have you looked at the templates of other projects of similar size? There# should be no reason to reinvent the wheel here. :-)

gvanrossum:
This is *not* fixed, PR #9345 did not work.

iamrajiv:
@gvanrossum The PR #9345 does not work because I have deleted the `about` section from issue template without  `about` section it does not work. I have made new PR #9367 please take a look.

PR #9345, opened 2020-08-23
updated issue templates and added pull_request templates

iamrajiv:
Fixes: #9342

PR #9367, opened 2020-08-27
Updated issue templates and added pull request templates

iamrajiv:
Fixes: #9342
Related: #9345


gvanrossum:
Okay, this seems to work. Thanks for the quick fix!

iamrajiv:
Thank you for guidance @gvanrossum

PR #9368, opened 2020-08-28
[mypyc] Merge two ops and obsolete more old registry

TH3CHARLie:
This PR merges `list_get_item_unsafe_op` and `get_module_dict_op` and also removes several old registry and functions.


JukkaL (approved):
I love seeing so much legacy code deleted. Looks good!

PR #9371, opened 2020-08-28
Make None compatible with Hashable

### LINKED ISSUE #8768: None is not considered Hashable
```python
from typing import Hashable


def f() -> Hashable:
    return None


print(isinstance(None, Hashable))
```
The print returns True, so Python understands that None is hashable, but MyPy complains about incompatible return types:
```
error: Incompatible return value type (got "None", expected "Hashable")
```

JukkaL:
Fixes #8768.

### DIFF
--- a/mypy/subtypes.py
+++ b/mypy/subtypes.py
@@ -207,10 +207,15 @@ class SubtypeVisitor(TypeVisitor[bool]):
 
     def visit_none_type(self, left: NoneType) -> bool:
         if state.strict_optional:
-            return (isinstance(self.right, NoneType) or
-                    is_named_instance(self.right, 'builtins.object') or
-                    isinstance(self.right, Instance) and self.right.type.is_protocol and
-                    not self.right.type.protocol_members)
+            if isinstance(self.right, NoneType) or is_named_instance(self.right,
+                                                                     'builtins.object'):
+                return True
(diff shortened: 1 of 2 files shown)


ilevkivskyi (approved):

erictraut:
We received a similar [feature request for pyright](https://github.com/microsoft/pyright/issues/985). I considered doing a spot fix specifically for `__hash__`, but this struck me as a bit hacky and not very generalizable, since there are other valid protocols that could also apply to None.

I thought of another approach that might work better. Could we define a `NoneType` in `typing.pyi` that defines the methods supported by `None`? Type checkers could then use this class definition when doing protocol validation. Interested in your thoughts.

If you're open to that approach, I'd be happy to submit a PR to typeshed that defines `NoneType`.

erictraut:
@JukkaL do you have any thoughts on my proposal above?

JukkaL:
It sounds reasonable, since the type of `None` is not available in the builtins or `types` in Python 3. I don't see a use case for making it available at runtime, however. It was removed from `types` in Python 3 (`types.NoneType` exists in Python 2.7), and I'd assume this happened for a reason. So maybe we should be name it `_NoneType` to make it more clear that it's an internal definition that's only supposed to be used by tools?

erictraut:
Thanks, naming it `_NoneType` is a great suggestion. I'll submit a PR to typeshed.

PR #9378, opened 2020-08-30
[mypyc] Merge new_list_op and new_set_op

TH3CHARLie:
This PR builds `new_list_op` in irbuild.

For empty list, we introduce a `c_custom_op` to do the job. To set the following list item, we rely on pointer addition and LoadMem


TH3CHARLie:
Unfortunately, the 32bit test suite is failing because the offset computation in IR is `foo = bar + 8` while on 32bit platform it should be `foo = bar + 4`.

JukkaL:
> Unfortunately, the 32bit test suite is failing because the offset computation in IR is foo = bar + 8 while on 32bit platform it should be foo = bar + 4.

Similar to how we have `native_int`, we could add `WORD_SIZE` as a constant that gets replaced with the number of bytes in a word in a test case description (4 on a 32-bit platform and 8 on a 64-bit platform).

mypyc/ir/ops.py
```diff
@@ -1457,7 +1457,7 @@ def to_str(self, env: Environment) -> str:
             base = env.format(', %r', self.base)
         else:
             base = ''
-        return env.format("%r = set_mem %r%s :: %r*", self.dest, self.src, base, self.type)
+        return env.format("set_mem %r = %r%s :: %r*", self.dest, self.src, base, self.type)
```
JukkaL, line 1460: Nit: This syntax looks a bit weird to me. I'd prefer something like this:

```
set_mem r1, r2, r3 :: type
```

mypyc/primitives/list_ops.py
```diff
@@ -27,25 +27,11 @@
...
-new_list_op = custom_op(arg_types=[object_rprimitive],
-                        result_type=list_rprimitive,
-                        is_var_arg=True,
-                        error_kind=ERR_MAGIC,
-                        steals=True,
-                        format_str='{dest} = [{comma_args}]',
-                        emit=emit_new)
-
+new_empty_list_op = c_custom_op(
+    arg_types=[c_int_rprimitive],
```
JukkaL, line 31: This takes a `Py_ssize_t` argument.

mypyc/irbuild/ll_builder.py
```diff
@@ -755,6 +756,27 @@ def make_dict(self, key_value_pairs: Sequence[DictEntry], line: int) -> Value:
 
         return result
 
+    def new_list_op(self, values: List[Value], line: int) -> Value:
+        length = self.add(LoadInt(len(values), line, rtype=c_int_rprimitive))
```
JukkaL, line 760: This type should be `Py_ssize_t`.

mypyc/irbuild/ll_builder.py
```diff
@@ -755,6 +756,27 @@ def make_dict(self, key_value_pairs: Sequence[DictEntry], line: int) -> Value:
...
+        if len(values) == 0:
+            return empty_list
+        args = [self.coerce(item, object_rprimitive, line) for item in values]
+        ob_item_ptr = self.add(GetElementPtr(empty_list, PyListObject, 'ob_item', line))
+        ob_item_base = self.add(LoadMem(pointer_rprimitive, ob_item_ptr, empty_list, line))
+        for i in range(len(values)):
+            if i == 0:
+                item_address = ob_item_base
+            else:
+                offset = self.add(LoadInt(PLATFORM_SIZE * i, line, rtype=pointer_rprimitive))
```
JukkaL, line 771: The type should be `Py_ssize_t`, since this is a displacement, not a pointer.

mypyc/irbuild/ll_builder.py
```diff
@@ -755,6 +756,27 @@ def make_dict(self, key_value_pairs: Sequence[DictEntry], line: int) -> Value:
...
+                item_address = self.add(BinaryIntOp(pointer_rprimitive, ob_item_base, offset,
+                                                    BinaryIntOp.ADD, line))
```
JukkaL, line 773: We can add a pointer and an integer, but adding two pointers doesn't make sense.

mypyc/irbuild/ll_builder.py
```diff
@@ -755,6 +756,27 @@ def make_dict(self, key_value_pairs: Sequence[DictEntry], line: int) -> Value:
...
+            self.add(SetMem(object_rprimitive, item_address, args[i], empty_list, line))
+        return empty_list
+
+    def new_set_op(self, values: List[Value], line: int) -> Value:
+        return self.call_c(new_set_op, [], line)
```
JukkaL, line 778: This isn't passing the values on?

mypyc/irbuild/ll_builder.py
```diff
@@ -755,6 +756,27 @@ def make_dict(self, key_value_pairs: Sequence[DictEntry], line: int) -> Value:
 
         return result
 
+    def new_list_op(self, values: List[Value], line: int) -> Value:
+        length = self.add(LoadInt(len(values), line, rtype=c_int_rprimitive))
+        empty_list = self.call_c(new_empty_list_op, [length], line)
```
JukkaL, line 761: Nit: The name of this variable is confusing, since this generally is not an empty (length 0) list.

mypyc/primitives/list_ops.py
```diff
@@ -27,25 +27,11 @@
...
-# Construct a list from values: [item1, item2, ....]
-new_list_op = custom_op(arg_types=[object_rprimitive],
-                        result_type=list_rprimitive,
-                        is_var_arg=True,
-                        error_kind=ERR_MAGIC,
-                        steals=True,
-                        format_str='{dest} = [{comma_args}]',
-                        emit=emit_new)
-
+new_empty_list_op = c_custom_op(
```
JukkaL, line 30: Here 'empty' is a bit confusing. What about calling this `new_list_op`, since that's what the underlying function is called?

mypyc/transform/refcount.py
```diff
@@ -103,7 +103,7 @@ def transform_block(block: BasicBlock,
         key = (block, i)
 
         assert op not in pre_live[key]
-        dest = op.dest if isinstance(op, Assign) else op
+        dest = op.dest if isinstance(op, Assign) or isinstance(op, SetMem) else op
```
JukkaL, line 106: I'm not sure if this is the right thing to do. SetMem doesn't modify the destination in a data-flow sense. In data-flow the destination is actually a source. 

(This may be related: It looks like the `type` attribute of `SetMem` is incorrect -- it should be void, and the `type` of the value being assigned needs to be stored in another attribute.)

JukkaL (commented):
Overall approach looks good. It's great that we can represent such complicated ops in the IR. Left various comments about details. The biggest issues are the type attribute of `SetMem` (see the comment in `refcount.py`) and handling of set items.

mypyc/irbuild/ll_builder.py
TH3CHARLie: that's simply because `new_set_op` takes no argument, so `values` here is just an empty list.

JukkaL (approved):
Thanks for the updates! Looks good now.

TH3CHARLie:
Thanks for the review!

Issue #9362, opened 2020-08-26
Support narrowing Union of List types via index and isinstance


sbdchd:
It would be nice to be able to narrow a union of two different list types via indexing into the array and using an `isinstance` check

```python
from typing import Union, List

x: Union[List[str], List[int]] = [10]

if isinstance(x[0], str):
    reveal_type(x)
    # expected: builtins.list[builtins.str]
    # actual: Union[builtins.list[builtins.str], builtins.list[builtins.int]]
else:
    reveal_type(x)
    # expected: builtins.list[builtins.int]
    # actual: Union[builtins.list[builtins.str], builtins.list[builtins.int]]
```

One caveat is that the array could be empty, but let's assume we make an emptiness check before indexing

gvanrossum:
This is a special case in mypy. If you rewrite the initialization like this it works:
```
x: Union[List[str], List[int]]
x = [10]
```
Honestly I don't think this is documented very well -- you might be able to infer this from [Explicit types for variables](https://mypy.readthedocs.io/en/stable/type_inference_and_annotations.html#explicit-types-for-variables) but it could be more explicit.

sbdchd:
I think I may have made my example too simple.

In my current use case I parse some data from a network response which is a union of two different list types: `Union[List[List[str]], List[SearchResult]]`

I'm hoping to refine this response via indexing and an `isinstance` check

```python
from typing import Union, List
from dataclasses import dataclass


@dataclass
class SearchResult:
    id: str
    score: float
    match: str


def get_search_results() -> Union[List[List[str]], List[SearchResult]]:
    ...


x = get_search_results()

if isinstance(x[0], SearchResult):
    reveal_type(x)
    # expected: builtins.list[main.SearchResult]
    # actual: Union[builtins.list[builtins.list[builtins.str]], builtins.list[main.SearchResult]]
else:
    reveal_type(x)
    # expected: builtins.list[builtins.str]
    # actual: Union[builtins.list[builtins.list[builtins.str]], builtins.list[main.SearchResult]]

```


which I think can be simplified to the following as a test case

```python
from typing import Union, List


def foo() -> Union[List[str], List[int]]:
    ...


x = foo()

if isinstance(x[0], str):
    reveal_type(x)
    # expected: builtins.list[builtins.str]
    # actual: Union[builtins.list[builtins.str], builtins.list[builtins.int]]
else:
    reveal_type(x)
    # expected: builtins.list[builtins.int]
[...]

PR #9382, opened 2020-08-30
add narrowing for lists via index isinstance check

sbdchd:
Allows for narrowing a union of list types by indexing into the list
and checking the type of an element.

Since mypy already narrows the index expression's type, we use this 
expression's type to compare against the type arguments of the lists 
in the union and generate a new type that is narrower than the original.

```python
from typing import Union, List

a: Union[List[List[int]], List[str]]

# before
if isinstance(a[0], list):
    reveal_type(a) # N: Revealed type is 'Union[builtins.list[builtins.list[builtins.int]], builtins.list[builtins.str]]'
else:
    reveal_type(a) # N: Revealed type is 'Union[builtins.list[builtins.list[builtins.int]], builtins.list[builtins.str]]'

# after
if isinstance(a[0], list):
    reveal_type(a) # N: Revealed type is 'builtins.list[builtins.list[builtins.int]]'
else:
    reveal_type(a) # N: Revealed type is 'builtins.list[builtins.str]'
```

fixes: https://github.com/python/mypy/issues/9362

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3986,7 +3986,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                 if len(node.args) != 2:  # the error will be reported elsewhere
                     return {}, {}
                 if literal(expr) == LITERAL_TYPE:
-                    if_map, else_map = self.conditional_type_map_with_intersection(
+                    if_map_base, else_map_base = self.conditional_type_map_with_intersection(
                         expr,
                         type_map[expr],
                         get_isinstance_type(node.args[1], type_map),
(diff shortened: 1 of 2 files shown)


mypy/checker.py
```diff
@@ -3986,11 +3986,33 @@ def find_isinstance_check_helper(self, node: Expression) -> Tuple[TypeMap, TypeM
...
+                    if isinstance(first_node_arg, IndexExpr):
+                        arg_type = get_proper_type(type_map[first_node_arg.base])
+                        if isinstance(arg_type, UnionType) and if_map is not None and else_map is not None:
+                            var_name = first_node_arg.base
+                            if_branch_union = []
+                            else_branch_union = []
+                            t = if_map[expr]
+                            for x in arg_type.items:
+                                x = get_proper_type(x)
+                                if not isinstance(x, Instance) or not x.args:
```
sbdchd, line 4004: I wasn't sure how to restrict this to only `list` or `sequence` like types, so ended up doing this which is probably incorrect.

mypy/checker.py
```diff
@@ -3986,11 +3986,33 @@ def find_isinstance_check_helper(self, node: Expression) -> Tuple[TypeMap, TypeM
...
+                                    return if_map, else_map
+                                if is_overlapping_types(x.args[0], t) is not covers_at_runtime(x.args[0], t, False):
```
sbdchd, line 4006: wasn't sure about this, essentially needed this special case to get `object` to work as expected in the test cases

PR #9387, opened 2020-08-31
Fix search paths for `-p`

### LINKED ISSUE #9386: `-p` fails to find packages that are found by `-m`
## 🐛 Bug Report

Search path logic mismatch between `-p` and `-m` inputs severely limits the usefulness of the former.

## To Reproduce

Install a package into a virtualenv in develop/editable mode

Try to typecheck the package within that venv using `mypy -p import.path.to.package`

## Expected Behavior

Package typechecks

## Actual Behavior

mypy fails with `Can't find package`

## Your Environment

Tested in both mypy 0.780 and mypy 0.790+dev.5faa011a632b9fd78822ba6f528e56a7297ac695
Python 3.7
macOS 10.14

huguesb:
`-p` is similar to `-m` except that it will recursively typecheck submodules.

Unfortunately the early module walk was being done with search paths that
did not benefit from site_packages for the selected python executable,
which resulted in some packages being typechecked fine with `-m`, but
rejected as `not found` by `-p`.

Fixes #9386

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -14,7 +14,10 @@ from mypy import build
 from mypy import defaults
 from mypy import state
 from mypy import util
-from mypy.modulefinder import BuildSource, FindModuleCache, mypy_path, SearchPaths
+from mypy.modulefinder import (
+    BuildSource, FindModuleCache, SearchPaths,
+    get_site_packages_dirs, mypy_path,
+)
 from mypy.find_sources import create_source_list, InvalidSourceList
 from mypy.fscache import FileSystemCache
 from mypy.errors import CompileError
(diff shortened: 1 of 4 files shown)


gvanrossum (commented):
This looks like it should work, but I'd like a small refactoring to reduce the code duplication.

Also, please add a test. Instructions in `test-data/unit/README.md`. You could add a test to `cmdline.test`.

gvanrossum:
(In the future, please don't amend or squash commits. Code reviewers prefer to see changes resulting from code review separate from the original changes. We'll squash when we land.)

huguesb:
> (In the future, please don't amend or squash commits. Code reviewers prefer to see changes resulting from code review separate from the original changes. We'll squash when we land.)

Sorry, I'm used to code review tools that  gracefully handle force pushes and can show diffs between successive updates... I will avoid force pushes here in the future.

huguesb:
> This looks like it should work, but I'd like a small refactoring to reduce the code duplication.
> 
> Also, please add a test. Instructions in `test-data/unit/README.md`. You could add a test to `cmdline.test`.

The framework for cmdline.test explicitly disables site-packages. I'm modifying the pep561 test suite instead, since it already provisions a typed pkg in a venv.

gvanrossum:
Thanks for this contribution!

PR #9389, opened 2020-08-31
Add typing.OrderedDict to type_aliases

septatrix:
### Description

In 3.7.2 [typing.OrderedDict](https://docs.python.org/3/library/typing.html#typing.OrderedDict) was added to the typing module. Beforehand the correct way to annotate an `OrderedDict` was as a string or using the `__future__` import. However since it has been added to the stdlib it has not been added to the *type_aliases* for mypy [as it should have](https://github.com/python/mypy/pull/7707#issuecomment-541774957).

## Sample File

```python
from typing import OrderedDict

# substituting with collections.OrderedDict soothes mypy but fails at runtime
d: OrderedDict[str, int] = OrderedDict()
```
```console
$ mypy test.py
test.py:3: error: Variable "typing.OrderedDict" is not valid as a type
test.py:3: note: See https://mypy.readthedocs.io/en/latest/common_issues.html#variables-vs-type-aliases
test.py:3: error: "_Alias" not callable
```

## Test Plan

I do not know if the other aliases are tested anywhere. If so it might be a good idea to add a test there...

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -141,6 +141,8 @@ reverse_builtin_aliases = {
 nongen_builtins = {'builtins.tuple': 'typing.Tuple',
                    'builtins.enumerate': ''}  # type: Final
 nongen_builtins.update((name, alias) for alias, name in type_aliases.items())
+# Drop OrderedDict from this for backward compatibility
+del nongen_builtins['collections.OrderedDict']
 
 RUNTIME_PROTOCOL_DECOS = ('typing.runtime_checkable',
                           'typing_extensions.runtime',
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -118,8 +118,8 @@ type_aliases = {
 }  # type: Final
 
 # This keeps track of the oldest supported Python version where the corresponding
-# alias _target_ is available.
-type_aliases_target_versions = {
+# alias source is available.
+type_aliases_source_versions = {
     'typing.List': (2, 7),
     'typing.Dict': (2, 7),
     'typing.Set': (2, 7),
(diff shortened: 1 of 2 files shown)


septatrix:
Sadly I do not know why the windows CI fails however it seems to be the same on master?

The travis CI fails for *mypyc* with the same errors as I listed above for the mypy version before this patch. Does it contain a different database for valid type aliases?

JukkaL:
Maybe the problem is that `typing.OrderedDict` is only defined on Python 3.7 and later: https://github.com/python/typeshed/blob/master/stdlib/3/typing.pyi#L97

The CI may have been fixed. Can you rebase and see if it still happens?

septatrix:
> Maybe the problem is that `typing.OrderedDict` is only defined on Python 3.7 and later: https://github.com/python/typeshed/blob/master/stdlib/3/typing.pyi#L97

However the test for Python 3.6 passes. Also 3.7 with mypyc fails. This would only explain the failure for the 3.5 test with mypyc.

> The CI may have been fixed. Can you rebase and see if it still happens?

Rebasing fixed the windows CI errors.

septatrix:
@ilevkivskyi is there maybe another place where this mapping gets stored?

septatrix:
This is a rather annoying flaw in `mypy` for us as this forces us to use type comments in many places which we want to avoid as they do not work together well with formatters and other tools. The root problem has been found but sadly `mypyc` behaves different from `mypy` for some reason. I have been tinkering around a bit but have not been able to find the cause of this in the last few days. If anyone has another idea where this could come from I would be pleased to get some help as I do not know what else to try. Maybe I missed an important part for version detection or similar in the `ordered_dict.py` file or similar...

msullivan:
Yeah mypyc is obnoxiously fiddly about conditional compilation stuff. That part of the code seems to be a distraction, though; there's no real need to change anything about how mypy uses OrderedDict. The bigger problem here is that the change doesn't actually fix the problem, unfortunately. Something weirder is going on with these aliases.

msullivan:
Alright I applied some brute force to fix this issue, which is annoying. There should be a better way to fix it.

PR #9394, opened 2020-09-01
typeops: extend make_simplified_union fast path to enums

huguesb:
In PR #9192 a fast path was created to address the slowness reported
in issue #9169 wherein large Union or literal types would dramatically
slow down typechecking.

It is desirable to extend this fast path to cover Enum types, as these
can also leverage the O(n) set-based fast path instead of the O(n**2)
fallback.

This is seen to bring down the typechecking of a single fairly simple
chain of `if` statements operating on a large enum (~3k members) from
~40min to 12s in real-world code! Note that the timing is taken from
a pure-python run of mypy, as opposed to a compiled version.


huguesb:
Ping, can I get a review?

For typechecking our entire codebase, this single change brings the mypy runtime from 12min down to 5min!

gvanrossum (commented):
This is a fairly complex algorithm (as can be inferred from your comments and local commit history :-).

Can you add some tests that show the correct result is obtained in various edge cases?

msullivan (changes requested):
This looks pretty plausible to me.

:+1: on Guido's comment. A few targeted tests would help here, I think. (Or a convincing argument that the existing test suite covers this well already, I suppose?)

huguesb:
I believe the existing test suite covers this code very well. It definitely caught the mistake I originally made of skipping the slow path entirely for literals (which breaks down when strict optional is false as indicated in the comment). And since there are no intended behavior changes, it's unclear to me what new tests to add.

bwo:
ping on this—does hugues' comment above about the existing test suite not convince?

bwo:
I see it now has conflicts anyway :/

bwo:
ping

msullivan:
Yeah, OK. I'll merge if someone fixes the conflicts.

msullivan:
(I just did it)

PR #9396, opened 2020-09-02
Add support for next-gen attrs API

euresti:
### Description

Adds handlers for attr.define, attr.mutable, and attr.frozen.
Adds support for frozen having a non False default.
Adds support the attrib maker: field
Also includes support for auto_attribs=None which means "auto detect which method to use depending on my class".

## Test Plan

Added a couple of simple tests.

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -40,11 +40,11 @@ attr_dataclass_makers = {
 }  # type: Final
 attr_frozen_makers = {
     'attr.frozen'
-} # type: Final
+}  # type: Final
 attr_define_makers = {
     'attr.define',
     'attr.mutable'
-} # type: Final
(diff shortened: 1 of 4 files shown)


euresti:
Hi. The windows failures seem to be not-related to my diff.  Is there some other thing necessary to get this looked at?

euresti:
@ilevkivskyi am I missing anything to get this reviewed?

msullivan (approved):

msullivan:
I'm sorry about the delay here. I'm going to merge this, and I hope we can stay on top of this sort of PR in the future, but if you want to drop the attrs plugin in favor of one maintained inside attrs or alternately add mechanisms for disabling the built-in one in favor of an external one, we can figure out the best way to do that.

PR #9391, opened 2020-09-01
unbreak CI

JelleZijlstra:

PR #9399, opened 2020-09-02
Fix CI by avoiding setuptools 50.0

gvanrossum:
Alternative to https://github.com/python/mypy/pull/9391; suggestion from https://github.com/pypa/setuptools/issues/2352#issuecomment-683734196

Test plan: see if all tests pass.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -12,5 +12,5 @@ pytest-cov>=2.10.0,<3.0.0
 typing>=3.5.2; python_version < '3.5'
 py>=1.5.2
 virtualenv<20
-setuptools
+setuptools!=50.0
 importlib-metadata==0.20


gvanrossum:
Well dang! The Travis build ran and passed but the Windows builds (are those the GitHub actions?) failed, seemingly due to mypyc compilation issues. I don't know what to try next. (Could these be legitimate failures that were landed because CI was failing anyways???)

hauntsaninja:
Probably not landed legitimate failures. The last PR landed was https://github.com/python/mypy/pull/9376 which kind of hinged on CI working :-)

gvanrossum:
So maybe some *different* change broke the environment for Windows? Jelle's attempt to fix this (#9391) had the same Windows failures, as does harmless-looking #9387.

I note that the build logs, under "install tox", list "setuptools-50.0.3", which IIUC is the fixed version. In Jelle's PR it listed "setuptools-50.0.0". The failure in both looks the same (a whole block of tests failing, starting with `testCompileMypyc`. The most suspicious error seems to be `LINK : fatal error LNK1104: cannot open file 'python37.lib'`. Honestly I don't even know where the Windows build is configured these days. Anyway, this makes me think that we are dealing with a different failure than the setuptools thing.

gvanrossum:
Also I'm guessing that technically excluding setuptools 50.0 isn't enough -- we need to exclude 50.0.0 through 50.0.2, since the fix happened in 5.0.3: https://github.com/pypa/setuptools/commit/400b71c9b932893ab97dd6368ae7bfb822064a87

hauntsaninja:
Closing since we merged #9400

PR #9400, opened 2020-09-02
Fix setuptools CI issue

hauntsaninja:


gvanrossum:
If this works, maybe combine with my idea and try "setuptools!=50.0" in both places?

hauntsaninja:
Great, this seems to work.

Errors like `cl : Command line warning D9002 : ignoring unknown option '/O0'` smelt like we weren't using the right compiler. In mypyc/build.py, we gets the C compiler from `from distutils import sysconfig, ccompiler`, which is close enough that I continued to be suspicious.

gvanrossum (approved):
Okay to land.

But we should eventually support newer setuptools versions too, otherwise we'll just break in a different way in the future, and it'll be another wild goose change.

hauntsaninja:
Sure, I'll file an issue. So continues the packaging treadmill...

PR #9403, opened 2020-09-03
Add MYPY_CONFIG_FILE_DIR to environment when config file is read

### LINKED ISSUE #7968: Set MYPY_CONFIG_FILE_DIR environment variable

aghast:
### Description

Fixes #7968

The variable MYPY_CONFIG_FILE_DIR is set to the absolute path of (**ADDED: the dirname of**) the config file, prior to it being parsed.  Existing environment-variable substitution should handle the rest.

## Test Plan

Added new file `test-data/unit/envvars.test` to list of test files in `mypy/test/testcmdline.py`.

Added one testcase, using $MYPY_CONFIG_FILE_DIR  to set `files=` in mypy.ini.

### DIFF
--- a/test-data/unit/envvars.test
+++ b/test-data/unit/envvars.test
@@ -1,13 +1,10 @@
 # Test cases related to environment variables
 [case testEnvvar_MYPY_CONFIG_FILE_DIR]
 # cmd: mypy --config-file=subdir/mypy.ini
-# flags: --save-failures-to=/tmp/mypy
-# To run just this test:
-#    pytest -n0 -k testEnvvar_MYPY_CONFIG_FILE_DIR
 [file bogus.py]
-FOO: int = 'x'
+FOO = 'x'		# type: int
 [file subdir/good.py]
-BAR: int = 0
+BAR = 0			# type: int
 [file subdir/mypy.ini]
 \[mypy]
 files=$MYPY_CONFIG_FILE_DIR/good.py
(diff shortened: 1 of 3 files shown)


test-data/unit/envvars.test
```diff
@@ -0,0 +1,14 @@
+# Test cases related to environment variables
+[case testEnvvar_MYPY_CONFIG_FILE_DIR]
+# cmd: mypy --config-file=subdir/mypy.ini
+# flags: --save-failures-to=/tmp/mypy
+# To run just this test:
+#    pytest -n0 -k testEnvvar_MYPY_CONFIG_FILE_DIR
+[file bogus.py]
+FOO: int = 'x'
```
gvanrossum, line 8: Have you looked at the Travis-CI failure? This test fails on Python 3.5.1 because that doesn't support variable declarations. You can use type comments instead.

gvanrossum (changes requested):
Otherwise this looks good to me.

gvanrossum (approved):
Thanks!

gvanrossum:
Dang. The tests failed, even though they passed earlier. I don't see how my whitespace-only commit could have caused this, but I'm going to have to revert it, and then we can look into a fix.

Issue #9407, opened 2020-09-03
Move to travis-ci.com? (from.org)


gvanrossum:
See this post by Victor Stinner:

https://mail.python.org/archives/list/python-dev@python.org/message/ZABQWCQKUNKNO5XUCNIMB7GELSXTA646/

gvanrossum:
(Turns out we already migrated, but the buttons in the README were still pointing to .org.)

PR #9412, opened 2020-09-04
Change travis-ci reference links to dot com

joncwong:
### Description

#9407 - More accurate reference links in the README

## Test Plan

N/A


gvanrossum (approved):
Thanks! It seems the actual migration already happened.

PR #9415, opened 2020-09-04
Update int_pow test to match new overloads in typeshed

jakebailey:
### Description

At the moment, python/typeshed#4473 is approved but fails the mypy self tests. This updates the `cmdline.test` file to match the new error message with overloads. Though, I'm sorta wondering if it would be less fragile to just not do this specific check at all.

Feel free to take this (and the other PR) whenever it's least annoying to merge both together. Or, let me know if I'm missing something. 🙂 

cc @JelleZijlstra


hauntsaninja:
Thanks for both of the PRs!

PR #9339, opened 2020-08-22
pep612: add semanal for paramspec

hauntsaninja:
Linking #8645

Taking Jukka's suggestion from #9250 and splitting up the PR so we can merge and review a piece at a time. If you've read what's been pushed to #9250, this is the first bit of it, except I've changed mypy_extensions to typing_extensions since PEP 612 has been accepted.

Issue #8645, opened 2020-04-07
PEP 612 tracking issue


hauntsaninja:
Hello!

I got a little scared off from working on this by the many other enthusiastic people volunteering their time in #8373. However, there doesn't seem to have been much progress; since I have some extra time now, I figured sharing what I've got could kickstart things. Apologies if someone — or whatever a GSoC is — is already working on it :-) Branch is visible at https://github.com/hauntsaninja/mypy/tree/pep612, happy to make it a draft PR if that would help anything.

Overall seems like the plan would be: 
- Focus on generic functions first, implement generic class support later, as pyre plans to. 
- Make the fairly straight forward changes in semanal.py and nodes.py so you have the AST. Make relevant visitors visit it (look at where we visit_type_var_expr).
- Change typeanal.py, tvar_scope.py, etc to get binding going to the types you create in types.py. Seems like this would be the place to enforce valid use location too.
- Implement the semantics of ParameterSpecification in checkexpr.py, infer.py, solve.py, etc.
- Add a runtime implementation to mypy_extensions

Since ParameterSpecification is just a juiced up type variable, it makes sense to try and re-use existing logic for TypeVar as much as possible (especially if later PEPs make use of the bounds and variance arguments that PEP 612 reserves). In my branch, I've chosen to add a base class TypeVarLikeX in a couple different class hierarchies, e.g. TypeVarLikeExpr, TypeVarLikeDef. Since I'm not super familiar with mypy's code, this also helps me stumble upon things that I might need to change (it feels great trusting mypy to help improve mypy!). But this is a fairly large refactor (the one-line change in the details snippet alone will require changes in a dozen files), and could maybe affect like plugins or something, so I thought I'd post this before exploring further.
<details>

```
--- a/mypy/types.py
+++ b/mypy/types.py
@@ -983,7 +983,7 @@ class CallableType(FunctionLike):
                  fallback: Instance,
                  name: Optional[str] = None,
                  definition: Optional[SymbolNode] = None,
-                 variables: Optional[List[TypeVarDef]] = None,
+                 variables: Optional[Sequence[TypeVarLikeDef]] = None,
```
</details>

I'm also a little worried about issues like #1317 and #5738. I don't yet know the code well enough to anticipate what issues we might run into, but I'd welcome any thoughts on interactions between existing TypeVar code and any PEP 612 implementation.

JukkaL:
Your plan looks good to me! 

[...]

PR #9422, opened 2020-09-06
pep612: more semantic analysis for paramspec

hauntsaninja:
Linking #8645

Like #9339, much of this is in #9250. However, this PR is a little less self contained, in that it causes several future TODOs.

A lot of the changes here are necessitated by changing the type of `CallableType.variables` to `Sequence[TypeVarLikeDef]`. This is nice, because it informs me where I need to make changes in the future / integrates a little bit better with existing type var stuff. But let me know if you think I should instead add another field to CallableType.

The other decision here that I'm less sure of is not adding upper_bound and variance fields to ParamSpecDef. These fields exist on ParamSpecExpr, since PEP 612 reserves that. I figured we should add these to ParamSpecDef only once they have semantics so that it's easier to figure out where we need to define the new semantics, but that means TypeVarLikeDef abstracts over less and might lead to unnecessary work if we expect those semantics to be defined any time soon.

### DIFF
--- a/mypy/tvar_scope.py
+++ b/mypy/tvar_scope.py
@@ -1,16 +1,20 @@
 from typing import Optional, Dict, Union
-from mypy.types import TypeVarDef
-from mypy.nodes import TypeVarExpr, SymbolTableNode
+from mypy.types import TypeVarLikeDef, TypeVarDef, ParamSpecDef
+from mypy.nodes import ParamSpecExpr, TypeVarExpr, TypeVarLikeExpr, SymbolTableNode
 
 
-class TypeVarScope:
-    """Scope that holds bindings for type variables. Node fullname -> TypeVarDef."""
+class TypeVarLikeScope:
+    """Scope that holds bindings for type variables and parameter specifications.
+
+    Node fullname -> TypeVarLikeDef.
+
+    """
 
     def __init__(self,
-                 parent: 'Optional[TypeVarScope]' = None,
+                 parent: 'Optional[TypeVarLikeScope]' = None,
                  is_class_scope: bool = False,
-                 prohibited: 'Optional[TypeVarScope]' = None) -> None:
-        """Initializer for TypeVarScope
+                 prohibited: 'Optional[TypeVarLikeScope]' = None) -> None:
+        """Initializer for TypeVarLikeScope
 
         Parameters:
           parent: the outer scope for this scope
@@ -18,7 +22,7 @@ class TypeVarScope:
           prohibited: Type variables that aren't strictly in scope exactly,
                       but can't be bound because they're part of an outer class's scope.
         """
-        self.scope = {}  # type: Dict[str, TypeVarDef]
+        self.scope = {}  # type: Dict[str, TypeVarLikeDef]
         self.parent = parent
         self.func_id = 0
         self.class_id = 0
@@ -28,9 +32,9 @@ class TypeVarScope:
             self.func_id = parent.func_id
             self.class_id = parent.class_id
 
-    def get_function_scope(self) -> 'Optional[TypeVarScope]':
+    def get_function_scope(self) -> 'Optional[TypeVarLikeScope]':
         """Get the nearest parent that's a function scope, not a class scope"""
-        it = self  # type: Optional[TypeVarScope]
+        it = self  # type: Optional[TypeVarLikeScope]
         while it is not None and it.is_class_scope:
             it = it.parent
         return it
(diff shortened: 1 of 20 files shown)


mypy/tvar_scope.py
```diff
@@ -1,24 +1,28 @@
...
+from mypy.nodes import ParamSpecExpr, TypeVarExpr, TypeVarLikeExpr, SymbolTableNode
 
 
-class TypeVarScope:
-    """Scope that holds bindings for type variables. Node fullname -> TypeVarDef."""
+class TypeVarLikeScope:
+    """Scope that holds bindings for type variables and parameter specifications.
+
+    Node fullname -> TypeVarLikeDef.
+
```
JukkaL, line 10: Style nit: extra empty line.

mypy/typeanal.py
```diff
@@ -641,8 +649,23 @@ def analyze_callable_type(self, t: UnboundType) -> Type:
...
+                    sym = self.lookup_qualified(callable_args.name, callable_args)
+                    if sym is None:
+                        raise ValueError
+                    tvar_def = self.tvar_scope.get_binding(sym)
+                    if not isinstance(tvar_def, ParamSpecDef):
+                        raise ValueError
+
+                    # TODO(shantanu): construct correct type for paramspec
+                    return AnyType(TypeOfAny.from_error)
+                except ValueError:
```
JukkaL, line 664: I don't like using exceptions for control flow. Some of the other called functions might rise a ValueError, resulting in a confusing error message. I'd suggest using nested if statements or a helper function (return on error).

mypy/typeanal.py
```diff
@@ -860,17 +885,22 @@ def anal_type(self, t: Type, nested: bool = True) -> Type:
...
+                var_def.name,
+                var_def.fullname,
+                var_def.id.raw_id,
+                self.anal_array(var_def.values),
+                var_def.upper_bound.accept(self),
+                var_def.variance,
+                var_def.line
+            )
+        else:
+            return var_def
```
JukkaL, line 900: Should we construct a new instance and set the full name, at least?

mypy/types.py
```diff
@@ -2057,15 +2096,19 @@ def visit_callable_type(self, t: CallableType) -> str:
...
+                        vals = '({})'.format(', '.join(val.accept(self) for val in var.values))
+                        vs.append('{} in {}'.format(var.name, vals))
+                    elif not is_named_instance(var.upper_bound, 'builtins.object'):
+                        vs.append('{} <: {}'.format(var.name, var.upper_bound.accept(self)))
+                    else:
+                        vs.append(var.name)
                 else:
-                    vs.append(var.name)
+                    # For other TypeVarLikeDefs, just use the repr
+                    vs.append(repr(var))
```
JukkaL, line 2111: Add a TODO comment about actually implementing suitable repr? In the longer term, it would make sense to use a similar way of constructing the repr of both cases.

test-data/unit/check-parameter-specification.test
```diff
@@ -0,0 +1,19 @@
...
+[builtins fixtures/tuple.pyi]
+
+[case testParamSpecLocations]
+from typing import Callable, List
+from typing_extensions import ParamSpec, Concatenate
+P = ParamSpec('P')
+
+x: P  # E: ParamSpec "P" is unbound
+def foo1(x: Callable[P, int]) -> Callable[P, str]:  ...
+def foo2(x: P) -> P: ...  # E: Invalid location for ParamSpec "P"
```
JukkaL, line 13: What about adding a hint about where it should be used, e.g. something along these lines:
```
# N: You can use it like this: Callable[P, <return type>]
```

JukkaL (commented):
Thanks for the PR! The overall approach looks good to me. Left a bunch of minor comments. Feel free to merge one you've addressed them, or you can ask for another review iteration.

A more meta comment (don't need to address right now): If a user tries to use ParamSpec they'll likely get failed assertions? Before we make a public release with this new code, if the implementation is still incomplete, it would be better to hide it behind a command line flag so that users aren't exposed to assertion failures.

mypy/typeanal.py
hauntsaninja: In general, TypeAnalyser visit methods seem pretty happy to return their input, so not sure I see the point (unless we were to add upper_bound to ParamSpecDef or TypeVarLikeDef, in which case we'd need to visit it and reconstruct).

mypy/types.py
hauntsaninja: TypeVarLikeDef already implements a suitable repr.
The code here is just a reimplementation of TypeVarDef's repr that additionally visits values / upper_bound. We'd only need to do something here if we added upper_bound to ParamSpecDef or TypeVarLikeDef.

hauntsaninja:
Thanks! Fixed all the review points that have been marked as resolved, replied to the other two.

Sure, adding a command line flag seems reasonable / easy. I don't think any of my changes are in 0.790, but I'll make sure users can't hit the assertions before the next release. And for now, ParamSpec doesn't yet exist in typing_extensions (or typing).

mypy/typeanal.py
JukkaL: I was confused -- no change is needed here.

JukkaL (approved):
Thanks for the updates! Looks good.

gvanrossum:
> And for now, ParamSpec doesn't yet exist in typing_extensions (or typing).

FYI, it will be in the next alpha of 3.10 in typing.py, and likely in the next release of typing_extensions too. And it's already in typeshed (as is `Concatenate`).

PR #9426, opened 2020-09-06
stubtest: use allowlist instead of whitelist

hauntsaninja:
Allows use of the old options to preserve backwards compatibility
A step towards https://github.com/python/typeshed/issues/4436


ethanhs (approved):
Thanks!

PR #9438, opened 2020-09-11
Make the new bug templates less markup-heavy

gvanrossum:
- Remove emoji
- Instead of `## H2 headings` just use `**bold**`
- Add link to docs
- Add suggestion for new users not to file a bug

To see what it looks like: https://github.com/gvanrossum/mypy/issues/3 (you can file new bugs there as well)


JukkaL:
Okay, let's try these out!

JelleZijlstra (approved):

PR #9439, opened 2020-09-13
Add missing operators to BINARY_MAGIC_METHODS

### LINKED ISSUE #9296: Arguments in `__truediv__` and `__itruediv__` is not specified as positional-only
When I name a function `__div__` or `__idiv__`, it seems to have different type signature from normal functions.
Is it intended behavior?

## Code to reproduce

```
$ cat main.py
class A:
    def f(self, other: int) -> int:
        pass

    def __div__(self, other: int) -> int:
        pass

    def __truediv__(self, other: int) -> int:
        pass

    def __idiv__(self, other: int) -> int:
        pass

    def __itruediv__(self, other: int) -> int:
        pass


reveal_type(A.f)
reveal_type(A.__div__)
reveal_type(A.__truediv__)
reveal_type(A.__idiv__)
reveal_type(A.__itruediv__)

$ mypy main.py
main.py:18: note: Revealed type is 'def (self: main.A, other: builtins.int) -> builtins.int'
main.py:19: note: Revealed type is 'def (main.A, builtins.int) -> builtins.int'
main.py:20: note: Revealed type is 'def (self: main.A, other: builtins.int) -> builtins.int'
main.py:21: note: Revealed type is 'def (main.A, builtins.int) -> builtins.int'
main.py:22: note: Revealed type is 'def (self: main.A, other: builtins.int) -> builtins.int'

$ mypy --version
mypy 0.782
```

momohatt:
Closes #9296.
* Added `__[i,r]truediv__` to `BINARY_MAGIC_METHODS` so that mypy can recognize that the arguments of these methods are positional-only.
* I also added `__[i,r]matmul__` because they were missing too. Please let me know if that was unnecessary.


JukkaL (commented):
Thanks for the PR! The changes look good.

It would be nice to have a test case to check that at least some of these work as expected. Would you like to do it?

momohatt:
Sure! I didn't find any tests that checks this feature for other operators so I've added a new test for `__[i,r]add__`. What do you think?

TH3CHARLie (approved):
LGTM! Thanks for the great work!

PR #9443, opened 2020-09-15
Predict enum value type for unknown member names.

mgilson:
### Description

It is very common for enums to have homogenous member-value types.
In the case where we do not know what enum member we are dealing
with, we should sniff for that case and still collapse to a known
type if that assumption holds.

As a canonical example:

```py
from enum import Enum

class Color(Enum):
    RED = 1
    BLUE = 2

print(Color.RED.value)
reveal_type(Color.RED.value)

def print_color_value(color: Color) -> None:
    reveal_type(color.value)  # mypy reveals this as "Any", but it should be "builtins.int"

print_color_value(Color.BLUE)
```

## Test Plan

New unit tests have been added and all existing tests still pass.

### DIFF
--- a/mypy/plugins/enums.py
+++ b/mypy/plugins/enums.py
@@ -65,6 +65,8 @@ def _infer_value_type_with_auto_fallback(
         return None
     if (isinstance(proper_type, Instance) and
             proper_type.type.fullname == 'enum.auto'):
+        if not isinstance(ctx.type, Instance):
+            raise ValueError("An incorrect ctx.type was passed.")
         info = ctx.type.type
         # Find the first _generate_next_value_ on the mro.  We need to know
         # if it is `Enum` because `Enum` types say that the return-value of
@@ -83,7 +85,7 @@ def _infer_value_type_with_auto_fallback(
             return ctx.default_attr_type
 
         # This should be a `CallableType`
-        node_type = stnode.type
+        node_type = get_proper_type(stnode.type)
         if isinstance(node_type, CallableType):
             if type_with_generate_next_value.fullname == 'enum.Enum':
                 int_type = ctx.api.named_generic_type('builtins.int', [])
@@ -128,14 +130,13 @@ def enum_value_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
             stnodes = (info.get(name) for name in info.names)
             # Enums _can_ have methods.
             # Omit methods for our value inference.
-            stnodes_non_method = (
-                n for n in stnodes if not isinstance(n.type, CallableType))
             node_types = (
                 get_proper_type(n.type) if n else None
-                for n in stnodes_non_method)
+                for n in stnodes)
             proper_types = (
                 _infer_value_type_with_auto_fallback(ctx, t)
-                for t in node_types)
+                for t in node_types
+                if t is None or not isinstance(t, CallableType))
             underlying_type = next(proper_types, None)
             if underlying_type is None:
                 return ctx.default_attr_type
@@ -153,13 +154,12 @@ def enum_value_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
     if stnode is None:
         return ctx.default_attr_type
 
-    underlying_type = get_proper_type(stnode.type)
+    underlying_type = _infer_value_type_with_auto_fallback(
+        ctx, get_proper_type(stnode.type))
     if underlying_type is None:
-        # TODO: Deduce the inferred type if the user omits adding their own default types.
-        # TODO: Consider using the return type of `Enum._generate_next_value_` here?
         return ctx.default_attr_type
 
-    return _infer_value_type_with_auto_fallback(ctx, underlying_type)
+    return underlying_type
 
 
 def _extract_underlying_field_name(typ: Type) -> Optional[str]:
--- a/mypy/plugins/enums.py
+++ b/mypy/plugins/enums.py
@@ -73,21 +73,17 @@ def _infer_value_type_with_auto_fallback(
     # `_generate_next_value_` is `Any`.  In reality the default `auto()`
     # returns an `int` (presumably the `Any` in typeshed is to make it
     # easier to subclass and change the returned type).
-    type_with_generate_next_value = next(
-        (type_info for type_info in info.mro
-            if type_info.names.get('_generate_next_value_')),
-        None)
-    if type_with_generate_next_value is None:
+    type_with_gnv = next(
+        (ti for ti in info.mro if ti.names.get('_generate_next_value_')), None)
+    if type_with_gnv is None:
         return ctx.default_attr_type
 
-    stnode = type_with_generate_next_value.get('_generate_next_value_')
-    if stnode is None:
-        return ctx.default_attr_type
+    stnode = type_with_gnv.names['_generate_next_value_']
 
     # This should be a `CallableType`
     node_type = get_proper_type(stnode.type)
     if isinstance(node_type, CallableType):
-        if type_with_generate_next_value.fullname == 'enum.Enum':
+        if type_with_gnv.fullname == 'enum.Enum':
             int_type = ctx.api.named_generic_type('builtins.int', [])
             return int_type
         return get_proper_type(node_type.ret_type)
--- a/mypy/plugins/enums.py
+++ b/mypy/plugins/enums.py
@@ -78,7 +78,7 @@ def enum_value_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
     """
     enum_field_name = _extract_underlying_field_name(ctx.type)
     if enum_field_name is None:
-        # We do not know the ennum field name (perhaps it was passed to a function and we only
+        # We do not know the enum field name (perhaps it was passed to a function and we only
         # know that it _is_ a member).  All is not lost however, if we can prove that the all
         # of the enum members have the same value-type, then it doesn't matter which member
         # was passed in.  The value-type is still known.
--- a/mypy/plugins/enums.py
+++ b/mypy/plugins/enums.py
@@ -63,36 +63,36 @@ def _infer_value_type_with_auto_fallback(
     """
     if proper_type is None:
         return None
-    if (isinstance(proper_type, Instance) and
-            proper_type.type.fullname == 'enum.auto'):
-        if not isinstance(ctx.type, Instance):
-            raise ValueError("An incorrect ctx.type was passed.")
-        info = ctx.type.type
-        # Find the first _generate_next_value_ on the mro.  We need to know
-        # if it is `Enum` because `Enum` types say that the return-value of
-        # `_generate_next_value_` is `Any`.  In reality the default `auto()`
-        # returns an `int` (presumably the `Any` in typeshed is to make it
-        # easier to subclass and change the returned type).
-        type_with_generate_next_value = next(
-            (type_info for type_info in info.mro
-                if type_info.names.get('_generate_next_value_')),
-            None)
-        if type_with_generate_next_value is None:
-            return ctx.default_attr_type
-
-        stnode = type_with_generate_next_value.get('_generate_next_value_')
-        if stnode is None:
-            return ctx.default_attr_type
-
-        # This should be a `CallableType`
-        node_type = get_proper_type(stnode.type)
-        if isinstance(node_type, CallableType):
-            if type_with_generate_next_value.fullname == 'enum.Enum':
-                int_type = ctx.api.named_generic_type('builtins.int', [])
-                return int_type
-            return get_proper_type(node_type.ret_type)
+    if not (isinstance(proper_type, Instance) or
+            proper_type.type.fullname != 'enum.auto'):
+        return proper_type
+    if not isinstance(ctx.type, Instance):
+        raise ValueError("An incorrect ctx.type was passed.")
+    info = ctx.type.type
+    # Find the first _generate_next_value_ on the mro.  We need to know
+    # if it is `Enum` because `Enum` types say that the return-value of
+    # `_generate_next_value_` is `Any`.  In reality the default `auto()`
+    # returns an `int` (presumably the `Any` in typeshed is to make it
+    # easier to subclass and change the returned type).
+    type_with_generate_next_value = next(
+        (type_info for type_info in info.mro
+            if type_info.names.get('_generate_next_value_')),
+        None)
+    if type_with_generate_next_value is None:
         return ctx.default_attr_type
-    return proper_type
+
+    stnode = type_with_generate_next_value.get('_generate_next_value_')
+    if stnode is None:
+        return ctx.default_attr_type
+
+    # This should be a `CallableType`
+    node_type = get_proper_type(stnode.type)
+    if isinstance(node_type, CallableType):
+        if type_with_generate_next_value.fullname == 'enum.Enum':
+            int_type = ctx.api.named_generic_type('builtins.int', [])
+            return int_type
+        return get_proper_type(node_type.ret_type)
+    return ctx.default_attr_type
 
 
 def enum_value_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
--- a/mypy/plugins/enums.py
+++ b/mypy/plugins/enums.py
@@ -10,7 +10,7 @@ Note that this file does *not* contain all special-cased logic related to enums:
 we actually bake some of it directly in to the semantic analysis layer (see
 semanal_enum.py).
 """
-from typing import Optional
+from typing import Iterable, Optional, TypeVar
 from typing_extensions import Final
 
 import mypy.plugin  # To avoid circular imports.
@@ -53,6 +53,19 @@ def enum_name_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
         return str_type.copy_modified(last_known_value=literal_type)
 
 
+_T = TypeVar('_T')
+
+
+def _first(it: Iterable[_T]) -> Optional[_T]:
+    """Return the first value from any iterable.
+
+    Returns ``None`` if the iterable is empty.
+    """
+    for val in it:
+        return val
+    return None
+
+
 def _infer_value_type_with_auto_fallback(
         ctx: 'mypy.plugin.AttributeContext',
         proper_type: Optional[ProperType]) -> Optional[Type]:
@@ -73,8 +86,8 @@ def _infer_value_type_with_auto_fallback(
     # `_generate_next_value_` is `Any`.  In reality the default `auto()`
     # returns an `int` (presumably the `Any` in typeshed is to make it
     # easier to subclass and change the returned type).
-    type_with_gnv = next(
-        (ti for ti in info.mro if ti.names.get('_generate_next_value_')), None)
+    type_with_gnv = _first(
+        ti for ti in info.mro if ti.names.get('_generate_next_value_'))
     if type_with_gnv is None:
         return ctx.default_attr_type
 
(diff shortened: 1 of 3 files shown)


mgilson:
Here I thought I was opening this PR against my fork -- I guess I need to work against public github more often.  Let me know if you want me to create an issue to track this.

mypy/plugins/enums.py
```diff
@@ -78,6 +78,22 @@ class SomeEnum:
     """
     enum_field_name = _extract_underlying_field_name(ctx.type)
     if enum_field_name is None:
+        # We do not know the ennum field name (perhaps it was passed to a function and we only
```
gvanrossum:
```suggestion
        # We do not know the enum field name (perhaps it was passed to a function and we only
```

mypy/plugins/enums.py
```diff
@@ -78,6 +78,22 @@ class SomeEnum:
     """
     enum_field_name = _extract_underlying_field_name(ctx.type)
     if enum_field_name is None:
+        # We do not know the enum field name (perhaps it was passed to a function and we only
+        # know that it _is_ a member).  All is not lost however, if we can prove that the all
+        # of the enum members have the same value-type, then it doesn't matter which member
+        # was passed in.  The value-type is still known.
+        if isinstance(ctx.type, Instance):
+            info = ctx.type.type
```
mgilson, line 86: Should we memoize this bit?  It'd probably be easy enough to pull this into a separate function an `@functools.lru_cache` it...

test-data/unit/check-enum.test
```diff
@@ -59,6 +59,26 @@ reveal_type(Truth.true.name)    # N: Revealed type is 'Literal['true']?'
...
+[builtins fixtures/bool.pyi]
+
+[case testEnumValueInhomogenous]
+from enum import Enum
+class Truth(Enum):
+    true = 'True'
+    false = 0
+
+def cannot_infer_truth(truth: Truth) -> None:
+    reveal_type(truth.value) # N: Revealed type is 'Any'
```
gvanrossum, line 79: I wonder if this shouldn't infer `object` instead of `Any`?

In favor of `Any`: That's what we did before in this case; it prevents false positives.

In favor of `object`: That's what you'd see for e.g. `reveal_type("" if a else 0)`, assuming `a` has an unknown value.

mypy/plugins/enums.py
gvanrossum: I wouldn't bother. It would be quite the program where you could show a measurable performance improvement for type checking by memoizing this.

mypy/plugins/enums.py
```diff
@@ -78,6 +78,22 @@ class SomeEnum:
...
+            stnodes = (info.get(name) for name in info.names)
+            first_node = next(stnodes, None)
+            if first_node is None:
+                return ctx.default_attr_type
+            first_node_type = first_node.type
+            if all(node is not None and node.type == first_node_type for node in stnodes):
+                underlying_type = get_proper_type(first_node_type)
+                if underlying_type is not None:
+                    return underlying_type
```
gvanrossum, line 95: If I read the docstring of `get_proper_type()` correctly, it suggest that here you should actually be returning `first_node_type`. And in fact, perhaps you should probably adjust the `all()` check to go through `get_proper_type()` instead?

mypy/plugins/enums.py
```diff
@@ -78,6 +78,22 @@ class SomeEnum:
...
+
         return ctx.default_attr_type
 
     assert isinstance(ctx.type, Instance)
```
gvanrossum, line 99: Below on lines 91-92 is a TODO that appears relevant. Unfortunately it appears hard to do something useful here -- while the default implementations of `auto()` always return ints, the docs specifically suggest that you can override it and return anything you like (the docs show it returning the field name).

So I'm not sure if you're going to be able to do much about that, but I thought I'd point it out as a potential improvement for your PR. It would certainly be nice if we got `int` for the following example:
```
class Truth(Enum):
    false = auto()
    true = auto()
```
As well as for a similar class where the first value is `0` instead of `auto()`.

But you'd have to check whether they're overriding `_generate_next_value_()` in the class.

gvanrossum (commented):
This is good work! Thanks for working on this. I have a few comments.

mypy/plugins/enums.py
mgilson: It took a little digging into the mypy guts here, but in the end, getting that working wasn't _too_ bad.  Running locally, the following code:

```py
from enum import Enum, auto
from typing import NamedTuple, List, Any


class AColorVal(NamedTuple):
    c: str


class Color(Enum):
    RED = 1
    BLUE = 2


def print_color_value(color: Color) -> None:
    reveal_type(color.value)


class Color2(Enum):
    RED = (1,)
    BLUE = (2,)


def print_color2_value(color: Color2) -> None:
    reveal_type(color.value)


class Color3(Enum):
    RED = AColorVal("RED")
    BLUE = AColorVal("BLUE")


def print_color3_value(color: Color3) -> None:
    reveal_type(color.value)


class Color4(Enum):
    RED = 10
    BLUE = auto()


def print_color4_value(color: Color4) -> None:
    reveal_type(color.value)


print(Color4.BLUE.value)


class AutoName(Enum):

    @staticmethod
    def _generate_next_value_(name: str, start: int, count: int, last_values: List[Any]) -> str:
        return name

class Color5(AutoName):
    NORTH = auto()
    SOUTH = auto()
    EAST = auto()
    WEST = auto()


def print_color5_value(color: Color5) -> None:
    reveal_type(color.value)


class Color6(Enum):
    RED = auto()
    BLUE = auto()

    def foo(self) -> str:
        return "bar"


def print_color6_value(color: Color6) -> None:
    reveal_type(color.value)
```

Produces the expected output:

```sh
/Users/mgilson/sandbox/foo.py:15: note: Revealed type is 'builtins.int'
/Users/mgilson/sandbox/foo.py:24: note: Revealed type is 'Tuple[builtins.int]'
/Users/mgilson/sandbox/foo.py:33: note: Revealed type is 'Tuple[builtins.str, fallback=foo.AColorVal]'
/Users/mgilson/sandbox/foo.py:42: note: Revealed type is 'builtins.int'
/Users/mgilson/sandbox/foo.py:62: note: Revealed type is 'builtins.str'
/Users/mgilson/sandbox/foo.py:74: note: Revealed type is 'builtins.int'
```

Unfortunately, the tests aren't quite working out (I keep getting complaints about `auto` not having the correct number of arguments.  I'm guessing that there is something interesting going on in the way the fixtures are set up that I haven't completely wrapped my head around (presumably sucking in a bunch of stuff from `typeshed` would slow them down a whole lot).

mypy/plugins/enums.py
mgilson: I took a stab at this, but I'm not completely sure what you're saying here :).  It's unclear to me whether the result of `enum_value_callback` should be a `ProperType` or not.  The existing code seems to try to return a `ProperType` so I just followed suit there.  I did change the "all-equal" to compare the entities after figuring out their `ProperType` though.

test-data/unit/check-enum.test
mgilson: I was trying to just follow the existing art.  I imagine that changing to `object` could cause some code to fail with false positives (If a user is actually _using_ the `.value`, it seems unlikely that they will be restricting themselves to the interface provided by `object` so this would probably force them to do additional `cast` or other methods of type-narrowing)

With that said ... once this is implemented, it shouldn't matter to me either way.  Hopefully `mypy` will know the types of all of my enum values since I see no reason for inhomogenous values and it should never bother me in either case :).  I'll do whatever you like.

test-data/unit/lib-stub/enum.pyi
```diff
@@ -21,6 +21,10 @@ class Enum(metaclass=EnumMeta):
     _name_: str
     _value_: Any
 
+    # In reality, _generate_next_value_ is python3.6 only and has a different signature.
```
mgilson, line 24: Shenanigans are happening here -- I'm not sure the best way to do this.  If I use the signature (as written in `typeshed`), we end up with:

```
        @staticmethod
        def _generate_next_value_(name: str, start: int, count: int, last_values: List[Any]) -> Any: pass
```

That pulls in `staticmethod` which doesn't have a stub in `lib-stub`.  From that point, there are a bunch of different ways we can lie (e.g. just add a `self` and drop the `staticmethod`), but it seems to me like a little lie is just as harmful as a big one (maybe more because it's less likely to be noticed).

test-data/unit/lib-stub/enum.pyi
```diff
@@ -37,4 +41,7 @@ class IntFlag(int, Flag):
 
 
 class auto(IntFlag):
-    value: Any
\ No newline at end of file
+
+    value: Any
+
+    def __init__(self) -> None: pass
```
mgilson, line 47: I don't know why I need to add this.  It _isn't_ present in `typeshed`, but my actual runs still seem to work.  Without it, I have two tests that fail with complaints that `auto` didn't get the right number of arguments:

```
[case testEnumValueSomeAuto]
from enum import Enum, auto
class Truth(Enum):
    true = 8675309
    false = auto()

def infer_truth(truth: Truth) -> None:
    reveal_type(truth.value) # N: Revealed type is 'builtins.int'
```

And

```
[case testEnumValueCustomAuto]
from enum import Enum, auto
from typing import List, Any
class AutoName(Enum):

    # In `typeshed`, this is a staticmethod and has more arguments,
    # but I have lied a bit to keep the test stubs lean.
    def _generate_next_value_(self) -> str:
        return "name"

class Truth(AutoName):
    true = auto()
    false = auto()

def infer_truth(truth: Truth) -> None:
    reveal_type(truth.value) # N: Revealed type is 'builtins.str'
```

If anyone can help me reason through why _that_ would be, I'm happy to dig deeper.

mypy/plugins/enums.py
gvanrossum: Hm, I think I misunderstood this. I thought that using `get_proper_type()` would distinguish `A` from `int` if we have
```
A = int
def f(a: A): ...
```
but it doesn't seem to work that way. So you're forgiven for not understanding me. :-)

I think what you have now is fine.

test-data/unit/check-enum.test
gvanrossum: Okay, let's stick with tradition and keep `Any`.

test-data/unit/lib-stub/enum.pyi
gvanrossum: I think this comment is all we need if in the future mypy and/or the test fixtures somehow get clever enough to notice the signature mismatch.

gvanrossum (commented):
This has mushroomed a bit, but I like how it's working now. I have a few style nits (you can ignore those if you want to) and then when you're ready I'll approve and merge. Thanks again for this PR!

test-data/unit/lib-stub/enum.pyi
mgilson: I had the same thought, but as far as I can tell, the same thing happens in the official stubs in [`typeshed`](https://github.com/python/typeshed/blob/master/stdlib/3/enum.pyi) -- but we don't see this complaint when I run it for real.  It also only complains for these two tests (which admittedly _are_ doing something a little special with `auto`) -- even though there are a lot of other tests that use `auto`.  I've never really seen `mypy` come up with a different required signature for something based on surrounding context.  Maybe there is something interesting happening in the plugin that I haven't quite grokked yet -- particularly when mixed with the slightly less-than-usual stub setup for these tests?

mypy/plugins/enums.py
```diff
@@ -53,6 +53,48 @@ def enum_name_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
...
+
+    This method is a no-op for a `None` proper_type and also in the case where
+    the type is not "enum.auto"
+    """
+    if proper_type is None:
+        return None
+    if (isinstance(proper_type, Instance) and
+            proper_type.type.fullname == 'enum.auto'):
+        if not isinstance(ctx.type, Instance):
+            raise ValueError("An incorrect ctx.type was passed.")
```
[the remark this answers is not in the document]
mgilson: Well ... It probably could be.  Looking at the call tree, I don't see how it _could_ be anything other than an `Instance` (the calling function _does_ use an `assert` for the same check).  Note that this check is on `ctx.type` -- not `proper_type` which was checked when entering this block.  So this function alone can't guarantee the type of `ctx.type` AFAICT.  I can go either way for an internal function like this though.  I'll change it in a separate commit and then we can revert it if we decide to waffle.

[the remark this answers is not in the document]
mgilson: Done in aae9de3

mypy/plugins/enums.py
```diff
@@ -53,6 +53,48 @@ def enum_name_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
...
+        type_with_generate_next_value = next(
+            (type_info for type_info in info.mro
+                if type_info.names.get('_generate_next_value_')),
+            None)
+        if type_with_generate_next_value is None:
+            return ctx.default_attr_type
+
+        stnode = type_with_generate_next_value.get('_generate_next_value_')
+        if stnode is None:
+            return ctx.default_attr_type
```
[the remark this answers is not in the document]
mgilson: I agree and it saves a branch.  Thanks for the suggestion.

mypy/plugins/enums.py
```diff
@@ -53,6 +53,48 @@ def enum_name_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
...
+        proper_type: Optional[ProperType]) -> Optional[Type]:
+    """Figure out the type of an enum value accounting for `auto()`.
+
+    This method is a no-op for a `None` proper_type and also in the case where
+    the type is not "enum.auto"
+    """
+    if proper_type is None:
+        return None
+    if (isinstance(proper_type, Instance) and
+            proper_type.type.fullname == 'enum.auto'):
```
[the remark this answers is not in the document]
mgilson: Yeah, I considered doing that initially ... then I didn't like the way the boolean expression worked out so I didn't ... then I didn't like the same things you were pointing out but I felt committed.  Anyway, I flip-flopped it as suggested.  With the shortened names and decreased indentation level, I was also able to reduce a little of the line-wrapping which was nice as well.

mypy/plugins/enums.py
```diff
@@ -53,6 +53,48 @@ def enum_name_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
...
+        if not isinstance(ctx.type, Instance):
+            raise ValueError("An incorrect ctx.type was passed.")
+        info = ctx.type.type
+        # Find the first _generate_next_value_ on the mro.  We need to know
+        # if it is `Enum` because `Enum` types say that the return-value of
+        # `_generate_next_value_` is `Any`.  In reality the default `auto()`
+        # returns an `int` (presumably the `Any` in typeshed is to make it
+        # easier to subclass and change the returned type).
+        type_with_generate_next_value = next(
```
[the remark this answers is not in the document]
mgilson: Done.

mypy/plugins/enums.py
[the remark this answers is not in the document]
gvanrossum: +1. Mypy doesn't catch ValueError, so the effect would be the same -- if it ever gets raised the user sees a crash.

mypy/plugins/enums.py
```diff
@@ -63,36 +63,31 @@ def _infer_value_type_with_auto_fallback(
...
-            return ctx.default_attr_type
-
-        # This should be a `CallableType`
-        node_type = get_proper_type(stnode.type)
-        if isinstance(node_type, CallableType):
-            if type_with_generate_next_value.fullname == 'enum.Enum':
-                int_type = ctx.api.named_generic_type('builtins.int', [])
-                return int_type
-            return get_proper_type(node_type.ret_type)
+    if not (isinstance(proper_type, Instance) or
```
gvanrossum, line 66: `and`!

test-data/unit/lib-stub/enum.pyi
gvanrossum: Adding `[builtins fixtures/primitives.pyi]` to those two tests resolves this. I think I have an idea: without that, you get `int` and `object` from the *very* basic `lib-stub/builtins.pyi`, and there `int` has no `__init__`. But in `fixtures/primitives.pyi` there is an `int.__init__`.

The secret is in the MRO of `auto`, which is (in either case) `auto`, `IntFlag`, `int`, `Flag`, `Enum`, `object`. With `primitives.pyi`, we pick up `__init__` from `int`. But without it, we pick it up from `Enum`... And `Enum.__init__` requires an argument.

So it's purely a test issue, and IMO the fix is to add `[builtins fixtures/primitives.pyi]` and get rid of `auto.__init__`. (And of the blank line before `value: Any`.)

gvanrossum:
I don't understand the Travis failures. These seem to be due to some bad code generated by mypyc. We may have to consult a mypy expert.

gvanrossum:
@TH3CHARLie Could you have a look at the mypyc test failures on Travis here? I see this error:
```
build/__native_dbbd1b679b3c31e8ff5a.c:972286:15: error: incompatible integer to
      pointer conversion assigning to 'PyObject *' (aka 'struct _object *') from
      'int' [-Werror,-Wint-conversion]
    cpy_r_r26 = 1;
              ^ ~
1 error generated.
```
How can we debug this further?

TH3CHARLie:
> @TH3CHARLie Could you have a look at the mypyc test failures on Travis here? I see this error:
> 
> ```
> build/__native_dbbd1b679b3c31e8ff5a.c:972286:15: error: incompatible integer to
>       pointer conversion assigning to 'PyObject *' (aka 'struct _object *') from
>       'int' [-Werror,-Wint-conversion]
>     cpy_r_r26 = 1;
>               ^ ~
> 1 error generated.
> ```
> 
> How can we debug this further?

I am guessing this is an incorrect true op assignment though I am not sure why the types are incompatible, I am now compiling this branch to find out.

TH3CHARLie:
@gvanrossum I think this is a mypyc bug that mypyc has incorrectly implement the `next` op since I can reproduce the error with commit b8bdc26127d7b7f3ac42b308350b0ec0fba01fcc (before all the GSOC work happens).

A minimal example:
```python
from typing import List, Optional

def foo(x: List[int]) -> Optional[int]:
    res = next(
        (i for i in x if i > 4), None)
    if res is None:
        return None
    else:
      return res
```

gvanrossum:
> I think this is a mypyc bug that mypyc has incorrectly implement the `next` op since I can reproduce the error with commit [b8bdc26](https://github.com/python/mypy/commit/b8bdc26127d7b7f3ac42b308350b0ec0fba01fcc) (before all the GSOC work happens).

Thanks @TH3CHARLie for the quick diagnosis! Sorry if it sounded like I was pointing a finger to your GSoC work -- I didn't mean that, I just thought you'd be quicker at figuring this out than me, and Jukka's asleep. :-) Do you want to file an issue or do you have a good idea on how to fix it already?

@mgilson If you can rewrite the code without using `next(<genexp>, None)` we can land this right away. Maybe define a function like this?
```
def first(it: Iterable[T]) -> Optional[T]:
    for x in it:
        return x
    return None
```

TH3CHARLie:
> Sorry if it sounded like I was pointing a finger to your GSoC work

I wasn't considering this at all, it's just to give a clear scope of where this would happen. Since the GSOC project involves plenty of refactoring, making sure it is not originating from that helps us to differentiate a migration miss and an existing bug.

> Do you want to file an issue or do you have a good idea on how to fix it already?
I am still trying to produce a fix but most likely this would require a specialized `next` op. I'll file an issue at the mypyc repository to track the progress.

mgilson:
I removed the `next` from in there.  Hopefully it passes on the next run 🤞

Issue #9290, opened 2020-08-11
Release 0.790 planning


JukkaL:
I plan to release mypy 0.790 around Aug 25 and cut the release branch this Thu (Aug 13).

Please add comments to this issue if you have any PR you'd like to be included. Ideally those PRs should be landed before the release branch is cut, to minimize cherry-picking.

(This will hopefully be the final major release with bundled monolithic typeshed. 0.800 should have third-party package stubs available from PyPI.)

JukkaL:
I'm encountering what looks like a hard crash (no traceback) on Linux with Python 3.6 when running the latest master against Dropbox internal code.

It started happening after https://github.com/python/mypy/commit/88eb84ec182c3e638aa15fe0ce7c9673059def7b was merged. Looking at that PR, I don't see anything particularly suspicious, but it's doing some pointer casting which might plausibly be a problem. If I can't fix the crash soon, I might cut the release branch from an earlier commit or wait a bit longer until I cut the release branch.

JukkaL:
I narrowed down the segfault to https://github.com/mypyc/mypyc/issues/756. It's a release-blocker, unless we cut the release branch from an earlier commit and do some cherry-picks.

msullivan:
Is the branch cut happening today? I've been seeing infinite loops with --follow-imports=normal on https://github.com/edgedb/ and will try to investigate and get a PR up today.

TH3CHARLie:
> Is the branch cut happening today? I've been seeing infinite loops with --follow-imports=normal on https://github.com/edgedb/ and will try to investigate and get a PR up today.

I think it will happen today since the fix for mypyc/mypyc#756 would be then included and no cherry-picking is needed.

msullivan:
I'd like https://github.com/python/mypy/pull/9302 to make it into the release

hauntsaninja:
I'm late to the party, but it'd be nice if this typeshed sync https://github.com/python/mypy/pull/9317 containing https://github.com/python/typeshed/pull/4447 made it in

TH3CHARLie:
> I'm late to the party, but it'd be nice if this typeshed sync #9317 containing [python/typeshed#4447](https://github.com/python/typeshed/pull/4447) made it in

according to the branches page, the release 0.790 branch was already cut last Friday.

JukkaL:
[...]

PR #9445, opened 2020-09-15
Recognize -stubs directories as valid package directories

### LINKED ISSUE #8229: mypy fails with *-stubs is not a valid Python package name
According to PEP 561 [stub-only package](https://www.python.org/dev/peps/pep-0561/#stub-only-packages):

> The name of the stub package MUST follow the scheme foopkg-stubs for type stubs for the package named foopkg. 

However if we invoke mypy on such package providing only path it fails with 

>  *-stubs is not a valid Python package name

This can be reproduced on the [example package](https://github.com/ethanhs/stub-package):

```
git clone https://github.com/ethanhs/stub-package
cd stub-package
mypy ./typedpkg-stubs
typedpkg-stubs is not a valid Python package name
```

Invoking `mypy` with `-p` option works

```
mypy -p typedpkg-stubs
Success: no issues found in 2 source files
```

but the former behavior is rather confusing.

Expected behavior:

- mypy should recognize `*-stubs` paths as roots of stub-only packages.

Tested with:

- mypy 0.761
- mypy-0.770+dev.9101707bd0c96624d09cb31fe573d7e25c89a35c

bryanforbes:
### Description

Currently, running `mypy typedpkg-stubs` (or any sub-file contained in the directory) in a project similar to https://github.com/ethanhs/stub-package results in an error:

```
mypy ./typedpkg-stubs
typedpkg-stubs is not a valid Python package name
```

This PR allows directories prefixed with `-stubs` which also contains a valid identifier before the `-` to be type checked.

Fixes #8229

### DIFF
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -127,6 +127,10 @@ class SourceFinder:
             res = ''
             base_dir = dir or '.'
         else:
+            # Remove -stubs to treat PEP-561 stub-only directories as packages
+            if base.endswith('-stubs'):
+                base = base[:-6]
+
             # Ensure that base is a valid python module name
             if not base.isidentifier():
                 raise InvalidSourceList('{} is not a valid Python package name'.format(base))
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -127,11 +127,9 @@ class SourceFinder:
             res = ''
             base_dir = dir or '.'
         else:
-            # Remove -stubs to treat PEP-561 stub-only directories as packages
-            if base.endswith('-stubs'):
-                base = base[:-6]
-
             # Ensure that base is a valid python module name
+            if base.endswith('-stubs'):
+                base = base[:-6]  # PEP-561 stub-only directory
             if not base.isidentifier():
                 raise InvalidSourceList('{} is not a valid Python package name'.format(base))
             parent, base_dir = self.crawl_up_dir(parent_dir)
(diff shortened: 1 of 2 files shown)


mypy/find_sources.py
```diff
@@ -127,6 +127,10 @@ def crawl_up_dir(self, dir: str) -> Tuple[str, str]:
             res = ''
             base_dir = dir or '.'
         else:
+            # Remove -stubs to treat PEP-561 stub-only directories as packages
+            if base.endswith('-stubs'):
+                base = base[:-6]
+
             # Ensure that base is a valid python module name
```
gvanrossum, line 134: Somehow I initially misread this as having a global effect -- but in fact `base` is only used in the check `base.isidentifier()` below, so it's a throwaway variable. How about moving stuff around so the other comment still leads?

```suggestion
            # Ensure that base is a valid python module name
            if base.endswith('-stubs'):
                base = base[:-6]  # PEP-561 stub-only directory
```

gvanrossum (commented):
Thanks! I have one style nit. Otherwise looks good.

mypy/find_sources.py
bryanforbes: Was the removal of the empty line after the assignment intentional?

gvanrossum: Yes.

gvanrossum (approved):

bryanforbes:
Is there any way this can be included in 0.790?

gvanrossum:
I very much doubt it -- according to https://github.com/python/mypy/issues/9290, the release branch was cut quite a while ago, and cherry-picking must be minimized so as not to invalidate the extensive testing that has already been applied to the release branch.

bryanforbes:
@gvanrossum makes sense. Thanks!

PR #9452, opened 2020-09-17
types overlap only if all type args match

### LINKED ISSUE #9451: Overload with Dict only checks overlap on keys
Using mypy 0.782, it looks like overlap is only checked on dictionary keys.

```
from typing import Dict, overload

class A: pass
class B: pass

@overload
def foo(x: Dict[A, A]) -> A: ...

@overload
def foo(x: Dict[A, B]) -> B: ...

def foo(x):
    for k, v in x: return v
    raise ValueError
```

gives `error: Overloaded function signatures 1 and 2 overlap with incompatible return types`. I get the same error when using `Mapping` instead of `Dict.

JelleZijlstra:
Fixes #9451

### Description

Previously, generic types were considered overlapping as soon as the first type argument matched. Now, we check all type arguments.

## Test Plan

The new test failed before my change and now passes.


hauntsaninja (approved):
Looks good to me!

mypy/meet.py
```diff
@@ -339,9 +339,9 @@ def _type_object_overlap(left: Type, right: Type) -> bool:
             # Or, to use a more concrete example, List[Union[A, B]] and List[Union[B, C]]
             # would be considered partially overlapping since it's possible for both lists
             # to contain only instances of B at runtime.
```
gvanrossum, line 341: Looking at this example I am wondering if the original behavior wasn't intentional?

JelleZijlstra: The example only has a single type argument; I think the case with multiple args was just not considered.

hauntsaninja: I stumbled a little on this too, unions are dealt with by the recursive call, about a hundred lines earlier in the function

JukkaL (approved):
Looks this was an oversight in the original implementation.

PR #9462, opened 2020-09-20
Consider AnyType for TypeVar upperbound to avoid missing attribute

### LINKED ISSUE #4351: TypeVar with Any upper bound generates missing attribute errors is used as Type[T]
Imagine the following code:

```
class Foo():
    NAME: ClassVar[str] = ""

T = TypeVar("T", bound=Foo)

def create_foo(object_class: Type[T]) -> T:
    print(T.NAME)  # Generates error in static analysis
    return T()
```

When this code is run through the static analysis, the marked line is flagged as an error because mypy thinks that `T` does not have the `NAME` attribute. When the same code is run through the dynamic analysis, the code works fine.

kamilturek:
Fixes #4351

Using TypeVar upperbound with imported type and disabled imports following was causing missing attribute error since AnyType was not considered as upperbound option.

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -280,6 +280,9 @@ def analyze_type_type_member_access(name: str,
             item = upper_bound
         elif isinstance(upper_bound, TupleType):
             item = tuple_fallback(upper_bound)
+        elif isinstance(upper_bound, AnyType):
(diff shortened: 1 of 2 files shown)


gvanrossum (approved):
Thanks for fixing this ancient bug!

PR #9470, opened 2020-09-23
Unpin package versions

### LINKED ISSUE #9401: Unpin setuptools
https://github.com/python/mypy/pull/9400 pins us to setuptools<50

Once the burning has stopped (to quote https://github.com/pypa/setuptools/pull/2376), we should unpin this. One day, everything will work.

Also while I'm here, a couple months ago I pinned virtualenv<20 to fix CI. virtualenv might work for us now, so another thing we could unpin.

Cyatos:
Unpinned versions of the  following packages:
 - setuptools
 ~~- virtualenv~~

Fixes #9401

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
         python-version: ${{ matrix.python }}
         architecture: ${{ matrix.arch }}
     - name: install tox
-      run: pip install --upgrade setuptools virtualenv tox==3.9.0
+      run: pip install --upgrade 'setuptools!=50' virtualenv tox==3.9.0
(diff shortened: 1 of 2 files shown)


hauntsaninja:
Thanks for doing this! The test failures are related to the virtualenv unpinning :-(
Unpinning setuptools is more important, so if you revert the virtualenv unpinning, I'll merge. Alternatively, feel free to figure out what needs to be done to make tests work with the newer virtualenv!

Cyatos:
Happy to help! Re-pinned virtualenv version for now. Will raise another PR if and when I fix the issue with virtualenv's versioning.

PR #9477, opened 2020-09-24
Speed up typechecking of dict, set and list expressions

huguesb:
Typechecking of dict, set, and list literals currently
goes through typechecking of the generic dict/set/list
constructor internally. This is usually fine but becomes
horrendously slow when the number of items is large:
 - for generic methods, `infer_arg_types_in_context` is
   called twice
 - `infer_arg_types_in_context` is O(n**2) where `n` is
   the number of arguments, which, in the case of a
   literal, is the number of items.

Add an `O(n)` fast path for deriving the type of simple
container literal expressions. This fast path only handle
a subset of cases but it provides a tremendous speedup for
the relatively common case of large literal constants.

The real-world example that motivated this change is a
1889 lines long dict constant representing the parsed value
of a mock JSON response from a 3rd party service, where
typechecking previously took upwards of 50s and is now
down to under 1s with this fast path.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3178,6 +3178,16 @@ class ExpressionChecker(ExpressionVisitor[Type]):
     def fast_container_type(
             self, items: List[Expression], container_fullname: str
     ) -> Optional[Type]:
+        """
+        Fast path to determine the type of a list or set literal,
+        based on the list of entries. This mostly impacts large
+        module-level constant definitions.
+
+        Limitations:
+         - no active type context
+         - no star expressions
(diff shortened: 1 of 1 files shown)


msullivan (commented):
This is fantastic and I'm thrilled to merge it. This has been a pretty silly pain point for a while. Could you add docstrings to the fast path functions explaining the rationale and what cases are covered by the fast paths?

huguesb:
Added docstrings as requested

msullivan:
Thank you!

hauntsaninja:
Just wanted to chime in and say this is great! :-) Thanks again!

PR #9481, opened 2020-09-25
Fix assignment of 'builtins.type' inferred as 'builtins.object'

### LINKED ISSUE #9476: regression: assignment of 'builtins.type' now type-inferred as 'builtins.object'
**Bug Report**

#7963 causes a serious regression

After this commit, the use of `from __future__ import annotations` causes assignments of `builtins.type` to be typed as `builtins.object`, which prevents the assigned variable from being used as a callable later on

This is especially bad because the semantic analyzer is shared and does not maintain the state of `__future__` flags on a per-module basis, therefore *any* file with that import will affect *all* files that are typechecked after it (which makes it hard to pinpoint in a large codebase, given the way the graph is processed...)

**To Reproduce**

Simple reproducer in a single file:

```
reveal_type(type)
_TYPE = type
reveal_type(_TYPE)
_TYPE('bar')
```

**Expected Behavior**

```
simple.py:2: note: Revealed type is 'Overload(def (o: builtins.object) -> builtins.type, def (name: builtins.str, bases: builtins.tuple[builtins.type], dict: builtins.dict[builtins.str, Any]) -> builtins.type)'
simple.py:4: note: Revealed type is 'Overload(def (o: builtins.object) -> builtins.type, def (name: builtins.str, bases: builtins.tuple[builtins.type], dict: builtins.dict[builtins.str, Any]) -> builtins.type)'
```

**Actual Behavior**

```
simple.py:2: note: Revealed type is 'Overload(def (o: builtins.object) -> builtins.type, def (name: builtins.str, bases: builtins.tuple[builtins.type], dict: builtins.dict[builtins.str, Any]) -> builtins.type)'
simple.py:4: note: Revealed type is 'builtins.object'
simple.py:5: error: "object" not callable
```

**Your Environment**

- Mypy version used: 0.790+dev.720b77e2ded8c46b66647cdde9218769d33aec41 (or any commit thereafter)
- Mypy command-line flags: `--python-version 3.7`
- Python version used: 3.7
- Operating system and version: macOS 10.14

TH3CHARLie:
Fixes #9476

### DIFF
--- a/mypy/typeanal.py
+++ b/mypy/typeanal.py
@@ -304,12 +304,17 @@ class TypeAnalyser(SyntheticTypeVisitor[Type], TypeAnalyzerPluginInterface):
             return make_optional_type(item)
         elif fullname == 'typing.Callable':
             return self.analyze_callable_type(t)
-        elif (fullname == 'typing.Type'):
+        elif (fullname == 'typing.Type' or
+             (fullname == 'builtins.type' and self.api.is_future_flag_set('annotations'))):
             if len(t.args) == 0:
-                any_type = self.get_omitted_any(t)
-                return TypeType(any_type, line=t.line, column=t.column)
+                if fullname == 'typing.Type':
+                    any_type = self.get_omitted_any(t)
+                    return TypeType(any_type, line=t.line, column=t.column)
+                else:
+                    return None
+            type_str = 'Type[...]' if fullname == 'typing.Type' else 'type[...]'
             if len(t.args) != 1:
-                self.fail('Type[...] must have exactly one type argument', t)
+                self.fail(type_str + ' must have exactly one type argument', t)
             item = self.anal_type(t.args[0])
             return TypeType.make_normalized(item, line=t.line)
         elif fullname == 'typing.ClassVar':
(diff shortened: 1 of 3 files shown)


huguesb:
I would suggest adding the example from #9476 as a regression test

TH3CHARLie:
> I would suggest adding the example from #9476 as a regression test

Thanks for the suggestion. I've added that test case. The different type signature is due to the simplified typeshed file used in tests, the point here is to prevent it from becoming `builtins.object` .

hauntsaninja (commented):
Thanks for patching quickly! Curious what you feel the real fix looks like?
It seems to me like adding another branch when `len(t.args) == 0`, `fullname == 'builtins.type'` and `is_future_flag_set('annotations')` and returning None could be the way to go.

TH3CHARLie:
> Thanks for patching quickly! Curious what you feel the real fix looks like?
> It seems to me like adding another branch when `len(t.args) == 0`, `fullname == 'builtins.type'` and `is_future_flag_set('annotations')` and returning None could be the way to go.

Thanks for the suggestion! Nice fix!

hauntsaninja (approved):
Looks good to me, though might be worth leaving a comment. But you're also welcome to just merge :-)

gvanrossum:
I'll trust @hauntsaninja, this area of the code would take me a lot of time to get into again.

TH3CHARLie:
Thanks for the review @hauntsaninja @gvanrossum! I've added a comment to point here.

Issue #9437, opened 2020-09-11
strict_optional=False makes checking scripts using xarray to fail


dalonsoa:
## 🐛 Bug Report

When using mypy with `strict_optional = False` in the config file or using the `--no-strict-optional` flag in the terminal to check a file using the `xarray` package, it fails with an scary INTERNAL ERROR message.

This might be a bug better suited for the `xarray` team, but thought to check with you first.

## To Reproduce

Write a python script just importing `xarray`:

```python
import xarray as xr
```

And check it with mypy using `mypy --no-strict-optional script.py`

## Expected Behavior

Everything works fine and I get: `Success: no issues found in 1 source file` . This is indeed the behavior if the `--no-strict-optional` flag is NOT used.

## Actual Behavior

I get the following output (personal info replaced by XXXXX):

```
/XXXXX/venv/lib/python3.7/site-packages/xarray/core/dataarray.py:2001: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
If this issue continues with mypy master, please report a bug at https://github.com/python/mypy/issues
version: 0.782
/XXXXX/venv/lib/python3.7/site-packages/xarray/core/dataarray.py:2001: : note: please use --show-traceback to print a traceback when reporting a bug
```

I've tried also the development version of mypy, as sugested, with identical results, and also used the `--show-traceback` option, with results in:

```
/XXXXX/venv/lib/python3.7/site-packages/xarray/core/dataarray.py:2001: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 0.782
Traceback (most recent call last):
  File "mypy/checker.py", line 401, in accept
  File "mypy/nodes.py", line 1129, in accept
  File "mypy/checker.py", line 3437, in visit_for_stmt
  File "mypy/checker.py", line 3470, in analyze_iterable_item_type
  File "mypy/checkexpr.py", line 2345, in check_method_call_by_name
  File "mypy/checkmember.py", line 126, in analyze_member_access
  File "mypy/checkmember.py", line 143, in _analyze_member_access
  File "mypy/checkmember.py", line 225, in analyze_instance_member_access
  File "mypy/checkmember.py", line 376, in analyze_member_var_access
  File "mypy/checkmember.py", line 578, in analyze_var
[...]

PR #9495, opened 2020-09-28
Fix partial type crash during protocol checking

hauntsaninja:
In particular, this affected hashables, as in #9437 

Maybe there's a more standard way of falling back the PartialType? We could also consider moving this logic into is_subtype, but I'd be hesitant to do so. I'm okay with returning False based on the fallback Instance constructed here, but not returning True.

### DIFF
--- a/test-data/unit/check-protocols.test
+++ b/test-data/unit/check-protocols.test
@@ -2536,3 +2536,38 @@ class EmptyProto(Protocol): ...
 def hh(h: EmptyProto) -> None: pass
 hh(None)
 [builtins fixtures/tuple.pyi]
+
+
+[case testPartialTypeProtocol]
+from typing import Protocol
+
+class Flapper(Protocol):
+    def flap(self) -> int: ...
+
+class Blooper:
+    flap = None
+
+    def bloop(self, x: Flapper) -> None:
+        reveal_type([self, x])  # N: Revealed type is 'builtins.list[builtins.object*]'
+
+class Gleemer:
+    flap = []  # E: Need type annotation for 'flap' (hint: "flap: List[<type>] = ...")
+
+    def gleem(self, x: Flapper) -> None:
+        reveal_type([self, x])  # N: Revealed type is 'builtins.list[builtins.object*]'
+[builtins fixtures/tuple.pyi]
+
+
+[case testPartialTypeProtocolHashable]
+# flags: --strict-optional
+from typing import Protocol
+
+class Hashable(Protocol):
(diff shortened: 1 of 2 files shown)


test-data/unit/check-protocols.test
```diff
@@ -2536,3 +2536,38 @@ class EmptyProto(Protocol): ...
...
+
+class Hashable(Protocol):
+    def __hash__(self) -> int: ...
+
+class DataArray:
+    __hash__ = None
+
+    def f(self, x: Hashable) -> None:
+        reveal_type([self, x])  # N: Revealed type is 'builtins.list[builtins.object*]'
+[builtins fixtures/tuple.pyi]
```
gvanrossum, line 2573: MIssing line ending.

test-data/unit/check-protocols.test
```diff
@@ -2536,3 +2536,38 @@ class EmptyProto(Protocol): ...
...
+class Gleemer:
+    flap = []  # E: Need type annotation for 'flap' (hint: "flap: List[<type>] = ...")
+
+    def gleem(self, x: Flapper) -> None:
+        reveal_type([self, x])  # N: Revealed type is 'builtins.list[builtins.object*]'
+[builtins fixtures/tuple.pyi]
+
+
+[case testPartialTypeProtocolHashable]
+# flags: --strict-optional
```
gvanrossum, line 2562: Do you need this? I thought it was the default. Or are the tests run multiple times with different default flags?

gvanrossum (approved):
This is a relief. Thanks very much for identifying the true cause and providing a fix!

test-data/unit/check-protocols.test
hauntsaninja: Oh, I meant to use `--no-strict-optional` as in https://github.com/python/mypy/issues/9437#issuecomment-699788746 — but a test a couple lines up used `--strict-optional` and I copied it by accident. Thanks!

Hmm. It actually looks like `--no-strict-optional` is the default in tests. I just discovered this file lurking in the shadows: https://github.com/python/mypy/blob/8bf770d63cc4ac656dde00904b61d7f1b6fba68e/test-data/unit/hacks.txt#L8

Also `object` in the test stubs doesn't have `__hash__`... so this test didn't actually test anything different from Blooper (and crashes regardless of `--strict-optional`). Fixed now (ie, the new test and old code only crashes with `--no-strict-optional`).

gvanrossum (approved):
Yay!

Issue #9497, opened 2020-09-29
False negative with calling a TypeVar


KotlinIsland:
```py
from typing import TypeVar

T = TypeVar("T")
T() # runtime error: TypeError: 'TypeVar' object is not callable
```

hauntsaninja:
Thanks, duplicate of #9244

hauntsaninja:
Opened a PR with a fix! Thanks again :-)

PR #9498, opened 2020-09-29
checkexpr: treat typevar values as object, not Any

### LINKED ISSUE #9244: mypy doesn't complain about TypeVars used as values in expressions
The following code currently type checks:
```
from typing import TypeVar
T = TypeVar("T")
x = 5
x + T
T()
def f(x): ...
f(T)
```

hauntsaninja:
Fixes #9244 / #9497

### DIFF
--- a/test-data/unit/check-expressions.test
+++ b/test-data/unit/check-expressions.test
@@ -2808,6 +2808,5 @@ def f() -> int:  # E: Missing return statement
 from typing import TypeVar
 T = TypeVar("T")
 x: int
-x + T  # E: Unsupported left operand type for + ("int")
+x + T  # E: Unsupported operand types for + ("int" and "object")
 T()  # E: "object" not callable
-[builtins fixtures/tuple.pyi]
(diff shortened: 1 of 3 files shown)


test-data/unit/check-expressions.test
```diff
@@ -2803,3 +2803,11 @@ def f() -> int:  # E: Missing return statement
     x: int
     assert isinstance(x, int), '...'
 [builtins fixtures/isinstance.pyi]
+
+[case testTypeVarAsValue]
+from typing import TypeVar
+T = TypeVar("T")
+x: int
+x + T  # E: Unsupported left operand type for + ("int")
```
gvanrossum, line 2811: Hm, isn't this a little weird? I'd think that the right operand type is more suspect... :-)

hauntsaninja: Good spot, it is a little weird. It took me some pdb, but the difference is that `int` in [builtins fixtures/tuple.pyi] doesn't have `__add__`. The fixture is actually unnecessary, so it's gone now :-)

gvanrossum (approved):

PR #9512, opened 2020-09-30
Fix runtime not generic note for type aliases, update for PathLike

hauntsaninja:
As a result of https://github.com/python/typeshed/pull/4586


gvanrossum (approved):

hauntsaninja:
This turned out to require some logic change, so re-requesting review :-)

hauntsaninja:
I'm going to go ahead and merge this to unblock typeshed, but happy to make any improvements should any come up!

PR #9528, opened 2020-10-03
Travis CI: install typing on python2 to restore some skipped tests

### LINKED ISSUE #9473: Some Python2 tests are skipped in CI
https://github.com/python/mypy/pull/9196 appears to have broken tests (https://github.com/python/typeshed/runs/1157402752?check_suite_focus=true, testAssignToComplexReal_python2). It looks like this is because it needs a Python 2 with typing installed. If that isn't the case, the test gets skipped.

vbarbaresi:
### Description

Fixes #9473

Typing used to be installed on python2 but was removed by [this change](https://github.com/python/mypy/commit/def2c63990dd#diff-354f30a63fb0907d4ad57269548329e3L15), I don't know why

in `mypy/util.py:try_find_python2_interpreter()` we try to find a valid python2 interpreter with typing installed.
If we don't find it, those tests are skipped.

However on [typeshed](https://github.com/python/typeshed/blob/master/tests/mypy_selftest.py#L14), the CI installs typing on python2 for mypy self test. Thus it doesn't skip the tests.

This caused the regression issue on typeshed

## Test Plan

I ran Travis CI on my forked repo.
I based a branch on 31862816ee87afc2b7514aec76249b884a9fdbef (the one that caused a python2 regression) and was able to reproduce the failing test: https://travis-ci.org/github/vbarbaresi/mypy/jobs/732473410
I noted that this change unskips around 40 tests (all from `testpythoneval.py`)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,6 +85,7 @@ install:
 - pip install -U pip setuptools
 - pip install -U 'virtualenv<20'
 - pip install -U tox==3.9.0
+- python2 -m pip install --user -U typing
 - tox --notest
 
 # This is a big hack and only works because the layout of our directories


vbarbaresi:
I just read in `testpythoneval.py`:
> Note: These test cases are *not* included in the main test suite, as including
      this suite would slow down the main suite too much.

So it looks like this was removed on purpose. Closing.

hauntsaninja:
Thanks for looking into this and fixing!
I think it might actually have been unintentional, eg since we do run the other pythoneval tests in CI / it's not called out in the PR you linked. I think the comment might be referring to the fact that the evaluation tests are separated out here: https://github.com/python/mypy/blob/3e77959eacf3d445a0cb4db5a4bc6dcf606fc040/runtests.py#L22
I'm going to reopen this, and maybe someone else who knows for sure can chime in :-)

JukkaL:
The removal of Python 2 typing seems accidental. It would be nice to run the tests in CI.

hauntsaninja:
Great! Thanks again, vbarbaresi

PR #9529, opened 2020-10-03
[mypyc] Add primitives for bitwise ops

JukkaL:
This adds primitives for `&`, `|`, `^`, `<<`, `>>` and `~`. These speed up a
bitwise op microbenchmark by about 10x on Linux.

Long integer operations are only slightly faster, except for certain cases
of mixed short/long integers, where we are significantly faster (but only
maybe 1.5x or so).

Also support `bool` when relevant, since it's a subclass of `int` and 
overrides some operations.

Work on mypyc/mypyc#751.


JukkaL:
Hmm I'm going to have to specialize some of the ops for booleans as well, as otherwise the integer ops will be used, resulting in runtime type errors.

TH3CHARLie (approved):
Overall LGTM and that's a great improvement!

PR #9530, opened 2020-10-04
semanal: propagate module_hidden for unknown symbols

### LINKED ISSUE #9517: Missing imports don't obey --no-implicit-reexport
Noticed this while making #9515, will fix once it's merged. Related to the problem behind #8649, which I fixed a while ago, but was timid about and hence didn't think about this bug at that time.

```
~/tmp/rexp λ cat x.py
from mystery import a, b as b, c as d
~/tmp/rexp λ cat y.py
from x import a
from x import b
from x import c
from x import d
~/tmp/rexp λ mypy --no-implicit-reexport .
x.py:1: error: Cannot find implementation or library stub for module named 'mystery'
x.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
y.py:3: error: Module 'x' has no attribute 'c'
Found 2 errors in 2 files (checked 2 source files)
```

whereas you'd expect:
```
~/tmp/rexp λ mypy --no-implicit-reexport . 
x.py:1: error: Cannot find implementation or library stub for module named 'mystery'
x.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
y.py:1: error: Module 'x' does not explicitly export attribute 'a'; implicit reexport disabled
y.py:3: error: Module 'x' has no attribute 'c'
y.py:4: error: Module 'x' does not explicitly export attribute 'd'; implicit reexport disabled
Found 4 errors in 2 files (checked 2 source files)
```

hauntsaninja:
Fixes #9517

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -1776,11 +1776,15 @@ class SemanticAnalyzer(NodeVisitor[None],
                 )
             elif module and not missing_submodule:
                 # Target module exists but the imported name is missing or hidden.
-                self.report_missing_module_attribute(module_id, id, imported_id, imp)
+                self.report_missing_module_attribute(
+                    module_id, id, imported_id, module_public=module_public,
+                    module_hidden=not module_public, context=imp
+                )
             else:
                 # Import of a missing (sub)module.
                 self.add_unknown_imported_symbol(
-                    imported_id, imp, target_name=fullname, module_public=module_public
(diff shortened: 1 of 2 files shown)


gvanrossum (approved):
Gosh, 'module_public' and 'module_hidden' are really confusing names. If it weren't for the docstring for SymbolTableNode explaining when they're not each other's opposite I'd have gone insane. :-)

PR #9534, opened 2020-10-05
stubgen: fix package imports with aliases

chadrik:
### Description

Running `stubgen` on modules that import multiple packages using aliases did not produce the correct results:

Input:

```python
import p.a as a
import p.b as b

x: a.X
y: b.Y
```

Result:

  ```python
  import p.b as a  # <--- wrong module!
  import p.b as b
  
  x: a.X
  y: b.Y
  ```

## Test Plan

I added a bunch of new tests to stubgen.test.  

I had to change one of the expected results: `testConditionalImportAll_semanal`.  Previously, if more than one module import introduced the same name into the namespace, the data structures used to track imports were simply allowed to become corrupted, and the generator would produce `import FIXME as whatever`.  After correcting the internal state to fix the issue that prompted this PR, I realized it would be difficult to maintain the current output for this test, so I chose to change the behavior.   I went with what I think is a sane solution:  if multiple import statements target the same name, rather than produce invalid output, use the last import encountered.  The last import would be the one used at runtime.   I also considered using the first import, since this is what mypy would use.

### DIFF
--- a/test-data/unit/stubgen.test
+++ b/test-data/unit/stubgen.test
@@ -1577,7 +1577,7 @@ else:
     import cookielib
 
 [out]
-import FIXME as cookielib
+import cookielib as cookielib
 
 [case testCannotCalculateMRO_semanal]
 class X: pass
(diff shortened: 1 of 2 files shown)


test-data/unit/stubgen.test
```diff
@@ -1577,7 +1577,7 @@ else:
     import cookielib
 
 [out]
-import FIXME as cookielib
+import cookielib as cookielib
```
chadrik, line 1580: for reference, here is the complete test:


```

[case testConditionalImportAll_semanal]
__all__ = ['cookielib']

if object():
    from http import cookiejar as cookielib
else:
    import cookielib

[out]
import cookielib as cookielib
```

hauntsaninja (approved):
Thanks for finding and fixing this! Despite the review request, this is actually my first time really looking at stubgen code...
Anyway, this looks good to me. Your (very thorough) test cases all check out and the changes make sense.

I was a little bit worried about the FIXME test case that needed changing... On first read, I thought it was intended as a feature for stubgen to highlight conditional imports with a FIXME. But it looks like that isn't what's happening here, since we're happy enough to clobber `reverse_alias`. I.e., the change in behaviour here seems consistent with things stubgen does in other places.

I'll leave open for a day or two, just in case someone else who's worked with stubgen more has something to say.

PR #9537, opened 2020-10-05
Add disable/enable error code opt to config_types

amikrop:
The `disable_error_code` and `enable_error_code` config file options were being cast to list, as they were not
included in the config_types dict in config_parser.py.


hauntsaninja (commented):
Lgtm, thanks for reporting and fixing! If you wanted to add a test, `test-data/unit/check-flags.test` would be the place to do it

amikrop:
OK, going to add a test

hauntsaninja:
Closing in favour of #9538 then :-)

PR #9539, opened 2020-10-05
Document --disable-error-code cli and config option

amikrop:
Added the --disable-error-code cli option and disable_error_code config option to the documentation.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -480,6 +480,11 @@ of the above sections.
            # 'items' now has type List[List[str]]
            ...
 
+.. option:: --disable-error-code CODE
+
+    This flag will cause mypy to suppress errors with error code ``CODE``.
+    This flag may be repeated.
+
 .. option:: --local-partial-types
 
     In mypy, the most common cases for partial types are variables initialized using ``None``,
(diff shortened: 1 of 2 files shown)


docs/source/command_line.rst
```diff
@@ -480,6 +480,11 @@ of the above sections.
            # 'items' now has type List[List[str]]
            ...
 
+.. option:: --disable-error-code CODE
```
hauntsaninja, line 483: The command line option is already documented on line 556.

amikrop: OK removed the duplicate entry I added

docs/source/config_file.rst
```diff
@@ -529,6 +529,12 @@ Miscellaneous strictness flags
     Allows variables to be redefined with an arbitrary type, as long as the redefinition
     is in the same block and nesting level as the original definition.
 
+.. confval:: disable_error_code
+
+    :type: comma-separated list of strings
+
+    Causes mypy to suppress errors with given error codes.
```
hauntsaninja, line 536: One nit: I'd steal the language from the command line docs here (for consistency, and for the mention of "globally").
```suggestion
    Allows disabling one or multiple error codes globally.
```

(More generally, duplication between command line docs and config docs seems like a bit of a maintenance headache)

hauntsaninja (approved):
Thanks again!

amikrop:
> Thanks again!

You are welcome.

PR #9543, opened 2020-10-07
Include conftest.py in source distributions

sbraz:
### Description
Hi,
This ensures `conftest.py` will be present in source distributions. When running tests from PyPI's `mypy-0.782.tar.gz`, only a fraction are collected.

## Test Plan
I don't think there is anything special to test. I have successfully run all the tests from PyPI's `mypy-0.782.tar.gz` after re-adding `conftest.py`.


msullivan:
Thanks! Looks like this got picked up as part of #9592, a big MANIFEST redo

PR #9547, opened 2020-10-07
Bump the typeshed pin

nipunn1313:
### Description

Bumps the typeshed pin to pick up latest changes

## Test Plan

Typeshed passes in CI - so can rely on existing CI here.


JukkaL (approved):
Thanks!

Issue #9536, opened 2020-10-05
Speed up mypy wheel builds


JukkaL:
Travis mypy release wheel builds may take around 1.5 hours to complete: https://travis-ci.org/github/mypyc/mypy_mypyc-wheels/builds

This makes iteration on mypy release builds very slow. 

Ideas about how to make these faster:
- Move at least some of the jobs to GitHub actions. It should give us more parallelism.
- Skip some tests that are slow and relatively low signal, at least in the slowest jobs.

Any help here would be appreciated, as this is one of the main causes of friction when making mypy releases. If we could get these to around 30 minutes, things would already be much better.

Note that the build scripts live in a different repository: https://github.com/mypyc/mypy_mypyc-wheels

hauntsaninja:
It seems just moving everything to Github Actions would do it. They allow for 5 concurrent macOS jobs, as opposed to Travis' 2. They also allow 20 concurrent jobs as opposed to Travis' 5.

While we're making changes here, we should also do what's needed for Python 3.9.

gvanrossum:
I'm not objecting to the move, but I note that perhaps this is an attempt by GitHub to kill the competition? After all nothing says GitHub can't reduce these limits after Travis is out of business (or even just heading in that direction).

hauntsaninja:
It's true; the change to have "details" click through to Github Actions for Travis build status rather than linking to Travis was something I found striking. More conspiratorially I think some of the difficulties Travis has had reporting PR status back to Github might be somewhat intentional.

I'd worry about it more if mypy_mypyc_wheels contributed positively to Travis' bottom line (I didn't even know that the project existed until today, so it isn't even providing them marketing value).

hauntsaninja:
I put together a working version of mypy_mypyc-wheels on Github Actions at https://github.com/hauntsaninja/mypy_mypyc-wheelsv2 . It currently takes pretty much exactly half an hour to run:
https://github.com/hauntsaninja/mypy_mypyc-wheelsv2/actions/runs/296713743
https://github.com/hauntsaninja/mypy_mypyc-wheelsv2/releases/tag/tag-7131a66aeed027daa6d83818870c437dbcbebaef

[...]

PR #9552, opened 2020-10-08
py39: fix mypyc complaint

hauntsaninja:
I was trying to build wheels for Python 3.9 as part of #9536, but ran
into this issue. You'll notice a couple hundred lines up msullivan
points out that mypyc can't handle conditional method definition, so
that's not an option here.

```
/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmpok7vxky0
    mypy/fastparse.py:1260: error: "ExtSlice" has no attribute "dims"
    mypy/fastparse.py:1264: error: "Index" has no attribute "value"
ERROR: Command errored out with exit status 1: /opt/_internal/cpython-3.9.0rc2/bin/python3.9 /opt/_internal/cpython-3.9.0rc2/lib/python3.9/site-packages/pip/_vendor/pep517/_in_process.py prepare_metadata_for_build_wheel /tmp/tmpok7vxky0 Check the logs for full command output.
```

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -1257,11 +1257,13 @@ class ASTConverter:
 
     # ExtSlice(slice* dims)
     def visit_ExtSlice(self, n: ast3.ExtSlice) -> TupleExpr:
-        return TupleExpr(self.translate_expr_list(n.dims))
+        # cast for mypyc's benefit on Python 3.9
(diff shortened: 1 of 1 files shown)


msullivan (approved):
Sigh!

PR #9557, opened 2020-10-08
[mypyc] optimise startswith and endswith

r3m0t:
### Description

Relates to https://github.com/mypyc/mypyc/issues/644

## Test Plan

Includes unit test

### DIFF
--- a/mypyc/lib-rt/str_ops.c
+++ b/mypyc/lib-rt/str_ops.c
@@ -53,6 +53,18 @@ PyObject *CPyStr_Split(PyObject *str, PyObject *sep, CPyTagged max_split)
     return PyUnicode_Split(str, sep, temp_max_split);
 }
 
+bool CPyStr_Startswith(PyObject *self, PyObject *subobj) {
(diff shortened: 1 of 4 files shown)


TH3CHARLie (approved):
LGTM and thanks for adding more primitives to mypyc!

@JukkaL Do you think is it worth to implement these two ops directly in the irbuild?(I've checked and it's possible to do so)

JukkaL:
@TH3CHARLie Except for the most performance critical primitives it's better to use a wrapper function, since they are easier to write and maintain. If these turn out to be bottlenecks in some (realistic) benchmarks, we can consider implementing these in irbuild.

JukkaL:
I checked that this improves the `str_methods` microbenchmark (in https://github.com/mypyc/mypyc-benchmarks) from 1.3x faster than CPython to 1.9x faster (on Linux)!

The results should automatically update overnight here so we can see how it performs on the benchmarks server: https://github.com/mypyc/mypyc-benchmark-results/blob/master/reports/summary-microbenchmarks.md

PR #9558, opened 2020-10-08
[mypyc] support unicode identifiers

r3m0t:
### Description

Fixes mypyc/mypyc#699

### DIFF
--- a/mypyc/namegen.py
+++ b/mypyc/namegen.py
@@ -5,9 +5,13 @@ from typing import List, Dict, Tuple, Set, Optional, Iterable, Match
 def make_c_compatible(name: str) -> str:
     """Replace characters that can't be used in C identifiers."""
     def as_compatible_char(match: Match[str]) -> str:
+        if match.group(0).startswith('_'):
+            return match.group(0) + '_'
+        if match.group(0) == '.':
+            return '___'
         codepoint = ord(match.group(0))
         return '_{}'.format(hex(codepoint))
-    name = re.sub(r'[^._a-zA-Z0-9]', as_compatible_char, name, flags=re.ASCII)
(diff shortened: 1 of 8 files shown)


mypyc/namegen.py
```diff
@@ -30,6 +40,9 @@ class NameGenerator:
       collides '___' with '.3_', but this is OK because names
       may not start with a digit.)
 
+    * Replace non-ASCII characters with their codepoints, such as '_e9'
+      for U+00E9.
```
msullivan, line 44: The downside here is that we'll get name collisions if `_e9` appears in a name. Can we come up with an encoding scheme that dodges this without uglifying normal names too much, like we did for `.` -> `___`. Would just using `___3[the unicode code point]_` work?

mypyc/namegen.py
```diff
@@ -36,9 +40,7 @@ class NameGenerator:
       prefixes will be 'bar_foo_' and 'baz_foo_'.
 
     * Replace '.' in the Python name with '___' in the C name. (And
-      replace the unlikely but possible '___' with '___3_'. This
-      collides '___' with '.3_', but this is OK because names
-      may not start with a digit.)
+      replace '___' with '____' and so on to prevent collisions)
```
msullivan, line 43: This doesn't seem to help with the unicode collisions, though?

msullivan: Oh, and it introduces new collisions. `a._` and `a___` (3 underscores) now both map to `a____` (4 underscores), for example.

r3m0t: Oops!

msullivan (changes requested):
I think restoring the old `___` -> `___3_` encoding and then mapping unicode to `___3[unicode hex]_` would work. There would be collisions with `.3[stuff]_` but that's OK because names can't start with digits.

PR #9562, opened 2020-10-08
py39: fix mypyc complaints part 2

hauntsaninja:
Necessary because I previously didn't actually fix mypyc's complaint +
mypyc has more complaints.
The sys.version_info aliasing works around us hitting
https://github.com/python/mypy/blob/08f207ef4a09f56d710d63775771ae921c41d4bc/mypyc/irbuild/expression.py#L44


msullivan (approved):
The sys.version_info one is really annoying.

Issue #9442, opened 2020-09-15
Support disallow_untyped_decorators = false option for 3rd party libraries


mvoitko:
**Feature**

Support `disallow_untyped_decorators` option for 3rd party libraries
Example:

```
[mypy-pytest-asyncio]
disallow_untyped_decorators = false
```

**Pitch**

Example
pytest decorators are widely used in the testing of Python code.  pytest is still untyped for now.
It would be nice to have the possibility to check the typing of own decorators but ignore untyped decorators of some 3rd party libraries.

gvanrossum:
You can already specify this flag on a per-module or per-package basis. How is your proposal different?

mvoitko:
There is an example described [here](https://mypy.readthedocs.io/en/stable/config_file.html#inverting-option-values), where it suppresses any error messages generated when the codebase tries importing the module.
I would like to use the same approach to suppress error messages, not about all decorators in my module or my package, but to suppress the error messages for all decorators from 3rd party library, but still detect my own untyped decorators.

gvanrossum:
Oh, so you want something like this:
```py
from some_third_party_package import decorator_one  # Untyped
from my_own_library import decorator_two  # Also untyped

# No error here
@decorator_one
def foo() -> None: ...

# But report an error here
@decorator_two
def bar() -> None: ...
```
That's not going to be an easy feature to add; I don't even know where to begin.

PR #9564, opened 2020-10-09
implement PEP 585 Generic Builtins (Partially solves #7907)

### LINKED ISSUE #7907: Implement PEP 585 (Generic builtins and __future__.annotations)
* Are you reporting a bug, or opening a feature request? Reporting a bug
  or a mock-up repro if the source is private. We would appreciate if you try to simplify your case to a minimal repro.
```py
from collections import defaultdict

squares: defaultdict[tuple, int] = defaultdict(lambda: 999999)
```
* What is the actual behaviour/output? MyPy tells me `"defaultdict" is not subscriptable, use "typing.DefaultDict" instead`
* What is the behaviour/output you expect? MyPy ignores the subscript error (and recognises it as equivalent) because of `from __future__ import annotations`
* What is the version of mypy and Python you are using? 0.740 on 3.8.0 (Ubuntu 18.04)
* No additional MyPy flags

AllanDaemon:
### Description

Partially Solves #7907 

Instead of using the nongen_builtins dictionary with the non generic builtins names, it uses a function to get it that takes the python version. It returns an empty dict if the version is 3.9 or greater.

I'm opening the PR to review the code and the way it solves. Some discussion was made in the #7907  thread. I'll add the new tests to the code.

## Test Plan

I'm writing some tests for this, taking in consideration the different versions. But I'm having some difficulties to get the testing stuff right, so it may take a little bit. If wanted, you can merge this and I'll open a PR with the new tests latter.


mypy/nodes.py
```diff
@@ -136,9 +136,13 @@ def get_column(self) -> int:
     'builtins.frozenset': 'typing.FrozenSet',
 }  # type: Final
 
-nongen_builtins = {'builtins.tuple': 'typing.Tuple',
+_nongen_builtins = {'builtins.tuple': 'typing.Tuple',
                    'builtins.enumerate': ''}  # type: Final
-nongen_builtins.update((name, alias) for alias, name in type_aliases.items())
+_nongen_builtins.update((name, alias) for alias, name in type_aliases.items())
+
+def get_nongen_builtins(python_version):
```
frost-nzcr4, line 143:
```suggestion
def get_nongen_builtins(python_version):  # type: (Tuple[int, int]) -> dict
```

mypy/nodes.py
```diff
@@ -136,9 +136,13 @@ def get_column(self) -> int:
     'builtins.frozenset': 'typing.FrozenSet',
 }  # type: Final
 
-nongen_builtins = {'builtins.tuple': 'typing.Tuple',
+_nongen_builtins = {'builtins.tuple': 'typing.Tuple',
                    'builtins.enumerate': ''}  # type: Final
-nongen_builtins.update((name, alias) for alias, name in type_aliases.items())
+_nongen_builtins.update((name, alias) for alias, name in type_aliases.items())
+
```
frost-nzcr4, line 142: PEP8 2 blank lines before function name

https://travis-ci.com/github/python/mypy/jobs/397269051

frost-nzcr4 (changes requested):

mypy/nodes.py
JelleZijlstra: Actually, this should use inline type annotations (not type comments), and a more precise type than `dict` (I haven't looked what it should be).

antonagestam: Perhaps it's a good idea to either type the function as returning `Mapping[str, str]`, or make it return either copies of the original dict or something actually immutable like `types.MappingProxyType`.

mypy/nodes.py
```diff
@@ -136,9 +136,13 @@ def get_column(self) -> int:
     'builtins.frozenset': 'typing.FrozenSet',
 }  # type: Final
 
-nongen_builtins = {'builtins.tuple': 'typing.Tuple',
+_nongen_builtins = {'builtins.tuple': 'typing.Tuple',
                    'builtins.enumerate': ''}  # type: Final
```
antonagestam, line 140:
```suggestion
_nongen_builtins = {'builtins.tuple': 'typing.Tuple',
                    'builtins.enumerate': ''}  # type: Final
```

(The change introduces misaligned indentation)

cdce8p:
@AllanDaemon I've written some test cases. Take a look at https://github.com/AllanDaemon/mypy/pull/1

mypy/typeanal.py
```diff
@@ -305,7 +308,8 @@ def try_analyze_special_unbound_type(self, t: UnboundType, fullname: str) -> Opt
         elif fullname == 'typing.Callable':
             return self.analyze_callable_type(t)
         elif (fullname == 'typing.Type' or
-             (fullname == 'builtins.type' and self.api.is_future_flag_set('annotations'))):
+             (fullname == 'builtins.type' and self.api.is_future_flag_set('annotations')) or
```
hauntsaninja, line 311: nit: this would read cleaner as
```
fullname == 'builtins.type' and (self.options.python_version >= (3, 9) or self.api.is_future_flag_set('annotations'))
```

hauntsaninja (commented):
Two quick comments:
1) Could you also gate the GENERIC_STUB_NOT_AT_RUNTIME_TYPES in get_omitted_any behind a python version  check?
2) Could you add some tests for tuple? It's a bit of a special snowflake and I'm concerned it won't work

test-data/unit/check-generic-alias.test
```diff
@@ -0,0 +1,49 @@
...
+t4: Iterator
+t5: Mapping
+
+[builtins fixtures/tuple.pyi]
+
+[case testGenericAliasImportBuiltIns]
+# flags: --python-version 3.9
+
+t1: type[int]
+
```
cdce8p, line 31: This should be enough to add `tuple` to the tests. And for good measure I've added `dict` as well. However `set` needs to be in its own test, as it is now. It seems that the fixtures don't work together for whatever reason.
```suggestion
t2: tuple[int, int]
t3: dict[int, str]

```

AllanDaemon:
> @AllanDaemon I've written some test cases. Take a look at [AllanDaemon#1](https://github.com/AllanDaemon/mypy/pull/1)

Thanks. I wrote a few tests but I got stuck with the issue that the `type` as it's on a different part of the code that I hadn't found yet, but your PR solves it. Now I gonna add some of my tests but the PR seems good to go.

AllanDaemon:
> Two quick comments:
> 
> 1. Could you also gate the GENERIC_STUB_NOT_AT_RUNTIME_TYPES in get_omitted_any behind a python version  check?
IDK. I'm not familiar yet with the code, but as far as related with PEP 585, those 2 classes are not mentioned there. I also couldn't find a code that triggers this. It this should be gated by python version, I think it would be better to open an new issue / PR.

> 2. Could you add some tests for tuple? It's a bit of a special snowflake and I'm concerned it won't work
I will.

AllanDaemon:
Well, I'm having some issues on writing tests, with some of them being bugs or things to do in code, not in the tests. The last commit I added introduces them.

> 2. Could you add some tests for tuple? It's a bit of a special snowflake and I'm concerned it won't work

And It doesn't. I added some tests that show it's not working. When I change `Tuple` with `tuple` it generates some pretty strange errors.


Also, there is the reveal_type() for the `collections.abc` types that are just returning `Any`. Running it outside the tests it usually says something like `typings.TYPE_NAME` instead, that is wrong now and I'll address it in the #7907  discussion.  Even the `collections.abc.Callable` that usually have a custom representation is returning Any here.

I'll investigate these errors, but as I'm a beginner with mypy internals I can take a long time to figure it out, so any help is appreciated.

AllanDaemon:
The  `tuple` problems seems when using variable-length tuple of homogeneous type. It works with `typings.Tuple` but not with `tuple`. 

```python
from typing import Tuple
T: Tuple[int, ...] = (1, 2)  # OK
t: tuple[int, ...] = (1, 2)  # error: Unexpected '...'
```

AllanDaemon:
By now, just the test that handles the `reveal_type()` is failing. As I wrote here (https://github.com/python/mypy/issues/7907#issuecomment-716094889), I think this should be handled in another PR, in particular after discussing a few things about how `reaveal_type()` should be handled in some cases (like `type`, and `tuple`).

So, for this PR, after I disable the tests that fail because of `reveal_type()`, this PR should be good to go.

So, is there anything else that should be handled before it could be merged?

mypy/typeanal.py
```diff
@@ -272,7 +275,7 @@ def try_analyze_special_unbound_type(self, t: UnboundType, fullname: str) -> Opt
             self.fail("Final can be only used as an outermost qualifier"
                       " in a variable annotation", t)
             return AnyType(TypeOfAny.from_error)
-        elif fullname == 'typing.Tuple':
+        elif fullname in ('typing.Tuple', 'builtins.tuple'):
```
cdce8p, line 278: As with `builtins.type` `builtins.tuple` is only allowed if you're using py39+ or `from __future__ import annotations` is used.
```suggestion
        elif (fullname == 'typing.Tuple' or
             (fullname == 'builtins.tuple' and (
                 self.options.python_version >= (3, 9) or self.api.is_future_flag_set('annotations')))):
```

AllanDaemon: Of course. Thanks.

AllanDaemon:
So, what else is missing for this PR that I need to do?

test-data/unit/check-generic-alias.test
```diff
@@ -0,0 +1,257 @@
...
+t17: tuple[int, ...] = (1, 2, 3, 4, 5, 6)
+t18: tuple[int, ...] = (1, 2, 3, 4, 5, 6, 7)
+t19: tuple[int, ...] = (1, 2, 3, 4, 5, 6, 7, 8)
+
+from typing import Tuple
+
+t20: Tuple[int, ...] = t19
+
+[builtins fixtures/tuple.pyi]
+[out]
```
ethanhs, line 257: Nit: missing newline at the end of file
```suggestion
[out]

```

ethanhs (approved):
This looks good, thank you for implementing PEP 585 support! I just had one minor change but otherwise I think this is good to land.

cdce8p:
Any updates?

hauntsaninja:
Thanks for your work on this!

I'm also closing #7907. We can open a more specific issue for the low priority additions of deprecation warnings / making reveal_type reveal collections classes.

wbadart:
Hey team- just upgraded to `0.800` and found that mypy might not recognize PEP585-type syntax in type aliases:

```python
T = list[int]
```

```
foo.py:1: error: "list" is not subscriptable  [misc]
    T = list[int]
        ^
```

Let me know if you have trouble reproducing the error and/or if there's anything I can do to help!

JelleZijlstra:
@wbadart that syntax doesn't work at runtime in 3.9 yet. See #9442 and #9880.

wbadart:
Yup. Sorry, I should have noted that I'm using [`from __future__ import annotations`][585] in the actual code where I noticed this, so no runtime errors. In any case, I just meant to note how mypy treats this syntax (regardless of what the Python interpreter makes of it).

[585]: https://www.python.org/dev/peps/pep-0585/#implementation

JelleZijlstra:
The future import doesn't make a difference to type aliases, because that code is still evaluated at runtime.

wbadart:
Ahh... Okay, I'm following now. That would explain why we don't want that code to type-check. Thanks for the clarification!

rben01:
> Hey team- just upgraded to `0.800` and found that mypy might not recognize PEP585-type syntax in type aliases:
> 
> ```python
> T = list[int]
> ```
> 
> ```
> foo.py:1: error: "list" is not subscriptable  [misc]
>     T = list[int]
>         ^
> ```
> 
> Let me know if you have trouble reproducing the error and/or if there's anything I can do to help!

Also seeing this on `0.812`

cdce8p:
@rben01 Please read the comments above. The PEP 585 syntax in Type Aliases requires at least Python **3.9**. Regardless if `from __future__ import annotations` is used or not. The mypy documentation has a good page on exactly that problem, since it can be quite confusing sometimes: https://mypy.readthedocs.io/en/stable/runtime_troubles.html

Once mypy supports [PEP 613](https://www.python.org/dev/peps/pep-0613/) you should however be able to declare them as strings in versions prior to **3.9**.
```py
T: TypeAlias = "list[int]"
```

rben01:
@cdce8p Sorry for skipping over the previous comments. But after (re-)reading all the prior material I'm still confused. Pep 585 says this:

> Starting with Python 3.7, when from __future__ import annotations is used, function and variable annotations can parameterize standard collections directly. Example:
>```python
>from __future__ import annotations
>def find(haystack: dict[str, list[int]]) -> int:
>    ...
>```
> Usefulness of this syntax before PEP 585 is limited as external tooling like Mypy does not recognize standard collections as generic. Moreover, certain features of typing like type aliases or casting require putting types outside of annotations, in runtime context. While these are relatively less common than type annotations, it's important to allow using the same type syntax in all contexts. This is why starting with Python 3.9, the following collections become generic using __class_getitem__() to parameterize contained types:

I think that that Pep is saying that using `list[int]` as a type annotation is allowed in Python 3.7+ as long as you import `annotations` from `__future__`, and that it's only disallowed in Python 3.7 and 3.8 when it used as an r-value, such as when it's the first argument to `cast` or in an expression such as `type_alias = list[int]`.

I tested this in an interpreter and it seems like this is correct:

Text available below

<details>
<summary> Text version </summary>

    In [66]: sys.version
    Out[66]: '3.8.7 (default, Feb 10 2021, 03:56:21) \n[Clang 7.1.0 (tags/RELEASE_710/final)]'
    
    In [67]: def find(haystack: dict[str, list[int]]) -> int:
        ...:     pass
        ...:     
    
    In [68]: type_alias = list[int]
    ---------------------------------------------------------------------------
    TypeError                                 Traceback (most recent call last)
    <ipython-input-68-1bf51d638f86> in <module>
    ----> 1 type_alias = list[int]
    
    TypeError: 'type' object is not subscriptable
</details>

Python's type system is evolving way too quickly for me to keep up, and the subtle behavior differences between Python versions are a lot to keep track of so I'll defer to the Mypy team. But this was my understanding of Pep 585: that `list[int]` is valid as a type annotation, but prior to Python 3.9 it may not appear as an r-value/be evaluated at runtime.

JelleZijlstra:
@rben01 That's right; I think the part that you're misunderstanding is that `type_alias = list[int]` isn't a type annotation, it's a runtime use of the type. Therefore, `from __future__ import annotations` doesn't help: when you run the code Python still tries to actually execute `list[int]` and fails.

rben01:
@JelleZijlstra Ok, sorry for the confusion. I am having the following problem, which I now realize is not the same as wdabart's, which is that mypy flags this as invalid for me:

```python
x: list[int] = [1]
```

I think this is definitely a bug with mypy? I can open a new issue if there isn't one for this specific case.

JelleZijlstra:
That should be allowed, yes (assuming mypy is running at least `--python-version 3.7`). If it doesn't that's a bug.

rben01:
Ah, sorry, the underlying issue was with having two versions of mypy installed and Visual Studio Code not using the right one when linting. Ignore me 🤦 .

PR #9570, opened 2020-10-09
misc: update wheel building and release scripts

hauntsaninja:
This goes together with https://github.com/mypyc/mypy_mypyc-wheels/pull/11 to resolve https://github.com/python/mypy/issues/9536

download-mypyc-wheels is now subsumed by upload-pypi, which uses the Github API instead of hardcoding filenames. This also means you won't need to make changes like https://github.com/python/mypy/pull/9559 If this feels fragile, let me know and I can add back a check against a hardcoded list of names (note CI won't make a release if one of the builds fails).

test_installed_version is part of mypyc_mypy-wheels and some of what it does is taken care of by cibuildwheel

upload-pypi can be made a lot simpler, since CI now builds the pure Python wheel and sdist. Since it's no longer building things, I assume we can be less picky about what Python we use. The version check now runs against the sdist, which is nice.

Some of this was a little hard to test end to end, but if there are kinks it shouldn't be too hard to work them out.


msullivan (approved):
This seems pretty reasonable to me and we should move forward with it. We should coordinate merging it so I can bother you about any "fine-tuning" that needs to be done.

Issue #9456, opened 2020-09-18
classmethod produces errors due to TypeVar


Dreamsorcerer:
A minimal test case:
```
[case testClassMethod]
from typing import Generic, TypeVar

T = TypeVar("T", int, None)

class A(Generic[T]):
    @classmethod
    def func(cls, arg: int):
        return cls(arg)

    def __init__(self, arg: T):
        self.attr = arg

[builtins fixtures/classmethod.pyi]
```

This gives an error complaining that `cls(arg)` expects `None`, due to it expanding the TypeVar on the class object.

Expected behaviour is that no error is produced, and that the type of `cls` does not have the expanded `TypeVar` (i.e. it should receive the same type that would be seen on a separate line of code like: `a = A(5)`.

One catch to changing this behaviour is correct handling when subclassing with a defined type: https://github.com/python/mypy/issues/9456#issuecomment-968201238

I've had a go at this, but was unable to figure out how to fix it. This is what I've figured out though:
https://github.com/python/mypy/blob/master/mypy/semanal.py#L641
Within `prepare_method_signature()`, `cls` gets changed from `Any` to the class type.
This is done by calling `fill_typevars()`, which results in the type including `T`, rather than just remaining as a `TypeInfo` (which appears to be the type in an example such as `a = A(5)`).
I've tried just removing the `fill_typevars()`, but then it runs into an unimplemented `RuntimeError`.

Because it includes `T` at this point, when it later tries to typecheck the code, it will check the class method twice, once as `A[int]` and again as `A[None]`, which obviously throws the error we see.

So, it seems like there needs to be some way to stop class methods getting the `TypeVar`s filled out on the `cls` argument.

Dreamsorcerer:
@gvanrossum I'm not sure I follow the reasoning on the `cls(day, day)`. My assumption is that this would be the equivalent of running `TimePeriod(date1, date2)`, where I am expecting the TypeVar to be defined by the date objects I'm passing in. i.e. The return type would be inferred as `TimePeriod[date, date]`.
The fact that the error states that it expected `None` makes me think that it's the same error as the operand one. i.e. Mypy thinks these TypeVars should always be `None`.

gvanrossum:
[...]

PR #9572, opened 2020-10-10
disable unreachable warnings in boolean ops on TypeVars with value restriction

vbarbaresi:
#### Description

This paragraph explains the limitation with TypeVars:
https://github.com/python/mypy/blob/eb50379defc13cea9a8cbbdc0254a578ef6c415e/mypy/checker.py#L967-#L974

We currently have no way of checking for all the type expansions,
and it's causing one part of the issue https://github.com/python/mypy/issues/9456

Using `self.chk.should_report_unreachable_issues()` honors the suppression of
the unreachable warning for TypeVar with value restrictions

#### Test Plan

I added a test statement that reproduces the unreachable false positive.
I added another test statement that is currently a false negative for the moment, because of the suppression for TypeVar type expansions.


msullivan:
Thanks!

PR #9573, opened 2020-10-10
Improve ambiguous **kwarg checking

### LINKED ISSUE #4708: Multiple dict unpacking for function arguments gives type error
When I run the following code through MyPy:
```
def fn(a: int, b: int, c: int) -> None:
    print((a, b, c))

fn(**{'a': 10}, **{'b': 20, 'c': 30})
```
I receive the following:
```
$ python -m mypy mypy_problem.py
mypy_problem.py:8: error: Too many arguments for "fn"
```

I expect this to not give an error because as far as I know it's valid python code - it works at least! A workaround is to wrap the whole thing in another dictionary, i.e.:
```
fn(**{**{'a': 10}, **{'b': 20, 'c': 30}})
```
But this is messy and possibly introduces an additional overhead.
```
$ conda list | grep 'python\|mypy\|typed-ast'
mypy                      0.570                     <pip>
python                    3.5.2                         0    conda-forge
typed-ast                 1.1.0                     <pip>
```
I'm using Windows 7 64-bit.

### LINKED ISSUE #9395: Position of double-starred kwargs affects the type errors when instantiating subclasses attr objects
## 🐛 Bug Report

I'm not sure this is the issue for typeshed or mypy, but with `attrs`, unpacking a dict to initialize a subclassed attr instance generates a weird type error or not depending on the position of the double-starred kwargs, where the runtime behavior is completely fine for both cases.

## To Reproduce

Run `mypy` with the following script:
```python
import attr
from typing import List

@attr.s(auto_attribs=True, slots=True)
class Point:
    x: int
    y: int

@attr.s(auto_attribs=True, slots=True)
class Point3D(Point):
    z: int
    opts: List[str]

p = Point(x=1, y=2)
q = Point3D(
    **attr.asdict(p),
    z=3,
    opts=['a', 'b'],
)
```

## Expected Behavior

There should be no type errors. (Or at least, some different error saying that it cannot validate keys from an arbitrary dict..)

## Actual Behavior

When `**` comes first:
```python
p = Point(x=1, y=2)
q = Point3D(  # <-- error: "Point3D" gets multiple values for keyword argument "opts"
    **attr.asdict(p),
    z=3,
    opts=['a', 'b'],
)
```
When `**` comes last:
```python
p = Point(x=1, y=2)
q = Point3D(  # no errors!
    z=3,
    opts=['a', 'b'],
    **attr.asdict(p),
)
```
where the runtime results are same.

Q: Are there other ways to initialize a subclassed attr instance from a super-attr instance with its values, instead of using `**attr.asdict()`?

## Your Environment

- Mypy version used: 0.783
- Mypy command-line flags: `python -m mypy test-attr.py`
```
[mypy]
ignore_missing_imports = true
namespace_packages = true
```
- Python version used: 3.8.5
- Operating system and version: Ubuntu 18.04

esoma:
# Description

Fixes #4708
Allows for multiple ambiguous **kwarg unpacking in a call -- all ambiguous **kwargs will map to all formal args that do not have a certain actual arg.

Fixes #9395 
Defers ambiguous **kwarg mapping until all other unambiguous formal args have been mapped -- order of **kwarg unpacking no longer affects the arg map.

## Test Plan

Additional tests included in PR, no existing tests were changed. Below are new tests that would exhibit different behaviors if they're run without code changes in the PR:

---
```python
[case testPassingMultipleKeywordVarArgs]
from typing import Any, Dict
def f1(a: 'A', b: 'A') -> None: pass
def f2(a: 'A') -> None: pass
def f3(a: 'A', **kwargs: 'A') -> None: pass
def f4(**kwargs: 'A') -> None: pass
d = None # type: Dict[Any, Any]
d2 = None # type: Dict[Any, Any]
f1(**d, **d2)
f2(**d, **d2)
f3(**d, **d2)
f4(**d, **d2)
class A: pass
[builtins fixtures/dict.pyi]
```

`f1(**d, **d2)` produces no error in PR. Current produces `Too many arguments for "f1"` (#4708)
`f2(**d, **d2)` produces no error in PR. Current produces `Too many arguments for "f2"` (#4708)

---

```python
[case testTypedDictAsStarStarAndDictAsStarStar]
from typing import Any, Dict

TD = TypedDict('TD', {'x': int, 'y': str})

def f1(x: int, y: str, z: bytes) -> None: ...
def f2(x: int, y: str) -> None: ...

td: TD
d = None # type: Dict[Any, Any]

f1(**td, **d)
f1(**d, **td)
f2(**td, **d) # E: Too many arguments for "f2"
f2(**d, **td) # E: Too many arguments for "f2"
[builtins fixtures/dict.pyi]
```
`f1(**d, **td)` produces no error in PR. Current produces `"f1" gets multiple values for keyword argument "x"` and `error: "f1" gets multiple values for keyword argument "y"` (#9395) 
`f2(**d, **td)` produces `Too many arguments for "f2"`. Current produces `"f2" gets multiple values for keyword argument "x"` and `"f2" gets multiple values for keyword argument "y"` (#9395)

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -4115,18 +4115,22 @@ def is_non_empty_tuple(t: Type) -> bool:
 def is_duplicate_mapping(mapping: List[int],
                          actual_types: List[Type],
                          actual_kinds: List[int]) -> bool:
-    # Multiple actuals can map to the same formal only if they both come from
-    # varargs (*args and **kwargs); in this case at runtime it is possible that
-    # there are no duplicates. We need to allow this, as the convention
-    # f(..., *args, **kwargs) is common enough.
-    return len(mapping) > 1 and not (
-        len(mapping) == 2 and
-        actual_kinds[mapping[0]] == nodes.ARG_STAR and
-        actual_kinds[mapping[1]] == nodes.ARG_STAR2) and not all(
-            actual_kinds[m] == nodes.ARG_STAR2 and
-            not isinstance(actual_types[m], TypedDictType)
-            for m in mapping
-        )
+    return (
+        len(mapping) > 1
+        # Multiple actuals can map to the same formal if they both come from
+        # varargs (*args and **kwargs); in this case at runtime it is possible
+        # that here are no duplicates. We need to allow this, as the convention
+        # f(..., *args, **kwargs) is common enough.
+        and not (len(mapping) == 2
+                 and actual_kinds[mapping[0]] == nodes.ARG_STAR
+                 and actual_kinds[mapping[1]] == nodes.ARG_STAR2)
+        # Multiple actuals can map to the same formal if there are multiple
+        # **kwargs which cannot be mapped with certainty (non-TypedDict
+        # **kwargs).
+        and not all(actual_kinds[m] == nodes.ARG_STAR2 and
+                    not isinstance(actual_types[m], TypedDictType)
+                    for m in mapping)
+    )
 
 
 def replace_callable_return_type(c: CallableType, new_ret_type: Type) -> CallableType:
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -1336,7 +1336,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                 ok = False
             elif kind in [nodes.ARG_POS, nodes.ARG_OPT,
                           nodes.ARG_NAMED, nodes.ARG_NAMED_OPT] and is_duplicate_mapping(
-                    formal_to_actual[i], actual_kinds):
+                    formal_to_actual[i], actual_types, actual_kinds):
                 if (self.chk.in_checked_function() or
                         isinstance(get_proper_type(actual_types[formal_to_actual[i][0]]),
                                    TupleType)):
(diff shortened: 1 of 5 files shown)


mypy/checkexpr.py
```diff
@@ -4112,15 +4112,25 @@ def is_non_empty_tuple(t: Type) -> bool:
...
+        # that here are no duplicates. We need to allow this, as the convention
+        # f(..., *args, **kwargs) is common enough.
+        and not (len(mapping) == 2
+                 and actual_kinds[mapping[0]] == nodes.ARG_STAR
+                 and actual_kinds[mapping[1]] == nodes.ARG_STAR2)
+        # Multiple actuals can map to the same formal if there are multiple
+        # **kwargs which cannot be mapped with certainty (non-TypedDict
+        # **kwargs).
+        and not all(actual_kinds[m] == nodes.ARG_STAR2 and
+                    not isinstance(actual_types[m], TypedDictType)
```
esoma, line 4131: `mypy/checkexpr.py:4131: error: Never apply isinstance() to unexpanded types;`

I don't see a terribly clean way to fix this. `# type: ignore` okay here? Or other suggestions?

LiuYuHui:
```suggestion
                    not isinstance(get_proper_type(actual_types[m]), TypedDictType)
```

esoma: Ah, thanks. I totally misunderstood the error.

msullivan (approved):
Looks great, thanks!

PR #9575, opened 2020-10-10
[mypyc] Make multiple assignment faster(1st stage)

TH3CHARLie:
relates mypyc/mypyc#729

This PR makes simple cases like `y, x = x, y` faster.

I tend to do some future improvement based on this PR(for example, supporting `ListExpr` as @JukkaL suggested), nevertheless, this PR itself should cover the most frequent use case.


TH3CHARLie:
Some imprecise benchmark results:

Before:
```
............
interpreted: 0.167262s (avg of 6 iterations; stdev 0.61%)
compiled:    0.163983s (avg of 6 iterations; stdev 0.48%)

compiled is 1.020x faster
```
After:
```
running multiple_assignment
............
interpreted: 0.164994s (avg of 6 iterations; stdev 0.52%)
compiled:    0.055589s (avg of 6 iterations; stdev 0.69%)

compiled is 2.968x faster
```

TH3CHARLie:
I believe the failed `StubgenCliParseSuite.test_walk_packages` is not related to this PR

JukkaL (approved):
Looks great! This fixes a bad performance bottleneck.

PR #9577, opened 2020-10-11
Require first argument of namedtuple to match with variable name

### LINKED ISSUE #4589: Check first argument of NamedTuple/namedtuple
Mypy doesn't complain about this code, even though the name arguments to `NamedTuple` and `namedtuple` are inconsistent with the assignment targets:

```py
from typing import NamedTuple
from collections import namedtuple

A = NamedTuple('X', [('a', int)])  # First argument should be 'A'?
B = namedtuple('X', ['a'])  # First argument should be 'B'?
```

I'd expect mypy to complain about the `'X'` arguments in both cases.

momohatt:
### Description

Closes #4589

This PR modifies `check_namedtuple` to return the internal name of the namedtuples (e.g. the content of the first argument of namedtuple/NamedTuple) so that the callers, especially `analyze_namedtuple_assign`, can check if the name of the variable on the l.h.s. matches with the first argument of the namedtuple.

### Test Plan

Added a code snippet from #4589 as the new test.
Also, I deleted `testReexportNamedTupleChange` (which was added in #6698 for https://github.com/python/mypy/issues/6548#issuecomment-484277202) since we wouldn't need to worry about this case when this is merged.

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -359,7 +359,7 @@ class SemanticAnalyzerPluginInterface:
 # A context for querying for configuration data about a module for
 # cache invalidation purposes.
 ReportConfigContext = NamedTuple(
-    'DynamicClassDefContext', [
+    'ReportConfigContext', [
         ('id', str),        # Module name
         ('path', str),      # Module file path
         ('is_check', bool)  # Is this invocation for checking whether the config matches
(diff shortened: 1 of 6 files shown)


msullivan (approved):
This looks great. Thanks.

PR #9583, opened 2020-10-12
Made the change "ever" to "even".Because even is more certain than ever

Abhishek-IOT:
### Description

I have just changed a word in the documentation because I thought that is more appropriate.

## Test Plan

This is a documentation change .


JelleZijlstra:
That doesn't seem necessary.

PR #9587, opened 2020-10-13
add mypyc.analysis to setup.py

JelleZijlstra:
Fixes #9584 

I haven't tested myself but this package should probably be here.

PR #9593, opened 2020-10-13
setup.py: use find_packages

hauntsaninja:
Is there a reason to not do this? It gets the right set of packages when
I run it. Saves chores like #9587 or #9061


JelleZijlstra (approved):
Also not sure if there was a reason for avoiding `find_packages()`.

Issue #9001, opened 2020-06-14
Segfault when working with TypeQuery


sobolevn:
* Are you reporting a bug, or opening a feature request?

There's a bug when working with `TypeQuery`. It is literally unusable. 

* Please insert below the code you are checking with mypy,
  or a mock-up repro if the source is private. We would appreciate
  if you try to simplify your case to a minimal repro.

Just take any type inside a plugin and pass it to `TypeQuery`:

```python
from mypy.plugin import FunctionContext
from mypy.types import Type as MypyType
from mypy.types import TypeQuery
from mypy.plugin import Plugin

def analyze(ctx: FunctionContext) -> MypyType:
    print(ctx.default_return_type.accept(TypeQuery(any)))
    return ctx.default_return_type

class _ReturnsPlugin(Plugin):
    def get_function_hook(self, fullname):
        if fullname == 'ex.test':
            return analyze

def plugin(version: str):
    return _ReturnsPlugin
```

And provide any source code to trigger this plugin:

```python
# ex.py
def test() -> int:
     return 1
```

* What is the actual behavior/output?

```
» PYTHONFAULTHANDLER=1 mypy ex.py
Fatal Python error: Segmentation fault

Current thread 0x0000000100b155c0 (most recent call first):
  File "/Users/sobolev/Documents/github/returns/returns/contrib/mypy/_features/decorators.py", line 18 in analyze
  File "/Users/sobolev/Documents/github/returns/.venv/lib/python3.7/site-packages/mypy/__main__.py", line 8 in console_entry
  File "/Users/sobolev/Documents/github/returns/.venv/bin/mypy", line 10 in <module>
[1]    22569 segmentation fault  PYTHONFAULTHANDLER=1 mypy ex.py

```

* What is the behavior/output you expect?

I would expect `TypeQuery` to work. Without the segfault.

* What are the versions of mypy and Python you are using?
  Do you see the same issue after installing mypy from Git master?

`0.780`

* What are the mypy flags you are using? (For example --strict-optional)

Full list: https://github.com/dry-python/returns/blob/master/setup.cfg#L122

ethanhs:
Mypy sets a higher recursion limit, which may be triggering this (if there is more recursion than the system stack depth) then mypy can segfault.

Can you try setting a higher stack limit?

sobolevn:
I have tried to add

```python
import sys
sys.setrecursionlimit(15000)
```

into the plugin code. Nothing changed.

ethanhs:
[...]

PR #9602, opened 2020-10-16
TypeVisitor and Co now support allow_interpreted_subclasses=True

sobolevn:
# Description

Now these types can be extended from plugin code.
Previously it was impossible:

```python
from mypy.type_visitor import TypeVisitor


class KindTranslator(TypeVisitor):
    ...
```

Was giving:

```
TypeError: interpreted classes cannot inherit from compiled traits
```

But, this is a very powerful tool. Some complex type manipulations do require this (or similar) tool.
For example:
- https://github.com/dry-python/returns/issues/391
- https://github.com/dry-python/returns/issues/631

More context: https://github.com/python/mypy/issues/9001#issuecomment-710315985

## Test Plan

I actually don't have any test plan at the moment. I would be very happy to receive any advice! 👍

### DIFF
--- a/mypy/type_visitor.py
+++ b/mypy/type_visitor.py
@@ -14,7 +14,7 @@ other modules refer to them.
 from abc import abstractmethod
 from mypy.ordered_dict import OrderedDict
 from typing import Generic, TypeVar, cast, Any, List, Callable, Iterable, Optional, Set, Sequence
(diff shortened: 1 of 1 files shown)


msullivan (approved):
Yeah, this is a good idea. It'll increase our generated code size but shouldn't slow things down.

Issue #7128, opened 2019-07-02
Speed up TypeQuery


JukkaL:
`mypy.type_visitor.TypeQuery` is constructed often with the `any` or `all` builtin function as an argument. The function is used as a callable inside the query. Calling callable variables is kind of slow when compiled with mypyc. Since type queries are used a lot, switching to an integer constant or a compiled object instead of a callable could speed up runtimes measurably, perhaps by as much as 0.5% or so (random guess). Since the optimization is pretty simple, this may be worth doing.

Discussion:
* Maybe have separate type query visitor which produces boolean values. The strategy could be specified as an integer constant.
* For non-bool type queries, experiment with using an object instead of a callable for the strategy, since callables are currently pretty slow when compiled.
* Maybe also special case `self.strategy([])` since it's used a lot.

mridullpandey:
hi @JukkaL i want to work on can you assign it to me
thanks :-)

mridullpandey:
can you please let me know how to get started

JukkaL:
@mridullpandey I've assigned this to you. Here are some hints:

* Figure out how to run mypy in compiled mode so that you can see whether you've improved performance. A good benchmark is type checking mypy. You need to figure out a way to measure the time consistently with accuracy of about 0.2% to validate whether you've made progress (or that at least you haven't regressed anything). Victor Stinner has written about how to do stable benchmarking, in case you want to learn about this.
* Create a copy of `TypeQuery` that is specialized for `bool` instead of accepting an arbitrary type.
* Replace the `strategy` constructor argument with an integer constant (corresponding to either `any` or `all`).
* Implement a `strategy` method that uses the integer constant to implement whatever is needed.

vbarbaresi:
I just opened a PR with my attempt at `TypeQuery` specialized for `bool`
In your first description of the issue, you wrote:
> For non-bool type queries, experiment with using an object instead of a callable for the strategy, since callables are currently pretty slow when compiled.

You mean replacing `strategy` type here
https://github.com/python/mypy/blob/ec76ccafa757c6d8fbb2c83195170bf75e0ff324/mypy/type_visitor.py#L257
by a new type of object that can be compiled by mypyc? There would still have to be a callable method someplace in that object, right?

Hyperion101010:
Hi there I am Shivam, I know Python to an intermediate level,
I am interested to solve this issue, will you assign me?

PR #9604, opened 2020-10-17
Create a copy of TypeQuery specialized for bool

vbarbaresi:
### Description
I tried implementing the optimization described in #7128.

I created `class TypeQueryBool(SyntheticTypeVisitor[bool])`.It's a specialized `TypeQuery`,
imlementating. `all` or `any` strategies.

Some code and comments are duplicated, so I'm not entirely satisfied with the implementation.
I'm opening the PR for some review about the implementation and the testing method.

## Test Plan

I forked `mypy_mypyc-wheels` project and I built mypyc wheels with travis, to benchmark.

I used `v0.800+dev.952ca239e6126aff52067a37869bd394eb08cf21` as the reference, and compared it with my exact changes added on top: https://github.com/vbarbaresi/mypy_mypyc-wheels/releases/tag/v0.800%2Bdev.f7129b21 

I'm using macOS 10.15, python 3.8.6 and the following benchmark script (I know it's not ideal because it counts the startup time of the python process):

```python
import sys
import pyperf

runner = pyperf.Runner()


runner.bench_command(
    'mypy_master',
    [
        '/tmp/venv_mypy_master/bin/mypy',
        '--config-file=mypy/mypy_self_check.ini',
        '/tmp/bench/mypy/mypy',
        '--no-incremental'
    ]
)


runner.bench_command(
    'mypy_typequerybool',
    [
        '/tmp/venv_mypy_typequerybool/bin/mypy',
        '--config-file=mypy/mypy_self_check.ini',
        '/tmp/bench/mypy/mypy',
        '--no-incremental'
    ]
)
```
I tried switching the order, running my branch first, then mypy master.
I wasn't able to get a 0.2 % accuracy, so I'm not entirely confident in the results.
I ran this bench more than 10 times and observed my branch performing very slightly better in a lot of runs

Typical run results:
```
Minimum:         10.9 sec
Median +- MAD:   11.0 sec +- 0.0 sec
Mean +- std dev: 11.0 sec +- 0.1 sec
Maximum:         11.1 sec

  0th percentile: 10.9 sec (-1% of the mean) -- minimum
  5th percentile: 11.0 sec (-1% of the mean)
 25th percentile: 11.0 sec (-0% of the mean) -- Q1
 50th percentile: 11.0 sec (+0% of the mean) -- median
 75th percentile: 11.1 sec (+0% of the mean) -- Q3
 95th percentile: 11.1 sec (+1% of the mean)
100th percentile: 11.1 sec (+1% of the mean) -- maximum

Number of outlier (out of 10.9 sec..11.2 sec): 0

mypy_master: Mean +- std dev: 11.0 sec +- 0.1 sec


Minimum:         10.9 sec
Median +- MAD:   11.0 sec +- 0.0 sec
Mean +- std dev: 11.0 sec +- 0.1 sec
Maximum:         11.1 sec

  0th percentile: 10.9 sec (-1% of the mean) -- minimum
  5th percentile: 10.9 sec (-1% of the mean)
 25th percentile: 10.9 sec (-0% of the mean) -- Q1
 50th percentile: 11.0 sec (+0% of the mean) -- median
 75th percentile: 11.0 sec (+0% of the mean) -- Q3
 95th percentile: 11.1 sec (+1% of the mean)
100th percentile: 11.1 sec (+1% of the mean) -- maximum

Number of outlier (out of 10.8 sec..11.1 sec): 0

mypy_typequerybool: Mean +- std dev: 11.0 sec +- 0.1 sec
```

﻿On some other runs the mean was 10.9 sec for my branch, and still 11.0 on master (but always with +- 0.1 sec std dev)

### DIFF
--- a/mypy/type_visitor.py
+++ b/mypy/type_visitor.py
@@ -14,6 +14,7 @@ other modules refer to them.
 from abc import abstractmethod
 from mypy.ordered_dict import OrderedDict
 from typing import Generic, TypeVar, cast, Any, List, Callable, Iterable, Optional, Set, Sequence
+from typing_extensions import Final
 from mypy_extensions import trait
 
 T = TypeVar('T')
@@ -245,7 +246,8 @@ class TypeQuery(SyntheticTypeVisitor[T]):
     """Visitor for performing queries of types.
 
     strategy is used to combine results for a series of types,
-    common use cases involve a boolean query using `any` or `all`.
+    For cases involving a boolean query using `any` or `all`, the specialized
+    TypeQueryBool is used.
 
     Note: this visitor keeps an internal state (tracks type aliases to avoid
     recursion), so it should *never* be re-used for querying different types,
@@ -349,11 +351,28 @@ class TypeQuery(SyntheticTypeVisitor[T]):
         return self.strategy(res)
 
 
+class BoolStrategies:
+    ANY_STRATEGY = 0  # type: Final[int]
+    ALL_STRATEGY = 1  # type: Final[int]
+
+
+def bool_strategy(res: List[bool], strategy_constant: int) -> bool:
+    if strategy_constant == BoolStrategies.ALL_STRATEGY:
+        return all(res)
+    else:
+        return any(res)
+
+
+def bool_strategy_empty(strategy_constant: int) -> bool:
+    if strategy_constant == BoolStrategies.ALL_STRATEGY:
+        return True
+    return False
+
 class TypeQueryBool(SyntheticTypeVisitor[bool]):
-    """Visitor for performing queries of types.
+    """Specialized visitor for boolean strategies
 
     strategy is used to combine results for a series of types,
-    common use cases involve a boolean query using `any` or `all`.
+    Cases involve a boolean query using `any` or `all`.
 
     Note: this visitor keeps an internal state (tracks type aliases to avoid
     recursion), so it should *never* be re-used for querying different types,
@@ -362,10 +381,10 @@ class TypeQueryBool(SyntheticTypeVisitor[bool]):
     # TODO: check that we don't have existing violations of this rule.
     """
 
-    def __init__(self, strategy: int) -> None:
+    def __init__(self, strategy_constant: int) -> None:
         # 0: any()
         # 1: all()
-        self.strategy = strategy
+        self.strategy_constant = strategy_constant
         # Keep track of the type aliases already visited. This is needed to avoid
         # infinite recursion on types like A = Union[int, List[A]].
         self.seen_aliases = set()  # type: Set[TypeAliasType]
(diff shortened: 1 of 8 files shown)


mypy/type_visitor.py
```diff
@@ -347,3 +349,121 @@ def query_types(self, types: Iterable[Type]) -> T:
...
+    def __init__(self, strategy: int) -> None:
+        # 0: any()
+        # 1: all()
+        self.strategy = strategy
+        # Keep track of the type aliases already visited. This is needed to avoid
+        # infinite recursion on types like A = Union[int, List[A]].
+        self.seen_aliases = set()  # type: Set[TypeAliasType]
+
+    def bool_strategy_empty(self) -> bool:
+        if self.strategy == self.STRATEGY_ALL:
```
msullivan, line 378: Try accessing the constants as `TypeQueryBool.STRATEGY_ALL` instead. mypyc should directly emit a constant in that case, and it'll emit an object lookup for `self`. (That's probably a bug.) The object lookup should be cheap, but is still a memory access and an error check, so the constant might be better.

mypy/type_visitor.py
```diff
@@ -347,3 +349,121 @@ def query_types(self, types: Iterable[Type]) -> T:
...
+    recursion), so it should *never* be re-used for querying different types,
+    create a new visitor instance instead.
+
+    # TODO: check that we don't have existing violations of this rule.
+    """
+    STRATEGY_ANY = 0  # type: Final[int]
+    STRATEGY_ALL = 1  # type: Final[int]
+
+    def __init__(self, strategy: int) -> None:
+        # 0: any()
```
msullivan, line 370: Drop this comment, since the constants have names

msullivan (commented):
This looks solid. The one question is whether we think it helps enough to be worth the extra code. I had one suggestion for a potential perf improvement, so try that out.

Also, it looks like you accidentally included a typeshed submodule update.

mypy/type_visitor.py
```diff
@@ -347,3 +349,121 @@ def query_types(self, types: Iterable[Type]) -> T:
...
+    strategy is used to combine results for a series of types,
+    Cases involve a boolean query using `any` or `all`.
+
+    Note: this visitor keeps an internal state (tracks type aliases to avoid
+    recursion), so it should *never* be re-used for querying different types,
+    create a new visitor instance instead.
+
+    # TODO: check that we don't have existing violations of this rule.
+    """
+    STRATEGY_ANY = 0  # type: Final[int]
```
sobolevn, line 366: Shouldn't it be just `: Final`?

```python


class Test(object):
    a = 1  # type: Final
    b = 2  # type: Final[int]

reveal_type(Test.a)  # Revealed type is 'Literal[1]?'
reveal_type(Test.b)  # Revealed type is 'builtins.int'
```

mypy/type_visitor.py
```diff
@@ -347,3 +349,121 @@ def query_types(self, types: Iterable[Type]) -> T:
                 self.seen_aliases.add(t)
             res.append(t.accept(self))
         return self.strategy(res)
+
+
+class TypeQueryBool(SyntheticTypeVisitor[bool]):
```
sobolevn, line 354: Related: #9602

sobolevn (commented):
👍

mypy/type_visitor.py
vbarbaresi: The doc says that if we omit the type, mypy infers it
https://mypy.readthedocs.io/en/stable/final_attrs.html#syntax-variants
I don't know what conclusion to draw from your example, is it better to have a Literal or builtins.int?

mypy/type_visitor.py
vbarbaresi: ✅

mypy/type_visitor.py
vbarbaresi: thanks for the headsup, I added the decorator for `allow_interpreted_subclasses=True`

mypy/type_visitor.py
sobolevn: I am not sure about the mypy team, but I personally prefer Literal, because it is more specific. And constants should be specific. 👍

msullivan:
Does this change the perf impact at all?

vbarbaresi:
I tried a few runs and I didn't see any perf impact, I have the exact same run times. The accuracy of my bench is probably not good enough to catch this kind of improvement

mypy/type_visitor.py
vbarbaresi: I changed them to Final 👌

vbarbaresi:
> The one question is whether we think it helps enough to be worth the extra code. 

I agree, it's not clear if it's worth it.

I want to give another shot at benchmarking this in the next few days.
I'll try another approach: running multiple iterations of mypy main entrypoint in a loop in the same Python process.
That could give more accurate results (and allow to run more iterations faster) than spawning a new process for each run.
Does that make sense? If you have other ideas, I'd be happy to try them as well

hauntsaninja:
Thanks for investigating this! Seems like we weren't able to get this to a point where it was worth it, so closing. But perf work / investigation is always appreciated :-)

PR #9606, opened 2020-10-17
[mypyc] Add 'bit' primitive type and streamline branching

JukkaL:
Add the `bit` primitive type that only has two valid values: 0 and 1. The existing
boolean primitive type is different as it also supports a third value (for error). 
This makes boolean  a poor fit for low-level IR operations which never raise an 
exception.

Use `bit` as the result type of comparison operations and various primitives.

Also simplify the branch operation to not have a special mode for branching on
negative values, since it's needlessly ad hoc. Primitives still support this mode,
but it gets converted away during IR building.

Work on mypyc/mypyc#748.


JukkaL:
I'm going to merge this as soon as I have a green build, since this will get stale very quickly. A review would still be appreciated, and I will respond to review feedback in a separate PR if this PR has already been merged.

TH3CHARLie (commented):
LGTM! Looking forward to seeing its boost on performance

JukkaL:
> Looking forward to seeing its boost on performance

This probably has no direct impact of performance, but this will be useful when working on non-C backends.

PR #9607, opened 2020-10-17
[mypyc] Simplify IR for tagged integer comparisons

JukkaL:
In a conditional context, such as in an if condition, simplify the 
IR for tagged integer comparisons. Also perform some additional 
optimizations if an operand is known to be a short integer. 

This slightly improves performance when compiling with no 
optimizations. The impact should be pretty negligible otherwise.

This is a bit simple-minded, and some further optimizations are 
possible. For example, `3 < x < 6` could be made faster. This 
covers the most common cases, however.

Closes mypyc/mypyc#758.


TH3CHARLie (approved):
LGTM! Just need to fix the `native_int` IR issue that is currently failing on 32bit platforms.

PR #9611, opened 2020-10-18
[mypyc] Add bytes primitive type

JukkaL:
This adds minimal support for the `bytes` primitive type. This won't 
improve performance until we implement some more primitives.
Actually, this will likely degrade performance a bit, since we now add 
type checks for `bytes` values.

Unfortunately, most of the `bytes` operations we'd need are private in 
CPython, so we'll need to either depend on implementation details or
reimplement (or copy+paste) a bunch of code.

Work on mypyc/mypyc#644.

### DIFF
--- a/mypyc/codegen/emit.py
+++ b/mypyc/codegen/emit.py
@@ -14,7 +14,7 @@ from mypyc.ir.rtypes import (
     is_list_rprimitive, is_dict_rprimitive, is_set_rprimitive, is_tuple_rprimitive,
     is_none_rprimitive, is_object_rprimitive, object_rprimitive, is_str_rprimitive,
     int_rprimitive, is_optional_type, optional_value_type, is_int32_rprimitive,
-    is_int64_rprimitive, is_bit_rprimitive
+    is_int64_rprimitive, is_bit_rprimitive, is_bytes_rprimitive
 )
 from mypyc.ir.func_ir import FuncDecl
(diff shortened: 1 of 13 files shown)


msullivan (approved):

JukkaL:
There were various test failures because of a bad annotation in `typed_ast` stubs. It should be fixed now.

hauntsaninja:
The match group thing keeps coming up, I'm in favour of a Union Any return (with maybe some overloads for when we don't need it): https://github.com/python/typeshed/issues/3902#issuecomment-628332769

TH3CHARLie:
What's the status quo of the match group thing? Is it still a blocker of this?

JukkaL:
I think that it's still an issue. I kind of forgot about this PR. The typeshed fix should be fairly straightforward (but it needs to be synced back to mypy).

JelleZijlstra:
`group` was already changed to return `AnyStr | Any`: https://github.com/python/typeshed/blob/master/stdlib/typing.pyi#L568

97littleleaf11:
Hi! @hauntsaninja, I'm working on bytes primitives, shall we sync the typeshed to fix current match.group issue?

hauntsaninja:
Sure, I opened https://github.com/python/mypy/pull/10862
If you're planning on building upon this PR, it'll need a merge

JukkaL:
Closing in favor of #10881.

PR #9613, opened 2020-10-19
stubgen: properly convert overloaded functions

chadrik:
### Description

Improves `stubgen` to convert functions that have been decorated with `@overload`.   `stubgen` is useful not just for creating a starting point for manually authoring stubs, but also to create automated workflows to continuously regenerate stubs for projects that want to provide stubs to third parties to hide runtime complexity from mypy. 

## Test Plan

Added two new tests.

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -559,10 +559,33 @@ class StubGenerator(mypy.traverser.TraverserVisitor):
             for name in sorted(undefined_names):
                 self.add('#   %s\n' % name)
 
-    def visit_func_def(self, o: FuncDef, is_abstract: bool = False) -> None:
+    def visit_overloaded_func_def(self, o: OverloadedFuncDef) -> None:
+        """@property with setters and getters, or @overload chain"""
+        overload_chain = False
+        for item in o.items:
+            if not isinstance(item, Decorator):
+                continue
+
+            if self.is_private_name(item.func.name, item.func.fullname):
+                continue
+
+            is_abstract, is_overload = self.process_decorator(item)
+
+            if not overload_chain:
+                self.visit_func_def(item.func, is_abstract=is_abstract, is_overload=is_overload)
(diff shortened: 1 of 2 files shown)


mypy/stubgen.py
```diff
@@ -599,7 +622,7 @@ def visit_func_def(self, o: FuncDef, is_abstract: bool = False) -> None:
                     and not is_cls_arg):
                 self.add_typing_import("Any")
                 annotation = ": {}".format(self.typing_name("Any"))
-            elif annotated_type and not is_self_arg:
+            elif annotated_type and not is_self_arg and not is_cls_arg:
```
chadrik, line 625: without this change I was getting an erroneous type for `cls`:

```python
class A:
    @typing.overload
    @classmethod
    def g(cls: Any, x: typing.Tuple[int, int]) -> int: ...
```

mypy/stubgen.py
```diff
@@ -723,7 +773,14 @@ def process_member_expr_decorator(self, expr: MemberExpr, context: Decorator) ->
                 self.add_coroutine_decorator(context.func,
                                              expr.expr.name + '.coroutine',
                                              expr.expr.name)
-        return is_abstract
+        elif (isinstance(expr.expr, NameExpr) and
+              (expr.expr.name == 'typing' or
+               self.import_tracker.reverse_alias.get('typing')) and
```
JukkaL, line 778: Should this be `expr.expr.name in ('typing', self.import_tracker.reverse_alias.get('typing'))`? Maybe add test case that uses `import typing as t` or similar.

JukkaL (commented):
Thanks, looks good! I only left one comment. This is ready to merge once it has been addressed.

chadrik:
This should be all ready to merge

chadrik:
Hi, this is still ready to merge!

Issue #4675, opened 2018-03-03
Add ability to ignore a file or directory without modifying it


adamtheturtle:
In [a project](https://github.com/mesosphere/dcos-e2e) I maintain, we vendor some third party Python code and the project includes generated files such as files created by [python-versioneer](https://github.com/warner/python-versioneer).

What I'd like to do is type check all files, including any new files, apart from those that I have explicitly ignored.

Right now I have the following hack workaround:

```python
import subprocess
import sys


def main() -> None:
    args = ['mypy', '.']
    ignore_paths = {
        'src/dcos_e2e/_vendor',
        'src/dcos_e2e/_version.py',
        'versioneer.py',
    }
    result = subprocess.run(args=args, stdout=subprocess.PIPE)
    result_lines = result.stdout.decode().strip().split('\n')
    error_lines = [
        line for line in result_lines
        if not any(line.startswith(path) for path in ignore_paths)
    ]
    print('\n'.join(error_lines))
    sys.exit(int(bool(error_lines)))


if __name__ == '__main__':
    main()
```

Various tools have mechanisms for doing this, for example:

```
Adam@MacBook-Pro ~/D/m/d/dcos-e2e> cat .isort.cfg 
[settings]
skip=_vendor,
     src/dcos_e2e/__init__.py,
     versioneer.py,
     setup.py,
```

```
[flake8]
exclude=./src/dcos_e2e/_vendor/,
        ./src/dcos_e2e/_version.py,
        ./versioneer.py,
```

```
Adam@MacBook-Pro ~/D/m/d/dcos-e2e> cat pylintrc | grep ignore= -A 3
ignore=CVS,
       _vendor,
      versioneer.py,
      _version.py,
```

Thank you for the software.

gvanrossum:
If you're just going to ignore the errors from specific paths, a more elegant solution is to use module-specific sections in your `mypy.ini` with `ignore_errors = True` -- see [config file docs](http://mypy.readthedocs.io/en/latest/config_file.html).

adamtheturtle:
I think that will work and it is something that I was not aware of and I probably should have read the docs closer. Thank you, I will close this.

danielbraun89:
> If you're just going to ignore the errors from specific paths, a more elegant solution is to use module-specific sections in your `mypy.ini` with `ignore_errors = True` -- see [config file docs](http://mypy.readthedocs.io/en/latest/config_file.html).

[...]

PR #9614, opened 2020-10-19
find_sources: find build sources recursively

### LINKED ISSUE #8548: mypy doesn't recurse through directories, but -h says it does
According to `mypy -h`:
```
Mypy is a program that will type check your Python code.

Pass in any files or folders you want to type check. Mypy will
recursively traverse any provided folders to find .py files:

    $ mypy my_program.py my_src_folder
```

But it doesn't recurse, it only checks what's explicitly provided on the command line:
```
$ python3 -m venv venv
$ . venv/bin/activate
(venv) $ pip install mypy
...
(venv) $ mkdir -p test/test/
(venv) $ touch test/test/a.py
(venv) $ mypy test
There are no .py[i] files in directory 'test'
(venv) $ touch test/b.py
(venv) $ mypy test
(venv) $ mypy test test/test
Success: no issues found in 2 source files
```

The fix might be to change the documentation.  This is easy to workaround with `find`.

hauntsaninja:
Fixes #8548

This is important to fix because it's a really common source of
surprising false negatives.

I also lay some groundwork for supporting passing in namespace packages
on the command line. The approach towards namespace packages that this
anticipates is that if you pass in files to mypy and you want mypy to
understand your namespace packages, you'll need to specify explicit
package roots.

We also make many fewer calls to crawl functions, since we just pass
around what we need.

### DIFF
--- a/test-data/unit/cmdline.test
+++ b/test-data/unit/cmdline.test
@@ -66,8 +66,10 @@ dir/a.py: error: Are you missing an __init__.py?
 # cmd: mypy dir/
 [file dir/a.py]
 undef
+import b
 [file dir/subdir/b.py]
 undef
+import a
 [out]
 dir/a.py:1: error: Name 'undef' is not defined
 dir/subdir/b.py:1: error: Name 'undef' is not defined
@@ -77,8 +79,10 @@ dir/subdir/b.py:1: error: Name 'undef' is not defined
 [file pkg/__init__.py]
 [file pkg/a.py]
 undef
+import a
 [file pkg/subdir/a.py]
 undef
+import pkg.a
 [out]
 pkg/a.py:1: error: Name 'undef' is not defined
 pkg/subdir/a.py:1: error: Name 'undef' is not defined
--- a/test-data/unit/cmdline.test
+++ b/test-data/unit/cmdline.test
@@ -45,19 +45,32 @@ pkg/subpkg/a.py:1: error: Name 'undef' is not defined
 # cmd: mypy dir
 [file dir/a.py]
 undef
-[file dir/subdir/a.py]
+[file dir/subdir/b.py]
 undef
 [out]
 dir/a.py:1: error: Name 'undef' is not defined
+dir/subdir/b.py:1: error: Name 'undef' is not defined
 
-[case testCmdlineNonPackageSlash]
-# cmd: mypy dir/
+[case testCmdlineNonPackageDuplicate]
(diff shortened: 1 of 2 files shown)


test-data/unit/cmdline.test
```diff
@@ -68,6 +81,7 @@ undef
 undef
 [out]
 pkg/a.py:1: error: Name 'undef' is not defined
+pkg/subdir/a.py:1: error: Name 'undef' is not defined
```
JukkaL, line 84: Test importing from modules found via recursive directory traversal. Test cases where the directory contains `__init__.py` and where it doesn't exist.

Maybe test this also with `--namespace-packages`?

JukkaL (approved):
Thanks for fixing this bug! This was pretty bad. Looks good -- my only comments are about additional test cases.

Additional thanks for cleaning up some of the code, even if it wasn't needed for this PR.

test-data/unit/cmdline.test
hauntsaninja: testCmdlinePackageSlash tests when the directory contains __init__.py and testCmdlinePackageContainingSubdir tests when it doesn't.
I added importing of the modules.

hauntsaninja: There currently isn't an interaction with `--namespace-packages` and cmdline tests are slow, so I won't do that unless told.
That said, I have another diff coming soon where I flesh out the package root code and make `--namespace-packages` the default. All these tests pass with that.

hauntsaninja: I'll also update the docs once all these changes (essentially important bug fixes + first three points of #8584) are shipped :-)

intgr:
Ugh, turns out many users were relying on the old behavior and this causes issues because mypy now finds unexpected Python files from `node_modules`, virtualenv paths (`venv`), temporary directories etc. And there isn't a good way to exclude those paths (feature request #4675).

For example, creating an **empty virtualenv** with Python 3.9 **breaks mypy**:
```
% mkdir test
% cd test
% python3 -m venv ./venv
% echo 'a: str = 10' > test.py
% mypy .
venv/lib/python3.9/site-packages/setuptools/_distutils/command/bdist_msi.py:355: error: expected an indented block
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I think this is a big enough regression that this change should be reverted in a hotfix release, until a longer-term solution can be figured out.

Issue #5759, opened 2018-10-09
Support namespace packages from command line


gvanrossum:
Currently the `--namespace-packages` option only applies to *imported* modules. Files, modules and packages specified on the command line (as files or with `-p` or `-m`) cannot be in namespace packages. We should fix this.

Follow-up to #5591.

gvanrossum:
Update: `-m` honors `--namespace-packages` just fine. It's just `-p` and specifying files on the command line that don't. With `-p` I get `Can't find package 'foo.bar'` and when specifying the full path mypy assumes the top-level namespace (foo) is not a package at all.

asottile:
I have a similar problem.  Here's a minimal reproduction:

```console
$ mypy --version
mypy 0.641
$ mkdir {a,b} && touch {a,b}/a.py
$ mypy a/a.py b/a.py --namespace-packages
b/a.py: error: Duplicate module named 'a'
```

I'm working around this by adding `__init__.py` files.   Note that I'm invoking in this way as I'm using `mypy` with [pre-commit](https://pre-commit.com) via [mirrors-mypy](https://github.com/pre-commit/mirrors-mypy)

gvanrossum:
It's on our list of issue, but that list is rather long. Perhaps one of you is interested in submitting a PR to help this along more quickly?

tsibley:
It seems as if @rafales started on addressing this: https://github.com/python/mypy/compare/master...rafales:namespace-packages

I might be able to find the time to make a PR, but would appreciate some guidance on where to start in order to save myself a lot of bootstrapping time understanding how mypy is put together.

russellwinstead:
It looks like it comes down to the options not being passed to the FindModuleCache object here:
https://github.com/python/mypy/blob/b027308116d19dc79bf80334ff06aab7c0001e11/mypy/main.py#L783-L784

rafales:
It's more complicated, right now mypy relies on the existence of
__init__.py files to figure out package name. If you end up with
subpackage named the same as a global package, you'll end up with
confusing errors (eg. "mycompany.graphql" and "graphql").

I did start looking into the problem and fixed it enough to work for
my usecase on the branch mentioned above. But it still needs some work
before turning it into PR.

sid-kap:
I'm getting the same error message when using `-p` on some code that uses old-style namespace packages (`__init__.py` files that contain `__import__('pkg_resources').declare_namespace(__name__)`).
[...]

PR #9616, opened 2020-10-19
modulefinder: make -p handle namespace packages correctly

hauntsaninja:
Fixes part of #5759
The other part is passing files arguments, which we make steps towards
in #9614

### DIFF
--- a/mypy/modulefinder.py
+++ b/mypy/modulefinder.py
@@ -387,7 +387,7 @@ class FindModuleCache:
                     if mod not in hits:
                         hits.add(mod)
                         result += self.find_modules_recursive(module + '.' + mod)
-        elif os.path.isdir(module_path) and any(module.startswith(p) for p in self.ns_packages):
+        elif os.path.isdir(module_path):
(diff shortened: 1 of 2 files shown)


mypy/modulefinder.py
```diff
@@ -387,13 +387,13 @@ def find_modules_recursive(self, module: str) -> List[BuildSource]:
                     if mod not in hits:
                         hits.add(mod)
                         result += self.find_modules_recursive(module + '.' + mod)
-        elif os.path.isdir(module_path) and module in self.ns_packages:
-            # Even more subtler: handle recursive decent into PEP 420
+        elif os.path.isdir(module_path) and any(module.startswith(p) for p in self.ns_packages):
```
JukkaL, line 390: I wonder if startswith is correct here. For example, consider `foobar` vs `foo`?

hauntsaninja: Good point, but now that I'm looking at the check, I think it's unnecessary (always true, and running tests with an assert seems to confirm that).

JukkaL (approved):

PR #9625, opened 2020-10-22
Improve annotated support with non types

### LINKED ISSUE #9315: Type alias for Annotated[some_type, non_type] not supported
The following code:

```py
from typing_extensions import Annotated
class Positive:...
PositiveInt = Annotated[int, Positive()]
```

triggers the following error:
```
test.py:3: error: Invalid type alias: expression is not a valid type
```

Mypy version: mypy 0.782
Python version: Python 3.7.7 (default, May 10 2020, 19:54:23)
[Clang 11.0.3 (clang-1103.0.32.59)] on darwin
Mypy config:
```ini
[mypy]
disallow_any_generics = True
disallow_untyped_calls = True
disallow_untyped_defs = True
disallow_incomplete_defs = True
disallow_untyped_decorators = True
no_implicit_optional = True
warn_return_any = True
strict_equality = True
ignore_missing_imports = False
strict_optional = True
warn_unused_ignores = True
warn_redundant_casts = True
warn_unused_configs = True
warn_unreachable = True
check_untyped_defs = True
new_semantic_analyzer = True
```

patrick91:
Closes #9315

This PR aims to support Annotated to create type Aliases when passing a non type as annotation, like this:

```python
class Meta:
    ...

x = Annotated[int, Meta()]
```

You can test with: 

```
pytest -k testAnnotatedSecondParamNonType --pdb
```

### DIFF
--- a/mypy/exprtotype.py
+++ b/mypy/exprtotype.py
@@ -65,7 +65,11 @@ def expr_to_unanalyzed_type(expr: Expression, _parent: Optional[Expression] = No
             if isinstance(expr.base, RefExpr) and expr.base.fullname in [
                 'typing.Annotated', 'typing_extensions.Annotated'
             ]:
-                base.args = (expr_to_unanalyzed_type(args[0], expr), args[1])
+                # TODO: this is not the optimal solution as we are basically getting rid
+                # of the Annotation definition and only returning the type information,
+                # losing all the annotations.
+
+                return expr_to_unanalyzed_type(args[0], expr)
             else:
                 base.args = tuple(expr_to_unanalyzed_type(arg, expr) for arg in args)
             if not base.args:
--- a/mypy/exprtotype.py
+++ b/mypy/exprtotype.py
@@ -61,7 +61,10 @@ def expr_to_unanalyzed_type(expr: Expression, _parent: Optional[Expression] = No
                 args = expr.index.items
             else:
                 args = [expr.index]
-            base.args = tuple(expr_to_unanalyzed_type(arg, expr) for arg in args)
+            if base.name == 'Annotated':
+                base.args = (expr_to_unanalyzed_type(args[0], expr), args[1])
+            else:
+                base.args = tuple(expr_to_unanalyzed_type(arg, expr) for arg in args)
             if not base.args:
                 base.empty_tuple_index = True
             return base
--- a/mypy/exprtotype.py
+++ b/mypy/exprtotype.py
@@ -3,7 +3,7 @@
 from typing import Optional
 
 from mypy.nodes import (
-    Expression, NameExpr, MemberExpr, IndexExpr, TupleExpr, IntExpr, FloatExpr, UnaryExpr,
+    Expression, NameExpr, MemberExpr, IndexExpr, RefExpr, TupleExpr, IntExpr, FloatExpr, UnaryExpr,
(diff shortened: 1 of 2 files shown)


mypy/exprtotype.py
```diff
@@ -61,7 +61,10 @@ def expr_to_unanalyzed_type(expr: Expression, _parent: Optional[Expression] = No
                 args = expr.index.items
             else:
                 args = [expr.index]
-            base.args = tuple(expr_to_unanalyzed_type(arg, expr) for arg in args)
+            if base.name == 'Annotated':
+                base.args = (expr_to_unanalyzed_type(args[0], expr), args[1])
```
patrick91, line 65: this seems to work, but it will probably fail the type checking since now the args are of different types.

mypy/exprtotype.py
```diff
@@ -61,7 +61,10 @@ def expr_to_unanalyzed_type(expr: Expression, _parent: Optional[Expression] = No
                 args = expr.index.items
             else:
                 args = [expr.index]
-            base.args = tuple(expr_to_unanalyzed_type(arg, expr) for arg in args)
+            if base.name == 'Annotated':
```
patrick91, line 64: is this a good way of checking the name? Looks like base here doesn't have a fullname

mypy/exprtotype.py
patrick91: it feels a bit hacky, let me know what you think!

patrick91: the check is potentially wrong too, but I think it might be worth checking this PR and maybe try to find a better way to do this 😊

hauntsaninja: You're correct that you're failing type checking. You can probably fix the first type error by narrowing / guarding expr.base to the type that it is when its typing.Annotated (probably `isinstance(expr.base, RefExpr)`?)

The second type error is trickier and might indicate a fundamental problem. If there isn't a way around this, we could band aid this problem by just returning `expr_to_unanalyzed_type(args[0], expr)`. I'd imagine the implication of this is that mypy and mypy plugins won't recognise the alias as being Annotated, but that's not a regression and we no longer have the false positive.

Finally, Annotated is variadic, so it can have arbitrarily many parameters, so you don't want to just hardcode it to two.

hauntsaninja (changes requested):

mypy/exprtotype.py
patrick91: > You're correct that you're failing type checking. You can probably fix the first type error by narrowing / guarding expr.base to the type that it is when its typing.Annotated (probably isinstance(expr.base, RefExpr)?)

Done :)

> The second type error is trickier and might indicate a fundamental problem. If there isn't a way around this, we could band aid this problem by just returning expr_to_unanalyzed_type(args[0], expr). I'd imagine the implication of this is that mypy and mypy plugins won't recognise the alias as being Annotated, but that's not a regression and we no longer have the false positive. 

I followed your suggestion for now, mostly to see if there's any other failing test.
I wonder if need the annotations anywhere else (I guess it might useful for plugins 🤔)

Happy to change it to something else, but I'm not sure what can we change now (I'm still unfamiliar with mypy's source code) :)

hauntsaninja (approved):
Thanks! I'm okay with this, but I'll leave it for a bit to see if someone else has something to say (I don't really know much about plugins)

JukkaL:
Looks good. Thanks for the PR!

PR #9634, opened 2020-10-24
[mypyc] Fix type of for loop index register in for over range

JukkaL:
It sometimes was short int even though it should have been int.


TH3CHARLie (approved):
LGTM

Issue #7276, opened 2019-07-31
"Intermittent" import error when checking a module that import another module which imports a 3rd party package


Quibi:
Hi,

Sorry for the title. I try to explain myself better.

I have found a strange behaviour in `mypy` related to an import error in a imported module. That is, an error in a module 'b' importing a module 'c' while checking module 'a', which imports 'b'. 'c', of course, is a 3rd party package.

In such a case, `mypy` reports nothing on a first round. But if it is run again, it reported a `error: Cannot find module named 'c'`. On a third run, nothing is reported again, while on a fourth run the error re-appears, and so on...

This happens even when the type check has been disabled in the import line.

A possible scenario to reproduce this "bug" is as follows:

1. Create a 'strange_report' package (with an empty `__init__.py`).

2. Create a subpackage called 'a' inside 'strange_report' (with an empty `__init__.py` too).

3. Write an `a.py` module (importer module) such as:
```
from a.b import B

b = B()  # just for visually check that a.py was run
```
4. Write a `b.py` module (imported module)
```
from ruamel.yaml import YAMLError  # type: ignore


class B:
    def __init__(self):
        print(f"New 'B' object {self} has been created")
```

5. Open terminal and change to directory 'strange_report'.

6. Module `a.py` can be run with `python -m a.a`

7. If you run `mypy a/a.py` for the first time, you do not get any reports (however, directory '.mypy_cache' is created).

8. But if you run `mypy a/a.py` once again, you get an error:

```
madera:strange_error jv$ mypy a/a.py
a/b.py:1: error: Cannot find module named 'ruamel'
a/a.py:1: error: Cannot find module named 'ruamel'
a/a.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```
9. You can repeat 7. and 8. again and again with similar results. Same results are gotten with or without `# type: ignore`.

However, **if cache directory '.mypy_cache' is deleted** before running mypy every time, it **does not report anything**.

Curiously enough, if you check `mypy a/b.py` it consistently reports nothing.

I am running on:

- macOS HIgh Sierra 10.13.6
- Python 3.7.4
- mypy 0.720
- ruamel.yaml 0.16.0 (so, yes, ruamel.yaml is installed; I use this 3rd party package because it is causing the bug in my system; no any particular purposes, I imagined that same bug can appeared with similar packages but cannot confirm that)
- Any mypy configuration beyond the default one.

Thank for your attention. I hope that will be helpful.

[...]

Issue #8545, opened 2020-03-16
Allow/disallow pep561 packages


emmatyping:
@alexander255 has requested a flag `--pep561-override` to allow listing packages as pep 561 compatible when they are not. This got me thinking and I think the best UX for this would be a pair of flags: `--allow-installed-package NAME` and `--disallow-installed-package NAME` (or something like that).

The idea is to solve two problems:

Some people publish PEP 561 packages that are either too unstable or flat out broken for some users. This allows the user to disable the package until things are fixed.

At the other end of the spectrum, some packagers are conservative on adding types, and while they exist, they are not public (yet). Some users would still like to use them.

This would eliminate `--no-site-packages`, which is a rather blunt instrument.

JukkaL:
I can see how this could be useful., and this shouldn't be hard to implement 

Random thoughts:

* What about `--use-installed-package X` as the option name? The inverse would be `--ignore-installed-package X`, perhaps. More ideas:
  * `--enable-installed-package X` and `--disable-installed-package X`
  * `--use-site-package X` and `--no-site-package X`
  * `--site-package X` and `--no-site-package X`
  * `--installed-package X` and `--no-installed-package X`
* We can start with `--use-site-package` (or whatever it would be called) and the inverse can be implemented later.  The inverse seems less commonly useful.
* Should this work for subpackages? What about things like `--use-site-package google.protobuf`?

ethanhs:
I think I like either `--enable-installed-packages X` or `--use-installed-packages X`, ideally I think this should be a list of packages, as people may want to allowlist/denylist groups of packages.

msullivan:
I am a "-0" on this; I don't like adding more complexity to our import system. But also I'd be happy to merge it. Let's make a decision on it so we can resolve #8512.

ethanhs:
Yeah I'm kinda -0 on this too. The complexity would be unfortunate, but I can kinda see the use case.

Vozf:
Just to clarify all selected flags will be usable via mypy.ini, correct?

embray:
:+1: for this.  I am using some packages that are fully typed and are PEP-561 compatible in principle (they have a `py.typed`) but due to a packaging bug (third-party package, outside my control) were missing the `py.typed` file when installed.

I can of course work around it by creating the `py.typed` file in the `site-packages/<package_name>` but it would be more convenient if I could tell mypy "yes, this package really does have type hints".

davidism:
[...]

PR #9636, opened 2020-10-24
Make --namespace-packages the default

hauntsaninja:
Unfortunately, a number of tests fail. Please help fixing them :-)
Especially the ones to do with cache inconsistencies... There's at least
one known bug here: #7276


JukkaL:
Great stuff! 

Some of the test failures can be tricky to debug; I can help with them.

hauntsaninja:
Yeah, I'd love help here, maybe you could take a look at testAddedMissingStubsPackage or testDeletePackage4?

JukkaL:
> Yeah, I'd love help here, maybe you could take a look at testAddedMissingStubsPackage or testDeletePackage4?

Sure, I should have time to look at these later this week.

hauntsaninja:
Spent some time looking at this today. I was able to fix the "cache data discrepancy" errors.

I also spent some time looking at mypy daemon / fine grained errors, but that turned out to be a real rabbit hole. Since `--follow-imports=normal` support is explicitly experimental, I think it should be non-blocking for this PR (ie, I can make tests pass by adding `--no-namespace-packages`).

JukkaL:
Sorry, I lost of track of this PR. Is there anything I can do to help move this forward?

nipunn1313:
#11257, #11259, and #11261 should all help unblock this

ethanhs:
We should probably add https://github.com/python/mypy/issues/10310 and other exclude usability improvements before putting this in a release because it seems that this is picking up tools/test/misc directories that people may not want checked.

Akuli:
It just feels wrong to have one global setting that affects every library you import, regardless of whether it's a dependency of a dependency, or a part of the project are working on where you forgot to put `__init__.py`.

hauntsaninja:
Thanks to Nipunn's efforts CI is now green. There's still some stuff to be fixed, see https://github.com/python/mypy/issues/11259

mypy/build.py
```diff
@@ -1321,8 +1321,10 @@ def validate_meta(meta: Optional[CacheMeta], id: str, path: Optional[str],
         st = manager.get_stat(path)
     except OSError:
         return None
-    if not (stat.S_ISREG(st.st_mode) or stat.S_ISDIR(st.st_mode)):
-        manager.log('Metadata abandoned for {}: file {} does not exist'.format(id, path))
+    if not stat.S_ISDIR(st.st_mode) and not stat.S_ISREG(st.st_mode):
+        manager.log(
+            'Metadata abandoned for {}: file or directory {} does not exist'.format(id, path)
```
nipunn1313, line 1326: this seems unrelated? Useful - but not related.

nipunn1313:
#11322 - is the remaining daemon issue.

I don't think we should make it a blocker as it only affects the rare scenario in which you delete an `__init__.py` file to intentionally convert into a namespace package. Workaround exists (restart daemon).

Code looks good to me!

nipunn1313 (approved):

mypy/build.py
hauntsaninja: I think there was a change here that was needed, but then got merged in another PR

JukkaL:
Great, let's get this merged! I'd like to play around with this a little before merging, since this kind of a big change, but don't wait for me if I get distracted and you don't hear anything from me in a few days.

mypy/build.py
nipunn1313: Yeah - I think remove this from this commit/PR before landing.
Better to keep the diff scoped to what's necessary.

nipunn1313:
Hey friends! Can we give this one a rebase and try to get it in? Thanks!

(I've been blocking https://github.com/python/typeshed/pull/6106 on this)

97littleleaf11:
cc @JukkaL @hauntsaninja

hauntsaninja:
Nipunn / typeshed are blocked on us and this change is ready, so we probably shouldn't delay past the next release. But given the breakage this will cause to people's setups I'd vaguely prefer it if @JukkaL were to hit merge.

I'm also curious if people have thoughts on how we could make this change go smoother. A couple thoughts:
- Maybe we can make "Duplicate module named ..." not cause fatal build errors, or make the error note give clearer advice?
- Similar thing for "Source file found twice under different module names"... I added a link to the docs in #11270, but we could more clearly say that the action to be taken is probably a) putting something on mypy path, b) adding an `__init__.py`, or c) using `exclude`
- Fix https://github.com/python/mypy/issues/11322
- Fix https://github.com/python/mypy/issues/9716
- If there are other breaking changes we could bundle with this in the next release (to avoid a drip-drip-drip of breaking changes). Candidates changes include: https://github.com/python/mypy/issues/8584#issuecomment-927361405

srittau:
I think @JukkaL gave his approval in #12021.

mypy/main.py
```diff
@@ -516,7 +516,7 @@ def add_invertible_flag(flag: str,
         title='Import discovery',
         description="Configure how imports are discovered and followed.")
     add_invertible_flag(
-        '--namespace-packages', default=False,
+        '--no-namespace-packages', dest="namespace_packages", default=True,
         help="Support namespace packages (PEP 420, __init__.py-less)",
```
97littleleaf11, line 520: Please also update the documentation. 
> https://mypy.readthedocs.io/en/stable/command_line.html#import-discovery
> https://mypy.readthedocs.io/en/stable/config_file.html?highlight=namespace-packages#confval-namespace_packages

hauntsaninja: Thanks, updated!

ethanhs:
Since we are already going to potentially have breakage due to https://github.com/python/mypy/pull/11143, and I plan on working on https://github.com/python/mypy/issues/8545 in the next week, perhaps this release is a good time to push this change?

ofek:
Can this be merged?

nipunn1313:
Wanted to check in here to see if its possible to go forward with this.

AlexWaygood:
Looks like this caused the selfcheck to start failing: https://github.com/python/mypy/actions/runs/3133286021/jobs/5086495727

AlexWaygood:
> Looks like this caused the selfcheck to start failing on master: https://github.com/python/mypy/actions/runs/3133286021/jobs/5086495727

Fixed in #13748

PR #9638, opened 2020-10-25
runtests.py: list tests to run in typeshed CI

hauntsaninja:
Running a subset of tests will help speed up typeshed CI. I don't think
typeshed really benefits from running all the other tests. Worst case if
something breaks, it'll be caught when typeshed is synced in mypy.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -79,6 +79,8 @@ cmds = {
          MYPYC_EXTERNAL,
          MYPYC_COMMAND_LINE,
          ERROR_STREAM]),
+    # Test cases to run in typeshed CI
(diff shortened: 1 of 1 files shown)


runtests.py
```diff
@@ -79,6 +79,8 @@
          MYPYC_EXTERNAL,
          MYPYC_COMMAND_LINE,
          ERROR_STREAM]),
+    # Test cases to run in typeshed CI
+    'typeshed-ci': 'pytest -k "%s"' % ' or '.join([CMDLINE, EVALUATION, SAMPLES, TYPESHED]),
```
JukkaL, line 83: Also update `DEFAULT_COMMANDS` below to not include this (or maybe get rid of it, since it's super error-prone, and use a flag here or something).

PR #9648, opened 2020-10-27
checkexpr: [minor] move node specific logic into its visit

hauntsaninja:
The has_uninhabited_component was formerly there to avoid clobbering the
store_type call in find_partial_type_ref_fast_path. But this interaction
is avoided by making the store_type call earlier.


JukkaL (approved):
I don't fully follow what is going on, but test are passing and it's good to move code away from `accept()` which is a very hot function.

PR #9651, opened 2020-10-27
Re-add fallback for zero terminal width

The-Compiler:
Contrary to what I assumed in #9143, shutil.get_terminal_size() doesn't
actually handle a 0 width from os.get_terminal_size() - it only handles
a 0 COLUMNS environment variable.

Thus, this caused #8144 to regress.

This change re-adds and uses DEFAULT_COLUMNS and also adds the test
which was deemed unnecessary in #8145 - this regression proves that I'd
have been a good idea to add it in the first place.
(Test written by Anthony Sottile)

Fixes https://github.com/pre-commit/mirrors-mypy/issues/29

### DIFF
--- a/mypy/test/testutil.py
+++ b/mypy/test/testutil.py
@@ -1,0 +1,12 @@
+import os
+from unittest import mock, TestCase
+
+from mypy.util import get_terminal_width
+
+
+class TestGetTerminalSize(TestCase):
+    def test_get_terminal_size_in_pty_defaults_to_80(self) -> None:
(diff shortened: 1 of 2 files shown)


The-Compiler:
FWIW I also opened a [Python bug](https://bugs.python.org/issue42174) about this now, since I feel like `shutil.get_terminal_size()` should handle this correctly (rather than mypy needing to work around it).

JukkaL (approved):
Thanks for the fix! This is clearly an area with some tricky edge cases, so adding a test makes sense.

PR #9660, opened 2020-10-29
[mypyc] Merge yield_from_except_op

TH3CHARLie:
relates mypyc/mypyc#753

This PR merges `yield_from_except_op` into the new IR, including several changes:

- A change to C wrapper's signature since the new IR currently has no unified way to represent the address of a pointer.
- A change to `LoadAddress`, allowing it to load local reg's address.
- A change to `uninit` pass, suppressing checks on `LoadAddress`


mypyc/lib-rt/CPy.h
```diff
@@ -458,7 +458,7 @@ static inline bool CPyFloat_Check(PyObject *o) {
 
 PyObject *CPy_GetCoro(PyObject *obj);
 PyObject *CPyIter_Send(PyObject *iter, PyObject *val);
-int CPy_YieldFromErrorHandle(PyObject *iter, PyObject **outp);
+int CPy_YieldFromErrorHandle(PyObject *iter, PyObject *outp);
```
JukkaL, line 461: `PyObject *` can be confusing here. What about using `void *`, and adding a comment saying that it's actually `PyObject **`, but we can't represent it in IR? Alternatively, we could perhaps use `CPyPtr`.

Or maybe the best option would be to add a new primitive type for `PyObject **` (see below for more about this).

mypyc/lib-rt/misc_ops.c
```diff
@@ -41,8 +41,9 @@ PyObject *CPyIter_Send(PyObject *iter, PyObject *val)
 // Returns false (0) if a value should be yielded.
 // In both cases the value is stored in outp.
 // Signals an error (2) if the an exception should be propagated.
-int CPy_YieldFromErrorHandle(PyObject *iter, PyObject **outp)
+int CPy_YieldFromErrorHandle(PyObject *iter, PyObject *_outp)
 {
+    PyObject **outp = (PyObject **)_outp;
```
JukkaL, line 46: Add comment explaining why we need to use a cast here.

mypyc/primitives/misc_ops.py
```diff
@@ -55,21 +55,19 @@
...
 # Op used for "yield from" error handling.
 # See comment in CPy_YieldFromErrorHandle for more information.
-yield_from_except_op = custom_op(
-    name='yield_from_except',
-    arg_types=[object_rprimitive],
-    result_type=yield_from_rtuple,
-    error_kind=ERR_MAGIC,
-    emit=simple_emit('{dest}.f0 = CPy_YieldFromErrorHandle({args[0]}, &{dest}.f1);'))
+yield_from_except_op = c_custom_op(
+    arg_types=[object_rprimitive, object_rprimitive],
```
JukkaL, line 67: Here the `object_rprimitive` is potentially confusing. Add comment explaining the real type of the second argument. Alternatively, if we'd use `pointer_rprimitive`, this would be arguably be cleaner. 

Another random idea: what about adding a new primitive type corresponding to `PyObject **`? We could call it `object_pointer_rprimitive`, perhaps.

mypyc/transform/uninit.py
```diff
@@ -45,7 +46,8 @@ def split_blocks_at_uninits(env: Environment,
                 # initialized is an operand to something other than a
                 # check that it is defined, insert a check.
                 if (isinstance(src, Register) and src not in defined
-                        and not (isinstance(op, Branch) and op.op == Branch.IS_ERROR)):
+                        and not (isinstance(op, Branch) and op.op == Branch.IS_ERROR)
+                        and not isinstance(op, LoadAddress)):
```
JukkaL, line 50: Why do we need this? Add comment explaining this, since this is a bit surprising. I wonder if we can somehow get rid of this special case.

mypyc/ir/ops.py
```diff
@@ -1505,16 +1505,22 @@ class LoadAddress(RegisterOp):
     error_kind = ERR_NEVER
     is_borrowed = True
 
-    def __init__(self, type: RType, src: str, line: int = -1) -> None:
+    def __init__(self, type: RType, src: Union[str, Register], line: int = -1) -> None:
```
JukkaL, line 1508: Add docstring explaining the difference between `str` and `Register` sources.

JukkaL (commented):
Thanks -- this is a tricky one! I have some thoughts about potentially making this cleaner. It's kind of awkward to have to use casts from `PyObject *` to `PyObject **`.

JukkaL:
Can you check that there is some existing case that uses the primitive? If not, it makes sense to add one (but I assume that this is tested somewhere).

TH3CHARLie:
> Can you check that there is some existing case that uses the primitive? If not, it makes sense to add one (but I assume that this is tested somewhere).

yes, there's `testYield` for this(but I think it's not completely for this op itself but also so other yield and exception stuff)

mypyc/transform/uninit.py
TH3CHARLie: We need this because we are taking the address of an uninitialized object, the automatically inserted initialization check will produce incorrect code, leading the control flow directly to uninitialized error rather than the following load address and other ops.

TH3CHARLie:
I've made some updates according to your review, including using `object_pointer_rprimitive` and a new IR test. Unfortunately, the special case in the uninit pass is somewhat inevitable

mypyc/test-data/irbuild-basic.test
```diff
@@ -3637,3 +3637,201 @@ L0:
     c = r2
     r3 = (c, b, a)
     return r3
+
+[case testYieldFromExcept]
+from typing import Any
```
JukkaL, line 3642: Yikes, the IR is quite verbose. I think it's fine to remove this and rely on a run test.

JukkaL (approved):
Thanks for the updates! Looks good.

I'd suggest dropping the test case (but even then it has been useful to see what sort of code is being generated).

PR #9662, opened 2020-10-29
[mypyc] Merge fast_isinstance_op and py_calc_meta_op

TH3CHARLie:
relates mypyc/mypyc#753

This PR merges two remaining misc ops.

The problem with these two is that they rely on inlined `PyTypeObject *` cast during the call, while the casted object has already gone through a cast from `PyTypeObject *` to `PyObject *` before the function call(due to the highly C-coupled design in some other parts of the IR).

Thus, I think the best way(at least for now) is to use inline wrappers to merge these two ops, avoiding introducing over-complicated designs for only these two ops.

### DIFF
--- a/mypyc/lib-rt/CPy.h
+++ b/mypyc/lib-rt/CPy.h
@@ -456,6 +456,14 @@ static inline bool CPyFloat_Check(PyObject *o) {
     return PyFloat_Check(o) || PyLong_Check(o);
 }
 
+static inline bool CPy_TypeCheck(PyObject *o, PyObject *type) {
+    return PyObject_TypeCheck(o, (PyTypeObject *)type);
+}
+
+static inline PyObject *CPy_CalculateMetaclass(PyObject *type, PyObject *o) {
+    return (PyObject *)_PyType_CalculateMetaclass((PyTypeObject *)type, o);
+}
+
 PyObject *CPy_GetCoro(PyObject *obj);
 PyObject *CPyIter_Send(PyObject *iter, PyObject *val);
 int CPy_YieldFromErrorHandle(PyObject *iter, PyObject **outp);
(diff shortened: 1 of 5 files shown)


mypyc/lib-rt/CPy.h
```diff
@@ -456,6 +456,14 @@ static inline bool CPyFloat_Check(PyObject *o) {
     return PyFloat_Check(o) || PyLong_Check(o);
 }
 
+static inline bool CPy_TypeCheck(PyObject *o, PyObject *type) {
```
JukkaL, line 459: Add TODO comment about figuring out how to avoid the inline function in non-C back ends that may not be able to use the inline function. This may be worth creating an issue about, since this is performance-critical, I think.

mypyc/lib-rt/CPy.h
```diff
@@ -456,6 +456,14 @@ static inline bool CPyFloat_Check(PyObject *o) {
...
+    return PyObject_TypeCheck(o, (PyTypeObject *)type);
+}
+
+static inline PyObject *CPy_CalculateMetaclass(PyObject *type, PyObject *o) {
+    return (PyObject *)_PyType_CalculateMetaclass((PyTypeObject *)type, o);
```
JukkaL, line 464: This use of casts is a good way to work around the pointer type mismatches, and it's not a problem if we can't inline this on non-C backends, since this isn't performance critical.

JukkaL (approved):
Looks good! The use of inline functions as a stopgap measure is a good idea, and it will allow us to remove the old primitive machinery.

Just one comment about a TODO command and/or issue about one of the inline functions.

TH3CHARLie:
The issue: https://github.com/mypyc/mypyc/issues/774

PR #9670, opened 2020-10-30
Improve Suggestion for empty TupleType syntax error

jonshea:
## Description

The mypy syntax for a function or method that takes zero parameters
is `() -> …`. Some new mypy users will reason by symmetry that the
syntax for a method that returns nothing is likely `(…) -> ()`.
A user who incorrectly annotates a function with `… -> ()` will be
given the suggestion `Use Tuple[T1, ..., Tn] instead of (T1, ..., Tn)`.
This suggestion is unlikely to help correct the user's misconception
about how to annotate the return type of a method that does not
explicitly return a value.

This PR adds a case to TupleType syntax error handling that returns a
helpful suggestion in the case where the tuple contains zero items.

Note: The error message casing in TupleType has grown large enough
that it likely warrants relocation to `MessageBuilder`, but as there
is not a `MessageBuilder` in accessible from `TypeAnalyzer` I have
decided to add this single error case the easy way. There is a
preexisting comment about the inaccessibility of `MessageBuilder`
in `RypeAnalyzer`'s `cannot_resolve_type method`.

## Test Plan

I have added a test to `check-functions.test` that verifies the new
suggestion is printed when `-> ()` is used as a return type
annotation. I have also tested that a valid return type of
`-> Tuple[()]` remains without error.

### DIFF
--- a/mypy/typeanal.py
+++ b/mypy/typeanal.py
@@ -538,7 +538,8 @@ class TypeAnalyser(SyntheticTypeVisitor[Type], TypeAnalyzerPluginInterface):
         if t.implicit and not self.allow_tuple_literal:
             self.fail('Syntax error in type annotation', t, code=codes.SYNTAX)
             if len(t.items) == 0:
-                self.note('Suggestion: Use None or Tuple[()] instead of ()', t, code=codes.SYNTAX)
+                self.note('Suggestion: Use Tuple[()] instead of () for an empty tuple, or '
+                'None for a function without a return value', t, code=codes.SYNTAX)
             elif len(t.items) == 1:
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):

hauntsaninja:
Thanks for this change!
Since the current language kind of conflates `Tuple[()]` and `None`, maybe we expand it a little to something like: `Suggestion: Use Tuple[()] instead of () for an empty tuple (or None for a function without a return value)`

jonshea:
Sounds good to me. It seemed like the existing error messages and suggestions are very terse, so I tried imitate that style, but I like your expanded version much better.

hauntsaninja:
Thanks again!

Issue #7672, opened 2019-10-09
Semi-verbose flag that shows which files mypy analyzes


gvanrossum:
Sometimes (example: #6385) it's useful to be able to see exactly which files mypy is processing. Currently the only way is to use `-v` and grep stderr for the word "Parsing".

Maybe this is common enough that we could spare a new command-line flag? It would print all the files for which a BuildSource object is created, *and* all the files that are reached by following imports (with some indicator of which is which).

The order in which files are printed should reflect the order in which they are processed (not the order in which they are specified).

I wouldn't mind adding extra info for each file (e.g. whether its cache entry is used), as long as it remains a single line per file.

Presumably there's also value in printing certain failures, e.g. modules which can't be found but for which no error message is printed (e.g. through `--follow-imports=silent`). But I would stop here.

JukkaL:
I can see how this could be useful. One option would be to switch `-v` to this mode, and move the existing verbose flag to `-vv` or something.

gvanrossum:
I’m okay with overloading -v for this.

A progress bar would be a separate issue, please file one.

hauntsaninja:
#9672 does the first part of this

JakeSummers:
In the short term, you can use:

```
mypy  <your-args>.              2> >(grep "Found source") | sed "s_.*path='\(.*\)py'.*_\1py_"
```

PR #9672, opened 2020-10-31
build: log build sources with -v

hauntsaninja:
With the changes I've been making to mypy's import handling, I think
this would be a useful thing to add preemptively.

Note that I would have found this very useful at points, and I think
others would too, eg #7672 and #8584

The existing logging ignores source_modules and source_text and doesn't
help with determining what mypy things the module name for a given file
is. This is useful for namespace package issues as in the complaint in #8584.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2541,36 +2541,24 @@ def skipping_ancestor(manager: BuildManager, id: str, path: str, ancestor_for: '
                           severity='note', only_once=True)
 
 
-def log_configuration(manager: BuildManager) -> None:
+def log_configuration(manager: BuildManager, sources: List[BuildSource]) -> None:
     """Output useful configuration information to LOG and TRACE"""
 
     manager.log()
     configuration_vars = [
         ("Mypy Version", __version__),
         ("Config File", (manager.options.config_file or "Default")),
-    ]
-
-    src_pth_str = "Source Path"
-    src_pths = list(manager.source_set.source_paths.copy())
-    src_pths.sort()
-
-    if len(src_pths) > 1:
-        src_pth_str += "s"
-        configuration_vars.append((src_pth_str, " ".join(src_pths)))
-    elif len(src_pths) == 1:
-        configuration_vars.append((src_pth_str, src_pths.pop()))
-    else:
-        configuration_vars.append((src_pth_str, "None"))
-
-    configuration_vars.extend([
         ("Configured Executable", manager.options.python_executable or "None"),
         ("Current Executable", sys.executable),
         ("Cache Dir", manager.options.cache_dir),
         ("Compiled", str(not __file__.endswith(".py"))),
-    ])
+    ]
 
     for conf_name, conf_value in configuration_vars:
         manager.log("{:24}{}".format(conf_name + ":", conf_value))
+
+    for source in sources:
+        manager.log("{:24}{}".format("Found source:", source))
 
     # Complete list of searched paths can get very long, put them under TRACE
     for path_type, paths in manager.search_paths._asdict().items():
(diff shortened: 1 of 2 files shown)


gvanrossum:
Hmm, I see that when I turn this on to mypy itself, it claims that some sources from typeshed/stdlib/2 and typeshed/thir_party/2 are found, e.g.
```
LOG:  Found source:           BuildSource(path='mypy/typeshed/stdlib/2/distutils/__init__.pyi', module='distutils', has_text=False, base_dir='mypy/typeshed/stdlib/2')
LOG:  Found source:           BuildSource(path='mypy/typeshed/stdlib/2/distutils/command/__init__.pyi', module='distutils.command', has_text=False, base_dir='mypy/typeshed/stdlib/2')
LOG:  Found source:           BuildSource(path='mypy/typeshed/stdlib/2/distutils/command/bdist.pyi', module='distutils.command.bdist', has_text=False, base_dir='mypy/typeshed/stdlib/2')
.
.
.
LOG:  Found source:           BuildSource(path='mypy/typeshed/third_party/2/ipaddress.pyi', module='ipaddress', has_text=False, base_dir='mypy/typeshed/third_party/2')
LOG:  Found source:           BuildSource(path='mypy/typeshed/third_party/2/pathlib2.pyi', module='pathlib2', has_text=False, base_dir='mypy/typeshed/third_party/2')
LOG:  Found source:           BuildSource(path='mypy/typeshed/third_party/2/pymssql.pyi', module='pymssql', has_text=False, base_dir='mypy/typeshed/third_party/2')
```
(A total of 265 files, out of 1436 reported.)

Any idea what's up with that?

gvanrossum:
Hm, I'm guessing this is the new, improved file search? This was from
```
python -m mypy -v mypy
```
It does the right thing when I run
```
python -m mypy -v -p mypy
```
IIRC the former used to work. Well, I guess I never liked typeshed being embedded inside mypy (nor the test subdir for that matter), but here we are.

gvanrossum:
IOW I realize I have to declare in favor of this PR, since this actually showed the error of my ways... :-)

hauntsaninja:
:-)
I also noticed the error message isn't accurate, for which I made #9674!

gvanrossum (approved):
But I still want a -v flag that just prints the list of sources, and not all the other stuff that’s logged by -v. That could become -vv.

Also, wouldn’t it be nice if each source file was logged just before it got checked, instead of all ahead of the checking? Then it would intersperse this activity log with the errors, as found.

hauntsaninja:
Yeah that would be nice, and there's still the "show all files reached by following imports" part of your original request.

It also occurred to me that `python -m mypy --config-file mypy_self_check.ini mypy --namespace-packages -v` should also end up checking all of typeshed, but it doesn't. So #9616 isn't quite the fix I thought it was...

gvanrossum:
Presuming you meant ‘-p mypy’, maybe it just stops descending when it finds a package named ‘mypy’?

hauntsaninja:
Yeah, I meant `-p mypy` (edited above). The issue is that modulefinder recurses differently in the presence of `__init__.py`; we hit [this check](https://github.com/python/mypy/blob/fd16f84a2d692ce3712d3da3251561d6af03261b/mypy/modulefinder.py#L380) even if we allow namespace packages, which is not what we want

samskiter:
> Also, wouldn’t it be nice if each source file was logged just before it got checked, instead of all ahead of the checking? Then it would intersperse this activity log with the errors, as found.

I know you posted this a while ago but definitely a big YES to this. - Currently trying to help fix this VSCode mypy extension and that feature would come in might handy for this bug https://github.com/microsoft/vscode-mypy/pull/123 

Any suggestions for a quick fix? If the output can include files checked, even if they have no error, it makes clearing errors on files a lot easier...

PR #9673, opened 2020-10-31
test-requirements: bump minimum pytest needed

### LINKED ISSUE #9671: mypy fails self check when run on Python 3.8
```
mypy/test/data.py:272: error: Call to untyped function "_prunetraceback" in typed context  [no-untyped-call]
                self.parent._prunetraceback(excinfo)
```

hauntsaninja:
pytest 6.0 causes type checking to fail against master.
Fixes #9671


JukkaL:
Thanks!

PR #9675, opened 2020-10-31
docs: remove mention of missing standard library stubs

hauntsaninja:
I was looking at #4542, an old issue that I think we've mostly
addressed, but nonetheless has an alarming number of likes. It occurs to
me that front and centre we mention an issue that most people never run
into in practice.

I also pulled the details of this. mypy's hardcoded list of stdlib
modules is incomplete, but here are the ones on the list that we do not
have stubs for (I only checked Python 3):
```
{'__dummy_stdlib1',
 '__dummy_stdlib2',
 'ossaudiodev',
 'sqlite3.dump',
 'turtledemo',
 'xml.dom.expatbuilder',
 'xml.dom.minicompat',
 'xml.dom.xmlbuilder',
 'xml.sax._exceptions',
 'xml.sax.expatreader',
 'xxlimited'}
 ```
 We should maybe consider getting rid of the hardcoded list altogether.

### DIFF
--- a/docs/source/running_mypy.rst
+++ b/docs/source/running_mypy.rst
@@ -133,9 +133,8 @@ follow the import.
 
 This can cause errors that look like the following::
 
-    main.py:1: error: No library stub file for standard library module 'antigravity'
-    main.py:2: error: Skipping analyzing 'django': found module but no type hints or library stubs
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):
It seems highly unlikely now that users would run into missing standard library stubs.

PR #9699, opened 2020-11-04
[mypyc] Cleanup old style primitive code

TH3CHARLie:
We have moved all the existing ops into the new IR, so this PR removes the old `PrimitiveOp` and related code

relates https://github.com/mypyc/mypyc/issues/709, closes https://github.com/mypyc/mypyc/issues/753


JukkaL (approved):
Excellent, this is a big milestone! I love seeing so much legacy code deleted. The old primitives are now officially dead.

PR #9701, opened 2020-11-05
disable_errors() and enable_errors() function pair refactored

### LINKED ISSUE #1184: Coding style suggestion about nesting constructs
There are a lot of different patterns in the mypy codebase where some stack is maintained by a pair of functions that push something onto the stack (or bump a counter or something) and pop it off again. So you'll see code like this:

``` python
            self.msg.disable_errors()
            arg_types = self.infer_arg_types_in_context(None, args)
            self.msg.enable_errors()
```

I wonder if this pattern couldn't be made clearer in the code by using context managers. In this case, we would write e.g.

``` python
            with self.msg.disable_errors():
                arg_types = self.infer_arg_types_in_context(None, args)
```

The implementation would be very simple:

``` python
@contextmanager
def disable_errors(self) -> None:
    self.disable_count += 1
    yield
    self.disable_count -= 1
```

(I'm picking on this example because it came up recently, but there are a lot of function pairs like this; e.g. check_func_item() (https://github.com/JukkaL/mypy/blob/master/mypy/checker.py#L542) has four different stacks it manipulates.)

kishdubey:
### Description
Fixes #1184. Refactored occurrences for disable_errors and enable_errors() as described in the issue. 
./runtests.py , runs without errors.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1992,10 +1992,9 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             if isinstance(s.expr, EllipsisExpr):
                 return True
             elif isinstance(s.expr, CallExpr):
-                self.expr_checker.msg.disable_errors()
-                typ = get_proper_type(self.expr_checker.accept(
-                    s.expr, allow_none_return=True, always_allow_any=True))
-                self.expr_checker.msg.enable_errors()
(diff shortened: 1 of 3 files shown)


gvanrossum:
@kishdubey Do you need help debugging the test failure?

JelleZijlstra:
#10569 was merged

PR #9705, opened 2020-11-06
Ambiguous **kwargs only map to unmapped arguments that have a matching type.

### LINKED ISSUE #9676: Spurious error with dynamic keyword arguments
**Bug Report**

The expression `datetime.timedelta(**{"seconds": 1.0})` works correctly in python, but mypy reports it as an error *if* the `python-version` parameter is 3.6 or greater. (note that the [typeshed definition of `timedelta.__init__` ](https://github.com/python/typeshed/blob/301325b55b1d523c67056be606d0bf771ca5551c/stdlib/2and3/datetime.pyi#L160-L183) is conditioned on the python version, which explains why that parameter has this effect)

**To Reproduce**

`test.py`:
```
import datetime

datetime.timedelta(**{"seconds": 1.0})
```

Run `mypy --python-version 3.5 test.py` and `mypy --python-version 3.6 test.py`

**Actual Behavior**

With `--python-version 3.5`, there are no errors. With `--python-version 3.6`, it reports:

```
test.py:3: error: Argument 1 to "timedelta" has incompatible type "**Dict[str, float]"; expected "int"
```

**Your Environment**

- Mypy version used: tested both 0.740 and 0.790
- Mypy command-line flags: `--python-version`
- Python version used: 3.8
- Operating system and version: linux 

Minimal test case: https://repl.it/@bdarnell/mypy-dynamic-keyword-args-bug-report
CI build in which this was initially discovered: https://travis-ci.org/github/tornadoweb/tornado/jobs/740396440

esoma:
### DO NOT MERGE

Just getting some eyes on this for now. There are some consequences to this change discussed in #9676 and that I'll continue here. In any case, should probably wait for #9629 to be merged as there will probably be some merge conflicts.

### Description

Closes #9676

- non-TypedDict **kwargs will only map to parameters which have a compatible type
- introduces a new check for non-TypedDict **kwargs that are not able to map to at least 1 parameter

## Test Plan

Added the `testCallKwargsDifferentTypes` test.

### DIFF
--- a/test-data/unit/check-kwargs.test
+++ b/test-data/unit/check-kwargs.test
@@ -299,8 +299,11 @@ d = None # type: Dict[str, A]
 f(**d)
 f(x=A(), **d)
 d2 = None # type: Dict[str, B]
-f(**d2)        # E: Argument 1 to "f" has incompatible type "**Dict[str, B]"; expected "A"
-f(x=A(), **d2) # E: Argument 2 to "f" has incompatible type "**Dict[str, B]"; expected "A"
+# TODO: "too many arguments" errors will not appear after #9629 is merged
+f(**d2)        # E: Too many arguments for "f" \
+               # E: Argument 1 in call to "f" cannot unpack into any parameters; no parameter accepts type "Dict[str, B]"
+f(x=A(), **d2) # E: Too many arguments for "f" \
+               # E: Argument 2 in call to "f" cannot unpack into any parameters; no parameter accepts type "Dict[str, B]"
 class A: pass
 class B: pass
 [builtins fixtures/dict.pyi]
@@ -363,7 +366,7 @@ def f(a: 'A', b: 'B') -> None: pass
 d = None # type: Dict[str, Any]
 f(**d)
 d2 = None # type: Dict[str, A]
-f(**d2) # E: Argument 1 to "f" has incompatible type "**Dict[str, A]"; expected "B"
+f(**d2) # E: Too few arguments for "f"
 class A: pass
 class B: pass
 [builtins fixtures/dict.pyi]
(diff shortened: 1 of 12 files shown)


esoma:
Continuing a line of conversation from #9676:

@hauntsaninja

>In fact I'd go a little further than your example and say errors in the following cases are valuable:
> ```python
> f(a=A(), **ad)
> f(a=A(), **ad, **bd)
> ```
Agreed, that was actually the first rule I tried. I changed it at some point, forgot I changed it and now I can't remember why 🙃. In any case, I've reverted it to have this behavior in my branch and its behaving well.

> But we might be in the minority here (or at least, on the wrong side of the do-ocracy). It looks like mypy changed its behaviour for the *args case in #7392 and consistency between args and kwargs seems like a good thing.

Right, I'm not sure I'm convinced of the need for that change, but I can understand it and it's not my position to be convinced anyway. I've thought about this a bit more and I've probably misinterpreted why @JukkaL reasons Case A should be an error and Case B/C should not. Let's say the following:
```python
d: Dict[str, str] = {}

def f0() -> None: pass
f0(**d) # no error

def f1(x: int = 0) -> None: pass
f1(**d) # no error

def f2(x: str = '') -> None: pass
f2(**d) # no error

def f3(x: str = '', **kwargs: int) -> None: pass
f3(**d) # no error

def f4(x: str = '', **kwargs: int) -> None: pass
f4(x='foo', **d) # error

def f5(x: int = 0, **kwargs: int) -> None: pass
f5(**d) # error
```
The rule here is that `d` can always be empty. However! If `d` is not compatible with any parameters **and** there is a `**kwargs` parameter it is an error. Since `**kwargs` will always consume the rest of the remaining arguments we can say that if `d` doesn't map anywhere else then the user *probably* means for it to be compatible with `**kwargs`.

I believe this would allow for Case A to raise an error, but B and C to not. This seems goofy to me, but I think it makes some practical sense, it aligns better with the `*args` behavior and doesn't step on #9269. If this is the way to go I'm happy to make the change.

test-data/unit/check-kwargs.test
```diff
@@ -485,7 +562,7 @@ def g(arg: int = 0, **kwargs: object) -> None:
 
 d = {} # type: Dict[str, object]
 f(**d)
-g(**d)  # E: Argument 1 to "g" has incompatible type "**Dict[str, object]"; expected "int"
+g(**d)
 
 m = {} # type: Mapping[str, object]
 f(**m)
```
esoma, line 568: On the line below there is a `# TODO: should be an error`
Presumably it shouldn't be anymore?

PR #9708, opened 2020-11-09
Fix exception using TypeVars with Generic

### LINKED ISSUE #9448: Internal error when handling generic class
I have some code which includes these lines:
```
_From = TypeVar("_From", date, None)
_Until = TypeVar("_Until", date, None)

class TimePeriod(ClauseElement, Generic[_From, _Until]):
    @classmethod
    def day(cls, day: date):
        return cls(day, day)  # Expected None?

    def __init__(self, from_day: _From = None, until_day: _Until = None, column="date"):
        if from_day and until_day and from_day > until_day:  # Right operand is never evaluated?
```

It's giving me several errors that don't make sense to me, plus hitting an internal error as well:
```
logic/analytics/time_period.py:38: error: Argument 1 to "TimePeriod" has incompatible type "date"; expected "None"  [arg-type]
logic/analytics/time_period.py:38: error: Argument 2 to "TimePeriod" has incompatible type "date"; expected "None"  [arg-type]
logic/analytics/time_period.py:47: error: Right operand of 'and' is never evaluated  [unreachable]
logic/analytics/time_period.py:56: error: Right operand of 'and' is never evaluated  [unreachable]
logic/analytics/time_period.py:51: error: INTERNAL ERROR -- Please try using mypy master on Github:
https://mypy.rtfd.io/en/latest/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues
version: 0.782
  File "mypy/checker.py", line 401, in accept
  File "mypy/nodes.py", line 1062, in accept
  File "mypy/checker.py", line 2011, in visit_assignment_stmt
  File "mypy/checker.py", line 2150, in check_assignment
  File "mypy/checker.py", line 2821, in infer_variable_type
  File "mypy/sametypes.py", line 31, in is_same_type
  File "mypy/types.py", line 1786, in accept
  File "mypy/sametypes.py", line 162, in visit_partial_type
RuntimeError:
logic/analytics/time_period.py:51: : note: use --pdb to drop into pdb
```

Why does it seem to think the value will always be None? Am I misunderstanding TypeVar?

The end goal is to have some methods like:
`def foo(self: TimePeriod[date, date]):`
which can only be used if the class was initialised without None.

Dreamsorcerer:
Fixes #9448 

This is problably not the right way to fix this, as I'm not entirely sure what I'm editing, however, it does pass all the tests.

The problem occurs in `infer_variable_type()` under specific circumstances:
https://github.com/python/mypy/blob/master/mypy/checker.py#L2859

The issue only occurs when a `TypeVar` contains `None` _and_ appears second in the `Generic`. i.e. There is no error if you change the test example to `Generic[S, T]`.

This seems to be down to the order in which it evaluates the types, so `self.attr` will be evaluated 4 times as `int`, `None`, `int`, `None` in the test, but switching the order around will evaluate it as `int`, `int`, `None`, `None`.

When it encounters the `None` value, it updates the value in `self.inferred_attribute_types` from `builtins.int*` to `<partial None>` (which happens in the call to `self.infer_partial_type()`). When this is then followed by a subsequent check for `int`, the code hits the `RuntimeError` while calling `is_same_type()`.

When changing `None` to other types (e.g. `str` or `bool`), I noticed that the value is never changed from `builtins.int*`, so my proposed fix is to avoid changing the value entirely if it already exists.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -2897,10 +2897,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                 return False
         else:
             return False
-        if (not isinstance(lvalue, MemberExpr)
-            or lvalue.def_var not in self.inferred_attribute_types):
-            self.set_inferred_type(name, lvalue, partial_type)
-            self.partial_types[-1].map[name] = lvalue
+        self.set_inferred_type(name, lvalue, partial_type)
+        self.partial_types[-1].map[name] = lvalue
         return True
 
     def is_valid_defaultdict_partial_value_type(self, t: ProperType) -> bool:
@@ -2938,7 +2936,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             var.is_inferred = True
             if isinstance(lvalue, MemberExpr) and self.inferred_attribute_types is not None:
                 # Store inferred attribute type so that we can check consistency afterwards.
-                if lvalue.def_var is not None:
+                if (lvalue.def_var is not None
+                    and lvalue.def_var not in self.inferred_attribute_types):
                     self.inferred_attribute_types[lvalue.def_var] = type
             self.store_type(lvalue, type)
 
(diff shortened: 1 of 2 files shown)


hauntsaninja:
#9448 was fixed on master

PR #9713, opened 2020-11-11
[mypyc] Faster min max

mehdigmira:
### Description

WIP

## Test Plan

WIP


mehdigmira:
This is a draft for making min/max operations faster (https://github.com/mypyc/mypyc/issues/773). For now the only specialized version is min(int, int).

```
running min_max_pair
..........
interpreted: 0.580921s (avg of 5 iterations; stdev 1.7%)
compiled:    0.346267s (avg of 5 iterations; stdev 1.3%)

compiled is 1.678x faster
```
I have some questions:
- Do I need to support min(*args: int) and min(x: Iterable[int]) ? If so, i think i should implement a C function that takes an array of int as an input. Is this supported with c_function_op ? WOuld smth like this make sense ?
```
int_min = c_function_op(
    name="builtins.min",
    arg_types=[List[int_rprimitive]]],
    return_type=int_rprimitive,
    c_function_name='Py_MIN',
    error_kind=ERR_NEVER)
```
- Do I need to support all types of int ? (btw what is CPyTagged ?)
- Do I need to support float as well ?
- The python documentation doesn't specify argument types for Py_MIN (https://docs.python.org/3/c-api/intro.html#c.Py_MIN). I tried building a generic version
```
generic_min = c_function_op(
    name="builtins.min",
    arg_types=[object_rprimitive, object_rprimitive],
    return_type=object_rprimitive,
    c_function_name='Py_MIN',
    error_kind=ERR_NEVER)
```
My test case passes but compiling fails with `subprocess.CalledProcessError: Command '['python3', '-c', 'import microbenchmarks.builtins; import benchmarking as bm; print("\\nelapsed:", bm.run_once("min_max_pair"))']' died with <Signals.SIGSEGV: 11>.` why is that ?

mypyc/primitives/int_ops.py
```diff
@@ -148,6 +148,14 @@ def int_unary_op(name: str, c_function_name: str) -> CFunctionDescription:
     c_function_name='CPyTagged_IsLt_',
     error_kind=ERR_NEVER)
 
+int_min = c_function_op(
+    name="builtins.min",
+    arg_types=[int_rprimitive, int_rprimitive],
+    return_type=int_rprimitive,
+    c_function_name='Py_MIN',
```
JukkaL, line 155: `int_rprimitive` is actually `CPyTagged` in the generated C, i.e. a tagged pointer. `Py_MIN` only works with plain integers, so it only works for a subset of possible values. You should test also with a big integer, such as a value higher than 2**63.

There are two ways around this:

* Add a function `CPyTagged_Min(x, y)` to `mypyc/lib-rt/int_ops.c` that uses tagged integer operations. This is the easiest option but only works with integers. We'd need separate functions for all supported argument types.
* Inline the implementation of `min` by generating IR operations in `specialize.py`. You can use the `binary_op` method, for example, so that this works for arbitrary types.

mypyc/irbuild/specialize.py
```diff
@@ -256,3 +258,12 @@ def translate_isinstance(builder: IRBuilder, expr: CallExpr, callee: RefExpr) ->
         if irs is not None:
             return builder.builder.isinstance_helper(builder.accept(expr.args[0]), irs, expr.line)
     return None
+
+
+@specialize_function('builtins.min')
+def faster_min(builder: IRBuilder, expr: CallExpr, callee: RefExpr):
```
JukkaL, line 264: You only need to use `@specialize_function` if you inline the body of `min` in the generated IR (see below). Otherwise using `c_function_op` should be sufficient.

mypyc/irbuild/specialize.py
```diff
@@ -36,20 +37,21 @@
...
 # compiled, and the RefExpr that is the left hand side of the call.
+
 Specializer = Callable[['IRBuilder', CallExpr, RefExpr], Optional[Value]]
 
 # Dictionary containing all configured specializers.
 #
 # Specializers can operate on methods as well, and are keyed on the
 # name and RType in that case.
-specializers = {}  # type: Dict[Tuple[str, Optional[RType]], Specializer]
+specializers = defaultdict(list)  # type: DefaultDict[Tuple[str, Optional[RType]], List[Specializer]]
```
JukkaL, line 47: I think that this change is not needed.

TH3CHARLie: I think the `defaultdict` is not needed while making the candidates specializers a list is necessary, otherwise in this case, `translate_safe_generator_call` and `faster_min` would compete with each other.

TH3CHARLie:
> * Do I need to support min(*args: int) and min(x: Iterable[int]) ? If so, i think i should implement a C function that takes an array of int as an input. Is this supported with c_function_op ? WOuld smth like this make sense ?
> 
To support op that takes variable-length args, you can take a look at `new_tuple_op` and pay attention to the `var_arg_type` attribute.

> * Do I need to support all types of int ? (btw what is CPyTagged ?)
> 
mypyc has short_int_rprimitive/int_rprimitve that maps to the Python int, while int32/int64 are only low-level concepts.
As the name suggests, short integer stores integer within a range that can directly be represented in binary. The larger ones, on the other hand, are represented as a pointer to `PyObject`, therefore `CPyTagged` is a tagged pointer to represent both at the same time.

> * Do I need to support float as well ?
>
For this PR, I think it's completely OK to support int and it should have a good performance boost, you can try float min/max later on.

JukkaL:
> Do I need to support min(*args: int)

The two-argument version is sufficient, since it's by far the most common one. If you want to support this, it's better to do it in a separate PR and add a different specializer for it -- but I wouldn't bother, since more than two arguments are rare.

>  If so, i think i should implement a C function that takes an array of int as an input. Is this supported with c_function_op ? WOuld smth like this make sense ?

You can use the `var_arg_type` argument of `c_function_op` (there are some examples in existing primitive definitions).

> Do I need to support all types of int ? (btw what is CPyTagged ?)

Only `int_rprimitive` (CPyTagged) needs to be supported, but as I explained in another comment, you can make this work for arbitrary types by generating the body of `min` as IR. I can give more pointers if you want to give this a try.

CPyTagged is either a C integer value shifted left by 1 or a tagged pointer to a CPython `int` object, depending on the value of the lowest bit. Hmm it looks like this isn't properly documented. I'll add this to the developer docs.

> Do I need to support float as well ?

If you generate IR, float will be trivial to support. Otherwise, it would be nice but not essential.

> The python documentation doesn't specify argument types for Py_MIN

As I mentioned above, `Py_MIN` doesn't work with `CPyTagged` values.

97littleleaf11:
Close this one since https://github.com/python/mypy/pull/10265 is going to fix the issue

PR #9714, opened 2020-11-11
[mypyc] Small improvements to developer docs

JukkaL:
Document tagged integers and some other updates.

### DIFF
--- a/mypyc/doc/dev-intro.md
+++ b/mypyc/doc/dev-intro.md
@@ -231,11 +231,25 @@ Installing a released version of mypy using `pip` (which is compiled)
 and using `dmypy` (mypy daemon) is a much, much faster way to type
 check mypyc during development.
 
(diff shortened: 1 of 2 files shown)


TH3CHARLie (approved):
Looks great! Thanks for the useful updates! I am thinking how about adding a pointer to mypyc dev docs to the `Quick start for contributing to mypy` section in the main README.md

JukkaL:
> I am thinking how about adding a pointer to mypyc dev docs to the Quick start for contributing to mypy section in the main README.md

Yeah, we should clearly have something in there. I've been slow to do this since I've been waiting until other docs are in decent shape. I'll do another pass over the user documentation soon, and then we should be ready to have a landing page with an intro and links to docs, contributor guide and benchmark results. We also need to publish the user docs.

PR #9715, opened 2020-11-11
Remove outdated paragraph about dmypy not supporting follow-imports=normal

matangover:
#8930 presumably forgot to remove this paragraph at the bottom of the page.


JukkaL (approved):
Yes, this is no longer relevant. Thanks!

PR #9723, opened 2020-11-14
[mypyc] Make intro to mypyc punchier

JukkaL:
This is a major update of the introduction section in the mypyc docs. Shorten
it and update the order of subsections to form a better narrative.


TH3CHARLie (approved):
Looks great! Thanks for the detailed improvement!

PR #9725, opened 2020-11-15
[mypyc] Remove c_ prefix in primitive registry

TH3CHARLie:
relates https://github.com/mypyc/mypyc/issues/781

remove the `c_` prefix since we no longer have the old registry.


JukkaL (approved):
Looks good!

PR #9728, opened 2020-11-17
build: disallow ignoring blocking errors

### LINKED ISSUE #9727: Error code misc and build errors
It probably shouldn't be possible to "disable" build errors at all, let alone under an overloaded and innocuous looking error code (note the new, clearer summary for blocking errors is unreleased / from https://github.com/python/mypy/pull/9674).

```
λ mkdir -p package/a  
λ mkdir -p package/b
λ echo 'x: int = "asdf"' >> package/b/x.py
λ echo 'y: str = 0' >> package/a/x.py
λ echo 'another_error: int = "adsf"' > package/unrelated.py

λ mypy package --show-error-codes                          
package/b/x.py: error: Duplicate module named 'x' (also at 'package/a/x.py')  [misc]

λ mypy package --show-error-codes --disable-error-code misc
Success: no issues found in 3 source files
```

Thanks to @ehossack for mentioning that people are doing this.

hauntsaninja:
Fixes #9727 (and effectively also adds a test for #9674)

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2748,7 +2748,8 @@ def load_graph(sources: List[BuildSource], manager: BuildManager,
             manager.errors.set_file(st.xpath, st.id)
             manager.errors.report(
                 -1, -1,
-                "Duplicate module named '%s' (also at '%s')" % (st.id, graph[st.id].xpath)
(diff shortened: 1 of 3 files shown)


JukkaL (approved):
Thanks for the fix! Looks good. This was potentially a pretty bad issue.

PR #9742, opened 2020-11-22
mypy: support namespace packages when passing files

### LINKED ISSUE #5759: Support namespace packages from command line

### LINKED ISSUE #8726: Relative imports from within the same package lead to "no type hints or stubs" error
Hi,

I have a problem with mypy and handling of imports. Basically, I have a Python package with several modules which import other modules of the package. When I run `mypy` (version 0.770) on the files, I get this error:
`error: Skipping analyzing '.file2': found module but no type hints or library stubs`

However, I don't understand that error message because (in this example) `file2` does contain type hints. 
For the purpose of this issue, I have created a minimal reproducer. I have two files `file1` and `file2` which both define a class. The class in `file1` (`Example1`) has a method which uses the class from `file2` (`Example2`), thus `file1` imports `file2` with a relative import (`from .file2 import Example2`). This is the full code:

- `file1.py`
```python
from .file2 import Example2

class Example1:
    def __init__(self, value: float) -> None:
        self.value = value

    def add(self, other: Example2) -> None:
        self.value += other.value
```
- `file2.py`
```python
class Example2:
    def __init__(self, value: float) -> None:
        self.value = value
```

When I run `mypy file1.py file2.py`, I am receiving the error mentioned above:
```
file1.py:1: error: Skipping analyzing '.file2': found module but no type hints or library stubs
file1.py:1: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```
Now I did visit that recommended link in the documentation but I couldn't figure out how to solve my problem. 

However, one thing that worked was to turn the relative into a global import, i.e. to write `from file2 import Example2`. However, I would like to avoid that because in my package I would prefer to use relative imports for ease of readability of the code (amongst others). Also, the error message does not seem to make sense to me because `file2.py` does contain type annotations.

Apologies for posting this because it seems like a basic thing, but I couldn't figure it out with the documentation or by browsing the issues here. 
It would be great if someone could explain how to do this properly or point me to a documentation of the solution.
Thanks a lot.

(Python version if 3.8.2).

### LINKED ISSUE #9677: Running mypy on the current directory without some OS-specific stuff
**Feature**

Have a way to pass the current directory to `mypy`. I can run `mypy dir`, but `mypy .` fails because `.` is not a valid Python package name.

**Pitch**

Would be very useful to have generic CI commands. I'm currently figuring out how I can pass the full path to the current directory in an OS-dependent manner. Other tools (flake8, black, ...) have no issue with "."

hauntsaninja:
This is the successor to https://github.com/python/mypy/pull/9632. Things should basically be as discussed in that PR. Since #9616 is merged, this should now resolve #5759.

We leave the Bazel integration with `--package-root` almost entirely untouched, save for a) one change that's a bugfix / doesn't affect the core of what `--package-root` is doing, b) another drive by bugfix that's not related to this PR.
Change a) fixes the package root `__init__.py` hackery when passed absolute paths. Change b) fixes the validation logic for package roots above the current directory; it was broken if you passed `..` as a package root

Since we're leaving `--package-root` alone for now, I named the new flag `--explicit-package-base` to try and avoid confusion. Doing so also matches the language used by BuildSource a little better.

The new logic is summarised in the docstring of `SourceFinder.crawl_up`.

Some commentary:
- I change `find_sources_in_dir ` to call `crawl_up` directly to construct the BuildSource. This helps codify the fact that files discovered will use the same module names as if you passed them directly.
- Doing so keeps things DRY with the more complicated logic and means, for instance, that we now do more sensible things in some cases when we recursively explore directories that have invalid package names.
- Speaking of invalid package names, if we encounter a directory name with an invalid package name, we stop crawling. This is necessary because with namespace packages there's no guarantee that what we're crawling was meant to be a Python package. I add back in a check in the presence of `__init__.py` to preserve current unit tests where we raise InvalidSourceList.
- The changes to modulefinder are purely cosmetic and can be ignored (there's some similar logic between the two files and this just makes sure they mirror each other closely)
- One notable change is we now always use absolute paths to crawl. This makes the behaviour more predictable and addresses a common complaint: fixes #9677, fixes #8726 and others.
- I figured this could use more extensive testing than a couple slow cmdline tests. Hopefully this test setup also helps clarify the behaviour :-)

cc @JukkaL

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2396,7 +2396,7 @@ def find_module_and_diagnose(manager: BuildManager,
                 and not options.use_builtins_fixtures
                 and not options.custom_typeshed_dir):
             raise CompileError([
-                'mypy: "%s" shadows library module "%s"' % (result, id),
+                'mypy: "%s" shadows library module "%s"' % (os.path.relpath(result), id),
                 'note: A user-defined top-level module with name "%s" is not supported' % id
             ])
         return (result, follow_imports)
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -50,7 +50,7 @@ def create_source_list(paths: Sequence[str], options: Options,
 def keyfunc(name: str) -> Tuple[bool, int, str]:
     """Determines sort order for directory listing.
 
-    The desirable propertes are:
+    The desirable properties are:
     1) foo < foo.pyi < foo.py
     2) __init__.py[i] < foo
     """
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -71,6 +71,15 @@ def normalise_package_base(root: str) -> str:
 
 
 def get_explicit_package_bases(options: Options) -> Optional[List[str]]:
+    """Returns explicit package bases to use if the option is enabled, or None if disabled.
+
+    We currently use MYPYPATH and the current directory as the package bases. In the future,
+    when --namespace-packages is the default could also use the values passed with the
+    --package-root flag, see #9632.
+
+    Values returned are normalised so we can use simple string comparisons in
+    SourceFinder.is_explicit_package_base
+    """
     if not options.explicit_package_bases:
         return None
     roots = mypy_path() + options.mypy_path + [os.getcwd()]
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -64,7 +64,7 @@ def keyfunc(name: str) -> Tuple[bool, int, str]:
 def normalise_package_base(root: str) -> str:
     if not root:
         root = os.curdir
-    root = os.path.normpath(os.path.abspath(root))
+    root = os.path.abspath(root)
     if root.endswith(os.sep):
         root = root[:-1]
     return root
@@ -127,7 +127,7 @@ class SourceFinder:
         We won't crawl past directories with invalid package names.
         The base directory returned is an absolute path.
         """
-        path = os.path.normpath(os.path.abspath(path))
+        path = os.path.abspath(path)
         parent, filename = os.path.split(path)
 
         module_name = strip_py(filename) or filename
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -47,16 +47,18 @@ def create_source_list(paths: Sequence[str], options: Options,
     return sources
 
 
-def keyfunc(name: str) -> Tuple[int, str]:
+def keyfunc(name: str) -> Tuple[bool, int, str]:
     """Determines sort order for directory listing.
 
-    The desirable property is foo < foo.pyi < foo.py.
+    The desirable propertes are:
+    1) foo < foo.pyi < foo.py
+    2) __init__.py[i] < foo
     """
     base, suffix = os.path.splitext(name)
     for i, ext in enumerate(PY_EXTENSIONS):
         if suffix == ext:
-            return (i, base)
-    return (-1, name)
+            return (base != "__init__", i, base)
+    return (base != "__init__", -1, name)
 
 
 def normalise_package_base(root: str) -> str:
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -1,11 +1,12 @@
 """Routines for finding the sources that mypy will check"""
 
-import os.path
+import functools
+import os
 
-from typing import List, Sequence, Set, Tuple, Optional, Dict
+from typing import List, Sequence, Set, Tuple, Optional
 from typing_extensions import Final
 
-from mypy.modulefinder import BuildSource, PYTHON_EXTENSIONS
+from mypy.modulefinder import BuildSource, PYTHON_EXTENSIONS, mypy_path
 from mypy.fscache import FileSystemCache
 from mypy.options import Options
 
@@ -24,7 +25,7 @@ def create_source_list(paths: Sequence[str], options: Options,
     Raises InvalidSourceList on errors.
     """
     fscache = fscache or FileSystemCache()
-    finder = SourceFinder(fscache)
+    finder = SourceFinder(fscache, options)
 
     sources = []
     for path in paths:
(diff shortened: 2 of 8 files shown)


hauntsaninja:
Summary of the mypy_primer issues:
1) While discovering files recursively, it turned out it's important that we list `__init__.py`s first. I haven't quite been able to get a minimal repro though, so no regression test :-( The real world code it's triggering on is often doing stuff like `from .x import x`, though.
2) I made a defensive change to prevent passing around invalid module names for cases like `pkg.invalid-name` if we have `pkg/__init__.py` and `pkg/invalid-name.py`. But this adds `pkg` as a base_dir, causing search paths to get muddied. Secondarily, it causes duplicate module errors if you have similarly named scripts (like zulip's migration scripts)

JukkaL (commented):
This looks awesome! I especially like the unit tests -- they both test the behavior in lot of detail and in a clean way, and also act as a nice specification. They are also written very clearly. I really enjoyed reviewing this.

Left some minor comments. The only thing that may be significant is the behavior with packages spanning multiple directories. Feel free to merge once you've addressed all the comments.

hauntsaninja:
Thanks for the review! Implemented all the suggestions. I still owe documentation updates, but I'll do that in a separate PR.

niander:
@hauntsaninja
I am trying to avoid having to specify explicit base folders for all the packages we have in our repo. According to the docstring in `craw_up`:

> If namespace packages is on, we crawl upwards until the nearest explicit base directory.
> Failing that, we return one past the highest directory containing an __init__.py

So that means that in a scenario where there is a namespace package at the root, to be able to correctly run `mypy` passing a file in the command line, it is required to explicitly specify all base directories in `mypy_path` and enable `--explicit-package-bases`. This seems odd to me because that means that `--namespace-packages` alone won't work in the common scenario of having a single top level namespace package unless you explicitly specify base directories.

Why was `crawl_up` implemented like that? While I trust there is a very good reason for it to be implemented the way it was. Maybe because it is impossible to know the namespace top folder because it could be other folders in the file system. I am particularly curious because I remember that before the introduction of `--explicit-package-bases` it was working for us with only `--namespace-packages` enabled. Or, maybe, it wasn't and I am totally wrong, but the curiosity is still there =)

hauntsaninja:
You might be confusing the directory the package is in with the package directory? Also mypy automatically adds the current directory as an explicit package base, which might be another piece you're missing. Automatically treating cwd as a base is why the common single top level namespace package scenario works.

Please read https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules which has examples.

The behaviour was previously completely broken when passing `--namespace-packages` and files. And the changes in behaviour here are backward compatible. You probably weren't checking anything (or are "totally wrong" in some other regard ;-) )

If you're still stuck, I recommend posting on gitter.im or making an issue.

niander:
@hauntsaninja thanks for the response. I am not stuck. It just seems that there is no way out of needing to define the base folder for all projects in the repo, which are using the same mypy.ini.

I don't think I am confusing directory and package. Yes, there is a directory that includes the packages and other files like setup.py. This is the base directory. And, because the top level package is a namespace package, the first folder inside it doesn't have `__init__.py` but the other subdirectories have. With just `--namespace-packages` and passing a file to check, `mypy` will consider the base directory as the namespace package folder and not the folder containing the namespace package. Which is pretty much what also happens when `--namespace-packages` is disabled.

Yes, maybe it was working before because the cwd was always the directory containing the package. But now, that is not true, especially since we mostly run mypy from the Python extension in VSCode and the cloned repository has several other python projects.

Anyways, I wonder, for instance, how pylint discover the correct base folder and the fully qualified name for the packages when checking a single file and if mypy could do the same.

hauntsaninja:
From what you're describing, it sounds like using `mypy --namespace-packages --explicit-package-bases` without any listing should work. I recommend creating an issue that has a `tree`-like description of your project layout.

The situation in which you'd need to list things is if you had a layout with something like:
```
root
    - projfolder1
        - setup.py
        - namespace_package1
    - projfolder2
        - setup.py
        - namespace_package2
```

niander:
That is pretty much what we have. When I run `mypy --namespace-packages --explicit-package-bases` for a file, lets say, `projfolder1/namespace_package1/package1/mymodule.py` and cwd being `root`, with `--verbose` I can see that the `base_dir` inferred is `root/projfolder1/namespace_package1`. And mypy finishes with success without detecting any error. With explicit `mypy_path`, the `base_dir` would be `root/projfolder1` and that works.

Do you think this is a bug worthy of an issue?

hauntsaninja:
I recommend creating an issue that has a `tree`-like description of your actual project layout (or a minimum repro) and the output with `--verbose`. Even better if your code is open source :-)

jaklan:
@niander have you resolved your issue? I have exactly the same problem in monorepo setup when using `pre-commit` (so passing file paths to mypy and cwd in `root`):
```
root/
-- projects/
---- project_a/
------ src/
-------- namespace/
----------- subnamespace/
------------- utils/
--------------- __init__.py
--------------- main.py
---- project_b/
------ src/
-------- namespace/
----------- other_subnamespace/
------------- utils/
--------------- __init__.py
--------------- main.py
```

will fail due to `error: Duplicate module named "utils.main"`.  

@hauntsaninja `--explicit-package-bases` doesn't help here, because then we get:
`module='projects.project_a.src.namespace.subnamespace.utils.main'`
instead of:
`module='namespace.subnamespace.utils.main'`
 and I don't see an easy way to modify `MYPYPATH` to include all needed paths automatically as well...

hauntsaninja:
As I said in my previous comment, this PR is not the place to discuss issues. I'll be locking this thread.

`--explicit-package-bases` plus setting `MYPYPATH=projects/project_a/src:projects/project_b/src` should do the trick. Feel free to write a wrapper script around mypy that computes the path appropriately, but I don't see a way for mypy to divine your project layout. If you have a recommendation, please open a feature request issue.

PR #9753, opened 2020-11-24
Remove outdated restriction about PyAnnotate

jugmac00:
PyAnnotate also supports type annotations, not only type comments.


hauntsaninja (approved):
Thanks! Looks like the lint failure is caused by a new release of flake8-bugbear

hauntsaninja:
Thanks again! (Fixed the lint in #9754)

PR #9757, opened 2020-11-25
TypedDict: Fix casefolding of 'missing keys' error message

intgr:
### Description

Previously the message accidentally converted all identifiers to
lowercase. By rephrasing the message, I avoided complicating the
`format_key_list()` function.

## Test Plan
Added a test case as well.

Code:
```python

Foo = TypedDict('Foo', {'camelCaseKey': str})
value: Foo = {}
```
Old output:
```
casetest.py:4: error: Key 'camelcasekey' missing for TypedDict "Foo"
```
Output after my changes:
```
casetest.py:4: error: Missing key 'camelCaseKey' for TypedDict "Foo"
```

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -1118,8 +1118,8 @@ class MessageBuilder:
             if actual_set < expected_set:
                 # Use list comprehension instead of set operations to preserve order.
                 missing = [key for key in expected_keys if key not in actual_set]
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):

hauntsaninja:
Thank you!

PR #9762, opened 2020-11-27
docs: miscellaneous improvements

hauntsaninja:
I was going through the docs to figure out where to document PEP 585 (and PEP 563 and 604 and so on), and here's a grab bag of small edits I made as I did so.


docs/source/builtin_types.rst
```diff
@@ -31,10 +31,10 @@ strings and ``Dict[Any, Any]`` is a dictionary of dynamically typed
...
 ``List`` are aliases for the built-ins ``dict`` and ``list``, respectively.
 
-``Iterable``, ``Sequence``, and ``Mapping`` are generic types that
-correspond to Python protocols. For example, a ``str`` object or a
-``List[str]`` object is valid
-when ``Iterable[str]`` or ``Sequence[str]`` is expected. Note that even though
-they are similar to abstract base classes defined in :py:mod:`collections.abc`
-(formerly ``collections``), they are not identical, since the built-in
-collection type objects do not support indexing.
+``Iterable``, ``Sequence``, and ``Mapping`` are generic types that correspond to
```
JelleZijlstra, line 34: Is this just realigning the lines? Not sure that's useful because the rendered output is the same and it makes the history harder to follow.

docs/source/existing_code.rst
```diff
@@ -114,8 +114,8 @@ A simple CI script could look something like this:
 
 .. code-block:: text
 
-    python3 -m pip install mypy==0.600  # Pinned version avoids surprises
-    scripts/mypy  # Runs with the correct options
+    python3 -m pip install mypy==0.790  # Pinned version avoids surprises
```
JelleZijlstra, line 117: Would be nice if we can automatically render the most recent release here, but probably not worth the effort

docs/source/python36.rst
```diff
@@ -1,67 +0,0 @@
-.. _python-36:
```
JelleZijlstra, line 1: Why was this deleted?

JelleZijlstra (changes requested):
Also you misspelled "miscellaneous" :)

docs/source/builtin_types.rst
hauntsaninja: I rephrase to mention Python 3.9, but yeah the top is just realignment. I'll make my editor do less next time.

docs/source/python36.rst
hauntsaninja: I figured a "what's new in Python 3.6" page doesn't make sense when Python 3.6 is the oldest supported Python and we don't have similar pages for any other Python release. Everything on this page is covered in another section already.

JelleZijlstra (approved):

PR #9763, opened 2020-11-28
Document PEP 585, 563, 604 and more.

### LINKED ISSUE #8523: Error when preserving generic type behaviour in subclasses (when not supported at runtime)

### LINKED ISSUE #8629: Document how to import types from stubs
I'm currently trying to type an API using the `signal` module. At some point, we pass a callback that has the type `signal._Handler` (see: https://github.com/python/typeshed/blob/master/stdlib/3/signal.pyi#L73), which is not in the  `signal` module, but in typeshed.

If I try to import it, I get an error, and I could not find the proper way to do so in the documentation.

hauntsaninja:
Fixes #8629, fixes #8523

This creates a new page to document issues arising from discrepancies
between the runtime and annotations. I felt this was better, rather than
force-fitting things into existing pages and "common issues", for
instance, it prevents us from having to explain PEP 563 in several
different places.

I do still list the runtime errors you'd get in the "common issues" page
to preserve SEO :-)

"String literal types", "Class name forward references", and "Import
cycles" are basically the same as where they were copied over from.

This also factors out the documentation of PEP 604 that I promised when
merging that PR (it seemed pretty verbose, particularly for the "kinds
of types" page). It's also a good place to document PEP 613, when we get
around to supporting that.

### DIFF
--- a/docs/source/builtin_types.rst
+++ b/docs/source/builtin_types.rst
@@ -38,3 +38,8 @@ though they are similar to abstract base classes defined in
 :py:mod:`collections.abc` (formerly ``collections``), they are not identical. In
 particular, prior to Python 3.9, the built-in collection type objects do not
 support indexing.
+
+In Python 3.9 and later, built-in collection type objects support indexing. This
+means that you can use built-in classes or those from :py:mod:`collections.abc`
+instead of importing from :py:mod:`typing`. See :ref:`generic-builtins` for more
+details.
--- a/docs/source/runtime_troubles.rst
+++ b/docs/source/runtime_troubles.rst
@@ -189,7 +189,7 @@ Using classes that are generic in stubs but not at runtime
 Some classes are declared as :ref:`generic<generic-classes>` in stubs, but not
 at runtime.
 
-In Python 3.8 and lower, there are several examples within the standard library,
+In Python 3.8 and earlier, there are several examples within the standard library,
 for instance, :py:class:`os.PathLike` and :py:class:`queue.Queue`. Subscripting
 such a class will result in a runtime error:
 
@@ -268,7 +268,7 @@ Using generic builtins
 Starting with Python 3.9 (:pep:`585`), the type objects of many collections in
 the standard library support subscription at runtime. This means that you no
 longer have to import the equivalents from :py:mod:`typing`; you can simply use
-the built-in collections or those from :py:mod:`collections.abc`
+the built-in collections or those from :py:mod:`collections.abc`:
 
 .. code-block:: python
 
@@ -305,7 +305,7 @@ module in earlier versions of Python than the addition, for example, using any
 of ``Literal``, ``Protocol``, ``TypedDict`` with Python 3.6.
 
 The easiest way to do this is to install and use the ``typing_extensions``
-package for the relevant imports, for example:
+package from PyPI for the relevant imports, for example:
 
 .. code-block:: python
 
--- a/docs/source/runtime_troubles.rst
+++ b/docs/source/runtime_troubles.rst
@@ -244,6 +244,9 @@ If your subclass is also generic, you can use the following:
    task_queue: MyQueue[str]
    reveal_type(task_queue.get())  # Reveals str
 
+In Python 3.9, we can just inherit directly from ``Queue[str]`` or ``Queue[T]``
+since its :py:class:`queue.Queue` implements :py:meth:`__class_getitem__`, so
+the class object can be subscripted at runtime without issue.
 
 Using types defined in stubs but not at runtime
 -----------------------------------------------
--- a/docs/source/runtime_troubles.rst
+++ b/docs/source/runtime_troubles.rst
@@ -111,8 +111,8 @@ defined (aka forward reference). Thus this code does not work as expected:
    def f(x: A) -> None: ...  # NameError: name 'A' is not defined
    class A: ...
 
-Starting from Python 3.7, you can add the special import ``from __future__ import annotations``,
-which makes the use of string literals in annotations unnecessary:
+Starting from Python 3.7, you can add ``from __future__ import annotations`` to
+resolve this, as discussed earlier:
 
 .. code-block:: python
 
@@ -133,9 +133,9 @@ For Python 3.6 and below, you can enter the type as a string literal or type com
 
    class A: ...
 
-Of course, instead of using a string literal type or special import, you could move the
-function definition after the class definition. This is not always
-desirable or even possible, though.
+Of course, instead of using future annotations import or string literal types,
+you could move the function definition after the class definition. This is not
+always desirable or even possible, though.
 
 .. _import-cycles:
 
--- a/docs/source/runtime_troubles.rst
+++ b/docs/source/runtime_troubles.rst
@@ -314,3 +314,19 @@ package from PyPI for the relevant imports, for example:
 
    from typing_extensions import Literal
    x: Literal["open", "close"]
+
+If you don't want to rely on ``typing_extensions`` being installed on newer
+Pythons, you could alternatively use:
+
+.. code-block:: python
+
+   import sys
+   if sys.version_info >= (3, 8):
+       from typing import Literal
+   else:
+       from typing_extensions import Literal
+
+   x: Literal["open", "close"]
+
+This plays nicely well with following :pep:`508` dependency specification:
+``typing_extensions; python_version<"3.8"``
--- a/docs/source/runtime_troubles.rst
+++ b/docs/source/runtime_troubles.rst
@@ -69,7 +69,8 @@ required to be valid Python syntax. For more details, see :pep:`563`.
     .. code-block:: python
 
         # base class example
-        class A(Tuple['B', 'C']): ... # OK
+        from __future__ import annotations
+        class A(Tuple['B', 'C']): ... # String literal types needed here
         class B: ...
         class C: ...
 
--- a/docs/source/runtime_troubles.rst
+++ b/docs/source/runtime_troubles.rst
@@ -43,8 +43,8 @@ with import cycles, see :ref:`import-cycles`.)
 
 .. _future-annotations:
 
-Annotations import from future (PEP 563)
-----------------------------------------
+Future annotations import (PEP 563)
+-----------------------------------
 
 Many of the issues described here are caused by Python trying to evaluate
 annotations. From Python 3.10 on, Python will no longer attempt to evaluate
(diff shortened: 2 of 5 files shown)


docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+
+Annotation issues at runtime
+============================
+
+Idiomatic use of type annotations can sometimes run up against what a given
+version of Python considers legal code. This section describes these scenarios
+and explains how to get your code running again. Generally speaking, we have
+three tools at our disposal:
+
+* For Python 3.7 through 3.9, use of ``from __future__ import annotations``
```
JelleZijlstra, line 11: Should string literals be in this list?

docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+a convenient way to tell mypy something without the code being evaluated at
+runtime. This is most useful for resolving :ref:`import cycles <import-cycles>`.
+
+.. note::
+
+   Python 3.5.1 and below don't have :py:data:`~typing.TYPE_CHECKING`. An
+   alternative is to define a constant named ``MYPY`` that has the value
+   ``False`` at runtime. Mypy considers it to be ``True`` when type checking.
+
+Class name forward references
```
JelleZijlstra, line 102: This sections seems repetitive with the previous sections

docs/source/runtime_troubles.rst
hauntsaninja: Yeah, I put string literals in the list because they're the easiest way to deal with issues in 3.6.

docs/source/runtime_troubles.rst
hauntsaninja: Do you have a specific thing we should cut? I do like that this section has somewhat standalone examples; it spells out what fixes should look like for users with this problem.

docs/source/builtin_types.rst
```diff
@@ -38,3 +38,8 @@ though they are similar to abstract base classes defined in
 :py:mod:`collections.abc` (formerly ``collections``), they are not identical. In
 particular, prior to Python 3.9, the built-in collection type objects do not
 support indexing.
+
+In Python 3.9 and later, built-in collection type objects support indexing. This
+means that you can use built-in classes or those from :py:mod:`collections.abc`
+instead of importing from :py:mod:`typing`. See :ref:`generic-builtins` for more
+details.
```
JukkaL, line 45: Since after a while a significant fraction (eventually a majority) of users will be using Python 3.9+, at some point we should make the 3.9+ way of doing things more prominent. Initially we could add another column such as "Type (Python 3.9)", which would have examples like `list[str]`.

No need to do this in this PR, though.

docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+Annotation issues at runtime
+============================
+
+Idiomatic use of type annotations can sometimes run up against what a given
+version of Python considers legal code. This section describes these scenarios
+and explains how to get your code running again. Generally speaking, we have
+three tools at our disposal:
+
+* For Python 3.7 through 3.9, use of ``from __future__ import annotations``
+  (:pep:`563`), made the default in Python 3.10 and later
```
JukkaL, line 12: Nit: Since 3.10 is not out yet, this would be better written as "... will be the default in Python 3.10 and later".

docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+    of forward references or generics in:
+
+    * :ref:`type aliases <type-aliases>`;
+    * :ref:`casts <casts>`;
+    * type definitions (see :py:class:`~typing.TypeVar`, :py:func:`~typing.NewType`, :py:class:`~typing.NamedTuple`);
+    * base classes.
+
+    .. code-block:: python
+
+        # base class example
```
JukkaL, line 71: Maybe add `from __future__ import annotations` to this example to make it explicit that string literals are still needed?

docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+
+Since code inside ``if TYPE_CHECKING:`` is not executed at runtime, it provides
+a convenient way to tell mypy something without the code being evaluated at
+runtime. This is most useful for resolving :ref:`import cycles <import-cycles>`.
+
+.. note::
+
+   Python 3.5.1 and below don't have :py:data:`~typing.TYPE_CHECKING`. An
+   alternative is to define a constant named ``MYPY`` that has the value
+   ``False`` at runtime. Mypy considers it to be ``True`` when type checking.
```
JukkaL, line 100: Maybe add a short example of how this works here?

docs/source/runtime_troubles.rst
JukkaL: Here's one idea:

* Move this section before the other sections dealing with forward references.
* Add a sentence to the end of the section along the lines of "The followings sections explain future annotations and string literal types in more detail.
* Remove/reduce redundant discussions in the other sections, or add something like "As discussed earlier, ...".

docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+   class BarClass:
+       def listifyme(self) -> 'List[BarClass]':
+           return listify(self)
+
+.. _not-generic-runtime:
+
+Using classes that are generic in stubs but not at runtime
+----------------------------------------------------------
+
+Some classes are declared as generic in stubs, but not at runtime.
```
JukkaL, line 188: Maybe add a short note explaining the term "generic class" or add a link to relevant section (we don't do this consistently, but I think that it would be useful to have more cross references like these).

docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+
+In Python 3.8 and lower, there are several examples within the standard library,
```
JukkaL, line 190: Tiny nit: Maybe "and earlier" instead of "lower"?

docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+   _T = TypeVar("_T")
+   if TYPE_CHECKING:
+       class _MyQueueBase(Queue[_T]): pass
+   else:
+       class _MyQueueBase(Generic[_T], Queue): pass
+
+   class MyQueue(_MyQueueBase[_T]): pass
+
+   task_queue: MyQueue[str]
+   reveal_type(task_queue.get())  # Reveals str
```
JukkaL, line 243: Optionally give a short Python 3.9 example, such as `class Tasks(Queue[str]):`.

docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+
+.. _generic-builtins:
+
+Using generic builtins
+----------------------
+
+Starting with Python 3.9 (:pep:`585`), the type objects of many collections in
+the standard library support subscription at runtime. This means that you no
+longer have to import the equivalents from :py:mod:`typing`; you can simply use
+the built-in collections or those from :py:mod:`collections.abc`
```
JukkaL, line 269: Add a colon or a period to the end of the paragraph.

docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+There is limited support for using this syntax in Python 3.7 and later as well.
+If you use ``from __future__ import annotations``, mypy will understand this
+syntax in annotations. However, since this will not be supported by the Python
+interpreter at runtime, make sure you're aware of the caveats mentioned in the
+notes at :ref:`future annotations import<future-annotations>`.
+
+Using X | Y syntax for Unions
+-----------------------------
+
+Starting with Python 3.10 (:pep:`604`), you can spell union types as ``x: int |
```
JukkaL, line 287: Nit: "... you will be able to spell", since Python 3.10 is not out yet.

docs/source/runtime_troubles.rst
```diff
@@ -0,0 +1,311 @@
...
+
+Using new additions to the typing module
+----------------------------------------
+
+You may find yourself wanting to use features added to the :py:mod:`typing`
+module in earlier versions of Python than the addition, for example, using any
+of ``Literal``, ``Protocol``, ``TypedDict`` with Python 3.6.
+
+The easiest way to do this is to install and use the ``typing_extensions``
+package for the relevant imports, for example:
```
JukkaL, line 306: "package" -> "package on PyPI"?

JukkaL (commented):
These look great! Runtime issues are sometimes painful to deal with, and it's a good to have them all covered in one place. Left a bunch of minor comments (mostly nits).

hauntsaninja:
Thanks, made changes everywhere that I marked a conversation as resolved. The two things I haven't done are:
1) changing tense for Python 3.10. I'd prefer not to, since it's an easy way for documentation to end up sounding really out of date. Plus we already have three alphas of 3.10; it'll be upon us before we know it!
2) adding an example for Python 3.5.1. We're just a release or two away from dropping support for 3.5 altogether; IIRC 3.5.1 doesn't even have typing.overload. If there's another reason to use a `MYPY` constant, I'd be happy to document it in more detail.
I don't feel that strongly though, so if you ask again I'll make the changes! :-)

JukkaL (approved):
Thanks for the updates! Looks good now. I agree with your reasoning, and other changes I proposed aren't important.

PR #9764, opened 2020-11-28
kinds_of_types: clearer sample bytestrings

ferdnyc:
### Description

Edit the code in the `AnyString` example `concat()` calls to use sample arguments that are easier to parse, because `concat(b'a', b'b')` just seems needlessly confusing.

## Test Plan

* Rebuild docs with `cd docs; make html`
* Review `build/html/kinds_of_types.html` for markup issues or syntax errors
* Verify code sample at `#text-and-anystr` in rendered output

### DIFF
--- a/docs/source/kinds_of_types.rst
+++ b/docs/source/kinds_of_types.rst
@@ -831,9 +831,9 @@ so use :py:data:`~typing.AnyStr`:
    def concat(x: AnyStr, y: AnyStr) -> AnyStr:
        return x + y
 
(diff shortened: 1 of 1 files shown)


ferdnyc:
(Note: I thought about using `foo` and `bar`, but that created a similar type of `b'bar'` issue with the second arg. Avoiding any strings beginning with `b` seemed prudent.)

hauntsaninja:
Thanks!

PR #9767, opened 2020-11-29
[mypyc] Generate names lazily when pretty-printing IR or generating C

JukkaL:
Instead of generating names for ops when building IR, we now generate names 
immediately before pretty-printing or generating C. This simplifies `Environment`
and reduces the amount of mutable state we need to maintain. This a step 
towards removing the `mypyc.ir.ops.Environment` class.

Move code related to pretty-printing to `mypyc.ir.pprint`.

Work on mypyc/mypyc#781.


JukkaL:
I'm going to merge this as soon as the builds are green, since I have several other PRs I want to create that depend on this (and this doesn't change behavior).

PR #9781, opened 2020-12-06
Allow dependabot to check github actions weekly

jlosito:
Allow dependabot to check github actions weekly

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/keeping-your-actions-up-to-date-with-dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


JukkaL:
Can you summarize the benefit of this? I quickly browsed the docs but I'm still not quite sure what the concrete impact would be.

jlosito:
This change would upgrade the version of the follow action this project uses.

https://github.com/python/mypy/blob/master/.github/workflows/test.yml#L40

It looks like the most recent version of this action is v2 https://github.com/marketplace/actions/setup-python

As new versions of this action comes out, PRs will be submitted automatically to keep it up to date as well as any other github actions which this project uses.

PR #9796, opened 2020-12-12
Adding more information to "Too Few Arguments"

### LINKED ISSUE #9340: Class __init__ "Too few arguments" should be a better error message

Feature request

```python
class Foo:
    def __init__(self, bar: int):
        pass


f = Foo()
```

```
➜ python test.py
  File "test.py", line 6, in <module>
    f = Foo()
TypeError: __init__() missing 1 required positional argument: 'bar'

~ via ⬢ 13.8.0 via 🐍 3.8.1 
➜ mypy --pretty test.py
test.py:6: error: Too few arguments for "Foo"
    f = Foo()
        ^
```

mypy should say which arguments are missing, like Python does. This was fixed for functions in #505, but it doesn't seem to work for class __init__.

0.782, 3.8.1

*  Do you see the same issue after installing mypy from Git master?

I have not tested this.

--pretty

  the full traceback below.

(You can freely edit this text, please remove all the lines
you believe are unnecessary.)

abhinaypandey02:
### Description

Fixes #9340 

As mentioned in the feature request, mypy now gives more information about the missing required arguments when creating an instance.
The problem was with the if statement that only gave information when the arguments provided were not empty and none. Removing that check gives information for calls with no argument too.

Before:
test.py:6: error: Too few arguments for "Foo"

After:
test.py:7: error: "Foo" missing 1 required positional argument: 'bar'

## Test Plan
All tests passing

### DIFF
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -58,7 +58,7 @@ a.foo(object(), A())  # Fail
 class A:
     def foo(self, x: 'A') -> None: pass
 [out]
-main:3: error: "foo" of "A" missing 1 required positional argument: 'x'
+main:3: error: Missing positional argument "x" in call to "foo" of "A"
 main:4: error: Too many arguments for "foo" of "A"
 main:4: error: Argument 1 to "foo" of "A" has incompatible type "object"; expected "A"
 
@@ -385,7 +385,7 @@ class B(A):
 [case testOverride__init_subclass__WithDifferentSignature]
 class A:
     def __init_subclass__(cls, x: int) -> None: pass
-class B(A):  # E: "__init_subclass__" of "A" missing 1 required positional argument: 'x'
+class B(A):  # E: Missing positional argument "x" in call to "__init_subclass__" of "A"
     def __init_subclass__(cls) -> None: pass
 
 [case testOverrideWithDecorator]
@@ -610,7 +610,7 @@ class B(A):
     def __init__(self) -> None: pass
 class C: pass
 [out]
-main:2: error: "A" missing 1 required positional argument: 'x'
+main:2: error: Missing positional argument "x" in call to "A"
 main:3: error: Too many arguments for "B"
 
 [case testConstructorWithReturnValueType]
@@ -857,7 +857,7 @@ class A:
     def f(self) -> None: pass
 A.f(A())
 A.f(object())     # E: Argument 1 to "f" of "A" has incompatible type "object"; expected "A"
-A.f()             # E: "f" of "A" missing 1 required positional argument: 'self'
+A.f()             # E: Missing positional argument "self" in call to "f" of "A"
 A.f(None, None)   # E: Too many arguments for "f" of "A"
 
 [case testAccessAttributeViaClass]
@@ -895,7 +895,7 @@ class A:
     def __init__(self, x: Any) -> None: pass
     def f(self) -> None: pass
 A.f(A())
-A.f()    # E: "f" of "A" missing 1 required positional argument: 'self'
+A.f()    # E: Missing positional argument "self" in call to "f" of "A"
 
 [case testAssignmentToClassDataAttribute]
 import typing
@@ -1005,7 +1005,7 @@ class A:
         b = B(A())
         if int():
             b = A() # E: Incompatible types in assignment (expression has type "A", variable has type "B")
-            b = B() # E: "B" missing 1 required positional argument: 'a'
+            b = B() # E: Missing positional argument "a" in call to "B"
 [out]
 
 [case testDeclareVariableWithNestedClassType]
@@ -1633,7 +1633,7 @@ class D:
     def __set__(self, inst, v, other): pass
 class A:
     f = D()
-A().f = 'x'  # E: "__set__" missing 3 required positional arguments: 'inst', 'v', 'other'
+A().f = 'x'  # E: Too few arguments for "__set__"
 
 [case testDescriptorDunderSetWrongArgTypes]
 class D:
@@ -1661,7 +1661,7 @@ class D:
     def __get__(self, inst, own, other): pass
 class A:
     f = D()
-A().f = 'x'  # E: "__get__" missing 3 required positional arguments: 'inst', 'own', 'other'
+A().f = 'x'  # E: Too few arguments for "__get__"
 
 [case testDescriptorDunderGetWrongArgTypeForInstance]
 from typing import Any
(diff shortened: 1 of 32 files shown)


abhinaypandey02:
I am sorry I don't have much knowledge about tests. Do I need to manually change all the tests because "Too few arguments" error is invoked plenty of times. Forgive me for sounding stupid I am new.

TH3CHARLie:
> Do I need to manually change all the tests because "Too few arguments" error is invoked plenty of times. 

Yes but you can do it smartly. Using `--update-data` flag can save a lot of trouble, see: https://github.com/python/mypy/blob/master/mypyc/doc/dev-intro.md#ir-tests

TH3CHARLie:
And you also need to fix your code formatting to pass flake8 tests

abhinaypandey02:
> > Do I need to manually change all the tests because "Too few arguments" error is invoked plenty of times.
> 
> Yes but you can do it smartly. Using `--update-data` flag can save a lot of trouble

None of `pytest --update-data mypy`, `pytest --update-data`, `pytest --update-data mypyc` worked for me when executed in the mypyc/ main directory.They all just add a newline above the test that fails. Am I missing something?(Eventually I built my own python script to modify test files)

Also I integrated flake8 with my Visual Studio code. It would be nice if you could help me figure out the flake8 arguments/settings used with mypy.

TH3CHARLie:
To submit a PR to mypy, all you need is to run `runtests.py` in your workspace. The script would run self type checking, formatting checking, and IR tests. There's no need to write your own custom script or use vscode's flake8 functionality.

For the `--update-data` flag, add the flag to `runtests.py` when you actually need it and remember to exclude this change when you are committing the changes.

This may help you have a better understanding of mypy testing workflow: https://github.com/python/mypy/blob/master/test-data/unit/README.md

abhinaypandey02:
> To submit a PR to mypy, all you need is to run `runtests.py` in your workspace. The script would run self type checking, formatting checking, and IR tests. There's no need to write your own custom script or use vscode's flake8 functionality.
> 
> For the `--update-data` flag, add the flag to `runtests.py` when you actually need it and remember to exclude this change when you are committing the changes.
> 
> This may help you have a better understanding of mypy testing workflow: https://github.com/python/mypy/blob/master/test-data/unit/README.md

Thanks a lot and sorry for not giving the readme a read.

test-data/unit/check-classes.test
```diff
@@ -3751,7 +3751,7 @@ int.__eq__(int)
 int.__eq__(3, 4)
 [builtins fixtures/args.pyi]
 [out]
-main:33: error: Too few arguments for "__eq__" of "int"
+main:33: error: Missing positional arguments
```
TH3CHARLie, line 3754: same here

test-data/unit/check-inference-context.test
```diff
@@ -623,7 +623,7 @@ reveal_type((lambda x, y: x + y)(1, 2))  # N: Revealed type is 'builtins.int'
 reveal_type((lambda s, i: s)(i=0, s='x')) # N: Revealed type is 'Literal['x']?'
 reveal_type((lambda s, i: i)(i=0, s='x')) # N: Revealed type is 'Literal[0]?'
 reveal_type((lambda x, s, i: x)(1.0, i=0, s='x')) # N: Revealed type is 'builtins.float'
-(lambda x, s, i: x)() # E: Too few arguments
+(lambda x, s, i: x)() # E: Missing positional arguments
```
TH3CHARLie, line 626: same for lambda functions

TH3CHARLie (commented):
The change to existing code needs some second thoughts, see comments below.

test-data/unit/check-classes.test
abhinaypandey02: I understand callable and lamda functions. But here are you pointing at magic methods?

abhinaypandey02: If I make my message something as `Missing positional argument for call to "__eq__" of "int"` then the only difference would be language. So is it about language or the missing function name here?

TH3CHARLie: I think the original language is sufficient. The new one does not make much improvement.

abhinaypandey02: So I think where the information about the arguments to be provided is not available, i.e. the message is just "Missing positional argument", there we can leave it to the old language? Should I code it that way?

TH3CHARLie: exactly. For callables, lambdas, magic methods (and maybe more), the names of the arguments are unknown so staying to the old language is totally fine. Let's improve the language only for user-defined functions and methods in this PR. If there's a need to improve other cases in the future, we can do them later in separate PRs.

test-data/unit/check-errorcodes.test
```diff
@@ -214,7 +214,7 @@ main:5: error: Invalid "type: ignore" comment  [syntax]
 
 [case testErrorCodeArgKindAndCount]
 def f(x: int) -> None: pass  # N: "f" defined here
-f()  # E: Too few arguments for "f"  [call-arg]
+f()  # E: Missing positional argument "x" in call to "f"  [call-arg]
```
TH3CHARLie, line 217: this is a perfect example of what's expected in this PR IMO

TH3CHARLie:
Thanks for the updates! Can you verify the case in #9340? For example, use the following code(copied from the issue):
```
class Foo:
    def __init__(self, bar: int):
        pass

c = Foo()

def foo(baz: int, bas: int):
    pass

foo()
```
You may also make this a new test case. I did some verification locally and looks like the changes don't work on this segment.

abhinaypandey02:
@TH3CHARLie I have tested the code in the issue #9340 and the new changes reflect. I made a new test case to check missing positional arguments using the code in the issue and it passes.

TH3CHARLie (approved):
Thanks for the updates and verifications, looks good to me. However I am not sure I am 100% in sync with the expectations in #9340, @gvanrossum Would you mind taking a look if you have time?

abhinaypandey02:
> Thanks for the updates and verifications, looks good to me. However I am not sure I am 100% in sync with the expectations in #9340, @gvanrossum Would you mind taking a look if you have time?

Thanks a lot for your review and approval.

gvanrossum (approved):
Hm, the number of tests that have to be fixed is a bit much, and in some cases the name of the missing argument isn't very useful, but I agree its most consistent. So weak +1.

TH3CHARLie:
> Hm, the number of tests that have to be fixed is a bit much, and in some cases the name of the missing argument isn't very useful, but I agree its most consistent. So weak +1.

Thanks for the review. I am merging this now but I'll keep an eye on this(and respond to any future feedbacks about it) since error messages are important and pervasive

PR #9797, opened 2020-12-12
Fix path escaping bug in HTML report (GH8979)

### LINKED ISSUE #8979: Path escaping bug in HTML report
Bug in MyPy 0.780 with Python 3.7.1 on Windows 10, including args `--html_report coverage_report/`

The template for the HTML report includes a path without normalizing the path separators, so on windows uses '\\' instead of '/', and this appears in the browser as %5C: 

    <a href="html/example%5Capp%5Cmain.py.html">example.app.main</a>

The browser then misbehaves when when navigating to this url with backslashes in it, appearing like so in the URL bar

    file:///C:/proj/example/coverage_report/html/example\app\main.py.html

It can be found [here](https://github.com/python/mypy/blob/793cf187f0e529f9d6623a512ae0149124ded542/mypy/xml/mypy-html.xslt#L40) in the template, and I believe is formatting [the name property](https://github.com/python/mypy/blob/793cf187f0e529f9d6623a512ae0149124ded542/mypy/nodes.py#L286-L288) of the `MyPyFile` class

rhkleijn:
### Description

Fixes #8979

One-liner fix for path escaping bug in HTML report on windows.

### DIFF
--- a/mypy/report.py
+++ b/mypy/report.py
@@ -522,7 +522,7 @@ class MemoryXmlReporter(AbstractReporter):
             etree.SubElement(root, 'file',
                              file_info.attrib(),
                              module=file_info.module,
-                             name=file_info.name,
+                             name=pathname2url(file_info.name),
                              total=str(file_info.total()))
         xslt_path = os.path.relpath('mypy-html.xslt', '.')
         transform_pi = etree.ProcessingInstruction('xml-stylesheet',


gvanrossum (approved):
(Are there any other places in this file where the same change would be beneficial?)

rhkleijn:
As far as I can tell, there are no obvious other places where the same change would be beneficial.

The fix is very specific to the conversion from a filename to a url, which only seems to happen in this particular line.
Since the fix is in the MemoryXmlReporter class which is used internally in the other XML-based reporters it however already extends to both --html-report and --xml-report.

PR #9798, opened 2020-12-13
Fix type check failure stemming from new pytest release

hauntsaninja:
The release of pytest 6.2 broke mypy's type checking of itself, since pytest 6.2 drops support for Python 3.5 (https://docs.pytest.org/en/stable/changelog.html#pytest-6-2-0-2020-12-12)

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -5,7 +5,8 @@ flake8-bugbear; python_version >= '3.5'
 flake8-pyi>=20.5; python_version >= '3.6'
 lxml>=4.4.0
 psutil>=4.0
-pytest>=6.1.0,<7.0.0
+# pytest 6.2 does not support Python 3.5
+pytest>=6.1.0,<6.2.0
 pytest-xdist>=1.34.0,<2.0.0
 pytest-forked>=1.3.0,<2.0.0
 pytest-cov>=2.10.0,<3.0.0


gvanrossum:
The subject line doesn’t really cover the fix?

hauntsaninja:
I didn't do a great job with the commit message... it looks like the release of pytest 6.2 broke mypy's type checking of itself: https://travis-ci.com/github/python/mypy/jobs/458991721

test-requirements.txt
```diff
@@ -5,7 +5,8 @@ flake8-bugbear; python_version >= '3.5'
 flake8-pyi>=20.5; python_version >= '3.6'
 lxml>=4.4.0
 psutil>=4.0
-pytest>=6.1.0,<7.0.0
+# pytest 6.2 does not support Python 3.5
+pytest>=6.1.0,<6.2.0
```
gvanrossum, line 9: Can’t you select pre-6.2 for Python 3.5 only?

hauntsaninja: Oh, oops... My urge to get things green was all consuming and I merged right before you posted this, sorry :-(
We definitely could, although in the past we've had to make changes to mypy source code based on pytest type annotation changes, so things are a little simpler if we stick to a single pytest version for all of mypy. Hopefully we can drop Python 3.5 support soon anyway.

gvanrossum: I worry that this declaration will stay behind even after we drop 3.5, and once we try to catch up with pytest we’d have a bunch of issues to debug all at once. But it’s fine, we can cross that bridge when we get to it.

bluetech:
IIUC, things broke here because the mypy self check runs on Python 3.7, which means the latest pytest gets installed (because it is compatible with Python 3.7, so pytest's `python_requires >= 3.6` doesn't apply), but mypy is run with `python_version = 3.5` which rejects the 3.6 syntax.

The easiest fix would be to run the CI job itself on Python 3.5, but that may not be desirable.

hauntsaninja:
I think this is the right fix. test-requirements.txt is used by mypy devs, not just CI, and assuming most mypy devs aren't using Python 3.5 it would still make sense for it to specify a pytest that allows mypy's self type check to pass.
(Note also that we've had the opposite problem once or twice; type checking of mypy has been broken on Python 3.8 and not showed up in CI, e.g. https://github.com/python/mypy/pull/9693)

PR #9799, opened 2020-12-13
Fix incomplete error codes documentation

revolter:

(Once you have, delete this section. If you leave it in, your PR may be closed without action.)

### Description

## Test Plan

### DIFF
--- a/docs/source/error_codes.rst
+++ b/docs/source/error_codes.rst
@@ -24,7 +24,7 @@ Displaying error codes
 ----------------------
 
 Error codes are not displayed by default.  Use :option:`--show-error-codes <mypy --show-error-codes>`
-to display error codes. Error codes are shown inside square brackets:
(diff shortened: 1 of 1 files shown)


TH3CHARLie:
This has been covered in https://mypy.readthedocs.io/en/stable/config_file.html?highlight=config%20file

revolter:
Indeed, but after Googling something, I found https://mypy.readthedocs.io/en/stable/error_codes.html instead. And, because there are 2 ways to do 1 thing, it doesn't make sense to mention only one in a generic article like the "Error codes" one.

revolter:
Do you think that this small change has any drawbacks?

JelleZijlstra (approved):

JelleZijlstra:
I don't think a small amount of redundancy in the documentation hurts.

PR #9801, opened 2020-12-13
[mypyc] Some tagged integer micro-optimizations

JukkaL:
Use `PyLong_FromLong` instead of `PyLong_FromSsize_t` since it's faster
(at least on Python 3.8) and happens to work the same on macOS and Linux. 
This is an extremely hot operation.

Add some `unlikely(...)` and `likely(...)` to tagged integer primitives.

Overall these made the Richards benchmark about 4% faster, which was a 
surprisingly big improvement. 

(I'm sure there are still some remaining easy performance wins in tagged 
integer operations.)

Work on mypyc/mypyc#757.


JukkaL:
Apparently on Windows `long` is 32 bits. Looks like this doesn't work on Windows, but we can still use it on supported platforms (the generated C isn't intended to be portable across platforms).

ilevkivskyi (approved):

PR #9804, opened 2020-12-15
fixing the --update-data flag for pytest

abhinaypandey02:
### Description

Fixing completely broken --update-data flag for pytest.

I had a great deal of pain to update tests as the --update-data flag didn't do a thing. On examining I realised that few logics are incorrect. I will try to explain them.

To determine the last line of the test it uses `item.line`. Well each `item` is the content between statements of the type [file], [case], [out], etc. `item.line` provides the first line of the item. Previously, testcase.lastline was defined as the first line of the last `item`. But the last item might consist of more than one line for example-
```
[out]
Error 1
Error 2
Error 3
```
Here the test case will be limited to Error 1. So we need to instead make `testcase.lastline` as `item.line + len(item.data) - 1` 

Also while performing an update to test it considered `test = '\n'.join(data_lines[testcase.line:testcase.lastline])` but here `testcase.lastline` is the length ant the line numer so it needs to be `test = '\n'.join(data_lines[testcase.line:testcase.line+testcase.lastline-1])`

## Test Plan

I made changes to messages.py and earlier it showed test failed but did not update them. But now it updates test cases with the output.(with --update-data flag)

### DIFF
--- a/mypy/test/data.py
+++ b/mypy/test/data.py
@@ -149,7 +149,7 @@ def parse_test_case(case: 'DataDrivenTestCase') -> None:
     case.input = input
     case.output = output
     case.output2 = output2
-    case.lastline = item.line+len(item.data)-1
+    case.last_line = case.line + item.line + len(item.data) - 2
     case.files = files
     case.output_files = output_files
     case.expected_stale_modules = stale_modules
@@ -185,7 +185,7 @@ class DataDrivenTestCase(pytest.Item):
     normalize_output = True
 
     # Extra attributes used by some tests.
-    lastline = None  # type: int
+    last_line = None  # type: int
     output_files = None  # type: List[Tuple[str, str]] # Path and contents for output files
     deleted_paths = None  # type: Dict[int, Set[str]]  # Mapping run number -> paths
     triggered = None  # type: List[str]  # Active triggers (one line per incremental step)
--- a/mypy/test/data.py
+++ b/mypy/test/data.py
@@ -149,7 +149,7 @@ def parse_test_case(case: 'DataDrivenTestCase') -> None:
     case.input = input
     case.output = output
     case.output2 = output2
-    case.lastline = item.line
+    case.lastline = item.line+len(item.data)-1
     case.files = files
     case.output_files = output_files
     case.expected_stale_modules = stale_modules
--- a/mypy/test/helpers.py
+++ b/mypy/test/helpers.py
@@ -148,7 +148,7 @@ def update_testcase_output(testcase: DataDrivenTestCase, output: List[str]) -> N
     testcase_path = os.path.join(testcase.old_cwd, testcase.file)
     with open(testcase_path, encoding='utf8') as f:
         data_lines = f.read().splitlines()
-    test = '\n'.join(data_lines[testcase.line:testcase.lastline])
+    test = '\n'.join(data_lines[testcase.line:testcase.line+testcase.lastline-1])
 
     mapping = {}  # type: Dict[str, List[str]]
     for old, new in zip(testcase.output, output):
@@ -168,7 +168,7 @@ def update_testcase_output(testcase: DataDrivenTestCase, output: List[str]) -> N
                 list(chain.from_iterable(zip(mapping[old], betweens[1:])))
             test = ''.join(interleaved)
 
-    data_lines[testcase.line:testcase.lastline] = [test]
+    data_lines[testcase.line:testcase.line+testcase.lastline-1] = [test]
     data = '\n'.join(data_lines)
     with open(testcase_path, 'w', encoding='utf8') as f:
         print(data, file=f)
--- a/mypy/test/helpers.py
+++ b/mypy/test/helpers.py
@@ -148,7 +148,7 @@ def update_testcase_output(testcase: DataDrivenTestCase, output: List[str]) -> N
     testcase_path = os.path.join(testcase.old_cwd, testcase.file)
     with open(testcase_path, encoding='utf8') as f:
         data_lines = f.read().splitlines()
-    test = '\n'.join(data_lines[testcase.line:testcase.line+testcase.lastline-1])
+    test = '\n'.join(data_lines[testcase.line:testcase.last_line])
 
     mapping = {}  # type: Dict[str, List[str]]
     for old, new in zip(testcase.output, output):
@@ -168,7 +168,7 @@ def update_testcase_output(testcase: DataDrivenTestCase, output: List[str]) -> N
                 list(chain.from_iterable(zip(mapping[old], betweens[1:])))
             test = ''.join(interleaved)
 
-    data_lines[testcase.line:testcase.line+testcase.lastline-1] = [test]
+    data_lines[testcase.line:testcase.last_line] = [test]
     data = '\n'.join(data_lines)
     with open(testcase_path, 'w', encoding='utf8') as f:
         print(data, file=f)


TH3CHARLie:
> Fixing completely broken --update-data flag for pytest.
>
> I made changes to messages.py and earlier it showed test failed but did not update them. But now it updates test cases with the output.(with --update-data flag)

I'd like to know about how you use the flag when you are updating the IR tests, because according to the best of my knowledge, the only thing abnormal(but not incorrect) is producing an extra newline at the end of some tests.

abhinaypandey02:
> > Fixing completely broken --update-data flag for pytest.
> > I made changes to messages.py and earlier it showed test failed but did not update them. But now it updates test cases with the output.(with --update-data flag)
> 
> I'd like to know about how you use the flag when you are updating the IR tests, because according to the best of my knowledge, the only thing abnormal(but not incorrect) is producing an extra newline at the end of some tests.

I added the --update-data flag to runtests.py `'pytest-fast': 'pytest --update-data -k "not (%s)"' % ' or '.join(ALL_NON_FAST),` same with `pytest-slow`. I also tried running just `pytest --update-data` but all of these produce the same output, i.e. adding just one newline before every test that needs to be changed. I debugged this and found that in `test = '\n'.join(data_lines[testcase.line:testcase.lastline])` `testcase.line` has the actual line number like 309 but the `testcase.lastline` is the length, something like 13 so its keeping the test blank.

gvanrossum:
I can repro this behavior. I made a small change to the error for Missing positional arguments and ran this:
```
pytest -n0 --update-data .\mypy\test\testcheck.py -k testInitSubclassWithImports
```
This failed and afterwards the diff looked like this:
```
diff --git a/test-data/unit/check-classes.test b/test-data/unit/check-classes.test
index 40d057ad..31a7c8ec 100644
--- a/test-data/unit/check-classes.test
+++ b/test-data/unit/check-classes.test
@@ -6447,6 +6447,7 @@ tmp/builtins.pyi:5: note: "__init_subclass__" of "object" defined here
 [builtins fixtures/object_with_init_subclass.pyi]

 [case testInitSubclassWithImports]
+
 from init_subclass.a import Base
 class Child(Base, thing=5):  # E: Missing positional argument "default_name" in call to "__init_subclass__" of "Base"
     pass
```
Maybe it's due to the -k flag? Or Windows?

abhinaypandey02:
> I can repro this behavior. I made a small change to the error for Missing positional arguments and ran this:
> 
> ```
> pytest -n0 --update-data .\mypy\test\testcheck.py -k testInitSubclassWithImports
> ```
> 
> This failed and afterwards the diff looked like this:
> 
> ```
> diff --git a/test-data/unit/check-classes.test b/test-data/unit/check-classes.test
> index 40d057ad..31a7c8ec 100644
> --- a/test-data/unit/check-classes.test
> +++ b/test-data/unit/check-classes.test
> @@ -6447,6 +6447,7 @@ tmp/builtins.pyi:5: note: "__init_subclass__" of "object" defined here
>  [builtins fixtures/object_with_init_subclass.pyi]
> 
>  [case testInitSubclassWithImports]
> +
>  from init_subclass.a import Base
>  class Child(Base, thing=5):  # E: Missing positional argument "default_name" in call to "__init_subclass__" of "Base"
>      pass
> ```
> 
> Maybe it's due to the -k flag? Or Windows?

@gvanrossum Thanks for your review and time. I had the same issue in Ubuntu, so I am not sure if it's about windows. Also running without -k and just `pytest --update data` gave the same issue. I strongly believe it's because of the fault in logic I indicated.

TH3CHARLie:
I see, I didn't encounter this before cuz I run the mypyc tests only.

mypy/test/data.py
```diff
@@ -149,7 +149,7 @@ def parse_test_case(case: 'DataDrivenTestCase') -> None:
     case.input = input
     case.output = output
     case.output2 = output2
-    case.lastline = item.line
+    case.lastline = item.line+len(item.data)-1
```
gvanrossum, line 152: For readability, please add spaces around the + and - operators.

Also, consider renaming this item to "numlines", and not subtracting 1. That would make the slices in helpers.py cleaner. It would also catch other places where this variable is used or set that might need adjusting.

abhinaypandey02: Well, I was not able to understand you completely.
For example-
```
239:[case testSample]
240:line 1
241:line 2
242:line 3
243:[out]
244:error 1
245:error 2
246:error 3
247:
```
**Before my changes**
`case.line=239`<br>
`case.lastline=6`<br>
`test = '\n'.join(data_lines[239:6])`<br>
**After my changes**
`case.line=239`<br>
`case.lastline=6 + 3 (len(item.data)) - 1 = 8 (exact length of the test case)`<br>
`test = '\n'.join(data_lines[ 239 : 239 + ( 6 + 3 - 1 ) - 1 ] ) =  '\n'.join(data_lines[ 239 : 246 ] )`<br>

Now if I change the class property `lastline` to `numlines` but do not subtract 1, then `case.numlines` will be 1 more than the number of lines in the test case and we will have to subtract 2 instead of 1 in the slices in helpers.py.
Instead, I have a suggestion to give `case.lastline` the actual meaning instead of number of lines. This will make the slices in helpers.py as [test.line:test.lastline]. 
For that we'll have to make `case.lastline=case.line + item.line + len(item.data) - 2` 

What do you think would be a better option?

gvanrossum: Okay, I like that suggestion. I hadn't noticed that both `-1` adjustment pointed in the same direction! The first one presumably compensates for the `[case foobar]` line, which isn't included (`case.line` points to the line after it).

Just rename the variable, please, so we know it's not used in the old sense elsewhere.

abhinaypandey02:
I have no clue about this test failure. `runtests.py` and flake8 passed with no issues but I don't understand this.

gvanrossum:
Wondering if the test failure is a Travis glitch. I'll close and reopen the issue.

gvanrossum:
@hauntsaninja Sorry to invoke your powers, but do you have any idea why the Travis-CI build here is failing?

@abhinaypandey02 One thing you could do is fetch the latest master and merge it into your branch, then push it again. Who knows this has been fixed recently... you are 7 commits behind.

hauntsaninja:
This smells to me like pip's new dependency resolver is doing something annoyingly over-smart. I think pip 20.3.2 might be unable to upgrade itself to pip 20.3.3... There's a chance pip 20.3.3 *is* the fix based on https://pip.pypa.io/en/stable/news/#id1 It's unclear to me why this hasn't affected any other builds, though.

The only sure shot solution here is to do what everyone at my work seems to be doing, which is to pin pip to <20.3. But if you wanted to try other things:
1) pin to latest pip version
2) install upgrade pip and setuptools separately, in the hope of pleasing the dependency resolver
3) run pip without progress bars. sure, it'll take 20 minutes for pip to figure out what to install, but travis won't kill the job due to excessive logging.

hauntsaninja:
It looks like in other builds Travis is using pip 20.3.3. So I'd try merging master and maybe you'll get a build with a pip version that doesn't need to update itself.

abhinaypandey02:
@gvanrossum I merged the changes. But I think I might have messed up because the new changes also show up in the Files Changed. Was this the expected output? But the test still failed :(

gvanrossum:
No, that usually doesn't happen when I merge from main. But I am not enough of a git+GitHub expert to be able to guess what you did differently from what I usually do (and honestly I don't recall the exact sequence of git commands I use for such things).

You could start creating a new PR and see if that fares better; that's probably faster than trying to debug this.

abhinaypandey02:
> No, that usually doesn't happen when I merge from main. But I am not enough of a git+GitHub expert to be able to guess what you did differently from what I usually do (and honestly I don't recall the exact sequence of git commands I use for such things).
> 
> You could start creating a new PR and see if that fares better; that's probably faster than trying to debug this.

Okay then I will close this and start a new PR

PR #9806, opened 2020-12-15
[mypyc] Attempt to fix macOS wheel builds

JukkaL:


ilevkivskyi:
I am going to land this simple fix for now to simplify internal pin move. We may consider a more robust fix later.

PR #9813, opened 2020-12-16
tests: Avoid side effect of COLUMNS env var

gmc444-b:
### Description

Several tests rely on getting the default width of a terminal for
rendering test output. Some PTY environments (in my case the Emacs
shell) set the env var `COLUMNS` to the terminal width, which can
cause several tests to fail due to formatting mismatches.

Prior to running thesee tests, mock out `sys.environ`, removing the
`COLUMNS` env var when present.

## Test Plan

The change should only effect environments like my own. After this change, tests pass in my environment. The change can be manually tested in other environments by running

``` Shell
COLUMNS=120 ./runtests.py
```

before, then after applying the change.


gvanrossum (approved):
LG. I recall seeing this problem occasionally; thanks for taking the time to track it down!

PR #9818, opened 2020-12-18
Make sure that enum members are treated as enum.

### LINKED ISSUE #9817: Enum members are Enum themselves. Yet mypy sometimes gets confused
Enum members are Enum themselves. Yet mypy sometimes gets confused.
For example:
```
from enum import Enum

class Spam(Enum):
    FOO = "Foo"

m = Spam.FOO
print(m.FOO.FOO)  # Valid python. mypy error: "str" has no attribute "FOO"
print(m == m.FOO)  # Valid python. mypy error: Non-overlapping equality check
print(m.FOO.startswith("Foo"))  # Invalid python. mypy does not complain.
```
The first two print calls are valid python code. mypy thinks that `m.FOO` is `str`. (mypy only shows an error on the second print if `--strict-equality` is specified.)
The last print raises an AttributeError. mypy does not complain because it things that `m.FOO` is `str`.
```
$ mypy --version
mypy 0.790
```

udifuchs:
The type of Spam.FOO is Spam in:
```py
class Spam(Enum):
    FOO = "Foo"
```
Yet mypy sometimes treats it as a string.
This commit makes sure that `Spam.FOO` is treated as `Spam`.

Fixes #9817.

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -596,6 +596,20 @@ def analyze_var(name: str,
             mx.not_ready_callback(var.name, mx.context)
         # Implicit 'Any' type.
         result = AnyType(TypeOfAny.special_form)
+
+    if info.is_enum and info == itype.type:
+        # enum members are enums themselves.
+        #
+        # class Spam(enum.Enum):
+        #    FOO = "Foo"
+        #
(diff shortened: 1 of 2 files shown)


gvanrossum:
See #9817. I don't think this is worth pursuing, and IMO it just promotes bad style.

PR #9834, opened 2020-12-23
allow packages/modules as args with files in cfg

trws:
### Description

Currently only files can be specified on the command line if files are
specified in mypy.ini.  This looks a great deal like a bug, especially
since packages and modules cannot be specified in the configuration
file. 

This PR changes the logic to only use the files from the ini
file if none of (modules/packages/files) are specified on the command
line and adds tests to verify the new behavior.

I would also like to add packages and modules keys to the configuration file so a project I work on can use packages as its default without arguments, but want to gauge interest in that first, where this looked like it was probably a mistake with a straightforward solution.

## Test Plan

Three new tests have been added to the codlin.test file that test that the conflict is still noted when both are placed on the command line, as well as two tests that ensure the right package/module is selected and checked when a package or module is on the command line with files in the config.

This is my first time working with this test format, pytest runs them and gives reliable results on my machine for these new tests, but if these are outside of style or similar I'll be happy to fix them.

### DIFF
--- a/mypy/main.py
+++ b/mypy/main.py
@@ -851,9 +851,9 @@ def process_options(args: List[str],
     if special_opts.no_executable or options.no_site_packages:
         options.python_executable = None
 
-    # Paths listed in the config file will be ignored if any paths are passed on
-    # the command line.
-    if options.files and not special_opts.files:
+    # Paths listed in the config file will be ignored if any paths, modules or packages
+    # are passed on the command line.
+    if options.files and not (special_opts.files or special_opts.packages or special_opts.modules):
         special_opts.files = options.files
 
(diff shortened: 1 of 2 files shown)


JukkaL:
> I would also like to add packages and modules keys to the configuration file so a project I work on can use packages as its default without arguments, but want to gauge interest in that first, where this looked like it was probably a mistake with a straightforward solution.

This sounds like a useful feature. If you want to pursue this, please open an issue and describe how this would work in some detail.

JukkaL (approved):
Thanks for the PR! The original behavior looks like a mistake, and the new behavior is clearly more logical.

Issue #9833, opened 2020-12-23
Mypy will not automatically check files with multiple dots in their names


jacksonokuhn:
**Bug Report**

Files with names of the form `x.y.py` don't get checked by mypy.

**To Reproduce**

```
mkdir mypy-test
touch mypy-test/a.py
touch mypy-test/a.b.py
mypy mypy-test
```

**Expected Behavior**

You'd expect mypy to check two source files with the message `Success: no issues found in 2 source files`

**Actual Behavior**

Mypy only checks one source file: `Success: no issues found in 1 source file`

**Your Environment**

- Mypy version used: 0.790
- Mypy command-line flags: N/A
- Mypy configuration options from `mypy.ini` (and other config files): N/A
- Python version used: 3.7.4
- Operating system and version: Mac 10.15.7

gvanrossum:
Those are not legal Python file names anyway.

hauntsaninja:
As it happens, this is no longer the behaviour on master.
This is a side effect of #9742 making `mypy dir/` behave more like `mypy file.py`. mypy already allowed `mypy a.b.py`, so it works with the recursive invocation now.

Note that all versions associate the file with the incorrect module name "a.b". I tried to improve this, but it caused a couple issues and I backtracked to keep the PR surface smaller. See point 2 of https://github.com/python/mypy/pull/9742#issuecomment-731715016

PR #9835, opened 2020-12-23
find_sources: deal more robustly with filenames with periods

hauntsaninja:
Note that we deal with the possibility of parent_module containing a
period in _crawl_up_helper

Brought up in #9833

For context on the preexisting comment about invalid module names, see https://github.com/python/mypy/pull/9742#issuecomment-731715016


hauntsaninja:
Actually, I don't really like this. If this causes trouble, we'll get a `Duplicate module named "a.b"` error with the two filenames.

PR #9839, opened 2020-12-27
Use Py_TRASHCAN_{BEGIN,END} in tp_dealloc functions

### LINKED ISSUE #789: Python 2.7 stub for subprocess

msullivan:
See https://github.com/python/cpython/blob/master/Include/cpython/object.h#L456

Fixes https://github.com/mypyc/mypyc/issues/789.

I didn't add a test because a test needs to create a lot of nodes.
Tested manually.

### DIFF
--- a/mypyc/codegen/emitclass.py
+++ b/mypyc/codegen/emitclass.py
@@ -627,10 +627,10 @@ def generate_dealloc_for_class(cl: ClassIR,
     emitter.emit_line('{}({} *self)'.format(dealloc_func_name, cl.struct_name(emitter.names)))
     emitter.emit_line('{')
     emitter.emit_line('PyObject_GC_UnTrack(self);')
-    emitter.emit_line('Py_TRASHCAN_BEGIN(self, {})'.format(dealloc_func_name))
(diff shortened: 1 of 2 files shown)


mypyc/codegen/emitclass.py
```diff
@@ -627,8 +627,10 @@ def generate_dealloc_for_class(cl: ClassIR,
     emitter.emit_line('{}({} *self)'.format(dealloc_func_name, cl.struct_name(emitter.names)))
     emitter.emit_line('{')
     emitter.emit_line('PyObject_GC_UnTrack(self);')
+    emitter.emit_line('CPy_TRASHCAN_BEGIN(self, {})'.format(dealloc_func_name))
```
JukkaL, line 630: Add short description of why we need to do this.

JukkaL (approved):
Thanks, this was a nasty bug. I assume that there's no easy way to fix this on Python 3.7 and earlier. Maybe we could implement something similar ourselves eventually?

Please add a follow-up issue to write a regression test to cover this case, once we have a way of adding expensive tests that only get run daily, for example.

msullivan:
This does work on on 3.7, but a new API got introduced in 3.8

PR #9846, opened 2020-12-29
[mypyc] Refactor method IR generation

JukkaL:
Add new helpers for generating methods that reduces the required
boilerplate. Switch several existing use cases to use the new helpers.

Summary of the new helpers:

1. Call `builder.enter_method(...)` to begin generating IR for a method.
2. Call `builder.add_argument(..)` for each non-self argument.
3. Generate method body normally.
4. Call `builder.leave_method()`.

Previously `enter()` and `leave()` had to be used, along with several
additional steps, such as defining `self` explicitly. The new approach 
is much easier to use and less error-prone. 

Since not all uses of the old interface have been removed, there is 
still some duplication. It would be great to eventually always use the 
higher-level interface to generate methods and functions.

Work on mypyc/mypyc#781.

### DIFF
--- a/mypyc/irbuild/generator.py
+++ b/mypyc/irbuild/generator.py
@@ -206,13 +206,10 @@ def add_throw_to_generator_class(builder: IRBuilder,
                                  fn_decl: FuncDecl,
                                  sig: FuncSignature) -> None:
     """Generates the 'throw' method for a generator class."""
-    builder.enter(fn_info)
-    self_reg = builder.read(builder.add_self_to_env(fn_info.generator_class.ir))
-
-    # Add the type, value, and traceback variables to the environment.
-    typ = builder.add_local_reg(Var('type'), object_rprimitive, True)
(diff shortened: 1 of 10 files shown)


mypyc/irbuild/generator.py
```diff
@@ -157,78 +157,51 @@ def add_helper_to_generator_class(builder: IRBuilder,
...
-    builder.functions.append(next_fn_ir)
+    builder.leave_method()
 
 
 def add_send_to_generator_class(builder: IRBuilder,
                                 fn_info: FuncInfo,
                                 fn_decl: FuncDecl,
                                 sig: FuncSignature) -> None:
     """Generates the 'send' method for a generator class."""
-    # FIXME: this is basically the same as add_next...
```
JukkaL, line 200: This function is a good example of how the old and new approaches differ.

JukkaL:
I'm going to merge this without review since this only refactors things and I have other changes that depend on this. 

If somebody wants to review this, I'm happy to create a follow-up PR that addresses review feedback.

PR #9854, opened 2020-12-29
[mypyc] Speed up 'if x' for int values

JukkaL:
This speeds up the `if_true` microbenchmark by about 5%. Code size is also
reduced.


TH3CHARLie (approved):
Looks great! Thanks!

PR #9865, opened 2020-12-30
Support TypeGuard (PEP 647)

gvanrossum:
While PEP 647 is still in draft mode, this is (or if it isn't will soon be) a full implementation.

This depends on a typeshed PR: https://github.com/python/typeshed/pull/4879.


mypy/checker.py
```diff
@@ -4001,6 +4002,15 @@ def find_isinstance_check_helper(self, node: Expression) -> Tuple[TypeMap, TypeM
                 if literal(expr) == LITERAL_TYPE:
                     vartype = type_map[expr]
                     return self.conditional_callable_type_map(expr, vartype)
+            else:
+                if (isinstance(node.callee, RefExpr)
+                        and isinstance(node.callee.node, SYMBOL_FUNCBASE_TYPES)
+                        and isinstance(node.callee.node.type, CallableType)
+                        and node.callee.node.type.type_guard is not None):
+                    if len(node.args) < 1:  # TODO: Is this an error?
```
hauntsaninja, line 4010: This seems fine to me, it's similar to `if isinstance()` / you have a TODO in your tests to error at the definition site.

hauntsaninja (commented):
This looks pretty reasonable to me!

Even though it's not in the direct path of changes, might be worth adding a test for interactions with the walrus operator.

Also, I'm guessing higher order use of type guarded functions doesn't work? E.g. something like:
```
def filter(fn: Callable[[T1], TypeGuard[T2]], it: Iterable[T1]) -> Iterator[T2]: ...

reveal_type(filter(...))
```

gvanrossum:
> Also, I'm guessing higher order use of type guarded functions doesn't work?

Hm, it doesn't, but it really should. Unfortunately that's going to require a different approach, and I'm not sure how to do it. There are an awful lot of places where a Callable's `ret_type` attribute is used, and I'm hesitant to try and audit them all to see whether I should redirect the code to the `type_guard` attribute or not.

hauntsaninja:
I don't have a general answer for you, but maybe add something to this line and you get lucky and it mostly just works? https://github.com/python/mypy/blob/aa97427dad5b8e61c23f59ee500e86d34d326a33/mypy/constraints.py#L460

gvanrossum:
> I don't have a general answer for you, but maybe add something to this line and you get lucky and it mostly just works?
> 
> https://github.com/python/mypy/blob/aa97427dad5b8e61c23f59ee500e86d34d326a33/mypy/constraints.py#L460

Thanks, right on the nose. It was this and a bit in expandtype.py.

gvanrossum:
However, a type guard that's a method doesn't work (no generics involved
even). How are those different?

test-data/unit/check-typeguard.test
```diff
@@ -0,0 +1,147 @@
...
+def main(a: object) -> None:
+    if guard.is_point(a):
+        reveal_type(a)  # N: Revealed type is 'points.Point'
+[file guard.py]
+from typing_extensions import TypeGuard
+import points
+def is_point(a: object) -> TypeGuard[points.Point]: pass
+[file points.py]
+class Point: pass
+[builtins fixtures/tuple.pyi]
```
JukkaL, line 147: Idea about additional tests (sorry if some of these already exist):

* Test a type guard function with a non-empty body. Check that a non-boolean return value is flagged as an error.
* Test a type guard that narrows down `Dict[str, object]` to a TypedDict.
* Test using type guard with an assert.
* Test using two type guards with 'or', such as `if is_int(x) or is_float(x):`.
* Test narrowing down from `Any` using a type guard.

mypy/types.py
```diff
@@ -270,6 +270,16 @@ def copy_modified(self, *,
             self.line, self.column)
 
 
+class TypeGuardType(Type):
+    """Only used by find_instance_check() etc."""
```
JukkaL, line 274: Add link to the PEP in the docstring.

JukkaL (commented):
Thanks for working on this! Overall looks good -- the changes are clean and localized. This may not be a very commonly needed feature, but in cases where this is useful the current alternatives are all quite bad.

I did a quick review pass and left suggestions about additional test cases. I'll read the PEP in detail and play around with the implementation, and I'll do another round of review afterwards.

JukkaL:
A few more ideas about tests:

* Test that `if not is_int(x)` does not narrow `Union[int, str]` to `str` (also in the else block of `if is_int(x)`).
* Test a type guard implemented as a class method.

gvanrossum:
Also (note to self) some tests with keyword args and *args, **kwargs (not
sure what's supposed to happen for those).

JukkaL:
Two more things that may be worth testing:
* Overloaded function where one item/all items have a type guard.
* Decorated type guard function (decorator has type T -> T).

JukkaL:
More possible things to test:

* If a method is a type guard, do we need to ensure that any overrides in subclasses are type guards (and compatible type guards)?
* If we have a higher-order function that takes a type guard, test that the passed function is a suitable type guard (currently it looks like we can pass an arbitrary function that returns `bool`).

(It's also okay to create follow-up issues for some of these if they don't work currently.)

gvanrossum:
I added tests for all your suggestions (the "overload" test doubles for overloading and higher-order-functions).

This found several problems -- for now I'm just skipping those 5 tests. I'll look into fixing them but I might take your offer of filing issues for some of them.

test-data/unit/check-typeguard.test
```diff
@@ -0,0 +1,299 @@
...
+        reveal_type(a)  # N: Revealed type is 'builtins.float'
+[builtins fixtures/classmethod.pyi]
+
+[case testTypeGuardRequiresPositionalArgs]
+from typing_extensions import TypeGuard
+def is_float(a: object, b: object = 0) -> TypeGuard[float]: pass
+def main1(a: object) -> None:
+    # This is debatable -- should we support these cases?
+
+    if is_float(a=a, b=1):  # E: type guard requires positional argument
```
JukkaL, line 230: Nit: Use upper case first letter in error message.

test-data/unit/check-typeguard.test
```diff
@@ -0,0 +1,299 @@
...
+
+def main(a: List[Optional[int]]) -> None:
+    aa = filter(None, a)
+    reveal_type(aa)  # N: Revealed type is 'typing.Iterator[builtins.int*]'
+    bb = filter(lambda x: x is not None, a)
+    reveal_type(bb)  # N: Revealed type is 'typing.Iterator[Any]'
+    cc = filter(is_int, a)
+    reveal_type(cc)  # N: Revealed type is 'typing.Iterator[builtins.int]'
+    # TODO: Make this work -- this matches the third overload instead of the second
+    dd = filter(non_tg, a)
```
JukkaL, line 274: What if you switch the order of the second and third overload items?

mypy/expandtype.py
```diff
@@ -96,8 +96,12 @@ def visit_type_var(self, t: TypeVarType) -> Type:
             return repl
 
     def visit_callable_type(self, t: CallableType) -> Type:
+        extra = {}  # type: Dict[str, Any]
+        if t.type_guard is not None:
+            extra['type_guard'] = t.type_guard.accept(self)
         return t.copy_modified(arg_types=self.expand_types(t.arg_types),
-                               ret_type=t.ret_type.accept(self))
+                               ret_type=t.ret_type.accept(self),
+                               **extra)
```
JukkaL, line 104: Using `**extra` could be slow here, since it's unoptimized in mypyc. Why not pass `None` here if no type guard is defined, avoiding the use of `**extra` and the creation of a dictionary?

JukkaL (approved):
Thanks for the updates! Looks good now. Left a few minor comments, feel free to merge once you've addressed them (or decided that they are not worth it).

gvanrossum:
Thanks, I will merge after the test pass. I hope this can be added to the 0.800 release. (Even if it's not perfect, it shouldn't change anything for users who don't use `TypeGuard`.)

The main significant issue is that overloads using a type guard as higher-order function don't work right. I'll file a separate issue for that.

JukkaL:
Sorry, this will miss 0.800, but we are hoping to make another release fairly soon after 0.800 that switches to modular typeshed and this will be included

mortoray:
>  This may not be a very commonly needed feature, but in cases where this is useful the current alternatives are all quite bad.

For reference, custom type guards are extremely useful in any testing library as well as any data validation and/or file loading.  In those domains type validation functions are frequently used. I think we'll find this feature being used everywhere once available.

I'll be happy when this is available!

PR #9866, opened 2020-12-31
[mypyc] always add implicit None return type to __init__ method

thomasjohns:
### Description

Fixes https://github.com/mypyc/mypyc/issues/792.

Always give `__init__` an implicit `None` return type so that programs like
```
class C:
    def __init__(self):  # missing `-> None` annotation
        pass
```
compile.

## Test Plan

Included a test program that doesn't compile with the existing application code.

### DIFF
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -683,10 +683,6 @@ class FuncDef(FuncItem, SymbolNode, Statement):
     def name(self) -> str:
         return self._name
 
-    @property
-    def is_module_level_func(self) -> bool:
-        return self.info == FUNC_NO_INFO
-
     def accept(self, visitor: StatementVisitor[T]) -> T:
         return visitor.visit_func_def(self)
 
--- a/mypy/nodes.py
+++ b/mypy/nodes.py
@@ -683,6 +683,10 @@ class FuncDef(FuncItem, SymbolNode, Statement):
     def name(self) -> str:
         return self._name
 
+    @property
+    def is_module_level_func(self) -> bool:
+        return self.info == FUNC_NO_INFO
+
     def accept(self, visitor: StatementVisitor[T]) -> T:
         return visitor.visit_func_def(self)
 
--- a/mypy/suggestions.py
+++ b/mypy/suggestions.py
@@ -437,7 +437,7 @@ class SuggestionEngine:
         if self.no_errors and orig_errors:
             raise SuggestionFailure("Function does not typecheck.")
 
-        is_method = bool(node.info) and not node.is_static
+        is_method = not node.is_module_level_func and not node.is_static
 
         with strict_optional_set(graph[mod].options.strict_optional):
             guesses = self.get_guesses(
--- a/mypy/suggestions.py
+++ b/mypy/suggestions.py
@@ -437,7 +437,7 @@ class SuggestionEngine:
         if self.no_errors and orig_errors:
             raise SuggestionFailure("Function does not typecheck.")
 
-        is_method = not node.is_module_level_func and not node.is_static
+        is_method = bool(node.info) and not node.is_static
 
         with strict_optional_set(graph[mod].options.strict_optional):
             guesses = self.get_guesses(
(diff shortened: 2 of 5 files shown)


mypyc/irbuild/mapper.py
```diff
@@ -132,7 +132,11 @@ def fdef_to_sig(self, fdef: FuncDef) -> FuncSignature:
         else:
             # Handle unannotated functions
             arg_types = [object_rprimitive for arg in fdef.arguments]
-            ret = object_rprimitive
+            # We at least know the return type for __init__ will be None.
+            if fdef.name == '__init__':
```
JukkaL, line 136: Could `fdef` be a module-level function with the (confusing) name `__init__`? Add test for this. We should allow top-level functions to return any values, even if they are named `__init__` for some reason. 

(You should be able to compare the `FuncDef` `info` attribute against `FUNC_NO_INFO` to check whether it's a method.)

JukkaL (commented):
Thanks for the fix! Looks good, just found one edge case that may require a minor tweak.

mypyc/irbuild/mapper.py
thomasjohns: Oh nice catch this will not work for module-level functions named `__init__`! I will add a check and test for this. Thank you for the `FUNC_NO_INFO` tip.

mypy/nodes.py
```diff
@@ -683,6 +683,10 @@ def __init__(self,
     def name(self) -> str:
         return self._name
 
+    @property
+    def is_module_level_func(self) -> bool:
+        return self.info == FUNC_NO_INFO
```
thomasjohns, line 688: @JukkaL my instinct was to "hide" the use of `FUNC_NO_INFO`. I'm not sure if that was a good instinct.

mypyc/test-data/run-classes.test
```diff
@@ -374,6 +374,11 @@ o = object()
 c1.d = o
 assert c1.d is o
 
+[case testInitMethodWithMissingNoneReturnAnnotation]
+class C:
+    def __init__(self):
+        pass
```
JukkaL, line 380: Also try constructing an instance of `C` (just make sure it produces a result).

mypy/nodes.py
JukkaL: Actually I think that this is unnecessary (see below).

mypy/suggestions.py
```diff
@@ -437,7 +437,7 @@ def get_suggestion(self, mod: str, node: FuncDef) -> PyAnnotateSignature:
         if self.no_errors and orig_errors:
             raise SuggestionFailure("Function does not typecheck.")
 
-        is_method = bool(node.info) and not node.is_static
+        is_method = not node.is_module_level_func and not node.is_static
```
JukkaL, line 440: Let's leave this as it was, since using an `info` attribute as a boolean is a common idiom.

mypyc/irbuild/mapper.py
```diff
@@ -132,7 +132,12 @@ def fdef_to_sig(self, fdef: FuncDef) -> FuncSignature:
         else:
             # Handle unannotated functions
             arg_types = [object_rprimitive for arg in fdef.arguments]
-            ret = object_rprimitive
+            # We at least know the return type for __init__ methods will be None.
+            is_init_method = fdef.name == '__init__' and not fdef.is_module_level_func
```
JukkaL, line 136: I think that you can just use `... and bool(fdef.info)`. Your suggestion to use a property is reasonable, but we have many checks where we do the equivalent of `if fdef.info` so it would add some inconsistency.

JukkaL (commented):
Thanks for the updates! This looks correct now. Left some some comments about style. (Sorry, my original suggestion to use `FUNC_NO_INFO` was a bad idea.)

mypyc/irbuild/mapper.py
thomasjohns: Okay thank you that makes sense. Applying this update and the suggestions above now.

JukkaL (approved):
Looks good now! Thanks for updates. This fixes a major issue.

PR #9874, opened 2021-01-03
PEP 612: hide from those who may seek to use it

hauntsaninja:
This prevents users from hitting asserts before PEP 612 support is fully
implemented.


hauntsaninja:
Sorry for the trouble you ran into, @gvanrossum :-( This flag will address Jukka's / your concern in https://github.com/python/mypy/pull/9422#issuecomment-753568200

gvanrossum (approved):
Thanks -- although I would rather see it work. :-) I spent some time trying to help but I would have to recover a lot of lost knowledge about how the checker works.

hauntsaninja:
Me too! I think I'll have more time soon + I've struck off most of the higher priority items off my list

PR #9877, opened 2021-01-05
[stubgenc] fixes and typos

sizmailov:
### Description

This PR consist of series of minor changes to `stubgenc`, including typos in comment and (unfortunately) in code, one bug fix in docstring parsing and minor tweak to generate a better stubs for pybind11 modules. Hopefully commit messages are explanatory enough. 

This might be vague PR of small but rather disconnected changes, let me know if this must be split in smaller once. 

## Test Plan

So far this PR does not include test, therefore is marked as draft. 
Locally I've additionally tested stubs generation for a small pybind11-based project (https://github.com/sizmailov/pybind11-project-example). As far as I can tell currently there is no tests in CI against actual pybind module. Not sure if it's desirable to have one. 

Please let me know if tests for this changes are required.


JukkaL:
Thanks for the PR and your attention to detail! 

It's okay to have all of these changes in a single PR, since they are all pretty simple. 

Having tests against a pybind module could be nice, assuming no extra big dependencies are needed and these tests aren't complex or slow. I'd prefer if the dependency could be included in `test-requirements.txt`, for consistency.

.github/workflows/test.yml
```diff
@@ -47,3 +47,33 @@ jobs:
       run: tox -e ${{ matrix.toxenv }} --notest
     - name: test
       run: tox -e ${{ matrix.toxenv }} --skip-pkg-install
+
+  stubgenc:
+    # Check stub file generation for a small pybind11 project
+    # (full text match is required to pass)
```
sizmailov, line 53: I'm pretty new to gh-actions, and I'm not sure if it's the right place for the check.

test-requirements.txt
```diff
@@ -15,3 +15,7 @@ py>=1.5.2
 virtualenv<20
 setuptools!=50
 importlib-metadata==0.20
+# stubgen compatibility with pybind
+pybind11
+numpy
+git+https://github.com/sizmailov/python_example.git@mypy-stubgen-test
```
sizmailov, line 21: I've took https://github.com/pybind/python_example as a starting point and added a number of peculiar examples to test capabilities of stubgen (spoiler: currently it's quite far from perfect, I'll submit more changes in next PRs).

I'm not sure if it's appropriate to have a git-requirement with link to private repo. The dependency is small and probably it's fine to copy it as whole (just needed parts) somewhere in mypy.

sizmailov (commented):
I've added basic check for generated stubs for real pybind11 project. 

The check is full-text match. This is probably imperfect but I don't immediately see what alternative we have.

The C++ code is included as external dependency to a specific branch to my own repo. This is obviously cannot be merged as is. Please tell how to deal with it. Setting up "official" repo seems to be overkill, so I'm inclined to full-text include of linked repo to mypy. Again, I don't have strong opinion here.

I've placed expected stubs under `test-data/stubgen/`. Seems fine to me, but could be wrong.

JukkaL:
Hmm the test set up feels rather complex and I can see it becoming painful to maintain. Also it would be nice if we can avoid numpy as a dependency, since it's fairly big and might slow down CI. 

Here are some alternatives (which aren't perfect though). Let me know what you think about them.

1. Low-tech, manual tests: Add a script to `misc/stubgenc-extra-tests/`, for example, that installs the deps (here we can depend on numpy etc.), builds and installs the wheel and runs the tests. We wouldn't have this running in CI, but reviewers can ask contributors to run this manually if they modify stubgenc.
2. Upload binary wheels for an example package that uses pybind11 to PyPI (for at least one platform) and add a regular PyPI dependency to `test-requirements.txt` (possibly conditional). The test can now be run as part of the normal test suite (skipped on non-supported platforms), and no build step should be needed.
3. Don't bother with having tests in the mypy repository.
4. Only write unit tests that don't need pybind11.

For option 2, you can pin a well-known, pre-existing package that uses pybind11 (so that the dep is already available and can be expected to stay relatively up-to-date). You could also set up a new PyPI package dedicated to this test only. Some mypy core devs would probably need write access to upload new versions of the package, and the build and upload scripts could perhaps live in the mypy repository.

(I'm not very familiar with pybind11 so it's possible that my suggestions don't make much sense.)

sizmailov:
I think `stubgenc` test shouldn't be run on CI unless related code is affected. Therefore it would be ok to include numpy tests to this relatively rare CI job. 

I would go with option 2 since some tests are better than none, and automatic test are better than manual.

I've contacted pybind11 maintainers in gitter chat and I think there is a good chance to have a maintainable repo dedicated to test stubgen's ability to precisely render pybind11 modules. The only remaining question is where this project should be hosted (e.g. under `github.com/python/` or `github.com/pybind/`). 

This might take some time to compose a concise "pybind11 mypy demo" repo along with PyPI package, but it's surely worth it. 

(My test repo with occasional constructions is poorly structured and I think needs complete rewrite from scratch to serve as demo in any sense)

cc: @henryiii, @YannickJadoul

sizmailov:
I've moved the CI stubgenc test to a separate workflow restricted to changes in stubgenc-related files, so it won't run for most of the time. The workflow steps are repeated in `misc/test-stubgenc.sh` to facilitate local tests.

The demo project is updated to be shorter and more structured than what I had before. I think it's already good to be shipped as version `0.0.1`. At the moment it's still installed from github and this awaits change to PyPI-install after we agree where the demo project should be named&hosted and who is responsible to maintain it.

@henryiii How do you feel about `pybind/pybind11-mymy-demo`? 
@JukkaL How do you feel about `python/pybind11-mymy-demo`?

.github/workflows/test_stubgenc.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    - name: initialize submodules
+      run: git submodule update --init
+
+    - name: Setup 🐍 3.8
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+
+    - name: Prepare env
```
hauntsaninja, line 30: Let's just directly run misc/test-stubgenc.sh here

YannickJadoul:
When this would be moved under `pybind`, we should (probably) ask @wjakob, btw. But if there's some good arguments, I don't think that ought to be an issue :-)

sizmailov:
I've released `pybind11-mypy-demo==0.0.1` on PyPI with number of wheels (thanks to [cibuildwheel](https://cibuildwheel.readthedocs.io) project!).
I'll happily share write rights (or completely transfer) the project when we will finally decide where it belongs.

At this point I see this PR as complete. I've removed intermediate commits and rebased on current master (which already fixed one the typos from PR :-) 
The `test-stubgen.sh` tests against `0.0.1` version of demo project, so it's completely safe to merge.

misc/test-stubgenc.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# This script is expected to be run from root of the mypy repo
+
+# Install dependencies, demo project and mypy
+python -m pip install -r test-requirements.txt
+python -m pip install pybind11 numpy pybind11-mypy-demo
+python -m pip install pybind11-mypy-demo==0.0.1
```
JelleZijlstra, line 7: Why do you first install pybind11-mypy-demo without a specific version? Also, it might be better to pin the versions of pybind11 and numpy too.

sizmailov: That's a typo, thanks for spotting it!

In fact, `pybind11` and `numpy` can be removed. `pybind11`  is required to build from source only. `numpy` tests are not there (yet), so can be omitted as well. 

Future note (for myself): `numpy` should be listed in `pybind11-mypy-demo` runtime requirements, not here)

sizmailov:
Hi! Is there is anything left that blocks the PR? Do we wait for rights share/transfer, etc.? 

It's fine if it's just lack of time :) No rush

JelleZijlstra (approved):

PR #9881, opened 2021-01-05
Add a class attribute hook to the plugin system

### LINKED ISSUE #9645: get_attribute_hook does not trigger for class attributes, get_base_class_hook cannot resolve class attr types
**Bug Report**

I wasn't sure how to best phrase title and/or description for this, so edits/suggestions/renames are welcomed.

The `get_attribute_hook` plugin hook is not triggered for attributes/variables on classes. It seems to only trigger for attributes/variables on instances. Additionally, I seem to be unable to resolve types of class variables in the hook return by `get_base_class_hook`.

**To Reproduce**

Given the source file

```python
from typing import Dict, TYPE_CHECKING, Tuple, cast


class Field:
    pass


class FieldBuilder:
    def with_name(self) -> 'FieldBuilder':
        return self

    # Imagine arbitrarily more builder methods

    def build(self) -> Field:
        return Field()


class BaseMeta(type):
    def __new__(mcs, name: str, bases: Tuple[type, ...], attrs: Dict[str, object]) -> 'BaseMeta':
        for key, val in attrs.items():
            if isinstance(val, FieldBuilder):
                attrs[key] = val.build()
        return cast(BaseMeta, type.__new__(mcs, name, bases, attrs))


class Base(metaclass=BaseMeta):
    pass


class Concrete(Base):
    foo = FieldBuilder().with_name()  # Can have arbitrary chain of builder methods
    bar = 1


if TYPE_CHECKING:
    reveal_type(Concrete.foo)  # Would like this to show up as type "Field"
    reveal_type(Concrete.bar)  # Would like this to show up as type "int"
print(type(Concrete.foo))
print(type(Concrete.bar))
```

and the plugin

```python
from typing import Callable, Optional
from typing import Type as TypingType

from mypy.plugin import AttributeContext, ClassDefContext, Plugin
from mypy.types import Type as MypyType


class ExamplePlugin(Plugin):
    def get_attribute_hook(self, fullname: str) -> Optional[Callable[[AttributeContext], MypyType]]:
        print('ATTRIBUTE HOOK FOR', fullname)
        return None

    def get_base_class_hook(self, fullname: str) -> Optional[Callable[[ClassDefContext], None]]:
        if not fullname.endswith('.Base'):
            return None

        def hook(ctx: ClassDefContext) -> None:
            print('BASE CLASS HOOK FOR', fullname, ctx.cls.fullname)
            if ctx.cls.info.names['foo'].type is None:
                if ctx.api.final_iteration:
                    print('CANNOT DEFER ANY MORE', fullname)
                else:
                    print('DEFERRING', fullname)
                    ctx.api.defer()

        return hook

def plugin(_mypy_info: str) -> TypingType[Plugin]:
    return ExamplePlugin
```

**Expected Behavior**

I would expect that an `ATTRIBUTE HOOK FOR` line would be printed for the `Concrete.foo` and `Concrete.bar` accesses. Alternatively, I'd expect to be able to eventually get a non-`None` type for the `foo` attribute in the base class hook.

**Actual Behavior**

The `ATTRIBUTE HOOK FOR` is never printed for the `Concrete.*` accesses and we never get a non-`None` type for `foo` (we reach the final iteration).

If there's a way to accomplish the `Would like this to show up as type` comments with the existing plugin hook behaviors, that also works for me! Note, I've seen [other plugins](https://github.com/dropbox/sqlalchemy-stubs/blob/55470ceab8149db983411d5c094c9fe16343c58b/sqlmypy.py#L79-L91) use the `get_function_hook` to modify the return type and therefore the type of the class var, but I didn't see a way to know that the result of the function would be stored in a class var, since I wouldn't want *every* builder method to be automatically converted.

**Your Environment**

- Mypy version used: `mypy 0.800+dev.f220ce50725ed7f117832b9f03aed4d1a9508e00`
- Mypy command-line flags: `--show-traceback --warn-unused-configs --config-file <file>`
```ini
[mypy]
plugins = <my plugin>
strict = True
```
- Python version used: `3.7.6`

FuegoFro:
### Description

This adds a `get_class_attribute_hook` to plugins to modify attributes on *classes* (as opposed to the existing `get_attribute_hook`, which is for attributes on *instances*). The modifications and added tests were modeled off of https://github.com/python/mypy/commit/3acbf3fe78a61c19ff96754233ada453472004c4.

Fixes #9645.

## Test Plan

A new test is added which exercises the new functionality.


mypy/checkmember.py
```diff
@@ -778,6 +778,13 @@ def analyze_class_attribute_access(itype: Instance,
         if not mx.is_lvalue:
             result = analyze_descriptor_access(mx.original_type, result, mx.builtin_type,
                                                mx.msg, mx.context, chk=mx.chk)
+
+        # Call the class attribute hook before returning.
+        fullname = '{}.{}'.format(info.fullname, name)
+        hook = mx.chk.plugin.get_class_attribute_hook(fullname)
+        if hook:
+            result = hook(AttributeContext(get_proper_type(mx.original_type),
+                                           result, mx.context, mx.chk))
```
FuegoFro, line 787: I wasn't sure if the hook should be invoked for every exit point of this function. If so, I can pull the hook invocation into a wrapper function that calls into this one (or similar). I'm not familiar enough with this code to be confident where it would and would not make sense to use the hook, so advice is appreciated!

FuegoFro (commented):
Hey there! Let me know if you think this is something that is even worth having. It does solve a need for me that I don't think is otherwise possible currently (see the issue linked in the description), but I don't know how much extra surface area/ongoing maintenance it'll cause for you. Also let me know if there's something else I should be updating (documentation, maybe?). As mentioned in the description, I looked to the PR that added `get_function_signature_hook` to determine which files needed changing, and I didn't see anything around documentation changed there. Thanks in advance, and happy new year!

test-data/unit/check-custom-plugin.test
```diff
@@ -730,3 +730,15 @@ reveal_type(dynamic_signature(1))  # N: Revealed type is 'builtins.int'
...
 plugins=<ROOT>/test-data/unit/plugins/function_sig_hook.py
+
+[case testClassAttrPluginFile]
+# flags: --config-file tmp/mypy.ini
+
+class Cls:
+    attr = 'test'
+
+reveal_type(Cls.attr)  # N: Revealed type is 'builtins.int'
+reveal_type(Cls().attr)  # N: Revealed type is 'builtins.str'
```
JukkaL, line 741: Ideas for more test coverage:

* Test other exit points as well.
* Test using `x.attr` so that `x` has type `Type[Cls]`. 
* Test accessing another attribute that doesn't have a hook associated with.

JukkaL (commented):
Thanks for the PR! A plugin hook for class attributes seems like a reasonable thing to have.

mypy/checkmember.py
FuegoFro: Just so I understand, calling through to another function like this is fine, but doing the "opposite" is unoptimized? Though, I realize I wasn't super clear in my previous comment haha. To be a bit more explicit, what I was suggesting was to rename `analyze_class_attribute_access` to something like `analyze_class_attribute_access_inner` and then create a new
```python
def analyze_class_attribute_access(...):
    result = analyze_class_attribute_access_inner(...)
    fullname = '{}.{}'.format(info.fullname, name)
    hook = mx.chk.plugin.get_class_attribute_hook(fullname)
    if hook:
        result = hook(AttributeContext(get_proper_type(mx.original_type),
                                       result, mx.context, mx.chk))
```
Would that pattern be unoptimized, or is it functions/closures declared *within* a function that's unoptimized?

In any case, when looking through this a bit more it does seem there's maybe a few exit points we *don't* want to call the hook for (which I wanted to double check with you) in addition to the `None` return, namely these two cases, which might make the whole `_inner` idea above moot anyway:

https://github.com/python/mypy/blob/6e8c0cd3ef7a375e143faf237312004adcfca7d6/mypy/checkmember.py#L782-L789

Do we want to call the hook for these two exit points?

test-data/unit/check-custom-plugin.test
FuegoFro: Ah, those are good ones! Would you recommend putting those in the same test case or lean more toward making separate test cases?

FuegoFro (commented):
Thank you for the review! 😀 I'll get the extra exit points and extra tests up soon 🙂

mypy/checkmember.py
```diff
@@ -778,32 +783,39 @@ def analyze_class_attribute_access(itype: Instance,
         if not mx.is_lvalue:
             result = analyze_descriptor_access(mx.original_type, result, mx.builtin_type,
                                                mx.msg, mx.context, chk=mx.chk)
-        return result
+
+        return apply_class_attr_hook(mx, hook, result)
     elif isinstance(node.node, Var):
+        # TODO RIGHT NOW NOMERGE - is it okay to not modify this?
```
FuegoFro, line 789: I had a few questions in here (marked by these `TODO RIGHT NOW NOMERGE` comments/asserts), both around whether certain return values should be modified and also how to exercise some of these branches, would love some advice on both! 🙂

FuegoFro:
@JukkaL I've got a few questions that I'd love some help/feedback on when you have a bit to look things over 🙂

mypy/checkmember.py
```diff
@@ -778,32 +783,39 @@ def analyze_class_attribute_access(itype: Instance,
...
     if isinstance(node.node, TypeInfo):
-        return type_object_type(node.node, mx.builtin_type)
+        assert False, "TODO RIGHT NOW NOMERGE - No tests hit this, how to trigger?"
+        return apply_class_attr_hook(mx, hook, type_object_type(node.node, mx.builtin_type))
 
     if isinstance(node.node, MypyFile):
         # Reference to a module object.
-        return mx.builtin_type('types.ModuleType')
+        assert False, "TODO RIGHT NOW NOMERGE - No tests hit this, how to trigger?"
+        return apply_class_attr_hook(mx, hook, mx.builtin_type('types.ModuleType'))
```
JukkaL, line 806: We shouldn't run the hook here, since it wouldn't get triggered consistently. Some references to module attributes are resolved in semantic analysis, before type checking. The hook should be applied either always for never for each particular attribute.

mypy/checkmember.py
```diff
@@ -778,32 +783,39 @@ def analyze_class_attribute_access(itype: Instance,
...
 
     if (isinstance(node.node, TypeAlias) and
             isinstance(get_proper_type(node.node.target), Instance)):
-        return instance_alias_type(node.node, mx.builtin_type)
+        return apply_class_attr_hook(mx, hook, instance_alias_type(node.node, mx.builtin_type))
```
JukkaL, line 810: Similar to above, we probably shouldn't apply the hook here.

mypy/checkmember.py
```diff
@@ -778,32 +783,39 @@ def analyze_class_attribute_access(itype: Instance,
...
 
     if is_decorated:
         assert isinstance(node.node, Decorator)
         if node.node.type:
-            return node.node.type
+            assert False, "TODO RIGHT NOW NOMERGE - No tests hit this, how to trigger?"
+            return apply_class_attr_hook(mx, hook, node.node.type)
         else:
+            # TODO RIGHT NOW NOMERGE - is it okay to not modify this?
```
JukkaL, line 818: Yeah, let's not call the hook here.

test-data/unit/check-custom-plugin.test
JukkaL: It's fine to group related checks in a single test case (this makes tests run faster). These are probably fine in this test case, but if the test case gets pretty long (use your judgment) it may be better to split it.

mypy/checkmember.py
```diff
@@ -778,32 +783,39 @@ def analyze_class_attribute_access(itype: Instance,
...
             isinstance(get_proper_type(node.node.target), Instance)):
-        return instance_alias_type(node.node, mx.builtin_type)
+        return apply_class_attr_hook(mx, hook, instance_alias_type(node.node, mx.builtin_type))
 
     if is_decorated:
         assert isinstance(node.node, Decorator)
         if node.node.type:
-            return node.node.type
+            assert False, "TODO RIGHT NOW NOMERGE - No tests hit this, how to trigger?"
+            return apply_class_attr_hook(mx, hook, node.node.type)
```
JukkaL, line 816: Hmm I'm not sure if this is actually ever used. Just to be safe, we can apply the hook here, but you don't have to come up with a way to exercise this in a test case.

JukkaL (commented):
Thanks for the updates! Overall looks good. Answered various questions.

FuegoFro:
@JukkaL Thanks so much for the response! I *think* this should be ready to go now 🙂

FuegoFro:
@JukkaL is there anything else you think needs to be changed here before merging? 🙂

FuegoFro:
Heya @JukkaL just checking in to see if we can get this merged if there's nothing else needed here 🙂

FuegoFro:
@JukkaL friendly ping, would love to get this merged in 🙂

charmoniumQ:
Any update on this?

mezz:
Hi @JukkaL, I think this PR is ready. Can you give it a final review?

mypy/plugin.py
```diff
@@ -662,6 +662,25 @@ class Derived(Base):
...
+        defined (or Var.info.fullname for generated attributes).
+
+        For example:
+
+            class Cls:
+                x: Any
+
+            Cls.x
+
+        get_class_attribute_hook is called with '__main__.Cls.x'.
```
sobolevn, line 680:
```suggestion
        get_class_attribute_hook is called with '__main__.Cls.x' as fullname.
```

test-data/unit/check-custom-plugin.test
```diff
@@ -902,3 +902,74 @@ reveal_type(f())  # N: Revealed type is "builtins.str"
...
+
+[case testClassAttrPluginMetaclass]
+# flags: --config-file tmp/mypy.ini
+
+from typing import Any, Type
+class M(type):
+    attr = 'test'
+
+B: Any
+class Cls(B, metaclass=M):
```
sobolevn, line 957: I suggest to use one more test example with real `B`, because subtyping `Any` is not the same as subtyping a regular class.

test-data/unit/check-custom-plugin.test
```diff
@@ -902,3 +902,74 @@ reveal_type(f())  # N: Revealed type is "builtins.str"
...
+\[mypy]
+plugins=<ROOT>/test-data/unit/plugins/class_attr_hook.py
+
+[case testClassAttrPluginPartialType]
+# flags: --config-file tmp/mypy.ini
+
+class Cls:
+    attr = None
+    def f(self) -> int:
+        return Cls.attr + 1
```
sobolevn, line 971: What about `self.__class__.attr`? Should we support that?

I think we can also check that `c: Type[Cls]; c.attr` works correctly.

sobolevn (changes requested):
Thank you! This mostly looks good to me, but I think that we should add a bit more tests. And don't forget about docs: https://mypy.readthedocs.io/en/stable/extending_mypy.html

FuegoFro (commented):
Thanks for the review! I added docs, here's a screenshot for convenience:
image

Let me know if there's anything else I can do to help get this PR merged! After ~a year, it'd be great to stop maintaining our own fork of mypy 🙏

hauntsaninja (approved):
Thanks for your patience, let's make this happen.

It wasn't clear to me why e.g. we don't apply the hook to things like type aliases, but then I saw Jukka's comments. I believe he's referring to stuff like:
https://github.com/python/mypy/blob/fc335cb16315964b923eb1927e3aad1516891c28/mypy/semanal.py#L4106

FuegoFro:
Excellent, thank you so much for the merge! 😀

PR #9882, opened 2021-01-06
Don't expand global variables in body of a function with constrained type variables

ilevkivskyi:
See first test case for the scenario where this causes issues (the second test case is already handled elsewhere and just added for completeness). This is an old issue, but it is important to fix it now because recent changes to PEP 484 about re-export force people to use function aliases.

### DIFF
--- a/mypy/treetransform.py
+++ b/mypy/treetransform.py
@@ -20,7 +20,7 @@ from mypy.nodes import (
     YieldFromExpr, NamedTupleExpr, TypedDictExpr, NonlocalDecl, SetComprehension,
     DictionaryComprehension, ComplexExpr, TypeAliasExpr, EllipsisExpr,
     YieldExpr, ExecStmt, Argument, BackquoteExpr, AwaitExpr, AssignmentExpr,
-    OverloadPart, EnumCallExpr, REVEAL_TYPE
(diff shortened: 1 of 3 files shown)


mypy/treetransform.py
```diff
@@ -358,7 +358,12 @@ def copy_ref(self, new: RefExpr, original: RefExpr) -> None:
         new.fullname = original.fullname
         target = original.node
         if isinstance(target, Var):
-            target = self.visit_var(target)
+            # Do not transform references to global variables.
+            # TODO: is it really the best solution? Should we instead make
+            # a trivial copy, or do something completely different? See
+            # testGenericFunctionAliasExpand for an example where this is important.
+            if original.kind != GDEF:
+                target = self.visit_var(target)
```
JukkaL, line 366: If you remove `visit_mypy_file` from the class (I think that it's unused), this will be completely reasonable, since this will be reference to something outside the AST node we are transforming.

JukkaL (approved):

mypy/treetransform.py
ilevkivskyi: OK.

JukkaL: Maybe also update the docstring that this can only be used to transform functions or classes, not top-level statements and such.

Issue #9761, opened 2020-11-26
Supporting Python 3.9


hauntsaninja:
The next release needs to support Python 3.9

Here's my checklist of what's left:
- [x] PEP 585 support (mostly #9564)
- [x] PEP 614 testing (#9759)
- [x] Merge and test new build process for 3.9 wheels (https://github.com/python/mypy/pull/9570, https://github.com/mypyc/mypy_mypyc-wheels/pull/11)
- [x] Add 3.9 to trove classifiers (https://github.com/python/mypy/pull/9887)
- [x] Improved documentation of PEP 563 + PEP 585 (#9763)
- [ ] mypy_extensions support for 3.9 (https://github.com/python/mypy_extensions/pull/20 / https://github.com/python/mypy_extensions/issues/21)

To clarify the state of support for users stumbling upon this issue, mypy 0.790 will mostly work with Python 3.9.
There are two caveats to running mypy with 3.9:
1) you won't be able to use `list[int]` syntax without mypy complaining,
2) mypy will be slower since we haven't released mypyc compiled wheels for 3.9.

Note that typeshed in 0.790 has pretty good support for Python 3.9, so using new stdlib functionality should work fine.

PS: if you're getting `error: syntax error in type comment` with Python 3.9, it's because you're using a really old mypy and you should upgrade to a recent release.

intgr:
> The next release needs to support Python 3.9

I wouldn't make all of these hard requirements. I'd very much like to see a mypy release with binary wheels for 3.9 even before all of these items are addressed.

gvanrossum:
Well, we still have the 0.790 release branch so we could just create wheels for that and push those. (Or make whatever tweak are necessary on that branch and push 0.791.)

hauntsaninja:
@intgr Agreed that I wouldn't block an otherwise scheduled release on this, but there isn't currently a scheduled release / at this point the checklist^ isn't much additional effort to complete.

If someone with the ability to release is interested in adding 3.9 wheels to 0.790, you'll need to cherry pick #9552 and #9562 (and merge the aforementioned mypyc wheel repo change). Note mypyc is broken in the 0.790 sdist so you should cherry pick #9587 as well (see https://github.com/python/mypy/issues/9290#issuecomment-711131865).

gvanrossum:
Here's hoping someone wants to take a stab!

JukkaL:
[...]

PR #9887, opened 2021-01-07
Add Python 3.9 to trove classifiers

hauntsaninja:
Since my checklist in https://github.com/python/mypy/issues/9761 is
almost complete, it's time to check off this box as well.


JukkaL (approved):
Excellent, supporting Python 3.9 is very important.

PR #9889, opened 2021-01-07
Fix type errors stemming from getattr

### LINKED ISSUE #9888: [stubgen] Crash on boost::python C extension
Hey team ! First of all, happy new year, and thanks for your great work !

**Crash Report**

When running `stubgen` on a compiled python module using boost::python, stubgen crashes.

See below for a minimal reproducible example. 

**Traceback**

```
  File "/install/bin/stubgen", line 8, in <module>
    sys.exit(main())
  File "mypy/stubgen.py", line 1564, in main
  File "mypy/stubgen.py", line 1457, in generate_stubs
  File "mypy/stubgenc.py", line 64, in generate_stub_for_c_module
  File "mypy/stubgenc.py", line 293, in generate_c_type_stub
  File "mypy/stubgenc.py", line 164, in generate_c_function_stub
TypeError: str object expected; got None
```

**To Reproduce**

I was able to break down a minimal example that breaks (thanks to [TNG/boost_python_examples](https://github.com/TNG/boost-python-examples) ). You find in the archive a sample C++ file and a basic CMakeLists to compile it

[example.tar.gz](https://github.com/python/mypy/files/5780439/example.tar.gz)

After extracting both files into a folder, build the library (requires python and libboost-python-dev)
```sh
mkdir build && cd build
cmake ..
make
```

Run stubgen on the generated lib
, sh
```sh
stubgen -p example_pywrap --search-path .
```

...and it crashes

**Your Environment**

- Mypy version used: 0.790
- Stubgen command-line flags:
- Python version used: 3.8
- Operating system and version: Ubuntu 20.04

**Remarks**

This closely looks like the bug fixed in #8888. And indeed, **if I install the mypy build mentionned in this PR, the problem is fixed.** (see this message: https://github.com/python/mypy/pull/8888#issuecomment-665253096)
- stubgen does not crash
- the output file `out/example_pywrap.pyi` exists and contains valid content

What I don't understand is that the PR has been merged 6 months ago, the resulting commit seems to be in the 0.790 tag (see commit https://github.com/python/mypy/commit/ea9291de044ee8930abbc58fded23fe413b91210), so I should not see the crash on mypy 0.790, shoud I ?

[EDIT] I tested again with the latest master commit of this repo (`mypy-0.800+dev.5f5d90eee23b809185145f6da7b18a3ca41b4879`) and it works (no crashes). So:
- If the crash was indeed fixed by #8888, can someone explain me why it was not in v0.790 even if the commit view mentions it ?
image
- Otherwise, I faced a different bug that was fixed after v0.790. 

Either way, is there somewhere I can go to see the ETA for 0.800, so I can know if I need how long my work-around is supposed to hold.

hauntsaninja:
Fixes #9888

I applied the following patch to typeshed and ran self check:
```
+@overload
+def getattr(__o: Any, name: str, __default: None) -> Optional[Any]: ...
+@overload
 def getattr(__o: Any, name: str, __default: Any = ...) -> Any: ...
 ```

### DIFF
--- a/mypy/moduleinspect.py
+++ b/mypy/moduleinspect.py
@@ -45,7 +45,7 @@ def get_package_properties(package_id: str) -> ModuleProperties:
         package = importlib.import_module(package_id)
     except BaseException as e:
         raise InspectError(str(e)) from e
-    name = getattr(package, '__name__')
+    name = package.__name__
     file = getattr(package, '__file__', None)
     path = getattr(package, '__path__', None)  # type: Optional[List[str]]
     if not isinstance(path, list):
--- a/mypy/moduleinspect.py
+++ b/mypy/moduleinspect.py
@@ -45,7 +45,7 @@ def get_package_properties(package_id: str) -> ModuleProperties:
         package = importlib.import_module(package_id)
     except BaseException as e:
         raise InspectError(str(e)) from e
-    name = getattr(package, '__name__', None)
+    name = getattr(package, '__name__')
     file = getattr(package, '__file__', None)
     path = getattr(package, '__path__', None)  # type: Optional[List[str]]
     if not isinstance(path, list):
--- a/mypy/moduleinspect.py
+++ b/mypy/moduleinspect.py
@@ -45,7 +45,7 @@ def get_package_properties(package_id: str) -> ModuleProperties:
         package = importlib.import_module(package_id)
     except BaseException as e:
         raise InspectError(str(e)) from e
-    name = getattr(package, '__name__', package_id) 
+    name = getattr(package, '__name__', package_id)
     file = getattr(package, '__file__', None)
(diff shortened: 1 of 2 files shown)


JukkaL (commented):
`getattr` is being a pain here. Maybe typeshed should use an overload so that `getattr(o, a, None)` produces `Optional[Any]`?

hauntsaninja:
Thanks, fixed! Yeah, that's the overload I used to find these. You can also use the overload `def getattr(__o: Any, name: str, __default: _T) -> Union[_T, Any]: ...` for a slightly more aggressive change.

Unfortunately, mypy_primer finds both potential typeshed changes somewhat problematic so I'm going to hold off (at least until we're back to having a release cycle)

JukkaL (approved):

PR #9891, opened 2021-01-08
add modules and packages keys to config

### LINKED ISSUE #9883: Add `packages` and `modules` keys to config file format
**Feature**

Offer configuration variables for `packages` and `modules` to match the command-line flags, much like `files` matches listing files on the command line.  Intended to work exactly like the command line flags, with the exception that they would not override a `files` key and having `files` and either of the new keys in the same config file would create a conflict.

Usage meant to be something like this (actual code this would work with linked below):

```ini
[mypy]
python_version = 3.7
packages=spack,builtin.packages,builtin_mock.packages
mypy_path=bin,lib/spack,lib/spack/external,var/spack/repos/builtin
```

To match a command line like `mypy -p spack -p builtin.packages -p builtin_mock.packages`

**Pitch**

The main use comes from, [spack](GitHub.com/llnl/spack), a project I'm adding type checking to.  It's a package manager where the libraries, tool and all package definition files are written in python.  Having mypy check the library without arguments is simple, I can just write a `files` key that covers all files and it will work fine, but the packages are in a structure like this:

```
./var/spack/repos/
├── builtin
│   ├── packages
│   │   ├── 3proxy
│   │   │   └── package.py
│   │   ├── abduco
│   │   │   └── package.py
│   │   ├── abi-compliance-checker
│   │   │   └── package.py
... etc.
```

If I use a `files` key for these, they are all treated as duplicate `package` modules, and mypy errors out (probably appropriately).  They can however be treated as a package, and mypy works quite well on the library and the whole package tree in that mode (especially after the recursive checking fixes).  The trick now is that I can't make running `mypy` check the packages tree without arguments or wrapping or similar.

I'd like to add support for specifying these in the configuration file to bring them up to the same level of support as listing files, as mentioned in PR #9834.

trws:
### Description

Fixes #9883 

This adds configuration file support for specifying packages and
modules.  I tried to preserve existing semantics between the config file
and command line.  If the command line specifies anything, it overrides
the config file, if the config file contains both package/module and
files keys, the later key is ignored.

## Test Plan

Added a new version of the package and module command line test that retrieves the values from the configuration file.


mypy/config_parser.py
```diff
@@ -278,6 +280,13 @@ def parse_section(prefix: str, template: Options,
                 if 'follow_imports' not in results:
                     results['follow_imports'] = 'error'
         results[options_key] = v
+        if key in ('files', 'packages', 'modules'):
+            if all(('packages' in results.keys() or 'modules' in results.keys(),
+                    results.get('files'))):
```
JukkaL, line 285: Style nit: This was a bit difficult to understand. I think it would better to use `and` instead of `all()`, and you can just use `'packages' in results`, without using `keys()`, I think.

mypy/config_parser.py
```diff
@@ -278,6 +280,13 @@ def parse_section(prefix: str, template: Options,
...
+                print("May only specify one of: module/package, files, or command. Ignoring key.",
```
JukkaL, line 286: What does 'command' here refer to?

Please include the name of the key in the error message. Currently it may not be clear what is being ignored.

JukkaL (commented):
Nice, this fixes an inconsistency in the features supported by the command line vs. the config file.

Overall looks good, just a few nits. Thanks for also updating the docs!

mypy/config_parser.py
trws: Good point, and I'll remove command, that's part of the corresponding CLI error message that I preserved but doesn't apply here.

mypy/config_parser.py
trws: That's fair.  I started with something that generated a list and tested it and didn't refactor it out as far as I should have.  Fixed.

trws:
Anything further here? I'd be happy to make other tweaks if necessary.

hauntsaninja:
Thanks for working on this, it looks like this was superseded by #13404

PR #9894, opened 2021-01-09
[mypyc] Use faster METH_FASTCALL wrapper functions on Python 3.7+

JukkaL:
Implement faster argument parsing based on METH_FASTCALL on supported
Python versions. 

Use `vgetargskeywordsfast` extracted from Python 3.9 with some modifications:
* Support required keyword-only arguments, `*args` and `**kwargs`
* Only support the 'O' type (to reduce code size and speed things up)

The modifications are very similar to what we have in the old-style
argument parsing logic.

The legacy calling convention is still used for `__init__` and `__call__`. I'll add
`__call__`  support in a separate PR. I haven't looked into supporting `__init__`
yet.

Here are some benchmark results (on Python 3.8)
* keyword_args_from_interpreted: 3.5x faster than before
* positional_args_from_interpreted: 1.4x faster than before

However, the above benchmarks are still slower when compiled. I'll continue 
working on further improvements after this PR.

Fixes mypyc/mypyc#578.

### DIFF
--- a/mypyc/codegen/emitwrapper.py
+++ b/mypyc/codegen/emitwrapper.py
@@ -55,6 +55,16 @@ def wrapper_function_header(fn: FuncIR, names: NameGenerator) -> str:
             name=fn.cname(names))
 
 
+def make_arg_groups(args: List[RuntimeArg]) -> List[List[RuntimeArg]]:
+    """Group arguments by kind."""
+    return [[arg for arg in args if arg.kind == k] for k in range(ARG_NAMED_OPT + 1)]
+
+
+def reorder_arg_groups(groups: List[List[RuntimeArg]]) -> List[RuntimeArg]:
+    """Reorder argument groups to match their order in a format string."""
+    return groups[ARG_POS] + groups[ARG_OPT] + groups[ARG_NAMED_OPT] + groups[ARG_NAMED]
+
+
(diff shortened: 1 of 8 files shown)


mypyc/codegen/emitwrapper.py
```diff
@@ -40,13 +102,92 @@ def generate_wrapper_function(fn: FuncIR,
...
+    real_args = list(fn.args)
+    if fn.class_name and not fn.decl.kind == FUNC_STATICMETHOD:
+        arg = real_args.pop(0)
+        emitter.emit_line('PyObject *obj_{} = self;'.format(arg.name))
+
+    # Need to order args as: required, optional, kwonly optional, kwonly required
+    # This is because CPyArg_ParseStackAndKeywords format string requires
+    # them grouped in that way.
+    groups = make_arg_groups(real_args)
+    reordered_args = reorder_arg_groups(groups)
```
ilevkivskyi, line 134: These preparatory steps here are very similar to the legacy wrapper below. Would it make sense to factor them out in a helper?

ilevkivskyi (approved):
LGTM, just one suggestion.

mypyc/codegen/emitwrapper.py
JukkaL: Refactored some of the shared code. I didn't share everything, since I'm planning further changes that may only be relevant for new-style wrapper functions.

PR #9900, opened 2021-01-12
Fixed Any and Dict missing from stubgen imports list

AWhetter:
### Description

This fixes an issue introduced in https://github.com/python/mypy/pull/9088/files where `Any` and `Dict` are missing from the list of default imports that stubgen can perform, due to a missing comma.

## Test Plan

Because this is such a small change I just did a quick manual test by using stubgen on a module that uses both `Any` and `Dict`.


gvanrossum (approved):
Great find! This is a bit of a misfeature in Python... :-(

PR #9903, opened 2021-01-12
[stubgenc] Render a bit better stubs

### LINKED ISSUE #9070: Ignore generic type signature in pybind11 functions and methods with overloads
I don't know if this is the correct place to address this issue, but I know that there's the expertise here to be able to discuss the correct approach to the problem. I've written this as a feature request, but there may be better approaches to the problem.

## Use Case
I have a C++ module that's exposed to Python with pybind11. In this module I have a class with a method that has multiple types. To use an example from the pybind11 documentation:

```cpp
py::class_<Pet>(m, "Pet")
   .def(py::init<const std::string &, int>())
   .def("set", (void (Pet::*)(int)) &Pet::set, "Set the pet's age")
   .def("set", (void (Pet::*)(const std::string &)) &Pet::set, "Set the pet's name");
```

This produces docstrings that look like the following:

```python
>>> help(example.Pet)

class Pet(__builtin__.object)
 |  Methods defined here:
 |
 |  __init__(...)
 |      __init__(self: Pet, arg0: str, arg1: int) -> None
 |
 |  set(...)
 |      set(*args, **kwargs) -> Any
 |      Overloaded function.
 |
 |      1. set(self: Pet, arg0: int) -> None
 |
 |      Set the pet's age
 |
 |      2. set(self: Pet, arg0: str) -> None
 |
 |      Set the pet's name
```

## Current Behaviour
When running stubgen against this module, the generated type stubs look like the following:

```python
from typing import overload

class Pet:
    def __init__(self: Pet, arg0: str, arg1: int) -> None: ...
    @overload
    def set(self: Pet, arg0: int) -> None: ...
    @overload
    def set(self: Pet, arg0: str) -> None: ...
    @overload
    def set(*args, **kwargs) -> Any: ...
```

The inclusion of the last overload essentially makes the type definition of the method useless because it'll accept anything.
The reason that this feels to me like an issue for stubgen to address is because I think that pybind11 is doing the right thing. The inclusion of `set(*args, **kwargs) -> Any` makes sense because tools like Sphinx need to be given a valid type signature and those tools expect that signature on the first line of the docstring.

## Expected Behaviour
When running stubgen against this module, the generated type stubs should look like the following:

```python
from typing import overload

class Pet:
    def __init__(self: Pet, arg0: str, arg1: int) -> None: ...
    @overload
    def set(self: Pet, arg0: int) -> None: ...
    @overload
    def set(self: Pet, arg0: str) -> None: ...
```

Note that the `set(*args, **kwargs) -> Any` type signature has been ignored.

## Implementation
If we decide that getting stubgen to ignore this specific type signature is the right thing to do, how should we do it? Ignoring `*args, *kwargs` here is not correct because having `*args` and `**kwargs` could be part of a valid signature: https://github.com/python/mypy/blob/007b7af9f96df9d23744d93fdf54e83bcdc01630/mypy/stubdoc.py#L193-L202
We could look for the "Overloaded signature." line and ignore the type signature in only that case. It feels brittle, but I don't know how else we can determine whether this situation exists.

Do we need to worry about the case where someone might decide to write the following:

```python
"""func(*args, **kwargs) -> Any
Do a thing.

func() -> None
Do something else.
```

## Versions

```
mypy 0.780
Python 3.7.7
```

sizmailov:
### Description

PR contains a couple of improvements of stubgenc generator:
 - render nested classes
 - shorten full-qualified types in properties
 - show concrete types of attributes
 - swap class/function render order (due to possible dependency on classes in functions arguments/return type)
 - remove redundant `(*args, **kwargs)` overloads in pybind11 modules (closes #9070)
 - render pybind11 static properties as `x: ClassVar[T]`
 - render rw-properties as annotated class attributes (e. g. `x: T`)

## Test Plan

Tests against `pybind11-mypy-demo` project are included


.github/workflows/test_stubgenc.yml
```diff
@@ -0,0 +1,31 @@
...
+
+on:
+  push:
+    branches: [master]
+    tags: ['*']
+  pull_request:
+    paths:
+    - 'mypy/stubgenc.py'
+    - 'mypy/stubdoc.py'
+    - 'test-data/stubgen/**'
```
JukkaL, line 11: This should also run for changes in `misc/test-stubgenc.sh`.

sizmailov: I'll address this in #9877

sizmailov:
Gentle ping

sizmailov:
Gentle ping

sizmailov:
Is this PR too big? Should it be split into smaller pieces, which can be reviewed? 

The individual changes are rather small and related, but are decoupled to some degree, so they can be applied one after another if needed. I thought it would be faster to ship them as one, rather than a series of dependent PRs. 

Please let me know if I can do anything to push the changes forward.

PR #9904, opened 2021-01-12
Fix an `issubclass` failure for protocols with overloaded methods

BvB93:
### Description

Mypy currently rejects `issubclass` checks for protocols containing non-method members. 
What I noticed is that this includes overloaded functions as well, the latter being incorrectly 
identified as a "non-method" due to a missing `isinstance(typ, mypy.types.Overloaded)` check. 

This PR attempts to remedy aforementioned issue.

## Examples

The `issubclass` behavior prior to this PR:
```python
from typing import Protocol, overload, Sequence, runtime_checkable

@runtime_checkable
class A(Protocol):
    @overload
    def get_item(self, a: int) -> int: ...
    @overload
    def get_item(self, a: slice) -> Sequence[int]: ...

# test.py:10:1: error: Only protocols that don't have non-method members can be used with issubclass()  [misc]
# test.py:10:1: note: Protocol "A" has non-method member(s): get_item
# Found 1 error in 1 file (checked 1 source file)
issubclass(int, A)
```

## Test Plan

A new `issubclass` test has been added for overload-containing protocols.


test-data/unit/check-protocols.test
```diff
@@ -2219,6 +2219,17 @@ if issubclass(cls, P):
...
+@runtime_checkable
+class POverload(Protocol):
+    @overload
+    def meth(self, a: int) -> float: ...
+    @overload
+    def meth(self, a: str) -> Sequence[float]: ...
+    def meth(self, a):
+        pass
+
+reveal_type(issubclass(int, POverload))  # N: Revealed type is 'builtins.bool'
```
JukkaL, line 2232: Could you also add a concrete class (e.g., `COverload`) that has a suitable overloaded method, and test that `issubclass` can be used to narrow down from `Type[Union[COverload, E]]`, similar to above around line 2218?

JukkaL (approved):
Thanks for the fix! 

Left an (optional) request to update the test case, otherwise looks good.

test-data/unit/check-protocols.test
BvB93: So while trying the add the test you proposed I seem to have stumbled across another bug.

The `if` block successfully narrows down the `Union` to `Type[COverload]` but the `else` block
on the other hand fails to do the same with `E`, revealing the type as the original `Union`.

Would you happen to know where in mypy the type inference of `if` / `else` statements is handled?
I imagine that change similar to `non_method_protocol_members()` will have to be made there,
but I'm not quite sure where to look.

Examples
---------
``` python
class COverload:
    x: str
    @overload
    def meth(self, a: int) -> float: ...
    @overload
    def meth(self, a: str) -> Sequence[float]: ...
    def meth(self, a):
        pass

cls_overload: Type[Union[COverload, E]]
if issubclass(cls_overload, POverload):
    reveal_type(cls_overload)  # N: Revealed type is 'Type[__main__.COverload]'
else:
    reveal_type(cls_overload)  # N: Revealed type is 'Type[__main__.E]'
```

The test output:
```
Expected:
  ...
  main:45: note: Revealed type is 'Type[__main__.COverload]'
  main:47: note: Revealed type is 'Type[__main__.E]' (diff)
Actual:
  ...
  main:45: note: Revealed type is 'Type[__main__.COverload]'
  main:47: note: Revealed type is 'Union[Type[__main__.COverload], Type[__main__.E]]' (diff)
```

97littleleaf11:
Passes all the tests now. Shall we merge this one? @hauntsaninja

JelleZijlstra:
Jukka already approved it, so let's land it.

test-data/unit/check-protocols.test
97littleleaf11: Could you please open a new issue about it?

BvB93: > Could you please open a new issue about it?

I just created an issue at https://github.com/python/mypy/issues/12228.

PR #9905, opened 2021-01-12
Fix TypedDict.get("missing_key") with string literal

### LINKED ISSUE #9902: TypedDict does not allow .get("missing_key", ...) to pass
**Bug Report**

MyPy does not allow `.get("key")` on a TypedDict if the key does not exist. This nullifies the usefulness of `.get`, and seems to be in direct contradiction to PEP 589:

"Operations with arbitrary str keys (instead of string literals or other expressions with known string values) should generally be rejected. This involves both destructive operations such as setting an item and read-only operations such as subscription expressions. As an exception to the above rule, d.get(e) and e in d should be allowed for TypedDict objects, for an arbitrary expression e with type str. The motivation is that these are safe and can be useful for introspecting TypedDict objects."

(Credit to @freundTech on python/typing gitter)

An example of where this is broken in a practical example is shown below.

**To Reproduce**

```python
from typing import TypedDict

class BaseDict(TypedDict):
    a: str

class CustomDict(BaseDict):
    b: str

def my_general_fun(x: BaseDict) -> None:
    print(x["a"], x.get("b"))
```

**Expected Behavior**

This should pass as far as I can tell from the PEP; a literal expression _is_ an arbitrary expression e with type str, just a special case of it. For `[]` non-literals are thrown out, but for `.get`, non-literals are included; I don't think this implies literals should be thrown out, but rather that the point of `.get` is for generic access. The type of `.get` should be inferred if a literal is passed and that literal is known, but it should it should the value of the second argument (`.get("b", "")` could be `str` if the TypedDict did not have a `"b"`, for example, normally None). This would be much more useful than simply not allowing this to pass at all.

From looking at this function:
 https://github.com/python/mypy/blob/6e99a2db100d794b1bf900746470527cd03fce16/mypy/plugins/default.py#L227

The check only happens for literals, so a workaround is to use the following instead:

```python
def my_general_fun(x: BaseDict) -> None:
    b = "b"
    print(x["a"], x.get(b))
```

Btu this should not be required, and could even lose some knowledge about the type.

I would expect the type of `x.get("b")` to be None.

**Actual Behavior**

```
tmp.py:11: error: TypedDict "BaseDict" has no key 'b'
```

**Your Environment**

- Mypy version used: 7.900
- Mypy command-line flags: None
- Python version used: 3.9
- Operating system and version: macOS 11

freundTech:
### Description

Fixes #9902

Previously this would throw an error, however according to PEP 589
.get() should be allowed for arbitrary expressions with type str, while
only [] throws an error.

## Test Plan

All affected tests have been updated to the new, intended behavior. This is my first contribution to mypy, so it would be good to have an extra pair of eyes checking if everything is correct.


freundTech:
Looks like I accidentally checked in some wrong files

PR #9909, opened 2021-01-13
Fix false unreachable error reports.

### LINKED ISSUE #9907: unreachable: false positive for isinstance
I have come across the problem when working with runtime checkable Protocols and overloading. Still, I can simplify it to the following example, where Mypy (version 0.79) reports a wrong "unreachable" error and thus misses the real bug in the last line:

```
from abc import abstractmethod


class A:
    f: Literal[0]


class B:
    @property
    @abstractmethod
    def f(self) -> Literal[0, 1]:
        ...

    def t(self) -> None:
        if isinstance(self, A):  # error: Subclass of "B" and "A" cannot exist: would have incompatible method signatures  [unreachable]
            self.f = 1  # error: Statement is unreachable  [unreachable]
```

On the other hand, when checking multiple inheritance, Mypy correctly detects that the definition of class `C` is  type-safe but the definition of class `D` is not:

```
class C(A, B):
    ...


class D(B, A):  # error: Definition of "f" in base class "B" is incompatible with definition in base class "A"  [misc]
    ...
```

So method `TypeChecker.check_multiple_inheritance` could eventually help to fix method `TypeChecker.find_isinstance_check_helper`?

As mentioned above, the problem would be the same if class `A` were a runtime checkable protocol.

tyralla:
Fixes #9907

For explanations, please see the commit message.

I hope changing the signature of method `TypeChecker.intersect_instances` is acceptable.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3725,7 +3725,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         return cdef, info
 
     def intersect_instances(self,
-                            instances: Sequence[Instance],
+                            instances: Tuple[Instance, Instance],
                             ctx: Context,
                             ) -> Optional[Instance]:
         """Try creating an ad-hoc intersection of the given instances.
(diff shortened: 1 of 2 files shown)


mypy/checker.py
```diff
@@ -3725,7 +3725,7 @@ def make_fake_typeinfo(self,
         return cdef, info
 
     def intersect_instances(self,
-                            instances: Sequence[Instance],
+                            instances: Tuple[Instance, Instance],
```
JelleZijlstra, line 3728: Why do you need this change? I don't see anything below that requires there to be exactly two.

tyralla: My change covers two permutations of the given `Instance` objects: it checks the original and the reversed order.  For the given order `(a, b, c)`, we would also need to check the order `(b, a, c)`, for example.

tyralla: Just a reminder. Or can I help with further explanations?

JelleZijlstra (approved):
Looks good but I'd also want @JukkaL to review this.

PR #9921, opened 2021-01-17
stubtest: fix signature construction for overload + implicit classmethod

hauntsaninja:
`__class_getitem__` and `__init_subclass__` are always classmethods, even if
they aren't decorated. We previously only handled this correctly if
these methods weren't overloaded.

This came up in https://github.com/python/typeshed/pull/4937

### DIFF
--- a/mypy/stubtest.py
+++ b/mypy/stubtest.py
@@ -377,6 +377,15 @@ def _verify_arg_default_value(
             )
 
 
+def maybe_strip_cls(name: str, args: List[nodes.Argument]) -> List[nodes.Argument]:
+    if name in ("__init_subclass__", "__class_getitem__"):
+        # These are implicitly classmethods. If the stub chooses not to have @classmethod, we
+        # should remove the cls argument
+        if args[0].variable.name == "cls":
+            return args[1:]
+    return args
+
+
 class Signature(Generic[T]):
     def __init__(self) -> None:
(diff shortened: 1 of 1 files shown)


mypy/stubtest.py
```diff
@@ -377,6 +377,15 @@ def _verify_arg_default_value(
             )
 
 
+def maybe_strip_cls(name: str, args: List[nodes.Argument]) -> List[nodes.Argument]:
+    if name in ("__init_subclass__", "__class_getitem__"):
+        # These are implicitly classmethods. If the stub chooses not to have @classmethod, we
+        # should remove the cls argument
+        if args[0].variable.name == "cls":
```
JelleZijlstra, line 384: What if I name it `_cls` instead? It sounds like we should just unconditionally remove the first arg.

hauntsaninja: Unconditionally removing the first argument here isn't right, because it'll do the wrong thing if you do choose to decorate with `@classmethod`, since those get their first argument stripped when resolving the decorator.

I don't think this is important enough to figure out how to keep that state.

Especially since stubtest currently is picky about classmethod names being cls, e.g. this not-actually-needed-potentially-problematic assert when resolving the classmethod decorator: https://github.com/python/mypy/blob/92923b2ed8085a38c353382f805c9e2e8716e717/mypy/stubtest.py#L814

PR #9923, opened 2021-01-18
Document new source finding behaviour

hauntsaninja:
This should cover the current state on master, as previously
discussed / implemented across #9742, #9683, #9632, #9616, #9614, etc.

This will need to be changed if we can make `--namespace-packages` the
default (#9636).

I haven't documented some of the finer points of the changes, since it
felt like an inappropriate level of detail (e.g. using absolute paths
when crawling, how directories with invalid package names affect
crawling, etc)

### DIFF
--- a/docs/source/running_mypy.rst
+++ b/docs/source/running_mypy.rst
@@ -352,15 +352,15 @@ is by providing mypy a list of paths. For example::
 This section describes how exactly mypy maps the provided paths
 to modules to type check.
 
-- mypy will check all paths provided that correspond to files.
+- Mypy will check all paths provided that correspond to files.
 
-- mypy will recursively discover and check all files ending in ``.py`` or
+- Mypy will recursively discover and check all files ending in ``.py`` or
   ``.pyi`` in directory paths provided.
 
 - For each file to be checked, mypy will attempt to associate the file (e.g.
   ``project/foo/bar/baz.py``) with a fully qualified module name (e.g.
   ``foo.bar.baz``). The directory the package is in (``project``) is then
-  add to mypy's module search paths.
+  added to mypy's module search paths.
 
 How mypy determines fully qualified module names depends on if the options
 :option:`--namespace-packages <mypy --namespace-packages>` and
--- a/docs/source/running_mypy.rst
+++ b/docs/source/running_mypy.rst
@@ -24,8 +24,11 @@ actual way mypy type checks your code, see our
 Specifying code to be checked
 *****************************
 
-Mypy lets you specify what files it should type check in several
-different ways.
+Mypy lets you specify what files it should type check in several different ways.
+
+Note that if you use namespace packages (in particular, packages without
+``__init__.py``), you'll need to specify :option:`--namespace-packages <mypy
+--namespace-packages>`.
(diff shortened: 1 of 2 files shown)


docs/source/running_mypy.rst
```diff
@@ -336,58 +339,79 @@ while this option can be quite powerful, it can also cause many
...
+is by providing mypy a list of paths. For example::
 
     $ mypy file_1.py foo/file_2.py file_3.pyi some/directory
 
 This section describes how exactly mypy maps the provided paths
 to modules to type check.
 
-- Files ending in ``.py`` (and stub files ending in ``.pyi``) are
-  checked as Python modules.
+- mypy will check all paths provided that correspond to files.
```
JukkaL, line 355: Style nit: At the start of a sentence, capitalize 'mypy' (mypy -> Mypy).

docs/source/running_mypy.rst
```diff
@@ -336,58 +339,79 @@ while this option can be quite powerful, it can also cause many
...
+
+- mypy will recursively discover and check all files ending in ``.py`` or
```
JukkaL, line 357: mypy -> Mypy

docs/source/running_mypy.rst
```diff
@@ -336,58 +339,79 @@ while this option can be quite powerful, it can also cause many
...
+  ``.pyi`` in directory paths provided.
+
+- For each file to be checked, mypy will attempt to associate the file (e.g.
+  ``project/foo/bar/baz.py``) with a fully qualified module name (e.g.
+  ``foo.bar.baz``). The directory the package is in (``project``) is then
+  add to mypy's module search paths.
```
JukkaL, line 363: Typo: add -> added

docs/source/running_mypy.rst
```diff
@@ -336,58 +339,79 @@ while this option can be quite powerful, it can also cause many
...
+   provides a solution.
 
-For example, suppose we run the command ``mypy foo/bar/baz.py`` where
-``foo/bar/__init__.py`` exists but ``foo/__init__.py`` does not.  Then
-the module name assumed is ``bar.baz`` and the directory ``foo`` is
-added to mypy's module search path.
+   With :option:`--explicit-package-bases <mypy --explicit-package-bases>`, mypy
+   will locate the nearest parent directory that is a member of the ``MYPYPATH``
+   environment variable, the :confval:`mypy_path` config or is the current
+   working directory. mypy will then use the relative path to determine the
```
JukkaL, line 400: Should you also mention module search path items from PEP 561 packages?

JukkaL (approved):
Thanks for documenting this! Specifying the files to type check can be a little tricky, especially when namespace packages are used, so documenting it in detail is very useful.

Looks good, just a few minor comments.

docs/source/running_mypy.rst
hauntsaninja: No, there isn't an interaction with PEP 561 packages when you pass files: https://github.com/python/mypy/blob/9c54cc3f1c73150992345350be834f2987b3967d/mypy/find_sources.py#L73

PR #9924, opened 2021-01-18
moduleinfo: delete

hauntsaninja:
I removed the documentation for this erorr already in #9675

The list of modules in moduleinfo that we don't have stubs for is now down to:
```
sqlite3.dump
turtledemo
xml.sax._exceptions
xml.sax.expatreader
```


JukkaL (approved):
It's good to not have to worry about missing stdlib stubs any more.

Issue #9915, opened 2021-01-16
Add a plugin hook for hadling function/method decorators


freundTech:
**Feature**

mypy currently has the `get_class_decorator_hook` plugin hook. I want to propose adding a `get_function_decorator_hook` and `get_method_decorator_hook`.

**Pitch**

Adding these hooks would allow special handling of decorated functions, as would for example be required to solve #9911 using a plugin. mypy currently has special handling for a few decorated such as `builtins.property`, `typing.final` and more, but from what I can see it's currently not possible to add additional cases using a plugin.

I have already experimented with this a bit and could probably open a PR in the near future.

Issue #9911, opened 2021-01-14
pyqtProperty setter always raise an error


BoboTiG:
**Bug Report**

It seems that mypy is not enjoying `pyqtProperty`setters. Someone already asked on the [Qt mailing-list](http://python.6.x6.nabble.com/Reconciling-pyqtProperty-and-mypy-td5279775.html).

**To Reproduce**

```python
from PyQt5.QtCore import QObject, pyqtProperty, pyqtSignal


class FeatureModel(QObject):

    stateChanged = pyqtSignal()

    def __init__(self, enabled: bool, /) -> None:
        super().__init__()
        self._enabled = enabled

    @pyqtProperty(bool, notify=stateChanged)
    def enabled(self):
        return self._enabled

    @enabled.setter
    def enabled(self, enabled):
        self._enabled = enabled
        self.stateChanged.emit()
```

**Expected Behavior**

Mypy should be happy.

**Actual Behavior**

```shell
$ python -m mypy file.py
file.py:16: error: Name 'enabled' already defined on line 12
Found 1 error in 1 file (checked 1 source file)
```

**Your Environment**

- Mypy version used: 0.790
- Mypy command-line flags: None
- Mypy configuration options from `mypy.ini` (and other config files): None
- Python version used: 3.8.6
- Operating system and version: macOS Big Sur (arm64) and Windows 10.

freundTech:
On the Qt mailing-list someone already posted these two links to the mypy source-code:
https://github.com/python/mypy/blob/v0.782/mypy/semanal.py#L964-L973
https://github.com/python/mypy/blob/v0.782/mypy/semanal.py#L842-L869

mypy has some special casing to make pythons @property (and a few other decorators) work.
The easy solution would probably be to add a special case for @pyqtProperty.
Cleaner solutions would be to either somehow make it possible to annotate that a decorator creates a property (This could be problematic, as I don't know if existing type annotations can handle this) or write a mypy plugin. I only took a quick look at the plugin interface and think it should be possible to do this. The problem here would be that the plugin API is unstable (See #6617).

gvanrossum:
The way to do that without baking knowledge of qt into mypy is to write a mypy extension.

freundTech:
Should the plugin hook I suggested in #9915 get added to mypy you could solve this using the following mypy plugin (function and class names subject to change):

```python
from typing import Callable, Optional
from typing import Type

from mypy.plugin import (
    Plugin, DecoratorContext,
)


[...]

PR #9925, opened 2021-01-18
[WIP] Add get_function_decorator_plugin_hook to plugin interface

freundTech:
### Description

This PR adds the plugin hook I proposes in #9915.

### Test Strategy

Two new unit tests have been added to test the new plugin hook

#### Why is this plugin hook needed?:
This hook can be used to solve problems like for example #9911, which can't be solved by existing plugin hooks.
There is already a plugin hook for handling class decorators, but none for method and function decorators.

#### Why is there no `get_method_decorator_plugin_hook`?:
From what I can tell the semantic analysis phase doesn't distinguish between functions and methods. Therefor only one hook has been added. The other hooks, which have `function` and `method` variants tun during the checking phase.

#### What can this hook do?
Mark decorated functions as properties, abstract, final and coroutines

#### What can't this hook do?
This hook currently can't disable type checking for decorated functions, despite the builtin '@typing.no_type_check' decorator being able to do so. Enabling this would require more code changes, as a local variable is used for this ad would have to be exposed in some way.

#### When does this hook run?
This hook runs after the builtin code has handled the decorators, but before handled decorators are removes and the decorated function is semantically analyzed.

## Additional considerations

* Make it possible for plugins to disable type checking for a function
* Expose `SematicAnalyzer.check_decorated_function_is_method` on the `SemanticAnalyzerPluginInterface`

### DIFF
--- a/mypy/plugin.py
+++ b/mypy/plugin.py
@@ -124,7 +124,7 @@ from typing import Any, Callable, List, Tuple, Optional, NamedTuple, TypeVar, Di
 from mypy_extensions import trait, mypyc_attr
 
 from mypy.nodes import (
-    Expression, Context, ClassDef, SymbolTableNode, MypyFile, CallExpr
+    Expression, Context, ClassDef, SymbolTableNode, MypyFile, CallExpr, Decorator
 )
 from mypy.tvar_scope import TypeVarLikeScope
 from mypy.types import Type, Instance, CallableType, TypeList, UnboundType, ProperType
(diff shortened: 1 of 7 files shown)


patrick91:
@freundTech I'd love to see this merged, is there anything I can do to push this forward?

freundTech:
I think the code is mostly finished and just needs some cleanup. I posted this as WIP, because I wanted some opinions on the "additional considerations" section first. Then a maintainer has to approve this.

I'll link this in the typing gitter to see if I can get some people to have a look at it.

patrick91:
@freundTech maybe you can send a message on the mailing list as well 😊 https://mail.python.org/archives/list/typing-sig@python.org/

97littleleaf11:
Any progress on this PR?

freundTech:
I think my last comment still stands. I haven't received any feedback on if this feature is needed/wanted/the best way to solve #9911 yet.

If there is consensus that this should be added I can rebase it and clean up the code.

embray:
This would indeed be useful.  It can be used also to create no-op decorators that mark a function or method as "this needs special typing support from my plugin", where the plugin can then determine based on additional context how to rewrite that function's signature.

I'm trying to figure out if there's a workaround to that without an explicit "get_function_decorator_hook" but it does make it a bit trickier...

Update: It does seem to be possible with just a `get_function_hook` matching on the special decorator name, and modifying the return type of the decorator.   But that's just one special use case I guess.

PR #9949, opened 2021-01-23
Fix Python 3.5 travis build by pinning pip

JukkaL:
It looks like pip just dropped Python 3.5 support, but we are not quite
yet ready to do it (as long as there's an easy workaround).

Example failure here:
https://travis-ci.com/github/python/mypy/jobs/474676438

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,6 +82,7 @@ jobs:
   #   - EXTRA_ARGS=
 
 install:
+# pip 21.0 no longer works on Python 3.5
 - pip install -U pip==20.3.4 setuptools
 - pip install -U 'virtualenv<20'
 - pip install -U tox==3.20.1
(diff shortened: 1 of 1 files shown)


hauntsaninja:
When would you accept a PR that drops Python 3.5 support?

.travis.yml
```diff
@@ -82,7 +82,7 @@ jobs:
   #   - EXTRA_ARGS=
 
 install:
-- pip install -U pip setuptools
+- pip install -U pip==20.3.4 setuptools
```
hauntsaninja, line 85: nit: could you add a comment that mentions Python 3.5 in some way so this is discoverable via grep

hauntsaninja (approved):

JukkaL:
> When would you accept a PR that drops Python 3.5 support?

Around the time that Ubuntu 16.04 LTS goes out of support (this April), assuming that it's not too much of a pain to support Python 3.5 until then.

.travis.yml
JukkaL: A good idea, done.

Issue #4814, opened 2018-03-28
Clean-up type variable scope and binding


ilevkivskyi:
Currently the code around type variable binding in functions is a bit unclear and does some strange things like binding typer variables before function type is analyzed and depending on `TypeVarQuery` returning short (unqualified) names.

ilevkivskyi:
Another related comment here is that we have a bit too many "things" related to type variables (in addition to `TypeVarScope` mentioned above):
* `TypeVarExpr`
* `TypeVarDef`
* `TypeVarType`
* `TypeVarId`
We can probably simplify a bit the model of how type variables work.

ilevkivskyi:
Probably it makes sense to clean up (or at least document) various things related to type variable scopes while we are working on the new semantic analyzer. I just discovered an inconsistency in new analyzer, it was using full type variable name where short one was expected. Without documentation, such errors are easy to make.

ilevkivskyi:
FWIW I more happy now with how type variables work (e.g. we use namespaces everywhere). There still may be bugs, I think `tvar_scope.py` is somewhat over-complicated, but we can handle issues on one-by-one basis now.

PR #9951, opened 2021-01-24
Merge TypeVarDef and TypeVarType

hauntsaninja:
In the spirit of https://github.com/python/mypy/issues/4814#issuecomment-414150557

I'm not sure what the distinction between TypeVarDef and TypeVarType was (note that there's also TypeVarExpr, whose purpose is clear to me).
In practice, it seems it boiled down to: a) TypeVarDef has a different `__repr__`, that's not user facing but was used in tests, b) in two places, we used the construction of a TypeVarType from a TypeVarDef as an opportunity to change the line number.

I ran into this because:
a) I was considering whether to create a corresponding ParamSpecDef and ParamSpecType
b) I was trying to fix an issue with applytype sometimes swallowing generics, which to fix I needed TypeVarType to have access to its TypeVarDef.
I was unable to find a reason for TypeVarType to not just *be* the TypeVarDef and here we are.


gvanrossum:
Do we even need the separation between TypeVarDef and ParamSpecDef? There seems a lot of duplication. A single flag bit or maybe making ParamSpecDef a trivial subclass of the other might simplify things there.

hauntsaninja:
Yeah, it's conceivable it could be made a flag. But having it be its own type makes my development much easier, since it allows mypy to tell me where I need to add functionality. I already extracted what's common into `TypeVarLikeDef` / `ParamSpecDef` is basically a trivial subclass of `TypeVarLikeDef`.

ethanhs:
We should probably give people advanced notice about this change in #6617 since I expect several plugins deal with TypeVars.

mypy/checkexpr.py
```diff
@@ -3277,13 +3277,12 @@ def check_lst_expr(self, items: List[Expression], fullname: str,
         # Used for list and set expressions, as well as for tuples
         # containing star expressions that don't refer to a
         # Tuple. (Note: "lst" stands for list-set-tuple. :-)
-        tvdef = TypeVarDef('T', 'T', -1, [], self.object_type())
-        tv = TypeVarType(tvdef)
+        tvdef = TypeVarType('T', 'T', -1, [], self.object_type())
```
msullivan, line 3280: Maybe keep the `tv` names (and similar below) instead of the `tvdef` ones?

msullivan (approved):
I made a suggestion but ignore it if you don't agree. 

And yeah, we should mention it in the plugin issue.

msullivan:
And this is great. Honestly I never understood the distinction distinction and am happy to learn that this confusion was warranted.

Issue #9954, opened 2021-01-24
Running mypy against a directory produces different outcome from running against all files in that directory


obi1kenobi:
**Bug Report**
Unexpectedly to me, running `mypy` against an entire directory does not produce the same results as running it explicitly against all files in that directory.

Specifically, it is possible to construct a scenario where the following two invocations differ in outcomes — one will claim that there are no type errors, while the other will find valid type errors, even though both runs scanned the same files and used the same `mypy` configuration:
```
mypy src/
```
vs
```
mypy src/foo.py src/bar.py < ... all other Python files inside src/ ... >
```

**To Reproduce**

I encountered this bug while working with `pip` and its use of `mypy`, this issue has all the details including detailed repro steps: https://github.com/pypa/pip/issues/9502

The gist of the repro is the following:
* Clone the `pip` [repo](https://github.com/pypa/pip). The `master` branch as of this post was at commit `8d042c448718355447f133c440f7a931a5ae57c0`, and that is where I found the issue — though I don't expect the exact commit matters.
* Install `mypy==0.790`.
* Run `mypy src/` and note the 51 type errors. This is confusing, because `pip` runs `mypy` in its CI environment, via a `pre-commit` hook.
* Observe that `pre-commit` executes `mypy` with a different invocation, explicitly listing all the files in `src/` instead of simply supplying the directory (the full invocation is quite long, you can find it in the linked issue above). This invocation finds no type errors.

**Expected Behavior**
When scanning the same set of files with the same configuration, I expected `mypy` to behave identically. I expected that specifying a directory and specifying all Python files within that directory should find identical sets of type errors.

It seems that the maintainers of the `pre-commit` project and perhaps the `pip` maintainers as well may have had a similar expectation, as the 51 type errors reported in the `pip` project when running `mypy` with `pip`'s configuration appear to be real type errors and not erroneous in any way.

**Actual Behavior**
`mypy` finds 51 type errors when scanning the `pip` project as a directory, but finds no type errors when invoked with an explicit list of all the files within the `src/` directory.

**Your Environment**

- Mypy version used: 0.790
- Mypy command-line flags: None (`--pretty` is used by `pip`'s configuration, but the same issue exists without it as well)
- Mypy configuration options from `mypy.ini` (and other config files): The following config is in `setup.cfg`:
```
[mypy]
follow_imports = silent
ignore_missing_imports = True
disallow_untyped_defs = True
disallow_any_generics = True
warn_unused_ignores = True

[...]

PR #9955, opened 2021-01-24
Emphasise that users probably want --follow-imports=normal

hauntsaninja:
For whatever reason, these options get misused a lot and cause a lot of
problems. #9954 is only the latest in a long history of
misunderstanding.

I speculate that this is because of a) the bad legacy of flake8-mypy,
b) people confusing it with ignore_missing_imports


gvanrossum (approved):
LGTM. As you probably know that flag was primarily designed for a transitional stage of introducing mypy at Dropbox gradually.

Issue #2608, opened 2016-12-25
Common python idiom hard to use with inner methods and strict-optional


euresti:
So everybody knows not to write
`def foo(arg={'bar': 'baz'}):`

So instead one usually writes something like:
```
def foo(arg=None):
    if arg is None:
        arg = {'bar', 'baz'}
```

Now add some types:
```
def func(arg=None):
    # type: (Dict[str, str]) -> Dict[str,str]
    reveal_type(arg)  # Revealed type is 'Union[builtins.dict[builtins.str, builtins.str], builtins.None]'
    if arg is None:
        arg = {'foo': 'bar'}
    reveal_type(arg)  #  Revealed type is 'builtins.dict[builtins.str, builtins.str]'
    return arg
```

So all is good, unless you try to create an inner method:
```
def func(arg=None):
    # type: (Dict[str, str]) -> Dict[str,str]
    reveal_type(arg)  # Revealed type is 'Union[builtins.dict[builtins.str, builtins.str], builtins.None]'
    if arg is None:
        arg = {'foo': 'bar'}
    reveal_type(arg)  #  Revealed type is 'builtins.dict[builtins.str, builtins.str]'

    def inner():
        # type: () -> Dict[str, str]
        reveal_type(arg)  # Revealed type is 'Union[builtins.dict[builtins.str, builtins.str], builtins.None]'
        return arg  # Incompatible return value type (got "Optional[Dict[str, str]]", expected Dict[str, str])

    return inner()
```

I guess there are lots of workarounds (assign to temporary variable, pull out the inner method, pass arguments into the inner) but they all require editing existing code which is always scary.  Hmm. I guess the easiest thing is to just add an `assert arg is not None` in the inner.  That's not too intrusive, anyway thought you might want to know.

gvanrossum:
This looks like the same as #2145 which was closed because the binder doesn't keep track of this -- maybe we should reopen that? Or at least keep this one open.

JukkaL:
Let's keep this open.

ilevkivskyi:
(Rising priority to normal, since this appeared third time.)

gvanrossum:
How should we do this? I think it's safe to trust the narrower type *if it
prevails over the entire rest of the (outer) function* -- maybe that's
enough? After all the "None" is an artifact of the idiom for mutable
default. (I almost wish there was a way to say "this None isn't really
here" like we have at the class level.)

carljm:
[...]

PR #9956, opened 2021-01-24
Add narrowing and closures to common issues

hauntsaninja:
This one is on our issue tracker's greatest hits.
Here's the issue I usually link: #2608

### DIFF
--- a/docs/source/common_issues.rst
+++ b/docs/source/common_issues.rst
@@ -772,3 +772,30 @@ False:
 
 If you use the :option:`--warn-unreachable <mypy --warn-unreachable>` flag, mypy will generate
 an error about each unreachable code block.
+
+
+Narrowing and inner functions
+-----------------------------
+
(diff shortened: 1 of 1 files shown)


sirosen:
Huge 👍 to adding this.

I think it would be helpful to have an example rather than just saying "assign `y = x` and use `y`".

You may need to annotate the type of the reassigned variable, depending on your case. Here's a minimal example of what I mean:
```
def outer(x: typing.Union[None, int, float]):
    if isinstance(x, float):
        x_: typing.Optional[int] = int(x)
    else:
        x_ = x

    def inner() -> int:
        if x_:
            return x_ + 1
        else:
            return 0

    return inner
```

Without `x_` annotated, mypy will deduce its type -- correctly -- from the first assignment as `int`. But we're really trying to narrow the union down to `Optional[int]` in this case.

So maybe it would be best to advise people that you "may need to annotate the reassignment in some cases" or something?

PR #9957, opened 2021-01-24
[mypyc] Remove unused fast argument parsing code

JukkaL:
Since we only use relatively few features supported by the original argument
parsing code, much of the code can be deleted.

This speeds up the `keyword_args_from_interpreted` benchmark by about 8%.


msullivan:
Fantastic

PR #9959, opened 2021-01-24
Clarify that mypy_path is relative to CWD

gustavgransbo:
### Description
I was unsure if the `mypy_path` entry in the config file was treated relative to the config file, and found the answer in #7967 (it is treated relative to the working directory of the mypy command).

This PR documents current behavior, and mentions how `MYPY_CONFIG_FILE_DIR` can be used to refer to path relative to the configuration file.

## Test Plan
Rebuilding docs creates expected results.


hauntsaninja:
Thanks!

PR #9964, opened 2021-01-25
Fix inconsistency in the disallow-untyped-defs doc

### LINKED ISSUE #9963: disallow-untyped-defs vs disallow_incomplete_defs doc inconsistency
https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_untyped_defs says:

> Disallows defining functions without type annotations or with incomplete type annotations.

Notice the extra ` or with incomplete type annotations`.

and https://mypy.readthedocs.io/en/stable/config_file.html#confval-disallow_incomplete_defs says:

> Disallows defining functions with incomplete type annotations.

While https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-untyped-defs says:

> This flag reports an error whenever it encounters a function definition without type annotations.

and https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-incomplete-defs says:

> This flag reports an error whenever it encounters a partly annotated function definition.

I believe the version from `command_line` doc is the correct one, and `config_file` doc should be updated.

ravwojdyla:
### Description

Fixes #9963

Fix inconsistency in the disallow-untyped-defs doc.

### DIFF
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -389,8 +389,7 @@ section of the command line docs.
     :type: boolean
     :default: False
 
-    Disallows defining functions without type annotations or with incomplete type
-    annotations.
+    Disallows defining functions without type annotations.
 
 .. confval:: disallow_incomplete_defs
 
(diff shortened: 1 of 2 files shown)


hauntsaninja:
Thanks, but I don't think this change is correct

PR #9966, opened 2021-01-25
Fix typo in error message

gustavgransbo:
### Description
Fixes a typo in the error raised when envoking mypy with `--explicit-package-bases` without also providing `--namespace-packages`. Currently the error refers to a non-existing parameter `--explicit-base-dirs`.


JelleZijlstra:
Thanks!

PR #9969, opened 2021-01-25
Make --explicit-package-bases invertible

### LINKED ISSUE #9968: Make `--explicit-package-bases` revertible
**Feature**

`--explicit-package-bases` should have an inverse: `--no-explicit-package-bases`.

**Pitch**

I have a `mypy.ini` that declares `explicit_package_bases = True`. It would be convenient to be able to override this setting from the command line by passing  `--no-explicit-package-bases`.

My use case is that I have a mono repo with a shared `mypy.ini`, where I mostly want to use `--explicit-package-bases`.
However, in some sub directories I have the following structure:
```
sub-dir
│
└───src
│   └───pkg
│   │   └─── # Package content
│   └───tests
│
└───noxfile.py
```
The `noxfile.py` is used to trigger a `mypy src/ tests/ --config {PATH_TO_CONFIG}` session with `sub-dir` as working directory.
I would like to use the same `mypy.ini` I use in the rest of the repo to not have to duplicate settings, but this sets `--explicit-package-bases`. Unfortunately this leads to `mypy` thinking that `src` is the namespace of `pkg`, since it is located at the working directory. The most convenient solution for me would be to pass `--no-explict-package-bases` when invoking `mypy` in `sub-dir`.

gustavgransbo:
### Description
Closes #9968

This PR adds the inverse of `--explicit-package-bases`: `--no-explicit-package-bases`.
The flag can be used if you want to override `explicit_package_bases = True` which was set in a configuration file.

## Test Plan

I verified that passing `--no-explicit-package-bases` with `explicit_package_bases = True` configured overrides the config.


JukkaL (approved):
Thanks for the PR! Looks good to me. And sorry for the slow response.

PR #9973, opened 2021-01-26
Use modular typeshed (unbundle third-party package stubs)

### LINKED ISSUE #9971: Switch to modular typeshed
The structure of typeshed is changing, and tools are no longer expected to bundle the entire typeshed. Instead, users will install relevant PEP 561 stub distributions for the packages that they need. 

Typeshed will continue to host third-party package stubs, and updates to third-party stubs will automatically uploaded to PyPI.

This will affect mypy in several ways:
* Mypy will only ship with stdlib stubs.
* Stubs for most third-party packages must be installed using pip, e.g. `pip install types-requests` for stubs for `requests`.
* Typeshed will no longer be a git submodule. Instead, we'll copy the stdlib stubs to the mypy repository.

This has some major benefits:
* It's easy to start using new stubs, or a new version of existing stubs, contributed to typeshed. Users don't need to wait for the next mypy release -- they can just install or update the relevant stubs.
* Users no longer have to update to the latest stubs when updating mypy. This can save a lot of friction, since new stubs may have errors, and updating to stubs with stricter types (even if they are precise) may require significant changes to code that uses the relevant features.
* This makes things more consistent. Currently some third-party stubs are bundled with mypy, while some must be installed from PyPI. In the future all third-party stubs are installed in a similar way using pip.
* It will be easier to maintain multiple versions of stubs (for example, separate stubs for versions of 1.x and 2.x of a library).

We can make the transition relatively smooth by suggesting how to install the relevant stubs when the stubs are missing. We can also support automatically installing the necessary stubs (`mypy --install-stubs`).

Discussion about the typeshed change, including motivation: https://github.com/python/typeshed/issues/2491

JukkaL:
Support the new structure of typeshed (python/typeshed#2491) and only
bundle stdlib stubs with mypy. 

Most stubs for third-party packages now need to be installed using pip (for 
example, `pip install types-requests`). Add stubs for `typing_extensions` 
and `mypy_extensions` as mypy dependencies, since these are 
needed for basic operation.

Suggest a pip command to run if we encounter known missing stubs.

Add `--install-types` option that installs all missing stub packages. This
can be used as `mypy --install-types` to install missing stub packages from
the previous mypy run (no need to provide a full mypy command line).

This also replaces the typeshed git submodule with a partial copy of the
typeshed that only includes stdlib stubs. 

Add a script to sync stubs from typeshed (`misc/sync-typeshed.py`). 
This is still incomplete since typeshed hasn't actually migrated to the new 
structure yet.

Work towards #9971.


docs/source/getting_started.rst
```diff
@@ -344,11 +345,40 @@ Without a library stub, mypy would have no way of inferring the type of ``x``
...
+
+.. code-block:: text
+
+  prog.py:1: error: Library stubs not installed for "yaml" (or incompatible with Python 3.8)
+  prog.py:1: note: Hint: "python3 -m pip install types-PyYAML"
+  ...
+
+.. note::
+
+   Starting in mypy 0.800, most third-party package stubs must be
```
ilevkivskyi, line 372: I assume you mean 0.900 here too.

docs/source/command_line.rst
```diff
@@ -841,6 +843,26 @@ format into the specified directory.
...
+    If you use this option without providing any files or modules to
+    type check, mypy will install stub packages suggested during the
+    previous mypy run. If there are files or modules to type check,
+    mypy first type checks those, and proposes to install missing
+    stubs at the end of the run, but only if any missing modules were
+    detected.
+
+    .. note::
+
+        This is new in mypy 0.800. Previous mypy versions included a
```
ilevkivskyi, line 862: I assume you mean 0.900 here.

ilevkivskyi (approved):
LGTM. thanks!

JukkaL:
I rebased and force pushed in the hope of fixing the brokenness.

JukkaL:
I'm going to ignore the mypy primer failures for now. If they persist, I'll disable mypy primer temporarily.

ethanhs:
>This also replaces the typeshed git submodule with a partial copy of the
typeshed that only includes stdlib stubs.

Is the idea here that we don't want to include all of typeshed for releases? If that is the case we can keep the submodule and only include the stdlib folder in typeshed I believe (which would be a bit easier I think).

E.g. If we change [this line](https://github.com/python/mypy/blob/ba037888eccd0f63e193d0fefd8f4e9265b8e3d3/setup.py#L75) to point to the stdlib folder of typeshed only that will be included on install.

JukkaL:
> Is the idea here that we don't want to include all of typeshed for releases?

That's part of it, but it's not the only thing. Now that we only need the stdlib stubs, the size is small enough that we can reasonably just commit them all in the repo. This lets us drop the submodule, which simplifies workflows a little -- especially for new contributors who might have no previous experience with git submodules.

Also, if modular typeshed is successful, we'll eventually have 10x the current number of third-party stubs, and having so many stubs (that we don't need) inside the mypy repo, even as a submodule, could be awkward.

PR #9983, opened 2021-01-27
Implement cloning typeshed in typeshed sync script

JukkaL:
Now you can sync the current typeshed master by just running
`misc/sync-typeshed.py` without any arguments.

### DIFF
--- a/misc/sync-typeshed.py
+++ b/misc/sync-typeshed.py
@@ -15,6 +15,7 @@ import subprocess
 import sys
 import tempfile
 import textwrap
+from typing import Optional
 
 
 def check_state() -> None:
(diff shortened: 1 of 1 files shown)


misc/sync-typeshed.py
```diff
@@ -53,12 +67,17 @@ def main() -> None:
...
+    if not args.typeshed_dir:
+        # Clone typeshed repo if no directory given.
+        with tempfile.TemporaryDirectory() as tempdir:
+            print('Cloning typeshed in {}...'.format(tempdir))
+            subprocess.run(['git', 'clone', 'https://github.com/python/typeshed.git'],
+                           check=True, cwd=tempdir)
+            repo = os.path.join(tempdir, 'typeshed')
+            commit = update_typeshed(repo, args.commit)
+    else:
+        commit = update_typeshed(args.typeshed_dir, args.commit)
```
ilevkivskyi, line 79: I think you can simplify this a bit as
```python
if args.typeshed_dir:
    repo = args. typeshed_dir
else:
    ...
    repo = os.path.join(tempdir, 'typeshed')
commit = update_typeshed(repo, args.commit)
```
But this is really minor.

ilevkivskyi (approved):

misc/sync-typeshed.py
JukkaL: Because the context manager deletes the temp directory on exit, the `update_typeshed` call must be within the context manager.

ilevkivskyi: Oooh, I see.

PR #9992, opened 2021-01-29
Add --exclude

### LINKED ISSUE #4675: Add ability to ignore a file or directory without modifying it

### LINKED ISSUE #9981: Please revert new source finding behaviour or add some way to exclude paths
**Bug Report**

I think the PR "find_sources: find build sources recursively" #9614 is a good and worthwhile change. But in practice it causes surprising regressions for many users. It's not just spurious warnings, mypy often refuses to run entirely ("*Found 1 error in 1 file (errors prevented further checking)*").

In many cases, there aren't good work-arounds, short of restructuring one's project directory tree.

I think the magnitude of this issue is large enough that the best course of action is to revert the change and issue a hotfix release, until some work-around for those cases is created.

There are many examples of common things that throw off mypy:
1. Creating even an empty virtualenv causes mypy to fail to run.
2. Having `__init__.py` files under any directory that's not a valid Python module name.
3. Many npm packages include .py files, which get installed in `node_modules` directory, some of which may have Python 2.7 code that causes syntax errors.
4. Having copies of directories or files with same `.py` file names.

Some more examples in comments here: https://github.com/python/mypy/issues/4675#issuecomment-765586163

**To Reproduce**

```
% mkdir test1
% cd test1
% echo 'a: str = 10' > test.py
% mkdir -p test/invalid-module
% touch test/invalid-module/__init__.py
% mypy .
invalid-module is not a valid Python package name
```

```
% mkdir test2
% cd test2
% python3 -m venv ./venv
% echo 'a: str = 10' > test.py
% mypy .
venv/lib/python3.9/site-packages/setuptools/_distutils/command/bdist_msi.py:355: error: expected an indented block
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

```
% mkdir test3
% cd test3
% echo 'a: str = 10' > test.py
% npm install gulp-sass
% mypy .
node_modules/node-sass/node_modules/node-gyp/gyp/pylib/gyp/__init__.py:37: error: invalid syntax
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

```
% mkdir test4
% cd test4
% echo 'a: str = 10' > test.py
% mkdir backup
% cp test.py backup/
% mypy .
test.py: error: Duplicate module named 'test' (also at './backup/test.py')
test.py: error: Are you missing an __init__.py?
Found 2 errors in 1 file (errors prevented further checking)
```

**Your Environment**

- Mypy version used: 0.800
- Mypy command-line flags: none
- Python version used: 3.9
- Operating system and version: macOS 10.15

hauntsaninja:
Resolves #4675, resolves #9981

Additionally, we always ignore site-packages and node_modules.
Also note that this doesn't really affect import discovery; it only
directly affects passing files or packages to mypy.

The additional check before suggesting "are you missing an
__init__.py" didn't make any sense to me, so I removed it, appended to
the message and downgraded the severity to note.

### DIFF
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -49,9 +49,9 @@ for full details, see :ref:`running-mypy`.
     Asks mypy to type check the provided string as a program.
 
 
-.. option:: --ignore-path
+.. option:: --exclude
 
-    Asks mypy to ignore a given file name, directory name or subpath while
+    Asks mypy to exclude a given file name, directory name or subpath while
     recursively discovering files to check. This flag may be repeated multiple
     times.
 
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -56,7 +56,7 @@ for full details, see :ref:`running-mypy`.
     Use forward slashes on all platforms.
 
     For instance, to avoid discovering any files named `setup.py` you could
-    pass ``--exclude '/setup.py$'``. Similarly, you can ignore discovering
+    pass ``--exclude '/setup\.py$'``. Similarly, you can ignore discovering
     directories with a given name by e.g. ``--exclude /build/`` or
     those matching a subpath with ``--exclude /project/vendor/``.
 
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -51,9 +51,9 @@ for full details, see :ref:`running-mypy`.
 
 .. option:: --exclude
 
-    Asks mypy to exclude a given file name, directory name or subpath while
-    recursively discovering files to check. This flag may be repeated multiple
-    times.
+    A regular expression that matches file names, directory names and paths
+    which mypy should ignore while recursively discovering files to check.
+    Use forward slashes on all platforms.
 
 
 Optional arguments
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -55,6 +55,22 @@ for full details, see :ref:`running-mypy`.
     which mypy should ignore while recursively discovering files to check.
     Use forward slashes on all platforms.
 
+    For instance, to avoid discovering any files named `setup.py` you could
+    pass ``--exclude '/setup.py$'``. Similarly, you can ignore discovering
+    directories with a given name by e.g. ``--exclude /node_modules/`` or
+    those matching a subpath with ``--exclude /project/vendor/``.
+
+    Note that this flag only affects recursive discovery, that is, when mypy is
+    discovering files within a directory tree or submodules of a package to
+    check. If you pass a file or module explicitly it will still be checked. For
+    instance, ``mypy --exclude '/setup.py$' but_still_check/setup.py``.
+
+    Note that mypy will never recursively discover files and directories named
+    "site-packages" or "__pycache__" or those whose name starts with a period,
+    exactly as ``--exclude '/(site-packages|__pycache__|\..*)$'`` would.
+    Mypy will also never recursively discover files with extensions other than
+    ``.py`` or ``.pyi``.
+
 
 Optional arguments
 ******************
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -68,7 +68,7 @@ for full details, see :ref:`running-mypy`.
     Note that mypy will never recursively discover files and directories named
     "site-packages", "node_modules" or "__pycache__", or those whose name starts
     with a period, exactly as ``--exclude
-    '/(site-packages|node_modules|__pycache__|\..*)$'`` would. Mypy will also
+    '/(site-packages|node_modules|__pycache__|\..*)/$'`` would. Mypy will also
     never recursively discover files with extensions other than ``.py`` or
     ``.pyi``.
 
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -49,6 +49,13 @@ for full details, see :ref:`running-mypy`.
     Asks mypy to type check the provided string as a program.
 
 
+.. option:: --ignore-path
+
+    Asks mypy to ignore a given file name, directory name or subpath while
+    recursively discovering files to check. This flag may be repeated multiple
+    times.
+
+
 Optional arguments
 ******************
 
--- a/docs/source/command_line.rst
+++ b/docs/source/command_line.rst
@@ -57,7 +57,7 @@ for full details, see :ref:`running-mypy`.
 
     For instance, to avoid discovering any files named `setup.py` you could
     pass ``--exclude '/setup.py$'``. Similarly, you can ignore discovering
-    directories with a given name by e.g. ``--exclude /node_modules/`` or
+    directories with a given name by e.g. ``--exclude /build/`` or
     those matching a subpath with ``--exclude /project/vendor/``.
 
     Note that this flag only affects recursive discovery, that is, when mypy is
@@ -66,10 +66,11 @@ for full details, see :ref:`running-mypy`.
     instance, ``mypy --exclude '/setup.py$' but_still_check/setup.py``.
 
     Note that mypy will never recursively discover files and directories named
-    "site-packages" or "__pycache__" or those whose name starts with a period,
-    exactly as ``--exclude '/(site-packages|__pycache__|\..*)$'`` would.
-    Mypy will also never recursively discover files with extensions other than
-    ``.py`` or ``.pyi``.
+    "site-packages", "node_modules" or "__pycache__", or those whose name starts
+    with a period, exactly as ``--exclude
+    '/(site-packages|node_modules|__pycache__|\..*)$'`` would. Mypy will also
+    never recursively discover files with extensions other than ``.py`` or
+    ``.pyi``.
 
 
 Optional arguments
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -197,6 +197,15 @@ section of the command line docs.
 
     This option may only be set in the global section (``[mypy]``).
 
+.. confval:: ignore_path
+
+    :type: comma-separated list of strings
+
+    A comma-separated list of file names, directory names or subpaths which mypy
+    should ignore while recursively discovering files to check.
+
+    This option may only be set in the global section (``[mypy]``).
+
 .. confval:: namespace_packages
 
     :type: boolean
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -199,10 +199,11 @@ section of the command line docs.
 
 .. confval:: exclude
 
-    :type: comma-separated list of strings
+    :type: regular expression
 
-    A comma-separated list of file names, directory names or subpaths which mypy
-    should ignore while recursively discovering files to check.
+    A regular expression that matches file names, directory names and paths
+    which mypy should ignore while recursively discovering files to check.
+    Use forward slashes on all platforms.
 
     This option may only be set in the global section (``[mypy]``).
 
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -205,6 +205,8 @@ section of the command line docs.
     which mypy should ignore while recursively discovering files to check.
     Use forward slashes on all platforms.
 
+    For more details, see :option:`--exclude <mypy --exclude>`.
+
     This option may only be set in the global section (``[mypy]``).
 
 .. confval:: namespace_packages
--- a/docs/source/config_file.rst
+++ b/docs/source/config_file.rst
@@ -197,7 +197,7 @@ section of the command line docs.
 
     This option may only be set in the global section (``[mypy]``).
 
-.. confval:: ignore_path
+.. confval:: exclude
 
     :type: comma-separated list of strings
 
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -2,13 +2,11 @@
 
 import functools
 import os
-import re
-import sys
 
 from typing import List, Sequence, Set, Tuple, Optional
 from typing_extensions import Final
 
-from mypy.modulefinder import BuildSource, PYTHON_EXTENSIONS, mypy_path
+from mypy.modulefinder import BuildSource, PYTHON_EXTENSIONS, mypy_path, matches_exclude
 from mypy.fscache import FileSystemCache
 from mypy.options import Options
 
@@ -94,7 +92,7 @@ class SourceFinder:
         self.explicit_package_bases = get_explicit_package_bases(options)
         self.namespace_packages = options.namespace_packages
         self.exclude = options.exclude
-        self.verbose = options.verbosity >= 2
+        self.verbosity = options.verbosity
 
     def is_explicit_package_base(self, path: str) -> bool:
         assert self.explicit_package_bases
@@ -106,16 +104,15 @@ class SourceFinder:
         seen = set()  # type: Set[str]
         names = sorted(self.fscache.listdir(path), key=keyfunc)
         for name in names:
+            # Skip certain names altogether
+            if name in ("__pycache__", "site-packages") or name.startswith("."):
+                continue
             subpath = os.path.join(path, name)
 
-            if self.exclude:
-                subpath_str = os.path.abspath(subpath).replace(os.sep, "/")
-                if self.fscache.isdir(subpath):
-                    subpath_str += "/"
-                if re.search(self.exclude, subpath_str):
-                    if self.verbose:
-                        print("TRACE: Excluding {}".format(subpath_str), file=sys.stderr)
-                    continue
+            if matches_exclude(
+                subpath, self.exclude, self.fscache, self.verbosity >= 2
+            ):
+                continue
 
             if self.fscache.isdir(subpath):
                 sub_sources = self.find_sources_in_dir(subpath)
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -91,6 +91,7 @@ class SourceFinder:
         self.fscache = fscache
         self.explicit_package_bases = get_explicit_package_bases(options)
         self.namespace_packages = options.namespace_packages
+        self.ignore_path = options.ignore_path
 
     def is_explicit_package_base(self, path: str) -> bool:
         assert self.explicit_package_bases
@@ -103,9 +104,15 @@ class SourceFinder:
         names = sorted(self.fscache.listdir(path), key=keyfunc)
         for name in names:
             # Skip certain names altogether
-            if name == '__pycache__' or name.startswith('.') or name.endswith('~'):
+            if (
+                name in ("__pycache__", "site-packages", "node_modules")
+                or name.startswith(".")
+                or name.endswith("~")
+            ):
                 continue
             subpath = os.path.join(path, name)
+            if any(subpath.endswith(pattern.rstrip("/")) for pattern in self.ignore_path):
+                continue
 
             if self.fscache.isdir(subpath):
                 sub_sources = self.find_sources_in_dir(subpath)
--- a/mypy/find_sources.py
+++ b/mypy/find_sources.py
@@ -6,7 +6,7 @@ import os
 from typing import List, Sequence, Set, Tuple, Optional
 from typing_extensions import Final
 
-from mypy.modulefinder import BuildSource, PYTHON_EXTENSIONS, mypy_path, matches_ignore_pattern
+from mypy.modulefinder import BuildSource, PYTHON_EXTENSIONS, mypy_path, matches_exclude_pattern
 from mypy.fscache import FileSystemCache
 from mypy.options import Options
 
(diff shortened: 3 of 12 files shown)


mypy/test/test_find_sources.py
```diff
@@ -252,3 +252,77 @@ def test_find_sources_namespace_multi_dir(self) -> None:
...
+        # paths
+        options.ignore_path = ["/pkg/a1"]
+        finder = SourceFinder(FakeFSCache(files), options)
+        assert find_sources(finder, "/") == [
+            ("a2", "/pkg"),
+            ("a2.b.c.d.e", "/pkg"),
+            ("a2.b.f", "/pkg"),
+        ]
+
+        options.ignore_path = ["b/c"]
```
hauntsaninja, line 299: This is a test case I'm not sure about.
We could complicate the behaviour, so that paths with slashes in the beginning or middle are treated as fixed paths to ignore, as opposed to patterns. That is, if there's a slash at the beginning it's just an absolute path and we ignore it. If there's a slash in the middle, then we treat it as a fixed path relative to the current directory and ignore it.
As opposed to right now, where we just treat "b/c" as a suffix and use it to ignore "CWD/x/y/b/c" as well as "CWD/b/c"

intgr:
There's a bikeshedding opportunity here for (re)naming --ignore-path :)

Arbitrary survey of tools that I commonly use:
* black uses --exclude
* flake8 uses --exclude  (flake8 --ignore refers to ignoring error codes, not paths)
* pytest uses --ignore, --ignore-glob
* coverage.py uses --omit
* Some Unix tools (tar, rsync) use --exclude. But recursive `ls` uses --ignore

I find that he word "ignore" could also refer to scanning files and ignoring errors therein. Like the existing `--ignore-missing-imports` or `ignore_errors=` in mypy.ini

I would slightly prefer --exclude or --exclude-path, as there is less potential for this confusion.

hauntsaninja:
Sure, `--exclude` makes sense to me. The `path` in `--ignore-path` could be somewhat misleading, since maybe people could interpret that as a fixed path. I just went ahead with it as a slightly more general version of the `--ignore-dir` that @JukkaL proposed in the issue. I'll change it to whatever Jukka decides.

I'll also call out that many other tools support regexes or globs. We can do this too, but it's not critical and I figure it's something we could always choose to add support for later.

hauntsaninja:
Oh also I guess I never fully replied to https://github.com/python/mypy/issues/9981#issuecomment-769126126.
I'm also happy to provide a PR that reverts some of these changes... Although I feel fairly strongly that we should continue to recursively explore with `--namespace-packages` (despite that being ad hoc and not a complete workaround), since the behaviour is really quite broken without it and it'd be hard to communicate to users what's expected to work and what's not on which version.

JukkaL:
I like `--exclude`. I also think that regular expressions would be good to support, since adding support later can break existing exclude patterns, especially on Windows, since `\` is special in regular expressions. Regular expressions would probably also be faster to match, and they should not be hard to implement.

I actually like how black does this, and I'd prefer to have something fairly similar. Here are some thoughts:
- If we'd use regexps, then only a single value for `--exclude` would ever be needed, as `|` would work. This would simplify things a little.
- `--exclude` could override the default exclusions instead of building on top of them. This could be useful in case our defaults don't work for some people. This would also speed up matching, since there would always be exactly one pattern.
- By using `re.compile` beforehand, matching should be pretty fast.
- If we'd always use `/` for path separators (even on Windows), the same exclude patterns would work across platforms. This is also nice on Windows, since otherwise `\` would have to be escaped, which is awkward in a regexp.

For example, the default pattern could be something like `/(site-modules|node_modules|\.[^/]*)/`. This works assuming directory paths always include a `/` suffix. We'd also need to make sure there's also `/` as a prefix.

Other things that could be useful to ignore by default (and black ignores these):
- `build` (`setup.py` can create this)
- `dist` (`setup.py` can create this)

JukkaL:
Also, when using verbose logging (`-v`), it would be good to log all excluded paths.

intgr:
> regular expressions would be good to support, since adding support later can break existing exclude patterns

Alternatively regex excludes could be a separate option (--exclude, --exclude-regex). But I have no preference on this matter.

marctorsoc:
> I like `--exclude`. I also think that regular expressions would be good to support, since adding support later can break existing exclude patterns, especially on Windows, since `\` is special in regular expressions. Regular expressions would probably also be faster to match, and they should not be hard to implement.
> 
> I actually like how black does this, and I'd prefer to have something fairly similar. Here are some thoughts:
> 
> * If we'd use regexps, then only a single value for `--exclude` would ever be needed, as `|` would work. This would simplify things a little.
> * `--exclude` could override the default exclusions instead of building on top of them. This could be useful in case our defaults don't work for some people. This would also speed up matching, since there would always be exactly one pattern.
> * By using `re.compile` beforehand, matching should be pretty fast.
> * If we'd always use `/` for path separators (even on Windows), the same exclude patterns would work across platforms. This is also nice on Windows, since otherwise `\` would have to be escaped, which is awkward in a regexp.
> 
> For example, the default pattern could be something like `/(site-packages|node_modules|\.[^/]*)/`. This works assuming directory paths always include a `/` suffix. We'd also need to make sure there's also `/` as a prefix.
> 
> Other things that could be useful to ignore by default (and black ignores these):
> 
> * `build` (`setup.py` can create this)
> * `dist` (`setup.py` can create this)

I think there are two paths here:

* do the simple case now (pass a list of directories to ignore + exclude the typical directories most people want to ignore) and work on a better version with regex, etc in a followup PR
* or do the one with all features within this PR

I would advocate for the former, as even though it would create some dirtier configs, it allows delivering something faster. Regex can be added as a separate option, as @intgr suggests. WDYT?

JukkaL:
I'd rather avoid the need for two options. We already have a large number of command-line options, and making sure all combinations of options work well is nontrivial. 

I think that it's also fine to turn off the new behavior (at least when not using `--namespace-packages`) in the next release and have the full-featured implementation supporting more complex patterns in the release after that. We can possibly also ignore some hard-coded paths in the first version but not make it configurable.

hauntsaninja:
@JukkaL I've made the changes as described.
However, I think it might be better if exclude adds to the default exclusions, instead of overriding. Fiddling around with the slashes is actually a little finnicky and effectively we're creating boilerplate for everyone who wants to use this option. In all the mypy issues, I've never seen someone complain about mypy not checking things that start with a period. I also think always excluding site-packages is a good idea that'll save a lot of confused users (the error messages can be surprising).

JukkaL:
Thanks for the updates!

> However, I think it might be better if exclude adds to the default exclusions, instead of overriding. Fiddling around with the slashes is actually a little finnicky and effectively we're creating boilerplate for everyone who wants to use this option. In all the mypy issues, I've never seen someone complain about mypy not checking things that start with a period. I also think always excluding site-packages is a good idea that'll save a lot of confused users (the error messages can be surprising and quite non-obvious).
>
> Users can always re-include by explicitly specifying files on the command line; --exclude only applies to recursive discovery.

You've convinced me that this is a better approach. Here's a summary:
* Default exclusions are always enabled, but they can be overridden by giving `.py`/`.pyi` files on the command line, or by using `-p`/`-m` and MYPYPATH. They can't be overridden by passing a directory, such as `mypy site-packages/`, right?
* `--exclude` is used for excluding additional files beyond the defaults.
* Default exclusions should only include things that are very rarely a good idea to type check. `__pycache__`, `site-packages` and `node_modules` plus anything starting with a dot seem safe.

This also has the benefit that if we add more default exclusions, everybody will get them automatically, without having to modify their `--exclude` patterns.

docs/source/command_line.rst
```diff
@@ -49,6 +49,13 @@ for full details, see :ref:`running-mypy`.
     Asks mypy to type check the provided string as a program.
 
 
+.. option:: --exclude
+
+    A regular expression that matches file names, directory names and paths
+    which mypy should ignore while recursively discovering files to check.
+    Use forward slashes on all platforms.
```
JukkaL, line 56: Add one or two examples of typical use cases, such as ignoring a `build` or `venv` directory (venvs can contain `.py` files outside `site-packages` that users may want to exclude). Another example might be `/setup.py$` to avoid duplicate `setup.py` files.

Maybe also mention which names are always excluded?

Mention that giving a path (or `-p`/`-m`?) overrides exclusions, and give an example, such as `mypy site-packages/example/*.py`).

docs/source/config_file.rst
```diff
@@ -197,6 +197,16 @@ section of the command line docs.
...
 
+.. confval:: exclude
+
+    :type: regular expression
+
+    A regular expression that matches file names, directory names and paths
+    which mypy should ignore while recursively discovering files to check.
+    Use forward slashes on all platforms.
+
+    This option may only be set in the global section (``[mypy]``).
```
JukkaL, line 208: Add a link to the longer explanation in `--exclude`?

mypy/modulefinder.py
```diff
@@ -443,10 +444,23 @@ def find_modules_recursive(self, module: str) -> List[BuildSource]:
         names = sorted(self.fscache.listdir(package_path))
         for name in names:
             # Skip certain names altogether
-            if name == '__pycache__' or name.startswith('.') or name.endswith('~'):
+            if (
+                name in ("__pycache__", "site-packages", "node_modules")
+                or name.startswith(".")
+                or name.endswith("~")
+            ):
```
JukkaL, line 451: Do we still need this once we have the regular expression? Or if the default exclusions are always enabled, this can be left and the code path below is only taken if there are non-default exclusions.

It may be worth having this code in a utility function that gets called from here and `find_sources.py` to avoid duplication. Already the implementations are slightly different, and it's easy to forget to update one of them.

mypy/modulefinder.py
```diff
@@ -443,10 +444,23 @@ def find_modules_recursive(self, module: str) -> List[BuildSource]:
         names = sorted(self.fscache.listdir(package_path))
         for name in names:
             # Skip certain names altogether
-            if name == '__pycache__' or name.startswith('.') or name.endswith('~'):
+            if (
+                name in ("__pycache__", "site-packages", "node_modules")
+                or name.startswith(".")
+                or name.endswith("~")
```
JukkaL, line 450: I don't remember seeing directories with a `~` suffix, and files will be ignored based on the extension, so I think that this can be safely removed.

mypy/modulefinder.py
```diff
@@ -443,10 +444,23 @@ def find_modules_recursive(self, module: str) -> List[BuildSource]:
...
             subpath = os.path.join(package_path, name)
 
+            if self.options and self.options.exclude:
+                subpath_str = os.path.abspath(subpath).replace(os.sep, "/")
+                if self.fscache.isdir(subpath):
+                    subpath_str += "/"
+                if re.search(self.options.exclude, subpath_str):
+                    if self.options.verbosity >= 2:
+                        print("TRACE: Excluding {}".format(subpath_str), file=sys.stderr)
+                    continue
```
JukkaL, line 462: Can you share the functionality with `find_sources.py` by moving the code to a utility function?

mypy/find_sources.py
```diff
@@ -102,11 +106,17 @@ def find_sources_in_dir(self, path: str) -> List[BuildSource]:
         seen = set()  # type: Set[str]
         names = sorted(self.fscache.listdir(path), key=keyfunc)
         for name in names:
-            # Skip certain names altogether
-            if name == '__pycache__' or name.startswith('.') or name.endswith('~'):
-                continue
```
JukkaL, line 107: If default exclusions are always considered, we may need to preserve this code (also this doesn't consider `site-packages`, etc.).

mypy/test/test_find_sources.py
```diff
@@ -252,3 +252,72 @@ def test_find_sources_namespace_multi_dir(self) -> None:
...
+        files = {
+            "/pkg/a1/b/c/d/e.py",
+            "/pkg/a1/b/f.py",
+            "/pkg/a2/__init__.py",
+            "/pkg/a2/b/c/d/e.py",
+            "/pkg/a2/b/f.py",
+        }
+
+        # file name
+        options.exclude = "/f.py"
```
JukkaL, line 273: Add `$` to match at end of string?

mypy_self_check.ini
```diff
@@ -19,3 +19,4 @@ pretty = True
 always_false = MYPYC
 plugins = misc/proper_plugin.py
 python_version = 3.5
+exclude = /(mypy/typeshed|__pycache__|site-packages)/|/(\.[^/]+)/?$
```
JukkaL, line 22: Nice!

If we always enable the defaults, this can be simplified to `/mypy/typeshed/`.

mypy/test/test_find_sources.py
```diff
@@ -252,3 +252,72 @@ def test_find_sources_namespace_multi_dir(self) -> None:
...
+
+        files = {
+            "pkg/a1/b/c/d/e.py",
+            "pkg/a1/b/f.py",
+            "pkg/a2/__init__.py",
+            "pkg/a2/b/c/d/e.py",
+            "pkg/a2/b/f.py",
+        }
+        finder = SourceFinder(FakeFSCache(files), options)
+        assert len(find_sources(finder, "/")) == len(files)
```
JukkaL, line 323: Add a test case that uses a more complex regexp, such as `/(a1|a3)/`.

mypy/options.py
```diff
@@ -97,6 +97,10 @@ def __init__(self) -> None:
         # sufficient to determine module names for files. As a possible alternative, add a single
         # top-level __init__.py to your packages.
         self.explicit_package_bases = False
+        # File names, directory names or subpaths to avoid checking
+        self.exclude = (
+            "/(__pycache__|site-packages|node_modules)/|/(\\.[^/]+|[^/]+~)/?$"
+        ) # type: str
```
JukkaL, line 103: If defaults are always included, this can default to `None` (or empty string).

mypy/main.py
```diff
@@ -808,6 +809,15 @@ def add_invertible_flag(flag: str,
     code_group.add_argument(
         '--explicit-package-bases', action='store_true',
         help="Use current directory and MYPYPATH to determine module names of files passed")
+    code_group.add_argument(
+        "--exclude",
+        metavar="PATH",
+        default=options.exclude,
+        help=(
+            "Regex to match file names, directory names or paths to avoid checking. "
```
JukkaL, line 817: Nit: Regx -> Regular expression (for extra clarity)

If default can't be overridden, remove "Defaults to ..." and maybe update to say "additional file names, ...".

Maybe explicitly add something about "recursing into directories" since this doesn't affect paths that refer to files.

docs/source/command_line.rst
```diff
@@ -49,6 +49,13 @@ for full details, see :ref:`running-mypy`.
     Asks mypy to type check the provided string as a program.
 
 
+.. option:: --exclude
+
+    A regular expression that matches file names, directory names and paths
+    which mypy should ignore while recursively discovering files to check.
```
JukkaL, line 55: Maybe emphasize "recursively discovering" since this could be surprising.

JukkaL (commented):
Left a bunch of minor comments.

hauntsaninja:
Great! I changed things to not override the default exclusions, addressed the comments and added some more tests.

JelleZijlstra:
I'm a bit late to the party, but for reference Black also got a feature request to support extending the default exclusions (psf/black#1571). I agree that for mypy it makes sense to default to extending the default excludes.

docs/source/command_line.rst
```diff
@@ -49,6 +49,30 @@ for full details, see :ref:`running-mypy`.
...
+    Use forward slashes on all platforms.
+
+    For instance, to avoid discovering any files named `setup.py` you could
+    pass ``--exclude '/setup.py$'``. Similarly, you can ignore discovering
```
JukkaL, line 59: This would be more correct as `--exclude '/setup\.py$'` (with the escape).

mypy/test/test_find_sources.py
```diff
@@ -236,19 +244,119 @@ def test_find_sources_namespace_explicit_base(self) -> None:
...
+
+        # default
+        fscache = FakeFSCache({"/dir/a.py", "/dir/venv/site-packages/b.py"})
+        assert find_sources(["/"], options, fscache) == [("a", "/dir")]
+        with pytest.raises(InvalidSourceList):
+            find_sources(["/dir/venv/"], options, fscache)
+        assert find_sources(["/dir/venv/site-packages"], options, fscache) == [
+            ("b", "/dir/venv/site-packages")
+        ]
+        assert find_sources(["/dir/venv/site-packages/b.py"], options, fscache) == [
```
JukkaL, line 276: Test `node_modules` and a directory starting with a dot as well?

mypy/test/test_find_sources.py
```diff
@@ -236,19 +244,119 @@ def test_find_sources_namespace_explicit_base(self) -> None:
...
+        files = {
+            "/pkg/a1/b/c/d/e.py",
+            "/pkg/a1/b/f.py",
+            "/pkg/a2/__init__.py",
+            "/pkg/a2/b/c/d/e.py",
+            "/pkg/a2/b/f.py",
+        }
+
+        # file name
+        options.exclude = "/f.py$"
```
JukkaL, line 289: Nit: Use `r"/f\.py$"`.

JukkaL (approved):
Thanks for the updates! Looks good, just a few small nits. I'll probably do the next mypy release next week (Mon, Tue or Wed), if everything goes as planned. If you don't have time to update the PR before that, I can do the minor tweaks myself.

hauntsaninja:
Changes made!

MatthiasLohr:
Any idea when this get merged and released? mypy checks are broken for me since 0.800 and I need this option to fix this.

JukkaL (approved):
Thanks for the updates, looks great!

JukkaL:
@MatthiasLohr I'm going to start to prepare the release now. It should be out today/tomorrow, assuming I encounter no unexpected issues.

MatthiasLohr:
Awesome, thanks!

rgommers:
The way this option is implemented as a regex makes use of it in `mypy.ini` quite weird - a single regex is inconsistent with how `files` works, and doesn't look scalable on a large code base. Here's what I just had to do in a PyTorch PR:
```
files =
    torch,
    caffe2,
    test/test_bundled_images.py,
    test/test_bundled_inputs.py,
    test/test_complex.py,
    test/test_datapipe.py,
    test/test_expecttest.py,
    test/test_futures.py,
    test/test_numpy_interop.py,
    test/test_torch.py,
    test/test_type_hints.py,
    test/test_type_info.py,
    test/test_utils.py,
    tools/clang_format_utils.py,
    tools/clang_tidy.py,
    tools/generate_torch_version.py,
    tools/stats_utils/*.py

#
# `exclude` is a regex, not a list of paths like `files` (sigh)
#
exclude = torch/include/|torch/csrc/|torch/distributed/elastic/agent/server/api.py
```

Using `exclude` as a list of paths doesn't work. 

Is this intentional, and if not should I open a new issue?

hauntsaninja:
The initial implementation didn't use regexes, but you can see Jukka's reasoning for regexes here: https://github.com/python/mypy/pull/9992#issuecomment-770893221 (tldr; matches how other tools do it, people are going to want it and it'd be a breaking change later)
Note a list of exact paths would be annoying here, e.g. with exact paths you wouldn't be able to do things like exclude all files named "setup.py" no matter where in a large directory tree they might be, so we weren't ever going for exact "files"-like semantics.

I'd open an issue for this. I think a good path forward here for making long regex's readable is allowing a list of regexes / the flag to be specified multiple times, and just chaining them together with "or"s.

PR #9994, opened 2021-01-29
Consider variance when joining generic types

### LINKED ISSUE #5269: Join doesn't treat correctly variance of generics
Currently, joins for the instances of same type are calculated using this code:
```python
        if is_subtype(t, s) or is_subtype(s, t):
            args = []
            for i in range(len(t.args)):
                args.append(join_types(t.args[i], s.args[i]))
            return Instance(t.type, args)
```
This has two problems:
* For a covariant generic `G`, the join of `G[int]` and `G[str]`, is currently just `object`, not `G[object]` as it (I believe) should be.
* This behaviour is wrong for contravariant types, in this case there should be a meet of arguments: so that `join(C[int], C[float])` should be `C[int]` for contravariant `C`, while currently it is `C[float]`.

A correct behaviour would to check for every type argument position:
* if the class is covariant in it, calculate join of args on the left and right
* if the class is contravariant in it, calculate meet of args
* if the class is invariant in it, check that the type on the left and right are the same
* if the last one didn't hold, fall back to `object`.
* verify bounds and values restrictions for new args, if some are not satisfied, fall back to `object`.
* (optional) if one of the meets in third bullet produced an `UninhabitedType`, fall back to `object` as well.

freundTech:
### Description
Fixes #5269

This PR improves joining of generic types by considering the variance of their TypeVars.

TODO:
- [x] Notify zulip about the changes (https://github.com/zulip/zulip/pull/17194)
- [x] Notify starlette about the changes (Notified on gitter)
- [x] Zulip implemented a fix (https://github.com/zulip/zulip/pull/17194, https://github.com/zulip/zulip/pull/17213)
- [ ] starlette implemented a fix

## Test Plan

~~I haven't finished writing tests and also broke some other stuff. Therefor this PR is WIP~~
~~Tests should now be good to go.~~
~~This currently is broken on recursive types and therefor a draft.~~


mypy/meet.py
```diff
@@ -1,9 +1,6 @@
 from mypy.ordered_dict import OrderedDict
 from typing import List, Optional, Tuple, Callable
 
-from mypy.join import (
```
freundTech, line 4: This has to be replaced with `from mypy import join`, to prevent recursive import error.

test-data/unit/check-inference.test
```diff
@@ -791,10 +791,10 @@ AnyStr = TypeVar('AnyStr', bytes, str)
 def f(x: Union[AnyStr, int], *a: AnyStr) -> None: pass
 f('foo')
 f('foo', 'bar')
-f('foo', b'bar') # E: Value of type variable "AnyStr" of "f" cannot be "object"
+f('foo', b'bar') # E: Value of type variable "AnyStr" of "f" cannot be "Sequence[object]"
```
freundTech, line 794: This is due to `str` being a `Sequence[str]` and `bytes` being `Sequence[int]`, so they get joined to `Sequence[object]`.

test-data/unit/check-newsemanal.test
```diff
@@ -1970,7 +1970,7 @@ class G(Generic[T]): pass
 
 t: Tuple[G[B], G[C]] # E: Type argument "__main__.B" of "G" must be a subtype of "__main__.A" \
                      # E: Type argument "__main__.C" of "G" must be a subtype of "__main__.A"
-reveal_type(t.__iter__) # N: Revealed type is 'def () -> typing.Iterator[__main__.G*[__main__.B]]'
+reveal_type(t.__iter__) # N: Revealed type is 'def () -> typing.Iterator[builtins.object*]'
```
freundTech, line 1973: Previously mypy ignored the typevar not being a subtype of the bound.
Now `object` is inferred instead. This shouldn't matter, as this can only happen with as the result of an error.

test-data/unit/check-classes.test
```diff
@@ -5436,7 +5436,7 @@ main:8:6: error: Type argument "builtins.int" of "G" must be a subtype of "built
 [case testExtremeForwardReferencing]
 from typing import TypeVar, Generic
 
-T = TypeVar('T')
+T = TypeVar('T', covariant=True)
```
freundTech, line 5439: Previous behavior was incorrect. With this change `covariant` is needed to keep the old behavior

freundTech:
All the existing tests have now been changed to test for the new behavior.

freundTech:
I added some more tests and think that this PR should now be ready. Feel free to review it.

freundTech:
I just noticed that this PR creates a giant mypy primer diff, which the workflow silently fails to post:  #10011

Converting back to draft for now

freundTech:
> ```diff
> zulip (https://github.com/zulip/zulip.git)
> + zerver/lib/markdown/__init__.py:2044: error: Value of type "object" is not indexable  [index]

This is due to 
```python
realm_filters: List[Tuple[str, str, int]],
realm_filters_key: int,
email_gateway: bool,
# define default configs
self.config = {
    "realm_filters": [realm_filters,
                      f"Realm-specific filters for realm_filters_key {realm_filters_key}"],
    "realm": [realm_filters_key, "Realm id"],
    "code_block_processor_disabled": [email_gateway,
                                      "Disabled for email gateway"],
}
```
`config` was previously inferred as `Dict[str, List[object]` and is now inferred as `Dict[str, object]`
As lists are invariant and the dict joins different typed lists this now results in `object` instead of `List[object]`, as it would if lists were covariant.

Possible solutons:
* Use tuples instead of lists. `config` is then inferred as `Dict[str, Tuple[object, str]]` (recommended)
* Annotate `config` as `Dict[str, List[object]`
* Make it so inline list definitions can have their type changed by surrounding code to allow the code to pass as is (complicated, not recommended)

freundTech:
> starlette (https://github.com/encode/starlette.git)
> + starlette/testclient.py:159: error: Dict entry 10 has incompatible type "str": "Dict[str, Dict[<nothing>, <nothing>]]"; expected "str": "Sequence[object]"
> ```

This is due to the two definitions
```python
scope = {
    "type": "websocket",
    "path": unquote(path),
    "root_path": self.root_path,
    "scheme": scheme,
    "query_string": query.encode(),
    "headers": headers,
    "client": ["testclient", 50000],
    "server": [host, port],
    "subprotocols": subprotocols,
}
```
and
```python
scope = {
    "type": "http",
    "http_version": "1.1",
    "method": request.method,
    "path": unquote(path),
    "root_path": self.root_path,
    "scheme": scheme,
    "query_string": query.encode(),
    "headers": headers,
    "client": ["testclient", 50000],
    "server": [host, port],
    "extensions": {"http.response.template": {}},
    }
```
They have nothing to do with each other, but are in the same python scope. Previously they had the same type by chance. Now the types differ, because inferred types are more accurate.

Possible solutions:
* Rename one or both of the variables (recommended)
* Annotate the variable with a general enough type such as `Dict[str, object]`

hauntsaninja (approved):
Thanks for this PR, looks great to me! cc @JukkaL if you want to take a look
I'm guessing the recursive issue in practice is `str` is `Sequence[str]`?

PR #10020, opened 2021-02-04
Add explicit_package_bases to docs

Dreamsorcerer:
This value appears to be supported in the config file, but is missing from the docs.


hauntsaninja (commented):
Thanks, looks good once CI passes. The duplication between command line and config file documentation is a bit of a pain to deal with :-)

PR #10040, opened 2021-02-06
[mypyc] Add LoadLiteral and use tables to construct and store literals

JukkaL:
Redesign the implementation of constant `PyObject *` values, such as string and integer 
literals.

There are several related changes:
* Add `LoadLiteral` op for loading literal values instead of using `LoadStatic`.
* No longer store all literals in a table while we construct IR to make it easier to remove
  literals no longer needed after optimizations in the future.
* Simplify the pretty-printed IR for loading literals.
* Rename the helpers used to construct literal references (e.g. `load_static_unicode` -> 
  `load_str`).
* Collect all used literals values after we have final IR.
* Store literal value initialization information in C arrays instead of generating code for 
  each value.
* Store constructed literal objects in a single `CPyStatics` array.

The goal is to speed up compiles (slightly) and to make it easy to support tuple literals
and efficient table-driven import statements. Both of these can use the `CPyStatics` array.

Closes mypyc/mypyc#794. Closes mypyc/mypyc#797.

### DIFF
--- a/mypyc/analysis/dataflow.py
+++ b/mypyc/analysis/dataflow.py
@@ -7,7 +7,7 @@ from typing import Dict, Tuple, List, Set, TypeVar, Iterator, Generic, Optional,
 from mypyc.ir.ops import (
     Value, ControlOp,
     BasicBlock, OpVisitor, Assign, Integer, LoadErrorValue, RegisterOp, Goto, Branch, Return, Call,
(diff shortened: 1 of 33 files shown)


JukkaL:
I'm not waiting for code review since I have a bunch of other changes that depend on this. If somebody would like to give feedback, I will respond to it in a separate PR.

PR #10050, opened 2021-02-08
Fix #7445 Use double quotes in error messages

### LINKED ISSUE #7445: Use double quotes in error messages for more consistent colors
Some messages use double quotes for quoting while other use single quotes. Maybe we should move to double quotes since the colored output only highlights double quoted strings.

Examples where this happens:

```
t.py:14: note: Revealed type is 'builtins.list[builtins.str*]'
t.py:27: error: Name 'asdf' is not defined
```

PrasanthChettri:
### Description

Fixes #7445 
Replaces all the instances where single quotes are used in messages with double quotes around the ```mpypy/messages.py```

## Test Plan
Only some minor changes in one file, have not done tests, pytest seems to freeze when running  ```pytest -q mypy```(not a hardware constraint I can assure), would be helpful if anybody could help me with this

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -747,7 +747,7 @@ class MessageBuilder:
         self.fail("Unpacking a string is disallowed", context)
 
     def type_not_iterable(self, type: Type, context: Context) -> None:
-        self.fail('\'{}\' object is not iterable'.format(type), context)
+        self.fail('\"{}\" object is not iterable'.format(type), context)
 
     def incompatible_operator_assignment(self, op: str,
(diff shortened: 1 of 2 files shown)


PrasanthChettri:
Im getting an Invalid Ouput Assertion error during testing<br>
Detail
```
  E: ...note: Revealed type is 'Any'
  A: ...note: Revealed type is "Any"
```
From what I gather this is the <b> intended</b> output we want i.e. error messages with double quotes  but the tests given above are failing because of it. If anybody would give a feedback it will be helpful

PR #10051, opened 2021-02-08
Expand user home directory for cache_dir

m-zat:
### Description

Currently cache_dir does not expand user home directory, which results
in counter-intuitive behavior when setting for instance

```MYPY_CACHE_DIR='~/.cache/mypy/'```

In this case a folder named `~` is created in the cwd where mypy is
launched.

This commit modifies the behavior and expands the ~ into the user home
directory.

## Test Plan

Tested by setting `MYPY_CACHE_DIR=~/.cache/mypy` and checked that the folder is picked correctly.


JukkaL (approved):
Thanks, the new behavior seems more user friendly (and less surprising).

PR #10061, opened 2021-02-10
Clarify callback protocol example

MarcoGorelli:
### Description

xref https://stackoverflow.com/q/66068202/4451315

The current example, when run without the double underscore prefix, doesn't error. I found this a bit confusing, and wasn't sure what the example was meant to teach me.
I'd like to think that the proposed change clarifies it.

Like this, remove the double underscore prefix gives:
```console
$ mypy t.py 
t.py:14: error: Incompatible types in assignment (expression has type "Callable[[T], T]", variable has type "Copy")
t.py:14: note: "Copy.__call__" has type "Callable[[Arg(T, 'origin')], T]"
```

## Test Plan

None (just simple doc change)

### DIFF
--- a/docs/source/protocols.rst
+++ b/docs/source/protocols.rst
@@ -490,6 +490,7 @@ a double underscore prefix is used. For example:
 
    copy_a: Callable[[T], T]
    copy_b: Copy
+   copy_c: Callable[[T], T]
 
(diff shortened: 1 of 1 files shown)


MarcoGorelli:
I presume this isn't wanted, sorry for the noise

PR #10080, opened 2021-02-12
Generic protocol documentation fix

hackaugusto:
The protocol itself must be generic, otherwise type checking will fail.

Here is a small snippet to test the definition from the documentation:

```python
def f(g: Copy):
    g("")

def copy(_origin: str):
    pass

f(copy)
```

### DIFF
--- a/docs/source/protocols.rst
+++ b/docs/source/protocols.rst
@@ -485,7 +485,7 @@ a double underscore prefix is used. For example:
 
    T = TypeVar('T')
 
-   class Copy(Protocol):
+   class Copy(Protocol[T]):
        def __call__(self, __origin: T) -> T: ...
 
    copy_a: Callable[[T], T]


sobolevn:
Sorry, but this is not the case for two reasons.

1. This already works fine:
```python

T = TypeVar('T')

class Copy(Protocol):
    def __call__(self, __origin: T) -> T: ...

copy_a: Callable[[T], T]
copy_b: Copy

copy_a = copy_b  # OK
copy_b = copy_a  # Also OK

def f(g: Copy):
    g("")

def copy(origin: T) -> T:
    pass

f(copy)

```

Notice, that I have changed `def copy(_origin: str):` to be `def copy(_origin: T) -> T:`, because `def copy(_origin: str):` does not match `Callable[[T], T]` protocol.

2. `Protocol[T]` means a different thing. Because it makes `T` type var bound and makes it required to be filled when writting `Copy[]`, otherwise `Copy` would be just `Copy[Any]`.

In my opinion the example is correct as is and should not be changed.
Anyway, thanks a lot @hackaugusto for bringing this up! Generics can be confusing 🙂

hackaugusto:
Hey @sobolevn , thanks for the reply :)

> Notice, that I have changed def copy(_origin: str): to be def copy(_origin: T) -> T:,

If I understood your explanation correctly, "hiding" the type variable forbids the callers from binding to it, so it forces the user code to provide a generic function too. Is that correct?

To me the documentation doesn't make clear what the sample is achieving, and I don't think I understand what `copy(_origin: T) -> T` means well enough to propose a text change. To me that reads like the `id` function. 

> otherwise Copy would be just Copy[Any].

I don't understand what you mean, if a non-generic `Copy` is the same as `Copy where T=Any`, then it should be equivalent to `Callable[[T], T] where T=Any`, right?

If that is the case I don't understand why `f2(copy)` type checks while `f(copy)` doesn't:

```python
from typing import Any, Callable, TypeVar

T = TypeVar('T')

class Copy(Protocol):
    def __call__(self, origin: T) -> T: ...

def f(g: Copy):
    g("")

# I don't think this is 100% of a valid translation since the return type
# and the argument type are not the same, however I don't know how
# to express this without generics ... None could Any I just tried to make
# it explicit that they are not the same by using a different type
def f2(g: Callable[[Any], None]): 
    g("")

def copy(origin: str):
    pass

f(copy)
f2(copy)
```

I ran mypy 0.820+dev.2160eb5d1d60f8adce9ea1308c9a42a03d93f340.

Thanks!

JukkaL:
@sobolevn is right.

The type variable in the original example is bound to the `__call__` method, not to the protocol. This non-generic `Copy` protocol means something that can accept arbitrary objects.

If we make `Copy` generic over `T`, the meaning changes: now `Copy[str]` means something that can only accept `str` objects, and `Copy` means `Copy[Any]`, i.e. we can't do much useful type checking.

I'm closing this PR now, but I could accept another PR that explains the example in more detail.

PR #10086, opened 2021-02-12
use sys.executable instead of `python3` for --install-types

asottile:
### Description

This (for example) fixes the option on windows where `python3.exe` is either not a thing or it is a stub that opens the windows store

## Test Plan

run mypy with `--install-types` on windows


asottile:
here's a screenshot of the issue courtesy of @ethanhs

ethanhs (approved):
I found this issue while reviewing https://github.com/python/mypy/pull/10052. Anthony is quick on the draw :)

PR #10093, opened 2021-02-15
Speed up case sensitive is-file check in file system cache

JukkaL:
This speeds up resolving imports somewhat.

Add fast paths to `isfile_case`. On Linux we skip all logic, since
the file system is almost always case sensitive.

Cache results of parent directories as well.

Add tests.

### DIFF
--- a/mypy/test/testfscache.py
+++ b/mypy/test/testfscache.py
@@ -25,7 +25,7 @@ class TestFileSystemCache(unittest.TestCase):
         self.make_file('bar.py')
         self.make_file('pkg/sub_package/__init__.py')
         self.make_file('pkg/sub_package/foo.py')
-        # Run twice to test bost cached and non-cached code paths.
+        # Run twice to test both cached and non-cached code paths.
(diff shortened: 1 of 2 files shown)


mypy/test/testfscache.py
```diff
@@ -0,0 +1,84 @@
...
+            assert not self.isfile_case('Bar.py')
+            assert not self.isfile_case('pkg/sub_package/__init__.PY')
+            assert not self.isfile_case('pkg/Sub_Package/foo.py')
+            assert not self.isfile_case('Pkg/sub_package/foo.py')
+
+    def test_isfile_case_2(self) -> None:
+        self.make_file('bar.py')
+        self.make_file('pkg/sub_package/__init__.py')
+        self.make_file('pkg/sub_package/foo.py')
+        # Run twice to test bost cached and non-cached code paths.
```
ilevkivskyi, line 47: same typo here.

ilevkivskyi (approved):

PR #10094, opened 2021-02-15
Don't try to follow imports to encodings.*

JukkaL:
`encodings` uses module-level `__getattr__`.  By not trying to follow
imports to the submodules, resolving imports is slightly faster.

This is usually pretty marginal, but there can be over 100 suppressed
submodules, so this may actually be significant.

### DIFF
--- a/mypy/dmypy_server.py
+++ b/mypy/dmypy_server.py
@@ -708,8 +708,15 @@ class Server:
             all_suppressed |= state.suppressed_set
 
         # Filter out things that shouldn't actually be considered suppressed.
+        #
+        # Various submodules of 'encodings' can be suppressed, since it uses
+        # module-level '__getattr__'. Skip them since there are many of them,
+        # and following imports to them is kind of pointless.
+        #
         # TODO: Figure out why these are treated as suppressed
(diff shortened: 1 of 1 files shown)


mypy/dmypy_server.py
```diff
@@ -708,8 +708,15 @@ def find_added_suppressed(self,
...
+        #
+        # Various submodules of 'encodings' can be suppressed, since it uses
+        # module-level '__getattr__'. Skip them since there are many of them,
+        # and following imports to them is kind of pointless.
+        #
         # TODO: Figure out why these are treated as suppressed
-        all_suppressed = {module for module in all_suppressed if module not in graph}
+        all_suppressed = {module
+                          for module in all_suppressed
+                          if module not in graph and not module.startswith('encodings.')}
```
ilevkivskyi, line 719: Maybe make this a top level constant, so later we can skip similar "special" modules?

ilevkivskyi (approved):

mypy/dmypy_server.py
JukkaL: I'll define a function that contains the check (as a micro-optimization and it also allows other checks than starts with).

PR #10096, opened 2021-02-16
Add Python script to build wheels using cibuildwheel

JukkaL:
The contents are extracted from the current GitHub action definition:
https://github.com/mypyc/mypy_mypyc-wheels/blob/master/.github/workflows/build.yml

This is a fairly direct translation of the existing behavior. The
behavior should be identical to the current workflow.

The idea is to make this easier to maintain and easier to test
locally.


ilevkivskyi (approved):

JukkaL:
I'm not going to wait for the build, since the code isn't actually used anywhere yet.

Issue #10074, opened 2021-02-11
Windows wheel builds run out of memory


JukkaL:
Windows wheel builds started running out of memory. Here's an extract from the logs:
```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\bin\HostX86\x64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -IC:\Users\RUNNER~1\AppData\Local\Temp\pip-req-build-b3o88uhy\mypyc\lib-rt -Ibuild -Ic:\cibw\python\python.3.5.4\tools\include -Ic:\cibw\python\python.3.5.4\tools\include "-IC:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\ATLMFC\include" "-IC:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.28.29333\include" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\include\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\winrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\cppwinrt" /Tcbuild\__native_35888ab3ae83eb1dd0bf.c /Fobuild\temp.win-amd64-3.5\Release\build\__native_35888ab3ae83eb1dd0bf.obj /O2 /wd4102 /wd4101 /wd4146 /GL- /wd9025
  cl : Command line warning D9025 : overriding '/GL' with '/GL-'
  cl : Command line warning D9014 : invalid value '9025' for '/wd'; assuming '5999'
  __native_35888ab3ae83eb1dd0bf.c
  build\__native_35888ab3ae83eb1dd0bf.c(4558): fatal error C1060: compiler is out of heap space
  error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Enterprise\\VC\\Tools\\MSVC\\14.28.29333\\bin\\HostX86\\x64\\cl.exe' failed with exit status 2
[...]

PR #10097, opened 2021-02-16
Use a 64-bit C compiler on Windows to build wheels

JukkaL:
This will hopefully fix out of memory errors.

Work on #10074.


ilevkivskyi (approved):

JukkaL:
Unfortunately this didn't actually work, since the environment variable settings aren't propagated to the build.

JukkaL:
Reverted the commit. We need another solution for this.

PR #10098, opened 2021-02-17
docs: add using typing.no_type_check decorator to silence mypy

chrisbeardy:


JukkaL (commented):
In my experience `no_type_check` is used quite rarely, so mentioning it under "common issues" is probably not the best location. However, we could discuss it under "Silencing type errors" in `type_inference_and_annotations.rst`. Would you like to move the new section there?

chrisbeardy:
I added it to the common issues section as that section already shows two ways to silence type errors and the `no_type_check` is just a third. It makes sense to keep the ways to silence errors together. I'm not sure it makes sense as a standalone section in `type_interence_and_notations.rst`. I could create a new section altogether with all the different ways to silence the type errors?

hauntsaninja:
Thanks for you contribution! I agree with Jukka that this shouldn't be in common_issues, I opened https://github.com/python/mypy/pull/12713 to move this to type_interence_and_notations.rst

PR #10107, opened 2021-02-18
Reduce the size of the build by not compiling stubgen

JukkaL:
This might fix the out-of-memory errors we are getting when
building Windows wheels. In any case these will speed up
builds slightly.


ilevkivskyi (approved):

JukkaL:
The Travis build doesn't seem to be making any progress. I hope that it runs eventually.

PR #10108, opened 2021-02-18
Fixed Auto Discovery of .test files Fixes #8650

### LINKED ISSUE #8650: Auto-discovery of .test files
Currently all the `.test` files (used for mypy self tests) are enumerated in the test case runners, such as in `mypy/test/testcheck.py`. This is error-prone and can be confusing.

It would be better to do auto-discovery of test files using file name prefixes. For example, all the files with names like `check-*.test` would be executed by `mypy.test.testcheck`.

(Additionally, if there is an unrecognized `.test` file, it would be nice to generate an error and give hints about how to fix the issue. Addressing this is optional for now, however, and may be a bit tricky.)

MiHarsh:
Fixes #8650

After this PR is merged, this while running testcheck.py , it would automatically create the list of .test files. The main benefit is that if in future there is addition of test files, we would not have to add the new file into testcheck.py manually, it would automatically grab them.

Further I have verified the changes on my local.

Please merge this PR. I am new to open source and this is my first PR.

Please review @JukkaL , @ethanhs and @davidzwa Sir

### DIFF
--- a/mypy/test/testcheck.py
+++ b/mypy/test/testcheck.py
@@ -24,87 +24,28 @@ from mypy.semanal_main import core_modules
 
 
 # List of files that contain test case descriptions.
-typecheck_files = [
-    'check-basic.test',
-    'check-union-or-syntax.test',
-    'check-callable.test',
-    'check-classes.test',
-    'check-statements.test',
-    'check-generics.test',
-    'check-dynamic-typing.test',
-    'check-inference.test',
-    'check-inference-context.test',
-    'check-kwargs.test',
-    'check-overloading.test',
-    'check-type-checks.test',
-    'check-abstract.test',
-    'check-multiple-inheritance.test',
-    'check-super.test',
-    'check-modules.test',
-    'check-typevar-values.test',
-    'check-unsupported.test',
-    'check-unreachable-code.test',
-    'check-unions.test',
-    'check-isinstance.test',
-    'check-lists.test',
-    'check-namedtuple.test',
-    'check-narrowing.test',
-    'check-typeddict.test',
-    'check-type-aliases.test',
-    'check-ignore.test',
-    'check-type-promotion.test',
-    'check-semanal-error.test',
-    'check-flags.test',
-    'check-incremental.test',
-    'check-serialize.test',
-    'check-bound.test',
-    'check-optional.test',
-    'check-fastparse.test',
-    'check-warnings.test',
-    'check-async-await.test',
-    'check-newtype.test',
-    'check-class-namedtuple.test',
-    'check-selftype.test',
-    'check-python2.test',
-    'check-columns.test',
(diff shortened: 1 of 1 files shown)


mypy/test/testcheck.py
```diff
@@ -24,87 +24,28 @@
...
+
+# Define the path where the check files are present.
+if sys.platform == 'win32':
+    path2loc = os.path.abspath(__file__).split('mypy\\test')[0] + 'test-data\\unit'
+else:
+    path2loc = os.path.join(os.path.abspath(__file__).split('mypy/test')[0], 'test-data/unit')
+
+for check_files in os.listdir(path2loc):
+    # Append only if it has the pattern ( check-*-*.test)
+    if re.match(r'check-\w+-?\w+?.test', check_files):
```
kamilturek, line 37: FIlename `check-union-or-syntax.test` is not matched by this regex.

kamilturek (changes requested):

mypy/test/testcheck.py
MiHarsh: @kamilturek thanks for pointing this. I'll update the regex.

mypy/test/testcheck.py
```diff
@@ -24,87 +24,28 @@
...
-    'check-generic-alias.test',
-    'check-typeguard.test',
-]
+typecheck_files = []
+
+# Define the path where the check files are present.
+if sys.platform == 'win32':
+    path2loc = os.path.abspath(__file__).split('mypy\\test')[0] + 'test-data\\unit'
+else:
+    path2loc = os.path.join(os.path.abspath(__file__).split('mypy/test')[0], 'test-data/unit')
```
JukkaL, line 33: This is hard to reason about. Can you unify the cases so that the same implementation works on both Windows and other platforms? You can probably use `os.path.join` and/or `os.sep`.

mypy/test/testcheck.py
```diff
@@ -24,87 +24,28 @@
...
+
+for check_files in os.listdir(path2loc):
+    # Append only if it has the pattern ( check-*-*.test)
+    if re.match(r'check-\w+-?\w+?-?\w+?.test', check_files):
+        typecheck_files.append(check_files)
```
JukkaL, line 38: Can you move this and the above logic to a helper module, so that this can be reused in other places easily?

For example, we could have something like this in this file:

```
typecheck_files = expand_test_files('test-data/unit/check-*.test')
```

You can use `glob.glob` to expand patterns instead of `re`, since regular expressions are more error-prone as they require escaping, etc.

JukkaL (commented):
Sorry for the slow response! Left some thoughts about how to make this cleaner and more reusable.

hauntsaninja:
Superseded by #13143

PR #10110, opened 2021-02-19
Use multi-file builds on Windows

JukkaL:
This may reduce memory use and thus fix the builds.


ilevkivskyi (approved):

JukkaL:
Looks like we were already using this, so this had no effect. I will revert this.

PR #10121, opened 2021-02-20
Migrate runtests.py from os.system to subprocess

freundTech:
Alternative PR to #10052. This version uses `subprocess`.

Also removes code, which was unreachable since mypy requires python 3.5.

## Test Plan

The changed have been manually tested on linux and a windows VM with python installed to `C:\Program Files\Python39` (So the path contains a space).

It was also tested that failure of individual tests still lead to the whole test script failing.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -55,7 +55,7 @@ cmds = {
                                                           STUBGEN_CMD,
                                                           STUBGEN_PY])],
     # Test cases that may take seconds to run each
-    'pytest-slow': ['pytest', '-q', '-k', '%s' % ' or '.join(
+    'pytest-slow': ['pytest', '-q', '-k', ' or '.join(
         [SAMPLES,
          TYPESHED,
          PEP561,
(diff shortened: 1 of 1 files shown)


runtests.py
```diff
@@ -93,10 +85,12 @@ def run_cmd(name: str) -> int:
     status = 0
     cmd = cmds[name]
     print('run %s: %s' % (name, cmd))
-    res = (system(cmd) & 0x7F00) >> 8
-    if res:
+    proc = subprocess.run(cmd,
+                          stderr=subprocess.STDOUT,
+                          stdout=stdout)
```
hauntsaninja, line 90: stdout here is always sys.stdout right? Why do we need to specify this?

runtests.py
```diff
@@ -55,30 +44,33 @@
...
+    'lint': ['flake8', '-j0'],
     # Fast test cases only (this is the bulk of the test suite)
-    'pytest-fast': 'pytest -q -k "not (%s)"' % ' or '.join(ALL_NON_FAST),
+    'pytest-fast': ['pytest', '-q', '-k', 'not (%s)' % ' or '.join(ALL_NON_FAST)],
     # Test cases that invoke mypy (with small inputs)
-    'pytest-cmdline': 'pytest -q -k "%s"' % ' or '.join([CMDLINE,
-                                                         EVALUATION,
-                                                         STUBGEN_CMD,
-                                                         STUBGEN_PY]),
+    'pytest-cmdline': ['pytest', '-q', '-k', '%s' % ' or '.join([CMDLINE,
```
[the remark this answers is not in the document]
freundTech: fixed.

runtests.py
freundTech: Removed. I added it while debugging a problem, hoping it would fix it and forgot to remove it after finding the (unrelated) cause of the problem. See https://github.com/python/mypy/pull/10052#issuecomment-782621949 for details.

hauntsaninja (approved):
Thanks!

PR #10125, opened 2021-02-21
Adds ability to have subclasses for NodeVisitor and TraverserVisitor

sobolevn:
### Description

Currently a very important tool - which `NodeVisitor` definitely is - is not available to be used in a 3rd party code.
Because currently inheriting from `NodeVisitor` raises an exception: `TypeError: interpreted classes cannot inherit from compiled traits`

A developer has a single choice to replicate the same visitor patter by theyself, which is not really convenient.
So, I propose to make this consistent with `TypeVisitor`, which already allows having interpreted subclasses.

Refs https://github.com/python/mypy/commit/a9fa9ab43a2655cf81c66c065d51db4618839d05
Refs https://github.com/python/mypy/issues/9001
Refs https://github.com/python/mypy/pull/9602

## Test Plan

As https://github.com/python/mypy/pull/9602 have shown, this is fine without tests.
Correct me if I am wrong, please, and I would contribute the required tests. CC @msullivan on this.

### DIFF
--- a/mypy/traverser.py
+++ b/mypy/traverser.py
@@ -1,6 +1,7 @@
 """Generic node traverser visitor"""
 
 from typing import List
+from mypy_extensions import mypyc_attr
 
 from mypy.visitor import NodeVisitor
 from mypy.nodes import (
@@ -16,6 +17,7 @@ from mypy.nodes import (
 )
 
 
+@mypyc_attr(allow_interpreted_subclasses=True)
 class TraverserVisitor(NodeVisitor[None]):
     """A parse tree visitor that traverses the parse tree during visiting.
 
--- a/mypy/visitor.py
+++ b/mypy/visitor.py
@@ -3,7 +3,7 @@
 from abc import abstractmethod
 from typing import TypeVar, Generic
 from typing_extensions import TYPE_CHECKING
-from mypy_extensions import trait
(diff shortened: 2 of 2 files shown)


TH3CHARLie (approved):
I like this as long as it wouldn't cause serious performance loss due to larger generated code.

stuhood:
I don't think that this is working in Mypy `0.971`:
```python
from mypy.visitor import ExpressionVisitor

class SafePositionVisiter(ExpressionVisitor[None]):
    ...
```

...fails to import as a plugin with:
```
pyproject.toml:1:1: error: Error importing plugin "example":
interpreted classes cannot inherit from compiled traits  [misc]
```

CarlosRDomin:
Any update here? I'm on `0.991 (compiled: yes)` and still can't seem to subclass `TraverserVisitor` (which has `@mypyc_attr(allow_interpreted_subclasses=True)`), as well as `TypeTraverserVisitor` or `TransformVisitor` which for some reason (I'm not familiar with mypy at all) are not marked as allowing subclasses. In my view this kind of defeats the purpose of extending the original type support of MyPy through custom plugins, but maybe I'm just missing something?

stuhood:
It looks like https://github.com/python/mypy/pull/10125#issuecomment-1201590910 is due #754. Thanks @JukkaL!

PR #10128, opened 2021-02-22
Speed up processing of new files in daemon by caching ASTs

JukkaL:
Processing newly installed stub files, in particular, could be quite slow incrementally
in mypy daemon. This is because adding N files results in N steps interally, each of 
which adds one file. However, each step parses all remaining files, resulting in 
an O(n**2) algorithm.

For example, processing `six` stubs could take about 40s (when not using a 
compiled mypy).

Partially address the issue by caching parsed ASTs during a single increment.
This speeds up the `import six` use case by about 3x when not using a compiled
mypy. It's still about 3x slower when using daemon, however.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2049,8 +2049,10 @@ class State:
             else:
                 # Reuse a cached AST
                 self.tree = manager.ast_cache[self.id][0]
-                manager.errors.set_file_ignored_lines(self.xpath, self.tree.ignored_lines,
-                                                      self.ignore_all or self.options.ignore_errors)
+                manager.errors.set_file_ignored_lines(
+                    self.xpath,
+                    self.tree.ignored_lines,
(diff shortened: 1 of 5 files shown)


mypy/build.py
```diff
@@ -1994,8 +2003,14 @@ def parse_file(self) -> None:
             return
 
         manager = self.manager
+
+        # Can we reuse a previously parsed AST? This avoids redundant work in daemon.
+        cached = self.id in manager.ast_cache and True
```
ilevkivskyi, line 2008: Why do you need `and True`? Leftover from debugging?

ilevkivskyi (approved):

mypy/build.py
JukkaL: Yeah, I'll remove it.

Issue #10137, opened 2021-02-26
stubgenc.py throw TypeError: unbound method type.mro() needs an argument


gen-xu:
**Bug Report**

When I was trying to use stubgen to generate pyi for the PyAV package, an error is being thrown.
```
  File "mypy\stubgenc.py", line 315, in generate_c_type_stub
TypeError: unbound method type.mro() needs an argument
```
After a bit investigation in stubgenc.py and PyAV. I found the issue is caused by these
https://github.com/python/mypy/blob/master/mypy/stubgenc.py#L299
https://github.com/python/mypy/blob/master/mypy/stubgenc.py#L355
Basically, when invoking `obj.mro()` with `EnumType` defined in https://github.com/PyAV-Org/PyAV/blob/main/av/enum.pyx#L25, it resolves into `<type>.mro()` and complains missing argument, because `EnumType` inherits `type`. 
The minimum case of this would just be `python -c 'type.mro()'`

**To Reproduce**

1. Install PyAV with `pip install av` I am using 8.0.3 but most versions should work though
2. Install mypy with `pip install mypy` I am using 0.812 but most versions with `stubgen` should behave similar
3. Run command `stubgen -p av` or more specifically `stubgen -m av.enum`

**Expected Behavior**

It supposed to generate `pyi` file without throwing errors

**Actual Behavior**

```
PS C:\Users\Gen\Repos> stubgen -m av.enum
Traceback (most recent call last):
  File "c:\users\gen\appdata\local\programs\python\python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\gen\appdata\local\programs\python\python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\Gen\AppData\Local\Programs\Python\Python39\Scripts\stubgen.exe\__main__.py", line 7, in <module>
  File "mypy\stubgen.py", line 1582, in main
  File "mypy\stubgen.py", line 1475, in generate_stubs
  File "mypy\stubgenc.py", line 64, in generate_stub_for_c_module
  File "mypy\stubgenc.py", line 315, in generate_c_type_stub
TypeError: unbound method type.mro() needs an argument
```

**Your Environment**

- Mypy version used: 0.812
- Mypy command-line flags: `stubgen -m av.enum`
- Mypy configuration options from `mypy.ini` (and other config files): N/A
- Python version used: 3.9.1
[...]

PR #10138, opened 2021-02-26
Fix mro invoking #10137

gen-xu:
## Description
Fixes https://github.com/python/mypy/issues/10137
Changed `obj.mro()` to `type.mro(obj)` so it properly invokes the `mro` of `type` when `obj` is `type`
## Test Plan
After this change, `stubgen` can properly generate pyi files for cython modules or native modules that include types inheriting the builtin `type`. Any package include such definition of class would be suitable for testing this.
Added a unittest for `generate_c_type_stub` as well.


JelleZijlstra:
Thanks! Would it be possible to write a unit test?

gen-xu:
> Thanks! Would it be possible to write a unit test?

Added unit test for `generate_c_type_stub`

JelleZijlstra (approved):

PR #10153, opened 2021-02-28
[mypyc] Use vectorcalls to call Python objects

JukkaL:
The vectorcall APIs can make calls to callable Python objects
significantly faster when using keyword arguments.

This makes a microbenchmark that calls interpreted functions using
keyword arguments about 1.9x faster (Python 3.8). The main benefit is
from not needing to construct a dictionary for the keyword arguments.

This also gives a minor speedup when only using positional arguments.
The nested_func microbenchmark is now about 9% faster.

The main PyObject_VectorCall API is available on Python 3.8+ (with an
underscore prefix on 3.8). On Python 3.9+ we can also use a dedicated
method vectorcall function (PyObject_VectorcallMethod).

The new APIs can be used for calls that don't use caller *args or
**kwargs. For other calls we continue to use the legacy APIs. On older
Python versions we also continue use the slower legacy APIs.

See https://docs.python.org/3.9/c-api/call.html for more details.

Vectorcalls require arguments to be stored in a C array. Add the
RArray primitive type to support them. Also add the AssignMulti op to
initialize an array. By having a dedicated op for this makes it easy
to generate compact C code. This is useful since calls are common
operations.

The reference count transform can't reason about references held in
RArray values. Add a no-op KeepAlive op that is only used to ensure
that the operands aren't freed until after the call op. KeepAlive must
also be used with SetMem and LoadMem, for consistency. Previously
these ops had similar functionality built in.

Generate vectorcall ops in IR build test cases only when the test case
opts into them by specifying a recent enough Python version through a
_python3_8 test name suffix (or _python_3_9). This avoids having to
modify a large number of test cases.

### DIFF
--- a/mypyc/codegen/emitfunc.py
+++ b/mypyc/codegen/emitfunc.py
@@ -514,7 +514,11 @@ class FunctionEmitterVisitor(OpVisitor[None]):
             val = reg.value
             if val == 0 and is_pointer_rprimitive(reg.type):
                 return "NULL"
-            return str(val)
+            s = str(val)
+            if val >= (1 << 31):
+                # Avoid overflowing signed 32-bit int
+                s += 'U'
+            return s
         else:
             return self.emitter.reg(reg)
 
--- a/mypyc/codegen/emitfunc.py
+++ b/mypyc/codegen/emitfunc.py
@@ -12,7 +12,7 @@ from mypyc.ir.ops import (
     LoadStatic, InitStatic, TupleGet, TupleSet, Call, IncRef, DecRef, Box, Cast, Unbox,
     BasicBlock, Value, MethodCall, Unreachable, NAMESPACE_STATIC, NAMESPACE_TYPE, NAMESPACE_MODULE,
     RaiseStandardError, CallC, LoadGlobal, Truncate, IntOp, LoadMem, GetElementPtr,
-    LoadAddress, ComparisonOp, SetMem, Register, LoadLiteral, AssignMulti
+    LoadAddress, ComparisonOp, SetMem, Register, LoadLiteral, AssignMulti, KeepAlive
 )
 from mypyc.ir.rtypes import (
     RType, RTuple, RArray, is_tagged, is_int32_rprimitive, is_int64_rprimitive, RStruct,
@@ -497,6 +497,10 @@ class FunctionEmitterVisitor(OpVisitor[None]):
         dest = self.reg(op)
         src = self.reg(op.src) if isinstance(op.src, Register) else op.src
         self.emit_line('%s = (%s)&%s;' % (dest, typ._ctype, src))
+
+    def visit_keep_alive(self, op: KeepAlive) -> None:
+        # This is a no-op.
+        pass
 
     # Helpers
 
--- a/mypyc/codegen/emitfunc.py
+++ b/mypyc/codegen/emitfunc.py
@@ -59,6 +59,8 @@ def generate_native_function(fn: FuncIR,
     for r in all_values(fn.arg_regs, fn.blocks):
         if isinstance(r.type, RTuple):
             emitter.declare_tuple_struct(r.type)
+        if isinstance(r.type, RArray):
+            continue  # Special: declared on first assignment
 
         if r in fn.arg_regs:
             continue  # Skip the arguments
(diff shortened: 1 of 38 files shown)


mypyc/codegen/emitfunc.py
```diff
@@ -498,7 +516,11 @@ def reg(self, reg: Value) -> str:
             val = reg.value
             if val == 0 and is_pointer_rprimitive(reg.type):
                 return "NULL"
-            return str(val)
+            s = str(val)
+            if val >= (1 << 31):
+                # Avoid overflowing signed 32-bit int
+                s += 'U'
```
ilevkivskyi, line 522: Do we have a test for this?

mypyc/ir/rtypes.py
```diff
@@ -145,6 +143,12 @@ def accept(self, visitor: 'RTypeVisitor[T]') -> T:
     def serialize(self) -> str:
         return 'void'
 
+    def __eq__(self, other: object) -> bool:
+        return isinstance(other, RVoid)
+
+    def __hash__(self) -> int:
+        return 12345
```
ilevkivskyi, line 150: This looks weird, maybe use something like `hash(RVoid)` (the class object)?

mypyc/ir/rtypes.py
```diff
@@ -689,6 +731,46 @@ def is_optional_type(rtype: RType) -> bool:
...
+    be only used for local variables that are initialized in one location.
+    """
+
+    def __init__(self,
+                 item_type: RType,
+                 length: int) -> None:
+        self.item_type = item_type
+        # Number of items
+        self.length = length
+        self._ctype = 'xxx'
```
ilevkivskyi, line 747: Why this is needed? I mean the `xxx`.

mypyc/ir/ops.py
```diff
@@ -226,6 +228,37 @@ def accept(self, visitor: 'OpVisitor[T]') -> T:
...
 
+class AssignMulti(Op):
+    """Assign multiple values to a Register (dest = src1, src2, ...).
+
+    This is used to initialize RArray values. It's provided to avoid
+    very verbose IR for common vectorcall operations.
+
+    Note that this interacts atypically with reference counting. We
+    assume that each RArray register is initialized exactly once
+    with this op.
```
ilevkivskyi, line 239: Maybe this can be enforced somehow, for example have a class level set of destination registers, and check in `__init__` that the destination is not in this set?

ilevkivskyi (approved):

mypyc/codegen/emitfunc.py
```diff
@@ -488,6 +502,10 @@ def visit_load_address(self, op: LoadAddress) -> None:
         src = self.reg(op.src) if isinstance(op.src, Register) else op.src
         self.emit_line('%s = (%s)&%s;' % (dest, typ._ctype, src))
 
+    def visit_keep_alive(self, op: KeepAlive) -> None:
+        # This is a no-op.
+        pass
```
TH3CHARLie, line 507: A total unrelated comment here, I like the idea of having `KeepAlive`, but I wonder whether we should keep this analysis-only IR all the way to codegen. If more analysis-only (or mid-end only) IRs are introduced in the future, it may become reasonable to eliminate them first.

TH3CHARLie (approved):
Generally looks great to me and this is a great improvement! Left a comment on potential long-term improvement

mypyc/ir/ops.py
JukkaL: That's something we should do once we have general validation of the generated IR. However, this is probably not the best starting point for IR validation. Anyway, the C compilation will fail if we try to initialize an array twice.

mypyc/codegen/emitfunc.py
JukkaL: This actually gets stripped along the way. I'll try raising an exception if we reach here.

mypyc/codegen/emitfunc.py
JukkaL: Added unit test.

mypyc/ir/rtypes.py
JukkaL: Done.

mypyc/ir/rtypes.py
JukkaL: Removed it. It seems useless.

PR #10162, opened 2021-03-04
Support @no_type_check for classes

### LINKED ISSUE #607: Support no_type_check and no_type_check_decorator

kamilturek:
### Description

Closes #607.

Skip checking class scope if `typing.no_type_check` decorator has been applied on class definition.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -1720,7 +1720,10 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
             self.binder = ConditionalTypeBinder()
             with self.binder.top_frame_context():
                 with self.scope.push_class(defn.info):
-                    self.accept(defn.defs)
+                    if not any(refers_to_fullname(decorator,
+                                                  'typing.no_type_check')
+                               for decorator in defn.decorators):
+                        self.accept(defn.defs)
             self.binder = old_binder
             if not (defn.info.typeddict_type or defn.info.tuple_type or defn.info.is_enum):
                 # If it is not a normal class (not a special form) check class keywords.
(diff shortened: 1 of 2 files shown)


mypy/checker.py
```diff
@@ -1720,7 +1720,10 @@ def visit_class_def(self, defn: ClassDef) -> None:
             self.binder = ConditionalTypeBinder()
             with self.binder.top_frame_context():
                 with self.scope.push_class(defn.info):
-                    self.accept(defn.defs)
+                    if not any(refers_to_fullname(decorator,
+                                                  'typing.no_type_check')
+                               for decorator in defn.decorators):
+                        self.accept(defn.defs)
```
JukkaL, line 1726: Type checking populates some information that other parts of the type checker may rely on. So this needs to be more elaborate.

For example, this shouldn't generate errors:

```py
import typing

@typing.no_type_check
class A:
    a = int()

    def f(self) -> None:
        self.b = int()

a = A()
a.a  # Cannot determine type of "a"
a.b  # Cannot determine type of "b"
```

One way to deal with this is to give everything in the class the type `Any`. We'd still no type checking, but it actually wouldn't do much. We use `in_checked_function()` in `mypy.checker` to do this. You can add some check for `no_type_check` decorator there (maybe via a `bool` attribute flag) so that `in_checked_function()` returns False inside the class.

AlexWaygood:
This now has merge conflicts, and Jukka's review from two years ago still hasn't been addressed,so I'm closing this PR. If you're still interested in working on this issue, please feel free to open a new PR @kamilturek!

Issue #6740, opened 2019-04-29
Stop saying mypy is beta software?


msullivan:
We're tagged as alpha in pypi and in our readme we say "Mypy is alpha software, but it has already been used in production for well over a year at Dropbox, and it has an extensive test suite."

While mypy certainly has its flaws I, I think it is pretty well established that it is usable in a serious way, and in fact we pretty seriously recommend it as such. 

EDIT (by Ethan): We now say we are beta, should we say we are production ready?

ilevkivskyi:
I think we can call it beta. Some people define beta when software is feature-complete w.r.t. subsequent major release, but IIUC this is rather a tradition than a specification.

gvanrossum:
We're being very self-deprecating here. I would not hesitate calling mypy production quality, i.e. beyond beta.

ethanhs:
I agree at minimum that mypy is beta quality, if not better. When adding mypy typing to a project, a friend was surprised mypy hasn't hit 1.0 yet! I do think we should explain that backwards incompatible changes may still happen, but we try not to do so?

Michael0x2a:
It'd be nice if we could align the announcement that mypy is now production-quality with a somewhat impressive release. Maybe we should just silently switch over to calling ourselves "beta" for now and wait one or two releases before announcing that mypy is production-quality/that we're releasing v1.0? We could perhaps combine that announcement with the "we have a new semantic analyzer" release or something.

This is only a suggestion though -- I don't really have a strong opinion on this apart from agreeing that we're definitely past the alpha stage.

JukkaL:
@Michael0x2a I like your idea!

wasilukm:
Hi, could you please consider also switching to semantic versioning?

ethanhs:
@wasilukm AIUI the plan is to switch to semver at 1.0. So if we go ahead with @michael0x2a's proposal that will be in a couple of releases.

With regards to Michael's plan, my only concern is that the new semantic analyzer should probably be battle tested for a couple of releases, as I worry we'd release 1.0 then have several bugs crop up, which would weaken the stability message of a 1.0.

JukkaL:
Applying semantic versioning is not trivial, since we'd need to document what the public API of mypy is. The command-line interface is well documented, but it's not clear which APIs plugins can safely depend on.

[...]

PR #10169, opened 2021-03-05
Enable local partial types by default

JukkaL:
Local partial types prevent inferring the variable type from two
separate scopes. For example, this is no longer accepted:

```
class A:
    x = []  # Need type annotation here

    def f(self) -> None:
        self.x.append(1)
```

Also, initialization with just `None` is usually not allowed and
requires a type annotation:

```
x = None  # Error
y: None = None  # OK
z: Optional[str] = None  # OK
```

The mypy daemon only supports local partial types. This makes the
daemon and non-daemon modes behave consistently.

Add the `--no-local-partial-types` flag to revert to the old behavior.

Also make local partial types behavior closer to the old behavior
in a few cases:

- allow `x = f(...)` even if the return type of `f(...)` is `None` when
  `f` is generic
- allow `x = None` in an enum class body
- allor for loop index variable to have an inferred `None` type

Related to #6740.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -133,9 +133,9 @@ TypeRange = NamedTuple(
         ('is_upper_bound', bool),  # False => precise type
     ])
 
-# Keeps track of partial types in a single scope. In fine-grained incremental
-# mode partial types initially defined at the top level cannot be completed in
-# a function, and we use the 'is_function' attribute to enforce this.
+# Keeps track of partial types in a single scope. By default, partial
+# types initially defined at the top level cannot be completed in a
+# function, and we use the 'is_function' attribute to enforce this.
 PartialTypeScope = NamedTuple('PartialTypeScope', [('map', Dict[Var, Context]),
                                                    ('is_function', bool),
                                                    ('is_local', bool),
@@ -1715,7 +1715,8 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
         for base in typ.mro[1:]:
             if base.is_final:
                 self.fail(message_registry.CANNOT_INHERIT_FROM_FINAL.format(base.name), defn)
-        with self.tscope.class_scope(defn.info), self.enter_partial_types(is_class=True):
+        with self.tscope.class_scope(typ), self.enter_partial_types(is_class=True,
+                                                                    is_enum=typ.is_enum):
             old_binder = self.binder
             self.binder = ConditionalTypeBinder()
             with self.binder.top_frame_context():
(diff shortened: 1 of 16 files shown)


mypy/checker.py
```diff
@@ -2868,9 +2882,14 @@ def infer_variable_type(self, name: Var, lvalue: Lvalue,
 
             self.set_inferred_type(name, lvalue, init_type)
 
-    def infer_partial_type(self, name: Var, lvalue: Lvalue, init_type: Type) -> bool:
+    def infer_partial_type(self, name: Var, lvalue: Lvalue, init_type: Type,
+                           init_expr: Expression) -> bool:
         init_type = get_proper_type(init_type)
         if isinstance(init_type, NoneType):
+            if isinstance(init_expr, CallExpr):
```
ilevkivskyi, line 2889: In PR description, you say it is only OK for generic functions, is this still true? Or a different error will be shown for non-generic? Add a more detailed comment here maybe.

ilevkivskyi (approved):

mypy/checker.py
JukkaL: For functions that always return `None` we generate error about a called function not returning a value, which happens elsewhere and isn't related to this code, which only deals with type inference.

JukkaL:
The new errors from mypy primer are expected. However, it may be possible to improve some of the error messages.

mypy/options.py
```diff
@@ -277,7 +277,7 @@ def __init__(self) -> None:
         self.dump_deps = False
         self.logical_deps = False
         # If True, partial types can't span a module top level and a function
-        self.local_partial_types = False
```
bl2ck2, line 280: True

bl2ck2 (commented):
True
[manylinux-main.zip](https://github.com/python/mypy/files/11337812/manylinux-main.zip)

wyattscarpenter:
It seems that this proposed change to mypy degrades its ability to infer types for the user, so I am not sure that this is a good change.

According to https://github.com/python/mypy/issues/6740#issuecomment-562194989 the motivation seems to be 

> Make --local-partial-types enabled the default (or maybe always). This is necessary to make daemon and non-daemon modes behave in a similar way.

https://github.com/python/mypy/issues/10634 suggests the motivation is that the partial type inference doesn't actually work correctly anyway.

I think that the daemon and the partial type inference should be fixed, instead.

JukkaL:
> I think that the daemon and the partial type inference should be fixed, instead.

That's a fair point. We should at least have another look at improving the daemon first. Last time I looked at this, it seemed too complex to fix, but even putting major effort into the implementation could be a good idea if it avoids migration effort from many thousands of users.

ilevkivskyi:
@wyattscarpenter 
The problem is not just with the daemon, non-local partial types cause three other important problems, they:
* Prevent significant speed-up of parallel checking by checking interfaces first.
* Significantly complicate implementation of deferring module top-levels.
* Cause spurious action at a distance, where class interface can depend on order of method definitions.

Maybe there are more things. Non-local partial types are conceptually problematic because they make interface dependent on implementation. This is IMO very wrong, we should not have added them in the first place.

wyattscarpenter:
It seems like a great feature to me for a non-manifestly-typed/duck-typed/gradually-typed language such as Python. Of course, there are implementation issues as well.

ilevkivskyi:
"seems" is a key word here: non-local partial types is one of those things that look like a cool feature to have superficially, but then turn out to be problematic. It is not much about spirit of the language, but more about requiring library APIs to be more explicit/local (and note many entries in `mypy_primer` are from libraries).

Btw, another important thing is `--allow-redefinition-new` requires local partial types as well. And we want to eventually make the former default behavior (because that one is definitely a pythonic behavior).

ilevkivskyi:
@JukkaL Btw, I checked what is the current impact from `--local-partial-types` few days ago. It looks like 90-95% of errors are from use case like `<type> | None`. Allowing this one particular use case in `--local-partial-types` should not cause problems for parallel type checking (because we can easily and reliably detect affected method targets during semantic analysis). Not sure about the daemon, but probably should be not too hard to support it there either. If this is what you meant when we spoke last time, I think this is actually a reasonable idea.

About the empty list/dict/etc use cases, I think they are just not worth the effort taking into account they cause only 5-10% of the impact together.

hauntsaninja:
(in case useful, you can find an up-to-date primer at https://github.com/hauntsaninja/mypy/pull/9 )

hauntsaninja:
Superseded by #21163 (and other previously merged PRs)

PR #10175, opened 2021-03-07
Refactor push type map

sobolevn:
### Description

While reading throught the `mypy`'s source code these two things caught my eye:
1. `push_type_map` was re-implemented in-place for `check_for_comp`
2. ~~`push_type_map` had special handling for case when `type_map` argument is `None` which should not be possible, because `type_map: 'TypeMap'` and `no_implicit_optional = True` and `strict_optional = True` are on.~~ Ok, I have missed that `TypeMap` alias is defined as `Optional`, excluded.

## Test Plan

Existing tests should cover this.


JukkaL (approved):
Thanks, looks good.

PR #10179, opened 2021-03-07
[mypyc] Updates to introduction in the documentation

JukkaL:
Improve clarity and add a comparison against Cython.


thomasjohns:
Looks good to me (for what its worth)! I remember the first time I heard about mypyc I was curious about how it compared to Cython, so having this in the introduction makes a lot of sense to me.

PR #10181, opened 2021-03-07
dmypy - warn instead of failing if report generation is configured

### LINKED ISSUE #10118: Have dmypy gracefully ignore report generation settings
**Feature**

If mypy.ini/setup.cfg enable report generation, running dmypy fails with an error
```
dmypy: start/restart cannot generate reports
```

Can we update dmypy to skip report generation and let the user know their report settings are being ignored?

**Pitch**

Currently, my project has
```ini
[mypy]
cobertura_xml_report = build/mypy/
```
Every time I want to run `dmypy`, I have to...
1. comment out the cobertura_xml_report setting in setup.cfg, 
2. run dmypy, and 
3. undo and/or remember not to commit my change to setup.cfg when I am ready to make a pull request.

This is a little nuisance I'd love to fix if you're okay taking a PR to make this non-fatal.

natemcmaster:
### Description
Log a warning instead of failing when `dmypy run` is called on a project that has report generation configured via setup.cfg

Resolves #10118 

## Test Plan

1. Built package locally
2. Installed into my virtualenv
3. Created test project with mypy.ini contents:
    ```ini
    [mypy]
    html_report = build/html
    ```
1. Ran `dmypy run .`

Result:

```
dmypy run .
dmypy: ignoring report generation settings. Start/restart cannot generate reports.
Daemon started
```


JelleZijlstra:
Why do we even need a warning? I'd expect `dmypy` to simply ignore any config file settings related to reports. However, if you pass an explicit command line argument around report generation to `dmypy`, we should generate an error.

I might be wrong here though since I don't regularly use either the daemon or reports. Here are my assumptions:
- the daemon doesn't know how to generate any sort of report
- users who are interested in reports would run non-daemon mypy to generate a report occasionally, but run `dmypy run` more frequently in order to do type checking. Such users would want to put their report generation settings in `mypy.ini`.

natemcmaster:
@JelleZijlstra my usage of dmypy and reports is:

* I want reports for CI builds which use the non-daemon mypy to help track progress towards full type annotation coverage
* I am using dmypy because the VS Code integration with mypy and Pycharm type checker still just aren't good enough yet, so dmypy is the fastest way to quickly check type hints in my large codebase

While trying to fix this, I initially thought I would try to make a bigger change, but I found that the code which raises this warning/error doesn't know the difference between an explicit command line option and a config file option, nor does it know which dmypy subcommand is being called. Both are possible to fix with some refactoring, but figured this small tweak solved my papercut and was a good starting point for more discussion.

mypy/dmypy_server.py
```diff
@@ -135,10 +135,10 @@ def process_start_options(flags: List[str], allow_sources: bool) -> Options:
         ['-i'] + flags, require_targets=False, server_options=True
     )
     if options.report_dirs:
-        sys.exit("dmypy: start/restart cannot generate reports")
+        print("dmypy: ignoring report generation settings. Start/restart cannot generate reports.")
```
JukkaL, line 138: Nit: Use capitalized 'Ignoring'.

mypy/dmypy_server.py
```diff
@@ -135,10 +135,10 @@ def process_start_options(flags: List[str], allow_sources: bool) -> Options:
...
     if options.junit_xml:
-        sys.exit("dmypy: start/restart does not support --junit-xml; "
-                 "pass it to check/recheck instead")
+        print("dmypy: ignoring report generation settings. "
+              "Start/restart does not support --junit-xml. Pass it to check/recheck instead")
```
JukkaL, line 141: Nit: Use capitalized 'Ignoring'.

I think it would be better to generate this warning and set `options.junit_xml` to `False`, even when using check/recheck.

natemcmaster: Updated. Thanks for the feedback, @JukkaL

PR #10183, opened 2021-03-07
Make Python 2 support an extra

### LINKED ISSUE #10124: Make Python 2 support an "extra" in 0.900
The motivation for this is that typed-ast is a pain to deal with (it's just broken again on latest Python 3.10, and was a pain to deal with for Python 3.9, e.g. preventing Windows installs for some months) and Python 2 usage is ever dwindling*. Since mypy 0.900 will basically force all users to revisit how they install mypy, it seems like a good time to force the remaining users of mypy with Python 2 to change to `pip install 'mypy[python2]'`. 

@JukkaL thoughts?

*pypistats.org tells me Python 2 usage is down 40% in the last six months (based on Python 2 downloads of six)

hauntsaninja:
Resolves #10124

### DIFF
--- a/docs/source/cheat_sheet.rst
+++ b/docs/source/cheat_sheet.rst
@@ -13,6 +13,12 @@ language represents various common types in Python 2.
    many of the examples have a dual purpose: show how to write the
    annotation, and show the inferred types.
 
+.. note::
+
+   To check Python 2 code with mypy, you'll need to install mypy with
+   ``pip install 'mypy[python2]'``.
+
+
 
 Built-in types
 **************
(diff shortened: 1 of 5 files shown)


JukkaL (approved):
Thanks! Looks good. As users update to Python 3.8+, eventually most mypy users won't need to install typed-ast anymore.

freundTech:
I just noticed that this PR adds the `python_version<'3.8'` environment marker to `install_requires`, but not to `mypy-requirements.txt`.

Is this on purpose?

This is inconsistent with the `psutil` requirement which is also in `extras_requires`, but is not in `mypy-requirements.txt`. It is only in `test-requirements.txt`

PR #10191, opened 2021-03-09
Support for python 3.10 match statement

freundTech:
### Description

This PR will add support for the python 3.10 match statement to mypy.

Current progress:
- [x] parser and ast nodes
- [x] semantic analysis
- [x] type-checking
    - [x] as-pattern
    - [x] or-pattern
    - [x] literal-pattern
    - [x] value-pattern
    - [x] capture-pattern
    - [x] wildcard-pattern
    - [x] sequence-pattern
    - [x] mapping-pattern
    - [x] class-pattern
- [x] Lots of tests

## Test Plan

Some tests have been added, but they don't pass yet. There will be a lot more tests in the future.

I'm currently developing and testing on cpython 3.10.0b1 and will make sure to update to newer versions every now and then.


mypy-requirements.txt
```diff
@@ -1,6 +1,6 @@
 typing_extensions>=3.7.4
 mypy_extensions>=0.4.3,<0.5.0
-typed_ast>=1.4.0,<1.5.0
+typed_ast>=1.4.0,<1.5.0; python_version<'3.8'
```
freundTech, line 3: typed_ast currently can't be compiled on python 3.10. This has already been fixed upstream (https://github.com/python/typed_ast/pull/155), but there hasn't been a release yet. Until there has been we just need to not install it.

brandtbucher:
Just curious: does our decision to reuse expression nodes as pattern nodes in the Python AST cause any trouble for you?

Our intuition is that creating duplicates of all of the expression nodes used by patterns would be less convenient than the current design, but we're open to feedback if not.

brandtbucher:
(Also, let me know if you have any questions or want reviews on this. I'm very much looking forward to mypy's support for this feature!)

freundTech:
@brandtbucher The reuse of expression nodes does create some problems, but it shouldn't be too bad.

I currently have two solutions in mind, but haven't decided on which would be better:

1. Duplicate the nodes in mypy: currently the mypy ast nodes very closely follow the python ast nodes. It would be possible to just duplicate the needed nodes in mypy and choose the correct one while converting from python ast to mypy ast.
2. Duplicate the visitors: It would also be possible to just create new visitors wherever needed. Then for example in SemanticAnalyzer `visit_match_stmt` wouldn't call `pattern.accept(self)`, but `pattern.accept(SemanticPatternAnalyzer(self))`. This would allow us to keep the mypy ast closer to the python ast and wouldn't produce any extra work for cases where we don't have to differentiate between patterns and expressions (like the already implemented `StrConv`)

Reviews and other feedback is always welcome, no matter how early.
If I push it it's ready for review 😉

freundTech:
I decided to go with the first option and create separate ast nodes for patterns. This will probably be a bit more work at the beginning, but will lead to better code and prevent problems down the line.

So far I created nodes for as-patterns, or-patterns and literal patterns.

Sometimes converting from python ast to my pattern ast isn't trivial.
For example `1+2j` is a valid literal pattern while `2j+1` is a syntax error (when used as a pattern). Python would represent them both as `BinOp(x, Add, y)`, just with the x and y swapped.

freundTech:
@brandtbucher With the ast conversion phase of this now being finished here's some more feedback on the reuse of expression nodes:

I ended up adding nodes for patterns to mypy and, during ast conversion, converting from expression nodes to pattern nodes.

This was a bit harder than ast conversion would have been if python had dedicated pattern nodes.
The biggest problem in my opinion was static types for expression nodes often being much broader than for the corresponding pattern nodes.

Take a mapping pattern for example:
```python
case {'k': v}:
    pass
```

This is represented as an ast.Dict, which, according to static types, can have arbitrary expressions as keys. When used as a pattern it can only have literal patterns and value patterns (a.k.a. constant expressions and member expressions) as keys though. This leads to me having to check or cast the types in order to make type checking work, when at runtime the invalid values are already caught by the python parser and don't even reach my code.

Another problem were literal patterns. A literal pattern, expressed as expression nodes, can be a Constant, UnaryOp or BinOp. At the same time not all Constant, UnaryOp and BinOp expressions are valid literal patterns. This makes conversion quite complicated.


I can understand why you went with not duplicating the nodes, but, at least for this usecase, it would have been a lot more comfortable if python would offer separate pattern nodes.

brandtbucher:
Thanks for your thoughts!

When I find the time, I might draw up a POC branch with dedicated pattern nodes to see how much that complicates or simplifies CPython's internals. If its a clear win both there *and* here, then we might as well just do it.

I'll take a closer look at what you've done here first. As you note, handling for things like mappings and numeric literals might be simplified dramatically.

freundTech:
I'm currently working on the type-checking portion of this and would like to hear some opinions on how "assigning" to capture variables might work.

PEP 634 allows capturing to existing variables. In this case we would have to make sure that the captured value fits the type.

PEP 634 also says that captured variables can be used after the match statement, which means that we need to produce a type that works for all match cases.

If we use the normal rules then something like this would not be allowed:
```
match m:
    case int(x):
        pass
    case str(x):
        pass
```
Instead two different variable names would have to be used.

Alternatively we could type x as the union or join of all the types it captures and narrow it within the case block. This would be more flexible for the user.

brandtbucher:
Not a typing expert, but I would expect the type of `x` to be inferred as `int` in the first case, `str` in the second, and `int | str` after.

...although I *can* see the possible benefit of requiring an explicit `x: int | str` before the `match` block to prevent errors. In that case, though, the type should still be narrowed in each case.

(We're sort of in uncharted territory here, since the names are not scoped to the case block like in many other languages with similar constructs.)

JelleZijlstra:
@freundTech The equivalent nonmatch code would be something like
```
if isinstance(m, int):
    x = m
elif isinstance(m, str):
    x = m
```
And that currently produces `Incompatible types in assignment (expression has type "str", variable has type "int")`. So the consistent thing would be to also produce an error for the `match` code you posted.

I feel like @brandtbucher's proposed semantics are more useful though, so I'd prefer to implement that.

gvanrossum:
Pyright already supports pattern matching. What does that do?

freundTech:
I just checked the behavior in pyright and found the following:

1. Pyright doesn't yet support type inference for class patterns, therefor in the above example x is inferred as `Unknown`
2. Taking another example
```python
a: List[int] = ...
match a:
    case [b, *x]:
        reveal_type(b)
    case b:
        reveal_type(b)

reveal_type(b)
````
I get `int`, `List[int]` and `int | List[int]`. So basically 

3. In pyright this is consistent with what it infers for @JelleZijlstra's example. There it infers `int | str`. In mypy this behaviour wouldn't be consistent with the if/elif behaviour.

My current opinion is that going with mypy's if/elif behaviour is probably too restrictive, so there's basically three options:

1. Being inconsistent between if/elif and match is ok. Use `Union` for match
2. Being inconsistent is not ok. Keep the existing semantics
3. Being inconststent is not ok. Make if/elif work like in pyright

I would prefer 1, because 2 is probably to restrictive and you are much more likely to use variables assigned in if/elif blocks after those blocks than you are to use capture variables after the match, so pyright's behaviour could hide potential errors.

In this case variables assigned inside the case blocks should still follow the if/elif behaviour though and the union behaviour should only apply to capture patterns.

**Comment by freundTech** [comment]
A second point of discussion would be the behaviour of Or Patterns. Here we again have the option of either allowing different types and typing the variable as the union of them or enforcing that capture patterns in or patterns have the same type.

Python already enforces that all subpatterns of or-patterns capture the same names, therefore it would be natural to also enforce that they have the same types. Pyright currently does not enforce this.

**Comment by gvanrossum** [comment]
Okay, I think it's fine to be inconsistent, and inferring a union when the paths merge seems fine.

There are some refinements possible for control flow which will certainly be requested by some users, but we can put those off till later. E.g.
```
def f(a: List[int]):
  match a:
    case [x, 0]: ...
    case x: return
  reveal_type(x)  # Ideally, int
```

**Comment by freundTech** [comment]
Here's the initial version of type checking. I also rebased on master to take advantage of the Travis CI run against python nightly.

It's still a bit rough around the edges, but most things work and it has a lot of tests.
Most importantly the following doesn't yet work:
* Self matching classes
* Inferring __match_args__ for dataclasses and NamedTuples

After PEP 653 has either been accepted or rejected I will also need to do some changes to the handling of sequence and mapping patterns. They currently follow neither PEP 634, nor the PEP 653 exactly.

To see the current progress I would recommend looking at https://github.com/python/mypy/blob/3effadc803c106361edb7dcd24660799e3b10327/test-data/unit/check-python310.test. All tests in that file except for `testClassPatternCaptureSelf` currently work.

**Comment by freundTech** [comment]
Here's another problem I've run into:
Matching generic classes with a class pattern.

Consider the following code:
```
class A(Generic[T]):
    a: T

m: object

match m:
    case A(a=1):
        reveal_type(m)
```

Obviously we can narrow m to type A, but that should we infer the typevar T as?

The simple but wrong answer would be `int`.
The complicated answer is it depends on the variance of T.
* For invariant T we can never infer T as anything other than`Any`.
* For contravariant T we can infer the most general type a sub pattern of type T is matched to (in this case `int`). But also matching contravariant typevars wouldn't make any sense, as contravariant usually means write-only while matching a pattern requires a read.
* For covariant T we could only infer anything more precise than `object` if all occurrences of the TypeVar are matched by the class pattern.
In the example above we could infer `int`, but if we just add another attribute `b: T` to A, then suddenly we can't.

Due to how complicated those rules are and the fact that inferring invariant typevars is impossible anyways I would go with inferring `Any` for all typevars for now.

I also though about possible changes to the match statement itself to enable matching generic classes, but due to the fact that generic types are not always present at runtime I don't think that's possible.

I'm not yet sure if mypy should report an error if generic typevars were inferred as `Any`.

I will however make sure that already known typevars are propagated:

```
m: A[int]

match m:
    case A(a=i):
        reveal_type(i)  # int
```

**Comment by brandtbucher** [comment]
@freundTech, you might be interested in sharing your thoughts in [this issue](https://bugs.python.org/issue43892).

**Comment by ncoghlan** [comment]
For the problematic cases that @freundTech mentions above, my draft AST in the issue Brandt linked would make the following changes:

* MatchMapping nodes instead of Dict nodes. Keys are still nominally arbitrary expressions at the AST level to avoid side effects on the compilation of other code
* MatchValue nodes wrapped around value patterns and literal patterns that compare by equality
* MatchConstant nodes for literal patterns that compare by identity

**Comment by JukkaL** [comment]
> Due to how complicated those rules are and the fact that inferring invariant typevars is impossible anyways I would go with inferring Any for all typevars for now.

Agreed. This is something we can refine later if this seems to be a problem, but I don't expect it to be.

> I'm not yet sure if mypy should report an error if generic typevars were inferred as Any.

I don't think that we need to report an error. Currently `isinstance` does something vaguely similar and doesn't generate an error:

```py
def f(x: object) -> None:
    if isinstance(x, list):
        reveal_type(x)  # list[Any]
```

erictraut:
@freundTech said that Pyright doesn't yet support type inference for class patterns. It actually does have full support for class patterns. I want to make sure that my understanding of class patterns is correct because it doesn't match some of what was said above.

@JelleZijlstra said:

```python
match m:
    case int(x):
        pass
    case str(x):
        pass
```

is equivalent to:
```python
if isinstance(m, int):
    x = m
elif isinstance(m, str):
    x = m
```

That doesn't match my understanding. 

From [PEP 634's description of Class Patterns](https://www.python.org/dev/peps/pep-0634/#class-patterns), the `isinstance` checks are correct, but the `x=m` statements are not. By my reading, `x` is a positional pattern, so it must be converted to a keyword pattern by consulting the `__match_args__` attribute on the class. Since `int` and `str` don't have such an attribute (at least not described in typeshed's stubs currently), there's no way to map the positional pattern to a keyword pattern. That's why Pyright infers an "Unknown" type in this case.

Is my understanding correct, or an I misreading the PEP?

Here's a code sample that's consistent with my understanding:
```python
class A:
    __match_args__ = ('a', 'b')
    a: int
    b: str

class B:
    __match_args__ = ('d', 'e')
    d: complex
    e: float

def func1(m: A | B):
    match m:
        case A(x):
            reveal_type(x) # int
            pass
        case A(b=x):
            reveal_type(x) # str
            pass
        case B(x):
            reveal_type(x) # complex
            pass
        case B(e=x):
            reveal_type(x) # float
            pass

    reveal_type(x) # int | complex
```

erictraut:
Never mind, I just noticed that there's a special case specified in the PEP for certain built-in types.

> As mentioned above, for the following built-in types the handling of positional subpatterns is different: bool, bytearray, bytes, dict, float, frozenset, int, list, set, str, and tuple.

I'll need to add support in Pyright for these special cases.

I also didn't notice that `__match_args__` was auto-generated for dataclass and namedtuple. I'll need to add that as well.

test-data/unit/check-python310.test
```diff
@@ -0,0 +1,807 @@
...
+    case [a]:
+        reveal_type(a)  # N: Revealed type is "builtins.int*"
+[builtins fixtures/list.pyi]
+
+[case testSequencePatternCapturesStarred]
+from typing import Iterable
+m: Iterable[int]
+
+match m:
+    case [a, *b]:
```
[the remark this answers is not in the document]
freundTech: Yes. My sequence and mapping pattern implementations aren't final yet, as PEP 653 (https://www.python.org/dev/peps/pep-0653/. Discussed here: https://mail.python.org/archives/list/python-dev@python.org/thread/YYIT3QXMLPNLXQAQ5BCXE4LLJ57EE7JV/#OAVTGQCFXNX47JH52ZMOAHVISDZ2JNMO) might make some changes to them.
I just did a rough implementation that matches neither semantics correctly and will adjust it once PEP 653 has been either accepted or rejected.

freundTech:
> @freundTech said that Pyright doesn't yet support type inference for class patterns. It actually does have full support for class patterns.

Ok. Good to know. I only tested with those special cases, as they are faster to write and because they weren't handled I just assumed that class patterns in general weren't handled. Sorry about that.

freundTech:
@brandtbucher @ncoghlan I just finished porting my code to the new pattern nodes. There were a few cases where pythons pattern nodes handled things differently than the mypy pattern nodes I had written, so I changed the mypy nodes to be closer to the python nodes.

Overall this saved 221 lines (+317 -538) and made parts of the code a lot cleaner. Thanks for the work on this.

ncoghlan:
@freundTech Thanks for confirming it helped make things clearer!

cpontvieux-systra:
Up ? 3.10rc2 is released and 3.10 final will be out in less than a month.

freundTech:
I'm still working on it. There's still a few edge cases left to handle, but I don't have that much time at the moment.
I'm doing this as part of my bachelors thesis and am currently spending the majority of my time on writing the actual thesis part.

I can't promise that I'll finish it before the python 3.10 release, but I'll try.

JukkaL:
@freundTech If there are only a few remaining edge cases that need to be fixed, it may be totally fine to create issues for them and they can fixed after this initial PR has been merged (and you don't need to commit to fixing them soon -- somebody else may be able to help). This assumes that none of them are major regressions to existing functionality.

cpontvieux-systra:
Works so far for my code using this PR.

freundTech:
@JukkaL I think it would be possible to merge this PR in the state it currently is in. I would just need to quickly push a commit to mark two tests as skip (CI passed, as it doesn't test with python 3.10 yet).

Current problems are:
* Exhaustiveness checking doesn't yet work correctly (https://github.com/python/mypy/pull/10191/commits/9d5d600114208dd7985a15f0aa92444946a2a82b#diff-9100781e1e8bb8997aa611cce8014d8680677e036e9652d59fbd86336f80c3feR1023-R1049). I have some code that fixes it, but haven't pushed it yet, as it causes a major regression in another part of match statement checking. This is the main holdup at right now.
* mypyc doesn't support match statements yet. I just made it output an error message instead of crashing when encountering a match statement (https://github.com/python/mypy/pull/10191/files#diff-4b30e16444eb14783107d506b5799b6141a2abddcfb10957f95b8dddcd5d38b9R183-R185). This is something someone else might be better at implementing, as I have no experience with the mypyc codebase yet.
* There are not test for match statements involving guards yet.

JelleZijlstra:
Thanks @freundTech! Here are my opinions on the holdups you identified:
* Exhaustiveness checking is important, but I think we can live without it initially. We can merge this PR without it, then open an issue to fix it later.
* mypyc support can also wait; most mypy users won't be using mypyc to compile their code.
* We should have some minimal tests for guards to make sure they don't crash mypy or anything like that. It's OK if narrowing doesn't work fully correctly yet, though.

freundTech:
Here's an edge case I've run into, that's a big problem for correctness.
When encountering a literal-pattern like this:
```python
m: object
match m:
    case 1:
        reveal_type(m)
```
I currently narrow `m` to `Literal[1]`. This isn't always correct, as m could also have a custom `__eq__` method that returns `True` for input `1`.
When going for 100% correctness a literal-pattern (except for True, False and None, which compare by identity) couldn't narrow anything, because it doesn't know anything about the `__eq__` method. This would however greatly impact usability in common cases in exchange for correctness in what are probably pretty rare edge cases.

For this reason I would suggest narrowing to literals anyways. This is also what pyright does.
If anyone has a different opinion on this please let me know.

97littleleaf11:
@freundTech There are several merge conflicts now, could you please fix them?

freundTech:
I noticed the merge conflicts and already started working on them, but was held up for a bit by one particular conflict.

I hope I can find the time to fix the rest this or next week.

JukkaL:
@freundTech This would be great to have in the next release after 0.920, even if not everything is working perfectly yet. I'm planning to cut the release branch either tomorrow (Tue 14 Dec) or the day after (Wed 15 Dec), and if you can get this up to date by then it should be possible to include this in the release.

gvanrossum:
Sorry to butt in here. Nice work. Over at Pyright we received user feedback about what should happen in this situation:
```py
def f(item: Any):
    match item:
        case str():
            # What is item's type? Any or str?
```
We concluded that this should be `str`, similar to how `if isinstance(item, str)` would be treated. (See https://github.com/microsoft/pyright/issues/2674.) If that's what you've coded up here, great. If not, consider doing that in a future PR (I don't want to hold up the 0.920 release!)

freundTech:
> @freundTech This would be great to have in the next release after 0.920, even if not everything is working perfectly yet. I'm planning to cut the release branch either tomorrow (Tue 14 Dec) or the day after (Wed 15 Dec), and if you can get this up to date by then it should be possible to include this in the release.

I didn't quite manage to finish fixing the conflicts today. Most things work, but two tests still fail. One introduced by this PR and one introduced by #11273. I will try getting them to work tomorrow, but this might need to be pushed to the release after that.

marcopullo:
FYI a field report as I've been extensively using the latest `freundTech/mypy@feature-match-stmt-exhaustiveness` branch (not this PR) in some personal toy projects. That branch combined with pattern matching and use of `assert_never` pattern to enforce exhaustive pattern matching has been working well on medium size examples (~5k lines of code). _(I cannot use a language without robust sum types, type narrowing, and exhaustive pattern matching — exactly all the things this PR finally pulls together for Python!)._ It's amazing!

However not sure all the changes in `freundTech/mypy@feature-match-stmt-exhaustiveness` have landed in `freundTech/mypy@feature-match-stmt` (this PR) or some type narrowing in the latter got broken — when switching from `feature-match-stmt-exhaustiveness` to `feature-match-stmt`  to the latter, mypy reports incorrect type errors due to not properly narrowing the type being matched.

See this example that works perfectly & correctly in the `feature-match-stmt-exhaustiveness` branch but reports the following incorrect errors in `feature-match-stmt`:

```python
def assert_never(value: NoReturn) -> NoReturn:
    raise AssertionError()

# Expected: no mypy error
def test_type_narrowing(x: int|str) -> bool: # Actual mypy: Missing return statement
    match x:
        case int():
            # Expected: Revealed type is "builtins.int"
            reveal_type(x) # Actual: Revealed type is "Union[builtins.int, builtins.str]"
            return True
        case str():
            # Expected: Revealed type is "builtins.str"
            reveal_type(x) # Revealed type is "Union[builtins.int, builtins.str]"
            return False
        case _:
            # Expected: no mypy error
            assert_never(x) # Actual: Argument 1 to "assert_never" has incompatible type "Union[int, str]"; expected "NoReturn"
```

Happy to provide more details as needed if that is not expected in the current `feature-match-stmt` branch.

Really looking forward to this landing into master — until then I've been extremely happy with `feature-match-stmt-exhaustiveness` in my personal projects as it allows Python to be a really ergonomic optionally typed language.

Awesome work @freundTech!

brandtbucher:
> ```py
> def assert_never(value: NoReturn) -> NoReturn:
>     raise AssertionError()
> ```

Hm, what does it even mean for `value` to be typed as `NoReturn` (and why do you need `value` here at all)?

henryiii:
This is a standard completeness check trick. If it's possible to reach "value" statically (that is, the type doesn't completely narrow), then this errors when typing checking. The raise is just in case. See https://github.com/python/mypy/issues/5818#issuecomment-431863917 from 2018 or https://hakibenita.com/python-mypy-exhaustive-checking. It was sadly never added directly, ironically, because match statements were going to be the future of completeness checking. It's still really useful, though.

brandtbucher:
But PEP 484 says that [even placing the annotation there is invalid](https://www.python.org/dev/peps/pep-0484/#the-noreturn-type). Is it being used as some hack-ier form of this?

```py
import typing

class Nothing:
    pass

def assert_never(x: Nothing) -> typing.NoReturn:
    assert False, x

del Nothing
```

gvanrossum:
Well, your class 'Nothing` still has instances. `NoReturn` is semantically an empty union, which is theoretically nice because it has no values. `NoReturn` is just a funny name.

henryiii:
Quick example:

```python
from typing import Literal, NoReturn


Platform = Literal["linux", "macos", "windows"]


def assert_never(x: NoReturn) -> NoReturn:
    raise AssertionError()


def f(platform: Platform) -> str:
    if platform == "linux":
        return "linux"
    elif platform == "macos":
        return "macos"
    else:
        assert_never(platform)
```

```console
$ mypy tmp.py --python-version=3.9
tmp.py:16: error: Argument 1 to "assert_never" has incompatible type "Literal['windows']"; expected "NoReturn"
```

If you end up with an empty union via narrowing, either by removing "windows" or adding a handler for it, then this passes. I do admit the name NoReturn is weird, and I rather wish an alias for it (for nicer error messages) + assert_never was available in typing, as it's a [really handy](https://github.com/search?l=Python&q=assert_never&type=Code) idiom.

marcopullo:
Also see similar manual annotation in Typescript's [never](https://patrickdesjardins.com/blog/typescript-exhaustive-check) to get the same exhaustive check behavior. As henryiii noted this ideally would be built-in and automatic like other languages (Rust, Haskell, etc.) and not require the manual annotation. In the meantime, the benefit of exhaustive checking on Unions far outweighs any manual annotation cost (by orders of magnitude!) IMO, so I always do it when available.

brandtbucher:
Hm, I guess it's just a limitation of my own (admittedly weak) understanding of type theory then. My working knowledge has always been that `NoReturn` is not *actually* a type, it's just a way of marking code following any call to that function as unreachable. PEP 484 seems to support that interpretation.

Using it as an annotation to mean "no type is valid here" seems nonsensical to me, through that lens. I understand the "empty union" argument a bit more clearly now, but it still feels like the `value: NoReturn` idiom is just relying on internal `mypy` implementation details.

marcopullo:
brandtbucher the empty type is a valid (and useful!) type. These are used in typed languages when you want to tell the typechecker something is impossible to inhabit and that you should never be able to instantiate/execute it.  e.g. see [absurd](https://stackoverflow.com/questions/14131856/whats-the-absurd-function-in-data-void-useful-for) in Haskell.

It goes by different names in different languages, but it's clearer to think about how many values can inhabit the type...
`Void` = the type of 0 instances allowed (e.g. `NoReturn` in mypy, `never` in Typescript, etc.)
`Unit` = the type of 1 possible instance (e.g. `Nothing` in mypy, `null` in Typescript, etc.)
`Bool` = 2 possible instances (e.g. `bool` in mypy, etc.)
et al.

In this case we want to communicate to the type checker a certain branch of code should never be executable. We do that by saying "create a thing which is uncreatable" which leads to a contradiction (aka type error).

brandtbucher:
Thanks, that explanation helps... and just to clarify, where you said `Nothing` above, did you mean `None`?

I definitely understand the value of exhaustiveness checking; I've just never seen it done this way before in Python, and the fact that it works at all completely broke my existing mental model of `NoReturn`.

But I'm also still pretty confident that this usage isn't legal in Python, so I'm not sure if I should bother updating my mental model at all. 🙃

srittau:
> Well, your class `Nothing`still has instances.`NoReturn`is semantically an empty union, which is theoretically nice because it has no values.`NoReturn` is just a funny name.

I think we should add `Never` as an alias to `NoReturn` to make it less weird.

marcopullo:
Correct — fixed comment.

I agree [PEP 484](https://www.python.org/dev/peps/pep-0484/#the-noreturn-type) does have an example of `NoReturn` in argument position as an error. In this case though I think the spec should be updated, rather than trying to conform with the spec which would only do harm as far as I can tell. 

IMO ideally 3 things happen:
1. Update spec to allow `NoReturn` as a valid empty type, including in a function argument position
2. Rename or alias or add it as a `Never` type (or other more descriptive name) to reduce confusion
3.  Automatically have this in `match` (and ideally other condition flow statements) so users don't need the manual annotation to begin with and can ignore all of these details for 99% of their code

gvanrossum:
Can someone remind me what the "manual annotation" (that you wish wasn't
necessary) refers to? Which part of the example is the "manual annotation"?

henryiii:
You have to opt into completeness checking currently - apparently even for match statements, which I thought was the reason a "nice" way to do completion checking was never added - it was supposed to come "free" with match statements.

That is,

```python
Platform = Literal["linux", "macos", "windows"]

def f(platform: Platform) -> str:
    match platform:
        case "linux":
            return "linux"
        case "macos":
            return "macos"
```

is currently not an error for not handling the "windows" case. I think it was/is still planned? I don't think it can be "automatic" on the other cases, like if statements, which is why a nicer way to write this would be nice. (`from typing import assert_never` or something).

marcopullo:
That last `case _: assert_never(…)` part is what I'd ideally not have to write to ensure mypy is checking that my `match` has covered all cases and gives me an error if I did not. 

Today I have to write
```python
def test(x: int | str) -> None:
    match x:
        case int():
            pass
        case str():
            pass
        case _:             # This is the manual annotation to get an exhaustive check
            assert_never(x) # This is the manual annotation to get an exhaustive check
```

Ideally I'd instead write:
```python
def test(x: int | str) -> None:
    match x:
        case int():
            pass
        case str():
            pass
```

See Rust's [match](https://doc.rust-lang.org/book/ch06-02-match.html#matches-are-exhaustive) or Haskell's `case` that give exhaustiveness checking without adding something like this `assert_never` branch manually.

gvanrossum:
Okay, so "manual annotation" refers to the need to add `case _: assert False` (or some other assertion). (An odd use of the term "annotation" in the context of Python type checking. :-)

We considered and rejected runtime exhaustiveness checking in an [earlier version of the `match` spec](https://www.python.org/dev/peps/pep-0622/#check-exhaustiveness-at-runtime).

But static checkers like mypy should be able to figure out on their own that a match for a union type that has cases covering all the union members is complete, shouldn't it? So this sounds like purely a mypy issue if it doesn't (in this case).

marcopullo:
Correct, item ([3](https://github.com/python/mypy/pull/10191#issuecomment-1002291833)) for removing the manual `assert_never` branch to get the exhaustive check is about the _static_ exhaustive check done by mypy (as implemented in this PR in its current state).

Per [freundTech's comment](https://github.com/python/mypy/pull/10191#issuecomment-933327249) earlier in this thread I think it is already their intention🤞.

> (An odd use of the term "annotation" in the context of Python type checking. :-)

Apologies for the confusion, I hadn't realized it already had a specific meaning from the PEP's.

erictraut:
@gvanrossum, are you saying that you think static type checkers should always enforce exhaustion for every match statement?

I've asserted that there are legitimate uses of a `match` statement without exhaustion, just like there are legitimate uses of a `case` statement in C/C++/Javascript without a `default` case.If you agree with that assertion, then a developer must indicate that exhaustion is intended. Adding an `assert_never` is the logical way to indicate this, although it could probably be made more ergonomic.

If, on the other hand, you think that non-exhaustive `match` statements leave too much room for bugs, then static type checkers could always enforce it and require developers to provide a wildcard or other "catch all" case statement within every `match`.

Do you have a viewpoint on this? I'd prefer not to leave it up to some optional flag that each type checker needs to expose. If there is adequate consensus that exhaustion is always required for `match`, then all static type checkers should always enforce it.

My experience with other languages is what led me to the assumption that there are legitimate uses that don't involve exhaustion, but the Python community could decide otherwise. This is a common source of bugs, after all.

Since you decided not to enforce exhaustion at runtime, my (slight) preference is to follow that precedent and not enforce it in static type checkers by default. We could explore more ergonomic ways to do this than the current `assert_never` or `assert False` techniques.

marcopullo:
Unsolicited 2 cents, in no particular order:

1. Absolutely there are many use cases without exhaustion. Languages that default static exhaustion (e.g. Rust) have simple ways to opt out as needed. Personally, I prefer to be explicit in those cases rather than the other way around. e.g. my exhaustively checked mypy codebase does this via a `case _: pass` line the times when I want to opt out.
2. So it's just which situation do we want the programmer to be more explicit about their intention when using `match`, exhaustive uses or non-exhaustive uses.
3. As a general principle, over time it's easier to relax constraints than strengthen them after the fact (read: easier to opt-out of strict checks later versus trying to impose them after you've written a bunch of code).
4. That there is no runtime exhaustive check by default may be all the more reason in favor of default static exhaustive checking, because there are no runtime guardrails to depend on.
5. There already exists an idiomatic statement (`if-elif-else`) that is default non-exhaustive.
6. All that said, not sure why static exhaustion checking would be treated differently than other [strict options](https://mypy.readthedocs.io/en/stable/config_file.html#miscellaneous-strictness-flags) under `mypy --strict` that static analyzer users would opt into via a flag

gvanrossum:
> @gvanrossum, are you saying that you think static type checkers should always enforce exhaustion for every match statement?

No. This seems to be the position of some others on this thread. My position is only that *if* the type checker can prove that the cases *are* exhaustive (e.g. when there are unguarded cases covering all members of an enum or union), *and* all the case blocks in the match statement raise or return a value, the type checker should know that the statement following the match is not reachable (which has consequences for e.g. "missing return statement" errors).

> I've asserted that there are legitimate uses of a `match` statement without exhaustion, just like there are legitimate uses of a `case` statement in C/C++/Javascript without a `default` case. If you agree with that assertion, then a developer must indicate that exhaustion is intended. Adding an `assert_never` is the logical way to indicate this, although it could probably be made more ergonomic.
> 
> If, on the other hand, you think that non-exhaustive `match` statements leave too much room for bugs, then static type checkers could always enforce it and require developers to provide a wildcard or other "catch all" case statement within every `match`.

This was the position of PEP 622: "For cases such as sealed classes and enums, where the patterns are all known to be members of a discrete set, static checkers can warn about missing patterns." But we didn't copy that language into PEP 634.

> Do you have a viewpoint on this? I'd prefer not to leave it up to some optional flag that each type checker needs to expose. If there is adequate consensus that exhaustion is always required for `match`, then all static type checkers should always enforce it.

I think the ship has sailed. (I admit that I wasn't so clear when I linked to PEP 622 above, sorry.)

> My experience with other languages is what led me to the assumption that there are legitimate uses that don't involve exhaustion, but the Python community could decide otherwise. This is a common source of bugs, after all.
> 
> Since you decided not to enforce exhaustion at runtime, my (slight) preference is to follow that precedent and not enforce it in static type checkers by default. We could explore more ergonomic ways to do this than the current `assert_never` or `assert False` techniques.

Agreed.  I think the problem with requiring exhaustion is that while it is easy enough to check in a type checker for unions and enums, it's unreasonable for other types (you'd end up always requiring a wildcard case at the end, which seems wrong, and in conflict with PEP 634). And since at runtime we never check exhaustion, requiring it for certain types in static checkers would be pretty inconsistent.

That said, at least mypy has a bunch of flags that are *very* opinionated (IIRC there's a flag that requires conditions to have type `bool`) so I wouldn't fight a flag that required exhaustion of union and enum matches only, as long as it was configurable per file. That doesn't mean other checkers need to follow though.

JelleZijlstra:
I'd like to see this landed so we can move forward with match support; the above discussion on exhaustiveness can be deferred for later.

@JukkaL above was OK with merging this and deferring any remaining issues, which is also what I'd choose. However, the mypy-primer results suggest that there are some changes in type narrowing behavior from this PR, so I am going to review those to make sure we don't have any regressions:

- https://github.com/Gobot1234/steam.py/blob/main/steam/profile.py#L204 this is a bug. We generate `__match_args__` for dataclasses, and because this class inherits from two dataclasses we generate a bunch of errors about incompatible overrides.
- https://github.com/pytest-dev/pytest/blob/main/src/_pytest/config/__init__.py#L312 is now marked as unreachable, which looks correct (the var is typed `None | PathLike | list`, and there are checks against all of those types).
- https://github.com/schemathesis/schemathesis/blob/master/src/schemathesis/specs/openapi/validation.py#L14 is now also marked as unreachable, which seems right.
- https://github.com/aio-libs/aioredis-py/blob/master/aioredis/connection.py#L1201 apparently the error code changed from "misc" to "literal-required". Not sure how this PR causes that, but seems like a positive change. There are a few other changes that look like they're from the same change.
- https://github.com/rotki/rotki/blob/develop/rotkehlchen/chain/ethereum/modules/adex/adex.py#L811 already has a type ignore on the previous line. Not sure what's going on exactly, but the change doesn't seem obviously wrong.

So only one clear bug. As a quick fix, we can probably just remove the final=True in https://github.com/python/mypy/pull/10191/files#diff-1fdad80f805993800524fbb9b684cc3d09879f54a0e51c71f2a447c7d46c1eb4R212, and iterate later on making inheritance + dataclasses + `__match_args__` safe.

mypy/plugins/dataclasses.py
```diff
@@ -200,6 +201,16 @@ def transform(self) -> None:
...
 
+        if (decorator_arguments['match_args'] and
+                ('__match_args__' not in info.names or
+                 info.names['__match_args__'].plugin_generated) and
+                attributes):
+            str_type = ctx.api.named_type("builtins.str")
+            literals: List[Type] = [LiteralType(attr.name, str_type)
+                        for attr in attributes if attr.is_in_init]
+            match_args_type = TupleType(literals, ctx.api.named_type("builtins.tuple"))
+            add_attribute_to_class(ctx.api, ctx.cls, "__match_args__", match_args_type, final=True)
```
JelleZijlstra, line 212:
```suggestion
            # TODO this should have final=True, but that breaks multiple inheritance
            # https://github.com/python/mypy/pull/10191#issuecomment-1013399068
            add_attribute_to_class(ctx.api, ctx.cls, "__match_args__", match_args_type)
```

sobolevn: We should probably special-case `__match_args__` to be ignored from `is_compatible_with_super` check 🤔

JukkaL:
I think that the remaining dataclass issue can be fixed after merge. I'm okay with merging this as is, if we prioritize fixing the regression after merge. Let's do it?

freundTech:
>     * https://github.com/aio-libs/aioredis-py/blob/master/aioredis/connection.py#L1201 apparently the error code changed from "misc" to "literal-required". Not sure how this PR causes that, but seems like a positive change. There are a few other changes that look like they're from the same change.

This was an intentional change. I introduced the error code "literal-required", because there are now multiple things that require literals. Previously the only place where literals were required were TypedDict and they emitted a misc error code.

Because __match_args__ also require literals I introduced the "literal-required" error code. The other points seem to match up with what I remember from when I last looked at this.

JelleZijlstra:
Thanks! I added https://github.com/python/mypy/labels/topic-match-statement to track future match-related bugs, and filed one for the dataclass override issue discussed above.

sobolevn:
Congrats! Long awaited feature 🎉

henryiii:
That issue seems to be about noticing that something has been exhaustively checked. But the other direction is very important too - it's important to exhaustively check in many/most cases. I think the two possibilities were to add a "strict" mode check to require literals & enums to error if all possibilities are not exhausted, and the other was to use a `case _: assert_never(...)` clause. At runtime, match statements are not checked for exhaustivity, but that would be hard and expensive, possibly impossible due to adding things after-the-fact - while it's easier for MyPy, and kind of the point of MyPy is to be stricter than runtime Python. :) It's also really easy and _much_ more explicit to opt out of strictness checking with a `case _: pass`.

If the explicit opt-in is chosen, though, I think `assert_never` or something like it should be added to `typing` - the reason it has not been added with nicer spelling to typing yet has been because case statements were supposed to be the way to check exhaustively! (Personally, I'd be happy with both a strictness check _and_ a typing "assert_never" addition, so that classic if's can be opt in, and case statements would be opt-out).

gvanrossum:
Why not use `case _: assert False`, which is already in the language? Sure, with `-O` it won't check, but the `assert` indicates that the code is not supposed to get there in the first place, so it's the user's responsibility to feel comfortable that the `case _` is indeed unreachable. And the type checker can help by letting you know if it sees a way that it *could* be reached (e.g. an enum that has no case).

If the user wants the `case _` reachable (say, in a validating parser) they can explicitly raise an exception: `case _: raise ValueError`. (Or a user-defined error or in some cases `TypeError`.)

JelleZijlstra:
Let's move further discussion of exhaustiveness to #12010.

henryiii:
`assert False` doesn't type check. If I'm operating on an enum or a literal, I want the type checker to tell me that I've handled all cases. For example, this is real code from cibuildwheel that currently is type-checked:

```python
PlatformName = Literal["linux", "macos", "windows"]
PLATFORMS: Final[Set[PlatformName]] = {"linux", "macos", "windows"}

...

if platform == "linux":
    cibuildwheel.linux.build(options, tmp_path)
elif platform == "windows":
    cibuildwheel.windows.build(options, tmp_path)
elif platform == "macos":
    cibuildwheel.macos.build(options, tmp_path)
else:
    assert_never(platform)
```

If a new platform is added, all the places this used (like the one above) will be flagged by MyPy as incomplete.

If this was written as a match statement, I'd ideally like this to be enforced by the type checker to not contain fallthrough (via a strictness flag in mypy):

```python
match platform:
    case "linux":
        cibuildwheel.linux.build(options, tmp_path)
    case "windows":
        cibuildwheel.windows.build(options, tmp_path)
    case "macos":
        cibuildwheel.macos.build(options, tmp_path)
```

Completeness checking for pattern matching is not done at runtime, since that would require checking against the Literal, would slow the program, etc, so it was not included. But it seems quite valid for a strict flag in mypy. If I explicitly didn't want to have completeness here, I could add `case _: pass` to make mypy happy with the strictness flag enabled; it would be explicit to the reader that I intended to fall through if unmatched.

But if such a flag was not added, and I have to add `case invalid: assert_never(invalid)` (which would also, for better or worse, catch fallthrough at runtime), then `assert_never` or something like it should be added to `typing`, as I believe part of the original reason for not adding it was because pattern matching was supposed to be the correct way to do completeness checking. (Or at least some way to avoid misusing `NoReturn` when implementing it - also the spelling `assert_never` might not be ideal, maybe `assert_unreachable`?)


Quoting <https://hakibenita.com/python-mypy-exhaustive-checking>:

> In 2018 [Guido though `assert_never` is a pretty clever trick](https://github.com/python/mypy/issues/5818#issuecomment-431863917), but it never made it into mypy. Instead, exhaustiveness checking will become officially available as part of mypy if/when PEP 622 - Structural Pattern Matching is implemented. Until then, you can use `assert_never` instead.

gvanrossum:
> `assert False` doesn't type check.

But we could easily make it so. To people who aren't deep into typing, `assert False` is immediately understandable; `assert_never()` they would have to look up.

> If this was written as a match statement, I'd ideally like this to be enforced by the type checker to not contain fallthrough (via a strictness flag in mypy):

I don't like the idea of such a strictness flag. It doesn't work for large applications (there's always some part of the code that doesn't follow the convention). It's much more scalable to require explicitly marking a case where you want this using `case _: assert False`.

> [...] part of the original reason for not adding it was because pattern matching was supposed to be the correct way to do completeness checking.

By some folks, perhaps. But in the end pattern matching was developed separate from type checking concerns (because it was already controversial enough, and tying it to type checking would probably have sunk it). So let's live with the status quo, which is that you must add an explicit `case _` to capture incompleteness, and `assert False` seems to be the most logical thing to put in code that you expect to be unreachable.

jstasiak:
For what it's worth I believe in most (if not all) codebases I worked with strict exhaustiveness checking (enable with a linter flag, of course, with a `case _: pass` escape hatch) would be very much appreciated and it would solve much more problems than it would create.

henryiii:
Also, I expect there are very few codebases using pattern matching yet, and even if there are, they are not statically typing those yet, since it was just merged. And no one would force them to enable the `--exhaustive-match` or whatever the flag would be. That's the idea of a strictness check, it forces a codebase to be more strict than runtime Python allows. In this case, you would only be able to leave off a `case _:` if all matches were accounted for, and by leaving that off, you are promising that you statically catch all cases. That doesn't mean `assert False` couldn't also be useful (such as for runtime checking), and it would not interfere with that flag at all, but I do think such a flag would make match statements more useful. Just my 2 cents, though.

gvanrossum:
Well, okay, as long as it's a per-module flag.

Also note that pyright (the type checker in VS Code's Python extension) has supported match/case for some time now (IIRC since 3.10 beta). We should ask on typing-sig what their experiences are.

Re: `assert False` -- maybe we could make it a mypy error only if a `case _: assert False` can be reached (I'm sure people use it to mark properties of their code that mypy cannot prove).

henryiii:
Not fond of special handling of `assert False` just inside a catch-all case statement. It's not discoverable - unless I know about this discussion, I wouldn't know it's doing something special for MyPy, while some custom `types.assert_unreachable` would be much easier to read as doing something special here. If it universally was caught, that would not be be an issue, but the special casing would be confusing, I think.

I'd also like to test for an empty union of types elsewhere - like we are already doing with `assert_never` via `NoReturn`. Besides the next 4 years while we are waiting for 3.9's EOL, there might[^1] be cases where you need to narrow down a union to nothing with an if statement and can't use case/match. I think if I saw that `assert False` was handled by MyPy here, again without reading this discussion here, I'd try it there, and it wouldn't work - in the worst possible way: it would not trigger a mypy error/warning.

Finally, `assert False` can be disabled at runtime, and if you can't reach it, there's no point in disabling it, it doesn't speed anything up. But if you do reach it, you don't want it disabled, that's still an error. So it doesn't really seem like the right runtime structure here - I'd use `raise AssertionError` instead, I think, if I were programming for runtime. `assert_never` does work at runtime even with assertions disabled, by the way. It also could directly print out the types that didn't get narrowed out, like reveal_type - but maybe that could be done in the special cased `assert False` too, since it's tied to a match structure.

PS: Though I don't absolutely hate it - catch all case statements are already slightly special in that you can only have one, it has to be at the end, etc.

I think the next step might be to summarize for typing-sig and see if anyone else has opinions, experience, or suggestions? Then the results of that can be added to @JukkaL's exhaustiveness issue?

[^1]: I have no idea if this is true or not - all cases I could think of could be written, probably best, as a match statement.

gvanrossum:
Please write fewer words.

Mypy's original philosophy was to piggyback as much as possible on existing constructs. E.g. `if isinstance(x, <type>): ...`, `assert x is not None`, etc.

Writing `case _: assert False` would fall squarely in that tradition.

henryiii:
I just didn't like the special casing inside the final match-all case statement. But I do like the simplicity of how it reads, as long as people don't expect it to be treated specially by mypy elsewhere if they read it here.

A little [code searching](https://cs.github.com/?scopeName=All+repos&scope=&q=%22assert+False%22+language%3Apython) also shows this is often how `assert False` is already used, at the end of an if chain. :+1:

> Please write fewer words.

Sorry! I know I do tend to write too much.

henryiii:
I've emailed typing-sig with a (hopefully short) summary, and I'll try to add any relevant information from responses to #12010.

tayler6000:
This may be desired functionality, however, returns from within a match statement are not recognized in my project.

```python
from enum import Enum

class CustomEnum(Enum):
  a = 0
  b = 1

class Foo:

  def bar(self, e: CustomEnum, b: bool) -> int:
    match (e, b):
      case (CustomEnum.a, True):
        return 0
      case (CustomEnum.b, True):
        return 1
      case (CustomEnum.a, False):
        return 2
      case (CustomEnum.b, False):
        return 3
```

Returns: `error: Missing return statement`  Which could be valid, because mypy may not realize that all possibilities were covered and there is no default case in this example.

```python
class Foo:

  def bar(self, e: CustomEnum, b: bool) -> int:
    match (e, b):
      case (CustomEnum.a, True):
        return 0
      case (CustomEnum.b, True):
        return 1
      case (CustomEnum.a, False):
        return 2
      case (CustomEnum.b, False):
        return 3
      case _:
        return 4
```

However, this example will always return 4, even if someone improperly uses the function, however mypy returns the same error `error: Missing return statement`.  It is not satisfied until there is a return outside of the match.

```python
class Foo:

  def bar(self, e: CustomEnum, b: bool) -> int:
    match (e, b):
      case (CustomEnum.a, True):
        return 0
      case (CustomEnum.b, True):
        return 1
      case (CustomEnum.a, False):
        return 2
      case (CustomEnum.b, False):
        return 3
      case _:
        return 4
    return 5
```

Apologies if this is intentional.  Additional apologies for my lack of PEP8.

erictraut:
@tayler6000, this is caused by the known issue https://github.com/python/mypy/issues/12010.

Tinche:
@freundTech This PR adds support for dataclasses but not attrs? Any plans on that front?

fredrikaverpil:
I noticed this project does not use GitHub releases. What's a good way to track match statement support in mypy?

richin13:
> I noticed this project does not use GitHub releases. What's a good way to track match statement support in mypy?

I recommend subscribing to this issue: https://github.com/python/mypy/issues/12021 (or to that effect, any future release planning issue where this gets included, if not 0.940)

bergkvist:
### I'm facing issues with match statement support

Consider the following:
```py
from typing import Literal, TypeAlias

Expr: TypeAlias = (
    tuple[Literal['lit'], int] |
    tuple[Literal['var'], str]
)

def square_lits(expr: Expr) -> Expr:
    match expr:
        case ('lit', x):
            return ('lit', square(x))
        case _:
            return expr

def square(x: int) -> int:
    return x * x
```

Here I get an error: 
```
test.py:11: error: Argument 1 to "square" has incompatible type "int | str"; expected "int"  [arg-type]
```

While the following works fine:
```py
from typing import Literal, TypeAlias

Expr: TypeAlias = (
    tuple[Literal['lit'], int] |
    tuple[Literal['var'], str]
)

def square_lits(expr: Expr) -> Expr:
    if expr[0] == 'lit':
        return ('lit', square(expr[1]))
    return expr

def square(x: int) -> int:
    return x * x
```

intgr:
@bergkvist Please report a new issue.

PR #10196, opened 2021-03-10
Fix strconv not displaying False and possibly other falsy values

freundTech:
### Description

This fixes strconv not displaying `False` and possibly other falsy values. I ran into this problem while writing tests for #10191.

## Test Plan

No currently existing tests are affected by this change.


JukkaL (approved):
Thanks!

PR #10203, opened 2021-03-11
Only import distutils if it is needed

freundTech:
### Description

distutils is deprecated in python 3.10, so we only import it if it is
needed, which should only be python < 3.2

## Test Plan

All tests still pass. This is a trivial change and shouldn't cause any problems.


JelleZijlstra:
Mypy doesn't support Python < 3.2, so maybe this branch can just be removed?

JelleZijlstra:
Never mind, this file has to work in Python 2 because it's used for finding site-packages in Python 2.

PR #10211, opened 2021-03-13
Add support for __setitem__

sohailsomani:
### Description

Support for __setitem__

## Test Plan

Test case attached `testAttrOverrides`. I also manually ran mypyc and validated the generated code.


mypyc/codegen/emitwrapper.py
```diff
@@ -269,6 +269,70 @@ def generate_dunder_wrapper(cl: ClassIR, fn: FuncIR, emitter: Emitter) -> str:
...
+    ))
+
+    error_code = 'return -1;'
+    arg_names = [arg.name for arg in fn.args]
+    for arg_name, arg in zip(arg_names, fn.args):
+        # Suppress the argument check for *args/**kwargs, since we know it must be right.
+        typ = arg.type if arg.kind not in (ARG_STAR, ARG_STAR2) else object_rprimitive
+        generate_arg_check(arg_name, typ, emitter, error_code)
+    native_args = ', '.join('arg_{}'.format(arg) for arg in arg_names)
+    assert fn.ret_type.is_unboxed
```
JukkaL, line 291: It's common for the return type to be `Any` (when compiling unannotated code), so boxed return types need to be supported.

mypyc/codegen/emitwrapper.py
```diff
@@ -269,6 +269,70 @@ def generate_dunder_wrapper(cl: ClassIR, fn: FuncIR, emitter: Emitter) -> str:
...
+    emitter.emit_line('{}retval = {}{}({});'.format(emitter.ctype_spaced(fn.ret_type),
+                                                    NATIVE_PREFIX,
+                                                    fn.cname(emitter.names),
+                                                    native_args))
+    emitter.emit_error_check('retval', fn.ret_type, error_code)
+    emitter.emit_line('return 0;')
```
JukkaL, line 297: Decrease the reference count before returning (if needed). Otherwise we may leak the returned value.

mypyc/codegen/emitwrapper.py
```diff
@@ -269,6 +269,70 @@ def generate_dunder_wrapper(cl: ClassIR, fn: FuncIR, emitter: Emitter) -> str:
...
+    ))
+
+    error_code = 'return -1;'
+    arg_names = [arg.name for arg in fn.args]
+    for arg_name, arg in zip(arg_names, fn.args):
+        # Suppress the argument check for *args/**kwargs, since we know it must be right.
+        typ = arg.type if arg.kind not in (ARG_STAR, ARG_STAR2) else object_rprimitive
+        generate_arg_check(arg_name, typ, emitter, error_code)
+    native_args = ', '.join('arg_{}'.format(arg) for arg in arg_names)
+    assert fn.ret_type.is_unboxed
```
JukkaL, line 323: Similar to above, the return value may be boxed. We also need to decref it (if it supports ref counting).

mypyc/test-data/run-classes.test
```diff
@@ -1636,3 +1636,56 @@ from native import A, B, C
 a = A()
 b = B()
 c = C()
+
+[case testAttrOverrides]
+from typing import Any
+
+class Proxy:
+    def __setattr__(self, name:str, value: Any) -> None:
+        print("__setattr__")
```
JukkaL, line 1645: Can you print the arguments here, to make it easier to follow the output? It would be good to do it in the other methods as well.

mypyc/test-data/run-classes.test
```diff
@@ -1636,3 +1636,56 @@ from native import A, B, C
 a = A()
 b = B()
 c = C()
+
+[case testAttrOverrides]
+from typing import Any
+
+class Proxy:
+    def __setattr__(self, name:str, value: Any) -> None:
```
JukkaL, line 1644: Use an unboxed type such as `int` for `value`, so that you'll also test that argument unboxing works correctly.

In addition to `None` return type, test the `Any` return type.

mypyc/test-data/run-classes.test
```diff
@@ -1636,3 +1636,56 @@ from native import A, B, C
...
+        print("__setattr__")
+
+    def __getattr__(self, name:str) -> Any:
```
JukkaL, line 1647: Use `int` for the return value to test return value boxing.

mypyc/test-data/run-classes.test
```diff
@@ -1636,3 +1636,56 @@ from native import A, B, C
...
+        print("__getattr__")
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        print("__setitem__")
```
JukkaL, line 1651: Use `int` for key and value arguments to test argument unboxing.

In addition to the `None` return type, test the `Any` return type.

mypyc/test-data/run-classes.test
```diff
@@ -1636,3 +1636,56 @@ from native import A, B, C
...
+
+    def __getitem__(self, key: Any) -> Any:
```
JukkaL, line 1653: Use `int` as the argument and return type to test unboxing/boxing.

mypyc/test-data/run-classes.test
```diff
@@ -1636,3 +1636,56 @@ from native import A, B, C
...
+        print("__getitem__")
+
+p = Proxy()
+p.a = 5
+a = p.b
+x = p[2]
+p[1] = 5
+
+y = p["object"]
```
JukkaL, line 1662: Use assert to check the return value.

mypyc/test-data/run-classes.test
```diff
@@ -1636,3 +1636,56 @@ from native import A, B, C
...
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        print("__setitem__")
+
+    def __getitem__(self, key: Any) -> Any:
+        print("__getitem__")
+
+p = Proxy()
+p.a = 5
+a = p.b
```
JukkaL, line 1658: Use assert to check the return value.

mypyc/test-data/run-classes.test
```diff
@@ -1636,3 +1636,56 @@ from native import A, B, C
...
+y = p["object"]
+p[p] = "Hello"
+print()
+
+[file driver.py]
+from native import Proxy
+
+p = Proxy()
+p.a = 5
+a = p.b
```
JukkaL, line 1671: Use assert to check the return value.

mypyc/test-data/run-classes.test
```diff
@@ -1636,3 +1636,56 @@ from native import A, B, C
...
+x = p[2]
```
JukkaL, line 1672: Use assert to check the return value.

mypyc/codegen/emitwrapper.py
```diff
@@ -269,6 +269,70 @@ def generate_dunder_wrapper(cl: ClassIR, fn: FuncIR, emitter: Emitter) -> str:
...
+    emitter.emit_line('}')
+
+    return name
+
+
+def generate_mp_setitem_wrapper(cl: ClassIR, fn: FuncIR, emitter: Emitter) -> str:
+    """Generates a wrapper for native __setitem__ methods to be able to fit into
+    the mp_ass_subscript protocol slot. This specifically means that the function
+    raises an exception and returns -1 on failure, 0 on success
+    """
```
JukkaL, line 308: The third argument may `NULL` for `__delitem__`. Check for this and raise an exception. Add a test for this. (No need to actually support `__delitem__` in this PR.)

If you want to support `__delitem__`, you'll need to implement the same dance as for `__delattr__`, since they are both handled in the same wrapper. However, I'd suggest doing this in a separate PR if you want to continue with this.

mypyc/codegen/emitwrapper.py
```diff
@@ -269,6 +269,70 @@ def generate_dunder_wrapper(cl: ClassIR, fn: FuncIR, emitter: Emitter) -> str:
     return name
 
 
+def generate_setattr_wrapper(cl: ClassIR, fn: FuncIR, emitter: Emitter) -> str:
+    """Generates a wrapper for native __setattr__ methods to be able to fit into
+    the tp_setattro protocol slot. This specifically means that the function
+    raises an exception and returns -1 on failure, 0 on success
+    """
+    input_args = ', '.join('PyObject *obj_{}'.format(arg.name) for arg in fn.args)
```
JukkaL, line 277: The third argument may be `NULL` for `__delattr__`. Check for this and raise an exception. Add a test for this. (No need to actually support `__delattr__` in this PR.)

If you want to support `__delattr__`, you can generate a separate wrapper method here from `__delattr__` if the class supports it. Then in the `__setattr__` wrapper you'll actually call the `__delattr__` wrapper if the third argument is `NULL`. Otherwise, you'll proceed as normal.

JukkaL (commented):
Thanks for the PR! Missing support for various dunder methods is perhaps the biggest gap in our Python feature support, so this is very much appreciated.

Overall looks good. Left a bunch of comments mostly about various special cases and test improvements.

JukkaL:
Are you still working on this? I have a branch that overlaps with some of this, and I'll probably a create a PR soon since dunder method support is a high-priority feature. (I will credit you and your PR in my PR.)

sohailsomani:
> Are you still working on this? I have a branch that overlaps with some of this, and I'll probably a create a PR soon since dunder method support is a high-priority feature. (I will credit you and your PR in my PR.)

I have not spent any more time on this, please go ahead, credit or no credit!

PR #10213, opened 2021-03-14
Add optional CI run against python nightly

freundTech:
### Description

This commit adds an optional CI run with python nightly. It is allowed to allow run failures, so even if the nightly run fails the whole CI still succeeds.  

This can be useful for detecting breaking changes in newer python versions and tracking progress towards supporting them.

### Runtime

This adds around 17 minutes to the total time calculated by travis.
Because it can run in parallel with other jobs the actual time will likely be much less, depending on the number of parallel jobs available.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,8 @@ jobs:
     python: 3.8
   - name: "run test suite with python 3.9"
     python: 3.9
+  - name: "run test suite with python nightly"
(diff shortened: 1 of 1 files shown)


JukkaL (approved):
Ok, let's try this out! Previously we've had issues keeping up with the nightlies, and this should help with that.

PR #10217, opened 2021-03-16
[mypyc] Faster creation of a tuple from list/tuple

97littleleaf11:
### Description

mypyc: https://github.com/mypyc/mypyc/issues/769


mypyc/irbuild/specialize.py
```diff
@@ -135,6 +135,9 @@ def translate_safe_generator_call(
                     + [builder.accept(arg) for arg in expr.args[1:]]),
                 builder.node_type(expr), expr.line, expr.arg_kinds, expr.arg_names)
         else:
+            if callee.fullname == "builtins.tuple":
+                length = builder.builder.builtin_len(builder.accept(expr.args[0].sequences[0]), expr.line)
```
97littleleaf11, line 139: DRAFT here. 
I am confused about:
-  how can I get the Value of the `list` from `GeneratorExpr` since `builtin_len` needs a `Value`
-  Where should I move these code to?

97littleleaf11:
I went through `translate_list_comprehension` and `comprehension_helper`. I would write similar code to fill in the blank tuple after I get `tuple_op`, is it correct?

mypyc/irbuild/specialize.py
JukkaL: You'll need to check that the sequence is something that supports `len()` and indexing. Initially it's okay to require is to be a list. Otherwise you can skip the optimization. It's possible to extend this to tuples at least. You can use `node_type(...)` and `is_list_rprimitive` to check if it's a list.

So this would be optimized:

```py
a = [1, 2, 3]
tuple(f(x) for x in a)  # Works because a is a list
```

This wouldn't work:

```py
a = iter([1, 2, 3])
# Can't optimize -- a is an iterable, and we don't know the length ahead of time
tuple(f(x) for x in a)
```

`builder.accept` is correct way to translate an `Expression` into a `Value`.

Since the `tuple(...)` specialization is going to be somewhat involved, you can create a helper function that you call from here. The helper function can check whether the optimization can be performed and return `None` if we should fall back to the default behavior. You can add the helper function in this same file.

JukkaL:
> I went through translate_list_comprehension and comprehension_helper. I should write similar code to fill in the blank tuple after I get tuple_op, is it correct?

That's right! It's also possible that you can reuse `ForSequence` (or `ForEnumerate` / `ForRange`) to simplify the generation of the for loop (not sure though whether it's worth it).

mypyc/irbuild/specialize.py
97littleleaf11: Thanks! 
Another question: how can I convert `short_int_rprimitive`(got from `builtin_len`) to `c_pyssize_t_rprimitive`?

JukkaL: They are the same size, so you only need to shift right by 1 (`IntOp` with `RIGHT_SHIFT`). You can find examples of `LEFT_SHIFT` -- `RIGHT_SHIFT` is pretty much the same.

mypyc/irbuild/ll_builder.py
```diff
@@ -1079,12 +1081,15 @@ def int_op(self, type: RType, lhs: Value, rhs: Value, op: int, line: int) -> Val
...
-    def builtin_len(self, val: Value, line: int) -> Value:
+    def builtin_len(self, val: Value, line: int,
+                    get_c_pyssize_t_rprimitive: bool = False) -> Value:
         typ = val.type
         if is_list_rprimitive(typ) or is_tuple_rprimitive(typ):
             elem_address = self.add(GetElementPtr(val, PyVarObject, 'ob_size'))
             size_value = self.add(LoadMem(c_pyssize_t_rprimitive, elem_address))
             self.add(KeepAlive([val]))
+            if get_c_pyssize_t_rprimitive:
+                return size_value
```
97littleleaf11, line 1092: @JukkaL Is it OK that I directly get `c_pyssize_t_rprimitive` from here?

JukkaL: This is fine, as long as it's supported for all types below. Otherwise it would be error-prone.

mypyc/irbuild/specialize.py
```diff
@@ -135,13 +140,48 @@ def translate_safe_generator_call(
...
+
+        def set_tuple_item() -> None:
+            e = builder.accept(gen.left_expr)
+            index_val = builder.accept(index)
+            offset = Integer(1, c_pyssize_t_rprimitive, gen.line)
+            index_pyssize_t = builder.int_op(c_pyssize_t_rprimitive, index_val, offset,
+                                             IntOp.RIGHT_SHIFT, gen.line)
+            index_pyssize_t = builder.int_op(c_pyssize_t_rprimitive, index_pyssize_t, offset,
+                                             IntOp.SUB, gen.line)
+            builder.call_c(new_tuple_set_item_op, [tuple_ops, index_pyssize_t, e], gen.line)
```
97littleleaf11, line 174: @JukkaL , I convert `int_rprimitive` to `c_pyssize_t_rprimitive` here. Another implementation would be writing a helper C function in `tuple_ops.c`. Which one is better?

mypyc/primitives/tuple_ops.py
```diff
@@ -27,6 +28,21 @@
...
+
+# PyTuple_SET_ITEM does no error checking,
+# and should only be used to fill in brand new tuples.
+new_tuple_set_item_op = custom_op(
+    arg_types=[tuple_rprimitive, c_pyssize_t_rprimitive, object_rprimitive],
+    return_type=void_rtype,
+    c_function_name='PyTuple_SET_ITEM',
+    error_kind=ERR_NEVER
+)
+
```
97littleleaf11, line 45: Using `PyTuple_SET_ITEM ` instead of `PyTuple_SetItem` since we only use it after initialization.

97littleleaf11:
It should work now, I will add some IR test tomorrow.

mypyc/irbuild/specialize.py
97littleleaf11: minus 1 for correct index

mypyc/irbuild/specialize.py
```diff
@@ -135,13 +140,48 @@ def translate_safe_generator_call(
...
                     + [builder.accept(arg) for arg in expr.args[1:]]))
     return None
 
 
+def tuple_from_generator_helper(builder: IRBuilder,
+                                gen: GeneratorExpr) -> Optional[Value]:
+
+    if len(gen.sequences) == 1 and len(gen.condlists[0]) == 0:
+        # Currently we only optimize for simplest generator expression
+        if not is_list_rprimitive(builder.node_type(gen.sequences[0])):
```
97littleleaf11, line 159: Here I limit the Generator with only one for-in loop and no conditions. Is it fine?

mypyc/irbuild/ll_builder.py
97littleleaf11: Got it! I add support for `dict` and `set` below

mypyc/irbuild/specialize.py
97littleleaf11: `IntOp.RIGHT_SHIFT ` results in a test failure

97littleleaf11: `index_val` on line 168 seems not to be recognized as `int_rprimitive`, how should I fix it?

97littleleaf11:
There is another implementation on a new personal branch: https://github.com/97littleleaf11/mypy/commit/4f0b92acb0401dc7ebc0051ef00ce9af3560c4fe

mypyc/irbuild/ll_builder.py
JukkaL: The generic case also needs to return a `c_pyssize_t_rprimitive`. For this you can call `PyObject_Size` directly instead of `CPyObject_Size` (via `generic_len_op`). Add a new primitive to `mypyc.primitives.generic_ops`, e.g. `generic_ssize_t_len_op` or similar.

mypyc/primitives/tuple_ops.py
JukkaL: This makes sense. 

Since `PyTuple_SET_ITEM` is a macro, non-C backend won't be able to use it, however (but currently we only have a C back end). It's okay to use it for now, but we'll need a follow-up issue to eventually replace this with lower-level IR operations such as `SetMem`. If you want to see how this is done, `new_list_op` does a similar thing as `PyTuple_SET_ITEM` but for lists, using low-level IR operations only.

mypyc/irbuild/specialize.py
```diff
@@ -135,13 +141,47 @@ def translate_safe_generator_call(
...
 
 
+def tuple_from_generator_helper(builder: IRBuilder,
+                                gen: GeneratorExpr) -> Optional[Value]:
+
+    if len(gen.sequences) == 1 and len(gen.condlists[0]) == 0:
+        # Currently we only optimize for simplest generator expression
+        rtype = builder.node_type(gen.sequences[0])
+        if ((is_list_rprimitive(rtype) or is_tuple_rprimitive(rtype))
+                and is_int_rprimitive(builder.node_type(gen.indices[0]))):
```
JukkaL, line 162: There is no need to restrict the index types to integers. I think that this check can be removed.

mypyc/irbuild/specialize.py
```diff
@@ -135,13 +141,47 @@ def translate_safe_generator_call(
...
+            tuple_ops = builder.builder.new_tuple_with_length(builder.accept(gen.sequences[0]),
+                                                              gen.line)
+            index, expr = gen.indices[0], gen.sequences[0]
+
+            def set_tuple_item() -> None:
+                e = builder.accept(gen.left_expr)
+                index_val = builder.accept(index)
```
JukkaL, line 169: I don't think that this works as expected. The naming is pretty confusing, but I think that this is actually a sequence item (`x` in `for x in my_list`), not the index. Actually `for_loop_helper` doesn't seem to give you access to the sequence index. That probably means that `for_loop_helper` isn't a good match for this use case, unfortunately. 

Directly using `ForEnumerate` or maybe `ForRange` could be better options. If using `ForEnumerate`, you'll have to create a new lvalue for `index1` (a register). If using `ForRange`, you'll need to use something similar to what `transform_assignment_stmt` does to assign the sequence item to the lvalue(s). The key part is this:

```
    target = builder.get_assignment_target(lvalue)
    builder.assign(target, rvalue_reg, line)
```

97littleleaf11: I have checked the generated IR, it seems that it dose work.

mypyc/irbuild/ll_builder.py
97littleleaf11: Thanks for pointing out! I will fix it later.

mypyc/irbuild/specialize.py
97littleleaf11: Otherwise the case `testComprehensionShadowBinder` would fail.

mypyc/irbuild/specialize.py
JukkaL: Yeah, that's fine.

mypyc/irbuild/specialize.py
JukkaL: I suspect that this means that there is a problem elsewhere. The optimization should work for cases like this:

```py
def f(a: List[str]) -> Tuple[str, ...]:
    return tuple(x + 'a' for x in a)
```

mypyc/irbuild/specialize.py
JukkaL: As discussed above, can you check that this works for types such as `List[str]`?

mypyc/primitives/generic_ops.py
```diff
@@ -241,6 +241,14 @@
     c_function_name='CPyObject_Size',
     error_kind=ERR_NEVER)
 
+# len(obj)
+# same as generic_len_op, however return py_ssize_t
+generic_ssize_t_len_op = custom_op(
+    arg_types=[object_rprimitive],
+    return_type=c_pyssize_t_rprimitive,
+    c_function_name='PyObject_Size',
+    error_kind=ERR_NEVER)
```
JukkaL, line 250: This should be `ERR_NEG_INT`, since this returns -1 on error. Actually `CPyObject_Size` should also be `ERR_MAGIC`, looks like.

mypyc/irbuild/specialize.py
97littleleaf11: Well, I didn't consider that. > <
I would add a new helper function similar to `for_loop_helper`.

JukkaL: Yeah, that should work. One option would be to have a similar helper that gives you access both to the index (starting from 0) and the item value. You don't need to support everything that `for_loop_helper` supports since this helper would be used in more specialized cases.

mypyc/primitives/generic_ops.py
97littleleaf11: Fixed!

mypyc/irbuild/for_helpers.py
```diff
@@ -86,6 +86,38 @@ def for_loop_helper(builder: IRBuilder, index: Lvalue, expr: Expression,
...
+    for_gen.gen_step()
+    builder.goto(condition_block)
+
+    for_gen.add_cleanup(exit_block)
+    builder.pop_loop_stack()
+
+    builder.activate_block(exit_block)
+
+
+
```
97littleleaf11, line 120: DRAFT version.
I found that I can directly get `index_target` from `ForSequence`
It seems work well now, but produces code duplication here. Also, the function name `for_loop_helper_simple ` should be changed.

mypyc/test-data/irbuild-tuple.test
```diff
@@ -258,3 +258,222 @@ L14:
...
+    r8 :: int32
+    r9, r10 :: bit
+    r11, a :: tuple
+L0:
+    r0 = (1, 1, 0, 1)
+    source = r0
+    r1 = PyList_New(0)
+    r2 = box(tuple[bool, bool, bool, bool], source)
+    r3 = PyObject_GetIter(r2)
+L1:
```
97littleleaf11, line 462: Bugs here.
I found that tuple `source` would be recognized as `tuple` instead of `builtin_tuple`, thus failed in `is_tuple_rprimitive (rtype)` in `specialize.py`

97littleleaf11:
IR test failed on 32bit

TH3CHARLie:
`c_pyssize_t` has platform-specific size so we use `native_int` in IR output and it will be replaced with either `int32` or `int64` in test textual matching. So a quick fix is to replace every `int64` in your code with `native_int`. Later you can try to improve https://github.com/mypyc/mypyc/issues/776 to solve this some-what ugly workflow

mypyc/test-data/irbuild-tuple.test
```diff
@@ -258,3 +258,222 @@ L14:
...
+def f(val: int) -> bool:
+    return val % 2 == 0
+
+def test() -> None:
+    source = [1, 2, 3]
+    a = tuple(f(x) for x in source)
+[out]
+def f(val):
+    val, r0 :: int
+    r1 :: int64
```
JukkaL, line 273: Change this to `native_int`. It may fix the 32-bit tests.

mypyc/test-data/irbuild-tuple.test
```diff
@@ -258,3 +258,222 @@ L14:
...
+    r4 = r5
+L3:
+    return r4
+def test():
+    r0 :: list
+    r1, r2, r3 :: object
+    r4, r5, r6, r7 :: ptr
+    source :: list
+    r8 :: ptr
+    r9 :: int64
```
JukkaL, line 297: Similar to above, try changing this to `native_int`.

mypyc/test-data/irbuild-tuple.test
```diff
@@ -258,3 +258,222 @@ L14:
...
+    r10 :: tuple
+    r11 :: short_int
+    r12 :: ptr
+    r13 :: int64
```
JukkaL, line 301: Like above.

mypyc/test-data/irbuild-tuple.test
```diff
@@ -258,3 +258,222 @@ L14:
...
+L3:
+    r21 = r11 + 2
+    r11 = r21
+    goto L1
+L4:
+    a = r10
+    return 1
+
+
+[case testTupleBuiltFromList]
```
JukkaL, line 353: The name is the same as above. Make it unique (e.g. `...FromList2`).

mypyc/test-data/irbuild-tuple.test
```diff
@@ -258,3 +258,222 @@ L14:
...
+    r24 = PyObject_CallFunctionObjArgs(r23, a, 0)
+    return 1
+
+
+[case testTupleBuiltFromTuple]
+def f(val: bool) -> bool:
+    return not val
+
+def test() -> None:
+    source = (True, True, False, True)
```
JukkaL, line 439: Maybe annotate this as a variable-length tuple: `source: Tuple[bool, ...] = (True, True, False, True)`.

mypyc/test-data/irbuild-tuple.test
JukkaL: The above type annotation should help, but it would be better if the specialization also can be made to work with fixed-length tuples.

You could add a check for `RTuple` (fixed-length tuple) and also handle that, instead of just supporting variable-length tuples and lists. You may need to coerce it to a variable-length tuple explicitly first, using `coerce`.

mypyc/irbuild/for_helpers.py
```diff
@@ -86,6 +86,37 @@ def for_loop_helper(builder: IRBuilder, index: Lvalue, expr: Expression,
     builder.activate_block(exit_block)
 
 
+def for_loop_helper_simple(builder: IRBuilder, index: Lvalue, expr: Expression,
+                           body_insts: Callable[[Value], None], line: int) -> None:
```
JukkaL, line 90: Add docstring here. Explain all the arguments and how this is different from `for_loop_helper`.

JukkaL (commented):
Left some comments (not a full review). 

Also, can you evaluate the performance impact using a microbenchmark? For example, test something like this (call `f` many times in a loop):

```py
def f(a: List[int]) -> Tuple[int, ...]):
    return tuple(x + 1 for x in a)
```

You can also check if this speeds up the `nqueens` benchmark in https://github.com/mypyc/mypyc-benchmarks. It's okay if there's no difference here, but it would be great if there is.

The above repository has a benchmark runner that can make it easier to measure performance. Just add a microbenchmark and run it using the `runbench.py` tool (no need to create a PR, you can experiment locally).

TH3CHARLie:
I kind of feel this issue is less good-first-issue. But that's good, you get to know more aspects about the codebase and workflow :D

97littleleaf11:
```py
@benchmark
def tuple_build_test() -> None:
    a = []
    for i in range(1000):
        a.append(i)
    for i in range(1000):
        b = tuple(x + 1 for x in a)
```

new:

interpreted: 0.059748s (avg of 16 iterations; stdev 1.2%)
compiled:    0.017971s (avg of 16 iterations; stdev 3.6%)

compiled is 3.325x faster


master: 

interpreted: 0.058382s (avg of 17 iterations; stdev 0.96%)
compiled:    0.025852s (avg of 17 iterations; stdev 2.5%)

compiled is 2.258x faster

TH3CHARLie:
you can use the resolve conversation button to make the PR page more concise

mypyc/test-data/irbuild-tuple.test
97littleleaf11: Should I support `RTuple` in this PR?

mypyc/test-data/irbuild-tuple.test
```diff
@@ -258,3 +258,233 @@ L14:
     r11 = r15
 L15:
     return r11
+
+
+[case testTupleBuiltFromList]
```
JukkaL, line 263: Also add a test where we execute the code, in `mypyc/test-data/run-tuples.test`. You can add more sub-cases to the existing `testTupleOps` test. Note that it uses the default driver, so you can just add `test_<whatever>` function(s). Test iterating over a list, a tuple and an unsupported iterable (e.g. set).

JukkaL (commented):
Thanks for the updates! The performance improvement is great, and we could get even more performance by inlining some more operations (no need to do it now, this is already god). This is almost ready.

Left a few more comments, mostly minor.

mypyc/test-data/irbuild-tuple.test
97littleleaf11: Thanks @JukkaL ! I just add a test function including iterating over a `list`, a `tuple` and a `RTuple`(unsupported).

mypyc/irbuild/for_helpers.py
```diff
@@ -86,6 +86,45 @@ def for_loop_helper(builder: IRBuilder, index: Lvalue, expr: Expression,
...
+    Args:
+        index: the loop index Lvalue
+        expr: the expression to iterate over
+        body_insts: a function that generates the body of the loop.
+                    It needs a index as parameter.
+    """
+    body_block, step_block, exit_block = BasicBlock(), BasicBlock(), BasicBlock()
+    condition_block = BasicBlock()
+
+    expr_reg = builder.accept(expr)
```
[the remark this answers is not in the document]
TH3CHARLie: This assertion hasn't been added yet

TH3CHARLie (approved):
With all minor comments addressed, I think it's LGTM now! Thanks!

JukkaL (approved):
Thanks for the final updates!

Issue #9950, opened 2021-01-23
Drop support for running on Python 3.5


JukkaL:
Python 3.5 has reached its end of life. However, Ubuntu 16.04 LTS still ships with
Python 3.5 and reaches its end of life in April 2021, and I'd like to keep supporting 3.5
until April (unless it becomes too difficult).

My motivation for waiting this long is that it's fairly common for organizations to leave 
upgrades until the last minute.

JelleZijlstra:
@JukkaL is there anything still blocking this? It would be good to do this sooner rather than later so contributors can stop worrying about Python 3.5 support (e.g. in #10219).

JukkaL:
I'd like to make at least one more public release with Python 3.5 support. We should be able to drop it (soon) afterwards. However, in the next release we can mark Python 3.5 as deprecated and any new features will no longer need to support 3.5.

hauntsaninja:
mypy no longer supports 3.5, as of #10706

PR #10219, opened 2021-03-17
Add support for pulling configuration from pyproject.toml files

### LINKED ISSUE #5205: Support for pyproject.toml

TheCleric:
### Description

Closes #5205 

This PR will add support to mypy for end users specifying configuration in a pyproject.toml file. . I also updated the documentation to indicate that this support has been added, along with some guidelines on adapting the ini configurations to the toml format.

## Test Plan

I've added tests as best I could to mirror existing tests that use a mypy.ini file, but with a pyrproject.toml file, to ensure the configuration is being pulled correctly with the desired results.


TheCleric:
Looking into the test failures. I didn't see these on my machine because I was using a `venv` and I think they got skipped.

TheCleric:
Might need some help from someone with more Travis CI experience than I. It seems to be throwing strange python segmentation faults, which I am not sure where they are coming from.

JukkaL:
It looks like you may have encountered a mypyc bug (mypyc is a compiler we use to compile mypy to C extensions). I'll try to debug this, since debugging mypyc issues can be non-trivial. Sorry about that!

TheCleric:
> It looks like you may have encountered a mypyc bug (mypyc is a compiler we use to compile mypy to C extensions). I'll try to debug this, since debugging mypyc issues can be non-trivial. Sorry about that!

Thanks so much!

mypy/config_parser.py
```diff
@@ -100,7 +122,22 @@ def check_follow_imports(choice: str) -> str:
     'cache_dir': expand_path,
     'python_executable': expand_path,
     'strict': bool,
-}  # type: Final
+}  # type: Final[Dict[str, Any]]
+
+# Reuse the ini_config_types and overwrite the diff
+toml_config_types = copy.deepcopy(ini_config_types)  # type: Final[Dict[str, Any]]
```
JukkaL, line 128: This seems to be the issue. Copying a compiled lambda doesn't work because of a mypyc bug: https://github.com/mypyc/mypyc/issues/804#issuecomment-802947572

Can you try replacing this with `copy.copy`? It should work.

JukkaL:
After fixing the above issue, compiled tests still fail for me. Here's what I see:
```
/Users/jukka/src/mypy/test-data/unit/check-custom-pyproject-plugin.test:557:
/Users/jukka/src/mypy/mypy/test/testcheck.py:138: in run_case
    self.run_case_once(testcase, ops, step)
/Users/jukka/src/mypy/mypy/test/testcheck.py:177: in run_case_once
    options = parse_options(original_program_text, testcase, incremental_step)
/Users/jukka/src/mypy/mypy/test/helpers.py:386: in parse_options
    targets, options = process_options(flag_list, require_targets=False)
mypy/main.py:855: in process_options
    ???
mypy/config_parser.py:200: in parse_config_file
    ???
mypy/config_parser.py:306: in parse_section
    ???
E   TypeError: str or None object expected; got float
```

This actually looks like it could be a legitimate error that mypyc was able to catch through strict runtime type checking. The type for argument `section: Mapping[str, str]` seems incorrect when using toml, as the value could be `float`, not just `str`.

TheCleric:
> This actually looks like it could be a legitimate error that mypyc was able to catch through strict runtime type checking. The type for argument `section: Mapping[str, str]` seems incorrect when using toml, as the value could be `float`, not just `str`.

Thanks, I'll fix!

TheCleric:
@JukkaL any clue on the strange failures on 3.5.1 with mypyc? These seem to pass everywhere else (including locally on 3.5.1), but for some reason the output prints out of order on these.

TheCleric:
I think I'm narrowing down the issues with 3.5. TOML doesn't use an OrderedDict by default like config_parser does, so I'm working on overriding that. In Python 3.6+ dicts are ordered by default which is why this is the only version with an issue. Will have a fix soon.

JelleZijlstra:
We might drop support for 3.5 soon (#9950), so maybe it's not worth worrying about? I'd also be OK with not supporting pyproject.toml on 3.5 while we wait to confirm that we can fully drop support for 3.5.

TheCleric:
> We might drop support for 3.5 soon (#9950), so maybe it's not worth worrying about? I'd also be OK with not supporting pyproject.toml on 3.5 while we wait to confirm that we can fully drop support for 3.5.

I'd be fine either way. I'll defer to those who may know more than I. 😄

JukkaL:
I'm fine with not supporting pyproject.toml on Python 3.5. In the next public release we can declare Python 3.5 as deprecated (even though Python 3.5 will likely still be mostly supported). Just give a clear error messages if somebody tries to use it on Python 3.5.

TheCleric:
> I'm fine with not supporting pyproject.toml on Python 3.5. In the next public release we can declare Python 3.5 as deprecated (even though Python 3.5 will likely still be mostly supported). Just give a clear error messages if somebody tries to use it on Python 3.5.

It actually works fine in Python 3.5, it's just one of the tests is failing because the order we store the data in the dict is variable since they aren't ordered. But the actual functionality is fine.

TheCleric:
Good news, 3.5 works now. Bad news, not sure what this mypy_primer error is all about (doesn't look like it is code related). @JukkaL any thoughts?

JelleZijlstra:
cc @hauntsaninja too, the error isn't very informative.

hauntsaninja:
Error was spurious, possibly related to https://www.githubstatus.com/incidents/fwtx9db6hz2y

TheCleric:
Alright. Seems good to go from my perspective. Ready for review!

mypy/config_parser.py
```diff
@@ -121,14 +160,29 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
...
-        except configparser.Error as err:
+            if config_file.endswith('.toml'):
+                toml_data = toml.load(config_file, _dict=OrderedDict)
+                # Filter down to just mypy relevant toml keys
+                toml_data = toml_data.get('tool', {})
+                other_keys = [key for key in toml_data.keys()
+                              if key.split('-')[0] != 'mypy']
+                for key in other_keys:
+                    del toml_data[key]
+                if toml_data.get('mypy') is None:
```
JukkaL, line 177: `if 'mypy' not in toml_data`?

mypy/config_parser.py
```diff
@@ -281,6 +340,18 @@ def parse_section(prefix: str, template: Options,
     return results, report_dirs
 
 
+def convert_to_boolean(value: Optional[Any]) -> bool:
+    """Return a boolean value translating from other types if necessary.
```
JukkaL, line 344: Style nit: Have both `"""` sequences on the same line.

test-data/unit/check-modules.pyproject.test
```diff
@@ -0,0 +1,92 @@
...
+
+[file a/__init__.pyi]
+from typing import Any
+def __getattr__(attr: str) -> Any: ...
+[file a/b/__init__.pyi]
+# empty (i.e. complete subpackage)
+
+[file pyproject.toml]
+\[tool.mypy]
+\[tool.'mypy-a.b.c']
```
JukkaL, line 17: This looks a bit clunky to me. Maybe we can come up with a better way to represent this?

What about using the approach that was taken in the first attempt to implement pyproject.toml support (#5208)? It was like this:

```toml
[tool.mypy.overrides.'a.b.c']
ignore_missing_imports = true
```

This looks a bit cleaner to me, but still not great. 

I like this one more, as it allows splitting long lists of globs:

```toml
[[tool.mypy.overrides]]
module = "a.b.c"
ignore_missing_imports = true

[[tool.mypy.overrides]]
module = [
   "b.*",
   "c",
]
ignore_missing_imports = true
```

What do you think?

test-data/unit/check-custom-pyproject-plugin.test
```diff
@@ -0,0 +1,737 @@
+-- Test cases for user-defined plugins
```
JukkaL, line 1: This and the following test files are pretty verbose. I think that it's better to try to keep the test files shorter so that each test covers some additional toml/config feature. A large number of tests slows local test runs and adds maintenance overhead.

It's good that you added so many tests, as it adds confidence to the implementation. Once this PR is close to ready, I'd recommend removing tests that are unlikely to increase test coverage significantly.

JukkaL (commented):
Thanks, looks mostly good! This has been popular feature request. It will be nice to finally have it.

Did a quick review pass. The most important things to resolve is the syntax for defining per-module configuration options and trimming down the test cases. 

Once those are done, I will do another review pass and test this manually a bit.

test-data/unit/check-custom-pyproject-plugin.test
TheCleric: > It's good that you added so many tests, as it adds confidence to the implementation. Once this PR is close to ready, I'd recommend removing tests that are unlikely to increase test coverage significantly.

I agree. I did these mostly to ensure that I covered as many use cases as I could, but realize they are a bit out of control. I'm definitely open to paring them down and/or combining a lot of the positive unit tests.

test-data/unit/check-modules.pyproject.test
TheCleric: My intention here was to keep it as close to the ini format as I could for ease of migration. I can certainly be swayed away from that. Either of the options shown look better. While I agree the second one looks cleaner, the downside to me is that now you're going to have to start checking things normally handled at the TOML level. For example, this would get flagged before it even gets parsed by our code in mypy:

```toml
[tool.mypy.overrides.'a.b.c']
ignore_missing_imports = true

[tool.mypy.overrides.'a.b.c']
ignore_missing_imports = false
```

However, under the second example, this is possible and would need code to account for it:

```toml
[[tool.mypy.overrides]]
module = "a.b.c"
ignore_missing_imports = true

[[tool.mypy.overrides]]
module = "a.b.c"
ignore_missing_imports = false
```

So now we'd have to start accounting for what happens when later overrides accidentally modify a prior override value. Do we keep oldest? Newest? Raise an error?

So out of laziness I would probably prefer the first as I'd like to lean on the TOML parser to catch situations like this.

It would take some work, but we could also support something like this, where the quotes are only needed for special characters:

```toml
[tool.mypy.overrides.a.b.c]
ignore_missing_imports = true

[tool.mypy.overrides.a.b.c.'*']
ignore_missing_imports = false
```

Let me know what you think @JukkaL

JukkaL: At work we have sections with many globs, which I find pretty ugly. Here's an example with dummy names:

```
[mypy-pkg1.mod1,pkg2.subpkg.*,...<dozens of similar items>...,pkg30.something.else.*]
<various strictness options>
```

This makes me prefer the `[[tool.mypy.overrides]]` alternative. With this syntax, pyproject.toml would look more appealing than mypy.ini, in my opinion, and we could possibly start recommending it over mypy.ini in the future.

Note that the duplicate module name issue is possibly already a problem if we have multiple patterns per section:

```
[tool.mypy.overrides.'a.b,a.b.c']
...

[tool.mypy.overrides.'a.b.c']
...
```

TheCleric: Okay, given that, what do we just say about duplicate behavior? Last one wins?

I'll start working on your second suggestion while we figure out this edge case.

JukkaL: If there are two identical globs/module names, I think it's best to give an error. Otherwise it could be confusing, if the user doesn't realize that there are two sections that match.

TheCleric: @JukkaL this should now be working the way you asked.

ErikBjare:
Looks like all the comments have been resolved. What's left?

mypy/config_parser.py
```diff
@@ -184,9 +242,84 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
...
+                disallow_untyped_defs: false,
+            },
+        }
+    """
+    if 'overrides' not in toml_data['mypy']:
+        return toml_data
+
+    if not isinstance(toml_data['mypy']['overrides'], list):
+        raise ValueError("tool.mypy.overrides sections must be an array. Please make sure you are "
+                         "using double brackets like so: [[tool.mypy.overrides]]")
```
JukkaL, line 282: Raise a more specific error (e.g. define a new exception class), and make sure the errors are reported cleanly and don't generate a traceback. It would be good to have tests for the error cases.

mypy/config_parser.py
```diff
@@ -184,9 +242,84 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
...
+
+    result = copy.copy(toml_data)
+    for override in result['mypy']['overrides']:
+        if 'module' not in override:
+            raise ValueError("toml config file contains a [[tool.mypy.overrides]] section, but no "
+                             "module to override was specified.")
```
JukkaL, line 288: Similar to above.

mypy/config_parser.py
```diff
@@ -184,9 +242,84 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
...
+
+        if isinstance(override['module'], str):
+            modules = [override['module']]
+        elif isinstance(override['module'], list):
+            modules = override['module']
+        else:
+            raise ValueError("toml config file contains a [[tool.mypy.overrides]] section with "
+                             "a module value that is not a string or a list of strings")
```
JukkaL, line 296: Similar to above.

mypy/config_parser.py
```diff
@@ -184,9 +242,84 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
...
+            if old_config_name not in result:
+                result[old_config_name] = module_overrides
+            else:
+                for new_key, new_value in module_overrides.items():
+                    if (new_key in result[old_config_name] and
+                            result[old_config_name][new_key] != new_value):
+                        raise ValueError("toml config file contains [[tool.mypy.overrides]] "
+                                         "sections with conflicting values. Module '%s' "
+                                         "has two different values for '%s'"
+                                         % (module, new_key))
```
JukkaL, line 311: Similar to above.

mypy/config_parser.py
```diff
@@ -184,9 +242,84 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
...
+                    result[old_config_name][new_key] = new_value
+
+    del result['mypy']['overrides']
+    # print(result)
```
JukkaL, line 315: Remove debug print.

mypy/config_parser.py
```diff
@@ -281,6 +417,17 @@ def parse_section(prefix: str, template: Options,
...
 
 
+def convert_to_boolean(value: Optional[Any]) -> bool:
+    """Return a boolean value translating from other types if necessary."""
+    if isinstance(value, bool):
+        return value
+    if not isinstance(value, str):
+        value = str(value)
+    if value.lower() not in configparser.RawConfigParser.BOOLEAN_STATES:
+        raise ValueError('Not a boolean: %s' % value)
```
JukkaL, line 427: Similar to above, raise a custom error and make sure we handle it without generating a traceback.

test-data/unit/cmdline.pyproject.test
```diff
@@ -0,0 +1,793 @@
...
+def g(a: int) -> int: return f(a)
+[file yx.py]
+def f(a): pass
+def g(a: int) -> int: return f(a)
+[file yy.py]
+def f(a): pass
+def g(a: int) -> int: return f(a)
+[out]
+pyproject.toml: [mypy-*x*]: Patterns must be fully-qualified module names, optionally with '*' in some components (e.g spam.*.eggs.*)
+pyproject.toml: [mypy-*y*]: Patterns must be fully-qualified module names, optionally with '*' in some components (e.g spam.*.eggs.*)
```
JukkaL, line 107: This error message is a bit confusing (though not a big deal).

docs/source/config_file.rst
```diff
@@ -885,5 +885,67 @@ These options may only be set in the global section (``[mypy]``).
...
+
+  * I.e., ``[mypy]`` would become ``[tool.mypy]``
+
+* The module specific sections should be moved into ``[[tool.mypy.overrides]]`` sections:
+
+  * For example, ``[mypy-packagename]`` would become:
+
+  ```
+  [[tool.mypy.overrides]]
+  module = 'packagename
```
JukkaL, line 905: Document `module = [...]` as well.

mypy/config_parser.py
```diff
@@ -121,14 +163,29 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
...
             continue
         try:
-            parser.read(config_file)
-        except configparser.Error as err:
+            if config_file.endswith('.toml'):
+                toml_data = toml.load(config_file, _dict=OrderedDict)
+                # Filter down to just mypy relevant toml keys
+                toml_data = toml_data.get('tool', {})
+                other_keys = [key for key in toml_data.keys()
+                              if key.split('-')[0] != 'mypy']
```
JukkaL, line 177: Do we still need the `split('-')`?

mypy/config_parser.py
```diff
@@ -121,14 +163,29 @@ def parse_config_file(options: Options, set_strict_flags: Callable[[], None],
...
+                    del toml_data[key]
+                if 'mypy' not in toml_data:
+                    continue
+                parser = destructure_overrides(toml_data)
+                config_types = toml_config_types
+            else:
+                config_parser.read(config_file)
+                parser = config_parser
+                config_types = ini_config_types
+        except (toml.TomlDecodeError, configparser.Error, ValueError) as err:
```
JukkaL, line 188: As mentioned elsewhere, it seems better to use a dedicated exception instead of `ValueError`, since it's possible to get a `ValueError` from a bug in the implementation, and in that case I'd rather get a traceback.

JukkaL (commented):
Thanks for the updates! The new config file syntax looks great to me.

Left some more minor comments. 

After you've done everything else, it would be great it you can trim the test cases.

mypy/config_parser.py
TheCleric: Nope. Will remove.

mypy/config_parser.py
TheCleric: Whoops!

mypy/config_parser.py
TheCleric: In this case, this is the same way that the `RawConfigParser` does it, so I'd recommend staying consistent with how it parses booleans vs. how we do so for the TOML file.

I.e., here's what's in `configparser.py`:

```python
    def _convert_to_boolean(self, value):
        """Return a boolean value translating from other types if necessary.
        """
        if value.lower() not in self.BOOLEAN_STATES:
            raise ValueError('Not a boolean: %s' % value)
        return self.BOOLEAN_STATES[value.lower()]
```

test-data/unit/cmdline.pyproject.test
TheCleric: Agreed. Will fix it up.

TheCleric:
@JukkaL I think other than the boolean func I left a comment on above, all the requested changes have been made.

test-data/unit/cmdline.pyproject.test
```diff
@@ -0,0 +1,82 @@
...
+--
+-- Note that # flags: --some-flag IS NOT SUPPORTED.
+-- Use # cmd: mypy --some-flag ...
+--
+-- '== Return code: <value>' is added to the output when the process return code
+-- is "nonobvious" -- that is, when it is something other than 0 if there are no
+-- messages and 1 if there are.
+
+-- Directories/packages on the command line
+-- ----------------------------------------
```
JukkaL, line 16: Could you add back some successful test cases as well that would cover all the essential features? It looks like all the test cases test failure modes.

Please add additional test cases to `check-*.test` files, since they are much faster than command-line tests.

JukkaL (commented):
Thanks for the updates! This is almost ready to merge. It would be good to have more tests (but not as many as there were originally). Using ValueError in the boolean value parser seems okay.

test-data/unit/cmdline.pyproject.test
TheCleric: Sorry, got a little overzealous in the removal. Will do.

TheCleric:
> Thanks for the updates! This is almost ready to merge. It would be good to have more tests (but not as many as there were originally). Using ValueError in the boolean value parser seems okay.

Alright @JukkaL, I tried to find a happy medium between too few and too many tests. 😆 

Let me know if I got a better balance this time.

dciborow:
as black is against supporting tox.ini, this is extra important to me to eventually have all of our configurations in a single spot...

https://github.com/psf/black/issues/2172

Right now we can not configure both black and mypy into the same file as black lacks support for tox.ini

JukkaL (approved):
Thanks for the final batch of updates! Looks good now.

mypy/config_parser.py
```diff
@@ -375,3 +532,17 @@ def set_strict_flags() -> None:
...
+
+
+def get_config_module_names(filename: Optional[str], modules: List[str]) -> str:
+    if not filename or not modules:
+        return ''
+
+    if not is_toml(filename):
+        return ", ".join("[mypy-%s]" % module for module in modules)
+
+    return "module = ['%s']" % ("', '".join(sorted(modules)))
```
xtofl, line 544: This looks a bit weird.  The quotes can be applied directly:

```python
quoted = '"{}"'.format
return "module = [%s]" % ", ".join(map(quoted, sorted(modules)))
```

xtofl (commented):
Thanks for this functionality!  I'm looking forward to having just one config file in my project.

I didn't find any errors, but I think some code may be expressed a bit nicer.

Kind regards

mypy/config_parser.py
```diff
@@ -100,7 +125,23 @@ def check_follow_imports(choice: str) -> str:
     'cache_dir': expand_path,
     'python_executable': expand_path,
     'strict': bool,
-}  # type: Final
+}  # type: Final[Dict[str, _INI_PARSER_CALLABLE]]
+
+# Reuse the ini_config_types and overwrite the diff
+toml_config_types = ini_config_types.copy()  # type: Final[Dict[str, _INI_PARSER_CALLABLE]]
+toml_config_types.update({
```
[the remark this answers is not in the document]
Gobot1234: Mypy works with 3.5+ and this change would break everything on <3.9

Issue #6760, opened 2019-05-03
Need plugin hooks for functions/methods


asvetlov:
For some frameworks like `pytest` it would be useful to have a plugin that works with functions.
Currently mypy has decorator hook for classes.

Some ideas for possible hooks: 
1. update function namespace to inject some names
2. modify function signature 
3. generate function signature (types)

For example pytest has the following fixture

```
from pathlib import Path

@pytest.fixture
def db_path(tmp_path: Path) -> Path:
    return tmp_path / 'test.db'
```

The fixture is used by name:
```
def test_a(db_path) -> None:
    ....
```
`db_path` here is a calucated result of `db_path()` fixture call.
It is bound by parameter name which is the same as fixture name.

Adding a type to `db_path` argument is tedious, especially if a fixture returns somethig complex like `Callable[[str, int], Awaitable[web.Response]`.

Technically all needed information can be extracted from fixture definition if the definition is type annotating.

A hook for pytest would be awesome but as @ilevkivskyi said `mypy` need a new hook to support it.

ilevkivskyi:
I like this idea, this will allow supporting some common dynamic features.

This will likely require local lookup functions to the public API (currently it is there on `SemanticAnalyzer` but not on `SemanticAnalyzerPluginInterface`).

blueyed:
JFI/bump: I wonder if this is only about having `get_function_signature_hook` (in line with `get_method_signature_hook`) - but from a quick test it seems like `get_method_signature_hook` also only gets called for method calls, not definitions.
I've also looked into this because of pytest.

JukkaL:
Brainstorming:
* We could have a hook that gets called after semantic analysis and before type checking that can adjust the signature of each function/method. 
* We could have a signature adjustment hook that gets called before type checking the body of a function. However, this seems more complicated -- for consistency, we also need another hook that gets called for function calls. The benefit here would be that the hook could access inferred types of other functions and attributes.
* We could have a hook that gets called for each module after semantic analysis, which would allow arbitrary changes to module namespaces. However, the implementation would have to be pretty tightly coupled with the mypy symbol table and AST implementation, so this might not the best option. Also, this might not work well with mypy daemon.

MrGreenTea:
> Brainstorming:
> 
>     * We could have a hook that gets called after semantic analysis and before type checking that can adjust the signature of each function/method.

I was quite suprised this is not already the case, as I understand it exists for classes / methods with the `get_method_signature_hook` already.

[...]

PR #10222, opened 2021-03-18
Update extending_mypy.rst

sobolevn:
Two things I have changed:
1. `get_function_signature_hook` was missing from docs, but it is defined https://github.com/python/mypy/blob/ac66403cb4374b858786350b96e7f5972db84bcd/mypy/plugin.py#L368-L376
2. While reading the docs, this `UserDefined` for `get_class_decorator_hook` example caught my eye. It was not clear how `UserDefined` and `@customize` work. I believe that `@dataclass` example is better for two reasons. First, it is an existing API and is already familiar to our users. Secondly, there's an actual plugin for it, so people can see how it is implemented

Refs https://github.com/python/mypy/issues/6760

### DIFF
--- a/docs/source/extending_mypy.rst
+++ b/docs/source/extending_mypy.rst
@@ -173,6 +173,8 @@ For example:
        ...
        yield timer()
 
+**get_function_signature_hook** is used to adjust the signature of a function.
+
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
Thank you!

PR #10223, opened 2021-03-18
Check kwargs key type as covariant

### LINKED ISSUE #10023: Mypy incorrectly warns when passing literals as kwarg keywords
**Bug Report**

```
from typing import Any, Dict, Literal
kw: Dict[Literal["a", "b"], Any]
def func(a, b):
    pass
func(**kw)
```
Leads to
```
main.py:8: error: Keywords must be strings
```

https://mypy-play.net/?mypy=latest&python=3.9&gist=582cf5e8c0a6e6a5881d391da64e1455

**Expected Behavior**
Expect this to be allowable - as these literals count as strings

kamilturek:
Closes #10023.

Assume mapping's key type is covariant when validating `**kwargs` argument type.

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -32,7 +32,7 @@ from mypy.nodes import (
     YieldFromExpr, TypedDictExpr, PromoteExpr, NewTypeExpr, NamedTupleExpr, TypeVarExpr,
     TypeAliasExpr, BackquoteExpr, EnumCallExpr, TypeAlias, SymbolNode, PlaceholderNode,
     ParamSpecExpr,
-    ARG_POS, ARG_OPT, ARG_NAMED, ARG_STAR, ARG_STAR2, LITERAL_TYPE, REVEAL_TYPE,
+    ARG_POS, ARG_OPT, ARG_NAMED, ARG_STAR, ARG_STAR2, COVARIANT, LITERAL_TYPE, REVEAL_TYPE,
 )
 from mypy.literals import literal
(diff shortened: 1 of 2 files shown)


mypy/checkexpr.py
```diff
@@ -3938,20 +3938,15 @@ def is_valid_var_arg(self, typ: Type) -> bool:
...
-                is_subtype(typ, self.chk.named_generic_type(
-                    'typing.Mapping',
-                    [self.named_type('builtins.unicode'),
-                     AnyType(TypeOfAny.special_form)])))
+            key_type = UnionType.make_union([
+                self.named_type('builtins.str'),
+                self.named_type('builtins.unicode')])
+        kwargs_type = self.chk.named_generic_type(
+            'typing.Mapping', [key_type, AnyType(TypeOfAny.special_form)])
+        kwargs_type.type.defn.type_vars[0].variance = COVARIANT
```
JelleZijlstra, line 3948: Wouldn't this affect all uses of Mapping after this code runs?

kamilturek: Ahh, you're right. Didn't realized that before.

vickyhuo: The fix I was thinking was something along these lines. 
```suggestion
        is_subtype(typ, self.chk.named_type('typing.Mapping')) and isinstance(typ.args[0], LiteralType)
```
This wouldn't cover the case of Literal["a", "b"] but you can try using the `try_getting_str_literals_from_type` method

PR #10226, opened 2021-03-20
[mypyc] Support class-based TypedDict definitions

JukkaL:
These need to be compiled as non-extension classes.

Also mypy has a special internal representation for these which doesn't
correspond to what's there at runtime and requires some hacks.

Fixes mypyc/mypyc#803.


TH3CHARLie (approved):

JukkaL:
In 3.9 `TypedDict` is no longer a real type object. It's not obvious what's the best way to support it. This is quite tricky.

PR #10231, opened 2021-03-21
double quotes in 'object is not iterable' error message

dixith:
### Description
This pull request fixes issue reported in #7445 for **object is not iterable** error message

## Test Plan
Updated test cases for the message. New test cases  not added.


hauntsaninja:
Thanks!

PR #10237, opened 2021-03-22
Pass literals as kwargs

### LINKED ISSUE #10023: Mypy incorrectly warns when passing literals as kwarg keywords

vickyhuo:
Closes #10023 

Checks if kwarg is Mapping type AND check if list of strings extracted from mapping key type is not empty

## Test Plan

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3937,11 +3937,9 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
     def is_valid_keyword_var_arg(self, typ: Type) -> bool:
         """Is a type valid as a **kwargs argument?"""
-        import pdb; pdb.set_trace()
-        str_type = self.named_type('builtins.str')
         if self.chk.options.python_version[0] >= 3:
             return (is_subtype(typ, self.chk.named_generic_type(
-                'typing.Mapping', [str_type,
+                'typing.Mapping', [self.named_type('builtins.str'),
                                    AnyType(TypeOfAny.special_form)]))
                     or
                     (is_subtype(typ, self.chk.named_type('typing.Mapping')) and
@@ -3951,7 +3949,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             return (
                 is_subtype(typ, self.chk.named_generic_type(
                     'typing.Mapping',
-                    [str_type,
+                    [self.named_type('builtins.str'),
                      AnyType(TypeOfAny.special_form)]))
                 or
                 is_subtype(typ, self.chk.named_generic_type(
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -3943,7 +3943,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                                    AnyType(TypeOfAny.special_form)]))
                     or
                     (is_subtype(typ, self.chk.named_type('typing.Mapping')) and
-                    try_getting_str_literals_from_type(typ.args[0]) is not None))
+                    try_getting_str_literals_from_type(typ.args[0])))
 
         else:
             return (
@@ -3958,7 +3958,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
                      AnyType(TypeOfAny.special_form)]))
                 or
                 (is_subtype(typ, self.chk.named_type('typing.Mapping')) and
-                 try_getting_str_literals_from_type(typ.args[0]) is not None))
+                 try_getting_str_literals_from_type(typ.args[0])))
 
     def has_member(self, typ: Type, member: str) -> bool:
         """Does type have member with the given name?"""
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -6,7 +6,7 @@ import itertools
 from typing import (
     Any, cast, Dict, Set, List, Tuple, Callable, Union, Optional, Sequence, Iterator
 )
-from typing_extensions import ClassVar, Final, overload
+from typing_extensions import ClassVar, Final, overload, Literal
 
 from mypy.errors import report_internal_error
 from mypy.typeanal import (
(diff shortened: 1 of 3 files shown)


test-data/unit/check-kwargs.test
```diff
@@ -491,3 +491,28 @@ m = {} # type: Mapping[str, object]
 f(**m)
 g(**m)  # TODO: Should be an error
 [builtins fixtures/dict.pyi]
+
+
+[case testGivingArgumentAsPositionalAndKeywordArg3]
```
JelleZijlstra, line 496: Should also error if you pass a kwargs that doesn't exist (e.g., you pass a dict with keys `Literal["c", "d"]` to a function defined as `def f(a, b)`).

vickyhuo: Hey @JelleZijlstra, for that case, should the error message be `Unexpected keyword argument` or `Missing positional argument`?

JelleZijlstra: Probably both? It should match what we do if you just call that function as `f(c=..., d=...)`.

mypy/checkexpr.py
```diff
@@ -1439,6 +1439,23 @@ def check_for_extra_actual_arguments(self,
                                                                             context)
                                 is_unexpected_arg_error = True
                         ok = False
+                elif isinstance(actual_type, Instance) and \
+                        actual_type.type.has_base('typing.Mapping'):
+                    if messages and actual_type.args:
```
JukkaL, line 1444: The type does not necessarily have the same args as `Mapping`. For example, we could have a class like this:

```py
class C(Mapping[str, T]): ...
```

To make this work in general, you can use `map_instance_to_supertype`.

mypy/checkexpr.py
```diff
@@ -3937,21 +3954,34 @@ def is_valid_var_arg(self, typ: Type) -> bool:
...
-                                   AnyType(TypeOfAny.special_form)]))
+            return (
+                (is_subtype(typ, self.chk.named_generic_type(
+                    'typing.Mapping',
+                    [self.named_type('builtins.str'),
+                     AnyType(TypeOfAny.special_form)])))
+                or
+                (is_subtype(typ, self.chk.named_type(
+                    'typing.Mapping')) and
+                    typ.args and  # type: ignore
```
JukkaL, line 3967: Similar to above, you'll need to use `map_instance_to_supertype`. Also, ensure that `typ` is an instance of `Instance`, and you can get rid of the `# type: ignore`. Right now this is unsafe, since TypedDict types can be subtypes of `Mapping` but they don't have `args`.

mypy/checkexpr.py
```diff
@@ -3937,21 +3954,34 @@ def is_valid_var_arg(self, typ: Type) -> bool:
...
+                     AnyType(TypeOfAny.special_form)])))
                 or
-                is_subtype(typ, self.chk.named_generic_type(
+                (is_subtype(typ, self.chk.named_generic_type(
                     'typing.Mapping',
                     [self.named_type('builtins.unicode'),
                      AnyType(TypeOfAny.special_form)])))
+                or
+                (is_subtype(typ, self.chk.named_type('typing.Mapping')) and
+                    typ.args and  # type: ignore
```
JukkaL, line 3983: Similar to above.

mypy/checkexpr.py
```diff
@@ -1439,6 +1439,23 @@ def check_for_extra_actual_arguments(self,
                                                                             context)
                                 is_unexpected_arg_error = True
                         ok = False
+                elif isinstance(actual_type, Instance) and \
```
JukkaL, line 1442: Style nit: We prefer extra parentheses instead of `\` to deal with expressions split into multiple lines.

test-data/unit/check-kwargs.test
```diff
@@ -491,3 +491,34 @@ m = {} # type: Mapping[str, object]
...
+v = {'a':2, 'b':1} # type: Mapping[Literal['a','b'], int]
+f(**v)
+
+w = {'c':2} # type: Mapping[Literal['d'], int] \
+# E: Dict entry 0 has incompatible type "Literal['c']": "int"; expected "Literal['d']": "int"
+f(**w)
+
+x = {'c':1, 'd': 2} # type: Mapping[Literal['c','d'], int]
+g(**x) # E: Unexpected keyword argument "c" for "g"
+h(**x) # E: Missing positional arguments "a", "b" in call to "h"
```
JukkaL, line 522: Test only giving one of the two positional arguments (e.g., only `b`).

97littleleaf11:
Previous PR: https://github.com/python/mypy/pull/9629

mypy/checkexpr.py
```diff
@@ -4007,12 +4027,19 @@ def is_valid_var_arg(self, typ: Type) -> bool:
...
-                is_subtype(typ, self.chk.named_generic_type('typing.Mapping',
-                    [UninhabitedType(), UninhabitedType()])) or
-                isinstance(typ, ParamSpecType)
+            is_subtype(typ, mapping_type) or
+            (isinstance(typ, Instance) and
+                is_subtype(typ, self.chk.named_type('typing.Mapping')) and
+                try_getting_str_literals_from_type(map_instance_to_supertype(
+                    typ, mapping_type.type).args[0]) is not None) or
+            is_subtype(typ, self.chk.named_generic_type('typing.Mapping',
+                                                        [UninhabitedType(), UninhabitedType()])) or
```
97littleleaf11, line 4041: Actually I don't know how this line works. Could somebody elaborate? @JukkaL @momohatt

momohatt: Sorry for being late.
As we discussed in #9629, this check is added to avoid false-positive errors when empty dictionaries are passed with double-stars (e.g., `**{}`). This condition is needed because the type of an empty dictionary is inferred to be `dict[<nothing>, <nothing>]`, which is not a subtype of `Mapping[str, Any]`.

97littleleaf11: @momohatt Thanks!

mypy/checkexpr.py
```diff
@@ -1490,6 +1490,27 @@ def check_for_extra_actual_arguments(self,
                                                                         context)
                             is_unexpected_arg_error = True
                         ok = False
+                elif (isinstance(actual_type, Instance) and
+                        actual_type.type.has_base('typing.Mapping')):
+                    any_type = AnyType(TypeOfAny.special_form)
+                    mapping_info = self.chk.named_generic_type('typing.Mapping',
+                                                               [any_type, any_type]).type
+                    supertype = map_instance_to_supertype(actual_type, mapping_info)
+                    if messages and supertype.args:
```
andersk, line 1499: #12699 deleted `messages` in favor of `self.msg`.

Issue #811, opened 2015-08-18
Support raising tuples


jingw:
```
error: Exception must be derived from BaseException
```

from:
```
raise exc[0], exc[1], exc[2]
```

JukkaL:
Fixed in the `hack` branch.

PR #10245, opened 2021-03-25
[mypyc] Don't coerce types checked with isinstance (#811)

sinback:
Prevent the coercion of things whose types are checked with isinstance
to other types - such coercion is not necessary and can lead to bugs.

Also, add test for avoiding coercion of isinstance() comparators.

Fixes mypyc/mypyc#811.

## Test Plan

Ensured via compiling by hand and running the example given in mypyc/mypyc#811 has been fixed; added a test for the new code path that is generated.

### DIFF
--- a/mypyc/irbuild/specialize.py
+++ b/mypyc/irbuild/specialize.py
@@ -248,7 +248,11 @@ def translate_next_call(builder: IRBuilder, expr: CallExpr, callee: RefExpr) -> 
 
 @specialize_function('builtins.isinstance')
 def translate_isinstance(builder: IRBuilder, expr: CallExpr, callee: RefExpr) -> Optional[Value]:
-    # Special case builtins.isinstance
+    # Special case for builtins.isinstance
+    # Prevent coercions on the thing we are checking the instance of - there is no need to coerce
+    # something to a new type before checking what type it is, and the coercion could lead to bugs.
+    builder.types[expr.args[0]] = AnyType(TypeOfAny.from_error)
+
     if (len(expr.args) == 2
             and expr.arg_kinds == [ARG_POS, ARG_POS]
             and isinstance(expr.args[1], (RefExpr, TupleExpr))):
--- a/mypyc/irbuild/specialize.py
+++ b/mypyc/irbuild/specialize.py
@@ -248,14 +248,15 @@ def translate_next_call(builder: IRBuilder, expr: CallExpr, callee: RefExpr) -> 
 
 @specialize_function('builtins.isinstance')
 def translate_isinstance(builder: IRBuilder, expr: CallExpr, callee: RefExpr) -> Optional[Value]:
-    # Special case for builtins.isinstance
-    # Prevent coercions on the thing we are checking the instance of - there is no need to coerce
-    # something to a new type before checking what type it is, and the coercion could lead to bugs.
-    builder.types[expr.args[0]] = AnyType(TypeOfAny.from_error)
-
     if (len(expr.args) == 2
             and expr.arg_kinds == [ARG_POS, ARG_POS]
             and isinstance(expr.args[1], (RefExpr, TupleExpr))):
+        # Special case for builtins.isinstance
+        # Prevent coercions on the thing we are checking the instance of - there is no need to
+        # coerceomething to a new type before checking what type it is, and the coercion could lead
+        # to bugs.
+        builder.types[expr.args[0]] = AnyType(TypeOfAny.from_error)
+
         irs = builder.flatten_classes(expr.args[1])
         if irs is not None:
             return builder.builder.isinstance_helper(builder.accept(expr.args[0]), irs, expr.line)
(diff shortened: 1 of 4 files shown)


mypyc/irbuild/specialize.py
```diff
@@ -248,7 +248,11 @@ def gen_inner_stmts() -> None:
 
 @specialize_function('builtins.isinstance')
 def translate_isinstance(builder: IRBuilder, expr: CallExpr, callee: RefExpr) -> Optional[Value]:
-    # Special case builtins.isinstance
+    # Special case for builtins.isinstance
+    # Prevent coercions on the thing we are checking the instance of - there is no need to coerce
+    # something to a new type before checking what type it is, and the coercion could lead to bugs.
+    builder.types[expr.args[0]] = AnyType(TypeOfAny.from_error)
```
JukkaL, line 254: This could be moved to within the if statement, so that if somehow there aren't any arguments, we won't get an `IndexError`.

mypyc/test-data/irbuild-isinstance.test
```diff
@@ -0,0 +1,97 @@
+[case testIsinstanceInt]
+def is_int(value):
+    return isinstance(value, int)
```
JukkaL, line 3: The IR could be a little simpler if the function would have an annotations, e.g. `def is_int(value: object) -> bool`. IF there is no annotation, more things have `Any` types, often resulting in some needless boxing.

Not a big deal though.

mypyc/test-data/irbuild-isinstance.test
```diff
@@ -0,0 +1,97 @@
...
+    r0 = load_address PyLong_Type
+    r1 = PyObject_IsInstance(value, r0)
+    r2 = r1 >= 0 :: signed
+    r3 = truncate r1: int32 to builtins.bool
+    r4 = box(int32, r1)
+    return r4
+
+[case testIsinstanceNotBool1]
+def is_not_bool(value):
+    return not isinstance(value, bool)
```
JukkaL, line 22: Here adding a type annotation to the function could be more useful, since the IR is pretty verbose right now.

mypyc/test-data/irbuild-isinstance.test
```diff
@@ -0,0 +1,97 @@
...
+    r7 = PyObject_Not(r6)
+    r8 = r7 >= 0 :: signed
+    r9 = truncate r7: int32 to builtins.bool
+    r10 = box(bool, r9)
+    return r10
+
+[case testIsinstanceNotBoolAndIsInt]
+# This test is to ensure that 'value' doesn't get coerced into a bool when we need it to be an int
+def is_not_bool_and_is_int(value):
+    return not isinstance(value, bool) and isinstance(value, int)
```
JukkaL, line 54: Similar to above, can you try adding an annotation (even simple one such as `Any` -> `bool`) in case it simplifies the IR?

Also, it would be good to have a run test case for this. This could go into `mypyc/test-data/run-integers.test`. No need to do this for the other test cases above.

JukkaL (commented):
Thanks for the PR! Looks good, just some minor comments for extra polish.

mypyc/test-data/run-integers.test
```diff
@@ -157,6 +157,15 @@ def check_bitwise(x: int, y: int) -> None:
             check_or(ll, rr)
             check_xor(ll, rr)
 
+[case testIsinstanceNotBoolAndIsInt]
+def test_isinstance_not_bool_and_is_int(value: object) -> bool:
+    return not isinstance(value, bool) and isinstance(value, int)
```
JukkaL, line 162: Actually this already seems to work correctly. Can you switch the checks so that it will reproduce the incorrect old behavior? I.e., like this:
```py
return isinstance(value, int) and not isinstance(value, bool)
```

mypyc/test-data/irbuild-isinstance.test
JukkaL: Similar to below -- to test that the bug fixed, can you switch the order of the `isinstance` checks?

JukkaL (commented):
Thanks for the updates! Left a few further comments about the test cases.

sinback:
Oops, good catch.

mypyc/test-data/irbuild-isinstance.test
```diff
@@ -0,0 +1,76 @@
...
+    r1 = 'bool'
+    r2 = CPyObject_GetAttr(r0, r1)
+    r3 = PyObject_IsInstance(value, r2)
+    r4 = r3 >= 0 :: signed
+    r5 = truncate r3: int32 to builtins.bool
+    r6 = r5 ^ 1
+    return r6
+
+[case testIsinstanceIntAndNotBool]
+# This test is to ensure that 'value' doesn't get coerced into a bool when we need it to be an int
```
JukkaL, line 41: Hmm the comment is a bit unclear. Maybe something like "ensure that 'value' doesn't get coerced to int when we are checking if it's a bool, since an int can never be an instance of bool"?

JukkaL (commented):
One more suggestion about a comment, then this is ready to merge.

JukkaL (approved):

PR #10251, opened 2021-03-27
Use context managers for disable_errors()

dixith:
### Description
This PR fixes one of the issue reported in #1184

## Test Plan
Since this change is coding style change, new cases are not added.


mypy/messages.py
```diff
@@ -136,8 +137,11 @@ def add_errors(self, messages: 'MessageBuilder') -> None:
                 for info in errs:
                     self.errors.add_error_info(info)
 
-    def disable_errors(self) -> None:
+    @contextmanager
+    def disable_errors(self) -> Iterator[None]:
         self.disable_count += 1
+        yield
```
JelleZijlstra, line 143: Should use try-finally

mypy/messages.py
```diff
@@ -136,8 +137,11 @@ def add_errors(self, messages: 'MessageBuilder') -> None:
...
+        self.disable_count -= 1
 
     def enable_errors(self) -> None:
```
JelleZijlstra, line 146: Can this method be removed?

JelleZijlstra (changes requested):
Thanks for your contribution!

mypy/messages.py
dixith: Removed enable_errors()

mypy/messages.py
dixith: Thank you for the review @JelleZijlstra 

yield should be inside try-finally? 

`@contextmanager
    def disable_errors(self) -> Iterator[None]:
        self.disable_count += 1
        try:
            yield
        finally:
            pass
        self.disable_count -= 1`

JelleZijlstra: And the `-= 1` should be in the finally block. The point is to make sure the cleanup code always happens.

dixith: Got it.

dixith: @JelleZijlstra Now that we have try-finally inside `enable_errors()` below try-finally is redundent right? Can be removed? https://github.com/python/mypy/blob/8860575a2134d87d890119015a9ead172f7fdfb1/mypy/checkexpr.py#L2811

```python
if right_map is None:
  with self.msg.disable_errors():
     try:
        right_type = self.analyze_cond_branch(right_map, e.right, left_type)
     finally:
        pass
```

dixith:
Getting below error. I should use `TypeMap` type for `right_map`?
```mypy/checkexpr.py:2772: error: Local variable 'right_map' has inferred type None; add an annotation```

mypy/messages.py
JelleZijlstra: A try-finally with an empty finally block doesn't do anything useful, so yes you can remove it.

dixith:
@JelleZijlstra 

Getting below error. Do you ahve any suggestions? 
```mypy/checkexpr.py:2772: error: Local variable 'right_map' has inferred type None; add an annotation```

JelleZijlstra:
Sorry, I'm not too familiar with that code. I'd try adding an annotation though, as the error message suggests.

JelleZijlstra:
Sorry for dropping this! We ended up merging #10569 instead.

PR #10255, opened 2021-03-28
Fix Generators typeshed test

hatal175:
### Description

typshed mistakenly set itertools.count as a function while it is actually a class. Meaning that it returns a class instance and then this clashes with the filter assignment in the test.
I want to fix the typeshed side but the mypy test on THEIR repo fails because of this.

## Test Plan

Mypy behaves the same but tests should be more in line with itertools actual typing.

### DIFF
--- a/test-data/samples/generators.py
+++ b/test-data/samples/generators.py
@@ -6,7 +6,7 @@ from typing import Iterator
 
 def iter_primes() -> Iterator[int]:
     # an iterator of all numbers between 2 and +infinity
-    numbers = itertools.count(2)
+    numbers: Iterator[int] = itertools.count(2)
 
     # generate primes forever
     while True:


hauntsaninja:
Thanks for this! Yeah, in general the mypy-typeshed dependency can cause some friction. In this case, I'm not sure that we should actually make this change in typeshed... Opened https://github.com/python/typeshed/issues/5145 to discuss :-)

JelleZijlstra:
Even if we don't make the change in typeshed, it seems harmless to make this change so that we make CI for the two projects a bit less interdependent, so I'm going to merge it.

TH3CHARLie:
This has caused `test-data/samples/generators.py:9: error: Variable annotation syntax is only supported in Python 3.6 and greater
` on master

JelleZijlstra:
Sorry for missing that, fixed it in #10315

PR #10257, opened 2021-03-28
[mypyc] Add a primitive for loading bool type

97littleleaf11:
### Description

Closes https://github.com/mypyc/mypyc/issues/821

## Test Plan

For this piece of python code:

```py
def f(x: object) -> bool:
    return isinstance(x, bool)
```

Previous version would generate:

```
def f(x):                                      
    x, r0 :: object                            
    r1 :: str                                  
    r2 :: object                               
    r3 :: int32                                
    r4 :: bit                                  
    r5 :: bool                                 
L0:                                            
    r0 = builtins :: module                    
    r1 = 'bool'                                
    r2 = CPyObject_GetAttr(r0, r1)             
    r3 = PyObject_IsInstance(x, r2)            
    r4 = r3 >= 0 :: signed                     
    r5 = truncate r3: int32 to builtins.bool   
    return r5   
```

And now it would only generate this:

```
def f(x):                                      
    x, r0 :: object                            
    r1 :: int32                                
    r2 :: bit                                  
    r3 :: bool                                 
L0:                                            
    r0 = load_address PyBool_Type              
    r1 = PyObject_IsInstance(x, r0)            
    r2 = r1 >= 0 :: signed                     
    r3 = truncate r1: int32 to builtins.bool   
    return r3    
```

### DIFF
--- a/mypyc/primitives/misc_ops.py
+++ b/mypyc/primitives/misc_ops.py
@@ -9,6 +9,11 @@ from mypyc.primitives.registry import (
     function_op, custom_op, load_address_op, ERR_NEG_INT
 )
 
+# Get the 'bool' type object.
+load_address_op(
+    name='builtins.bool',
+    type=object_rprimitive,
+    src='PyBool_Type')
 
 # Get the boxed Python 'None' object
 none_object_op = load_address_op(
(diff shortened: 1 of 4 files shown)


JukkaL (approved):
Thanks, looks good!

PR #10258, opened 2021-03-28
stubtest: check overloads are functions, don't early return

### LINKED ISSUE #10254: overload hides 'is not a function' error in stubtest
**Bug Report**

I've noticed that overloading a function can hide a mismatch between the runtime type and the stub type. I noticed this as I was going over typeshed stubtest exceptions, specifically for itertools.accumulate.
Tests for python 3.6 and 3.7 failed because the stubs declared accumulate as a function while it was actually a class. However for 3.8 it is defined as two functions with overload and the tests passed successfully.

**To Reproduce**

(Write your steps here:)

1. Run stubtest for typeshed in python 3.8.
2. Test passes.
3. Remove one of the overloads in itertools.pyi
4. run again and see it fails.

**Expected Behavior**

I t should fail on a type mismatch like it does for no overloads.

**Actual Behavior**

passed.

**Your Environment**

Ran from latest python 3.8 and latest mypy from github.

hauntsaninja:
Fixes #10254

### DIFF
--- a/mypy/stubtest.py
+++ b/mypy/stubtest.py
@@ -265,7 +265,7 @@ def verify_typeinfo(
 
 
 def _verify_static_class_methods(
-    stub: nodes.FuncItem, runtime: types.FunctionType, object_path: List[str]
+    stub: nodes.FuncBase, runtime: Any, object_path: List[str]
 ) -> Iterator[str]:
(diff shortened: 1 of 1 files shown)


JukkaL (approved):
LGTM

PR #9707, opened 2020-11-08
[mypyc] `set(f(x) for x in it)` no longer uses list comprehension (#771)

mehdigmira:
### Description

Solves https://github.com/mypyc/mypyc/issues/771

## Test Plan

Added statements to the run-sets.test file

PR #10261, opened 2021-03-29
[mypyc] Faster set creation with generator expression

97littleleaf11:
### Description

Closes https://github.com/mypyc/mypyc/issues/771 and https://github.com/python/mypy/pull/9707

Changes:
* Move code from `tranform_set_comprehension` to `translate_set_comprehension`
* Fix the unnecessary set creation (https://github.com/python/mypy/pull/9707#issuecomment-725549735)

## Test Plan

Both generated IR test and functional test:
* Set built from iterators (`list` `tuple` `dict` `range`)
* Set built from nested generator

### DIFF
--- a/mypyc/irbuild/specialize.py
+++ b/mypyc/irbuild/specialize.py
@@ -28,8 +28,8 @@ from mypyc.primitives.dict_ops import dict_keys_op, dict_values_op, dict_items_o
 from mypyc.primitives.tuple_ops import new_tuple_set_item_op
 from mypyc.irbuild.builder import IRBuilder
 from mypyc.irbuild.for_helpers import (
-    translate_list_comprehension, comprehension_helper,
-    for_loop_helper_with_index
+    translate_list_comprehension, translate_set_comprehension,
+    comprehension_helper, for_loop_helper_with_index,
 )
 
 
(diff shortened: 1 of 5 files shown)


JukkaL:
Thanks for the PR! Have you benchmarked this against master? (If not, no worries, I can do some quick benchmarking tomorrow.)

97littleleaf11:
> Thanks for the PR! Have you benchmarked this against master? (If not, no worries, I can do some quick benchmarking tomorrow.)

Not yet.

mypyc/irbuild/specialize.py
```diff
@@ -143,6 +143,9 @@ def translate_safe_generator_call(
                 val = tuple_from_generator_helper(builder, expr.args[0])
                 if val is not None:
                     return val
+            if callee.fullname == "builtins.set":
+                return translate_set_comprehension(builder, expr.args[0])
+
```
97littleleaf11, line 148: Maybe I should move `builtins.set` to a new specialize function ?

JukkaL: That sounds reasonable, since the implementation isn't sharing much with the other cases.

JukkaL:
I measured the performance and now the speed of calling `set` with a generator expression is the same as a set comprehension. Nice! Calling `set` was 62% faster in the ad-hoc benchmark that I used, compared to master.

JukkaL (approved):
Thanks, looks good! This is a nice performance win.

PR #10265, opened 2021-03-30
[mypyc] Faster min

ChetanKhanna:
### Description
I have tried to specialize `min(x, y)` in this PR as per https://github.com/mypyc/mypyc/issues/773

## Test Plan
I ran the complete test suite locally as well as https://github.com/mypyc/mypyc-benchmarks
However, I'm still unsure as to how am I supposed to report benchmark results. After running the benchmark runner on both master and this branch for 5 time, I got the following results (averaged):
*master*: 11.7058x faster
*current branch*: 11.8452x faster


mypyc/irbuild/specialize.py
```diff
@@ -291,3 +291,16 @@ def translate_isinstance(builder: IRBuilder, expr: CallExpr, callee: RefExpr) ->
...
     return None
+
+
+@specialize_function('builtins.min')
+def faster_min(builder: IRBuilder, expr: CallExpr, callee: RefExpr) -> Optional[Value]:
+    if (len(expr.args) == 2
+            and expr.arg_kinds == [ARG_POS, ARG_POS]):
+        x, y = builder.accept(expr.args[0]), builder.accept(expr.args[1])
+        comparison = builder.binary_op(x, y, '<', expr.line)
+        if comparison == Integer(1,  bool_rprimitive):
```
JukkaL, line 302: This isn't right. The if statement runs during compilation, but it should instead generate an if statement that is executed when the compiled code is run.

mypyc/test-data/run-integers.test
```diff
@@ -328,3 +328,14 @@ def test_int_as_bool() -> None:
     for x in 1, 55, -1, -7, 1 << 50, 1 << 101, -(1 << 50), -(1 << 101):
         assert is_true(x)
         assert not is_false(x)
+
+[case testMin]
+def check_min_int() -> None:
+    x: int = 200
+    y: int = 30
+    assert min(x, y) == 30
```
JukkaL, line 336: Also test the case where the first argument is the minimum. I think that it's going to fail.

To help debug this, also add an irbuild test case for this. This way you can validate that the code you are generating is what you expect it to be.

JukkaL (changes requested):

TH3CHARLie:
If I recall correctly, the specializer logic would pick the first specializer of a function name. Because there is already a registered `builtins.min` for generator call, I doubt you actually specialize the call and the performance numbers look like noise to me. As Jukka suggested, write a simple IR test and see if the IR matches your desired case.

ChetanKhanna:
Okay, I think I'll have to dig in deeper then. I'll keep on updating this thread with all the progress. Thank you :)

ChetanKhanna:
Hello! Okay so I was looking into it and I have a couple of doubts:-
1. Writing an `irbuild` test helped. However, I feel there is an extra block in the output and I'm not sure how to avoid it from getting generated while compiling. Here's the output and I think block `L6` shouldn't be there. If anyone could suggest how to possibly handle this please?

```
def f(x, y):
      x, y :: int
      r0 :: int64
      r1 :: bit
      r2 :: int64
      r3, r4, r5 :: bit
      r6 :: bool
      r7 :: bit
      r8 :: object
      r9 :: str
      r10, r11, r12, r13 :: object
      r14 :: int
  L0:
      r0 = x & 1
      r1 = r0 == 0
      r2 = y & 
      r3 = r2 == 0
      r4 = r1 & r3
      if r4 goto L1 else goto L2 :: bool
  L1:
      r5 = x < y :: signed
      r6 = r5
      goto L3
  L2:
      r7 = CPyTagged_IsLt_(x, y)
      r6 = r7
  L3:
      if r6 goto L4 else goto L5 :: bool
  L4:
      return x
  L5:
      return y
  L6:
      r8 = builtins :: module
      r9 = 'min'
      r10 = CPyObject_GetAttr(r8, r9)
      r11 = box(int, x)
      r12 = box(int, y)
      r13 = PyObject_CallFunctionObjArgs(r10, r11, r12, 0)
      r14 = unbox(int, r13)
      return r14
```

2. As @TH3CHARLie said that `builtins.min` is already registered and shouldn't be registered again on the new function. Then should the new function be shifted inside the `translate_safe_generator_call` which originally registers `min`?

TH3CHARLie:
For Q1: I suggest you update your local commit so we can figure out why L6 appears. Your guess is correct, it shouldn't be here, it's a generic implementation that is usually slower than direct integer comparison.

For Q2: my advice is to make the specializers from a map of <name, function> to <name, list of functions> and update the lookup logic accordingly. But this seems a little bit heavy, not sure how @JukkaL thinks.

ChetanKhanna:
> For Q1: I suggest you update your local commit so we can figure out why L6 appears. Your guess is correct, it shouldn't be here, it's a generic implementation that is usually slower than direct integer comparison.

I've pushed my local commits, although they are still WIP

> For Q2: my advice is to make the specializers from a map of <name, function> to <name, list of functions> and update the lookup logic accordingly. But this seems a little bit heavy, not sure how @JukkaL thinks.

Okay, that makes sense. I can proceed that way if that's okay.

mypyc/test-data/run-integers.test
ChetanKhanna: I get why you think that way. I've updated the logic a bit to get a better IR. I tested locally and the other case seem to work fine as well. I'll add that test case before pushing next time.

mypyc/irbuild/specialize.py
ChetanKhanna: I've updated this. However, I posted a question below. Any suggestions please?

97littleleaf11:
> For Q2: my advice is to make the specializers from a map of <name, function> to <name, list of functions> and update the lookup logic accordingly. But this seems a little bit heavy, not sure how @JukkaL thinks.

Agree with this idea. Recent speed-up commits and issues have and will change a large range of code structure. Maybe we might open a new issue to discuss this?

JukkaL:
> Maybe we can open a new issue to discuss this?

This is a good idea. My first impression is that having a list of functions is probably fine, but it's worth thinking about it a bit more.

mypyc/test-data/irbuild-int.test
```diff
@@ -80,3 +80,49 @@ L14:
...
+    return y                                 
+L6:                                          
+    r8 = builtins :: module                  
+    r9 = 'min'                               
+    r10 = CPyObject_GetAttr(r8, r9)          
+    r11 = box(int, x)                        
+    r12 = box(int, y)                        
+    r13 = PyObject_CallFunctionObjArgs(r10, r11, r12, 0)
+    r14 = unbox(int, r13)                    
+    return r14
```
ChetanKhanna, line 128: This is obviously wrong. However, the same code when compiled using mypyc separately gives a different IR. Here's what's generated:
```
def f_min(x, y): 
    x, y :: int 
    r0 :: int64
    r1 :: bit 
    r2 :: int64
    r3, r4, r5 :: bit 
    r6 :: bool
    r7 :: bit 
L0:
    r0 = x & 1 
    r1 = r0 == 0
    r2 = y & 1 
    r3 = r2 == 0
    r4 = r1 & r3
    if r4 goto L1 else goto L2 :: bool
L1:
    r5 = x < y :: signed
    r6 = r5
    goto L3
L2:
    r7 = CPyTagged_IsLt_(x, y)
    r6 = r7
L3:
    if r6 goto L4 else goto L5 :: bool
L4:
    inc_ref x :: int 
    return x
L5:
    inc_ref y :: int 
    return y
```
Not sure  why L6 gets generated in the test suite though :confused:

mypyc/irbuild/specialize.py
```diff
@@ -302,3 +302,20 @@ def translate_isinstance(builder: IRBuilder, expr: CallExpr, callee: RefExpr) ->
         if irs is not None:
             return builder.builder.isinstance_helper(builder.accept(expr.args[0]), irs, expr.line)
     return None
+
+
+@specialize_function('builtins.min')
+def faster_min(builder: IRBuilder, expr: CallExpr, callee: RefExpr) -> None:
+    if (len(expr.args) > 0
```
97littleleaf11, line 309: The return value should be `Optional[Value]`. The the call translation functions in `irbuild/expression.py` can get the correct result and stop further trying.

97littleleaf11 (changes requested):

lastthyme (approved):

97littleleaf11:
lgtm! This PR brings a promising speed-up:
```
running min_max_pair
..........
interpreted: 0.456433s (avg of 5 iterations; stdev 3.1%)
compiled:    0.264439s (avg of 5 iterations; stdev 0.33%)

compiled is 1.726x faster
```
Master branch is about 0.9x.

mypyc/test-data/run-integers.test
```diff
@@ -328,3 +328,14 @@ def test_int_as_bool() -> None:
     for x in 1, 55, -1, -7, 1 << 50, 1 << 101, -(1 << 50), -(1 << 101):
         assert is_true(x)
         assert not is_false(x)
+
+[case testMin]
+def check_min_int() -> None:
+    x: int = 200
```
97littleleaf11, line 334: I think functions with prefix `check` wouldn't trigger a run test? Maybe you should change it to `test_*`. Also, please add more run tests about other types, such as strings, floats.

97littleleaf11 (changes requested):

mypyc/test-data/run-integers.test
ChetanKhanna: I'm having some issues with my system right now. I'll try to update the branch by the weekend.

ChetanKhanna:
If this is good, then I can the `max` counterpart as well.

mypyc/test-data/run-floats.test
```diff
@@ -20,3 +20,9 @@ def test_abs() -> None:
     assert abs(44324.732) == 44324.732
     assert abs(-23.4) == 23.4
     assert abs(-43.44e-4) == 43.44e-4
+
+[case testFloatMin]
+def test_float_min() -> None:
```
97littleleaf11, line 25: I think merging these test cases into one might be better.

mypyc/test-data/fixtures/ir.py
```diff
@@ -289,6 +289,12 @@ def zip(x: Iterable[T], y: Iterable[S], z: Iterable[V]) -> Iterator[Tuple[T, S,
 def eval(e: str) -> Any: ...
 def abs(x: float) -> float: ...
 def exit() -> None: ...
+@overload
+def min(x: int, y: int) -> int: ...
+@overload
+def min(x: float, y: float) -> float: ...
+@overload
+def min(x: str, y: str) -> str: ...
```
97littleleaf11, line 297: You can use `T` (already defined in this file) to make them simple.

mypyc/test-data/run-strings.test
```diff
@@ -617,3 +617,9 @@ def test_encode() -> None:
         assert False
     except UnicodeEncodeError:
         pass
+
+[case testStringMin]
+def test_str_min() -> None:
+    x: str = 'aaa'
+    y: str = 'bbb'
+    assert min(x, y) == 'aaa'
```
97littleleaf11, line 625: you can also test the minimun of user defined class which has `__lt__`.

97littleleaf11 (changes requested):
Thanks for contributing! I have several suggesions on testing. The specializer looks good to me and you could support `max` in this PR.

btw, we don't recommend force-push since it would break previous reviews.

mypyc/test-data/run-floats.test
ChetanKhanna: Done

mypyc/test-data/fixtures/ir.py
ChetanKhanna: Added

mypyc/test-data/run-strings.test
ChetanKhanna: Added

JukkaL:
Please also test heterogeneous operands for `min` (e.g. `Any` and `int`; also `int` and `Any`). The `int` operands needs to be coerced to `object`. Another interesting case is `int` and `bool`.

Finally, test `Any` and `Any` with a few different types, and test `Any` and `Any` with incompatible types (e.g. `int` and `str`).

ChetanKhanna:
Sorry for the late update. I tried all the tests as @JukkaL suggested, they all failed to compile. I get an error like:
`error: assignment to ‘CPyTagged’ {aka ‘long unsigned int’} from ‘PyObject *’ {aka ‘struct _object *’} makes integer from pointer without a cast`

Pushing the commit so that everyone can see as well.

> The `int` operands needs to be coerced to `object`.

Do we need to explicitly perform coercion in the new `faster_min` function?

ChetanKhanna:
I made changes to `faster_min`, the tests are running fine now.

mypyc/test-data/run-floats.test
```diff
@@ -13,10 +13,15 @@ assert str_to_float("44324") == 44324.0
...
     assert abs(0.0) == 0.0
     assert abs(-1.234567) == 1.234567
     assert abs(44324.732) == 44324.732
     assert abs(-23.4) == 23.4
     assert abs(-43.44e-4) == 43.44e-4
+
+def test_float_min() -> None:
+    x: float = 20.0
+    y: float = 30.0
+    assert min(x, y) == 20.0
```
JukkaL, line 27: Test also a case where the first operand is greater.

mypyc/test-data/run-strings.test
JukkaL: Test the inverse here as well.

mypyc/test-data/run-strings.test
```diff
@@ -626,3 +626,21 @@ def test_encode() -> None:
...
+    def __init__(self, val: str) -> None:
+        self.val = val
+    
+    def __lt__(self, x: 'SomeItem') -> bool:
+        return self.val < x.val
+
+def test_dunder_min() -> None:
+    x = SomeItem('aaa')
+    y = SomeItem('bbb')
+    assert min(x, y).val == 'aaa'
```
JukkaL, line 646: Similarly, test the inverse as well.

JukkaL (commented):
Left a few comments (not a full review).

mypyc/irbuild/specialize.py
```diff
@@ -185,6 +185,33 @@ def translate_set_from_generator_call(
     return None
 
 
+@specialize_function('builtins.min')
+def faster_min(builder: IRBuilder, expr: CallExpr, callee: RefExpr) -> Optional[Value]:
+    if (len(expr.args) > 0
```
JukkaL, line 190: I think that this check is redundant (and not sufficient, since one argument is not enough), since if `arg_kinds` has two positional items, there will be exactly two arguments.

mypyc/irbuild/specialize.py
```diff
@@ -185,6 +185,33 @@ def translate_set_from_generator_call(
...
+            and expr.arg_kinds == [ARG_POS, ARG_POS]):
+        x, y = builder.accept(expr.args[0]), builder.accept(expr.args[1])
+        result = Register(builder.node_type(expr))
+        comparison = builder.binary_op(x, y, '<', expr.line)
```
JukkaL, line 194: Based on a quick experiment, it seems that CPython does actually evaluate `y < x` instead of `x < y` when doing `min(x, y)`. Can you double check this and update the implementation accordingly if that's the case? (This is not a big deal, but it's better to remain as close to CPython as possible.)

JukkaL (approved):
Looks good! I checked that `min` is now much faster than previously -- it can be over 10x faster than before.

I have just a two minor comments. Feel free to merge once you've addressed them.

Extending this to support `max` should be pretty easy and a good follow-up PR :-)

mypyc/irbuild/specialize.py
97littleleaf11: Thanks for pointing out! I checked the `bltinmodule/min_max()` in CPython and it does eval `y < x` when doing `min(x, y)`. To be more specific, it evaluates arguments from last to first.

PR #10267, opened 2021-03-30
add stubs for missing Logger attributes

dod-wdwilcox:
### Description

Fixes #43673
https://bugs.python.org/issue43673

Add type stubs for two undocumented attributes, which nonetheless exist and I need.
https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L1925

## Test Plan
It's not clear to me how/if I need to add this to the dynamic, introspective test setup going on.

### DIFF
--- a/mypy/typeshed/stdlib/logging/__init__.pyi
+++ b/mypy/typeshed/stdlib/logging/__init__.pyi
@@ -901,6 +901,12 @@ class RootLogger(Logger):
 
 root: RootLogger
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra:
You'll want to make this PR in the github.com/python/typeshed repo; mypy just uses a copy of the code.

Also, I don't think this is syntactically valid. You'll have to add new fields inside the Logger class instead.

PR #10269, opened 2021-03-31
Optimize truth value testing for strings

pranavrajpal:
This causes code that tests for the boolean value of a string to test the length directly instead of going through PyObject_IsTrue, as mentioned in mypyc/mypyc#768.

This should make code that uses this faster, but my benchmarks seem to show that this makes code slower than it would be without this optimization. I was hoping you could take a look at my code and the benchmarks and see if I'm doing anything wrong.

## Test Plan

I added a test that compiles and runs code that uses this, as well as an irbuild test, and ran all of the mypyc tests to make sure they were still working.

## Benchmarks

This is my benchmark code:
```python
import time

def run(s: str) -> None:
    t0 = time.time()
    for i in range(1_000_000):
        if s:
            pass
    t1 = time.time()
    print(f'{t1 - t0:.3f}')

run('')
run('A' * 100)
```

My results from running this benchmark several times each:

Interpreted version without mypyc: 0.022 to 0.024 seconds
Compiled version, from master (without these changes): 0.006 to 0.007 seconds
Compiled version, with these changes: 0.008 to 0.009 seconds

### DIFF
--- a/mypyc/test-data/irbuild-basic.test
+++ b/mypyc/test-data/irbuild-basic.test
@@ -196,24 +196,22 @@ def f(x: object, y: object) -> str:
 def f(x, y):
     x, y :: object
     r0 :: str
-    r1 :: int32
+    r1 :: int
     r2 :: bit
-    r3 :: bool
-    r4, r5 :: str
+    r3, r4 :: str
 L0:
     r0 = PyObject_Str(x)
-    r1 = PyObject_IsTrue(r0)
-    r2 = r1 >= 0 :: signed
-    r3 = truncate r1: int32 to builtins.bool
-    if r3 goto L1 else goto L2 :: bool
+    r1 = CPyObject_Size(r0)
+    r2 = r1 != 0
+    if r2 goto L1 else goto L2 :: bool
 L1:
-    r4 = r0
+    r3 = r0
     goto L3
 L2:
-    r5 = PyObject_Str(y)
-    r4 = r5
+    r4 = PyObject_Str(y)
+    r3 = r4
 L3:
-    return r4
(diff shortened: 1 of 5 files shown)


mypyc/test-data/irbuild-basic.test
```diff
@@ -196,24 +196,22 @@ def f(x: object, y: object) -> str:
...
-    r3 :: bool
-    r4, r5 :: str
+    r3, r4 :: str
 L0:
     r0 = PyObject_Str(x)
-    r1 = PyObject_IsTrue(r0)
-    r2 = r1 >= 0 :: signed
-    r3 = truncate r1: int32 to builtins.bool
-    if r3 goto L1 else goto L2 :: bool
+    r1 = CPyObject_Size(r0)
```
JukkaL, line 204: This is probably the bottleneck. This may be about as fast/slow as `PyObject_IsTrue`, so you aren't seeing an improvement. You'd need to specialize this operation for `str` values, and access the size of the string directly using C struct access. There are examples of how to do this for `list` and `tuple` values. Strings should be fairly similar (details may be different).

pranavrajpal:
Ok, I have moved the check for whether a string's length is 0 to a primitive, and this appears to have made it faster.

Using the previously mentioned benchmark, I got these times:
Interpreted: 0.022 to 0.024 seconds
Compiled version, without these changes: 0.006 to 0.007 seconds
Compiled version, with these changes: 0.003 to 0.004 seconds

That appears to show that these changes make that benchmark approximately twice as fast.

TH3CHARLie (approved):
LGTM, good job!

PR #10284, opened 2021-04-06
Use type(x) == T for type narrowing

### LINKED ISSUE #4445: "if type(x) == T" not being used for type narrowing
Given this code:

```
import random

x = random.randint(1, 4)
if x == 0:
    y = None
elif x == 1:
    y = MyType()
elif x == 2:
    y = MyType2()
elif x == 3:
    y = MyType3()
else:
    assert x == 4
    y = None

if isinstance(y, MyType2):
    reveal_type(y)
    y.b()
```

Then as expected, mypy gives this output:

```
error: Revealed type is 'MyType2'
```

However, if we change this line:

```
if isinstance(y, MyType2):
```

Over to this line:

```
if type(y) == MyType2:
```

Then mypy gives this output:

```
error: Revealed type is 'Union[MyType, builtins.None]'
error: Item "MyType" of "Optional[MyType]" has no attribute "b"
error: Item "None" of "Optional[MyType]" has no attribute "b"
```

But I'd expect that in this case to get the same output as when using isinstance

As a workaround, I currently need to do this:

```
if type(y) == MyType2:
    assert isinstance(y, MyType2)
    reveal_type(y)
    y.b()
```

Which produces the expected mypy output:

```
error: Revealed type is 'MyType2'
```

I shouldn't need to use that workaround.

pranavrajpal:
Fixes #4445

This makes mypy use expressions like `type(x) == T` during type narrowing similar to how it already does for `isinstance(x, T)`.

This also adds support for using `type(x) is T`, as well as checking the types of multiple variables at the same time, like `type(x) == type(y) == T`.

This doesn't do any type narrowing when multiple literal types are present in the comparison, such as in `type(x) == int == str` or `type(x) == int == int`. Expressions of that kind should always be False if those literal types aren't all the same, and are relatively pointless if all of the literal types are the same, so I think it's fine not supporting those. I can try adding support for the case when those literal types are the same if necessary.

## Test Plan

I added several tests to make sure that mypy was actually using checks of this kind to narrow types inside if statements.

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3993,10 +3993,15 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                     # type of x from that check
                     return {}, {}
                 else:
-                    # FIXME: crashes if no fullname attribute (check if NameExpr?)
-                    if isinstance(expr, RefExpr) and expr.fullname is not None:
-                        info = self.lookup_typeinfo(expr.fullname)
-                        is_final = info.is_final
+                    try:
+                        if isinstance(expr, RefExpr) and expr.fullname is not None:
+                            info = self.lookup_typeinfo(expr.fullname)
+                            is_final = info.is_final
+                    # lookup_typeinfo sometimes fails with a KeyError
+                    except KeyError:
+                        # assume the expression isn't final unless we can confirm
+                        # otherwise
+                        is_final = False
                     type_being_compared = current_type
 
         if not exprs_in_type_calls:
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -4128,9 +4128,9 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                     # explicit type(x) == some_type check
                     if if_map == {} and else_map == {}:
                         # exprs that are being passed into type
-                        exprs_in_type_calls = [] # type: List[Expression]
+                        exprs_in_type_calls = []  # type: List[Expression]
                         # type that is being compared to type(expr)
-                        type_being_compared = None # type: Optional[List[TypeRange]]
+                        type_being_compared = None  # type: Optional[List[TypeRange]]
 
                         for index in expr_indices:
                             expr = node.operands[index]
@@ -4142,16 +4142,17 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                                 if current_type is None:
                                     continue
                                 if type_being_compared is not None:
-                                    # It doesn't really make sense to have several types being compared 
-                                    # to the output of type (like type(x) == int == str) because whether
-                                    # that's true is solely dependent on what the types being compared are,
-                                    # so we don't try to narrow types any further because we can't really get
-                                    # any information about the type of x from that check
+                                    # It doesn't really make sense to have several types being
+                                    # compared to the output of type (like type(x) == int == str)
+                                    # because whether that's true is solely dependent on what the
+                                    # types being compared are, so we don't try to narrow types any
+                                    # further because we can't really get any information about the
+                                    # type of x from that check
                                     break
                                 else:
                                     type_being_compared = current_type
 
-                        if_maps = [] # type: List[TypeMap]
+                        if_maps = []  # type: List[TypeMap]
                         for expr in exprs_in_type_calls:
                             current_if_map, _ = self.conditional_type_map_with_intersection(
                                 expr,
@@ -4159,6 +4160,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                                 type_being_compared
                             )
                             if_maps.append(current_if_map)
+
                         def combine_maps(list_maps: List[TypeMap]) -> TypeMap:
                             """Combine all typemaps in list_maps into one typemap"""
                             result_map = {}
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -3954,9 +3954,14 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
 
         return None, {}
 
-    def find_type_equals_check(self, node: Expression, expr_indices: List[int]
+    def find_type_equals_check(self, node: ComparisonExpr, expr_indices: List[int]
                                ) -> Tuple[TypeMap, TypeMap]:
-        """Narrow types based on any checks of the type ``type(x) == T``"""
+        """Narrow types based on any checks of the type ``type(x) == T``
+
+        :param node: The node that might contain the comparison
+
+        :param expr_indices: The list of indices of expressions in ``node`` that are being compared
+        """
         type_map = self.type_map
 
         def is_type_call(expr: CallExpr) -> bool:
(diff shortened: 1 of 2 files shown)


test-data/unit/check-isinstance.test
```diff
@@ -2573,3 +2573,27 @@ if issubclass(x, B):
...
+if type(x) == type(y) == int:
+    reveal_type(y) # N: Revealed type is "builtins.int"
+    reveal_type(x) # N: Revealed type is "builtins.int"
+
+[case testTypeEqualsCheckUsingIs]
+from typing import Any
+
+y: Any
+if type(y) is int:
+    reveal_type(y) # N: Revealed type is "builtins.int"
```
JukkaL, line 2599: Add also test case where you narrow from `Union[int, str]` to `int`, and there is an else block. For example, something like this:

```py
def f(x: Union[int, str]) -> None:
    if type(x) is int:
        reveal_type(x)  # int
    else:
        reveal_type(x)  # Union[int, str]
```

The else block should use the original type, since instances of `int` subclasses (e.g. `bool`) will be handled in the else block.

JukkaL (commented):
Thanks for the PR! Please have a look my comment about dealing with 'else' blocks and union types. I'll do a full review once that use case works and is tested.

test-data/unit/check-isinstance.test
pranavrajpal: Ok, I have added a test case for this and modified the type checker to handle this. Could you take a look at my changes when you get the chance?

mypy/checker.py
```diff
@@ -4118,6 +4118,61 @@ def has_no_custom_eq_checks(t: Type) -> bool:
...
+                                    # compared to the output of type (like type(x) == int == str)
+                                    # because whether that's true is solely dependent on what the
+                                    # types being compared are, so we don't try to narrow types any
+                                    # further because we can't really get any information about the
+                                    # type of x from that check
+                                    break
+                                else:
+                                    type_being_compared = current_type
+
+                        if_maps = []  # type: List[TypeMap]
```
JukkaL, line 4155: Can you move the remaining operations behind `if exprs_in_type_calls:`, as an optimization? This way the remaining code will only be run if there is something to do. It also makes it easier to see that the code does nothing unless there is a `type(x)` somewhere.

test-data/unit/check-isinstance.test
```diff
@@ -2573,3 +2573,36 @@ if issubclass(x, B):
...
+
+[case testTypeEqualsNarrowingUnionWithElse]
+from typing import Union
+
+x: Union[int, str]
+if type(x) is int:
+    reveal_type(x)  # N: Revealed type is "builtins.int"
+else:
+    reveal_type(x)  # N: Revealed type is "Union[builtins.int, builtins.str]"
```
JukkaL, line 2608: Test `type(x) is not int` with `Union[int, str]` (also `!=`). It shouldn't do anything, since the type could be a subclass of `int`.

Optionally, if we are dealing with a `@final` class (`TypeInfo` with `is_final` true), we can actually restrict the type away in the else block. If the type of `x` is `Union[C, D]`, and `C` is final, the type of `x` in the else block could be just `D`, since there can be no subclasses of `C`:

```py
x: Union[C, D]
if type(x) is C:
    reveal_type(x)  # C
else:
    reveal_type(x)  # D, but only if C is final
```

It's fine to leave the handling of final classes as a follow-up issue, though. The current PR already improves things.

JukkaL (commented):
Thanks for the updates! This adds a useful new feature that removes some false positives, which is always nice. Looks good, just a few comments.

test-data/unit/check-isinstance.test
pranavrajpal: When you say that `type(x) is not int` shouldn't do anything, do you mean that it shouldn't do any narrowing, like this:
```python
x: Union[int, str]
if type(x) is not int:
    reveal_type(x)  # Union[int, str]
else:
    reveal_type(x)  # Union[int, str]
```
Or do you mean that it should flip what it infers for the if and else cases:
```python
x: Union[int, str]
if type(x) is not int:
    reveal_type(x)  # Union[int, str]
else:
    reveal_type(x)  # int
```

I think it would be the second one because I think the only way for `type(x) is not int` to be false is for x to be an `int`.

JukkaL: Yeah, the second is right, I was confused!

pranavrajpal:
I've added the tests for `type(x) is not int` and `type(x) != int` and I made the type narrowing stop early if it didn't find any `type(x)` calls.

I'm still working on handling comparing the type of a variable to a final class.

pranavrajpal:
I have added support for restricting the type of a variable in the else case if the type being compared is final.

Could you take a look and let me know what you think?

mypy/checker.py
```diff
@@ -3954,6 +3954,88 @@ def conditional_callable_type_map(self, expr: Expression,
...
+                    # further because we can't really get any information about the
+                    # type of x from that check
+                    return {}, {}
+                else:
+                    try:
+                        if isinstance(expr, RefExpr) and expr.fullname is not None:
+                            info = self.lookup_typeinfo(expr.fullname)
+                            is_final = info.is_final
+                    # lookup_typeinfo sometimes fails with a KeyError
+                    except Exception:
```
JelleZijlstra, line 4001: Catching any Exception isn't great because it can easily mask a bug. If lookup_typeinfo raises an unexpected exception, it might be better to fix that inside the method.

pranavrajpal: I made the changes that you mentioned by not catching all Exceptions. Could you take a look at my changes?

JelleZijlstra (approved):
Looks good to me but I'd want @JukkaL to do the final review.

hauntsaninja:
Not at all blocking, but if you want something related to work on, I think this code still doesn't work after this PR: https://github.com/python/mypy/issues/10302#issuecomment-816851557

mypy/checker.py
```diff
@@ -3954,6 +3954,81 @@ def conditional_callable_type_map(self, expr: Expression,
 
         return None, {}
 
+    def find_type_equals_check(self, node: ComparisonExpr, expr_indices: List[int]
+                               ) -> Tuple[TypeMap, TypeMap]:
+        """Narrow types based on any checks of the type ``type(x) == T``
+
+        :param node: The node that might contain the comparison
+
+        :param expr_indices: The list of indices of expressions in ``node`` that are being compared
```
JukkaL, line 3963: Style nit: We generally use this style for documenting arguments:

```
Args:
    node: The node that might ...
    expr_indiced: The list of indices ...
```

This is a minor thing, however, and this doesn't block merging this PR.

JukkaL (approved):
Thanks for the updates and the refactoring! Looks great.

mypy/checker.py
pranavrajpal: I made this change in #10378 because I couldn't figure out how to add more commits to this branch after it's already been merged.

PR #10286, opened 2021-04-06
[mypyc] Implement dict setdefault primitive

97littleleaf11:
### Description

Related issue: https://github.com/mypyc/mypyc/issues/644

* Add a new primitive for `Dict.setdefault()`.
* Move some dict creation primitives to beginning of `dict_ops.py` for better code structure.

## Test Plan

Please refer to the test-data changes

### DIFF
--- a/mypyc/test-data/run-dicts.test
+++ b/mypyc/test-data/run-dicts.test
@@ -217,3 +217,10 @@ def test_dict_copy() -> None:
     dd['a'] = 1
     assert dd.copy() == dd
     assert isinstance(dd.copy(), defaultdict)
+
+def test_dict_setdefault() -> None:
+    d: Dict[str, int] = {'a': 1, 'b': 2}
+    assert d.setdefault('a', 2) == 1
+    assert d.setdefault('c', 3) == 3
+    assert d['a'] == 1
+    assert d['c'] == 3
--- a/mypyc/primitives/dict_ops.py
+++ b/mypyc/primitives/dict_ops.py
@@ -17,6 +17,40 @@ load_address_op(
     type=object_rprimitive,
     src='PyDict_Type')
 
+# Construct an empty dictionary.
+dict_new_op = custom_op(
+    arg_types=[],
+    return_type=dict_rprimitive,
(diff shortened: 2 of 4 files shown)


97littleleaf11:
Failed in mypyc compiled version.

TH3CHARLie:
The compiled mypy fails usually because the newly added primitive breaks sth. Try to comment out the new op and see if the error disappears. If that's the case, you need to double-check if your implementation matches cpython

97littleleaf11:
@TH3CHARLie thanks! I just found that the second parameter of `setdefault`is optional.

TH3CHARLie:
Actually for `Py_None`, you can find `none_object` in `ll_builder` to represent that, which means you can build the IR itself without the need for a C wrapper function. But in that way, the two variants of the same op are implemented differently. A better solution would be somehow unify the `load_address_op` into the op registration.

mypyc/test-data/run-dicts.test
```diff
@@ -217,3 +217,13 @@ def test_dict_copy() -> None:
...
+
+def test_dict_setdefault() -> None:
+    d: Dict[str, int] = {'a': 1, 'b': 2}
+    assert d.setdefault('a', 2) == 1
+    assert d.setdefault('c', 3) == 3
+    assert d['a'] == 1
+    assert d['c'] == 3
+    assert d.setdefault('a') == 1
+    assert d.setdefault('e') == None
+    assert d.setdefault('e', 100) == None
```
JukkaL, line 229: Also test a subclass of `dict` that overrides `setdefault`. We should use the overridden method (example where this probably matters: `OrderedDict`). You can create a `dict` subclass that overrides `setdefault` that behaves differently and use that in the test. If this doesn't work correctly right now, you'll probably need to add a helper similar to `CPyDict_KeysView` that has a generic fallback implementation for `dict` subclasses.

JukkaL (approved):
Thanks for the updates, looks good now! This made a microbenchmark I came up with over 2x faster, which is a great result.

PR #10288, opened 2021-04-06
Consider reversed operands order when comparing version info

### LINKED ISSUE #10264: Python version check is sensitive to comparison order
**Bug Report**

Whether mypy successfully detects a Python version check is dependent on the order of the comparison.

**To Reproduce**

```python
"""This will not pass on Python 3.8."""
import sys

if (3, 8) <= sys.version_info:
    pass
else:
    import invalid_library

```

```sh
% mypy type.py --python-version 3.8
type.py:6: error: Cannot find implementation or library stub for module named 'invalid_library'
type.py:6: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```

```python
"""This will pass on Python 3.8."""
import sys

if sys.version_info >= (3, 8):
    pass
else:
    import invalid_library
```

```sh
% mypy type.py --python-version 3.8
```

**Expected Behavior**

I expected mypy to correctly parse the Python version check regardless of the order of the comparison.

**Actual Behavior**

`(3, 8) <= sys.version_info` is not correctly detected by mypy, while the logically equivalent `sys.version_info >= (3, 8)` is.

**Your Environment**

- Mypy version used: 0.812
- Mypy command-line flags: `--python-version 3.8`
- Mypy configuration options from `mypy.ini` (and other config files): Empty
- Python version used: 3.8.6

kamilturek:
### Description

Closes #10264.

Consider reversed order of operands when trying to compare version info. When reversed order is used, operator is reversed as well for correct comparison.

### DIFF
--- a/mypy/reachability.py
+++ b/mypy/reachability.py
@@ -27,6 +27,14 @@ inverted_truth_mapping = {
     MYPY_FALSE: MYPY_TRUE,
 }  # type: Final
 
+reverse_op = {"==": "==",
+              "!=": "!=",
+              "<":  ">",
+              ">":  "<",
+              "<=": ">=",
+              ">=": "<=",
+              }  # type: Final
+
 
 def infer_reachability_of_if_statement(s: IfStmt, options: Options) -> None:
     for i in range(len(s.expr)):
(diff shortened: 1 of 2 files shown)


JelleZijlstra (approved):

mypy/reachability.py
```diff
@@ -127,10 +135,15 @@ def consider_sys_version_info(expr: Expression, pyversion: Tuple[int, ...]) -> i
...
         return TRUTH_VALUE_UNKNOWN
+
+    index = contains_sys_version_info(expr.operands[0])
     thing = contains_int_or_tuple_of_ints(expr.operands[1])
-    if thing is None:
+    if index is None or thing is None:
+        index = contains_sys_version_info(expr.operands[1])
+        thing = contains_int_or_tuple_of_ints(expr.operands[0])
+        op = reverse_op[op]
+    if index is None or thing is None:
```
JelleZijlstra, line 145: This check is redundant, since the conditions below won't trigger if either value is None.

JelleZijlstra (changes requested):
Please remove the redundant check.

mypy/reachability.py
kamilturek: That's right. Good point 👍

JelleZijlstra (approved):

kamilturek:
@JelleZijlstra One job on Travis CI has failed for some weird reason, doesn't look like thing related to this PR.

JelleZijlstra:
Yeah that looks harmless. I'll merge this PR, thanks for your contributions!

PR #10290, opened 2021-04-07
Only lookup vtable indices if needed for method call emits

rtpg:
The vtable index is only needed for when we actually go through a  vtable lookup. I don't believe this is actually a major performance improvement, but it should make it clearer what's important for each branch.

This was found while working through [this mypyc issue](https://github.com/mypyc/mypyc/issues/826)


JukkaL (approved):
Thanks! The build failure seems unrelated/

PR #10292, opened 2021-04-07
Extend the dataclass plugin to deal with callable properties

aecay:
### Description

At runtime, the callable properties of dataclasses are handled in the
way one would expect: they are not passed a `self` argument.  Mypy,
however, just sees them as callable class attributes and generates
errors about missing arguments.  This is a special case of what is
discussed in #708.  I donʼt have a general solution for that problem,
but for dataclasses, I can fix it by automatically converting the
callable entries in a data class into (settable) properties.  That makes
them work properly via-a-vis the typechecker.

## Test Plan

Added unit tests which verify:
- callable dataclass properties no longer generate mypy errors about arity mismatches
  - neither as class init arguments
  - nor via property assignment
  - nor when calling a property on an instance
- callable dataclass properties reject incompatible types
  - as class init arguments
  - and via property assignment

### DIFF
--- a/mypy/plugins/dataclasses.py
+++ b/mypy/plugins/dataclasses.py
@@ -13,7 +13,10 @@ from mypy.plugins.common import (
     add_method, _get_decorator_bool_argument, deserialize_and_fixup_type,
 )
 from mypy.typeops import map_type_from_supertype
-from mypy.types import Type, Instance, NoneType, TypeVarDef, TypeVarType, CallableType, get_proper_type
+from mypy.types import (
+    Type, Instance, NoneType, TypeVarDef, TypeVarType, CallableType,
+    get_proper_type
+)
 from mypy.server.trigger import make_wildcard_trigger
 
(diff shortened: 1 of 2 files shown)


aecay:
The travis failure is in the python3.5 unit test job, which fails before the tests even run:
```
 Could not find a version that satisfies the requirement typing (from versions: )
No matching distribution found for typing
```

This same job passed on fb07856, and I can't see where the changes in 6149640 would have caused it to start failing...  So it looks like a transient issue.  When I try to rerun the job on Travis, I get an error ("could not rerun job") -- maybe a permissions issue?  In any case, I figured it makes sense to wait for reviews and maybe a little help coaxing the CI to rerun rather than keep bashing away at it for the moment.

JelleZijlstra:
I used the powers vested in me to click the "Re-run job button" in Travis. For future reference, closing and reopening the PR tends to work too.

JukkaL (approved):
Thanks for the PR! Looks good.

PR #10293, opened 2021-04-08
[mypyc] Add a more detailed MRO explanation.

rtpg:
(This` include #10290)

This is inspired by Rust's `--explain` mechanism. Instead of having huge multi-line error outputs by default, you need to pass in `explain` (here through setting `MYPYC_EXPLAIN=1` when running `mypyc`) in order to get a more detailed explanation for less obvious error messages from the mypyc compilation itself.

This is more of a discussion piece at this stage, but I think that having detailed explanations for most of the `mypyc` errors would help a lot with debugging errors quickly in general, especially since they're more restrictive than Python's normal rues.


JelleZijlstra:
You could leverage the existing `--show-error-context` mypy flag perhaps.

JukkaL:
This is a reasonable idea. It's quite a big change to our error reporting philosophy, so we should discuss it in an issue first and brainstorm different approaches and other error conditions where this might be useful. It's probably not worth it just for this particular message, but if we can come up with a few more common error conditions where we could use this, it could be worthwhile.

We should probably also have a shorter summary message corresponding to each detailed message. In this case, the short message that we'd show by default could be something like this:

```
... error: A native class can't inherit from two non-trait native/extension classes ("Base1" and "Base2")
... note: Consider using @trait to declare some base classes as traits
... note: See https:/<some doc link>
```

PR #10295, opened 2021-04-08
[mypyc] Pre-allocate space in list comprehensions

97littleleaf11:
### Description

Closes https://github.com/mypyc/mypyc/issues/473

* Reconstruct `tuple_creation_from_generator_helper` to `preallocate_space_helper`
* Rewrite `new_tuple_with_length`. Now we need only to pass in length.
* Add new primitive for list initialization.
* Support faster list comprehension in some simple situations.

### DIFF
--- a/mypyc/irbuild/for_helpers.py
+++ b/mypyc/irbuild/for_helpers.py
@@ -16,7 +16,8 @@ from mypyc.ir.ops import (
 )
 from mypyc.ir.rtypes import (
     RType, is_short_int_rprimitive, is_list_rprimitive, is_sequence_rprimitive,
-    RTuple, is_dict_rprimitive, short_int_rprimitive, int_rprimitive
+    is_tuple_rprimitive, is_dict_rprimitive,
+    RTuple, short_int_rprimitive, int_rprimitive
 )
 from mypyc.primitives.registry import CFunctionDescription
 from mypyc.primitives.dict_ops import (
@@ -129,6 +130,30 @@ def for_loop_helper_with_index(builder: IRBuilder, index: Lvalue, expr: Expressi
     builder.pop_loop_stack()
 
     builder.activate_block(exit_block)
+
+
+def preallocate_space_helper(builder: IRBuilder,
+                             gen: GeneratorExpr,
+                             empty_op_llbuilder: Callable[[Value, int], Value],
+                             set_item_op: CFunctionDescription) -> Optional[Value]:
+    """Currently we only optimize for simplest generator expression"""
+    if len(gen.sequences) == 1 and len(gen.condlists[0]) == 0:
+        rtype = builder.node_type(gen.sequences[0])
+        if is_list_rprimitive(rtype) or is_tuple_rprimitive(rtype):
+
+            length = builder.builder.builtin_len(builder.accept(gen.sequences[0]),
+                                                 gen.line, use_pyssize_t=True)
+            target_op = empty_op_llbuilder(length, gen.line)
+
+            def set_item(item_index: Value) -> None:
+                e = builder.accept(gen.left_expr)
+                builder.call_c(set_item_op, [target_op, item_index, e], gen.line)
+
+            for_loop_helper_with_index(builder, gen.indices[0], gen.sequences[0],
+                                       set_item, gen.line)
+
+            return target_op
+    return None
 
 
 def translate_list_comprehension(builder: IRBuilder, gen: GeneratorExpr) -> Value:
--- a/mypyc/irbuild/for_helpers.py
+++ b/mypyc/irbuild/for_helpers.py
@@ -137,7 +137,9 @@ def sequence_from_generator_preallocate_helper(
         gen: GeneratorExpr,
         empty_op_llbuilder: Callable[[Value, int], Value],
         set_item_op: CFunctionDescription) -> Optional[Value]:
-    """Currently we only optimize for simplest generator expression, which means that
+    """Generate a new tuple or list from a simple generator expression.
+
+    Currently we only optimize for simplest generator expression, which means that
     there is no condition list in the generator and only one original sequence with
     one index is allowed.
 
--- a/mypyc/irbuild/for_helpers.py
+++ b/mypyc/irbuild/for_helpers.py
@@ -136,8 +136,11 @@ def preallocate_space_helper(builder: IRBuilder,
                              gen: GeneratorExpr,
                              empty_op_llbuilder: Callable[[Value, int], Value],
                              set_item_op: CFunctionDescription) -> Optional[Value]:
-    """Currently we only optimize for simplest generator expression"""
-    if len(gen.sequences) == 1 and len(gen.condlists[0]) == 0:
+    """Currently we only optimize for simplest generator expression.
+
+    "... for index in list/tuple"
+    """
+    if len(gen.sequences) == 1 and len(gen.indices) == 1 and len(gen.condlists[0]) == 0:
         rtype = builder.node_type(gen.sequences[0])
         if is_list_rprimitive(rtype) or is_tuple_rprimitive(rtype):
 
@@ -157,6 +160,7 @@ def preallocate_space_helper(builder: IRBuilder,
 
 
 def translate_list_comprehension(builder: IRBuilder, gen: GeneratorExpr) -> Value:
+    # Try simplest list comprehension, otherwise fall back to general one
     val = preallocate_space_helper(builder, gen,
                                    empty_op_llbuilder=builder.builder.new_list_op_with_length,
                                    set_item_op=list_set_item_op)
--- a/mypyc/irbuild/for_helpers.py
+++ b/mypyc/irbuild/for_helpers.py
@@ -24,7 +24,7 @@ from mypyc.primitives.dict_ops import (
     dict_next_key_op, dict_next_value_op, dict_next_item_op, dict_check_size_op,
     dict_key_iter_op, dict_value_iter_op, dict_item_iter_op
 )
-from mypyc.primitives.list_ops import list_append_op, list_get_item_unsafe_op, list_set_item_op
+from mypyc.primitives.list_ops import list_append_op, list_get_item_unsafe_op, new_list_set_item_op
 from mypyc.primitives.set_ops import set_add_op
 from mypyc.primitives.generic_ops import iter_op, next_op
 from mypyc.primitives.exc_ops import no_err_occurred_op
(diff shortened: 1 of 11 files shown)


mypyc/irbuild/specialize.py
```diff
@@ -92,6 +93,12 @@ def dict_methods_fast_path(
     if not (len(expr.args) == 1 and expr.arg_kinds == [ARG_POS]):
         return None
     arg = expr.args[0]
+    if isinstance(arg, GeneratorExpr):
+        val = preallocate_space_helper(builder, arg,
+                                       empty_op_llbuilder=builder.builder.new_list_op_with_length,
+                                       set_item_op=list_set_item_op)
+        if val is not None:
+            return val
```
97littleleaf11, line 101: Should also consider `[ ]` form, which is in `transform_list_comprehension`

97littleleaf11:
Encounter segment fault locally when i want to import module compiled by this branch

JukkaL:
Can you share the example which is failing for you? I can see if I can spot the issue.

97littleleaf11:
Here is my local test file `tmp.py`:

```py
def f(x: str) -> str:
    return x

def test() -> None:
    source = ["a", "b", "c"]
    a = list(f(x) for x in source)
    print(a)

test()
```

compile and run: 
```
python -c "import tmp"
[1]    58318 segmentation fault  python -c "import tmp"
```
If I comment line 96 - 101 of `specialize.py` it could work normally.

mypyc/irbuild/specialize.py
97littleleaf11: Not a good idea to deal with list comprehension here, since this function is specialized for dict view.

JelleZijlstra:
Did you try to find the cause in gdb?

JukkaL:
Hmm we should probably document how to debug C-level crashes on different platforms. It's pretty easy on Linux but I'm actually not sure what's the best way to do it on macOS or Windows. I'll create an issue.

97littleleaf11:
I am using `lldb` on Mac. Then I got this
```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00000001011e2c44 tmp4.cpython-37m-darwin.so`CPyDef_test at list_ops.c:77:9 [opt]
   74               }
   75           }
   76           // PyList_SET_ITEM doesn't decref the old element, so we do
-> 77           Py_DECREF(PyList_GET_ITEM(list, n));
   78           // N.B: Steals reference
   79           PyList_SET_ITEM(list, n, value);
   80           return true;
```

Since it's newly created list, there is no element in list. Thus `Py_DECREF` failed. 
Maybe we could change this to `Py_XDECREF`?

JelleZijlstra:
I was wondering whether this could lead to reading from uninitialized memory, but https://docs.python.org/3/c-api/list.html#c.PyList_New says the list is initialized to NULL.

Changing this code to XDECREF would incur a small performance hit for the additional check. If that matters, you could write separate setitem functions for existing lists (can use DECREF) and new lists (don't need to do anything).

97littleleaf11:
@JelleZijlstra Thanks for your advice! I will add a new primitive later.

JelleZijlstra:
To be clear I'm not sure the performance hit is big enough to be worth the additional complexity. You could try some benchmarks, or maybe someone else reading here has a stronger intuition.

If you created a new list and you're filling it up, you shouldn't need to do bounds checking.

mypyc/irbuild/ll_builder.py
```diff
@@ -891,6 +891,9 @@ def make_dict(self, key_value_pairs: Sequence[DictEntry], line: int) -> Value:
 
         return result
 
+    def new_list_op_with_length(self, length: Value, line: int) -> Value:
+        return self.call_c(new_list_op, [length], line)
```
JukkaL, line 895: Add docstring. Document that the new list is uninitialized and needs to be initialized with `CPyList_SetItemUnsafe` ops. Also document the expected rtype of `length`.

mypyc/irbuild/for_helpers.py
```diff
@@ -131,7 +132,41 @@ def for_loop_helper_with_index(builder: IRBuilder, index: Lvalue, expr: Expressi
     builder.activate_block(exit_block)
 
 
+def preallocate_space_helper(builder: IRBuilder,
```
JukkaL, line 135: Let's try to make the name more descriptive. Maybe `sequence_from_generator_preallocate_helper`?

mypyc/irbuild/for_helpers.py
```diff
@@ -131,7 +132,41 @@ def for_loop_helper_with_index(builder: IRBuilder, index: Lvalue, expr: Expressi
...
+                             gen: GeneratorExpr,
+                             empty_op_llbuilder: Callable[[Value, int], Value],
+                             set_item_op: CFunctionDescription) -> Optional[Value]:
+    """Currently we only optimize for simplest generator expression.
+
+    "... for index in list/tuple"
+    """
```
JukkaL, line 142: Document the arguments, in particular `empty_op_llbuilder` and `set_item_op`.

mypyc/irbuild/for_helpers.py
```diff
@@ -131,7 +132,41 @@ def for_loop_helper_with_index(builder: IRBuilder, index: Lvalue, expr: Expressi
     builder.activate_block(exit_block)
 
 
+def preallocate_space_helper(builder: IRBuilder,
+                             gen: GeneratorExpr,
+                             empty_op_llbuilder: Callable[[Value, int], Value],
+                             set_item_op: CFunctionDescription) -> Optional[Value]:
+    """Currently we only optimize for simplest generator expression.
+
```
JukkaL, line 140: Add an example of what sort things this can be used to handle (e.g. `tuple(x + 1 for x in a)` if `a` is a list).

mypyc/lib-rt/list_ops.c
```diff
@@ -79,7 +79,19 @@ bool CPyList_SetItem(PyObject *list, CPyTagged index, PyObject *value) {
         PyList_SET_ITEM(list, n, value);
         return true;
     } else {
-        PyErr_SetString(PyExc_IndexError, "list assignment index out of range");
+        PyErr_SetString(PyExc_OverflowError, "Python int too large to convert to C ssize_t");
```
JukkaL, line 82: Brainstorming: "int too large for list index" could be a slightly better message? We generally want to avoid C concepts in error messages.

mypyc/lib-rt/list_ops.c
```diff
@@ -79,7 +79,19 @@ bool CPyList_SetItem(PyObject *list, CPyTagged index, PyObject *value) {
...
+}
+
+// This function should only be used to fill in brand new lists.
+bool CPyList_SetItemUnsafe(PyObject *list, CPyTagged index, PyObject *value) {
+    if (CPyTagged_CheckShort(index)) {
+        Py_ssize_t n = CPyTagged_ShortAsSsize_t(index);
+        PyList_SET_ITEM(list, n, value);
+        return true;
+    } else {
+        PyErr_SetString(PyExc_OverflowError, "Python int too large to convert to C ssize_t");
```
JukkaL, line 94: Similar to above.

mypyc/primitives/list_ops.py
```diff
@@ -65,6 +65,15 @@
     error_kind=ERR_FALSE,
     steals=[False, False, True])
 
+# PyList_SET_ITEM does no error checking,
+# and should only be used to fill in brand new tuples.
```
JukkaL, line 69: tuples -> lists

JukkaL (commented):
I tried this with a microbenchmark that does lots of list comprehensions it was 50% faster, which is great!

Looks good overall. Left various minor comments.

mypyc/lib-rt/tuple_ops.c
```diff
@@ -55,7 +55,7 @@ bool CPySequenceTuple_SetItemUnsafe(PyObject *tuple, CPyTagged index, PyObject *
         PyTuple_SET_ITEM(tuple, n, value);
         return true;
     } else {
-        PyErr_SetString(PyExc_OverflowError, "Python int too large to convert to C ssize_t");
+        PyErr_SetString(PyExc_OverflowError, "set_item index too large for tuple");
```
JukkaL, line 58: Hmm since the original error message was duplicated in multiple places, a better approach would be to use a `#define` for the error message. 

That being said, it looks like the original message was copied verbatim from CPython, so I now think that it would better to preserve the old message (here and elsewhere). Some existing test cases could test the exception message, and even if the new message is more informative, we could still break some existing tests, which we typically want to avoid.

If you add a `#define` and then add a comment saying that the message is copied from CPython and shouldn't be changed, things will be clearer in the future.

mypyc/irbuild/ll_builder.py
```diff
@@ -1124,14 +1134,13 @@ def new_tuple(self, items: List[Value], line: int) -> Value:
         size = Integer(len(items), c_pyssize_t_rprimitive)  # type: Value
         return self.call_c(new_tuple_op, [size] + items, line)
 
-    def new_tuple_with_length(self, val: Value, line: int) -> Value:
-        """Generate a new empty tuple with length from a list or tuple
+    def new_tuple_with_length(self, length: Value, line: int) -> Value:
+        """Generate a new empty tuple with length
```
JukkaL, line 1138: Document here that the object must be initialized afterwards before it can be visible to user code.

mypyc/test-data/run-lists.test
```diff
@@ -125,7 +125,7 @@ def test_insert() -> None:
             l.insert(long_int, 5)
         except Exception as e:
             assert type(e).__name__ == 'OverflowError'
-            assert str(e) == 'Python int too large to convert to C ssize_t'
+            assert str(e) == 'insert index too large for list'
```
JukkaL, line 128: As discussed above, after thinking about this more carefully, it looks better to preserve the old message. Add comment here saying that the message is used by CPython.

mypyc/test-data/run-lists.test
```diff
@@ -338,3 +338,16 @@ assert list_in_mixed(0.0)
 assert not list_in_mixed([1])
 assert not list_in_mixed(object)
 assert list_in_mixed(type)
+
+[case testListBuiltFromGenerator]
+def test() -> None:
+    source = ["a", "b", "c"]
+    a = list(x + "f2" for x in source)
+    b = [x + "f2" for x in source]
```
JukkaL, line 346: Make all these lists distinct (e.g. replace `f2` with `f3` here), for a bit more testing coverage.

JukkaL (commented):
Thanks for the updates! Some more minor comments, but this should be almost ready to merge.

JukkaL:
> Could you please give me more hints about how to solve it?

Add a `builtins.list` specialization in `mypyc.irbuild.specialize`, perhaps? Or, more specifically, add a new case to the existing specialization where there is only one argument that is a generator, similar to what we do for `builtins.tuple` and others.)

JukkaL (approved):
Looks good now! I like how this gives a significant speedup to list comprehensions, which are very common operations.

PR #10296, opened 2021-04-08
Use double quotes errors in 'dont mix' and 'has no attributes'

dixith:
### Description
This pull request fixes issue reported in #7445  for `dont mix` and `has no attributes` error messages

## Test Plan

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -212,7 +212,7 @@ def _determine_eq_order(ctx: 'mypy.plugin.ClassDefContext') -> bool:
     order = _get_decorator_optional_bool_argument(ctx, 'order')
 
     if cmp is not None and any((eq is not None, order is not None)):
-        ctx.api.fail('Don\'t mix "%s" with "%s" and "%s"' % (cmp, eq, order), ctx.reason)
+        ctx.api.fail('Don\'t mix "cmp" with "eq" and "order"', ctx.reason)
 
     # cmp takes precedence due to bw-compatibility.
     if cmp is not None:
@@ -226,7 +226,7 @@ def _determine_eq_order(ctx: 'mypy.plugin.ClassDefContext') -> bool:
         order = eq
 
     if eq is False and order is True:
-        ctx.api.fail('eq must be "True" if order is "True"', ctx.reason)
+        ctx.api.fail('eq must be True if order is True', ctx.reason)
 
     return order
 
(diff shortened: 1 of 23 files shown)


dixith:
@JelleZijlstra 
```
# E: eq must be "True" if order is "True"
```

Please let me know if you feel double quotes are not needed for the above error message

mypy/plugins/attrs.py
```diff
@@ -212,7 +212,7 @@ def _determine_eq_order(ctx: 'mypy.plugin.ClassDefContext') -> bool:
     order = _get_decorator_optional_bool_argument(ctx, 'order')
 
     if cmp is not None and any((eq is not None, order is not None)):
-        ctx.api.fail("Don't mix `cmp` with `eq' and `order`", ctx.reason)
+        ctx.api.fail('Don\'t mix "%s" with "%s" and "%s"' % (cmp, eq, order), ctx.reason)
```
hauntsaninja, line 215: This change makes the error message worse. We shouldn't be interpolating the values for these arguments here.
```suggestion
        ctx.api.fail('Don\'t mix "cmp" with "eq" and "order"', ctx.reason)
```

mypy/plugins/attrs.py
```diff
@@ -226,7 +226,7 @@ def _determine_eq_order(ctx: 'mypy.plugin.ClassDefContext') -> bool:
         order = eq
 
     if eq is False and order is True:
-        ctx.api.fail("eq must be True if order is True", ctx.reason)
+        ctx.api.fail('eq must be "True" if order is "True"', ctx.reason)
```
hauntsaninja, line 229: No need for quotes here, I think just plain True reads more clearly as a Python literal

mypy/plugins/attrs.py
dixith: @hauntsaninja Below code looks good? 
```python
ctx.api.fail('Don\'t mix "{}" with "{}" and "{}"'.format(cmp, eq, order), ctx.reason)
```

dixith: or should i change it to 
```python
ctx.api.fail('Don\'t mix {} with {} and {}'.format(cmp, eq, order), ctx.reason)
```

mypy/plugins/attrs.py
```diff
@@ -226,7 +226,7 @@ def _determine_eq_order(ctx: 'mypy.plugin.ClassDefContext') -> bool:
         order = eq
 
     if eq is False and order is True:
-        ctx.api.fail("eq must be True if order is True", ctx.reason)
+        ctx.api.fail('eq must be True if order is True', ctx.reason)
```
JelleZijlstra, line 229: You could undo this quote change but I don't feel strongly

JelleZijlstra (approved):

dixith:
@hauntsaninja Can you please review and approve the PR? JelleZijlstra has approved.

JelleZijlstra:
(Sorry, I forgot to click merge. I don't think Shantanu needs to review too; these changes are pretty simple.

PR #10300, opened 2021-04-08
Issue an error when overriding typeguard with non-typeguard in subclass

### LINKED ISSUE #9928: PEP 647: Error if overriding typeguard with non-typeguard in subclass
See https://github.com/python/mypy/pull/9865. (**UPDATED**)

I believe we should issue an error in this case.

See (disabled) test: https://github.com/python/mypy/blob/fffbe88fc54807c8b10ac40456522ad2faf8d350/test-data/unit/check-typeguard.test#L290-L296

kamilturek:
### Description

Closes #9928.

Issue an error about incompatible signature when trying to override typeguard method with non-typeguard one.


JukkaL (approved):
Thanks! Looks good.

PR #10305, opened 2021-04-10
[WIP] Basic Arithmetic operation for floater point

pranshugrover:
I am trying to add arithmetic operations for floater point. I have added basic addition but I am uncertain if I am heading in the right direction. My main concern is the C helper for float since we do not have a representation for float like we do for int.
Please review and guide me if and where I am going wrong so that I can add other arithmetic operations.


mypyc/lib-rt/float_ops.c
```diff
@@ -0,0 +1,17 @@
+// Floater point primitive operations
+
+#include <Python.h>
+
+#include "CPy.h"
+
+
+PyObject PyFloat_Add(float left, float right ){
```
JelleZijlstra, line 8: This function doesn't make much sense: are the arguments floats or PyFloatObjects? The types say one thing, but PyFloat_Check doesn't work on floats, only on PyObjects.

Also, the return type should if anything be a `PyObject *`, not a PyObject.

pranshugrover: That's exactly where I was having my doubts 
Ints were defined as CPyTagged but there wasn't anything for float and I got confused when I looked at the python C API docs. Can you tell me more problems in my code so I can correct things? And once again,  do I change the PyObject or the float part of the function definition?

pranshugrover: Also do I have to make an IRBuild test for this? Because I feel I've missed that part as well.

JelleZijlstra: Sorry, in this area I know enough to see problems but not enough to come up with solutions. Most likely you'll just have to use boxed floats (PyObjects), since as you said only ints are represented as native objects by mypyc.

Also, I'm pretty sure Python floats are `double`s at the C level, not `float`s.

pranshugrover: Thank you regardless, I see the error you pointed out, I guess it should be PyFloatObject instead of a float.

TH3CHARLie:
This is incorrect. mypyc supports floating-point arithmetic via `PyNumber_` protocols (`generic_ops.py`). If you want to specialize them, you should check how CPython supports them internally (`floatobject.c`) and mimic the behavior. Simply adding two floats will not result in the correct python object.

pranshugrover:
> This is incorrect. mypyc supports floating-point arithmetic via `PyNumber_` protocols (`generic_ops.py`). If you want to specialize them, you should check how CPython supports them internally (`floatobject.c`) and mimic the behavior. Simply adding two floats will not result in the correct python object.

Thank you for this, I will look into it.

JukkaL:
The internal type for `floats` is defined like this in `mypyc.ir.rtypes`:

```py
# Floats are represent as 'float' PyObject * values. (In the future
# we'll likely switch to a more efficient, unboxed representation.)
float_rprimitive = RPrimitive('builtins.float', is_unboxed=False,
                              is_refcounted=True)  # type: Final
```

Based on this, you can see that `PyObject *` is (usually) the correct type for floats. 

As mentioned in the comment, using `PyObject *` is not super efficient, but switching to a more efficient representation requires some non-trivial changes to mypyc and likely won't be feasible in the next few months, at least.

You can cast `PyObject *` to `PyFloatObject *` if you want to access the contents of the object directly. You should only do this if it's faster this way, or if the Python C API doesn't provide some features you require.

Here are docs for the relevant Python C API features: https://docs.python.org/3/c-api/float.html

In particular, `PyFloat_AS_DOUBLE` and `PyFloat_FromDouble` seem useful. From these you can see that the corresponding C type for the value of a Python float object is indeed `double`.

TH3CHARLie:
closing this as the author deleted their account (see @ghost)

PR #10307, opened 2021-04-11
[mypyc] Add 'range' primitive type

97littleleaf11:
### Description

Closes https://github.com/mypyc/mypyc/issues/770

## Test Plan

```py
def range_object() -> None:
    r = range(4, 12, 2)
    sum = 0
    for i in r:
        sum += I
```
This should generate a range object and a iterator when looping on it.

```py
def range_in_loop() -> None:
    sum = 0
    for i in range(4, 12, 2):
        sum += I
```
This should maintain original specialized IR result.


mypyc/test-data/irbuild-basic.test
```diff
@@ -3646,3 +3646,69 @@ L0:
...
     return r3
+
+[case testRangeObject]
+def range_object() -> None:
+    r = range(4, 12, 2)
+    sum = 0
+    for i in r:
+        sum += i
+
+def range_in_loop() -> None:
```
JukkaL, line 3657: There probably is an existing test case for range in a for loop, so this can be removed.

mypyc/test-data/run-loops.test
```diff
@@ -452,3 +452,19 @@ def bar(x: Optional[str]) -> None:
 [file driver.py]
 from native import bar
 bar(None)
+
+[case testRangeObject]
+r1 = range(4, 12, 2)
```
JukkaL, line 457: Can you change this to use a `test_` function without a custom driver, as that's the "modern" way to write tests.

Test also that coercion to `range` works, for example something like this:

```
def f(r: range) -> None:
    pass

r: Any = range(...)
f(r)  # OK

r = 'x'
f(r)  # Should be a TypeError

ff: Any = f
ff(r)  # Again should be a TypeError; this uses the wrapper function since the callee has type Any
```

JukkaL (commented):
Thanks for the PR! Looks good, just a few comments about tests.

mypyc/test-data/run-loops.test
97littleleaf11: I got this error after adding function `def f(r: range) -> None:`
```
run-loops.test:460: error: Function "builtins.range" is not valid as a type
run-loops.test:460: note: Perhaps you need "Callable[...]" or a callback protocol?
```

JukkaL: You'll need to change the definition of `range` into a class in `mypyc/test-data/fixtures/ir.py`.

mypyc/test-data/irbuild-basic.test
97littleleaf11: I cannot find the irbuild case and I'll keep this.

mypyc/test-data/irbuild-basic.test
```diff
@@ -3646,3 +3646,71 @@ L0:
...
+L2:
+    r2 = CPyTagged_Add(sum, i)
+    sum = r2
+L3:
+    r3 = r0 + 4
+    r0 = r3
+    i = r3
+    goto L1
+L4:
+    return 1
```
97littleleaf11, line 3719: Two cases seem to work correctly.

JukkaL (approved):
LGTM

PR #10308, opened 2021-04-11
Fix function `is_protocol_implementation` for quite similar protocol classes

### LINKED ISSUE #9771: INTERNAL ERROR when defining "__add__" in two related Protocol classes
The following minimal example causes a crash:

```
from typing import Protocol, TypeVar, Union

T1 = TypeVar("T1", bound=float)
T2 = TypeVar("T2", bound=float)


class Vector(Protocol[T1]):

    def __len__(self) -> int:
        ...

    def __add__(
        self: "Vector[T1]",
        other: Union[T2, "Vector[T2]"],
    ) -> "Vector[Union[T1, T2]]":
        ...


class Matrix(Protocol[T1]):

    def __add__(
        self: "Matrix[T1]",
        other: Union[T2, Vector[T2], "Matrix[T2]"],
    ) -> "Matrix[Union[T1, T2]]":
        ...
```

The traceback:

```
  File "C:\Users\tyralla\AppData\Local\Programs\Python\Python38\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\tyralla\AppData\Local\Programs\Python\Python38\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "mypy\checker.py", line 401, in accept
  File "mypy\nodes.py", line 940, in accept
  File "mypy\checker.py", line 1756, in visit_class_def
  File "mypy\checker.py", line 1824, in check_protocol_variance
  File "mypy\subtypes.py", line 93, in is_subtype
  File "mypy\subtypes.py", line 135, in _is_subtype
  File "mypy\types.py", line 794, in accept
  File "mypy\subtypes.py", line 264, in visit_instance
  File "mypy\subtypes.py", line 538, in is_protocol_implementation
  File "mypy\subtypes.py", line 605, in find_member
  File "mypy\subtypes.py", line 701, in find_node_type
  File "mypy\expandtype.py", line 29, in expand_type_by_instance
  File "mypy\expandtype.py", line 16, in expand_type
  File "mypy\types.py", line 1098, in accept
  File "mypy\expandtype.py", line 97, in visit_callable_type
  File "mypy\expandtype.py", line 143, in expand_types
  File "mypy\types.py", line 1724, in accept
  File "mypy\expandtype.py", line 123, in visit_union_type
  File "mypy\typeops.py", line 366, in make_simplified_union
  File "mypy\subtypes.py", line 1156, in is_proper_subtype
  File "mypy\subtypes.py", line 1177, in _is_proper_subtype
  File "mypy\types.py", line 794, in accept
  File "mypy\subtypes.py", line 1273, in visit_instance
  File "mypy\subtypes.py", line 556, in is_protocol_implementation
  File "mypy\subtypes.py", line 1156, in is_proper_subtype
  File "mypy\subtypes.py", line 1177, in _is_proper_subtype
  File "mypy\types.py", line 1098, in accept
  File "mypy\subtypes.py", line 1294, in visit_callable_type
  File "mypy\subtypes.py", line 839, in is_callable_compatible
  File "mypy\subtypes.py", line 1068, in unify_generic_callable
  File "mypy\constraints.py", line 101, in infer_constraints
  File "mypy\constraints.py", line 174, in _infer_constraints
  File "mypy\types.py", line 794, in accept
  File "mypy\constraints.py", line 399, in visit_instance
  File "mypy\subtypes.py", line 554, in is_protocol_implementation
  File "mypy\subtypes.py", line 93, in is_subtype
  File "mypy\subtypes.py", line 135, in _is_subtype
  File "mypy\types.py", line 1098, in accept
  File "mypy\subtypes.py", line 299, in visit_callable_type
  File "mypy\subtypes.py", line 839, in is_callable_compatible
  File "mypy\subtypes.py", line 1068, in unify_generic_callable
  File "mypy\constraints.py", line 101, in infer_constraints
  File "mypy\constraints.py", line 174, in _infer_constraints
  File "mypy\types.py", line 794, in accept
  File "mypy\constraints.py", line 401, in visit_instance
  File "mypy\constraints.py", line 437, in infer_constraints_from_protocol_members
```

How I use Mypy:

- Mypy version used: 0.790
- Mypy command-line flags: --follow-imports=silent --show-error-codes --warn-unused-ignores --warn-redundant-casts --strict --show-traceback
- Python version used: 3.8.5
- Operating system and version: Windows 10

tyralla:
Fixes #9771

```
from typing import Protocol, TypeVar, Union

T1 = TypeVar("T1", covariant=True)
T2 = TypeVar("T2")

class P1(Protocol[T1]):
    def b(self) -> int: ...
    def a(self, other: "P1[T2]") -> T1: ...

class P2(Protocol[T1]):
    def a(self, other: Union[P1[T2], "P2[T2]"]) -> T1: ...
```

Method `infer_constraints_from_protocol_members` of class `ConstraintBuilderVisitor` raises an assertion for the above code when we pass `P2` as `instance` and `P1` as `template` to it (inline comment: _The above is safe since at this point we know that 'instance' is a subtype of (erased) 'template', therefore it defines all protocol members_).

So, the actual error happens before. Method `visit_instance` of class `ConstraintBuilderVisitor` should never apply the method `infer_constraints_from_protocol_members` on `P1` and `P2`. It nevertheless does so due to function `is_protocol_implementation` indicating `P2` actually implements `P1`.

In my naive understanding, the problem seems to lie in the recursive approach underlying function `is_protocol_implementation`. This function first analyses method `a` to check if `P2` is a subclass of `P1`. Therefore, it needs to check if `P1` is a subclass of `P2`. This can result in a situation where the "'assuming' structural subtype matrix" is already filled with `P1` and `P2`.  `is_protocol_implementation` then returns `True` (for avoiding infinite recursion in other cases) without checking any other members, which is incorrect for the given case.

My simple solution is first to check if all members of `P1` (the potential supertype) define a subset of `P2` (the potential subtype). If this does not hold, we do not need to perform the complex recursive approach and can return `False` immediately. Maybe not the most general solution, but it works on the additional test `testTwoUncomfortablyIncompatibleProtocolsWithoutRunningInIssue9771` (I can't think of a better name...) and seems not to impact any other tests.


JelleZijlstra:
I'm not sure how, but this PR actually fixes a more basic bug. This code passes on master:

```python
from typing import Hashable, Iterable

def f(x: Hashable) -> None:
    pass

def g(x: Iterable[str]) -> None:
    f(x)
```
But with your PR it correctly produces `lruc.py:20: error: Argument 1 to "f" has incompatible type "Iterable[str]"; expected "Hashable"`. The error mypy-primer found in Black above alerted me to this.

hauntsaninja (approved):
I think Jelle's fixed issue is explained by:
- https://github.com/python/mypy/blob/9ab622cf557f1184dd040ff4b282c7a13ea3d976/mypy/subtypes.py#L598 finds that Iterable does have a `__hash__` method in its MRO.
- The protocol_members attribute doesn't look at object https://github.com/python/mypy/blob/fffbe88fc54807c8b10ac40456522ad2faf8d350/mypy/nodes.py#L2497

I'm not the most familiar with this code, but your write up makes sense, this passes tests and fixes a crash multiple people have run into, so I'll merge soon if no one else chimes in. Thanks! :-)

JelleZijlstra:
I feel like we should add an explicit test for the Hashable/Iterable case to make sure it doesn't regress.

tyralla:
I tried to add the following regression test you suggested: 

```python
[case testHashable]

from typing import Hashable, Iterable

def f(x: Hashable) -> None:
    pass

def g(x: Iterable[str]) -> None:
    f(x)   # E: Argument 1 to "f" has incompatible type "Iterable[str]"; expected "Hashable"

```

Additional content of typing-full.pyi:

```python
@runtime_checkable
class Hashable(Protocol, metaclass=ABCMeta):
    @abstractmethod
    def __hash__(self) -> int: pass
```

However, this test passes both with and without the PR.  Am I missing something?

hauntsaninja:
I'm guessing object doesn't have __hash__ in the test fixtures

tyralla:
Yes, you are right. Defining a fixture for such an essential feature did not come into my mind. 

I added the minimal stub file `object_hashable.pyi`. Now the test actually fails after uncommenting the additional check that I introduced into `is_protocol_implementation`.

JelleZijlstra:
Thanks for adding the test!

hauntsaninja:
Thank you!

PR #10141, opened 2021-02-26
Fixes Final docs

sobolevn:
### Description

There were several things to imporve in `Final` docs:
1. For some reason it was `ID: Final[float] = 1`, while `1` is clearly `int`
2. There was also a quite misleading phrase about `ID: Final = 1` being `int` and not `Literal[1]`

PR #10316, opened 2021-04-12
Now Final types are better

sobolevn:
Related #10141

### DIFF
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -2,6 +2,7 @@ import re
 import sys
 import warnings
 
+import typing  # for typing.Type, which conflicts with types.Type
 from typing import (
     Tuple, Union, TypeVar, Callable, Sequence, Optional, Any, Dict, cast, List, overload
 )
@@ -397,7 +398,7 @@ class ASTConverter:
         ast3.BitXor: '^',
         ast3.BitAnd: '&',
         ast3.FloorDiv: '//'
-    }  # type: Final
+    }  # type: Final[Dict[typing.Type[AST], str]]
 
     def from_operator(self, op: ast3.operator) -> str:
         op_name = ASTConverter.op_map.get(type(op))
@@ -417,7 +418,7 @@ class ASTConverter:
         ast3.IsNot: 'is not',
         ast3.In: 'in',
         ast3.NotIn: 'not in'
-    }  # type: Final
+    }  # type: Final[Dict[typing.Type[AST], str]]
 
     def from_comp_operator(self, op: ast3.cmpop) -> str:
         op_name = ASTConverter.comp_op_map.get(type(op))
--- a/mypy/fastparse.py
+++ b/mypy/fastparse.py
@@ -2,7 +2,6 @@ import re
 import sys
 import warnings
 
-import typing  # for typing.Type, which conflicts with types.Type
 from typing import (
     Tuple, Union, TypeVar, Callable, Sequence, Optional, Any, Dict, cast, List, overload
 )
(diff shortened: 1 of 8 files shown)


mypy/main.py
```diff
@@ -164,7 +164,7 @@ def _fill_text(self, text: str, width: int, indent: str) -> str:
     ('allow', 'disallow'),
     ('show', 'hide'),
 ]  # type: Final
-flag_prefix_map = {}  # type: Final[Dict[str, str]]
+flag_prefix_map = {}  # type: Final
```
JukkaL, line 167: I'd prefer to leave this as is, since the type is not immediately obvious. The type comments acts as documentation.

mypy/report.py
```diff
@@ -41,7 +41,7 @@
     (TypeOfAny.from_error, "Error"),
     (TypeOfAny.special_form, "Special Form"),
     (TypeOfAny.implementation_artifact, "Implementation Artifact"),
-])  # type: Final[collections.OrderedDict[int, str]]
```
JukkaL, line 44: Again here the redundant type is kind of useful as documentation, so maybe leave it here.

mypyc/irbuild/classdef.py
```diff
@@ -205,7 +205,7 @@ def allocate_class(builder: IRBuilder, cdef: ClassDef) -> Value:
 MAGIC_TYPED_DICT_CLASSES = (
     'typing._TypedDict',
     'typing_extensions._TypedDict',
-)  # type: Final[Tuple[str, ...]]
```
JukkaL, line 208: This changes the semantics, so we want to preserve the type. We want this to be an arbitrary-length tuple instead of a fixed-length tuple.

mypy/fastparse2.py
```diff
@@ -254,7 +253,7 @@ def translate_type_comment(self, n: ast27.stmt,
         ast27.BitXor: '^',
         ast27.BitAnd: '&',
         ast27.FloorDiv: '//'
-    }  # type: Final[Dict[typing.Type[AST], str]]
```
JukkaL, line 257: Let's keep this since otherwise it may be unclear what the inferred type would be.

mypy/fastparse2.py
```diff
@@ -276,7 +275,7 @@ def from_operator(self, op: ast27.operator) -> str:
         ast27.IsNot: 'is not',
         ast27.In: 'in',
         ast27.NotIn: 'not in'
-    }  # type: Final[Dict[typing.Type[AST], str]]
```
JukkaL, line 279: Similar to above, this type annotation seems somewhat useful.

mypy/fastparse.py
```diff
@@ -398,7 +397,7 @@ def translate_type_comment(self,
         ast3.BitXor: '^',
         ast3.BitAnd: '&',
         ast3.FloorDiv: '//'
-    }  # type: Final[Dict[typing.Type[AST], str]]
+    }  # type: Final
```
JukkaL, line 400: Let's keep this since otherwise it may be unclear what the inferred type would be.

mypy/fastparse.py
```diff
@@ -418,7 +417,7 @@ def from_operator(self, op: ast3.operator) -> str:
         ast3.IsNot: 'is not',
         ast3.In: 'in',
         ast3.NotIn: 'not in'
-    }  # type: Final[Dict[typing.Type[AST], str]]
+    }  # type: Final
```
JukkaL, line 420: Similar to above.

JukkaL (commented):
Thanks! Looks good, though on reflection I actually like some of the explicit types and would like to keep them.

sobolevn:
Done! Thanks a lot for the feedback!

JukkaL (approved):
Thanks for the updates!

PR #10318, opened 2021-04-13
[mypyc] Add "is" check for "in" operations

DunderBird:
### Description
Fix semantic difference mentioned in https://github.com/mypyc/mypyc/issues/726.

CPython will try `is` test before compare. For example, `b in [a]` actually means `b is a or b == a`. The following example show the difference:
```python
from typing import ClassVar
class B:
    a: ClassVar[int] = 1
    def __eq__(self, other) -> bool:
        return self.a == 1

b = B()
print(b == b)
print(b in [b])
B.a = 2
print(b == b)
print(b in [b])
```
cpython(python3.8) outputs:
```
True
True
False
True
```
while the program compiled outputs:
```
True
True
False
False
```

### Problems
In `mypyc/irbuild/expression.py: transform_comparison_expr` I check `is` statically before generating expressions(line 446 and line 454). This assumes if `a`'s node is same to `b`'s then `a is b` is always true. I don't know whether this assumption always holds. I add this mainly to avoid generate redundant IR like `r1 == r1` which gcc will show warnings.

### DIFF
--- a/mypyc/irbuild/expression.py
+++ b/mypyc/irbuild/expression.py
@@ -467,6 +467,7 @@ def transform_comparison_expr(builder: IRBuilder, e: ComparisonExpr) -> Value:
                 item_type = builder.node_type(item)
                 eq_expr = ComparisonExpr([cmp_op], [lhs, item])
                 # only add `is` check for two boxed types
+                expr: Union[OpExpr, ComparisonExpr]
                 if item_type.is_unboxed or lhs_type.is_unboxed:
                     expr = eq_expr
                 else:
--- a/mypyc/irbuild/expression.py
+++ b/mypyc/irbuild/expression.py
@@ -441,6 +441,7 @@ def transform_comparison_expr(builder: IRBuilder, e: ComparisonExpr) -> Value:
         # 16 is arbitrarily chosen to limit code size
         if 1 <= n_items < 16:
             lhs = e.operands[0]
+            lhs_type = builder.node_type(lhs)
             if e.operators[0] == 'in':
                 # check `is` relation statically
                 for item in items:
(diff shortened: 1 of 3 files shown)


mypyc/test-data/irbuild-tuple.test
```diff
@@ -195,70 +195,142 @@ def f(i: int) -> bool:
...
-    r13, r14 :: bit
-    r15 :: bool
-    r16 :: bit
+    r0, r1 :: object
+    r2 :: bit
+    r3 :: object
+    r4 :: int32
+    r5 :: bit
+    r6, r7, r8 :: bool
+    r9 :: int64
```
TH3CHARLie, line 204: change all `int64` in IR to `native_int` will fix the CI failure.

mypyc/test-data/run-lists.test
```diff
@@ -338,3 +364,6 @@ assert list_in_mixed(0.0)
 assert not list_in_mixed([1])
 assert not list_in_mixed(object)
 assert list_in_mixed(type)
+
+assert inop_is_check(glob_eqFalse)
+assert not inop_is_check(EqFalse())
```
TH3CHARLie, line 369: `test_inop_is` is not tested

mypyc/irbuild/expression.py
```diff
@@ -436,23 +436,36 @@ def transform_comparison_expr(builder: IRBuilder, e: ComparisonExpr) -> Value:
...
-        # x in y -> x == y[0] or ... or x == y[n]
-        # x not in y -> x != y[0] and ... and x != y[n]
+        # x in y -> (x is y[0] or x == y[0]) or ... or (x is y[n] or x == y[n])
+        # x not in y -> (x is not y[0] and x != y[0]) and ... and (x is not y[n] and x != y[n])
         # 16 is arbitrarily chosen to limit code size
-        if 1 < n_items < 16:
+        if 1 <= n_items < 16:
+            lhs = e.operands[0]
             if e.operators[0] == 'in':
+                # check `is` relation statically
```
JukkaL, line 445: We shouldn't use the `is` check for any unboxed types, such as `int_rprimitive`, since it's mostly redundant and can cause unnecessary boxing. For example, look at the type of the operands, and use the extra `is` check only if both of the operands are boxed types (use `is_unboxed` of `RType`).

As an optional optimization, if both types are straightforward native classes, and we know all the subclasses, and there is no custom `__eq__` in any of the classes or subclasses, we know that there is no `__eq__` override. This means that the `==` check is the same as an `is` check, and we can skip the latter one.

DunderBird: Sorry to respond late. Thank you for your review and I commit to solve the first problem. As for the optional optimization, I can't figure out what do you mean by straightforward native classes and how can we know all the subclasses.

JukkaL:
Self type check is failing. Can you rebase to master (to fix Python 3.5 build failures) and fix the build?

PR #10322, opened 2021-04-13
[WIP] Optimize setdefault to only construct default collection when needed #830 

pranshugrover:
I attempted to resolve this, I am having trouble with doing it for all 3 data types mentioned simultaneously. Do we make separate function for that? Also what test are we looking at for this.


mypyc/lib-rt/dict_ops.c
```diff
@@ -394,3 +394,22 @@ tuple_T4CIOO CPyDict_NextItem(PyObject *dict_or_iter, CPyTagged offset) {
     Py_INCREF(ret.f3);
     return ret;
 }
+
+PyObject*CPyDict_SetDefault_list(CPyTagged x,  PyList_Type l ){ 
+
+    int v;
+    v = int CPyDict_Contains(PyObject* k, CPyTagged* x){
```
TH3CHARLie, line 401: this is not valid C code

TH3CHARLie:
closing this as the author deleted their account (see @ghost)

PR #10323, opened 2021-04-13
[mypyc] optimize setdefault: create empty list and dict lazily

DunderBird:
### Description
Optimize `setdefault` according to https://github.com/mypyc/mypyc/issues/830. Only list and dict are implemented.

I add `match` and `arg_converter` to CFunctionDescription. `match` is used to check if this primitive function is matched(this work is done in ll_builder originally). `arg_converter` is used to generate argument list for current primitive function.

`CPyDict_SetDefaultWithEmptyCollection` is a new primitive function which will generate empty list/dict lazily. It's "`match`" is `_setdefault_empty_match` in which I just test the arg's function_name. It's "arg_converter" is `_setdefault_empty_create_args`.

I find a bug(https://github.com/mypyc/mypyc/issues/831) during my development but I'm not sure whether my code causes it because after I switch to master branch the bug still exists.


TH3CHARLie:
The implementation is overcomplicated. Usually, we don't change the core logic that all existing primitives are relying on before discussion because it can easily lead to bugs. And adding new primitive should always be small and incremental changes.

The idea of lazy allocation is good but I don't think it should be implemented using a primitive, especially when we are using integer code to represent containers which will break the consistency of the semantics IMO. I prefer to implement this optimization directly using IR build.

JukkaL:
Agreed. In particular, this could be implemented in `mypyc.irbuild.specialize` as a specialization of `dict.setdefault`.

PR #10324, opened 2021-04-13
Expose proper plugin plugin from within mypy for external usage

### LINKED ISSUE #10306: Expose the proper plugin in the published mypy version
**Feature**
The misc/proper_plugin.py is currently not packaged with mypy. Can we expose it as part of mypy itself?
 
**Pitch**
I think it would be wise to have a single source of truth for the proper_plugin.py. As the mypy project already uses it to validate internal plugins like the dataclasses, it should stay up to date. Some projects copy the file to their project, but that can cause confusing from not being up to date (i.e. https://github.com/python/mypy/issues/8047#issuecomment-560124874). I think externally developed plugins could benefit from the plugin. So perhaps exposing it as `mypy.misc.proper_plugin`?

MaicoTimmerman:
### Description

This PR moves the proper_plugin.py into the mypy package itself, such that it can be used by developers creating their own mypy plugin directly.

Fixes #10306

## Test Plan
The plugin was no covered by tests, however it is used to verify the internal plugins and continues to do so.


JukkaL:
Hmm this is probably fine, but I don't think I understand the use case here. What would you use this plugin for? It seems pretty specific to mypy internals.

MaicoTimmerman:
> Hmm this is probably fine, but I don't think I understand the use case here. What would you use this plugin for? It seems pretty specific to mypy internals.

It has been used it to detect some problems in https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/ext/mypy/plugin.py.

MaicoTimmerman:
@JukkaL Anything you want to do or clarify so that this can be merged?

MaicoTimmerman:
@JukkaL I noticed you triggered the checks, but meanwhile this PR has sunk to the second page, can we merge it?

MaicoTimmerman:
I'm closing this MR. Sqlalchemy moved off mypy plug-ins, to be type checker agnostic.

PR #10331, opened 2021-04-16
[WIP] __annotations__ is no longer undefined

### LINKED ISSUE #10330: Module-level `__annotations__` variable is flagged as undefined
To repro, run mypy on the following:

```python
print(__annotations__)
```

```
test.py:1: error: Name '__annotations__' is not defined
```

A pyright user recently noted [the same issue in pyright](https://github.com/microsoft/pylance-release/issues/1161), which I have fixed.

Axeinator:
### Description
Allows __annotations__ to be accessed without error

Fixes #10330 

## Test Plan

Waiting for test plan design...


JelleZijlstra:
That sounds like an issue with the test fixtures. You can add a different fixture in your tests by using `[fixture ...]` or add dict to the one you're already using.

Axeinator:
I'm also getting a different error here, with the tests again.

`Implicit generic "Any". Use "typing.Dict" and specify generic parameters`

PR #10333, opened 2021-04-16
[mypyc] Dict true test

97littleleaf11:
### Description

Closes https://github.com/mypyc/mypyc/issues/768

* Modify `builtin_len(dict)`.
    * In `cpython/dictobject.h`
      ```c
       /* Get the number of items of a dictionary. */
      #define PyDict_GET_SIZE(mp)  (assert(PyDict_Check(mp)),((PyDictObject *)mp)->ma_used)
      ```
    * Thus replace `call_c` with `LoadMem` and add `PyDictObject`. Related issue: https://github.com/mypyc/mypyc/issues/835

## Test Plan

An IR test and a functional test.
Similar to https://github.com/python/mypy/commit/1e96f1df48a6e06ba598fa8230b96671b206cba0

### DIFF
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -801,5 +801,14 @@ PySetObject = RStruct(
 PyListObject = RStruct(
     name='PyListObject',
     names=['ob_base', 'ob_item', 'allocated'],
-    types=[PyObject, pointer_rprimitive, c_pyssize_t_rprimitive]
+    types=[PyVarObject, pointer_rprimitive, c_pyssize_t_rprimitive]
+)
+
+# ma_keys: PyDictKeysObject*
+# ma_values: PyObject**
+PyDictObject = RStruct(
+    name='PyDictObject',
+    names=['ob_base', 'ma_used', 'ma_version_tag', 'ma_keys', 'ma_values'],
+    types=[PyObject, c_pyssize_t_rprimitive, uint64_rprimitive, pointer_rprimitive,
+           pointer_rprimitive]
 )
--- a/mypyc/ir/rtypes.py
+++ b/mypyc/ir/rtypes.py
@@ -801,14 +801,5 @@ PySetObject = RStruct(
 PyListObject = RStruct(
     name='PyListObject',
     names=['ob_base', 'ob_item', 'allocated'],
-    types=[PyVarObject, pointer_rprimitive, c_pyssize_t_rprimitive]
-)
-
-# ma_keys: PyDictKeysObject*
-# ma_values: PyObject**
-PyDictObject = RStruct(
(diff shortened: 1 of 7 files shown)


JukkaL (commented):
Hmm the original implementation may actually be buggy. We should allow subclasses to override `__len__`, for consistency with other dict operations, and this means that we should have a type guard. This probably makes it too complex to be done directly in IR, and instead we could use a C helper that uses `PyDict_GET_SIZE` behind a type guard. Sorry about that!

This should work similar to CPython and print 'here':

```py
class D(dict):
    def __len__(self) -> int:
        return 1

if D():
    print('here') 
```

97littleleaf11:
Got it. I will start a new branch from beginning later.

97littleleaf11:
I tried adding this:

```py
Py_ssize_t CPyDict_Size(PyObject *dict) {
    if (PyDict_CheckExact(dict)) {
        return PyDict_Size(dict);
    }
    return PyObject_Size(dict);
}
```

```py
class MyDict(dict):
    def __init__(self, *args, **kwargs):
        self.update(*args, **kwargs)

    def __len__(self):
        return 1


def test_dict_subclass_to_bool() -> None:
    d = MyDict()
    if d:
        print("true")
```

However `PyObject_Size` won't call the overrided `__len__`. 
Does the `cast` operation cause that?

```
L0:
    r0 = tmp5.MyDict :: type
    r1 = PyObject_CallFunctionObjArgs(r0, 0)
    if is_error(r1) goto L7 (error at test_dict_subclass_to_bool:45) else goto L1
L1:
    r2 = cast(dict, r1)
    if is_error(r2) goto L7 (error at test_dict_subclass_to_bool:45) else goto L2
L2:
    d = r2
    r3 = CPyDict_Size(d)
    dec_ref d
    r4 = r3 >= 0 :: signed
    if not r4 goto L7 (error at test_dict_subclass_to_bool:45) else goto L3 :: bool
```

JukkaL:
> However PyObject_Size won't call the overrided __len__.

Hmm I wonder if this is a bug in `PyObject_Size`. Anyway, it's already broken for `len(d)`, so not supporting it for boolean checks seems okay. So let's drop the `__len__` override test case and use the original implementation for now. Sorry about the back and forth -- there are some gray areas with respect to CPython compatibility where it's often unclear what's the best course of action.

But can you add a new issue to track the issue of overriding `__len__`? To properly fix it, we should fix it for `len()` as well.

> Does the cast operation cause that?

No, the cast does not modify the operand.

JukkaL (approved):
Thanks for the updates! This looks good, and it made a quick benchmark I wrote about 50% faster, which is a nice speedup.

PR #10334, opened 2021-04-16
[mypyc] Add incref in CPyDict_SetDefault

97littleleaf11:
### Description

Closes https://github.com/mypyc/mypyc/issues/831


mypyc/lib-rt/dict_ops.c
```diff
@@ -69,7 +69,9 @@ PyObject *CPyDict_GetWithNone(PyObject *dict, PyObject *key) {
 
 PyObject *CPyDict_SetDefault(PyObject *dict, PyObject *key, PyObject *value) {
     if (PyDict_CheckExact(dict)){
-        return PyDict_SetDefault(dict, key, value);
+        PyObject* ret = PyDict_SetDefault(dict, key, value);
+        Py_INCREF(ret);
```
JelleZijlstra, line 73: I think this should be Py_XINCREF in case SetDefault returned NULL.

JukkaL (approved):
Thanks, looks good!

PR #10335, opened 2021-04-16
Use double quotes errors in `already defined`, `is nonlocal and global`, `No binding for nonlocal` and `is already defined in local`

dixith:
### Description
This pull request fixes issue reported in #7445   for `already defined`, `is nonlocal and global`, `No binding for nonlocal` and `is already defined in local` error messages

## Test Plan


JukkaL (approved):
Thanks for the PR! Looks good. The CI failure seems like a flake that we've been seeing recently.

PR #10338, opened 2021-04-17
Fixes #10083.

### LINKED ISSUE #10083: str.fomat index_string too restricted
**Bug Report**

The following code is flagged by `mypy-0.800`:
```python
"{0[k-y]}".format({"k-y": "val"})
```
The error is
> Invalid index expression in format field accessor "[k-y]"

According to [Python Format String Syntax](https://docs.python.org/3/library/string.html#formatstrings):
> replacement_field ::=  "{" [field_name] …
> field_name        ::=  arg_name ( … | "[" element_index "]")*
> element_index     ::=  … | index_string
> index_string      ::=  <any source character except "]"> +

the `-` should be okay. Similar when using `+` or `/`. Looks like only `[0-9A-Z_a-z]` are allowed with `mypy`.

**To Reproduce**

```
mypy -c '"{0[k-y]}".format({"k-y": ""})'
<string>:1: error: Invalid index expression in format field accessor "[k-y]"
```

**Expected Behavior**

No error should be detected.

**Actual Behavior**

`mypy` prints an error.

**Your Environment**

Debian-10-Buster with self-built  `mypy-0.800`

hatal175:
### Description

Fixes #10083.

Use the same string formatting index splitting function that python uses instead of relying only on ast parsing so it would
have the same logic as python.
It also changes some errors since it catches syntax errors.

## Test Plan

I've added some positive assertion tests and changed tests where error came out with different text. I had to fix the testFormatCallAccessorsIndices test that assumed that string format can evaluate functions inside the index (it can't).

I believe the current testing + mypy primer should cover our bases.

### DIFF
--- a/mypy/checkstrformat.py
+++ b/mypy/checkstrformat.py
@@ -11,12 +11,14 @@ implementation simple.
 """
 
 import re
+import string
 
 from typing import (
     cast, List, Tuple, Dict, Callable, Union, Optional, Pattern, Match, Set, Any
 )
 from typing_extensions import Final, TYPE_CHECKING
 
+from mypy.traverser import TraverserVisitor
 from mypy.types import (
     Type, AnyType, TupleType, Instance, UnionType, TypeOfAny, get_proper_type, TypeVarType,
     LiteralType, get_proper_types
@@ -162,6 +164,30 @@ class ConversionSpecifier:
     def has_star(self) -> bool:
         return self.width == '*' or self.precision == '*'
 
+class LineNoChanger(TraverserVisitor):
+    def __init__(self, line: int, end_line: Optional[int], column: int) -> None:
+        self.line = line
+        self.end_line = end_line
+        self.column = column
+
+    def set_context_info(self, ctx: Context) -> None:
+        ctx.line = self.line
+        ctx.end_line = self.end_line
+        ctx.column = self.column
+
+    def visit_index_expr(self, expr: IndexExpr) -> None:
+        self.set_context_info(expr)
+        super().visit_index_expr(expr)
+
+    def visit_member_expr(self, o: MemberExpr) -> None:
+        self.set_context_info(o)
+        super().visit_member_expr(o)
+
+    def visit_int_expr(self, o: IntExpr) -> None:
+        self.set_context_info(o)
+
+    def visit_str_expr(self, o: StrExpr) -> None:
+        self.set_context_info(o)
 
 class StringFormatterChecker:
     """String interpolation/formatter type checker.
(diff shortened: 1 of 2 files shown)


mypy/checkstrformat.py
```diff
@@ -518,7 +546,21 @@ def apply_field_accessors(self, spec: ConversionSpecifier, repl: Expression,
...
-        dummy = DUMMY_FIELD_NAME + spec.field[len(spec.key):]
+
+        try:
+            first_obj, split_fields = string._string.formatter_field_name_split(
+                spec.field[len(spec.key):])
+            split_fields_str = [
+                ("." + str(field)) if is_attr else
+                '[{}]'.format(field) if isinstance(field, int) else
+                '["{}"]'.format(field.replace('"', '\\"')) for is_attr, field in split_fields]
+        except Exception:
```
JukkaL, line 557: We avoid catching `Exception` since it can be triggered by unexpected error conditions. Can you make this more specific? Also, if you narrow this down to the actual call that can fail instead of several lines of code, it would be easier to see what is likely to raise an exception.

mypy/checkstrformat.py
```diff
@@ -518,7 +546,21 @@ def apply_field_accessors(self, spec: ConversionSpecifier, repl: Expression,
 
         # This is a bit of a dirty trick, but it looks like this is the simplest way.
         temp_errors = self.msg.clean_copy().errors
-        dummy = DUMMY_FIELD_NAME + spec.field[len(spec.key):]
+
+        try:
+            first_obj, split_fields = string._string.formatter_field_name_split(
+                spec.field[len(spec.key):])
+            split_fields_str = [
+                ("." + str(field)) if is_attr else
```
JukkaL, line 554: This list comprehension is a bit difficult to understand. Can you change it into a for loop, to improve readability?

JukkaL (commented):
Thanks for the PR! Left a few ideas about how to improve this PR.

mypy/checkstrformat.py
```diff
@@ -163,6 +165,32 @@ def has_star(self) -> bool:
         return self.width == '*' or self.precision == '*'
 
 
+class LineNoChanger(TraverserVisitor):
```
hauntsaninja, line 168: nit: can we name this LineNumChanger (or make the n lower case?) sounds like it's not changing the lines :P

hatal175: Done.

mypy/checkstrformat.py
hatal175: Done

mypy/checkstrformat.py
hatal175: I narrowed it down to a ValueError but either formatter_field_name_split or the iteration itself can throw it so I don't see much point separating them.

97littleleaf11:
Can `LineNumChanger` also benefit `--pretty`?

97littleleaf11:
could you please resolve the conflicts? the string formatting tests have been moved to a new file.

hatal175:
Rebased.

hatal175:
I don't know why mypyc fails on string._string. I don't have the time to go through this - this used to work so something in the meanwhile must have changed.

PR #10345, opened 2021-04-20
Stop running Python 3.5 in CI

hauntsaninja:
Python 3.5 always fails in CI these days. So a) it's annoying and a bad contributor experience, b) we're effectively not testing against Python 3.5 anyway
Linking https://github.com/python/mypy/issues/9950

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,14 +24,6 @@ env:
 jobs:
   fast_finish: true
   include:
-  # Specifically request 3.5.1 because we need to be compatible with that.
-  - name: "run test suite with python 3.5.1 (compiled with mypyc)"
-    python: 3.5.1
-    dist: trusty
-    env:
-    - TOXENV=py
-    - EXTRA_ARGS="-n 2"
-    - TEST_MYPYC=1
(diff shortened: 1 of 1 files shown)


ethanhs (approved):
I look forward to dropping support so we can use TypeVar without needing to guard it.

TH3CHARLie:
Agreed. Should we keep the number of compiled suites by making 3.6 compiled?

hauntsaninja:
You probably have a better sense of that than I do — how likely is it that 3.6 compiled will catch issues not caught by 3.7 compiled and 3.6?

hauntsaninja:
I do seem to remember we had a CI false negative once because we didn't have a compiled >= 3.8 (due to some typeshed change to ast), so maybe we make Python 3.9 compiled

TH3CHARLie (approved):
A compiled 3.9 LGTM!

JukkaL:
I think that we should run compiled tests on the earliest supported/working Python version and the latest, so it would be 3.6 and 3.9 after this change.

PR #10347, opened 2021-04-21
Remove types-typing-extensions dependency

hauntsaninja:
This is due to https://github.com/python/typeshed/pull/5233


JelleZijlstra (approved):
Looks good, but curious what's going on with the mypy-primer change. It's on https://github.com/Akuli/porcupine/blob/master/porcupine/settings.py#L388, cc @Akuli.

hauntsaninja:
It's because porcupine uses a custom typeshed dir, which hasn't been updated to include typing_extensions in the stdlib. I assume this prevents mypy from matching the Literal overloads.

PR #10349, opened 2021-04-21
Run compiled tests on the earliest supported Python (3.6)

TH3CHARLie:
relates https://github.com/python/mypy/pull/10345#issuecomment-823896985


JukkaL (approved):
Thanks!

PR #10350, opened 2021-04-21
Make TypeMap use literal_hash instead of Expression as key

freundTech:
The comment above TypeMap reads:
```
# NB: The keys of this dict are nodes in the original source program,
# which are compared by reference equality--effectively, being *the
# same* expression of the program, not just two identical expressions
# (such as two references to the same variable). TODO: it would
# probably be better to have the dict keyed by the nodes' literal_hash
# field instead.
```
While developing #10191 I ran into multiple cases where typemaps using literal_hash would have made my code both cleaner and faster. Instead I has to resort to calculating literal_hash for all elements of a typemap on the spot in multiple places.

With this PR I tried to check what it would take to use a literal_hash as a key. The result works, but it isn't pretty, resorting to some hacks in other places. Therefore this PR is mostly meant as a PoC and I wouldn't recommend merging it.

The main hold-ups are:
* `checker.type_map` needs to keep expressions as key, as it also maps non-literal expressions to types
* propagate_up_typemap_info needs to access values from `checker.type_map`, but only has access to literal_hashes instead of expressions
* `binder` needs to get the declared type for elements of a `type_map`, which is easy with the expression, but not with the literal_hash.

My recommendation would be to replace the whole `literal_hash` system with an `expression_hash`, that still maps equivalent literal expressions to equal hashes, but instead of mapping non-literals expressions to `None` maps them to unique hashes (For example to themselves). That would solve the first two problems.
I'm still thinking about how to solve the third one.

For now I'll focus my time on #10191, but I might come back to this in the future.

### DIFF
--- a/mypy/binder.py
+++ b/mypy/binder.py
@@ -127,17 +127,13 @@ class ConditionalTypeBinder:
                 return self.frames[i].types[key]
         return None
 
-    def put(self, expr: Expression, typ: Type) -> None:
-        if not isinstance(expr, (IndexExpr, MemberExpr, NameExpr)):
+    def put(self, lhash: Key, typ: Type) -> None:
+        if lhash[0] not in ('Var', 'Member', 'Index'):
             return
-        if not literal(expr):
-            return
-        key = literal_hash(expr)
-        assert key is not None, 'Internal error: binder tried to put non-literal'
-        if key not in self.declarations:
-            self.declarations[key] = get_declaration(expr)
-            self._add_dependencies(key)
-        self._put(key, typ)
+        if lhash not in self.declarations:
+            self.declarations[lhash] = get_declaration_from_literal(lhash)
+            self._add_dependencies(lhash)
+        self._put(lhash, typ)
 
     def unreachable(self) -> None:
         self.frames[-1].unreachable = True
@@ -265,6 +261,8 @@ class ConditionalTypeBinder:
             return None
         if not literal(expr):
             return
+        lhash = literal_hash(expr)
+        assert lhash is not None
         self.invalidate_dependencies(expr)
 
         if declared_type is None:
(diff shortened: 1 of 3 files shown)


JukkaL:
> My recommendation would be to replace the whole literal_hash system with an expression_hash, that still maps equivalent literal expressions to equal hashes, but instead of mapping non-literals expressions to None maps them to unique hashes (For example to themselves).

This sounds worth experimenting with. However, we have some checks against `None` literal hashes -- we only use the binder for expressions that are "literal", I think. How would this idea work with that? Maybe we'd check if the expression hashes to itself to check if it's a literal expression?

Also, I'd like to see some examples where you have to calculate literal hashes redundantly. Maybe there's a third approach.

[Somewhat related: It might be worth it to rename this use of the term "literal", since it predates literal types, and can easily be confused with those.]

freundTech:
The following checks would be made easier when TypeMap would use literal_hash:

Inferring types after match statements:
```python
match m:
    case list(x):
        pass
    case dict(x):
        pass
```
I need to know that those two x NameExpr are the same x.

Checking for duplicate captures:
```python
match m:
    case [x, x]:
        pass
````

Merging Or-Pattern branches:
```python
match m:
    case [x] | {'value': x}:
        pass
```

Currently I don't actually use literal_hash for this, as I only learned about it's existence later. I instead manually compare the Var of the NameExpr, as I know that TypeMaps produced by match patterns contain only NameExpr with Var as node. This uses a bunch of ugly asserts though (https://github.com/python/mypy/blob/5228bc7c64d59ccccea736e134707a180571a317/mypy/checkpattern.py#L464)

Note that literal_hash also just maps NameExpr to their node, so the result is the same.


I'm not sure if there really are any checks that check literal_hash against None. One common pattern I saw was
```python
if literal(x) == LITERAL_YES:
    key = literal_hash(x)
    assert key is not None
```
Is there any case where it would make sense to check `literal_hash(x) is not None` instead of `literal(x) == LITERAL_YES`?

But I don't have enough experience with the codebase to know if the expression_hash idea would really work.

JukkaL:
> This uses a bunch of ugly asserts though

I don't think that it's a problem, if the rest of the code is pretty clean. Returning `Optional[Var]` (`None` on failure) might be slightly cleaner -- though you'd probably need a `None` check.

> Is there any case where it would make sense to check literal_hash(x) is not None instead of literal(x) == LITERAL_YES?

To be honest, I don't remember the details of this code. Maybe `LITERAL_TYPE` makes things more complicated? 

> But I don't have enough experience with the codebase to know if the expression_hash idea would really work.

My initial impression is that it may be better to try this refactoring after the match statement PR has been merged. That way it would be easier to evaluate the pros and cons, as we'd see the impact to all affected code.

PR #10355, opened 2021-04-22
Require setuptools, test-data/packages/typedpkg_ns/typedpkg_ns/__init__.py imports pkg_resources

hrnciar:
### Description
I found there is a missing runtime dependency on setuptools so I propose to add it.


ethanhs:
Hi, thank you for the PR. The `pkg_resources` import is part of tests, so it is a test dependency. We already list it here: https://github.com/python/mypy/blob/master/test-requirements.txt#L17, so I am going to close this

PR #10362, opened 2021-04-23
Use Python 3.9 for mypy self check

srittau:
Using obsolete Python 3.5 breaks typeshed's execution of mypy's self
check, as typeshed does not support Python 3.5 anymore.

Cf. python/typeshed#5242

### DIFF
--- a/mypy_self_check.ini
+++ b/mypy_self_check.ini
@@ -18,5 +18,5 @@ show_error_codes = True
 pretty = True
 always_false = MYPYC
 plugins = misc/proper_plugin.py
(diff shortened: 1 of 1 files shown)


TH3CHARLie (approved):

JukkaL:
I think that we should use Python 3.6 instead as the target -- we want to use the oldest Python target for self check to avoid depending on features added in later Python versions.

srittau:
Using 3.9 actually causes another problem with the typeshed's mypy self check. I will send another PR.

Issue #9940, opened 2021-01-22
mypy section now required in setup.cfg differs from documentation


electrofelix:
**Bug Report**

Existing setup.cfg with a section that looks like `[mypy-myapp.mymodule]` containing an `ignore_errors = True` setting is no longer picked up as expected by mypy 0.800.

The behaviour documented at https://mypy.readthedocs.io/en/stable/config_file.html#config-file-format indicates that a section `[mypy]` is optional for `setup.cfg` files. However with a recent change it is now required.

PR #9114 introduces a change to skip setup.cfg to avoid another issue where use of setup.cfg was shadowing any additional user configs covered in #9113.

To maintain compatibility with existing configs potentially the check should be:
```
    if (config_file in defaults.SHARED_CONFIG_FILES and
            not any(section.startswith("mypy-") for section in parser.sections()))
        ....
```

It may be preferred to deprecate the existing documented behaviour, and be able to rely on needing there always to be a `mypy` section instead of treating setup.cfg as special. In the short term however it might be better to add a deprecation notice when this is detected and retain compatibility with existing setup.cfg that do not include an `[mypy]` section for now.

**To Reproduce**

(Write your steps here:)

1. Install mypy locally or via virtualenv
2. clone https://github.com/python/mypy
3. cd to mypy/test-data/packages/typedpkg
4. Run mypy to confirm detect error
   ```
   mypy -p typedpkg
   ```
5. Create a setup.cfg that should ignore these errors and execute mypy:
   ```
   cat <<EOF > setup.cfg
   [mypy-typedpkg.sample]
   ignore_errors = True
   EOF
   mypy -p typedpkg
   ```
6. Replace with a setup.cfg that contains an empty `[mypy]`
   ```
   cat <<EOF > setup.cfg
   [mypy]

   [mypy-typedpkg.sample]
   ignore_errors = True
   EOF
   mypy -p typedpkg
   ```

**Expected Behavior**

Error should be produce for the first execution of mypy, and success reported for the latter two

Output from step 4:
```
typedpkg/sample.py:7: error: Incompatible return value type (got "List[str]", expected "Tuple[str, ...]")
Found 1 error in 1 file (checked 5 source files)
```

Output from step 5:
```
Success: no issues found in 5 source files
```

Output from step 6:
```
Success: no issues found in 5 source files
```

**Actual Behavior**

Output from step 4:
```
typedpkg/sample.py:7: error: Incompatible return value type (got "List[str]", expected "Tuple[str, ...]")
[...]

PR #10364, opened 2021-04-24
Remove invalid note in docs: setup.cfg *does* need [mypy] section

jamescooke:
### Description

As per https://github.com/python/mypy/issues/9940, a `[mypy]` section is now required in `setup.cfg` in order for config to be loaded from that file. However, the docs are out of date.

This PR removes that out of date clause.

I'm not sure if this fixes the linked issue, but I know that removing this clause would have saved me a few hours of head banging - and therefore would save others too.

## Test Plan

GitHub renders the change OK, so I'm assuming sphinx will also build it fine.


TH3CHARLie (approved):
Sorry for the late review. Thanks for updating the docs!

Issue #2220, opened 2016-10-05
Mypy doesn't recognize a class as iterable when it implements only `__getitem__`


akaptur:
The following code runs fine, but mypy produce an error:
``` python
from typing import List

class Foo(object):
    def __init__(self):
        self.version = [1, 2, 3]

    def __getitem__(self, k):
        return self.version[k]


def list_foo(f):
    # type: (Foo) -> List[int]
    return list(f)

print list_foo(Foo())
```
```
example.py: note: In function "list_foo":
example.py:13: error: No overload variant of "list" matches argument types [example.Foo]
Fail
```

gvanrossum:
We should support this. It is part of the Python spec for `iter()`: https://docs.python.org/3/library/functions.html#iter

> Without a second argument, object must be a collection object which supports the iteration protocol (the __iter__() method), _or it must support the sequence protocol (the __getitem__() method with integer arguments starting at 0)_. [emphasis mine]

JukkaL:
We currently require an explicit `Iterable` base class, as mypy doesn't have structural subtyping. A potential quick fix would be to not complain about unimplemented `__iter__` in a subclass of `Iterable` if there is a suitable `__getitem__`.

A better fix would require structural subtyping (or protocols), but that's going to be harder. And we'd need another special case for `__getitem__`.

chadrik:
I ran into this problem with [`sre_parse.SubPattern`](https://github.com/python/cpython/blob/master/Lib/sre_parse.py#L109) using mypy 0.580.  Is there an adjustment that still needs to be made to structural subtyping to recognize this, or should the [stubs](https://github.com/python/typeshed/blob/master/stdlib/3/sre_parse.pyi#L45) be "fixed" for objects like these?

pmhahn:
Same problem for Exceptions in Python2:
```python
d = {0: 0}
try:
  d[1]
except KeyError as ex:
  (key,) = ex
# error: 'builtins.KeyError' object is not iterable
```
I can work-around that by explicitly accessing `ex.args`, but I still get that error for existing code.
But as exceptions in Python3 no longer work like this, I will have to update that code anyway in the future.

Recursing:
Has there been any progress on this issue?

glesica:
This problem still appears to exist in version 0.790. Specifically, the `sorted` function appears not to like getting an `ItemsView` instance even though it is technically iterable and the whole thing works at runtime.

```python
from typing import Dict, List

[...]

PR #10386, opened 2021-04-30
Make classes with __getitem__ work in for context

hatal175:
### Description

The for analysis searches for __iter__ to decide if it is iterable. I've added a search for __getitem__ as well.
This is a step in the right direction for #2220 but it doesn't really make classes with __getitem__ be accepted as arguments marked as Iterable - I'm not sure how to solve this without being intrusive in either typeshed or mypy type inference.

Please note that I've made it work with __getitem__ that accepts int (as the docs say) so if you have **only** __getitem__ with string, this will generate an error.

## Test Plan

I've added a few tests and mypy primer seems to have fewer errors :)

### DIFF
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -39,7 +39,8 @@ from mypy.types import (
 from mypy.sametypes import is_same_type
 from mypy.messages import (
     MessageBuilder, make_inferred_type_note, append_invariance_notes, pretty_seq,
-    format_type, format_type_bare, format_type_distinctly, SUGGESTED_TEST_FIXTURES
+    format_type, format_type_bare, format_type_distinctly, SUGGESTED_TEST_FIXTURES,
+    temp_message_builder
 )
 import mypy.checkexpr
 from mypy.checkmember import (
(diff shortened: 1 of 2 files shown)


mypy/checker.py
```diff
@@ -3499,20 +3500,40 @@ def analyze_iterable_item_type(self, expr: Expression) -> Tuple[Type, Type]:
         """Analyse iterable expression and return iterator and iterator item types."""
```
sobolevn, line 3500: Let's explain how iterables are handled in CPython. Priorities and possibilities.

mypy/checker.py
```diff
@@ -3499,20 +3500,40 @@ def analyze_iterable_item_type(self, expr: Expression) -> Tuple[Type, Type]:
...
+                expr,
+                getitem_msg_builder
+                )[0]
+
+            if not getitem_msg_builder.is_errors():
+                # We found __getitem__
+                return self.named_generic_type("typing.Iterator", [getitem_type]), getitem_type
+
+            self.msg.add_errors(iter_msg_builder)
+        # Non-tuple iterable.
```
sobolevn, line 3531: This comment does not really explain what is going on anymore.

mypy/checker.py
```diff
@@ -3499,20 +3500,40 @@ def analyze_iterable_item_type(self, expr: Expression) -> Tuple[Type, Type]:
...
+                iterable,
+                [arg],
+                [nodes.ARG_POS],
+                expr,
+                getitem_msg_builder
+                )[0]
+
+            if not getitem_msg_builder.is_errors():
+                # We found __getitem__
+                return self.named_generic_type("typing.Iterator", [getitem_type]), getitem_type
```
sobolevn, line 3528: Looks like it should be `typing.Iterable` and not `Iterator`.

test-data/unit/check-protocols.test
```diff
@@ -1940,6 +1940,17 @@ reveal_type(list(c for c in C))  # N: Revealed type is "builtins.list[__main__.C
 reveal_type(list(C))  # N: Revealed type is "builtins.list[__main__.C*]"
 [builtins fixtures/list.pyi]
 
+[case testIterableGetItemOnClass]
+class A:
```
sobolevn, line 1944: Let's also test:
- Metaclasses with `__getitem__` 
- `for` statements
- `list(iter(A()))` and `list(iter(B()))`

sobolevn (commented):
Thank you!

mypy/checker.py
```diff
@@ -3499,20 +3500,40 @@ def analyze_iterable_item_type(self, expr: Expression) -> Tuple[Type, Type]:
...
 
         if isinstance(iterable, TupleType):
             joined = UninhabitedType()  # type: Type
             for item in iterable.items:
                 joined = join_types(joined, item)
             return iterator, joined
+
+        if iter_msg_builder.is_errors():
+            # We couldn't find __iter__ so let's try __getitem__
+            getitem_msg_builder = temp_message_builder()
```
sobolevn, line 3515: You can try to use `.clean_copy()` here.

PR #10389, opened 2021-04-30
Allow booleans to be narrowed to literal types

### LINKED ISSUE #6113: Treat bool as equivalent to Literal[True, False] to allow better type inference
Similar to this https://github.com/python/mypy/issues/6027, but more basic use case.

```python
from typing import Collection, Union, List


def foo(bar: Union[bool, Collection[int]]) -> Union[int, List[int]]:
    if bar is False:  # or   if not bar:
        return 10

    if bar is True:
        return 20

    # otherwise, bar must be a collection (list, tuple, ...)
    return [i*2 for i in bar]


print(foo(True))
print(foo(False))
print(foo([1, 2, 3]))
print(foo((1, 2, 3,)))
```

Mypy is giving the error

```error: Item "bool" of "Union[bool, Collection[int]]" has no attribute "__iter__" (not iterable)```

But it should be able to infer that it can't be a `bool`?
using if / elif /else also gives the error.

### LINKED ISSUE #11043: Incompatible types: `Union[bool, A]` and `Optional[A]`
**Bug Report**

```python
from typing import Union, Optional


class Foo:
    ...
    
    
class Bar(Foo):
    ...
    
    
class Thing:
    thing: Optional[Foo]
    
    def errors(self, x: Union[bool, Foo]) -> None:
        if x is True:
            self.thing = Bar()
        elif x is False:
            self.thing = None
        else:
            self.thing = x
            
    def works(self, x: Union[bool, Foo]) -> None:
        if isinstance(x, bool):
            self.thing = Bar() if x else None
        else:
            self.thing = x
```

I would expect that `errors` and `works` are equal, but `errors` produces an error.

Sorry if this is a duplicate, I tried googling it!

**To Reproduce**

https://mypy-play.net/?mypy=latest&python=3.10&gist=f17cd6e5962f2edd358143dcce83ab51

**Expected Behavior**

No error

**Actual Behavior**

```python
main.py:21: error: Incompatible types in assignment (expression has type "Union[bool, Foo]", variable has type "Optional[Foo]")
```

**Your Environment**

See link above - can be reproduced in all versions

ethan-leba:
### Description

This PR allows boolean types to be narrowed to literals, in the same fashion
that Enums can be narrowed.

Closes #6113, closes #11043

## Test Plan

New unit tests included!


mypy/checker.py
```diff
@@ -4534,10 +4535,11 @@ def refine_identity_comparison_expression(self,
         if singleton_index == -1:
             singleton_index = possible_target_indices[-1]
 
-        enum_name = None
+        sum_type_name = None
         target = get_proper_type(target)
-        if isinstance(target, LiteralType) and target.is_enum_literal():
-            enum_name = target.fallback.type.fullname
+        if (isinstance(target, LiteralType) and
+                (target.is_enum_literal() or target.fallback.type.fullname == "builtins.bool")):
```
ethan-leba, line 4541: is there a better way to check if the type is a bool than looking for this string?

JukkaL: `isinstance(target.value, bool)`?

test-data/unit/check-narrowing.test
```diff
@@ -1027,5 +1027,27 @@ if str_or_bool_literal is not True and str_or_bool_literal is not False:
...
+[builtins fixtures/primitives.pyi]
+
+[case testNarrowingBoolean]
+# flags: --strict-optional
+from typing import Optional
+from typing_extensions import Literal
+
+bool_val: bool
 
+if bool_val is not False:
```
JukkaL, line 1039: Should `if bool_val` and `if not bool_val` also do narrowing? These can't be overridden for `bool`.

JukkaL (commented):
Thanks for the PR! Left some comments (not a full review).

Note that since we'll hopefully have a release soon, we probably won't merge this PR until after the release branch has been cut, since changes to type inference are a bit risky just before a release.

test-data/unit/check-narrowing.test
ethan-leba: Yeah, that seems useful, i'll look into it 👍

JelleZijlstra:
The homeassistant line 414 error seems like a bug. The code is at https://github.com/home-assistant/core/blob/dev/homeassistant/config_entries.py#L403. There is an assertion that `result` is a bool, but somehow mypy still thinks it's Any according to the error message.

ethan-leba:
> The homeassistant line 414 error seems like a bug. The code is at https://github.com/home-assistant/core/blob/dev/homeassistant/config_entries.py#L403. There is an assertion that `result` is a bool, but somehow mypy still thinks it's Any according to the error message.

Hoom, looks like this bug occurs on master with enums as well.

```
from enum import Enum

class Foo(Enum):
    A = 1
    B = 2

z: Any

assert isinstance(z, Foo)
reveal_type(z)

if z is Foo.A:
    reveal_type(z)
reveal_type(z)
```

```
note: Revealed type is "scratch.Foo"
note: Revealed type is "Literal[scratch.Foo.A]"
note: Revealed type is "Any"
```

Didn't make much progress after briefly investigating this bug. Any ideas what places
might be good to look in the codebase?

ethan-leba:
> The homeassistant line 414 error seems like a bug. The code is at https://github.com/home-assistant/core/blob/dev/homeassistant/config_entries.py#L403. There is an assertion that `result` is a bool, but somehow mypy still thinks it's Any according to the error message.

It looks the error is stemming from the type binder here:
https://github.com/python/mypy/blob/master/mypy/binder.py#L201

A (narrowed) Any type which does not have the same type on each side of a
conditional will be converted back to Any. Not sure about the context of this
decision, is this something we want to change?

ethan-leba:
@JukkaL ping

ethan-leba:
Hi @hauntsaninja, is there any interest in re-reviewing this PR? If so, I can rebase, etc.

hauntsaninja:
Thanks, yes! This PR is great and something I'm very interested in merging, although I would feel more comfortable doing so if @JukkaL had time to okay.

I fixed the broken test and merge conflict, hopefully it's enough to get CI green.

hauntsaninja:
Thanks again for this PR, super useful!

Dreamsorcerer:
This change seems to break basic object-oriented programming (but, could be a wider issue with type narrowing). e.g.
```
assert not thing.frozen
thing.freeze()
assert thing.frozen  # Mypy: thing.frozen is Literal[False]
```

I'd suggest disabling this feature for attributes, or if not too complex, reset the type narrowing after a method of that object is called.

PR #10404, opened 2021-05-03
Fix a bunch of tests broken by python 3.10

freundTech:
Python 3.10 changes a bunch of parser error messages. This commit introduces the version>= argument to [out] sections of tests, which allows testing against the new messages on newer python versions.

[out] sections are overwritten by other [out] sections following them.

I also thought about supporting other comparisons such as `<`, `==` and others, while not having later out sections overwrite earlier out sections, but decided on this approach for simplicity.

This was tested on `Python 3.10.0a7+ (heads/master:0cad068, Apr 30 2021, 14:45:23)` (shortly before 3.10.0b1). Note that Travis CI nighly is still stuck on  3.10.0a5+.


JukkaL (approved):
Thanks! This is a nice approach for dealing with differences between Python versions. Looks good.

PR #10406, opened 2021-05-03
Fixed a teststubtest test failing in python 3.10

freundTech:
I couldn't track down the exact change that caused this test to fail, but python now stores the full path to the module in `test_module.bad.__code__.co_filename` instead of just the filename, despite the file being in the current working directory.

Because the working directory is unknown where the output is actually checked we just remove it from the output. This only affects tests.


JelleZijlstra (approved):

hauntsaninja:
Thanks!

PR #10410, opened 2021-05-04
Use typeddict-item error code for more errors

JukkaL:
Previously this was inconsistent, and some TypedDict operations used
the 'misc' error code, which is dangerous to ignore. TypedDict related
errors seem to be some of most commonly ignored error with the 'misc'
error code.


ilevkivskyi (approved):
Maybe add a short description for `object_type` to one of functions docstrings where it was added?

JukkaL:
> Maybe add a short description for object_type to one of functions docstrings where it was added?

Added a reference to another docstring where it's described to avoid repetition.

PR #10412, opened 2021-05-04
Fix inference of IntEnum value attribute type

### LINKED ISSUE #10411: IntEnum with NewType value type regression
The type of `E.A.value` in this program has changed since the last release:

```py
from typing import NewType
from enum import IntEnum

N = NewType("N", int)

class E(IntEnum):
    A = N(0)

# mypy 0.812: N
# master:     int
reveal_type(E.A.value)
```

The original inferred type `N` is more precise, so this looks like a regression. This causes false positives with mypy-protobuf.

JukkaL:
If the enum value initializer has a NewType type, it should be
reflected in the type of the `value` attribute. It was broken because
the special casing for `__new__` introduced in #10057 didn't consider
the `__new__` in `IntEnum` as special.

Fixes #10411.

### DIFF
--- a/mypy/plugins/enums.py
+++ b/mypy/plugins/enums.py
@@ -112,7 +112,7 @@ def _implements_new(info: TypeInfo) -> bool:
     type_with_new = _first(ti for ti in info.mro if ti.names.get('__new__'))
     if type_with_new is None:
         return False
-    return type_with_new.fullname not in ('enum.Enum', 'enum.IntEnum')
+    return type_with_new.fullname not in ('enum.Enum', 'enum.IntEnum', 'enum.StrEnum')
 
 
 def enum_value_callback(ctx: 'mypy.plugin.AttributeContext') -> Type:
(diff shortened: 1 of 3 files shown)


ilevkivskyi (approved):

mypy/plugins/enums.py
```diff
@@ -112,7 +112,7 @@ def _implements_new(info: TypeInfo) -> bool:
     type_with_new = _first(ti for ti in info.mro if ti.names.get('__new__'))
     if type_with_new is None:
         return False
-    return type_with_new.fullname != 'enum.Enum'
+    return type_with_new.fullname not in ('enum.Enum', 'enum.IntEnum')
```
JelleZijlstra, line 115: Maybe also consider StrEnum, which is being added in 3.10
```suggestion
    return type_with_new.fullname not in ('enum.Enum', 'enum.IntEnum', 'enum.StrEnum')
```

PR #10418, opened 2021-05-04
Fix TypedDict test case on Python 3.5

JukkaL:
Class-based TypedDict syntax requires Python 3.6+.


JukkaL:
Not going to wait for the builds to finish, since the Python 3.5 wheel build is already broken.

PR #10444, opened 2021-05-07
[mypyc] Match evaluation order of multiple assignment from iterable (#793)

ChetanKhanna:
### Description

Refactored to iterate rvalues first before performing any assignments to match with Python semantics.

Closes mypyc/mypyc#793.

## Test Plan

Added a new test case to test the unpacking with `*x` type lvalues

### DIFF
--- a/mypyc/test-data/irbuild-statements.test
+++ b/mypyc/test-data/irbuild-statements.test
@@ -443,9 +443,9 @@ L0:
 def from_any(a):
     a, r0, r1 :: object
     r2 :: bool
-    x, r3 :: object
+    r3 :: object
     r4 :: bool
-    y, r5 :: object
+    x, y, r5 :: object
     r6 :: bool
 L0:
     r0 = PyObject_GetIter(a)
@@ -455,13 +455,13 @@ L1:
     r2 = raise ValueError('not enough values to unpack')
     unreachable
 L2:
-    x = r1
     r3 = PyIter_Next(r0)
     if is_error(r3) goto L3 else goto L4
 L3:
     r4 = raise ValueError('not enough values to unpack')
     unreachable
 L4:
+    x = r1
     y = r3
     r5 = PyIter_Next(r0)
     if is_error(r5) goto L6 else goto L5
@@ -504,9 +504,9 @@ L0:
 def from_any(a):
     a, r0, r1 :: object
     r2 :: bool
-    r3, x :: int
-    r4 :: object
-    r5 :: bool
+    r3 :: object
+    r4 :: bool
+    r5, x :: int
     y, r6 :: object
     r7 :: bool
 L0:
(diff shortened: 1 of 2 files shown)


mypyc/test-data/irbuild-statements.test
```diff
@@ -607,6 +607,91 @@ L5:
...
+    r7 :: bool
+    r8 :: str
+    r9 :: int32
+    r10 :: bit
+    r11 :: str
+    r12 :: int32
+    r13 :: bit
+    r14 :: list
+    r15 :: ptr
+    r16 :: int64
```
ChetanKhanna, line 634: I tested this on lots different test cases from both this branch and using master. Each time it generates a register whose size is platform dependent (int64 for 64 bit platform and int32 otherwise)  This is actually decided by `c_pyssize_t_rprimitive` inside `ll_builder.py.builtin_len()`. Not sure how to write a system independent irbuild test case for it :thinking:

TH3CHARLie: ummm that's due to https://github.com/mypyc/mypyc/issues/776. A quick fix to pass the CI is to change all `int64` in your generated IR output into `native_int`.

ChetanKhanna: Thanks, it works now :)

mypyc/test-data/irbuild-statements.test
```diff
@@ -607,6 +607,91 @@ L5:
 L6:
     return 1
 
+[case testStarUnpack]
+from typing import Any, List, Iterator
```
97littleleaf11, line 611: testMultiAssignmentStarUnpack might be a better name?

mypyc/test-data/irbuild-statements.test
```diff
@@ -607,6 +607,91 @@ L5:
...
+    r7 = raise ValueError('not enough values to unpack')
+    unreachable
+L4:
+    r8 = 'x'
+    r9 = PyObject_SetAttr(a, r8, r4)
+    r10 = r9 >= 0 :: signed
+    r11 = 'y'
+    r12 = PyObject_SetAttr(a, r11, r6)
+    r13 = r12 >= 0 :: signed
+    r14 = PySequence_List(r3)
```
JukkaL, line 671: Hmm should we do this before settings any attributes as well? It seems that we only partially evaluate the values to assign ahead of time if using star unpacking.

ChetanKhanna: Hi @JukkaL , sorry about the extreme delay in addressing the review comment :) I have only recently resumed contributing to mypy. 

I have now refactored the code to evaluate all values before any assignment takes place.

(Not sure if the PR or the issue is still relevant, I was just trying to pick up where I left)

PR #10446, opened 2021-05-08
[mypyc] Emit native_int instead of int64/int32 in IR test output 

TH3CHARLie:
relates https://github.com/mypyc/mypyc/issues/776

This is a super quick workaround. So far the issue mostly (if not completely) occurs in `c_pyssize_t` so making this rtype emits `native_int` would solve the issue. As a result, the `replace_native_int` pass in IR test is now removed.


97littleleaf11:
lgtm

JukkaL (approved):
Very nice! Confusion between different integer types in tests has been a common pain point. This will reduce friction when writing new features and changing the generated IR significantly.

PR #10448, opened 2021-05-08
[mypyc] Modify specializers and add int.__pow__ in test-data/fixture

97littleleaf11:
### Description

* Move simple `tuple` and `list` creation to new functions. Each specializer only focus on one situation now.
* Add `int.__pow__` in `test-data/fixture`

## Test Plan

This change should pass all existing tests.


97littleleaf11:
The following test case leads to a test failure:

```py
source_e = [0, 1, 2]
e = list((x ** 2) for x in (y + 2 for y in source_e))
assert e == [4, 9, 16]
```

Got this when testing 

```
Unsupported left operand type for ** ("int")
```

However I cannot reproduce this by directly compile it.

JukkaL (approved):
Thanks for the PR! Looks good. It's easier to understand specializers if each of them does one thing only.

PR #10453, opened 2021-05-09
Fix “they is” typo in Final docs

foolip:
### Description

Spotted in https://mypy.readthedocs.io/en/stable/final_attrs.html


hauntsaninja (approved):
Thanks!

foolip:
Thanks for these excellent docs! I've been reading it cover-to-cover and it's really amazing how few errors I've spotted, just two so far :)

PR #10458, opened 2021-05-11
Infer unreachability for await no_return()

### LINKED ISSUE #10454: Awaiting on `Awaitable[Tuple[NoReturn, ...]]`
**Bug Report**

Mypy raises "Implicit return in function which does not return  [misc]" when awaiting on `Awaitable[Tuple[NoReturn, ...]]` (and similar) inside `NoReturn` coroutine.

**To Reproduce**

1) Run Mypy against following example:

```
from typing import TypeVar, NoReturn, Tuple, Awaitable

T1 = TypeVar('T1')
T2 = TypeVar('T2')


async def gather(
    f1: Awaitable[T1], f2: Awaitable[T2]
) -> Awaitable[Tuple[T1, T2]]:
    raise NotImplementedError


async def no_return() -> NoReturn:
    raise TypeError


async def main() -> NoReturn:
    await gather(no_return(), no_return())
```

2) Mypy will raise "Implicit return in function which does not return"

**Expected Behavior**

Here I expect that awaiting on `Awaitable[Tuple[NoReturn, ...]]` will be threated exactly same as awaiting on plain `Awaitable[NoReturn]`.

**Actual Behavior**

I don't know how to inspect inferred types, but almost sure that `gather(no_return(), no_return())` are inferred correctly as `Awaitable[Tuple[NoReturn, NoReturn]]`, but Mypy doesn't treat it same as `Awaitable[NoReturn]`.

**Your Environment**

- Mypy version used: 0.812
- Mypy command-line flags: no additional flags required to reproduce the bug
- Mypy configuration options from `mypy.ini` (and other config files): same as for field above
- Python version used: 3.9.0
- Operating system and version: Ubuntu 20.04

hauntsaninja:
Fixes #10454

The logic here is similar to what we do in visit_call_expr_inner

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -4042,7 +4042,10 @@ class ExpressionChecker(ExpressionVisitor[Type]):
             return AnyType(TypeOfAny.special_form)
         else:
             generator = self.check_method_call_by_name('__await__', t, [], [], ctx)[0]
-            return self.chk.get_generator_return_type(generator, False)
+            ret_type = self.chk.get_generator_return_type(generator, False)
+            if isinstance(ret_type, UninhabitedType) and not ret_type.ambiguous:
+                self.chk.binder.unreachable()
+            return ret_type
 
     def visit_yield_from_expr(self, e: YieldFromExpr, allow_none_return: bool = False) -> Type:
         # NOTE: Whether `yield from` accepts an `async def` decorated
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -4043,6 +4043,7 @@ class ExpressionChecker(ExpressionVisitor[Type]):
         else:
             generator = self.check_method_call_by_name('__await__', t, [], [], ctx)[0]
             ret_type = self.chk.get_generator_return_type(generator, False)
+            ret_type = get_proper_type(ret_type)
             if isinstance(ret_type, UninhabitedType) and not ret_type.ambiguous:
                 self.chk.binder.unreachable()
             return ret_type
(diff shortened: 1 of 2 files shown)


mypy/checkexpr.py
```diff
@@ -4042,7 +4042,10 @@ def check_awaitable_expr(self, t: Type, ctx: Context, msg: str) -> Type:
             return AnyType(TypeOfAny.special_form)
         else:
             generator = self.check_method_call_by_name('__await__', t, [], [], ctx)[0]
-            return self.chk.get_generator_return_type(generator, False)
+            ret_type = self.chk.get_generator_return_type(generator, False)
+            if isinstance(ret_type, UninhabitedType) and not ret_type.ambiguous:
```
TH3CHARLie, line 4046: expand `ret_type` before `isinstance`

test-data/unit/check-flags.test
```diff
@@ -334,6 +334,20 @@ from mypy_extensions import NoReturn
...
+from mypy_extensions import NoReturn
+
+async def f() -> NoReturn: ...
+
+async def g() -> NoReturn:
+  await f()
+
+async def h() -> NoReturn:  # E: Implicit return in function which does not return
+  f()
+[builtins fixtures/dict.pyi]
```
TH3CHARLie, line 348: I guess `dict.pyi` here is for `isinstance`

TH3CHARLie (approved):
LGTM! Just have one minor confusion that would not block merging this.

test-data/unit/check-flags.test
TH3CHARLie: I see. Thanks for the clarification.

PR #10460, opened 2021-05-12
Fix several typos 

ArjunSahlot:
### Description
More typo fixes and 1 code change and 1 suggestion.

## Test Plan
No testing required.

### DIFF
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -332,7 +332,6 @@ class MessageBuilder:
                 # The checker passes "object" in lieu of "None" for attribute
                 # checks, so we manually convert it back.
                 typ_format, orig_type_format = format_type_distinctly(typ, original_type)
-                # Not fully sure but I think we can replace "type(item) == NoneType" with "item is None"
                 if typ_format == '"object"' and \
                         any(type(item) == NoneType for item in original_type.items):
                     typ_format = '"None"'
--- a/mypy/messages.py
+++ b/mypy/messages.py
@@ -332,6 +332,7 @@ class MessageBuilder:
                 # The checker passes "object" in lieu of "None" for attribute
                 # checks, so we manually convert it back.
                 typ_format, orig_type_format = format_type_distinctly(typ, original_type)
+                # Not fully sure but I think we can replace "type(item) == NoneType" with "item is None"
                 if typ_format == '"object"' and \
                         any(type(item) == NoneType for item in original_type.items):
                     typ_format = '"None"'
--- a/mypy/mro.py
+++ b/mypy/mro.py
@@ -43,7 +43,7 @@ def linearize_hierarchy(info: TypeInfo,
 
 
 def merge(seqs: List[List[TypeInfo]]) -> List[TypeInfo]:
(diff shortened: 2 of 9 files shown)


mypy/messages.py
```diff
@@ -332,6 +332,7 @@ def has_no_attr(self,
                 # The checker passes "object" in lieu of "None" for attribute
                 # checks, so we manually convert it back.
                 typ_format, orig_type_format = format_type_distinctly(typ, original_type)
+                # Not fully sure but I think we can replace "type(item) == NoneType" with "item is None"
```
JelleZijlstra, line 335: Nope, this is mypy's internal NoneType, not the type of None.

mypy/mro.py
```diff
@@ -43,7 +43,7 @@ def linearize_hierarchy(info: TypeInfo,
 
 
 def merge(seqs: List[List[TypeInfo]]) -> List[TypeInfo]:
-    seqs = [s[:] for s in seqs]
+    seqs = [s.copy() for s in seqs]
```
JelleZijlstra, line 46: Seems unnecessary

JelleZijlstra (changes requested):
Thanks, most of these changes look good

mypy/messages.py
hauntsaninja, line 335: We can't... This is https://github.com/python/mypy/blob/fffbe88fc54807c8b10ac40456522ad2faf8d350/mypy/types.py#L698 , not `type(None)` (we could maybe use isinstance instead of `type(x) == ...`, but I'd prefer to just leave this alone)

mypy/mro.py
hauntsaninja, line 46: This change isn't necessary. The slice is a common idiom

ArjunSahlot: Yeah, doesn't really change anything. I thought it would make more readable though.

hauntsaninja (commented):
Thanks, mostly looks good!

hauntsaninja:
Oh lol, Github review race. Never mind me.

ArjunSahlot:
Oh lol, its fine. I think I have fixed the issues now.

JelleZijlstra:
> Oh lol, Github review race. Never mind me.

At least we said the same things :)

JelleZijlstra (approved):

ArjunSahlot:
Does anyone know why [this commit](https://github.com/python/mypy/pull/10460/commits/2d2478b944dd428101d56f37baf574541976471c) didn't pass the tests/github actions?

JelleZijlstra:
> Does anyone know why this commit didn't pass the tests/github actions?

https://travis-ci.com/github/python/mypy/jobs/504824045 has some errors around git fetch; probably a race conditions due to new commits.

ArjunSahlot:
Ok, I just wanted to make sure it wasn't a problem.

mypy/messages.py
ArjunSahlot: I think PEP recommends using `isinstance` over `type() == `. Would that be a better option?

PR #10467, opened 2021-05-13
Add `__attrs_attrs__` and `fields` to attrs

Tinche:
My first contribution, so sorry if it's not super well-coded.

This improved the attrs plugin in two ways.

Firstly, it makes Mypy aware of the `__attrs_attrs__` class variable which attrs generates, and populates it with the proper `attr.Attribute`s. This is not generally used much by attrs users, but is useful for our version of `attr.fields`, and can be used to write attrs protocols.

Secondly, it modified the return value of `attr.fields` to return properly instantiated `attr.Attributes` (which are generic). So now, this type checks:

```
from typing import TypeVar
from attr import define, Attribute, fields as f

@define
class A:
    a: int

C = TypeVar("C")
T1 = TypeVar("T1")

def test_projection(model: Type[C], field: Attribute[T1]) -> T1:
    pass

reveal_type(test_projection(A, f(A).a))
```
```
t.py:28: note: Revealed type is "builtins.int*"
```

This makes it possible to create ORM/ODM frameworks based on attrs that support statically checked projections, for one.

This would be just the initial implementation for soliciting early feedback, with docs and tests to follow.

### DIFF
--- a/mypy/plugins/attrs.py
+++ b/mypy/plugins/attrs.py
@@ -1,5 +1,6 @@
 """Plugin for supporting the attrs library (http://www.attrs.org)"""
 
+from mypy.lookup import lookup_fully_qualified
 from mypy.ordered_dict import OrderedDict
 
 from typing import Optional, Dict, List, cast, Tuple, Iterable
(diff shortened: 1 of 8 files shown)


euresti:
This looks like a reasonable approach.  As you can tell from all the errors there are some issues.  I'm guessing there needs to be a lot of `isinstance` checks and lots of error handlings (e.g. I think `lookup_fully_qualified("attr.Attribute",...)` fails if you don't have the right stubs.  

Oh and lots of comments because the code is definitely not self explanatory.  :)

mypy/plugins/attrs.py
```diff
@@ -257,6 +258,14 @@ def _get_decorator_optional_bool_argument(
         return default
 
 
+def _make_var(n: str, i: Type, fullname: Optional[str] = None, is_classvar: bool = False) -> Var:
```
sobolevn, line 261: Maybe using `name` and `instance` as var names would be better?

mypy/plugins/attrs.py
```diff
@@ -319,6 +328,51 @@ def attr_class_maker_callback(ctx: 'mypy.plugin.ClassDefContext',
...
+                        [Instance(ctx.cls.info, []), a.argument(ctx).type_annotation]
+                        if len(attr_type.type_vars) == 2 else
+                        [a.argument(ctx).type_annotation]),
+                ),
+            )
+            for a in attributes
+        },
+    )
+    cd = ClassDef(f"{info.name}Attributes", Block([]))
+    cd.fullname = f"{info.name}Attributes"
```
sobolevn, line 356: Shouldn't it be `f"{info.fullname}Attributes"`? https://github.com/python/mypy/blob/master/mypy/nodes.py#L2471

Tinche:
I tried running the plugin when `attrs` itself was not installed:
```
t.py:1: error: Library stubs not installed for "attr" (or incompatible with Python 3.9)
t.py:1: note: Hint: "python3 -m pip install types-attrs"
t.py:1: note: (or run "mypy --install-types" to install all missing stub packages)
t.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
t.py:28: note: Revealed type is "<nothing>"
```
It doesn't crash Mypy and basically tells you what to do, so I think this case is fine.

Next, I tried deleting the stubs; `attr.Attribute` ceases to be generic but again, no crashes. (`lookup_fully_qualified` works, but returns the non-stub version.)

I'm going to add a check anyway, to future-proof just in case.

mypy/plugins/attrs.py
Tinche: Looking at the runtime behavior, looks like the inner class is actually just called `{name}Attributes`, and the `__module__` is 'builtins', always.

```python3
>>> import a
>>> a.A
<class 'a.A'>
>>> a.A.__attrs_attrs__
(Attribute(name='x', default=NOTHING, validator=None, repr=True, eq=True, eq_key=None, order=True, order_key=None, hash=None, init=True, metadata=mappingproxy({}), type=<class 'str'>, converter=<function converter at 0x7fcd03db2040>, kw_only=False, inherited=False, on_setattr=None),)
>>> a.A.__attrs_attrs__.__class__
<class 'AAttributes'>
>>> a.A.__attrs_attrs__.__module__
'builtins'
>>>
```

Tinche:
@euresti I have fixed almost all of the tests by changing the fixtures a little (they need `dict` now, for example).

I couldn't fix the incremental tests though. As part of creating the `__attrs_attrs__` attribute, I create a namedtuple subclass in the plugin (just like attrs does at runtime). The incremental tests cannot find this class, for some reason. Now I need to figure out why. Does it have to do with de/serialization?

Tinche:
Fixed by https://github.com/python/mypy/pull/8578

PR #10474, opened 2021-05-14
mypy_primer: use 'needs: mypy_primer' for comment job

Akuli:
### Description

Simplifies mypy_primer running. Port of https://github.com/python/typeshed/pull/5428

## Test Plan

https://github.com/Akuli/typeshed/pull/17


.github/workflows/mypy_primer.yml
```diff
@@ -48,10 +48,58 @@ jobs:
...
+            )
+
+            console.log("Diff from mypy_primer:")
+            console.log(data)
+
+            let body
+            if (data.trim()) {
+              body = 'Diff from [mypy_primer](https://github.com/hauntsaninja/mypy_primer), showing the effect of this PR on open source code:\n```diff\n' + data + '```'
+            } else {
+              body = 'According to [mypy_primer](https://github.com/hauntsaninja/mypy_primer), this change has no effect on the checked open source code. 🤖🎉'
```
hauntsaninja, line 91: nit: can we get rid of the no-op message? It can be noisy (e.g. for all the mypyc work), even with comment hiding, and no one on this repo has asked for such a thing yet.

Akuli: As it is now, this PR also introduces comment hiding to mypy. I'll delete that too, and it can be added back in a separate PR if someone wants it.

hauntsaninja: I think that's fine to keep — or at least, I want comment hiding! :D

Akuli: I'll make a separate PR once this is done

Akuli:
With no message for "diff is empty", the comment hider is a bit of a problem. In typeshed, I set it to hide everything except the last comment. But now it should hide all comments when the diff is empty, and everything except last when the diff is non-empty. This could be done by setting an output from the `github-script` step. Something like `console.log("::set-output foo=1")` (not sure about the syntax details).

hauntsaninja (approved):
That's a good point, it's fine to leave it then :-)

Akuli:
(I don't have merge permission on mypy)

hauntsaninja:
Thanks again!

PR #10475, opened 2021-05-14
Update README.md, MIT License link is not working

SamirPaul1:

(YES)

### Description

When I was going through the README.md of mypy, I saw that a link in the README.md file is not working. This is a link about the License of the project. Although all the links are working perfectly and anyone can just click on the link and see relevant informations.

The "LICENSE" should be embedded with the link  https://github.com/python/mypy/blob/master/LICENSE

---

Before Update

---

After Update

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,5 +276,5 @@ For more details, see the file [CONTRIBUTING.md](CONTRIBUTING.md).
 License
 -------
 
-Mypy is licensed under the terms of the MIT License (see the file
-LICENSE).
+Mypy is licensed under the terms of the MIT License, see the file[
+LICENSE](https://github.com/python/mypy/blob/master/LICENSE).


README.md
```diff
@@ -276,5 +276,5 @@ For more details, see the file [CONTRIBUTING.md](CONTRIBUTING.md).
 License
 -------
 
-Mypy is licensed under the terms of the MIT License (see the file
-LICENSE).
+Mypy is licensed under the terms of the MIT License, see the file[
+LICENSE](https://github.com/python/mypy/blob/master/LICENSE).
```
JukkaL, line 280: Nit: It would be better to have a space/newline before `[` (e.g. move `[` to the next line).

SamirPaul1: Yes, @JukkaL

PR #10476, opened 2021-05-15
mypy_primer: fix comment permissions

hauntsaninja:
Port https://github.com/python/typeshed/pull/5452
We didn't have the global permission set, so maybe it wasn't broken, but
explicitly setting permissions seems like a good thing


hauntsaninja:
cc @Akuli

Akuli (approved):

TH3CHARLie (approved):

PR #10478, opened 2021-05-15
Do not allow to use Final and ClassVar at same time

### LINKED ISSUE #10477: Variable can be annotated with Final and ClassVar at the same time
**Bug Report**

Variable can be annotated with `Final` and `ClassVar` at the same time.

But `PEP 591` says that this is forbidden:
>  Variables should not be annotated with both ClassVar and Final.

**To Reproduce**

1. Declarate class variable with both `Final` and `ClassVar` annotations.

```python
from typing_extensions import Final, ClassVar


class Foo:
    a: Final[ClassVar[int]] = 10
```

**Expected Behavior**

`mypy` should fail.

```
error: Variable should not be annotated with both ClassVar and Final
```

**Actual Behavior**

`mypy` do not raise any error. 

```
```

**Your Environment**

- Mypy version used: 0.812
- Python version used: 3.6.8
- Operating system and version: macOS Big Sur 11.2.1

uriyyo:
### Description

Fixes #10477

Do not allow to use Final and ClassVar at same time.

## Test Plan

Code bellow:

```python
from typing import Final, ClassVar

class A:
    a: Final[ClassVar[int]]
```

After this PR will produce an error:
```
error: Variable should not be annotated with both ClassVar and Final
```

But currently, it will not raise any error.

### DIFF
--- a/test-data/unit/check-final.test
+++ b/test-data/unit/check-final.test
@@ -1082,3 +1082,10 @@ class A:
         self.x = 10  # type: Final
         undefined  # type: ignore
 [builtins fixtures/tuple.pyi]
+
+[case testFinalUsedWithClassVar]
+from typing import Final, ClassVar
+
+class A:
+    a: Final[ClassVar[int]]  # E: Variable should not be annotated with both ClassVar and Final
+[out]
--- a/test-data/unit/check-final.test
+++ b/test-data/unit/check-final.test
@@ -1088,4 +1088,6 @@ from typing import Final, ClassVar
 
 class A:
     a: Final[ClassVar[int]]  # E: Variable should not be annotated with both ClassVar and Final
+    b: ClassVar[Final[int]]  # E: Final can be only used as an outermost qualifier in a variable annotation
+    c: ClassVar[Final] = 1  # E: Final can be only used as an outermost qualifier in a variable annotation
 [out]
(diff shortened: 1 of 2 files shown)


test-data/unit/check-final.test
```diff
@@ -1082,3 +1082,10 @@ class A:
         self.x = 10  # type: Final
         undefined  # type: ignore
 [builtins fixtures/tuple.pyi]
+
+[case testFinalUsedWithClassVar]
+from typing import Final, ClassVar
+
+class A:
+    a: Final[ClassVar[int]]  # E: Variable should not be annotated with both ClassVar and Final
```
JelleZijlstra, line 1090: Can you test some other variants, such as `a: ClassVar[Final] = 1`, `a: ClassVar[Final[int]]`?

uriyyo: Such cases will produce an error with message:
```
Final can be only used as an outermost qualifier in a variable annotation.
```

Should it be changed to `Variable should not be annotated with both ClassVar and Final`?

JelleZijlstra: If it's easy to do, then yes. The existing error message would make users first swap the qualifiers, then get an error anyway from your new check, which is not very user-friendly. But don't worry about changing the error message if it's complicated.

Either way I think it would be useful to test all these variations here. They can all go in the same test case.

uriyyo: Looks like it is not easy to change message for a `ClassVar[Final[int]]` case.

I will skip this for this moment.

JelleZijlstra (approved):
Looks good, but I'm going to hold off on merging because GitHub Actions are broken at the moment.

PR #10486, opened 2021-05-16
Remove assertion from TypeGuard serialization

JelleZijlstra:
This came up in python/typeshed#5473: mypy started crashing when
I made a function return TypeGuard[Type[Any]]. I wasn't able to
reproduce this locally so far in a standalone file, but verified that removing the
assertion fixes the crash.


JukkaL (approved):
LGTM! It may be possible to test this using incremental mode test cases somehow.

PR #10494, opened 2021-05-17
Allow running mypy if Python 2 typeshed stubs aren't installed

JukkaL:
This enables custom mypy configurations where only Python 3 is supported.

### DIFF
--- a/mypy/modulefinder.py
+++ b/mypy/modulefinder.py
@@ -749,7 +749,7 @@ def load_stdlib_py_versions(custom_typeshed_dir: Optional[str]
     """Return dict with minimum and maximum Python versions of stdlib modules.
 
     The contents look like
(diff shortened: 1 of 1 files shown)


ilevkivskyi (approved):

srittau:
You could actually split the Python 2 and 3 loading logic completely: You only need to check `VERSIONS` when checking Python 3, and only look at `@python2` when checking Python 2.

JukkaL:
> You could actually split the Python 2 and 3 loading logic completely: You only need to check VERSIONS when checking Python 3, and only look at @python2 when checking Python 2.

Yes, this would be possible once we no longer have shared Python 2/3 stdlib stubs.

JelleZijlstra:
> this would be possible once we no longer have shared Python 2/3 stdlib stubs

We made this change recently in typeshed already.

PR #10495, opened 2021-05-17
Sync typeshed

JukkaL:
Source commit:
https://github.com/python/typeshed/commit/17bc1d83ea5ba4c3fb8363259993047727caf6b3

This doesn't include some recent typeshed commits. In particular, this is from before
the split of stdlib stubs to Python 2 and 3 variants. This way we can more easily narrow
down issues it might have caused.


JukkaL:
Looks like GitHub is having some issues?

PR #10498, opened 2021-05-18
Fix spurious name undefined error in class body within import cycle

JukkaL:
This could sometimes happen with protobuf stubs. The issue is a quite
tricky one, since it only happens if files in an import cycle are
checked in a specific order.

The problem was with code like this, if `Foo` and `Base` are defined
in different files within an import cycle:

```
"""m.py"""
from m2 import Base

class Foo(Base):
    x: Bar  # <<-- Unexpected error: "Bar" undefined
    class Bar: ...
```

Due to the import cycle, `Base` could be a placeholder node when
semantically analyzing `m` for the first time. This caused another
pass over `m`. On the second pass `Bar` was repoted as undefined,
because of an incorrect namespace completeness check. We were checking
the completeness of the *module-level* namespace, when we should have
looked at the completeness of the *class* namespace.

If `Base` was ready during the first pass, the example worked as
expected, since neither the module nor the class namespace was
complete.

Errors about undefined things are only supposed to be generated when
the target namespace is complete (i.e., all names are included in
the symbol table, possibly as placholders).

This fixes the issue by keeping track of whether a class body is being
processed for the first time. During the first time the namespace is
being built, so it's incomplete.

This may not work in some very tricky scenarios where we need to
process the body of a class more than twice, but these cases are
probably very rare, so this fix should get us most of the way there.

### DIFF
--- a/mypy/semanal.py
+++ b/mypy/semanal.py
@@ -250,6 +250,8 @@ class SemanticAnalyzer(NodeVisitor[None],
         self.imports = set()
         self.type = None
         self.type_stack = []
+        # Are the namespaces of classes being processed complete?
+        self.incomplete_type_stack = []  # type: List[bool]
         self.tvar_scope = TypeVarLikeScope()
         self.function_stack = []
         self.block_depth = [0]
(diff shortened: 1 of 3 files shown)


ilevkivskyi (approved):

JukkaL:
Closing and reopening to rerun builds. Seems like GitHub actions is still being flaky.

hauntsaninja:
Just changed mypy_primer to use `--show-traceback` so the error should be more immediately helpful next time :-)

PR #10505, opened 2021-05-19
Port over latest GA mypy_primer changes

hauntsaninja:
This is mostly the same as typeshed. The biggest difference is that we
don't post the "no effect" message. As a side effect, we also have to
run the comment hider first.

### DIFF
--- a/.github/workflows/mypy_primer_comment.yml
+++ b/.github/workflows/mypy_primer_comment.yml
@@ -1,0 +1,100 @@
+name: Comment with mypy_primer diff
+
+on:
+  workflow_run:
+    workflows:
+      - Run mypy_primer
+    types:
+      - completed
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  comment:
+    name: Comment PR from mypy_primer
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download diffs
+        uses: actions/github-script@v3
+        with:
+          script: |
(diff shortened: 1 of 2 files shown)


hauntsaninja:
cc @Akuli , if you have time to take a look

.github/workflows/mypy_primer_comment.yml
```diff
@@ -0,0 +1,100 @@
...
+              .join('')
+              .substr(0, 30000)  // About 300 lines
+            )
+
+            console.log("Diff from mypy_primer:")
+            console.log(data)
+
+            let body
+            if (data.trim()) {
+              body = 'Diff from [mypy_primer](https://github.com/hauntsaninja/mypy_primer), showing the effect of this PR on open source code:\n```diff\n' + data + '```'
```
Akuli, line 93: Nit:

```suggestion
            if (data.trim()) {
              const body = 'Diff from [mypy_primer](https://github.com/hauntsaninja/mypy_primer), showing the effect of this PR on open source code:\n```diff\n' + data + '```'
```

Akuli (approved):
The only difference seems to be that comment hiding is before creating the new comment. A possible disadvantage is how failing to hide comments will then prevent the comment from being posted. If that's a concern, you can add something like `run: cat diff_*.txt` before the comment hiding, so that even if comment hiding fails, you can still read the diff without having to manually download and extract the artifact.

PR #10507, opened 2021-05-19
Include stubs for mypy_extensions in bundled typeshed

JukkaL:
Also don't install a stub package for `mypy_extensions`.

There are two reasons why we're doing this:

* We make assumptions about what's defined in `mypy_extensions`, and
  if users are free to modify it, various things may break, and this
  could even cause mypy crashes.

* Older mypy versions complain about installed stubs for
  `mypy_extensions`. Updating to an earlier mypy versions in the same
  environment is broken if we use a stub package.


srittau:
Would it make sense to bundle the mypy_extensions stub directly with mypy_extensions instead of having it in typeshed?

JukkaL:
> Would it make sense to bundle the mypy_extensions stub directly with mypy_extensions instead of having it in typeshed?

This has the same problem that updating to a more recent version of `mypy_extensions` could break mypy. By having the stub vendored, we are free to update the definitions in typeshed without having to worry about breaking older mypy versions. 

This is similar how we are also still shipping bundled `typing_extensions` stubs.

hauntsaninja:
This is problematic for `--custom-typeshed-dir`

hauntsaninja:
Forward compatibility if mypy_extensions ships with types might be less of an issue if mypy pins the mypy_extension version exactly (we already pin to < 0.5.0, so I think it could work if we released 0.5.0 and pinned it)

JukkaL:
We can include `typeshed/mypy-extensions` along with `typeshed/stdlib` under `mypy/`. This should fix `--custom-typeshed-dir`. I'd prefer to avoid the extra dependency. Alternatively, we can include `typeshed/mypy-extensions` in the search path when using `--custom-typeshed-dir`.

PR #10511, opened 2021-05-20
Double quotes in errors "Inconsistent use of * in function", "is a type variable and only valid in type context" and "Import of x ignored"

dixith:
### Description
This pull request fixes issue reported in #7445 for errors related to `Inconsistent use of * in function`, `is a type variable and only valid in type context`, `Import of x ignored`.

Also updated unit test data readme file(test-data/unit/README.md) for error `Name "zzz" already defined`

## Test Plan


hauntsaninja:
Thanks!

PR #10513, opened 2021-05-21
Double quotes in errors `Source file found twice`, `Duplicate module named`, `Setting "strict" not supported`, `bound by an outer class` and `cannot subclass`

dixith:
### Description
This pull request fixes issue reported in #7445 for errors related to `Source file found twice`, `Duplicate module named`, `Setting "strict" not supported`, `bound by an outer class` and `cannot subclass`

## Test Plan


JelleZijlstra (approved):
Thanks! How many more are left?

dixith:
> 
> 
> Thanks! How many more are left?

@JelleZijlstra Not many. I'll try to finish the rest in 2 or 3 PRs

Issue #5485, opened 2018-08-15
Class with function fields incorrectly thinks first argument is self


Dentosal:
This is a bug report.

# Minimal repro

```python
from dataclasses import dataclass
from typing import Callable

@dataclass
class Repro:
    fn: Callable[[int], str]

r = Repro(fn=repr)
assert "1" == r.fn(1)
```

Works correctly when dataclass is not used, i.e. when `Repro` is

```python
class Repro:
    def __init__(self, fn: Callable[[int], str]) -> None:
        self.fn = fn
```

## Actual behavior

```python
repro.py:9: error: Invalid self argument "Repro" to attribute function "fn" with type "Callable[[int], str]"
repro.py:9: error: Too many arguments
```

When running with Python, works correctly. Doesn't throw any exceptions.

## Expected behavior

Successful type check

# Environment

## Versions

Tested with two versions, 0.620 and current version from GitHub.

* Python 3.7.0
* mypy 0.620
* mypy 0.630+dev-a2ffde90635baf05e264f8d899f5705c52b27c2a

## Mypy config

```ini
[mypy]
warn_redundant_casts = True
warn_unused_ignores = True
incremental = True

follow_imports = normal
ignore_missing_imports = True
disallow_untyped_calls = True
disallow_untyped_defs = True
check_untyped_defs = True
warn_return_any = True
warn_no_return = True
no_implicit_optional = True
strict_optional = True
```

ilevkivskyi:
This is not something specific to dataclasses. There is a simpler repro that gives same errors:
```python
class Repro:
    fn: Callable[[int], str]

r = Repro()
r.fn(1)
```
The problem is the same as in https://github.com/python/mypy/issues/708. I would however _not_ close this as a duplicate, because this issue highlights another aspect of the problem.

vim345:
I just came across the same issue. I guess a lot of people are going to face this issue very soon, as it's very easy to reproduce it using dataclasses.

ilevkivskyi:
I guess this should be not hard to fix, mypy cares whether a callable was assigned in class (to match what Python does), so we should just tweak the `Var`s in plugin to pretend they were assigned on `self`. But we still need a volunteer who will fix this.

ojii:
If someone needs a workaround, this works:

```python
from dataclasses import dataclass
from typing import *

T = TypeVar("T")


@dataclass
class Box(Generic[T]):
    inner: T

    @property
    def unboxed(self) -> T:
        return self.inner


[...]

PR #10536, opened 2021-05-26
Fix support of dataclass callable fields

### LINKED ISSUE #708: Cannot assign to field of Callable type

wyfo:
### Description
This PR is related to #708 and #5485, but fix them for dataclasses only. (Is there really an issue for regular classes?)

*mypy* currently handle dataclass callable fields as instance methods, as they are annotated at class level the same way a method would be. As a consequence, field node attribute `is_initialized_in_class` is `True`. 

Here is an illustration of the current mypy behavior, comparing "regular" instance field annotation and dataclass one:
```python
from dataclass import dataclasses
from typing import Callable

class A:
    def __init__():
        self.a: Callable[[int], int] = lambda i: i

@dataclass
class B:
    b: Callable[[int], int] = lambda i: i

A().a(0)  # is_initialized_in_class is False for variable `A().a` -> no error because A().a is instance variable access
B().b(0)  # is_initialized_in_class is True for variable `b().b`  -> error because B().b is a method access
```

By setting field node `is_initialized_in_class` to `False`, the field is then treated the same way it would be when declared with annotated assignment in `__init__` method.

A first PR #10292 by @aecay concerning this issue was merged one month ago, but its implementation was more complex and did not fix the bug for *frozen* dataclasses, so I've reverted it. I hope there is no problem with that.

## Test Plan

Add unit tests:
- mypy/test/testcheck.py::TypeCheckSuite::testDataclassCallableFieldAccess
- mypy/test/testcheck.py::TypeCheckSuite::testDataclassCallableFieldAssignment


wyfo:
I had to modify `mypy/test/testdeps.py::GetDependenciesSuite::testDataclassDeps` because of a side effect of the code modification. Dependency computation is indeed impacted, but I think this is not a real issue. 
In fact, the dependency `<m.B.y> -> m` appears as soon as `y` field is used in the module (for example in the test, by adding `B(0, 0).y` below the class definition), so it should be there when needed.
However, I'm new to *mypy* code, so I'm not sure of my understanding of dependencies.

wyfo:
Should be closed because a more general fix is done #10548

PR #10546, opened 2021-05-28
Resolves #10283; check for module level ignore_missing_imports on missing stubs

### LINKED ISSUE #10283: --ignore-missing-imports not working for modularized typeshed types
**Bug Report**

Mypy does not respect the `--ignore-missing-imports` flag/`ignore_missing_imports = True` config for types that are now modularized out of typeshed (eg the [`legacy_bundled_packages`](http://github.com/FuegoFro/mypy/blob/98edc4148ac9fb3bdb601419ab2bafbac848fc6c/mypy/stubinfo.py#L4-L86)).

I don't *think* this is a typeshed bug, but let me know if you disagree and I can make an issue there instead.

**To Reproduce**

Given the following file:

```python
from redis.sentinel import Sentinel
```

I always get the following output:

```
python -m mypy --ignore-missing-imports example.py
example.py:1: error: Library stubs not installed for "redis.sentinel" (or incompatible with Python 3.7)
example.py:1: note: Hint: "python3 -m pip install types-redis"
example.py:1: note: (or run "mypy --install-types" to install all missing stub packages)
example.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

It doesn't matter whether `--ignore-missing-imports` is included or whether `types-redis` is installed, it will always give me exactly that output.

Note that a line like `from foo_asdf import bar` *is* correctly ignored by the `--ignore-missing-imports` flag and types that *are* defined by `types-redis` (like `redis.StrictRedis`) will work correctly (with or without the flag) when that package is installed.

**Your Environment**

- Mypy version used: `mypy 0.820+dev.e46aa6df6126cf47bea97bf9f94d7cba093d0103` (latest `master` at time of filing)
- Mypy command-line flags: `--ignore-missing-imports`
- Python version used: Python 3.7.6

TheCleric:
### Description

Fixes #10283 

For missing type stubs for legacy bundled packages, we will now look for module level overrides for ignore_missing_imports. If present, we will not issue an error like: `Cannot find implementation or library stub for module named "redis.sentinel"`

## Test Plan

3 new test scenarios were added for this PR in `check-modules.test` for a missing type stub for a legacy bundled package:
Test Name | `ignore_missing_imports` global | `ignore_missing_imports` module override | Result
-|-|-|-
testIgnoreMissingImportsGlobalWithMissingStub | True | False | Error
testIgnoreMissingImportsWildcardWithMissingStub | False | True | No Error
testIgnoreMissingImportsGlobalAndWildcardWithMissingStub | True | True | No Error

The first two tests were passing before the code change and are there to help ensure no regression. The last test was the use case from #10283 to make sure the new code accomplished its mission.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -11,6 +11,7 @@ The function build() is the main interface to this module.
 # TODO: More consistent terminology, e.g. path/fnam, module/id, state/file
 
 import contextlib
+import copy
 import errno
 import gc
 import json
@@ -2440,16 +2441,25 @@ def find_module_and_diagnose(manager: BuildManager,
         # Could not find a module.  Typically the reason is a
         # misspelled module name, missing stub, module not in
         # search path or the module has not been installed.
-
         ignore_missing_imports = options.ignore_missing_imports
         top_level = file_id.partition('.')[0]
+
         # Don't honor a global (not per-module) ignore_missing_imports
         # setting for modules that used to have bundled stubs, as
         # otherwise updating mypy can silently result in new false
-        # negatives.
+        # negatives. However, if ignore_missing_imports was set
+        # at the module level, we will honor it.
+        # https://github.com/python/mypy/issues/10283
         global_ignore_missing_imports = manager.options.ignore_missing_imports
         if top_level in legacy_bundled_packages and global_ignore_missing_imports:
-            ignore_missing_imports = False
+            # To see if this ignore_missing_imports was due to the global or due to the
+            # module, let's copy the options, set global ignore_missing_imports to False
+            # and then look again.
+            options_copy = copy.deepcopy(manager.options)
+            options_copy.ignore_missing_imports = False
+            module_copy = options_copy.clone_for_module(id)
+            if not module_copy.ignore_missing_imports:
+                ignore_missing_imports = False
 
         if skip_diagnose:
             raise ModuleNotFound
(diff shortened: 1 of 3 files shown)


mypy/build.py
```diff
@@ -2440,16 +2441,25 @@ def find_module_and_diagnose(manager: BuildManager,
...
+        # negatives. However, if ignore_missing_imports was set
+        # at the module level, we will honor it.
+        # https://github.com/python/mypy/issues/10283
         global_ignore_missing_imports = manager.options.ignore_missing_imports
         if top_level in legacy_bundled_packages and global_ignore_missing_imports:
-            ignore_missing_imports = False
+            # To see if this ignore_missing_imports was due to the global or due to the
+            # module, let's copy the options, set global ignore_missing_imports to False
+            # and then look again.
+            options_copy = copy.deepcopy(manager.options)
```
JukkaL, line 2458: This seems potentially dangerous, since `Options` can have significant state (`_per_module_cache`) and it's hard to reason about the performance and correctness implications of the cache. I came up with a simple alternative approach that doesn't use `deepcopy`, and since there is not much time before the release, I'll create a PR now (but I'll give you credit for the first implementation).

JukkaL:
Thanks for the PR! I'm closing this in favor of #10582 (see my comment above) primarily since options cloning has previously been a real performance bottleneck for very large projects, and I want to be careful not to regress performance.

TheCleric:
> Thanks for the PR! I'm closing this in favor of #10582 (see my comment above) primarily since options cloning has previously been a real performance bottleneck for very large projects, and I want to be careful not to regress performance.

Completely understand. I wasn't super happy with my approach anyway. 😂

PR #10548, opened 2021-05-28
Fix callable instance variable support

### LINKED ISSUE #708: Cannot assign to field of Callable type

### LINKED ISSUE #5485: Class with function fields incorrectly thinks first argument is self

wyfo:
Fixes #708 and Fixes #5485

Prevent handling as bounded method of callable members declared as instance variables.

### DIFF
--- a/mypy/checkmember.py
+++ b/mypy/checkmember.py
@@ -530,6 +530,15 @@ def instance_alias_type(alias: TypeAlias,
     tp = type_object_type(target.type, builtin_type)
     return expand_type_by_instance(tp, target)
 
+def is_instance_var(var: Var, info: TypeInfo) -> bool:
+    """Return if var is an instance variable according to PEP 526."""
+    return (
+        # check the type_info node is the var (not a decorated function, etc.)
+        var.name in info.names and info.names[var.name].node is var
+        and not var.is_classvar
+        # variables without annotations are treated as classvar
+        and not var.is_inferred
+    )
 
 def analyze_var(name: str,
                 var: Var,
@@ -559,7 +568,12 @@ def analyze_var(name: str,
         t = get_proper_type(expand_type_by_instance(typ, itype))
         result = t  # type: Type
         typ = get_proper_type(typ)
-        if var.is_initialized_in_class and isinstance(typ, FunctionLike) and not typ.is_type_obj():
+        if (
+            var.is_initialized_in_class
+            and not is_instance_var(var, info)
+            and isinstance(typ, FunctionLike)
+            and not typ.is_type_obj()
+        ):
             if mx.is_lvalue:
                 if var.is_property:
                     if not var.is_settable_property:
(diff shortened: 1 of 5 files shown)


JelleZijlstra:
This is great! It's a very long-lasting bug so it's wonderful to get it fixed. The mypy-primer output looks very promising. https://github.com/willmcgugan/rich/blob/master/rich/pager.py#L20 seems like it's missing a `ClassVar`.

wyfo:
Yes, *mypy-primer* looks very promising! (I'm also excited, thinking about all the `# type: ignore` I will remove from my code)
`ClassVar` would also be missing here https://github.com/sphinx-doc/sphinx/blob/830b3fbe2babcc8df33f767ce3a406b16c0cac1c/sphinx/domains/std.py#L54

msullivan (approved):
Fantastic, thank you!

hauntsaninja:
Just wanted to say thank you again! This fixes a longstanding issue!

PR #10550, opened 2021-05-29
[mypyc] Avoid crash when importing unknown module with from import

pranavrajpal:
Fixes mypyc/mypyc#851

This fixes a bug where code compiled with mypyc would crash on from imports (`from x import y`) if:
- y is a module
- mypy doesn't know that y is a module (due to an ignore_missing_imports configuration option or something else)

The bug was caused by using getattr to import modules (i.e. `y = getattr(x, 'y')`) and changing this to `import x.y as y` when it can determine that y is a module. This doesn't work when we don't know that y is a module.

I changed the from import handling to use something similar to the method shown in the [`__import__` docs](https://docs.python.org/3/library/functions.html#__import__). I also removed the special casing of from imports for modules (`from x import y` where y is a module) mentioned earlier, because these changes make that special casing unnecessary.

## Test Plan

I added 2 tests to mypyc/test-data/run-imports.test.

### DIFF
--- a/mypyc/irbuild/builder.py
+++ b/mypyc/irbuild/builder.py
@@ -46,7 +46,7 @@ from mypyc.primitives.list_ops import to_list, list_pop_last, list_get_item_unsa
 from mypyc.primitives.dict_ops import dict_get_item_op, dict_set_item_op
 from mypyc.primitives.generic_ops import py_setattr_op, iter_op, next_op
 from mypyc.primitives.misc_ops import (
-    import_op, check_unpack_count_op, get_module_dict_op, import_extra_args_op, get_locals
+    import_op, check_unpack_count_op, get_module_dict_op, import_extra_args_op
 )
 from mypyc.crash import catch_errors
 from mypyc.options import CompilerOptions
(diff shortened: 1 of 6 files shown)


mypyc/irbuild/builder.py
```diff
@@ -286,19 +288,45 @@ def add_to_non_ext_dict(self, non_ext: NonExtClassInfo,
         key_unicode = self.load_str(key)
         self.call_c(dict_set_item_op, [non_ext.dict, key_unicode, val], line)
 
+    def gen_import_from(self, id: str, line: int, imported: List[str]) -> None:
+        self.imports[id] = None
+
+        globals_dict = self.load_globals_dict()
+        locals_dict = self.call_c(get_locals, [], line)
```
msullivan, line 295: Drop the whole `'PyEval_GetLocals` thing and just pass in `None` for that. `PyEval_GetLocals` won't do anything that's even a little bit like being correct in our setting (I think it will return the locals dict for whatever the enclosing interpreted function is?), and the docs for `__import__` reports " The standard implementation does not use its locals argument at all"

msullivan (commented):
Looks great, just drop the locals dict stuff.

mypyc/irbuild/builder.py
pranavrajpal: This should be fixed in 8294834c6619ed0453ab6d2b3265a9377c435a1e. I also removed the primitive for getting the locals dict because we aren't using it anymore.

PR #10551, opened 2021-05-29
Upadate

kkshitish:
I do change because the code is not necessary. If we do something in one line then, why we need more lines code. I tested this after update. it was running well. There is no come after changing.

 Have you read the [Contributing Guidelines](https://github.com/python/mypy/blob/master/CONTRIBUTING.md)?

Yes 

Description:

This changing required because after update code become short and I remove the unwanted lines code. If we do something one line then, why we need more lines of code.


JelleZijlstra:
This changes the meaning of the code and doesn't solve any issue.

Issue #8646, opened 2020-04-07
Idea: warn if an object without a "significant" boolean representation is used in a boolean context


brianmaissy:
This is a bit of a long shot, because this isn't really a type error per se. But consider the following:
```
numbers: List[int] = []
my_generator: Iterable = (i for i in numbers)
if not my_generator:
    print("empty")
```

Obviously this will not print "empty", because a is a `generator`, which is always truthy. However, 99.9% of the times this is not what the user intended, and it would be great if a tool could identify it and warn about it, and mypy seems well-placed to be able to do it.

Beyond the fact that this isn't really a type "error", there are some more complications. For example, how to more generally define what objects don't have a "significant" boolean representation. One could argue that this is just as much a "mistake" as the example with the generator:
```
class A:
    pass

a: A = A()
if not a:
    print("empty")
```

As well as the question of what's considered "using something in a boolean context". Should calling `bool()` on a generator be the same as using it in the conditional expression of an `if` or `while`?

So I'm not at all sure this is a feasible or reasonable suggestion, but I didn't find any mention if it in past issues, so thought it was worth bringing up for discussion.

JukkaL:
I've encountered similar issues myself, though I can't remember if there has been any online discussion about these. Stricter boolean checks would at least be nice as an opt-in check.

hauntsaninja:
#10666 will add an opt-in check for this.

PR #10557, opened 2021-05-30
Fail when conditioning on implicit bool

ikonst:
### Description

Fails conditionals where the expression's type does not implement `__bool__` nor `__len__` since such expressions would always evaluate as truthy and likely an error (e.g. the common pitfall of conditioning on a callable rather than its result).

This feature was proposed by @pkch in https://github.com/python/mypy/issues/3195#issuecomment-295341122 as an alternative to the deprecated ``--strict-bool``.

## Test Plan

Test case was added.

### DIFF
--- a/docs/source/error_code_list.rst
+++ b/docs/source/error_code_list.rst
@@ -665,6 +665,24 @@ consistently when using the call-based syntax. Example:
     # Error: First argument to namedtuple() should be "Point2D", not "Point"
     Point2D = NamedTuple("Point", [("x", int), ("y", int)])
 
+
+
+Implicitly converting to bool always evaluates as True [implicit-bool]
+----------------------------------------------------------------------
+
+In contexts where an expression has to be converted to a bool, an object
+which does not implement __bool__ nor __len__ will always evaluate to True.
+
+.. code-block:: python
+
+    class Foo:
+      pass
+
+    foo = Foo()
+    # Error: "{}" does not implement __bool__ or __len__ so the expression is always truthy
+    if foo:
+       ...
+
 Report syntax errors [syntax]
 -----------------------------
 
--- a/mypy/message_registry.py
+++ b/mypy/message_registry.py
@@ -84,6 +84,11 @@ DUPLICATE_TYPE_SIGNATURES = 'Function has duplicate type signatures'  # type: Fi
 DESCRIPTOR_SET_NOT_CALLABLE = "{}.__set__ is not callable"  # type: Final
 DESCRIPTOR_GET_NOT_CALLABLE = "{}.__get__ is not callable"  # type: Final
 MODULE_LEVEL_GETATTRIBUTE = '__getattribute__ is not valid at the module level'  # type: Final
+TYPE_CONVERTS_TO_BOOL_IMPLICITLY = \
+    '"{}" does not implement __bool__ or __len__ so the expression is always truthy'  # type: Final
+ALL_TYPES_IN_UNION_CONVERT_TO_BOOL_IMPLICITLY = \
+    'Neither of {} implements __bool__ or __len__ '\
+    'so the expression is always truthy'  # type: Final
 
 # Generic
 GENERIC_INSTANCE_VAR_CLASS_ACCESS = \
--- a/mypy/test/helpers.py
+++ b/mypy/test/helpers.py
@@ -8,6 +8,7 @@ import contextlib
 from typing import List, Iterable, Dict, Tuple, Callable, Any, Optional, Iterator
 
 from mypy import defaults
+from mypy import errorcodes as codes
 import mypy.api as api
 
 import pytest
@@ -372,7 +373,6 @@ def assert_type(typ: type, value: object) -> None:
 def parse_options(program_text: str, testcase: DataDrivenTestCase,
                   incremental_step: int) -> Options:
     """Parse comments like '# flags: --foo' in a test case."""
-    options = Options()
     flags = re.search('# flags: (.*)$', program_text, flags=re.MULTILINE)
     if incremental_step > 1:
         flags2 = re.search('# flags{}: (.*)$'.format(incremental_step), program_text,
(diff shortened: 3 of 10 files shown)


docs/source/error_code_list.rst
```diff
@@ -665,6 +665,24 @@ consistently when using the call-based syntax. Example:
     # Error: First argument to namedtuple() should be "Point2D", not "Point"
     Point2D = NamedTuple("Point", [("x", int), ("y", int)])
 
+
+
+Implicitly converting to bool always evaluates as True [implicit-bool]
```
ikonst, line 670: I'm not sure whether "implicit-bool" should be the code, and whether "implicit conversion" captures what's happening here. Perhaps there's some accepted terminology for this in cpython?

mypy/test/helpers.py
```diff
@@ -372,16 +373,19 @@ def assert_type(typ: type, value: object) -> None:
...
     """Parse comments like '# flags: --foo' in a test case."""
-    options = Options()
     flags = re.search('# flags: (.*)$', program_text, flags=re.MULTILINE)
     if incremental_step > 1:
         flags2 = re.search('# flags{}: (.*)$'.format(incremental_step), program_text,
                            flags=re.MULTILINE)
         if flags2:
             flags = flags2
 
+    disabled_error_codes = {codes.IMPLICIT_BOOL}
```
ikonst, line 383: Too many fixtures would need multiple changes to add `__len__` to list, `__bool__` to int, etc. and I understand that making fixtures more "complete" across the board is discouraged (to keep tests lean).

docs/source/error_code_list.rst
hauntsaninja: The Python docs seem to use the term "boolean operations"

mypy/message_registry.py
```diff
@@ -84,6 +84,11 @@
 DESCRIPTOR_SET_NOT_CALLABLE = "{}.__set__ is not callable"  # type: Final
 DESCRIPTOR_GET_NOT_CALLABLE = "{}.__get__ is not callable"  # type: Final
 MODULE_LEVEL_GETATTRIBUTE = '__getattribute__ is not valid at the module level'  # type: Final
+TYPE_CONVERTS_TO_BOOL_IMPLICITLY = \
+    '"{}" does not implement __bool__ or __len__ so the expression is always truthy'  # type: Final
+ALL_TYPES_IN_UNION_CONVERT_TO_BOOL_IMPLICITLY = \
+    'Neither of {} implements __bool__ or __len__ '\
```
hauntsaninja, line 90: nit: 
```suggestion
    'No members of {} implement __bool__ or __len__ '\
```

hauntsaninja (commented):
Looks reasonable. This is more #8646 than pkch's comment, since e.g. you don't warn about `x: Optional[int]; if x: ...`. I'd also check for `__bool__` that returns a Literal, a la #9297 (although maybe complicated, e.g. for Unions you'd need to check that all members aren't returning the same Literal)

ikonst:
> I'd also check for __bool__ that returns a Literal

So, this opens the door for a more generic always-true/false check, and this was actually my first approach - to leverage the `can_be_true` and `can_be_false` attributes, i.e.
```python
if t.can_be_true and not t.can_be_false:
  self.fail('expression is always true',  ...)
elif not t.can_be_true and t.can_be_false:
  self.fail('expression always false',  ...)
```

For this to detect the "implicit" case, I had to teach `Instance` to flag it:
```python
class Instance(ProperType):
    ...
    def __init__(self, ...):
       ...
       if self.type and not has_readable_member('__bool__') and not t.type.has_readable_member('__len__'):
           self.can_be_true = True
           self.can_be_false = False
```

This has the benefit of handling many of the details (literals, unions). However:
1. We'd need to explicitly allow patterns which are obviously intentional like `while True`. Can you think of other such patterns? Should `if True` becopme an error?
2. The current code propagates the `can_be_true` and `can_be_false` to the expression's root. To leverage it for errors, we'd need more context to produce different errors for different conditions, e.g. instead of "expression is always true", have it say "literal value 42 is always true" or "type Foobar has no __bool__ or __len__". For this to happen, perhaps I shouldn't handle it in the "if" and "conditional" visitors, but rather in the code that evaluates the expressions, so that I could flag it right in, say, `42 or get_data()` (regardless of whether you end up doing `if 42 or get_data()`).

Initially I hesitated trying to make a generic "always-true/false" checker simply because I felt the error wouldn't be as useful to flag a very common pitfall of using an object as a boolean.

docs/source/error_code_list.rst
ikonst: "context of Boolean operation" is also used

ikonst:
I'll try a different approach.

Akuli:
In the past, I have used `if True:` when a linter complains about something unless it's hidden in a conditional (lol), and `if False:` to disable code. But I think that's it: if the condition is always true or always false, and the condition is not literally `True` or `False`, then it should be an error.

Mypy already has a flag that makes it warn about code that never runs. If the scope of this PR grows, maybe consider how it interacts with that?

ikonst:
> In the past, I have used if True: when a linter complains about something ...
> Mypy already has a flag that makes it warn about code that never runs.

Yes, and there's special treatment for "always true" and "always false" that's established during semantic analysis, e.g. so that `if TYPE_CHECKING` and `if MYPY` wouldn't result in errors. `True` and `False` literals, however, aren't handled there (unless you pass e.g. `--always-true True`). We would need to see how we should handle patterns like those.

Right now there's no special handling, so e.g.
```python
if False:
  print('x') 
```
would result in:
```
$ mypy --warn-unreachable test.py
test.py:2: error: Statement is unreachable
```

Akuli:
Perhaps it's good that mypy warns about `if False:`, as it's something used for temporarily disabling something, and you shouldn't leave it in when cleaning up the code.

ikonst:
Merged #10666 instead since it had a cooler number.

PR #10562, opened 2021-05-31
Unreachable expressions: treat non-zero ints as truthy

ikonst:
Makes `42 or False` result in `Error: Right operand of "or" is never evaluated` when invoked with [--warn-unreachable](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-unreachable).


ikonst:
👋  @hauntsaninja

PR #10566, opened 2021-06-01
check_boolean_op: clean up semanal/type-based reachability

ikonst:
`ExpressionChecker.check_boolean_op` supports two potentially conflicting ways of determining whether an expression is always true/false: name-based checks from semantic analysis (e.g. `name in options.always_true`) and type narrowing in `TypeChecker.find_isinstance_check` (*). The two ways can be set to conflict, e.g.
```python
# flags: --always-true COMPILE_TIME_TRUE
COMPILE_TIME_TRUE: None
```
In the above case, `COMPILE_TIME_TRUE` would be always true per semanal and always false per `find_isinstance_check`, which results in an assertion failure here:
https://github.com/python/mypy/blob/5eb4de443269b906d2e8b2292dbf472ce17d0e1f/mypy/checkexpr.py#L2828
and here:
https://github.com/python/mypy/blob/5eb4de443269b906d2e8b2292dbf472ce17d0e1f/mypy/checkexpr.py#L2832

This change reconciles the two, by prioritizing the semanal flags (e.right_always, e.right_unreachable) over type-narrowing (if_map, else_map). Then we explicitly avoid flagging reachability errors when the semanal flags are set, since they're obviously used intentionally.

(*) which seems to do slightly more than `isinstance` checks

### DIFF
--- a/mypy/checkexpr.py
+++ b/mypy/checkexpr.py
@@ -2774,38 +2774,32 @@ class ExpressionChecker(ExpressionVisitor[Type]):
 
         assert e.op in ('and', 'or')  # Checked by visit_op_expr
 
-        if e.op == 'and':
+        if e.right_always:
+            left_map, right_map = None, {}
+        elif e.right_unreachable:
+            left_map, right_map = {}, None
+        elif e.op == 'and':
             right_map, left_map = self.chk.find_isinstance_check(e.left)
-            restricted_left_type = false_only(left_type)
-            result_is_left = not left_type.can_be_true
         elif e.op == 'or':
             left_map, right_map = self.chk.find_isinstance_check(e.left)
-            restricted_left_type = true_only(left_type)
-            result_is_left = not left_type.can_be_false
 
         # If left_map is None then we know mypy considers the left expression
         # to be redundant.
-        #
-        # Note that we perform these checks *before* we take into account
-        # the analysis from the semanal phase below. We assume that nodes
-        # marked as unreachable during semantic analysis were done so intentionally.
-        # So, we shouldn't report an error.
-        if codes.REDUNDANT_EXPR in self.chk.options.enabled_error_codes:
-            if left_map is None:
-                self.msg.redundant_left_operand(e.op, e.left)
+        if (
+            codes.REDUNDANT_EXPR in self.chk.options.enabled_error_codes
+            and left_map is None
+            # don't report an error if it's intentional
+            and not e.right_always
+        ):
+            self.msg.redundant_left_operand(e.op, e.left)
 
-        # Note that we perform these checks *before* we take into account
-        # the analysis from the semanal phase below. We assume that nodes
-        # marked as unreachable during semantic analysis were done so intentionally.
-        # So, we shouldn't report an error.
-        if self.chk.should_report_unreachable_issues():
-            if right_map is None:
-                self.msg.unreachable_right_operand(e.op, e.right)
-
-        if e.right_unreachable:
-            right_map = None
-        elif e.right_always:
-            left_map = None
+        if (
+            self.chk.should_report_unreachable_issues()
+            and right_map is None
+            # don't report an error if it's intentional
+            and not e.right_unreachable
+        ):
+            self.msg.unreachable_right_operand(e.op, e.right)
 
         # If right_map is None then we know mypy considers the right branch
         # to be unreachable and therefore any errors found in the right branch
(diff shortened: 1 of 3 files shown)


ikonst:
(still 🤔 how to do that, but I wanted to have the regression test in place)

mypy/checkexpr.py
```diff
@@ -2774,38 +2774,34 @@ def check_boolean_op(self, e: OpExpr, context: Context) -> Type:
 
         assert e.op in ('and', 'or')  # Checked by visit_op_expr
 
-        if e.op == 'and':
+        left_map: mypy.checker.TypeMap
+        right_map: mypy.checker.TypeMap
+        if e.right_always:
```
ikonst, line 2779: For always-true/false cases, `find_isinstance_check` returns a `{}, None` or a `None, {}` pair. We do the same here, but without even calling `find_isinstance_check`.

mypy/checkexpr.py
```diff
@@ -2822,6 +2818,13 @@ def check_boolean_op(self, e: OpExpr, context: Context) -> Type:
             assert right_map is not None  # find_isinstance_check guarantees this
             return right_type
 
+        if e.op == 'and':
```
ikonst, line 2821: Can safely move it down here since nothing uses `restricted_left_type` and `result_is_left` until this point.

ikonst:
@hauntsaninja @JukkaL want to take a look?

ikonst:
@hauntsaninja @JukkaL gentle ping

hauntsaninja (commented):
Sorry for the delay and thanks for this well explained PR that fixes a crash!
However, the reveal types in your test cases aren't quite what I'd expect. E.g., in your first test case I'd expect `reveal_type(b)` to be bool (`Literal[True]` and `None` shouldn't really be appearing anywhere, since mypy should always think the COMPILE_TIME_* are True or False). 
I think we should maybe set `left_type` and `right_type` if `e.right_always` or `e.right_unreachable`? Also happy to merge as is!

ikonst:
> However, the reveal types in your test cases aren't quite what I'd expect. E.g., in your first test case I'd expect reveal_type(b) to be bool (Literal[True] and None shouldn't really be appearing anywhere, since mypy should always think the COMPILE_TIME_* are True or False).

You're saying you'd expect:
```diff
 indeterminate: bool
 COMPILE_TIME_FALSE: Literal[True]
 b = indeterminate or COMPILE_TIME_FALSE
-reveal_type(b)  # N: Revealed type is "Union[builtins.bool, Literal[True]]"
+reveal_type(b)  # N: Revealed type is "builtins.bool"
```
but we don't know if `indeterminate` is truthy; if it is, the expression's type is LHS type, else, it's the RHS type (never mind if RHS is truthy or not).

Or in this case:
```diff
 indeterminate: bool
 COMPILE_TIME_TRUE: None
 b = indeterminate or COMPILE_TIME_TRUE
-reveal_type(b)  # N: Revealed type is "Union[builtins.bool, None]"
+reveal_type(b)  # N: Revealed type is "builtins.bool"
```

Same logic. `COMPILE_TIME_TRUE` is typed `None`, never mind if it's truthy or not. The expression should have either LHS or RHS type -- since LHS is indeterminate, a union of both.

hauntsaninja:
`mypy --help` says:
```
  --always-true NAME        Additional variable to be considered True (may be repeated)
  --always-false NAME       Additional variable to be considered False (may be repeated)
```
which I interpreted as the variable should be interpreted as that literal boolean value, not just something that's truthy or falsey (TYPE_CHECKING and MYPY are often used to make bold lies to the type checker, so just wiping out the value seemed like a valid interpretation).
But https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-always-true reads much more ambiguously, it seems you didn't read it that way, and changing it would require more work for something that's kind of niche. So like I said, happy to merge, and thank you again!

ikonst:
> the variable should be interpreted as that literal boolean value, not just something that's truthy or falsey

That's what I'd assume too, and maybe that'd be a better way to implement it. As for now, those options manifest in `OpExpr. right_always` and `OpExpr.right_unreachable` (a semanal artifact) which are booleans used purely for this feature. I've added comments to document that:
```diff
-    # Is the right side going to be evaluated every time?
+    # Per static analysis only: Is the right side going to be evaluated every time?
     right_always = False
-    # Is the right side unreachable?
+    # Per static analysis only: Is the right side unreachable?
     right_unreachable = False
```

PR #10568, opened 2021-06-01
Fix typo in config_file.rst

wwuck:
### Description

One example references `[[tools.mypy.overrides]]` when all other examples reference `[[tool.mypy.overrides]]`.  I'm guessing the odd one out is incorrect.


ethanhs (approved):
Yeah the PEP specifies `tool`, thanks for catching this!

PR #3666, opened 2017-07-05
Suppress any errors found in a boolean expression after filtering out all possible types.

rowillia:
This fixes https://github.com/python/mypy/issues/3643

PR #10569, opened 2021-06-01
Convert MessageBuilder.disable_errors to a context manager

ikonst:


mypy/messages.py
```diff
@@ -136,11 +137,13 @@ def add_errors(self, messages: 'MessageBuilder') -> None:
                 for info in errs:
                     self.errors.add_error_info(info)
 
-    def disable_errors(self) -> None:
+    @contextmanager
+    def disable_errors(self) -> ContextManager[None]:
```
JelleZijlstra, line 141: Return type should be `Iterable[None]` (or something with `Generator`). The function returns an iterable, the decorator turns it into a ContextManager.

ikonst: Oops, thanks. I guess I was trying to placate PyCharm which complained that it doesn't have `__enter__()` when I tried to use it "manually" within `analyze_type_type_member_access`.

Maybe I should just bite the bullet and use `with` in `analyze_type_type_member_access` (and indent all the code).

mypy/checkexpr.py
```diff
@@ -2815,13 +2807,11 @@ def check_boolean_op(self, e: OpExpr, context: Context) -> Type:
         elif e.right_always:
             left_map = None
 
-        if right_map is None:
-            self.msg.disable_errors()
-        try:
+        # If right_map is None then we know mypy considers the right branch
```
ikonst, line 2810: Oops, the original comment added in https://github.com/python/mypy/pull/3666 got shifted around over the years.

mypy/checkexpr.py
```diff
@@ -1,7 +1,7 @@
 """Expression type checker. This file is conceptually part of TypeChecker."""
 
 from mypy.ordered_dict import OrderedDict
-from contextlib import contextmanager
+from contextlib import contextmanager, nullcontext
```
hauntsaninja, line 4: CI will tell you that 3.6 doesn't have nullcontext. Easy enough to redefine though

hauntsaninja (approved):

mypy/checkexpr.py
ikonst: 🙏

mypy/nullcontext.py
```diff
@@ -0,0 +1,10 @@
+import sys
+from contextlib import contextmanager
+from typing import Iterator
+
+if sys.version_info < (3, 7):
+    @contextmanager
+    def nullcontext() -> Iterator[None]:
+        yield
+else:
+    from contextlib import nullcontext as nullcontext, contextmanager  # noqa: F401
```
hauntsaninja, line 10: nit: don't think you need to import contextmanager again. I think you could also move this to `mypy/util.py` instead of making a separate file.

ikonst: Thanks, copy-pasta, will fix.

Re `mypy/util.py`, I followed the pattern put in place by `mypy/ordered_dict.py`. Maybe we want to put both of those in `mypy/backports.py` or `mypy/compat.py`?

hauntsaninja:
Thank you!

atakiar:
@ikonst we might be able to close #1184? Also PRs #9701 and #10251

ikonst:
Oops, I didn't notice there were other PRs and issues! Perhaps an admin can close them.

JelleZijlstra:
I closed the other two PRs. I think the issue can remain open for now since there are other similar refactors to be done (e.g., #10685).

PR #10579, opened 2021-06-04
Restrict the number of errors shown when there are missing stubs

### LINKED ISSUE #10529: Limit the number of generated errors if stubs are missing
If stubs for third-party libraries have not been installed, mypy may generate thousands of errors, making it hard to figure out what's going on. One reason for this is that missing stubs cause many `Any` types to be inferred, and these may trigger additional errors, depending on the strictness options being used.

One way to address this is to use some heuristic to silence errors that are *not* about missing stubs after a threshold. For example, if we've reported N errors about missing stubs and at least M errors overall, hide any remaining errors that are not about stubs.

This could be a significant issue for some users after the next mypy release (0.900), which unbundles stubs for third-party libraries. Thus users are likely to see errors about missing stubs as soon they update to mypy 0.900.

JukkaL:
When upgrading to mypy 0.900, most projects will see some errors about
missing stubs. Projects with strict settings could see thousands of errors, 
since missing stubs will generate many additional `Any` types. 

After 200 errors we will now only show errors about unresolved imports or 
missing stubs, so that the likely root causes won't be obscured in a high 
volume of errors.

Fixes #10529.


mypy/errors.py
```diff
@@ -10,14 +10,19 @@
...
 from mypy import errorcodes as codes
 from mypy.util import DEFAULT_SOURCE_OFFSET, is_typeshed_file
 
 T = TypeVar('T')
 allowed_duplicates = ['@overload', 'Got:', 'Expected:']  # type: Final
 
 
+# Threshold after which we may filter out most errors to avoid very
+# verbose output
+MANY_ERRORS_THRESHOLD = 200
```
ilevkivskyi, line 23: I think this should be tunable by user (e.g. with a flag). For example, what if the errors are caused by something else (not missing imports). I always prefer to see all the errors especially when working with the PyCharm plugin.

Finally, testing will be simpler when it is tunable.

test-data/unit/check-modules.test
```diff
@@ -2982,3 +2982,212 @@ T = TypeVar("T")
...
+certifi.x  # E: Expression has type "Any"
+certifi.x  # E: Expression has type "Any"
+certifi.x  # E: Expression has type "Any"
+certifi.x  # E: Expression has type "Any"
+# Only show 200 errors
+certifi.x  # N: (Skipping most remaining errors due to unresolved imports or missing stubs; fix these first)
+certifi.x
+certifi.x
+certifi.x
+certifi.x
```
ilevkivskyi, line 3193: Add couple more tests maybe:
* Checking that new import errors are still shown
* Checking that limit does not apply if there were no import errors

ilevkivskyi (approved):

ilevkivskyi (approved):

PR #10582, opened 2021-06-04
Make ignore_missing_imports work for libs which had bundled stubs

### LINKED ISSUE #10283: --ignore-missing-imports not working for modularized typeshed types

JukkaL:
It has to be specified as a per-module option, if the library used
to have bundled stubs. This way mypy won't silently start ignoring
these missing stubs.

This makes `ignore_missing_imports` special by making it work a bit
differently, depending on whether it's set globally or per module.

The new behavior is less surprising. This makes it possible to ignore
arbitrary missing third-party stub packages, even those which used to
have bundled stubs.

Fixes #10283.

Originally implemented by @TheCleric in #10283. This is somewhat 
simpler alternative implementation which is easier to reason about and
may perform better in some cases of very large configs.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2449,7 +2449,9 @@ def find_module_and_diagnose(manager: BuildManager,
         # otherwise updating mypy can silently result in new false
         # negatives.
         global_ignore_missing_imports = manager.options.ignore_missing_imports
-        if top_level in legacy_bundled_packages and global_ignore_missing_imports:
+        if (top_level in legacy_bundled_packages
+                and global_ignore_missing_imports
+                and not options.ignore_missing_imports_per_module):
             ignore_missing_imports = False
 
         if skip_diagnose:
--- a/mypy/options.py
+++ b/mypy/options.py
@@ -82,6 +82,8 @@ class Options:
         self.no_silence_site_packages = False
         self.no_site_packages = False
         self.ignore_missing_imports = False
+        # Is ignore_missing_imports set in a per-module section
+        self.ignore_missing_imports_per_module = False
         self.follow_imports = 'normal'  # normal|silent|skip|error
         # Whether to respect the follow_imports setting even for stub files.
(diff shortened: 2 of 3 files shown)


ilevkivskyi (approved):

tadeu:
I think that this broke `dmypy`: https://github.com/python/mypy/issues/10709

PR #10586, opened 2021-06-07
[mypyc] Fix segfault when top level raises exception

pranavrajpal:
Fixes mypyc/mypyc#813

This prevents the compiled code from segfaulting when the top level code raises an exception. In that case, previously, the compiled code would store an internal pointer to the module object and would just return that pointer if the same module was initialized more than once. However, if an exception was raised while running top level code, the internal pointer would still point to the module object even though not everything was completely initialized, leading future attempts to import that module to use functions and classes that were never completely initialized.

This fixes that bug by resetting the internal module pointer to NULL when an error occurs so that the top level code is run again on future attempts to import that same function.

## Test Plan

I added a run test to run-misc.test that reproduces this segfault and made sure it passed after my changes.

I wasn't really sure where I should put that test so I ended up putting it in run-misc.test, but let me know if you want me to move it somewhere else.

### DIFF
--- a/mypyc/codegen/emitmodule.py
+++ b/mypyc/codegen/emitmodule.py
@@ -888,7 +888,7 @@ class GroupGenerator:
 
         emitter.emit_lines('{} = PyModule_Create(&{}module);'.format(module_static, module_prefix),
                            'if (unlikely({} == NULL))'.format(module_static),
-                           '    return NULL;')
+                           '    goto fail;')
         emitter.emit_line(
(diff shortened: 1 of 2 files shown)


msullivan (approved):
OK, this looks great, and is much better behavior than segfaulting.

Thinking about it a bit more, the fully right thing to do would be to also free the module and any types/statics that were created before the error occured. 

I'm going to merge this now, but could you file a follow-up issue about doing that? (Actually *doing* it is not super high priority)

PR #10589, opened 2021-06-08
[mypyc] Add mypyc run tests for singledispatch

pranavrajpal:
This adds some run tests to run-singledispatch.test for situations that are likely to be common or that seem like they would be edge cases that we don't handle correctly.

Most of these tests are commented out because mypyc doesn't properly support singledispatch yet, and the few that aren't commented out only work because they're written in a way that causes mypyc to use the standard library implementation of singledispatch.

There are some situations where I added tests that check whether mypyc sticks to what the standard library's implementation does, even though those situations are probably bugs that should probably be mypy errors. For example, `testTypeAnnotationsDisagreeWithRegisterArgument` checks whether mypyc uses the argument to register for dispatch instead of the type annotation when they disagree, even though having different types for both of those is probably a bug.

## Test Plan

I ran all of the run tests that I added and commented out the ones that weren't passing.

### DIFF
--- a/mypyc/test-data/run-singledispatch.test
+++ b/mypyc/test-data/run-singledispatch.test
@@ -1,10 +1,8 @@
 # Test cases related to the functools.singledispatch decorator
-# Most of these tests are skipped out because mypyc doesn't support singledispatch yet
+# Most of these tests are marked as xfails because mypyc doesn't support singledispatch yet
 # (These tests will be re-enabled when mypyc supports singledispatch)
 
-# TODO: change these skipped tests to xfails
-
-[case testSpecializedImplementationUsed-skip]
+[case testSpecializedImplementationUsed-xfail]
 from functools import singledispatch
 
 @singledispatch
@@ -19,7 +17,7 @@ def test_specialize() -> None:
     assert fun('a')
     assert not fun(3)
 
-[case testSubclassesOfExpectedTypeUseSpecialized-skip]
+[case testSubclassesOfExpectedTypeUseSpecialized-xfail]
 from functools import singledispatch
 class A: pass
 class B(A): pass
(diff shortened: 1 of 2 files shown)


msullivan (commented):
I think you should be able to add `-skip` to the test names instead of commenting them out

msullivan:
> I think you should be able to add `-skip` to the test names instead of commenting them out

... we ought to add an `-xfail` sigil that can be put on test names, and register those tests as xfails to pytest; then the tests will be run as part of the test suite, but not fail it, and pytest will report if they unexpectedly pass.

pranavrajpal:
> we ought to add an `-xfail` sigil that can be put on test names, and register those tests as xfails to pytest

I think this is a good idea, but I think skipping these tests will work for now, so I just added `-skip` to all the previously commented out tests.

I'll create another issue to add xfails to the test runner.

msullivan:
These are great, and cover a lot of cases. I'd suggest adding one with a few more classes in it, and some slightly interesting computation. Maybe something like `testBenchmarkVisitorTree` or `testSimulateMypy`, but adapted to use singledispatch instead of visitor patterns?

pranavrajpal:
I added 2 more run tests in 9e0548b986ad464c69ca150824cf9d6320f1f62a. The first one (`testSingledispatchTreeSumAndEqual`) is meant to be a modified version of `testBenchmarkVisitorTree` that uses singledispatch and removes the benchmarking (I think https://github.com/mypyc/mypyc-benchmarks would be a better place for benchmarks, right?).

The other test (`testSimulateMypySingledispatch`) is meant to simulate mypy's use of singledispatch in stubtest.py.

msullivan (approved):
Looks good. Switch them to `xfail` and I'll merge it!

PR #10590, opened 2021-06-08
[mypyc] Add cleanup at end of init function

pranavrajpal:
This frees everything created in the generated initialization function so that we don't leak memory in the case of an error during initialization, as mentioned in https://github.com/python/mypy/pull/10586#pullrequestreview-678090095.

### DIFF
--- a/mypyc/codegen/emitmodule.py
+++ b/mypyc/codegen/emitmodule.py
@@ -921,14 +921,13 @@ class GroupGenerator:
 
         emitter.emit_line('return {};'.format(module_static))
         emitter.emit_lines('fail:',
-                           'Py_XDECREF({});'.format(module_static),
+                           'Py_CLEAR({});'.format(module_static),
                            'Py_XDECREF(modname);')
         # the type objects returned from CPyType_FromTemplate are all new references
         # so we have to decref them
         for t in type_structs:
             emitter.emit_line('Py_XDECREF({});'.format(t))
-        emitter.emit_lines('{} = NULL;'.format(module_static),
-                           'return NULL;')
+        emitter.emit_line('return NULL;')
         emitter.emit_line('}')
 
     def generate_top_level_call(self, module: ModuleIR, emitter: Emitter) -> None:
--- a/mypyc/codegen/emitmodule.py
+++ b/mypyc/codegen/emitmodule.py
@@ -923,14 +923,14 @@ class GroupGenerator:
         emitter.emit_lines('fail:',
                            'Py_CLEAR({});'.format(module_static),
                            'Py_CLEAR(modname);')
-        # the type objects returned from CPyType_FromTemplate are all new references
-        # so we have to decref them
         for name, typ in module.final_names:
             static_name = emitter.static_name(name, module_name)
             if emitter.ctype(typ) == 'PyObject *':
                 emitter.emit_line('Py_CLEAR({});'.format(static_name))
             elif is_tagged(typ):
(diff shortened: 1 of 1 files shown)


mypyc/codegen/emitmodule.py
```diff
@@ -899,9 +899,11 @@ def generate_module_def(self, emitter: Emitter, module_name: str, module: Module
                            '    goto fail;')
 
         # HACK: Manually instantiate generated classes here
+        type_structs = []  # type: List[str]
         for cl in module.classes:
             if cl.is_generated:
                 type_struct = emitter.type_struct_name(cl)
+                type_structs.append(type_struct)
```
msullivan, line 906: I think we need to collect the type structs even for the not generated ones. The `is_generated` classes (like those created for nested functions) get initialized here, but most classes get initialized by code we generated as IR, as part of the module initialization function. 

(I can't remember why we don't arrange to do that for the generated ones; obviously I didn't feel great about that fact, which is why this block of code is labeled `HACK` :))

mypyc/codegen/emitmodule.py
```diff
@@ -918,7 +920,13 @@ def generate_module_def(self, emitter: Emitter, module_name: str, module: Module
 
         emitter.emit_line('return {};'.format(module_static))
         emitter.emit_lines('fail:',
-                           '{} = NULL;'.format(module_static),
+                           'Py_XDECREF({});'.format(module_static),
```
msullivan, line 923: Do `Py_CLEAR` for these, which will do an Py_XDECREF and NULL out the variable. (By virtue of being a macro that does not behave like a function...)

mypyc/codegen/emitmodule.py
```diff
@@ -918,7 +920,13 @@ def generate_module_def(self, emitter: Emitter, module_name: str, module: Module
...
+                           'Py_XDECREF(modname);')
+        # the type objects returned from CPyType_FromTemplate are all new references
+        # so we have to decref them
+        for t in type_structs:
```
msullivan, line 927: We should also do the finals, too. Those need to be reset to whatever the proper error value for the type is.

msullivan: I think the right thing to do is to call out to `emit_dec_ref`, which knows how to decref everything: https://github.com/python/mypy/blob/master/mypyc/codegen/emit.py#L344

And then assign the result of `c_undefined_value` to the variable, to clear it out

mypyc/codegen/emitmodule.py
```diff
@@ -918,8 +921,19 @@ def generate_module_def(self, emitter: Emitter, module_name: str, module: Module
 
         emitter.emit_line('return {};'.format(module_static))
         emitter.emit_lines('fail:',
-                           '{} = NULL;'.format(module_static),
-                           'return NULL;')
+                           'Py_CLEAR({});'.format(module_static),
+                           'Py_CLEAR(modname);')
+        # the type objects returned from CPyType_FromTemplate are all new references
```
msullivan, line 926: I think this comment has drifted away from the code it is describing

pranavrajpal: This should be fixed in ff8d287f196b98cc55809ecb72a4f4b674bc09bc.

mypyc/codegen/emitmodule.py
pranavrajpal: Yeah, that does look a lot cleaner. This should be fixed in ebe20ed7d03802584fdd593ac77224e9be3f8022.

PR #10603, opened 2021-06-08
Don't suggest installing types-pycrypto

JelleZijlstra:
It doesn't exist, since it was removed from typeshed in October (python/typeshed#4606). Similarly, pynamodb was removed in January.

I haven't tracked down what happened to pyre-extensions in typeshed, but https://pypi.org/project/types-pyre-extensions/ doesn't exist and in any case mypy users presumably don't need it.


JelleZijlstra:
Looks like I'll need to update some tests as well.

TH3CHARLie (approved):

JukkaL:
Thanks for fixing this and uploading the dummy packages! I forgot to double check that the suggestions are still up to date.

@JelleZijlstra uploaded PyPI dummy packages for these so that there is no security risk. Still, I think it makes to release 0.902 with this fix, but I think that it can wait a few days at least so that we can include other fixes as well if we want.

PR #10610, opened 2021-06-09
Don't suggest incompatible stub packages

### LINKED ISSUE #10602: `--install-types` suggests incompatible stub packages
**Bug Report**

Running `mypy --install-types` suggest installation of stubs that are ot compatible with the current running version of python.
**To Reproduce**

(Write your steps here:)

1. Create a project containing `import OpenSSL.crypto`
2. `mypy --install-types`

**Expected Behavior**

Does not suggest installing library stubs incompatible with python version.

**Actual Behavior**

Suggests (and installs) stubs incompatible with the current version.

**Your Environment**

- Mypy version used: 0.901
- Mypy command-line flags: none
- Python version used: 3.8, 3.9
- Operating system and version: Debian

JukkaL:
Keep track of supported Python versions of legacy bundled packages, and
only suggest a stub package if the major Python version is compatible.

Fixes #10602.


ilevkivskyi (approved):
LG

JelleZijlstra (approved):

PR #10615, opened 2021-06-10
Add boto3 and botocore legacy stubs

vemel:
### Description

Adds `types-botocore` and `types-boto3` stub packages for `botocore` and `boto3` respectively.

This is a great idea to install missing type annotations with `mypy`. thank you!


JukkaL:
Thanks for the PR! This seems like a good idea. 

However, we are still figuring out the best way to suggest stubs from outside of typeshed (see #10619). If/when we decide on how to deal with adding suggested stubs generally, we can add boto3 and botocore stubs to suggestions.

vemel:
@JukkaL thank you for a quick response! Probably, `mypy` can suggest to install `types-<library>` or `<library>-stubs` automatically if they exist in PyPI and have `Typing :: Typed` classifier.

I am not sure if it should be done on a regular `mypy` run, p[probably only with `--install-types` flag.

I will keep the PR open for now.

JelleZijlstra:
@vemel that's too risky because anyone can upload things to PyPI, and installing packages can execute arbitrary code. In addition, low-quality stubs packages can lead to more problems than they solve. We need some sort of vetting.

vemel:
Hello!

Any update on this?

PR #10624, opened 2021-06-10
stubgen: Use PEP 604 instead of Optional

srittau:
### Description

stubgen: Use PEP 604 syntax instead of `Optional`.

## Test Plan

I changed the existing unit tests.


JelleZijlstra (approved):

JelleZijlstra:
Looks like stubgen will only generate this in function parameters though, where it should be safe.

PR #10625, opened 2021-06-10
stubgen: Use NamedTuple class syntax

srittau:
### Description

Use `typing.NamedTuple` in stubgen, instead of `collections.namedtuple`.

## Test Plan

I modified the existing tests and added another test for an empty namedtuple, to ensure that `Any` is not imported unnecessarily.

### DIFF
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -900,7 +900,6 @@ class StubGenerator(mypy.traverser.TraverserVisitor):
     def process_namedtuple(self, lvalue: NameExpr, rvalue: CallExpr) -> None:
         if self._state != EMPTY:
             self.add('\n')
-        name = repr(getattr(rvalue.args[0], 'value', ERROR_MARKER))
         if isinstance(rvalue.args[1], StrExpr):
             items = rvalue.args[1].value.split(" ")
         elif isinstance(rvalue.args[1], (ListExpr, TupleExpr)):
--- a/mypy/stubgen.py
+++ b/mypy/stubgen.py
@@ -912,7 +912,7 @@ class StubGenerator(mypy.traverser.TraverserVisitor):
         self.import_tracker.require_name('NamedTuple')
         self.add('{}class {}(NamedTuple):'.format(self._indent, lvalue.name))
         if len(items) == 0:
-            self.add('    ...\n')
+            self.add(' ...\n')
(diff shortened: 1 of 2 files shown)


mypy/stubgen.py
```diff
@@ -900,18 +900,24 @@ def is_namedtuple(self, expr: Expression) -> bool:
...
         else:
             self.add('%s%s: Any' % (self._indent, lvalue.name))
             self.import_tracker.require_name('Any')
             return
-        self.import_tracker.require_name('namedtuple')
-        self.add('%s%s = namedtuple(%s, %s)\n' % (self._indent, lvalue.name, name, items))
+        self.import_tracker.require_name('NamedTuple')
+        self.add('{}class {}(NamedTuple):'.format(self._indent, lvalue.name))
+        if len(items) == 0:
+            self.add(' ...\n')
```
JelleZijlstra, line 915:
```suggestion
            self.add('    ...\n')
```
Or maybe this should be an assertion failure? Empty NamedTuples aren't meaningful.

srittau: Not meaningful, but legal. Someone could be using it as a placeholder for something. But I can raise an assertion if you prefer.

JelleZijlstra: That's true. Let's just leave it be in that case and fix the indentation.

mypy/stubgen.py
```diff
@@ -900,18 +900,24 @@ def is_namedtuple(self, expr: Expression) -> bool:
...
         else:
             self.add('%s%s: Any' % (self._indent, lvalue.name))
             self.import_tracker.require_name('Any')
             return
-        self.import_tracker.require_name('namedtuple')
-        self.add('%s%s = namedtuple(%s, %s)\n' % (self._indent, lvalue.name, name, items))
+        self.import_tracker.require_name('NamedTuple')
+        self.add('{}class {}(NamedTuple):'.format(self._indent, lvalue.name))
+        if len(items) == 0:
+            self.add('    ...\n')
```
JelleZijlstra, line 915: Actually, I just realized my spacing change was wrong because this is on the same line as the `:`. Sorry for the noise.
```suggestion
            self.add(' ...\n')
```

JelleZijlstra (approved):

PR #10626, opened 2021-06-10
stubgen: Don't annotate unknown argument and return types

srittau:
### Description

Don't annotate unknown argument and return types with explicit `Any`. Also, fix formatting of functions that have an explicit type and default argument.

## Test Plan

I adapted the existing tests. I also added tests to ensure that explicit `Any`s remain in the stub. I also added a test for the formatting fix.


mypy/stubgen.py
```diff
@@ -624,26 +624,25 @@ def visit_func_def(self, o: FuncDef, is_abstract: bool = False,
...
-                    and not is_cls_arg):
-                self.add_typing_import("Any")
-                annotation = ": {}".format(self.typing_name("Any"))
-            elif annotated_type and not is_self_arg and not is_cls_arg:
-                annotation = ": {}".format(self.print_annotation(annotated_type))
-            else:
-                annotation = ""
+            annotation = ""
+            if annotated_type and not is_self_arg and not is_cls_arg:
+                print(annotated_type, get_proper_type(annotated_type))
```
JelleZijlstra, line 629: remove
```suggestion
```

srittau: Oops, now the whole world can see that I use printf debugging. :P

JelleZijlstra (approved):

mypy/stubgen.py
JelleZijlstra: I think everyone does it :)

PR #10627, opened 2021-06-11
Add tests for singledispatch

pranavrajpal:
This adds mypy tests for singledispatch, most of which are either based on the use of singledispatch in stubtest.py or seem like they would be common errors. All of these tests are currently skipped because mypy unsurprisingly doesn't catch these errors yet due to the lack of singledispatch support.

I ended up adding some new error messages for some tests, but I'm not intending for those to be the final error messages that we actually end up using. I just wanted to put something for those errors as a stand-in that we can replace later if anyone has any other preferences.

### DIFF
--- a/mypy/test/testcheck.py
+++ b/mypy/test/testcheck.py
@@ -95,6 +95,7 @@ typecheck_files = [
     'check-generic-alias.test',
     'check-typeguard.test',
     'check-functools.test',
+    'check-singledispatch.test',
 ]
 
 # Tests that use Python 3.8-only AST features (like expression-scoped ignores):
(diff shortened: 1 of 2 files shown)


msullivan (approved):
This looks pretty good; switch it to the new xfail and then we can do it

sobolevn:
Related: I am working on `singledispatch` on steroids: https://github.com/dry-python/classes/

It supports even generic functions with type restrictions.
Examples: https://github.com/dry-python/classes/tree/master/typesafety/test_typeclass

It might be useful while working on `singledispatch` support.

msullivan (approved):

PR #10635, opened 2021-06-12
Add support for using xfail in test cases

### LINKED ISSUE #10604: Allow marking test cases in .test files with -xfail
**Feature**

Allow marking test cases in *.test files as xfails by adding -xfail to the end of the test name.

**Pitch**

If someone is adding tests for a feature that hasn't been implemented yet, it's useful to be able to mark it as an xfail so that:
- the test failing doesn't fail the entire test suite
- the test is still run so that people can easily notice if the test starts passing as they work on that feature
    
Adding -skip to the test name achieves the first part, but it's easy to accidentally forget about a test and keep skipping it even after the feature has been implemented if the test is not run regularly.

This came up when writing some mypyc run tests while working on singledispatch support for mypyc, where it was first mentioned in https://github.com/python/mypy/pull/10589#issuecomment-856520979.

pranavrajpal:
Closes #10604

This adds support for marking test cases in *.test files as [xfails](https://docs.pytest.org/en/latest/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) by adding `-xfail` to the end of the test case name.

I wasn't sure if I should turn on [`xfail_strict`](https://docs.pytest.org/en/latest/how-to/skipping.html#strict-parameter) in the pytest config. Turning that on would probably make it less likely that we would end up with tests that are still marked as xfails long after they're fixed, but it might be annoying have the test suite fail because a test marked as xfail started passing.

## Test Plan

I couldn't find a way to write tests for this change, so I changed some existing tests by marking them as xfail and made sure they were treated as xfails when they failed, and xpasses when they succeeded.

### DIFF
--- a/mypy/test/data.py
+++ b/mypy/test/data.py
@@ -223,6 +223,7 @@ class DataDrivenTestCase(pytest.Item):
                  only_when: str,
                  platform: Optional[str],
                  skip: bool,
+                 xfail: bool,
                  data: str,
                  line: int) -> None:
         super().__init__(name, parent)
@@ -234,6 +235,7 @@ class DataDrivenTestCase(pytest.Item):
                 or (platform == 'posix' and sys.platform == 'win32')):
             skip = True
         self.skip = skip
+        self.xfail = xfail
         self.data = data
         self.line = line
         self.old_cwd = None  # type: Optional[str]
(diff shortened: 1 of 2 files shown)


msullivan:
Thanks for doing this!

Let's try xfail_strict for now but we'll drop it as soon as it annoys us, I figure

PR #10641, opened 2021-06-14
[mypyc] Raise ImportError instead of AttributeError when from-import fails

97littleleaf11:
### Description
Fixes https://github.com/mypyc/mypyc/issues/707

### DIFF
--- a/mypyc/lib-rt/misc_ops.c
+++ b/mypyc/lib-rt/misc_ops.c
@@ -647,10 +647,9 @@ CPy_Super(PyObject *builtins, PyObject *self) {
 // This helper function is a simplification of cpython/ceval.c/import_from()
 PyObject *CPyImport_ImportFrom(PyObject *module, PyObject *package_name,
                                PyObject *import_name, PyObject *as_name) {
-
-    PyObject *x;
     // check if the imported module has an attribute by that name
-    if (_PyObject_LookupAttr(module, import_name, &x) == 0) {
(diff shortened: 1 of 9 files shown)


msullivan (commented):
This looks good, thank you! Could you fix the ops test that fail (passing `--update-data` to pytest will automatically update them) and add a run test that tests the behavior, and we should be able to merge it.

97littleleaf11:
@msullivan Thanks for your review! I will fix those tests later.

mypyc/lib-rt/misc_ops.c
```diff
@@ -643,3 +643,46 @@ CPy_Super(PyObject *builtins, PyObject *self) {
...
+    // check if the imported module has an attribute by that name
+    if (_PyObject_LookupAttr(module, import_name, &x) == 0) {
+        // if not, attempt to import a submodule with that name
+        PyObject *fullmodname = PyUnicode_FromFormat("%U.%U", package_name, import_name);
+        if (fullmodname == NULL) {
+            goto fail;
+        }
+
+        // The following code is a simplification of cpython/import.c/PyImport_GetModule()
+        Py_INCREF(module);
```
JukkaL, line 661: This incref and the decref below seem redundant, since the caller will continue to hold a reference to these.

JukkaL (approved):
Thanks, looks good now!

JukkaL:
Actually, this may be missing a test. You should validate that you get `ImportError` on missing attribute.

PR #10642, opened 2021-06-14
Add Python 3.9 support in logging.basicConfig

joooeey:
links to relevant docs: https://docs.python.org/3/library/logging.html#logging.basicConfig https://docs.python.org/3/library/functions.html#open

I glanced over them.

### Description

In Python 3.9 the two new keyword arguments `encoding` and `errors` were added to `logging.basicConfig`. This PR reflects that change.

## Test Plan


JelleZijlstra:
Thanks for your contribution! This change should be made in https://github.com/python/typeshed, not directly in mypy.

joooeey:
I proposed this change in typeshed: https://github.com/python/typeshed/pull/5639

PR #10652, opened 2021-06-16
Don't suggest to install stubs with py.typed files

srittau:
### Description

The pallets libraries are now bundled with a `py.typed` file. Suggesting to install these stubs only leads to confusion. See python/typeshed#5423 and python/typeshed#5642 for context.


srittau:
I've added cryptography and pyjwt (which was misnamed before) to the list.

JelleZijlstra:
Ideally we should still suggest these stubs if users are using an old version of Click that doesn't have py.typed yet. But that's probably too complicated to make sense.

cjolowicz:
> Ideally we should still suggest these stubs if users are using an old version of Click that doesn't have py.typed yet. But that's probably too complicated to make sense.

The stubs are not suggested if click >= 8.0 is installed:

```sh
❯ venv/bin/mypy <(echo "import click")

/dev/fd/63:1: error: Library stubs not installed for "click" (or incompatible with Python 3.9)
/dev/fd/63:1: note: Hint: "python3 -m pip install types-click"
/dev/fd/63:1: note: (or run "mypy --install-types" to install all missing stub packages)
/dev/fd/63:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports

❯ venv/bin/pip install click 

Collecting click
  Using cached click-8.0.1-py3-none-any.whl (97 kB)
Installing collected packages: click
Successfully installed click-8.0.1

❯ venv/bin/mypy <(echo "import click")

```

But people who upgrade mypy before upgrading click can be misled by the warning.

JelleZijlstra:
Oh, nice! Then I don't see any problems with this PR. Perhaps mypy should also warn if it sees both types-click and click with a py.typed, but that's a different problem.

PR #10655, opened 2021-06-16
Address #10648

### LINKED ISSUE #10648: Docs: Postponed evaluation of annotations deferred to 3.11
**Documentation**

As per [the mailing list](https://mail.python.org/archives/list/python-dev@python.org/thread/CLVXXPQ2T2LQ5MP2Y53VVQFCXYWQJHKZ/), the postponed evaluation of annotations has been deferred to 3.11.
However, the docs still refer to it becoming default in 3.10: https://mypy.readthedocs.io/en/stable/runtime_troubles.html

It also seems like the current behaviour is that `mypy` considers all type annotations to be using postponed evaluation, even when missing the future statement.  For example, `mypy` reports no issues with the following code:

```python3
class X:
    def func(self) -> X:
        pass
```

As expected, it will not run without `from __future__ import annotations`.

Based on the title of the linked documentation page ("Annotation issues at runtime") and the opening sentence, it suggests that this might be intended behaviour? If so, can this please be added to the documentation? (If it's not intended behaviour, I can file a bug, but didn't want to open extra issues unnecessarily.)

hellocoldworld:
### Description

Postponed evaluation of annotations was deferred to 3.11 - update the docs.

Closes #10648

### DIFF
--- a/docs/source/runtime_troubles.rst
+++ b/docs/source/runtime_troubles.rst
@@ -9,7 +9,7 @@ and explains how to get your code running again. Generally speaking, we have
 three tools at our disposal:
 
 * For Python 3.7 through 3.9, use of ``from __future__ import annotations``
-  (:pep:`563`), made the default in Python 3.10 and later
+  (:pep:`563`), made the default in Python 3.11 and later
 * Use of string literal types or type comments
 * Use of ``typing.TYPE_CHECKING``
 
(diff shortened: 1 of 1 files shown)


JelleZijlstra (approved):

ethanhs (approved):
I believe we also might need to tweak mypy's behavior on this as well.

hauntsaninja:
Took a quick look, but didn't see any code that needed changing.

PR #10658, opened 2021-06-16
Narrow type with typevar when the lower bound of its upper bound and that type is same

### LINKED ISSUE #10605: regression: typevar with bound and type check causes narrowing to nothing
**To Reproduce**
```py
from typing import TypeVar, Type


class A:
    a: int = 1


T = TypeVar("T", bound=A)


def foo(t: Type[T], a: A) -> None:
    if type(a) is it:
        reveal_type(a) # <nothing>, with 0.8xx it was A
        a.a #  error: <nothing> has no attribute "a"
```
**Expected Behavior**

should not be narrowed to nothing, should be narrowed to A

**Your Environment**
- Mypy version used: 0.900 and 0.8xx
- Mypy command-line flags: --strict
- Mypy configuration options from `mypy.ini` (and other config files): no
- Python version used: 3.10
- Operating system and version: Windows 10

ethframe:
### Description

Fixes #10605 

This PR extends type narrowing by a type from a generic argument. This is achieved by handling a specific case while narrowing some type with `TypeVarType`: if the lower bound of that type and `upper_bound` of type variable is the same type as `upper_bound`, then the type can be narrowed to that type variable.

### DIFF
--- a/mypy/meet.py
+++ b/mypy/meet.py
@@ -7,6 +7,7 @@ from mypy.types import (
     UninhabitedType, TypeType, TypeOfAny, Overloaded, FunctionLike, LiteralType,
     ProperType, get_proper_type, get_proper_types, TypeAliasType, TypeGuardType
 )
+from mypy.sametypes import is_same_type
 from mypy.subtypes import is_equivalent, is_subtype, is_callable_compatible, is_proper_subtype
 from mypy.erasetype import erase_type
 from mypy.maptype import map_instance_to_supertype
@@ -77,6 +78,10 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
           and narrowed.type.is_metaclass()):
         # We'd need intersection types, so give up.
         return declared
+    elif isinstance(narrowed, TypeVarType):
+        if is_same_type(narrowed.upper_bound, meet_types(narrowed.upper_bound, declared)):
+            return narrowed
+        return meet_types(declared, narrowed)
     elif isinstance(declared, (Instance, TupleType, TypeType, LiteralType)):
         return meet_types(declared, narrowed)
     elif isinstance(declared, TypedDictType) and isinstance(narrowed, Instance):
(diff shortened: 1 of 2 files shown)


mypy/meet.py
```diff
@@ -77,6 +78,10 @@ def narrow_declared_type(declared: Type, narrowed: Type) -> Type:
           and narrowed.type.is_metaclass()):
         # We'd need intersection types, so give up.
         return declared
+    elif isinstance(narrowed, TypeVarType):
+        if is_same_type(narrowed.upper_bound, meet_types(narrowed.upper_bound, declared)):
```
msullivan, line 82: Does it work to check `is_subtype(narrowed.upper_bound, declared)` instead?

Mathematically, the two should be equivalent: x ≤ y if and only if x ∧ y = x (see https://en.wikipedia.org/wiki/Join_and_meet).

In practice, mypy's implementations fall short of mathematical perfection, and I think that `is_subtype` is more reliable than `meet_types`, and reads more naturally.

ethframe: The difference from `is_subtype` is how the `UnboundType` is treated. When type of `declared` is `UnboundType`, `is_subtype` will always return `True`. But since the type of `meet_type` result in this case is either `AnyType`, `NoneType` or `UninhabitedType`, `is_same_type` will result in `True` only if `narrowed.upper_bound` is also of `AnyType`, `NoneType`, `UninhabitedType` or `UnboundType`. I thought it is better not to unconditionally narrow unbound type with type variable.

msullivan: Hm. OK, that seems plausible. Do you have an example where we get different behavior between the two? I think that would make it easier for me to consider the difference.

ethframe: I tried to come up with an example and while I was skimming through `typeanal.py` to figure out how one can get an unbound type during type checking phase, I've learned about [this issue](https://github.com/python/mypy/issues/4987). Seems that there is no need to consider unbound types in the type checker, unless their presence leads to crashes. And since  `is_subtype` works fine with unbound types, I think it is actually a good idea to rewrite the entire PR code like this:

```python
elif isinstance(narrowed, TypeVarType) and is_subtype(narrowed.upper_bound, declared):
    return narrowed
```

msullivan: OK, I agree!

ethframe: Done.

msullivan:
Sorry for the delay, merging!

PR #10664, opened 2021-06-17
Add ignore missing type hints

### LINKED ISSUE #10660: Adding an --ignore-missing-type-hints config
**Feature**

We introduce a new configuration option, namely `--ignore-missing-type-hints` as a lighter alternative to `--ignore-missing-imports`. Such a config would be allowed in both the global level and to package-specific configs.

The config would simply disable "Skipping analyzing "offending.package": found module but no type hints or library stubs" error messages. More specifically, we'd be disabling errors on `ModuleNotFoundReason.FOUND_WITHOUT_TYPE_HINTS`.

**Pitch**

Having to `ignore-missing-imports` each individual package I import that hasn't built type hints is really frustrating. [Several of the notes here](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-type-hints-for-third-party-library) for resolving this error are largely non-actionable -- many libraries don't have types associated with them, nor stubs in typeshed.

There seem to be a few related issues, for example https://github.com/python/mypy/issues/4542, https://github.com/python/mypy/issues/9789 (thanks @hauntsaninja)

Given that a specific ModuleNotFoundReason exists, it makes me wonder why this flag hasn't been implemented yet -- is there a specific technical reason for it?

evinism:
### Description

Closes #10660

Draft PR that adds `--ignore-missing-type-hints`

## Test Plan

Will write automated tests after clarifying feature behavior, before un-drafting. Tested manually on one of my codebases.

### DIFF
--- a/mypy/build.py
+++ b/mypy/build.py
@@ -2443,6 +2443,7 @@ def find_module_and_diagnose(manager: BuildManager,
         # search path or the module has not been installed.
 
         ignore_missing_imports = options.ignore_missing_imports
+
         top_level, second_level = get_top_two_prefixes(file_id)
         # Don't honor a global (not per-module) ignore_missing_imports
         # setting for modules that used to have bundled stubs, as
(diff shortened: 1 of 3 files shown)


hauntsaninja:
Thanks for the PR, but closing in favour of https://github.com/python/mypy/pull/14740 (this PR is also not quite ready yet, e.g. merge conflicts and missing tests)

PR #10666, opened 2021-06-18
Check expr that are implicitly true for lack dunder bool/len

ikonst:
Optional check which disallows using implicitly true expressions (whose type has no `__bool__` nor `__len__`) in boolean context. The check is only enabled when using strict optionals.

This does not affect type narrowing / reachability which would continue seeing such types as "can be true / can be false". The rationale is that it's not an error per the data model to evaluate any object as boolean, and an always-true behavior could be overridden in a derived type.

Some builtin fixtures are updated to make sure common types (str, bool, int) are not considered to be always true.


mypy/types.py
```diff
@@ -842,6 +842,14 @@ def __init__(self, typ: mypy.nodes.TypeInfo, args: Sequence[Type],
         # Literal context.
         self.last_known_value = last_known_value
 
+        if (
+            self.type and
+            not self.type.has_readable_member('__bool__') and
```
ikonst, line 847: Do we have to check for Python 2 `__nonzero__`?

test-data/unit/check-expressions.test
```diff
@@ -375,8 +376,10 @@ if int():
     d = c or d # E: Incompatible types in assignment (expression has type "C", variable has type "D")
 if int():
     d = d or c # E: Incompatible types in assignment (expression has type "C", variable has type "D")
-class C: pass
-class D(C): pass
+class C:
+    def __bool__(self) -> bool: pass
+class D(C):
+    def __bool__(self) -> bool: pass
```
ikonst, line 382: Not needed

JukkaL:
> Teaches mypy to recognize instances that don't implement `__bool__` or `__len__` as always true.

Since subclasses can freely add `__bool__`, this doesn't sounds right. This is something we could perhaps infer for final classes, but the benefit would be marginal, since most users don't use `@final` much.

ikonst:
> Since subclasses can freely add __bool__, this doesn't sounds right.

I see that. Factoring such "knowledge" into unreachable code detection would be practically incorrect.

```python
def f(a: object) -> None:
   a.foobar()
   _ = len(a)
```
For the cases above it's easy to justify a type error since, while `foobar` and `__len__` could be added in a subclass, those operations would raise an error on object, while using object in boolean context is succeeds "blindly".

I wonder how it'll be practical to add such check, perhaps through an optional warning, i.e. make it more of a linting rule and not something that affects analysis?

I keep seeing this in real world code, e.g.
```python
def f(a: Iterable) -> None:
   if a:  # signature says Iterable but we're really expecting Collection and up
       ...
```
and while it's not technically incorrect, it's likely an error in typing (s/Iterable/Collection/).

Here's another favorite:
```python
class MyMessage(google.protobuf.Message):
    ...

def f(a: MyMessage) -> None:
    if a.sub_message:  # user assumes missing Protobuf field manifests as None but it's an eagerly created empty message
        ...
```
Per [Protobuf guidance](https://developers.google.com/protocol-buffers/docs/pythontutorial#parsing-and-serialization), one shouldn't inherit from messages, so those could indeed be marked `@final` (e.g. in [mypy-protobuf](https://github.com/dropbox/mypy-protobuf)) but I agree `@final` is not widely used and not very practical.

JukkaL:
This sounds like a good match for an optional check that can be activated via enabling some error code.  Also `--strict` could enable it. If it's an optional check, it should have no impact on type checking or type inference when disabled, but that seems doable.

ikonst:
@JukkaL Would it be OK to bolt this unto `find_isinstance_check_helper` like in 801f274b837a254557d2e39890d1de59b190744b? This checker seems to be already in the business of figuring out types in boolean context.

Also, is there anything that can print a node in a way that's short enough for messages, e.g.
```
note: NameExpr(m [l]) has type 'typing.Match[builtins.str*]' which does not implement __bool__ or __len__ so it will always be truthy in boolean context
```
Obviously I'd like something friendlier than `NameExpr(m [l])`. For some nodes getting a "name" is more obvious than others. Perhaps the source code of the node is the closest to what I'd expect. It's also possible that I'm worrying too much and that printing the type in context (line, column) is enough to explain what's going on.

hauntsaninja:
It's unprecedented (and in my eyes, undesirable) to print out a mypy node. Most mypy error messages just print out a type and it seems to be fine. If you want, consider special casing RefExpr, and not printing the node for anything else.

ikonst:
@JukkaL I've updated the code to only perform a check (conditioned an error code).

A few follow-ups:
* I'm not sure I understand where it's established which error codes are enabled/disabled by default. Is there something except for `default_enabled=False`?
* I'd appreciate some word-smithing of the messages, as well as the error-code itself.
* There are a handful of fixtures which define `class bool: pass` and `class int: pass`. Should I change them too?

ikonst:
Phew, and now passing on CI :)

ikonst:
@JukkaL gentle ping

ikonst:
@JukkaL ping

ikonst:
@hauntsaninja would you have time to review?

ikonst:
@hauntsaninja @JukkaL any suggestions?

I'd also love if I could offer this as a plugin, but we don't have a plugin interface but general purpose "linters" and semantic analyzers right now, even though I think having type information at hand allows a new class of such checkers to be easily implemented.

test-data/unit/check-enum.test
```diff
@@ -894,13 +894,13 @@ else:
 
 [case testEnumReachabilityWithNone]
 # flags: --strict-optional
-from enum import Enum
+from enum import Flag
```
hauntsaninja, line 897: Why do we need this change?

test-data/unit/check-errorcodes.test
```diff
@@ -807,3 +807,61 @@ from typing_extensions import TypedDict
 
 Foo = TypedDict("Bar", {})  # E: First argument "Bar" to TypedDict() does not match variable name "Foo"  [name-match]
 [builtins fixtures/dict.pyi]
+[case testImplicitBool]
+# flags: --enable-error-code implicit-bool
+from typing import List, Union
```
hauntsaninja, line 812: nit: could be good to reduce the whitespace in this test

test-data/unit/check-expressions.test
```diff
@@ -323,7 +323,8 @@ if int():
     b = b or a  # E: Incompatible types in assignment (expression has type "Union[bool, A]", variable has type "bool")
 if int():
     b = a or b  # E: Incompatible types in assignment (expression has type "Union[A, bool]", variable has type "bool")
-class A: pass
+class A:
```
hauntsaninja, line 326: Do we need these changes if the error code is disabled by default?

test-data/unit/fixtures/bool.pyi
```diff
@@ -10,9 +10,11 @@ class object:
 class type: pass
 class tuple(Generic[T]): pass
 class function: pass
-class int: pass
+class int:
```
hauntsaninja, line 13: Similarly, it seems like we only should need fixture changes in fixtures/list.pyi

hauntsaninja (commented):
Sorry for the delay, from a skim looks mostly reasonable.

I'm okay to merge this when disabled by default (if we were to enable by default I think this would need more justification, e.g., an analysis of mypy primer errors). Given disabled by default, would be good to undo changes to unrelated tests and fixtures.

test-data/unit/check-enum.test
ikonst: Hmm... let me think. I think it's because `Flag` implements `__bool__`, so with base `Enum` the following change would take place:
```diff
 if x:
     reveal_type(x)  # N: Revealed type is "__main__.Foo"
 else:
-     reveal_type(x)  # N: Revealed type is "Union[__main__.Foo, None]"
+     reveal_type(x)  # N: Revealed type is "None"
```
This is probably a vestige from my previous approach where lack of bool overload was taken as a signal that the class cannot be false-y. In this implementation I only emit a notice, but it doesn't affect always-true/false checks.

Long story short, I'll revert this change.

test-data/unit/check-expressions.test
ikonst: (Would you rather me not do this right now?)

hauntsaninja: Yes, I'd prefer not making changes not required. We'll make the changes when / if the error code is ever enabled by default.

test-data/unit/check-errorcodes.test
```diff
@@ -807,3 +807,48 @@ from typing_extensions import TypedDict
...
+if good_union:
+    pass
+
+bad_union: Union[Foo, object] = Foo()
+if bad_union:  # E: "__main__.bad_union" has type "Union[__main__.Foo, builtins.object]" none of which implement __bool__ or __len__ so it will always be true in boolean context  [implicit-bool]
+    pass
+
+def f():
+    pass
+if f:  # E: function "def () -> Any" will always be true in boolean context  [implicit-bool]
```
hauntsaninja, line 847:
```suggestion
if f:  # E: Function "def () -> Any" will always be true in boolean context  [implicit-bool]
```

test-data/unit/check-errorcodes.test
```diff
@@ -807,3 +807,48 @@ from typing_extensions import TypedDict
...
+    pass
+conditional_result = 'foo' if f else 'bar'  # E: function "def () -> Any" will always be true in boolean context  [implicit-bool]
```
hauntsaninja, line 849:
```suggestion
conditional_result = 'foo' if f else 'bar'  # E: Function "def () -> Any" will always be true in boolean context  [implicit-bool]
```

test-data/unit/check-errorcodes.test
```diff
@@ -807,3 +807,48 @@ from typing_extensions import TypedDict
...
+s = ''
+if s:
+    pass
+
+good_union: Union[str, int] = 5
+if good_union:
+    pass
+
+bad_union: Union[Foo, object] = Foo()
+if bad_union:  # E: "__main__.bad_union" has type "Union[__main__.Foo, builtins.object]" none of which implement __bool__ or __len__ so it will always be true in boolean context  [implicit-bool]
```
hauntsaninja, line 842:
```suggestion
if bad_union:  # E: "__main__.bad_union" has type "Union[__main__.Foo, builtins.object]" of which no members implement __bool__ or __len__ so it will always be true in boolean context  [implicit-bool]
```

ikonst:
Thanks for updating the copy @hauntsaninja.

hauntsaninja:
@ikonst Just made some more copy changes (including changing the error code name) and added some test case for `if not truthy_expression`. If it looks good to you, I'll merge.

My concerns with the phrasing:
1) The phrasing of "always" wasn't quite true, since subclasses could implement `__bool__`, so I tried to fudge that a little.
2) I don't know that "implicit bool" clearly describes the feature. I was also a little concerned about the potential for confusion with another requested mypy feature: to not treat bool as a subtype of int. Lmk if you have other suggestions.

hauntsaninja:
Oh actually, I think I like the error code "bool-conversion" for this.

ikonst:
Didn't yet review all changes, but "bool-conversion" rings like that "strict bool" feature we used to have. That is, it insinuates that relying on bool conversion is wrong perhaps even in cases where it's idiomatic Python.

docs/source/error_code_list2.rst
```diff
@@ -214,3 +214,20 @@ mypy generates an error if it thinks that an expression is redundant.
 
         # Error: If condition in comprehension is always true  [redundant-expr]
         [i for i in range(x) if isinstance(i, int)]
+
+
+Check that expression is not implicitly true in boolean context [truthy-bool]
+--------------------------------------------------------------------------------
+
+Warn when an expression whose type does not implement __bool__ or __len__ is used in boolean context
+since it could always evaluate to true (if one also assumes that no subtype implements __bool__ or
```
ikonst, line 223: "You should always take an umbrella since it could always rain" -- not exactly what we mean :)

Maybe something like "since by default objects evaluate as true"?

The thing with subtypes is correct but I wonder if we can position it differently? Essentially we're warning due to lack of type-based **guarantee** of meaningful bool. What would we say if the python data model would be defined in such a way that `len(object()) == 0)`? What would our warning, if any, be like?

ikonst:
@hauntsaninja ping :)

hauntsaninja:
Yeah, something like "since by default objects always evaluate to true (this warning assumes that no subtype implements __bool__ ..." seems good to me.
Feel free to make phrasing changes, or if you like it as is, let me know and I'll merge. I'm happy with most versions of the error code and docs now that we have the subtype thing mentioned.

hauntsaninja:
Thanks for making this happen!

PR #10668, opened 2021-06-18
[mypyc] Only construct default collection when needed

97littleleaf11:
### Description

Closes https://github.com/mypyc/mypyc/issues/830

* Add `dict_setdefault_spec_init_op`

## Test Plan

* Add new run tests for empty collection as defaults
* Add irbuild tests


TH3CHARLie:
Do you have a measured performance?

97littleleaf11:
running dict_set_default
..........
interpreted: 0.400383s (avg of 5 iterations; stdev 0.52%)
compiled:    0.108074s (avg of 5 iterations; stdev 0.63%)

compiled is 3.705x faster
..........

The current branch is 2.50x faster

mypyc/lib-rt/dict_ops.c
```diff
@@ -80,6 +80,31 @@ PyObject *CPyDict_SetDefaultWithNone(PyObject *dict, PyObject *key) {
     return CPyDict_SetDefault(dict, key, Py_None);
 }
 
+PyObject *CPyDict_SetDefaultWithEmptyDatatype(PyObject *dict, PyObject *key,
+                                              int data_type) {
+    PyObject *res = CPyDict_GetItem(dict, key);
+    if (!res) {
+        PyErr_Clear();
```
JukkaL, line 87: Do you have this because the function may return `NULL` without raising an exception in case `data_type` is unsupported? It would be cleaner to instead raise `RuntimeError` or similar instead of just returning `NULL`. At least it would be better to have this just before `return NULL;` to make it clearer why we are doing this.

mypyc/primitives/dict_ops.py
```diff
@@ -135,6 +135,13 @@
     is_borrowed=True,
     error_kind=ERR_MAGIC)
 
+# dict.setdefault(key, empty tuple/list/set)
```
JukkaL, line 138: Document the valid values for the third argument and what they mean.

JukkaL (commented):
Great, looks good! Just left a few minor comments.

mypyc/lib-rt/dict_ops.c
```diff
@@ -80,6 +80,34 @@ PyObject *CPyDict_SetDefaultWithNone(PyObject *dict, PyObject *key) {
     return CPyDict_SetDefault(dict, key, Py_None);
 }
 
+PyObject *CPyDict_SetDefaultWithEmptyDatatype(PyObject *dict, PyObject *key,
+                                              int data_type) {
+    PyObject *res = PyDict_GetItemWithError(dict, key);
```
JukkaL, line 85: `PyDict_` functions usually can't be used directly, since they don't work for all dict subclasses. So it's better to continue to use `CPyDict_GetItem` as before, and only add a comment about why we clear the error.

JukkaL (approved):
Thanks!

JukkaL:
The code looks fine but Travis CI is failing. Can you investigate?

97littleleaf11:
It seems to be a HTTPError:
```
Warning, treated as error:
failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://six.readthedocs.io/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 502 Server Error: Bad Gateway for url: https://six.readthedocs.io/objects.inv
ERROR: InvocationError for command /home/travis/build/python/mypy/.tox/docs/bin/sphinx-build -d /home/travis/build/python/mypy/.tox/docs_doctree docs/source /home/travis/build/python/mypy/.tox/docs_out --color -W -bhtml (exited with code 2)
```

JelleZijlstra:
I pushed the button restart that job. Only the docs build failed, so it seems unlikely to be related to this PR which doesn't touch the docs.

Issue #10665, opened 2021-06-17
Crash on assigning to guarded variable in failed TypeGuard branch


p7g:
The title is probably not fully accurate, but I'm unable to fully intuit what's going on.

**Crash Report**

I'm using a TypeGuard to narrow the type of a variable from a more general one (not Any). If I add a branch like `if not guard(var): var = ...` there's an AssertionError.

I suspect something weird happens to the type of the variable after the failed type guard.

**Traceback**

```
test4.py:10: error: INTERNAL ERROR -- Please try using mypy master on Github:                                                                                                                      
https://mypy.readthedocs.io/en/stable/common_issues.html#using-a-development-mypy-build
Please report a bug at https://github.com/python/mypy/issues                                                                                                                                       
version: 0.910+dev.7f54bd064399ee4e38586c311a2796ff3f255276
Traceback (most recent call last):
  File "/Users/patrick/.local/venvs/fellow/bin/mypy", line 8, in <module>
    sys.exit(console_entry())
  File "/Users/patrick/.local/venvs/fellow/lib/python3.8/site-packages/mypy/__main__.py", line 11, in console_entry
    main(None, sys.stdout, sys.stderr)
  File "/Users/patrick/.local/venvs/fellow/lib/python3.8/site-packages/mypy/main.py", line 103, in main
    res = build.build(sources, options, None, flush_errors, fscache, stdout, stderr)
  File "/Users/patrick/.local/venvs/fellow/lib/python3.8/site-packages/mypy/build.py", line 179, in build
    result = _build(
  File "/Users/patrick/.local/venvs/fellow/lib/python3.8/site-packages/mypy/build.py", line 254, in _build
    graph = dispatch(sources, manager, stdout)
  File "/Users/patrick/.local/venvs/fellow/lib/python3.8/site-packages/mypy/build.py", line 2696, in dispatch
    process_graph(graph, manager)
  File "/Users/patrick/.local/venvs/fellow/lib/python3.8/site-packages/mypy/build.py", line 3020, in process_graph
    process_stale_scc(graph, scc, manager)
  File "/Users/patrick/.local/venvs/fellow/lib/python3.8/site-packages/mypy/build.py", line 3118, in process_stale_scc
    graph[id].type_check_first_pass()
  File "/Users/patrick/.local/venvs/fellow/lib/python3.8/site-packages/mypy/build.py", line 2165, in type_check_first_pass
    self.type_checker().check_first_pass()
[...]

PR #10496, opened 2021-05-18
Fix crash on TypeGuard plus "and"

JelleZijlstra:
In python/typeshed#5473, I tried to switch a number of `inspect` functions to use the new `TypeGuard` functionality. Unfortunately, mypy-primer found a number of crashes in third-party libraries in places where a TypeGuard function was ANDed together with some other check. Examples:
- https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/util/inspect.py#L252
- https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/ext/coverage.py#L212
- https://github.com/streamlit/streamlit/blob/develop/lib/streamlit/elements/doc_string.py#L105

The problems trace back to the decision in #9865 to make TypeGuardType not inherit from ProperType: in various conditions that are more complicated than a simple `if` check, mypy wants everything to become a ProperType. Therefore, to fix the crashes I had to make TypeGuardType a ProperType and support it in various visitors.

PR #10671, opened 2021-06-19
Fix typeguards crash when assigning guarded value in if statement

A5rocks:
### Description

Fixes https://github.com/python/mypy/issues/10665

This PR implements joining type guard types (which can happen when leaving an if statement). (it continues along the lines of https://github.com/python/mypy/pull/10496)

## Test Plan

Added the crash case the test cases. They all passed locally, except some mypyd ones that complained about restarting, which makes no sense. Hopefully they work in GitHub actions.

### DIFF
--- a/mypy/join.py
+++ b/mypy/join.py
@@ -433,7 +433,7 @@ class TypeJoinVisitor(TypeVisitor[ProperType]):
         assert False, "This should be never called, got {}".format(t)
 
     def visit_type_guard_type(self, t: TypeGuardType) -> ProperType:
-        assert False, "This should be never called, got {}".format(t)
+        return t.type_guard.accept(self)
 
     def join(self, s: Type, t: Type) -> ProperType:
         return join_types(s, t)
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -4102,7 +4102,7 @@ class TypeChecker(NodeVisitor[None], CheckerPluginInterface):
                         self.fail("Type guard requires positional argument", node)
                         return {}, {}
                     if literal(expr) == LITERAL_TYPE:
-                        return {expr: TypeGuardType(node.callee.type_guard)}, {}
+                        return {expr: node.callee.type_guard}, {}
         elif isinstance(node, ComparisonExpr):
             # Step 1: Obtain the types of each operand and whether or not we can
             # narrow their types. (For example, we shouldn't try narrowing the
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -35,7 +35,7 @@ from mypy.types import (
     UnionType, TypeVarId, TypeVarType, PartialType, DeletedType, UninhabitedType, TypeVarDef,
     is_named_instance, union_items, TypeQuery, LiteralType,
     is_optional, remove_optional, TypeTranslator, StarType, get_proper_type, ProperType,
-    get_proper_types, is_literal_type, TypeAliasType, TypeGuardType)
+    get_proper_types, is_literal_type, TypeAliasType)
 from mypy.sametypes import is_same_type
 from mypy.messages import (
     MessageBuilder, make_inferred_type_note, append_invariance_notes, pretty_seq,
--- a/mypy/checker.py
+++ b/mypy/checker.py
@@ -35,7 +35,7 @@ from mypy.types import (
     UnionType, TypeVarId, TypeVarType, PartialType, DeletedType, UninhabitedType, TypeVarDef,
     is_named_instance, union_items, TypeQuery, LiteralType,
     is_optional, remove_optional, TypeTranslator, StarType, get_proper_type, ProperType,
-    get_proper_types, is_literal_type, TypeAliasType)
+    get_proper_types, is_literal_type, TypeAliasType, TypeGuardType)
 from mypy.sametypes import is_same_type
 from mypy.messages import (
     MessageBuilder, make_inferred_type_note, append_invariance_notes, pretty_seq,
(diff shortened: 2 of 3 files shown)


mypy/join.py
```diff
@@ -433,7 +433,7 @@ def visit_type_alias_type(self, t: TypeAliasType) -> ProperType:
         assert False, "This should be never called, got {}".format(t)
 
     def visit_type_guard_type(self, t: TypeGuardType) -> ProperType:
-        assert False, "This should be never called, got {}".format(t)
+        return t.type_guard.accept(self)
```
JelleZijlstra, line 436: I don't think this is right, since for `join(T, TypeGuard[U])` it will end up with `join(T, U)`, which is approximately `Union[T, U]` in practice. Computing `join(T, bool)` may make more sense.

I'm not sure what practical effect this has though. It's at least clear that we'll have to make sure we handle TypeGuard correctly in more places in the codebase.

A5rocks: The thing is, it's (in the example crash code) saying `a` is `TypeGuard[B]` -- when it's actually `B` (?). However, when I changed that (https://github.com/python/mypy/pull/10671/commits/3f068f5246b9cc5ef6d9f78a5875a8fc939d5809) some of the test cases started erroring... in specific, some of the narrowing ones:

```
_____________________________________________________________ testTypeGuardNonOverlapping _____________________________________________________________
[gw6] linux -- Python 3.8.2 /home/a5/.virtualenvs/tmp-7d419bcf0dbd427/bin/python
data: /home/a5/oss/mypy/test-data/unit/check-typeguard.test:78:
/home/a5/oss/mypy/mypy/test/testcheck.py:137: in run_case
    self.run_case_once(testcase)
/home/a5/oss/mypy/mypy/test/testcheck.py:236: in run_case_once
    assert_string_arrays_equal(output, a, msg.format(testcase.file, testcase.line))
E   AssertionError: Unexpected type checker output (/home/a5/oss/mypy/test-data/unit/check-typeguard.test, line 78)
---------------------------------------------------------------- Captured stderr call -----------------------------------------------------------------
Expected:
  main:6: note: Revealed type is "builtins.list[builtins.str]" (diff)
Actual:
  main:6: note: Revealed type is "None"         (diff)

  E: ...te: Revealed type is "builtins.list[builtins.str]"
  A: ...te: Revealed type is "None"
                              ^
______________________________________________________________ testTypeGuardNonzeroFloat ______________________________________________________________
[gw6] linux -- Python 3.8.2 /home/a5/.virtualenvs/tmp-7d419bcf0dbd427/bin/python
data: /home/a5/oss/mypy/test-data/unit/check-typeguard.test:105:
/home/a5/oss/mypy/mypy/test/testcheck.py:137: in run_case
    self.run_case_once(testcase)
/home/a5/oss/mypy/mypy/test/testcheck.py:236: in run_case_once
    assert_string_arrays_equal(output, a, msg.format(testcase.file, testcase.line))
E   AssertionError: Unexpected type checker output (/home/a5/oss/mypy/test-data/unit/check-typeguard.test, line 105)
---------------------------------------------------------------- Captured stderr call -----------------------------------------------------------------
Expected:
  main:5: note: Revealed type is "builtins.float" (diff)
Actual:
  main:5: note: Revealed type is "builtins.int" (diff)

  E: ...ed type is "builtins.float"
  A: ...ed type is "builtins.int"
                             ^
___________________________________________________________ testTypeGuardNarrowToTypedDict ____________________________________________________________
[gw6] linux -- Python 3.8.2 /home/a5/.virtualenvs/tmp-7d419bcf0dbd427/bin/python
data: /home/a5/oss/mypy/test-data/unit/check-typeguard.test:155:
/home/a5/oss/mypy/mypy/test/testcheck.py:137: in run_case
    self.run_case_once(testcase)
/home/a5/oss/mypy/mypy/test/testcheck.py:236: in run_case_once
    assert_string_arrays_equal(output, a, msg.format(testcase.file, testcase.line))
E   AssertionError: Unexpected type checker output (/home/a5/oss/mypy/test-data/unit/check-typeguard.test, line 155)
---------------------------------------------------------------- Captured stderr call -----------------------------------------------------------------
Expected:
  main:10: note: Revealed type is "TypedDict('__main__.User', {'name': builtins.str, 'id': builtins.int})" (diff)
Actual:
  main:10: note: Revealed type is "None"        (diff)

  E: ...ote: Revealed type is "TypedDict('__main__.User', {'name': builtins.s...
  A: ...ote: Revealed type is "None"...
                               ^
```

Because of that, I kind of assumed that `TypeGuard[T]` is just `T` -- I guess that isn't the case though? If so, then this might be more annoying to fix...

A5rocks:
Alright, so I'm somewhat confused on how to approach this. `TypeGuardType` seems to be used for *both* type guard functions (?) *and* type guarded variables. This makes sense since type guarded variables are a little special, but at the same time is massively confusing. See for example:

https://github.com/python/mypy/blob/790ab35ca936d04c484d427704d5acb17904012b/mypy/checkexpr.py#L4178-L4181

This was why my [initial approach](https://github.com/python/mypy/pull/10671/commits/3f068f5246b9cc5ef6d9f78a5875a8fc939d5809) of saying type guarded variables were the same type as their actual type didn't work, because it would not *just* take the new type, it would instead apply some narrowing.

However, if a `TypeGuard[T]` is a `TypeGuardType(T)`, then something has to change. I'm unsure on how to resolve this -- maybe I could add a boolean to `TypeGuardType` saying whether it is a guarded value? Maybe make an entirely new `ProperType` (which seems like a pain to do)? I'd love a bit of guidance!

JukkaL:
I'm pretty sure that we shouldn't leak type guard types into various type operations. If we try to join a type guard type, the bug is probably elsewhere.

JukkaL:
FWIW, #10683 fixes the crash in a simple but kind of hacky way. A better understanding of what's going on would be nice. Maybe the binder should handle type guard types in a different way.

PR #10675, opened 2021-06-20
Call get_method_hook when methods are used in decorators

pranavrajpal:
Currently, when methods are used in decorators, instead of correctly calling `get_method_hook` with `'{module_name}.{class_name}.{method_name}'` (with `module_name`, `class_name`, and `method_name` filled in with their correct values), `get_function_hook` is called instead with `'{method_name} of {class_name}'`. This PR fixes this bug by calling `get_method_hook` correctly.

Technically, this may break some plugins if they relied on the fact that `get_function_hook` would be called on methods used in decorators, because with this PR those calls get moved to `get_method_hook`. I'm guessing that this is rare, but I don't know for sure if any plugins rely on that.

Currently, this only works when you access the method in the decorator itself, instead of assigning the method to a local variable. So, for example (assuming `obj` is an object that has a method `meth`), this works:

```python
@obj.meth
def f() -> None:
    pass
```

but this doesn't:

```python
method = obj.meth
@method
def f() -> None:
    pass
```

Currently, the second case causes mypy to pass `__main__.method` to `get_function_hook` instead of passing the same thing as the first case. I added a test that's currently marked as xfail for the second case, but I'm not really sure how to fix it.

## Test Plan

I added a test to check-custom-plugin.test to make sure that the correct method name is getting passed to plugins and that it is getting passed to the correct plugin hook.

### DIFF
--- a/test-data/unit/check-custom-plugin.test
+++ b/test-data/unit/check-custom-plugin.test
@@ -810,3 +810,31 @@ reveal_type(f())  # N: Revealed type is "builtins.str"
 [file mypy.ini]
 \[mypy]
 plugins=<ROOT>/test-data/unit/plugins/method_in_decorator.py
+
+[case testPluginMethodHookMethodAssignedToVariableInDecorator-xfail]
+# flags: --config-file tmp/mypy.ini
+from typing import overload, TypeVar, Callable
+
+T = TypeVar('T')
+class Foo:
+    @overload
+    def a(self, x: Callable[[], T]) -> Callable[[], T]: ...
+
+    @overload
+    def a(self, x: str) -> None: ...
+
+    def a(self, x):
+        pass
+
+b = Foo()
+
+meth = b.a
+@meth
+def f() -> None:
+    pass
+
+reveal_type(f())  # N: Revealed type is "builtins.str"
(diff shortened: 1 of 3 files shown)


test-data/unit/check-custom-plugin.test
```diff
@@ -783,3 +783,44 @@ reveal_type(dynamic_signature(1))  # N: Revealed type is "builtins.int"
...
+def f() -> None:
+    pass
+
+reveal_type(f())  # N: Revealed type is "builtins.str"
+
+[file mypy.ini]
+\[mypy]
+plugins=<ROOT>/test-data/unit/plugins/method_in_decorator.py
+
+[case testPluginMethodHookMethodAssignedToVariableInDecorator-xfail]
```
msullivan, line 807: I would drop this test, it's fine.

msullivan (approved):
Looks great, thanks! Drop the xfailed test, and then I can merge it

test-data/unit/check-custom-plugin.test
pranavrajpal: Ok, I have removed that test.

PR #10676, opened 2021-06-20
fix testSemanticAnalysisFalseButTypeNarrowingTrue

ikonst:
It was indeed weird that `Union[builtins.bool, Literal[True]]` wasn't simplified by `make_simplified_union` to `builtins.bool`. This was simply because `builtins.bool` wasn't defined in test-data/unit/lib-stub/builtins.pyi.

It's important to keep fixtures nimble, but this is too much of a gotcha IMO not to have `bool` defined. (Also, I intend to define it in a follow-up PR anyway so it could have a `__bool__` method.)

### DIFF
--- a/test-data/unit/check-unreachable-code.test
+++ b/test-data/unit/check-unreachable-code.test
@@ -619,23 +619,26 @@ if MYPY or mypy_only:
 # flags: --always-false COMPILE_TIME_FALSE
 from typing import Literal
 
-indeterminate: bool
+indeterminate: str
 COMPILE_TIME_FALSE: Literal[True]  # type-narrowing: mapped in 'if' only
 a = COMPILE_TIME_FALSE or indeterminate
-reveal_type(a)  # N: Revealed type is "builtins.bool"
+reveal_type(a)  # N: Revealed type is "builtins.str"
 b = indeterminate or COMPILE_TIME_FALSE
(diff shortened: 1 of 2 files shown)


test-data/unit/check-unreachable-code.test
```diff
@@ -619,23 +619,26 @@ if MYPY or mypy_only:
...
 [typing fixtures/typing-medium.pyi]
 
 [case testSemanticAnalysisTrueButTypeNarrowingFalse]
 # flags: --always-true COMPILE_TIME_TRUE
-indeterminate: bool
-COMPILE_TIME_TRUE: None  # type narrowed to `else` only
+from typing import Literal
+
+indeterminate: str
+COMPILE_TIME_TRUE: Literal[False]  # type narrowed to `else` only
```
ikonst, line 635: Changing `None` to `Literal[False]` to make it symmetric to `testSemanticAnalysisFalseButTypeNarrowingTrue`

test-data/unit/check-unreachable-code.test
```diff
@@ -619,23 +619,26 @@ if MYPY or mypy_only:
 # flags: --always-false COMPILE_TIME_FALSE
 from typing import Literal
 
-indeterminate: bool
+indeterminate: str
```
ikonst, line 622: `str` both can be true and false since it has `__len__`

msullivan (approved):
Sure. We can add bool to the fixtures.

PR #10679, opened 2021-06-20
[mypyc] Support various number-related dunders

JukkaL:
This adds support for these unary dunders:
* `__neg__`
* `__invert__`
* `__int__`
* `__float__`

Also add support for binary, reversible dunders, such as `__add__` and `__radd__`.

Finally, add support for in-place operator dunders such as `__iadd__`.

The semantics of the binary dunders don't always match Python semantics, but
many common use cases should work.

There is one significant difference from Python that is not easy to remove: if a
forward dunder method is called with an incompatible argument, it's treated the
same as if it returned `NotImplemented`. This is necessary since the body of
the method is never reached on incompatible argument type and there is no
way to explicitly return `NotImplemented`. However, it's still recommended that 
the body returns `NotImplemented` as expected for Python compatibility.

If a dunder returns `NotImplemented` and has a type annotation, the return
type should be annotated as `Union[T, Any]`, where `T` is the return value 
when `NotImplemented` is not returned.

Work on mypyc/mypyc#839.


mypyc/codegen/emitwrapper.py
```diff
@@ -259,16 +261,126 @@ def generate_dunder_wrapper(cl: ClassIR, fn: FuncIR, emitter: Emitter) -> str:
...
+
+    Both arguments and the return value are PyObject *.
+    """
+    gen = WrapperGenerator(cl, emitter)
+    gen.set_target(fn)
+    gen.arg_names = ['left', 'right']
+    wrapper_name = gen.wrapper_name()
+
+    gen.emit_header()
+    if fn.name not in reverse_op_methods and fn.name in reverse_op_methods.values():
```
msullivan, line 287: It doesn't really matter but let's make a set in operators with the results of `reverse_op_methods.values()` so we don't need to scan

mypyc/codegen/emitwrapper.py
```diff
@@ -574,18 +691,30 @@ def generate_wrapper_core(fn: FuncIR, emitter: Emitter,
...
     The object is named PyObject *obj_{}. This is expected to generate
     a value of name arg_{} (unboxed if necessary). For each primitive a runtime
     check ensures the correct type.
     """
+    error = error or AssignHandler()
     if typ.is_unboxed:
         # Borrow when unboxing to avoid reference count manipulation.
-        emitter.emit_unbox('obj_{}'.format(name), 'arg_{}'.format(name), typ,
-                           error_code, declare_dest=True, borrow=True, optional=optional)
+        emitter.emit_unbox('obj_{}'.format(name),
```
msullivan, line 710: I am not asking for a change, but just making a remark: I've really come around to "black-style" for multi-line function calls, where you put the first argument on a new line, indented four spaces

mypyc/codegen/emitwrapper.py
```diff
@@ -596,10 +725,129 @@ def generate_arg_check(name: str, typ: RType, emitter: Emitter,
...
+        emitter.emit_cast('obj_{}'.format(name),
+                          'arg_{}'.format(name),
+                          typ,
+                          declare_dest=True,
+                          raise_exception=raise_exception,
+                          error=error,
+                          optional=optional)
+
+
+class WrapperGenerator:
```
msullivan, line 737: I like this

msullivan (approved):
Great!

PR #10681, opened 2021-06-21
pkg_resources is now in types-setuptools

srittau:
Depends on python/typeshed#5669


JukkaL (approved):
I just merged the typeshed commit that this depends on. This can be merged once `types-setuptools` is available on PyPI.

JukkaL:
`types-setuptools` is now on PyPI: https://pypi.org/project/types-setuptools/

PR #10687, opened 2021-06-21
[mypyc] Fix suffix assignment when a module contains a submodule with the same name.

DragonMinded:
## Description

Code that attempts to assign the shortest suffix for the C name of each module currently trips up if you have a module with a submodule that shares the same name. For instance, if you have a module "test" and a submodule "test.test", and run `mypyc test/` instead of compiling you will get a cryptic error message. It appears that this is because of a logic bug that's attempting to assign non-ambiguous suffixes. It sees that "test." is ambiguous, but it is also the only valid choice in this case. So, fix that by allowing the somewhat ambiguous suffix "test." to be given to the "test" module instead of the "test.test" module, fixing the assertion and allowing the code to compile.

This fixes a bug that I noticed and mentioned in https://github.com/mypyc/mypyc/issues/857 but did not open a separate issue for.

## Test Plan

Tested on the repro from https://github.com/mypyc/mypyc/issues/857 to verify that the code compiles. Also, added a new test that failed without the fix and succeeded with the fix applied. Also, all other tests continue to pass and there looks to be decent coverage of the file in question. You can verify by running the mypyc tests.

## Formatting

I ran `black -S` over the changed files and it looks like it changed more than just the code I added, so I reverted the changes black made to sections that I didn't touch.

### DIFF
--- a/mypyc/namegen.py
+++ b/mypyc/namegen.py
@@ -95,13 +95,47 @@ def make_module_translation_map(names: List[str]) -> Dict[str, str]:
         for suffix in candidate_suffixes(name):
             num_instances[suffix] = num_instances.get(suffix, 0) + 1
     result = {}
-    for name in names:
+    possibly_ambiguous_suffixes = set()  # type: Set[str]
+    # Find suffixes from logest to shortest, so that in the case of module
+    # names like "bla" and "bla.bla", we will not prematurely claim the
+    # "bla." suffix for the "bla.bla" module, leaving no candidates for
+    # the "bla" module.
+    for name in sorted(names, key=lambda name: len(name), reverse=True):
+        # First, try to find a suffix, preferring those that are not possibly
+        # ambiguous given other selections that were already made. If we have
+        # "foo.bar" and "biz.bar", we would prefer to return suffixes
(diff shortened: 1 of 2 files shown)


msullivan (commented):
Thanks for looking at this!

This looks like a solid fix, but I think it might be possible to solve the problem a little more simply: I think it might work to just always take the fullname (plus a dot) as the suffix in the else case?

DragonMinded:
Probably, yeah. However, I tried to preserve the existing functionality since a comment in this module claimed that it was trying to keep prefixes short for debugging readability. I wasn't sure if anyone else was relying on that functionality, or lengthening the prefix would cause headaches elsewhere. It certainly failed a few other tests and I wasn't sure if they were just there to verify functionality or be resistance against breaking the current guarantees.

msullivan:
Sure, that makes sense. It's possible @JukkaL disagrees with me here, but I think that I would prioritize simplicity over maintaing the current guarantees

97littleleaf11:
I think we may still keep suffix as short as possible.

An easy-to-read implementation might be construct a suffix tree (use Dict[str, Node] to store sons at each node). Then we can easily assign different names to each module after one traversal along the tree.

PR #10690, opened 2021-06-22
Replace variable annotation with type comment to fix Python 3.5

JukkaL:


JukkaL:
I skipped waiting for CI since wheel builds are broken and the change is trivial.

